PR #13399, opened 2024-01-10
docs: socket: Note that socket.getaddrinfo() doesn't take kwargs.

### LINKED ISSUE #8039: ESP32: getaddrinfo() does not conform to documentation
Micropython v1.17, compiled with ESP-IDF v4.3

While trying #8038, I read the documentation of getaddrinfo() at https://docs.micropython.org/en/latest/library/socket.html?highlight=socket#socket.getaddrinfo

There I find
``` 
socket.getaddrinfo(host, port, af=0, type=0, proto=0, flags=0, /)
```
However, trying this at home gives me
```
>>> getaddrinfo('eris', 80, type=SOCK_DGRAM)
TypeError: function doesn't take keyword arguments
```

projectgus:
Closes #8039.

I'm not sure if it's viable to note this for every MP function where CPython takes kwargs and MP doesn't, but this is one where a lot of existing sample code is likely to pass kwargs.

This work was funded through GitHub Sponsors.

### DIFF
--- a/docs/library/socket.rst
+++ b/docs/library/socket.rst
@@ -106,6 +106,12 @@ Functions
       from an exception object). The use of negative values is a provisional
       detail which may change in the future.
 
+   .. admonition:: Difference to CPython
+      :class: attention
+
+      The optional arguments to this function are not keyword arguments and must
+      be passed as positional arguments.
+
 .. function:: inet_ntop(af, bin_addr)
 
    Convert a binary network address *bin_addr* of the given address family *af*


dpgeorge:
I'm not sure this is worth adding, because the `/` at the end of the argument list means that it only accepts positional arguments.

Yes, I understand that's not helpful for people who don't know the syntax, but it would be better to teach that syntax rather than add notes like this everywhere.

See https://peps.python.org/pep-0570/

projectgus:
@dpgeorge Oh, I didn't know that either! Yes, I agree.

I'll submit a PR for the "Differences from CPython" docs page, instead.

PR #8050, opened 2021-11-30
py/mpprint.h: Add  MP_DEBUG_PRINT() macro and  MP_DEBUG_PRINT_LEVEL.

IhorNehrutsa:
Debugging macro for developers.

How to use:
```
// Important! Set MP_DEBUG_PRINT_LEVEL in *.c or *.cpp development file BEFORE any "MicroPython's *.h" includes.
// For example:
#define MP_DEBUG_PRINT_LEVEL MP_DEBUG_PRINT_TRACE // show all messages
// Include mpprint.h after defining the MP_DEBUG_PRINT_LEVEL
#include "py/mpprint.h"
...
#include "py/obj.h"
#include "py/runtime.h"
...
// Add MP_DEBUG_PRINT() macro in code, like
void foo(int arg) {
    MP_DEBUG_PRINT(MP_DEBUG_PRINT_TRACE, "Enter foo()")
    if (arg < 0) {
        MP_DEBUG_PRINT(MP_DEBUG_PRINT_WARNING, "arg=%d less zero", arg)
        ...
    }
    ...
    int value;
    ...
    // calculate value
    ...
    MP_DEBUG_PRINT(MP_DEBUG_PRINT_INFO, "See a value=%d", value)
    ...
    MP_DEBUG_PRINT(MP_DEBUG_PRINT_TRACE, "Exit foo()")
}

// It is not a dogma. You may start debugging from level 30.
#define MP_DEBUG_PRINT_LEVEL 30
// Then add MP_DEBUG_PRINT(30, ...) and when gets too many messages then change some messages to the next level MP_DEBUG_PRINT(40, ...), or MP_DEBUG_PRINT(20, ...) etc.
// Then you may change MP_DEBUG_PRINT_LEVEL to 20(reduce printing), and finally to 0(suppress printing).

// Usually, you will debug one or two source files. Debug printing from other files is suppressed if MP_DEBUG_PRINT_LEVEL is 0 or undefined.
```

PR #13400, opened 2024-01-10
py/mpprint.h: Add do_printf() macro.

IhorNehrutsa:
Move `do_printf()` from `drivers/esp-hosted/esp_hosted_hal.h` to `py/mpprint.h`

### DIFF
--- a/drivers/esp-hosted/esp_hosted_hal.h
+++ b/drivers/esp-hosted/esp_hosted_hal.h
@@ -37,28 +37,6 @@
 #define PROTOBUF_C_UNPACK_ERROR(...)    error_printf(__VA_ARGS__);
 #endif
 
-#define ANSI_C_RED          "\x1B[31m"
-#define ANSI_C_GREEN        "\x1B[32m"
-#define ANSI_C_YELLOW       "\x1B[33m"
-#define ANSI_C_BLUE         "\x1B[34m"
-#define ANSI_C_MAGENTA      "\x1B[35m"
-#define ANSI_C_CYAN         "\x1B[36m"
-#define ANSI_C_WHITE        "\x1B[37m"
-#define ANSI_C_DEFAULT      "\x1B[0m"
-
-#if ESP_HOSTED_DEBUG
-#define do_printf(...)      mp_printf(&mp_plat_print, __VA_ARGS__)
-#else
-#define do_printf(...)
-#endif
-
-// Logging macros.
-#define debug_printf(...)   do_printf(ANSI_C_BLUE); do_printf(__VA_ARGS__); do_printf(ANSI_C_DEFAULT);
-#define info_printf(...)    do_printf(ANSI_C_GREEN); do_printf(__VA_ARGS__); do_printf(ANSI_C_DEFAULT);
-#define warn_printf(...)    do_printf(ANSI_C_YELLOW); do_printf(__VA_ARGS__); do_printf(ANSI_C_DEFAULT);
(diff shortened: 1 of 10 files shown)


ports/renesas-ra/mpbthciport.c
```diff
@@ -37,9 +39,6 @@
 
 #if MICROPY_PY_BLUETOOTH
 
-#define debug_printf(...) // mp_printf(&mp_plat_print, "mpbthciport.c: " __VA_ARGS__)
-#define error_printf(...) mp_printf(&mp_plat_print, "mpbthciport.c: " __VA_ARGS__)
```
projectgus, line 41: These debug macros have lost the context of what file the log line is coming from.

projectgus:
@IhorNehrutsa Could you please explain the reasoning behind this feature?

My understanding is that the API in `esp_hosted_hal.h` is for integrating an arbitrary program with the ESP Hosted logging functions, so it's designed around ESP Hosted (and ESP-IDF ) logging conventions. Moving that ESP Hosted feature (and design) to become a core part of MicroPython is a big change.

Suggest a good place to start is these two [questions from last time](https://github.com/micropython/micropython/pull/8050#issuecomment-1734746745):

* What problem does this solve?
* Can you please walk step by step through a hypothetical example that explains how a developer would use this to find and debug some MicroPython behaviour?

Plus two new questions:

* Why choose this API for MicroPython logging?
* What's the plan for documenting this API and ensuring it's adopted, if it's going to be a core part of MP?

drivers/esp-hosted/esp_hosted_hal.h
```diff
@@ -29,36 +29,6 @@
 #ifndef MICROPY_INCLUDED_DRIVERS_ESP_HOSTED_HAL_H
 #define MICROPY_INCLUDED_DRIVERS_ESP_HOSTED_HAL_H
 
-#ifndef ESP_HOSTED_DEBUG
-#define ESP_HOSTED_DEBUG    (0)
-#endif
-
-#if ESP_HOSTED_DEBUG
-#define PROTOBUF_C_UNPACK_ERROR(...)    error_printf(__VA_ARGS__);
-#endif
```
iabdalkader, line 38: This block shouldn't be removed.

iabdalkader:
> My understanding is that the API in `esp_hosted_hal.h` is for integrating an arbitrary program with the ESP Hosted logging functions, so it's designed around ESP Hosted (and ESP-IDF ) logging conventions

Those macros were actually just meant to be used internally by the esp-hosted driver as a debugging aid, they are also unrelated to ESP-IDF. It was just becoming hard to see useful error output from debugging, warnings etc.. and I wanted some colorful output :) It would be nice if they are standardized in MicroPython, and replace all internal debugging macros/helpers at some point, but they could probably use some work first. Re the context, it can be maintained by using `__FILE__, `__LINE__` etc..

drivers/esp-hosted/esp_hosted_hal.h
IhorNehrutsa: Reverted

ports/renesas-ra/mpbthciport.c
IhorNehrutsa: See  do_print_line()

dpgeorge:
Thanks for the contribution, but we don't need all this additional complexity in MicroPython.  I know it's not adding any code size, but it adds lines to the source files and that adds to the complexity for someone reading and trying to understand the code.

The printf-style debugging that we do have in MicroPython is already plenty, and should be done using local-to-the-file `DEBUG_printf()` macros.

PR #13409, opened 2024-01-11
tools/mpremote: Reduce dependencies.

jspricke:
No longer require importlib_metadata on new Python versions as it is
included in the standard distribution.


robert-hh:
Which Python version do you call "newer". People might use older versions, and even then mpremote should work.

jspricke:
@robert-hh it was integrated in Python 3.8 and the conditional dependency makes sure that it works in older as well ;).

jimmo:
Thanks @jspricke this looks good!

PR #13413, opened 2024-01-11
rp2: Make result of thread.get_ident() a non-zero integer.

projectgus:
- [CPython says thread identifier is a "nonzero integer"](https://docs.python.org/3.5/library/_thread.html#_thread.get_ident), so rp2 should use a 1-indexed core number rather than 0-indexed.

- Fixes the thread/thread_ident1 test failure on rp2 port. Follow-up to #13351.

Unfortunately this may be a breaking change for any existing rp2 MP code which makes a hard-coded comparison of thread identifier to 0 or 1.


dpgeorge:
Thank you!

PR #13415, opened 2024-01-12
ports/ra: Fix SysTick clock source.

iabdalkader:
The SysTick_Config function is using PCLK instead of the system clock, to configure the ticks.


dpgeorge:
Thanks.

(I split it into two commits, one for removing the `min_delay` decl.)

PR #13425, opened 2024-01-14
py/py.mk: Remove extra build dir created for frozen_content.

iabdalkader:
The current rule creates a extra dir nested in the build dir, and frozen_content ends up in `build-BOARD/build-BOARD/frozen_content.o`. I don't think this is intentional, is there a reason for this ?

### DIFF
--- a/py/py.mk
+++ b/py/py.mk
@@ -203,7 +203,7 @@ PY_O += $(PY_CORE_O)
 
 # object file for frozen code specified via a manifest
 ifneq ($(FROZEN_MANIFEST),)
-PY_O += $(BUILD)/$(BUILD)/frozen_content.o
+PY_O += $(BUILD)/frozen_content.o
 endif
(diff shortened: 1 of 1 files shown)


jimmo:
@iabdalkader This stumped me for a while, but I think the reason this must have been done initially is that our `.c --> .o` rule is

```
$(BUILD)/%.o: %.c
```

and because the generated `frozen_content.c` is inside `build-FOO`, it must therefore generate `build-FOO/build-FOO/frozen_content.o`.

But why does your change work? It's because we inadvertently fixed this during the make-pins cleanup, because the generated `pins.c` is in the same situation. See 2eda5138701d6a7d36f8d8e3700d136b3c1161b7

So yes, this looks good, thanks for the PR!

iabdalkader:
Thanks @jimmo for the context.

PR #11516, opened 2023-05-16
ports: mimxrt: Switch to official GitHub SDK (WIP)

imi415:
This (WIP) PR will replace nxp_sdk submodule with official mcux_sdk released [here](https://github.com/nxp-mcuxpresso/mcux-sdk)

The current nxp_sdk was used for both mpy and TinyUSB project, given the fact that TinyUSB project has switched to mcux_sdk already (At least for Kinetis, LPC55, and i.MXRT devices), and nxp_sdk drivers are kept on an fairly old version with limited devices supported, the mcux_sdk might be a better choice for existing and future devices.

The PR also updated the (Config Tool Generated) clock_config presets and pinmux table for RT1052 series, which was updated due to device macro names has changed in recent revisions.

The CMSIS core headers is nolonger shipped in official mcux_sdk, so mpy's CMSIS lib is used instead.

Currently tested platform:
* MIMXRT1020EVK

Other EVKs only performed compilation tests, will be updated here after testing.

PR #13429, opened 2024-01-15
ports: mcx: Initial port for MCXN947 series MCU.

nxp-yilin:
This PR adds initial support for NXP MCXN947 MCU series, with the following boards:
FRDM_MCXN947

This port currently using nxp_driver submodule with a on-going [PR](https://github.com/hathach/nxp_driver/pull/8), since there could be potential testability and repository size issues if we just introduce the whole mcux-sdk repository (https://github.com/micropython/micropython/pull/11516), although it could be the preferable way of keeping the SDK updated.


nxp-yilin:
It seems [updating the submodule itself](https://github.com/micropython/micropython/actions/runs/7528719375/job/20491503138#step:4:620) is causing problems on mimxrt ports. I guess either we need to fix the port using new SDK or find another way of adding the new drivers.

robert-hh:
There is an open PR #11516 to add the mcux library as submodule instead of the library of Hatach. I tried this PR and the mimxrt port builds well with it. Only that the mcux lib is huge.

robert-hh:
When I include your submodule I get the error message at the build as well.  So there is a missing conditional compile in machine_rtc.c, which is required with the new lib. That came with lib v2.11, but MicroPython uses 2.10 atm.

jimmo:
Thanks @nxp-yilin for the contribution!

For my own education, is there any possibility that this could be combined with the `mimxrt` port? (Given that both ports use the `nxp_driver` submodule).

Taking on a new port is a big undertaking for the MicroPython project. The PR specifically mentions the `MCXN947`, but will this port work (without too much additional effort) on other chips in the MCX series (e.g. N94x, N54x)? What's the availability of these chips, are there dev boards and other products available? (We will need to obtain some for testing)

That said, this does look like a very well done port, thank you for making it so consistent with the other ports.

robert-hh:
>  is there any possibility that this could be combined with the mimxrt port

At first glance the answer is no. The source structure is completely different. For instance it adds a drv_xxx.c layer between the machine_xxx.c files and the mcux library. Some modules like PWM, I2S, networking are still t.b.d.

Besides that, adding that as a separate port will require updating lib/nxp-driver to v2.11 at least. I prepared it, but it not done yet since no one missed it. And there is PR #11516 to used the official NXP library instead of the clone hosted by Hatach in the tinyusb environment.

nxp-yilin:
> > is there any possibility that this could be combined with the mimxrt port
> 
> At first glance the answer is no. The source structure is completely different. For instance it adds a drv_xxx.c layer between the machine_xxx.c files and the mcux library. Some modules like PWM, I2S, networking are still t.b.d.

The MCX series uses LPC-style pin naming as well as different PORT/Pin structures than i.MXRT's IOMUXC, this requires different make-pins scripts and object structs, so the thought was ditched some while ago.
Some peripherals are not fully compatible between these series, such as CTimer and ADC etc., and there would be too much macros controlling which peripherals are get used. That drv_ layer was initially added to move these HW related part out of machine class code, however since there is only one MCX part for now so these files are mostly wrappers for MCXN94x/N54x. The missing drivers could be added once the port layout and submodule selection are confirmed.
 
> Besides that, adding that as a separate port will require updating lib/nxp-driver to v2.11 at least. I prepared it, but it not done yet since no one missed it. And there is PR #11516 to used the official NXP library instead of the clone hosted by Hatach in the tinyusb environment.

That repository contains way more parts than MicroPython project currently supports, however it seems there's approach of selectively initializing certain submodules being used by adding a `submodule` target, e.g. picosdk only for rp2 port. If this approach is preferred then we can switch to it, and the current mimxrt port can also following the latest mcux-sdk by taking the changes from #11516 or continuing using the current nxp_driver as transitional stage.

ports/mcx/pendsv.c
```diff
@@ -0,0 +1,154 @@
...
+        // Check if there is an active object to throw via nlr_jump
+        "ldr r1, pendsv_object_ptr\n"
+        "ldr r0, [r1]\n"
+        "cmp r0, #0\n"
+        "beq .no_obj\n"
+        #if defined(PENDSV_DEBUG)
+        "str r0, [sp, #8]\n"            // store to r0 on stack
+        #else
+        "str r0, [sp, #0]\n"            // store to r0 on stack
+        #endif
```
iabdalkader, line 134: The stacked registers layout doesn't change if debugging is enabled. I've been trying for a while to draw attention to the same issue in the stm32 port with no luck, but it shouldn't propagate to new ports.

nxp-yilin:
Thanks @jimmo , here's some answers regarding the previous questions:

> For my own education, is there any possibility that this could be combined with the `mimxrt` port? (Given that both ports use the `nxp_driver` submodule).

This is explained in [this comment](https://github.com/micropython/micropython/pull/13429#issuecomment-1893480467). It's possible but will add a lot of confusing macro to control which peripheral we are using.

> Taking on a new port is a big undertaking for the MicroPython project. The PR specifically mentions the `MCXN947`, but will this port work (without too much additional effort) on other chips in the MCX series (e.g. N94x, N54x)? What's the availability of these chips, are there dev boards and other products available? (We will need to obtain some for testing)

Generally, this port is for the 'MCXNx4x' series with 4 sub variants:
* MCXN947
* MCXN946
* MCXN547
* MCXN546

However the MCXN947 will be the MCU for official boards (FRDM_MCXN947, MCXN947EVK), and there's currently no official boards for the N5 series. So technically this port also works for other variants mentioned above but for now only MCXN947 boards are available to use with this port.

ports/mcx/pendsv.c
nxp-yilin: Thanks, as this macro is not controlled by DEBUG flags anyway, so I just removed this.

projectgus:

nxp-yilin:
As this PR has been hanging around for a some time, I'd like to know whether there's anything we can do to proceed with the next steps, or whether there's any concern which needs to be solved first.

For previous discussions on the SDK submodule dependency, we have the following options available:

* Continue using Hatach's cut-down version
* Bring the [mcux-sdk](https://github.com/nxp-mcuxpresso/mcux-sdk) submodule in for mcx port, while eventually migrate mimxrt port to this.

Please kindly advise the best approach so the code can be updated accordingly.

robert-hh:
About the SDK: Relying on Hatach's version may not be the best option. There is a PR #11516 dangling since a while for switching to the MCUX library as submodule. I tested that with the mimxrt port and it looked doable. Only the size of the SDK is an obstacle. And naturally the MicroPython project would NOT follow each and every update of the SDK. That's too much ongoing test effort.

nxp-yilin:
> About the SDK: Relying on Hatach's version may not be the best option. There is a PR #11516 dangling since a while for switching to the MCUX library as submodule. I tested that with the mimxrt port and it looked doable. Only the size of the SDK is an obstacle. And naturally the MicroPython project would NOT follow each and every update of the SDK. That's too much ongoing test effort.

Thanks Robert, I personally also prefer using the mcux-sdk repo. I'll try bumping the version of that submodule to somewhere contains the latest MCX parts on that PR and see whether we can make this work.

iabdalkader:
I tested this and it seems to work fine, good job! However, I wasn't sure how to deploy the firmware, I eventually just used jlink, so you may want to add deploy instructions to `boards/<board>/deploy.md` (and possibly `board.json` files too).

raveslave:
team nxp, any plans to complete this?

PR #13446, opened 2024-01-17
docs: Add note about position-only arguments in CPython vs MicroPython.

projectgus:
Required modifying the `gen-cpydiff.py` code to allow a "common" section to be inserted at the top of any of the generated files.

Alternative to #13399, as originally reported in #8039.


docs/differences/modules_common.txt
```diff
@@ -0,0 +1,33 @@
...
+
+For example, in CPython 3.4 this is the signature of the constructor ``socket.socket``::
+
+    socket.socket(family=AF_INET, type=SOCK_STREAM, proto=0, fileno=None)
+
+However, the signature documented in :func:`MicroPython<socket.socket>` is::
+
+    socket(af=AF_INET, type=SOCK_STREAM, proto=IPPROTO_TCP, /)
+
+The ``/`` at the end of the parameters indicates that they are all positional-only in MicroPython. The following code works in CPython but not in most MicroPython ports::
```
projectgus, line 21: Note: I said "most" here because actually this works in the unix port, different modsocket!

dpgeorge:
Thanks!

PR #13253, opened 2023-12-21
 ports/esp32/boards: Add support for popular ESP32_CAM boards.

gitcnd:
Tested working boards definition for ESP32 CAM boards.
Includes build instructions, image, and sample working camera-capture micropython demo script.

---

*@projectgus added the following information:*

Has matching micropython-media PR: https://github.com/micropython/micropython-media/pull/58

PR #11076, opened 2023-03-19
pyboard: Fix espxx boards hanging in bootloader after reset button.

Josverl:
## Updated Logic
The updated logic will only set the dtr/rts signals for boards that do not use standard Microsoft drivers( based on the manufacturer)
It would also be possible to check against a list of known driver manufactures (like 'wch.cn' or 'Silicon Labs' ) but this would require a list of known drivers for all ports. (which I do not have)

fixes: https://github.com/micropython/micropython/issues/11075

PR #10347, opened 2022-12-27
tools/pyboard.py : Set DTR on Windows to avoid ESPxx hard reset.

Josverl:
Fixes https://github.com/micropython/micropython/issues/9659

Based on @mgsb's proposal in https://github.com/micropython/micropython/issues/9659#issuecomment-1312791767  with added logic to _only apply the fix to Windows hosts_ 

## Tests

Tested on : 
| Host | MCU  |  result  |
|---|---|---|
| Windows  | ESP32  | ✅ |
| Windows  | ESP8266 | ✅ |
| Windows  | PYBD v1 | ✅ |
| Windows  | rp2 | ✅ |
| Windows  | samd | ✅ |
| Ubuntu 22.04 | ESP32  | ✅ |
| Ubuntu 22.04 | ESP8266 | ✅ |
| Ubuntu 22.04  | PYBD v1 | ✅ |
| Ubuntu 22.04  | rp2 | ✅ |
| Ubuntu 22.04 | samd | ✅ |

As there are currently no tests for mpremote in the repo I have created a stand alone test for this using pytest :
https://gist.github.com/Josverl/57159ee096dc941f294c745056fe341c

Happy to add test, but currently the build / test process for mpremote is not clear to me

PR #3083, opened 2017-05-12
esp8266: fix bugs in esp8266/modnetwork.c when connecting to ap

puuu:
This patches should fix the bugs discussed in #3061 .

PR #13455, opened 2024-01-17
tools/pyboard.py: Fix RTS esp32 lockups.

NoosaHydro:
this workaround prevents python .Serial(...) from unwantedly pulsing the RTS line and/or holding it high (which resets or locks up many ESP32/ESP8266 boards)

### DIFF
--- a/tools/mpremote/mpremote/transport_serial.py
+++ b/tools/mpremote/mpremote/transport_serial.py
@@ -59,7 +59,14 @@ def reraise_filesystem_error(e, info):
 
 
 class SerialTransport(Transport):
-    def __init__(self, device, baudrate=115200, wait=0, exclusive=True):
+    def __init__(
+        self,
+        device,
+        baudrate=115200,
+        wait=0,
+        exclusive=True,
+        hard_reset=False,
+    ):
         self.in_raw_repl = False
         self.use_raw_paste = True
         self.device_name = device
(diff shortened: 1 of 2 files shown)


projectgus:
@NoosaHydro Which ESP boards are you finding that lock up with the current mpremote/pyboard behaviour, and with which host OS?

Most development boards have some version of the "[automatic bootloader](https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/boot-mode-selection.html#automatic-bootloader)" circuit, meaning that default serial port behaviour (assert RTS and DTR together) doesn't trigger a reset. In recent years I don't recall too many where this doesn't work as expected, except for when wiring RTS and DTR pins directly to the chip.

andrewleech:
@projectgus There's more context to this change in the original PR it was included in: https://github.com/micropython/micropython/pull/13253#issuecomment-1893326692

projectgus:
@andrewleech Thanks Andrew, that's useful context.

I'm still curious how widespread this problem is. I have an embarrassing number of ESP32 boards and some ESP8266 ones, and I debugged these kind of issues a *lot* when I was the esptool maintainer at Espressif (and before). My impression is that it was more common back around 2016/2017, when more of the cheaper development boards weren't implementing the automatic reset circuit properly. These days I had the impression it was quite rare.

I'm not denying it happens, it's quite likely there are some families of boards that have this problem and maybe it's more widespread than I know. However, I'm curious exactly which boards do exhibit it, on which host OSes. That helps everyone understand how important it is to work around.

Flipje1955:
wtr this discussion and the inquiry as to how impactful the matter is: a couple of days ago I wanted to play with a ESP32_CAM board with an ambition to get to object detection & recognition and had a difficult time getting it to work icw Thonny. I managed  - in the end - following up on a suggestion somewhere on the net to add RTS / DTR settings in Thonny's configuration file. 

Then I had some ESP8266 Wemos D1's lying (ESP8266EX chips), once bought to supplement my collection of RPI Pico's with WIFI connectivity (now obsolete given the Pico_W availability) and ran into the same issue as the ESP32_CAM: flashing works but the Micropython prompt never shows up. Played with RTS etc here too but have not gotten the ESP8266 to run Micropython (tried different images, Thonny settings, several boards). Logged it as issue at Thonny (#3083). 
Both the ESP32 and ESP8266 end the flash process with: Hard resetting via RTS pin... Done!  Which appears to connect to the topic discussed here.
I realize these two examples do not yet represent a widespread issue (but still). .

Thx for the efforts all you guys continue to make to improve Micropython !

projectgus:
For additional reference, here is some previous work on this as done by @Josverl and discussed on Discord recently.

Note that according to Jos' research there are some (non-ESP) boards where de-asserting DTR may break functionality, and we'd need to check the same is not true for RTS.

https://github.com/micropython/micropython/pull/11076

https://github.com/micropython/micropython/pull/10347

Josverl:
@gitcnd 
I run ESP32/S3/ESP8266 with mpremote without any problems ( Flashing , connecting , resuming, rebooting) 
Can you tell exactly what serial interface your boards are using ?

My fixes determine based on OS=="win" and portinfo[0].manufacturer != "Microsoft"  as that was the best rule i could make based on my hardware. 
The part of fixes that you propose appear to be a copy of the exact code I put in earlier, . 

I ask as I spend quite a lot of time in determining, then fixing and and validating across ports and OSSes.
Also a change to one problem then can create another ( that's why #11076 was needed) 
As there are (unfortunately) no integration of HIL tests for this in the MicroPython repo, and there was some refactoring since by @jimmo, possibly errors could have crept back in.

So before merging any changes I would propose to: 

1. make sure to do such tests (I can dig up my test suite if needed)
2. make sure that we do not have two different sections of code applying a similar fix under different conditions
3. it _does_ make sense to me to add some sort of `port-reset` option, across all host platforms to be consistent.
(FWIW "reset = hard reset" is already used in mpremote, so perhaps best avoid confusion)

dpgeorge:
In light of 1aaf6eddd82170209ca90586160ac265b9412229 and 3dd8073c290c077f17ffdee17a019763ad82604d this PR is outdated and probably not needed anymore.  So I'll close it.

Issue #13380, opened 2024-01-08
rp2: 1.22.0, 1.22.1: Lockup on machine.I2S initialization


roman-yepishev:
STR:

1. Flash Raspberry Pi Pico W with `RPI_PICO_W-20231227-v1.22.0.uf2` or `RPI_PICO_W-20240105-v1.22.1.uf2`:
    ```
    MicroPython v1.22.1 on 2024-01-05; Raspberry Pi Pico W with RP2040
    ```

2. Connect over USB serial console (e.g. with `mpremote`).
3. Execute the following (no need to actually connect any I2S device):

    ```
    import machine
    i2s = machine.I2S(0, sck=machine.Pin(13), ws=machine.Pin(14),
                      sd=machine.Pin(15), mode=machine.I2S.TX,
                      bits=16, format=machine.I2S.MONO, rate=8000, ibuf=2000)
    ```

Expected results:

* I2S is initialized and the object is returned.

Actual results:

* The board stops transmitting data over USB serial console and does not return from that call (i.e. if I attempt to play audio afterwards as part of `main.py`, there is no output).

Re-flashing `RPI_PICO_W-20231005-v1.21.0.uf2` allows the use of I2S on the same pins without lockup:

```
MicroPython v1.21.0 on 2023-10-06; Raspberry Pi Pico W with RP2040
Type "help()" for more information.
>>> import machine
>>> i2s = machine.I2S(0, sck=machine.Pin(13), ws=machine.Pin(14),
...                   sd=machine.Pin(15), mode=machine.I2S.TX,
...                               bits=16, format=machine.I2S.MONO, rate=8000, ibuf=2000)
>>> help(i2s)
object I2S(id=0,
...
```

dpgeorge:
Thanks for the report.  I can confirm the issue.  It was introduced by commit cfc212b10810863b7dee4f9228ce824ceca837c6, this line:
```c
irq_set_exclusive_handler(DMA_IRQ_0, rp2_dma_irq_handler);
```

The I2S driver uses DMA IRQs, but the new `rp2.DMA` class claims the IRQ handlers for itself.

The fix is probably to change `irq_set_exclusive_handler()` to `irq_add_shared_handler()`.

@nickovs what do you think?

nickovs:
I think that should be just fine. The IRQ handlers are only used for Python callbacks, not for any driver processing, so the added latency of the shared handler is unlikely to be noticeable.

dpgeorge:
[...]

PR #13468, opened 2024-01-18
tests: rename port-specific test dirs to start with `port_`, and add rp2 specific tests

dpgeorge:
There are three commits here:
- rename port-specific tests dirs to begin with `port_`, to keep the tests dir more organised and prevent possible name clashes with the directories
- add rp2 specific tests for `rp2.DMA`
- extend the existing `machine.I2S` test to test multiple I2S instances

The aim here is to test more of the rp2 ports peripherals.  The new set of tests test currently fail on rp2 due to #13380.


jimmo:
This looks good!

Just wondering though... is there any reason why we can't do `ports/*` rather than `port_*`? (i.e. mirror the way the top-level ports dir is structured).

dpgeorge:
> is there any reason why we can't do `ports/*` rather than `port_*`?

Because there were no nested directories in `tests/` and I didn't want to start doing that (nested dirs).  But that's a pretty weak reason, and it would be neater for tests to match the top-level dir structure.

I will change it to nested dirs.

PR #13470, opened 2024-01-18
ports/rp2: Add a means to set mass-storage filesystem label

Gadgetoid:
This addresses discussion over at https://github.com/orgs/micropython/discussions/9590

For other boards which more conventionally use MSC, they call `f_setlabel` from the factory reset code paths in C, eg:

https://github.com/micropython/micropython/blob/16c6bc47cf2b9247418bca8c0d1361167d08594e/ports/renesas-ra/factoryreset.c#L88

The C method uses `MICROPY_HW_FLASH_FS_LABEL`,  but I wasn't sure by what means we could incorporate that into RP2's Python filesystem bootstrapping. Perhaps calling `vfs.label()` without an argument could use this as the default value?

### DIFF
--- a/extmod/vfs_fat.c
+++ b/extmod/vfs_fat.c
@@ -39,6 +39,7 @@
 #include <string.h>
 #include "py/runtime.h"
 #include "py/mperrno.h"
+#include "py/objstr.h"
 #include "lib/oofatfs/ff.h"
 #include "extmod/vfs_fat.h"
 #include "shared/timeutils/timeutils.h"
@@ -378,6 +379,22 @@ static mp_obj_t fat_vfs_statvfs(mp_obj_t vfs_in, mp_obj_t path_in) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_2(fat_vfs_statvfs_obj, fat_vfs_statvfs);
 
+static mp_obj_t fat_vfs_label(mp_obj_t vfs_in, mp_obj_t label_in) {
+    #ifdef MICROPY_FATFS_USE_LABEL
+    mp_obj_fat_vfs_t *self = MP_OBJ_TO_PTR(vfs_in);
+    const char *label = mp_obj_str_get_str(label_in);
+
+    FRESULT res = f_setlabel(&self->fatfs, label);
+
+    if (FR_OK != res) {
+        mp_raise_OSError(fresult_to_errno_table[res]);
+    }
+
+    #endif
+    return mp_const_none;
+}
+static MP_DEFINE_CONST_FUN_OBJ_2(fat_vfs_label_obj, fat_vfs_label);
+
 static mp_obj_t vfs_fat_mount(mp_obj_t self_in, mp_obj_t readonly, mp_obj_t mkfs) {
     fs_user_mount_t *self = MP_OBJ_TO_PTR(self_in);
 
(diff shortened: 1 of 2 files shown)


dpgeorge:
I'm not sure that this is the best way to do it, because it adds a method, code size, and differs to other Vfs filesystems that don't support a label.

How about instead copy the way the stm32 port does it, have a `MICROPY_HW_FLASH_FS_LABEL` that can be configured for any port, and in `extmod/vfs_fat.c` whenever `f_mkfs()` is called it also calls `f_setlabel(..., MICROPY_HW_FLASH_FS_LABEL)`?  Then the label will always be set to the port-defined value whenever a FAT filesystem is formatted.

Gadgetoid:
As it stands, you're probably right and I'll refactor this PR to fix the omission without addressing our specific use-case- 

I've got a fork with some minimal changes that allow USB Mass Storage mode builds to expose *two* filesystems (currently hard-coded, but then the whole USB MSC thing is a bit of an exception to the do-everything-USB-in-Python goal at the moment).

In our - admittedly quite contrived - demo case, these two filesystems have distinct labels.

The conceit is that we can have one filesystem that's writable from MicroPython, and one that's writeable from the host and use the formere as a means to support logging, saves, config changes and so forth while code is written into the latter from the host. Of course this is all highly experimental but it *works* and suggests there might be scope for some more esoteric FAT/MSC builds if we find the experience... tolerable. I am keenly aware of all the problems with filesystem syncing across multiple hosts from my tinkering with PiratePython (a CPython RAMdisk for Pi Zero).

It might be that MicroPython is unconcerned with the slippery slope that is FAT/MSC and the answer is for us to build our own C module to implement this functionality (ala the Factory Reset style of some ports). I'm happy with that.

(on the off chance anyone wants to replicate this dual FS weirdness, diff here - https://github.com/micropython/micropython/compare/master...pimoroni:micropython:feature/multi-msc)

projectgus:

Gadgetoid:
I have rebased/pushed this because I'm experimenting with FAT again.

I believe the multiple-filesystems issue still presides, since as of https://github.com/micropython/micropython/commit/6aa3c946347281875165392c09753547d8c77fc3 you can select a portion of RP2s flash on which to create a filesystem, and potentially create as many filesystems as you like.

That makes use of `MICROPY_HW_FLASH_FS_LABEL` a little contentious, since all creates vfat filesystems would have the same name with no means of overriding it.

Perhaps a `label` argument could be added to `vfs.VfsFat()`, and other filesystems that support labels? Defaulting to `MICROPY_HW_FLASH_FS_LABEL`?

Edit: I should note that my current focus is on a *single* filesystem with a boot-time switch to select between mass-storage and runtime modes, avoiding any read/write conflicts between the host and device. So the `MICROPY_HW_FLASH_FS_LABEL` would work fine for that. I am - perhaps unnecessarily -considering potential future complications.

dpgeorge:
Commit 390390ec37d01c3f0ae0af7c489d36ac1e84fbe1 made it so the label is set to `MICROPY_HW_FLASH_FS_LABEL` when formatting the filesystem via `VfsFat.mkfs()`.

I can see how it would still be useful to allow Python to set the label (eg for multiple filesystems with different labels).  That could either be done as per this PR (a new method, `VfsFat.label()`) or a new kwarg to `VfsFat.mkfs(..., label)`.  With a new method, it might also be useful to be able to get the label, eg no args would get it `vfs.label()`.

Whatever is added, it's going to cost a bit.  But I guess that's just the cost of having good filesystem support.

Gadgetoid:
I discovered the hard way that changing/writing the filesystem label on boot will result in a corrupt filesystem more often than not, so making it part of `mkfs` makes a lot of sense. It's otherwise not super obvious that it changes the FAT (well in hindsight maybe it is, but it wasn't to me) and would cause issues if there were sudden power loss (like tapping reset twice by accident).

I've since switched to preparing a pre-loaded filesystem up front and setting the label "offline" as such, so it's already set when the board boots.

Anyway TLDR I think setting label during `mkfs` is safer than a `label` method on an existing filesystem object.

dpgeorge:
> Anyway TLDR I think setting label during `mkfs` is safer than a `label` method on an existing filesystem object.

OK, sounds good.

Then I'll close this PR and in the future someone can implement the `mkfs(..., label)` argument.

PR #13000, opened 2023-11-17
extmod/asyncio: Add `Task` methods from CPython

imnotjames:
This updates the way that Task is defined in the C module for asyncio which implements tasks & task queues.  The goal is to bring `asyncio` in MicroPython slightly closer to CPython.

This adds the following methods to `Task`:

* `get_coro()` - how CPython exposes the coroutine backing the task ([CPython Docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.get_coro))
* `result()` - a helper method from CPython for returning the successful response ([CPython Docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.result))
* `exception()` - a helper method from CPython for returning the raised values ([CPython Docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.exception))
* `cancelled()` - helper for true / false if the task has been cancelled ([CPython Docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancelled))
* `add_done_callback()` - adds a callback to the task for completion or if already complete executes it ([CPython docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.add_done_callback))
* `remove_done_callback()` - removes a specific "done" callback from the task ([CPython docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.remove_done_callback))

~Adds a `hash` unary operation so `Task`s can be added to sets like in CPython.~

~Also adds `set_result` and `set_exception` for consistency but neither of these do anything - they raise a `RuntimeError` because they aren't supported on Tasks. :shrug:~

PR #13474, opened 2024-01-19
extmod/asyncio: Support gather of tasks that finish early.

dpgeorge:
Adds support to asyncio.gather() for the following cases:
- one or more sub-tasks finish before the gather starts
- one or more sub-tasks raise an exception before the gather starts

### DIFF
--- a/extmod/asyncio/core.py
+++ b/extmod/asyncio/core.py
@@ -219,6 +219,11 @@ def run_until_complete(main_task=None):
             elif t.state is None:
                 # Task is already finished and nothing await'ed on the task,
                 # so call the exception handler.
+
+                # Save exception raised by the coro for later use.
+                t.data = exc
+
+                # Create exception context and call the exception handler.
                 _exc_context["exception"] = exc
                 _exc_context["future"] = t
                 Loop.call_exception_handler(_exc_context)
--- a/extmod/asyncio/funcs.py
+++ b/extmod/asyncio/funcs.py
@@ -63,9 +63,6 @@ class _Remove:
 
 # async
 def gather(*aws, return_exceptions=False):
-    if not aws:
-        return []
-
     def done(t, er):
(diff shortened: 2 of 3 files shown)


dpgeorge:
@iabdalkader FYI

dpgeorge:
I found and fixed an edge case where all sub-tasks finish early.

iabdalkader:
@dpgeorge Thank you very much for this much needed fix 🙏🏼  I had a chance to test this today, and even though it fixes the main issue for me (`gather()` raising "can't gather"), I still see the following output... Is this expected, should I just ignore it ?

```
Task exception wasn't retrieved
future: <Task> coro= <generator object 'conn_task' at 240437a0>
  File "asyncio/core.py", line 1, in run_until_complete
  File "ucloud.py", line 341, in conn_task
DoneException: 
```

Also it seems the task exception is now saved, and I can retrieve it after `gather()` raises with `t.data`, but could we also retrieve it with `Task.exception()` for more compatibility with CPython's `Task` ? https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.exception
```Python
# Save exception raised by the coro for later use.
t.data = exc
```

dpgeorge:
> I still see the following output... Is this expected, should I just ignore it ?

Yes that's expected.  Because of the difference in GC vs reference counting, MicroPython and CPython necessarily differ in when exceptions like this are printed.  You can see in the test in this PR that those messages are sent to a dummy `custom_handler()` which just ignores them.  That way the test runs the same under CPy and uPy.

> Also it seems the task exception is now saved, and I can retrieve it after gather() raises with t.data

You should also be able to retrieve the exception from the return value of the `gather()` call.  Is that enough?  If not, why not?


> Also it seems the task exception is now saved, and I can retrieve it after `gather()` raises with `t.data`

If the return value of `gather()` is not enough we could consider adding `Task.exception()`.  See #13000.

PR #13475, opened 2024-01-19
lib/cmsis: Update to CMSIS 5.9.0.

iabdalkader:
This update brings support for Cortex-M55 and Cortex-M85, zero and copy tables and more.

Note CMSIS release 5.9.0 is the last one to contain all CMSIS components, CMSIS-6 is split into multiple smaller repos (Core, NN, DSP, RTOS etc..), so I think we can add the CMSIS 6 core repo as a submodule at some point since it's much smaller now. I avoided CMSIS 6 in this PR because it has some minor breaking changes, that will require more fixes to ports (notably core debug was renamed, and some registers were removed).

### DIFF
--- a/lib/cmsis/inc/cachel1_armv7.h
+++ b/lib/cmsis/inc/cachel1_armv7.h
@@ -1,0 +1,411 @@
+/******************************************************************************
+ * @file     cachel1_armv7.h
+ * @brief    CMSIS Level 1 Cache API for Armv7-M and later
+ * @version  V1.0.1
+ * @date     19. April 2021
+ ******************************************************************************/
+/*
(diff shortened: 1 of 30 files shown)


dpgeorge:
I think this is a good approach, updating to CMSIS 5.9.0, instead of moving to CMSIS 6.  We can move to CMSIS 6 later on if that becomes necessary.

dpgeorge:
It's interesting that the stm32 and samd port builds have changed.  I looked into the stm32 change on PYBV10 and there are minor differences in code generation in `pyb_i2c.c` and `lib/stm32lib/STM32F4xx_HAL_Driver/Src/stm32f4xx_hal_rcc.c`.

Testing this PR on PYBV10 the tests pass and everything looks OK.

PR #13478, opened 2024-01-19
asyncio/stream: Catch and ignore ECONNRESET.

ned-pcs:
This commit deals with ECONNRESET (and EPIPE) exceptions while reading from sockets.
For more context, see https://github.com/micropython/micropython-lib/issues/788

### DIFF
--- a/extmod/asyncio/stream.py
+++ b/extmod/asyncio/stream.py
@@ -3,6 +3,13 @@
 
 from . import core
 
+from errno import ECONNRESET
+
+try:
+    from errno import EPIPE
+except ImportError:
+    EPIPE = 32
+
 
 class Stream:
     def __init__(self, s, e={}):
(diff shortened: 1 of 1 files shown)


ned-pcs:
I see that the Unix tests that expected to see the ECONNRESET exception have failed.
Without this commit, my ESP32 client reading from a Unix micropython server would get ECONNRESET errors, though a Unix client didn't.

Carglglz:
@ned-pcs this makes the those tests pass 
```py
...
 except OSError as exc:
    if exc.errno in (ECONNRESET, EPIPE) and r:
        return r
    raise

```

Could this work in ESP32? 👀

ned-pcs:
@Carglglz I tried the logic you showed (`    if exc.errno in (ECONNRESET, EPIPE) and r:`) but that didn't eliminate the problem on the ESP32 side.
@dpgeorge My current test program uses Microdot as a server; I will write a simplified version that doesn't use Microdot when I get some time.

Carglglz:
@ned-pcs I think this can be closed, see #13533 for a possible fix.

ned-pcs:
@Carglglz I have tested your #13533 fix and it works. Closing.

PR #13487, opened 2024-01-20
ports/esp32: Use read_uv() to get battery voltage on UM_FEATHERS3 board.

Joey232:
Used read_uv() to get the battery voltage because it uses the
On-chip calibraton values.

Signed-off-by: Joey232 <Joey@jsconsulting.com>

### DIFF
--- a/ports/esp32/boards/UM_FEATHERS3/modules/feathers3.py
+++ b/ports/esp32/boards/UM_FEATHERS3/modules/feathers3.py
@@ -65,10 +65,11 @@ def get_battery_voltage():
     This is an approximation only, but useful to detect if the charge state of the battery is getting low.
     """
     adc = ADC(Pin(VBAT_SENSE))  # Assign the ADC pin to read
-    adc.atten(ADC.ATTN_2_5DB)  # Needs 2.5DB attenuation for max voltage of 1.116V w/batt of 4.2V
-    measuredvbat = adc.read()
-    measuredvbat /= 3657  # Divide by 3657 as we are using 2.5dB attenuation, which is max input of 1.25V = 4095 counts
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks, this looks OK to me.  What do you think @UnexpectedMaker ?

UnexpectedMaker:
> Thanks, this looks OK to me. What do you think @UnexpectedMaker ?

Sorry for the delay - not had a chance to test this until now. I'm not really sure of the point of this change as I ran them side by side and it yielded the same result as my original code, on 3 different batteries at different charge levels.

If this is changed here, it also needs to be changed in all of my helper libraries, so it's a bigger question than just this one change.

I guess there's less `magic numbers` involved. What do you think @dpgeorge ?

Joey232:
It depends upon the individual ESP32 chip as to how much of a difference this change makes. Of the two boards I have one was pretty much spot on and one was off by about 5%. After this change the first one didn't change much ,but the one that was off got much better. It just makes sense to use the calibration data if it's there.

UnexpectedMaker:
> It depends upon the individual ESP32 chip as to how much of a difference this change makes. Of the two boards I have one was pretty much spot on and one was off by about 5%. After this change the first one didn't change much ,but the one that was off got much better. It just makes sense to use the calibration data if it's there.

Ok, fair enough - I tried 5 boards here and didn't see such a discrepancy, but agree that using calibrated readings is better.

@dpgeorge I'm happy for this to get merged, and I'll get the other helpers updated soon as well. 

Cheers :)

dpgeorge:
OK, sounds good.

Rebased and merged in 3e48d24576f8a44a20e65a5782204944de2467b3

PR #13497, opened 2024-01-22
nrf/modules/machine/pwm.c: Tag a PWM device as used in the constructor.

robert-hh:
When PWM constructor was created without specifying a device or setting both freq and duty rate, it was not tagged as used, and further calls to get a PWM object may get the same PWM device assigned.
Fixes #13494.


dpgeorge:
Thanks, this fix looks good.

PR #13498, opened 2024-01-22
Zero intialize externed mp_state_ctx

Shootfast:
Added zero initialization of extern mp_state_ctx to satisfy clang on OSX

### DIFF
--- a/py/mpstate.c
+++ b/py/mpstate.c
@@ -30,4 +30,4 @@
 mp_dynamic_compiler_t mp_dynamic_compiler = {0};
 #endif
 
-mp_state_ctx_t mp_state_ctx;
+mp_state_ctx_t mp_state_ctx = {0};


dpgeorge:
This should already be in the BSS, it shouldn't need an explicit initializer.

What is the exact error that clang gives (and what version of clang, what platform)?

Shootfast:
When built via the embed module, I think clang's optimizer might be a touch too aggressive. 
I'm building the embed module .c files into libmicropython.a , and when linking I then get:

```
/Library/Developer/CommandLineTools/usr/bin/cc  -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk -mmacosx-version-min=12.5 -Wl,-search_paths_first -Wl,-headerpad_max_install_names CMakeFiles/test.dir/main.c.o -o main  ../libutil/libutil.a  ../../ext/libmicropython.a 
Undefined symbols for architecture arm64:
  "_mp_state_ctx", referenced from:
      _mp_init in libmicropython.a(runtime.c.o)
      _mp_locals_set in libmicropython.a(runtime.c.o)
      _mp_globals_set in libmicropython.a(runtime.c.o)
      _mp_locals_get in libmicropython.a(runtime.c.o)
      _mp_globals_get in libmicropython.a(runtime.c.o)
      _mp_iternext in libmicropython.a(runtime.c.o)
```

mpstate.o is definitely a part of libmicropython.a, and just applying this patch corrects the error.

Apple clang version 14.0.0 (clang-1400.0.29.202) on MacOSX Monteray (12.5.1)

projectgus:
This feels like a problem with the way libmicropython.a is being built, somehow. 

As Damien says, both the "no explicit initializer" and "explicit initializer to zero" versions should end up in the same .bss section, but maybe adding the initializer is moving it to .data section (unnecessarily) and working around something else. I'm familiar with the nuances of linking on macOS, so maybe there's some macOS-specific magic here.

Can you share the build method for libmicropython.a, please? It'd be interesting to look into the "good" and "bad" libmicropython.a libraries as well, if you have them handy and don't mind sharing them.

dpgeorge:
> When built via the embed module, I think clang's optimizer might be a touch too aggressive.
> I'm building the embed module .c files into libmicropython.a , and when linking I then get:

Hmm, that seems like a compiler/linker bug to me.  Or possibly some flags (eg LTO) that interfere.  Maybe you need to compile with `-fno-common`?

Shootfast:
My build steps are just via a very simple CMake script.

I am running the makefile in examples/embed and then taking all the .c/.h files generated and sticking them into a CMake list, and building like:
```
set(MICROPYTHON_SOURCES
	port/embed_util.c
	port/mphalport.c
	py/argcheck.c
	py/asmarm.c
	py/asmbase.c
	py/asmthumb.c
	py/asmx64.c
...
	py/vstr.c
	py/warning.c
	shared/runtime/gchelper_generic.c
)

add_library(micropython, ${MICROPYTHON_SOURCES})
target_include_directories(micropython PUBLIC ${MICROPYTHON_EXTRACT_DIR})
target_include_directories(micropython PUBLIC ${MICROPYTHON_EXTRACT_DIR}/port)
set_target_properties(micropython PROPERTIES LINKER_LANGUAGE C)
set_target_properties(micropython PROPERTIES C_STANDARD 99)
```
I request to link the library like usual:
```
target_link_libraries(main
    PRIVATE
        micropython
)
```

The compiler command that generates the mpstate.o file is here:
```
cd /Users/me/dev/proj/_build/ext && /Library/Developer/CommandLineTools/usr/bin/cc  -I/Users/me/dev/proj/_build/ext/dist/micropython -I/Users/me/dev/proj/_build/ext/dist/micropython/port -O3 -DNDEBUG -std=gnu99 -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX13.1.sdk -mmacosx-version-min=12.5 -MD -MT ext/CMakeFiles/micropython.dir/dist/micropython/py/mpstate.c.o -MF CMakeFiles/micropython.dir/dist/micropython/py/mpstate.c.o.d -o CMakeFiles/micropython.dir/dist/micropython/py/mpstate.c.o -c /Users/me/dev/proj/_build/ext/dist/micropython/py/mpstate.c
```
And when it is archived into micropython.a:
```
/Library/Developer/CommandLineTools/usr/bin/ar qc libmicropython.a CMakeFiles/micropython.dir/dist/micropython/extmod/vfs.c.o CMakeFiles/micropython.dir/dist/micropython/extmod/vfs_posix.c.o CMakeFiles/micropython.dir/dist/micropython/extmod/vfs_posix_file.c.o CMakeFiles/micropython.dir/dist/micropython/extmod/vfs_reader.c.o CMakeFiles/micropython.dir/dist/micropython/port/embed_util.c.o CMakeFiles/micropython.dir/dist/micropython/port/mphalport.c.o CMakeFiles/micropython.dir/dist/micropython/py/argcheck.c.o CMakeFiles/micropython.dir/dist/micropython/py/asmarm.c.o 
...
CMakeFiles/micropython.dir/dist/micropython/py/vstr.c.o CMakeFiles/micropython.dir/dist/micropython/py/warning.c.o CMakeFiles/micropython.dir/dist/micropython/shared/runtime/gchelper_generic.c.o
/Library/Developer/CommandLineTools/usr/bin/ranlib libmicropython.a
```
There aren't any exotic compiler flags being passed around, and I get the same error in both Debug and Release builds.

Shootfast:
> > When built via the embed module, I think clang's optimizer might be a touch too aggressive.
> > I'm building the embed module .c files into libmicropython.a , and when linking I then get:
> 
> Hmm, that seems like a compiler/linker bug to me. Or possibly some flags (eg LTO) that interfere. Maybe you need to compile with `-fno-common`?

Adding this flag worked! Which then sent me down this rabbit hole:
https://stackoverflow.com/questions/19398742/os-x-linker-unable-to-find-symbols-from-a-c-file-which-only-contains-variables
Apparently the mach-o binary format behaves differently to others and requires this explicit instruction vs elf.

https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/1-Articles/executing_files.html#//apple_ref/doc/uid/TP40001829-98432-TPXREF120

Thanks for the help, I guess we can abandon this merge request if explicitly initializing these structs goes against other design considerations for the library.

projectgus:
Glad you got things working, @Shootfast!

> Apparently the mach-o binary format behaves differently to others and requires this explicit instruction vs elf.

The second answer to the linked question is interesting, it looks like it's macos `ar` that's different: https://stackoverflow.com/a/26581710

So the other option would be to modify CMake to pass `-c` option to ranlib (look like maybe [STATIC_LIBRARY_OPTIONS](https://stackoverflow.com/a/52124651)). However, `-fno-common` seems like a very good solution.

@dpgeorge Does MP rely on common symbols anywhere? If not, could we put this in the embedded makefile without any adverse effect?

dpgeorge:
> Does MP rely on common symbols anywhere?

No, we don't rely on having the same symbol (variable) defined in different .o files.



> If not, could we put this in the embedded makefile without any adverse effect?

Yes I think that would be fine.

PR #13499, opened 2024-01-22
ports/mimxrt: Fix deepsleep wakeup pin ifdef.

kwagyeman:
The ifdef for the wakeup pin was not being defined. https://github.com/openmv/micropython/blob/f13c0bc8a9b922ef856a454592366d5b428f532f/ports/mimxrt/modmachine.c#L145

Note, waking up via deepsleep on the mimxrt is broken without this.


dpgeorge:
Thanks, merged in bf6214505fb477890416f4c5ec5912611fc361d6 with minor code style changes.

PR #13500, opened 2024-01-23
tests/extmod: Remove asyncio .exp files that match CPython output.

dpgeorge:
These were added back in commit c4935f30490d0446e16a51dbf7a6397b771cf804 because the tests required CPython 3.8, which was quite new at the time. But CPython 3.8 was released over 4 years ago (October 2019) and the CI test runners, and developers, have this (or a more recent) CPython version.

Removing the .exp files also helps keep MicroPython semantics the same as CPython.


dpgeorge:
Looks like there are some CI failures here on macOS and Appveyor.

dpgeorge:
The tests with timing are failing on macOS: `asyncio_basic.py`, `asyncio_wait_task.py`, the parts with "took" logic.

And `asyncio_fair.py` is failing on AppVeyor, probably the timing in the test is a little too tight.

dpgeorge:
I tweaked `asyncio_fair.py` to be more deterministic, and added back .exp files for those tests with timing.

PR #13502, opened 2024-01-23
github: Convert issue templates to forms.

projectgus:
## What problem does this PR solve?

The existing issue reports have a template, but many issues don't address the items mentioned there or don't provide enough detail. This ultimately leads to long back-and-forth discussions to tease out relevant details.

Some issues are still opened that should really be questions asked in Discussions or Discord.

## How does this PR solve the problem?

Using the GitHub [Issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) feature means we can ask structured questions with required answers, include links to other resources which are easy to click.

We can't force people to slow down and provide good answers, but we can strongly encourage it and make it harder to skim over essential details.

## How is/was this PR tested?

I have this branch set as the default on my fork, so the functionality can be tested here:
https://github.com/projectgus/micropython/issues/new/choose

(Feel free to submit dummy issues.)


Josverl:
@projectgus , 


**Documentation**: could you add the `docs` label for documentation issues as a default ?
**Security** : I think there may be a better way to set this up so that any security concerns can be reported open , and safe.
the recommended way is to publish a security policy.

Responsible security reporting in GitHub is a process of privately and collaboratively disclosing security vulnerabilities in projects hosted on GitHub. [The recommended way to set up responsible security reporting in GitHub is to enable private vulnerability reporting for your repository, and to create a security policy file (SECURITY.md) that provides clear instructions for how to report a security vulnerability in your project](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/about-coordinated-disclosure-of-security-vulnerabilities)[1](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/about-coordinated-disclosure-of-security-vulnerabilities).

stinos:
The principle is certainly ok, and I understand that everything in there is there for a reason (namely: people having reported issues lacking all the needed information in the past), but even when trying to look at this as an inexperienced first-time user overall this feels bit much, almost off-putting.

For example there are some seemingly duplicate things (e.g. the last 2 Checks essentially duplicate the numbered text 1-2), 2 COC links on consecutive lines, there's 'MicroPython version' then the next line is 'What exact version of MicroPython are you using?', or 'Please include all details' then 'Please provide as many details as possible' couple of lines below), the use of italic is not super consistent (sometimes it's for stressing things, otherwise it's more like a description, it kinda looks like the intent is to have italic before each text box but that's not the case), combined with the different font sizes etc makes everything a bit awkward to read, sending people back all the way to the beginning is a bit strange (could say that right underneath the title perhaps?) and in general there's just a lot of text and a lot of newlines.

If you think all of this is still needed then I'd suggest to at least reduce the number of newlines (what is under 'Expected behaviour' and 'Observed Behaviour' for instance could easily be one paragraph, no?), use plain text in the same font everywhere except bold for the headings as it is now. Or if you want I can have a go at it based on what you have now to explain what I mean.

projectgus:
Thanks @Josverl and @stinos for taking a look.

> Documentation: could you add the docs label for documentation issues as a default ?

I think this should be there - it's transferred from the markdown template to the yaml. It doesn't work in my fork because it seems like labels don't transfer into forked repos.

> Security : I think there may be a better way to set this up so that any security concerns can be reported open , and safe.
the recommended way is to publish a security policy.

I agree, and I thought about it as well when converting the markdown template to the YAML file (the "security" form is pretty closely based on the existing "security" template.). I'll talk this over with @dpgeorge and @jimmo. It might end up in its own PR, though.

> even when trying to look at this as an inexperienced first-time user overall this feels bit much, almost off-putting.

@stinos you're right, and I agree with most of your specific comments. The problem I had in mind while writing these was to make it harder for folks to skim past the questions and miss the various "sign posts" to other resources. This happens pretty often with the current template. However, by adding more obstacles to submit a bad issue report I've made it quite difficult to submit a good issue report! I agree it should still feel easy/welcome to submit a good issue report.

I also don't particularly like the default GitHub styling for the form, which is out of our control, but you're right that I can cut back on the italics, etc.

I'll give it another pass today, and update the PR.

projectgus:
> Documentation: could you add the docs label for documentation issues as a default ?

Oh, actually the label should be `documentation` not `docs`. I translated it incorrectly. Oops!

projectgus:
OK, have simplified it down and hopefully it's a bit less daunting. PTAL!

(I also discovered the hard way that there's no nice way to wrap Markdown lines in the YAML file. Neither the `|` nor the `>` multi-line style maps to what you'd get if the contents were in a plain markdown file. So I've stuck with long lines.)

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,59 @@
...
+labels: ['enhancement']
+body:
+  - type: markdown
+    attributes:
+      value: |
+        This form is for requesting features or improvements in MicroPython.
+
+        #### Not a MicroPython core feature?
+
+        * If you have a question \"How Do I ...?\", please post it on [GitHub Discussions](https://github.com/orgs/micropython/discussions/) or [Discord](https://discord.gg/RB8HZSAExQ) instead of here.
```
jimmo, line 12: Many feature requests would be better starting off life as a discussion. So maybe an extra line below this one also linking to discussions/discord to suggest "get feedback on your idea first" etc.

jimmo (commented):
Thanks @projectgus, I really like this.

.github/ISSUE_TEMPLATE/feature_request.yml
projectgus: @jimmo Done, PTAL!

projectgus:
@stinos If you have time, it'd be great to hear your thoughts about the new version of this.

stinos:
Looks pretty neat. Only the 'not a bug report?' looks somewhat off but that might just be me. Maybe combine this all into one bullet point list with no bold header, like

- If you have a question "How Do I ...?", please post it on [GitHub Discussions](https://github.com/orgs/micropython/discussions/) or [Discord](https://discord.gg/RB8HZSAExQ) instead of here.
- For missing or incorrect documentation, please submit a [documentation request](https://github.com/micropython/micropython/issues/new?assignees=&labels=documentation&projects=&template=documentation.yml&title=docs%3A+) instead of a bug report.
- Please provide as much detail as you can, it really helps us find and fix bugs faster.

? Also minor, but I'm wondering why the documentation request is listed separately here but not feature request and security issue etc? On the other hand they are already under 'Report a bug or unexpected behaviour. If this doesn't look right, choose a different type. ' but that one is easy to miss.

projectgus:
> Looks pretty neat.

Thanks! It got a lot better after your initial feedback, so thanks again for that too.

> Only the 'not a bug report?' looks somewhat off but that might just be me. Maybe combine this all into one bullet point list with no bold header,

I agree, but also I think it's potentially good for this to be slightly out of place. Having a heading that slightly stands out that says "Not a bug report?" is (I hope) a point that might catch someone who is otherwise scrolling past all of the information in order to submit their issue.

> I'm wondering why the documentation request is listed separately here but not feature request and security issue etc?

This is a good question! I copied this part from the existing templates. @jimmo added it in commit 316008046a689b59e7a521c811d6e381bd03fb5a, maybe he remembers? My guess is that a number of "not a bug report" issues were describing things that weren't in the docs...?

I notice it's actually currently mentioned in the feature request template as well and I didn't transfer this into the new form. @jimmo, do you have a feel for which form(s) - if any - should specifically prompt about missing docs?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,107 @@
+name: Bug Report
```
jimmo, line 1: Minor nitpick -- this was previously "Bug **r**eport" (matching "Enhancement request", "Documentation issue", etc)

jimmo:
> I notice it's actually currently mentioned in the feature request template as well and I didn't transfer this into the new form. @jimmo, do you have a feel for which form(s) - if any - should specifically prompt about missing docs?

@projectgus The original text was specifically in the context of a "how do I?" type question (i.e. the best outcome is the documentation is expanded to cover whatever the person is stuck on).

But I guess in the new layout the context is the "Not a bug report?" heading which is why @stinos 's question makes sense.

Maybe you could just add a third bullet point to link back to the issue choose page? e.g. "For other issue types see here"

projectgus:
Thanks @jimmo, I understand the context change better now. How about this version?

jimmo:
LGTM!

dpgeorge:
Thanks @projectgus !

Issue #10947, opened 2023-03-06
RP2040: ADC value different in ADC(channel) vs ADC(Pin(#))


scruss:
A user on the the Raspberry Pi forum noticed this issue: [ADC channel number vs. GPIO pin number - Raspberry Pi Forums](https://forums.raspberrypi.com/viewtopic.php?p=2087648#p2087330). In short, if an ADC is initialized in this manner:

    adc = ADC(2)

(that is, by ADC channel) it will return a higher reading than if it is initialized by Pin using:

    adc = ADC(Pin(28))

A user who cannot post issues here noted that initializing by channel [leaves the pull-down enabled](https://forums.raspberrypi.com/viewtopic.php?p=2087648#p2087378), while initializing by pin does not. There is example code in the link that demonstrates the issue.

The Raspberry Pi MicroPython introductory documentation uses the channel method, and thus may be producing spurious results for users.

 Version information shown in the REPL:
```
MicroPython v1.19.1-746-gf2de289ef on 2022-12-13; Raspberry Pi Pico with RP2040
```

th3w:
Out of  curiosity I have also tried to find the cause today:

In port/rp2/machine_adc.c -> machine_adc_make_new at  line 88 ,  the sdk function adc_gpio_init(channel) will only be called if value of channel is between 26-29. But the channel value  can also be a number between 0 and 4. (4 is a special case because this is the internal ADC_TEMPSENSOR)
So  a given channel value between  0-3 must first translated to gpio number 26-29 before calling adc_gpio_init().

scruss:
Thanks for the PR. I suspect we may need to do something for Channel 4, as it has always read far higher than I'd expect. I'm not sure that it's exposed as a Pin number, though

dpgeorge:
Honestly, I think the original behaviour is correct and gives the most flexibility:
- you can specify low channel numbers to get access to the raw channel without any configuration, eg `ADC(2)`
- you can specify high channel numbers to initialise the GPIO, eg `ADC(28)`
- you can pass in a Pin to use ADC on that pin, eg `ADC(Pin(28))`

scruss:
But where does it say "use the Pin or high channel number or you'll get unexpected readings"? If this is not made utterly clear in how ADC is accessed, it can't be the correct behaviour.

dpgeorge:
The problem is, initialising the ADC for GPIO reading when specifying a low channel number may break existing use of the ADC this way.

[...]

PR #12417, opened 2023-09-11
rp2/boards/ARDUINO_NANO_RP2040_CONNECT: Add support for reading the Nina ADC pins.

iabdalkader:
This PR provides access to the ADC pins controlled by the Nina module, which couldn't be used before by `machine.ADC`. With this change it's now possible to use those pins as regular ADC GPIOs:

```python
import time
from machine import ADC

adc = ADC("A4")
while True:
    print(adc.read_u16())
    time.sleep_ms(10)
```

PR #13509, opened 2024-01-24
rp2: Fix the ADC pin initialisation.

robert-hh:
- Fix the ADC pin initialization:  the change closes the gap when an integer was used as Pin reference. With that change, e.g. ADC(26), ADC(Pin(26)) and ADC("GP26") behave identical and the GPIO is initialized in high-Z mode. Only when using ADC channel numbers 0 - 4 the ADC pin is not initialized and that step is left to the user code.

### DIFF
--- a/docs/rp2/quickref.rst
+++ b/docs/rp2/quickref.rst
@@ -187,6 +187,14 @@ Use the :ref:`machine.ADC <machine.ADC>` class::
     adc = ADC(Pin(26))     # create ADC object on ADC pin
     adc.read_u16()         # read value, 0-65535 across voltage range 0.0v - 3.3v
 
+The argument of the constructor ADC specifies either a Pin by number, name of as
+Pin object, or a channel number in the range 0 - 3 or ADC.CORE_TEMP for the
+internal temperature sensor. If a pin is specified,
+the pin is initialized in high-Z mode. If a channel number is used, the pin
+is not initialized and configuring is left to the user code. After hard reset,
+RP2040 pins operate in current sink mode at about 60µA. If the pin is not
+otherwise configured, that may lead to wrong ADC readings.
+
 Software SPI bus
 ----------------
 
--- a/ports/rp2/machine_adc.c
+++ b/ports/rp2/machine_adc.c
@@ -74,14 +74,14 @@ static mp_obj_t mp_machine_adc_make_new(const mp_obj_type_t *type, size_t n_args
     const machine_pin_obj_t *pin = NULL;
 
     if (mp_obj_is_int(source)) {
-        // Get and validate channel number.
         channel = mp_obj_get_int(source);
-        if (ADC_IS_VALID_GPIO(channel)) {
-            channel = ADC_CHANNEL_FROM_GPIO(channel);
-        } else if (!(channel >= 0 && channel <= ADC_CHANNEL_TEMPSENSOR)) {
-            mp_raise_ValueError(MP_ERROR_TEXT("invalid channel"));
+        if (!(channel >= 0 && channel <= ADC_CHANNEL_TEMPSENSOR)) {
+            // Not a valid ADC channel, fallback to searching for a pin.
+            channel = -1;
         }
-    } else {
+    }
+
+    if (channel == -1) {
         // Get GPIO and check it has ADC capabilities.
         pin = machine_pin_find(source);
         bool valid_adc_pin = false;


iabdalkader:
I think this was intentional, there were long discussions about that, but see these comments in particular:

https://github.com/micropython/micropython/pull/12417#issuecomment-1770226362

And this one:

https://github.com/micropython/micropython/issues/10947#issuecomment-1457417404

TL;DR: The idea was Not to initialize any GPIOs if you pass raw channel numbers.

robert-hh:
I understand this use case. I do not know if that happens often. But the way it is now, the ADC input behaves erratic, with the returned values depending on the source impedance at the input, the temperature of the MCU, firmware versions, .... This erratic behavior seems worse to me that a potential double use of the ADC pin. At least it must be documented, that in this case the ADC return values are highly erratic.

robert-hh:
Interestingly this "undefined" state after power.on seems to apply to all GPIO pins. It is a kind of constant current mode of ~60µA to GND. This is different to other MCUs which have the GPIO pins in High Z input mode after power on, and may be important for low power applications.

robert-hh:
> L;DR: The idea was Not to initialize any GPIOs if you pass raw channel numbers.

There is a slight inconsistency then. At all other places one can either use a Pin number, a Pin name or a Pin object if a Pin is designated. Here, there is a difference between ADC(26) and ADC("GP26") resp ADC(Pin(26). The first one is a RAW ADC, the other are configured ones.

iabdalkader:
Yes, a int here should be a channel number not a GPIO number, see

https://docs.micropython.org/en/latest/library/machine.ADC.html#machine.ADC

Either way, it's not my choice, I just wanted to point out that it was done this way for the reasons mentioned by Damien in those threads.

dpgeorge:
> So I changed the PR to leave the ADC GPIO uninitialized when the channel number 0-4 is used, and initialize it when a Pin reference is used, e.g. either ADC(26) or ADC(Pin(26)) or ADC("GP26").

This does seem reasonable.  And adds a little more consistency in the way you can specify a pin as either an integer, string or `Pin` object, and get the same configuration of the ADC.

This PR as it currently stands does not contradict my earlier comment here: https://github.com/micropython/micropython/issues/10947#issuecomment-1457239642 .  Actually, this PR makes that comment correct, because the statement "you can specify high channel numbers to initialise the GPIO, eg ADC(28)" in that comment is currently not true, and this PR makes it a true statement.

And this PR does not contradict this comment either: https://github.com/micropython/micropython/issues/10947#issuecomment-1457417404

The only issue with this PR is that it is a slight breaking change.  Code that originally did `ADC(26)` will now behave like `ADC(Pin(26))`.  If we are going to make that change then we need to make it clear in the release notes.

I'm happy to make this change, ie merge this PR as it currently stands.

dpgeorge:
If we merge this PR I'd say we can then close #10947.

docs/rp2/quickref.rst
```diff
@@ -187,6 +187,13 @@ Use the :ref:`machine.ADC <machine.ADC>` class::
     adc = ADC(Pin(26))     # create ADC object on ADC pin
     adc.read_u16()         # read value, 0-65535 across voltage range 0.0v - 3.3v
 
+The argument of the constructor ADC specifies either a Pin by number, name of as
+Pin object, or a channel number in the range 0 - 3. If a pin is specified,
```
dpgeorge, line 191: I think it should also state you can pass in `ADC.CORE_TEMP` as a channel (channel 4).  That will configure the internal temperature sensor for reading on the ADC.

ports/rp2/machine_adc.c
```diff
@@ -73,12 +73,9 @@ static mp_obj_t mp_machine_adc_make_new(const mp_obj_type_t *type, size_t n_args
...
 
-    if (mp_obj_is_int(source)) {
-        // Get and validate channel number.
-        channel = mp_obj_get_int(source);
-        if (ADC_IS_VALID_GPIO(channel)) {
-            channel = ADC_CHANNEL_FROM_GPIO(channel);
-        } else if (!(channel >= 0 && channel <= ADC_CHANNEL_TEMPSENSOR)) {
+    if (mp_obj_is_int(source) && (channel = mp_obj_get_int(source)) <= ADC_CHANNEL_TEMPSENSOR) {
+        // Validate channel number.
+        if (channel < 0) {
```
dpgeorge, line 78: You could probably eliminate the below "invalid channel" error entirely by doing something like this:
```c
if (mp_obj_is_int(source)) {
    channel = mp_obj_get_int(source);
    if (!(channel >= 0 && channel <= ADC_CHANNEL_TEMPSENSOR)) {
        // Not a valid ADC channel, fallback to searching for a pin.
        channel = -1;
    }
}

if (channel == -1) {
    // Get GPIO and check it has ADC capabilities.
    pin = machine_pin_find(source);
    ...
}
```

docs/rp2/quickref.rst
robert-hh: Done.

ports/rp2/machine_adc.c
robert-hh: Looks good, saving some bytes. Changed by updating the respective commit(s).

dpgeorge:
Thanks for updating.

docs/rp2/quickref.rst
```diff
@@ -187,6 +187,14 @@ Use the :ref:`machine.ADC <machine.ADC>` class::
     adc = ADC(Pin(26))     # create ADC object on ADC pin
     adc.read_u16()         # read value, 0-65535 across voltage range 0.0v - 3.3v
 
+The argument of the constructor ADC specifies either a Pin by number, name of as
+Pin object, or a channel number in the range 0 - 3 or ADC.CORE_TEMP for the
```
lurch, line 191: "name of as Pin object" looks like a typo?

robert-hh: It is.

PR #13515, opened 2024-01-25
Ports: Add LED pin aliases for all Arduino boards.

iabdalkader:
The standard Arduino pinout uses LED_R/G/B and LED_BUILTIN (if available). This patch adds aliases to match the standard pinout, while retaining LED_RED/GREEN/BLUE for compatibility with existing scripts and examples.

CC @sebromero


sebromero:
@iabdalkader Thanks for the PR! In the Arduino core the RGB LEDs don't use an underscore though. I suggest we keep it consistent: `LEDR`, `LEDG`, `LEDB`.

See:

https://github.com/arduino/ArduinoCore-renesas/blob/main/variants/PORTENTA_C33/pins_arduino.h#L74-L76
https://github.com/arduino/ArduinoCore-mbed/blob/main/variants/GIGA/pins_arduino.h#L30-L32
https://github.com/arduino/ArduinoCore-mbed/blob/main/variants/PORTENTA_H7_M7/pins_arduino.h#L30-L32
https://github.com/arduino/ArduinoCore-mbed/blob/main/variants/NICLA_VISION/pins_arduino.h#L30-L32
https://github.com/espressif/arduino-esp32/blob/master/variants/arduino_nano_nora/pins_arduino.h#L81-L83

iabdalkader:
Updated.

sebromero (approved):
LGTM

PR #13516, opened 2024-01-25
core: Add support for the :_b format specifier.

jepler:
This groups non-decimal values by fours, bbb_bbbb_bbbb.

The SHOW_COMMA (now SHOW_SEP) flag is overloaded for both decimal and binary contexts; use of incorrect `:,b` is not diagnosed, and for `:_d` the separator is still printed as `,` not `_`.

Over in CircuitPython this was recently raised as an issue: https://github.com/adafruit/circuitpython/issues/8837

My gut says that _I_ would find the 4-character grouping useful so I went ahead and implemented it. Let's see what the code size cost is.


jepler:
I rebased this. If it's not worth it given the code size growth or differences from cpython, I'm fine with closing this PR. Thanks for taking another look.

dpgeorge:
I had a look at this to see if I could reduce code size by:
- changing `flags` from `int` to `unsigned int`
- storing the separation character in `flags` as the upper bits (shifted left by 10)

That makes the code simpler, eg in `mp_print_mp_int`: `char comma = flags >> PF_FLAG_SEP_POS`

And in `mp_obj_str_format_helper`:
```
           if (*s == ',' || *s == '_') {
               flags |= *s << PF_FLAG_SEP_POS;
               s++;
           }
```
With that change, code size is down by about 30 bytes.  And it now supports `{:_d}`.

jepler:
Aha! I had considered putting the separator character in the flags but simply couldn't get it to save code size.

The calling convention of mp_print_strn and mp_print_float and the PF_FLAG values don't seem to be part of the dynruntime API but I refrained from changing them. Should I remove PF_FLAG_SHOW_SEP and renumber the rest?

dpgeorge:
I think the signature of `mp_print_mp_int()` also needs changing for the `flags` argument.

We also need to guarantee that there are enough bits in `unsigned int` to hold the separator character.  On 16-bit systems we might need to explicitly widen this type?



> The calling convention of mp_print_strn and mp_print_float and the PF_FLAG values don't seem to be part of the dynruntime API but I refrained from changing them

Correct!

> Should I remove PF_FLAG_SHOW_SEP and renumber the rest?

Yes, I think it's worth doing that.

jepler:
OK, so a shift of 9 is possible after renumbering flags, and `'_' << 9` still fits in unsigned 16 bits.

I also added a cpydiff entry for this, since non-standard `{:,b}` is now accepted.

dpgeorge (approved):
Looks really good now, thank you!

It's now a minor increase in code size, for a little bit more CPython compatibility.

dpgeorge:
Squashed, rebased and merged in 9032491efd8a63e0b13dbcfb8579cde4791c03af and f77fd6257cc14ff9a05d702406501a49b7bacbd0

jepler:
Thanks!

PR #13519, opened 2024-01-25
py/builtinimport: Simplify calls to stat_path().

smurfix:
stat_path is only called with stringified vstr_t objects.

Thus, pulling the stringification into the function replaces three function calls with one, saving a few bytes.

### DIFF
--- a/py/builtinimport.c
+++ b/py/builtinimport.c
@@ -57,16 +57,17 @@
 // uses mp_vfs_import_stat) to also search frozen modules. Given an exact
 // path to a file or directory (e.g. "foo/bar", foo/bar.py" or "foo/bar.mpy"),
 // will return whether the path is a file, directory, or doesn't exist.
-STATIC mp_import_stat_t stat_path(const char *path) {
+STATIC mp_import_stat_t stat_path(vstr_t *path) {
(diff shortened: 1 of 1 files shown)


robert-hh:
The unix coverage check fails always. So do not care. The code size savings are marginal. The mimxrt port even grows.

jimmo:
> replaces three function calls with one

gcc's optimiser works in mysterious ways :)

I'm pretty neutral on this... the original context is that `stat_path` is a wrapper for `mp_import_stat` so that's why the function signatures match, but the code has evolved a bit since then so I don't think this argument is very strong.

If every callsite is having to use `stat_path(vstr_null_terminated_str(x))` then I agree that's a good argument for changing the signature of `stat_path`. Ideally we could make it not require null-terminated strings at all, but that would mean changing the signature of `mp_import_stat` and `mp_find_frozen_module`.

dpgeorge:
> If every callsite is having to use `stat_path(vstr_null_terminated_str(x))` then I agree that's a good argument for changing the signature of `stat_path`

I also agree with this point.  It reduces code duplication by a little, and puts the null-termination logic/requirement in one location.

PR #13525, opened 2024-01-26
esp32: Allow to use the DEBUG_printf.

IhorNehrutsa:


IhorNehrutsa:
py/mpprint.h: Add do_printf() macro. #13400

ports/esp32/mpconfigport.h
```diff
@@ -69,6 +69,7 @@
 #define MICROPY_STREAMS_POSIX_API           (1)
 #define MICROPY_USE_INTERNAL_ERRNO          (0) // errno.h from xtensa-esp32-elf/sys-include/sys
 #define MICROPY_USE_INTERNAL_PRINTF         (0) // ESP32 SDK requires its own printf
+#define MICROPY_DEBUG_PRINTERS              (1)
```
dpgeorge, line 72: I don't think it makes sense to enable this by default.  When you build your own board you can easily enable it in the boards config file.

PR #13533, opened 2024-01-27
extmod/asyncio: Fix `Stream.wait_closed`.

Carglglz:
Leaving pending bytes at server side will cause the socket to be closed promptly when `Stream.wait_closed` is called which raises `ECONNRESET` at client side.

The fix is to make sure there is no pending data and then close the socket.

### DIFF
--- a/extmod/asyncio/stream.py
+++ b/extmod/asyncio/stream.py
@@ -2,6 +2,7 @@
 # MIT license; Copyright (c) 2019-2020 Damien P. George
 
 from . import core
+from .funcs import wait_for_ms
 
 
 class Stream:
@@ -9,6 +10,7 @@ class Stream:
         self.s = s
         self.e = e
         self.out_buf = b""
+        self._server = False
 
     def get_extra_info(self, v):
         return self.e[v]
@@ -17,7 +19,15 @@ class Stream:
         pass
 
     async def wait_closed(self):
-        # TODO yield?
+        while True and self._server:
+            try:
+                # this makes sure there is no pending data in the socket
+                # to avoid ECONNRESET at client side.
+                if not await wait_for_ms(self.read(), 10):
+                    break
+            except Exception:  # ECONNRESET, TimeoutError
+                break
+
         self.s.close()
 
     # async
@@ -171,6 +181,7 @@ class Server:
                     continue
             s2.setblocking(False)
             s2s = Stream(s2, {"peername": addr})
+            s2s._server = True
             core.create_task(cb(s2s, s2s))
 
 
--- a/tests/multi_net/asyncio_tcp_server_client_drain.py
+++ b/tests/multi_net/asyncio_tcp_server_client_drain.py
@@ -1,0 +1,59 @@
+# Test asyncio TCP server and client using start_server() and open_connection()
+
+try:
+    import asyncio
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+PORT = 8000
+
+
+async def handle_connection(reader, writer):
+    # Test that peername exists (but don't check its value, it changes)
+    writer.get_extra_info("peername")
+
+    data = await reader.read(11)  # leaving bytes unread triggers the error
+    # data = await reader.read(13)  # reading everything
+    print("echo:", data.replace(b"\r\n", b""))
+    writer.write(data)
+    await writer.drain()
+
+    print("close")
+    writer.close()
+    await writer.wait_closed()
+
+    print("done")
+    ev.set()
+
+
+async def tcp_server():
+    global ev
+    ev = asyncio.Event()
+    server = await asyncio.start_server(handle_connection, "0.0.0.0", PORT)
+    print("server running")
+    multitest.next()
+    async with server:
+        await asyncio.wait_for(ev.wait(), 10)
+
+
+async def tcp_client(message):
+    reader, writer = await asyncio.open_connection(IP, PORT)
+    print("write:", message.replace(b"\r\n", b""))
+    writer.write(message)
+    await writer.drain()
+    data = await reader.read()  # if the client doesn't read exactly or is fast
+    # enough e.g.
+    # data = await reader.read(11)  # this works on unix because is
+    # fast enough, however esp32 fails.
+    print("read:", data.replace(b"\r\n", b""))
+
+
+def instance0():
+    multitest.globals(IP=multitest.get_network_ip())
+    asyncio.run(tcp_server())
+
+
+def instance1():
+    multitest.next()
+    asyncio.run(tcp_client(b"client data\r\n"))
--- a/tests/multi_net/asyncio_tcp_server_client_drain.py.exp
+++ b/tests/multi_net/asyncio_tcp_server_client_drain.py.exp
@@ -1,0 +1,8 @@
+--- instance0 ---
+server running
+echo: b'client data'
+close
+done
+--- instance1 ---
+write: b'client data'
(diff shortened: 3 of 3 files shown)


dpgeorge:
Thanks for the patch.

As per https://github.com/micropython/micropython/pull/13478#issuecomment-1902831806, this will need a minimal example to show the problem.

Carglglz:
> As per https://github.com/micropython/micropython/pull/13478#issuecomment-1902831806, this will need a minimal example to show the problem.

Yes the example is the test (see the comments)

```py
async def handle_connection(reader, writer):
     # Test that peername exists (but don't check its value, it changes)
     writer.get_extra_info("peername")

     data = await reader.read(11)  # leaving bytes unread triggers the error
     # data = await reader.read(13)  # reading everything
     print("echo:", data.replace(b"\r\n", b""))
     writer.write(data)
     await writer.drain()

     print("close")
     writer.close()
     await writer.wait_closed()

     print("done")
     ev.set()


 async def tcp_server():
     global ev
     ev = asyncio.Event()
     server = await asyncio.start_server(handle_connection, "0.0.0.0", PORT)
     print("server running")
     multitest.next()
     async with server:
         await asyncio.wait_for(ev.wait(), 10)


 async def tcp_client(message):
     reader, writer = await asyncio.open_connection(IP, PORT)
     print("write:", message.replace(b"\r\n", b""))
     writer.write(message)
     await writer.drain()
     data = await reader.read()  # if the client doesn't read exactly or is fast
     # enough e.g.
     # data = await reader.read(11)  # this works on unix because is
     # fast enough, however esp32 fails.
     print("read:", data.replace(b"\r\n", b""))

def instance0():
     multitest.globals(IP=multitest.get_network_ip())
     asyncio.run(tcp_server())


 def instance1():
     multitest.next()
     asyncio.run(tcp_client(b"client data\r\n"))
```

Without the patch this test fails with `OSError: [Errno 104] ECONNRESET`

Also I'm not sure if this is the right fix or if we should be looking deeper? 🤔

dpgeorge:
> Yes the example is the test

Oh, sorry, right!  I will take a proper look at it.

bikeNomad:
I have tested this patch with my code (and with @Carglglz recent `aiohttp` fixes) and it appears to work.

dpgeorge:
Testing under CPython, the behaviour on master seems to be correct (ie without this fix): I can get CPython asyncio to raise ECONNRESET on the client side under similar conditions of the server not reading all the data.  It's much harder to trigger, I needed to send ~100kb of data from the client.  But it seems to fail in the same way.

So I don't think this fix is needed / correct.  Instead the client needs to be able to handle ECONNRESET when it tries to read data.

dpgeorge:
Here's my test showing that MicroPython (unix port at least!) has the same behaviour as CPython:
```py
# Test behaviour of read on a stream that closed.

import sys
import asyncio

PORT = 8000

N = 130000


async def handle_connection(reader, writer):
    # Change the "if" to run separate tests:
    # - if 1: read 1 byte and leave the rest, triggers ECONNRESET on client side
    # - if 0: read all data, client will finish cleanly
    if 1:
        print("read 1 byte")
        sys.stdout.flush()
        data = await reader.read(1)
    else:
        n = 0
        while n < N:
            data = await reader.read(10_000)
            n += len(data)
            print("read", len(data), n)
            sys.stdout.flush()
            await asyncio.sleep(0.05)

    print("close")
    writer.close()
    await writer.wait_closed()

    print("done")
    ev.set()


async def tcp_server():
    global ev
    ev = asyncio.Event()
    server = await asyncio.start_server(handle_connection, "0.0.0.0", PORT)
    print("server running")
    multitest.next()
    async with server:
        await asyncio.wait_for(ev.wait(), 10)


async def tcp_client():
    reader, writer = await asyncio.open_connection(IP, PORT)
    print(f"write {N} bytes")
    sys.stdout.flush()
    writer.write(b"x" * N)
    await writer.drain()
    await asyncio.sleep(0.5)
    try:
        print("read", await reader.read(1))
    except OSError as er:
        print("read error", er.errno)


def instance0():
    multitest.globals(IP=multitest.get_network_ip())
    asyncio.run(tcp_server())


def instance1():
    multitest.next()
    asyncio.run(tcp_client())
```

There are two scenarios that test can test:

#### "if 1": server will read 1 byte and leave the rest, causing client to have ECONNRESET
The output looks like this (the 104 is ECONNRESET):
```
$ ./run-multitests.py -c multi_net/asyncio_tcp_server_client_drain.py
multi_net/asyncio_tcp_server_client_drain.py on micropython|micropython:
TRACE micropython|micropython:
   105 i0 : SET IP = '10.0.0.127'
   105 i0 : server running
   105 i0 : NEXT
   207 i1 : NEXT
   207 i0 : read 1 byte
   207 i1 : write 130000 bytes
   208 i0 : close
   209 i0 : done
   616 i1 : read error 104
TRACE python3|python3:
   104 i0 : SET IP = '10.0.0.127'
   104 i0 : server running
   104 i0 : NEXT
   206 i1 : NEXT
   206 i0 : read 1 byte
   206 i1 : write 130000 bytes
   206 i0 : close
   207 i0 : done
   716 i1 : read error 104
pass
1 tests performed
1 tests passed
```
Both unix MicroPython and CPython clien have the ECONNRESET in this case.  They both behave exactly the same.

#### "if 0": server will read all incoming data, client will finish cleanly
The output looks like this:
```
$ ./run-multitests.py -c multi_net/asyncio_tcp_server_client_drain.py
multi_net/asyncio_tcp_server_client_drain.py on micropython|micropython:
TRACE micropython|micropython:
   105 i0 : SET IP = '10.0.0.127'
   105 i0 : server running
   105 i0 : NEXT
   208 i1 : NEXT
   208 i0 : read 10000 10000
   208 i1 : write 130000 bytes
   208 i0 : read 10000 20000
   312 i0 : read 10000 30000
   314 i0 : read 10000 40000
   315 i0 : read 10000 50000
   419 i0 : read 10000 60000
   420 i0 : read 10000 70000
   525 i0 : read 10000 80000
   526 i0 : read 10000 90000
   630 i0 : read 10000 100000
   631 i0 : read 10000 110000
   735 i0 : read 10000 120000
   737 i0 : read 10000 130000
   840 i0 : close
   841 i1 : read b''
   841 i0 : done
TRACE python3|python3:
   104 i0 : SET IP = '10.0.0.127'
   104 i0 : server running
   104 i0 : NEXT
   206 i1 : NEXT
   206 i0 : read 10000 10000
   206 i1 : write 130000 bytes
   309 i0 : read 10000 20000
   310 i0 : read 10000 30000
   414 i0 : read 10000 40000
   415 i0 : read 10000 50000
   519 i0 : read 10000 60000
   521 i0 : read 10000 70000
   522 i0 : read 10000 80000
   627 i0 : read 10000 90000
   628 i0 : read 10000 100000
   732 i0 : read 10000 110000
   733 i0 : read 10000 120000
   837 i0 : read 10000 130000
   837 i1 : read b''
   837 i0 : close
   839 i0 : done
pass
1 tests performed
1 tests passed
```
Both MicroPython and CPython behave exactly the same here.

dpgeorge:
Given the above, it looks like MicroPython (asyncio) is behaving correctly here (at least, as per CPython).  So I will close this PR.

The ECONNRESET needs to be handled at the layer above asyncio, eg in Microdot, or in user code.

(Feel free to discuss/reopen the PR if you think the above analysis is incorrect.)

PR #13546, opened 2024-01-29
extmod/modssl_mbedtls: Fix cipher iteration in SSLContext.get_ciphers.

dpgeorge:
Prior to this commit it would skip every second cipher returned from mbedtls.

The corresponding test is also updated and now passes on esp32, rp2, stm32 and unix.


dpgeorge:
@Carglglz FYI

Carglglz:
@dpgeorge  Thanks for this! 😓 I see now that the leftover code comes from the previous while loop that I didn't realise it was double counting now in the for loop. 🤦🏼‍♂️

PR #13555, opened 2024-01-30
rp2: Change machine.I2S and rp2.DMA to use shared DMA IRQ handlers.

dpgeorge:
These separate drivers must share the DMA resource with each other.

The rp2.DMA test is also updated to be more reliable.

Fixes issue #13380.


dpgeorge:
@nickovs FYI

PR #13556, opened 2024-01-30
Fix compilation warnings for Windows 64 bits

srinsoz:
Hi

Here is some missing cast and other minor change that prevent compilation warning if ````mp_int_t```` is defined as integer 64 bits.

### DIFF
--- a/shared/timeutils/timeutils.h
+++ b/shared/timeutils/timeutils.h
@@ -60,7 +60,7 @@ mp_uint_t timeutils_mktime_2000(mp_uint_t year, mp_int_t month, mp_int_t mday,
 
 static inline void timeutils_seconds_since_epoch_to_struct_time(uint64_t t, timeutils_struct_time_t *tm) {
     // TODO this will give incorrect results for dates before 2000/1/1
-    return timeutils_seconds_since_2000_to_struct_time(t - TIMEUTILS_SECONDS_1970_TO_2000, tm);
+    timeutils_seconds_since_2000_to_struct_time(t - TIMEUTILS_SECONDS_1970_TO_2000, tm);
 }
 
 static inline uint64_t timeutils_mktime(mp_uint_t year, mp_int_t month, mp_int_t mday, mp_int_t hours, mp_int_t minutes, mp_int_t seconds) {
--- a/extmod/modrandom.c
+++ b/extmod/modrandom.c
@@ -86,7 +86,7 @@ STATIC uint32_t yasmarang_randbelow(uint32_t n) {
 #endif
 
 STATIC mp_obj_t mod_random_getrandbits(mp_obj_t num_in) {
-    int n = mp_obj_get_int(num_in);
+    mp_int_t n = mp_obj_get_int(num_in);
     if (n > 32 || n < 0) {
         mp_raise_ValueError(MP_ERROR_TEXT("bits must be 32 or less"));
     }
@@ -111,7 +111,7 @@ STATIC mp_obj_t mod_random_seed(size_t n_args, const mp_obj_t *args) {
     } else {
         seed = mp_obj_get_int_truncated(args[0]);
     }
-    yasmarang_pad = seed;
+    yasmarang_pad = (uint32_t)seed;
     yasmarang_n = 69;
     yasmarang_d = 233;
     yasmarang_dat = 0;
@@ -126,7 +126,7 @@ STATIC mp_obj_t mod_random_randrange(size_t n_args, const mp_obj_t *args) {
     if (n_args == 1) {
         // range(stop)
         if (start > 0) {
-            return mp_obj_new_int(yasmarang_randbelow(start));
+            return mp_obj_new_int(yasmarang_randbelow((uint32_t)start));
         } else {
             goto error;
         }
@@ -135,7 +135,7 @@ STATIC mp_obj_t mod_random_randrange(size_t n_args, const mp_obj_t *args) {
         if (n_args == 2) {
             // range(start, stop)
             if (start < stop) {
-                return mp_obj_new_int(start + yasmarang_randbelow(stop - start));
+                return mp_obj_new_int(start + yasmarang_randbelow((uint32_t)(stop - start)));
             } else {
                 goto error;
             }
@@ -151,7 +151,7 @@ STATIC mp_obj_t mod_random_randrange(size_t n_args, const mp_obj_t *args) {
                 goto error;
             }
             if (n > 0) {
-                return mp_obj_new_int(start + step * yasmarang_randbelow(n));
+                return mp_obj_new_int(start + step * yasmarang_randbelow((uint32_t)n));
             } else {
                 goto error;
             }
@@ -167,7 +167,7 @@ STATIC mp_obj_t mod_random_randint(mp_obj_t a_in, mp_obj_t b_in) {
     mp_int_t a = mp_obj_get_int(a_in);
     mp_int_t b = mp_obj_get_int(b_in);
     if (a <= b) {
-        return mp_obj_new_int(a + yasmarang_randbelow(b - a + 1));
+        return mp_obj_new_int(a + yasmarang_randbelow((uint32_t)(b - a + 1)));
     } else {
         mp_raise_ValueError(NULL);
     }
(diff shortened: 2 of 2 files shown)


stinos:
Can you explain how you're building for windows i.e. where these warnings prevent compilation? Things is: I'm pretty sure there are quite a lot more locations where these truncating assignments happen so I'm wondering why you don't have issues there.

srinsoz:
For the context : We are looking to integrate micropython in our products. We have our own GNU makefile that compile the C/C++ code using Microsoft cl compiler. 


Here is the compile command with the output of ``extmod/modrandom.c``
 
````
compiling 64 bits windows/amd64/objs/extmod/modrandom.obj
cl /nologo /permissive- /GS /W3 /WX- /O2 /Oi /Oy- /GL /D "WIN32" /D "NDEBUG" /D "_CONSOLE" /D "_MBCS" /D "_USING_V110_SDK71_" /D "VIRTUAL_HUB" /Gm- /sdl /EHsc /MT /GS /Gy /Zc:inline /fp:precise /Zc:wchar_t /Zc:forScope /analyze- /Gd /Iwindows/amd64/objs/ /I../../../yoctolib/v2.0/Public/cpp/Sources/yapi/mbedtls/library /I../../../mpy/micropython /I../../../yoctolib/v2.0/Public/cpp/Sources/yapi/mbedtls/include /Fowindows/amd64/objs/extmod/modrandom.obj /c ../../../mpy/micropython/extmod/modrandom.c
modrandom.c
../../../mpy/micropython/extmod/modrandom.c(89): warning C4244: 'initializing': conversion from 'mp_int_t' to 'int', possible loss of data
../../../mpy/micropython/extmod/modrandom.c(114): warning C4244: '=': conversion from 'mp_uint_t' to 'uint32_t', possible loss of data
../../../mpy/micropython/extmod/modrandom.c(129): warning C4244: 'function': conversion from 'mp_int_t' to 'uint32_t', possible loss of data
../../../mpy/micropython/extmod/modrandom.c(138): warning C4244: 'function': conversion from 'mp_int_t' to 'uint32_t', possible loss of data
../../../mpy/micropython/extmod/modrandom.c(154): warning C4244: 'function': conversion from 'mp_int_t' to 'uint32_t', possible loss of data
../../../mpy/micropython/extmod/modrandom.c(170): warning C4244: 'function': conversion from 'mp_int_t' to 'uint32_t', possible loss of data
../../../mpy/micropython/extmod/modrandom.c(180): warning C4244: 'function': conversion from 'mp_int_t' to 'uint32_t', possible loss of data
````

And the definition of mp_int_t
````cpp
// type definitions for the specific machine
#ifdef __64BITS__
    typedef signed long long mp_int_t;
    typedef unsigned long long mp_uint_t;
    #define UINT_FMT "%"FMTu64
    #define INT_FMT "%"FMTu64
#else
	....
#endif
````

**Comment by stinos** [comment]
The maximum warning level which works with msvc is `/W1`. Or maybe `/W2` but I doubt it otherwise it would have been set like that in ports/windows/common.props probably. What we do to workaround when embedding is compiling all micropython code with its own flags and further guard the headers like

```
#ifdef _MSC_VER
#pragma warning ( disable : 4200 ) //nonstandard extension used : zero-sized array in struct/union
#endif
#ifdef __cplusplus
extern "C"
{
#endif
#include <py/objfun.h>
#include <py/objint.h>
#include <py/objmodule.h>
#include <py/objtype.h>
#include <py/runtime.h>
#ifdef __cplusplus
}
#endif
#ifdef _MSC_VER
#pragma warning ( default : 4200 )
#endif
```

Anyway, I'm still wondering how it's possible you only have to fix 2 files to get this working? For reference if I use `/W3` I get >200 warnings from >10 different files, all core micropython code. Hence my view is that either these should all be fixed, or none.

**Comment by srinsoz** [comment]
Yes we have ~100 more warnings. I will look at it next week, I will probably change the warning level for the compilation of micropyhton at the end.  

These ones where prety safe and obvious. so I've decided to push them on the main repo.

**Comment by projectgus** [comment]
Is it going to be feasible to update the MSVC guards and the CI Windows build to keep these type of warnings fixed in the future, after all of the warning fixes are applied?

**Comment by dpgeorge** [comment]
Thanks for the contribution.

> Is it going to be feasible to update the MSVC guards and the CI Windows build to keep these type of warnings fixed in the future, after all of the warning fixes are applied?

Yes, this is also my thought.  These issues will creep back into the code if there isn't some CI to prevent them.  Can we get similar warnings from gcc/clang with some combination of compiler flags, maybe?

File: shared/timeutils/timeutils.h [hunk 1]
```diff
@@ -60,7 +60,7 @@ mp_uint_t timeutils_mktime_2000(mp_uint_t year, mp_int_t month, mp_int_t mday,
 
 static inline void timeutils_seconds_since_epoch_to_struct_time(uint64_t t, timeutils_struct_time_t *tm) {
     // TODO this will give incorrect results for dates before 2000/1/1
-    return timeutils_seconds_since_2000_to_struct_time(t - TIMEUTILS_SECONDS_1970_TO_2000, tm);
+    timeutils_seconds_since_2000_to_struct_time(t - TIMEUTILS_SECONDS_1970_TO_2000, tm);
```
dpgeorge, line 63: I was under the impression that this is valid C code, to return a void value in a void function.

projectgus: @dpgeorge C99 says not so:

> 6.8.6.4 The return statement
> Constraints
> A return statement with an expression shall not appear in a function whose return type is void. A return statement without an expression shall only appear in a function whose return type is void.

dpgeorge: Well, it does compile fine with `gcc -std=c99`.


> A return statement with an expression shall not appear in a function whose return type is void.

Technically this sentence doesn't apply here because `timeutils_seconds_since_2000_to_struct_time()` is not an expression, it doesn't return anything....?

dpgeorge: The reason I care is because at some point I wanted to replace all `mp_raise_XXX()` calls with `return mp_raise_XXX()` (to get rid of NLR), and that would sometimes introduce this case of return-void-expression.

projectgus: > Technically this sentence doesn't apply here because timeutils_seconds_since_2000_to_struct_time() is not an expression, it doesn't return anything....?

I might be misreading the standard, but I believe this is a *void expression (an expression that has type **void**)* as per section 6.3.2.2 void.

Section 6.5 Expressions is also illustrative, particularly the first point:

> 1. An expression is a sequence of operators and operands that specifies computation of a value, or that designates an object or a function, or that generates side effects, or that performs a combination thereof.

> Well, it does compile fine with gcc -std=c99.

Looks like it's enabled by `-pedantic`: https://gcc.godbolt.org/z/qzTejd66x

For my 2c, it doesn't seem like it'd necessarily be problematic to explicitly allow this pattern in MicroPython. Although maybe it's worth noting such exceptions in the coding style docs somewhere. Also, I don't know if there's a way to make MSVC allow it...

dpgeorge: > Looks like it's enabled by `-pedantic`

Yes you are right.

OK, I'm happy to change this then, so it doesn't return a void expression.

Side note: `-pedantic` raises a lot of warnings in the core code (eg when enabled for `bare-arm`).  I'm not sure we can eliminate all those warnings.

projectgus: > Side note: -pedantic raises a lot of warnings in the core code (eg when enabled for bare-arm). I'm not sure we can eliminate all those warnings.

I don't think we need to, those warnings are classed pedantic for a reason. :grin:

stinos:
> Can we get similar warnings from gcc/clang with some combination of compiler flags, maybe?

Yes, but IIRC maybe not all exactly the same as msvc but definitely close. There's `-Wconversion` for starters, but be prepared for a lot of red text :)

dpgeorge:
> There's `-Wconversion` for starters, but be prepared for a lot of red text :)

I had a go at enabling this warning on the `bare-arm` port, and indeed it gives a lot of warnings.  But actually it finds some interesting things that maybe do need improving.  Eg `qstr_add()` should have the type of its first argument as `qstr_len_t`, not `mp_uint_t`.

So I'd be interested to start fixing things that `-Wconversion` brings up, and eventually enable that warning on CI.

PR #13564, opened 2024-01-30
ARDUINO_PORTENTA_H7/mpconfigboard.h: Configure SPI1.

loopback:
Currently, only the processor's SPI2 bus is enabled (though the related pins are labeled SPI1 in the Portenta H7 documentation). This enables the processor's SPI1 bus, which is accessible via the board's high-density connectors.

### DIFF
--- a/ports/stm32/boards/ARDUINO_PORTENTA_H7/mpconfigboard.h
+++ b/ports/stm32/boards/ARDUINO_PORTENTA_H7/mpconfigboard.h
@@ -170,7 +170,12 @@ extern struct _spi_bdev_t spi_bdev;
 #define MICROPY_HW_I2C3_SCL         (pin_H7)
 #define MICROPY_HW_I2C3_SDA         (pin_H8)
 
-// SPI
+// SPI buses
+#define MICROPY_HW_SPI1_NSS         (pin_C13)
+#define MICROPY_HW_SPI1_SCK         (pin_B3)
+#define MICROPY_HW_SPI1_MISO        (pin_B4)
+#define MICROPY_HW_SPI1_MOSI        (pin_D7)
+
 #define MICROPY_HW_SPI2_NSS         (pin_I0)
 #define MICROPY_HW_SPI2_SCK         (pin_I1)
 #define MICROPY_HW_SPI2_MISO        (pin_C2)


dpgeorge:
@iabdalkader when you get a chance can you please check that this is OK?

iabdalkader:
This is okay I guess, but note that some of those pins are already used for the SD card (SDMMC2 ), on both the breakout and shields, so you can't use them both at the same time. If it's okay to enable this then it's fine @dpgeorge 

```C
// SD Card SDMMC
#define MICROPY_HW_SDCARD_SDMMC     (2)
#define MICROPY_HW_SDCARD_CK        (pin_D6)
#define MICROPY_HW_SDCARD_CMD       (pin_D7)
#define MICROPY_HW_SDCARD_D0        (pin_B14)
#define MICROPY_HW_SDCARD_D1        (pin_B15)
#define MICROPY_HW_SDCARD_D2        (pin_B3)
#define MICROPY_HW_SDCARD_D3        (pin_B4)
#define MICROPY_HW_SDCARD_MOUNT_AT_BOOT (0)
```
>> though the related pins are labeled SPI1 in the Portenta H7 documentation

Arduino's pinout is usually logical not physical, the first and only SPI is SPI1 (even thought it's actually SPI2), I know it's confusing, but can't fix it.

loopback:
Thanks, @iabdalkader.

> note that some of those pins are already used for the SD card (SDMMC2 ), on both the breakout and shields, so you can't use them both at the same time. If it's okay to enable this then it's fine @dpgeorge

I think this sort of thing is pretty common with the major micropython boards. As a user, I favor having all the functions available for the pins that are exposed in the headers or high-density connectors, and view making mutually-exclusive choices in how to use them as part of the development process.

> Note that `sdcard_init` is called on boot, unconditionally, so the pins will be configured as SDMMC2 first.

I tested this with a Portenta H7 and verified the SPI bus works properly on those pins once it is initialized.

>  > though the related pins are labeled SPI1 in the Portenta H7 documentation
> 
> @loopback Arduino's pinout is usually logical not physical, the first and only SPI is SPI1 (even thought it's actually SPI2), 

Understood. Thanks for clarifying.

dpgeorge:
OK, let's go with this then.  It doesn't change anything unless you explicitly use `machine.SPI(1)`.



> Arduino's pinout is usually logical not physical, the first and only SPI is SPI1 (even thought it's actually SPI2), I know it's confusing, but can't fix it.

@iabdalkader you can name SPI's, so you could do something like:
```c
MICROPY_HW_SPI2_NAME "SPI1"
```
Then it can be used as `machine.SPI("SPI1")`.  It's confusing but maybe less confusing that accessing it as `machine.SPI(2)`.

iabdalkader:
> Then it can be used as `machine.SPI("SPI1")`. It's confusing but maybe less confusing that accessing it as `machine.SPI(2)`.

Yeah I know, but I think it will just be more confusing this way.

PR #13567, opened 2024-01-31
ci: Move codespell into GitHub action workflow, apply codespell version.

projectgus:
* Similar to ruff.yaml, it's simpler to run the `codespell` command directly from a workflow file.
* Applies a specific version to codespell, same as pre-commit (introduced in #12157)

~~Additional, optional commit in this branch:~~

EDIT: Dropped this commit as `yq` is only packaged in Ubuntu starting from "lunar", 23.04.

* ~~Read the CI tool versions (ruff, codespell) from `.pre-commit-config.yaml`. This means there is now a single canonical source for these versions, at the cost of some ugly "yq" YAML parsing in the workflow steps.~~

~~Can drop the second commit if it's too fiddly, and~~ will add comments instead to remind anyone updating that the version is copied in two places.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,12 @@
+name: Check spelling with codespell
+
+on: [push, pull_request]
+
+jobs:
+  codespell:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - run: pip install --user codespell==2.2.6 tomli
+    - run: codespell
+
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -7,6 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    # codespell version should be kept in sync with .pre-commit-config.yml
     - run: pip install --user codespell==2.2.6 tomli
     - run: codespell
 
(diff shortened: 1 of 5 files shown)


projectgus:
@Josverl This overlaps with some of the (very useful) tooling & CI improvements you've been making, so PTAL.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,12 @@
+# https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python
```
dpgeorge, line 1: Why is this comment needed?  Other workflow files don't have it.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,12 @@
...
+name: Check spelling with codespell
+on: [push, pull_request]
+jobs:
```
dpgeorge, line 4: Please put blank lines around the top-level sections, to be consistent with, eg, `code_formatting.yml`.

.github/workflows/codespell.yml
projectgus: I copied it from ruff.yml which has it, but actually both these workflow files don't use much of what's documented there.

Will remove from both.

.github/workflows/codespell.yml
projectgus: As above, this file is a copy of ruff.yml. I'll change them both.

dpgeorge: Oh, I see.  Yeah, would be good to change both.

PR #13568, opened 2024-01-31
examples/embedding: Add -fno-common to the sample compiler flags.

projectgus:
This makes no difference when files are linked directly into a target application, but on macOS additional steps are needed to index common symbols in static libraries. See https://stackoverflow.com/a/26581710

By not creating any common symbols, this problem is bypassed.

This will also trigger linker errors if there are cases where the same symbol is defined in the host application.

Alternative to the fix proposed in #13498 and confirmed to work by @Shootfast.


dpgeorge:
Thanks, this looks good.

PR #9497, opened 2022-10-04
tinyusb: Support USB device drivers written in Python

projectgus:
The goal is to allow implementing USB device drivers in Python via TinyUSB at runtime, in addition to any existing compile-time TinyUSB class drivers (i.e. CDC and optionally MSC).

Python API with the more interesting user-facing code is at https://github.com/micropython/micropython-lib/pull/558 - if checking out this branch in micropython, go to the `lib/micropython-lib` subdirectory and check out the `feature/usbd_python` branch there as well. 

## Port Support

This PR implements this for the `rp2` and `samd` ports.

The port-level requirement is that the TinyUSB task is run in a "scheduled" static callback node, as implemented in #12846 for rp2 & samd.) I think this approach can be extended to most of the other ports pretty easily.

Challenging ports (future work):

* esp32. ESP-IDF runs TinyUSB in its own FreeRTOS Task. I think it can be made to work here by using a cross-task Python callback (see jimmo's [implementation here](https://github.com/micropython/micropython/pull/8945/commits/843204c717b7c8446398fffe4ebcdb0bd13765df)), or by changing the TinyUSB implementation to work similar to the other ports (this will save some RAM and on single core ESP32s it might even improve performance. On dual core it might reduce performance a little but it also might not.)
* stm32. Currently doesn't use TinyUSB. It may be possible to port this approach to the ST USB stack, but prefer to make a compile-time option to build stm32 port with TinyUSB. This will be less work and more maintainable into the future.
* zephyr. The Zephyr [USB device driver API](https://docs.zephyrproject.org/latest/connectivity/usb/device/api/usb_device.html) (example: [hid core](https://github.com/zephyrproject-rtos/zephyr/blob/main/subsys/usb/device/class/hid/core.c)) looks like it would adapt to this approach easily, but haven't verified this yet.

## Binary Size Impact

Where possible, the implementation has been placed into micropython-lib to reduce the baseline binary size impact. If enabled at compile time (currently not the default) the code size change is:

| Port | Board | Code Size  | Delta | Static RAM Size| Delta |
| ------ | -------- | -------------------- | ------- | ------------ |  ---- |
| rp2 | RPI_PICO | 322808 -> 325704 | +2896| 5872 -> 5876 | +4 |
| samd | SEEED_WIO_TERMINAL (samd51) | 261136 -> 263984 | +2848 | 3068 -> 3072 | +4 |

Note: This doesn't include freezing in the "usbd" module from micropython-lib. Currently that module is approximately 15KB.

[...]

PR #13571, opened 2024-01-31
all tinyusb ports: On cold boot, enable USB after boot.py completes.

projectgus:
This implements similar behaviour to the stm32 port.

Currently this doesn't add any useful functionality (and may break workflows that depend on USB-CDC being live in boot.py), however it's a precondition for more usable workflows with USB devices defined in Python (allows setting up USB interfaces in boot.py before the device enumerates for the first time).

#9497 is the PR for the eventual dynamic USB support on TinyUSB. Currently this is only implemented for samd & rp2, but changing it for all TinyUSB ports here maintains consistency and makes it easier to expand dynamic USB support to the other ports in the future.

## Breaking Changes

The documented advice for MicroPython is to put only early initialisation code into `boot.py` and the main application into `main.py`. For these cases it's unlikely this change will break anything. (Specifically: USB-CDC is not initialised and open during an early cold boot, so any output from a short-running `boot.py` after hard reset won't ever have been available on the USB-CDC port.)

However, the *de facto* situation for most of these ports is that `boot.py` and `main.py` were interchangeable until now. Any application that put all of its functionality into `boot.py` may find that it no longer works as expected.

## Testing

I only have hardware to test this on `rp2` and `samd`, so I haven't tested this on mimxrt, nrf or renesas-ra.

This work was funded through GitHub Sponsors.

### DIFF
--- a/ports/rp2/main.c
+++ b/ports/rp2/main.c
@@ -38,7 +38,7 @@
 #include "shared/runtime/gchelper.h"
 #include "shared/runtime/pyexec.h"
 #include "shared/runtime/softtimer.h"
-#include "tusb.h"
+#include "shared/tinyusb/mp_usbd.h"
 #include "uart.h"
 #include "modmachine.h"
 #include "modrp2.h"
@@ -86,11 +86,8 @@ int main(int argc, char **argv) {
     #endif
     #endif
 
-    #if MICROPY_HW_ENABLE_USBDEV
-    #if MICROPY_HW_USB_CDC
+    #if MICROPY_HW_ENABLE_USBDEV && MICROPY_HW_USB_CDC
     bi_decl(bi_program_feature("USB REPL"))
-    #endif
-    tusb_init();
     #endif
 
     #if MICROPY_PY_THREAD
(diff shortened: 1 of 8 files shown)


robert-hh:
Delaying USB init after running boot.py was run can be very useful. I implemented similar in the (closed) PR for MSC support, calling tusb_init() after running _boot.py, allowing to configure MSC support in _boot.py.
I can make tests with mimxrt, nrf and renesas boards.

ports/rp2/main.c
```diff
@@ -181,6 +180,9 @@ int main(int argc, char **argv) {
 
         // Execute user scripts.
         int ret = pyexec_file_if_exists("boot.py");
+
+        tusb_init();
```
dpgeorge, line 184: This needs to be guarded with `#if MICROPY_HW_ENABLE_USBDEV`.

projectgus: Oops, thanks!

projectgus:
Thanks @robert-hh for testing nrf and mimxrt, much appreciated.

projectgus:
Temporarily mark this as a draft as @dpgeorge noted that TinyUSB is always enabled, even if the case where there are zero interfaces defined (i.e. runtime USB devices are on, but the user didn't initialise any). It should instead not enable TinyUSB at all in this case.

We need to filter the call to `tusb_init()` in this case, so I'll introduce a (no-op) macro here and then override it in the runtime USB PR.

projectgus:
OK, ready again. This updated PR should produce 100% same binary output as before, but there's now a wrapper function `md_usbd_init()` that can be made more complex when we add runtime USB support (and therefore may want to enable/disable TinyUSB based on the state of the runtime USB device.)

dpgeorge:
Thanks, now merged.

PR #13576, opened 2024-01-31
core: Throw an exception for invalid int literals like "01".

### LINKED ISSUE #13342: CPython incompatible: should error on integer literals like "01"
Recently I noticed that micropython accepts literals like `01` in a source file, while standard Python rejects them saying "SyntaxError: leading zeros in decimal integer literals are not permitted; use an 0o prefix for octal integers".

However, `int("01")` should succeed, parsing the number as decimal.

I implemented this check in a CircuitPython pull request (https://github.com/adafruit/circuitpython/pull/8783) but it feels a bit clunky so I did not pull request the same change in micropython, but filed this issue instead.

jepler:
Including making int("01") parse in base 10 like standard Python

Closes: #13342


dpgeorge:
That's a lot of code size increase for a very minor feature... I guess simplifying the error message string would be a good way to make it smaller.  But still it's larger than I would expect.

dpgeorge:
For some reason the nanbox CI is failing here.

jepler:
Ah, the nanbox test has `int("0x...")` which is now an error since the default radix of `int()` was modified to be 10 to match cpython, rather than being 0.

jepler:
This is re-worked to use less code and hopefully green now.

Please feel free to close this up if the cost is still too high and you don't have an obvious suggestion to make to reduce the code growth.

The nanbox test fix might be worth taking regardless.

jepler:
Please feel free to close this up if the error detecting isn't worth the trouble/code size.

dpgeorge:
I think this fix is worth getting in, and I think I can reduce code size:

Change the `is_all_zeros()` function to:
```c
static bool is_all_zeros(const char *str, size_t len) {
    while (len--) {                                                        
        if (*str++ != '0') {                         
            return false;
        }                                
    }
    return true;                                                            
}                 
```
and then the adjusted logic in `mp_parse_num_base()` is:
```c
        } else {     
            if (*base == 0 && is_all_zeros(str, len)) {
                *base = 10;
            }             
            p -= 2;
        }                                                                    
```
@jepler do you want to make that change and see if the CI still all passes?

jepler:
Sadly, assuming I implemented your suggestion correctly, it didn't pass the tests locally.

```diff
FAILURE /home/jepler/src/micropython/tests/results/basics_int1.py
--- /home/jepler/src/micropython/tests/results/basics_int1.py.exp	2025-01-22 08:15:13.045944568 -0600
+++ /home/jepler/src/micropython/tests/results/basics_int1.py.out	2025-01-22 08:15:13.045944568 -0600
@@ -69,6 +69,6 @@
 ValueError
 ValueError
 ValueError
-ValueError
+1
 31
 TypeError

FAILURE /home/jepler/src/micropython/tests/results/basics_lexer.py
--- /home/jepler/src/micropython/tests/results/basics_lexer.py.exp	2025-01-22 08:15:12.849942152 -0600
+++ /home/jepler/src/micropython/tests/results/basics_lexer.py.out	2025-01-22 08:15:12.853942202 -0600
@@ -23,5 +23,8 @@
 SyntaxError
 SyntaxError
 SyntaxError
-0
-SyntaxError
+Traceback (most recent call last):
+  File "/home/jepler/src/micropython/tests/basics/lexer.py", line 88, in <module>
+  File "<string>", line 1
+SyntaxError: invalid syntax for integer with base 0: '00'
+CRASH
\ No newline at end of file
```

```diff
diff --git a/py/parsenumbase.c b/py/parsenumbase.c
index a706b7cfb0..683b4e35a7 100644
--- a/py/parsenumbase.c
+++ b/py/parsenumbase.c
@@ -30,7 +30,7 @@
 #include "py/runtime.h"
 
 static bool is_all_zeros(const char *str, size_t len) {
-    while (len-- && unichar_isdigit(*str)) {
+    while (len--) {
         if (*str++ != '0') {
             return false;
         }
@@ -56,13 +56,10 @@ size_t mp_parse_num_base(const char *str, size_t len, int *base) {
         } else if (*base == 0 && (c | 32) == 'b') {
             *base = 2;
         } else {
-            if (*base == 0) {
+            if (*base == 0 && !is_all_zeros(str, len)) {
                 *base = 10;
             }
             p -= 2;
-            if (!is_all_zeros(str, len)) {
-                *base = 0;
-            }
         }
     } else if (*base == 8 && c == '0') {
         c = *(p++);
```

dpgeorge:
You need to remove the `!` in front of the call to `is_all_zeros()`.

jepler:
Updated.

jepler:
ah I think the new code doesn't match cpython behavior exactly?

micropython, this change:
```
>>> int("000 ", 0)
ValueError: invalid syntax for integer with base 0: '000 '
```

standard python:
```
ValueError: invalid literal for int() with base 0: '018 '
>>> int("000 ", 0)
0
```

dpgeorge:
Hmm, you're right, looks like the behaviour of `int("000 ", 0)` is a regression with the current state of this PR.  I guess we will need a test for that case.

dpgeorge:
A simpler way to handle `"000"` and reject invalid bases might be to set `*base = 1` instead of `*base = 10` for the case the base is 0.  Then the integer parsing routine will only accept the character 0 in the number.

py/parsenumbase.c
```diff
@@ -27,38 +27,31 @@
 #include "py/mpconfig.h"
 #include "py/misc.h"
 #include "py/parsenumbase.h"
+#include "py/runtime.h"
 
 // find real radix base, and strip preceding '0x', '0o' and '0b'
 // puts base in *base, and returns number of bytes to skip the prefix
+// puts 0 in *base to indicate an invalid C-style octal literal
```
jepler, line 34: this comment is now wrong

py/parsenum.c
```diff
@@ -151,13 +151,13 @@ mp_obj_t mp_parse_num_integer(const char *restrict str_, size_t len, int base, m
         raise_exc(exc, lex);
         #elif MICROPY_ERROR_REPORTING == MICROPY_ERROR_REPORTING_NORMAL
         mp_obj_t exc = mp_obj_new_exception_msg_varg(&mp_type_ValueError,
-            MP_ERROR_TEXT("invalid syntax for integer with base %d"), base);
+            MP_ERROR_TEXT("invalid syntax for integer with base %d"), base == 1 ? 10 : base);
```
jepler, line 154: wait, should this be `base == 1 ? 0 : base`?

jepler (commented):
it's late so consider this a note to self...

jepler:
@dpgeorge How do you feel about where things are at now?

py/parsenumbase.c
```diff
@@ -27,38 +27,32 @@
 #include "py/mpconfig.h"
 #include "py/misc.h"
 #include "py/parsenumbase.h"
+#include "py/runtime.h"
```
dpgeorge, line 30: I don't think this header is needed anymore.

tests/basics/int1.py
```diff
@@ -79,6 +81,8 @@ def test(value, base):
 test('0xg', 16)
 test('1 1', 16)
 test('123', 37)
+test('01', 0)
+test('000 ', 0)
```
dpgeorge, line 85: This test is supposed to pass, so I suggest it goes above as simply `print(int('000 ', 0))`.  Also maybe test with non-zero bases as well:
```py
print(int('000 ', 10))
print(int('000 ', 16))
```

jepler:
Updated again. Want me to squash it all too?

dpgeorge:
> Want me to squash it all too?

Yes please (otherwise I can do it during merge).

dpgeorge (approved):

dpgeorge:
I like how you reorganised the logic in `mp_parse_num_base()`.  It's a lot cleaner now and easier to understand.

jepler:
I've rebased & squashed it.

Thanks for taking the time with me to make this the best change it could be.

dpgeorge:
Rebased and merged in 7b3f189b1723fe642f122a3b7826d16fe32f801a and 13b13d1fdd05549d504eeded0b5aa8871d5e5dcf

jepler:
I belatedly notice that gcc (x86_64 and arm, -Os) both optimize `(b == 0 || b == 16`) into `!(b & ~16)` so that readability reduction was not needed to get the code size savings :-/

robert-hh:
> readability reduction was not needed to get the code size savings

That's my general rule: Make code readable and leave optimization to the compiler.

dpgeorge:
Yes, I was wondering about that optimisation, if the compiler could do it.

It's not to late to change it in a follow up PR, if you'd like to.

PR #13583, opened 2024-02-02
webassembly: Add proxying between Python and JavaScript objects, and change top-level interface for PyScript use

dpgeorge:
This PR extends, improves and changes the webassembly port so that it's suitable for use in PyScript, (https://pyscript.net/, https://github.com/pyscript/pyscript).

The main things in this PR are:
- Add transparent proxying of Python and JavaScript objects between JavaScript and Python.  That allows passing objects and making function/method calls between the two languages.
- Change the top-level interface for this port to generate a `micropython.mjs` module, rather than `micropython.js`.  This module exposes a `loadMicroPython()` function which returns an object that can be used to execute Python, access the Python globals dict, and other things.
- Improve `run-tests.py` so it can automatically run webassembly tests via node, both .py tests and .js/.mjs tests.
- Add a set of .js/.mjs tests to test the proxying in the webassembly port.
- Enhance the `random` and `time` modules on the webassembly port.

A lot of the proxying and top-level interface is made to match the API of Pyodide (https://github.com/pyodide/pyodide/).

TODO:
- [x] remove remaining stray TODO's and commented-out code
- [x] finish cleaning up commits
- [x] possibly add support for variants, and make standard and pyscript variants


dpgeorge:
@ntoll @WebReflection FYI

ports/webassembly/pyscript/proxy_js.js
```diff
@@ -0,0 +1,266 @@
...
+
+function proxy_js_init() {
+    globalThis.proxy_js_ref = [globalThis];
+}
+
+function proxy_call_python(target, argumentsList) {
+    //console.debug("APPLY", target, argumentsList);
+    let args = 0;
+
+    // TODO: is this the correct thing to do, strip trailing "undefined" args?
```
WebReflection, line 17:
```js
while (argumentList.at(-1) === void 0) // or undefined
  argumentList.pop();
```

WebReflection: alternatively ...

```js
let { length } = argumentList;
while (length && argumentList[length - 1] === undefined) length--;

// drop all at once if needed
if (length !== argumentList.length)
  argumentList.splice(length);
```

WebReflection: ... or even terser ...

```js
let { length } = argumentList;
while (length-- && argumentList[length] === undefined);

// drop all at once if needed
if (++length !== argumentList.length)
  argumentList.splice(length);
```

you pick whatever you prefer :-)

ports/webassembly/pyscript/proxy_js.js
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+}
+
+const py_proxy_handler = {
+    isExtensible() {
+        return true;
+    },
+    has(target, prop) {
+        throw Error("has not implemented");
```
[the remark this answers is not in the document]
WebReflection: I am a bit confused (but hadn't had my coffee yet ...) ... no change was made to the Proxy traps and this specific error has been reported all over the place ... is this code meant to actually go or something still need to be changed? 🤔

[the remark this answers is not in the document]
dpgeorge: Oh, sorry, I meant to say that this bit was still to come.

The PR here is the current version that's published to npm.

But, that said, I've just now pushed a commit to implement `has`, along with a new test for this feature.  Let me know what you think about it.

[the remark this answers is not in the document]
WebReflection: I'll have a look, thank you!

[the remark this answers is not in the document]
WebReflection: I've no idea where that is though ... 😅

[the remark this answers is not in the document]
dpgeorge: If you look at the most recent commits, the 4 most recent ones implement `PyProxy.has` and `PyProxy.ownKeys` and add tests for these.

tests/ports/webassembly/promise_with_resolvers.mjs
```diff
@@ -0,0 +1,19 @@
+// Test polyfill of a method on a built-in.
+
```
WebReflection, line 2: this reminded me I should explicitly put a license [in here](https://github.com/ungap/with-resolvers) 🤣

jokes a part, glad my poly helped testing this stuff as we do use it all over the place 👍

WebReflection:
beside the terser suggestion to split all at once arguments (usually faster than N pop() but I don't think that's critical path for performance anyway) my only hint, from a JS maintainer point of view, is to somehow normalize the usage of `'` VS `"` or add a linter but it's also not a blocker and not so crucial ... overall this would be a :+1: to me and I can maybe help later on implementing that `deleteProperty` and `set` trap in the near future.

Thanks!

tests/ports/webassembly/promise_with_resolvers.mjs
dpgeorge: We take licensing/copyright seriously.  Should I add anything here?

dpgeorge:
> normalize the usage of `'` VS `"` or add a linter

Yes that would be a good thing to do.

What style do you use?  Can you recommend a light-weight linter?

tests/ports/webassembly/promise_with_resolvers.mjs
WebReflection: those 3 LOC aren't anything special and the purpose of that polyfill is to *die* ASAP (once all browsers support it) so don't worry about it, I haven't invented anything and the pattern is pretty simple too, the only trick there is the usage of `var` (on purpose) but that's also just part of the specs.

We're good, and you have the author of that poly stating that in here 👍

WebReflection:
we use *eslint* but not because we chose it (or I did) because it was there already and it's somehow the industry standard.

Although, I think these days I would instead consider [quick-lint-js](https://quick-lint-js.com/) or [biome](https://biomejs.dev/) as these are both faster and less opinionated or easier to configure / bootstrap.

As this project is used in *polyscript* I can at least point at how we configured it ... then again, mostly comes from original PyScript config: https://github.com/pyscript/polyscript/blob/main/.eslintrc.json

WebReflection:
@dpgeorge so I've built latest and things are pretty exciting ... but still there's something missing on the MicroPython side ... assuming the `micropython.mjs` is the one generated by this branch (which is in my case) I have this successfully working:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>MicroPython - Latest</title>
    <link rel="stylesheet" href="https://pyscript.net/releases/2024.1.3/core.css">
    <script type="module" src="https://pyscript.net/releases/2024.1.3/core.js"></script>
</head>
<body>
    <script>
        function ownKeys(py_or_js) {
            console.log("a" in py_or_js);
            console.log(Reflect.ownKeys(py_or_js));
        }
    </script>
    <mpy-config>
        interpreter = "http://localhost:8080/micropython.mjs"
    </mpy-config>
    <script type="mpy" worker>
        from pyscript import window
        import sys
        print(sys.version)

        window.ownKeys({ "a": 1 })
        # true
        # ["length", "a"]
        #  "length" should not be there but it's OK
    </script>
</body>
</html>
```

but as soon as I remove the `worker` attribute I have:

```
false
['clear', 'copy', 'get', 'items', 'keys', 'pop', 'popitem', 'setdefault', 'update', 'values', 'fromkeys', 'set', 'delete']
```

Which is somehow expected as I get all the dictionary keys but I don't have a way to convert via Python directly to JS objects, which happens instead automatically with workers, as those references are converted before being proxied around via the `PyProxy.toJs()` ... without any `ffi.to_js` on Python side, and without a worker, I can't easily pass around dictionaries to JS functions that are expecting just plain JS object literals, produced by that `PyProxy.toJs(pythonThing)`.

Is there any plan or chance that the `PyProxy`, or better its static `toJs()` utility, can be used also within the Python code? 🤔

Thank you!

dpgeorge:
@WebReflection I have added formatting and linting using Biome (with indenting set to 4 spaces... at least that matches the Python and C code, and PyScript's style!).

But Biome lint is complaining about your promise with resolvers code:
```
  × The assignment should not be in an expression.
  
     3 │ Promise.withResolvers ||
   > 4 │     (Promise.withResolvers = function withResolvers() {
       │      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   > 5 │         let a;
        ...
  > 11 │         return { resolve: a, reject: b, promise: c };
  > 12 │     });
       │     ^
    13 │ 
    14 │ const mp = await (await import(process.argv[2])).loadMicroPython();
  
  i The use of assignments in expressions is confusing.
    Expressions are often considered as side-effect free.
```
Do you know a good way to fix that?  (I already converted the `var` into `let` to make Biome happier.)

dpgeorge:
> Is there any plan or chance that the `PyProxy`, or better its static `toJs()` utility, can be used also within the Python code?

I have now implemented `jsffi.to_js()` on the Python side.  Hopefully it has the correct semantics for you.

As part of this I added a new `jsffi` module and moved things out of the existing `pyodide` module, and then removed the `pyodide` module.  I don't think it makes sense for MicroPython to try and implement the `pyodide` module because our API is different in semantics.  So I've put the relevant functionality in `jsffi`:
- `jsffi.create_proxy()`
- `jsffi.to_js()`
- `jsffi.JsProxy` -- a type

Hopefully that change (removal of `pyodide` module) won't impact pyscript too much :grimacing:

dpgeorge:
@WebReflection regarding your test above with `in` and `ownKeys()` applied (on the JS side) to a Python dictionary: I guess I must have misunderstood the semantics of `has` and `ownKeys` (as JS concepts).

In Python you have:
- `item in collection` which is equivalent to `collection.__contains__(item)`
- `hasattr(obj, method)

They are quite different although can be confused with each other.  If the subject object is a dict, then `in`/`__contains__` checks if the dict contains within it the given key.  But `hasattr` will check if the dict responds to the given method (also it will check for attributes, but dict's don't have any attributes).

In the webassembly port at the moment, I implemented the latter, ie `hasattr` semantics.  So that's why you are getting a list of methods rather than a list of the keys/elements of the dict.

Should I instead implement `in`/`__contains__`?

WebReflection:
@dpgeorge the `in` operator in JS does something completely different from Python: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in

`'key' in {key: false}` is **True** in JS, it's about knowing if the object or anything down its inheritance / prototypal chain has a *key* with a name `'key'`, it's not about the value ... ever ... even with arrays, it's `0 in []` is **false** but `0 in [0]` is **true**.

Reflect.ownKeys returns keys that belong to the current object, not keys inherited via class or anything else ... this is still about keys.

I think `hasattr` in that regard makes sense, or better, it is what `in` operator expects so thumb up here, thank you!

WebReflection:
@dpgeorge sorry for late reply here too ...

> Do you know a good way to fix that? (I already converted the var into let to make Biome happier.)

you *can't* convert my code to a `var` expression or it breaks as *var* is **hoisted** (available in the whole scope out of the box) while `let` is not, so my patch won't work and it will fail hard on browsers that don't have `Promise.withResolvers` ... my code is rarely accidental 😉 ... you need to revert that **and** you need to put an *ignore comment* for your linter right there and never care about that code ever again in your future.

WebReflection:
@dpgeorge sorry for the further ping but maybe interesting ... the reason there was `var` there is this one:

```js
// this works as expected, ok and error are hoisted
var promise = new Promise((...args) => { [resolve, reject] = args }), resolve, reject;

// the same would deadly fail with `let`
let promise = new Promise((...args) => { [resolve, reject] = args }), resolve, reject;

Uncaught (in promise) ReferenceError: Cannot access 'resolve' before initialization
    at <anonymous>:1:43
    at new Promise (<anonymous>)
    at <anonymous>:1:15
```

That is why the code was what it is but now I think I'll change it and call it a day because if I just pre-declare before the promise everything would be fine.

dpgeorge:
> I think `hasattr` in that regard makes sense, or better, it is what `in` operator expects so thumb up here, thank you!

OK, I will leave the `has`/`ownKeys` implementation as it currently is.

> This code should work without linting complains ... please let me know if it doesn't.

The linter still complains: it wants the `let a, b` split into two `let` statements, and an arrow function used instead for the very inner function.  This is what it wants:
```js
Promise.withResolversCustom = function withResolversCustom() { 
    let a;
    let b;
    const c = new this((resolve, reject) => { 
        a = resolve;
        b = reject;
    });
    return { resolve: a, reject: b, promise: c };
};
```
And I renamed it to `withResolversCustom` so that I can be sure the test that follows actually uses this implementation.

WebReflection:
I think you can configure the linter to allow that and I don’t find empty grouped declarations particularly confusing but then again, linter is for consistent code so multiple `let` it is (and that’s fine). I will do a second round of tests today and let you know how that goes.

ntoll:
Yes * 1000 to moving to a proper namespace (and PyScript will shim over the top). It's great to see this taking shape. :rocket:

WebReflection:
@dpgeorge I can confirm the jsffi does what it needs to do on both main and workers (whre it's implicit, but that's another story):

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>MicroPython - Latest</title>
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@pyscript/core/dist/core.css">
    <script type="module" src="https://cdn.jsdelivr.net/npm/@pyscript/core/dist/core.js"></script>
</head>
<body>
    <script>
        function ownKeys(py_or_js) {
            console.log("a" in py_or_js, py_or_js.length);
            console.log(Reflect.ownKeys(py_or_js));
        }
    </script>
    <mpy-config>
        interpreter = "http://localhost:8080/micropython.mjs"
    </mpy-config>
    <script type="mpy">
        from pyscript import window
        from jsffi import to_js

        import sys
        print(sys.version)

        window.ownKeys(to_js({ "a": 1 }))
    </script>
</body>
</html>
```

I've built this PR and moved it into the same folder where this `index.html` lands ... then I've used `npx mini-coi .` to bootstrap a server that provides all headers needed to have workers available too.

**My only concern** is that there is still a *leaky* `length` attribute that should not be there and its value is `0`, no matter if there is no or one or more fields in the dictionary.

Do you want me to double-check why that leaky property happens and/or when and/or why?

WebReflection:
@dpgeorge I am not sure but it looks to me that `proxy_call_python` should *not* add the `length` property with value `0` when no arguments (which also I am not sure where these come from) are present ... but if you think you can give this module a try and return `new LiteralMap([['a', 1]])` with the given `index.html` case we should all be good and the data type or the resulted type should work cross interpreter with ease. You never want to add a `length`, an object literal with a `length` is not common **except** for the `arguments` reference itself.

```js
const length = (...args) => Reflect.ownKeys(args);
console.log(length());
// ['length']

function args() {
  'use strict';
  return Reflect.ownKeys(arguments);
}
console.log(args());
// ['length', 'callee', Symbol(Symbol.iterator)]
```

dpgeorge:
I have fixed the leaky `length`.  This was a very simple mistake on my behalf, creating a JS `[]` instead of a `{}` as the basis for the converted Python `dict`.  So it was returning an array instead of an object.


> if you think you can give this module a try and return `new LiteralMap([['a', 1]])` with the given `index.html` case we should all be good

I'm happy to incorporate this, it should be easy to do.

Can I just cut-and-paste verbatim your `LiteralMap` code into the code for this PR?

WebReflection:
@dpgeorge 

> Can I just cut-and-paste verbatim your LiteralMap code into the code for this PR?

if you want to cut and paste there is another version in here with minor tweaks I will port to current module:
https://github.com/pyodide/pyodide/blob/6cd25019d5f091e9c3d396c57cc01d145adfa869/src/js/literal-map.js

I am not super sure why thse WASM targeting projects seem to have an issue with just importing or requiring modules but I don't mind any copy/paste, it's just harder to eventually fix the same thing everywhere in case any fix is needed.

dpgeorge:
> am not super sure why thse WASM targeting projects seem to have an issue with just importing or requiring modules

We could add it as a submodule, but it's a tiny piece of code for a submodule, and submodules are not the most fun to deal with.  Usually for small things like that we copy it in verbatim to `lib/<third-party-repo-name>`.

WebReflection:
> We could add it as a submodule

I meant as *npm* module but I guess the toolchain here is quite different so ... fair enough. If you want to go out as this is it's fine to me as we need to align over some gotchas around *LiteralMap*, notably the fact its *Map* methods can be, and should be allowed to be, overridden as the *object literal* shape is preferred over the *literal as Map* edge case ... if we could have both consistently across interpreters then it's cool, if we cannot or some interpreter still wants to impose *Map* first, I prefer this interpreter to just stick with *object literal* instead, as it's doing now.

> This was a very simple mistake on my behalf, creating a JS [] instead of a {}

Please keep in mind `Array.isArray` in *JS* should return *true* for lists / tuples but *false* for object literals when the proxy passes over it ... `Array.isArray(new Proxy([], handler))` is `true`, as example, while `new Proxy({}, handler)` would return `false` and this is also crucial for third party libraries. This beside the *LiteralMap* suggestion.

dpgeorge:
> If you want to go out as this is it's fine to me as we need to align over some gotchas around _LiteralMap_,

OK, for now I will leave `LiteralMap` out and just keep using `{}` as the base for conversion from a Python dict to JS entity.

> Please keep in mind `Array.isArray` in _JS_ should return _true_ for lists / tuples but _false_ for object literals when the proxy passes over it

I have now added a test for this behaviour.

ntoll:
There's not much I can add from a technical perspective (but I am following along). But I just wanted to say how great it is to see this working moving forward and acknowledge, affirm and celebrate that you're collaborating so well and bringing respective areas of expertise together. Bravo..! :rocket: 

(As you can probably tell, I feel like the cheerleader on the edge of the game shouting "whoop whoop"... not really anything to do with what's happening on the pitch, but hopefully you feel it's good to be appreciated! :smile: )

WebReflection:
@dpgeorge

> OK, for now I will leave LiteralMap out and just keep using {} as the base for conversion from a Python dict to JS entity.

the idea is that the day we move to LiteralMap, beside returning just that as proxy for the literal, there won't be any other change to consider from MicroPython side, if its default is already *object-literal* so I am OK with this resolution for now.

> I have now added a test for this behaviour.

Thank you! the last Proxy to consider in terms of native JS APIs drilling proxies (where they shouldn't but that's late to debate) is the function (lambda and others) but I think you got that covered already ... you can pass along `new Proxy(callable, handler)` and only then you have enabled `construct` and `apply` traps.

I am probably stating the obvious there but "*that's all folks*" about Proxy invariants in JS.

P.S. I wrote more here but it looks to me you are already covering all cases https://gist.github.com/WebReflection/58ce94a0d2c4118a4f3f26934c9582f4

dpgeorge:
@WebReflection I have implemented proxying of Python dicts so they behave on the JavaScript side as objects, eg `obj = {'a':1}` on the Python side works as `obj.a` on the JS side.  This actually turned out quite neat, and I also retained the ability to do  `interpreter.globals.get("...")`.

dpgeorge:
I have published the latest version of this PR (which includes REPL support, and the above-mentioned dict/object behaviour) at npm: https://www.npmjs.com/package/@micropython/micropython-webassembly-pyscript/v/1.22.0-224

@ntoll @WebReflection FYI

WebReflection:
> I also retained the ability to do  interpreter.globals.get("...").

this looks like the best form all worlds, as I was hoping *LiteralMap* idea would've brought in Pyodide too ... if only they'd follow your solution here we'd have way happier users with way less surprises! Thank You!

WebReflection:
@dpgeorge I've just updated polyscript and I've noticed one improvement and some "*regression*" in there ...

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <script type="module" src="../../dist/index.js"></script>
  <script>
    function keys(ref) {
      console.log('a' in ref, ref.a);
      console.log(Object.keys(ref));
      // crickets
      for (const k in ref)
        console.log(k, ref[k]);
    }
  </script>
  <script type="micropython">
    import js

    py = {"a": 1}

    js.keys(py)
  </script>
</head>
</html>
```

The `Object.keys(ref)` produces an empty array ... this is the `ownKeys` trap involved that should return the list of dictionary own keys (not its methods from the Dict class) ... it returns nothing.

The `in` trap seems to work but then again, because `ownKeys` trap is not functional, a `for/in` loop returns nothing.

It is also possible that `getOwnPropertyDescriptor` trap should be implemented and return `{enumerable: true, configurable: true, writable: true, value: ....}` otherwise more *JS* code won't work.

The explicit access though works, but I think the Proxy is still incomplete on your side.

If you need any help from me, I'd be happy ... also now I gonna check latest Proxy status and hint where/what should change.

WebReflection:
@dpgeorge maybe this helps more/better:

```js
const dictionaryHandler = {
  deleteProperty(ref, field) {
    // perform anything you need to remove `field`
    // from the reference and then return `true`
    // if that was successful, `false` if you want to throw.
    // keep in mind in JS `delete ({}).any` does not throw.
  },
  get(ref, field, refProxy) {
    // return the `ref[field]` value if `hasattr(ref, field)`
    // is true. Feel free to return the dictionary native method
    // or field if that's not the case. For functions, remember to
    // bind the context which can be either the `ref` or the `refProxy`
    // it's up to you how you handle methods on dictionaries.
    // if the `constructor` is the `field` value remember in JS
    // that's expected to be just the `Object` global field,
    // unless you want to provide a special `PyDict` instead,
    // which I think would be fine.
  },
  getOwnPropertyDescriptor(ref, field) {
    // you need to return something like this:
    // {
    //   value: ref[field],
    //   enumerable: true,
    //   writable: true,
    //   configurable: true
    // }
    // This is for loops and enumerability within loops,
    // it allows `for/in` loops to work as expected or it
    // allows `Object.keys` and `Object.assign` or even
    // `{...spread}` operation to provide the enumerable fields
    // and value to copy over. THIS IS IMPORTANT!
  },
  has(ref, field) {
    // this is to satisfy the `"key" in ref` condition.
    // it is not enough to satisfy `for/in` loops but also
    // it looks like you got this right as the result is `true`
    // when I do check this.
    // The `has` trap is meant to also expose inherited methods
    // or values so it should be `true` also for `get` or anything
    // else that the Python dictionary has in its inheritance.
    // Maybe for JS this is not super important, but in theory
    // both `"get" in {get() {}}` and `"get" in {}`, where `{}`
    // is a Python dictionary, should return `true` here.
  },
  ownKeys(ref) {
    // this should return a list of *own* properties!
    // in a Python dictionary case that's `["a"]` if the
    // dictionary was created as `{"a": 1}`.
    // no inherited methods ever should be exposed in here
    // or shenanigans will happen on the JS introspection side.
    // This should satisfy both `Reflect.ownKeys(pyDictionary)`
    // and `Object.keys(pyDictionary)` returning only what has
    // been explicitly defined in the dictionary and nothing that
    // comes from the `__class__`
  },
  set(ref, field, newValue, refProxy) {
    // this should update the Python dictionary
    // with ref[field] = newValue and return `true`
    // unless that `field` was forbidden or something.
    // I don't think you'll have to deal ever with forbidden
    // fields from the JS side, as those would be always
    // "own properties" attached *but* you can throw if
    // the field is `__proto__` as that won't work in JS,
    // although this is also very unlikely to happen and
    // absolutely an edge case we should not care about right now
  },
};
```

If you want to be defensive there to preserve your Python dictionary you can also add these:

```js
{
  defineProperty(ref, field, descriptor) {
    // triggered when `Object.defineProperty(proxy, field, descriptor)`
    // operation is performed on the JS side of affair.
    // You can throw errors with accessors (the descriptor has `get` or `set`)
    // or you can handle all cases but also the `enumerable` and `configurable`
    // makes this way trickier than it should be to deal with for our use cases.
    // I think you might just ignore this or simply throw if ever encountered.
  },
  getPrototypeOf(ref) {
    // for Python dictionaries consumed in the JS world
    // I think it's fair to assume this will always simply
    // return `Object.prototype` and, unless you think otherwise,
    // you don't really need to provide this trap as that's the default
  },
  isExtensible(ref) {
    // I don't know if Python has immutable dictionaries but
    // if that's the case this should return `false`
    // You can also return `false` for special cases, as example
    // the `interpreter.globals` as you probably don't want JS
    // to mess up with that and keep your globals handled only
    // by your interpreter runtime trusted code
  },
  preventExtensions(ref) {
    // similarly to `isExtensible(ref)` you probably should not
    // care about this trap but if you think some object should
    // return `false`, meaning it can't be extended anymore in JS
    // feel free to play around, although I think this should be
    // `false` by default for Python dictionaries handled in the JS
    // world, as it makes little sense to prevent changes on these,
    // IMHO
  },
  setPrototypeOf(ref, newPrototype) {
    // this should probably always return `false` for Python
    // dictionaries as it makes no sense to change their nature
    // out there in the JS field ... it should rather throw *and*
    // this should also cover the previously mentioned `__proto__`
    // field definition, as that's an ugly legacy hack for prototype
    // that IIRC this trap should prevent anyway ... just stick
    // a `return false;` inside this trap and call it a day, IMHO
  },
}
```

At this point the whole JS Proxy API should be hopefully clearer and fully covered for you and I am sure you can provide the best of all worlds when it comes to JS consuming Python dictionaries 🥳

dpgeorge:
Thanks for the detailed hints @WebReflection .  To get iteration and `Object.keys()` working I needed to implement `getOwnPropertyDescriptor()`.  That's now done and your test above now works.

WebReflection:
@dpgeorge

> That's now done and your test above now works.

do you mind publishing it to *npm* otherwise I can't grab latest automatically from there and current state is not ideal? Thanks!

dpgeorge:
@WebReflection latest version is here: https://www.npmjs.com/package/@micropython/micropython-webassembly-pyscript/v/1.22.0-225

PR #13584, opened 2024-02-02
extmod: Add new `vfs` module with mount/umount and Vfs classes.

dpgeorge:
This PR adds a new `vfs` module.  The following functions and classes have been placed in this new module:
- `vfs.mount()`
- `vfs.umount()`
- `vfs.VfsFat`
- `vfs.VfsLfs1`
- `vfs.VfsLfs2`
- `vfs.VfsPosix`

These are from the `os` module.  They are still available in the `os` module for now, but the intention is to deprecate the `os.mount` etc use and favour `vfs.mount` instead.  Then eventually these functions/classes will be removed from the `os` module.  This is to make the `os` module (eventually) strictly a subset of the CPython `os` module.

Docs, test and ports have all been updated to use `vfs` instead of `os` in the appropriate places.

TODO:
- [x] Decide whether/when to update `pyboard.py` and `mpremote` to use `vfs` instead of `os`.  That will make these tools not work with older versions of MicroPython firmware (one will need to use older versions of `mpremote`, which is easy to do with `pip`).

### DIFF
--- a/docs/esp32/quickref.rst
+++ b/docs/esp32/quickref.rst
@@ -650,15 +650,15 @@ SD card
 
 See :ref:`machine.SDCard <machine.SDCard>`. ::
 
-    import machine, os
+    import machine, os, vfs
 
     # Slot 2 uses pins sck=18, cs=5, miso=19, mosi=23
     sd = machine.SDCard(slot=2)
-    os.mount(sd, '/sd')  # mount
+    vfs.mount(sd, '/sd') # mount
 
     os.listdir('/sd')    # list directory contents
 
-    os.umount('/sd')     # eject
+    vfs.umount('/sd')    # eject
 
 RMT
 ---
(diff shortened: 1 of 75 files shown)


jimmo:
LGTM!

On the `mpremote` / `pyboard.py` question I think we are eventually going to need some sort of board-autodetect functionality. I think it would be nice if current versions did keep working on older boards. In this case it's relatively easy to detect this (either explicitly before the first command or as a fallback if the command fails). It would be also nice to use this to support e.g. CircuitPython.

For now perhaps we could just change them to do `import os as vfs` ?

dpgeorge:
> For now perhaps we could just change them to do `import os as vfs` ?

pyboard/mpremote both need `os` and `vfs` (they need `os.chdir`).

Since `os.mount` etc aren't going anywhere until MicroPython 2.0, maybe we can just leave pyboard/mpremote as they are?  I guess we would delete `pyboard.py` as part of MP 2.0, so that never needs to know about `vfs` module.  And `mpremote` refactor is WIP, so eventually can do auto-detection, and for now continue to use just `os`.

PR #13620, opened 2024-02-08
tools/manifestfile.py: Change library search to use a list of paths.

dpgeorge:
This PR changes how library packages are searched for when a manifest file is loaded: there is now simply a list of library paths that is searched in order for the given package.  This list defaults to the main directories in micropython-lib, but can be added to -- either appended or prepended -- by using `add_library()`.
    
In particular the way unix-ffi library packages are searched has changed, because the `unix_ffi` argument to `require()` is now removed.  Instead, if a build wants to include packages from micropython-lib/unix-ffi, then it must explicitly add this to the list of paths to search using:
    
        add_library("unix-ffi", "$(MPY_LIB_DIR)/unix-ffi")
    
Work done in collaboration with Jim Mussared.

### DIFF
--- a/docs/reference/manifest.rst
+++ b/docs/reference/manifest.rst
@@ -95,6 +95,17 @@ Note: The ``opt`` keyword argument can be set on the various functions, this con
 the optimisation level used by the cross-compiler.
 See :func:`micropython.opt_level`.
 
+.. function:: add_library(library, library_path, prepend=False)
+
+    Register the path to an external named *library*.
+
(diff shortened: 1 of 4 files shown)


dpgeorge:
Tested by adding `require("glob")` to the unix manifest.py.  It now works (previously did not).

Should fix https://github.com/micropython/micropython-lib/pull/672 and https://github.com/micropython/micropython-lib/issues/795

PR #13630, opened 2024-02-09
stm32/ethernet: A few improvements.

robert-hh:
- Set the output speed of RMII_TXD0 and RMII_TXD1 to medium. That matches better the frequency of the signals that may occur at these pins. For configurations with a high capacitive load at these pins medium driving speed may still be too low.
- Allow to set the address of the PHY in the LAN object constructor with the keyword argument  `phy_addr=x`, e.g. `lan = network.LAN(phy_addr=1`. The default value is 0, matching the current behavior.
- Remove some code duplication.
- Allow setting the type of the PHY with the keyword option `phy_type=<model>`. Supported model names are PHY_LAN8742, PHY_LAN8720 and PHY_DP83848. The default is PHY_LAN8742. Tested with a LAN8720 (=LAN8742) and DP83848 device.

### DIFF
--- a/ports/stm32/Makefile
+++ b/ports/stm32/Makefile
@@ -251,6 +251,7 @@ SRC_C += \
 	pyb_can.c \
 	usb.c \
 	eth.c \
+	eth_phy.c \
 	gccollect.c \
 	help.c \
 	machine_bitstream.c \
--- a/ports/stm32/eth.c
+++ b/ports/stm32/eth.c
@@ -40,8 +40,7 @@
 #include "lwip/dhcp.h"
 #include "netif/ethernet.h"
 
-// ETH PHY register definitions (for LAN8742)
-
+// ETH PHY register definitions (for LAN8742 and LAN8720/LAN8710)
 #undef PHY_BCR
 #define PHY_BCR                 (0x0000)
 #define PHY_BCR_SOFT_RESET      (0x8000)
@@ -137,6 +136,7 @@ typedef struct _eth_t {
     uint32_t trace_flags;
     struct netif netif;
     struct dhcp dhcp_struct;
+    uint32_t phy_addr;
 } eth_t;
 
 static eth_dma_t eth_dma __attribute__((aligned(16384)));
@@ -146,11 +146,13 @@ eth_t eth_instance;
 static void eth_mac_deinit(eth_t *self);
 static void eth_process_frame(eth_t *self, size_t len, const uint8_t *buf);
 
-static void eth_phy_write(uint32_t reg, uint32_t val) {
+static void eth_phy_write(uint32_t phy_addr, uint32_t reg, uint32_t val) {
     #if defined(STM32H5) || defined(STM32H7)
     while (ETH->MACMDIOAR & ETH_MACMDIOAR_MB) {
     }
     uint32_t ar = ETH->MACMDIOAR;
+    ar &= ~ETH_MACMDIOAR_PA_Msk;
+    ar |= (phy_addr << ETH_MACMDIOAR_PA_Pos);
     ar &= ~ETH_MACMDIOAR_RDA_Msk;
     ar |= reg << ETH_MACMDIOAR_RDA_Pos;
     ar &= ~ETH_MACMDIOAR_MOC_Msk;
@@ -165,18 +167,20 @@ static void eth_phy_write(uint32_t reg, uint32_t val) {
     }
     ETH->MACMIIDR = val;
     uint32_t ar = ETH->MACMIIAR;
-    ar = reg << ETH_MACMIIAR_MR_Pos | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MW | ETH_MACMIIAR_MB;
+    ar = (phy_addr << ETH_MACMIIAR_PA_Pos) | (reg << ETH_MACMIIAR_MR_Pos) | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MW | ETH_MACMIIAR_MB;
     ETH->MACMIIAR = ar;
     while (ETH->MACMIIAR & ETH_MACMIIAR_MB) {
     }
     #endif
 }
 
-static uint32_t eth_phy_read(uint32_t reg) {
+static uint32_t eth_phy_read(uint32_t phy_addr, uint32_t reg) {
     #if defined(STM32H5) || defined(STM32H7)
     while (ETH->MACMDIOAR & ETH_MACMDIOAR_MB) {
     }
     uint32_t ar = ETH->MACMDIOAR;
+    ar &= ~ETH_MACMDIOAR_PA_Msk;
+    ar |= (phy_addr << ETH_MACMDIOAR_PA_Pos);
     ar &= ~ETH_MACMDIOAR_RDA_Msk;
     ar |= reg << ETH_MACMDIOAR_RDA_Pos;
     ar &= ~ETH_MACMDIOAR_MOC_Msk;
@@ -190,7 +194,7 @@ static uint32_t eth_phy_read(uint32_t reg) {
     while (ETH->MACMIIAR & ETH_MACMIIAR_MB) {
     }
     uint32_t ar = ETH->MACMIIAR;
-    ar = reg << ETH_MACMIIAR_MR_Pos | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MB;
+    ar = (phy_addr << ETH_MACMIIAR_PA_Pos) | (reg << ETH_MACMIIAR_MR_Pos) | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MB;
     ETH->MACMIIAR = ar;
     while (ETH->MACMIIAR & ETH_MACMIIAR_MB) {
     }
@@ -198,9 +202,10 @@ static uint32_t eth_phy_read(uint32_t reg) {
     #endif
 }
 
-void eth_init(eth_t *self, int mac_idx) {
+void eth_init(eth_t *self, int mac_idx, uint32_t phy_addr) {
     mp_hal_get_mac(mac_idx, &self->netif.hwaddr[0]);
     self->netif.hwaddr_len = 6;
+    self->phy_addr = phy_addr;
 
     // Configure GPIO
     mp_hal_pin_config_alt_static(MICROPY_HW_ETH_MDC, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, STATIC_AF_ETH_MDC);
@@ -362,7 +367,7 @@ static int eth_mac_init(eth_t *self) {
     #endif
 
     // Reset the PHY
-    eth_phy_write(PHY_BCR, PHY_BCR_SOFT_RESET);
+    eth_phy_write(self->phy_addr, PHY_BCR, PHY_BCR_SOFT_RESET);
     mp_hal_delay_ms(50);
 
     // Wait for the PHY link to be established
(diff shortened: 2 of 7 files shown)


ports/stm32/eth.c
```diff
@@ -210,8 +210,8 @@ void eth_init(eth_t *self, int mac_idx) {
     mp_hal_pin_config_alt_static(MICROPY_HW_ETH_RMII_RXD0, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, STATIC_AF_ETH_RMII_RXD0);
     mp_hal_pin_config_alt_static(MICROPY_HW_ETH_RMII_RXD1, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, STATIC_AF_ETH_RMII_RXD1);
     mp_hal_pin_config_alt_static(MICROPY_HW_ETH_RMII_TX_EN, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, STATIC_AF_ETH_RMII_TX_EN);
-    mp_hal_pin_config_alt_static(MICROPY_HW_ETH_RMII_TXD0, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, STATIC_AF_ETH_RMII_TXD0);
-    mp_hal_pin_config_alt_static(MICROPY_HW_ETH_RMII_TXD1, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, STATIC_AF_ETH_RMII_TXD1);
+    mp_hal_pin_config_alt_static_speed(MICROPY_HW_ETH_RMII_TXD0, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, MP_HAL_PIN_SPEED_MEDIUM, STATIC_AF_ETH_RMII_TXD0);
+    mp_hal_pin_config_alt_static_speed(MICROPY_HW_ETH_RMII_TXD1, MP_HAL_PIN_MODE_ALT, MP_HAL_PIN_PULL_NONE, MP_HAL_PIN_SPEED_MEDIUM, STATIC_AF_ETH_RMII_TXD1);
```
dpgeorge, line 209: The default when using just `mp_hal_pin_config_alt_static()` is to use "high speed", so this reduces the speed.  Why not configure it to "very high"?

ports/stm32/eth.c
```diff
@@ -165,14 +167,14 @@ STATIC void eth_phy_write(uint32_t reg, uint32_t val) {
     }
     ETH->MACMIIDR = val;
     uint32_t ar = ETH->MACMIIAR;
-    ar = reg << ETH_MACMIIAR_MR_Pos | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MW | ETH_MACMIIAR_MB;
+    ar = reg << ETH_MACMIIAR_MR_Pos | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MW | ETH_MACMIIAR_MB | (phy_addr << ETH_MACMIIAR_PA_Pos);
```
dpgeorge, line 153: please put the pyh_addr bit at the start of this expression, to retain the ordering that most-significant-bits are on the left

ports/stm32/eth.c
```diff
@@ -190,17 +194,18 @@ STATIC uint32_t eth_phy_read(uint32_t reg) {
     while (ETH->MACMIIAR & ETH_MACMIIAR_MB) {
     }
     uint32_t ar = ETH->MACMIIAR;
-    ar = reg << ETH_MACMIIAR_MR_Pos | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MB;
+    ar = reg << ETH_MACMIIAR_MR_Pos | (ar & ETH_MACMIIAR_CR_Msk) | ETH_MACMIIAR_MB | (phy_addr << ETH_MACMIIAR_PA_Pos);
```
dpgeorge, line 180: as above, please reorder

ports/stm32/eth.c
```diff
@@ -190,17 +194,18 @@ STATIC uint32_t eth_phy_read(uint32_t reg) {
...
     ETH->MACMIIAR = ar;
     while (ETH->MACMIIAR & ETH_MACMIIAR_MB) {
     }
     return ETH->MACMIIDR;
     #endif
 }
 
-void eth_init(eth_t *self, int mac_idx) {
+void eth_init(eth_t *self, int mac_idx, int phy_addr) {
```
dpgeorge: please make the type of phy_addr a `uint32_t` to match elsewhere

ports/stm32/Makefile
```diff
@@ -251,6 +251,7 @@ SRC_C += \
 	pyb_can.c \
 	usb.c \
 	eth.c \
+	phy.c \
```
dpgeorge, line 254: I suggest calling this new file `eth_phy.c`.

ports/stm32/eth.c
```diff
@@ -26,12 +26,14 @@
 
 #include <string.h>
 #include "py/mphal.h"
+#include "py/runtime.h"
```
dpgeorge, line 29: Why is this header needed?

ports/stm32/eth.c
```diff
@@ -146,7 +129,7 @@ eth_t eth_instance;
 STATIC void eth_mac_deinit(eth_t *self);
 STATIC void eth_process_frame(eth_t *self, size_t len, const uint8_t *buf);
 
-STATIC void eth_phy_write(uint32_t phy_addr, uint32_t reg, uint32_t val) {
+void eth_phy_write(uint32_t phy_addr, uint32_t reg, uint32_t val) {
```
dpgeorge, line 132: Does it make sense to move this function and `eth_phy_read()` to the new `eth_phy.c` file?

ports/stm32/eth.h
```diff
@@ -29,12 +29,19 @@
...
-void eth_init(eth_t *self, int mac_idx, int phy_addr);
+void eth_init(eth_t *self, int mac_idx, int phy_addr, int phy_type);
 void eth_set_trace(eth_t *self, uint32_t value);
 struct netif *eth_netif(eth_t *self);
 int eth_link_status(eth_t *self);
 int eth_start(eth_t *self);
 int eth_stop(eth_t *self);
 void eth_low_power_mode(eth_t *self, bool enable);
 
+enum {
```
dpgeorge, line 40: Please put this enum at the top of this file.

ports/stm32/eth.h
```diff
@@ -29,12 +29,19 @@
...
+    PHY_LAN8742 = 0,
```
dpgeorge, line 41: I suggest prefixing these names with ETH_, eg `ETH_PHY_LAN8742`.

ports/stm32/phy.h
```diff
@@ -0,0 +1,67 @@
...
+#define PHY_SPEED_10HALF   (1)
+#define PHY_SPEED_10FULL   (5)
+#define PHY_SPEED_100HALF  (2)
+#define PHY_SPEED_100FULL  (6)
+#define PHY_DUPLEX         (4)
+
+uint32_t eth_phy_read(uint32_t phy_addr, uint32_t reg);
+void eth_phy_write(uint32_t phy_addr, uint32_t reg, uint32_t val);
+
+int16_t phy_lan87xx_get_link_status(uint32_t phy_addr);
```
dpgeorge, line 62: I suggest prefixing these with `eth_`.

ports/stm32/eth.c
```diff
@@ -381,6 +371,14 @@ STATIC int eth_mac_init(eth_t *self) {
                 break;
             case 1:
                 if (bsr & PHY_BSR_LINK_STATUS) {
+                    // Announce all modes
+                    eth_phy_write(self->phy_addr, PHY_ANAR,
+                        PHY_ANAR_SPEED_10HALF |
+                        PHY_ANAR_SPEED_10FULL |
+                        PHY_ANAR_SPEED_100HALF |
+                        PHY_ANAR_SPEED_100FULL |
+                        PHY_ANAR_IEEE802_3);
```
dpgeorge, line 380: This new setting is a change from the previous behaviour, which just left this register in its default state.  How does the change you've made impact things?  (It looks like a good change, I'm just wondering why you made it.)

ports/stm32/eth.c
robert-hh: Actually I was not sure what state it is and changed it to the slowest speed that fits according to the data sheet and that worked with my "ball of wires" test set-up. But if the default is "high speed", it can be left as that. "Very high" speed is according to the data sheet not needed at 3.3V and it just increases the radiation and cross-talk. So I can drop this commit.

ports/stm32/eth.c
robert-hh: Done, as well for the H7 case.

ports/stm32/Makefile
robert-hh: Done, in contrast to usb_phy.*

ports/stm32/eth.c
robert-hh: For raising an error about a wrong phy_type:
`mp_raise_ValueError(MP_ERROR_TEXT("Invalid phy_type"));`

ports/stm32/eth.c
robert-hh: I considered that forth and back and decided to leave it in eth.c, because:
- eth_phy.c implements only the writes to registers that are different between the PHY devices. All common register accesses are still in eth.c.
- eth_phy_read() and eth_phy_write() access directly the MAC controller, like all the other code in eth.c. They are not specific for a different type of PHY device.

ports/stm32/eth.h
robert-hh: Done.

ports/stm32/phy.h
robert-hh: Done

ports/stm32/eth.c
robert-hh: I made it to be sure that all possible modes that are supported by the Ethernet controller are announced. The default mode can as well be set by strapping pins at the PHY, which by chance/bad luck may be more restrictive.

ports/stm32/eth.h
robert-hh: Done.

robert-hh:
The PR is updated. Tested with a PYBD_SF6 and STM32H750 device and LAN8720 and DP83848 PHY devices. The performance using iperf3 is about 80MBit/s for TCP and 10.5MBit/s for UDP.

dpgeorge:
> 10.5MBit/s for UDP

I think this is just the default bitrate setting for UDP tests.  You should be able to get a lot more bandwidth than this by using the `-b` setting on the client side (assuming you run the client on a PC).  Eg `iperf3 -u -b 80M`.  The thing to look out for with UDP tests is the packet loss.

robert-hh:
At the board version the bandwidth can be set as well. Setting it to 80M results in a data rate of 80M for udp. Using `reverse=True` yields an UDP loss rate of 0 at a 10M bandwidth, and a loss rate of 0.2% at 80M bandwidth. That seems fine.

ports/stm32/eth.c
dpgeorge: I see.

It would be better to separate the layers here: low-level hardware init vs higher-level Python interface.

I suggest changing `eth_init()` to return `int`, with `0` being success and a negative number being an error code.  Then callers of this function must test this error code and raise as appropriate.

As part of this, note that `ARDUINO_PORTENTA_H7`'s board init function calls `eth_init()`, and that should definitely not raise an exception (because that code runs outside the main MicroPython framework).  So that use will need to be updated to pass the correct PHY.

ports/stm32/eth.c
dpgeorge: OK.

ports/stm32/eth.c
robert-hh: Done

ports/stm32/eth.c
robert-hh: Moved the raise() call to network_lan.c and added the proper parameters to board_init.c of the Portenta H7. Thank you for the hint about the Portenta H7. I did not expect that.

robert-hh:
Updated.

dpgeorge:

robert-hh:
Thank you very much. I tested and it works.

PR #13638, opened 2024-02-11
esp32/machine_hw_spi: Common SPI argument parsing.

xyzzy42:
This combines the argument parsing and checking for the machine.SPI.init() and machine.SPI() interfaces.

The only real difference was unspecified arguments in init() mean to keep the same value, while in new they get default values.

Behavior has changed for passing the "id" argument to init().  On other ports this isn't allowed.  But on esp32 it would change the SPI controller of the static SPI instance to the new id.  This results in multiple static spi objects for the same controller and they would fight over which one has inconsistent mpy vs esp-idf state.  This has been changed to not allow "id" with init(), like other ports.

In a few causes, a loop is used over arguments that are handled the same way instead of cut & pasting the same stanza of code for each argument.

The init_internal function had a lot of arguments, which is not efficient to pass.  Pass the args mp_arg_val_t array instead as a single argument. This reduced both the number of C lines and the compiled code size.

Summary of code size change:  Two argument lists of 72 bytes are replaced by a single shared 72 byte list.  New shared argument parsing code is 155 bytes, and using it reduces existing functions by 465 bytes.

```
add/remove: 3/0 grow/shrink: 0/5 up/down: 227/-609 (-382)
Function                                     old     new   delta
machine_hw_spi_argcheck                        -     135    +135
spi_allowed_args                               -      72     +72
defaults$0                                     -      20     +20
allowed_args$1                               424     352     -72
allowed_args$0                              1112    1040     -72
machine_hw_spi_init                          122      36     -86
machine_hw_spi_make_new                      255     118    -137
machine_hw_spi_init_internal                 698     456    -242
Total: Before=1480477, After=1480095, chg -0.03%
```

### DIFF
--- a/ports/esp32/machine_hw_spi.c
+++ b/ports/esp32/machine_hw_spi.c
@@ -80,9 +80,15 @@
 #define MP_HW_SPI_MAX_XFER_BITS (MP_HW_SPI_MAX_XFER_BYTES * 8) // Has to be an even multiple of 8
 
 typedef struct _machine_hw_spi_default_pins_t {
-    int8_t sck;
-    int8_t mosi;
-    int8_t miso;
+    union {
+        int8_t array[3];
+        struct {
+            // Must be in enum's ARG_sck, ARG_mosi, ARG_miso, etc. order
+            int8_t sck;
+            int8_t mosi;
+            int8_t miso;
+        } pins;
+    };
 } machine_hw_spi_default_pins_t;
 
 typedef struct _machine_hw_spi_obj_t {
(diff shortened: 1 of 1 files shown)


ports/esp32/machine_hw_spi.c
```diff
@@ -392,123 +398,87 @@ STATIC void machine_hw_spi_print(const mp_print_t *print, mp_obj_t self_in, mp_p
...
-    } else if (args[ARG_mosi].u_obj == mp_const_none) {
-        mosi = -1;
-    } else {
-        mosi = machine_pin_get_id(args[ARG_mosi].u_obj);
-    }
+STATIC void machine_hw_spi_init(mp_obj_base_t *self_in, size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    machine_hw_spi_obj_t *self = (machine_hw_spi_obj_t *)self_in;
+
+    mp_arg_val_t args[MP_ARRAY_SIZE(spi_allowed_args)];
+    mp_arg_parse_all(n_args, pos_args, kw_args, MP_ARRAY_SIZE(spi_allowed_args),
```
dpgeorge, line 461: It might work (and would be simpler) to disallow the `id` argument here, by doing:
```py
mp_arg_parse_all(n_args, pos_args, kw_args, MP_ARRAY_SIZE(spi_allowed_args) - 1, spi_allowed_args + 1, args + 1);
```
That is, skip the `id` argument when passing the `spi_allowed_args` array.

Then, `machine_hw_spi_argcheck()` won't need to check `args[ARG_id].u_int` in this case.

xyzzy42: Yes, that should work.  And I can avoid the need to put error messages already part of mp_arg_parse_all in the code here too.  I bet that saves a bunch of code.

xyzzy42:
I've added quad SPI initi support to this driver, which added two more pin arguments to both functions.

dpgeorge:
> I've added quad SPI initi support to this driver, which added two more pin arguments to both functions.

Please be sure to do that in a separate commit, after the argument refactoring commit(s).

ports/esp32/machine_hw_spi.c
trentp-igor: And done.  Saved a bunch of code.  I hated copying the error messages out of the library argument parsing code.

After this, there was this block of `if(new) { }` and the common function, and there was really no point to moving that block out of make_new, so it put it back, which saved some more code.

All this only saved 19 more bytes that the previous version.  This is because the refactored arg checking code was small enough gcc decided to inline it into both call sites.  It be a bit more without that.

xyzzy42: And done. Saved a bunch of code. I hated copying the error messages out of the library argument parsing code.

After this, there was this block of if(new) { } and the common function, and there was really no point to moving that block out of make_new, so it put it back, which saved some more code.

All this only saved 19 more bytes that the previous version. This is because the refactored arg checking code was small enough gcc decided to inline it into both call sites. It be a bit more without that.

xyzzy42:
> > I've added quad SPI initi support to this driver, which added two more pin arguments to both functions.
> 
> Please be sure to do that in a separate commit, after the argument refactoring commit(s).

Of course, I didn't include it intentionally.  But this refactor saves even more cut & paste when more pin arguments get added for quad or octal spi.

xyzzy42:
esp32-c3 build error is totally unrelated problem.  I can't re-run the script.

dpgeorge:
I reran the CI and everything passed.

Now merged.

PR #13640, opened 2024-02-11
esp32/network_ppp: Make PPP support optional.

xyzzy42:
PPP is not that commonly used, let it be turned off in the board config to save space.  It is still on by default.

On an basic ESP32-S3 build, turning off PPP with LWIP still on saves ~35 kB of codend 4 kB of data.

 ```
1321257  304296 2941433 4566986  45afca before-ppp-off.elf
1285101  299920 2810305 4395326  43113e after-ppp-off.elf
-------------------------------
-36156    -4376     -56
```

Note that the BSS segment size includes all NOBITS sections in ELF file. Some of these are aligned to 64kB chunk sized dummy blocks, I think for alignment to MMU boundaries, and these went down by 1 block each, so 128 kiB of BSS is not really part of the binary size reduction.


dpgeorge:
Thanks!

PR #11430, opened 2023-05-05
Enable more features in the embed port

cwalther:
In preparation for using the _embed_ port in a project, I am trying to enable more features in it than the minimalistic `examples/embedding` demonstrates. I am making good progress, however for some features I was unable to achieve my goal completely out-of-tree and required patches to the MicroPython source tree itself. This PR is my attempt to get these patches upstream.

I am not done yet with enabling more features (next on the list is a filesystem), hence the draft status, but I still wanted to publish my work in progress for public scrutiny, in case people more familiar with MicroPython internals than me have better ideas on how to achieve my goals. Accordingly, in the short term, I plan to update this branch by force push occasionally, to incorporate new insights or to rebase on current master. Once I consider it done, I will remove the draft status.

PR #13653, opened 2024-02-13
embed: improve stack top estimation

### LINKED ISSUE #11781: ports/embed: segfault in mp_compile()
There is a bug in the `embed` port that can cause segfaults while compiling Python code in memory-constrained situations. I have not managed to construct a simple reproducer yet, since it depends on memory allocation patterns, but the root cause is easy to verify, so I’m reporting it already.

`mp_stack_ctrl_init()`, called two levels deep in `main()`→`mp_embed_init()`→`mp_stack_ctrl_init()`, captures a too low `stack_top`, so that the local variable `parse_tree` of `mp_embed_exec_str()`, called only one level deep in `main()`→`mp_embed_exec_str()`, ends up outside of the stack range looked at by the garbage collector in `gc_helper_collect_regs_and_stack()`. When garbage collection is invoked at just the wrong moment during compilation, the blocks pointed at by the contents of `parse_tree` are therefore freed prematurely, and when they are later reused and overwritten, that corrupts the pointers inside them, which is likely to lead to memory access failures in `mp_parse_tree_clear()` at the latest.

This is easily verified with
```diff
diff --git a/ports/embed/port/embed_util.c b/ports/embed/port/embed_util.c
index 14f508977..816e28a36 100644
--- a/ports/embed/port/embed_util.c
+++ b/ports/embed/port/embed_util.c
@@ -49,6 +49,7 @@ void mp_embed_exec_str(const char *src) {
         mp_lexer_t *lex = mp_lexer_new_from_str_len(MP_QSTR__lt_stdin_gt_, src, strlen(src), 0);
         qstr source_name = lex->source_name;
         mp_parse_tree_t parse_tree = mp_parse(lex, MP_PARSE_FILE_INPUT);
+        assert((char*)&parse_tree <= MP_STATE_THREAD(stack_top));
         mp_obj_t module_fun = mp_compile(&parse_tree, source_name, true);
         mp_call_function_0(module_fun);
         nlr_pop();
```
This assert fails for me on macOS 13 x86_64, compiling with clang 14.

I am not sure how to best fix this. Ideas so far:
* Remove `mp_stack_ctrl_init()` from `mp_embed_init()` and require the application itself to call it, at the same or higher level than it will later call `mp_embed_exec_str()`.
* Have `mp_stack_ctrl_init()` apply some offset to the detected stack top to account for its own call depth. I have no idea how to generically and portably determine what that offset should be, though. (Underestimating will fail to fix the problem, overestimating might run into unmapped memory above the stack.)
* `mp_embed_exec_str()` could check whether `parse_tree` is above `stack_top` and if so adjust the latter.

Any opinions?

yamt:
Fixes: https://github.com/micropython/micropython/issues/11781

### DIFF
--- a/examples/embedding/main.c
+++ b/examples/embedding/main.c
@@ -31,7 +31,14 @@ static char heap[8 * 1024];
 
 int main() {
     // Initialise MicroPython.
-    mp_embed_init(&heap[0], sizeof(heap));
+    //
+    // Note: &stack_top below should be good enough for many cases.
+    // However, depending on environment, there might be more appropriate
+    // ways to get the stack top value.
+    // eg. pthread_get_stackaddr_np, pthread_getattr_np,
+    // __builtin_frame_address/__builtin_stack_address, etc.
+    int stack_top;
+    mp_embed_init(&heap[0], sizeof(heap), &stack_top);
 
     // Run the example scripts (they will be compiled first).
     mp_embed_exec_str(example_1);
--- a/ports/embed/port/embed_util.c
+++ b/ports/embed/port/embed_util.c
@@ -34,8 +34,8 @@
 #include "port/micropython_embed.h"
 
 // Initialise the runtime.
-void mp_embed_init(void *gc_heap, size_t gc_heap_size) {
-    mp_stack_ctrl_init();
+void mp_embed_init(void *gc_heap, size_t gc_heap_size, void *stack_top) {
+    mp_stack_set_top(stack_top);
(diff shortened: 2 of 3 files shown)


dpgeorge:
Thanks for the patch, this looks like a good solution to that issue.  It forces the user to think about the stack top pointer.

cwalther:
Thanks! I will integrate this into #11430 when I get the time.

PR #13656, opened 2024-02-13
mimxrt: Fix headers include guards typos.

iabdalkader:


dpgeorge:
Maybe worth also fixing `cyw43_configport.h` at the same time?

iabdalkader:
Fixed!

PR #4404, opened 2019-01-09
add esp32 smartconfig

Lupino:
#3427 

```python
import smartconfig
import network
wlan = network.WLAN(network.STA_IF)
if not wlan.active():
    wlan.active(True)

smartconfig.set_type(smartconfig.ESPTOUCH)
smartconfig.start()
while smartconfig.status() != smartconfig.SC_STATUS_LINK_OVER:
    pass
smartconfig.stop()

print(smartconfig.get_ssid())
print(smartconfig.get_password())
print(smartconfig.get_phoneip())
```

PR #13658, opened 2024-02-13
esp32/modsmartconfig: Add smartconfig module.

walklinewang:
Here is the test code which shown all methods and constants belong to smartconfig module, I was compiled firmware with `micropython master` (synced yesterday) and `esp-idf v5.0.4`, separately tested on `esp32` and `esp32c3` develop board.

```python
import network
import time
import smartconfig

import esp
esp.osdebug(0)


TYPES = {
    smartconfig.TYPE_ESPTOUCH: 'ESPTOUCH',
    smartconfig.TYPE_AIRKISS: 'AIRKISS',
    smartconfig.TYPE_ESPTOUCH_AIRKISS: 'ESPTOUCH_AIRKISS',
    smartconfig.TYPE_ESPTOUCH_V2: 'ESPTOUCH_V2'
}

TIMEOUT = 120_000 # ms
TESTING_ESPTOUCH_V2 = False

def run_test():
    sta = network.WLAN(network.STA_IF)
    _ = sta.active(True)

    print('smartconfig')
    print(f'  - version: {smartconfig.version()}')

    if TESTING_ESPTOUCH_V2:
        smartconfig.v2_key('1234567890123456')
        print(f'  - v2 key: {smartconfig.v2_key()}')

        smartconfig.type(smartconfig.TYPE_ESPTOUCH_V2)
    else:
        smartconfig.type(smartconfig.TYPE_ESPTOUCH_AIRKISS)
    print(f'  - type: {TYPES[smartconfig.type()]}')

    smartconfig.fast_mode(True)
    print(f'  - fast mode: {smartconfig.fast_mode()}')

    smartconfig.timeout(15)
    print(f'  - timeout: {smartconfig.timeout()}')

    smartconfig.start()

    start_ms = time.ticks_ms()

    while not smartconfig.done():
        if time.ticks_ms() - start_ms > TIMEOUT:
            print('smartconfig timeout')
            smartconfig.stop()
            return

        time.sleep_ms(100)

    print('smartconfig done, ', end='')

    if sta.status() == network.STAT_GOT_IP:
        print('and connected to ap')
        print(f'smartconfig info: {smartconfig.info()}')
        print(f'  - ssid: "{smartconfig.ssid()}"')
        print(f'  - password: "{smartconfig.password()}"')
        print(f'  - bssid: {smartconfig.bssid()}')
        print(f'  - type: {smartconfig.type()} ({TYPES[smartconfig.type()]})')

        if smartconfig.v2_data():
            print(f'  - v2_data: "{smartconfig.v2_data()}"') # EspTouch V2 custom data
    else:
        # maybe wrong password or other situations
        print('but failed connect to ap')

    smartconfig.stop()


if __name__ == '__main__':
    run_test()

```

Update: Tested  `v5.1.2` and `v5.2.2` in addition to `esp-idf v5.0.4`, compiled the firmware using the latest MicroPython code.

### DIFF
--- a/ports/esp32/modsmartconfig.c
+++ b/ports/esp32/modsmartconfig.c
@@ -1,0 +1,248 @@
+#include <string.h>
+
+#include "py/runtime.h"
+
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "freertos/event_groups.h"
+
+#include "esp_wifi.h"
+#include "esp_event.h"
+#include "esp_log.h"
+#include "esp_smartconfig.h"
+
+#include "mphalport.h"
+#include "modsmartconfig.h"
+
+static EventGroupHandle_t wifi_event_group;
+
+static bool smartconfig_process_done = false;
+
+// smartconfig info variables
+static uint8_t ssid[33] = {0};
+static uint8_t bssid[6] = {0};
+static uint8_t password[65] = {0};
+static uint8_t rvd_data[33] = {0};
+static int8_t type = SC_TYPE_ESPTOUCH_AIRKISS;
+
+// event_handler used for esp_event_handler_register()
+static void event_handler(void *arg, esp_event_base_t event_base,
+    int32_t event_id, void *event_data) {
+    if (event_base == WIFI_EVENT &&
+        event_id == WIFI_EVENT_STA_START) {
+        xTaskCreate(smartconfig_task, "smartconfig_task", 4096, NULL, 3, NULL);
+    } else if (event_base == WIFI_EVENT &&
+               event_id == WIFI_EVENT_STA_DISCONNECTED) {
+        // wrong password or other situations may occurs this event
+        xEventGroupSetBits(wifi_event_group, SMARTCONFIG_DONE_BIT);
+    } else if (event_base == IP_EVENT &&
+               event_id == IP_EVENT_STA_GOT_IP) {
+        xEventGroupSetBits(wifi_event_group, CONNECTED_BIT);
+    } else if (event_base == SC_EVENT &&
+               event_id == SC_EVENT_SCAN_DONE) {
+        ESP_LOGI(TAG, "Scan done");
+    } else if (event_base == SC_EVENT &&
+               event_id == SC_EVENT_GOT_SSID_PSWD) {
+        smartconfig_event_got_ssid_pswd_t *evt = (smartconfig_event_got_ssid_pswd_t *)event_data;
+        wifi_config_t wifi_config;
+
+        bzero(&wifi_config, sizeof(wifi_config_t));
+        memcpy(wifi_config.sta.ssid,     evt->ssid,     sizeof(wifi_config.sta.ssid));
+        memcpy(wifi_config.sta.password, evt->password, sizeof(wifi_config.sta.password));
+        wifi_config.sta.bssid_set = evt->bssid_set;
+
+        if (wifi_config.sta.bssid_set == true) {
+            memcpy(wifi_config.sta.bssid, evt->bssid, sizeof(wifi_config.sta.bssid));
+        }
+
+        memcpy(ssid,     evt->ssid,     sizeof(evt->ssid));
+        memcpy(bssid,    evt->bssid,    sizeof(evt->bssid));
+        memcpy(password, evt->password, sizeof(evt->password));
+        type = evt->type;
+
+        if (evt->type == SC_TYPE_ESPTOUCH_V2) {
+            check_esp_err(esp_smartconfig_get_rvd_data(rvd_data, sizeof(rvd_data)));
+        }
+
+        check_esp_err(esp_wifi_disconnect());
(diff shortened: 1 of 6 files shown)


Walkline80:
You can go with `MicroMsg` or [EspTouch App](https://www.espressif.com.cn/en/support/download/apps?keys=&field_technology_tid%5B%5D=20) for smartconfig tests.

Walkline80:
In the present, because of 2.4G-5G mixed WiFi and WiFi6 routers, `AirKiss` protocol is no longer easy to use, even unusable, and using `EspTouch` and `EspTouch V2` protocol requires additional installation of app, I'm afraid that many people don't like to use smartconfig now, but I think there should still be such a function, and it only takes up about 40K of firmware space :)

Walkline80:
Rebase all commits to single one, and append document for smartconfig.

Walkline80:
Hi @projectgus, I didn't keep working on #4404 , because the ESP32 port requires ESP-IDF v5, which is a lot different than before, so I chose to start over.

Walkline80:
According to PR #13763  removed all `STATIC` macro in source files.

ports/esp32/modsmartconfig.c
```diff
@@ -0,0 +1,248 @@
...
+static EventGroupHandle_t wifi_event_group;
+
+static bool smartconfig_process_done = false;
+
+// smartconfig info variables
+static uint8_t ssid[33] = {0};
+static uint8_t bssid[6] = {0};
+static uint8_t password[65] = {0};
+static uint8_t rvd_data[33] = {0};
+static int8_t type = SC_TYPE_ESPTOUCH_AIRKISS;
```
mattytrentini, line 26: Is there a reason why a couple of these arrays (`ssid`, `password`) [have an extra byte](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_smartconfig.html#_CPPv433smartconfig_event_got_ssid_pswd_t)?

mattytrentini:
It looks like a nice feature, though I hadn't come across it before. I presume you've tested it? :) 

I would guess that the maintainers would prefer this to be an optional component (even if it default to on by default).

Walkline80:
@mattytrentini thanks for your review.

> I would guess that the maintainers would prefer this to be an optional component (even if it default to on by default).

About the optional component, here is my thoughts:

1. Keep the modification of [ports/esp32/esp32_common.cmake](https://github.com/micropython/micropython/pull/13658/files#diff-0449b9894deaa2526024c3f5e37b76adcfb012346bd412f9a83ae8b6d74bf6efR90)
2. Add a macro definition to (every) `ports/esp32/boards/XXX/mpconfigboard.h`, looks like:

   ```c
   #define MICROPY_ENABLE_SMARTCONFIG (1)
   ```
3. Finally, surround a preprocessor command to the whole `modsmartconfig.c` file, looks like:

    ```c
    #if MICROPY_ENABLE_SMARTCONFIG
    ...
    ...
    #endif // MICROPY_ENABLE_SMARTCONFIG
    ````

I have tried these steps, ` #define MICROPY_ENABLE_SMARTCONFIG (1)` and ` #define MICROPY_ENABLE_SMARTCONFIG (0)` can make this component optionally appear or disappear. Please let me know if I'm correct, thanks~

> they can read about this feature by looking at the Espressif SmartConfig documentation.

I see what you mean, aligning the function names in the component to the function names in the idf documentation.

But I can't fully agree with you, even if you do align the function names, there're still some differences between them, for example, the IDF API doesn't directly return the information, such as `ssid` and `password`, as well as the status of the smartconfig process, so you still have to refer to the component's documentation.

ports/esp32/modsmartconfig.c
Walkline80: I made the following attempt:

1. Create a new cell phone hotspot

2. Setup `ssid` as long as possible (my phone can set 32 characters, but my colleague's phone can only set 30 characters)

3. Go through the `smartconfig` process, and connected to the hotspot correctly

4. Use `smartconfig.ssid()` (or other functions returned with variable length) to return the obtained ssid

    1. If the length of the hotspot ssid is **less than 32**, then the correct contents will be printed
    2. If the length of the hotspot ssid is **equals to 32**, then it will print more than 32 characters, as for the exact length, I guess it needs to find the next `\0` character?

Walkline80:
The following changes were made in this update:

* Added `version()`, `timeout()`, `fast_mode()` and `v2_key()` functions.
* Renamed function `rvd_data()` to `v2_data()`.
* Change the return value type of `v2_data()` to `str`.
* Make module optional (already selected by default)

ports/esp32/modsmartconfig.c
Walkline80: These attempts presuppose that the array size is set to the same size as the array declared in the document, without an extra byte.

PIBSAS:
Hope this will add to normal Micropython too!

Walkline80:
@PIBSAS thanks for your reply.

Walkline80:
I have no idea why there are 4 more files in the file changed list and the number of checks has increased to 67 and one of them fails....

PR #13660, opened 2024-02-14
esp32/mpnimbleport: Release the GIL while doing NimBLE port deinit.

dpgeorge:
In case callbacks must run (eg a disconnect event happens during the deinit) and the GIL must be obtained to run the callback.

Fixes issue #12349.

TODO:
- [ ] maybe write a multitest for this case (soft reset while connected with IRQ handler active)


dpgeorge:
This only fixes part of issue #12349.

PR #13664, opened 2024-02-14
stm32/boards: Standardize pin names to always have P{PORT}{NUMBER} available

AsensioL:
I noticed that some STM32 Nucleo boards were missing the PA0 notation that is used by machine.Pin to control GPIO pins.

I have added those pins to the pins.csv of the boards that were missing them.

I have not checked that all of the pins are externally exposed, only that they were on the second column (and, of course, removed duplicates).


AsensioL:
Here's the script that I used to search for projects with missing pins:
```bash
cd ports/stm32/boards
for board in $boards; do
    grep -E "^(PA0)*,PA0" $board/pins.csv > /dev/null;
    has_pa0=$?;
    echo $has_pa0 $board | grep NUCLEO;
done
```
Where `1` indicates missing `PA0,PA0` followed by the folder name

robert-hh:
The Pin name scheme is indeed not consistent across the STM32 boards. Some use PAxx, some use Axx, some use both. Taking the Pyboard devices as reference, it should be Axx. But enforcing a common scheme may break existing code. So first a decision must be made about how to proceed.

iabdalkader:
This is a known issue with the stm32 port, CPU pins names should match the datasheet, but in the stm32 port the "P" prefix was removed from to save code space, so boards use a workaround by redefining `Pxy,Pxy` as board pins (ironically using even more code space). The proper fix would be to use the full CPU datasheet names (i.e `Pxy`), and remove all those extra pins, but that may break some existing scripts that use `Px`. Note 27/68 boards redefine `Pxy,Pxy`.

AsensioL:
I could be totally biased by only having used MicroPython on STM32 Nucleo boards, but, here was my thought:

On many Nucleo 64 boards (see [UM1724](https://www.st.com/resource/en/user_manual/dm00105823.pdf), starting page 28, but there's other with different layouts) there is an Arduino header on each side of the board, with pins named A0 to A5 (left) and D0 to D15 (right).
With some exceptions, these pins do not map one-to-one to the CPU's PA0 to PA5 or PD0 to PD15 respectively. Also the Port+PinNumber pins are denoted using Pxy in the same PDF, hinting at two different naming systems:
`{A,D}y` for aliases, and `Pxy` for CPU pins

I had previously only initialize pins using the `machine.Pin('Pxy')` constructor, so when I switched to a board missing the `Pxy` notation I assumed it was missing and proceeded to add them.
Then I thought that they might be missing on other Nucleo boards as well, checked and found out that that was the case.

However, I have also recently learned that I could access the CPU pins from `machine.cpu.xy` and aliased pins from `machine.board.alias`, so this PR might be completely unnecessary.

------

TLDR; This change shouldn't break anyone's code, it just makes sure all STM32 Nucleo boards can have their pins consistently initialized using `machine.Pin('Pxy')`

iabdalkader:
I understand this is a non-breaking change, and I also would like to keep `machine.Pin('Pxy')` working, as we have many scripts that use it, but at this point I'm wondering if this workaround should be done at the pin-generation level (in `make-pins.py` scripts), this way all boards have it by default and we can remove all of the aliases.

PR #13674, opened 2024-02-15
Avoid including system headers when the functionality is disabled

yamt:
Background:
When I ported ports/unix to WASI, i disabled unrelevant functionalities.
eg. MICROPY_PY_TERMIOS=0 because WASI doen't have termios.
however, it still tries to include the corresponding system headers. (eg. termios.h, which WASI doen't provide.)

### DIFF
--- a/ports/unix/modsocket.c
+++ b/ports/unix/modsocket.c
@@ -25,6 +25,10 @@
  * THE SOFTWARE.
  */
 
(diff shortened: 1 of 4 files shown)


dpgeorge:
Thanks for the patch.

I have squashed these 4 commits together.  Generally I do prefer separated commits, but in this case they were all closely related and make sense as a single commit.

PR #11723, opened 2023-06-07
mpy-cross: WASM target and Python wheel.

jimmo:
This PR adds:
 - A way to compile mpy-cross to a self-contained WASM (not WASI) binary. This binary exposes an entry point to take Python source as bytes (as well as the various options), and return the generated .mpy file bytes. This could be used directly from Javascript/Node (will add an example).

 - A way to use this binary from the mpy_cross Python wrapper (currently via the `wasmtime`, `pywasm3` and `wasmer` runtimes). Between these three runtimes, all of the host targets and architectures we need should be covered. The performance in both these runtimes is excellent. Initial tests show very little difference to the native binary. It also works with pywasm (which is a pure-Python WASM runtime), but too slow to be of practical use.

This is an alternative to #10834. (cc @dlech @andrewleech). Instead of having to go through the whole process to generate wheels for every architectures and platforms, we generate a single .wasm binary, and rely on there being a WASM runtime available instead for the target (which there seems to be). We can also ship multiple .wasm binaries in the (single) wheel for all the different bytecode versions we want to support.

In more detail:
* Split mpy-cross Makefile and mpconfigport.h into native and wasm versions (with shared common config).
* Implement the wasm entry point. This disables GC (to avoid needing to worry about register tracking) and NLR (not necessary for compilation).
* Makes the Python wrapper able to use different compiler backends, and implements a native and pywasm3 backend.
* Add argument parsing to the `__main__` entry point (which can now use either backend).
* Add a pyproject.toml to generate the wheel

Future work:
* Auto-select backend based on desired bytecode version.
* javascript example (i.e. browser-based mpy-cross).
* Investigate performance optimisations for pywasm.

_This work was funded through GitHub Sponsors._

PR #13676, opened 2024-02-15
Exprimental WASI support for ports/unix

yamt:
~~See comments in build-wasi.sh~~

### DIFF
--- a/ports/unix/mpconfigport.h
+++ b/ports/unix/mpconfigport.h
@@ -164,7 +164,9 @@ typedef long mp_off_t;
 #define MICROPY_PY_SYS_PATH_ARGV_DEFAULTS (0)
 
 // Enable sys.executable.
+#ifndef MICROPY_PY_SYS_EXECUTABLE
 #define MICROPY_PY_SYS_EXECUTABLE (1)
+#endif
 
 #define MICROPY_PY_SOCKET_LISTEN_BACKLOG_DEFAULT (SOMAXCONN < 128 ? SOMAXCONN : 128)
 
--- a/ports/unix/unix_mphal.c
+++ b/ports/unix/unix_mphal.c
@@ -43,7 +43,13 @@
 #endif
 #endif
 
-#ifndef _WIN32
+#if defined(_WIN32) || defined(__wasi__)
+#define HAS_SIGNAL 0
+#else
+#define HAS_SIGNAL 1
+#endif
+
+#if HAS_SIGNAL
 #include <signal.h>
 
 static void sighandler(int signum) {
@@ -75,7 +81,7 @@ static void sighandler(int signum) {
 void mp_hal_set_interrupt_char(char c) {
     // configure terminal settings to (not) let ctrl-C through
     if (c == CHAR_CTRL_C) {
-        #ifndef _WIN32
+        #if HAS_SIGNAL
         // enable signal handler
         struct sigaction sa;
         sa.sa_flags = 0;
@@ -84,7 +90,7 @@ void mp_hal_set_interrupt_char(char c) {
         sigaction(SIGINT, &sa, NULL);
         #endif
     } else {
-        #ifndef _WIN32
+        #if HAS_SIGNAL
         // disable signal handler
         struct sigaction sa;
         sa.sa_flags = 0;
--- a/ports/unix/variants/mpconfigvariant_common.h
+++ b/ports/unix/variants/mpconfigvariant_common.h
@@ -95,7 +95,9 @@
 #define MICROPY_PY_OS_INCLUDEFILE      "ports/unix/modos.c"
 #define MICROPY_PY_OS_ERRNO            (1)
 #define MICROPY_PY_OS_GETENV_PUTENV_UNSETENV (1)
+#ifndef MICROPY_PY_OS_SYSTEM
 #define MICROPY_PY_OS_SYSTEM           (1)
+#endif
 #define MICROPY_PY_OS_URANDOM          (1)
 
 // Enable the unix-specific "time" module.
@@ -105,11 +107,15 @@
 #define MICROPY_PY_TIME_INCLUDEFILE    "ports/unix/modtime.c"
 
 // The "select" module is enabled by default, but disable select.select().
+#ifndef MICROPY_PY_SELECT_POSIX_OPTIMISATIONS
 #define MICROPY_PY_SELECT_POSIX_OPTIMISATIONS (1)
+#endif
 #define MICROPY_PY_SELECT_SELECT       (0)
 
 // Enable the "websocket" module.
+#ifndef MICROPY_PY_WEBSOCKET
 #define MICROPY_PY_WEBSOCKET           (1)
+#endif
 
 // Enable the "machine" module, mostly for machine.mem*.
 #define MICROPY_PY_MACHINE             (1)
--- a/ports/unix/variants/wasi/mpconfigvariant.mk
+++ b/ports/unix/variants/wasi/mpconfigvariant.mk
@@ -45,3 +45,18 @@ MICROPY_USE_READLINE = 0
 MICROPY_VFS_FAT = 0
 MICROPY_VFS_LFS1 = 0
 MICROPY_VFS_LFS2 = 0
+
+PROG ?= micropython
+
+# LLVM still uses the older version of EH proposal. ("phase 3")
+# Convert to the latest version of EH proposal with exnref.
+$(BUILD)/$(PROG).spilled.exnref: $(BUILD)/$(PROG).spilled
+	$(WASM_OPT) --translate-to-exnref --enable-exception-handling \
+	-o $(BUILD)/$(PROG).spilled.exnref $(BUILD)/$(PROG).spilled
+
+# Unlike emscripten, WASI doesn't provide a way to scan GC roots
+# like WASM locals. Hopefully --spill-pointers can workaround it.
+$(BUILD)/$(PROG).spilled: $(BUILD)/$(PROG)
+	$(WASM_OPT) --spill-pointers -o $(BUILD)/$(PROG).spilled $(BUILD)/$(PROG)
+
+all: $(BUILD)/$(PROG).spilled.exnref
--- a/ports/unix/variants/wasi/mpconfigvariant.mk
+++ b/ports/unix/variants/wasi/mpconfigvariant.mk
@@ -1,0 +1,47 @@
+# prerequisites
+#
+# WASI_SDK: wasi-sdk 25.0 or later
+# WASM_OPT: binaryen wasm-opt version_117 or later
+
+# Note:
+# we specify the target "--target=wasm32-wasi" explicitly below for the case
+# where $CLANG is built with a configuration different from wasi-sdk.
+# ditto for "-B $(WASI_SDK)/bin/".
+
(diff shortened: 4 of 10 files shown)


ports/unix/mpconfigport.h
```diff
@@ -154,8 +154,12 @@ typedef long mp_off_t;
 // Don't default sys.argv and sys.path because we do that in main.
 #define MICROPY_PY_SYS_PATH_ARGV_DEFAULTS (0)
 
+#if defined(__wasi__)
```
stinos, line 157: Should be done like 

```
#ifndef  MICROPY_PY_SYS_EXECUTABLE
#define MICROPY_PY_SYS_EXECUTABLE (1)
#endif
```

so it works generically, then make a separate mpconfigvariant_wasi.h where it gets set to t 0.

ports/unix/variants/mpconfigvariant_common.h
```diff
@@ -91,7 +91,11 @@
 #define MICROPY_PY_OS_ERRNO            (1)
 #define MICROPY_PY_OS_GETENV_PUTENV_UNSETENV (1)
 #define MICROPY_PY_OS_SEP              (1)
+#if defined(__wasi__)
```
stinos, line 94: Similar comment as above: instead of modifying this variant it would be better to create it separately

ports/unix/mpconfigport.h
yamt: who would include mpconfigvariant_wasi.h?

stinos: That would be done automatically if you use `make VARIANT=wasi` (though, my mistake, it should actually be unix/variants/wasi/mpconfigvariant.h). Likewise instead of build.sh you'd implement the specific makefile options in unix/variants/wasi/mpconfigvariant.mk). Check out the other variants for examples. Also I think to have any chance of merging this ulab should probably not be included unconditionally because this makes it harder to build out of the box; on the other hand if this only works with non-approved upstream wasi patches it's not super useful for others anyway.

dpgeorge: Yes, please make this a new variant.  It should then be possible to build it via `make VARIANT=wasi`.

yamt: ok. i was not aware of the mechanism. thank you. let me try to adapt.

yamt: > Also I think to have any chance of merging this ulab should probably not be included unconditionally because this makes it harder to build out of the box; on the other hand if this only works with non-approved upstream wasi patches it's not super useful for others anyway.

sure.

wrt wasi patches, alternatively it's possible to put the setjmp/longjmp runtime within micropython itself. at least until when setjmp/longjmp gets ready in wasi. how do you think?

anyway, it won't be super useful until wasm EH is widely implemented in runtimes though.

dpgeorge: If building unix for wasi requires patches to wasi/binaryen then IMO this PR should stay as a draft until those patches are no longer needed.  You could put the setjmp/longjmp code here, but I guess the other patches are still needed.

dpgeorge:
See semi-related #11723, which gets `mpy-cross` compiling to WASM (but not WASI at this stage).

ports/unix/mpconfigport.h
yamt: i adapted to use VARIANT. it simplify a lot. thank you.

i'm not sure what should be done for UNAME_S logic in Makefile though.
a build for WASI is actually a cross compile. but the existing CROSS_COMPILE seems geared to "arm-none-eabi-gcc" style cross toolchain, which doesn't seem suitable to wasi-sdk style cross toolchain.

yamt: > If building unix for wasi requires patches to wasi/binaryen then IMO this PR should stay as a draft until those patches are no longer needed. You could put the setjmp/longjmp code here, but I guess the other patches are still needed.

FYI, all toolchain patches have been merged in the upstream since then.

yamt:
is there any TODO for this PR while waiting for LLVM 19?

yamt:
the last push is a conflict resolution

yamt:
now wasi-sdk 25 has been released and this only requires released versions of the toolchain.
can you consider to make this land?

yamt:
the ci failure doesn't seem related to this PR.
```
Cloning into '/home/runner/work/micropython/micropython/lib/asf4'...
fatal: unable to access 'https://github.com/adafruit/asf4/': The requested URL returned error: 403
fatal: clone of 'https://github.com/adafruit/asf4' into submodule path '/home/runner/work/micropython/micropython/lib/asf4' failed
Failed to clone 'lib/asf4'. Retry scheduled
```

yamt:
@dpgeorge @stinos  does this have any blockers?

ports/unix/variants/wasi/mpconfigvariant.mk
```diff
@@ -0,0 +1,28 @@
+# Build for WASI
+
+FROZEN_MANIFEST ?= $(VARIANT_DIR)/manifest.py
+
+# WASI doesn't have FFI
+MICROPY_PY_FFI=0
```
stinos, line 6: Should have whitespace around the ` = `

ports/unix/variants/wasi/mpconfigvariant.mk
```diff
@@ -0,0 +1,28 @@
+# Build for WASI
```
stinos, line 1: Imo not needed to state the obvious.,

ports/unix/variants/wasi/manifest.py
```diff
@@ -0,0 +1,3 @@
+# include("$(PORT_DIR)/variants/manifest.py")
```
stinos, line 1: Since this file only has comments can it just be omitted?

ports/unix/unix_mphal.c
```diff
@@ -43,7 +43,13 @@
...
 #endif
 
-#ifndef _WIN32
+#if defined(_WIN32) || defined(__wasi__)
+#define HAS_SIGNAL 0
+#else
+#define HAS_SIGNAL 1
+#endif
+
+#if HAS_SIGNAL != 0
```
stinos, line 52: Just `#if HAS_SIGNAL` is good enough?

ports/unix/build-wasi.sh
```diff
@@ -0,0 +1,55 @@
...
+STRIP="${WASI_SDK}/bin/strip" \
+SIZE="${WASI_SDK}/bin/size" \
+UNAME_S=wasi \
+VARIANT=wasi \
+"$@"
+
+PROG=build-wasi/micropython
+
+# Unlike emscripten, WASI doesn't provide a way to scan GC roots
+# like WASM locals. Hopefully --spill-pointers can workaround it.
```
stinos, line 36: Did you try this with e.g. long-running scripts?

stinos:
Apart from some cosmetic changes (see somments):
- can you squash this so only relevant commits are left? It probably comes down to a commit which makes some PP defs in the unix port configurable, a commit which fixes some unix build stuff, and then one commit which adds the WASI variant.
- it seems like almost none of the other ports/variants require a .sh file to build them, the goal is to have builds like `make VARIANT=wasi WASI_SDK=/opt/wasi-sdk-25.0`. So I wonder how much of build-wasi.sh is really needed: all build flags can probably go in mpconfigvariant.mk? That would be more consistent with how variants are used in general. And I think, not 100% sure though, you could also get those post-build `${WASM_OPT}` call in there by changing the `STRIP` definition to first strip then call wasm-opt.
- in general ports/variants shouldn't go stale and the first step in making that happen is having the CI build it so errors can be caught early. Which would come down to building the WASI variant as part of the github workflows (see tools/ci.sh and workflows/ports_unix.yml), and then running it, for instance have it run all unit tests. Would that be possible somehow?

yamt:
@stinos thank you for comments. i will take my time to digest.

ports/unix/unix_mphal.c
yamt: done

ports/unix/build-wasi.sh
yamt: no

yamt:
> Apart from some cosmetic changes (see somments):
> 
>     * can you squash this so only relevant commits are left? It probably comes down to a commit which makes some PP defs in the unix port configurable, a commit which fixes some unix build stuff, and then one commit which adds the WASI variant.

i will consider squashing later.
maybe one of the last things to do.
 
>     * it seems like almost none of the other ports/variants require a .sh file to build them, the goal is to have builds like `make VARIANT=wasi WASI_SDK=/opt/wasi-sdk-25.0`. So I wonder how much of build-wasi.sh is really needed: all build flags can probably go in mpconfigvariant.mk? That would be more consistent with how variants are used in general. And I think, not 100% sure though, you could also get those post-build `${WASM_OPT}` call in there by changing the `STRIP` definition to first strip then call wasm-opt.

partly done.
but i'm not sure if it's a good idea to abuse STRIP for this purpose
because STRIP seems to be documented as user-overridable.

>     * in general ports/variants shouldn't go stale and the first step in making that happen is having the CI build it so errors can be caught early. Which would come down to building the WASI variant as part of the github workflows (see tools/ci.sh and workflows/ports_unix.yml), and then running it, for instance have it run all unit tests. Would that be possible somehow?

ok.

i have run some of test cases in this repo with toywasm locally.

yamt:
@stinos i think i have solved most of your concerns. thank you.

yamt:
the last push was a conflict resolution.

yamt:
@dpgeorge @stinos can you revisit when/if you have time? thank you.

stinos:
Seems pretty much ok to me, just wondering if it's not possible to leave the Makefile as-is and put those changes in mpconfigvariant.mk (which gets included in the Makefile anyway)? For the rest it's up to Damien.

yamt:
> Seems pretty much ok to me, just wondering if it's not possible to leave the Makefile as-is and put those changes in mpconfigvariant.mk (which gets included in the Makefile anyway)? For the rest it's up to Damien.

do you mean to put the extra rules there? it might be possible.
but i thought mpconfigvariant.mk was meant only for declarative things like setting variables.

stinos:
> but i thought mpconfigvariant.mk was meant only for declarative things like setting variables

That's mostly how it gets currently used, but I don't think that it's a fixed rule, could put anything in there

yamt:
> > but i thought mpconfigvariant.mk was meant only for declarative things like setting variables
> 
> That's mostly how it gets currently used, but I don't think that it's a fixed rule, could put anything in there

ok. i moved the rules. i had to duplicate `PROG ?= micropython`.

yamt:
the last push was a mechanical rebase

yamt:
to me, the ci failure (Address already in use / Connection refused) seems unrelated to the changes in this PR.

yamt:
the last push was a conflict resolution.

yamt:
@dpgeorge @stinos this PR is now almost two years old. maybe you can consider to merge?

matrey:
FYI I just stumbled upon this: https://github.com/agentskillmania/build-micro-python-wasi/blob/main/micro-python-wasi.README.md "MicroPython (v1.27.0) compiled to WebAssembly (WASI Preview2) for sandboxed execution."

PR #13680, opened 2024-02-15
esp32: Increase NimBLE task stack size and overflow detection headroom.

dpgeorge:
The Python BLE IRQ handler will most likely run on the NimBLE task, so its C stack must be large enough to accommodate reasonably complicated Python code (eg a few call depths).  So increase this stack size.

Also increase the headroom from 1024 to 2048 bytes.  This is needed because (1) esp32 architecture uses a fair amount of stack in general; (2) by the time execution gets to setting the Python stack top via `mp_stack_set_top()` in this interlock code, about 600 bytes of stack are already used, which reduces the amount available for Python.

Fixes issue #12349.


dpgeorge:
@antirez From my tests this should fix your issue with BLE task stack overflow.

Can you please test this PR?

antirez:
Appreciated! Will test it later today and report back.

antirez:
Good, news @dpgeorge! Now the BLE issue is fixed testing with latest master commit + your fix. Thank you.

I noticed that the device still crashes when I try to call the memory functions you adviced me to call. Not a problem for me, but just as a pure report, I added at the end of this message what I get.

There is now to understand if this also fixes the problem I have in the LoRa T-WATCH device from Lilygo. In the watch, the issue was different: as soon as I initialized the BLE stack, MicroPython restarts, yet I'll test this last commit and in case will open a new issue, because it is unrelated to what was reported/fixed here. Thanks.

```
Guru Meditation Error: Core  0 panic'ed (StoreProhibited). Exception was unhandled.

Core  0 register dump:
PC      : 0x4009d9e5  PS      : 0x00060033  A0      : 0x8009d59f  A1      : 0x3fff7270  
A2      : 0x00000000  A3      : 0x000006a0  A4      : 0x00000400  A5      : 0x00000100  
A6      : 0x0078f971  A7      : 0x00060023  A8      : 0x00000000  A9      : 0x3fff27fc  
A10     : 0x00000008  A11     : 0x3ffe4364  A12     : 0x3ffe4388  A13     : 0x3ffe43b8  
A14     : 0x00000020  A15     : 0x0000000a  SAR     : 0x0000001a  EXCCAUSE: 0x0000001d  
EXCVADDR: 0x0000000c  LBEG    : 0x4000c2e0  LEND    : 0x4000c2f6  LCOUNT  : 0xffffffff  


Backtrace: 0x4009d9e2:0x3fff7270 0x4009d59c:0x3fff7290 0x40082258:0x3fff72b0 0x400822a0:0x3fff72d0 0x4009e149:0x3fff72f0 0x40092749:0x3fff7310 0x40204f7d:0x3fff7330 0x40207c94:0x3fff7360 0x40208ab0:0x3fff7390 0x40208faf:0x3fff73d0 0x40209063:0x3fff7420 0x40204006:0x3fff7460
```

dpgeorge:
> Now the BLE issue is fixed testing with latest master commit + your fix. Thank you.

Very good!  Thanks for testing.

> I noticed that the device still crashes when I try to call the memory functions you adviced me to call.

Hmm, I think that might be because the IDF doesn't allow calling it's heap query functions from within the BLE task??  Well, that's a separate issue and can be dealt with later if needed.

dpgeorge:
I added a test for BLE IRQ handler stack overflow.  It fails on ESP32_GENERIC prior to this PR, and passes with this PR applied.

PR #13682, opened 2024-02-16
py/obj: Introduce mp_obj_malloc_with_finaliser to allocate and set type.

dpgeorge:
Following 709e8328d9812a2e02da6fba65a03f9a873d60a2
    
Using this helps to reduce code size.  And it ensure that the type is always set as soon as the object is allocated, which is important for the GC to function correctly.

### DIFF
--- a/extmod/modtls_mbedtls.c
+++ b/extmod/modtls_mbedtls.c
@@ -210,11 +210,10 @@ STATIC mp_obj_t ssl_context_make_new(const mp_obj_type_t *type_in, size_t n_args
 
     // Create SSLContext object.
     #if MICROPY_PY_SSL_FINALISER
-    mp_obj_ssl_context_t *self = m_new_obj_with_finaliser(mp_obj_ssl_context_t);
+    mp_obj_ssl_context_t *self = mp_obj_malloc_with_finaliser(mp_obj_ssl_context_t, type_in);
     #else
-    mp_obj_ssl_context_t *self = m_new_obj(mp_obj_ssl_context_t);
+    mp_obj_ssl_context_t *self = mp_obj_malloc(mp_obj_ssl_context_t, type_in);
     #endif
-    self->base.type = type_in;
 
     // Initialise mbedTLS state.
     mbedtls_ssl_config_init(&self->conf);
@@ -488,11 +487,10 @@ STATIC mp_obj_t ssl_socket_make_new(mp_obj_ssl_context_t *ssl_context, mp_obj_t 
     mp_get_stream_raise(sock, MP_STREAM_OP_READ | MP_STREAM_OP_WRITE | MP_STREAM_OP_IOCTL);
 
     #if MICROPY_PY_SSL_FINALISER
-    mp_obj_ssl_socket_t *o = m_new_obj_with_finaliser(mp_obj_ssl_socket_t);
+    mp_obj_ssl_socket_t *o = mp_obj_malloc_with_finaliser(mp_obj_ssl_socket_t, &ssl_socket_type);
     #else
-    mp_obj_ssl_socket_t *o = m_new_obj(mp_obj_ssl_socket_t);
+    mp_obj_ssl_socket_t *o = mp_obj_malloc(mp_obj_ssl_socket_t, &ssl_socket_type);
     #endif
-    o->base.type = &ssl_socket_type;
     o->ssl_context = ssl_context;
     o->sock = sock;
     o->poll_mask = 0;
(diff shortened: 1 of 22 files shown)


dpgeorge:
Needs #13683 merged first.

extmod/vfs_lfsx_file.c
```diff
@@ -90,11 +90,11 @@ mp_obj_t MP_VFS_LFSx(file_open)(mp_obj_t self_in, mp_obj_t path_in, mp_obj_t mod
     }
 
     #if LFS_BUILD_VERSION == 1
-    MP_OBJ_VFS_LFSx_FILE *o = m_new_obj_var_with_finaliser(MP_OBJ_VFS_LFSx_FILE, file_buffer, uint8_t, self->lfs.cfg->prog_size);
+    MP_OBJ_VFS_LFSx_FILE *o = mp_obj_malloc_var_with_finaliser(MP_OBJ_VFS_LFSx_FILE, uint8_t, self->lfs.cfg->prog_size, type);
     #else
-    MP_OBJ_VFS_LFSx_FILE *o = m_new_obj_var_with_finaliser(MP_OBJ_VFS_LFSx_FILE, file_buffer, uint8_t, self->lfs.cfg->cache_size);
-    #endif
+    MP_OBJ_VFS_LFSx_FILE *o = mp_obj_malloc_var_with_finaliser(MP_OBJ_VFS_LFSx_FILE, uint8_t, self->lfs.cfg->cache_size, type);
     o->base.type = type;
```
andrewleech, line 96: This one's still here

py/obj.h
```diff
@@ -916,6 +916,15 @@ extern const struct _mp_obj_exception_t mp_const_GeneratorExit_obj;
 #define mp_obj_malloc_var(struct_type, var_type, var_num, obj_type) ((struct_type *)mp_obj_malloc_helper(sizeof(struct_type) + sizeof(var_type) * (var_num), obj_type))
 void *mp_obj_malloc_helper(size_t num_bytes, const mp_obj_type_t *type);
 
+#if MICROPY_ENABLE_FINALISER
+#define mp_obj_malloc_with_finaliser(struct_type, obj_type) ((struct_type *)mp_obj_malloc_with_finaliser_helper(sizeof(struct_type), obj_type))
+#define mp_obj_malloc_var_with_finaliser(struct_type, var_type, var_num, obj_type) ((struct_type *)mp_obj_malloc_with_finaliser_helper(sizeof(struct_type) + sizeof(var_type) * (var_num), obj_type))
+void *mp_obj_malloc_with_finaliser_helper(size_t num_bytes, const mp_obj_type_t *type);
+#else
+//#define mp_obj_malloc_with_finaliser(struct_type, obj_type) mp_obj_malloc(struct_type, obj_type)
+//#define mp_obj_malloc_var_with_finaliser(struct_type, var_type, var_num, obj_type) mp_obj_malloc_var(struct_type, var_type, var_num, obj_type)
```
andrewleech, line 925: Were these supposed to be left commented out?

andrewleech: Ah I see this is just part of the test commit

andrewleech:
I do think where you've moved the macros to makes sense, I struggled a bit to decide where I should put the new functionality.... the consolidation is sensible.

extmod/modtls_mbedtls.c
```diff
@@ -210,11 +210,10 @@ STATIC mp_obj_t ssl_context_make_new(const mp_obj_type_t *type_in, size_t n_args
 
     // Create SSLContext object.
     #if MICROPY_PY_SSL_FINALISER
-    mp_obj_ssl_context_t *self = m_new_obj_with_finaliser(mp_obj_ssl_context_t);
+    mp_obj_ssl_context_t *self = mp_obj_malloc_with_finaliser(mp_obj_ssl_context_t, type_in);
     #else
-    mp_obj_ssl_context_t *self = m_new_obj(mp_obj_ssl_context_t);
+    mp_obj_ssl_context_t *self = mp_obj_malloc(mp_obj_ssl_context_t, type_in);
     #endif
```
andrewleech, line 216: I did think this one and the other like it could just have the `#if MICROPY_PY_SSL_FINALISER` block removed and have just the `mp_obj_malloc_with_finaliser` version here as it falls back to `mp_obj_malloc` if disabled.

extmod/vfs_lfsx_file.c
dpgeorge: Thanks!  Now fixed.

py/obj.h
dpgeorge: It's a test to see if they are needed...

dpgeorge: I've left this in.

Issue #13672, opened 2024-02-15
unix port: intermittent infinite loop in import / vfs_reader handling.


andrewleech:
We're running unix micropython with `unittest` from micropython-lib to test our application codebase.
This is running micropython @ 4a2e510a

I'm seeing an intermittent problem where the tests sometimes lock up part way through. 

I've finally managed to capture it while running micropython through gdb and the deadlock is during a module import. This happens to be importing an mpy file, not sure if it happens with py files as well at this stage.

Call Stack:
```
mp_reader_vfs_readbyte(void * data) (\code\src\micropython\extmod\vfs_reader.c:55)
read_uint(mp_reader_t * reader) (\code\src\micropython\py\persistentcode.c:154)
load_raw_code(mp_reader_t * reader, mp_module_context_t * context) (\code\src\micropython\py\persistentcode.c:226)
load_raw_code(mp_reader_t * reader, mp_module_context_t * context) (\code\src\micropython\py\persistentcode.c:316)
load_raw_code(mp_reader_t * reader, mp_module_context_t * context) (\code\src\micropython\py\persistentcode.c:316)
mp_raw_code_load(mp_reader_t * reader, mp_compiled_module_t * cm) (\code\src\micropython\py\persistentcode.c:439)
mp_raw_code_load_file(qstr filename, mp_compiled_module_t * context) (\code\src\micropython\py\persistentcode.c:462)
do_load(mp_module_context_t * module_obj, vstr_t * file) (\code\src\micropython\py\builtinimport.c:248)
process_import_at_level(qstr full_mod_name, qstr level_mod_name, mp_obj_t outer_module_obj, _Bool override_main) (\code\src\micropython\py\builtinimport.c:512)
mp_builtin___import___default(size_t n_args, const mp_obj_t * args) (\code\src\micropython\py\builtinimport.c:607)
```

In the current `mp_reader_vfs_readbyte()` function:
```
reader.bufpos = 53
reader.buflen = 0
reader.bufsize = 53
```
so it's returning `MP_READER_EOF (-1)`

One function up, in:
```
STATIC size_t read_uint(mp_reader_t *reader) {
    size_t unum = 0;
    for (;;) {
        byte b = reader->readbyte(reader->data);
        unum = (unum << 7) | (b & 0x7f);
        if ((b & 0x80) == 0) {
            break;
        }
    }
    return unum;
}
```
There's no detection / handling of `0xFF` from `reader->readbyte()` so it's looking forever.

[...]

PR #13684, opened 2024-02-16
extmod/vfs_posix_file: Ensure file object has safe default fd.

andrewleech:
If file open fails the object will have fd = -1 (closed).

For detailed investigation see https://github.com/micropython/micropython/issues/13672

### DIFF
--- a/extmod/vfs_posix_file.c
+++ b/extmod/vfs_posix_file.c
@@ -92,6 +92,7 @@ mp_obj_t mp_vfs_posix_file_open(const mp_obj_type_t *type, mp_obj_t file_in, mp_
     }
 
     mp_obj_vfs_posix_file_t *o = mp_obj_malloc_with_finaliser(mp_obj_vfs_posix_file_t, type);
+    o->fd = -1; // In case open() fails below, initialise this as a "closed" file object.
 
     mp_obj_t fid = file_in;
 


extmod/vfs_posix_file.c
```diff
@@ -92,18 +91,21 @@ mp_obj_t mp_vfs_posix_file_open(const mp_obj_type_t *type, mp_obj_t file_in, mp_
         }
     }
 
-    o->base.type = type;
-
     mp_obj_t fid = file_in;
 
     if (mp_obj_is_small_int(fid)) {
+        // When opening an existing file description, don't close it on deletion
+        mp_obj_vfs_posix_file_t *o = m_new_obj(mp_obj_vfs_posix_file_t);
```
dpgeorge, line 98: description -> descriptor

And please use `mp_obj_malloc()`.

andrewleech: Thanks, fix pushed as suggested!

extmod/vfs_posix_file.c
```diff
@@ -92,18 +91,17 @@ mp_obj_t mp_vfs_posix_file_open(const mp_obj_type_t *type, mp_obj_t file_in, mp_
...
+        fd = MP_OBJ_SMALL_INT_VALUE(fid);
+    } else {
+        const char *fname = mp_obj_str_get_str(fid);
+        MP_HAL_RETRY_SYSCALL(fd, open(fname, mode_x | mode_rw, 0644), mp_raise_OSError(err));
     }
-
-    const char *fname = mp_obj_str_get_str(fid);
-    int fd;
-    MP_HAL_RETRY_SYSCALL(fd, open(fname, mode_x | mode_rw, 0644), mp_raise_OSError(err));
+    mp_obj_vfs_posix_file_t *o = m_new_obj_with_finaliser(mp_obj_vfs_posix_file_t);
```
dlech, line 103: If this raises an exception, e.g. due to out of memory, the open file handle will leak.

dlech (commented):
Does this work?

```diff
diff --git a/extmod/vfs_posix_file.c b/extmod/vfs_posix_file.c
index 2f7078954..8f0e7aba4 100644
--- a/extmod/vfs_posix_file.c
+++ b/extmod/vfs_posix_file.c
@@ -64,6 +64,9 @@ STATIC void vfs_posix_file_print(const mp_print_t *print, mp_obj_t self_in, mp_p
 
 mp_obj_t mp_vfs_posix_file_open(const mp_obj_type_t *type, mp_obj_t file_in, mp_obj_t mode_in) {
     mp_obj_vfs_posix_file_t *o = m_new_obj_with_finaliser(mp_obj_vfs_posix_file_t);
+    o->base.type = type;
+    o->fd = -1;
+
     const char *mode_s = mp_obj_str_get_str(mode_in);
 
     int mode_rw = 0, mode_x = 0;
@@ -92,8 +95,6 @@ mp_obj_t mp_vfs_posix_file_open(const mp_obj_type_t *type, mp_obj_t file_in, mp_
         }
     }
 
-    o->base.type = type;
-
     mp_obj_t fid = file_in;
 
     if (mp_obj_is_small_int(fid)) {
```

extmod/vfs_posix_file.c
andrewleech: the type needs to be set down where it was originally as it gets changed by the switch in the function, but yes pre-seeing fd to -1 was an earlier version of the fix I had. I was thinking it's safer to not leak the object at all if the open fails but hadn't considered the opposite issue!

andrewleech: Thanks I've updated as such

andrewleech:
I've updated the commit message and title to reflect the way this fix now works.

extmod/vfs_posix_file.c
```diff
@@ -96,14 +97,13 @@ mp_obj_t mp_vfs_posix_file_open(const mp_obj_type_t *type, mp_obj_t file_in, mp_
 
     mp_obj_t fid = file_in;
 
+    int fd;
```
dpgeorge, line 100: I don't think this line or any below it needs to change.  The only change should be `o->fd=-1` just after creating the object.

andrewleech: Yeah ok, the other changes could go in a separate commit (/pr); they're just a code size improvement with no functional change so I left them in.

dpgeorge: Are they size improvements?  I would think the compiler is pretty good at optimising the `return`.

andrewleech: Oh, I was sure I saw a size improvement, but sure enough I just compiled Unix before and after this change with no difference in size.
Maybe I had previously looked at size before / after a rebase as well.

Regardless, I've simplified this PR as suggested!

PR #13686, opened 2024-02-16
py/emitnative: Simplify how the prelude is found for native functions, and support `__name__' on them

dpgeorge:
With this PR, native functions (via `@micropython.native`) are now a little smaller because they don't have to load their prelude pointer, that's handled instead by the runtime.

And, native functions now support `__name__` to get their function name.


dpgeorge:
This is a follow-up to #13628, which decreased code size by a very large amount, so the small increase here is acceptable.

PR #13689, opened 2024-02-16
extmod/network: Implement IPv6 API to set and retrieve nic configuration.

### LINKED ISSUE #12600: IPv6 APIs
There are already various bugs tracking adding IPv6 to the various ports:

* ESP32: #3683, #11111
* RP2: #9108
* Other: #7563
* Maybe more...?

But, even with these changes, there are some critical missing parts: the APIs. For example, `<netif>.ifconfig()` returns a tuple of the IPv4 address configuration. However, there is no API to get the IPv6 addresses of the device, and it looks like such a thing isn't even defined yet.

So I wanted to open this issue to discuss potential API additions to help make IPv6 a first-class citizen in micropython.

Specifically, we need the ability to:

1. List the current IPv6 addresses assigned to an interface, along with the status of each address
2. Statically configure IPv6 addresses, along with adding the associated on-link routes
3. Turn on/off SLAAC (if support is compiled in)
4. Turn on/off DHCPv6 (if support is compiled in)
5. Set a preference for IPv6 vs IPv4 when performing name lookups that have both A and AAAA records.

We currently don't have any way to do any of this from Micropython, so I think we should at least start to define what this should look like.

For compatibility reasons, I think we should avoid extending the `ifconfig()` method. Lots of code seems pretty dependent on how this method works. I was thinking something like this as a starting point:

* `<netif>.MAX_IPV6_ADDRS` - constant describing the maximum number of IPv6 addresses that can be assigned to this interface. (Compile-time default seems to be 3)
* `<netif>.v6_addrs()` - Returns a list of tuples, one tuple per address. Each tuple would have the following fields:
    1. IPv6 Address (as a string?)
    2. Flags (Temporary, Deprecated, SLAAC, DHCPv6, Static, on-link, etc)
    3. remaining preferred lifetime (or `None`)
    4. remaining valid lifetime (or `None`)
* `<netif>.v6_del_addr(<addr>)` - Removes the given IPv6 address.
* `<netif>.v6_add_addr(<addr>, <prefix-len>)` - Adds the given IPv6 address, optionally adding a route for other on-link devices with the same prefix.
* `<netif>.config('host-lookup-pref', <'v6'|'v4'>)` - Configures the host lookup preference when there are both A and AAAA records.
* `<netif>.config('slaac_en', <True|False>)` - Turns on or off SLAAC.
* `<netif>.config('dhcpv6_en', <True|False>)` -  Turns on or off DHCPv6.
 
I'm sure this could be improved upon, but wanted to go ahead and put this out there as a starting point.

Thoughts?

felixdoerre:
fixes #12600

This change implements a new `<AbstractNIC>.ipconfig` function for the CYW43 NIC class, to set and retrieve network configuration for a NIC. Currently this change supports:

- `ipconfig("addr4")` obtain a tuple `(addr, netmask)` of the currently configured ipv4 address
- `ipconfig("addr6")` obtain a list of tuples `(addr, state, prefered_lifetime, valid_lifetime)` of all currently active ipv6 addresses. This includes static, slaac and link-local addresses
- `ipconfig(addr4="1.2.3.4/24")` to set the ipv4 address and netmask
- `ipconfig(addr6="2a01::2")` to set a static ipv6 address. Note that this does not configure an interface route, as this does not seem supported by lwip
- `ipconfig(autoconf6=True)` to enable ipv6 network configuration with slaac (and wait up to 10s for an address to be received)
- `ipconfig(gw4="1.2.3.1")` to set the ipv4 gatewas
- `network.ipconfig(dns="2a01::2")` set the primary dns server (can be a ipv4 or ipv6 address).
- `ipconfig(dhcp4=True)` enable, and wait for (up to 10s), ipv4 dhcp. This sets ipv4 address, netmask, gateway and a dns server. `ipconfig(dhcp4=False)` stops dhcp, releases the ip, and clears the configured ipv4 address.
- `ipconfig(dhcp6=True)` enable stateless dhcpv6 to obtain a dns server
- `netwok.ipconfig(prefer=6)` to prefer ipv6 addresses to be returned as dns responses (falling back to ipv4 if the host does not have an ipv6 address). Note that this does not flush the dns cache, so if a host is already in the dns cache with its v4 address, subsequent lookups will return that address even if `prefer=6` is set.

This interface replaces `network.ifconfig` completely, and  `ifconfig` should be marked as deprecated and removed in a future version.

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -38,6 +38,7 @@
 #if MICROPY_PY_LWIP
 
 #include "shared/netutils/netutils.h"
+#include "modnetwork.h"
 
 #include "lwip/init.h"
 #include "lwip/tcp.h"
@@ -353,8 +354,12 @@ static void lwip_socket_free_incoming(lwip_socket_obj_t *socket) {
     }
 }
 
+#if LWIP_VERSION_MAJOR < 2
+#define IPADDR_STRLEN_MAX 46
+#endif
 mp_obj_t lwip_format_inet_addr(const ip_addr_t *ip, mp_uint_t port) {
-    char *ipstr = ipaddr_ntoa(ip);
+    char ipstr[IPADDR_STRLEN_MAX];
+    ipaddr_ntoa_r(ip, ipstr, sizeof(ipstr));
     mp_obj_t tuple[2] = {
         tuple[0] = mp_obj_new_str(ipstr, strlen(ipstr)),
         tuple[1] = mp_obj_new_int(port),
@@ -1750,7 +1755,11 @@ static mp_obj_t lwip_getaddrinfo(size_t n_args, const mp_obj_t *args) {
     state.status = 0;
 
     MICROPY_PY_LWIP_ENTER
+    #if LWIP_VERSION_MAJOR < 2
     err_t ret = dns_gethostbyname(host, (ip_addr_t *)&state.ipaddr, lwip_getaddrinfo_cb, &state);
+    #else
+    err_t ret = dns_gethostbyname_addrtype(host, (ip_addr_t *)&state.ipaddr, lwip_getaddrinfo_cb, &state, mp_mod_network_prefer_dns_use_ip_version == 4 ? LWIP_DNS_ADDRTYPE_IPV4_IPV6 : LWIP_DNS_ADDRTYPE_IPV6_IPV4);
+    #endif
     MICROPY_PY_LWIP_EXIT
 
     switch (ret) {
--- a/extmod/network_cyw43.c
+++ b/extmod/network_cyw43.c
@@ -321,6 +321,12 @@ static mp_obj_t network_cyw43_ifconfig(size_t n_args, const mp_obj_t *args) {
 }
 MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(network_cyw43_ifconfig_obj, 1, 2, network_cyw43_ifconfig);
 
+static mp_obj_t network_cyw43_ipconfig(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs) {
+    network_cyw43_obj_t *self = MP_OBJ_TO_PTR(args[0]);
+    return mod_network_nic_ipconfig(&self->cyw->netif[self->itf], n_args - 1, args + 1, kwargs);
+}
+static MP_DEFINE_CONST_FUN_OBJ_KW(network_cyw43_ipconfig_obj, 1, network_cyw43_ipconfig);
+
 static mp_obj_t network_cyw43_status(size_t n_args, const mp_obj_t *args) {
     network_cyw43_obj_t *self = MP_OBJ_TO_PTR(args[0]);
     (void)self;
@@ -532,6 +538,7 @@ static const mp_rom_map_elem_t network_cyw43_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_disconnect), MP_ROM_PTR(&network_cyw43_disconnect_obj) },
     { MP_ROM_QSTR(MP_QSTR_isconnected), MP_ROM_PTR(&network_cyw43_isconnected_obj) },
     { MP_ROM_QSTR(MP_QSTR_ifconfig), MP_ROM_PTR(&network_cyw43_ifconfig_obj) },
+    { MP_ROM_QSTR(MP_QSTR_ipconfig), MP_ROM_PTR(&network_cyw43_ipconfig_obj) },
     { MP_ROM_QSTR(MP_QSTR_status), MP_ROM_PTR(&network_cyw43_status_obj) },
     { MP_ROM_QSTR(MP_QSTR_config), MP_ROM_PTR(&network_cyw43_config_obj) },
 
--- a/extmod/network_lwip.c
+++ b/extmod/network_lwip.c
@@ -26,6 +26,7 @@
 
 #include "py/runtime.h"
 #include "py/mphal.h"
+#include "py/parsenum.h"
 
 #if MICROPY_PY_NETWORK && MICROPY_PY_LWIP
 
@@ -40,9 +41,19 @@
 #include "lwip/timeouts.h"
 #include "lwip/dns.h"
 #include "lwip/dhcp.h"
+#include "lwip/nd6.h"
+#include "lwip/dhcp6.h"
+#include "lwip/prot/dhcp.h"
+#include "lwip/prot/dhcp6.h"
+#include <string.h>
+#include <stdlib.h>
+
+int mp_mod_network_prefer_dns_use_ip_version = 4;
 
 // Implementations of network methods that can be used by any interface.
 
+// This function provides the implementation of nic.ifconfig, is deprecated and will be removed.
+// Use network.ipconfig and nic.ipconfig instead.
 mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_obj_t *args) {
     if (n_args == 0) {
         // Get IP addresses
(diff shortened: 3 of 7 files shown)


dpgeorge:
> `ipconfig(autoconf6=True)` to enable ipv6 network configuration with slaac (and wait up to 10s for an address to be received)

This setting (and `dhcp4`) needs to be non-blocking, ie asynchronous.  Eg you may want to update this setting during the middle of an `asyncio`-based program, and you don't want it to block other running tasks.

So I guess configuring `autoconf6` would kick off the setting and then return immediately, and the user would need some way to tell/poll for when it's completed.

projectgus:
(I made a small edit on the PR description to clarify this is a per-NIC instance method.)

dpgeorge:
In https://github.com/micropython/micropython/issues/12600#issuecomment-1762433838 you distinguished between NIC vs global settings, eg `prefer` can be a global network setting.

Does it make sense to keep this separation and put `prefer` (and others) in `network.ipconfig`?

jimmo:
Thanks for the PR @felixdoerre !

It's going to be a bit confusing having both an `nic.ipconfig` and `nic.ifconfig` method (even if one is slated for deprecation), but from discussion with @dpgeorge and @projectgus I think maybe a good way forward is to just continue to use the existing `ifconfig`, and add `ipconfig` as an alias. (I agree, `ipconfig` is definitely the better name).

Fortunately the existing `ifconfig` can be made to work with the new behavior:
* No arguments: Existing behavior, return v4 `(addr,mask,gw,dns)` tuple.
* Single positional argument, `args[0] == "dhcp"`: Existing behavior, enable v4 dhcp.
* Any other positional argument: New behavior, return config option.
* Any keyword arguments: New behavior, set config option(s).

So the change would be:
* Rename `mod_network_nic_ifconfig` to `mod_network_lwip_ipconfig`
* Add new argument handling to this function.
* In each of the `network_foo.c`, rename the `ifconfig` function and wrapper object and locals dict entry to `ipconfig`
* Add an `ifconfig` alias to each of the locals dicts.

We discussed the relative merits of the way the `config()` methods work in MicroPython (see also BLE.config), it's a bit of an odd API, and it would be nice to explore how to improve it. One backwards-compatible option is to make the nic type implement `attr`, which for the `ipconfig` attribute would return an instance of a new type `LwipIpConfig` which is callable, but also itself implements `attr` allowing direct get/set access to the various options.  (Historically a type couldn't support having both a `locals_dict` and `attr`, which I think may have influenced this design originally, but this has been supported for a couple of years now)

So, the existing cases would invoke the call operator on the config object, allowing:

```py
print(nic.ipconfig("foo"))   # nic->attr->lwipipconfig->call("foo")
nic.ipconfig(foo=1)
```

but now you'd also be able to write

```
print(nic.ipconfig.foo)   # nic->attr->lwipipconfig->attr->foo
nic.ipconfig.foo = 1
```

this also lets us retain the `nic.ipconfig(**config_dict)` behavior we currently have.

Let's do that part in a future PR though :)

Some other comments:
* Do we need a way to get/set the v6 gateway? If you have slaac and dhcp6 disabled (weird, but allowed) then how does this work?
* Should we call it `slaac6` or `autoconf6`? The feature is called SLAAC right, but I guess it's generally known as auto configuration. I think `slaac6`.
* Can you clarify (maybe add some comments) the way setting `addr6` works? My understanding is that it removes any existing static address, and then sets the first free address slot to the new static address?
* When you query `addr4` you get an `("IP", "MASK")` tuple , but it appears that setting is expected to be done as "IP/PREFIXLEN". I feel like these should be consistent, although I can see why you've done it this way. I'm not sure what the right answer is...
* As Damien mentioned above, `dns` (and I think also `prefer`) is a global option. Should we introduce a top-level `network.ipconfig` (which for now only supports those two), or should we make all nics just provide a view onto the same global state?
* Should we support setting/getting multiple DNS servers? LWIP seems to use the zeroth DNS server for everything (i.e. when receiving a DHCP response), so I think just a single DNS server is probably fine, although hard to fix the API in the future.
* I agree with Angus' suggestion above about the default interface, although this isn't specific to v6 and this PR. It's a good feature to add though, but happy to save for another PR. (Right now for example if you enable both AP and station, you must be careful which order you bring them up in otherwise the AP will be the default interface).
* Any other considerations with respect to the non-LWIP builds... Right now because `mod_network_find_nic` isn't implemented the support for multiple interfaces isn't very complete, but for example, in this mode DNS/resolving is handled per-interface.

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                };
+                return mp_obj_new_tuple(2, tuple);
+            }
+            #if LWIP_IPV6
+            case MP_QSTR_addr6: {
+                mp_obj_t addrs[LWIP_IPV6_NUM_ADDRESSES];
+                size_t n_addrs = 0;
+                for (int i = 0; i < LWIP_IPV6_NUM_ADDRESSES; i++) {
+                    if (ip6_addr_isvalid(netif_ip6_addr_state(netif, i))) {
+                        const char *addr_str = ipaddr_ntoa(netif_ip_addr6(netif, i));
```
jimmo, line 149: Note that `ipaddr_ntoa` writes to a global static buffer... This is safe here because we immediately `mp_obj_new_str` it before calling again (but means this function is not thread safe).

If we use `ipaddr_ntoa_r` instead (with a stack-allocated buffer) then (assuming nothing else calls `ipaddr_ntoa`) we can save 56 bytes of bss by having the linker GC those functions.

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                        mp_obj_t tuple[4] = {
+                            mp_obj_new_str(addr_str, strlen(addr_str)),
+                            MP_OBJ_NEW_SMALL_INT(netif_ip6_addr_state(netif, i)),
```
jimmo, line 152: One of the API quirks elsewhere in micropython networking is not standardising the constant values across ports/implementations. (e.g. wireless security modes, or error codes).

So if e.g. we need to implement this on a non-LWIP NIC, we should implement a static mapping to defined constant values.

But in this case, are these values actually bool?

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                            mp_raise_ValueError(MP_ERROR_TEXT("invalid prefer argument"));
+                        }
+                        mp_mod_network_prefered_dns_ip = value;
+                        break;
+                    }
+                    #endif
+                    case MP_QSTR_addr4:
+                    case MP_QSTR_addr6: {
+                        ip_addr_t ip_addr;
+                        int prefix = 32;
```
jimmo, line 266: Please rename to `prefix_bits` or `prefix_len` to convey that it's the `/N`.

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                        if (e->value != mp_const_none) {
+                            size_t addr_len;
+                            const char *addr_str = mp_obj_str_get_data(e->value, &addr_len);
+                            char *split = strchr(addr_str, '/');
+                            if (split) {
+                                *split = 0;
```
jimmo, line 272: Unfortunately this is mutating the original string in `e->value`, as `mp_obj_str_get_data` does not copy the data. (And will outright fail if the string is in ROM, e.g. frozen str or qstr).

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                            if (e->value != mp_const_none) {
+                                netif->ip_addr = ip_addr;
+                                uint32_t mask = -(1u << (32 - prefix));
+                                ip_addr_set_ip4_u32_val(netif->netmask, ((mask & 0xFF) << 24) | ((mask & 0xFF00) << 8) | ((mask >> 8) & 0xFF00) | ((mask >> 24) & 0xFF));
+                            } else {
+                                ip4_addr_set_any(ip_2_ip4(&netif->ip_addr));
+                            }
+                        #if LWIP_IPV6
+                        } else if (mp_obj_str_get_qstr(args[0]) == MP_QSTR_addr6) {
+                            for (i = 1; i < LWIP_IPV6_NUM_ADDRESSES; i++) {
```
jimmo, line 294: Add comment here as mentioned in the main comment.

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                    case MP_QSTR_dhcp6: {
+                        dhcp6_disable(netif);
+                        dhcp6_enable_stateless(netif);
+                        break;
+                    }
+                    #endif
+                    #if LWIP_IPV6_AUTOCONFIG
+                    case MP_QSTR_autoconf6: {
+                        netif_set_ip6_autoconfig_enabled(netif, e->value == mp_const_true);
+                        if (e->value == mp_const_true) {
```
jimmo, line 211: Please use `mp_obj_is_true(e->value)` (here and the line above)

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                            nd6_restart_netif(netif);
+                            uint32_t start = mp_hal_ticks_ms();
+                            while (1) {
+                                int found = 0;
```
jimmo, line 215: `bool` instead of `int` (and `while (true)` above for consistency)

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                                for (i = 1; i < LWIP_IPV6_NUM_ADDRESSES; i++) {
```
jimmo, line 216: Please add a comment for this loop... I think it's "See if we now have a valid non-static address on this interface".

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                                        break;
+                                    }
+                                }
+                                if (found) {
+                                    break;
+                                }
+                                if (mp_hal_ticks_ms() - start > 10000) {
+                                    mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("timeout waiting for DHCP to get IP address"));
+                                }
+                                mp_hal_delay_ms(100);
```
jimmo, line 229: In general rather than delaying for a fixed time, just do `MICROPY_EVENT_POLL_HOOK`

(In more detail, that's exactly what `mp_hal_delay_ms` does anyway, and it means you don't end up waiting longer than necessary)

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                                    if (ip6_addr_isvalid(netif_ip6_addr_state(netif, i)) &&
+                                        !netif_ip6_addr_isstatic(netif, i)) {
+                                        found = 1;
+                                        break;
+                                    }
+                                }
+                                if (found) {
+                                    break;
+                                }
+                                if (mp_hal_ticks_ms() - start > 10000) {
```
jimmo, line 226: Please make `10000` a constant at the top of the file.

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                                    mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("timeout waiting for DHCP to get IP address"));
+                                }
+                                mp_hal_delay_ms(100);
+                            }
+                        } else {
+                            // Skip link-local address in slot 0
+                            for (i = 1; i < LWIP_IPV6_NUM_ADDRESSES; i++) {
```
jimmo, line 233: Similarly comment to explain "Clear out any non-static addresses, skipping the link-local address in slot 0".

(That said, what if slaac is disabled, does it still reserve slot 0?)

extmod/network_lwip.c
jimmo: Also I see the comment later explains why this loop starts at 1, might be worth duplicating here (wasn't obvious to me).

extmod/modlwip.c
```diff
@@ -1711,6 +1711,10 @@ STATIC void lwip_getaddrinfo_cb(const char *name, const ip_addr_t *ipaddr, void
     }
 }
 
+#if LWIP_VERSION_MAJOR >= 2
+extern int mp_mod_network_prefered_dns_ip;
```
jimmo, line 1715: I misunderstood this initially to be an ip address of a DNS server. Please rename to `mp_mod_network_prefer_dns_use_ip_version`

extmod/network_lwip.c
```diff
@@ -26,7 +26,7 @@
 
 #include "py/runtime.h"
 #include "py/mphal.h"
-
+#include "py/parsenum.h"
 #if MICROPY_PY_NETWORK && MICROPY_PY_LWIP
```
jimmo, line 30: Please add extra line before the `#if`.

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+    if (kwargs->used == 0) {
+        // Get config value
+        if (n_args != 1) {
+            mp_raise_TypeError(MP_ERROR_TEXT("must query one param"));
+        }
+
+        switch (mp_obj_str_get_qstr(args[0])) {
+            case MP_QSTR_dhcp4: {
+                struct dhcp *dhcp = netif_dhcp_data(netif);
+                return dhcp != NULL && dhcp->state != DHCP_STATE_OFF ? mp_const_true : mp_const_false;
```
jimmo, line 114: These can use `mp_obj_new_bool`

extmod/network_lwip.c
```diff
@@ -90,6 +100,241 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                return dhcp != NULL && dhcp->state != DHCP6_STATE_OFF ? mp_const_true : mp_const_false;
+            }
+            #endif
+            #if LWIP_IPV6_AUTOCONFIG
+            case MP_QSTR_autoconf6: {
+                return netif->ip6_autoconfig_enabled ?  mp_const_true : mp_const_false;
+            }
+            #endif
+            case MP_QSTR_dns: {
+                const char *addr_str = ipaddr_ntoa(dns_getserver(0));
```
jimmo, line 128: See comment later about `ipaddr_ntoa` (would be good to use `ipaddr_ntoa_r` instead).

extmod/network_lwip.c
felixdoerre: `addr_state` is a flags variable, containing multiple bits. They are defined in `ip6_addr.h:283` following. They contain information if an address is valid, tentative, duplicated, and if it is preferred.

The other values `pref_life`/`valid_life` are numeric timers that are decreasing over timed.

felixdoerre:
First of all, thank you very much for your detailed responses and review comments. I've resolved most comments, except the one on the "wait"-logic for dhcp4 and autoconf6, as it does not seem clear if we want to keep it. And the concerns regarding compatibility. Regarding the questions posed:

> This setting (and dhcp4) needs to be non-blocking, ie asynchronous. Eg you may want to update this setting during the middle of an asyncio-based program, and you don't want it to block other running tasks.

I made it blocking, to mimic the current "ifconfig" behavior. Both "autoconf6" and "dhcp4", can easily be made non-blocking, and checking if they completed can be done, like in C-code from python, by checking if corresponding IP-addresses are there. If that's wanted, I'll remove the blocking logic.

> Does it make sense to keep this separation and put prefer (and others) in network.ipconfig?
> As Damien mentioned above, dns (and I think also prefer) is a global option. Should we introduce a top-level network.ipconfig (which for now only supports those two), or should we make all nics just provide a view onto the same global state?

Yes, I'll just skipped it out of lazyness. I've added the separation.

> Do we need a way to get/set the v6 gateway? If you have slaac and dhcp6 disabled (weird, but allowed) then how does this work?

The gateway is set with a router advertisement. That is independent of slaac and of dhcp6 and cannot be disabled. (see nd6.c:569). You can only disable the automatic address derivation.

> Should we call it slaac6 or autoconf6? The feature is called SLAAC right, but I guess it's generally known as auto configuration. I think slaac6.

I think autoconf6 is the better name. It's called this was in lwip, and it is symmetrical to "autoconf4" (which we could want to implement/enable later). In linux the feature is also called `autoconf` (`/proc/sys/net/ipv6/conf/.../autoconf`).

> Can you clarify (maybe add some comments) the way setting addr6 works? My understanding is that it removes any existing static address, and then sets the first free address slot to the new static address?
> When you query addr4 you get an ("IP", "MASK") tuple , but it appears that setting is expected to be done as "IP/PREFIXLEN". I feel like these should be consistent, although I can see why you've done it this way. I'm not sure what the right answer is...

> Should we support setting/getting multiple DNS servers? LWIP seems to use the zeroth DNS server for everything (i.e. when receiving a DHCP response), so I think just a single DNS server is probably fine, although hard to fix the API in the future.

I'd leave that out for the future. I am not completely sure how one would extend the api, but one option would be to have `dns2`. Or to allow lists/tuples for `dns` arguments/values.

> I agree with Angus' suggestion above about the default interface, although this isn't specific to v6 and this PR. It's a good feature to add though, but happy to save for another PR. (Right now for example if you enable both AP and station, you must be careful which order you bring them up in otherwise the AP will be the default interface).

Yes, we can implement this in a future PR in the "global" function.

> Any other considerations with respect to the non-LWIP builds... Right now because mod_network_find_nic isn't implemented the support for multiple interfaces isn't very complete, but for example, in this mode DNS/resolving is handled per-interface.

I'm not sure how non-LWIP builds handle network.

Regarding compatibility: I'd want to have a clear path for deprecation and removal of `ifconfig` and the corresponding logic, even if it is, for example, an optionally installable python module to mimic the original behavior. I can merge the code, to have the new `ipconfig` logic call the old function, if the arguments are as it expects, and have `ipconfig` be an alias then, but I'd say that makes deprecation (and a drop-in module when we remove it eventually) even harder, so I am not sure, if an alias in either direction is helpful here. 

So regarding the 2 points I'd suggest this:
*Blocking* I'd suggest to just remove the blocking logic, and requiring users to wait in python code. I'm not that sure, if that is good, as the "typical usecase" for enabling dhcp pretty likely includes waiting for the ip.

*Compatibility* I'd suggest to keep the both methods separate and drop `ifconfig` in a couple of major releases, providing a python drop-in replacement for it, to install via `mip`.

projectgus:

greezybacon:
Could we please also add a `blocking=TRUE|false` kw option to make the networking API more compatible with async code?It would already be possible to monitor the progress of DHCP and autoconfig by polling the corresponding `addr4` or `addr6` endpoints. Having the code block for several seconds every time an address (auto) configuration is attempted is really counterproductive for asyncio-based workloads.

felixdoerre:
Yes, sure. Keeping blocking without being able to deactivate is not good, I think no-one disputed that. I'm hesitant to add a `blocking=true/false` option, as this is not another network-config option, but rather modifies existing behavior. Also it is not order-independent like all the other config options that we are using here. 

Thinking about it again, I see two options:
Option 1: make the call only non-blocking.
- Pro: clean api
- Con: blocking for dhcp requires manual polling

Option 2: Make the `dhcp4`/`autoconf6` options "tri-valued". For example with `True/False/"nonblock"`. Deactivating is non-blocking anyways.
- Pro: blocking is included in the API and is the default behavior making it easier to use for beginners
- Con: Having a 3rd value for the Option, that can never be returned, makes this option "special", but it is also in-line with the "special" `dhcp`-Option the previous method had, so this is not a big weirdness. Also people using async, are usually more advanced, so they can figure out, how to call the api, if it is not wrapped in a library.

What do you all think about "Option 2"? If we all agree, I'd go forward with that.

extmod/network_lwip.c
```diff
@@ -90,6 +99,279 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                                }
+                            }
+                        }
+                        break;
+                    }
+                    #endif
+                    case MP_QSTR_addr4:
+                    case MP_QSTR_addr6: {
+                        ip_addr_t ip_addr;
+                        int prefix_bits = 32;
```
jimmo, line 296: Following up to the earlier discussion about `"ip/N"` vs `("ip", "mask")`, I think we should make this function accept both.

i.e. if `e->value` is str, then use what you've implemented below, else use `mp_obj_get_array_fixed_n` and extract the ip and mask.

The rationale is that the setter should accept the same type as the getter, and I think that the tuple makes sense for the getter.

extmod/network_lwip.c
```diff
@@ -90,6 +99,279 @@ mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_o
...
+                switch (mp_obj_str_get_qstr(e->key)) {
+                    case MP_QSTR_dhcp4: {
+                        if (mp_obj_is_true(e->value)) {
+                            if (dhcp_supplied_address(netif)) {
+                                dhcp_renew(netif);
+                            } else {
+                                dhcp_release_and_stop(netif);
+                                dhcp_start(netif);
+                            }
+                            uint32_t start = mp_hal_ticks_ms();
```
jimmo, line 239: From talking to @dpgeorge and @projectgus about this:

The only other thing in network/nic config that blocks right now is explicitly doing `nic.ifconfig("dhcp")`. Note that the initial dhcp fetch doesn't block. Seeing as we're deprecating `ifconfig`, I think we should just declare that the `network` module is entirely non-blocking.

This is exactly your proposed "Option 1". However, we need to be able from Python to poll for an autoconf/dhcp address, i.e. an equivalent of `dhcp_supplied_address` or `!netif_ip6_addr_isstatic`. I think that means we need to add another field to the return tuple of `addr4` and `addr6` ? Or other suggestions welcome... `nic.ipconfig("has_autoconf6")` / `nic.ipconfig("has_dhcp4")` ? Not sure...

In the future we should consider adding callbacks (or a `irq` method, similar to the machine APIs) -- this is a long-requested feature.

jimmo (commented):
Thanks @felixdoerre for updating!

Happy to go with the new `ipconfig` being separate to the now-deprecated `ifconfig`, and also using `autoconf6` as the name.

Now that the API is agreed on, are you able to write some initial documentation for this new method? (Or let us know and we can do this in a follow-up PR). Thanks

felixdoerre (commented):
Hi, thank you for the review.

Regarding adding documentation, I'd like to do that in a separate PR, as `network.ifconfig` is not only explicitly documented, but referenced throughout the documentation (in 12+ files), including the documentation of specific NICs.

That brought my thought to the implementation of non-cyw43-nics. Wiznet5k and esphosted seem to have an LWIP-compatible interface so I added the corresponding (conversion) method. Wiznet5 in the non-lwip mode does not have such an interface, and for `ninaw10` there is no comparable interface at all (also the `dhcp` special case is not handled). So I'd be inclined to leave that `ipconfig` method untouched for those variants.

extmod/network_lwip.c
dpgeorge: For the `pref_life`/`valid_life` values, I think it's fine to leave them as pure numbers, representing a count.  So they don't need constants.

For `addr_state`, that could have different values when/if `ipconfig()` is implemented on non-lwIP NICs.  So we either need to:
- provide constants in the network module itself
- standardise the bit fields/values of `addr_state` and enforce them to be the same for all implementations of all NICs, and then document these bits/values

I think the latter option is the best.  And then for the "standard" bits/values that we choose, it can be the same as what lwIP exposes.  That means that for now the code here can stay as-is, but if lwIP changes these values we will need to provide a translation mapping the new back to the old, so users can always rely on the same values working.  And then we need to document these lwIP values as our standard.

extmod/network_cyw43.c
```diff
@@ -321,6 +321,13 @@ STATIC mp_obj_t network_cyw43_ifconfig(size_t n_args, const mp_obj_t *args) {
 }
 MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(network_cyw43_ifconfig_obj, 1, 2, network_cyw43_ifconfig);
 
+mp_obj_t network_nic_config(struct netif *netif, size_t n_args, const mp_obj_t *args, mp_map_t *kwargs);
```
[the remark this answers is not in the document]
jimmo: > So you would call even the method that is used to implement nic.ipconfig "static"?

No, this cannot be `static` because it's in a different compilation unit. I agree the convention is a bit loose (but I think it's at least consistent across all of the network code) is that the `mod_` prefix is used only on non-static methods.

So the convention as it stands is:

* Generally, in modfoo.c, the methods should be named `foo_bar`, and the wrapper objects named `foo_bar_obj`.
* Where possible, both the methods and the wrappers should be `static`.
* Where these functions need to call into hooks provided (e.g. by the port or elsewhere), the name should be something like `mp_foo_`.

There is remnants of an older convention (modjson.c, modbinascii.c, modbtree.c) where `mod_` was used as a prefix for methods. In general, almost all of these are `static`, because typically most modules are self-contained.

However, the two conventions seem to have blended in the network code, and the current situation is that the `mod_network_` prefix is used on non-static methods (i.e. because they are required to be implemented elsewhere), and just `network_` on static methods (the default).

Anyway, in this case please just match the existing way that `mod_network_nic_ifconfig` works (non-static, `mod_` prefix).

> Concluding, I will name the static method mod_network_ipconfig and the non-static method network_nic_ipconfig

Both of these methods are non-static, and they both work in the same way.

If I were to overhaul this convention, the main thing I would fix here is that both of these methods (and the existing `mod_network_nic_ifconfig` should have `lwip` somewhere in their name. In other works, `network_lwip.c` is a collection of helper methods that can be used by lwip-based nics. (And the especially confusing thing here is that `network_lwip.c` is not like, say, `network_cyw43.c`, rather it provides helper methods for e.g. `network_cyw43.c`)

> These are both static methods, one is prefixed with mod this other isn't. (gathering from the TODO, in L125), this probably occurred during refactoring.

`network_country` _is_ `static`, but `mod_network_hostname` is non-static. Nothing else calls `network_country` other than via Python, whereas `mod_network_hostname` is used in a bunch of places around the codebase.

Perhaps the confusion here (and what that TODO on L125) is about, is that the _wrapper_ object (the actual Python function object) is non-static. This allows a locals dict elsewhere to use the wrapper directly.

Also possibly we're talking at cross-purposes about the meaning of `static`. I am exclusively using it in the sense of the C keyword (i.e. private to this compilation unit). The fact that `network.ipconfig` is "static" in sort of the Python/C++ sense (versus an instance method like nic.ipconfig) has no bearing on the "mod_" prefix.

dpgeorge:
Merged!  Thanks @felixdoerre for the effort on this one.

PR #13691, opened 2024-02-16
fix websocket clean shutdown

felixdoerre:
When the websocket closes currently, it does not send a proper "close"-frame but encodes the `0x8800`-sequence inside a binary packet, which is wrong. The close packet is a different kind of websocket frame. This change resolves the error I get in Firefox when the websocket closes (and the change also seems logical to me, according to the websocket standard)


dpgeorge:
Thanks, this looks correct according to https://www.rfc-editor.org/rfc/rfc6455 section 5.5.1.

PR #13703, opened 2024-02-19
py/emitnative: Simplify how the prelude is found for native functions, and support `__name__' on them (v2)

dpgeorge:
Alternative to #13686.  This version is more involved but also adds support for `__name__` on viper functions.


dpgeorge:
This is too complicated.  Closing.

PR #13717, opened 2024-02-21
stm32/dma,spi: Fix invalid SPI reads due to D-Cache coherency.

### LINKED ISSUE #13471: STM32: SPI.read(n) fails if n > 1 (regression)

projectgus:
Adds a new DMA API to protect against cache coherency issues on chips with D-Cache, when working with arbitrary buffers (i.e. supplied by Python code).

The API is added to SPI, fixes #13471. Future work may be needed to call this API from other STM32 drivers that use DMA with arbitrary buffers.

Explanation:

1) It's necessary to invalidate D-Cache after a DMA RX operation completes, in case the CPU read (or speculatively accessed) memory in the DMA region during the operation. This seems to have been the root cause of #13471 (and only when src==dest, for that case.)

2) More generally, it is also necessary to temporarily mark the first and last cache lines of a DMA RX operation as "uncached", in case the DMA buffer shares this cache line with unrelated data. The CPU could otherwise write the other data at any time during the DMA operation (for example from an interrupt handler), creating a dirty cache line that's inconsistent with the DMA result.

## Test code

Regression test included in the PR, based on the work done by @peterhinch and @robert-hh in the linked issue.


projectgus:
@dpgeorge @jimmo Is there any reasonable place to include those tests in the code base? I saw there is a `tests/ports/stm32/spi.py` but it doesn't rely on any hardware configuration. These tests require the MISO pin to be connected to GND.

dpgeorge:
> Is there any reasonable place to include those tests in the code base? I saw there is a `tests/ports/stm32/spi.py` but it doesn't rely on any hardware configuration. These tests require the MISO pin to be connected to GND.

We should definitely include those tests in the repo, and `tests/ports/stm32/` is the best location for them.

I suggest making these tests separate file(s) to the existing tests, adding a comment at the top of the file of the require hardware connections, and then possibly trying to auto-detect those hardware connections (eg toggle IO and see if it changes) and SKIP'ing if they are not there.

Alternatives:
- don't do any auto-detection of the hardware connection and let the test fail if it's not set up properly, then the user is made aware that they must make hardware connections
- make a new `tests/ports/stm32_hardware` directory for these kinds of tests, so they can be easily included/excluded from a test run

Note there are already some tests in `stm32` that require specific hardware, like a PYBv1 with accelerometer.  Also the `tests/ports/cc3200` directory has similar hardware-dependent tests.

projectgus:
Have added regression test in `ports/stm32_hardware`. Would have preferred auto-skip idea, but couldn't think of a good way to programmatically find the GPIO for SPI2 MOSI to toggle it for a given SoC.

dpgeorge:
I tested a PYBD_SF2 with the test in this PR, both before and after the DMA/SPI fix in this PR.  The test fails without the fix and passes with the fix.  Very good!

Issue #8208, opened 2022-01-25
RP2: Overclock or CPU_MHZ as a board config option


Gadgetoid:
For PicoSystem we shipped a port of MicroPython that clocks the RP2040 to 250MHz at 1.2v (up from the stock 133MHz 1.1v settings) and have (short of needing to apply the slight overvolt in response to some units not being stable) generally seen these settings to be stable and safe. I accomplished this by maintaining a fork full of hacks where I've modified `main.c` :grimacing: 

Many RP2040-based projects leveraging MicroPython could probably benefit from this same overclock, particularly if the manufacturers are happy to support boards at these speeds.

As such I'd like to propose a board config approach to setting CPU frequency and overvolt, with a view to supporting this in the RP2 port at first and perhaps expanding to other boards (we've encountered MCUs that are ostensibly one speed, but get a manufacturer certified OC frequency later).

This is part of a multi-step process to untie all the hacks I needed to implement for PicoSystem, justify their existence and bring them upstream for the benefit of everyone.

It also seems a shame to leave that 2x+ speedup on the table.

1. What settings are common across MCUs for achieving overclock? I guess CPU_FREQ and VOLTAGE perhaps, but the latter could be derived from the former in most cases
2. Is there anywhere in the MicroPython codebase where assumptions are made about CPU frequency that this would horribly break?
3. Would it be appropriate to configure an "Overclocked Pico" board variant, for those who want to opt in to more performance?

Gadgetoid:
Looks like this is tangentially related: https://github.com/micropython/micropython/pull/6941

I think there's still a good case for an overclock-by-default option on a board by board basis.

dpgeorge:
> This is part of a multi-step process to untie all the hacks I needed to implement for PicoSystem, justify their existence and bring them upstream for the benefit of everyone.

A great goal!

> As such I'd like to propose a board config approach to setting CPU frequency and overvolt, with a view to supporting this in the RP2 port at first and perhaps expanding to other boards

This sounds fine to me.

There is the existing `machine.freq()` function which can be used to change the CPU frequency, and that allows overclocking on ports/MCUs that support it (eg stm32, rp2).  So the user can dynamically opt-in to overclocking.

> 1. What settings are common across MCUs for achieving overclock?

I think this is very hardware dependent.  Apart from providing a common `machine.freq()` function (which already exists) I don't think it's worth trying to define a common way to build firmware with overclocking enabled.

[...]

PR #13718, opened 2024-02-21
rp2/main.c: Set the default clock frequency at boot.

robert-hh:
As a side effect, the peripheral clock will be set to 48Mhz and both UART and I2C baud rates will not be affected any more by CPU speed changes using machine.freq().

With the change the UART baud rate range is 50 to 3_000_000. That is a somewhat breaking change since baud rates beyond 3MBaud are not supported any more. 

Addresses the discussion #13679.

### DIFF
--- a/ports/rp2/main.c
+++ b/ports/rp2/main.c
@@ -80,6 +80,9 @@ int main(int argc, char **argv) {
     pendsv_init();
     soft_timer_init();
 
+    // Set the MCU frequency and as a side effect the peripheral clock to 48 MHz.
+    set_sys_clock_khz(125000, false);
+
     #if MICROPY_HW_ENABLE_UART_REPL
     bi_decl(bi_program_feature("UART REPL"))
     setup_default_uart();
@@ -231,6 +234,10 @@ int main(int argc, char **argv) {
 
         gc_sweep_all();
         mp_deinit();
+        #if MICROPY_HW_ENABLE_UART_REPL
+        setup_default_uart();
+        mp_uart_init();
+        #endif
     }
 
     return 0;
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -96,6 +96,20 @@ static void mp_machine_set_freq(size_t n_args, const mp_obj_t *args) {
     if (!set_sys_clock_khz(freq / 1000, false)) {
         mp_raise_ValueError(MP_ERROR_TEXT("cannot change frequency"));
     }
+    if (n_args > 1) {
+        mp_int_t freq_peri = mp_obj_get_int(args[1]);
+        if (freq_peri != (USB_CLK_KHZ * KHZ)) {
+            if (freq_peri == freq) {
+                clock_configure(clk_peri,
+                    0,
+                    CLOCKS_CLK_PERI_CTRL_AUXSRC_VALUE_CLKSRC_PLL_SYS,
+                    freq,
+                    freq);
+            } else {
+                mp_raise_ValueError(MP_ERROR_TEXT("peripheral freq must be 48_000_000 or the same as the MCU freq"));
+            }
+        }
+    }
     #if MICROPY_HW_ENABLE_UART_REPL
     setup_default_uart();
     mp_uart_init();
--- a/docs/rp2/quickref.rst
+++ b/docs/rp2/quickref.rst
@@ -31,12 +31,25 @@ The MicroPython REPL is accessed via the USB serial port. Tab-completion is usef
 find out what methods an object has. Paste mode (ctrl-E) is useful to paste a
 large slab of Python code into the REPL.
 
-The :mod:`machine` module::
+The :mod:`machine` module:
+
+machine.freq() allows to change the MCU frequency and control the peripheral
+frequency for UART and SPI. Usage::
+
+    machine.freq(MCU_frequency[, peripheral_frequency=48_000_000])
+
+The MCU frequency can be set in a range from less than 48 MHz to about 250MHz.
+The default at boot time is 125 MHz. The peripheral frequency must be either
+48 MHz or identical to the MCU frequency, with 48 MHz as the default.
+If the peripheral frequency is changed, any already existing instance of
+UART and SPI will change it's baud rate and may have to be re-configured::
 
     import machine
 
     machine.freq()          # get the current frequency of the CPU
-    machine.freq(240000000) # set the CPU frequency to 240 MHz
+    machine.freq(240000000) # set the CPU frequency to 240 MHz and keep
+                            # the UART frequency at 48MHz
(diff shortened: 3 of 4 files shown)


ports/rp2/main.c
```diff
@@ -223,6 +227,12 @@ int main(int argc, char **argv) {
         soft_timer_deinit();
         gc_sweep_all();
         mp_deinit();
+        // Reset the MCU frequency
+        set_sys_clock_khz(125000, false);
```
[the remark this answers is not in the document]
robert-hh: The intention here is to have the device at the same state as when entering the `for (;;) {` loop. If code e.g. in boot.py and main.py changes the frequency, it will still do it.

Gadgetoid:
Could this possibly also include some accommodation for https://github.com/micropython/micropython/issues/8208

I'd very much like a board config option to change that default, please!

robert-hh:
@dpgeorge I dropped resetting the clock at soft reboot. Only at cold boot the clock it set to have the side effect of changing the peripheral frequency to 48MHz.

ports/rp2/modmachine.c
```diff
@@ -95,6 +95,12 @@ static void mp_machine_set_freq(size_t n_args, const mp_obj_t *args) {
     if (!set_sys_clock_khz(freq / 1000, false)) {
         mp_raise_ValueError(MP_ERROR_TEXT("cannot change frequency"));
     }
+    // Option: Set PERI clock to system clock again
+    // clock_configure(clk_peri,
+    //                 0,
+    //                 CLOCKS_CLK_PERI_CTRL_AUXSRC_VALUE_CLK_SYS,
+    //                 SYS_CLK_KHZ * KHZ,
+    //                 SYS_CLK_KHZ * KHZ);
```
dpgeorge, line 103: I'm not sure what this commented-out code is aiming to do.  When `set_sys_clock_khz()` is called, does that update the PERI clock or not?

robert-hh: The whole change is about the inconsistency, that when machine.freq() is called to change the frequency, then PERI clock is set to 48 MHz as result of calling `set_sys_clock_khz()`. The change I suggested would set PERI clock always to 48MHz. Then, changing the CPU frequency would not affect e.g. the UART or PWM settings, at the penalty of limiting the peripherals baud rates.
 
This commented code is about the alternative suggestion to set PERI clock always to the CPU clock. Then there would be changes to the baud rates of the peripherals when changing the CPU clock, but at known rates.

dpgeorge: OK, I understand more.

So without this PR (ie current master), what is the value of PERI at start up, is it the system clock, or system clock divided by some factor?

Note that `machine.freq()` takes multiple arguments to specify the frequency of the buses (see eg stm32 port).  So maybe it's a good idea to allow the user to specify the PERI clock as the second (optional) argument to `machine.freq()`?  Something like:
```py
machine.freq(sys, peri=48000000)
```

robert-hh: At the moment the PERI freq at boot is the system clock 125Mhz and it will be set to 48Mhz when calling machine.freq(n). 

> Note that machine.freq() takes multiple arguments to specify the frequency of the buses (see eg stm32 port).

Actually, the stm32 port lost this feature with commit 7d39db2.  machine.freq() takes just 1 argument.

robert-hh: I changed machine.freq() to have a second boolean type agrument. If set to False or omitted, the peripheral frequency will be 48MHz, which is the default value now at boot. If set to True, the peripheral frequency is the same as the MCU. The clock section for the peripheral clock has no divider. The documentation is wrong. So the peripheral clock can only be one of the other clock sources, like USB clock or SYS clock. 
The API could also be changed to accept a frequency instead of a boolean value, with 48_000_000 and the value of the first argument as the permitted values. But then, an error check & message would have to be added.
As a required side change, machine.freq() now accepts up to 4 arguments. I made a separate PR for that 1 char change as well.

dpgeorge: > The API could also be changed to accept a frequency instead of a boolean value, with 48_000_000 and the value of the first argument as the permitted values. But then, an error check & message would have to be added.

It looks like this is how you ended up implementing it (ie not as a boolean).  I think what you did -- specifying a freq -- is the right way to do it.  That way is more general, and more consistent with stm32.

But do you think it's better to make the peri frequency a keyword-only argument?  That forces the user to be explicit and makes it even more future proof/general.  Eg `machine.freq(125000000, peri=48000000)`.

robert-hh: > But do you think it's better to make the peri frequency a keyword-only argument? That forces the user to be explicit and makes it even more future proof/general. Eg machine.freq(125000000, peri=48000000).

The Python API is is extmod/modmachine.c. Making peri=xxx a keyword only argument would either require to change that or to move the API back to the rp2 port.

ports/rp2/modmachine.c
```diff
@@ -88,14 +88,32 @@ NORETURN void mp_machine_bootloader(size_t n_args, const mp_obj_t *args) {
 }
 
 static mp_obj_t mp_machine_get_freq(void) {
-    return MP_OBJ_NEW_SMALL_INT(mp_hal_get_cpu_freq());
+    mp_obj_t tuple[] = {
+        mp_obj_new_int(mp_hal_get_cpu_freq()),
+        mp_obj_new_int(clock_get_hz(clk_peri)),
+    };
+    return mp_obj_new_tuple(MP_ARRAY_SIZE(tuple), tuple);
```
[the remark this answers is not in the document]
robert-hh: Keeping the old behavior is easy. That's why it's a separate commit.

ports/rp2/modmachine.c
dpgeorge: > The Python API is is extmod/modmachine.c. Making peri=xxx a keyword only argument would either require to change that

Yes, if we want to make this a keyword-only argument, we'd need to change the API in `extmod/modmachine.c`.  We can do that if we think it's the best option.  All other ports can remain as they are (stm32 accepting 4 positional args, all other ports accepting 1 positional arg).

dpgeorge: OK, let's leave it as a positional argument.

ports/rp2/modmachine.c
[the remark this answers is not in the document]
dpgeorge: Thanks for removing the additional commit (and returning just a single int).

dpgeorge:
> Could this possibly also include some accommodation for #8208
> 
> I'd very much like a board config option to change that default, please!

@Gadgetoid that was not done in this PR (a config option for default CPU freq).  That should be pretty easy to do, but someone needs to do it :smile:

robert-hh:
Thank you.

PR #13725, opened 2024-02-22
extmod/modlwip: Allow empty ip addresses.

felixdoerre:
Previously, binding a socket with
```
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.bind(("", PORT))
```
was possible, but the empty string is not regarded as valid ip address by lwip. This PR adds a special case for the empty ip string, restoring cpython behavior.


dpgeorge:
Thanks, this looks fine.  esp32 already has such a check for the empty string, similar to this.

But please try to get the esp8266 CI passing.

dpgeorge:
Ah, I missed the point that this is fixing a regression introduced by 628abf8f25a7705a2810fffe2ca6ae652c532896

PR #13732, opened 2024-02-22
ports: Fix ports missing the vfs module.

robert-hh:
In PR #13584 the VFS module was not enabled for the SAMD21 port. This  is done now. For some reasons, the builds on the download page have VFS provided.

Note: The same should apply to all boards with the NRF52832 MCU. So it might be better to change the default rule in py/mpconfig.py to set the level limit for VFS to MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_BASIC_FEATURES.

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -1746,7 +1746,7 @@ typedef double mp_float_t;
 
 // Whether to provide the "vfs" module
 #ifndef MICROPY_PY_VFS
-#define MICROPY_PY_VFS (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_EXTRA_FEATURES && MICROPY_VFS)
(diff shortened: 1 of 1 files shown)


dpgeorge:
Good that you noticed this, I definitely missed it!



> So it might be better to change the default rule in py/mpconfig.py to set the level limit for VFS to MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_BASIC_FEATURES.

Yes, I think this is a better solution.  Actually, because it's also guarded by having `MICROPY_VFS` enabled (the module is useless without the VFS subsystem) then I suggest enabling it at the core feature level.  That will actually fix the B_L072Z_LRWAN1 board, which has `os` enabled and the VFS subsystem, but currently no `vfs` module.

robert-hh:
The PR is updated to at least core features. A test build for SAMD21 works.

dpgeorge:
Thanks for updating and testing.

Issue #11530, opened 2023-05-17
uasyncio: current_task behaviour differs from CPython


peterhinch:
Background: I am looking for a way detect whether a function has been called in a `uasyncio` context. The `asyncio_running` function works in CPython, but in MP it fails if a `uasyncio` script has previously run. In the absence of a fix, a non-hacky workround would be appreciated :)

Repro:
```py
try:
    import uasyncio as asyncio
except:
    import asyncio

def asyncio_running():
    try:
        _ = asyncio.current_task()
    except:
        return False
    return True

def prun(txt):
    print(f"{txt}: asyncio {'' if asyncio_running() else 'not'} running")

prun("Start")

async def foo():
    prun("In task")
    await asyncio.sleep(1)

try:
    asyncio.run(foo())
finally:
    _ = asyncio.new_event_loop()

prun("At end")
```
On CPython 3.8.10:
```
>>> import rats25
Start: asyncio not running
In task: asyncio  running
At end: asyncio not running
```
On MP 1.20 (tested on Unix and RP2):
```
>>> import rats25
Start: asyncio not running
In task: asyncio  running
At end: asyncio  running
```

dpgeorge:
Fixed by 8692d2602a7775fea22d75ebf5b50971dc2b968f

PR #11562, opened 2023-05-19
uasyncio/core.py: Make current_task raise when there is no task.

jimmo:
To match CPython it should raise RuntimeError if no loop is currently active. Previously it returned the most recent task that ran.

I compared doing it this way versus setting it to None, but then we need to add the extra line to declare `cur_task`.

It's +60 bytes, but only +24 bytes if you don't include the error message for the `RuntimeError`.

Fixes #11530  CC @peterhinch

PR #13752, opened 2024-02-26
extmod/asyncio: Make current_task raise exception when there is no task.

dpgeorge:
Matches CPython behaviour.

Fixes issue #11530.


dpgeorge:
Alternative to #11562.

Compared to that PR:
- the one here is +56 bytes on PYBv1.0, and #11562 is +68 (built locally with arm-none-eabi-gcc 13.2.0)
- the one here does not add any code to the hot path/loop of the asyncio scheduler so doesn't reduce performance of asyncio

jimmo:
LGTM (see https://github.com/micropython/micropython/pull/11562#issuecomment-1965685130)

PR #13755, opened 2024-02-26
ports/mimxrt:Add thread support.

tonyzhangnxp:
Adds thread support on ports/mimxrt, mimicks it from stm32.


tonyzhangnxp:
Hello Robert

Thanks for your time. Some customers requested the threading feature in micropython on MIMXRT platform.
It's a good idea that this feature can be merged into the main branch.

- I did not enable threading for the MIMXRT 10xx boards, since these MCUs are low on RAM.
_[Tony] Which one has low memory, do you mean MIMXRT1010 evk ? which has 128KB RAM, it's not enough for threading ?_ 
- You do not have to set/unset threading in each board's .mk file. You can do that in Makefile.
_[Tony] In my understanding, mk file is used to dis/enable the feature, so I added it for all the boards. If add it in Makefile, that means threading is default enable for all the boards?_

robert-hh:
> which has 128KB RAM, it's not enough for threading 

Of the 128k only 64 k are available for the OCRM section, which holds the MicroPython heap. The stack of threads is taken from the  heap, like all other dynamic data for Python scripts. 

>  If add it in Makefile, that means threading is default enable for all the boards?

You can enable it if not already defines otherwise in the board's .mk file. May be fewer files changed.

tonyzhangnxp:
hello
Set MICROPY_PY_THREAD = 1 if undefined in mpconfigport.h, and set MICROPY_PY_THREAD=0 in mk of mimxrt1010 platforms.

robert-hh:
Thanks, well noticed. The interesting information is, that there are customers using MicroPython. Do you know why they did not ask at the MicroPython site?

robert-hh:
The test coverage for tests/thread is fine. The same as Pyboard:

```
25 tests performed (139 individual testcases)
24 tests passed
5 tests skipped: mutate_bytearray mutate_dict mutate_instance mutate_list mutate_set
1 tests failed: thread_ident1
```

tonyzhangnxp:
> Thanks, well noticed. The interesting information is, that there are customers using MicroPython. Do you know why they did not ask at the MicroPython site?

As I know the customer is using MicroPython in some industrial products, and they need quick support, so the device vendor is the first option.
I thought maybe merging this feature in the main branch should be good for others.

projectgus:

tonyzhangnxp:
All STATIC had been replaced with static

PR #13764, opened 2024-02-27
stm32: Add support for analog-only "_C" pins.

dpgeorge:
This PR adds support for the analog-only pads/pins on STM32H7 parts.  These pads/pins are called PA0_C/PA1_C/PC2_C/PC3_C in the datasheet.  They each have an analog switch that can optionally connect them to their normal pin (eg PA0).  When the switch is open, the normal and _C pin are independent pins/pads.

The approach taken in this PR to make these _C pins available to Python is:
- put them in their own, independent row in the stm32h7_af.csv definition file, with only the ADC column defined (this seems logical to me, since they are going to be separate machine.Pin entities, and doing it this way keeps `make-pins.py` pretty clean)
- allow a board to reference these pins in the board's `pins.csv` file by the name `PA0_C` etc (so a board can alias them, for example)
- these pins (when enabled in `pins.csv`) now become available like any other `machine.Pin` through both `machine.Pin.board` and `machine.Pin.cpu`
- BUT these _C pins have a separate pin type which doesn't have any methods, because they don't have any functionality
- these _C pins can be used with `machine.ADC` to construct the appropriate ADC object, either by passing the string as `machine.ADC("PA0_C")` or by passing the object as `machine.ADC(machine.Pin.cpu.PA0_C)`
- if a board defines both the normal and _C pin (eg both PA0 and PA0_C) in `pins.csv` then it must not define the analog switch to be closed (this is a sanity check for the build, because it doesn't make sense to close the switch and have two separate pins)

To show how this works, the ARDUINO_PORTENTA_H7 board is updated to define its analog pins, using these new _C pins.  The code is also tested on this board to work correctly.

TODO:
- [x] update all relevant af.csv files with separate _C rows


dpgeorge:
@iabdalkader FYI, this should allow full support for the dual-pad analog pins on Arduino boards.

ports/stm32/boards/make-pins.py
```diff
@@ -59,6 +59,11 @@ class Stm32Pin(boardgen.Pin):
     def __init__(self, cpu_pin_name):
         super().__init__(cpu_pin_name)
 
+        # Pins ending in "_C" are special analog-only pins.
```
jimmo, line 62: Can you add "H7" and "Dual-pad" to this comment -- if you don't know about dual pad, it would be very difficult to figure out what an "analog-only pin" is.

ports/stm32/boards/stm32f4xx_prefix.c
```diff
@@ -30,3 +30,19 @@
...
+        .port = PORT_##p_port, \
+        .pin = (p_pin), \
+        .num_af = (sizeof(p_af) / sizeof(pin_af_obj_t)), \
+        .pin_mask = (1 << ((p_pin) & 0x0f)), \
+        .gpio = GPIO##p_port, \
+        .af = p_af, \
+        .adc_num = p_adc_num, \
+        .adc_channel = p_adc_channel, \
+    }
+#endif
```
jimmo, line 48: Could we add an `#else` here that did something like

```
#define PIN_ANALOG DUAL_PAD_SUPPORT_NOT_ENABLED__CONFIGURE__MICROPY_HW_ANALOG_SWITCH_PAx
```

ports/stm32/pin.c
```diff
@@ -601,6 +605,44 @@ MP_DEFINE_CONST_OBJ_TYPE(
...
+/******************************************************************************/
+// Special analog-only pin.
+
+#if MICROPY_HW_ENABLE_ANALOG_ONLY_PINS
+
+// When both normal and _C pins are defined in pins.csv, force the corresponding
+// analog switch to be open.  The macro code below will produce a compiler warning
+// if the board defines the switch as closed.
+#if defined(pin_A0) && defined(pin_A0_C)
+#define MICROPY_HW_ANALOG_SWITCH_PA0 (SYSCFG_SWITCH_PA0_OPEN)
```
jimmo, line 617: Is the idea that this triggers a redefinition of the macro?

Would doing this achieve the same:

```c
#if defined(pin_A0) && defined(pin_A0_C) && defined(MICROPY_HW_ANALOG_SWITCH_PA0) && MICROPY_HW_ANALOG_SWITCH_PA0 == SYSCFG_SWITCH_PA0_CLOSED
#error "Both pads for dual-pad A0 are enabled, switch must be closed."
#endif
```

(It's a bit confusing to see a .c file defining a macro that it doesn't use)

dpgeorge: As the comment says, this is intended to trigger a compiler warning if the board defines the switch as closed.

I originally had tried what you suggested above but it doesn't work because the comparison `MICROPY_HW_ANALOG_SWITCH_PA0 == SYSCFG_SWITCH_PA0_CLOSED` is too complex for the preprocessor.

ports/stm32/boards/make-pins.py
dpgeorge: Done.

ports/stm32/boards/stm32f4xx_prefix.c
dpgeorge: Done.

iabdalkader:
@dpgeorge I tested this by adding all analog/analog-only pins to the Giga board, and I can read values independently on dual pads pins, so it seems to be working fine, thanks!

```C
A0,PC4
A1,PC5
A2,PB0
A3,PB1
A4,PC3
A5,PC2
A6,PC0
A7,PA0
A8,PC2_C
A9,PC3_C
A10,PA1_C
A11,PA0_C
```

Test:
```Python
import time
from machine import ADC
from machine import Pin

adc1 = ADC("A5")  # PC2
adc2 = ADC("A8")  # PC2_C
while True:
    print(adc1.read_u16(), adc2.read_u16())
    time.sleep_ms(500)
```

dpgeorge:
Thanks @iabdalkader for testing.

I have now added the Giga `pins.csv` code as part of this PR.

Does anything need to be added for Nicla Vision?

iabdalkader:
The Nicla closes 2 switches, and breaks out 3 single pad pins analog pins::

```Python
A0,PC4
A1,PF13
A2,PF3
```

Pins for all boards can be found here in `<board>/variant.cpp` https://github.com/arduino/ArduinoCore-mbed/tree/main/variants

Should we also add digital board pins in this PR ?

dpgeorge:
> The Nicla closes 2 switches, and breaks out 3 single pad pins analog pins

OK, thanks, I added those analog pins.

> Should we also add digital board pins in this PR ?

I also added digital pins to this PR.  Can you please check them?  I didn't add everything for the Giga, just the standard set up to D14.

ports/stm32/boards/ARDUINO_NICLA_VISION/pins.csv
```diff
@@ -166,6 +166,20 @@ PK4,PK4
...
 PK6,PK6
 PK7,PK7
+
+# Arduino digital pins
+D0,PG12
+D1,PA9
+D2,PA10
+D3,PG1
+D4,PC7
+D5,PC6
```
iabdalkader, line 176:
```suggestion
```
Note those two pins are broken out on the camera connector. I don't think it makes much sense to define them, since the camera connector isn't a generic B2B connector, but if we do then we should define the other pins on that connector.

iabdalkader:
> Can you please check them? I didn't add everything for the Giga, just the standard set up to D14.

I double checked all pins, they all look good. No need to add the HD pins there's another open PR that adds them

dpgeorge:
Hmm, maybe I should merge #14022 first, then this PR here just adds the _C analog pins.

iabdalkader:
> Hmm, maybe I should merge #14022 first, then this PR here just adds the _C analog pins.

Either way is fine with me, but re #14022, in the stm32 port, it's conventional to define actual board pins near the end of the file, after all `Pxy,Pxy`

sebromero:
I just added the suggestions form the review in #14022  ✅

dpgeorge:
I have removed all pin additions to Arduino boards.  That will be done as part of #14022.

ports/stm32/boards/ARDUINO_NICLA_VISION/pins.csv
dpgeorge: These pins have not been added.

PR #13766, opened 2024-02-27
ports/rp2: Use MICROPY_BOARD_DIR to find pins.csv.

Gadgetoid:
It's not valid to assume that `${MICROPY_PORT_DIR}/boards/${MICROPY_BOARD}` is equal to `${MICROPY_BOARD_DIR}`, because the latter could be *outside* of `${MICROPY_PORT_DIR}`, eg: https://github.com/pimoroni/badger2040/tree/main/firmware/PIMORONI_BADGER2040W

This PR replaces this path with the canonical `${MICROPY_BOARD_DIR}` so that `pins.csv` is correctly located when building against out-of-tree board definitions.

Additionally I've removed `MICROPY_BOARDS_DIR` to discourage similar mistakes.

This bug caused `pins.csv` in our out-of-tree board definitions for Badger 2040W to not be detected- https://github.com/pimoroni/badger2040/issues/60

### DIFF
--- a/ports/rp2/CMakeLists.txt
+++ b/ports/rp2/CMakeLists.txt
@@ -519,15 +519,14 @@ endforeach()
 # Include the main MicroPython cmake rules.
 include(${MICROPY_DIR}/py/mkrules.cmake)
 
-set(MICROPY_BOARDS_DIR "${MICROPY_PORT_DIR}/boards")
-set(GEN_PINS_AF_CSV "${MICROPY_BOARDS_DIR}/rp2_af.csv")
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Is there anything I can/should test that would help move this along?

dpgeorge:
Sorry I didn't get to this earlier.  This is indeed a proper fix for the issue, and I also like the removal of `MICROPY_BOARDS_DIR`, because the name of that variable is too similar to `MICROPY_BOARD_DIR`.

Rebased and merged in 932f76c6ba64c5a3e68de3324556d9979f09303b

PR #13364, opened 2024-01-05
ESP32: Use ESP-IDF v5.0.5.

IhorNehrutsa:

PR #11869, opened 2023-06-26
esp32c6 support (IDF v5.2)

andrewleech:
The is esp32c6 officially supported in IDF v5.1 but requires a number of changes to work...
Pulling together changes and generic board support needed here.

PR #13775, opened 2024-02-28
esp32: Add support for IDF version v5.2.

DvdGiessen:
This MR adds support for building the ESP32 port with ESP-IDF v5.2. I've been using these for a little bit the past few weeks on three different ESP32 variants and it seems to work well so far. These changes are implemented in the second commit of this MR.

The first commit in this MR makes the build work with IDF v5.0.5 and later. Included here since all that needed was a small touch-up on the version ranges in existing version checks.

The new IDF v5.2 deprecated the task cleanup callback we use, so I implemented support for the new option. This also requires a change in the `sdkconfig`, for the third and last commit in this MR implements an new variable `${SDKCONFIG_IDF_VERSION_SPECIFIC}` which is used in all `mpconfigboard.cmake` files to include an extra `sdkconfig` file based on the IDF version in use.

EDIT: Added IDF-dependant `sdkconfig` file as suggested by @dpgeorge.


dpgeorge:
See related #11869 and #13364.

dpgeorge:
> But because I don't think we can conditionally configure these based on a detected IDF version it is not part of the patch.

This might be possible with some cmake code like this:
```cmake
if(IDF_VERSION_MINOR >= 2)
    set(SDKCONFIG_DEFAULTS
        ${SDKCONFIG_DEFAULTS}
        boards/sdkconfig.idf52
    )
else()
    set(SDKCONFIG_DEFAULTS
        ${SDKCONFIG_DEFAULTS}
        boards/sdkconfig.idf50
    )
endif()
```

DvdGiessen:
> This might be possible with some cmake code

Ah, yes, that does (with some changes) work nicely.

I ended up implementing it slightly different: Instead of adding always the extra configuration file in the CMake code, I opted to make a specific variable `${SDKCONFIG_IDF_VERSION_SPECIFIC}` that is used in all the `mpconfigboard.cmake` files, similar to how we also explicitly specify the `sdkconfig.base` file in every one of those.

Pushed an extra commit that implements this change.

I think that makes it more explicit and clear for developers which `sdkconfig` files are being used for their board, without also having to go through the `CMakeList.txt` file to see if any additional ones are being appended there. Let me know if you'd prefer the solution where we don't modify all board files.

Also updated MR description and added explicit commit that adds compatibility with v5.0.5 and v5.0.6, since that was just a matter of setting the correct version range in an existing check.

ports/esp32/modnetwork_globals.h
```diff
@@ -31,9 +31,13 @@
 { MP_ROM_QSTR(MP_QSTR_AUTH_WPA2_WPA3_PSK), MP_ROM_INT(WIFI_AUTH_WPA2_WPA3_PSK) },
 { MP_ROM_QSTR(MP_QSTR_AUTH_WAPI_PSK), MP_ROM_INT(WIFI_AUTH_WAPI_PSK) },
 { MP_ROM_QSTR(MP_QSTR_AUTH_OWE), MP_ROM_INT(WIFI_AUTH_OWE) },
-#if ESP_IDF_VERSION > ESP_IDF_VERSION_VAL(5, 1, 1)
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 5) || ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 1, 2)
```
dpgeorge, line 34: Won't the second condition be always true if the first is true?

kdschlosser:
I had thought that MIcroPython was no longer going to support more than a single version of the IDF at any given time. It would eliminate all of the IDF version checking that needs to be done.... what happened there?

andrewleech:
@DvdGiessen could you check if master still builds with v5.2.1 for you?

It's fine for S3 etc but `ESP32_GENERIC` build is failing with:
```
ld: region `iram0_0_seg' overflowed by 84
 bytes
ninja: build stopped: subcommand failed.
HINT: The applications static IRAM usage is larger than th
e available IRAM size.
For more information on how to reduze IRAM usage run 'idf.
py docs -sp api-guides/performance/ram-usage.html#optimizi
ng-iram-usage'
ninja failed with exit code 1, output of the command is in
 the /micropython/ports/esp32/build-ESP32_GENERIC/log/idf_
py_stderr_output_774 and /micropython/ports/esp32/build-ES
P32_GENERIC/log/idf_py_stdout_output_774
```

DvdGiessen:
> There is a bunch of things that have been deprecated between 5.1 and 5.2 and yet those things are still being used. I2C is one of the very large changes.. There are others as well.

Correct. I did see those changes and looked into how much work it would be to rework the current code to use those new interfaces, and estimated doing so would result in a lot more work (both for me in rewriting a lot of code, and for the maintainers needing to review a much larger PR). The old interfaces, although marked as deprecated, are not removed and work fine (they might not contain the improvements made in the newer interfaces, but it's usually not going to be worse than things already were by being on an older IDF version all together).

So while indeed it might indeed be nice to move to these newer interfaces, I concluded that shouldn't delay the smaller changes needed to run on the lastest IDF (which aside from these big interface changes also includes a lot of other stuff that is useful to have in the meantime).

> @DvdGiessen could you check if master still builds with v5.2.1 for you?

Hm, the ESP32_GENERIC does indeed seem to slightly overflow the static IRAM space:

```
Total sizes:
Used static DRAM:   55704 bytes (  68876 remain, 44.7% used)
      .data size:   23280 bytes
      .bss  size:   32424 bytes
Used static IRAM:  131154 bytes (    -82 remain, 100.1% used) Overflow detected! You can run idf.py size-files for more information.
      .text size:  130127 bytes
   .vectors size:    1027 bytes
Used Flash size : 1561271 bytes
           .text: 1285067 bytes
         .rodata:  275948 bytes
Total image size: 1715705 bytes (.bin may be padded larger)
```

PR #13785, opened 2024-02-29
extmod/os_dumpterm: Handle exception correctly when it happens on read and on write.

felixdoerre:
When disconnecting from the webrepl abruptly, micropython crashes. Inspection with the debugger reveals, that `dupterm_objs` is already `NULL` and so deactivation/closing segfaults. This happens because the `dupterm_obj` is already removed from the exception that was handled when trying to read from the socket.

### DIFF
--- a/extmod/os_dupterm.c
+++ b/extmod/os_dupterm.c
@@ -45,6 +45,10 @@ void mp_os_deactivate(size_t dupterm_idx, const char *msg, mp_obj_t exc) {
     if (exc != MP_OBJ_NULL) {
         mp_obj_print_exception(&mp_plat_print, exc);
     }
+    if (term == MP_OBJ_NULL) {
+        // Dupterm was already closed.
+        return;
+    }
     nlr_buf_t nlr;
     if (nlr_push(&nlr) == 0) {
         mp_stream_close(term);


dpgeorge:
Thanks for the patch.

I would think it's a bit more robust, and clearer code, to do the following instead:
```diff
--- a/extmod/os_dupterm.c
+++ b/extmod/os_dupterm.c
@@ -45,6 +45,10 @@ void mp_os_deactivate(size_t dupterm_idx, const char *msg, mp_obj_t exc) {
     if (exc != MP_OBJ_NULL) {
         mp_obj_print_exception(&mp_plat_print, exc);
     }
+    if (term == MP_OBJ_NULL) {
+        // Dupterm was already closed.
+        return;
+    }
     nlr_buf_t nlr;
     if (nlr_push(&nlr) == 0) {
         mp_stream_close(term);
```
That would handle all possible cases of the dupterm disappearing.  Eg it could be that the user somehow manages to call `os.dupterm(None)` during a dupterm read.

It also makes sure all error messages are printed, even if the dupterm is already closed.

felixdoerre:
That sounds reasonable. I've tested your suggestion and it seems to work identically well for me. I've changed this PR's contents to your suggestion.

dpgeorge:
Thanks for updating.

Issue #13540, opened 2024-01-28
Webrepl feature parity with mpremote


felixdoerre:
The current webrepl-code seems to be a bit outdated lack behind the feature progress of `mpremote`. Additionally, the current file-transfer logic that aims to work while the REPL is blocked, however this can lead to filesystem corruption if the blocking code accesses the filesystem at the same time.

I propose the following:
* move the password-checking logic from `_webrepl` to `webrepl.py`
* remove the current file-transfer protocol, and with it the `_webrepl` module.
* move the repl interaction from websocket "text" packages to "binary" packages, as the output from the repl is not required to be valid UTF-8, but just a byte sequence.
* implement the same logic that `mpremote` uses for file transfers in the webrepl client.

Removing `_webrepl` and moving the password logic to `webrepl.py` even seems to make the firmware smaller:
```
   text    data     bss     dec     hex filename                                                                       
 606100       0  381308  987408   f1110 /home/jenkins/tmp/micropython/ports/rp2/build-RPI_PICO_W/firmware.elf
 605668       0  381292  986960   f0f50 /home/jenkins/tmp/micropython/ports/rp2/build-RPI_PICO_W/firmware.elf
```

I've implemented a proof-of-concept here: https://github.com/felixdoerre/webreplv2 together with a new webrepl client.

The benefits over the current webrepl are:
* Allow binary data transfers over the websocket (this also helps other usages of the webrepl, like Thonny, or using the websocket as backend for `mpremote` instead of a serial device)
* Implement a File-Browser in the webrepl-client, with file up- and download using the same primitives as `mpremote`
* Implement `mip` to install packages directly from the webrepl-client
* Implement the filesystem-mounting feature
* Implement direct mounting of a local directory through the browser (without uploading). Due to webapi limitations this is only a readonly mount.
* Allow browser-saving of the webrepl password
* Auto-connect the webrepl, if the target (and password) is known.

Do you want to include this into micropython, replacing the current webrepl module?

felixdoerre:
Just to give an overview of the related PRs and their dependencies.

- (No PR yet, I'm not sure if I should open it against micropython/webrepl or micropython-lib or somewhere else): Deploy webrepl_v2.js (https://github.com/felixdoerre/webreplv2) to https://micropython.org/webrepl/webrepl_v2.js
  - https://github.com/micropython/micropython-lib/pull/814
    - https://github.com/micropython/micropython/pull/13786
    - https://github.com/micropython/micropython/pull/14141

[...]

PR #9297, opened 2022-09-13
ports/rp2 implement dupterm_notify

felixdoerre:
`dupterm_notify` allows to send data (especially keyboard interrupts) via a dupterm mechanism (most notably webrepl). This callback is already implemented for `esp32`, `esp8266` and `mimxrt`. This pull request implements the same logic for the `rp2`-port.

PR #13786, opened 2024-02-29
_webrepl: Remove the _webrepl module.

felixdoerre:
When https://github.com/micropython/micropython-lib/pull/814 is merged, all functionality form the `_webrepl` module that is still needed, is part of the python module. The special filesystem protocol is not needed anymore as the web client implements filetransfer exactly as `mpremote` does. Also the filesystem protocol was not safe to use while the repl is busy with other file I/O anyways.

While making way for more features in the webrepl, this change makes the micropython firmware smaller (see the numbers in #13540), the benefit only shows in builds with networking enabled.


dpgeorge:
> The special filesystem protocol is not needed anymore

The original idea for WebREPL (conceived when esp8266 was new and it needed a minimal way to get a remote REPL) was to combine both Telnet and FTP into one protocol (so only one port needed to be listening), and also support accessing the device from a browser (hence websocket instead of normal sockets).

Removing filesystem support is quite a big breaking change.

But I agree that it's a good thing to do!  I think it makes more sense to have filesystem access done using the standard mpremote mechanism (even if that means filesystem acces can no longer be done in the background).

> Also the filesystem protocol was not safe to use while the repl is busy with other file I/O anyways.

Is that really true?  WebREPL callbacks are done at thread level and so filesystem operations are atomic.

One benefit of this PR is that it removes the large C-stack-allocated `filebuf[512]` variable.

felixdoerre:
I am not sure that filesystem operations are atomic. See, for example, this stacktrace: https://github.com/micropython/micropython/pull/9297#issuecomment-1261520716, showing a `mp_handle_pending` in `rp2_flash_writeblocks`. That particular problem was resolved by locking the scheduler in `dupterm_rx_chr`: https://github.com/micropython/micropython/blob/bc424ddc4165ffd8203b4e781f7c686598601e15/extmod/os_dupterm.c#L109

But this does not address possible races of the interrupt with main code (I've not tried to reproduce such a race, but I am pretty sure this is possible). This might be resolvable, but I am actually not sure which guarantee exactly we would need to verify, in order to ensure that file-system operations in a foreground job and an "unsolicited" file-transfer from `dupterm_notify` are safe.

If background operations are actually needed/required, isn't a more stable way to use the async-repl package, and then use the regular transfer logic? (I haven't tested that one yet, but this seems plausible to me).

dpgeorge:
> If background operations are actually needed/required, isn't a more stable way to use the async-repl package, and then use the regular transfer logic?

Yes.  Essentially your application is creating its own "background FTP" task (really an `asyncio` task that handles file transfer).  That does seem more general than having the file transfer baked into the low-level `_webrepl` module.

vshymanskyy:
While working on [`ViperIDE`](https://viper-ide.org), I used the plain `mpremote` protocol for the File Management, and it turns out to work well.

felixdoerre:
From what I see in your code, you don't use the "exact" fragments from `mpremote` but just something similar. In particular you hex-encode the files to download them, which is a workaround against the problem, that `webrepl` as-is cannot send arbitrary bytes on stdout, only utf-8 chars. The adjustments that I suggest to webrepl here make this unnecessary and allow a more efficient binary transfer.

vshymanskyy:
@felixdoerre wow! could you describe briefly the mechanism behind this optimization?

PR #13763, opened 2024-02-27
global: Remove the STATIC macro.

projectgus:
@dpgeorge has said for a while that he wants to get rid of the `STATIC` macro. No time like the present!

The macro is fully removed rather than deprecated. I think this is reasonable as it's a relatively quick refactor to clean up from, but I have a script (see below) to post a heads-up to everyone with an open PR that adds the `STATIC` macro.

If that's not acceptable then I think we could add a CI check that blocks addition of `STATIC`, but keep the macro defined otherwise. But I think this way is probably less fiddly in some ways (doesn't leave contributors bouncing off the CI.)

## Methodology

1) `git ls-files | egrep '\.[ch]$' | xargs sed -Ei "s/(^| )STATIC($| )/\1static\2/"`
2) Do some manual cleanup in the diff by searching for the word STATIC in comments and changing those back.
3) `git-grep STATIC docs/`, manually fixed those cases
4) `rg -t python STATIC`, manually fixed the codegen lines that used STATIC.
5) A few files in renesas-ra port flagged codespell so there are some spelling fixes on top of everything else.
6) Few one-off cases (coverage.cpp file, a lone STATIC in a parenthesis, etc.) were caught by CI.

## Automated heads-up

I have knocked up a [quick script](https://gist.github.com/projectgus/0aa5264d193bd58017205aa5486a9217) to leave a comment on every PR that seems to add `STATIC` in the diff, letting them know that when they rebase they'll need to change it to `static`.

(Of course, there will also be some inevitable merge conflicts from changed lines. There isn't really a way around that, apart from not moving away from the macro.)

## Follow-up work

Separate PR https://github.com/georgerobotics/cyw43-driver/pull/109 to remove STATIC from cyw43-driver repo. Once merged then can revert the commit in this branch that manually sets the macro when building those files.

*This work was funded through GitHub Sponsors.*

PR #14004, opened 2024-03-02
Esp32 bootloader fixes

xyzzy42:
Currently only the Arduino Nano ESP32 defines a machine.bootloader handler for ESP32.  All other boards will intentionally hang.

There is no error message, nor is a NotImplementedError raised.  There's no indication if Micropython has crashed, or if the bootloader was entered but USB is not working, which is a real problem the ESP32 bootloader has.  It's not possible escape from this hang with ^C or any other means besides physical access to the reset pin or the ability to cycle power.

Change this to only define an implementation of machine.bootloader() when there is a handler for it.

Support bootloader on ESP32-S2/S3/C2/C3.

On these targets it's possible to enter the bootloader by setting a bit in an RTC register before resetting.

Structure it in a way that a board can still provide a custom bootloader handler.  The handler here will be the default if none is provided, for any board based on the supported targets.

### DIFF
--- a/ports/esp32/mpconfigport.h
+++ b/ports/esp32/mpconfigport.h
@@ -268,6 +268,14 @@ typedef long mp_off_t;
 #define MICROPY_HW_ENABLE_MDNS_RESPONDER    (1)
 #endif
 
+#ifndef MICROPY_BOARD_ENTER_BOOTLOADER
+// RTC has a register to trigger bootloader on these targets
+#if CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C2 || CONFIG_IDF_TARGET_ESP32C3
+#define MICROPY_ESP32_USE_BOOTLOADER_RTC    (1)
+#define MICROPY_BOARD_ENTER_BOOTLOADER(nargs, args) machine_bootloader_rtc()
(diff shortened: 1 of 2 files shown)


ports/esp32/mpconfigport.h
```diff
@@ -270,7 +269,17 @@ typedef long mp_off_t;
 #endif
 
 #ifndef MICROPY_BOARD_ENTER_BOOTLOADER
-#define MICROPY_BOARD_ENTER_BOOTLOADER(nargs, args)
+// RTC has a register to trigger bootloader on these targets
+#if CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32C2 || CONFIG_IDF_TARGET_ESP32C3
+#define MICROPY_ESP32_USE_BOOTLOADER_RTC    (1)
+#define MICROPY_BOARD_ENTER_BOOTLOADER(nargs, args) machine_bootloader_rtc()
```
xyzzy42, line 275: I structured it this way on the assumption that there will be `#elif` blocks for other targets.  While I found no documentation, there are different bits in different registers that looks like they'd work on the C6, H2, etc.

projectgus:

dpgeorge:
Thanks, this is a nice addition.

PR #14006, opened 2024-03-02
unix: Fix support for Bluetooth via libusb, add HCI dump feature

yasnim24:
ports/unix did not build with
"make MICROPY_PY_BLUETOOTH=1 MICROPY_BLUETOOTH_BTSTACK=1".

This PR fixes it. If libusb-1.0 exists on the machine "MICROPY_BLUETOOTH_BTSTACK_USB=1" is
automatically set and Bluetooth over USB should work.

Now also a log file "/tmp/hci_dump.pklg" could be activated by setting
"MICROPY_BLUETOOTH_BTSTACK_HCI_DUMP=1".

### DIFF
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -33,6 +33,25 @@ QSTR_GLOBAL_DEPENDENCIES += $(VARIANT_DIR)/mpconfigvariant.h
 # OS name, for simple autoconfig
 UNAME_S := $(shell uname -s)
 
+# If the variant enables it, enable modbluetooth.
+ifeq ($(MICROPY_PY_BLUETOOTH),1)
+ifeq ($(MICROPY_BLUETOOTH_BTSTACK),1)
+HAVE_LIBUSB := $(shell (which pkg-config > /dev/null && pkg-config --exists libusb-1.0) 2>/dev/null && echo '1')
+
+# Figure out which BTstack transport to use.
+ifeq ($(HAVE_LIBUSB),1)
+# Default to btstack-over-usb.
+MICROPY_BLUETOOTH_BTSTACK_USB ?= 1
+MICROPY_BLUETOOTH_BTSTACK_H4 ?= 0
+else
+# Fallback to HCI controller via a H4 UART (e.g. Zephyr on nRF) over a /dev/tty serial port.
+MICROPY_BLUETOOTH_BTSTACK_USB ?= 0
+MICROPY_BLUETOOTH_BTSTACK_H4 ?= 1
+endif
+
+endif
+endif
+
 # include py core make definitions
 include $(TOP)/py/py.mk
 include $(TOP)/extmod/extmod.mk
@@ -144,20 +163,8 @@ ifeq ($(MICROPY_SSL_AXTLS),1)
 endif
 endif
 
-# If the variant enables it, enable modbluetooth.
 ifeq ($(MICROPY_PY_BLUETOOTH),1)
 ifeq ($(MICROPY_BLUETOOTH_BTSTACK),1)
-HAVE_LIBUSB := $(shell (which pkg-config > /dev/null && pkg-config --exists libusb-1.0) 2>/dev/null && echo '1')
-
-# Figure out which BTstack transport to use.
-ifeq ($(HAVE_LIBUSB),1)
-# Default to btstack-over-usb.
-MICROPY_BLUETOOTH_BTSTACK_USB ?= 1
-else
-# Fallback to HCI controller via a H4 UART (e.g. Zephyr on nRF) over a /dev/tty serial port.
-MICROPY_BLUETOOTH_BTSTACK_H4 ?= 1
-endif
-
 SRC_BTSTACK_C += lib/btstack/platform/embedded/btstack_run_loop_embedded.c
 endif
 endif
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -49,6 +49,12 @@ MICROPY_BLUETOOTH_BTSTACK_USB ?= 0
 MICROPY_BLUETOOTH_BTSTACK_H4 ?= 1
 endif
 
+# MICROPY_BLUETOOTH_BTSTACK_HCI_DUMP could be set to 1 to activate logging
+# to file "/tmp/hci_dump.pklg" for btstack-over-usb.
+ifndef MICROPY_BLUETOOTH_BTSTACK_HCI_DUMP
+MICROPY_BLUETOOTH_BTSTACK_HCI_DUMP ?= 0
+endif
+
 # Set the file that will be included by "extmod/btstack/btstack_config.h"
 # 
 # The original file "extmod/btstack/btstack_config_common.h"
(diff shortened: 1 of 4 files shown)


yasnim24:
Several minutes ago I saw a commit that could be used to replace the init_mp_state_thread() function in "ports/unix/mpbtstackport_usb.c":
commit bc424dd: "py/modthread: Move thread state initialisation to shared function."

projectgus:

yasnim24:
The first three commits #29f6e66, #7acadbd, #5393f63 are identical with the original commits except that the <author> has been fixed.
The other two commits #b4c69d8 and #0e75c12 are new.

ports/unix/Makefile
```diff
@@ -144,21 +169,27 @@ ifeq ($(MICROPY_SSL_AXTLS),1)
...
-HAVE_LIBUSB := $(shell (which pkg-config > /dev/null && pkg-config --exists libusb-1.0) 2>/dev/null && echo '1')
 
-# Figure out which BTstack transport to use.
-ifeq ($(HAVE_LIBUSB),1)
-# Default to btstack-over-usb.
-MICROPY_BLUETOOTH_BTSTACK_USB ?= 1
-else
-# Fallback to HCI controller via a H4 UART (e.g. Zephyr on nRF) over a /dev/tty serial port.
-MICROPY_BLUETOOTH_BTSTACK_H4 ?= 1
+ifndef MICROPY_BLUETOOTH_BTSTACK_HCI_DUMP
```
projectgus, line 175: Please add a comment here about what this value is for, similar to the other config items.

Suggest even moving it up with the other user-settable values.

ports/unix/Makefile
```diff
@@ -33,6 +33,31 @@ QSTR_GLOBAL_DEPENDENCIES += $(VARIANT_DIR)/mpconfigvariant.h
...
+endif
+
+endif
+endif
+
+# Set the file that will be included by "extmod/btstack/btstack_config.h"
+# 
+# The original file "extmod/btstack/btstack_config_common.h"
+# will be included by it. 
+MICROPY_BLUETOOTH_BTSTACK_CONFIG_FILE ?= '"ports/unix/btstack_config.h"'
```
projectgus, line 59: Suggest moving this up before the `endif` for bluetooth

ports/unix/Makefile
```diff
@@ -33,6 +33,31 @@ QSTR_GLOBAL_DEPENDENCIES += $(VARIANT_DIR)/mpconfigvariant.h
...
+MICROPY_BLUETOOTH_BTSTACK_USB ?= 1
+MICROPY_BLUETOOTH_BTSTACK_H4 ?= 0
+else
+# Fallback to HCI controller via a H4 UART (e.g. Zephyr on nRF) over a /dev/tty serial port.
+MICROPY_BLUETOOTH_BTSTACK_USB ?= 0
+MICROPY_BLUETOOTH_BTSTACK_H4 ?= 1
+endif
+
+endif
+endif
```
projectgus, line 53:
```suggestion
endif  # MICROPY_PY_BLUETOOTH
```

(I know the original version doesn't have this, but it makes it all a lot easier to follow.)

projectgus (commented):
Thanks for submitting this PR, @yasnim24. 

I was a little confused by the title - the unix port already has support for Bluetooth via libusb, but it's broken on master and this PR fixes it, yes? The only new functionality is the HCI dump option.

I think the changes look reasonable to me. We should probably add a bluetooth variant (or two) to the unix port and build them in CI, though, to catch any future bitrot of this kind. You don't have to do that part unless you're keen, though.

yasnim24:
Thank you for your comment, @projectgus.

Your suggested changes are OK for me.
But I'm not sure what to do. Should I add new commits for the changes? Or do I have to change the original commits and do a force push?

Can you also further explain what you mean with:
> ... We should probably add a bluetooth variant (or two) to the unix port and build them in CI, though, to catch any future bitrot of this kind. ...

projectgus:
> But I'm not sure what to do. Should I add new commits for the changes? Or do I have to change the original commits and do a force push?

Best to rebase, change the commits, and force push to keep clean commit history. Jimmo has pointers published at https://github.com/jimmo/git-and-micropython#README

> Can you also further explain what you mean with:
> 
> > ... We should probably add a bluetooth variant (or two) to the unix port and build them in CI, though, to catch any future bitrot of this kind. ...

We have some "variants" for the unix port with different configs:
https://github.com/micropython/micropython/tree/master/ports/unix/variants

However I don't think any of the variants enable the Bluetooth feature. Therefore, when the code broke ("bitrotted") at some point in the past nobody noticed until now. If we had a variant to build Bluetooth via GitHub Actions (Continuous Integration aka CI) then we would have known that this feature no longer compiled.

Don't worry about that for this PR, though.

yasnim24:
Merged changes as suggested by @projectgus to commits [afe2867](https://github.com/micropython/micropython/pull/14006/commits/afe286737f19b3a5bbcdfc452abfb5224db8137d), [60a5e47](https://github.com/micropython/micropython/pull/14006/commits/60a5e47114769f5c20713b48b9b68829e45b80ed).

projectgus (approved):
Sorry for the long delay for re-review, @yasnim24 . This looks good to me, and useful!

agatti:
I happen to have re-done this fix independently (#18972), and I don't mind to close my PR in favour of this one :)

However the SCO compilation workaround I made in https://github.com/micropython/micropython/pull/18972/changes/ae219af730670101d98e589a22d99be39325d61e#diff-b4cac84a1198963d177aafaf45041027d46361cdd1492dabfe17b797c313f91d seems to still be needed, and the workaround proposed in this PR doesn't seem to work for me.  I'm building this via  `make MICROPY_PY_BLUETOOTH=1 MICROPY_BLUETOOTH_BTSTACK=1 MICROPY_BLUETOOTH_BTSTACK_USB=1 MICROPY_BLUETOOTH_BTSTACK_H4=0` but it shouldn't make a difference as far as the SCO code inclusion is concerned.

@yasnim24 can you please check if this still works on your end (and maybe split the dump functionality in its own commit while you're at it)?

yasnim24:
@agatti I read your comment but unfortunately I dont't know when I can check it. The last time I worked on the subject is now over one year ago. That makes it more difficult for me.

agatti:
Thanks for your response!  I guess it all depends on how you plan to handle this for the minor changes needed to make it work on more recent Linux environment (eg. current Arch Linux).

I don't mind taking over and make the changes needed myself whilst keeping your authorship details in (and rework the changes into two build fix + hci dump commits) if this isn't something you plan to get back to and you're fine with me doing that.

yasnim24:
@agatti I'm perfectly fine if you make the changes yourself.

dpgeorge:
@agatti your PR #18972 does fix the SCO gcc warning in a cleaner way.  But it seems there are still some improvements in this PR which yours does not cover.

It would be good to merge both this and your PR, but there are probably going to be clashes.  What do you think?

agatti:
I was planning to rework this into at least two commits, one for the build fix and the other one for the HCI dump.  Since I've got no control over this branch I'll probably add the necessary commits into my repo whilst keeping the original authorship information.

If my SCO fix looks nicer, then I'll keep that and retain the rest.  Let me work on this a bit and see how it ends up looking.

agatti:
OK, I've pushed the modified commits to https://github.com/agatti/micropython/tree/bt-usb-fix-tmp.

I've split the patch into three commits, each using commit subjects that follow the repo guidelines:

- Build fix: mostly the same as the original, with my SCO compilation fix.
- New event loop for BT-over-USB: `pkg-config` is searched using `command -v` first, as per POSIX specs, made the stack size configurable, replaced `log_info` calls with `DEBUG_printf`
- HCI dump: the packet file name directory is looked up using `$TMPDIR` first, and if that's not the case then it defaults to `/tmp`, plus if `$MICROPY_BLUETOOTH_HCI_DUMP` is set, it will use that as the file path.

I've retained the authorship and reused as much of the original commit messages as possible - timestamps may not have been properly preserved.

If I didn't butcher the original PR too much, let me know so those commits can be cherry-picked as they are or if I have to replace my PR's content with these new commits.

dpgeorge:
@agatti thanks for doing that.  Your 3 commits look good, except that it seems you missed the 5th commit from this PR, which removes `init_mp_state_thread()` and instead uses `mp_thread_init_state()`.  Was there a reason to leave that out?

agatti:
It was probably too late at night and I missed that :|

I've merged commit 5 of the original PR into commit 2 of the new patch set.  The commit message doesn't really need to be changed, I think.

PR #14015, opened 2024-03-03
Networking: Add network constants to WLAN classes.

iabdalkader:


dpgeorge:
I just noticed that esp8266 and esp32 have these constants named `AUTH_OPEN` and `AUTH_WPA_PSK`.  Furthermore, they live in the `network` module, not the `network.WLAN` class.

We'll need to make the cyw43 (and nina, esp-hosted) drivers match that.

iabdalkader:
We've had those for a few years now, and there's many scripts out there that use them, if we must use the prefix `AUTH_` can we add aliases so we don't break scripts ?

```C
    // Network is not secured.
    { MP_ROM_QSTR(MP_QSTR_OPEN),                MP_ROM_INT(NINA_SEC_OPEN) },
    // Security type WEP (40 or 104).
    { MP_ROM_QSTR(MP_QSTR_WEP),                 MP_ROM_INT(NINA_SEC_WEP) },
    // Network secured with WPA/WPA2 personal(PSK).
    { MP_ROM_QSTR(MP_QSTR_WPA_PSK),             MP_ROM_INT(NINA_SEC_WPA_PSK) },
```

dpgeorge:
You can keep the existing constants in the nina/esp-hosted `WLAN` objects, for backwards compatibility.

But adding new ones, they should go in `network` and be prefixed with `AUTH_`.

iabdalkader:
> But adding new ones, they should go in `network` and be prefixed with `AUTH_`.

That would still break the networking scripts that have been using those constants, because the same scripts run on CYW43, Nina and ESP hosted.

dpgeorge:
> That would still break the networking scripts that have been using those constants, because the same scripts run on CYW43, Nina and ESP hosted.

But cyw43 doesn't currently have any constants, so those scripts don't work on cyw43 at the moment.

Is it possible to update the networking scripts to use `network.AUTH_xxx`, and then require a firmware update to use these new scripts?

iabdalkader:
> But cyw43 doesn't currently have any constants, so those scripts don't work on cyw43 at the moment.

That true, this PR is to make CYW43 compatible with the existing networking scripts by adding the missing constants. If we add different constants, existing scripts won't run and we'll need to change every other interface to match. Right now Nina, ESP hosted use those constants, and one more driver that's not in MicroPython (a WINC1500 very similar to Nina driver in structure). 

> Is it possible to update the networking scripts to use `network.AUTH_xxx`, and then require a firmware update to use these new scripts?

I will have to change 3 network interfaces and some examples but it will still break user code.

dpgeorge:
Do the network scripts currently work on `ARDUINO_NANO_ESP32`?

iabdalkader:
> How will it break user code if the old constants are left as they are?

The networking scripts expect those constants to be in `network.WLAN.xxx` so they will not run on `CYW43`. The goal is to have a common set of networking scripts, which we have right now.

Note if we add constants `network.AUTH_xxx` they will be enums, and then network interface drivers will need to be changed to map those constants to internal values, right ? It's a good idea, I like it better than duplicating the standard constants, but I'm just pointing out that it will require changes to all drivers.

> Do the network scripts currently work on `ARDUINO_NANO_ESP32`?

No they don't but we don't support any ESP boards.

dpgeorge:
Just discussing this with @jimmo and we may actually want to go the other way: move everything out of `network` and into individual classes.  Ie, this PR is correct (although I still think the constants need to be named `AUTH_xxx`).

Also then need to move things like `network.PHY_xxx` to `network.LAN.PHY_xxx`, etc.

jimmo:
The main justification for the above is the way constants work elsewhere, e.g. `Pin.X` rather than on `machine` directly. We should do the same for `network`.

iabdalkader:
There's one more issue with `AUTH_xxx` it doesn't match the keyword arg, `security=AUTH_xxx` should be `SEC_` if we must have a prefix.

iabdalkader:
> Also then need to move things like network.PHY_xxx to network.LAN.PHY_xxx, etc.

> The main justification for the above is the way constants work elsewhere, e.g. `Pin.X` rather than on `machine` directly. We should do the same for `network`.

machine groups many modules, each one has its own constants, that aren't shared with any other module. `network`'s constants on the other hand are common, standard constants that can be shared between all networking interfaces, and will result in less duplicate constants. I assume this change would also include `networking.STA/AP_IF` ? This is a very breaking change, every script assumes those are accessible from `network`. I don't think this is a good change, those constants should all be in network.

iabdalkader:
> It happens to me regularly that I type WLAN.STA_IF, because I expect it there

If all constants are moved to one module, regardless of what that module is, instead of being split between `network` and `WLAN`, there should be less confusion about this.  However I still think they should be in `network`, because this way they can be shared, instead of having to add the same constants to every network interface. Note that the ports in question, esp32 and esp8266,  already have them in `network` via `MICROPY_PY_NETWORK_MODULE_GLOBALS_INCLUDEFILE`. I guess this config option, and include files, will not be needed either.

dpgeorge:
Note that stm32 and mimxrt already have `network.LAN.PHY_xxx` constants.  It makes sense to have these PHY constants in the LAN class because:
1. Other LAN drivers may have a different set of PHYs that they support, and hence different constants.
2. Other LAN drivers may have different values for the PHY constants.

Similarly for `network.WLAN`, if you have multiple WLAN drivers compiled in then each one may have a different set of constants (different AUTH modes available, and they may have different values for their constants).

iabdalkader:
> Similarly for `network.WLAN`, if you have multiple WLAN drivers compiled in then each one may have a different set of constants (different AUTH modes available, and they may have different values for their constants).

It's true some drivers will not support all modes, and constants have different values, but that's not a big problem we can use an enum and map to the actual value in drivers and fail if security mode is not supported.

I've pushed a change to move the constants to WLAN, they have to be duplicated in every driver. Note I used `AUTH_` but this doesn't match the keyword arg, it should be `SEC_`.

dpgeorge:
I think we should put the constants on the class.  It keeps the constants grouped together where they belong and with the class that they are used for.

There aren't going to be that many cases where there are common constants, eg WLAN and ETH have pretty much nothing in common.

> Note I used `AUTH_` but this doesn't match the keyword arg, it should be `SEC_`.

Yes, I agree, the constants should start with `SEC_`.  I think we should do that here.

> I assume this change would also include `networking.STA/AP_IF` ?

Yes... I think we should also move those constants to their class.  And while we are moving them, we can rename them to `IF_STA` and `IF_AP`, to have a common prefix, to match all the other groups of constants.

Of course we would retain the old constants for backwards compatibility (eg until MicroPython 2.0).

extmod/network_cyw43.c
```diff
@@ -536,6 +536,11 @@ STATIC const mp_rom_map_elem_t network_cyw43_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_config), MP_ROM_PTR(&network_cyw43_config_obj) },
 
     // Class constants.
+    { MP_ROM_QSTR(MP_QSTR_STA_IF), MP_ROM_INT(MOD_NETWORK_STA_IF) },
+    { MP_ROM_QSTR(MP_QSTR_AP_IF), MP_ROM_INT(MOD_NETWORK_AP_IF) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_OPEN), MP_ROM_INT(CYW43_AUTH_OPEN) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_WPA_PSK), MP_ROM_INT(CYW43_AUTH_WPA2_AES_PSK) },
```
dpgeorge, line 542: I suggest using `SEC_WPA_WPA2` for this constant, because:
- It should explicitly state that it's mixed WPA/WPA2 mode.
- The "psk" (pre shared key) suffix isn't really needed.  It's only to distinguish with other auth modes like enterprise.  But enterprise can be distinguished with a `_ENT` suffix.

Also, the corresponding cyw43 constant should be `CYW43_AUTH_WPA2_MIXED_PSK`.

Can add `SEC_WPA2` with `CYW43_AUTH_WPA2_AES_PSK` if needed.

extmod/network_esp_hosted.c
```diff
@@ -301,9 +301,13 @@ STATIC const mp_rom_map_elem_t network_esp_hosted_locals_dict_table[] = {
...
-    { MP_ROM_QSTR(MP_QSTR_OPEN),                MP_ROM_INT(ESP_HOSTED_SEC_OPEN) },
-    { MP_ROM_QSTR(MP_QSTR_WEP),                 MP_ROM_INT(ESP_HOSTED_SEC_WEP) },
-    { MP_ROM_QSTR(MP_QSTR_WPA_PSK),             MP_ROM_INT(ESP_HOSTED_SEC_WPA_WPA2_PSK) },
+
+    // Class constants.
+    { MP_ROM_QSTR(MP_QSTR_STA_IF),              MP_ROM_INT(MOD_NETWORK_STA_IF) },
+    { MP_ROM_QSTR(MP_QSTR_AP_IF),               MP_ROM_INT(MOD_NETWORK_AP_IF) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_OPEN),           MP_ROM_INT(ESP_HOSTED_SEC_OPEN) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_WEP),            MP_ROM_INT(ESP_HOSTED_SEC_WEP) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_WPA_PSK),        MP_ROM_INT(ESP_HOSTED_SEC_WPA_WPA2_PSK) },
```
dpgeorge, line 310: Constant should be `SEC_WPA_WPA2`.

extmod/network_wiznet5k.c
```diff
@@ -1013,6 +1013,9 @@ STATIC const mp_rom_map_elem_t wiznet5k_locals_dict_table[] = {
     #if WIZNET5K_WITH_LWIP_STACK
     { MP_ROM_QSTR(MP_QSTR_send_ethernet), MP_ROM_PTR(&send_ethernet_obj) },
     #endif
+    // Class constants.
+    { MP_ROM_QSTR(MP_QSTR_STA_IF), MP_ROM_INT(MOD_NETWORK_STA_IF) },
+    { MP_ROM_QSTR(MP_QSTR_AP_IF), MP_ROM_INT(MOD_NETWORK_AP_IF) },
```
dpgeorge, line 1018: Wiznet is a LAN and shouldn't have these constants at all.

ports/esp32/network_wlan.c
```diff
@@ -711,6 +711,9 @@ STATIC const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_PM_NONE), MP_ROM_INT(WIFI_PS_NONE) },
     { MP_ROM_QSTR(MP_QSTR_PM_PERFORMANCE), MP_ROM_INT(WIFI_PS_MIN_MODEM) },
     { MP_ROM_QSTR(MP_QSTR_PM_POWERSAVE), MP_ROM_INT(WIFI_PS_MAX_MODEM) },
+    #ifdef MICROPY_PY_NETWORK_MODULE_GLOBALS_INCLUDEFILE
+    #include MICROPY_PY_NETWORK_MODULE_GLOBALS_INCLUDEFILE
+    #endif
```
dpgeorge, line 716: This will include everything, including LAN constants, which is wrong.

I suggest to just copy across the relevant WLAN constants.

extmod/network_ninaw10.c
```diff
@@ -849,12 +849,15 @@ STATIC const mp_rom_map_elem_t nina_locals_dict_table[] = {
...
+    { MP_ROM_QSTR(MP_QSTR_AP_IF),               MP_ROM_INT(MOD_NETWORK_AP_IF) },
     // Network is not secured.
-    { MP_ROM_QSTR(MP_QSTR_OPEN),                MP_ROM_INT(NINA_SEC_OPEN) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_OPEN),           MP_ROM_INT(NINA_SEC_OPEN) },
     // Security type WEP (40 or 104).
-    { MP_ROM_QSTR(MP_QSTR_WEP),                 MP_ROM_INT(NINA_SEC_WEP) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_WEP),            MP_ROM_INT(NINA_SEC_WEP) },
     // Network secured with WPA/WPA2 personal(PSK).
-    { MP_ROM_QSTR(MP_QSTR_WPA_PSK),             MP_ROM_INT(NINA_SEC_WPA_PSK) },
+    { MP_ROM_QSTR(MP_QSTR_AUTH_WPA_PSK),        MP_ROM_INT(NINA_SEC_WPA_PSK) },
```
dpgeorge, line 860: Is this mixed mode?  `SEC_WPA_WPA2`.

ports/esp8266/network_wlan.c
```diff
@@ -519,6 +519,9 @@ STATIC const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_PM_NONE), MP_ROM_INT(NONE_SLEEP_T) },
     { MP_ROM_QSTR(MP_QSTR_PM_PERFORMANCE), MP_ROM_INT(MODEM_SLEEP_T) },
     { MP_ROM_QSTR(MP_QSTR_PM_POWERSAVE), MP_ROM_INT(LIGHT_SLEEP_T) },
+    #ifdef MICROPY_PY_NETWORK_MODULE_GLOBALS_INCLUDEFILE
+    #include MICROPY_PY_NETWORK_MODULE_GLOBALS_INCLUDEFILE
+    #endif
```
dpgeorge, line 524: Please add the `IF_xxx` and `SEC_xxx` constants explicitly, not by including this header.

extmod/network_ninaw10.c
jimmo: When starting the AP mode, it looks like only WEP is supported (in our driver, `nina_start_ap` only supports the open and WEP constants).

But the command we send to the Nina doesn't actually specify the mode. Our driver and the Arduino driver use the same command (0x19, which they call `SET_AP_PASSPHRASE_CMD` and document as WPA, and we call `NINA_CMD_START_AP_WEP`).

iabdalkader:
> Of course we would retain the old constants for backwards compatibility (eg until MicroPython 2.0).

Does this also mean that we should keep both `AUTH_` and `SEC_` ?

extmod/network_ninaw10.c
iabdalkader: > Our driver and the Arduino driver use the same command

No we use different commands: https://github.com/micropython/micropython/blob/master/drivers/ninaw10/nina_wifi_drv.c#L402
It's seems that WPA is also supported in AP mode, in the firmware, but there's no command to actually connect use it. Maybe I will push a firmware update later to fix it.

> Is this mixed mode? `SEC_WPA_WPA2`.

@dpgeorge The security mode is not set in station mode, it seems to connect to all modes, except for deprecated/unsecure ones like WEP.

extmod/network_cyw43.c
iabdalkader: @dpgeorge So remove `_PSK` from any other WPA constant as well ? Like :
```
    { MP_ROM_QSTR(MP_QSTR_SEK_WPA_PSK), MP_ROM_INT(AUTH_WPA_PSK) },
    { MP_ROM_QSTR(MP_QSTR_SEK_WPA2_PSK), MP_ROM_INT(AUTH_WPA2_PSK) },
    { MP_ROM_QSTR(MP_QSTR_SEK_WPA_WPA2_PSK), MP_ROM_INT(AUTH_WPA_WPA2_PSK) },
```

ports/esp32/network_wlan.c
iabdalkader: Okay

iabdalkader:
Updated. I've kept the existing constants to maintain backwards compatibility. As for esp ports, the constants in `network.XXX` are the same, included via the config file, so it should still be backwards compatible.

extmod/network_ninaw10.c
```diff
@@ -860,6 +860,17 @@ static const mp_rom_map_elem_t nina_locals_dict_table[] = {
...
     { MP_ROM_QSTR(MP_QSTR_ioctl),               MP_ROM_PTR(&network_ninaw10_ioctl_obj) },
 
+    // Class constants.
+    { MP_ROM_QSTR(MP_QSTR_IF_STA),              MP_ROM_INT(MOD_NETWORK_STA_IF) },
+    { MP_ROM_QSTR(MP_QSTR_IF_AP),               MP_ROM_INT(MOD_NETWORK_AP_IF) },
+    // Network is not secured.
+    { MP_ROM_QSTR(MP_QSTR_SEC_OPEN),            MP_ROM_INT(NINA_SEC_OPEN) },
+    // Security type WEP (40 or 104).
+    { MP_ROM_QSTR(MP_QSTR_SEC_WEP),             MP_ROM_INT(NINA_SEC_WEP) },
+    // Network secured with WPA/WPA2 personal(PSK).
```
dpgeorge, line 870: Do we need these comments?  I understand they are copied from below, but none of the other NICs have comments.  Better to add this info in the documentation, it would be more useful there and apply to all NICs.

ports/esp8266/network_wlan.c
```diff
@@ -519,6 +519,15 @@ static const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
...
     { MP_ROM_QSTR(MP_QSTR_PM_POWERSAVE), MP_ROM_INT(LIGHT_SLEEP_T) },
+
+    { MP_ROM_QSTR(MP_QSTR_IF_STA), MP_ROM_INT(STATION_IF)},
+    { MP_ROM_QSTR(MP_QSTR_IF_AP), MP_ROM_INT(SOFTAP_IF)},
+
+    { MP_ROM_QSTR(MP_QSTR_SEK_OPEN), MP_ROM_INT(AUTH_OPEN) },
+    { MP_ROM_QSTR(MP_QSTR_SEK_WEP), MP_ROM_INT(AUTH_WEP) },
+    { MP_ROM_QSTR(MP_QSTR_SEK_WPA), MP_ROM_INT(AUTH_WPA_PSK) },
+    { MP_ROM_QSTR(MP_QSTR_SEK_WPA2), MP_ROM_INT(AUTH_WPA2_PSK) },
+    { MP_ROM_QSTR(MP_QSTR_SEK_WPA_WPA2), MP_ROM_INT(AUTH_WPA_WPA2_PSK) },
```
dpgeorge, line 530: SEK->SEC

mattytrentini:
Documentation, folks! As far as I can tell these are not documented _anywhere_. 

I'm happy to add documentation but it's not clear to me what is port-specific and what is general...

This is a real example: I was trying to create a simple, secured AP on ESP32 recently and couldn't figure out the appropriate configuration. Old forum posts with now-obsolete items make things worse - so I think we _need_ to add a little more to the docs.

iabdalkader:
> I'm happy to add documentation but it's not clear to me what is port-specific and what is general...

When ports register network interfaces they either uses network.LAN or network.WAN, with the exception of things like `WIZNET5K` which uses its own name. So I think the constants should be documented here https://docs.micropython.org/en/latest/library/network.WLAN.html with notes that some modes might not be supported.

Issue #14013, opened 2024-03-03
Ethernet fails on the Olimex ESP32-POE-ISO Revision L


PhantomPhoton:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

ESP32-POE-ISO Revision L

### MicroPython version

MicroPython v1.22.2 on 2024-02-22; Olimex ESP32 ETH with ESP32

### Reproduction

lan = network.LAN(mdc=machine.Pin(23), mdio=machine.Pin(18), power=machine.Pin(12),
                  phy_type=network.PHY_LAN8720, phy_addr=0,
                  ref_clk=machine.Pin(17), ref_clk_mode=machine.Pin.OUT)
lan.active(True)

### Expected behaviour

Expected ethernet to enable

### Observed behaviour

OSError: ethernet enable failed

### Additional Information

I've pulled the expected working code from https://docs.micropython.org/en/latest/esp32/quickref.html#lan

I know the Ethernet on the device works, when i flash it with a example program from arduino ide, it gets an ip and works correctly.

robert-hh:
It works here if I use a slightly reduced call:
```
import machine, network, time
lan=network.LAN(mdc=machine.Pin(23), mdio=machine.Pin(18), power=machine.Pin(12),
    phy_type=network.PHY_LAN8720, phy_addr=0)

lan.active(True)
```
Adding the settings for ref_clk_mode causes an error. That's stumping, because it used to work. Maybe related to  ESP IDF version change.
Note: When it failed once, the board must be reset with a power cycle.

robert-hh:
Strange. After have the board powered for a while, even your variant with specifying ref_clk works. That has to be further analyzed.

robert-hh:
There is a strange behaviour at the MDC line of LAN8720. At the first MDIO message after Reset, the level of the clock signal slowly ramps up. Frequency and rise/fall times are not affected.  Further messages are fine. This behavior can be repeated several times. So it's not that the probe was not on spot
First message MDC
![esp32_mdc_pkt1](https://github.com/micropython/micropython/assets/12476868/7143775a-5ef1-49da-a1cb-e5952a60a2da)

Further messages MDC

[...]

PR #14016, opened 2024-03-03
esp/network_lan.c: Provide separate arguments for phy_reset and phy_power.

robert-hh:
Before, the pin defined for power would be used by the esp_idf driver to reset the PHY. That worked, but sometimes the MDIO configuration started before the power was fully settled, leading to an error. With that change, the power for the PHY is independently enabled in network_lan.c with a 100ms delay to allow the power to settle. A separate define for a reset Pin is provided, even if the PHY reset pin is rarely connected.

It is assumed, that setting the power control pin high enables the power.
Tested with an Olimex ESP32 Ethernet POE board.
Addresses Issue #14013


dpgeorge:
Thanks for the fix!

PR #14019, opened 2024-03-03
py/emitglue: Add explicit cast of proto_fun to uint8_t pointer.

dpgeorge:
Otherwise C++ compilers may complain when this header is included in an extern "C" block.


dcdunn:
Thank  you Damien.

PR #14022, opened 2024-03-04
boards/Arduino: Add missing pin names from documentation.

sebromero:
This PR adds missing pin names from the documentation on https://docs.arduino.cc/. For the Arduino GIGA R1 WiFi it changes the DAC pin numbers to be consistent with the pinout document.


ports/stm32/boards/ARDUINO_GIGA/pins.csv
```diff
@@ -1,3 +1,115 @@
+D0,PB7
```
dpgeorge, line 1: Please shift these down to after `PK7`, ie just before the UART definitions begin.

Also, please describe them with a comment so it's clear what they are for, eg:
```
PK7,PK7

# Arduino digital/analog pins
D1,PA9
...
```

ports/stm32/boards/ARDUINO_NICLA_VISION/pins.csv
```diff
@@ -1,3 +1,22 @@
+D0,PG12
```
dpgeorge, line 1: Please shift down and add a comment, as for `ARDUINO_GIGA`.

ports/stm32/boards/ARDUINO_PORTENTA_H7/pins.csv
```diff
@@ -1,3 +1,23 @@
+A4,PC2
```
dpgeorge, line 1: Please shift down and add a comment, as for `ARDUINO_GIGA`.

dpgeorge:
Thanks for this.  Now merged (and I added some missing _C pins on Portenta H7 from #13764).

PR #14023, opened 2024-03-04
Nina WiFi network interface updates.

iabdalkader:
A couple of small fixes/changes:

* Allow the network interface to be configured before/after it's activated.
* Set the default security mode if none is provided.


iabdalkader:
> can you please improve the error messages in the ninaw10 driver

@dpgeorge https://github.com/micropython/micropython/pull/14094

PR #14025, opened 2024-03-04
esp32/machine_uart: Apply timeout_char default.

DvdGiessen:
If the `timeout_char` parameter is not given, we should still configure the UART to ensure the UART is always initialized consistently. So the default of `0`get applied correctly, or if for example the baud rate was changed the char timeout isn't still based on the old baud rate causing weird behaviour, etc.


dpgeorge:
Thanks, this fix looks correct.

PR #14029, opened 2024-03-05
py/objarray: Fix use-after-free if extending a slice from itself.

### LINKED ISSUE #13283: use after free found at micropython/py/objarray.c:509 [micropython@a5bdd39127]
Hi all, I'm Wonil Jang, from the research group S2Lab in UNIST.
We found a use after free bug from micropython by our custom tool. The detailed information is as follows.

## Environment
- OS: Ubuntu 22.04
- Version: micropython at commit a5bdd39127
- Build: unix ports
- Bug Type: use-after-free
- Bug Location: [micropython/py/objarray.c:509](https://github.com/micropython/micropython/blob/master/py/objarray.c#L509)
- Credits: Junwha Hong and Wonil Jang, from S2Lab in UNIST

## Proof-of-Concept (PoC)
```python
b = bytearray(b'1234567')
for i in range(3):
     b[-1:] = b
```

## Problem Statement
By PoC, 

```cpp

STATIC mp_obj_t array_subscr(mp_obj_t self_in, mp_obj_t index_in, mp_obj_t **value**) {
...
size_t src_len;
void *src_items;
size_t item_sz = mp_binary_get_size('@', o->typecode & TYPECODE_MASK, NULL);
if (mp_obj_is_obj(value) && MP_OBJ_TYPE_GET_SLOT_OR_NULL(((mp_obj_base_t *)MP_OBJ_TO_PTR(value))->type, subscr) == array_subscr) {
    // value is array, bytearray or memoryview
    **mp_obj_array_t *src_slice = MP_OBJ_TO_PTR(value);**
    if (item_sz != mp_binary_get_size('@', src_slice->typecode & TYPECODE_MASK, NULL)) {
    compat_error:
        mp_raise_ValueError(MP_ERROR_TEXT("lhs and rhs should be compatible"));
    }
    **src_len = src_slice->len;**
    **src_items = src_slice->items;**
    #if MICROPY_PY_BUILTINS_MEMORYVIEW
    if (mp_obj_is_type(value, &mp_type_memoryview)) {
        src_items = (uint8_t *)src_items + (src_slice->memview_offset * item_sz);
    }
    #endif
}
...
if ((size_t)len_adj > o->free) {
    // TODO: alloc policy; at the moment we go conservative
    **o->items = m_renew(byte, o->items, (o->len + o->free) * item_sz, (o->len + len_adj) * item_sz);**
    o->free = len_adj;
    dest_items = o->items;
}
mp_seq_replace_slice_grow_inplace(dest_items, o->len,
    slice.start, slice.stop, **src_items**, src_len, len_adj, item_sz);

```
If you run the PoC, ```self_in``` and ```value``` can be same in ```array_subscr``` function. This is because when ```subscr()``` function is called, It just uses reference without copying. and then, if the address is rebased after calling realloc, only ```self_in``` changed and ```value```(src) becomes a dangling pointer. So when copying it, the use after free bug occurs.

## Patch
It should keep source before calling fixed realloc().

## Log
```cpp
    #1 0x5555557340a4 in array_subscr /home/qbit/testing-2023/micropython/ports/unix/../../py/objarray.c:509:21
    #2 0x555555731de6 in mp_obj_subscr /home/qbit/testing-2023/micropython/ports/unix/../../py/obj.c:536:24
    #3 0x555555781d05 in mp_execute_bytecode /home/qbit/testing-2023/micropython/ports/unix/../../py/vm.c:487:21
    #4 0x55555574261b in fun_bc_call /home/qbit/testing-2023/micropython/ports/unix/../../py/objfun.c:273:42
    #5 0x555555903f3d in execute_from_lexer /home/qbit/testing-2023/micropython/ports/unix/main.c:161:13
    #6 0x555555902ad5 in do_file /home/qbit/testing-2023/micropython/ports/unix/main.c:310:12
    #7 0x555555902ad5 in main_ /home/qbit/testing-2023/micropython/ports/unix/main.c:722:19
    #8 0x7ffff7c29d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #9 0x7ffff7c29e3f in __libc_start_main csu/../csu/libc-start.c:392:3
    #10 0x555555593a34 in _start (/home/qbit/testing-2023/micropython/ports/unix/build-standard/micropython+0x3fa34)

```
Thank you for reading my report.

projectgus:
Closes https://github.com/micropython/micropython/issues/13283 which reported this bug when assigning to a slice, also fixes a related bug when a bytearray is extended by itself (h/t @dpgeorge for spotting.) Unit tests added for both.

There is a related much more nefarious set of bugs when a bytearray or similar is targeted by a memoryview and then resized. For now these are not fixed, but have pushed a commit here to document this case as a difference from CPython.

Reproducing this bug and confirming the fix was done by running the unix port under valgrind with GC-aware extensions (PR at https://github.com/micropython/micropython/pull/14196).

Note in default configurations with GIL this bug exists but has no impact (the free buffer won't have new data written to  it while the function is still executing, and is no longer referenced after the function returns).


py/objarray.c
```diff
@@ -504,13 +505,18 @@ STATIC mp_obj_t array_subscr(mp_obj_t self_in, mp_obj_t index_in, mp_obj_t value
                         // TODO: alloc policy; at the moment we go conservative
                         o->items = m_renew(byte, o->items, (o->len + o->free) * item_sz, (o->len + len_adj) * item_sz);
                         o->free = len_adj;
-                        dest_items = o->items;
+                        if (o->items != dest_items) { // m_renew moved o->items
```
dpgeorge, line 508: I don't think this check is needed, the code in this `if` can just be executed unconditionally.

tests/basics/bytearray_slice_assign.py
```diff
@@ -61,8 +61,14 @@
...
+b.append(57)  # expand and add a bit of unused space at end of the bytearray
 for i in range(400):
-    b[-1:] = b'ab' # grow slowly into the unused space
+    b[-1:] = b"ab"  # grow slowly into the unused space
+print(len(b), b)
+
+# Growth of bytearray via slice extension from itself
+b = bytearray(b"1234567")
+for i in range(3):
+    b[-1:] = b
```
dpgeorge, line 73: I think there needs to be an additional test that extends a bytearray by a memoryview of itself, where that memoryview is offset by non-zero.  That will test the case where `src_items` was realloc'd by the GC and `src_offs` is non-zero.

py/objarray.c
projectgus: ah yeah, ofc! New patch incoming.

tests/basics/bytearray_slice_assign.py
projectgus: Good catch, I'd missed there's no existing coverage for this.

Interestingly, this turns up a twist: CPython doesn't support resizing any buffer object which has an active memoryview into it. This code:

```py
# Growth of bytearray via slice extension of a memoryview to itself
b = bytearray(b"1234567")
m = memoryview(b)[2:5]
for _ in range(3):
    b[-1:] = m
print(len(b), b)
```

Triggers:

```
  File "/home/gus/ry/george/micropython/tests/basics/bytearray_slice_assign.py", line 80, in <module>
    b[-1:] = m
    ~^^^^^
BufferError: Existing exports of data: object cannot be re-sized
```

[Some explanation here](https://bugs.python.org/issue41097).

You're no doubt more familiar with this than me, but AFAIK the related memory safety issue also exists in MicroPython - if there's a memoryview into a buffer object and an unrelated resize moves the buffer then the memoryview will point to invalid memory. My understanding is this a known (but undocumented?) limitation of memoryviews in Micropython.

In this particular case (buffer extended by memoryview to itself) this patch makes it memory safe, so perhaps I can make a separate test file for this with its own .exp? Plus add some documentation around the difference to CPython. What do you recommend?

dpgeorge: > if there's a memoryview into a buffer object and an unrelated resize moves the buffer then the memoryview will point to invalid memory.

The memoryview will still point to "valid" memory, and importantly memory that won't be GC'd because the memoryview object retains a pointer to the head of the buffer.



> In this particular case (buffer extended by memoryview to itself) this patch makes it memory safe, so perhaps I can make a separate test file for this with its own .exp? Plus add some documentation around the difference to CPython. What do you recommend?

Yes, that all sounds good: separate test with .exp, and cpydiff documentation.

projectgus: > The memoryview will still point to "valid" memory, and importantly memory that won't be GC'd because the memoryview object retains a pointer to the head of the buffer.

That's a pretty elegant solution, but I don't think this is how it works at the moment. As per the bug this PR is fixing, resizing calls `m_renew` which calls through to `gc_realloc`. If resizing in place fails, `gc_realloc` allocates a new buffer, copies into it, and then frees the old one with `gc_free`. So any remaining pointers into that buffer from other memoryviews will become pointers to freed blocks.

Maybe it should work in the way you describe, where the old buffer is left for the GC to pick up once it's no longer used?

dpgeorge: > If resizing in place fails, `gc_realloc` allocates a new buffer, copies into it, and then frees the old one with `gc_free`. So any remaining pointers into that buffer from other memoryviews will become pointers to freed blocks.

Ahh, indeed!  That's a problem.  But that's kind of separate to this PR, so maybe we can think about that and fix that separately?

projectgus: Up to you. If we were to change resizing to not explicitly free the old buffer then this PR wouldn't be necessary (no use after free would exist here, as the old buffer hadn't been freed yet).

dpgeorge: > If we were to change resizing to not explicitly free the old buffer then this PR wouldn't be necessary

Yes, but I feel like that's quite a big change that doesn't benefit many cases.  In most cases the code won't be resizing a memoryview and it's definitely beneficial (??) to explicitly free memory doing a realloc.

Also, are there other places in the code (probably confined to `py/objarray.c`) where we'd also need to be careful not to free memory that may be pointed to by a memoryview?

projectgus: I've updated the PR to address the cases with bytearrays. The cases with memoryviews are much more complex, as we discussed offline. Have added a commit here to document the requirement of not resizing a bytearray that has a memoryview pointing to it, at least until we have a better fix.

PR #14040, opened 2024-03-06
docs/library/machine.UART: Fix UART.irq docs to match current code.

dpgeorge:
These docs now match the code in `extmod/machine_uart.c`.

PR #14041, opened 2024-03-06
ports: Support UART.irq for the RP2, ESP32, MIMXRT, NRF, Renesas-RA and SAMD ports.

robert-hh:
Adds UART IRQ functionality for 4 additional ports: RP2, ESP32, MIMXRT and SAMD. Available triggers:

IRQ_RX: called every time when a byte is received. Not useful for high data rates (like > 9600 baud).
IRQ_RXIDLE: called when the receive line gets silent after a burst of data. Very useful!
IRQ_TXIDLE: called when a block of data has been sent.

Not all ports provide all triggers:

- rp2: IRQ_RXIDLE, IRQ_TXIDLE and IRQ_BREAK. IRQ_RXIDLE works as it should. IRQ_TXIDLE is only called when the data length is mor than 5 bytes, and it triggers when still 4-5 bytes are to be sent.
- MIMXRT: IRQ_RXIDLE and IRQ_TXIDLE. Work as they should with a fix to the NXP UART driver. which is included in this PR.
- SAMD: IRQ_RX and IRQ_TXIDLE. The SAMD port's hardware does not support IRQ_RXIDLE. Based on the IRQ_RX trigger a simple class can be made which implements a IRQ_RXIDLE handler call. IRQ_TXIDLE triggers while the last byte is sent.
- ESP32: IRQ_RX and IRQ_BREAK: The ESP32 does not support IRQ_RXIDLE. The option `hard=True` is not supported.
- NRF: IRQ_RX and IRQ_TXIDLE. Enabled for boards with nrf52480. I could not test other boards.

Tested with RP2040, ESP32. ESP32-S2, ESP32-S3, ESP32-C3, SAMD21, SAMD51, NRF52840, and all supported MIMXRT MCUs except i.MX RT 1064 (I do not have access to a board).

Test observations for other boards:

- STM32: IRQ_RXIDLE. For F4xx MCU (e.g. Pyboard), the RXIDLE interrupt triggers twice with every message longer than 1 byte starting with the second message after UART init. The first trigger happens at the first byte of the message, the second after the message. 1 Byte messages get only one trigger, but with different delays after the message. Tests with F7xx and H7xx boards showed proper behavior.
- Renesas-RA: Trigger IRQ_RX.
- CC2300: Trigger RX_ANY. IRQ_RX was added as alias in this PR.

### DIFF
--- a/docs/library/machine.UART.rst
+++ b/docs/library/machine.UART.rst
@@ -152,31 +152,6 @@ Methods
    Send a break condition on the bus. This drives the bus low for a duration
    longer than required for a normal transmission of a character.
 
-.. method:: UART.irq(trigger, priority=1, handler=None, wake=machine.IDLE)
-
-   Create a callback to be triggered when data is received on the UART.
-
-       - *trigger* can only be ``UART.RX_ANY``
-       - *priority* level of the interrupt. Can take values in the range 1-7.
-         Higher values represent higher priorities.
-       - *handler* an optional function to be called when new characters arrive.
-       - *wake* can only be ``machine.IDLE``.
-
-   .. note::
-
-      The handler will be called whenever any of the following two conditions are met:
-
-          - 8 new characters have been received.
-          - At least 1 new character is waiting in the Rx buffer and the Rx line has been
-            silent for the duration of 1 complete frame.
-
-      This means that when the handler function is called there will be between 1 to 8
-      characters waiting.
-
-   Returns an irq object.
-
-   Availability: WiPy.
-
 .. method:: UART.flush()
 
    Waits until all data has been sent. In case of a timeout, an exception is raised. The timeout
@@ -203,11 +178,42 @@ Methods
 
    Availability: rp2, esp32, esp8266, mimxrt, cc3200, stm32, nrf ports, renesas-ra
 
+.. method:: UART.irq(handler=None, trigger=0, hard=False)
+
+   Configure an interrupt handler to be called when a UART event occurs.
+
+   The arguments are:
+
+     - *handler* is an optional function to be called when the interrupt event
+       triggers.  The handler must take exactly one argument which is the
+       ``UART`` instance.
+
+     - *trigger* configures the event which can generate an interrupt.
+       Possible values are:
+
+       - ``UART.IRQ_RXIDLE`` interrupt after receiving at least one character
+         and then the RX line goes idle.
+
+     - *hard* if true a hardware interrupt is used.  This reduces the delay
+       between the pin change and the handler being called.  Hard interrupt
+       handlers may not allocate memory; see :ref:`isr_rules`.
+
+   Returns an irq object.
+
+   Availability: renesas-ra, stm32.
+
 Constants
 ---------
 
-.. data:: UART.RX_ANY
+.. data:: UART.RTS
+          UART.CTS
 
-    IRQ trigger sources
+    Flow control options.
 
-    Availability: WiPy.
+    Availability: esp32, mimxrt, renesas-ra, rp2, stm32.
+
+.. data:: UART.IRQ_RXIDLE
+
+    IRQ trigger sources.
+
+    Availability: stm32.
--- a/docs/library/machine.UART.rst
+++ b/docs/library/machine.UART.rst
@@ -188,19 +188,59 @@ Methods
        triggers.  The handler must take exactly one argument which is the
        ``UART`` instance.
 
-     - *trigger* configures the event which can generate an interrupt.
-       Possible values are:
+     - *trigger* configures the event(s) which can generate an interrupt.
+       Possible values are a mask of one or more of the following:
 
        - ``UART.IRQ_RXIDLE`` interrupt after receiving at least one character
          and then the RX line goes idle.
+       - ``UART.IRQ_RX`` interrupt after each received character.
+       - ``UART.IRQ_TXIDLE`` interrupt after or while the last character(s) of
+         a message are or have been sent.
+       - ``UART.IRQ_BREAK`` interrupt when a break state is detected at RX
 
      - *hard* if true a hardware interrupt is used.  This reduces the delay
-       between the pin change and the handler being called.  Hard interrupt
+       between the pin change and the handler being called. Hard interrupt
        handlers may not allocate memory; see :ref:`isr_rules`.
 
    Returns an irq object.
 
-   Availability: renesas-ra, stm32.
+   Due to limitations of the hardware not all trigger events are available on all ports.
+
+   .. table:: Availability of triggers
+      :align: center 
+
+      ============== ========== ====== ========== =========
+      Port / Trigger IRQ_RXIDLE IRQ_RX IRQ_TXIDLE IRQ_BREAK
+      ============== ========== ====== ========== =========
+      CC3200                      yes
+      ESP32            yes        yes                yes
+      MIMXRT           yes                yes
+      NRF                         yes     yes
+      RENESAS-RA       yes        yes
+      RP2              yes                yes        yes
+      SAMD             yes        yes     yes
+      STM32            yes        yes
+      ============== ========== ====== ========== =========
+
+
+   .. note::
+     - The ESP32 port does not support the option hard=True.
+
+     - The rp2 port's UART.IRQ_TXIDLE is only triggered when the message
+       is longer than 5 characters and the trigger happens when still 5 characters
+       are to be sent.
+
+     - The rp2 port's UART.IRQ_BREAK needs receiving valid characters for triggering
+       again.
+
+     - The SAMD port's UART.IRQ_TXIDLE is triggered while the last character is sent.
+
+     - On STM32F4xx MCU's, using the trigger UART.IRQ_RXIDLE the handler will be called once
+       after the first character and then after the end of the message, when the line is
+       idle.
+
+
+   Availability: cc3200, esp32, mimxrt, nrf, renesas-ra, rp2, samd, stm32.
 
 Constants
 ---------
@@ -213,7 +253,10 @@ Constants
     Availability: esp32, mimxrt, renesas-ra, rp2, stm32.
 
 .. data:: UART.IRQ_RXIDLE
+          UART.IRQ_RX
+          UART.IRQ_TXIDLE
+          UART.IRQ_BREAK
 
     IRQ trigger sources.
 
-    Availability: stm32.
+    Availability: renesas-ra, stm32, esp32, rp2040, mimxrt, samd, cc3200.
--- a/ports/esp32/machine_timer.c
+++ b/ports/esp32/machine_timer.c
@@ -38,6 +38,7 @@
 #include "hal/timer_hal.h"
 #include "hal/timer_ll.h"
 #include "soc/timer_periph.h"
+#include "machine_timer.h"
 
 #define TIMER_DIVIDER  8
 
@@ -46,27 +47,8 @@
 
 #define TIMER_FLAGS    0
 
-typedef struct _machine_timer_obj_t {
-    mp_obj_base_t base;
-
-    timer_hal_context_t hal_context;
-    mp_uint_t group;
-    mp_uint_t index;
-
-    mp_uint_t repeat;
-    // ESP32 timers are 64 or 54-bit
-    uint64_t period;
-
-    mp_obj_t callback;
-
-    intr_handle_t handle;
-
-    struct _machine_timer_obj_t *next;
-} machine_timer_obj_t;
-
 const mp_obj_type_t machine_timer_type;
 
-static void machine_timer_disable(machine_timer_obj_t *self);
 static mp_obj_t machine_timer_init_helper(machine_timer_obj_t *self, mp_uint_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args);
 
 void machine_timer_deinit_all(void) {
@@ -91,17 +73,16 @@ static void machine_timer_print(const mp_print_t *print, mp_obj_t self_in, mp_pr
     #endif
 }
 
-static mp_obj_t machine_timer_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
-    mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
-    #if CONFIG_IDF_TARGET_ESP32C3
-    mp_uint_t group = mp_obj_get_int(args[0]) & 1;
-    mp_uint_t index = 0;
-    #else
-    mp_uint_t group = (mp_obj_get_int(args[0]) >> 1) & 1;
-    mp_uint_t index = mp_obj_get_int(args[0]) & 1;
-    #endif
+machine_timer_obj_t *machine_timer_create(mp_uint_t timer) {
 
     machine_timer_obj_t *self = NULL;
+    #if CONFIG_IDF_TARGET_ESP32C3
+    mp_uint_t group = timer & 1;
+    mp_uint_t index = 0;
+    #else
+    mp_uint_t group = (timer >> 1) & 1;
+    mp_uint_t index = timer & 1;
+    #endif
 
     // Check whether the timer is already initialized, if so use it
     for (machine_timer_obj_t *t = MP_STATE_PORT(machine_timer_obj_head); t; t = t->next) {
@@ -120,6 +101,14 @@ static mp_obj_t machine_timer_make_new(const mp_obj_type_t *type, size_t n_args,
         self->next = MP_STATE_PORT(machine_timer_obj_head);
         MP_STATE_PORT(machine_timer_obj_head) = self;
     }
+    return self;
+}
+
+static mp_obj_t machine_timer_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
+
+    // Create the new timer.
+    machine_timer_obj_t *self = machine_timer_create(mp_obj_get_int(args[0]));
 
     if (n_args > 1 || n_kw > 0) {
         mp_map_t kw_args;
@@ -130,7 +119,7 @@ static mp_obj_t machine_timer_make_new(const mp_obj_type_t *type, size_t n_args,
     return self;
 }
 
-static void machine_timer_disable(machine_timer_obj_t *self) {
+void machine_timer_disable(machine_timer_obj_t *self) {
     if (self->hal_context.dev != NULL) {
         // Disable the counter and alarm.
         timer_ll_enable_counter(self->hal_context.dev, self->index, false);
@@ -162,7 +151,7 @@ static void machine_timer_isr(void *self_in) {
     }
 }
 
-static void machine_timer_enable(machine_timer_obj_t *self) {
+void machine_timer_enable(machine_timer_obj_t *self, void (*timer_isr)) {
     // Initialise the timer.
     timer_hal_init(&self->hal_context, self->group, self->index);
     timer_ll_enable_counter(self->hal_context.dev, self->index, false);
@@ -176,7 +165,7 @@ static void machine_timer_enable(machine_timer_obj_t *self) {
     timer_ll_clear_intr_status(self->hal_context.dev, TIMER_LL_EVENT_ALARM(self->index));
     ESP_ERROR_CHECK(
         esp_intr_alloc(timer_group_periph_signals.groups[self->group].timer_irq_id[self->index],
-            TIMER_FLAGS, machine_timer_isr, self, &self->handle)
+            TIMER_FLAGS, timer_isr, self, &self->handle)
         );
     timer_ll_enable_intr(self->hal_context.dev, TIMER_LL_EVENT_ALARM(self->index), true);
 
@@ -234,7 +223,7 @@ static mp_obj_t machine_timer_init_helper(machine_timer_obj_t *self, mp_uint_t n
     self->callback = args[ARG_callback].u_obj;
     self->handle = NULL;
 
-    machine_timer_enable(self);
+    machine_timer_enable(self, machine_timer_isr);
 
     return mp_const_none;
 }
--- a/ports/mimxrt/hal/fsl_lpuart.c
+++ b/ports/mimxrt/hal/fsl_lpuart.c
@@ -1,0 +1,2013 @@
+/*
+ * Copyright (c) 2015-2016, Freescale Semiconductor, Inc.
+ * Copyright 2016-2020 NXP
+ * All rights reserved.
+ *
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+#include "fsl_lpuart.h"
+
+/*******************************************************************************
+ * Definitions
+ ******************************************************************************/
+
+/* Component ID definition, used by tools. */
+#ifndef FSL_COMPONENT_ID
+#define FSL_COMPONENT_ID "platform.drivers.lpuart"
+#endif
+
+/* LPUART transfer state. */
+enum
+{
+    kLPUART_TxIdle, /*!< TX idle. */
+    kLPUART_TxBusy, /*!< TX busy. */
+    kLPUART_RxIdle, /*!< RX idle. */
+    kLPUART_RxBusy  /*!< RX busy. */
+};
+
+/*******************************************************************************
+ * Prototypes
+ ******************************************************************************/
+/*!
+ * @brief Check whether the RX ring buffer is full.
+ *
+ * @userData handle LPUART handle pointer.
+ * @retval true  RX ring buffer is full.
+ * @retval false RX ring buffer is not full.
+ */
+static bool LPUART_TransferIsRxRingBufferFull(LPUART_Type *base, lpuart_handle_t *handle);
+
+/*!
+ * @brief Write to TX register using non-blocking method.
+ *
+ * This function writes data to the TX register directly, upper layer must make
+ * sure the TX register is empty or TX FIFO has empty room before calling this function.
+ *
+ * @note This function does not check whether all the data has been sent out to bus,
+ * so before disable TX, check kLPUART_TransmissionCompleteFlag to ensure the TX is
+ * finished.
+ *
+ * @param base LPUART peripheral base address.
+ * @param data Start address of the data to write.
+ * @param length Size of the buffer to be sent.
+ */
+static void LPUART_WriteNonBlocking(LPUART_Type *base, const uint8_t *data, size_t length);
+
+/*!
+ * @brief Read RX register using non-blocking method.
+ *
+ * This function reads data from the TX register directly, upper layer must make
+ * sure the RX register is full or TX FIFO has data before calling this function.
+ *
+ * @param base LPUART peripheral base address.
+ * @param data Start address of the buffer to store the received data.
+ * @param length Size of the buffer.
+ */
+static void LPUART_ReadNonBlocking(LPUART_Type *base, uint8_t *data, size_t length);
+
+/*******************************************************************************
+ * Variables
+ ******************************************************************************/
+/* Array of LPUART peripheral base address. */
+static LPUART_Type *const s_lpuartBases[] = LPUART_BASE_PTRS;
+/* Array of LPUART handle. */
+void *s_lpuartHandle[ARRAY_SIZE(s_lpuartBases)];
+/* Array of LPUART IRQ number. */
+#if defined(FSL_FEATURE_LPUART_HAS_SEPARATE_RX_TX_IRQ) && FSL_FEATURE_LPUART_HAS_SEPARATE_RX_TX_IRQ
+static const IRQn_Type s_lpuartRxIRQ[] = LPUART_RX_IRQS;
+const IRQn_Type s_lpuartTxIRQ[] = LPUART_TX_IRQS;
+#else
+const IRQn_Type s_lpuartIRQ[] = LPUART_RX_TX_IRQS;
+#endif
+#if !(defined(FSL_SDK_DISABLE_DRIVER_CLOCK_CONTROL) && FSL_SDK_DISABLE_DRIVER_CLOCK_CONTROL)
+/* Array of LPUART clock name. */
+static const clock_ip_name_t s_lpuartClock[] = LPUART_CLOCKS;
+
+#if defined(LPUART_PERIPH_CLOCKS)
+/* Array of LPUART functional clock name. */
+static const clock_ip_name_t s_lpuartPeriphClocks[] = LPUART_PERIPH_CLOCKS;
+#endif
+
+#endif /* FSL_SDK_DISABLE_DRIVER_CLOCK_CONTROL */
+
+/* LPUART ISR for transactional APIs. */
+#if defined(__ARMCC_VERSION) && (__ARMCC_VERSION >= 6010050)
+lpuart_isr_t s_lpuartIsr = (lpuart_isr_t)DefaultISR;
+#else
+lpuart_isr_t s_lpuartIsr;
+#endif
+
+/*******************************************************************************
+ * Code
+ ******************************************************************************/
+/*!
+ * brief Get the LPUART instance from peripheral base address.
+ *
+ * param base LPUART peripheral base address.
+ * return LPUART instance.
+ */
+uint32_t LPUART_GetInstance(LPUART_Type *base) {
+    uint32_t instance;
+
+    /* Find the instance index from base address mappings. */
+    for (instance = 0U; instance < ARRAY_SIZE(s_lpuartBases); instance++) {
+        if (s_lpuartBases[instance] == base) {
+            break;
+        }
+    }
+
+    assert(instance < ARRAY_SIZE(s_lpuartBases));
+
+    return instance;
+}
+
+/*!
+ * brief Get the length of received data in RX ring buffer.
+ *
+ * userData handle LPUART handle pointer.
+ * return Length of received data in RX ring buffer.
+ */
+size_t LPUART_TransferGetRxRingBufferLength(LPUART_Type *base, lpuart_handle_t *handle) {
+    assert(NULL != handle);
+
+    size_t size;
+    size_t tmpRxRingBufferSize = handle->rxRingBufferSize;
+    uint16_t tmpRxRingBufferTail = handle->rxRingBufferTail;
+    uint16_t tmpRxRingBufferHead = handle->rxRingBufferHead;
+
+    if (tmpRxRingBufferTail > tmpRxRingBufferHead) {
+        size = ((size_t)tmpRxRingBufferHead + tmpRxRingBufferSize - (size_t)tmpRxRingBufferTail);
+    } else {
+        size = ((size_t)tmpRxRingBufferHead - (size_t)tmpRxRingBufferTail);
+    }
+
+    return size;
+}
+
+static bool LPUART_TransferIsRxRingBufferFull(LPUART_Type *base, lpuart_handle_t *handle) {
+    assert(NULL != handle);
+
+    bool full;
+
+    if (LPUART_TransferGetRxRingBufferLength(base, handle) == (handle->rxRingBufferSize - 1U)) {
+        full = true;
+    } else {
+        full = false;
+    }
+    return full;
+}
+
+static void LPUART_WriteNonBlocking(LPUART_Type *base, const uint8_t *data, size_t length) {
+    assert(NULL != data);
+
+    size_t i;
+
+    /* The Non Blocking write data API assume user have ensured there is enough space in
+    peripheral to write. */
+    for (i = 0; i < length; i++) {
+        base->DATA = data[i];
+    }
+}
+
+static void LPUART_ReadNonBlocking(LPUART_Type *base, uint8_t *data, size_t length) {
+    assert(NULL != data);
+
+    size_t i;
+    #if defined(FSL_FEATURE_LPUART_HAS_7BIT_DATA_SUPPORT) && FSL_FEATURE_LPUART_HAS_7BIT_DATA_SUPPORT
+    uint32_t ctrl = base->CTRL;
+    bool isSevenDataBits = (((ctrl & LPUART_CTRL_M7_MASK) != 0U) ||
+        (((ctrl & LPUART_CTRL_M_MASK) == 0U) && ((ctrl & LPUART_CTRL_PE_MASK) != 0U)));
+    #endif
+
+    /* The Non Blocking read data API assume user have ensured there is enough space in
+    peripheral to write. */
+    for (i = 0; i < length; i++) {
+        #if defined(FSL_FEATURE_LPUART_HAS_7BIT_DATA_SUPPORT) && FSL_FEATURE_LPUART_HAS_7BIT_DATA_SUPPORT
+        if (isSevenDataBits) {
+            data[i] = (uint8_t)(base->DATA & 0x7FU);
+        } else {
+            data[i] = (uint8_t)base->DATA;
+        }
+        #else
+        data[i] = (uint8_t)(base->DATA);
+        #endif
+    }
+}
+
+/*!
+ * brief Initializes an LPUART instance with the user configuration structure and the peripheral clock.
+ *
+ * This function configures the LPUART module with user-defined settings. Call the LPUART_GetDefaultConfig() function
+ * to configure the configuration structure and get the default configuration.
+ * The example below shows how to use this API to configure the LPUART.
+ * code
+ *  lpuart_config_t lpuartConfig;
+ *  lpuartConfig.baudRate_Bps = 115200U;
+ *  lpuartConfig.parityMode = kLPUART_ParityDisabled;
+ *  lpuartConfig.dataBitsCount = kLPUART_EightDataBits;
+ *  lpuartConfig.isMsb = false;
+ *  lpuartConfig.stopBitCount = kLPUART_OneStopBit;
+ *  lpuartConfig.txFifoWatermark = 0;
+ *  lpuartConfig.rxFifoWatermark = 1;
+ *  LPUART_Init(LPUART1, &lpuartConfig, 20000000U);
+ * endcode
+ *
+ * param base LPUART peripheral base address.
+ * param config Pointer to a user-defined configuration structure.
+ * param srcClock_Hz LPUART clock source frequency in HZ.
+ * retval kStatus_LPUART_BaudrateNotSupport Baudrate is not support in current clock source.
+ * retval kStatus_Success LPUART initialize succeed
+ */
+status_t LPUART_Init(LPUART_Type *base, const lpuart_config_t *config, uint32_t srcClock_Hz) {
+    assert(NULL != config);
+    assert(0U < config->baudRate_Bps);
+    #if defined(FSL_FEATURE_LPUART_HAS_FIFO) && FSL_FEATURE_LPUART_HAS_FIFO
+    assert((uint8_t)FSL_FEATURE_LPUART_FIFO_SIZEn(base) >= config->txFifoWatermark);
+    assert((uint8_t)FSL_FEATURE_LPUART_FIFO_SIZEn(base) >= config->rxFifoWatermark);
+    #endif
+
+    status_t status = kStatus_Success;
+    uint32_t temp;
+    uint16_t sbr, sbrTemp;
+    uint8_t osr, osrTemp;
+    uint32_t tempDiff, calculatedBaud, baudDiff;
+
+    /* This LPUART instantiation uses a slightly different baud rate calculation
+     * The idea is to use the best OSR (over-sampling rate) possible
+     * Note, OSR is typically hard-set to 16 in other LPUART instantiations
+     * loop to find the best OSR value possible, one that generates minimum baudDiff
+     * iterate through the rest of the supported values of OSR */
+
+    baudDiff = config->baudRate_Bps;
+    osr = 0U;
+    sbr = 0U;
+    for (osrTemp = 4U; osrTemp <= 32U; osrTemp++) {
+        /* calculate the temporary sbr value   */
+        sbrTemp = (uint16_t)((srcClock_Hz * 10U / (config->baudRate_Bps * (uint32_t)osrTemp) + 5U) / 10U);
+        /*set sbrTemp to 1 if the sourceClockInHz can not satisfy the desired baud rate*/
+        if (sbrTemp == 0U) {
+            sbrTemp = 1U;
+        }
+        /* Calculate the baud rate based on the temporary OSR and SBR values */
+        calculatedBaud = (srcClock_Hz / ((uint32_t)osrTemp * (uint32_t)sbrTemp));
+        tempDiff = calculatedBaud > config->baudRate_Bps ? (calculatedBaud - config->baudRate_Bps) :
+            (config->baudRate_Bps - calculatedBaud);
+
+        if (tempDiff <= baudDiff) {
+            baudDiff = tempDiff;
+            osr = osrTemp;      /* update and store the best OSR value calculated */
+            sbr = sbrTemp;      /* update store the best SBR value calculated */
+        }
+    }
+
+    /* Check to see if actual baud rate is within 3% of desired baud rate
+     * based on the best calculate OSR value */
+    if (baudDiff > ((config->baudRate_Bps / 100U) * 3U)) {
+        /* Unacceptable baud rate difference of more than 3%*/
+        status = kStatus_LPUART_BaudrateNotSupport;
+    } else {
+        #if !(defined(FSL_SDK_DISABLE_DRIVER_CLOCK_CONTROL) && FSL_SDK_DISABLE_DRIVER_CLOCK_CONTROL)
+
+        uint32_t instance = LPUART_GetInstance(base);
+
+        /* Enable lpuart clock */
+        (void)CLOCK_EnableClock(s_lpuartClock[instance]);
+        #if defined(LPUART_PERIPH_CLOCKS)
+        (void)CLOCK_EnableClock(s_lpuartPeriphClocks[instance]);
+        #endif
+
+        #endif /* FSL_SDK_DISABLE_DRIVER_CLOCK_CONTROL */
+
+        #if defined(FSL_FEATURE_LPUART_HAS_GLOBAL) && FSL_FEATURE_LPUART_HAS_GLOBAL
+        /*Reset all internal logic and registers, except the Global Register */
+        LPUART_SoftwareReset(base);
+        #else
+        /* Disable LPUART TX RX before setting. */
+        base->CTRL &= ~(LPUART_CTRL_TE_MASK | LPUART_CTRL_RE_MASK);
+        #endif
+
+        temp = base->BAUD;
+
+        /* Acceptable baud rate, check if OSR is between 4x and 7x oversampling.
+         * If so, then "BOTHEDGE" sampling must be turned on */
+        if ((osr > 3U) && (osr < 8U)) {
+            temp |= LPUART_BAUD_BOTHEDGE_MASK;
+        }
+
+        /* program the osr value (bit value is one less than actual value) */
+        temp &= ~LPUART_BAUD_OSR_MASK;
+        temp |= LPUART_BAUD_OSR((uint32_t)osr - 1UL);
+
+        /* write the sbr value to the BAUD registers */
+        temp &= ~LPUART_BAUD_SBR_MASK;
+        base->BAUD = temp | LPUART_BAUD_SBR(sbr);
+
+        /* Set bit count and parity mode. */
+        base->BAUD &= ~LPUART_BAUD_M10_MASK;
+
+        temp = base->CTRL & ~(LPUART_CTRL_PE_MASK | LPUART_CTRL_PT_MASK | LPUART_CTRL_M_MASK | LPUART_CTRL_ILT_MASK |
+            LPUART_CTRL_IDLECFG_MASK);
+
+        temp |= (uint8_t)config->parityMode | LPUART_CTRL_IDLECFG(config->rxIdleConfig) |
+            LPUART_CTRL_ILT(config->rxIdleType);
+
+        #if defined(FSL_FEATURE_LPUART_HAS_7BIT_DATA_SUPPORT) && FSL_FEATURE_LPUART_HAS_7BIT_DATA_SUPPORT
+        if (kLPUART_SevenDataBits == config->dataBitsCount) {
+            if (kLPUART_ParityDisabled != config->parityMode) {
+                temp &= ~LPUART_CTRL_M7_MASK; /* Seven data bits and one parity bit */
+            } else {
+                temp |= LPUART_CTRL_M7_MASK;
+            }
+        } else
+        #endif
+        {
+            if (kLPUART_ParityDisabled != config->parityMode) {
+                temp |= LPUART_CTRL_M_MASK; /* Eight data bits and one parity bit */
+            }
+        }
+
+        base->CTRL = temp;
+
+        #if defined(FSL_FEATURE_LPUART_HAS_STOP_BIT_CONFIG_SUPPORT) && FSL_FEATURE_LPUART_HAS_STOP_BIT_CONFIG_SUPPORT
+        /* set stop bit per char */
+        temp = base->BAUD & ~LPUART_BAUD_SBNS_MASK;
+        base->BAUD = temp | LPUART_BAUD_SBNS((uint8_t)config->stopBitCount);
+        #endif
+
+        #if defined(FSL_FEATURE_LPUART_HAS_FIFO) && FSL_FEATURE_LPUART_HAS_FIFO
+        /* Set tx/rx WATER watermark
+           Note:
+           Take care of the RX FIFO, RX interrupt request only assert when received bytes
+           equal or more than RX water mark, there is potential issue if RX water
+           mark larger than 1.
+           For example, if RX FIFO water mark is 2, upper layer needs 5 bytes and
+           5 bytes are received. the last byte will be saved in FIFO but not trigger
+           RX interrupt because the water mark is 2.
+         */
+        base->WATER = (((uint32_t)(config->rxFifoWatermark) << 16U) | config->txFifoWatermark);
+
+        /* Enable tx/rx FIFO */
+        base->FIFO |= (LPUART_FIFO_TXFE_MASK | LPUART_FIFO_RXFE_MASK);
+
+        /* Flush FIFO */
+        base->FIFO |= (LPUART_FIFO_TXFLUSH_MASK | LPUART_FIFO_RXFLUSH_MASK);
+        #endif
+
+        /* Clear all status flags */
+        temp = (LPUART_STAT_RXEDGIF_MASK | LPUART_STAT_IDLE_MASK | LPUART_STAT_OR_MASK | LPUART_STAT_NF_MASK |
+            LPUART_STAT_FE_MASK | LPUART_STAT_PF_MASK);
+
+        #if defined(FSL_FEATURE_LPUART_HAS_LIN_BREAK_DETECT) && FSL_FEATURE_LPUART_HAS_LIN_BREAK_DETECT
+        temp |= LPUART_STAT_LBKDIF_MASK;
+        #endif
+
+        #if defined(FSL_FEATURE_LPUART_HAS_ADDRESS_MATCHING) && FSL_FEATURE_LPUART_HAS_ADDRESS_MATCHING
+        temp |= (LPUART_STAT_MA1F_MASK | LPUART_STAT_MA2F_MASK);
+        #endif
+
+        #if defined(FSL_FEATURE_LPUART_HAS_MODEM_SUPPORT) && FSL_FEATURE_LPUART_HAS_MODEM_SUPPORT
+        /* Set the CTS configuration/TX CTS source. */
+        base->MODIR |= LPUART_MODIR_TXCTSC(config->txCtsConfig) | LPUART_MODIR_TXCTSSRC(config->txCtsSource);
+        if (true == config->enableRxRTS) {
+            /* Enable the receiver RTS(request-to-send) function. */
+            base->MODIR |= LPUART_MODIR_RXRTSE_MASK;
+        }
+        if (true == config->enableTxCTS) {
+            /* Enable the CTS(clear-to-send) function. */
+            base->MODIR |= LPUART_MODIR_TXCTSE_MASK;
+        }
+        #endif
+
+        /* Set data bits order. */
+        if (true == config->isMsb) {
+            temp |= LPUART_STAT_MSBF_MASK;
+        } else {
+            temp &= ~LPUART_STAT_MSBF_MASK;
+        }
+
+        base->STAT |= temp;
+
+        /* Enable TX/RX base on configure structure. */
+        temp = base->CTRL;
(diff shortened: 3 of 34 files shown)


dpgeorge:
Thanks for this, it's a great addition!  Definitely needed.

By chance I was working on the docs side of this today: #14040.

robert-hh:
A few discussion topics:
- Some alternative names for the triggers:
  IRQ_RXANY instead of IRQ_RX
  IRQ_TXDONE instead of IRQ_TXIDLE

- All but the ESP32 port have the same section of code now, which implement the UART.irq  method. I wonder if that could not be moved for example into extmod/machine_uart.c.

dpgeorge:
Thanks again for working on this.  After thinking more about it here are my general comments.

I don't think it makes sense to have both `IRQ_RX` and `IRQ_RXIDLE`, especially because it seems that the ports either implement one or the other, but never both.

We need to make it as easy as possible to write portable code that runs across all the ports, and having different IRQs, or the same IRQ but with slightly different semantics, makes that hard.

So I suggest to not have `IRQ_RX` at all, and make `IRQ_RXIDLE` mandatory.  Users should be able to rely on thisIRQ to have the following semantics:

    IRQ_RXIDLE: every character received by the UART will have this interrupt
    triggered at some point in time after that character is received

The idea is that if you only ever do `uart.read()` within the `IRQ_RXIDLE` callback, then you should always be able to (eventually) read all available characters.  And ideally in a timely manner.

Even if the hardware/SDK doesn't support this concept we should still be able to synthesize it.  Eg if the hardware supports `IRQ_RX` then in combination with a timer we can implement `IRQ_RXIDLE`.

Having `IRQ_BREAK` (when the hardware supports it) makes sense, the semantics for that seem obvious.

For `IRQ_TXIDLE`/`IRQ_TXDONE`: as with `IRQ_RXIDLE` the semantics of this need to be well specified, and then all ports should implement it.

robert-hh:
So you suggest that IRQ_RX is renamed to IRQ_RXIDLE for those ports that do not support a true IRQ_RXIDLE, and for those that do so IRQ_RX is dropped?

B.t.w: I made two Python classes as example that implement IRQ_RXIDLE based on an existing IRQ_RX. The simpler one is below with uart.irq() being a method. The other implements uart.irq() as a class. I considered implementing that in the machine_uart module itself, but dealing with the timer is easier in Python.
```
#
# Simple wrapper class for UART, adding the trigger IRQ_RXIDLE for port
# which just support IRQ_RX
# Missing: Handling of the irq.xxx methods in this class.
#
import machine

class UART(machine.UART):
    def __init__(self, id, *args, **kwargs):
        self.irq_timeout_count = 0
        self.irq_rx_count = 0
        self.timeout_idle = 5
        self.irq_timer = None
        self.irq_handler = None
        self.IRQ_RXIDLE = 4096

    def irq(self, handler=None, trigger=None, hard=False, period=5):
        if handler is not None:
            if trigger == self.IRQ_RXIDLE:
                self.irq_handler = handler
                self.irq_timer = machine.Timer(callback=self.irq_timer_callback,
                    mode=machine.Timer.PERIODIC, period=period)
                return super().irq(handler=self.irq_rx_callback, trigger=self.IRQ_RX, hard=True)
            else:
                if self.irq_timer is not None:
                    self.irq_timer.deinit()
                self.irq_handler = None
                return super().irq(handler=handler, trigger=trigger, hard=hard)
        else:
            if self.irq_timer is not None:
                self.irq_timer.deinit()
            self.irq_handler = None
            return super().irq(handler=None, trigger=0, hard=False)

    def irq_rx_callback(self, uart):
        self.irq_rx_count += 1
        self.irq_timeout_count = 0

    def irq_timer_callback(self, timer):
        if self.irq_rx_count > 0:
            self.irq_timeout_count += 1
            if self.irq_timeout_count == 2:
                if self.irq_handler is not None:
                    self.irq_handler(self)
                self.irq_rx_count = 0
                self.irq_timeout_count = 0
```

robert-hh:
@dpgeorge Thinking about further, I prefer to support the behaviour with proper names. Looking at the list of ports, three of them support true IRQ_RXIDLE, and eight support IRQ_RX. Both the rp2040 and the MIMXRT port can support IRQ_RX mode. The rp2040 commit would get much simpler and more close to the existing state. Without support for IRQ_RXIDLE, the MIMXRT port would not need a change to the NXP UART driver. 
So I suggest to have IRQ_RX as the common method and have support for IRQ_RXIDLE as a derived class method.

dpgeorge:
> So I suggest to have IRQ_RX as the common method and have support for IRQ_RXIDLE as a derived class method.

I understand it's simpler to implement what the hardware provides.  But:
- IRQ_RX is not that useful for large data rates (as you already mentioned).  So it will be impossible to implement IRQ_RXIDLE behaviour in Python using IRQ_RX (because IRQ_RX will be called too fast for Python to keep up with it).  Even if the implementation does work, it won't be efficient.
- We need to provide something portable that works the same across all ports.  Eg so the `network.PPP` driver can work and use the UART IRQ.

robert-hh:
For the existing PR is see then two options:
a) Drop the changes for IRQ_RX and keep only the changes for RP2040 and MIMXRT for IRQ_RXIDLE. More boards may be added later it a IRQ_RXIDLE is implemented based on an internal IRQ_RX and a timer. Maybe the easiest way would be to compare the ms time stamp at an char RX with that of the previous char RX.
b) Rename IRQ_RX to IRQ_RXIDLE and have a note in the documentation about the difference to real IRQ_RXIDLE.

Obviously the behavior of RXIDLE can as well be implemented in Python with a timer interrupt, which checks if uart.any() values stall at a value > 0.

projectgus (commented):
Left a couple of small comments in the docs, and agree about thinking through the most useful RX interrupt behaviour, but overall this looks very useful. Thanks for adding it on so many ports, @robert-hh!

dpgeorge:
After discussion with @jimmo and @projectgus we came to the conclusion that it makes sense to expose multiple UART IRQs, and then the user can decide which is best for their application.  But we do need to have consistence across ports for the behaviour of the interrupts.

- IRQ_RX: all ports will be able to implement this in hardware (I assume?) and it's a pretty useful interrupt to have access to, although only for slow baud rates.
- IRQ_RXFIFO: for when the RX input buffer is half full.  This can be easily synthesized (in C) as long as IRQ_RX is available.  Note that we don't need to implement this one yet but there is scope to add it in the future.
- IRQ_RXIDLE: when the RX line remains high for whole character period after receiving at least one character.  This can be synthesized (in C) for ports that don't have hardware support for it.
- IRQ_BREAK: useful and obvious semantics.
- IRQ_TXIDLE: not so sure about this one, it's a bit harder to define and we don't yet have consistency across ports for TX, whether it's blocking or runs in the background.

So in the end, based on the above points, I think the PR as it stands is in pretty good shape!  The main thing is to have good documentation that specifies the behaviour of each IRQ well, and makes sure the implementations for each port match that behaviour so it can be relied on.

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+   Configure an interrupt handler to be called when a UART event occurs.
+
+   The arguments are:
+
+     - *handler* is an optional function to be called when the interrupt event
+       triggers.  The handler must take exactly one argument which is the
+       ``UART`` instance.
+
+     - *trigger* configures the event which can generate an interrupt.
+       Possible values are:
```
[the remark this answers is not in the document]
robert-hh: Thanks. Updated accordingly.

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+
+       - ``UART.IRQ_RXIDLE`` interrupt after receiving at least one character
+         and then the RX line goes idle.
+       - ``UART.IRQ_RX`` interrupt after each received character.
+       - ``UART.TX_IDLE`` interrupt after or while the last character(s) of
+         a message are or have been sent.
+       - ``UART.BREAK`` interrupt when a break state is detected at RX
```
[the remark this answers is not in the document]
robert-hh: Thanks. Updated accordingly. In the code, these names were used.

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+
+   Returns an irq object.
+
+   Due to limitations of the hardware not all trigger events are available at all ports.
+
+   .. table:: Availability of triggers
+      :align: center 
+
+      ============== ======= === ====== =====
+      Port / Trigger RX_IDLE RX  TXIDLE BREAK
```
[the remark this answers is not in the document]
robert-hh: Thanks. Updated accordingly

robert-hh:
Thank you for the confirmation. Some notes:

> IRQ_RX: all ports will be able to implement this in hardware

Actually, at the RP2040 and MIMXRT port one has to chose between IRQ_RX and IRQ_RXIDLE. For the RP2040, it's the way interrupts are created and the FIFO is working. Unfortunately it is not possible to tell how many bytes are in the FIFO. And IRQ_TXIDLE happens only if at least one byte is in the FIFO. For that reason, the FIFO full IRQ is set to 75% (24 Byte), and in an interrupt not more than 23 bytes are removed from the FIFO. read() takes all data from the FIFO. but with an IRQ_RX at 24Byte this is not so useful. The MIMXRT UART driver has no IRQ_RX event, since that is all handled in the driver itself. But I can check again if some of the other Event match.

> IRQ_RXFIFO: for when the RX input buffer is half full.

That looks doable to some extend for most ports, if the RX ring buffer is taken as reference.

> IRQ_RXIDLE: when the RX line remains high for whole character period after receiving at least one character. This can be synthesized (in C) for ports that don't have hardware support for it.

Principally yes. The challenge will be to use the right timer mechanism with little overhead. The time the RX line has to stay high is 30 bit times for RP2040, which is hardwired, and for MIMXRT I have configured it to 4 byte times starting from the most recent start bit, resulting in about 30 bit times after the last byte. If I recall right, the STM32 timing is similar.
The Python code above works well for slow communication like that from a GPS receiver or other sensors. So it has it's use case.

> IRQ_BREAK: useful and obvious semantics.

I looked at the datasheets for all eight ports. For only two ports I found a hardware interrupt or signalling mechanism for break. I'm still wondering why this is not available at the MIMXRT or STM32 port.

> IRQ_TXIDLE: not so sure about this one, it's a bit harder to define and we don't yet have consistency across ports for TX, whether it's blocking or runs in the background.

Yes. This one should happen when all data of a "message" is sent. But yes, if the driver buffers internally without being able to tell, that could be tricky. And we have already uart.flush() and uart.txdone(), which could be used for that purpose, with the same limitation.

I updated the documentation according to Angus' comments.

robert-hh:
The PR is now updated with a  sample implementation of IRQ_RXIDLE for the SAMD port using softtimer. The same approach can be used for the Renesas port. 
For that purpose, the softtimer data structure is extended by an additional pointer, which can port at context data for the timer interrupt. In this case it's the respective UART data structure. That allows to have a single timer callback and not a separate callback for each UART. At the moment that does not interfere with any other use of softtimer. To make is consistent, the function softtimer_static_init() may get another argument, the context pointer. That requires to change all call of that function. Not a big thing, but I wanted to wait until the approach of a context pointer is accepted.

shared/runtime/softtimer.h
```diff
@@ -44,6 +44,7 @@ typedef struct _soft_timer_entry_t {
         void (*c_callback)(struct _soft_timer_entry_t *);
         mp_obj_t py_callback;
     };
+    void *context;
```
dpgeorge, line 47: To avoid adding this to all users of this struct it might be possible to instead create a derived struct like this:
```c
typedef struct _soft_timer_entry_extended_t {
    soft_timer_entry_t base;
    void *context;
} soft_timer_entry_extended_t;
```
Then you can pass `&entry->base` to the existing soft-timer functions, but when you get a callback you can cast the `soft_timer_entry_t*` back to a `soft_timer_entry_extended_t*` and access the context.

robert-hh: Thanks for the hint. The code is changed now and tested with the SAMD and Renesas port, which are affected by the change.
I could not test with the Pico2, since I have no board yet. But when adding the RP2 commit to the RP2350 PR, it still builds fine.

ports/nrf/boards/ARDUINO_NANO_33_BLE_SENSE/mpconfigboard.h
```diff
@@ -17,6 +17,7 @@
 #define MICROPY_PY_MACHINE_I2C          (1)
 #define MICROPY_PY_MACHINE_ADC          (1)
 #define MICROPY_PY_MACHINE_TEMP         (1)
+#define MICROPY_PY_MACHINE_UART_IRQ     (1)
```
dpgeorge, line 20: Is there a way to automatically set this option in `mpconfigport.h`, so it doesn't need to be done per board?  For example, all nRF52 boards get it, or use `EXTRA_FEAT`.

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+       again.
+
+     - The ESP32 port does not support the option hard=True.
+
+     - At the STM32F4xx using the trigger UART.IRQ_RXIDLE the handler will be called once
+       after the first character and then after the end of the message, when the line is
+       idle.
+
+
+   Availability: renesas-ra, stm32, esp32, rp2040, mimxrt, samd, nrf, cc3200.
```
dpgeorge, line 243: Please sort these in alphabetical order.

robert-hh: Done

ports/nrf/boards/ARDUINO_NANO_33_BLE_SENSE/mpconfigboard.h
robert-hh: It can be based on the #define of `NRF52`. But there is also a define for `MICROPY_PY_MACHINE_UART` in every board file. That could be moved to mpconfigport.h as well, especially since the uart includefile is set in mpconfigport.h.

dpgeorge:
@robert-hh I have added tests for all the IRQs added in this PR.  I pushed those tests to your branch.  Feel free to modify the tests, eg add more configurations for other ports.

Each test must have configuration for a given port, eg the UART id and UART pins.  Some tests can run without extra hardware connections, but others require connecting TX and RX.  The tests are documented at their top to say what they need.

I ran the tests on a few different ports (stm32, rp2, esp32, samd).  Not all of them pass.  It would be great if all tests could pass on all ports (or skip if the port doesn't support that IRQ), but I know that could be difficult to achieve.  If you have some time to look into making them pass that would be fantastic!

--------

For example, running the tests on various hardware:

#### ESP32

```
$ ./run-tests.py --target minimal --device /dev/ttyUSB0 extmod/machine_uart*.py extmod_hardware/machine_uart*.py
skip  extmod/machine_uart_irq_txidle.py
skip  extmod/machine_uart_tx.py
FAIL  extmod_hardware/machine_uart_irq_break.py
pass  extmod_hardware/machine_uart_irq_rxidle.py
pass  extmod_hardware/machine_uart_irq_rx.py
3 tests performed (36 individual testcases)
2 tests passed
2 tests skipped: machine_uart_irq_txidle machine_uart_tx
1 tests failed: machine_uart_irq_break
```

#### SAMD21

```
$ ./run-tests.py --target minimal extmod/machine_uart*.py extmod_hardware/machine_uart*.py             
FAIL  extmod/machine_uart_irq_txidle.py
skip  extmod/machine_uart_tx.py
skip  extmod_hardware/machine_uart_irq_break.py
pass  extmod_hardware/machine_uart_irq_rxidle.py
FAIL  extmod_hardware/machine_uart_irq_rx.py
3 tests performed (30 individual testcases)
1 tests passed
2 tests skipped: machine_uart_irq_break machine_uart_tx
2 tests failed: machine_uart_irq_rx machine_uart_irq_txidle
```

#### rp2 (Pico)

```
$ ./run-tests.py --target minimal extmod/machine_uart_irq*.py extmod_hardware/machine_uart_irq*.py 
FAIL  extmod/machine_uart_irq_txidle.py
pass  extmod_hardware/machine_uart_irq_break.py
pass  extmod_hardware/machine_uart_irq_rxidle.py
skip  extmod_hardware/machine_uart_irq_rx.py
3 tests performed (33 individual testcases)
2 tests passed
1 tests skipped: machine_uart_irq_rx
1 tests failed: machine_uart_irq_txidle
```

#### stm32 (PYBv1.0)

```
$ ./run-tests.py --target minimal extmod/machine_uart_irq*.py extmod_hardware/machine_uart_irq*.py
skip  extmod/machine_uart_irq_txidle.py
skip  extmod_hardware/machine_uart_irq_break.py
FAIL  extmod_hardware/machine_uart_irq_rxidle.py
pass  extmod_hardware/machine_uart_irq_rx.py
2 tests performed (21 individual testcases)
1 tests passed
2 tests skipped: machine_uart_irq_break machine_uart_irq_txidle
1 tests failed: machine_uart_irq_rxidle
```

ports/nrf/boards/ARDUINO_NANO_33_BLE_SENSE/mpconfigboard.h
dpgeorge: > But there is also a define for `MICROPY_PY_MACHINE_UART` in every board file.

Right... ideally that would be moved to `mpconfigport.h`, but maybe that can be left for another PR, to clean these up.

Do as you see fit.

robert-hh: I'll leave that for a separate clean-up. There is more to re-arrange.

robert-hh:
I'll see what can be done for testing. In my tests I used two modules. Using a single device with a kind of loopback was sometimes not possible..

robert-hh:
The test scripts are now reworked to some extend. Tests pass for almost all supported boards. Some boards cannot pass the test even if the feature works:
- ESP32S2 and ESP32C3 have no UART2 and cannot run the IRQ_BREAK test. IRQ_BREAK cannot be performed in a UART loopback mode. IRQ_BREAK works an can be triggered with a GPIO pulse, but then there is no serial data as expected by the test.
- STM32 (PyBoard) fails at the IRQ_RXIDLE test even if the mechanism works. It just triggers more than one IRQ, and then the state of the flag and the expected data mismatch.
- CC3200 is too slow for the IRQ_RX test, and it locks up in the callback when attempting to allocate RAM. Given that the CC3200 is legacy, the test support is not included.

dpgeorge:
> The test scripts are now reworked to some extend. Tests pass for almost all supported boards.

Excellent, thanks for working on that!  I will test them myself.

robert-hh:
Re-testing this morning when running all tests in a row with the loopback jumper attached I see a few hiccups:
- the NRF port locks up when trying to run the second test. It seems to require a hard reset between tests.
- SAMD fails the IRQ_TXIDLE test with the loopback jumper attached. In that case, the IRQ callback is called twice instead of once. There is a similar multiple call situation with STM32 and IRQ_RXIDLE.

robert-hh:
I updated the SAMD port such that the double callbacks are gone.

Gadgetoid:
Looking at RP2 (regular Pico), trying to send 512 bytes I get multiple IRQs firing but only 288 bytes of data in the first, and nothing in the subsequent IRQ(s). Is this normal, related to my issues with LTE/PPP?

```python
from machine import UART
import random
import time, sys


def irq(u):
    if u.irq().flags() & u.IRQ_RXIDLE:
        data = u.read()
        if data:
            print("IRQ_RXIDLE:", bool(u.irq().flags() & u.IRQ_RXIDLE), "data:", data, len(data))
        else:
            print("IRQ_RXIDLE: NO DATA!?")


BUF_LEN = 512
text = "".join(chr(random.randint(ord('A'), ord('z'))) for _ in range(BUF_LEN))
 
uart = UART(0, 115200, tx=0, rx=1, timeout=1)
uart.irq(irq, uart.IRQ_RXIDLE)

while True:
    print(f"\nWriting {len(text)} bytes...")
    uart.write(text)
    time.sleep_ms(5000)
```

eg:

```
Writing 512 bytes...
IRQ_RXIDLE: True data: b'b]Ubf_s`zOGscbZvdifV^FFsZyxK^gXh_yOjxHTYwOJOCiPyYfV^RYKx`ZPN[eBbDACGjumymFqvLi^X[swHFBGucRZjUkdNuiDixdpidqTDyD_W^rPngQIkkMFmJwycUdgpzgMwWW]LSMYtzgGpjCznIhP[Kpg_]Qa]TLxwmcRDDRjDJqkCVhUZ]AijGRuWwY_iUHqOimvXdbqEfgKHwvDlWpvsWFuhBwZsTV]vGycxutlZXZI\\HQe\\UmAbFeefjZZPmONI[ai]ImmDXWyCpxXedF]rZkHG' 288
IRQ_RXIDLE: NO DATA!?
IRQ_RXIDLE: NO DATA!?

Writing 512 bytes...
IRQ_RXIDLE: True data: b'b]Ubf_s`zOGscbZvdifV^FFsZyxK^gXh_yOjxHTYwOJOCiPyYfV^RYKx`ZPN[eBbDACGjumymFqvLi^X[swHFBGucRZjUkdNuiDixdpidqTDyD_W^rPngQIkkMFmJwycUdgpzgMwWW]LSMYtzgGpjCznIhP[Kpg_]Qa]TLxwmcRDDRjDJqkCVhUZ]AijGRuWwY_iUHqOimvXdbqEfgKHwvDlWpvsWFuhBwZsTV]vGycxutlZXZI\\HQe\\UmAbFeefjZZPmONI[ai]ImmDXWyCpxXedF]rZkHG' 288
IRQ_RXIDLE: NO DATA!?
```

robert-hh:
I'll check. Did you have a loopback connection at UART 0?

robert-hh:
You have to set rxbuf to a larger value. The default is 256. It should be 512 at least.

Gadgetoid:
> The default is 256. It should be 512 at least.

Hmm I suppose I should probe deeper into what my LTE module tries to send, since this implies it could be overwhelming the rx buffer and might explain why I must use polling.

That said, a generously large (4k) rxbuf does not appear to help. Which only really suggests this is unrelated.

tests/extmod_hardware/machine_uart_irq_break.py
```diff
@@ -0,0 +1,67 @@
...
+    UART.IRQ_BREAK
+except (ImportError, AttributeError):
+    print("SKIP")
+    raise SystemExit
+
+import time, sys
+
+# Configure pins based on the target.
+if "esp32" in sys.platform and not (
+    "ESP32S2" in sys.implementation._machine or "ESP32C3" in sys.implementation._machine
```
dpgeorge, line 18: I suggest to SKIP s2/s3 like this:
```py
if "esp32" in sys.platform:
    if "ESP32S2" in sys.implementation._machine or "ESP32C3" in sys.implementation._machine:
        print("SKIP")
        raise SystemExit
    ... set up pins for esp32
elif "rp2" in sys.platform:
    ... set up pins for rp2
else:
    print("Please add support for this test on this platform.")
    raise SystemExit
```
That way any new ports in the future that support IRQ_BREAK have an appropriate warning about adding support to this test.

tests/extmod_hardware/machine_uart_irq_break.py
```diff
@@ -0,0 +1,67 @@
...
+):
+    # The ESP32 needs separate UART instances for the test
```
dpgeorge, line 20: It's a shame that esp32 needs two instances... I see why, it's because sendbreak() changes the TX and RX baudrate.

I think we should change sendbreak() so it pulls the TX GPIO low for the correct period of time.  Otherwise in real applications one could lose RX chars while doing a sendbreak.  (But this is for a separate PR in the future.)

tests/extmod_hardware/machine_uart_irq_break.py
```diff
@@ -0,0 +1,67 @@
...
+    raise SystemExit
+
+
+def irq(u):
+    print("IRQ_BREAK:", bool(u.irq().flags() & u.IRQ_BREAK), "data:", u.read(1))
+
+
+# Test that the IRQ is called for each break received.
+for bits_per_s in (2400, 9600, 57600):
+    if recv_tx_pin is None:
```
dpgeorge, line 43: This block of if-code is not needed.

tests/extmod_hardware/machine_uart_irq_break.py
```diff
@@ -0,0 +1,67 @@
...
+    recv_uart.irq(irq, recv_uart.IRQ_BREAK)
+
+    print("write", bits_per_s)
+    for i in range(3):
+        send_uart.write(str(i))
+        send_uart.flush()
+        time.sleep_ms(10)
+        send_uart.sendbreak()
+        time.sleep_ms(10)
+        recv_uart.read()
```
dpgeorge, line 66: It seems this read is only needed for esp32, so I suggest doing this:
```py
if "esp32" in sys.platform:
    # On esp32 a read is needed to read in the break byte.
    recv_uart.read()
```

tests/extmod_hardware/machine_uart_irq_rx.py
```diff
@@ -0,0 +1,78 @@
...
+for bits_per_s in (2400, 9600):
+    if tx_pin is None:
+        uart = UART(uart_id, bits_per_s)
+    else:
+        uart = UART(uart_id, bits_per_s, tx=tx_pin, rx=rx_pin)
+
+    uart.irq(irq, uart.IRQ_RX)
+
+    print("write", bits_per_s)
+    if byte_by_byte:
```
dpgeorge, line 69: I suggest to just always write byte-by-byte.  That doesn't really change the test for those ports that are fast enough.

tests/extmod_hardware/machine_uart_irq_rx.py
```diff
@@ -0,0 +1,78 @@
...
+elif "nrf" in sys.platform:
+    uart_id = 0
+    tx_pin = None
+    rx_pin = None
+elif "renesas-ra" in sys.platform:
+    uart_id = 9
+    tx_pin = None  # P602 @ RA6M2
+    rx_pin = None  # P601 @ RA6M2
+else:
+    print("SKIP")
```
dpgeorge, line 48: I suggest keeping this line as `print("Please add support for this test on this platform.")`, because the test should not really be skipped on boards that do support this feature.

tests/extmod_hardware/machine_uart_irq_rxidle.py
```diff
@@ -0,0 +1,71 @@
...
+    uart_id = 0
+    tx_pin = "D1"
+    rx_pin = "D0"
+    byte_by_byte = True
+elif "samd" in sys.platform and "ItsyBitsy M4" in sys.implementation._machine:
+    uart_id = 3
+    tx_pin = "D1"
+    rx_pin = "D0"
+else:
+    print("SKIP")
```
dpgeorge, line 46: I suggest keeping `print("Please add support for this test on this platform.")`.

tests/extmod_hardware/machine_uart_irq_rxidle.py
```diff
@@ -0,0 +1,71 @@
...
+    raise SystemExit
+
+
+def irq(u):
+    if u.irq().flags() & u.IRQ_RXIDLE:
```
dpgeorge, line 51: Is stm32 the only port that needs this check?  If so then I suggest to remove the check and let the stm32 port fail.  That will serve as a reminder to fix stm32 (in a separate PR).

tests/extmod/machine_uart_irq_txidle.py
```diff
@@ -0,0 +1,69 @@
...
+elif "mimxrt" in sys.platform and "Teensy 4" in sys.implementation._machine:
+    uart_id = 1
+    tx_pin = None
+    min_window = 0
+elif "nrf" in sys.platform:
+    uart_id = 0
+    tx_pin = None
+    min_window = 0
+else:
+    print("SKIP")
```
dpgeorge, line 41: I suggest keeping `print("Please add support for this test on this platform.")`.

tests/extmod/machine_uart_irq_txidle.py
```diff
@@ -0,0 +1,69 @@
...
+    raise SystemExit
+
+
+def irq(u):
+    print("IRQ_TXIDLE:", bool(u.irq().flags() & u.IRQ_TXIDLE))
```
dpgeorge, line 46: Can this be a `u.irq().flags() == u.IRQ_TXIDLE`?  Which ports require the `&`?

tests/extmod/machine_uart_irq_txidle.py
```diff
@@ -0,0 +1,69 @@
...
+    if tx_pin is None:
+        uart = UART(uart_id, bits_per_s)
+    else:
+        uart = UART(uart_id, bits_per_s, tx=tx_pin, rx=rx_pin)
+
+    uart.irq(irq, uart.IRQ_TXIDLE)
+
+    bits_per_char = 10  # 1(startbit) + 8(bits) + 1(stopbit) + 0(parity)
+    start_time_us = (len(text) - 10) * bits_per_char * 1_000_000 // bits_per_s
+    window_ms = max(min_window, 20 * bits_per_char * 1_000 // bits_per_s + 1)
```
dpgeorge, line 62: It's probably worth commenting on the algorithm used in these two calculations.  Why `-10` for `start_time_us`, and why `20 *` for `window_ms`, etc?

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+      RP2040           yes                yes        yes
+      SAMD             yes        yes     yes
+      STM32            yes        yes
+      ============== ========== ====== ========== =========
+
+
+   .. note::
+     - The ESP32 port does not support the option hard=True.
+
+     - At the RP2040 port UART.IRQ_TXIDLE is only triggered when the message
```
dpgeorge, line 229: I suggest wording like "The rp2 port's UART.IRQ_TXIDLE is ...".

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+       is longer than 5 characters and the trigger happens when still 5 characters
+       are to be sent.
+
+     - At the RP2040 port UART.IRQ_BREAK needs receiving valid characters for triggering
```
dpgeorge, line 233: Suggest "The rp2 port's UART.IRQ_BREAK needs ...".

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+       again.
+
+     - At the SAMD port UART.IRQ_TXIDLE is triggered while the last character is sent.
```
dpgeorge, line 236: "The samd port's ..."

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+
+     - At the STM32F4xx using the trigger UART.IRQ_RXIDLE the handler will be called once
```
dpgeorge, line 238: "On STM32F4xx MCU's, using the trigger ..."

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+       after the first character and then after the end of the message, when the line is
+       idle.
+
+
+   Availability: cc3200, esp32, mimxrt, nrf, renesas-ra, rp2040, samd, stm32.
```
dpgeorge, line 243: rp2040 -> rp2

docs/library/machine.UART.rst
```diff
@@ -203,11 +178,85 @@ Methods
...
+.. data:: UART.IRQ_RXIDLE
+          UART.IRQ_RX
+          UART.IRQ_TXIDLE
+          UART.IRQ_BREAK
 
-    IRQ trigger sources
+    IRQ trigger sources.
 
-    Availability: WiPy.
+    Availability: renesas-ra, stm32, esp32, rp2040, mimxrt, samd, cc3200.
```
dpgeorge, line 262: rp2040 -> rp2

tests/extmod_hardware/machine_uart_irq_break.py
robert-hh: I kept it for a possible future board that does not allow specifying the rx/tx pins. I still wonder why the MIMXRT lib does not provide the `break` event.

tests/extmod_hardware/machine_uart_irq_break.py
robert-hh: If there is no data, the read returns `None`. So it does not hurt.

tests/extmod_hardware/machine_uart_irq_rx.py
robert-hh: But then the test would not reflect the real operation. Better then to skip testing slow devices like SAMD21.

docs/library/machine.UART.rst
robert-hh: Changed

docs/library/machine.UART.rst
robert-hh: changed

docs/library/machine.UART.rst
robert-hh: changed

docs/library/machine.UART.rst
robert-hh: changed

tests/extmod_hardware/machine_uart_irq_rx.py
robert-hh: Done. But we know it will fail for the CC3200, since it supports IRQ_RX but cannot pass the test.

tests/extmod_hardware/machine_uart_irq_rxidle.py
robert-hh: Done

tests/extmod/machine_uart_irq_txidle.py
robert-hh: Done

tests/extmod/machine_uart_irq_txidle.py
robert-hh: I just made that identical to the other test scripts, which use the `bool(a & b)` term.

tests/extmod_hardware/machine_uart_irq_rxidle.py
robert-hh: Removed.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: Added. The value of 20 could probably be smaller.  SAMD devices use busy waiting for sleep_ms(n)
for n <= 10 and for sleep_us(), preventing the IRQ to get scheduled. For these I added the minimum wait time. But I prepared a change for SAMD to make that obsolete. Will be in a separate PR.

docs/library/machine.UART.rst
robert-hh: Done

docs/library/machine.UART.rst
robert-hh: Done

tests/extmod_hardware/machine_uart_irq_break.py
robert-hh: OK

tests/extmod_hardware/machine_uart_irq_rx.py
dpgeorge: OK, for cc3200 it can have an explicit section here that skips that port.  And a comment.  Eg:
```py
elif sys.platform == "cc3200":
    # CC3200 doesn't work because it's too slow and has an allocation error in the handler.
    print("SKIP")
    raise SystemExit
```

tests/extmod/machine_uart_irq_txidle.py
dpgeorge: It's stricter to test with `==`, which I would prefer to do if possible.

The tests that require `bool(a & b)` probably need it for a (good?) reason.

tests/extmod_hardware/machine_uart_irq_break.py
dpgeorge: I would rather go the other way and make sure all ports allow specifying the tx/rx pins (not required but allowed).  That makes ports more consistent.

For a port that can't change the tx/rx pin then all it needs to do is verify that the passed-in pin matches the pre-configured pin.

Anyway, all that is for another day.  For now I don't think we need this bit of code in the test.

tests/extmod_hardware/machine_uart_irq_rx.py
robert-hh: Then the .exp file does not match.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: Changed back for IRQ_TXIDLE.

tests/extmod_hardware/machine_uart_irq_rx.py
robert-hh: Done.

tests/extmod_hardware/machine_uart_irq_break.py
robert-hh: Done

tests/extmod_hardware/machine_uart_irq_break.py
robert-hh: Removed

tests/extmod_hardware/machine_uart_irq_rx.py
dpgeorge: That's true...

OK, then let's leave it as you have it.

robert-hh:
I made all changes, but will a test run with all tests and board again.

Gadgetoid:
> * Patched by `setting hard=True` in these cases.

AFAIK the PPP module does not set `hard=True`:

https://github.com/micropython/micropython/blob/654711c699e9f27d37442f10ee01827662c4a2bf/extmod/network_ppp_lwip.c#L230-L235

Should it? Or is the PPP `poll()` method likely to do something verboten by hardware interrupts?

robert-hh:
Most likely that is not suitable. In hard irq handlers allocating memory is not possible.

Gadgetoid:
There's no reference held to the receieve buffer:

https://github.com/micropython/micropython/blob/0d313fb721cdd2516cf3c65333f5d89c6ce632d2/ports/mimxrt/machine_uart.c#L339-L340

So if you add a `gc.collect()` to the txidle test, it will crash.

Minimal repro for RP2:

```python
# Test machine.UART.IRQ_TXIDLE firing after a transmission.
# Does not require any external connections.

from machine import UART
import time, sys
import gc

# Configure pins based on the target.
hard_irq = False

uart_id = 0
tx_pin = "GPIO0"
rx_pin = "GPIO1"
min_window = 1
hard_irq = True


def irq(u):
    print("IRQ_TXIDLE:", u.irq().flags() == u.IRQ_TXIDLE)


text = "Hello World" * 30

# Test that the IRQ is called after the write has completed.
for bits_per_s in (2400, 9600, 115200):
    uart = UART(uart_id, bits_per_s, tx=tx_pin, rx=rx_pin, rxbuf=512)
    uart.irq(irq, uart.IRQ_TXIDLE, hard=True)
    
    # Comment me in/out for crashy funtimes!
    gc.collect()

    bits_per_char = 10  # 1(startbit) + 8(bits) + 1(stopbit) + 0(parity)
    start_time_us = (len(text) - 10) * bits_per_char * 1_000_000 // bits_per_s
    window_ms = max(min_window, 20 * bits_per_char * 1_000 // bits_per_s + 1)

    print("write", bits_per_s)
    uart.write(text)
    time.sleep_us(start_time_us)
    print("ready")
    time.sleep_ms(window_ms)
    print("done")

```

robert-hh:
About the NRF port: testing with a Seeed XIAO NRF52840 instead of the Arduino Nano 33 BLE, test batches started by run-tests.py work,  but the test for IRQ_TXIDLE fails if the rx/tx loopback is present.

robert-hh:
> So if you add a gc.collect() to the txidle test, it will crash.

The crash is not related to `hard=True`. Let's look for a free floating pointer.

Gadgetoid:
> > So if you add a gc.collect() to the txidle test, it will crash.
> 
> The crash is not related to `hard=True`. Let's look for a free floating pointer.

You mean this one?  😆 

https://github.com/micropython/micropython/blob/0d313fb721cdd2516cf3c65333f5d89c6ce632d2/ports/mimxrt/machine_uart.c#L339-L340

Sorry crossed wires, this is *another* thing I ran into while monkeying with PPP.

robert-hh:
I found the culprit. It is the mp_irq_obj pointer. Since the RP2 has static UART objects, these were not linked to root pointers and a gc.collect() would free them.

Gadgetoid:
Awesome! And here I am still staring glassy eyed at ring buffer code chasing one of those read herons.... who would read a heron anyway?!

robert-hh:
@Gadgetoid I updated the RP2 file yesterday evening to address the crash with gc.collect(). SO you can try the most recent version of the PR.
@dpgeorge There was an issue with non-rooted mp_irq_obj objects and the fact, that with RP2 and NRF the UART objects are static.  That is fixed now. It did not reduce the behavior differences of the two NRF boards I have.

robert-hh:
@dpgeorge There is a new commit now which fixes a bug in the NRF UART, which did not allow to re-instantiate the UART with a different baud rate. Changing the baud rate required a reset. I can squash this commit with the other NRF commit, or even make a separate PR, if you prefer.

robert-hh:
Tested with a RPI PICO2 in both ARM and RISCV mode. The test cases pass.

ports/rp2/machine_uart.c
```diff
@@ -422,6 +448,9 @@ static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_arg
 
     // Get static peripheral object.
     machine_uart_obj_t *self = (machine_uart_obj_t *)&machine_uart_obj[uart_id];
+    self->mp_irq_obj = NULL;
+    MP_STATE_PORT(rp2_uart_irq_obj[self->uart_id]) = self->mp_irq_obj;
```
dpgeorge, line 452: Does this irq obj need to be cleared?  Usually something like `machine.UART(1)` is used to retrieve an existing object and not reset it.  But here, the IRQ will be reset (actually the hardware IRQ stays active, if it's enabled).

ports/nrf/modules/machine/uart.c
```diff
@@ -293,6 +335,58 @@ static bool mp_machine_uart_txdone(machine_uart_obj_t *self) {
...
+const mp_irq_methods_t uart_irq_methods = {
+    .trigger = uart_irq_trigger,
+    .info = uart_irq_info,
+};
+
+static mp_irq_obj_t *mp_machine_uart_irq(machine_uart_obj_t *self, bool any_args, mp_arg_val_t *args) {
+    if (self->mp_irq_obj == NULL) {
+        self->mp_irq_trigger = 0;
+        self->mp_irq_obj = mp_irq_new(&uart_irq_methods, MP_OBJ_FROM_PTR(self));
+        MP_STATE_PORT(nrf_uart_irq_obj) = self->mp_irq_obj;
```
dpgeorge, line 373: In case one day more UARTs are supported by this port, this should be generalised to an array, and do something like:
```c
MP_STATE_PORT(nrf_uart_irq_obj)[self->uart_id] = self->mp_irq_obj;
```

tests/extmod/machine_uart_irq_txidle.py
```diff
@@ -0,0 +1,82 @@
...
+for bits_per_s in (2400, 9600, 115200):
+    if tx_pin is None:
+        uart = UART(uart_id, bits_per_s)
+    else:
+        uart = UART(uart_id, bits_per_s, tx=tx_pin, rx=rx_pin)
+
+    if hard_irq is True:
+        uart.irq(irq, uart.IRQ_TXIDLE, hard=True)
+    else:
+        uart.irq(irq, uart.IRQ_TXIDLE)
```
dpgeorge, line 64: It should be possible to do `uart.irq(irq, uart.IRQ_TXIDLE, hard=hard_irq)`, unconditionally.

tests/extmod/machine_uart_irq_txidle.py
```diff
@@ -0,0 +1,82 @@
...
+
+    UART.IRQ_TXIDLE
+except (ImportError, AttributeError):
+    print("SKIP")
+    raise SystemExit
+
+import time, sys
+
+# Configure pins based on the target.
+hard_irq = False
```
dpgeorge, line 15: Maybe add a comment: "use hard IRQ on ports that need to respond immediately to catch the TXIDLE flag"

ports/rp2/machine_uart.c
```diff
@@ -460,6 +489,77 @@ static void mp_machine_uart_sendbreak(machine_uart_obj_t *self) {
...
+const mp_irq_methods_t uart_irq_methods = {
+    .trigger = uart_irq_trigger,
+    .info = uart_irq_info,
+};
+
+static mp_irq_obj_t *mp_machine_uart_irq(machine_uart_obj_t *self, bool any_args, mp_arg_val_t *args) {
+    if (self->mp_irq_obj == NULL) {
+        self->mp_irq_trigger = 0;
+        self->mp_irq_obj = mp_irq_new(&uart_irq_methods, MP_OBJ_FROM_PTR(self));
+        MP_STATE_PORT(rp2_uart_irq_obj[self->uart_id]) = self->mp_irq_obj;
```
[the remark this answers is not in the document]
robert-hh: Done. But there are more and older places where the above bracket placement is used. Should these be changed as well? Looking into the generated code, I did not see a difference.

ports/rp2/machine_uart.c
robert-hh: The behavior of `machine.UART(1)` is not consistent over the  ports. Ig a port has static UART objects like RP2 or NRF, it will refer to the existing state. If a port has dynamic objects like STM32, ESP32, MIMXRT, Renesas, SAMD,  will instantiate a new object. It seems to be more consistent to see it as a new object and clear previous states.

ports/nrf/modules/machine/uart.c
robert-hh: There is no uart_id element in the `machine_uart_obj_t` data type. I can add it, or use a constant index of 0.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: The CC200 port for instance does not support the `hard=xxx` argument. But CC3200 does not provide the IRQ_TXIDLE event, so the test will be skipped.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: OK.

ports/nrf/modules/machine/uart.c
robert-hh: Added uart_id.

ports/rp2/machine_uart.c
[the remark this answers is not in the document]
dpgeorge: > But there are more and older places where the above bracket placement is used. Should these be changed as well?

No, leave them, we can clean them up separately (separate PR).  That's low priority though.



> Looking into the generated code, I did not see a difference.

The generated code is indeed equivalent.  But it's more conceptual.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: Nevertheless changed.

ports/rp2/machine_uart.c
dpgeorge: > If a port has static UART objects like RP2 or NRF, it will refer to the existing state.

Yes, exactly, on rp2 it will refer to the existing instance.  And so in this case (eg accessing `machine.UART(1)`) the IRQ settings should not be changed, for consistency.

> If a port has dynamic objects like STM32, ESP32, MIMXRT, Renesas, SAMD, will instantiate a new object.

Actually, on stm32 any previous UART instance is also retrieved unchanged with a `machine.UART(1)`.

> Yes, it needs to be cleared to have a clean state after soft reset

Then that probably needs to be done in a `machine_uart_deinit()` function.

tests/extmod/machine_uart_irq_txidle.py
dpgeorge: Ideally we could remove these kinds of small inconsistencies between ports and make the IRQ behave as similar as possible across the ports.

Just to be clear, the hard IRQ is needed only on rp2 when TX/RX have a connection between them (which isn't required for this test but is for the other ones)?

And for samd?

Maybe would be good to make the comment here more detailed, with the above details, so someone who wants to eventually fix things can know exactly what to look for, what tests to run to check if hard IRQ is really needed.

ports/rp2/machine_uart.c
robert-hh: It can indeed be done there. Thanks for the hint. But uart.deinit() is not automatically called on soft reset.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: > Just to be clear, the hard IRQ is needed only on rp2 when TX/RX have a connection between them (which isn't required for this test but is for the other ones)?

Yes. The NRF port has similar issues, but does not support hard irq. The SAMD port is fine.

ports/rp2/machine_uart.c
robert-hh: I could call `machine_uart_deinit()` in `mp_machine_uart_make_new` to settle the ground.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: I assume that the IRQ_TXIDLE is triggered while still data is sent. For RP2 this is doumented as 4-5 bytes, but for SAMD21 this is true for one byte as well. Then, when the IRQ callback is called, the IRQ_TXIDLE flag of self->irq_flags is overwritten by the byte TX interrupt.

dpgeorge: > we could change the NRF port to accept & ignore the hard irq argument instead raising an error.

All ports should (eventually) be using the shared code in `shared/runtime/mpirq.c`, and then they would all support the `hard` argument (I guess).  But getting all this consistent now may be too big of a job and we'd never finish :)

> Then, when the IRQ callback is called, the IRQ_TXIDLE flag of self->irq_flags is overwritten by the byte TX interrupt.

So maybe there's something to fix here: if the user doesn't want to respond to TX IRQ's (there's no such thing as far as the user is concerned) then `self->irq_flags` should not be updated when this IRQ occurs.

ports/rp2/machine_uart.c
dpgeorge: I would think the UART can behave the same way as Pin, in the way IRQs are handled, and the way they are deinitialised on soft reset.

tests/extmod/machine_uart_irq_txidle.py
robert-hh: > All ports should (eventually) be using the shared code in shared/runtime/mpirq.c, and then they would all support the hard argument (I guess). 

The NRF port uses the code from  shared/runtime/mpirq.c but raises an error if the `hard` argument is used.

> if the user doesn't want to respond to TX IRQ's (there's no such thing as far as the user is concerned) then self->irq_flags should not be updated when this IRQ occurs.

With the RX/TX loopback it's the RX event which is triggered. And the uart.irq() callback and the uart IRQ handler are not synchronized. Ideally the respective flag for an event would only be cleared if the callback has finished. But that is not possible for a scheduled callback.

ports/rp2/machine_uart.c
robert-hh: So I added a `machine_uart_deinit_all()`  function and call it in main() upon soft reset,
and cleared the uart.irq() state in uart.deinit().

tests/extmod/machine_uart_irq_txidle.py
robert-hh: I reworked the handling of the the flags reported by irq.flags(). It will only be changed when a event is present which is expected by the trigger. With that change, hard=True is not needed for the TXIDLE test, and the RX/TX loopback does not cause any problem.

ports/esp32/machine_timer.c
```diff
@@ -130,7 +112,33 @@ static mp_obj_t machine_timer_make_new(const mp_obj_type_t *type, size_t n_args,
     return self;
 }
 
-static void machine_timer_disable(machine_timer_obj_t *self) {
+machine_timer_obj_t *machine_timer_create(mp_uint_t timer) {
```
dpgeorge, line 115: I think this new create function should be used by the above `machine_timer_make_new`, because otherwise there's a lot of duplicated code.

ports/rp2/machine_uart.c
dpgeorge: Great, this is much better, thanks!

tests/extmod/machine_uart_irq_txidle.py
dpgeorge: Excellent!

ports/mimxrt/hal/fsl_lpuart.c
```diff
@@ -0,0 +1,2013 @@
+/*
```
dpgeorge, line 1: Which MCU source file is this taken from?  Doesn't it need to be different for each MCU?

Is there a better way to fix the code here rather than copying the file?

dpgeorge:
@robert-hh this is looking really good now, thank you for all the effort on it, especially with the testing.

Just two remaining comments, see above.

ports/mimxrt/hal/fsl_lpuart.c
robert-hh: As far as I recall, that was from the 1062 lib file. But this file is identical for all MIMXRT1xxx MCUs. Newer Versions of the lib use a common UART driver file for all MIMXRT MCU, and this hal/fsl_lpuart.c file is besisdes the changes identical to the one I used a baseline.
I tested it with different MCUs, but can re-test it again.
The change was needed to make IRQ_RXIDLE work as intended by the MCU developers.

dpgeorge: If you tested it on different MCUs then I'm happy with that.

ports/esp32/machine_timer.c
robert-hh: Done.

ports/mimxrt/hal/fsl_lpuart.c
robert-hh: I repeated the test with MIMXRT1010, MIMXRT1015, MIMXRT1021, MIMXRT1052, MIMXRT1062 and MIMXRT1172. All work fine. The test script is set up for Teensy 4.x, but that restriction can be dropped, since the MIMXRT port does not need to specify UART pins.

ports/esp32/machine_timer.c
robert-hh: For ESP32, SAMD, Renesas-RA and NRf there are still two commits. I can join them, leaving just one commit per port.

robert-hh:
The change to esp32/machine_timer.c is made and the MIMXRT MCU's re-tested. I removed at the test scripts the restrictions for Teensy 4.x. So any MIMXRT board can be used.

ports/esp32/machine_timer.c
dpgeorge: Thanks for cleaning up the esp32 timer functions.  The diff is smaller now, which is great!

ports/mimxrt/hal/fsl_lpuart.c
dpgeorge: Very good.

dpgeorge:
> I removed at the test scripts the restrictions for Teensy 4.x. So any MIMXRT board can be used.

OK, great.  I also removed the test configuration for the samd `SEEED_XIAO_SAMD21` because that doesn't have UART IRQs enabled by default (I was enabling them manually to test on this boad).  I have now ordered an `ADAFRUIT_ITSYBITSY_M0_EXPRESS` to use for testing instead.

I also did some very minor fixes to this PR:
- remove some stray double blank lines
- add a few blank lines around `#if`/`#endif`'s
- make `const mp_irq_methods_t uart_irq_methods` static

And I kept all 16 commits.  They are well separated commits making independent changes.

Thanks @robert-hh for all your work on this!  And @Gadgetoid for testing, and @projectgus for additional reviews.

robert-hh:
Thank you for the improvements and for merging.

iabdalkader:
Note this doesn't build anymore with more recent SDKs because the isr variable is now defined:
```
extern lpuart_isr_t s_lpuartIsr[];
```

robert-hh:
> Note this doesn't build anymore with more recent SDKs

Which port?

iabdalkader:
> Which port?

mimxrt, with any recent SDK release.

robert-hh:
We are still at sdk 2.10, and we use a modified fsl_lpuart.c to get a proper IRQ_RXIDLE behavior. Updating to a newer SDK is still due. When that happens I expect more changes to be done. No too many, as I noticed when trying that last year.

iabdalkader:
> We are still at sdk 2.10, and we use a modified fsl_lpuart.c to get a proper IRQ_RXIDLE behavior.

I understand. I'm just saying if you try to use a more recent SDK this doesn't build anymore. However, it seems everything else works. I'm not sure which SDK I'm using but it's more recent than the one in MicroPython.

ports/mimxrt/hal/fsl_lpuart.c
andrewleech: Hi @robert-hh I'm looking at this file from the perspective of pulling in a new version of it for SDK update. With the auto-formatting done to micropython code it's hard to be sure of the changes you've made.
Was it just the changes in `LPUART_TransferHandleIRQ()` where the large block in the middle `/* If IDLE flag is set and the IDLE interrupt is enabled. */` was removed and replaced with the smaller version at the end?

Before:
``` c
    /* If IDLE flag is set and the IDLE interrupt is enabled. */
    if ((0U != ((uint32_t)kLPUART_IdleLineFlag & status)) &&
        (0U != ((uint32_t)kLPUART_IdleLineInterruptEnable & enabledInterrupts)))
    {
#if defined(FSL_FEATURE_LPUART_HAS_FIFO) && FSL_FEATURE_LPUART_HAS_FIFO
        count = ((uint8_t)((base->WATER & LPUART_WATER_RXCOUNT_MASK) >> LPUART_WATER_RXCOUNT_SHIFT));

        while ((0U != handle->rxDataSize) && (0U != count))
        {
            tempCount = (uint8_t)MIN(handle->rxDataSize, count);

            /* Using non block API to read the data from the registers. */
            LPUART_ReadNonBlocking(base, handle->rxData, tempCount);
            handle->rxData = &handle->rxData[tempCount];
            handle->rxDataSize -= tempCount;
            count -= tempCount;

            /* If rxDataSize is 0, invoke rx idle callback.*/
            if (0U == (handle->rxDataSize))
            {
                handle->rxState = (uint8_t)kLPUART_RxIdle;

                if (NULL != handle->callback)
                {
                    handle->callback(base, handle, kStatus_LPUART_RxIdle, handle->userData);
                }
            }
        }
#endif
        /* Clear IDLE flag.*/
        base->STAT = ((base->STAT & 0x3FE00000U) | LPUART_STAT_IDLE_MASK);

        /* If rxDataSize is 0, disable rx ready, overrun and idle line interrupt.*/
        if (0U == handle->rxDataSize)
        {
            /* Disable and re-enable the global interrupt to protect the interrupt enable register during
             * read-modify-wrte. */
            irqMask = DisableGlobalIRQ();
            base->CTRL &= ~(uint32_t)(LPUART_CTRL_RIE_MASK | LPUART_CTRL_ILIE_MASK | LPUART_CTRL_ORIE_MASK);
            EnableGlobalIRQ(irqMask);
        }
        /* Invoke callback if callback is not NULL and rxDataSize is not 0. */
        else if (NULL != handle->callback)
        {
            handle->callback(base, handle, kStatus_LPUART_IdleLineDetected, handle->userData);
        }
        else
        {
            /* Avoid MISRA 15.7 */
        }
    }
```
after 
``` c
    /* If IDLE flag is set and the IDLE interrupt is enabled. */
    if ((0U != ((uint32_t)kLPUART_IdleLineFlag & status)) &&
        (0U != ((uint32_t)kLPUART_IdleLineInterruptEnable & enabledInterrupts))) {
        /* Clear IDLE flag.*/
        base->STAT |= LPUART_STAT_IDLE_MASK;
        if (NULL != handle->callback) {
            handle->callback(base, handle, kStatus_LPUART_IdleLineDetected, handle->userData);
        } else {
            /* Avoid MISRA 15.7 */
        }
    }
```

robert-hh: Comparinf is easier if you run the initial fsl_lpuart.c through uncrustify as well before comparing. So yes, the changes focus on handling the line_idle signal. I removed most of the initial code dealing with it and just kept the part which handles the callback mechanism. So it looks like the code you show is identical to the blocks below.
So this block is removed, which is an `if() .....`:
```
    /* If IDLE flag is set and the IDLE interrupt is enabled. */
    if ((0U != ((uint32_t)kLPUART_IdleLineFlag & status)) &&
        (0U != ((uint32_t)kLPUART_IdleLineInterruptEnable & enabledInterrupts))) {
        #if defined(FSL_FEATURE_LPUART_HAS_FIFO) && FSL_FEATURE_LPUART_HAS_FIFO
        count = ((uint8_t)((base->WATER & LPUART_WATER_RXCOUNT_MASK) >> LPUART_WATER_RXCOUNT_SHIFT));

        while ((0U != handle->rxDataSize) && (0U != count)) {
            tempCount = (uint8_t)MIN(handle->rxDataSize, count);

            /* Using non block API to read the data from the registers. */
            LPUART_ReadNonBlocking(base, handle->rxData, tempCount);
            handle->rxData = &handle->rxData[tempCount];
            handle->rxDataSize -= tempCount;
            count -= tempCount;

            /* If rxDataSize is 0, invoke rx idle callback.*/
            if (0U == (handle->rxDataSize)) {
                handle->rxState = (uint8_t)kLPUART_RxIdle;

                if (NULL != handle->callback) {
                    handle->callback(base, handle, kStatus_LPUART_RxIdle, handle->userData);
                }
            }
        }
        #endif
        /* Clear IDLE flag.*/
        base->STAT = ((base->STAT & 0x3FE00000U) | LPUART_STAT_IDLE_MASK);

        /* If rxDataSize is 0, disable rx ready, overrun and idle line interrupt.*/
        if (0U == handle->rxDataSize) {
            /* Disable and re-enable the global interrupt to protect the interrupt enable register during
             * read-modify-wrte. */
            irqMask = DisableGlobalIRQ();
            base->CTRL &= ~(uint32_t)(LPUART_CTRL_RIE_MASK | LPUART_CTRL_ILIE_MASK | LPUART_CTRL_ORIE_MASK);
            EnableGlobalIRQ(irqMask);
        }
        /* Invoke callback if callback is not NULL and rxDataSize is not 0. */
        else if (NULL != handle->callback) {
            handle->callback(base, handle, kStatus_LPUART_IdleLineDetected, handle->userData);
        } else {
            /* Avoid MISRA 15.7 */
        }
    }
```
And this on is added, which was part of the removed block:
```
    /* If IDLE flag is set and the IDLE interrupt is enabled. */
    if ((0U != ((uint32_t)kLPUART_IdleLineFlag & status)) &&
        (0U != ((uint32_t)kLPUART_IdleLineInterruptEnable & enabledInterrupts))) {
        /* Clear IDLE flag.*/
        base->STAT |= LPUART_STAT_IDLE_MASK;
        if (NULL != handle->callback) {
            handle->callback(base, handle, kStatus_LPUART_IdleLineDetected, handle->userData);
        } else {
            /* Avoid MISRA 15.7 */
        }
    }
```

PR #14042, opened 2024-03-06
stm32/mboot: Add support for a raw filesystem.

dpgeorge:
Mboot currently supports loading firmware from a filesystem, like FAT or littlefs.  But in some cases it's more robust to simply use raw block storage for the firmware, rather than a filesystem.

This PR adds support for such a thing: a new `MBOOT_VFS_RAW` option which is a simple, contiguous area of storage that contains the firmware file to update with.  There is no metadata on this raw filesystem, just the data of the file.

To use this feature the application just needs to make sure the firmware to update with is available somewhere contiguous in flash storage (internal or external flash) for mboot to read, and then tells mboot the starting location and size.  Then mboot will be able to do a firmware update as usual.

This feature is enabled by default if `MBOOT_FSLOAD` is enabled, although a board can explicitly disable it by `#define MBOOT_VFS_RAW (0)`.

The cost for enabling this (if `MBOOT_FSLOAD` is already enabled) is +96 bytes of mboot firmware.


dpgeorge:
LEGO_HUB_NO6 has been updated to use this new feature.  Tests have been done on this board with this PR, using the `appupdate` script.

dpgeorge:
Updated to allow the raw filesystem to wrap around, to support wear levelling of the flash that the firmware update is stored to.

PR #14045, opened 2024-03-06
py: Add support for function doc strings.

dpgeorge:
This PR adds support for function doc strings.  It's disabled by default.

```py
def f():
    "doc string"
    pass

print(f.__doc__)
```

TODO:
- [x] add tests
- [x] ~~maybe have a separate compile-time option for this~~ no, let's keep it under the existing MICROPY_ENABLE_DOC_STRING option
- [ ] check that the native .mpy ABI is unchanged

### DIFF
--- a/py/compile.c
+++ b/py/compile.c
@@ -797,6 +797,32 @@ static qstr compile_funcdef_helper(compiler_t *comp, mp_parse_node_struct_t *pns
     // compile the function definition
     compile_funcdef_lambdef(comp, fscope, pns->nodes[1], PN_typedargslist);
 
+    #if MICROPY_ENABLE_DOC_STRING
+    // look for the first statement
+    mp_parse_node_t pn = pns->nodes[3]; // the function body/suite
+    if (MP_PARSE_NODE_IS_STRUCT_KIND(pn, PN_suite_block_stmts)) {
+        // a list of statements; get the first one
+        pn = ((mp_parse_node_struct_t *)pn)->nodes[0];
+    }
+
+    // check the first statement for a doc string
+    if (MP_PARSE_NODE_IS_STRUCT_KIND(pn, PN_expr_stmt)) {
+        mp_parse_node_struct_t *pns2 = (mp_parse_node_struct_t *)pn;
+        if ((MP_PARSE_NODE_IS_LEAF(pns2->nodes[0])
+             && MP_PARSE_NODE_LEAF_KIND(pns2->nodes[0]) == MP_PARSE_NODE_STRING)
+            || (MP_PARSE_NODE_IS_STRUCT_KIND(pns2->nodes[0], PN_const_object)
+                && mp_obj_is_str(mp_parse_node_extract_const_object((mp_parse_node_struct_t *)pns2->nodes[0])))) {
+            // duplicate the function on the stack
+            EMIT(dup_top);
+            // compile the doc string
+            compile_node(comp, pns2->nodes[0]);
+            // store the doc string into the function
+            EMIT(rot_two);
+            EMIT_ARG(attr, MP_QSTR___doc__, MP_EMIT_ATTR_STORE);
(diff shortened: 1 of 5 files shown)


py/compile.c
```diff
@@ -782,6 +782,32 @@ STATIC qstr compile_funcdef_helper(compiler_t *comp, mp_parse_node_struct_t *pns
     // compile the function definition
     compile_funcdef_lambdef(comp, fscope, pns->nodes[1], PN_typedargslist);
 
+    #if MICROPY_ENABLE_DOC_STRING
+    // look for the first statement
+    mp_parse_node_t pn = pns->nodes[3]; // the function body/suite
+    if (MP_PARSE_NODE_IS_STRUCT_KIND(pn, PN_suite_block_stmts)) {
```
jimmo, line 788: This check already seems to be done by `check_for_doc_string`?

i.e. this whole thing could be a call to `check_for_doc_string`, the only difference is how it emits the store of `__doc__` ?

py/compile.c
```diff
@@ -782,6 +782,32 @@ STATIC qstr compile_funcdef_helper(compiler_t *comp, mp_parse_node_struct_t *pns
     // compile the function definition
     compile_funcdef_lambdef(comp, fscope, pns->nodes[1], PN_typedargslist);
 
+    #if MICROPY_ENABLE_DOC_STRING
```
jimmo, line 785: Can we update the comment in `mpconfig.h` to explain that this enables docstrings on functions, modules, and classes?

(I think we should keep them as one combined option)

dpgeorge:
I really want to find a way to make this use less RAM for functions that don't have doc strings.  Because at the moment all functions increase by 1 GC block size, regardless of whether they use the new doc string slot.

But I don't know how to do that in a simple way.  It probably requires a separate function type like `mp_type_fun_bc_with_doc_string` and the ability to upgrade a normal function to this type when the doc string entry is stored to (and eg repurpose the `const byte *bytecode` entry in `mp_obj_fun_bc_t` to point to a separate struct that points to the doc string and the bytecode).

If we can find a way to do that then we could enable this feature on the large ports, because it would only cost something if docs strings are used.

Also could extend things to allow any attribute to be stored to a function object, not just the `__doc__` entry.  Doc strings would then just be one user of this feature (to store attrs to functions).

Josverl:
Out of interest wrt to the Micropython-stubs project I'd like to understand what the goal is for this feature

Is it targeting stdlib/micropython code, frozen (. Mpy) code, or regular micropython.py modules?

I assume the interface would be the cpython style `__doc__` property, is that correct?

Is there something similar considered for modules or classes.?

py/compile.c
dpgeorge: Good idea.  I've now done this code refactor.

py/compile.c
dpgeorge: Comment updated.

dpgeorge:
> Out of interest wrt to the Micropython-stubs project I'd like to understand what the goal is for this feature
> 
> Is it targeting stdlib/micropython code, frozen (. Mpy) code, or regular micropython.py modules?

It's mainly by request of users.  Eg pyscript is very interested to have this enabled in its build, so pyscript users have access to docstrings.

It will work on both frozen code and regular Python code.  But the docstrings need to be there already, otherwise `X.__doc__` just returns `None` (or raises an attribute exception).


> I assume the interface would be the cpython style `__doc__` property, is that correct?

Yes.  I've added a test to show how it's compatible with CPython, with the `__doc__` attribute.

> Is there something similar considered for modules or classes.?

`__doc__` already works for modules and classes, it's just never been enabled.  This PR adds support for `__doc__` on functions and methods.

Josverl:
Thanks for clarifying.
That means that I should look into inspecting the availability of the `__doc__` property in the firmware stub creation script.

PR #13777, opened 2024-02-29
global: Prune trailing whitespace.

Gadgetoid:
Prune trailing whitespace across the whole project, this is done automatically with:

```
grep -IUrl --color "[[:blank:]]$" --exclude-dir=.git --exclude=*.exp | xargs sed -i 's/[[:space:]]*$//'
```

Raised as a draft because this touches 113 files and may break things so I do not expect it to be merged as is, but I think we should be cognisant of these... issues. I've avoided .exp files, but much of the remaining diffs are still near inscrutable.

Though GitHub will filter out whitespace from diffs if you ask nicely- https://github.com/micropython/micropython/pull/13777/files?diff=unified&w=1

If there's any interest in this kind of tidyup and this PR isn’t suitable for merging, I would recommend individual authors using some variation of the above command or editor config to clean up files as they are otherwise modified.

PR #14046, opened 2024-03-06
global: Run uncrustify on all .c source files. 

Gadgetoid:
Enforce .c code consistency by running uncrustify -c tools/uncrustify.cfg across the whole* codebase, with:

```
find . -name *.c | xargs uncrustify -c tools/uncrustify.cfg --replace --no-backup
```

Followed by tools/codeformat.py to fix all the erroneous ifdef indents.

* Skip third-party code in lib/ and drivers/cc3100/
* Skip third-party code in ports/cc3200/FreeRTOS/
* Skip all exclusions listed in tools/codeformat.py

As mentioned in https://github.com/micropython/micropython/pull/13777 it looks like there are files that are not covered either by the `PATHS` or `EXCLUSIONS` in `tools/codeformat.py`, resulting in some files that have not been considered for code consistency reformatting.

I think there are some valid changes here, but also a few left that raise an eyebrow.

Should some of these paths be added to the code formatting tool? Eg:

* `drivers/bus/*.[ch]`
* `drivers/memory/*.[ch]`
* `examples/**/*.[ch]`

Also it looks like `extmod/nimble/nimble/nimble_npl_os.c` is missed, since the glob is `extmod/nimble/*.[ch]` and, as such, does not include that subdirectory.

Other files should either be formatted, or explicitly excluded.

The nrf port has been ignored because it's not fully formatted, but if it's ignored will it ever be fully formatted?

### DIFF
--- a/extmod/nimble/nimble/nimble_npl_os.c
+++ b/extmod/nimble/nimble/nimble_npl_os.c
@@ -79,8 +79,8 @@ static void *m_malloc_bluetooth(size_t size) {
     return alloc->data;
 }
 
-static mp_bluetooth_nimble_malloc_t* get_nimble_malloc(void *ptr) {
-    return (mp_bluetooth_nimble_malloc_t*)((uintptr_t)ptr - sizeof(mp_bluetooth_nimble_malloc_t));
+static mp_bluetooth_nimble_malloc_t *get_nimble_malloc(void *ptr) {
+    return (mp_bluetooth_nimble_malloc_t *)((uintptr_t)ptr - sizeof(mp_bluetooth_nimble_malloc_t));
 }
 
 static void m_free_bluetooth(void *ptr) {
@@ -117,7 +117,7 @@ static bool is_valid_nimble_malloc(void *ptr) {
 
 void *nimble_malloc(size_t size) {
     DEBUG_MALLOC_printf("NIMBLE malloc(%u)\n", (uint)size);
-    void* ptr = m_malloc_bluetooth(size);
+    void *ptr = m_malloc_bluetooth(size);
     DEBUG_MALLOC_printf("  --> %p\n", ptr);
     return ptr;
 }
--- a/extmod/nimble/nimble/nimble_npl_os.c
+++ b/extmod/nimble/nimble/nimble_npl_os.c
@@ -93,11 +93,11 @@ static void m_free_bluetooth(void *ptr) {
     } else {
         MP_STATE_PORT(bluetooth_nimble_memory) = NULL;
     }
-    m_free(alloc
     #if MICROPY_MALLOC_USES_ALLOCATED_SIZE
-           , alloc->size
+    m_free(alloc, alloc->size);
+    #else
+    m_free(alloc);
     #endif
-    );
 }
 
 // Check if a nimble ptr is tracked.
--- a/tools/codeformat.py
+++ b/tools/codeformat.py
@@ -34,11 +34,13 @@ import subprocess
 
 # Relative to top-level repo dir.
 PATHS = [
-    "drivers/ninaw10/*.[ch]",
+    "drivers/**/*.[ch]",
+    "examples/**/*.[ch]",
     "extmod/*.[ch]",
     "extmod/btstack/*.[ch]",
-    "extmod/nimble/*.[ch]",
+    "extmod/nimble/**/*.[ch]",
     "lib/mbedtls_errors/tester.c",
+    "shared/libc/*.[ch]",
     "shared/netutils/*.[ch]",
     "shared/timeutils/*.[ch]",
     "shared/runtime/*.[ch]",
@@ -49,7 +51,10 @@ PATHS = [
 ]
 
 EXCLUSIONS = [
+    # Fixups broken by preprocessor macro
+    "shared/readline/*.[ch]",
     # The cc3200 port is not fully formatted yet.
+    "drivers/cc3100/*/*.[ch]",
     "ports/cc3200/*/*.[ch]",
     # ESP-IDF downloads 3rd party code.
     "ports/esp32/managed_components/*",
--- a/drivers/bus/softqspi.c
+++ b/drivers/bus/softqspi.c
@@ -57,7 +57,7 @@ static void nibble_write(mp_soft_qspi_obj_t *self, uint8_t v) {
 }
 
 static int mp_soft_qspi_ioctl(void *self_in, uint32_t cmd, uintptr_t arg) {
-    mp_soft_qspi_obj_t *self = (mp_soft_qspi_obj_t*)self_in;
+    mp_soft_qspi_obj_t *self = (mp_soft_qspi_obj_t *)self_in;
     (void)arg;
 
     switch (cmd) {
@@ -68,7 +68,7 @@ static int mp_soft_qspi_ioctl(void *self_in, uint32_t cmd, uintptr_t arg) {
             // Configure pins
             mp_hal_pin_write(self->clk, 0);
             mp_hal_pin_output(self->clk);
-            //mp_hal_pin_write(self->clk, 1);
+            // mp_hal_pin_write(self->clk, 1);
             mp_hal_pin_output(self->io0);
             mp_hal_pin_input(self->io1);
             mp_hal_pin_write(self->io2, 1);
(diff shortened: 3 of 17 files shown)


jimmo:
Thanks @Gadgetoid

> Should some of these paths be added to the code formatting tool?

Definitely.

The initial list of inclusions was partially just to keep the scope of the transition manageable.

We should add `"drivers/**/*.[ch]"` to the paths, and add an exclusion for `drivers/cc3100`.

Also add `"examples/**/*.[ch]"`, `"shared/libc/*.[ch]"`, and `"shared/readline/*.[ch]"`.

And then finally the nimble glob should be updated to include a `**`.

Then I think this PR should just be the results of running that, no other changes (i.e. revert the changes to `drivers/cc3100` and `shared/memzip` and `ports/cc3200`).

nRF we can sort out later (but replacing it with Zephyr is still the plan there...)

projectgus:

Gadgetoid:
@jimmo split into two commits, one to update the rules and one to apply them.

And, of course, rebased to avoid conflicts with the `STATIC` and whitespace fixes.

extmod/nimble/nimble/nimble_npl_os.c
```diff
@@ -94,10 +94,10 @@ static void m_free_bluetooth(void *ptr) {
         MP_STATE_PORT(bluetooth_nimble_memory) = NULL;
     }
     m_free(alloc
-    #if MICROPY_MALLOC_USES_ALLOCATED_SIZE
-           , alloc->size
-    #endif
-    );
+        #if MICROPY_MALLOC_USES_ALLOCATED_SIZE
+        , alloc->size
+        #endif
```
Gadgetoid, line 99: This is a bit of an odd case- looks like uncrustify bumps up the indent but the fixups don't catch it. Should they?

shared/readline/readline.c
```diff
@@ -293,16 +293,16 @@ int readline_process_char(int c) {
                 break;
             #if MICROPY_REPL_EMACS_WORDS_MOVE
             case 'b':
-#if MICROPY_REPL_EMACS_EXTRA_WORDS_MOVE
-backward_word:
-#endif
+                #if MICROPY_REPL_EMACS_EXTRA_WORDS_MOVE
+            backward_word:
+                #endif
```
Gadgetoid, line 298: ???

Gadgetoid: The following addition to the `fixup_c` function catches these edge cases, but I don't love it-

```patch
diff --git a/tools/codeformat.py b/tools/codeformat.py
index e2e5c0cc0..97b466d06 100755
--- a/tools/codeformat.py
+++ b/tools/codeformat.py
@@ -109,8 +109,13 @@ def fixup_c(filename):
                 directive = m.group(2)
                 if directive in ("if ", "ifdef ", "ifndef "):
                     l_next = lines[0]
+                    label = re.match(f"^ +[a-z_]+:", l_next)
                     indent_next = len(re.match(r"( *)", l_next).group(1))
-                    if indent - 4 == indent_next and re.match(r" +(} else |case )", l_next):
+                    if label:
+                        l = l[indent:]
+                        lines[0] = lines[0].lstrip()
+                        dedent_stack.append(0)
+                    elif indent - 4 == indent_next and re.match(r" +(} else |case )", l_next):
                         # This #-line (and all associated ones) needs dedenting by 4 spaces.
                         l = l[4:]
                         dedent_stack.append(indent - 4)
```

extmod/nimble/nimble/nimble_npl_os.c
Gadgetoid: Looking at other examples I think the preference here would be to rewrite this code as follows:

```c
#if MICROPY_MALLOC_USES_ALLOCATED_SIZE
m_free(alloc, alloc->size);
#else
m_free(alloc);
#endif
```

shared/readline/readline.c
projectgus: What about the slightly more verbose:

```c
#if MICROPY_REPL_EMACS_EXTRA_WORDS_MOVE
            case 'b':
            backward_word:
#elif MICROPY_REPL_EMACS_WORDS_MOVE
            case 'b':
#endif
```

... to avoid nesting multiple #ifdefs in these?

projectgus: Or, as `forward_word` and `backward_word` are labels, maybe they could always be included but marked as unused to avoid the warning (not sure how to do that for labels, though...)

projectgus: I don't know what level of crime this is, but it seems to work and doesn't rely on any compiler-specific pragmas:

```c
            #if MICROPY_REPL_EMACS_WORDS_MOVE
            case 'b':
                goto backward_word; // Ensure label always used
backward_word:
                redraw_step_back = cursor_count_word(0);
                rl.escape_seq = ESEQ_NONE;
                break;
            case 'f':
                goto forward_word; // Ensure label is always used
forward_word:
                redraw_step_forward = cursor_count_word(1);
                rl.escape_seq = ESEQ_NONE;
                break;
            case 'd':
                vstr_cut_out_bytes(rl.line, rl.cursor_pos, cursor_count_word(1));
                redraw_from_cursor = true;
                rl.escape_seq = ESEQ_NONE;
                break;
            case 127:
                goto backward_kill_word; // Ensure label is always used
backward_kill_word:
                redraw_step_back = cursor_count_word(0);
                vstr_cut_out_bytes(rl.line, rl.cursor_pos - redraw_step_back, redraw_step_back);
                redraw_from_cursor = true;
                rl.escape_seq = ESEQ_NONE;
                break;
            #endif
```

Gadgetoid:
I should probably rebase this :grimacing:

dpgeorge:
> I should probably rebase this 😬

Yes please!

Gadgetoid:
Hmm.

1. This seems to have dramatically increased in scope since last I ran ci_c_code_formatting_run
2. My computer and CI disagree about code formatting
3. My local version of Uncrustify (Homebrew on a Mac) is 0.79.0_f
4. CI is running 0.72.0_f

Trying again from my Pi which has 0.72.0.

Looks like rules can change quite dramatically between versions, so we should look into some way of enforcing either the specific Uncrustify version, or the rules from a specific version- otherwise well-meaning contributors could run the code formatting run and submit code that fails CI... as I just did.

projectgus:
> Looks like rules can change quite dramatically between versions, so we should look into some way of enforcing either the specific Uncrustify version, or the rules from a specific version- otherwise well-meaning contributors could run the code formatting run and submit code that fails CI... as I just did.

Yeah, it's not ideal and can be generally frustrating. I put a [small note in the docs](https://github.com/micropython/micropython/blob/master/CODECONVENTIONS.md#uncrustify) when I ran across the same thing, but I think most contributors won't expect it.

I did a bit of work earlier this year building uncrustify for WASM, and had it running under wasmtime. This creates the interesting possibility of publishing uncrustify packages on pypi for easy install, and switching the pre-commit hook to use the correct uncrustify version under WASM by default. I stopped working it, but maybe I should pick it back up.

(Performance is... OK. The overhead is mostly startup time unfortunately, so it does add a perceptible delay when running pre-commit against small patches. It's probably not enough of a delay to make it infeasible, though, and would always include an option to use native uncrustify instead.)

dlech:
> publishing uncrustify packages on pypi for easy install

It's not WASM, but it is already there. https://pypi.org/project/micropython-uncrustify/ The binary wheels should cover the vast majority of developer machines and it has a proper source package if anyone want to build it somewhere exotic.

Happy to transfer the project to MicroPython maintainership and save you the work.

projectgus:
@dlech Ah sorry I'm crossing back and forth the various discussion streams about this. Have just left some related replies at https://github.com/micropython/micropython/pull/15221#issuecomment-2159599968 and https://github.com/orgs/micropython/discussions/14496#discussioncomment-9733050

Gadgetoid:
I'm still missing the proposed changes to `shared/readline/readline.c` but I've rebased this to catch up with the last *checks notes* four months of updates.

shared/readline/readline.c
```diff
@@ -330,9 +330,9 @@ int readline_process_char(int c) {
         } else {
             rl.escape_seq = ESEQ_NONE;
             if (c == 'A') {
-#if MICROPY_REPL_EMACS_KEYS
-up_arrow_key:
-#endif
+                #if MICROPY_REPL_EMACS_KEYS
+            up_arrow_key:
+                #endif
```
Gadgetoid, line 335: Curses!

Gadgetoid: I'm tempted to skip changes to readline.c in favour of making a separate PR to clean up these warts. It feels a little out of scope to apply a manual rewrite here?

projectgus: @Gadgetoid That seems like a very good idea to me. If you're going to make that change then feel free to request a review after that and I'll take another pass over it. Thanks for perservering!

dpgeorge:
@Gadgetoid I see this PR is still in draft status.  I think the only thing remaining to do was skip the changes to `readline.c`, ie do that at a later time.

But also this will need rebasing on latest master, which may pick up some new things.

Do you have time to do this?  If not I can pick it up (but I'll need to make a new PR, I cannot edit this one).  Just let me know either way.

Gadgetoid:
Rebased, removed changes to readline and added it as an exclusion, re-run over the latest files. Wish me luck.

tools/codeformat.py
dpgeorge, line 47: Now that most of `shared/` is formatted, it's probably worth combining all these into:
```
    "shared/**/*.[ch]",
```
That way, any new directories added to `shared/` are not missed.

(You may need to add an exclusion for `shared/memzip` though... or reformat that as well).

dpgeorge:
Thanks @Gadgetoid for updating the PR.  I guess it can be brought out of draft now?

dpgeorge (approved):

dpgeorge:
Rebased and merged in 828a3b8946ed94216185c6f2f0ff017c788d6acf, 8d31abc79df7f719227e5e0a0227d2f9115078ec and 48c7daaa3d7068c127c5bc395ffa72029ab9bba1

PR #14074, opened 2024-03-12
extmod/nimble: Check for active before setting address mode.

DvdGiessen:
`BLE().config(addr_mode=...)` is not safe to call if the NimBLE stack is not yet active (because it tries to acquire mutexes which should be initialized first).

### DIFF
--- a/extmod/nimble/modbluetooth_nimble.c
+++ b/extmod/nimble/modbluetooth_nimble.c
@@ -728,6 +728,9 @@ void mp_bluetooth_get_current_address(uint8_t *addr_type, uint8_t *addr) {
 }
 
 void mp_bluetooth_set_address_mode(uint8_t addr_mode) {
+    if (!mp_bluetooth_is_active()) {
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks, this looks OK.  I see that in the `examples/bluetooth/ble_bonding_peripheral.py` example it explicitly activates the BLE before setting `addr_mode`:
```py
        self._ble.irq(self._irq)
        self._ble.config(bond=True)
        self._ble.config(le_secure=True)
        self._ble.config(mitm=True)
        self._ble.config(io=_IO_CAPABILITY_DISPLAY_YESNO)
        self._ble.active(True)
        self._ble.config(addr_mode=2)
```
So, that example still works with the change in this PR.

DvdGiessen:
Indeed, this only adds a proper error that can be caught in Python (instead of likely triggering a null pointer dereference and crashing in NimBLE), so it only serves to prevent the user from doing something they can't really do anyway and giving a clear error message, similar to how such "protection" exist for other methods of the BLE class and as is already noted in the documentation.

PR #14075, opened 2024-03-12
ports/esp32: Check if main heap allocation failed.

DvdGiessen:
If the heap allocation fails we will crash if we continue, so at least we can show a clear error message so one can figure out memory allocation was the problem (instead of just seeing some arbitrary null pointer error later).


dpgeorge:
Thanks.

PR #14076, opened 2024-03-12
ports/Makefile: Reinstate -O0 option on demand.

stinos:
-Og introduced in for example 9ffb1ad and c5dbbf7 results in a binary with debug symbols, but essentially breaks debugging with gdb since it still allows the compiler to optimize temporaries for instance, so gdb will report 'optimized out' for a lot of variables and breakpoints will often halt at the wrong line.
-O0 has none of these problems so allow it by specifying DEBUG=2, leaving the original DEBUG=1 functionality as-is.

See for instance https://stackoverflow.com/questions/63386189/whats-the-difference-between-a-compilers-o0-option-and-og-option for some more information.

I'm not sure if DEBUG=2 is a common way of doing this, but it's definitely pretty simple. Likewise: don't know if all ports I changed this for even work with gdb?

In theory `-g3` is also better than `-g` but at least with `-O0 -g` stepping through the source and looking at variables and memory all just seems to work fine enough, whereas `-Og` is just not usable for getting detailed information on a problem quickly.

### DIFF
--- a/mpy-cross/Makefile
+++ b/mpy-cross/Makefile
@@ -25,9 +25,9 @@ CFLAGS += -fdata-sections -ffunction-sections -fno-asynchronous-unwind-tables
 # Debugging/Optimization
 ifdef DEBUG
 CFLAGS += -g
-COPT = -O0
+COPT ?= -O0
 else
-COPT = -Os #-DNDEBUG
+COPT ?= -Os #-DNDEBUG
 endif
 
 # On OSX, 'gcc' is a symlink to clang unless a real gcc is installed.
--- a/ports/bare-arm/Makefile
+++ b/ports/bare-arm/Makefile
@@ -16,13 +16,14 @@ PYDFU ?= $(TOP)/tools/pydfu.py
 CFLAGS += -I. -I$(TOP) -I$(BUILD)
 CFLAGS += -Wall -Werror -std=c99 -nostdlib
 CFLAGS += -mthumb -mtune=cortex-m4 -mcpu=cortex-m4 -msoft-float
-CSUPEROPT = -Os # save some code space for performance-critical code
+CFLAGS += $(COPT)
+CSUPEROPT ?= -Os # save some code space for performance-critical code
 
 # Select debugging or optimisation build.
 ifeq ($(DEBUG), 1)
-CFLAGS += -Og
+COPT ?= -Og
 else
-CFLAGS += -Os -DNDEBUG
+COPT ?= -Os -DNDEBUG
 CFLAGS += -fdata-sections -ffunction-sections
 endif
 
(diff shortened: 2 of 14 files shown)


iabdalkader:
Can we consider adding `-ggdb3` to include macro definitions ? Without those you have to calculate register addresses (some base + some other base + offset) manually and then cast just to get a register value.

stinos:
> Can we consider adding -ggdb3 

You mean in the same way, e.g `DEBUG=1` -> `-g`, `DEBUG=2` -> -ggdb3 ?

> Without those you have to calculate register addresses (some base + some other base + offset) manually

Does that also work with `-g3`? I'm not sure we want to hardcode `gdb` here.

robert-hh:
How about an additional compile flags option, which csn be set at the make command line. Maybe it exists already.

iabdalkader:
> You mean in the same way, e.g `DEBUG=1` -> `-g`, `DEBUG=2` -> -ggdb3 ?

I'm not sure, maybe, but if you use `O0` it's likely you also want macros.

> Does that also work with `-g3`? I'm not sure we want to hardcode `gdb` here.

The manual says:
`Level  3  includes  extra  information,  such as all the macro definitions present in the program.  Some debuggers support macro expansion when you use -g3.`

stinos:
> How about an additional compile flags option, which csn be set at the make command line. ~Maybe it exists already.~ Seem you can use CFLAGS for that purpose, since all settings for CFLAGS in makefile are `CFLAGS +=`.

Since the docs say

> If you use multiple -g options, with or without level numbers, the last such option is the one that is effective. 

then that's probably not going to work since the ones passed on the commandline come first?

dlech:
Can we just use `COPT` to override the optimization consistently across ports to allow any optimization rather than trying to make new options to try to cover all cases?

I'm already used to doing `make DEBUG=1 COPT=-O0` for the unix port.

dlech:
> is the idea behing `COPT` that it stands for `C optimization options'?

Yes, that is the idea. Basically, just the `-O` flag.

> And why is `-g` in CFLAGS not COPT?

Because it isn't the `-O` flag.

dpgeorge:
I think the simplest thing is to make more consistent use of `COPT` across the ports, allow it to be overridden on the command line, and document it.

And, I'm happy to change `-g` to `-g3` in the relevant Makefile's.  But I think `-ggdb3` should be specified by the user on the command line if they need that.

stinos:
> I think the simplest thing is to make more consistent use of COPT across the ports, allow it to be overridden on the command line

Ok, changed that already.


>  and document it.

In gettingstarted.rst?

It's now only in unix/README.md but that also seems to be the only one which documents DEBUG=1 whereas most ports have that.


> But I think -ggdb3 should be specified by the user on the command line if they need that.

Sounds reasonable but that should be via `CFLAGS_EXTRA` then I guess, which not all ports have now and the ones which do have it do not use it consistently as 'append to CFLAGS', making it impossible to override `-g` (e.g. unix Makefile does a `CFLAGS += -g` after the line where it appends `CFLAGS_EXTRA` to `CFLAGS`)

Issue #14005, opened 2024-03-02
Different behavior on V1.22.2 over V1.19


Tangerino:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

ESP32

### MicroPython version

Using the official V1.22.2

### Reproduction

This is a very complete program using:
- Ethernet
- 4G
- 2 serial ports
- WiFI
- BLE

### Expected behaviour

Not panic attacks

### Observed behaviour

The same code used on 2K devices on V1.18 and V1.19

![CleanShot 2024-03-02 at 09 43 02](https://github.com/micropython/micropython/assets/3164100/7a64d536-f444-45a0-84a8-22f1e5ceefa4)

Now trying V1.22 (the official image) in few seconds running I get this.

But don't know where to look

Guru Meditation Error: Core  1 panic'ed (Cache disabled but cached memory region accessed). 

Core  1 register dump:
PC      : 0x401ec634  PS      : 0x00060035  A0      : 0x800901c6  A1      : 0x3ffc03a0  
A2      : 0x3fffa4b8  A3      : 0x3fff057c  A4      : 0x00000006  A5      : 0x3ffc03e8  
A6      : 0x00000001  A7      : 0x3fffa59c  A8      : 0x8009a475  A9      : 0x3ffc0380  
A10     : 0x3fffa4b8  A11     : 0x3fffa3e4  A12     : 0x3fffa3ea  A13     : 0x3fffa3e4  
A14     : 0xfffffff7  A15     : 0x00060023  SAR     : 0x00000018  EXCCAUSE: 0x00000007  
EXCVADDR: 0x00000000  LBEG    : 0x400973d4  LEND    : 0x400973f0  LCOUNT  : 0xffffffff  

[...]

PR #12729, opened 2023-10-18
extmod/network_ppp: Add stream config parameter.

DvdGiessen:
This makes the stream that the PPP object wraps, which is normally only set once via the constructor, accessible and configurable via the `.config()` method.

It also allows the stream to be set to `None`, which essentially stops all PPP communication without disconnecting the session.

This allows retrieving the stream, and more interesting, replacing it on-the-fly to suspend it, for example to send AT commands to a modem without completely disconnecting and re-establishing the PPP connection:

```py
uart = ppp.config('stream')
ppp.config(stream=None)
uart.write(b'+++')
# do some AT commands
uart.write(b'ATO\r\n')
ppp.config(stream=uart)
```

Any attempted communication by PPP while the UART is not connected will register as simple packet loss to the LwIP stack because we return 0 for any write calls, and protocols like TCP will then automatically handle retrying.

EDIT: Reworked to also support the generic network.PPP implementation alongside the ESP32 one.

PR #14086, opened 2024-03-13
ports/esp32: Print support information on panic.

DvdGiessen:
When a fatal error occurs it's important to know which precise version it occurred on in order to be able to decode the crash dump information such as the backtrace.

I noticed it today in https://github.com/micropython/micropython/issues/14005#issuecomment-1979962621, and quickly prototyped a small improvement.

By wrapping around the built-in IDF panic handler we can print some extra information whenever a fatal error occurs. The message link to [a new wiki page](https://github.com/micropython/micropython/wiki/ESP32-debugging) which contains additional information on how to debug ESP32 issues links to the bug reporting issue template.

Example of current output:

```

MPY version : v1.23.0-preview.213.gfbf668cf3d.dirty on 2024-03-13
IDF version : v5.2.1
Machine     : GlobalLink 520 Bx with ESP32S3

Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

... (snip)
```

(Another suggestion is we could perhaps mention the SHA256 hash of each ESP32 image ELF file on the website. The IDF prints the "ELF file SHA256" at the end of each fatal error message, making it easy to look up the corresponding ELF file.)

### DIFF
--- a/ports/esp32/panichandler.c
+++ b/ports/esp32/panichandler.c
@@ -1,0 +1,68 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2024 Mindhash B.V.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "genhdr/mpversion.h"
+#include "py/mpconfig.h"
+
+#ifdef MICROPY_BUILD_TYPE
+#define MICROPY_BUILD_TYPE_PAREN " (" MICROPY_BUILD_TYPE ")"
+#else
+#define MICROPY_BUILD_TYPE_PAREN
+#endif
+
+#ifdef CONFIG_ESP_PANIC_HANDLER_IRAM
+#define MICROPY_WRAP_PANICHANDLER_FUN(f) IRAM_ATTR f
+#define MICROPY_WRAP_PANICHANDLER_STR(s) DRAM_STR(s)
+#else
+#define MICROPY_WRAP_PANICHANDLER_FUN(f) f
+#define MICROPY_WRAP_PANICHANDLER_STR(s) (s)
+#endif
+
+void __real_esp_panic_handler(void *);
+void esp_panic_handler_reconfigure_wdts(uint32_t timeout_ms);
+void panic_print_str(const char *str);
+
+void MICROPY_WRAP_PANICHANDLER_FUN(__wrap_esp_panic_handler)(void *info) {
+    esp_panic_handler_reconfigure_wdts(1000);
+
+    const static char *msg = MICROPY_WRAP_PANICHANDLER_STR(
+        "\r\n"
+        "A fatal error occurred. The crash dump printed below may be used to help\r\n"
+        "determine what caused it. If you are not already running the most recent\r\n"
+        "version of MicroPython, consider upgrading. New versions often fix bugs.\r\n"
+        "\r\n"
+        "To learn more about how to debug and/or report this crash visit the wiki\r\n"
+        "page at: https://github.com/micropython/micropython/wiki/ESP32-debugging\r\n"
+        "\r\n"
+        "MPY version : " MICROPY_GIT_TAG " on " MICROPY_BUILD_DATE MICROPY_BUILD_TYPE_PAREN "\r\n"
+        "IDF version : " IDF_VER "\r\n"
+        "Machine     : " MICROPY_HW_BOARD_NAME " with " MICROPY_HW_MCU_NAME "\r\n"
+        "\r\n"
+        );
+    panic_print_str(msg);
+
+    __real_esp_panic_handler(info);
+}
--- a/ports/esp32/CMakeLists.txt
+++ b/ports/esp32/CMakeLists.txt
@@ -59,5 +59,8 @@ include($ENV{IDF_PATH}/tools/cmake/project.cmake)
 # Set the location of the main component for the project (one per target).
 list(APPEND EXTRA_COMPONENT_DIRS main_${IDF_TARGET})
 
+# Enable the panic handler wrapper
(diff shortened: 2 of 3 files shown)


andrewleech:
This does look truly helpful, thanks. Do you know a convenient way to trigger a panic at the moment to test it?

DvdGiessen:
One I know of is to make use of the fact that the PPP implementation can call into MicroPython code in its own task without first setting up a proper thread state or holding the GIL (see my notes in https://github.com/micropython/micropython/pull/12729#issuecomment-1969045333). So it we give it a zero-sized BytesIO, then make it write to it, that will trigger an reallocation of the BytesIO buffer which will fail and crash:

```py
from network import PPP
from io import BytesIO
ppp = PPP(BytesIO(0))
ppp.active(True)
ppp.connect()
```

andrewleech:
I tested this on my esp32c6 and it worked as intended.

I would suggest though rather than the message saying to raise an issue, it might be better to print a link to a (new) wiki page about these crashes, that could then describe how to raise an issue (if needed) and could even include the details on how to decode the dump.

DvdGiessen:
> But I guess it doesn't work well on esp32-s2/s3, which have native USB. Because the USB disappears when there's a crash.

That might indeed be a limitation depending on how the chip is connected, but that's not necessarily always the case for these chips. For example: I've tested this on a ESP32-S3 based board which uses the native USB serial/JTAG console (`CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG=y`) and there it works fine (the chip resets but doesn't lose the USB connection while doing so).

> it might be better to print a link to a (new) wiki page about these crashes

Agreed, that does indeed seem like a better solution. Something like this? Open to suggestions.

```
```

DvdGiessen:
I've just written a basic wiki page on debugging crashes on the ESP32, and updated the panic message in this MR accordingly.

https://github.com/micropython/micropython/wiki/ESP32-debugging

ports/esp32/panichandler.c
```diff
@@ -0,0 +1,42 @@
+#include "genhdr/mpversion.h"
```
dpgeorge, line 1: Can you please add a license/copyright header?  You can copy it from, say, `esp32_partition.c` but replace the copyright line with your name and the current year.

Walkline80:
> ```python
> from network import PPP
> from io import BytesIO
> ppp = PPP(BytesIO(0))
> ppp.active(True)
> ppp.connect()
> ```

I tested this code on my esp32c3 board (idf5.0.4) with this pr code, it's crashed, but there's no extra message showed, I wonder why?

DvdGiessen:
> it's crashed, but there's no extra message showed

Did the default ESP-IDF panic handler show up, or did it just reboot? The default panic handler output usually starts with `Guru Meditation Error` followed by some debug information, and the final line is `Rebooting...`. (See the documentation for v5.0.4 on the C3 [here](https://docs.espressif.com/projects/esp-idf/en/v5.0.4/esp32c3/api-guides/fatal-errors.html) and the code [here](https://github.com/espressif/esp-idf/blob/v5.0.4/components/esp_system/panic.c#L228-L407).)

If that didn't show, then you're likely losing some output, perhaps as mentioned above due to the USB connection going away before the error output has been received. That is not something this specific PR aims to change.

dpgeorge:

Walkline80:
> or did it just reboot?

Yes, just like a normally hard reset.

I tested another board that connected to PC by using USB CDC mode, and this time the extra message and `Guru Meditation Error` showed up.

DvdGiessen:
@andrewleech Indeed, I wrote the wiki page mostly based on the output of the Xtensa-based chips. For the RISC-V baseed chips there isn't a straight up backtrace printed, and I think instead `idf.py monitor` just decodes the MEPC address.

To be more precise, the monitor doesn't seem to know what it is decoding. I think it looks for things that look like addresses in the output and then prints a line for those, without actually knowing something is a backtrace, register dump, etc. The package implementing it is [`esp_idf_panic_decoder`](https://github.com/espressif/esp-idf-panic-decoder), and it's used through the [`handle_possible_pc_address_in_line`](https://github.com/search?q=repo%3Aespressif%2Fesp-idf-monitor+handle_possible_pc_address_in_line&type=code) method.

So for RISC-V based chips, the wiki should probably indeed explain to look at the `MEPC` (Machine Exception Program Counter) register. That value can be passed to `addr2line` to find the source code line where the crash happened. To get a full backtrace, attaching a debugger as described at the bottom of the page is probably the easiest way?

projectgus:
Thanks for all your work on this, @DvdGiessen! Even though I wrote the first version of idf_monitor many years ago, I even learned something new reading your wiki page.

> For the RISC-V baseed chips there isn't a straight up backtrace printed, and I think instead `idf.py monitor` just decodes the MEPC address.

Actually idf.py monitor can do a bit more than this. Here's the backtrace from ESP32-C3 if I insert an `abort()` into the IDF hello_world:

```
Hello world!

abort() was called at PC 0x42007655 on core 0
0x42007655: app_main at /home/gus/ry/george/esp-idf-v5/examples/get-started/hello_world/main/hello_world_main.c:40

Stack dump detected
Core  0 register dump:
MEPC    : 0x403806d0  RA      : 0x40383a70  SP      : 0x3fc8fc50  GP      : 0x3fc8b600  
0x403806d0: panic_abort at /home/gus/ry/george/esp-idf-v5/components/esp_system/panic.c:466
0x40383a70: __ubsan_include at /home/gus/ry/george/esp-idf-v5/components/esp_system/ubsan.c:313

TP      : 0x3fc87230  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130  
S0/FP   : 0x00000004  S1      : 0x3fc8fcb4  A0      : 0x3fc8fc7c  A1      : 0x3fc8fcb2  
A2      : 0x00000000  A3      : 0x3fc8fca9  A4      : 0x00000001  A5      : 0x3fc8d000  
A6      : 0x7a797877  A7      : 0x76757473  S2      : 0x00000000  S3      : 0x00000000  
S4      : 0x00000000  S5      : 0x00000000  S6      : 0x00000000  S7      : 0x00000000  
S8      : 0x00000000  S9      : 0x00000000  S10     : 0x00000000  S11     : 0x00000000  
T3      : 0x6e6d6c6b  T4      : 0x6a696867  T5      : 0x66656463  T6      : 0x62613938  
MSTATUS : 0x00001881  MTVEC   : 0x40380001  MCAUSE  : 0x00000007  MTVAL   : 0x00000000  
0x40380001: _vector_table at ??:?

MHARTID : 0x00000000  


Backtrace:


panic_abort (details=details@entry=0x3fc8fc7c "abort() was called at PC 0x42007655 on core 0") at /home/gus/ry/george/esp-idf-v5/components/esp_system/panic.c:466
466         *((volatile int *) 0) = 0; // NOLINT(clang-analyzer-core.NullDereference) should be an invalid operation on targets
#0  panic_abort (details=details@entry=0x3fc8fc7c "abort() was called at PC 0x42007655 on core 0") at /home/gus/ry/george/esp-idf-v5/components/esp_system/panic.c:466
#1  0x40383a70 in esp_system_abort (details=details@entry=0x3fc8fc7c "abort() was called at PC 0x42007655 on core 0") at /home/gus/ry/george/esp-idf-v5/components/esp_system/port/esp_system_chip.c:93
#2  0x4038961a in abort () at /home/gus/ry/george/esp-idf-v5/components/newlib/abort.c:38
#3  0x42007658 in app_main () at /home/gus/ry/george/esp-idf-v5/examples/get-started/hello_world/main/hello_world_main.c:40
#4  0x42014b8a in main_task (args=<error reading variable: value has been optimized out>) at /home/gus/ry/george/esp-idf-v5/components/freertos/app_startup.c:208
#5  0x40384686 in vPortTaskWrapper (pxCode=<optimized out>, pvParameters=<optimized out>) at /home/gus/ry/george/esp-idf-v5/components/freertos/FreeRTOS-Kernel/portable/riscv/port.c:255
ELF file SHA256: 22ad941aa
```

The monitor actually recognises the block of stack memory, reprocesses it, and dumps it back out as a gdb backtrace.

I think the relevant code is here:
https://github.com/espressif/esp-idf-monitor/blob/master/esp_idf_monitor/base/serial_handler.py#L218

I assume this is the same on C6, but I don't actually have a C6 at hand yet to test (on its way).

I'm not sure what the best MicroPython-friendly solution here is. The easiest one is obviously to run `idf.py -B build-generic_C6 monitor` and make it crash in the monitor, but that's obviously a bit limiting.

I think the best would be to build a little command line decoder tool that someone can paste a stack trace into, that calls the idf_panic_handler functions internally.

Issue #13092, opened 2023-11-29
ESP32-S3 compile warnings


kdschlosser:
Compiling version 1.21.0 using ESP-IDF 5.0.4 for an esp32-s3 with octal SPIRAM I am getting the following warnings. 

This is the make command I am using

```
make BOARD=ESP32_GENERIC_S3 BOARD_VARIANT=SPIRAM_OCTAL
```

And here are the warnings. 
```
In file included from  /micropython/ports/esp32/managed_components/espressif__tinyusb/src/osal/osal_freertos.h:33,
                 from  /micropython/ports/esp32/managed_components/espressif__tinyusb/src/osal/osal.h:57,
                 from  /micropython/ports/esp32/managed_components/espressif__tinyusb/src/tusb.h:38,
                 from  /micropython/ports/esp32/managed_components/espressif__esp_tinyusb/include/tinyusb.h:11,
                 from  /micropython/ports/esp32/usb.c:35:
 /micropython/lib/berkeley-db-1.xx/PORT/include/queue.h:85: warning: "LIST_INIT" redefined
   85 | #define LIST_INIT(head) {                                               \
      |
In file included from  /esp-idf/components/esp_wifi/include/esp_private/esp_wifi_types_private.h:20,
                 from  /esp-idf/components/esp_wifi/include/esp_wifi_types.h:11,
                 from  /micropython/ports/esp32/mpconfigport.h:13,
                 from  /micropython/py/mpconfig.h:75,
                 from  /micropython/py/mpstate.h:31,
                 from  /micropython/py/runtime.h:29,
                 from  /micropython/ports/esp32/usb.c:27:
 /.espressif/tools/xtensa-esp32s3-elf/esp-2022r1-11.2.0/xtensa-esp32s3-elf/xtensa-esp32s3-elf/sys-include/sys/queue.h:563: note: this is the location of the previous definition
  563 | #define LIST_INIT(head) do {                                            \
      |
In file included from  /micropython/ports/esp32/managed_components/espressif__tinyusb/src/osal/osal_freertos.h:33,
                 from  /micropython/ports/esp32/managed_components/espressif__tinyusb/src/osal/osal.h:57,
                 from  /micropython/ports/esp32/managed_components/espressif__tinyusb/src/tusb.h:38,
                 from  /micropython/ports/esp32/managed_components/espressif__esp_tinyusb/include/tinyusb.h:11,
                 from  /micropython/ports/esp32/usb.c:35:
[...]

Issue #6186, opened 2020-06-24
Pyboard D build does not include btree


peterhinch:
The MICROPY_PY_BTREE option is specified for ESP8266 and ESP32, but not for the Pyboard D. Is this intentional?

dpgeorge:
The main reason is due to the licensing conditions of the berkeley-db code, one of the clauses is:
```
 * 3. All advertising materials mentioning features or use of this software
 *    must display the following acknowledgement:
 *      This product includes software developed by the University of
 *      California, Berkeley and its contributors.
```
and that puts constraints on marketing of pyboard.

But from what I understand this clause was removed from all Berkeley code, so would be good to make that change to the license in this library code, then it could be included in builds without a problem.

In the meantime, `btree` can be built as a native .mpy file:
```
$ cd examples/natmod/btree
$ make ARCH=armv7m
```
Then copy `btree_armv7m.mpy` to the board as `btree.mpy`.

peterhinch:
This is another example of https://github.com/micropython/micropython/issues/5564. In response to a forum query I suggested **btree**, only for the user to discover that it wasn't available on his PBD. Embarrassing.

We need a quick way to check whether feature X is available on platform Y. The unappetising choice is to trawl the build system or plug in a board (if we have one) and try it.

fxwiegand:
> This is another example of #5564. In response to a forum query I suggested **btree**, only for the user to discover that it wasn't available on his PBD. Embarrassing.
> 
> We need a quick way to check whether feature X is available on platform Y. The unappetising choice is to trawl the build system or plug in a board (if we have one) and try it.

Almost three years later and this is still a problem. Took me quite some time to find out that `btree` isn't available on the pico. Not to mention that its a proper adventure to create a firmware that includes it only finding out that it's not working properly on the pico.

robert-hh:
btree was added to the esp8266 port after a poll in the user community. I never understood why, because the ESP8266 is very limited in resources, resulting in a limited implementation. Did you try if it works different on the Pico than on the ESP8266 or ESP32?

fxwiegand:
I haven't tried it on any of the ESP ones. Only on the pico compared to the unix port. I have talked to @iabdalkader on the micropython discord who also mentioned that there are problems with the btree module on the pico.

iabdalkader:
[...]

PR #14099, opened 2024-03-15
lib/berkeley-db: Update submodule to improve header include paths, remove symlinks, update BSD license

dpgeorge:
This PR makes the following changes to the berkeley-db submodule and associated `modbtree.c` code:
- moves the berkeley-db header files to a single directory within that submodule, and references all these headers with a much fuller path, which prevents clashes (eg with esp32 and `queue.h`)
- removes unused/non-working files from berkeley-db, which removes all symlinks in that repo (symlinks don't play well under Windows)
- allows injecting an external configuration header into berkeley-db, so the configuration doesn't have to be provided by `-Dxx-yy` flags to the compiler (and possibly clashing with other symbols)
- removes the advertising clause from the BSD 4-clause license of berkeley-db (see relevant commit and README.Impt.License.Change for details)
- updates `modbtree.c` and `esp8266` port to build with these changes

TODO:
- [x] need to create a new berkeley-db repo within the micropython github organisation to host these berkeley-db changes


dpgeorge:
This will fix #13092 (ESP32 compiler warnings) and help with #6186 (enabling btree on stm32).

PR #14103, opened 2024-03-15
ports/esp32: Skip codespell for IDF managed components.

DvdGiessen:
During a build the ESP-IDF downloads managed components in the `ports/esp32/managed_components` directory, which shouldn't be spellchecked by us.


dpgeorge:
Thanks for this, codespell checking the IDF managed components has annoyed me a few times!

PR #14082, opened 2024-03-13
Add external Arduino MicroPython library.

iabdalkader:
This library contains modules used by Arduino boards. Depends on:

- [x] https://github.com/micropython/micropython/pull/12961
- [x] https://github.com/arduino/arduino-lib-mpy/pull/7

PR #14120, opened 2024-03-18
extmod/modopenamp: Implement MIMXRT port and log handler fixes.

iabdalkader:
* Logging is fixed by using the existing metal log handling mechanism instead of overriding the metal_log, which causes build issues when logging is enabled.
* A port is implemented for `mimxrt` and tested with IMXRT1176 custom board. A test firmware for M4 can be found here https://github.com/iabdalkader/openamp_vuart (build with `make PORT=mimxrt`).


dpgeorge:
@iabdalkader Is this streaming interface important for your use of the `openamp` module?  If not, then this would be a low priority for us.

extmod/modopenamp.c
```diff
@@ -140,6 +144,12 @@ static int endpoint_recv_callback(struct rpmsg_endpoint *ept, void *data, size_t
     endpoint_obj_t *self = metal_container_of(ept, endpoint_obj_t, ep);
     if (self->callback != mp_const_none) {
         mp_call_function_2(self->callback, mp_obj_new_int(src), mp_obj_new_bytearray_by_ref(len, data));
+    } else {
+        for (size_t i=0; i<len; i++) {
+            if (ringbuf_put(&self->ringbuf, ((uint8_t *) data)[i]) == -1) {
```
dpgeorge, line 149: The problem with a ring buffer used in this way is that the boundaries of the message are lost.

The behaviour of the endpoint should match UDP sockets, and a recv/read should return a single packet/message.  Eg if 2 messages are pending on the endpoint then the code needs to do 2 separate reads to get them both.

iabdalkader:
> @iabdalkader Is this streaming interface important for your use of the `openamp` module? If not, then this would be a low priority for us.

No not at all, it's just something @jimmo asked for in the reviews.

extmod/modopenamp.c
iabdalkader: Yeah I thought about it, but what if you call `recv` with a size that's less than the size of the next pending message ?

dpgeorge: Then the message is truncated to fit the requested number of bytes, and the rest of that message is discarded (that's how UDP recv/recvfrom work).

dpgeorge:
> However you may want to look at the log issue, second commit.

I had a look at this and see the problem.  Can we fix that by using the proper logging interface, and calling `metal_set_log_handler()` and passing in a function that wraps `mp_printf()`?

iabdalkader:
> Can we fix that by using the proper logging interface

Yes, I'll send it in a different PR.

> Is this streaming interface important for your use of the openamp module?

Note the only blocker right now is https://github.com/micropython/micropython/pull/14082

extmod/modopenamp.c
iabdalkader: Okay, updated. I've also found a way to avoid double copying, see comment message.

dpgeorge:
I see that this now has mimxrt support, which is good, but it means that will be blocked on the streaming part of the API.

If you want to get the logging and/or mimxrt support merged faster then please put them in a separate PR.

iabdalkader:
> If you want to get the logging and/or mimxrt support merged faster then please put them in a separate PR.

I kept them here in this PR and sent a new one for endpoint streaming https://github.com/micropython/micropython/pull/14181

ports/mimxrt/mpremoteprocport.c
```diff
@@ -0,0 +1,164 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * modremoteproc mimxrt port.
+ */
+
+#include <stdio.h>
+#include <stdint.h>
```
dpgeorge, line 30: I don't think either of these headers are needed (this file doesn't use `printf`).

dpgeorge:
There are some remaining `DEBUG_printf` in `extmod/modopenamp_remoteproc_store.c` that should probably be changed to `metal_log()` calls.  Also `extmod/modopenamp_remoteproc.c` has a `#define DEBUG_printf` that's never used.

iabdalkader:
> There are some remaining `DEBUG_printf` in `extmod/modopenamp_remoteproc_store.c` that should probably be changed to `metal_log()` calls. Also `extmod/modopenamp_remoteproc.c` has a `#define DEBUG_printf` that's never used.

I removed them all, as well as some `debug_printf`'s in `extmod/modopenamp.c` that we've missed in the initial review. I think it's much better to control the whole log with a single switch.

PR #14123, opened 2024-03-18
lwip: socket: Use Nagle algorithm and add support for TCP_NODELAY.

greezybacon:
This adds support to use the Nagle algorithm implemented already in LwIP to determine when a TCP should be sent.

I was confused as to why there was a constant ~200ms lag when sending small packets from TCP in Micropython. After too much searching, I arrived at the TCP_TMR_INTERVAL which sends packets at its usual timeout interval of 250ms. Changing this to 50ms makes a corresponding difference in the lag but still doesn't offer a solid solution on reducing the latency when small packets are desired.

I looked at adding the TCP_NODELAY socket option for LwIP, but that also didn't work until I found that it isn't used without calling `tcp_output_nagle`. And Micropython currently has a static algorithm to issue packet sends with the send buffer is below 25%. 

It seems to me we should be using the common Nagle algorithm to determine when best to send packets based on best performance / latency? I'm of course open to other ideas and happy to provide some before+after tests based on performance and/or latency if interested.

Maybe not said enough, thanks to all who have put together such an incredible project!

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -69,12 +69,20 @@
 #define IP_ADD_MEMBERSHIP 0x400
 #define IP_DROP_MEMBERSHIP 0x401
 
+#define TCP_NODELAY TF_NODELAY
+
 // For compatibilily with older lwIP versions.
 #ifndef ip_set_option
 #define ip_set_option(pcb, opt)   ((pcb)->so_options |= (opt))
 #endif
 #ifndef ip_reset_option
 #define ip_reset_option(pcb, opt) ((pcb)->so_options &= ~(opt))
+#endif
+#ifndef tcp_set_flags
+#define tcp_set_flags(pcb, set_flags) do { (pcb)->flags |= (set_flags); } while (0)
+#endif
+#ifndef tcp_clear_flags
+#define tcp_clear_flags(pcb, clear_flags) do { (pcb)->flags &= ~(clear_flags); } while (0)
 #endif
 
 // A port can define these hooks to provide concurrency protection
@@ -742,9 +750,10 @@ static mp_uint_t lwip_tcp_send(lwip_socket_obj_t *socket, const byte *buf, mp_ui
         MICROPY_PY_LWIP_REENTER
     }
 
-    // If the output buffer is getting full then send the data to the lower layers
-    if (err == ERR_OK && tcp_sndbuf(socket->pcb.tcp) < TCP_SND_BUF / 4) {
-        err = tcp_output(socket->pcb.tcp);
+    // Use nagle algorithm to determine when to send segment buffer (can be
+    // disabled with TCP_NODELAY socket option)
+    if (err == ERR_OK) {
+        err = tcp_output_nagle(socket->pcb.tcp);
     }
 
     MICROPY_PY_LWIP_EXIT
@@ -1440,6 +1449,17 @@ static mp_obj_t lwip_socket_setsockopt(size_t n_args, const mp_obj_t *args) {
             break;
         }
 
+        // level: IPPROTO_TCP
+        case TCP_NODELAY: {
+            mp_int_t val = mp_obj_get_int(args[3]);
+            if (val) {
+                tcp_set_flags(socket->pcb.tcp, opt);
+            } else {
+                tcp_clear_flags(socket->pcb.tcp, opt);
+            }
+            break;
+        }
+
         default:
             printf("Warning: lwip.setsockopt() not implemented\n");
     }
@@ -1829,6 +1849,9 @@ static const mp_rom_map_elem_t mp_module_lwip_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR_IPPROTO_IP), MP_ROM_INT(0) },
     { MP_ROM_QSTR(MP_QSTR_IP_ADD_MEMBERSHIP), MP_ROM_INT(IP_ADD_MEMBERSHIP) },
     { MP_ROM_QSTR(MP_QSTR_IP_DROP_MEMBERSHIP), MP_ROM_INT(IP_DROP_MEMBERSHIP) },
+
+    { MP_ROM_QSTR(MP_QSTR_IPPROTO_TCP), MP_ROM_INT(IP_PROTO_TCP) },
+    { MP_ROM_QSTR(MP_QSTR_TCP_NODELAY), MP_ROM_INT(TCP_NODELAY) },
 };
 
 static MP_DEFINE_CONST_DICT(mp_module_lwip_globals, mp_module_lwip_globals_table);


greezybacon:
@dpgeorge sorry if my story-telling got the best of the situation.

250ms is too long. 50ms is also too long. If the controller provides a small response to a small message/request, it should be sent immediately- or at least be able to.

dpgeorge:
> If the controller provides a small response to a small message/request, it should be sent immediately- or at least be able to.

Yes, I agree.  But it's hard to balance that requirement against good throughput when there is lots of data being written in small chunks.

Well, that's what this PR/discussion is about!

greezybacon:
But that’s the reason for the no-delay “option”. It allows the application designer to make the decision on throughput / latency for the IP stack.

greezybacon:
[Nagle’s algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) is a well studied, general solution to the latency / throughput problem. It was published in RFC 896.

I don’t pretend to be an expert- just want to provide context:

When using it, it determines that packets should NOT be sent immediately if there is unacknowledged data previously sent. This allows the application to buffer data until either (a) the send buffer exceeds the max-segment-size (MSS) or (b) the receiver acknowledges all previously-sent packets.

This allows the sender to buffer writes using a dynamic threshold based on the latency of the receiver.

As previously written, Micropython would only send packets if there was <25% remaining in the send buffer. Using it, sending a small message of ~50 bytes would not trigger a send. So it would remained queued until the TCP interval timer expires.

Using Nagle’s algorithm, the first write, no matter how small, will generate a packet. Until that data is acknowledged, all further writes will be queued until one of the two criteria is met. Additionally, when the ACK is received from the receiver, it will trigger the queued data to be sent. So, if the MSS has not yet been reached by the application, then the maximum throughput of the application is still reached along with a reasonable latency given the delay distance of the receiver. If the MSS is reached by the app in the meantime, packets may be sent until the TCP window size remains unacknowledged.

In addition to this, an application designer may choose to always create packets for every write by setting the TCP_NODELAY socket option. But even without that option, Nagle’s algorithm will provide much more reasonable latency from Micropython.

greezybacon:
@dpgeorge I wanted to provide a somewhat scientific approach to testing the throughput/latency problem. I constructed these two programs to setup a simple character-generator and echo-service to test the send throughput and response latency respectively. The character generator service sends 16-byte packets continuously to a connected client. The echo-service will read a 16-byte message and then respond with the received message.

Results on RP2040 with WIZnet W5500
-----

|build|character generator|echo service|
|------|------|-------|
| v1.23.0-preview.237.g7dff38fdc | 371 kiB/s | 21 msg/s  (~48ms response time)|
| This PR (without TCP_NODELAY) | 357 kiB/s | 240 msgs/s (~4.2ms response time) |
|------|------|-------|
| Diff | 96.2% (-3.77%) | 1143% |


So, to your point, it does impact the maximal throughput of an application that makes many small writes. However, that impact is rather small considering that for message-based applications it makes an 11 fold increase in the performance 
and response time.

------

Server:
```python
import socket
def network_service(handler, port=8888, backlog=2):
    host = socket.getaddrinfo('0.0.0.0', port)[0]
    s = socket.socket()
    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
    s.bind(host[-1])
    s.listen(backlog)
    #
    while True:
        try:
            s2, addr = s.accept()
            handler(s2)
        except KeyboardInterrupt:
            s.close()
            break

def chargen(s):
    data = b'a' * 16
    try:
        while True:
            for char in range(33, 65):
                s.write(data)
    except KeyboardInterrupt:
        raise
    finally:
        print("Client disconnected")
        s.close()

def echo(s):
    empty = b''
    try:
        while True:
            buf = s.read(16)
            if buf == empty:
                break
            s.write(buf)
    except KeyboardInterrupt:
        raise
    finally:
        print("Client disconnected")
        s.close()

import network
nic = network.WIZNET5K()
nic.active(True)
nic.ifconfig('dhcp')
print(f"Network up at {nic.ifconfig()[0]}")

network_service(chargen, port=8888)
```

And the client (running CPython)
```python
async def downloader(host, port, timeout=10):
    reader, writer = await asyncio.open_connection(host, port)
    #
    start = time.monotonic()
    end = start + timeout
    bytes = 0
    try:
        while time.monotonic() < end:
            bytes += len(await reader.read(512))
    finally:
        now = time.monotonic()
        avg = bytes / (now - start)
        print(f"Read {bytes >> 10}kiB bytes (avg {int(avg) >> 10}kiB/s) in {now - start:.1f}s")
        writer.close()

async def echo(host, port, timeout=10):
    reader, writer = await asyncio.open_connection(host, port)
    #
    start = time.monotonic()
    end = start + timeout
    msgs = 0
    try:
        while time.monotonic() < end:
            writer.write(b'a' * 16)
            await writer.drain()
            await reader.read(16)
            msgs += 1
    finally:
        now = time.monotonic()
        avg = msgs / (now - start)
        print(f"Read {msgs} msgs (avg {int(avg)}msgs/s) in {now - start:.1f}s")
        writer.close()
```

For the `chargen` service, I got these results on a RP2040 with WIZnet W5500:

This PR:
```
>>> asyncio.run(downloader(mphost, 8888))
Read 3229kiB bytes (avg 358kiB/s) in 10.0s
>>> asyncio.run(downloader(mphost, 8888))
Read 3238kiB bytes (avg 359kiB/s) in 10.0s
>>> asyncio.run(downloader(mphost, 8888))
Read 3206kiB bytes (avg 356kiB/s) in 10.0s
```

v1.23.0-preview.237.g7dff38fdc
```
>>> asyncio.run(downloader(mphost, 8888))
Read 3316kiB bytes (avg 368kiB/s) in 10.0s
>>> asyncio.run(downloader(mphost, 8888))
Read 3421kiB bytes (avg 380kiB/s) in 10.0s
>>> asyncio.run(downloader(mphost, 8888))
Read 3355kiB bytes (avg 372kiB/s) in 10.0s
```

For the `echo` service:

This PR:
```
>>> asyncio.run(echo(mphost 8888))
Read 1984 msgs (avg 220msgs/s) in 0.9s
>>> asyncio.run(echo(mphost, 8888))
Read 2358 msgs (avg 262msgs/s) in 0.9s
>>> asyncio.run(echo(mphost, 8888))
Read 2194 msgs (avg 243msgs/s) in 0.9s
>>> asyncio.run(echo(mphost, 8888))
Read 2140 msgs (avg 237msgs/s) in 0.9s
```

v1.23.0-preview.237.g7dff38fdc
```
>>> asyncio.run(echo(mphost, 8888))
Read 186 msgs (avg 20msgs/s) in 0.4s
>>> asyncio.run(echo(mphost, 8888))
Read 190 msgs (avg 21msgs/s) in 0.5s
>>> asyncio.run(echo(mphost, 8888))
Read 204 msgs (avg 22msgs/s) in 0.0s
```

greezybacon:
@dpgeorge if it would help, I'd be happy to remove the TCP_NODELAY option.

dpgeorge:
This is a very good improvement for short TCP messages, while still maintaining good link efficiency for lots of data.  And I think it's useful to have the TCP_NODELAY option so users have some control over how TCP output works, which can be important for IoT devices.

So, this is merged!

vshymanskyy:
Now, it should be fairly easy to port it to ESP32.
Is it planned or a PR would be welcome?

greezybacon:
@vshymanskyy I believe this code will work consistently for any NIC which is using lwIP stack.

vshymanskyy:
Yeagh but ESP32 uses a different socket implementation:
https://github.com/micropython/micropython/blob/master/ports/esp32/modsocket.c

So I guess it needs to be adjusted separately

dpgeorge:
> do you know the reason the ESP32 architecture couldn't be adjusted to use the `modlwip.c` implementation?

Because `extmod/modlwip.c` is for non-OS configurations/builds of lwIP.  ESP32 uses an RTOS and configures lwIP in OS mode, so ESP32 cannot use this extmod code.

But actually ESP32 should already enable the Nagle algorithm as part of its configuration.  Then it's just a matter of exposing the TCP_NODELAY constant.

greezybacon:
That makes sense. I implemented it in #15107

PR #14127, opened 2024-03-19
Support board level configurations

gentijo:
This PR allows ESP32 board level configuration at both the idf_component.yml file and the USER_C_MODULES.

The idf_component.yml control occurs by looking for a main directory that also incorporates the MICROPY_BOARD variable as part of the name  main_{arch} (i.e. main_esp32) , if a folder does not exist then it falls back to previous behaviour.

For the USER_C_MODULES, the usermod.cmake will look in the USER_C_MODULES directory for ${MICROPY_BOARD}_micropython.cmake, if that does not exist then it will look for a micropython.cmake after that it will fall back to the previous behavior to do a directory scan.  

This will allow boards to specify which USER_C_MODULES will be incorporated and allow unused modules to exist without adding extra code to the build.

### DIFF
--- a/ports/esp32/CMakeLists.txt
+++ b/ports/esp32/CMakeLists.txt
@@ -66,4 +66,4 @@ else()
 endif()
 
 # Define the project.
-project(micropython)
+project(micropython_${MICROPY_BOARD})
--- a/ports/esp32/CMakeLists.txt
+++ b/ports/esp32/CMakeLists.txt
@@ -57,7 +57,13 @@ set(SDKCONFIG_DEFAULTS ${CMAKE_BINARY_DIR}/sdkconfig.combined)
 include($ENV{IDF_PATH}/tools/cmake/project.cmake)
 
 # Set the location of the main component for the project (one per target).
-list(APPEND EXTRA_COMPONENT_DIRS main_${IDF_TARGET})
+if(EXISTS ${CMAKE_CURRENT_LIST_DIR}/${MICROPY_BOARD}_main_${IDF_TARGET})
+    message("Including main Dir: ${MICROPY_BOARD}_main_${IDF_TARGET}")
+    list(APPEND EXTRA_COMPONENT_DIRS ${MICROPY_BOARD}_main_${IDF_TARGET})
+else()
+    message("Including main Dir: main_${IDF_TARGET}")
+    list(APPEND EXTRA_COMPONENT_DIRS main_${IDF_TARGET})
+endif()
 
 # Define the project.
 project(micropython)
--- a/py/usermod.cmake
+++ b/py/usermod.cmake
@@ -36,7 +36,15 @@ function(usermod_gather_sources SOURCES_VARNAME INCLUDE_DIRECTORIES_VARNAME INCL
 endfunction()
 
 # Include CMake files for user modules.
-if (USER_C_MODULES)
+if (EXISTS ${USER_C_MODULES}/${MICROPY_BOARD}_micropython.cmake)
+    message("Including User C Module(s) from ${USER_C_MODULE_PATH} for Board ${MICROPY_BOARD}")
+    include(${USER_C_MODULES}/${MICROPY_BOARD}_micropython.cmake)
+
+elseif(EXISTS ${USER_C_MODULES}/micropython.cmake)
+    message("Including User C Module(s) from ${USER_C_MODULE_PATH} from micropython.cnake")
+    include(${USER_C_MODULES}/micropython.cmake)
+
+elseif (USER_C_MODULES)
     foreach(USER_C_MODULE_PATH ${USER_C_MODULES})
         message("Including User C Module(s) from ${USER_C_MODULE_PATH}")
         include(${USER_C_MODULE_PATH})


dpgeorge:
For customising the set of user C modules for a board, you should already be able to just definite `USER_C_MODULES` in the boards `mpconfigboard.cmake` file.

For the IDF component customisation, I also think you can use the `mpconfigboard.cmake` file for that.  Can you explain exactly what you are trying to do here?

gentijo:
Hi @dpgeorge thank you for your time providing feedback. 

At the BOARD_DIR level, each of the CMakeLists.txt include esp32_common.cmake, at the bottom of esp32_common, it defines the "main" component where the granularity is at the IDF_TARGET level, main_esp32, esp32C*, esp32S*

idf_component.yml files are an ESP IDF thing and can only be specified within a component where the main component is the component that we have the most control of.  
The board directory is not considered an idf component so the build process won't look for an idf_component.yml there.
 
What I have done is added logic to test if a main directory exists consisting of the name ${BOARD_NAME}_${IDF_TARGET}_main, if it exists then set that as the main component directory otherwise fallback to the ${IDF_TARGET}_main directory as it originally did.  This allows for a for a build to utilize a board specific main directory which opens the possibility to have a per board idf_component.yml

USER_C_MODULES are similar but different in scope.  The build process allows me to declare a USER_C_MODULE directory where the build process with scan that directory for other module directories and compile them in. 

What it doesn't allow me to do is have one USER_C_MODULE that has multiple modules, then at the board level specify which of those modules should be included.  

The change here was to test for the existence of cmake file with the naming of ${BOARD_NAME}_micropython.cmake, if that file exists then include it, and allow that file to selectively include  modules in the directory. Otherwise fall back to the previous behavior. 

TLDR;
My project is focused around Micropython support for MicroROS and I currently have 3 distinct build profiles.

1) Core profile,  includes ROSMicroPy as a USER_C_MODULE that provides the micropython interface for the microros idf component.

2) Add to core, LVGL as an IDF component and my C Module to provide the MP interface, also a C Module to provide access to Nintendo JoyCon that requires the Bluetooth modules.

3) a ROS camera version which includes the Core module plus esp32-camera as an IDF module and my User module to provide the MP interface. 

This page provide a better picture of my build profiles https://github.com/gentijo/ROSMicroPy/blob/main/README.md

An interesting fact that took me down this path is that, on the ESP platform, IDF components can not be User C Modules because IDF components are compiled before the QSTR scan occurs causing unknown symbol errors.  So for modules that use the IDF version of a CMakeList.txt file are treated differently than User C Modules that have a different CMakeList.txt structure.  For my build profiles, I need to both specify a idf_component.yml to include the necessary IDF style components (LVGL / microROS) and the User C module that provides the Micropython interface for the module.

cnadler86:
Is this still ongoing? I found this approach interesting, for example to include esp32 cam driver in the IDF yaml config on board level, rather than on Port Variant level. And I can imagine there are more use cases than this.

gentijo:
Hi @cnadler86,

I did not hear back on the acceptance of my PR so I have not perused it. With that said, I think I can acheive most of it without any mods. 

1) I would move the board dir to my project dir, a parent of micropython, then specify a relative board dir in ../../../board-dir.  Micropython is a git submodule of my project.   Then in my  mconfigboard.cmake file, I would would add another component dir similar to what is in ports/esp32/CMakelist.txt  i.e.
    list(APPEND EXTRA_COMPONENT_DIRS {my_board_main})
So that I have another component that allows me to have a idf_component.yml file, and I think the board dir and the main dir can be one and the same.

One thing I learned in looking at this a second time is that c_user_module can be a list if you specify it in your mpconfigboard.cmake, this eliminates the need to point to a cmake file so that you could be selective.  I was under the assumption that if you set a user_c_module path, it looks at all the subdirs, making it hard to have one set of modules for config a vs a different set of modules for config b.   It think it does scan the directory for modules, but if you c_user_module list points to individual directories, it will work. 

The only thing you don't get is a way to specify the name for the build file which is a concern for me because for rosmicropy. I currently have 5 different build configurations for different hardware configs and soon that will 4x in size. 

It would be nice if we could provide a variable in our mpconfigboard dirs that would override the output filename if supplied otherwise it would default to micropython.bin. 

I haven't tested what I just said, I will soon but I think it would work.

-John

cnadler86:
Hi John,

Thank you for your advice on how to possibly register a User Module as IDF component. After reading the documentation I think I finally understand what you mean (I am relative new in the embedded field). 

I will try this out in the next days and let you know if I succeed building my User Module als IDF component with its own sdkconfig file and dependencies in a separate yml file.

gentijo:
@cnadler86 glad to help, feel free to reach out if you have any questions. 

You may want to look at this module I wrote, it's unique in that it is a hybrid C User module, where the critical parts are written in C++ but the rest of it is written in Python then frozen (converted to assembler) and included with the module.  It really came in handy to reduce the C Module complexity as the Python components did a lot of the heavy lifting dealing with dict / list classes, while the lower level is in C. Had this all been implemented in C++  it would of been a lot more code because I did not need to extract and validate variables as much
https://github.com/gentijo/libMicroPy_JSONForms


It's all part of this project that has several other modules that depend on ESP32 components. 
https://github.com/ROSMicroPy/ROSMicroPy/

PR #14133, opened 2024-03-19
samd: Some fixes to the clock configuration.

robert-hh:
- Set the clock config option for the ADAFRUIT TRINKET M0 and MINISAM M4 boards to use the USB SOF as reference.
- Fix the clock configuration after (not because) enabling Python USB drivers. This change affects the clock generation on SAMD board without a crystal. These try to use the USB SOF signal as reference. But if after a while no USB connection is established, it falls back to an internal clock reference. It turns out that on some SAMD21 board the clock is then far off the proper value, and USB fails. This change moves as well the check for the fallback after USB has been initialized.
- Change a wrong MCU name in README.md.


dpgeorge:
Thanks for this, it all looks pretty good (although I did not test it).

I have rebased and merged this with the following minor changes:
- renamed MICROPY_HW_SPIFLASH_BAUD to MICROPY_HW_SPIFLASH_BAUDRATE to match other config options that have BAUDRATE in their name
- made `sync_dfll48_with_xosc32kulp()` static
- fixed small formatting error
- removed MICROPY_HW_XOSC32K comment at end of `#endif` because the `#if` was changed to MICROPY_HW_DFLL_USB_SYNC

robert-hh:
Thank you very much.

PR #14134, opened 2024-03-20
Add new samd21 board Adafruit QT PY

charkster:
New board definition. Adafruit QT PY (samd21e18a).


charkster:
I am not experienced in modifying commit comments to conform to automatic testing. I guess I have to amend it. Is there any way to do this in the web GUI?
Thanks.

PR #14135, opened 2024-03-20
Add new board. Adafruit QT PY.

charkster:
New board added Adafruit QT PY.


robert-hh:
A few things:
The commit message is wrong. a) The header line has the wrong format, and B) the signature must have a personal E-Mail address.
The header line could be e.g. `samd/boards: Add support for the Adafruit QT Py board.`

About the board files: Since the QT Py board can optionally be equipped with am external flash chip, you could forsee a build variant with support for that device. That requires additional changes to all board files. and testing. I will come up with an untested suggestion later.

robert-hh:
So here are two zip file. One with the modified board files and one with a modified Makefile. The latter was required to change the build path for variants. The files build, but lacking boards I could not test it.
[Makefile.zip](https://github.com/micropython/micropython/files/14663789/Makefile.zip)
[ADAFRUIT_QT_PY.zip](https://github.com/micropython/micropython/files/14663792/ADAFRUIT_QT_PY.zip)

charkster:
> So here are two zip files. One with the modified board files and one with a modified `Makefile`. The latter change was required to change the build path for variants. The files build, but lacking boards I could not test it. [Makefile.zip](https://github.com/micropython/micropython/files/14663789/Makefile.zip) [ADAFRUIT_QT_PY.zip](https://github.com/micropython/micropython/files/14667359/ADAFRUIT_QT_PY.zip)
> 
> Edit: Tested by wiring a 8M flash chip to a SAMD21 board. I had a typo in mpconfigboard.mk, which is fixed now.

Thank you. I will attach an external flash tomorrow (GD25Q16 2M) and verify the additional pins and make options. 
Would I need to edit mpconfigboard.mk with this change:

`MICROPY_HW_CODESIZE = 2M`

I have not used a make option before... would it look like this:

`make BOARD=ADAFRUIT_QT_PY BOARD_VARIANT=SPIFLASH`

Would it be simpler to replace BOARD_VARIANT with SPIFLASH and then specify the memory size? Something like this:

`make BOARD=ADAFRUIT_QT_PY SPIFLASH=2M`

charkster:
> Would it be simpler to replace BOARD_VARIANT with SPIFLASH and then specify the memory size? Something like this:
> 
> `make BOARD=ADAFRUIT_QT_PY SPIFLASH=2048K`

I was thinking that if SPIFLASH is specified on the make command line (including a value), it would take precedence over MICROPY_HW_CODESIZE specified in mpconfigboard.mk . I prefer to specify the flash size on the make command line, otherwise I have to remember to check the value of MICROPY_HW_CODESIZE in mpconfigboard.mk even if I don't use the BOARD_VARIANT in the make command. If I first build a variant with an external memory and then build another version without a memory, I must remember to edit mpconfigboard.mk the second time. I also like having the exact external flash size specified in the build directory name.
```
# Strip the letter 'K' from MICROPY_HW_CODESIZE for use by C code.
ifneq ($(SPIFLASH),)
CFLAGS += -DMICROPY_HW_CODESIZE=$(strip $(subst K,' ', $(SPIFLASH)))
LDFLAGS += --defsym=_codesize=$(SPIFLASH)
else
CFLAGS += -DMICROPY_HW_CODESIZE=$(strip $(subst K,' ', $(MICROPY_HW_CODESIZE)))
LDFLAGS += --defsym=_codesize=$(MICROPY_HW_CODESIZE)
endif

LDFLAGS += -nostdlib $(addprefix -T,$(LD_FILES)) -Map=$@.map --cref
```

robert-hh:
> I will attach an external flash tomorrow (GD25Q16 2M) and verify the additional pins and make options.

You can use any external flash with a size between 1M and 16M. I used an 8M flash (GD25Q64). The actual flash size will be determined at boot-up time. The external flash will be used only for the file system, not for code.

> Would I need to edit mpconfigboard.mk with this change: `MICROPY_HW_CODESIZE = 2048K`

No. Do not touch MICROPY_HW_CODESIZE. It is NOT related to the size of the external flash. It specifies, how much of the internal 256k flash is used for the firmware. At devices without external flash for the file system, this is 184k. At devices with external flash, it's 248k (256k - 8k for the bootloader), and then more MicroPython features are enabled.

The BOARD_VARIANT scheme is used to ensure, that all variants are built for the firmware download site. So yes, the make command is:

make BOARD=ADAFRUIT_QT_PY BOARD_VARIANT=SPIFLASH -j8

The symbol `SPIFLASH` must match the name in board.json. The -j8 speeds up compiling by using 8 threads for the build.

Summary: The settings in the board file and Makefile are sufficient. It happened here in the test of the "board + flash on a breadboard" set-up, that the SPI baud rate was too high. That can be changed in the file samd_spiflash.c, line 139. It is hard coded at 24MHz. Changing it to 12MHz worked. I made a PR to configure that by a compile symbol, but that may take a while to get merged.

charkster:
Thanks @robert-hh. I have my QT PY board with the 2M SPIFLASH soldered and new uf2 programmed. I can see lots of spiflash.py libraries for other MCU, but nothing included in SAMD. Is the memory (and size) detectable with built-in classes? I made [my own detailed driver a couple of years ago](https://github.com/charkster/GD25Q16C), but I would prefer to use the standard MicroPython classes.
I did see your [Samd on-board flash support](https://github.com/orgs/micropython/discussions/10137) post. I like your idea about using the STM32 spiflash on SAMD. If you have any example code for SPI Flash, I would like to try it (Google searches did not yield any good examples).

robert-hh:
There is no need to add any driver for SPI flash. If you have built the firmware with the board files I have sent you with the command:

`make BOARD=ADAFRUIT_QT_PY BOARD_VARIANT=SPIFLASH -j8`

then the built-in spiflash driver for the flash based file system (file samd_spiflash.c) is active and will use the flash. You can check with:

```
import os
print(os.statvfs("/"))
```

It should return something like:

`(4096, 4096, 512, 502, 502, 0, 0, 0, 0, 255)`

Total size 512*4096 bytes.

robert-hh:
I got a QT PY board today, soldered a flash chip to it, and installed the SPIRAM variant. It works fine. I had only 4MB, 8MB  and 16MB flash chips, so I soldered 8Mb chip.

charkster:
> There is no need to add any driver for SPI flash. If you have built the firmware with the board files I have sent you with the command:
> 
> `make BOARD=ADAFRUIT_QT_PY BOARD_VARIANT=SPIFLASH -j8`
> 
> then the built-in spiflash driver for the flash based file system (file samd_spiflash.c) is active and will use the flash. You can check with:
> 
> ```
> import os
> print(os.statvfs("/"))
> ```
> 
> It should return something like:
> 
> `(4096, 4096, 512, 502, 502, 0, 0, 0, 0, 255)`
> 
> Total size 512*4096 bytes. If you get something like `(0,0,0,0,0,0,0,0,0,255`, then the flash chip is not working. I had that in the beginning with my breadboard test set-up, requiring to reduce the clock speed. (samd_spiflash.py, line 139).

@robert-hh Thank you.
My GigaDevices GD25Q16C 2MB flash IC was not detected, even after reducing the speed to 12MHz (and then 6Mhz) in samd_spiflash.py . The capacity should be readable in id[2] (3rd byte read back from command 0x9F)... SFDP is also supported by this memory (same command 0x5A)
```
import os
print(os.statvfs("/"))
(0, 0, 0, 0, 0, 0, 0, 0, 0, 256)
```
I was careful to solder it with the correct orientation (CS is pin 1, which matches the dot on the QT PT's silkscreen). The SECTOR_SIZE is definitely different (512 vs 4096), and it uses 24bit addresses. I could not see any issues with your pin.csv (matches the schematic perfectly). 

Any ideas you have about this would be appreciated. Thanks again.

robert-hh:
> The SECTOR_SIZE is definitely different (512 vs 4096), and it uses 24bit addresses. 

That's strange. The same chip is e.g. used by the ItsyBitsyM0 and works there. According to the data sheet of that device, chapter 1 and 3, the sector size is 4096. This is hard defined in the code. So if you changed that, change it back, and check again the soldering.
If you want to test with Python driver, you can use the firmware w/o spiflash support and try e.g.t this driver: https://github.com/robert-hh/SPI_Flash
I used that alternatively to the C-code version. At the QT PY board, the SPI device to be used is 3.

charkster:
> > The SECTOR_SIZE is definitely different (512 vs 4096), and it uses 24bit addresses.
> 
> That's strange. The same chip is e.g. used by the ItsyBitsyM0 and works there. According to the data sheet of that device, chapter 1 and 3, the sector size is 4096. This is hard defined in the code. So if you changed that, change it back, and check again the soldering. If you want to test with Python driver, you can use the firmware w/o spiflash support and try e.g.t this driver: https://github.com/robert-hh/SPI_Flash I used that alternatively to the C-code version. At the QT PY board, the SPI device to be used is 3.

Sorry, you are correct that the SECTOR_SIZE is 4096 bytes. I was thinking about the sector count being 512 (512 * 4096 = 2097152 bytes). The only schematic difference that I can see between my QT PY and Itsybitsy M0 is the resistor on the CS pin. HOLD and WP both have the 3.3V connection (I am measuring that correctly).

Can you commit your files for the QT PY board support? It's possible that I have an error somewhere in my clone. I'll just plan on pulling the changes.
Thanks.

robert-hh:
I created a branch in my repository which contains all changes to the board files and Makefile. I've just build and tested the firmware to be sure, that it detects the external flash. The change to Makefile just takes care, that the build directory is named after the board variant. SO you may skip that for the moment.

https://github.com/robert-hh/micropython/tree/samd_qt_py

dpgeorge:
Support for QT Py was added in 6cbe145ca8b46857cd2ba5289021022bb92cb02e

PR #14136, opened 2024-03-20
py/parse: Zero dangling parse tree pointer to fix possible leak.

projectgus:
This fixes a bug where a random Python object may become un-garbage-collectable until an enclosing Python file (compiled on device) finishes executing:

## Details

The `mp_parse_tree_t` structure is stored on the stack in top-level functions such as parse_compile_execute() in pyexec.c (and others).

Although it quickly falls out of scope in these functions, it is usually still in the current stack frame when the compiled code executes. (Compiler dependent, but usually it's one stack push per function.)

This means if any Python object happens to allocate at the same address as the (freed) root parse tree chunk, it's un-garbage-collectable as there's a (dangling) pointer up the stack referencing this same address.

As reported by @GitHubsSilverBullet here:
https://github.com/orgs/micropython/discussions/14116#discussioncomment-8837214

### DIFF
--- a/py/parse.c
+++ b/py/parse.c
@@ -1386,6 +1386,7 @@ void mp_parse_tree_clear(mp_parse_tree_t *tree) {
         m_del(byte, chunk, sizeof(mp_parse_chunk_t) + chunk->alloc);
         chunk = next;
     }
+    tree->chunk = NULL; // Avoid dangling pointer that may live on stack
 }
 
 #endif // MICROPY_ENABLE_COMPILER


projectgus:
I did a quick audit of other functions calling `m_del` in similar context and didn't find any obvious ones, however I think it's possible for something similar to happen with any normal C automatic variables:

1. Function A allocates a 16 byte stack frame, uses 3 words of it.
2. The highest word is a pointer to the heap which is freed.
3. Function A returns.
4. Function B is called, allocates a 16 byte stack frame (same address on the stack as A), only uses 2 words of it, calls into child function C.
5. The pointer from step 2 is still present in valid stack memory from now until function B returns.

I think the only way around this reliably would be a `m_free_and_null(p)` function that nulls out the argument pointer. May be worth considering?

projectgus:
> This means if any Python object happens to allocate at the same address 

It's not even "at" same address I realise, it only has to overlap the dangling pointer address.

dpgeorge:
Thank you!



> I think the only way around this reliably would be a `m_free_and_null(p)` function that nulls out the argument pointer. May be worth considering?

Maybe...

dpgeorge:
> I think the only way around this reliably would be a `m_free_and_null(p)` function that nulls out the argument pointer. May be worth considering?

That wouldn't have helped this case if it were written as:
```c
    m_free_and_null(&chunk);
```
because `chunk` is a local variable.  Instead it would have needed some more sophisticated code for the un-link list logic.

projectgus:
@dpgeorge good point, so yeah maybe it's OK to do it piecemeal if/when these things come up.

PR #14137, opened 2024-03-20
wiznet5k: w5100s: Properly enable interrupt signal

greezybacon:
According to [the datasheet](https://docs.wiznet.io/img/products/w5100s/w5100s_ds_v125e.pdf), page 29, the IEN bit to enable the interrupt is in the MR2 register, not the MR register.

This is just cleanup as the interrupt appears to be enabled by default after resetting the chip.

### DIFF
--- a/extmod/network_wiznet5k.c
+++ b/extmod/network_wiznet5k.c
@@ -221,7 +221,7 @@ static void wiznet5k_init(void) {
         setSn_IMR(0, Sn_IR_RECV);
         #if _WIZCHIP_ == W5100S
         // Enable interrupt pin
-        setMR(MR2_G_IEN);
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the fix.  It looks correct but I did not test it.  Did you test it on hardware?

greezybacon:
I did. I would be very happy to provide any sort of output. I have the
W5100S-EVB-PICO I tested it on.

Cheers,
Jared

dpgeorge:
> I have the W5100S-EVB-PICO I tested it on.

OK, great, thanks for confirming.  The fact you tested it is enough.

greezybacon:
Of course. It’s exciting to be able to contribute.

PR #14140, opened 2024-03-20
docs: Update docs to replace ifconfig with ipconfig.

felixdoerre:
Follow up for #13689, to update the documentation to refer to `ipconfig` instead of `ifconfig` where appropriate.

I've left `wipy` out, because I couldn't find, which network stack it is using. Also, I am not sure which network stack exactly `network.WLAN` and `network.LAN` are referring to. Should I just copy the adjustments form `network.rst` over to `network.WLAN`/`network.LAN`?

### DIFF
--- a/docs/esp8266/tutorial/network_basics.rst
+++ b/docs/esp8266/tutorial/network_basics.rst
@@ -19,10 +19,10 @@ You can check if the interfaces are active by::
 
 You can also check the network settings of the interface by::
 
-    >>> ap_if.ifconfig()
-    ('192.168.4.1', '255.255.255.0', '192.168.4.1', '8.8.8.8')
+    >>> ap_if.ipconfig('addr4')
+    ('192.168.4.1', '255.255.255.0')
 
-The returned values are: IP address, netmask, gateway, DNS.
+The returned values are: IP address and netmask.
 
 Configuration of the WiFi
 -------------------------
@@ -45,8 +45,8 @@ To check if the connection is established use::
 
 Once established you can check the IP address::
 
-    >>> sta_if.ifconfig()
-    ('192.168.0.2', '255.255.255.0', '192.168.0.1', '8.8.8.8')
+    >>> sta_if.ipconfig('addr4')
+    ('192.168.0.2', '255.255.255.0')
 
 You can then disable the access-point interface if you no longer need it::
 
@@ -64,7 +64,7 @@ connect to your WiFi network::
             sta_if.connect('<ssid>', '<key>')
             while not sta_if.isconnected():
                 pass
-        print('network config:', sta_if.ifconfig())
+        print('network config:', sta_if.ipconfig('addr4'))
 
 Sockets
 -------
--- a/docs/library/network.WIZNET5K.rst
+++ b/docs/library/network.WIZNET5K.rst
@@ -13,7 +13,7 @@ Example usage::
 
     import network
     nic = network.WIZNET5K(pyb.SPI(1), pyb.Pin.board.X5, pyb.Pin.board.X4)
-    print(nic.ifconfig())
+    print(nic.ipconfig("addr4"))
 
     # now use socket as usual
     ...
@@ -51,20 +51,7 @@ Constructors
 Methods
 -------
 
-.. method:: WIZNET5K.isconnected()
-
-   Returns ``True`` if the physical Ethernet link is connected and up.
-   Returns ``False`` otherwise.
-
-.. method:: WIZNET5K.ifconfig([(ip, subnet, gateway, dns)])
-
-   Get/set IP address, subnet mask, gateway and DNS.
-
-   When called with no arguments, this method returns a 4-tuple with the above information.
-
-   To set the above values, pass a 4-tuple with the required information.  For example::
-
-    nic.ifconfig(('192.168.0.4', '255.255.255.0', '192.168.0.1', '8.8.8.8'))
+This class implements most methods from `AbstractNIC <AbstractNIC>`, which are documented there. Additional methods are:
 
 .. method:: WIZNET5K.regs()
 
--- a/docs/library/network.rst
+++ b/docs/library/network.rst
@@ -24,7 +24,7 @@ For example::
         print("Waiting for connection...")
         while not nic.isconnected():
             time.sleep(1)
-    print(nic.ifconfig())
+    print(nic.ipconfig("addr4"))
 
     # now use socket as usual
     import socket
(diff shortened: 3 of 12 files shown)


docs/esp8266/tutorial/network_basics.rst
```diff
@@ -19,8 +19,8 @@ You can check if the interfaces are active by::
 
 You can also check the network settings of the interface by::
 
-    >>> ap_if.ifconfig()
-    ('192.168.4.1', '255.255.255.0', '192.168.4.1', '8.8.8.8')
+    >>> ap_if.ipconfig('addr4')
+    ('192.168.4.1', '255.255.255.0')
 
 The returned values are: IP address, netmask, gateway, DNS.
```
robert-hh, line 25: According to the example, gateway and DNS are not returned.

dpgeorge:
Thanks for writing these docs, definitely needed!

But, `ipconfig()` is not yet implemented on esp32 or esp8266 ports.  So the docs there shouldn't change.  And it means `mpremote` still needs to use the old `ifconfig()`.

Well, probably it would be better to implement `ipconfig()` on all the ports first, then we can make a consistent move to that new method, in the docs and everywhere.

dpgeorge:
These are the network interfaces that need `ipconfig()` added to them:
- `extmod/network_ninaw10.c`
-  `ports/cc3200/mods/modwlan.c`
- `ports/esp32/network_lan.c`
- `ports/esp32/network_ppp.c`
- `ports/esp32/network_wlan.c`
- `ports/esp8266/network_wlan.c`
- `ports/mimxrt/network_lan.c`
- `ports/stm32/network_lan.c`

Some of those (eg stm32, mimxrt) are very easy.

robert-hh:
> These are the network interfaces that need ipconfig() added to them:

I can start working in these. I have all the boards needed for testing.

felixdoerre:
Yes, @robert-hh, that would be very nice. I have none of these boards :). So I will not touch the networking code for the other ports, that you can do and test the necessary changes.

dpgeorge:
To start with I suggest just implementing the IPv4 config options for `ipconfig()`, on those additional NICs.  That should be relatively straightforward, and would be enough to allow this docs PR to be merged.

robert-hh:
So mimxrt and stm32 were indeed very easy. There others are more work, when they do not use the common LWIP stack. So the ipconfig() method has to be fully implemented, which will take a while.

felixdoerre:
I've started a PR (#14199) to update the configuration function for the other network interfaces. Their current state is also not too consistent. The lwip-variants should all be included in there. I am not sure how consistent we will be able to make the new `ipconfig` interface.

> But, ipconfig() is not yet implemented on esp32 or esp8266 ports. So the docs there shouldn't change. And it means mpremote still needs to use the old ifconfig().

Judging from what is documented (and being a kind-of devil's advocate), we'd only need to implement the `esp-netif` netconfig for esp32 and esp8266. All the other network interfaces are not documented explicitly anyways, right?

Regarding mpremote: note that this is just a configuration example. If this has `ipconfig` of `ifconfig` should not really matter, but I have no strong preference, which one is better.

docs/library/network.rst
```diff
@@ -113,8 +113,46 @@ parameter should be `id`.
...
+         * ``0x10`` The address is deprecated (but still valid)
+         * ``0x30`` The address is preferred (and valid)
+         * ``0x40`` The address is duplicated and can not be used.
+
+       This method can be used to set a static IPv6
+       address.
+
 .. method:: AbstractNIC.ifconfig([(ip, subnet, gateway, dns)])
 
+       Deprecated, only non-lwip ports.
```
projectgus, line 155: This is a bit ambiguous for MicroPython users to parse out. How about

```suggestion
       Deprecated, recommended to use ipconfig on ports which support it.
```

docs/library/network.WIZNET5K.rst
```diff
@@ -56,16 +56,6 @@ Methods
    Returns ``True`` if the physical Ethernet link is connected and up.
    Returns ``False`` otherwise.
 
-.. method:: WIZNET5K.ifconfig([(ip, subnet, gateway, dns)])
```
dpgeorge, line 59: This is being removed because it's documented in the AbstractNIC class, right?

Maybe there needs to be some note in this file that this class does extend the AbstractNIC type and so does have the ifconfig/ipconfig methods.

docs/library/network.rst
```diff
@@ -113,8 +113,46 @@ parameter should be `id`.
          connected to the AP.  The list contains tuples of the form
          (MAC, RSSI).
 
+.. method:: AbstractNIC.ipconfig('param')
+            AbstractNIC.ipconfig(param=value, ...)
+
+       Get or set interface-specific ip-configuration interface parameters.
```
dpgeorge, line 119: Maybe make "ip" uppercase?  So `IP-configuration`.

docs/library/network.rst
```diff
@@ -113,8 +113,46 @@ parameter should be `id`.
...
+       Supported parameters are (availability subject to compile-time flags):
```
dpgeorge, line 120: I suggest "(availability of a particular parameter depends on the port and the specific network interface)".  Because users see the port/NIC and not so much which compile time flags are enabled.

docs/library/network.rst
```diff
@@ -113,8 +113,46 @@ parameter should be `id`.
...
+         ``state`` indicates the current state of the address:
+
+         * ``0x08`` - ``0x0f`` indicates the address is tentative, counting the
+           number of probes sent.
+         * ``0x10`` The address is deprecated (but still valid)
+         * ``0x30`` The address is preferred (and valid)
+         * ``0x40`` The address is duplicated and can not be used.
+
+       This method can be used to set a static IPv6
+       address.
```
dpgeorge, line 151: I think this sentence needs to be indented so it comes under the above dot point.

docs/library/network.WIZNET5K.rst
felixdoerre: Yes, also there is also `.active`, and `.status` neither are directly documented here and only implicitly inherited. Interestingly the wiznet5k-class does not implement all functions from AbstractNIC (like `.scan`) but even before the change it implemented more than documented here.

To make it kind-of consistent we could remove the docs for `WIZNET5K.isconnected` and then add a paragraph linking to `AbstractNIC` for: isconnected, active, ifconfig (maybe leave that out to encourage ipconfig instead), ipconfig, config and status.

dpgeorge: > To make it kind-of consistent we could remove the docs for `WIZNET5K.isconnected` and then add a paragraph linking to `AbstractNIC` for: isconnected, active, ifconfig (maybe leave that out to encourage ipconfig instead), ipconfig, config and status.

Yes.  But that sounds like something for a separate PR.

felixdoerre:
I think I've incorporated all suggestions and was a bit more thorough to remove `ifconfig` from the docs as now all nics should have `ipconfig`.

docs/library/network.rst
```diff
@@ -113,8 +115,47 @@ parameter should be `id`.
...
+         ``state`` indicates the current state of the address:
+
+         * ``0x08`` - ``0x0f`` indicates the address is tentative, counting the
+           number of probes sent.
+         * ``0x10`` The address is deprecated (but still valid)
+         * ``0x30`` The address is preferred (and valid)
+         * ``0x40`` The address is duplicated and can not be used.
+
+         This method can be used to set a static IPv6
+         address.
```
dpgeorge, line 154: I think this needs more detail on how to set the IPv6 address.  Is the thing you set it to a 4 tuple in the same format as returned?

docs/library/network.rst
```diff
@@ -113,8 +115,47 @@ parameter should be `id`.
...
+
 .. method:: AbstractNIC.ifconfig([(ip, subnet, gateway, dns)])
 
+       Deprecated, recommended to use ipconfig on ports which support it.
```
dpgeorge, line 158: Please make this deprecation notice consistent with others by using:

    .. note:: This function is deprecated.  Please use :meth:`AbstractNIC.ipconfig` instead.

(See eg `machine.rst` and `sleep()`, and also `pyb.rst` has more examples.)

docs/library/network.rst
felixdoerre: ooh yes, I changed the other occurrence. Now I should have both.

docs/library/network.rst
felixdoerre: I have an example immediately after the property name, but I'll add it again here.

dpgeorge:

With this done we can now officially move forward with using `ipconfig()` over `ifconfig()`, and also say we support IPv6 (at least on ports that enable it).

Thanks for all your efforts on this @felixdoerre !

felixdoerre:
So, is it now time to enable ipv6 by default on some ports, e.g. pico-W?

dpgeorge:
> So, is it now time to enable ipv6 by default on some ports, e.g. pico-W?

Yes!  That would be a good idea.  Feel free to make a PR for that, or otherwise I'll eventually get to it.

razvanphp:
Seems like in here, `ifconfig` is not marked as deprecated and `ipconfig` is not even mentioned:
https://docs.micropython.org/en/latest/library/network.LAN.html#network.LAN.ifconfig

PR #14179, opened 2024-03-26
re: Add support for start- and endpos. 

greezybacon:
This adds support to match the CPython `re` module which supports `search`ing and `match`ing a substring using a start- and end-pos specified in the corresponding Pattern method.

Pattern objects have two additional parameters for the ::search and ::match methods to define the starting and ending position of the subject within the string to be searched.

This allows for searching a sub-string without creating a slice of the string, which is advantageous for performance and also makes inroads for something like the `finditer` method.

However, one caveat of using the start-pos rather than a slice is that the start anchor (`^`) remains anchored to the beginning of the text.

### DIFF
--- a/extmod/modre.c
+++ b/extmod/modre.c
@@ -196,10 +196,11 @@ static void re_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t 
 
 // Note: this function can't be named re_exec because it may clash with system headers, eg on FreeBSD
 static mp_obj_t re_exec_helper(bool is_anchored, uint n_args, const mp_obj_t *args) {
-    (void)n_args;
     mp_obj_re_t *self;
+    bool was_compiled = false;
     if (mp_obj_is_type(args[0], (mp_obj_type_t *)&re_type)) {
         self = MP_OBJ_TO_PTR(args[0]);
+        was_compiled = true;
     } else {
         self = MP_OBJ_TO_PTR(mod_re_compile(1, args));
     }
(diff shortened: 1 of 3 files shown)


stinos:
> This adds support to match the CPython re module which supports searching and matching a substring using a start- and end-pos specified in the corresponding Pattern method

Just to be clear: 'match' doesn't mean it's CPython-compatible (and hence the use of a .exp file for the tests)?

greezybacon:
Oh- it was my intention that it would be CPython compatible. I took the
tests from their documentation.

I didn’t realize that EXP files were optional for tests. I’ll take it out.
Thanks.

greezybacon:
I was a bit excited and went ahead and added a `finditer` implementation as well. It's also aimed at compatibility with CPython's `re` module. If it should remain, I'd like a little direction on the `#define` to include support for it. Do we need it? Did I get it in the right place?

extmod/modre.c
```diff
@@ -195,17 +195,39 @@ static void re_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t
...
+        subj.begin += startpos;
+        if (n_args > 3) {
+            // Arg #3 is ending-pos
+            size_t endpos = mp_obj_get_int(args[3]);
+            if (endpos > len) {
+                endpos = len;
+            }
+            else if (endpos < startpos) {
+                endpos = startpos;
+            }
```
dpgeorge, line 226: Please add a test case that tests this line.

greezybacon: Thanks, @dpgeorge, I did.

dpgeorge:
Sorry, this PR got a bit lost...

> I was a bit excited and went ahead and added a `finditer` implementation as well.

OK, so that increases the scope somewhat of this PR :sweat_smile: 

It would be good to know what the code size impact is for just the start/endpos change.  To that end, probably best to keep this PR just for start/endpos (including tests for that), and have a separate PR for finditer.  That makes it easier to review.

greezybacon:
Thanks, @dpgeorge. I really appreciate the review and feedback. I have separated the `finditer` method into a separate branch and will propose it as a separate PR if this gets merged.

dpgeorge:
Thanks for updating this PR.

It looks fine to me, it's always good to (1) extend compatibility with CPython, and (2) provide a way to do something without allocating on the heap.

It's also good that you've made it so that the arguments are not accepted when using the module level `re.search` and `re.match` functions (ie they are only accepted on precompiled re types, as per CPython).

Considering the entire `re` module is behind a configuration option, I would say this addition is acceptable in terms of code size increase.

dpgeorge (approved):

PR #14180, opened 2024-03-26
py/makeqstrdata.py: Ensure that scope names get low qstr values.

### LINKED ISSUE #14093: Unsigned conversion overflow for QSTRs during build
### Checks

Applies to all ports and boards

### MicroPython version

Micropython v1.22

### Reproduction

* Add a new ext_mod with lots of QSTRs

### Expected behaviour

Build succeeds

### Observed behaviour

Build creates the following error messages

    micropython-v1.22.2/py/scope.c:36:22: warning: unsigned conversion from 'int' to 'unsigned char' changes value from '256' to '0' [-Woverflow]
    36 |     [SCOPE_LAMBDA] = MP_QSTR__lt_lambda_gt_,
    micropython-v1.22.2/py/scope.c:46:5: note: in expansion of macro 'MP_STATIC_ASSERT'
    46 |     MP_STATIC_ASSERT(MP_QSTR__lt_lambda_gt_ <= UINT8_MAX);
      |     ^~~~~~~~~~~~~~~~

and a few similar errors.

### Additional Information

In `py/scope.c`, a comment says that "these low numbered qstrs should fit in 8 bits.  See assertions below." However, during the QSTR generation, there is nothing to force this.

It seems that there is a way to do this in `py/makeqstrdefs.py`: there is a list of operators `operator_qstr_list` that must have low values.

My solution was to add these values to this list.

    "<module>",
    "<lambda>",
    "<listcomp>",
    "<dictcomp>",
    "<setcomp>",
    "<genexpr>",

However, the name of the list and the comment above it are now incorrect, as they refer to 'operators'. 

Please indicate whether it would be better to create _another_ list with additional QSTRs to keep below 255, and add another list to the test around line 410 in `print_qstr_data`, changing

    pool = 0 if qstr in operator_qstr_list else 1

to something like

    pool = 0 is qstr in operator_qstr_list or qstr in more_low_qstr_list else 1

jimmo:
Originally implemented in a patch file provided by @ironss-iotec.

Fixes #14093

### DIFF
--- a/py/makeqstrdata.py
+++ b/py/makeqstrdata.py
@@ -223,11 +223,11 @@ static_qstr_list = [
     "zip",
 ]
 
-# Additional QSTRs that must have index <255 because they are stored in
-# `mp_binary_op_method_name` and `mp_unary_op_method_name` (see py/objtype.c).
+# Additional QSTRs that must have index <255 because they are stored as `byte` values.
 # These are not part of the .mpy compatibility list, but we place them in the
 # fixed unsorted pool (i.e. QDEF0) to ensure their indices are small.
-operator_qstr_list = {
+unsorted_qstr_list = {
+    # From py/objtype.c: used in the `mp_binary_op_method_name` and `mp_unary_op_method_name` tables.
     "__bool__",
     "__pos__",
     "__neg__",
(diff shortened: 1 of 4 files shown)


py/makeqstrdata.py
```diff
@@ -286,6 +286,13 @@
     "__get__",
     "__set__",
     "__delete__",
+    # From py/scope.c: used in `scope_simple_name_table` table.
+    "<module>",
```
dpgeorge, line 290: I'm not sure this should be added, because it's already in `static_qstr_list`.  And so it may be possible for there to be a duplicate definition of it as QDEF0?

Also, operators from `py/objtype.c` that are in `static_qstr_list` are not listed in this set, so for consistency `<module>` shouldn't be.

jimmo: Updated.

PR #14181, opened 2024-03-26
extmod/modopenamp: Implement streaming protocol for endpoints.

iabdalkader:
This patch implements a streaming protocol for Endpoints, to allow them to be used with select/poll or asyncio. It also implements `Endpoint.recv()` function, whose behavior matches that of UDP sockets' recv(). Calling `Endpoint.recv()` returns a single message, truncated to fit the requested number of bytes. If two messages are pending on the endpoint, the code must do two separate reads to get them both. Received RPMsg messages are saved by holding the buffers in the endpoint receive callback, and releasing them when they are no longer needed (i.e., when `Endpoint.recv` is called). This avoids the copying buffers twice, saves memory, but also allows the host to block the remote if it's not receiving messages (assuming the queue size is the same as the number of buffers in the VRING). Note that the queuing of RPMsg messages is only enabled if `callback=None` (i.e., if the application is not processing messages asynchronously).

Alternative Python implementation:
```Python
class EndpointIO(io.IOBase):
    def __init__(self, ept):
        self.ept = ept
        
    def write(self, buf):
        if buf != b"\r\n":
            self.ept.send(buf, timeout=0)
        
    def ioctl(self, op, arg):
        if op == _MP_STREAM_POLL and self.ept.is_ready():
            return _MP_STREAM_POLL_WR
        return 0
```

### DIFF
--- a/extmod/modopenamp.c
+++ b/extmod/modopenamp.c
@@ -33,6 +33,8 @@
 #include "py/nlr.h"
 #include "py/runtime.h"
 #include "py/mpprint.h"
+#include "py/stream.h"
+#include "py/mperrno.h"
 
 #include "metal/sys.h"
 #include "metal/alloc.h"
@@ -44,6 +46,7 @@
 #include "openamp/open_amp.h"
 #include "openamp/remoteproc.h"
 #include "openamp/remoteproc_loader.h"
+#include "lib/rpmsg/rpmsg_internal.h"
 #include "modopenamp.h"
 
 #if !MICROPY_ENABLE_FINALISER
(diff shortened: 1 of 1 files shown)


dpgeorge (commented):
Are you actively using this patch?  Has your implementation of it changed since you made this PR?

It's nice to see endpoints behaving as streams.  But it's kind of strange that while it's a stream, it doesn't have read/write methods like other streams do.  That means that C code can read/write the endpoint like it's a stream, but Python code cannot.

iabdalkader:
> Are you actively using this patch? Has your implementation of it changed since you made this PR?

No, and I'm not sure anymore it's useful. I managed to make an Endpoint stream like so:

```Python
class EndpointIO(io.IOBase):
    def __init__(self, ept):
        self.ept = ept
        
    def write(self, buf):
        if buf != b"\r\n":
            self.ept.send(buf, timeout=0)
        
    def ioctl(self, op, arg):
        if op == _MP_STREAM_POLL and self.ept.is_ready():
            return _MP_STREAM_POLL_WR
        return 0
```

So I think this PR's only value is having streaming supported out of the box. If that's something desirable, I can address issues and update it.

iabdalkader:
Closing this because the Python implementation works fine and keeps the C code minimal. A future improvement would be to make the module extensible and add streaming support in an extension.

PR #10263, opened 2022-12-18
github: Move windows CI builds to GitHub Actions.

dlech:
By moving to GitHub actions, all MicroPython CI builds are now on GitHub actions. This allows faster parallel builds and saves time by not building when no relevant files changed.

This reveals a few failing tests, so those are temporarily disabled until they can be fixed.

The `misc_sys_settrace_features.py` is disable on visual studio build because it is failing with:

```
--- "D:/a/micropython/micropython/tests/results\\misc_sys_settrace_features.py.exp"	2022-12-17 23:24:57.940564100 +0000
+++ "D:/a/micropython/micropython/tests/results\\misc_sys_settrace_features.py.out"	2022-12-17 23:24:57.940564100 +0000
@@ -40,799 +40,17 @@
  1:   @__main__:factorial => sys_settrace_features.py:91
  2:   @__main__:factorial => sys_settrace_features.py:91
  3:   @__main__:do_tests => sys_settrace_features.py:97
- 4:   @__main__:<module> => sys_settrace_features.py:106
-### trace_handler::main event: line
- 0:   @__main__:factorial => sys_settrace_features.py:88
- 1:   @__main__:factorial => sys_settrace_features.py:91
- 2:   @__main__:factorial => sys_settrace_features.py:91
- 3:   @__main__:do_tests => sys_settrace_features.py:97
- 4:   @__main__:<module> => sys_settrace_features.py:106
[snipped repeating]

------------------- script exited ------------------
-Total traces executed:  155
+Traceback (most recent call last):
+  File "misc/sys_settrace_features.py", line 106, in <module>
+  File "misc/sys_settrace_features.py", line 97, in do_tests
+  File "misc/sys_settrace_features.py", line 91, in factorial
+  File "misc/sys_settrace_features.py", line 91, in factorial
+  File "misc/sys_settrace_features.py", line 69, in trace_tick_handler
+  File "misc/sys_settrace_features.py", line 38, in trace_tick
+  File "misc/sys_settrace_features.py", line 30, in print_stacktrace
+  File "misc/sys_settrace_features.py", line 30, in print_stacktrace
+  File "misc/sys_settrace_features.py", line 30, in print_stacktrace
+  File "misc/sys_settrace_features.py", line 30, in print_stacktrace
+  File "misc/sys_settrace_features.py", line 25, in print_stacktrace
+RuntimeError: maximum recursion depth exceeded
+CRASH
\ No newline at end of file

[...]

PR #14188, opened 2024-03-26
Improve MSYS2-based CI builds.

stinos:
This builds on #10263, fixing a mistake I once made when writing the MSYS2 installation instructions: the correct  (well, for what we need) package to install for Python on MSYS2 is `<mingwprefix>-python3` not plain `python3`. This way one single Python version can be used both for running run-tests.py and getting the expected test output.

Note these days MSYS2 also has an 'ucrt64' [environment ](https://www.msys2.org/docs/environments/) which apparently has a bunch of floating point related fixes we've sort of long been waiting for, but it's unclear to me if `mingw64` is going to be deprecated so I only added a comment about that.

@dlech FYI

Now, I only figured this out after RDP-ing into the action runner because I didn't notice at first that the way I setup my local MSYS2 environment used different instructions from the one in windows/README.md. However the process to RDP into a github action runner mentioned in #10263 does not work anymore since that repository has been disabled due to a violation of GitHub's terms of service. I skimmed over those terms but really have no idea which terms exactly.. @dpgeorge these instructions can be pretty useful imo, and the basic principle is simple enough and there are multiple solutions to implement it, but I have no idea if I can post them here and/or include them in the docs somewhere - what do you think?

### DIFF
--- a/.github/workflows/ports_windows.yml
+++ b/.github/workflows/ports_windows.yml
@@ -42,6 +42,8 @@ jobs:
           configuration: Debug
         - visualstudio: '2019'
           configuration: Debug
+    env:
+      CI_BUILD_CONFIGURATION: ${{ matrix.configuration }}
     runs-on: ${{ matrix.runner }}
     steps:
     - name: Install Visual Studio 2017
--- a/.github/workflows/ports_windows.yml
+++ b/.github/workflows/ports_windows.yml
@@ -108,16 +108,6 @@ jobs:
       run:
         shell: msys2 {0}
     steps:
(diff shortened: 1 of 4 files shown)


dpgeorge:
Thanks for the patch, this is a nice clean up / improvement.

> @dpgeorge these instructions can be pretty useful imo, and the basic principle is simple enough and there are multiple solutions to implement it, but I have no idea if I can post them here and/or include them in the docs somewhere - what do you think?

I agree the RDP instructions are useful.

Maybe we can add them in the wiki here on github?

PR #14194, opened 2024-03-27
tinyusb: Don't disconnect on soft reset unless runtime USB was active.

projectgus:
Previously, constructing the singleton USBDevice object was enough to trigger a USB disconnect on soft reset. Now it also has to be active.

The only case where this changes the behaviour is if the USBDevice object has been constructed but never set to active (no more disconnect in this case). Otherwise, behaviour is the same.

This change was [requested by hippy on the raspberrypi forums](https://forums.raspberrypi.com/viewtopic.php?p=2205682#p2205682).

This PR also includes a second commit that updates some comments related to runtime USB, but doesn't change any code.

### DIFF
--- a/shared/tinyusb/mp_usbd_runtime.c
+++ b/shared/tinyusb/mp_usbd_runtime.c
@@ -431,11 +431,12 @@ void mp_usbd_init(void) {
 // Top-level USB device deinit.
 //
 // This variant is called from soft reset, NULLs out the USB device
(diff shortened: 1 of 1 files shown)


projectgus:
@dpgeorge Would be nice to get this small patch in 1.23 if possible :grin:

projectgus:
@dpgeorge That's right. That should be the only case where the behaviour changes.

PR #7722, opened 2021-08-27
gc.c: Ensure a gap of one byte before the finaliser table.

jepler:
.. or, for !MICROPY_ENABLE_FINALISER, before the first block of the pool.
    
Closes: adafruit/circuitpython#5021
Closes: micropython/micropython#7716

I tested this with a version of https://github.com/micropython/micropython/issues/7716#issuecomment-906872738 and there were no problems detected. ~~I haven't tested it on our real HW yet.~~ it also fixes the original problem on our real hw

PR #14196, opened 2024-03-27
gc: Support valgrind tracking of Python heap allocations

projectgus:
Extends [Valgrind](https://valgrind.org/) memcheck support in the unix port to track the Python heap, allowing Valgrind to report use-after-free and (in some cases) out of bounds reads and writes when accessing Python heap memory.

Valgrind support is implemented in a dedicated header, to try and keep gc.c relatively clean.

Also adds an additional valgrind build configuration: `MICROPY_DEBUG_VALGRIND_MAX_ADDRSPACE`. When set, the heap will never reuse the address ranges of freed memory - new allocations are always at a higher address than any previous allocation. Assuming there's enough total heap, this means that use after free errors can never be masked by the freed address being already reallocated for something else.

This branch was used when reproducing and verifying #14029, and seems like it will be useful for other purposes as well.

## TODO

* [ ] Add support for valgrind in run-tests.py.
* [x] Check the existing valgrind logic in `gc_get_ptr()` that was added in #7722 (84071590b3d76a647fe9cd9055a089605fedc219). I need to verify but this logic appears to prevent the GC marking any "pointer" regions if valgrind is enabled. The side effect seems to be that objects retained on the stack (like the running bytecode!), and root pointers, are being garbage collected prematurely. I think we can implement this differently to still avoid warnings if the GC reads uninitialised stack words, but without incorrect GC behaviour. EDIT: Done, have implemented this in a different way so normal gc scans can continue.

## Limitations

* Valgrind supports defining a "redzone" between each allocation so out of bounds read or write is always caught. This branch does not have this, so if an allocation is a block-aligned length and the next sequential block in the heap is also allocated then Valgrind can't detect an out of bounds read or write that spills over into the next block. I couldn't find an unobtrusive way to patch gc.c to add a redzone, but someone might have more luck than me...


projectgus:
Thanks very interesting, thanks @stinos! I hadn't seen MTuner and I agree it could be a very useful tool for some MicroPython analysis.

> how about making memory profiling in gc.c generic, simply by name, i.e. keep everything as it is now but instead of VALGRIND_MP_MALLOC call it MP_MEMPROFILER_ALLOC etc, and then at the top of the file use

I would quibble with calling valgrind's memcheck a profiler, it's more of a runtime correctness checker (i.e. closer to tools like ASan than tools like MTuner). This is kind of a philosophical point, but kind of not because there's a bunch of places in this PR where we have to tell valgrind "we're about to access some memory that's supposed to be inaccessible now, but it's not an error" (for example, see around gc.c:868). This concept doesn't really have an equivalent when doing memory profiling.

That said, I do think it's a good idea to make the macros generic if possible. I will take a look and see how well they can be made to fit. I really like the idea of making it easier to integrate multiple different tools with the MicroPython core, while keeping the main source files pretty clean.

projectgus:
> there's a bunch of places in this PR where we have to tell valgrind "we're about to access some memory that's supposed to be inaccessible now, but it's not an error" (for example, see around gc.c:868)

Actually maybe this is less complex than I thought, can probably collapse down to something like:

* `MP_GC_TRACE_INIT()`
* `MP_GC_TRACE_PRE_ALLOC()`
* `MP_GC_TRACE_ALLOC()`
* `MP_GC_TRACE_REALLOC()`
* `MP_GC_TRACE_FREE()`

Will try this soon, thanks!

stinos:
> I would quibble with calling valgrind's memcheck a profiler, it's more of a runtime correctness checker

Sure; I didn't give the naming a lot of thought but indeed `GC_TRACE` does cover the various checkers/viewers/... in a better way.

PR #13340, opened 2024-01-03
nrf: Implement time.time() and machine.RTC

cwalther:
Optionally adds `time()` and `time_ns()` to the `time` module, as well as a `machine.RTC` class that only implements the `datetime()` method (following the example of the _rp2_ port), whose sole purpose is to provide the ability to set the time. Also provides the basis for enabling `gmtime()`, `localtime()`, `mktime()` in the time module.

The nRF52 does not have a dedicated real-time clock peripheral, but timekeeping can be done by the same real-time counter that already powers the `time.ticks_*` and related functions. For reasonable accuracy, a suitable LFCLK source is required: The internal RC oscillator (`BLUETOOTH_LFCLK_RC`) by itself is insufficient, but any of the following work fine:

- external 32kHz crystal (default)
- synthesis from HFCLK (`BLUETOOTH_LFCLK_SYNTH`) when HFXO (external 32MHz crystal) is enabled
- `BLUETOOTH_LFCLK_RC` + periodic calibration from HFXO (automatically done by the SoftDevice while enabled using `ble.enable()`)

Boards can enable this by defining both configuration options `MICROPY_PY_TIME_TICKS` and `MICROPY_PY_TIME_TIME_TIME_NS`. Additionally, they may want to enable `MICROPY_PY_TIME_GMTIME_LOCALTIME_MKTIME`. I have not enabled it on any of the existing boards, because my main use case is an [external](https://github.com/micropython/micropython/pull/13071) one (and I don’t have any of them to test), but can do that if desired.

This includes a generic implementation of `mp_time_localtime_get()` and `mp_time_time_get()` in terms of `mp_hal_time_ns()`, which could also be used by other ports. In particular by the _embed_ port, for which I originally wrote it in #11430 (https://github.com/micropython/micropython/pull/11430/commits/82b8679e91e90f65b9c5c30fc7cfbeb0eed28d3b), noting the following:

> I'm unsure where to put _modtime_mphal.h_, it could also be in _extmod_. The important thing is that for `MICROPY_PY_TIME_INCLUDEFILE` to work it must be at the same path in both the port build (original source tree) and the application build (_micropython_embed_ distribution), therefore not in _ports/embed/port_.
> 
> It is named _.h_, mismatching the corresponding _ports/*/modtime.c_, because it must not be compiled separately, which naming it _.c_ would make harder for users of the embed port - they would need to explicitly exclude it, whereas this way they can continue to just compile all the _.c_ files found in the _micropython_embed_ distribution except those in _lib_.

[...]

PR #10400, opened 2023-01-03
ports/nrf: Add RTC support to NRF ports.

RetiredWizard:
Updated after reviewing RTCounter.py. It turns out that RTCounter starts up and uses the second nrf rtc unit so piggybacking off of that this update doesn't need most of the Circuitpython logic the earlier PR was using and takes care of starting up the rtc which had been omitted.

Since RTCounter uses rtc unit 1 this PR compiles on all of the currently supported nrf series boards.

Signed-off-by: RetiredWizard <github@retiredwizard.com>

PR #14198, opened 2024-03-27
ports/nrf: Add RTC support to NRF ports.

RetiredWizard:
Resubmitting after accidentally closing #10400

Updated after reviewing RTCounter.py. It turns out that RTCounter starts up and uses the second nrf rtc unit so piggybacking off of that this update doesn't need most of the Circuitpython logic the earlier PR was using and takes care of starting up the rtc which had been omitted.

Since RTCounter uses rtc unit 1 this PR compiles on all of the currently supported nrf series boards.

This is the same update as the accidentally closed PR but there is an alternate version to evaluate as well #13340.

Signed-off-by: RetiredWizard [github@retiredwizard.com](mailto:github@retiredwizard.com)


ports/nrf/mphalport.c
```diff
@@ -129,7 +129,10 @@ static void rtc_irq_time(nrfx_rtc_int_type_t event) {
 
 void rtc1_init_time_ticks(void) {
     // Start the low-frequency clock (if it hasn't been started already)
-    mp_nrf_start_lfclk();
+    if (!nrf_clock_lf_is_running(NRF_CLOCK)) {
+        nrf_clock_task_trigger(NRF_CLOCK, NRF_CLOCK_TASK_LFCLKSTART);
+        rtc_offset_check();
+    }
```
cwalther, line 135: Please don’t undo my change from b10182b, it was done for a reason.

RetiredWizard: That was probably a rebasing error, I'll put it back. Thanks!

RetiredWizard:
@cwalther I was thinking about the offset check from my PR and was wondering if it's a carry over from the IMX boards that support an RTC battery to keep the clock running when the reset of the board is shutdown. I don't know if the nrf boards have any similar capability.

cwalther:
Found it. The validation thing is from https://github.com/adafruit/circuitpython/pull/2924/ and the purpose is indeed to maintain the clock over certain kinds of reset that clear the BSS section but not all of RAM (not a MicroPython soft reboot, that is already achieved by my approach). However, the parts of it that are in this PR are not sufficient to make this work, adjusting the linker scripts and heap initialization for setting up the `.uninitialized` section is also needed, as well as saving the RTC counter (https://github.com/adafruit/circuitpython/commit/daf7c2857da3f0cc61f20474bdd1ce2824cb409c).

I am undecided whether this is a desirable thing, so I will not add it to #13340 for now. It is useful in some cases, such as an unattended device in the field that occasionally crashes and is restarted by a watch-dog timer. On the other hand, it is a complication and violates the expectation that a reset resets everything.

RetiredWizard:
Thanks! Great detective work!

jkorte-dev:
I had to add MICROPY_PY_MACHINE_RTC_ENTRY to the MICROPY_PY_MACHINE_EXTRA_GLOBALS marco to make RTC show up in the machine module 

time.localtime() and rtc.datetime() are not in sync. 
According to https://docs.micropython.org/en/latest/library/time.html
"If secs is not provided or None, then the current time from the RTC is used."

time.localtime(time.mktime(machine.RTC().datetime())) works 

Thank you for your work!

PR #14199, opened 2024-03-27
Add nic.ipconfig for other ports (like esp32, stm32 and others)

felixdoerre:
In order to make network configuration extensible and consistent across ports, we implemented `ipconfig` for the other implemented nics:

   - [x] extmod/network_ninaw10.c -> nina_ifconfig
   - [x] ports/cc3200/mods/modwlan.c -> sl_NetCfgSet
   - [x] ports/esp32/network_lan.c --> esp-netif
   - [x] ports/esp32/network_ppp.c --> (old?) lwip, can only be used to set dns.
   - [x] ports/esp32/network_wlan.c --> esp-netif (setting the IP in CIDR notation requries more work)
   - [x] ports/esp8266/network_wlan.c -> esp-netif, copy of the esp32 code(?)
   - [x] ports/mimxrt/network_lan.c -> lwip
   - [x] ports/stm32/network_lan.c -> lwip

### DIFF
--- a/extmod/network_ninaw10.c
+++ b/extmod/network_ninaw10.c
@@ -41,6 +41,7 @@
 #include "py/runtime.h"
 #include "py/misc.h"
 #include "py/mperrno.h"
+#include "py/parsenum.h"
 #include "shared/netutils/netutils.h"
 #include "shared/runtime/softtimer.h"
 #include "extmod/modnetwork.h"
@@ -75,6 +76,8 @@ typedef struct _nina_obj_t {
 #define SO_NO_CHECK             (0x100a)
 #define NINAW10_POLL_INTERVAL   (100)
 
+#define IPADDR_STRLEN_MAX       46
+
 #define is_nonblocking_error(errno) ((errno) == MP_EAGAIN || (errno) == MP_EWOULDBLOCK || (errno) == MP_EINPROGRESS)
 
 #define debug_printf(...) // mp_printf(&mp_plat_print, __VA_ARGS__)
@@ -86,6 +89,8 @@ static nina_obj_t network_nina_wl_ap = {{(mp_obj_type_t *)&mod_network_nic_type_
 static mp_sched_node_t mp_wifi_poll_node;
 static soft_timer_entry_t mp_wifi_poll_timer;
 static void network_ninaw10_deinit(void);
+
+static bool network_ninaw10_dhcp_active = false;
 
 static bool network_ninaw10_poll_list_is_empty(void) {
     return MP_STATE_PORT(mp_wifi_poll_list) == NULL ||
@@ -199,6 +204,7 @@ static mp_obj_t network_ninaw10_active(size_t n_args, const mp_obj_t *args) {
                 mp_raise_msg_varg(&mp_type_OSError,
                     MP_ERROR_TEXT("failed to initialize Nina-W10 module, error: %d"), error);
             }
+            network_ninaw10_dhcp_active = true;
             // check firmware version
             uint8_t semver[NINA_FW_VER_LEN];
             if (nina_fw_version(semver) != 0) {
@@ -367,10 +373,154 @@ static mp_obj_t network_ninaw10_ifconfig(size_t n_args, const mp_obj_t *args) {
         netutils_parse_ipv4_addr(items[2], ifconfig.gateway_addr, NETUTILS_BIG);
         netutils_parse_ipv4_addr(items[3], ifconfig.dns_addr, NETUTILS_BIG);
         nina_ifconfig(&ifconfig, true);
+        network_ninaw10_dhcp_active = false;
         return mp_const_none;
     }
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(network_ninaw10_ifconfig_obj, 1, 2, network_ninaw10_ifconfig);
+
+mp_obj_t network_ninaw10_ipconfig(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs) {
+    nina_ifconfig_t ifconfig;
+    // get ifconfig info
+    nina_ifconfig(&ifconfig, false);
+
+    if (kwargs->used == 0) {
+        // Get config value
+        if (n_args != 1) {
+            mp_raise_TypeError(MP_ERROR_TEXT("must query one param"));
+        }
+        switch (mp_obj_str_get_qstr(args[0])) {
+            case MP_QSTR_dns: {
+                return netutils_format_ipv4_addr(ifconfig.dns_addr, NETUTILS_BIG);
+            }
+            default: {
+                mp_raise_ValueError(MP_ERROR_TEXT("unexpected key"));
+                break;
+            }
+        }
+    } else {
+        // Set config value(s)
+        if (n_args != 0) {
+            mp_raise_TypeError(MP_ERROR_TEXT("can't specify pos and kw args"));
+        }
+
+        for (size_t i = 0; i < kwargs->alloc; ++i) {
+            if (MP_MAP_SLOT_IS_FILLED(kwargs, i)) {
+                mp_map_elem_t *e = &kwargs->table[i];
+                switch (mp_obj_str_get_qstr(e->key)) {
+                    case MP_QSTR_dns: {
+                        netutils_parse_ipv4_addr(e->value, ifconfig.dns_addr, NETUTILS_BIG);
+                        nina_ifconfig(&ifconfig, true);
+                        break;
+                    }
+                    default: {
+                        mp_raise_ValueError(MP_ERROR_TEXT("unexpected key"));
+                        break;
+                    }
+                }
+            }
+        }
+    }
+    return mp_const_none;
+}
+
+static mp_obj_t network_ninaw10_nic_ipconfig(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs) {
+    nina_ifconfig_t ifconfig;
+    // get ifconfig info
+    nina_ifconfig(&ifconfig, false);
+
+    if (kwargs->used == 0) {
+        // Get config value
+        if (n_args != 2) {
+            mp_raise_TypeError(MP_ERROR_TEXT("must query one param"));
+        }
+
+        switch (mp_obj_str_get_qstr(args[1])) {
+            case MP_QSTR_dhcp4: {
+                return mp_obj_new_bool(network_ninaw10_dhcp_active);
+            }
+            case MP_QSTR_has_dhcp4: {
+                uint16_t ip_sum =
+                    ifconfig.ip_addr[0] + ifconfig.ip_addr[1] + ifconfig.ip_addr[2] + ifconfig.ip_addr[3];
+                if (network_ninaw10_dhcp_active) {
+                    return mp_obj_new_bool(ip_sum != 0);
+                } else {
+                    return mp_const_false;
+                }
+            }
+            case MP_QSTR_addr4: {
+                mp_obj_t tuple[2] = {
+                    netutils_format_ipv4_addr(ifconfig.ip_addr, NETUTILS_BIG),
+                    netutils_format_ipv4_addr(ifconfig.subnet_addr, NETUTILS_BIG),
+                };
+                return mp_obj_new_tuple(2, tuple);
+            }
+            case MP_QSTR_gw4: {
+                return netutils_format_ipv4_addr(ifconfig.gateway_addr, NETUTILS_BIG);
+            }
+            default: {
+                mp_raise_ValueError(MP_ERROR_TEXT("unexpected key"));
+                break;
+            }
+        }
+        return mp_const_none;
+    } else {
+        // Set config value(s)
+        if (n_args != 1) {
+            mp_raise_TypeError(MP_ERROR_TEXT("can't specify pos and kw args"));
+        }
+
+        for (size_t i = 0; i < kwargs->alloc; ++i) {
+            if (MP_MAP_SLOT_IS_FILLED(kwargs, i)) {
+                mp_map_elem_t *e = &kwargs->table[i];
+                switch (mp_obj_str_get_qstr(e->key)) {
+                    case MP_QSTR_dhcp4: {
+                        mp_raise_ValueError(MP_ERROR_TEXT("DHCP control unsupported"));
+                        break;
+                    }
+                    case MP_QSTR_addr4: {
+                        int prefix_bits = 32;
+                        if (e->value != mp_const_none && mp_obj_is_str(e->value)) {
+                            size_t addr_len;
+                            const char *input_str = mp_obj_str_get_data(e->value, &addr_len);
+                            char *split = strchr(input_str, '/');
+                            if (split) {
+                                mp_obj_t prefix_obj = mp_parse_num_integer(split + 1, strlen(split + 1), 10, NULL);
+                                prefix_bits = mp_obj_get_int(prefix_obj);
+                                uint32_t mask = -(1u << (32 - prefix_bits));
+                                ifconfig.subnet_addr[0] = (mask >> 24) & 0xFF;
+                                ifconfig.subnet_addr[1] = (mask >> 16) & 0xFF;
+                                ifconfig.subnet_addr[2] = (mask >> 8) & 0xFF;
+                                ifconfig.subnet_addr[3] = mask & 0xFF;
+                            }
+                            netutils_parse_ipv4_addr(e->value, ifconfig.ip_addr, NETUTILS_BIG);
+                        } else if (e->value != mp_const_none) {
+                            mp_obj_t *items;
+                            mp_obj_get_array_fixed_n(e->value, 2, &items);
+                            netutils_parse_ipv4_addr(items[0], ifconfig.ip_addr, NETUTILS_BIG);
+                            netutils_parse_ipv4_addr(items[1], ifconfig.subnet_addr, NETUTILS_BIG);
+                        }
+                        nina_ifconfig(&ifconfig, true);
+                        network_ninaw10_dhcp_active = false;
+                        break;
+                    }
+                    case MP_QSTR_gw4: {
+                        netutils_parse_ipv4_addr(e->value, ifconfig.gateway_addr, NETUTILS_BIG);
+                        nina_ifconfig(&ifconfig, true);
+                        network_ninaw10_dhcp_active = false;
+                        break;
+                    }
+                    default: {
+                        mp_raise_ValueError(MP_ERROR_TEXT("unexpected key"));
+                        break;
+                    }
+                }
+            }
+        }
+    }
+    return mp_const_none;
+}
+static MP_DEFINE_CONST_FUN_OBJ_KW(network_ninaw10_nic_ipconfig_obj, 1, network_ninaw10_nic_ipconfig);
 
 static mp_obj_t network_ninaw10_config(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs) {
     nina_obj_t *self = MP_OBJ_TO_PTR(args[0]);
@@ -856,6 +1006,7 @@ static const mp_rom_map_elem_t nina_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_disconnect),          MP_ROM_PTR(&network_ninaw10_disconnect_obj) },
     { MP_ROM_QSTR(MP_QSTR_isconnected),         MP_ROM_PTR(&network_ninaw10_isconnected_obj) },
     { MP_ROM_QSTR(MP_QSTR_ifconfig),            MP_ROM_PTR(&network_ninaw10_ifconfig_obj) },
+    { MP_ROM_QSTR(MP_QSTR_ipconfig),            MP_ROM_PTR(&network_ninaw10_nic_ipconfig_obj) },
     { MP_ROM_QSTR(MP_QSTR_config),              MP_ROM_PTR(&network_ninaw10_config_obj) },
     { MP_ROM_QSTR(MP_QSTR_status),              MP_ROM_PTR(&network_ninaw10_status_obj) },
     { MP_ROM_QSTR(MP_QSTR_ioctl),               MP_ROM_PTR(&network_ninaw10_ioctl_obj) },
--- a/ports/esp32/network_common.c
+++ b/ports/esp32/network_common.c
@@ -34,6 +34,7 @@
 #include <string.h>
 
 #include "py/runtime.h"
+#include "py/parsenum.h"
 #include "py/mperrno.h"
 #include "shared/netutils/netutils.h"
 #include "modnetwork.h"
@@ -42,7 +43,7 @@
 #include "esp_netif.h"
 #include "esp_wifi.h"
 #include "lwip/sockets.h"
-// #include "lwip/dns.h"
+#include "lwip/dns.h"
 
 NORETURN void esp_exceptions_helper(esp_err_t e) {
     switch (e) {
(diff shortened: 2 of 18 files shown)


robert-hh:
Tested your changes for MIMXRT (MIMXRT1020_EVK) and STM32 (LAN: PYBD_SF6 with DP83848, WiFi: Built-in CYW43) and they work, as expected. I did not test other boards because the changes do not touch board-specific features.

robert-hh:
Tested a little bit more. 
ESP32:
The ESP32 port builds and I can query parameters. 
- Setting e.g. addr4 failed. 
```
>>> lan.ipconfig(addr4="10.0.0.144/16")
TypeError: can't specify pos and kw args
```
and
```
>>> lan.ipconfig(addr4=("10.0.0.144", "255.255.255.0"))
TypeError: can't specify pos and kw args
```
- network.ipconfig() is missing.

MIMXRT & STM32:
- Setting addr4 using CIDR notation failed:
```
>>> lan.ipconfig(addr4="10.0.0.216/24")
ValueError: invalid arguments
```

ports/esp32/network_common.c
```diff
@@ -154,6 +155,148 @@ static mp_obj_t esp_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+            }
+            default: {
+                mp_raise_ValueError(MP_ERROR_TEXT("unexpected key"));
+                break;
+            }
+        }
+        return mp_const_none;
+    } else {
+        // Set config value(s)
+        if (n_args != 0) {
```
robert-hh, line 200: should be: `n_args != 1` or `n_args > 1`

robert-hh:
I have updated my branch with the implementation for NINAW10. It includes code for network.ipconfig(), even if the effect with NINAW10 is unclear, which may allow to set the DNS, but it cannot be read back from the module.

ports/esp32/network_common.c
```diff
@@ -154,6 +155,148 @@ static mp_obj_t esp_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+                                mp_obj_t prefix_obj = mp_parse_num_integer(split + 1, strlen(split + 1), 10, NULL);
+                                prefix_bits = mp_obj_get_int(prefix_obj);
+                            }
+                            uint32_t mask = -(1u << (32 - prefix_bits));
+                            uint32_t *m = (uint32_t *)&info.netmask;
+                            *m = esp_netif_htonl(mask);
+                            // TODO: netutils code does not allow inputing a C-string, but requires a mp_string
+                            // This makes parsing CIDR-Notation a bit more complex.
+                            // netutils_parse_ipv4_addr(addr_str, (void *)&info.ip, NETUTILS_BIG);
+                        } else if (e->value != mp_const_none) {
```
robert-hh, line 247: What I did for NINAW10 is to create a temporary string object with to_copy set to addr_len as default. So if the / is missing, just the address is taken.
```
                            mp_obj_t ip_addr_obj = mp_obj_new_str(addr_str, to_copy);
                            netutils_parse_ipv4_addr(ip_addr_obj, ifconfig.ip_addr, NETUTILS_BIG);
```
The other option would be to change `netutils_parse_ipv4_addr()` to stop parsing a number at '/' like it does for '.'. Then e->value can be used.

robert-hh: So I changed `netutils_parse_ipv4_addr()` in netutils.c to stop the IP scanning at a '/' by adding the lines
```
    char *split = strchr(addr_str, '/');
    const char *s_top = split ? split : addr_str + addr_len;
```
Then, no interim string object is needed and e->value can be used as input to `netutils_parse_ipv4_addr()`.

robert-hh: I changed `netutils_parse_ipv4_addr()` again to stop at any character not being a digit or '.'. That generally improves the quality of parsing and has the side effect, that anything up from a '/' is ignored, which we use here.

extmod/network_ninaw10.c
```diff
@@ -372,6 +375,147 @@ static mp_obj_t network_ninaw10_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+
+    if (kwargs->used == 0) {
+        // Get config value
+        if (n_args != 2) {
+            mp_raise_TypeError(MP_ERROR_TEXT("must query one param"));
+        }
+
+        switch (mp_obj_str_get_qstr(args[1])) {
+            case MP_QSTR_dhcp4: {
+                return mp_const_true;
```
felixdoerre, line 436: If I'm reading this (https://github.com/adafruit/nina-fw/blob/47875b775aee74c5f948ef2b955c9786b023a89e/arduino/libraries/WiFi/src/WiFi.cpp#L297) correctly, the ninaw10 has a dhcp client enabled, but disables it once you set a static IP (i.e. once you call `NINA_CMD_SET_IF_CONFIG`). And then it would need to be re-initialized to activate the dhcp client again. Maybe this behavior is something we should reflect here?

extmod/network_ninaw10.c
```diff
@@ -372,6 +375,147 @@ static mp_obj_t network_ninaw10_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+            }
+            case MP_QSTR_has_dhcp4: {
```
felixdoerre, line 438: I am not sure if it is guaranteed that we have a dhcp4 supplied address. Maybe obtaining the address and checking if it is zero could work. If we keep this keyword, I would add the same logic to the esp code. Otherwise, I would argue, that we could remove that keyword to indicate that the nic driver code "does not know".

robert-hh:
OK. What's next. I could either continue with ESP8266 or CC3200.

ports/esp32/network_common.c
```diff
@@ -154,6 +155,146 @@ static mp_obj_t esp_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+                        }
+                        touched_ip_info = 1;
+                        break;
+                    }
+                    case MP_QSTR_gw4: {
+                        netutils_parse_ipv4_addr(e->value, (void *)&info.gw, NETUTILS_BIG);
+                        touched_ip_info = 1;
+                        break;
+                    }
+                    case MP_QSTR_dns: {
```
felixdoerre, line 259: For esp, i've moved `dns` to the interface-local variant, it seems dns config is set per-interface here.

extmod/network_ninaw10.c
```diff
@@ -372,6 +375,147 @@ static mp_obj_t network_ninaw10_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+        if (n_args != 1) {
+            mp_raise_TypeError(MP_ERROR_TEXT("can't specify pos and kw args"));
+        }
+
+        for (size_t i = 0; i < kwargs->alloc; ++i) {
+            if (MP_MAP_SLOT_IS_FILLED(kwargs, i)) {
+                mp_map_elem_t *e = &kwargs->table[i];
+                switch (mp_obj_str_get_qstr(e->key)) {
+                    case MP_QSTR_dhcp4: {
+                        // NINAW10 does not allow to control DHCP setting
```
felixdoerre, line 468: Or should we check if, what the user sets is sensible (i.e. if the user sets it to `true` when its already enabled, and only ignore it if the user sets it to `false`), and throw an error if we cannot communicate that to the nic (i.e. if dhcp has been disabled by setting a static IP)?

felixdoerre:
I've incorporated the suggestions that were in the comments, so the esp32 code could work now. You can re-test that one. And/or start working on esp8266. I am not sure if esp8266 needs different code, or if that can re-use code from esp32, and if we can re-use code, should we try to extract the common logic somewhere?

extmod/network_ninaw10.c
robert-hh: I can do that. @iabdalkader since you made that driver, do you have more information?

robert-hh: > And then it would need to be re-initialized to activate the dhcp client again. Maybe this behavior is something we should reflect here?

That would require some static flag, that `NINA_CMD_SET_IF_CONFIG` was called.

ports/esp32/network_common.c
robert-hh: At least that's less code.

robert-hh:
OK. I'll re-test the ESP32. If you want to work at the ESP8266 code, I can go for the CC3200.
Or do you want to go for the esp32_ppp first?

robert-hh:
The ESP32 code works fine. One little hiccup.
After setting DHCP first True, then False, the dns was still set, which seems wrong. Then, after setting addr4 and gw, dns was shown as  0.0.0.0. Nothing that needs a fix.

extmod/network_ninaw10.c
robert-hh: The only reasonable alternative to ignoring is throwing an error, easiest by not supplying the key. That saves code space.

felixdoerre:
> OK. I'll re-test the ESP32. If you want to work at the ESP8266 code, I can go for the CC3200.
Or do you want to go for the esp32_ppp first?

I can go for `esp32_ppp` first, but that one also seems wired to me. It can only be used to set `dns` and not ip addresses, and retreive the whole 4-tuple. Also it does not set dns with something interface-specific, but with the lwip-global `dns_setserver`. Do you know if that would work as well instead of `esp_netif_set_dns_info`? And if so, maybe the best way is:

- make `dns` network-global, and only use `dns_setserver()`.
- for ppp implement a method to get addr+gw
- remove the dns set/get logic from `network_common`.

If that sounds sane to you, I'd continue with those changes, and you can pick esp8266 or cc3200 whatever you prefer.

robert-hh:
I added a commit to my branch for NINAW109 dealing with dhcp4. For better visibility, I kept that separate. You can pick that commit and squash it to the existing NINAW10 commit.

robert-hh:
I have no knowledge and no position about esp32_ppp. About dns setting per interface or global: that has to be tested. I can do that, if you have the code. dns setting was otherwise a network option. Although having a rather large function and some tweaky code in modnetwork.c seems not appropriate.

extmod/network_ninaw10.c
iabdalkader: > I can do that. @iabdalkader since you made that driver, do you have more information?

Hi, information about what specifically ? Can you provide more context ?

robert-hh: The discussion was about enabling and disabling DHCP in the NINAW10 module and whether there is a dedicated command/function to control it, or if it happens implicitly. The assumption now is: DHCP is enabled, when the NINAW10 module is intialized. And on any manual setting of the IP configuration it's disabled.

robert-hh:
A few build errors with the last commit for ESP32_PPP. The last one is just a typo.
```
/home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c: In function 'esp_network_ipconfig':
/home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c:168:31: error: implicit declaration of function 'dns_getserver' [-Werror=implicit-function-declaration]
  168 |                 ipaddr_ntoa_r(dns_getserver(0), addr_str, sizeof(addr_str));
      |                               ^~~~~~~~~~~~~
/home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c:168:31: warning: passing argument 1 of 'ipaddr_ntoa_r' makes pointer from integer without a cast [-Wint-conversion]
  168 |                 ipaddr_ntoa_r(dns_getserver(0), addr_str, sizeof(addr_str));
      |                               ^~~~~~~~~~~~~~~~
      |                               |
      |                               int
In file included from /home/robert/Downloads/esp-idf/components/lwip/lwip/src/include/lwip/sockets.h:46,
                 from /home/robert/Downloads/esp-idf/components/lwip/include/lwip/sockets.h:8,
                 from /home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c:45:
/home/robert/Downloads/esp-idf/components/lwip/lwip/src/include/lwip/ip_addr.h:245:38: note: expected 'const ip_addr_t *' {aka 'const struct ip_addr *'} but argument is of type 'int'
  245 | char *ipaddr_ntoa_r(const ip_addr_t *addr, char *buf, int buflen);
      |                     ~~~~~~~~~~~~~~~~~^~~~
/home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c:193:25: error: implicit declaration of function 'dns_setserver' [-Werror=implicit-function-declaration]
  193 |                         dns_setserver(0, &dns);
      |                         ^~~~~~~~~~~~~
In file included from /home/robert/Downloads/MicroPython/micropython/py/mpstate.h:35,
                 from /home/robert/Downloads/MicroPython/micropython/py/runtime.h:29,
                 from /home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c:36:
/home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c: At top level:
/home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c:206:57: error: 'network_ipconfig' undeclared here (not in a function); did you mean 'esp_network_ipconfig'?
  206 | MP_DEFINE_CONST_FUN_OBJ_KW(esp_network_ipconfig_obj, 1, network_ipconfig);
      |                                                         ^~~~~~~~~~~~~~~~
/home/robert/Downloads/MicroPython/micropython/py/obj.h:391:104: note: in definition of macro 'MP_DEFINE_CONST_FUN_OBJ_KW'
  391 |     {{&mp_type_fun_builtin_var}, MP_OBJ_FUN_MAKE_SIG(n_args_min, MP_OBJ_FUN_ARGS_MAX, true), .fun.kw = fun_name}
      |                                                                                                        ^~~~~~~~
/home/robert/Downloads/MicroPython/micropython/ports/esp32/network_common.c:158:17: warning: 'esp_network_ipconfig' defined but not used [-Wunused-function]
  158 | static mp_obj_t esp_network_ipconfig(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs) {
      |                 ^~~~~~~~~~~~~~~~~~~~
```

robert-hh:
Just needed top uncomment `#include "lwip/dns.h"`.

ports/esp32/network_common.c
```diff
@@ -34,6 +34,7 @@
 #include <string.h>
 
 #include "py/runtime.h"
+#include "py/parsenum.h"
```
robert-hh, line 37: A little bit below `#include "lwip/dns.h"` must be uncommented.

ports/esp32/network_common.c
```diff
@@ -154,6 +155,196 @@ static mp_obj_t esp_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+                        mp_raise_ValueError(MP_ERROR_TEXT("unexpected key"));
+                        break;
+                    }
+                }
+            }
+        }
+    }
+    return mp_const_none;
+}
+MP_DEFINE_CONST_FUN_OBJ_KW(esp_network_ipconfig_obj, 1, esp_network_ipconfig);
```
robert-hh, line 206: Another change at this spot:
`MP_DEFINE_CONST_FUN_OBJ_KW(esp_network_ipconfig_obj, 0, esp_network_ipconfig);`
1 -> 0. There may be 0 positional arguments.

ports/esp32/network_common.c
robert-hh: You have dns setting in network.ipconfig. Why do you keep it here at nic.ipconfig (and removed dns query here)?

felixdoerre: ooh, yes. I missed that. Does the `network.ipconfig(dns=)` work? Then I'll remove the remainder here.

robert-hh: `network.ipconfig(dns=)` works.

ports/esp32/network_ppp.c
```diff
@@ -266,6 +266,40 @@ static mp_obj_t ppp_ifconfig(size_t n_args, const mp_obj_t *args) {
 }
 MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(ppp_ifconfig_obj, 1, 2, ppp_ifconfig);
 
+static mp_obj_t ppp_ipconfig(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs) {
+
```
robert-hh, line 270: The PPP must be active, otherwise all requests fail. So better start this function with e.g.:
```
    ppp_if_obj_t *self = MP_OBJ_TO_PTR(args[0]);

    if (!self->active) {
        mp_raise_ValueError(MP_ERROR_TEXT("PPP not active"));
    }
```

felixdoerre: `ppp_ifconfig` also doesn't have such a guard, right?

ports/esp32/network_ppp.c
```diff
@@ -266,6 +266,40 @@ static mp_obj_t ppp_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+    if (kwargs->used == 0) {
+        ppp_if_obj_t *self = MP_OBJ_TO_PTR(args[0]);
+        struct netif *netif = ppp_netif(self->pcb);
+        // Get config value
+        if (n_args != 1) {
```
robert-hh, line 275: Must be: `if (n_args != 2) {`. The first argument is the PPP object.

ports/esp32/network_ppp.c
```diff
@@ -266,6 +266,40 @@ static mp_obj_t ppp_ifconfig(size_t n_args, const mp_obj_t *args) {
...
+            mp_raise_TypeError(MP_ERROR_TEXT("must query one param"));
+        }
+
+        switch (mp_obj_str_get_qstr(args[0])) {
```
robert-hh, line 279: Must be `(args[1]`, the second positional argument.

ports/esp32/network_ppp.c
robert-hh: It does not, but it does also not crash with a core dump if not active. The guard here is to prevent a core dump crash. I assume it crashes here:
`struct netif *netif = ppp_netif(self->pcb);`

felixdoerre: aah right, the other path has an `self->pcb != NULL`. Then let's check for the same condition.

robert-hh: Fine, then let's go for that test,

robert-hh:
Good morning, Felix. Some topics:
- The PPP ipconfig for ESP32 seems fine. I tried as well to run PPP with a GSM modem, but that does not seem to work. But the hardware is old, and maybe not supported any more by the network. Nevertheless, ipconfig() seems fine.
- There are a few CI build error. I cannot replicate them locally. Here. it builds fine. And the errors are related to TinuyUSB. So maybe the CI has to be updated.
- I made another change to NINAW10, rolling back the modification to extmod/modnetwork.* and putting the respective definitions in the RP2 boards files. The RP2 port is currently the only one using NINAW10, but I have two PRs open for SAMD and MIMXRT which use NINAW10 as well. That again would be am argument to make changes at a single place, like it is now.
- If you want to carry on with ESP8266, I'll take care with the CC3200. The ESP8266 is more close to the ESP32 than the CC3200, which is a very old port and different approach in many aspects.

extmod/network_ninaw10.c
iabdalkader: Yes that seems to be the case. Although I can't find an explicit call to `tcpip_adapter_dhcpc_start` in the nina firmware, but it seems to be enabled by default and disabled on static IP config.

robert-hh:
I did not push the change for NINAW10 ye. I'm not sure if that is a good one. It "just" moves the changes for extmod/modnetwork.c from the file itself to files included by modnetwork.c. But these have to be adapted for every port using NINAW10.

I see you're fast with ESP8266. But there are quite a few build errors. You can see them here: https://github.com/micropython/micropython/actions/runs/8479596701/job/23233876028?pr=14199

extmod/network_ninaw10.c
robert-hh: Thanks.

robert-hh:
There were too many changes to the ESP8266 network_wlan.c to add comments. So it's attached here. It compiles and runs, and the ipconfig() options work. I tested only the AP mode.

[network_wlan.zip](https://github.com/micropython/micropython/files/14803158/network_wlan.zip)

robert-hh:
Updated one line:
[network_wlan.zip](https://github.com/micropython/micropython/files/14803248/network_wlan.zip)

felixdoerre:
> I see you're fast with ESP8266. But there are quite a few build errors. 

I got distracted and didn't look for compile errors yet. Thx for fixing my code :).

robert-hh:
@felixdoerre The support for the CC3200 is now in my branch. Tested with a WiPy board and STA mode.
Just take the last commit with CC3200 code and NOT the commits for NINAW10.

robert-hh:
If you think it's ready, you can change the state from draft to open.

robert-hh:
@felixdoerre There was still something to fix in extmod/network_lwip.c, when using ipconfig(addr="a.b.c.d") without the number for the netmask. The other ports would then just set the IP addr, but for LWIP ports the value for the address was undefined.
You should merge that change with the other change for extmod/network_lwip.c

felixdoerre:
Hmm, when you now set the address without netmask, it assumes `/32`. Is that reasonable? Maybe we should assume `/24` as a sane default.

robert-hh:
Right & thanks. The intention was to not touch the netmask at all, like the ESPxx, NINAW10 and CC3200 does at the moment.

robert-hh:
Updated. I was caught by the deferred setting of ip_addr and netmask. Please cross-check.

robert-hh:
@felixdoerre While adding ipconfig() to another port which I support I noticed, that the code setting addr4 can be simplified. addr_str and to_copy are in most ports not needed. I pushed already the changes for C3200 and NINAW10 to the usual place. ESP32 and ESP8266 can be simplified similarly and had the behavior, that when skipping the /nnn in CIDR notation the netmask was 255.255.255.255. Updated files to ESP32 and ESP8266:
[network_common.zip](https://github.com/micropython/micropython/files/14813386/network_common.zip)  for the ESP32 port
[network_wlan.zip](https://github.com/micropython/micropython/files/14813387/network_wlan.zip) for the ESP8266 port.

felixdoerre:
Thx for cleaning up. Yes. This looks reasonable. I've incorporated all those changes.

robert-hh:
Thanks. One thing that I not really comfortable with is the dns setting in network.ipconfig(). Even if I understand the reason, it's kind of bumpy, because one cannot set all parameters with one call. It requires network.ipconfig() for dns and nic.ipconfig() for the other parameters.
Is there a common practice of reference for ipconfig() and it's parameters?

felixdoerre:
We discussed options in #12600. I am not aware of any other `ipconfig` method (like e.g. from cpython) that we modeled this after. Maybe the thing that fits the closest is the low-level interface to configure networking on linux. Do you know how the esp interfaces behave, if you have multiple and set/receive via dhcp different DNS servers on both? From what I have seen in the lwip library, the two dns servers should override each other, which can be a bit irritating, when you configure two different nics.

When you have a more complete complex linux system with `resolved`, that service manages the per-link dns servers and tries to intelligently switch between them, but given we don't have such a component, I'd feel that per-link dns servers are irritating.

Initially, I also envisioned to set the default gw globally, so one can actively choose where packages are routed per default, but it seems we left that out.

(and additionally, `prefer` is also global and only makes sense globally. That would be even more irritating if set per link, so if we have ipv6 we need that global method anyways).

robert-hh:
Interestingly the last comment of Damien in #12600 uses a DNS setting in the nic.ipconfig() method.

felixdoerre:
Yes, and when I implemented the first round for lwip, I had dns in the nic method, but Damien hinted to keep the split between global and nic-config: https://github.com/micropython/micropython/pull/13689#issuecomment-1965742005

robert-hh:
Maybe that can be re-considered. I most cases just one NIC is used. And most ports allow just one NIC well at a time. The ESP32 port for instance allows to use both WiFi and Ethernet at the same time in a good fashion, but for other ports this either does not work or is somewhat unresponsive. At least it's that what I have seen.

felixdoerre:
A similar thing, is that activating dhcp is now non-blocking, meaning that if you want to make sure, you have an IP before you continue requires additional calls. Maybe the easiest way for "common usage" would be to have one "network-config" method, that can accept one dict, and then also deal with wifi-phy setup and similar? Because the goal, to configure networking in one call, is already not possible in most cases.

projectgus:
These code changes LGTM, although I think at some point (not this PR) we should look at refactoring out some of the common parts - for example splitting an IPV4/NETMASK string is implemented at least four times and they're not all the same.

There might even be a way to do this like the recent extmod/machine changes where the port supplies some static functions for each action, and there's one common argument parsing and verification routine.

But not for this PR, this LGTM.

> Maybe the easiest way for "common usage" would be to have one "network-config" method, that can accept one dict, and then also deal with wifi-phy setup and similar? Because the goal, to configure networking in one call, is already not possible in most cases.

How difficult do you think this would be to implement on top of what's here? Agree it'd be a nice helper method for the common use case, although I worry a bit about the complexity of all the smaller details.

felixdoerre:
> How difficult do you think this would be to implement on top of what's here? Agree it'd be a nice helper method for the common use case, although I worry a bit about the complexity of all the smaller details.

I am not sure how difficult this will turn out. Some challenge might be, that the order of applying config options matters sometimes. I'd imagine something like this:

```
{
 "wlan": {
  "module": "WLAN",
  "params": [network.STA_IF],
  "connect": ["ssid", "password"],
  "config": {"txpower": 7},
  "ipconfig": {"addr4":"10.0.4.2/24", "gw4": "10.0.4.1", "autoconf6": True}
  },
   "wlanAP": {
  "module": "WLAN",
  "params": [network.AP_IF],
  "connect": ["ssid", "password"],
  "config": {"security": 1},
  "ipconfig": {"addr4":"10.0.5.1/24", "addr6": "...."}
  },
  "ipconfig": {"dns": "10.0.4.1", "prefer": 4)
}
```

Applying this, should be possible in a few simple calls. I am not sure how one would convey, that you want to block until `dhcp`/`autoconf` is ready.

But I guess testing all the "rough corners" and making sure, every thinkable network configuration works, will be the tougher challenge. Also making sure that the semantics of this dict is the same across all boards. I am pretty sure, the format and semantics would need quite a few more discussions.

dpgeorge:
Now merged!

Regarding "easy network configuration": this is probably a good idea to have, to help users navigate the new complexity in address configuration.  At the very least we need good docs (there's #14140 for that).  Then eventually we can work towards some helper functions (written in Python) to make network configuration simpler.

razvanphp:
I've tested those changes on my Portenta H7 (STM32) and my problem with setting a static IP address still happens in `v1.24.0-preview.315.g8feb714b4`

Maybe you guys could have a look? https://github.com/micropython/micropython/issues/15877

Thank you!

PR #14206, opened 2024-03-28
extmod/modtls_axtls: Add verify_mode and CERT_NONE constant.

dpgeorge:
Supporting `verify_mode` and `CERT_NONE` is required for the new `ssl.py` code, as well as `requests` to work.


dpgeorge:
This also requires https://github.com/micropython/micropython-lib/pull/836

Then SSL requests work again on esp8266.

Issue #14092, opened 2024-03-14
RP2 - lightsleep() fails/does not exit when started from inside a thread - solution


mungewell:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

rp2 - official Pico (non-wifi)

### MicroPython version

MicroPython v1.22.2 on 2024-02-22; Raspberry Pi Pico with RP2040

### Reproduction

Run the attached script...

[blink_thread.py.txt](https://github.com/micropython/micropython/files/14596234/blink_thread.py.txt)

### Expected behaviour

When called within (only) a function, `lightsleep()` correct exits and the rest of the function are run.
```
MPY: soft reboot
Blink in thread test
CPU Freq 100000000
entering lightsleep
wake from sleep
CPU Freq 125000000
```

### Observed behaviour

The `lightsleep()` call does not exit....
```
MPY: soft reboot
Blink in thread test
CPU Freq 100000000
entering lightsleep

...
```

### Additional Information

No, I've provided everything above.

mungewell:
I commented out various bits of the `lightsleep()` code, and noted that it DOES work from within a thread - if it does NOT call `__wfi()`
```
         scb_hw->scr |= M0PLUS_SCR_SLEEPDEEP_BITS;
+#if 0 //SDW
         __wfi();
+#endif //SDW
         scb_hw->scr &= ~M0PLUS_SCR_SLEEPDEEP_BITS;
```

So does this mean that the (previously set) timer interrupt does not occur, if running from within a thread? And if so, why?

mungewell:
Seems that IRQs are NOT enabled, on the core that the code is running.... if we enable them, then the `lightsleep()` functions OK.

Patch (mixed in with my other stuff) is something like
```
@@ -165,18 +233,25 @@ STATIC void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
     } else {
         uint32_t sleep_en0 = clocks_hw->sleep_en0;
         uint32_t sleep_en1 = clocks_hw->sleep_en1;
-        clocks_hw->sleep_en0 = CLOCKS_SLEEP_EN0_CLK_RTC_RTC_BITS;
[...]

PR #14209, opened 2024-03-28
ports/rp2: Prevent lock-up when `lightsleep()` called within thread.

mungewell:
As per Bug #14092 when called from within a thread the interrupts may not be enabled on current core, and thus the call to `lightsleep()` never completes.


dpgeorge:
Thanks for the fix!

PR #14212, opened 2024-03-29
samd/mcu: Guard static function with appropriate #if.

dpgeorge:


robert-hh:
Thanks. Looks good and saves some bytes, when not needed.

PR #14222, opened 2024-04-01
wiznet5k: Properly enable socket buffers for W5100(S).

greezybacon:
The W5100 and W5100S only have 4 available sockets and 16kB of socket buffer. Allocating 16kB to either the receive or transmit buffer of a single socket is not allowed, so the current setup does not change the allocation for socket 0 from the reset default. `ctlwizchip` is returning `-1` to indicate the error, but the response isn't being inspected and probably doesn't need to be.


dpgeorge:
Thanks for the fix!

PR #14238, opened 2024-04-03
py/obj.h: Fix C++ compatibility of _mp_obj_type_t struct initialisers.

vonasmic:


stinos:
I'm not sure something is actually wrong here; the macro does have a rather complex expansion though. Yet if you look at the preprocessed code everything seems fine, for instance here's the expansion of

```
MP_DEFINE_CONST_OBJ_TYPE(
    mp_type_vfs_posix_fileio,
    MP_QSTR_FileIO,
    MP_TYPE_FLAG_ITER_IS_STREAM,
    print, vfs_posix_file_print,
    protocol, &vfs_posix_fileio_stream_p,
    locals_dict, &vfs_posix_rawfile_locals_dict
    );
```
->
```
const mp_obj_type_t mp_type_vfs_posix_fileio = {
 .base = { &mp_type_type },
 .name = MP_QSTR_FileIO,
 .flags = ( ( 0x0080 ) | ( 0x0100 ) ),
 .slot_index_print = 1,
 .slot_index_protocol = 2, 
 .slot_index_locals_dict = 3,
 .slots = {
    vfs_posix_file_print,
    &vfs_posix_fileio_stream_p,
    &vfs_posix_rawfile_locals_dict,
} };
```

So can you elaborate what exactly is wrong here?
Also note that _mp_obj_empty_type_t and _mp_obj_full_type_t also use this order, so not chaging them doesn't seem right.

vonasmic:
> So can you elaborate what exactly is wrong here? 
I am trying to create a new Gyro class in my .cpp file.
```
MP_DEFINE_CONST_OBJ_TYPE(
    sensors_type_Sensor,
    MP_QSTR_Gyro,
    MP_TYPE_FLAG_NONE,
    make_new, (void *)sensors_Gyro_make_new,
    locals_dict, &sensors_Gyro_locals_dict);
```

But if I then try to build it, the build exits with
```bash
error: designator order for field '_mp_obj_type_t::flags' does not match declaration order in 'const mp_obj_type_t' {aka 'const _mp_obj_type_t'}
  760 | #define MP_DEFINE_CONST_OBJ_TYPE_NARGS_2(_struct_type, _typename, _name, _flags, f1, v1, f2, v2) const _struct_type _typename = { .base = { &mp_type_type }, .name = _name, .flags = _flags, .slot_index_##f1 = 1, .slot_index_##f2 = 2, .slots = { v1, v2, } }```
  
  When I change the order of mp_obj_type_t members, to match the definition then it works correctly

**Comment by stinos** [comment]
Aha. I think it then makes more sense to change the definitions of all `MP_DEFINE_CONST_OBJ_TYPE_NARGS_XX` macros so those get the correct order. And it's best to mention the reason explicitly in the commit message.

**Comment by vonasmic** [comment]
> MP_DEFINE_CONST_OBJ_TYPE

I tried that and it breaks something. For example now the self argument isn't passed when calling functions. Not sure why that happens though

**Comment by robert-hh** [comment]
This scheme was set up by @jimmo . So better let him look at it. He seems to be off today/this week.

**Comment by vonasmic** [comment]
but this solution also breaks something in unix port / coverage

**Comment by projectgus** [comment]
> Aha. I think it then makes more sense to change the definitions of all `MP_DEFINE_CONST_OBJ_TYPE_NARGS_XX` macros so those get the correct order. 

Agree, this is a lot safer to avoid accidental regression bugs like the ones @stinos noticed.

@vonasmic if you have the version that changes the macro expansion but doesn't work, feel free to push it here instead and we can take a look.

> And it's best to mention the reason explicitly in the commit message.

Yes, agree. I'd also point out calling this the "wrong order" is not the whole story.  This initialisation order is incompatible with C++, but it's correct C code. MicroPython doesn't promise full compatibility with C++, this is why the provided `cppexample` keeps most of the Python binding code in `examplemodule.c` instead of directly in the .cpp file.

(That said, I think Damien is generally happy to merge a PR to improve the situation for C++ if it doesn't risk regression bugs or make the code unnecessarily complex. So changing the macro implementations like @stinos suggests should be OK. But this might not be the last piece of C++ incompatibility that you find, the easier path might be to put the Python bindings in a .c file that calls your .cpp file.)

**Comment by vonasmic** [comment]
> @vonasmic if you have the version that changes the macro expansion but doesn't work, feel free to push it here instead and we can take a look.
>
I made a new push with changed order in MP_DEFINE_CONST_OBJ_TYPE_NARGS_XX

**Comment by stinos** [comment]
@vonasmic you only have to swap the flags in the definition, not the macro argument order, so

```
#define MP_DEFINE_CONST_OBJ_TYPE_NARGS_0(_struct_type, _typename, _name, _flags) \
  const _struct_type _typename = { .base = { &mp_type_type }, .flags = _flags, .name = _name }
```

not

```
#define MP_DEFINE_CONST_OBJ_TYPE_NARGS_0(_struct_type, _typename, _flags, _name) \
  const _struct_type _typename = { .base = { &mp_type_type }, .flags = _flags, .name = _name }
```

with the latter you're actually assigning the name to the flags so nothing works anymore.

vonasmic:
oh, that makes sense. Thanks @stinos, it works now

vonasmic:
It still fails on the Unix port / coverage

stinos:
No worries, that's looks like something not directly related to Micropython

projectgus:
This approach looks very low risk to me, thanks! :+1:

PR #14245, opened 2024-04-03
esp32/machine_timer.c: Restrict timer numbers for ESP32C3 to 0 and 1.

robert-hh:
The ESP32C3 has only two timers in one group. Previously trying to use timer 3 caused a core dump.

Addresses the issue uncovered in discussion #14235.


dpgeorge:
Thanks!

PR #14253, opened 2024-04-05
shared/tinyusb: Fix dynamic USB control callbacks for wLength==0.

projectgus:
In the case where an OUT control transfer triggers with wLength==0 (i.e. all data sent in the SETUP phase, and no additional data phase) the callbacks were previously implemented to return b"" (i.e. an empty buffer for the data phase).

However this didn't actually work as intended because b"" can't provide a RW buffer (needed for OUT transfers with a data phase to write data into), so actually the endpoint would stall.

The symptom was often that the device process the request (if processing it in the SETUP phase when all information was already available), but the host sees the endpoint stall and eventually returns an error.

This commit changes the behaviour so returning True from the SETUP phase of a control transfer queues a zero length status response.

### DIFF
--- a/docs/library/machine.USBDevice.rst
+++ b/docs/library/machine.USBDevice.rst
@@ -130,15 +130,25 @@ Methods
 
       Second argument is a memoryview to read the USB control request
       data for this stage. The memoryview is only valid until the
-      callback function returns.
+      callback function returns. Data in this memoryview will be the same
+      across each of the three stages of a single transfer.
+
+      A successful transfer consists of this callback being called in sequence
+      for the three stages. Generally speaking, if a device wants to do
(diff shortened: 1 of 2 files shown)


projectgus:
@dpgeorge Have taken the liberty of marking this fix for 1.23. It can be handled on the Python side without a code change, but it's uglier.

(Specifically, without this fix OUT control transfers with wLength==0 need the callback to return `bytearray(0)`. With this change they can return True which makes the Python code a lot easier to follow.)

dpgeorge:
I tested this with a DFU device that I wrote in Python using `machine.USBDevice` and it allowed me to simplify my code a little, which did need to handle the case of `wLength==0`.

PR #14265, opened 2024-04-07
Add nonblocking support for WIZnet 5k NIC

greezybacon:
This adds support for the WIZnet chip:
  - Non-blocking support (ie. `settimeout(0)` or `setblocking(False)`)
  - DHCP runs in the background and will auto-renew the address or update it if the DHCP server assigns a different one
  - DHCP runs on a dynamic socket, so it won't corrupt other usage of socket `#1`
  - DNS runs on a dynamic socket, so it won't corrupt other usage of socket `#2`
  - Interrupt pin is supported
  - Sockets are otherwise polled on a 64ms interval (based on the lwIP-based interface)
  - `send` and `recv` is supported on UDP sockets. (`send` requires a call to `connect`)
  - TCP_NODELAY option is supported
  - Listener socket is automatically reinstated if no sockets are available when a client connection is accepted.

Using the WIZnet TCP/IP offload (without lwIP) reduces the code size by about 37k (on the RP2040), and static RAM usage by 28k.


extmod/network_wiznet5k.c
```diff
@@ -110,9 +119,21 @@ typedef struct _wiznet5k_obj_t {
     wiz_NetInfo netinfo;
     uint8_t socket_used;
     bool active;
+    uint8_t *dhcp_buf;
+    uint8_t dhcp_retry;
```
greezybacon, line 123: `dhcp_retry` isn't used

extmod/network_wiznet5k.c
```diff
@@ -556,11 +712,28 @@ static mp_uint_t wiznet5k_socket_send(mod_network_socket_obj_t *socket, const by
...
+        }
     }
     return ret;
 }
 
 static mp_uint_t wiznet5k_socket_recv(mod_network_socket_obj_t *socket, byte *buf, mp_uint_t len, int *_errno) {
+    if (socket->timeout > 0) {
+        mp_uint_t start = mp_hal_ticks_ms();
+        while (getSn_SR(socket->fileno) == SOCK_ESTABLISHED && getSn_RX_RSR(socket->fileno) == 0) {
+            if (mp_hal_ticks_ms() - start > socket->timeout) {
```
dpgeorge, line 729: Because `socket->timeout` is signed `int32_t` this comparison may be signed, but it should be unsigned.  So it needs to be `(uint32_t)(mp_hal_ticks_ms() - start) > (uint32_t)socket->timeout`.

extmod/network_wiznet5k.c
```diff
@@ -590,12 +771,27 @@ static mp_uint_t wiznet5k_socket_sendto(mod_network_socket_obj_t *socket, const
...
     return ret;
 }
 
 static mp_uint_t wiznet5k_socket_recvfrom(mod_network_socket_obj_t *socket, byte *buf, mp_uint_t len, byte *ip, mp_uint_t *port, int *_errno) {
     uint16_t port2;
+
+    if (socket->timeout > 0) {
+        mp_uint_t start = mp_hal_ticks_ms();
+        while (getSn_SR(socket->fileno) != SOCK_CLOSED && getSn_RX_RSR(socket->fileno) == 0) {
+            if (mp_hal_ticks_ms() - start > socket->timeout) {
```
dpgeorge, line 787: Needs `uint32_t` casts.

ports/rp2/mpconfigport.h
```diff
@@ -218,6 +218,10 @@ extern const struct _mp_obj_type_t mod_network_nic_type_nina;
 #if MICROPY_PY_NETWORK_WIZNET5K
 extern const struct _mp_obj_type_t mod_network_nic_type_wiznet5k;
 #define MICROPY_HW_NIC_WIZNET5K             { MP_ROM_QSTR(MP_QSTR_WIZNET5K), MP_ROM_PTR(&mod_network_nic_type_wiznet5k) },
+// This Network interface requires the extended socket state for timeouts and non-blocking
```
dpgeorge, line 221: I think the extended state is only needed for UDP sendto functionality?

dpgeorge:
Thanks for this, in general it looks good.

I assume you tested it?

ports/rp2/mpconfigport.h
greezybacon: I think so.

The additional state is for POSIX semantics using `connect` with a UDP socket. For `connect`, then remote host and port are stashed in the extended state and used for future calls to `send`. The `sendto` method continues to require a remote address for each call.

Do I need a better comment there?

greezybacon:
@dpgeorge Thank you greatly for the review. I made the sign changes as requested. I had started some additional changes for DHCP renewal which I also completed and pushed.

Concerning testing, yes, I have tested with the W5500 and the RP2040 and also with the W5100S-EVB-PICO board (W5100S + the RP2040). The performance is similar to lwIP. I mostly tested:
- DHCP leasing and auto-renewal (leasing was previously implemented using a fixed socket)
- DNS A-record lookup (previously implemented using a fixed socket)
- Chargen service (download from device)
- Echo service (send/receive 16-byte packets from the device)
- With and without the interrupt pin configured

Concerning completeness, timeouts for sending are not implemented. I spent a long time working with the retry transmission registers on the device which ultimately implement TCP Retransmission Timeout (RTO). However, calculating the proper setting requires knowledge of the apparent round-trip time (RTT), which the chip does not make available. So I'm not sure how to set it properly and then also not how to ensure that packets are definitely retransmitted within the (user-configured) timeout period. I'm game to implement something simple like what is implemented for the receive timeouts.

ports/rp2/mpconfigport.h
dpgeorge: I think it needs a more accurate comment, because when I read this comment I was confused.

Actually, instead of repeating the comment in each port that uses this, it would be better to just enable the extended sockets here in the `mpconfigport.h` file without any comment, and then in `extmod/network_wiznet5k.c` have a check for this option and a comment, something like:
```c
// Requires extended socket state for send...
#if !MICROPY_PY_SOCKET_EXTENDED_STATE
#error WIZNET5K_PROVIDED_STACK requires MICROPY_PY_SOCKET_EXTENDED_STATE
#endif
```

dpgeorge:
> Concerning completeness, timeouts for sending are not implemented
> ...
> I'm game to implement something simple like what is implemented for the receive timeouts.

That sounds OK, implementing something very simple, like just wait for the send to succeed within the given timeout and if it doesn't succeed then raise an exception.

The timeout code in `ports/esp32/modsocket.c` may give some hints here.

greezybacon:
Okay- I added simple send timeouts based on the implementation for the receive timeouts. It watches the transmit free-space register (TX_FSR) to be non-zero within the timeout period. If it remains zero, then a timeout is raised. Otherwise, the number of bytes able to be sent/put in the buffer is returned.

Would you have any interest in a similar implementation for the W6100S (the IPv6 model of the chip)? I would enjoy implementing it.

greezybacon:
As of 314a6c23, the text/bss sizes are (approx. for RP2040):
```
wiznet5k + lwIP: text=499500 bss=33944
wiznet5k + ioLibrary: text=461764 bss=5840 
```
So, about 37k less flash and 28k less ram.

dpgeorge:
> Would you have any interest in a similar implementation for the W6100 (the IPv6 model of the chip)? I would enjoy implementing it.

It it much of an extra burden supporting this chip, is it similar enough to the W5x00 chips?  It doesn't look like it's supported by `lib/wiznet5k`.  Well, that should anyway be done in a separate PR.

greezybacon:
> It it much of an extra burden supporting this chip, is it similar enough to the W5x00 chips? It doesn't look like it's supported by `lib/wiznet5k`. Well, that should anyway be done in a separate PR.

There’s an io6Library that supports it, and because it can work in dual stack mode and so has a different state machine, it definitely makes sense to track it separately.

dpgeorge:
> There’s an io6Library that supports it, and because it can work in dual stack mode and so has a different state machine, it definitely makes sense to track it separately.

Do you mean it would need a completely separate `extmod/network_wiznet5k_ip6.c` driver?

Note that the lwIP version of the Wiznet driver can already support IPv6 (at least in principle it should be able to, it may need a few tweaks to the lwIP bindings to work).

------

Anyway, let's get this PR ready to merge based on the non-blocking support.  I see that there are quite a few commits in this PR.  That's good if they are all making independent changes, but some need to be squashed, like the formatting one.

@greezybacon can you pleas fix the failing CI and prepare the commits so they are ready for final review and merge?

greezybacon:
Sure thing. I’ll squash the commits and address the CI issues. 

I have no idea on the effort for the W6100 to used with lwIP or ioLibrary. Totally agree- I’ll work on that separately.

dpgeorge:
> I’ll squash the commits and address the CI issues.

You don't have to squash everything into one commit.  Use your judgement for what you think is reasonable!

greezybacon:
@dpgeorge I got it down to seven potentially interesting commits. I'd be happy to squash further if you'd like.

I also spent a little time working on multicast support via the `IP_ADD_MEMBERSHIP` socket option and think I have it working. Additionally, the TCP_NO_DELAY socket option is not implemented (although also technically not exactly implemented in the chip). Would you prefer for me to rebase on a later commit of the `master` branch and add those two in, or would you prefer a follow-up PR?

dpgeorge:
> Would you prefer for me to rebase on a later commit of the `master` branch and add those two in, or would you prefer a follow-up PR?

If these extra features are ready, please add them to this PR.

greezybacon:
@dpgeorge I pushed the support for the multicast and tcp-nodelay.

There's another pitfall I think we should address. The WIZnet chips only support a certain number of concurrent sockets. When one is registered as a TCP socket in listen mode, the chip will covert it to an established connection when the chip receives a SYN packet to the listening port, and the same, established socket is returned from the `accept` call. POSIX semantics expect that the listening socket will remain in a state to accept further connections. So the `accept` method will create a new socket and start listening on the same port.

There are three problems here. First, if there are no more sockets available, then it is not possible to re-listen. The established socket is returned from `accept`; however, the listening socket will no longer ever be in a state to accept connections. The second is that, even if there are more sockets available, there is a small albeit non-zero time when there is no listening socket on the chip. If a SYN packet is received within this time, the chip will issue an ICMP destination-unreachable packet which will trigger a connection-refused error on the client side.

The second issue is an unavoidable caveat of using the chip and I think should just be documented on the driver documentation page. The client should just retry a short time later.

Since the `asyncio` module (at least) will expect the listening socket to permanently remain in an `accept`able state (pun intended), I think the listening socket should be put into a dormant state where a call to `accept` will never fail; however, there may not be a listening socket at the present. In the background, the driver could automatically re-add the listening socket when there is an available socket on the device. Thereafter, the `accept` call will yield further clients.

But this gives way to the third issue. Considering that only so many sockets can be concurrently open on the chip at one time, perhaps an application designer would only want to allocate a small number of sockets to any particular service (such as one or two). In this case, I think the call to `accept` should NOT create a new listening socket if there are presently _n_ number of established client sockets. Then, when a client disconnects and there are fewer than _n_ established clients, the driver could open the listening socket again. If there are already _n_ clients connected, then there would be no listening socket and client connection attempts would fail naturally with connection-refused errors from the ICMP destination-unreachable packet sent.

Supposing this makes sense, I would propose we use the `backlog` argument for the `listen` method to define this limit. Instead of the maximum queue depth of sockets accepted by the system but not yet handled by the software, I propose we use it to define the maximum number of simultaneous established client connections (for the WIZnet listening sockets only). When the `backlog` limit is reached, the driver stops re-adding the listening socket until a client disconnects.

If it doesn't make sense, I have a solution I'm working on to implement the dormant listening socket when there are no available sockets on the chip without a limit on accepted connections.

greezybacon:
@dpgeorge after more thought on this one, I think it would make more sense to add a socket option to define the simultaneous connection limit. It turns out that the `backlog` parameter has a default, so not specifying it from the application would confusingly impose an arbitrary limit on connections.

I went ahead and added support to hold the socket in a dormant state while it is considered to be listening but not able to have an associated socket on the chip due to unavailability.

dpgeorge:
@greezybacon thanks for the detailed explanation above regarding `accept` and the socket limit.

These issues, and the effort you needed to put into this PR, are the reason I much prefer to put the Wiznet in MAC-raw mode and use the on-MCU version of lwIP (rather than the Wiznet TCP/IP stack).

> Using the WIZnet TCP/IP offload (without lwIP) reduces the code size by about 37k (on the RP2040), and static RAM usage by 28k.

Is this the main motivation for this PR, to reduce code size and RAM usage?  Note that it's possible to reduce RAM usage of on-MCU lwIP by tuning the options in `lwipopts.h` (but it comes at the expense of performance).

-----------------

Regardless of the above comment, I'm happy that you've done this work and happy to merge it.

Is this PR ready for final testing/review?  It would be good not to increase the scope of this PR anymore.

greezybacon:
First of all, the multicast feature does not work. I think we should take it out and add it via an independent PR. I've tried methods outlined in the data sheet and several forum posts and continue to be unable to receive a multicast mDNS packet. I'll go ahead and remove it for final testing/review.

----

Concerning motivation, there are several motivating factors:
1. According to WIZnet, when using the chip this way, if a vulnerability in the network stack on the chip were exploited, it would not result in the compromise of the MCU itself.
2. It seems like this would be a more efficient architecture for an MCU, since the overhead of dealing with IP packets is completely offloaded
3. It was a firmware option to use the chip this way, and I noticed it didn't work. I was curious as to the benefits and wanted to explore using it
4. I ended up with a rather large Python project that compiled to about 100kB of memory space and was running into memory pressure on the RP2040. This was one angle of fixing it. But moving the project to the manifest files and including it with the firmware was a simple solution.

Interestingly, the chip-socket setup operates more deterministically than the LWIP version. Consider having a server which accepts short-lived client connections (say, for testing), the LWIP will hold onto the sockets after disconnect for some time and will quickly arrive at a state where it will no longer accept connections for quite a while (like 60s). However, using the chip-based sockets, the socket connection is torn down and reset quickly and deterministically, and the device will continue to accept further client connections immediately.

I saw that tuning options for LWIP is possible in `lwipopts.h`, however, the file would have to be edited directly to do so, and I generally setup projects to use Micropython as a submodule and attempt use the code exactly as is. Would you consider wrapping many of the options there with `#ifndef` lines to allow a project using Micropython as submodule to define them outside the project? In case an example would help, here's a typical project layout:

```
- boards/
  - REV1/
    - boot/
      - ioconfig.py
      - main.py
    - Makefile
    - manifest.py
    - mpconfigboard.cmake
    - mpconfigboard.h
- lib/
  - micropython -> submodule
- src/
  - package/
    - __init__.py
    - manifest.py
```

I would prefer to define the overrides in the `mpconfigboard.h`, but it's not possible since Micropython will forcibly define the options in the `rp2` port.

Issue #14307, opened 2024-04-15
[webassembly][pyscript] Unable to process over 88Kb of code


WebReflection:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

webassemply

### MicroPython version

latest for pyscript

### Reproduction

We bootstrap our virtual env via FS operations and we port that whole operation via strings (as in Python strings) but we reached a memory limit so that everything breaks. Under 60Kb of single *run* we don't have issues but beyond that it's breaking.

### Expected behaviour

MicroPython should be able to run up to 256Kb of code or more to me or we should be able to raise that limit. We tried to explicitly increase the `heapsize` without any success.

### Observed behaviour

`runPython(moreThan88KCode)` fails, it doesn't on Pyodide, it doesn't if we manually remove some line to evaluate from that code.

### Additional Information

This file contains our bootstrap logic which is automatically generated and it travels across workers or projects to be sure our namespace is up and running by the time users' code executes.

It served us well as logic to date and we never had issues with MicroPython but a recent branch that would like to bring more is failing hard on it.

[micropython-limit.txt](https://github.com/micropython/micropython/files/14979983/micropython-limit.txt)

WebReflection:
This file, once renamed as `.html` will show the limit. This is without PyScript around.

Remove the line line 47 and see no error ... please note: you need to remove that line, as even commenting it, will still show the *memory out of bound* error, just to be clear it's not the code the issue, rather how much input one can *run*.

[micropython-limit.html.txt](https://github.com/micropython/micropython/files/14980167/micropython-limit.html.txt)

WebReflection:
@dpgeorge I don't know if this is useful but in this line maybe there's not enough memory allocated https://github.com/micropython/micropython/blob/53d0050255ef92df5adc9255380f170eb33bd2c9/ports/webassembly/api.js#L134 but I also wonder if this should be `JSFLAGS += -s ALLOW_MEMORY_GROWTH=1` instead https://github.com/micropython/micropython/blob/53d0050255ef92df5adc9255380f170eb33bd2c9/ports/webassembly/variants/pyscript/mpconfigvariant.mk#L1

dpgeorge:
[...]

PR #14318, opened 2024-04-17
webassembly/api: Allocate code data on C heap when running Python code.

dpgeorge:
Otherwise Emscripten allocates it on the Emscripten C stack, which will overflow for large amounts of code.

Fixes issue #14307.

### DIFF
--- a/ports/webassembly/api.js
+++ b/ports/webassembly/api.js
@@ -127,23 +127,31 @@ export async function loadMicroPython(options) {
         },
         pyimport: pyimport,
         runPython(code) {
+            const len = Module.lengthBytesUTF8(code);
+            const buf = Module._malloc(len + 1);
+            Module.stringToUTF8(code, buf, len + 1);
             const value = Module._malloc(3 * 4);
             Module.ccall(
                 "mp_js_do_exec",
                 "number",
-                ["string", "pointer"],
-                [code, value],
+                ["pointer", "number", "pointer"],
+                [buf, len, value],
             );
+            Module._free(buf);
             return proxy_convert_mp_to_js_obj_jsside_with_free(value);
         },
         runPythonAsync(code) {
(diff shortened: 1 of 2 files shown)


dpgeorge:
Find a compiled version of this PR here for testing: https://micropython.org/resources/wasm/micropython.mjs

ports/webassembly/api.js
```diff
@@ -131,23 +131,31 @@ export async function loadMicroPython(options) {
...
-                ["string", "pointer"],
-                [code, value],
+                ["pointer", "number", "pointer"],
+                [buf, len, value],
             );
+            Module._free(buf);
             return proxy_convert_mp_to_js_obj_jsside_with_free(value);
         },
         runPythonAsync(code) {
+            const len = Module.lengthBytesUTF8(code);
```
WebReflection, line 148: if `len` always needs a `+ 1` (for null code `\x00` I suppose) why isn't this just `const len = Module.lengthBytesUTF8(code) + 1;` instead? Not a blocker, just a silly question!

WebReflection:
I can confirm this fixes the presented use-case/issue we had 🙏

[index.txt](https://github.com/micropython/micropython/files/15013165/index.txt)

ports/webassembly/api.js
ntoll: Bravo

dpgeorge: `Module.stringToUTF8` will always add a null byte at the end, so there must be enough room for it.

But, below in the call to `mp_js_do_exec_async`, we pass just `len`.

dpgeorge:
Also added a fix for initialising `Module` with the latest Emscripten.

WebReflection:
@dpgeorge do you mind publishing this latest version to *npm* so we can also bump it? Thanks!

PR #14323, opened 2024-04-18
py/objfun.h: Fix C++ compatibility.

stinos:
Explicit casts are needed.  Fixes recent changes from 648a7578, 94002297.


dpgeorge:
Thanks for this.

I feel like we should have some simple CI that builds a C++ project which uses MicroPython, to catch these issues.

stinos:
> I feel like we should have some simple CI that builds a C++ project which uses MicroPython, to catch these issues.

A good start would be to `#include` the complete public API in examples/usercmodule/cppexample/example.cpp: that would have caught this particular issue. I'll do that shortly.

But apart from that getting complete coverage like for ce491ab0 there's no other way than to write out *all* macro invocations I think.

projectgus:
Maybe the usermodule.c in the cppexample should be made into a .cpp file as well, as this seems to be how C++ developers prefer to structure it? That will at least cover some possible macro expansions.

stinos:
> That will at least cover some possible macro expansions.

It's not a lot though.. Plus the initial idea was to have both .c and .cpp code so the .mk file would also show how to use both.

PR #14332, opened 2024-04-19
shared/readline: Implement CTRL-L (clear screen, redraw prompt)

### LINKED ISSUE #11354: REPL / built-in readline: Add CTRL-L support (clear screen)
Right now, the built-in readline implementation ([`shared/readline/readline.c`](https://github.com/micropython/micropython/blob/master/shared/readline/readline.c)) does not support CTRL-L, which is a somewhat common key to clear the the screen, and move the prompt and command line to the top of the screen. At least [bash supports it](https://ss64.com/bash/syntax-keyboard.html).

This feature could be optional, and features like `MICROPY_REPL_EMACS_KEYS` already are, but the size savings wouldn't very much.

I might implement this myself If I get around to it. If anyone else want to do it, please let me know here.

neuschaefer:
This commit implements CTRL-L, a somewhat common key to clear the the screen and move the prompt and command line to the top of the screen. At least zsh, fish, ruby, and GNU readline support it.

Fixes: https://github.com/micropython/micropython/issues/11354

### DIFF
--- a/shared/readline/readline.c
+++ b/shared/readline/readline.c
@@ -188,6 +188,11 @@ int readline_process_char(int c) {
             redraw_step_back = rl.cursor_pos - rl.orig_line_len;
             redraw_from_cursor = true;
         #endif
+        } else if (c == CHAR_CTRL_L) {
+            mp_hal_stdout_tx_str("\x1b[2J\x1b[H");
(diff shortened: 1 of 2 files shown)


dpgeorge:
Can you please rebase again on latest master, to pick up #14349?

PR #14336, opened 2024-04-20
pyboard.py: Default to the last available serial port on the system

daid:
A small addition to the pyboard.py upload tool. This makes it default to the last serial port available on the system instead of `/dev/ttyACM0`.

My main reason for adding this is that port numbers on windows are not very predictable, and looking up the port each time is quite annoying.
The reason for picking the last port in the list, is that physical RS232 ports (even if they are rare) are first in the list before USB ports. So the last thing in the list is the thing you most recently plugged in. So very likely your target.

End result of this change is that not specifying a port on Windows works. And not specifying a port on Linux will act slightly different if you have multiple boards connected.

### DIFF
--- a/tools/pyboard.py
+++ b/tools/pyboard.py
@@ -766,12 +766,19 @@ del _injected_buf, _FS
 
 def main():
     import argparse
+    import serial.tools.list_ports
+
+    default_device = "/dev/ttyACM0"
+    available_devices = serial.tools.list_ports.comports()
+    if available_devices:
+        default_device = available_devices[-1].name
+    default_device = os.environ.get("PYBOARD_DEVICE", default_device)
 
     cmd_parser = argparse.ArgumentParser(description="Run scripts on the pyboard.")
     cmd_parser.add_argument(
         "-d",
         "--device",
-        default=os.environ.get("PYBOARD_DEVICE", "/dev/ttyACM0"),
+        default=default_device,
         help="the serial device or the IP address of the pyboard",
     )
     cmd_parser.add_argument(


Josverl:
Why?
This would break the common workflow and expectation for a reverse one which would be rather surprising, to say the least

And if you are talking about the last one, should COM11: be fefire, or after COM3?

daid:
> Not specifing a port never has failed me on Windows, nor on Linux.

Then most likely, you have an environment variable (`PYBOARD_DEVICE`) set on Windows (or something else set it for you). As it cannot work without a parameter or environment setting in Windows. On Linux, it's quite likely that your board is /dev/ttyACM0

> This would break the common workflow and expectation for a reverse one which would be rather surprising, to say the least

I think you are wrong here, it doesn't break a common workflow. As the default of /dev/ttyACM0 isn't valid on Windows, and very likely the one and only or last detected serial port on Linux.

> And if you are talking about the last one, should COM11 be before, or after COM3?

COM11 comes after COM3 in the listed serial ports from the pyserial utils.

> Also I find that port numbers on windows are highly predictable and stable across plug in, usb hubs and reboots of both host and MCU

Windows ensures the same device produces the same port number. However, I'm not plugging in the same device all the time, I'm switching between a lot of devices. In which case Windows produces different port numbers for each.

> while my limited experience on Ubuntu has thought me that is not always the case, and plug-in order and MCU restarts can result in differ port names.

Per default (unless you setup udev rules) Linux just numbers them by detection order, so the first becomes /dev/ttyACM0, 2nd /dev/ttyACM1, and if you unplug the first one and plug in a different board that now becomes /dev/ttyACM0. UNLESS something still had the port /dev/ttyACM0 open, in which case that is still "reserved" and then it becomes /dev/ttyACM2. Which is why it can feel unpredictable on Linux.

> What is the root cause of the issue you are trying to solve? Is there perhaps a BT set rial port set at a lower port than your MCU?

The root cause of my issue is that I'm swapping a lot of different boards and Windows all gives these different port numbers.


A different solution that wouldn't change the current (IMHO broken on Windows) default behavior is to have a `--device auto` as parameter to auto find which serial port to use.

andrewleech:
I too use a _lot_ of different micropython devices, on both Windows and Linux (WSL2)

Yes on windows each device gets its own COM number which is stable across connect / disconnect for most devices. For me, `mpremote` finds them just fine and automatically connects:
```
C:\Users\anl>mpremote --version
mpremote 1.22.0
C:\Users\anl>mpremote
Connected to MicroPython at COM40

>>>
C:\Users\anl>mpremote
Connected to MicroPython at COM18
MicroPython v1.22.0-preview.87.gcf0a8ce702.dirty on 2023-11-03; WEACTV20_F411 with STM32F411CE
```
That was me plugging in a RPI Pico first, run mpremote, found it at COM40. I then unplugged it and connected a STM32 Black Pill instead, mpremote found that just fine at COM18.
I've done _nothing_ to pre-configure mpremote or any other windows setting / environment relating to these com ports or where to find them, this should be all default behavior here already. 

It's true I don't have any physical COM ports, however as per [the docs](https://docs.micropython.org/en/latest/reference/mpremote.html#auto-connection-and-soft-reset) and this [check in the code](https://github.com/micropython/micropython/blob/880f7bc0406a1ef249c833e1a88c023cd339ecf6/tools/mpremote/mpremote/commands.py#L38) mpremote auto-connect only looks at USB serial ports, it should skip over any built in ones already.

In terms of making it easier to lookup what's connected, you can do `mpremote connect list` to see what it sees.

As an alternate suggestion, I also use my [WSL USB Manager](https://gitlab.com/alelec/wsl-usb-gui/-/releases) app to forward USB devices from Windows to WSL to use linux tools to communicate / program devices. The gui app is also handy as a quick visual reference of usb devices connected, when not forwarded to linux it displays their com port, eg. 
image

Issue #14339, opened 2024-04-20
[RP2] machine.soft_reset() hangs when i2s.deinit() is not called


marecl:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

RPI_PICO

### MicroPython version

MicroPython v1.23.0-preview.324.gd11ca092f.dirty on 2024-04-20; Raspberry Pi Pico with RP2040

### Reproduction

1. Initialize I2S device
2. Call `machine.soft_reset()`

### Expected behaviour

machine.soft_reset() resets machine

### Observed behaviour

device hangs, requiring hardware reset or reconnecting USB

### Additional Information

This issue is work-in-progress. I've observed it during development of my application. At this time please treat this like a general observation.
Firmware is slightly customized with added MSC in board configuration. It has it's own issues but I didn't observe any connection.
I've observed that `main.c` calls I2S init on startup, but doesn't deinit it during soft reset.

marecl:
Solved in mentioned PR. Waiting for feedback on the issue.

marecl:
#14345

dpgeorge:
Thanks for the detailed report and reproduction.  I can confirm the issue on a Pico.  And the same issue exists on the stm32 port (tested with a PYBv1.0), and I assume it also exists on the mimxrt port.

The fix would be to enable finalisers on I2S objects for all the ports, following how esp32 does it.  That would also fix other cases of I2S instances being GC'd without `.deinit()` being called on them by the user.

robert-hh:
The mimxrt port de-inits I2S on soft reset. But as for enabling finalizers: during soft reset `gc_sweep_all()` is called. I had the impression that this call clears each and every object from the heap, whether created with finalizer or not.

robert-hh:
The i2s object contains usually pointers to buffers, allocated with m_new(). Assuming that `gc_sweep_all()` does not free objects in a specific order it may happen, machine_is2_deinit() as part of calling __del__ with it is called after the buffers have already been cleared. 
What happens is  m_free() is called on already free'd object?

dpgeorge:
> The mimxrt port de-inits I2S on soft reset

[...]

PR #14345, opened 2024-04-21
ports/rp2/machine_i2s.c: Detect and deinitialize I2S instances on boot.

marecl:
In reference to #14339  

`machine_i2s_init0` makes all of these pointers NULL without checking if there was a pointer already. Furthermore, garbage collector doesn't care enough to call `I2S.deinit()`.
This created an issue, where invoking `machine.soft_reset()` would hard crash an entire board.  
Fix is a check if `machine_i2s_obj` contains an object (assumed to be a valid `machine_i2s_obj_t`). If memory address is not NULL - object gets deinitialized. No further crashes are observed.


dpgeorge:
> Furthermore, garbage collector doesn't care enough to call I2S.deinit().

To fix that, finalisers should be enabled on the I2S object on all ports, following how esp32 does it.  That requires two things:
1. Remove the `#if MICROPY_PY_MACHINE_I2S_FINALISER` guard from `extmod/machine_i2s.c`.
2. Make rp2, stm32, mimxrt ports all use `mp_obj_malloc_with_finaliser()` instead of `mp_obj_malloc()` in their `mp_machine_i2s_make_new_instance()` implementations.

robert-hh:
If you do not want to follow the recommendation of Damien, this deinit should better be placed at the end of main() between the label `soft_reset_exit:` and the call to `gc_sweep_all();`.  At that point the I2S data structure is still intact and `mp_machine_i2s_deinit()` can be called safely. After `gc_sweep_all()` has run, all allocated data elements can not be considered any more as valid, which includes the I2S object and it's buffers. So they must not be used in machine_i2s_init0(). You might have to create a function like `machine_i2c_deinit_all()` in machine_i2s.c which is called at behind label `soft_reset_exit:` in main.c.

marecl:
> If you do not want to follow the recommendation of Damien, this deinit should better be placed at the end of main() between the label `soft_reset_exit:` and the call to `gc_sweep_all();`. At that point the I2S data structure is still intact and `mp_machine_i2s_deinit()` can be called safely. After `gc_sweep_all()` has run, all allocated data elements can not be considered any more as valid, which includes the I2S object and it's buffers. So they must not be used in machine_i2s_init0(). You might have to create a function like `machine_i2c_deinit_all()` in machine_i2s.c which is called at behind label `soft_reset_exit:` in main.c.

I'll work on this, just don't have a lot of free time. I'd rather go with Damien's solution as it seems more appropriate and portable. I don't have other devboards, so most likely fixing it globally (like calling `__del__` or `__exit__`) would free us from meddling with hardware.

dpgeorge:
This was fixed differently in d2e33fe3096eec60d7017b0f17c4ddb6910a4d0b

PR #14370, opened 2024-04-24
examples/usercmodule: Include API headers in the C++ example.

stinos:
Make the CI builds compile the public API as C++ to catch accidental introductions of incompatible code.

See #14323

### DIFF
--- a/ports/unix/coveragecpp.cpp
+++ b/ports/unix/coveragecpp.cpp
@@ -1,5 +1,20 @@
 extern "C" {
-#include "py/obj.h"
+// Include the complete public API to verify everything compiles as C++.
+#include <py/gc.h>
+#include <py/obj.h>
+#include <py/objarray.h>
+#include <py/objexcept.h>
+#include <py/objfun.h>
+#include <py/objgenerator.h>
+#include <py/objint.h>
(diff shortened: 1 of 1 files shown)


projectgus:
Adding CI build coverage for C++ is good, but adding it as code in a user-facing example (that's primarily something someone can take, learn from, and adapt into their own code) seems not ideal. It's a lot of extra noise for someone who's just starting out with the C API.

MicroPython doesn't (to my knowledge) have a category of "build tests" (i.e. things that are functionally examples, but exist for project development and test purposes.) Maybe this could be the use case for the first one?

stinos:
Fair point. Practically: create a tests/ subdirectory for this code then? Preference for a name?

> i.e. things that are functionally examples, but exist for project development and test purposes

The tests/ directory has a bunch of things which fall under that category imo. The socket multi tests etc.

dpgeorge:
How about moving this into `ports/unix/coveragecpp.cpp`?  Seems like an obvious place for it.

stinos:
> How about moving this into ports/unix/coveragecpp.cpp? Seems like an obvious place for it.

Ok.

dpgeorge:

PR #14372, opened 2024-04-24
extmod/network_nina: Fix the AP security mode constants.

iabdalkader:
I was looking at the Nina firmware recently when I noticed that the command opcode `0x19` actually starts the AP in WPA/WPA2 mode. This has been long misunderstood because its handler was called `setApPassPhrase` implying that it's WEP. Even though a function exists in the firmware that attempts to set up an AP with WEP, it seems that this mode is not supported for AP.

I've verified this and the AP is using WPA/WPA2:
```bash
> nmcli dev wifi
BSSID              SSID  MODE   CHAN  RATE        SIGNAL  BARS  SECURITY  
84:CC:A8:30:4C:51  NINA  Infra  1     135 Mbit/s  100     ▂▄▆█  WPA1 WPA2 
```

There are no functional changes in this PR, it just fixes the constants names and removes the useless sanity checks for `WEP`.

### DIFF
--- a/drivers/ninaw10/nina_wifi_drv.c
+++ b/drivers/ninaw10/nina_wifi_drv.c
@@ -95,7 +95,7 @@ typedef enum {
 
     // AP mode commands.
     NINA_CMD_START_AP_OPEN          = 0x18,
-    NINA_CMD_START_AP_WEP           = 0x19,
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks!

Issue #14342, opened 2024-04-20
[RP2] When building with external board definition directory, machine.Pin.board is not populated.


ricksorensen:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

RP2

### MicroPython version

```
$ git describe --dirty
v1.22.0-325-gd11ca092f
$ mpremote
Connected to MicroPython at /dev/ttyACM0
Use Ctrl-] or Ctrl-x to exit this shell

>>> 
MPY: soft reboot
MicroPython v1.23.0-preview.324.gd11ca092f on 2024-04-20; XIAO RP2040 Local with RP2040
Type "help()" for more information.
```

### Reproduction

I am using linux to build, with an external definition of board in :
`/work/domicropy/rjsboards/SEEED_XIAO_RP2040` with contents:
```
$ ls /work/domicropy/rjsboards/SEEED_XIAO_RP2040/
board.json  mpconfigboard.cmake  mpconfigboard.h  pins.csv

comment: from ports/rp2, after prelims
$ make BOARD_DIR=/work/domicropy/rjsboards/SEEED_XIAO_RP2040/
$ mpremote bootloader 
# example for my linux:
$ cp build-SEEED_XIAO_RP2040/firmware.uf2  /media/rick/RPI-RP2
# reset board  
$ mpremote
>>> import machine
>>> dir(machine.Pin.board)
```

### Expected behaviour

Expected list of pins described in `pins.csv`, which is stored in external board definition 
```
>>>import machine
>>>dir(machine.Pin.board)
[...]

PR #14375, opened 2024-04-25
rp2/CMakeLists.txt: Enable pins.csv usage for out-of-tree boards

mgsb:
The current generation of pins.h hard-codes rp2/boards as the directory for the board being built. Change this to use the value of MICROPY_BOARD_DIR to find pins.csv for both in-tree and out-of-tree boards.

### DIFF
--- a/ports/rp2/CMakeLists.txt
+++ b/ports/rp2/CMakeLists.txt
@@ -537,8 +537,8 @@ set(GEN_PINS_MKPINS "${MICROPY_BOARDS_DIR}/make-pins.py")
 set(GEN_PINS_SRC "${CMAKE_BINARY_DIR}/pins_${MICROPY_BOARD}.c")
 set(GEN_PINS_HDR "${MICROPY_GENHDR_DIR}/pins.h")
 
-if(EXISTS "${MICROPY_BOARDS_DIR}/${MICROPY_BOARD}/pins.csv")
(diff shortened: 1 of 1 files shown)


Gadgetoid:
I proposed more or less the same thing here https://github.com/micropython/micropython/pull/13766

Though I also dropped `MICROPY_BOARDS_DIR` since it doesn't get used often enough to warrant the extra obfuscation.

So yeah 👍 for this, all of our builds are using out of tree boards dirs and finding pins.csv fails and fails silently.

ricksorensen:
Addresses/closes #14342 also

mgsb:
> I proposed more or less the same thing here #13766
> 
> Though I also dropped `MICROPY_BOARDS_DIR` since it doesn't get used often enough to warrant the extra obfuscation.
> 
> So yeah 👍 for this, all of our builds are using out of tree boards dirs and finding pins.csv fails and fails silently.

Ooops, sorry I missed yours. Maybe two PRs will help expedite merging :)

Gadgetoid:
> Maybe two PRs will help expedite merging :)

I hope so! 😆

dpgeorge:
The alternative PR was merged in 932f76c6ba64c5a3e68de3324556d9979f09303b

mgsb:
Thanks!

Issue #14371, opened 2024-04-24
Mbedtls: load_cert_chain NULL terminator uncertainty


peterzuger:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

unix, stm32, pybd

### MicroPython version

MicroPython v1.20.0-1189.g49ce7a607 on 2024-04-24; linux [GCC 13.2.1] version

### Reproduction

Generate the specific type of DER encoded KEY/CERT pair.
```
openssl genrsa -out key.pem 1024

openssl req -x509 -newkey rsa:1024 -keyout key.pem -out cert.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=CommonNameOrHostname"

openssl rsa -in key.pem -outform DER -out key.der
openssl x509 -in cert.pem -out cert.der -outform DER
```

Run interactivley or via script on unix or any STM32 port (all mbedtls ports will be affected)
```
import tls
ctx = tls.SSLContext(tls.PROTOCOL_TLS_CLIENT)

key = open("key.der", "rb").read()
cert = open("cert.der", "rb").read()

ctx.load_cert_chain(cert, key)
```

### Expected behaviour

This script is expected to run without issue.

### Observed behaviour

```
Traceback (most recent call last):
  File "test.py", line 8, in <module>
ValueError: invalid key
```

### Additional Information

As described [here (mbedtls_pk_parse_key)](https://mbed-tls.readthedocs.io/projects/api/en/development/api/file/pk_8h/?highlight=mbedtls_pk_parse_key#_CPPv420mbedtls_pk_parse_keyP18mbedtls_pk_contextPKh6size_tPKh6size_tPFiPvPh6size_tEPv):

> Size of key in bytes. For PEM data, this includes the terminating null byte, so keylen must be equal to strlen(key) + 1

But for DER data it must:

> The buffer must contain the input exactly, with no extra trailing material.

The behaviour of mbedtls has changed since the switch to v3.5, it now checks that the provided key buffer does not contain any additional data but only of unencrypted PKCS8 data.

[...]

PR #14385, opened 2024-04-27
extmod/modtls_mbedtls: Fix key_len passed to mbedtls_pk_parse_key.

peterzuger:
See issue #14371:

mbedtls_pk_parse_key() expects key_len to include the NULL terminator for PEM data but not for DER encoded data.

Since all PEM data starts with "-----BEGIN" this is used to check if the data is PEM.

This can be done for both v2 and v3 of mbedtls since the fundamental behaviour/expectation did not change. 

What changed is that in v3 the `PKCS#8` DER parser now checks that the passed key buffer was fully utilized and no bytes are remaining (all other DER formats still do not check this).


dpgeorge:
Thanks for the patch.  This looks OK, except that actually mbedtls is not configured to be able to parse PEM files at all.  To enable that requries:
```c
#define MBEDTLS_PEM_PARSE_C
#define MBEDTLS_BASE64_C
```

As such, it's not possible to write a test for the code added in this PR.

What I suggest is to change all cert/key parsing code in `modtls_mbedtls.c` so that it does not add a +1 at all, because only DER is supported.

And then add optional support for PEM parsing via the following (essentially what this PR does but with a if guard):
```c
    #if MBEDTLS_PEM_PARSE_C
    // len should include terminating null if the data is PEM encoded
    if ((key_len >= 10) && (memcmp(key, "-----BEGIN", 10) == 0)) {
        key_len += 1;
    }
    #endif
```

peterzuger:
Yes, I enabled PEM parsing while testing for this bug/patch. The support for PEM parsing was mostly added to this PR to not break the code of people who use PEM.

This Patch is mostly to fix the PKCS8 DER parsing that was broken by the update of mbedtls.

I have added the ```#if``` guard to the patch.

dpgeorge:
Do the calls to `mbedtls_x509_crt_parse()` also need a similar fix?  It looks like that function can parse both DER and PEM as well (the latter only when PEM is enabled), and so +1 should only be added to the length if it's PEM.

> This Patch is mostly to fix the PKCS8 DER parsing that was broken by the update of mbedtls.

Is it possible to add a test which fails prior to this PR but passes after this PR?  In other words, can we add a test that has a PKCS8 DER certificate?  How to generate such a certificate?

peterzuger:
After taking a look at `mbedtls_x509_crt_parse()` and its documentation [here](https://os.mbed.com/teams/Arcola/code/mbedtls/docs/tip/group__x509__module.html#ga033567483649030f7f859db4f4cb7e14).

When `MBEDTLS_PEM_PARSE_C` is disabled this function simply passes the arguments to `mbedtls_x509_crt_parse_der()`

The documentation states the same thing as `mbedtls_pk_parse_key()`:
> `buflen	size of the buffer (including the terminating null byte for PEM data)`

The problem seems to be possible, at the end of the function `x509_crt_parse_der_core()` in `x509_crt.c` mbedtls does a similar check to ensure the entire buffer was used. But they also do some custom length checking/parsing that seems to bypass this by modifying the `end` ptr ?

```
if( p != end )
{
    mbedtls_x509_crt_free( crt );
    return( MBEDTLS_ERROR_ADD( MBEDTLS_ERR_X509_INVALID_FORMAT,
            MBEDTLS_ERR_ASN1_LENGTH_MISMATCH ) );
}
```
After some time searching I could not find a way to create a Certificate that triggers this condition causing the parsing to fail.


As for the test, in the next iteration I will add a certificate generated like described in #14371 to the tests which will currently fail but will be fixed with this PR.

Should I include the fix for `mbedtls_x509_crt_parse()` in this PR or should I open a separate one since there is currently no test I can think of to see the difference for the Certificate side of things (no known way to produce a failure).

dpgeorge:
> Should I include the fix for `mbedtls_x509_crt_parse()` in this PR or should I open a separate one since there is currently no test I can think of to see the difference for the Certificate side of things (no known way to produce a failure).

Yes, please include the fix in this PR.  Even if we can't find a test that fails here (passing DER with an extra null byte counted in the length), the documentation is quite clear:

> buflen	size of the buffer (including the terminating null byte for PEM data)

So, if it's not PEM data, it shouldn't include a terminating null byte.

peterzuger:
I have addde the fix for `mbedtls_x509_crt_parse()` and added a test using a PKCS#8 key which fails without the fix for `mbedtls_pk_parse_key()`.

extmod/modtls_mbedtls.c
```diff
@@ -100,6 +100,12 @@ static void mbedtls_debug(void *ctx, int level, const char *file, int line, cons
 }
 #endif
 
+#if defined(MBEDTLS_PEM_PARSE_C)
+static int mbedtls_is_pem(const byte *data, size_t len) {
+    return (len >= 10) && (memcmp(data, "-----BEGIN", 10) == 0);
```
projectgus, line 105: > Since all PEM data starts with "-----BEGIN" this is used to check if the data is PEM.

This is frustratingly not quite true, because PEM is an outgrowth of a mail format it can contain preceding plaintext.

Suggest using `strstr()` like mbedTLS does here: https://github.com/Mbed-TLS/mbedtls/blob/cfe8c4c57d7e71ea4763252e76b20fe019040eb8/tf-psa-crypto/drivers/builtin/src/pkparse.c#L1337

tests/extmod/ssl_keycert.py.exp
```diff
@@ -1,3 +1,4 @@
 ValueError('invalid key',)
 TypeError("can't convert 'NoneType' object to str implicitly",)
 ValueError('invalid cert',)
+ValueError('invalid cert',)
```
projectgus, line 4: This test result really confuses me! Most of the ports (all but esp32) don't have MBEDTLS_PARSE_PEM enabled by default, so you'd expect `mbedtls_parse_key()` to fail and the exception to come back `invalid key` not `invalid cert`. Wouldn't you?

From a casual look in the code I can't see why it's accepting PEM here (i.e. on the unix port default config). Do you know?

projectgus (commented):
Thanks for digging into this, @peterzuger!

As damien noted, I ran into the same problem last week after updating ESP-IDF v5.2.2 and dropped in a separate fix here:
https://github.com/micropython/micropython/pull/15523/commits/b180f1fcb1af54505a170f61a3d08645f2ccd1bf

... I should have looked at open PRs first and saved myself some time! :grin: 

I like the idea of increasing the test coverage to include this, although as per my comment I'm not actually sure about what behaviour we should expect on most ports.

tests/extmod/ssl_keycert.py.exp
peterzuger: This is not a fix for PEM, this data passed here is a DER formatted `PKCS#8` key.

See #14371 , this tests DER data in the `PKCS#8` format which without the fix in this or your PR fails because the +1 in the key length passes additional data which mbedtls now checks against (since v3.5). 
But with this it should pass on all ports (I only tested unix and stm32).

projectgus: Ah yes, my mistake that test data is very clearly not in PEM format! Got it, looks good.

extmod/modtls_mbedtls.c
projectgus: It's relatively minor, but I'd say if there's a PEM certificate file that is accepted by desktop mbedTLS and openSSL then it should be accepted by MicroPython as well. Which means allowing preceding text, even if it's not something commonly used.

What do you think about combining my commit and your test addition for this? My version of the fix already uses strstr and it's a little more factored (less duplicate code), and they both do more or less the same thing.

dpgeorge: @peterzuger The change https://github.com/micropython/micropython/commit/b180f1fcb1af54505a170f61a3d08645f2ccd1bf that @projectgus refers to is a little cleaner than the changes here, because it factors the extraction of cert data into a helper function, and only has one location with a conditional check on `MBEDTLS_PEM_PARSE_C`.  So I would prefer that simpler commit.

The test from this PR is very good though, we definitely want that.

What do you think, shall we take @projectgus 's commit and your test?

dpgeorge: I updated this to use `strstr`.

PR #13087, opened 2023-11-29
py/objint: Fix int.to_bytes() buffer size checks.

projectgus:
This "little fix" ended up growing and growing, so this might be too much code size impact. In which case this should be closed. There's a much smaller patch which only fixes the first point from the list below.

* No longer overflows if byte size is 0 (closes #13041)
* Raises OverflowError in any case where number won't fit into byte length. (Now matches CPython, previously MicroPython would return a truncated bytes object.)
* Document that micropython `int.to_bytes()` doesn't implement the optional `signed` kwarg, but will behave as if `signed=True` when the integer is negative (this is the current behaviour). Add tests for this also.

Changes are implemented for small ints, MPZ large ints, and "long long" large ints.

Adds a new set of unit tests for ints between 32 and 64 bits to increase coverage of "long long" large ints, which are otherwise untested.

Tested on unix port (64 bit small ints, MPZ long ints) and Zephyr STM32WB board (32 bit small ints, long long large ints).

Untested on a port whose native format is big endian (don't have one at hand).

*This work was funded through GitHub Sponsors.*

PR #14387, opened 2024-04-28
py/objint: Make byteorder argument optional in byte-conversion methods.

AmirHmZz:
This PR intends to make `byteorder` argument optional in `int.to_bytes()` and `int.from_bytes()` methods. According to [CPython documentation](https://docs.python.org/3/library/stdtypes.html#int.to_bytes), `byteorder` argument is optional and set to "big" by default. This PR makes Micropython compatible with CPython in this case.


projectgus:
Hi @AmirHmZz,

Thanks for sending the PR for this!

>

projectgus:
Hi @AmirHmZz,

Thanks for noticing this and submitting this PR.

> According to [CPython documentation](https://docs.python.org/3/library/stdtypes.html#int.to_bytes), byteorder argument is optional and set to "big" by default. This PR makes Micropython compatible with CPython in this case.

To expand a bit on this, these `byteorder` arguments only became optional in CPython 3.11 ([3.10 version is here](https://docs.python.org/3.10/library/stdtypes.html#int.to_bytes)). MicroPython only advertises full compatibility with [CPython 3.4](https://docs.micropython.org/en/latest/genrst/index.html) at this point.

That said, there are a number of improvements from later CPython versions that are supported in MicroPython and this one might be useful to have. :grin:

I was going to suggest adding unit test cases for these as well, although it will be a little fiddly because the unit tests compare behaviour with CPython by default only some CPython versions will accept the optional argument.

projectgus:
I asked @dpgeorge about this change and it seems like it really will depend on the code size hit as to whether it's desirable to merge. It's currently small, which is encouraging.

However, Damien noted that in Python 3.11 the `int.to_bytes()` function also gained a default value for the `length` argument. What's the size impact if this is added as well? It's desirable to keep the CPython compatibility changes consistent as possible, although if that additional change somehow has much higher code size impact then it can be dropped.

> Where should I add unit tests? I will be glad to contribute!

Thanks, that'd be great!

`tests/basic/int_bytes.py` has the current test cases for `int.to_bytes` and `int.from_bytes`. However, this test file works by comparing output to CPython which won't be consistent for this change. You'll need to do something like `tests/basics/string_format_cp310.py` and `tests/basics/string_format_cp310.exp` which has the test case and expected output in separate files.

Suggest something like `int_bytes_optional_args_cp311.py`.

> making that arg optional is reducing redundant allocations and deallocations on "big" string which is used in most cases. I use int>bytes and bytes>int conversions many many times in my projects and most of them are in the loops.

I would have expected if you're either freezing your .py files into the firmware, or compiling to .mpy with mpy-cross, then the "big" string should end up interned as a QSTR with minimal performance impact. If you've got some "real world" type code where this change makes a significant performance difference, please share it as it'd be interesting to look into.

AmirHmZz:
@projectgus Thanks for feedback. I've added tests and also made length optional in `int.to_bytes()`.

projectgus:
Nice work @AmirHmZz, thanks! Implementation and tests look good to me. It's encouraging this implementation is free or almost free in terms of code size on most of the smaller microcontroller ports.

projectgus (approved):
Looks good to me, I think whether this is merged will come down to @dpgeorge's call about the binary size cost.

projectgus:
One thing about the optional length argument for `int.to_bytes()`, currently MicroPython will truncate to the length if the int doesn't fit (CPython raises an exception.) Which may be a bit counter-intuitive. However, I have a PR open (#13087) which changes the behaviour to match CPython on this.

(Regrettably, that PR adds more than 8 bytes binary size! :laughing:)

AmirHmZz:
I've rebased this PR against `mater`.

projectgus:
This still looks good to me, and the code size impact is pretty small, but we'll need to wait for @dpgeorge to take a look now.

projectgus (approved):

dpgeorge:
Thanks for this, it does look good now: CPython compatibility with minimal code size increase.

Rebased and merged!

PR #10062, opened 2022-11-23
extmod/modussl_mbedtls: Wire in support for DTLS

damz:
## What is this PR trying to solve?

This PR enables support for DTLS (i.e. TLS over datagram transport protocols like UDP). While support for DTLS is absent in cpython, it is worth supporting it in micropython because it is the basis of the ubiquitous [CoAP protocol](https://www.rfc-editor.org/rfc/rfc7252), used in many IOT projects. See #5270

## How does this PR solve the problem?

A new set of "protocols" are added to `SSLContext`:

* `ssl.PROTOCOL_DTLS_CLIENT`
* `ssl.PROTOCOL_DTLS_SERVER`

They act as you expect. If one of these is set, the library assumes that the underlying socket is a datagram-like socket (i.e. UDP or similar).

Implement our own timer callbacks as the out of the box implementation relies on `gettimeofday()`.

TODO: To fully support asyncio for DTLS socket, we will need to return a readable or writable event in `poll(MP_STREAM_POLL, ...)` if `_mbedtls_timing_get_delay(self) >= 1`. This is left for future work so as not to interfere with #9871.

PR #5544, opened 2020-01-16
extmod/modussl_mbedtls: add support for PSK cipher suites (ESP32 only)

tve:
This PR proposes to add support for PSK cipher suites to ussl, specifically for the esp32 but it should be supportable for other platforms that use mbedtls as well (there is nothing esp32-specific in this PR except for enabling the cipher suites in the mbedtls config, which is port-specific).
The PSK cipher suites are awesome because they do away with all the cert and key stuff. Instead each client can get an id and hex-key and use that for bidirectional auth when connecting to the server. Very useful for MQTT, for example.
Docs can be read at: https://micropython-tve.readthedocs.io/en/tls-psk/library/ussl.html

PR #14396, opened 2024-04-30
Adds TLS-PSK support to Python SSL context

aminop1us:
This pull request adds support to the ssl module for TLS-PSK (pre-shared key) authentication. This implementation is code compatible with CPython 3.13 for both client and server side, for example as described here:
https://docs.python.org/3.13/library/ssl.html#ssl.SSLContext.set_psk_server_callback
https://docs.python.org/3.13/library/ssl.html#ssl.SSLContext.set_psk_client_callback

The strong motivation for adding this functionality to MicroPython is that typical TSL using PKI on microcontrollers is heavy and resource intensive. For the kinds of platforms that MicroPython is targeting, many developers will prefer the lighter and simpler PSK alternative, particularly for development and testing. In fact this is exactly the target that TLS-PSK was designed for, so MicroPython would benefit greatly from this support being built-in to the ssl module. Since CPython 3.13 added this as a standard, we tried to be as compliant as possible. We are currently using these TLS-PSK capabilities in several in-house projects, and wanted to share it back in the hope that others might benefit from this, too.

Note: due to the underlying MBEDTLS library not supporting client side callbacks for TLS-PSK, we kept the client-side callback API compatible with CPython, but call the callback immediately and pass the returned identity and key to MBEDTLS when it creates the TLS connection. We tried to keep to CPython compatibility as much as possible for interoperability. We have tested with simple server and client code that runs on both our MicroPython branch as well as CPython 3.13.0a3.

We have tried to follow the MicroPython standard coding conventions, but if there are any requests for changes before it can be merged, please let us know and we will do our best to assist in any changes required.

In summary, this implementation allows the application to set a TLS PSK callback function on SSLContext for server-side connections, so the callback function is called for each handshake. Here is the documentation from the set_psk_server_callback link above:
"The parameter callback is a callable object with the signature: def callback(identity: str | None) -> bytes. The identity parameter is an optional identity sent by the client which can be used to select a corresponding PSK. The return value is a bytes-like object representing the pre-shared key. Return a zero length PSK to reject the connection."


Carglglz:
@aminop1us since this would be a new feature, could you add a test (see `tests/multi_net` for examples) so the code coverage can pass, thanks 👍🏼

aminop1us:
> @aminop1us since this would be a new feature, could you add a test (see `tests/multi_net` for examples) so the code coverage can pass, thanks 👍🏼

thank for your suggestion, I will add a new test case.

aminop1us:
> @aminop1us since this would be a new feature, could you add a test (see `tests/multi_net` for examples) so the code coverage can pass, thanks 👍🏼

I added the test and updated the pull request

aminop1us:
please let me know if there is anything else I can do

dpgeorge:
Thank you for the contribution, this is a good feature that's definitely very useful in embedded contexts.

Please see a prior attempt at this in #5544.

I did not realise that CPython had very recently added support for PSK.  That's great... although we did move to a custom `tls` module so we could implement things like PSK and DTLS (see #10062) without having to wait for CPython do it.

So, since we have our own `tls` module, we can implement PSK in the best (minimal) way for embedded, ie we don't need to follow CPython's API.  Then the `ssl` wrapper code in `micropython-lib` can be updated to support the CPython compatible PSK API (which I think requires calling `ssl_context.set_ciphers("PSK")`).

@aminop1us for your use case, do you find it beneficial for your MicroPython code to use PSK in a CPython compatible way, ie with this callback function?  Or would you be just as happy using a custom MicroPython `tls` PSK API?  Eg maybe instead of a callback function one just passes in a dictionary mapping the identity to the key, like `ssl_context.psk_keys = my_dict`.  In your use case is a dict of identity/key pairs good enough, or do you use the full power of a callback?

dpgeorge:
For reference, this is the PR that added PSK support in CPython: https://github.com/python/cpython/pull/103181

aminop1us:
> Thank you for the contribution, this is a good feature that's definitely very useful in embedded contexts.
> 
> Please see a prior attempt at this in #5544.
> 
> I did not realise that CPython had very recently added support for PSK. That's great... although we did move to a custom `tls` module so we could implement things like PSK and DTLS (see #10062) without having to wait for CPython do it.
> 
> So, since we have our own `tls` module, we can implement PSK in the best (minimal) way for embedded, ie we don't need to follow CPython's API. Then the `ssl` wrapper code in `micropython-lib` can be updated to support the CPython compatible PSK API (which I think requires calling `ssl_context.set_ciphers("PSK")`).
> 
> @aminop1us for your use case, do you find it beneficial for your MicroPython code to use PSK in a CPython compatible way, ie with this callback function? Or would you be just as happy using a custom MicroPython `tls` PSK API? Eg maybe instead of a callback function one just passes in a dictionary mapping the identity to the key, like `ssl_context.psk_keys = my_dict`. In your use case is a dict of identity/key pairs good enough, or do you use the full power of a callback?

As long as it is possible for us to write a wrapper to be CPython compatible, we are fine with that.

keenanjohnson:
Hey @aminop1us and @dpgeorge I was just inquiring what the status here was and if you wanted anyone to jump in to finish this task out? 

Having a PSK would micropython to fully support DTLS which we merged the main work for earlier this year in https://github.com/micropython/micropython/pull/15764.

Our project previously implemented a similiar version of this in the way that @dpgeorge described above I think spread through these commits:

- https://github.com/Ribbit-Network/micropython/commit/8dbcda55930eb447c7cca53899f9626550380f4e
- https://github.com/Ribbit-Network/micropython/commit/f1d2b811be3b36f9e2bfa722c06b09d65d60e9a2
- https://github.com/Ribbit-Network/micropython/commit/88ae0b19238d2b0970d49ecc2b45052d07fea974

If it is useful for someone help complete this, I'm happy to jump in or submit our work in a new PR or whatever is most useful to see this closed out.

Thanks all!

keenanjohnson:
I am assuming the original author does not intend to finish this, so I'll try to pick it up.

keenanjohnson:
I have a draft of the approach suggested by @dpgeorge above here: https://github.com/micropython/micropython/pull/17074.

If that looks in line with your suggested approach, I'm happy to clean up the PR (docs, squash commits) so it's ready for merge.

keenanjohnson:
Hey @dpgeorge ! Just wondering if you had a chance to look at the progress in https://github.com/micropython/micropython/pull/17074 and if that seems in line with the direction you would like to head?

projectgus:
Just to catch up where TLS-PSK is at between this PR and #17074, as it would be great to have this support in MicroPython:

On balance we either need the CPython-style callbacks built-in, or (as @dpgeorge and @aminop1us have commented) we need a design which can implement a CPython compatibility layer on top of whatever we have in MicroPython. This PR has CPython-style callbacks already and looks to be pretty lightweight, so that actually seems pretty good to me.

One thing I'm unsure about: do we also need a version of CPython's `ctx.set_ciphers('PSK')` that calls through to `mbedtls_ssl_conf_ciphersuites()` and limits the accepted ciphers to *only* PSK ones (as implemented in #17074)? It's unclear to me if calling `mbedtls_ssl_conf_psk()` enforces a PSK requirement on both client and server, or if this leaves a gap where a connection using a non-PSK suite could be accepted (therefore breaking the expectation of mutual auth). The mbedTLS `ssl_client2.c` and `ssl_server2.c` examples don't seem to do anything special unless the user also passes a ciphersuite on the command line, but that's example code not a production application. :shrug: 

Related point, it's fantastic this PR has a test (thank you) but probably we also need some basic "adversarial" testing for cases like the above (i.e. that a PSK-enabled client won't connect to a non-PSK server, and vice versa).

As no one is actively working on this I'm going to take it off the next milestone, but TLS-PSK support would be very welcome if someone does want to pick it up and submit a new PR which takes the above into consideration.

dpgeorge:
To add to @projectgus 's comments above, I think it is very useful to allow a callback for the server side (at the very least so we can implement CPython compatible `set_psk_server_callback()` in `ssl.py`).

This PR already implements a server callback, but if we want to do that in a more MicroPython-y way, as mentioned in https://github.com/micropython/micropython/pull/14396#issuecomment-2227687650, we could have an attribute to which you store a dictionary that maps identity string to the required PSK:
```py
ssl_ctx.server_psk_keys = dict
```
Allowing this also allows callbacks to user code: the user could pass in an object that implements `__getitem__` like this:
```py
class PSKDict:
    def __getitem__(self, identity_str):
        # called when mbedTLS needs a PSK for the given identity_str
        return ...

ssl_ctx.server_psk_keys = PSKDict()
```
This approach is relatively minimal, allows the simple case of supplying a fixed key (or keys per client), but also allows a callback if needed.

Note: I haven't tested this idea and maybe there are some caveats that make it unworkable.

dpgeorge:
Closing in favour of #17074.

PR #14402, opened 2024-05-01
rp2/boards/SEEED_XIAO_RP2040: New board definition.

mattytrentini:
A board definition for the [Seeed XIAO RP2040](https://www.seeedstudio.com/XIAO-RP2040-v1-0-p-5026.html).

Pin list:

```bash
>>> from machine import Pin
>>> help(Pin.board)
object <class 'board'> is of type type
  D6 -- Pin(GPIO0, mode=ALT, pull=PULL_DOWN, alt=31)
  D7 -- Pin(GPIO1, mode=ALT, pull=PULL_DOWN, alt=31)
  CSn -- Pin(GPIO1, mode=ALT, pull=PULL_DOWN, alt=31)
  D8 -- Pin(GPIO2, mode=ALT, pull=PULL_DOWN, alt=31)
  D10 -- Pin(GPIO3, mode=ALT, pull=PULL_DOWN, alt=31)
  D9 -- Pin(GPIO4, mode=ALT, pull=PULL_DOWN, alt=31)
  D4 -- Pin(GPIO6, mode=ALT, pull=PULL_DOWN, alt=31)
  D5 -- Pin(GPIO7, mode=ALT, pull=PULL_DOWN, alt=31)
  NEOPIXEL_POWER -- Pin(GPIO11, mode=ALT, pull=PULL_DOWN, alt=31)
  NEOPIXEL -- Pin(GPIO12, mode=ALT, pull=PULL_DOWN, alt=31)
  LED_G -- Pin(GPIO16, mode=ALT, pull=PULL_DOWN, alt=31)
  LED_R -- Pin(GPIO17, mode=ALT, pull=PULL_DOWN, alt=31)
  LED_B -- Pin(GPIO25, mode=ALT, pull=PULL_DOWN, alt=31)
  A0_D0 -- Pin(GPIO26, mode=ALT, pull=PULL_DOWN, alt=31)
  A1_D1 -- Pin(GPIO27, mode=ALT, pull=PULL_DOWN, alt=31)
  A2_D2 -- Pin(GPIO28, mode=ALT, pull=PULL_DOWN, alt=31)
  A3_D3 -- Pin(GPIO29, mode=ALT, pull=PULL_DOWN, alt=31)
```


mattytrentini:
Minor nit: There are 3x LEDs (arranged as RGB closely together) on this board but they're active _low_. To initialise them to off:

```bash
>>> leds = (Pin.board.LED_R, Pin.board.LED_G, Pin.board.LED_B)
>>> for l in leds:
...     l.init(Pin.OUT)
...     l.on()  # Turn LED *off*
```
And to turn them on (purple here):

```bash
>>> leds[0].off()
>>> leds[2].off()
```

Is there a better way to present these LEDs? ie to default them to off but to turn them on with appropriate syntax? Should I just create a module for this board?

ricksorensen:
Curiosity- why is CSn defined in `pins.csv`?  It is defined as GPIO1 (from SEEED pinout) which is also UART RX.  I believe this CSn was thought of for general SPI0 use- but I'm guessing as device SPI will have its CS connected to an arbitrary pin.
I like clearly defining the I2C0 and SPI0 pins.

mattytrentini:
> Curiosity- why is CSn defined in `pins.csv`? 

Just because that's the pin nominated as CS on [Seeed's pinout diagram](https://wiki.seeedstudio.com/XIAO-RP2040/#hardware-overview). You're right that it's largely arbitrary, but it also doesn't _have_ to be used for that purpose. It seems best to define it so users will find it by default if they're looking for a CS pin. 

> It is defined as GPIO1 (from SEEED pinout) which is also UART RX.

Note that it can be used for either purposes - UART(0) is bound to that pin too.

(I did inadvertently add `RX_LED` and `TX_LED` which were copy 'n' paste issues. Removed now!)

dpgeorge:
Sorry @mattytrentini this PR got a bit lost :sweat_smile: 

> Is there a better way to present these LEDs?

You can use `machine.Signal("LED_R", invert=True)`.  That allows on/off to do the right thing.

To make that easier for the user we could either add a `board.py` (frozen) which defines the LEDs as `Signal` objects.  Or add a `machine.LED` class.  The latter (an `LED` class) would need a little thought and also need to be added to all ports (mimxrt already has it).

Regarding pin names: IMO `A0_D0` is a bit confusing and also inconsistent with the other names.  Looking at the pinout, most physical pins have multiple names, eg `A0` and `D0`.  So I think either:
1. just use the D0-D10 names and nothing else (following SEEED_XIAO_RP2350)
2. use D0-D10 and separately add all the aliases like A0, MISO, CS, TX, RX.

robert-hh:
>  Or add a machine.LED class. The latter (an LED class) would need a little thought and also need to be added to all ports 

The LED class had been removed from the SAMD port a while ago in favor of the Signal class. It seems that only the MIMXRT port has it. And there it may be kept for legacy. But do not know how many people use it.

robert-hh:
> Regarding pin names: IMO A0_D0 is a bit confusing 

The XIAO_SAMD21 uses the `A0_D0` scheme, because that's what is printed in the schematics, and indeed it confuses me every time I'm going to use that board. . The board just shows numbers. So I could change that to offering Ax and Dx as options.

pyramids:
> > Regarding pin names: IMO A0_D0 is a bit confusing
> 
> The XIAO_SAMD21 uses the `A0_D0` scheme, because that's what is printed in the schematics, and indeed it confuses me every time I'm going to use that board. . The board just shows numbers. So I could change that to offering Ax and Dx as options.

Just because the XIAO_SAMD21 uses a very unfortunate convention does not mean you cannot instead follow the much better convention of SEEED_XIAO_RP2350 (only have D0, D1, etc., without any A0, ...). Otherwise, you would have to instead use the equally unfortunate convention set forth by SEEED_XIAO_NRF52 and change to the D0_A0 scheme since you could not possibly deviate from that pre-established scheme, either! And then change back to A0_D0, of course...for more concerns along those lines, see https://github.com/orgs/micropython/discussions/18186#discussioncomment-14734274

With that out of the way:

All XIAO boards have (in rather decent, accompanying documentation in a wiki) D0, D1, D2, through D10 as pin labels. IMHO that is totally sufficient (as all that have ADCs also have labels like A0, A1, etc., next to D0, D1, etc., plus labels for other functionality that AFAIK always(?) maps to the same Dx pins on all XIAO boards). So I would suggest to drop aliases like A0, A1, CSn, etc., because a user specifically targetting this board already knows that A4 is also called D4, and a user curious if an unknown-to-him/her XIAO board has A4 (analog) functionality on D4 can readily find out by checking if machine.ADC("D4") does not raise an exception (which I have not verified to happen but if it does not, that would be a bug in machine.ADC).

I would, however, like to see an alias called "LED" because most boards define one and having it (possibly as an alias for LED_B) would probably make many simple MicroPython scripts/programs portable (or else allow them to be written to be portable).

ports/rp2/boards/SEEED_XIAO_RP2040/pins.csv
```diff
@@ -0,0 +1,17 @@
+A0_D0,GPIO26
+A1_D1,GPIO27
+A2_D2,GPIO28
+A3_D3,GPIO29
```
dpgeorge: Suggest separating these lines out to D0, D1, D2, D3 and A0, A1, A2, A3.

ports/rp2/boards/SEEED_XIAO_RP2040/pins.csv
```diff
@@ -0,0 +1,17 @@
...
+D7,GPIO1
+D8,GPIO2
+D9,GPIO4
+D10,GPIO3
+NEOPIXEL_POWER,GPIO11
+NEOPIXEL,GPIO12
+LED_R,GPIO17
+LED_G,GPIO16
+LED_B,GPIO25
+CSn,GPIO1
```
dpgeorge: Suggest removing `CSn`, it's not labelled on the pinout (is it?) so any pin could be used for that.

mattytrentini:
Updated to use suggested naming for pins (`D0`/`A0` etc are now separate and `CSn` is removed). This is now consistent with #18958. 

Also added `LED` as an alias to `LED_B`.

mattytrentini:
See https://github.com/micropython/micropython-media/pull/114 for the relevant image of the board.

dpgeorge (approved):
Thanks for updating, looks good now!  Nice to have consistency with pin naming.

PR #14405, opened 2024-05-01
stm32: fix flashing of final words on H5 and H7 MCUs

dpgeorge:
There are two fixes here for H5 and H7 MCUs regarding minimum flash write size:
- fix flash writing so it doesn't drop the last few words if the length of the write is not a multiple of the hardware flash write size (which is 16 bytes on H5, and 16 or 32 bytes on H7)
- fix mboot packing when it writes back the initial part of the firmware, so it uses the hardware flash write size

### DIFF
--- a/ports/stm32/flash.c
+++ b/ports/stm32/flash.c
@@ -24,6 +24,7 @@
  * THE SOFTWARE.
  */
 
+#include <string.h>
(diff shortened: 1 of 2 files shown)


dpgeorge:
Tested first commit on a H5 board, and second commit on a NUCLEO_WB55.

Issue #14416, opened 2024-05-03
docs: Fix typos in README.md


th4ruka:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) and didn't find any that matched.

### Documentation URL

https://github.com/micropython/micropython/blob/master/README.md

### Description

I have found a few typos in the README.md. Following is an example,

> Select ports have support for ` _thread` module (multithreading), `socket` and `ssl` for networking, and `asyncio`. Note that only a subset of Python 3 functionality is implemented for the data types and modules.

Here, `Select` should be changed to `Selected`. 
I would like to work on these typos and make a contribution to MicroPython.

ma261065:
While I'm sure there are typos in the documentation, you picked a controversial use of English as your example.

Many people now use "select" to imply a carefully chosen subset, such as "The awards were attended by a select group of celebrities". 

One piece of guidance I have seen says:

_Use "select" as an adjective to connote a desirable group of objects. Use "selected" as adjective to connote a neutral (even if purposeful) subsetting of objects from a larger universe._

So while I agree that "selected" would be the better choice here, perhaps we would be better off using more common language such as "**Some** ports have support for  _thread module" and avoid the use of "advertising speak" for technical documentation.

dpgeorge:
Yes, the word "select" as used here is correct English.  But it would be better to simplify the language so it doesn't confuse (eg it could be confused as the `select` module, since the previous sentence was listing module names).

dpgeorge:
Fixed in 206dc09456727415248c8d10f60b0121d391d38c

PR #14417, opened 2024-05-03
Fix typos in README.md 

th4ruka:
This handles the fix addressed in issue #14416. Following changes have been made in README.md,

> Selected ~~Select~~ ports have support for `_thread` module (multithreading), `socket` and `ssl` for networking, and `asyncio`.

* Changed `Select` to `Selected` in the above sentence.

> See the [online documentation](https://docs.micropython.org/) for API reference ~~references~~ and information about using MicroPython and information about how it is implemented.

* Changed `references` to `reference` in the above sentence.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Python 3.5 and some select features from later versions). The following core
 datatypes are provided: `str`(including basic Unicode support), `bytes`,
 `bytearray`, `tuple`, `list`, `dict`, `set`, `frozenset`, `array.array`,
 `collections.namedtuple`, classes and instances. Builtin modules include
-`os`, `sys`, `time`, `re`, and `struct`, etc. Select ports have support for
+`os`, `sys`, `time`, `re`, and `struct`, etc. Some ports have support for
 `_thread` module (multithreading), `socket` and `ssl` for networking, and
 `asyncio`. Note that only a subset of Python 3 functionality is implemented
 for the data types and modules.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,7 @@ Python 3.5 and some select features from later versions). The following core
 datatypes are provided: `str`(including basic Unicode support), `bytes`,
 `bytearray`, `tuple`, `list`, `dict`, `set`, `frozenset`, `array.array`,
 `collections.namedtuple`, classes and instances. Builtin modules include
-`os`, `sys`, `time`, `re`, and `struct`, etc. Select ports have support for
+`os`, `sys`, `time`, `re`, and `struct`, etc. Selected ports have support for
```
dlech, line 22: Maybe more clear to say "Some" or "Certain"?

"Select ports" makes more sense to me than "Selected ports" though.

README.md
```diff
@@ -36,7 +36,7 @@ Getting started
 ---------------
 
 See the [online documentation](https://docs.micropython.org/) for API
```
dlech, line 38:
```suggestion
See the [online documentation](https://docs.micropython.org/) for the API
```

If we are changing to "reference" it seems like we should add "the" here to make it sound correct.

README.md
tharuka-pavith: Hi! Thanks for the feedback. I agree that `Some` or `Certain` could work well here. `Some` would be a better choice.

tharuka-pavith:
I have made following changes according to the feedback given,
* Replaced `Selected` with `Some` in port description.
* Updated the online documentation sentence.
Would appreciate another review. @dlech

dlech (approved):
Looks good to me.

dpgeorge:
Thanks for the contribution.

PR #14419, opened 2024-05-03
runtime/abort: Add support for vm_abort in standard runtime

RealJohnSmith:
Add abort setup code (`nlr_set_abort`) to standard runtime. This makes the standard runtime respond to abort signal without any further modifications.

When aborted, the program exits with 255 exit code, to differentiate from a normal shutdown.

### DIFF
--- a/shared/runtime/pyexec.c
+++ b/shared/runtime/pyexec.c
@@ -45,6 +45,9 @@
 
 pyexec_mode_kind_t pyexec_mode_kind = PYEXEC_MODE_FRIENDLY_REPL;
 int pyexec_system_exit = 0;
+#if MICROPY_ENABLE_VM_ABORT
+int pyexec_abort = 0;
+#endif
 
 #if MICROPY_REPL_INFO
 static bool repl_display_debugging_info = 0;
@@ -80,6 +83,10 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
     nlr_buf_t nlr;
     nlr.ret_val = NULL;
     if (nlr_push(&nlr) == 0) {
+        #if MICROPY_ENABLE_VM_ABORT
+        nlr_set_abort(&nlr);
+        #endif
+
         mp_obj_t module_fun;
         #if MICROPY_MODULE_FROZEN_MPY
         if (exec_flags & EXEC_FLAG_SOURCE_IS_RAW_CODE) {
@@ -143,15 +150,22 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
             mp_hal_stdout_tx_strn("\x04", 1);
         }
 
-        // check for SystemExit
-        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) {
+        #if MICROPY_ENABLE_VM_ABORT
+        if (nlr.ret_val == NULL) { // abort
+            ret = pyexec_abort;
+        } else
+        #endif
+        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) { // system exit
             // at the moment, the value of SystemExit is unused
             ret = pyexec_system_exit;
-        } else {
+        } else { // other exception
             mp_obj_print_exception(&mp_plat_print, MP_OBJ_FROM_PTR(nlr.ret_val));
             ret = 0;
         }
     }
+    #if MICROPY_ENABLE_VM_ABORT
+    nlr_set_abort(NULL);
+    #endif
 
     #if MICROPY_REPL_INFO
     // display debugging info if wanted
--- a/shared/runtime/pyexec.c
+++ b/shared/runtime/pyexec.c
@@ -45,9 +45,12 @@
 
 pyexec_mode_kind_t pyexec_mode_kind = PYEXEC_MODE_FRIENDLY_REPL;
 int pyexec_system_exit = 0;
+int pyexec_unhandled_exception = 0; // kept at 0 for backward compatibility
 #if MICROPY_ENABLE_VM_ABORT
-int pyexec_abort = 0;
+int pyexec_abort = 0; // kept at 0 for backward compatibility
 #endif
+int pyexec_keyboard_interrupt = 0; // kept at 0 for backward compatibility
+
 
 #if MICROPY_REPL_INFO
 static bool repl_display_debugging_info = 0;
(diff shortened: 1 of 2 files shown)


RealJohnSmith:
I see that some checks/ports are failing because the function `nlr_set_abort` is not defined. However, it is guarded by `#ifdef MICROPY_ENABLE_VM_ABORT` . I am not sure what the problem is. Can you please help me how to fix it?

shared/runtime/pyexec.c
```diff
@@ -80,6 +80,10 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
     nlr_buf_t nlr;
     nlr.ret_val = NULL;
     if (nlr_push(&nlr) == 0) {
+        #ifdef MICROPY_ENABLE_VM_ABORT
```
dlech, line 83:
```suggestion
        #if MICROPY_ENABLE_VM_ABORT
```

It is always defined as 0 or 1.

shared/runtime/pyexec.c
```diff
@@ -143,11 +147,15 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
             mp_hal_stdout_tx_strn("\x04", 1);
         }
 
-        // check for SystemExit
-        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) {
+        #ifdef MICROPY_ENABLE_VM_ABORT
```
dlech, line 150:
```suggestion
        #if MICROPY_ENABLE_VM_ABORT
```

shared/runtime/pyexec.c
```diff
@@ -143,11 +147,15 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
...
+        if (nlr.ret_val == NULL) { // abort
+            ret = 255;
```
dlech, line 152: Probably want to define a macro for this value and document it.

Unhanded exceptions still return ret = 0 though, so why is this different?

shared/runtime/pyexec.c
```diff
@@ -143,11 +147,15 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
...
+        } else
+        #endif
+        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) { // system exit
             // at the moment, the value of SystemExit is unused
             ret = pyexec_system_exit;
-        } else {
+        } else { // other exception
             mp_obj_print_exception(&mp_plat_print, MP_OBJ_FROM_PTR(nlr.ret_val));
             ret = 0;
         }
```
dlech, line 161: missing `nlr_set_abort(NULL);` somwhere at the end

shared/runtime/pyexec.c
dlech: Also, it seems like this should be immediately after `} else {`, otherwise we are still using the MicroPython runtime (printing, etc.) and the runtime is not aborted immediately.

shared/runtime/pyexec.c
dlech: And this should probably be followed by a `goto` that jumps to a label just before `MICROPY_BOARD_AFTER_PYTHON_EXEC()` at the end of the function so that no other MicroPython runtime stuff runs.

shared/runtime/pyexec.c
RealJohnSmith: Ok, thank you.

shared/runtime/pyexec.c
RealJohnSmith: It is not aborted immediately, however are those things dependent on user code preventing abort somehow? From what I understand, there are some useful things to do, e.g. flush the stdout before the abort.

shared/runtime/pyexec.c
RealJohnSmith: > Probably want to define a macro for this value and document it.

Ok. Where is a good place to define it?

RealJohnSmith: > Unhanded exceptions still return ret = 0 though, so why is this different?

Is this on purpose? I consider it a bug. In my port/usecase I want the exceptions to throw a non-success exit code and I was planning opening an additional PR later. As well as exposing the exit code from system exit, so `os.exit(1)` in python actually exits with the exit code according to the argument.

RealJohnSmith: > And this should probably be followed by a goto that jumps to a label just before MICROPY_BOARD_AFTER_PYTHON_EXEC() at the end of the function so that no other MicroPython runtime stuff runs.

Is it wrong to run the repl stuff? And the `mp_hal_stdout_tx_strn` function. Those are the only two things that are potentially run

RealJohnSmith: > > Probably want to define a macro for this value and document it.
> 
> Ok. Where is a good place to define it?

At the moment, I implemented it the same way as `extern int pyexec_system_exit;` is set up. Do you think this is appropriate?

dlech: > Where is a good place to define it?

Maybe in `mpconfig.h` right after `MICROPY_ENABLE_VM_ABORT`? Then users could override the default value if needed.

>  I implemented it the same way as extern int pyexec_system_exit; Do you think this is appropriate?

Only if there is an a actual need to know different possible reasons for the abort. But if we do need that info, then it seems like the way to do it.

> Is this on purpose?

Not sure, but if you could explain more about your use case that could help the maintainers decide if this is something we should change or not.


> Is it wrong to run the repl stuff?

Again, I think it depends on the use case. Perhaps this is why Damien left this code out when he implemented the feature. There isn't are obvious "right way" to me.

shared/runtime/pyexec.c
```diff
@@ -143,11 +150,16 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
             mp_hal_stdout_tx_strn("\x04", 1);
         }
 
-        // check for SystemExit
-        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) {
+        #if MICROPY_ENABLE_VM_ABORT
+        if (nlr.ret_val == NULL) { // abort
+            ret = pyexec_abort;
+            nlr_set_abort(NULL);
```
dlech, line 156: I think `nlr_set_abort(NULL);` needs to be called unconditionally after the closing brace matching `if (nlr_push(&nlr) == 0) {`. This is a bit tricky though since things in the exception handler could potentially raise an exception which would cause the `nlr_set_abort(NULL);` to be skipped. On the other hand, if that does happen, we probably end up crashing on an unhandled nlr jump anyway.

RealJohnSmith: Fixed.

shared/runtime/pyexec.c
RealJohnSmith: My use case is this:

I am building a device which is programmable by micro python by the end users. The device/firmware provides an additional "OS" layer - display/UI/menu/settings/file system/... and API for the user program. When the user program is being executed, it is rendered on the display, sending stdout to the display and other info. When the program exits, I want to be able to show a screen "Finished successfully" or "Crashed" (triggered by an unhandled exception or os.exit(non-zero) or "Killed/aborted". The abort is triggered from the device itself/button combination. I need the abort to be instant (as perceived by the user). Which means, there is some time to finish some instructions, flush std out, ...; However no more python instructions should be run (so the program wont be able to "prevent" being killed)

RealJohnSmith: For this, it makes sense to me to return exit codes on the exit. Not just always zero. I am not sure if there is a use case, where you would want to have always zero exit code. If the host fw/vm doesn't care about it, it can just ignore it. Right?

Am I missing something?

dlech: > I am building a device which is programmable by micro python by the end users.

We did something similar with Pybricks. What we ended up doing was just raise `SystemExit` to stop user programs. We only use the abort if we are powering off the device. (If a user manages to somehow always catch the `SystemExit`, they will have to power off the device, but this doesn't really seem to happen that we've notified).

Here is the source if you are interested: https://github.com/pybricks/pybricks-micropython/blob/79a0e010ac36fb64593d5ef5b213b64f1a03985b/bricks/_common/micropython.c#L229

> If the host fw/vm doesn't care about it, it can just ignore it. Right?

Makes sense to me.

RealJohnSmith: I see. Should one be preferred over the other? Or is this approach OK as well?
Is there anything blocking this merge at the moment?

dlech: I think we will have to wait for @dpgeorge to chime in here (sometimes it can take a while before he has time).

RealJohnSmith:
Ok, in the meantime, I have implemented the functionality, which sets exit code according to the SystemExit exception. It is set up such, that it is reverse compatible (the current exit codes won't change for anyone, unless they decide to use this by setting the appropriate exit code values)

RealJohnSmith:
There is a unit test failing, something with thread locking, but I am not sure what the output of the test means or how is it related to my code. Can you please advice/help? Thanks

dlech:
That is a flaky test that fails randomly, so nothing to worry about.

RealJohnSmith:
Ok, thanks

RealJohnSmith:
What is the current status? All is ready and we are waiting for @dpgeorge? Is there something that I can do in the meantime, to prepare or fix something, to speed things up?

RealJohnSmith:
Hi. Is there any update on this? Can we merge this?

dpgeorge:
Sorry this didn't get any further attention.  Will focus on the revised #15392.

PR #14435, opened 2024-05-06
webassembly/objpyproxy: Implement JS iterator protocol for Py iterables.

dpgeorge:
This allows using JavaScript for..of on Python iterables.


ports/webassembly/objpyproxy.js
```diff
@@ -162,10 +162,39 @@ const py_proxy_handler = {
...
+
+        if (prop === Symbol.iterator) {
+            // Get the Python object iterator, and return a JavaScript iterator.
+            const iter_ref = Module.ccall(
+                "proxy_c_to_js_get_iter",
+                "number",
+                ["number"],
+                [target._ref],
+            );
+            return () => ({
```
WebReflection, line 174: nit: this "*screams*" for a generator to me ... or better, that's usually [how iterators are implemented in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator):

```js
return function* () {
    while (true) {
        const value = Module._malloc(3 * 4);
        const valid = Module.ccall(
            "proxy_c_to_js_iternext",
            "number",
            ["number", "pointer"],
            [iter_ref, value],
        );
        if (valid) {
            yield proxy_convert_mp_to_js_obj_jsside_with_free(value);
        }
        else {
            Module._free(value);
            break;
        }
    }
};
```

Maybe it doesn't save much typing but at least it grants that the shape of the object is consistent, that is `{ done: true, value: void 0 }` or `{ done: false, value: ... }`.

dpgeorge: Thanks!  I was hoping you would tell me the "right" way to do this :smile: 

Now updated as suggested.

WebReflection:
virtually approved 👍

please remember to push incremental updates on *npm* for quicker feedback in case something ain't fully right (but to me this looks right) 🙏

PR #14441, opened 2024-05-07
shared/tinyusb: Stall the CDC IN endpoint while disconnecting.

projectgus:
Only when dynamic USB devices are enabled.

The issue here is that when the USB reset triggers, the dynamic USB device reset callback is called from inside the TinyUSB task.

If that callback tries to print something then it'll call through to tud_cdc_write_flush(), but TinyUSB hasn't finished updating state yet to know it's no longer configured. Subsequently it may try to queue a transfer and then the low-level DCD layer panics.

By explicitly stalling the endpoint first, usbd_edpt_claim() will fail and tud_cdc_write_flush() returns immediately. The stall remains until the reset finishes processing.

Tested by running the dfu example from #14439 (no longer crashes), and also re-ran high level USB examples on rp2 and samd.


dpgeorge:
I have tested this with the USB DFU example and it works.

Thanks!

PR #12853, opened 2023-11-01
ports: Add RISC-V 32 bit QEMU port.

agatti:
This adds a QEMU-based bare metal RISC-V 32 bits port.  For the time being only QEMU's "virt" 32 bits board is supported, using the ilp32 ABI and the RV32IM architecture.

Tests run and pass except for two instances I'll debug and fix shortly, and the test generator framework was modified to remove hardcoded references to the "qemu-arm" port.  If you want to test this on your machine I suggest you build your own toolchain from RISC-V's reference toolchain following [these instructions](https://github.com/riscv-collab/riscv-gnu-toolchain#installation-newliblinux-multilib).  In my experience pre-built toolchains' support for 32 bits target is a hit and miss situation, ranging from missing headers to mismatched ABIs and more.

I'm not sure on whether this PR title (and commit message) is ok being prefixed with just "ports", as it touches several places and the list would be too long.  If that's not ok let me know what to change that into to follow the repo's conventions.

This is part of the RISC-V work I'm doing on MicroPython; so instead of coming up with a very large and complicated patch to merge I'm splitting things up in more manageable chunks (the ESP32 Xtensa/RISC-V build changes in #12839 is part of this).

If this gets merged then the changes related to NLR/Native GC and the native code emitter (including compressed instructions support - RISC-V's equivalent of Arm's Thumb-2) can be tested on something more manageable than an ESP32-C3, and can also help for future RISC-V microcontroller ports.  Plus if needed I can help to provide a PR to bring this into the CI as a build target.

The changes in `tinytest.c` are due to the fact that referencing `stdout` without libc being linked sort of confuses the compiler.  The test framework isn't impacted otherwise.

By the way, is there any way to let the CI spellchecker know that certain words are not misspelled?  Right now it complains about a register name in an `__asm volatile` block, for example.

PR #14456, opened 2024-05-09
libm/libm: Do not force floating point type size evaluation.

agatti:
MicroPython's vendored `libm` and `libm_dbl` definition of `FLT_EVAL_METHOD` break compilation on embedded targets that use picolibc as their libc provider (see discussion: https://github.com/micropython/micropython/pull/12853#discussion_r1591815622).

One case where this is needed is the RISC-V toolchain available on the CI image.  Going forward, other toolchains may prefer to use picolibc instead of newlib, and users may want to be able to use said library with MicroPython.

Starting from C99 `FLT_EVAL_METHOD` should be left to the compiler/libc to define, according to the appropriate floating point types' representation for the platform.  Choosing floating point types representations should be done via compiler options rather than with `#define`s.

Tests pass on both `qemu-arm` and `qemu-riscv` with this change applied.  Though I saw no build breakages for ports/boards that used either `libm` or `libm_dbl`, I cannot tell whether this PR breaks code on hardware platforms I have no access to.

### DIFF
--- a/lib/libm/libm.h
+++ b/lib/libm/libm.h
@@ -19,7 +19,10 @@
 #include <stdint.h>
 #include <math.h>
 
-#define FLT_EVAL_METHOD 0
+// These lines verify that FLT_EVAL_METHOD==0, MicroPython's libm requires this.
+// If compilation fails here then check the host compiler's FLT_EVAL_METHOD.
+typedef float float_t;
+typedef double double_t;
 
 #define FORCE_EVAL(x) do {                        \
 	if (sizeof(x) == sizeof(float)) {         \
--- a/lib/libm_dbl/libm.h
+++ b/lib/libm_dbl/libm.h
@@ -15,7 +15,10 @@
 #include <stdint.h>
 #include <math.h>
 
-#define FLT_EVAL_METHOD 0
+// These lines verify that FLT_EVAL_METHOD==0, MicroPython's libm requires this.
+// If compilation fails here then check the host compiler's FLT_EVAL_METHOD.
+typedef float float_t;
+typedef double double_t;
 
 #define FORCE_EVAL(x) do {                        \
 	if (sizeof(x) == sizeof(float)) {         \


projectgus:
> Starting from C99 FLT_EVAL_METHOD should be left to the compiler/libc to define, according to the appropriate floating point types' representation for the platform. Choosing floating point types representations should be done via compiler options rather than with #defines.

As I understand it, the maybe-tricky part here is that MicroPython libm provides a bunch of optimised math routines that rely on a particular binary representation of floats. So if `FLT_EVAL_METHOD != 0` then probably `float_t != float` and possibly `double_t != double` and therefore MicroPython's libm may produce garbage results on that compiler.

To keep a check for this, I think it might be enough to insert this into both libm headers where the define previously was:

```c
// These lines verify that FLT_EVAL_METHOD==0, MicroPython's libm requires this.
// If compilation fails here then check the host compiler's FLT_EVAL_METHOD.
typedef float float_t;
typedef double double_t;
```

Redundant typedefs are allowed, so provided math.h makes matching definitions then this should compile. If a particular platform is configured differently, math.h should have defined float_t or double_t to something else so a compiler error should appear at this point. I think it's preferable to (maybe) compiling and producing garbage results

That said, I don't know if MicroPython supports any build configurations where this would fail.

agatti:
@projectgus, I've looked a bit into it as well.

On RISC-V I believe `FLT_EVAL_METHOD` may change depending on the combination of architecture and abi being passed to GCC via `-march` and `-mabi`.  If my understanding is correct (and please do not quote me :)), if a CPU implements a combination of "D", "Q", or "Zfh"/"Zfhmin"  extensions (double, quad, half precision) but *not* the "F" extension (single precision) then MicroPython's `libm` would not work as either floats or doubles will be aliased to something else.  Same goes for `libm_dbl` on a CPU that only implements "F", "Q", or "Zfh"/"Zfhmin" but not "D", doubles will be aliased to floats.

That alone makes your point valid, I guess.

I haven't looked much into how Arm handles this, but for example what happens if you use `libm_dbl` targetting a Cortex-M4F (whose FPU only supports single-precision floats) passing `-mfloat-abi=hard`?  No idea whether the compiler will still use soft-float code when using doubles or will just alias doubles to floats.

That said, I'll update the patch following your suggestion, thanks!

projectgus:
@agatti These are good questions that I don't know the answers to! At minimum, I think these changes won't make anything not work that previously would have worked...

projectgus (approved):

PR #14462, opened 2024-05-10
shared/tinyusb: Add common cdc tx/rx functions.

andrewleech:
There are a few tinyusb CDC functions used for stdio that are currently replicated across a number of ports. 
Not surprisingly in a couple of cases these have started to diverge slightly, with additional features added to one of them.

This PR consolidates a couple of key shared functions used directly by tinyusb based ports.

### DIFF
--- a/ports/esp32/boards/ARDUINO_NANO_ESP32/mpconfigboard.cmake
+++ b/ports/esp32/boards/ARDUINO_NANO_ESP32/mpconfigboard.cmake
@@ -18,7 +18,7 @@ set(SDKCONFIG_DEFAULTS
 set(MICROPY_SOURCE_BOARD
     ${MICROPY_BOARD_DIR}/board_init.c
     ${MICROPY_BOARD_DIR}/double_tap.c
-    ${MICROPY_DIR}/shared/tinyusb/mp_cdc_common.c
+    ${MICROPY_DIR}/shared/tinyusb/mp_usbd_cdc.c
 )
 
 set(MICROPY_FROZEN_MANIFEST ${MICROPY_BOARD_DIR}/manifest.py)
--- a/ports/mimxrt/Makefile
+++ b/ports/mimxrt/Makefile
@@ -232,6 +232,8 @@ SHARED_SRC_C += \
 	shared/runtime/stdout_helpers.c \
 	shared/runtime/sys_stdio_mphal.c \
 	shared/timeutils/timeutils.c \
+	shared/tinyusb/mp_usbd.c \
+	shared/tinyusb/mp_usbd_cdc.c \
 
 # Set flash driver name, base address and internal flash flag, based on the flash type.
 ifeq ($(MICROPY_HW_FLASH_TYPE),$(filter $(MICROPY_HW_FLASH_TYPE),qspi_nor_flash))
--- a/ports/mimxrt/mpconfigport.h
+++ b/ports/mimxrt/mpconfigport.h
@@ -148,6 +148,7 @@ uint32_t trng_random_u32(void);
 #endif
 
 #define MICROPY_HW_ENABLE_USBDEV            (1)
+#define MICROPY_HW_USB_CDC                  (1)
 
 // Hooks to add builtins
 
--- a/ports/mimxrt/mphalport.c
+++ b/ports/mimxrt/mphalport.c
@@ -30,6 +30,7 @@
 #include "py/mphal.h"
 #include "shared/timeutils/timeutils.h"
 #include "shared/runtime/interrupt_char.h"
+#include "shared/tinyusb/mp_usbd_cdc.h"
 #include "extmod/misc.h"
 #include "ticks.h"
 #include "tusb.h"
@@ -44,51 +45,9 @@
 static uint8_t stdin_ringbuf_array[MICROPY_HW_STDIN_BUFFER_LEN];
 ringbuf_t stdin_ringbuf = {stdin_ringbuf_array, sizeof(stdin_ringbuf_array), 0, 0};
 
-uint8_t cdc_itf_pending; // keep track of cdc interfaces which need attention to poll
-
-void poll_cdc_interfaces(void) {
-    // any CDC interfaces left to poll?
-    if (cdc_itf_pending && ringbuf_free(&stdin_ringbuf)) {
-        for (uint8_t itf = 0; itf < 8; ++itf) {
-            if (cdc_itf_pending & (1 << itf)) {
-                tud_cdc_rx_cb(itf);
-                if (!cdc_itf_pending) {
-                    break;
-                }
-            }
-        }
-    }
-}
-
-
-void tud_cdc_rx_cb(uint8_t itf) {
-    // consume pending USB data immediately to free usb buffer and keep the endpoint from stalling.
-    // in case the ringbuffer is full, mark the CDC interface that need attention later on for polling
-    cdc_itf_pending &= ~(1 << itf);
-    for (uint32_t bytes_avail = tud_cdc_n_available(itf); bytes_avail > 0; --bytes_avail) {
-        if (ringbuf_free(&stdin_ringbuf)) {
-            int data_char = tud_cdc_read_char();
-            if (data_char == mp_interrupt_char) {
-                mp_sched_keyboard_interrupt();
-            } else {
-                ringbuf_put(&stdin_ringbuf, data_char);
-            }
-        } else {
-            cdc_itf_pending |= (1 << itf);
-            return;
-        }
-    }
-}
-
 uintptr_t mp_hal_stdio_poll(uintptr_t poll_flags) {
     uintptr_t ret = 0;
-    poll_cdc_interfaces();
-    if ((poll_flags & MP_STREAM_POLL_RD) && ringbuf_peek(&stdin_ringbuf) != -1) {
-        ret |= MP_STREAM_POLL_RD;
-    }
-    if ((poll_flags & MP_STREAM_POLL_WR) && tud_cdc_connected() && tud_cdc_write_available() > 0) {
-        ret |= MP_STREAM_POLL_WR;
-    }
+    ret |= mp_usbd_cdc_poll_interfaces(poll_flags);
     #if MICROPY_PY_OS_DUPTERM
     ret |= mp_os_dupterm_poll(poll_flags);
     #endif
@@ -97,7 +56,7 @@ uintptr_t mp_hal_stdio_poll(uintptr_t poll_flags) {
 
 int mp_hal_stdin_rx_chr(void) {
     for (;;) {
-        poll_cdc_interfaces();
+        mp_usbd_cdc_poll_interfaces(0);
         int c = ringbuf_get(&stdin_ringbuf);
         if (c != -1) {
             return c;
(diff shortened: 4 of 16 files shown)


shared/tinyusb/mp_cdc_common.h
```diff
@@ -0,0 +1,35 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#ifndef MICROPY_INCLUDED_SHARED_TINYUSB_MP_CDC_COMMON_H
+#define MICROPY_INCLUDED_SHARED_TINYUSB_MP_CDC_COMMON_H
+
+void tud_cdc_poll_interfaces(void);
```
projectgus, line 30: I know you didn't choose these names, but this might be a good time to clean this up. `tud_` is the prefix used by the TinyUSB device API itself.

What about `mp_usbd_cdc_poll_interfaces()`, and `mp_usbd_cdc.h` for the header & source file names?

If that's too verbose then we could go for a prefix of `mp_cdc...` in both places, instead?

projectgus (commented):
This is really useful, thanks @andrewleech! One small change suggestion, but I think we should merge it with or without the additional name change.

shared/tinyusb/mp_cdc_common.h
andrewleech: Hi Angus, thanks that's a good catch - I kind of did choose this name / add the `tud_` myself but you're right, that's a tinyusb header, we should definitely go with a mp header. I think verbose is fine here, it's not going to push out line lengths crazy in any places it's used.

FYI this change is a precurser to the main thing I've been working on in this space - getting the microypthon banner to be buffered at startup / printed on initial connection of mpremote, just like stm port does. This change is in final stages of testing now and should be pushed soon....

projectgus: Ooh, that'll be nice to have!

andrewleech: The file / function names have been updated now in latest push.

andrewleech:
Note: I've only been able to test this on RP2, I don't own samd / mimxrt / renesas-ra boards.

projectgus (approved):

andrewleech:
> > Note: I've only been able to test this on RP2, I don't own samd / mimxrt / renesas-ra boards.
> 
> It may be redundant as I saw the follow-on PR has been tested on samd already, but I've tested this PR on samd and it seems to work as expected. 👍

Thanks I've just updated my previous post to reflect what I've now been able to test! Regardless, I certainly appreciate any additional testing as I'm not very familiar with some of these ports!

FYI thanks to this incredibly helpful tidbit of info, I'm also looking at bringing esp32s2 and s3 into using this same shared tinyusb implementation, as well as fixing the auto-bootloader jump there for S3 which is currently _incredibly_ frustrating!
https://github.com/hathach/tinyusb/pull/2647#issuecomment-2119178403

projectgus:
@andrewleech Oh wow, that's fantastic news! I had anticipated this being a much bigger task, very exciting.

andrewleech:
So now I've tested everything here except the renesas port changes.

Can anyone provide guidance on how to update the auto generated files in a renesas board (in particular the interrupt vector header) to enable USB on a evk board? @TakeoTakahashi2020  @dpgeorge 

Alternatively @iabdalkader could you perhaps test this PR doesn't break anything on a Portenta C33 (I can really afford one myself for testing with).

iabdalkader:
@andrewleech Yes I can test it, perhaps later today. What do I need to test ?

andrewleech:
> What do I need to test ?

Thanks, on this PR there's no intended functional change so it's just a basic check that stdio / repl still seems to work. 

If possible, the bigger change is in the follow on PR https://github.com/micropython/micropython/pull/14485
For that one is checking at startup whether the banner is buffered and shown on initial connection. The description there should show pretty clearly. 

If you know, I'd love some pointers on how to create (or better, modify) the auto generated code in a renesas board so I can test / document it myself too.

andrewleech:
> You'll need to install the e2studio, load the configuration, edit and then generate the code. For this you need the configuration.xml, I think it's this one: configuration.xml.zip

Ok thanks! I previously installed RA Smart Configurator which can open a "project", not sure if that will be the XML. I'll try tomorrow.

 I _really_ think this XML/project file needs to be included in the board folder as the source of the auto gen files.

andrewleech:
> REPL is still functional on the Portenta C33.

That's great thanks!

iabdalkader:
> not sure if that will be the XML. I'll try tomorrow.

There's a cfg.txt file too, those are the only two non-generated files.
[ra_cfg.txt](https://github.com/micropython/micropython/files/15435799/ra_cfg.txt)

andrewleech:
Thanks @iabdalkader that "configuration.xml" file opened just fine in the standalone "RA Smart Configurator" application!
 Looking at how everything for your board was configured in that cleared up a lot for me. 
I was then able to start with a basic new configuration for my ek board and enable the USB pins and interrupts in that, following the USB FS configuration of your project as a guide. 
Running the "generate files" before and after making my USB config showed only changes to pin and vector files, I was then able to use these changes as a reference to modify the existing EK board generated files to enable USB there. I needed to disable a couple of existing interrupts as that chip only supports 32 user interrupts but that's fine for this test.

That got it going for me, with a cut up usb cable wired to the USB pins and a jumper from 3v to vbus to enable it I had cdc working (quicker than wiring resister divider from USB 5v to vbus pin).

More than just replicating your testing on this PR, this more importantly helped me finish off the updates in tinyusb to support my startup buffer feature. 

Thanks again! I'll try to get a more succinct version of these notes into a readme update pr to help the next person. Would you mind me also including your configuration.xml in that pr (in the matching board folder) to reference as an example?

iabdalkader:
> Would you mind me also including your configuration.xml in that pr (in the matching board folder) to reference as an example?

No please go ahead. The configuration should have been there with the board files.

andrewleech:
I found a further consolidation that made sense in moving more of the cdc stdin handling into the `mp_usbd_cdc_poll_interfaces()` function, de-duplicating this code everywhere else.

With that change, I've restested on all platforms covered here and all appear to be working (basic read/write on repl and mpremote mount still works).

I previously had nrf also included here but it was a larger change which grew even larger as I tested and cleaned it up so moved it to its own PR.

With this, I consider this PR complete and ready for final review.

dpgeorge:
Thanks, this looks really good.  It was always on my list to eventually factor these functions, and I'm glad someone else got to it first :)

ricksorensen:
NRF52840 does not compile with these changes on my system. It cannot find the `ringbuf` functions.  Looked at the samd mods, and updated:
```
diff --git a/ports/nrf/mphalport.h b/ports/nrf/mphalport.h
index 7efe05a15..a51f84ef5 100644
--- a/ports/nrf/mphalport.h
+++ b/ports/nrf/mphalport.h
@@ -28,6 +28,7 @@
 #define __NRF52_HAL
 
 #include "py/mpconfig.h"
+#include "py/ringbuf.h"   // RJS
 #include <nrfx.h>
 #include "pin.h"
 #include "nrf_gpio.h"
@@ -42,6 +43,8 @@ typedef enum
 } HAL_StatusTypeDef;
 
 extern const unsigned char mp_hal_status_to_errno_table[4];
+extern int mp_interrupt_char;     //RJS
+extern ringbuf_t stdin_ringbuf;   //RJS
 
 NORETURN void mp_hal_raise(HAL_StatusTypeDef status);
 void mp_hal_set_interrupt_char(int c); // -1 to disable

```
and
```
diff --git a/ports/nrf/mphalport.c b/ports/nrf/mphalport.c
index 06c6ba5cc..682ceff80 100644
--- a/ports/nrf/mphalport.c
+++ b/ports/nrf/mphalport.c
@@ -40,6 +40,12 @@
 #include "nrf_clock.h"
 #endif
 
+// RJS 
+#define MICROPY_HW_STDIN_BUFFER_LEN 128
+#endif
+static uint8_t stdin_ringbuf_array[MICROPY_HW_STDIN_BUFFER_LEN];  //RJS
+ringbuf_t stdin_ringbuf = { stdin_ringbuf_array, sizeof(stdin_ringbuf_array), 0, 0 };   //RJS
 #if !defined(USE_WORKAROUND_FOR_ANOMALY_132) && \
     (defined(NRF52832_XXAA) || defined(NRF52832_XXAB))
 // ANOMALY 132 - LFCLK needs to avoid frame from 66us to 138us after LFCLK stop.
```
My code now compiles (BOARD=SEEED_XIAO_NRF52), and the REPL works with new build.
 
I do not think these are the optimal patches as I did them simply to resolve link/compile errors based on the samd code... and I have little knowledge of `tinyusb`.  I am particularly concerned with the extern definitions.

robert-hh:
Not surprisingly, the `ARDUINO_NANO_33_BLE_SENSE` build shows the same problem. Looking into the code the implementations seems somewhat confusing, because there in Makefile both `drivers/usb/usb_cdc.c` and `lib/tinyusb/src/class/cdc/cdc_device.c` are referred. The former uses properly ringbuffers.

andrewleech:
Sorry for the breakage @ricksorensen and @robert-hh !
My nrf change was initially much larger, so I split them intoa separate pull request. 
I thought I'd left just the minimum needed here for changed filename but must have got it wrong.

If either of you get a chance could you take a look at https://github.com/micropython/micropython/pull/15158 ?

In that follow up PR I remove the nrf specific USB CDC file and consolidate all this functionality. 

But yes it makes sense to get the master build fixed quickly with or without that PR

robert-hh:
> If either of you get a chance could you take a look at https://github.com/micropython/micropython/pull/15158

Will do.

ricksorensen:
For me both boards compile okay with #15158.  (Time on history is CDT (-5)
```
  980  2024/06/01 21:33:05 git clone https://github.com/andrewleech/micropython.git trynrf
  982  2024/06/01 21:33:57 cd trynrf/
  984  2024/06/01 21:34:17 make -C mpy-cross
  985  2024/06/01 21:34:47 cd ports/nrf/
  991  2024/06/01 21:35:56 ./drivers/bluetooth/download_ble_stack.sh 
  993  2024/06/01 21:36:23 make BOARD=SEEED_XIAO_NRF52 submodules
  995  2024/06/01 21:37:05 make BOARD=SEEED_XIAO_NRF52
```
I built yesterday but did not have access to my SEEED_XIAO_NRF52 board until this morning ... the REPL works.  The only difference I see in the nrf port is in the `ports/nrf/Makefile`
```
diff  [thispr]/ports/nrf/Makefile [currupy]/ports/nrf/Makefile 
188c188
< 	tinyusb/mp_cdc_common.c \             # this PR
---
> 	tinyusb/mp_usbd_cdc.c \               # may 31 micropython
```

I also tried  the pull request from this morning (#15189) which also worked.  Same differences as above.

dpgeorge:
The esp32 and nrf build issues introduced by this PR should be fixed by #15189.

andrewleech:
> The Portenta C33 firmware has been broken after 1.23.0

Can you give more details about the failures? Does repl not respond? Or is it related more to the 1200bps bootloader?

Wifi is a tricky one, maybe at least one of the boards using an esp-corprocessor based wifi could be added to the boards @dpgeorge or others involved in release testing has for running multitests on? Though I'm aware that's a laborious process already... hopefully the automated testing discussed in the meetup the other night can progress to help here!

oh, is the failure you're referring to the compile error?
```
../../lib/mbedtls_errors/mp_mbedtls_errors.c:275:13: error: 'MBEDTLS_ERR_PK_BUFFER_TOO_SMALL' undeclared here (not in a function); did you mean 'MBEDTLS_ERR_MPI_BUFFER_TOO_SMALL'?
  275 |         { -(MBEDTLS_ERR_PK_BUFFER_TOO_SMALL), "PK_BUFFER_TOO_SMALL" },
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |             MBEDTLS_ERR_MPI_BUFFER_TOO_SMALL
```
Maybe this is a board that should be in CI, it's arguably the renesas board with the most features currently so covers the widest range of supported libraries for that port really?

iabdalkader:
> oh, is the failure you're referring to the compile error?

No, the firmware builds fine but USB doesn't enumerate. When the firmware is loaded and the board exists DFU the device becomes dead.

andrewleech:
> I think it's already in CI.

You're right, it is. It was a submodules issue on my end with the compile.

USB not enumerating at all? That sounds like either 
* tinyusb isn't being serviced correctly anymore; it's currently done in mpconfigport.h / MICROPY_HW_ENABLE_USBDEV / MICROPY_EVENT_POLL_HOOK which should possibly get moved to an interrupt wrap like most other ports now.
* tinyusb update has broken the HS driver for this port?

I just retested master-ish on the "`EK-RA4W1` with usb cable patched on" I've got , but it's just the HS peripheral which is a somewhat different driver to the HS one you're using iirc.

andrewleech:
@iabdalkader if you can it would be helpful to test https://github.com/micropython/micropython/pull/15550 to see if that changes anything for you?

Also you could try downgrading tinyusb submodule back to 1fdf2907 I think that's the commit it was for the previous release.

iabdalkader:
I did but it's still not working. I also reverted https://github.com/micropython/micropython/commit/d144f0699de10b10ce42f3013f46d46b8fe97c13 (had to comment out SOF enable), and doesn't work either.

iabdalkader:
Reverting this https://github.com/micropython/micropython/commit/2d33071b171e2de4a3e04f48d1f19b5bbe87c663
Fixes it.

```
>>> 
MPY: sync filesystems
MPY: soft reboot
MicroPython v1.24.0-preview.143.ge1fe62f4f.dirty on 2024-07-29; Arduino Portenta C33 with RA6M5
>>> 
```

andrewleech:
@iabdalkader that's really surprising that the shared code broke the USB support. There must be some difference between the FS and HS peripheral that's conflicting with that commit. If there's some way I could get my hands on a portenta I'd be happy to investigate the problem, though I don't know anyone local to me in Melbourne with one I could borrow.

PR #14470, opened 2024-05-13
Fix compilation warning with GCC

srinsoz:


andrewleech:
Hi, your first few commits looked like good cleanups of compiler warnings thanks!
The latest couple however look like they belong on a personal project / personal branch instead perhaps?

andrewleech:
It shouldn't need a new pull request, you could start a new branch for the extra commits, then reset this branch to the commits meant to share.

srinsoz:
That should be right now. Sorry for the extra work. I'm not use to Git.

py/obj.h
```diff
@@ -188,8 +188,8 @@ static inline bool mp_obj_is_small_int(mp_const_obj_t o) {
 #define mp_const_float_pi MP_ROM_PTR((mp_obj_t)(((0x40490fdb & ~3) | 2) + 0x80800000))
 #if MICROPY_PY_MATH_CONSTANTS
 #define mp_const_float_tau MP_ROM_PTR((mp_obj_t)(((0x40c90fdb & ~3) | 2) + 0x80800000))
-#define mp_const_float_inf MP_ROM_PTR((mp_obj_t)(((0x7f800000 & ~3) | 2) + 0x80800000))
-#define mp_const_float_nan MP_ROM_PTR((mp_obj_t)(((0xffc00000 & ~3) | 2) + 0x80800000))
+#define mp_const_float_inf MP_ROM_PTR((mp_obj_t)(((0x7f800000 & ~3) | 2) + 0x80800000ull))
+#define mp_const_float_nan MP_ROM_PTR((mp_obj_t)(((0xffc00000 & ~3) | 2) + 0x80800000ull))
```
dpgeorge, line 192: What exactly is the error/warning that's lead to this change?  And why isn't the change also needed for `tau` , `e` and `pi` constants?

Is appending a `u` enough, or maybe just `ul`?

srinsoz: Actually the warning does not come form gcc but from Microsoft cl compiler.

The exact output is:

````
../../../mpy/micropython/py/modmath.c(376): warning C4307: '+': integral constant overflow
../../../mpy/micropython/py/modmath.c(377): warning C4307: '+': integral constant overflow
````

We have fix only these two line because they are the ones that generate the warnings, but 
It may be a good idea to update all the definitions.

Comment (dpgeorge) [reply py/obj.h:192]: I guess it's because these constants start with `0x7f` and `0xff`, but the others start with `0x40`.

Is it possible to use `u` or `ul` instead of `ull`?

Comment (srinsoz) [reply py/obj.h:192]: Nop. Only `ull` will work. 
Which is the way to specify an unsigned 64 bit integer.
https://gcc.gnu.org/onlinedocs/gcc/Long-Long.html

Comment (projectgus) [reply py/obj.h:192]: All of the expression results fit inside an unsigned 32 bit integer, don't they?

(The problem is, as @dpgeorge hinted, you can't set the high bit of a 32 bit *signed* integer safely. However you shouldn't need to scale up all the way to a `long long` to resolve this.)

Comment (projectgus) [reply py/obj.h:192]: (Deleted previous comment as it looks like mp_const_float_nan is intended to overflow, so maybe promoting to unsigned 64-bit and then truncating with a cast is the best way.)

Comment (dpgeorge) [reply py/obj.h:192]: Thinking about this some more, these additions are intended to overflow (as @projectgus points out), and I don't think the `ull` is needed if the operation is done differently.

Eg, does this work:
```c
#define MAKE_FLOAT_OBJ(f) MP_ROM_PTR((mp_obj_t)((((f) & ~3) | 2) + 0x80800000) & 0xffffffff)
#define mp_const_float_e MAKE_FLOAT_OBJ(0x402df854)
...
```
?

dpgeorge: I think we should use `UINT16_C(...)` here.

projectgus: ... `UINT64_C`, yes?

dpgeorge: Of course!  My typo.

dpgeorge: We should start using `UINTxx_C(...)` macros everywhere, but for now let's leave this as it currently stands in this PR.

dpgeorge (approved):

PR #14477, opened 2024-05-14
examples/usb: Add README that points out the alternative usb modules.

projectgus:
I didn't get a chance to review #14439, but one thing I think is important is to provide a link to the micropython-lib usb modules from here.

Otherwise if someone starts by looking in this directory then they won't know that the alternative exists.

(In general there are probably also opportunities to mention this example in the docs or micropython-lib, but I feel like this is the most important link as it's the most likely to turn a new user away if they're looking for something more high-level.)


dpgeorge:
Thanks!

PR #14483, opened 2024-05-14
webassembly: Revert conversion of Python None to JavaScript null, and make dict proxy return undefined when key doesn't exist

dpgeorge:
There are 3 related commits in this PR:
- Revert back to converting Python None to JavaScript null.  This essentially reverts fa23e4b093f81f03a24187c7ea0c928a9b4a661b which made `None` convert to `undefined`.  So now it's back to how it was before that commit, and Py `None` converts to JS `null`.  That's consistent with how the `json` module converts these values.
- Create a special "undefined" type/instance for Python visibility.  This corresponds exactly to JS `undefined`.  It's accessible via `js.undefined`.  Passing this from Py to JS, JS will receive `undefined`.  And when JS passes `undefined` to Py, Py will see the proxy `js.undefined` value.
- Make proxy Py dicts return `undefined` to JS when JS looks up a key that doesn't exist (instead of raising `KeyError`).  These semantics match better the JS behaviour of `{}` objects.

The first two changes mean the following:
- There's a one-to-one correspondence between Py `None` and JS `null`.
- There's a one-to-one correspondence between Py `js.undefined` and JS `undefined`.
- Conversion of these values matches the `json` module.

All three changes together mean that, given the following JavaScript function with an optional argument:
```js
function withDefault({ value = 'OK' } = {}) {
  console.log(value);
}
```
the JavaScript and Python code to call this function have equivalent behaviour, namely, in JavaScript:
```js
withDefault();
withDefault({});
withDefault({ value: null });
withDefault({ value: undefined });
```
and in Python:
```py
js.withDefault()
js.withDefault({})
js.withDefault({"value": None})
js.withDefault({"value": js.undefined})
```

### DIFF
--- a/ports/webassembly/proxy_c.c
+++ b/ports/webassembly/proxy_c.c
@@ -59,6 +59,16 @@ enum {
     PROXY_KIND_JS_PYPROXY = 7,
 };
 
+MP_DEFINE_CONST_OBJ_TYPE(
+    mp_type_undefined,
+    MP_QSTR_undefined,
+    MP_TYPE_FLAG_NONE
+    );
+
+static const mp_obj_base_t mp_const_undefined_obj = {&mp_type_undefined};
+
+#define mp_const_undefined (MP_OBJ_FROM_PTR(&mp_const_undefined_obj))
+
 MP_DEFINE_EXCEPTION(JsException, Exception)
 
 void proxy_c_init(void) {
@@ -80,7 +90,7 @@ static inline mp_obj_t proxy_c_get_obj(uint32_t c_ref) {
 
 mp_obj_t proxy_convert_js_to_mp_obj_cside(uint32_t *value) {
     if (value[0] == PROXY_KIND_JS_UNDEFINED) {
-        return mp_const_none;
+        return mp_const_undefined;
     } else if (value[0] == PROXY_KIND_JS_NULL) {
         return mp_const_none;
     } else if (value[0] == PROXY_KIND_JS_BOOLEAN) {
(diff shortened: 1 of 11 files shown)


dpgeorge:
@WebReflection please take a look at this PR and see what you think.  It's a bit different to what we discussed on Discord, the difference being that `undefined` is not converted to Py `None`, but rather to a distinct object `js.undefined`.  That means that JS functions that return their default value of `undefined` will pass that value through to Python.  But, that has very little impact (no tests we have are impacted by that), and doing it this way keeps a proper separation between `undefined` and `null` on both sides (Py and JS sides).  So IMO it feels nice and consistent, a one-to-one correspondence between `null` and `None`.

ports/webassembly/proxy_c.c
```diff
@@ -228,7 +241,15 @@ void proxy_c_to_js_lookup_attr(uint32_t c_ref, const char *attr_in, uint32_t *ou
         qstr attr = qstr_from_str(attr_in);
         mp_obj_t member;
         if (mp_obj_is_dict_or_ordereddict(obj)) {
-            member = mp_obj_dict_get(obj, MP_OBJ_NEW_QSTR(attr));
+            // Lookup the requested attribute as a dict key (not a method),
```
WebReflection, line 244: that `not as method` kinda bothers me ... `function some({ thing = () => {} } = {}) {}` is a valid *JS* expectation, if this is covered in here I am OK with that comment (if explained somewhere else), otherwise I wonder "*why*"

WebReflection:
it's true that methods (or utilities?) are rare, but not that much neither for JS related APIs ... I've left a comment that, once resolved or explained, would make me virtually approve this PR as I think it's awesome in both intent and expectations 👍

ports/webassembly/proxy_c.c
dpgeorge: I think it's just that the comment I wrote here is confusing.  I meant that here the code is looking up the attribute as a key within the dict (or ordereddict).  That's opposed to the standard case of looking up the attribute as the name of a member/function/method of the target object.

Eg looking up `attr` on `target` will have the following logic:
- If `target` is a `dict` or `ordereddict` instance then lookup `target[attr]`, and return the result.  If the lookup failed, return `js.undefined`.
- Otherwise, lookup `target.foo`, raising an exception if it can't be found.

I've updated the comments to try and make it clearer.

I also added a test for passing through a function in the dict.

WebReflection: Awesome, :+1: for me!

PR #14485, opened 2024-05-15
shared/tinyusb: Buffer startup stdout (banner) to send once usb cdc is connected.

andrewleech:
At startup, buffer stdout / microypthon banner so that it can be printed on initial connection of mpremote, just like stm port does.

Requires: https://github.com/hathach/tinyusb/pull/2629 (merged) and https://github.com/hathach/tinyusb/pull/2647 (merged)

Extra details: https://github.com/hathach/tinyusb/discussions/2595

Sitting on top of https://github.com/micropython/micropython/pull/14462

# functional change
This change is most obvious when you've first plugged in a micropython device (or hit reset) when it's a board that uses USB (CDC) serial in the chip itself for the repl interface. This doesn't apply to UART going via a separate usb-serial chip.

On all ports other than stm32 currently, nothing happens when you first run a terminal like mpremote on a freshly booted board. 
A user might typically first hit enter to look for a repl prompt `>>>` to ensure they are connected to a micropython board, then maybe hit ctrl-d to soft reboot and see the banner to check they're connected to the right board. 
Code_dZZf1qJ3Un

On stm however, on a freshly started board when you connect mpremote the banner is immediately shown!.
Code_Wsv7praNrC

This PR extends this behaviour to rp2, mimxrt, samd and renesas. It's also extended to esp32s2 / s3 in the follow up PR.


andrewleech:
```
anl@STEP: ~ $ mpremote a1
Connected to MicroPython at /dev/ttyACM1
MicroPython v1.23.0-preview.377.g754ae8b76d.dirty on 2024-05-15; Raspberry Pi Pico with RP2040
>>>
```

andrewleech:
```
anl@STEP: ~ $ mpremote
MicroPython v1.23.0-preview.380.g7ab4754386.dirty on 2024-05-16; Seeed Xiao with SAMD21G18A
>>> 
```

andrewleech:
```
anl@STEP: ~ $ mpremote
MicroPython v1.23.0-preview.378.g9958203b3d.dirty on 2024-05-15; i.MX RT1010 EVK with MIMXRT1011DAE5A
>>> 
```

andrewleech:
```
anl@STEP: ~ $ mpremote
MicroPython v1.23.0-preview.381.gd3a95507b6.dirty on 2024-05-16; PCA10059 with NRF52840
>>>
```

andrewleech:
@mattytrentini 
```
anl@STEP: ~ $ mpremote
MicroPython v1.23.0-preview.378.g526af512b4.dirty on 2024-05-16; ItsyBitsy M4 Express with SAMD51G19A
>>>
```

shared/tinyusb/mp_usbd_cdc.c
```diff
@@ -83,40 +85,50 @@ void tud_cdc_rx_cb(uint8_t itf) {
...
-            uint32_t n2 = tud_cdc_write(str + i, n);
-            tud_cdc_write_flush();
-            i += n2;
+            // Explicitly run the USB stack as the scheduler may be locked (eg we
+            // are in an interrupt handler), while there is data pending.
+            mp_usbd_task();
+        }
+        n = MIN(n, tud_cdc_write_available());
+        if (n == 0) {
+            break;
```
projectgus, line 106: IIUC, if disconnected then after the tinyusb FIFO is full then further output is dropped until USB-CDC connects.

Could this lead to some misleading situations where USB host connects and receives:

> banner
> early program output
> [unknown number of dropped bytes here]
> contemporary output that was sent after USB host connected

Is that right? Does it seem like a problem?

The best solution I can see from this perspective would be to rotate bytes out of the overflowing FIFO as needed, but (a) this is quite slow if USB never connects and (b) not supported by current TinyUSB API.

However there is `tud_cdc_n_write_clear()` which empties the write FIFO. What do you think about doing this each time the FIFO fills up while disconnected? It means maybe some buffered bytes are lost, but the "new after old with missing middle" situation shown above can't happen.

(Note this isn't a request necessarily, I'm not fully convinced this is worth worrying about.)

Related request I do have is that it'd be good to have a short comment or so in here please, outlining the intended execution flow for this "disconnected, buffering" case as it may not be obvious to future readers why it's structured like this.

shared/tinyusb/mp_usbd_cdc.c
```diff
@@ -125,6 +137,13 @@ mp_usbd_line_state_cb
 tud_cdc_line_state_cb
 #endif
     (uint8_t itf, bool dtr, bool rts) {
+    if (dtr) {
+        // The host has started to open the cdc serial port.
+        // Wait a few ms for host to be ready then send tx buffer.
+        cdc_connected_flush_delay = (tud_speed_get() == TUSB_SPEED_HIGH) ? 64 : 8;
+        tud_sof_cb_enable(true);
```
projectgus, line 144: Kind of surprised you need to do this, the write_flush() call only prepares the IN endpoint for the host to request it and I'd have thought that the host driver would be prepared to cleanly receive bytes sent after this point. What behaviour do you see if you flush from here, the data goes missing?

EDIT: Oh, I see stm32/usbd_cdc_interface.c does exactly the same thing and has a comment `The actual connection state is delayed to give the host a chance to configure its serial port (in most cases to disable local echo)`. I see in the case of stm32 it delays marking the port as connected at all, so maybe this is for some other purpose that doesn't apply here? As I don't see how local echo matters for data going into the host...?

projectgus (commented):
This will be a really nice addition, thanks @andrewleech :grin:

shared/tinyusb/mp_usbd_cdc.c
projectgus: Oh right, maybe it's actually this:
https://github.com/pyserial/pyserial/blob/master/serial/serialposix.py#L351

pyserial default behaviour is to discord the contents of the input buffer immediately on open. :bomb:

andrewleech: Yes you've found the pieces of the puzzle - I also found that other non-python applications also often have a similar flush/discard at the end of the connection process, as might be seen in the ramblings on https://github.com/hathach/tinyusb/discussions/2595

So yeah it's simply a delay to give the PC application time to finish doing it's port open and clear the buffers before we send our carefully stored data.

shared/tinyusb/mp_usbd_cdc.c
andrewleech: Yeah you raise a good question. My thoughts were that keeping the banner (and maybe a little after it) is a good thing to display on connection. 
But yes it could be somewhat confusing if there's a running application with more information, the disconnect in log could seem quite random. I'm not too sure how STM handles this currently, whether it's the initial data persisted or the latest data. @dpgeorge do you have a preference here?

dpgeorge: In stm32 it's a rolling ring buffer, so new data is always placed into the pending CDC output buffer, and if the ring buffer is full then the oldest data is lost.  That means on connection to a USB host, the host sees a consistent history of the most recent data, possibly truncated at the beginning of that data.

andrewleech: Thanks @dpgeorge, the consistent view of "latest data" does make sense really. I'll have a closer look to see if I can make it automatically overwrite, I actually think this might be possible. 

If not, the suggestion of "clear on full" might be worth doing instead.

andrewleech:
Note this PR currently has the TinyUSB submodule pinned to my PR with required SOF updates, so it won't currently build easily.

This upstream PR will need to be merged first, it's blocked by the need to test on renesas.

projectgus:
> Note this PR currently has the TinyUSB submodule pinned to my PR with required SOF updates, so it won't currently build easily.

FWIW I think we could emulate something similar by adding a callback in the mp_usbd_task, or maybe even soft timer. However I think better to stick with the approach you've already used, if it's going to be available soon.

shared/tinyusb/mp_usbd_cdc.c
andrewleech: Hah, this is convenient: https://github.com/hathach/tinyusb/blob/1f259b3ab0f1b9587dee9c0bfb5574d70d021640/src/class/cdc/cdc_device.c#L250

So yeah if I basically remove the space in buffer check when not connected, it'll pretty much "just work". Though I'll need to confirm behavior when you try to write more than a buffer's worth in one go, whether it keeps the start or end of the data.

andrewleech:
Yes I'd originally thought of using a soft timer however it does have some overheads, and I'm not sure if all needed ports have it enabled. 

I think the upstream change should be achievable, I've had very productive discussions on it already. 

I'm just blocked by renesas port testing, I can't figure out the auto generated code being used in the board profiles there enough to enable USB on the evk board I've loaned.

andrewleech:
I've now got a renesas board working finally and have finished the implementation there, it's all working too.

I just need to clean up some of the commented code in the nrf stuff and it's all good to go!

shared/tinyusb/mp_usbd_cdc.c
projectgus: That is certainly convenient!

It may actually be impossible to debug this with mpremote, given the pyserial hard-coded flush of the input buffer on open. :facepalm: But it should be visible with other serial programs, I'd expect.

andrewleech: I made the change and it seemed to work fine! Even with mpremote.... I added a main.py on the samd51 board:  
`print(list(range(512)))`

Now when I start mpremote I see:
```
 $ mpremote
 486, 487, 488, 489, 490, 491, 492, 493, 494, 495, 496, 497, 498, 499, 500, 501, 502, 503, 504, 505, 506, 507, 508, 509, 510, 511]
MicroPython v1.23.0-preview.379.g130e763226.dirty on 2024-05-25; ItsyBitsy M4 ExpreType "help()" for more information.
>>>
```

andrewleech:
This PR has tested working on `rp2`, `mimxrt`, `samd` and `renesas`. 

`esp32` support is in follow-up PR https://github.com/micropython/micropython/pull/15108.
`nrf` support is in follow-up PR https://github.com/micropython/micropython/pull/15158.

The required updates to TinyUSB have been merged upstream and the submodule here is now pinned to the matching master commit.

As such I consider this PR complete and ready for review.

andrewleech:
Branch has been rebased and re-tested on:
* renesas EK-RA4W1 (build modded to enable usb port)
* samd ADAFRUIT_ITSYBITSY_M4_EXPRESS
* rp2 RPI_PICO_W
* mimxrt MIMXRT1010_EVK

shared/tinyusb/mp_usbd_cdc.c
```diff
@@ -58,8 +58,10 @@ uintptr_t mp_usbd_cdc_poll_interfaces(uintptr_t poll_flags) {
     if ((poll_flags & MP_STREAM_POLL_RD) && ringbuf_peek(&stdin_ringbuf) != -1) {
         ret |= MP_STREAM_POLL_RD;
     }
-    if ((poll_flags & MP_STREAM_POLL_WR) && tud_cdc_connected() && tud_cdc_write_available() > 0) {
-        // When connected operate as blocking, only allow if space is available.
+    if ((poll_flags & MP_STREAM_POLL_WR) &&
+        (!tud_cdc_connected() || (tud_cdc_connected() && tud_cdc_write_available() > 0))) {
+        // Always allow write when not connected, fifo will retain latest.
+        // When connected operate as blocking, only allow if space is available
```
dpgeorge, line 64: Full stop at end of sentence for consistency.

shared/tinyusb/mp_usbd_cdc.c
```diff
@@ -93,42 +95,59 @@ void tud_cdc_rx_cb(uint8_t itf) {
...
+        i += n2;
     }
     return i;
 }
 
+static int8_t cdc_connected_flush_delay = 0;
+
+void tud_sof_cb(uint32_t frame_count) {
+    if (--cdc_connected_flush_delay < 0) {
+        // Finished on-connection delat, disable SOF interrupt again.
```
dpgeorge, line 133: delat -> delay

dpgeorge (commented):
I see tinyusb has two extra commits compared to this PR's updated submodule.  Is it worth pulling in those two extra commits?  They seem to fix an esp32 include.

andrewleech:
> They seem to fix an esp32 include.

That's in espressif's custom driver, which is not the one used in https://github.com/micropython/micropython/pull/15108
There's no harm in bringing it in as well though.

shared/tinyusb/mp_usbd_cdc.c
dpgeorge: This behaviour is perfect!

dpgeorge:
Now merged!  This is a very nice quality-of-life improvement, thanks for the effort getting it all worked out!

(I made some minor changes to this PR before merge, to get the ARDUINO_NANO_33_BLE_SENSE board building, same issue as last time.)

andrewleech:
Ah yes, that's a good point thanks! The `ARDUINO_NANO_33_BLE_SENSE ` has been fixed "properly" in https://github.com/micropython/micropython/pull/15158

PR #14493, opened 2024-05-16
py/mpz: Ignore calls to to_bytes with zero length.

agatti:
Requesting a zero-sized bytes representation of large integers (using mpz rather than `long long`) would attempt to sign-extend a zero-bytes buffer with no bounds checking.  This PR sidesteps the whole issue by not performing any operation if the requested length is zero, with the same runtime behaviour as smaller integers in the same situation.

For the record, code like `int.from_bytes(bytes(range(20)), "big").to_bytes(0, "big")` would crash MicroPython with:

```
#0  __memset_avx2_unaligned_erms ()
    at ../sysdeps/x86_64/multiarch/memset-vec-unaligned-erms.S:330
#1  0x000055555557a9c9 in mpz_as_bytes (z=<optimized out>, 
    big_endian=big_endian@entry=true, len=<optimized out>, len@entry=0, 
    buf=<optimized out>, buf@entry=0x7ffff7a06fe0 "") at ../../py/mpz.c:1631
#2  0x0000555555598eec in mp_obj_int_to_bytes_impl (self_in=<optimized out>, 
    big_endian=big_endian@entry=true, len=len@entry=0, 
    buf=buf@entry=0x7ffff7a06fe0 "") at ../../py/objint_mpz.c:118
#3  0x0000555555598525 in int_to_bytes (n_args=<optimized out>, 
    args=0x7fffffffab78) at ../../py/objint.c:440
[...]
```

### DIFF
--- a/py/mpz.c
+++ b/py/mpz.c
@@ -1590,6 +1590,11 @@ bool mpz_as_uint_checked(const mpz_t *i, mp_uint_t *value) {
 }
 
 void mpz_as_bytes(const mpz_t *z, bool big_endian, size_t len, byte *buf) {
+    if (len == 0) {
+        // No output is necessary.
+        return;
+    }
+
     byte *b = buf;
     if (big_endian) {
         b += len;


agatti:
Is `thread/thread_gc1` a flaky test?  The function I modified shouldn't really impact on the garbage collector's operation (it doesn't allocate any memory at all).

dpgeorge:
> Is `thread/thread_gc1` a flaky test?

Yes, it's flaky.  I have rerun it.

agatti:
Sorry, this looks like it is already part of #13087 - I was looking for pending PRs that made changes to py/mpz and I missed that one :(

projectgus:
@agatti Your patch is a **_lot_** smaller, so if the bigger one in 13087 turns out too big then it might be a better pick. :laughing:

PR #15094, opened 2024-05-22
py: Add RISC-V RV32I NLR.

agatti:
This PR adds native NLR support for 32 bits RISC-V RV32I targets.  When setting `MICROPY_NLR_SETJMP` to 0, tests passed on `qemu-riscv` and on an ESP32-C3 board I get the same test failures I'd get when setjmp/longjmp implementation is used instead.

### DIFF
--- a/py/nlrrv32.c
+++ b/py/nlrrv32.c
@@ -1,0 +1,81 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2024 Alessandro Gatti
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


py/nlrrv32.c
```diff
@@ -0,0 +1,81 @@
...
+        "lw  x24, 44(x10)  \n" // Retrieve S8.
+        "lw  x25, 48(x10)  \n" // Retrieve S9.
+        "lw  x26, 52(x10)  \n" // Retrieve S10.
+        "lw  x27, 56(x10)  \n" // Retrieve S11.
+        "lw   x2, 60(x10)  \n" // Retrieve SP.
+        "addi x10, x0, 1   \n" // Return 1 for a non-local return.
+        "jalr  x0, x1, 0   \n" // Return.
+        :                      // Outputs.
+        : "r" (top)            // Inputs.
+        :                      // Clobbered.
```
dpgeorge, line 75: This line will need to have `"memory"` added to it for gcc 14 to work.

agatti: PR updated, thank you.

PR #15096, opened 2024-05-22
shared/runtime/gchelper: Add RISC-V RV32I native gchelper.

agatti:
This PR adds native gchelper support for 32 bits RISC-V RV32I targets.

Tests pass on `qemu-riscv` when disabling the setjmp implementation, for both native assembler and C fallback versions.  On an ESP32-C3 board I see no differences in test failures count between the three versions (setjmp, native, and fallback).

### DIFF
--- a/shared/runtime/gchelper_generic.c
+++ b/shared/runtime/gchelper_generic.c
@@ -150,6 +150,37 @@ static void gc_helper_get_regs(gc_helper_regs_t arr) {
     arr[10] = x29;
 }
 
+#elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 32)
+
+// Fallback implementation for RV32I, prefer gchelper_rv32i.s
+
+static void gc_helper_get_regs(gc_helper_regs_t arr) {
+    register long s0 asm ("x8");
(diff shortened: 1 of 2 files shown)


shared/runtime/gchelper_generic.c
```diff
@@ -150,6 +150,37 @@ static void gc_helper_get_regs(gc_helper_regs_t arr) {
     arr[10] = x29;
 }
 
+#elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 32)
+
+// Fallback implementation, prefer gchelper_rv32.s
```
dpgeorge, line 155: Maybe a comment about the arch: "Fallback implementation for RV32I, prefer ..."

Also, should the assembler file have the "i" in it as well, `gchelper_rv32i.s`?  In case one day we need to support RV32E.

agatti: Oh, I copied the comment from the thumb section - I'll fix that.  Regarding the ISA, as far as I remember you can have ifdef blocks in GAS files (at least when built using `gcc`).  If that's a concern I can rename the file, that's not a problem at all.

PR #15097, opened 2024-05-22
shared/runtime/semihosting: Add RISC-V semihosting code.

agatti:
This PR contains a full-featured RISC-V semihosting interface implementation.

Even though it targets RISC-V, the specifications say they follow Arm's semihosting call interface and thus this bit of code can be easily modified to also support Arm/Thumb and Aarch64.  To keep breakages at a minimum to users of the Arm-specific semihosting code I chose to put this in a different file instead, plus there are no guarantees they

If needed I can make the necessary modifications to have one single RV32/RV64/Aarch32/Aarch64 semihosting implementation, although there are no guarantees they won't diverge in the future (the RISC-V specifications are at version 0.3, after all...)

### DIFF
--- a/shared/runtime/semihosting_arm.c
+++ b/shared/runtime/semihosting_arm.c
@@ -24,7 +24,7 @@
  * THE SOFTWARE.
  */
 
-#include "semihosting.h"
+#include "semihosting_arm.h"
 
 // Resources:
 // http://embed.rs/articles/2016/semi-hosting-rust/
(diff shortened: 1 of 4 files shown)


dpgeorge:
Did you implement all of this code yourself, from scratch?  (We need to be careful with licensing/copyright.)

And did you make use of this (or test it) with your qemu-riscv work #12853?

agatti:
Yes I did write it myself.  If you prefer a more authoritative source I guess you can take Arm's implementation and simply replace the entry call with the one provided by the RISC-V specification.  This specification is actually quite simple to implement.

For testing, I used it quite a bit when debugging my code emitter (heap dumping and file I/O), although I haven't tested more exotic calls like get command line, elapsed time, invoke system command, etc.  Fun fact, QEMU switches stack start and stack limit pointers in the structure when calling get heap info.

Haven't tested it on OpenOCD, but I guess its implementation should be at least as complete as QEMU's.

dpgeorge:
OK, thanks for the clarifications.  I'm happy to merge this, since it's useful for debugging.

PR #8932, opened 2022-07-20
py/_thread: Add support for lock.acquire timeout.

andrewleech:
This working on https://github.com/micropython/micropython-lib/pull/503 I ran into the limitation that while the `_thread.Lock.acquire()` function supports the `waitflag` argument, it doesn't handle the `timeout` arg - while you can pass it it's silently ignored and instead blocks forever.

This PR adds support for the `timeout` arg. It does so in a fairly naive way I feel... but it does work.
1. try to get lock (unblocking)
2. if success, return True
3. if fail, sleep 1ms
4. if timeout expired, return False
5. else go back to 1.

Ref: https://docs.python.org/3/library/_thread.html#thread.lock.acquire

Similarly the test is not exactly best practice, what with having a static sleep in the middle, but it does test the behaviour.

PR #5599, opened 2020-02-01
Timed lock aquisition

mcdeoliveira:
This PR handles the timeout parameter in `thread_lock_acquire`.

It provides a set of compilation flags, tests, and an implementation for the unix and the esp32 ports. Other ports based on freertos should be easy to implement based on the esp32 port, because freertos lock_acquire already handles the timeout parameter. Unfortunately I do not have access to other hardware, so I think it is better to leave to others the task of replicating the functionality on other ports.

Let me just add a word about the relevance of handling the timeout parameter in lock_acquire. The only feature used by the current CPython implementation of threading.py that is nor yet supported by _thread is precisely a call to lock_acquire with a timeout parameter. In fact, a _thread_lock_acquire with timeout is issued in RLock and all subsequent classes in threading.py are based on RLock! Implementing the timeout parameter essentially enables a micropython version of Threading.

PR #15099, opened 2024-05-22
py/_thread: Add support for lock.acquire timeout.

### LINKED ISSUE #3332: timeout parameter is not implemented in _thread.allocate_lock()
https://github.com/micropython/micropython/blob/62849b7010abffb7b0a9c9875930efe7cb77519c/py/modthread.c#L69

---

Background:

I'm looking for a way to implement a cancelable timeout on Linux, e.g something like threading.Timer from the standard library.

Being on Linux, I'm thinking of trying a timerfd with uasyncio instead. But if implementing the timeout parameter here is not too hard, I could give it a try. It looks to be platform dependent though, so it might be beyond my capabilities.

DvdGiessen:
Fixes #3332. Based on previous work in #5599 and #8932.

Differences to previous PR's:

- Uses a per-port resolution
  - This prevents us from always converting to microseconds and back on ports that don't have that level of precision, no longer limiting the maximum timeout.
  - Also used to implement `_thread.TIMEOUT_MAX` (see [CPython docs](https://docs.python.org/3/library/_thread.html#thread.TIMEOUT_MAX)), which in turn is used to raise an OverflowError if a too large timeout is given.
- Maintains the `MICROPY_PY_THREAD_LOCK_TIMEOUT` feature flag from @mcdeoliveira, since some ports do not support timeouts yet.
  - Adds POSIX feature flags check since apparently my MacOS system doesn't support `pthread_mutex_timedlock` (see [here](https://web.archive.org/web/20080201140752/http://lists.apple.com/archives/Unix-porting/2008/Jan/msg00014.html)).
  - This flag also controls the availability of `_thread.TIMEOUT_MAX`, which is used in the tests to determine whether timeouts are supported or the test should be skipped.
- Takes into account comments from @dpgeorge in https://github.com/micropython/micropython/issues/3332#issuecomment-971109197
  - Modifies the existing `mp_thread_mutex_lock` to add full timeout support instead of adding an extra function.
  - Uses `int64_t` for the timeout value (see note below).

It still has some open issues / questions:

- The code for adding an constant float/int to the `_thread` class is wrong. Looking at other examples like `math.PI` and `sys.maxsize`, it looks like currently all these are defined in the global `obj.h`/`objint.h` so they can have separate implementations for each supported storage format?
  - Even nicer would be if we had some helpers / defines to make it easier to define large integer / float constants, without having to put them into the global int/float implemenation files. But that's probably out of scope for this PR.
- I've switched to `int64_t` because on the Unix port the max timeout would otherwise be a mere 35 minutes. However on many other ports the resolution method (see above) already allows for longer timeouts. We could add another per-port define for the type so we could use a smaller type on some ports? I'm not sure if that would be a meaningful improvement for the smaller microcontrollers.
- The TIMEOUT_MAX value is not entirely accurate; it does for example not take into account the `mp_int_t` size limit, or on the Unix port the fact that we cannot use the full uint32 range because the current time should be subtracted.

And it needs testing on all the platforms; I've don't have hardware for all of them.

Feedback is welcomed!


py/gc.c
```diff
@@ -114,7 +114,7 @@
 
 #if MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL
 #define GC_MUTEX_INIT() mp_thread_recursive_mutex_init(&MP_STATE_MEM(gc_mutex))
-#define GC_ENTER() mp_thread_recursive_mutex_lock(&MP_STATE_MEM(gc_mutex), 1)
+#define GC_ENTER() mp_thread_recursive_mutex_lock(&MP_STATE_MEM(gc_mutex), -1)
```
dpgeorge, line 117: Is it worth defining a constant for this `-1`, to make it clear what it means?  Eg `MP_THREAD_MUTEX_TIMEOUT_FOREVER`.

tests/run-tests.py
```diff
@@ -1281,6 +1281,7 @@ def main():
                     "unicode",
                     "cmdline",
                     "ports/unix",
+                    "thread",
```
dpgeorge, line 1284: I guess it makes sense to add this here.  The tests are actually run on the unix port as port of CI, but it's done separately, see line 268 of `ports/unix/Makefile`.  So probably that line should now be removed, otherwise the thread tests are run twice (on CI at least).

tests/thread/thread_locktimed1.py
```diff
@@ -0,0 +1,40 @@
+# test timed _thread lock object using a single thread
+#
+# MIT license; Copyright (c) 2016 Damien P. George on behalf of Pycom Ltd
```
dpgeorge, line 3: This copyright line can be removed.  There's nothing of substance in this test copied from the other thread tests.

tests/thread/thread_locktimed2.py
```diff
@@ -0,0 +1,42 @@
+# test timed _thread lock object using multiple threads
+#
+# MIT license; Copyright (c) 2016 Damien P. George on behalf of Pycom Ltd
```
dpgeorge, line 3: Copyright can be removed.

dpgeorge:
I haven't done a full review of this yet, but made some comments to start with.  Also, the CI needs to be fixed.

DvdGiessen:
Thank you for the first look! I've addressed the minor ones for now.

The CI failing is a bit inconsistent. A few failures are expected, for example for the cc3200 port the code simply cannot be compiled due to the mentioned open question about defining numeric constants. Some others seem to be unrelated failures.

This PR touches a lot of ports, and for some the changes were merely tested to build / checked against platform documentation but not run on real hardware yet.

tests/run-tests.py
DvdGiessen: It did break a few other tests in CI such as the Windows port which also uses this set, so I've reverted it.

PR #15101, opened 2024-05-22
ports/esp32: Make LAN.active(bool) idempotent.

elvis-epx:
Do not raise an exception when LAN.active(False) is called but LAN is already inactive, or LAN.active(True) is called but LAN is already active.

Aligns this LAN API with with WLAN.active(bool), which is already idempotent.

Within ESP-IDF, it can be seen at [1] and [2] that esp_eth_start() and esp_eth_stop() return ESP_ERR_INVALID_STATE only when the state contradicts the intention, not for any actual error.

[1] https://github.com/espressif/esp-idf/blob/master/components/esp_eth/src/esp_eth.c#L278
[2] https://github.com/espressif/esp-idf/blob/master/components/esp_eth/src/esp_eth.c#L301


elvis-epx:
Deferred to https://github.com/micropython/micropython/pull/15385

PR #15107, opened 2024-05-24
esp32: Add support for TCP_NODELAY.

greezybacon:
This adds support for the `TCP_NODELAY` socket option for lwIP sockets. Generally, TCP sockets use the Nagle algorithm and will send data when an ACK is received or after all previously-sent data has already been ACKed.

If the `TCP_NODELAY` option is set for a socket, every write to the socket will trigger a packet to be sent.


dpgeorge:
Thanks!

PR #15108, opened 2024-05-24
ports/esp32: Use shared/tinyusb integration.

andrewleech:
Uses newer TinyUSB `synopsys/dwc2` driver for esp32s2 and esp32s3 rather than the IDF tinyusb component.

This allows re-use of other tinyusb integration code / features shared between ports.

Also add / fix automatic bootloader handling for S2 and S3 by adding support for the ESP standard DTR/RTS based reboot to bootloader.

On S2 and S3 this requires esptool to be run twice (fails on first run as the chip resets and usb drops out) unless used with: https://github.com/espressif/esptool/pull/980

This PR also switches from OTG to Serial/Jtag mode on S3 to workaround issue discussed in:
https://github.com/espressif/arduino-esp32/issues/6762

Also, after flashing with esptool, auto-reboot back to application (micropython) works when used with new enough esptool containing https://github.com/espressif/esptool/commit/0215786283660480e9ec85dd077e6fc2f46919e9
This is not in the current release 4.8, so 4.9 beta release or install from GitHub url is needed.


andrewleech:
This is now tested and working on both S2 and S3.

ports/esp32/mpconfigport.h
```diff
@@ -197,6 +197,70 @@
...
+
+#ifndef MICROPY_HW_USB_PRODUCT_FS_STRING
+#ifdef CONFIG_TINYUSB_DESC_PRODUCT_STRING
+#define MICROPY_HW_USB_PRODUCT_FS_STRING CONFIG_TINYUSB_DESC_PRODUCT_STRING
+#else
+#define MICROPY_HW_USB_PRODUCT_FS_STRING "Board in FS mode"
+#endif
+#endif
+
+#define MICROPY_HW_USBDEV_TASK_HOOK extern void mp_usbd_task(void); mp_usbd_task();
```
dpgeorge, line 259: Does it need to poll `tud_task()` like this?  Surely it can be done by scheduling from a USB IRQ?  That's how the rp2 port works.

How did the IDF handle calling `tud_task()` prior to this PR, did it have a dedicated FreeRTOS thread to run TinyUSB?

ports/esp32/mphalport.h
```diff
@@ -52,6 +52,7 @@
 
 extern TaskHandle_t mp_main_task_handle;
 
+extern int mp_interrupt_char;
```
dpgeorge, line 55: I think it's better to include `shared/runtime/interrupt_char.h` rather than declare this variable.

shared/tinyusb/mp_usbd_cdc.c
```diff
@@ -162,6 +172,15 @@ tud_cdc_line_state_cb
         tud_sof_cb_enable(true);
     }
     #endif
+    #if MICROPY_HW_ESP_AUTOMATIC_BOOTLOADER
```
dpgeorge, line 175: I reckon this new config option should be called `MICROPY_HW_USB_CDC_DTR_RTS_BOOTLOADER`, to indicate that it's part of the USB CDC code.

And then put a default value in `shared/tinyusb/mp_usbd_cdc.h` of 0 (disabled) with a comment that this should only be used on ESP32 boards.

ports/esp32/Makefile
```diff
@@ -29,8 +29,12 @@ BUILD ?= build-$(BOARD)
...
 
 # Device serial settings.
-PORT ?= /dev/ttyUSB0
-BAUD ?= 460800
+ifneq ($(PORT),)
+PORT_ARG := -p $(PORT)
+endif
+ifneq ($(BAUD),)
+BAUD_ARG := -b $(BAUD)
+endif
```
dpgeorge, line 37: So this now leaves the selection of port and baud to `idf.py`?  Does that give a better user experience?

andrewleech: As a default, yes. Esptool does have some auto detection logic that has worked very well for me during development.

ports/esp32/mpconfigport.h
andrewleech: Yeah I'm pretty sure IDF has a low priority tight loop task servicing the USB.

I'll have to look into whether we can use the interrupt, you're right it would certainly be a better option.

dpgeorge: You should be able to look at the source for that tight-loop task, and see what semaphore it waits on.  Then see where that semaphore is set.  That's the location we can schedule our tud task callback.

andrewleech: Oh, huh, the driver we're using here does that automatically for us! 

https://github.com/espressif/tinyusb/blob/9554283b03cb4af00119ccd5a38e539749b32d53/src/portable/espressif/esp32sx/dcd_esp32sx.c#L867

I'll just want to confirm that this is definitely getting called to configure it.

dpgeorge: You'll need to wrap `_dcd_int_handler` at the linker level, I guess.

dpgeorge:
@andrewleech I see you made changes.  Is this ready for re-review?

andrewleech:
Not quite sorry @dpgeorge I need to test them still.

andrewleech:
Ok now it's re-tested on S2 and S3 with another rebase.

I've fixed up based on review feedback and yes it all appears to work correctly without the polling of the USB task so that's great!

dpgeorge:
I tested this on a FeatherS2 with ESP32_GENERIC_S2 firmware.

I can get a REPL but it's easy to lock it up, eg by typing "help()" at the prompt:
```
$ mpremote
MicroPython v1.24.0-preview.152.g88e404dd5 on 2024-07-26; Generic ESP32S2 module with ESP32S2
>>> help()
Welcome to MicroPython on the ESP32!

For online docs please visit http://docs.micropython.org/

For access to the hardware use the 'machine' module:

import machine
pin12 = machine.Pin(12, machine.Pin.OUT)
pin12.value(1)
pin13 = machine.Pin(13, machine.Pin.IN, machine.Pin.PULL_UP)
```
At that point it locks up and requires a hard reset.

dpgeorge:
I also tested a TinyS3 with ESP32_GENERIC_S3 firmware, and it's also easy to lock up, eg with `help()`, or running the test suite eg `tests/basics/int_big1.py` fails.

andrewleech:
Thanks @dpgeorge thankfully the fix was easy to find and fairly obvious when looking at it! https://github.com/micropython/micropython/compare/c7d865170da1946b9dae2416dbed9fc0586fbd99..949495e5c0e370772883404c94a08f9bbba1aaf1#diff-e356b2686148d4a1bb2533f6af83551c711c223cca6e01e395ee91af6b255cb0

```
anl@STEP: ~/micropython/tests $ ./run-tests.py --target esp32 --device /dev/ttyACM0 basics/int_big1.py
pass  basics/int_big1.py
1 tests performed (178 individual testcases)
1 tests passed
```

dpgeorge:
> TLDR stdout was releasing the gil for larger transfers, which made sense when IDF was doing it completely itself. Now there's more micropython interaction and idle sleeps it certainly doesn't make sense anymore.

Ah, right!

The reason the GIL must be obtained (ie not released) is because `mp_usbd_cdc_tx_strn()` may release the GIL while waiting for USB FIFOs to drain.

But, I still think it's a good idea to release the GIL when doing UART output, and also probably USB JTAG output, because both of those may take some time for large output.

So I suggest to add back the GIL release in `mp_hal_stdout_tx_strn()`, but only for USB serial JTAG and UART output.

dpgeorge:
I've tested this on an S2 and S3 and it's looking a lot better.

I have a test which tests throughput (and integrity) of the USB CDC.  Prior to this PR this test would fail on S2 and S3 (not sure why...).  With this PR the test now passes!  So that's great, the USB CDC is more robust.

But, performance is a lot worse with this PR.  This can be improved by about a factor of 10 by waking the main task whenever there is a USB IRQ.  This can be achieved by calling `mp_hal_wake_main_task_from_isr()`, at the end of  `__wrap_dcd_event_handler()`.

With that change through put on USB CDC is much better:
- Data from device to host PC: around 3 Mbit/sec
- Data from host PC to device: around 1.1 Mbit/sec

For PC-to-device that's on par with RP2040.  But for device-to-PC it could be better,  RP2040 gets around 4.5 Mbit/sec.  But I think for now that's an acceptable rate.

@andrewleech can you please add a call to `mp_hal_wake_main_task_from_isr()` as described above?  This will require a little thought how to make it work with other ports that don't define this function.

andrewleech:
Thanks for the testing @dpgeorge that's a good outcome. I too would have thought faster speeds should be possible but that's still quite functional (and certainly better than UART) so a good first step. It could be an interesting exercise in future to identify where any speed restrictions come from.

Are your speed testing scripts easy enough to share? I still want to work on porting stm over to tinyusb and it would be good to compare speeds before and after. 

Regardless, I've reworked the GIL handling as suggested and added `mp_hal_wake_main_task_from_isr()` along with a weak stub that should satisfy other ports.

pillo79:
Not sure I understand the CI error, I'm not able to reproduce it locally even with the exact same build command (`ci_esp32_build_s3_c3`).
Building on the `espressif/idf:release-5.0` Docker image (ESP-IDF v5.0.7-86-gb87b33c97f).

ports/esp32/usb.c
```diff
@@ -28,76 +28,77 @@
...
-        #endif
+    // Configure USB PHY
+    usb_phy_config_t phy_conf = {
+        .controller = USB_PHY_CTRL_OTG,
+        .otg_mode = USB_OTG_MODE_DEVICE,
     };
-    ESP_ERROR_CHECK(tusb_cdc_acm_init(&acm_cfg));
 
-}
+    #if EXTERNAL_PHY
```
projectgus, line 53: Does MicroPython support any boards with external PHY? If not then maybe safer to drop this for now?

dpgeorge:
@andrewleech can you please rebase this on latest master, which has had some changes, eg due to #15727.  Then it should be ready for final review and merge.

ports/esp32/esp32_common.cmake
```diff
@@ -52,6 +52,45 @@ list(APPEND MICROPY_SOURCE_DRIVERS
     ${MICROPY_DIR}/drivers/dht/dht.c
 )
 
+if(MICROPY_PY_TINYUSB)
+    string(CONCAT GIT_SUBMODULES "${GIT_SUBMODULES} " lib/tinyusb)
+
+    if(ECHO_SUBMODULES)
+    # No-op, we're just doing submodule/variant discovery.
+    # Cannot run the add_library/target_include_directories rules (even though
+    # the build won't run) because IDF will attempt verify the files exist.
```
projectgus, line 61: There's an obscure failure mode here where if someone does:

```
make submodules
make BOARD=ESP32_GENERIC_S2
```

then the build may fail, because the submodules step didn't fetch TinyUSB. (I note the ci.sh script needed to be updated not to trigger exactly this failure mode!)

Suggest we always append lib/tinyusb to the submodules list for esp32 port. The downside is small, this check and explanation can be removed, and I don't think anyone will ever notice they got a maybe-unnecessary submodule (but definitely someone somewhere will avoid running into a weird build error).

ports/esp32/mpconfigport.h
```diff
@@ -200,6 +200,63 @@
 
 #define MP_STATE_PORT MP_STATE_VM
 
+#if CONFIG_USB_OTG_SUPPORTED
+#ifndef MICROPY_HW_ENABLE_USBDEV
+#define MICROPY_HW_ENABLE_USBDEV            (1)
```
projectgus, line 205: This is a nitpick, but you can do

```suggestion
#define MICROPY_HW_ENABLE_USBDEV            (SOC_USB_OTG_SUPPORTED)
```
... and avoid the outer `#if CONFIG_USB_OTG_SUPPORTED` statement

(The existence of `CONFIG_USB_OTG_SUPPORTED` in ESP-IDF is really only a kluge so the config system in ESP-IDF can know the value of `SOC_USB_OTG_SUPPORTED`)

projectgus (commented):
This looks really good, @andrewleech! I'm particularly looking forward to the ESP32-S3 port resetting after flashing, and to MicroPython being able to make use of the idf.py port detection features! Not to mention the potential for runtime USB devices on ESP32!!

Long term I'm a little concerned about depending on private ESP-IDF APIs, but most of these look pretty low-level and therefore should be pretty stable.

ports/esp32/esp32_common.cmake
andrewleech: This is a failure case on many ports for a number of submodules, `nimble` is a common one for that too. But yeah I'm happy to always include it just to simplify things!

ports/esp32/usb.c
andrewleech: These code paths hadn't been tested either so yes safest to remove for now. They were referenced from TinyUSB example code, which can be referred to again in future if these features are needed.

andrewleech:
Hi @dpgeorge this has been rebased now. 
Thanks for the review and useful comments @projectgus - I've implemented those suggestions too.

Yeah I didn't love needing to hit the couple of esp functions but that still seemed better than replicating anything extra.

I'm dissapointed that https://github.com/espressif/esptool/pull/980 got 99% of the way there before being merged in a state that's no better than without it in my testing, but the reboot stuff I could handle from the micropython side does improve quality of life certainly!

ports/esp32/mpconfigport.h
```diff
@@ -201,6 +201,61 @@
...
+#define MICROPY_HW_ENABLE_USBDEV            (SOC_USB_OTG_SUPPORTED)
+#endif
+
+#if MICROPY_HW_ENABLE_USBDEV
+#define MICROPY_SCHEDULER_STATIC_NODES        (1)
+#define MICROPY_HW_USB_CDC_DTR_RTS_BOOTLOADER (1)
+
+// Enable USB-CDC serial port
+#ifndef MICROPY_HW_USB_CDC
+#define MICROPY_HW_USB_CDC                  (1)
```
projectgus, line 214: There's a second definition of this on L319 that should probably be merged with this one, somehow.

(maybe add it around line 207, and define it to the value of MICROPY_HW_ENABLE_USBDEV if otherwise undefined?)

projectgus (commented):
This looks good to me! One non-critical note about the port header, but also please double-check the `ARDUINO_NANO_ESP32` config. I think a6c35aee will need reverting here for it to work correctly.

dpgeorge:
Let's update TinyUSB to 0.17.0 separately in #15902.

andrewleech:
> This looks good to me! One non-critical note about the port header, but also please double-check the `ARDUINO_NANO_ESP32` config. I think [a6c35ae](https://github.com/micropython/micropython/commit/a6c35aeee873953bd9016cf63feb1a6772bdd6af) will need reverting here for it to work correctly.

Thanks @projectgus I hadn't really taken a look at this board properly in the past, I'm glad I did now. The entire support around `MICROPY_HW_USB_EXTERNAL_TINYUSB` was essentially there just for this board, just so that a custom wrapper could be declared to connect our `tud_cdc_line_state_cb()` to the tinyusb stack provided by esp idf.

Now that I've seen that, I can delete all of this and clean up both the declaration of `tud_cdc_line_state_cb()` and the custom code in that board because we don't need to support IDF tinyusb anymore.

ports/esp32/mpconfigport.h
andrewleech: Yeah thanks, I'd missed this duplication. I've consolidated this a bit better now, I think it's fairly clear.

pillo79:
Sorry for noticing so late. Yes, all of the changes introduced for the Nano were only to reuse the common "1200 touch" code despite having to import the IDF TinyUSB implementation.
Thanks for this, looks a lot cleaner now! :rocket: :clap:

dpgeorge:
@andrewleech can you please rebase on latest master to pick up the tinyusb update.  Then hopefully that's the final bit!

shared/tinyusb/mp_usbd_cdc.c
```diff
@@ -149,13 +149,7 @@ static void usbd_cdc_run_bootloader_task(mp_sched_node_t *node) {
...
 #endif
 
-void
-#if MICROPY_HW_USB_EXTERNAL_TINYUSB
-mp_usbd_line_state_cb
-#else
-tud_cdc_line_state_cb
-#endif
-    (uint8_t itf, bool dtr, bool rts) {
+void tud_cdc_line_state_cb(uint8_t itf, bool dtr, bool rts) {
```
dpgeorge, line 162: This change (in the first commit here) needs a corresponding change in `mp_usbd_cdc.h` to remove the public declaration of `mp_usbd_line_state_cb()`.

andrewleech: Ah yes I only just realised that earlier today myself! I'll fix it and rebase asap thanks.

sebromero:
Amazing work @andrewleech 👏 Will this allow us to use the existing USB HID classes?

andrewleech:
@sebromero the intention is to support the runtime USB subsystem yes! However I haven't enabled or tested that yet as part of this PR.

I'm not sure if I should just enable the runtime support defines here by default for the S2 and S3, what do you think @projectgus ?

sebromero:
@andrewleech Nice! Let me know if / how I can help testing this. I can also get my students to run an experiment so we would have plenty of test data :-)

dpgeorge:
> I'm not sure if I should just enable the runtime support defines here by default for the S2 and S3

Let's leave that for a separate PR.

shared/tinyusb/mp_usbd.c
```diff
@@ -49,12 +49,17 @@ void mp_usbd_task_callback(mp_sched_node_t *node) {
 
 extern void __real_dcd_event_handler(dcd_event_t const *event, bool in_isr);
 
+// Provide stub for ports that don't have/need the wake main task function.
+void __attribute((weak)) mp_hal_wake_main_task_from_isr(void) {
```
dpgeorge, line 53: I suggest to do this slightly differently and force all ports using these shared TinyUSB bindings to define this function.  The ports can just add the following to their `mphalport.h`:
```c
static inline void mp_hal_wake_main_task_from_isr(void) {
    // Nothing needs to be done here.
}
```

Reasons:
1. Reduce code size and don't need to call a useless empty function (see rp2 code size report).
2. Don't rely on weak linking, which may not be that portable and could lead to issues if the wrong function is linked instead.
3. Make it explicit that a port doesn't need to implement anything for this MP HAL function, and provide a clear place to add it if needed (self documenting).

dpgeorge:
I tested this on ESP32 S2 and S3 and it works, and it looks all good except the comment above about `mp_hal_wake_main_task_from_isr()`.

shared/tinyusb/mp_usbd.c
andrewleech: Thanks @dpgeorge I'm not a huge fan on boilerplate but yes it's true weak functions can be a little unclear / non-deterministic, it's hard to be certain which version is actually used. 

I considered using a define instead to set if this function should be needed, but that's hard to document / describe as well. 

This probably is the best compromise, I've implemented it for each of the TinyUSB ports now.

dpgeorge: > I considered using a define instead to set if this function should be needed, but that's hard to document / describe as well.

Yeah, I had the same thoughts as you.

Thanks for updating!

dpgeorge:
Merged!

Thanks very much @andrewleech for all your efforts on this, and everyone else who contributed and tested.

sebromero:
> I'm not sure if I should just enable the runtime support defines here by default for the S2 and S3, what do you think

@andrewleech Any hint on how to do this? 🙏  Would love to test this.

andrewleech:
@sebromero keep an eye on this: https://github.com/micropython/micropython/issues/15564#issuecomment-2412918671

gampam2000:
Hi, recently I can't compile micropython anymore. I came across this PR in my search and could it be that the errors are related to this? I Use an ESP32S3. 

I get errors like this:

```
In file included from /micropython/ports/esp32/usb.c:36:
/micropython/shared/tinyusb/mp_usbd.h: In function 'mp_usbd_init_tud':
/micropython/shared/tinyusb/mp_usbd.h:46:5: error: unknown type name 'tud_cdc_configure_fifo_t'
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
/micropython/shared/tinyusb/mp_usbd.h:46:38: error: field name not in record or union initializer
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |                                      ^
/micropython/shared/tinyusb/mp_usbd.h:46:38: note: (near initialization for 'cfg')
/micropython/shared/tinyusb/mp_usbd.h:46:58: error: field name not in record or union initializer
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |                                                          ^
/micropython/shared/tinyusb/mp_usbd.h:46:58: note: (near initialization for 'cfg')
/micropython/shared/tinyusb/mp_usbd.h:46:75: warning: excess elements in scalar initializer
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |                                                                           ^
/micropython/shared/tinyusb/mp_usbd.h:46:75: note: (near initialization for 'cfg')
In file included from /micropython/ports/esp32/main.c:55:
/micropython/shared/tinyusb/mp_usbd.h: In function 'mp_usbd_init_tud':
/micropython/shared/tinyusb/mp_usbd.h:46:5: error: unknown type name 'tud_cdc_configure_fifo_t'
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
/micropython/shared/tinyusb/mp_usbd.h:46:38: error: field name not in record or union initializer
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |                                      ^
/micropython/shared/tinyusb/mp_usbd.h:46:38: note: (near initialization for 'cfg')
/micropython/shared/tinyusb/mp_usbd.h:46:58: error: field name not in record or union initializer
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |                                                          ^
/micropython/shared/tinyusb/mp_usbd.h:46:58: note: (near initialization for 'cfg')
/micropython/shared/tinyusb/mp_usbd.h:46:75: warning: excess elements in scalar initializer
   46 |     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
      |                                                                           ^
/micropython/shared/tinyusb/mp_usbd.h:46:75: note: (near initialization for 'cfg')
/micropython/shared/tinyusb/mp_usbd.h:47:5: error: implicit declaration of function 'tud_cdc_configure_fifo' [-Werror=implicit-function-declaration]
   47 |     tud_cdc_configure_fifo(&cfg);
      |     ^~~~~~~~~~~~~~~~~~~~~~
/micropython/shared/tinyusb/mp_usbd.h:47:5: error: implicit declaration of function 'tud_cdc_configure_fifo' [-Werror=implicit-function-declaration]
   47 |     tud_cdc_configure_fifo(&cfg);
      |     ^~~~~~~~~~~~~~~~~~~~~~
cc1: some warnings being treated as errors
.....
```

Any ideas how to fix this?

Thanks & Best regards

vshymanskyy:
> I can't compile micropython anymore

For some reason, i needed to run this manually:
```sh
make -f ../../py/mkrules.mk GIT_SUBMODULES="lib/berkeley-db-1.xx lib/tinyusb lib/micropython-lib" submodules
```

This didn't do anything useful (probably this is a regression):
```sh
make BOARD=ESP32_GENERIC_S3 submodules
```

gampam2000:
`make BOARD=ESP32_GENERIC_S3 submodules` does work 

```
make[1]: Entering directory '/lib/micropython/ports/esp32'
Updating submodules: lib/berkeley-db-1.xx lib/tinyusb lib/micropython-lib
Synchronizing submodule url for 'lib/berkeley-db-1.xx'
Synchronizing submodule url for 'lib/micropython-lib'
Synchronizing submodule url for 'lib/tinyusb'
Submodule path 'lib/tinyusb': checked out '5217cee5de4cd555018da90f9f1bcc87fb1c1d3a'
make[1]: Leaving directory '/lib/micropython/ports/esp32'
```

It checks out `5217cee5de4cd555018da90f9f1bcc87fb1c1d3a` of tinyusb, is this the correct Git ID?

gampam2000:
I found the issue: I needed to add `set(MICROPY_PY_TINYUSB ON)` in my boards `CMakeLists.txt`

gampam2000:
> Just to clarify, you were referring to building the `ESP32_GENERIC_S3` board earlier, when you now say you needed to set `MICROPY_PY_TINYUSB` is this on a custom board definition of yours? 

Yes I used this only to initialize the submodules.

> Normally that setting is configured automatically in
> 
> https://github.com/micropython/micropython/blob/1b89c503db690967d50699abe0bfa942f6f6b15e/ports/esp32/main_esp32s3/CMakeLists.txt#L10
> so I'm not sure why it's not working for you.

Thanks for pointing that out.
I'm using the [micropython-example-boards](https://github.com/micropython/micropython-example-boards) setup, I guess that file is than not automatically included. Maybe this is an issue in the micropython-example-boards?

cnadler86:
> > I can't compile micropython anymore
> 
> For some reason, i needed to run this manually:
> 
> ```shell
> make -f ../../py/mkrules.mk GIT_SUBMODULES="lib/berkeley-db-1.xx lib/tinyusb lib/micropython-lib" submodules
> ```
> 
> This didn't do anything useful (probably this is a regression):
> 
> ```shell
> make BOARD=ESP32_GENERIC_S3 submodules
> ```

I had also problems while compiling on my board definition. This solved also my problem. Maybe on a fresh micropython repo-checkout, this is not needed, because the submodules are automatically registered and that is why the pipeline is passing, but on existing repos the submodule is not been registered automatically.

Vincent1-python:
But now this driver is too old for ESP32P4 driver, and the code needs to be updated.

PR #15111, opened 2024-05-24
rp2: Selectively leave the USB clocks enabled in light sleep.

projectgus:
Without this change going to lightsleep stops the USB peripheral clock, and can lead to either the device going into a weird state or the host deciding to issue a bus reset.

This change only keeps the USB peripheral clocks enabled if the USB device is currently active and a host has configured the device. Otherwise USB clocks are disabled same as before.

Increases power consumption in light sleep (measured at the USB port for a PICO board as 1.2-1.3mA without this change, and 3.4-3.5mA with this change).  I think can argue that if you have a working active USB host connection, an extra 2mA is unlikely to be significant.

(EDIT: Above measurement was originally off by a factor of 10, corrected in comments.)

As discussed at https://github.com/orgs/micropython/discussions/14401

## Tested with

* The code supplied in the linked discussion, via `mpremote run`
* `mpremote a1 exec "import machine; machine.lightsleep(5000)"` (plus some longer and shorter delays, up to 30 seconds).
* Same command as above, via REPL (allows interactively sending bytes to the USB device to try and bring it back.)

# Details

Logic analyser shows that without this change the USB Host continues sending SoF and other requests to the sleeping device, and it doesn't respond so every transfer is an error. On my system, the host seems to keep trying indefinitely. After waking, TinyUSB ends up in a bad state and doesn't set up any new transfers itself (i.e. the USB-CDC doesn't send any data to the host.) If using a REPL, pressing a key to send some data from the host to the device is enough to make it come back from this weird state and it seems to resume normal operations.

Behaviour seems to vary between host controllers - the discussion includes an [example of the host deciding the device is no longer responding and resetting it](https://github.com/orgs/micropython/discussions/14401#discussioncomment-9282604). I wasn't able to reproduce this here, can only guess it depends on the host controller (and/or maybe the hub controller, especially for superspeed hubs.)

With this change applied, the USB device continues to respond to host transfers and (presumably) will even complete pending endpoint transfers. All other requests are NAKed while still asleep, but the interaction with the host seems to resume correctly on wake.

# Alternatives

One alternative approach would be to disconnect the USB device pins when going into lightsleep. This would cause the host to definitely see the device as disconnected, rather than ending up in the current weird in-between states. Also no increase in power usage. If we do the planned work to support transparent USB serial disconnect/reconnect in mpremote then it will probably be possible for mpremote to cleanly recover the connection on wake, also.

I also did some experiments with getting TinyUSB and the host to recover from the weird in-between state, without the host needing to initiate an OUT transfer. This obviously doesn't help for hosts which trigger a full bus reset, but might help for some hosts. I had a little unreliable success with this, but it seems flaky.

### DIFF
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -28,6 +28,7 @@
 // extmod/modmachine.c via MICROPY_PY_MACHINE_INCLUDEFILE.
 
 #include "py/mphal.h"
+#include "mp_usbd.h"
 #include "modmachine.h"
 #include "uart.h"
 #include "hardware/clocks.h"
@@ -134,8 +135,17 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
         return;
     }
     #endif
-    // Disable USB and ADC clocks.
-    clock_stop(clk_usb);
+
+    #if MICROPY_HW_ENABLE_USBDEV
+    // Only disable the USB clock if a USB host has not configured the device
+    bool disable_usb = !tud_mounted();
+    #else
+    bool disable_usb = true;
+    #endif
+    if (disable_usb) {
+        clock_stop(clk_usb);
+    }
+
     clock_stop(clk_adc);
 
     // CLK_REF = XOSC
@@ -152,7 +162,9 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
 
     // Disable PLLs.
     pll_deinit(pll_sys);
-    pll_deinit(pll_usb);
+    if (disable_usb) {
+        pll_deinit(pll_usb);
+    }
 
     // Disable ROSC.
     rosc_hw->ctrl = ROSC_CTRL_ENABLE_VALUE_DISABLE << ROSC_CTRL_ENABLE_LSB;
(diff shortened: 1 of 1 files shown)


dpgeorge:
This is a really great enhancement (and very simple)!  It's not very user friendly that USB resumes in a broken state after lightsleep finishes.

Tested it and it works very well.

But I think you're current measurements are a factor of 10 too high.  Idle REPL consumption is about 15mA, lightsleep is about 1.35mA without this PR, and about 3.67mA with this PR.  IMO that's totally acceptable, a few extra mA to keep the USB alive.

If users really need to get lower power consumption with USB connected they can use `machine.USBDevice` to turn off the USB before entering lightsleep.

projectgus:
> But I think you're current measurements are a factor of 10 too high. Idle REPL consumption is about 15mA, lightsleep is about 1.35mA without this PR, and about 3.67mA with this PR. IMO that's totally acceptable, a few extra mA to keep the USB alive.

Oh you're right, that does sound more reasonable! Thanks.

mungewell:
@projectgus 
I submitted a PR which exposes all the 'sleep_en[01]' bits to userland, so that I can tell Pico to keep the PIO clocks running.
https://github.com/micropython/micropython/pull/16183

How would you feel if I replaced the 'disable_usb' parameter, and included checking 'sleep_en[]' for the appropriate bits - thus keeping USB clock running.

projectgus:
@mungewell Thanks for the heads-up. I will aim to get back to you on your PR soon.

mungewell:
> they can use machine.USBDevice to turn off the USB before entering lightsleep

Can someone please explain what command I can use to turn of the USB. I have enabled (and connected to) the UART REPL, the USB is still showing connected when I `lsub` on my Linux box.

PR #15158, opened 2024-05-29
ports/nrf: Consolidate stdio functions.

andrewleech:
Consolidate CDC, UART and NUS stdio interfaces into the one handler. 
Allows any/all of them to be enabled separately.

Note there are some conflicts between UART and CDC REPL handling, particularly if hwfc is enabled on the UART. 
This can block for long enough to cause the CDC connection to be broken. This is not a new bug as such, the current release of nrf port is restricted to not allow both CDC and UART to be enabled at the same time at all.

Updates UART REPL to use similar define to other platforms:
MICROPY_HW_ENABLE_UART_REPL

Uses shared/tinyusb cdc implementation.

Thie PR builds upon (and therefore include commits from) https://github.com/micropython/micropython/pull/14462 and https://github.com/micropython/micropython/pull/14485

It has currently only been tested on a nrf52840 dongle, however uart, cdc and nus support have individually all worked on this board.


robert-hh:
This PR does not even build for a Arduino BLE 33 Sense or a SEEED XIAO NRF52 or NRF52840_MDK_USB_DONGLE. Compile errors:

rivers/usb/usb_cdc.c: In function 'mp_usbd_port_get_serial_number':
drivers/usb/usb_cdc.c:108:46: error: 'MICROPY_HW_USB_DESC_STR_MAX' undeclared (first use in this function)
  108 |     MP_STATIC_ASSERT(sizeof(deviceid) * 2 <= MICROPY_HW_USB_DESC_STR_MAX);
      |                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~
../../py/misc.h:54:61: note: in definition of macro 'MP_STATIC_ASSERT'
   54 | #define MP_STATIC_ASSERT(cond) ((void)sizeof(char[1 - 2 * !(cond)]))
      |                                                             ^~~~
drivers/usb/usb_cdc.c:108:46: note: each undeclared identifier is reported only once for each function it appears in
  108 |     MP_STATIC_ASSERT(sizeof(deviceid) * 2 <= MICROPY_HW_USB_DESC_STR_MAX);
      |                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~
../../py/misc.h:54:61: note: in definition of macro 'MP_STATIC_ASSERT'
   54 | #define MP_STATIC_ASSERT(cond) ((void)sizeof(char[1 - 2 * !(cond)]))
      |                                                             ^~~~
CC ../../lib/tinyusb/src/portable/nordic/nrf5x/dcd_nrf5x.c
drivers/usb/usb_cdc.c:109:5: error: implicit declaration of function 'mp_usbd_hex_str'; did you mean 'mp_obj_new_str'? [-Werror=implicit-function-declaration]
  109 |     mp_usbd_hex_str(serial_buf, (uint8_t *)deviceid, sizeof(deviceid));
      |     ^~~~~~~~~~~~~~~
      |     mp_obj_new_str
drivers/usb/usb_cdc.c: In function 'usb_cdc_init':
drivers/usb/usb_cdc.c:124:5: error: implicit declaration of function 'mp_usbd_init'; did you mean 'mp_set_init'? [-Werror=implicit-function-declaration]
  124 |     mp_usbd_init();
      |     ^~~~~~~~~~~~
      |     mp_set_init
CC device/startup_nrf52840.c
drivers/usb/usb_cdc.c: In function 'USBD_IRQHandler':
drivers/usb/usb_cdc.c:143:5: error: implicit declaration of function 'tud_int_handler' [-Werror=implicit-function-declaration]
  143 |     tud_int_handler(0);
      |     ^~~~~~~~~~~~~~~

Rebasing to the current master branch fails as well due to conflicts in the make files of the other ports. Nothings difficult, just duplicating references to files, but it breaks rebase.

If you look at the CI of this PR, you see that it fails as well in building the firmware.

andrewleech:
Ah yes it seems I rushed ahead with this, it was building previously before I went and consolidated some more settings to make UART and ble repl all work the same way. 
A little more work to do to get it cleaned up!

robert-hh:
The mpconfigboard.h file for the SEEED and Arduino board miss enabling of MICROPY_HW_ENABLE_USBDEV and MICROPY_HW_USB_CDC.

andrewleech:
Ah thanks for looking into that. I should make that define enabled by default if USB is enabled.

robert-hh:
Adding `#define MICROPY_HW_ENABLE_USBDEV    (1)` makes both the Seeed NRF52 board and the Arduino Nano 33 BLE Sense compile & build properly, given that  the old version of lib/tinyusb is used and the branch is **not** rebased to master. The boards boot and works fine. 
When building with the actual master branch, the Seeed board refuses to work. The USB device does not register.

robert-hh:
Does not build for Arduino Nano 33 BLE. Which commit of the tinyusb lib do you use for your builds? The master branch uses 
When updating the submodules in your PR, I get now `d10b65ada4be7d5754b3128e80a9b4db72bdb23f`
The master branch is at `1fdf29075d4e613eacfa881166015263797db0f6`

robert-hh:
Both SEEED NRF52 and Arduino Nanon 33 BLE build and run. Only I had to define MICROPY_HW_ENABLE_USBDEV as 1.

andrewleech:
Ah, I see now, those boards had just `MICROPY_HW_USB_CDC` defined, not `MICROPY_HW_ENABLE_USBDEV`. 

I think it makes sense more broadly to have `MICROPY_HW_ENABLE_USBDEV` as the "main" define to enable usb, so I've added it to those two boards.  All other nrf boards with USB seemed to have both defines already.

For reference, the tinyusb update came in for https://github.com/micropython/micropython/pull/14485 and is needed for those features. I don't think it should matter for just this PR though, I really would have thought the previous tinyusb commit would be fine for the changes just made here.

I also don't know what's going on with the CI build failure
```
LINK build-PCA10056-s140/firmware.elf
/usr/lib/gcc/arm-none-eabi/9.2.1/../../../arm-none-eabi/bin/ld: /tmp/firmware.elf.6R2Ffu.ltrans0.ltrans.o: in function `__wrap_dcd_event_handler':
<artificial>:(.text+0x3f3c): undefined reference to `dcd_event_handler'
```
I can build this board with s140 just fine on both `Arm GNU Toolchain 12.3.Rel1` and `GNU Tools for Arm Embedded Processors 7-2018-q2-update) 7.3.1 20180622`
I see the CI build is performed with the arm compile from ubuntu repos, I might have to try with that.

andrewleech:
Ok that build failure:
```
LINK build-PCA10056-s140/firmware.elf
/usr/lib/gcc/arm-none-eabi/9.2.1/../../../arm-none-eabi/bin/ld: /tmp/firmware.elf.6R2Ffu.ltrans0.ltrans.o: in function `__wrap_dcd_event_handler':
<artificial>:(.text+0x3f3c): undefined reference to `dcd_event_handler'

```
Is caused by:
>  [Bug 24406](https://sourceware.org/bugzilla/show_bug.cgi?id=24406) - -Wl,--wrap= incompatible with -flto

Which was apparently fixed in binutils 2.33
The CI build is on ubuntu focal, which has `binutils-arm-none-eabi amd64 2.34-4ubuntu1+13ubuntu1` so should have had that particular fix, though there's a few linked tickets from that bug report.

Of note I don't have the issue at home with gcc 12.x so it's been fixed since then, but it's hard to tell exactly when.

andrewleech:
Also thanks @robert-hh your reports made it clear there were a number of build failures on nrf52840 when `MICROPY_HW_USB_CDC = 1` but `MICROPY_HW_ENABLE_USBDEV = 0`
I've cleaned them up too, as there's definitely cases where you might want that chip without USB being used.

robert-hh:
Building with that PR works fine for the SEEED and Arduino boards, linking is now almost instantly. Both run. However rebasing on top of the current master branch causes conflicts in shared/tinyusb/mp_usbd_cdc.c and differences in lib/tinyusb.

P.S:: I'm using GCC 12.2 as the standard version of the OS.

andrewleech:
Thanks I've rebased again onto master now my previous PR has been merged!

robert-hh:
I compiled and ran this PR on two NRF52 boards. They build fine and work most of the times. But sometimes the boards do not register as USB devices any more, even when I do a full power cycle.  The boards are connected to an external hub. Unplugging the hub from the PC and re-plugging helps most of the time. Connection the boards directly to the PC does not work. There are similar reports in the MIMXRT section on Discord/micropython. So it may be a generic problem of TinyUSB. Note to @projectgus.
Can you suggest something to trace the state more, now that it happens here?

robert-hh:
@andrewleech  SO I see a strange behavior on the RP2040 port. If in main.py a script is started which enables a Pin IRQ (and disables), then after a hard reset or power up USB is blocked like detailed above. I have to disconnect the USB hub and reconnect it to get USB working again, and then I see at the next boot fragments of the messages created by scripts started from main.py. If I do not add this test script, everything is fine. Once booted, soft reset works without problems.
The trigger does not have to occur. I can tie that input pin to GND and still the problem shows up.
Since I noticed the same behavior at the NRF52 board with this PR, the reason may be identical. People also mentioned problems with the MIMXRT port. I have to check.
Test script:
```
from machine import Pin
from time import sleep_ms, sleep_us

pin7 = Pin(7, Pin.OUT)

def callback(_, p=pin7):
    p(1)  # Trigger
    p(0)

pin3 = Pin(3, Pin.IN)
pin3.irq(callback, trigger=Pin.IRQ_FALLING)
sleep_ms(100)
pin3.irq(handler=None)
```

robert-hh:
The issue starts with the following commit:
`8809ae716 shared/tinyusb: Buffer startup CDC data to send to host on connection.`

andrewleech:
Thanks for the report and testing @robert-hh 
That's a shame, I thought I'd tested that change pretty thoroughly, looks like I missed some use cases though.
I'll try to make the to replicate your test and hopefully see where the problem occurs, might be related to how the tiny USB tx buffer switches from overwrite to blocking mode

robert-hh:
Some information about my test case: The above script is imported in main.py after having imported another python script. Sometimes it locks up, sometimes strange error messages pop up, like the one below, where importing the script should raise an error message. It was iperf3.py, which is not working at a plain RPI Pico. That all looks like having a uninitialized symbol.
```
ror: unexpected indent
>>> Tr
NameError: name 'Tr' isn't defined
>>> 
>>> Tr
NameError: name 'Tr' isn't defined
>>> 
>>>  
>>> Trac  acrTrac
  File "<stdin>", line 1
SyntaxError: invalid syntax
```

robert-hh:
It seems that it does not have to be that specific script. It often happens if the imported script raises an exception. Then, the output it want to create plus the usual boot banner get lengthy, like 674 bytes.

robert-hh:
It may have been a false alarm, at least to some extend. The problem is obvious only one one of my RP2040 Pico boards. All others behave fine. 
And the one misbehaving board does only when using Picocom. with mpremote things seems fine. And even with Picocom, if I close Picocom in the state when there is no response and open it again, communication works.

andrewleech:
Interesting, I wonder if picocom does something different with the port opening process.
The buffering stuff leans into a practice where detecting "port open" is done via DTR signal being asserted. Maybe picocom doesn't do this?
I haven't looked at the reports on discord yet either, wonder if that's related or not.

robert-hh:
Another strange observation:
I nuked the board which showed most of the phenomenon and copied back a small main,py and  the sample file. Then, the problem was gone! 
Then, I copied back like 40 more files to the board, and the issue as back again.

robert-hh:
With mprmote I get a REPL prompt, but with a part of the main.py being in the REPL input buffer, like:
```
ysh import *
  File "<stdin>", line 1
SyntaxError: invalid syntax
```
The main.py I'm using now for tests is:
```
from upysh import *
from pye import pye

from machine import Pin
import cs1237
power=Pin(13, Pin.OUT, value=1)
cs=cs1237.CS1237(Pin(15), Pin(14))
```
With upysh.py being from mircopython-lib, pye.py is my on-board editor, both being frozen in the firmware, and cs1237.py being the driver for a cs1237 ADC, at which I noticed the problem. attached. When starting it you must NOT attach the sensor. Without the sensor an error is raised. With the sensor, all is fine. Maybe a timing issue. In some tests, instead of importing that file a simple `time.sleep_ms(500)` was sufficient to cause that misbehavior. The cs1237 driver waits for ~500ms to get a response from the ADC on instantiation.

[cs1237.zip](https://github.com/user-attachments/files/15856577/cs1237.zip)

So it''s still kind of spooky. The problem may as well be somewhere in the file system or timing dependent.

andrewleech:
The "correct" behaviour of that new buffering change _is_ to to keep the most recent "full stdout" buffer worth of data from before a terminal is connected for display on-connection. 

So a failure in main that drops to repl should show the tail of the error message along with the repl prompt when you open a terminal. 

So yeah I'd expect that snippet of traceback, though probably not the content from main itself.... It would be interesting to increase the size of the buffer and see what else is shown then. This is defined in micropython/shared/tinyusb/tusb_config.h


The lockups seen on picocom are definitely an issue that warrants deeper investigation also.

robert-hh:
With Picocom, the REPL history buffer contains e.g.:

from upysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fiberoupysh impo Fibe

That looks very strange!

andrewleech:
Really strange! Was that after a soft reset or a hard boot? I'm wondering if there's some static buffers / memory pointers being reused without appropriate clearing on soft reset? I'm really not sure how the buffer change could do this either unless it's a buffer (mis) allocation issue that the explicit clear in tinyusb was hiding previously.
Though maybe my disabling the buffer clear in tinyusb is also relating to initial reboot, it might be missing a reset on the tx ringbuffer head and tail pointers on a cold boot now too.

Repl history buffer though seems like a separate issue to me, though might be a similar root cause. I don't think there should be any direct relation between repl history buffer and tinyusb tx buffer though.

robert-hh:
That all happens only after hard reset resp power cycle. Once it's working, soft reset behaves fine. It seems to be the initial set-up, which fails. And only if the start-up is delayed and/or an exception is raised by the code executed in main.py. May it be that some initialization only happens when input is requested from REPL? Or the opposite? It could be a timing effect.

projectgus:
I think the root cause is in the linked issue above (found by debugging @robert-hh's case above and then refining it.)

ports/nrf/modules/machine/uart.c
```diff
@@ -319,8 +320,17 @@ static mp_uint_t mp_machine_uart_read(mp_obj_t self_in, void *buf_in, mp_uint_t
 
 static mp_uint_t mp_machine_uart_write(mp_obj_t self_in, const void *buf_in, mp_uint_t size, int *errcode) {
     machine_uart_obj_t *self = self_in;
+    uint8_t *buf = (uint8_t *)buf_in;
+    #if !NRFX_UART_ENABLED
+    if (!nrfx_is_in_ram(buf_in)) {
+        // EasyDMA requires that transfer buffers are placed in DataRAM,
+        // NRFX_ERROR_INVALID_ADDR if the are not.
+        buf = alloca(size);
```
dpgeorge, line 328: This seems pretty dangerous, the stack may overflow if the buffer is large.

Maybe limit it to 256 bytes alloca, and do a loop over the buffer to break it up and send it in chunks?

ports/nrf/Makefile
```diff
@@ -259,6 +261,11 @@ SRC_C += $(addprefix lib/tinyusb/src/,\
 	portable/nordic/nrf5x/dcd_nrf5x.c \
 	)
 
+CFLAGS += \
+	-DCFG_TUSB_MCU=OPT_MCU_NRF5X \
```
dpgeorge, line 273: Instead of defining this at this level I suggest (this may not work but worth a try):
- don't add the `INC += -I../../shared/tinyusb` line above (this will prevent the default `tusb_config.h` from being included)
- add `INC += -I./drivers/usb`
- reuse the existing `drivers/usb/tusb_config.h` file and define this `CFG_TUSB_MCU` option there, and then do `#include "shared/tinyusb/tusb_config.h"` to get the remaining config from the common file

ports/nrf/boards/PCA10056/mpconfigboard.mk
```diff
@@ -2,6 +2,8 @@ MCU_SERIES = m4
 MCU_VARIANT = nrf52
 MCU_SUB_VARIANT = nrf52840
 SOFTDEV_VERSION = 6.1.1
+FLASHER ?= jlink
```
dpgeorge, line 5: Is this line needed, since jlink is the default flasher?

ports/nrf/mpconfigport.h
```diff
@@ -332,10 +349,25 @@ long unsigned int rng_generate_random_word(void);
...
-#include "device/usbd.h"
-#define MICROPY_HW_USBDEV_TASK_HOOK extern void tud_task(void); tud_task();
-#define MICROPY_EXCLUDE_SHARED_TINYUSB_USBD_CDC (1)
+#if MICROPY_HW_ENABLE_USBDEV
+#ifndef MICROPY_HW_USB_CDC
+#define MICROPY_HW_USB_CDC (1)
+#endif
+#endif
+
+#if MICROPY_HW_ENABLE_USBDEV
```
dpgeorge, line 358: Any reason why this extra #if is needed here, can't the config below be part of the block just above with the `MICROPY_HW_USB_CDC` config?

ports/nrf/mpconfigport.h
```diff
@@ -332,10 +349,25 @@ long unsigned int rng_generate_random_word(void);
...
+
+#ifndef MICROPY_HW_USB_VID
+#define MICROPY_HW_USB_VID  (0xf055)
+#endif
+#ifndef MICROPY_HW_USB_PID
+#define MICROPY_HW_USB_PID  (0x9802)
+#endif
+
+#ifndef MICROPY_HW_USB_CDC_TX_TIMEOUT
+#define MICROPY_HW_USB_CDC_TX_TIMEOUT (500)
```
dpgeorge, line 368: This is the default so I'd say just leave this define out.

ports/nrf/mphalport.h
```diff
@@ -43,6 +44,9 @@ typedef enum
 
 extern const unsigned char mp_hal_status_to_errno_table[4];
 
+extern int mp_interrupt_char;
```
dpgeorge, line 47: Please include `shared/runtime/interrupt_char.h` instead of declaring this variable.

ports/nrf/mphalport.c
```diff
@@ -50,6 +54,14 @@
 #include "soc/nrfx_coredep.h"
 #endif
 
+#ifndef MICROPY_HW_STDIN_BUFFER_LEN
+#define MICROPY_HW_STDIN_BUFFER_LEN 1024
```
dpgeorge, line 58: This seems large, especially since nrf parts don't usually have that much RAM.

Can we drop it down to 512 like rp2?

I know it's currently 1024 in master, but the new code in this PR uses `ringbuf_free(&stdin_ringbuf)` to never overflow this buffer, and also the CFG_TUD_CDC_RX_BUFSIZE/CFG_TUD_CDC_TX_BUFSIZE values are now 256 (in this PR) vs 64 (in master), so there's already more buffering.

ports/nrf/boards/PCA10056/mpconfigboard.mk
andrewleech: Ah, not anymore that's true.

ports/nrf/mpconfigport.h
andrewleech: Oh thanks, yes I'd moved the block above to that spot to consolidate the two, then forget the final cleanup.

dpgeorge:
@andrewleech I see you made changes.  Is this ready for re-review?

andrewleech:
Not quite sorry @dpgeorge I need to test them still.

ports/nrf/modules/machine/uart.c
andrewleech: Good catch! I had been thinking this was really just stuff like exception messages where it's sending const strings from the firmware. But looking forward, certainly when RomFS comes in there will be plenty of opportunity to send data directly from flash.
I've reworked this in a chunked send loop which was tested on a nrf52840 dev board (repl on uart).

ports/nrf/Makefile
andrewleech: Thanks, I had thought it was confusing to have two levels of this same header - but having the port one explicitely import the shared one does make it clearer for sure. It's certainly better to have this define in the header it belongs in rather than a global cflag.

I've made this change and yes it does work!

andrewleech:
@dpgeorge I've addressed your review feedback thanks and re-tested on the nrf52840 dongle and dev board.

I also tested USB without the poll after adding the `LDFLAGS += -Wl,--wrap=dcd_event_handler` and it appears to work correctly with just the interrupt (like on other ports) which is great!

dpgeorge:
Thanks for updating, the changes now look good.

I tested this on an `ARDUINO_NANO_33_BLE_SENSE` and it works well.  Tested:
- CDC upload and download speed and integrity
- mpremote cp files to and from the device over USB CDC
- run the test suite

ports/nrf/Makefile
```diff
@@ -128,7 +129,15 @@ CFLAGS_MCU_m4 = $(CFLAGS_CORTEX_M) -mtune=cortex-m4 -mcpu=cortex-m4 -mfpu=fpv4-s
 
 CFLAGS_MCU_m0 = $(CFLAGS_CORTEX_M) -fshort-enums -mtune=cortex-m0 -mcpu=cortex-m0 -mfloat-abi=soft
 
+# linker wrap does not work with lto on older gcc/binutils: https://sourceware.org/bugzilla/show_bug.cgi?id=24406
+GCC_VERSION = $(shell arm-none-eabi-gcc --version | sed -n -E 's:^arm.*([0-9]+\.[0-9]+\.[0-9]+).*$$:\1:p')
+GCC_MAJOR_VERS = $(word 1,$(subst ., ,$(GCC_VERSION)))
+ifeq ($(shell test $(GCC_MAJOR_VERS) -ge 10; echo $$?),0)
```
dpgeorge, line 135: @andrewleech this looks a bit funny to me.  On my Arch system I have:
```
arm-none-eabi-gcc (Arch Repository) 14.1.0
```
But `GCC_VERSION` here evaluates to `4.1.0`.

andrewleech: Ah drats, it worked on a few systems I'd tested on but yes my arch server is also at 14.1.0 and I get the same thing. Silly greedy regex. Easy fix...
`arm-none-eabi-gcc --version | sed -n -E 's:^arm.*([0-9]+\.[0-9]+\.[0-9]+).*$$:\1:p'` ->  
`arm-none-eabi-gcc --version | sed -n -E 's:^arm.* ([0-9]+\.[0-9]+\.[0-9]+).*$$:\1:p'`

andrewleech: https://github.com/micropython/micropython/pull/15553

I wish there was a way to do this in a proper programatic way, but I couldn't find one.

PR #15159, opened 2024-05-29
mpremote/transport_serial: Invert flow control settings on windows.

andrewleech:
The RTS/DTR lines are being overrided for windows to stop esp  chips from resetting unexpectedly. 

The settings can conflict with other ports (eg. stm32) if they are not  using the standard "Microsoft" usbser driver by blocking the port due to it handling flow control in the normal way.

More information:
https://docs.espressif.com/projects/esptool/en/latest/esp32/advanced-topics/boot-mode-selection.html#automatic-bootloader

image

https://dl.espressif.com/dl/schematics/ESP32-Core-Board-V2_sch.pdf

### DIFF
--- a/tools/mpremote/mpremote/transport_serial.py
+++ b/tools/mpremote/mpremote/transport_serial.py
@@ -84,10 +84,12 @@ class SerialTransport(Transport):
                     portinfo = list(serial.tools.list_ports.grep(device))  # type: ignore
                     if portinfo and portinfo[0].manufacturer != "Microsoft":
                         # ESP8266/ESP32 boards use RTS/CTS for flashing and boot mode selection.
-                        # DTR False: to avoid using the reset button will hang the MCU in bootloader mode
-                        # RTS False: to prevent pulses on rts on serial.close() that would POWERON_RESET an ESPxx
-                        self.serial.dtr = False  # DTR False = gpio0 High = Normal boot
-                        self.serial.rts = False  # RTS False = EN High = MCU enabled
+                        # When both lines are the same, no operation is performed on the esp.
+                        # Note: Both of these lines are "active low", so True here results on 0 on the line.
+                        # When DTR=True, RTS=False | DTR:0 / RTS:1 | IO0 is 0 (bootloader mode)
+                        # When DTR=False, RTS=True | DTR:1 / RTS:0 | EN is 0 (reset chip)
+                        self.serial.dtr = True
+                        self.serial.rts = True
                     self.serial.open()
                 else:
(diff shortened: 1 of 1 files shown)


dpgeorge:
Is it an option to just delete this Windows+ESP specific code, and leave the RTS/DTR lines alone?

Probably not, because then ESP boards will behave erratically (which is the reason these RTS/DTR settings were applied in the first place).

So maybe we should have command-line options to select the behaviour of these lines.  We could default to `True`/`True` (is in this PR), but then have an option to override that default.

andrewleech:
Yeah I threw this change up quickly partially as just a way to document:
* some learnings from the recent work I did with handling this in SW on the esp32s3 
* where / how this can cause conflicts.

The command line option is certainly most flexible, could add options there to actually use the lines to forcibly reset if desired.
The comment in the existing code leads me to think I should look deeping into windows pyserial code to see where the so called glitch comes from to see if that could be directly fixed to match linux behaviour and not need any special handling.

Josverl:
> Is it an option to just delete this Windows+ESP specific code, and leave the RTS/DTR lines alone?

I think that will leave most if not all users using Windows & ESP32, and certainly myself  in an unusable sparsely documented state with long or failing delays between soft resets ( the default mpremote behaviour after any command).
I agree the current code is quite ugly  but its the simplest I could find that actually works reliably across all ports and platforms I have access to.

There are no actual HIL tests covering the pyboard.py/mpremote behavior across ports and platforms, and I think that is a 

I have created and ran pytests test for these scenarios in the past against a range of different ports and platforms for connect after flash , soft reset and hard reset 
Even if they cannot be run in CI, then at least there is a consistent way to test, and assess the impact of proposed changes.
I can probably find these in the git history of some of my repos , or recreate them.

Is there interest in such tests ?

andrewleech:
Thanks @Josverl yes I've since realised that True / True essentially is the default in pyserial anyway, so certainly isn't the right setting.

Even though I think it "should" be fine for the Espressif schematic shown (plus the normally recommended cap across lines) I guess in practice lots of boards are slightly different in behaviour? 

Regardless, unless I could figure out a way deeper in pyserial to fix the glitches on the lines causing the problems on windows, it does look like this probably needs to stay unchanged here in mpremote

Josverl:
I'll create a PR for the mpremote tests and put them in `tools\mpremote\tests` with a quick readme.md and requirements-tests.txt.
will need to wait a week or so - some of them are here : https://github.com/Josverl/mpremote/tree/main/tests

PR #15179, opened 2024-05-31
rp2/modmachine: Use atomic section macros

drowfx:
To avoid undefined references to
mp_thread_begin_atomic_section/mp_thread_end_atomic_section, replace them with the MICROPY_BEGIN_ATOMIC_SECTION / MICROPY_END_ATOMIC_SECTION macros. That way, the support for building with MICROPY_PY_THREAD disabled introduced by commit efa54c27b9ea ("rp2/mpconfigport: Allow MICROPY_PY_THREAD to be disabled by a board.") is useable again.

Fixes commit 19844b498306 ("rp2/modmachine: Prevent lock-up when lightsleep() called within thread.").


dpgeorge:
Thanks for the fix!

PR #15189, opened 2024-06-02
shared/tinyusb: Allow ports to use 1200bps-touch without other CDC code.

dpgeorge:
This fixes the build for some esp32 and nrf boards (for example `ARDUINO_NANO_33_BLE_SENSE` and `ARDUINO_NANO_ESP32`) due to commit c98789a6d8e05acb608afe4b30cf3ca563419b2d.  Changes are:

- Allow the CDC TX/RX functions in `mp_usbd_cdc.c` to be enabled separately to those needed for `MICROPY_HW_USB_CDC_1200BPS_TOUCH`.

- Add `MICROPY_EXCLUDE_SHARED_TINYUSB_USBD_CDC` option as a temporary workaround for the nrf port to use.

- Declare `mp_usbd_line_state_cb()` in a header as a public function.

- Fix warning with type cast of `.callback_line_state_changed`.

### DIFF
--- a/ports/esp32/boards/ARDUINO_NANO_ESP32/board_init.c
+++ b/ports/esp32/boards/ARDUINO_NANO_ESP32/board_init.c
@@ -26,6 +26,7 @@
 
 #include <string.h>
 #include "py/mphal.h"
+#include "shared/tinyusb/mp_usbd_cdc.h"
 
 #include <esp_system.h>
 #include <esp_ota_ops.h>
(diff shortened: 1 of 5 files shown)


projectgus:
@dpgeorge Would this have been caught pre-merge if we added one or two more boards in CI?

dpgeorge:
> Would this have been caught pre-merge if we added one or two more boards in CI?

Yes.  But there are 160 boards that we support (not including variants) and not all can be built for the CI...

Probably though it would be worth adding one of the ones that failed here.  Preferably one that's quick to build and doesn't slow down the CI more.

PR #12540, opened 2023-09-28
ports/stm32/boards: Add WEACT_STM32H743 board.

mattytrentini:
Adds a board definition for the [WeAct Studio STM32H743](https://github.com/WeActStudio/MiniSTM32H7xx/tree/master#stm32h743vit6-2048kb-rom-1mb-ram) board.

PR #15191, opened 2024-06-02
Add DEVEBOX_H750 Board

ArduinoTM:


robert-hh:
It is not clear for me for which MCU the PR was intended. You use both STM32H743VIT6 and STM32H750VB. The H750VB's flash is not large enough for the MicroPython firmware. So I assume you worked with the H743VI. Then you should use this name consistently. 
In board.json the URL of the board vendor must be more specific. It should point to a web site with detailed technical information for this board.
About a PR for a STM32H743 board, please have a look at PR #12540.

PR #15197, opened 2024-06-03
tools/metrics.py: Change rp2 board selection to RPI_PICO_W.

dpgeorge:
This tool is used to compute size differences in the firmware (eg as part of CI), but it doesn't currently check any firmware that has bare-metal lwIP/networking, making it hard to see how firmware size changes when networking related changes are made.

So, change the board selection for the rp2 port to RPI_PICO_W.  Changes in size to standard RPI_PICO firmware will be very similar to other bare-metal boards like PYBV10.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -55,7 +55,7 @@ function ci_code_size_setup {
 function ci_code_size_build {
     # check the following ports for the change in their code size
     PORTS_TO_CHECK=bmusxpd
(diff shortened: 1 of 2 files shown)


dpgeorge:
The correct code-size-diff output won't work until this is merged to master (because it's trying to compare previous `RPI_PICO` builds with `RPI_PICO_W`).

Issue #14204, opened 2024-03-28
esp8266: Error OSError: -40 https request


andreyu-git:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

esp-8266

### MicroPython version

MicroPython v1.22.2 on 2024-02-22; ESP module with ESP8266

### Reproduction
```

import esp

import network
import urequests as requests

esp.osdebug(0)

wlan = network.WLAN(network.STA_IF)
wlan.active(True)

if not wlan.isconnected():
    print('connecting to network...')
    wlan.connect('RT166', 'mypass')

    while not wlan.isconnected():
        pass

print('network config:', wlan.ifconfig()[0])

bot_token = 'mytoken'

res = requests.get(url=f'https://api.telegram.org/bot{bot_token}/sendMessage', data={
    chat_id: '000000',
    text: 'hello'
})

print(res.text)
```

### Observed behaviour

```
Traceback (most recent call last):
  File "<stdin>", line 26, in <module>
  File "requests/__init__.py", line 184, in post
  File "requests/__init__.py", line 93, in request
OSError: -40
```

### Additional Information

Hello!

I'm trying to make an http request to the telegram api and I get the OSError error: -40

I searched for information on the Internet, but I never found a solution!

Please tell me how to fix this problem!

dpgeorge:
The site you are trying to access, `api.telegram.org`, requires the client to support a very modern TLS (https) protocols.  The protocols that the site supports are:
```
$ nmap --script ssl-enum-ciphers -p 443 api.telegram.org
...

PORT    STATE SERVICE
443/tcp open  https
| ssl-enum-ciphers: 
|   TLSv1.2: 
|     ciphers: 
|       TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
|       TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519) - A
|       TLS_DHE_RSA_WITH_AES_128_GCM_SHA256 (dh 2048) - A
[...]

Issue #10485, opened 2023-01-12
Enable MBEDTLS_GCM_C by default


sylvinus:
Hello,

Today on `master`, the UNIX port (among others I haven't tested) is not able to connect to most TLS endpoints because it lacks support for [GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode).

Indeed, GCM is required for what seem to be by far the 4 most popular Cipher Suites on the web currently:
https://scotthelme.co.uk/top-1-million-analysis-june-2022/

This can be easily tested by this script: ( [SSLLabs for reference](https://www.ssllabs.com/ssltest/analyze.html?d=en.wikipedia.org&s=198.35.26.96&latest) )
```
import ussl
import socket

ai = socket.getaddrinfo("en.wikipedia.org", 443)
s = socket.socket(ai[-1][0])
s.connect(ai[-1][-1])
s = ussl.wrap_socket(s, server_hostname="en.wikipedia.org")
s.write(b"GET / HTTP/1.0\r\nHost: en.wikipedia.org\r\n\r\n")
print(s.read(1024))
s.close()
```

This will fail the handshake with `OSError: (-30592, 'MBEDTLS_ERR_SSL_FATAL_ALERT_MESSAGE')` unless `MBEDTLS_GCM_C` is defined in `extmod/mbedtls/mbedtls_config_common.h`.

For some reason, it seems to be enabled only in the `rp2` port.

I know there will be some consequences on the flash size, but if micropython ships with TLS support it makes sense to support today's web. Actually, there could probably be a few flags for known weak ciphers to be disabled if we want to recoup the increase due to enabling GCM!

Thanks!

dpgeorge:
OK, we probably should enable `MBEDTLS_GCM_C` on all ports that use mbedTLS.  That will make them all consistent (with rp2) and be able to access more sites.

> Actually, there could probably be a few flags for known weak ciphers to be disabled

This was already done recently, see 68f166dae9ad6dfd94038d5f4394defbb44238af and previous commit.

If you know of any others that can be disabled, please comment.

sylvinus:
@dpgeorge great! How to feel confident about disabling more? Should there be an automated test on the top N websites? Probably TLS is also used to connect to other types of endpoints (APIs?) we must continue to support?

dpgeorge:
> How to feel confident about disabling more? Should there be an automated test on the top N websites?

[...]

PR #15211, opened 2024-06-05
extmod/mbedtls: Enable modern cipher suites

sylvinus:
Following the comments in https://github.com/micropython/micropython/issues/14204#issuecomment-2024366349 and https://github.com/micropython/micropython/issues/10485#issuecomment-1396426824 , I've run some tests on the [top 500 domains from moz.com](https://moz.com/top500).

Without this patch, 155 out of 500 fail to connect because of TLS issues. This patch fixes them all!

One of the suites it enables is TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, which is actually required by hosts like bloomberg.com for TLS1.2:
https://www.ssllabs.com/ssltest/analyze.html?d=bloomberg.com&s=15.197.146.156&hideResults=on&ignoreMismatch=on

I tried disabling some other flags but it seems all of them are needed to get good coverage of those top 500 domains.


dpgeorge:
Thanks for the contribution!  Very nice to see your approach of testing the top 500 domains.

Although code size will increase quite a lot with these additional ciphers, I think it's unavoidable.  MicroPython needs to be able to talk to sites on the internet.

Also, the rp2 port already enables both of these options, so this PR brings all ports together to have the same mbedtls options enabled.

Can you please then remove the two now-redundant `#define` lines from `ports/rp2/mbedtls/mbedtls_config_port.h`?

dpgeorge:
> Can you please then remove the two now-redundant `#define` lines from `ports/rp2/mbedtls/mbedtls_config_port.h`?

I have now done this.

I also updated the `ssl_poll.py` test so it works again, by regenerating the key/cert with a higher number of bits.

dave4jr:
@dpgeorge Hi Damien, sorry to ask here but can't find it elsewhere. I need this update on the ESP32 so I can send using HTTPS to our external API. How can I load the nightly or this master branch onto the ESP32? I downloaded it but it's 10mb whereas the 1.23 release is much larger (I believe without submodules). Will that mean I will be losing out on having key functionality / other modules if i load up from master directly? Any help would be super appreciated! Thank you!

dpgeorge:
> I need this update on the ESP32 so I can send using HTTPS to our external API.

@dave4jr The ESP32 port uses a different configuration to this PR.  The ESP32 already had these mbedTLS options enabled for some time, so release v1.23.0 should already be able to access HTTPS sites using these ciphers.

dave4jr:
Ahhh okay cool.. Thank you very much for the quick response @dpgeorge !

PR #15217, opened 2024-06-06
py/lexer: Support concatenation of adjacent f-strings.

dpgeorge:
This is quite a simple and small change to support concatenation of adjacent f-strings, and improve compatibility with CPython.

### DIFF
--- a/tests/basics/string_fstring.py
+++ b/tests/basics/string_fstring.py
@@ -65,3 +65,13 @@ print(f"{a!s:8s}")
 # Still allow ! in expressions.
 print(f"{'1' if a != '456' else '0'!r:8s}")
 print(f"{'1' if a != '456' else '0'!s:8s}")
+
+# Concatenation of adjacent f-strings.
+print(f"" f"")
+print(f"a" f"b")
+print(f"{x}" f"{y}")
+print(
+    f"a{x}b---------------------------------"
+    f"cd---------------------------------"
+    f"e{y}f---------------------------------"
+)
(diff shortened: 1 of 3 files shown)


kdschlosser:
You are the man. Thanks for adding this as it has been something I have been tripping over from time to time. I do like to keep my code as close to pep8 compliant as possible and one of the things I am really a stickler about is keeping the line length to 80 characters. Not being able to concatenate the f strings added more to my code because I would have to make a list and the join the list together in order to keep the line length at <= 80.

Glad it was an easy fix.

tests/basics/string_fstring.py
```diff
@@ -65,3 +65,9 @@ def foo(a, b):
 # Still allow ! in expressions.
 print(f"{'1' if a != '456' else '0'!r:8s}")
 print(f"{'1' if a != '456' else '0'!s:8s}")
+
+# Concatenation of adjacent f-strings.
+print(f"" f"")
+print(f"a" f"b")
+print(f"{x}" f"{y}")
+print(f"a{x}b" f"cd" f"e{y}f")
```
kdschlosser, line 73: I would add one additional test to this. 

```python
print(
    f"a{x}b" 
    f"cd" 
    f"e{y}f"
)
```

This is a useful thing to do if wanting to keep code pep8 compliant, under the 80 character width for a line.
Want to make sure that the lexer properly handles the line breaks as well.

dpgeorge: Good idea.  I've now done that (although slightly different to the way you suggested).

kdschlosser: I thought it could possibly be a failure mode if something went screwy with the lexer.. Can't hurt to have a test for it.

question on the tests.. Does it check what actually gets output to make sure nothing got truncated or added during the concatenation?

dpgeorge: > Does it check what actually gets output to make sure nothing got truncated or added during the concatenation?

Yes.  It runs the same test with CPython and makes sure the output from that matches the output from MicroPython (unless there's a corresponding .py.exp file which is used as the expected output instead of running CPython).

PR #15221, opened 2024-06-06
tools: Ensure codeformat.py uses uncrustify 0.72.

Gadgetoid:
As identified in https://github.com/micropython/micropython/pull/14046 it's really easy to install a mismatched version of Uncrustify and format source in a way that the CI (which just installs whatever mystery meat Uncrustify apt cares to offer) disagrees with.

Eg: when I develop on macOS, brew installs Uncrustify-0.79.0_f.

This may also vary across Linux distributions, and is also prone to some future break if `runs-on: ubuntu-22.04` is ever changed.

This change is a proposal for installing Uncrustify in a consistent fashion, so that developers and CI get the same version, the same rules and the same output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ user.props
 
 # MacOS desktop metadata files
 .DS_Store
+
+# CI Tools
+tools/uncrustify
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -21,8 +21,11 @@ function ci_gcc_arm_setup {
 # c code formatting
 
 function ci_c_code_formatting_setup {
-    sudo apt-get install uncrustify
-    uncrustify --version
+  wget https://github.com/uncrustify/uncrustify/archive/refs/tags/uncrustify-0.72.0.tar.gz
+  tar -xf uncrustify-0.72.0.tar.gz
+  cmake -B tools/uncrustify -S uncrustify-uncrustify-0.72.0
+  cmake --build tools/uncrustify -j
+  ./tools/uncrustify/uncrustify --version
 }
 
 function ci_c_code_formatting_run {
--- a/tools/codeformat.py
+++ b/tools/codeformat.py
@@ -171,7 +171,7 @@ def main():
 
     # Format C files with uncrustify.
     if format_c:
-        command = ["uncrustify", "-c", UNCRUSTIFY_CFG, "-lC", "--no-backup"]
+        command = ["./tools/uncrustify/uncrustify", "-c", UNCRUSTIFY_CFG, "-lC", "--no-backup"]
         if not args.v:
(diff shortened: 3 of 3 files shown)


Gadgetoid:
Note: It may be prudent to try the "apt install uncrustify" step first and only install from source if there's a version mismatch. This would avoid any performance impact both on CI and to developers who already have the correct version.

dhalbert:
@dlech just built a pypi package that installs the proper version:
https://pypi.org/project/micropython-uncrustify/

See https://github.com/orgs/micropython/discussions/14496

dlech:
FYI, I recently published https://pypi.org/project/micropython-uncrustify/ to have an easy way to get a pre-compiled binary with the correct version of uncrustify. This would be more efficient than everyone having to compile it themselves.

Gadgetoid:
Haha, now that's an approach. D'oh!

Downside is that you probably now have to contend with cursed virtualenvs somehow 🫠  ... but that's not my problem *runs away*

Happy to close this and defer to whatever solution you're cookin' up! I got a little carried away 😆

Gadgetoid:
Hmmm, well it *works* -

```
brew install virtualenvwrapper pyenv-virtualenvwrapper
source virtualenvwrapper.sh
mkvirtualenv micropython
pip install micropython-uncrustify
~/.virtualenvs/micropython/bin/uncrustify --version 
Uncrustify-f44f717
```

And bonus points for Windows builds!

I guess it's supposed to huck it into /bin if I do naughty things with system python?

dlech:
You can also run it like this if you don't want to make a vritualenv or pollute your PATH. 

    pipx run --spec micropython-uncrustify uncrustify --version

Gadgetoid:
`pipx` aha, that's the one people keep telling me about that my brain never retains. TIL, thank you.

Though, uh, remind me not to try and `brew install pipx` again... kicked off some kind of dependency chain chaos involving ffmpeg.

projectgus:
Oh nice work, @dlech! I mentioned in the linked issue that I did some work getting uncrustify to build under wasm, but it's even better if you've got it building native binaries for all of the relevant architectures.

> Downside is that you probably now have to contend with cursed virtualenvs somehow 🫠 ... but that's not my problem runs away

pre-commit manages its own sandbox Python environments, from what I can tell, and it seems pretty seamless. So I think what we can probably do is make a pre-commit hook that installs and runs the correct micropython-uncrustify version and that should be pretty seamless for most users.

dpgeorge:
I'm going to close this.  The pipx solution https://github.com/micropython/micropython/pull/15221#issuecomment-2153556941 or just `uv pip install micropython-uncrustify` in a venv is a great way to get the right version.

PR #15231, opened 2024-06-08
py/asmrv32: Add a RISC-V RV32IMC native code emitter.

agatti:
This PR adds a native code emitter backend for RV32I CPUs that also support the M extension (multiplication/division).

Right now native and viper tests pass on both `qemu-riscv` and on ESP32-C3 boards.  For `qemu-riscv`, all excluded entries in `tests_profile.txt` should be removed to enable native and viper tests, and `MICROPY_EMIT_RV32` must be enabled.

For ESP32-C3, `ports/esp32/mpconfigport.h` must have `MICROPY_EMIT_RV32` enabled and `MICROPY_NLR_SETJMP` disabled.  Then tests can be run via `run-tests.py --target esp32 --device <usb device name>`

<hr>

The big thing missing from this PR is compressed instructions support.  This stems from the fact that generated code blocks cannot change length after the code size computation pass (more details [here](https://github.com/orgs/micropython/discussions/15212)).  To try to solve that I've tried a couple of things in the meantime.

* Having a second `MP_ASM_PASS_EMIT` pass to fix up labels (changing the ASM API to make `ASM_END_PASS` return a boolean value and forwarding that to the compiler backend to reschedule passes) didn't work.
* Then I've added a new compilation pass between `MP_PASS_CODE_SIZE` and `MP_PASS_EMIT` in both the native emitter and the asm base emitter, where the labels would be fixed up before the final emit pass and the assertion would check size difference between this pass and `MP_PASS_EMIT`.  That made it work in all but three tests, and I wasn't able to figure out why that happened even though everything looked fine.

I haven't had the time to try to add an extra compilation pass after MP_PASS_EMIT/MP_ASM_PASS_EMIT to fix up labels and re-do the emit process, but I'm not sure that would work.

This brings us back to the lack of compressed instructions.  Those instructions only complement part of the instruction set, so depending on which values/addresses are evaluated then they may or may not be used by the emitter.  This would only complicate keeping generated code sequences the same length in every case (in fact sequences may also have different alignment boundaries depending on the input values), so I'd love to see how that issue can be worked around.


ricksorensen:
Thanks 

Initial testing with my XIAO ESP32C3 and (@peterhinch) Peter Hinch's [https://github.com/peterhinch/micropython-nano-gui](https://github.com/peterhinch/micropython-nano-gui) using the default `st7789_4bit` driver with `viper` worked great.  I just ran a few examples, but will try more.   My initial impression is the  speed is getting closer to that with (@russhughes) Russ Hughes compiled module, but I have not yet done accurate timing yet.

agatti:
@ricksorensen thanks for trying it out!

Consider that right now generated code may be executing at 2/3rd of its potential maximum speed or even less (see the discussion post I linked in the PR description why).  Moreover there are more than ten registers that are currently unused, and I guess they could be repurposed to hold locals instead.

I plan to address the latter point once this gets more testing and is hopefully part of MicroPython and enabled by default.

dpgeorge:
This is fantastic!  Thanks for working on it.  (I was going to ask if you were working on this feature, and if not I would have started it myself.)

There are many possible subsets of RV32 that the native emitter could target, based on all the extensions to RV32.  Discussing with @projectgus we thought that targetting RV32IMC would be a good place to start, as that seems to be a pretty common set of extensions (I - full registers, M - mult/div, C -compressed instructions) and allows the emitter to use some relatively powerful instructions to get good code generation.

(Eg we could go fully minimal and only support RV32E (reduced registers) but IMO that's not necessary at this point and too restrictive.)

So I think what you did here with RV32IM is good, and it would be great to extend  it with compressed instructions where possible.  A good thing with RV32 vs Thumb2 is that RV32-C supports both standard and compressed instructions at the same time (whereas Thumb2 has exclusive 32-bit or 16-bit modes).  So that means it's easy to fall back to the 32-bit instructions where needed, and also allows gradually adding support for more and more compressed instructions in a backwards compatible way.

Also, chips with compressed support must allow all instructions to start on a 16-bit alignment boundary (as far as I understand), which puts less restrictions on the code generator.

mpy-cross/main.c
```diff
@@ -312,6 +313,9 @@ MP_NOINLINE int main_(int argc, char **argv) {
                 } else if (strcmp(arch, "xtensawin") == 0) {
                     mp_dynamic_compiler.native_arch = MP_NATIVE_ARCH_XTENSAWIN;
                     mp_dynamic_compiler.nlr_buf_num_regs = MICROPY_NLR_NUM_REGS_XTENSAWIN;
+                } else if (strcmp(arch, "rv32i") == 0) {
+                    mp_dynamic_compiler.native_arch = MP_NATIVE_ARCH_RV32I;
+                    mp_dynamic_compiler.nlr_buf_num_regs = MICROPY_NLR_NUM_REGS_RV32I;
```
dpgeorge, line 318: I would call the architecture `rv32imc` to indicate the extensions required.  This will allow room to grow in the future if we also need to support strict `rv32i` without extensions.

dpgeorge:
Regarding compressed instruction support: for now I would think that it should be possible and easy to emit compressed instructions for those cases where it's guaranteed that the instruction can always be encoded in the C version.

For example this bit:
```c
define ASM_MOV_REG_REG(state, rd, rs) asm_rv32_opcode_add(state, rd, ASM_RV32_REG_ZERO, rs)
```
Surely that can just emit `c.mv rd, rs` instead of `add rd, x0, rs`?  So it could be:
```c
#define ASM_MOV_REG_REG(state, rd, rs) asm_rv32_opcode_cmv(state, rd, rs)
```

Similarly for other instructions like `c.add`, `c.jr`, `c.jalr`.

But also note that emitters like Thumb2 output instructions of varying size, and don't have any problem with determining the instruction size early on in the set of native compilation passes.  Eg `ASM_MOV_REG_IMM` calls `asm_thumb_mov_reg_i32_optimised()` which can output 2 bytes, 4 bytes, 6 bytes or 8 bytes.  I don't see why the RV32I-compressed instructions would be that different to require additional passes?

agatti:
[Replying with a single comment to both points]

The original emitter was written last year, and was then rewritten from scratch a few weeks ago due to an SSD failure and me not having pushed the final code version to my fork.  Still, thanks for taking a look at this!

That said, I agree that the IMC extensions set is the one to target, and that was my original goal indeed.  I also planned to add inline assembler support (I do have a prototype for IM but need to sort a few things out first) for those instructions, including the Zicr extension (special registers access).  The C extension adds a bit more complexity when encoding certain instructions as there are register choice limitations, like Thumb1 mode not being able to easily access R8 and up for example.  I'll add C extension support, one way or another :)

Since we're discussing extensions, would it make sense to have an extmod module that exposes the extensions list to Python code?  I can submit a new PR for such a thing that's independent of this PR if so.

> But also note that emitters like Thumb2 output instructions of varying size, and don't have any problem with determining the instruction size early on in the set of native compilation passes. Eg `ASM_MOV_REG_IMM` calls `asm_thumb_mov_reg_i32_optimised()` which can output 2 bytes, 4 bytes, 6 bytes or 8 bytes. I don't see why the RV32I-compressed instructions would be that different to require additional passes?

Unless I'm mistaken, Thumb2 always outputs sequences of the same length when handling labels.  Not so on RV32, or at least in my implementation.  Considering that immediates that are in the ±2Kbytes range can be encoded as part of the opcode and are always treated as signed integers, here is what happens:

- On non-emit asm passes, label targets are not final and may point to 0xFFFFFFFF.  Before the changes I just pushed, code that emitted conditional or unconditional jumps would always encode an offset of -1 with a multi-opcode sequence.  I've since added a special case for loading immediates when the value is between -2048 and 0, shortening the generated sequence.
- On the emit asm pass, label targets are known (or maybe can change when the stack isn't settled and code gets rewritten?), and have offsets that can be usually encoded in a single-opcode sequence.  That may invalidate already computed label targets that point past the code offset in question if a longer code sequence was used previous computation passes to figure out where the label should point to.

Most native functions are small and label offsets stay within ±2Kbytes range so they can be always be encoded in a single opcode, making NOPs unnecessary.  However, the moment said assumption does not hold true anymore, I ended up triggering the assertion at py/asmbase.c:93 (I always built the test firmware with DEBUG=1).  Hence my attempts to perform manual label fixup by moving label targets up or down depending on what fixup offset the current code sequence applied (you can still see remnants of those attempts, `fixup_block`).  If that assertion can be safely ignored I can add its removal to this PR.

mpy-cross/main.c
agatti: Makes sense, although I wonder whether there are recently produced MCUs that don't have the M extension present.  Except soft-cores, that is, but that doesn't really count, does it?

Anyway, I'll add the platform renaming to the new batch of commits.

dpgeorge: You may be right that everything has M extension support, and probably also C.  But at this stage (early stage for RISC-V lifetime) it's hard to tell and so I'd go conservative and spell it out as `rv32imc`.

And supporting soft-cores is definitely in scope for MicroPython.

dpgeorge:
> Since we're discussing extensions, would it make sense to have an extmod module that exposes the extensions list to Python code?

How about the existing `platform` module in `extmod/modplatform.c` for that?

> Unless I'm mistaken, Thumb2 always outputs sequences of the same length when handling labels

Yes.  The logic is (schematically):
```c
void asm_thumb_b_label(asm_thumb_t *as, uint label) {
    mp_uint_t dest = get_label_dest(as, label);
    mp_int_t rel = dest - as->base.code_offset;

    if (dest != (mp_uint_t)-1 && rel <= 0) {
        // is a backwards jump, so we know the size of the jump on the first pass
        if (SIGNED_FIT12(rel)) {
            // emit 16-bit short jump
            return;
        }
    }

    // is a large backwards jump, or a forwards jump (that must be assumed large)
    // emit 32-bit long jump
}
```

And it sounds like you've now got similar logic for the RV32 generation?

----

But what do you think about my comment:

> Surely that can just emit c.mv rd, rs instead of add rd, x0, rs?

py/asmrv32.c
```diff
@@ -0,0 +1,761 @@
...
+    ((op & 0b1111111) | ((imm & 0b11111) << 7) | ((ft3 & 0b111) << 12) |         \
+    ((rs1 & 0b11111) << 15) | ((rs2 & 0b11111) << 20) |                         \
+    ((imm & 0b111111100000) << 20))
+
+#define RV32_ENCODE_TYPE_U(op, rd, imm)                                        \
+    ((op & 0b1111111) | ((rd & 0b11111) << 7) |                                  \
+    (imm & 0b11111111111111111111000000000000))
+
+// ADD RD, RS1, RS2
+void asm_rv32_opcode_add(asm_rv32_t *state, mp_uint_t rd, mp_uint_t rs1, mp_uint_t rs2) {
```
dpgeorge, line 106: I would think that all of these functions could be made `static inline` in the header file, and it would result in smaller code.  You can see that other emitters like `asmxtensa.h` have them inline.

But that would only be true if the `saved_registers_mask` was either removed (and a different way found to work out which registers to save), or made more general, eg the `asm_rv32_emit_instruction()` function may be able to tell which register is written to in a general way, and then mark it as used.

py/asmrv32.h
```diff
@@ -0,0 +1,291 @@
...
+#define ASM_RV32_REG_S8 (ASM_RV32_REG_X24)
+#define ASM_RV32_REG_S9 (ASM_RV32_REG_X25)
+#define ASM_RV32_REG_S10 (ASM_RV32_REG_X26)
+#define ASM_RV32_REG_S11 (ASM_RV32_REG_X27)
+
+typedef struct _asm_rv32_t {
+    // Opaque emitter state.
+    mp_asm_base_t base;
+    // Which registers are tainted and need saving/restoring.
+    mp_uint_t saved_registers_mask;
```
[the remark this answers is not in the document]
agatti: This was a cheap method to minimise the number of stack push and pops (8 bytes each right now, probably less with C opcodes), there are a couple more registers that can be used or not that are not part of the REG_LOCALs set (for example RA/x1 if a jalr opcode is emitted and S1/x9 for handling large immediates without clobbering REG_TEMPs).  For some small functions, saving and restoring all potentially used registers would take more code than the function body itself.

I can take this out and keep the same tracking in place, see the other review comment.

py/asmrv32.c
agatti: Luckily RV32 opcodes are quite regular when it comes to register locations within opcodes and their encoding.  To remove the register usage tracking in the opcode functions I can have something like `asm_rv32_emit_instruction(asm_rv32_t *emit, uint32_t opcode, rv32_opcode_kind_t kind)` where the latter is just an enumeration indicating which encoding to follow.

agatti:
> > Since we're discussing extensions, would it make sense to have an extmod module that exposes the extensions list to Python code?
> 
> How about the existing `platform` module in `extmod/modplatform.c` for that?

Sure, `platform.cpu_extensions()` or `.riscv_extensions()` returning a tuple like ('C', 'I', etc.) would fit in there.

> > Unless I'm mistaken, Thumb2 always outputs sequences of the same length when handling labels
> 
> Yes. The logic is (schematically):
> [...]
> And it sounds like you've now got similar logic for the RV32 generation?

I do, but I then I wonder what would cause that assertion to trigger.  Oh well, I'll take out the NOPs then.

> But what do you think about my comment:
> 
> > Surely that can just emit c.mv rd, rs instead of add rd, x0, rs?

Sure, that's doable.  Will take this into account, thank you!

agatti:
A subset of the C instruction set is now implemented, and native tests pass on QEMU and ESP32C3.

I'll look into implementing the opcodes that have limitations on which registers they can operate upon, as I'll need to fill up a table to see which registers `emitnative.c` uses for code sequences involving those opcodes.

Please ignore the \[fixup ...\] bits in the Git commit messages, it's for my scripts to rearrange history once things are ready to be merged, while keeping commits small enough to be looked at as discrete units.

agatti:
@dpgeorge I've bumped into the misaligned code assertion at asmbase.c:93 again.  In this bit of code:

```c
void asm_rv32_emit_mov_reg_pcrel(asm_rv32_t *state, mp_uint_t rd, mp_uint_t label) {
    ptrdiff_t displacement = (ptrdiff_t)(state->base.label_offsets[label] - state->base.code_offset);
    mp_uint_t upper = 0;
    mp_uint_t lower = 0;
    split_immediate(displacement, &upper, &lower);

    // auipc rd, HI(relative)
    asm_rv32_opcode_auipc(state, rd, upper);
    if (lower != 0) {
		if (SIGNED_FIT6(lower)) {
			// c.addi rd, LO(relative)
			asm_rv32_opcode_caddi(state, rd, lower);
		} else {
			// addi rd, rd, LO(relative)
			asm_rv32_opcode_addi(state, rd, rd, lower);
		}
    }
}
```

Removing the half-word `c.addi` opcode makes it work, same as adding a `c.nop` after the `c.addi` opcode to keep things aligned.  Is there anything I should look for in the debugger since I can reproduce this at will now?

agatti:
This should be version of the code ready for review.  This attempts to use compressed opcodes as much as possible, except for `asm_rv32_emit_mov_reg_pcrel`.  Having code sequences there that may change the emitted code size between passes trips the assertion at asmbase.c:93.  This happens on different compilers, and is also reproducible on the CI machine environment.

Some more code sequences have been tightened up, some more could take fewer bytes if some restrictions on the registers being used are applied.  Unfortunately the most interesting C opcodes have a window of 8 registers, and it cannot fit all the ones used unless some assumptions have to be made.  That may be the subject of further work, but right now this is more or less as good as it can get without having to add conditionals everywhere checking which registers are being operated upon.

agatti:
Just one minor note, `mpy_ld.py` support for RISC-V is currently missing (I've incorrectly updated `natmod.rst` saying the architecture was supported - now fixed).  The way RISC-V handle relocations may involve having potential entries whose appearance should modify the behaviour of the immediately preceding entry, altering its semantics.  Not sure if Xtensa has a similar mechanism, but as far as I remember neither Arm nor x86/x64 has anything similar - it's been a while.

Since the changes needed may be a bit too involved for a PR like this, I'd rather have the changes in `mpy_ld.py` submitted as their own PR if that's fine with you.

ricksorensen:
For release, will the `esp32/mpconfigport.h` be updated to automatically include the RISC-V emitter for those MCUs that support it?

agatti:
Hrm, this problem only occurs on the gcc version shipped with Ubuntu 22.04.  On both my own toolchain and the one provided by Espressif it works.  Weird...

agatti:
When building in release mode the compiler made a few assumptions regarding registers usage and rearranged things around a bit.  Mapping an internal temporary register to another index fixed the issue, making registers usage tracking not as effective as it was supposed to be (it's been removed).

Anyway, now things work as expected on both Qemu and on ESP32, in both debug and release mode.

py/asmrv32.h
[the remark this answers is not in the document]
agatti: The tracking was removed in 4d0c3f964c2f8fb6bb3d5eaff95b989bf6e1e7f8 so this doesn't apply anymore.

py/asmrv32.c
agatti: The tracking was removed in 4d0c3f964c2f8fb6bb3d5eaff95b989bf6e1e7f8 so this doesn't apply anymore.

dpgeorge:
Now merged!  Thanks @agatti for your efforts on this.

PR #15237, opened 2024-06-09
nrf/modules/machine/pin.c: Disable the IRQ with pin.irq(handler=None).

robert-hh:
Before, the input was still set to pin.irq() mode, only the handler was disabled. That prevented switching the Pin between input and output mode.


dpgeorge:
Thanks, the change looks correct.

PR #15238, opened 2024-06-09
mimxrt: Bug fixes, small changes and additions.

robert-hh:
Some small changes and bug fixes.

- Change the RAM Pin configuration and fix the comment for the MIMXRT1176 port.
- Clear pending IRQ requests when Pin IRQ is enabled or disabled. Otherwise the IRQ handler may be called prematurely.

### DIFF
--- a/docs/mimxrt/quickref.rst
+++ b/docs/mimxrt/quickref.rst
@@ -438,6 +438,10 @@ The i.MXRT MCU supports battery backup of the RTC.  By connecting a battery of
 current drawn from the battery is ~20µA, which is rather high.  A CR2032 coin
 cell will last for about one year.
 
+Note: In v1.23.0 the support for subseconds was removed. When reading the RTC, 0 will
+be returned as value for subsecond, When setting the RTC time, the subsecond
+field is ignored. The RTC itself does not provide a microsecond value.
+
 SD card
 -------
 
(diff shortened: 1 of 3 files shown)


docs/mimxrt/quickref.rst
```diff
@@ -438,6 +438,10 @@ The i.MXRT MCU supports battery backup of the RTC.  By connecting a battery of
 current drawn from the battery is ~20µA, which is rather high.  A CR2032 coin
 cell will last for about one year.
 
+Note: In v1.23.0 the support for subseconds was removed. When reading the RTC, 0 will
```
dpgeorge, line 441: Can you please link to the commit which removed subseconds?

Is it that the hardware just doesn't support subseconds?  But then how was it implemented before 1.23.0?

robert-hh:
The hardware does not support subseconds. In 1.23.0 and earlier a subsecond mechanism was simulated using ticks_us(). But since RTC and ticks_us() use different clocks, that would drift away.
The change was done in commit 913f9ad5ad9bb446c6e4310b6b5eda310098a1cf

dpgeorge:
OK, thanks for the info about subseconds.

Now merged.

robert-hh:
Thank you.

PR #15239, opened 2024-06-10
samd/mcu/samd21/mpconfigmcu.h:Allow user to enable additional options.

ricksorensen:
Currently for samd21 some features are disable because of limited memory. With the ability to trade firmware and filesystem space, a user may wish to selectively enable some of these features.  This change allows them to be enabled in board mpconfigboard.h or on the build command line for example.  The selectively enable functions are:

 * MICROPY_PY_FRAMEBUF
 * MICROPY_PY_SELECT
 * MICROPY_PY_ONEWIRE
 * MICROPY_PY_ASYNCIO


robert-hh:
That's fine. As part of the make process also the setting for `MICROPY_HW_CODESIZE` has to be adapted accordingly.
Note: The commit message need a space character after the colon. Just change the commit locally with `git commit --amend` and then push it with `git push <your_forks_name> samd_func -f`.

dpgeorge:
Thanks, now merged.

PR #15240, opened 2024-06-10
extmod/modplatform: Add picolibc to the recognised libcs list.

agatti:
Picolibc was not yet recognised by the `platform` module, this PR adds it to the known libc implementations MicroPython knows about.


dpgeorge:
Thanks!

PR #15251, opened 2024-06-11
github: Add Pull Request template.

projectgus:
### Summary

Provides pull request submitters with contributor documentation, and prompts them to provide relevant information about testing, and trade-offs or alternatives to their change.

Sections are deliberately small so they don't crowd out the space available in the GitHub Pull Request description text field.

### Testing

I've set the template as the default branch on my fork, so clicking through the following link allows testing it out:
https://github.com/projectgus/micropython/compare/master...projectgus:micropython:github/pr_template

There's a trade-off between explaining everything in the template, has chance it is "TLDR" ignored or a submitter accidentally submits their text inside one of the long HTML comments (although a maintainer can fix that formatting easily).

Some projects, for [example vscode](https://raw.githubusercontent.com/microsoft/vscode/main/.github/pull_request_template.md), have a really short template that probably helps with that.

I think it is worth keeping some detail in the MicroPython template, though, particularly if it means less back and forth to find out about testing or the reason for making a change.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,33 @@
+<!-- Thanks for submitting a Pull Request! We appreciate you spending the
+     time to improve MicroPython. Please provide enough information so that
+     others can review your Pull Request.
+
+     Before submitting, please read:
+     https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md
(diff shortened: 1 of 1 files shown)


dpgeorge:
Looks good, thanks!

PR #15259, opened 2024-06-11
Document limited numerical precision

tdunning:
Added limited doc changes to alert users about the difference in numerical precision relative to standard Python.

### DIFF
--- a/docs/differences/numerics.rst
+++ b/docs/differences/numerics.rst
@@ -1,0 +1,49 @@
+.. _numerics:
+
+In Python 3.x, floating point numbers are implemented using double 
+precision. This provides roughly 53 significant bits (17 digits or so)
+which means that many programmers can often (but not always) simply
+ignore the issue of limited floating-point precision without much
+risk of being caught out. The cost of putting floating point numbers
+into the heap is considered to be high enough that the slightly 
+higher performance that single precision numbers would provide is
+reckoned to be immaterial.
+
+MicroPython, on the other hand, typically runs on much lower performance
+hardware that typically lacks double precision floating point hardware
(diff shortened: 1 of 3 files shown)


docs/differences/numerics.rst
```diff
@@ -0,0 +1,49 @@
...
+ignore the issue of limited floating-point precision without much
+risk of being caught out. The cost of putting floating point numbers
+into the heap is considered to be high enough that the slightly 
+higher performance that single precision numbers would provide is
+reckoned to be immaterial.
+
+MicroPython, on the other hand, typically runs on much lower performance
+hardware that typically lacks double precision floating point hardware
+and is commonly memory-constrained. There is also the perception that
+the microcontrollers that are typically used with MicroPython typically
```
peterhinch, line 15: If the microcontroller includes 32-bit floating point hardware, MicroPython uses this by default on grounds of speed. It is usually possible to specify 64-bit software FP as a build option to prioritise precision over performance.

peterhinch (commented):
I question whether this requires such detailed coverage. The issue of floating point precision is both elementary and well known. For example the CPython `array` module has `f` and `d` type codes.

tdunning:
It is a fair comment that there may be too much detail in the numerics doc, but some justification and explanation is definitely needed and the Differences from CPython page (and the README) definitely need to mention it.

stinos:
It definitely needs mentioning, but can't we point to the canonical 'what every programmer must know about FP' or similar ?

tdunning:
That's a great idea. I pointed at a decent fixed point tutorial, but I don't know of a good FP intro off-hand.

Suggestions welcome!!

PR #15262, opened 2024-06-12
extmod/machine_usb_device: Add remote_wakeup.

felixdoerre:
I'm using usb-"remote_wakeup" on one of my projects and adding it to the existing usb_device implementation was just a matter of exposing the tusb-funcion and a tiny addition to micropython lib to allow setting the attribute indicating support, here:

https://github.com/micropython/micropython-lib/blob/50ed36fbeb919753bcc26ce13a8cffd7691d06ef/micropython/usb/usb-device/usb/device/core.py#L192

@projectgus, maybe you're interested.

### DIFF
--- a/extmod/machine_usb_device.c
+++ b/extmod/machine_usb_device.c
@@ -157,6 +157,11 @@ static mp_obj_t usb_device_active(size_t n_args, const mp_obj_t *args) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(usb_device_active_obj, 1, 2, usb_device_active);
 
+static mp_obj_t usb_remote_wakeup(mp_obj_t self) {
+    return mp_obj_new_bool(tud_remote_wakeup());
+}
+static MP_DEFINE_CONST_FUN_OBJ_1(usb_remote_wakeup_obj, usb_remote_wakeup);
+
 static mp_obj_t usb_device_stall(size_t n_args, const mp_obj_t *args) {
     mp_obj_usb_device_t *self = (mp_obj_usb_device_t *)MP_OBJ_TO_PTR(args[0]);
     int epnum = mp_obj_get_int(args[1]);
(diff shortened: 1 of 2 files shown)


extmod/machine_usb_device.c
```diff
@@ -157,6 +157,12 @@ static mp_obj_t usb_device_active(size_t n_args, const mp_obj_t *args) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(usb_device_active_obj, 1, 2, usb_device_active);
 
+static mp_obj_t usb_remote_wakeup(mp_obj_t self) {
+    tud_remote_wakeup();
```
projectgus, line 161: This functions return true if wake up succeeded (device was suspended, and host supports remote wakeup). Is there any value returning this to the caller, do you think?

projectgus (commented):
This looks very useful, thanks @felixdoerre!

We also don't have a way to query the suspend state from Python (i.e. a wrapper around `tud_suspended()`, or a callback for suspend and resume events). Are you doing OK without that?

(If it's OK without it then it's better not to add it, to keep code size down. I'm asking because of curiousity, I haven't done much with USB suspend/resume myself.)

dpgeorge:
Would it be more convenient for the Python user to add an optional argument to `submit_xfer(..., remote_wakeup=False)` which triggers the remote wakeup after the IN transfer is queued?

felixdoerre (commented):
> We also don't have a way to query the suspend state from Python (i.e. a wrapper around tud_suspended(), or a callback for suspend and resume events). Are you doing OK without that?

I don't have any need for that. In my usecase a rp-pico is connected to mqtt and an usb host and will wake up the usb host on an mqtt message, just like a wake-on-wlan workaround :)

> `submit_xfer(..., remote_wakeup=False)`

I also don't need to send any data via USB. I only use the "default" USB-CDC, and don't call `submit_xfer` myself as far as I am aware, so that interface would be a bit more wired for me.

extmod/machine_usb_device.c
projectgus: Yes please, as you say this should have minimal size impact and may be useful later if we add more suspend/resume awareness.

projectgus:
@felixdoerre Please also add this function to `docs/library/machine.USBDevice.rst`. 

Otherwise it LGTM, should be good to add! Thanks for contributing it.

PR #15264, opened 2024-06-12
rp2: Fix recursive atomic sections when core1 is active.

projectgus:
## Summary

mp_thread_begin_atomic_section() is expected to be recursive (i.e. for nested machine.disable_irq() calls, or if Python code calls disable_irq() and then the Python runtime calls mp_handle_pending() which also enters an atomic section to check the scheduler state).

On rp2 when not using core1 the atomic sections are recursive.

However when core1 was active (i.e. _thread) then there was a bug that caused the core to live-lock if an atomic section recursed.

Adds a test case specifically for mutual exclusion and recursive atomic sections when using two threads. Without this fix the test immediately hangs on rp2.

This was found while testing a fix for https://github.com/micropython/micropython-lib/issues/874 (but it's only a partial fix for that issue).

## Testing

* Re-ran test suite, including the new unit test, on rp2 with this change.
* Re-ran the test code from the linked issue, it no longer randomly hangs in live-lock.
* Also ran the new `thread/disable_irq.py` test on esp32 port and verified correct output (via `mpremote run`, currently thread tests are disabled on this port.)

* `recursive_mutex_enter_blocking` is also compiled into the firmware, so it might be possible to call `save_and_disable_interrupts` and then `recursive_mutex_enter_blocking` in order to save a little code size. Not sure, though.
* Possibly shouldn't be calling the scheduler hook at all when interrupts are disabled. However this fix would still be needed for the recursive disable_irq() case.

### DIFF
--- a/ports/rp2/mpthreadport.c
+++ b/ports/rp2/mpthreadport.c
@@ -46,13 +46,13 @@ static uint32_t *core1_stack = NULL;
 static size_t core1_stack_num_words = 0;
 
 // Thread mutex.
-static mutex_t atomic_mutex;
+static recursive_mutex_t atomic_mutex;
 
 uint32_t mp_thread_begin_atomic_section(void) {
     if (core1_entry) {
         // When both cores are executing, we also need to provide
         // full mutual exclusion.
-        return mutex_enter_blocking_and_disable_interrupts(&atomic_mutex);
+        return recursive_mutex_enter_blocking_and_disable_interrupts(&atomic_mutex);
     } else {
         return save_and_disable_interrupts();
     }
@@ -60,7 +60,7 @@ uint32_t mp_thread_begin_atomic_section(void) {
 
 void mp_thread_end_atomic_section(uint32_t state) {
     if (atomic_mutex.owner != LOCK_INVALID_OWNER_ID) {
-        mutex_exit_and_restore_interrupts(&atomic_mutex, state);
+        recursive_mutex_exit_and_restore_interrupts(&atomic_mutex, state);
     } else {
         restore_interrupts(state);
     }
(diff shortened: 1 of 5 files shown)


projectgus:
Updated so the test is a bit more aggressive about testing the nesting of disable_irq.

dpgeorge:
Thanks, this looks like a necessary fix.  The `machine.disable_irq/enable_irq` functions should be able to be nested.

>  `recursive_mutex_enter_blocking` is also compiled into the firmware, so it might be possible to call `save_and_disable_interrupts` and then `recursive_mutex_enter_blocking` in order to save a little code size. Not sure, though.

I'm pretty sure the reason we needed custom mutex+irq functions is still valid, so we can't separate them again.  See dc2a4e3cbd1755228ae3de2d15ec8f95721b9ac9

projectgus:
> > 
> 
> I'm pretty sure the reason we needed custom mutex+irq functions is still valid, so we can't separate them again. See [dc2a4e3](https://github.com/micropython/micropython/commit/dc2a4e3cbd1755228ae3de2d15ec8f95721b9ac9)

That deadlock happens from taking the mutex before disabling interrupts, is that right? I don't think there's a similar deadlock from disabling interrupts before trying to take the mutex, because it's no longer possible for that core to be interrupted at the same point.

I think the main limitation of changing it to two calls is that if a core is waiting for the mutex, the current code restores interrupts each time around the loop so it doesn't starve interrupts on that core. If we disable interrupts before trying to take the mutex then interrupts will remain disabled on that core until the mutex is taken. Which I think is probably often a small period of time, but it could be longer in some cases.

However it's a small enough piece of code that keeping it like it is seems like the best course of action. :+1:

PR #11604, opened 2023-05-24
py/map: Make map rehashing an atomic operation.

dpgeorge:
This simple change to map rehashing makes it so the old map is kept intact during the rehash, and only after the rehash is complete is the old map updated to the new one.  And this update is a very quick memory copy, which is done inside an atomic section.

This means that hard IRQs handlers can access dict members (eg globals) even during the update (add/del) of that dict.

See related #10620.

PR #15265, opened 2024-06-12
tests/thread: Restore gc before test case ends.

projectgus:
## Summary

Otherwise GC stays disabled (not re-enabled by soft reset) and later test runs fail with MemoryError.

Note this test case is currently disabled in CI due to "reliability issues" (as of 02b44a0154). Not sure if this is the root cause of the issue in question.

## Testing

* Tested on rp2 port and verified subsequent test failures went away.
* Also re-ran tests on unix port with `run-tests.py -d thread`, but no change expected here as a new micropython process is run for each test file.

We could have soft reset re-enable automatic gc, instead? Would be more resilient than this fix, but might have other implications.

### DIFF
--- a/tests/thread/stress_schedule.py
+++ b/tests/thread/stress_schedule.py
@@ -52,6 +52,8 @@ while n < _NUM_TASKS and time.ticks_diff(time.ticks_ms(), t) < _TIMEOUT_MS:
 thread_run = False
 time.sleep_ms(20)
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks, this fix is good.

The CI reliability of these stress tests is more likely related to globals not being protected by a mutex, see related #11604.

PR #15268, opened 2024-06-12
extmod/modplatform: Expose CPU features/extensions.

agatti:
This PR adds the ability to expose CPU-specific features/extensions to scripts when the `platform` module is compiled in.

Right now this is only available on RISC-V RV32, but I looked into extending it for other architectures.  Some Cortex CPUs have a pair of registers called `ID_PFR` but no flags are defined in CMSIS.  For x86/x64 accessing those flags would be OS-dependent, but still doable if you limit yourself to Linux and Windows.  Finally, for Aarch64 the `ID_AA64PFR` registers have documented flags, but I have no Armv8 to test an eventual implementation on.

The function name is still somewhat tentative as different architectures may have different names for the same thing after all.

For RISC-V, it is not implemented as a singleton with precomputed values like other functions in the `platform` module because it is technically possible to disable extensions by clearing the bits in the MISA special register rather than reading them.

### DIFF
--- a/docs/library/platform.rst
+++ b/docs/library/platform.rst
@@ -36,3 +36,11 @@ Functions
    Returns a tuple of strings *(lib, version)*, where *lib* is the name of the
    libc that MicroPython is linked to, and *version* the corresponding version
    of this libc.
+
+.. function:: processor()
+
+   Returns a string with a detailed name of the processor, if one is available.
+   If no name for the processor is known, it will return an empty string
+   instead.
+
+   This is currently available only on RISC-V targets (both 32 and 64 bits).
--- a/extmod/modplatform.c
+++ b/extmod/modplatform.c
@@ -61,11 +61,49 @@ static mp_obj_t platform_libc_ver(size_t n_args, const mp_obj_t *pos_args, mp_ma
 }
 static MP_DEFINE_CONST_FUN_OBJ_KW(platform_libc_ver_obj, 0, platform_libc_ver);
 
+#ifdef __riscv
+static mp_obj_t platform_processor(void) {
+    #if (__riscv_xlen <= 64) && !defined(__linux__)
+    uintptr_t misa_csr = 0;
+
+    // Load the MISA CSR directly.
+    __asm volatile (
+        "csrr %0, misa \n"
+        : "+r" (misa_csr)
+        :
+        :
+        );
+
+    char processor_buffer[31] = { // "RV32"/"RV64" + up to 26 chars.
+        #if (__riscv_xlen < 64)
+        "RV32"
+        #else
+        "RV64"
+        #endif
+    };
+    mp_uint_t offset = 4;
+    for (mp_uint_t bit = 0; bit < 26; bit++) {
+        if (misa_csr & (1U << bit)) {
+            processor_buffer[offset++] = 'A' + bit;
+        }
+    }
+
+    return mp_obj_new_str(processor_buffer, offset);
+    #else
+    return MP_OBJ_NEW_QSTR(MP_QSTR_);
+    #endif
+}
+static MP_DEFINE_CONST_FUN_OBJ_0(platform_processor_obj, platform_processor);
+#endif
+
 static const mp_rom_map_elem_t modplatform_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_platform) },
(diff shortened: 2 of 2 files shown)


dpgeorge:
The `platform` module is a CPython standard module, so we can't add custom extensions to it.

Existing CPython functions such as `platform.machine()` and `platform.processor()` may be good places to put this info.

Otherwise, we already do have `sys.implementation._machine` which could also store this info.  Eg on stm32 it stores the full MCU name, such as:
```py
>>> import sys
>>> sys.implementation
(name='micropython', version=(1, 24, 0, 'preview'), _machine='PYBD-SF6W with STM32F767IIK', _mpy=8966)
```

agatti:
Understood.  I moved it to `platform.processor()`, and it now returns an empty string if it's not running on a RV32 processor.  This should be in line with CPython's behaviour.

vshymanskyy:
Currently I'm not aware about any way to distinguish between ESP32, STM32, SAMD flavors.
I think this is really important. For example in [ViperIDE](https://viper-ide.org), I'm running the `mpy-cross` directly in browser and I'd like to pass the `-march=` argument depending on the actual board that is connected.

Initially I started by creating a map like this:
```js
const archMap = {
    "rp2": "armv6m",
    "esp32": "xtensawin",
    "webassembly": null,
    "unix": "x86",
    ...
}
```
but it's not gonna work for `esp32c3`, `stm32` (`pyb`), etc.

dpgeorge:
> For example in [ViperIDE](https://viper-ide.org), I'm running the `mpy-cross` directly in browser and I'd like to pass the `-march=` argument depending on the actual board that is connected.

You should be able to use `sys.implementation._mpy` for this.  See https://docs.micropython.org/en/latest/reference/mpyfiles.html#versioning-and-compatibility-of-mpy-files

vshymanskyy:
@dpgeorge Thank you, this is exactly what I need

agatti:
I've updated it for RV64 so if it will ever get its own bare-metal target it's ready to go.

I've also disabled extensions enumeration code on Linux as it needs a different method to obtain that information (cannot access machine registers directly).

dpgeorge:
@agatti to get this PR over the line I suggest making this new function only available if it actually returns a non-empty string (so code size increase is +0 for most existing ports).  At the moment that's only for RISC-V machines.

extmod/modplatform.c
```diff
@@ -61,11 +61,45 @@ static mp_obj_t platform_libc_ver(size_t n_args, const mp_obj_t *pos_args, mp_ma
...
+    #endif
+}
+static MP_DEFINE_CONST_FUN_OBJ_0(platform_processor_obj, platform_processor);
+
 static const mp_rom_map_elem_t modplatform_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_platform) },
     { MP_ROM_QSTR(MP_QSTR_platform), MP_ROM_PTR(&platform_platform_obj) },
     { MP_ROM_QSTR(MP_QSTR_python_compiler), MP_ROM_PTR(&platform_python_compiler_obj) },
     { MP_ROM_QSTR(MP_QSTR_libc_ver), MP_ROM_PTR(&platform_libc_ver_obj) },
+    { MP_ROM_QSTR(MP_QSTR_processor), MP_ROM_PTR(&platform_processor_obj) },
```
dpgeorge, line 105: Wrap this in a `#if` that enables it only for RISC-V.

docs/library/platform.rst
```diff
@@ -36,3 +36,9 @@ Functions
    Returns a tuple of strings *(lib, version)*, where *lib* is the name of the
    libc that MicroPython is linked to, and *version* the corresponding version
    of this libc.
+
+.. function:: processor()
+
+   Returns a string with a detailed name of the processor, if one is available.
+   If no name for the processor is known, it will return an empty string
+   instead.
```
dpgeorge, line 44: Add a line that this function is only available on RISC-V processors.

agatti:
Here's an updated version that only brings that function bit in on RISC-V targets, along with the platform availability remark in the documentation.

dpgeorge (approved):
Thanks for updating.

I tested on RPI_PICO2_W in RISCV mode:
```
MicroPython v1.27.0-preview.123.gc224e976f7 on 2025-09-11; Raspberry Pi Pico 2 W with RP2350-RISCV
>>> import platform
>>> platform.processor()
'RV32ACIMUX'
```

PR #15271, opened 2024-06-13
py/objarray.c: Fix buffer overflow in case of memory allocation failure.

srinsoz:
If array.append fails with an exception due to heap exhaustion, the next attempt to grow the buffer will cause a buffer overflow because the free slot count is increased before performing the allocation, and will stay as if the allocation succeeded.

### DIFF
--- a/py/objarray.c
+++ b/py/objarray.c
@@ -392,8 +392,9 @@ static mp_obj_t array_append(mp_obj_t self_in, mp_obj_t arg) {
     if (self->free == 0) {
         size_t item_sz = mp_binary_get_size('@', self->typecode, NULL);
         // TODO: alloc policy
-        self->free = 8;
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the fix.

I managed to write a test that would fail before this patch is applied, and succeed after this patch is applied.  The test is:
```py
b = bytearray(16)
for _ in range(100000):
    try:
        while True:
            b.append(0)
    except:
        print('full')
```

That test works well on the unix port, at least.  But I couldn't make it reliable on the rp2 port, and I know it won't work well on the webassembly port (which keeps growing its heap), and on the esp32 with SPIRAM it will also be a difficult test because of the large amount of RAM.  So, I don't think it's worth adding this test to the test suite, it would be too much to maintain moving forward.

PR #15284, opened 2024-06-14
tests/extmod/machine_timer: Fix tests.

agatti:
`machine.Timer` requires a timer id in its constructor, which was not passed when building objects.

This PR makes the test pass on an ESP32C3 at least, as it would consistently fail before.

### DIFF
--- a/tests/extmod/machine_timer.py
+++ b/tests/extmod/machine_timer.py
@@ -9,30 +9,30 @@ except:
     raise SystemExit
 
 # create and deinit
-t = machine.Timer(freq=1)
+t = machine.Timer(0, freq=1)
 t.deinit()
 
 # deinit again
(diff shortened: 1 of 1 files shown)


dpgeorge:
Unfortunately this now fails on other ports that use `extmod/machine_timer.c`, eg stm32, which only support `-1` as the id (to create a software timer).  OTOH, esp32 does not support `-1`, ie doesn't support software timers.

I think we should introduce a `machine.SoftTimer` class, like we did with `machine.SoftI2C` and `machine.SoftSPI`.  But that's out of scope for this PR.

dpgeorge:
Fixed in a different way in 9ba04cc7563ec934ca14d66aa18ae3563c8d1aea

Issue #15275, opened 2024-06-13
Cannot instantiate FrameBuffer from a bytes object


peterhinch:
### Port, board and/or hardware

RP2, Pyboard 1.1

### MicroPython version

MicroPython v1.22.1 on 2024-01-05; Raspberry Pi Pico with RP2040

### Reproduction

Enter the following at the REPL:
```py
from framebuf import FrameBuffer, GS8
b = b'\x00' * 100
f = FrameBuffer(b, 10, 10, GS8)
```

### Expected behaviour

I would expect the instantiation of the `FrameBuffer` to succeed (as it does if a `bytearray` is passed). 

### Observed behaviour

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object with buffer protocol required
```

### Additional Information

Use case:
Instantiating from a `bytes` object (or a `memoryview` of `bytes`) would enable graphical objects frozen as bytecode to be blitted onto an existing frame buffer.

A further observation: the error message is most misleading.

### Code of Conduct

Yes, I agree

massimosala:
And from the
[Python docs](https://docs.python.org/3/c-api/buffer.html)
 it seems _bytes_ should implement the buffer protocol:

massimosala:
This doesn't give errors... if you wanna try it and report back:

> f = FrameBuffer(bytearray(b), 10, 10, GS8)

peterhinch:
@massimosala Perhaps my report wasn't clear enough.

See "Expected behaviour" above: a `bytearray` does work.

A `bytes` object does support the buffer protocol. That is why I described the error message as misleading.

Gadgetoid:
Looks like `bytes` fails the inside check in `mp_get_buffer`:

```c
MP_OBJ_TYPE_GET_SLOT(type, buffer)(obj, bufinfo, flags & MP_BUFFER_RW) == 0
```

Because it's effectively a string, and you can't write to a string:

```c
mp_int_t mp_obj_str_get_buffer(mp_obj_t self_in, mp_buffer_info_t *bufinfo, mp_uint_t flags) {
    if (flags == MP_BUFFER_READ) {
        GET_STR_DATA_LEN(self_in, str_data, str_len);
        bufinfo->buf = (void *)str_data;
        bufinfo->len = str_len;
        bufinfo->typecode = 'B'; // bytes should be unsigned, so should unicode byte-access
        return 0;
    } else {
[...]

PR #15285, opened 2024-06-14
extmod/modframebuf.c: handle read-only storage types.

Gadgetoid:
Edit: ignore this, skip to https://github.com/micropython/micropython/pull/15285#issuecomment-2171119725

Handle a FrameBuffer wrapper around a read-only type, such as bytes() frozen as bytecode.

This is an extremely crude attempt at supporting the use-case described in #15275 without having every single method on FrameBuffer have to check the writability of the underlying storage buffer.

I don't particularly like this solution, but I'm raising it as a PR to hopefully promote discussion about how this kind of (very valid in my opinion) use case can be solved.

Between 32blit, PicoSystem and PicoGraphics I have some not-trivial experience dealing with graphics and asset conversion on embedded platforms, but most of these are rather more complex and wider in scope than MicroPython's FrameBuffer.

Some tests:

```python
from framebuf import FrameBuffer, GS8

buf = FrameBuffer(bytes(25), 5, 5, GS8)

print(dir(buf))

buf2 = FrameBuffer(bytearray(bytes(25)), 5, 5, GS8)

print(dir(buf2))

buf2.blit(buf, 0, 0)
buf.blit(buf2, 0, 0)
```

Outputs:

```
./build-standard/micropython test.py 
['__class__', 'blit', 'ellipse', 'fill', 'fill_rect', 'hline', 'line', 'pixel', 'poly', 'rect', 'scroll', 'text', 'vline']
['__class__', 'blit', 'ellipse', 'fill', 'fill_rect', 'hline', 'line', 'pixel', 'poly', 'rect', 'scroll', 'text', 'vline']
  File "test.py", line 12, in <module>
TypeError: read-only
```

Other potential solutions:

* Expand `blit` to accept raw bytes
* Add a `raw_blit` to do what blit does, using faster scanline copies, from raw bytes

### DIFF
--- a/examples/natmod/framebuf/framebuf.c
+++ b/examples/natmod/framebuf/framebuf.c
@@ -10,6 +10,7 @@ void *memset(void *s, int c, size_t n) {
 #endif
 
 mp_obj_full_type_t mp_type_framebuf;
+mp_obj_full_type_t mp_type_staticbuf;
 
 #include "extmod/modframebuf.c"
 
--- a/extmod/modframebuf.c
+++ b/extmod/modframebuf.c
@@ -44,6 +44,7 @@ typedef struct _mp_obj_framebuf_t {
 
 #if !MICROPY_ENABLE_DYNRUNTIME
 static const mp_obj_type_t mp_type_framebuf;
+static const mp_obj_type_t mp_type_staticbuf;
 #endif
 
 typedef void (*setpixel_t)(const mp_obj_framebuf_t *, unsigned int, unsigned int, uint32_t);
@@ -231,6 +232,18 @@ static void gs8_fill_rect(const mp_obj_framebuf_t *fb, unsigned int x, unsigned 
     }
 }
 
+static mp_obj_t get_framebuf_or_staticbuf(mp_obj_t buf_in) {
+    mp_obj_t buf_out;
+    buf_out = mp_obj_cast_to_native_base(buf_in, MP_OBJ_FROM_PTR(&mp_type_framebuf));
+    if (buf_out == MP_OBJ_NULL) {
+        buf_out = mp_obj_cast_to_native_base(buf_in, MP_OBJ_FROM_PTR(&mp_type_staticbuf));
+    }
+    if (buf_out == MP_OBJ_NULL) {
+        mp_raise_TypeError(NULL);
+    }
+    return buf_out;
+}
+
 static mp_framebuf_p_t formats[] = {
     [FRAMEBUF_MVLSB] = {mvlsb_setpixel, mvlsb_getpixel, mvlsb_fill_rect},
     [FRAMEBUF_RGB565] = {rgb565_setpixel, rgb565_getpixel, rgb565_fill_rect},
@@ -318,7 +331,13 @@ static mp_obj_t framebuf_make_new(const mp_obj_type_t *type, size_t n_args, size
     }
 
     mp_buffer_info_t bufinfo;
-    mp_get_buffer_raise(args_in[0], &bufinfo, MP_BUFFER_WRITE);
+    bool status = mp_get_buffer(args_in[0], &bufinfo, MP_BUFFER_WRITE);
+    if (!status) {
+        mp_get_buffer_raise(args_in[0], &bufinfo, MP_BUFFER_READ);
+        if (type != (mp_obj_type_t *)&mp_type_staticbuf) {
+            mp_raise_ValueError(MP_ERROR_TEXT("buffer must be writable"));
+        }
+    }
 
     if ((strides_required * stride + (height_required - strides_required) * width_required) * bpp / 8 > bufinfo.len) {
         mp_raise_ValueError(NULL);
@@ -709,10 +728,7 @@ static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(framebuf_poly_obj, 5, 6, framebuf_pol
 
 static mp_obj_t framebuf_blit(size_t n_args, const mp_obj_t *args_in) {
     mp_obj_framebuf_t *self = MP_OBJ_TO_PTR(args_in[0]);
-    mp_obj_t source_in = mp_obj_cast_to_native_base(args_in[1], MP_OBJ_FROM_PTR(&mp_type_framebuf));
-    if (source_in == MP_OBJ_NULL) {
-        mp_raise_TypeError(NULL);
-    }
+    mp_obj_t source_in = get_framebuf_or_staticbuf(args_in[1]);
     mp_obj_framebuf_t *source = MP_OBJ_TO_PTR(source_in);
 
     mp_int_t x = mp_obj_get_int(args_in[2]);
@@ -723,7 +739,8 @@ static mp_obj_t framebuf_blit(size_t n_args, const mp_obj_t *args_in) {
     }
     mp_obj_framebuf_t *palette = NULL;
     if (n_args > 5 && args_in[5] != mp_const_none) {
-        palette = MP_OBJ_TO_PTR(mp_obj_cast_to_native_base(args_in[5], MP_OBJ_FROM_PTR(&mp_type_framebuf)));
+        mp_obj_t palette_in = get_framebuf_or_staticbuf(args_in[5]);
+        palette = MP_OBJ_TO_PTR(palette_in);
     }
 
     if (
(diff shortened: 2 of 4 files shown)


Gadgetoid:
Okay now actually having thrown this against CI and some platform builds, I don't think it's tenable at all. Though I am open to being convinced otherwise 😆

In light of this, would `raw_blit` be an acceptable addition to FrameBuffer?

peterhinch:
This may be a dumb suggestion as my knowledge of the C code is limited.

The only legitimate thing to to with a read-only frame buffer is to use it to blit onto a r/w `FrameBuffer`. Is there a way to make its methods that imply writing (e.g. `.fill` etc.) invisible to Python so that an `AttributeError` is thrown?

Gadgetoid:
> This may be a dumb suggestion as my knowledge of the C code is limited.
> 
> The only legitimate thing to to with a read-only frame buffer is to use it to blit onto a r/w `FrameBuffer`. Is there a way to make its methods that imply writing (e.g. `.fill` etc.) invisible to Python so that an `AttributeError` is thrown?

Yes you can just drop them from the locals dict in C, which is just a lookup table of all the functions that exist on a MicroPython object.

In this case I’ve copied the entire locals dict and stubbed out every method with the same C function- a simple helper to catch any use and raise whatever error we’d deem appropriate. If the FrameBuffer constructor detects a read-only source it throws this cut-down dummy class at you instead.

This doesn’t seem to play nice with the dynamic bindings, though 😬

That really leaves us with just accepting a read-only buffer, saving a “read_only” flag into the FrameBuffer instance and just checking that and raising an exception in every single draw method. Sure would be nice if C had decorators 🤣 the end result to the user would feel the same, but it’s a bit of a burden on future programmers to make sure any write methods include this check.

Gadgetoid:
IMHO the only way for this *not* to be a burden on future programmers is to raise a more appropriate exception when “FrameBuffer” is given a read-only bytes object, and then have a “StaticBuffer” class. This would smell like a FrameBuffer but would make this use case more explicit.

Since a StaticBuffer could still encode width/height/type it would allow for cross-format blits in the same way using FrameBuffer would, but just lets you be more explicit about its 
expected use. 🤔

Gadgetoid:
I have blown away my previous bonkers changes and introduced a `StaticBuffer` type.

For a crude example, here's how you might create a static, 5px by 5px "white" (depends on pallete really) square asset for compiling into bytecode:

```python
buf = StaticBuffer(b"\xff" * 25, 5, 5, GS8)
```

Then you can create a regular `FrameBuffer` and blit a `StaticBuffer` to it as if it were a `FrameBuffer`.

```python
from framebuf import FrameBuffer, StaticBuffer, GS8

my_asset = StaticBuffer(b"\xff" * 25, 5, 5, GS8)

buf = bytearray(10 * 10)
canvas = FrameBuffer(buf, 10, 10, GS8)
canvas.blit(my_asset, 2, 2)

# Print out the underlying buffer, showing the static asset blit'd inside
for y in range(0, len(buf), 10):
    print(",".join((f"{x:02x}" for x in buf[y:y+10])))
```

Outputs:

```
00,00,00,00,00,00,00,00,00,00
00,00,00,00,00,00,00,00,00,00
00,00,ff,ff,ff,ff,ff,00,00,00
00,00,ff,ff,ff,ff,ff,00,00,00
00,00,ff,ff,ff,ff,ff,00,00,00
00,00,ff,ff,ff,ff,ff,00,00,00
00,00,ff,ff,ff,ff,ff,00,00,00
00,00,00,00,00,00,00,00,00,00
00,00,00,00,00,00,00,00,00,00
00,00,00,00,00,00,00,00,00,00
```

Of course my changes are still fraught with issues, notably sharing the same constructor between `StaticBuffer` and `FrameBuffer` seems tricksy, but I don't really want to copy all those useful checks and balances.

peterhinch:
Looks good to me.

For some reason I couldn't fetch the PR to try it. Is this because it's failing CI?
```bash
[adminpete@meerkat]: ~/micropython
$ git fetch origin pull/15285/head:staticbuf
fatal: couldn't find remote ref pull/15285/head
```

Gadgetoid:
Uh that’s a good question- I don’t know!

Branch should be patch-readonly-framebuffer on the Pimoroni fork because CircuitPython broke my ability to have a personal fork 🫠

I don’t know how likely this change is to be approved and it obviously still needs work, but it feels like the best of a bad selection of solutions to this particular very valid use case.

Gadgetoid:
If you’ve got a local copy you’re hacking on, you can always just grab the patch and slap that over the top - https://patch-diff.githubusercontent.com/raw/micropython/micropython/pull/15285.patch

peterhinch:
That works fine, with informative error messages if you try to create a `FrameBuffer` from a `bytes` instance or modify a `StaticBuffer`.

Gadgetoid:
Did you manage to bake the `my_static_asset = StaticBuffer(b"", w, y, type)` pattern into bytecode directly, or are you just storing the `bytes`?

An asset conversion tool could probably do the former, so the user can just blit `my_static_asset` and it would have a known size and datatype.

peterhinch:
I'm storing the `bytes`. My image file is frozen: `img` below is a `bytes` object physically located in Flash.
```py
import my_image as img  # Application code is not currently frozen
from framebuf import FrameBuffer, GS8, StaticBuffer
fbuf = StaticBuffer(img.data, 240, 240, GS8)
```
I then issue
```py
ssd.blit(fbuf, 0, 0)
```
The `ssd` object is the display driver instance, which is subclassed from `FrameBuffer`.

peterhinch:
It would be good to get feedback from the maintainers on this. It seems like a good solution to a practical problem. @dpgeorge ?

Gadgetoid:
Currently I'm missing explicit tests for the new class, too, though the constructor is enough to please the test coverage checks. I'm not totally sure where tests for this could fit in, but if this is suitable for inclusion then tests should probably be added.

dpgeorge:
@projectgus and I discussed this PR.  We think that it's a useful -- and maybe even necessary -- feature.  The point is, it's not currently possible to blit directly from ROM to a frame buffer, you need to go via RAM to do that (and wrap the RAM in a `FrameBuffer`).

@peterhinch in your font library https://github.com/peterhinch/micropython-font-to-py your writer code seems to get around the need to create a glyph in RAM by using `bytearray_at`.  That's neat but still not ideal.  And your code still needs to create a new `FrameBuffer` each time it renders a character.

This PR adds a new `StaticBuffer` object which acts like a `FrameBuffer` except it has no methods, only the internal state like width, height, stride and format, and a pointer to the read-only buffer.  That's a pretty neat way of doing things!  But, still, you need to create a new `StaticBuffer` (in RAM) each time you want to blit something different to a `FrameBuffer`.  In the case of @peterhinch 's font writer code, it's not going to be much better off using `StaticBuffer` than it currently is using `FrameBuffer` wrapping a `bytearray_at`.

@Gadgetoid suggests some alternatives to `StaticBuffer`:

> Other potential solutions:
> 
>     * Expand `blit` to accept raw bytes
> 
>     * Add a `raw_blit` to do what blit does, using faster scanline copies, from raw bytes

@projectgus and I were thinking that the first alternative here might be a good solution.  But it's not enough just to accept raw bytes, you also need to pass in the width/height/stride/format of the bytes data.  And our proposal is that all this information is passed in through a tuple/list.

Essentially instead of a `StaticBuffer` wrapping a ROM bytes object with width/height/etc, you use a tuple/list instead:
```py
glyph = (glyph_bytes, glyph_width, glyph_height, glyph_stride, glyph_format)
frame_buffer.blit(glyph, x, y)
```
You could also use a list instead of a tuple, and then you could reuse that list each time you draw a new glyph.  That keeps memory churn down to a minimum (could probably do it all without allocating anything on the GC heap).

Using a tuple/list is not as elegant/Pythonic as an object like `StaticBuffer`, but it is simple, and crucially allows zero allocations.  It should also be relatively small code size to implement.

@Gadgetoid @peterhinch what do you think about that idea?

Gadgetoid:
> And our proposal is that all this information is passed in through a tuple/list.

It probably wont surprise you to learn that I've done something like this. With PicoVision IIRC. I had a method for loading data from an image file which returns a tuple with dimensions and the binary data, though in this case it was for decoding assets into RAM (for performance)- https://github.com/pimoroni/picovision/blob/0d9c413128341ab784a8df9852929c4011265737/examples/sneks_and_ladders/pvgame.py#L8-L26

From the MicroPython perspective this is nice, since it gives a formal interface for programmers to get data into the framebuffer and doesn't really have any constraints over how. You could produce a `Spritesheet` class that loads some data from *somewhere* and just overloads the right magic methods to make it convert to a tuple.

I'd put width, height, stride and format before the data so it feels like a "header", but otherwise it sounds good to me.

dpgeorge:
> I'd put width, height, stride and format before the data so it feels like a "header"

Yes, that sounds good.

peterhinch:
Looks good so long as `.blit` still supports the `palette` arg.

dpgeorge:
Thanks for rebasing.  The adjusted code size report now looks a lot more consistent.  Compared with #17358, the implementation here is about 50% bigger on bare-metal targets.

dpgeorge:
Closing in favour of #17358.

PR #15293, opened 2024-06-17
webassembly/objjsproxy: Implement proxying of JS iterable protocol.

dpgeorge:
This allows Python to iterate over JavaScript objects that provide Symbol.iterator.


WebReflection:
virtual :+1: to me ... if you merge and you don't mind, please publish to *npm* so we can scratch one more issue from our side, thank you!

PR #15299, opened 2024-06-18
webassembly: Support top-level await of asyncio Task and Event.

dpgeorge:
This change allows doing a top-level await on an asyncio primitive like Task and Event.

This feature enables a better interaction and synchronisation between JavaScript and Python, because `api.runPythonAsync` can now be used (called from JavaScript) to await on the completion of asyncio primitives.


dpgeorge:
@ntoll FYI

WebReflection:
ruff a part, and for what is worth it, this looks good to me

ntoll:
ruff..?

ntoll:
Thank you so much @dpgeorge for this. Can't wait to give it a spin.

PR #15301, opened 2024-06-19
rp2: Fix USB PLL glitch during wake from light sleep.

projectgus:
Follow-up to #15111, that change works most of the time but has an intermittent bug where USB doesn't resume as expected after waking from light sleep. Thanks to @GitHubsSilverBullet who reported it here: https://github.com/orgs/micropython/discussions/14401#discussioncomment-9675017

Turns out waking calls `clocks_init()` which will re-initialise the USB PLL. Most of the time this is OK but occasionally it seems like the clock glitches the USB peripheral and it stops working until the next hard reset.

Adds a `machine.lightsleep()` test that consistently hangs in the first two dozen iterations of a single test run on rp2 without this fix. Passed over 100 full test runs in a row with this fix.

The test is currently rp2-only as it seems similar lightsleep USB issues exist on other ports (both pyboard and ESP32-S3 native USB don't send any data to the host after waking, until they receive something from the host first.)

EDIT: Added one more small commit to always disable USB if there is no timeout (i.e. entering DORMANT mode where the external oscillator is disabled.)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -73,6 +73,7 @@ used during the build process and is not part of the compiled source code.
             /ppp_set_auth.* (Apache-2.0)
         /rp2
             /mutex_extra.c (BSD-3-clause)
+            /clocks_extra.c (BSD-3-clause)
         /stm32
             /usbd*.c (MCD-ST Liberty SW License Agreement V2)
             /stm32_it.* (MIT + BSD-3-clause)
--- a/ports/rp2/CMakeLists.txt
+++ b/ports/rp2/CMakeLists.txt
@@ -117,6 +117,7 @@ set(MICROPY_SOURCE_DRIVERS
 )
 
 set(MICROPY_SOURCE_PORT
+    clocks_extra.c
     fatfs_port.c
     help.c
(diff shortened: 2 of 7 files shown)


dpgeorge:
Maybe we should report this to pico-sdk, and potentially they can add this argument to `clocks_init()`?

But for now, patching it here is OK.

Can you please split out the addition of our copy of `clocks_init()` to a separate commit and update the top-level LICENSE?  See 8438c8790c0ad11545371a07f0b0ce4c65c62101

projectgus:
> Can you please split out the addition of our copy of `clocks_init()` to a separate commit and update the top-level LICENSE? See [8438c87](https://github.com/micropython/micropython/commit/8438c8790c0ad11545371a07f0b0ce4c65c62101)

No worries, done. PTAL!

projectgus:
Added one more small commit to always disable USB if going to DORMANT mode with no external oscillator.

Have also raised a feature request with pico-sdk, see link above.

dpgeorge:
All looks good, thank you!

Issue #13770, opened 2024-02-27
Async randomly dying


desultory:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

RP2

### MicroPython version

MicroPython v1.22.2 on 2024-02-22; Raspberry Pi Pico with RP2040

### Reproduction

```from machine import UART, Pin
from uasyncio import StreamReader, StreamWriter, Event, Lock, sleep_ms, run


LoRa_bands = {'EU': 868.1, 'US': 915.0}


class LoRa:
    def __init__(self, uart=0, tx_pin=0, rx_pin=1, reset_pin=5, baudrate=115200,
                 band='US', address=2):
        self.uart = UART(uart, baudrate=baudrate, tx=tx_pin, rx=rx_pin)
        self.reset_pin = Pin(reset_pin, Pin.OUT)
        self.reader = StreamReader(self.uart)
        self.writer = StreamWriter(self.uart, {})

        self._band = band
        self._address = address

        self.initialized = False
        self.seat = Lock()
        self._recv = Event()
        self._received = Event()

    async def send_query(self, query):
        return await self.send_command(query, True)

    async def send_command(self, cmd, query=False):
        """ Sends an AT command to the device """
        cmd = cmd.upper()
        if not cmd.startswith('AT+'):
            orig_cmd = cmd
            cmd = 'AT+' + cmd
        else:
            orig_cmd = cmd.removeprefix('AT+')

        if query and cmd[-1] != '?':
            cmd += '?'

        cmd += '\r\n'

        async with self.seat:
            print("Sending command: ", cmd.rstrip())
            await self.writer.awrite(cmd)
            await sleep_ms(10 if query else 50)
            while not self.uart.any():
                print("Waiting for response...")
                await sleep_ms(250)
            response = await self.recv()

        response = response[:-2]
        print(response)

        if query:
            if response.startswith(f"+{orig_cmd}="):
                response = response[len(f"+{orig_cmd}="):]
            else:
                raise ValueError("Invalid response: %s" % response)
        elif response != '+OK':
            raise ValueError("Error: %s" % response)
        return response

    async def reset(self):
        print("Resetting LoRa module.")
        self.reset_pin.on()
[...]

PR #15302, opened 2024-06-19
webassembly/asyncio: Schedule run loop when tasks are pushed to queue.

dpgeorge:
In the webassembly port there is no asyncio run loop running at the top level.  Instead the Python asyncio run loop is scheduled through setTimeout and run by the outer JavaScript event loop.  Because tasks can become runable from an external (to Python) event (eg a JavaScript callback), the run loop must be scheduled whenever a task is pushed to the asyncio task queue, otherwise tasks may be waiting forever on the queue.


ntoll:
Bravo @dpgeorge

WebReflection:
if happy about this, please remember to publish on *npm*, thank you! 🙏

ntoll:
What @WebReflection said, @dpgeorge :rofl:

EmilioMartinez:
Hi @dpgeorge, is there any chance to implement this change and the [top-level support for Event](https://github.com/micropython/micropython/pull/15299) to the RP2 port? Should I open an issue?

Just to clarify if it applies, the behaviour I see is that on the PiPico top-level waiting for an Event terminates the program (unless something else is running), and setting events seems to queue them without inmediate result, only to resume when something else wakes the scheduler.

I think it might be related to [this issue] (https://github.com/micropython/micropython/issues/16318), or even [this other](https://github.com/micropython/micropython/issues/13770)

dpgeorge:
> is there any chance to implement this change and #15299 to the RP2 port?

No, it's not possible to implement this on anything other than the webassembly port, because that port has an async-like scheduler in the "operating system", ie browser/Node.



> the behaviour I see is that on the PiPico top-level waiting for an Event terminates the program (unless something else is running), and setting events seems to queue them without inmediate result, only to resume when something else wakes the scheduler.

If you think this is a bug then please make a minimal reproduction of your code that still has the bug and create a new issue for it.

PR #7370, opened 2021-06-08
various: Fix -fsanitize=undefined errors.

jepler:
[messages truncated to pass ci_commit_formatting_run]
[the line numbers match 0e87459e2bfd07]
```
../../extmod/crypto-algorithms/sha256.c:49:19: runtime error: left shif...
../../extmod/moduasyncio.c:106:35: runtime error: member access within ...
../../py/binary.c:210:13: runtime error: left shift of negative value -...
../../py/mpz.c:744:16: runtime error: negation of -9223372036854775808 ...
../../py/objint.c:109:22: runtime error: left shift of 1 by 31 places c...
../../py/objint_mpz.c:374:9: runtime error: left shift of 4611686018427...
../../py/objint_mpz.c:374:9: runtime error: left shift of negative valu...
../../py/parsenum.c:106:14: runtime error: left shift of 46116860184273...
../../py/runtime.c:395:33: runtime error: left shift of negative value ...
../../py/showbc.c:177:28: runtime error: left shift of negative value -...
../../py/vm.c:321:36: runtime error: left shift of negative value -1
```
My testing was done on an amd64 debian buster system using gcc-8.3 and
these settings:
```
CFLAGS += -g3 -Og -fsanitize=undefined
LDFLAGS += -fsanitize=undefined
```

The introduced PAIRHEAP macro's conditional (x ? &x->i : NULL)
assembles (under amd64 gcc 8.3 -Os) to the same as &x->i, since i is
the initial field of the struct. However, for the purposes of undefined
behavior analysis the conditional is needed.

This was originally filed as #7237 but for this PR everything except the "null pointer passed as argument" diagnostics are addressed.

Signed-off-by: Jeff Epler <jepler@gmail.com>

PR #7369, opened 2021-06-08
various: Fix null pointer errors

jepler:
[messages truncated to pass ci_commit_formatting_run]
[the line numbers match 0e87459e2bfd07]
```
mpz.c:711:5: runtime error: null pointer passed as argumen...
objarray.c:130:9: runtime error: null pointer passed as ar...
objarray.c:377:5: runtime error: null pointer passed as ar...
objarray.c:457:21: runtime error: null pointer passed as a...
objarray.c:457:21: runtime error: null pointer passed as a...
objarray.c:461:21: runtime error: null pointer passed as a...
objdict.c:250:5: runtime error: null pointer passed as arg...
objdict.c:250:5: runtime error: null pointer passed as arg...
objset.c:181:5: runtime error: null pointer passed as argu...
objset.c:181:5: runtime error: null pointer passed as argu...
qstr.c:181:78: runtime error: null pointer passed as argum...
runtime.c:395:33: runtime error: left shift of negative va...
runtime.c:534:17: runtime error: left shift of negative va...
sequence.c:108:15: runtime error: null pointer passed as a...
sequence.c:108:15: runtime error: null pointer passed as a...
showbc.c:177:28: runtime error: left shift of negative val...
vm.c:321:36: runtime error: left shift of negative value -...
```

My testing was done on an amd64 debian buster system using gcc-8.3 and these settings:
```
CFLAGS += -g3 -Og -fsanitize=undefined
LDFLAGS += -fsanitize=undefined
```

The changes are intended/expected to produce no runtime overhead, as the behavior of the mem* functions is only actually modified for the unix coverage run (MICROPY_NONNULL_COMPLIANT).

This was originally filed as #7237 but for this PR only the "null pointer passed as argument" diagnostics are addressed.

Signed-off-by: Jeff Epler <jepler@gmail.com>

PR #7237, opened 2021-05-10
various: Fix gcc -fsanitize=undefined errors

jepler:
```
../../extmod/crypto-algorithms/sha256.c:49:19: runtime error: left shift of # by 24 places cannot be represented in type 'int'
../../extmod/moduasyncio.c:106:35: runtime error: member access within null pointer of type 'struct mp_obj_task_t'
../../py/binary.c:210:13: runtime error: left shift of negative value -1
../../py/mpz.c:711:5: runtime error: null pointer passed as argument 2, which is declared to never be null
../../py/mpz.c:744:16: runtime error: negation of -9223372036854775808 cannot be represented in type 'long long int'; cast to an unsigned type to negate this value to itself
../../py/objarray.c:130:9: runtime error: null pointer passed as argument 1, which is declared to never be null
../../py/objarray.c:377:5: runtime error: null pointer passed as argument 1, which is declared to never be null
../../py/objarray.c:457:21: runtime error: null pointer passed as argument 1, which is declared to never be null
../../py/objarray.c:457:21: runtime error: null pointer passed as argument 2, which is declared to never be null
../../py/objarray.c:461:21: runtime error: null pointer passed as argument 1, which is declared to never be null
../../py/objdict.c:250:5: runtime error: null pointer passed as argument 1, which is declared to never be null
../../py/objdict.c:250:5: runtime error: null pointer passed as argument 2, which is declared to never be null
../../py/objint.c:109:22: runtime error: left shift of 1 by 31 places cannot be represented in type 'int'
../../py/objint_mpz.c:374:9: runtime error: left shift of 4611686018427387904 by 1 places cannot be represented in type 'long int'
../../py/objint_mpz.c:374:9: runtime error: left shift of negative value -#
../../py/objset.c:181:5: runtime error: null pointer passed as argument 1, which is declared to never be null
../../py/objset.c:181:5: runtime error: null pointer passed as argument 2, which is declared to never be null
../../py/parsenum.c:106:14: runtime error: left shift of 4611686018427387904 by 1 places cannot be represented in type 'long int'
../../py/qstr.c:181:78: runtime error: null pointer passed as argument 2, which is declared to never be null
../../py/runtime.c:395:33: runtime error: left shift of negative value -#
../../py/sequence.c:108:15: runtime error: null pointer passed as argument 1, which is declared to never be null
../../py/sequence.c:108:15: runtime error: null pointer passed as argument 2, which is declared to never be null
[...]

PR #3799, opened 2018-05-19
emitbc: Avoid undefined behavior calling memset()

jepler:
When micropython is built with 'clang -fsanitize=undefined', a diagnostic like the following will occur:
```
$ UBSAN_OPTIONS=abort_on_error=1 ./micropython_fuzzing  -c 'print(1)'
../../py/emitbc.c:319:16: runtime error: null pointer passed as argument 1, which is declared to never be null
/usr/include/string.h:62:62: note: nonnull attribute specified here
Aborted
```
Traditionally, memset(NULL, value, 0) has been accepted without causing problems.  However, it is not standards-compliant behavior; and for instance Ted Unangst of the OpenBSD project notes that "A smart C compiler may observe a call to memcpy, flag both pointers as valid, and then delete any null checks. Forwards and backwards."
    https://www.tedunangst.com/flak/post/zero-size-objects

Since micropython is using -fdelete-null-pointer-checks ("enabled by default on most targets") and it is probably giving good code size improvements, we have to pay a modest price and add a few checks.

PR #15303, opened 2024-06-19
tools,unix: Enable Undefined Behavior Sanitizer (UBSan) for unix port and in CI.

projectgus:
This PR enables [UBSan](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dundefined) on the coverage build by default, so it will be checked in CI.

Includes two supporting commits to fix UB that appeared in test runs. Kudos to @jepler who did a ton of previous work with UBSan in #7237, #7370, #3799, probably others - hence why this PR is pretty small.

## Null argument checks

The biggest commit in this PR is to fix cases where UBSan detects a null pointer being passed to a function argument decorated as non-null. In every case this was a libc function (memset, memcpy, etc) where the length argument was zero, so a no-op in principle.

@jepler did a good job of explaining the potential risk in #3799 so I'll quote that here:

> Traditionally, `memset(NULL, value, 0)` has been accepted without causing problems.  However, it is not standards-compliant behavior; and for instance Ted Unangst of the OpenBSD project notes that "A smart C compiler may observe a call to memcpy, flag both pointers as valid, and then delete any null checks. Forwards and backwards."
>    https://www.tedunangst.com/flak/post/zero-size-objects
>
> Since micropython is using -fdelete-null-pointer-checks ("enabled by default on most targets") and it is probably giving good code size improvements, we have to pay a modest price and add a few checks.

There are multiple ways we could handle this:

1. Current approach in this PR is to add the checks in the code, and incur a code size increase. It's relatively small, as most of the checks can be optimised out again by the compiler, but it's still there and may be an unacceptably high cost.
2. The approach @jepler chose in #7369 was to have a compile time flag that adds these checks via macros, disabled in production builds to save code size. I think this is an OK approach, but worry that it adds similar source code complexity as 1 without adding protection against the optimisation issue described above - i.e. UBSan builds won't see the UB due to the conditional check, so if it does happen then we'll find it the old fashioned way.
3. Disable all non-null argument checks in UBSan. In practice MicroPython really only encounters these when calling libc functions, so there's a similar benefit and risk as approach 2 but the code stays simpler. This was actually my first choice, there's a branch at https://github.com/projectgus/micropython/tree/feature/ubsan_unix_skip_nonnull that implements it. Then I got nervous about the risk of not checking for it...

@jepler I'd be very interested in your thoughts on this, if you're around. I can see benefits for each approach.

### DIFF
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -91,6 +91,17 @@ ifndef DEBUG
 CFLAGS += -U _FORTIFY_SOURCE
 endif
 
+# Optional Undefined Behavior Sanitizer
+#
+# Note that if running MicroPython with UBSan outside of the Makefile,
+# the environment variable UBSAN_OPTIONS should be set as shown below.
+UBSAN ?= 0
+ifeq ($(UBSAN),1)
+CFLAGS += -fsanitize=undefined
+LDFLAGS += -lubsan
+export UBSAN_OPTIONS=suppressions=$(TOP)/tools/ubsan-suppressions.txt
+endif
+
 # On OSX, 'gcc' is a symlink to clang unless a real gcc is installed.
 # The unix port of MicroPython on OSX must be compiled with clang,
 # while cross-compile ports require gcc, so we test here for OSX and
--- a/ports/unix/README.md
+++ b/ports/unix/README.md
@@ -73,6 +73,21 @@ deplibs`. To actually enable/disable use of dependencies, edit the
 options. For example, to build the SSL module, `MICROPY_PY_SSL` should be
 set to 1.
 
+Build Variants
+--------------
+
+The default `standard` variant is suitable for most uses. Additional build
+variants can be found in the `variants` subdirectory and built by setting
+`VARIANT=name` environment variable (i.e. `make test VARIANT=coverage`).
+
+Note that each variant is built in its own build directory.
+
+Additional variants include:
+
+* `minimal` variant has minimum code size and reduced features.
+* `coverage` variant enables almost all possible features and is suitable for
+  code coverage testing.
+
 Debug Symbols
 -------------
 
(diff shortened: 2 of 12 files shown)


ports/unix/Makefile
```diff
@@ -91,6 +91,17 @@ ifndef DEBUG
 CFLAGS += -U _FORTIFY_SOURCE
 endif
 
+# Optional Undefined Behavior Sanitizer
```
projectgus, line 94: In a display of pedantry befitting a computer programmer, I've used US English spelling for the proper noun Undefined Behavior Sanitizer but used British English spelling elsewhere. If this is too silly then I can change it.

projectgus:
Few things still needed:

* coverage 32-bit build breaks when UBSan enabled. Weirdly in CI it prints an (odd) warning for a read overflow. Locally building with UBSan causes a segfault on `nlr_raise`. Guessing UBSan is inserting checks into at least one of the pseudo-assembler NLR C functions, but so far haven't found a root cause...
* The unix coverage builds for qemu mips and qemu_arm have toolchains that don't ship libubsan. Might be able to cross-install the Ubuntu packages for the respective architectures, but probably simpler to disable UBSAN except for x86/x64.

jepler:
I don't have any detailed comments but thank you for picking up this work and trying to move it forward!

projectgus:
> stm32:   +64 +0.016% PYBV10
>      mimxrt:   +64 +0.018% TEENSY40
>         rp2:  +128 +0.015% RPI_PICO_W
>        samd:   +64 +0.024% ADAFRUIT_ITSYBITSY_M4_EXPRESS

Hmm OK, this is more code size increase than I saw in my testing as well (GCC version dependent, maybe?) So I think something else to tackle here...

projectgus:
Some notes-to-self for when I get back to this:

* Damien is on board with changing the coverage build to setjmp/longjmp for NLR, so we don't need another build and test config. Native NLR is still tested in the standard builds.
* Some builds got smaller from the extra NULL checks, so it may be worth digging into which checks add vs remove code size.

jepler:
There's standards work afoot to change the status of zero-length memcpy calls & others that I think would mean micropython usage was no longer undefined behavior. However, this will require a compiler in "C2y" mode, i.e., with support for the not yet finalized next ISO C standard.

Article:
https://developers.redhat.com/articles/2024/12/11/making-memcpynull-null-0-well-defined

Committee document:
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3322.pdf

projectgus:
This has been picked up again in #17414 so closing in favour of that PR.

PR #15322, opened 2024-06-20
Update modhashlib.c : adler32 checksum.

massimosala:


andrewleech:
hashlib is intended to be a subset of the official cpython hashlib, as are all builtins that have a matching name cpython module.

mattytrentini:
Yes, as @andrewleech has mentioned, this should be [added to zlib, same as in CPython](https://docs.python.org/3/library/zlib.html#zlib.adler32) and the signature needs to match. 

That adds a slight complication since it's recommended that [deflate should be preferred to zlib](https://docs.micropython.org/en/latest/library/zlib.html) for compression - and so the zlib library is less likely to be included. May be worth a discussion to carefully specify `#define`'s to allow for `adler32` to be included separate from `compress`/`decompress`.

It also needs documentation.

PR #15336, opened 2024-06-24
ports: rework board variant support so it requires a `mpconfigvariant_XXX.mk/cmake` file

dpgeorge:
This PR reworks board variants on all ports that currently support it: esp32, esp8266, rp2, stm32.

It's a simple change that moves the board variant configuration from an if statement within `mpconfigboard.mk/cmake` into separate files for each variant `mpconfigvariant_XXX.mk/cmake`.  There are two benefits to this:
- The build system now gives an error if the variant that you specified doesn't exist (because the mpconfigvariant file must exist with the variant name you specify).
- No more error-prone if logic needed in the `.mk/cmake` files.

The way you build a variant is unchanged, still via `make BOARD_VARIANT=XXX`.


dpgeorge:
@Gadgetoid FYI

PR #15337, opened 2024-06-24
webassembly/objjsproxy: Lookup attributes without testing they exist.

dpgeorge:
In JavaScript when accessing an attribute such as `obj.attr` a value of `undefined` is returned if the attribute does not exist.  This is unlike Python semantics where an `AttributeError` is raised.  Furthermore, in some cases in JavaScript `attr in obj` can return false yet `obj.attr` is still valid and returns something other than `undefined`.

To more closely match these JavaScript semantics when proxying a JavaScript object through to Python, change the attribute lookup logic on a `JsProxy` so that it immediately attempts `obj.attr` instead of first testing if the attribute exists via `attr in obj`.

This allows JavaScript objects which dynamically create attributes to work correctly on the Python side, with both `obj.attr` and `obj["attr"]`.  Note that `obj["attr"]` already works in all cases because it immediately does the subscript access without first testing if the attribute exists.

As a benefit, this new behaviour matches the Pyodide behaviour.

### DIFF
--- a/ports/webassembly/objjsproxy.c
+++ b/ports/webassembly/objjsproxy.c
@@ -46,8 +46,14 @@ EM_JS(bool, has_attr, (int jsref, const char *str), {
 EM_JS(bool, lookup_attr, (int jsref, const char *str, uint32_t * out), {
     const base = proxy_js_ref[jsref];
     const attr = UTF8ToString(str);
-    if (attr in base) {
-        let value = base[attr];
+
+    // Attempt to lookup the requested attribute from the base object:
+    // - If the value is not `undefined` then the attribute exists with that value.
+    // - Otherwise if the value is `undefined` and the `in` operator returns true, then
+    //   that attribute does exist and is intended to have a value of `undefined`.
+    // - Otherwise, the attribute does not exist.
+    let value = base[attr];
+    if (value !== undefined || attr in base) {
         if (typeof value === "function") {
             if (base !== globalThis) {
                 if ("_ref" in value) {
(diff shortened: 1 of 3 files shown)


WebReflection:
> This allows JavaScript objects which dynamically create attributes to work correctly on the Python side, with both obj.attr and obj["attr"]. Note that obj["attr"] already works in all cases because it immediately does the subscript access without first testing if the attribute exists.
> 
> As a benefit, this new behaviour matches the Pyodide behaviour.

OK, if this doesn't break what used to work before, I think this is good 👍

Actually, every single attribute access should just go for it, imho, the `in` trap is usually an internal implementation for other JS things ... a Proxy is a proxy in JS and if it defines **just a get trap** that's gotta work without needing to be worried about `in`, `hasOwn`, or any other trap used only for introspection.

ports/webassembly/objjsproxy.c
```diff
@@ -46,8 +46,14 @@ EM_JS(bool, has_attr, (int jsref, const char *str), {
...
-    if (attr in base) {
-        let value = base[attr];
+
+    // Attempt to lookup the requested attribute from the base object:
+    // - If the value is not `undefined` then the attribute exists with that value.
+    // - Otherwise if the value is `undefined` and the `in` operator returns true, then
+    //   that attribute does exist and is intended to have a value of `undefined`.
+    // - Otherwise, the attribute does not exist.
+    let value = base[attr];
+    if (value !== undefined || attr in base) {
```
WebReflection, line 56: yes, nailed it! Thank you!

dpgeorge:
> OK, if this doesn't break what used to work before, I think this is good

Correct, this should not break any existing code.



> Actually, every single attribute access should just go for it

Makes sense.  With this PR that will be how it works for both `obj.attr` and `obj["attr"]`8

WebReflection (approved):

Issue #15298, opened 2024-06-18
rp2: Nonsense REPL input on Linux due to tty serial input echo race


projectgus:
### Port, board and/or hardware

rp2 port, PICO and WeACTStudio boards

### MicroPython version

MicroPython v1.24.0-preview.39.g411d66586c on 2024-06-18; WeAct Studio RP2040 with RP2040
(current master)

### Reproduction

1. Copy this code block to main.py on board:

```
from machine import Pin
import time
led = Pin(Pin.board.LED, Pin.OUT)
for i in range(100):
    print("abcdefghijklmnopqrstuvwxyz")
time.sleep_ms(100)
led(1)  # connect while LED is on to trigger bug
for i in range(100):
    print("ABC ", i)
    time.sleep_ms(50)
led(0)
```

2. Connect the board to USB or hard reset it.
3. While the LED is on, connect with "mpremote a0" or similar.

### Expected behaviour

Since PR #14485 merged, expected output is something like:

```
[... similar lines cut from here ...]
ABC  96
ABC  97
ABC  98
ABC  99
MicroPython v1.24.0-preview.39.g411d66586c on 2024-06-18; WeAct Studio RP2040 with RP2040
Type "help()" for more information.
>>>
```

### Observed behaviour

This is a race condition so it depends on timing, but on Linux the output sometimes shows the REPL behaving as if some random input taken from the earlier output has been sent to it, i.e.

```
[... similar lines cut from here ...]
ABC  96
ABC  97
ABC  98
ABC  99
MicroPython v1.24.0-preview.39.g411d66586c on 2024-06-18; WeAct Studio RP2040 with RP2040
Type "help()" for more information.
>>> stuvwxyz
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'stuvwxyz' isn't defined
>>> 
>>> abcdefg
```

If the output printed earlier from main.py includes valid Python statements, these might even be executed...

### Additional Information

The on-the-wire behaviour can be seen in usbmon capture output where the USB captures

1. The host sets DTR, RTS via USB-CDC control transfers.
2. The rp2 sends all of the buffered data to the host (correctly)
3. The host unexpectedly echoes back some of the data to the rp2, one character at a time.
4. If step 3 includes a newline then the device might try to evaluate it in the REPL, and returns an error.

Output of `strace -tt mpremote a0` shows the last part of the puzzle:

```
[...]

PR #15339, opened 2024-06-24
shared/tinyusb/mp_usbd_cdc: Fix short CDC TX timeouts.

dpgeorge:
The `mp_event_wait_ms()` function may return earlier than the requested timeout, and if that happens repeatedly (eg due to lots of USB data and IRQs) then the loop waiting for CDC TX FIFO space to become available may exit much earlier than MICROPY_HW_USB_CDC_TX_TIMEOUT, even when there is no space.

Fix this by using `mp_hal_ticks_ms()` to compute a more accurate timeout.

The `basics/int_big_mul.py` test fails on RPI_PICO without this fix.

### DIFF
--- a/shared/tinyusb/mp_usbd_cdc.c
+++ b/shared/tinyusb/mp_usbd_cdc.c
@@ -102,9 +102,9 @@ mp_uint_t mp_usbd_cdc_tx_strn(const char *str, mp_uint_t len) {
             n = CFG_TUD_CDC_EP_BUFSIZE;
         }
         if (tud_cdc_connected()) {
-            int timeout = 0;
             // If CDC port is connected but the buffer is full, wait for up to USC_CDC_TIMEOUT ms.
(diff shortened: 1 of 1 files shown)


dpgeorge:
@andrewleech @projectgus FYI

andrewleech:
Ah that makes sense, thanks the fix looks good to me!

robert-hh:
I do not know if that was the intention, but it does not change the odd behavior which I brought up in the discussion PR #15158 and to which @projectgus referred in issue #15298.

andrewleech:
> I do not know if that was the intention, but it does not change the odd behavior which I brought up in the discussion PR #15158 and to which @projectgus referred in issue #15298.

No I wouldn't expect this to affect / help with that one.
The fix for that one is still going to be somewhat complicated I feel, likely needing further changes in tinyusb itself to support.

projectgus:
LGTM!

PR #15345, opened 2024-06-25
rp2,docs: Stop machine.idle() blocking indefinitely, update docs.

projectgus:
# Summary

Updates rp2 port to always resume from idle within 1ms max.

When rp2 port went tickless the behaviour of machine.idle() changed as there is no longer a tick interrupt to wake it up every millisecond. On a quiet system without interrupts it can now block indefinitely. No other port does this.

A lot of existing code (i.e. micropython-lib lps22h, lcd160cr sensor drivers, lora sync_modem driver, usb-device-hid) calls `machine.idle()` inside a tight loop that is polling some condition. This reduces the power usage compared to constantly looping, but can be faster than calling a sleep function. However there's not always an interrupt before the condition they are polling for, so it's difficult to restructure this code if `machine.idle()` doesn't have any upper limit on execution time.

Also changes the machine module documentation to explain the new behaviour and recommended usage. I believe all ports will pause for at most 1ms now, and suggest this is good behaviour to stick with.

This fix depends on #15398 for correct behaviour, and that PR should be merged before this.

# Testing

Tested on rp2 PICO and PICO_W:

* Added a unit test to verify the `machine.idle()` wakeup time. This test fails on master (30ms average wakeup time on PICO-W, never completes on PICO). The test also fails in this branch without the linked PR (average wakeup time of 0.02ms). Passes after this branch is rebased with the WFE fix.
* Ran `while True: machine.idle()` and verified power consumption was approximately the same as tunning `time.sleep(10)`. This is also depends on the linked PR with the WFE fix.

# Trade-offs and Alternatives

* Keeping the current behaviour of "machine.idle() blocks indefinitely on some ports" rules out a lot of use cases, and the power consumption is only very marginally less. Lightsleep is always going to be a better fit for long idle periods, and saves much more power.
* Could add an optional `machine.idle(sleep_ms=1)` argument but it's hard to come up with a use case to justify the extra complexity. This isn't a sleep function so an interrupt can resume execution at any time, and looping with 1ms periods has basically the same power consumption as looping with 1000ms idle periods.


dpgeorge:
I can confirm this change increases power consumption... here's my test results on a RPI_PICO:

| code | master | this PR |
|-|-|-|
| idle REPL | 15.1 mA | 15.1 mA |
| while True: pass | 17.7 mA | 17.7 mA|
| while True: machine.idle() | 15.1 mA | 19.0 mA|
| while True: sleep_ms(1) | 17.7 mA | 17.7 mA|
| while True: sleep_ms(10) | 17.7 mA | 17.7 mA|

That's the current drawn over the USB.  I run the command and then exit the terminal program (mpremote repl) because there's about a 0.5mA difference between USB CDC active and inactive.

The results show that:
1. This PR makes `machine.idle()` worse than a tight loop!
2. The existing `time.sleep_ms()` calls are also broken because they are as bad as a tight loop.

Going back to v1.23.0, point (2) is fixed, ie `time.sleep_ms()` is the same as idle REPL.

So, we need to investigate what went wrong since v1.23.0.

projectgus:
> So, we need to investigate what went wrong since v1.23.0.

The regression is 3af006efb39. Calling `recursive_mutex_exit()` calls the default [lock_internal_spin_unlock_with_notify()](https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/hardware_sync/include/hardware/sync.h#L115) implementation which executes SEV after unlocking. This is intended to wake up the other CPU if it's executed WFE while waiting for the lock. However, if there's no contention for the lock then it causes the next WFE to resume immediately.

Because this happens in `pendsv_resume()` now, it happens on every call to `mp_wfe_or_timeout()` and the WFE resumes immediately every time. :facepalm: 

I actually don't know what the best solution is here because I don't 100% understand the relevant RP2040 docs. [Have asked for clarification on the Raspberry Pi forums](https://forums.raspberrypi.com/viewtopic.php?t=373094).

projectgus:
This PR is ready now but depends on https://github.com/micropython/micropython/pull/15398 for correct behaviour.

projectgus:
@dpgeorge Rebased to pick up the WFE fix, and split into two commits. I ended up making the new test rp2-only for now as there are several ports it won't pass on.

dpgeorge:
Thanks!  I tested it and the test works well (on rp2, and also stm32 just for my interest).

PR #15347, opened 2024-06-25
py/asmrv32: Make some code sequences smaller.

agatti:
This PR changes a few code sequences to use more compressed opcodes where possible.  The sequences in question are the ones that show up the most in the test suite and require the least amount of code changes, namely short offset loads from memory to RET/ARG registers, indirect calls through the function table, register-based jumps, locals' offset calculation, reg-is-null jumps, and register comparisons.

There are a couple more places where things could be shortened further but they either require some potentially unsafe register rearranging (TEMP0/TEMP1 to A4/A5 and use `c.sw` for short stores and `c.lw` for more short loads), or for some minor cases (use `c.lw`/`c.sw` for load/stores if offsets are larger than 12 bits and have bits 11:8 cleared, not that frequent).  I guess it's a matter of diminishing returns from here on, but if anybody else wants to squeeze the last bytes out of this I won't mind at all.

That said, on average there should be a 15-20% code reduction with no loss of speed (but no gains either), which isn't that bad considering the amount of code involved.

As usual, native and viper tests pass on both QEMU and on an ESP32C3.

### DIFF
--- a/py/asmrv32.h
+++ b/py/asmrv32.h
@@ -151,9 +151,20 @@ void asm_rv32_end_pass(asm_rv32_t *state);
     ((op & 0b1111111) | ((rd & 0b11111) << 7) |                                  \
     (imm & 0b11111111111111111111000000000000))
 
+#define RV32_ENCODE_TYPE_CB(op, ft3, rs, imm) \
+    ((op & 0b11) | ((ft3 & 0b111) << 13) | ((rs & 0b111) << 7) | \
+    (((imm) & 0b100000000) << 4) | (((imm) & 0b11000000) >> 1) | \
+    (((imm) & 0b100000) >> 3) | (((imm) & 0b11000) << 7) | \
+    (((imm) & 0b110) << 2))
+
 #define RV32_ENCODE_TYPE_CI(op, ft3, rd, imm) \
     ((op & 0b11) | ((ft3 & 0b111) << 13) | ((rd & 0b11111) << 7) | \
     (((imm) & 0b100000) << 7) | (((imm) & 0b11111) << 2))
+
+#define RV32_ENCODE_TYPE_CIW(op, ft3, rd, imm) \
+    ((op & 0b11) | ((ft3 & 0b111) << 13) | ((rd & 0b111) << 2) | \
+    ((imm & 0b1111000000) << 1) | ((imm & 0b110000) << 7) | \
+    ((imm & 0b1000) << 2) | ((imm & 0b100) << 4))
 
 #define RV32_ENCODE_TYPE_CJ(op, ft3, imm) \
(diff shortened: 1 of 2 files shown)


py/asmrv32.h
```diff
@@ -151,23 +151,40 @@ void asm_rv32_end_pass(asm_rv32_t *state);
     ((op & 0b1111111) | ((rd & 0b11111) << 7) |                                  \
     (imm & 0b11111111111111111111000000000000))
 
+#define RV32_ENCODE_TYPE_CB(op, ft3, rs, imm) \
+    ((op & 0b11) | ((ft3 & 0b111) << 13) | ((rs & 0b111) << 7) | \
```
dpgeorge, line 155: I only just realised now that you are using binary literals.  AFAIK this is not in the C standard, at least not in C99 which we aim to support.

I think it's only working because gcc (and other compilers) has an extension to support binary literals.

I think we need to use hex literals instead.

agatti: Fair enough, but consider that all known compilers that can build code for RISC-V (GCC, LLVM, IAR, and Segger) have that turned on by default.

Turning all those binary literal into their hex counterparts is not a problem, but I wonder what's the best course of action.  Merge this PR after review and then I submit another one to turn binary literals into hex, or to hold this off for a bit until I send the conversion PR and it gets merged, then I update this one.

I'd rather send a conversion PR that covers both this and the previous one, but if you prefer the other option (or something else entirely) that'd be fine for me too.

dpgeorge: I would prefer to stay away from binary literals (even if they are very neat in situations like this).  And that's because it would be too easy for them to accidentally be copied to code that is not purely for RV32.  We need consistency and it's hard to say "binary literals can only be used for RV32 code".

Furthermore, `mpy-cross` actually builds the RV32 native emitter so it can target all archs.  And that builds with an arbitrary compiler, and can run on any target machine (even in the browser through WASM).  So there may be an issue building `mpy-cross` if someone uses a compiler that doesn't support binary literals.

dpgeorge: > I'd rather send a conversion PR that covers both this and the previous one,

That's OK by me.

dpgeorge:
> I guess it's a matter of diminishing returns from here on, but if anybody else wants to squeeze the last bytes out of this I won't mind at all.

If a MicroPython-based project is using a lot of native code, then optimising the generation of that code is usually worth it, since the cost is a one-off in the emitter but paid back many times over in all the generated code that benefits from that optimisation.  Even better is if the code is generated offline using `mpy-cross`.

But I think what you've done in this PR is a good compromise between complexity/size of the RV32 emitter and optimised code generation.  At least for now.

Thank you!

PR #15349, opened 2024-06-26
webassembly: Reuse PyProxy objects when they are the same Python object.

dpgeorge:
### Summary

This change makes it so that PyProxy objects are reused (on the JavaScript side) when they correspond to an existing Python object that is the same object.  Eg proxying the same Python function to JavaScript, the same PyProxy instance is now used.  That means that if `foo` is a Python function then accessing it on the JavaScript side such as `api.globals().get("foo")` has the property that:
```js
api.globals().get("foo") === api.globals().get("foo")
```

### Testing

Tests added to the webassembly port-specific tests.

TODO:
- [x] test case when there are many proxy objects created, that they still get reclaimed by the finalizer mechanism

This adds some code size, memory use and performance overhead, but they are all minor.  Not making this change means it's hard to write code that expects identity on the JavaScript side when the proxied object on the Python side is identical.


dpgeorge:
@WebReflection FYI

ports/webassembly/proxy_js.js
```diff
@@ -244,6 +246,10 @@ function proxy_convert_mp_to_js_obj_jsside(value) {
     } else {
         // obj
         const id = Module.getValue(value + 4, "i32");
+        const existing_obj = globalThis.proxy_js_dict[id];
+        if (existing_obj !== undefined) {
+            return existing_obj.deref();
```
WebReflection, line 251: watch out `.deref()` doesn't guarantee the proxy is returned and, when that's the case, this branch would fail expectations.

An easier way is to reach out for that proxy instead and keep going if not there as it means nobody used that before but now it's needed again, it should be re-registered and re-assigned.

In code words:

```js
const existing_proxy = globalThis.proxy_js_dict[id]?.deref();
if (existing_proxy) return existing_proxy;
```

Another hint would be to make `proxy_js_dict` a `proxy_js_map` instead as JS is not optimized to `delete` entries from dictionaries but this might be premature optimization too, yet the check would be a bit more elegant via `proxy_js_map.has(id)` and `proxy_js_map.get(id).deref()` but it's OK as it is now, after all.

dpgeorge:
Thanks @WebReflection  for those links.  I think I'm on top of the WeakRef behaviour now, and this PR should address the previous issues about `deref()` returning `undefined`.

Now merged, and available on npmjs at https://www.npmjs.com/package/@micropython/micropython-webassembly-pyscript/v/1.24.0-preview-114

WebReflection:
@dpgeorge thanks! it's already up and running via *npm* *canary* distribution of PyScript https://cdn.jsdelivr.net/npm/@pyscript/core@0.4.56/dist/core.js

dpgeorge:
@WebReflection very good!  Let me know if you see any issues, this was a pretty deep change.

WebReflection:
all automation tests are green and I could close a related bug without issues https://github.com/pyscript/pyscript/issues/2108#issuecomment-2238947849

so far, so good!

PR #15350, opened 2024-06-26
tools/mpy-tool.py : Fix "freezing int to long-long is not implemented".

srinsoz:
### Summary

Allow inclusion of large integer constants in frozen files using longlong representation  (mpy-cross option -mlongint-impl=longlong)

### Testing

use mpy-cross with  option -mlongint-impl=longlong


dpgeorge:
Thanks, this looks good!

PR #11143, opened 2023-03-28
ports/rp2: Use nano.specs for C++ modules.

Gadgetoid:
Remove C++ stack unwinding and exception handling in all cases on user C++ modules. There's no hook to surface exceptions to MicroPython so they're basically useless?

Set options as per https://github.com/raspberrypi/pico-sdk/blob/master/src/rp2_common/pico_cxx_options/CMakeLists.txt

Despite the options linked above, without the addition of `-specs=nano.specs`, this doesn't make much difference. Since Pico SDK sets `nosys.specs` elsewhere this is slightly odd, but I'm in a little over my head. Does anyone know what could be happening here?

Save around 10,880 bytes of RAM in our Pico builds (with a number of C++ modules) and returns it to MicroPython's heap for end-users.

PR #15353, opened 2024-06-26
C++ Compatibility For Minimal Necessary Header Files

feeqra:
### Summary

This Pull Request adds C++ functionality to the minimal set of header files needed to integrate MicroPython into a Zephyr project using C++. Without these changes all users with C++ applications must manually add C++ compatibility themselves.

### Testing

Testing was done on an out-of-tree custom ARM board running Zephyr. With these changes, the code was built successfully and MicroPython ran correctly.

### DIFF
--- a/py/builtin.h
+++ b/py/builtin.h
@@ -40,6 +40,10 @@ typedef enum {
 
 #define mp_builtin_open_obj mp_vfs_open_obj
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 mp_import_stat_t mp_vfs_import_stat(const char *path);
 mp_obj_t mp_vfs_open(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs);
 
@@ -138,5 +142,9 @@ extern const mp_obj_module_t mp_module_uctypes;
 extern const mp_obj_module_t mp_module_machine;
 
 extern const char MICROPY_PY_BUILTINS_HELP_TEXT[];
+
+#ifdef _cplusplus
+}
+#endif
 
 #endif // MICROPY_INCLUDED_PY_BUILTIN_H
--- a/py/compile.h
+++ b/py/compile.h
@@ -30,6 +30,10 @@
 #include "py/parse.h"
 #include "py/emitglue.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 #if MICROPY_COMP_ALLOW_TOP_LEVEL_AWAIT
 // set to `true` to allow top-level await expressions
 extern bool mp_compile_allow_top_level_await;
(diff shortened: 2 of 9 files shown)


stinos:
The idea is sound, but this implementation isn;t ideal. For instance: imo external code shouldn't include py/stackctrl.h directly, but py/runtime.h instead. On the other hand here's the list of headers I've been using for embedding the windows and unix ports:

```
#ifdef __cplusplus
extern "C"
{
#endif
#include <py/objfun.h>
#include <py/objint.h>
#include <py/objmodule.h>
#include <py/objtype.h>
#include <py/runtime.h>
#ifdef __cplusplus
}
#endif
```

and some of thes are lacking here. So I think either the situation remains as-is, or we'd have to settle on which files form the toplevel api and then modify those. Or maybe even provide one common header which is C++ compatible, but that has the (slight) disadvantage that it's going to include more than needed for most users.

Rafeeq-Muhammad:
I think we should provide one common header file which is C++ compatible. It's my understanding that the developers who need C++ compatibility could just remove the unnecessary header files that they don't need. If that's the case, in my opinion this outweighs the developers having to manually implement C++ compatibility for the files they do need. I'll have to look into this and ask the discord.

Rafeeq-Muhammad:
I'm unclear how to give this PR a Zephyr tag. Also I'd like to mention that this PR doesn't attempt to implement C++ compatibility for all ports, just Zephyr.

dpgeorge:
> I'm unclear how to give this PR a Zephyr tag. Also I'd like to mention that this PR doesn't attempt to implement C++ compatibility for all ports, just Zephyr.

The changes here are not zephyr specific, they apply to everything (all ports), so they need to work with everything.

In general this should be usable by anyone embedding MicroPython in a C++ project.

Gadgetoid:
I've been embedding C++ code into MicroPython builds since https://github.com/micropython/micropython/commit/0cf12dd59c9ddddbd602d4267410033cb5a9d265, this being the core of our (probably not wise in retrospect) software strategy at Pimoroni.

I've found that introducing C++ code into MicroPython also brings a lot of C++ baggage and footguns- while these may be manageable for a seasoned C++ developer, wider support could open the gates for more naive user (such as me!) to grapple with the problems that I did and still do.

If C++ is "officially" endorsed by this change and works across ports, I think documentation should also be updated to reflect these pitfalls.

Chiefly there are two issues that have plagued me:

* malloc/new, including allocations that happen at runtime via static initialization (std::vector etc)  - need to document that `MICROPY_C_HEAP_SIZE > 0` (or similar) is required for C++ builds not to potentially explode/trounce `gc_heap` - I don't know if the Zephyr port or others have this concept
* C++ Stack unwinding/exception handling bloat - I've been using a hack in my `CMakeLists.txt`  to stop our (already very overloaded) builds exploding with C++ bloat.  I'd love more eyes on why this is and how to fix it - see: https://github.com/micropython/micropython/pull/11143

Note: I also use a C++ module which redirects C++ malloc/free to MicroPython's GC_HEAP and allows me to track allocations. It's not ideal but it's a little more debuggable than MicroPython's "eh just throw C some space, it'll be fine" approach. You can see it in its full horror here: https://github.com/pimoroni/pimoroni-pico/blob/main/micropython/modules/cppmem/cppmem.cpp

I guess embedding MicroPython into something else is rather an about-face of my memory management problems, but I figure my painfully earned two cents would be better said than unsaid here.

stinos:
> need to document that MICROPY_C_HEAP_SIZE > 0 (or similar) is required for C+

Didn't know what that option was so found out it's for rp2 only, to be fair it says `If a board uses malloc then it must set this to at least 4096.` so that's pretty close already

PR #15355, opened 2024-06-27
stm32: allow building with -O2 without warnings with latest gcc

dpgeorge:
### Summary

Compiling any stm32 board with -O2 using arm-none-eabi-gcc 14.1.0 leads to warnings.

This PR fixes those warnings, either disabling them or tweaking the code.

### Testing

PYBD_SF6 now builds with `make BOARD=PYBD_SF6 COPT=-O2` without any warnings.

And CI has been updated to build this board with -O2.


ports/stm32/pin.h
```diff
@@ -45,8 +45,8 @@ typedef struct {
 typedef struct {
     mp_obj_base_t base;
     qstr name;
-    uint32_t port   : 4;
-    uint32_t pin    : 5;    // Some ARM processors use 32 bits/PORT
+    uint32_t port   : 4; // Allows GPIOA through GPIOP
+    uint32_t pin    : 4; // ST MCUs have a maximum of 16 pins pert port
```
dlech, line 49:
```suggestion
    uint32_t pin    : 4; // ST MCUs have a maximum of 16 pins per port
```

dpgeorge: Thanks!  Now fixed.

PR #15359, opened 2024-06-27
ports/rp2: Replace CMSIS with Pico SDK equivalents.

Gadgetoid:
This change replaces a couple of CMSIS holdouts with their Pico SDK equivalents. AFAICT these are functionally identical, fancy decorators around their ASM counterparts.

I have updated the comment in `mphalport.h` to clarify what's still coming from CMSIS. I believe `__WFI` has long since been replaced with `__wfi()` so `__DSB()` and `__SEV()` should probably follow suit.

Let me know if there's a reason things are as they are (obviously wrong stale comment notwithstanding.)

It might be worth evaluating what bumping the priority of `PendSV_IRQn` and https://github.com/micropython/micropython/blob/b4213c9c920afb8da5d3d123b26527e26dceb877/ports/rp2/main.c#L77C8-L77C39 are intended to accomplish, and whether this is generalisable and could make a good case for a Pico SDK function. 

I'm a little over my head, and probably fighting a losing battle, but the CMSIS inclusion feels like an ugly wart.

### DIFF
--- a/ports/rp2/mphalport.h
+++ b/ports/rp2/mphalport.h
@@ -30,7 +30,6 @@
 #include "pico/time.h"
 #include "hardware/clocks.h"
 #include "hardware/structs/systick.h"
-#include "RP2040.h" // cmsis, for NVIC_SetPriority, PendSV_IRQn and SCB/SCB_SCR_SEVONPEND_Msk
 #include "pendsv.h"
 
 #define SYSTICK_MAX (0xffffff)
--- a/ports/rp2/mphalport.h
+++ b/ports/rp2/mphalport.h
@@ -30,7 +30,7 @@
 #include "pico/time.h"
 #include "hardware/clocks.h"
 #include "hardware/structs/systick.h"
-#include "RP2040.h" // cmsis, for __WFI
+#include "RP2040.h" // cmsis, for NVIC_SetPriority, PendSV_IRQn and SCB/SCB_SCR_SEVONPEND_Msk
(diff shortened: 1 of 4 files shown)


ports/rp2/mphalport.h
```diff
@@ -30,7 +30,7 @@
 #include "pico/time.h"
 #include "hardware/clocks.h"
 #include "hardware/structs/systick.h"
-#include "RP2040.h" // cmsis, for __WFI
+#include "RP2040.h" // cmsis, for NVIC_SetPriority, PendSV_IRQn and SCB/SCB_SCR_SEVONPEND_Msk
```
dpgeorge, line 33: I agree including this here is ugly.

With your changes, is it possible to move the inclusion of this header to the .c files that need it?  That would be less ugly.

Gadgetoid: I don't see why not. I'll give it a shot. It will also help prevent future (unintentional) reliance on CMSIS defines over the Pico SDK equivalents.

Gadgetoid: Changed, rebased and push'd. I kept separate commits for each change but do with them as you please!

dpgeorge: Looks good now, a lot cleaner.

dpgeorge:
Rebased, squashed and merged in 0e261443be9a9201fd786b54aa2d0198f21543f7

PR #15385, opened 2024-06-30
esp32/network_lan: LAN.active(True) should succeed if already active.

glenn20:
This PR ensures that `network.LAN.active(True/False)` will not raise an exception if the LAN is already in the desired state.

### Summary

Currently, `lan.active(True)` will raise an `OSError` exception if the LAN is already in the desired state. 

This is inconsistent with `network.WLAN.active(True/False)` and causes `lan.active(True)` to raise an exception after a soft reset (causing commonly used network startup scripts to fail for LAN interfaces after a soft reset).

### Testing

I have tested this on a [LILYGO-T-ETH-LITE ESP32S3](https://www.lilygo.cc/products/t-eth-lite) board using the W5500 LAN driver.

The PR checks the return status of esp_eth_start/stop() and does not raise exception on ESP_ERR_INVALID_STATE as this is [documented as indicating the device is already in the desired state](https://docs.espressif.com/projects/esp-idf/en/v4.2/esp32/api-reference/network/esp_eth.html#_CPPv413esp_eth_start16esp_eth_handle_t).

While an exception is no longer raised the espressif drivers still print a warning message:

- `E (677173) esp_eth: esp_eth_start(286): driver started already`.

### DIFF
--- a/ports/esp32/network_lan.c
+++ b/ports/esp32/network_lan.c
@@ -312,13 +312,14 @@ static mp_obj_t lan_active(size_t n_args, const mp_obj_t *args) {
     lan_if_obj_t *self = MP_OBJ_TO_PTR(args[0]);
 
     if (n_args > 1) {
-        if (mp_obj_is_true(args[1])) {
+        bool make_active = mp_obj_is_true(args[1]);
+        if (make_active && !self->base.active) {
             esp_netif_set_hostname(self->base.netif, mod_network_hostname_data);
             self->base.active = (esp_eth_start(self->eth_handle) == ESP_OK);
             if (!self->base.active) {
                 mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("ethernet enable failed"));
             }
-        } else {
+        } else if (!make_active && self->base.active) {
             self->base.active = !(esp_eth_stop(self->eth_handle) == ESP_OK);
             if (self->base.active) {
                 mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("ethernet disable failed"));


elvis-epx:
Similar/equal to https://github.com/micropython/micropython/pull/15101

glenn20:
@elvis-epx Thanks. I missed your PR (I must  not have looked as hard as I thought).

I note that with your PR, it appears that `self->base.active` will be set to `False` after calling `LAN.active(True)` (if the interface was already active). This will cause a following call to `isconnected()` to always return `False`.

If you are happy to address that, I will close this PR in deferrence to yours.

elvis-epx:
Good catch @glenn20, have closed my PR in favor of yours.

dpgeorge:
> While an exception is no longer raised the espressif drivers still print a warning message

That doesn't seem right.  I think we should try to make it so no warnings are printed at all.

For example, is it possible to use `self->base.active` to determine if the IDF start/stop should be called?  Something like:
```c
make_active = mp_obj_is_true(args[1]);
if (make_active && !self->base.active) {
    self->base.active = (esp_eth_start(self->eth_handle) == ESP_OK);
    ...
} else if (!make_active && self->base.active) {
    self->base.active = !(esp_eth_stop(self->eth_handle) == ESP_OK);
    ...
}
```
Would that work?

glenn20:
Yes, it will help in the common case - I will fix that.

Though it wont prevent the messages after a soft reset as we have no way of determining if the ethernet is already active or not.

glenn20:
@dpgeorge - I have changed the code to (and rebased):

```py
        esp_err_t err = ESP_OK;
        bool make_active = mp_obj_is_true(args[1]);
        if (make_active && !self->base.active) {
            esp_netif_set_hostname(self->base.netif, mod_network_hostname_data);
            err = esp_eth_start(self->eth_handle);
            self->base.active = (err == ESP_OK || err == ESP_ERR_INVALID_STATE);
            if (!self->base.active) {
                mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("ethernet enable failed"));
            }
        } else if (!make_active) {
            err = esp_eth_stop(self->eth_handle);
            self->base.active = !(err == ESP_OK || err == ESP_ERR_INVALID_STATE);
            if (self->base.active) {
                mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("ethernet disable failed"));
            }
        }
```

which prevents the warning message in the common case that `lan.active(True)` is called on an already active interface (except after a soft reset).

We can't guard against repeated calls to `lan.active(False)` without making it impossible to deactivate the interface after a soft reset (without calling `lan.active(True)` first, or some other logic to detect soft resets).

Without the checks for ESP_ERR_INVALID_STATE, exceptions are raised after a soft reset and the interface is already active.

Tested on LILYGO T-ETH-LITE device.

glenn20:
Doh. While updating the LAN MAC address PR, I realised that the `network.LAN` singleton is statically allocated (not from the heap), so we **CAN** rely on `self->base.active` persisting across soft resets and that your suggested simpler fix should work fine. I will test to confirm.

glenn20:
> ```c
> make_active = mp_obj_is_true(args[1]);
> if (make_active && !self->base.active) {
>     self->base.active = (esp_eth_start(self->eth_handle) == ESP_OK);
>     ...
> } else if (!make_active && self->base.active) {
>     self->base.active = !(esp_eth_stop(self->eth_handle) == ESP_OK);
>     ...
> }
> ```
> 
> Would that work?

Yes, that works and I have pushed a change to implement that (now that I realise that `self->base.active` is persistent across soft resets).

dpgeorge:
> Yes, that works and I have pushed a change to implement that (now that I realise that `self->base.active` is persistent across soft resets).

Ah, yeah, I was wondering (in the context of the PR here) why the `if (self->initialized) {` bit in `get_lan()` worked and didn't reinstall the LAN driver after soft reset, and hence give a warning about double installation.

So, this PR is much better now, it shouldn't ever warn, no matter how many times you call `active()` with true or false.

dpgeorge:
Sorry, this got lost.  Now merged.

PR #15389, opened 2024-07-01
py/asmrv32: Do not use binary literals.

agatti:
### Summary

As per discussion in https://github.com/micropython/micropython/pull/15347#discussion_r1658693954, non-standard binary literals have been removed in favour of their hexadecimal counterparts.

### Testing

This was tested with the Qemu RISC-V port.


dpgeorge:
Thanks for this.  It looks good although I did not check all of the bin->hex rewrites.  I trust the tests.

PR #15392, opened 2024-07-02
Enable abort support

RealJohnSmith:
### Summary

This PR updates previous PR/discussion at: https://github.com/micropython/micropython/pull/14419 but in this new PR, the code is rebased on top of the current master.

### Testing


shared/runtime/pyexec.c
```diff
@@ -143,15 +155,41 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
...
-        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) {
-            // at the moment, the value of SystemExit is unused
-            ret = pyexec_system_exit;
-        } else {
+        #if MICROPY_ENABLE_VM_ABORT
+        if (nlr.ret_val == NULL) { // abort
+            ret = pyexec_abort;
+        } else
+        #endif
+        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) { // system exit
```
dpgeorge, line 163: This updated `SystemExit` handling code (the second commit in the PR) adds quite a bit of code size to existing builds, which don't really use the functionality.

How useful is this new code?  Does it make sense to guard it with a compile-time configuration flag?

dlech: Or perhaps to make it even more flexible, use a macro similar to other port-specific hooks.

I.e. `MICROPY_VM_HOOK_ABORT`. This way, the code can be customized without modifying `pyexec.c`.

RealJohnSmith: I am not sure how to do the part with the hook. What do you mean, can you please explain?

RealJohnSmith: It doesnt feel like a lot of code size to me, but I can add some compile time guards, that is not a problem

RealJohnSmith: I implemented the guard using a new macro "MICROPY_ENABLE_EXIT_CODE_HANDLING"

if set to 0, the behavior should be the same as before. As well as code size. This means, the exit code is written into the variable `pyexec_system_exit` internally, before the SystemExit exception is raised (e.g. soft reset method).

if set to 1, the values defined in macros such as `PYEXEC_KEYBOARD_INTERRUPT` are used for the appropriate exit value.

RealJohnSmith:
What do you think, @dpgeorge @dlech? Is this ok for merge?

shared/runtime/pyexec.c
```diff
@@ -45,6 +45,11 @@
 
 pyexec_mode_kind_t pyexec_mode_kind = PYEXEC_MODE_FRIENDLY_REPL;
 int pyexec_system_exit = 0;
+#if MICROPY_ENABLE_VM_ABORT
+int pyexec_abort = 0; // kept at 0 for backward compatibility
+#endif
+int pyexec_keyboard_interrupt = 0; // kept at 0 for backward compatibility
```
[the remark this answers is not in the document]
RealJohnSmith: we dont, I forgot it there

shared/runtime/pyexec.c
```diff
@@ -45,6 +45,11 @@
 
 pyexec_mode_kind_t pyexec_mode_kind = PYEXEC_MODE_FRIENDLY_REPL;
 int pyexec_system_exit = 0;
+#if MICROPY_ENABLE_VM_ABORT
+int pyexec_abort = 0; // kept at 0 for backward compatibility
```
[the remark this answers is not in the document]
RealJohnSmith: Ok, removing the comment. The original value was 0 and I think we should keep it there.

shared/runtime/pyexec.h
```diff
@@ -35,13 +35,33 @@ typedef enum {
...
+//
+// If guard MICROPY_ENABLE_EXIT_CODE_HANDLING is enabled, system exit code will be set
+// according to the reason for exit and whether it was considered a success or not,
+// consistent with python sys.exit() call - https://docs.python.org/3/library/exceptions.html#SystemExit
 extern int pyexec_system_exit;
 
+#if MICROPY_ENABLE_VM_ABORT
+// Set this to the value (eg PYEXEC_ABORT) that will be propagated through
+// the pyexec functions if the VM is aborted for immediate exit.
+// It will reset to 0 at the start of each execution (eg each REPL entry).
```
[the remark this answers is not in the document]
RealJohnSmith: Here:
https://github.com/micropython/micropython/blob/d2bfc85fcb43412ab703fe6207ace6021a5d373c/shared/runtime/pyexec.c#L82-L83

This is the original part (e.g. relevant without the `MICROPY_ENABLE_EXIT_CODE_HANDLING`) I didn't modify it in any way.

RealJohnSmith:
If I understand correctly the conflict that occurred since the last time, the `pyexec_system_exit` was removed. I removed it as well, which also solves the comments you had

RealJohnSmith:
Can we merge it now?

dpgeorge:
Superseded by #15730.

PR #15393, opened 2024-07-02
ports/rp2: Replace PIO_NUM with pio_get_index.

Gadgetoid:
The `PIO_NUM` macro had been defined ~3 years ago when `rp2_pio.c` was first conceived.

There's now a Pico SDK function for this `pio_get_index()`, which is already used in some parts of `rpi_pio.c`.

This change:

* removes `PIO_NUM` in favour of using `pio_get_index()` everywhere.
* fixes `rp2_pio_print` to use `pio_get_index()` too, since it had its own copy of the ternary expression
* remove a ternary from `rp2_pio_state_machine` and calculate it from `pio_get_index`. If it's good enough for `pio_irq0` it's good enough for here.
*  Remove a ternary on `GPIO_FUNC_PIO0` vs `GPIO_FUNC_PIO1` these are, helpfully, sequentially ordered so we can calculate them too.

### DIFF
--- a/ports/rp2/rp2_pio.c
+++ b/ports/rp2/rp2_pio.c
@@ -37,8 +37,6 @@
 #include "hardware/irq.h"
 #include "hardware/pio.h"
 
-#define PIO_NUM(pio) ((pio) == pio0 ? 0 : 1)
-
 typedef struct _rp2_pio_obj_t {
     mp_obj_base_t base;
     PIO pio;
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks, looks good.  Rebased, squashed some, and merged in 462fa5f94ff7a90afb42e72ce145309fa671ddce and f61fac0ba6e2684b8bb4b11c7a768b0cb805c198

(And thanks for keeping the commits separate.)

PR #15395, opened 2024-07-02
esp32/mpconfigport: Enable the RV32 emitter for ESP32C3 targets.

agatti:
### Summary

The RV32 code emitter assumed that the arch-specific NLR was used instead of the setjmp/longjmp based NLR code.  If the latter NLR provider was chosen, the emitter would allocate space on the stack for the NLR buffer but would not fill it in.

This change makes the emitter backend aware of the setjmp-based NLR being enabled and makes sure the native calls to setjmp() are added to the generated code.

### Testing

This was tested on the QEMU RISC-V port using the default configuration and `MICROPY_NLR_SETJMP` enabled, and on an ESP32C3 with the ESP32 port using the default configuration and `MICROPY_EMIT_RV32` enabled.

This adds up to 16 bytes to each generated RV32 function that requires setjmp-based NLR (adds two native calls to MP_F_SETJMP in the function setup code).  The main alternative would be to stop compilation if both the setjmp-based NLR and the RV32 emitter are requested at the same time, but that would add a bit more complexity on the port configuration/documentation side.  As an aside, the RV32-specific NLR should probably be used anyway as it takes up 64 bytes on the stack as opposed to newlib's/picolibc's 304 bytes, but that's more of an issue on RV32 MCUs with smaller RAM space than what an ESP32C3 has.

I have limited code changes to the RV32 emitter, even though the Xtensa windowed emitter unconditionally brings in setjmp support in `py/emitnxtensawin.c`.  I wasn't sure on whether said emitter would ever work without a setjmp NLR so I left it alone.

### DIFF
--- a/ports/esp32/mpconfigport.h
+++ b/ports/esp32/mpconfigport.h
@@ -18,9 +18,8 @@
 
 // object representation and NLR handling
 #define MICROPY_OBJ_REPR                    (MICROPY_OBJ_REPR_A)
+#if !CONFIG_IDF_TARGET_ESP32C3
 #define MICROPY_NLR_SETJMP                  (1)
-#if CONFIG_IDF_TARGET_ESP32C3
-#define MICROPY_GCREGS_SETJMP               (1)
 #endif
 
 // memory allocation policies
@@ -44,6 +43,8 @@
 #define MICROPY_PERSISTENT_CODE_LOAD        (1)
 #if !CONFIG_IDF_TARGET_ESP32C3
 #define MICROPY_EMIT_XTENSAWIN              (1)
+#else
+#define MICROPY_EMIT_RV32                   (1)
 #endif
 
 // workaround for xtensa-esp32-elf-gcc esp-2020r3, which can generate wrong code for loops


dpgeorge:
This won't work well with `mpy-cross` when it generates rv32 code, because the emitter will use `mpy-cross`'s NLR settings, not the target settings.

At the moment we don't have a CLI option to use with `mpy-cross` to specify which NLR implementation to use, it's just assumed that there's a fixed option.  Currently that fixed option is that all bare-metal ports use the custom NLR code, except xtensawin which uses setjmp/longjmp.

A way forward here would be to do:
```c
#if MICROPY_NLR_SETJMP && !MICROPY_DYNAMIC_COMPILER
#define N_NLR_SETJMP (1)
#endif
```

agatti:
> Currently that fixed option is that all bare-metal ports use the custom NLR code, except xtensawin which uses setjmp/longjmp.

I see, I missed that bit.  Given the only officially supported port for the RV32 emitter shares its main configuration file with an Xtensa-based platform I wanted to make it easier for ESP32C3 users to enable the native emitter - just add a define and off you go.

Anyway, maybe this can be a bit off topic but is there a plan to eventually enable the RV32 emitter by default for the ESP32 port when applicable?

In the meantime I'll update the PR with your code suggestion, thank you.

dpgeorge:
I'm having second thoughts about this PR.  The problem is it makes it hard to know if you're on an RV32 system whether setjmp will be used or the native NLR.  And, mpy-cross may generate the wrong native code (eg using NLR, but target uses setjmp).

How about instead, we enable native NLR for ESP32C3?

> Anyway, maybe this can be a bit off topic but is there a plan to eventually enable the RV32 emitter by default for the ESP32 port when applicable?

Yes, we can do that now.

agatti:
Ok, I've made the appropriate changes to `mpconfigport.h` to disable setjmp()-based NLR and GCREGS for the ESP32C3, along with enabling the native emitter by default.

Just to be sure I re-ran the tests on an ESP32C3 board and everything works as expected.

dpgeorge:
Thanks for updating the PR.  I tested the changes on an ESP32C3 and they work well, the native and viper tests in `tests/micropython/` all pass.

dpgeorge:
Running the tests in native emitter mode leads to some crashes, see #15551.

PR #13329, opened 2024-01-03
rp2: Fix hang triggered by timing of short sleeps and soft timer events

projectgus:
The alarm pool on rp2 pico-sdk stops triggering at times, until an additional interrupt is received to wake the CPU from a WFE state. In systems without regular interrupts this can hang indefinitely. See #12873, and also linked pico-sdk issue https://github.com/raspberrypi/pico-sdk/issues/1552.

This PR removes most MicroPython dependencies on the pico-sdk alarm pool, as a fix for the readily reproducible versions of this bug:

* Soft timer is refactored to use a standalone hardware timer alarm. To support this, `pendsv_suspend()` & `pendsv_resume()` is made safe to run from their core, as these functions protect access to the soft timer data structures.
* Functions for sleeping & waiting for events are refactored to use the soft timer instead of pico-sdk functions.

There are two remaining places that MicroPython uses the pico-sdk alarm pool on rp2:

* `machine.Timer()` uses the default alarm pool. A future PR is planned to refactor this to use soft timer, but this requires deeper soft timer changes to support microsecond resolution.
* pico-sdk invokes alarm pool functions internally. Notably `multicore_lockout_start_blocking()` which adds an alarm `at_end_of_time`. This modifies the alarm pool - but with a timer that never expires, so unlikely to trigger the bug. There is also a USB workaround timeout `hw_enumeration_fix_wait_se0()` that creates a short-lived timeout using the alarm pool.

After `machine.Timer()` is refactored then a fix for the second point is to build pico-sdk with `PICO_TIME_DEFAULT_ALARM_POOL_DISABLED=1`. This converts the other timeouts into busy-waits inside the SDK, and will presumably save some code size as alarm pool code will no longer be compiled into MicroPython. This is not possible yet, as `machine.Timer()` needs an alarm pool and pico-sdk doesn't provide a way to statically initialise a new alarm pool (there is `alarm_pool_create()`, but this calls pico-sdk's malloc to allocate memory for the alarm pool... non-functional commit for this is at https://github.com/projectgus/micropython/commit/9c8ddd8dee8a6a07ed325c214317303771b23e62)

Closes #12873 (Rationale: although the alarm pool is still used in the places described, the access patterns that made it easy to hang the rp2 in a WFE state should no longer be triggered unless `machine.Timer()` is used a lot for very short delays.)

## Future work

1. Refactor soft_timer to support 64-bit microsecond timeouts
2. Refactor `machine.Timer()` to use these
[...]

PR #15398, opened 2024-07-03
rp2: Fix power consumption when sleeping with a timeout.

projectgus:
### Summary

Fixes a regression introduced in 3af006e (part of #13329), where WFE never blocked in `mp_wfe_or_timeout()` function and would busy-wait instead. This increases power consumption measurably. Originally found when testing https://github.com/micropython/micropython/pull/15345

Root cause is that `mp_wfe_or_timeout()` calls soft timer functions that (after the regression) call `recursive_mutex_enter()` and `recursive_mutex_exit()`. Mutex exit calls `lock_internal_spin_unlock_with_notify()` and the default pico-sdk implementation of this macro issues a SEV which negates the WFE that follows it, meaning the CPU never suspends.

See https://forums.raspberrypi.com/viewtopic.php?p=2233908 for more details.

The fix in this comment adds a custom "nowait" variant mutex that doesn't do WFE/SEV, and uses this one for PendSV. This will use more power when there's contention for the PendSV mutex as the other core will spin, but this shouldn't happen very often.

Additional commit fixes a bug that was masked by this bug, specifically that core1 didn't have any way to resume from WFE (because the soft timer interrupt only triggers on core0).

### Testing

Tested on PICO board while measuring USB power consumption. Prior to this fix, `time.sleep(1)` uses 3-4mA more power than sitting at an idle REPL. After this fix, the power consumption is approximately the same (actually approx 0.3mA higher while sleeping, I assume due to the extra clock or maybe because of USB traffic).

Also ran the full rp2 test suite multiple times. The thread_sleep2 test successfully found the core1 bug mentioned above.

Also connected a PICO-W to a Wi-Fi network.

This is the third fix I wrote for this issue(!)

* The first version of this PR replaced the soft timer in `mp_wfe_or_timeout()` with a dedicated hardware timer alarm. This works fine for single core, but in dual core there's potential for both cores to enter this function at the same time meaning this approach is no longer simple.
* I also tried making all the recursive mutexes the "nowait" variety, instead of adding a new variant. This works fine, but it means if there is contention over any mutex then the waiting core will spin. Mostly this probably doesn't matter as there won't be a lot of contention, but there are some pathological cases: For example a program where core1 spends most of its time blocked on a lock or similar.

This approach adds a bit more code size, but it means the low power mutex variant is still used everywhere except for PendSV - and there shouldn't be a lot of lock contention over PendSV.

### DIFF
--- a/ports/rp2/mphalport.c
+++ b/ports/rp2/mphalport.c
@@ -228,6 +228,13 @@ static void soft_timer_hardware_callback(unsigned int alarm_num) {
     // The timer alarm ISR needs to call here and trigger PendSV dispatch via
     // a second ISR, as PendSV may be currently suspended by the other CPU.
     pendsv_schedule_dispatch(PENDSV_DISPATCH_SOFT_TIMER, soft_timer_handler);
+
+    // This ISR only runs on core0, but if core1 is running Python code then it
+    // may be blocked in WFE so wake it up as well. Unfortunately this also sets
+    // the event flag on core0, so a subsequent WFE on this core will not suspend
+    if (core1_entry != NULL) {
+        __sev();
+    }
 }
 
 void soft_timer_init(void) {
--- a/ports/rp2/mpthreadport.c
+++ b/ports/rp2/mpthreadport.c
@@ -39,7 +39,7 @@ extern uint8_t __StackTop, __StackBottom;
 void *core_state[2];
 
 // This will be non-NULL while Python code is executing.
-static void *(*core1_entry)(void *) = NULL;
+core_entry_func_t core1_entry = NULL;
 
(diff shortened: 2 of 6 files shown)


projectgus:
Ah... this method doesn't work if both cores are used, there's a race for the shared alarm timeout. :facepalm:

projectgus:
>   rp2:   -64 -0.007% RPI_PICO_W

Not sure why this is, I assume some lucky linker combo? There is definitely new code being added in this PR.

dpgeorge:
I've tested this PR on a Pico and Pico W and the power consumption looks good for `machine.idle()` and `time.sleep_ms(1)`.

PR #15415, opened 2024-07-06
extmod/modmachine.c: Allow up to 4 arguments for machine.freq().

robert-hh:
The STM32 port supports 4 arguments. At all other ports, only the first argument is used and others are ignored.


dpgeorge:
Thank you!  This was a pretty big regression, not supporting more than 1 argument on stm32...

PR #15426, opened 2024-07-08
esp32/network_lan: Check for valid LAN MAC address.

### LINKED ISSUE #15425: Micropython allows uninitialised, invalid LAN MAC addresses on network

esp32 port, LILYGO T-ETH-LITE ESP32S3 (W5500 SPI LAN)

### MicroPython version

MicroPython v1.23.0 on 2024-06-02; Generic ESP32S3 module with Octal-SPIRAM with ESP32S3

### Reproduction

The following code will activate the W5500 interface on the network and initiate a DHCP request with a MAC address of `b'\x00\x00\x00\x00\x00\x00'`. 

```py
import machine, time
from network import LAN, PHY_W5500

spi = machine.SPI(1, sck=10, miso=11, mosi=12)
lan = LAN(0, spi=spi, phy_type=PHY_W5500, phy_addr=1, cs=9, int=13, reset=14)
lan.active(True)
```

### Expected behaviour

`lan.active()` should fail to activate the interface if the MAC address has not been initialised with `lan.config(mac=...)` (eg. raise an exception).

### Observed behaviour

In general, DHCP servers will refuse the request and `lan.isconnected()` never returns `True` (eg. see #15294). 

If a static IP address is assigned with `lan.ifconfig()`, then `lan.isconnected()` will return `True` and the device will have full network access, though the device will have an invalid MAC address (which would also conflict with other similarly configured micropython LAN devices on the local network).

### Additional Information

This could be addressed by:

1. adding appropriate documentation reminding users to initialise the MAC address, OR
2. adding a check in `network_lan.c:lan_active()` that would raise an exception if the MAC address is uninitialised.

I prefer option 2, as this mitigates the risk of naive use of micropython devices impacting a local network.

### Code of Conduct

Yes, I agree

glenn20:
### Summary

`LAN.active()`: Check for a valid MAC address before enabling the LAN interface. 

SPI LAN devices may be initialised with a MAC address of 00:00:00:00:00:00 which will be ignored by DHCP servers and may conflict with other uninitialised devices on the local network.

This PR checks that a valid unicast MAC address has been set (using `LAN.config(mac=...)`) before enabling the LAN interface and will raise OSError("invalid MAC address") if the MAC address is not valid.

Fixes #15425.

### Testing

Tested with LILYGO T-ETH-LITE ESP32S3 board (with initialised and uninitialised MAC addresses). Also tested that devices initialised with a multicast MAC address will raise an OSError("invalid MAC address").

Alternatively, guidance could be provided in the LAN network docs to properly initialise the MAC address before calling `LAN.active(True)`.

### DIFF
--- a/ports/esp32/network_lan.c
+++ b/ports/esp32/network_lan.c
@@ -36,6 +36,7 @@
 
 #include "esp_eth.h"
 #include "esp_eth_mac.h"
+#include "esp_mac.h"
 #include "esp_event.h"
 #include "esp_log.h"
 #include "esp_netif.h"
@@ -90,6 +91,17 @@ static void eth_event_handler(void *arg, esp_event_base_t event_base,
             break;
         default:
             break;
+    }
+}
+
+static void set_mac_address(lan_if_obj_t *self, uint8_t *mac, size_t len) {
+    if (len != 6) {
+        mp_raise_ValueError(MP_ERROR_TEXT("invalid buffer length"));
+    }
+    if (((mac[0] & 0x01) != 0) ||
+        (esp_eth_ioctl(self->eth_handle, ETH_CMD_S_MAC_ADDR, mac) != ESP_OK) ||
+        (esp_netif_set_mac(self->base.netif, mac) != ESP_OK)) {
+        mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("failed setting MAC address"));
     }
 }
 
@@ -302,6 +314,14 @@ static mp_obj_t get_lan(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_ar
         mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("esp_netif_attach failed"));
     }
 
+    // If MAC address is unset, set it to the address reserved for the ESP32 ETH interface
+    uint8_t mac_addr[6];
+    esp_eth_ioctl(self->eth_handle, ETH_CMD_G_MAC_ADDR, mac_addr);
+    if ((mac_addr[0] | mac_addr[1] | mac_addr[2] | mac_addr[3] | mac_addr[4] | mac_addr[5]) == 0) {
+        esp_read_mac(mac_addr, ESP_MAC_ETH);  // Get ESP32 MAC address for ETH iface
+        set_mac_address(self, mac_addr, sizeof(mac_addr));
+    }
+
     eth_status = ETH_INITIALIZED;
 
     return MP_OBJ_FROM_PTR(&lan_obj);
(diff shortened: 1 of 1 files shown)


dpgeorge:
Does the user always have to set the MAC address, or does the MAC usually get set automatically?

glenn20:
It seems that for many (probably not all) SPI LAN devices, they do not have an allocated MAC address and one must be supplied by the user (or the MCU). This will probably be the case for non-esp32 mcus as well.

For ESP32 devices, Espressif [reserves 4 IP addresses](https://docs.espressif.com/projects/esp-idf/en/v4.4/esp32s3/api-reference/system/system.html#mac-address), (wlan1, wlan2, ble and ethernet). I have been using the following code to allocate a MAC address for my devices:

```py
mac = bytearray(machine.unique_id())  # returns BASE_MAC without starting wifi
mac[-1] += 3
lan.config(mac=mac)
```

I believe espressif's intention was to reserve the IP address for use by their internal ethernet interface, but I figure it is a good default for attached ethernet devices.

glenn20:
> It seems that for many (probably not all) SPI LAN devices, they do not have an allocated MAC address...

Because I beleive/suspect some external devices may have their own MAC address, I query the MAC, rather than just testing that `LAN.config(mac=...)` has been called.

projectgus:
> I believe espressif's intention was to reserve the IP address for use by their internal ethernet interface, but I figure it is a good default for attached ethernet devices.

There's a note in the docs (second note under the linked heading) that shows that using it with attached devices is their intention too: "Although ESP32-S3 has no integrated Ethernet MAC, it is still possible to calculate an Ethernet MAC address. However, this MAC address can only be used with an external ethernet interface such as an SPI-Ethernet device.". The note is there for all the SoCs that don't have an integrated MAC.

What about fixing this by reading back the MAC after initialising the driver, and setting it to the Espressif "Ethernet" MAC value if it's all zeroes? That's more user friendly than initialising an all zero MAC or returning an error, and it'll give "reasonable defaults" behaviour in most cases.

glenn20:
@projectgus That seems like a reasonable approach. It is certainly more user friendly.

The only issue I can see is the risk of initialising an external ethernet device with a MAC address also allocated to an internal ethernet peripheral. This code may need to be flagged for additional checks for future esp32 mcus which may also have an internal ethernet. 

Because of my lack of knowledge about that risk, I just opted for raising an exception as an improvement on the vagaries of initialising a device with an uninitialised mac address.

I'll make the change.

glenn20:
I have updated so that instead of raising an exception:

- Get the MAC address that is reserved for the internal ETH interface ([`esp_read_mac()`](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/misc_system_api.html#_CPPv412esp_read_macP7uint8_t14esp_mac_type_t))
- Set the device MAC address to this reserved address.

The code to set the MAC address has been extracted to the function `set_mac_address()` which is called from `lan_active()` and `lan_config()`.

NOTES

- It may be more appropriate to move this check to the end of `get_lan()`, instead of `lan_active()`. I'm happy to do that if it is preferred.
- `lan.config(mac=xxxx)` now has an additional check - it will raise an exception if the MAC address is not a unicast MAC address (ie. broadcast or multicast).

Tested with a LILYGO T-ETH-LITE-ESP32S3 (W5500 SPI) with multiple soft and hard resets (with and without manually setting a MAC address with `lan.config(mac=)`).

dpgeorge:
> * It may be more appropriate to move this check to the end of `get_lan()`,

That would be my preference, to do it within `get_lan()`.  You should be able to do the following (I hope it works!):
```c
lan = network.LAN()
print(lan.config('mac'))
```
That is, the MAC should be valid after creation.

glenn20:
> That would be my preference, to do it within `get_lan()`. You should be able to do the following (I hope it works!):
> 
> ```c
> lan = network.LAN()
> print(lan.config('mac'))
> ```

Done.

dpgeorge:

PR #15429, opened 2024-07-08
ports/unix/modtermios.c: add baudrate options.

LnnrtS:
### Summary

This adds some more baudrate option as they are available in the linux kernel - up to a certain point that seams reasonable in an embedded context. 

### Testing

Compiled on linux 6.9.7 and opened a connection to a USB-to-Serial adapter with one of the higher baudrates (via patched [pycopy-serial/](https://github.com/pfalcon/pycopy))

The number of possible options could probably be reduced to reduce code size since such high speeds might not be very common. Obviously just having a maximum of 115200 seemed to be good enough until now.

### DIFF
--- a/ports/unix/modtermios.c
+++ b/ports/unix/modtermios.c
@@ -141,6 +141,27 @@ static const mp_rom_map_elem_t mp_module_termios_globals_table[] = {
     #ifdef B115200
     C(B115200),
     #endif
+    #ifdef B230400
+    C(B230400),
+    #endif
+    #ifdef B460800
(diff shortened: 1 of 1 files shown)


LnnrtS:
Assuming the failed checked for macos is a false negative..

dpgeorge:
Thanks for the contribution.  This looks like a reasonable addition, because the constant values are not obvious (eg `termios.B115200` is 4098) and so it's hard to select a higher baudrate without having them.

PR #15431, opened 2024-07-08
ports/rp2: Lazy PIO handlers.

Gadgetoid:
### Summary

This change has (or should have, famous last words) no impact on vanilla MicroPython builds, but is intended to avoid RP2's PIO implementation from trampling PIO usage in USER_C_MODULES.

This is consistent with PIOs tracking of used state machines and managed programs, and makes working with PIO in USER_C_MODULES much less of an uphill battle.

Since PIO deinit runs before gc_sweep_all it's impossible to work around this wrinkle otherwise. A module finalizer does not get the opportunity to put the PIOs back into a state which wont crash rp2_pio_deinit.

rp2_pio.c: 
- init: Avoid exclusive handlers being added to all PIOs and add them when needed.
- deinit: Only remove handlers we have set.
- rp2_pio_irq: Add the exlusive handler if needed.
- rp2_state_mahcine_irq: Add the exclusive handler if needed.

If a PIO's interrupt is in use, it will raise:

```python
ValueError: irq claimed by external resource
```

### Testing

I've performed some basic tests running `pio_1hz.py` with both `sm.irq` and `pio.irq` functions. Both IRQs still fire as expected, and I can soft reset.

In my work-in-progress that uses this feature, `rp2_pio.c` no longer trounces our C module code.

I have not yet tested the impact this has on Pico W (none, hopefully).

This probably comes at a small cost in additional code size, but brings PIO handlers in line with the managed PIO instruction memory and management of individual state machines.

### DIFF
--- a/ports/rp2/rp2_pio.c
+++ b/ports/rp2/rp2_pio.c
@@ -104,6 +104,23 @@ static void pio1_irq0(void) {
     pio_irq0(pio1);
 }
 
+// Returns the correct irq0 handler wrapper for a given pio
+static inline irq_handler_t rp2_pio_get_irq_handler(PIO pio) {
+    return pio == pio0 ? pio0_irq0 : pio1_irq0;
+}
+
+// Add the irq0 handler if it's not added, and if no other handler is present
+void rp2_pio_irq_set_exclusive_handler(PIO pio, uint irq) {
+    irq_handler_t current = irq_get_exclusive_handler(irq);
+    // If the IRQ is set and isn't our handler, or a shared handler is set, then raise an error
+    if ((current && current != rp2_pio_get_irq_handler(pio)) || irq_has_shared_handler(irq)) {
+        mp_raise_ValueError("irq claimed by external resource");
+        // If the IRQ is not set, add our handler
+    } else if (!current) {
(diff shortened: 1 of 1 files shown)


ports/rp2/rp2_pio.c
```diff
@@ -872,6 +893,7 @@ static mp_obj_t rp2_state_machine_irq(size_t n_args, const mp_obj_t *pos_args, m
         }
 
         if (self->pio->inte0) {
+            rp2_pio_irq_set_exclusive_handler(self->pio, self->pio == pio0 ? PIO0_IRQ_0 : PIO1_IRQ_0);
```
dpgeorge, line 896: Couldn't the second argument here just be `self->irq`?

dpgeorge:
This looks like a good change.  I tested it on Pico W and the WiFi and BLE still works.

ports/rp2/rp2_pio.c
Gadgetoid: Good spot! Thank you. No idea how I managed to do this two different ways.

Your thorough reviews are much appreciated!

dpgeorge:
Rebased and merged in ad3552ae2cb5be65a8d61dc1c3bb2014d7ef4cae

Issue #11404, opened 2023-05-03
RP2: machine.spi LSB first is unsupported (also SoftSPI)


peterhinch:
This arose [here](https://github.com/peterhinch/micropython-micro-gui/issues/31). The driver sets up a passed SPI instance as follows:
```py
    def __init__(self, spi, pincs, height=240, width=400, vcom=False):
        spi.init(baudrate=2_000_000, firstbit=machine.SPI.LSB)
        self._spi = spi
```
which fails with `NotImplementedError: LSB` (hard SPI) or `ValueError:firstbit must be MSB` (soft SPI).

The constant `LSB` does exist.
```py
>>> dir(SPI)
['__class__', '__name__', 'read', 'readinto', 'write', '__bases__', '__dict__', 'LSB', 'MSB', 'deinit', 'init', 'write_readinto']
```

slabua:
Any update?

robert-hh:
I made a PR once for LSB first with SoftSPI, but closed it since there were too many PRs open. I can re-open that PR, such that people can build their own firmware based on it.

robert-hh:
See PR #15436.

slabua:
@robert-hh thank you!

dpgeorge:
The RP2040 hardware SPI doesn't support LSB mode.  And it's quite difficult to synthesize LSB mode: eg if the driver transforms all the bytes before sending them out that won't be possible if the data to send is a read-only bytes object (and in ROM).

So I think the rp2 port just has to remain without SPI.LSB for hardware SPI.  Instead one would need to use software SPI if LSB mode is needed (and software SPI.LSB will be implemented in #15436).

slabua:
@dpgeorge i see thanks. Currently I am using hardware SPI by preprocessing the data thru a function that converts MSB to LSB. Not optimal but the only way I found.

dpgeorge:
> Currently I am using hardware SPI by preprocessing the data thru a function that converts MSB to LSB. Not optimal but the only way I found.

Once #15436 is merged I suggest you use `machine.SoftSPI` in LSB mode.

slabua:
@dpgeorge thanks, will do!

dpgeorge:
LSB for SoftSPI was implemented in ee1036023ef199636d96e70c6c7ed587ccaab92e

[...]

Issue #5340, opened 2019-11-17
ESP32 software SPI LSB first wanted


emard:
Currently software SPI channel no. -1 doesn't allow LSB first operation.
What only works is MSB first. For performance reasons, I need both
MSB first and  LSB first in software SPI so I'd like here to post a feature request :-)

nevercast:
Am I mistaken? I don't think there is any software SPI on ESP32. Its all hardware SPI. Can you show an example of the code you're using and the output you get when running it?

emard:
hm you should lnow better :)
this is how I initialize what I think is software SPI, it has -1 channel number
and it accepts only firstbit=SPI.MSB and I'd like to set SPI.LSB sometimes:

self.swspi=SPI(-1, baudrate=self.spi_freq, polarity=1, phase=1,
bits=8, firstbit=SPI.MSB, sck=Pin(self.gpio_tck),
mosi=Pin(self.gpio_tdi), miso=Pin(self.gpio_tdo))

If I set firstbit=SPI.LSB it returns error:
ValueError: firstbit must be MSB

nevercast:
Software SPI support was an external module all this time and I'd never noticed. Interesting! I was incorrect. extmod/machine_spi.c should be your soft SPI

emard:
Great, if it can be fixed to accept LSB and MSB I'll be happy :)

dpgeorge:
LSB for SoftSPI was implemented in ee1036023ef199636d96e70c6c7ed587ccaab92e

PR #15436, opened 2024-07-09
extmod/machine_spi.c: Support firstbit=machine.LSB for machine.SoftSPI.

robert-hh:
Easy to add, sometimes needed and sought for.

For the change the definition of MICROPY_PY_MACHINE_SPI_MSB/ -LSB was moved to py/mpconfig.h, making them available to all ports. The identical defines in esp32/mpconfigport.h were deleted.

Tested with the RP2 and ESP32 port.


extmod/machine_spi.c
```diff
@@ -185,9 +179,8 @@ static mp_obj_t mp_machine_soft_spi_make_new(const mp_obj_type_t *type, size_t n
     if (args[ARG_bits].u_int != 8) {
         mp_raise_ValueError(MP_ERROR_TEXT("bits must be 8"));
     }
-    if (args[ARG_firstbit].u_int != MICROPY_PY_MACHINE_SPI_MSB) {
-        mp_raise_ValueError(MP_ERROR_TEXT("firstbit must be MSB"));
-    }
+    self->spi.firstbit = args[ARG_firstbit].u_int == MICROPY_PY_MACHINE_SPI_MSB ?
+        MICROPY_PY_MACHINE_SPI_MSB : MICROPY_PY_MACHINE_SPI_LSB;
```
dpgeorge, line 183: I don't think there's any need to check for valid inputs like this.  Instead doing `self->spi.firstbit = args[ARG_firstbit].u_int;` is fine (that will anyway give equivalent behaviour because `mp_soft_spi_transfer` is validating the value as well).

Alternatively, could replace the firstbit variable here with `bool msb` and do:
```c
self->spi.msb = args[ARG_firstbit].u_int == MICROPY_PY_MACHINE_SPI_MSB;
```
Then in `mp_soft_spi_transfer` it's simply `self->spi.msb ? src[i] : swap_bits(src[i])`.  That may save a few bytes of code size.

extmod/machine_spi.c
```diff
@@ -227,6 +221,10 @@ static void mp_machine_soft_spi_init(mp_obj_base_t *self_in, size_t n_args, cons
     if (args[ARG_phase].u_int != -1) {
         self->spi.phase = args[ARG_phase].u_int;
     }
+    if (args[ARG_firstbit].u_int != -1) {
+        self->spi.firstbit = args[ARG_firstbit].u_int == MICROPY_PY_MACHINE_SPI_MSB ?
+            MICROPY_PY_MACHINE_SPI_MSB : MICROPY_PY_MACHINE_SPI_LSB;
```
dpgeorge, line 226: As above, no need to validate the value here.

py/mpconfig.h
```diff
@@ -1779,6 +1779,12 @@ typedef double mp_float_t;
 #define MICROPY_PY_MACHINE_SOFTSPI (0)
 #endif
 
+// if a port didn't define MSB/LSB constants then provide them
```
dpgeorge, line 1782: I suggest "Values of SPI.MSB and SPI.LSB constants" (no need to say anything about a port providing them or not, because this whole file is about defining defaults when a port doesn't specify config options).

dpgeorge:
Thanks for this.  I always wondered how LSB could be implemented in `SoftSPI` without too much overhead and this is a neat way of doing it.  Good to support LSB for the cases that need it (and if the hardware doesn't support it / not implemented).

Links to relevant issues: #5340, #11404.

py/mpconfig.h
robert-hh: Changed.

extmod/machine_spi.c
robert-hh: Done.

extmod/machine_spi.c
robert-hh: Changed. The first variant is kept, such that the name of the symbol matches the name of the option. Since the value of MICROPY_PY_MACHINE_SPI_MSB is 0 and `self->spi.msb ? .....,` is actually self->spi.msb != 0 ? .....` most likely the compiler will generate the same code for both variants in mp_soft_spi_transfer.

robert-hh:
The review comments are integrated and the value of firstbit is shown by the object print method.

extmod/machine_spi.c
dpgeorge: > The first variant is kept, such that the name of the symbol matches the name of the option.

Yes, good idea.

And I checked code size and indeed it doesn't change when using `bool msb`.  So the compiler is optimising it as we would expect it to.

PR #8838, opened 2022-06-30
py/vm.c: Only raise pending exceptions from the main thread.

jimmo:
Originally we had a global `MP_STATE_VM(mp_pending_exception)`, which
meant that any thread could handle the pending exception.

In issue #7026 it was decided that only the main thread should handle
pending exceptions (e.g. keyboard interrupt), and PR #7051
(259d9b69f & ca920f721, FYI @dlech) changed this such that the pending exception
became per-thread, and keyboard interrupts were delivered only to the main
thread.

However, the `MP_STATE_VM(sched_state)` was still global, and this meant
that a non-main thread could race to clear `MP_STATE_VM
(sched_state)`, which meant that the main thread would never notice its
`MP_STATE_VM(mp_pending_exception)`.

This could be fixed by having mp_sched_unlock also check if the main thread
still has a pending exception, and leaving `MP_STATE_VM(sched_state)`
pending. (It currently just checks whether the current thread has a pending exception).

However, there's no mechanism for pending exceptions to be delivered to
anything other than the main thread, so instead, this commit makes
`MP_STATE_VM(mp_pending_exception)` global again, but allows only the main
thread to process it. On non-threading builds this has no effect. On
threading builds, a port is able to provide its own "am I the main thread"
function, but a fallback is provided.

PR #15448, opened 2024-07-11
py/scheduler,rp2: Avoid scheduler race conditions when GIL disabled, fix "TinyUSB callback can't recurse" error

### LINKED ISSUE #15390: Raspberry Pi Pico  _thread OSError: TinyUSB callback can't recurse

Raspberry Pi Pico

### MicroPython version

### Reproduction

>>>import time                                                              
>>> import _thread                                                           
>>> def test(name):
...     print(1)                                                             
...     time.sleep(1)
...                                                                          
...                                                                          
... 
>>> _thread.start_new_thread(test,('name',))
2                                                                               
1                                                                               
>>> FATAL: uncaught exception 200127f0                                          
OSError: TinyUSB callback can't recurse 

### Expected behaviour

I found that this problem does not exist in version 1.22.2 

### Observed behaviour

My previous program didn't work after I updated the firmware. I was sure the program would work, so I changed the firmware to 1.22.2 and it worked.I found when the "while True"  in "_thread.start_new_thread" , this problem must occur. I guess it is caused by USB-related changes.

### Additional Information

### Code of Conduct

Yes, I agree

projectgus:
### Summary

Looks like there was a bug since v1.22 when using rp2 threads, where either CPU may poll CDC input and trigger the TinyUSB task. TinyUSB could run on both CPUs concurrently, which may have lead to some incorrect behaviour.

The race started triggering an exception when runtime USB support was added, and a check was added for the USB task recursing on itself from a Python handler function. This check can be incorrectly triggered by the race (as the flag that indicates the TinyUSB task is running is set by the other CPU).

The race is most commonly triggered when working from the interactive REPL, even a minimal running thread can trigger it. This commit adds a test case that triggers it in a different way (polling stdin from a thread).

The root cause is that the scheduler on threaded ports without GIL can run on any thread, and creates potential for race conditions that are hard to avoid. So the first fix is to run all scheduler callbacks on the main thread if GIL disabled.

The secondary fix is not to run the TinyUSB task when polled from another task, if the GIL is disabled. Instead we schedule the TinyUSB task to run on the main thread.

Closes #15390.

Could make the scheduler always only run on the main thread, but with the GIL this has quite an impact on scheduler latency - and isn't necessary for correct behaviour. There is a branch linked from a comment below that mostly works around this, but it's a more complex change (and will still have higher latency than running the callback immediately in whatever thread is currently holding the GIL).

### Testing

* Ran all thread tests, including the new test case, on rp2 port with and without `MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE` set, verified no longer raises an exception.
* Ran all thread tests on esp32 port, verified no regression when GIL enabled.

### DIFF
--- a/py/scheduler.c
+++ b/py/scheduler.c
@@ -236,7 +236,13 @@ void mp_handle_pending(bool raise_exc) {
 
     // Handle any pending callbacks.
     #if MICROPY_ENABLE_SCHEDULER
-    if (MP_STATE_VM(sched_state) == MP_SCHED_PENDING) {
+    bool run_scheduler = (MP_STATE_VM(sched_state) == MP_SCHED_PENDING);
+    #if MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL
+    // Avoid races by running the scheduler on the main thread, only.
+    // (Not needed if GIL enabled, as GIL ensures thread safety here.)
+    run_scheduler = run_scheduler && mp_thread_is_main_thread();
+    #endif
+    if (run_scheduler) {
         mp_sched_run_pending();
     }
     #endif
--- a/shared/tinyusb/mp_usbd_runtime.c
+++ b/shared/tinyusb/mp_usbd_runtime.c
@@ -501,6 +501,15 @@ void mp_usbd_task_callback(mp_sched_node_t *node) {
 // Task function can be called manually to force processing of USB events
 // (mostly from USB-CDC serial port when blocking.)
 void mp_usbd_task(void) {
+    #if MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL
+    if (!mp_thread_is_main_thread()) {
+        // Avoid race with the scheduler callback by scheduling TinyUSB to run
+        // on the main thread.
+        mp_usbd_schedule_task();
+        return;
+    }
+    #endif
+
     if (in_usbd_task) {
         // If this exception triggers, it means a USB callback tried to do
         // something that itself became blocked on TinyUSB (most likely: read or
--- a/tests/thread/thread_stdin.py
+++ b/tests/thread/thread_stdin.py
@@ -1,0 +1,44 @@
+# Test that having multiple threads block on stdin doesn't cause any issues.
+#
+# The test doesn't expect any input on stdin.
+#
+# This is a regression test for https://github.com/micropython/micropython/issues/15230
+# on rp2, but doubles as a general property to test across all ports.
+import sys
+import _thread
+
+try:
+    import select
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+
+class StdinWaiter:
+    def __init__(self):
+        self._done = False
+
+    def wait_stdin(self, timeout_ms):
+        poller = select.poll()
+        poller.register(sys.stdin, select.POLLIN)
+        poller.poll(timeout_ms)
+        # Ignoring the poll result as we don't expect any input
+        self._done = True
+
+    def is_done(self):
+        return self._done
+
+
+thread_waiter = StdinWaiter()
+_thread.start_new_thread(thread_waiter.wait_stdin, (1000,))
+StdinWaiter().wait_stdin(1000)
+
+# Spinning here is mostly not necessary but there is some inconsistency waking
+# the two threads, especially on CPython CI runners where the thread may not
+# have run yet. The actual delay is <20ms but spinning here instead of
+# sleep(0.1) means the test can run on MP builds without float support.
+while not thread_waiter.is_done():
+    pass
+
+# The background thread should have completed its wait by now.
+print(thread_waiter.is_done())
--- a/docs/library/micropython.rst
+++ b/docs/library/micropython.rst
@@ -136,6 +136,14 @@ Functions
    the heap may be locked) and scheduling a function to call later will lift
    those restrictions.
 
+   On multi-threaded ports, the scheduled function's behaviour depends on
+   whether the Global Interpreter Lock (GIL) is enabled for the specific port:
+
+   - If GIL is enabled, the function can preempt any thread and run in its
+     context.
+   - If GIL is disabled, the function will only preempt the main thread and run
+     in its context.
+
    Note: If `schedule()` is called from a preempting IRQ, when memory
    allocation is not allowed and the callback to be passed to `schedule()` is
    a bound method, passing this directly will fail. This is because creating a


ports/rp2/mphalport.h
```diff
@@ -68,6 +68,11 @@
 extern int mp_interrupt_char;
 extern ringbuf_t stdin_ringbuf;
 
+#if MICROPY_PY_THREAD
+// Only run the TinyUSB task on CPU 0 to avoid data races
+#define MICROPY_HW_USBD_TASK_CAN_RUN() (get_core_num() == 0)
```
projectgus, line 73: @dpgeorge I wasn't sure what the best place for this would be. We could add a mp_usbd_port.h header, but it seems like overkill while there's only a single macro.

dpgeorge: I think it's fine here.

dpgeorge:
> Looks like there was a long standing underlying bug here when using rp2 threads, where either CPU may poll CDC input and trigger the TinyUSB task. TinyUSB could run on both CPUs concurrently, which may have lead to some incorrect behaviour.

I guess that was introduced in bcbdee235719d459a4cd60d51021454fba54cd0f ?  Prior to that commit, tinyusb was guarded so it ran only on core 0.

projectgus:
> > 
> 
> I guess that was introduced in [bcbdee2](https://github.com/micropython/micropython/commit/bcbdee235719d459a4cd60d51021454fba54cd0f) ? Prior to that commit, tinyusb was guarded so it ran only on core 0.

I think that commit set up the conditions for the bug. TinyUSB still only runs on CPU0 as of this commit because that's where the scheduler runs it. However, that commit introduces a different bug - waiting for data in places where the scheduler doesn't run will hang. Commit a00c9d56db775ee5fc14c2db60eb07bab8e872dd fixed that by re-introducing a TinyUSB task hook, but without the CPU check.

projectgus:
> TinyUSB still only runs on CPU0 as of this commit because that's where the scheduler runs it.

Oh wait, I have this wrong don't I?  mp_handle_pending() can run on both CPUs too? I didn't realise this...

dpgeorge:
> mp_handle_pending() can run on either CPU

Correct. Eg called from the VM itself.




> The bug still doesn't happen in that commit because the scheduler has a lock, but I think that this fix is incomplete as it doesn't handle the case where mp_usb_task is scheduled to run, the scheduler runs on CPU1, and simultaneously is polled from code running on CPU0...)

Are you saying this PR is not correct ("this fix is incomplete"), or are you referring to the commit a00c9d56db775ee5fc14c2db60eb07bab8e872dd not being fully correct?

projectgus:
> Are you saying this PR is not correct ("this fix is incomplete")

This PR isn't a complete fix. I think there's still a race where:

* CPU0 is calling poll_cdc_interfaces() which runs mp_usbd_task (allowed by the CPU0 check)
* CPU1 runs the scheduler and this runs mp_usbd_task

The simple fix is to always guard the scheduled mp_usb_task to not run on CPU1, but because mp_usbd_task is run on-demand then there needs to be a way to re-schedule mp_usbd_task to run later, and keep retrying until it runs on CPU0. I have a fix for this here now, but I'm not sure about it (also requires a scheduler change).

Or, add proper locking around mp_usbd_task so it can run on either CPU.

projectgus:
I've pushed two commits which are a possible fix for this, but I don't like it much.

I think the smart fix here might be to revert bcbdee235 and re-introduce the dedicated TinyUSB task hook instead of calling TinyUSB via the scheduler. If the performance hit of that change is high then can keep the DCD event handler function and use it to set a flag that we check before calling into TinyUSB.

dpgeorge:
We need to think about this in depth, because all other bare-metal ports that use TinyUSB need to do the same thing.

projectgus:
> We need to think about this in depth, because all other bare-metal ports that use TinyUSB need to do the same thing.

Agree about thinking it through. Currently we have rp2 and samd using the scheduler approach, and all of the other ports using a task hook.

The queue-via-scheduler approach should work fine on all the single core bare metal ports. It's only made complex in the dual core case where it's possible for both CPUs to be trying to call into the TinyUSB task at once.

Will think about it some more. Maybe adding another recursive mutex on rp2 is the way to go.

dpgeorge:
@projectgus and @jimmo and I discussed this and decided to fix it by only allowing scheduled functions to run on the main thread.  See related #8838 and https://github.com/pybricks/micropython/commit/b196c783c081c5444749995df789d1620fa90814

projectgus:
>  decided to fix it by only allowing scheduled functions to run on the main thread

I've implemented this "scheduler only on the main thread" change and it's working but I don't think it's a good solution after all. See this branch:
https://github.com/projectgus/micropython/commits/bugfix/rp2_tinyusb_recursive_mainthread/

It's a problem for ports with GIL and multiple threads (i.e. ESP32). The `stress_schedule.py` test really falls over, and I think real workloads will also suffer:

* vm.c:1358 only bounces the GIL if the scheduler is idle, not pending, which is a problem if only main thread can run the scheduler and another thread is running. Changing this to also bounce the GIL if pending on a non-main thread helps but it's still quite slow because the GIL only bounces every `MICROPY_PY_THREAD_GIL_VM_DIVISOR` branches. Therefore scheduled function latency can be quite high if another thread is running.
* Replaced that with a change in the linked branch, where scheduler.c bounces the GIL instead of running the scheduler on a non-main threads. This improves things a lot, but it's still higher latency because now the main thread has to take the GIL, then execute another iteration of the main VM loop, and only then it handles the scheduled operation. Whereas before the non-main thread would have run the scheduler immediately.

Going to take another look at this next week. @jimmo mentioned the idea of having a flag so some scheduler nodes only run on the main thread, this might be the best compromise. Otherwise maybe implementing a mutex for TinyUSB (with some caution around ensuring race conditions can't lose TinyUSB events despite the mutex).

dpgeorge:
> I've implemented this "scheduler only on the main thread" change and it's working but I don't think it's a good solution after all.

Hmmm, OK, I'll also have a think about this.  We do need to try and reduce the latency of scheduled callbacks, because they are being used more and more for background workloads.

projectgus:
Finally got a version of this fix I'm happy with, after @dpgeorge pointed out that this problem only exists where GIL is disabled.

(An aside, I also noticed that there is no user-facing documentation for the GIL at all. I made a note to submit something.)

projectgus:
@dlech This may be of interest, although I'm unsure if its equivalent to https://github.com/pybricks/micropython/commit/b196c783c081c5444749995df789d1620fa90814 or not (specifically, I don't know if pybricks enables the GIL or not).

dlech:
Yeah, we are using that patch on the Unix port with the GIL enabled. So it looks like this isn't quite the same. The problems were were trying to solve are A) Unix signal handlers should always run on the main thread, like in CPython and B) if there is an unhandled exception in any background task, we want it to stop the whole program by throwing in the main thread rather than throwing in a random thread.

projectgus:
@dlech That makes sense, thanks for the extra detail.

dpgeorge:
OK, so PyBrick's use case is more about functionality, rather than fixing a bug.  And because it's on the unix port there's less concern about loss of performance / increase latency of scheduled code.

------

I have tested this latest PR on rp2 and it fixes the original report in #15390.

py/scheduler.c
```diff
@@ -236,7 +236,13 @@ void mp_handle_pending(bool raise_exc) {
 
     // Handle any pending callbacks.
     #if MICROPY_ENABLE_SCHEDULER
-    if (MP_STATE_VM(sched_state) == MP_SCHED_PENDING) {
+    bool run_scheduler = (MP_STATE_VM(sched_state) == MP_SCHED_PENDING);
+    #if !MICROPY_PY_THREAD_GIL
```
dpgeorge, line 240: Is it worth making this `#if MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL`, to be clear that it's only needed when threading is enabled?

I know it makes no difference, due to `mp_thread_is_main_thread()` being `false` when thread is disabled.  So could go either way here.

shared/tinyusb/mp_usbd_runtime.c
```diff
@@ -501,6 +501,15 @@ void mp_usbd_task_callback(mp_sched_node_t *node) {
 // Task function can be called manually to force processing of USB events
 // (mostly from USB-CDC serial port when blocking.)
 void mp_usbd_task(void) {
+    #if !MICROPY_PY_THREAD_GIL
```
dpgeorge, line 504: Similar comment to above.

tests/thread/thread_stdin.py
```diff
@@ -0,0 +1,42 @@
+# test that having multiple threads block on stdin doesn't cause any issues.
```
dpgeorge, line 1: Upper case Test, since the sentence ends in a full stop.

tests/thread/thread_stdin.py
```diff
@@ -0,0 +1,42 @@
...
+#
+# The test doesn't expect any input on stdin.
+#
+# This is a regression test for https://github.com/micropython/micropython/issues/15230
+# on rp2, but doubles as a general property to test across all ports.
+import select
```
dpgeorge, line 7: Zephyr has `_thread` but not `select`, so this test fails on zephyr boards with threading enabled.

Probably best to try-except on this import, and skip accordingly (or add an explicit zephyr skip in `run-tests.py`).

tests/thread/thread_stdin.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+thread_waiter = StdinWaiter()
+_thread.start_new_thread(thread_waiter.wait_stdin, (1000,))
+StdinWaiter().wait_stdin(1000)
+
+# add a short delay to account for inconsistency waking the two threads (on
+# CPython CI runners there seems to be some intermittent difference in thread
+# wakeup delays.) On Linux we expect both threads should resume more or less
+# concurrently - aka the "thundering herd" effect"
```
dpgeorge, line 35: Grammar nitpicks:
- upper case at start of paragraph
- full-stop outside the closing )
- remove additional " after herd
- full-stop at end

tests/thread/thread_stdin.py
```diff
@@ -0,0 +1,42 @@
...
+try:
+    time.sleep_ms(100)
+except AttributeError:
+    time.sleep(0.1)
```
dpgeorge, line 39: How about just doing `time.sleep(0.1)` without any try-except?  All ports should support that.

projectgus: Oh, I somehow had it in my head for many years that there was a MicroPython configuration that didn't support floats.

Easy!

tests/thread/thread_stdin.py
projectgus: Happy to make this edit, but do want to point out the existing tests are *very* inconsistent in this regard. IIRC, the reason it's written like this is that I copied one of the other test files and tried to match its style.

py/scheduler.c
projectgus: I had the same thought, but erred on the other side - given we both thought it then I've made the change!

tests/thread/thread_stdin.py
dpgeorge: I agree there's a lot of inconsistency with comments.  But this jumped out at me, and I don't think any other tests have a lowercase start and a full stop (at least none of the thread tests).

Would be good to clean this up, but not really worth the time.

tests/thread/thread_stdin.py
dpgeorge: You are right, there are configs without float.  CC3200 is the main example that has threading but no float, and these thread tests were originally written to run on that target.

It's probably worth retaining that ability, ie don't use floats here.

But note that if floats are enabled on a build, then even parsing `0.1` will raise a `SyntaxError` before the code runs.  So your original code won't support non-float builds.

How about just a busy loop waiting for the done flag to be set?  (Also don't want a dependency on `machine` here.)  The test running will time out after 30 seconds if the flag is never set.

projectgus: Have changed as suggested.

dpgeorge (approved):

dpgeorge:
This turned out to be a relatively small patch, which is great.  Thank you!

PR #15450, opened 2024-07-11
extmod/modopenamp: Add support for building Open-AMP on device side.

iabdalkader:
Adds support for building Open-AMP in device mode, in addition to a couple of misc fixes. This was tested with two VMs each running on a different core. To build for device:

```basn
make -j<n> BOARD=my_board CORE=M4 MICROPY_PY_OPENAMP_DEVICE=1
```
Of course, whichever port you use, it must support building for a second core. For example, the main core will initialize USB, but the second one will not and so on.


extmod/modopenamp.c
```diff
@@ -411,6 +422,6 @@ const mp_obj_module_t openamp_module = {
 };
 
 MP_REGISTER_ROOT_POINTER(struct _virtio_dev_obj_t *virtio_device);
-MP_REGISTER_MODULE(MP_QSTR_openamp, openamp_module);
+MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR_openamp, openamp_module);
```
dpgeorge, line 425: Why do you want to make this module extensible?  In general, this feature is only used to support CPython modules that we don't fully support in C, so they can be extended in Python.

For the `openamp` module we define it ourselves, so there's nothing missing that needs adding.

iabdalkader: I extend in Python to add utils and classes that can be accessed with `openamp.x`. For example I have an `EndPointIO` class which is a subclass of `IOBase` that you can use for `dupterm`.

iabdalkader:
I pushed one more commit for `stm32` to define the last used MPU region number. This macro allows custom/outside code to find a free MPU region to use. For example you can start at the max (H7 that's 15) loop down to the `MPU_REGION_LAST` or start from `MPU_REGION_LAST+1` knowing that it will not clash with regions used by MicroPython.

extmod/modopenamp.c
iabdalkader: Hm I guess my `EndpointIO` class would not be needed if https://github.com/micropython/micropython/pull/14181 is merged but I still have other extensions.

dpgeorge: > Note that there's an openamp.Endpoint in the C-module, so openamp.EndpointIO belongs there too but as an extension.

I don't think it's a good idea to add things to this module outside this repo.  We should fully define the `openamp` module here, and have the docs here.  Any extensions that work with `openamp` but are outside this repo (which would be third-party extensions, relative to this repo) should be placed in a separate module, eg `openamp_ext`, with separate docs.  Otherwise it will get confusing what functions/classes are available in this module.

Eventually we want to get rid of this extensible module stuff (it's a backwards compatible way to support the old u-module naming scheme), so I don't want to add more modules that use this.

iabdalkader: > Eventually we want to get rid of this extensible module stuff (it's a backwards compatible way to support the old u-module naming scheme), so I don't want to add more modules that use this.

Whoa that will be a very breaking change. We heavily depend on extending modules for adding missing functionality or for portability. For example, we extend machine to add an LED that's used throughout or examples. We'll also extend it to mask missing machine.PWM for `stm32` port. Can this feature be kept for such uses cases ?

iabdalkader: Did I understand correctly that we'll no longer be able to extend C-Modules from Python ? Note that we've switched to using the extensible macro a while ago, so we have accumulated a bit of an ecosystem around it, extensions, example scripts and docs. I'm mentioning this because I really think decisions to remove things like this should consider the side-effects to users of the C-API. It's been hard to keep up with the C-API breaking changes, but it's all for the greater good. However, I don't understand why a feature like this should be removed. We maintain a fork anyway, we can adapt, but I've been working really hard to align with the upstream, in the hope that one day we can ditch the fork.

dpgeorge: > Eventually we want to get rid of this extensible module stuff (it's a backwards compatible way to support the old u-module naming scheme), so I don't want to add more modules that use this.

Sorry, I didn't really express myself clearly in this comment.

What I intended to say is that we want to eventually drop the u-module naming backwards compatibility feature, which is tied in to the extensible module macro.  But you're right that we can't get rid of the ability to extend C modules from Python.

At the same time, I think it's important to limit the number of extensible modules because  extensible ones are slower to import than non-extensible ones, due to needing to search the filesystem first.  So their use should be limited and only when necessary.

You can see that the extensible modules are strictly defined and documented here: https://docs.micropython.org/en/latest/library/index.html#extending-built-in-libraries-from-python .  We really do want to limit their use.

See also the comment in `extmod/modbluetooth.c`:
```c
// This module should not be extensible (as it is not a CPython standard
// library nor is it necessary to override from the filesystem), however it
// has previously been known as `ubluetooth`, so by making it extensible the
// `ubluetooth` alias will continue to work.
MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR_bluetooth, mp_module_bluetooth);
```

For the case here of `openamp`, I really think it should not be extensible.  If there's important additional C functionality then it should be added in this repo, with docs, so it's available on all ports.  Or if that functionality is best implemented in Python, and it makes more sense to have it in `openamp` rather than a separate module, then it should go in `micropython-lib` and then, yes, the module would be extensible.

IMO it's best to limit the scope of the low-level `openamp` module to just the necessary C functionality, and then have a different higher-level module written in Python that uses `openamp` to provide more functionality.  For example that's done with `_thread` and `threading`, `bluetooth` and `aioble`, `machine.USBDevice` and `usb-device`, to name a few.

dpgeorge: > Note that we've switched to using the extensible macro a while ago, so we have accumulated a bit of an ecosystem around it, extensions, example scripts and docs

That's interesting... this is not really what the feature was intended for.  As I mention above, I think modules should have well defined scope and built-in C modules left to provide basic functionality, while other Python modules build upon the low-level modules and provide more features.

projectgus: > I think this is all subjective

This is a tricky topic for sure, but I suggest the measurable aspect of what Damien is talking about is consistency. The extension mechanism is used to ensure consistency between MicroPython and CPython (i.e. extending os with `os.path`), or between different MP ports (i.e. extending `machine.Pin` to the unix port). In both cases there are docs for those modules already, and the extensions mechanism keeps the implementation (and therefore user experience) as consistent with those docs as possible.

> However, the rest of the extensions, which BTW are Not yet ready for sharing, may or may not be welcome by the maintainers. 

Either way, I think the preferred path for MicroPython and user experience would be to have a separate module here:

* If the additional Python-based functionality gets merged to micropython-lib, then this sounds like it fits well with the common pattern Damien already mentioned of bluetooth/aioble, machine.USBDevice/usb-device, etc.
* If the additional Python functionality is not merged to micropython-lib, then using the same module name would cause the MicroPython docs for the `openamp` module to be inconsistent with the docs for the external extended `openamp` module. Using a different name removes the ambiguity and improves the user experience.

iabdalkader: > If the additional Python functionality is not merged to micropython-lib, then using the same module name would cause the MicroPython docs for the openamp module to be inconsistent with the docs

But what if we don't want to push our extensions to `micropython-lib` even if they are in line with the upstream policy and will be merged ? Maybe we'd want to maintain them privately, or push them somewhere else, like `arduino-lib` and then we could also maintain our own docs for the module.
The reason for sending work like this upstream, in addition to contributing something useful, is to Not have fork and maintain it ourselves, but if we still have to fork and change the code that we ourselves put upstream just to be able to extend it, then what's the point of pushing it upstream in the first place.

iabdalkader: Note that you can drop the extensible commit (it's in its own commit) if you don't want it, if this is ready to be merged.

projectgus: > then we could also maintain our own docs for the module.

It's of course your choice if you want to do that. I'd note that if there are two modules called `openamp` with different functions in them, and different docs for that module in two places, then this can easily lead to a bad user experience. i.e. Finding the docs for `openamp` (even `openamp` docs for the matching MP version) won't necessarily find the correct docs, because there will be two different implementations with the same name.

Sometimes that's unavoidable because of external constraints (and the history of MP has plenty of confusing incidents arising from those unavoidable name clashes). In this case it seems like it is avoidable, by giving different modules different names. Maybe I don't understand the use case here, though.

> Note that you can drop the extensible commit (it's in its own commit) if you don't want it, if this is ready to be merged.

:thumbsup:

ports/stm32/mpmetalport.h
```diff
@@ -70,7 +70,7 @@ static inline int __metal_sleep_usec(unsigned int usec) {
 }
 
 static inline void metal_generic_default_poll(void) {
-    MICROPY_EVENT_POLL_HOOK
+    mp_event_handle_nowait();
```
dpgeorge, line 73: This changes the semantics of the poll: previously it would do a WFI, now it doesn't.

Where exactly is this function used?  I can see some code in the libmetal submodule (function called `metal_condition_wait`) that call this poll function with IRQs disabled.  Is that ever actually used?

Note that if IRQs are disabled then a WFE may never return (but a WFI will).

extmod/modopenamp.c
```diff
@@ -181,13 +192,13 @@ static mp_obj_t endpoint_send(uint n_args, const mp_obj_t *pos_args, mp_map_t *k
...
         if (bytes > 0 || timeout == 0) {
-            MICROPY_EVENT_POLL_HOOK
+            metal_generic_default_poll();
             break;
         }
         if (timeout > 0 && (mp_hal_ticks_ms() - start > timeout)) {
             mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("timeout waiting for a free buffer"));
         }
-        MICROPY_EVENT_POLL_HOOK
+        metal_generic_default_poll();
```
dpgeorge, line 201: This loop, waiting to send, will now be a very busy loop and use lots of CPU power.  It would be better to use `mp_event_wait_ms(1)` here.

And then use `mp_event_handle_nowait()` above just before the `break`.  Then `metal_generic_default_poll()` could continue to call `mp_event_handle_nowait()` (if that `mp_event_handle_nowait` function is ever even used).

(As a side remark, IMO it's better to use the `mp_event_*` functions here in the extmod code because it's well defined what they do, and it makes reasoning about this code easier.  Whereas if `metal_generic_default_poll` is used the behaviour is dependent on the port.)

extmod/extmod.mk
```diff
@@ -536,12 +536,19 @@ ifeq ($(MICROPY_PY_OPENAMP_REMOTEPROC),1)
 CFLAGS += -DMICROPY_PY_OPENAMP_REMOTEPROC=1
 endif
 
+ifeq ($(MICROPY_PY_OPENAMP_DEVICE),1)
+CFLAGS += -DMICROPY_PY_OPENAMP_DEVICE=1
+CFLAGS_THIRDPARTY += -DVIRTIO_DEVICE_ONLY
+else
```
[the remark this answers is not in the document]
iabdalkader: Any of those options will do, I'd prefer:
```
ifeq ($(MICROPY_PY_OPENAMP_MODE),0)
// host
else ifeq ($(MICROPY_PY_OPENAMP_MODE),1)
// device
else
error "OpenAMP mode build mode is not specified."
endif
```

ports/stm32/mpmetalport.h
iabdalkader: Yes you have a point, but there's no `mp_event` that uses WFI, did I miss it ? This function is called from libmetal, why/where specifically I'm not really sure, but we need to provide an implementation.

extmod/extmod.mk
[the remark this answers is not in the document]
iabdalkader: Actually this was intended to be passed on the command line to rebuild the port in device mode, so you won't have to edit files to rebuild a port in device mode. So there should be a default mode.

ports/stm32/mpmetalport.h
iabdalkader: The equivalent would be:
```
mp_event_handle_nowait();
WFI();
```

iabdalkader: Fixed.

ports/stm32/mpu.h
```diff
@@ -36,12 +36,19 @@
...
-#define MPU_REGION_OPENAMP  (MPU_REGION_NUMBER15)
 
 // Only relevant on CPUs with D-Cache, must be higher priority than SDRAM
 #define MPU_REGION_DMA_UNCACHED_1 (MPU_REGION_NUMBER6)
 #define MPU_REGION_DMA_UNCACHED_2 (MPU_REGION_NUMBER7)
 
+// Last used MPU region.
+#ifdef MPU_REGION_NUMBER8
+#define MPU_REGION_OPENAMP  (MPU_REGION_NUMBER8)
+#define MPU_REGION_LAST     (MPU_REGION_NUMBER8)
```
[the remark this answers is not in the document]
iabdalkader: Done.

dpgeorge:
Thanks for updating.  I've now merged this in 1216f2c313fbf68cb7f18c4914ddf843c31d9191 through 91f4a6b9e902c066c98bda3799e01c8c6c0783ea, all commits here in this PR **except** the commit that makes the `openamp` module extensible.  That needs further discussion.

PR #15467, opened 2024-07-15
ports/zephyr: Add threading support on the nrf52840DK.

danicampora:
This  PR attempts to implement the _thread module on the Zephyr port. Three important remarks about this PR:

1. Due to the fact that we are still using a rather old version of Zephyr, CONFIG_DYNAMIC_THREAD is not available and therefore the stack for threads cannot be allocated dynamically, only at compile time. So for the time being and for the purpose of these PR, a maximum of 4 Zephyr threads (besides the main thread) can be created. Once we manage to update to the latest version of Zephyr this won't be a problem anymore.

2. For some reason, threads get blocked forever before calling `mp_thread_start();` inside `thread_entry()` (see modthread.c, line 174), because `MP_THREAD_GIL_ENTER()` never returns. I'm unable to understand why this is happening, since this implementation of `mpthreadport` for the Zephyr port is following the same approach used by the ESP32 port. I suspect there's some kind of incompatibility with the rest of the Micropython options enabled for the Zephyr port, and I hope that @dpgeorge can shed some light in here. For this reason `MICROPY_PY_THREAD_GIL` is disabled at the moment. This obviously needs to be fixed before this PR is merged.

3. Some of the tests in the `tests/thread` folder run successfully, and some other don't most likely because `MICROPY_PY_THREAD_GIL` is not enabled.

A simple test that runs OK:

```
import _thread
import time

def thread_entry(num):
    for i in range(3):
        time.sleep(1 + (num * 0.5))
        print("Thread {} running...".format(num))

_thread.start_new_thread(thread_entry, (0,))
_thread.start_new_thread(thread_entry, (1,))
_thread.start_new_thread(thread_entry, (2,))
_thread.start_new_thread(thread_entry, (3,))
```


dpgeorge:
> For some reason, threads get blocked forever before calling `mp_thread_start();`

This is because the zephyr port does not release the GIL at the REPL.

With the GIL enabled you can see that things do work if you run some code at the REPL.  Eg a loop like `for _ in range(100): pass` will let the other threads obtain the GIL.

Ports like esp32 have a GIL release (via `MICROPY_EVENT_POLL_HOOK`) in their `mp_hal_stdin_rx_chr()` functions, which polls for chars at the REPL.  Zephyr doesn't have this because it polls for input using the zephyr subsystem.

This limitation of zephyr also has other unwanted consequences, eg scheduled events (like BLE irq callbacks) don't run at the REPL.

In the simplest case (to fix threading) the zephyr port needs release the GIL in its `mp_hal_stdin_rx_chr()` function.  Eventually it should be fixed properly by also allowing scheduled events to run.

ports/zephyr/mpconfigport_full.h
```diff
@@ -0,0 +1,164 @@
...
+// Included here to get basic Zephyr environment (macros, etc.)
+#include <zephyr/zephyr.h>
+#include <zephyr/drivers/spi.h>
+
+// Usually passed from Makefile
+#ifndef MICROPY_HEAP_SIZE
+#define MICROPY_HEAP_SIZE (16 * 1024)
+#endif
+
+#define MICROPY_ENABLE_SOURCE_LINE  (1)
```
dpgeorge, line 39: There's a lot of config in this file repeated in `mpconfigport.h`.  Ideally there'd be a common base file with common config options, included by other config files.

ports/zephyr/mpconfigport_full.h
```diff
@@ -0,0 +1,164 @@
...
+    { MP_ROM_QSTR(MP_QSTR_open), MP_ROM_PTR(&mp_builtin_open_obj) },
+
+#if MICROPY_PY_THREAD
+#define MICROPY_EVENT_POLL_HOOK \
+    do { \
+        extern void mp_handle_pending(bool); \
+        extern void mp_hal_main_sem_take(); \
+        mp_handle_pending(true); \
+        MP_THREAD_GIL_EXIT(); \
+        mp_hal_main_sem_take(); \
```
dpgeorge, line 154: I'm not sure what `mp_hal_main_sem_take()` is used for... it seems that this semaphore is never given so I'm not sure what it's supposed to do?

dpgeorge:
This is great, thank you!

I was able to test it with the `qemu_cortex_m3` target.  But I had to modify some config files so that that target would build correctly with threading enabled.  We need to make it easier to enable threading on any given board, eg by a single config option.

ports/zephyr/mpconfigport_full.h
danicampora: @dpgeorge thanks for pointing this out, I will clean it up.

ports/zephyr/mpconfigport_full.h
danicampora: I saw this on the ESP32 port and figured it would be needed here sooner rather than later, but forgot to remove it when submitting this PR. I'll clean this up and add it again when (if) required.

danicampora:
> This is great, thank you!
> 
> I was able to test it with the `qemu_cortex_m3` target. But I had to modify some config files so that that target would build correctly with threading enabled. We need to make it easier to enable threading on any given board, eg by a single config option.

Good point. I'll look into it. Thanks!

danicampora:
> This limitation of zephyr also has other unwanted consequences, eg scheduled events (like BLE irq callbacks) don't run at the REPL.

This explains some of the mixed results I had when testing BLE on the Zephyr port. Now a lot of stuff makes sense :-)

danicampora:
> In the simplest case (to fix threading) the zephyr port needs release the GIL in its mp_hal_stdin_rx_chr() function

Sounds good! I will work on this.

> Eventually it should be fixed properly by also allowing scheduled events to run.

I would very much like to do this as well, I'll add it to my list. Thanks for all the details @dpgeorge, this helps me a lot in understanding what's the current state of the Zephyr port and how certain things have been done in there.

dpgeorge:
> > Eventually it should be fixed properly by also allowing scheduled events to run.
> 
> I would very much like to do this as well, I'll add it to my list. Thanks for all the details @dpgeorge, this helps me a lot in understanding what's the current state of the Zephyr port and how certain things have been done in there.

Ideally we'd make an event sub-system for zephyr that made efficient use of semaphores to signal when events are ready to process.  I think zephyr can do this with it's polling system which can poll things for read/write and also wait on a semaphore.

Then we'd use this, eg at the REPL to wait for either char input or an event (eg BLE) to process.

But, that's quite a bit of work!

danicampora:
> Ideally we'd make an event sub-system for zephyr that made efficient use of semaphores to signal when events are ready to process. I think zephyr can do this with it's polling system which can poll things for read/write and also wait on a semaphore.

That makes a lot of sense. But as a first approach to get threading working I was looking into changing the way the Zephyr REPL reads characters from the console. Instead of using `console_getchar`, I plan to use `console_read` which does return if no characters are available. What do you think of that?

dpgeorge:
> I plan to use `console_read` which does return if no characters are available

That sounds reasonable.

danicampora:
@dpgeorge can you take another look at this PR? I added several more commits and this now runs like a well-oiled embedded machine. 
Happy to squash all the commits into one before merging, but I think like this is easier to review what changed since the last time you checked.

danicampora:
> This limitation of zephyr also has other unwanted consequences, eg scheduled events (like BLE irq callbacks) don't run at the REPL.

I can confirm that with the changes in this PR BLE callbacks are able to run fine at the REPL.

dpgeorge:
I have tested this on a NUCLEO_WB55RG board, with threading enabled using the `-DOVERLAY_CONFIG=thread.conf` option to west.

The BLE events work a lot better now, they can be received at an idle REPL.  That's great!

Running the thread tests... sometimes they pass and other times they do not.  For example:
```
$ ./run-tests.py --target pyboard thread/thread_ident1.py thread/thread_coop.py thread/thread_ident1.py                                                                     
pass  thread/thread_ident1.py
pass  thread/thread_coop.py
pass  thread/thread_ident1.py
3 tests performed (8 individual testcases)
3 tests passed
```
That's good, but then if I access the REPL after running those tests and do a soft reset, it hangs.

@danicampora it would be great if we could get at least a small subset of the thread tests running successfully over and over again.  For those that fail, we can eventually try to fix them.

ports/zephyr/README.md
```diff
@@ -63,6 +63,11 @@ To build for QEMU instead:
 
     $ west build -b qemu_x86 ~/micropython/ports/zephyr
 
+To build any board with the `_thread` module enabled, 
+add `-DOVERLAY_CONFIG=thread.conf`, for instance:
+
+$ west build -b frdm_k64f ~/micropython/ports/zephyr -DOVERLAY_CONFIG=thread.conf
```
dpgeorge, line 69: Please indent this line with 4 spaces.

Also, might be worth replacing `~/micropython/ports/zephyr` with `--` so it works regardless of where the MP repo is (but then this example would be inconsistent with other examples in this file... probably worth changing them all).

danicampora:
@dpgeorge sorry for the delay in responding. I solved the issue with the soft reset, and now working on the failed tests. Will come back to this asap.

danicampora:
@dpgeorge I solved the problem with the soft reset and with several other tests. At the moment there are 3 tests failing:

- `thread_gc1.py` this one passes if I run it manually on the REPL, but when done via the run_tests script, it hangs.
- `thread_lock3.py` fails because it tries to spawn 10 threads, and Zephyr supports a maximum of 4 at the moment.
- `thread_lock4.py` the output is different from the expected one, also not sure why this is happening.

dpgeorge:
I have pushed two commits to try and improve things:
- collect registers during a GC scan (this is definitely needed, but may not help with tests on this PR)
- enable mpz big ints, this should get `thread_lock4.py` working (which computes large factorials)

But still I see failures.  Actually I see more than what's reported above when I run this PR on a board configured with less heap (eg 16k or even 32k heap): `thread_gc1.py`, `thread_lock4.py` (after enabling mpz) and `thread_qstr1.py` all fail.  And I think it's all because of the same underlying issue with the GC scanning other threads.

I investigated it a bit using `qemu_cortex_m3` as the target (after disabling networking and enabling threading) but couldn't work out the issue.  We definitely need to fix it because if `thread_gc1.py` fails then there's something wrong with the GC and that will lead to unreliable firmware.

ports/zephyr/main.c
```diff
@@ -174,17 +180,23 @@ int real_main(void) {
     machine_pin_deinit();
     #endif
 
+    #if MICROPY_PY_THREAD
+    mp_thread_deinit();
+    gc_sweep_all();
+    gc_collect();
```
dpgeorge, line 186: Why exactly do you need to call `gc_sweep_all()` and `gc_collect()` here?  You shouldn't really call `gc_collect()` after `gc_sweep_all()`, that's not really well defined I don't think...

ports/zephyr/mpthreadport.c
```diff
@@ -0,0 +1,310 @@
...
+
+    // TODO: we need to support for CONFIG_DYNAMIC_THREAD in order to dynamically create allocate the stack of a thread
+    // if (*stack_size == 0) {
+    //     *stack_size = MP_THREAD_DEFAULT_STACK_SIZE; // default stack size
+    // } else if (*stack_size < MP_THREAD_MIN_STACK_SIZE) {
+    //     *stack_size = MP_THREAD_MIN_STACK_SIZE; // minimum stack size
+    // }
+
+    // in case some threads have finished but their stack has not been collected yet
+    mp_thread_gc_others();
```
dpgeorge, line 196: I don't think this works, having this here.  You can't call `gc_collect_root()` without first calling `gc_collect_start()` and then finishing up with `gc_collect_end()`.

What exactly is this line trying to do?  If it's needed, maybe a call to `gc_collect()` is what you want?

ports/zephyr/main.c
danicampora: Somehow I thought that was required. I'll look into this and fix it.

ports/zephyr/mpthreadport.c
danicampora: That line tries to free the allocated stack for threads that have already finished, so that we can give it to a new thread. 

> You can't call gc_collect_root() without first calling gc_collect_start() and then finishing up with gc_collect_end()

Perhaps this is the problem with `thread_gc1.py`. I'll find a better way to do this.

danicampora:
@dpgeorge thanks for the patience with this one. I believe I solved the problem when collecting other threads. This is now what I get when running ALL the _thread tests on zephyr:

```
danielcampora@Daniels-MacBook-Pro-3 tests % python3 run-tests.py --target pyboard --device /dev/tty.usbmodem0010502934581  thread/*.py
skip  thread/disable_irq.py
pass  thread/mutate_bytearray.py
pass  thread/mutate_dict.py
pass  thread/mutate_instance.py
pass  thread/mutate_list.py
skip  thread/mutate_set.py
pass  thread/stress_aes.py
pass  thread/stress_create.py
FAIL  thread/stress_heap.py
pass  thread/stress_recurse.py
pass  thread/stress_schedule.py
pass  thread/thread_exc1.py
pass  thread/thread_exc2.py
pass  thread/thread_exit1.py
pass  thread/thread_exit2.py
pass  thread/thread_gc1.py
pass  thread/thread_heap_lock.py
pass  thread/thread_ident1.py
pass  thread/thread_lock1.py
pass  thread/thread_lock2.py
FAIL  thread/thread_lock3.py
pass  thread/thread_lock4.py
pass  thread/thread_lock5.py
pass  thread/thread_qstr1.py
pass  thread/thread_shared1.py
pass  thread/thread_shared2.py
pass  thread/thread_sleep1.py
pass  thread/thread_sleep2.py
pass  thread/thread_stacksize1.py
pass  thread/thread_start1.py
pass  thread/thread_start2.py
29 tests performed (144 individual testcases)
27 tests passed
2 tests skipped: disable_irq mutate_set
2 tests failed: stress_heap thread_lock3
```
The reason for the 2 failing tests is because they want to run 10 threads, and a max of 4 are supported. This limitation can be fixed once we upgrade to the latest Zephyr and thus the stack for the threads can be allocated from the GC heap.

dpgeorge:
Thanks @danicampora, that's looking great now.  I will test and review.

ports/zephyr/mpconfigport.h
```diff
@@ -43,11 +43,16 @@
...
 #define MICROPY_KBD_EXCEPTION       (1)
-#define MICROPY_CPYTHON_COMPAT      (0)
 #define MICROPY_PY_ASYNC_AWAIT      (0)
+#ifdef CONFIG_THREAD_CUSTOM_DATA            // this one enables the _thread module
+#define MICROPY_CPYTHON_COMPAT      (1)
+#define MICROPY_PY_ATTRTUPLE        (1)
+#else
+#define MICROPY_CPYTHON_COMPAT      (0)
 #define MICROPY_PY_ATTRTUPLE        (0)
+#endif
```
dpgeorge, line 53: Let's just enable `MICROPY_CPYTHON_COMPAT` and `MICROPY_PY_ATTRTUPLE` (and `MICROPY_PY_BUILTINS_ENUMERATE` below) unconditionally, ie regardless of the thread setting.

And because all these are enabled by default in `py/mpconfig.h`, you can just remove their entries from this file.

(Eventually we'll move to using a `MICROPY_CONFIG_ROM_LEVEL` setting for this port, and this moves us in that direction.)

ports/zephyr/mpconfigport.h
```diff
@@ -113,6 +118,12 @@
 #define MICROPY_COMP_CONST (0)
 #define MICROPY_COMP_DOUBLE_TUPLE_ASSIGN (0)
 
+#ifdef CONFIG_THREAD_CUSTOM_DATA
```
dpgeorge, line 121: It's a bit unconventional that a Zephyr option controls a MicroPython option.  But I can't really see a better way to do it.

Maybe add a comment here along the lines of "When the Zephyr CONFIG_THREAD_CUSTOM_DATA feature is enabled, the MicroPython MICROPY_PY_THREAD feature is enabled automatically"

ports/zephyr/main.c
dpgeorge: This bit still needs investigation.

danicampora: Done. I need to call `gc_collect()` after `mp_thread_deinit()` otherwise soft-reset starts giving issues. But now `gc_sweep_all()` is called after (not before) and added a call to `mp_deinit()` to be in-line with the rest of the ports.

ports/zephyr/mpconfigport.h
danicampora: OK. Done. Also enabled MICROPY_PY_BUILTINS_SET.

dpgeorge:
I tested this again on `nucleo_wb55rg` and also `frdm_k64f`, and it works.

I updated `run-tests.py` so it skips the two failing thread tests, that require more that 4 threads.

I also increased the default GC heap to 48k so that thread tests can pass on a default build.

dpgeorge:
@danicampora I squashed and cleaned up the commit history and force pushed to your branch here.  So you'll need to `git pull --rebase` if you check it out again.

I was about to merge it, but then noticed that when threading is disabled the REPL is unreliable, if you type too fast then it locks up.  At least, I noticed this on a `nucleo_wb55rg`.  That will need investigating.

danicampora:
@dpgeorge thank so much!! I will investigate the issue with the unreliable REPL asap.

danicampora:
@dpgeorge I can't reproduce the REPL problem when threading is disabled (on a `nrf52840dk`). Could this be specific to the `nucleo_wb55rg`? I don't have one with me at the moment, but I can get one next week.

Is the REPL lock-up issue also happening to you on the `frdm_k64f`?

dpgeorge:
> can't reproduce the REPL problem when threading is disabled (on a `nrf52840dk`). Could this be specific to the `nucleo_wb55rg`?

It looks like it's related to `CONFIG_CONSOLE_SUBSYS`.  The situation seems to be:
- `nucleo_wb55rg` has `CONFIG_CONSOLE_SUBSYS=n`
- on master, this board works fine
- with this PR, this board locks up quite easily at the REPL after entering ~100 characters (just hold down any key, eg "a", to see it lock up)
- with this PR and `CONFIG_CONSOLE_SUBSYS=y`, this board works fine, the REPL is good and all tests can be run
- BUT, with `CONFIG_CONSOLE_SUBSYS=y` ctrl-C no longer works, it can't break out of a loop (for example)
- setting `CONFIG_CONSOLE_SUBSYS=n` on the `frdm_k64f` and with this PR, it locks up immediately and you can't type anything at the REPL

I'm not familiar enough with the latest Zephyr to know the best way to get a serial port working.  Is it best to have `CONFIG_CONSOLE_SUBSYS` enabled or disabled?  If it's enabled, how can we intercept ctrl-C?

This is a bit out of scope for this PR, but we do need to fix the issue with the `nucleo_wb55rg`.

dpgeorge:
I'm not sure why, but if I revert just the changes to `zephyr_getchar()` made in this PR, then `nucleo_wb55rg` works fine again at the REPL.  So the issue is probably something to do with:
```c
    mp_hal_wait_sem(&uart_sem, 0);
    if (k_sem_take(&uart_sem, K_MSEC(0)) == 0) {
```
It looks like the call to `mp_hal_wait_sem()` does almost nothing when the timeout is 0, it never gets to the `k_poll` call.  That's probably OK, but also probably not the intention.

ports/zephyr/mpconfigport.h
```diff
@@ -139,3 +142,20 @@ typedef long mp_off_t;
...
 #define MICROPY_PORT_BUILTINS \
     { MP_ROM_QSTR(MP_QSTR_open), MP_ROM_PTR(&mp_builtin_open_obj) },
+
+#if MICROPY_PY_THREAD
+#define MICROPY_EVENT_POLL_HOOK \
+    do { \
+        extern void mp_handle_pending(bool); \
+        mp_handle_pending(true); \
+        MP_THREAD_GIL_EXIT(); \
+        MP_THREAD_GIL_ENTER(); \
```
dpgeorge, line 152: Maybe this should do a `k_msleep(1)` between GIL_EXIT and GIL_ENTER, otherwise things like `mp_hal_stdin_rx_chr()` become busy loops with no chance to power-save the MCU.

danicampora:
@dpgeorge at the moment the only board that I have with me is the `nrf52840dk`. On the master branch, with `CONFIG_CONSOLE_SUBSYS` disabled, the REPL on this board locks immediately when trying to receive the first character right after boot. It goes to the hardfault handler after calling `k_yield()` in line 49 of `zephyr_getchar.c` (**console_irq_input_hook**). If I remove the `k_yield()` there, then the REPL works fine.

1. Can you check if removing this `k_yield()` resolves the issues on the `nucleo_wb55rg`?
2. Besides that, why can't we just have `CONFIG_CONSOLE_SUBSYS=y` for all Zephyr boards? Is there a reason some boards can't have that option enabled?

danicampora:
@dpgeorge I pushed 2 new commits. Specifically with this one: https://github.com/micropython/micropython/pull/15467/commits/4fe7a1452d7be446f3928b64c655344048df253b

I can run Zephyr Micropython on the `nrf52840dk` with or without `CONFIG_CONSOLE_SUBSYS` enabled and also with and without `_thread` enabled, and the REPL is always stable. I can also run the entire `_thread` test suit successfully with `CONFIG_CONSOLE_SUBSYS` disabled on the  `nrf52840dk`.

danicampora:
> I'm not sure why, but if I revert just the changes to `zephyr_getchar()` made in this PR, then `nucleo_wb55rg` works fine again at the REPL. So the issue is probably something to do with:
> 
> ```c
>     mp_hal_wait_sem(&uart_sem, 0);
>     if (k_sem_take(&uart_sem, K_MSEC(0)) == 0) {
> ```
> 
> It looks like the call to `mp_hal_wait_sem()` does almost nothing when the timeout is 0, it never gets to the `k_poll` call. That's probably OK, but also probably not the intention.

I pushed https://github.com/micropython/micropython/pull/15467/commits/a0767837ad3885ca776dec092ec3cefbfab6dd7a to address that.

dpgeorge:
> Can you check if removing this `k_yield()` resolves the issues on the `nucleo_wb55rg`?

Yes, this fixes it for me, great!

> Besides that, why can't we just have CONFIG_CONSOLE_SUBSYS=y for all Zephyr boards?

`CONFIG_CONSOLE_SUBSYS=y` doesn't allow ctrl-C to break out of running code, which is a pretty handy feature.  We eventually need to get ctrl-C working for SUBSYS enabled (or switch everything to SUBSYS disabled).

> I can run Zephyr Micropython on the `nrf52840dk` with or without `CONFIG_CONSOLE_SUBSYS` enabled and also with and without `_thread` enabled, and the REPL is always stable. I can also run the entire `_thread` test suit successfully with `CONFIG_CONSOLE_SUBSYS` disabled on the `nrf52840dk`.

Very good!

I also tested the following:
- `nucleo_wb55rg` as-is
- `nucleo_wb55rg` with `CONFIG_CONSOLE_SUBSYS=y`
- `nucleo_wb55rg` with `_thread` enabled (SUBSYS disabled)
- `frdm_k64f` as-is (`_thread` disabled, SYBSYS enabled)

All of these cases work fine, the REPL is stable, test suite can be run (although about 10 tests fail but that's unrelated to this PR), and `_thread` tests pass on `nucleo_wb55rg` when threading is enabled.

And the fix to `mp_hal_wait_sem()` looks good as well, thank you!

So, this PR looks ready now.

dpgeorge (approved):

dpgeorge:
Merged!

Thanks a lot @danicampora for working on this, it's a great improvement!

PR #15476, opened 2024-07-17
esp32: Fix uneven GIL allocation between Python threads.

### LINKED ISSUE #15423: Inconsistent thread behavior in different versions of MicroPython on ESP32

ESP-WROOM-32 (chip ESP32-D0WD-V3)

### MicroPython version

- [Works well] MicroPython v1.20.0 on 2023-04-26; ESP32 module with ESP32
- [Bug] MicroPython v1.22.2 on 2024-02-22; Generic ESP32 module with ESP32
- [Bug] MicroPython v1.23.0 on 2024-06-02; Generic ESP32 module with ESP32

### Reproduction

```
import _thread
import time
        
def thread2task():
    print("thread 2 started")
    while True:
        pass


if __name__ == "__main__":
    _thread.start_new_thread(thread2task,())
        
    while 1:
        t = time.ticks_ms()
        time.sleep(1)
        print(time.ticks_ms() - t)
```

### Expected behaviour

The main loop should consistently produce similar timing intervals regardless of the thread's state. For example, the sleep interval should be close to 1000ms as in version 1.20.

Output:
- MicroPython v1.20.0 on 2023-04-26; ESP32 module with ESP32:
```
MPY: soft reboot
thread 2 started
1006
1010
1010
1010
1010
1010
1010
1010
1010
1010
[...]
```

### Observed behaviour

Creating a simple thread in versions above 1.20 causes variable time intervals, which affects the timing accuracy of the main loop, suspends the connection to the computer, and often prevents the script from being restarted.

Output:
- MicroPython v1.22.2 on 2024-02-22; Generic ESP32 module with ESP32:
```
MPY: soft reboot
thread 2 started
[Nothing else appears, this is where the script gets stuck]
```
- MicroPython v1.23.0 on 2024-06-02; Generic ESP32 module with ESP32:
**Once I managed to get this result, other times nothing is printed, as in version 1.22.2**
```
MPY: soft reboot
thread 2 started
1843
1230
1830
2060
1500
1500
2020
1980
2020
1980
[...]
```

### Additional Information

### Code of Conduct

Yes, I agree

projectgus:
### Summary

Explicitly yield each time a thread mutex is unlocked. Closes #15423.

Key to understanding this bug is that Python threads run at equal RTOS priority, and although ESP-IDF FreeRTOS (and I think vanilla FreeRTOS) scheduler will round-robin equal priority tasks in the ready state it does not make a similar guarantee for tasks moving between ready and waiting.

The pathological case of this bug is when one Python thread task is busy (i.e. never blocks) it will hog the CPU more than expected, sometimes for an unbounded amount of time. This happens even though it periodically unlocks the GIL to allow another task to run.

Assume T1 is busy and T2 is blocked waiting for the GIL. T1 is executing and hits a condition to yield execution:

1. T1 calls MP_THREAD_GIL_EXIT
2. FreeRTOS sees T2 is waiting for the GIL and moves it to the Ready list (but does not preempt, as T2 is same priority, so T1 keeps running).
3. T1 immediately calls MP_THREAD_GIL_ENTER and re-takes the GIL.
4. Pre-emptive context switch happens, T2 wakes up, sees GIL is not available, and goes on the waiting list for the GIL again.

To break this cycle step 4 must happen before step 3, but this may be a very narrow window of time so it may not happen regularly - and quantisation of the timing of the tick interrupt to trigger a context switch may mean it never happens.

Yielding at the end of step 2 maximises the chance for another task to run.

### Testing

Adds a test case which is based on the code in the linked bug report. This test fails consistently on esp32 without this fix, and passes afterwards. Test case also passes on rp2, which is the only other port where thread tests are currently enabled, and passes if run manually on the unix port.

This PR also includes a commit to enable the thread tests on the esp32 port.

Could make a port-specific `MP_THREAD_GIL_EXIT` macro that yields, to avoid the overhead of yielding any time a thread mutex is unlocked. However almost all of the thread mutex lock/unlock events in the esp32 port are the GIL (the other thread mutexes are the GC mutex and the mutex that protects thread creation and cleanup).


dpgeorge:
> Could make a port-specific `MP_THREAD_GIL_EXIT` macro that yields, to avoid the overhead of yielding any time a thread mutex is unlocked. However almost all of the thread mutex lock/unlock events in the esp32 port are the GIL (the other thread mutexes are the GC mutex and the mutex that protects thread creation and cleanup).

When the GIL is enabled the GC mutex is not used (neither the qstr mutex).  The GIL is "global" so is enough to make everything exclusive across thread access.

But mutex's are also used for `_thread.allocate_lock()` (regardless of the GIL being enabled/disabled).  And I think it's also good that they will now yield when releasing, that may improve thread cooperation on esp32 when the user code uses locks at the Python level.

dpgeorge:
Also good that you enabled thread tests on esp32.  I think if they were already enabled a long time ago, we probably would have caught this issue when updating the IDF to v5.

projectgus:
> I think if they were already enabled a long time ago, we probably would have caught this issue when updating the IDF to v5.

Probably, at least one of the existing tests hangs intermittently without this fix.

dpgeorge:
I tested this and it works well, fixes the original issue.

Thank you!

Issue #15471, opened 2024-07-16
rp2: Board does not enable USB after hard reset if an error is raised in boot.py.


robert-hh:
### Port, board and/or hardware

rp2, mimxrt, SAMD, but maybe other's as well

### MicroPython version

MicroPython v1.24.0-preview.23.g8809ae716.dirty on 2024-07-16; Raspberry Pi Pico with RP2040

### Reproduction

Just create a boot.py file, which raises an exception and push reset or do a power cycle. On Soft reset this problem does not occur. I used the following test script with the option to bypass the test:
```
from machine import Pin
p=Pin(2, Pin.IN, Pin.PULL_UP)
if p():
    1/0
```

### Expected behaviour

Just boot with an exception message and bypass main.py.

### Observed behaviour

Board does not register USB and seems bricked. Uploading an older firmware including v1.23.0 helps.
Just a few interim commits: 
Fails with commit 8809ae716 shared/tinyusb: Buffer startup CDC data to send to host on connection.
Runs with the previous commit d144f0699 lib/tinyusb: Update to the most recent master.

### Additional Information

No, I've provided everything above.
@pi-anl 

### Code of Conduct

Yes, I agree

robert-hh:
Going with a debugger through the code, it hangs when trying to write an error message to the FIFO, but the TX FIFO is not overwritable. File tusb_fifo.c, line 484.  In that line, the flag `f->overwritable` is `false` at that time. And thus it locks up. Overwritable will be set to `true` later in the call to `mp_tusb_init()`, which is called **after** boot.py is executed.

robert-hh:
b.t.w.:  It does not have to be an exception in boot.py. Any output (e.g. print) by boot.py locks up the board.

dpgeorge:
Fixed by a734ee9057a760b0316eca110d71db1524142fec

PR #15481, opened 2024-07-17
mp_usbd_cdc.c: Skip writing to an uninitialized USB device.

robert-hh:
During execution of boot.py the USB device is not yet initialized. Any attempt to write would lock up the device. This change skips writing. Any output from boot.py is lost, but the device does not lock up.

Attempt to fix issue #15471.

### DIFF
--- a/shared/tinyusb/mp_usbd.h
+++ b/shared/tinyusb/mp_usbd.h
@@ -125,7 +125,6 @@ inline static bool mp_usb_device_builtin_enabled(const mp_obj_usb_device_t *usbd
 
 static inline void mp_usbd_init(void) {
     // Without runtime USB support, this can be a thin wrapper wrapper around tusb_init()
-    extern bool tusb_init(void);
     tusb_init();
 }
 
--- a/shared/tinyusb/mp_usbd_cdc.c
+++ b/shared/tinyusb/mp_usbd_cdc.c
@@ -95,6 +95,9 @@ void tud_cdc_rx_cb(uint8_t itf) {
 }
 
 mp_uint_t mp_usbd_cdc_tx_strn(const char *str, mp_uint_t len) {
+    if (!tusb_inited()) {
+        return 0;
+    }
     size_t i = 0;
     while (i < len) {
         uint32_t n = len - i;


projectgus:
This looks like probably a regression from #14485 when MicroPython started buffering output to send to CDC on connect. Combined with not initialising TinyUSB until after boot.py finishes as a way to support clean runtime USB initialisation.

I think any early print output was probably lost anyhow, as the TinyUSB CDC driver zeroes its data structures on driver init. So I think this is a good way to fix it, thanks @robert-hh .

cc @andrewleech

shared/tinyusb/mp_usbd.h
```diff
@@ -67,6 +67,8 @@ extern const uint8_t mp_usbd_builtin_desc_cfg[MP_USBD_BUILTIN_DESC_CFG_LEN];
 
 void mp_usbd_task_callback(mp_sched_node_t *node);
 
+bool tusb_inited(void);
```
dpgeorge, line 70: This shouldn't be needed, it's already defined in `tusb.h`.

robert-hh: OK. For the same reason, the declaration of `tusb_init()` further down is also not needed.

andrewleech:
Ah yes, previously the stdout would skip USB if it's not connected, but now it tries to write to the USB buffer. I hadn't considered situations before USB init occurred; I hadn't realised that was possible! 

Yep it's certainly reasonable to add this check to skip if it's not init, that's much safer.

dpgeorge:
I tested this by adding a print to `boot.py`: before this PR the Pico W would not start up, and with this PR it does start up (but the print is lost).  So that's good.

The alternative would be to somehow pre-initialise TinyUSB's CDC buffers very early on (before `boot.py`) and then not touch them when doing a proper `tusb_init()`, to retain any data printed.  But that would most likely require changes to TinyUSB (which we could eventually do...).

robert-hh:
Thanks. I started looking into TinyUSB for the change, but then I preferred the simple patch in the MicroPython Domain. Changing TinyUSB would require to update at it's source repository. I'm sure Hatach would accept it, but it would take more time.

PR #12802, opened 2023-10-25
unix/main: Use standard pyexec/repl for unix port.

andrewleech:
This improves repl usage consistency across ports. 
Only enabled when MICROPY_USE_READLINE == 1 (default).

In particular I noticed that `aiorepl` didn't work properly on the unix port currently, it would echo the "just entered line" on a newline each time you hit enter, eg.
```
>>> import aiorepl, asyncio;asyncio.run(aiorepl.task())
Starting asyncio REPL...
--> print("hello")
print("hello")
hello
```
I feel there were also a few other inconsistencies with how I'm used to the repl working on the stm port / via mpremote which are alsos hopefully resolved now.

I've got an optional couple of lines to restore/reset stdio_raw mode when running code - however if raw mode is disabled the aiorepl issue above remains. I think for consistency with other ports it's better to keep stdio in raw mode, but not sure if this is likely to have other side effects?

This PR consilidates the repl functionality and fixes the above aiorepl glitch.

PR #15486, opened 2024-07-18
Make raising `SystemExit` do a soft reset on bare-metal targets

dpgeorge:
### Summary

The current situation with `SystemExit` and soft reset is the following:
- `sys.exit()` follows CPython and just raises `SystemExit`
- on the unix port, raising `SystemExit` quits the application / MicroPython, whether at the REPL or in code (this follows CPython behaviour)
- on bare-metal ports, raising `SystemExit` at the REPL does nothing, raising it in code will stop the code and drop into the REPL
- `machine.soft_reset()` raises `SystemExit` but with a special flag set, and bare-metal targets check this flag when it propagates to the top-level and do a soft reset when they receive it

The original idea here was that a bare-metal target can't "quit" like the unix port can, and so dropping to the REPL was considered the same as "quit".  But this bare-metal behaviour is arguably inconsistent with unix.

This PR proposes to change the behaviour to the following, which is more consistent:
- raising `SystemExit` on a bare-metal port will do a soft reset (unless the exception is caught by the application)
- `machine.soft_reset()` is now equivalent to `sys.exit()`
- unix port behaviour remains unchanged

### Testing

Tested running the test suite on PYBD-SF6 and everything still passes (in particular tests that skip by raising `SystemExit` still correctly skip).

This decreases code size, simplifies behaviour and makes unix and bare-metal behaviour match (and match CPython).

It is a breaking change though, if users are expecting `SystemExit` to just drop to the REPL.  The alternative would be to not make this change.

### DIFF
--- a/extmod/modmachine.c
+++ b/extmod/modmachine.c
@@ -62,7 +62,6 @@ NORETURN static void mp_machine_deepsleep(size_t n_args, const mp_obj_t *args);
 #endif
 
 static mp_obj_t machine_soft_reset(void) {
-    pyexec_system_exit = PYEXEC_FORCED_EXIT;
     mp_raise_type(&mp_type_SystemExit);
 }
 static MP_DEFINE_CONST_FUN_OBJ_0(machine_soft_reset_obj, machine_soft_reset);
--- a/extmod/modmachine.c
+++ b/extmod/modmachine.c
@@ -24,6 +24,7 @@
  * THE SOFTWARE.
  */
 
+#include "py/builtin.h"
 #include "py/runtime.h"
 
 #if MICROPY_PY_MACHINE
@@ -33,6 +34,10 @@
 
 #if MICROPY_PY_MACHINE_DHT_READINTO
 #include "drivers/dht/dht.h"
+#endif
+
+#if !MICROPY_PY_SYS_EXIT
+#error MICROPY_PY_MACHINE requires MICROPY_PY_SYS_EXIT
 #endif
 
 // The port must provide implementations of these low-level machine functions.
@@ -60,11 +65,6 @@ NORETURN static void mp_machine_deepsleep(size_t n_args, const mp_obj_t *args);
 #ifdef MICROPY_PY_MACHINE_INCLUDEFILE
 #include MICROPY_PY_MACHINE_INCLUDEFILE
 #endif
-
-static mp_obj_t machine_soft_reset(void) {
-    mp_raise_type(&mp_type_SystemExit);
-}
-static MP_DEFINE_CONST_FUN_OBJ_0(machine_soft_reset_obj, machine_soft_reset);
 
 #if MICROPY_PY_MACHINE_BOOTLOADER
 NORETURN mp_obj_t machine_bootloader(size_t n_args, const mp_obj_t *args) {
@@ -156,7 +156,7 @@ static const mp_rom_map_elem_t machine_module_globals_table[] = {
     #endif
 
     // Reset related functions.
-    { MP_ROM_QSTR(MP_QSTR_soft_reset), MP_ROM_PTR(&machine_soft_reset_obj) },
+    { MP_ROM_QSTR(MP_QSTR_soft_reset), MP_ROM_PTR(&mp_sys_exit_obj) },
     #if MICROPY_PY_MACHINE_BOOTLOADER
     { MP_ROM_QSTR(MP_QSTR_bootloader), MP_ROM_PTR(&machine_bootloader_obj) },
     #endif
--- a/ports/qemu-arm/modmachine.c
+++ b/ports/qemu-arm/modmachine.c
@@ -27,9 +27,6 @@
 // This file is never compiled standalone, it's included directly from
 // extmod/modmachine.c via MICROPY_PY_MACHINE_INCLUDEFILE.
 
-// This variable is needed for machine.soft_reset(), but the variable is otherwise unused.
-int pyexec_system_exit = 0;
-
 static void mp_machine_idle(void) {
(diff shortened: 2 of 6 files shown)


dpgeorge:
This is related to #12802 which makes a similar change, but I think it's worth putting this in a dedicated PR for visibility.

andrewleech:
Ah, for some reason I had in my head that `pyexec_system_exit` was only being used on unix port, not for the bare metal ones. 
I see now in this isolated PR how it's used in the embedded ports to control the soft reset.

Regardless, this looks good to me and I think it makes sense for `SystemExit` to soft-reboot by default. 

It's certainly true that there could be other users out there expecting to use it as a silent drop to repl, but that seems fairly unlikley to me? 
Either way, said users could still catch it and drop to repl themselves in their `main.py` as a fairly simple workaround.

dpgeorge:
> It's certainly true that there could be other users out there expecting to use it as a silent drop to repl, but that seems fairly unlikley to me?

Yes, it's possible some code is relying on this behaviour.  But it would be an uncommon use, I'm guessing.


> Either way, said users could still catch it and drop to repl themselves in their `main.py` as a fairly simple workaround.

Exactly.  And doing it that way (with an `except SystemExit`) would be more explicit in the code, what the intended behaviour is.

danicampora:
I like this change! since an embedded system is supposed to run indefinitely after being deployed, so doing a soft reset after a SystemExit seems more consistent and logical.

Gadgetoid:
+1 to this. It makes sense and prevents surprises related to DMA, PIO, etc allocation on rp2 at least. I don't believe we have any deployments or devices using SystemExit to mean "exit to repl"  at the moment, but if we do it's just another thing for us to fix.

dpgeorge:
Thanks for the reviews!

andrewleech:
Nice one, I'll rebase my PR onto master

dpgeorge:
Unfortunately this change is not fully compatible with CPython behaviour.

Running CPython with `-i` for inspect mode, and running a script that doe `raise SystemExit` (or `sys.exit()`), the script finishes and then it drops to the REPL.  You need to issue a second `raise SystemExit` to actually exit from the REPL back to the shell:
```
$ cat test.py
print(123)
raise SystemExit
$ python -i test.py
123
  File "test.py", line 2, in <module>
    raise SystemExit
SystemExit
>>>
>>> raise SystemExit
$
```
But with this PR (which is merged), on bare-metal a `raise SystemExit` in a running program, eg main.py, will exit the program and skip the REPL and go straight to a soft reset (similar CPython exiting).

If we consider the REPL on bare-metal to be the equivalent of `-i` inspect mode in CPython, then `raise SystemExit` should drop to the REPL, which is what it used to do, but now it doesn't :disappointed: 

-------

I see two options:
1. Leave things as they currently are, `raise SystemExit` does a soft reset (ie doesn't drop to the REPL).  The justification is that bare-metal ports are already different enough to CPython on a desktop (eg no `sys.args`, raw vs friendly REPL, etc) that this additional difference doesn't matter.
2. Revert the change here so `raise SystemExit` drops to the REPL.  And then `machine.soft_reset()` is used to skip the REPL and go straight to a soft reset.

Note that option (1) means there's no clean way for an application to abort and drop to the REPL.  You would need to create a custom exception derived from `BaseException` and raise that (eg `raise SystemAbort`).  I guess that's acceptable.

projectgus:
> If we consider the REPL on bare-metal to be the equivalent of -i inspect mode in CPython, then raise SystemExit should drop to the REPL, which is what it used to do, but now it doesn't 😞

@dpgeorge I don't quite follow: if this is optional behaviour on CPython (enabled by `-i`) then why would we consider MicroPython's default behaviour to be the equivalent of that rather than the equivalent of CPython's default behaviour?

dpgeorge:
> I don't quite follow: if this is optional behaviour on CPython (enabled by `-i`) then why would we consider MicroPython's default behaviour to be the equivalent of that rather than the equivalent of CPython's default behaviour?

That's a good point, and maybe that is the right way to look at it.

My thinking was that bare-metal has inspect mode that is switched on by default.  Considering you can't specify command-line options on bare-metal, this seems reasonable?  Bare-metal is kind of like CPython's `python -i boot.py main.py` (except you can't run two scripts in succession like that in the same namespace).

projectgus:
> Bare-metal is kind of like CPython's `python -i boot.py main.py` (except you can't run two scripts in succession like that in the same namespace).

Oh, I see the reasoning now. I guess, as you say, MicroPython on bare metal doesn't cleanly map to either "inspect mode" or regular CPython behaviour at the moment - but I can see the resemblance to inspect mode.

> Note that option (1) means there's no clean way for an application to abort and drop to the REPL. You would need to create a custom exception derived from BaseException and raise that (eg raise SystemAbort). I guess that's acceptable.

It's hacky but another option would be to programmatically `raise KeyboardInterrupt`, yes?

For my 2c, having an automatically executed embedded application automatically drop to a REPL on "exit" seems more like a footgun than useful behaviour in most circumstances, but it's possible I'm not thinking through all the ways people might want to use this.

dpgeorge:
> It's hacky but another option would be to programmatically `raise KeyboardInterrupt`, yes?

Yes.  Or just `raise BaseException` (which usually nothing ever catches).



> For my 2c, having an automatically executed embedded application automatically drop to a REPL on "exit" seems more like a footgun than useful behaviour in most circumstances

Yes, I tend to agree with that.  You could then argue that we shouldn't ever drop to the REPL (eg on an uncaught exception), but then things like `mpremote` would never work.  So we need to make some compromise somewhere.

PR #15495, opened 2024-07-19
extmod/vfs_fat: Allow setting label of partition when creating FAT fs

tstenvold:
### Summary

During a factory reset, a label for the flash drive can be set via `MICROPY_HW_FLASH_FS_LABEL`. However, it would be convenient to be able to set this value when using `vfs.VfsFat.mkfs(p1)`. This PR allows an optional label argument to `vfs.VfsFat.mkfs(p1, "label")`

### Testing
Tested on a PYBV11 for creating a new label for the mounted flash with no label and with a label

Slight additional overhead for `mkfs()`. Incompatible function call for littlefsX. 
An alternative, would be to add a `vfs.VfsFat.setlabel("label")`, I decided against this as it would be an additional function call adding more overhead.

### DIFF
--- a/extmod/vfs_fat.c
+++ b/extmod/vfs_fat.c
@@ -100,7 +100,9 @@ static mp_obj_t fat_vfs_del(mp_obj_t self_in) {
 static MP_DEFINE_CONST_FUN_OBJ_1(fat_vfs_del_obj, fat_vfs_del);
 #endif
 
-static mp_obj_t fat_vfs_mkfs(mp_obj_t bdev_in) {
+static mp_obj_t fat_vfs_mkfs(size_t n_args, const mp_obj_t *args) {
+    mp_obj_t bdev_in = args[0];
+
     // create new object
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the contribution.

This feature has actually been requested a few times, and was implemented as a separate `.label()` method in #13470.  There I argued for the C-time macro to be added  to the rp2 port, instead of a new function.

I actually think a separate function is more useful than adding a keyword argument to `mkfs()` (as done here).  Because what if you want to relabel the filesystem without formatting it?  Or maybe that's not a common use case, and that labelling it during formatting is all that's needed?  After all, in #13470 the use-case was to label during formatting.

If we did go with a separate `.label()` method, it could be possible to retrieve the label if no arguments were specified, eg `if vfs.label() == "ABC": ...`.

tstenvold:
This was the trade-off I was also thinking. The microcontroller rarely cares (or should) about the volume label, so it's more a quality of life for when the controller is plugged into a computer, and is likely not changed often.

A separate function to set/get the label is more overhead than I think is required, and getting the label has many edge cases that for when volume labels aren't supported, but yes there isn't a straight forward solution.

Issue #15501, opened 2024-07-19
Call __set_name__ for class members on class creation.


AJMansfield:
### Description

Add support for the `__set_name__` data model method defined in Python 3.6.
[`https://docs.python.org/3/reference/datamodel.html#object.__set_name__`](https://docs.python.org/3/reference/datamodel.html#object.__set_name__)

This functionality is important for using descriptors effectively. Often, it's used to construct associated attribute names for storing per-instance per-descriptor state:

```python
class Descriptor:
    def __set_name__(self, owner, name):
        self.attr_name = "_{}_value".format(name)

    def __get__(self, instance, owner=None):
        return getattr(instance, self.attr_name)
    
    def __set__(self, instance, value):
        setattr(instance, self.attr_name, value)

    def __delete__(self, instance):
        delattr(instance, self.attr_name)
```

Or for registering instances of a descriptor with some abstracted meta-functionality in the parent:

```python
class Field:
    def __set_name__(self, owner, name):
        self.name = name
        owner._fields[self.name] = self
    def __get__(self, instance, owner=None):
        return instance._data[self.name]

class Structure:
    _fields: dict = {}
    def __init__(self, buf):
        self._data = parse(buf, self._fields)

class S(Structure):
    f = Field()
    g = Field()
```
In order to implement this, code needs to be added to run at the end of a class-definition block, iterate through the class's namespace, and then call this method on any of the class members that define it.

### Code Size

This could probably be gated behind whatever feature flag exists for descriptors more generally, but the implementation ought to be rather lightweight anyway.

### Implementation

I intend to implement this feature and submit a pull request. I could use some guidance on locating where in the codebase class creation is handled.

In the interest of TDD, I've already created a test to check this feature here: #15500

### Code of Conduct

Yes, I agree

AJMansfield:
Note, I've already created a test for this feature here: #15500

PR #15503, opened 2024-07-19
py/objtype: Add support for __set_name__. (hazard version)

AJMansfield:
### Summary

This PR implements the feature described in #15501, adding support for the `__set_name__` data model method.

### Testing

This PR includes and passes the unit test originally submitted in #15500 to verify the feature's absence.

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -1124,7 +1124,7 @@ typedef time_t mp_timestamp_t;
 #define MICROPY_PY_FUNCTION_ATTRS_CODE (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_FULL_FEATURES)
 #endif
 
-// Whether to support the descriptors __get__, __set__, __delete__
+// Whether to support the descriptors __get__, __set__, __delete__, __set_name__
 // This costs some code size and makes load/store/delete of instance
 // attributes slower for the classes that use this feature
 #ifndef MICROPY_PY_DESCRIPTORS
--- a/py/objtype.c
+++ b/py/objtype.c
@@ -661,8 +661,8 @@ static void mp_obj_instance_load_attr(mp_obj_t self_in, qstr attr, mp_obj_t *des
     // try __getattr__
     if (attr != MP_QSTR___getattr__) {
         #if MICROPY_PY_DESCRIPTORS
-        // With descriptors enabled, don't delegate lookups of __get__/__set__/__delete__.
-        if (attr == MP_QSTR___get__ || attr == MP_QSTR___set__ || attr == MP_QSTR___delete__) {
+        // With descriptors enabled, don't delegate lookups of __get__/__set__/__delete__/__set_name__.
+        if (attr == MP_QSTR___get__ || attr == MP_QSTR___set__ || attr == MP_QSTR___delete__ || attr == MP_QSTR___set_name__) {
             return;
         }
         #endif
@@ -960,7 +960,7 @@ static bool check_for_special_accessors(mp_obj_t key, mp_obj_t value) {
     #endif
     #if MICROPY_PY_DESCRIPTORS
     static const uint8_t to_check[] = {
-        MP_QSTR___get__, MP_QSTR___set__, MP_QSTR___delete__,
+        MP_QSTR___get__, MP_QSTR___set__, MP_QSTR___delete__, // not needed for MP_QSTR___set_name__ tho
     };
     for (size_t i = 0; i < MP_ARRAY_SIZE(to_check); ++i) {
         mp_obj_t dest_temp[2];
@@ -1240,6 +1240,23 @@ mp_obj_t mp_obj_new_type(qstr name, mp_obj_t bases_tuple, mp_obj_t locals_dict) 
             elem->value = static_class_method_make_new(&mp_type_staticmethod, 1, 0, &elem->value);
         }
     }
+
+    #if MICROPY_PY_DESCRIPTORS
+    // call __set_name__ on all entries (especially descriptors)
+    for (size_t i = 0; i < locals_map->alloc; i++) {
+        if (mp_map_slot_is_filled(locals_map, i)) {
+            elem = &(locals_map->table[i]);
+
+            mp_obj_t set_name_method[4];
+            mp_load_method_maybe(elem->value, MP_QSTR___set_name__, set_name_method);
+            if (set_name_method[1] != MP_OBJ_NULL) {
+                set_name_method[2] = MP_OBJ_FROM_PTR(o);
+                set_name_method[3] = elem->key;
+                mp_call_method_n_kw(2, 0, set_name_method);
+            }
+        }
+    }
+    #endif
 
     return MP_OBJ_FROM_PTR(o);
 }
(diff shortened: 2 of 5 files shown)


dpgeorge:
Thanks for the contribution.

I think this is a good addition.  It's important to support this `__set_name__` special-method to properly make use of descriptors (which MicroPython does support, at least with get/set/delete).

I'm just wondering about the performance hit.  It means that each class that's defined will need to iterate through all of its members.  I wonder if there's a way to make use of `MP_TYPE_FLAG_HAS_SPECIAL_ACCESSORS` to decide whether to do this search or not.  There's already `check_for_special_accessors()` that's run on class creation in a loop over all members.  Maybe that can be reused instead of essentially duplicating the work?

AJMansfield:
> I'm just wondering about the performance hit.

I do think the extra loop is necessary, for reasons described below -- but also, this code only runs at class-creation time, something that in most cases should only happen a small fixed number of times during setup. (At least, outside of intractable runtime dynamic metaclassing scenarios for which performance is already a lost cause...)

> I wonder if there's a way to make use of `MP_TYPE_FLAG_HAS_SPECIAL_ACCESSORS` to decide whether to do this search or not.

That could be done if we are only interested in addressing instance-attribute-like descriptors, but that would miss classattr- and classmethod-like descriptors that only use `__set_name__` to bind to the containing class. We could try to special-case that out too, but it's much simpler to just check the thing itself and look up `MP_QSTR___set_name__` directly.

> There's already check_for_special_accessors() that's run on class creation in a loop over all members. Maybe that can be reused instead of essentially duplicating the work?

A check for a `__set_name__` sub-member _could_ be added as part of that loop, but there could be a sequencing hazard with invoking user code anywhere except right at the end -- and even then, invoking before attribute access is fully ready (incl. having the special accessor flag correctly set) would become a sequencing hazard if/when metaclasses become supported.

My opinion, is that it's better to just keep the loop separate. It does mean chasing down member pointers a second time, but the expensive part of the loop is looking up `MP_QSTR___set_name__` on each member, which needs to be done whether or not it's part of the same loop with `check_for_special_accessors`.

AJMansfield:
Update: I'm currently investigating some corner-case behavior I recently found around what happens (and what _should_ happen) when a `__set_name__` function inserts or removes elements from the class being initialized.

At the moment I'm still waiting on python/cpython#122381 to define what the exact semantics should be in that case -- but this PR is still valid for a merge.

Whatever the correct semantics are, they'll need at least a small code-size increase. And with how rarely that gets used, I feel like it'll probably be better gate them behind the `MICROPY_PY_METACLASSES` flag from #15511, and make that a separate PR.

@dpgeorge should I spend the time adding a cpydiff to this PR about what the exact mismatch is?

dpgeorge:
> should I spend the time adding a cpydiff to this PR about what the exact mismatch is?

Yes please.  Then at least all this work/knowledge is encoded in a test and can be improved later on.

AJMansfield:
Not exactly proud of needing to resort to a stochastic test to reliably show off the sequence hazard I was concerned about, but I have a reliable cpydiff now for the issue I was worried about.

Fundamentally, modify-while-iterating bug. CPython avoids this bug in its `__set_name__` implementation because it effectively iterates on `dict.keys()`, which creates a copy of the current set of keys instead of just a view. However, my proposed implementation for Micropython iterates through the raw allocated slots in `locals_map` (the same way the other steps that iterate its namespace do).

As a result, in my `cpydiff/core_class_setname_hazard` test case, involving a class with 26 descriptors with names `a` through `z` that each insert 10 additional attributes into the class at random, CPython ends up calling `__set_name__` exactly once on only these original descriptors -- but my proposed micropython code calls some of them multiple times, and some not at all, as the order of the dict entries being iterated on shifts in response to the new insertions:

```
+-------------+-------------+
| CPy output: | uPy output: |
+-------------+-------------+
|     a 1     |     a 2     |
|     b 1     |     b 1     |
|     c 1     |     c 2     |
|     d 1     |     d 3     |
|     e 1     |     e 4     |
|     f 1     |     f 3     |
|     g 1     |     g 4     |
|     h 1     |     h 2     |
|     i 1     |     i 1     |
|     j 1     |     j 3     |
|     k 1     |     k 2     |
|     l 1     |     l 1     |
|     m 1     |     m 1     |
|     n 1     |     n 1     |
|     o 1     |     o 1     |
|     p 1     |     p 0     |
|     q 1     |     q 1     |
|     r 1     |     r 0     |
|     s 1     |     s 0     |
|     t 1     |     t 1     |
|     u 1     |     u 3     |
|     v 1     |     v 0     |
|     w 1     |     w 1     |
|     x 1     |     x 0     |
|     y 1     |     y 0     |
|     z 1     |     z 0     |
+-------------+-------------+
```

A similar effect can happen when `__set_name__` removes instead of adds; but the worst case is still just descriptors having their `__set_name__` missed or called multiple times.

There's no hazard in cases where there are no class-mutating descriptors, and this might not be a use-case it's worthwhile to support. I'll see if I can create an alternative pull-request for a version that exactly matches CPython, though, so we can compare the performance cost of creating that copy on more than just speculation.

AJMansfield:
I've done some benchmarking using a new suite of internalbench class creation benchmarks (see PR #16825) and have compared the benchmark times from the base branch to this branch.

There's a lot of noise in this (I'll see if I can run these on real hardware at some point), but overall this patch makes processing classes take about 12% longer:

```
internal_bench/class_create:
    0.349 -> 0.391 (+12%) internal_bench/class_create-0-empty.py
    0.476 -> 0.544 (+14%) internal_bench/class_create-1-slots.py
    0.490 -> 0.553 (+13%) internal_bench/class_create-1.1-slots5.py
    0.432 -> 0.483 (+12%) internal_bench/class_create-2-classattr.py
    0.776 -> 0.868 (+12%) internal_bench/class_create-2.1-classattr5.py
    0.461 -> 0.533 (+16%) internal_bench/class_create-3-instancemethod.py
    0.477 -> 0.535 (+12%) internal_bench/class_create-4-classmethod.py
    0.452 -> 0.511 (+13%) internal_bench/class_create-4.1-classmethod_implicit.py
    0.494 -> 0.537 (+09%) internal_bench/class_create-5-staticmethod.py
    0.458 -> 0.512 (+12%) internal_bench/class_create-6-getattribute.py
    0.474 -> 0.518 (+09%) internal_bench/class_create-6.1-getattr.py
    0.386 -> 0.447 (+16%) internal_bench/class_create-6.2-descriptor.py
    0.517 -> 0.625 (+21%) internal_bench/class_create-6.3-descriptor_setname.py
    0.419 -> 0.480 (+15%) internal_bench/class_create-6.4-property.py
    0.363 -> 0.420 (+16%) internal_bench/class_create-7-inherit.py
    0.368 -> 0.399 (+09%) internal_bench/class_create-7.1-inherit_initsubclass.py
```

There were also two other benchmark tests that got concerningly slower:
```
internal_bench/arrayop:
    0.174 -> 0.194 (+12%) internal_bench/arrayop-3-bytearray_inplace.py
internal_bench/loop_count:
    0.268 -> 0.310 (+16%) internal_bench/loop_count-2-range_iter.py
```

None of the other tests in their families seemed to be affected, but my tests for #16806 and #16816 both showed these same tests slowed by near-enough the exact same margins as class creation.

AJMansfield:
Closed in favor of #16806

PR #15509, opened 2024-07-20
tests/cpydiff/core_class: Document issue with super in classmethod.

AJMansfield:
### Summary

This documents an existing problem with the way `super()` is implemented that became evident while I was implementing `__init_subclass__`, where when used inside a classmethod it defers to the ancestor of the class object itself (i.e. `type`), rather than parent classes it inherits from.

This is probably an issue with super itself not treating the second argument differently when its a type in the way that CPython does.

### Testing

I've verified this against CPython versions 3.8, 3.10, and 3.12.


dpgeorge:
Thanks for the test case / report.

I was curious about why the difference existed, and upon investigating managed to fix it.  See #15533.

PR #15511, opened 2024-07-20
py/objtype: Add basic __init_subclass__ metaclass support.

AJMansfield:
### Summary

This PR adds partial support for the `__init_subclass__` data model method defined in [PEP487](https://peps.python.org/pep-0487/), and gates this added features behind a new feature flag, `MICROPY_PY_METACLASSES`.
While this PR does not implement "full" metaclasses, it does implement the main mechanism used in practice for most types of meta-class-like class customization.

The behavior implemented here varies from the CPython behavior for two reasons:
- Due to #15509, user-written `__init_subclass__` implementations are not able to recursively call `super().__init_subclass__()` the way PEP487 specifies. Therefore, this implementation applies a workaround: instead of invoking `__init_subclass__` only for the last base class, it invokes it on all direct base classes.
- Micropython does not support passing kwargs as bases of a class, therefore `__init_subclass__` currently has no kwargs functionality.

This PR uses #15503 as its base to ensure the `__init_subclass__` code remains in the correct order with `__set_name__` in order to match CPython. It also adds `MICROPY_PY_METACLASSES` as an alternative condition for enabling #15503's functionality, for the rare cases of metaclass code that relies on `__set_name__` class customization but doesn't need the per-lookup overhead of descriptors.

### Testing

This PR includes a test verifying this feature, and three cpydiff's precisely outlining the mismatch with CPython. I've also manually verified the functionality using the unix port.

This increases code size, but only when the corresponding feature flag is enabled.


dpgeorge:
Thanks for this, it looks like a pretty useful feature (I've never used it myself though in CPython).

Especially if MicroPython will never support full CPython metaclasses, this is a good (and more modern) approach to it.

AJMansfield:
I'm currently looking to see how to address the issue with the init_subclass recursion now given #15533.

dpgeorge:
> I'm currently looking to see how to address the issue with the init_subclass recursion now given #15533.

Oh, that's interesting, that the fix in 15533 breaks this PR.

I think 15533 is necessary and should be merged.  It's a very simple fix and fixes an edge case that's very difficult to notice in a running program, let alone debug and work around.

AJMansfield:
> Oh, that's interesting, that the fix in 15533 breaks this PR.

No, mostly it just makes one of the behavioral compromises in the original version of this PR obsolete and unnecessary -- since it was possible to fix that fairly quickly, merging the original half-measure is unnecessary.

t actually _fixes_ cpydiff/core_class_initsubclass_recursive, so I'm moving it back to tests -- though frustratingly, it wasn't quite enough to fix cpydiff/core_class_initsubclass_multi the way I'd originally hoped, even if I think I understand why now.

I think the fix for that will be tweaking to `super_attr` to special-case `__init_subclass__` lookups to default to a no-op function instead of failing. (Alternatively, we _could_ just add a default `__init_subclass__` implementation to base object; I just think it's better to limit the scope of the performance impact and tweak super instead.)

dpgeorge:
> For that matter, it actually _fixes_ cpydiff/core_class_initsubclass_recursive, so I'll be moving it back to tests

Very good!  Then merging #15533 is the right way to move forward.

> though frustratingly, it wasn't quite enough to fix cpydiff/core_class_initsubclass_multi the way I'd originally hoped, even if I think I understand why now.
> 
> I think the fix for that will be tweaking to `super_attr` to special-case `__init_subclass__` lookups to default to a no-op function instead of failing. (Alternatively, we _could_ just add a default `__init_subclass__` implementation to base object; I just think it's better to limit the scope of the performance impact and tweak super instead.)

Hmm, I'm not sure at this stage what the best option is... do what you think is best then we can review it.

AJMansfield:
Retracting in favor of further discussion and consensus about this feature.

PR #15517, opened 2024-07-21
py/py.mk: Add USERMOD_LIB_ASM to include assembly files.

george-hopkins:
### Summary

Introduce `USERMOD_LIB_ASM` to allow users to include assembly files as part of their user modules. It could be used to include optimized functions or outputs of other programming languages.

### Testing

Tested on `unix`.

* Assemble files yourself and use `LDFLAGS_USERMOD` to include the compiled functions as a library.


dpgeorge:
Thanks for the contribution, this is a useful addition.

(Assembly files can be .s or .S.  The latter are compiled with the C compiler and can contain macros, which is more general/useful than .s, so I think supporting .S is the right thing.)

PR #15519, opened 2024-07-22
extmod/vfs_fat.c: Add default volume label to mkfs if it exists

tstenvold:
### Summary

Using mkfs doesn't set a volume label for FAT filesystems, this PR will set the default volume label if it exists.

### Testing
Tested on a PYBV11

Alternatives would be to add a new function for setting/getting the label or adding a second optional argument to mfks for setting the volume label. Both of which would add more code for little gain.


dpgeorge:
Thanks, this is a good solution.

PR #15520, opened 2024-07-22
py/misc: Fix msvc and C++ compatibility.

stinos:
### Summary

Fix 3 warning types which occur when compiling misch.h with a C++ compiler, msvc with most warnings enabled in this case, since 908ab1ce

### Testing

Compiles with msvc in C++ mode now.

The 'conditional expression is constant' on `if (sizeof(mp_int_t) == sizeof(long)) {` was fixed by adding making this

```
static inline bool mp_check(bool value) {
    return value;
}

... if (mp_check(sizeof(mp_int_t) == sizeof(long))) { ...
``` 

instead of disabling the warning by surrounding the code with something like
```
#if _MSC_VER
#pragma warning(push)
#pragma warning(disable: 4127)
#endif
if (sizeof(mp_int_t) == sizeof(long)
#if _MSC_VER
#pragma warning(pop)
#endif
```
because the latter is ugly and a lot of text for something simple. I don't know of another cross-platform way to do this (C preprocessor cannot compare types nor can use sizeof) and `if constexpr` is only available fo C++-17.

### DIFF
--- a/py/misc.h
+++ b/py/misc.h
@@ -338,38 +338,44 @@ typedef const char *mp_rom_error_text_t;
 #ifdef _MSC_VER
 #include <intrin.h>
 
-static uint32_t mp_clz(uint32_t x) {
+static inline uint32_t mp_clz(uint32_t x) {
     unsigned long lz = 0;
     return _BitScanReverse(&lz, x) ? (sizeof(x) * 8 - 1) - lz : 0;
 }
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks, this looks good.

Not sure why all those static functions in `py/misc.h` weren't `static inline` to begin with.... that's an oversight by the person who merged that PR :grimacing: :joy:

PR #15529, opened 2024-07-23
docs/neopixel: Mention bitstream timing tuple

scy:
### Summary

I had to communicate with a NeoPixel-like strip with slightly non-standard timings (neither the usual 400 nor 800 kHz ones) and only noticed the `NeoPixel` constructor accepting a timing tuple when reading the source code. It should be mentioned in the docs, too.

### Testing

None, documentation change.


dpgeorge:
Thanks for the contribution!

Issue #15528, opened 2024-07-23
Update libffi to 3.3 or later to unblock CI for RV64.


agatti:
### Description

After the fix for #12838 has landed, I've looked into adding CI support for the Unix port on RISC-V 64 bits.

Whilst normal builds work (as in, no cross compilation involved), cross-compilation using the normal features set for the Unix port to an RV64 target fails because libffi is too old - RISC-V support was added in version 3.3 of libffi (almost five years ago) and MicroPython uses 3.2.1.  I see that MicroPython's libffi submodule points to a fork that backports a few commits, maybe those changes are not needed anymore on a more recent version of libffi?

I've marked this as an improvement/enhancement rather than a bug as things still work as they are at the moment and whilst RISC-V 64 bits is not a priority for MicroPython, an update to libffi would impact on all other Unix ports at the very least.  On the other hand, having some sort of early warning for code breakages on RISC-V may prove helpful, especially as MicroPython for RISC-V 64 is now packaged in the OpenWRT distribution (see the original bug report), and is probably brought in somewhere else.

### Code Size

I have no idea on whether more recent versions of libffi would increase the code size.  Also, libffi would be used on systems much larger than a microcontroller, where modest size increases don't hurt as much as they would elsewhere (although it'd be better if size still stays small anyway).

### Implementation

I hope the MicroPython maintainers or community will implement this feature

### Code of Conduct

Yes, I agree

dlech:
Orthogonal to updating the submodule, if the CI is Ubuntu, it should be possible to install the `libffi-dev:riscv64` package and use that for cross-compiling as well (probably need some extra steps to configure apt for riscv64 foreign arch).

agatti:
That's interesting - the Unix/i386 build installs `libffi-dev:i386` when setting its build environment up so it should be doable.  I'll look into that, thanks!

agatti:
A PR that puts the Unix RV64 port under CI, without having to update the libffi submodule, was submitted.  Thank you, @dlech for your suggestion!

I guess this can be closed then.

dpgeorge:
I think it would be a good idea to update our `lib/libffi` submodule to point to a more recent libffi version.

>  I see that MicroPython's libffi submodule points to a fork that backports a few commits,

Where do you see these commits?  The submodule we point to is `https://github.com/atgreen/libffi`, and following that link now goes to https://github.com/libffi/libff`.  And as far as I can tell, the latter is the official release and its master branch directly descends from the atgreen repo.

[...]

Issue #12838, opened 2023-10-31
Unix port: Error when building 1.21.0 for riscv64


jefferyto:
I'm updating the micropython package in OpenWrt (https://github.com/openwrt/packages/pull/22564) and I'm getting a build error when compiling for riscv64:

```
CC main.c
[...]

PR #15531, opened 2024-07-24
github/workflows: Add RISC-V 64 bits Unix port to CI.

agatti:
### Summary

The Unix port can be built for RISC-V 64 bits but it is not under CI. There are documented usages of the Unix port for RISC-V 64 bits, and until recently it required some external workarounds to build successfully.  Now that it can build out of the box it's not a bad idea to have it under CI.

See the discussion in #15528, and issue #12838 that started it all.

Regarding adding repositories to the CI instances to fetch the RV64 `libffi-dev` binary package from (along with its dependencies), I'm not entirely satisfied with the way it's currently done - it doesn't look safe enough to me.  `add-apt-repository` is able to add repositories using separate command line arguments for each apt source line field, but I haven't found anything that allows to set a specific architecture for the source (or I might have just missed it...).

### Testing

The CI jobs completed successfully when pushing to a local branch in my fork: https://github.com/agatti/micropython/actions/runs/10068525047/job/27834049532

Something worth mentioning is that the test run passes without the explicit listdir/ilistdir and ffi test exceptions, so maybe those exceptions for the other ports should be revised?

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -427,6 +427,11 @@ CI_UNIX_OPTS_QEMU_ARM=(
     MICROPY_STANDALONE=1
 )
 
+CI_UNIX_OPTS_QEMU_RISCV64=(
+    CROSS_COMPILE=riscv64-linux-gnu-
+    VARIANT=coverage
+)
+
 function ci_unix_build_helper {
     make ${MAKEOPTS} -C mpy-cross
     make ${MAKEOPTS} -C ports/unix "$@" submodules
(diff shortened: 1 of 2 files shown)


tools/ci.sh
```diff
@@ -692,6 +697,27 @@ function ci_unix_qemu_arm_run_tests {
     (cd tests && MICROPY_MICROPYTHON=../ports/unix/build-coverage/micropython ./run-tests.py --exclude 'vfs_posix.*\.py')
 }
 
+function ci_unix_qemu_riscv64_setup {
+    for repository in jammy jammy-updates jammy-security
```
dlech, line 701: To make this future-proof, we could `. /etc/os-release` at the beginning of this file. Then we can use `$VERSION_CODENAME` instead of `jammy`.

dpgeorge:
Thanks for this, it looks good.  Nice to be able to test RV64.

> Something worth mentioning is that the test run passes without the explicit listdir/ilistdir and ffi test exceptions, so maybe those exceptions for the other ports should be revised?

Feel free to open a PR to remove those exclusions and see if things pass (they may, they didn't when it was written).

PR #15533, opened 2024-07-24
py/runtime: Fix self arg passed to classmethod when accessed via super.

dpgeorge:
### Summary

MicroPython was not conforming to CPython behaviour when using `super()` within a `classmethod`.  This patch fixes this behaviour.

### Testing

New test added which did not pass before, but now passes.

Thanks to @AJMansfield for the original test case.

This increases code size a little bit, but if not fixed it can lead to some very difficult to debug problems, if you even notice there's a problem in the first place.


dpgeorge:
Won't need #15509 if this is merged.

PR #15523, opened 2024-07-23
esp32: Update ESP-IDF to v5.2.2, optimise size

projectgus:
### Summary

This PR updates the esp32 port's ESP-IDF version to v5.2.2 and changes the version build in CI from v5.0.4 to v5.2.2.

This is in anticipation of adding support for new SoCs ESP32-C6 (see #11869) and ESP32-C2 (see #15440) in the linked PRs.

Propose after merging this and changing the nightly version then we wait a couple of weeks and if no regressions are reported then remove support for older ESP-IDF, including some code cleanup to suit.

### Code Size

The usual problem with updating ESP-IDF is an increase in code size. This PR applies some config changes to reduce resource usage, and also adds a utility script to quickly measure the static resource usage of multiple boards & ESP-IDF versions.

Current master branch resource usage:

| BOARD            | BOARD_VARIANT | IDF Version | Binary Size | Static IRAM Size | Static DRAM Size |
|------------------|---------------|-------------|-------------|------------------|------------------|
| ESP32_GENERIC    |               | v5.0.4      |     1669037 |           128286 |            52892 |
|                  |               | v5.0.5      |     1682469 |           125822 |            53732 |
|                  |               | v5.1.2      |     1708393 |           129398 |            54276 |
|                  |               | v5.2        |     1707709 |           131150 |            55588 |
|                  |               | v5.2.2      |     1663573 |           121966 |            55236 |
| ESP32_GENERIC    | D2WD          | v5.0.4      |     1425121 |           121270 |            52768 |
|                  |               | v5.0.5      |     1435673 |           118666 |            53608 |
|                  |               | v5.1.2      |     1456773 |           121306 |            54160 |
|                  |               | v5.2        |     1446181 |           121578 |            54776 |
|                  |               | v5.2.2      |     1399309 |           112318 |            54340 |
| ESP32_GENERIC    | SPIRAM        | v5.0.4      |     1545837 |           130618 |            53308 |
|                  |               | v5.0.5      |     1557441 |           127802 |            54124 |
|                  |               | v5.1.2      |     1551761 |           128770 |            54540 |
|                  |               | v5.2        |     1543413 |           126870 |            55208 |
|                  |               | v5.2.2      |     1497321 |           117170 |            54788 |
| ESP32_GENERIC_S3 |               | v5.0.4      |     1564905 |           108674 |            49080 |
|                  |               | v5.0.5      |     1581793 |           104290 |            49344 |
[...]

PR #15538, opened 2024-07-25
esp32: Fix heap corruption triggered by bluetooth.active(0).

projectgus:
### Summary

It seems like at some point Espressif NimBLE team changed nimble_port_init and nimble_port_deinit to manage HCI init internally:
https://github.com/espressif/esp-nimble/commit/f8a79b04c9743543b8959727d7

This change is now included in all the IDF versions that current MicroPython supports.

As a result, existing code that called esp_nimble_hci_deinit() explicitly would trigger a use-after-free bug and heap corruption (specifically this calls through to ble_transport_deinit() which calls os_mempool_free(). The second time this writes out to a bunch of memory pools where the backing buffers have already been freed.)

Symptoms were intermittent random crashes after de-activating Bluetooth. Setting Heap Poisoning to Comprehensive in menuconfig caused the bug to be detected every time.

Found while testing #15523.

### Testing

Ran multi_bluetooth tests on ESP32 and ESP32-S3 with and without Comprehensive heap poisoning, and on IDF v5.0.4 and v5.2.2. Failures (crashes) in ble_gatt_data_transfer.py are present without this fix.

### DIFF
--- a/ports/esp32/mpnimbleport.c
+++ b/ports/esp32/mpnimbleport.c
@@ -32,7 +32,6 @@
 
 #define DEBUG_printf(...) // printf("nimble (esp32): " __VA_ARGS__)
 
-#include "esp_nimble_hci.h"
 #include "nimble/nimble_port.h"
 #include "nimble/nimble_port_freertos.h"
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks, this looks good.  I did a quick test with an ESP32 (against PYBD-SF6) with IDF v5.0.5.  Current master does indeed crash on `ble_gatt_data_transfer.py`, and is fixed with this PR.

PR #15539, opened 2024-07-25
esp32: Workaround math.gamma(-inf) result.

projectgus:
### Summary

Without this workaround, `math.gamma(-float("inf"))` returns inf instead of raising a math domain ValueError.
 
Found while testing #15523.

Upstream issue reported as https://github.com/espressif/esp-idf/issues/14273

### Testing

tests/float/math_domain_special.py fails on esp32 without this fix.

### DIFF
--- a/py/modmath.c
+++ b/py/modmath.c
@@ -196,7 +196,17 @@ MATH_FUN_1(erf, erf)
 // erfc(x): return the complementary error function of x
 MATH_FUN_1(erfc, erfc)
 // gamma(x): return the gamma function of x
+#if MICROPY_PY_MATH_GAMMA_FIX_NEGINF
+static mp_float_t MICROPY_FLOAT_C_FUN(tgamma_func)(mp_float_t x) {
+    if (isinf(x) && x < 0) {
+        math_error();
+    }
+    return MICROPY_FLOAT_C_FUN(tgamma)(x);
+}
+MATH_FUN_1(gamma, tgamma_func)
+#else
 MATH_FUN_1(gamma, tgamma)
+#endif
 // lgamma(x): return the natural logarithm of the gamma function of x
 MATH_FUN_1(lgamma, lgamma)
 #endif
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -1393,6 +1393,11 @@ typedef double mp_float_t;
 #define MICROPY_PY_MATH_POW_FIX_NAN (0)
 #endif
 
+// Whether to provide fix for gamma(-inf) to raise ValueError
+#ifndef MICROPY_PY_MATH_GAMMA_FIX_NEGINF
+#define MICROPY_PY_MATH_GAMMA_FIX_NEGINF (0)
+#endif
+
 // Whether to provide "cmath" module
 #ifndef MICROPY_PY_CMATH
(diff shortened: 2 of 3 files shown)


dpgeorge:
This is related to https://github.com/espressif/esp-idf/issues/7282, where I reported that `math.tgamma(-inf)` was broken.

I guess it's fine to work around the `math.gamma(-inf)` issue with our custom config option (otherwise we could report it to IDF and hope for an eventual fix).

projectgus:
> This is related to [espressif/esp-idf#7282](https://github.com/espressif/esp-idf/issues/7282), where I reported that `math.tgamma(-inf)` was broken.

This is calling tgamma, so according to that issue it should be fixed... :thinking:

projectgus:
Looks like tgammaf() wasn't fixed when tgamma() was. Have opened another ESP-IDF issue and linked it from the PR description. 

Suggest we merge this workaround until an upstream fix is available.

py/mpconfig.h
```diff
@@ -1393,6 +1393,11 @@ typedef double mp_float_t;
 #define MICROPY_PY_MATH_POW_FIX_NAN (0)
 #endif
 
+// Whether to provide fix for gamma(-inf) to raise ValueError
+#ifndef MICROPY_PY_MATH_GAMMA_FIX_NEG_INF
+#define MICROPY_PY_MATH_GAMMA_FIX_NEG_INF (0)
```
dpgeorge, line 1398: I suggest to call it `MICROPY_PY_MATH_GAMMA_FIX_NEGINF` (no underscore near the end) to match other names, like `NEGZERO`.

py/modmath.c
```diff
@@ -196,7 +196,17 @@ MATH_FUN_1(erf, erf)
 // erfc(x): return the complementary error function of x
 MATH_FUN_1(erfc, erfc)
 // gamma(x): return the gamma function of x
+#if MICROPY_PY_MATH_GAMMA_FIX_NEG_INF
+static mp_float_t tgamma_funcf(mp_float_t x) {
+    if (__builtin_isinf_sign(x) == -1) {
```
dpgeorge, line 201: We don't use builtin's for floats anywhere else.  Is this portable?  Can we use `isinf(x) && x < 0` instead?  It may compile to the same thing anyway.

projectgus: Yeah, I settled on this because in theory `isinf(x) == -1` should work but it doesn't on ESP32. `isinf` is actually a BSD extension, the actual C99 version is way more verbose to test the same thing: https://www.gnu.org/software/libc/manual/html_node/Floating-Point-Classes.html

However gcc does ship a builtin for BSD `isinf`, so I used that.

But your suggestion seems good, will change it.

dpgeorge:
@projectgus for a change like this, ideally it would be split into two commits, one for the new option added to modmath.c, and a second for the esp32 port to enable this option to work around the bug; eg the first commit would have a message like `py/modmath: Add option to work around -inf bug in a port's tgamma.`.

Can you please split the commit that way?

projectgus:
@dpgeorge Sure, have split and re-ran float tests on unix port and esp32 port.

dpgeorge (approved):

Issue #2971, opened 2017-03-21
RFC: defining machine.Timer


dpgeorge:
This issue is to discuss the API of the machine.Timer class.  There are currently 3 different version of Timer which don't have much in common:

esp8266:
```python
# id must be -1, period is in milliseconds, mode is PERIODIC or ONE_SHOT
Timer(id, *, period, mode, callback) # constructor
Timer.init(*, period, mode, callback) # reinitialise all parameters
Timer.deinit() # stop the timer
```

cc3200/wipy:
```python
# mode is PERIODIC, ONE_SHOT or PWM
Timer(id, mode, *, width=16) # constructor
Timer.init(mode, *, width=16) # reinitialise
Timer.deinit() # stop

# each timer has one or more channels
# each timer channel is associated with a specific IO pin (eg for PWM)
Timer.channel(*, freq, period, polarity, duty_cycle) # create channel assoc with TImer
TimerChannel.freq([freq]) # get/set
TimerChannel.period([period]) # get/set
TimerChannel.duty_cycle([duty]) # get/set
TimerChannel.irq(...) # get or set up the IRQ callback
```

stmhal/pyboard (actually pyb.Timer):
```python
# freq is in Hz (can be float <1), prescaler/period are raw values, mode is UP/DOWN/CENTER
Timer(id, *, freq, prescaler, period, mode, div, callback, deadtime) # constructor
Timer.init(*, freq, prescaler, period, mode, div, callback, deadtime) # reinit
Timer.counter([value]) # get/set the raw timer counter
Timer.source_freq() # get source frequency that clocks this timer
Timer.freq([value]) # get/set
Timer.prescaler([value]) # get/set
Timer.period([value]) # get/set
Timer.callback(callable) # get/set callback (pass None to disable)

# each timer has multiple chanels (usually 4)
# pin is the pin to associate with the channel (eg for PWM)
Timer.channel(mode, *, callback, pin, pulse_width, pulse_width_percent, compare, polarity) # create
TimerChannel.pulse_width([value]) # get/set
TimerChannel.pulse_width_percent([value]) # get/set
TimerChannel.capture([value]) # get/set
TimerChannel.compare([value]) # get/set
TimerChannel.callback(callable) # get/set callback
```

To compare current usage, for a simple periodic timer at 10Hz one would do:
```python
# esp8266
t = Timer(-1)
t.init(period=100, mode=Timer.PERIODIC, callback=lambda t:print(t))

# stmhal/pyboard
[...]

PR #15540, opened 2024-07-25
tests,esp32: Skip soft timer test on esp32 port, rename.

projectgus:
### Summary

Rename the extmod/machine_timer.py test to machine_soft_timer.py (as this is what it tests). Skip it on esp32, which has no soft timer support.

Found while testing #15523.

### Testing

Allows this test to not fail on esp32.

### DIFF
--- a/tests/extmod/machine_soft_timer.py
+++ b/tests/extmod/machine_soft_timer.py
@@ -1,4 +1,11 @@
-# test machine.Timer
+# test "soft" machine.Timer (no hardware ID)
+import sys
+
+
+if sys.platform == "esp32":
+    print("SKIP")  # TODO: Implement soft timers for esp32 port
+    raise SystemExit
+
 
 try:
     import time, machine as machine


dpgeorge:
Seems like this PR accidentally got some I2S stuff committed as well?

projectgus:
> Seems like this PR accidentally got some I2S stuff committed as well?

Oops, thanks. Fixed.

dpgeorge:
See related #15284.  In particular my comment https://github.com/micropython/micropython/pull/15284#issuecomment-2172004106

The reason I never got around to fixing this test is because I kinda wanted it to be a reminder to fix this properly on esp32, ie allow it to create timers without an id.  Because at the moment `machine.Timer` is really not very portable, and for such a basic thing like making a timer we should aim to make the API fully portable, see #2971.

Ultimately I'd like to see:
- esp32 support soft timers
- this existing test would then pass on esp32, probably best renamed to `tests/extmod/machine_soft_timer.py`
- add a separate `tests/extmod/machine_timer.py` test which tests hardware timers on various ports

In the meantime... maybe just let this test fail on esp32, or at least skip?

tests/extmod/machine_timer.py
```diff
@@ -1,38 +1,53 @@
...
 t.deinit()
 
 # deinit again
 t.deinit()
 
 # create 2 and deinit
-t = machine.Timer(freq=1)
-t2 = machine.Timer(freq=1)
+t = machine.Timer(TIMER_ID, freq=1)
+t2 = machine.Timer(TIMER_ID, freq=1)
```
dpgeorge, line 31: Since this is testing having multiple timers at once, probably this would need a separate `TIMER_ID2`.

projectgus:
> probably best renamed to tests/extmod/machine_soft_timer.py
> or at least skip?

What about these two bits for now? That's what I've done here as they're the low hanging fruit. Skipping seems appropriate given the feature isn't implemented on that port for now. I would also like to see soft timers on esp32 port, in the long term.

dpgeorge:
> What about these two bits for now?

Yes, looks good now.

PR #15541, opened 2024-07-25
esp32: Fix I2S init error handling, add test coverage.

projectgus:
### Summary

* Fix I2S initialisation when requested DMA buffer size is small (driver always needs 2 buffers). 
* Fixes silent failure during I2S init, will now raise an exception if initialisation fails.
* Adds support for ESP32 in the extmod/machine_i2s_rate test.

Found while testing #15523.

### Testing

With these changes machine_i2s_rate test now passes on esp32.

### DIFF
--- a/ports/esp32/machine_i2s.c
+++ b/ports/esp32/machine_i2s.c
@@ -151,7 +151,7 @@ static uint32_t get_dma_buf_count(uint8_t mode, i2s_data_bit_width_t bits, forma
 
     uint32_t dma_buf_count = ibuf / (DMA_BUF_LEN_IN_I2S_FRAMES * dma_frame_size_in_bytes);
 
-    return dma_buf_count;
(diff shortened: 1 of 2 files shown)


ports/esp32/machine_i2s.c
```diff
@@ -151,7 +151,7 @@ static uint32_t get_dma_buf_count(uint8_t mode, i2s_data_bit_width_t bits, forma
 
     uint32_t dma_buf_count = ibuf / (DMA_BUF_LEN_IN_I2S_FRAMES * dma_frame_size_in_bytes);
 
-    return dma_buf_count;
+    return MAX(dma_buf_count, 2);
```
dpgeorge, line 154: I think this requires a small comment like "ESP-IDF driver always requires at least two DMA buffers".

PR #15543, opened 2024-07-25
tests: Add machine_spi_rate test.

projectgus:
### Summary

Based on machine_i2s_rate, allows testing basic SPI functionality and timings.

Implemented for rp2, esp32 and pyboard.

- [x] Fix on pyboard (currently shows up to 14ms difference. Maybe need to account for quantisation error due to available SPI clock dividers.)

Implemented while testing #15523.

### Testing

Tested on PyBoard V1.1, RP2 Pico, ESP32, ESP32-S3, ESP32-C3.


dpgeorge:
> * Fix on pyboard (currently shows up to 14ms difference. Maybe need to account for quantisation error due to available SPI clock dividers

I suggest to print the SPI object (eg something like `print(spi, file=stringio)`) and from that parse the actual baudrate, then use that to compute the expected timing.

projectgus:
> I suggest to print the SPI object (eg something like `print(spi, file=stringio)`) and from that parse the actual baudrate, then use that to compute the expected timing.

Yeah, I was hoping to avoid the complexity of this but I think it's going to be the way forward.

tests/extmod/machine_spi_rate.py
```diff
@@ -0,0 +1,128 @@
...
+elif "rp2" in sys.platform:
+    spi_instances = ((0, Pin(18), Pin(19), Pin(16)),)
+elif "esp32" in sys.platform:
+    spi_instances = [
+        (1, Pin(18), Pin(19), Pin(21)),
+    ]
+    if "ESP32C3" not in str(sys.implementation):
+        spi_instances.append((2, Pin(18), Pin(19), Pin(21)))
+else:
+    raise SystemExit("Please add support for this test on this platform.")
```
dpgeorge, line 29: I don't think this message ends up anywhere the user (person running the test) can see it.  It's not in the generated .out file.  The test just fails but the .out has no indication of why.  Probably better to make this an assert like `assert False, "platform not supported"`.

projectgus: Oh right, thanks - I don't know why I thought this worked on MP.

I went for "print and then SystemExit" as that's already a common pattern.

projectgus:
Updated, re-tested on the implemented ports.

tests/extmod/machine_spi_rate.py
```diff
@@ -0,0 +1,152 @@
...
+    # Return the 'real' baudrate for a SPI object, from parsing 'print' output
+    # i.e.
+    # SPI(id=1, baudrate=500000, polarity=0, phase=0, bits=8, firstbit=0, sck=14, mosi=13, miso=12)
+    #
+    # Note the 'real' rate may be quite different to the requested rate, i.e.
+    # on ports where the SPI hardware only supports power of 2 clock dividers.
+    #
+    # Implementation is somewhat fiddly and inefficient to avoid dependency on
+    # 're' module,
+    s = str(spi)
```
projectgus, line 44: @dpgeorge I went with str() here rather than the suggestion of printing to StringIO because it seems like this should work everywhere (string constructor calls through to mp_obj_print_helper()`.)

dpgeorge: Oh yeah, that's much simpler!

dpgeorge (approved):
Looks good.  I tested on Pico (rp2 port), an ESP32 board, PYBv1.0 and PYBD-SF6.  Works well.

PR #15549, opened 2024-07-25
shared/runtime/gchelper: Add RISC-V RV64I native gchelper.

agatti:
### Summary

This adds native gchelper support to RV64, given that said platform is now under CI and the code changes required are minimal.  The Unix port already enables platform-specific gchelper code (albeit the generic version), so the RV64 platform deserved its own entry in there.

### Testing

This has been tested on the CI environment both using the generic gchelper code and using the assembler version with this change in the Unix port makefile:

```diff
diff --git a/ports/unix/Makefile b/ports/unix/Makefile
index d5bd6d409..2fab170bf 100644
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -214,16 +214,19 @@ SRC_C += \
        $(wildcard $(VARIANT_DIR)/*.c)
 
 SHARED_SRC_C += $(addprefix shared/,\
-       runtime/gchelper_generic.c \
+       runtime/gchelper_native.c \
        timeutils/timeutils.c \
        $(SHARED_SRC_C_EXTRA) \
        )
 
+SRC_O += shared/runtime/gchelper_rv64i.o
 SRC_CXX += \
 
 OBJ = $(PY_O)
 OBJ += $(addprefix $(BUILD)/, $(SRC_C:.c=.o))
 OBJ += $(addprefix $(BUILD)/, $(SRC_CXX:.cpp=.o))
+OBJ += $(addprefix $(BUILD)/, $(SRC_O:.s=.o))
 OBJ += $(addprefix $(BUILD)/, $(SHARED_SRC_C:.c=.o))
```

On RV64, a setjmp frame consumes close to half a kilobyte, so having an alternative implementation for gchelper on that platform would give some interesting memory savings.  The proposed implementation takes around 128 bytes of memory, which is far less than the setjmp version.

### DIFF
--- a/ports/unix/mpconfigport.h
+++ b/ports/unix/mpconfigport.h
@@ -105,7 +105,7 @@ typedef long mp_off_t;
 // Always enable GC.
 #define MICROPY_ENABLE_GC           (1)
 
-#if !(defined(MICROPY_GCREGS_SETJMP) || defined(__x86_64__) || defined(__i386__) || defined(__thumb2__) || defined(__thumb__) || defined(__arm__))
+#if !(defined(MICROPY_GCREGS_SETJMP) || defined(__x86_64__) || defined(__i386__) || defined(__thumb2__) || defined(__thumb__) || defined(__arm__) || (defined(__riscv) && (__riscv_xlen == 64)))
 // Fall back to setjmp() implementation for discovery of GC pointers in registers.
 #define MICROPY_GCREGS_SETJMP (1)
 #endif
--- a/shared/runtime/gchelper_generic.c
+++ b/shared/runtime/gchelper_generic.c
@@ -150,23 +150,24 @@ static void gc_helper_get_regs(gc_helper_regs_t arr) {
     arr[10] = x29;
 }
 
-#elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 32)
+#elif defined(__riscv) && (__riscv_xlen <= 64)
 
-// Fallback implementation for RV32I, prefer gchelper_rv32i.s
+// Fallback implementation for RV32I and RV64I, prefer gchelper_rv32i.s
+// for RV32I targets or gchelper_rv64i.s for RV64I targets.
 
 static void gc_helper_get_regs(gc_helper_regs_t arr) {
-    register long s0 asm ("x8");
-    register long s1 asm ("x9");
-    register long s2 asm ("x18");
-    register long s3 asm ("x19");
-    register long s4 asm ("x20");
-    register long s5 asm ("x21");
-    register long s6 asm ("x22");
-    register long s7 asm ("x23");
-    register long s8 asm ("x24");
-    register long s9 asm ("x25");
-    register long s10 asm ("x26");
-    register long s11 asm ("x27");
+    register uintptr_t s0 asm ("x8");
+    register uintptr_t s1 asm ("x9");
+    register uintptr_t s2 asm ("x18");
+    register uintptr_t s3 asm ("x19");
+    register uintptr_t s4 asm ("x20");
+    register uintptr_t s5 asm ("x21");
+    register uintptr_t s6 asm ("x22");
+    register uintptr_t s7 asm ("x23");
+    register uintptr_t s8 asm ("x24");
+    register uintptr_t s9 asm ("x25");
+    register uintptr_t s10 asm ("x26");
+    register uintptr_t s11 asm ("x27");
     arr[0] = s0;
     arr[1] = s1;
     arr[2] = s2;
--- a/shared/runtime/gchelper_rv64i.s
+++ b/shared/runtime/gchelper_rv64i.s
@@ -3,7 +3,7 @@
  *
  * The MIT License (MIT)
  *
- * Copyright (c) 2019 Damien P. George
+ * Copyright (c) 2024 Alessandro Gatti
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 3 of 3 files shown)


ports/unix/mpconfigport.h
```diff
@@ -105,7 +105,7 @@ typedef long mp_off_t;
 // Always enable GC.
 #define MICROPY_ENABLE_GC           (1)
 
-#if !(defined(MICROPY_GCREGS_SETJMP) || defined(__x86_64__) || defined(__i386__) || defined(__thumb2__) || defined(__thumb__) || defined(__arm__))
+#if !(defined(MICROPY_GCREGS_SETJMP) || defined(__x86_64__) || defined(__i386__) || defined(__thumb2__) || defined(__thumb__) || defined(__arm__) || (defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 64)))
```
dpgeorge, line 108: Does this need the `__riscv_xlen` part of the logic test?  If you build the unix port on RV32 then it should also work with the native NLR.

shared/runtime/gchelper.h
```diff
@@ -41,7 +41,7 @@ typedef uintptr_t gc_helper_regs_t[4];
 typedef uintptr_t gc_helper_regs_t[10];
 #elif defined(__aarch64__)
 typedef uintptr_t gc_helper_regs_t[11]; // x19-x29
-#elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 32)
+#elif defined(__riscv) && defined(__riscv_xlen) && ((__riscv_xlen == 32) || (__riscv_xlen == 64))
```
dpgeorge, line 44: Is the `__riscv_xlen` logic needed at all?  Is `#elif defined(__riscv)` enough?  (Is there any other RV arch other than 32 and 64 bits?)

dpgeorge:
As a side note: do you think it's possible to add CI to build and run tests on unix with riscv32?

shared/runtime/gchelper.h
agatti: There's also [RV128](https://github.com/riscv/riscv-isa-manual/blob/main/src/rv128.adoc), which as you can imagine has 128-bits wide registers.  Can't say I've ever seen that in the wild though.

ports/unix/mpconfigport.h
agatti: You are indeed correct, however consider that having Linux on RV32 is really uncommon.  I don't think I've ever seen an RV32 chip that can run a Linux environment for real work, short of a soft-core running on a high-end FPGA.

If you don't mind I'd leave this here and then provide a PR that handles both review points and brings RV32 under CI.

agatti:
> As a side note: do you think it's possible to add CI to build and run tests on unix with riscv32?

In theory it should just be a matter of changing the arch in the GCC command line, as I guess the `libffi` package brought in to build RV64 would handle RV32 as well.  As mentioned in the first review point, I'd prefer sending a PR with just the RV32 bits in, code changes and all.

ports/unix/mpconfigport.h
dpgeorge: OK.

shared/runtime/gchelper.h
dpgeorge: Ah, RV128!  Wow.

I guess the check could then be `__riscv_xlen <= 64`, just to make it a tiny bit simpler.

dpgeorge:
> As mentioned in the first review point, I'd prefer sending a PR with just the RV32 bits in, code changes and all.

For sure, a separate PR would be good for that.

(If needed you can disable ffi altogether for the unix RV32 CI.)

shared/runtime/gchelper.h
agatti: But what if somebody comes up with [RV16](https://dlnext.acm.org/doi/10.1007/s11390-022-0910-x)?  Jokes aside, `__riscv_xlen <= 64` can work, and according to the [RISC-V C API document](https://github.com/riscv-non-isa/riscv-c-api-doc/blob/main/riscv-c-api.md), `__riscv_xlen` must always be defined.  I'll change this bit, thanks for reminding me.

agatti:
I've cleaned up the `__riscv_xlen` bits so all currently testable configurations are covered.  I've also changed the data type holding the register contents to something more appropriate, so in the remote eventuality somebody wants to use this with RV128 all they have to do is update the `__riscv_xlen` check.

agatti:
Seems like the RISC-V Linux compiler toolchain that's available in the CI environment cannot be used to build 32-bits binaries.  GCC itself claims it's multilib-aware but there's no multilib compiler package from the repo, and no 32-bits include/libs are installed with the compiler.

dpgeorge:
> Seems like the RISC-V Linux compiler toolchain that's available in the CI environment cannot be used to build 32-bits binaries. GCC itself claims it's multilib-aware but there's no multilib compiler package from the repo, and no 32-bits include/libs are installed with the compiler.

OK, so let's abandon that idea (unix RV32 CI).

Instead I'm getting closer to having the qemu-* ports run native .mpy files as part of CI.  That will help test more of the native emitter.

agatti:
> Instead I'm getting closer to having the qemu-* ports run native .mpy files as part of CI. That will help test more of the native emitter.

That'd be great to have.  I'm also working on natmod support for RV32 and being able to build and run .mpy files in QEMU would help a lot (right now only a few natmods from the `examples/` directory work out of the box, but some show similar issues to the failed RV32 tests: wrong parameter types, crashes during init, etc.).

dpgeorge (approved):

PR #15550, opened 2024-07-26
renesas-ra/usb: Use interrupt rather than polling for usb task.

andrewleech:
### Summary

Most ports using TinyUSB now service the usb tasks from the usb interrupt. Based on issues reported in https://github.com/micropython/micropython/pull/14462#issuecomment-2251555449 I noticed the renesas port was still using the older polling pattern for this.

A quick test on a  "EK-RA4W1 with usb cable patched on" I've got showed that the interrupt wrap approach appears to work correctly here, so submitted this change.

### Testing

So far just flashing this change to "EK-RA4W1 with usb cable patched on" and tested that repl is still available in mpremote 

The board I have uses the renesas full-speed peripheral, this change has not been tested on renesas board with high-speed usb.

### DIFF
--- a/ports/renesas-ra/Makefile
+++ b/ports/renesas-ra/Makefile
@@ -157,6 +157,9 @@ LIBSTDCPP_FILE_NAME = "$(shell $(CXX) $(CXXFLAGS) -print-file-name=libstdc++.a)"
 LDFLAGS += -L"$(shell dirname $(LIBSTDCPP_FILE_NAME))"
 endif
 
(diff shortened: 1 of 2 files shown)


iabdalkader:
@andrewleech I finally had a chance to test this and USB is working fine with this change.

andrewleech:
That's great thanks @iabdalkader

dpgeorge:
Thanks @andrewleech, it's good to make all TinyUSB ports use the same mechanism.

And thanks to @iabdalkader for testing.

PR #15553, opened 2024-07-26
nrf/Makefile: Fix GCC_VERION regex.

andrewleech:
See https://github.com/micropython/micropython/pull/15158#discussion_r1692520901


agatti (commented):
Instead of using a regex and invoking sed, how about something like `$(shell $(CC) -dumpversion)`?  It's been available in GCC since pretty much forever as far as I know.

andrewleech:
> how about something like $(shell $(CC) -dumpversion)

Oh that's an easy yes! Thanks, not sure why I couldn't find that when I searched initially. Stack overflow let me down ;-)

Cheers, I'll update this shortly as such.

Issue #15551, opened 2024-07-26
RV32 native emitter fails compiling some of the tests


dpgeorge:
### Port, board and/or hardware

Any RISC-V 32 architecture

### MicroPython version

Current master version (commit 6007f3e2062cc65fc8416f241c682e37eb956c11)

### Reproduction

Using an ESP32C3 board, run the test suite using the following:
```
$ cd tests
$ ./run-tests.py --target esp32 --device /dev/ttyUSB0 --via-mpy --emit native --mpy-cross-flags="-march=rv32imc" -d basics
```
The output is:
```
518 tests performed (16146 individual testcases)
488 tests passed
18 tests skipped: annotate_var builtin_next_arg2 builtin_range_binop del_deref del_local exception_chain gen_yield_from_close memoryview_itemsize namedtuple_asdict nanbox_smallint scope_implicit subclass_native_call sys_path sys_tracebacklimit try_finally_return2 try_reraise try_reraise2 unboundlocal
30 tests failed: array1 array_construct array_construct_endian array_intbig builtin_property class_ordereddict class_super dict1 dict_fixed errno1 frozenset_binop fun_calldblstar3 gc1 ifcond int1 logic_constfolding memoryview_intbig namedtuple1 set_binop stopiteration string_format_modulo string_format_modulo_int struct1 struct1_intbig struct2 struct_micropython sys_getsizeof true_value try_finally_return unpack1
```
The test failures are due to `mpy-cross` crashing:
```
mpy-cross: ../py/asmbase.c:94: mp_asm_base_label_assign: Assertion `as->label_offsets[label] == as->code_offset' failed.
```

Alternatively, a much simpler way to see the problem:
```
$ ./mpy-cross/build/mpy-cross -march=rv32imc -X emit=native tests/basics/array1.py 
mpy-cross: ../py/asmbase.c:94: mp_asm_base_label_assign: Assertion `as->label_offsets[label] == as->code_offset' failed.
```

### Expected behaviour

`mpy-cross` should be able to compile all the tests with the rv32imc emitter selected.

### Observed behaviour

`mpy-cross` crashes with an assertion failure.

### Code of Conduct

Yes, I agree

dpgeorge:
@agatti you may be interested to look into this issue.

agatti:
Thanks for bringing this to my attention, I'll look into this right away.

[...]

PR #15573, opened 2024-07-31
py/emitnative: Fix case of clobbered REG_TEMP0 when loading const obj.

dpgeorge:
### Summary

The `emit_load_reg_with_object()` helper function will clobber `REG_TEMP0`. This is currently OK on architectures where `REG_RET` and `REG_TEMP0` are the same (all architectures except RV32), because all callers of `emit_load_reg_with_object()` use either `REG_RET` or `REG_TEMP0` as the destination register.  But on RV32 these registers are different and so when `REG_RET` is the destination, `REG_TEMP0` is clobbered, leading to incorrectly generated machine code.

This commit fixes the issue simply by using `REG_TEMP0` as the destination register for all uses of `emit_load_reg_with_object()`, and adds a comment to make sure the caller of this function is careful.

### Testing

This is currently tested only by inspecting output of `mpy-cross -march=debug` for the following code:
```py
import array

@micropython.native
def a():
        array.array('B', b'12')

a()
```
See https://github.com/micropython/micropython/issues/15551#issuecomment-2259068640 for background.

TODO: would be good to improve the qemu-riscv (and qemu-arm) ports so they can test this case by generating .mpy files using mpy-cross with a native emitter for all their tests, and running the .mpy.  But that requires a bit of upfront leg work.

### DIFF
--- a/py/emitnative.c
+++ b/py/emitnative.c
@@ -357,6 +357,7 @@ static void emit_native_mov_reg_qstr(emit_t *emit, int arg_reg, qstr qst) {
     #endif
 }
 
+// This function may clobber REG_TEMP0 (and `reg_dest` can be REG_TEMP0).
 static void emit_native_mov_reg_qstr_obj(emit_t *emit, int reg_dest, qstr qst) {
     #if MICROPY_PERSISTENT_CODE_SAVE
     emit_load_reg_with_object(emit, reg_dest, MP_OBJ_NEW_QSTR(qst));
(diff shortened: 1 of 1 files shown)


jimmo (approved):
> TODO: would be good to improve the qemu-riscv (and qemu-arm) ports so they can test this case by generating .mpy files using mpy-cross with a native emitter for all their tests, and running the .mpy. But that requires a bit of upfront leg work.

Can you raise an issue describing this in more detail? (I'm not quite sure from the description above what needs to be done).

This change looks good otherwise, although given that this is not the first example of this sort of clobbering it would be good to know whether the suggested improvement to the qemu-testing is likely to be able to catch other cases?

dpgeorge:
> Can you raise an issue describing this in more detail?

See #15609 which sets the stage to run native .mpy tests on the qemu-* ports.

> it would be good to know whether the suggested improvement to the qemu-testing is likely to be able to catch other cases?

We currently only run full native tests under CI for the x86 and x64 emitters.  Enabling all the native tests under CI using qemu-arm and qemu-riscv, it's highly likely that would expose some of the bugs that we already found and fixed (which were found by running native tests on hardware, mostly ARM Thumb and Xtensa hardware).

PR #15575, opened 2024-07-31
py/asmrv32: Fix RV32 extended test failures.

agatti:
### Summary

This PR contains fixes for the native emitter test failures uncovered in #15551.  These fixes depend on #15573 to be merged in to properly fix test some test failures.

### Testing

Ran the whole `esp32` test suite via `mpy-cross` on an ESP32C3 board with `./run-tests.py --target esp32 --device /dev/ttyUSB0 --via-mpy --emit native --mpy-cross-flags="-march=rv32imc"`.

### DIFF
--- a/py/asmrv32.c
+++ b/py/asmrv32.c
@@ -312,8 +312,8 @@ void asm_rv32_emit_call_ind(asm_rv32_t *state, mp_uint_t index) {
     if (FIT_UNSIGNED(offset, 11)) {
         // lw     temporary, offset(fun_table)
         // c.jalr temporary
-        asm_rv32_opcode_lw(state, INTERNAL_TEMPORARY, REG_FUN_TABLE, offset);
-        asm_rv32_opcode_cjalr(state, INTERNAL_TEMPORARY);
+        asm_rv32_opcode_lw(state, REG_TEMP2, REG_FUN_TABLE, offset);
+        asm_rv32_opcode_cjalr(state, REG_TEMP2);
         return;
     }
 
@@ -321,16 +321,14 @@ void asm_rv32_emit_call_ind(asm_rv32_t *state, mp_uint_t index) {
     mp_uint_t lower = 0;
     split_immediate(offset, &upper, &lower);
 
-    // TODO: Can this clobber REG_TEMP[0:2]?
-
     // lui    temporary, HI(index) ; Or c.lui if possible
     // c.add  temporary, fun_table
     // lw     temporary, LO(index)(temporary)
     // c.jalr temporary
-    load_upper_immediate(state, INTERNAL_TEMPORARY, upper);
-    asm_rv32_opcode_cadd(state, INTERNAL_TEMPORARY, REG_FUN_TABLE);
-    asm_rv32_opcode_lw(state, INTERNAL_TEMPORARY, INTERNAL_TEMPORARY, lower);
-    asm_rv32_opcode_cjalr(state, INTERNAL_TEMPORARY);
+    load_upper_immediate(state, REG_TEMP2, upper);
+    asm_rv32_opcode_cadd(state, REG_TEMP2, REG_FUN_TABLE);
+    asm_rv32_opcode_lw(state, REG_TEMP2, REG_TEMP2, lower);
+    asm_rv32_opcode_cjalr(state, REG_TEMP2);
 }
 
 void asm_rv32_emit_jump_if_reg_eq(asm_rv32_t *state, mp_uint_t rs1, mp_uint_t rs2, mp_uint_t label) {
@@ -350,15 +348,13 @@ void asm_rv32_emit_jump_if_reg_eq(asm_rv32_t *state, mp_uint_t rs1, mp_uint_t rs
     mp_uint_t lower = 0;
     split_immediate(displacement, &upper, &lower);
 
-    // TODO: Can this clobber REG_TEMP[0:2]?
-
     // bne   rs1, rs2, 12                      ; PC + 0
     // auipc temporary, HI(displacement)       ; PC + 4
     // jalr  zero, temporary, LO(displacement) ; PC + 8
     // ...                                     ; PC + 12
     asm_rv32_opcode_bne(state, rs1, rs2, 12);
-    asm_rv32_opcode_auipc(state, INTERNAL_TEMPORARY, upper);
-    asm_rv32_opcode_jalr(state, ASM_RV32_REG_ZERO, INTERNAL_TEMPORARY, lower);
+    asm_rv32_opcode_auipc(state, REG_TEMP2, upper);
+    asm_rv32_opcode_jalr(state, ASM_RV32_REG_ZERO, REG_TEMP2, lower);
 }
 
 void asm_rv32_emit_jump_if_reg_nonzero(asm_rv32_t *state, mp_uint_t rs, mp_uint_t label) {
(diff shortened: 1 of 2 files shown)


py/asmrv32.c
```diff
@@ -326,10 +326,11 @@ void asm_rv32_emit_call_ind(asm_rv32_t *state, mp_uint_t index) {
 }
 
 void asm_rv32_emit_jump_if_reg_eq(asm_rv32_t *state, mp_uint_t rs1, mp_uint_t rs2, mp_uint_t label) {
-    ptrdiff_t displacement = (ptrdiff_t)(state->base.label_offsets[label] - state->base.code_offset);
+    mp_uint_t label_offset = state->base.label_offsets[label];
+    ptrdiff_t displacement = (ptrdiff_t)(label_offset - state->base.code_offset);
+    bool emit_short_jump = (label_offset != (mp_uint_t)-1) && (displacement < 0);
```
dpgeorge, line 331: I suggest making a local (static inline?) function that does these three lines of computation, and reused, to reduce code duplication.

agatti: Good point!  I'll update my local copy and push this change along with fixes for the other failing tests, thanks.

py/asmrv32.c
```diff
@@ -270,6 +270,15 @@ static void emit_function_epilogue(asm_rv32_t *state, mp_uint_t registers) {
     state->saved_registers_mask = old_saved_registers_mask;
 }
 
+static void calculate_displacement_for_label(asm_rv32_t *state, mp_uint_t label, ptrdiff_t *displacement, bool *emit_short_jump) {
```
dpgeorge, line 273: I suggest making this return `bool` for the `emit_short_jump` variable, instead of taking that as a pointer-to-return.  That makes the code a little neater, and maybe more efficient because it's not taking the address of a local variable.

Also, I suggest calling the variable `can_emit_short_jump` because it's not a requirement to emit a short jump, just a hint that it's possible.

py/asmrv32.c
```diff
@@ -354,31 +364,25 @@ void asm_rv32_emit_jump_if_reg_eq(asm_rv32_t *state, mp_uint_t rs1, mp_uint_t rs
 }
 
 void asm_rv32_emit_jump_if_reg_nonzero(asm_rv32_t *state, mp_uint_t rs, mp_uint_t label) {
-    ptrdiff_t displacement = (ptrdiff_t)(state->base.label_offsets[label] - state->base.code_offset);
+    ptrdiff_t displacement = 0;
+    bool emit_short_jump = false;
+    calculate_displacement_for_label(state, label, &displacement, &emit_short_jump);
 
-    if (FIT_SIGNED(displacement, 9) && IS_IN_C_REGISTER_WINDOW(rs)) {
+    if (FIT_SIGNED(displacement, 8) && emit_short_jump && IS_IN_C_REGISTER_WINDOW(rs)) {
```
dpgeorge, line 371: I suggest to put the `emit_short_jump` (or `can_emit_short_jump`) first in the if-logic, because it's the most important thing to check first, whether this optimisation can be done or not.  It'll also be a little more efficient when executing.

py/asmrv32.c
```diff
@@ -445,12 +458,14 @@ void asm_rv32_emit_mov_reg_local(asm_rv32_t *state, mp_uint_t rd, mp_uint_t loca
...
-    // lw    rd, LO(offset)(rd)
-    load_upper_immediate(state, rd, upper);
-    asm_rv32_opcode_cadd(state, rd, ASM_RV32_REG_SP);
-    asm_rv32_opcode_lw(state, rd, rd, lower);
+    // TODO: Can this clobber REG_TEMP[0:2]?
+
+    // lui   temporary, HI(offset) ; Or c.lui if possible
+    // c.add temporary, sp
+    // lw    rd, LO(offset)(temporary)
+    load_upper_immediate(state, INTERNAL_TEMPORARY, upper);
```
dpgeorge, line 466: Why did this change to use `INTERNAL_TEMPORARY` instead of `rd`?  It seems like using `rd` is OK and then it doesn't require an additional register.

agatti: Too eager when copying and pasting I guess.  `INTERNAL_TEMPORARY` is always saved and restored as part of the function prologue/epilogue and its clobbering is always function-scoped so it's safe.  I'll fix that to use `rd` though, thanks.

py/asmrv32.c
agatti: I've implemented your suggestion, but the compiler emitted the very same code when optimisations were enabled as `can_emit_short_jump` is the least computationally-intensive thing to check for if you need to invalidate the whole expression (a non-volatile boolean), so it was checked first anyway.

py/asmrv32.c
agatti: The function not returning a `bool` was deliberate, as it's more difficult to forget about the second returned value that way, since it is forced to be declared for the code to compile (and the compiler would reorder everything anyway when optimising things - there are *a lot* of unused registers on RV32 after all).  I agree with the variable renaming, it is definitely clearer that way!

That said, I've implemented your suggestions, thanks for taking a look.

agatti:
This PR now contains fixes for all RV32-specific test failures found in #15551, it's now marked as ready for review.

py/asmrv32.c
dpgeorge: Oh, that's interesting that the compiler swaps the operands of the short-circuiting operator.  I guess it can do that if it's all local variables and arithmetic (no function calls).

Anyway, thanks for updating it.

dpgeorge:
I tested this PR on a RP2350 in RISC-V mode and it works well, all `basic`, `extmod` and `float` tests now pass (they didn't before):
```
$ ./run-tests.py --target rp2 --via-mpy --emit native --mpy-cross-flags="-march=rv32imc" -d basics extmod float
...
706 tests performed (22699 individual testcases)
706 tests passed
66 tests skipped
```

py/asmrv32.c
agatti: There are other places where the temporary register is used, like in `asm_rv32_emit_jump` and `asm_rv32_emit_call_ind` for example.

If I have to pick another register to replace that with, I'd say I could use REG_TEMP2 instead.  I understand that right now is only used inside `emit_native_rot_three` and in viper load/stores for RV32, but I wonder if there are any chances of that register being used elsewhere in the future.  If that's OK for me to use then this may allow few more size optimisations, as `S0` is inside the compressed opcodes register window.

dpgeorge: If using `INTERNAL_TEMPORARY` instead of `rd` allows for more size optimisations, then that's a good reason to use `INTERNAL_TEMPORARY`.

agatti: I've looked a bit more into this to see if by shuffling things around there are more avenues for size optimisation.

Something that can be done (not in this PR, of course), would be to free up the register taken by `INTERNAL_TEMPORARY` (`S0`) and map it to `REG_TEMP0`.  Whilst this will still require a push/pop of the register in question (S-registers have to be saved and restored in this case), it would allow usage of compressed opcodes like `C.SUB`, `C.AND`, `C.OR` and `C.XOR`, and make it viable to use `C.SW` in more places.  There would be a penalty when calling out to the MicroPython runtime as REG_TEMP0 would probably be already clobbered when calling `asm_rv32_emit_call_ind`.

I'd need to come up with some numbers to see if such a thing is justified, though.

Anyway, I'll change the code to use `rd` and `REG_TEMP2` whenever it won't impact on code size.  Hopefully there's a way to make this bit of code better than it is now.

agatti:
I've cleaned the PR up a bit, and did the same successful test verification you did but on an ESP32C3.

dpgeorge:
This looks really good now, thank you.  I also reconfirmed the latest changes on RP2350-RISCV.

PR #15577, opened 2024-08-01
qemu-arm/qemu-riscv: consolidate Makefile.test into main Makefile

dpgeorge:
### Summary

Currently, the `qemu-arm` and `qemu-riscv` ports have separate `Makefile.test` files needed to build and run the tests.  It's a bit awkward (and unconventional within this project) to have to run `make -f Makefile.test test` to run the tests.

The files don't really need to be separate, and this PR consolidates `Makefile.test` into the main `Makefile`.

Now the tests are run simply with `make test`.

### Testing

Locally ran `make test` and it had the same output as before doing `make -f Makefile.test test`.

CI is also updated to work with this new, simplified make target.

The alternative is to leave it as-is, but I think this change makes it simpler, easier to understand, and more like other ports (which only have one Makefile).


agatti:
Out of curiosity, wouldn't the `powerpc` port need to be renamed to `qemu-powerpc` and have similar build machinery set up to run tests?

dpgeorge:
> wouldn't the `powerpc` port need to be renamed to `qemu-powerpc`

The `powerpc` port was intended to be a proper, generic PowerPC port.  But it never really got off the ground.

dpgeorge:
Closing in favour of #15624.

PR #15583, opened 2024-08-02
esp32: Use configured stack-limit-margin when creating threads.

dpgeorge:
### Summary

This gives more stack margin for threads on ESP32C3 (it should have been done this way from the beginning).  That's important for IDF v5.2, see https://github.com/micropython/micropython/pull/15523#issuecomment-2264356843

### Testing

Ran the main test suite on an ESP32C3 using IDF v5.0.5.  All tests pass (sans known failure `math_domain_special.py`).

### DIFF
--- a/ports/esp32/mpthreadport.c
+++ b/ports/esp32/mpthreadport.c
@@ -38,7 +38,7 @@
 #if MICROPY_PY_THREAD
 
 #define MP_THREAD_MIN_STACK_SIZE                        (4 * 1024)
-#define MP_THREAD_DEFAULT_STACK_SIZE                    (MP_THREAD_MIN_STACK_SIZE + 1024)
+#define MP_THREAD_DEFAULT_STACK_SIZE                    (MP_THREAD_MIN_STACK_SIZE + MP_TASK_STACK_LIMIT_MARGIN)
 #define MP_THREAD_PRIORITY                              (ESP_TASK_PRIO_MIN + 1)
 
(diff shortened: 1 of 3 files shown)


ports/esp32/mpthreadport.c
```diff
@@ -161,7 +161,7 @@ mp_uint_t mp_thread_create_ex(void *(*entry)(void *), void *arg, size_t *stack_s
     thread = th;
 
     // adjust the stack_size to provide room to recover from hitting the limit
-    *stack_size -= 1024;
+    *stack_size -= MP_TASK_STACK_LIMIT_MARGIN;
```
projectgus, line 164: This code has a pre-existing race condition, because stack_size points to inside the thread 'arg' structure (and the thread is already created. If a context switch happens before here then it will set the limit to the full task stack size.

It still failed for me on IDF V5.2.2 until I moved the adjustment before creating the RTOS thread. However I think I found a simpler way to handle it, will submit PR after lunch.

(As an aside, I thought I saw the stress_recurse test failing intermittently on esp32 - maybe this is why.)

dpgeorge: Great find!  I'll wait for your fix then.

projectgus: (Alternative now at https://github.com/micropython/micropython/pull/15605 )

dpgeorge:
Closing in favour of #15605.

PR #15585, opened 2024-08-02
ports/rp2: Add i2s deinit on soft reset.

### LINKED ISSUE #14339: [RP2] machine.soft_reset() hangs when i2s.deinit() is not called

Gadgetoid:
### Summary

Code using i2s required a try/finally in order to avoid a hard fault on soft reset.

Add `machine_i2s_deinit_all` to teardown any active i2s instances on reset.

Fixes #14339

Probably replaces #14345

### Testing

I've "fixed" and tested the RP2 port, using our TinyFX/picofx libraries and confirmed that subsequent to this change, forgetting to explicitly `deinit()` your i2s instance will not result in a burst of static and a hardfault.

As far as I'm aware, this bring i2s in line with other interfaces such that they're torn down automatically on soft reset.

I don't believe these are any cases where you wouldn't want to do this, but obviously doing so automatically comes at a small code size cost.

It's arguable that this should be done (as mentioned in #14339) with a finalizer, but there are some issues around that topic which would take much longer to work out than simply adding this method.

Even executing all finalizers first does not guarantee they happen in the right order, so since we're using `MP_STATE_PORT` to track active i2s instances, we might as well be nice and clean them up first.

### DIFF
--- a/ports/rp2/machine_i2s.c
+++ b/ports/rp2/machine_i2s.c
@@ -171,6 +171,15 @@ void machine_i2s_init0(void) {
     }
 }
 
+void machine_i2s_deinit_all(void) {
+    for (uint8_t i = 0; i < MAX_I2S_RP2; i++) {
+        machine_i2s_obj_t *i2s = MP_STATE_PORT(machine_i2s_obj[i]);
+        if (i2s) {
+            mp_machine_i2s_deinit(i2s);
+        }
+    }
+}
+
 static int8_t get_frame_mapping_index(int8_t bits, format_t format) {
     if (format == MONO) {
         if (bits == 16) {
--- a/ports/rp2/main.c
+++ b/ports/rp2/main.c
@@ -214,6 +214,7 @@ int main(int argc, char **argv) {
         #if MICROPY_PY_NETWORK
         mod_network_deinit();
         #endif
+        machine_i2s_deinit_all();
         rp2_dma_deinit();
         rp2_pio_deinit();
         #if MICROPY_PY_BLUETOOTH
--- a/ports/rp2/modmachine.h
+++ b/ports/rp2/modmachine.h
@@ -6,6 +6,7 @@
 void machine_pin_init(void);
 void machine_pin_deinit(void);
 void machine_i2s_init0(void);
+void machine_i2s_deinit_all(void);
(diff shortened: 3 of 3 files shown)


dpgeorge:
> It's arguable that this should be done (as mentioned in #14339) with a finalizer, but there are some issues around that topic which would take much longer to work out than simply adding this method.

I like your pragmatism here!

> Even executing all finalizers first does not guarantee they happen in the right order, so since we're using `MP_STATE_PORT` to track active i2s instances, we might as well be nice and clean them up first.

So, the mimxrt port uses a `machine_i2s_deinit_all()` function like added in this PR.  But the esp32 port uses finalisers.  stm32 is the only other port that supports I2S and it does not do any clean up (it should though).

I was originally thinking that finalisers are the better option (eg files and sockets rely heavily on finalisers working properly).  But in the case of peripherals like I2S that have their instances tracked in explicit root pointers, finalisers actually don't work as expected: because there's a perpetual root pointer holding the instance, the GC will never reclaim it even if the user no longer has a pointer to the instance in a Python variable.  The finaliser will only run during a call to `gc_sweep_all()`, which is good but not what I expected to happen.

Therefore, I think explicitly managing the deinit (what's done in this PR) is a good approach.  In particular, the finaliser approach doesn't add anything extra (it doesn't help clean up I2S instances during runtime of an application).

So +1 from me.

dpgeorge:
Rebased and merged in d2e33fe3096eec60d7017b0f17c4ddb6910a4d0b

Issue #15571, opened 2024-07-30
frambuf initalizes with ValueError


hoihu:
### Port, board and/or hardware

all variants that use the framebuf module

### MicroPython version

MicroPython v1.24.0-preview.149.g6007f3e20 on 2024-07-26; Raspberry Pi Pico W with RP2040

### Reproduction

I'm using a simple frambuffer with one bit per pixel. This code segment used to work on earlier micropython versions but fails now: 

```
import framebuf
w=21; h=8
data = bytearray(w*h // 8)
fb = framebuf.FrameBuffer(data, w, h, framebuf.MONO_HMSB)
```

### Expected behaviour

Expected to accept the input buffer without a `ValueError`

### Observed behaviour

this causes a 

```
 File "<stdin>", line 1, in <module>
ValueError: 
```

ValueError is raised with no further details

### Additional Information

I tracked this due to this input validation here:

https://github.com/micropython/micropython/blob/master/extmod/modframebuf.c#L318

and I think that's due to 
https://github.com/micropython/micropython/blob/master/extmod/modframebuf.c#L294

which modifies the `stride` value to be 24 (`(21 + 7) & ~7 == 24`) and therefore the provided input buffer is too small.  Modifying the example above with a input buffer of 24 then works as expected.

I'm not sure why this check was introduced. Maybe due to other functionalities in the framebuffer. But in any case it's now difficult to provide a matching input buffer size without the knowledge of what is checked inside the module.

This could be addressed via a doc update. Or raise a ValueError that indicates what went wrong (e.g. "expected buffer length of xxx, got yyy")

### Code of Conduct

Yes, I agree

hoihu:
seems to be introduced by https://github.com/micropython/micropython/commit/d040478d8aad7a05aac98651ec6ac5c2e9265a22 which solved https://github.com/micropython/micropython/issues/12562

Gadgetoid:
```
>>> 21/8
2.625
>>> ((21+7) & ~7)/8
3.0
```

The code mentioned seems to make `stride` a multiple of 8, presumably because `mono_horiz_setpixel` couldn't handle pixels wrapping on the Y axis to an adjacent byte.

21 is not a multiple of 8, so if you had a buffer 21x2 you would get:

```
111111111111111111111222
222222222222222222xxxxxx
```

Where 1 and 2 are rows 1 and 2 (Y axis) respectively. (might be that X/Y are swapped by its academic.

[...]

PR #15587, opened 2024-08-02
docs/framebuf: Add minimal buffer size requirements.

hoihu:
### Summary

To improve the framebuf documentation regarding buffer size requirements

see https://github.com/micropython/micropython/issues/15571

### Testing

only doc updates

none

### DIFF
--- a/docs/library/framebuf.rst
+++ b/docs/library/framebuf.rst
@@ -53,6 +53,9 @@ Constructors
     optionally *stride*.  Invalid *buffer* size or dimensions may lead to
     unexpected errors.
 
+    Buffer size depends on the pixel format. See below for additional details.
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the patch, and sorry this sat for ages without getting a review.

Commit d1574de3b6a050ef07c39a131203305f731d39d8 -- fixing issue #15944 -- actually made the calculation for the size more complicated!  So the docs here would need to be adjusted to take that into account.

Maybe it's easier to just say something like "the backing buffer must have enough bytes to store the requested width/height/stride and a ValueError will be raised if its too small"?

dpgeorge:
Closing due to inactivity.  Feel free to reopen if you want to work on it.

Issue #15580, opened 2024-08-01
examples/network/https_client.py + cloudflare.com = OSError: 97


CardealRusso:
### Port, board and/or hardware

Unix

### Reproduction

https://raw.githubusercontent.com/micropython/micropython/master/examples/network/https_client.py

If i use cloudflare.com in place of google.com o get OSError: 97 Any tips?

dpgeorge:
It's because the `getaddrinfo()` is returning an IPv6 address as the first one.

To force IPv4 use:
```
ai = socket.getaddrinfo("cloudflare.com", 443, socket.AF_INET)
```

CardealRusso:
thanks

dpgeorge:
I've improved the examples in #15588, so they work with both IPv4 and IPv6.

PR #15588, opened 2024-08-03
examples/network: Improve HTTP client examples to work with IPv4 and IPv6, and add more comments

dpgeorge:
### Summary

The main changes here are to pass the address family and socket type to `getaddrinfo()`, and then use the result of the address lookup when creating the socket, so it has the correct address family.
    
This allows both IPv4 and IPv6 to work, because the socket is created with the correct AF_INETx type for the address.
    
Also add some more comments to the examples to explain what's going on.

And use `SSLContext` instead of `ssl.wrap_socket()` because the latter is deprecated (in CPython at least).
    
Fixes issue #15580.

### Testing

Tested on CPython (only `http_client.py` and `https_client.py`), MicroPython unix and Pico W.

### DIFF
--- a/examples/network/http_client.py
+++ b/examples/network/http_client.py
@@ -1,16 +1,35 @@
+# Very simple HTTP client example:
+# - Connects to a server.
+# - Sends a HTTP request.
+# - Reads the result.
+#
+# This example works in both MicroPython and CPython.
+#
+# To implement an HTTP client using less code, use mip to install the requests package:
+# https://github.com/micropython/micropython-lib/tree/master/python-ecosys/requests
+
 import socket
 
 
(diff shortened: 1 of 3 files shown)


projectgus (approved):
These changes look good!

The only other thing that I'd suggest is to add an extra comment along these lines to both examples:

```py
# To implement an HTTP client using less code, use mip to install the requests package:
# https://github.com/micropython/micropython-lib/tree/master/python-ecosys/requests
```

dpgeorge:
Thanks for the review.  I've added that comment to the HTTP and HTTPS client examples (not the non-blocking one, because that doesn't generalise to requests).

PR #15589, opened 2024-08-03
esp32/main: Store native code as linked list instead of list on GC heap.

dpgeorge:
### Summary

Finalisers that run during `gc_sweep_all()` may run native code, for example if an open file is closed and the underlying block device is implemented in native code, then the filesystem driver (eg FAT) may call into the native code.

Therefore, native code must be freed after the call to `gc_sweep_all()`. That can only be achieved if the GC heap is not used to store the list of allocated native code blocks.  Instead, this commit makes the native code blocks a linked list.

### Testing

Tested on an ESP32C3, running the native tests as per #15551:
```
./run-tests.py --target esp32 --device /dev/ttyUSB0 --via-mpy --emit native --mpy-cross-flags="-march=rv32imc" -d basics extmod
```

All the tests pass.

Note: requires #15573, #15575.

This approach uses 4 bytes extra iRAM per native function, but keeps the code simple and understandable.  An alternative would be to still use the GC heap to store the list of native memory blocks, and make `gc_sweep_all()` only call the finalisers, not actually reclaim/clear memory.  That's a complex change to make though.

### DIFF
--- a/ports/esp32/main.c
+++ b/ports/esp32/main.c
@@ -78,6 +78,15 @@
 #define MP_TASK_STACK_LIMIT_MARGIN (1024)
 #endif
 
+typedef struct _native_code_node_t {
+    struct _native_code_node_t *next;
+    uint32_t data[];
+} native_code_node_t;
+
+static native_code_node_t *native_code_head = NULL;
+
+static void esp_native_code_free_all(void);
+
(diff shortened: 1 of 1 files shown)


projectgus (approved):
This approach looks good to me!

PR #8304, opened 2022-02-14
stm32/powerctrl: Add USB support to lightsleep.

andrewleech:
Currently on stm32, if usb cdc repl is in use when `machine.lightsleep()` is run, the reply locks up and never recovers.
The micro itself will wake up and continue on with the code, but no data can be read/written to USB.

The PC thinks the USB is still connected, however no data is transferred. 

If the USB is unplugged / replugged the connection recovers - assuming the device is self-powered it continues working with a normal re-connection of the USB, a reboot is not required.

This PR fixes the issue by explicitely disconnecting / reconnecting the USB in code before and after the STOP call.

While it would be nicer to not need the PC connection to disconnect / reconnect I'm not sure this is possible.

PR #15592, opened 2024-08-03
stm32/usb: Add support for using TinyUSB stack.

andrewleech:
### Summary

This PR adapts the stm32 port to allow switching from STM USB stack to TinyUSB stack. 

Using TinyUSB improves consistancy with other micropython ports and brings in the ability to use the runtime usb definition support recently added to other TinyUSB based ports.

TinyUSB can be enabled in a build with `#define MICROPY_HW_TINYUSB_STACK (1)` 
It's also enabled by default with `MICROPY_PREVIEW_VERSION_2`

From the command line it can be enabled with:
``` bash
make -C ports/stm32 BOARD=PYBV11 CFLAGS_EXTRA='-DMICROPY_HW_TINYUSB_STACK=1'
```

### Testing

Development and testing so far has been on the STM32WB55 and STM32F765

I've been more recently using and testing this in conjunction with [shared/tinyusb: Add support for USB Network (NCM) interface. #16459](https://github.com/micropython/micropython/pull/16459) on the STM32F765 where the builtin CDC and NCM endpoints are both working very well.

I haven't yet done significant testing of dynamic / runtime usb devices.

Currently the TinyUSB integration matches other ports for features, so CDC is enabled for repl use and [machine.USBDevice](https://docs.micropython.org/en/latest/library/machine.USBDevice.html) support is available by default.

This means however the existing stm32 / pyb usb support has been removed, including [pyb.USB_VCP](https://docs.micropython.org/en/latest/library/pyb.USB_VCP.html) and [pyb.usb_mode](https://docs.micropython.org/en/latest/library/pyb.html#pyb.usb_mode). 

Features like mass storage, hid and a second vcp could in theory be added via `machine.USBDevice` however this hasn't been tested and is a breaking change for the stm32 port.
 
TODO: test lightsleep behaviour / compared with https://github.com/micropython/micropython/pull/8304#issuecomment-2303453613

### DIFF
--- a/ports/stm32/Makefile
+++ b/ports/stm32/Makefile
@@ -256,6 +256,7 @@ SRC_C += \
 	usbd_hid_interface.c \
 	usbd_msc_interface.c \
 	mphalport.c \
+	msc_disk.c \
 	mpnetworkport.c \
 	mpthreadport.c \
 	irq.c \
--- a/ports/stm32/Makefile
+++ b/ports/stm32/Makefile
@@ -58,7 +58,7 @@ MBOOT_TEXT0_ADDR ?= 0x08000000
 include $(TOP)/py/py.mk
 include $(TOP)/extmod/extmod.mk
 
-GIT_SUBMODULES += lib/libhydrogen lib/stm32lib
+GIT_SUBMODULES += lib/libhydrogen lib/stm32lib lib/tinyusb
 
 CROSS_COMPILE ?= arm-none-eabi-
 LD_DIR=boards
@@ -110,6 +110,8 @@ INC += -I$(STM32LIB_CMSIS_ABS)/Include
 INC += -I$(STM32LIB_HAL_ABS)/Inc
 INC += -I$(USBDEV_DIR)/core/inc -I$(USBDEV_DIR)/class/inc
 #INC += -I$(USBHOST_DIR)
+INC += -I$(TOP)/lib/tinyusb/src
+INC += -I$(TOP)/shared/tinyusb/
 INC += -Ilwip_inc
 
 CFLAGS += $(INC) -Wall -Wpointer-arith -Werror -Wdouble-promotion -Wfloat-conversion -std=gnu99 -nostdlib $(CFLAGS_EXTRA)
@@ -213,6 +215,10 @@ SHARED_SRC_C += $(addprefix shared/,\
 	runtime/stdout_helpers.c \
 	runtime/sys_stdio_mphal.c \
 	timeutils/timeutils.c \
+	tinyusb/mp_usbd.c \
+	tinyusb/mp_usbd_cdc.c \
+	tinyusb/mp_usbd_descriptor.c \
+	tinyusb/mp_usbd_runtime.c \
 	)
 
 ifeq ($(MICROPY_FLOAT_IMPL),double)
@@ -242,7 +248,9 @@ SRC_C += \
 	boardctrl.c \
 	main.c \
 	stm32_it.c \
+	usbd.c \
 	usbd_conf.c \
+	usb.c \
 	usbd_desc.c \
 	usbd_cdc_interface.c \
 	usbd_hid_interface.c \
@@ -277,7 +285,6 @@ SRC_C += \
 	can.c \
 	fdcan.c \
 	pyb_can.c \
-	usb.c \
 	eth.c \
 	eth_phy.c \
 	gccollect.c \
@@ -460,6 +467,16 @@ USBDEV_SRC_C += $(addprefix $(USBDEV_DIR)/,\
 	class/src/usbd_msc_scsi.c \
 	)
 
+# TinyUSB Stack source
+-include $(TOP)/lib/tinyusb/src/tinyusb.mk
+TINYUSB_SRC_C := $(addprefix lib/tinyusb/, \
+	$(TINYUSB_SRC_C) \
+	src/portable/st/stm32_fsdev/dcd_stm32_fsdev.c \
+	src/portable/synopsys/dwc2/dcd_dwc2.c \
+	src/portable/synopsys/dwc2/dwc2_common.c \
+	src/portable/synopsys/dwc2/hcd_dwc2.c \
+	)
+
 ifeq ($(MICROPY_SSL_MBEDTLS),1)
 LIB_SRC_C += mbedtls/mbedtls_port.c
 endif
@@ -499,6 +516,7 @@ OBJ += $(addprefix $(BUILD)/, $(SHARED_SRC_C:.c=.o))
 OBJ += $(addprefix $(BUILD)/, $(DRIVERS_SRC_C:.c=.o))
 OBJ += $(addprefix $(BUILD)/, $(HAL_SRC_C:.c=.o))
 OBJ += $(addprefix $(BUILD)/, $(USBDEV_SRC_C:.c=.o))
+OBJ += $(addprefix $(BUILD)/, $(TINYUSB_SRC_C:.c=.o))
 OBJ += $(addprefix $(BUILD)/, $(SRC_C:.c=.o))
 OBJ += $(addprefix $(BUILD)/, $(SRC_CXX:.cpp=.o))
 OBJ += $(GEN_PINS_SRC:.c=.o)
--- a/ports/stm32/main.c
+++ b/ports/stm32/main.c
@@ -280,7 +280,7 @@ static bool init_sdcard_fs(void) {
                 }
             }
 
-            #if MICROPY_HW_ENABLE_USB && !MICROPY_HW_TINYUSB_STACK
+            #if MICROPY_HW_USB_MSC
             if (pyb_usb_storage_medium == PYB_USB_STORAGE_MEDIUM_NONE) {
                 // if no USB MSC medium is selected then use the SD card
                 pyb_usb_storage_medium = PYB_USB_STORAGE_MEDIUM_SDCARD;
@@ -639,7 +639,7 @@ soft_reset:
     }
     #endif
 
-    #if MICROPY_HW_STM_USB_STACK
+    #if MICROPY_HW_USB_MSC
     // if the SD card isn't used as the USB MSC medium then use the internal flash
     if (pyb_usb_storage_medium == PYB_USB_STORAGE_MEDIUM_NONE) {
         pyb_usb_storage_medium = PYB_USB_STORAGE_MEDIUM_FLASH;
--- a/ports/stm32/main.c
+++ b/ports/stm32/main.c
@@ -88,6 +88,11 @@
 #include "pyb_can.h"
 #include "subghz.h"
 
+#if MICROPY_HW_TINYUSB_STACK
+#include "usbd_conf.h"
+#include "shared/tinyusb/mp_usbd.h"
+#endif
+
 #if MICROPY_PY_THREAD
 static pyb_thread_t pyb_thread_main;
 #endif
@@ -275,14 +280,12 @@ static bool init_sdcard_fs(void) {
                 }
             }
 
-            #if MICROPY_HW_ENABLE_USB
+            #if MICROPY_HW_ENABLE_USB && !MICROPY_HW_TINYUSB_STACK
             if (pyb_usb_storage_medium == PYB_USB_STORAGE_MEDIUM_NONE) {
                 // if no USB MSC medium is selected then use the SD card
                 pyb_usb_storage_medium = PYB_USB_STORAGE_MEDIUM_SDCARD;
             }
-            #endif
 
-            #if MICROPY_HW_ENABLE_USB
             // only use SD card as current directory if that's what the USB medium is
             if (pyb_usb_storage_medium == PYB_USB_STORAGE_MEDIUM_SDCARD)
             #endif
@@ -606,7 +609,12 @@ soft_reset:
     #endif
 
     #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_TINYUSB_STACK
+    pyb_usbd_init();
+    mp_usbd_init();
+    #else
     pyb_usb_init0();
+    #endif
     #endif
 
     #if MICROPY_PY_MACHINE_I2S
@@ -631,7 +639,7 @@ soft_reset:
     }
     #endif
 
-    #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_STM_USB_STACK
     // if the SD card isn't used as the USB MSC medium then use the internal flash
     if (pyb_usb_storage_medium == PYB_USB_STORAGE_MEDIUM_NONE) {
         pyb_usb_storage_medium = PYB_USB_STORAGE_MEDIUM_FLASH;
@@ -665,7 +673,7 @@ soft_reset:
     // or whose initialisation can be safely deferred until after running
     // boot.py.
 
-    #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_STM_USB_STACK
     // init USB device to default setting if it was not already configured
     if (!(pyb_usb_flags & PYB_USB_FLAG_USB_MODE_CALLED)) {
         #if MICROPY_HW_USB_MSC
@@ -769,6 +777,9 @@ soft_reset_exit:
     MP_STATE_PORT(pyb_stdio_uart) = &pyb_uart_repl_obj;
     #else
     MP_STATE_PORT(pyb_stdio_uart) = NULL;
+    #endif
+    #if MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE && MICROPY_HW_TINYUSB_STACK
+    mp_usbd_deinit();
     #endif
 
     MICROPY_BOARD_END_SOFT_RESET(&state);
--- a/ports/stm32/modmachine.c
+++ b/ports/stm32/modmachine.c
@@ -47,6 +47,7 @@
 #include "rtc.h"
 #include "i2c.h"
 #include "spi.h"
+#include "shared/tinyusb/mp_usbd.h"
 
 #if defined(STM32G0)
 // G0 has BOR and POR combined
@@ -297,9 +298,13 @@ MP_NORETURN static void mp_machine_reset(void) {
 
 // Activate the bootloader without BOOT* pins.
 MP_NORETURN void mp_machine_bootloader(size_t n_args, const mp_obj_t *args) {
-    #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_STM_USB_STACK
     pyb_usb_dev_deinit();
     #endif
+    #if MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE && MICROPY_HW_TINYUSB_STACK
+    mp_usbd_deinit();
+    #endif
+
     #if MICROPY_HW_ENABLE_STORAGE
     storage_flush();
     #endif
--- a/ports/stm32/modos.c
+++ b/ports/stm32/modos.c
@@ -52,7 +52,7 @@ bool mp_os_dupterm_is_builtin_stream(mp_const_obj_t stream) {
            #if MICROPY_PY_MACHINE_UART
            || type == &machine_uart_type
            #endif
-           #if MICROPY_HW_ENABLE_USB
+           #if MICROPY_HW_STM_USB_STACK
            || type == &pyb_usb_vcp_type
            #endif
     ;
@@ -64,7 +64,7 @@ void mp_os_dupterm_stream_detached_attached(mp_obj_t stream_detached, mp_obj_t s
         uart_attach_to_repl(MP_OBJ_TO_PTR(stream_detached), false);
     }
     #endif
-    #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_STM_USB_STACK
     if (mp_obj_get_type(stream_detached) == &pyb_usb_vcp_type) {
         usb_vcp_attach_to_repl(MP_OBJ_TO_PTR(stream_detached), false);
     }
@@ -75,7 +75,7 @@ void mp_os_dupterm_stream_detached_attached(mp_obj_t stream_detached, mp_obj_t s
         uart_attach_to_repl(MP_OBJ_TO_PTR(stream_attached), true);
     }
     #endif
-    #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_STM_USB_STACK
     if (mp_obj_get_type(stream_attached) == &pyb_usb_vcp_type) {
         usb_vcp_attach_to_repl(MP_OBJ_TO_PTR(stream_attached), true);
     }
--- a/ports/stm32/modpyb.c
+++ b/ports/stm32/modpyb.c
@@ -167,7 +167,7 @@ static const mp_rom_map_elem_t pyb_module_globals_table[] = {
     // Deprecated (use network.country instead).
     { MP_ROM_QSTR(MP_QSTR_country), MP_ROM_PTR(&mod_network_country_obj) },
 
-    #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_STM_USB_STACK
     { MP_ROM_QSTR(MP_QSTR_usb_mode), MP_ROM_PTR(&pyb_usb_mode_obj) },
     #if MICROPY_HW_USB_HID
     { MP_ROM_QSTR(MP_QSTR_hid_mouse), MP_ROM_PTR(&pyb_usb_hid_mouse_obj) },
--- a/ports/stm32/mpconfigboard_common.h
+++ b/ports/stm32/mpconfigboard_common.h
@@ -277,7 +277,7 @@
 #define MICROPY_HW_USB_CDC_NUM (1)
 #endif
 #ifndef MICROPY_HW_USB_MSC
-#define MICROPY_HW_USB_MSC (MICROPY_HW_STM_USB_STACK)
+#define MICROPY_HW_USB_MSC (MICROPY_HW_ENABLE_USB)
 #endif
 #ifndef MICROPY_HW_USB_HID
 #define MICROPY_HW_USB_HID (MICROPY_HW_STM_USB_STACK)
@@ -695,8 +695,13 @@
 #define MICROPY_HW_BDEV_WRITEBLOCKS(src, bl, n) spi_bdev_writeblocks(MICROPY_HW_BDEV_SPIFLASH, (src), MICROPY_HW_BDEV_SPIFLASH_OFFSET_BLOCKS + (bl), (n))
 #endif
 
+// Define the FATFS maximum sector size.
+#ifndef MICROPY_FATFS_MAX_SS
 #if defined(STM32N6)
 #define MICROPY_FATFS_MAX_SS                    (4096)
+#else
+#define MICROPY_FATFS_MAX_SS                    (512)
+#endif
 #endif
 
 // Whether to enable caching for external SPI flash, to allow block writes that are
--- a/ports/stm32/mpconfigboard_common.h
+++ b/ports/stm32/mpconfigboard_common.h
@@ -250,12 +250,50 @@
 #error "Old USBD_xxx configuration option used, renamed to MICROPY_HW_USB_xxx"
 #endif
 
+// Select whether TinyUSB or legacy STM stack is used to provide USB.
+#ifndef MICROPY_HW_TINYUSB_STACK
+#define MICROPY_HW_TINYUSB_STACK (0)
+#endif
+
+// Central definition for STM USB stack (when not using TinyUSB)
+#define MICROPY_HW_STM_USB_STACK (MICROPY_HW_ENABLE_USB && !MICROPY_HW_TINYUSB_STACK)
+
+#if MICROPY_HW_TINYUSB_STACK
+#ifndef MICROPY_HW_ENABLE_USBDEV
+#define MICROPY_HW_ENABLE_USBDEV (1)
+#endif
+
+#ifndef MICROPY_HW_USB_CDC
+#define MICROPY_HW_USB_CDC (1)
+#endif
+
+#ifndef MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE
+#define MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE    (1) // Support machine.USBDevice
+#endif
+#endif
+
+// Configure maximum number of CDC VCP interfaces, and whether MSC/HID are supported
+#ifndef MICROPY_HW_USB_CDC_NUM
+#define MICROPY_HW_USB_CDC_NUM (1)
+#endif
+#ifndef MICROPY_HW_USB_MSC
+#define MICROPY_HW_USB_MSC (MICROPY_HW_STM_USB_STACK)
+#endif
+#ifndef MICROPY_HW_USB_HID
+#define MICROPY_HW_USB_HID (MICROPY_HW_STM_USB_STACK)
+#endif
+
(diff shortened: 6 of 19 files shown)


dpgeorge:
> For now I (intend to) maintain back-compatibility with the `MICROPY_HW_ENABLE_STM_USB` flag to revert to the stm stack instead of tinyusb.

Yes, please do allow support for either the STM or TinyUSB stack.

shared/runtime/pyexec.c
```diff
@@ -583,7 +583,7 @@ int pyexec_friendly_repl(void) {
     for (;;) {
     input_restart:
 
-        #if MICROPY_HW_ENABLE_USB
+        #if MICROPY_HW_ENABLE_USB && 0
         if (usb_vcp_is_enabled()) {
             // If the user gets to here and interrupts are disabled then
             // they'll never see the prompt, traceback etc. The USB REPL needs
```
andrewleech, line 589: @dpgeorge This block in the shared pyexec code appears to be stm32 specific! Other ports use the `MICROPY_HW_ENABLE_USBDEV` define and don't define `usb_vcp_is_enabled` etc. 
This little feature to re-enable any disabled interrupts once a device hits repl does sound convenient, would it be worth extending this to all ports? Or does this feel like something users haven't needed and should be removed?

dpgeorge: I reckon we can just remove this.  It's pretty hacky and doesn't generalise well to other ports.

ports/stm32/usbd_conf.c
```diff
@@ -51,213 +51,230 @@ PCD_HandleTypeDef pcd_hs_handle;
...
+    #endif // !MICROPY_HW_USB_HS_IN_FS
+
+    // Configure and enable USB HS interrupt
+    NVIC_SetPriority(OTG_HS_IRQn, IRQ_PRI_OTG_HS);
+    HAL_NVIC_EnableIRQ(OTG_HS_IRQn);
+    #endif
+}
+
+
+void pyb_usbd_init(void) {
```
andrewleech: This isn't needed for stm hal, adds code size.

ports/stm32/usbd_conf.c
```diff
@@ -51,213 +51,230 @@ PCD_HandleTypeDef pcd_hs_handle;
...
-                       PCD BSP Routines
-*******************************************************************************/
-
-/**
-  * @brief  Initializes the PCD MSP.
-  * @param  hpcd: PCD handle
-  * @retval None
-  */
-void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd) {
+void pyb_usbd_init_fs(void) {
```
andrewleech: make these static

shared/tinyusb/tusb_config.h
```diff
@@ -59,8 +59,25 @@
 #define MICROPY_HW_USB_MSC_INQUIRY_REVISION_STRING "1.00"
 #endif
 
+#if MICROPY_HW_USB_HS
+#define OPT_MODE_PORT_SPEED OPT_MODE_HIGH_SPEED
```
dpgeorge: I'm not sure this works, because a lot of boards (eg PYBD_SF2/6) define `MICROPY_HW_USB_HS` yet also define `MICROPY_HW_USB_HS_IN_FS` and so run at FS speed.

dpgeorge:
Code size change for stm32 is still very high....

shared/tinyusb/tusb_config.h
andrewleech: Yeah valid point; I've removed all the HS changes from this file as they're not directly related to this PR anyway and I don't have a good way to test them.

andrewleech:
> Code size change for stm32 is still very high....

Yes it is! I've had to do a fairly deep dive into the code / build to figure this one out, it's more than the tinyusb files being always compiled (to avoid promoting this config to a Makefile setting) - it seems most of this size difference is coming from a reordering in the linker/map causing changes in padding between symbols, or in some cases its' from which module first includes a common element etc. A minor example of this; This diff alone causes a +8 bytes in total flash use: 
image

Regardless, it looks like I'll be able to fix this with reording some elements and adding more `#ifdef` around stuff I missed in the last update (eg. the new `static inline void mp_hal_wake_main_task_from_isr(void)` function in mphalport.h)

ports/stm32/mpconfigport.h
```diff
@@ -179,6 +179,9 @@
 #define MICROPY_FATFS_USE_LABEL        (1)
 #define MICROPY_FATFS_RPATH            (2)
 #define MICROPY_FATFS_MULTI_PARTITION  (1)
+#if MICROPY_HW_TINYUSB_STACK && CFG_TUD_MSC
+#define MICROPY_FATFS_MAX_SS           (4096)
```
andrewleech: @dpgeorge It looks like this was the primary culprit for the flash size increase; with this define added it bumps up the FATFS cache size and caused the ~350 byte increase. The guards around it resolve that.

dpgeorge: Ah, yes, that makes sense!

andrewleech:
I believe the size issue (when building with stm usb stack) has been resolved finally. Though I want to check exactly how it's now smaller...

On a positive note, switching to TinyUSB actually reduces the size by ~5.4KB.

dpgeorge:
> I believe the size issue (when building with stm usb stack) has been resolved finally. Though I want to check exactly how it's now smaller...

Yes, that's a bit concerning that it's smaller.  Probably something has been unintentionally removed/disabled.

hoihu:
nice work, heads up from my side. would be cool to use all the „dynamic python-USB configuration“ staff in the stm32 port!

andrewleech:
> > I believe the size issue (when building with stm usb stack) has been resolved finally. Though I want to check exactly how it's now smaller...
> 
> Yes, that's a bit concerning that it's smaller. Probably something has been unintentionally removed/disabled.

Whoops, I sliced and diced the feature commit every which way trying to find the size difference befrore remember the other commit at the base of this branch: [runtime/pyexec: Remove legacy usb irq enable code.](https://github.com/micropython/micropython/pull/15592/commits/e2f26a272c2946ee2ebaa33286e910a705f9d7ae)

That's the drop in size.

dpgeorge:
I tried to test this on PYBD_SF6, PYBV10 and NUCLEO_WB55 but it did not work:
- PYBD_SF6 uses both FS and HS ports, and the main one is the HS (secondary port), and from what I can tell this isn't supported by TinyUSB (or at least our configuration).  So I gave up on this board.
- PYBV10 I expected to work because it's simple and just uses FS.  But enabling TinyUSB, nothing happens, no USB device appears.
- NUCLEO_WB55 I also expected to work, but similar to PYBV10 no USB device appears.  The UART REPL is still there and working, so the board starts up, but no native USB.

To enable TinyUSB I changed `mpconfigboard_common.h`:
```diff
--- a/ports/stm32/mpconfigboard_common.h
+++ b/ports/stm32/mpconfigboard_common.h
@@ -142,7 +142,7 @@
 
 // Select whether TinyUSB or legacy STM stack is used to provide USB.
 #ifndef MICROPY_HW_TINYUSB_STACK
-#define MICROPY_HW_TINYUSB_STACK (0)
+#define MICROPY_HW_TINYUSB_STACK (1)
 #endif
```
and it's definitely using that config option.

Did I do something wrong?

ports/stm32/mpconfigboard_common.h
```diff
@@ -136,6 +136,24 @@
 #define MICROPY_HW_ENABLE_USB (0)
 #endif
 
+#if MICROPY_HW_ENABLE_USB
+#define MICROPY_HW_USB_CDC (1)
+#define MICROPY_HW_USB_FS (1)
```
dpgeorge: These options cannot be enabled unconditionally, because they may clash with board configurations (when using the STM USB stack).

Either just remove these lines (my preference, if that works) or at least only enable them if `MICROPY_HW_TINYUSB_STACK` is enabled.

ports/stm32/mpconfigboard_common.h
```diff
@@ -256,6 +274,8 @@
 // Windows needs a different PID to distinguish different device configurations.
 #ifndef MICROPY_HW_USB_VID
 #define MICROPY_HW_USB_VID              (0xf055)
+#define MICROPY_HW_USB_PID              (0x9802)
```
dpgeorge, line 294: Maybe add a comment that this is only used in the TinyUSB configuration (and eventually we'll need to change this value depending on whether MSC is enabled etc).

andrewleech:
Thanks for the feedback @dpgeorge I'll address your specific review comments and test / fix it on the nucleo and pybd - I'll let you know when it's ready to test again

andrewleech:
@dpgeorge it turns out the WB55 / tinyusb issue was both more complicated and simpler than expected. 

The 16bit register access issue I found with tinyusb/1cf8e34ae5f had actually already been solved in a newer version of TinyUSB (I should have looked there first) but that alone didn't fix the issue.

Claude then figured out that the wrong USB base register was being used; in tinyusb it uses `#define FSDEV_REG_BASE USB_BASE` but for WB55 it needed to be `#define FSDEV_REG_BASE USB1_BASE` and that fixed micropython support.
... except then a recursive grep for `USB1_BASE` also found: 
```
./hw/mcu/st/cmsis_device_wb/Include/stm32wb55xx.h                  
969:#define USB1_BASE             (APB1PERIPH_BASE + 0x00006800UL)  // ((0x40000000UL) + 0x00006800UL)
1091:#define USB                 ((USB_TypeDef *) USB1_BASE)       
13070:#define USB_BASE                             (USB1_BASE)     
```
Turns out that's a newer version of the WB hal that tinyusb pulls, our one in microypthon is older and doesn't have https://github.com/STMicroelectronics/cmsis-device-wb/commit/d1b860584dfe24d40d455ae624ed14600dfa93c9#diff-a7f3443ed971e94fc67ce2c7fcdfde0030f573aeaaab4412180d4ef6d7ca7689R12165 
In our copy we have `#define USB_BASE (0x40005C00UL)` whereas `USB1_BASE` expands to `0x40006800UL` . 

TLDR no tinyusb PR is needed, but we need to update both tinyusb to the latest release _and_ [lib/stm32lib/CMSIS/STM32WBxx](https://github.com/micropython/stm32lib/pull/26)

dpgeorge:
Interesting!

Does that fix PYBV10 as well?

andrewleech:
I'm not sure yet, I actually don't have one at home (really should buy one) but I do have a F411 which I should be able to test. I obviosly need to clean up the build failures first regardless!

dpgeorge:
Oh, I should definitely give you a PYBV1x!

If you get this PR passing CI, then I can test PYBV10 myself.

dpgeorge:
Maybe it's worth splitting out the TinyUSB update to a separate PR?  Then it can be tested against existing TinyUSB-based ports.

andrewleech:
> Maybe it's worth splitting out the TinyUSB update to a separate PR? Then it can be tested against existing TinyUSB-based ports.

I've reordered the branch to split this out into https://github.com/micropython/micropython/pull/18160

andrewleech:
Hi @dpgeorge I found why this would have been failing on the pyboard; 403634cbd5dcb3d634653f54fb48f70db10a7f8d - vbus sensing for F4/F7 wasn't configured correctly. This has been fixed (tested on the F429) which I expect should apply equally to the pyboard.

Once the rest of the stmlib updates go through I'll rebase on top of that/master.

dpgeorge:
> Once the rest of the stmlib updates go through I'll rebase on top of that/master.

stm32lib is ready, so let's get that in first via #18268.

dpgeorge:
I tested this PR on PYBV10.  The CDC works!  That's a really god milestone.

Testing performance with `tests/serial_test.py`, the DATA IN throughput (in to the host PC) is the same as with the stm32 USB driver.  But the DATA OUT throughput is about 4 times slower with TinyUSB compared to the existing stm32 USB driver.  That doesn't really matter for now, but is definitely something to investigate later on.

I also tried out `machine.USBDevice`.  Unfortunately that didn't work: trying to create a USB HID keyboard device, the host PC fails to fully enumerate the PYBV10 (dmesg shows `can't set config #1, error -32`).  So that's also something to investigate, but could wait until after this PR as well.

dpgeorge:
> I also tried out `machine.USBDevice`. Unfortunately that didn't work

Aah, that's simply because it was trying to create a CDC+MSC+HID device and there are not enough hardware endpoints on STM32F405 for that to work.  If I disable MSC (so the builtin is now `BUILTIN_CDC` instead of `BUILTIN_CDC_MSC`) then a custom HID keyboard using `machine.USBDevice` works!

That's excellent.

iabdalkader:
> But the DATA OUT throughput is about 4 times slower with TinyUSB compared to the existing stm32 USB driver.

Try increasing the CDC TX/RX buffers:
```C
#define CFG_TUD_CDC_EP_BUFSIZE  (4096)
#define CFG_TUD_CDC_RX_BUFSIZE  (4096)
#define CFG_TUD_CDC_TX_BUFSIZE  (4096)
```

dpgeorge:
> #define CFG_TUD_CDC_EP_BUFSIZE  (4096)

I don't think it's possible to change this option, it needs to be the actual size of the endpoint (64 bytes for full speed).  I tried changing it and CDC stopped working.

> ```c
> #define CFG_TUD_CDC_RX_BUFSIZE  (4096)
> #define CFG_TUD_CDC_TX_BUFSIZE  (4096)
> ```

I tried this.  Unfortunately it didn't make much difference to the CDC throughput.

My guess is that the loss in performance comes from the way the USB IRQ is handled/scheduled.

iabdalkader:
> I don't think it's possible to change this option, it needs to be the actual size of the endpoint (64 bytes for full speed). I tried changing it and CDC stopped working.

We use it in Alif and mimxrt ports, they do improve the throughput by a lot. I think you just need to increase all of them at the same time (needs to match TX/RX_BUFSIZE).

andrewleech:
> But the DATA OUT throughput is about 4 times slower with TinyUSB compared to the existing stm32 USB driver.

I think this is because with your speed test script it checks for stm in particular and uses the direct usb VCP object to send data from device to pc; it's able to read directly from the provided buffer or something iirc. That was part of the reason I was wanting to replicate the concept of a direct usb object for cdc access from python on the tinyusb devices.

Oh, or it might be because of the (lack of) accelerated pathway in dupterm, again I was going to add that in the cdc class exposed to python (follow on pr if it's worth the size-add).

> not enough hardware endpoints on STM32F405 for that to work.

It would be good to test for this and raise a helpful error!

andrewleech:
> > #define CFG_TUD_CDC_EP_BUFSIZE  (4096)
> 
> I don't think it's possible to change this option, it needs to be the actual size of the endpoint (64 bytes for full speed). I tried changing it and CDC stopped working.

See https://github.com/hathach/tinyusb/discussions/2707 and this https://github.com/hathach/tinyusb/discussions/979

ports/stm32/Makefile
```diff
@@ -238,16 +245,29 @@ DRIVERS_SRC_C += $(addprefix drivers/,\
 	dht/dht.c \
 	)
 
+# TinyUSB Stack source
+-include $(TOP)/lib/tinyusb/src/tinyusb.mk
```
dpgeorge: Please move this new section to down just below the definition of `USBDEV_SRC_C`.

ports/stm32/main.c
```diff
@@ -88,6 +87,15 @@
 #include "pyb_can.h"
 #include "subghz.h"
 
+#if MICROPY_HW_TINYUSB_STACK
+#include "usbd_conf.h"
+#include "shared/tinyusb/mp_usbd.h"
+#endif
+
+#if MICROPY_HW_ENABLE_USB
+#include "usb.h"
```
dpgeorge: Couldn't we just unconditionally include this particular header, like it was before this PR?  Keeps it simple.

ports/stm32/modmachine.c
```diff
@@ -297,9 +298,13 @@ MP_NORETURN static void mp_machine_reset(void) {
 
 // Activate the bootloader without BOOT* pins.
 MP_NORETURN void mp_machine_bootloader(size_t n_args, const mp_obj_t *args) {
-    #if MICROPY_HW_ENABLE_USB
+    #if MICROPY_HW_ENABLE_USB && !MICROPY_HW_TINYUSB_STACK
```
dpgeorge: I think it'll be clearer if this was `MICROPY_HW_STM_USB_STACK`.

There are a few other cases of this as well.

ports/stm32/modpyb.c
```diff
@@ -49,14 +49,17 @@
...
-#include "usb.h"
 #include "portmodules.h"
 #include "modmachine.h"
 #include "extmod/modmachine.h"
 #include "extmod/modnetwork.h"
 #include "extmod/vfs.h"
 #include "extmod/modtime.h"
 
+#if !MICROPY_HW_TINYUSB_STACK
+#include "usb.h"
```
dpgeorge: I think just include this unconditionally, like it was before this PR.

ports/stm32/mpconfigboard_common.h
```diff
@@ -136,6 +136,32 @@
 #define MICROPY_HW_ENABLE_USB (0)
 #endif
 
+#if MICROPY_HW_ENABLE_USB
```
dpgeorge: I suggest moving this new config section down into the existing section that has the header:
```c
/*****************************************************************************/
// USB configuration
```
Maybe just after the `#error "Old USBD_xxx configuration option used, renamed to MICROPY_HW_USB_xxx"` bit.

Also, suggest moving the existing
```c
// Configure maximum number of CDC VCP interfaces, and whether MSC/HID are supported
#ifndef MICROPY_HW_USB_CDC_NUM
#define MICROPY_HW_USB_CDC_NUM (1)
#endif
#ifndef MICROPY_HW_USB_MSC
#define MICROPY_HW_USB_MSC (MICROPY_HW_ENABLE_USB)
#endif
#ifndef MICROPY_HW_USB_HID
#define MICROPY_HW_USB_HID (MICROPY_HW_STM_USB_STACK)
#endif
```
part from this file to the same location, so all the USB is consolidated into one location.

(Except for the `#define MICROPY_HW_ENABLE_USB (0)`, that's the top-level config option that should stay where it is.)

ports/stm32/mpconfigport.h
```diff
@@ -176,6 +176,15 @@
 #define MICROPY_FATFS_USE_LABEL        (1)
 #define MICROPY_FATFS_RPATH            (2)
 #define MICROPY_FATFS_MULTI_PARTITION  (1)
+#if MICROPY_HW_USB_MSC && MICROPY_HW_TINYUSB_STACK
```
dpgeorge: I think this whole bit that's added here can go.

512 is the default, so that will be used for everything except N6.

For N6... I think it's a separate problem to fix that if something doesn't work with MSC on that SoC.

ports/stm32/mphalport.c
```diff
@@ -8,6 +8,17 @@
 #include "usb.h"
 #include "uart.h"
 
+#if MICROPY_HW_TINYUSB_STACK
+#include "shared/tinyusb/mp_usbd_cdc.h"
+
+#ifndef MICROPY_HW_STDIN_BUFFER_LEN
+#define MICROPY_HW_STDIN_BUFFER_LEN 128
```
dpgeorge: Suggest making this 512, following rp2 and alif ports.

ports/stm32/mphalport.c
```diff
@@ -26,6 +37,9 @@ MP_NORETURN void mp_hal_raise(HAL_StatusTypeDef status) {
 
 MP_WEAK uintptr_t mp_hal_stdio_poll(uintptr_t poll_flags) {
     uintptr_t ret = 0;
+    #if MICROPY_HW_ENABLE_USBDEV && MICROPY_HW_USB_CDC && MICROPY_HW_TINYUSB_STACK
```
dpgeorge: I think you can remove the `MICROPY_HW_ENABLE_USBDEV &&` bit, that's guaranteed to be true if TinyUSB is enabled.

ports/stm32/mphalport.c
```diff
@@ -53,6 +67,13 @@ MP_WEAK int mp_hal_stdin_rx_chr(void) {
         if (dupterm_c >= 0) {
             return dupterm_c;
         }
+        #if MICROPY_HW_ENABLE_USBDEV && MICROPY_HW_USB_CDC && MICROPY_HW_TINYUSB_STACK
```
dpgeorge: Can remove `MICROPY_HW_ENABLE_USBDEV &&`.

ports/stm32/mphalport.c
```diff
@@ -64,6 +85,13 @@ MP_WEAK mp_uint_t mp_hal_stdout_tx_strn(const char *str, size_t len) {
         uart_tx_strn(MP_STATE_PORT(pyb_stdio_uart), str, len);
         did_write = true;
     }
+    #if MICROPY_HW_ENABLE_USBDEV && MICROPY_HW_USB_CDC && MICROPY_HW_TINYUSB_STACK
```
dpgeorge: Can remove `MICROPY_HW_ENABLE_USBDEV &&`.

ports/stm32/usbd.c
```diff
@@ -0,0 +1,42 @@
...
+
+#include "mp_usbd.h"
+#include "py/mpconfig.h"
+#include "string.h"
+#include "mphalport.h"
+
+void mp_usbd_port_get_serial_number(char *serial_buf) {
+    uint8_t *id = (uint8_t *)MP_HAL_UNIQUE_ID_ADDRESS;
+    MP_STATIC_ASSERT(sizeof(MP_HAL_UNIQUE_ID_ADDRESS) * 2 <= MICROPY_HW_USB_DESC_STR_MAX);
+    mp_usbd_hex_str(serial_buf, id, sizeof(MP_HAL_UNIQUE_ID_ADDRESS));
```
dpgeorge: I think these `sizeof`'s need to be replaced with a constant 12.  Otherwise they are just returning 4 probably, which is the size of the integer constant `MP_HAL_UNIQUE_ID_ADDRESS`.

ports/stm32/usbd_conf.c
```diff
@@ -56,18 +62,17 @@ PCD_HandleTypeDef pcd_hs_handle;
...
-void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd) {
+#if MICROPY_HW_TINYUSB_STACK
+void pyb_usbd_init(void)
+#else
+void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd)
+#endif
+{
     #if MICROPY_HW_USB_FS
-    if (hpcd->Instance == USB_OTG_FS) {
+    #if !MICROPY_HW_TINYUSB_STACK
```
dpgeorge: Suggest `#if MICROPY_HW_STM_USB_STACK` instead.

ports/stm32/usbd_conf.c
```diff
@@ -172,12 +177,33 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd) {
...
+
+        #if !MICROPY_HW_TINYUSB_STACK
         return;
+        #endif
     }
     #endif
 
     #if MICROPY_HW_USB_HS
-    if (hpcd->Instance == USB_OTG_HS) {
+    #if !MICROPY_HW_TINYUSB_STACK
```
dpgeorge: Suggest `#if MICROPY_HW_STM_USB_STACK` instead.

ports/stm32/usbd_conf.c
```diff
@@ -290,6 +316,8 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd) {
     #endif // MICROPY_HW_USB_HS
 }
 
+#if !MICROPY_HW_TINYUSB_STACK
```
dpgeorge: Suggest `#if MICROPY_HW_STM_USB_STACK` instead.

tools/ci.sh
```diff
@@ -515,6 +515,7 @@ function ci_stm32_nucleo_build {
     make ${MAKEOPTS} -C ports/stm32 BOARD=NUCLEO_H743ZI COPT=-O2 CFLAGS_EXTRA='-DMICROPY_PY_THREAD=1'
     make ${MAKEOPTS} -C ports/stm32 BOARD=NUCLEO_L073RZ
     make ${MAKEOPTS} -C ports/stm32 BOARD=NUCLEO_L476RG DEBUG=1
+    make ${MAKEOPTS} -C ports/stm32 BOARD=STM32H573I_DK CFLAGS_EXTRA='-DMICROPY_HW_TINYUSB_STACK=1'
```
dpgeorge: Instead of adding yet another build to the CI, I think just replace the existing STM32H573I_DK with this line, ie just build that existing board with TinyUSB enabled.

dpgeorge:
@andrewleech this is looking really good, thanks.  Just some comments to address (see above) which should be mostly uncontroversial.  Then it'll be ready to merge.

andrewleech:
> @andrewleech this is looking really good, thanks. Just some comments to address (see above) which should be mostly uncontroversial. Then it'll be ready to merge.

Thank you for the comments - yeah the `#if !MICROPY_HW_TINYUSB_STACK` must have crept back in from some wayward rebasing / testing on different computers. Too many branches on different computers to keep it all in check!

I'll clean up as per your comments cheers!

dpgeorge:
The CI is failing.  I suggest the following patch to hopefully fix that:
```diff
--- a/ports/stm32/mpconfigboard_common.h
+++ b/ports/stm32/mpconfigboard_common.h
@@ -635,8 +635,13 @@
 #define MICROPY_HW_BDEV_WRITEBLOCKS(src, bl, n) spi_bdev_writeblocks(MICROPY_HW_BDEV_SPIFLASH, (src), MICROPY_HW_BDEV_SPIFLASH_OFFSET_BLOCKS + (bl), (n))
 #endif
 
+// Define the FATFS maximum sector size.
+#ifndef MICROPY_FATFS_MAX_SS
 #if defined(STM32N6)
 #define MICROPY_FATFS_MAX_SS                    (4096)
+#else
+#define MICROPY_FATFS_MAX_SS                    (512)
+#endif
 #endif
```

dpgeorge:
Thanks for updating, this is looking pretty good now.  I'll do some final testing on my side, then hopefully merge.

ports/stm32/modos.c
```diff
@@ -52,7 +52,7 @@ bool mp_os_dupterm_is_builtin_stream(mp_const_obj_t stream) {
            #if MICROPY_PY_MACHINE_UART
            || type == &machine_uart_type
            #endif
-           #if MICROPY_HW_ENABLE_USB
+           #if MICROPY_HW_STM_USB_STACK
            || type == &pyb_usb_vcp_type
```
andrewleech, line 56: @dpgeorge this will be one of the serial port speed regressions on tinyusb; I've got ideas on how to fix this that will then apply to all tinyusb ports but that can be a follow up change!

dpgeorge: OK!  A follow-up PR sounds good for that.

dpgeorge (approved):
Tested on PYBV11, with STM USB and TinyUSB.  It works.

dpgeorge:
Merged!

Thanks @andrewleech for your efforts with this, it's a great step forward for the stm32 port!

iabdalkader:
@dpgeorge @andrewleech I tested this with STM32H7 and STM32N6 (the N6 needs tinyusb 0.19.0), it works flawlessly on the H7, but has some issues on N6. Nevertheless, awesome work!

On the N6, I only had to define `CFG_TUSB_RHPORT0_MODE (OPT_MODE_DEVICE | OPT_MODE_HIGH_SPEED)` to make it work. It seems to work fine until soft-reboot (ctrl+d) then the device disconnects. I didn't test it with upstream code so I'm not sure if it's our code or something else that's misconfigured. @dpgeorge Could you give it a try and let me know if it works for you?

I notice a few other issues:
1- The board serial number changes after switching.
2- Can we reuse the old `MICROPY_HW_USB_CDC_TX/RX_DATA_SIZE` to set `CFG_TUD_CDC_TX/RX_BUFSIZE`? It feels more consistent with other board config macros, than `CFG_TUD_xxx`, and this way boards can switch between stacks seamlessly. We'll need a new one for `CFG_TUD_CDC_EP_BUFSIZE` 
3- You may want to update to tinyusb 0.19.0 while you're at it, it's the first release with N6 support.

andrewleech:
Thanks @iabdalkader I've got another branch under testing that adds high speed support, configured automatically from the existing board defines. The buffer size define suggestion is good, I'll add it as well.

I'll change of serial number was unintended, I'd like to look into that - thanks for reporting it.

A disconnect over soft reset is normal if USB runtime settings are used in boot.py but not expected otherwise, that behaviour is curious. Perhaps though I've got a re-init of some kind after the soft reset that's being ignored on older boards but triggers a reset on n6.

iabdalkader:
> A disconnect over soft reset is normal if USB runtime settings are used in boot.py but not expected otherwise, that behaviour is curious. Perhaps though I've got a re-init of some kind after the soft reset that's being ignored on older boards but triggers a reset on n6.

@andrewleech  I'm not using the USB runtime. I can't be 100% sure if it's our code/fork or something specific to N6 in TinyUSB or MicroPython, but given that the H7 works fine, it's likely Not our code.  I've also tested tinyusb `0.19.0` just now with H7, and it works fine.

iabdalkader:
It seems the problem is calling `pyb_usbd_init()` multiple times, it seems the N6 doesn't like that.

iabdalkader:
@dpgeorge Might be unrelated, but there's an issue with `mp_usbd_cdc.c:mp_usbd_cdc_tx_strn`: if the FIFO is full *and* the device is disconnected, it will never timeout or break. Admittedly, I can only reproduce this with our firmware, probably because I send FIFO-sized packets with our protocol stuff, but suppose upstream can get into that state somehow, you need to run the `tud_task` for `tud_cdc_connected()` to change, to even run that inner timeout loop.

Removing `if (tud_cdc_connected())` completely, or just adding the following to the loop tail, fixes this issue for me:

```C
        if (tud_task_event_ready()) {
            mp_usbd_task();
        }
```

Is this something you're interested in improving/fixing, can I send a PR for this?

andrewleech:
@iabdalkader thanks for looking into that, it's certainly something I'm interested in having improved! These kind of corner-cases are the hardest to nail down and you've done most of the hard work already.

I don't remember the full context of that area of code clearly but a lot of the tud_cdc_connected() logic there was to support the fifo being able to be filled and maintained before it's connected so that on first connected you can get the startup banner etc displayed that was captured earlier in startup. It was supposed to properly fifo that buffer though to only show the most recent data.

I expect that removing if (tud_cdc_connected()) will break the startup buffering, however adding the tud_task_event_ready() check feels like relatively low overhead to add to fix the lock up!

iabdalkader:
> I expect that removing if (tud_cdc_connected()) will break the startup buffering, however adding the tud_task_event_ready() check feels like relatively low overhead to add to fix the lock up!

I realized I was looking at the file from the 1.26.0 release, and it's very different from the upstream one, so I think someone may have already fixed it. I also see that the task is always called now, which probably resolves the issue. It would take me a while to retest this, but it looks promising.

The only issue now is with `pyb_usbd_init`. If I call it just once (not on every soft reboot), the N6 works fine, and it’s running in high-speed mode. I can also switch TinyUSB submodule in our fork, so nothing is really needed here.

Found the fix: https://github.com/micropython/micropython/commit/a43e38544bf707831350dd53091c7d47eddd41bf

dpgeorge:
See #18402 for some additional USB testing which uncovers a bug with stm32 and TinyUSB.

PR #15598, opened 2024-08-05
qemu-arm: Fix tinytest test profile when updating set of dirs/files.

dpgeorge:
### Summary

Updating a set must use `.update()` rather than `.add()`.

Also apply the same pattern to qemu-riscv to prevent the same issue when directories/files are added to that port's `tests_profile.txt` file.

### Testing

Prior to this change, the qemu-arm port was not running any of the `inlineasm` or `ports/qemu-arm` tests.  Now it does (verified by inspecting the `build/genhdr/test.h` and `build/console.out` files).

### DIFF
--- a/ports/qemu-arm/tests_profile.txt
+++ b/ports/qemu-arm/tests_profile.txt
@@ -1,10 +1,10 @@
 # Port-specific test directories.
 
-test_dirs.add(("inlineasm", "qemu-arm"))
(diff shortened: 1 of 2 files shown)


projectgus (approved):
LGTM! (I also ran the tests locally to confirm I still understood the process.)

As discussed, I think that long term finding a way to interact with the qemu REPL via stdin/stdout would be a great step.

PR #15603, opened 2024-08-05
natmod: Add native modules support for RV32 code.

agatti:
### Summary

This commit adds support for RV32IMC native modules, which is one of the few remaining bits left to bring RV32 support to parity with Thumb or x86 (the other is inline assembler). 

### Testing

All tests were performed on an ESP32C3 board, with cross-compilation done in an environment similar to the CI machine used to build the ESP32 and Qemu-RISCV ports.  The natmods being built are the ones from the `examples/natmod/` directory, using the tests in `tests/extmod/` to exercise the code in a known way whenever possible.

There isn't any alternative to this in order to get natmod support on RV32, and there are no real trade-offs in execution speed or in code size.  A full-fledged linker would be able to convert some two-opcodes relocations into a single opcode or perform similar optimisations (the infamous R_RISCV_RELAX relocation marker), but this isn't it.  The RV32-specific relocations handling code had to be moved to its own function to please the linter when it measures function complexity.

One thing worth mentioning is the insane amount of [possible static relocation types](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#reloc-table) that must be handled (around 35 in total).  Right now I've implemented all the ones I've seen from building the sample natmods, for the ones I didn't see yet, I'll have to add them in blindly and hope the documentation is accurate.

The makefile changes to add native compilation are pretty much the same as what was originally written for the Qemu-RISCV port.  It's quite convoluted but I couldn't find any better way to simplify that and get the same out-of-the-box compilation output under various Linux environments.

### DIFF
--- a/examples/natmod/btree/Makefile
+++ b/examples/natmod/btree/Makefile
@@ -7,7 +7,7 @@ MOD = btree_$(ARCH)
 # Source files (.c or .py)
 SRC = btree_c.c
 
-# Architecture to build for (x86, x64, armv7m, xtensa, xtensawin)
+# Architecture to build for (x86, x64, armv7m, xtensa, xtensawin, rv32imc)
 ARCH = x64
 
 BTREE_DIR = $(MPY_DIR)/lib/berkeley-db-1.xx
--- a/ports/qemu/README.md
+++ b/ports/qemu/README.md
@@ -71,8 +71,9 @@ To access the REPL directly use:
 
     $ make repl
 
-This will start `qemu-system-arm` with the UART redirected to stdio.  It's also
-possible to redirect the UART to a pty device using:
+This will start `qemu-system-arm` (or `qemu-system-riscv32`) with the UART
+redirected to stdio.  It's also possible to redirect the UART to a pty device
+using:
 
     $ make run
 
@@ -84,7 +85,7 @@ for example `mpremote`:
 
 You can disconnect and reconnect to the serial device multiple times.  Once you
 are finished, stop the `make run` command by pressing Ctrl-C where that command
-was started (or execute `machine.reset()` at the REPL).
+was started (or execute `import machine; machine.reset()` at the REPL).
 
 The test suite can be run against the firmware by using the UART redirection.
 You can either do this automatically using the single command:
@@ -96,6 +97,17 @@ tests against the serial device, for example:
 
     $ cd ../../tests
     $ ./run-tests.py -t /dev/pts/1
+
+Selected native modules that come as examples with the MicroPython source tree
+can also be tested with this command (this is currently supported only for the
+`VIRT_RV32` board):
+
+    $ make test_natmod
+
+The same remarks about manually running the tests apply for native modules, but
+`run-natmodtests.py` should be run instead of `run-tests.py`.  In this case you
+also have to explicitly pass the architecture you are running native modules to
+`run-natmodtests.py` ("--arch rv32imc" for the `VIRT_RV32` board).
 
 Extra make options
 ------------------
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -273,6 +273,7 @@ function ci_qemu_setup_arm {
 }
 
 function ci_qemu_setup_rv32 {
+    ci_mpy_format_setup
     ci_gcc_riscv_setup
     sudo apt-get update
     sudo apt-get install qemu-system
@@ -292,6 +293,10 @@ function ci_qemu_build_rv32 {
     make ${MAKEOPTS} -C mpy-cross
     make ${MAKEOPTS} -C ports/qemu BOARD=VIRT_RV32 submodules
     make ${MAKEOPTS} -C ports/qemu BOARD=VIRT_RV32 test
+
+    # Test building and running native .mpy with rv32imc architecture.
+    ci_native_mpy_modules_build rv32imc
+    make ${MAKEOPTS} -C ports/qemu BOARD=VIRT_RV32 test_natmod
 }
 
 ########################################################################################
@@ -476,10 +481,14 @@ function ci_native_mpy_modules_build {
         arch=$1
     fi
     make -C examples/natmod/features1 ARCH=$arch
-    make -C examples/natmod/features2 ARCH=$arch
+    if [ $arch != rv32imc ]; then
+        # This requires soft-float support on rv32imc.
+        make -C examples/natmod/features2 ARCH=$arch
+        # This requires thread local storage support on rv32imc.
+        make -C examples/natmod/btree ARCH=$arch
+    fi
     make -C examples/natmod/features3 ARCH=$arch
     make -C examples/natmod/features4 ARCH=$arch
-    make -C examples/natmod/btree ARCH=$arch
     make -C examples/natmod/deflate ARCH=$arch
     make -C examples/natmod/framebuf ARCH=$arch
     make -C examples/natmod/heapq ARCH=$arch
--- a/tools/mpy_ld.py
+++ b/tools/mpy_ld.py
@@ -47,6 +47,7 @@ MP_NATIVE_ARCH_ARMV7EMSP = 7
 MP_NATIVE_ARCH_ARMV7EMDP = 8
 MP_NATIVE_ARCH_XTENSA = 9
 MP_NATIVE_ARCH_XTENSAWIN = 10
+MP_NATIVE_ARCH_RV32IMC = 11
 MP_PERSISTENT_OBJ_STR = 5
 MP_SCOPE_FLAG_VIPERRELOC = 0x10
 MP_SCOPE_FLAG_VIPERRODATA = 0x20
@@ -56,6 +57,7 @@ MP_FUN_TABLE_MP_TYPE_TYPE_OFFSET = 73
 
 # ELF constants
 R_386_32 = 1
+R_RISCV_32 = 1
 R_X86_64_64 = 1
 R_XTENSA_32 = 1
 R_386_PC32 = 2
@@ -70,15 +72,57 @@ R_386_GOTOFF = 9
 R_386_GOTPC = 10
 R_ARM_THM_CALL = 10
 R_XTENSA_ASM_EXPAND = 11
+R_RISCV_BRANCH = 16
+R_RISCV_JAL = 17
+R_RISCV_CALL = 18
+R_RISCV_CALL_PLT = 19
 R_XTENSA_DIFF32 = 19
 R_XTENSA_SLOT0_OP = 20
+R_RISCV_GOT_HI20 = 20
+R_RISCV_TLS_GD_HI20 = 22
+R_RISCV_PCREL_HI20 = 23
+R_RISCV_PCREL_LO12_I = 24
+R_RISCV_PCREL_LO12_S = 25
 R_ARM_BASE_PREL = 25  # aka R_ARM_GOTPC
 R_ARM_GOT_BREL = 26  # aka R_ARM_GOT32
 R_ARM_THM_JUMP24 = 30
+R_RISCV_HI20 = 26
+R_RISCV_LO12_I = 27
+R_RISCV_LO12_S = 28
+R_RISCV_TPREL_HI20 = 29
+R_RISCV_TPREL_LO12_I = 30
+R_RISCV_TPREL_LO12_S = 31
+R_RISCV_TPREL_ADD = 32
+R_RISCV_ADD8 = 33
+R_RISCV_ADD16 = 34
+R_RISCV_ADD32 = 35
+R_RISCV_ADD64 = 36
+R_RISCV_SUB8 = 37
+R_RISCV_SUB16 = 38
+R_RISCV_SUB32 = 39
+R_RISCV_SUB64 = 40
+R_RISCV_GOT32_PCREL = 41
 R_X86_64_GOTPCREL = 9
 R_X86_64_REX_GOTPCRELX = 42
 R_386_GOT32X = 43
+R_RISCV_ALIGN = 43
+R_RISCV_RVC_BRANCH = 44
+R_RISCV_RVC_JUMP = 45
+R_RISCV_RELAX = 51
+R_RISCV_SUB6 = 52
+R_RISCV_SET6 = 53
+R_RISCV_SET8 = 54
+R_RISCV_SET16 = 55
+R_RISCV_SET32 = 56
+R_RISCV_32_PCREL = 57
+R_RISCV_PLT32 = 59
 R_XTENSA_PDIFF32 = 59
+R_RISCV_SET_ULEB128 = 60
+R_RISCV_SUB_ULEB128 = 61
+R_RISCV_TLSDESC_HI20 = 62
+R_RISCC_TLSDESC_LOAD_LO12 = 63
+R_RISCV_TLSDESC_ADD_LO12 = 64
+R_RISCV_TLSDESC_CALL = 65
 
 ################################################################################
 # Architecture configuration
@@ -128,6 +172,18 @@ def asm_jump_xtensa(entry):
     jump_offset = entry - 4
     jump_op = jump_offset << 6 | 6
     return struct.pack("<BH", jump_op & 0xFF, jump_op >> 8)
+
+
+def asm_jump_rv32(entry):
+    # This could be 6 bytes shorter, but the code currently cannot
+    # support a trampoline with varying length depending on the offset.
+
+    # auipc t6, HI(entry)
+    # jalr  zero, t6, LO(entry)
+    upper, lower = split_riscv_address(entry)
+    return struct.pack(
+        "<II", (upper | 0x00000F97) & 0xFFFFFFFF, ((lower << 20) | 0x000F8067) & 0xFFFFFFFF
+    )
 
 
 class ArchData:
@@ -199,6 +255,13 @@ ARCH_DATA = {
         asm_jump_xtensa,
         separate_rodata=True,
     ),
+    "rv32imc": ArchData(
+        "EM_RISCV",
+        MP_NATIVE_ARCH_RV32IMC << 2,
+        4,
+        (R_RISCV_32, R_RISCV_GOT_HI20, R_RISCV_GOT32_PCREL),
+        asm_jump_rv32,
+    ),
 }
 
 ################################################################################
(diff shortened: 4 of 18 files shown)


agatti:
For `features2` and `random` I have no idea at the moment on how to tweak the build process to get the soft-float code work on Picolibc in the CI environment.  `MICROPY_FLOAT_IMPL` is set to the appropriate value for RV32 natmods I think, and it works when building with Newlib on a different toolchain (the one provided by Arch linux in my case), so it should work for Picolibc too but it doesn't - unless the RISC-V toolchain shipped by Ubuntu 22.04 not only doesn't provide `__builtin_alloca` but has some issues with soft-float too?

Suggestions are welcome :)

agatti:
@dpgeorge I found an "interesting" bug, which I'm not sure if it is caused by my relocation handling.

This happens when you build the `examples/deflate` natmod for rv32imc and then push it to an ESP32C3.  If you take a look at https://gist.github.com/agatti/84476520c808095575321769dfe64138 you'll see it's `tests/extmod/deflate_decompress.py` with a modified module import to pick up the natmod and with most of the test code commented out except for the last dozen or so calls to the deflate module.

Now, if you run that file via `mpremote run` as-is, the device crashes when executing the code at line 194 or 195.  However, if you remove the chunk of code between lines 53 and 170 - *which is already commented out* - and run the test again, there's no crash at all.

More interestingly, if you leave the commented block in, it's only the sequence of having the last three marked blocks together that make the device crash.  With any other two blocks combination and the commented block it works.

This doesn't make much sense unless the patched relocations are incorrect and by accident something overwrites some structures in RAM, but I'm not entirely sure that's what happening, because as it's the same functions called over and over I suppose it would create issues right away.  I'm starting to doubt it's my device misbehaving but until the new C3s I've ordered arrive I cannot rule that out (all the boards I have are busy doing other things).

agatti:
The `framebuf` natmod crashes because `pyelftools` may have a bug in handling `R_RISCV_SUB32` relocations.

The `switch` statement in `framebuf_make_new` is compiled as a table whose entries contains two relocations each, one positive (ADD32) and one negative (SUB32).  In this case the positive relocation contains the offset between the table entry and the parameters calculation code (calculated from the beginning of the segment), and then the negative relocation contains the offset of the table start, to make sure the offset is relative to the `auipc/jalr` pair of instruction that performs the table jump.

According to [the official relocation descriptions](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#reloc-table), in both cases one must add or subtract the value of the relocation plus the addend to the original word present in the binary (in this case 0x00000000).  When performing the relocations for that switch table, ADD32 relocations have the correct `s["st_value"]`/`r_addend` values pair, however SUB32 relocations have 0 in `s["st_value"]` instead of the table offset:

```
# ADD32 relocation, st_value contains the offset of ".L250" (the table entry offset) as it is expected
R: <Relocation (RELA): Container({'r_offset': 0, 'r_info': 63523, 'r_info_sym': 248,
                                  'r_info_type': 35, 'r_addend': 0})> 
S: Container({'st_name': 1705, 'st_value': 4644, 'st_size': 0,
              'st_info': Container({'bind': 'STB_LOCAL', 'type': 'STT_NOTYPE'}),
              'st_other': Container({'visibility': 'STV_DEFAULT'}), 'st_shndx': 1}), '.L250'
  00001580 .L250
# SUB32 relocation, st_value is 0 instead of containing the offset of ".L245" (the table start offset)
R: <Relocation (RELA): Container({'r_offset': 0, 'r_info': 23591, 'r_info_sym': 92,
                                  'r_info_type': 39, 'r_addend': 0})> 
S: Container({'st_name': 951, 'st_value': 0, 'st_size': 0,
              'st_info': Container({'bind': 'STB_LOCAL', 'type': 'STT_NOTYPE'}),
              'st_other': Container({'visibility': 'STV_DEFAULT'}), 'st_shndx': 6}), '.L245'
  00001580 .L245
```

Once the relocations are applied and there's an attempt to create a framebuffer, the code tries to set up the parameters using a code fragment outside the natmod code space, and things crash...

agatti:
@dpgeorge here's a version that makes all tests pass for the natmods that build (`features0`, `features1`, `features3`, `features4`, `framebuf`, `heapq`, `re`) except for the ones in `deflate` and `framebuf_ellipse` and `framebuf_polygon`.  `re_debug`, `re_sub`, `re_sub_unmatched` need more build flags to include debug and substitute functionalities, and those were not in the `re` natmod makefile.

That said, for those failures I'm not sure they are due to my changes, as for example, in `framebuf_ellipse` it crashes when attempting to display the first clipped ellipse - however if you remove the rest of the test operations, those ellipses draw just fine.  Same goes for `deflate`, functions fail after being called in the middle of a test case, but if moved around or isolated they work fine.

I made all test verifications by hand (editing the tests' import lines to pick up the mpy version), hopefully when Qemu ports will run mpy files this can be added to CI as well.

agatti:
Can reproduce the same crashes on `xtensawin` and an ESP32-D0WD-V3.

dpgeorge:
Thanks for this, it's a great addition!

I haven't had time to look at it properly, but it will definitely need some CI tests:

> I made all test verifications by hand (editing the tests' import lines to pick up the mpy version), hopefully when Qemu ports will run mpy files this can be added to CI as well.

Yes, we should try to get the qemu ports to test this.  See #15624 for the first step in that direction.

vshymanskyy:
I can confirm that this PR works with [`wasm2mpy`](https://github.com/vshymanskyy/wasm2mpy).
We were even able to run the CoreMark tests:
```
ESP32-C3 160MHz: 179.791   <---
ESP32 240MHz: 228.363
ESP32-S3 240MHz: 271.573
BL616 320MHz: 344.293         <--- on Zephyr
```

examples/natmod/btree/Makefile
```diff
@@ -32,6 +32,8 @@ SRC += $(addprefix $(realpath $(BTREE_DIR))/,\
 	mpool/mpool.c \
 	)
 
+CLEAN_EXTRA = "$(MOD).mpy"
```
dpgeorge, line 35: This is not really related to this PR.  Please put it in a separate commit, at least.

And maybe it can go in `py/dynruntime.mk` so it doesn't need to be repeated for each example?

tools/mpy_ld.py
```diff
@@ -590,40 +670,75 @@ def do_relocation_text(env, text_addr, r):
...
-    elif reloc_type == "xtensa_l32r":
-        l32r = unpack_u24le(env.full_text, r_offset)
-        assert l32r & 0xF == 1  # RI16 encoded l32r
-        l32r_imm16 = l32r >> 8
-        l32r_imm16 = (l32r_imm16 + reloc >> 2) & 0xFFFF
-        l32r = l32r & 0xFF | l32r_imm16 << 8
-        pack_u24le(env.full_text, r_offset, l32r)
-    else:
-        assert 0, reloc_type
+    if env.arch.name != "EM_RISCV":
```
dpgeorge, line 708: To minimise diff and make it easier to change in the future, I suggest:
```py
if env.arch.name == "EM_RISCV":
    # (comment why we don't do anything here)
    pass
elif reloc_type == "le32":
    ...
```

tools/mpy_ld.py
```diff
@@ -590,40 +670,75 @@ def do_relocation_text(env, text_addr, r):
...
     if log_name is None:
         if s_type == "STT_SECTION":
             log_name = s.section.name
         else:
             log_name = s.name
-    log(LOG_LEVEL_3, "  {:08x} {} -> {:08x}".format(r_offset, log_name, addr))
+    if addr is not None:
+        log(LOG_LEVEL_3, "  {:08x} {} -> {:08x}".format(r_offset, log_name, addr))
+    else:
+        log(LOG_LEVEL_3, "  {:08x} {} {:08x}".format(r_offset, log_name, value))
```
dpgeorge, line 741: Maybe `"  {:08x} {} == {:08x}"`?

tools/mpy_ld.py
```diff
@@ -684,6 +803,276 @@ def do_relocation_data(env, text_addr, r):
...
+    R_RISCV_PCREL_LO12_I: "riscv_lo12i",
+    R_RISCV_PCREL_LO12_S: "riscv_lo12s",
+    R_RISCV_LO12_I: "riscv_lo12i",
+    R_RISCV_LO12_S: "riscv_lo12s",
+    R_RISCV_32_PCREL: "riscv_32pcrel",
+    R_RISCV_PLT32: "riscv_32pcrel",
+}
+
+
+def process_riscv32_relocation(env, text_addr, r):
```
dpgeorge, line 835: Does it make sense to split this function into "compute" and "rewrite" parts?  That might follow better the existing logic above (where this function is called) used by other archs.

Alternatively, maybe it's better to have a single function like this for all the other archs?  Although I think in their case the different archs share a lot of the rewrite logic.

agatti: It's been a while, but the only reason was to keep the method length/complexity below the threshold set in the precommit checks running on python code.

As you point out, other archs more or less share the same relocation handling primitives when it comes to relocation handling with RISC-V being the outlier here (and MIPS and probably AArch64 too).  I'm OK with doing the split myself (either with a set of functions or classes with just enough inheritance to reuse even more code), whether that is submitted as its own PR or as an entry in this set of commits is fine - whichever way takes less effort for you to review.

dpgeorge: I'm happy to leave this refactoring for another day (ie keep it as you have it).

This one-file program is getting quite large so probably needs some more serious factoring, eventually.

tools/mpy_ld.py
agatti: This should be handled in the latest revision of this PR.  Thanks for the suggestion!

examples/natmod/btree/Makefile
agatti: Good point!  The change in question is now limited to `py/dynruntime.mk` and was moved to its own commit.

tools/mpy_ld.py
agatti: Makes sense, I've updated the PR to reflect this suggestion.  Thanks!

dpgeorge:
@agatti this PR is looking really good now, and I've tested it with the natmod examples.  Everything passes locally for me, except compiling `features2` (because it needs float) and running `extmod/random_extra_float.py` (because it needs float).

As the remaining piece of the puzzle, I think we should get CI to run these tests.  Here's a minimal patch that works locally for me and hopefully works for the CI.  If you could add it to this PR, that would be great (feel free to modify to get it working and/or improve it as you see fit):
```diff
diff --git a/ports/qemu/Makefile b/ports/qemu/Makefile
index 237e1e951..f60e8f58b 100644
--- a/ports/qemu/Makefile
+++ b/ports/qemu/Makefile
@@ -168,6 +168,11 @@ test: $(BUILD)/firmware.elf
        $(eval DIRNAME=ports/$(notdir $(CURDIR)))
        cd $(TOP)/tests && ./run-tests.py -t execpty:"$(QEMU_SYSTEM) $(QEMU_ARGS) -serial pty -kernel ../$(DIRNAME)/$<" $(RUN_TESTS_ARGS) $(RUN_TESTS_EXTRA)
 
+.PHONY: test_natmod
+test_natmod: $(BUILD)/firmware.elf
+       $(eval DIRNAME=ports/$(notdir $(CURDIR)))
+       cd $(TOP)/tests && ./run-natmodtests.py -p -d execpty:"$(QEMU_SYSTEM) $(QEMU_ARGS) -serial pty -kernel ../$(DIRNAME)/$<" $(RUN_NATMODTESTS_ARGS) extmod/{btree,deflate,framebuf,heapq,random_basic,re}*.py
+
 $(BUILD)/firmware.elf: $(LDSCRIPT) $(OBJ)
        $(Q)$(CC) $(LDFLAGS) -o $@ $(OBJ) $(LIBS)
        $(Q)$(SIZE) $@
diff --git a/ports/qemu/boards/VIRT_RV32.mk b/ports/qemu/boards/VIRT_RV32.mk
index a61b659fa..355a09c3d 100644
--- a/ports/qemu/boards/VIRT_RV32.mk
+++ b/ports/qemu/boards/VIRT_RV32.mk
@@ -12,3 +12,5 @@ MPY_CROSS_FLAGS += -march=rv32imc
 
 # These Thumb tests don't run on RV32, so exclude them.
 RUN_TESTS_ARGS = --exclude 'inlineasm|qemu/asm_test'
+
+RUN_NATMODTESTS_ARGS = --arch rv32imc
diff --git a/tools/ci.sh b/tools/ci.sh
index c67aeed0c..63ffe58b8 100755
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -292,6 +292,10 @@ function ci_qemu_build_rv32 {
     make ${MAKEOPTS} -C mpy-cross
     make ${MAKEOPTS} -C ports/qemu BOARD=VIRT_RV32 submodules
     make ${MAKEOPTS} -C ports/qemu BOARD=VIRT_RV32 test
+
+    # Test building and running native .mpy with rv32imc architecture.
+    ci_native_mpy_modules_build rv32imc
+    make ${MAKEOPTS} -C ports/qemu BOARD=VIRT_RV32 test_natmod
 }
 
 ########################################################################################
@@ -476,7 +480,9 @@ function ci_native_mpy_modules_build {
         arch=$1
     fi
     make -C examples/natmod/features1 ARCH=$arch
-    make -C examples/natmod/features2 ARCH=$arch
+    if [ $arch != rv32imc ]; then
+        make -C examples/natmod/features2 ARCH=$arch
+    fi
     make -C examples/natmod/features3 ARCH=$arch
     make -C examples/natmod/features4 ARCH=$arch
     make -C examples/natmod/btree ARCH=$arch
```

agatti:
> Everything passes locally for me, except compiling `features2` (because it needs float) and running `extmod/random_extra_float.py` (because it needs float).

If you merge #15838 first then RV32 natmods requiring floating point support should compile and work.  I tried that back in September, but I don't recall if I had to change makefiles or any other file.

> As the remaining piece of the puzzle, I think we should get CI to run these tests. Here's a minimal patch that works locally for me and hopefully works for the CI. If you could add it to this PR, that would be great (feel free to modify to get it working and/or improve it as you see fit):

Thanks, I'll look at it shortly and update the PR accordingly!

dpgeorge:
> I guess you've tested building modules using newlib outside the CI?

I'm using the `riscv32-embecosm-ubuntu2204-gcc13.2.0` toolchain (used for the Pico 2) and I guess that uses newlib?  `btree` builds and runs for me using that toolchain so I guess it must be newlib.

> if that's OK with you I'll let the CI skip the `btree` natmod when testing RV32 natmod builds

Yes that's totally fine.

> I can try to come up with a PR later for MicroPython's berkeley-db fork to work around this if possible.

I'm not sure if that's worth the time/effort.  We should instead improve `mpy_ld.py` so that it can find the `errno` variable (but that's not a priority right now).

------

My main aim here is to just get some form of CI working that tests the functionality of this PR.  Compiling and running just a few of the `examples/natmod/` modules is enough for that.

agatti:
> I'm using the `riscv32-embecosm-ubuntu2204-gcc13.2.0` toolchain (used for the Pico 2) and I guess that uses newlib? 

So far the only picolibc-only RV32/RV64 toolchain I've found is the one shipping with Ubuntu 22.04 (and possibly 24.04, haven't yet looked at it) - which made adding QEMU RISC-V support so much of a prolonged effort...

> I'm not sure if that's worth the time/effort. We should instead improve `mpy_ld.py` so that it can find the `errno` variable (but that's not a priority right now).

The thing is, the CI toolchain uses a picolibc version that's explicitly set up with global errno variable support disabled.  That is, errno is instead a thread-local variable and that is thus compiled as a static uninitialised variable.

I've had a go at adding support for static initialised data to `mpy_ld.py` and it sort of worked except for ArmV6 bringing in absolute relocations into the mix.  I believe this is the same issue but for a different segment (.DATA vs .BSS) - if GCC for Arm isn't sneaking in absolute relocations for this kind of variables then I may probably be able to add a patch for that, otherwise things may require a more coordinated effort I guess.

Still, thanks for looking at this PR :)

dpgeorge:
OK, sounds like there are a few things to improve here.  But let's just take it a step at a time :smile: That makes it much easier to review and get things merged.

Adding a bit of CI testing to this PR, it should then be good to go.

agatti:
Now the CI should work for RV32.  In theory this could work for Arm too but my Arm dev VM needs to be brought back up and I'm on a metered connection right now, so I limited my changes to make sure RV32 works.

Regarding the `mpy_ld.py` changes, I guess TLS support for RV32 can be added when doing the per-arch processing split.  It also requires having init jump code sequences that may change their length after the first evaluation (and changes to QEMU's RV32 platform code), so it's a whole new PR indeed.

Glad all of this works now!

ports/qemu/README.md
```diff
@@ -110,3 +122,5 @@ The following options can be specified on the `make` command line:
 - `QEMU_DEBUG_ARGS`: defaults to `-s` (gdb on TCP port 1234), but can be overridden
   with different qemu gdb arguments.
 - `QEMU_DEBUG_EXTRA`: extra options to pass to qemu when `QEMU_DEBUG=1` is used.
+- `RUN_NATMODTESTS_ARGS`: pass in extra flags for `run-natmodtests.py` when invoked
+  via `make test_natmod`.
```
dpgeorge, line 126: I'm not sure if `RUN_NATMODTESTS_ARGS` should be exposed to the user as a public-facing variable.  It's set by the board to configure things like the arch.

If we want to give control to the user, it should probably be via a `RUN_NATMODTESTS_EXTRA` variable.  And maybe that should default to `extmod/{deflate,framebuf,heapq,random_basic,re}*.py` for the `VIRT_RV32` board (with the idea to add more tests for ARM-based boards in the future).

tools/ci.sh
```diff
@@ -476,10 +481,14 @@ function ci_native_mpy_modules_build {
         arch=$1
     fi
     make -C examples/natmod/features1 ARCH=$arch
-    make -C examples/natmod/features2 ARCH=$arch
+    if [ $arch != rv32imc ]; then
+        # This requires soft-float support.
```
dpgeorge, line 485: Maybe make it clear that it's rv32imc that needs soft-float support, eg "This requires soft-float support on rv32imc.".

Same comment with the TLS line below, eg "This requires thread local storage support on rv32imc.".

ports/qemu/README.md
agatti: Valid point.  I opted to remove the entry from the documentation instead of adding another user-facing variable as that, if needed, can be done when enabling natmod testing on Arm.

tools/ci.sh
agatti: I found it rather clear already since the comments in question are inside an explicit rv32imc conditional block, but more clarity never hurts.  I've updated the comments as such, thanks for the suggestion.

dpgeorge (approved):

dpgeorge:
Thanks for updating!  Now merged.

PR #15604, opened 2024-08-05
Update pyboard.py fixing get_time() bug

rufusclark:
### Summary

The current script evaluates "pyb.RTC().datetime()" resulting in a remote side exception being called as pyb is not defined.

The script should evaluate "machine.RTC().datetime()" and hence return the current time.

This reflects the API in the docs, https://docs.micropython.org/en/latest/library/time.html#module-time

### Testing

Tested with the Raspberry Pi Pico 4MB

### DIFF
--- a/tools/pyboard.py
+++ b/tools/pyboard.py
@@ -506,7 +506,7 @@ class Pyboard:
         return self.exec_(pyfile)
 
     def get_time(self):
-        t = str(self.eval("pyb.RTC().datetime()"), encoding="utf8")[1:-1].split(", ")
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thanks for the contribution.

`pyboard.py` is mostly deprecated now, in favour of `mpremote`.  And to access the RTC you can do:
```
mpremote rtc
mpremote rtc --set
```

Regardless of that, this PR is still valid and is a good fix, because stm32 does also have `machine.RTC` so it still works on that port.

PR #15605, opened 2024-08-06
py,esp32,rp2: Add new cstack API, check margin macro, fix esp32 stack check.

projectgus:
### Summary

* Motivation for this was testing the ESP32-C3 bug fix in https://github.com/micropython/micropython/pull/15583/ and realising there is a subtle race condition on esp32 port where the stack check limit may not be set reliably, depending on function timing.
* On looking at the whole situation, every port adds some "stack limit margin" but it's usually hard-coded in the call to `mp_stack_set_limit()`. Threaded ports have the extra challenge of adding a margin in the thread module. However only the value passed to `mp_stack_set_limit()` needs to have the margin applied (i.e. if the user sets a thread stack size of N bytes, the tasks will have a real stack size of N bytes and the stack limit checker sets its value to `N - margin` bytes.)
* After discussing with Damien this PR has expanded to add both a `MICROPY_STACK_CHECK_MARGIN` macro and a new cstack API which sets the stack origin and the stack limit (based off the stack size) in one call.
* To offset the code size impact on ports without threads, new cstack API can be inlined in some cases.
* Old stackctrl.h API is kept unchanged, apart from a note about deprecation and macros to disable when building `MICROPY_PREVIEW_VERSION_2`.
* `esp32`, `unix` and `rp2` ports are converted in this PR as initial examples.

### Testing

* Ran `stress` and `thread` tests on unix port, esp32 port (S3 and C3), and on rp2 port.
* A commit in this PR enables the `stress` tests for the esp32 port.

Could have made a much smaller fix for the esp32 port race condition by adjusting the stack size before creating the thread, or by applying a mutex. However this approach seems easier to understand overall, and implements a single source of truth for the stack limit margin.
 
### Follow-up work

* Refactor other ports to use the new API.
* Suspect the stack limitmargin is set quite conservatively (especially for ESP32-C3 where it is 2KB). Could experiment with reducing this to save some RAM.

### DIFF
--- a/ports/rp2/mpconfigport.h
+++ b/ports/rp2/mpconfigport.h
@@ -81,6 +81,7 @@
 #define MICROPY_TRACKED_ALLOC                   (MICROPY_SSL_MBEDTLS || MICROPY_BLUETOOTH_BTSTACK)
 #define MICROPY_READER_VFS                      (1)
 #define MICROPY_ENABLE_GC                       (1)
+#define MICROPY_STACK_CHECK_MARGIN              (256)
 #define MICROPY_ENABLE_EMERGENCY_EXCEPTION_BUF  (1)
 #define MICROPY_LONGINT_IMPL                    (MICROPY_LONGINT_IMPL_MPZ)
 #define MICROPY_FLOAT_IMPL                      (MICROPY_FLOAT_IMPL_FLOAT)
--- a/py/cstack.c
+++ b/py/cstack.c
@@ -4,6 +4,7 @@
  * The MIT License (MIT)
  *
  * Copyright (c) 2014 Paul Sokolovsky
+ * Copryight (c) 2024 Angus Gratton
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -24,31 +25,22 @@
  * THE SOFTWARE.
  */
 
-// This API is deprecated, please use py/cstack.h instead
-
 #include "py/runtime.h"
+#include "py/cstack.h"
 
-#if !MICROPY_PREVIEW_VERSION_2
-
-#include "py/stackctrl.h"
-
-void mp_stack_ctrl_init(void) {
+void mp_cstack_init_with_sp_here(size_t stack_size) {
     #if __GNUC__ >= 13
     #pragma GCC diagnostic push
     #pragma GCC diagnostic ignored "-Wdangling-pointer"
     #endif
     volatile int stack_dummy;
-    MP_STATE_THREAD(stack_top) = (char *)&stack_dummy;
+    mp_cstack_init_with_top((void *)&stack_dummy, stack_size);
     #if __GNUC__ >= 13
     #pragma GCC diagnostic pop
     #endif
 }
 
-void mp_stack_set_top(void *top) {
-    MP_STATE_THREAD(stack_top) = top;
-}
-
-mp_uint_t mp_stack_usage(void) {
+mp_uint_t mp_cstack_usage(void) {
     // Assumes descending stack
     volatile int stack_dummy;
     return MP_STATE_THREAD(stack_top) - (char *)&stack_dummy;
@@ -56,16 +48,10 @@ mp_uint_t mp_stack_usage(void) {
 
 #if MICROPY_STACK_CHECK
 
-void mp_stack_set_limit(mp_uint_t limit) {
-    MP_STATE_THREAD(stack_limit) = limit;
-}
-
-void mp_stack_check(void) {
-    if (mp_stack_usage() >= MP_STATE_THREAD(stack_limit)) {
+void mp_cstack_check(void) {
+    if (mp_cstack_usage() >= MP_STATE_THREAD(stack_limit)) {
         mp_raise_recursion_depth();
     }
 }
 
 #endif // MICROPY_STACK_CHECK
-
-#endif // !MICROPY_PREVIEW_VERSION_2
--- a/py/cstack.h
+++ b/py/cstack.h
@@ -4,6 +4,7 @@
  * The MIT License (MIT)
  *
  * Copyright (c) 2014 Paul Sokolovsky
+ * Copyright (c) 2024 Angus Gratton
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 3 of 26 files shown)


py/stackctrl.h
```diff
@@ -26,15 +26,20 @@
...
+#include "py/mpstate.h"
 
 void mp_stack_ctrl_init(void);
 void mp_stack_set_top(void *top);
 mp_uint_t mp_stack_usage(void);
 
 #if MICROPY_STACK_CHECK
 
-void mp_stack_set_limit(mp_uint_t limit);
+inline static void mp_stack_set_limit(mp_uint_t limit) {
```
dpgeorge, line 37: This is no longer setting the limit, instead it's setting the total available stack space.  The limit is then that number minus the margin.

So I suggest renaming this function to something like `mp_stack_set_size()`.  If we do that, it would force users to notice this change in API.  OTOH maybe it's nicer for them to provide a backwards compatible macro `#define mp_stack_set_limit mp_stack_set_size`?

py/stackctrl.h
```diff
@@ -26,15 +26,20 @@
...
+    assert(limit > MICROPY_STACK_CHECK_MARGIN); // Should be enforced by port
```
dpgeorge, line 38: Suggest using `MP_STATIC_ASSERT` (if possible here...).

dpgeorge:
An alternative to this would be to still have the `MICROPY_STACK_CHECK_MARGIN` macro but use it explicitly at the call site for `mp_stack_set_limit()`.  Eg in `mp_thread_init_state()` it would do:
```c
    mp_stack_set_limit(stack_size - MICROPY_STACK_CHECK_MARGIN);
```
That makes it more obvious that the limit is the stack size minus an explicit margin.  Otherwise the margin is kind of hidden in the `stackctrl.h` implementation.

But, this alternative approach is more verbose (more duplicated code with the subtraction of the margin everywhere) so probably not any better than what this PR does.

py/stackctrl.h
projectgus: Good point... what about `mp_stack_check_init(mp_uint_t stack_size)`. Then the name still indicates the purpose, but more accurate about what it will do.

py/stackctrl.h
projectgus: Not possible here, limit is not guaranteed to be compile-time constant.

However the compiler will be smart, if assertions are enabled and limit *is* constant then the assert will be elided if it passes, and if it fails then dead code elimination will conveniently remove everything that comes after. :laughing:

py/stackctrl.h
dpgeorge: I think `mp_stack_check_init()` is confusing because there's already `mp_stack_ctrl_init()`... with two init functions it's confusing which one to call.

The current situation is that you either do one of the following to initialise stack checking:
- call `mp_stack_ctrl_init()` to use the current stack pointer to set the top of the stack, then call `mp_stack_set_limit()` to set the limit from there
- call `mp_stack_set_top()` to set an explicit top pointer, then call `mp_stack_set_limit()`

Then you can call `mp_stack_usage()` and `mp_stack_check()`.

If we are going to rejig the API, I'd propose to have two separate init functions:
- `mp_stack_check_init_with_sp_here(size_t stack_size)`
- `mp_stack_check_init_with_top(void *top, size_t stack_size)`

(Note the `size_t` type.)

Then it's much clearer what to call, and you only need to call one function to init it.

py/stackctrl.h
```diff
@@ -26,15 +26,20 @@
 #ifndef MICROPY_INCLUDED_PY_STACKCTRL_H
 #define MICROPY_INCLUDED_PY_STACKCTRL_H
 
-#include "py/mpconfig.h"
+#include "py/mpstate.h"
```
dpgeorge, line 29: I was a bit worried this would introduce a circular header dependency, but luckily it does not.

So, now it would also be possible to make `mp_stack_set_top()` static-inline, which should also help to reduce code size.

py/stackctrl.h
dpgeorge: Aah, right.

py/stackctrl.h
projectgus: Have pushed a commit here to do this as well.

projectgus:
@dpgeorge Have updated according to our discussion earlier, and updated the PR description.

The project coverage check is failing because stackctrl.c is no longer covered. What's the best way to deal with that?

dpgeorge:
> The project coverage check is failing because stackctrl.c is no longer covered. What's the best way to deal with that?

Hmmmm... three options I can think of:
1. Add code to `ports/unix/coverage.c` to call the old functions to get coverage.
2. Add a `#if MICROPY_STACK_CHECK_LEGACY` around everything in `py/stackctrl.c` and disable this on unix coverage builds, to exclude that code from coverage checks.
3. Implement the new functions in terms of the old ones.

Option 3 makes it harder to easily remove the old functions when the time comes.  Probably option 1 is the best.

projectgus:
> Option 3 makes it harder to easily remove the old functions when the time comes. Probably option 1 is the best.

Thanks, have push a commit to try that approach (and coverage seems happy, at least.)

ports/rp2/mpconfigport.h
```diff
@@ -158,6 +158,8 @@
 // Number 3 is currently used by pico-sdk (PICO_TIME_DEFAULT_ALARM_POOL_HARDWARE_ALARM_NUM)
 #define MICROPY_HW_SOFT_TIMER_ALARM_NUM         (2)
 
+#define MICROPY_STACK_CHECK_MARGIN              (256)
```
dpgeorge, line 161: Please move this config option to where it is relative to how it's ordered in `py/mpconfig.h`.

That would mean placing it between `MICROPY_ENABLE_GC` and `MICROPY_ENABLE_EMERGENCY_EXCEPTION_BUF`.

py/cstack.c
```diff
@@ -0,0 +1,56 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2014 Paul Sokolovsky, 2024 Angus Gratton
```
dpgeorge, line 6: The convention in this project is to put separate authors' copyright on separate lines, ie:
```c
* Copyright (c) 2014 Paul Sokolovsky
* Copyright (c) 2024 Angus Gratton
```

py/cstack.h
```diff
@@ -0,0 +1,62 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2014 Paul Sokolovsky, 2024 Angus Gratton
```
dpgeorge, line 6: separate lines, as above

py/cstack.h
```diff
@@ -0,0 +1,62 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#ifndef MICROPY_INCLUDED_PY_CSTACK_H
+#define MICROPY_INCLUDED_PY_CSTACK_H
+
+#include "py/mpstate.h"
+
+// Both init functions accept the full stack size. Set the
```
dpgeorge, line 31: "Both init functions below accept..."

py/stackctrl.c
```diff
@@ -25,6 +25,11 @@
  */
 
 #include "py/runtime.h"
+
+// This API is deprecated, please use py/cstack.h instead
```
dpgeorge, line 29: Suggest putting this comment above the include of py/runtime.h, to mirror `stackctrl.h`.

dpgeorge:
This is looking really good now!  Just a few minor comments.

ports/rp2/mpconfigport.h
projectgus: Done. Do you want the same thing for esp32 mpconfigport.h? It would mean splitting up a block of defines with an ifdef.

dpgeorge: Yes, please move it.  Consistency is better here.

projectgus: Done!

dpgeorge:

dpgeorge:
@projectgus looks like this may have broken `multi_bluetooth/ble_irq_calls.py` on ESP32-C3.  That test tests for recursion within a BLE IRQ handler, and now it seems that ESP32-C3 has less stack for Python calls because of the increased margin, so the test fails.

The BLE IRQ handler on esp32 has `CONFIG_BT_NIMBLE_TASK_STACK_SIZE=6144` minus 2048 minus `MICROPY_STACK_CHECK_MARGIN` bytes.  So that's very small on ESP32-C3/RISC-V.

To fix it I suggest decreasing `MICROPY_STACK_CHECK_MARGIN` on RISC-V to a less conservative value.  If that still doesn't fix `multi_bluetooth/ble_irq_calls.py` then we should probably increase `CONFIG_BT_NIMBLE_TASK_STACK_SIZE` on RISC-V.

PR #12131, opened 2023-07-31
py: Prevent overflowing size_t when calling mp_seq_multiply.

FelixNumworks:
Previously, it could led to a situation where `m_new(type, BIG_NUMBER)` would call `m_alloc(zero)` because `sizeof(type)*BIG_NUMBER` would return `0`. This change ensures that a MemoryError is raised instead.

EDIT:
The inputs [0]*2\*\*30 or "blue"*2\*\*30 could create a situation where size_t was overflowed and malloc was called with a corrupted value that would then lead to a bad access in memory.

PR #15606, opened 2024-08-06
py/misc.h: prevent m_new from overflowing size_t when multiplying type size and number of elements

FelixNumworks:
### Summary

Previously, it could led to a situation where m_new(type, BIG_NUMBER) would call m_alloc(zero) because sizeof(type)*BIG_NUMBER would return 0. This change ensures that a MemoryError is raised.

This is an alternative to [this pull request](https://github.com/micropython/micropython/pull/12131)

See discussion in the [other pull request](https://github.com/micropython/micropython/pull/12131)

### DIFF
--- a/py/misc.h
+++ b/py/misc.h
@@ -69,24 +69,30 @@ typedef unsigned int uint;
 
 /** memory allocation ******************************************/
 
+
+// Ensure that m_new does not overflow the multiplication when computing sizeof(type) * num.
+#define CAPPED_SIZE_NUM_OBJ(type, num) ((sizeof(type) * (num)) / sizeof(type) != (size_t)(num) ? SIZE_MAX : sizeof(type) * (num))
+// Ensure that the multiplication and the addition do not overflow.
+#define CAPPED_SIZE_OBJ_NUM_VAR(obj_type, var_type, var_num) ((sizeof(var_type) * (var_num)) / sizeof(var_type) != (size_t)(var_num) || sizeof(obj_type) + sizeof(var_type) * (var_num) < sizeof(var_type) * (var_num) ? SIZE_MAX : sizeof(obj_type) + sizeof(var_type) * (var_num))
+
 // TODO make a lazy m_renew that can increase by a smaller amount than requested (but by at least 1 more element)
 
-#define m_new(type, num) ((type *)(m_malloc(sizeof(type) * (num))))
-#define m_new_maybe(type, num) ((type *)(m_malloc_maybe(sizeof(type) * (num))))
-#define m_new0(type, num) ((type *)(m_malloc0(sizeof(type) * (num))))
+#define m_new(type, num) ((type *)(m_malloc(CAPPED_SIZE_NUM_OBJ(type, num))))
+#define m_new_maybe(type, num) ((type *)(m_malloc_maybe(CAPPED_SIZE_NUM_OBJ(type, num))))
+#define m_new0(type, num) ((type *)(m_malloc0(CAPPED_SIZE_NUM_OBJ(type, num))))
 #define m_new_obj(type) (m_new(type, 1))
 #define m_new_obj_maybe(type) (m_new_maybe(type, 1))
-#define m_new_obj_var(obj_type, var_field, var_type, var_num) ((obj_type *)m_malloc(offsetof(obj_type, var_field) + sizeof(var_type) * (var_num)))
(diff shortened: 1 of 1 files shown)


projectgus:
You've done a good job of finding all the places where a multiply could overflow! As @dpgeorge mentioned in the other PR, this kind of approach would still need a lot of work to get the code size low enough (and maybe even then it won't be low enough).

FWIW I would suggest a possible approach could look like this:

1) Implement a `static inline bool mp_mul_overflow(size_t a, size_t b, size_t *res)`[*] function in `py/misc.h`. For gcc and Clang this can be a thin wrapper around `__builtin_mul_overflow`. However it will need a fallback for MSVC and other compilers. Suggest a good fallback would be to cast both args to uint64_t, multiply as 64-bit, and then check if the result is greater than UINT32_MAX. A smart compiler will hopefully find a way to optimise that to avoid the 64-bit multiplies, too.

2) Implement another non-inline function like `size_t mp_checked_mul(size_t a, size_t b)` that multiplies two values and raises OverflowError if they overflow. (Raising OverflowError is what CPython does in this case.) Implement mp_checked_mul() using the macro from step 1.

3) Change all of these vulnerable multiplies to be calls to `mp_checked_mul()`. I think changing each multiply to a non-inlined function call has the least possible code size impact (at best, a single Thumb multiply instruction becomes the function call branch so no size impact. At worst, each instance adds 3 mov instructions and a literal load so 1 instruction becomes 4, +6 bytes each time.)  

3) If the code size impact is still significant (probably it will be) then add a macro like `MICROPY_HEAP_OVERFLOW_CHECKS`. If this is set to 0, `mp_checked_mul` is implemented as an inline `return a * b;`. That means this check can be optionally turned off to save code size.

This is a significant amount of work, and I can't promise the result would definitely be merged, so it's obviously up to you if you're interested to do something like this or not. However, I'd like to help you if you're keen. Let me know if you have any other questions. 

[*] size_t may not be right type for these, maybe uint32_t or mp_uint_t or something else.

projectgus:
Oh, and some unit tests that trigger the overflow cases for code coverage.

py/misc.h
```diff
@@ -69,24 +69,30 @@ typedef unsigned int uint;
...
 #if MICROPY_MALLOC_USES_ALLOCATED_SIZE
-#define m_renew(type, ptr, old_num, new_num) ((type *)(m_realloc((ptr), sizeof(type) * (old_num), sizeof(type) * (new_num))))
-#define m_renew_maybe(type, ptr, old_num, new_num, allow_move) ((type *)(m_realloc_maybe((ptr), sizeof(type) * (old_num), sizeof(type) * (new_num), (allow_move))))
-#define m_del(type, ptr, num) m_free(ptr, sizeof(type) * (num))
-#define m_del_var(obj_type, var_field, var_type, var_num, ptr) (m_free(ptr, offsetof(obj_type, var_field) + sizeof(var_type) * (var_num)))
+#define m_renew(type, ptr, old_num, new_num) ((type *)(m_realloc((ptr), CAPPED_SIZE_NUM_OBJ(type, old_num), CAPPED_SIZE_NUM_OBJ(type, new_num))))
+#define m_renew_maybe(type, ptr, old_num, new_num, allow_move) ((type *)(m_realloc_maybe((ptr), CAPPED_SIZE_NUM_OBJ(type, old_num), CAPPED_SIZE_NUM_OBJ(type, new_num), (allow_move))))
+#define m_del(type, ptr, num) m_free(ptr,  CAPPED_SIZE_NUM_OBJ(type, num))
+#define m_del_var(obj_type, var_field, var_type, var_num, ptr) (m_free(ptr, CAPPED_SIZE_OBJ_NUM_VAR(offsetof(obj_type, var_field), var_type, var_num)))
```
dpgeorge, line 92: Probably the free's don't need checking, because if it already succeeded in allocation the multiply should be the same for the free (and should definitely be within the available address space of the architecture).

dpgeorge:
In addition to what @projectgus says above, there may be other ways to minimise the impact of this change.  For example:
- Instead of `mp_checked_mul()` raising `OverflowError` it could return a magic number like -1 to indicate overflow, and then the `m_malloc` etc functions could check for this magic number and raise `OverflowError` there.
- It looks like there are only a handful of allocation functions that are involved here, so maybe they can all be extended to take arguments like `calloc`: a number and a size and then they do the multiplication and overflow checking themselves (at the start of `m_malloc`, `m_malloc0`, `m_renew` etc).

I think it's worth doing this, but we need to find an approach which has minimal impact in both code size and performance (less so for performance, because doing an allocation is already costly so these code paths are not usually in a performance-critical path).

Regardless, we'd still want to have this feature optional via a config macro.

PR #15609, opened 2024-08-07
qemu-arm/qemu-riscv: improve tinytest runner so it doesn't need to generate .exp files on disk

dpgeorge:
### Summary

The qemu-arm and qemu-riscv ports generate a whole lot of .exp files in the `tests/` directory when you run the tests on these ports.  This leaves the repo in an unclean state.

This PR improves this situation by:
- generating the .exp files from within the `tools/tinytest-codegen.py` script by calling out to Python to run them and collecting the resulting stdout output
- then putting the output into a big dictionary of results
- then generating the resulting `build/genhdr/tests.h` file

This results in a clean repo after running the tests.  It also means that `run-tests.py` no longer needs a special case to handle the allowed tests for qemu-arm/riscv targets.

A bonus of this approach is that it's not much extra work to enable running the tests on qemu-*  port using the via .mpy and via native emitter test framework that already exists in `run-tests.py`.

Note: this PR builds on #15577.

TODO:
- [ ] test zephyr changes

### Testing

The qemu-arm tests pass like they did before, but no stray .exp files are generated.

It's a little slower this way when generating the .exp files but not by much.

An alternative to this approach would be to make the qemu-* ports have a stdio to the outside world for the REPL, then treat the qemu process as a board with a serial port, and run the test suite as it would normally be run for an attached hardware board.


dpgeorge:
Closing in favour of #15624.

Issue #15408, opened 2024-07-04
esp32: Enabling verbose debugging triggers a crash at boot.


agatti:
### Port, board and/or hardware

esp32 port, generic ESP32C3 board

### MicroPython version

Without the changes to `ports/esp32/mpconfigport.h` mentioned later the banner is:

`MicroPython v1.24.0-preview.96.g92484d882 on 2024-07-04; ESP32C3 module with ESP32C3`

### Reproduction

Starting from a fresh git checkout from the master branch, enable both `MICROPY_DEBUG_PRINTERS` and `MICROPY_DEBUG_VERBOSE` in `ports/esp32/mpconfigport.h`, then build the ESP32 port for an `ESP32_GENERIC_C3` board with ESP-IDF 5.0.4 (*Edit* - this also occurs on 5.0.5, 5.1.2, and 5.2.0).

### Expected behaviour

The board should show a regular REPL prompt on boot.

### Observed behaviour

The device enters into a boot loop with `RTC_SW_CPU_RST` as the reset cause:

```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x1 (POWERON),boot:0xf (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5820,len:0xe8c
load:0x403cc710,len:0x6ec
load:0x403ce710,len:0x2b24
entry 0x403cc710
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0xc (RTC_SW_CPU_RST),boot:0xf (SPI_FAST_FLASH_BOOT)
Saved PC:0x403807a8
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5820,len:0xe8c
load:0x403cc710,len:0x6ec
load:0x403ce710,len:0x2b24
entry 0x403cc710
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0xc (RTC_SW_CPU_RST),boot:0xf (SPI_FAST_FLASH_BOOT)
Saved PC:0x403807a8
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5820,len:0xe8c
load:0x403cc710,len:0x6ec
load:0x403ce710,len:0x2b24
entry 0x403cc710
# Keeps printing boot messages
```

[...]

PR #15634, opened 2024-08-10
esp32/mphalport: Print debug strings even before the GIL is ready.

agatti:
### Summary

If verbose debugging is enabled there is some stdout output happening before the GIL is ready, and the code assumed that no string printing occurred before the interpreter was fully initialised.  Printing long strings would operate without holding the GIL, which would crash if string output would happen too early in the startup process.

This commit addresses that issue, making sure verbose debugging output will work even before the interpreter is fully initialised (as if it is not yet ready there's no GIL to take care of).

Also, the threshold that would indicate whether a string is "long" (and thus requiring a GIL release/lock operation) or not was hardcoded to 20 bytes.  This commit makes that configurable, maintaining 20 bytes as a default.

This fixes bug #15408.

### Testing

Testing was done with an ESP32C3 board, with both `MICROPY_DEBUG_PRINTERS` and `MICROPY_DEBUG_VERBOSE` enabled.

This patch adds around 32 bytes of code, but only when `MICROPY_DEBUG_PRINTERS`, `MICROPY_DEBUG_VERBOSE`, and `MICROPY_PY_THREAD_GIL` are enabled.  Given that the first two flags are off by default, and enabling them means printing extra debugging information, a minor code size increase and even minor slowdown when printing strings is not only tolerated but expected.

### DIFF
--- a/ports/esp32/mphalport.c
+++ b/ports/esp32/mphalport.c
@@ -47,6 +47,10 @@
 #include "usb_serial_jtag.h"
 #include "uart.h"
 
+#if MICROPY_PY_STRING_TX_GIL_THRESHOLD < 0
+#error "MICROPY_PY_STRING_TX_GIL_THRESHOLD must be positive"
+#endif
+
 TaskHandle_t mp_main_task_handle;
 
 static uint8_t stdin_ringbuf_array[260];
@@ -129,7 +133,13 @@ mp_uint_t mp_hal_stdout_tx_strn(const char *str, size_t len) {
     // Only release the GIL if many characters are being sent
     mp_uint_t ret = len;
     bool did_write = false;
-    bool release_gil = len > 20;
+    bool release_gil = len > MICROPY_PY_STRING_TX_GIL_THRESHOLD;
+    #if MICROPY_DEBUG_PRINTERS && MICROPY_DEBUG_VERBOSE && MICROPY_PY_THREAD_GIL
+    // If verbose debug output is enabled some strings are printed before the
+    // GIL mutex is set up.  When that happens, no Python code is running and
+    // therefore the interpreter doesn't care about the GIL not being ready.
+    release_gil = release_gil && (MP_STATE_VM(gil_mutex).handle != NULL);
+    #endif
     if (release_gil) {
         MP_THREAD_GIL_EXIT();
     }
--- a/ports/esp32/mpconfigport.h
+++ b/ports/esp32/mpconfigport.h
@@ -314,3 +314,8 @@ void boardctrl_startup(void);
 #endif
 #endif
 #endif
+
+// The minimum string length threshold for string printing to stdout operations to be GIL-aware.
+#ifndef MICROPY_PY_STRING_TX_GIL_THRESHOLD
+#define MICROPY_PY_STRING_TX_GIL_THRESHOLD  (20)
+#endif


dpgeorge:
What exactly is being printed before `mp_init()` runs?  From my understanding, `printf()` on the esp32 port does not go via `mp_hal_stdout_tx_strn()`, so the printing before `mp_init()` must come from a call to `mp_printf()`, but where from exactly?

ports/esp32/mphalport.c
```diff
@@ -129,7 +133,13 @@ mp_uint_t mp_hal_stdout_tx_strn(const char *str, size_t len) {
     // Only release the GIL if many characters are being sent
     mp_uint_t ret = len;
     bool did_write = false;
-    bool release_gil = len > 20;
+    bool release_gil = len > MICROPY_PY_STRING_TX_GIL_THRESHOLD;
+    #if MICROPY_DEBUG_PRINTERS && MICROPY_DEBUG_VERBOSE && MICROPY_PY_THREAD_GIL
+    // If verbose debug output is enabled some strings are printed before the
+    // GIL mutex is set up.  When that happens, no Python code is running and
+    // therefore the interpreter doesn't care about the GIL not being ready.
+    release_gil = release_gil && (MP_STATE_VM(gil_mutex.handle) != NULL);
```
dpgeorge, line 141: it should be `MP_STATE_VM(gil_mutex).handle`

agatti: Sorry, fixed.

agatti:
> What exactly is being printed before `mp_init()` runs? From my understanding, `printf()` on the esp32 port does not go via `mp_hal_stdout_tx_strn()`, so the printing before `mp_init()` must come from a call to `mp_printf()`, but where from exactly?

I don't have a module ready at the moment, but for example (taken from the original bug report gdb session):

```
#0  0x40380080 in _panic_handler ()
#1  0x40390988 in xQueueGenericSend (xQueue=0x0, 
    pvItemToQueue=pvItemToQueue@entry=0x0, xTicksToWait=<optimized out>, 
    xTicksToWait@entry=0, xCopyPosition=xCopyPosition@entry=0)
    at /home/agatti/.local/esp-idf/components/freertos/FreeRTOS-Kernel/queue.c:832
#2  0x42008c2e in mp_thread_mutex_unlock (mutex=<optimized out>)
    at /home/agatti/src/micropython/ports/esp32/mpthreadport.c:223
#3  0x42006924 in mp_hal_stdout_tx_strn (
    str=0x3c1425f8 "Initializing GC heap: %p..%p = %u bytes\n", len=22)
    at /home/agatti/src/micropython/ports/esp32/mphalport.c:134
#4  0x4200ba38 in mp_vprintf (print=0x3c142908 <mp_plat_print>, 
    fmt=<optimized out>, 
    fmt@entry=0x3c1425f8 "Initializing GC heap: %p..%p = %u bytes\n", 
    args=args@entry=0x3fca8f74) at /home/agatti/src/micropython/py/mpprint.c:396
#5  0x4201b864 in DEBUG_printf (
    fmt=fmt@entry=0x3c1425f8 "Initializing GC heap: %p..%p = %u bytes\n")
    at /home/agatti/src/micropython/shared/libc/printf.c:44
#6  0x4200a1b8 in gc_init (start=start@entry=0x3fca9018, end=0x3fcb9010, 
    end@entry=0x3fcb9018) at /home/agatti/src/micropython/py/gc.c:192
#7  0x42006268 in mp_task (pvParameter=<optimized out>)
    at /home/agatti/src/micropython/ports/esp32/main.c:128
```

I'll get a full list of messages printed before `mp_init` when I get back home.

dpgeorge:
> Edit: here's what the default build prints before `mp_init`:

Ah, I see, it's the initialisation of the GC heap.

So then it does make sense to check if the GIL is ready when debug printing is enabled.

PR #12937, opened 2023-11-09
mimxrt: Provide WiFi and BLE support for MIMXRT boards.

robert-hh:
This support uses ESP32 modules with either NINAW10 firmware or esp-hosted firmware. For MIXRT101x devices, the NINAW10 firmware is expected. For MIMXRT102x, MIMXRT105x and MIMXRT106x the esp_hosted firmware has to be used. The esp_hosted firmware is faster and has a more robust BLE support, but the MIMXRT to support  it gets too larger for MIMXRT101x devices. This PR includes Pin definitions and enables WiFI and BLE for:

- ADAFRUIT_METRO_M7_AIRLIFT
- MIMXRT1010_EVK
- MIMXRT1015_EVK
- OLIMEX RT1010py
- MIMXRT1020_EVK
- MIMXRT1050_EVK
- Seeed Arch Mix
- Teensy 4.0
- Teensy 4.1

As ESP32 device, e.g. Adafruit Airlift modules, Nina W10 modules or any generic ESP32 breakout can be used. The NINAW10 or esp_hosted firmware has to be installed to that device. Therefore there are dependencies of this device to a few other PR's to micropython-lib, especially https://github.com/micropython/micropython-lib/pull/642 and https://github.com/micropython/micropython-lib/pull/654. The PR https://github.com/micropython/mynewt-nimble/pull/2 prevents a rare lock-up situation in the ble-nimble stack, which occurred during testing with the BLE test suite.

PR #15639, opened 2024-08-11
Remove old board variants

mattytrentini:
I noticed a few unusual board variants and think they should be removed.

IDF3 builds are _very_ old now (it seems like the last successful builds are from 2021?). Can we remove them? If we want to retain the old builds for [ESP32_GENERIC](https://micropython.org/download/ESP32_GENERIC/) then we'll need to leave `ESP32_GENERIC/board.json` intact but we should at least remove it from `UM_TINYPICO/board.json` since there are [no IDF3 builds for that board](https://micropython.org/download/UM_TINYPICO/). 

If we remove IDF3 entirely we could also slightly simplify [build-boards.sh:61](https://github.com/micropython/micropython/blob/master/tools/autobuild/build-boards.sh#L61). I haven't made that change but can if we decide to remove IDF3.

Finally, I noticed a `wlan` variant for `ADAFRUIT_METRO_M4_EXPRESS`. I believe the samd port doesn't support variants so I think this is a mistake - and appears to be causing an unnecessary build.

### DIFF
--- a/tools/autobuild/build-boards.sh
+++ b/tools/autobuild/build-boards.sh
@@ -55,10 +55,8 @@ function build_board {
     $MICROPY_AUTOBUILD_MAKE BOARD=$board BUILD=$build_dir && copy_artefacts $dest_dir $descr $fw_tag $build_dir $@
     rm -rf $build_dir
 
-    # Query variants from board.json and build them. Ignore the special "IDF3"
-    # variant for ESP32 boards (this allows the downloads page to still have
-    # the idf3 files for older releases that used to be explicitly built).
-    for variant in `cat $board_json | python3 -c "import json,sys; print(' '.join(v for v in json.load(sys.stdin).get('variants', {}).keys() if v != 'IDF3'))"`; do
+    # Query variants from board.json and build them.
+    for variant in `cat $board_json | python3 -c "import json,sys; print(' '.join(json.load(sys.stdin).get('variants', {}).keys()))"`; do
         local variant_build_dir=$build_dir-$variant
         echo "building variant $descr $board $variant"
         $MICROPY_AUTOBUILD_MAKE BOARD=$board BOARD_VARIANT=$variant BUILD=$variant_build_dir && copy_artefacts $dest_dir $descr-$variant $fw_tag $variant_build_dir $@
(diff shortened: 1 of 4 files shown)


dpgeorge:
Thanks, this looks good.

> IDF3 builds are _very_ old now (it seems like the last successful builds are from 2021?). Can we remove them?

Yes I think we should remove them.  If needed, one can build it themselves using IDF v3 (and this repo checked out at the appropriate commit in the far past).

> If we remove IDF3 entirely we could also slightly simplify [build-boards.sh:61](https://github.com/micropython/micropython/blob/master/tools/autobuild/build-boards.sh#L61). I haven't made that change but can if we decide to remove IDF3.

Yes, please do this as well.

Please can you split this PR into two commits:
- remove IDF v3 variants, and the bit from `build-boards.sh`
- remove the variant from `ADAFRUIT_METRO_M4_EXPRESS`

robert-hh:
There is a PR #12937 to add WLAN and BLE support for the SAMD51 family. The WLAN variant was created for that PR, and it seems to have sneaked into the master branch. But it harndly can be built, since the remainder of the PR is missing.

mattytrentini:
This should be ready now @dpgeorge 👍🏻

tools/autobuild/build-boards.sh
```diff
@@ -55,10 +55,8 @@ function build_board {
     $MICROPY_AUTOBUILD_MAKE BOARD=$board BUILD=$build_dir && copy_artefacts $dest_dir $descr $fw_tag $build_dir $@
     rm -rf $build_dir
 
-    # Query variants from board.json and build them. Ignore the special "IDF3"
-    # variant for ESP32 boards (this allows the downloads page to still have
-    # the idf3 files for older releases that used to be explicitly built).
-    for variant in `cat $board_json | python3 -c "import json,sys; print(' '.join(v for v in json.load(sys.stdin).get('variants', {}).keys() if v != 'IDF3'))"`; do
+    # Query variants from board.json and build them.
+    for variant in `cat $board_json | python3 -c "import json,sys; print(' '.join(v for v in json.load(sys.stdin).get('variants', {}).keys()))"`; do
```
dpgeorge, line 59: This list comprehension is simple enough that the `v for v in` bit can now be removed.

mattytrentini: Of course; fixed. 🤦🏻‍♂️

dpgeorge (approved):

PR #5642, opened 2020-02-14
stm32/boards/WEACTV20_F411: Add support for WeAct v2.0 'blackpill' board

jeremyherbert:
The WeAct v2.0 board (also sometimes known as the "blackpill") is a micropython-capable board based on the STM32F411CE (512kB flash, 128kB SRAM) that costs only a few US dollars and is widely available from various Chinese suppliers. It is a breadboard compatible breakout-style board that has reset/BOOT0/user switches, a user LED and a USB C connector, along with 25MHz and 32.768kHz crystals.

Although the board has an unpopulated flash footprint, I don't have any SPI flash ICs around to test it and add support, so this only supports internal flash for the time being. I have ordered some and will try to test the external flash when I get a chance.

moved from https://github.com/micropython/micropython/pull/5582 to a branch

PR #12676, opened 2023-10-12
stm32/boards/ARDUINO_PORTENTA_H7: Update pin definitions.

loopback:
Changes made to pins.csv:
- Removed processor pin designations from board pin names.
- Added board designations for the high-density connectors.

Refer to Discord discussion [here](https://discord.com/channels/574275045187125269/1012726390501671002/1161099839858282576).

PR #15646, opened 2024-08-13
stm32/WEACT_F411_BLACKPILL: Add WeAct F411 'blackpill' board.

andrewleech:
### Summary

Adds board profile for the WeAct F411 v2.0 'blackpill' which is a quite popular low cost stm dev board.

This board also has optional spiflash so can be purchased in a few different configurations. I've added a variant for 4MB flash.

Existing board profiles for F411 don't work on this board as it's built with a 25Mhz crystal.

This was built on top of https://github.com/micropython/micropython/pull/5642 refactored and updated for current compatibility.

### Testing

I've used builds from this on two BlackPill boards I own, one with loaded flash, one without


mattytrentini:
Looks good @andrewleech ! 

Hey @mcauser, you've worked with these boards a bit haven't you? Would you be able to review?

robert-hh:
I have a F411 board here, sold as Weact Studio F411 Blackpill V3.1. There is one difference in the wiring: for my V3.1 board SPI_MOSI is at PA6, not PB4 (and the flash size is 8MB). Otherwise the PR works.

andrewleech:
Thanks @robert-hh that sounds like something that would work as another variant!

robert-hh:
That requires a little bit more awkward approach, since the changes have start in pins.csv and reflected then in a variant-specific mpconfigboard-VARIANT.mk.  Mayb by defining two variants of the MISO pin in pins,csv and selecting the right one later.

robert-hh:
Thanks. These new definitions seem to work, with two minor hiccups.

- In line 2 of `mpconfigvariant_V3_FLASH_8MB.mk` the trailing \\ is missing
- the board name `WEACTV20_F411` indicates, it's for V2 only. Maybe better just use `WEACT_F411` or` WEACT_F411_BLACKPILL`.

andrewleech:
Thanks @robert-hh good catch and suggestions, I've updated as such.

ports/stm32/boards/WEACT_F411_BLACKPILL/pins.csv
```diff
@@ -0,0 +1,44 @@
...
+SWCLK,PA14
+OSC32_IN,PH0
+OSC32_OUT,PH1
+USB_DM,PA11
+USB_DP,PA12
+FLASH_CS,PA4
+FLASH_SCK,PA5
+FLASH_MOSI,PA7
+FLASH_MISO_V2,PB4
+FLASH_MISO_V3,PA6
```
dpgeorge, line 44: Are these flash pins intended to be exposed via board name to the user?  If not then you can do `-FLASH_CS,PA4` (for example); it's still exposed as `pyb_pin_FLASH_CS` but the user won't see `Pin.board.FLASH_CS`.

I don't mind either way (exposing to user or not).

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigboard.h
```diff
@@ -0,0 +1,122 @@
...
+
+#define MICROPY_HW_RTC_USE_LSE      (1)
+#define MICROPY_HW_RTC_USE_US       (0)
+#define MICROPY_HW_RTC_USE_CALOUT   (1)
+
+// SPI flash pins
+#define MICROPY_HW_SPIFLASH_CS          (pyb_pin_FLASH_CS)
+#define MICROPY_HW_SPIFLASH_SCK         (pyb_pin_FLASH_SCK)
+#define MICROPY_HW_SPIFLASH_MOSI        (pyb_pin_FLASH_MOSI)
+#ifdef WEACT_F411_V3
```
dpgeorge, line 90: Maybe `#if WEACT_F411_V3`?

davefes:
@andrewleech I built the firmware using your mpconfigboard.h from:
https://github.com/micropython/micropython/commit/686a415fefd2832595526e5ba1ffe73ed8622250
and it seems to work. I didn't change any of the other files in the WEACTF411CE folder that were supplied by:
https://github.com/nspsck/STM32F411CEU6_BlackPill_Micropython.

>>> micropython.mem_info()
stack: 476 out of 15360
GC: total: 86976, used: 18304, free: 68672
 No. of 1-blocks: 214, 2-blocks: 66, max blk sz: 72, max free sz: 3377

Is there something else I need to do to get the 8MB? I'd be keen to build for the 8MB flash to see if it solves a comms issue for me.

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigboard.h
```diff
@@ -0,0 +1,122 @@
...
+#ifndef MICROPY_HW_SPIFLASH_SIZE_BITS
+#define MICROPY_HW_SPIFLASH_SIZE_BITS (128 * 1024 * 1024)
+#endif
+
+#if !MICROPY_HW_ENABLE_INTERNAL_FLASH_STORAGE
+
+extern const struct _mp_spiflash_config_t spiflash_config;
+extern struct _spi_bdev_t spi_bdev;
+#define MICROPY_HW_SPIFLASH_ENABLE_CACHE (1)
+#define MICROPY_HW_BDEV_IOCTL(op, arg) ( \
```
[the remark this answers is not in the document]
andrewleech: Oh I missed that consolidating, nice one thanks!

andrewleech:
I've restructured the board and variant configuration a bit in the latest push.

v3.1 is now the default as it's what people will get if they buy them new currently, no flash configured.
There are then two extra v3.1 variants for 8MB flash and 8 MHz xtal (no flash) matching the configurations available for sale. 

After that there's a variant for v2.0 with 4MB flash - the only difference between v2.0 and v3.1 boards is the pinout of the spiflash I believe so users with v2.0 (no flash) will also be able to use the default build. 

I then found the boards I own are actually v1.3 which were the original version released a few years ago. On these the spiflash pinout matches v3.1 but does not have a user switch 

I noticed that on v1.3 and v3.1 boards the spiflash pinout aligns with the spi hardware peripheral so that's now being used. I don't know why they changed it in v2.0, regardless it's cinched with softspi.

I still need to do a final sanity check on the boards I have, I'd also appreciate if anyone can test on v2.0 or v3.1 boards.

@davefes  I'm not quite following if you're trying to copy some of this PR into a different micropython source?  This whole PR is the board folder which uses the variant build option to select the alternate configurations. The readme in the stm port folder should give guidance on how this works iirc.

ports/stm32/boards/WEACT_F411_BLACKPILL/pins.csv
andrewleech: There's a few pins here that probably don't really add value for a user, I'd forgotten about the - in this file thanks for the reminder.

robert-hh:
> If anyone can I'd appreciate a test on v2.0 or v3.1 boards!

I tested the V3.1, 25MHz variant with and without flash. It builds and runs. Next week I will get a 8MHz V3.1 board which I then will test. Looking at the config files I would have expected a value of 8 for the 8MHz version at 
`#define MICROPY_HW_CLK_PLLM (8)`. I will see.

davefes:
After the following change:
```
In file included from ./mpconfigport.h:31,
                 from ../../py/mpconfig.h:91,
                 from ../../py/mpstate.h:31,
                 from ../../py/runtime.h:29,
                 from usrsw.c:29:
usrsw.c: In function 'switch_init0':
boards/WEACTF411CE/mpconfigboard.h:66:38: error: 'pyb_pin_SW' undeclared (first use in this function); did you mean 'pyb_pin_B1'?
   66 | #define MICROPY_HW_USRSW_PIN        (pyb_pin_SW)
```
it built and runs the base-version.

I don't see a readme in the /stm32 folder.  How do I change:
`make -j BOARD=WEACT411CE` 
to build for the 8MB version?

ports/stm32/boards/WEACT_F411_BLACKPILL/pins.csv
```diff
@@ -0,0 +1,45 @@
...
+PB7,PB7
+PB8,PB8
+PB9,PB9
+PB10,PB10
+PB12,PB12
+PB13,PB13
+PB14,PB14
+PB15,PB15
+PC14,PC14
+PC15,PC15
```
dpgeorge, line 31: I suggest to not expose these pins with board names, ie just do `,PC15` etc.

We can always add them in the future if they are needed, but we can't take them away.

See related #12676 and #13664.

ports/stm32/boards/WEACT_F411_BLACKPILL/stm32f4xx_hal_conf.h
```diff
@@ -0,0 +1,23 @@
...
+ * The MIT License (MIT)
+ * Copyright (c) 2024 Andrew Leech
+ */
+#ifndef MICROPY_INCLUDED_STM32F4XX_HAL_CONF_H
+#define MICROPY_INCLUDED_STM32F4XX_HAL_CONF_H
+
+#include "boards/stm32f4xx_hal_conf_base.h"
+
+// Oscillator values in Hz
+#if MICROPY_HW_HSE_VALUE
```
dpgeorge, line 11: In this case I think it's better to do `#ifdef MICROPY_HW_HSE_VALUE`, because if that value is defined it should be used (even if it's 0...).

davefes:
Should I be able to make a LFS2 filesystem?
```
import make_LFS2
  File "make_LFS2.py", line 4, in <module>
AttributeError: 'module' object has no attribute 'VfsLfs2'
```
The base version works on `rsusb` but not using `mpremote`.

Still can't work-out how to build for 8MB.

robert-hh:
>  Next week I will get a 8MHz V3.1 board which I then will test.

Looking at the order confirmations, I did not order a 8MHz version of the board, like I thought I did. Shhhh.... But looking into the F411 data sheet, it is obvious that for the 8MHz version the value of `MICROPY_HW_CLK_PLLM` has to be 8 to get 1 MHz at the input of the PLL.

ports/stm32/boards/WEACT_F411_BLACKPILL/bdev.c
```diff
@@ -0,0 +1,45 @@
...
+
+static const mp_soft_spi_obj_t soft_spi_bus = {
+    .delay_half = MICROPY_HW_SOFTSPI_MIN_DELAY,
+    .polarity = 0,
+    .phase = 0,
+    .sck = MICROPY_HW_SPIFLASH_SCK,
+    .mosi = MICROPY_HW_SPIFLASH_MOSI,
+    .miso = MICROPY_HW_SPIFLASH_MISO,
+};
+#define
```
dpgeorge, line 17: stray line

ports/stm32/boards/WEACT_F411_BLACKPILL/board.json
```diff
@@ -0,0 +1,22 @@
...
+    "mcu": "stm32f411",
+    "product": "WeAct F411 'blackpill'. Default variant is v3.1 with no SPI Flash.",
+    "thumbnail": "",
+    "url": "https://github.com/WeActStudio/WeActStudio.MiniSTM32F4x1",
+    "variants": {
+        "V13": "v1.3 board with no SPI Flash",
+        "V13_FLASH_4M": "v1.3 board with 4MB SPI Flash",
+        "V20_FLASH_4M": "v2.0 board with 4MB SPI Flash",
+        "V31_FLASH_8M": "v3.1 board with 8MB SPI Flash",
+        "V31_XTAL_8M": "v3.1 board with 8MHz crystal"
```
dpgeorge, line 19: Just to be clear, this 8MHz crystal version doesn't have SPI flash?

How do people distinguish v3.1 with the different crystals?

ports/stm32/boards/WEACT_F411_BLACKPILL/bdev.c
```diff
@@ -0,0 +1,45 @@
+#include "storage.h"
+#include "spi.h"
+
+#if !MICROPY_HW_ENABLE_INTERNAL_FLASH_STORAGE
+
+#if WEACT_F411_V2
```
dpgeorge, line 6: I think this should be `WEACT_F411_V20`.

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigboard.h
```diff
@@ -0,0 +1,112 @@
...
+// Disable internal filesystem to use spiflash.
+#define MICROPY_HW_ENABLE_INTERNAL_FLASH_STORAGE (0)
+
+// SPI flash pins
+#define MICROPY_HW_SPIFLASH_CS          (pyb_pin_FLASH_CS)
+#define MICROPY_HW_SPIFLASH_SCK         (pyb_pin_FLASH_SCK)
+#define MICROPY_HW_SPIFLASH_MOSI        (pyb_pin_FLASH_MOSI)
+#if WEACT_F411_V13
+#define MICROPY_HW_SPIFLASH_MISO        (pyb_pin_FLASH_MISO_V13)
+#elif WEACT_F411_V2
```
dpgeorge, line 95: Should be `WEACT_F411_V20`.

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigvariant_V13_FLASH_4M.mk
```diff
@@ -0,0 +1,3 @@
+CFLAGS += \
+  -DWEACT_F411_V13=1 \
```
dpgeorge, line 2: This is a makefile, so shouldn't there technically be tab indenting here?

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigvariant_V31_FLASH_8M.mk
```diff
@@ -0,0 +1 @@
+CFLAGS += -DMICROPY_HW_SPIFLASH_SIZE_BYTES="(8 * 1024 * 1024)"
```
dpgeorge, line 1: Probably worth adding `-DWEACT_F411_V31=1` for consistency.

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigvariant_V31_XTAL_8M.mk
```diff
@@ -0,0 +1 @@
+CFLAGS += -DMICROPY_HW_HSE_VALUE="(8000000)"
```
dpgeorge, line 1: Add `-DWEACT_F411_V31=1`?

ports/stm32/boards/WEACT_F411_BLACKPILL/board.json
davefes: I can only suggest the way nspsck does here:
https://github.com/nspsck/STM32F411CEU6_BlackPill_Micropython/blob/main/README.md

> For 8MHz crystal and 16MB spi flash:

make -j BOARD=WEACTF411CE BOARD_VARIANT=VARIANTS CRYSTAL_FREQ=8 SPI_FLASH_SIZE=16

For 25MHz crystal and 16MB spi flash, you have to leave CRYSTAL_FREQ out:

make -j BOARD=WEACTF411CE BOARD_VARIANT=VARIANTS SPI_FLASH_SIZE=16

maybe ignore BOARD_VARIANT depending on how Andrew builds these variants.

davefes: So far there is only one variant with 8MHz xtal ... no external flash.

dpgeorge: To build a variant with this PR use:
```
make BOARD=WEACT_F411_BLACKPILL BOARD_VARIANT=V31_XTAL_8M
```

davefes: Do I need to re-load any of the build files since `203a2a9`?

dpgeorge: You probably need to re-pull this PR.  But some of the comments here need addressing before it will build correctly for all variants.

davefes: Thank you, I'll wait.

robert-hh: > Just to be clear, this 8MHz crystal version doesn't have SPI flash?

According to the store pages: No. But the footprint for the flash chip is present, so people could add a flash chip. But then they have to adapt the build files and build the firmware themselves.

> How do people distinguish v3.1 with the different crystals?

They have to look at the crystal. 

I re-ordered a 8MHz crystal board and will get it next week. Then I can check whether the configuration works. ATM it does not look right, since `MICROPY_HW_CLK_PLLM` is defined unconditionally as 25 for all boards. But maybe that definition is not effective.

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigvariant_V13_FLASH_4M.mk
andrewleech: Oh, yeah technically it doesn't matter after a line-continuation `\` however by convention it is better to use a tab for consistency.

ports/stm32/boards/WEACT_F411_BLACKPILL/bdev.c
andrewleech: Oof, that's what I get for changing my mind on the formatting of these version defines, thanks for catching it.

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigvariant_V13_FLASH_4M.mk
dpgeorge: Oh, I did not realise that when lines are continued they don't need tabs.

But still good to use tabs for consistency.

andrewleech:
Note in the latest push I've changed the way the make args are handled to simplify end user building for customised boards, but it's not quite right yet. Don't bother reviewing too much, I'll let everyone here know when it's working.

robert-hh:
You could set the default value of MICROPY_HW_HSE_SPEED_MHZ together with the HSE_VALUE. The the change between 8MHz and 25Mhz is only in a single file. Like in stm32f4xx_hal_conf.h:

```
// Oscillator values in Hz
#ifndef MICROPY_HW_HSE_SPEED_MHZ
#define MICROPY_HW_HSE_SPEED_MHZ (25)
#endif
#define HSE_VALUE (MICROPY_HW_HSE_SPEED_MHZ * 1000000)
```

And in mpconfigboard.h there is only:
`#define MICROPY_HW_CLK_PLLM (MICROPY_HW_HSE_SPEED_MHZ)`

Besides that and even without the above change, Makefile has to be changed for include mpconfigboard.mk after the respective mpconfigvariant_xxx.mk. That's because `XTAL_FREQ_MHZ` is set in the mpconfigvariant_xxx.mk and used in mpconfigboard.mk. But I do not know if that has implications in other ports.

Note: Funny that such a seemingly simple board addition needs so much attention.

robert-hh:
It seems that the definition of `#define MICROPY_HW_CLK_PLLM (MICROPY_HW_HSE_SPEED_MHZ)` fails. Since in `plli2svalues.py` the search expression is `#define +(HSE_VALUE) +\((\(uint32_t\))?([0-9]+)\)"`. So the above #define is ignored and always `#define MICROPY_HW_CLK_PLLM (25)` is used.

davefes:
Well, that is the first time I have software built to use 8MB flash that actually works. I am use to changing to a LFS2 filesystem on a previous distro, however:

import make_LFS2
  File "make_LFS2.py", line 4, in <module>
AttributeError: 'module' object has no attribute 'VfsLfs2'

VfsLFs2 appears to no longer be in os?

robert-hh:
Getting 'os.VfsLfs2' requires at least to define
`MICROPY_VFS_LFS2 = 1`
in mpconfigboard.h

andrewleech:
> VfsLFs2 appears to no longer be in os?

Timely test, thanks for letting me know. Yes the entry to enable that was missing so I've added that now in the latest push.

robert-hh:
Now, with the 8M Flash version I have the same problem with the sequence of including mpconfigboard.mk and mpconfigvariant_xxx.mk. mpconfigvariant_xxx.mk defines symbols, which are needed in mpconfigboard.mk for proper configuration, like the flash and clock speed. But if mpconfigboard.mk is included first in Makefile, these symbols never get active,

andrewleech:
> It seems that the definition of `#define MICROPY_HW_CLK_PLLM (MICROPY_HW_HSE_SPEED_MHZ)` fails. Since in `plli2svalues.py` the search expression is `#define +(MICROPY_HW_CLK_PLLM) +\((\(uint32_t\))?([0-9]+)\)",`. So the above #define is ignored and always `#define MICROPY_HW_CLK_PLLM (25)` is used.

Oh that is a complication, I wasn't aware of the pll code generation scripts.
As it stands it looks like the i2s one takes `mpconfigboard.h` as the input and just finds the last define like that, suggesting there's no way to make it configurable at all. 
As such even changing it to:
```
#if MICROPY_HW_HSE_XTAL_8M
#define MICROPY_HW_CLK_PLLM (8)
#else
#define MICROPY_HW_CLK_PLLM (25) // Default HSE is 25MHz
#endif
```
doesn't work, the script looks over every line,  keeping the data from the last line that matches the regex. 
This will need some more thought...

andrewleech:
> I have the same problem with the sequence of including ...

Yes I found that issue too, I think I have it all resolved in the last push though (except the pll script issue)

robert-hh:
The last push still includes mpconfigboard.mk before e.g. mpconfigvariant_V31_FLASH_8M.mk, and so the 8M flash setting is never propagated.

davefes:
I think I spoke too soon:
```
MicroPython v1.24.0-preview.220.g185116ea4 on 2024-09-02; WEACT_F411_BLACKPILL with STM32F411CE
import micropython
micropython.mem_info()
stack: 468 out of 15360
GC: total: 83072, used: 1984, free: 81088
 No. of 1-blocks: 29, 2-blocks: 8, max blk sz: 40, max free sz: 5054
```
... only 80K on an 8MB flash or I am not doing this correctly.

robert-hh:
Maybe you have to recreate the file system manually. If that fails, try to reload the PR and build again. There was an interim state where the external flash was not used.

davefes:
Looking through `66579b1` I don't see `MICROPY_VFS_LFS2 = 1`  Should I be looking somewhere else?

andrewleech:
@dpgeorge as far as I can see there's currently no way to support if/else on HSE, HSI or PLL variables due to the way they're parsed in python scripts to generate pll tables. 

I've addressed this in https://github.com/micropython/micropython/commit/f62389cc9509a14e62d9e4abfb6f867cc548897d which adjusts things to run the python scripts on pre-processed outputs, with some extra variables defined to ensure the configured values are calculated and populated in the `qstr.i.last` outputs file.

I did trial changing `CPP` to include headers in the output `-E -dD` however this blew `qstr.i.last` from ~90MB up to ~350MB and didn't actually expand "nested" defines so still only worked in certain situations, eg. `#define MICROPY_HW_CLK_PLLM (MICROPY_HW_HSE_SPEED_MHZ)` wasn't expanded so still wasn't captured by the regex in the python scripts.

andrewleech:
> I don't see `MICROPY_VFS_LFS2 = 1`

Ah, yeah, I added it and somehow lost it again. Too much juggling multiple changes on my end I suspect. It's in the latest push now. 

I do think I might have all the issues resolved now...
`build-WEACT_F411_BLACKPILL-V31_XTAL_8M/genhdr/plli2stable.h`
```
// MAKE generated file, created by plli2svalues.py: DO NOT EDIT
// This table is used in machine_i2s.c
// HSE_VALUE = 8000000
// MICROPY_HW_CLK_PLLM = 8 
...
```
`build-WEACT_F411_BLACKPILL-V13_FLASH_4M/genhdr/plli2stable.h`
```
// MAKE generated file, created by plli2svalues.py: DO NOT EDIT
// This table is used in machine_i2s.c
// HSE_VALUE = 25000000
// MICROPY_HW_CLK_PLLM = 25 
...
```
``` bash
anl@STEP: ~/micropython $ mpremote bootloader
anl@STEP: ~/micropython $ make -C ports/stm32 BOARD=WEACT_F411_BLACKPILL BOARD_VARIANT=V13_FLASH_4M deploy
...
anl@STEP: ~/micropython $ mpremote
MicroPython v1.24.0-preview.272.gf62389cc95 on 2024-09-03; WEACT_F411_BLACKPILL with STM32F411CE
>>> import os
>>> os.statvfs(".")
(2048, 2048, 2038, 2033, 2033, 0, 0, 0, 0, 255)
>>>

**Comment by davefes** [comment]
As I would have quite a bit to learn to continue ... is the intention that pre-built firmware will be available on the Downloads page?

**Comment by robert-hh** [comment]
@andrewleech The Flash 8M build works, and os.VfsLfs2 is present as well. Only I did not succeed formatting the external Flash with a LFS file system. But that may be another story nor related to the board. 

A minor comment to the board files. Since there is the line:

`CFLAGS += -DMICROPY_HW_HSE_SPEED_MHZ=$(XTAL_FREQ_MHZ)`

in `mpconfigboard.h`, the symbol `MICROPY_HW_HSE_SPEED_MHZ` is always defined, and the respective `#ifndef MICROPY_HW_HSE_SPEED_MHZ` in `mpconfigboard.h` is always False. But that does not hurt.

**Comment by andrewleech** [comment]
Thanks @robert-hh I actually hadn't tested formatting lfs myself, but in my case it seemed to work as per: https://docs.micropython.org/en/latest/reference/filesystem.html#littlefs

``` python
anl@STEP: ~/micropython $ mpremote
>>>
MPY: sync filesystems
MPY: soft reboot
MicroPython v1.24.0-preview.272.gf62389cc95 on 2024-09-03; WEACT_F411_BLACKPILL
with STM32F411CE
>>> import os, vfs, pyb
>>> vfs.umount('/flash')
>>> vfs.VfsLfs2.mkfs(pyb.Flash(start=0))
>>> vfs.mount(pyb.Flash(start=0), '/flash')
>>> os.chdir('/flash')
>>> os.statvfs('/flash')
(4096, 4096, 1024, 1022, 1022, 0, 0, 0, 0, 255)
>>>
```

andrewleech:
@davefes sorry I haven't provided much guidance at all during this PR but your efforts have certainly been inspiring me throughout to make this easier! 
In a recent update I added a readme with some example build commands. It was based on your suggestion I added similar arguments that can be used in that make command to set custom flash sizes.

Maybe with a fresh pull of the latest commit along with the examples you might get a better working build? 

But regardless, yes if this PR is accepted then builds of all the defined variants will end up being on the downloads page.

davefes:
No problem.  I never did a pull of your commits, but just copy/pasted the board definition ones.  Getting the changes you have made to the files on `micropython master` are beyond me.  Am I correct in thinking that those changes will work their way into `master`?

andrewleech:
My goal is to get all these changes to a high enough standard to be merged into master yes, but these things take time because there are a lot of other active pr's so we have to be patient!

davefes:
Just looking through the files and the changes made it did seem a BIG job.  Would it be possible for you to just zip up your latest changes to `Makefile`, `plli2svalues.py`, `pllvalues.py', `machine_i2s.c` and 'powerctrl.c` and I could drop them into my `micropython` and do a build? If not, there is no problem waiting until the changes make it into prime-time.

robert-hh:
@davefes What stops you from pulling this PR into your clone of MicroPython?

Assuming the remote's name is `origin`:
```
git fetch origin pull/15646/head:weact_blackpill
git checkout weact_blackpill
```

davefes:
Not having done it before!  That part worked, thanks.  How do I get the updated files I mentioned?

davefes:
Built for 8MB flash, made a LFS2 filesystem and the application runs fine.  Still, when running:
```
import micropython
micropython.mem_info()
stack: 468 out of 15360
GC: total: 83072, used: 17120, free: 65952
 No. of 1-blocks: 173, 2-blocks: 60, max blk sz: 72, max free sz: 3183
```

andrewleech:
@davefes `micropython.mem_info()` shows free ram, not flash.  The way to check flash is 
```
>>> os.statvfs('/flash')
(4096, 4096, 1024, 1022, 1022, 0, 0, 0, 0, 255)
```
The output is somewhat confusing https://docs.python.org/3/library/os.html#os.statvfs
basically the first element is block size (in bytes), the third element is total number of blocks.  So for me it's 4096 * 1024 bytes = 4MB

davefes:
O dear, thanks for the correction.

davefes:
I only have 2 boards, at the moment.  One board without flash and using the base-version and VFAT works, when I change to LFS2 I get the dreaded timeout error on file transfer.  The other board (8MB flash) works properly. Suspect a faulty board.

robert-hh:
@andrewleech The 8MHz V3.1 board arrived. The firmware builds and works. USB registers, sleep functions show a reasonable behavior. So I conclude that the clocks are set up properly.

davefes:
5 out of 5 new 8MB 25MHz boards work.  LFS2 as well.  Great work.

Thanks for all the effort you put into this.

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigboard.mk
```diff
@@ -0,0 +1,26 @@
...
+
+ifdef SPI_FLASH_SIZE_MB
+# Add flash size to build folder name if set from command line
+BUILD := $(BUILD)_FLASH_$(SPI_FLASH_SIZE_MB)M
+endif
+
+ifdef XTAL_FREQ_MHZ
+# Add xtal speed to build folder name if set from command line
+BUILD := $(BUILD)_XTAL_$(XTAL_FREQ_MHZ)M
+endif
```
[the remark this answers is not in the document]
andrewleech: Because of the makefile ordering this isn't used / run when a variant is configured - this is just for when one of these settings is passed on the `make` command line. I thought this would be most helpful for people manually configuring the FLASH size from command line, it will put the flash size in the build folder name.

andrewleech:
@dpgeorge I've finished simplifying the pll generation script args and addressed the other comments to date thanks.

davefes:
I have managed to corrupt the 8MB flash on 3 devices running some simple code. https://github.com/orgs/micropython/discussions/15969
Could you suggest how to erase the external flash?  STM32CubeProgrammer or ST-Link
Thanks

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigboard.mk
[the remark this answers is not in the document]
davefes: Could I suggest a warning in the README.md to not use A4, A5, A6 or A7 on the 8MB variant.  I was vaguely aware that those pins were used for the SPI flash, but did not appreciate the consequences of "accidentally" using them.

dpgeorge:
I have verified that with this PR:
- PYBV10 firmware is unchanged
- PYBD_SF6 firmware is mostly unchanged (it's the same size, and only around 20 lines of an xxd dump differ), and at least the generated pllfreqtable.h and plli2stable.h files are unchanged

ports/stm32/boards/plli2svalues.py
```diff
@@ -137,8 +137,11 @@ def search_header(filename, re_include, re_define, lookup, val):
             m = regex_define.match(line)
             if m:
                 # found lookup value
+                found = m.group(3)
+                if "*" in found or "/" in found:
+                    found = eval(found)
```
dpgeorge, line 142: Did you find cases where this arithmetic was needed?  Can you give an example of a board that needs it?

ports/stm32/boards/plli2svalues.py
```diff
@@ -166,35 +169,41 @@ def main():
...
 
         if argv[0].startswith("hse:"):
-            # extract HSE_VALUE from header file
+            hse = int(argv[0][len("hse:") :])
+
+        if argv[0].startswith("pllm:"):
+            pllm = int(argv[0][len("pllm:") :])
+
+        if argv[0].startswith("file:"):
+            # extract HSE_VALUE and from header file
```
dpgeorge, line 183: "extract HSE_VALUE from header file"

ports/stm32/boards/plli2svalues.py
```diff
@@ -166,35 +169,41 @@ def main():
...
             (hse,) = search_header(
-                argv[0][len("hse:") :],
+                argv[0][len("file:") :],
                 r'#include "(boards/[A-Za-z0-9_./]+)"',
-                r"#define +(HSE_VALUE) +\((\(uint32_t\))?([0-9]+)\)",
```
dpgeorge, line 178: I don't think this include regex and its associated logic in `search_header` is needed anymore.  That simplifies things, and the `vals` argument is then also not needed.

ports/stm32/boards/pllvalues.py
```diff
@@ -230,7 +230,9 @@ def print_table(hse, valid_plls):
 
 def search_header_for_hsx_values(filename, vals):
     regex_inc = re.compile(r'#include "(boards/[A-Za-z0-9_./]+)"')
-    regex_def = re.compile(r"#define +(HSE_VALUE|HSI_VALUE) +\((\(uint32_t\))?([0-9]+)\)")
```
dpgeorge, line 233: The regex_inc is no longer needed, neither the `vals` argument here.

davefes:
Will this distro end up on the Micropython Downloads page?

ports/stm32/boards/plli2svalues.py
andrewleech: This logic is to support usage like in this PR for https://github.com/micropython/micropython/blob/aecdfc198723e45cf0a88baf8970f361a39f4014/ports/stm32/boards/WEACT_F411_BLACKPILL/stm32f4xx_hal_conf.h#L11
Where the clock speed define is reused in two locations, one location needing it is Hz the other needing it in MHz
https://github.com/micropython/micropython/blob/aecdfc198723e45cf0a88baf8970f361a39f4014/ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigboard.h#L23

dpgeorge: OK, thanks for that.  We should keep this bit of Python logic then.

andrewleech: Thanks but yes I'll add a comment to the section to clarify its purpose

ports/stm32/boards/plli2svalues.py
andrewleech: Ah, yes I follow what you mean; the recursive search really isn't needed at all now that it's parsing the pre-processed / merged headers. Stripping that out certainly simplifies the logic!

ports/stm32/boards/WEACT_F411_BLACKPILL/mpconfigboard.mk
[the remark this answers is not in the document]
andrewleech: I've added a note to the [readme](https://github.com/micropython/micropython/blob/2b7cd314ad22ba65bb249d82d72deef32b09c679/ports/stm32/boards/WEACT_F411_BLACKPILL/README.md) that's a good suggestion; it's not immediately obvious that the pins needed for "on-board" functions are also exposed to the end user.

andrewleech:
@dpgeorge I've updated the scripts, thanks for the suggestions!

dpgeorge (approved):

dpgeorge:
Thanks for updating.  Looks good now.

dpgeorge:
@andrewleech unfortunately this board does not build:
```py
ValueError: file:build-WEACT_F411_BLACKPILL/genhdr/qstr.i.last does not contain a definition of micropy_hw_hse_value
```
It looks like the regex needs to be updated to allow parenthesis within the expression for the value.  Eg it can look like this:
```c
static uint32_t __attribute__((unused)) micropy_hw_hse_value = ((25) * 1000000);
```
A regex like this is simple and seems to work (the `uint32_t` bit isn't needed):
```py
r"static.* +(micropy_hw_hs[ei]_value) = +([0-9 +-/\*()]+);"
```

dpgeorge:
I posted a fix for the above in #16444.

andrewleech:
Oh sorry and thanks, my final cleanup of the board configs was more hassle than I expected!

PR #15652, opened 2024-08-13
shared/tinyusb: Allow ports to define CDC TX/RX buffer sizes.

iabdalkader:
This change allows ports to define CDC TX and RX buffer sizes. Alternatively we could just set TX/RX buffer sizes to match `CFG_TUD_CDC_EP_BUFSIZE` size:
```C
#ifdef CFG_TUD_CDC_EP_BUFSIZE
#define CFG_TUD_CDC_RX_BUFSIZE  (CFG_TUD_CDC_EP_BUFSIZE)
#define CFG_TUD_CDC_TX_BUFSIZE  (CFG_TUD_CDC_EP_BUFSIZE)
#else
#define CFG_TUD_CDC_RX_BUFSIZE  ((CFG_TUD_MAX_SPEED == OPT_MODE_HIGH_SPEED) ? 512 : 256)
#define CFG_TUD_CDC_TX_BUFSIZE  ((CFG_TUD_MAX_SPEED == OPT_MODE_HIGH_SPEED) ? 512 : 256)
#endif
```
This way you only need to define one variable in ports to override all three buffer sizes. I prefer this option. Note that bigger buffers increase the bandwidth.


dpgeorge:
> Alternatively we could just set TX/RX buffer sizes to match `CFG_TUD_CDC_EP_BUFSIZE` size:

I don't think that would work well because tinyusb defines:
```c
#ifndef CFG_TUD_CDC_EP_BUFSIZE
  #define CFG_TUD_CDC_EP_BUFSIZE    (TUD_OPT_HIGH_SPEED ? 512 : 64)
#endif
```
so in our case where a port/board doesn't define `CFG_TUD_CDC_EP_BUFSIZE` it may pick up the tinyusb default, which  is too small for full speed device (only 64 bytes).

Also, I don't think it makes any difference to increase `CFG_TUD_CDC_EP_BUFSIZE` beyond 64 bytes for full-speed devices, the CDC throughput is already maxed out AFAIK.

(If you're low on RAM the main thing that matters is `CFG_TUD_CDC_RX_BUFSIZE`, it should be at least 256 for things like `pyboard.py` and `mpremote` to work properly.)

iabdalkader:
> in our case where a port/board doesn't define `CFG_TUD_CDC_EP_BUFSIZE` it may pick up the tinyusb default, which is too small for full speed device (only 64 bytes).

Note that all ports/boards are already using TinyUSB's defaults (`CFG_TUD_CDC_EP_BUFSIZE` isn't defined anywhere). The alternate change above sets the CDC TX/RX buffers to match the size of `CFG_TUD_CDC_EP_BUFSIZE` *if* it's defined.

> Also, I don't think it makes any difference to increase CFG_TUD_CDC_EP_BUFSIZE beyond 64 bytes for full-speed devices, the CDC throughput is already maxed out AFAIK.

I didn't test FS, but on one HS board, setting `CFG_TUD_CDC_EP_BUFSIZE` higher (to match TX/RX) increases the throughput by about 15-20%. This might be due to less memcpy between buffers or because the EP will more likely have something to send when polled by the host.

I think the current change (in the PR commit) is more flexible because it allows defining all 3 buffers individually.

dpgeorge:
> I think the current change (in the PR commit) is more flexible because it allows defining all 3 buffers individually.

Yes I agree.

PR #15655, opened 2024-08-13
ports/rp2: Skip core1_entry if thread disabled.

Gadgetoid:
If MICROPY_PY_THREAD is set to 0 (ie: a user C module wishes to use core1 exclusively) then the test of `core1_entry` would fail to compile with an "undeclared identifier" error. Fix it by wrapping in MICROPY_PY_THREAD.

Note: `core1_entry` is defined in `mpthreadport.c` *only* if `MICROPY_PY_THREAD` is enabled.


dpgeorge (approved):

projectgus (approved):

dpgeorge:
Rebased and merged in 26d91b897ea46a28f3d32cbe094b607124c1657d

PR #15657, opened 2024-08-14
extmod/mbedtls: Optimise the DER certificate parsing fix.

projectgus:
### Summary

Small code size and binary size optimisation for the fix merged in #14385.

### Testing

* Re-ran unit tests that cover DER parsing, including on ESP-IDF V5.2.2 (see #15523)
* Re-tested PEM parsing manually on ESP32 (I think only port where PEM is enabled by default):

```
$ openssl req -x509 -newkey rsa:2048 -keyout rsa_key.pem -out rsa_cert.pem -days 3650 -nodes -subj '/CN=micropython.local/O=MicroPython/C=AU'
$ mpremote cp rsa_key.pem rsa_cert.pem
$ mpremote repl
```

```
>>> import io,ssl
>>> ssl.wrap_socket(io.BytesIO(), key="rsa_key.pem", cert="rsa_key.pem", do_handshake=False)
<SSLSocket>
```

### DIFF
--- a/extmod/modtls_mbedtls.c
+++ b/extmod/modtls_mbedtls.c
@@ -100,11 +100,22 @@ static void mbedtls_debug(void *ctx, int level, const char *file, int line, cons
 }
 #endif
 
-#if defined(MBEDTLS_PEM_PARSE_C)
-static int mbedtls_is_pem(const byte *data, size_t len) {
-    return (len >= 10) && (strstr((const char *)data, "-----BEGIN") != NULL);
+// Given a string-like object holding PEM or DER formatted ASN.1 data, return a
+// pointer to its buffer and the correct length for mbedTLS APIs.
+//
+// (mbedTLS >= 3.5 rejects DER formatted data with trailing bytes within keylen,
+// but PEM must include a terminating NUL byte in the keylen...)
+static const unsigned char *asn1_get_data(mp_obj_t obj, size_t *out_len) {
+    size_t len;
+    const char *str = mp_obj_str_get_data(obj, &len);
+    #if defined(MBEDTLS_PEM_PARSE_C)
+    if (strstr(str, "-----BEGIN ") != NULL) {
+        ++len;
+    }
+    #endif
+    *out_len = len;
+    return (const unsigned char *)str;
 }
-#endif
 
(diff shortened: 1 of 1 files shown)


extmod/modtls_mbedtls.c
```diff
@@ -100,11 +100,22 @@ static void mbedtls_debug(void *ctx, int level, const char *file, int line, cons
...
 
-#if defined(MBEDTLS_PEM_PARSE_C)
-static int mbedtls_is_pem(const byte *data, size_t len) {
-    return (len >= 10) && (strstr((const char *)data, "-----BEGIN") != NULL);
+// Given a string-like object holding PEM or DER formatted ASN.1 data, return a
+// pointer to its buffer and the correct length for mbedTLS APIs.
+//
+// (mbedTLS >= 3.5 rejects DER formatted data with trailing bytes within keylen,
+// but PEM must include a terminating NUL byte in the keylen...)
+static const byte *asn1_get_data(mp_obj_t obj, size_t *out_len) {
```
dpgeorge, line 108: Considering the mbedtls API uses `const unsigned char *` as the data type for key/cert, perhaps it's worth using that here as the return type, so it matches exactly?  (I know it doesn't really matter, but it may aid in self-documenting the code?)

extmod/modtls_mbedtls.c
```diff
@@ -100,11 +100,22 @@ static void mbedtls_debug(void *ctx, int level, const char *file, int line, cons
...
+    size_t len;
+    const char *str = mp_obj_str_get_data(obj, &len);
+    #if defined(MBEDTLS_PEM_PARSE_C)
+    if (strstr(str, "-----BEGIN ") != NULL) {
+        len++;
```
dpgeorge, line 113: Very minor, but the (weak) convention in this project is to prefer pre-increment `++len` if it doesn't matter the order (pre/post) (because pre-increment is a theoretically simpler operation than post-increment).

projectgus: Sure, have changed.

dpgeorge:
@peterzuger FYI, this should be just an optimisation of your commit.

dpgeorge (approved):

PR #15659, opened 2024-08-14
test,esp32: Fix machine_spi_rate test on ESP32-C3.

projectgus:
### Summary

Update to the test added in #15543, changes the SPI pins for ESP32-C3. Found while testing #15523, not sure if it's a regression there or if I missed it when testing the first time (IO 18 and 19 are the native USB pins).

### Testing

Test passes on ESP32, ESP32-C3 and ESP32-S3.


dpgeorge:
Tested on an ESP32-C3-DevKitC-02:
- prior to this PR the `machine_spi_rate` test would fail
- after this PR it succeeds

PR #15660, opened 2024-08-14
ports/zephyr: Allow a BLE peripheral to create services at runtime.

danicampora:
This PR adds the required functionality for a peripheral to create services at runtime. Easiest way to test this is with the example code from:

https://github.com/micropython/micropython/blob/master/examples/bluetooth/ble_simple_peripheral.py


danicampora:
The commit message formatting check seems to be failing but not for the commit on this PR, for an older one, not sure what to do about it...

ports/zephyr/modbluetooth_zephyr.c
```diff
@@ -69,6 +121,102 @@ static struct k_timer mp_bluetooth_zephyr_gap_scan_timer;
...
+static void add_service(const struct bt_uuid *u, struct bt_gatt_attr *attr);
+static void add_characteristic(struct add_characteristic *ch, struct bt_gatt_attr *attr_chrc, struct bt_gatt_attr *attr_value);
+static void ccc_cfg_changed(const struct bt_gatt_attr *attr, uint16_t value);
+static void add_ccc(struct bt_gatt_attr *attr, struct bt_gatt_attr *attr_desc);
+static void add_cep(const struct bt_gatt_attr *attr_chrc, struct bt_gatt_attr *attr_desc);
+static void add_descriptor(struct bt_gatt_attr *chrc, struct add_descriptor *d, struct bt_gatt_attr *attr_desc);
+static void mp_bt_zephyr_gatt_indicate_done(struct bt_conn *conn, struct bt_gatt_indicate_params *params, uint8_t err);
+static struct bt_gatt_attr *mp_bt_zephyr_find_attr_by_handle(uint16_t value_handle);
+
+
```
dpgeorge, line 140: Please only use one blank line to separate sections.

ports/zephyr/modbluetooth_zephyr.c
```diff
@@ -404,6 +809,131 @@ int mp_bluetooth_gap_peripheral_connect_cancel(void) {
...
+    } else if (uuid->type == MP_BLUETOOTH_UUID_TYPE_32) {
+        bt_uuid_create(result, uuid->data, 4);
+    } else {    //  MP_BLUETOOTH_UUID_TYPE_128
+        bt_uuid_create(result, uuid->data, 16);
+    }
+    return result;
+}
+
+static void gatt_db_add(const struct bt_gatt_attr *pattern, struct bt_gatt_attr *attr, size_t user_data_len) {
+
```
dpgeorge, line 826: Please don't add blank lines at the start of functions.

ports/zephyr/modbluetooth_zephyr.c
```diff
@@ -404,6 +809,131 @@ int mp_bluetooth_gap_peripheral_connect_cancel(void) {
...
+            ch->permissions & GATT_PERM_MASK,
+            mp_bt_zephyr_gatts_attr_read, mp_bt_zephyr_gatts_attr_write, NULL), attr_value, 0);
+
+    chrc_data = attr_chrc->user_data;
+    chrc_data->properties = ch->properties;
+    chrc_data->uuid = attr_value->uuid;
+}
+
+static void ccc_cfg_changed(const struct bt_gatt_attr *attr, uint16_t value) {
+    // TODO
```
dpgeorge, line 888: Is it possible to raise an exception here, eg `OSError(MP_EOPNOTSUPP)`?  Or will that raise through zephyr's BLE bindings in a bad way?

ports/zephyr/mpconfigport.h
```diff
@@ -72,18 +71,22 @@
...
 #define MICROPY_PY_BLUETOOTH        (1)
+#define MICROPY_PY_BUILTINS_SET     (1)
+#define MICROPY_PY_STRUCT           (1)
 #ifdef CONFIG_BT_CENTRAL
 #define MICROPY_PY_BLUETOOTH_ENABLE_CENTRAL_MODE (1)
 #endif
 #define MICROPY_PY_BLUETOOTH_ENABLE_GATT_CLIENT (0)
+#else
+#define MICROPY_PY_BUILTINS_SET     (0)
+#define MICROPY_PY_STRUCT           (0)
```
dpgeorge, line 89: I think it's best to just remove these settings from this config file completely and let them take their default values at all times (which means they'll be enabled for all boards, which is fine, they are core parts of Python).

dpgeorge:
This looks great, thanks!  I'll need to test it though.

ports/zephyr/modbluetooth_zephyr.c
danicampora: OK

ports/zephyr/modbluetooth_zephyr.c
danicampora: OK

ports/zephyr/modbluetooth_zephyr.c
danicampora: No, that's not a good idea, I will better implement what's pending on that TODO.

ports/zephyr/mpconfigport.h
danicampora: OK, will do.

dpgeorge:
Can you please rebase this on latest master, to pick up the threading changes?

danicampora:
> Can you please rebase this on latest master, to pick up the threading changes?

@dpgeorge done! sorry for the delay.

danicampora:
@dpgeorge I just fixed a problem I discovered with gc memory on this. This is now ready for you to test/review.

dpgeorge:
I've run a few tests with this PR on a Nucleo WB55RG (after enabling BT for it on zephyr, following how it's done for nrf52840dk in this PR).

From what I understand, this PR allows the zephyr board to be a peripheral, advertise some services, and have a central connect to it.  So the multitests that I ran are the following.  Note the zephyr board is always instance 0 so that it's the peripheral.  Eg:
```
$ ./run-multitests.py -i pyb:a0 -i pyb:a1 -t multi_bluetooth/ble_gap_connect.py
```
where `/dev/ttyACM0` is the zephyr nucleo_wb55rg, and `/dev/ttyACM1` is another MicroPython board with BLE (PYBD-SF2 in my case but it shouldn't matter, an ESP32 or Pico W should work fine).

#### `multi_bluetooth/ble_gap_connect.py`

This will pass the first time, but run the second time in a row it will fail with `ValueError: Timeout waiting for 7`, and after that the zephyr board has locked up and must be hard reset.  I suspect there's an issue with soft reset not clearing out the zephyr BLE state.

Note that sometimes on the first run it fails with `ValueError: Timeout waiting for 1`, which means it's waiting for the central to connect to it.  Not sure why.

#### `multi_bluetooth/ble_gattc_discover_services.py`

This will pass the first time, but fail the second and on the third will either fail again with an error, or just lock up.

#### `multi_bluetooth/ble_gatt_data_transfer.py`

On the first run, this almost passes with the following diff of the expected output:
```diff
-gatts_read: b'central0central1central2central3'
+gatts_read: b'central3'
```
So somehow the data is being lost/overwritten.

On subsequent runs it either has a timeout or locks up.

Sometimes on the first run (after a hard reset) instance 1 (not the zephyr board) fails with `KeyError: (11, UUID('00000002-1111-2222-3333-444444444444'))` which means that it wasn't able to find this expected characteristic in the discovered services.  So there's something wrong with service discovery.

-------------

Other tests I either didn't try because I didn't expect them to pass, or they didn't pass after trying multiple times.

It's great that the above three tests can almost pass!  I think there's an issue with soft reset not clearing the services (if that's even possible...) and with that fixed we should be able to get some tests passing repeatedly and reliably.

ports/zephyr/modbluetooth_zephyr.c
```diff
@@ -56,9 +77,42 @@ enum {
...
+};
+
+typedef struct _mp_bt_zephyr_conn_t {
+    struct bt_conn *conn;
+    struct _mp_bt_zephyr_conn_t *next;
+} mp_bt_zephyr_conn_t;
+
 typedef struct _mp_bluetooth_zephyr_root_pointers_t {
+    // list of objects to be tracked by the gc
+    mp_obj_t objs_list;
```
dpgeorge, line 105: It might be better to use `m_tracked_calloc()` for this, rather than reimplement that functionality (unless you need specific functionality that `m_tracked_calloc()` is not providing).

danicampora: Thanks Damien. I see how `m_tracked_calloc()` works but it might be a bit of a hassle to free all the objects created during the service registrations (attributes, UUIDs, etc.) when de-initialising BLE, and with my approach I just have to make that `objs_list` variable point to NULL.

danicampora:
> It's great that the above three tests can almost pass! I think there's an issue with soft reset not clearing the services (if that's even possible...) and with that fixed we should be able to get some tests passing repeatedly and reliably.

Thanks for testing! @dpgeorge, working on fixing these problems now.

danicampora:
@dpgeorge I solved the problem with the soft reset and appending writes. I can not run those 3 tests that you mentioned over and over again.

I'm going to try to get other tests passing as well...

danicampora:
@dpgeorge I managed to make a few more tests passing, but when rebasing with master to use Zephyr 3.7.0, the behaviour of BLE has changed a bit, and now a gatt server only sends notifications if the clients are actually subscribed to it. In the past it would also send it either way (just like NimBLE seems to do). Now it doesn't, and this is a problem for many tests which expect notifications in several cases without any subscription. Not sure how to proceed now, any suggestions?

jonnor:
@dpgeorge do you have any feedback on dealing with the above test failures that @danicampora is seeing?

jonnor:
Support for Zephyr 4.x has now been merged. Perhaps some of the issues has been resolved there? Or can be?

dpgeorge:
I had trouble testing BLE with Zephyr 3.7, it just locked up when initializing the BLE.

But now with Zephyr 4.0.0 it seems to work again, I can at least do a GAP scan (on a nucleo_wb55rg).

So, this PR needs to be rebased and tested with the latest zephyr port.   Then we can decide how to proceed.

dpgeorge:
> I managed to make a few more tests passing, but when rebasing with master to use Zephyr 3.7.0, the behaviour of BLE has changed a bit, and now a gatt server only sends notifications if the clients are actually subscribed to it.

I did some investigation into this and found that an option has been added to disable this feature, see: https://github.com/zephyrproject-rtos/zephyr/commit/fae5dd36431106c328e000d676f1a3d8d987275f .  And it looks like this option was added in Zephyr 3.5.0.  So that's good, we can use it!

I did some local rebasing and testing of this PR, and it's looking really good.  I tested on `nucleo_wb55rg` with the following config additions:
```
CONFIG_BT_GATT_DYNAMIC_DB=y
CONFIG_BT_GATT_CLIENT=y
CONFIG_BT_L2CAP_TX_MTU=252
CONFIG_BT_BUF_ACL_RX_SIZE=256
CONFIG_BT_GATT_ENFORCE_SUBSCRIPTION=n
```
Note the last option, to fix the subscription issue.

Prior to this PR, only `multi_bluetooth/ble_gap_advertise.py` would pass (on zephyr), and it would pass both as instance0 and instance1.

Now with this PR, that test still passes, and also the following additional tests now pass when run as instance0 (ie in peripheral role):
```
multi_bluetooth/ble_descriptor.py
multi_bluetooth/ble_gap_connect.py
multi_bluetooth/ble_gap_device_name.py
multi_bluetooth/ble_gattc_discover_services.py
multi_bluetooth/ble_gatt_data_transfer.py
```
(When run as instance1 none of the above pass, but that's expected, it's just not implemented.)

-------------

So this PR is almost good to go.  The things that it needs are:
- rebase on latest master
- add the `CONFIG_BT_GATT_ENFORCE_SUBSCRIPTION=n` option
- revert the changes in this PR that were needed for the enforced subscription (I think that's the second last commit here, but not exactly sure it all should be reverted)

@danicampora are you able to do that?  If not I can work on it, just let me know, thanks!

danicampora:
@dpgeorge thanks for testing and the detailed info. I can work on this and update the PR so that it can be ready to merge.

ports/zephyr/modbluetooth_zephyr.c
```diff
@@ -312,24 +641,156 @@ int mp_bluetooth_gatts_write(uint16_t value_handle, const uint8_t *value, size_t
...
         return ERRNO_BLUETOOTH_NOT_ACTIVE;
     }
-    return MP_EOPNOTSUPP;
+
+    int err = MP_ENOENT;
+    mp_bt_zephyr_conn_t *connection = mp_bt_zephyr_find_connection(conn_handle);
+
+    if (connection) {
+        struct bt_gatt_attr *attr_val = mp_bt_zephyr_find_attr_by_handle(value_handle + 1);
+        mp_bluetooth_gatts_db_entry_t *entry = mp_bluetooth_gatts_db_lookup(MP_STATE_PORT(bluetooth_zephyr_root_pointers)->gatts_db, value_handle + 2);
```
dpgeorge, line 762: This `entry` variable is no longer needed.

In fact, I think the commit "ports/zephyr: Only send notifications/indications when subscribed." can be almost fully reverted, because you no longer need to create this additional database entry at `value_handle + 2`.  Is that correct?

danicampora: Yes, that's true, that `entry` variable is no longer needed. Will remove it.

But some of the changes in that commit are needed in my opinion, because in the case of the function `mp_bluetooth_gatts_write` you only want to send the notification/indication if the subscription is actually in place. While in the case of `mp_bluetooth_gatts_notify_indicate` we want to send the notification/indication regardless of the subscription state.

dpgeorge: > But some of the changes in that commit are needed in my opinion, because in the case of the function `mp_bluetooth_gatts_write` you only want to send the notification/indication if the subscription is actually in place

Yes, OK, let's keep it that additional functionality.

dpgeorge (approved):
I retested this on nucleo_wb55rg and it works really well.

Thanks @danicampora, this is a really nice addition, it brings zephyr closer to the other ports in terms of BLE functionality!

dpgeorge:
This PR was squashed and merged in fe28cd78fefde7da1bc7614982223ce9e522143a

(Hopefully GitHub detects that and automatically closes this.)

jonnor:
Has been merged, closing.

PR #15676, opened 2024-08-18
wiznet5k: Add support for IPv6.

greezybacon:
### Summary

This adds support for the WIZNET5K nic to use IPv6 with the LWIP stack. Additionally, if `LWIP_IPV6` is disabled, the device is configured to drop all IPv6 packets to reduce load on the MCU.

Changes borrowed in spirit from the LWIP "Mainloop mode ("NO_SYS")" page, https://www.nongnu.org/lwip/2_0_x/group__lwip__nosys.html

### Testing

Testing was performed on the RP2040 with the Wiznet W5500 chip.

- IPv4 address via dhcp4
- IPv6 address via autoconf6
- TCP network service via IPv4 any address ('0.0.0.0')
  - Was able to connect via both IPv4 and IPv6 address
- TCP network service via IPv6 any address ('::')
  - Was able to connect via both IPv4 and IPv6 address
- PING to both the IPv4 and IPv6 addresses
- PING to the IPv6 link-local address

What seems not to work

- DHCPv6 (I use a delegated IPv6 prefix). LWIP is not able to retrieve an address over DHCPv6. Tested with
```python
>>> nic = network.WIZNET5K()
>>> nic.active(True)
>>> nic.ipconfig(dhcp6=True)
>>> # Wait 30 seconds
>>> nic.ipconfig('addr6')
[('FE80::D2:CBFF:FE6D:6192', 48, 0, 0)]
```
I suspect this is a pre-existing issue with the `ipconfig` function and LWIP.

Positively, the device will discard all IPv6 packets if IPv6 support is disabled in the firmware.

### DIFF
--- a/extmod/network_wiznet5k.c
+++ b/extmod/network_wiznet5k.c
@@ -61,6 +61,7 @@
 #include "lwip/err.h"
 #include "lwip/dns.h"
 #include "lwip/dhcp.h"
+#include "lwip/ethip6.h"
 #include "netif/etharp.h"
 
 #define TRACE_ETH_TX (0x0002)
@@ -297,12 +298,20 @@ static err_t wiznet5k_netif_init(struct netif *netif) {
     netif->hwaddr_len = sizeof(netif->hwaddr);
     int ret = WIZCHIP_EXPORT(socket)(0, Sn_MR_MACRAW, 0, 0);
     if (ret != 0) {
-        printf("WIZNET fatal error in netifinit: %d\n", ret);
+        printf("WIZNET fatal error in netif_init: %d\n", ret);
         return ERR_IF;
     }
 
     // Enable MAC filtering so we only get frames destined for us, to reduce load on lwIP
     setSn_MR(0, getSn_MR(0) | Sn_MR_MFEN);
+
+    #if LWIP_IPV6
+    netif->output_ip6 = ethip6_output;
+    netif->flags |= NETIF_FLAG_MLD6;
+    #else
+    // Drop IPv6 packets if firmware does not support it
+    setSn_MR(0, getSn_MR(0) | Sn_MR_MIP6B);
+    #endif
 
     return ERR_OK;
 }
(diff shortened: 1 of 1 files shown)


extmod/network_wiznet5k.c
```diff
@@ -63,6 +63,10 @@
 #include "lwip/dhcp.h"
 #include "netif/etharp.h"
 
+#if LWIP_IPV6
+#include "lwip/ethip6.h"
+#endif
```
dpgeorge, line 68: I suggest to just include this header unconditionally.  That header has its own `#if LWIP_IPV6` guard.  That keeps the code here a tiny bit simpler / easier to read.

dpgeorge:
> DHCPv6 (I use a delegated IPv6 prefix). LWIP is not able to retrieve an address over DHCPv6

Using a Pico W, connecting as usual to a WiFi AP (just using `wlan.connect(SSID, KEY)`), I'm able to automatically obtain an IPv6 address.

But I suspect that's different to your case because you seem to have enabled `LWIP_IPV6_DHCP6`, and that's not enabled on Pico W.

Even after enabling `LWIP_IPV6_DHCP6` on Pico W I'm not sure how to test it.  The device is assigned an IPv6 address automatically even without first running `wl.ipconfig(dhcp6=True)`.  And after running that command, nothing changes (still has the same `wl.ipconfig('addr6')` result).

greezybacon:
To your point, yes, I enabled `LWIP_IPV6_DHCP6`, which probably isn't typically enabled. So my test is probably unnecessary.

To answer your question, I think the device constructor may setup IPv6 auto configuration. I believe you can bring the interface down and back up and then request a different addressing strategy:

```python3
nic = network.DEVICE()
nic.active(False)
nic.active(True)
nic.ipconfig(...)
```

felixdoerre:
> LWIP is not able to retrieve an address over DHCPv6

@greezybacon, stateful dhcpv6 seems not implemented in lwip: https://github.com/particle-iot/lwip/blob/master/src/include/lwip/opt.h#L2739

> DHCPv6 (I use a delegated IPv6 prefix)

using a delegated IPv6 prefix does not contradict to also use SLAAC. You can just use DHCPv6 to obtain the prefix and then redistribute it using router advertisements. And using SLAAC would work with lwip.

greezybacon:
Well, this helps the Wiznet work over SLAAC. I added the header unconditionally as requested.

dpgeorge:
This PR is good as-is, to enable IPv6 support.

Any further improvements can be made later, if they are needed.

PR #15678, opened 2024-08-19
zephyr: Add CI to run tests against zephyr with qemu.

dpgeorge:
### Summary

Prior to this PR, the zephyr port was only compiled under CI, but no tests were run.  There was some provision to run tests on zephyr using `make-bin-testsuite`, but this hasn't been working since the zephyr port moved fully to cmake (and it was never run under CI, only manually).

With this change, the zephyr port is tested against the standard test suite via the following scheme:
- the zephyr port is built with the `qemu_cortex_m3` board and the `prj_minimal.conf` configuration
- `qemu-system-arm` runs `zephyr.elf`
- the zephyr console is redirected to a pts/pty
- `tests/run-tests.py` is run in bare-metal mode against the pts/pty device
    
This allows testing the zephyr port as though it were a physical board attached over a serial port.

### Testing

CI has been added and should run and pass.

The existing tinytest framework for the zephyr port has been removed in favour of this new approach of running `qemu-system-arm` and redirecting console to a pts/pty device, then running `run-tests.py` against that serial device.  This makes running the tests just like any other bare-metal or unix port.

I don't think it's worth continuing to maintain the tinytest framwork on the zephyr port because it's currently broken and needs effort to fix, it's different to everything else, puts an extra burden on the maintainers, and one day won't fit the test suite because it will overflow flash.


tools/pyboard.py
```diff
@@ -237,7 +237,7 @@ def __init__(self, cmd):
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
         )
-        pty_line = self.subp.stderr.readline().decode("utf-8")
+        pty_line = self.subp.stdout.readline().decode("utf-8")
```
projectgus, line 240: Is there any chance this might break some other process that prints the PTY to stderr not stdout?

(Maybe fixable by changing `stderr=subprocess.STDOUT` in the call to Popen? Although if it's not something that anyone uses then might as well ignore.)

projectgus (approved):
Less code, more test coverage! :rocket:

tools/pyboard.py
dpgeorge: I was wondering why this was stderr to begin with... it looks like this commit used it: a9e6f08adb6dbe84ee59e9545895aa9dfc5ade17 , but that's since been removed due to zephyr moving to cmake.


(Note, this change is also needed by the qemu-arm port.)

dpgeorge: I've updated to use `stderr=subprocess.STDOUT`.

Issue #15465, opened 2024-07-14
super().__init__(*args, **kwargs) doesn't pass kwargs to parent C class.


kwagyeman:
### Port, board and/or hardware

openmv

### MicroPython version

v1.23.0

### Reproduction

See below.

### Expected behaviour

There should be no error.

### Observed behaviour

We have a class like this:

```
class Model(uml.Model):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
```

Its parent is a custom C class that implements `make_new` and accepts positional and keyword arguments.

When you pass just positional arguments to `Model()` in code using this subclass things work as expected. When you pass `kwargs` you get this error:

`TypeError: function doesn't take keyword arguments`

It appears MicroPython refuses to pass the `kwargs` to the parent class. Note that the error is not coming from code being executed in our `make_new` in C. It appears something else has rejected the `kwargs` being passed.

### Code of Conduct

Yes, I agree

ricksorensen:
This discussion on the micropython site might be pertinent:
https://github.com/orgs/micropython/discussions/10236

And the documentation on CPython differences:
https://docs.micropython.org/en/latest/genrst/modules.html#positional-only-parameters

kwagyeman:
Thanks, I think this clears it up.

stinos:
FWIW only did a quick test but this patch makes passing kwargs to a native base work for me; 

```
diff --git a/py/objtype.c b/py/objtype.c
index b6d600e94..e468ff116 100644
--- a/py/objtype.c
+++ b/py/objtype.c
@@ -84,14 +84,14 @@ static int instance_count_native_bases(const mp_obj_type_t *type, const mp_obj_t

 // This wrapper function is allows a subclass of a native type to call the
 // __init__() method (corresponding to type->make_new) of the native type.
-static mp_obj_t native_base_init_wrapper(size_t n_args, const mp_obj_t *args) {
+static mp_obj_t native_base_init_wrapper(size_t n_args, const mp_obj_t *args, mp_map_t *kw_args) {
     mp_obj_instance_t *self = MP_OBJ_TO_PTR(args[0]);
     const mp_obj_type_t *native_base = NULL;
     instance_count_native_bases(self->base.type, &native_base);
-    self->subobj[0] = MP_OBJ_TYPE_GET_SLOT(native_base, make_new)(native_base, n_args - 1, 0, args + 1);
[...]

PR #15688, opened 2024-08-21
py/objtype: Allow passing keyword arguments to native base __init__.

stinos:
### Summary

Allowing passing keyword arguments to a native base's `__init__` (i.e. `make_new` in the C code); previously only positional arguments allowed. See #15465 and https://github.com/orgs/micropython/discussions/10236 where this came up.

### Testing

Test has been added to the test suite.

Every call to the native base's `__init__` going to be result in allocating a `mp_map_t` on the C stack and initialize it with `mp_map_init_fixed_table` (see `fun_builtin_var_call`), even though `make_new` will only use the resulting table to access it's `used` member i.e. to get the number of keyword arguments.

### DIFF
--- a/py/objtype.c
+++ b/py/objtype.c
@@ -85,14 +85,14 @@ static int instance_count_native_bases(const mp_obj_type_t *type, const mp_obj_t
 
 // This wrapper function allows a subclass of a native type to call the
 // __init__() method (corresponding to type->make_new) of the native type.
-static mp_obj_t native_base_init_wrapper(size_t n_args, const mp_obj_t *args) {
+static mp_obj_t native_base_init_wrapper(size_t n_args, const mp_obj_t *args, mp_map_t *kw_args) {
     mp_obj_instance_t *self = MP_OBJ_TO_PTR(args[0]);
     const mp_obj_type_t *native_base = NULL;
     instance_count_native_bases(self->base.type, &native_base);
-    self->subobj[0] = MP_OBJ_TYPE_GET_SLOT(native_base, make_new)(native_base, n_args - 1, 0, args + 1);
+    self->subobj[0] = MP_OBJ_TYPE_GET_SLOT(native_base, make_new)(native_base, n_args - 1, kw_args->used, args + 1);
     return mp_const_none;
 }
-static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(native_base_init_wrapper_obj, 1, MP_OBJ_FUN_ARGS_MAX, native_base_init_wrapper);
+static MP_DEFINE_CONST_FUN_OBJ_KW(native_base_init_wrapper_obj, 1, native_base_init_wrapper);
 
 #if !MICROPY_CPYTHON_COMPAT
 static
--- a/py/obj.h
+++ b/py/obj.h
@@ -516,6 +516,9 @@ typedef mp_obj_t (*mp_fun_3_t)(mp_obj_t, mp_obj_t, mp_obj_t);
 typedef mp_obj_t (*mp_fun_var_t)(size_t n, const mp_obj_t *);
 // mp_fun_kw_t takes mp_map_t* (and not const mp_map_t*) to ease passing
 // this arg to mp_map_lookup().
+// Note that the mp_obj_t* array will contain all arguments, positional and keyword, with the keyword
+// ones starting at offset n, like: arg0 arg1 ... arg<n> key0 value0 key1 value1 ..., and the mp_map_t*
+// gets those same keyword arguments but as a map for convenience; see fun_builtin_var_call.
 typedef mp_obj_t (*mp_fun_kw_t)(size_t n, const mp_obj_t *, mp_map_t *);
 
 // Flags for type behaviour (mp_obj_type_t.flags)
--- a/py/objfun.c
+++ b/py/objfun.c
@@ -107,7 +107,9 @@ static mp_obj_t fun_builtin_var_call(mp_obj_t self_in, size_t n_args, size_t n_k
     if (self->sig & 1) {
         // function allows keywords
 
-        // we create a map directly from the given args array
+        // we create a map directly from the given args array; self->fun.kw does still
+        // expect args to have both positional and keyword arguments, ordered as:
+        // arg0 arg1 ... arg<n_args> key0 value0 key1 value1 ... key<n_kw> value<n_kw>
         mp_map_t kw_args;
         mp_map_init_fixed_table(&kw_args, n_kw, args + n_args);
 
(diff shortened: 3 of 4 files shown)


py/objtype.c
```diff
@@ -85,14 +85,14 @@ static int instance_count_native_bases(const mp_obj_type_t *type, const mp_obj_t
 
 // This wrapper function allows a subclass of a native type to call the
 // __init__() method (corresponding to type->make_new) of the native type.
-static mp_obj_t native_base_init_wrapper(size_t n_args, const mp_obj_t *args) {
+static mp_obj_t native_base_init_wrapper(size_t n_args, const mp_obj_t *args, mp_map_t *kw_args) {
     mp_obj_instance_t *self = MP_OBJ_TO_PTR(args[0]);
     const mp_obj_type_t *native_base = NULL;
     instance_count_native_bases(self->base.type, &native_base);
-    self->subobj[0] = MP_OBJ_TYPE_GET_SLOT(native_base, make_new)(native_base, n_args - 1, 0, args + 1);
+    self->subobj[0] = MP_OBJ_TYPE_GET_SLOT(native_base, make_new)(native_base, n_args - 1, kw_args->used, args + 1);
```
dpgeorge, line 92: I'm not sure how this works: when calling `make_new`, the keyword arguments must be passed in the same `args` array as the positional arguments.  But upon entry to this function the positional in `args` are separate to the keyword in `kw_args`.

Is it implicitly relying on the fact that the keyword args do actually sit in the `args` array, but only by accident?

stinos: > Is it implicitly relying on the fact that the keyword args do actually sit in the args array, but only by accident?

It's not by accident, the convention seems to be such that this is always the case. But there are 2 different ways in which this happens which is why my original comment about some things not being needed wasn't accurate because I only tested one case (I removed that part of the comment plus added extra tests to cover the case where `__init__` is not called with start arguments).

The part which this commit chages is indicated with `// ****`, rest is copy-paste of the actual code.

- calling `super().__init__()` or `super().__init__(pos0)` or `super().__init__(kw0=0)` or `super().__init__(pos0, kw0)` i.e. no star arguments, all result in dispatching `MP_BC_CALL_METHOD` which leads to the following call chain:

```
ENTRY(MP_BC_CALL_METHOD)
....

// args contains: fun  self/NULL  arg(0)  ...  arg(n_args-2)  arg(n_args-1)  kw_key(0)  kw_val(0)  ... kw_key(n_kw-1)  kw_val(n_kw-1)
mp_obj_t mp_call_method_n_kw(size_t n_args, size_t n_kw, const mp_obj_t *args);

// args contains, eg: arg0  arg1  key0  value0  key1  value1
mp_obj_t mp_call_function_n_kw(mp_obj_t fun_in, size_t n_args, size_t n_kw, const mp_obj_t *args) {
    // get the type
    const mp_obj_type_t *type = mp_obj_get_type(fun_in);
    // do the call
    if (MP_OBJ_TYPE_HAS_SLOT(type, call)) {
        return MP_OBJ_TYPE_GET_SLOT(type, call)(fun_in, n_args, n_kw, args);
    }
}

mp_obj_t fun_builtin_var_call(mp_obj_t self_in, size_t n_args, size_t n_kw, const mp_obj_t *args) {
    ...
    if (self->sig & 1) {
        // function allows keywords

        // we create a map directly from the given args array
        mp_map_t kw_args;
        mp_map_init_fixed_table(&kw_args, n_kw, args + n_args);

        // **** after this commit: actual call to native_base_init_wrapper(size_t n_args, const mp_obj_t *args, mp_map_t *kw_args)
        return self->fun.kw(n_args, args, &kw_args);

    } else {
        // function takes a variable number of arguments, but no keywords

        // **** before this commit: actual call to native_base_init_wrapper(size_t n_args, const mp_obj_t *args)
        return self->fun.var(n_args, args);
    }
}
```

- calling `super.__init__(*args)` or `super().__init__(**kwargs)` or combinations thereof i.e. anything with star arguments result in dispatching `MP_BC_CALL_METHOD_VAR_KW` which leads to the same call chain as above except that there's first an extra call to unpack the star arguments:

```
ENTRY(MP_BC_CALL_METHOD_VAR_KW):
 // We have following stack layout here:
 // fun self arg0 arg1 ... kw0 val0 kw1 val1 ... bitmap <- TOS
...

mp_obj_t mp_call_method_n_kw_var(bool have_self, size_t n_args_n_kw, const mp_obj_t *args) {
    mp_call_args_t out_args;
    mp_call_prepare_args_n_kw_var(have_self, n_args_n_kw, args, &out_args);
    mp_obj_t res = mp_call_function_n_kw(out_args.fun, out_args.n_args, out_args.n_kw, out_args.args);
}

...
```

So the main trade-off by this commit is that every call to the native base's make_new is going to be preceded by a call to `mp_map_init_fixed_table` even though most of what that does is unused and instead it merely serves as a way to pass the number of keyword arguments.

And the difference between `make_new` and 'normal' functions taking keyword arguments is that those are usually going to use the `kwargs` map directly and ignore that starting at `args + n_args` the keyword arguments also happen to be available.
Whereas make_new implementations which do use keyword arguments will usually repeat the
```
mp_map_t kwargs;
mp_map_init_fixed_table(&kwargs, n_kw, args + n_args)
```
part.

dpgeorge: Thanks for the detailed explanation.  You are right.

A different way to see the same thing is the following: if you use `MP_DEFINE_CONST_FUN_OBJ_KW` to wrap a C function, then that C function is always and only ever called from `fun_builtin_var_call`, along the path that has `self->sig & 1` true.  And since `fun_builtin_var_call` is passed the positional and kw arguments adjacently in an array, they are passed through to the wrapped C function.

So I think we can rely on this behaviour, and this PR is good.  But I suggest adding some comments:
- a comment just above the typedef for `mp_fun_kw_t` stating that, even though an `mp_map_t*` is passed, the `mp_obj_t*` array will actually contain the keyword arguments as well
- a comment in `fun_builtin_var_call` that `self->fun.kw` expects the `args` array to have both positional and keyword adjacent

(I guess one day we could change `mp_fun_kw_t`'s type so that it just takes the big `args` array and doesn't pass in a `mp_map_t`... but that's a big change.  Or could go the other way and make everything take a `mp_map_t`, but that's also a big change.)

stinos: Ok thanks for checking, added comments, hope that's fine.

dpgeorge: That looks good, thanks!

> So the main trade-off by this commit is that every call to the native base's make_new is going to be preceded by a call to mp_map_init_fixed_table even though most of what that does is unused and instead it merely serves as a way to pass the number of keyword arguments.

Regarding this: yes, there will now be a slight additional overhead calling native base `make_new`, but IMO that's worth it for the added benefit of being able to pass in keyword arguments.  Constructors are anyway always going to be relatively slow due to memory allocation, and aren't as performance critical as normal functions/methods.

dpgeorge (approved):

PR #15697, opened 2024-08-23
Nano ESP32: allow `machine.bootloader` to run either hardware or DFU bootloader 

pillo79:
### Summary

The `machine.bootloder`implementation on the Arduino Nano ESP32 is different from other Espressif-based boards in that it calls a custom "bootloader" (actually, a different application) which allows DFU uploads. This allows users to easily switch between Micropython and Arduino C environments. 

However in some cases the hardware bootloader is still useful - for example when a full flash is required. This PR introduces argument parsing of the `machine.bootloader()` implementation so that when it is called with `1` as argument, the hardware bootloader is started - in all other cases, as before, the Arduino bootloader will run.

### Testing

Tested on Arduino Nano ESP32:
- `machine.bootloader()` starts the Arduino bootloader (for use with `dfu-tool`)
- `machine.bootloader(1)` starts the ESP32-S3 hardware bootloader (for use with `esptool`)

### DIFF
--- a/ports/esp32/boards/ARDUINO_NANO_ESP32/board_init.c
+++ b/ports/esp32/boards/ARDUINO_NANO_ESP32/board_init.c
@@ -27,6 +27,7 @@
 #include <string.h>
 #include "py/mphal.h"
 #include "shared/tinyusb/mp_usbd_cdc.h"
+#include "modmachine.h"
 
 #include <esp_system.h>
 #include <esp_ota_ops.h>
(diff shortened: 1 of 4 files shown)


dpgeorge:
Thanks for the contribution.  This will need to wait until #15108 is merged.

pillo79:
@dpgeorge Rebased to include the now-merged #15108.

dpgeorge (approved):
Sorry this got lost.  Looks good!

PR #15708, opened 2024-08-24
shared/runtime/sys_stdio_mphal: Fix docstring for stdio.

timdechant:
### Summary
Docstring for stdio indicates "FileIO", which is a binary IO stream.
As detailed in adafruit#553, stdio is not binary by design;
its docstring should indicate "StringIO".
More detail in issue adafruit#9543.

### Testing
Pico build on Waveshare RP2040-Zero.

### DIFF
--- a/shared/runtime/sys_stdio_mphal.c
+++ b/shared/runtime/sys_stdio_mphal.c
@@ -54,7 +54,7 @@ static const sys_stdio_obj_t stdio_buffer_obj;
 
 static void stdio_obj_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     sys_stdio_obj_t *self = MP_OBJ_TO_PTR(self_in);
-    mp_printf(print, "<io.FileIO %d>", self->fd);
+    mp_printf(print, "<io.%s %d>", mp_obj_get_type_str(self_in), self->fd);
 }
 
 static mp_uint_t stdio_read(mp_obj_t self_in, void *buf, mp_uint_t size, int *errcode) {
@@ -122,7 +122,7 @@ static const mp_stream_p_t stdio_obj_stream_p = {
 
 MP_DEFINE_CONST_OBJ_TYPE(
     stdio_obj_type,
-    MP_QSTR_FileIO,
+    MP_QSTR_TextIOWrapper,
     MP_TYPE_FLAG_ITER_IS_STREAM,
     print, stdio_obj_print,
     protocol, &stdio_obj_stream_p,
--- a/tests/basics/sys_stdio.py
+++ b/tests/basics/sys_stdio.py
@@ -2,6 +2,14 @@
 
 import sys
 
+try:
+    sys.stdout
+    sys.stdin
+    sys.stderr
+except AttributeError:
+    print("SKIP")
+    raise SystemExit
+
 # CPython is more verbose; no need to match exactly
 
 print('TextIOWrapper' in str(sys.stdout))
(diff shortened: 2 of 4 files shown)


shared/runtime/sys_stdio_mphal.c
```diff
@@ -122,7 +122,7 @@ static const mp_stream_p_t stdio_obj_stream_p = {
 
 MP_DEFINE_CONST_OBJ_TYPE(
     stdio_obj_type,
-    MP_QSTR_FileIO,
+    MP_QSTR_StringIO,
```
dpgeorge, line 125: This should probably be `TextIOWrapper`, which matches CPython and also the unix port of MicroPython.

shared/runtime/sys_stdio_mphal.c
```diff
@@ -54,7 +54,7 @@ static const sys_stdio_obj_t stdio_buffer_obj;
 
 static void stdio_obj_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     sys_stdio_obj_t *self = MP_OBJ_TO_PTR(self_in);
-    mp_printf(print, "<io.FileIO %d>", self->fd);
+    mp_printf(print, "<io.StringIO %d>", self->fd);
```
dpgeorge, line 57: This will now print the wrong value for things like `sys.stdin.buffer`, which should remain a `FileIO`.

Could fix that with the following:
```c
mp_printf(print, "<io.%s %d>", mp_obj_get_type_str(self_in), self->fd);
```

dpgeorge (commented):
Thanks for the contribution.

It would be good to add some tests for this, to make sure it doesn't regress.  Are you able to do that?  I suggest a new test file `tests/basics/sys_stdio.py`.

shared/runtime/sys_stdio_mphal.c
timdechant: Ah, even better.  I'll update and add some tests.

timdechant:
hmmm... qemu-arm is failing my test:

> basics/sys_stdio.py: 
> Traceback (most recent call last):
>   File "<stdin>", line 7, in <module>
> AttributeError: 'module' object has no attribute 'stdout'

Did not know this was optional; updating test.

tests/basics/sys_stdio.py
```diff
@@ -0,0 +1,33 @@
...
+
+print('TextIOWrapper' in str(sys.stdout))
+print('TextIOWrapper' in str(sys.stderr))
+print('TextIOWrapper' in str(sys.stdin))
+
+print('TextIOWrapper' in str(type(sys.stdout)))
+print('TextIOWrapper' in str(type(sys.stderr)))
+print('TextIOWrapper' in str(type(sys.stdin)))
+
+# # .buffer member is optional
```
dpgeorge, line 23: I suggest putting the following code (uncommented) into a separate test `tests/basics/sys_stdio_buffer.py`, and skip that test by first checking if `sys.stdout.buffer` exists (like this file does for `sys.stdout`).

dpgeorge: You can also add a `tests/basics/sys_stdio_buffer.py.exp` file if the output doesn't match CPython.

timdechant: > separate test `tests/basics/sys_stdio_buffer.py`
Done.

> add a `tests/basics/sys_stdio_buffer.py.exp` file if the output doesn't match CPython
A weak match to CPython is appropriate for this.

dpgeorge (approved):

dpgeorge:
Squashed, rebased and merged in 455415b1e1199c4364e84cc31905452f7c4ec399

Thanks for the contribution!

PR #15714, opened 2024-08-25
inlineasm: Add inline assembler support for RV32.

agatti:
### Summary

This commit adds support for writing inline assembler functions when targeting a RV32IMC processor.

An almost complete test suite is also present, covering 95% of the opcodes (leaving out things like breakpoint triggers and the like).

### Testing

This was tested both on QEMU-RISCV and on an ESP32-C3 after manually enabling the inline assembler emitter in `mpconfigport.h`.

Due to the "interesting" limitations about immediates' size and registers set that vary from opcode to opcode, the code has a lot of checks for invalid input, along with (hopefully) helpful error messages as part of the exceptions it raises.  However, unless I'm doing something really wrong, I cannot seem to be able to trap those exceptions in tests.  That would be helpful for further validation of range and registers checking.

As mentioned in the commit message, this bit of code takes up a fair amount of .rodata space (last time I checked it was something like 4.5k) and thus it is not enabled by default unless it's running on anything that's not a microcontroller.  Since the Pico2 port has Arm inline assembler support enabled, this is also enabled for said port to maintain feature parity across MCU architectures.

Syntax-wise, there are a few things that may need a bit of explanation

* Compressed opcodes start with the `c.` prefix, which I replaced with `c_` - so for example `c.srai` turns into `c_srai`.
* When defining load/store opcodes the immediate is in the last operand, so `lw a0, 10(a1)` is defined as `lw(a0, a1, 10)`.  This is to simplify syntax parsing, as I don't think the original syntax can be parsed as such, and should be easier to grasp for folks coming from Arm assembler.
* Since writing RISC-V assembly gets painful quickly if you have to deal with inline constants, two meta opcodes have been implemented (following GAS), namely `li Rd, Imm32` and `la Rd, Label`.  The first loads the given immediate into the given register using as few bytes as possible, the second loads the address of the given label into the given register.
* `c_mov(Rd, Rs)` has been aliased to `mov(Rd, Rs)`.
* To keep code size down (most of the code in this PR is error checking anyway), there is no support for opcodes with multiple argument types (eg. `add(a0, a1)` becoming `C.ADD A0, A1` or `ADD(A0, A0, A1)` behind the scenes, or `add(a0, 100)` turning into `ADDI A0, A0, 100`).


agatti:
@dpgeorge I've resolved the conflict blocking this PR, but now the issue with Arm tests has gotten worse.  Before the qemu-arm refactoring it'd take longer than 30 seconds, but would still end.  Now it's been stuck for a couple hours already.

Am I missing something in [`e2b5402` (#15714)](https://github.com/micropython/micropython/pull/15714/commits/e2b540231db68e24f80feeebe0cd3030b88e0517) when moving files around?  I can't see what's wrong, unless there's some extra logic elsewhere I haven't updated.

dpgeorge:
> but now the issue with Arm tests has gotten worse. Before the qemu-arm refactoring it'd take longer than 30 seconds, but would still end. Now it's been stuck for a couple hours already.

It looks like this issue with CI is resolved now?  The qemu-arm CI job now passes without issue, it seems.

agatti:
Yes, that issue was indeed resolved - however the inlineasm tests for RV32 won't be picked up until #15784 is merged (and until this PR gets rebased once more).

agatti:
Here's the rebased version taking the merged `qemu` port into account.  I've run the tests locally in a VM being as close as possible to the CI environment, and on an ESP32C3 board.

Running tests for `MPS2_AN385`, `SABRELITE`, and `VIRT_RV32` qemu boards work fine, but `MICROBIT` and `NETDUINO2` targets do not work at all (even without this PR).  Are those boards meant to be compile-only?

`MICROBIT` test run:

<details>

```
dev@mpy-dev-rv32:/micropython/ports/qemu$ make BOARD=MICROBIT test V=1
python3 ../../py/makeversionhdr.py build-MICROBIT/genhdr/mpversion.h
test -e "../../lib/micropython-lib/README.md" || (echo -e "\033[1;31mError: micropython-lib submodule is not initialized.\033[0m Run 'make submodules'"; false)
python3 ../../tools/makemanifest.py -o build-MICROBIT/frozen_content.c -v "PORT_DIR=/micropython/ports/qemu" -v "MPY_DIR=../.." -v "BOARD_DIR=" -v "MPY_LIB_DIR=../../lib/micropython-lib" -b "build-MICROBIT" -f"-march=armv7m" --mpy-tool-flags="" "freeze('test-frzmpy', ('frozen_asm_thumb.py', 'frozen_const.py', 'frozen_viper.py', 'native_frozen_align.py'))"
cd ../../tests && ./run-tests.py --target qemu --device execpty:"qemu-system-arm -global nrf51-soc.flash-size=1048576 -global nrf51-soc.sram-size=262144 -machine microbit -nographic -monitor null -semihosting  -serial pty -kernel ../ports/qemu/build-MICROBIT/firmware.elf" --exclude 'inlineasm/rv32imc' 
# HANGS HERE
```

</details>

`NETDUINO2` test run:

<details>

```
dev@mpy-dev-rv32:/micropython/ports/qemu$ make BOARD=NETDUINO2 test V=1
python3 ../../py/makeversionhdr.py build-NETDUINO2/genhdr/mpversion.h
test -e "../../lib/micropython-lib/README.md" || (echo -e "\033[1;31mError: micropython-lib submodule is not initialized.\033[0m Run 'make submodules'"; false)
python3 ../../tools/makemanifest.py -o build-NETDUINO2/frozen_content.c -v "PORT_DIR=/micropython/ports/qemu" -v "MPY_DIR=../.." -v "BOARD_DIR=" -v "MPY_LIB_DIR=../../lib/micropython-lib" -b "build-NETDUINO2" -f"-march=armv7m" --mpy-tool-flags="" "freeze('test-frzmpy', ('frozen_asm_thumb.py', 'frozen_const.py', 'frozen_viper.py', 'native_frozen_align.py'))"
cd ../../tests && ./run-tests.py --target qemu --device execpty:"qemu-system-arm -machine netduino2 -nographic -monitor null -semihosting  -serial pty -kernel ../ports/qemu/build-NETDUINO2/firmware.elf" --exclude 'inlineasm/rv32imc' 
b'MicroPython v1.24.0-preview.297.g34fbf1d55 on 2024-09-06; netduino2 with STM32\r\nType "help()" for more information.\r\n>>> '
  File "/micropython/tests/./run-tests.py", line 1188, in <module>
    main()
  File "/micropython/tests/./run-tests.py", line 1084, in main
    pyb.enter_raw_repl()
  File "/micropython/tools/pyboard.py", line 365, in enter_raw_repl
    raise PyboardError("could not enter raw repl")
pyboard.PyboardError: could not enter raw repl
make: *** [Makefile:169: test] Error 1
```

</details>

dpgeorge:
> but `MICROBIT` and `NETDUINO2` targets do not work at all (even without this PR). Are those boards meant to be compile-only?

You are right, these boards don't run the tests on master.

Looking into it it seems the problems are:
- NETDUINO2: requires a `time.sleep(0.1)` in `tools/pyboard.py`, `ProcessPtyToTerminal::__init__`, just before creating the `serial.Serial(pty, ...)` instance.  That gives qemu enough time to start up before trying to access the serial port.
- MICROBIT: probably needs the same patch as NETDUINO2, but then also locks up running the `tests/feature_check/native_check.py`.  Solution would be to either debug why it crashes on that and fix it, or just disable the native emitter on this board.

This is unrelated to this PR and so out of scope.  But feel free to fix it if you like :smile:

agatti:
@dpgeorge fyi, MICROBIT crashes when accessing the function table, which doesn't make much sense.

```
...
(gdb) 
0x20000f74 in gc_heap ()
=> 0x20000f74 <gc_heap+3192>:   00 a8   add     r0, sp, #0
(gdb) 
0x20000f76 in gc_heap ()
=> 0x20000f76 <gc_heap+3194>:   d7 f8 b4 40     ldr.w   r4, [r7, #180]  ; 0xb4
(gdb) print/a $r7
$3 = 0x28fe8 <mp_fun_table>
(gdb) ni
0x0001ff8a in HardFault_Handler ()
=> 0x0001ff8a <HardFault_Handler+0>:    03 20   movs    r0, #3
(gdb) x/4a mp_fun_table+180
0x2909c <mp_fun_table+180>:     0x1b689 <mp_setup_code_state_native>    0x1bb9b <mp_small_int_floor_divide>     0x1bb7d <mp_small_int_modulo>   0xe23d <mp_native_yield_from>
```

agatti:
This should now work with the changes to the test runner made after 1.24.0.

This PR also makes the test runner able to detect if inline assembler support is enabled and if so for which architecture.  If so, the `tests/inlineasm/$INLINEASM_ARCH/` directory is automatically added to the test suite.  On esp32, qemu, and rp2 it will pick up the correct test files according to what arch is currently in use (allowing for future inlineasm additions like RV64 and other architectures/variants).

agatti:
I'm wondering on what's the policy for new test files after the introduction of `unittest`.  

Should the test files contained in this PR moved to that framework (as in, is it a blocking issue) or is it something that can be done at a later stage (not necessarily by myself, either)?

dpgeorge:
> I'm wondering on what's the policy for new test files after the introduction of `unittest`.

Using unittest is only for convenience, if it makes writing the test easier.  So you don't need to change anything in this PR.

ports/qemu/Makefile
```diff
@@ -19,16 +19,20 @@ QSTR_DEFS = qstrdefsport.h
 MICROPY_ROM_TEXT_COMPRESSION ?= 1
 
 ifeq ($(QEMU_ARCH),arm)
-FROZEN_MANIFEST ?= "require('unittest'); freeze('test-frzmpy')"
+QEMU_GC_HEAP_SIZE ?= 100
+FROZEN_MANIFEST ?= "require('unittest'); freeze('test-frzmpy', ('frozen_asm_thumb.py', 'frozen_const.py', 'frozen_viper.py', 'native_frozen_align.py'))"
 endif
 ifeq ($(QEMU_ARCH),riscv32)
-FROZEN_MANIFEST ?= "require('unittest'); freeze('test-frzmpy', ('frozen_const.py', 'frozen_viper.py', 'native_frozen_align.py'))"
+QEMU_GC_HEAP_SIZE ?= 120
```
dpgeorge, line 26: Can we just increase the heap size to 120k for all boards/archs?  That would keep it simple (if it's possible to do, if the boards all have enough RAM for that, even though it's all virtual).

I would prefer to keep it simple and use the one size.  But if you think it's needed (or a much better idea) to have this configurable, then I suggest a different name for the variable.  The prefix `QEMU_` indicates that it's a qemu-related config option, when it's actually a MicroPython option.  So maybe just `GC_HEAP_SIZE`, or `MICROPY_GC_HEAP_SIZE`.

agatti: Sharing the same default for both Arm and RV32 could work but then that value would need to be changed once more for #15931, to a value that may not reflect the reality of the physical boards the code under test will run on.  Unless RV64 ends up with a different default, and at that point it's pretty much the same as having a per-arch default.

Besides fixing the RV32 tests (for which I still have no idea why that happens), I've extensively used this feature to see how my code behaves in low-memory conditions, due to a simulated testing scenario, or due to a fragmented heap after long periods of (simulated) execution time.  I found it much faster to script a command line argument to `make` than messing with `CFLAGS`.

W.r.t. the prefix, you've got valid points, but on the other hand this flag is specific to the QEMU test runner.  I'd say that keeping the prefix sort of makes it more explicit that it is strictly a testing-only configuration flag, but I'm not 100% against any changes on that front.

agatti: No worries, I missed the bit about those other two ports - I rarely interact with minimal or with zephyr, if at all.

I've updated the user-facing name to `MICROPY_HEAP_SIZE` in both the code and the port's README file.

ports/qemu/main.c
```diff
@@ -34,7 +34,11 @@
 #include "shared/runtime/gchelper.h"
 #include "shared/runtime/pyexec.h"
 
-#define HEAP_SIZE (100 * 1024)
+#if MICROPY_HEAP_SIZE <= 0
+#error MICROPY_HEAP_SIZE must be a positive integer.
+#endif
+
+#define HEAP_SIZE (MICROPY_HEAP_SIZE * 1024)
```
dpgeorge, line 41: The minimal and zephyr ports specify `MICROPY_HEAP_SIZE` in bytes, so it should also be bytes here, not kibytes.

dpgeorge:
> * Compressed opcodes start with the `c.` prefix, which I replaced with `c_` - so for example `c.srai` turns into `c_srai`

It would be possible to allow `c.` as the prefix, because it's valid Python syntax.  It would just be a little bit of extra code to identify in the emitter.

Do you think it's worth trying to match the syntax more closely here, and use `c.` as the prefix?

Note that the Thumb emitter uses `_w` suffix for wide instructions which are `.w` in ARM syntax, eg `beq.w label` is `beq_w(label)` in MicroPython.  So for consistency maybe just worth sticking with `c_` for RISC-V. 

> * When defining load/store opcodes the immediate is in the last operand, so `lw a0, 10(a1)` is defined as `lw(a0, a1, 10)`.  This is to simplify syntax parsing, as I don't think the original syntax can be parsed as such, and should be easier to grasp for folks coming from Arm assembler.

It would be possible to support `lw(a0, 10(a1))`, because that's valid Python syntax.

Note that the Thumb emitter uses eg `ldr(r0, [r1, 12])` for specifying an offset to load/store, so it matches actual Thumb assembler (see `py/emitinlinethum.c:get_addr_arg`).  So I'm thinking it would be good to match RISC-V assembler as well and support `10(a1)` syntax.  It would be quite simple to do, just make a helper function like `get_addr_arg` that matches the `RULE_atom_expr_normal` with the second element of that node being a `RULE_trailer_paren`.

> * `c_mov(Rd, Rs)` has been aliased to `mov(Rd, Rs)`.

Do you mean that using `mov()` it will emit a `c_mov()` instruction?  (That makes more sense to me than the opposite.)


> * To keep code size down (most of the code in this PR is error checking anyway), there is no support for opcodes with multiple argument types

That sounds very reasonable.

---------

> the code has a lot of checks for invalid input, along with (hopefully) helpful error messages as part of the exceptions it raises
> However, unless I'm doing something really wrong, I cannot seem to be able to trap those exceptions in tests. That would be helpful for further validation of range and registers checking.

You can probably get these tests working by using `exec("<bad inline code>")`, eg:
```py
def test_asm(code):
    try:
        exec(code)
    except SyntaxError:
        print(...)

test_asm("""
@micropython.asm_rv32
def f(a0):
    mov(a100, a200)
""")
```

agatti:
> > * Compressed opcodes start with the `c.` prefix [...]

> Do you think it's worth trying to match the syntax more closely here, and use `c.` as the prefix?

Given that the Thumb emitter doesn't use dots in opcode names, as you point out, I'd say that leaving `c_` as a prefix right now is a decent compromise to get this out of the door sooner.

Maybe the dot notation can be introduced for both RISC-V and Thumb in v2, as some breakage may be expected then.

> It would be possible to support `lw(a0, 10(a1))`, because that's valid Python syntax.

Is it?  I thought integers were validated in their own specialised way in the code to speed things up - ints do not implement `__call__` so I expected that having an open parenthesis right after an int would trigger an error before the inline assembler parser would take control of the evaluation.  Good to know, I'll see what I can do for that, and thanks for the hints about the specialised parsing.

> Do you mean that using `mov()` it will emit a `c_mov()` instruction? (That makes more sense to me than the opposite.)

Indeed, `mov` isn't even an official RISC-V opcode - sorry for the unwanted confusion (it made sense in Italian :)).

> You can probably get these tests working by using `exec("<bad inline code>")`, eg: [...]

I'll give that a shot, thanks!

agatti:
`RULE_atom_expr_normal` unfortunately fails on expressions like `lw(a0, -4(sp))`, in that case it matches `RULE_factor_2`.  I'll work around that somehow.

agatti:
This is still missing the syntax error tests - I just pushed the changes so the C code changes can be reviewed as I'm not 100% sure whether the new parse node handling code is correct.

ports/qemu/main.c
agatti: This should be handled in the latest changes I pushed.

py/emitinlinerv32.c
```diff
@@ -0,0 +1,774 @@
...
+            opcode_qstr, node_index + 1));
+    return false;
+}
+
+static bool parse_register_offset_node(emit_inline_asm_t *emit, qstr opcode_qstr, const opcode_t *opcode_data, mp_parse_node_t node, mp_uint_t node_index, mp_parse_node_t *register_node, mp_parse_node_t *offset_node, bool *negative) {
+    assert(register_node != NULL && "Register node pointer is NULL.");
+    assert(offset_node != NULL && "Offset node pointer is NULL.");
+    assert(negative != NULL && "Negative pointer is NULL.");
+
+    if (!MP_PARSE_NODE_IS_STRUCT_KIND(node, PN_atom_expr_normal) && !MP_PARSE_NODE_IS_STRUCT_KIND(node, PN_factor_2)) {
```
dpgeorge, line 508: This validation logic can be simplified, in pseudo code:
```
if MP_PARSE_NODE_IS_STRUCT_KIND(node, PN_factor_2):
    if MP_PARSE_NODE_IS_TOKEN_KIND(node_struct->nodes[0], MP_TOKEN_OP_MINUS):
        *negative = true;
    elif MP_PARSE_NODE_IS_TOKEN_KIND(node_struct->nodes[0], MP_TOKEN_OP_PLUS):
        // pass
    else:
        goto invalid;
    node = node_struct->nodes[1]

if !MP_PARSE_NODE_IS_STRUCT_KIND(node, PN_atom_expr_normal):
    goto invalid
```

Note that `PN_factor_2` and `PN_atom_expr_normal` always have 2 elements so there's no need to check that.

dpgeorge:
When an rv32 inline-asm function is entered into, what registers are automatically saved?  Is there anything else done in the prelude?

With Thumb inline-asm, a few registers are saved/restored on function entry/exit.  This was arguably not a good choice, it might have been a better choice to have a very minimal prelude for inline-asm function.

py/emitinlinerv32.c
agatti: Thanks, I ended up doing something similar, but I left the elements count check.  I'll take that out.

Right now I'm cleaning up the range checks, as negative offsets are still seen as positive due to the offset node being without its minus sign and you can imagine the rest.  Right now there are lots of repeated code for the special cases and I'm merging together.

agatti:
> When an rv32 inline-asm function is entered into, what registers are automatically saved? Is there anything else done in the prelude?

The prelude saves the function table pointer and the three local registers.  The reason behind that is purely because they're stored in S-registers and those registers must be restored on exit (in other words, A0-A7 and T0-T6 can be trashed as much as you want, but S0-S11 must be preserved).

In theory T-registers or even the A4 to A7 could be have been used for those, but then half of the compressed opcodes wouldn't be able to access them.

In fact, the emitter just calls `asm_rv32_entry`/`asm_rv32_exit`	and one of the changes to `py/asmrv32.c` present in this PR is to save an internal temporary S-register only whenever there's an emitted function call (that's required to shorten most function calls to just 2 opcodes/4 bytes).

For inline assembly code it's up to the user to keep track of either RA or S0-S11, but that's expected of the user in the first place no matter the assembler used.

agatti:
OK, this should be it I guess.  I wish the offset-to-register syntax support wasn't bolted on this way, but it works :)  That can be cleaned up for v2 anyway.

dpgeorge:
> the offset-to-register syntax support wasn't bolted on this way

What do you mean, that you'd prefer it was designed it from the start?

agatti:
Indeed, everything else fits in nicely without special cases and exceptions.  It's more of a personal preference, really, but re-design everything to cover that case in a generic way as well would take a bit more time and delay this PR for the 1.25.0 milestone.

Anyway, I'll rework that for v2 when adding dotted opcodes support - in the end it works and the code isn't *that* bad, but it could be better for sure :)

dpgeorge:
OK, sounds fine to me.  I'm glad that the syntax `10(a0)` is now supported, I think that's a good improvement.

dpgeorge:
> The prelude saves the function table pointer and the three local registers.

Just to be clear, this is necessary for @native/@viper code generation.  But for inline-rv32-asm it's not strictly needed, but done just because inline-rv32-asm is using `asm_rv32_entry`/`asm_rv32_exit` for convenience.  Is that true?

Would it be possible for inline-rv32-asm to use a different entry/exit set of functions and not save any registers automatically?  As you say, it's up to the assembler user to keep track of and restore needed registers.

agatti:
Done.  Maybe when more targets will support `cm.push`/`cm.popret` opcodes (right now only the pico2 does) it can be feasible to automatically save/restore RA + S0..S11 with an explicit opt-out via an argument to the decorator (that's 2 opcodes/4 bytes vs ~60 bytes/~25 opcodes).

agatti:
Apologies for the extra push, it's just a fix for a minor issue I found when reviewing the code for the last time (the check for negative values' range didn't take into account that the negative flag was a pointer), plus some commented test code was removed.

py/asmrv32.c
```diff
@@ -126,7 +100,6 @@ static void split_immediate(mp_int_t immediate, mp_uint_t *upper, mp_uint_t *low
     // Turn the lower half from unsigned to signed.
     if ((*lower & 0x800) != 0) {
         *upper += 0x1000;
-        *lower -= 0x1000;
```
dpgeorge, line 129: Is removing this line intentional?  Is it a bug fix or refactoring?

tests/ports/qemu/asm_test.py
```diff
@@ -1,4 +1,13 @@
-import frozen_asm
+import sys
+
+arch = getattr(sys.implementation, "_mpy", 0) >> 10
+if arch >= 3 and arch <= 8:
+    import frozen_asm_thumb as frozen_asm
+elif arch == 11:
+    import frozen_asm_rv32 as frozen_asm
+else:
+    raise ImportError
```
dpgeorge, line 9: Would it be simpler and equivalent to do:
```py
try:
    import frozen_asm_thumb as frozen_asm
except ImportError:
    try:
        import frozen_asm_rv32 as frozen_asm
    except ImportError:
        raise ImportError
```
?

Then you don't need to know about the different arch values.

py/emitinlinerv32.c
```diff
@@ -0,0 +1,818 @@
...
+    { MP_QSTR_la,         opcode_la,                 CALL_RL,  R,   0, L,   0,  N,  0, 0xFFFFFFFF, 0xFFFFFFFF, 0x00000000 },
+    { MP_QSTR_lb,         asm_rv32_opcode_lb,        CALL_RIR, R,   0, I,   0,  R,  0, 0xFFFFFFFF, 0x00000FFF, 0xFFFFFFFF },
+    { MP_QSTR_lbu,        asm_rv32_opcode_lbu,       CALL_RIR, R,   0, I,   0,  R,  0, 0xFFFFFFFF, 0x00000FFF, 0xFFFFFFFF },
+    { MP_QSTR_lh,         asm_rv32_opcode_lh,        CALL_RIR, R,   0, I,   0,  R,  0, 0xFFFFFFFF, 0x00000FFF, 0xFFFFFFFF },
+    { MP_QSTR_lhu,        asm_rv32_opcode_lhu,       CALL_RIR, R,   0, I,   0,  R,  0, 0xFFFFFFFF, 0x00000FFF, 0xFFFFFFFF },
+    { MP_QSTR_li,         opcode_li,                 CALL_RI,  R,   0, I,   0,  N,  0, 0xFFFFFFFF, 0xFFFFFFFF, 0x00000000 },
+    { MP_QSTR_lui,        asm_rv32_opcode_lui,       CALL_RI,  R,   0, I,   12, N,  0, 0xFFFFFFFF, 0xFFFFF000, 0x00000000 },
+    { MP_QSTR_lw,         asm_rv32_opcode_lw,        CALL_RIR, R,   0, I,   0,  R,  0, 0xFFFFFFFF, 0x00000FFF, 0xFFFFFFFF },
+    { MP_QSTR_mov,        asm_rv32_opcode_cmv,       CALL_RR,  R,   0, R,   0,  N,  0, 0xFFFFFFFF, 0xFFFFFFFF, 0x00000000 },
```
dpgeorge, line 310: I suggest calling this pseudo/convenience opcode `mv`.

That matches GAS (as far as I can tell, which aliases `mv` to `c.mv`) and also matches the fact that `c.mv` is called `c.mv` rather than `c.mov`.

dpgeorge (commented):
This is looking very good now.  I particularly like how the tests suite can now auto-detect the inline-asm capability and arch, that's nice!

I have a few final comments and then it should be good to go in.

py/asmrv32.c
agatti: More like both a bug fix and a refactoring.  `*lower` would contain the lowest 12 bits of a 32 bits immediate, so applying a further offset isn't correct in that case even if it leaves the lowest part alone; however the variable has its upper 20 bits masked out on every usage so the extra offset didn't really matter.

tests/ports/qemu/asm_test.py
agatti: Makes sense.  This should be fixed in the latest iteration of this PR.

py/emitinlinerv32.c
agatti: I reckoned GAS used `mov`, but I've always used `c.mv` or `add(rd, rm, zero)` myself depending on my needs.  Anyway, I've just checked and GAS indeed uses `mv`, thanks for spotting that!  

This should be fixed now.

py/asmrv32.c
dpgeorge: OK

dpgeorge (approved):
Tested on `RPI_PICO2` in RISC-V mode.  Works well.

dpgeorge:
@agatti thanks for your efforts on this, it's a very nice feature to have, especially because RISC-V will become more and more prevalent in MCUs in the future.

BTW, if you want to have a go at reducing the code size, you should be able to cut off at least 1k with the following tricks:
- use `qstr_short_t` for the type of `REGISTERS_QSTR_TABLE`
- for the `argumentX_mask` variables of `opcode_t`, make them indirect enums into a table of masks (because there are only a limited set of possible masks), eg `enum { MASK_00000000, MASK_00000FFF, MASK_FFFFFFFF, ...}` then access via `MASK_TABLE[opcode->argument1_mask]`.
- combined with the above, use `qstr_short_t` for the `qstring` entry in `opcode_t` and 5-bit fields for `argumentX_mask`, which should allow fitting the qstr and the 3 argument masks in 32 bits

agatti:
>     * use `qstr_short_t` for the type of `REGISTERS_QSTR_TABLE`

Oh, I don't think I've ever seen `qstr_short_t` being used before.  So I must assume there's a hard limit of 64k qstrs?

>     * for the `argumentX_mask` variables of `opcode_t`, make them indirect enums into a table of masks [...].

That's nice - wish I've thought of that :)  This will end up as a PR after 1.25.0 is out, so it can be featured in a later version along with eventual fixes for bugs that may be found in the meantime.

All things considered, reducing the footprint by 30% or more may probably make this a candidate for being enabled by default on ESP32Cx targets too.  I can probably shorten some error messages' text and coalesce some messages as well to further reduce the space taken up by this feature, I'll see what can be done on that.

Anyway, thanks for the help on this - glad it's been merged!

dpgeorge:
> I can probably shorten some error messages' text and coalesce some messages as well to further reduce the space taken up by this feature,

Note that there are configurable error message levels:
```
// Exception messages are removed
#define MICROPY_ERROR_REPORTING_NONE     (0)                                   
// Exception messages are short static strings                                 
#define MICROPY_ERROR_REPORTING_TERSE    (1)                                 
// Exception messages provide basic error details     
#define MICROPY_ERROR_REPORTING_NORMAL   (2)                                 
// Exception messages provide full info, e.g. object names                       
#define MICROPY_ERROR_REPORTING_DETAILED (3)                                                                                
```
The default for most ports is NORMAL.  It's possible to provide different error messages for the different levels.

But actually I think what you have done already is pretty good, it's a much better user experience with more detailed error messages for tricky things like inline assembler.  Maybe it could be tweaked to reduce size by a little, eg I just noticed a few messages end in `.` which should be removed.

PR #15719, opened 2024-08-26
samd: A few minor improvements.

robert-hh:
- Fix a execution order bug in mp_hal_ticks_us_64().
     The upper 32 bit of the 64 bit ticks register was taken before  disabling the interrupts. That may have caused a wrong return value. 
- Simplify mp_hal_delay_ms().
    Do NOT use mp_hal_delay_us() for short delays. This was initially done to make short delays precise, but mp_hal_delay_us() does not allow for scheduling. Leave using mp_hal_delay_us() to user code if needed.

Both changes do not affect the API of the calls.


dpgeorge (approved):

dpgeorge:
Thanks, this is good, the UART test is now simpler!

Issue #15670, opened 2024-08-16
Binary operations on undefined variables crash the native emitter.


agatti:
### Port, board and/or hardware

unix

### MicroPython version

MicroPython v1.24.0-preview.206.ge9814e987.dirty on 2024-08-16; linux [GCC 14.2.1] version

### Reproduction

Run the following code:

```python
@micropython.native
def f():
    a += 0  # Or anything else, really
f()
```

### Expected behaviour

A `NameError` exception should be raised.  (CPython raises `NameError: name 'a' is not defined`)

### Observed behaviour

This is the backtrace of the crash when running the interpreter under GDB:

```
Program received signal SIGSEGV, Segmentation fault.
mp_obj_get_type (o_in=o_in@entry=0x0) at ../../py/obj.c:61
61              return o->type;
(gdb) bt
#0  mp_obj_get_type (o_in=o_in@entry=0x0) at ../../py/obj.c:61
#1  0x0000555555591333 in mp_binary_op (op=MP_BINARY_OP_INPLACE_ADD, lhs=0x0, rhs=0x1) at ../../py/runtime.c:630
#2  0x00007ffff7fc0053 in ?? ()
#3  0x00007ffff7a06440 in ?? ()
#4  0x00007ffff7fc006d in ?? ()
#5  0x00007fffffff9610 in ?? ()
#6  0x0000000000000003 in ?? ()
#7  0x00007fffffff9760 in ?? ()
#8  0x0000000000000000 in ?? ()
```

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

mendenm:
Is anyone working on this bug?  I can confirm it still exists:

```
@micropython.native
def rgb_hsv():
    a = 120 - a    
x = rgb_hsv()
```

crashes.  It took me a long time to find the typo in my code that was hard-hanging the machine.  In the real code, `a = 120 - a` was supposed to be `a = 120 - b`.

Version: MicroPython v1.24.1 on 2024-11-29; Raspberry Pi Pico with RP2040

agatti:
I'm still (slowly) working on a more general fix for this issue but it will take a while as it changes a quite a bit of things inside the emitter/compiler code.

In the meantime your best bet is to double down on static analysis to catch these issues at "compile" time.

peterhinch:
Or test the code before adding the decorator.

mendenm:
@peterhinch Test the code?  Never!  ;-)

PR #15723, opened 2024-08-26
py/runtime: Check for null objects coming from native functions.

agatti:
### Summary

When performing inplace binary operations (eg +=, -=, etc) from a native emitted context, the left hand side operand would be NULL if it wasn't defined earlier.  The code assumed the operand would always be non-NULL and hence a crash occurred when accessing the operand object.

This commit turns that crash into an exception being raised, with the same name and message as if the same error occurred in a bytecode context.  Also, the exception is raised at runtime, following the same behaviour as if the function was not decorated as native.

To reproduce the issue just try to execute the following Python code:

```python
@micropython.native
def f():
	a += 10

f()
```

Whilst this is a minor issue, the very fact that it would crash the whole interpreter because of a typo in a native function makes it quite annoying to have...

This fixes issue #15670.

### Testing

This was tested on the Unix port, but the issue would happen on any target that supports native compilation.

This patch is only enabled if `MICROPY_ENABLE_COMPILER` is enabled, and should add a hundred bytes or so of code.  Rather than patching the runtime this issue could maybe be fixed when compiling the incorrect statement that emits the binary operation, but I'm quite new to that part of the codebase.  At least it follows the same behaviour as the VM, better than nothing...

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -500,6 +500,12 @@
 #define MICROPY_COMP_RETURN_IF_EXPR (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_EXTRA_FEATURES)
 #endif
 
+// Whether to enable extra arguments validation when running native code
+// Requires MICROPY_DYNAMIC_COMPILER and costs 160 bytes (Thumb2)
+#ifndef MICROPY_COMP_EXTRA_VALIDATION
+#define MICROPY_COMP_EXTRA_VALIDATION (0)
+#endif
+
 /*****************************************************************************/
 /* Internal debugging stuff                                                  */
 
--- a/py/runtime.c
+++ b/py/runtime.c
@@ -285,6 +285,12 @@ void mp_delete_global(qstr qst) {
 mp_obj_t mp_unary_op(mp_unary_op_t op, mp_obj_t arg) {
     DEBUG_OP_printf("unary " UINT_FMT " %q %p\n", op, mp_unary_op_method_name[op], arg);
 
+    #if MICROPY_COMP_EXTRA_VALIDATION
+    if (arg == MP_OBJ_NULL) {
+        goto unbounded_local_variable;
+    }
+    #endif
+
     if (op == MP_UNARY_OP_NOT) {
         // "not x" is the negative of whether "x" is true per Python semantics
         return mp_obj_new_bool(mp_obj_is_true(arg) == 0);
@@ -362,10 +368,21 @@ mp_obj_t mp_unary_op(mp_unary_op_t op, mp_obj_t arg) {
             mp_unary_op_method_name[op], mp_obj_get_type_str(arg));
         #endif
     }
+
+    #if MICROPY_COMP_EXTRA_VALIDATION
+unbounded_local_variable:
+    mp_raise_msg(&mp_type_NameError, MP_ERROR_TEXT("cannot access local variable"));
+    #endif
 }
 
 mp_obj_t MICROPY_WRAP_MP_BINARY_OP(mp_binary_op)(mp_binary_op_t op, mp_obj_t lhs, mp_obj_t rhs) {
     DEBUG_OP_printf("binary " UINT_FMT " %q %p %p\n", op, mp_binary_op_method_name[op], lhs, rhs);
+
+    #if MICROPY_COMP_EXTRA_VALIDATION
+    if (lhs == MP_OBJ_NULL || rhs == MP_OBJ_NULL) {
+        goto unbounded_local_variable;
+    }
+    #endif
 
     // TODO correctly distinguish inplace operators for mutable objects
     // lookup logic that CPython uses for +=:
@@ -685,6 +702,11 @@ unsupported_op:
 
 zero_division:
     mp_raise_msg(&mp_type_ZeroDivisionError, MP_ERROR_TEXT("divide by zero"));
+
+    #if MICROPY_COMP_EXTRA_VALIDATION
+unbounded_local_variable:
+    mp_raise_msg(&mp_type_NameError, MP_ERROR_TEXT("local variable referenced before assignment"));
+    #endif
 }
 
 mp_obj_t mp_call_function_0(mp_obj_t fun) {
@@ -709,6 +731,12 @@ mp_obj_t mp_call_function_n_kw(mp_obj_t fun_in, size_t n_args, size_t n_kw, cons
 
     DEBUG_OP_printf("calling function %p(n_args=" UINT_FMT ", n_kw=" UINT_FMT ", args=%p)\n", fun_in, n_args, n_kw, args);
 
+    #if MICROPY_COMP_EXTRA_VALIDATION
+    if (fun_in == MP_OBJ_NULL) {
+        goto missing_function;
+    }
+    #endif
+
     // get the type
     const mp_obj_type_t *type = mp_obj_get_type(fun_in);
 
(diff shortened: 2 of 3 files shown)


agatti:
Added test to keep codecov happy...

dpgeorge:
Thanks for attempting to fix this.

The reason this hasn't been fixed before is because (1) it's difficult to do properly, and (2) it will lead to a decrease in performance (for either just native code, or both native and bytecode).

There are actually quite a few tests which test for unbound locals, eg `tests/basics/unboundlocal.py`.  All of these tests are currently disabled when the native emitter is used, because they crash it.

For example, how about `x = ~x`, or `print(x)`?

agatti:
Thanks for the explanation!

From some random testing I've made the only other place where I was able to reproduce the issue on the REPL is when handling unary operations (like in `x = ~x` as you mentioned) and can be fixed in the same way as binary operations were.  Everything else was handled properly, but I'll take a deeper look at the test cases you mentioned - if I find the time I may even try to see whether a few hours of `afl-fuzz` yield different crash cases.

Regarding the performance hit, maybe adding a NULL check in a couple of places can probably end up too small of a slowdown to be really noticed - those checks can be optimised/moved around by the C compiler, after all.  If performance in those bits still critical even after the compiler did its magic, would putting those extra checks being an additional define work?

agatti:
Ok, I've let `afl-fuzz` go wild for a couple of hours using `annotate_var`, `del_deref`, `del_local`, `scope_implicit`, and `unboundlocal` with all functions marked as native and `native_misc` with the extra test I added first, and all it found were crashes in `mp_unary_op`.  Once that was patched too, nothing interesting was found so far but I'll let it go until tomorrow morning to see if it can come up with something else.

Regarding `unboundlocal`, I ran it on a patched interpreter after marking both functions as native, and I saw no crashes.  Is there any other invalid behaviour I should have noticed?

dpgeorge:
I'm surprised fuzzing didn't find more cases.  For example the following easily crashes the unix port:
```py
@micropython.native
def f():
    print(x)
    x = 1
f()
```

agatti:
> I'm surprised fuzzing didn't find more cases. For example the following easily crashes the unix port:

It started finding a few more when I upgraded AFL++ to the latest stable version, but just two or three more cases.

Anyway, I've cleaned this up and put the whole lot behind its own define, not enabled by default.  Maybe this should be brought in when the `DEBUG` variable is set in the makefile or something like that.

py/runtime.c
```diff
@@ -285,6 +285,12 @@ void mp_delete_global(qstr qst) {
 mp_obj_t mp_unary_op(mp_unary_op_t op, mp_obj_t arg) {
     DEBUG_OP_printf("unary " UINT_FMT " %q %p\n", op, mp_unary_op_method_name[op], arg);
 
+    #if MICROPY_COMP_EXTRA_VALIDATION
+    if (arg == MP_OBJ_NULL) {
+        goto unbounded_local_variable;
```
stinos, line 290: Why `goto` instead of a direct call?

py/mpconfig.h
```diff
@@ -500,6 +500,12 @@
 #define MICROPY_COMP_RETURN_IF_EXPR (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_EXTRA_FEATURES)
 #endif
 
+// Whether to enable extra arguments validation when running native code
+// Requires MICROPY_DYNAMIC_COMPILER and costs 160 bytes (Thumb2)
+#ifndef MICROPY_COMP_EXTRA_VALIDATION
```
stinos, line 505: 'COMP' is 'COMPILER' here, right? How does this require MICROPY_DYNAMIC_COMPILER? Still this doesn't seem to describe well what is actually implemented here: adding checks in runtime.c means they get are used whenever running any type of code, be it Python or native or calling the functions directly from C. To me this seems as if it should be just called 'extra argument validation'. But then I'm wondering if this shouldn't be an 'all or nothing' approach. I didn't check in detail, but it seems not unlikely that not all possible cases have been covered yet?

agatti: The rationale is that the issues these checks handle seem to come only from emitted native code, hence the "dependency" on MICROPY_DYNAMIC_COMPILER and why it is disabled by default.  That should be probably be reworded, as you point out.  If you're not using any native emitters or you already have other error checking in place at the source level then you don't need this.  The name has COMP in it because I didn't find any better stanza for the define in the configuration file and I just followed the naming convention.

It is true this may not cover all possible cases, but it handles all the ones I've bumped into plus a few more found by fuzzing after the first @dpgeorge comment.  I'm quite new to that bit of MicroPython internals after all.

dpgeorge:
I really don't think this is the right approach.  There are so many combinations / cases of unbound variables being passed in as arguments.  Basically every C function needs to check it's arguments if they are `MP_OBJ_NULL` or not.

stinos:
> Basically every C function needs to check it's arguments if they are MP_OBJ_NULL or not.

Just to be clear: is it so that micropython.native emits bytecode and hence essentially all functions which can get called in vm.c would need a check? Which in turn is indeed pretty much everything from the public API?

dpgeorge:
> is it so that micropython.native emits bytecode and hence essentially all functions which can get called in vm.c would need a check?

Not exactly.

The point is, if a local variable is not populated in a `@micropython.native` function then it has an undefined value (probably zero'd out to `MP_OBJ_NULL`), and that null object can be passed anywhere a normal object may be passed, via native calls to functions in `mp_fun_table`.

So to cover every case I think it's sufficient if every function in `mp_fun_table` checks it's arguments for non-null.

stinos:
> every function in mp_fun_table

Ok thanks, that's what I wanted to know.

agatti:
> I really don't think this is the right approach. There are so many combinations / cases of unbound variables being passed in as arguments. Basically every C function needs to check it's arguments if they are `MP_OBJ_NULL` or not.

Fair point, but I can only think of two other ways to solve this, and I suspect both of those have already been considered before.

The first one is to emit the NULL check along with the native opcodes, as the number and types of arguments for each function in `mp_fun_table` is known beforehand and can be looked up from a table somehow, the other one would be to perform checks at the level before the actual opcodes are emitted - for the latter case I can't say I know where those bits should go or how they should look like, though.

I'd say the first one is slightly more approachable, but maybe it's because I've recently worked with native emitters.  It may very well be I'm missing something important to invalidate that assumption.

So, is there any realistic way to sort this out?

agatti:
I guess I can just emit the necessary null checks when recompiling the debug/json emitter output, so MicroPython emitters/runtime can be left untouched.  This probably is the better option, so I guess this can be closed as the solution proposed here didn't go anywhere :)

dpgeorge:
Just to follow up with some more examples that don't work with the PR in it's current state:
```py
@micropython.native
def f():
    x = 0
    print(x, y)
    y = 1
f()
```
and
```py
@micropython.native
def f():
    l = []
    l.insert(0, x)
    print(l)
    x = 0
f()
```
Both those examples crash the unix port with this PR and `MICROPY_COMP_EXTRA_VALIDATION=1`.

> So to cover every case I think it's sufficient if every function in `mp_fun_table` checks it's arguments for non-null.

Given the `list.insert` example above, that original statement of mine is wrong.  Basically every single native function needs to have a NULL check for all of its arguments.

> So, is there any realistic way to sort this out?

I don't think so.  All known ways to fix this are difficult:
1. put NULL checks everywhere, for all input arguments to all C functions
2. put NULL checks in the code generated by the native emitter
3. do compile time analysis to work out which variables may be accessed before being assigned

agatti:
Thank you for the follow-up!

In the recompiler I went for option 2, adding the NULL checks in the compiled output on each emitted call.

To match the function call name to its table index I already need to have the call table in some form anyway, I just added argument types to that representation.  With that in place then it's relatively easy to emit the NULL checks for pointer arguments when emitting the function call, all it takes is to inject a series of `jump_if_reg_zero` opcodes before each `call_ind` when parsing the opcodes stream, all going to a bit of code that emits a `ValueError` (the last bit of code may even get shared across each function in the same module when doing intra-module optimisations).

After having worked on the native emitters for a short while, I think emitting NULL checks at call sites after marking which parameter on which call needs to be checked is probably the easiest way to get it done.  However, in the end it'd probably require too many modifications and would bloat the compiler for something that, after all, has a marginal benefit if you use code analysis tools that can spot these issues for you anyway.

I guess very few people would bother to explicitly enable this, so it's better left to be done elsewhere :)

Issue #14217, opened 2024-03-30
ESP32-S3 forcing use of USB OTG


kdschlosser:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

ESP32-S3

### MicroPython version

MicroPython: 1.22.2
board: ESP32_GENERIC_S3
board_variant: SPIRAM_OCT

### Reproduction

compile for the ESP32-S3 and try and use the pins for I2C. Using them results in a timeout or if doing a scan it takes the scan more than a minute to complete and returns an empty list.

### Expected behaviour

for I2C to work using pins 19 and 20

### Observed behaviour

I2C attached to pins 19 and 20 does not work I believe this issue is caused by the forced use of USB OTG. I have tried to turn it off by setting `CONFIG_USB_OTG_SUPPORTED=n` in the `ports/esp32/boards/ESP32_GENERIC_S3/sdkconfig.board` file and also removing `boards/sdkconfig.usb` from the `ports/esp32/boards/ESP32_GENERIC_S3/mpconfigboard.cmake` file. For whatever reason `CONFIG_USB_OTG_SUPPORTED` still gets set to `y` when being compiled.

Normally this wouldn't be an issue but because of this code in MicroPython

```c
void mp_task(void *pvParameter) {
    volatile uint32_t sp = (uint32_t)esp_cpu_get_sp();
    #if MICROPY_PY_THREAD
    mp_thread_init(pxTaskGetStackStart(NULL), MICROPY_TASK_STACK_SIZE / sizeof(uintptr_t));
    #endif
    #if CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG
    usb_serial_jtag_init();
    #elif CONFIG_USB_OTG_SUPPORTED
    usb_init();
    #endif
    #if MICROPY_HW_ENABLE_UART_REPL
    uart_stdout_init();
    #endif
    machine_init();
```

[...]

PR #15727, opened 2024-08-27
esp32: Fix ESP32-C3 usb serial/jtag on IDF v5.0.4, clean up native USB-CDC

### LINKED ISSUE #15701: ESP32_GENERIC_C3 binary does not run if compiled with IDF 5.0.4

ESP32C3, GENERIC_ESP32_C3, SEEED XIAO ESP32C3  IDF 5.0.4

### MicroPython version

[v1.24.0-preview.224.g6c3dc0c0b (2024-08-22) .bin](https://micropython.org/resources/firmware/ESP32_GENERIC_C3-20240822-v1.24.0-preview.224.g6c3dc0c0b.bin)

Started with branch https://github.com/micropython/micropython/commit/0a11832cddbeed57bf0267afa0c833584817312d

### Reproduction

(Presumes use of IDF 5.0.4 for working binary downloads:
```
$strings ~/Downloads/ESP32_GENERIC_C3-20240820-v1.24.0-preview.213.g76dd4facb.bin | grep IDF
[0;32mI (%lu) %s: ESP-IDF:          %s
IDF version : v5.0.4
MicroPython-1.24.0-preview-riscv-IDFv5.0.4-with-newlib4.1.0
```
)
1. For ESP32_GENERIC_C3, Download latest preview version [v1.24.0-preview.224.g6c3dc0c0b (2024-08-22) .bin](https://micropython.org/resources/firmware/ESP32_GENERIC_C3-20240822-v1.24.0-preview.224.g6c3dc0c0b.bin)
2. Write to board:
`esptool --chip esp32c3 --port /dev/ttyACM0 --baud 460800 write_flash -z 0x0 ~/Downloads/ESP32_GENERIC_C3-20240820-v1.24.0-preview.213.g76dd4facb.bin`
3.  Reset board
4. `mpremote` 

### Expected behaviour

Expected `>>>` prompt after carriage return

### Observed behaviour

No response 

### Additional Information

The first revision that fails with IDV 5.0.4 is:
https://github.com/micropython/micropython/commit/0a11832cddbeed57bf0267afa0c833584817312d `esp32: Use the ESP-IDF default esp_console config for ESP32-C3 `

If I use IDF 5.2.2 to compile the most recent development versions the  binary loads and runs successfully.  

### Code of Conduct

Yes, I agree

projectgus:
### Summary

* Closes #15701 which is a regression introduced with #15523 when using IDF v5.0.x.
* Fixes a lower impact regression from the same PR where on ESP-IDF >= 5.1 the ESP32-S3 would come up using the USB Serial/JTAG interface instead of native USB. The obvious fixes for the first regression all made this second regression appear on all ESP-IDF versions.
* Overall the logic for how the USB function is assigned were a bit hard to follow, and depended on some default ESP-IDF config selections.
* There are now two new MicroPython config macros, fully independent of ESP-IDF config:
  -  `MICROPY_HW_USB_CDC` for native USB-CDC which defaults to 1 on supported SoCs (currently ESP32-S2,S3).
  - `MICROPY_HW_ESP_USB_SERIAL_JTAG` to use the serial/JTAG peripheral which defaults to 1 on supported SoCs that don't have native USB-CDC enabled.
* Combined with the existing `MICROPY_HW_ENABLE_UART_REPL` this gives full control in the MicroPython board header over which serial options are enabled. This somewhat helps with #14217 although it's not a fix for it.
* Support in the code for setting `CONFIG_ESP_CONSOLE_USB_CDC` on S3 has been removed. This already didn't work on master, and no boards used it. Getting both ESP-IDF stdio and MicroPython stdio on the USB-CDC is actually now possible with this PR (didn't work without it), but requires setting both the sdkconfig option and the MicroPython build option independently.

### Testing

Build and verified REPL works with ESP-IDF v5.0.4 and v5.2.2 on:

* ESP32-C3 DevKitC (external USB/serial)
* Seeed XIAO C3 (internal USB/serial)
* ESP32-S3 DevKitC (both "UART" and "native USB" ports)

We could drop ESP-IDF v5.0.4 support now, but nightly builds have only just switched over so good to wait a little while and see if any other issues crop up. There is still the S3 issue to resolve.

There is a weird inconsistency where ESP-IDF stdout (and logs) go to different places than MicroPython stdout. However that's already the case on master and this PR makes it a little easier to reason about (because the ESP-IDF stdio config and the MicroPython stdio config are now fully independent of each other). If necessary then we could add a hook that routes ESP-IDF stdout into MicroPython stdout as part of app_main(), but I think safe to leave for now.

### Future Work

* Disabling both USB config macros doesn't disable the USB port after MicroPython starts (currently, or after this change). PR to follow for this.


ports/esp32/mpconfigport.h
```diff
@@ -259,6 +259,24 @@ typedef long mp_off_t;
...
+#define MICROPY_HW_USB_CDC (CONFIG_USB_OTG_SUPPORTED)
+#endif
+
+// Enable serial over USB/JTAG serial peripheral
+#ifndef MICROPY_HW_ESP_USB_SERIAL_JTAG
+#define MICROPY_HW_ESP_USB_SERIAL_JTAG (!MICROPY_HW_USB_CDC && \
+    (CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED || (        \
+    ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 1, 0) && ( \
+    CONFIG_ESP_CONSOLE_SECONDARY_USB_SERIAL_JTAG || \
+    CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG))))
```
projectgus, line 273: This is pretty ugly but after we drop 5.0.x support it can be simplified to `(!MICROPY_HW_USB_CDC && CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED)`

projectgus:
Limitation of this PR on ESP32-S3, even if you set:

```
#define MICROPY_HW_USB_CDC (0)
#define MICROPY_HW_ESP_USB_SERIAL_JTAG (0)
```

Then the USB serial/JTAG still enables. This is no worse than the behaviour on master, though, so will submit a follow-up PR for it.

projectgus:
I had some second thoughts about this last night, because it commits MicroPython to being the canonical source for its stdio routing (i.e. whether to connect stdio to UART serial, USB Serial/JTAG, and/or USB OTG peripheral). The ESP-IDF stdio is not really kept consistent with this (in current master and in this PR). Currently on master the ESP-IDF/libc stdout can be routed to UART serial and USB serial/JTAG, but not to the USB-CDC serial device - IIUC that "port" only ever gets MicroPython stdio.

This is *mostly* fine, I think it only really comes up when enabling ESP-IDF logging or if someone links a user C module that calls libc printf() or similar.

However currently in master the ESP-IDF sdkconfig is the canonical source for both stdio configs, and this PR changes it to make MicroPython build macros the canonical source for the MicroPython stdio config. So there's potential for more divergence.

I think the main limitation is that there are currently MicroPython configs that ESP-IDF's stdio can't support with its default Kconfig settings, i.e. output to UART0 and USB-CDC simultaneously. Maybe of more concern, configuring everything in ESP-IDF probably makes it fiddlier to change stdio around at runtime.

So maybe this PR doubling down on the current approach is OK. If we want to sync up ESP-IDF/libc output (so e.g. IDF logs can go to CDC port) then we can do it by installing custom stdout or logging handlers when MicroPython starts up.

ports/esp32/mpconfigport.h
```diff
@@ -259,6 +259,14 @@ typedef long mp_off_t;
 // board specifics
 #define MICROPY_PY_SYS_PLATFORM "esp32"
 
+// Enable serial over USB/JTAG serial peripheral
+#ifndef MICROPY_HW_ESP_USB_SERIAL_JTAG
+#define MICROPY_HW_ESP_USB_SERIAL_JTAG (CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED || ( \
+    ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 1, 0) && (         \
+    CONFIG_ESP_CONSOLE_SECONDARY_USB_SERIAL_JTAG ||         \
```
dpgeorge: I don't think we ever checked/used `CONFIG_ESP_CONSOLE_SECONDARY_USB_SERIAL_JTAG`, so could probably leave that out of this macro definition.  But also doesn't hurt to include it.

dpgeorge: Actually... maybe the inclusion of this SECONDARY option is why S3 changes behaviour?

dpgeorge:
> Applying this fix unexpectedly caused ESP32-S3 to come up using the USB Serial/JTAG interface instead of native USB. The config options here were a bit hard to follow, so I added another commit that refactors this as well.

I don't understand why this happened with S3.  Also, is this problem with S3 only on IDF 5.0.4, or only on IDF 5.2.2, or both?

ports/esp32/mpconfigport.h
projectgus: Yes, this is why S3 changed behaviour. That option is set by default in the ESP-IDF config.

dpgeorge: So, isn't it possible to just exclude `CONFIG_ESP_CONSOLE_SECONDARY_USB_SERIAL_JTAG` from the check, and then it will work on S3?

Said differently: without this PR, S3 works.  And the only thing that needs fixing is a rename of `CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED` to `CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG` for IDF < 5.1.  And the latter can be achieved probably with:
```c
#if CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG
#define CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED 1
#endif
```

ports/esp32/mpconfigport.h
```diff
@@ -259,6 +259,24 @@ typedef long mp_off_t;
 // board specifics
 #define MICROPY_PY_SYS_PLATFORM "esp32"
 
+// Enable native USB peripheral CDC via TinyUSB
+#ifndef MICROPY_HW_USB_CDC
+#define MICROPY_HW_USB_CDC (CONFIG_USB_OTG_SUPPORTED)
+#endif
+
+// Enable serial over USB/JTAG serial peripheral
```
andrewleech, line 267:
```suggestion
// Enable stdio over USB Serial / JTAG  peripheral```

**Comment by andrewleech** [comment]
I for one fully support stdio management in micropython, rather than via IDF.

Tangential to this change, I've been looking at the way all stm32 stdio is currently run via dupterm (if enabled) during my in-progress efforts to migrate stm32 to tinyusb. 

The advantage of this is full runtime / application support for enabling & disabling any combination of peripherals for stdio. 
Also, once disconnected from stdio, the cdc object can be used by application code as a standard stream object for custom protocol use.

This will require a stream class wrapper around the usb cdc code and registering it along with any uarts etc with dupterm, which I'm keen to re-implement in the shared tinyusb codebase.

File: ports/esp32/mpconfigport.h (hunk 2 above)
Comment (projectgus) [reply ports/esp32/mpconfigport.h]: I think that suggested patch is actually equivalent behaviour-wise to the first commit in this PR, although I didn't fully understand that until just now when I went to look into it again:

Serial over S3 native USB port currently works on master in two different ways depending on which ESP-IDF version you have, with <5.1 you get native USB and with >=5.1 you get USB Serial/JTAG peripheral (because both `CONFIG_ESP_CONSOLE_SECONDARY_USB_SERIAL_JTAG` and `CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED` are set by default on both S3 and C3). 
That particular regression (switching to the USB/JTAG peripheral) was introduced by me unwittingly.

With the patch (either the three line one or the first commit in this PR by itself) then S3 would become USB Serial/JTAG regardless of ESP-IDF version.

So I think the three line patch would indeed fix the immediate issue with C3 on v5.0.4 and wouldn't make anything else obviously worse. Maybe that is worth doing now as a quick PR to unbreak C3 on nightly builds.

However, it also seems pretty clear that we should clean this up so it's easier to reason about. Either with the approach in this PR, or with something else.

**Comment by projectgus** [comment]
> I for one fully support stdio management in micropython, rather than via IDF.

@andrewleech @dpgeorge After thinking about this some more I've rewritten this PR to completely decouple the ESP-IDF stdio config and the MicroPython stdio config. They were already not equivalent(*), so I think this actually makes it way easier to reason about.

(*) And could not be made equivalent because if you switch stdout to USB-CDC in ESP-IDF  config then it disables UART stdout.

Have rewritten the PR description as well, and re-ran the tests mentioned there. PTAL.

**Comment by andrewleech** [comment]
> MICROPY_HW_USB_CDC for native USB-CDC which defaults to 1 on supported SoCs (currently ESP32-S3).

This should be S2 as well I believe?

File: ports/esp32/main.c [hunk 4]
```diff
@@ -99,9 +99,9 @@ void mp_task(void *pvParameter) {
     #if MICROPY_PY_THREAD
     mp_thread_init(pxTaskGetStackStart(NULL), MICROPY_TASK_STACK_SIZE / sizeof(uintptr_t));
     #endif
-    #if CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED
+    #if MICROPY_HW_ESP_USB_SERIAL_JTAG
     usb_serial_jtag_init();
-    #elif CONFIG_USB_OTG_SUPPORTED
+    #elif MICROPY_HW_USB_CDC
```
dpgeorge, line 104: I like this, using `MICROPY_HW_xxx` config options exclusively makes it easier to follow.

But, it seems this will clash with #15108, which uses `MICROPY_HW_ENABLE_USBDEV` here instead of `MICROPY_HW_USB_CDC`.  Maybe the former is better because you want USB to be initialised if any USB mode is used (eg just MSC)?

I guess a lot of this PR will clash with #15108, eg the changes to `mphalport.c`.  Don't want to have our efforts here wasted if #15108 will need to change things anyway.

ports/esp32/usb.c
```diff
@@ -28,7 +28,7 @@
 #include "py/mphal.h"
 #include "usb.h"
 
-#if CONFIG_USB_OTG_SUPPORTED && !CONFIG_ESP_CONSOLE_USB_CDC && !CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG_ENABLED
+#if MICROPY_HW_USB_CDC
```
dpgeorge, line 31: I like how much simpler this has become.  But again it will clash with #15108, and maybe should be `MICROPY_HW_ENABLE_USBDEV`?  (That's not what #15108 does, but might be what #15108 needs.)

ports/esp32/mpconfigport.h
```diff
@@ -260,6 +260,20 @@ typedef long mp_off_t;
...
 
+// Enable stdio over native USB peripheral CDC via TinyUSB
+#ifndef MICROPY_HW_USB_CDC
+#define MICROPY_HW_USB_CDC (SOC_USB_OTG_SUPPORTED)
+#endif
+
+// Enable stdio over USB Serial/JTAG peripheral
+#ifndef MICROPY_HW_ESP_USB_SERIAL_JTAG
+#define MICROPY_HW_ESP_USB_SERIAL_JTAG (SOC_USB_SERIAL_JTAG_SUPPORTED && !MICROPY_HW_USB_CDC)
+#endif // MICROPY_HW_ESP_USB_SERIAL_JTAG
```
dpgeorge, line 271: The comment on this line is not needed, because the if-endif is so short, and it's not consistent with the other if-endif blocks in this file.

projectgus:
> This should be S2 as well I believe?

I always forget about the ESP32-S2, you're entirely correct! Have updated the post.

ports/esp32/main.c
projectgus: Yes, agree that eventually we want to split this into `MICROPY_HW_ENABLE_USBDEV` and `MICROPY_HW_USB_CDC` the same as the rp2 port. This did cross my mind when creating this, but I figured at the moment these two macros would be equivalent - so leave splitting them out for when introducing the functionality that needs to differentiate them. I can see the case for adding it now, though - a little less churn.

@andrewleech Is doing this one way or the other way going to be easier for you?

projectgus: Looked at #15108 again and can see the point, so I've pushed commit e1fd26d82 that adds `MICROPY_HW_ENABLE_USBDEV`. It's a bit odd because if you build with `#define MICROPY_HW_ENABLE_USBDEV (1)` and `#define MICROPY_HW_USB_CDC (0)` then the port enumerates as a USB-CDC device (as the descriptor is configured in ESP-IDF) but the port doesn't do anything.

It might be better if we drop that commit from here, and @andrewleech adds it to his branch the next time he rebases #15108. Not sure.

dpgeorge: Let's make the change to use `MICROPY_HW_ENABLE_USBDEV` in #15108 instead.  That keeps this PR simple.

dpgeorge (approved):

dpgeorge:
Unfortunately it looks like this PR broke `ARDUINO_NANO_ESP32`:
```
micropython/shared/tinyusb/mp_usbd_cdc.c: In function 'tud_sof_cb':
micropython/shared/tinyusb/mp_usbd_cdc.c:134:9: error: implicit declaration of function 'tud_sof_cb_enable'; did you mean 'dcd_sof_enable'? [-Werror=implicit-function-declaration]
  134 |         tud_sof_cb_enable(false);
      |         ^~~~~~~~~~~~~~~~~
      |         dcd_sof_enable
```
That board defines `MICROPY_HW_ENABLE_USBDEV`...

The option `!MICROPY_EXCLUDE_SHARED_TINYUSB_USBD_CDC` is no longer used anywhere, maybe it needs to be enabled for this board?

PR #15730, opened 2024-08-27
runtime/exitcodes: Set exit code according to the SystemExit exception.

RealJohnSmith:
### Summary

Third time is a charm, this PR continues previous PRs, originally at https://github.com/micropython/micropython/pull/14419 and later https://github.com/micropython/micropython/pull/15392

### Testing

### DIFF
--- a/shared/runtime/pyexec.c
+++ b/shared/runtime/pyexec.c
@@ -76,6 +76,10 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
     nlr_buf_t nlr;
     nlr.ret_val = NULL;
     if (nlr_push(&nlr) == 0) {
+        #if MICROPY_PYEXEC_ENABLE_VM_ABORT
+        nlr_set_abort(&nlr);
+        #endif
+
         mp_obj_t module_fun;
         #if MICROPY_MODULE_FROZEN_MPY
         if (exec_flags & EXEC_FLAG_SOURCE_IS_RAW_CODE) {
@@ -120,7 +124,7 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
         mp_hal_set_interrupt_char(-1); // disable interrupt
         mp_handle_pending(true); // handle any pending exceptions (and any callbacks)
         nlr_pop();
-        ret = 1;
+        ret = PYEXEC_NORMAL_EXIT;
         if (exec_flags & EXEC_FLAG_PRINT_EOF) {
             mp_hal_stdout_tx_strn("\x04", 1);
         }
@@ -139,15 +143,41 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
             mp_hal_stdout_tx_strn("\x04", 1);
         }
 
-        // check for SystemExit
-        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) {
-            // at the moment, the value of SystemExit is unused
+        #if MICROPY_PYEXEC_ENABLE_VM_ABORT
+        if (nlr.ret_val == NULL) { // abort
+            ret = PYEXEC_ABORT;
+        } else
+        #endif
+        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) { // system exit
+            #if MICROPY_PYEXEC_ENABLE_EXIT_CODE_HANDLING
+            mp_obj_t val = mp_obj_exception_get_value(MP_OBJ_FROM_PTR(nlr.ret_val));
+            if (val != mp_const_none) {
+                if (mp_obj_is_int(val)) {
+                    ret = (int)mp_obj_int_get_truncated(val);
+                } else {
+                    mp_obj_print_helper(MICROPY_ERROR_PRINTER, val, PRINT_STR);
+                    mp_print_str(MICROPY_ERROR_PRINTER, "\n");
+                    ret = PYEXEC_UNHANDLED_EXCEPTION;
+                }
+            } else {
+                ret = PYEXEC_NORMAL_EXIT;
+            }
+            #else
             ret = PYEXEC_FORCED_EXIT;
-        } else {
+            #endif
+        } else { // other exception
             mp_obj_print_exception(&mp_plat_print, MP_OBJ_FROM_PTR(nlr.ret_val));
-            ret = 0;
+            ret = PYEXEC_UNHANDLED_EXCEPTION;
+            #if MICROPY_PYEXEC_ENABLE_EXIT_CODE_HANDLING
+            if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_KeyboardInterrupt))) { // keyboard interrupt
+                ret = PYEXEC_KEYBOARD_INTERRUPT;
+            }
+            #endif
         }
     }
+    #if MICROPY_PYEXEC_ENABLE_VM_ABORT
+    nlr_set_abort(NULL);
+    #endif
 
     #if MICROPY_REPL_INFO
     // display debugging info if wanted
--- a/shared/runtime/pyexec.h
+++ b/shared/runtime/pyexec.h
@@ -37,6 +37,17 @@ extern pyexec_mode_kind_t pyexec_mode_kind;
 
 #define PYEXEC_FORCED_EXIT (0x100)
 
+#if MICROPY_PYEXEC_ENABLE_EXIT_CODE_HANDLING
+#define PYEXEC_NORMAL_EXIT (0)
+#define PYEXEC_UNHANDLED_EXCEPTION (1)
+#define PYEXEC_KEYBOARD_INTERRUPT (128 + 2) // same as SIG INT exit code
+#define PYEXEC_ABORT (128 + 9)              // same as SIG KILL exit code
+#else
+#define PYEXEC_NORMAL_EXIT (1)
+#define PYEXEC_UNHANDLED_EXCEPTION (0)
+#define PYEXEC_ABORT PYEXEC_FORCED_EXIT
+#endif
+
 int pyexec_raw_repl(void);
 int pyexec_friendly_repl(void);
 int pyexec_file(const char *filename);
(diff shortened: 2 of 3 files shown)


RealJohnSmith:
@dpgeorge @dlech Can you please review this code, if the changes made in the previous PR (https://github.com/micropython/micropython/pull/15392) satisfy the comments? If not, we can change it in some way, but please tell me in what way.

dpgeorge:
There are two independent things in this PR:
1. Add support for VM abort to the pyexec code.
2. Do more sophisticated handling of uncaught exceptions in pyexec.

Both of these things seem reasonable.  But for (1) it would be great if this was in a separate commit, to show exactly what's changed for that.  And it also needs to have a separate option to control it, eg `MICROPY_PYEXEC_ENABLE_VM_ABORT`.  That's because some users of this function may want to enable `MICROPY_ENABLE_VM_ABORT` but have the abort go out to a higher level than pyexec code.

shared/runtime/pyexec.c
```diff
@@ -139,15 +147,46 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
...
+        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) { // system exit
+            #if MICROPY_ENABLE_EXIT_CODE_HANDLING
+            mp_obj_t val = mp_obj_exception_get_value(MP_OBJ_FROM_PTR(nlr.ret_val));
+            if (val != mp_const_none) {
+                if (mp_obj_is_small_int(val)) {
+                    ret = MP_OBJ_SMALL_INT_VALUE(val);
+                } else {
+                    mp_obj_print(val, PRINT_STR);
+                    mp_print_str(&mp_plat_print, "\n");
+                    ret = PYEXEC_UNHANDLED_EXCEPTION;
```
dpgeorge, line 170: What's the intention of printing the value here?  This is quite unconventional to return a non-int value in a `SystemExit`, so maybe this bit can either do nothing, or call `mp_obj_print_exception`.

But, see also how the unix port handles this: `ports/unix/main.c:handle_uncaught_exception`.  Maybe it's worth copying that code?

shared/runtime/pyexec.c
```diff
@@ -139,15 +147,46 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
...
+                }
+            } else {
+                ret = 0;
+            }
+            #else
+            ret = PYEXEC_FORCED_EXIT;
+            #endif
+        } else { // other exception
+            mp_obj_print_exception(&mp_plat_print, MP_OBJ_FROM_PTR(nlr.ret_val));
```
dpgeorge, line 179: To eliminate duplication of code for this line in the case of a `KeyboardInterrupt` I suggest:
```c
} else { // other exception
    ret = PYEXEC_UNHANDLED_EXCEPTION;
    #if MICROPY_ENABLE_EXIT_CODE_HANDLING
    if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_KeyboardInterrupt))) {
        ret = PYEXEC_KEYBOARD_INTERRUPT;
    }
    #endif
}
```
And then define `PYEXEC_UNHANDLED_EXCEPTION` as 0 when `MICROPY_ENABLE_EXIT_CODE_HANDLING` is disabled.

shared/runtime/pyexec.c
RealJohnSmith: It is to keep the behavior consistent with real python:

https://python.readthedocs.io/en/latest/library/exceptions.html#SystemExit
> If the value is an integer, it specifies the system exit status (passed to C’s exit() function); if it is None, the exit status is zero; if it has another type (such as a string), the object’s value is printed and the exit status is one.

I know it strictly doesn't have to be and that it is something a little different, but I don't see any reason with not honoring this particular behavior.

Do you want me to change it in some way?

stinos: > If the value is an integer,

Makes sense, but minor remark: I don't think it is 100% guaranteed (because of different object representations) that `mp_obj_is_small_int(val)` expresses 'is an integer value' exactly, i.e. there might be integers whos value is larger than what a small int can represent but are still a valid exit code. So perhaps this should try to extract an integer value without forcing it to be a small int?

RealJohnSmith:
Thank you for your time and input. I changed it according to your comments. 
Originally it used to be in two commits, but some time ago, there was a merge conflict as master changed in the meantime quite some bit, in this handling. I fixed the conflict in one go, as it would have been complicated cherry picking specific lines in which belongs where and how it resolves that conflict. Is it ok to have it as one commit after all?

shared/runtime/pyexec.c
RealJohnSmith: I believe that a valid exit code is only one byte long anyway, 0 to 255 (sometimes -1, which is equivalent to 255). So I don't think this should be an issue. Is there a case where bigger exit codes might be needed? Can you point to some documentation where they are used? What function do you suggest instead of `mp_obj_is_small_int(val)` ?

RealJohnSmith: the int is passed to the C exit function.

According to its docs, https://man7.org/linux/man-pages/man3/exit.3.html, only the 0xFF part (the least significant byte, 0-255) is masked and used.

stinos: > is masked and used.

Those are some version linux-specific docs so in my understanding they only specify one particular use case 'returned to the parent when using wait'. It's easy enough to find conflicting information which shows the full 32bit can in fact be used both on unix and especially on windows where it's not uncommon; first search result: https://stackoverflow.com/questions/179565/exit-codes-bigger-than-255-possible.  

> What function do you suggest instead

Hmm, good question. Also depends on whether it should be an actual integer (`mp_obj_get_int`), or something which can be converted to an integer (`mp_obj_get_int_maybe`). But in any case another small issue is that those use `mp_int_t` not `int`. Probably fine to just cast the result to `int` though for completeness I'd document that somewhere.

RealJohnSmith: `mp_obj_is_small_int` should be fine up to 32bits, right? I understand that this is a "pointer-size int". Is it correct?

re `mp_obj_get_int_maybe` - is there a method/helper, which can check whether it is castable? To run the check before the get and maybe do something else with the object.

stinos: > mp_obj_is_small_int should be fine up to 32bits, right? I understand that this is a "pointer-size int". Is it correct?

It's stored in a machine word, but it has 1 or 2 bits less available for the actual representation, see mpconfig.h. So not necessarily 32 bits.


> re mp_obj_get_int_maybe - is there a method/helper, which can check whether it is castable? To run the check before the get and maybe do something else with the object.

On second thought mp_obj_get_int_maybe should not be used here because it can fall back to calling `MP_UNARY_OP_INT_MAYBE` which is probably not intended, and I doubt anyone is going to use it anyway. So probably:

```
if (mp_obj_is_int(val)) {
    val= (int) mp_obj_int_get_truncated(arg);
}
```

is ok here.

RealJohnSmith: Ok, I applied this and force pushed. Should be ok resolved now.

RealJohnSmith:
Is it ok now, in its current form?

RealJohnSmith:
What blocks it from mergin?

shared/runtime/pyexec.c
```diff
@@ -139,15 +143,41 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
...
+        #endif
+        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t *)nlr.ret_val)->type), MP_OBJ_FROM_PTR(&mp_type_SystemExit))) { // system exit
+            #if MICROPY_PYEXEC_ENABLE_EXIT_CODE_HANDLING
+            mp_obj_t val = mp_obj_exception_get_value(MP_OBJ_FROM_PTR(nlr.ret_val));
+            if (val != mp_const_none) {
+                if (mp_obj_is_int(val)) {
+                    ret = (int)mp_obj_int_get_truncated(val);
+                } else {
+                    mp_obj_print(val, PRINT_STR);
+                    mp_print_str(&mp_plat_print, "\n");
```
dpgeorge, line 159: CPython prints this message to stderr, so if we are aiming to match CPython (which I think is a good idea), this should be:
```c
mp_obj_print_helper(MICROPY_ERROR_PRINTER, val, PRINT_STR);
mp_print_str(MICROPY_ERROR_PRINTER, "\n");
```

shared/runtime/pyexec.c
```diff
@@ -139,15 +143,41 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
...
+                    ret = PYEXEC_UNHANDLED_EXCEPTION;
+                }
+            } else {
+                ret = 0;
```
dpgeorge, line 163: This can probably be `ret = PYEXEC_NORMAL_EXIT`.

shared/runtime/pyexec.h
```diff
@@ -37,6 +37,23 @@ extern pyexec_mode_kind_t pyexec_mode_kind;
 
 #define PYEXEC_FORCED_EXIT (0x100)
 
+#if MICROPY_PYEXEC_ENABLE_EXIT_CODE_HANDLING
+#define PYEXEC_NORMAL_EXIT (0)
+#define PYEXEC_UNHANDLED_EXCEPTION (1)
+#define PYEXEC_KEYBOARD_INTERRUPT (128 + 2) // same as SIG INT exit code
+#define PYEXEC_ABORT (128 + 9)              // same as SIG KILL exit code
+#else
+#ifndef PYEXEC_NORMAL_EXIT
```
dpgeorge, line 46: I don't think you need to provide user configurability of these constants.  It can just be:
```
#define PYEXEC_NORMAL_EXIT (1)
#define PYEXEC_UNHANDLED_EXCEPTION (0)
#define PYEXEC_ABORT PYEXEC_FORCED_EXIT
```

shared/runtime/pyexec.h
```diff
@@ -37,6 +37,23 @@ extern pyexec_mode_kind_t pyexec_mode_kind;
 
 #define PYEXEC_FORCED_EXIT (0x100)
 
+#if MICROPY_PYEXEC_ENABLE_EXIT_CODE_HANDLING
+#define PYEXEC_NORMAL_EXIT (0)
+#define PYEXEC_UNHANDLED_EXCEPTION (1)
```
dpgeorge, line 42: How does one distinguish between an unhandled exception, and a `SystemExit(1)`?

RealJohnSmith: One doesn't. That is ok. It is not distinguishable in other programs as well. Only by additional stderr output

RealJohnSmith:
Ok, I updated the code according to the comments

RealJohnSmith:
Is there anything else I should update, or is this ok to accept and merge? @dpgeorge

Issue #14430, opened 2024-05-05
mpy_ld.py: Unsupported relocation when using cos/sin functions


jonnor:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

esp32/xtensawin and armv7emsp

### MicroPython version

MicroPython 1.22.2 from git

### Reproduction

1. Have a dynamic native module using sinf/cosf functions

Note, you have to include the object files from libm and libgcc manually in the Makefile. This means extracting the .o files from the archives, and then adding them to SRC_O make variable. See this for an example:
https://github.com/emlearn/emlearn-micropython/blob/bd85ae1fade99b34a626c9fcaae0aa4f2ec87803/src/emlfft/Makefile

```
ifeq ($(ARCH), xtensawin)
    LIBM_O := lib_a-sf_cos.o lib_a-sf_sin.o lib_a-ef_rem_pio2.o lib_a-kf_cos.o lib_a-kf_sin.o lib_a-sf_fabs.o lib_a-kf_rem_pio2.o lib_a-sf_scalbn.o lib_a-sf_floor.o
endif
ifeq ($(ARCH), armv7emsp)
    LIBM_O := libm_a-sf_cos.o libm_a-sf_sin.o libm_a-ef_rem_pio2.o libm_a-kf_cos.o libm_a-kf_sin.o libm_a-sf_fabs.o libm_a-kf_rem_pio2.o libm_a-sf_scalbn.o libm_a-sf_floor.o
endif

ifeq ($(ARCH), armv7emsp)
	SOFTFP_O := _arm_addsubdf3.o _arm_muldivdf3.o _arm_muldivdf3.o _arm_truncdfsf2.o
endif
ifeq ($(ARCH), xtensawin)
	SOFTFP_O := _floatsidf.o _muldf3.o _divdf3.o _truncdfsf2.o
endif
```

2. Try to build the module

### Expected behaviour

Linking succeeds without errors, module is functional.

Alternatively: Error and/or documentation provides hints as to the possible reasons why this could happen, and ways of resolving.

### Observed behaviour

Exception when linking. With ARCH=xtensawin

```
[...]

PR #15738, opened 2024-08-28
tools/mpy_ld: Ignore R_XTENSA_ASM_EXPAND relocation entries.

agatti:
### Summary

As reported in #14430 the Xtensa compiler can add R_XTENSA_ASM_EXPAND relocation relaxation entries in object files, and they were not supported by mpy_ld.

This commit adds handling for that entry, doing nothing with it, as it is only of real use for an optimising linker.

### Testing

I didn't directly test this, although it was reported as working by @vshymanskyy in https://github.com/micropython/micropython/issues/14430#issuecomment-2315230596

### DIFF
--- a/tools/mpy_ld.py
+++ b/tools/mpy_ld.py
@@ -69,6 +69,7 @@ R_XTENSA_PLT = 6
 R_386_GOTOFF = 9
 R_386_GOTPC = 10
 R_ARM_THM_CALL = 10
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks, this looks OK to me.

It would be quite hard to add a test for this, so I'm happy for there not to be a test.

dpgeorge (approved):

PR #15743, opened 2024-08-29
Merge qemu-riscv port functionality into qemu-arm port

dpgeorge:
### Summary

Currently both the `qemu-arm` and `qemu-riscv` share a lot of code and functionality.  In particular the recent improvement to running tests on `qemu-arm` in #15624 would need to be duplicated in the `qemu-riscv` port.

This PR merges the `qemu-riscv` port into the `qemu-arm` port.  The only real differences between the two are the toolchains used to build the code, and the initialisation/startup framework.  Everything else is pretty much the same, so this brings the following benefits:
- less code duplication
- less burden on maintenance
- generalised qemu port, could in the future support other architectures

A new board `VIRT_RV32` has been added to the `qemu-arm` port which is the existing RISC-V from the `qemu-riscv` port.  To build it is simply:
```
$ make BOARD=VIRT_RV32 repl
```

To cleanly separate the code for the different architectures, startup code has been moved to `ports/qemu-arm/mcu/<arch>/`.

### Testing

The CI has been updated to run the `VIRT_RV32` board against the test suite.

We could keep the ports separate, but there's a lot of code duplication, and an extra burden on maintenance.

Eventually we would rename the `qemu-arm` port to `qemu`, but I didn't do that in this PR.


dpgeorge:
@agatti I would appreciate your comments/review of this, thanks!

ports/qemu-arm/mpconfigport.h
```diff
@@ -47,7 +49,11 @@
 #define MICROPY_FLOAT_IMPL          (MICROPY_FLOAT_IMPL_FLOAT)
 #define MICROPY_WARNINGS            (1)
 #define MICROPY_PY_IO_IOBASE        (0)
+#if defined(__ARM_ARCH)
 #define MICROPY_PY_SYS_PLATFORM     "qemu-arm"
+#elif defined(__riscv)
+#define MICROPY_PY_SYS_PLATFORM     "qemu-riscv32"
+#endif
```
dpgeorge, line 56: Probably this would be replaced with simply `"qemu"` when we rename the port.

ports/qemu-arm/Makefile
```diff
@@ -51,16 +59,69 @@ endif
...
+
+RV32_ABI = ilp32
+
+QEMU_ARGS += -bios none
+
+# GCC 10 and lower do not recognise the Zicsr extension in the
+# architecture name.  "Make" unfortunately does not provide any simple way
+# to perform numeric comparisons, so to keep things simple we assume that
+# GCC is at least version 10 for the time being.
+ifeq ($(GCC_VERSION),10)
```
agatti, line 82: Maybe you may want to replace this bit with `ifeq ($(shell test $(GCC_VERSION) -ge 10; echo $$?),0)` and remove the second sentence in the original comment.  I found a solution about that issue long after the code was committed...

ports/qemu-arm/Makefile
```diff
@@ -86,6 +147,11 @@ repl: $(BUILD)/firmware.elf
 run: $(BUILD)/firmware.elf
 	$(QEMU_SYSTEM) $(QEMU_ARGS) -serial pty -kernel $<
 
+# `make debug` will block QEMU until a debugger is connected to port 1234.
+.PHONY: debug
+debug: $(BUILD)/firmware.elf
+	$(QEMU_SYSTEM) $(QEMU_ARGS) -serial mon:stdio -S -s -kernel $<
+
 .PHONY: test
 test: $(BUILD)/firmware.elf
```
agatti, line 156: I'd add a `testdebug` target too.  How would you be able to see things like #15589 in a debugger otherwise?

agatti (commented):
Besides a couple of minor Makefile things, that looks good to me!

ports/qemu-arm/Makefile
```diff
@@ -90,65 +59,107 @@ endif
...
+# `make debug` will block QEMU until a debugger is connected to port 1234.
+.PHONY: debug
+debug: $(BUILD)/firmware.elf
+	$(QEMU_SYSTEM) $(QEMU_ARGS) -serial mon:stdio -S -s -kernel $<
 
 .PHONY: test
 test: $(BUILD)/firmware.elf
 	$(eval DIRNAME=ports/$(notdir $(CURDIR)))
-	cd $(TOP)/tests && ./run-tests.py --target qemu-arm --device execpty:"qemu-system-arm $(QEMU_ARGS) -serial pty -kernel ../$(DIRNAME)/$<" $(TESTS_EXCLUDE)
+	cd $(TOP)/tests && ./run-tests.py --target qemu-arm --device execpty:"$(QEMU_SYSTEM) $(QEMU_ARGS) -serial pty -kernel ../$(DIRNAME)/$<" $(TESTS_EXCLUDE)
```
agatti, line 158: I'd add another argument to this for additional user arguments for `run-tests.py`, or rename `TESTS_EXCLUDE` into something else and have it kick in only if it's not defined.

ports/qemu-arm/Makefile
dpgeorge: It's not possible to used both `-S` and the `run-tests.py` infrastructure, because the latter consumes stdio of qemu-system-X and so you won't be able to start the emulator because the monitor is not accessible.

> That issue was reproduced by running several tests in a row -

I think we could make this `debug` target use `-serial pty` and then you can run the tests manually as many times as you like against the exposed serial device, eg:
```
$ make debug
<in a separate terminal>
$ gdb ...
<in a separate terminal>
$ ./run-tests.py --target qemu-arm --device /dev/pts/123
```

ports/qemu-arm/Makefile
dpgeorge: Good idea, done.  But I changed `-ge` to `-le` because I think that's what you meant.

ports/qemu-arm/Makefile
dpgeorge: OK, how about adding `$(RUN_TESTS_EXTRA)`?  If it's not defined it won't do anything.  The user can set it from the command line, eg to `-d basics`.

ports/qemu-arm/Makefile
agatti: I just tried the version I committed a few days ago as https://github.com/agatti/micropython/commit/ce8a877310dba2c09fcf985758a524ed3f7cbb79, using `-serial pty`, and it seems to work when running `make testdebug` in the first terminal and when it's ready spawning another one and run `gdb` attaching to the qemu instance.  Unless something else changed there's no need to use three terminals :)

ports/qemu-arm/Makefile
agatti: That would work, too.  Would it help if these extra arguments (along with the other targets) are documented or are niche enough that if you need them you'd have probably figured them out already?

dpgeorge: I can document the extra args, and the possible options for `BOARD=`.

ports/qemu-arm/Makefile
agatti: > But what I'm also saying above is that it's possible to rerun the test suite over and over, or whatever combination of tests you like. That requires 3 terminals.

Oh, never figured that was a possibility.  Almost every time I've had the need for this it turned out to be a fatal crash so there was no point in rerunning the test without restarting qemu as well.  In that case then yes, you're absolutely right.

ports/qemu-arm/Makefile
```diff
@@ -86,6 +147,11 @@ repl: $(BUILD)/firmware.elf
 run: $(BUILD)/firmware.elf
 	$(QEMU_SYSTEM) $(QEMU_ARGS) -serial pty -kernel $<
 
+# `make debug` will block QEMU until a debugger is connected to port 1234.
+.PHONY: debug
+debug: $(BUILD)/firmware.elf
+	$(QEMU_SYSTEM) $(QEMU_ARGS) -serial mon:stdio -S -s -kernel $<
```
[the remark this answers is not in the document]
dpgeorge: I added a `$(QEMU_DEBUG_ARGS)` option, defaulting to `-s`.

ports/qemu-arm/Makefile
dpgeorge: You are right, you can use `-S -s` with the `make test` target.

Instead of duplicating all targets with a debug version, I added an orthogonal option `QEMU_DEBUG` which can be enabled for any qemu target, eg:
```
$ make QEMU_DEBUG=1 test
```
That will run the test suite with `-S -s`.

I also added `QEMU_DEBUG_EXTRA`, and updated the README with descriptions of these options.

ports/qemu-arm/Makefile
dpgeorge: I added `RUN_TEST_EXTRA`, and documented it in the README, along with a list of available boards.

dpgeorge:
@agatti this has been updated based on your feedback, if you'd like to take another look, thanks!

tools/ci.sh
```diff
@@ -259,26 +261,8 @@ function ci_qemu_arm_build {
...
-
-function ci_qemu_riscv_build {
-    make ${MAKEOPTS} -C mpy-cross
-    make ${MAKEOPTS} -C ports/qemu-riscv submodules
-    make ${MAKEOPTS} -C ports/qemu-riscv
-    make ${MAKEOPTS} -C ports/qemu-riscv clean
-    make ${MAKEOPTS} -C ports/qemu-riscv -f Makefile.test submodules
-    make ${MAKEOPTS} -C ports/qemu-riscv -f Makefile.test test
+    make ${MAKEOPTS} -C ports/qemu-arm BOARD=SABRELITE test
+    make ${MAKEOPTS} -C ports/qemu-arm BOARD=VIRT_RV32 test
```
projectgus, line 265: Is it worth keeping these split into two jobs? The advantages I can see are runtime (can be concurrent) and extra context when debugging (can see immediately if something fails on ARM and/or RISC-V, independently of each other).

projectgus (approved):
These changes LGTM! Have one question about CI, but that's all.

I support the idea of renaming the port to `qemu` as a follow-up.

tools/ci.sh
dpgeorge: Yeah, that would be a good idea, to keep them separate.

The main motivation to combine them is to only need to run one setup step.  At the moment that takes about 45 seconds, but I guess if we split it then the ARM one only needs to install the ARM toolchain, and RV32 one only needs to install the RISC-V toolchain.  Then the only thing in common is installation of qemu, which is probably very fast.

dpgeorge: Now done, two separate CI jobs.

agatti:
@dpgeorge the new changes look good to me as well!

Just one minor thing: given that picolibc for Arm is also an option on the CI machine, would it make sense to extract the libc selection bit that's currently in place for RV32 and make it available for all ports?

I haven't tried to build qemu-arm with picolibc but it'd be an interesting experiment to try (@projectgus expressed some interest back then if memory serves).  All the known libm incompatibilities should be already sorted out, or it wouldn't build the RV32 target at all, would it.

agatti (approved):

dpgeorge:
> Just one minor thing: given that picolibc for Arm is also an option on the CI machine, would it make sense to extract the libc selection bit that's currently in place for RV32 and make it available for all ports?

I have now applied this suggestion and moved the picolibc bit of the Makefile so that it's accessible by both ARM and RISC-V targets.  As part of this I needed to make ARM use CC for linking instead of LD, to match RISC-V.  That anyway makes things more consistent.

dpgeorge:
And of course the CI fails now with picolibc...

agatti:
Interestingly, Arm targets with picolibc are ever so slightly larger than their newlib counterparts:

```
Picolibc

181640      16  103124  284780   4586c build-MPS2_AN385/firmware.elf
264044      16  103124  367184   59a50 build-SABRELITE/firmware.elf

Newlib

181632      16  103124  284772   45864 build-MPS2_AN385/firmware.elf
264036      16  103124  367176   59a48 build-SABRELITE/firmware.elf
```

Not sure if that's due to some missing flags/ifdefs/whatever, but that's unexpected to say the least.

agatti:
@dpgeorge I was also wondering, given that there's no RV32 platform in the code size report then maybe adding the size of the qemu repl build there could be nice to have.  The qemu build isn't influenced by too much third party code so it would reflect core code increase more accurately I think.

If that's not in the scope for this PR, then if I can add that functionality myself I'll send one after this is merged (and the port gets renamed from qemu-arm to qemu).

projectgus:
> Interestingly, Arm targets with picolibc are ever so slightly larger than their newlib counterparts:

Huh, that's indeed interesting! I suppose MicroPython is pretty light-on with the libc features it uses, especially in these ports, so there may not be a lot of scope for size optimisation.

dpgeorge:
> I was also wondering, given that there's no RV32 platform in the code size report then maybe adding the size of the qemu repl build there could be nice to have. The qemu build isn't influenced by too much third party code so it would reflect core code increase more accurately I think.

That's a good idea.  But we need to be aware of the time that the code-size CI job takes to run, it's currently one of the slowest jobs, and adding more build to it will make it even slower.

Feel free to open a PR for it.

dpgeorge:
See #15784 for the rename of the qemu-arm port to qemu.

agatti:
@dpgeorge you may want to increase the heap size for RV32, I just bumped into https://github.com/micropython/micropython/pull/12853#issuecomment-1798784848 again.

dpgeorge:
> you may want to increase the heap size for RV32, I just bumped into [#12853 (comment)](https://github.com/micropython/micropython/pull/12853#issuecomment-1798784848) again.

The tests are no longer embedded in the firmware, but rather run individually over the serial port.  Can you please explain what the issue is that you see?  Something I can reproduce?

agatti:
Sure, take a look at the test results of https://github.com/micropython/micropython/pull/15714: https://github.com/micropython/micropython/actions/runs/10700895755/job/29665631825?pr=15714 - I rebased the code to be compatible with the merged qemu port.

```
740 tests performed (22638 individual testcases)
739 tests passed
134 tests skipped: ...
1 tests failed: deflate_decompress
make: Leaving directory '/home/runner/work/micropython/micropython/ports/qemu-arm'
```

dpgeorge:
Hmm, that's strange the failure comes up there, with seemingly unrelated changes.  All that PR does is enable inline-riscv-asm, and that shouldn't change the amount of GC heap available, or how the GC heap is used by that failing test.

I guess you can just increase the qemu-arm GC heap size by a little bit in your PR, to get that test to pass.  If that doesn't do it, it needs investigating.

agatti:
I can also reproduce the issue locally on an Ubuntu 22.04 VM when running tests through `source tools/ci.sh && ci_qemu_build_rv32`.  Increasing `HEAP_SIZE` from 100k to 120k makes the test pass (it has to be between 117k and 120k, with 116k it would still fail).

If a GC sweep is performed after each test, I wonder what's lingering in the GC heap when that test is scheduled to run.  I'm not sure decompressing the zlib data from that failing test should take that much memory to begin with.

If needed I can easily attach a debugger, reproduce the issue, and report back on what's happening - but I'd need a couple of pointers on what to look for and where to look.

PR #15747, opened 2024-08-29
extmod/libmetal: Fix libmetal rule.

iabdalkader:
Dependency on auto-generated libmetal should be an order only prerequisite.

### DIFF
--- a/extmod/libmetal/libmetal.mk
+++ b/extmod/libmetal/libmetal.mk
@@ -4,13 +4,13 @@
 # This replicates the basic functionality of the pre-processor, including adding a "micropython"
 # platform that is almost identical to the built-in "generic" platform with a few small changes
 # provided by the files in extmod/libmetal.
-$(BUILD)/openamp: $(BUILD)
+$(BUILD)/openamp: | $(BUILD)
 	$(MKDIR) -p $@
 
-$(BUILD)/openamp/metal: $(BUILD)/openamp
+$(BUILD)/openamp/metal: | $(BUILD)/openamp
 	$(MKDIR) -p $@
 
-$(BUILD)/openamp/metal/config.h: $(BUILD)/openamp/metal $(TOP)/$(LIBMETAL_DIR)/lib/config.h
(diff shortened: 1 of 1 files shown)


extmod/libmetal/libmetal.mk
```diff
@@ -10,7 +10,7 @@ $(BUILD)/openamp: $(BUILD)
 $(BUILD)/openamp/metal: $(BUILD)/openamp
 	$(MKDIR) -p $@
 
-$(BUILD)/openamp/metal/config.h: $(BUILD)/openamp/metal $(TOP)/$(LIBMETAL_DIR)/lib/config.h
+$(BUILD)/openamp/metal/config.h: | $(BUILD)/openamp/metal $(TOP)/$(LIBMETAL_DIR)/lib/config.h
```
dpgeorge, line 13: Wouldn't you want to rebuild this target if `lib/libmetal` changes, ie keep the second part a normal prerequisite?  Eg if you check out a different commit from `lib/libmetal` you'd want this rule to run again.

But probably the two mkdir rules above can be changed to order-only prerequisites.

iabdalkader: > ie keep the second part a normal prerequisite

I thought this is what this change is doing, i.e., make the first order-only prerequisite and the second normal prerequisite, no ?

iabdalkader: I can change it to the following:
```
$(BUILD)/openamp/metal/config.h: $(TOP)/$(LIBMETAL_DIR)/lib/config.h | $(BUILD)/openamp/metal
```

iabdalkader: Fixed.

dpgeorge: Yes, that looks better now.

Do you also want to make the rules above this line for `$(BUILD)/openamp` and `$(BUILD)/openamp/metal` also order-only?

iabdalkader: Done. I've tested the build and it's working fine. Also changing `$(TOP)/$(LIBMETAL_DIR)/lib/config.h` triggers a rebuild.

dpgeorge (approved):

PR #12241, opened 2023-08-15
mpy_ld.py: Support modules larger than 4KiB on armv6m

jonnor:
Hit a problem with modules larger than 4 KiB when working on dynamic native modules for RP2040. This should fix that for ARM Thumb targets, using the similar strategy used for ARM Thumb 2 targets.

Have tested it on RP2040, and it seems to work: My module gets initialized, and the defined functions can be called. I have never written Thumb assembler before, so please do sanity check it!

PR #15749, opened 2024-08-30
tools/mpy_ld.py: Support long entry jumps for armv6m.

agatti:
### Summary

This commit adds support for armv6m native modules larger than 4k in code size, by adding a new jump trampoline that does not depend on a single jump instruction address range.  This was reported in https://github.com/micropython/micropython/issues/14430#issuecomment-2316162839 as being a blocker for a reproducible way to trigger the Arm side of the unhandled relocation errors reported in #14430.

### Testing

This was tested on a RPi Pico by @vshymanskyy in https://github.com/micropython/micropython/issues/14430#issuecomment-2319485402

Due to a current mpy_ld.py limitation, entry jumps must not change the length of the emitted code sequence between invocations.  This means the current short jump had to be padded to match the long jump in length (the padding is the defined `undefined instruction` opcode according to the armv6m technical reference manual).  From a cursory look at mpy_ld.py's code I couldn't find a quick and easy way to work around that limitation.

Unfortunately this adds 18 bytes to the jump trampoline on armv6m even when the code in question is well within the initial `b` opcode jump range (maybe this can be shorter but I've been focusing on RISC-V for the past 18 months or so, bear with me here :)).  The trampoline could be four bytes smaller if `R6` and `R7` are safe to be trashed when entering `mpy_init` but if I remember correctly, according to the Arm EABI calling convention it's up to the callee to save and restore R4..R7 (in this case the jump sequence).  If I'm mistaken about that, then this can be easily shortened by removing R6/R7 stack operations and fixing up the `ldr` and `str` offsets.

### DIFF
--- a/tools/mpy_ld.py
+++ b/tools/mpy_ld.py
@@ -88,10 +88,22 @@ def asm_jump_x86(entry):
 
 
 def asm_jump_thumb(entry):
-    # Only signed values that fit in 12 bits are supported
     b_off = entry - 4
-    assert b_off >> 11 == 0 or b_off >> 11 == -1, b_off
-    return struct.pack("<H", 0xE000 | (b_off >> 1 & 0x07FF))
+    if b_off >> 11 == 0 or b_off >> 11 == -1:
+        return struct.pack("<H9H", 0xE000 | (b_off >> 1 & 0x07FF), *([0xDE00] * 9))
+    else:
+        # sub   sp, #4
+        # push  {r6, r7}
+        # ldr   r7, [pc, #8]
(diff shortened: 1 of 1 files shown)


dpgeorge:
Please see #12241, which attempts the same thing.  And see also my comment https://github.com/micropython/micropython/pull/12241#issuecomment-2227540493 for a possible way to do it with less bytes.

agatti:
I see - well if this can be made shorter then by all means let's go with the shorter version!  I've had to deal with RV32 assembly for too much I guess...

Thanks!

PR #15752, opened 2024-08-31
py/gc: Return max_free size in bytes.

iabdalkader:
I think this was an oversight, but regardless of that, for consistency, max_free should be in bytes, not block count. This matches the info's total, used, and free sizes. Also, while it's currently possible to use `MICROPY_BYTES_PER_GC_BLOCK` in outside code to get the maximum free size in bytes, in the future, different areas could have different block sizes, which would make `max_free` useless.

### DIFF
--- a/py/gc.c
+++ b/py/gc.c
@@ -715,6 +715,7 @@ void gc_info(gc_info_t *info) {
 
     info->used *= BYTES_PER_BLOCK;
     info->free *= BYTES_PER_BLOCK;
+    info->max_free *= BYTES_PER_BLOCK;
 
     #if MICROPY_GC_SPLIT_HEAP_AUTO
     info->max_new_split = gc_get_max_new_split();


dpgeorge:
I think the existing code is correct.  It was added in 749cbaca7f5406e23bbf0009477c73e14fa8a9ff where it explicitly stated that the value is number of blocks, not bytes.

The fact it measures number of blocks is consistent with the following output:
```
>>> micropython.mem_info()
mem: total=9060, current=2845, peak=4314
stack: 896 out of 80000
GC: total: 2072832, used: 3968, free: 2068864
 No. of 1-blocks: 66, 2-blocks: 12, max blk sz: 9, max free sz: 64640
```
On the last line there, all numbers measure blocks (not bytes).

iabdalkader:
> On the last line there, all numbers measure blocks (not bytes).

Yes that makes sense, but back then GC only had one area. I could use `MICROPY_BYTES_PER_GC_BLOCK` for now, with multiple areas, the maximum run of blocks could come from any of those areas.
Perhaps we should add a new field just in case someone is parsing/using this info ?

dpgeorge:
At the moment `MICROPY_BYTES_PER_GC_BLOCK` is a fixed constant.  So I don't think we should change anything until that becomes a variable per area.

iabdalkader:
> At the moment `MICROPY_BYTES_PER_GC_BLOCK` is a fixed constant. So I don't think we should change anything until that becomes a variable per area.

If/when that happens, it will break outside code, but I can use `MICROPY_BYTES_PER_GC_BLOCK` for now.

PR #9871, opened 2022-11-07
ussl: Fix polling of SSL sockets

damz:
### What does this PR solve?

There are three issues in the current implementation `ioctl(MP_STREAM_POLL)` for SSL sockets:

* mbedtls might have buffered data (which we can check with `mbedtls_ssl_check_pending`) which needs to be considered when checking for `MP_STREAM_POLL_RD`;
* During handshake (and renegotiation) the state machine might need to write in order to read (or read in order to write);
* An error raised during `read()` or `write()` needs to be stored so that we don't block on polling (e.g. we should not wait to read data on an SSL socket that is in error).

### How does this PR solve the problem?

* Store the direction the library told us it needs, and the last error when raised;
* Only consider the direction the library to us it needed during `ioctl(MP_STREAM_POLL)`;
* Implement a full unit test in unix/coverage

Issue #5270, opened 2019-10-28
Implement DTSL (encrypted UDP sockets)


beyonlo:
Hello.

Are there any example about to use DTLS? I would like to use UDP protocol to communicate, but in the secure mode, like as a TCP socket using TLS. I found on docs just about ussl for TCP sockets.

Thank you.

MrSurly:
I don't believe it's implemented.  A while back, I [took a look](https://github.com/micropython/micropython-esp32/issues/188) at doing so, and despite what I said in that thread, I did indeed abandon it.

IIRC, the underlying stack (mbedtls) supports this.

beyonlo:
@dpgeorge Any plans to implement DTLS?

Thank you.

dpgeorge:
>  Any plans to implement DTLS?

Not in the short term, but in the medium term (~3 months) it could be done if there's enough interest.

dpgeorge:
For reference, using DTLS on CPython is possible via https://github.com/rbit/pydtls

dmartauz:
I am also interested in DTLS support, however I would prefer it to be done similarly to ucryptolib.aes with encrypt/decrypt methods. The reason is that the output would be used as UDP payload for NB-IoT modem controlled via AT commands which is not bound to lwip.

dpgeorge:
> The reason is that the output would be used as UDP payload for NB-IoT modem controlled via AT commands which is not bound to lwip.

Note that the `ssl` module can currently be used stand-alone, ie without lwip/sockets.  It just requires a suitable stream (eg user class derived from `io.IOBase`) to be passed in to `ssl.wrap_socket`  I would envisage a similar thing for DTLS.

Sebbb:
> > Any plans to implement DTLS?
> 
> Not in the short term, but in the medium term (~3 months) it could be done if there's enough interest.

I'm interested as well, I'm searching for an option to encrypt/decrypt UDP packets on an ESP32.

The possibility to communicate securely definitely makes sense, also in the IoT/embedded world :)

PR #15764, opened 2024-09-02
extmod/modussl_mbedtls: Updated Wire in support for DTLS.

keenanjohnson:
This is an update to a previous PR: #10062, which updates the PR based on the latest refactoring of the micropython repo.

I opened this PR as I wasn't sure if the original author of #10062 was still interested in pushing the code change through, but I am quite interested in seeing it happen.

## What is this PR trying to solve?
This PR enables support for DTLS (i.e. TLS over datagram transport protocols like UDP). While support for DTLS is absent in cpython, it is worth supporting it in micropython because it is the basis of the ubiquitous [CoAP protocol](https://www.rfc-editor.org/rfc/rfc7252), used in many IOT projects. See https://github.com/micropython/micropython/issues/5270

## How does this PR solve the problem?
A new set of "protocols" are added to SSLContext:

ssl.PROTOCOL_DTLS_CLIENT
ssl.PROTOCOL_DTLS_SERVER
They act as you expect. If one of these is set, the library assumes that the underlying socket is a datagram-like socket (i.e. UDP or similar).

Implement our own timer callbacks as the out of the box implementation relies on gettimeofday().

TODO: To fully support asyncio for DTLS socket, we will need to return a readable or writable event in poll(MP_STREAM_POLL, ...) if _mbedtls_timing_get_delay(self) >= 1. This is left for future work so as not to interfere with https://github.com/micropython/micropython/pull/9871.


tests/ports/esp32/ssl_dtls.py
```diff
@@ -0,0 +1,11 @@
+# Very basic test of ssl.SSLContext DTLS.
+
+try:
+    import ssl
```
projectgus, line 4: Minor, but so this test doesn't fail on all ports without DTLS support then need to also check for a DTLS name here (i.e. `ssl.PROTOCOL_DTLS_CLIENT` and catch `NameError` below to skip.

projectgus (commented):
Thanks for keeping this PR alive, @keenanjohnson! I can see it being quite useful for anyone using CoAP, we could even consider enabling it by default on some of the bigger ports (such as ESP32) provided the code size impact is not too big ( :crossed_fingers: ).

The main concern I have here is adding all these extra elements to the `ssl` module and its types. This is a CPython specific module, and we're trying to cut down on incompatibilities in these.

(If nothing else, there's also the awkward possibility that some day CPython adds DTLS support and then we have to figure out how to make this API compatible with whatever CPython adopts.)

There is potentially a way forward here. As of b802f0f8abf, there is a `tls` module (non-CPython compatible) and an `ssl` module (CPython compatible). The [current ssl module](https://github.com/micropython/micropython-lib/blob/master/python-stdlib/ssl/ssl.py) just does `from tls import *` but it doesn't have to do that, it could be changed to only import the CPython-compatible names.

To take the code in this direction probably means significant additional work, though. I think applying the tls/ssl split here would look something like this:

* Make a separate `tls.DTLSSocket` (or similar) class, so a CPython-compatible `ssl.socket` doesn't gain any of the other functions. As a bonus, `tls.DTLSSocket` can also be minimal and only expose the functions used by DTLS.
* Can probably leave ` ssl_context_make_new` implementation the same, apart from having it return a `tls.DTLSSocket` object if a DTLS protocol is requested. The behaviour will stay CPython-compatible unless the caller passes in an SSLContext with the DTLS-only type set.
* Update `ssl.py` in micropython-lib so it only imports the CPython-compatible names rather than everything.
* Add a doc for `tls` and describe how it's the non-CPython-compatible version of the `ssl` module. Docs for DTLS can be placed there.

How does that sound? Does it sound reasonable? I admit that it does seem like it could become confusing (i.e. two very similar modules with slightly different behaviour), but it's an awkward situation either way.

We can probably help with some of the steps as well (in particular the last two points), if they become sticking points.

projectgus:
>  Make a separate tls.DTLSSocket (or similar) class, so a CPython-compatible ssl.socket doesn't gain any of the other functions. As a bonus, tls.DTLSSocket can also be minimal and only expose the functions used by DTLS.

I mentioned this to @dpgeorge and he's not keen on a new class for size reasons, instead he suggested not adding the new methods and using `read()` and `write()` methods (possibly plus `read1`/`write1`) for DTLS so that the size stays even smaller. So the main compatibility change may just be to move the new protocol enum values from `ssl` to `tls`.

keenanjohnson:
Hey @projectgus yes that all makes sense. I see that #15905 has been merged as well. Other than fixing the test above, is there any other outstanding changes I can make here before we bring this in?

keenanjohnson:
Tests are now fixed. The outstanding items are improving the test coverage and fixing the commit messages.

keenanjohnson:
@projectgus it seems like no matter what I do I can't appease the code coverage metric.

projectgus:
> It seems like the main issue is that the coverage test is skipping my dtls test: https://github.com/micropython/micropython/actions/runs/12241647550/job/34147207388#step:5:880

It's a bit fiddly, what is happening is that the test runner is running the test in CPython, where it prints "SKIP". Then it's running it in MicroPython, where it passes, and then it's comparing the output and saying "the test has failed"!

The documentation for this has just improved, see the explanation of test types here: https://github.com/micropython/micropython/tree/master/tests#micropython-test-suite

You can run this test locally by:

1) Build the micropython unix port. You can build the exact "coverage" variant by passing `VARIANT=coverage` is you need, but the test also fails on the standard variant which is easier to run as its the default that the test runner uses.
2) Go to the tests directory and run ` ./run-tests.py -i ports/unix/ssl_dtls`

What I suggest doing is having the test print something on success, and then adding an .exp file with that expected output. For example:

```diff
diff --git c/tests/ports/unix/ssl_dtls.py i/tests/ports/unix/ssl_dtls.py
index ec45eb877a..a40736c785 100644
--- c/tests/ports/unix/ssl_dtls.py
+++ i/tests/ports/unix/ssl_dtls.py
@@ -14,3 +14,4 @@ except NameError:
 # Test constructing with valid arguments
 dtls_client = ssl.SSLContext(PROTOCOL_DTLS_CLIENT)
 dtls_server = ssl.SSLContext(PROTOCOL_DTLS_SERVER)
+print("OK")
diff --git c/tests/ports/unix/ssl_dtls.py.exp i/tests/ports/unix/ssl_dtls.py.exp
new file mode 100644
index 0000000000..d86bac9de5
--- /dev/null
+++ i/tests/ports/unix/ssl_dtls.py.exp
@@ -0,0 +1 @@
+OK
```

projectgus:
Once you have the CI passing, the other step to do is to rebase the branch and squash all these commits down to one. Jimmo has a MicroPython-specific guide here: https://github.com/jimmo/git-and-micropython#README - although any git-related guide about "rebasing" and "squashing" will be able to give you the gist.

If you get totally stuck, one of us can help you. Please don't solve the problem by closing this PR and opening a new one, this shouldn't be necessary! Cheers.

tests/ports/esp32/ssl_dtls.py
```diff
@@ -0,0 +1,17 @@
+# Very basic test of ssl.SSLContext DTLS.
```
dpgeorge, line 1: There only needs to be one common test for all ports.  Please move this test to `tests/extmod/tls_dtls.py`, and then remove the unix specific one.

tests/ports/esp32/ssl_dtls.py
```diff
@@ -0,0 +1,17 @@
...
+
+try:
+    import ssl
```
dpgeorge, line 4: I suggest to use `import tls` because this test will not run on CPython, the `ssl` module doesn't have DTLS support.

tests/ports/esp32/ssl_dtls.py
```diff
@@ -0,0 +1,17 @@
...
+    import ssl.PROTOCOL_DTLS_CLIENT
+    import ssl.PROTOCOL_DTLS_SERVER
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+except NameError:
```
dpgeorge, line 10: Instead of separate except clauses, use `except (ImportError, NameError)` (actually `NameError` might need to be `AttributeError`).

extmod/modtls_mbedtls.c
```diff
@@ -520,6 +542,41 @@ static int _mbedtls_ssl_recv(void *ctx, byte *buf, size_t len) {
...
+static int _mbedtls_timing_get_delay(void *ctx) {
+    mp_obj_ssl_socket_t *o = (mp_obj_ssl_socket_t *)ctx;
+
+    if (o->timer_fin_ms == 0) {
+        return -1;
+    }
+
+    mp_uint_t now = mp_hal_ticks_ms() & (MICROPY_PY_TIME_TICKS_PERIOD - 1);
+    mp_int_t elapsed_ms = ((now - o->timer_start_ms + MICROPY_PY_TIME_TICKS_PERIOD / 2) & (MICROPY_PY_TIME_TICKS_PERIOD - 1))
+        - MICROPY_PY_TIME_TICKS_PERIOD / 2;
```
dpgeorge, line 566: Does it need to use truncated wrapping logic like this?  Or can it just use the full (wrapping) range of `mp_hal_ticks_ms()`?  Ie you can just do `mp_hal_ticks_ms() - o->timer_start_ms`.

keenanjohnson:
OK

> Once you have the CI passing, the other step to do is to rebase the branch and squash all these commits down to one. Jimmo has a MicroPython-specific guide here: https://github.com/jimmo/git-and-micropython#README - although any git-related guide about "rebasing" and "squashing" will be able to give you the gist.
> 
> If you get totally stuck, one of us can help you. Please don't solve the problem by closing this PR and opening a new one, this shouldn't be necessary! Cheers.

Thanks @projectgus ! I'm comfortable with rebasing, so I'll do that once all the builds are working :)

tests/ports/esp32/ssl_dtls.py
keenanjohnson: Done!

tests/ports/esp32/ssl_dtls.py
keenanjohnson: Done!

tests/ports/esp32/ssl_dtls.py
keenanjohnson: Done!

tests/ports/esp32/ssl_dtls.py
keenanjohnson: Actually I don't believe that the ssl or tls module in cpython supports DTLS out of the box in this same way, so I'm a bit confused about how to proceed here for the comparison test.

extmod/modtls_mbedtls.c
keenanjohnson: I think the truncated logic should work.

keenanjohnson:
Ok I have reduced the duplicated test, added the basic .exp file, and moved the enable flag to the correct file as suggested by @projectgus in discord, but it still seems to be skipping the test whether I use the tls or ssl module in the test. Am I missing something else obvious?

tests/extmod/tls_dtls.exp
```diff
@@ -0,0 +1 @@
+OK
```
dpgeorge, line 1: This file needs to be called `tls_dtls.py.exp`.  Also please add a newline on the last line (but not a trailing blank line!).

tests/extmod/tls_dtls.py
```diff
@@ -0,0 +1,13 @@
+# Very basic test of DTLS.
+
+try:
+    from tls import PROTOCOL_DTLS_CLIENT, PROTOCOL_DTLS_SERVER, SSLContext
+except (ImportError, NameError):
```
dpgeorge, line 5: Only `ImportError` is needed here when using `from ... import ...`.

keenanjohnson:
I think the main implementation here is all done, but just trying to get the code coverage metric up. Just one more tricky timing line to hit before the code coverage bot is satisfied. Am I doing the testing in the preferred way for micropython @projectgus ?

extmod/modtls_mbedtls.c
```diff
@@ -520,6 +542,39 @@ static int _mbedtls_ssl_recv(void *ctx, byte *buf, size_t len) {
...
 
+#ifdef MBEDTLS_SSL_PROTO_DTLS
+static void _mbedtls_timing_set_delay(void *ctx, uint32_t int_ms, uint32_t fin_ms) {
+    mp_obj_ssl_socket_t *o = (mp_obj_ssl_socket_t *)ctx;
+
+    o->timer_int_ms = int_ms;
+    o->timer_fin_ms = fin_ms;
+
+    if (fin_ms != 0) {
+        o->timer_start_ms = mp_hal_ticks_ms() & (MICROPY_PY_TIME_TICKS_PERIOD - 1);
```
dpgeorge, line 553: I don't think this needs the mask, it should be possible to just do `o->timer_start_ms = mp_hal_ticks_ms();`.

(Because in the function below that computes `elapsed_ms` there is no masking either.)

tests/extmod/tls_dtls.py
```diff
@@ -0,0 +1,55 @@
...
+    import socket, io
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+class DummySocket(io.IOBase):
+    def __init__(self):
+        self.write_buffer = bytearray()
+        self.read_buffer = bytearray()
+        self.elapsed_ms = 0  # Add elapsed time tracking
```
dpgeorge, line 14: It looks like this variable is not used or needed?

dpgeorge:
> just trying to get the code coverage metric up. Just one more tricky timing line to hit before the code coverage bot is satisfied.

I don't think it's possible to hit this last line, it looks highly dependent on internals of mbedTLS.  So I think we can just leave that line untested.

extmod/modtls_mbedtls.c
```diff
@@ -520,6 +542,39 @@ static int _mbedtls_ssl_recv(void *ctx, byte *buf, size_t len) {
...
+}
+
+static int _mbedtls_timing_get_delay(void *ctx) {
+    mp_obj_ssl_socket_t *o = (mp_obj_ssl_socket_t *)ctx;
+
+    if (o->timer_fin_ms == 0) {
+        return -1;
+    }
+
+    mp_int_t elapsed_ms = mp_hal_ticks_ms() - o->timer_start_ms;
```
projectgus, line 564: Nitpick, but suggest `mp_uint_t` to avoid potential signed integer overflow.

The timer_fin_ms and timer_int_ms fields should probably be unsigned as well, to match.

keenanjohnson: Sounds good. Will adjust

extmod/modtls_mbedtls.c
keenanjohnson: Will adjust.

keenanjohnson:
Just getting back around to this after, the holidays. 

Thanks for the feedback @dpgeorge and @projectgus. I believe I have addressed all of it and I've just squashed all my commits down to one that conforms to the guidelines (at least I think so).

With that I think everything is ready to go? As noted above, it's likely not possible to hit the one line that the code coverage bot is complaining about in test.

Any other lingering thoughts or anything before it's ready for merge that I can help tackle? Also I don't feel any particular ownership over anything here, so if there is something basic, feel free to fix it or whatever you need to do in order to merge this in, though I'm also happy to help if you let me know.

keenanjohnson:
Hmmm though now it does seem like the qemu tests are failing for some reason...

dpgeorge:
> Hmmm though now it does seem like the qemu tests are failing for some reason

It looks like GitHub Actions have updated ubuntu-latest to ubuntu-24.04, and that's causing the failures, which are unrelated to this PR's changes.

I've tried to fix those CI issues in #16564.

keenanjohnson:
Ah I see @dpgeorge . Thanks for helping me understand what was going on. Once the other CI issues PR is merged, I can update mine here to sit on top of that change.

dpgeorge:
> Once the other CI issues PR is merged, I can update mine here to sit on top of that change.

The CI should now be fixed, so please rebase and force push.

keenanjohnson:
Rebased onto master. Thanks @dpgeorge !

keenanjohnson:
Hey @projectgus and @dpgeorge just checking that there aren't any other outstanding tasks or asks from your end at this time? No rush as I know you are busy and I appreciate you, just wanted to make sure you weren't waiting on me for something.

dpgeorge:
I have a few comments:
1. The docs need fixing in the sense that they are still documenting the `ssl` module, whereas this new functionality is in the MicroPython-specific `tls` module.  Fixing that is out of scope of this PR, but it still needs to be done.
2. If this is useful functionality then I think it should be enabled on all ports that use mbedtls, eg rp2, stm32, mimxrt.
3. Does anything need to be done to get this working with asyncio?  What happens if you poll a DTLS socket, does that work?

tests/extmod/tls_dtls.py
```diff
@@ -0,0 +1,53 @@
...
+    def readinto(self, buf):
+        l = min(len(self.read_buffer), len(buf))
+        if l == 0:
+            return None
+        buf[:l] = self.read_buffer[:l]
+        self.read_buffer = self.read_buffer[l:]
+        return l
+
+    def ioctl(self, req, arg):
+        return 0
```
dpgeorge, line 29: Is this ioctl method needed?  If so, for which `req`uests?  Probably it should be:
```py
if req == 4:  # MP_STREAM_CLOSE
    return 0
return -1
```
Which follows a3128f89ccb1d24d0ddaec7164d2fa8a101b8ac3

ports/unix/mbedtls/mbedtls_config_port.h
```diff
@@ -32,6 +32,9 @@
 // Enable mbedtls modules
 #define MBEDTLS_TIMING_C
 
+// Enable DTLS
+#define MBEDTLS_SSL_PROTO_DTLS
```
dpgeorge, line 36: To enable this feature on all ports that use mbedtls, this can be moved to `extmod/mbedtls/mbedtls_config_common.h`.

keenanjohnson:
Thanks @dpgeorge for the feedback. Understood on the docs, and I'm happy to try and split the docs modules as a way to give back as a next PR after this one lands.

I see your other comment on how to enable for all ports and I can do that now. 

As for asyncio, it does work well! The project where I am using DTLS for COAP is all built in async. You can see how we use it here as everything is open-source: https://github.com/Ribbit-Network/ribbit-network-frog-software/blob/main/modules/ribbit/coap/__init__.py

ports/unix/mbedtls/mbedtls_config_port.h
keenanjohnson: Moved to the common location!

tests/extmod/tls_dtls.py
keenanjohnson: I believe it's just close yes.

keenanjohnson:
Looks like the windows build had some sort of temporary failure yesterday downloading the windows SDK, but after force pushing the branch again, the windows builds seem resolved.

keenanjohnson:
I saw that you self requested the review again @projectgus and I appreciate it! I believe I've addressed all previous feedback, so to should be ready. Appreciate you and all the work you do!

projectgus:
> I saw that you self requested the review again @projectgus and I appreciate it! I believe I've addressed all previous feedback, so to should be ready. Appreciate you and all the work you do!

Looks like Damien will do the final review for this, but thanks for the kind words!

dpgeorge:
@keenanjohnson I'm doing a final review here and trying to test it, but not having any luck.  Do you have some (simple) example (Micro)Python code for both a DTLS server and DTLS client that can talk to each other, just send a small amount of data between each other?

extmod/modtls_mbedtls.c
```diff
@@ -37,6 +37,8 @@
 #include "py/stream.h"
 #include "py/objstr.h"
 #include "py/reader.h"
+#include "py/smallint.h"
```
dpgeorge, line 40: This header shouldn't be needed anymore.

tests/extmod/tls_dtls.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+class DummySocket(io.IOBase):
+    def __init__(self):
+        self.write_buffer = bytearray()
+        self.read_buffer = bytearray()
+
+    def write(self, data):
+        self.write_buffer.extend(data)
+        return len(data)
```
dpgeorge, line 18: The `write_buffer` is never actually used.  So this function could just do `return len(data)`.

tests/extmod/tls_dtls.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def readinto(self, buf):
+        l = min(len(self.read_buffer), len(buf))
+        if l == 0:
+            return None
```
dpgeorge, line 23: The `read_buffer` is always empty, so this function could just always do `return None`.

dpgeorge:
I finally managed to write a test where MicroPython is both a DTLS server and client, and they connect to each other, do a handshake, and transfer some data between themselves.

It works well for unix<->unix.

It works well for unix<->PYBD_SF2 and unix<->RPI_PICO_W, when unix is the DTLS server.

But when PYBD_SF2 or RPI_PICO_W is the DTLS server then it rarely works.  I'm pretty sure this is because the lwIP driver on those boards can only queue one UDP packet at a time.  With a quick hack to queue up to two UDP packets, these boards as a DTLS server work much more reliably.

Note: to properly implement a DTLS server, we need to implement `socket.recvfrom(n, MSG_PEEK)` (I worked around this in my test, but it would be good to add that peek feature, it's quite easy to add).

keenanjohnson:
Hey @dpgeorge ! Thanks for taking a look. I was traveling for a few days and didn't get around to emails I guess. 

That makes sense that there might be some issues with the DTLS server on all ports. I've only tested on unix and esp32 and mostly with the device as a client rather than the server (which is the more common usecase I think). 

I've addressed your small comments above and as for the message peek feature should we try to implement that in this PR or would you prefer I remove the server functionality in this PR for now and address adding that with the peak in a separate changeset as I imagine the desire for the server side of DTLS is pretty small in the community?

tests/extmod/tls_dtls.py
keenanjohnson: Changed!

extmod/modtls_mbedtls.c
keenanjohnson: Removed!

tests/extmod/tls_dtls.py
keenanjohnson: Changed!

keenanjohnson:
If we do want to add the peek feature here, I assume we would modify the extmod/modsocket.c implementation of recvfrom()? Also if you post your test code @dpgeorge I can try to verify with the exact same code here to save the hassle of me trying to recreate what you've already done.

keenanjohnson:
Resolved a small merge conflict.

dpgeorge:
> Also if you post your test code @dpgeorge I can try to verify with the exact same code here to save the hassle of me trying to recreate what you've already done.

I've now added a commit to this PR with the test that I made.  It runs well under unix/unix with the multi-test running:
```
$ cd tests
$ ./run-multitests.py multi_net/tls_dtls_server_client.py
```
That should pass.  It should run as part of the CI as well.

> If we do want to add the peek feature here,

No, let's not add that here, it's a separate feature.  And it's not necessary for the test.

And let's keep this PR as-is with both server and client support.  It works fine with the unix port.

keenanjohnson:
Great and thanks for your support! I think the summarize there two follow up change-sets we've discussed after this is merged:

1) Splitting out the TLS / SSL docs to make them more reasonable.
2) Implementing the socket.recvfrom(n, MSG_PEEK) to fully support the DTLS server on PYBD_SF2 or RPI_PICO_W

Is there anything else open on this PR that I can help with?

dpgeorge:
> Is there anything else open on this PR that I can help with?

Were you able to successfully run the test that I added?  If you are happy with that test, then this PR should be good to go.

keenanjohnson:
Yes the test runs locally for me! I can also confirm that the CI system runs it successfully! 
https://github.com/micropython/micropython/actions/runs/13318657734/job/37198676006#step:4:3454

dpgeorge (approved):

dpgeorge:
Thanks for testing.

Now merged!

keenanjohnson:
Thank you for the review!

PR #15767, opened 2024-09-02
ports/unix/modffi: Fix signed integer cast in return_ffi_value.

mchlswyr:
### Summary

This PR builds on #7358. In `return_ffi_value`, integers of type `char`, `short`, and `long` are cast to type `int`. In an x86-64 Linux build, `long` is 64-bit and `int` is 32-bit, so the value is truncated.

### Testing

I added some `long` tests. The tests could not be added to the existing ffi_types test because two .exp files were required for the 32-bit and 64-bit results. Code common to both the ffi_types and `long` tests was factored into ffi_int_base. ffi_types was renamed to ffi_int_types to group the related tests under the "ffi_int" prefix.

I ran the tests in the following four cases:

1. Old modffi.c and `MICROPY_FORCE_32BIT=0`: ffi_int_long64 will fail;
2. Old modffi.c and `MICROPY_FORCE_32BIT=1`: all tests will pass or skip;
3. New modffi.c and `MICROPY_FORCE_32BIT=0`: all tests will pass or skip;
4. New modffi.c and `MICROPY_FORCE_32BIT=1`: all tests will pass or skip.


projectgus (approved):
This looks very good to me, especially the additional test coverage. Thanks, @mchlswyr!

dpgeorge (approved):
Thanks for the contribution!

dpgeorge:
Rebased and merged in b05983ff6d684561c8db6257ff816ef2e8114db5 and 230e521515d357816b925f13df4c5100a0e720d9

PR #15782, opened 2024-09-04
esp32: Fix ARDUINO_NANO_ESP32 build configuration.

projectgus:
### Summary

Regression introduced by https://github.com/micropython/micropython/pull/15727#issuecomment-2326340698 now that `MICROPY_HW_USB_CDC` is set.

The ARDUINO_NANO_ESP32 specifically builds shared/tinyusb/mp_usb_cdc.c for the 1200bps reset behaviour. However MicroPython esp32 doesn't yet use the rest of the shared/tinyusb functionality.

### Testing

I've verified this fixes the build for this board, and have eyeballed mp_usb_cdc.c to see that it should compile an equivalent binary to before `MICROPY_HW_USB_CDC` was defined. However I don't have a suitable board to test this on.


projectgus:
@andrewleech Heads-up about this board for your work in #15108, unless you've already looked into it. I think most of the board-specific changes (i.e. explicitly adding `${MICROPY_DIR}/shared/tinyusb/mp_usbd_cdc.c` from mpconfigboard.cmake) can be cleaned up when this is added, plus this commit can be reverted - so it should end up a lot tidier.

dpgeorge:
Thanks, this looks good to me.  The change is confined to the `ARDUINO_NANO_ESP32` board so that limits its scope, which is good.

For reference: commit 84a8f7ea6dfa3529b024d31fa25bb91153bbfc35 added `MICROPY_EXCLUDE_SHARED_TINYUSB_USBD_CDC` for the nrf port, but that was recently no longer used by the nrf port in 19075695da82e30850e8f595c58521d2b3787b4a .  So this option can be deleted once esp32 moves to the shared TinyUSB code.

dpgeorge (approved):
I verified that this compiled successfully with IDF 5.2.2.

PR #15784, opened 2024-09-04
qemu: Rename qemu-arm port to qemu.

dpgeorge:
Because this port now supports multiple architectures.

Follow-up to #15743.


dpgeorge:
@agatti does this rename look good to you?

agatti:
At a first glance it looks OK, although for #15714 I will have to rename `ports/qemu/test-frzmpy/frozen_asm.py` to make it unique to the Arm port and add a similar file for RV32 (and modify the makefiles to reflect that of course).

dpgeorge:
> I will have to rename `ports/qemu/test-frzmpy/frozen_asm.py` to make it unique to the Arm port

That sounds fine, you can do that as part of #15714.

dpgeorge:
@agatti this is now merged.

PR #15785, opened 2024-09-04
rp2/modules: Fix FatFS boot script.

### LINKED ISSUE #15779: rp2: FatFS filesystem is not recreated after erasing the flash.

rp2 port, Arduino Nano-RP2040

### MicroPython version

MicroPython v1.24.0-preview.260.g09d070aa5

### Reproduction

This particular board has a 16MBytes flash, 2MBytes of which are reserved for the firmware, the rest is for the filesystem. A write of about ~3MBs should be enough to erase the firmware + the filesystem's partition table, but I tested with 16MBytes:

```bash
dd if=/dev/zero of=flash.bin bs=1M count=16
```

Get the board in bootloader mode, with double tap (on Arduino board) or jumper on other boards etc.. then write the binary to flash (erasing everything):
```bash
picotool load flash.bin
```

After it's done, and reset, the ROM bootloader runs. Load MicroPython, for example:
```bash
picotool load -x ARDUINO_NANO_RP2040_CONNECT.uf2
```

Try to use the filesystem:
```Python
>>> import os
>>> os.stat(".")
OSError: [Errno 19] ENODEV
```

### Expected behaviour

I expected it to detect that there's no valid filesystem and re-create it.

### Observed behaviour

The board boots MicroPython, the storage device is mounted, however it doesn't have a valid filesystem now:

```Python
>>> import os
>>> os.stat(".")
OSError: [Errno 19] ENODEV

>>> import vfs
>>> import machine, rp2
>>> bdev = rp2.Flash()
>>> fs = vfs.VfsFat(bdev)
>>> vfs.mount(fs, "/")
OSError: [Errno 19] ENODEV
```

### Additional Information

### Code of Conduct

Yes, I agree

iabdalkader:
This change helps detect if the filesystem is invalid and reformat the FAT if needed.

Fixes #15779

### DIFF
--- a/ports/rp2/modules/_boot_fat.py
+++ b/ports/rp2/modules/_boot_fat.py
@@ -5,10 +5,9 @@ import machine, rp2
 # Try to mount the filesystem, and format the flash if it doesn't exist.
 bdev = rp2.Flash()
 try:
-    fs = vfs.VfsFat(bdev)
+    vfs.mount(vfs.VfsFat(bdev), "/")
 except:
     vfs.VfsFat.mkfs(bdev)
-    fs = vfs.VfsFat(bdev)
-vfs.mount(fs, "/")
+    vfs.mount(vfs.VfsFat(bdev), "/")
 
-del vfs, bdev, fs
+del vfs, bdev


ports/rp2/modules/_boot_fat.py
```diff
@@ -6,9 +6,10 @@
 bdev = rp2.Flash()
 try:
     fs = vfs.VfsFat(bdev)
+    vfs.mount(fs, "/")
```
dpgeorge, line 9: Maybe this (and the corresponding lines in the "except" below) can be `vfs.mount(vfs.VfsFat(bdev), "/")`, and then the `fs` variable is not needed and the `del fs` can be removed below.

iabdalkader: Done.

dpgeorge (approved):

dpgeorge:

PR #15786, opened 2024-09-04
add support for new custom board - CircuitArt ESP32S3 zero

CircuitART:
This is a custom board that i developed it using the same footprint as pi zero. 
the PID was obtained from Espressif   https://github.com/espressif/usb-pids/pull/33
I have already build and tested circuitpython on it, and so far all is good


CircuitART:
wrong repo

CircuitART:
wrong repo

PR #15787, opened 2024-09-04
samd/machine_uart.c: Add full support for 9bit data.

robert-hh:
Before, 9bit data could be specified in the constructor and was transmitted, but the 9th bit was set to 0 when sending, and ignored when receiving. This PR completes 9 bit support in that the 9th bit is taken from the data.

9 bit data has to be provided with uart.write() and and read  with uart.read() as two bytes for each transmitted item, low order byte first. The data length supplied with uart.write() and requested by uart.read() has to be even, which is checked. The size of the UART buffers will be transparently doubled to cater for 9bit data.


dpgeorge (approved):

dpgeorge:
Thank you!

robert-hh:
And you.

PR #15788, opened 2024-09-04
Remove tinytest framework, and `--write-exp` and `--list-tests` options from `run-tests.py`

dpgeorge:
### Summary

With the recent qemu (#15624 and #15743) and zephyr (#15678) changes to how their tests are run, two things became unused:
- the tinytest framework, which embedded a set of tests and their expected output within firmware, so these tests could be run stand-alone
- the `--write-exp` and `--list-tests` options to `tests/run-tests.py`, which were needed primarily to generated the expected test output for tinytest (also the associated `tests/run-tests-exp.py`/`.sh` scripts are unused)

This PR removes both of these things.

### Rationale

Removing tinytest removes now-unused code and reduces a maintenance burden.

Removing `--write-exp` and `--list-tests` options helps to simplify the rather complex logic in `run-tests.py`.  In particular this opens the door to make it so `run-tests.py` can automatically detect the target/platform that it's running on (that's for a follow-up PR).

### Testing

All code removed here is unused.  The existing tests and CI should not change.

We could leave this functionality here, in case it's useful for people to use in their own project.  But IMO that's a burden on this project that we don't need.  And we'd need to add code to test these now-unused components, otherwise they will eventually break as other things change around them.


dpgeorge:
@stinos @dlech I would appreciate your review/comments on this PR.  You are the most likely to be using these now-obsolete features, and if you're not then it's unlikely anyone else is :smile: 

I'd rather not keep these unused parts, but if you have a strong reason to keep them please let me know.

stinos:
I'm not actively using --list-tests and don't think I ever used --write-exp so as far as I'm concerned these can be removed.

dlech:
Thanks for the heads-up. We are using `lib/tinytest` in our Pybricks tree. But it will be no problem to just copy that directory to our own tree.

dpgeorge:
> We are using `lib/tinytest` in our Pybricks tree

Ah, that's interesting.  Do you find it easier/necessary to use tinytest over just running the tests through the serial port?

dlech:
We actually just use it for unit testing C code, not MicroPython stuff.

dpgeorge:
> We actually just use it for unit testing C code

Right.  So you don't use `upytest_execute_test`, or probably any of those helper functions.

Then it is better for you to just maintain a copy of the tinytest library.  Not really much point in us having it here.

projectgus (approved):
Always a good day when you can delete code without losing functionality!

(And in general I think this set of changes to run all tests through the same codepaths is really beneficial!)

Issue #15795, opened 2024-09-05
Build fails for STM32L452XX with USB enabled


higstar:
### Port, board and/or hardware

stm32 STM32L452xx

### MicroPython version

v1.23.0-287-g042693496

### Reproduction

Edit NUCLEO_L452RE's mpconfigboard.h adding the following:
```c
#define MICROPY_HW_ENABLE_USB       (1)
#define MICROPY_HW_USB_FS           (1)

```

### Expected behaviour

Expected the build to complete successfully.

### Observed behaviour

Build fails as follows:
```
...
stm32_it.c: In function 'OTG_CMD_WKUP_Handler':
stm32_it.c:396:9: error: implicit declaration of function '__HAL_PCD_UNGATE_PHYCLOCK' [-Werror=implicit-function-declaration]
  396 |         __HAL_PCD_UNGATE_PHYCLOCK(pcd_handle);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~
```

### Additional Information

Patching as follows gets things underway:
```bash
sed -i 's_#elif defined(STM32L432xx)$_#elif defined(STM32L432xx) || defined(STM32L452xx)_g' ports/stm32/usbd_conf.c
sed -i 's_defined(STM32L1) || defined(STM32L432xx) || defined(STM32WB)_defined(STM32L1) || defined(STM32L432xx) || defined(STM32L452xx) || defined(STM32WB)_g' ports/stm32/mpconfigboard_common.h
sed -i 's_#elif defined(STM32L0) || defined(STM32L432xx)$_#elif defined(STM32L0) || defined(STM32L432xx) || defined(STM32L452xx)_g' ports/stm32/stm32_it.c
```

### Code of Conduct

Yes, I agree

Tico06:
Hi @higstar ,

I made a test by connecting an USB module to a NUCLEO-L452RE.

I patched the code with your sed proposal without any error.

And updated mpconfigboard.h, commenting REPL config, and adding USB config:
```c
// USART2 is connected to the ST-LINK USB VCP
// #define MICROPY_HW_UART_REPL        PYB_UART_2
// #define MICROPY_HW_UART_REPL_BAUD   115200

// USB config
#define MICROPY_HW_ENABLE_USB       (1)
#define MICROPY_HW_USB_FS           (1)
```

Then, first run a make clean:
```
$ make clean BOARD=NUCLEO_L452RE
[...]

PR #15796, opened 2024-09-05
ports/stm32: Updated STM32L452XX Multi OTG handling

higstar:
Added custom Multi OTG handling for STM32L452XX
based on STM32L432XX handling.
https://github.com/micropython/micropython/issues/15795

### Summary

To enable USB when using an L4 which is not based on the L432, Multi OTG handling seems to fall through the cracks, and results in the build not completing.

### Testing

I do not yet have HW, so have not been able to test on a L452 target.

Ideally the multi OTG handling would not involve piecemeal handling for specific targets, which then needs springling around various files.

### DIFF
--- a/ports/stm32/mpconfigboard_common.h
+++ b/ports/stm32/mpconfigboard_common.h
@@ -619,7 +619,7 @@
 // Whether the USB peripheral is device-only, or multiple OTG
 // For STM32G0 and STM32H5 the USB peripheral supports device and host mode,
 // but otherwise acts like a non-multi-OTG peripheral.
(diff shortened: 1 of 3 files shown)


andrewleech:
On a tangentially related note, if / when this is completed it should simplify USB compatibility and remove the need for per-chip-type handling of all these: https://github.com/micropython/micropython/pull/15592

mattytrentini:
I couldn't find it after a quick look, but I do have a [Blues Wireless Swan](https://blues.com/products/swan/) board _somewhere_. The micro it's built on is the [STM32L4R5ZIY6](https://dev.blues.io/datasheets/swan-datasheet) - if that's useful to test this PR on then let me know and I'll hunt down the board...

dpgeorge (approved):
Thanks for the contribution!

This looks fine to me.  I haven't tested it but it's a very low risk change.

PR #15800, opened 2024-09-05
tools/ci.sh: Clean up Arm, MIPS, and RV64 Unix port build steps.

agatti:
### Summary

These are some minor patches I've made on my tree when I added the Unix RV64 target to the CI scripts.  I forgot to submit them back then, so here they are, rebased onto the latest master and all.  In short, the MIPS target now doesn't require any workarounds to run under CI and passes FFI tests, the RV64 target now has an FFI helper, and the Arm build doesn't require an environment variable to run anymore (now shares the same post-installation setup commands as MIPS).

### Testing

When building and running the Arm and MIPS targets locally in a VM environment equivalent to the CI image (same OS, packages, and run commands), the VFS tests pass.  No idea on why they must be disabled in the CI environment - the RV64 build passes them anyway.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -410,7 +410,6 @@ CI_UNIX_OPTS_QEMU_MIPS=(
     CROSS_COMPILE=mips-linux-gnu-
     VARIANT=coverage
     MICROPY_STANDALONE=1
-    LDFLAGS_EXTRA="-static"
 )
 
 CI_UNIX_OPTS_QEMU_ARM=(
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):

dpgeorge:
Thanks for this, it's a good improvement.

I struggled a lot to get all this working initially.  Good to see some of the remaining issues cleaned up!

> When building and running the Arm and MIPS targets locally in a VM environment equivalent to the CI image (same OS, packages, and run commands), the VFS tests pass. No idea on why they must be disabled in the CI environment - the RV64 build passes them anyway.

That vfs test failure on CI only is strange...

dpgeorge:
@agatti it seems like a combination of this and #15767 has lead to CI failures, eg see https://github.com/micropython/micropython/actions/runs/10936894284/job/30361527498

Are you able to look into that?

agatti:
From a first look, the function call's result seems to be trashed by libffi's post-call cleanup code which sort of lets me think those extra tests might have exposed an incompatibility with the version of libffi RV64 uses.

RV64 uses a newer, pre-packaged version of libffi (see #15528), as the libffi linked by the MicroPython repo is just one or two versions before RV64 support was introduced.

Later today I'll try to see if using a source-built libffi version fixes the problem or not, although in that case the newer libffi Arm build issue still needs to be sorted out (or FFI tests are going to be disabled for RV64 for the time being, either works).

agatti:
@dpgeorge The prebuilt Ubuntu libffi package is probably at fault here.

After updating the libffi module to 3.4.6 and changing the build configuration to pick up the local libffi copy the failures I get are different, but not due to trashed results.  

Now the error diff is as such:

```diff
--- ports_unix_ffi_int_types.py.exp     2024-09-19 17:58:30.422200400 +0200
+++ ports_unix_ffi_int_types.py.out     2024-09-19 17:58:30.423200405 +0200
@@ -93,12 +93,12 @@
 f32u(ffff) = fffe
 f32u(10000) = 10001
 f32u(7fffffff) = 7ffffffe
-f32u(80000000) = 80000001
-f32u(ffffffff) = fffffffe
+f32u(80000000) = ffffffff80000001
+f32u(ffffffff) = fffffffffffffffe
 f32u(100000000) = 1
-f32u(7fffffffffffffff) = fffffffe
+f32u(7fffffffffffffff) = fffffffffffffffe
 f32u(8000000000000000) = 1
-f32u(ffffffffffffffff) = fffffffe
+f32u(ffffffffffffffff) = fffffffffffffffe
 f32u(10000000000000000) = 1
 f64i(0) = 1
 f64i(7f) = 7e
```

PR #15803, opened 2024-09-06
qemu/arm: Let `NETDUINO2` and `MICROBIT` boards run tests.

agatti:
### Summary

The NETDUINO2 and MICROBIT boards would not run the test suite in their current form.  The former would make the test runner bail out before a REPL is ready, the latter would trigger a CPU hard fault due to Qemu's soft MMU enabled for that particular machine definition (see details).

As a bonus, Thumb targets now have a slightly more appealing crash reporting, with the exception cause and the registers content dumped to STDOUT.

<details>

(`armv7m_nvic_neg_prio_requested` triggers the exception - a NVIC priority of -1 signals a hard fault)

```
Thread 1 "qemu-system-arm" hit Breakpoint 1, 0x0000565ed7ad5fb4 in armv7m_nvic_neg_prio_requested (secure=<optimized out>, opaque=<optimized out>) at ../../hw/intc/armv7m_nvic.c:404
404     ../../hw/intc/armv7m_nvic.c: No such file or directory.
(gdb) bt
#0  0x0000565ed7ad5fb4 in armv7m_nvic_neg_prio_requested (secure=<optimized out>, opaque=<optimized out>) at ../../hw/intc/armv7m_nvic.c:404
#1  arm_v7m_mmu_idx_for_secstate_and_priv (priv=<optimized out>, secstate=<optimized out>, env=<optimized out>) at ../../target/arm/m_helper.c:2889
#2  arm_v7m_mmu_idx_for_secstate (env=0x565ef2870940, secstate=false) at ../../target/arm/m_helper.c:2900
#3  0x0000565ed7acff9c in arm_mmu_idx (env=0x565ef2870940) at ../../target/arm/helper.c:13063
#4  arm_cpu_get_phys_page_attrs_debug (cs=<optimized out>, addr=70656, attrs=0x7fff80f5be64) at ../../target/arm/helper.c:12591
#5  0x0000565ed7b9d2dc in cpu_memory_rw_debug (cpu=<optimized out>, addr=70952, ptr=<optimized out>, len=2, is_write=<optimized out>)
    at ../../softmmu/physmem.c:3473
#6  0x0000565ed7bb8ddb in target_memory_rw_debug (is_write=false, len=2, buf=0x565ef2b84d40 "", addr=70952, cpu=0x565ef2866bc0) at ../../gdbstub.c:85
#7  target_memory_rw_debug (is_write=false, len=2, buf=0x565ef2b84d40 "", addr=70952, cpu=0x565ef2866bc0) at ../../gdbstub.c:65
#8  handle_read_mem (user_ctx=<optimized out>, params=<optimized out>) at ../../gdbstub.c:1803
#9  handle_read_mem (params=<optimized out>, user_ctx=<optimized out>) at ../../gdbstub.c:1787
#10 0x0000565ed7e677de in process_string_cmd.constprop.0 (data=<optimized out>, cmds=<optimized out>, num_cmds=<optimized out>, user_ctx=0x0)
    at ../../gdbstub.c:1499
#11 0x0000565ed7bbaa0f in run_cmd_parser (data=<optimized out>, cmd=0x565ed8357c80 <read_mem_cmd_desc>) at ../../gdbstub.c:1517
#12 gdb_handle_packet (line_buf=<optimized out>) at ../../gdbstub.c:2728
#13 gdb_read_byte (ch=99 'c') at ../../gdbstub.c:3064
#14 gdb_chr_receive (opaque=<optimized out>, buf=<optimized out>, size=<optimized out>) at ../../gdbstub.c:3365
#15 0x0000565ed7dacc02 in tcp_chr_read (chan=<optimized out>, cond=<optimized out>, opaque=<optimized out>) at ../../chardev/char-socket.c:564
#16 0x0000781affd56c44 in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#17 0x0000565ed7e56248 in glib_pollfds_poll () at ../../util/main-loop.c:232
#18 os_host_main_loop_wait (timeout=1000000000) at ../../util/main-loop.c:255
#19 main_loop_wait (nonblocking=<optimized out>) at ../../util/main-loop.c:531
#20 0x0000565ed7ba222c in qemu_main_loop () at ../../softmmu/runstate.c:726
#21 0x0000565ed771ffb2 in main (argc=<optimized out>, argv=<optimized out>, envp=<optimized out>) at ../../softmmu/main.c:50
```

</details>

### Testing

Ran the tests for the two boards in question under Qemu (where else?).

There should be a better way to disable soft MMU in Qemu and let MICROBIT still emit native code, but I haven't found a proper solution at the moment.

### DIFF
--- a/ports/qemu/mcu/arm/startup.c
+++ b/ports/qemu/mcu/arm/startup.c
@@ -28,6 +28,8 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "py/runtime.h"
+
 #include "shared/runtime/semihosting_arm.h"
 #include "uart.h"
 
@@ -50,7 +52,7 @@ __attribute__((naked)) void Reset_Handler(void) {
     _start();
 }
 
-void Default_Handler(void) {
+NORETURN void Default_Handler(void) {
     for (;;) {
     }
 }
@@ -74,25 +76,35 @@ __attribute__((naked, section(".isr_vector"))) void isr_vector(void) {
 
 #elif defined(__ARM_ARCH_ISA_THUMB)
 
+extern void NMI_Handler(void);
+extern void HardFault_Handler(void);
+extern void MemManage_Handler(void);
+extern void BusFault_Handler(void);
+extern void UsageFault_Handler(void);
+extern void SVC_Handler(void);
+extern void DebugMon_Handler(void);
+extern void PendSV_Handler(void);
+extern void SysTick_Handler(void);
+
 // ARM architecture with Thumb-only ISA.
 
 const uint32_t isr_vector[] __attribute__((section(".isr_vector"))) = {
     (uint32_t)&_estack,
     (uint32_t)&Reset_Handler,
-    (uint32_t)&Default_Handler, // NMI_Handler
-    (uint32_t)&Default_Handler, // HardFault_Handler
-    (uint32_t)&Default_Handler, // MemManage_Handler
-    (uint32_t)&Default_Handler, // BusFault_Handler
-    (uint32_t)&Default_Handler, // UsageFault_Handler
+    (uint32_t)&NMI_Handler,
+    (uint32_t)&HardFault_Handler,
+    (uint32_t)&MemManage_Handler,
+    (uint32_t)&BusFault_Handler,
+    (uint32_t)&UsageFault_Handler,
     0,
     0,
     0,
     0,
-    (uint32_t)&Default_Handler, // SVC_Handler
-    (uint32_t)&Default_Handler, // DebugMon_Handler
+    (uint32_t)&SVC_Handler,
+    (uint32_t)&DebugMon_Handler,
     0,
-    (uint32_t)&Default_Handler, // PendSV_Handler
-    (uint32_t)&Default_Handler, // SysTick_Handler
+    (uint32_t)&PendSV_Handler,
+    (uint32_t)&SysTick_Handler,
 };
 
 #endif
--- a/ports/qemu/Makefile
+++ b/ports/qemu/Makefile
@@ -44,6 +44,7 @@ LDFLAGS += -nostdlib
 LIBS = $(shell $(CC) $(CFLAGS) -print-libgcc-file-name)
 
 SRC_C += \
+	mcu/arm/errorhandler.c \
 	mcu/arm/startup.c \
 	shared/runtime/semihosting_arm.c \
 
--- a/ports/qemu/Makefile
+++ b/ports/qemu/Makefile
@@ -20,7 +20,11 @@ MICROPY_ROM_TEXT_COMPRESSION ?= 1
 
 ifeq ($(QEMU_ARCH),arm)
 MICROPY_HEAP_SIZE ?= 122880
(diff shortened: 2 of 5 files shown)


agatti:
This PR is focused on making the two boards in question run the test suite, although whilst `NETDUINO2` tests pass `MICROBIT` now yield failures when running inline asm tests (expected) but "framebuf_polygon", "int_big_div", and "int_big_mul" also fail - which as far as I can remember they used to pass before (but don't quote me on that).

The asm tests issues are going to disappear once the inline asm detector in "run-tests.py" gets merged from the inline rv32 asm PR, but no real clue about the other test failures - haven't attached gdb yet.  Will do soon.

ports/qemu/mcu/arm/startup.c
```diff
@@ -27,6 +27,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdnoreturn.h>
```
dpgeorge, line 30: Is this header C99?

Can we instead just use NORETURN which is already defined by us?

ports/qemu/boards/MICROBIT.mk
```diff
@@ -11,3 +11,6 @@ LDSCRIPT = mcu/arm/nrf51.ld
 SRC_BOARD_O = shared/runtime/gchelper_native.o shared/runtime/gchelper_thumb1.o
 
 MPY_CROSS_FLAGS += -march=armv7m
+
+# Inline assembler tests are disabled for this board, so exclude them.
+RUN_TESTS_ARGS = --exclude 'inlineasm|qemu/asm_test'
```
dpgeorge, line 16: Doesn't `run-tests.py` now include logic to skip these (or at least the `inlineasm`) tests if the target doesn't support them?

ports/qemu/mcu/arm/startup.c
agatti: I thought this was C99 but turns out it's C11.  I recall I looked for `MP_NORETURN` but found nothing, so I guess that's an exception to the naming convention.

Still, thanks for noticing, should be addressed now.

ports/qemu/boards/MICROBIT.mk
agatti: Missed this during the last rebase, thanks for noticing!  This should be fixed now.

ports/qemu/mcu/arm/startup.c
dpgeorge: I also first grep'd for `MP_NORETURN` before realising it's `NORETURN`.

It would probably be good for us to rename it to `MP_NORETURN`.

dpgeorge (approved):

ports/qemu/mcu/arm/startup.c
agatti: > It would probably be good for us to rename it to `MP_NORETURN`.

I came up with https://github.com/agatti/micropython/commit/981f81e0e1d913bf08f40c2f970b2f70b554f697 but I'll wait for this PR to be merged and update my commit before creating a new PR for the macro rename.

PR #15812, opened 2024-09-09
tools/mpy_ld.py: Support jumping more than 2k on armv6m architectures.

dpgeorge:
### Summary

Native .mpy files targetting armv6m (eg RP2040) cannot currently have more than about 2kiB of native code (between the start of the file and the init function).

This PR fixes that by using bigger jumps to jump to the init function.

### Testing

Tested with `examples/natmod/features0/` by forcing `mpy_ld.py` to always emit the long jump.  It works.

### DIFF
--- a/tools/mpy_ld.py
+++ b/tools/mpy_ld.py
@@ -89,10 +89,26 @@ def asm_jump_x86(entry):
 
 
 def asm_jump_thumb(entry):
-    # Only signed values that fit in 12 bits are supported
+    # This function must return the same number of bytes for the encoding of the jump
+    # regardless of the value of `entry`.
     b_off = entry - 4
-    assert b_off >> 11 == 0 or b_off >> 11 == -1, b_off
-    return struct.pack("<H", 0xE000 | (b_off >> 1 & 0x07FF))
+    if b_off >> 11 == 0 or b_off >> 11 == -1:
+        # Signed value fits in 12 bits.
+        b0 = 0xE000 | (b_off >> 1 & 0x07FF)
+        b1 = 0
+        b2 = 0
+        b3 = 0
+    else:
+        # Use bl to do a large jump/call:
+        #   push {r0, lr}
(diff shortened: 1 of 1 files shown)


dpgeorge:
This is an alternative to #12241 that uses less bytes (8) to encode the jump.

jonnor:
This looks nice and simple. I will test it later. Probably the -2 adjustment of the offset is what I missed when I tried to implement this.

> # This function must return the same number of bytes for the encoding of the jump regardless of the value of `entry`.

As it stands, the function does not obey this property - it's either 2 or 4 bytes. As you mentioned in the review of my code, this can cause issues because the length of instructions is used else where to compute the layout. Is the plan to skip the conditional and always do the long jump (once verified)?

dpgeorge:
> As it stands, the function does not obey this property - it's either 2 or 4 bytes

No, it always emits 8 bytes.  See the `struct.pack("<HHHH", ...)` at the end of the function.  There is no other return in the function.

vshymanskyy:
I have run this change with wasm2mpy tests (with and without forcing long jumps) and noticed no regressions.
Coremark builds fine, but still doesn't run probably due to unknown reason.

jonnor:
> > As it stands, the function does not obey this property - it's either 2 or 4 bytes
> 
> No, it always emits 8 bytes. See the `struct.pack("<HHHH", ...)` at the end of the function. There is no other return in the function.

You are right, of course. I do not know where my head was at.

jonnor:
Tested with a module that exceeds the 2k limitation (before this commit, fails with `AssertionError: 3295` in `assert b_off >> 11 == 0 or b_off >> 11 == -1, b_off`). Builds successfully and looks to run ok on a RPI Pico. So I think this is good to go @dpgeorge !

dpgeorge:
Thanks @jonnor and @vshymanskyy for testing and review!

PR #15813, opened 2024-09-09
examples/natmod: Fix links in README.md.

mattytrentini:
I noticed that the links in the examples/natmod/README.md file were incorrectly formatted (can't split lines between the square brackets and parentheses) and weren't rendering correctly when viewed in Github. The first commit resolves those links.

I used [mdl](https://github.com/markdownlint/markdownlint) to lint the README and tidied up a couple of other suggestions too (specifically [MD014](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md014---dollar-signs-used-before-commands-without-showing-output), and [MD007](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md007---unordered-list-indentation)). These issues are tidied up in the second commit.

### DIFF
--- a/examples/natmod/README.md
+++ b/examples/natmod/README.md
@@ -1,12 +1,12 @@
 # Dynamic Native Modules
 
 Dynamic Native Modules are .mpy files that contain native machine code from a
-language other than Python. For more info see [the documentation]
-(https://docs.micropython.org/en/latest/develop/natmod.html).
+language other than Python. For more info see [the
+documentation](https://docs.micropython.org/en/latest/develop/natmod.html).
 
-This should not be confused with [User C Modules]
-(https://docs.micropython.org/en/latest/develop/cmodules.html) which are a
-mechanism to add additional out-of-tree modules into the firmware build.
+This should not be confused with [User C
+Modules](https://docs.micropython.org/en/latest/develop/cmodules.html) which are
+a mechanism to add additional out-of-tree modules into the firmware build.
 
 ## Examples
 


examples/natmod/README.md
```diff
@@ -1,33 +1,33 @@
...
-        - relocated pointers in rodata
+       * defining simple functions exposed to Python
+       * defining local, helper C functions
+       * defining constant integers and strings exposed to Python
+       * getting and creating integer objects
+       * creating Python lists
+       * raising exceptions
+       * allocating memory
+       * BSS and constant data (rodata)
+       * relocated pointers in rodata
```
dpgeorge, line 30: Would be good to stick with 4 space indentation.  That matches the rest of the project, and other .md files.

Does the linter also complain about * vs - bullet points?

examples/natmod/README.md
```diff
@@ -67,8 +67,8 @@ with `pip`.
...
 make (one of x86, x64, armv6m, armv7m, xtensa, xtensawin):
 
-```
-$ cd features0
-$ make ARCH=armv7m
-$ mpremote cp features0.mpy :
+```text
+cd features0
+make ARCH=armv7m
+mpremote cp features0.mpy :
```
dpgeorge, line 73: How about just making this a block indented by 4 spaces, without the three ticks?

Also, prefixing with dollar signs matches all the other README.md files, so I'd prefer to keep them.  They indicate you need to use the REPL.

examples/natmod/README.md
mattytrentini: I'll roll it back to four spaces. It generates one warning [MD007](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md007---unordered-list-indentation) for the unordered list; it really wants 3 spaces. But it's a benign warning.

Yes, it raises [MD004](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md004---unordered-list-style) warnings if I use '-'. Reading the warning though, 'consistent' should be the default style and that I don't think we're breaking that rule; could be a false positive? If you prefer '*' I can revert that change.

examples/natmod/README.md
mattytrentini: > How about just making this a block indented by 4 spaces, without the three ticks?

Can do but that it raises [MD046](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md046---code-block-style). I personally prefer the 'fencing' style (with backticks) because the language can be specified and so fencing can be used consistently, everwhere. But, it's up to you!

> Also, prefixing with dollar signs matches all the other README.md files, so I'd prefer to keep them. They indicate you need to use the REPL.

This raises [MD014](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md014---dollar-signs-used-before-commands-without-showing-output) and there's some justification for it. Probably the biggest issue is that copy/paste is more useful without the '$'. But again, up to you. 😄

mattytrentini:
I've backed out the more debatable minor issues. What's left is just the - more important! - fixes for the links.

dpgeorge (approved):

dpgeorge:
Thanks!

PR #15818, opened 2024-09-09
py/objset: Allow intersection() to take multiple sets as args.

AmirHmZz:
CPython allows `set.intersection()` to take multiple iterables as arguments and It seems `CPython 2.6` implemented this mechanism. Micropython currently allows only one iterable as argument and this PR aims to increase compatibility with CPython.

### DIFF
--- a/py/objset.c
+++ b/py/objset.c
@@ -227,47 +227,57 @@ static mp_obj_t set_diff_update(size_t n_args, const mp_obj_t *args) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR(set_diff_update_obj, 1, set_diff_update);
 
-static mp_obj_t set_intersect_int(mp_obj_t self_in, mp_obj_t other, bool update) {
-    if (update) {
-        check_set(self_in);
-    } else {
-        check_set_or_frozenset(self_in);
-    }
-
-    if (self_in == other) {
-        return update ? mp_const_none : set_copy(self_in);
-    }
-
-    mp_obj_set_t *self = MP_OBJ_TO_PTR(self_in);
-    mp_obj_set_t *out = MP_OBJ_TO_PTR(mp_obj_new_set(0, NULL));
-
-    mp_obj_t iter = mp_getiter(other, NULL);
+static mp_obj_set_t *set_intersect_one(mp_set_t *src, mp_obj_t iterable) {
+    mp_obj_set_t *out_ptr = MP_OBJ_TO_PTR(mp_obj_new_set(0, NULL));
+    mp_obj_t iter = mp_getiter(iterable, NULL);
     mp_obj_t next;
     while ((next = mp_iternext(iter)) != MP_OBJ_STOP_ITERATION) {
-        if (mp_set_lookup(&self->set, next, MP_MAP_LOOKUP)) {
-            set_add(MP_OBJ_FROM_PTR(out), next);
+        if (mp_set_lookup(src, next, MP_MAP_LOOKUP)) {
+            set_add(MP_OBJ_FROM_PTR(out_ptr), next);
+        }
+    }
+    return out_ptr;
+}
+
+static mp_obj_t set_intersect_int(size_t n_args, const mp_obj_t *args, bool update) {
+    mp_obj_set_t *in_ptr;
+    size_t start = 1;
+
+    if (update) {
+        check_set(args[0]);
+        in_ptr = MP_OBJ_TO_PTR(args[0]);
+    } else {
+        check_set_or_frozenset(args[0]);
+        if (n_args == 1) {
+            return set_copy(args[0]);
+        }
+        mp_obj_set_t *src = MP_OBJ_TO_PTR(args[0]);
+        in_ptr = set_intersect_one(&src->set, args[1]);
+        start = 2;
+    }
+
+    for (size_t i = start; i < n_args; i++) {
+        if (MP_OBJ_FROM_PTR(in_ptr) != args[i]) {
+            mp_obj_set_t *out_ptr = set_intersect_one(&in_ptr->set, args[i]);
+            m_del(mp_obj_t, in_ptr->set.table, in_ptr->set.alloc);
+            in_ptr->set.alloc = out_ptr->set.alloc;
+            in_ptr->set.used = out_ptr->set.used;
+            in_ptr->set.table = out_ptr->set.table;
         }
     }
 
-    if (update) {
-        m_del(mp_obj_t, self->set.table, self->set.alloc);
-        self->set.alloc = out->set.alloc;
-        self->set.used = out->set.used;
-        self->set.table = out->set.table;
-    }
-
-    return update ? mp_const_none : MP_OBJ_FROM_PTR(out);
+    return update ? mp_const_none : MP_OBJ_FROM_PTR(in_ptr);
 }
 
-static mp_obj_t set_intersect(mp_obj_t self_in, mp_obj_t other) {
-    return set_intersect_int(self_in, other, false);
+static mp_obj_t set_intersect(size_t n_args, const mp_obj_t *args) {
+    return set_intersect_int(n_args, args, false);
 }
-static MP_DEFINE_CONST_FUN_OBJ_2(set_intersect_obj, set_intersect);
+static MP_DEFINE_CONST_FUN_OBJ_VAR(set_intersect_obj, 1, set_intersect);
 
-static mp_obj_t set_intersect_update(mp_obj_t self_in, mp_obj_t other) {
-    return set_intersect_int(self_in, other, true);
+static mp_obj_t set_intersect_update(size_t n_args, const mp_obj_t *args) {
+    return set_intersect_int(n_args, args, true);
 }
-static MP_DEFINE_CONST_FUN_OBJ_2(set_intersect_update_obj, set_intersect_update);
+static MP_DEFINE_CONST_FUN_OBJ_VAR(set_intersect_update_obj, 1, set_intersect_update);
 
 static mp_obj_t set_isdisjoint(mp_obj_t self_in, mp_obj_t other) {
     check_set_or_frozenset(self_in);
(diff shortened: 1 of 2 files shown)


py/objset.c
```diff
@@ -227,29 +227,45 @@ static mp_obj_t set_diff_update(size_t n_args, const mp_obj_t *args) {
...
-            set_add(MP_OBJ_FROM_PTR(out), next);
+    mp_obj_t self_iter = mp_getiter(args[0], NULL);
+    mp_obj_t self_next;
+    while ((self_next = mp_iternext(self_iter)) != MP_OBJ_STOP_ITERATION) {
+        bool skip = false;
+        for (size_t i = 1; i < n_args; i++) {
+            mp_obj_t iter = mp_getiter(args[i], NULL);
+            mp_obj_t next;
+            bool present = false;
+            while ((next = mp_iternext(iter)) != MP_OBJ_STOP_ITERATION) {
```
dlech, line 251: This looks like it is doing a linear search now instead of `mp_set_lookup()` (which uses a hash-based search), so we would need to see some performance measurements to compare this to the previous implementation for large sets.

AmirHmZz: I've taken a look at CPython and it seems CPython has a function for single intersection and a function for multi intersections (takes multiple iterables) which calculates `n-1` intersections. I guess code size increase to define a new function is more than this linear search BTW.

dpgeorge: We definitely don't want to reduce the performance of the existing hash-based approach.

projectgus: Hmm... What about keeping the existing while loop structure around the first argument, and checking any additional args inside the loop?

Semi-pseudocode:

```
self = args[0];
other = args[1];

 mp_obj_t iter = mp_getiter(other, NULL);
 mp_obj_t next;
 while ((next = mp_iternext(iter)) != MP_OBJ_STOP_ITERATION) {
        if (mp_set_lookup(&self->set, next, MP_MAP_LOOKUP)) {
            for (int i = 2; i < n_args; i++) {
                if (!mp_binary_op(MP_BINARY_OP_CONTAINS(args[i], next)) {
                    continue; // Item not in all arguments
               }
            }
            set_add(MP_OBJ_FROM_PTR(out), next);
```

This should perform roughly the same as the existing code with one arg (due to the set lookup), and multiple args should be faster than calling the existing function multiple times (as the binary "contains" can be faster if the arguments are large sets).

AmirHmZz:
I've modified the `set_intersect_int()` and avoided linear search too. This might not reduce the performance of the existing hash-based approach.

AmirHmZz:
I've also set minimum accepted arg count to 1 in order to make `set.intersection()` compatibe with CPython in that case.

py/objset.c
```diff
@@ -227,47 +227,47 @@ static mp_obj_t set_diff_update(size_t n_args, const mp_obj_t *args) {
...
+                    set_add(MP_OBJ_FROM_PTR(out_ptr), next);
+                }
+            }
 
-    mp_obj_t iter = mp_getiter(other, NULL);
-    mp_obj_t next;
-    while ((next = mp_iternext(iter)) != MP_OBJ_STOP_ITERATION) {
-        if (mp_set_lookup(&self->set, next, MP_MAP_LOOKUP)) {
-            set_add(MP_OBJ_FROM_PTR(out), next);
+            m_del(mp_obj_t, in_ptr->set.table, in_ptr->set.alloc);
```
projectgus, line 252: Do I understand correctly that in the `!update` case, the behaviour is:

1. Make a full copy of set `self`.
2. For each argument, allocate a new set and build it up to have the items in both `self` and the argument.
3. Delete `self`'s set data, and then swap in the new set.

This seems like a lot of temporary memory churn, especially in the simplest case where there's only one argument.

I know when you first asked me about this on Discord I said it'd be better to build up the new set by adding instead of copying it and then removing items, to limit the peak memory usage. However if this is the alternative then maybe it is better to copy the set and then remove items?

The only downsize I can see is that the final result table might be too big (i.e. started with many entries and then removed most). Currently `mp_set_rehash` (which re-allocates and rebuilds the map) only gets called on growth never when making the set smaller. Could still emulate this though, by checking at the end if the set has a lot of free space and copying it if it does...?

AmirHmZz: > This seems like a lot of temporary memory churn, especially in the simplest case where there's only one argument.

If this is the only problem, I can solve it easily without significant code size increase.

About the idea of creating a set and delete items, we should do some linear searches because we want to make sure all items of set `X` are present in `Iterable[i]` and If it's not, then we remove it from `X`. It would be much easier if all of `args` were sets.
IMHO we should choose between these options:
1. Focus on reducing memory footprints with cost of decreasing performance (Linear search)
2. Focus on increasing performance with cost of increasing temporary memory usages.

projectgus: If there's a simple way to avoid the unncessary copy in the single arg version, then that sounds good.

I don't understand how copying the set once and then removing items requires linear searches though. What about something like this:

```c
p_obj_set_t *out_ptr = MP_OBJ_TO_PTR(set_copy(self_in));
for (size_t i = 1; i < n_args; i++) {
        if (MP_OBJ_FROM_PTR(in_ptr) != args[i]) {
            mp_obj_t iter = mp_getiter(args[i], NULL);
            mp_obj_t next;
            while ((next = mp_iternext(iter)) != MP_OBJ_STOP_ITERATION) {
                 if (!mp_set_lookup(&in_ptr->set, next, MP_MAP_LOOKUP)) {
                    set_discard(MP_OBJ_FROM_PTR(out_ptr), next);
                }
       }
}
```

This is the same structure as the current PR, only with the inner `set_add` swapped to a `set_discard`. Because any item stays removed once you discard it, you only have to copy the set once no matter how many args are provided.

You have thought about this problem a lot more than me, though. I am probably missing an important detail.

AmirHmZz: Let me show you what is wrong with your solution with an example: If we call `set.intesection({1,2},{2})` then:

1. You're making a copy of `{1,2}.`
2. Then iterating over `{2}` and checking 2 is present in `{1,2}` (which is present, so you do nothing)
3. Finally, the result will be `{1,2}` which is not correct.

As I told you earlier, we should check all items of `self_in` are present in args[i] and remove those are not. That's why this approach leads us to linear search.

projectgus: :facepalm: Makes sense, thanks for the clear explanation.

projectgus (approved):
This looks good to me, the code size is similar to the current version and the temporary memory usage looks like it won't be significantly worse than calling the single argument version multiple times in a row.

AmirHmZz:
`If this case (having one arg) is the only problem, I can solve it easily without significant code size increase.`
It seems a little bit more code is needed to avoid unnecessary copy when `n_args == 2` than I expected, which may increase code size. So I think it's better to wait for @dpgeorge review.

dpgeorge:
> It seems a little bit more code is needed to avoid unnecessary copy when `n_args == 2` than I expected, which may increase code size

I think it's worth improving things so it doesn't need the unnecessary copy.  The new implementation should match the existing implementation (in terms of memory usage) for `update == false` and `n_args == 2`, ie something like `s.intersect({1, 2})`.

And fixing that will actual improve the general case for `update == false` and `n_args >= 2`, because it won't need to copy the input set at all.

Please try to do that, and we'll see how much firmware size changes.  It shouldn't be too bad.

tests/basics/set_intersection.py
```diff
@@ -1,7 +1,21 @@
 s = {1, 2, 3, 4}
 print(sorted(s))
+print(sorted(s.intersection()))
+print(sorted(s))
```
dpgeorge, line 4: Needs an additional test to ensure that `s.intersection()` is a copy of `s`.

dpgeorge:
@AmirHmZz do you have any interest/time to continue working on this PR?  If not, I suggest to close it.

AmirHmZz:
@dpgeorge I'll do my best to see what can I do. Just give me some time cause currently I'm busy with something but I'll be back in 4-5 days.

AmirHmZz:
@dpgeorge @projectgus I've moved "intersect a source set against one iterable" logic into a small helper, `set_intersect_one()` and I've used it both for the first pass (reading directly from `args[0]`, no copy) and subsequent passes (reading from the accumulated result), instead of duplicating the loop body inline.

projectgus (approved):
This looks good to me, thanks @AmirHmZz.

tests/basics/set_intersection.py
AmirHmZz: done.

dpgeorge: Actually, I think this test still needs to be added.

The point is, you need to test that `s.intersection()` is not exactly the same object as `s`, that it's an independent copy.

Eg `s.intersection() is not s`.

tests/basics/set_intersection.py
```diff
@@ -1,7 +1,22 @@
...
+print(s.intersection_update())
+print(sorted(s))
 print(sorted(s.intersection({1, 3})))
+print(sorted(s))
 print(sorted(s.intersection([3, 4])))
+print(sorted(s))
+print(sorted(s.intersection({1, 2, 3}, {1, 4, 5}, {1})))
+print(sorted(s))
+print(sorted(s.intersection([1, 3], [1, 3, 5])))
+print(sorted(s))
```
dpgeorge, line 15: I don't think you need any of these interspersed `print(sorted(s))` lines.  If `s` ever changes then subsequent lines below won't match CPython output.

AmirHmZz: Actually it depends on how `s` might change. Assume [L12](https://github.com/AmirHmZz/micropython/blob/19204678ff20fdbd6fd86e4be609b988bc09871d/tests/basics/set_intersection.py#L12) pops `2` from `s` for any reason, then `L14` won't feel any difference. Now they might seem unnecessary, but I don't know what's going to change later. BTW if you think they are redundant, I will delete them. Based on your perspective, I think lines 7,9,11,13,15 should be omitted, am I right?

tests/basics/set_intersection.py
AmirHmZz: I get it now.  done.

tests/basics/set_intersection.py
dpgeorge: OK, it's fine to leave them in.

dpgeorge (approved):

dpgeorge:
Thanks @AmirHmZz for updating!

PR #15819, opened 2024-09-09
unix/README: Fix typo in build dependencies.

AmirHmZz:
`libffi-dev` is written as `libbfi-dev` by mistake.


projectgus (approved):
Thanks @AmirHmZz, not sure how that typo snuck in!

I re-ran the fixed steps in a clean ubuntu 20.04 container, and confirmed I could build the unix port OK.

PR #9653, opened 2022-10-17
Adding Pin.toggle() to docs

millosolomillo:
`Pin.toggle()` works with the pyboard and the Raspberry Pi Pico (RP2040), so I am adding the method to the documentation.

Apparently it is not working with the ESP32 port #8789, so feel free to move it under this section:

> The following methods are not part of the core Pin API and only implemented on certain ports.

PR #15822, opened 2024-09-10
doc: Document machine.Pin.toggle() method.

### LINKED ISSUE #9652: docs: Pin.toggle() method is not documented
`Pin.toggle()` method works (at least for RP2040 platform) but it is not documented here: 

https://docs.micropython.org/en/latest/library/machine.Pin.html#methods

projectgus:
### Summary

Document the `pin.toggle()` method that's available on some ports.

This is an amended version of the commit from #9653. Closes #9652.

### DIFF
--- a/docs/library/machine.Pin.rst
+++ b/docs/library/machine.Pin.rst
@@ -238,6 +238,12 @@ The following methods are not part of the core Pin API and only implemented on c
 
    Availability: cc3200 port.
 
+.. method:: Pin.toggle()
+
+   Toggle output pin from "0" to "1" or vice-versa.
+
+   Availability: mimxrt, samd, rp2 ports.
+
 Constants
 ---------
 


projectgus:
Huh, the auto-generated github email address in the original commit failed CI. Have reset it...

docs/library/machine.Pin.rst
```diff
@@ -238,6 +238,12 @@ The following methods are not part of the core Pin API and only implemented on c
 
    Availability: cc3200 port.
 
+.. method:: Pin.toggle()
+
+   Toggle output pin from ``0`` to ``1`` or vice-versa.
```
dpgeorge, line 243: Docs above use `"0"` and `"1"`, so suggest using that here for consistency.

dpgeorge (approved):

PR #15832, opened 2024-09-11
RFC: make source code generation better at picking up source changes

stinos:
### Summary

This is a first attempt to reduce the number of build failures which originate from the build system not properly picking up certain changes, typcially while developing and switching between branches (also see https://github.com/orgs/micropython/discussions/15693 where it's mentioned this is not uncommon). It's RFC currently because:
- I'm not sure this fixes all possible problems, so if someone else knows of a case not covered here and easy to reproduce then I can look into that and hopefully come up with a fix
- the main fix is pretty drastic and changes behavior

About that last part: makeqstrdefs.py seems to have been designed such that one can run `makeqstrdefs split ...` multiple times followed by a single `makeqstrdefs cat ...`. As far as I'm aware this isn't used as such in the micropython source itself. But it is also not possible anymore now., because basically my change is that `split` now creates a file which serves as input to tell `cat` what to do, and that file is overwritten each time. I can make it such that `split` appends to that file instead so multiple calls are possible again, but that would require an extra build step because at one point the file obviously has to be deleted again else we'd just keep appending to it, and the original problem only worsens. Thoughts?

### Testing

I verified changing between branches where one has more moduleds/frozen content than the other now works as expected.

The only alternative is seeing a build error, doing `make clean` and seeing if that fixes it :)

### DIFF
--- a/docs/develop/qstr.rst
+++ b/docs/develop/qstr.rst
@@ -60,8 +60,8 @@ Processing happens in the following stages:
    means that ``qstr.i.last`` will only contain data from files that have
    changed since the last compile.
 
-2. ``qstr.split`` is an empty file created after running ``makeqstrdefs.py split``
-   on qstr.i.last. It's just used as a dependency to indicate that the step ran.
+2. ``qstr.split`` is a file created after running ``makeqstrdefs.py split``
+   on qstr.i.last. It's used as a dependency to indicate that the step ran.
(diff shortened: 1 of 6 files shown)


dpgeorge:
I think the overall goal of this PR is good.

I just haven't had time to review this in detail.

stinos:
No worries, but before an actual code review we should decide if the current behavior can be broken or not, see description, because that might change parts of the implementation again.

Issue #10432, opened 2023-01-06
ESP32: 64-bit integer in python variable parsed into 32-bit in dynamic module written in C


shuki25:
## Issue
I'm writing a chess engine wrapper which uses a 64-bit bitboard for board positions, so it is important that it works on ESP32.

While writing a native dynamic module in C, I ran into two issues: (1) It appears the code is not being executed properly on a 32-bit architecture (ESP32 for example) when using 64-bit integer; (2) generating a 64-bit mask using left/right shift operator `1ULL << n` compiles successfully on ESP32 platform but fails to link to a `.mpy` file. The module works as expected when running on 64-bit linux.

I ran into some issues passing a 64-bit integer to a native module function and passed values being parsed into 32-bit. However, ESP32 does support 64-bit integer as evident shown below.  Looks like the Micropython native module functions are not converting numbers correctly.

If i assign a large value (e.g. 0xf0c00c000030c0f0) to a python variable as integer, it is stored correctly and output correctly.  But if I pass that value into a function, `mp_obj_get_int()` converted it to a 32-bit integer instead of 64-bit integer. Even casting it to `uint64_t` doesn't do a thing.

## Example testing output:
```python
MPY: soft reboot
MicroPython 699477d12 on 2022-12-28; ESP32 module (spiram) with ESP32
Type "help()" for more information.
>>> import uchess
>>> uchess.test_64bit()
default hex value: 0xf0c00c000030c0f0
num bits: 64
Using left shift with '& 0x8000000000000000 bitwise' mask
1111000011000000000011000000000000000000001100001100000011110000
Correct expected bit output:
1111000011000000000011000000000000000000001100001100000011110000

Using right shift with '& 0x0000000000000001 bitwise' mask (expected output should be flipped from above)
0000111100000011000011000000000000000000001100000000001100001111
Correct expected bit output:
0000111100000011000011000000000000000000001100000000001100001111
>>> a = 0xf0c00c000030c0f0
>>> a
17347878958773879024
>>> hex(a)
'0xf0c00c000030c0f0'
>>> uchess.test_64bit(a)
parameter passed: 3195120
num bits: 64
Using left shift with '& 0x8000000000000000 bitwise' mask
0000000000000000000000000000000000000000001100001100000011110000

Using right shift with '& 0x0000000000000001 bitwise' mask (expected output should be flipped from above)
0000111100000011000011000000000000000000000000000000000000000000
>>> uchess.test_64bit_lshift()
[...]

PR #15838, opened 2024-09-12
natmod: Allow linking with static libraries

### LINKED ISSUE #5629: Software floating point unsupported in dynamic native .mpy modules
Anytime I want to do a division or comparison with floats, I get an undefined symbol error by the linker. How can I include them?

E.g. a division yields this error:
`LinkError: build/mylib.o: undefined symbol: __divsf3`

vshymanskyy:
### Summary

When building non-trivial native modules, the compiler runtime needs to be linked manually, which is a tedious process. This PR allows `mpy_ld` to automatically resolve dependencies and link with `libgcc.a` and `libm.a` (or other user-specified static libraries). It also improves reporting format of `multiple definitions` and `unresolved symbol` errors.

Also, the automatic process ensures that only the required object files from archives are being linked into the mpy file. When doing it manually, one can include unneeded obj files which will just inflate the binary.

Loading large `.a` files takes quite some time, so the implementation caches the parsing results to minimize the impact on the developer experience (and the planet 🌱).

`MICROPY_ARCH_CFLAGS` was added so it can be conveniently used in the `Makefile` (i.e. it can be used to cross-build third-party libs)

### Testing

This was developed as part of [`wasm2mpy`](https://github.com/vshymanskyy/wasm2mpy) initiative.
As a result, I was able to produce builds without including the runtime object files in wasm2mpy repository:
https://github.com/vshymanskyy/wasm2mpy/actions/runs/10834266784/job/30063030978

```
┌────────────────┬───────┬───────┬──────────┬──────────┬─────────────┬─────────────┬──────────┬─────────────┐
│                │ x86   │ x64   │ armv6m   │ armv7m   │ armv7emsp   │ armv7emdp   │ xtensa   │ xtensawin   │
├────────────────┼───────┼───────┼──────────┼──────────┼─────────────┼─────────────┼──────────┼─────────────┤
│ assemblyscript │ 🟢    │ 🟢    │ 🟢       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
│ cpp            │ 🟢    │ 🟢    │ 🟢       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
│ rust           │ 🟢    │ 🟢    │ 🟢       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
│ tinygo         │ 🟢    │ 🟢    │ 🟢       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
│ zig            │ 🟢    │ 🟢    │ 🟢       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
│ virgil         │ 🟢    │ 🟢    │ 🟢       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
│ wat            │ 🟢    │ 🟢    │ 🟢       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
│ coremark       │ 🟢    │ 🟢    │ 🟥       │ 🟢       │ 🟢          │ 🟢          │ 🟢       │ 🟢          │
└────────────────┴───────┴───────┴──────────┴──────────┴─────────────┴─────────────┴──────────┴─────────────┘
```

Note: **armv6m/coremark** build fails due to unrelated [reason1](https://github.com/micropython/micropython/pull/15812), [reason2](https://github.com/micropython/micropython/issues/14430#issuecomment-2326857677)

Also, `examples/natmod/features2` was updated to use this functionality so it is included in the regular MicroPython CI tests.

Also, this was successfully used by @agatti and @jonnor . See comments below.

### Motivation

- Fixes #5629
- #6186
- Removes [bloat and hacks from `emlearn`](https://github.com/emlearn/emlearn-micropython/pull/15/files)
- https://github.com/micropython/micropython/issues/10432
- https://github.com/micropython/micropython/issues/14430
- https://github.com/orgs/micropython/discussions/9640
- https://github.com/orgs/micropython/discussions/11730
- https://forums.raspberrypi.com/viewtopic.php?t=320367
- https://forum.micropython.org/search.php?keywords=%22undefined+symbol%22&terms=all&author=&sc=1&sf=all&sr=posts&sk=t&sd=d&st=0&ch=300&t=0&submit=Search

- Alternatively, one can do it manually, i.e. find and unpack relevant `.a`, resolve dependencies, include required object files in the build process
- `mpy_ld` gets a new (optional) dependency on [ar](https://pypi.org/project/ar) package.  
`ar` is not required unless you either set `LINK_RUNTIME=1`, or pass `-l` option to the `mpy_ld`.
- I've added some basic support for weak symbols, but it should be addressed via a separate PR.

### DIFF
--- a/.github/workflows/ports_esp32.yml
+++ b/.github/workflows/ports_esp32.yml
@@ -34,6 +34,7 @@ jobs:
       run: source tools/ci.sh && echo "IDF_VER=$IDF_VER" | tee "$GITHUB_OUTPUT"
 
     - name: Cached ESP-IDF install
+      if: 0  # Temporary: skip cache for PR verification
       id: cache_esp_idf
       uses: actions/cache@v4
       with:
--- a/examples/natmod/features2/Makefile
+++ b/examples/natmod/features2/Makefile
@@ -10,5 +10,8 @@ SRC = main.c prod.c test.py
 # Architecture to build for (x86, x64, armv7m, xtensa, xtensawin)
 ARCH = x64
 
+# Link with libm.a and libgcc.a from the toolchain
+LINK_RUNTIME = 1
+
 # Include to get the rules for compiling and linking the module
 include $(MPY_DIR)/py/dynruntime.mk
--- a/examples/natmod/features2/main.c
+++ b/examples/natmod/features2/main.c
@@ -1,5 +1,6 @@
 /* This example demonstrates the following features in a native module:
     - using floats
+    - calling math functions from libm.a
     - defining additional code in Python (see test.py)
     - have extra C code in a separate file (see prod.c)
 */
@@ -9,6 +10,9 @@
 
 // Include the header for auxiliary C code for this module
 #include "prod.h"
+
+// Include standard library header
+#include <math.h>
 
 // Automatically detect if this module should include double-precision code.
 // If double precision is supported by the target architecture then it can
@@ -40,6 +44,12 @@ static mp_obj_t add_d(mp_obj_t x, mp_obj_t y) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_2(add_d_obj, add_d);
 #endif
+
+// A function that uses libm
+static mp_obj_t call_round(mp_obj_t x) {
+    return mp_obj_new_float_from_f(roundf(mp_obj_get_float_to_f(x)));
+}
+static MP_DEFINE_CONST_FUN_OBJ_1(round_obj, call_round);
 
 // A function that computes the product of floats in an array.
 // This function uses the most general C argument interface, which is more difficult
@@ -74,6 +84,7 @@ mp_obj_t mpy_init(mp_obj_fun_bc_t *self, size_t n_args, size_t n_kw, mp_obj_t *a
     #if USE_DOUBLE
     mp_store_global(MP_QSTR_add_d, MP_OBJ_FROM_PTR(&add_d_obj));
     #endif
+    mp_store_global(MP_QSTR_round, MP_OBJ_FROM_PTR(&round_obj));
 
     // The productf function uses the most general C argument interface
     mp_store_global(MP_QSTR_productf, MP_DYNRUNTIME_MAKE_FUNCTION(productf));
--- a/py/dynruntime.mk
+++ b/py/dynruntime.mk
@@ -29,9 +29,10 @@ CFLAGS += -Wall -Werror -DNDEBUG
 CFLAGS += -DNO_QSTR
 CFLAGS += -DMICROPY_ENABLE_DYNRUNTIME
 CFLAGS += -DMP_CONFIGFILE='<$(CONFIG_H)>'
-CFLAGS += -fpic -fno-common
-CFLAGS += -U _FORTIFY_SOURCE # prevent use of __*_chk libc functions
-#CFLAGS += -fdata-sections -ffunction-sections
+
+CFLAGS_ARCH += -fpic -fno-common
+CFLAGS_ARCH += -U_FORTIFY_SOURCE # prevent use of __*_chk libc functions
+#CFLAGS_ARCH += -fdata-sections -ffunction-sections
 
 MPY_CROSS_FLAGS += -march=$(ARCH)
 
@@ -47,72 +48,71 @@ ifeq ($(ARCH),x86)
 
 # x86
 CROSS =
-CFLAGS += -m32 -fno-stack-protector
+CFLAGS_ARCH += -m32 -fno-stack-protector
 MICROPY_FLOAT_IMPL ?= double
 
 else ifeq ($(ARCH),x64)
 
 # x64
 CROSS =
-CFLAGS += -fno-stack-protector
+CFLAGS_ARCH += -fno-stack-protector
 MICROPY_FLOAT_IMPL ?= double
 
 else ifeq ($(ARCH),armv6m)
 
 # thumb
 CROSS = arm-none-eabi-
-CFLAGS += -mthumb -mcpu=cortex-m0
+CFLAGS_ARCH += -mthumb -mcpu=cortex-m0
 MICROPY_FLOAT_IMPL ?= none
 
 else ifeq ($(ARCH),armv7m)
 
 # thumb
 CROSS = arm-none-eabi-
-CFLAGS += -mthumb -mcpu=cortex-m3
+CFLAGS_ARCH += -mthumb -mcpu=cortex-m3
 MICROPY_FLOAT_IMPL ?= none
 
 else ifeq ($(ARCH),armv7emsp)
 
 # thumb
 CROSS = arm-none-eabi-
-CFLAGS += -mthumb -mcpu=cortex-m4
-CFLAGS += -mfpu=fpv4-sp-d16 -mfloat-abi=hard
+CFLAGS_ARCH += -mthumb -mcpu=cortex-m4
+CFLAGS_ARCH += -mfpu=fpv4-sp-d16 -mfloat-abi=hard
 MICROPY_FLOAT_IMPL ?= float
 
 else ifeq ($(ARCH),armv7emdp)
 
 # thumb
 CROSS = arm-none-eabi-
-CFLAGS += -mthumb -mcpu=cortex-m7
-CFLAGS += -mfpu=fpv5-d16 -mfloat-abi=hard
+CFLAGS_ARCH += -mthumb -mcpu=cortex-m7
+CFLAGS_ARCH += -mfpu=fpv5-d16 -mfloat-abi=hard
 MICROPY_FLOAT_IMPL ?= double
 
 else ifeq ($(ARCH),xtensa)
 
 # xtensa
 CROSS = xtensa-lx106-elf-
-CFLAGS += -mforce-l32
+CFLAGS_ARCH += -mforce-l32
 MICROPY_FLOAT_IMPL ?= none
 
 else ifeq ($(ARCH),xtensawin)
 
 # xtensawin
 CROSS = xtensa-esp32-elf-
-CFLAGS +=
 MICROPY_FLOAT_IMPL ?= float
 
 else ifeq ($(ARCH),rv32imc)
 
 # rv32imc
 CROSS = riscv64-unknown-elf-
-CFLAGS += -march=rv32imac -mabi=ilp32 -mno-relax
+CFLAGS_ARCH += -march=rv32imac -mabi=ilp32 -mno-relax
 # If Picolibc is available then select it explicitly.  Ubuntu 22.04 ships its
 # bare metal RISC-V toolchain with Picolibc rather than Newlib, and the default
 # is "nosys" so a value must be provided.  To avoid having per-distro
 # workarounds, always select Picolibc if available.
 PICOLIBC_SPECS = $(shell $(CROSS)gcc --print-file-name=picolibc.specs)
 ifneq ($(PICOLIBC_SPECS),picolibc.specs)
-CFLAGS += --specs=$(PICOLIBC_SPECS)
+CFLAGS_ARCH += --specs=$(PICOLIBC_SPECS)
 endif
 
 MICROPY_FLOAT_IMPL ?= none
@@ -122,7 +122,13 @@ $(error architecture '$(ARCH)' not supported)
 endif
 
 MICROPY_FLOAT_IMPL_UPPER = $(shell echo $(MICROPY_FLOAT_IMPL) | tr '[:lower:]' '[:upper:]')
-CFLAGS += -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_$(MICROPY_FLOAT_IMPL_UPPER)
+CFLAGS += $(CFLAGS_ARCH) -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_$(MICROPY_FLOAT_IMPL_UPPER)
+
+ifeq ($(LINK_RUNTIME),1)
+LIBGCC_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-libgcc-file-name))
+LIBM_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-file-name=libm.a))
+MPY_LD_FLAGS += $(addprefix -l, $(LIBGCC_PATH) $(LIBM_PATH))
+endif
 
 CFLAGS += $(CFLAGS_EXTRA)
 
@@ -165,7 +171,7 @@ $(BUILD)/%.mpy: %.py
 # Build native .mpy from object files
 $(BUILD)/$(MOD).native.mpy: $(SRC_O)
 	$(ECHO) "LINK $<"
-	$(Q)$(MPY_LD) --arch $(ARCH) --qstrs $(CONFIG_H) -o $@ $^
+	$(Q)$(MPY_LD) --arch $(ARCH) --qstrs $(CONFIG_H) $(MPY_LD_FLAGS) -o $@ $^
 
 # Build final .mpy from all intermediate .mpy files
 $(MOD).mpy: $(BUILD)/$(MOD).native.mpy $(SRC_MPY)
--- a/tools/ar_util.py
+++ b/tools/ar_util.py
@@ -1,0 +1,230 @@
+#!/usr/bin/env python3
+#
+# This file is part of the MicroPython project, http://micropython.org/
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2024 Volodymyr Shymanskyy
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import os
+import re
+import hashlib
+import functools
+import pickle
+
+from elftools.elf import elffile
+from collections import defaultdict
+
+try:
+    from ar import Archive
+except:
+    Archive = None
+
+
+class PickleCache:
+    def __init__(self, path, prefix=""):
+        self.path = path
+        self._get_fn = lambda key: os.path.join(path, prefix + key[:24])
+
+    def store(self, key, data):
+        os.makedirs(self.path, exist_ok=True)
+        with open(self._get_fn(key), "wb") as f:
+            pickle.dump(data, f)
+
+    def load(self, key):
+        with open(self._get_fn(key), "rb") as f:
+            return pickle.load(f)
+
+
+def cached(key, cache):
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+            cache_key = key(*args, **kwargs)
+            try:
+                d = cache.load(cache_key)
+                if d["key"] != cache_key:
+                    raise Exception("Cache key mismatch")
+                return d["data"]
+            except Exception:
+                res = func(*args, **kwargs)
+                try:
+                    cache.store(
+                        cache_key,
+                        {
+                            "key": cache_key,
+                            "data": res,
+                        },
+                    )
+                except Exception:
+                    pass
+                return res
+
+        return wrapper
+
+    return decorator
+
+
+class CachedArFile:
+    def __init__(self, fn):
+        if not Archive:
+            raise RuntimeError("Please run 'pip install ar' to link .a files")
+        self.fn = fn
+        self._archive = Archive(open(fn, "rb"))
+        info = self.load_symbols()
+        self.objs = info["objs"]
+        self.symbols = info["symbols"]
+
+    def open(self, obj):
+        return self._archive.open(obj, "rb")
+
+    def _cache_key(self):
+        sha = hashlib.sha256()
+        with open(self.fn, "rb") as f:
+            for chunk in iter(lambda: f.read(4096), b""):
+                sha.update(chunk)
+        # Change this salt if the cache data format changes
+        sha.update(bytes.fromhex("45155db4bc868fa78cb99c3448b2bf2b"))
+        return sha.hexdigest()
+
+    @cached(key=_cache_key, cache=PickleCache(path=".mpy_ld_cache", prefix="ar_"))
+    def load_symbols(self):
+        print("Loading", self.fn)
+        objs = defaultdict(lambda: {"def": set(), "undef": set(), "weak": set()})
+        symbols = {}
+        for entry in self._archive:
+            obj_name = entry.name
+            elf = elffile.ELFFile(self.open(obj_name))
+            symtab = elf.get_section_by_name(".symtab")
+            if not symtab:
+                continue
+
+            obj = objs[obj_name]
+
+            for symbol in symtab.iter_symbols():
+                sym_name = symbol.name
+                sym_bind = symbol["st_info"]["bind"]
+
+                if sym_bind in ("STB_GLOBAL", "STB_WEAK"):
+                    if symbol.entry["st_shndx"] != "SHN_UNDEF":
+                        obj["def"].add(sym_name)
+                        symbols[sym_name] = obj_name
+                    else:
+                        obj["undef"].add(sym_name)
+
+                    if sym_bind == "STB_WEAK":
+                        obj["weak"].add(sym_name)
+
+        return {"objs": dict(objs), "symbols": symbols}
+
+
+def resolve(archives, symbols):
+    resolved_objs = []  # Object files needed to resolve symbols
+    unresolved_symbols = set()
+    provided_symbols = {}  # Which symbol is provided by which object
+    symbol_stack = list(symbols)
+
+    # A helper function to handle symbol resolution from a particular object
+    def add_obj(archive, symbol):
+        obj_name = archive.symbols[symbol]
+        obj_info = archive.objs[obj_name]
+
+        obj_tuple = (archive, obj_name)
+        if obj_tuple in resolved_objs:
+            return  # Already processed this object
+
+        resolved_objs.append(obj_tuple)
+
+        # Add the symbols this object defines
+        for defined_symbol in obj_info["def"]:
+            if defined_symbol in provided_symbols and not defined_symbol.startswith(
+                "__x86.get_pc_thunk."
+            ):
+                if defined_symbol in obj_info["weak"]:
+                    continue
+                else:
+                    raise RuntimeError(f"Multiple definitions for {defined_symbol}")
+            provided_symbols[defined_symbol] = obj_name  # TODO: mark weak if needed
+
+        # Recursively add undefined symbols from this object
+        for undef_symbol in obj_info["undef"]:
+            if undef_symbol in obj_info["weak"]:
+                print(f"Skippping weak dependency: {undef_symbol}")
+                continue
+            if undef_symbol not in provided_symbols:
+                symbol_stack.append(undef_symbol)  # Add undefined symbol to resolve
+
+    while symbol_stack:
+        symbol = symbol_stack.pop(0)
+
+        if symbol in provided_symbols:
+            continue  # Symbol is already resolved
+
+        found = False
+        for archive in archives:
+            if symbol in archive.symbols:
+                add_obj(archive, symbol)
+                found = True
+                break
+
+        if not found:
+            unresolved_symbols.add(symbol)
+
+    return resolved_objs, list(unresolved_symbols)
+
+
+def expand_ld_script(fn):
+    # This function parses a subset of ld scripts
+    # Typically these are just groups of static lib references
+    group_pattern = re.compile(r"GROUP\s*\(\s*([^\)]+)\s*\)", re.MULTILINE)
+    output_format_pattern = re.compile(r"OUTPUT_FORMAT\s*\(\s*([^\)]+)\s*\)", re.MULTILINE)
+    comment_pattern = re.compile(r"/\*.*?\*/", re.MULTILINE | re.DOTALL)
+
+    with open(fn, "r") as f:
+        content = f.read()
+        content = comment_pattern.sub("", content).strip()
+
+    # Ensure no unrecognized instructions
+    leftovers = content
+    for pattern in (group_pattern, output_format_pattern):
+        leftovers = pattern.sub("", leftovers)
+    if leftovers.strip():
+        raise ValueError("Invalid instruction found in the ld script:" + leftovers)
+
+    # Extract files from GROUP instructions
+    files = []
+    for match in group_pattern.findall(content):
+        files.extend([file.strip() for file in re.split(r"[,\s]+", match) if file.strip()])
+
+    return files
+
+
+def load_archive(fn):
+    # Woah, some .a files are actually ld scripts!
+    def is_ld_script():
+        with open(fn, "rb") as f:
+            header = b"/* GNU ld script"
+            return header == f.read(len(header))
+
+    if is_ld_script():
+        return [CachedArFile(item) for item in expand_ld_script(fn)]
+    else:
+        return [CachedArFile(fn)]
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -140,12 +140,15 @@ IDF_VER=v5.2.2
 export IDF_CCACHE_ENABLE=1
 
 function ci_esp32_idf_setup {
-    pip3 install pyelftools
     git clone --depth 1 --branch $IDF_VER https://github.com/espressif/esp-idf.git
     # doing a treeless clone isn't quite as good as --shallow-submodules, but it
     # is smaller than full clones and works when the submodule commit isn't a head.
     git -C esp-idf submodule update --init --recursive --filter=tree:0
     ./esp-idf/install.sh
+    # Install additional packages into the IDF env
+    source esp-idf/export.sh
+    pip3 install pyelftools
+    pip3 install ar
 }
 
 function ci_esp32_build_common {
@@ -362,6 +365,7 @@ function ci_samd_build {
 function ci_stm32_setup {
     ci_gcc_arm_setup
     pip3 install pyelftools
+    pip3 install ar
     pip3 install pyhy
 }
 
@@ -529,6 +533,7 @@ function ci_unix_standard_v2_run_tests {
 function ci_unix_coverage_setup {
     sudo pip3 install setuptools
     sudo pip3 install pyelftools
+    sudo pip3 install ar
     gcc --version
     python3 --version
 }
@@ -577,6 +582,7 @@ function ci_unix_32bit_setup {
     sudo apt-get install gcc-multilib g++-multilib libffi-dev:i386
     sudo pip3 install setuptools
     sudo pip3 install pyelftools
+    sudo pip3 install ar
     gcc --version
     python3 --version
 }
--- a/tools/mpy_ld.py
+++ b/tools/mpy_ld.py
@@ -30,6 +30,7 @@ Link .o files to .mpy
 
 import sys, os, struct, re
 from elftools.elf import elffile
+import ar_util
 
 sys.path.append(os.path.dirname(__file__) + "/../py")
 import makeqstrdata as qstrutil
@@ -664,7 +665,7 @@ def do_relocation_text(env, text_addr, r):
         R_XTENSA_PDIFF32,
         R_XTENSA_ASM_EXPAND,
     ):
-        if s.section.name.startswith(".text"):
+        if not hasattr(s, "section") or s.section.name.startswith(".text"):
             # it looks like R_XTENSA_[P]DIFF32 into .text is already correctly relocated,
             # and expand relaxations cannot occur in non-executable sections.
             return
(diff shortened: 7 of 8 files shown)


agatti:
@vshymanskyy nice! On RV32 it's able to pull softfp symbols from libgcc when building `examples/natmod/random`, and the resulting mpy file seems to work (I didn't test all the exposed functions yet).

A couple of things worth mentioning...  Is a new pip dependency really needed?  The archive format isn't that complex to write a reader for, as you already know, and the dependency in question isn't so much code to require a separate installation step.

Regarding collecting all duplicate symbol messages and print them all at once, (warning: personal preference) it might be useful whilst developing the feature but I'm not sure how much use a final user would get out of it.

vshymanskyy:
@agatti No, I really don't want to write yet another AR parser. The format is not actually simple, it's tricky to get right as seen from https://github.com/vidstige/ar/pull/17 and similar pull requests on that repo (and alternatives, which don't work due to the same reasons).

Collecting unresolved symbol messages and print them all at once provides much better insight into the scope of work and leads to much better decisions, which is why all "big" linkers do this (llvm, gcc, even tinycc afaik)

agatti:
@vshymanskyy for the symbol messages, I meant showing them all to the user, my bad.  Collecting is one thing, but displaying the whole lot may be a bit too much for the final user, especially if there are a lot of them - it's still an exception being raised so there's possibly quite a bit of text to read (traceback plus entries).  Again, personal preference here.

I'm still a bit torn about having an extra dependency.  I mean, it does work, don't get me wrong!  It's just that it looks like a small dependency compared to pyelftools, that's it.  But it's not up to me to decide :)

Still, this just unlocked one more natmod working on RV32, thanks!

vshymanskyy:
It turns out, that some `.a` and `.so` files are actually LD scripts, so I've also updated this PR to handle this.
Bumped into it on `x64`.

jonnor:
@vshymanskyy This looks like a super contribution to the natmod functionality. In addition to the mentioned issue, I think this would also close https://github.com/micropython/micropython/issues/5629 

I am not a reviewer for the MicroPython project, but I took a quick look at the code, and the integration into mpy_ld looks very clean. I also like that it is enabled by default without needing special configuration, so that it will just-work for native module writers.

It might be useful to now use floating point in one of the existing modules (or a new one) - such that this functionality gets tested automatically by CI - and prevent breakage in the future.
A couple of lines to the natmod documentation could also be useful. For example, to explain that on targets without an FPU, floating point library code will be copied into the native module. And that such code will not be shared between multiple modules. I have found the code size to be very unproblematic in practice, but it is good to inform people - so they for example can compare/contrast with user C modules (where only a single copy of these symbols will exist in an entire firmware image).

vshymanskyy:
Thanks for comments, I think these are all very relevant

jonnor:
I noticed now that it was opt-in. That is probably a smart choice, since there is an added Python dependency (ar). And that actually depends on Python 3.11, which I believe is more recent than the minimum version required by the MicroPython tooling per now.

I tested this briefly in emlfft module in https://github.com/emlearn/emlearn-micropython (where before I had a lot of manual .a files). It built fine on all platforms, and I could delete all the dirty hacks! Hope to test more soon :)

vshymanskyy:
I don't think `ar` depends on `Python 3.11`, as it's [verified for versions back to 3.7](https://github.com/vidstige/ar/actions/runs/10828329931).
However, I didn't test the code of my PR for such compatibility. I'll start by adding some tests to the CI

jonnor:
I do not know the details of ar. But on the default Python version on ubuntu-latest Docker image I got the following exception. The function file_digest was added to hashlib in Python 3.11, and updating to Python 3.11 resolved the issue.

```
  File "/home/runner/work/emlearn-micropython/emlearn-micropython/micropython/tools/mpy_ld.py", line 1060, in do_link
    archives.extend(ar_util.load_archive(item))
  File "/home/runner/work/emlearn-micropython/emlearn-micropython/micropython/tools/ar_util.py", line 221, in load_archive
    return [CachedArFile(fn)]
  File "/home/runner/work/emlearn-micropython/emlearn-micropython/micropython/tools/ar_util.py", line 92, in __init__
    info = self.load_symbols()
  File "/home/runner/work/emlearn-micropython/emlearn-micropython/micropython/tools/ar_util.py", line 61, in wrapper
    cache_key = key(*args, **kwargs)
  File "/home/runner/work/emlearn-micropython/emlearn-micropython/micropython/tools/ar_util.py", line 101, in _cache_key
    digest = hashlib.file_digest(f, "sha3_256")
AttributeError: module 'hashlib' has no attribute 'file_digest'
```

vshymanskyy:
@jonnor yes, that was used in my code. But it's already fixed in the latest version of this PR

jonnor:
@vshymanskyy thanks, can confirm that the lastest works with Python 3.10. It also builds all the modules in emlearn-micropython (5 different DSP/ML algorithms) on armv7m/armv7emsp/xtensawin. However *one* of the modules fails to find one symbol (out of many) on armv6m. See build log attached below

```
(venv) [jon@jon-thinkpad emlearn-micropython]$ make tinymaix_cnn.results ARCH=armv6m
make -C src/tinymaix_cnn/ ARCH=armv6m MPY_DIR=/home/jon/projects/micropython  V=1 clean dist
make[1]: Entering directory '/home/jon/projects/emlearn-micropython/src/tinymaix_cnn'
/bin/rm -rf build 
/bin/mkdir -p build/
GEN build/tinymaix_cnn.config.h
python3 /home/jon/projects/micropython/tools/mpy_ld.py '-vvv' --arch armv6m --preprocess -o build/tinymaix_cnn.config.h mod_cnn.c
CC mod_cnn.c
arm-none-eabi-gcc -I. -I/home/jon/projects/micropython -std=c99 -Os -Wall -Werror -DNDEBUG -DNO_QSTR -DMICROPY_ENABLE_DYNRUNTIME -DMP_CONFIGFILE='<build/tinymaix_cnn.config.h>' -fpic -fno-common -U _FORTIFY_SOURCE  -mthumb -mcpu=cortex-m0 -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_NONE  -I../../dependencies/TinyMaix/include -I../../dependencies/TinyMaix/src -Wno-error=unused-variable -Wno-error=multichar -o build/mod_cnn.o -c mod_cnn.c
In file included from mod_cnn.c:9:
../../dependencies/TinyMaix/src/tm_layers.c: In function 'tml_conv2d_dwconv2d':
../../dependencies/TinyMaix/src/tm_layers.c:159:26: warning: unused variable 's_step' [-Wunused-variable]
  159 |                 uint32_t s_step = (_ky1-_ky0)*(_kx1-_kx0);
      |                          ^~~~~~
In file included from mod_cnn.c:10:
../../dependencies/TinyMaix/src/tm_model.c: In function 'tm_load':
../../dependencies/TinyMaix/src/tm_model.c:20:26: warning: multi-character character constant [-Wmultichar]
   20 |     if(mdl_bin->magic != TM_MDL_MAGIC)   return TM_ERR_MAGIC;   //FIXME: big-endian not compatible
      |                          ^~~~~~~~~~~~
../../dependencies/TinyMaix/src/tm_model.c: In function 'tm_run':
../../dependencies/TinyMaix/src/tm_model.c:112:24: warning: unused variable 'l' [-Wunused-variable]
  112 |             tml_gap_t* l = (tml_gap_t*)(mdl->layer_body);
      |                        ^
../../dependencies/TinyMaix/src/tm_model.c:121:28: warning: unused variable 'l' [-Wunused-variable]
  121 |             tml_softmax_t* l = (tml_softmax_t*)(mdl->layer_body);
      |                            ^
../../dependencies/TinyMaix/src/tm_model.c:125:28: warning: unused variable 'l' [-Wunused-variable]
  125 |             tml_reshape_t* l = (tml_reshape_t*)(mdl->layer_body);
      |                            ^
LINK build/mod_cnn.o
python3 /home/jon/projects/micropython/tools/mpy_ld.py '-vvv' --arch armv6m --qstrs build/tinymaix_cnn.config.h -l/usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a -l/usr/arm-none-eabi/lib/thumb/v6-m/nofp/libm.a -o build/tinymaix_cnn.native.mpy build/mod_cnn.o
qstr vals: __del__, new, run, tinymaixcnn
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:mulsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:truncdfsf2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:extendsfdf2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:floatsisf.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_fixunssfsi.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:subdf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:fixsfsi.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:addsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_thumb1_case_uqi.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_divsi3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_arm_cmpsf2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:divsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:subsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_clzsi2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_dvmd_tls.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:eqsf2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:lesf2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:gesf2.o
LinkError: /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_divsi3.o: undefined symbol: __aeabi_idiv0
make[1]: *** [/home/jon/projects/micropython/py/dynruntime.mk:155: build/tinymaix_cnn.native.mpy] Error 1
make[1]: Leaving directory '/home/jon/projects/emlearn-micropython/src/tinymaix_cnn'
make: *** [Makefile:26: dist/armv6m_6.3/tinymaix_cnn.mpy] Error 2
```

vshymanskyy:
@jonnor please try the latest version. These symbols are weak and `mpy_ld` never really handled them.
I added some basic support for weak symbols, but it should be addressed via a separate PR.

vshymanskyy:
The documentation and CI tests are added.

One test fails: `esp32 port / build_idf (esp32_build_cmod_spiram_s2)`
This is due to `Cached ESP-IDF install`. Once cache is dropped, all should work as expected.

jonnor:
@vshymanskyy I tested your latest commit now (git SHA 02d50609d78becb9a410358baf9268312ffe0d42). The module above that failed to build now builds. However, at least one of the other modules now fails with a duplicate symbol:

```
(venv) [jon@jon-thinkpad emlearn-micropython]$ make emltrees.results ARCH=armv6m
make -C src/emltrees/ ARCH=armv6m MPY_DIR=/home/jon/projects/micropython  V=1 clean dist
make[1]: Entering directory '/home/jon/projects/emlearn-micropython/src/emltrees'
/bin/rm -rf build 
/bin/mkdir -p build/
GEN build/emltrees.config.h
python3 /home/jon/projects/micropython/tools/mpy_ld.py '-vvv' --arch armv6m --preprocess -o build/emltrees.config.h trees.c trees.py
CC trees.c
arm-none-eabi-gcc -I. -I/home/jon/projects/micropython -std=c99 -Os -Wall -Werror -DNDEBUG -DNO_QSTR -DMICROPY_ENABLE_DYNRUNTIME -DMP_CONFIGFILE='<build/emltrees.config.h>' -fpic -fno-common -U _FORTIFY_SOURCE  -mthumb -mcpu=cortex-m0 -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_NONE  -I/home/jon/projects/emlearn-micropython/venv/lib/python3.12/site-packages/emlearn -o build/trees.o -c trees.c
^[[ALINK build/trees.o
python3 /home/jon/projects/micropython/tools/mpy_ld.py '-vvv' --arch armv6m --qstrs build/emltrees.config.h -l/usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a -l/usr/arm-none-eabi/lib/thumb/v6-m/nofp/libm.a -o build/emltrees.native.mpy build/trees.o
qstr vals: __del__, addleaf, addnode, addroot, emltrees, new, predict, setdata
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:addsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:floatsisf.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:divsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_udivsi3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_arm_cmpsf2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_clzsi2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_thumb1_case_uqi.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_dvmd_tls.o
LinkError: duplicate symbol: __aeabi_idiv0
make[1]: *** [/home/jon/projects/micropython/py/dynruntime.mk:155: build/emltrees.native.mpy] Error 1
make[1]: Leaving directory '/home/jon/projects/emlearn-micropython/src/emltrees'
make: *** [Makefile:14: dist/armv6m_6.3/emltrees.mpy] Error 2
(venv) [jon@jon-thinkpad emlearn-micropython]$ make emltrees.results ARCH=armv6m
make -C src/emltrees/ ARCH=armv6m MPY_DIR=/home/jon/projects/micropython  V=1 clean dist
make[1]: Entering directory '/home/jon/projects/emlearn-micropython/src/emltrees'
/bin/rm -rf build 
/bin/mkdir -p build/
GEN build/emltrees.config.h
python3 /home/jon/projects/micropython/tools/mpy_ld.py '-vvv' --arch armv6m --preprocess -o build/emltrees.config.h trees.c trees.py
CC trees.c
arm-none-eabi-gcc -I. -I/home/jon/projects/micropython -std=c99 -Os -Wall -Werror -DNDEBUG -DNO_QSTR -DMICROPY_ENABLE_DYNRUNTIME -DMP_CONFIGFILE='<build/emltrees.config.h>' -fpic -fno-common -U _FORTIFY_SOURCE  -mthumb -mcpu=cortex-m0 -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_NONE  -I/home/jon/projects/emlearn-micropython/venv/lib/python3.12/site-packages/emlearn -o build/trees.o -c trees.c
LINK build/trees.o
python3 /home/jon/projects/micropython/tools/mpy_ld.py '-vvv' --arch armv6m --qstrs build/emltrees.config.h -l/usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a -l/usr/arm-none-eabi/lib/thumb/v6-m/nofp/libm.a -o build/emltrees.native.mpy build/trees.o
qstr vals: __del__, addleaf, addnode, addroot, emltrees, new, predict, setdata
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_udivsi3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_arm_cmpsf2.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:divsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:addsf3.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:floatsisf.o
using /usr/lib/gcc/arm-none-eabi/14.1.0/thumb/v6-m/nofp/libgcc.a:_dvmd_tls.o
LinkError: duplicate symbol: __aeabi_idiv0
make[1]: *** [/home/jon/projects/micropython/py/dynruntime.mk:155: build/emltrees.native.mpy] Error 1
make[1]: Leaving directory '/home/jon/projects/emlearn-micropython/src/emltrees'
make: *** [Makefile:14: dist/armv6m_6.3/emltrees.mpy] Error 2
```

vshymanskyy:
As I mentioned, mpy_ld never actually handled weak symbols so this is expected

vshymanskyy:
As I mentioned, mpy_ld never actually handled weak symbols so this is expected.
Most probably, your other modules have these symbols defined explicitly. Could you check if that's the case and remove them?

vshymanskyy:
@jonnor this is no longer needed:
https://github.com/search?q=repo%3Aemlearn%2Femlearn-micropython%20__aeabi_idiv0&type=code

jonnor:
@vshymanskyy ah right you are. Thanks, I had forgotten about those hacks. Removing them makes everything build succesfully!
For reference, this testing is ongoing in this branch/MR, https://github.com/emlearn/emlearn-micropython/pull/15
Looking forward to merging it master - as soon as this linking support is in MicroPython master :D

vshymanskyy:
@projectgus could you please resolving the `Cached ESP-IDF install` issue https://github.com/micropython/micropython/pull/15838#issuecomment-2350962956
And maybe assigning someone to review this?

vshymanskyy:
I have temporarily removed the IDF installation cache, to verify the PR.
**All checks have passed**

vshymanskyy:
Force-pushed a dummy change just to get `All checks have passed`.

jonnor:
@dpgeorge will you consider this for MicroPython 1.24 ? It fits in very well with the other work/bugfixes to make native modules work in practice.

Chris-F5:
This is an awesome feature. I'm having a problem though.
Steps to reproduce:

1. Clone your branch at `78bde7e0a8`
2. Navigate into `examples/natmod/features2`
3. Modify the Makefile to set `ARCH=armv7m` or `ARCH=armv7m` (both arent working for me).
4. Go to the `call_round` function in`main.c` and swap out the `round` function call with a `sqrtf`
5. Run `make`

Then make outputs this error (which means nothing to me!)
```
LINK build/main.o
LinkError: /usr/arm-none-eabi/lib/thumb/v6-m/nofp/libm.a:lib_a-s_lib_ver.o: .data.__fdlib_version non-empty
make: *** [../../../py/dynruntime.mk:155: build/features2.native.mpy] Error 1
```

Here is my output of `make -n` after a clean.
```
/bin/mkdir -p build/
echo "GEN build/features2.config.h"
python3 ../../../tools/mpy_ld.py --arch armv6m --preprocess -o build/features2.config.h main.c prod.c test.py
echo "CC main.c"
arm-none-eabi-gcc -I. -I../../.. -std=c99 -Os -Wall -Werror -DNDEBUG -DNO_QSTR -DMICROPY_ENABLE_DYNRUNTIME -DMP_CONFIGFILE='<build/features2.config.h>' -fpic -fno-common -U _FORTIFY_SOURCE  -mthumb -mcpu=cortex-m0 -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_NONE  -o build/main.o -c main.c
echo "CC prod.c"
arm-none-eabi-gcc -I. -I../../.. -std=c99 -Os -Wall -Werror -DNDEBUG -DNO_QSTR -DMICROPY_ENABLE_DYNRUNTIME -DMP_CONFIGFILE='<build/features2.config.h>' -fpic -fno-common -U _FORTIFY_SOURCE  -mthumb -mcpu=cortex-m0 -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_NONE  -o build/prod.o -c prod.c
echo "LINK build/main.o"
python3 ../../../tools/mpy_ld.py --arch armv6m --qstrs build/features2.config.h -l/usr/lib/gcc/arm-none-eabi/9.3.0/thumb/v6-m/nofp/libgcc.a -l/usr/arm-none-eabi/lib/thumb/v6-m/nofp/libm.a -o build/features2.native.mpy build/main.o build/prod.o
echo "MPY test.py"
../../../mpy-cross/build/mpy-cross -march=armv6m -o build/test.mpy test.py
echo "GEN features2.mpy"
python3 ../../../tools/mpy-tool.py --merge -o features2.mpy build/features2.native.mpy build/test.mpy
```

vshymanskyy:
This happens because the `lib_a-s_lib_ver.o` has a non-empty data section, and currently `mpy_ld` doesn't support this (as stated in the documentation). These limitations still apply to static libraries.

This not an issue with this PR, as this is exactly what would happen if you try linking all the required object files manually.

.github/workflows/ports_esp32.yml
```diff
@@ -34,6 +34,7 @@ jobs:
       run: source tools/ci.sh && echo "IDF_VER=$IDF_VER" | tee "$GITHUB_OUTPUT"
 
     - name: Cached ESP-IDF install
+      if: 0  # Temporary: skip cache for PR verification
```
dpgeorge, line 37: Does this need to be permanently disabled to get CI to work, or can this line be removed now?

examples/natmod/features2/Makefile
```diff
@@ -10,5 +10,8 @@ SRC = main.c prod.c test.py
 # Architecture to build for (x86, x64, armv7m, xtensa, xtensawin)
 ARCH = x64
 
+# Link with libm.a and libgcc.a from the toolchain
+MICROPY_LINK_RUNTIME = 1
```
dpgeorge, line 14: The other Make variables like `ARCH` are simple, short names, so how about calling this just `LINK_RUNTIME` to match them?

py/dynruntime.mk
```diff
@@ -29,9 +29,10 @@ CFLAGS += -Wall -Werror -DNDEBUG
 CFLAGS += -DNO_QSTR
 CFLAGS += -DMICROPY_ENABLE_DYNRUNTIME
 CFLAGS += -DMP_CONFIGFILE='<$(CONFIG_H)>'
-CFLAGS += -fpic -fno-common
-CFLAGS += -U _FORTIFY_SOURCE # prevent use of __*_chk libc functions
-#CFLAGS += -fdata-sections -ffunction-sections
+
+MICROPY_ARCH_CFLAGS += -fpic -fno-common
+MICROPY_ARCH_CFLAGS += -U_FORTIFY_SOURCE # prevent use of __*_chk libc functions
```
dpgeorge, line 34: I'm unsure why these options need to go under a different variable name?  And if they do, how about calling it `CFLAGS_ARCH`?

py/dynruntime.mk
```diff
@@ -45,66 +46,71 @@ ifeq ($(ARCH),x86)
...
 endif
 
 MICROPY_FLOAT_IMPL_UPPER = $(shell echo $(MICROPY_FLOAT_IMPL) | tr '[:lower:]' '[:upper:]')
-CFLAGS += -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_$(MICROPY_FLOAT_IMPL_UPPER)
+CFLAGS += $(MICROPY_ARCH_CFLAGS) -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_$(MICROPY_FLOAT_IMPL_UPPER)
+
+ifeq ($(MICROPY_LINK_RUNTIME),1)
+  LIBGCC_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-libgcc-file-name))
+  LIBM_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-file-name=libm.a))
+  MPY_LD_FLAGS += $(addprefix -l, $(LIBGCC_PATH) $(LIBM_PATH))
```
dpgeorge, line 112: Please de-dent these lines to match the other if blocks in this file.

tools/ar_util.py
```diff
@@ -0,0 +1,268 @@
...
+from collections import defaultdict
+
+try:
+    from ar import Archive
+except:
+    Archive = None
+
+
+class PickleCache:
+    def __init__(self, path=".cache", prefix=""):
```
dpgeorge, line 43: I suggest calling the cache path `.mpy_ld_cache`, so that it's clearer what program generated it and when it can safely be deleted.

tools/ar_util.py
```diff
@@ -0,0 +1,268 @@
...
+            header = b"/* GNU ld script"
+            return header == f.read(len(header))
+
+    if is_ld_script():
+        return [CachedArFile(item) for item in expand_ld_script(fn)]
+    else:
+        return [CachedArFile(fn)]
+
+
+if __name__ == "__main__":
```
dpgeorge, line 233: Is this script ever used as a stand-alone CLI program?  If not then it might be best to remove this code, at the very least because it's untested by CI and can easily get broken with changes in the future.

tools/mpy_ld.py
```diff
@@ -584,7 +585,7 @@ def do_relocation_text(env, text_addr, r):
         R_XTENSA_PDIFF32,
         R_XTENSA_ASM_EXPAND,
     ):
-        if s.section.name.startswith(".text"):
+        if not hasattr(s, "section") or s.section.name.startswith(".text"):
```
dpgeorge, line 588: Was this change necessary?

examples/natmod/features2/main.c
```diff
@@ -41,6 +45,12 @@ static mp_obj_t add_d(mp_obj_t x, mp_obj_t y) {
 static MP_DEFINE_CONST_FUN_OBJ_2(add_d_obj, add_d);
 #endif
 
+// A function that uses libm
+static mp_obj_t call_round(mp_obj_t x) {
+    return mp_obj_new_float_from_f(round(mp_obj_get_float_to_f(x)));
```
dpgeorge, line 50: `round` is a double function, so I suggest using `roundf`.

dpgeorge (commented):
Thanks @vshymanskyy for this, it's a really great enhancement.  I tested `features2` with `armv7emsp` and it works well.

py/dynruntime.mk
vshymanskyy: I've renamed it to `CFLAGS_ARCH`. In some cases, it is convenient to be able to use it in the natmod `Makefile`.
I remember it was handy during my experiments.

.github/workflows/ports_esp32.yml
vshymanskyy: This change was needed to actually run the PR test, due to the way `ci_esp32_idf_setup` works.
I think we can remove it just before merging this PR.

tools/mpy_ld.py
```diff
@@ -1075,8 +1076,8 @@ def process_riscv32_relocation(env, text_addr, r):
     return addr, value
 
 
-def load_object_file(env, felf):
-    with open(felf, "rb") as f:
+def load_object_file(env, f, felf):
+    if 1:  # Temporary, to preserve indent
```
vshymanskyy, line 1080: @dpgeorge should I de-indent it?

dpgeorge: Yes please.

tools/mpy_ld.py
vshymanskyy: I think it was breaking some of the builds in [wasm2mpy](https://github.com/vshymanskyy/wasm2mpy).
However I do not recall exactly what it was. I need to revive some work on this and test with and without this change.

py/dynruntime.mk
vshymanskyy: Real-life use case: https://github.com/vshymanskyy/lbm-micropython/blob/a106761c6b43e4f8e64010cc7a6eccd113c1e847/Makefile#L52

.github/workflows/ports_esp32.yml
vshymanskyy: Clarification: It may be needed to manually purge the cache **once** after this PR is merged.

tools/mpy_ld.py
vshymanskyy: Yes it does fix some linking issues with ESP8266. 
@dpgeorge I realize this is out of scope for this PR, should I revert this change?

dpgeorge: OK, we can keep this change in this commit.

py/dynruntime.mk
dpgeorge: OK, I see why it's useful to have these options in a separate variable.

So let's keep it as you have it now, `CFLAGS_ARCH`.

dpgeorge (approved):

.github/workflows/ports_esp32.yml
dpgeorge: This PR is ready to be merged, so what do we do with this line?  Can you just remove this now and push to the branch, then I can merge it? And then possibly follow up with a cache purge?

tools/ar_util.py
```diff
@@ -0,0 +1,230 @@
...
+    def open(self, obj):
+        return self._archive.open(obj, "rb")
+
+    def _cache_key(self):
+        sha = hashlib.sha256()
+        with open(self.fn, "rb") as f:
+            for chunk in iter(lambda: f.read(4096), b""):
+                sha.update(chunk)
+        # Change this salt if the cache data format changes
+        sha.update(bytes.fromhex("45155db4bc868fa78cb99c3448b2bf2b"))
```
projectgus, line 105: This is something of a nitpick, but can't this be a cache version number like `sha.update(b"1")` rather than a random hex string?

tools/ar_util.py
```diff
@@ -0,0 +1,230 @@
...
+        files.extend([file.strip() for file in re.split(r"[,\s]+", match) if file.strip()])
+
+    return files
+
+
+def load_archive(fn):
+    # Woah, some .a files are actually ld scripts!
+    def is_ld_script():
+        with open(fn, "rb") as f:
+            header = b"/* GNU ld script"
```
projectgus, line 224: I'm a little worried about whether every .a ld script will start with this comment.

It looks like GNU ld assumes anything which isn't in a known BFD format is a linker script:
https://github.com/bminor/binutils-gdb/blob/bf5e5514bd841f4b4837435c46c67862ba3bdfe5/ld/ldfile.c#L407

(although it then goes on to try and parse the linker script, and changes its mind if it can't be parsed... :grimacing: )

Would it be reasonable to invert the logic: check for the [AR magic bytes](https://github.com/bminor/binutils-gdb/blob/bf5e5514bd841f4b4837435c46c67862ba3bdfe5/include/aout/ar.h#L30) and assume anything which doesn't have these is a linker script?

tools/ci.sh
```diff
@@ -140,12 +140,15 @@ IDF_VER=v5.2.2
 export IDF_CCACHE_ENABLE=1
 
 function ci_esp32_idf_setup {
-    pip3 install pyelftools
     git clone --depth 1 --branch $IDF_VER https://github.com/espressif/esp-idf.git
     # doing a treeless clone isn't quite as good as --shallow-submodules, but it
     # is smaller than full clones and works when the submodule commit isn't a head.
     git -C esp-idf submodule update --init --recursive --filter=tree:0
     ./esp-idf/install.sh
+    # Install additional packages into the IDF env
```
projectgus, line 148:
```suggestion
    # Install additional packages for mpy_ld into the IDF env
```

projectgus (approved):
This is really impressive and useful, @vshymanskyy!

I have some minor inline comments, but as far as I'm concerned none of them are blockers for merging this as-is.

agatti:
You may also want to delete this fragment of `tools/ci.sh` as part of the PR since soft-float is supposed to be working out of the box for that specific native module.

https://github.com/micropython/micropython/blob/e44a2c6921919d0d19356a70a23bf21cd301856c/tools/ci.sh#L496-L499

jonnor:
@projectgus @dpgeorge I believe @vshymanskyy said on Discord that he is rather preoccupied at the moment (the situation in Ukraine remains difficult). So I am sure he would appreciate it if anyone can help out with the small things remaining to get this unblocked and merged :)

agatti:
If nobody else has time for this I can put the finishing touch to this PR to let it go through, the hard part is already done anyway.

tools/mpy_ld.py
```diff
@@ -1465,6 +1490,7 @@ def main():
     cmd_parser.add_argument("--arch", default="x64", help="architecture")
     cmd_parser.add_argument("--preprocess", action="store_true", help="preprocess source files")
     cmd_parser.add_argument("--qstrs", default=None, help="file defining additional qstrs")
+    cmd_parser.add_argument("-l", dest="libs", action="append", help="Static .a libraries to link")
```
dpgeorge, line 1493: Suggest adding a long form name `--libs` or `--lib` for this (as well as keeping `-l`).

And lowercase "static .a ..." to match the help of other commands.

dpgeorge:
@agatti yes please, if you could pick this up that would be great!

I guess you'll need to make a new PR because you won't have write access to this branch.

.github/workflows/ports_esp32.yml
projectgus: I think what we need is: 
* Remove this line
* Merge PR
* Delete IDF caches manually

If this happens a lot then we can make the caching more specific (i.e. use the hash of ci.sh as the cache key, that's a common pattern) - but if it's very occasional then probably not worth the extra churn.

tools/ar_util.py
agatti: I guess `b"1"` doesn't have enough bits set to make much of a difference in this case.  If we want to go for the version number then I guess it should be embedded in the resulting cache file.

dpgeorge: Maybe use something like `b"000000000000000000000000001"`?

agatti:
Right, I have a version that incorporates most of the changes requested in this PR - however the code changed a bit since the first time I tried and whilst this fully works on ARMv7m and normal soft-float stuff works on RV32 (adding `MICROPY_FLOAT_IMPL=float` to the command line), linking in functions from libm does not work on Picolibc (`features2` was changed to reference `roundf` after I first tried it).

On Picolibc there's no real libm file to speak of, as the floating point code is in `libc.a` which is somewhere else.  Usually this is not a problem, as the specs file add the necessary includes to the GCC command line, but path resolution with `--print-file-name` doesn't seem to take the specs's extra paths into account.  So, for example, `$(CROSS)gcc -specs=<picolibc.specs path here> --print-file-name=libc.a` won't find anything worthwhile even though normal compilation and linking works otherwise.

So, I can see two options for this:

* either I keep working on this to make it work on Picolibc, delaying this even further
* I submit this with the new changes, making this official only for ARMv7m, while Picolibc support is added at a later date.

I'd go for the former option, so I can also look at #16042 if there aren't any takers for that one as well, but I need some input on this to better fit within 1.25.0's release schedule.

agatti:
> I'm pretty sure more than just ARMv7m is supported by this PR. According to the top post, there are many architectures supported. But that's with newlib, of course.

Indeed they're all newlib-based.  The only other platform that requires soft-float as far as I know is xtensa as far as I can know, but I'm looking at CI targets right now and natmods aren't built for it.

> I suggest getting this PR across the line first, then adding Picolibc support as a follow up. That could all still fit within the 1.25.0 release. (I don't think the amount of work would be that different if you added Picolibc support together with this PR, or as a follow up, right?)

The amount of work would be the same, but I do have a chunk of contiguous free time available now so I'd rather finish this off in one go if I can.  For other things and follow up PRs, I may or may not have the time for them in the next few weeks or so (although w.r.t. inline LX6/LX7 assembler I might have a business case for it so it's different).  If I happen to not have the time I don't want to keep you folks waiting :)

> The release is at least one week away, so you have that much time, at least.
>
> Note that this PR is not critical for the release and can be left until after the release if needed.

Fair enough, I'll see what I can do then.  I'll clean up newlib support and submit a PR so there's at least something that can be merged in the meantime.

vshymanskyy:
@agatti I'm pretty sure that `-specs` DOES have impact on `--print-file-name`.

agatti:
> @agatti I'm pretty sure that `-specs` DOES have impact on `--print-file-name`.

Glad you're here :)  Anyway, I'm quite sure this doesn't seem to happen on the RV32 compiler version for Ubuntu 22.04, the same running in CI jobs - unless I'm missing something that is.  Here's what's going on my environment:

```bash
[/micropython/examples/natmod/features2] $ riscv64-unknown-elf-gcc --version
riscv64-unknown-elf-gcc () 10.2.0

[/micropython/examples/natmod/features2] $ riscv64-unknown-elf-gcc -march=rv32imac -mabi=ilp32 --print-file-name=picolibc.specs
/usr/lib/gcc/riscv64-unknown-elf/10.2.0/picolibc.specs

[/micropython/examples/natmod/features2] $ riscv64-unknown-elf-gcc -specs=/usr/lib/gcc/riscv64-unknown-elf/10.2.0/picolibc.specs -march=rv32imac -mabi=ilp32 --print-file-name=libgcc.a
/usr/lib/gcc/riscv64-unknown-elf/10.2.0/rv32imac/ilp32/libgcc.a

[/micropython/examples/natmod/features2] $ riscv64-unknown-elf-gcc -specs=/usr/lib/gcc/riscv64-unknown-elf/10.2.0/picolibc.specs -march=rv32imac -mabi=ilp32 --print-file-name=libm.a  
libm.a

[/micropython/examples/natmod/features2] $ riscv64-unknown-elf-gcc -specs=/usr/lib/gcc/riscv64-unknown-elf/10.2.0/picolibc.specs -march=rv32imac -mabi=ilp32 --print-file-name=libc.a  
libc.a

[/micropython/examples/natmod/features2] $ cat /usr/lib/gcc/riscv64-unknown-elf/10.2.0/picolibc.specs
[...]
*link:
[...] -L%{-picolibc-prefix=*:%*/lib/picolibc/riscv64-unknown-elf/lib/%M;:/usr/lib/picolibc/riscv64-unknown-elf/lib/%M} [...]

[/micropython/examples/natmod/features2] $ ll /usr/lib/picolibc/riscv64-unknown-elf/lib/rv32imac/ilp32 
[...]
-rw-r--r-- 1 root root 9644250 Nov 19  2021 libc.a
[...]
-rw-r--r-- 1 root root    1444 Nov 19  2021 libm.a
```

I only have a picolibc-based environment on a Ubuntu 22.04 VM - I'll create a different environment to see if this is still the case elsewhere.

My changes to `py/dynruntime.mk` are as such, by the way:

```
diff --git i/py/dynruntime.mk w/py/dynruntime.mk
index 5592db5fa..95ddbfad1 100644
--- i/py/dynruntime.mk
+++ w/py/dynruntime.mk
@@ -112,7 +112,8 @@ CFLAGS_ARCH += -march=rv32imac -mabi=ilp32 -mno-relax
 # workarounds, always select Picolibc if available.
 PICOLIBC_SPECS = $(shell $(CROSS)gcc --print-file-name=picolibc.specs)
 ifneq ($(PICOLIBC_SPECS),picolibc.specs)
-CFLAGS_ARCH += --specs=$(PICOLIBC_SPECS)
+CFLAGS_ARCH += -specs=$(PICOLIBC_SPECS)
+USE_PICOLIBC := 1
 endif
 
 MICROPY_FLOAT_IMPL ?= none
@@ -125,8 +126,13 @@ MICROPY_FLOAT_IMPL_UPPER = $(shell echo $(MICROPY_FLOAT_IMPL) | tr '[:lower:]' '
 CFLAGS += $(CFLAGS_ARCH) -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_$(MICROPY_FLOAT_IMPL_UPPER)
 
 ifeq ($(LINK_RUNTIME),1)
+ifeq ($(USE_PICOLIBC),1)
+LIBM_NAME := libc.a
+else
+LIBM_NAME := libm.a
+endif
 LIBGCC_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-libgcc-file-name))
-LIBM_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-file-name=libm.a))
+LIBM_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-file-name=$(LIBM_NAME)))
 MPY_LD_FLAGS += $(addprefix -l, $(LIBGCC_PATH) $(LIBM_PATH))
 endif
```

tools/ar_util.py
projectgus: > I guess `b"1"` doesn't have enough bits set to make much of a difference in this case

Actually this isn't the case, a vital property of a secure hash function is collision resistance via  the avalanche effect. Even a 1-bit difference in the input will cause a significant difference in the output.

So a single byte is more than enough (the hash implementation may pad it out internally to fill a block, but we don't need to try and second-guess this - the hash designers have done it for us).

agatti:
OK, unless I manage to get the specs-driven path resolution working in the CI environment (and if somebody can help...) I can at least get things to compile with this:

```diff
diff --git a/py/dynruntime.mk b/py/dynruntime.mk
index 5592db5fa..807befb46 100644
--- a/py/dynruntime.mk
+++ b/py/dynruntime.mk
@@ -110,9 +110,12 @@ CFLAGS_ARCH += -march=rv32imac -mabi=ilp32 -mno-relax
 # bare metal RISC-V toolchain with Picolibc rather than Newlib, and the default
 # is "nosys" so a value must be provided.  To avoid having per-distro
 # workarounds, always select Picolibc if available.
-PICOLIBC_SPECS = $(shell $(CROSS)gcc --print-file-name=picolibc.specs)
+PICOLIBC_SPECS := $(shell $(CROSS)gcc --print-file-name=picolibc.specs)
 ifneq ($(PICOLIBC_SPECS),picolibc.specs)
-CFLAGS_ARCH += --specs=$(PICOLIBC_SPECS)
+CFLAGS_ARCH += -specs=$(PICOLIBC_SPECS)
+USE_PICOLIBC := 1
+PICOLIBC_ARCH := rv32imac
+PICOLIBC_ABI := ilp32
 endif
 
 MICROPY_FLOAT_IMPL ?= none
@@ -125,8 +128,42 @@ MICROPY_FLOAT_IMPL_UPPER = $(shell echo $(MICROPY_FLOAT_IMPL) | tr '[:lower:]' '
 CFLAGS += $(CFLAGS_ARCH) -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_$(MICROPY_FLOAT_IMPL_UPPER)
 
 ifeq ($(LINK_RUNTIME),1)
+# All of these picolibc-specific directives are here to work around a
+# limitation of Ubuntu 22.04's RISC-V bare metal toolchain.  In short, the
+# specific version of GCC in use (10.2.0) does not seem to take into account
+# extra paths provided by an explicitly passed specs file when performing name
+# resolution via `--print-file-name`.
+#
+# If Picolibc is used and libc.a fails to resolve, then said file's path will
+# be computed by searching the Picolibc libraries root for a libc.a file in a
+# subdirectory whose path is built using the current `-march` and `-mabi`
+# flags that are passed to GCC.  The `PICOLIBC_ROOT` environment variable is
+# checked to override the starting point for the library file search, and if
+# it is not set then the default value is used, assuming that this is running
+# on an Ubuntu 22.04 machine.
+#
+# This should be revised when the CI base image is updated to a newer Ubuntu
+# version (that hopefully contains a newer RISC-V compiler) or to another Linux
+# distribution.
+ifeq ($(USE_PICOLIBC),1)
+LIBM_NAME := libc.a
+else
+LIBM_NAME := libm.a
+endif
 LIBGCC_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-libgcc-file-name))
-LIBM_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-file-name=libm.a))
+LIBM_PATH := $(realpath $(shell $(CROSS)gcc $(CFLAGS) --print-file-name=$(LIBM_NAME)))
+ifeq ($(USE_PICOLIBC),1)
+ifeq ($(LIBM_PATH),)
+# The CROSS toolchain prefix usually ends with a dash, but that may not be
+# always the case.  If the prefix ends with a dash it has to be taken out as
+# Picolibc's architecture directory won't have it in its name.  GNU Make does
+# not have any facility to perform character-level text manipulation so we
+# shell out to sed.
+CROSS_PREFIX := $(shell echo $(CROSS) | sed -e 's/-$$//')
+PICOLIBC_ROOT ?= /usr/lib/picolibc/$(CROSS_PREFIX)/lib
+LIBM_PATH := $(PICOLIBC_ROOT)/$(PICOLIBC_ARCH)/$(PICOLIBC_ABI)/$(LIBM_NAME)
+endif
+endif
 MPY_LD_FLAGS += $(addprefix -l, $(LIBGCC_PATH) $(LIBM_PATH))
 endif
```

Some assumptions are made, but they are at least documented.  Is this approach good enough for this situation?

If the path resolution issues I'm seeing depend on the compiler version then my workaround won't be used, and if this happens to be run on a system that uses a "broken" compiler but with Picolibc installed elsewhere this should still be recoverable somewhat - I assume the user *will* look at the Makefile to at least edit the hardcoded path, see the workaround, and act accordingly in that case.

tools/ar_util.py
agatti: Right, I haven't had the need to look into hash algorithm details and the like in almost two decades so I might be a bit rusty on the subject.  Anyway, I got nerd-sniped and I found this https://www.mdpi.com/2410-387X/8/3/40 so if I read the paper correctly, for SHA-256, the more bits set the better? :)

Anyway, I went with @dpgeorge's suggestion, it still encodes the same value with a few more bits set.  Nevertheless, thanks for giving me the opportunity to get re-acquainted with this stuff!

agatti:
Yes, the workaround kicks in only if `USE_PICOLIBC` is set to 1 from inside the makefile itself, and right now only the RV32 target does that.

tools/ar_util.py
projectgus: > for SHA-256, the more bits set the better? :)

We don't need a secure hash function which achieve a particular cryptographic strength level here, we just need one which won't trivially produce collisions. If you could produce collisions by flipping a single bit or adding a single byte to a SHA input then there'd be problems with things a lot more important than this. :grin: 

> Anyway, I went with @dpgeorge's suggestion, it still encodes the same value with a few more bits set.

No worries, it's a relatively minor point anyway - one way or another we just need to change this arbitrary string if the on-disk format ever changes.

dpgeorge:
Merged via 16863, see commit 51976110e2da32b7a5b7035b7d5f17ab49bbd40e

Thanks @vshymanskyy for this great addition, and @agatti for making the final steps!

dubiousjim:
Cool, I was just coming to this forum to ask into implementing this functionality. Glad that someone's already done the work of building it, and that it's already been reviewed and committed. Before I start experimenting with it, can I just ask for someone more familiar with the changes to say: is it known how this will work with musl? (I'm using Alpine Linux/armv7.)

agatti:
> is it known how this will work with musl? (I'm using Alpine Linux/armv7.)

You may need to modify `py/dynruntime.mk` to enable the same behaviour that is currently followed for Picolibc.  Picolibc needs to process `libc.a` for making math functions available, but maybe you can leverage that to let the linker pick stack protection symbols up instead - there is no limitation on which symbols are linked from external libraries, as far as I know.

PR #15843, opened 2024-09-12
tools/metrics.py: Add VIRT_RV32 to the code size metrics.

agatti:
### Summary

This commit adds the Qemu-based RISC-V 32 bits `VIRT_RV32` board to the list of ports/boards to be built for measuring code size changes, as per the discussion in https://github.com/micropython/micropython/pull/15743#issuecomment-2328024612.

### Testing

Ran the tool locally on a CI-like environment.

In the original discussion it was mentioned the potential for the metrics collection step to take too long.  The RV32 port in question brings in as little third party code as possible and it should be relatively quick to build, all things considered (a single full build takes around a minute on an underclocked i5 laptop).  However, until there are real measurements being taken in the final environment this is all speculation.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -55,11 +55,12 @@ function ci_code_size_setup {
     sudo apt-get install gcc-multilib
     gcc --version
     ci_gcc_arm_setup
+    ci_gcc_riscv_setup
 }
 
 function ci_code_size_build {
     # check the following ports for the change in their code size
-    PORTS_TO_CHECK=bmusxpd
+    PORTS_TO_CHECK=bmusxpdv
     SUBMODULES="lib/asf4 lib/berkeley-db-1.xx lib/btstack lib/cyw43-driver lib/lwip lib/mbedtls lib/micropython-lib lib/nxp_driver lib/pico-sdk lib/stm32lib lib/tinyusb"
 
     # starts off at either the ref/pull/N/merge FETCH_HEAD, or the current branch HEAD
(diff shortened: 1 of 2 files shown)


dpgeorge:
This won't actually be reported by the github-actions bot (see its comment immediately above) unless "v" is added to the `PORTS_TO_CHECK` list in `tools/ci.sh:ci_code_size_build`.

agatti:
Sorry, I thought I added the second commit when I force updated the branch last night.  The second commit is now part of the PR.

dpgeorge:
The CI for the code size check is failing, but I think that will only work once this is merged, so that it has the "v" option on master to check the PR against.

dpgeorge (approved):

dpgeorge:
@agatti are you sure this will work, won't it need to install a RISC-V toolchain in `ci_code_size_setup`?

agatti:
You're right, turned out I tested that with a VM that had already built the RV32 code once.  This should now be addressed.

dpgeorge:
> Wait, that runs on 20.04? Is it OK if I update the workflow to use 22.04 or is there a particular reason that workflow is still on that?

See commit 0a079155e46d588095e4419b1095fd4ae691ab50 ... something to do with gcc-multilib.

You're welcome to try updating that to 22.04, but it would need to be a separate PR because that CI step needs to be tested, and it won't be tested here because it fails due to missing "v" support.

agatti:
That should be fixed in #15884 - besides the RV64 and macOS failures (expected and occasional), the code size task seems to work just fine on 22.04.

dpgeorge (approved):

PR #15845, opened 2024-09-13
ports/rp2: Add support for MICROPY_PY_NETWORK_PPP_LWIP.

Gadgetoid:
Add the new PPP LWIP interface into RP2's mpconfigport.h as a first-class interface.

This joins `MICROPY_PY_NETWORK_NINAW10`, `MICROPY_PY_NETWORK_WIZNET5K` and `MICROPY_PY_NETWORK_CYW43` and ensures that any interface specific caveats - so far we just have a dependency on `MICROPY_PY_SOCKET_EXTENDED_STATE` - are dealt with.

Note: I've also included commit to add the socket state constants into the PPP module, since these are returned verbatim by `ppp.status()` but there's currently no means to disambiguate them. I figured I'd get away with this, but shout if you'd prefer a separate PR.

### Testing

So far this bringup has been tested downstream in my RP2350 builds.

TODO: Build and test a regular Pico (RP2040) firmware depending upon this change.


dpgeorge:
I must be missing something here... I thought `network.PPP` would be available automatically when `MICROPY_PY_NETWORK_PPP_LWIP` is enabled due to the following entry in `mp_module_network_globals_table`:
```c
    #if MICROPY_PY_NETWORK_PPP_LWIP                                         
    { MP_ROM_QSTR(MP_QSTR_PPP), MP_ROM_PTR(&mp_network_ppp_lwip_type) },        
    #endif                                                                  
```
I don't see why it needs to be explicitly added?

Also, why does the lwIP PPP driver need `MICROPY_PY_SOCKET_EXTENDED_STATE`?

Regarding the state constants: I'm happy to add these but:
1. The esp32 port's PPP driver may also want to add such constants, and we need to think if the ones here are general enough to extend to that port.
2. Existing network status/state constants actually start with `STAT_`.  Eg run `git grep MP_QSTR_STAT_` to see them all.  The PPP constants should try to match the naming pattern of those existing ones.

Gadgetoid:
D’oh! 

Looking into it, I think  “MICROPY_PY_SOCKET_EXTENDED_STATE” might be a holdout from my experimental Python sockets driver, as - perhaps - is the practise of manually adding to the list of port interfaces. Will test without and drop this commit.

Acknowledged regarding the constants.

dpgeorge:
I'll close this, because PPP can already be enabled, per comments above.

See also #16445.

PR #15851, opened 2024-09-16
extmod/vfs_semihosting: Add a VFS using semihosting for file I/O.

agatti:
### Summary

This contains a VFS implementation that uses semihosting as a file I/O call interface, allowing the code to access files located either in the host machine (if using QEMU), or in the machine where the debugger is running on (if using SWD/JTAG).  The implementation is meant to be only used to aid in debugging in very specific cases, and it is rather limited due to what the Arm/RISC-V semihosting specification offers.  The following features are not available: flush, poll, stat (both file and vfs), ilistdir, mkdir, and rmdir.

The Arm and RISC-V semihosting implementations have been merged into one single file with the same interface, with the only difference being the semihosting call entry point.  I've added the Arm semihosting call code to the RISC-V implementation because all the file I/O functions were already implemented there.  Semihosting support was also extended to AArch64 and RV64 targets.  The old platform-specific files have been left alone, not sure if they have any further use, but I can always update the PR to remove them.

This feature was originally written for a situation where it made sense to be available on real hardware, but then I realised it could help with ~~natmod testing, so I migrated this to run on QEMU.  For that, the import machinery may require some changes to accommodate for semihosting limitations.  The reason being that it seems there is no way to know if a path points to a file, to a directory, or to some other special entity.  With QEMU's semihosting implementation, opening a directory as if it was a file returns ENOENT (instead of EISDIR), the same as if the file wasn't there, but maybe it is enough for MPY files?~~ a subset of generic filesystem operations.

### Testing

VFS testing has been done manually on QEMU (VIRT_RV32, MPS2_AN385, and SABRELITE boards), as I'm not sure what is the proper way to add this to the CI pipeline though.  Regarding the rest of the semihosting calls, most of them were tested on RV32 when debugging the native RV32 emitter back then, the code is virtually the same.

On real boards, using `mpremote mount` can be an option, but it requires allocating a serial port to let it talk with the device's REPL - if a REPL is actually present to begin with.  With semihosting all serial ports are available for the board to use as the SWD/JTAG communication channel will be used for file I/O instead.

This has potential security implications if used incorrectly.  Since the usage of this feature is restricted to very specific cases, there may be issues regarding unexpected path traversals and I/O on files outside the local root.  It is disabled by default except for the QEMU port (and must be explicitly activated on the command line), as it can be helpful when testing.

### DIFF
--- a/extmod/extmod.cmake
+++ b/extmod/extmod.cmake
@@ -66,6 +66,7 @@ set(MICROPY_SOURCE_EXTMOD
     ${MICROPY_EXTMOD_DIR}/vfs_posix.c
     ${MICROPY_EXTMOD_DIR}/vfs_posix_file.c
     ${MICROPY_EXTMOD_DIR}/vfs_reader.c
+    ${MICROPY_EXTMOD_DIR}/vfs_semihosting.c
     ${MICROPY_EXTMOD_DIR}/virtpin.c
     ${MICROPY_EXTMOD_DIR}/nimble/modbluetooth_nimble.c
 )
--- a/extmod/extmod.mk
+++ b/extmod/extmod.mk
@@ -69,6 +69,7 @@ SRC_EXTMOD_C += \
 	extmod/vfs_posix.c \
 	extmod/vfs_posix_file.c \
 	extmod/vfs_reader.c \
+	extmod/vfs_semihosting.c \
 	extmod/virtpin.c \
 	shared/libc/abort_.c \
 	shared/libc/printf.c \
--- a/extmod/modvfs.c
+++ b/extmod/modvfs.c
@@ -33,6 +33,7 @@
 #include "extmod/vfs_lfs.h"
 #include "extmod/vfs_posix.h"
 #include "extmod/vfs_rom.h"
+#include "extmod/vfs_semihosting.h"
 
 #if !MICROPY_VFS
 #error "MICROPY_PY_VFS requires MICROPY_VFS"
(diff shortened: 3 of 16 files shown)


iabdalkader:
> using `mpremote mount` can be an option

`mpremote mount` doesn't work on qemu because stdio_poll, ticks, delay etc.. are not implemented. I was about to send a PR to add all of that to get `mpremote mount` working, then I saw this PR. I think this is a much better approach. My use case is: I need to use qemu in CI to run library unit-tests.
@dpgeorge Is this something that will be merged or should I go with `mpremote mount` approach?

agatti:
I'd say that for your specific use case #18234 is probably a better option.  

`mpremote`'s filesystem support is much more complete than what this can offer as it is limited by Arm's semihosting specifications: listdir(), stat(), and others are not implemented simply because they're not part of the specs.  I assume regular Python testing code may leverage more I/O functions than basic read/write/seek/mkdir, plus in your testing harness you probably have the REPL starting up at boot time anyway.

This was originally written to solve an issue I had on a specific set of devices where I had no REPL and ran out of serial ports with no bluetooth or network available.  It got retrofitted to QEMU for natmod testing, but ultimately it got superseded by changes in the testing framework.  This doesn't make this any less useful though, but it doesn't really fit your use case.

iabdalkader:
> `mpremote`'s filesystem support is much more complete than what this can offer as it is limited by Arm's semihosting specifications: listdir(), stat(), and others are not implemented

Yeah I know, I read your comments in `extmod/vfs_semihosting.c`. Either one works for me really, the unit tests only reads files stored on the filesystem, so open, read, close is all I need.

agatti:
Fair enough!  Merging both PRs could still be an option, though.  #18234 can be helpful to have around, in any case.

Still, I've updated the PR to also cover RV64 as that target arch wasn't part of the QEMU port back then, along with some really minor fixes.

PR #15854, opened 2024-09-16
ports/rp2: Include check for CYW43_GPIO in machine_pin_cyw43.c

ZodiusInfuser:
When doing some experiments with IO expander support for the Pimoroni Yukon, I discovered that the `#if` check in `machine_pin_cyw43.c` only checks that `MICROPY_PY_NETWORK_CYW43` and `MICROPY_HW_PIN_EXT_COUNT` are defined.

This is a reasonable assumption for the Pico W, but causes conflicts if someone (such as myself) wants to attach an external IO expander to their Pico W and have its pins appear as `Pin` objects.

This PR addresses this by adding the additional checks for `defined(CYW43_GPIO) && CYW43_GPIO == 1` to `machine_pin_cyw43.c` letting board builds include wireless but separately choose whether the external IO pins come from the cyw43 or not.

The long term solution of course would be to allow for multiple external pin types, but that feels like a bigger discussion, and is more than I currently require anyway.

### DIFF
--- a/ports/rp2/machine_pin_cyw43.c
+++ b/ports/rp2/machine_pin_cyw43.c
@@ -30,7 +30,7 @@
 #include "py/runtime.h"
 #include "py/mphal.h"
 
-#if defined(MICROPY_PY_NETWORK_CYW43) && defined(MICROPY_HW_PIN_EXT_COUNT)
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):

dpgeorge:
Thanks, rebased and merged in ded8bbdd5efc50d2b86e3020462ced8bbcf4d25c

PR #15861, opened 2024-09-17
esp32/i2c: Use idf function for calculating main timeout.

andrewleech:
### Summary

The esp32 I2C driver sets a global timeout during init. The functionality for this is currently only correct for original esp32 and esp32-s2, newer parts have a slightly different argument.

When searching through the IDF code for this driver I found this function that handles the chip-specific differences.

ESP32-S2
image

ESP32-S3
image

### Testing

This has been implemented from the reference manual & inspecting the IDF codebase, I haven't pysically tested it so far.

### DIFF
--- a/ports/esp32/machine_i2c.c
+++ b/ports/esp32/machine_i2c.c
@@ -257,7 +257,7 @@ static void machine_hw_i2c_init(machine_hw_i2c_obj_t *self, bool first_init) {
         .master.clk_speed = self->freq,
     };
     i2c_param_config(self->port, &conf);
-    int timeout = I2C_SCLK_FREQ / 1000000 * self->timeout_us;
+    int timeout = i2c_ll_calculate_timeout_us_to_reg_val(I2C_SCLK_FREQ, self->timeout_us);
     i2c_set_timeout(self->port, (timeout > I2C_LL_MAX_TIMEOUT) ? I2C_LL_MAX_TIMEOUT : timeout);
     i2c_driver_install(self->port, I2C_MODE_MASTER, 0, 0, 0);
 }


dpgeorge:
Thanks, this looks correct.  But would be good to test it on esp32 and esp32s3, eg a scope measuring SCL.

robert-hh:
> Thanks, this looks correct. But would be good to test it on esp32 and esp32s3, eg a scope measuring SCL.

I can do that, but struggling on where the timeout should get visible. Trying different actions, I see with the current master a timeout of 100ms - 300ms not related to the timeout setting in the constructor.

dpgeorge:
> I can do that, but struggling on where the timeout should get visible.

Hmm, I'm not sure.  Maybe you can try your test with another port that has timeout (eg rp2, stm32) and see how that behaves.

robert-hh:
I see a action now with the ESP32S3. The SCL timeout setting is recognized and the level is properly handled, BUT only for the first timeout event after setting up the I2C module. Further timeout events are not handled. So it seems that the timeout flag is not reset or not re-enabled.

Side note: At the Python API level this scl low/high timeout setting seems of little importance. When thinking about a slave mode, another timeout will be more important - the time the slave waits for an incoming message. Then, this scl timeout would better be renamed.

andrewleech:
Thanks so much for your testing @robert-hh I really appreciate it! 

Your findings are very interesting - both the inconsistency in the max value and the "auto clearing" of the timeout. I suspect many of the other newer parts will likely behave similar to S3 where there's these couple of glitches we really should address somehow.

On an important side-note though, the i2c API being used here is largely deprecated. IDF has a new "device / transfer centric" API we will presumably need to migrate to anyway. 

I made a brief start towards implementing it while looking at this, but didn't get far before thinking it'll likely need a bit broader architectural change to support.
I think this same timeout function is still used in the new API though perhaps it's set per-transfer now ... I don't remember for sure.

dpgeorge (approved):
This looks like a necessary change:
- `i2c_set_timeout()` is a thin layer around `i2c_ll_set_tout()`
- `i2c_ll_set_tout()` writes directly to the hardware register, and it needs to be in units calculated by `i2c_ll_calculate_timeout_us_to_reg_val()`

PR #15862, opened 2024-09-17
ports/esp32: Use capability defines to configure features.

andrewleech:
### Summary

Was broken out from https://github.com/micropython/micropython/pull/11869 

This updates esp32 code where appropriate to replace ifdefs based on a list of specific chips with a feature SOC_* definition, eg.
```
#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
```
to
```
#if SOC_TOUCH_SENSOR_SUPPORTED
```

### Testing

TBD - each of these changes will requires review and test to confirm no unexpected regressions.

This should simplify adding new esp32-* chips in future, deferring chip feature support to the IDF.


andrewleech:
Hi @projectgus the soc capability changes commit has been moved here and I think is ready for review / test again.
I've fixed your previous suggestions from the original MR.

ports/esp32/machine_adc.c
```diff
@@ -94,6 +94,14 @@ static const machine_adc_obj_t madc_obj[] = {
     {{&machine_adc_type}, ADCBLOCK1, ADC_CHANNEL_3, GPIO_NUM_3},
     {{&machine_adc_type}, ADCBLOCK1, ADC_CHANNEL_4, GPIO_NUM_4},
     {{&machine_adc_type}, ADCBLOCK2, ADC_CHANNEL_0, GPIO_NUM_5},
+    #elif CONFIG_IDF_TARGET_ESP32C6
```
andrewleech, line 97: Ah whoops this belongs in the C6 MR now, I'll move it...

andrewleech:
@projectgus I've rebased this now that C6 has merged, but haven't tested anything beyond it being able to compile for C6

ports/esp32/adc.c
```diff
@@ -33,30 +33,27 @@
...
-            break;
-        #endif
-        #if CONFIG_IDF_TARGET_ESP32S2
-        case 13:
-            self->width = ADC_WIDTH_BIT_13;
-            break;
-        #endif
-        default:
-            mp_raise_ValueError(MP_ERROR_TEXT("invalid bits"));
+    if (bits >= SOC_ADC_RTC_MIN_BITWIDTH && bits <= SOC_ADC_RTC_MIN_BITWIDTH) {
```
projectgus, line 36:
```suggestion
    if (bits >= SOC_ADC_RTC_MIN_BITWIDTH && bits <= SOC_ADC_RTC_MAX_BITWIDTH) {
```

ports/esp32/machine_pwm.c
```diff
@@ -241,10 +241,13 @@ static void set_freq(machine_pwm_obj_t *self, unsigned int freq, ledc_timer_conf
...
-        #if SOC_LEDC_SUPPORT_XTAL_CLOCK
-        timer->clk_cfg = LEDC_USE_XTAL_CLK;
-        #else
+
+        #if SOC_LEDC_SUPPORT_PLL_DIV_CLOCK
+        timer->clk_cfg = LEDC_USE_PLL_DIV_CLK;
+        #elif SOC_LEDC_SUPPORT_APB_CLOCK
         timer->clk_cfg = LEDC_USE_APB_CLK;
+        #elif SOC_LEDC_SUPPORT_XTAL_CLOCK
+        timer->clk_cfg = LEDC_USE_XTAL_CLK;
```
projectgus, line 250: This changes the behaviour for all chips except for ESP32, I think? but probably in a good way... as it breaks down as:

```
❯ rg -g soc_caps.h "SOC_LEDC_SUPPORT_.+_CLOCK" components/soc
components/soc/esp32h2/include/soc/soc_caps.h
267:#define SOC_LEDC_SUPPORT_PLL_DIV_CLOCK      (1)
268:#define SOC_LEDC_SUPPORT_XTAL_CLOCK         (1)

components/soc/esp32c2/include/soc/soc_caps.h
166:#define SOC_LEDC_SUPPORT_PLL_DIV_CLOCK  (1)
167:#define SOC_LEDC_SUPPORT_XTAL_CLOCK     (1)

components/soc/esp32/include/soc/soc_caps.h
231:#define SOC_LEDC_SUPPORT_APB_CLOCK       (1)

components/soc/esp32s2/include/soc/soc_caps.h
223:#define SOC_LEDC_SUPPORT_APB_CLOCK       (1)
225:#define SOC_LEDC_SUPPORT_XTAL_CLOCK      (1)

components/soc/esp32c3/include/soc/soc_caps.h
219:#define SOC_LEDC_SUPPORT_APB_CLOCK       (1)
220:#define SOC_LEDC_SUPPORT_XTAL_CLOCK      (1)

components/soc/esp32s3/include/soc/soc_caps.h
227:#define SOC_LEDC_SUPPORT_APB_CLOCK       (1)
228:#define SOC_LEDC_SUPPORT_XTAL_CLOCK      (1)

components/soc/esp32c6/include/soc/soc_caps.h
267:#define SOC_LEDC_SUPPORT_PLL_DIV_CLOCK      (1)
268:#define SOC_LEDC_SUPPORT_XTAL_CLOCK         (1)

components/soc/esp32p4/include/soc/soc_caps.h
274:#define SOC_LEDC_SUPPORT_PLL_DIV_CLOCK      (1)
275:#define SOC_LEDC_SUPPORT_XTAL_CLOCK         (1)
```

Which I read as:

* ESP32 only supports APB clock source.
* All chips apart from ESP32 support both XTAL clock and one of APB or PLL_DIV.

And:

* In current master all chips apart from ESP32 get the slower XTAL clock source (40MHz).
* With this PR they now get either APB or PLL_DIV, whichever is supported. (I believe 80MHz or 96MHz, respectively.)

However there is a bunch of (uncommented) code earlier in this function which looks wrong - i.e. it calculates the duty resolution, but it calculates it from the assumption the source clock is 80MHz. This is only true on ESP32 on current master, plus now true for some other chips in this PR.

ESP-IDF driver provides a function `ledc_find_suitable_duty_resolution`, and the function doc says to find the source clock frequency via `esp_clk_tree_src_get_freq_hz` - so probably all that earlier code can be replaced with calls to these two functions?

We can potentially leave this for a future PR, though. As I think this PR already makes the behaviour more  correct, not less correct...

ports/esp32/adc.c
projectgus: (I like this trick to keep the switch statement readable, though!)

projectgus (commented):
These changes look good, thanks @andrewleech. I have a couple of comments but it's pretty straightforward.

I can help test this PR on some of the other SoCs, although I haven't done this yet.

ports/esp32/adc.c
andrewleech: Oh drats, I've fixed that one already in the past. Must have accidentally reverted it in one of the rebase fixups.

andrewleech:
Thanks @projectgus I've fixed that adc issue and reworked the PWM over in https://github.com/micropython/micropython/pull/16090 as suggested.

ports/esp32/adc.c
projectgus: I thought it looked familiar. :grin:

projectgus:
@andrewleech When you have a moment to rebase this and drop the PWM change then it'd be great to get it merged.

andrewleech:
Ah yes, thanks for the reminder @projectgus I've rebased it now.

projectgus (commented):
This looks great, @andrewleech! I have one small suggestion but otherwise looks good.

I can help do some testing on various boards and then I think we should get this in, should streamline future ESP32 additions. :rocket:

ports/esp32/machine_adc_block.c
```diff
@@ -32,13 +32,8 @@
...
 machine_adc_block_obj_t madcblock_obj[] = {
-    #if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6 || CONFIG_IDF_TARGET_ESP32S3
-    {{&machine_adc_block_type}, ADC_UNIT_1, 12, -1, {0}},
-    {{&machine_adc_block_type}, ADC_UNIT_2, 12, -1, {0}},
-    #elif CONFIG_IDF_TARGET_ESP32S2
-    {{&machine_adc_block_type}, ADC_UNIT_1, 13, -1, {0}},
-    {{&machine_adc_block_type}, ADC_UNIT_2, 13, -1, {0}},
-    #endif
+    {{&machine_adc_block_type}, ADC_UNIT_1, SOC_ADC_RTC_MAX_BITWIDTH, -1, {0}},
+    {{&machine_adc_block_type}, ADC_UNIT_2, SOC_ADC_RTC_MAX_BITWIDTH, -1, {0}},
```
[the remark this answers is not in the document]
andrewleech: Oh good catch, I'm not sure how that one got missed considering most of my initial testing was on C6!

Hmm... ok it compiles on C6 even without this cleanup so the `ADC_UNIT_2` definition must exist, it just obviously wouldn't actually work. With the new #ifdef though the ram isn't wasted by the definition so well worth doing.

ports/esp32/modmachine.c
```diff
@@ -147,30 +147,34 @@ static void machine_sleep_helper(wake_type_t wake_type, size_t n_args, const mp_
         esp_sleep_enable_timer_wakeup(((uint64_t)expiry) * 1000);
     }
 
-    #if !(CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6)
-
+    #if SOC_PM_SUPPORT_EXT0_WAKEUP
```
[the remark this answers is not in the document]
andrewleech: Oh yeah that's a good point. Pin irq / sleep settings that do / don't work seem to come up in issues regularly so a concerted effort to clean up and test these would be valuable.  This arguably applies to all ports though, just just esp32...

ports/esp32/machine_bitstream.c
```diff
@@ -42,7 +42,7 @@
 // This is a translation of the cycle counter implementation in ports/stm32/machine_bitstream.c.
 static void IRAM_ATTR machine_bitstream_high_low_bitbang(mp_hal_pin_obj_t pin, uint32_t *timing_ns, const uint8_t *buf, size_t len) {
     uint32_t pin_mask, gpio_reg_set, gpio_reg_clear;
-    #if !CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6
+    #if SOC_GPIO_PIN_COUNT > 32
```
dpgeorge, line 45: looks like this logic was actual broken prior to this change, should have been `!(CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6)`

ports/esp32/machine_pin.c
```diff
@@ -43,7 +43,7 @@
 #include "modesp32.h"
 #include "genhdr/pins.h"
 
-#if CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6
+#if SOC_USB_SERIAL_JTAG_SUPPORTED
```
dpgeorge, line 46: Looks like ESP32S3 also defines `SOC_USB_SERIAL_JTAG_SUPPORTED`, but I guess that's OK, it'll just include the header and won't change any functionality.

ports/esp32/adc.c
```diff
@@ -33,30 +33,27 @@
...
+            case 11:
+                self->width = ADC_BITWIDTH_11;
+                break;
+            case 12:
+                self->width = ADC_BITWIDTH_12;
+                break;
+            case 13:
+                self->width = ADC_BITWIDTH_13;
+                break;
+            default:
```
dpgeorge, line 53: Should this default do anything, raise an exception, or just not exist?  Is it possible to be within the min and max but still not match one of the case statements?

ports/esp32/machine_bitstream.c
andrewleech: Yes I think so, something else I missed in the C6 PR I expect.

ports/esp32/adc.c
andrewleech: I guess even if there aren't any chips currently supporting bits outside of this range (I'm not sure if there is) the next one released might change this.
The gap in the logic should be filled to be safe.

andrewleech: I've reworked this slightly now. I think the logic is a bit clearer now too (with a new comment added)

dpgeorge: Looks good now!

ports/esp32/machine_pin.c
projectgus: Some people build S3 with USB Serial/JTAG peripheral instead of the native one, as well. (No supported boards do this, but we've had bug reports for it.)

projectgus (approved):
Great work @andrewleech! I think we should merge this to progress some of the other ESP32-related PRs that add new chips or tweak features.

(I ended up doing less testing than I anticipated, although did try out the ADC on the ESP32-C6 which seems to work correctly. Have re-read the other diff items and I don't see any potential for subtle bugs like the PWM clock source - the I2C clock sources are disjoint so that one shouldn't change anything.)

dpgeorge (approved):

PR #15863, opened 2024-09-17
runtime/exitcodes: Set exit code according to the SystemExit exception.

RealJohnSmith:
### Summary

This PR continues previous PRs, originally at https://github.com/micropython/micropython/pull/14419 and later https://github.com/micropython/micropython/pull/15392 and https://github.com/micropython/micropython/pull/15730

### Testing

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -986,6 +986,16 @@ typedef double mp_float_t;
 #define MICROPY_ENABLE_VM_ABORT (0)
 #endif
 
+// Whether to handle abort behavior in pyexec code
+#ifndef MICROPY_PYEXEC_ENABLE_VM_ABORT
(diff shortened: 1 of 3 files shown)


RealJohnSmith:
@dpgeorge can you please have a look? It is the same code as last time. Can we merge it already? Is there something that blocks it, that I should change? This drags along for half a year or so already...

RealJohnSmith:
Check commit message fails because of some other commit in the history not being signed off. This commit is OK and signed off

RealJohnSmith:
@dpgeorge, what do you think about this PR?

RealJohnSmith:
@dpgeorge ?

RealJohnSmith:
@dpgeorge Hi, do you think you could find some time to review this case? Thank you

dpgeorge:
Replaced by #18275.

Issue #15868, opened 2024-09-17
PIC16 ports does not compile (without modification)


SirTobias:
### Port, board and/or hardware

PIC16

### MicroPython version

Micropython master branch, Microchip xc16-gcc v2.10 for PIC16

### Reproduction

1. Check out master branch
2. Cross-compile, if needed
3. Go to ports/pic16, either make or make BOARD=dsPICSK leads to the same issue

### Observed behaviour

With micropython (master branch, PIC16 port), XC16-gcc v2.10, I get on an Ubuntu 22.04 LTS:

with either `make` or `make BOARD=dsPICSG`

`                                                                             
/opt/microchip/xc16/v2.10/bin/xc16-gcc  -DFFCONF_H=\"lib/oofatfs/ffconf.h\" -I. -I../.. -Ibuild -I/opt/microchip/xc16/v2.10/include -I/opt/microchip/xc16/v2.10/support/dsPIC33F/h -Wall -Werror -std=gnu99 -nostdl
ib -mcpu=33FJ256GP506 -mlarge-code  -O1 -DNDEBUG -c -MD -MF build/py/mpstate.d -o build/py/mpstate.o ../../py/mpstate.c || (echo -e "See \033[1;31mhttps://github.com/micropython/micropython/wiki/Build-Troublesho
oting\033[0m"; false)                                                                                                                                                                                              
In file included from ../../py/mpstate.h:33:0,                                                                                                                                                                     
                 from ../../py/mpstate.c:27:                                                               
../../py/misc.h: In function 'mp_clz_mpi':                                                                 
../../py/misc.h:383:5: error: size of unnamed array is negative ` 

When I comment out the assertion in py/misc.h:383, then it compiles. Does anybody have an idea?

Thank you very much in advance.

### Additional Information

Ubuntu 22.04 LTS, 64 bit

### Code of Conduct

Yes, I agree

agatti:
Weird, I've also had to deal with this recently.  This seems to be working for me:

```diff
diff --git a/ports/pic16bit/Makefile b/ports/pic16bit/Makefile
index 6d061514f..dda48fa3c 100644
--- a/ports/pic16bit/Makefile
+++ b/ports/pic16bit/Makefile
@@ -7,7 +7,7 @@ QSTR_DEFS = qstrdefsport.h
 include $(TOP)/py/py.mk
 include $(TOP)/extmod/extmod.mk
 
-XCVERSION ?= 1.35
+XCVERSION ?= 2.10
[...]

PR #15872, opened 2024-09-18
pic16bit: Make it build with recent XC16 versions.

agatti:
### Summary

The `pic16bit` port was left behind and didn't build anymore on recent compilers.  This PR contains the necessary modifications to make the code build on the currently latest version of xc16 (2.10).

The only code change needed is an implementation of `mp_clz_mpi`, which it requires its own implementation.  The build will fail with a missing reference to `__clzhi2` otherwise, and that symbol doesn't seem to be defined anywhere.

*I've been trying to repurpose a PIC24F board with a new firmware and since MicroPython seems to have a port for it, why not.  I understand that maybe the `pic16bit` port is still lingering around for historical reasons and that this PR is now out of scope for the project, but letting these changes build up dust on my machine isn't much better :)*

### Testing

This PR contains only the platform-generic changes I've made to the port to make it work on a PIC24F.  However, as I don't have a dsPIC33 target board to test this against, I can at least claim it now compiles out of the box without errors.

That said, the code builds and there is at least [one report](https://github.com/micropython/micropython/issues/15868#issuecomment-2358350157) of these changes working on a dsPIC33 in #15868.

### DIFF
--- a/ports/pic16bit/mpconfigport.h
+++ b/ports/pic16bit/mpconfigport.h
@@ -93,7 +93,3 @@ typedef int mp_off_t;
 #define MICROPY_MPHALPORT_H "pic16bit_mphal.h"
 #define MICROPY_HW_BOARD_NAME "dsPICSK"
 #define MICROPY_HW_MCU_NAME "dsPIC33"
-
-// XC16 toolchain doesn't seem to define these
-typedef int intptr_t;
-typedef unsigned int uintptr_t;
--- a/ports/pic16bit/Makefile
+++ b/ports/pic16bit/Makefile
@@ -7,7 +7,7 @@ QSTR_DEFS = qstrdefsport.h
 include $(TOP)/py/py.mk
 include $(TOP)/extmod/extmod.mk
 
-XCVERSION ?= 1.35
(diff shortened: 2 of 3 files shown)


ports/pic16bit/mpconfigport.h
```diff
@@ -93,7 +93,3 @@ typedef int mp_off_t;
 #define MICROPY_MPHALPORT_H "pic16bit_mphal.h"
 #define MICROPY_HW_BOARD_NAME "dsPICSK"
 #define MICROPY_HW_MCU_NAME "dsPIC33"
-
-// XC16 toolchain doesn't seem to define these
-typedef int intptr_t;
-typedef unsigned int uintptr_t;
```
dpgeorge, line 99: These are needed for XCVERSION=1.35, but I'm more than happy to stop supporting that old version, and remove these typedefs.

dpgeorge:
Thanks for this fix.  Doing the release v1.24.0 I did notice that pic16bit did not compile anymore, due to the clz issue pointed out above.

I've tested this PR using XCVERSION=2.00 and it compiles.  I did not try running it on hardware though.

PR #15874, opened 2024-09-19
stm32/mboot: Add mboot version string.

victorallume:
### Summary

This adds a configurable version string to a known location at the end of mboot flash section. 

Micropython v1.22 saw littlefs being upgraded to an incompatible version, requiring mboot to be updated prior to micropython. Except that without a way to determine (from micropython) what version mboot is on a device, it is difficult to make a decision to update mboot or not (in an OTA situation). This change writes a version string to a known location at the end of the mboot flash area (by default, this is 64 bytes from the end of the 32k flash block). By default, the version string is "mboot-upy-" plus the git ref as given by `git describe --dirty` but can be over-ridden by defining `MBOOT_VER` in your own build system (for example, it may be useful to include mboot configuration information such as supported filesystems in this version string)

To access the mboot version from micropython, something like the following python could be used:
```
import stm

MBOOT_BASE = 0x08000000
MBOOT_LEN = 32 * 1024
VER_LEN=64
s=''
base = MBOOT_BASE + MBOOT_LEN - VER_LEN
for i in range(VER_LEN):
    c = stm.mem8[base + i]
    if c == 0:
        break
    s += chr(c)
print(f'mboot version is {s}')
```
At a later point, this could be incorporated directly into the micropython or machine modules.

### Testing
Tested build for BOARD=PYBD_SF6. The file `firmware.bin` only differs after the end of the original; the new version includes zero padding up until the final 64 bytes then a null-terminated version string.
```
2988,2990c2988,2999
< 0057120    2400    00f4
<          \0   $ 364  \0
< 0057124
---
> 0057120    2400    00f4    0000    0000    0000    0000    0000    0000
>          \0   $ 364  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0
> 0057140    0000    0000    0000    0000    0000    0000    0000    0000
>          \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0  \0
> *
> 0077700    626d    6f6f    2d74    7075    2d79    3176    322e    2e33
>           m   b   o   o   t   -   u   p   y   -   v   1   .   2   3   .
> 0077720    2d30    3133    2d31    3467    6138    3664    3161    3338
>           0   -   3   1   1   -   g   4   8   a   d   6   a   1   8   3
> 0077740    0000
>          \0
> 0077741
```

Flashing the new mboot onto the pyboard works normally, and running the above sample code shows `mboot version is mboot-upy-v1.23.0-311-g48ad6a183`

If the mboot build is pushing the limit of the available mboot flash space, this will take away some of that, possibly preventing the build from working. In this case, the size of the version string can be decreased, or removed entirely.

This will also increase the size of the dfu (or bin or hex) file, as the payload is now padded out to the entire size of the flash block (32k). In my testing, this resulted in an increase in the dfu file of about 5k.


victorallume:
@Josverl Moving discussion to the PR. `git describe` does output the version number (tag) + hash unless the latest commit is tagged. Happy to change it if there's a particular format that's more sensible, but it's configurable (to allow storing configuration info in there as well)

dpgeorge:
Thanks for this, it's definitely a good addition.  Really mboot should have been versioned from the very beginning.

> This will also increase the size of the dfu (or bin or hex) file, as the payload is now padded out to the entire size of the flash block (32k)

Yes, that's an unfortunate side effect of doing it this way.  Eg if mboot is much larger (say 64k) then appending the version can increase the mboot firmware by quite a bit.

An alternative way to do it is put the null-terminated version string anywhere in the firmware, and then have a 4-byte pointer to the version string in a known location.  That known location could be in an unused slot in the IRQ vector table.  Or, an approach I've used before is to put that pointer in the `ResetHandler` function as the second word.  The first word of that function jumps over that pointer.  Then the application can retrieve this pointer by a few indirections:
```c
uintptr_t **irq_table = (uintptr_t **)0x08000000;
uintptr_t *reset_handler = irq_table[1];
const char *version_str = (const char *)reset_handler[1];
```

victorallume:
Thanks Damien, I've got the example code running, but have a couple of thoughts:

- Even though the NMI is not currently used, if it were to be used down the track, it's going to be more time sensitive than reset, so I'd be inclined to put it in the reset handler. Unless you think that the NMI will never get used in the future?
- We need the ability to determine if there is no mboot version stored (i.e. using an mboot prior to this being included). I'm figuring in this case, the pointer will either be to a junk location, or to a valid location, with unrelated data. I figure that prefixing the build-defined version with a magic string (e.g. "MBOOTVER") should solve this, unless you can think of a better approach?

victorallume:
I've modified the approach to using several levels of indirection from the NMI vector table. @dpgeorge can you please comment if you think this approach will be the best long-term approach, given my comments above? If so, I'll tidy up the commit.

I have also added a function to retrieve the version: `pyb.mboot_ver()` - this will return None if it can't find anything (i.e. a version of mboot prior to version info being available, or and empty string if it isn't defined. I've placed it in the `pyb` module as I figure this is specific to stm32, but if you think there's any better places for it (machine? machine.info?) I'll move it.

dpgeorge:
After thinking about this and considering the pros and cons of the different approaches, I think it's probably best to use the original scheme that you had @victorallume , of just putting the version at the end of the flash segment.

Having it at the end of flash:
- pro: simple and robust
- pro: easy to find (you just need to know the length of the mboot segment, which will usually be just before the VTOR/vector table of the application)
- pro: will be all 0xff's if the version doesn't exist
- con: uses a fixed amount of space in the flash even if the version string is not that long
- con: pads out the bootloader to the maximum flash space

Doing it with a pointer somewhere in the vector table:
- pro: version string can be as short or long as needed and only uses that many bytes
- pro: doesn't pad out the bootloader
- con: complicated to look up the pointer, and need to validate every step of indirection in case the pointer/version is not there
- con: hard to tell if the version is not there
- con: wastes space with the pointer, and also any prefix to help identify that the pointer is there (this could be 12 bytes wasted, which is significant compared to the length of the version string itself)

I feel that the complexity of the solution is the important deciding factor here, a bootloader should stay simple.  Just appending it at the end is simplest and IMO what we should do.

If the firmware padding out to the full size (eg 32k) is an issue then the bootloader can be compressed, or split into first and second (version) parts.

Also I think 32 bytes is enough space for the version string.  Eg `mboot-v1.24.0-73-g2e796d6c3` is 27 bytes and contains enough information.

victorallume:
> Also I think 32 bytes is enough space for the version string. Eg `mboot-v1.24.0-73-g2e796d6c3` is 27 bytes and contains enough information.

The output of `git describe --dirty` will end up going over 32 bytes in this case, but more importantly, I was thinking that build configuration could be added to that string (specifically, what filesystems are supported), so 64 strikes me as a better default. Without someone changing the config, how likely is it that you'd run out of space with a length of 64?

dpgeorge:
> I was thinking that build configuration could be added to that string (specifically, what filesystems are supported)

Yes, that sounds like a good idea.  Would you add such info to the default string?

> so 64 strikes me as a better default (it is configurable, but you probably wouldn't want to change it once it's set). Without someone changing the config, how likely is it that you'd run out of space with a length of 64?

OK, we can go for 64 as the default.  And I agree you wouldn't want to change it once it's set.

It's unlikely that a build would run out of space at this stage.  And anyone doing custom builds can easily shrink or disable this version string.

victorallume:
Yeah, I was thinking to include the supported filesystems as a default. Are there any other options that would be useful to include here? Packing is the only one I see of a quick scan through the code.

A somewhat ugly implementation of this is below. I would have preferred to define this at the Makefile level, but the macros for the included filesystems are defined in header files not Makefiles, so have to go this way. This creates a version string such as `mboot-v1.24.0-1-g28793b513-dirty[fat,raw,]` What do you think?

Also, are you happy with `mboot-` as the prefix to the version? I was thinking possibly `mboot-upy-` as the version is micropython, not mboot, but that difference is academic.

```
#include "mboot.h"

#ifdef MBOOT_VER

#if MBOOT_VER_FS_TYPES
#define MBOOT_VER_0 MBOOT_VER "["

#if defined MBOOT_VFS_FAT && MBOOT_VFS_FAT
#define MBOOT_VER_FAT MBOOT_VER_0 "fat,"
#else
#define MBOOT_VER_FAT MBOOT_VER_0
#endif

#if defined MBOOT_VFS_LFS1 && MBOOT_VFS_LFS1
#define MBOOT_VER_LFS1 MBOOT_VER_FAT "lfs1,"
#else
#define MBOOT_VER_LFS1 MBOOT_VER_FAT
#endif

#if defined MBOOT_VFS_LFS2 && MBOOT_VFS_LFS2
#define MBOOT_VER_LFS2 MBOOT_VER_LFS1 "lfs2,"
#else
#define MBOOT_VER_LFS2 MBOOT_VER_LFS1
#endif

#if defined MBOOT_VFS_RAW && MBOOT_VFS_RAW
#define MBOOT_VER_RAW MBOOT_VER_LFS2 "raw,"
#else
#define MBOOT_VER_RAW MBOOT_VER_LFS2
#endif

#define MBOOT_VER_COMPLETE MBOOT_VER_RAW "]"

#else //MBOOT_VER_FS_TYPES

#define MBOOT_VER_COMPLETE MBOOT_VER

#endif //MBOOT_VER_FS_TYPES

const char mboot_ver[] __attribute__((section(".mboot_ver"))) __attribute__ ((__used__)) = MBOOT_VER_COMPLETE;
#endif  //MBOOT_VER
```

dpgeorge:
> I was thinking to include the supported filesystems as a default. Are there any other options that would be useful to include here? Packing? USB support?

USB is always enabled, so no need to list that.  I would include:
- I2C support
- 64-bit address space `MBOOT_ADDRESS_SPACE_64BIT`
- enabled filesystems

> This creates a version string such as `mboot-v1.24.0-1-g28793b513-dirty[fat,raw,]` What do you think?

I think use a `+` to separate version and metadata, so it's somewhat semver.org compatible.  Eg:
```
mboot-v1.24.0-1-g28793b513+i2c.fat.raw
```
If it's tricky to insert `.` between the words but not at the end, then could add USB support at the start of the string (then any options always begin with `.`), eg:
```
mboot-v1.24.0-1-g28793b513+usb.i2c.fat.raw
```

> Also, are you happy with `mboot-` as the prefix to the version?

Yes I think just `mboot-` is fine.  It's short and identifiable.

victorallume:
I have added the options support; only thing is, with all current options enabled, we're starting to push 64 bytes: `mboot-v1.24.0-2-g0b8551277-dirty+usb.i2c.fat.raw.lfs2.lfs1.64` is 62 bytes including the null terminator. If that's acceptable, I'll clean up the PR. Otherwise we could abbreviate further (e.g. ub.i2.fa.l1.l2.64) to eke a few more bytes out, or drop the dirty from the git description (though I'm currently reworking the build to use the same version info as micropython itself via `makeversionhdr.py`)

dpgeorge:
> I have added the options support; only thing is, with all current options enabled, we're starting to push 64 bytes:

The "dirty" bit won't be used in production, it should always be built on a clean commit (otherwise the hash means nothing because there could be arbitrary changes to the code).

Also don't need the null terminator.  Either it ends at 64 bytes, or ends with 0xff's.

> It would also be preferable to define `MBOOT_VER_LEN` in individual board directories in `mpconfigboard.mk`, so it can be over-ridden more easily as required

You can use:
```
MBOOT_VER_LEN ?= 64
```
to provide a default.

ports/stm32/mboot/Makefile
```diff
@@ -36,6 +36,11 @@ include $(BOARD_DIR)/mpconfigboard.mk
 # A board can set MBOOT_TEXT0_ADDR to a custom location where mboot should reside.
 MBOOT_TEXT0_ADDR ?= 0x08000000
 
+# The string in MBOOT_VER will be stored in the final MBOOT_VER_LEN bytes of mboot flash.
+# Defaults to git version. Can make this region smaller by reducing MBOOT_VER_LEN
```
dpgeorge, line 40: "A board can change the size of this region by defining MBOOT_VER_LEN."

ports/stm32/mboot/Makefile
```diff
@@ -36,6 +36,11 @@ include $(BOARD_DIR)/mpconfigboard.mk
...
+MBOOT_VER_LEN ?= 64
+MBOOT_VER_INC_OPTS ?= 1  # if set to 1, this will append build options to version string (see ver.c)
```
dpgeorge, line 42: I suggest making these more explicitly named, for example:
```
MBOOT_VERSION_ALLOCATED_BYTES
MBOOT_VERSION_INCLUDE_OPTIONS
```

ports/stm32/mboot/Makefile
```diff
@@ -121,6 +130,7 @@ SRC_C += \
 	vfs_fat.c \
 	vfs_lfs.c \
 	vfs_raw.c \
+	ver.c \
```
dpgeorge, line 133: Suggest `version.c`.

ports/stm32/mboot/Makefile
```diff
@@ -206,7 +216,7 @@ deploy-stlink: $(BUILD)/firmware.dfu
 
 $(BUILD)/firmware.dfu: $(BUILD)/firmware.elf
 	$(ECHO) "Create $@"
-	$(Q)$(OBJCOPY) -O binary -j .isr_vector -j .text -j .data $^ $(BUILD)/firmware.bin
+	$(Q)$(OBJCOPY) -O binary -j .isr_vector -j .text -j .data -j .mboot_ver_data $^ $(BUILD)/firmware.bin
```
dpgeorge, line 219: `.mboot_version_text` (use text because it's in the text section)

ports/stm32/mboot/README.md
```diff
@@ -89,6 +89,16 @@ How to use
    the beginning of the chunk when the end is reached.  Then use a split
    raw filesystem to inform mboot of this wrapping.
 
+   The version and config options that mboot was built with are stored in at 
+   the end of the mboot flash block. The length of this is section defaults
```
dpgeorge, line 93: "The version and config options that mboot was built with are stored in a small, fixed section of bytes at the end of the flash region allocated for mboot.  The length of the fixed section defaults ..."

ports/stm32/mboot/README.md
```diff
@@ -89,6 +89,16 @@ How to use
...
+   to 64 bytes, but can be over-ridden by setting MBOOT_VER_LEN. If running
```
dpgeorge, line 94: overridden (without hyphen)

ports/stm32/mboot/README.md
```diff
@@ -209,6 +219,33 @@ and signed firmware, and can be deployed via USB DFU, or by copying it to the de
 internal filesystem (if `MBOOT_FSLOAD` is enabled). `firmware.dfu` is still unencrypted
 and can be directly flashed with jtag etc.
 
+Retrieving the mboot version in micropython
+--------------------------------------
+The following python code retrieves the version of mboot currently on the device:
+```python
+import stm
+
+def get_mboot_version(include_opts=True):
```
dpgeorge, line 228: I suggest putting this code in `fwupdate.py`, and adding arguments for the base, length and version length parameters (with defaults).

ports/stm32/mboot/README.md
```diff
@@ -209,6 +219,33 @@ and signed firmware, and can be deployed via USB DFU, or by copying it to the de
...
+    MBOOT_BASE = const(0x0800_0000)
+    MBOOT_LEN = const(32 * 1024)  # this will need to updated if mboot is a different size
+    VER_LEN = const(64)  # this will need to be updated if the length/location of version string changes
+    base = const(MBOOT_BASE + MBOOT_LEN - VER_LEN)
+    s = ''
+    for i in range(VER_LEN):
+        c = stm.mem8[base + i]
+        if c == 0xFF or c == 0:  # have hit empty flash or a null-terminator
```
dpgeorge, line 236: I don't think checking for null is needed.  The way the string is generated, null won't ever be there.

dpgeorge:
Thanks for updating this.  I still need to test it but it's looking good now.

ports/stm32/mboot/README.md
dpgeorge: I added back the null check.  It's needed because in some cases the end of the firmware is padded with 0x00's by the DFU programmer (to align the flash writes up to 8 bytes).

dpgeorge:
Tested on PYBV10 and PYBD_SF6, both work (although PYBV10 needs mboot to be 16k so it doesn't overwrite the filesystem).

I also fixed `version.c` so that the version string is excluded completely when `MBOOT_VERSION_ALLOCATED_BYTES=0` (previously it would attempt to build with a 0-length allocated string and hence hit the static assert).

dpgeorge (approved):

PR #15879, opened 2024-09-19
unix/modffi: Let RV64 pass FFI tests again.

### LINKED ISSUE #15528: Update libffi to 3.3 or later to unblock CI for RV64.

agatti:
### Summary

PR #15800 combined with #15767 would trigger an incompatibility in the libffi version used by the CI environment for testing the RISC-V 64 Unix port.  Using a source-built libffi library would also bring up a RISC-V 64 specific issue when handling unsigned 32 bit values in certain cases.

The first problem is solved by updating the vendored libffi version to v3.6.4 (the latest stable at this point), and pointing the libffi submodule to the official libffi repository.  This, along with some FFI callback memory management changes, closes #15528.

The second problem stems from how RISC-V 64 holds 32-bit integers inside a 64 bits wide register.  Quoting from the "RISC-V Instruction Set Manual Volume 1", section 4.2:

> The compiler and calling convention maintain an invariant that all 32-bit values are held in a sign-extended format in 64-bit registers. Even 32-bit unsigned integers extend bit 31 into bits 63 through 32. Consequently, conversion between unsigned and signed 32-bit integers is a no-op, as is conversion from a signed 32-bit integer to a signed 64-bit integer.

The compiler can handle this since it knows what is going to happen to values whose lifetime it manages, and can generate code that takes this situation into account.  However, for FFI functions the compiler has no idea on what's in store for the value and so it just leaves it alone.  The modffi code wasn't aware of this and would happily return 64-bit signed values if the `uint32_t` value returned by the function had its most significant bit set.

### Testing

Testing occurred in a CI-like environment using the same commands used in `ci.sh` to build both the Qemu Unix RISC-V 64, Arm, and MIPS ports.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,9 @@ build/
 build-*/
 docs/genrst/
 
-# Test failure outputs
+# Test failure outputs and intermediate artefacts
 tests/results/*
+tests/ports/unix/ffi_lib.so
 
 # Python cache files
 __pycache__/
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/micropython/axtls.git
 [submodule "lib/libffi"]
 	path = lib/libffi
-	url = https://github.com/atgreen/libffi
+	url = https://github.com/libffi/libffi
 [submodule "lib/lwip"]
 	path = lib/lwip
 	url = https://github.com/lwip-tcpip/lwip.git
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -168,7 +168,7 @@ ifeq ($(MICROPY_STANDALONE),1)
 # Build libffi from source.
 GIT_SUBMODULES += lib/libffi
 DEPLIBS += libffi
-LIBFFI_CFLAGS := -I$(shell ls -1d $(BUILD)/lib/libffi/out/lib/libffi-*/include)
+LIBFFI_CFLAGS := -I$(shell ls -1d $(BUILD)/lib/libffi/include)
  ifeq ($(MICROPY_FORCE_32BIT),1)
   LIBFFI_LDFLAGS = $(BUILD)/lib/libffi/out/lib32/libffi.a
  else
--- a/ports/unix/alloc.c
+++ b/ports/unix/alloc.c
@@ -85,25 +85,6 @@ void mp_unix_mark_exec(void) {
     }
 }
 
-#if MICROPY_FORCE_PLAT_ALLOC_EXEC
-// Provide implementation of libffi ffi_closure_* functions in terms
-// of the functions above. On a normal Linux system, this save a lot
-// of code size.
-void *ffi_closure_alloc(size_t size, void **code);
-void ffi_closure_free(void *ptr);
-
-void *ffi_closure_alloc(size_t size, void **code) {
-    size_t dummy;
-    mp_unix_alloc_exec(size, code, &dummy);
-    return *code;
-}
-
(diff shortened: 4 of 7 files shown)


dpgeorge:
> The second problem stems from how RISC-V 64 holds 32-bit integers inside a 64 bits wide register. Quoting from the "RISC-V Instruction Set Manual Volume 1", section 4.2:
> 
> > The compiler and calling convention maintain an invariant that all 32-bit values are held in a sign-extended format in 64-bit registers. Even 32-bit unsigned integers extend bit 31 into bits 63 through 32. Consequently, conversion between unsigned and signed 32-bit integers is a no-op, as is conversion from a signed 32-bit integer to a signed 64-bit integer.
> 
> The compiler can handle this since it knows what is going to happen to values whose lifetime it manages, and can generate code that takes this situation into account. However, for FFI functions the compiler has no idea on what's in store for the value and so it just leaves it alone. The modffi code wasn't aware of this and would happily return 64-bit signed values if the `uint32_t` value returned by the function had its most significant bit set.

Wow, very interesting, and great debugging!


> The alternative is to disable FFI tests for RISC-V 64.

I think now that you've done the work and have the FFI tests working, we should keep them enabled.  After all, it was these tests that picked up the above issue with signed/unsigned.  That's exactly why tests and large test coverage is important.

Great work!

dpgeorge:
Reviewing this PR in more detail, I think we should first fix #15528 and updated the libffi submodule to v3.4.6.  That will make things much simpler here.  Otherwise we'll just be adding workarounds on workarounds.

agatti:
Fair point, I've updated libffi and changed the callbacks' memory management code.  This should be now ready for merging, as all tests pass with the new libffi.

dpgeorge (approved):

dpgeorge:
Thanks for updating, this looks really good now, simpler and we have a more recent libffi version.

I tested it with the micropython-lib `signal` module, both with a dynamically linked libffi and in `MICROPY_STANDALONE=1` mode, and the ffi callback works.

dpgeorge:
Also good to see that the unix port did not grow in size, it actually shrunk by 6 bytes.  So the original motivation to have custom `ffi_closure_alloc` functions seems to no longer be relevant, at least for the `MICROPY_STANDALONE=0` build (see c9598604c6872d87b6c818c92efc4bce357d988c).

PR #11104, opened 2023-03-22
ports: adafruit grand central board support

graeme-winter:
Extend existing support for SAMD51 based boards to Adafruit Grand Central. 

Includes fixes to DAC implementation to support two channels, fixes to the RTC module which are probably applicable to other boards.

PR #15889, opened 2024-09-21
ports: adafruit grand central board support

graeme-winter:
### Summary

Adds board support for Adafruit Grand Central M4 (this is an updated version of #11104 which I am about to close, and includes contributions from @robert-hh)

Board: https://www.adafruit.com/product/4064

### Testing

Derived from PR which had fairly extensive testing, but re-built against current head of `master` to resurrect support: tested against hardware, DAC tests reported in previous PR.


graeme-winter:
Creating as draft PR to allow CI to run

graeme-winter:
Following suggestion from @robert-hh on #11104 have split off the DAC fix commit into a separate PR #15890 and this one is rebased against that

robert-hh:
The PR still includes the commit for the changed machine_dac.c. You can remove it with `git rebase -i`

graeme-winter:
> The PR still includes the commit for the changed machine_dac.c. You can remove it with `git rebase -i`

Removed - I had left it in for testing assuming #15890 will get merged first

graeme-winter:
Will need to add TC to generate PWM signal alongside TCC, since some conspicuous pins (e.g. `D13`) are not connected to TCC peripherals on this board. Also need to extend `pins.csv` to cover all the pins.

graeme-winter:
Note: also need to check / update `ports/samd/mcu/samd51/pin-af-table.csv`

graeme-winter:
> Will need to add TC to generate PWM signal alongside TCC, since some conspicuous pins (e.g. `D13`) are not connected to TCC peripherals on this board. Also need to extend `pins.csv` to cover all the pins.

From the notes: _TC_ - These device are currently not assigned to Pin. the TC device pair 0/1 is used for `ticks_us()`. PB01 is connected to:

```
33 PB01          D13 IRQ: 1 ADC0:13 ADC1: - Serial1:  -  Serial2: 5/3 TC: 7/1 PWM1:  -  PWM2:  -
```

So 7/1 probably available

robert-hh:
> Will need to add TC to generate PWM signal alongside TCC, since some conspicuous pins (e.g. D13) are not connected to TCC peripherals on this board. Also need to extend pins.csv to cover all the pins.

I was an intentional decision to use only TCC for PWM and keep TC for other tasks. TC is used for instance by ADC timed and DAC timed. That fact that not all pins can be used for PWM seems acceptable. That applies to other ports as well. Why do you need a PWM signal at D13?

robert-hh:
> also need to check / update ports/samd/mcu/samd51/pin-af-table.csv

What is missing in pin-af-table.csv and adafruit grand central's pins.csv?

robert-hh:
What's missing in this PR is a chapter in the pins assignment document `micropython/docs/samd/pinout.rst`.

graeme-winter:
> > also need to check / update ports/samd/mcu/samd51/pin-af-table.csv
> 
> What is missing in pin-af-table.csv and adafruit grand central's pins.csv?

Now I see that I was looking for D53 but this had been differently aliased in your csv file, as CS, will explore if we can have both.

dpgeorge:
The DAC PR has now been merged.

This board definition looks OK to me, but as @robert-hh mentioned it needs a section added to `docs/samd/pinout.rst`.

robert-hh:
@graeme-winter To make your life a little bit easier, here is an updated pinout.rst file. I do not have the Grand Cemtral board, so I loaded the Image to another board with a SAMD51x20 using your pins.csv file. So it should match.

[pinout.zip](https://github.com/user-attachments/files/18517703/pinout.zip)

robert-hh:
@dpgeorge Since this is about boards: since a while I have a set of GENERIC board definitions. Is it worth adding these? And there are a few other SAMD21 board definitions which could be added, if "Just another board" is not too boring.

dpgeorge:
> since a while I have a set of GENERIC board definitions. Is it worth adding these?

For the samd port?  Is it possible to make them generic, eg with generic clock/crystal settings (and I guess not SPI flash)?  I guess that could be useful, but it would also be useful to then have a list of boards for which the generic firmware is known to work with, so that users can be somewhat assured that it's worth trying and won't brick their board (even though it's very unlikely to brick something).

> And there are a few other SAMD21 board definitions which could be added, if "Just another board" is not too boring.

If you have the hardware for these and they are relatively common, then, yes, it's worth adding them.

But, please move this discussion to a separate PR so we don't derail this one.

robert-hh:
@graeme-winter Sorry, I should follow my own recipe of testing before posting. There was an error in the .rst file.
[pinout.zip](https://github.com/user-attachments/files/18520116/pinout.zip)

@dpgeorge Will open a new PR. The generic boards use the internal oscillator, like the ItsyBitsy boards, and only the internal flash.

PR #15902, opened 2024-09-24
lib/tinyusb: Update TinyUSB to release 0.17.0

dpgeorge:
### Summary

TinyUSB (finally!) has a new release, 0.17.0.  Usually we track latest master but now there's a release it would be good to update to that.

More importantly, an updated TinyUSB is needed for both RP2350 support #15619 and ESP32 shared TinyUSB integration #15108.  Both those PRs update TinyUSB to a commit prior to the 0.17.0 release, and it would be good to consolidate those updates into a single update of TinyUSB, hence this PR.

### Testing

Testing is needed for all existing TinyUSB ports:
- [x] mimxrt
- [x] nrf
- [x] renesas-ra FS
- [x] renesas-ra HS
- [x] rp2
- [x] samd

### DIFF
--- a/shared/tinyusb/mp_usbd.h
+++ b/shared/tinyusb/mp_usbd.h
@@ -40,6 +40,13 @@
 #include "device/dcd.h"
 #endif
 
+// Initialise TinyUSB device.
+static inline void mp_usbd_init_tud(void) {
+    tusb_init();
+    tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
+    tud_cdc_configure_fifo(&cfg);
+}
+
 // Run the TinyUSB device task
 void mp_usbd_task(void);
 
@@ -125,7 +132,8 @@ inline static bool mp_usb_device_builtin_enabled(const mp_obj_usb_device_t *usbd
 
 static inline void mp_usbd_init(void) {
     // Without runtime USB support, this can be a thin wrapper wrapper around tusb_init()
-    tusb_init();
+    // which is called in the below helper function.
+    mp_usbd_init_tud();
 }
 
 #endif
--- a/shared/tinyusb/mp_usbd_runtime.c
+++ b/shared/tinyusb/mp_usbd_runtime.c
@@ -428,8 +428,10 @@ void mp_usbd_init(void) {
     }
 
     if (need_usb) {
-        tusb_init(); // Safe to call redundantly
-        tud_connect(); // Reconnect if mp_usbd_deinit() has disconnected
+        // The following will call tusb_init(), which is safe to call redundantly.
+        mp_usbd_init_tud();
+        // Reconnect if mp_usbd_deinit() has disconnected.
+        tud_connect();
     }
 }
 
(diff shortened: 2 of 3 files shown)


robert-hh:
Tested with

- MIMXRT 1062 (Teensy 4.0), 
- SAMD21 with and without external flash , 
- SAMD51, 
- RP2, 
- NRF52840
 
all build and work. NRF seems better than before. Code size increase for SAMD21 is 48bytes. Lacking a ARDUINO_PORTENTA_C33 board I could not test it with Renesas-RA. But at least the firmware builds.
Will test it with other MIMXRT variants as well.

andrewleech:
Thanks @robert-hh I've got a renesas FS USB board I'll be able to test with. 

@iabdalkader no pressure but if you get a chance to test on the portenta renesas or would be good to confirm it hasn't been broken by this change.

dpgeorge:
@robert-hh thanks for testing!  That's very helpful.

I also tested with the following:
- rp2 / Pico
- samd / Itsy Bitsy M0 Express
- mimxrt / Teensy 4.0

I didn't see any regressions.  USB CDC throughput is unchanged.

andrewleech:
> USB CDC throughput is unchanged.

Hi @dpgeorge do you have a script/pattern you could share that you use to test CDC throughout? I'd like to check that before/after on renesas and esp boards.

dpgeorge:
> do you have a script/pattern you could share that you use to test CDC throughout?

Yes!  See #15909.

andrewleech:
Thanks for that @dpgeorge ! 
Renesas FS USB is slightly slower (right is with this PR applied), but not dramatically:
image

dpgeorge:
You probably need to do a few runs before you can compare code changes.  Eg the load of your PC can change the throughput.

dpgeorge:
@andrewleech so you were able to test the `renesas-ra` port?  If it worked then I think we are good to go with this PR.

andrewleech:
@dpgeorge sorry I wasn't entirely clear - yes renesas ra port is tested for FS connections and it does work. I don't have hardware to test HS myself though. FS is slightly slower but not enough to prevent moving forward I don't think.

dpgeorge:
@andrewleech OK, thanks.  So the remaining test would be for `renesas-ra` port in HS mode.

@iabdalkader are you able to test this PR on an Arduino Portenta C33?

iabdalkader:
> @iabdalkader are you able to test this PR on an Arduino Portenta C33?

@andrewleech @dpgeorge Yes I can test it today. I tested this too https://github.com/micropython/micropython/pull/15550 and it works fine, if you merge it first it will be easier for me to test both.

dpgeorge:
> I tested this too #15550 and it works fine, if you merge it first it will be easier for me to test both.

OK, that's now merged.

iabdalkader:
> OK, that's now merged.

This PR still needs to be rebased on master.

iabdalkader:
I pulled this PR locally, and rebased on master and tested C33, and it seems to work fine.

dpgeorge:
Thanks @iabdalkader for testing.

PR #15903, opened 2024-09-24
esp32/boards: add mac address selector option

budavariam:


budavariam:
I did not intend to add it here

PR #15440, opened 2024-07-10
esp32c2 support (IDF v5.2.2) 

TrekMax:
### Summary

The is esp32c2 officially supported in IDF v5.2 but requires a number of changes to work...
Pulling together changes and generic board support needed here.

### Testing

Functions verified by actual hardware include:
- machine.pin
- machine.i2c
- machine.adc
- machine.pwm
- machine.timer

PR #15906, opened 2024-09-24
esp32c2 support (IDF v5.3) 

TrekMax:
### Summary

The is esp32c2 officially supported in IDF v5.2 but requires a number of changes to work...
Pulling together changes and generic board support needed here.

### Testing

Functions verified by actual hardware include:
- machine.pin
- machine.i2c
- machine.adc
- machine.pwm
- machine.timer
- bluetooth


sanderjo:
> Isn't this already covered by #11869?

That is about esp32c6

This PR has subject esp32c2.

So c6 versus c2?

andrewleech:
Nice one @sanderjo there's a bunch of new chips getting added to IDF these days! 

It still needs testing but this commit on the top of my C6 branch is intended to make it easier to bring in new chips like this one if you want to take a look: https://github.com/micropython/micropython/pull/15862

agatti:
@sanderjo I saw the ESP32C2 definition in `mpconfigport.h` in @andrewleech's branch so I thought it had basic C2 integration too, along with IDF 5.2 - my bad.

agatti:
@TrekMax `MICROPY_EMIT_RV32` being enabled should work on ESP32C2 too, it has the same RV32 core as C3 after all.  Same for `MICROPY_NLR_SETJMP` being disabled.

andrewleech:
@agatti the latest changes on C6 (about to push) should enable the emitter on all risk chips based on the soc defines, rather than listing individual chips. 

I suggest this PR would be well placed to rebase on top of the C6 work once it's merged as lots of people have contributed to that PR to help it fill gaps / support that will then naturally apply to many of the newer chips.

agatti:
@andrewleech oh, that's nice to hear.  Hopefully that would make support for the C5 easier to add once that chip shows up in production :)

projectgus:
Thanks @TrekMax for updating your PR! (Note for housekeeping: this is replacing earlier PR #15440).

Similar to discussions above, suggest we tackle in this order:

1. Merge #11869 
2. Damien will do the upcoming 1.24 MicroPython release.
3. Merge #15862 - this should make adding future SoCs much simpler, but has some risk of bugs in other SoCs.
4. Rebase this branch, then review and merge ESP32-C2 support.

@TrekMax Let us know if we can help with any part of this process. It should be possible to update this PR with totally new commits without needing to close and open a new PR. There is a guide to doing this online at https://github.com/jimmo/git-and-micropython#README

TrekMax:
Thanks. I will wait for esp32c6 support (IDF v5.2) #11869 to be merged, then rebase this branch, and then review and commit ESP32-C2 support.

TrekMax:
@projectgus Hey , the commit has been rebased to the latest master, please review it. Thanks!

andrewleech:
Hi @TrekMax could you also take a look at this which is intended to make it much easier to bring in new chips: https://github.com/micropython/micropython/pull/15862

TrekMax:
@andrewleech Of course. I have ESP32C3, ESP32C2 and ESP32S3 to verify it.

dpgeorge:
@TrekMax please can you rebase this PR on the latest master, because there have been a lot of changes to the esp32 port, in particular support for IDF v5.3.

projectgus:
@TrekMax Further to damien's request: now that #15862 is merged a lot of the C2 specific #ifdefs are probably not needed, so the PR will hopefully become very simple. :grin:

ports/esp32/mpconfigport.h
```diff
@@ -34,6 +34,8 @@
 #define MICROPY_GC_INITIAL_HEAP_SIZE        (56 * 1024)
 #elif CONFIG_IDF_TARGET_ESP32S2 && !CONFIG_SPIRAM
 #define MICROPY_GC_INITIAL_HEAP_SIZE        (36 * 1024)
+#elif CONFIG_IDF_TARGET_ESP32C2
+#define MICROPY_GC_INITIAL_HEAP_SIZE        (40 * 1024)
```
projectgus, line 38: Just curious, is there a reason not to choose 36KB, same as ESP32-S2?

ports/esp32/machine_bitstream.c
```diff
@@ -172,16 +173,20 @@ static void machine_bitstream_high_low_rmt(mp_hal_pin_obj_t pin, uint32_t *timin
     // Cancel RMT output to GPIO pin.
     esp_rom_gpio_connect_out_signal(pin, SIG_GPIO_OUT_IDX, false, false);
 }
-
+#endif
 /******************************************************************************/
 // Interface to machine.bitstream
 
 void machine_bitstream_high_low(mp_hal_pin_obj_t pin, uint32_t *timing_ns, const uint8_t *buf, size_t len) {
+    #if CONFIG_SOC_RMT_SUPPORTED
```
projectgus, line 181:
```suggestion
    #if SOC_RMT_SUPPORTED
```

ports/esp32/modesp32.c
```diff
@@ -231,7 +231,9 @@ static const mp_rom_map_elem_t esp32_module_globals_table[] = {
 
     { MP_ROM_QSTR(MP_QSTR_NVS), MP_ROM_PTR(&esp32_nvs_type) },
     { MP_ROM_QSTR(MP_QSTR_Partition), MP_ROM_PTR(&esp32_partition_type) },
+    #if CONFIG_SOC_RMT_SUPPORTED
```
projectgus, line 234:
```suggestion
    #if SOC_RMT_SUPPORTED
```

ports/esp32/machine_bitstream.c
```diff
@@ -90,6 +90,7 @@ static void IRAM_ATTR machine_bitstream_high_low_bitbang(mp_hal_pin_obj_t pin, u
     mp_hal_quiet_timing_exit(irq_state);
 }
 
+#if CONFIG_SOC_RMT_SUPPORTED
```
projectgus, line 93:
```suggestion
#if SOC_RMT_SUPPORTED
```

(This is a nitpick, but the only reason the `CONFIG_` prefix copy exists is for KConfig.)

ports/esp32/esp32_rmt.c
```diff
@@ -47,9 +47,12 @@
 // This current MicroPython implementation lacks some major features, notably receive pulses
 // and carrier output.
 
+#if CONFIG_SOC_RMT_SUPPORTED
 // Last available RMT channel that can transmit.
 #define RMT_LAST_TX_CHANNEL (SOC_RMT_TX_CANDIDATES_PER_GROUP - 1)
-
+#else
+#define RMT_LAST_TX_CHANNEL (-1)
```
projectgus, line 54: Instead of defining this macro to a dummy value, is it possible to put `#if SOC_RMT_SUPPORTED` around the entire C file so it's not compiled on C2?

ports/esp32/boards/sdkconfig.ble
```diff
@@ -17,4 +17,4 @@ CONFIG_BT_NIMBLE_PINNED_TO_CORE=1
 
 # Increase NimBLE task stack size from the default, because Python code
 # (BLE IRQ handlers) will most likely run on this task.
-CONFIG_BT_NIMBLE_TASK_STACK_SIZE=6144
+CONFIG_BT_NIMBLE_TASK_STACK_SIZE=4096
```
projectgus, line 20: This change probably should go in a chip-specific `sdkconfig.c2`, unless we're very sure all SoCs can run with the smaller stack size?

(The git log shows it had to be increased in cd66aa05c so there is some risk of regression if we make it smaller on all chips.)

ports/esp32/boards/ESP32_GENERIC_C2/mpconfigboard.cmake
```diff
@@ -0,0 +1,8 @@
+set(IDF_TARGET esp32c2)
+
+set(SDKCONFIG_DEFAULTS
+    boards/sdkconfig.base
+    ${SDKCONFIG_IDF_VERSION_SPECIFIC}
+    boards/sdkconfig.ble
+    boards/sdkconfig.26mhz
```
projectgus, line 7: Do all ESP32-C2 modules use 26MHz crystals? I see the docs say it can support both 26MHz and 40MHz.

projectgus (changes requested):
Thanks for your patience and for the quick update of this PR, @TrekMax ! I have some comments below, mostly minor.

ports/esp32/boards/ESP32_GENERIC_C2/board.md
```diff
@@ -0,0 +1,3 @@
+The following files are firmware images that should work on most
+ESP32-C2-based boards with 4MiB of flash, including WROOM and MINI modules,
+that use the revision 3 silicon (or newer).
```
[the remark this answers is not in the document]
TrekMax: Yes, This copy-paste from ESP32-C3, I will modify it. ESP32C2

ports/esp32/.gitignore
```diff
@@ -1,2 +1,8 @@
 dependencies.lock
 managed_components/
+
+# MacOS desktop metadata files
+.DS_Store
+
+# Clangd cache
+.cache/
```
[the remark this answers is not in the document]
TrekMax: okays!

ports/esp32/boards/ESP32_GENERIC_C2/mpconfigboard.cmake
projectgus: I see you moved this into the `sdkconfig.c2` file, which is fine.

Do we need to add a comment in `board.md` that is supports boards with 26MHz crystals? Or do the modules all have 26MHz crystals, already?

projectgus:
@TrekMax Thanks for updating the PR in response to the review! It's looking pretty tidy.

There are a couple of unresolved questions above, do you know the answer to them?

ports/esp32/boards/deploy_c2.md
```diff
@@ -0,0 +1,14 @@
+Program your board using the esptool.py program, found [here](https://github.com/espressif/esptool).
```
projectgus, line 1: A reminder for the future that #16535 deletes these copy-paste deploy.md files, so depending on which PR merges first there is a cleanup step for the PR which is merged second.

projectgus: The linked PR just merged, so it's OK to rebase this, remove this file, and link to the generic `deploy.md` now.

AmirHmZz:
@TrekMax Any updates on this PR?

ports/esp32/boards/ESP32_GENERIC_C2/mpconfigboard.cmake
projectgus: Looks like all current "ESP32-C2 series" chips are ESP8684 and all ESP8684 modules use 26MHz crystals for compatibility with ESP8266.

Have added the crystal frequency to board.md in case someone gets one which uses 40MHz.

projectgus:
I've rebased this PR on master and done a number of tests. 

Good news: basic unit tests pass, peripheral unit tests pass.

Bad news: Wi-Fi is almost unusable, Bluetooth appears totally unusable.

* The `multi_net/` tests fail intermittently for non-TLS tests (timeouts, packet loss) and always fail for TLS tests. (ESP32-C3 passes these tests 100%).
* The `multi_inet/` tests mostly fail with ENOMEM for TLS. I tried tuning the heap size to get the `test_tls_sites_simple.py` to make a single TLS connection but could not (with 16KB Python heap the test fails with ENOMEM, with 12KB Python heap the test fails with MemoryError - i.e. there's either not enough ESP-IDF RAM or not enough Python RAM).
* The multi_bluetooth tests hard fault during initialisation (crash is in nimble_port, looks like the NimBLE host event queue is uninitialised.) Unsure what the root cause for this one is.

I think more work is needed for ESP32-C2 to be a supported MicroPython target. It has only 272KB of RAM vs 400KB for ESP32-C3, but we may be able to tune some ESP-IDF parameters to free up more RAM at runtime.

projectgus:
@dpgeorge Recommend we retarget this PR for 1.27 unless there's some miracle fix for the Wi-Fi & Bluetooth problems. :crying_cat_face: 

EDIT: Miracle fix may be possible, it turns out!

projectgus:
> miracle fix for the Wi-Fi & Bluetooth problems

Turns out this was possible! By tweaking various config settings it was possible to free up 40KB of static RAM (went from 60KB free static RAM to 100KB) and also reduce the runtime memory usage from Wi-Fi and TCP/IP. These tweaks are added in their own `sdkconfig.free_ram` file, as they're potentially useful for other chips as well.

With the free RAM increases implemented, tests pass:

* multi_net - passing [*][^]
* net_inet - passing
* multi_bluetooth - passing [*]
* unit tests - defaults, extmod_hardware & ports/esp32 - passing

[*] I had to tweak the run-multitests.py script for ESP32C2 as (like ESP8266) the ESP8684 prints its early boot log at 74880bps which is interpreted as line noise by the script and can trigger encoding errors. Will submit another PR for this.

[^] Except DTLS tests, it seems DTLS support has been disabled on ESP-IDF v5.4.2 (on all targets). Will investigate and submit a fix separately.

ports/esp32/boards/ESP32_GENERIC_C2/mpconfigboard.h
```diff
@@ -0,0 +1,13 @@
+// This configuration is for a generic ESP32C2 board with 4MiB (or more) of flash.
+
+#define MICROPY_HW_BOARD_NAME               "ESP32C2 module"
+#define MICROPY_HW_MCU_NAME                 "ESP32C2"
+
+#define MICROPY_HW_ENABLE_SDCARD            (0)
+#define MICROPY_PY_MACHINE_I2S              (0)
+
+// Network config
+#define MICROPY_PY_NETWORK_HOSTNAME_DEFAULT "esp32c2"
```
dpgeorge: I feel like this should go in `mpconfigport.h` with all the other ones.

ports/esp32/boards/ESP32_GENERIC_C2/mpconfigboard.h
```diff
@@ -0,0 +1,13 @@
...
+
+// Enable UART REPL for modules that have an external USB-UART and don't use native USB.
+#define MICROPY_HW_ENABLE_UART_REPL         (1)
```
dpgeorge: Does the C2 have USB?  If not, the comment here is wrong.

ports/esp32/machine_adc.c
```diff
@@ -87,7 +87,7 @@ static const machine_adc_obj_t madc_obj[] = {
     {{&machine_adc_type}, ADCBLOCK2, ADC_CHANNEL_7, GPIO_NUM_27},
     {{&machine_adc_type}, ADCBLOCK2, ADC_CHANNEL_8, GPIO_NUM_25},
     {{&machine_adc_type}, ADCBLOCK2, ADC_CHANNEL_9, GPIO_NUM_26},
-    #elif CONFIG_IDF_TARGET_ESP32C3
+    #elif CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32C6 || CONFIG_IDF_TARGET_ESP32C2
```
dpgeorge: Why is C6 added here?  There's already a C6 section below.

If the C6 should be here instead, then the one below can be removed.

ports/esp32/main.c
```diff
@@ -245,6 +245,13 @@ void boardctrl_startup(void) {
 }
 
 void MICROPY_ESP_IDF_ENTRY(void) {
+    #if CONFIG_IDF_TARGET_ESP32C2 && CONFIG_SOC_XTAL_SUPPORT_26M
```
dpgeorge: Is the `CONFIG_SOC_XTAL_SUPPORT_26M` needed?  It seems like that's unconditionally set on the C2.

Also, maybe this bit can be moved into `MICROPY_BOARD_STARTUP()`?  That would allow a board to not print this newline.

dpgeorge:
Thanks @projectgus for getting this working, and tweaking the RAM for WiFi/BLE.  That's really good, and this PR is relatively simple.

Just a few minor comments above, but otherwise I think this is good to go in.

ports/esp32/main.c
projectgus: It's not needed at the moment, but it is possible for a C2-based board to be configured for 40MHz crystal in which case it won't need this workaround.

Putting it in `MICROPY_BOARD_STARTUP()` seems like a good place.

dpgeorge: Isn't there a difference between supporting 26MHz and selecting 26MHz?

Eg, it looks like both `CONFIG_SOC_XTAL_SUPPORT_26M` and `CONFIG_SOC_XTAL_SUPPORT_40M` are currently enabled on the C2.

ports/esp32/machine_adc.c
projectgus: Good catch, this whole section needed changing (the C2 has one fewer ADC pin than the C3, the C6 has many more than either of the other two).

projectgus:
@dpgeorge Thanks for the detailed review, have updated!

ports/esp32/main.c
projectgus: Oh you're right, I picked the wrong config flag! My mistake.

The new version of the PR scopes it to the board so I've removed any `#if` for now.

projectgus:
With these RAM tweaks there's almost as much free RAM on the C2 as the original ESP32 (without SPIRAM) - 142KiB vs 125KiB:

Original:
```
>>> micropython.mem_info()
stack: 704 out of 15360
GC: total: 56000, used: 3200, free: 52800, max new split: 90112
 No. of 1-blocks: 59, 2-blocks: 10, max blk sz: 19, max free sz: 3287
```

C2:
```
>>> import micropython
>>> micropython.mem_info()
stack: 500 out of 15360
GC: total: 35968, used: 1232, free: 34736, max new split: 92160
 No. of 1-blocks: 22, 2-blocks: 7, max blk sz: 19, max free sz: 2158

File: ports/esp32/boards/ESP32_GENERIC_C2/board_init.c [hunk 15]
```diff
@@ -0,0 +1,37 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * Development of the code in this file was sponsored by Microbric Pty Ltd
```
dpgeorge: This line does not need to be here.

ports/esp32/boards/ESP32_GENERIC_C2/board_init.c
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#include "mpconfigport.h"
```
dpgeorge: Should include `py/mpconfig.h` instead.

ports/esp32/machine_bitstream.c
```diff
@@ -172,16 +173,20 @@ static void machine_bitstream_high_low_rmt(mp_hal_pin_obj_t pin, uint32_t *timin
...
 void machine_bitstream_high_low(mp_hal_pin_obj_t pin, uint32_t *timing_ns, const uint8_t *buf, size_t len) {
+    #if SOC_RMT_SUPPORTED
     if (esp32_rmt_bitstream_channel_id < 0) {
         machine_bitstream_high_low_bitbang(pin, timing_ns, buf, len);
     } else {
         machine_bitstream_high_low_rmt(pin, timing_ns, buf, len, esp32_rmt_bitstream_channel_id);
     }
+    #else
+    machine_bitstream_high_low_bitbang(pin, timing_ns, buf, len);
+    #endif
```
dpgeorge, line 186: I'm probably being too picky here, so feel free to ignore this suggestion, but this could be a little cleaner as:
```c
    #if SOC_RMT_SUPPORTED
    if (esp32_rmt_bitstream_channel_id >= 0) {
        machine_bitstream_high_low_rmt(pin, timing_ns, buf, len, esp32_rmt_bitstream_channel_id);
        return;
    }
    #endif
    machine_bitstream_high_low_bitbang(pin, timing_ns, buf, len);
}
```

ports/esp32/boards/ESP32_GENERIC_C2/board.md
```diff
@@ -0,0 +1,3 @@
+The following files are firmware images that should work on most
+ESP32-C2-based boards with 4MiB of flash and 26MHz crystal frequency.
```
dpgeorge: Maybe worth saying "with at least 4MiB of flash"?

dpgeorge (commented):
A few more minor comments.

ports/esp32/boards/ESP32_GENERIC_C2/board_init.c
projectgus: Oops, copied wrong boilerplate!

ports/esp32/boards/ESP32_GENERIC_C2/board.md
projectgus: I changed this because currently all ESP8684 chips have either 2MB or 4MB of embedded flash, and I actually thought the flash controller only supports those sizes. But it looks like it can at least theoretically support more.

projectgus:
@dpgeorge Updated!

dpgeorge (approved):
Looks good now, thanks!

PR #15907, opened 2024-09-24
esp32: Remove old compiler workaround.

agatti:
### Summary

The ESP32 port contains a workaround to avoid having a certain function in `py/parse.c` being generated incorrectly.  The compiler in question is not part of any currently supported version of ESP-IDF anymore, and the problem inside the compiler (well, assembler in this case) has been corrected a few years ago.

### Testing

The default test suite (as in `run-tests.py --target esp32`) was executed successfully on an ESP32-D0WD and on an ESP32C3 - using a firmware image built with both ESP-IDF 5.0.4 and 5.2.2.

### DIFF
--- a/ports/esp32/mpconfigport.h
+++ b/ports/esp32/mpconfigport.h
@@ -47,13 +47,6 @@
 #define MICROPY_EMIT_RV32                   (1)
 #endif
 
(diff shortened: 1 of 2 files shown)


dpgeorge (approved):
Thanks, this looks good to me.  Now we use IDF 5.2.2 and the xtensa toolchain version is much more recent, at esp-13.2.0_20230928

Issue #6769, opened 2021-01-14
esp32: crash when calling dynamically loaded native module function


tve:
I started writing a dynamically loadable native module and before I got to add any real code the "stub testing" I did caused a hard crash (Guru meditation error, etc). A backtrace shows:
```
PC: 0x400df854: mp_call_function_n_kw at /home/src/esp32/mpy-1.13/micropython/ports/esp32/../../py/r
untime.c:649
BT-0: 0x400df851: mp_call_function_n_kw at /home/src/esp32/mpy-1.13/micropython/ports/esp32/../../py
/runtime.c:646
BT-1: 0x400df98e: mp_call_method_n_kw at /home/src/esp32/mpy-1.13/micropython/ports/esp32/../../py/r
untime.c:666
BT-2: 0x400ed47d: mp_execute_bytecode at /home/src/esp32/mpy-1.13/micropython/ports/esp32/../../py/v
m.c:1085
```
it crashes on the `type->call`:
```
    // do the call
    if (type->call != NULL) {
        return type->call(fun_in, n_args, n_kw, args);
    }
```
Sometimes as I modify unrelated code, instead of a crash I get a non-sensical "tuple cannot be called" or "bytearray cannot be called" exception (I forget the exact exception text). I suspect that something that gets loaded as part of the module gets garbage collected or perhaps not properly sized... Notice the explicit GC calls in the test code, which I added to replace a gc.threshold I had in my original code.

I spent time to reduce to a small reproducible test case, which is in the attached zip file. For me this fails with both the released generic v1.12 and v1.13 firmwares (`esp32-idf4-20191220-v1.12.bin`, `esp32-idf4-20200902-v1.13.bin`). To reproduce:
- expand the zip archive [mpy-zzz.tar.gz](https://github.com/micropython/micropython/files/5816219/mpy-zzz.tar.gz)
- perform an erase_all on your esp32
- flash either v1.12 or v1.13
- `pyboard.py -f cp *.py modu8g2.mpy :/`
- start minicom, reset the board, and watch it crash

The crash looks like this:
```
MicroPython 1.12.0 v1.12 on 2019-12-20
initialising module self=3ffe5310
done with 3ffe5310
*** modinfo initial
access(): [0, 0, 0, 0]
access(1): 0
draw_glyph_GS4: start
draw_glyph_GS4(...): (nil)
modu8g2: initial ['__class__', '__name__', '__file__', 'access', 'draw_glyph_GS4']
[...]

PR #15908, opened 2024-09-25
py: Properly track root pointers for text, BSS and rodata imported from native .mpy files

### LINKED ISSUE #6592: .bss corruption with native modules on ESP32
Modify features1.c to change `uint16_t data16[4];` to `uint16_t data16[32];`.
Compile for xtensawin
Copy over features1.mpy and test.py containing:

```py
import gc
import features1

def demo():
    print(features1.access())
    features1.access()
    features1.access()
    gc.collect()

demo()
```

Import test then run test.demo() a few times. The list printed out by `features1.access()` should be all-zero but contains modified entries.

```
>>> import test
initialising module self=3ffe5440
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
>>> test.demo()
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
>>> test.demo()
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
>>> test.demo()
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 3, 0, 1, 0, 3, 0, 1, 0, 7, 0, 1, 0, 3, 0, 1, 0]
>>> test.demo()
[7, 0, 1, 0, 3, 0, 1, 0, 15, 0, 1, 0, 3, 0, 1, 0, 7, 0, 1, 0, 3, 0, 1, 0, 31, 0, 1, 0, 3, 0, 1, 0]
>>> test.demo()
[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 7, 0, 1, 0, 3, 0, 1, 0, 31, 0, 1, 0, 3, 0, 1, 0]
>>> test.demo()
[7, 0, 1, 0, 3, 0, 1, 0, 15, 0, 1, 0, 3, 0, 1, 0, 7, 0, 1, 0, 3, 0, 1, 0, 31, 0, 1, 0, 3, 0, 1, 0]
```

This happens on both spiram and non-spiram builds. Making `data16` larger (or smaller) seems to make the issue go away?

dpgeorge:
### Summary

When importing a `.mpy` file with native code, memory is allocated to hold the text of the native code (the machine code), and also optionally a BSS/rodata section.

On certain architectures, the text may be allocated in a special region, outside the MicroPython GC heap.  On other architectures it's allocated on the GC heap.

The BSS/rodata is always allocated on the GC heap.

The trick is making sure pointers to GC-allocated memory are retained and reachable by the GC.  This is difficult because in a lot of cases a pointer to the head of the allocated block is not the thing that's passed around.  Rather a pointer inside the block is used, eg pointing to a native function when the module itself was deleted.

This PR attempts to make the GC scanning work correctly on all architectures.  It was previously broken on esp32, but cleaning it up also changed how the unix port does things.  ARM architectures are not known to have any bugs and are functionally unchanged with this PR.

Should fix #6592 and #6769.

### Testing

The existing test was updated to support esp32 (xtensawin), and also be more comprehensive so the bug could be visible on other architectures.

Tested on unix, esp32 and stm32 ports.  To test that the bug is there I disabled the `track_root_pointer()` function added in this PR and indeed the new test added here failed on all three of these platforms.

Even though there's quite a bit of code change, the only real change in the compiled firmware is:
- esp32: the BSS/rodata memory is now tracked explicitly
- unix: the BSS/rodata memory is tracked explicitly, and the mmap'd executable memory is no longer scanned by the GC


py/mpconfig.h
```diff
@@ -1992,15 +2015,17 @@ typedef double mp_float_t;
...
-// If these MP_PLAT_*_EXEC macros are overridden then the memory allocated by them
-// must be somehow reachable for marking by the GC, since the native code
-// generators store pointers to GC managed memory in the code.
+// If these macros are defined then the memory allocated by them does not need to be
+// traced by the GC.  But if they are left undefined then the GC heap will be used as
+// the allocator and the memory must be traced by the GC.
 #ifndef MP_PLAT_ALLOC_EXEC
-#define MP_PLAT_ALLOC_EXEC(min_size, ptr, size) do { *ptr = m_new(byte, min_size); *size = min_size; } while (0)
+// Leave as undefined.
+// #define MP_PLAT_ALLOC_EXEC(min_size, ptr, size) do { *ptr = malloc(min_size); *size = min_size; } while (0)
```
dpgeorge, line 2023: Not so happy how this bit turned out, needing to leave these undefined to get the correct behaviour.

The point is, if you define this macro then there is different tracking requirements.

Maybe a better way would be something like:
```c
#ifndef MP_PLAT_ALLOC_EXE
// Text tracked by the GC.
#define MP_PLAT_ALLOC_EXEC(min_size, ptr, size) do { *ptr = m_new(byte, min_size); *size = min_size; } while (0)
#define MICROPY_PERSISTENT_CODE_TRACK_FUN_DATA (1)
#define MICROPY_PERSISTENT_CODE_TRACK_BSS_RODATA (0)
#else
// Text not tracked by the GC, explicitly track BSS/rodata.
#define MICROPY_PERSISTENT_CODE_TRACK_FUN_DATA (0)
#define MICROPY_PERSISTENT_CODE_TRACK_BSS_RODATA (1)
#endif
```

projectgus: Yeah, I like this suggestion for how to structure it. A bunch of other #if/#else should simplify out as well.

py/persistentcode.c
```diff
@@ -72,6 +72,20 @@ typedef struct _bytecode_prelude_t {
...
+#if MICROPY_PERSISTENT_CODE_TRACK_FUN_DATA || MICROPY_PERSISTENT_CODE_TRACK_BSS_RODATA
+
+// An mp_obj_list_t that tracks native text/BSS/rodata to prevent the GC from reclaiming them.
+MP_REGISTER_ROOT_POINTER(mp_obj_t persistent_code_root_pointers);
+
+static void track_root_pointer(void *ptr) {
+    if (MP_STATE_PORT(persistent_code_root_pointers) == MP_OBJ_NULL) {
+        MP_STATE_PORT(persistent_code_root_pointers) = mp_obj_new_list(0, NULL);
+    }
+    mp_obj_list_append(MP_STATE_PORT(persistent_code_root_pointers), MP_OBJ_FROM_PTR(ptr));
```
projectgus, line 84: I'm guessing there's a good reason this doesn't work, but I'm curious what it is: What about extending `mp_raw_code_t` (possibly by creating a new wrapper struct that's optionally subbed in for `mp_raw_code_t *rc = mp_emit_glue_new_raw_code();`), and then append the code pointer to a field at the end of it? That way couldn't the raw code pointers stay valid until all of the raw code objects were GCed, but then be GCed themselves if the code falls out of scope?

py/mpconfig.h
dpgeorge: Actually, it wasn't that simple that I could apply my above idea.  But I've adjusted things a bit to keep the existing default definition of `MP_PLAT_ALLOC_EXEC`.

iabdalkader:
Will this conflict with or break the changes here: https://github.com/micropython/micropython/compare/master...dpgeorge:micropython:py-micropython-fun-to-mpy ? Maybe now would be a good time to add this feature upstream ?

py/persistentcode.c
projectgus: Got it, thanks.

py/mpconfig.h
projectgus: New update still a good simplification, I think!

projectgus (approved):

dpgeorge:
> Will this conflict with or break the changes here

No, the changes here are only related to native functions with machine code.  That branch only supports bytecode functions, so is independent to this PR.

> Maybe now would be a good time to add this feature upstream ?

There's a lot to do just to get 1.24.0 done, so I'd rather wait with that feature until after the release is made.

PR #15909, opened 2024-09-25
tools/test_serial.py: Add test for serial throughput.

dpgeorge:
### Summary

This is a test script used to test USB CDC (or USB UART) serial reliability and throughput.

### Testing

Run against any MicroPython board with:
```
$ python test_serial.py <device>
```

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -203,6 +203,18 @@ internal_bench/bytebuf:
 1 tests performed (3 individual testcases)
 ```
 
+## Serial reliability and performance test
+
+Serial port reliability and performance can be tested using the `serial_test.py` script.
+Pass the name of the port to test against, for example:
+
+    $ ./serial_test.py -t /dev/ttyACM0
+
+If no port is specified then `/dev/ttyACM0` is used as the default.
+
+The test will send data out to the target, and receive data from the target, in various
+chunk sizes.  The throughput of the serial connection will be reported for each sub-test.
+
 ## Test key/certificates
 
 SSL/TLS tests in `multi_net` and `net_inet` use self-signed key/cert pairs
--- a/tests/serial_test.py
+++ b/tests/serial_test.py
@@ -1,0 +1,263 @@
+#!/usr/bin/env python
+#
+# Performance and reliability test for serial port communication.
+#
+# Basic usage:
+#   serial_test.py [-t serial-device]
+#
+# The `serial-device` will default to /dev/ttyACM0.
+
+import argparse
+import serial
+import sys
+import time
+
+run_tests_module = __import__("run-tests")
+
+read_test_script = """
+bin = True
+try:
+    wr=__import__("pyb").USB_VCP(0).send
+except:
+    import sys
+    if hasattr(sys.stdout,'buffer'):
+        wr=sys.stdout.buffer.write
+    else:
+        wr=sys.stdout.write
+        bin = False
+b=bytearray(%u)
+if bin:
+    wr('BIN')
+    for i in range(len(b)):
+        b[i] = i & 0xff
+else:
+    wr('TXT')
+    for i in range(len(b)):
+        b[i] = 0x20 + (i & 0x3f)
+for _ in range(%d):
+    wr(b)
+"""
+
+
+write_test_script_verified = """
+import sys
+try:
+    rd=__import__("pyb").USB_VCP(0).recv
+except:
+    rd=sys.stdin.readinto
+b=bytearray(%u)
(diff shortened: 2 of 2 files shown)


andrewleech:
I've just tested on a esp32-S3 CDC with my shared/tinyusb migration PR: #15108
It works correctly without timeouts (@dpgeorge reported similar results previously on that PR) 
`DATA IN: 0.09 MBits/sec` / `DATA OUT: verify=0, bufsize=9999` `1.14 MBits/sec`

@robert-hh I've seen the same  0.09MBit/s DATA IN rates on renesas board, not sure why they're so low.

I also get lock-ups with the black pill board. I suspect it might be due to the 25mhz crystal causing instability - that is said to be the reason the built in DFU struggles to connect to usb more often than not.

On other stm32 boards though, I get much better speeds. WB55 for instance:
```
DATA IN: bufsize=256, read 32768 bytes in 63.21 msec = 506.22 kibytes/sec = 4.15 MBits/sec
DATA OUT: verify=0, bufsize=9999, wrote 639936 bytes in 1286.46 msec = 485.78 kibytes/sec = 3.98 MBits/sec
```
However for this I needed to increase the delay before the initial drain as the soft reboot took a little longer to complete.
```
def drain_input(ser):
    time.sleep(0.5)  # <---
    while ser.inWaiting() > 0:
        data = ser.read(ser.inWaiting())
        time.sleep(0.1)
```

rp2 (tinyusb)
```
DATA IN: bufsize=256, read 32768 bytes in 67.66 msec = 472.96 kibytes/sec = 3.87 MBits/sec
DATA OUT: verify=0, bufsize=9999, wrote 639936 bytes in 4998.38 msec = 125.03 kibytes/sec = 1.02 MBits/sec
```

andrewleech:
Ah... some of the slow data rates are caused by UART!
On renesas-ra, while I'm using the USB port only, the stdout goes to all ports and uart is a blocking write. Disabling uart repl raised the DATA IN rates at least from 0.09 MBits/sec up to 3-5 MBits/sec. The DATA OUT rates are still slow though.

dpgeorge:
> Ah... some of the slow data rates are caused by UART!

Oh, yes, that's bitten me a few times.  You need to disable UART if you want to test USB throughput.

You could also use this tool to test UART reliability, if you connect up a USB-UART convertor.

projectgus:
I just noted some very interesting results testing a PICO2 on the DATA IN test for #17960.

| Computer | Port | Port Type | Max Speed Observed |
| ----------- | ----- | ---------- | ------------ |
| Desktop (Ryzen 9900) | USB 3.2 Root Port | On-CPU controller | 5.40Mbit/sec |
| Desktop (Ryzen 9900) | USB 3 Hub Ports | Hubs plugged to on-CPU controller | 5.40Mbit/sec |
| Desktop (Ryzen 9900) | USB 3.1 Root Port | Motherboard controller (off-CPU) | 1.35Mbit/sec (!) |
| Laptop (i5-8265U) | USB 3.1 Root Port | Integrated chipset controller | 4.95Mbit/sec | 

So I think we're inadvertantly (and probably uncontrollably) benchmarking a few things here:

* Host Python performance (both these computers have CPython 3.13).
* Kernel I/O performance (`time` says none of these tests are running on CPU more than 250ms for the whole ~30 second run, about 85% user and 15% system).
* USB link quality (bad cable, noise issues, etc.)
* USB controller chipset quirks (I suspect that's the only thing that can explain my one very slow result). As I understand it, USB 3.x hubs implement an entire USB 2.0 controller inside them (and then encapsulate downstream packets as USB 3.x Superspeed), so plugging a device like a dev board into a USB 3.x hub depends a lot on the hub's controller.

Still seems like a good test, I think the thing to keep in mind is that it's only good for comparing between runs on the exact same hardware (even down to the port choice). :shrug:

dpgeorge:
Interesting data!  And interesting you got pretty much the same result as me for your laptop.  For reference my laptop is an i7-8550U.

Apart from performance, this test is also useful to test link reliability.  If a port can pass this test then it's REPL is most likely quite stable.

dpgeorge:
I've now updated this PR:
- renamed to `tests/serial_test.py`
- used `argparse` instead of hand crafting the CLI parsing
- used the same `--test-instance` argument as all the other tests to select the target device
- removed unused code
- added a small bit of documentation to `tests/README.md`

It's now ready to merge.

Note: the script now depends on `tests/run-tests.py` to get the helper function for the `--test-instance` argument.  That means it's no longer a stand-alone test.  Hopefully that's OK.

PR #15910, opened 2024-09-25
esp32: Fix hang in taskYIELD() on riscv CPUs when IRQs disabled.

### LINKED ISSUE #15846: esp32c3: calling ticks_us() for >=32 times will cause hang when IRQ is disabled

esp32c3

### MicroPython version

up to commit f1bdac375240941edfc0aa04b646dc3c53d6b371

### Reproduction

st_irq = machine.disable_irq()
tm_til = ticks_us()+1000
while ticks_us()<tm_til: pass
machine.enable_irq(st_irq)

### Expected behaviour

The program can run both with and without IRQ disabled (i.e., with and without the 1st and 4th line).

### Observed behaviour

Without disabling IRQ (by removing the 1st and 4th line), the program can run. However, with IRQ disabled, the program crashes due to WDT timeout, and system restarts.

```
1

MPY version : v1.24.0-preview.351.g0f65d98dd.dirty on 2024-09-14
IDF version : v5.0.4
Machine     : ESP32C3 module with ESP32C3

Guru Meditation Error: Core  0 panic'ed (Interrupt wdt timeout on CPU0). 

Core  0 register dump:
MEPC    : 0x4038f69e  RA      : 0x4038f68a  SP      : 0x3fca7f60  GP      : 0x3fc97400  
TP      : 0x3fc6d790  T0      : 0x3fca7dc0  T1      : 0x4038fc48  T2      : 0xffffffff  
S0/FP   : 0x00000022  S1      : 0x3c142750  A0      : 0x00000000  A1      : 0x00000000  
A2      : 0x600c0000  A3      : 0x00000000  A4      : 0x00000001  A5      : 0x00000001  
A6      : 0x00000002  A7      : 0x00000000  S2      : 0x3fc9b000  S3      : 0x3c140000  
S4      : 0x3c173000  S5      : 0x3c14c000  S6      : 0x3c14c000  S7      : 0x3c14c000  
S8      : 0x3c14c000  S9      : 0x3c142000  S10     : 0x3c142000  S11     : 0x3fca80b0  
T3      : 0x00000031  T4      : 0x00000001  T5      : 0x00000061  T6      : 0x00000009  
MSTATUS : 0x00001881  MTVEC   : 0x40380001  MCAUSE  : 0x00000018  MTVAL   : 0x0000c701  
MHARTID : 0x00000000  

Stack memory:
3fca7f60: 0x00000000 0x3fcaf890 0x3fc9a6bc 0x420078b6 0x3c142773 0x0000000a 0x3c142773 0x4202b2c6
3fca7f80: 0x3fcaf710 0x3fcb15f4 0x3fca7f90 0x3fcaf890 0x3c142740 0x3c142750 0x00000023 0x4200c212
3fca7fa0: 0x00000000 0x00000100 0x3c142740 0x4200cd2e 0x3fcb15e0 0x3fcab460 0x00000003 0x3fcb0510
3fca7fc0: 0x3c14c000 0x3c14c000 0x3c14c000 0x3c140000 0x00000000 0x00000100 0x00000000 0x4202b3fc
3fca7fe0: 0x3c14c000 0x3c14c000 0x3c14c000 0x0000002b 0x0000002a 0x3fc9a6bc 0x3fcb2348 0x00000000
3fca8000: 0x00000000 0x3fcaf890 0x4202b314 0x00000016 0x3fca8070 0x00000000 0x3c140000 0x3c14c000
3fca8020: 0x3c14c000 0x3c14c000 0x3c14c000 0x3c14c000 0x3c142000 0x3c142000 0x3fca80b0 0x3fca7fe0
3fca8040: 0x00000002 0x00000003 0x00000004 0x3c140000 0x00000002 0x00000003 0x00000004 0x4202b7ce
3fca8060: 0x3c141b0c 0x00000001 0x3fc9a6bc 0x42007000 0x00000038 0x0000002a 0x3fcb1780 0x3c142000
3fca8080: 0x3c142000 0x3c142000 0x3c142000 0x3fcb812c 0x00003c00 0x3fca812c 0x42007000 0x3fc9a6bc
3fca80a0: 0x00000001 0x3fc9f000 0x3fc9ef70 0x420072dc 0x00000000 0x00000000 0x00000000 0x3fca80b0
3fca80c0: 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000
3fca80e0: 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000
3fca8100: 0x00000000 0x00000000 0x00000000 0x00000000 0x00000000 0xa5a5a5a5 0xa5a5a5a5 0xa5a5a5a5
3fca8120: 0xa5a5a5a5 0xa5a5a5a5 0x00010000 0x55555555 0xaaaaaa56 0x9aaa96aa 0xa5559aaa 0xaa5556aa
3fca8140: 0xaaaaaaaa 0xaaaaaaaa 0x5596aaaa 0x9aaa9559 0x5556aaaa 0x9aaa9555 0x5a6a555a 0xaaaaaaa6
3fca8160: 0x6aaaa9aa 0xa5aaaaaa 0xaa9aaaaa 0x6a9aaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaa9aa 0xaaaaaaaa
3fca8180: 0xa9aaaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaa9aa 0x6595aaaa 0x55555555
3fca81a0: 0x55965955 0x55556555 0x55555559 0xaa9596a5 0xa996555a 0x65aaaaaa 0x5a5aaaa5 0x6aaa5aaa
3fca81c0: 0x5aa96aa9 0x96aaa596 0x5aa5a96a 0xaaaa5aaa 0xaaaaaaaa 0x5a96aa5a 0x596aa96a 0x695a595a
3fca81e0: 0xaaaa955a 0xaaaaaaaa 0xaaaaaaaa 0xaaa9aaaa 0x565aaa59 0x65965596 0x59659659 0x56596596
3fca8200: 0x96596595 0xa5965965 0xa965aa56 0xaaaaa96a 0xaa5a6aaa 0x55aaa995 0x5555aa65 0x9aaaaa95
3fca8220: 0xaa555aaa 0xaaaa9aaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaa9aa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa
3fca8240: 0xaa9aaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa 0xaaaaaaaa
3fca8260: 0xaaaaaaaa 0xaaa9aaaa 0xaaaaaaaa 0xaaaaaaaa 0x655956aa 0x55565559 0xa5aaaaa5 0x55965a96
3fca8280: 0x59555556 0x95659565 0x69555aaa 0xaaaaaaaa 0x56656aaa 0x95965696 0x6aaa95a5 0xaaa59659
3fca82a0: 0x6aa5aa96 0xa96956a9 0x66aa5aaa 0x96a9a666 0xaaaaaaaa 0xaaaaaaaa 0x6aaaaaaa 0x6aaaaaa9
3fca82c0: 0xa96aaaa9 0x55555555 0x5a955556 0x696aa96a 0xaaaaaa96 0x6aa96666 0x965a9659 0xaa5aaaa5
3fca82e0: 0xa96aa5aa 0xaaa96aaa 0x6a5aaa5a 0x59696969 0x66696a5a 0xa6a9a555 0xaaa999aa 0xaaaaaaaa
3fca8300: 0xaaaaaaaa 0xaaaaaaaa 0x6555aaaa 0xa566aa6a 0x5669aaaa 0x95555655 0x59655566 0x555a9a6a
3fca8320: 0x65559555 0xaa95a959 0x55954aaa 0x64965659 0x99995956 0x96aa9a99 0xa6aaaaaa 0x55aaaaaa
3fca8340: 0xaaaa5645 0x166649aa 0xaaaaaa49 0xaaaaaaaa 0xaaaa9a9a 0x669a9aa9 0xa5152a9a 0xa990949a



ELF file SHA256: 35be7c8f87ac78c1

Rebooting...

```

### Additional Information

### Code of Conduct

Yes, I agree

projectgus:
### Summary

 Closes #15846, fixing a regression introduced in #15476.

The hang occurs because the esp32 port was calling "from ISR" port-layer functions to set/clear the interrupt mask. FreeRTOS kernel therefore doesn't know the CPU is in a critical section. In `taskYIELD()` the riscv port layer [blocks after yielding until it knows the yield has happened](https://github.com/espressif/esp-idf/blob/v5.2.2/components/freertos/FreeRTOS-Kernel/portable/riscv/port.c#L652), and would block indefinitely if IRQs are disabled (until INT WDT triggers).

Moving to the "public" `portENTER_CRITICAL`/`portEXIT_CRITICAL API` means that FreeRTOS knows we're in a critical section and can react accordingly.

Adds a regression test for this case (should be safe to run on all ports).

On single core CPUs, this should result in almost exactly the same behaviour apart from fixing this case.

On dual core CPUs, we now have cross-CPU mutual exclusion for atomic sections. This also shouldn't change anything, mostly because all the code which enters an atomic section runs on the same CPU. If it does change something, it will be to fix a thread safety bug.

### Testing

Ran unit tests on ESP32, ESP32-C3 and ESP32-S2 and verified all passing (one vfat related test is failing on S2, but it also fails on master - will investigate separately). Tested with IDF V5.2.2 on all three chips, and also V5.0.4 on ESP32-C3 only.

There is some risk that this change triggers a FreeRTOS crash where there is a call to a blocking FreeRTOS API with interrupts disabled. Previously this code might have worked, but was probably thread unsafe and would have hung in some circumstances - now FreeRTOS knows it's in a critical section so it may crash outright.

There's no FreeRTOS API to verify if interrupts are disabled, but the alternative would have been to track this manually as per the workaround commit https://github.com/micropython/micropython/issues/15846#issuecomment-2355999409 . However I think using the "proper" critical section is more future-proof, and might avoid other subtle bugs in SMP configs. There is a little more runtime overhead to using the critical section, but this code is pretty optimised in ESP-IDF.


tests/extmod/machine_disable_irq.py
```diff
@@ -0,0 +1,18 @@
...
+    from time import ticks_us, ticks_add, ticks_diff
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+for _ in range(5):
+    st_irq = disable_irq()
+    print("disabled")
+    # busy-wait in a tight loop for 1ms, to simulate doing some work in a critical section
+    now = ticks_us()
```
robert-hh, line 13: This line seems not to be needed.

projectgus: Oops, thanks!

projectgus:
Replying to this question asked on the linked issue:

> > I've submitted a candidate fix in the linked PR,
> Where exactly? I do not see it in the linked PR.

It's a bit subtle. The hang [happens in this line of the port layer](https://github.com/espressif/esp-idf/blob/v5.2.2/components/freertos/FreeRTOS-Kernel/portable/riscv/port.c#L652):

```
system_cpu_int_reg = SYSTEM_CPU_INTR_FROM_CPU_0_REG;
[...]
while (port_xSchedulerRunning[coreID] && port_uxCriticalNesting[coreID] == 0 && REG_READ(system_cpu_int_reg + 4 * coreID) != 0) {}
```

Which is looping until the interrupt for the yield has triggered and cleared the interrupt register. This is needed because the CPU keeps running for a few cycles otherwise, and can lead to weird problems like a yielding function continuing to execute a few instructions before it actually yields. There's a long comment there (that I think I wrote in 2020, lol) with some explanation.

If interrupts are disabled then the interrupt never runs to clear that register, so it loops here indefinitely.

This PR changes it so that now FreeRTOS knows it's in a critical section, specifically `port_uxCriticalNesting` value is set and the loop exits immediately.

tests/extmod/machine_disable_irq.py
```diff
@@ -0,0 +1,17 @@
+# Test executing Python code while IRQs disabled.
+try:
+    from machine import disable_irq, enable_irq
+    from time import ticks_us, ticks_add, ticks_diff
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+for _ in range(5):
+    st_irq = disable_irq()
```
dpgeorge, line 10: Although it's not explicitly documented as such, the intention is that calls to this function can be nested.

I tested that with this PR on ESP32 and ESP32C3, and it seems to work, so that's great!

I suggest adding a test for that nested calling.  Eg:
```py
def f(n):
    st_irq = disable_irq()
    if n:
        f(n - 1)
    else:
        <tight loop code>
    enable_irq(st_irq)

for nest in range(3):
    for _ in range(5):
        f(nest)
```

projectgus: Done, have swapped test and confirmed it still fails before and passes after the fix is applied on C3.

tests/extmod/machine_disable_irq.py
```diff
@@ -0,0 +1,26 @@
...
+# Structured to also nest disable_irq() calls
+def f(n):
+    st_irq = disable_irq()
+    if n:
+        f(n - 1)
+    else:
+        # busy-wait in a tight loop for 1ms, to simulate doing some work in a critical section
+        tm_til = ticks_add(ticks_us(), 1000)
+        while ticks_diff(tm_til, ticks_us()) > 0:
+            pass
```
dpgeorge, line 19: I usually write this kind of pattern without using `ticks_add()` as:
```py
t0 = ticks_us()
while ticks_diff(ticks_us(), t0) < 1000:
    pass
```
I see that the way you've done it is the way the example is written in the `ticks_add()` docs... but you'll see in the `ticks_diff()` docs that it's done similar to what I wrote above.

Anyway, this is not really important, just FYI if you wanted to make these kinds of loops a little simpler.

projectgus: Yeah, I agree that's cleaner. Have changed, also rebased and re-ran tests on C3.

tests/extmod/machine_disable_irq.py
```diff
@@ -0,0 +1,26 @@
+# Test executing Python code while IRQs disabled.
+try:
+    from machine import disable_irq, enable_irq
+    from time import ticks_us, ticks_add, ticks_diff
```
dpgeorge, line 4: ticks_add is no longer needed

projectgus: Removed

dpgeorge (dismissed):

dpgeorge:
Testing this again on stm32, unfortunately the new test does not reliably pass.  That's because on stm32 the ticks-us counter is effectively frozen when IRQs are disabled.  It does still count a little bit but can't wrap around past 1ms.

robert-hh:
>  That's because on stm32 the ticks-us counter is effectively frozen when IRQs are disabled. It does still count a little bit but can't wrap around past 1ms.

To a much lesser extend that should apply to SAMD and MIMXRT port as well. But there  it's the counter roll-over interrupt, which happens rarer.

tests/extmod/machine_disable_irq.py
```diff
@@ -0,0 +1,26 @@
...
+
+# Structured to also nest disable_irq() calls
+def f(n):
+    st_irq = disable_irq()
+    if n:
+        f(n - 1)
+    else:
+        # busy-wait in a tight loop for 1ms, to simulate doing some work in a critical section
+        t0 = ticks_us()
+        while ticks_diff(ticks_us(), t0) < 1000:
```
dpgeorge, line 18: I suggest just making this a fixed-iteration loop calling `ticks_us()` each time (so it actually does some work).  Eg:
```py
for _ in range(100):
    ticks_us()
```
That should still trigger the original bug because there are more than 32 iterations, so the VM will try to release the GIL.

Then separately (one day) we could add a specific test for using `ticks_us()` while interrupts are disabled, and specify which ports that actually works on, and maybe fix other ports.

projectgus: Argh, of course!

Have implemented the suggestion, confirmed that the test still fails on ESP32-C3 on master and passes with the fix. Pyboard V1.1 also pases this test now.

dpgeorge (approved):

tests/extmod/machine_disable_irq.py
dpgeorge: Thanks for updating.  I also tested on PYBD-SF2 and it works now.

PR #15911, opened 2024-09-25
extmod: Implement a minimal typing module.

stinos:
### Summary

The compiler already ignores type hints but to improve runtime support a minimal `typing` (and `abc` and `typing_extensions`) module is implemented in C because code-size wise that is the cheapest to do. Also see https://github.com/micropython/micropython-lib/pull/584 for background.

Note @Josverl asked to support `__future__` and `collections.abc` as well but I did not (yet) do that because:
- `__future__` is easy to add as an alias but the result would be that anything you ask is possible,i.e. `if __future__.just_inventing_something:` would pass. Unclear if we want that.
- I'm not sure how to implement `collections.abc` correctly in the presence of micropython-lib's `collections` module

### Testing

A basic test is added.

- a Python implementation; see https://github.com/micropython/micropython-lib/pull/584, and the Python code mentioned in modtyping.c. Works as well, clearer, but even when frozen the code size is larger.
- ignoring typing-related imports in the compiler: might even be smaller in code size, possibly faster as well, but it doesn't practically seem possible to make type aliases (`Nested = Iterable[Tuple[MyAlias, ...]]`) work because the compiler cannot distinguish them from normal Python code.
- ignoring everything typing-related in the compile as long as it's compatible with Python 3.12: since Python 3.12 got support for the [type statement](https://docs.python.org/3/reference/simple_stmts.html#type) which is easy to distinguish so the compiler could ignore that, and hence type aliases using that syntax become possible. I'd say this is the way forward, but it's just too soon for that.

Note I really have no idea how many people would actually use this so maybe makes more sense to disable this by default in all but the PC ports.

Also note that `dir(typing)` or REPL completion on `typing.` will list _all_ qstrs :)

### DIFF
--- a/extmod/modtyping.c
+++ b/extmod/modtyping.c
@@ -1,0 +1,157 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2024 Damien P. George
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "py/obj.h"
+
+#if MICROPY_PY_TYPING
+
+// Implement roughly the equivalent of the following minimal Python typing module, meant to support the
+// typing syntax at runtime but otherwise ignoring any functionality:
+//
+//  class _AnyCall:
+//    def __init__(*args, **kwargs):
+//        pass
+//
+//    def __call__(self, *args, **kwargs):
+//        return self
+//
+//    def __getitem__(self, attr):
+//        return self
+//
+//  _typing_obj = _AnyCall()
+//
+//  def __getattr__(attr):
+//      return _typing_obj
+//
+// Note this works together with the micropython compiler itself ignoring type hints, i.e. when encountering
+//
+// def hello(name: str) -> None:
+//    pass
+//
+// both str and None hints are simply ignored.
+
+typedef struct _mp_obj_any_call_t
+{
+    mp_obj_base_t base;
+} mp_obj_any_call_t;
+
+extern const mp_obj_type_t mp_type_any_call_t;
+
+
+// Can be used both for __new__ and __call__: the latter's prototype is
+// (mp_obj_t self_in, size_t n_args, size_t n_kw, const mp_obj_t *args)
+// so this function works as long as the argument size matches.
+static mp_obj_t any_call_new(const mp_obj_type_t *arg0, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    #if MICROPY_OBJ_REPR != MICROPY_OBJ_REPR_D
+    MP_STATIC_ASSERT(sizeof(mp_obj_type_t *) == sizeof(mp_obj_t));
+    #endif
+    // If it's an actual instance we're used for __call__ so return self_in.
+    if (mp_obj_get_type(MP_OBJ_FROM_PTR(arg0)) == &mp_type_any_call_t) {
+        return MP_OBJ_FROM_PTR(arg0);
+    }
+    // Could also be we're being called as a function/decorator, return the decorated thing then.
+    // TODO obviously a bit of a hack, plus doesn't work for decorators with arguments.
+    // Note could test mp_obj_is_fun on the first arg here, then being called as decorator etc that
+    // is true, but turns out just returning the last argument works in more cases, like
+    // UserId = typing.NewType("UserId", int)
+    // assert UserId(1) == 1
+    if (n_args != 0) {
+        return args[n_args - 1];
+    }
+    return mp_obj_malloc(mp_obj_any_call_t, arg0);
+}
+
+#if MICROPY_OBJ_REPR == MICROPY_OBJ_REPR_D
+static mp_obj_t any_call_call(mp_obj_t self_in, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    if (mp_obj_get_type(self_in) == &mp_type_any_call_t || n_args == 0) {
+        return self_in;
+    }
+    return args[n_args - 1];
+}
+#else
+#define any_call_call any_call_new
+#endif
+
+#if defined(MICROPY_UNIX_COVERAGE)
+void any_call_attr(mp_obj_t self_in, qstr attr, mp_obj_t *dest);
+#endif
+
+static mp_obj_t any_call_subscr(mp_obj_t self_in, mp_obj_t index_in, mp_obj_t value) {
+    return self_in;
+}
+
+// TODO could probably apply same trick as in any_call_new here and merge any_call_module_attr,
+// but still have to test if that's worth it code-size wise.
+static void any_call_attr(mp_obj_t self_in, qstr attr, mp_obj_t *dest) {
+    // Only loading is supported.
+    if (dest[0] == MP_OBJ_NULL) {
+        if (attr != MP_QSTR___str__ && attr != MP_QSTR___repr__) {
+            dest[0] = self_in;
+        }
+    }
+}
+
+void any_call_module_attr(mp_obj_t self_in, qstr attr, mp_obj_t *dest) {
+    // Only loading is supported.
+    if (dest[0] == MP_OBJ_NULL) {
+        dest[0] = MP_OBJ_FROM_PTR(&mp_type_any_call_t);
+    }
+}
+
+static MP_DEFINE_CONST_OBJ_TYPE(
+    mp_type_any_call_t,
+    MP_QSTR_any_call,
+    MP_TYPE_FLAG_NONE,
+    make_new, any_call_new,
+    attr, any_call_attr,
+    subscr, any_call_subscr,
+    call, any_call_call
+    );
+
+
+static const mp_rom_map_elem_t mp_module_typing_globals_table[] = {
+    { MP_ROM_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_typing) },
+};
+
+static MP_DEFINE_CONST_DICT(mp_module_typing_globals, mp_module_typing_globals_table);
+
+const mp_obj_module_t mp_module_typing = {
+    .base = { &mp_type_module },
+    .globals = (mp_obj_dict_t *)&mp_module_typing_globals,
+};
+
+// Extensible such that a typing module implemented in Python still has priority.
+MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR_typing, mp_module_typing);
+MP_REGISTER_MODULE_DELEGATION(mp_module_typing, any_call_module_attr);
+
+
+#if MICROPY_PY_TYPING_EXTRA_MODULES
+MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR_abc, mp_module_typing);
+MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR___future__, mp_module_typing);
+MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR_typing_extensions, mp_module_typing);
+#endif
+
+#endif // MICROPY_PY_TYPING
--- a/extmod/modtyping.c
+++ b/extmod/modtyping.c
@@ -58,7 +58,7 @@ typedef struct _mp_obj_any_call_t
     mp_obj_base_t base;
 } mp_obj_any_call_t;
 
-extern const mp_obj_type_t mp_type_any_call_t;
+static const mp_obj_type_t mp_type_any_call_t;
 
 
 // Can be used both for __new__ and __call__: the latter's prototype is
--- a/extmod/modtyping.c
+++ b/extmod/modtyping.c
@@ -25,12 +25,14 @@
  */
 
 #include "py/obj.h"
+#include "py/runtime.h"
 
 #if MICROPY_PY_TYPING
 
 // Implement roughly the equivalent of the following minimal Python typing module, meant to support the
 // typing syntax at runtime but otherwise ignoring any functionality:
 //
+//  TYPE_CHECKING = False
 //  class _AnyCall:
 //    def __init__(*args, **kwargs):
 //        pass
@@ -59,6 +61,78 @@ typedef struct _mp_obj_any_call_t
 } mp_obj_any_call_t;
 
 static const mp_obj_type_t mp_type_any_call_t;
+static const mp_obj_type_t mp_type_typing_alias;
+
+// Lightweight runtime representation for objects such as typing.List[int].
+// The alias keeps track of the original builtin type and the tuple of
+// parameters so that __origin__ and __args__ can be queried at runtime.
+typedef struct _mp_obj_typing_alias_t {
+    mp_obj_base_t base;
+    mp_obj_t origin;
+    mp_obj_t args; // tuple or MP_OBJ_NULL when not parametrised
+} mp_obj_typing_alias_t;
+
+// Maps a qstr name to the builtin type that should back the alias.
+typedef struct {
+    qstr name;
+    const mp_obj_type_t *type;
+} typing_alias_spec_t;
+
+static mp_obj_t typing_alias_from_spec(const typing_alias_spec_t *spec_table, size_t spec_len, qstr attr);
+
+static mp_obj_t typing_alias_new(mp_obj_t origin, mp_obj_t args) {
+    mp_obj_typing_alias_t *self = mp_obj_malloc(mp_obj_typing_alias_t, &mp_type_typing_alias);
+    self->origin = origin;
+    self->args = args;
+    return MP_OBJ_FROM_PTR(self);
+}
+
+static void typing_alias_attr(mp_obj_t self_in, qstr attr, mp_obj_t *dest) {
+    // Only handle reads that we recognise: __origin__ and __args__.  Anything
+    // else is delegated back to the VM where it will fall through to the
+    // generic AnyCall behaviour.
+    if (dest[0] != MP_OBJ_NULL) {
+        return;
+    }
+
+    mp_obj_typing_alias_t *self = MP_OBJ_TO_PTR(self_in);
+    if (attr == MP_QSTR___origin__) {
+        dest[0] = self->origin;
+    } else if (attr == MP_QSTR___args__) {
+        dest[0] = self->args == MP_OBJ_NULL ? mp_const_empty_tuple : self->args;
+    }
+}
+
+static mp_obj_t typing_alias_subscr(mp_obj_t self_in, mp_obj_t index_in, mp_obj_t value) {
+    if (value != MP_OBJ_SENTINEL) {
+        mp_raise_TypeError(MP_ERROR_TEXT("typing alias does not support assignment"));
+    }
+
+    mp_obj_typing_alias_t *self = MP_OBJ_TO_PTR(self_in);
+    mp_obj_t args_obj;
+    if (mp_obj_is_type(index_in, &mp_type_tuple)) {
+        args_obj = index_in;
+    } else {
+        mp_obj_t items[1] = { index_in };
+        args_obj = mp_obj_new_tuple(1, items);
+    }
+
+    return typing_alias_new(self->origin, args_obj);
+}
+
+static mp_obj_t typing_alias_call(mp_obj_t self_in, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_obj_typing_alias_t *self = MP_OBJ_TO_PTR(self_in);
+    return mp_call_function_n_kw(self->origin, n_args, n_kw, args);
+}
+
+static MP_DEFINE_CONST_OBJ_TYPE(
+    mp_type_typing_alias,
+    MP_QSTR_typing_alias,
+    MP_TYPE_FLAG_NONE,
+    attr, typing_alias_attr,
+    subscr, typing_alias_subscr,
+    call, typing_alias_call
+    );
 
 
 // Can be used both for __new__ and __call__: the latter's prototype is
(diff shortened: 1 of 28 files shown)


stinos:
Test for failing Codecov added.

dhalbert:
We added `from __future__ import annotations` (and nothing else) in CircuitPython a while ago: https://github.com/adafruit/circuitpython/pull/6117. We had to implement it as a native module for the reasons mentioned in the PR.

Josverl:
Agree that currently the most important use case for `__futures__`  is indeed `annotations`. 

I see a lot of downloads of the type stubs, and it's growing steadily. Thr use will only increase when the stubs are integrated with the documentation. So I would prefer this to be enabled on the major ports, and disabled on the memory constrained ports.

For `collections.abc`, could that be a .mpy module that imports the singleton from typing ?

stinos:
> - `__future__` is easy to add as an alias but the result would be that anything you ask is possible,i.e. `if __future__.just_inventing_something:` would pass. Unclear if we want that.

On second thought I don't think `__future__` is ever used like that anyway, so added it as well.


> - I'm not sure how to implement `collections.abc` correctly in the presence of micropython-lib's `collections` module
> For collections.abc, could that be a .mpy module that imports the singleton from typing ?

Essentially, yes. It requires that collections/abc.py is added in micropython-lib and contains just `from typing import __getattr__`. If micropython-lib/collections is not used at all, having the option `MICROPY_MODULE_BUILTIN_SUBPACKAGES` allows `collections.abc` with just one extra entry in collection's globals (see updated implementation).

Likewise `abc` should be removed from micropython-lib or adjusted.

Josverl:
Hi @stinos, I build and tested a RPI_PICO image with 
```c
// enable typing module in C
#define MICROPY_PY_TYPING           (1)
#define MICROPY_PY_TYPING_EXTRA_MODULES (1)
```

I can the same notebook with tests that I used to validate the `.py` versions ( that have some changes compared to your originals) 
and I find that `modtyping.c` and its aliasses do help , but they are significantly less compatible with the typing.mpy implementation.

a short breakdown:
| category | `.mpy` versions| `modtyping.c` | comment
|--------|--------|--------|---
| Pass  | 42| 22 | 
| Fail | 0| 6 | failing runtime checks : isinstance , reveal_type , no_type_check, NewType
| `collection.abc` | 0 | 9 | similar tests to abc, expect that they can be made to pass 
| Unsupported, python 3.12 Fail | 5| 5 |


of the failing runtime checks I think that `reveal_type` is not that important,
I think 
-  `isinstance`  is a must have 
- `NewType` and `@no_type_check` would be very good to have.


for complete test results see : [verify_type_checking.ipynb](https://github.com/Josverl/rt_typing/blob/main/verify_type_checking.ipynb)

for the tests that have different results between CPython , the .mpy amd the .c implementation I have added the results in 
comments. One such example is : 
``` py
# %%micropython --reset
from typing import TYPE_CHECKING


from abc import ABC, abstractmethod
from math import pi


class Shape(ABC):
    @abstractmethod
    def get_area(self) -> float:
        pass

    @abstractmethod
    def get_perimeter(self) -> float:
        pass


class Circle(Shape):
    def __init__(self, radius) -> None:
        self.radius = radius

    def get_area(self) -> float:
        return pi * self.radius**2

    def get_perimeter(self) -> float:
        return 2 * pi * self.radius


class Square(Shape):
    def __init__(self, side) -> None:
        self.side = side

    def get_area(self) -> float:
        return self.side**2

    def get_perimeter(self) -> float:
        return 4 * self.side


c1 = Circle(5)
s1 = Square(5)

for shape in [c1, s1]:
    a = shape.get_area()
    p = shape.get_perimeter()

    print(a, p)
    print(f"{type(a)=}")
    print(f"{type(p)=}")

    assert isinstance(a, (float, int)), "Area should be a float"
    assert isinstance(p, (float, int)), "Perimeter should be a float"

# CPython
# ----------------------
# 78.53981633974483 31.41592653589793
# type(a)=<class 'float'>
# type(p)=<class 'float'>
# 25 20
# type(a)=<class 'int'>
# type(p)=<class 'int'>

# MicroPython typing.py
# ----------------------
# 78.53982 31.41593
# type(a)=<class 'float'>
# type(p)=<class 'float'>
# 25 20
# type(a)=<class 'int'>
# type(p)=<class 'int'>

# Micropython - modtyping.c
# ----------------------
# <any_call> <any_call>
# type(a)=<class 'any_call'>
# type(p)=<class 'any_call'>
# WARNING |Traceback (most recent call last):
# WARNING |  File "<stdin>", line 52, in <module>
# ERROR   |AssertionError: Area should be a float
#

```

stinos:
Great, thanks for testing! Do you mind checking the code size the .c implementation takes (pico isn't included in the ci build code size report)? Thing is still: the closer it gets to just freezing the .py implementations (and for whatever reason on the unix build it seems already way over it), the less sense it makes to follow this path.

> for complete test results see : [verify_type_checking.ipynb](https://github.com/Josverl/rt_typing/blob/main/verify_type_checking.ipynb)

I think that repository is private since I get a 404 trying to access it.


> isinstance is a must have

Well, it's a bit more than just 'isinstance does not work', this is *not* good.. I didn't even test that because I had no idea the ability to derive from a native type could even be broken and assumed it would first look in the derived class for attributes and only then in the base class. 


> collection.abc

The implementation is likely not active; either you have to make sure there's no `collections` module like the one from micropython-lib is in your module search path and `#define MICROPY_MODULE_BUILTIN_SUBPACKAGES ( 1 )` or you need a collections/abc.py which has a `from typing import __getattr__`

Josverl:
- 404 🤐, indeed private , now public. I assumed it was all along
- While adding the test cases I did many iterations on trying to match the CPython runtime behavior (within reason) 
   for `abc` the simplest I have found is [`from typing import __Ignore as ABC`](https://github.com/Josverl/rt_typing/blob/main/mp/abc.py) and same in in `collections.abc`
- there are a few more tweaks 
( I use `__Ignore `  as I find that a bit clearer than AnyCall and is saves a few bytes compared to `_TypingIgnore`


**Firmware size.**
Im not quite sure which figures I should compare,
and when I compare the size of the added frozen modules (547 bytes total)  , to the growth of the firmware.elf  (892) - it does not match , so Im probably looking at the incorrect number, or there is additional overhead / module 
here is wat I collected: 
```
##### No Typing
 332704       0    5524  338228   52934 /home/builder/micropython/ports/rp2/build-RPI_PICO/firmware.elf

#### C typing
 332704       0    5524  338228   52934 /home/builder/micropython/ports/rp2/build-RPI_PICO/firmware.elf
 
##### .py typing - frozen 

 334048       0    5524  339572   52e74 /home/builder/micropython/ports/rp2/build-RPI_PICO/firmware.elf

##### .py typing - frozen -O3 

 333792       0    5524  339316   52d74 /home/builder/micropython/ports/rp2/build-RPI_PICO_T/firmware.elf

##### .py typing - frozen -O3 - reduced 
no abc but including collections.abc
removed the less useful reveal_type, get_origin, get_args from typing.py

 333596       0    5524  339120   52cb0 /home/builder/micropython/ports/rp2/build-RPI_PICO_T/firmware.elf

```

stinos:
> for abc the simplest I have found is [from typing import __Ignore as ABC](https://github.com/Josverl/rt_typing/blob/main/mp/abc.py) and same in in collections.abc

Yes that works, but you'd have to repeat that for every single type, whereas `from typing import __getattr__` would do that automatically I think?


> Firmware size. Im not quite sure which figures I should compare,

Essentially what you're posted, except that 'No Typing' and 'C typing' have the same size so they're both 'C typing' probably?

Anyway while fixing the implementation (code sample you posted works now) I found a blocker, see your `no_type_check` test for instance: it is not possible to distinguish between any_call (or TypeIgnore for that matter) being called as a decorator/function (because `@decorator` is merely syntactic sugar for `func = decorator(func)`, and because `@decorator(1)` is `func = decorator(1)(func)`) or being called to create a new instance.


So that means our typing implementation, whatever it is, *must* have *all* possible typing-related decorators implemented explicitly (and they definitely do not currently) otherwise the runtime behavior of code using an unlisted decorator is incorrect because the current implementations which make the module's `__getattr__` the fallback will turn the decorated thing into a TypeIgnore instance.

No-one probably really thought this through so far, but this is a rather serious issue.
So serious, that I now think that our current implementations might simply be too dangerous to use (even if we list all decorators now, we'd have to be constantly checking CPythone doesn't introduce new ones). And reverting to an implementation without `__getattr__` but instead listing all types manually, so that at least if something isn't known it cannot run, is the only sane thing to do.

Josverl:
> whereas from typing import __getattr__ would do that automatically I think?

No `__getattr__`  will produce the instance `__ignore` rather than the Type  `___Ignore`
we could add logic though to check for Uppercase in the name, but a few classes/types may use lowercase
and yes , I did think that through , hence the testing 🤔

>  'No Typing' and 'C typing' have the same size so they're both 'C typing' probably?

yeah, noticed that, may have fumbled the definitions, here are numbers for Variants I created of the   PIMORONI_PICOLIPO-FLASH
```
 330824       0    5524  336348   521dc /home/jos/micropython/ports/rp2/build-PIMORONI_PICOLIPO-FLASH_16M/firmware.elf
 331096       0    5524  336620   522ec /home/jos/micropython/ports/rp2/build-PIMORONI_PICOLIPO-TYPING_C/firmware.elf
 331968       0    5524  337492   52654 /home/jos/micropython/ports/rp2/build-PIMORONI_PICOLIPO-TYPING_PY/firmware.elf 
```

- .c increase = 272 
- .mpy increate = 1.144 for a .mpy payload increase of 700

The .c implementation , even if it is for part of the functionality, and currently does not pass all tests, is significantly lower.

.mpy Package size breakdown : 
| module| size|
|--------|--------|
| __future__.mpy| 50|
| abc.mpy| 70|
| typing_extensions.mpy| 66|
| typing.mpy | 394 |
| collections/__init__.mpy| 38| 
| collections/abc.mpy| 82 | 
| **TOTAL**| 700 | 
| _mysterious overhead_| 444| 


> implementations might simply be too dangerous to use

I think the risk is acceptable, as with the implementation of the typing @decorators, we are actually 'unbreaking' code, rather than breaking it.
And to further reduce the risk, I think the best ways it to have the typing modules as: 
optional , 
easy to add (mpremote mip installable) python packages.
even at the cost of 700 or 

The way I have been using `__getattr__ `and `__Ignore `aka `_Any_Call` are to optimize for the size, to reduce from the 1.685 bytes for just `typing.mpy` otherwise.
But fully agree that we [need some guardrails ](https://github.com/Josverl/rt_typing/blob/869a1861ee14fdcb7bbfe98477fd1edc568c2f24/mp/typing.py#L86-L96)to protect against unwanted scenarios and tests to validate these, even for an opt-in approach.

Josverl:
>  > easy to add (mpremote mip installable) python packages,
> 
> Ok but if that is a hard requirement, the C implementation is a no-go?

The general rule in MicroPython is that if something can be reasonable implemented in Python, then it should.  that is why there are several `_something` in .c combined with `something.py` modules in micropython.lib 
That is actually also true for the CPython version of the `_typing` and `typing`  module(s) .

As static and runtime typing in Python is still rapidly developing, my worry is that a size optimized C-implementation will be harder to maintain and extend, than a Python based solution. Also the mip installable approach adds a degree of freedom, as it can be installed on (almost) any port/board/family of micropython.

perhaps we need to rank/ weight what is, or we find, important, in order to make a good-enough choice wrt to the implementation:
- type checking: support type annotations [ list of typing features] with documented exceptions, verified with pyright and mypy  
- runtime : adherance to CPython runtime behaviour , with documented exceptions 
- flash size overhead 
- maintanability 
-

Josverl:
@stinos, 
I have some 30+ testcases for MicroPython runtime typing that I think make sense adding to this PR. 
Each is a short snippet of code with some type annotations that should run without errors (if the modules are available) 
If your interested in adding these, let me know.

extmod/modtyping.c
```diff
@@ -0,0 +1,157 @@
...
+//    pass
+//
+// both str and None hints are simply ignored.
+
+typedef struct _mp_obj_any_call_t
+{
+    mp_obj_base_t base;
+} mp_obj_any_call_t;
+
+extern const mp_obj_type_t mp_type_any_call_t;
```
Josverl: should this not be `static const mp_obj_type_t mp_type_any_call_t;` ?

stinos: It is possible I made it extern with the idea that would allow using this in other C code for extending typing stuff. Don't remember, and currently not needed, so indeed can be static.

stinos:
>  I have some 30+ testcases for MicroPython runtime typing that I think make sense adding to this PR.

Yes we definitely need tests. In fact if I remember correctly I simply stopped going further with this PR because I figured that without a rather complete test suite it's all a bit useless. For example there might be a wanted feature which is actually impossible to implement in and then we should know that and because it will also make it easier to decide between the various possible solutions (this vs your implementation vs Andrew's) - or perhaps some combination thereof depending on what is more important for the user.

Now, because the typing module is quite large it can make it pretty hard to figure out if everything (i.e., from CPython) is covered.

So my idea was that the test code should closely follow the [CPython docs](https://docs.python.org/3/library/typing.html#module-contents) and would basically duplicate most code listed there (no idea if allowed license-wise but perhaps manually typing a code sample which essentially replicates the original might just be fine?) and leave comments for pieces which aren't supported (but, like your code, can document that in cpydiff as well).

I realize that might be some work, but I have a feeling Python's typing is still in active development and I'm afraid otherwise things might get really hard to follow. And since from the earlier comments it sounds as if there might be some caveats with potentially dangerous behavior, I don't feel very comfortable with this PR unless it's very clear what is ok and what is not.

Or does this sound like overkill?

<br/><br/>

> As static and runtime typing in Python is still rapidly developing, my worry is that a size optimized C-implementation will be harder to maintain and extend, than a Python based solution.

It definitely is. I checked my last incantation (was never pushed) and it's like magic. At least needs more code comments.


>  Also the mip installable approach adds a degree of freedom, as it can be installed on (almost) any port/board/family of micropython.
> perhaps we need to rank/ weight what is, or we find, important, in order to make a good-enough choice wrt to the implementation:

I briefly went over all comments and all things considered find it still hard to make a decision or even rank. Yet I think it might simply come down to the code size again since the difference seems substantial.

Josverl:
> I'm not sure how to implement `collections.abc` correctly in the presence of micropython-lib's collections module

I think that could be done by adding this to `collections/__init__.py` on micropython-lib :

```py
# avoid overwriting built-in collections.abc typing dummy module
try:
    import typing as abc
    import sys
    sys.modules['collections.abc'] = abc
except ImportError:
    pass
```

not sure how to register it in modcollections.c though

I currently get an import error 
```
$ ./ports/unix/build-standard/micropython -c "import collections;import collections.abc"
ImportError: no module named 'collections.abc'
```

**Testing**
> the test code should closely follow the [CPython docs](https://docs.python.org/3/library/typing.html#module-contents) and would basically duplicate most code listed there (no idea if allowed license-wise but perhaps manually typing a code sample which essentially replicates the original might just be fine?) and leave comments for pieces which aren't supported (but, like your code, can document that in cpydiff as well).

I have indeed been trawling the CPython docs and adding that as test material for the [runtime tests](https://github.com/Josverl/micropython/blob/builtintypingmodule/tests/extmod/typing_runtime.py) , as well as the [cpy-diffs](https://github.com/micropython/micropython/commit/119a2ad40009c184d1eb37cbba85d969f54dcd08).
Perhaps we should that per python version, or at least mention the relevant Python version / feature supported for runtime.

>  I have a feeling Python's typing is still in active development a

It is, but as long as there will be new python versions planned, typing will never be "Completed".
`typing` is in the `stdlib` though  and is definitely getting more and more traction, while still being optional to use. 
And that optionality should be the same for MicroPython

> I realize that (creating tests) might be some work.
I've been at this for a few years, so agree, however we do not need to start from scratch.


~~If you can merge to PR to this PR with the tests (not sure if there is a simpler way) ,~~ 
then we can build out the test suite 
- and use that to: 
- get clarity on what works / fails
- compare the different implementations against the same tests
- decide whether or not any implementation is good enough for

Josverl:
I have been working on adding tests based on the [list of typing PEPs](https://peps.python.org/topic/typing/)
Starting from Python 3.5 up to 3.8 

**Tests:**
* copied the code samples for these PEPS to individual test documents
* added try/excepts to print things that need fixing  `FIXME :` 
**modtyping.c**
* added type alias for type subscriptions `List[int]` , `Literal`, `Dict`, `Tuple`, `Set` and `FrozenSet` 
* added `typing.TYPE_CHECKING = False`

# Current State 
## Test state: 
pass  extmod/typing_pep_0484.py
pass  extmod/typing_pep_0526.py
FAIL  extmod/typing_pep_0544.py
pass  extmod/typing_pep_0560.py
pass  extmod/typing_pep_0586.py
FAIL  extmod/typing_pep_0589.py
FAIL  extmod/typing_pep_0591.py
FAIL  extmod/typing_runtime.py
pass  extmod/typing_syntax.py
9 tests performed (129 individual testcases)
5 tests passed
4 tests failed: typing_pep_0544 typing_pep_0589 typing_pep_0591 typing_runtime

## cpy-diffs 
6 cpydiffs test defined for missing / different functionality
 - User Defined Generics
 - get_args()
 - get_origin()
 - reveal_type()
 - runtime_checkable()
 - TypeDict checks 

## big item issues:
- [ ] importing from collections.abc rather than from abc 
- [ ] User Defined Generic Classes not supported in Micropython due to the lack of metaclasses support
      https://peps.python.org/pep-0544/#generic-protocols


### Miscellaneous
typing.Callable, ParamSpec
- [ ] FIXME: from collections.abc import Callable

typing.get_origin()
- [ ] FIXME: cpy_diff - get_origin(str) should be None
typing.get_args()
- [ ] FIXME: cpy_diff - get_args(int) should be ()

# Python 3.8

### PEP 544 - Protocols: Structural subtyping (static duck typing)
Merging and extending protocols: 
- [ ] FIXME: Difference or Crash - multiple bases have instance lay-out conflict: multiple bases have instance lay-out conflict
- [ ] FIXME: Difference or Crash - multiple bases have instance lay-out conflict: multiple bases have instance lay-out conflict


### PEP 589 - TypedDict: Type hints for dictionaries with a fixed set of keys

Totality and optional keys
- [ ] FIXME: cpy_diff : `total` parameter not supported by runtime TypedDict implementation: function doesn't take keyword arguments
- [ ] FIXME: cpy_diff : `total` parameter not supported by runtime TypedDict implementation: function doesn't take keyword arguments

Inheritance and required/optional mix
- [ ] FIXME: cpy_diff : `total` parameter not supported by runtime TypedDict implementation for Point3D: function doesn't take keyword arguments

Runtime checks: TypedDict cannot be used with isinstance/class checks
- [ ] FIXME: TypedDict class allowed in isinstance (unexpected)

Alternative functional syntax and constructors
- [ ] FIXME: cpy_diff Functional TypedDict constructor call failed at runtime (expected): 'dict' object isn't callable

Totality and mixing with Required/NotRequired
- [ ] FIXME: cpy_diff - total parameter not supported by runtime TypedDict implementation for MovieMix: function doesn't take keyword arguments

extra_items and closed examples
- [ ] FIXME: extra_items not supported by runtime typing implementation: function doesn't take keyword arguments
- [ ] FIXME: closed parameter not supported by runtime typing implementation: function doesn't take keyword arguments

Interaction with Mapping and dict conversions
- [ ] FIXME: extra_items not supported by runtime typing implementation function doesn't take keyword arguments


### PEP 0591 - Final qualifier for types
The Final annotation
Semantics and examples
- [ ] FIXME: Final cannot be used with container types

Josverl:
I reworked the python version of the typing module to reduce the impact to the firmware size while freezing, ( See linked PR above) and ran the same tests against both.
 
To my surprise there is not much difference in code size between for the unix port ( though I could be measuring the wrong thing) 
but I expect the .C approach could be optimized further.

stinos:
I rebased this for completeness. But as noted here and in related issues/pull requests, for me personally the blocker is that it's easy to break runtime behavior when doing anything more than annotating (`: SomeType`): decorating with or deriving from types from the typing module is a mess I still need to look into.

AJMansfield:
> deriving from types from the typing module is a mess I still need to look into.

Indeed, dealing with this sort of thing...:
```python
T = TypeVar('T')
L = list[T] # lowercase list
G = Generic[T]
class XX(L,G):
    ...
```

I've already got reasons to want to refactor the whole call tree under `__build_class__` even without trying to deal with that, though --- specifically, to deal with the inheritance-kwargs situation for an implementation of PEP487's `__init_subclass__`, since at present in my prototype implementation the kwargs to `__build_class__` end up needing to be copied and reformatted like three different times between the two different formats used for passing kwargs to C functions, in order to relay them all the way to `mp_obj_new_type` and present them to the actual `mp_call_method_n_kw` calls to user-defined `__init_subclass__` methods.

Adding actual support for PEP560 `__class_getitem__` doesn't feel like the wildest idea, either. There's a few ways of doing it I can think of, but the simplest might be to just add a case to `type`'s `type_attr` method to convert a lookup of `__getitem__` into a class-member lookup of `__class_getitem__`.

I don't think there's a small and efficient way to make lowercase `list[whatever]` and `DerivedFromList[whatever]` just idempotently return `list` and `DerivedFromList` while making `int[whatever]` and `OrdinaryClass[whatever]` still error, but I don't think MicroPython needs to enforce that; we can just provide a blanket idempotent `__class_getitem__` on `type` itself.

And then for the typing objects themselves, assuming there's a cheap test this can add for whether or not any given object is a type-hint object, we could have `mp_obj_new_type` filter them out entirely while  processing through the list of base classes (i.e. as it's setting inherited flags and copying the base class list into the class's struct); or perhaps make `__Ignore` a self-deleting base class by adding an appropriate `__init_subclass__`.

AJMansfield:
> * ignoring everything typing-related in the compile as long as it's compatible with Python 3.12: since Python 3.12 got support for the [type statement](https://docs.python.org/3/reference/simple_stmts.html#type) which is easy to distinguish so the compiler could ignore that, and hence type aliases using that syntax become possible. I'd say this is the way forward, but it's just too soon for that.

Honestly, type statements absolutely feel like the sort of thing that should've already become a `from __future__ import type` thing on earlier CPythons too --- with how the whole point was to help address all the version compatibility problems that CPython has even just with itself when you use type annotations. Maybe worth proposing, so it can become the way forward sooner rather than later.

Josverl:
As there are a lot of flavors of Typing, and the [Discussions](https://discuss.python.org/c/typing/32) are quite active. 
there are some ideas on introducing a mini DSL specifically for typing - but not sure if that would be a help or a hindrance.
I think the only option is to consider the typing PEP-xxxx features one by one - and possibly the logical order is by ascending CPython version. 
https://discuss.python.org/c/typing/32

Another thing I notice is that the typing community has a strong preference to implement typing code in Python where possible C is just too much work for too few people, as C-Devs < Py-Devs. I think for MicroPython that is no different. 

Im all for implementing a minimal `_typing` module in C where optimization is needed / desired, and keep the rest as a mip installable `typing.mpy` in micropython-lib

stinos:
Btw I started adding/enabling m,ore tests to get a clearer picture of what is needed: https://github.com/stinos/micropython/tree/typingtests

Josverl:
I have included test results for a rebased version of this PR in the reports in the test matrix that I have run with the updated typing tests.

This PR is currently the smallest ( not counting mip or ROMFS installed) , but does not pass all the typing tests , although I am sure that it is possible to match the same quality with additional work.

| Variant | PASS | xfail | skip | FAIL | False Positive | ERROR | total | text | data | bss | size | Δsize |
|:---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| standard | 3 | 1 | 183 | 0 | 0 | 0 | 187 | 784_958 | 70_904 | 2_960 | 858_822 | ref |
| 1_mp-stubs | 122 | 29 | 36 | 0 | 0 | 0 | 187 | 790_390 | 72_056 | 2_960 | 865_406 | +6_584 |
| 2a_typing | 107 | 27 | 53 | 0 | 0 | 0 | 187 | 786_830 | 71_352 | 2_960 | 861_142 | +2_320 |
| 2b_bundle_typing_S | 122 | 29 | 36 | 0 | 0 | 0 | 187 | 787_614 | 71_512 | 2_960 | 862_086 | +3_264 |
| 2c_bundle_typing_L | 127 | 29 | 31 | 0 | 0 | 0 | 187 | 788_486 | 71_704 | 2_960 | 863_150 | +4_328 |
| **3_builtintypingmodule** | 128 | 31 | 8 | 16 | 2 | 2 | 187 | 786_870 | 71_352 | 2_960 | 861_182 | +2_360 |
| 5_standard_mip_installed | 152 | 33 | 2 | 0 | 0 | 0 | 187 | 784_958 | 70_904 | 2_960 | 858_822 | +0 |
| 6_coverage | 152 | 33 | 2 | 0 | 0 | 0 | 187 | 1_844_803 | 353_696 | 264_608 | 2_463_107 | +1_604_285 |
| RPI_PICO2 | 152 | 33 | 2 | 0 | 0 | 0 | 187 | - | - | - | - | - |
| ESP32_C5 | 122 | 29 | 36 | 0 | 0 | 0 | 187 | - | - | - | - | - |

Legend:
 - **ok**=passed, **xfail**=expected failure, **skip**=test skipped
 - **FAIL**=test failed, **False Positive**=unexpected success
 - **ERROR**=test errored (also used when the test module crashed).
 - Firmware sizes are reported in bytes from `size <binary>`.
 
 for the full report see : micropython/micropython-lib#1051

PR #15914, opened 2024-09-26
rp2: Fix dropped UART REPL bytes on soft reset.

projectgus:
### Summary

Necessary to fix "mpremote run" over hardware UART (i.e. when `MICROPY_HW_ENABLE_UART_REPL` is set).

Bisect shows bug was introduced by d420b4e4 (#13718), but looks like made more complex by 01c046d2 (#14041). Specifically: resetting and re-initialising the REPL UART during soft reset clears the FIFO before it's done printing the "MPY: soft reboot" line.

Fixed by adding a UART TX flush in the deinit path.

### Testing

1. Build rp2 port with `MICROPY_HW_ENABLE_UART_REPL` set.
2. Open interactive REPL and type Ctrl-D. Note the "MPY: soft reboot" line is not printed, only one invalid character is received.
3. Alternatively, try to use "mpremote run" over the hardware UART. It times out waiting for the "soft reboot" line.

Fails without this PR, works with this PR.

* Alternative would be to not reset the REPL UART on soft reset. This may be better from the "principle of least surprise" point of view (i.e. if the user hasn't changed the UART settings then it makes no difference, and if they have changed them then having Soft Reboot change them and lose the REPL seems a bit unexpected.) However doing it that way has less clean layering, as the machine.UART code would need to know about the REPL configuration.

### DIFF
--- a/ports/rp2/machine_uart.c
+++ b/ports/rp2/machine_uart.c
@@ -458,6 +458,7 @@ static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_arg
 }
 
 static void mp_machine_uart_deinit(machine_uart_obj_t *self) {
+    uart_tx_wait_blocking(self->uart); // Flush TX FIFO if necessary
     uart_deinit(self->uart);
     if (self->uart_id == 0) {
         irq_set_enabled(UART0_IRQ, false);


dpgeorge (approved):
Thanks for the fix.  I verified the bug and that this PR fixes it.

It's true that the alternative would be to not reset the UARTs and leave them as they are for a soft reset.  That's not too difficult, considering they already have static state.  The only thing that really needs deiniting is their IRQ handlers because the Python callback is freed on soft reset.

Maybe that's a nice change, not to deinit UARTs, but it's a much bigger change (and stm32 for example does reset all UARTs on soft reset).

So let's go with this fix!

dpgeorge:
> The only thing that really needs deiniting is their IRQ handlers because the Python callback is freed on soft reset.

Actually, no, the RX and TX buffers are only on the GC heap.  So the UART really must be deinited on soft reset.

projectgus:
I guess I was thinking to still deinit the machine.UART objects (including the IRQs and the buffers) but leave the hardware UART peripheral setting alone if that UART is also used for REPL (i.e. the part implemented in rp2/uart.c which AFAIK doesn't interact with any of the machine module code.)

This still seems cleaner overall, though.

robert-hh:
The change looks good for me. For a moment I had doubts that using `uart_tx_wait_blocking()` was suitable, because is only looks at the FIFO. But as long as IRQ is enabled, the FIFO will eventually get refilled from the TX buffer.

robert-hh (approved):

dpgeorge (approved):

Issue #7179, opened 2021-04-28
Missing country settings for ESP WIFI 


karfas:
In micropython, there is no way to define in which country the ESP WIFI transmitters operate.
This might create legal problems for anyone using the ESP devices and micropython "out of the box".

The ESP-IDF function call is esp_wifi_set_country(), described in https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/wifi.html#wi-fi-country-code.

Current uPy sources have no call to esp_wifi_set_country() - I searched *.[hc] for it.

The issue had been discussed in https://forum.micropython.org/viewtopic.php?f=18&t=10388

andypiper:
Interestingly, per #8417 the country option is available in wifi on the pyboard / stm port, but apparently not exposed in the ESP port as of now.

jonnor:
This still seems to be relevant. Anyone wants to make a MR to add it?

PR #15916, opened 2024-09-26
esp32/network_wlan: Get and set the WLAN country code.

agatti:
### Summary

This provides an implementation for #7179, by allowing to set (and get) the country code for the ESP32's WLAN radio interface.

No user-visible API changes have been introduced as this uses `network.country` rather than having its own extra method somewhere else (I initially thought about bolting it onto `network.WLAN` to minimise code changes, but in for a penny, in for a pound).  Ports can now properly set the WLAN country code if they have the APIs for it, rather than just keeping track of the code itself.

There are a couple of minor things one should be aware of when using this on an ESP32, and those have been added as notes to `network.country`'s documentation section:

* ESP-IDF unconditionally stores the chosen country code into the device's non-volatile storage
* setting the country code with no WLAN interface ever being instantiated will raise an exception
* the country code will default to "worldwide" if it was never set in the device or if no WLAN interface was instantiated yet.

### Testing

Testing was done manually on an ESP32C3 using ESP-IDF 5.0.4 and via `tests/ports/esp32/country_wlan.py`.  This is tricky to write a reliable test for because the board's WLAN interface may not be in a known state when the test runs.

Right now the test only passes if it is run when no WLAN has been instantiated before and if the WLAN country code is set to either `XX` (the default when the flash is cleared) or to `AU`.  I was tempted to put `IT` there instead but I managed to resist :)

The provided test cycles between those two known country codes, so it is still independent from the existing device configuration, as long as the first test run was done with an empty NVS.

This feature slightly increases the code size so maybe it can be put behind a define of its own, if setting the WLAN country code is not needed.  The existing implementation doesn't do anything interesting for ESP32, so some code space was already lost to begin with, which makes the extra space taken by this feature a bit easier to let go.

### DIFF
--- a/extmod/modnetwork.h
+++ b/extmod/modnetwork.h
@@ -149,4 +149,6 @@ void mod_network_register_nic(mp_obj_t nic);
 mp_obj_t mod_network_find_nic(const uint8_t *ip);
 #endif
 
+mp_obj_t mp_network_country(size_t n_args, const mp_obj_t *args);
+
 #endif // MICROPY_INCLUDED_MODNETWORK_H
--- a/ports/esp32/network_wlan.c
+++ b/ports/esp32/network_wlan.c
@@ -82,6 +82,7 @@ static bool mdns_initialised = false;
 
 static uint8_t conf_wifi_sta_reconnects = 0;
 static uint8_t wifi_sta_reconnects;
+static uint8_t network_country_code_set = 0;
 
 // The rules for this default are defined in the documentation of esp_wifi_set_protocol()
 // rather than in code, so we have to recreate them here.
@@ -262,6 +263,19 @@ static mp_obj_t network_wlan_make_new(const mp_obj_type_t *type, size_t n_args, 
     mp_arg_check_num(n_args, n_kw, 0, 1, false);
 
     esp_initialise_wifi();
+
+    #ifdef MICROPY_PY_NETWORK_COUNTRY
+    char country_code[4] = { '0', '1', ' ', '\0' };
+    if (!(mod_network_country_code[0] == 'X' && mod_network_country_code[1] == 'X')) {
+        country_code[0] = mod_network_country_code[0];
+        country_code[1] = mod_network_country_code[1];
+    }
+    if (network_country_code_set) {
+        // Always follow the AP's country if one is set.
+        esp_exceptions(esp_wifi_set_country_code(country_code, true));
+        network_country_code_set = 0;
+    }
+    #endif
 
     int idx = (n_args > 0) ? mp_obj_get_int(args[0]) : ESP_IF_WIFI_STA;
     if (idx == ESP_IF_WIFI_STA) {
@@ -771,6 +785,58 @@ unknown:
     mp_raise_ValueError(MP_ERROR_TEXT("unknown config param"));
 }
 MP_DEFINE_CONST_FUN_OBJ_KW(network_wlan_config_obj, 1, network_wlan_config);
+
+#ifdef MICROPY_PY_NETWORK_COUNTRY
+mp_obj_t mp_network_country(size_t n_args, const mp_obj_t *pos_args) {
+    // ESP-IDF docs state the country code is three characters long.
+    char country_code[4] = { '0', '1', ' ', '\0' };
+    esp_err_t state;
+
+    // This function can be called at any time once the `network` module is
+    // imported.  Currently the static wifi flags indicate only whether the
+    // radio interface is active or not, whilst we need to know if the
+    // interface is initialised.
+
+    if (n_args == 0) {
+        state = esp_wifi_get_country_code(country_code);
+        if (state != ESP_ERR_WIFI_NOT_INIT) {
+            esp_exceptions(state);
+            if (country_code[0] == '0' && country_code[1] == '1') {
+                mod_network_country_code[0] = 'X';
+                mod_network_country_code[1] = 'X';
+            } else {
+                mod_network_country_code[0] = country_code[0];
+                mod_network_country_code[1] = country_code[1];
+            }
+        }
+
+        return mp_obj_new_str((const char *)mod_network_country_code, 2);
+    }
+
+    if (n_args == 1) {
+        size_t length = 0;
+        const char *code = mp_obj_str_get_data(pos_args[0], &length);
+        if (length < 2) {
+            mp_raise_ValueError(NULL);
+        }
+        mod_network_country_code[0] = code[0];
+        mod_network_country_code[1] = code[1];
+        // Maintain compatibility with the default implementation.
+        if (!(mod_network_country_code[0] == 'X' && mod_network_country_code[1] == 'X')) {
+            country_code[0] = mod_network_country_code[0];
+            country_code[1] = mod_network_country_code[1];
+        }
+        state = esp_wifi_set_country_code(country_code, true);
+        if (state != ESP_ERR_WIFI_NOT_INIT) {
+            esp_exceptions(state);
+        } else {
+            network_country_code_set = 1;
+        }
+    }
+
+    return mp_const_none;
+}
+#endif
 
 static const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_active), MP_ROM_PTR(&network_wlan_active_obj) },
--- a/tests/ports/esp32/network_wlan_country.py
+++ b/tests/ports/esp32/network_wlan_country.py
@@ -1,0 +1,40 @@
+# Test WLAN country code get/set - this must be run before any WLAN
+# instantiation, ideally right after a board reboot.
+
+import network
+
+network.country("XX")
+print(network.country())
+
+network.country("QQ")
+try:
+    network.WLAN(network.STA_IF)
+    print("Failed to raise")
+except OSError as e:
+    print(str(e))
+print(network.country())
+
+network.country("AU")
+print(network.country())
+
+wlan = network.WLAN(network.STA_IF)
+print(network.country())
+
+try:
+    network.country("!!")
+    print("Failed to raise")
+except OSError as e:
+    print(str(e))
+print(network.country())
+
+wlan = network.WLAN(network.AP_IF)
+print(network.country())
+network.country("XX")
+print(network.country())
+
+try:
+    network.country("FP")
+    print("Failed to raise")
+except OSError as e:
+    print(str(e))
+print(network.country())
--- a/docs/library/network.rst
+++ b/docs/library/network.rst
@@ -210,6 +210,12 @@ The following are functions available in the network module.
 
     The default code ``"XX"`` represents the "worldwide" region.
 
+    On ESP32, setting an invalid country code before instantiating the radio
+    interface (ie. calling ``network.WLAN(…)`` the first time) will succeed but
+    then all calls to ``network.WLAN(…)`` will fail until a valid country code is
+    set.  It is best to delay calling ``network.country`` until the radio is
+    initialised first.
+
 .. function:: hostname([name])
 
     Get or set the hostname that will identify this device on the network. It will
--- a/extmod/modnetwork.c
+++ b/extmod/modnetwork.c
@@ -122,6 +122,9 @@ MP_REGISTER_ROOT_POINTER(mp_obj_list_t mod_network_nic_list);
 #endif // MICROPY_PORT_NETWORK_INTERFACES
 
 static mp_obj_t network_country(size_t n_args, const mp_obj_t *args) {
+    #if defined(MICROPY_PY_NETWORK_COUNTRY)
+    return MICROPY_PY_NETWORK_COUNTRY(n_args, args);
+    #else
     if (n_args == 0) {
         return mp_obj_new_str(mod_network_country_code, 2);
     } else {
@@ -134,6 +137,7 @@ static mp_obj_t network_country(size_t n_args, const mp_obj_t *args) {
         mod_network_country_code[1] = str[1];
         return mp_const_none;
     }
+    #endif
 }
 // TODO: Non-static to allow backwards-compatible pyb.country.
 MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mod_network_country_obj, 0, 1, network_country);
--- a/ports/esp32/mpconfigport.h
+++ b/ports/esp32/mpconfigport.h
@@ -191,6 +191,7 @@
 #define MICROPY_PY_NETWORK_MODULE_GLOBALS_INCLUDEFILE "ports/esp32/modnetwork_globals.h"
 #ifndef MICROPY_PY_NETWORK_WLAN
 #define MICROPY_PY_NETWORK_WLAN             (1)
+#define MICROPY_PY_NETWORK_COUNTRY          mp_network_country
 #endif
 #ifndef MICROPY_HW_ENABLE_SDCARD
 #define MICROPY_HW_ENABLE_SDCARD            (1)
--- a/ports/esp32/network_common.c
+++ b/ports/esp32/network_common.c
@@ -79,6 +79,8 @@ MP_NORETURN void esp_exceptions_helper(esp_err_t e) {
             mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Wifi Not Connected"));
         case ESP_ERR_NO_MEM:
             mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("WiFi Out of Memory"));
+        case ESP_ERR_INVALID_ARG:
+            mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("WiFi Invalid Argument"));
         default:
             mp_raise_msg_varg(&mp_type_RuntimeError, MP_ERROR_TEXT("Wifi Unknown Error 0x%04x"), e);
     }
--- a/tests/ports/esp32/network_wlan_country.py.exp
+++ b/tests/ports/esp32/network_wlan_country.py.exp
@@ -1,0 +1,11 @@
+XX
+WiFi Invalid Argument
+XX
+AU
+AU
+WiFi Invalid Argument
+AU
+AU
+XX
+WiFi Invalid Argument
+XX


dpgeorge:
> * setting the country code with no WLAN interface ever being instantiated will raise an exception

I think it's important to be able to do this, set the country at the very start of the program.  That's then compatible with other ports, and means the user doesn't have to track when the WLAN comes up, so they can set the country afterwards.

Is it possible to follow how cyw43 works, use the existing implementation of `network_country()`, and set the country code in the IDF when `WLAN.active(1)` is called?

dpgeorge:
@agatti any comments on my above comment?  If you don't have time for this I think we should close it, and it can be revisited in the future if needed.

agatti:
I might have missed the comment, sorry :(

I recall that back then I did some tests and it didn't work with the ESP-IDF version I was using at the time, I believe it was v5.0.3.  I'll need to try this again on a more modern ESP-IDF SDK.

Thanks for the reminder, I'll update the PR this weekend since it needs rebasing anyway.

agatti:
It's still the case: "`ESP_ERR_WIFI_NOT_INIT: WiFi is not initialized by esp_wifi_init`".

What I can do, though, is to keep it in a buffer in memory and set it at the end of `esp_initialise_wifi` if the buffer is not empty.  I assume the extra 4 bytes of RAM taken by the country code shouldn't be a problem, but I'll also update the PR to make the country code bits be disabled via a define.

dpgeorge:
Can you reuse the `extmod/modnetwork.c:mod_network_country_code` global variable?

agatti:
My patch had its own buffer because `esp_wifi_set_country` and `esp_wifi_get_country` operate on three bytes long strings [1], so I tried to optimise for firmware size and avoiding the extra byte copies.

I guess I'll need to evaluate the footprint increase for the extra copy then.

[1] From the function documentation of `esp_wifi_set_country`: `The third octet of country code string is one of the following: ' ', 'O', 'I', 'X', otherwise it is considered as ' '.`
A similar remark appears in `esp_wifi_get_country` documentation.

Incidentally, I have no idea what that third character actually means, I've looked around a bit and found no reference explaining that - the IEEE802.11 specifications seem to be paywalled so I'll never know.

agatti:
I've reworked the PR, this should be shorter than the original version, using `mod_network_country_code` as storage.

This feature takes ~1.3KiB when built for `ESP32_GENERIC` on ESP-IDF 5.5.1, the same remarks as the bandwidth parameter addition for the D2WD variant apply here as well.

Now the country can be set at any time, with just one (documented) caveat: since there's no country code validation, if an invalid value is set before the WiFi radio is initialised all calls to `network.WLAN` will fail with no recovery until the country code is either reset back to 'XX' or to something more appropriate.

Setting the wrong country code in the situation mentioned above or after the WiFi radio is initialised used to fail with an unknown error 0x0102, I've added a human-readable entry in `ports/esp32/network_common.c` for said error code.

You'll notice that the conditional guards around the country setter/getter code use `#ifdef` rather than `#if`.  Since `MICROPY_PY_NETWORK_COUNTRY` needs to be defined in the first place for `extmod/modnetwork.c` to call the port's entry point, doing that avoids having to add another entry in `ports/esp32/mpconfigport.h`.  I can rework this part if using `#if` is preferred - removing the `MICROPY_PY_NETWORK_COUNTRY` definition will make setting the country code a no-op.

This should also work in hosted mode (I've succesfully built the firmware for `ESP32_GENERIC_P4`), but I cannot test such a configuration yet.

Finally, the test was rewritten as WLAN support now stores configuration information in RAM.  It still requires to be executed before the `network.WLAN` constructor gets invoked.

ports/esp32/network_wlan.c
```diff
@@ -772,6 +782,63 @@ static mp_obj_t network_wlan_config(size_t n_args, const mp_obj_t *args, mp_map_
...
+        // Maintain compatibility with the default implementation.
+        if (!(mod_network_country_code[0] == 'X' && mod_network_country_code[1] == 'X')) {
+            country_code[0] = mod_network_country_code[0];
+            country_code[1] = mod_network_country_code[1];
+        }
+        state = esp_wifi_set_country_code(country_code, true);
+        if (state != ESP_ERR_WIFI_NOT_INIT) {
+            esp_exceptions(state);
+        }
+    } else {
```
dpgeorge: This function can only be passed 0 or 1 arguments, due to:
```
MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mod_network_country_obj, 0, 1, network_country);
```
so this "else" is not reachable and can be removed.

agatti: Oh, I missed that!  The else block was removed, thanks for noticing that.

dpgeorge:
> * ESP-IDF unconditionally stores the chosen country code into the device's non-volatile storage

Hmm, that could be quite an unexpected and bad side effect of this change.  I see that `esp_wifi_set_country_code()` is called unconditionally in the constructor of `network.WLAN()`, so does that mean it unconditionally writes to flash every time WLAN is created??  If so, that could potentially wear out the flash page used to store this information.

To work around that, maybe can call `esp_wifi_get_country_code()` first, and only if the country code is not the expected one call `esp_wifi_set_country_code()` to set it?  Hopefully the same fix can be used for `network.country()` as well, so that if you have `network.country("IT")` in your `boot.py` then it will only update NVS once (subsequent calls will be a no-op).

agatti:
Oh, I believe that's not the case since ESP-IDF 5.1.x something - I updated the documentation in the PR but forgot to update the description (and also one of the commit messages).  It's been a while since I looked at that :|

After applying this simple change:

```diff
diff --git i/ports/esp32/network_wlan.c w/ports/esp32/network_wlan.c
index 4cbd21d5e..ba390e1dc 100644
--- i/ports/esp32/network_wlan.c
+++ w/ports/esp32/network_wlan.c
@@ -269,6 +269,10 @@ static mp_obj_t network_wlan_make_new(const mp_obj_type_t *type, size_t n_args,
         country_code[0] = mod_network_country_code[0];
         country_code[1] = mod_network_country_code[1];
     }
+
+    esp_exceptions(esp_wifi_get_country_code(country_code));
+    printf("COUNTRY CODE READ FROM IF: %s\n", country_code);
+
     // Always follow the AP's country if one is set.
     esp_exceptions(esp_wifi_set_country_code(country_code, true));
     #endif
```

This happens on my no-name ESP32_GENERIC board:

```
$ picocom -b 115200 /dev/ttyUSB0
picocom v3.1
[...]
Type [C-a] [C-h] to see available commands
Terminal ready
MicroPython v1.29.0-preview.339.gd72d8c4f3f.dirty on 2026-06-17; Generic ESP32 module with ESP32
>>> import network
>>> network.WLAN(network.AP_IF)
COUNTRY CODE READ FROM IF: 01 
<WLAN>
>>> network.country('IT')  # This will call `esp_wifi_set_country_code`
>>> network.country()
'IT'
>>> # Board unplugged from USB!
FATAL: read zero bytes from port
term_exitfunc: reset failed for dev UNKNOWN: Input/output error

# Wait a few seconds, then plug the board back in.

$ picocom -b 115200 /dev/ttyUSB0
picocom v3.1
[...]
Type [C-a] [C-h] to see available commands
Terminal ready
MicroPython v1.29.0-preview.339.gd72d8c4f3f.dirty on 2026-06-17; Generic ESP32 module with ESP32
>>> import network
>>> network.WLAN(network.AP_IF)
COUNTRY CODE READ FROM IF: 01 
<WLAN>
>>> 
Terminating...
Thanks for using picocom
```

The second run should have read 'IT' from the IF if it was saved into NVS.

I'll update the commit message right now, with no code changes.

dpgeorge:
> Oh, I believe that's not the case since ESP-IDF 5.1.x something

Where did you get the information that it's not the case anymore?

> The second run should have read 'IT' from the IF if it was saved into NVS.

Hmm, that's interesting.  Looking at IDF 5.5.2 source code file `components/esp_wifi/include/esp_wifi.h` it says:
```
  * @attention 4. The country configuration is stored into flash.
  *
```
https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_wifi.h#L1543

agatti:
> > Oh, I believe that's not the case since ESP-IDF 5.1.x something
> Where did you get the information that it's not the case anymore?

I might be confusing this with another WiFi feature that changed behaviour between 5.0 and 5.2, but when I initially wrote this PR the country was always written to NVS and now it is not, so it changed its behaviour in either 5.1 or 5.2.  (Explained below).

> > The second run should have read 'IT' from the IF if it was saved into NVS. 
> Hmm, that's interesting. Looking at IDF 5.5.2 source code file `components/esp_wifi/include/esp_wifi.h` it says:
> 
> ```
>   * @attention 4. The country configuration is stored into flash.
>   *
> ```
> 
> https://github.com/espressif/esp-idf/blob/master/components/esp_wifi/include/esp_wifi.h#L1543

There's a bug in my code too!  But the documentation is both correct and incorrect at the same time.

Whenever the interface is turned on the previously set country code was always written to the IF but no checking was done whether a code was *actually* set first.  So this makes it work "as intended" (there's more, though):

```diff
diff --git i/ports/esp32/network_wlan.c w/ports/esp32/network_wlan.c
index 4cbd21d5e..509c4b444 100644
--- i/ports/esp32/network_wlan.c
+++ w/ports/esp32/network_wlan.c
@@ -82,6 +82,7 @@ static bool mdns_initialised = false;
 
 static uint8_t conf_wifi_sta_reconnects = 0;
 static uint8_t wifi_sta_reconnects;
+static uint8_t network_country_code_set = 0;
 
 // The rules for this default are defined in the documentation of esp_wifi_set_protocol()
 // rather than in code, so we have to recreate them here.
@@ -269,8 +270,11 @@ static mp_obj_t network_wlan_make_new(const mp_obj_type_t *type, size_t n_args,
         country_code[0] = mod_network_country_code[0];
         country_code[1] = mod_network_country_code[1];
     }
-    // Always follow the AP's country if one is set.
-    esp_exceptions(esp_wifi_set_country_code(country_code, true));
+    if (network_country_code_set) {
+        // Always follow the AP's country if one is set.
+        esp_exceptions(esp_wifi_set_country_code(country_code, true));
+        network_country_code_set = 0;
+    }
     #endif
 
     int idx = (n_args > 0) ? mp_obj_get_int(args[0]) : ESP_IF_WIFI_STA;
@@ -825,6 +829,8 @@ mp_obj_t mp_network_country(size_t n_args, const mp_obj_t *pos_args) {
         state = esp_wifi_set_country_code(country_code, true);
         if (state != ESP_ERR_WIFI_NOT_INIT) {
             esp_exceptions(state);
+        } else {
+            network_country_code_set = 1;
         }
     }
```

**but** if what I see in my tests is correct, `esp_wifi_set_country_code` will set the country code into the Wifi NVS (which seems to be inaccessible, according to ieee80211_nvs.o the NVS name is "nvs.net80211" and the key is simply "country") but this time it seems that `esp_wifi_set_storage(WIFI_STORAGE_RAM)` is actually taken into account unlike back when I wrote the PR, so the internal call to `g_osi_funcs_p->_nvs_commit()` is skipped.

You can try this: apply the diff mentioned above, uncomment the call to `esp_wifi_set_storage()`, then execute [1], press CTRL+D, wait a couple of seconds and unplug the board, wait some more time, plug the board back in and execute [2].  Then, bring the call to `esp_wifi_set_storage()` back in, and re-do the same procedure but replacing 'IT' with 'AU'.  [2] will still read 'IT' even if you explicitly set 'AU' in the modified version of [1].

[1]

```py
import esp
esp.osdebug(0)
import network
w = network.WLAN(network.AP_IF)
w.active(True)
network.country('IT')
w.active(False)
```

[2]

```py
import esp
esp.osdebug(0)
import network
w = network.WLAN(network.AP_IF)
print(network.country())
```

dpgeorge:
Thanks for the very detailed response, and code to show the behaviour.

I tested it out as you describe above (with IDF 5.5.2, on ESP32_GENERIC) and am convinced that `esp_wifi_set_storage(WIFI_STORAGE_RAM)` works, and the country code *is not* stored to flash.

That's great!

extmod/modnetwork.h
```diff
@@ -149,4 +149,6 @@ void mod_network_register_nic(mp_obj_t nic);
 mp_obj_t mod_network_find_nic(const uint8_t *ip);
 #endif
 
+mp_obj_t mp_network_country(size_t n_args, const mp_obj_t *args);
```
dpgeorge, line 152: This doesn't seem right, the port could use a different function name.

Suggest either:
```c
#if defined(MICROPY_PY_NETWORK_COUNTRY)
mp_obj_t MICROPY_PY_NETWORK_COUNTRY(size_t n_args, const mp_obj_t *args);
#endif
```
Or, any ports using this feature must declare the function in their `mphalport.h` (which is included indirectly into `extmod/modnetwork.c`).

ports/esp32/network_wlan.c
```diff
@@ -772,6 +786,58 @@ static mp_obj_t network_wlan_config(size_t n_args, const mp_obj_t *args, mp_map_
...
+        }
+        mod_network_country_code[0] = code[0];
+        mod_network_country_code[1] = code[1];
+        // Maintain compatibility with the default implementation.
+        if (!(mod_network_country_code[0] == 'X' && mod_network_country_code[1] == 'X')) {
+            country_code[0] = mod_network_country_code[0];
+            country_code[1] = mod_network_country_code[1];
+        }
+        state = esp_wifi_set_country_code(country_code, true);
+        if (state != ESP_ERR_WIFI_NOT_INIT) {
```
dpgeorge, line 830: This handling of `ESP_ERR_WIFI_NOT_INIT` and the global variable `network_country_code_set` seems a little awkward.  It also means that if the country code is (for some reason) stored in NVRAM then `network.country()` returns `"XX"` until you create a WLAN instance.

To improve this, I suggest to simply call `esp_initialise_wifi()` at the start of this `mp_network_country()` function.  That should make it behave pretty well, the special note in the docs is no longer needed, the variable `network_country_code_set` is not needed, and the code added to `network_wlan_make_new()` is also probably not needed anymore.

If users are setting/getting the country code, then they probably want wifi initialised anyway (or it'll be initialised soon enough when they create a WLAN instance).

tests/ports/esp32/network_wlan_country.py
```diff
@@ -0,0 +1,40 @@
...
+network.country("XX")
+print(network.country())
+
+network.country("QQ")
+try:
+    network.WLAN(network.STA_IF)
+    print("Failed to raise")
+except OSError as e:
+    print(str(e))
+print(network.country())
```
dpgeorge, line 15: This line actually failed for me, it printed `IT` 😂 

That's because I had `IT` set in the NVRAM from running your test, and due to the way the country setting is delayed until WLAN creation, it loaded up the NVRAM value.

With my suggestion above to run `esp_initialise_wifi()`, that should fix the issue.

ports/esp32/network_wlan.c
agatti: This makes sense, but I wonder whether calling `esp_initialise_wifi` behind the scenes is desirable.

That function allocates a fairly sizeable chunk of heap: when dealing with heavy libraries like microdot, on non-SPIRAM units you usually end up with ~22k total left to MicroPython or something like that last time I checked.  The fix there is to delay creating the WLAN object until everything is initialised and a GC pass is performed.

If code is also meant to work on boards with cyw43 or other network adapters, then at the very least documentation has to be modified to mention this and suggest calling `network.country` after a WLAN is created (which is fine for STA, but for AP you may briefly break regulations if you start with the wrong settings).

dpgeorge: The original intention for `network.country()` was to set/get the region for *all* networking adaptors, which includes WiFi and potentially others, and anything else network-related that has a country setting.

Ideally you set the country very early on in your code, so that it can be picked up by the necessary network components.

Unfortunately the esp32 doesn't work well with that model :disappointed: 

I know you tried hard to make it work, and maybe that's the best we can do.  But right now the test here fails on my board unless I erase NVRAM... and maybe that's all I need to do, the test failing is a very minor point.

agatti: There are two scenarios here:

- The device acts as a STA node: the second argument to `esp_wifi_set_country_code` actually indicates whether the country code coming in from an AP takes precedence or not, and right now it is configured to do so in all cases.  Considering that it is relatively uncommon for an AP to not broadcast its country code, unless you're also using other (mis)configured MCU boards to be APs acting that way, calling `network.country` ends up being a no-op anyway.

- The device acts as an AP node: if you're in AP mode and do care about these sort of things you probably know better than leaving this setting alone in your hardware if you're broadcasting, which is more or less your argument about who is going to call `network.country` in the first place.  However, AP mode is when things like microdot are going to be used.  This is where documentation changes could address the situation, as somebody's bound to get bitten by this when they update their MicroPython version - but at least there will be something to point them to so they can make the appropriate changes in their code.

After all, the call was a no-op before this, so users need to consider the trade-off, which could be placed in the documentation section.  Ultimately they're free to undefine the appropriate entry in `mpconfigport.h` and they won't even notice this was even here.

So, I can apply the suggested changes to keep things consistent, but I'll have to expand the documentation section - which I don't mind doing.  If in the future (v2?) `network.country` changes its characteristics, then it'll be easy to revisit this too.

dpgeorge: Thinking about this more, I feel you are right that `network.country()` should *not* initialize the wifi.

As I mentioned, `network.country()` is intended to be used for anything needing a country setting.  Let's consider that maybe LoRa, Zigbee and HaLow all need country settings as well (that may or may not be strictly true, but let's assume for now).  And assume I have pure-Python drivers for all these wireless protocols (we already do for LoRa).  Then that driver may want to do `c = network.country()` to get the current country (system-wide) settings that the user configured, and apply that to the driver.  Eg in LoRa:
```py
import network

COUNTRY_SETTINGS = {
    "AU": 916_000,
    "IT": 868_000,
    ...
}

lora_cfg = {...}
lora_cfg["freq_khz"] = COUNTRY_SETTINGS.get(network.country(), DEFAULT_FREQ)
```
I know that's a little contrived (the LoRa bands are actually named like AU916, EU868 etc and you probably want to select them with those names) but in principal this is what `network.country()` could be used for.

In that case, `network.country(value)` should set a source of truth value for the country setting, and `network.country()` should retrieve that source of truth.  Then, for this PR, when starting up wifi it should try to set the wifi country from that source of truth.  But I don't think the source of truth should be updated from the wifi country value, because that will interfere with potential use in, eg, LoRa drivers.

Following that logic through to code in this PR, the only change would be that getting the country should not call `esp_wifi_get_country_code()`.  Rather, it should just return the value stored in `mod_network_country_code`.  (And then MICROPY_PY_NETWORK_COUNTRY is probably not needed, instead we just need a hook at the end of the existing `network_country` function like `mp_network_country_changed()`).

(I understand it may be useful to know what country the STA is using because it was set by the AP's country.  In that case maybe we need to expose that differently, eg `wlan.status('country')`).

PR #15917, opened 2024-09-26
extmod/asyncio/stream.py: Improve loading of ssl  and socket for Async IO.

srinsoz:
### Summary
Reduce overhead when creating new sockets. The current version imports socket and ssl every time a socket is created, which has a significant overhead, in particular on devices with a filesystem in flash memory (3 filestat access for each import). In this new version, the code first checks if the module is already imported. If not, the exception is caught, and the module is imported.

This improvement speeds up the execution of code that frequently opens sockets (outgoing or incoming).

### Testing

Use Async IO function open_connection.
````
 reader, writer = await asyncio.open_connection('192.168.1.1', 8888)
````

### CODE REVIEW

**Comment by peterhinch** [comment]
Python already performs this check: importing a module which is already loaded has no effect.

**Comment by sosi-deadeye** [comment]
CPython reads during an import only from disk, if the module were not found in `sys.modules`.
I would be surprised if Micropython deviates from this.

```python
>>> import sys
>>> sys.modules
{}
>>> import os
>>> sys.modules
{'os.path': <module 'os.path' from '/home/deadeye/.micropython/lib/os/path.mpy'>, 'os': <module 'os' from '/home/deadeye/.micropython/lib/os/__init__.mpy'>}
```

srinsoz:
I've just checked with the Windows port of MicroPython, and I do not experience the same behavior.

See output:
```python
C:\tmp\micropython\ports\windows>build-standard\micropython.exe
MicroPython v1.24.0-preview.363.g17d823445.dirty on 2024-09-30; win32 [MSC v.1941 32 bit] version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> import sys
>>> sys.modules
{}
>>> import os
>>> sys.modules
{}
>>>
```

I believe that there is something wrong with the handling of [``mp_loaded_modules_dict``](https://github.com/micropython/micropython/blob/17d82344581ad3a76033fae54c5d3304e17f185f/py/builtinimport.c#L380). At the start of the function ``process_import_at_level``, the code checks if the module is loaded. If not, it load it, but then the ``mp_loaded_modules_dict`` dictionary is never updated.



```c
static mp_obj_t process_import_at_level(qstr full_mod_name, qstr level_mod_name, mp_obj_t outer_module_obj, bool override_main) {
    // Immediately return if the module at this level is already loaded.
    mp_map_elem_t *elem;

    #if MICROPY_PY_SYS
    // If sys.path is empty, the intention is to force using a built-in. This
    // means we should also ignore any loaded modules with the same name
    // which may have come from the filesystem.
    size_t path_num;
    mp_obj_t *path_items;
    mp_obj_get_array(mp_sys_path, &path_num, &path_items);
    if (path_num)
    #endif
    {
        elem = mp_map_lookup(&MP_STATE_VM(mp_loaded_modules_dict).map, MP_OBJ_NEW_QSTR(full_mod_name), MP_MAP_LOOKUP);
        if (elem) {
            return elem->value;
        }
    }
```

peterhinch:
On Linux:
```py
MicroPython v1.24.0-preview.106.g55e75c4ad on 2024-07-15; linux [GCC 11.4.0] version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> import sys
>>> sys.modules
{}
>>> import asyncio
>>> sys.modules
{'asyncio.core': <module 'asyncio.core' from 'asyncio/core.py'>, 'asyncio': <module 'asyncio' from 'asyncio/__init__.py'>}
```
I suggest you try this test: `os` may be a special case.

srinsoz:
You are right. I've dug a bit deeper and I have a better understanding of the "issue".

Our main problem is that, when importing ``socket``, three calls to [``stat_module``](https://github.com/micropython/micropython/blob/17d82344581ad3a76033fae54c5d3304e17f185f/py/builtinimport.c#L102) occurs. On normal operation this is not a big deal since it happens only one time when the file is loaded. But in the case of [``asyncio.open_connection``](https://github.com/micropython/micropython/blob/17d82344581ad3a76033fae54c5d3304e17f185f/extmod/asyncio/stream.py#L101) this leads to three calls to ``stat_module`` at every connection!

This problem comes from multiple factors. 

The first factor is that ``socket`` module is imported inside the ``open_connection``. This is great to reduce the memory footprint and do not load unnecessary modules, but this leads to multiple calls to ``import socket`` during the execution of the program.

The second factor is that the ``socket`` module is declared as an extensible module, but has no extension. 

As far I understand we have three scenario for imports.

1. The module is not builtin and is loaded from the fs.
2. The module is builtin and does not support extension (declared with ``MICROPY_REGISTERED_MODULES``). 
3. The module is builtin and does  support extension (declared with ``MP_REGISTER_EXTENSIBLE_MODULE``). 

Case 1 is not an issue. The first import executes the calls to ``stat_module`` then load the file and store the modules in the ``mp_loaded_modules_dict`` variable. These modules are then listed in the ``sys.modules``. Following calls to import skip the file stat and the loading because they are already present in ``mp_loaded_modules_dict``.

Case 2 is also not an issue, because import find the builtin modules without calling ``stat_module``. These modules are not added to ``mp_loaded_modules_dict`` (that the case of ``import os``).  Following calls do the same, but since no IO are required in this case, it's not a performance issue.

The problematic case is the number 3. To allow user to extend the modules from the filesystem, the code test if a module is present on the FS using ``stat_module``. And we have two sub case

3a. A file is present. The modules is loaded and added to ``mp_loaded_modules_dict`` (like in case 1). Following import will not trigger any ``stat_module`` since the module is listed in ``mp_loaded_modules_dict``. 

3b. The problematic situation is when no modules is present on the files system. In this situation, the code calls 3 time ``stat_module`` to test if a external module is present. Then since no file is found, it loads the build version of the module, BUT does not add it to ``mp_loaded_modules_dict``! So following imports will redo the  full process. It will call 3 times ``stat_module`` and load again the modules :-(

On normal operation, this penalty is not a big deal, since imports are not used frequently, but on this case it's called every new connection and this has a significant impact on performance, specially if your filesystem use an external flash.

In my point of view, the case 3b should also update ``mp_loaded_modules_dict``. Any suggestion?

dpgeorge:
I think we should try to fix case 3b.  That will improve import time for a lot of common modules, eg `io`, `struct`, `os`, `socket`, `json`.

PR #15919, opened 2024-09-26
uos: Add BUILD_ID in the firmware

IlorDash:
It'll be convenient to store build ID in device firmware to help track builds more effectively with building Micropython in CI/CD environments such as Zuul or Jenkins.

This commit adds support for including the BUILD_ID in the firmware. BUILD_ID is retrieved from similar environment variable during the build process. If BUILD_ID is not present, nothing is included. MICROPY_BUILD_ID_STR is defined in makeversionhdr.py and used everywhere with MICROPY_BUILD_DATE.

You can access BUILD_ID in REPL, e.g. for BUILD_ID = 1105413164430710189699:
```
>>> os.uname()
(sysname='pyboard', nodename='pyboard', release='1.23.0', version='v1.23.0-102.g288b6be46.dirty on 2024-09-26, build: 1105413164430710189699', machine='MY_BOARD')
>>> sys.version
'3.4.0; MicroPython v1.23.0-102.g288b6be46.dirty on 2024-09-26, build: 1105413164430710189699'
```


stinos:
Unlike git tag and build date which are always available, and git hash which is often going to be available, the build id is probably only going to be available when building on specific CI. So I'm wondering if perhaps instead of showing `build: no build-id>` everywhere, which isn't super useful, this should be made such that if there's no build id it is also not shown?

PR #15924, opened 2024-09-27
py/nlrrv64: Add RISC-V RV64I NLR implementation.

agatti:
### Summary

Following the RV32 port, I updated the NLR code for RV64.  It is nothing more than the same code but with doubled byte offsets inside the structure and double-words load/store opcodes.

### Testing

The Qemu RV64 Unix tests pass - I'm also in the process to set up a MicroPython-enabled buildroot for a Milk-V Duo so that RV64 can also be tested locally on my end in the near future (one of my RV64 boards is now free for MicroPython development).

### DIFF
--- a/py/nlr.h
+++ b/py/nlr.h
@@ -45,6 +45,7 @@
 #define MICROPY_NLR_NUM_REGS_XTENSA         (10)
 #define MICROPY_NLR_NUM_REGS_XTENSAWIN      (17)
 #define MICROPY_NLR_NUM_REGS_RV32I          (14)
+#define MICROPY_NLR_NUM_REGS_RV64I          (14)
 
 // *FORMAT-OFF*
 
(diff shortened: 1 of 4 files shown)


py/nlr.h
```diff
@@ -89,9 +90,16 @@
 #elif defined(__mips__)
     #define MICROPY_NLR_MIPS (1)
     #define MICROPY_NLR_NUM_REGS (MICROPY_NLR_NUM_REGS_MIPS)
-#elif defined(__riscv) && defined(__riscv_xlen) && (__riscv_xlen == 32)
-    #define MICROPY_NLR_RV32I (1)
-    #define MICROPY_NLR_NUM_REGS (MICROPY_NLR_NUM_REGS_RV32I)
+#elif defined(__riscv)
+    #if (__riscv_xlen == 32)
```
dpgeorge, line 94: We usually don't put parenthesis around logic expressions that are on their own, in preprocessor if's.

agatti: Sorry, this should be fixed now.

dpgeorge (approved):

py/nlr.h
dpgeorge: Thanks!

PR #15929, opened 2024-09-29
ports/renesas: Replace MICROPY_EVENT_POLL_HOOK with mp_event_wait.

andrewleech:
### Summary

Basic update to the renesas to replace the traditional `MICROPY_EVENT_POLL_HOOK` with the newer `mp_event_wait_indefinite()` / `mp_event_wait_ms()` as appropriate.

### Testing
Renesas imxrt1010 EVK

`python3 test_serial.py` from #15909 has been run on both usb and uart and works without any failures.

`~/micropython/tests$ ./run-tests.py --target renesas-ra --device /dev/ttyACM0` has been run before and after MR with no changes (fails on `builtin_pow3_intbig machine_spi_rate` in both cases).

### DIFF
--- a/ports/renesas-ra/mphalport.h
+++ b/ports/renesas-ra/mphalport.h
@@ -35,6 +35,14 @@
 #define MICROPY_PY_PENDSV_ENTER   uint32_t atomic_state = raise_irq_pri(IRQ_PRI_PENDSV)
 #define MICROPY_PY_PENDSV_EXIT    restore_irq_pri(atomic_state)
 
+// Port level Wait-for-Event macro
+//
+// Do not use this macro directly, include py/runtime.h and
+// call mp_event_wait_indefinite() or mp_event_wait_ms(timeout).
+// Uses WFI which will wake up from regular systick interrupt if not
+// before from any other source.
+#define MICROPY_INTERNAL_WFE(TIMEOUT_MS) __WFI()
+
 #define MICROPY_PY_LWIP_ENTER
 #define MICROPY_PY_LWIP_REENTER
 #define MICROPY_PY_LWIP_EXIT
--- a/drivers/esp-hosted/esp_hosted_bthci_uart.c
+++ b/drivers/esp-hosted/esp_hosted_bthci_uart.c
@@ -72,7 +72,7 @@ int esp_hosted_hci_cmd(int ogf, int ocf, size_t param_len, const uint8_t *param_
     // Receive HCI event packet, initially reading 3 bytes (HCI Event, Event code, Plen).
     for (mp_uint_t start = mp_hal_ticks_ms(), size = 3, i = 0; i < size;) {
         while (!mp_bluetooth_hci_uart_any()) {
-            MICROPY_EVENT_POLL_HOOK
+            mp_event_wait_ms(1);
             // Timeout.
             if ((mp_hal_ticks_ms() - start) > HCI_COMMAND_TIMEOUT) {
                 error_printf("timeout waiting for HCI packet\n");
@@ -126,7 +126,7 @@ int mp_bluetooth_hci_controller_init(void) {
         if (mp_bluetooth_hci_uart_any()) {
             mp_bluetooth_hci_uart_readchar();
         }
-        MICROPY_EVENT_POLL_HOOK
+        mp_event_wait_ms(1);
     }
 
     #ifdef MICROPY_HW_BLE_UART_BAUDRATE_SECONDARY
--- a/drivers/esp-hosted/esp_hosted_wifi.c
+++ b/drivers/esp-hosted/esp_hosted_wifi.c
@@ -243,7 +243,7 @@ static int esp_hosted_request(CtrlMsgId msg_id, void *ctrl_payload) {
 
 static CtrlMsg *esp_hosted_response(CtrlMsgId msg_id, uint32_t timeout) {
     CtrlMsg *ctrl_msg = NULL;
-    for (mp_uint_t start = mp_hal_ticks_ms(); ; mp_hal_delay_ms(10)) {
+    for (mp_uint_t start = mp_hal_ticks_ms(); ; mp_event_wait_ms(10)) {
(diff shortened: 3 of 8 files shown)


dpgeorge:
Thanks.  We need to eventually do this for all ports.

andrewleech:
@iabdalkader I had to also make a similar change in `drivers/esp_hosted_wifi: Replace EVENT_POLL_HOOK with mp_event_wait_ms.`, included in this PR. I don't have any hardware to test this on though, does the change look ok to you?

I will endevour to get something set up to test esphosted as I'm keen to use it in the near future, but not sure when I'll have it running.

andrewleech:
@dpgeorge I've now done some basic testing and switched to `mp_event_wait_ms()` in a few places similar to a quick review of other ports.

drivers/esp-hosted/esp_hosted_bthci.c
```diff
@@ -126,7 +126,7 @@ int mp_bluetooth_hci_controller_init(void) {
         if (mp_bluetooth_hci_uart_any()) {
```
andrewleech, line 126: Bit off topic, but @iabdalkader does this just block for 2.5 seconds at boot every time? Does the esp chip take that long to boot, or is it just extra timeout in case?

iabdalkader: Yes, because the firmware is configured to output debug messages on boot.

iabdalkader: I think you meant to comment on this line: https://github.com/micropython/micropython/blob/acecb0f81909dc02f4ce931ac29a2fe6c58a6818/drivers/esp-hosted/esp_hosted_bthci.c#L125

andrewleech: Thanks, I did, but github only seems to let me start a comment on a line x rows from an actual change... this was as close as I could get :-D

iabdalkader:
> @iabdalkader I had to also make a similar change in `drivers/esp_hosted_wifi: Replace EVENT_POLL_HOOK with mp_event_wait_ms.`, included in this PR. I don't have any hardware to test this on though, does the change look ok to you?

I can test it next week and let you know.

iabdalkader:
I tested this, sorry for the delay. I ran multitests and they pass just as before. Unrelated, but most of the sslcontext tests are skipped not sure why (this was the case before this change).

andrewleech:
Thanks @iabdalkader for the testing! No apologies necessary, this was just a bit of opportunistic cleanup for me and not in a rush for any reason.

andrewleech:
> Although this PR has correct behaviour I think it will increase power consumption at an idle REPL (and other idle locations) because it's now doing a busy loop rather than executing WFI.
> 
> To fix that you'll need to define MICROPY_INTERNAL_WFE. Since renesas-ra is not tickless (unlike rp2 which is tickless) this macro probably needs to be:
> 
> ```c
> #define MICROPY_INTERNAL_WFE(TIMEOUT_MS) __WFI()
> ```
> 
> And please put that in `mphalport.h`.

Thanks for the review @dpgeorge 
I added a more slightly more detailed variant of this define which uses __WFI() if no timeout is given, else __WFI() for a timeout.

ports/renesas-ra/mphalport.h
```diff
@@ -35,6 +35,22 @@
...
+// Port level Wait-for-Event macro
+//
+// Do not use this macro directly, include py/runtime.h and
+// call mp_event_wait_indefinite() or mp_event_wait_ms(timeout)
+// Uses WFI when timeout is given so it wakes up from systick
+// otherwise it may not wake up within given timeout.
+#define MICROPY_INTERNAL_WFE(TIMEOUT_MS)                        \
+    do {                                                        \
+        if ((TIMEOUT_MS) < 0) {                                 \
+            __WFE();                                            \
```
dpgeorge, line 48: I'm not sure WFE is correct to use on this port.  It doesn't do SEV anywhere, and SCB_SCR_SEVONPEND is not enabled (as far as I can tell), so it's unlikely that the WFE will ever wake up.

Eg this WFE will be run each time `mp_hal_stdin_rx_chr()` loops around, and so it's supposed to wake whenever a new character is available, ie an "event" occurs.  But I can't see how that would work in the current code (would need to add SEV's in the right locations).

andrewleech: Oh interesting, I don't know much about these internals.
I had done some minimal testing at repl (over uart) which I just repeated:
```
anl@STEP:~/micropython/ports/renesas-ra$ mpremote u0
Connected to MicroPython at /dev/ttyUSB0

>>>
MPY: sync filesystems
MPY: soft reboot
MicroPython v1.26.0-preview.25.g0f1c2c1f25.dirty on 2025-04-24; EK-RA4W1 with RA4W1
>>>
>>> import time
>>> time.sleep(1)
>>> while True:
...   time.sleep(1)
...
KeyboardInterrupt:
>>>
>>> while True:
...   pass
...
KeyboardInterrupt:
>>>
```

dpgeorge: It may be that interrupts automatically set the event flag on these MCUs.  In that case, WFI and WFE are equivalent, in which case this macro should just unconditionally use one of them.

We need to get this right because it's tricky and can have lots of impacts on the system if not done the right way.

andrewleech: I went hunting on their website for a reference manual that covers the core in this way, found nothing.

Looking at the source, these do just both go to cmsis that implements them as the two `wfi`/`wfe` assembly commands which doesn't really help. 

Based on the lack of data I'm certainly inclined to take the route of simple / safe and just use the one-liner WFI command instead.

dpgeorge: Yes, I think just a simple WFI would be best here.

andrewleech: I've pushed the simplified change now thanks.

dpgeorge (approved):

PR #15931, opened 2024-09-29
qemu/riscv64: Add RISC-V 64 bits port.

agatti:
### Summary

This is a port for bare-metal RV64, which is largely the same as the RV32 port but without native code generation.  Only a few tweaks here and there were needed to get this to the REPL stage, and a few more to let it pass the test suite.

### Testing

This PR also adds a CI target to run the test suite on bare-metal RV64.

It's pretty rare for RV64 boards to not have hard floating point support built-in (usually double-precision).  Using RV64IMC is still something, but if people are going to use the bare-metal version I suspect they want to use hard floating point as well.

Now, that alone isn't much of an issue if not for the fact that the NLR and the GC helper need to take care of floating point registers as well (native code generation comes later).  This may require a bit of thought and planning...

### DIFF
--- a/docs/reference/mpyfiles.rst
+++ b/docs/reference/mpyfiles.rst
@@ -58,7 +58,7 @@ If importing an .mpy file fails then try the following:
     sys_mpy = sys.implementation._mpy
     arch = [None, 'x86', 'x64',
         'armv6', 'armv6m', 'armv7m', 'armv7em', 'armv7emsp', 'armv7emdp',
-        'xtensa', 'xtensawin', 'rv32imc'][sys_mpy >> 10]
+        'xtensa', 'xtensawin', 'rv32imc', 'rv64imc'][sys_mpy >> 10]
     print('mpy version:', sys_mpy & 0xff)
     print('mpy sub-version:', sys_mpy >> 8 & 3)
     print('mpy flags:', end='')
--- a/ports/qemu/README.md
+++ b/ports/qemu/README.md
@@ -1,8 +1,8 @@
 MicroPython port to qemu
 ========================
 
-This is experimental, community-supported port for Cortex-M and RISC-V RV32IMC
-emulation as provided by QEMU (http://qemu.org).
+This is experimental, community-supported port for Cortex-M and RISC-V
+RV32IMC/RV64IMC emulation as provided by QEMU (http://qemu.org).
 
 The purposes of this port are to enable:
 
@@ -26,9 +26,9 @@ Dependencies
 For ARM-based boards the build requires a bare-metal ARM toolchain, such as
 `arm-none-eabi-gcc`.
 
-### RISC-V
+### RISC-V 32
 
-For RISC-V-based boards the build requires a bare metal RISC-V toolchain with GCC 10
+For RV32-based boards the build requires a bare metal RISC-V toolchain with GCC 10
 or later, either with multilib support or 32 bits specific (M, C, and Zicsr
 extensions must be supported, along with ilp32 ABI).  Both newlib and picolibc are
 supported, with the latter having precedence if found.
@@ -36,6 +36,47 @@ supported, with the latter having precedence if found.
 Most pre-built toolchains should work out of the box, either coming from your
 Linux distribution's package manager, or independently packaged ones like
 [xPack](https://xpack.github.io/dev-tools/riscv-none-elf-gcc/).
+
+### RISC-V 64
+
+For RV64-based boards the build requires a bare metal RISC-V toolchain with GCC 10
+or later, either with multilib support or 64 bits specific (M, C, and Zicsr
+extensions must be supported, along with lp64 ABI).  In other words, if your
+compiler can build a binary that's compliant to any ratified RVA profiles,
+you *might* be good to go: if you run your cross-compiler `gcc` binary with `-v` as
+its only argument and you don't see mentions of `-mcmodel=medany`, your
+toolchain is probably not going to work for this situation (more later).  Both newlib
+and picolibc are supported, with the latter having precedence if found.
+
+RISC-V 64 toolchains are a bit more temperamental.  Some toolchains have their `libc`,
+`libm`, and `libgcc` libraries built in a way that they cannot be used with code
+placed at an offset beyond 0x7FFFF7FF.  QEMU's `VIRT_RV64` board code address space
+starts at 0x80000000, making those toolchains not viable if code from those libraries
+is brought in by the build process.  A lot of readily available RISC-V toolchains are
+unfortunately broken from this point of view.
+
+However, to work around this, a Docker container file is provided to be able to run
+tests using a known working compiler (the same one running MicroPython's RV32 and RV64
+CI jobs).
+
+To build the provided toolchain image you can use the following commands:
+
+```bash
+cd $MICROPYTHON_SOURCE_ROOT/ports/qemu
+docker build -t micropython/mpy-qemu-riscv -f Dockerfile.riscv .
+```
+
+and then tests can be run with:
+
+```bash
+cd $MICROPYTHON_SOURCE_ROOT
+docker container run -v .:/micropython --rm -it micropython/mpy-qemu-riscv make -C ports/qemu -- BOARD=VIRT_RV64 test
+```
+
+(If you use Podman, you can replace `docker` with `podman` in the commands above).
+
+Also, if you replace `VIRT_RV64` with `VIRT_RV32` you should be able to run
+32-bits tests as well!
 
 Build instructions
 ------------------
@@ -63,6 +104,7 @@ Available boards are:
 | `NETDUINO2`       | `arm`        | `netduino2`              |
 | `SABRELITE`       | `arm`        | `sabrelite`              |
 | `VIRT_RV32`       | `riscv32`    | `virt`                   |
+| `VIRT_RV64`       | `riscv64`    | `virt`                   |
 
 Running
 -------
@@ -100,8 +142,8 @@ tests against the serial device, for example:
     $ ./run-tests.py -t /dev/pts/1
 
 Selected native modules that come as examples with the MicroPython source tree
-can also be tested with this command (this is currently supported only for the
-`VIRT_RV32` board):
+can also be tested with this command (this is currently not supported for the
+`VIRT_RV64` board):
 
     $ make test_natmod
 
--- a/.github/workflows/ports_qemu.yml
+++ b/.github/workflows/ports_qemu.yml
@@ -50,3 +50,15 @@ jobs:
     - name: Print failures
       if: failure()
       run: tests/run-tests.py --print-failures
+
+  build_and_test_rv64:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install packages
+      run: source tools/ci.sh && ci_qemu_setup_rv64
+    - name: Build and run test suite
+      run: source tools/ci.sh && ci_qemu_build_rv64
+    - name: Print failures
+      if: failure()
+      run: tests/run-tests.py --print-failures
--- a/mpy-cross/main.c
+++ b/mpy-cross/main.c
@@ -130,7 +130,8 @@ static int usage(char **argv) {
         "Target specific options:\n"
         "-msmall-int-bits=number : set the maximum bits used to encode a small-int\n"
         "-march=<arch> : set architecture for native emitter;\n"
-        "                x86, x64, armv6, armv6m, armv7m, armv7em, armv7emsp, armv7emdp, xtensa, xtensawin, rv32imc, host, debug\n"
+        "                x86, x64, armv6, armv6m, armv7m, armv7em, armv7emsp,\n"
+        "                armv7emdp, xtensa, xtensawin, rv32imc, rv64imc, host, debug\n"
         "\n"
         "Implementation specific options:\n", argv[0]
         );
@@ -316,6 +317,9 @@ MP_NOINLINE int main_(int argc, char **argv) {
                 } else if (strcmp(arch, "rv32imc") == 0) {
                     mp_dynamic_compiler.native_arch = MP_NATIVE_ARCH_RV32IMC;
                     mp_dynamic_compiler.nlr_buf_num_regs = MICROPY_NLR_NUM_REGS_RV32I;
+                } else if (strcmp(arch, "rv64imc") == 0) {
+                    mp_dynamic_compiler.native_arch = MP_NATIVE_ARCH_RV64IMC;
+                    mp_dynamic_compiler.nlr_buf_num_regs = MICROPY_NLR_NUM_REGS_RV64I;
                 } else if (strcmp(arch, "debug") == 0) {
                     mp_dynamic_compiler.native_arch = MP_NATIVE_ARCH_DEBUG;
                     mp_dynamic_compiler.nlr_buf_num_regs = 0;
@@ -329,6 +333,9 @@ MP_NOINLINE int main_(int argc, char **argv) {
                     #elif defined(__arm__) && !defined(__thumb2__)
                     mp_dynamic_compiler.native_arch = MP_NATIVE_ARCH_ARMV6;
                     mp_dynamic_compiler.nlr_buf_num_regs = MICROPY_NLR_NUM_REGS_ARM_THUMB_FP;
+                    #elif defined(__riscv) && (__riscv_xlen == 64)
+                    mp_dynamic_compiler.native_arch = MP_NATIVE_ARCH_RV64IMC;
+                    mp_dynamic_compiler.nlr_buf_num_regs = MICROPY_NLR_NUM_REGS_RV64I;
                     #else
                     mp_printf(&mp_stderr_print, "unable to determine host architecture for -march=host\n");
                     exit(1);
--- a/mpy-cross/mpy_cross/__init__.py
+++ b/mpy-cross/mpy_cross/__init__.py
@@ -43,6 +43,7 @@ NATIVE_ARCHS = {
     "NATIVE_ARCH_XTENSA": "xtensa",
     "NATIVE_ARCH_XTENSAWIN": "xtensawin",
     "NATIVE_ARCH_RV32IMC": "rv32imc",
+    "NATIVE_ARCH_RV64IMC": "rv64imc",
 }
 
 globals().update(NATIVE_ARCHS)
--- a/ports/qemu/Dockerfile.riscv
+++ b/ports/qemu/Dockerfile.riscv
@@ -1,0 +1,39 @@
+# This file is part of the MicroPython project, https://micropython.org/
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2025 Alessandro Gatti
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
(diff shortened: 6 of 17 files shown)


agatti:
After a few changes this seems to work now, marking it as ready for review.

@dpgeorge I've had to include the GC heap size changes here to fix the `micropython/memstats` tests as it would run out of memory when allocating a 10K elements list.  This should also fix the RV32 test failure you see in #15937.

agatti:
Regarding the fact that RV64 cores usually come with an FPU on board and whatnot...

Would a PR that bakes in the extensions set an MPY file is built for be accepted?

I thought about adding an extra raw code/sub-code element with the extensions set the code is compiled for, with the default being IMC (on both RV32 and RV64) if missing - to not break existing MPY files that have been built with the -pre version of MicroPython and to not delay the current release date.

That would allow rejecting code at runtime on RV32 platforms that do not support a particular extension, maintaining file format compatibility and avoiding crashes, in addition to limit the number of RISC-V targets in MPY files to just two: RV32 (aka RV32IMC) and RV64.  My main target right now is enhancing generated code for RP2, as it has `Zba`, `Zbb`, `Zcb`, and `Zcmp` extensions that would make generated code smaller and faster on that platform.  I'm also working on reviving the CH32V port, which has floating point and other potentially useful extensions.

On the other hand, this would require a few changes to mpy-cross to add extra -X parameters in order to specify the extensions set to compile viper/native code for.  Ideally such a scheme could be extended for Arm and slowly deprecate having one constant per architecture and floating point kind combination, in order to consolidate those into AArch32 and AArch64.

dpgeorge:
> Would a PR that bakes in the extensions set an MPY file is built for be accepted?
> 
> I thought about adding an extra raw code/sub-code element with the extensions set the code is compiled for, with the default being IMC (on both RV32 and RV64) if missing - to not break existing MPY files

I understand the motivation here (eg `armv7emsp` and `armv7emdp` would be just `armv7m` plus floating-point features), but (1) how exactly do you propose to do that, how would the mpy file specify the extensions without needing to bump the mpy version to v7; and (2) is there really any difference between having just one architecture option with many possibilities (as it is now with `armv7emXX`) and two options with less possibilities for each option?  In the end, if there are, say, 8 bits to encode all needed information about the arch and its features, it doesn't make a difference if that's one single 8-bit number, or two 4-bit numbers (for example).  (Or maybe there is a difference that I didn't think of.)

agatti:
> I understand the motivation here (eg `armv7emsp` and `armv7emdp` would be just `armv7m` plus floating-point features), but (1) how exactly do you propose to do that, how would the mpy file specify the extensions without needing to bump the mpy version to v7; and (2) is there really any difference between having just one architecture option with many possibilities (as it is now with `armv7emXX`) and two options with less possibilities for each option? In the end, if there are, say, 8 bits to encode all needed information about the arch and its features, it doesn't make a difference if that's one single 8-bit number, or two 4-bit numbers (for example). (Or maybe there is a difference that I didn't think of.)

TL;DR: There are simply too many extensions and sub-extensions combinations that are exposed by RISC-V processors (current and future) that using an integer to enumerate all seen combinations may not be the best choice.

Note: this may be much less relevant for Arm (or even not at all) so I'll describe this feature set from the RISC-V point of view, as going forward I think it may become more important as time goes by.

Given that RISC-V vendors have a certain degree of freedom of mixing and matching "main" extensions (e.g. C, M, F, D, etc.) on two distinct word sizes, having to encode all possible combinations when they are found in the wild may become a bit of a burden.  Not to mention that "main" extensions can be enumerated at runtime (see #15268), but sub-extensions cannot (not the correct terminology but it's good enough) [1].

For example, on RISC-V function prologues and epilogues require storing each register separately onto the stack and then load it back - two opcodes per register.  However, the RISC-V ISA folks standardised an extension to the `C` extension (I know it sounds weird) called `Zcmp`, which provides opcodes with equivalent functionality to Thumb's multi-register `push` and `pop`.  This is something you'd *really* want to have enabled if your CPU supports it - the Pico2 does, by the way - but unfortunately there's no easy way to know beforehand if that's available.  (Let's ignore setting an exception handler for illegal instructions, let the CPU execute one instruction per sub-extension and then record which extension opcodes triggered an illegal instruction exception)

Assuming the rationale is something that we can work with, the idea here is to:

1. Make `mpy-cross` aware that emitted code must make use of certain CPU extensions when packaging MPY files.
2. Make the emitter framework (for both native and viper) that generated code must make use of certain CPU extensions if they are explicitly enabled in `mpy-cross`.
3. Make the MPY loader reject files that claim to depend on one or more extensions that are not explicitly enabled in the runtime.

`mpy-cross` would be probably extended to recognise an additional `-X` argument, `-X exts=....` for example.  The extensions can be either comma- or underscore-separated.  So assuming you want the `Zba` and `Zcmp` extension if possible you'd use `-X exts=Zba,Zcmp`.  Whilst figuring out if an extension is supported in the runtime can currently be done via a set of defines, to prevent hardcoding extension names in `mpy-cross` a series of qstr lookups can be made to see whether those strings are actually present in a separate internal qstr pool or not.  I believe this applies for both native code (that will crash if an unsupported opcode is executed) and for bytecode that expects a certain extensions set to be present when emitted at runtime.

Then the emitter framework has to be extended to receive that extensions list from `mpy-cross` and make use of it.  Given that the set of defines and the entries in the separate qstr pool must match, the main problem here is how to get the parameters to the emitter in the first place (and let the emitter generate those opcodes, of course).  Code decorated with `@micropython.native` and `@micropython.viper` when executed at runtime is supposed to make use of all the enabled extensions anyway, as they're enabled via defines.

Finally, an extra check should be performed on MPY files when loaded to make sure the extensions set they claim to depend upon is actually compatible with what is made available by the runtime, and reject the file if not.  After all, there already are explicit checks on the MPY file version and its architecture identifier.  Whether the extensions are added in their own MPY section as raw ASCII text, as a 6-bits bitpacked list of indices (0b000000=A, 0b000001=B, etc.), or something else, that's not something I have a preference for yet.

I understand that this may be a bit too much and I've probably forgotten some important bits, but this is just a proposal and no work has been done on this yet.  If accommodating RV32/RV64 at such a level is out of scope for MicroPython, that's fine - maybe this can be reworked at a later time when the need for something like this arises :)

<hr>

[1] There's a [proposal](https://github.com/riscv/configuration-structure) for that but it's going nowhere - the moment they brought ASN.1 and MAC into the mix they really lost the plot if you ask me.

dpgeorge:
@agatti this PR in general looks good, and to make progress with it let's put aside the issue of RISC-V extensions, including floating point.  That's anyway being discussed in #17635.

Can you please rebase this PR on latest master and force push?  Then we can see how the CI goes and I'll do a full review.

agatti:
Sure, I've rebased it and CI didn't detect any failures for the QEMU/RV64 job.  It should be ready for review.

agatti:
Sorry, last minute change in the semihosting asm block (wrong offset for the exit code, now fixed).

dpgeorge:
Trying to build this locally I get a few errors:
```
/usr/lib/gcc/riscv64-elf/14.1.0/rv64imac/lp64/libgcc.a(_clzsi2.o): in function `__clzdi2':
/build/riscv64-elf-gcc/src/gcc-14.1.0/libgcc/libgcc2.c:690:(.text.__clzdi2+0x18): relocation truncated to fit: R_RISCV_HI20 against symbol `__clz_tab' defined in .rodata.__clz_tab section in /usr/lib/gcc/riscv64-elf/14.1.0/rv64imac/lp64/libgcc.a(_clz.o)
/usr/lib/gcc/riscv64-elf/14.1.0/rv64imac/lp64/libgcc.a(divsf3.o): in function `__divsf3':
/build/riscv64-elf-gcc/src/gcc-14.1.0/libgcc/soft-fp/divsf3.c:44:(.text.__divsf3+0x7e): relocation truncated to fit: R_RISCV_HI20 against `.L10'
make: *** [Makefile:269: build-VIRT_RV64/firmware.elf] Error 1
```

agatti:
I'm not sure Arch's toolchain is able to generate a working image.

Arch's toolchain support libraries haven't been compiled for the `medany` memory model, which allows for code being placed anywhere in the address space.  Said toolchain built its libraries using the `medlow` memory model instead (the default), so code can not be placed past the 0x7FFF_FFFF boundary [1] [2] [3].

In fact if you change the start address in the linkerscript from 0x8000_0000 to anything lower (say, 0x7000_0000) the code will link.  However, QEMU's `VIRT_RV64` RAM space starts at 0x8000_0000 (see https://github.com/qemu/qemu/blob/190d5d7fd725ff754f94e8e0cbfb69f279c82b5d/hw/riscv/virt.c#L103) so the code won't work.

[1] See also https://www.sifive.com/blog/all-aboard-part-4-risc-v-code-models for an explanation of the various RISC-V memory models if that's of any interest.

[2] ...or for a more rigorous specification: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#medium-low-code-model

[3] On Ubuntu, running `riscv64-unknown-elf-gcc -v` yields this fragment: `'CFLAGS_FOR_TARGET=-Os -mcmodel=medany' 'CXXFLAGS_FOR_TARGET=-Os -mcmodel=medany'`, Arch's `riscv64-elf-gcc -v` doesn't have any overrides.

<hr/>

@VynDragon do you know of a workaround for cases like this?  I think Zephyr solved it by rebuilding its toolchains forcing `medany` for RV64 (https://github.com/zephyrproject-rtos/zephyr/blob/0c4e9d29bbc1b142ca738f86facb6e88218b629d/cmake/compiler/gcc/target_riscv.cmake#L10), but maybe an alternative way to make this work has been found in the meantime.

dpgeorge:
OK, thanks for the information about medany.

It's unfortunate that RV64 is awkward in this way.  Is it possible to pass a flag to qemu to change the memory layout?  I vaguely remember something like that was possible when I was trying to get other qemu profiles to run, but maybe I misremember.

If there's no easy way to get Arch's toolchain to work in this situation then could you please add some info about this to the qemu README?  Is there a docker image (or similar) that we can recommend to use which does have a working toolchain?

agatti:
The only workaround I found to make it work with that toolchain is to figure out how to make the MicroPython image work when used as the machine's BIOS - that way the text segment is mapped at 0x2000_0000 instead and thus can be built using `medlow`.  However, that's probably a bit too involved for such a feature that seems to work when the toolchain works as expected.

For Docker, that's a possibility indeed...  Either let users work with that or point them to a reputable toolchain.  If there's no official/authoritative Docker image and the Embecosm toolchain - the one used by the RP2350 in RV32 mode - doesn't work, I'll come up with my Dockerfile and add that to the repo (in the beginning the QEMU RV32 port's README had instructions on how to set a custom toolchain up, I can re-use that to build the Dockerfile).

In any case I'll update the README to mention how to set things up if the toolchain doesn't work as intended.

agatti:
I've tried quite a few toolchains and short of rebuilding the one from https://github.com/riscv-collab/riscv-gnu-toolchain using `./configure --with-arch=rv64imac --with-cmodel=medany`, the one from Ubuntu that's also used in the CI is probably the easiest choice.

I've added a Dockerfile and some instructions to the README file - I don't have "proper" Docker set up on this machine as I'm already using Incus as my hypervisor/VM/container manager and I'm a bit scared to mess things up, so I provided working instructions for Podman instead.

Incidentally, this could have been an interesting opportunity to promote either https://github.com/mattytrentini/mpbuild for a hands-off solution for cases like this (if it will ever get around to build the QEMU port), or something like https://codeberg.org/agatti/micropython-builder if extreme flexibility is a requirement and customising things like usernames and timezones isn't a problem :)

dpgeorge:
> I've added a Dockerfile and some instructions to the README file

OK, thanks, that looks good.  I tried to test it but ran out of disk space (!) so I'll assume it's working correctly.

> Incidentally, this could have been an interesting opportunity to promote either ...

It might be a bit of effort for mpbuild to work with the qemu port.  And I'd need to investigate micropython-builder to see exactly what it does.  So for now, what you have with the Dockerfile seems like a good approach.

ports/qemu/README.md
```diff
@@ -37,6 +37,47 @@ Most pre-built toolchains should work out of the box, either coming from your
 Linux distribution's package manager, or independently packaged ones like
 [xPack](https://xpack.github.io/dev-tools/riscv-none-elf-gcc/).
 
+### RISC-V 64
+
+For RV64-based boards the build requires a bare metal RISC-V toolchain with GCC 10
+or later, either with multilib support or 64 bits specific (M, C, and Zicsr
+extensions must be supported, along with lp64 ABI).  In other words, if your
+compiler can build a binary that's compliant to any ratified RVA profiles,
+you're *might* be good to go: if running your cross-compiler `gcc` binary with
```
dpgeorge: you're -> you

docs/reference/mpyfiles.rst
```diff
@@ -58,7 +58,7 @@ If importing an .mpy file fails then try the following:
     sys_mpy = sys.implementation._mpy
     arch = [None, 'x86', 'x64',
         'armv6', 'armv6m', 'armv7m', 'armv7em', 'armv7emsp', 'armv7emdp',
-        'xtensa', 'xtensawin', 'rv32imc'][sys_mpy >> 10]
+        'xtensa', 'xtensawin', 'rv32imc', 'rv64imc'][sys_mpy >> 10]
```
dpgeorge, line 61: It looks like none of this native rv64imc architecture selection (and in .mpy files) is actually used, right?  I guess it's OK to include it here, but in that case it would be good to also update `tests/feature_check/target_info.py` to add the corresponding change that's made here.

ports/qemu/README.md
agatti: Thanks!  I've touched up that paragraph and the ones later to make it slightly clearer too.

docs/reference/mpyfiles.rst
agatti: Good point, this should be addressed in the latest revision of the PR.

agatti:
For the record, here's how much space the container image takes on my machine:

```bash
$ podman image tree mpy-qemu-riscv
Image ID: 12989dfa2914
Tags:     [localhost/micropython/mpy-qemu-riscv:latest]
Size:     6.89GB
Image Layers
├── ID: f9f52dc133e2 Size: 80.64MB Top Layer of: [docker.io/library/ubuntu:24.04]
└── ID: 9f2f57b1afcc Size: 3.405GB Top Layer of: [localhost/micropython/mpy-qemu-riscv:latest]
```

Anyway, `micropython-builder` - which should probably be called `micropython-dev-containers` as it doesn't just build stuff - is a set of scripts that builds the isolated environments I use to work on MicroPython (hence the hardcoded GitHub username and timezone/NTP settings in `mpy-base-image.yaml`, along with other tools' configuration).

It leverages `https://github.com/lxc/incus`, `https://github.com/mikefarah/yq`, and mDNS to build a series of containers that mimic target-specific CI environments as closely as possible, with said containers being accessible via SSH on the local machine.  If you happen to have incus already set up and you've changed `mpy-base-image.yaml` to suit your needs, `./rebuild-images.sh $MPY_ROOT <platform>` should create a container named `mpy-dev-<platform>` with $MPY_ROOT mounted under `/micropython` (that's writable from the container, by the way).  Then, `ssh dev@mpy-dev-<platform>.local` should get you in using the GitHub user's public key(s), and from now on it's as if you're running commands in a CI task container.

dpgeorge:
> Then, `ssh dev@mpy-dev-<platform>.local` should get you in using the GitHub user's public key(s), and from now on it's as if you're running commands in a CI task container.

Wow, that's neat!  @projectgus might be interested in that.

VynDragon:
> @VynDragon do you know of a workaround for cases like this? I think Zephyr solved it by rebuilding its toolchains forcing `medany` for RV64 (https://github.com/zephyrproject-rtos/zephyr/blob/0c4e9d29bbc1b142ca738f86facb6e88218b629d/cmake/compiler/gcc/target_riscv.cmake#L10), but maybe an alternative way to make this work has been found in the meantime.

Sorry no clue, only did rv32 on zephyr, and only rv64 on arch for linux target but with custom built toolchain.

dpgeorge:
> For the record, here's how much space the container image takes on my machine:

I eventually managed to build the container, it needed about 12GB free space to actually build it, even if the resulting size is much smaller.

I ran the `podman container create` command from the README, and that seemed to work (it returned a big hash).  But then actual make command failed:
```
$ podman container run -v .:/micropython --userns=keep-id -it mpy-qemu-riscv make -C ports/qemu -- BOARD=VIRT_RV64 test
Error: container create failed (no logs from conmon): conmon bytes "": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...
```
I've not used `podman` before, so don't know what's going on here.

agatti:
> I ran the `podman container create` command from the README, and that seemed to work (it returned a big hash).

That's correct, the hash is the id of the container it created.  The arguments in that command line are more or less the same as docker except for `--userns=keep-id`, which keeps the UID of the user inside the container the same as the UID of the user who ran podman (so filesystem permissions issues shouldn't happen).  Podman doesn't need root access to function unless it needs to expose hardware devices to a container, so it's usually a safer choice overall.

> But then actual make command failed:
> Error: container create failed (no logs from conmon): conmon bytes "": readObjectStart: expect { or n, but found , error found in #0 byte of ...||..., bigger context ...||...

That may be conmon acting up: https://github.com/containers/conmon/issues/475.  You may want to reinstall it if it's too old (current version on the Arch package repos is 1:2.1.13-1).

Does running `podman container run --log-level=trace <rest of the usual arguments>` provide any insight on what is happening?  For reference https://gist.github.com/agatti/897cf78d2f7ee5e5f67541dca305d77a is the output of the command with verbose mode when I run it on my Arch machine, and I have to stop it as it starts running tests.

dpgeorge:
> That may be conmon acting up: [containers/conmon#475](https://github.com/containers/conmon/issues/475). You may want to reinstall it if it's too old (current version on the Arch package repos is 1:2.1.13-1).

I had the latest version of conmon.

> Does running `podman container run --log-level=trace <rest of the usual arguments>` provide any insight on what is happening?

Yes, that allowed me to see it was using `runc` and I changed that to `crun` (by doing `pacman -S crun`) to match your set up.  That got me a bit further, but then it complained about OCI permissions for `containers/storage/overlay/<hash>`.  After battling that for a while -- it would always create a new directory there with uid/gid 100000:100000 which I could not access -- I just removed the `--userns=keep-id` option and it started working.  In particular the generated `build-VIRT_RV64` directory had the correct uid/gid for my proper Linux username.

So maybe remove the `--userns=keep-id` from the README that's added here??

agatti:
In the case of the filesystem permissions issues you've seen, if I don't add that argument things don't work on my end and I wonder if it is due to some other configuration changes I've made on this machine when setting Incus up.

That said, I've reformatted a spare laptop to make sure things work with Docker as well.  I've replaced the previous instructions with a shorter set of commands that should work both on Docker and Podman (just by changing the command name).  That assumes *no* extra configuration changes were made except for (optionally) setting up permissions for a non-root user to use Docker in the first place.

dpgeorge:
OK, thanks for updating the README.  I tested the new set of commands (using podman) and they work (although I didn't start from fresh, I had the old podman image there so it reused that).

Anyway, I think that's enough wrangling of toolchains for now!  CI works, and there's a documented way for developers to replicate CI.

dpgeorge (approved):

PR #15940, opened 2024-10-01
py/objtype: Don't delegate lookup of descriptor methods to __getattr__.

dpgeorge:
### Summary

When descriptors are enabled, lookup of the `__get__`, `__set__` and `__delete__` descriptor methods should not be delegated to `__getattr__`.  That follows CPython behaviour.

Issue initially found by @ntoll.

### Testing

Added a test for these cases.


projectgus (approved):
I'm not up with the relevant details of Python behaviour, but the code change and the test coverage look good to me!

PR #15942, opened 2024-10-01
samd/ports/SAMD21_XPLAINED_PRO: Update the deploy instructions and use the external flash for the file system.

robert-hh:
### Summary

The SAMD21 Xplained Pro board is delivered without a bootloader, making it impossible to run MicroPython. This PR tells, how to install a .uf2 bootloader and provides samples of bootloaders, which can be used for the board. In addition it
adapts and fixes MicroPython to use the on-board 1MByte flash for the file system. Commits in this PR:

- Update the deploy instructions
- Proved two bootloaders as Intel Hex files. One of the is a generic SAMD21 bootloader, that should work on any SAMD21 board. The other has the Volume labels and USB identifiers changed to reflect the Xplained Pro board, but is besides that identical to the generic bootloader.
- Use the 1 MByte flash for the file system.

### Testing

- Building the firmware. 
- Updating the bootloader following the steps of the instructions.
- Updating the Firmware using all methods to activate the bootloader.
- Creating the file system at the 1MB flash (done automatically at the first boot)
- Creating files , storing data and deleting files with a stress test.


dpgeorge:
Would it be better to instead change the linker script for this board so that the firmware is placed at the beginning of flash, so that a bootloader is not needed at all?

For a dev board like this one, I would assume that someone using it would perhaps want to use MicroPython on it as-is, without a bootloader.  After all, you need to use the debug port to program the bootloader initially.  So why not just program MicroPython using the debug port?

dpgeorge:
> That's possible, but then this board would behave different than all other SAMD ports, and it would need the .hex version of the firmware. At the moment there are only .uf2 versions available for the SAMD port.

OK, that's a good point.  So it would be easier to make this board behave like all other samd boards.

In that case it would be good to provide a bootloader.  But, I don't think it's appropriate to check it into this repo.  Instead can we just host it at micropython.org/resources/firmware?  I can copy it there and we can link to it in the deploy.md file.

robert-hh:
> Instead can we just host it at micropython.org/resources/firmware? I

That's good. I was wondering if there is a place to host support files likes this one. Should I keep the file in the PR?

robert-hh:
@dpgeorge The file deploy.md is updated to point to micropython.org/resources/firmware. And since it's a .hex files already, the step of converting a .bin file to Intel hex is not needed any more.

ports/samd/boards/SAMD21_XPLAINED_PRO/deploy_xplained_pro.md
```diff
@@ -0,0 +1,25 @@
+The SAMD21 Xplained Pro board is shipped without a bootloader. For use
+with MicroPyhton a suitable bootloader has be be installed first. The
+following procedure has been found to work and be simple:
+
+1. Get the bootloader file bootloader-xplained-pro-v3.16.0-15-gaa52b22.hex
+from https://micropython.org/resources/firmware.
```
dpgeorge, line 6: I think it would be better if this URL was the full URL of the .hex file, like:
```
Get the bootloader from https://micropython.org/resources/firmware/bootloader-xplained-pro-v3.16.0-15-gaa52b22.hex
```

dpgeorge (approved):

dpgeorge:
@robert-hh I have merged this and the .hex file for the bootloader is copied across to the micropython.org URL.

Actually, I think it would be good to add a README.md to the new `ports/samd/mboot/` directory which describes how that .hex file is generated (which repository it comes from), and also giving the license of that file (which would probably just be the license of the source code it was generated from).  Would you mind making a PR for that README?

robert-hh:
Sure. I will do that,

PR #15946, opened 2024-10-01
pydfu: Add support for USB serial number argument to handle multiple DFU devices

IlorDash:
In automated testing platforms that use multiple Micropython devices and run within CI/CD pipelines, there can be cases where two or more devices are left in DFU mode, causing the `pydfu` flashing process to fail with the following error:

```
pydfu.py", line 117, in init
    raise ValueError("Multiple DFU devices found")
ValueError: Multiple DFU devices found
```

To allow distinguishing between DFU devices a serial number argument was added. It can be retrieved using the dmesg:

```
usb 1-7.3: Product: STM32  BOOTLOADER
usb 1-7.3: Manufacturer: STMicroelectronics
usb 1-7.3: SerialNumber: 315C34853133
```

The serial number should be passed as is, for example:

```
./pydfu.py -u firmware.dfu --serial 315C34853133
```

With the serial number argument, users and CI/CD pipelines can now flash specific devices, even when multiple DFU devices are present.

### DIFF
--- a/tools/pydfu.py
+++ b/tools/pydfu.py
@@ -107,15 +107,23 @@ def find_dfu_cfg_descr(descr):
     return None
 
 
-def init(**kwargs):
+def init(serial_number, **kwargs):
     """Initializes the found DFU device so that we can program it."""
     global __dev, __cfg_descr
     devices = get_dfu_devices(**kwargs)
     if not devices:
         raise ValueError("No DFU device found")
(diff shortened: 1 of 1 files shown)


Josverl (commented):
I find `--serial` an overloaded term that can be simply confused with the serial port.
I would suggest to use a `--sn/--serial-number` to avoid that confusion

PR #15952, opened 2024-10-02
esp32: Apply the LWIP active TCP socket limit.

### LINKED ISSUE #12819: esp32: sockets become unresponsive after several larger transactions
hello,
im encountering massive issues with using microdot or microwebsrv2 with serving ~5 request of ~10kb each on esp32 wroom-32 (no additional memory). Worst case is an immediate unresponsiveness of any new connection. It stops serving the opened connections and doesnt accept any new ones which forces me to do a hard reset to recover functionality. Reducing backlog of async web_server to 2 doesnt really help.
This is issue is consistently replicable with the latest micropython build with idf 5.1 or the latest µP bin from the website.
When using 1.19.1 with idf 4.4.4 it runs for hours serving 100kb per website with two clients connected and requests send every 10s for hours without any issue.
I already made an example for the microdot dev available here: https://github.com/Tobinator32/microdot_crash_demo
Any advices how to get it up and running with the latest µP and IDF would be greatly appreciated.

### LINKED ISSUE #15844: esp32: Webserver Socket stops responding after a minute.

esp32 port, ESP32 and ESP32C3

ESP32 (lolin32-Lite, esp32 devkit v1)  and ESP32C3 (esp32c3 supermini)
both board have same issues

### MicroPython version
MicroPython v1.23.0 on 2024-06-02; ESP32C3 module with ESP32C3
in older versions than MicroPython v1.20.0 code works fine withou issues

MicroPython v1.19.1 works OK
MicroPython v1.20.0 works OK
MicroPython v1.21.0 NOT working
MicroPython v1.22.0 NOT working
MicroPython v1.23.0 NOT working

### Reproduction

Copy content of zip file to your ESP device. Set correct SSID and Password and reset your board.
[issue.zip](https://github.com/user-attachments/files/16995089/issue.zip)

The web server will offer you index.html, which cyclically send requests to server and server answers back.
in terminal you can see positions of virtual joysticks on webpage.
image

After about a minute, the server stops responding. There is enough free RAM, I don't know where the problem is, everything works correctly on older versions.

There is another version of the web server available in the Zip archive, without the use of threads, but the problem is exactly the same.

### Expected behaviour

I expect that the server will behave the same in all versions. It should respond indefinitely to web browser requests.

### Observed behaviour

In Chrome, using developer tools, I check the responses from the server. After 1-5 minutes, the server stops responding and sometimes restarts.
image

### Additional Information

Web server must run in background, without blocking of REPL and WebRepl

Code of webserver with threads:
```
import network
import usocket
import _thread
from os import stat, listdir
from time import sleep_ms
import gc


class WebServer:
    def __init__(self, web_folder='/www', port=80):
        self.WEB_FOLDER = web_folder
        self.MIMETYPES = {
            "txt"   : "text/plain",
            "htm"   : "text/html",
            "html"  : "text/html",
            "css"   : "text/css",
            "csv"   : "text/csv",
            "js"    : "application/javascript",
            "xml"   : "application/xml",
            "xhtml" : "application/xhtml+xml",
            "json"  : "application/json",
            "zip"   : "application/zip",
            "pdf"   : "application/pdf",
            "ts"    : "application/typescript",
            "ttf"   : "font/ttf",
            "jpg"   : "image/jpeg",
            "jpeg"  : "image/jpeg",
            "png"   : "image/png",
            "gif"   : "image/gif",
            "svg"   : "image/svg+xml",
            "ico"   : "image/x-icon",
            "cur"   : "application/octet-stream",
            "tar"   : "application/tar",
            "tar.gz": "application/tar+gzip",
            "gz"    : "application/gzip",
            "mp3"   : "audio/mpeg",
            "wav"   : "audio/wav",
            "ogg"   : "audio/ogg"
        }
        self.webserv_sock = None
        self.url_handlers = {}
        self.port = port

    def _file_exists(self, path):
        try:
            stat(path)
            return True
        except:
            return False

    def get_mime_type(self, filename):
        try:
            _, ext = filename.rsplit(".", 1)
            return self.MIMETYPES.get(ext, "application/octet-stream")
        except:
            return "application/octet-stream"
        
    def read_in_chunks(self, file_object, chunk_size=1024):
        while True:
            data = file_object.read(chunk_size)
            if not data:
                break
            yield data

    def serve_file(self, client, path):
        try:
            
            if path.startswith("/*GET_FILE"):
                file_path = path.replace("/*GET_FILE", "")
            else:
                if path == "/":
                    path = "/index.html"
                if '?' in path:
                    path = path.split('?')[0]
                file_path = self.WEB_FOLDER + path
            
            mime_type = self.get_mime_type(file_path)
            filestatus = 0 # 0=Not found  1=Found  2=found in GZip

            if self._file_exists(file_path + '.gz'):
                filestatus = 2
                file_path += '.gz'
            elif self._file_exists(file_path):
                filestatus = 1
                        
            if filestatus > 0:
                with open(file_path, 'rb') as file:
                    client.write(b'HTTP/1.1 200 OK\r\n')
                    client.write(b"Content-Type: " + mime_type.encode() + b"\r\n")
                    if filestatus == 2:
                        client.write(b'Content-Encoding: gzip\r\n')
                    client.write(b'\r\n')
                    for piece in self.read_in_chunks(file):
                        client.write(piece)
            else:
                client.write(b"HTTP/1.0 404 Not Found\r\n\r\nFile not found.")
        except OSError as e:
            print("OSError:", e)
            client.write(b"HTTP/1.0 500 Internal Server Error\r\n\r\nInternal error.")
        except Exception as e:
            print("Exception:", e)
            client.write(b"HTTP/1.0 500 Internal Server Error\r\n\r\nInternal error.")

    def handle(self, pattern):
        """Decorator to register a handler for a specific URL pattern."""
        def decorator(func):
            self.url_handlers[pattern] = func
            return func
        return decorator

    def client_handler(self, client):
        try:
            gc.collect()
            sleep_ms(0)
            request = client.recv(2048)
            if request:
                _, path, _ = request.decode("utf-8").split(" ", 2)
                for pattern, handler in self.url_handlers.items():
                    if path.startswith(pattern):
                        try:
                            gc.collect()
                            handler(client, path, request)
                        except Exception as e:
                            print("Handler Exception:", e)
                        client.close()
                        return
                # Default file serving if no handler matches
                self.serve_file(client, path)
        except Exception as e:
            sleep_ms(0)
            #print("Webserver Exception:", e)
        finally:
            client.close()

    def web_thread(self):
        while True:
            try:
                cl, addr = self.webserv_sock.accept()
                cl.settimeout(2)  # time in seconds
                self.client_handler(cl)
            except Exception as ex:
                sleep_ms(0)

    def start(self):
        addr = usocket.getaddrinfo('0.0.0.0', self.port)[0][-1]
        self.webserv_sock = usocket.socket()
        self.webserv_sock.setsockopt(usocket.SOL_SOCKET, usocket.SO_REUSEADDR, 1)
        self.webserv_sock.bind(addr)
        self.webserv_sock.listen(5)
        _thread.start_new_thread(self.web_thread, ())
        for interface in [network.AP_IF, network.STA_IF]:
            wlan = network.WLAN(interface)
            if not wlan.active():
                continue
            ifconfig = wlan.ifconfig()
            print("Web server spusten na adrese {}:{}".format(ifconfig[0], self.port))

    def stop(self):
        if self.webserv_sock:
            self.webserv_sock.close()
 
```

Code of webserver without threads:

```
import network
import socket
from os import stat, listdir
from time import sleep_ms
import gc
from micropython import alloc_emergency_exception_buf

# Define constant for registering the handler function
_SO_REGISTER_HANDLER = 20

class WebServer:
    def __init__(self, web_folder='/www', port=80):
        self.WEB_FOLDER = web_folder
        self.MIMETYPES = {
            "txt"   : "text/plain",
            "htm"   : "text/html",
            "html"  : "text/html",
            "css"   : "text/css",
            "csv"   : "text/csv",
            "js"    : "application/javascript",
            "xml"   : "application/xml",
            "xhtml" : "application/xhtml+xml",
            "json"  : "application/json",
            "zip"   : "application/zip",
            "pdf"   : "application/pdf",
            "ts"    : "application/typescript",
            "ttf"   : "font/ttf",
            "jpg"   : "image/jpeg",
            "jpeg"  : "image/jpeg",
            "png"   : "image/png",
            "gif"   : "image/gif",
            "svg"   : "image/svg+xml",
            "ico"   : "image/x-icon",
            "cur"   : "application/octet-stream",
            "tar"   : "application/tar",
            "tar.gz": "application/tar+gzip",
            "gz"    : "application/gzip",
            "mp3"   : "audio/mpeg",
            "wav"   : "audio/wav",
            "ogg"   : "audio/ogg"
        }
        self.webserv_sock = None
        self.url_handlers = {}
        self.port = port
        self.client_sockets = []

    def _file_exists(self, path):
        try:
            stat(path)
            return True
        except:
            return False

    def get_mime_type(self, filename):
        try:
            _, ext = filename.rsplit(".", 1)
            return self.MIMETYPES.get(ext, "application/octet-stream")
        except:
            return "application/octet-stream"
        
    def read_in_chunks(self, file_object, chunk_size=1024):
        while True:
            data = file_object.read(chunk_size)
            if not data:
                break
            yield data

    def serve_file(self, client, path):
        try:
            
            if path.startswith("/*GET_FILE"):
                file_path = path.replace("/*GET_FILE", "")
            else:
                if path == "/":
                    path = "/index.html"
                if '?' in path:
                    path = path.split('?')[0]
                file_path = self.WEB_FOLDER + path
            
            mime_type = self.get_mime_type(file_path)
            filestatus = 0 # 0=Not found  1=Found  2=found in GZip

            if self._file_exists(file_path + '.gz'):
                filestatus = 2
                file_path += '.gz'
            elif self._file_exists(file_path):
                filestatus = 1
                        
            if filestatus > 0:
                with open(file_path, 'rb') as file:
                    client.write(b'HTTP/1.1 200 OK\r\n')
                    client.write(b"Content-Type: " + mime_type.encode() + b"\r\n")
                    if filestatus == 2:
                        client.write(b'Content-Encoding: gzip\r\n')
                    client.write(b'\r\n')
                    for piece in self.read_in_chunks(file):
                        client.write(piece)
            else:
                client.write(b"HTTP/1.0 404 Not Found\r\n\r\nFile not found.")
        except OSError as e:
            print("OSError:", e)
            client.write(b"HTTP/1.0 500 Internal Server Error\r\n\r\nInternal error.")
        except Exception as e:
            print("Exception:", e)
            client.write(b"HTTP/1.0 500 Internal Server Error\r\n\r\nInternal error.")

    def handle(self, pattern):
        """Decorator to register a handler for a specific URL pattern."""
        def decorator(func):
            self.url_handlers[pattern] = func
            return func
        return decorator

    
    def accept_websocket(self, sock):
        try:
            client_sock, client_addr = sock.accept()
            #client_sock.setblocking(False)
            client_sock.settimeout(2)
            client_sock.setsockopt(socket.SOL_SOCKET, _SO_REGISTER_HANDLER, self.client_handler)
            self.client_sockets.append(client_sock)
            # Debug output
            #print("Received new connection from:", client_addr)
            #print(self.client_sockets)
        except Exception as e:
            print("Error accepting connection:", e)
    
    
    
    
    def client_handler(self, client_sock):
        try:
            request = client_sock.recv(2048)
            #print(request)
            if request:
                _, path, _ = request.decode("utf-8").split(" ", 2)
                for pattern, handler in self.url_handlers.items():
                    if path.startswith(pattern):
                        try:
                            gc.collect()
                            handler(client_sock, path, request)
                        except Exception as e:
                            print("Handler Exception:", e)
                        self.close_client(client_sock)
                        return
                # Default file serving if no handler matches
                self.serve_file(client_sock, path)
            self.close_client(client_sock)
        except Exception as e:
            print("Error in client_handler:", e)
            self.close_client(client_sock)

    def close_client(self, client_sock):
        try:
            client_sock.setsockopt(socket.SOL_SOCKET, _SO_REGISTER_HANDLER, None)
            client_sock.close()
        except Exception as e:
            print("Error closing client socket:", e)
        finally:
            if client_sock in self.client_sockets:
                self.client_sockets.remove(client_sock) 

    def start(self):
        alloc_emergency_exception_buf(100)
        addr = socket.getaddrinfo('0.0.0.0', self.port)[0][-1]
        self.webserv_sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        self.webserv_sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
        self.webserv_sock.bind(addr)
        self.webserv_sock.listen(1)
        self.webserv_sock.settimeout(2)
        self.webserv_sock.setsockopt(socket.SOL_SOCKET, _SO_REGISTER_HANDLER, self.accept_websocket)
        for interface in [network.AP_IF, network.STA_IF]:
            wlan = network.WLAN(interface)
            if not wlan.active():
                continue
            ifconfig = wlan.ifconfig()
            print("Web server started at address {}:{}".format(ifconfig[0], self.port))

    def stop(self):
        if self.webserv_sock:
            self.webserv_sock.setsockopt(socket.SOL_SOCKET, _SO_REGISTER_HANDLER, None)
            self.webserv_sock.close()
        for client_sock in self.client_sockets:
            self.close_client(client_sock)

```

### Code of Conduct

Yes, I agree

projectgus:
### Summary

This is a workaround for a bug in ESP-IDF where the configuration setting for LWIP maximum active TCP sockets (PCBs) is not applied. See https://github.com/espressif/esp-idf/issues/9670

Fixes cases where a lot of short-lived TCP connections can cause:

- Excessive memory usage (unbounded number of sockets in TIME-WAIT).
- Much higher risk of stalled connections due to repeated port numbers. The maximum number of active TCP PCBs is reduced from 16 to 12 to further reduce this risk.

This is not a 100% fix for the second point: a peer can still reuse a port number while a previous socket is in TIME-WAIT, and LWIP will reject that connection (in an RFC compliant way) causing the peer to stall.

* Closes #15844
* Closes #12819

(Note that this may not be a complete fix for every failure reported in those issues, but it should fix most of them. Will need additional info to reproduce any remaining problems, so probably worth opening a new issue.)

### Testing

Using the test program supplied in the issue report for #15844, and this test client:

```py
#!/usr/bin/env python
import time
import requests

IP = '192.168.66.125:80'

start = time.time()

for i in range(1000_000):
    r = requests.get(f'http://{IP}/*JOY;{i};0;0;0;0;0', timeout=20)
    r.close()
    print(i, time.time() - start)
    time.sleep(0.15)
```

* On master branch this test fails reliably, often in the first 500 iterations and always before 1000 iterations. Packet captures show a local port is reused by the client and that connection stalls (until either the client times out or the server socket ages out of TIME-WAIT).
* With the fix in this PR applied but the default max number of TCP PCBs (16), failures were seen between 1000 and 2000 iterations.
* With all changes in the PR applied, including reducing the max TCP PCBs to 12, a failure happened after 28,000 iterations and 1h40m (unclear if this was due to port reuse or another transient network issue). A follow up test has run over 7,000 iterations without failing. EDIT: This test stopped after 16,579 iterations due to reused port.

The frequency of TCP local port reuse depends on the client system (in this case desktop Linux), so may vary depending on the host OS and other network usage.

* The trade-off to sockets being cleaned up out of TIME-WAIT early is the possibility of delayed packets sent to the old socket being received by a new socket on the same port (see RFC1337). This patch compromises by only cleaning up sockets early when a lot of short-lived sockets are created, otherwise the full TIME-WAIT period should apply.
* As discussed in the linked issues it is possible to lower the MSL instead, so TIME-WAIT always ends more rapidly (say reduced from 120 seconds to 10 seconds). On quality network links this is probably fine, but may lead to problems on poor network links.
* LWIP could implement the behaviour of many other TCP/IP stacks: accept an out-of-window SYN on a TIME-WAIT socket so the reused port can be opened. Then port reuse would (mostly) not trigger stalled connections. See the [last heading in this comment](https://github.com/micropython/micropython/issues/15844#issuecomment-2384979336) for some discussion. However, out of scope for MicroPython to be able to change this - would need to be discussed with the LWIP maintainers, then wait for the change to be adopted by esp-lwip.

### DIFF
--- a/ports/esp32/CMakeLists.txt
+++ b/ports/esp32/CMakeLists.txt
@@ -74,5 +74,9 @@ list(APPEND EXTRA_COMPONENT_DIRS main_${IDF_TARGET})
 # Enable the panic handler wrapper
 idf_build_set_property(LINK_OPTIONS "-Wl,--wrap=esp_panic_handler" APPEND)
 
+# Patch LWIP memory pool allocators (see lwip_patch.c)
+idf_build_set_property(LINK_OPTIONS "-Wl,--wrap=memp_malloc" APPEND)
+idf_build_set_property(LINK_OPTIONS "-Wl,--wrap=memp_free" APPEND)
+
 # Define the project.
 project(micropython)
--- a/ports/esp32/boards/sdkconfig.base
+++ b/ports/esp32/boards/sdkconfig.base
@@ -133,3 +133,7 @@ CONFIG_NEWLIB_NANO_FORMAT=y
 # Due to limitations in the PMP system this feature breaks native emitters
 # so is disabled by default.
 CONFIG_ESP_SYSTEM_PMP_IDRAM_SPLIT=n
+
+# Further limit total sockets in TIME-WAIT when there are many short-lived
+# connections.
+CONFIG_LWIP_MAX_ACTIVE_TCP=12
--- a/ports/esp32/esp32_common.cmake
+++ b/ports/esp32/esp32_common.cmake
@@ -108,6 +108,7 @@ list(APPEND MICROPY_SOURCE_PORT
     network_wlan.c
     mpnimbleport.c
     modsocket.c
+    lwip_patch.c
     modesp.c
     esp32_nvs.c
     esp32_partition.c
--- a/ports/esp32/lwip_patch.c
+++ b/ports/esp32/lwip_patch.c
@@ -1,0 +1,64 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2024 Angus Gratton
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#include "lwip/memp.h"
+
+// This is a link-time patch to enforce the limit of max active TCP PCBs. A
+// workaround for upstream issue https://github.com/espressif/esp-idf/issues/9670
+//
+// Without this limit the number of TCP PCBs in TIME-WAIT is unbounded, which can
+// have two problems on systems with a lot of short-lived TCP connections:
+//
+// - Higher memory usage.
+// - Increased chance of stalled TCP connections due to port reuse.
+
+static unsigned active_tcp_pcbs;
+
+void *__real_memp_malloc(memp_t type);
+void __real_memp_free(memp_t type, void *mem);
+
+void *__wrap_memp_malloc(memp_t type) {
+    if (type != MEMP_TCP_PCB) {
+        return __real_memp_malloc(type);
+    }
+
+    if (active_tcp_pcbs >= MEMP_NUM_TCP_PCB) {
+        return NULL;
+    }
+
+    void *res = __real_memp_malloc(MEMP_TCP_PCB);
+    if (res != NULL) {
+        ++active_tcp_pcbs;
+    }
+    return res;
+}
+
+void __wrap_memp_free(memp_t type, void *mem) {
+    __real_memp_free(type, mem);
+    if (type == MEMP_TCP_PCB && mem != NULL) {
+        assert(active_tcp_pcbs);
+        --active_tcp_pcbs;
+    }
+}


projectgus:
Uploading some builds for generic boards here if anyone would like to test:

* [ESP32_GENERIC_TCP_limit_patch.zip](https://github.com/user-attachments/files/17220212/ESP32_GENERIC_TCP_limit_patch.zip)
* [ESP32_GENERIC_C3_TCP_limit_patch.zip](https://github.com/user-attachments/files/17220213/ESP32_GENERIC_C3_TCP_limit_patch.zip)
* [ESP32_GENERIC_S3_TCP_limit_patch.zip](https://github.com/user-attachments/files/17220216/ESP32_GENERIC_S3_TCP_limit_patch.zip)

dpgeorge:
Ahh, our good old friend `--wrap` :joy: 

This looks pretty good to me, a relatively small and self-contained workaround.

I will test it.

projectgus:
> Ahh, our good old friend --wrap 😂

Yeah :sweat_smile: . I originally thought I could do this by calling an internal LWIP API, but it was a ton fiddlier than this.

projectgus:
As this TIME-WAIT phenomenon seems to have haunted me on and off for the past decade, I've decided that I'll put some of my own time into it if the lwIP developers are amenable: https://lists.nongnu.org/archive/html/lwip-devel/2024-10/msg00000.html

dpgeorge:
> I've decided that I'll put some of my own time into it if the lwIP developers are amenable

Very good!

TRadigk:
> Uploading some builds for generic boards here if anyone would like to test:
> 
>     ...

Tried the C3 image on two different ESP32 C3 boards, both of them got stuck in bootloop:
```
Invalid image block, can't boot.
ets_main.c 333 
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x7 (TG0WDT_SYS_RST),boot:0xd (SPI_FAST_FLASH_BOOT)
Saved PC:0x40047ed2
SPIWP:0xee
mode:DIO, clock div:1
load:0x3c150020,len:0x3cdb8
load:0x3fc96600,len:0x3044
load:0x40380000,len:0x1ec
load:0x42000020,len:0x14c3d4
Invalid image block, can't boot.
ets_main.c 333 
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x7 (TG0WDT_SYS_RST),boot:0xd (SPI_FAST_FLASH_BOOT)
Saved PC:0x40047ed2
SPIWP:0xee
mode:DIO, clock div:1
load:0x3c150020,len:0x3cdb8
load:0x3fc96600,len:0x3044
load:0x40380000,len:0x1ec
load:0x42000020,len:0x14c3d4
Invalid image block, can't boot.
ets_main.c 333 
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x10 (RTCWDT_RTC_RST),boot:0xd (SPI_FAST_FLASH_BOOT)
SPIWP:0xee
mode:DIO, clock div:1
load:0x3c150020,len:0x3cdb8
load:0x3fc96600,len:0x3044
load:0x40380000,len:0x1ec
load:0x42000020,len:0x14c3d4
Invalid image block, can't boot.
ets_main.c 333 
```

flashing https://micropython.org/resources/firmware/ESP32_GENERIC_C3-20241003-v1.24.0-preview.378.gca6723b14.bin  brought both of them back to life.
The flash command used (path shortened):
```
esptool --chip esp32c3 --port COM8 --baud 460800 write_flash -z 0x0 "D:\...\ESP32_GENERIC_C3-20241003-v1.24.0-preview.378.gca6723b14.bin"
```

dpgeorge:
### Testing

I have tested this PR.  My test setup is:
- server running on the esp32 is `examples/network/http_server.py` from this repository, unchanged (ie not using the server from #15844)
- client on a PC running code below
- using ESP32_GENERIC board definition
- using a TinyPICO board

I ran the following tests:
- current master with ESP32_GENERIC for 10 minutes to prove it fails
- this PR with ESP32_GENERIC for 37 hours
- current master on PYBD-SF6 for 37 hours (parallel with above test)

#### Current master with ESP32_GENERIC

I ran this for 10 minutes and got 9 failures, all of which were `ConnectTimeout` errors from `requests`.

That's about 54 failures per hour.

#### This PR with ESP32_GENERIC

I ran this test for 37 hours.  There were 400297 iterations of the client loop, ie that many attempted requests.  Out of those, there were 49 which had an error, and all of those were of this form:
```
ConnectTimeout(MaxRetryError("HTTPConnectionPool(host='192.168.0.61', port=8080): Max retries exceeded with url: /*JOY;1338;0;0;0;0;0 (Caused by ConnectTimeoutError(<urllib3.connection.HTTPConnection object at 0x79bc338ef440>, 'Connection to 192.168.0.61 timed out. (connect timeout=20)'))"))
```
The first such error was at connection number 1338, after 466 seconds.  Then more or less evenly distributed over the 37 hours.

Failure rate here is 1.32 failures per hour, about 40 times less than the failure rate on master.

#### Current master on PYBD-SF6

As a comparison to this PR I also ran in parallel the same test on master on PYBD-SF6 (using the same PC as the client, same access point).  This board did nearly twice as many connections in the same time.

There were 715801 iterations of the client loop.  And 9 failures.

One failure (the first failure, at iteration 36012) was:
```
ConnectionError(ProtocolError('Connection aborted.', ConnectionResetError(104, 'Connection reset by peer')))
```
The other 8 failures (the first at iteration 53639, and more or less distributed throughout the 37 hours) were the same `ConnectTimeout` as above.

Failure rate here is 0.24 failures per hour.  But because this board had more iterations, to compare with ESP32_GENERIC the equivalent failure rate would be 0.14, which is about 10 times less than ESP32_GENERIC with this PR.

### Summary

- This PR definitely makes things a lot better on esp32, but it doesn't completely fix the problem.
- Probably this PR would approach the PYBD-SF6 result if `MEMP_NUM_TCP_PCB` were reduced to 5 to match PYBD-SF6.
-  I do not know if the `ConnectTimeout` errors correspond to port reuse, I only tracked the error raised by `requests` on the client.
- There are always going to be errors with networking, so clients and server must both be written to be as robust as possible.

It would be interesting to see if implementing "accept an out-of-window SYN on a TIME-WAIT socket" in lwIP would really reduce the failure rate to zero.  But that's a lot of work and probably impossible for us to integrate into MicroPython without forking lwIP and ESP-IDF.

### Client test code

Modified version of @projectgus client code above to track errors:
```py
import sys
import time
import requests

IP = sys.argv[1]

start = time.time()
errors = []

try:
    for i in range(1000_000):
        try:
            r = requests.get(f'http://{IP}/*JOY;{i};0;0;0;0;0', timeout=20)
            r.close()
        except Exception as er:
            print("Exception:", repr(er))
            errors.append((i, time.time() - start, er))
        print(i, time.time() - start, len(errors))
        time.sleep(0.15)
except KeyboardInterrupt:
    for er in errors:
        print(er)
```

projectgus:
> Tried the C3 image on two different ESP32 C3 boards, both of them got stuck in bootloop:

Sorry @TRadigk for the missing instruction. These .bin files attached above are only the app, and the .bin files published on the website are the full flash contents which incorporates some other binary files into one.

To test: First flash a "full" MicroPython firmware .bin from the website, and then do `esptool.py -p PORT -b 460800 write_flash 0x10000 micropython.bin` to flash one of the app binaries attached above. (The difference is the starting address.)

projectgus:
Very thorough testing, @dpgeorge! Nice.

>     * I do not know if the `ConnectTimeout` errors correspond to port reuse, I only tracked the error raised by `requests` on the client.

FWIW I think most of them probably do, the ones I tested with a running packet capture did 100% (on an otherwise quiet Wi-Fi network with good signal strength). However, fully agree with your other point that you can't ever assume a robust network and code should be prepared for some failures.

>     * Probably this PR would approach the PYBD-SF6 result if `MEMP_NUM_TCP_PCB` were reduced to 5 to match PYBD-SF6.

I think so too. Just want to make the point for anyone reading along that, similar to reducing MSL, reducing `MEMP_NUM_TCP_PCB` (even to 12) is really only a workaround. We're trading off reducing one kind of network robustness (correctly handling lost or delayed packets in the TIME-WAIT state) to increase another kind of network robustness (dealing with port reuse on repeated connections).

TRadigk:
> > Tried the C3 image on two different ESP32 C3 boards, both of them got stuck in bootloop:
> 
> Sorry @TRadigk for the missing instruction. These .bin files attached above are only the app, and the .bin files published on the website are the full flash contents which incorporates some other binary files into one.
> 
> To test: First flash a "full" MicroPython firmware .bin from the website, and then do `esptool.py -p PORT -b 460800 write_flash 0x10000 micropython.bin` to flash one of the app binaries attached above. (The difference is the starting address.)

Thank you so much, @projectgus for clearing this up. Now I was able to fully test and compare ESP32_GENERIC_C3-20241003-v1.24.0-preview.378.gca6723b14 and the provided "app". The result is pretty clear to me. On the previous version I could execute (at best) **73** consecutive requests until wifi broke down.
With the new app I could run **1000** requests successfully, and only sometimes my resilience was triggered (unexpectedly closed connection encountered), where I could recover within 20 ms (not tuned down to the very last millisecond, but "felt" like that this should be enough to create a new HTTP client and reopen connection).

dpgeorge (approved):

dpgeorge:
Although not a complete fix, this PR improves things dramatically.

Merged.

Issue #15944, opened 2024-10-01
FrameBuffer Initialisation fails with ValueError on `memoryview`


corranwebster:
### Port, board and/or hardware

RP2; Unix (but I believe this affects any board with framebuffer support)

### MicroPython version

MicroPython v1.23.0 on 2024-06-02; Raspberry Pi Pico with RP2040

### Reproduction

Run the code below in the REPL, or run a script with the same code in it:
``` python
import array
import framebuf
buf = array.array('H', bytearray(2*8*8))
sub_buf = memoryview(buf)[4:]
fbuf = framebuf.FrameBuffer(sub_buf, 4, 8, framebuf.RGB565, 8)
```

### Expected behaviour

I expected the code to work without error.

### Observed behaviour

I get an unexpected `ValueError` with no explanation:
``` python-console
>>> import array
>>> import framebuf
>>> buf = array.array('H', bytearray(2*8*8))
>>> sub_buf = memoryview(buf)[4:]
>>> fbuf = framebuf.FrameBuffer(sub_buf, 4, 8, framebuf.RGB565, 8)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: 
```

### Additional Information

The "what am I actually trying to do" here is to create a frame buffer on a subsection of another buffer using a memory view to avoid copying data, either:
- to clip out a sprite from a larger sprite-sheet buffer to display using `FrameBuffer.blit`
- to clip drawing to a sub-region of a larger image (for this you _have_ to have a memory view into the same underlying buffer)

This is likely related to, but different from #15571, as that is an issue with bit-alignment.  As with that issue, I think that the lines at fault are here: https://github.com/micropython/micropython/blob/17d82344581ad3a76033fae54c5d3304e17f185f/extmod/modframebuf.c#L317-L319

If you look at the underlying buffers, you have something like the following:
```
....xxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
```
where `.` are values in the underlying buffer that are skipped by the memoryview, `x` are values that should be visible in the framebuffer, and `s` are values which are skipped by the stride of 8.

It appears that what the `FrameBuffer` code is expecting is the following:
```
....xxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssssxxxx
ssss
```
even though the last 4 values are never written to.

## Work-around

[...]

PR #15954, opened 2024-10-02
extmod/modframebuf.c: Fix `FrameBuffer` stride corner-cases.

corranwebster:
### Summary

This is a fix for issue #15944 which handles corner cases in the FrameBuffer code when using stride values where the last line's stride may extend past the end of the underlying buffer.  This PR includes extra tests for these corner cases and a slightly more useful error message when the underlying buffer is in fact too small.

For example a GS8 format FrameBuffer with a width of 8, height of 2 and stride of 10 should be able to fit into a buffer of size 18 (10 bytes for the first horizontal line, and 8 bytes for the second---the full 10 bytes are not needed).

Similarly a 1 by 9 FrameBuffer in MONO_VLSB format with a stride of 10 should be able to fit into a buffer of length 11 (10 bytes for the first 8 lines, and then one byte for the 9th line.

Being able to do this is particularly important when cropping the corner of an existing FrameBuffer, either to copy a sprite or to clip drawing.

This PR should not break any existing code with the changed behaviour.

### Testing

This PR includes additional tests that exercise these corner-cases.

The tests have been run with the unix version of micropython, which should be sufficient as the code is in a generic core
module.


dpgeorge:
Thanks for the contribution!

> Being able to do this is particularly important when cropping the corner of an existing FrameBuffer, either to copy a sprite or to clip drawing.

Yes, that is very good motivation for this change.  I can see how that would fail without this PR.

I'm just thinking if there are any parts of the existing code that assume that there are stride bytes available on the last line.  Probably not but that needs to be checked.

dpgeorge (approved):

dpgeorge:
Thanks for the detailed analysis.  I'm happy with this change.

dpgeorge:
Tested on PYBD-SF2 hardware, the test here fails prior to this commit, then passes with the fix in this commit.

Rebased and merged in d1574de3b6a050ef07c39a131203305f731d39d8

PR #15956, opened 2024-10-03
docs: Add reference for reset and boot sequence, document factory reset for esp32 & rp2.

### LINKED ISSUE #15230: If boot.py doesn't exit then rp2 port will never initialise USB (soft-bricked)

Raspberry Pico

### MicroPython version

### Reproduction

I took my program code and stripped it down to a working repro. There are more lines that could be deleted but help with visualization. The content is saved to boot.py. There is no other file saved on the microcontroller.
With any version **prior** 1.23 this will yield a serial device in device manager, whenever the device is connected to the computer. With version 1.23 it will not and can only recovered by flashing a firmware version prior 1.23:

```python
import _thread
import asyncio  # type: ignore
from time import sleep
import gc
import machine

synchronousLoopAlive: bool = True
anyLoopAlive: bool = True
led = machine.Pin(25, machine.Pin.OUT)


async def yieldingBackgroundTaskOne():
    while anyLoopAlive:
        print("One")
        led.value(not led.value())

        await asyncio.sleep_ms(1000)


def synchronousTask():
    while synchronousLoopAlive:
        sleep(2)


if __name__ == "__main__":
    try:
        sleep(3)
        loop = asyncio.get_event_loop()
        gc.collect()

        programThread = _thread.start_new_thread(synchronousTask, ())
        print("program thread started")
        taskOne = loop.create_task(yieldingBackgroundTaskOne())
        print("task One started")

        loop.run_forever()
    except KeyboardInterrupt:
        synchronousLoopAlive = False
        anyLoopAlive = False
        loop.stop()
        print("INTERRUPTED - STOPPING EVERYTHING")
    except Exception as e:
        print(f"error occurred: {e}")

```

### Expected behaviour

Expected a serial device in device manager as in all versions prior 1.23.

### Observed behaviour

No serial device in device manager with micropython 1.23, but with any version prior 1.23.

### Additional Information

### Code of Conduct

Yes, I agree

projectgus:
### Summary

This started out as a task to document `boot.py` and `main.py` on the rp2 port (Closes #15230). However, I realised the boot sequence and reset types were documented per-port and sometimes inconsistent so I decided to consolidate these into a common reference page that each port can link to.

In follow-up commits I added port-specific factory reset docs for esp32 and rp2 ports. These ports don't have any MicroPython-specific recovery features like Safe Mode boots, the factory reset process is to erase the entire flash. However, now it's explicitly documented as such.

### DIFF
--- a/docs/esp32/tutorial/reset.rst
+++ b/docs/esp32/tutorial/reset.rst
@@ -1,0 +1,25 @@
+Factory reset
+=============
+
+If something unexpected happens and your ESP32-based board no longer boots
+MicroPython, then you may have to factory reset it. For more details, see
+:ref:`soft_bricking`.
+
+Factory resetting the MicroPython esp32 port involves fully erasing the flash
+and resetting the flash memory, so you will need to re-flash the MicroPython
+firmware afterwards and copy any Python files to the filesystem again.
+
+1. You will need the Espressif `esptool`_ installed on your system. This is the
+   same tool that you may have used to initially install MicroPython on your
+   board (see :ref:`installation instructions <esp32_flashing>`).
+2. Find the serial port name of your board, and then use esptool to erase the
+   entire flash contents::
+
+       esptool.py -p PORTNAME erase_flash
+
+3. Use esptool to flash the MicroPython file to your board again. If needed,
+   this file and flashing instructions can be found on the `MicroPython
+   downloads page`_.
+
+.. _esptool: https://github.com/espressif/esptool
+.. _MicroPython downloads page: https://micropython.org/download/?port=esp32
--- a/docs/library/pyb.rst
+++ b/docs/library/pyb.rst
@@ -147,10 +147,10 @@ Power related functions
    (internal oscillator) directly.  The higher frequencies use the HSE to
    drive the PLL (phase locked loop), and then use the output of the PLL.
 
-   Note that if you change the frequency while the USB is enabled then
-   the USB may become unreliable.  It is best to change the frequency
-   in boot.py, before the USB peripheral is started.  Also note that sysclk
-   frequencies below 36MHz do not allow the USB to function correctly.
+   Note that if you change the frequency while the USB is enabled then the USB
+   may become unreliable. It is best to change the frequency in :ref:`boot.py`,
+   before the USB peripheral is started. Also note that sysclk frequencies below
+   36MHz do not allow the USB to function correctly.
 
 .. function:: wfi()
 
@@ -205,8 +205,9 @@ Miscellaneous functions
 
 .. function:: main(filename)
 
-   Set the filename of the main script to run after boot.py is finished.  If
-   this function is not called then the default file main.py will be executed.
+   Set the filename of the main script to run after :ref:`boot.py` is finished.
+   If this function is not called then the default file :ref:`main.py` will be
+   executed.
 
    It only makes sense to call this function from within boot.py.
 
--- a/docs/reference/reset_boot.rst
+++ b/docs/reference/reset_boot.rst
@@ -244,6 +244,7 @@ two processes:
 The specific process(es) are different on each port:
 
 - :doc:`pyboard and stm32 port instructions </pyboard/tutorial/reset>`
+- :doc:`esp32 port instructions </esp32/tutorial/reset>`
 - :doc:`renesas-ra port instructions </renesas-ra/tutorial/reset>`
 - :doc:`rp2 port instructions </rp2/tutorial/reset>`
 - :doc:`wipy port instructions </wipy/tutorial/reset>`
--- a/docs/reference/reset_boot.rst
+++ b/docs/reference/reset_boot.rst
@@ -1,0 +1,260 @@
+Reset and Boot Sequence
+=======================
+
+A device running MicroPython follows a particular boot sequence to start up and
+initialise itself after a reset.
+
+.. _hard_reset:
+
+Hard reset
+----------
+
+Booting from hard reset is what happens when a board is first powered up, a cold
+boot. This is a complete reset of the MCU hardware.
+
+The MicroPython port code initialises all essential hardware (including embedded
+clocks and power regulators, internal serial UART, etc), and then starts the
+MicroPython environment. Existing :doc:`RTC </library/machine.RTC>`
+configuration may be retained after a hard reset, but all other hardware state
+is cleared.
+
+The same hard reset boot sequence can be triggered by a number of events such as:
+
+- Python code executing :func:`machine.reset()`.
+- User presses a physical Reset button on the board (where applicable).
+- Waking from deep sleep (on most ports).
+- MCU hardware watchdog reset.
+- MCU hardware brown out detector.
+
+The details of hardware-specific reset triggers depend on the port and
+associated hardware. The :func:`machine.reset_cause()` function can be used to
+further determine the cause of a reset.
+
+.. _soft_reset:
+
+Soft Reset
+----------
+
+When MicroPython is already running, it's possible to trigger a soft reset by
+:ref:`typing Ctrl-D in the REPL <repl_soft_reset>` or executing
+:func:`machine.soft_reset()`.
+
+A soft reset clears the Python interpreter, frees all Python memory, and starts
+the MicroPython environment again.
+
+State which is cleared by a soft reset includes:
+
+- All Python variables, objects, imported modules, etc.
+- Most peripherals configured using the :doc:`machine module
+  </library/machine>`. There are very limited exceptions, for example
+  :doc:`machine.Pin </library/machine.Pin>` modes (i.e. if a pin is input or
+  output, high or low) are not reset on most ports. More advanced configuration
+  such as :func:`Pin.irq()` is always reset.
+- Bluetooth.
+- Network sockets. Open TCP sockets are closed cleanly with respect to the other party.
+- Open files. The filesystem is left in a valid state.
+
+Some system state remains the same after a soft reset, including:
+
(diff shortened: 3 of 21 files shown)


projectgus:
@TradigK Interested if you have any comments on this PR. Does it look like it resolves your comment from here: https://github.com/micropython/micropython/issues/15230#issuecomment-2328175797 ?

ChiYuLmiunia (approved):

TRadigk:
@projectgus  I've checked all changes and fully agree with them. The boot sequence is clearly laid out and to be found on a top level 👍 (which helps to understand that this is not port port specific, but micropython specific). You added a way to recover from soft-bricking for every port allowing it to be found with the search and providing guidance to recover, which I also like very much.
I think this is a great addition to the docs! Thank you! 😎

projectgus:
After discussion with @dpgeorge have added a commit which deprecates `machine.soft_reset()` and recommends `sys.exit()` instead.

docs/reference/reset_boot.rst
```diff
@@ -0,0 +1,216 @@
...
+
+main.py
+^^^^^^^
+
+Similar to ``boot.py``, a file named ``main.py`` can be copied to the board's
+internal :ref:`filesystem <filesystem>`. If found then it is executed next in the
+startup process.
+
+``main.py`` is for any Python code that you want to run each time your device
+starts.
```
dpgeorge, line 119: It might be worth adding somewhere in this file that `main.py` execution is skipped when you soft-reset in raw-repl mode (see docs/reference/repl.rst for more details about raw-repl).

(I'm surprised we don't already document that fact somewhere... at least I couldn't find it.)

docs/esp32/tutorial/reset.rst
```diff
@@ -0,0 +1,25 @@
...
+1. You will need the Espressif `esptool`_ installed on your system. This is the
+   same tool that you may have used to initially install MicroPython on your
+   board (see :ref:`installation instructions <esp32_flashing>`).
+2. Find the serial port name of your board, and then use esptool to erase the
+   entire flash contents::
+
+       esptool.py -p PORTNAME erase_flash
+
+3. Use esptool to flash the MicroPython firmware to your board again. If needed,
+   this file and flashing instructions can be found on the `MicroPython
```
dpgeorge, line 21: file -> firmware

docs/library/pyb.rst
```diff
@@ -205,8 +205,9 @@ Miscellaneous functions
 
 .. function:: main(filename)
 
-   Set the filename of the main script to run after boot.py is finished.  If
-   this function is not called then the default file main.py will be executed.
+   Set the filename of the main script to run after :ref:`boot_py` is finished.
+   If this function is not called then the default file main.py will be
```
dpgeorge, line 209: Maybe also make `main.py` here a ref-link?

docs/reference/reset_boot.rst
```diff
@@ -0,0 +1,216 @@
...
+contents unless you really know what you're doing.
+
+.. _boot_py:
+
+boot.py
+^^^^^^^
+
+A file named ``boot.py`` can be copied to the board's internal :ref:`filesystem
+<filesystem>` using :doc:`mpremote <mpremote>` or :doc:`pyboard.py
+<pyboard.py>`.
```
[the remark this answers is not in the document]
andrewleech: Can they be both frozen on all ports now? I know it only worked on a few ports previously

docs/esp32/tutorial/reset.rst
```diff
@@ -0,0 +1,25 @@
...
+and resetting the flash memory, so you will need to re-flash the MicroPython
+firmware afterwards and copy any Python files to the filesystem again.
+
+1. You will need the Espressif `esptool`_ installed on your system. This is the
+   same tool that you may have used to initially install MicroPython on your
+   board (see :ref:`installation instructions <esp32_flashing>`).
+2. Find the serial port name of your board, and then use esptool to erase the
+   entire flash contents::
+
+       esptool.py -p PORTNAME erase_flash
```
andrewleech, line 18: For most users these days the auto-detection of port works well - it's often only advanced users that have more than one com port on their computer at any one time. As such I'd remove the need to "find serial port name" as new users wont know how to do this. 
If desired this could be split to show the basic `esptool.py erase_flash` first and then a short follow up paragraph on setting the port name as well if needed?

docs/reference/reset_boot.rst
```diff
@@ -0,0 +1,216 @@
...
+initialise itself after a reset.
+
+.. _hard_reset:
+
+Hard reset
+----------
+
+Booting from hard reset happens when the device is powered on, if Python code
+executes :func:`machine.reset()`, or if a physical Reset button is pressed on
+the board (where applicable).
```
andrewleech, line 14:
```suggestion

Booting from hard reset is what happens when a board is first powered up, a cold boot, this is a complete reset of the MCU. 

The same hard reset boot sequence can be triggered by a number of other sources such as:
* Python code executing :func:`machine.reset()`
* User presses a physical Reset button on the board (where applicable)
* MCU watchdog reset
* Waking from deep sleep (on most ports)
* MCU brown out detector (on some ports)
```

docs/reference/reset_boot.rst
```diff
@@ -0,0 +1,216 @@
...
+
+   Depending on the port, some hardware initialisation is delayed until after
+   ``boot.py`` exits. This includes initialising USB on the stm32 port and all
+   ports which support :ref:`machine.USBDevice <machine.USBDevice>`. On these
+   ports, output printed from ``boot.py`` may not be visible on the built-in USB
+   serial port until after ``boot.py`` finishes running.
+
+   The purpose of this late initialisation is so that it's possible to
+   pre-configure particular hardware in ``boot.py``, and then have it start with
+   the correct configuration.
```
andrewleech, line 104: It might be also worth mentioning it's common to configure network / wifi in boot.py though as that means it will be available on all boots for use with repl / mpremote when there's no main.py used.

andrewleech (commented):
These changes look great thanks for writing them!

docs/esp32/tutorial/reset.rst
projectgus: *Tries to imagine a world where there aren't nine serial devices plugged into a computer at all times*
image

:grin: In all seriousness, I think I agree. However, I suggest I submit a separate PR for this that also changes it in the esp32 tutorial. Also there are the download page instructions, although I'm not sure where these are generated from - maybe damien has to make that change somewhere internally.

docs/esp32/tutorial/reset.rst
projectgus: Have changed, but would like to note that this term was adopted from the linked esp32 tutorial which uses it extensively, for example "And then deploy the new firmware using::". (I did change "deploy" to "flash", though.)

(Can change the tutorial as well, but suggest I do it in the PR mentioned above.)

docs/esp32/tutorial/reset.rst
dpgeorge: > Also there are the download page instructions, although I'm not sure where these are generated from

They should be in the various `deploy.md` files (part of the `board.json` and `board.md` set of files).

docs/reference/reset_boot.rst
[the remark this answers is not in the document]
projectgus: > Also might be worth mentioning that boot.py (and main.py below) can be frozen into the firmware.

The note about freezing is in the list of `main.py` tips (but it mentions you can freeze both files.)

> Can they be both frozen on all ports now? I know it only worked on a few ports previously

Oh..? I've used the phrase "on most ports" and "and on some ports" in other places, should I add it here?

(This is frustratingly non-specific for users, but it's still useful to document where something may be port-dependent than not to document it at all.)

projectgus:
@andrewleech @dpgeorge Thanks for the great reviews! I've made some edits based on your suggestions, please take a look. (Apologies, I rebased so the "compare" feature in GitHub may show a lot of unrelated edits.)

docs/reference/reset_boot.rst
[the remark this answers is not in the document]
dpgeorge: It looks like all ports now use `pyexec_file_if_exists()` to run `boot.py` and `main.py`, and if freezing is enabled on that port then that function will first search frozen code.

So, all ports now support freezing of these two.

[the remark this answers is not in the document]
projectgus: Thanks for checking this, @dpgeorge

dpgeorge:
I have reviewed the changes here and am happy with this PR now.

Except the commit that deprecates `machine.soft_reset()` in favour of `sys.exit()`, that needs reconsideration in the light of https://github.com/micropython/micropython/pull/15486#issuecomment-2445553809

projectgus:
After discussion with @dpgeorge, `machine.soft_reset()` is no longer deprecated as the `sys.exit()` behaviour may change in the future. Updated the relevant commit in this PR to document the current behaviour of `sys.exit()`.

projectgus:
Also rewrote the suggested `try/except/finally` structure for main.py as the original version locked the user out of the REPL (reset in finally block). This version allows KeyboardInterrupt to drop to a REPL, a safer recommendation...

dpgeorge (approved):

dpgeorge:
> Also rewrote the suggested `try/except/finally` structure for main.py as the original version locked the user out of the REPL (reset in finally block). This version allows KeyboardInterrupt to drop to a REPL, a safer recommendation...

It's tricky to recommend something because in production you'd usually want the app to be as robust and secure as possible and hence not allow ctrl-C or anything to stop it, but that contradicts with having a convenient development cycle, eg using `mpremote`.

Anyway, what you have now is good!

PR #15891, opened 2024-09-21
zephyr: Track mainline zephyr

Ayush1325:
### Summary

The purpose of this PR is to track mainline zephyr for micropython zephyr port. It provides a base for upcoming Zephyr boards that are not supported in a stable Zephyr version yet. 

The reason for creating this is because in the past, BeagleBoard.org has maintained different forks for upcoming boards, which ended up becoming unmaintained with time due to having low manpower. So this PR also is sort of testing ground and staging for upcoming (and existing) boards from BeagleBoard.org, which already do have some random fork.

It can also help with supporting the latest version of Zephyr in the future.

This is built on top of the amazing work https://github.com/micropython/micropython/pull/9335

### Plans

The boards that are planned to be used for testing are the following:
1. [BeagleConnect Freedom](https://www.beagleboard.org/boards/beagleconnect-freedom)
2. [BeagleBone AI-64](https://www.beagleboard.org/boards/beaglebone-ai-64)
3. [BeagleY-AI](https://www.beagleboard.org/boards/beagley-ai)

Currently, the plan is to keep this as a Draft and create new PRs to move the board support to upstream once we support some future v3.8.0 of Zephyr or something.

PR #15959, opened 2024-10-03
ports: zephyr: Add support for BeagleConnect Freedom.

Ayush1325:
- Add build CI.
- Add the following capabilities
  - I2C
  - SPI
  - FLASH
- Cherry picked from https://github.com/micropython/micropython/pull/15891

### DIFF
--- a/ports/zephyr/boards/beagleconnect_freedom.conf
+++ b/ports/zephyr/boards/beagleconnect_freedom.conf
@@ -1,0 +1,7 @@
+# Hardware features
+CONFIG_I2C=y
+CONFIG_SPI=y
+
+# Flash drivers
+CONFIG_FLASH=y
+CONFIG_FLASH_MAP=y


tools/ci.sh
```diff
@@ -759,6 +759,7 @@ function ci_zephyr_build {
     docker exec zephyr-ci west build -p auto -b frdm_k64f
     docker exec zephyr-ci west build -p auto -b mimxrt1050_evk
     docker exec zephyr-ci west build -p auto -b nucleo_wb55rg # for bluetooth
+    docker exec zephyr-ci west build -p auto -b beagleconnect_freedom
```
dpgeorge, line 762: I don't think we should add this board to CI, because the zephyr CI here already takes a long time (and not all boards are tested anyway).

Ayush1325: @dpgeorge Where are the images present https://micropython.org/download/ built?

dpgeorge: They are build by the scripts in `tools/autobuild`.

But currently zephyr is not included in this, we don't provide pre-built firmware for any zephyr board.

We can start to do that, for boards that are known to work well with MicroPython and which have been tuned to enable the correct features.

dpgeorge (approved):

tools/ci.sh
Ayush1325: Sure, Let's stick to board configs for now and after some community testing, we can revisit providing pre-built firmware in the future.

PR #15976, opened 2024-10-07
docs/esp8266/tutorial/neopixel.rst: fix wrong index into array in example

poWer4aiX:
### Summary

While checking for neopixel support I noticed the example code provided in the docs. When testing the same by copy/paste, I got an array index exception. at line 51. The for loop on the `range(n)` is starting at `1` while the NP pixel array starts at `0`. The provided fix is shifting the index access by -1.

### Testing

The original sample code was throwing an indexing exception in line 51, the new one is running without.

### DIFF
--- a/docs/esp8266/tutorial/neopixel.rst
+++ b/docs/esp8266/tutorial/neopixel.rst
@@ -48,7 +48,7 @@ The following demo function makes a fancy show on the LEDs::
         # cycle
         for i in range(4 * n):
             for j in range(n):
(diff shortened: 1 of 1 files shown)


dpgeorge:
> The for loop on the `range(n)` is starting at `1` while the NP pixel array starts at `0`

That's not true, a `for j in range(n)` will start `j` at `0` and go up to `n - 1`.

So there must be something else wrong here.

Issue #3842, opened 2018-06-05
Cortex-M cores: pendSV handler need to modify the auto stacked XPSR value to clear LDM/STM progress


RockySong:
If pendSV is serviced during an LDM/STM instruction, which is interruptible and uses XPSR "ICI/IT" bitfield to save load/store progress, it may causes INVSTATE type hardfault when CPU finally returns to nlr_jump().
This is because on exception return, the Cortex-M core is expecting to resume LDM/STM when it sees the valid progress information in unstacked ICI/IT field of XPSR register 
However, pendSV returns to nlr_jump(). So there is no LDM/STM instruction and the Cortex-M CPU generates hardfault with INVSTATE.
My workaround, insert below instructions in the beginning of PendSV_Handler:
		"mov r2,	#0x01000000 \n"
		"str r2, [sp, #28]      \n"    // modify stacked XPSR to make sure possible LDM/STM progress is cleared

So when nlr_jump() is entered, it always has XPSR unstacked to a "clean" state.

Note: This may have something to do with the INVSTATE hardfault found in openMV project.

iabdalkader:
Is this going to break multithreading ? BTW another solution is to disable interruption of multicycle instructions:

> DISMCYCINT : Disables interruption of multi-cycle instructions. This increases the interrupt latency of the processor because load/store and multiply/divide operations complete before interrupt stacking occurs

I'm almost sure I tried this before but didn't work (maybe I didn't upload the right fw image when testing)

RockySong:
I did captured real case that LDM/STM is interrupted and ICI field saved its progress. I think DISMCYCINT should also work as it makes LDM/STM can't be interrupted.

iabdalkader:
@dpgeorge Any comment on this issue ?

xiabizha:
would any plan of micropython to support resolve this issue?

dpgeorge:
This has been fixed by ece950d9bed18554e945188c5a31908c8bfa9a7b, by removing preemptive NLR jump via PendSV.

Issue #3807, opened 2018-05-21
Hardfault when mp_parse is interrupted with pendsv.


iabdalkader:
Hello, I am getting a hardfault (INVSTATE) in nlrthumb.c if mp_parse is interrupted with pendsv. Is this behavior expected ? Could someone shed more light on the cause ?

Here's some info from GDB:
```
Breakpoint 1, HardFault_C_Handler (regs=0x2001fcc8) at stm32_it.c:144
144	    if (!pyb_hard_fault_debug) {
(gdb) backtrace 
#0  HardFault_C_Handler (regs=0x2001fcc8) at stm32_it.c:144
#1  <signal handler called>
#2  nlr_jump (val=0x200117e0) at ../../py/nlrthumb.c:85
#3  0x08092dd4 in mp_parse (lex=0x20012140, input_kind=(MP_PARSE_FILE_INPUT | unknown: 134819284)) at ../../py/parse.c:925
#4  0x0000004e in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)

(gdb) print /x *regs
$1 = {r0 = 0x200117e0, r1 = 0x20, r2 = 0x4, r3 = 0x4, r12 = 0x2000, lr = 0x8092dd5, pc = 0x808ff84, xpsr = 0x100f000}

(gdb) print /x *SCB
$3 = {CPUID = 0x411fc270, ICSR = 0x400803, VTOR = 0x8020000, AIRCR = 0xfa050300, SCR = 0x0, CCR = 0x70200, SHPR = {0x0, 0x0, 0x0, 0x0, 0x0, 
    0x0, 0x0, 0x0, 0x0, 0x0, 0xf0, 0x0}, SHCSR = 0x0, CFSR = 0x20000, HFSR = 0x40000000, DFSR = 0xb, MMFAR = 0x0, BFAR = 0x0, AFSR = 0x0, 
  ID_PFR = {0x30, 0x200}, ID_DFR = 0x100000, ID_AFR = 0x0, ID_MFR = {0x110030, 0x0, 0x1000000, 0x0}, ID_ISAR = {0x1101110, 0x2112000, 
    0x20232231, 0x1111131, 0x1310132}, RESERVED0 = {0x0}, CLIDR = 0x9000003, CTR = 0x8303c003, CCSIDR = 0xf00fe019, CSSELR = 0x0, 
[...]

Issue #294, opened 2014-02-15
pendsv needs to play nice with GC


dpgeorge:
Pressing CTRL-C on the pyboard generates an exception, in order to break out of any running code and return to the REPL.  It does this using pendsv, running at the lowest interrupt priority.  The GC is currently not protected from this, and so it can be interrupted at an arbitrary point.  This needs to be fixed.  Probably also other parts of py that modify global state (eg `rt_store_global`).

dpgeorge:
This has been fixed by ece950d9bed18554e945188c5a31908c8bfa9a7b, by removing preemptive NLR jump via PendSV.

PR #15988, opened 2024-10-10
stm32: Remove preemptive keyboard interrupt via PendSV

dpgeorge:
### Summary

Since the very beginning, the `stm32` port (first called `stm`, then `stmhal` now `stm32`) has had a special keyboard interrupt feature which works by using PendSV to break out of any running code.

This preemptive ctrl-C was added long ago in commit 01156d510c728408be4fd3100bcee18e8985ac00

The `stm32` port still uses that ancient code, and current does this:
- If ctrl-C is received on UART or USB then `mp_sched_keyboard_interrupt()` is called (like all other ports) to set a flag for the VM to see, and then the VM (or any loop calling `mp_handle_pending(true)`) will eventually handle the `KeyboardInterrupt` exception, raising it via NLR.
- If another ctrl-C is received while the existing scheduled keyboard interrupt is still pending (ie the VM has not yet processed it) then a special hard NLR jump will activate, that preempts the calling code.  Within the PendSV interrupt the stack is adjusted and an NLR jump is made to the most recent `nlr_push()` location.  This is like a normal NLR except *it is called from an interrupt context and completely annihilates the code that was interrupted by the IRQ*.

The reason for the preemptive interrupt was to handle ctrl-C properly before the VM was able to handle it.  Then when the esp8266 port came along, it needed a more port-agnostic way to handle ctrl-C, and so the mechanism was added to the VM and runtime to be able to check for pending interrupts.  Then the `stm32` port was updated to use this mechanism, with a fallback to the old preemptive way if a second ctrl-C was received (without the first one being processed).

This preemptive NLR jump is problematic because it can interrupt long-running instructions (eg store multiple, usually used at the end of a function to restore registers and return).  If such an instruction is interrupted the CPU remembers that with some flags, and can resume the long-running instruction when the interrupt finishes.  But the preemptive NLR does a long jump to different code at thread level and so the long-running interrupt is never resumed.  This leads to a CPU fault.

This fault has been previously reported in #3807 and #3842 (see also #294).  I was never able to reproduce the issues reported there, but finally (after 6 years!) I can reproduce it.

The bad behaviour of preemptive NLR jump has become apparent since #15486.  If you run the test suite over and over again on any stm32 board, it will eventually crash the board (see testing section below).

The point is, a skipped test now soft resets the board and so the board must run `boot.py` again.  The test runner may then interrupt the execution of `boot.py` with the double-ctrl-C that it sends (in `tools/pyboard.py:enter_raw_repl()`) in order to get the board into a known good state for the next test.  If the timing is right, this can trigger the preemptive PendSV in an unfortunate location and hard fault the board.

The fix in this PR is to just remove the preemptive NLR jump feature.  No other port has this feature and it's not needed, ctrl-C works very well on those ports.  Preemptive NLR jump is a very dangerous thing and is obviously buggy.

See this comment here which suggests what's done in this PR: https://github.com/micropython/micropython/issues/3842#issuecomment-396935908

### Testing

I have been using this to trigger the fault:
```
while true; do ./run-tests.py --target pyboard -d extmod stress float micropython basics misc --exclude extreme_exc --exclude mtime --exclude pow3_intbig --exclude recursive_iternext  --exclude float_parse --exclude math_constants --exclude vfs_blockdev_invalid || break; done
```
That can trigger the error on PYBv1.x, but it happens more regularly on PYBD-SF2/6.

With this PR, the board no longer hard faults.  (But it does show a new issue, it can still interrupt `boot.py` with a single ctrl-C, and that will be fixed in a follow-up PR.)

Could clear the CPU state for the long-running instruction as suggested in #3842.  But it's much simpler to just remove this code, which can have other issues as per #294.

### DIFF
--- a/ports/renesas-ra/pendsv.c
+++ b/ports/renesas-ra/pendsv.c
@@ -31,13 +31,6 @@
 #include "pendsv.h"
 #include "irq.h"
 
-// This variable is used to save the exception object between a ctrl-C and the
-// PENDSV call that actually raises the exception.  It must be non-static
-// otherwise gcc-5 optimises it away.  It can point to the heap but is not
-// traced by GC.  This is okay because we only ever set it to
-// mp_kbd_exception which is in the root-pointer set.
-void *pendsv_object;
-
 #if defined(PENDSV_DISPATCH_NUM_SLOTS)
 uint32_t pendsv_dispatch_active;
 pendsv_dispatch_t pendsv_dispatch_table[PENDSV_DISPATCH_NUM_SLOTS];
@@ -49,24 +42,6 @@ void pendsv_init(void) {
     #endif
     // set PendSV interrupt at lowest priority
     NVIC_SetPriority(PendSV_IRQn, IRQ_PRI_PENDSV);
-}
-
-// Call this function to raise a pending exception during an interrupt.
-// It will first try to raise the exception "softly" by setting the
-// mp_pending_exception variable and hoping that the VM will notice it.
-// If this function is called a second time (ie with the mp_pending_exception
-// variable already set) then it will force the exception by using the hardware
-// PENDSV feature.  This will wait until all interrupts are finished then raise
-// the given exception object using nlr_jump in the context of the top-level
-// thread.
-void pendsv_kbd_intr(void) {
-    if (MP_STATE_MAIN_THREAD(mp_pending_exception) == MP_OBJ_NULL) {
-        mp_sched_keyboard_interrupt();
-    } else {
-        MP_STATE_MAIN_THREAD(mp_pending_exception) = MP_OBJ_NULL;
-        pendsv_object = &MP_STATE_VM(mp_kbd_exception);
-        SCB->ICSR = SCB_ICSR_PENDSVSET_Msk;
-    }
 }
 
 #if defined(PENDSV_DISPATCH_NUM_SLOTS)
@@ -90,10 +65,7 @@ void pendsv_dispatch_handler(void) {
 __attribute__((naked)) void PendSV_Handler(void) {
     // Handle a PendSV interrupt
     //
-    // For the case of an asynchronous exception, re-jig the
-    // stack so that when we return from this interrupt handler
-    // it returns instead to nlr_jump with argument pendsv_object
-    // note that stack has a different layout if DEBUG is enabled
+    // Calls any pending functions in pendsv_dispatch_table.
     //
     // For the case of a thread switch, swap stacks.
     //
@@ -132,27 +104,6 @@ __attribute__((naked)) void PendSV_Handler(void) {
         ".no_dispatch:\n"
         #endif
 
-        // Check if there is an active object to throw via nlr_jump
-        "ldr r1, pendsv_object_ptr\n"
-        "ldr r0, [r1]\n"
-        "cmp r0, #0\n"
-        "beq .no_obj\n"
-        #if defined(PENDSV_DEBUG)
-        "str r0, [sp, #8]\n"            // store to r0 on stack
-        #else
-        "str r0, [sp, #0]\n"            // store to r0 on stack
-        #endif
-        "mov r0, #0\n"
-        "str r0, [r1]\n"                // clear pendsv_object
-        "ldr r0, nlr_jump_ptr\n"
-        #if defined(PENDSV_DEBUG)
-        "str r0, [sp, #32]\n"           // store to pc on stack
-        #else
-        "str r0, [sp, #24]\n"           // store to pc on stack
-        #endif
-        "bx lr\n"                       // return from interrupt; will return to nlr_jump
-        ".no_obj:\n"                    // pendsv_object==NULL
-
         #if MICROPY_PY_THREAD
         // Do a thread context switch
         "push {r4-r11, lr}\n"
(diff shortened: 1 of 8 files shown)


dpgeorge:
@iabdalkader @dlech are you using this preemptive PendSV interrupt feature?

dlech:
> you using this preemptive PendSV interrupt feature?

Nope, we use `mp_sched_keyboard_interrupt()`.

andrewleech:
That's a sneaky problem to track down @dpgeorge nice work! 

I've noticed that mpremote/pyboard.py do a double ctrl-c https://github.com/micropython/micropython/blob/197becbdcc0f8e6fd03ddaa921c32193a4a409db/tools/mpremote/mpremote/transport_serial.py#L124 was this originally done to use this hard interrupt? Maybe after sending the ctrl-c a slight timeout after sending would be warranted now as it could take a little longer for the vm to respond? Though as you say, if it's working well on other ports already it's probably not worth worrying about!

Regardless, this change improves consistency with other ports and simplifies behavior so I'm all for it!

dpgeorge:
> I've noticed that mpremote/pyboard.py do a double ctrl-c
> was this originally done to use this hard interrupt?

Yes it was.  Because originally a single ctrl-C would trigger the preemptive NLR jump, but then when softer interrupt handling was added, `pyboard.py` was updated to issue a double ctrl-c so that it could stop anything on `stm32`.

Implementation of generic ctrl-C handling and change `stm32` to try "soft" interrupt first: 124df6f8d07b53542b6960dbeea9b63bff469a67

Change `pyboard.py` to issue a double ctrl-C (most relevant to your question): bc1488a05f509cd5be8bfab9574babfcb993806f

IMO we no longer need the double ctrl-C, we should remove it.

ports/stm32/boards/common_isr_ram/common_isr.ld
```diff
@@ -14,7 +14,7 @@
        flash (in main.c) along with the isr_vector above.
     */
     . = ALIGN(4);
-    *(.text.pendsv_kbd_intr)
+    *(.text.mp_sched_keyboard_interrupt)
```
dpgeorge, line 17: @andrewleech it's curious that you only needed to put `pendsv_kbd_intr` in RAM.  That function calls `mp_sched_keyboard_interrupt` so I wonder why you didn't need to put the latter in RAM as well?

projectgus (approved):
This looks good to me! Rationale makes sense, and the other ports have been doing without it just fine.

Apart from the very legitimate problems with CPU register state, the idea of randomly jumping out of slow-running or stuck C code seems convenient on the one hand but terrifying on the other - too many opportunities for hardware and software to wind up left in some invalid state with no cleanup path.

dpgeorge:
> But it does show a new issue, it can still interrupt `boot.py` with a single ctrl-C, and that will be fixed in a follow-up PR.

See #15990 for full description of this problem, and a fix for that.

dpgeorge:
> IMO we no longer need the double ctrl-C, we should remove it.

See #15991 for that.

ports/stm32/boards/common_isr_ram/common_isr.ld
andrewleech: Oh that is interesting... maybe neither of them actually need to be in ram... or more likely my testing didn't catch the scenario where ctrl-c was being handled by the UART isr at the exact moment flash was locked. Just before or just after it wouldn't matter. 
I think I added that function to the ram list based just on review of things called from the isr function but forgot to follow it down the call stack to subsequent functions.

dpgeorge: OK, thanks for confirming.  I'll leave this as it is in this PR because it makes sense to have this function in RAM.

iabdalkader:
> This fault has been previously reported in https://github.com/micropython/micropython/issues/3807 and https://github.com/micropython/micropython/issues/3842 (see also https://github.com/micropython/micropython/issues/294). I was never able to reproduce the issues reported there, but finally (after 6 years!) I can reproduce it.
> @iabdalkader @dlech are you using this preemptive PendSV interrupt feature?

Sorry for the late reply. As a workaround for this issue, I've been clearing the stacked LDM/STM, but I've recently switched to the new VM abort feature and no longer need the hard-pendsv.

ports/renesas-ra/pendsv.c
```diff
@@ -132,27 +104,6 @@ __attribute__((naked)) void PendSV_Handler(void) {
         ".no_dispatch:\n"
         #endif
 
-        // Check if there is an active object to throw via nlr_jump
-        "ldr r1, pendsv_object_ptr\n"
-        "ldr r0, [r1]\n"
-        "cmp r0, #0\n"
-        "beq .no_obj\n"
-        #if defined(PENDSV_DEBUG)
```
iabdalkader, line 140: Note the comment just above this, about the debug layout has been invalid and now outdated. Glad to see the `PENDSV_DEBUG` removed, it should be removed from all ports.

dpgeorge:
> As a workaround for this issue, I've been clearing the stacked LDM/STM, but I've recently switched to the new VM abort feature and no longer need the hard-pendsv.

OK, great, thanks for confirming.

ports/renesas-ra/pendsv.c
dpgeorge: It is now removed from all ports (although I see there's still a few stray `-DPENDSV_DEBUG` remaining in Makefile's that should be removed).

As for the comments for the stack layout: feel free to open a PR to fix those.

iabdalkader: > As for the comments for the stack layout: feel free to open a PR to fix those.
I think we should remove, it was never correct; the stack layout doesn't change with different builds, the registers are stacked *by the hardware* so the layout is always the same. I did some extensive testing here:

https://github.com/micropython/micropython/issues/315#issuecomment-577692954

Issue #12337, opened 2023-08-30
New machine.CAN driver


projectgus:
## Summary

Planned cross-platform `machine.CAN` driver for Micropython.

## Background

MicroPython currently only has one driver for a CAN bus controller, the [pyb.CAN](https://docs.MicroPython.org/en/latest/library/pyb.CAN.html) driver on STM32. Various folks have contributed functional and useful pull requests with CAN drivers for other ports. Many implement a `machine.CAN` interface. Quite sensibly, these drivers have all based their new APIs on the `pyb.CAN` driver. These efforts are hugely appreciated 🙏.

@jimmo, @dpgeorge, and I discussed this recently. We all agree MicroPython should have a cross-platform `machine.CAN` interface. None of us want to use the `pyb.CAN` API for this:

* The conventions used by `pyb.CAN` don't match the best conventions of `machine` APIs.
* It's not very ergonomic, compared to libraries such as [CircuitPython canio](https://docs.circuitpython.org/en/latest/shared-bindings/canio/index.html) or [python-can](https://python-can.readthedocs.io/en/stable/index.html).
* Some functions assume behaviour of the ST CAN hardware peripheral that is not common across platforms. Trying to adapt these to each port in turn is messy and difficult for both users and maintainers.

## Goals

MicroPython should have a `machine.CAN` interface which is:

* Similar to other cross-platform `machine` APIs.
* Relatively easy to use.
* Still fast and light on resource usage.
* Easy to implement on new hardware platforms.
* Easy to extend for use with `asyncio`.
* Easy to integrate with higher-level CAN-based driver layers such as canopen.

Also:

* The interface should be easy for the existing un-merged driver PRs to adapt to.

## Plan

We plan to do the following:

1. [ ] Write a `machine.CAN` API as a docs PR, discuss with other maintainers and the community, and merge it.
2. [ ] Implement common cross-port interface helpers as `extmod/machine_can.c` (similar to `extmod/machine_pwm.c`, etc).
3. [ ] Create a compliant `machine.CAN` implementation on STM32 (this will be a thin wrapper around `pyb.CAN`, which will remain for legacy compatibility.)
4. [ ] Work with the existing PR authors as needed/wanted to merge `machine.CAN` drivers for new ports.

GitHub sponsor funds have been allocated for the first three steps, and I plan to work on it. Currently other work (USB device driver functionality) is the priority, but this is the next priority after that.

This issue post will be updated with overall status of this effort.

[...]

PR #15989, opened 2024-10-10
stm32: Refactor pyb.CAN implementation, add tests.

projectgus:
### Summary

I've been gradually working on stm32 as the first implementation of the new CAN API (#12337), and first step was to refactor the low-level part of the `pyb.CAN` driver to be reusable from `machine.CAN` as well. In the course of doing this I also wrote some simple automated tests for `pyb.CAN`.

~~After doing the refactors and adding the tests I discovered some small bugs in various STM32 boards. This PR contains bugfix commits also submitted in #16543, #16545, #16546, #16547. Only the top two commits in this branch are unique to this PR (but it's difficult to run the tests without the other fixes).~~

can.h API changes (only relevant for anyone writing C code against this API):

* Peripheral argument is now `CAN_HandleTypeDef *` instead of `pyb_can_obj_t *`, to decouple the two layers.
* Adds a new `can_rx_fifo_t` enum for the RX FIFO selection. The numeric value of this enum is the same as the previous values used by bxCAN (0, 1), but they are different to the enums previously used by FDCAN. The FDCAN ST HAL enums are buffer offset values but these are now moved into the implementation and are no longer exposed in the API.
* Responsibility for re-enabling RX interrupts has been moved up to the caller of `can_receive()`. See pyb_can.c for the necessary code.

As part of the refactoring some `pyb.CAN` behaviour changed on FDCAN, to make it the same as bxCAN:

* FDCAN peripherals no longer send rx callbacks with ids 3,4,5 when the bus error state changes (this behaviour wasn't documented).
* FDCAN peripherals now increment the error counters the same as bxCAN-based peripherals. 

### Testing

Ran the new multi-tests via command line:

```
./run-multitests.py -i pyb:/dev/ttyACM0 -i pyb:/dev/ttyACM1 -p2 multi_pyb_can/*.py
```

Tests pass on various pairings of the following boards:

* PyBoard V1.1 (bxCAN).
* NUCLEO_H723ZG board (FDCAN).
* ARDUINO_NICLA_VISION (STM32H747, FDCAN)
* NUCLEO_G474RE board (FDCAN, but G series has different peripheral design to H5+H7).

* I think the only real alternative would have been to ignore the existing `pyb.CAN` driver and make the STM32 `machine.CAN` driver from scratch. Given how much refactoring was needed here then that does look like a somewhat appealing alternative in hindsight, especially as I'm not done yet! However, this way the code size will be smaller for as long as both `pyb.CAN` and the future `machine.CAN` drivers are included in MicroPython.

### Follow-up work (not in this PR)

* Add multi-tests around CAN-FD mode.
* Finish the stm32 machine.CAN implementation based on this lower layer.

### DIFF
--- a/ports/stm32/can.c
+++ b/ports/stm32/can.c
@@ -28,29 +28,15 @@
 #include "py/mperrno.h"
 #include "py/mphal.h"
 #include "can.h"
+#include "pyb_can.h"
 #include "irq.h"
 
 #if MICROPY_HW_ENABLE_CAN
 
-void can_init0(void) {
-    for (uint i = 0; i < MP_ARRAY_SIZE(MP_STATE_PORT(pyb_can_obj_all)); i++) {
-        MP_STATE_PORT(pyb_can_obj_all)[i] = NULL;
-    }
-}
-
-void can_deinit_all(void) {
-    for (int i = 0; i < MP_ARRAY_SIZE(MP_STATE_PORT(pyb_can_obj_all)); i++) {
-        pyb_can_obj_t *can_obj = MP_STATE_PORT(pyb_can_obj_all)[i];
-        if (can_obj != NULL) {
-            can_deinit(can_obj);
-        }
-    }
-}
-
 #if !MICROPY_HW_ENABLE_FDCAN
 
-bool can_init(pyb_can_obj_t *can_obj, uint32_t mode, uint32_t prescaler, uint32_t sjw, uint32_t bs1, uint32_t bs2, bool auto_restart) {
-    CAN_InitTypeDef *init = &can_obj->can.Init;
+bool can_init(CAN_HandleTypeDef *can, int can_id, uint32_t mode, uint32_t prescaler, uint32_t sjw, uint32_t bs1, uint32_t bs2, bool auto_restart) {
+    CAN_InitTypeDef *init = &can->Init;
     init->Mode = mode << 4; // shift-left so modes fit in a small-int
     init->Prescaler = prescaler;
     init->SJW = ((sjw - 1) & 3) << 24;
@@ -67,7 +53,7 @@ bool can_init(pyb_can_obj_t *can_obj, uint32_t mode, uint32_t prescaler, uint32_
     uint32_t sce_irq = 0;
     const machine_pin_obj_t *pins[2];
 
-    switch (can_obj->can_id) {
+    switch (can_id) {
         #if defined(MICROPY_HW_CAN1_TX)
         case PYB_CAN_1:
             CANx = CAN1;
@@ -107,21 +93,16 @@ bool can_init(pyb_can_obj_t *can_obj, uint32_t mode, uint32_t prescaler, uint32_
     uint32_t pin_mode = MP_HAL_PIN_MODE_ALT;
     uint32_t pin_pull = MP_HAL_PIN_PULL_UP;
     for (int i = 0; i < 2; i++) {
-        if (!mp_hal_pin_config_alt(pins[i], pin_mode, pin_pull, AF_FN_CAN, can_obj->can_id)) {
+        if (!mp_hal_pin_config_alt(pins[i], pin_mode, pin_pull, AF_FN_CAN, can_id)) {
             return false;
         }
     }
 
     // init CANx
-    can_obj->can.Instance = CANx;
-    HAL_CAN_Init(&can_obj->can);
+    can->Instance = CANx;
+    HAL_CAN_Init(can);
 
-    can_obj->is_enabled = true;
-    can_obj->num_error_warning = 0;
-    can_obj->num_error_passive = 0;
-    can_obj->num_bus_off = 0;
-
-    __HAL_CAN_ENABLE_IT(&can_obj->can, CAN_IT_ERR | CAN_IT_BOF | CAN_IT_EPV | CAN_IT_EWG);
+    __HAL_CAN_ENABLE_IT(can, CAN_IT_ERR | CAN_IT_BOF | CAN_IT_EPV | CAN_IT_EWG);
 
     NVIC_SetPriority(sce_irq, IRQ_PRI_CAN);
     HAL_NVIC_EnableIRQ(sce_irq);
@@ -129,10 +110,9 @@ bool can_init(pyb_can_obj_t *can_obj, uint32_t mode, uint32_t prescaler, uint32_
     return true;
 }
 
-void can_deinit(pyb_can_obj_t *self) {
-    self->is_enabled = false;
-    HAL_CAN_DeInit(&self->can);
-    if (self->can.Instance == CAN1) {
+void can_deinit(CAN_HandleTypeDef *can) {
+    HAL_CAN_DeInit(can);
+    if (can->Instance == CAN1) {
         HAL_NVIC_DisableIRQ(CAN1_RX0_IRQn);
         HAL_NVIC_DisableIRQ(CAN1_RX1_IRQn);
         HAL_NVIC_DisableIRQ(CAN1_SCE_IRQn);
@@ -140,7 +120,7 @@ void can_deinit(pyb_can_obj_t *self) {
         __HAL_RCC_CAN1_RELEASE_RESET();
         __HAL_RCC_CAN1_CLK_DISABLE();
     #if defined(CAN2)
-    } else if (self->can.Instance == CAN2) {
+    } else if (can->Instance == CAN2) {
         HAL_NVIC_DisableIRQ(CAN2_RX0_IRQn);
         HAL_NVIC_DisableIRQ(CAN2_RX1_IRQn);
         HAL_NVIC_DisableIRQ(CAN2_SCE_IRQn);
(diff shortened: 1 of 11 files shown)


projectgus:
@iabdalkader @H-Grobben It's been a while, but this PR touches a bunch of driver code you've both worked on. Would appreciate any comments if you have the time.

projectgus:
> stm32:    +0 +0.000% PYBV10

:raised_eyebrow:  It seems very coincidental that this PR refactors so much code and ends up with exactly the same binary size...

iabdalkader:
> @iabdalkader @H-Grobben It's been a while, but this PR touches a bunch of driver code you've both worked on. Would appreciate any comments if you have the time.

It's been years since I worked on FDCAN and CAN, so I don't remember much, but I'll try to take a look as soon as I have some free time. @kwagyeman is very interested in CAN support he might also want to take a look.

kwagyeman:
@projectgus - I'll update my PR for CAN on the IMX after an API is approved and merged.

projectgus:
> @projectgus - I'll update my PR for CAN on the IMX after an API is approved and merged.

Sounds great. Appreciate your patience with the CAN support, @kwagyeman.

dpgeorge:
I didn't get time for a full review of this yet, but one thing I suggest is renaming the `multi_pyb_can` directory to just `multi_pyb`.  It can potentially include tests for other thing in the `pyb` module in the future.

projectgus:
> I didn't get time for a full review of this yet, but one thing I suggest is renaming the `multi_pyb_can` directory to just `multi_pyb`. It can potentially include tests for other thing in the `pyb` module in the future.

I can change this, but I wanted to check you're sure about it? From a usability perspective, it seems fiddly: Once other tests are added, it'd be necessary to run `./run-multitests.py multi_pyb/pyb_can_*.py` most of the time as the PyBoard CAN tests require CAN transceivers connected to the boards in the test. Other `pyb` module tests may have very different environment requirements.

projectgus:
To make the review and merge process a bit simpler I've split out four individual PRs with the bugfixes. The changes unique to this PR are the two head commits in this branch. The other fixes are still in this branch because it's hard to run the tests without them.

H-Grobben:
User modules can not access the 'pyb_can_obj_t' struct if it is defined in the pyb_can.c file. Is the struct also defined as a machine.CAN struct, in a .h file, which then be used by user modules?

If it's put in the pyb_can.h file, it is still usable.

I will try to review the code and run some tests on the can changes if I can find the time.

projectgus:
Thanks for taking a look, @H-Grobben.

> User modules can not access the 'pyb_can_obj_t' struct if it is defined in the pyb_can.c file.

Ah, sorry - I actually hadn't thought of the use case that a user C module might be using `pyb_can_obj_t`. I'll move it to pyb_can.h to minimise headaches on upgrade.

> Is the struct also defined as a machine.CAN struct, in a .h file, which then be used by user modules?

Everything in `pyb_can.*` is to-be-eventually-deprecated code for `pyb` module compatibility.  The new `machine.CAN` driver is still pending (unfortunately), but it will have a different implementation that uses `can.h` but not `pyb_can.h`.

I don't think any of the current `machine` module hardware drivers expose a C interface, so probably the new CAN driver won't do this either (apart from the new `can.h`). Can explain more about your user C module's use case, please? Is the code published anywhere?

> -- update: need more time, but initial test locks up the micropython on the G4, with our custom c module, with this PR

It wouldn't surprise me that there might be other subtle behaviour changes that impact a custom C module. If you figure out what's happening then I'm happy to try and tweak things to keep compatibility intact.

kwagyeman:
@projectgus - Is CAN for the machine module ever going to be completed? It's been over a year now on this right?

projectgus:
> @projectgus - Is CAN for the machine module ever going to be completed?

I know it's frustrating, I'm sorry it's dragging so long. I have not forgotten about it, but there's been a lot of other things to take care of as well. There has been some progress behind the scenes, hopefully will be visible in a PR draft soon.

projectgus:
@H-Grobben Have rebased this PR and moved the `pyb_can_obj_t` struct into pyb_can.h

projectgus:
(Rebased, not other changes.)

dpgeorge:
I have run the tests from this PR on PYBV10+PYBV10 and PYBV10+PYBD_SF6 (so just standard CAN, not FDCAN).  They pass before the refactoring in this PR and with this PR.  Very good!

I do like how the `can.h` API is now operate on `CAN_HandleTypeDef*` rather than `pyb_can_obj_t*`.  That's a good refactoring.

ports/stm32/can.c
```diff
@@ -312,119 +310,124 @@ HAL_StatusTypeDef CAN_Transmit(CAN_HandleTypeDef *hcan, uint32_t Timeout) {
...
 
-    switch (*state) {
-        case RX_STATE_FIFO_EMPTY:
-            __HAL_CAN_DISABLE_IT(&self->can,  (fifo_id == CAN_FIFO0) ? CAN_IT_FMP0 : CAN_IT_FMP1);
-            irq_reason = MP_OBJ_NEW_SMALL_INT(0);
-            *state = RX_STATE_MESSAGE_PENDING;
-            break;
-        case RX_STATE_MESSAGE_PENDING:
-            __HAL_CAN_DISABLE_IT(&self->can, (fifo_id == CAN_FIFO0) ? CAN_IT_FF0 : CAN_IT_FF1);
-            __HAL_CAN_CLEAR_FLAG(&self->can, (fifo_id == CAN_FIFO0) ? CAN_FLAG_FF0 : CAN_FLAG_FF1);
```
dpgeorge, line 339: I can see where `__HAL_CAN_DISABLE_IT` has been refactored to (the `instance->IER &= ~...` code below).

But, I can't see where `__HAL_CAN_CLEAR_FLAG` has been moved to.  For FDCAN it's replaced with `instance->IR` but for standard CAN I don't see any place that clears the interrupt flags.

ports/stm32/can.c
```diff
@@ -219,16 +211,22 @@ int can_receive(CAN_HandleTypeDef *can, int fifo, CanRxMsgTypeDef *msg, uint8_t
     // Release (free) message from FIFO
     *rfr |= CAN_RF0R_RFOM0;
 
+    // Re-enable any interrupts that were disabled in RX IRQ handlers
+    can_enable_rx_interrupts(can, fifo, can_rx_pending(can, fifo) == 0);
```
dpgeorge, line 215: Note that in the original implementation (prior to this refactoring), RX interrupts (message pending, full, overflow) were only re-enabled if a Python callback was registered, eg `self->rxcallback0 != mp_const_none`.

But now, they are always re-enabled.

I guess this changes behaviour in a subtle way?  Because now the RX state machine (`rx_state0`, `rx_state1`) is managed even if there's no Python callback (prior, it was not managed).

Not sure if this matters too much, but worth thinking a bit more about it.

ports/stm32/pyb_can.c
```diff
@@ -27,6 +27,8 @@
 #include <string.h>
 #include <stdlib.h>
 
+#include "py/mpconfig.h"
```
dpgeorge, line 30: You don't need to include this, it's included by all other `py/xxx.h` headers.

dpgeorge (commented):
I've now done a full review of this PR.  Apart from the above comments, it looks good.

ports/stm32/can.c
projectgus: It was intended to be done on the next line:

```c
        instance->IER &= ~full_int;
        *fifo_reg &= ~full_flag;
```

`fifo_reg` is either RF0R or RF1R and `full_flag` is either `CAN_RF0R_FULL0` or `CAN_RF1R_FULL1`. This is the equivalent to the register.

However this is wrong as this bit is "write one to clear", will fix and make the behaviour clearer.

projectgus: I've rewritten this to use the HAL macros again, as less prone to mistakes like missing "write 1 to clear". See what you think.

ports/stm32/can.c
projectgus: As discussed, this is now delegated to the caller in pyb_can.c - which knows if interrupts are needed or not.

projectgus:
Marked as draft again as I need to add a bit more functionality to the lower layer for machine.CAN to work.

projectgus:
> Marked as draft again as I need to add a bit more functionality to the lower layer for machine.CAN to work.

Actually this is probably OK, can always do a second refactor PR if needed.

ports/stm32/pyb_can.h
```diff
@@ -0,0 +1,58 @@
...
+    mp_obj_base_t base;
+    mp_obj_t rxcallback0;
+    mp_obj_t rxcallback1;
+    mp_uint_t can_id : 8;
+    bool is_enabled : 1;
+    byte rx_state0;
+    byte rx_state1;
+    uint16_t num_error_warning;
+    uint16_t num_error_passive;
+    uint16_t num_bus_off;
```
IhorNehrutsa, line 46: Is it necessary to count hardware errors?
It may be enough to perform bus recovering inside the driver.

See RM0008 Reference manual 24.7.6 Error management
See https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/twai.html#error-states-and-counters

dpgeorge: This is useful for the user to get an idea of if/when the bus is failing.  They can access these "performance counters" via `can.info()`.

(This is part of the Python API of `pyb.CAN`, so can't really be changed.)

ports/stm32/can.c
dpgeorge: Looks good now.

dpgeorge (approved):
I retested on PYBV10+PYBV10.  The new multitests still pass.

PR #15990, opened 2024-10-10
tests/run-tests.py: Wait for soft reset if a target skips a test.

dpgeorge:
### Summary

Commit 69c25ea8653566ec97690b5121bd10b753c89426 made raising `SystemExit` do a soft reset (on bare-metal targets).  This means that any test which is skipped by a target (by raising `SystemExit`) will trigger a soft reset on that target, and then it must execute its startup code, such as `boot.py`.

If the timing is right, this startup code can be unintentionally interrupted by the test runner when preparing the next test, because the test runner enters the raw REPL again via a Ctrl-C Ctrl-A ctrl-D sequence (in `Pyboard.enter_raw_repl()`).

When this happens (`boot.py` is interrupted) the target may not be set up correctly, and it may (in the case of stm32 boards) flash LEDs and take extra time, slowing down the test run.

Fix this by explicitly waiting (with a short 1 second timeout) for the target to finish its soft reset when it skips a test.

### Testing

Run the test suite many times over and stm32 boards will occasionally have their boot.py interrupted and flash their LEDs to indicate that.  With this patch applied you don't ever see that behaviour.

### DIFF
--- a/tests/run-tests.py
+++ b/tests/run-tests.py
@@ -824,6 +824,12 @@ def run_tests(pyb, tests, args, result_dir, num_threads=1):
         output_mupy = run_micropython(pyb, args, test_file, test_file_abspath)
 
         if output_mupy == b"SKIP\n":
+            if pyb is not None and hasattr(pyb, "read_until"):
+                # Running on a target over a serial connection, and the target requested
+                # to skip the test.  It does this via a SystemExit which triggers a soft
+                # reset.  Wait for the soft reset to finish, so we don't interrupt the
+                # start-up code (eg boot.py) when preparing to run the next test.
+                pyb.read_until(1, b"raw REPL; CTRL-B to exit\r\n")
             print("skip ", test_file)
             skipped_tests.append(test_name)
             return


tests/run-tests.py
```diff
@@ -824,6 +824,12 @@ def run_one_test(test_file):
         output_mupy = run_micropython(pyb, args, test_file, test_file_abspath)
 
         if output_mupy == b"SKIP\n":
+            if pyb is not None and hasattr(pyb, "read_until"):
+                # Running on a target over a serial connection, and the target requested
+                # to skip the test.  It does this via a SystemExit which triggers a soft
+                # reset.  Wait for the soft reset to finish, so we don't interrupt the
+                # start-up code (eg boot.py) when preparing to run the next test.
+                pyb.read_until(1, b"raw REPL; CTRL-B to exit\r\n", timeout=1)
```
andrewleech, line 832: I'm not sure if 1 second is really long enough here? 

I'm thinking of the likes of a wifi board that's used for multitests where it's common to have wifi & network setup in boot.py, this can take longer. 

Considering this will continue sooner when that startup message is received I'd be inclined to set a reasonably long timeout for the less common cases that need it?

dpgeorge: This is not the total timeout, but actually the timeout between characters received over the serial connection.

Well, you're right that it could still take a while, with nothing printed for many seconds.

I can increase it.  All boards are expected to do a soft reset after skipping a test, so it should only time out if something is wrong.

dpgeorge: I've now changed this to use the default timeout of 10 seconds.

Tested on stm32, rp2 and esp32 boards and the wait for the soft reset after a skip is quick (because they don't do much in their boot.py).

andrewleech: That seems reasonable, looks good!

projectgus (approved):
Looks like a good change to me, I'm pretty sure I've seen this behaviour before.

dpgeorge:
Thanks for the reviews!

PR #16006, opened 2024-10-13
esp32/mphalport: Always poll stdin ringbuffer to include UART use.

andrewleech:
### Summary

Fixes https://github.com/micropython/micropython/discussions/15994#discussioncomment-10927232

### DIFF
--- a/ports/esp32/mphalport.c
+++ b/ports/esp32/mphalport.c
@@ -109,12 +109,6 @@ uintptr_t mp_hal_stdio_poll(uintptr_t poll_flags) {
     uintptr_t ret = 0;
     #if MICROPY_HW_ESP_USB_SERIAL_JTAG
     usb_serial_jtag_poll_rx();
-    if ((poll_flags & MP_STREAM_POLL_RD) && ringbuf_peek(&stdin_ringbuf) != -1) {
-        ret |= MP_STREAM_POLL_RD;
-    }
-    if (poll_flags & MP_STREAM_POLL_WR) {
-        ret |= MP_STREAM_POLL_WR;
-    }
     #endif
     #if MICROPY_HW_USB_CDC
     ret |= mp_usbd_cdc_poll_interfaces(poll_flags);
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for this.  It looks like a regression from 4247921c4ebfced79aa7ccfbb835c9f60ad6f0fc .  So maybe a simpler fix, that just fixes what that commit broke, is:
```diff
--- a/ports/esp32/mphalport.c
+++ b/ports/esp32/mphalport.c
@@ -109,13 +109,13 @@ uintptr_t mp_hal_stdio_poll(uintptr_t poll_flags) {
     uintptr_t ret = 0;
     #if MICROPY_HW_ESP_USB_SERIAL_JTAG
     usb_serial_jtag_poll_rx();
+    #endif
     if ((poll_flags & MP_STREAM_POLL_RD) && ringbuf_peek(&stdin_ringbuf) != -1) {
         ret |= MP_STREAM_POLL_RD;
     }
     if (poll_flags & MP_STREAM_POLL_WR) {
         ret |= MP_STREAM_POLL_WR;
     }
-    #endif
     #if MICROPY_HW_USB_CDC
     ret |= mp_usbd_cdc_poll_interfaces(poll_flags);
     #endif
```

That's slightly semantically different to this PR, but much closer to what it was prior to the change above.

dpgeorge:
Oh, actually, maybe what's done in this PR is actually better, to test ringbuf after polling all devices.  Because 4247921c4ebfced79aa7ccfbb835c9f60ad6f0fc did add the CDC and dupterm polling code.

projectgus (approved):

andrewleech:
Yes I thought it was clearer / more robust to have this at the bottom of the function, in case other charges in those polling functions ended up relying on the ringbuffer.

dpgeorge:
I've tested this PR with ESP32_GENERIC firmware:
- on current master doing `mpremote mount .` then `import os` then `os.listdir()` is problematic (the import takes ages then the listdir has a timeout)
- also on current master running the new `mpremote` tests fails the mounting test
- with this PR applied, both the above tests work/pass

dpgeorge:
Thanks @andrewleech for the fix!

PR #16011, opened 2024-10-14
wiznet: Reset mDNS when interface is brought up.

greezybacon:
### Summary

The LwIP interface is removed in wiznet5k_deinit() which is called as part of the init sequence. Therefore, if using mDNS, then the interface will need to be re-added when bringing the interface up.

Additionally, this allows to set the hostname from Micropython code prior to bringing the interface up and mDNS responding to the (new) hostname. This allows the hostname to be configured and saved on the flash or be based on dynamic information such as the MAC or unique_id().

### Testing

This was tested with the W5100S_EVB_PICO board.

### DIFF
--- a/extmod/network_wiznet5k.c
+++ b/extmod/network_wiznet5k.c
@@ -58,6 +58,7 @@
 #include "shared/netutils/netutils.h"
 #include "lib/wiznet5k/Ethernet/wizchip_conf.h"
 #include "lib/wiznet5k/Ethernet/socket.h"
+#include "lwip/apps/mdns.h"
 #include "lwip/err.h"
 #include "lwip/dns.h"
 #include "lwip/dhcp.h"
@@ -201,6 +202,9 @@ static void wiznet5k_config_interrupt(bool enabled) {
 void wiznet5k_deinit(void) {
     for (struct netif *netif = netif_list; netif != NULL; netif = netif->next) {
         if (netif == &wiznet5k_obj.netif) {
+            #if LWIP_MDNS_RESPONDER
+            mdns_resp_remove_netif(&wiznet5k_obj.netif);
+            #endif
             netif_remove(netif);
             netif->flags = 0;
             break;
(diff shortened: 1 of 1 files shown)


tpwrules:
Does `mdns_resp_remove_netif` need to be called at any point? I think it should be called in `wiznet5k_deinit`, otherwise it will be initialized twice and leak things. I have had it set up like that for a bit and everything seems to be working.

tpwrules:
This is my patch on top of 1.23.0 for reference:
```diff
diff --git a/extmod/network_wiznet5k.c b/extmod/network_wiznet5k.c
index 1eaabe9eb..5dc1e93e8 100644
--- a/extmod/network_wiznet5k.c
+++ b/extmod/network_wiznet5k.c
@@ -61,6 +61,7 @@
 #include "lwip/err.h"
 #include "lwip/dns.h"
 #include "lwip/dhcp.h"
+#include "lwip/apps/mdns.h"
 #include "netif/etharp.h"
 
 #define TRACE_ETH_TX (0x0002)
@@ -198,6 +199,10 @@ static void wiznet5k_config_interrupt(bool enabled) {
 }
 
 void wiznet5k_deinit(void) {
+#if LWIP_MDNS_RESPONDER
+    mdns_resp_remove_netif(&wiznet5k_obj.netif);
+#endif
+
     for (struct netif *netif = netif_list; netif != NULL; netif = netif->next) {
         if (netif == &wiznet5k_obj.netif) {
             netif_remove(netif);
@@ -243,6 +248,11 @@ static void wiznet5k_init(void) {
 
     // register with network module
     mod_network_register_nic(&wiznet5k_obj);
+
+#if LWIP_MDNS_RESPONDER
+    mdns_resp_add_netif(&wiznet5k_obj.netif, mod_network_hostname_data, 60);
+    // no need to call anything further as LWIP_NETIF_EXT_STATUS_CALLBACK == 1
+#endif
 }
 
 static void wiznet5k_send_ethernet(wiznet5k_obj_t *self, size_t len, const uint8_t *buf) {
```

greezybacon:
@tpwrules thank you for your review. I really appreciate it. I tried tracing call through `lwip_remove()`, it looks like the callback for `remove_callback` should handle removing it also from mDNS, but I think the clarity and failsafe is much better.

I added the call to remove it as you suggested.

extmod/network_wiznet5k.c
```diff
@@ -199,6 +200,10 @@ static void wiznet5k_config_interrupt(bool enabled) {
 }
 
 void wiznet5k_deinit(void) {
+    #if LWIP_MDNS_RESPONDER
+    mdns_resp_remove_netif(&wiznet5k_obj.netif);
```
dpgeorge, line 204: I think this should be moved down to just before the `netif_remove(netif)` line.

greezybacon: Thanks for the review, @dpgeorge. I moved it as requested.

PR #16015, opened 2024-10-15
py/malloc: Fix failed memory allocations by more aggressively freeing buffers.

projectgus:
### Summary

This issue was originally reported on [Discord by ondrej1024](https://discord.com/channels/574275045187125269/1024252597839859734/1292796291247636553): on RPI_PICO_W a TLS connection via micropython-lib umqtt.simple would fail consistently on reconnect with `OSError(ENOMEM)`. A workaround was to explicitly close the old socket before connecting again. This doesn't seem like it should be necessary.

Was able to generalise into the new test case in this PR: `extmod/ssl_noleak`. This fails when trying to allocate sequential SSL socket objects (one at a time) on rp2 and also other ports.

Different fixes were needed, in separate commits in this PR. First approach was replaced with a new approach at suggestion of @dpgeorge, see inlinecomments below.

Changes:

* For rp2 and other mbedTLS bare metal ports, TLS nodes are "tracked" memory allocations associated with the SSL socket object. If an SSL socket object runs its finaliser during garbage collection then the "tracked" nodes can't be immediately freed (GC is locked) so they were left for the next GC. As a result, allocating a new SSL socket could fail even though enough memory was available (but not yet freed). Fix is to support `gc_free()` during a GC sweep by tracking gc collection separately from "heap is locked".
* This change led to a double-free assertion failure during gc_sweep_all (soft-reset), as the memory buffer might already be free. Fix for this was to split gc_sweep into two phases - one which runs finalisers and then one which frees memory. This works around other potential finaliser issues when memory has already been freed, too. Due to the sparse nature of FTB entries, this could be optimised to more than trade-off adding a second pass, with only a small code size increase.
* Similarly, GC mutex needed to be made recursive on ports which support `MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL`. Have added an optional generic recursive mutex interface.
* For esp32 something similar happens where the small SSL socket object in the MicroPython heap holds large buffers outside MicroPython in the ESP-IDF heap. Allocatinga new SSL socket object could fail to allocate new buffers as the ESP-IDF heap is too full, even if there are SSL socket objects eligible for garbage collection that would free up lots of ESP-IDF heap when garbage collected. Fix is to run a gc_collect() and retry if SSL socket setup fails.

### Testing

* New unit test `extmod/ssl_noleak` fails consistently on master (using RPI_NANO_W and ESP32_GENERIC_C3 boards). Probably also fails on other ports using `MICROPY_MBEDTLS_CONFIG_BARE_METAL`, unless they have huge amounts of RAM.
* With fixes in this PR, new test passes (as well as all other unit tests).
* Separate commit enables mbedTLS "bare metal" memory management in the coverage build, to cover these changes and increase the coverage of MicroPython-specific code (the default for unix port being libc calloc and free).

### Performance

Performance benchmark run on PYB_V11 with all cache init commented out:

```
diff of scores (higher is better)
N=168 M=100                ./pybv11_master.txt -> ./pybv11_pr.txt         diff      diff% (error%)
bm_chaos.py                    240.23 ->     242.18 :      +1.95 =  +0.812% (+/-0.00%)
bm_fannkuch.py                  48.62 ->      48.98 :      +0.36 =  +0.740% (+/-0.01%)
bm_fft.py                     1479.57 ->    1492.25 :     +12.68 =  +0.857% (+/-0.00%)
bm_float.py                   4082.97 ->    4080.69 :      -2.28 =  -0.056% (+/-0.00%)
bm_hexiom.py                    31.85 ->      31.80 :      -0.05 =  -0.157% (+/-0.01%)
bm_nqueens.py                 2840.25 ->    2866.36 :     +26.11 =  +0.919% (+/-0.01%)
bm_pidigits.py                 372.74 ->     376.82 :      +4.08 =  +1.095% (+/-0.36%)
bm_wordcount.py                 30.88 ->      30.88 :      +0.00 =  +0.000% (+/-0.01%)
core_import_mpy_multi.py       361.54 ->     363.66 :      +2.12 =  +0.586% (+/-0.01%)
core_import_mpy_single.py       57.34 ->      57.34 :      +0.00 =  +0.000% (+/-0.00%)
core_locals.py                  24.96 ->      24.96 :      +0.00 =  +0.000% (+/-0.00%)
core_qstr.py                    78.47 ->      78.46 :      -0.01 =  -0.013% (+/-0.00%)
core_str.py                     12.56 ->      12.56 :      +0.00 =  +0.000% (+/-0.00%)
core_yield_from.py             220.01 ->     220.00 :      -0.01 =  -0.005% (+/-0.01%)
misc_aes.py                    268.31 ->     268.31 :      +0.00 =  +0.000% (+/-0.01%)
misc_mandel.py                2069.87 ->    2090.56 :     +20.69 =  +1.000% (+/-0.01%)
misc_pystone.py               1366.78 ->    1366.45 :      -0.33 =  -0.024% (+/-0.01%)
misc_raytrace.py               246.10 ->     248.10 :      +2.00 =  +0.813% (+/-0.01%)
viper_call0.py                 332.95 ->     332.96 :      +0.01 =  +0.003% (+/-0.01%)
viper_call1a.py                324.59 ->     324.59 :      +0.00 =  +0.000% (+/-0.00%)
viper_call1b.py                252.76 ->     252.76 :      +0.00 =  +0.000% (+/-0.01%)
viper_call1c.py                254.68 ->     254.67 :      -0.01 =  -0.004% (+/-0.01%)
viper_call2a.py                319.64 ->     319.65 :      +0.01 =  +0.003% (+/-0.01%)
viper_call2b.py                222.30 ->     222.30 :      +0.00 =  +0.000% (+/-0.01%)
```

* For future performance gains we could look at adding a similar "fast path" optimisation to the ATB lookups in GC, and/or expanding the ATB and FTB table entries to word size (currently individual bytes).
* For bare metal, @dpgeorge pointed out that it's possible mbedTLS doesn't need "tracked node" allocations (i.e. if mbedTLS structures always contain pointers to the head of each buffer, then they could be tracked as normal GC objects). However the mbedTLS implementation has used "tracked node" allocations for as long as it's been in MicroPython, and seems like good insurance rather than auditing every mbedTLS allocation.
* For ESP32, it's tempting to try and extend the "GC and then retry" behaviour to all ESP-IDF heap allocations. However, this is much more fiddly and has potential thread safety issues for allocations made from non-Python threads. Most of the large ESP-IDF memory allocations happen as part of mbedTLS, at least.

### DIFF
--- a/ports/unix/mbedtls/mbedtls_config_port.h
+++ b/ports/unix/mbedtls/mbedtls_config_port.h
@@ -32,7 +32,18 @@
 // Enable mbedtls modules
 #define MBEDTLS_TIMING_C
 
+#if defined(MICROPY_UNIX_COVERAGE)
+// Test the "bare metal" memory management in the coverage build
+#define MICROPY_MBEDTLS_CONFIG_BARE_METAL (1)
+#endif
+
 // Include common mbedtls configuration.
 #include "extmod/mbedtls/mbedtls_config_common.h"
+
+#if defined(MICROPY_UNIX_COVERAGE)
+// See comment above, but fall back to the default platform entropy functions
+#undef MBEDTLS_ENTROPY_HARDWARE_ALT
+#undef MBEDTLS_NO_PLATFORM_ENTROPY
+#endif
 
 #endif /* MICROPY_INCLUDED_MBEDTLS_CONFIG_H */
--- a/ports/unix/mpthreadport.c
+++ b/ports/unix/mpthreadport.c
@@ -65,7 +65,7 @@ static pthread_key_t tls_key;
 // The mutex is used for any code in this port that needs to be thread safe.
 // Specifically for thread management, access to the linked list is one example.
 // But also, e.g. scheduler state.
-static pthread_mutex_t thread_mutex;
+static mp_thread_recursive_mutex_t thread_mutex;
 static mp_thread_t *thread;
 
 // this is used to synchronise the signal handler of the thread
@@ -78,11 +78,11 @@ static sem_t thread_signal_done;
 #endif
 
 void mp_thread_unix_begin_atomic_section(void) {
-    pthread_mutex_lock(&thread_mutex);
+    mp_thread_recursive_mutex_lock(&thread_mutex, true);
 }
 
 void mp_thread_unix_end_atomic_section(void) {
-    pthread_mutex_unlock(&thread_mutex);
+    mp_thread_recursive_mutex_unlock(&thread_mutex);
 }
 
 // this signal handler is used to scan the regs and stack of a thread
@@ -113,10 +113,7 @@ void mp_thread_init(void) {
 
     // Needs to be a recursive mutex to emulate the behavior of
     // BEGIN_ATOMIC_SECTION on bare metal.
-    pthread_mutexattr_t thread_mutex_attr;
-    pthread_mutexattr_init(&thread_mutex_attr);
-    pthread_mutexattr_settype(&thread_mutex_attr, PTHREAD_MUTEX_RECURSIVE);
-    pthread_mutex_init(&thread_mutex, &thread_mutex_attr);
+    mp_thread_recursive_mutex_init(&thread_mutex);
 
     // create first entry in linked list of all threads
     thread = malloc(sizeof(mp_thread_t));
@@ -320,6 +317,26 @@ void mp_thread_mutex_unlock(mp_thread_mutex_t *mutex) {
     pthread_mutex_unlock(mutex);
     // TODO check return value
 }
+
+#if MICROPY_PY_THREAD_RECURSIVE_MUTEX
+
+void mp_thread_recursive_mutex_init(mp_thread_recursive_mutex_t *mutex) {
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
+    pthread_mutex_init(mutex, &attr);
+    pthread_mutexattr_destroy(&attr);
+}
+
+int mp_thread_recursive_mutex_lock(mp_thread_recursive_mutex_t *mutex, int wait) {
+    return mp_thread_mutex_lock(mutex, wait);
+}
+
+void mp_thread_recursive_mutex_unlock(mp_thread_recursive_mutex_t *mutex) {
+    mp_thread_mutex_unlock(mutex);
+}
+
+#endif // MICROPY_PY_THREAD_RECURSIVE_MUTEX
 
 #endif // MICROPY_PY_THREAD
 
--- a/py/gc.c
+++ b/py/gc.c
@@ -113,9 +113,12 @@
 #endif
 
 #if MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL
-#define GC_ENTER() mp_thread_mutex_lock(&MP_STATE_MEM(gc_mutex), 1)
-#define GC_EXIT() mp_thread_mutex_unlock(&MP_STATE_MEM(gc_mutex))
+#define GC_MUTEX_INIT() mp_thread_recursive_mutex_init(&MP_STATE_MEM(gc_mutex))
+#define GC_ENTER() mp_thread_recursive_mutex_lock(&MP_STATE_MEM(gc_mutex), 1)
+#define GC_EXIT() mp_thread_recursive_mutex_unlock(&MP_STATE_MEM(gc_mutex))
 #else
+// Either no threading, or assume callers to gc_collect() hold the GIL
+#define GC_MUTEX_INIT()
 #define GC_ENTER()
 #define GC_EXIT()
 #endif
@@ -210,9 +213,7 @@ void gc_init(void *start, void *end) {
     MP_STATE_MEM(gc_alloc_amount) = 0;
     #endif
 
-    #if MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL
-    mp_thread_mutex_init(&MP_STATE_MEM(gc_mutex));
-    #endif
+    GC_MUTEX_INIT();
 }
 
 #if MICROPY_GC_SPLIT_HEAP
--- a/py/gc.c
+++ b/py/gc.c
@@ -477,29 +477,20 @@ static void gc_deal_with_stack_overflow(void) {
     }
 }
 
-static void gc_sweep(void) {
-    #if MICROPY_PY_GC_COLLECT_RETVAL
-    MP_STATE_MEM(gc_collected) = 0;
-    #endif
-    // free unmarked heads and their tails
-    int free_tail = 0;
-    #if MICROPY_GC_SPLIT_HEAP_AUTO
-    mp_state_mem_area_t *prev_area = NULL;
-    #endif
-    for (mp_state_mem_area_t *area = &MP_STATE_MEM(area); area != NULL; area = NEXT_AREA(area)) {
-        size_t end_block = area->gc_alloc_table_byte_len * BLOCKS_PER_ATB;
-        if (area->gc_last_used_block < end_block) {
-            end_block = area->gc_last_used_block + 1;
-        }
-
-        size_t last_used_block = 0;
-
-        for (size_t block = 0; block < end_block; block++) {
-            MICROPY_GC_HOOK_LOOP(block);
-            switch (ATB_GET_KIND(area, block)) {
-                case AT_HEAD:
-                    #if MICROPY_ENABLE_FINALISER
-                    if (FTB_GET(area, block)) {
+// Run finalisers for all to-be-freed blocks
+static void gc_sweep_run_finalisers(void) {
+    #if MICROPY_ENABLE_FINALISER
+    for (const mp_state_mem_area_t *area = &MP_STATE_MEM(area); area != NULL; area = NEXT_AREA(area)) {
+        assert(area->gc_last_used_block <= area->gc_alloc_table_byte_len * BLOCKS_PER_ATB);
+        // Small speed optimisation: skip over empty FTB blocks
+        size_t ftb_end = area->gc_last_used_block / BLOCKS_PER_FTB; // index is inclusive
+        for (size_t ftb_idx = 0; ftb_idx <= ftb_end; ftb_idx++) {
+            byte ftb = area->gc_finaliser_table_start[ftb_idx];
+            size_t block = ftb_idx * BLOCKS_PER_FTB;
+            while (ftb) {
+                MICROPY_GC_HOOK_LOOP(block);
+                if (ftb & 1) { // FTB_GET(area, block) shortcut
+                    if (ATB_GET_KIND(area, block) == AT_HEAD) {
                         mp_obj_base_t *obj = (mp_obj_base_t *)PTR_FROM_BLOCK(area, block);
                         if (obj->type != NULL) {
                             // if the object has a type then see if it has a __del__ method
@@ -519,7 +510,35 @@ static void gc_sweep(void) {
                         // clear finaliser flag
                         FTB_CLEAR(area, block);
                     }
-                    #endif
+                }
+                ftb >>= 1;
+                block++;
+            }
+        }
+    }
+    #endif // MICROPY_ENABLE_FINALISER
+}
+
+static void gc_sweep(void) {
+    #if MICROPY_PY_GC_COLLECT_RETVAL
+    MP_STATE_MEM(gc_collected) = 0;
+    #endif
+    // free unmarked heads and their tails
+    int free_tail = 0;
+    #if MICROPY_GC_SPLIT_HEAP_AUTO
+    mp_state_mem_area_t *prev_area = NULL;
+    #endif
+
+    gc_sweep_run_finalisers();
+
+    for (mp_state_mem_area_t *area = &MP_STATE_MEM(area); area != NULL; area = NEXT_AREA(area)) {
+        size_t last_used_block = 0;
+        assert(area->gc_last_used_block <= area->gc_alloc_table_byte_len * BLOCKS_PER_ATB);
+
+        for (size_t block = 0; block <= area->gc_last_used_block; block++) {
+            MICROPY_GC_HOOK_LOOP(block);
+            switch (ATB_GET_KIND(area, block)) {
+                case AT_HEAD:
                     free_tail = 1;
                     DEBUG_printf("gc_sweep(%p)\n", (void *)PTR_FROM_BLOCK(area, block));
                     #if MICROPY_PY_GC_COLLECT_RETVAL
--- a/py/gc.c
+++ b/py/gc.c
@@ -334,12 +334,12 @@ void gc_lock(void) {
     // - each thread has its own gc_lock_depth so there are no races between threads;
     // - a hard interrupt will only change gc_lock_depth during its execution, and
     //   upon return will restore the value of gc_lock_depth.
-    MP_STATE_THREAD(gc_lock_depth)++;
+    MP_STATE_THREAD(gc_lock_depth) += (1 << GC_LOCK_DEPTH_SHIFT);
 }
 
 void gc_unlock(void) {
     // This does not need to be atomic, See comment above in gc_lock.
-    MP_STATE_THREAD(gc_lock_depth)--;
+    MP_STATE_THREAD(gc_lock_depth) -= (1 << GC_LOCK_DEPTH_SHIFT);
 }
 
 bool gc_is_locked(void) {
@@ -581,13 +581,18 @@ static void gc_sweep(void) {
     }
 }
 
-void gc_collect_start(void) {
+static void gc_collect_start_common(void) {
     GC_ENTER();
-    MP_STATE_THREAD(gc_lock_depth)++;
+    assert((MP_STATE_THREAD(gc_lock_depth) & GC_COLLECT_FLAG) == 0);
+    MP_STATE_THREAD(gc_lock_depth) |= GC_COLLECT_FLAG;
+    MP_STATE_MEM(gc_stack_overflow) = 0;
+}
+
+void gc_collect_start(void) {
+    gc_collect_start_common();
     #if MICROPY_GC_ALLOC_THRESHOLD
     MP_STATE_MEM(gc_alloc_amount) = 0;
     #endif
-    MP_STATE_MEM(gc_stack_overflow) = 0;
 
     // Trace root pointers.  This relies on the root pointers being organised
     // correctly in the mp_state_ctx structure.  We scan nlr_top, dict_locals,
@@ -658,14 +663,12 @@ void gc_collect_end(void) {
     for (mp_state_mem_area_t *area = &MP_STATE_MEM(area); area != NULL; area = NEXT_AREA(area)) {
         area->gc_last_free_atb_index = 0;
     }
-    MP_STATE_THREAD(gc_lock_depth)--;
+    MP_STATE_THREAD(gc_lock_depth) &= ~GC_COLLECT_FLAG;
     GC_EXIT();
 }
 
 void gc_sweep_all(void) {
-    GC_ENTER();
-    MP_STATE_THREAD(gc_lock_depth)++;
-    MP_STATE_MEM(gc_stack_overflow) = 0;
+    gc_collect_start_common();
     gc_collect_end();
 }
 
@@ -902,10 +905,13 @@ found:
 // force the freeing of a piece of memory
 // TODO: freeing here does not call finaliser
 void gc_free(void *ptr) {
-    if (MP_STATE_THREAD(gc_lock_depth) > 0) {
-        // Cannot free while the GC is locked. However free is an optimisation
-        // to reclaim the memory immediately, this means it will now be left
-        // until the next collection.
+    // Cannot free while the GC is locked, unless we're only doing a gc sweep.
+    // However free is an optimisation to reclaim the memory immediately, this
+    // means it will now be left until the next collection.
+    //
+    // (We have the optimisation to free immediately from inside a gc sweep so
+    // that finalisers can free more memory when trying to avoid MemoryError.)
+    if (MP_STATE_THREAD(gc_lock_depth) & ~GC_COLLECT_FLAG) {
         return;
     }
 
@@ -930,7 +936,8 @@ void gc_free(void *ptr) {
     #endif
 
     size_t block = BLOCK_FROM_PTR(area, ptr);
-    assert(ATB_GET_KIND(area, block) == AT_HEAD);
+    assert(ATB_GET_KIND(area, block) == AT_HEAD
+        || (ATB_GET_KIND(area, block) == AT_MARK && (MP_STATE_THREAD(gc_lock_depth) & GC_COLLECT_FLAG)));
 
     #if MICROPY_ENABLE_FINALISER
     FTB_CLEAR(area, block);
--- a/py/gc.c
+++ b/py/gc.c
@@ -123,6 +123,18 @@
 #define GC_EXIT()
 #endif
 
+// Static functions for individual steps of the GC mark/sweep sequence
+static void gc_collect_start_common(void);
+static void *gc_get_ptr(void **ptrs, int i);
+#if MICROPY_GC_SPLIT_HEAP
+static void gc_mark_subtree(mp_state_mem_area_t *area, size_t block);
+#else
+static void gc_mark_subtree(size_t block);
+#endif
+static void gc_deal_with_stack_overflow(void);
+static void gc_sweep_run_finalisers(void);
+static void gc_sweep_free_blocks(void);
+
 // TODO waste less memory; currently requires that all entries in alloc_table have a corresponding block in pool
 static void gc_setup_area(mp_state_mem_area_t *area, void *start, void *end) {
     // calculate parameters for GC (T=total, A=alloc table, F=finaliser table, P=pool; all in bytes):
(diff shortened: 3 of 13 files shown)


ports/unix/coverage.c
```diff
@@ -150,6 +150,43 @@ static MP_DEFINE_CONST_OBJ_TYPE(
 static const mp_obj_str_t str_no_hash_obj = {{&mp_type_str}, 0, 10, (const byte *)"0123456789"};
 static const mp_obj_str_t bytes_no_hash_obj = {{&mp_type_bytes}, 0, 10, (const byte *)"0123456789"};
 
+#if MICROPY_ENABLE_FINALISER && MICROPY_TRACKED_ALLOC
+// Dummy class which contains a tracked allocation freed from its finaliser
```
projectgus, line 154: I realised after writing this custom coverage class & test that the other way to get coverage would be to build the `coverage` variant with `MICROPY_MBEDTLS_CONFIG_BARE_METAL`, so the new unit test covers the tracked allocation changes.

This might actually be a better approach, as I think this config doesn't disable anything apart from calling libc functions... What do you think, @dpgeorge?

py/malloc.c
```diff
@@ -248,12 +257,7 @@ void *m_tracked_calloc(size_t nmemb, size_t size) {
     size_t n = m_tracked_count_links(&nb);
     DEBUG_printf("m_tracked_calloc(%u, %u) -> (%u;%u) %p\n", (int)nmemb, (int)size, (int)n, (int)nb, node);
     #endif
-    if (MP_STATE_VM(m_tracked_head) != NULL) {
-        MP_STATE_VM(m_tracked_head)->prev = node;
-    }
-    node->prev = NULL;
-    node->next = MP_STATE_VM(m_tracked_head);
-    MP_STATE_VM(m_tracked_head) = node;
+    m_tracked_node_insert(&MP_STATE_VM(m_tracked_head), node);
```
projectgus, line 260: Do we need to worry about thread safety for any of these tracked node linked list operations? I think they're always called holding the GIL on GIL ports, but on non-GIL threaded ports (i.e. rp2) there's nothing stopping two threads from calling in here at the same time for unrelated objects, is there?

(Comment applies to the existing functions as much as the changes in this PR, although now we also potentially traverse a linked list during a GC pass.)

ports/unix/coverage.c
dpgeorge: If enabling `MICROPY_MBEDTLS_CONFIG_BARE_METAL` on the unix port (coverage variant) is possible, then yes that seems like a good approach.  In general less code in this `coverage.c` file is better (although it's unavoidable at times), and also it's better to emulate bare-metal for testing where possible, because then this unix-coverage build becomes a good proxy for coverage testing of bare-metal code.

projectgus: It all seems to pass here, have just pushed it to let the coverage analysis have the final word...

py/malloc.c
dpgeorge: Yes, I think you're right that it needs to be protected.  I guess everything in `mp_state_vm_t` does (`mp_state_mem_t` should already be protected by `gc_mutex`, and `mp_state_thread_t` is by definition per-thread so doesn't need protecting).

ports/unix/coverage.c
projectgus: Needed some tweaking, but seems to be working well. As a bonus it found a bug in my original implementation (the cleanup was running on soft reset `gc_sweep_all` and triggering double frees as the tracked nodes were already freed in this case!)

py/malloc.c
projectgus: Understood. Will look at fixing in a separate PR.

projectgus: Have submitted this draft:
https://github.com/micropython/micropython/pull/16071

Is stacked on top of this PR, currently.

py/malloc.c
```diff
@@ -287,16 +303,36 @@ void m_tracked_free(void *ptr_in) {
...
+
+#if MICROPY_ENABLE_FINALISER
+void m_tracked_free_pending(void) {
+    m_tracked_node_t *head;
+
+    if (MP_STATE_THREAD(gc_lock_depth) == 0) {
+        while ((head = MP_STATE_VM(m_tracked_pending_free_head))) {
+            MP_STATE_VM(m_tracked_pending_free_head) = head->next;
+            // During gc_sweep_all() these nodes are already free, so check first
+            if (gc_nbytes(head)) {
```
dpgeorge, line 329: I see why you needed to do this.  It feels a bit fragile but I think would be robust in practice.

You actually hit an issue which cropped up before in a different way, and may crop up again: `gc_sweep_all()` is too aggressive in the way it frees.  It should probably be changed to:
- first pass: run finalisers of all alive objects on the GC heap
- second pass: free all remaining memory on the GC heap

Bu, that's for another time.

dpgeorge:
This looks really clean @projectgus !  Thanks for the thorough implementation and testing.

That said, I can't help but think of the alternative where `gc_free()` could possibly be made to work when the GC is locked.  That would solve the bare-metal case without needing a list of pending tracked nodes to free, and also maybe improve other cases where `gc_free()` doesn't work if the heap is locked.

It could be tricky to implement though, eg it probably needs a recursive mutex for `GC_ENTER()`.  Or, could have a flag that's set when running a finaliser (eg `gc_finaliser_active = true`) and if this flag is set when `gc_free()` is called it runs only this bit of the function:
```c
    // free head and all of its tail blocks
    do {
        ATB_ANY_TO_FREE(area, block);
        block += 1;
    } while (ATB_GET_KIND(area, block) == AT_TAIL);
```
That would be more of a self-contained fix, contained to `gc.c`.

py/malloc.c
projectgus: That seems like a good idea, safer all around for finalisers.

projectgus:
>  `gc_free()` could possibly be made to work when the GC is locked

I think I started out with the same idea, and came to a sticking point that I couldn't seem to resolve. Unfortunately I no longer remember what that sticking point was, but I'll take a quick look and get back to you.

projectgus:
OK, maybe I didn't try this approach befote... I just implemented it and it works better than I expected!

Currently fails on ports with `MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL` as `GC_ENTER` isn't recursive yet, that needs updating and is a bit of work as different on each port.

Otherwise, seems like `gc_free()` can run unmodified in this case. It updates `gc_last_free_area` and `gc_last_free_atb_index` but these are both reset immediately after `gc_sweep` so changing them here seems pretty harmless.

### Running finalisers early

I had the same problem of potential for double-frees during `gc_sweep_all()`, so decided to try splitting out finalisers into an earlier pass as suggested above. Then I went back and optimised the pass a little bit for speed.

This seems to have turned out great!

Code size impact is small (this is the biggest code size impact in this PR, but the total size impact is *less* than the previous approach and we get the "run finalisers early on soft reset" behaviour as a bonus!)

| Iteration | qemu ARM text size | uni minimal text size |
| ------------ | -------------------------- | --------------- |
| baseline | 182052 | 175563 |
| split finaliser pass | 182072 | 175675 |
| optimised split finaliser pass | 182072 (!) | 175731 |

Garbage collecting large number of objects is also faster with this change, due to the rapid skipping of finaliser checks.  Ran [this test program](https://gist.github.com/projectgus/56af097dbe55898fda0bb18064b749c0) on ESP32_GENERIC+SPIRAM and ESP32_GENERIC_S3 (assuming ESP32+SPIRAM is close to worst case due to lots of slow RAM).

| Variant + Test | GC Time (master) | GC Time (this PR) | Delta |
| -------------------- | ----------- | ---- | ---- |
| ESP32_GENERIC+SPIRAM | | | |
| Baseline | 666us | 685us | +19us |
| objects | 477149us | 476484us | -665us |
| objects collected | 42260us | -8514us |
| ESP32_GENERIC_S3 | | | |
| Baseline | 281us | 274us | -7us |
| objects | 227321us | 226257us | -1064us |
| objects collected | 25271us | 18445us | -6826us |

I was going to make another test where there were thousands of objects with finalisers, but then I realised this doesn't really happen... all of the objects with finalisers are relatively limited system resources so at most there could be maybe a couple of dozen objects with finalisers live in the system. So optimising for a case where most blocks don't contain finalisers seems to be fine!

projectgus:
This should be ready for review, have updated the PR description with the new approach and the results of `run-perfbench.py` (quite good!)

dpgeorge:
I ran some performance benchmark tests with this PR, on a PYBV11.

With master vs this PR:
```
diff of scores (higher is better)
N=100 M=100                    master ->       PR           diff      diff% (error%)
bm_chaos.py                    353.68 ->     357.70 :      +4.02 =  +1.137% (+/-0.02%)
bm_fannkuch.py                  73.26 ->      74.12 :      +0.86 =  +1.174% (+/-0.00%)
bm_fft.py                     2322.68 ->    2349.30 :     +26.62 =  +1.146% (+/-0.00%)
bm_float.py                   5747.36 ->    5771.89 :     +24.53 =  +0.427% (+/-0.01%)
bm_hexiom.py                    46.13 ->      46.63 :      +0.50 =  +1.084% (+/-0.00%)
bm_nqueens.py                 4195.70 ->    4238.32 :     +42.62 =  +1.016% (+/-0.00%)
bm_pidigits.py                 696.50 ->     697.86 :      +1.36 =  +0.195% (+/-0.03%)
bm_wordcount.py                 46.12 ->      47.38 :      +1.26 =  +2.732% (+/-0.00%)
core_import_mpy_multi.py       601.76 ->     602.88 :      +1.12 =  +0.186% (+/-0.01%)
core_import_mpy_single.py       97.89 ->      98.07 :      +0.18 =  +0.184% (+/-0.01%)
core_locals.py                  40.77 ->      41.01 :      +0.24 =  +0.589% (+/-0.01%)
core_qstr.py                   202.79 ->     202.60 :      -0.19 =  -0.094% (+/-0.00%)
core_str.py                     27.13 ->      27.29 :      +0.16 =  +0.590% (+/-0.00%)
core_yield_from.py             351.04 ->     354.30 :      +3.26 =  +0.929% (+/-0.01%)
misc_aes.py                    400.40 ->     408.19 :      +7.79 =  +1.946% (+/-0.00%)
misc_mandel.py                3019.45 ->    3062.86 :     +43.41 =  +1.438% (+/-0.00%)
misc_pystone.py               2365.20 ->    2379.80 :     +14.60 =  +0.617% (+/-0.01%)
misc_raytrace.py               368.86 ->     371.13 :      +2.27 =  +0.615% (+/-0.01%)
viper_call0.py                 550.19 ->     502.52 :     -47.67 =  -8.664% (+/-0.00%)
viper_call1a.py                526.06 ->     526.04 :      -0.02 =  -0.004% (+/-0.01%)
viper_call1b.py                390.37 ->     381.52 :      -8.85 =  -2.267% (+/-0.00%)
viper_call1c.py                394.95 ->     385.90 :      -9.05 =  -2.291% (+/-0.00%)
viper_call2a.py                513.19 ->     513.18 :      -0.01 =  -0.002% (+/-0.02%)
viper_call2b.py                343.28 ->     336.42 :      -6.86 =  -1.998% (+/-0.00%)
```

The same thing but with icache disabled (`__HAL_FLASH_INSTRUCTION_CACHE_ENABLE()` commented out):
```
diff of scores (higher is better)
N=100 M=100           master-noicache -> PR-noicache        diff      diff% (error%)
bm_chaos.py                    282.44 ->     286.04 :      +3.60 =  +1.275% (+/-0.01%)
bm_fannkuch.py                  56.54 ->      57.57 :      +1.03 =  +1.822% (+/-0.01%)
bm_fft.py                     1729.54 ->    1763.51 :     +33.97 =  +1.964% (+/-0.00%)
bm_float.py                   4857.99 ->    4850.01 :      -7.98 =  -0.164% (+/-0.00%)
bm_hexiom.py                    36.78 ->      36.47 :      -0.31 =  -0.843% (+/-0.00%)
bm_nqueens.py                 3366.39 ->    3409.25 :     +42.86 =  +1.273% (+/-0.01%)
bm_pidigits.py                 465.57 ->     468.37 :      +2.80 =  +0.601% (+/-0.02%)
bm_wordcount.py                 35.43 ->      36.24 :      +0.81 =  +2.286% (+/-0.01%)
core_import_mpy_multi.py       425.80 ->     417.08 :      -8.72 =  -2.048% (+/-0.00%)
core_import_mpy_single.py       68.07 ->      65.41 :      -2.66 =  -3.908% (+/-0.01%)
core_locals.py                  30.46 ->      30.16 :      -0.30 =  -0.985% (+/-0.00%)
core_qstr.py                   112.58 ->     103.57 :      -9.01 =  -8.003% (+/-0.03%)
core_str.py                     16.50 ->      15.32 :      -1.18 =  -7.152% (+/-0.00%)
core_yield_from.py             268.53 ->     261.46 :      -7.07 =  -2.633% (+/-0.02%)
misc_aes.py                    306.40 ->     315.68 :      +9.28 =  +3.029% (+/-0.01%)
misc_mandel.py                2463.90 ->    2512.41 :     +48.51 =  +1.969% (+/-0.01%)
misc_pystone.py               1809.19 ->    1770.53 :     -38.66 =  -2.137% (+/-0.01%)
misc_raytrace.py               303.08 ->     305.20 :      +2.12 =  +0.699% (+/-0.01%)
viper_call0.py                 370.48 ->     377.14 :      +6.66 =  +1.798% (+/-0.02%)
viper_call1a.py                389.31 ->     393.89 :      +4.58 =  +1.176% (+/-0.00%)
viper_call1b.py                273.83 ->     287.90 :     +14.07 =  +5.138% (+/-0.00%)
viper_call1c.py                276.08 ->     290.38 :     +14.30 =  +5.180% (+/-0.01%)
viper_call2a.py                382.23 ->     386.64 :      +4.41 =  +1.154% (+/-0.00%)
viper_call2b.py                239.12 ->     252.41 :     +13.29 =  +5.558% (+/-0.01%)
```

This is not as convincing as the results obtained by @projectgus in https://github.com/micropython/micropython/pull/16015#issue-2587678758 ... but at least it doesn't look like a big regression in performance.

----------

I also tested on PYBD_SF2.  Before this PR it fails the `ssl_noleak.py` test.  With this PR it passes that test.

tests/extmod/ssl_noleak.py
```diff
@@ -0,0 +1,40 @@
+# Ensure that SSL sockets can be allocated sequentially
+# without running out of available memory.
+try:
+    import io
+    import ssl
```
dpgeorge, line 5: Since this test doesn't run under CPython (and has a .exp) is it worth just doing `import tls`?

Also, could use unittest for this test (although it is nice and simple without unittest).

projectgus: Have refactored to use unittest, good idea.

We have to import the `ssl` module here in order to use `wrap_socket()`.

projectgus:
Now up to the third version of this PR. The previous version couldn't account for the possibility that a hard IRQ triggers and locks the heap during a collection sweep - `gc_free()` should be denied in this case.

This version now splits `gc_lock_depth` into a flag for "in collection" and the heap lock depth. This allows `gc_free()` to tell the difference between "called from a finaliser with heap unlocked" and "heap locked due to hard IRQ".

Have re-run the unit tests mentioned in the PR description, but haven't re-run performance tests yet. Will do that soon and then mark as Ready for review again.

projectgus:
This is odd, these failing jobs are checking out a different commit. Will force push again.

tests/extmod/ssl_noleak.py
dpgeorge: OK, fair point that it needs `ssl.wrap_socket()`.

(Although, that wrapper function doesn't do anything more than create a context and call `ctx.wrap_socket`, so `test_without_context` is not testing anything more than `test_with_context` is.)

tests/extmod/ssl_noleak.py
```diff
@@ -0,0 +1,48 @@
...
+            s = TestSocket()
+            s = ssl.wrap_socket(s, do_handshake=False)
+
+    def test_with_context(self):
+        # Single SSLContext, multiple sockets
+        ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
+        for n in range(ITERS):
+            print(n)
+        s = TestSocket()
+        s = ctx.wrap_socket(s, do_handshake_on_connect=False)
```
dpgeorge, line 44: I think these two lines need to be indented :grimacing:

ports/esp32/main.c
```diff
@@ -96,6 +96,25 @@ time_t platform_mbedtls_time(time_t *timer) {
...
+    // (The ESP-IDF default is internal-only, partly for physical security to prevent
+    // possible information leakage from unencrypted PSRAM contents on the original
+    // ESP32 - no PSRAM encryption on that chip. MicroPython doesn't support flash
+    // encryption and is already storing the Python heap in PSRAM so this isn't a
+    // significant factor in overall physical security.)
+    void *p = calloc(c, len);
+    if (p == NULL) {
+        // A Python object (i.e. SSL socket) may be holding references
+        // to some ESP-IDF buffers which will be freed after collection
+        gc_collect();
```
dpgeorge, line 112: Are we really 100% certain that this `tls_calloc` function is only ever called from within MicroPython context (ie GIL is held), and so it's safe to call `gc_collect` here?

(I'm thinking about things like mDNS services that run in the background independently from the MP thread and may need to make mbedtls calls, which may then end up here.  Also maybe user C modules that start a C thread that uses TLS.)

ports/unix/mbedtls/mbedtls_config_port.h
```diff
@@ -32,7 +32,18 @@
 // Enable mbedtls modules
 #define MBEDTLS_TIMING_C
 
+#if MICROPY_UNIX_COVERAGE
```
dpgeorge, line 35: Existing uses of this feature macro are `#if defined(MICROPY_UNIX_COVERAGE)`.

I know it's probably equivalent, but this macro is not defined (even to 0) when coverage is not being built, and probably good to keep consistency with existing uses of it.

py/gc.c
```diff
@@ -477,29 +477,21 @@ static void gc_deal_with_stack_overflow(void) {
...
-    MP_STATE_MEM(gc_collected) = 0;
-    #endif
-    // free unmarked heads and their tails
-    int free_tail = 0;
-    #if MICROPY_GC_SPLIT_HEAP_AUTO
-    mp_state_mem_area_t *prev_area = NULL;
-    #endif
+// Run finalisers for all to-be-freed blocks first
+// (necessary during gc_sweep_all() to avoid accessing already-freed memory)
+static void sweep_run_finalisers(void) {
```
dpgeorge, line 482: I'd call this `gc_sweep_run_finalisers()`.

Also, I think it's necessary to run the finalisers fist for all GC sweeps, not just `gc_sweep_all`: if object X has a finaliser and is not reachable, then all of the things Y it points to are also not reachable.  If Y is before X in the GC heap then it'll be reclaimed before `X.__del__` is run.

This has probably never been an issue in practice because objects with finalisers don't access Y.

(And the corresponding commit comment would need to be updated to reflect this situation, if what I say above is true.)

tests/extmod/ssl_noleak.py
projectgus: Good point... I'll simplify the other case out.

ports/esp32/main.c
projectgus: mDNS is all plaintext, isn't it?

But I think the user C module case is a very strong one, will refactor to check the current thread is a Python thread.

py/gc.c
projectgus: You're right, it's currently easy to reproduce the failure with `gc_sweep_all()` but it could happen with other object graphs.

py/gc.c
```diff
@@ -477,29 +477,21 @@ static void gc_deal_with_stack_overflow(void) {
...
-        for (size_t block = 0; block < end_block; block++) {
-            MICROPY_GC_HOOK_LOOP(block);
-            switch (ATB_GET_KIND(area, block)) {
-                case AT_HEAD:
-                    #if MICROPY_ENABLE_FINALISER
-                    if (FTB_GET(area, block)) {
+        assert(area->gc_last_used_block <= area->gc_alloc_table_byte_len * BLOCKS_PER_ATB);
+        // Small speed optimisation: skip over empty FTB blocks
+        size_t ftb_end = (area->gc_last_used_block + BLOCKS_PER_FTB) / BLOCKS_PER_FTB;
+        for (size_t ftb_idx = 0; ftb_idx < ftb_end; ftb_idx++) {
```
dpgeorge, line 488: This arithmetic is a little tricky, considering `gc_last_used_block` is inclusive.

I think `(area->gc_last_used_block + BLOCKS_PER_FTB) / BLOCKS_PER_FTB` is equivalent to `area->gc_last_used_block / BLOCKS_PER_FTB + 1`, so the above two lines are equivalent to:
```c
        size_t ftb_end = area->gc_last_used_block / BLOCKS_PER_FTB;
        for (size_t ftb_idx = 0; ftb_idx <= ftb_end; ftb_idx++) {
```
That might be a little clearer when reading the code (although the compiler is probably already making this optimisation).

py/gc.c
```diff
@@ -113,8 +113,8 @@
 #endif
 
 #if MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL
-#define GC_ENTER() mp_thread_mutex_lock(&MP_STATE_MEM(gc_mutex), 1)
-#define GC_EXIT() mp_thread_mutex_unlock(&MP_STATE_MEM(gc_mutex))
+#define GC_ENTER() mp_thread_recursive_mutex_lock(&MP_STATE_MEM(gc_mutex), 1)
```
dpgeorge, line 116: I suggest adding a `GC_MUTEX_INIT` macro here, and using it below in `gc_init`, so that the condition `MICROPY_PY_THREAD && !MICROPY_PY_THREAD_GIL` doesn't need to be repeated below (otherwise the init and use of the mutex may get out of sync wrt which config options it's enabled for).

ports/unix/mpthreadport.c
```diff
@@ -321,6 +321,26 @@ void mp_thread_mutex_unlock(mp_thread_mutex_t *mutex) {
     // TODO check return value
 }
 
+#if MICROPY_PY_THREAD_RECURSIVE_MUTEX
+
+void mp_thread_recursive_mutex_init(mp_thread_recursive_mutex_t *mutex) {
```
dpgeorge, line 326: Looks like the initialisation of `thread_mutex` at the top of `mp_thread_init` can use this new function.

ports/esp32/main.c
projectgus: Have rewritten to handle only the special case of `mbedtls_ssl_setup()` as this is the function most likely to fail to allocate.

I think there is a possible generic version of this (maybe applicable to all C-level mallocs), but need to:

* Make GIL mutex recursive on esp32 so we can take it while calling gc_collect().
* Handle non-Python threads (either by skipping GC on these, or by doing what the BLE stack does and making a temporary fake Python thread context.)

But this way is much smaller scoped change, covers the most obvious failure pattern.

ports/unix/mbedtls/mbedtls_config_port.h
projectgus: Have changed.

Would you be open to a PR that cleaned up the few `MICROPY_` boolean flags that are treated as ("defined means set" and changed them to match the pattern more commonly used in MicroPython of "1 means set, 0 means unset")?

Quick grep suggests there are not many (pattern is mostly used when the macro holds a value like a pin number).

py/gc.c
projectgus: Have changed! Oddly, this did change something in the code generation on -Os (got slightly bigger) :shrug:

tests/extmod/ssl_noleak.py
projectgus: Have rewritten to not use `ssl.wrap_socket()` and changed the import to `tls`.

tests/extmod/ssl_noleak.py
projectgus: Eek! Have fixed, and double checked both tests fail with ENOMEM on master and pass in this branch.

ports/esp32/main.c
dpgeorge: > Have rewritten to handle only the special case of `mbedtls_ssl_setup()` as this is the function most likely to fail to allocate.

That is a lot cleaner and more generic (not tied to esp32).



> * Make GIL mutex recursive on esp32 so we can take it while calling gc_collect().

I don't think a recursive GIL is necessary.  See `extmod/modbluetooth.c` and grep for `MICROPY_PY_BLUETOOTH_USE_SYNC_EVENTS_WITH_INTERLOCK`: it only obtains the GIL if it's not already within Python context.

ports/unix/mbedtls/mbedtls_config_port.h
dpgeorge: > Would you be open to a PR that cleaned up the few `MICROPY_` boolean flags that are treated as ("defined means set" and changed them to match the pattern more commonly used in MicroPython of "1 means set, 0 means unset")?

Yes!

py/gc.c
dpgeorge: Looking at the latest edits on commit https://github.com/micropython/micropython/pull/16015#issuecomment-2412939085, it seems that 3 bare-metal ports got smaller by 8 bytes, which is good, and I assume is due to this change.

py/gc.c
projectgus: Good idea, have made the change

ports/unix/mpthreadport.c
projectgus: Oh yeah, I hadn't noticed that mutex! Have refactored.

ports/esp32/main.c
projectgus: > I don't think a recursive GIL is necessary.

I was thinking of protecting against the case where a user C module releases the GIL on a Python thread. But maybe this isn't something a user C module is able to do?

dpgeorge: > I was thinking of protecting against the case where a user C module releases the GIL on a Python thread

Yes, a user C module can definitely do that.

So I think what you have now is a good solution.

py/gc.c
```diff
@@ -519,7 +511,35 @@ static void gc_sweep(void) {
...
+        }
+    }
+    #endif // MICROPY_ENABLE_FINALISER
+}
+
+static void gc_sweep(void) {
+    #if MICROPY_PY_GC_COLLECT_RETVAL
+    MP_STATE_MEM(gc_collected) = 0;
+    #endif
+    // free unmarked heads and their tails
```
dpgeorge, line 527: This comment belongs down just after the call to `gc_sweep_run_finalisers()`.  Also probably worth moving the related variables `free_tail` and `prev_area` down as well, to the top of that for loop where they are used.

Actually, on second thought, might be worth putting the call to `gc_sweep_run_finalisers()` in `gc_collect_end()`.  Then this function (`gc_sweep`) is just doing the sweep.  Eg:
```c
void gc_collect_end(void) {               
    gc_deal_with_stack_overflow();   
    gc_sweep_run_finalisers();                   
    gc_sweep();                                         
    ...
}
```
Could even rename `gc_sweep` to `gc_sweep_reclaim_blocks` or something like that.

projectgus: Good idea. I've gone through and made those changes, and then went a step further and reordered the static functions so they read top-to-bottom in roughly the order of the GC flow, instead of the order being determined by the declaration order.

Have put this in its own commit, as it creates even more churn in the diff, but it's probably better for long-term maintainability.

dpgeorge (approved):
Looks good now, this has come a long way since its first version!

Please can you just modify the commit subject line for the two "gc: ..." commits to "py/gc: ..." for consistency (I still need to revisit #15823 for that).

projectgus:
> Please can you just modify the commit subject line for the two "gc: ..." commits to "py/gc: ..." for consistency (I still need to revisit #15823 for that).

Done!

PR #16017, opened 2024-10-15
Draft: tinyusb cdc class in python / dupterm

andrewleech:
### Summary
In the stm port the built-in USB CDC / virtual com port is exposed as a class in python. It's also connected to repl by being registered in dupterm by default.

The benefits of this are that at runtime the user can  de-register it from dupterm to disable repl in that port and then also use it as a custom stream if desired. 

This PR replicates this behaviour on other ports which use TinyUSB.

### Testing

This has been tested so far on: 
* Renesas

This does come at a cost to code size. 
Currently it's written such that it dupterm is not enabled the previous behaviour is used, with the class no longer created. 

A define `MICROPY_HW_UART_REPL` is commonly used to specify whether a UART should be connected to repl, I'm considering whether a matching `MICROPY_HW_USBD_CDC_REPL` define should be added to specify whether the CDC should be registered in dupterm automatically or not. 

Then there's also a question about whether a separate define should be added to specify if the class should be created anyway in case people want access to that for custom data.

### DIFF
--- a/extmod/modmachine.c
+++ b/extmod/modmachine.c
@@ -32,6 +32,10 @@
 #include "extmod/modmachine.h"
 #include "shared/runtime/pyexec.h"
 
+#if MICROPY_HW_USB_CDC
+#include "shared/tinyusb/mp_usbd_cdc.h"
+#endif
+
 #if MICROPY_PY_MACHINE_DHT_READINTO
 #include "drivers/dht/dht.h"
 #endif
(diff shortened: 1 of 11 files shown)


andrewleech:
I started adding `send` and `recv` to match the pyb module, particularly to use these with the test_serial.py script and see how much speed improvement you get vs stdin in particular. While this works, it's really not neccesary when there's already the normal stream operations so I will remove them again - users should be able to do everything needed with the stream api and having these it arguably just confusing and costing extra flash.

PR #16023, opened 2024-10-16
py/malloc: Add m_tracked_realloc().

andrewleech:
### Summary
This PR provides a `m_tracked_realloc()` function to compliment the existing `m_tracked_calloc()` and `m_tracked_free()` functions. 

I'm working on improvements to the LVGL micropython bindings (as user c module) and these functions look safer to provide to that library for dynamic memory support.

I may be wrong here in my assumptions about how the various forms of alloc/free work, but if these are #defined in to be used in a large user c module that expects "standard" malloc/callloc/free function semantics would this tracked version be safer in the sense of ensuring the gc doesn't collect things that might still be in use deep in said C library?

### Testing

The main testing done so far is via the updated coverage test included here. 
I'm intending to test this further in user C modules like LVGL.

There's already a non-tracked `m_realloc()` function however when `MICROPY_MALLOC_USES_ALLOCATED_SIZE` is enabled this function requires the previous size to also be provided so no longer matches C standard lib signature. This setting is on by default in the unix port.

### DIFF
--- a/ports/unix/coverage.c
+++ b/ports/unix/coverage.c
@@ -343,6 +343,13 @@ static mp_obj_t extra_coverage(void) {
             gc_collect();
         }
 
+        // resize one of the blocks
+        void *before = ptrs[1];
+        ptrs[1] = FLIP_POINTER(m_tracked_realloc(FLIP_POINTER(ptrs[1]), 2 * NUM_BYTES));
+        void *after = ptrs[1];
+        bool location_changed = before != after;
+        mp_printf(&mp_plat_print, "%d\n", location_changed);
+
         // check the memory blocks have the correct content
         for (size_t i = 0; i < NUM_PTRS; ++i) {
             bool correct_contents = true;
@@ -361,6 +368,65 @@ static mp_obj_t extra_coverage(void) {
         }
 
         mp_printf(&mp_plat_print, "m_tracked_head = %p\n", MP_STATE_VM(m_tracked_head));
+
+        // Test realloc with black-box behavioral testing
+        mp_printf(&mp_plat_print, "# tracked realloc\n");
+
+        // Test 1: Basic realloc with data preservation
+        uint8_t *test_ptr = m_tracked_calloc(1, 32);
+        for (int i = 0; i < 32; i++) {
+            test_ptr[i] = i;
+        }
+
+        test_ptr = m_tracked_realloc(test_ptr, 64);  // Grow
+        bool data_preserved = (test_ptr[0] == 0 && test_ptr[31] == 31);
+        mp_printf(&mp_plat_print, "grow preserves data: %d\n", data_preserved);
+
+        test_ptr = m_tracked_realloc(test_ptr, 16);  // Shrink
+        bool shrink_ok = (test_ptr[0] == 0 && test_ptr[15] == 15);
+        mp_printf(&mp_plat_print, "shrink preserves data: %d\n", shrink_ok);
+
+        m_tracked_free(test_ptr);
+
+        // Test 2: Multiple allocations + reallocs + GC stability
+        uint8_t *realloc_ptrs[5];
+        for (int i = 0; i < 5; i++) {
+            realloc_ptrs[i] = m_tracked_calloc(1, 32);
+            realloc_ptrs[i][0] = 'A' + i;  // Mark each
+        }
+
+        // Realloc some in different positions
+        realloc_ptrs[0] = m_tracked_realloc(realloc_ptrs[0], 64);  // First allocated (tail of list)
+        realloc_ptrs[2] = m_tracked_realloc(realloc_ptrs[2], 64);  // Middle
+        realloc_ptrs[4] = m_tracked_realloc(realloc_ptrs[4], 64);  // Last allocated (head of list)
+
+        // Run GC - if list corrupted, this might crash/fail
+        gc_collect();
+
+        // Verify markers intact
+        bool markers_ok = true;
+        for (int i = 0; i < 5; i++) {
+            if (realloc_ptrs[i][0] != 'A' + i) {
+                markers_ok = false;
+                break;
+            }
+        }
+        mp_printf(&mp_plat_print, "realloc gc stable: %d\n", markers_ok);
+
+        // Cleanup
+        for (int i = 0; i < 5; i++) {
+            m_tracked_free(realloc_ptrs[i]);
+        }
+
+        // Test 3: Edge cases
+        uint8_t *null_alloc = m_tracked_realloc(NULL, 32);
+        null_alloc[0] = 'X';
+        mp_printf(&mp_plat_print, "realloc(NULL) ok: %d\n", null_alloc[0] == 'X');
+
+        void *free_result = m_tracked_realloc(null_alloc, 0);
+        mp_printf(&mp_plat_print, "realloc(ptr, 0) returns NULL: %d\n", free_result == NULL);
+
+        mp_printf(&mp_plat_print, "m_tracked_head after cleanup: %p\n", MP_STATE_VM(m_tracked_head));
     }
 
     // vstr
--- a/py/malloc.c
+++ b/py/malloc.c
@@ -245,6 +245,15 @@ typedef struct _m_tracked_node_t {
     uint8_t data[];
 } m_tracked_node_t;
 
+// Helper to get data size of a tracked node, abstracting MICROPY_TRACKED_ALLOC_STORE_SIZE.
+static inline size_t get_tracked_node_size(m_tracked_node_t *node) {
+    #if MICROPY_TRACKED_ALLOC_STORE_SIZE
+    return node->size;
+    #else
+    return gc_nbytes(node) - sizeof(m_tracked_node_t);
+    #endif
+}
+
 #if MICROPY_DEBUG_VERBOSE
 static size_t m_tracked_count_links(size_t *nb) {
     m_tracked_node_lock();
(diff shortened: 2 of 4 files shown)


py/malloc.c
```diff
@@ -263,6 +263,47 @@ void *m_tracked_calloc(size_t nmemb, size_t size) {
...
+    if (n_bytes == 0) {
+        m_tracked_free(ptr_in);
+        return NULL;
+    }
+    m_tracked_node_t *node = (m_tracked_node_t *)((uint8_t *)ptr_in - sizeof(m_tracked_node_t));
+    size_t prev_bytes;
+    #if MICROPY_TRACKED_ALLOC_STORE_SIZE
+    prev_bytes = node->size;
+    #else
+    prev_bytes = gc_nbytes(node);
```
andrewleech, line 281: Hi @dpgeorge I copied this logic from `free` below but feel there's a discrepancy here?  When `node->size` is set in alloc above it's set to `nmemb * size` so just the requested allocation size. 
I would have thought the `gc_nbytes(node)` version would return the entire allocated size, ie `sizeof(m_tracked_node_t) + nmemb * size` ?
Is this true, and if so does it not really matter as its only used for metrics reporting? I just found it a little confusing so would be happy to add some clarifying comments if needed, or update the impl to be consistent in both cases.

py/malloc.c
```diff
@@ -263,6 +263,49 @@ void *m_tracked_calloc(size_t nmemb, size_t size) {
...
+    #endif
+    #endif
+    node = m_realloc(node,
+        #if MICROPY_MALLOC_USES_ALLOCATED_SIZE
+        sizeof(m_tracked_node_t) + prev_bytes,
+        #endif
+        sizeof(m_tracked_node_t) + n_bytes
+        );
+    if (node == NULL) {
+        node->prev->next = node->next;
```
projectgus, line 297: This line dereferences a NULL pointer. You might need to use a second variable for the result, to disambiguate.

(Can also cover these lines in the coverage test by trying to `m_tracked_realloc()` to an unreasonably large size.)

py/malloc.c
```diff
@@ -263,6 +263,49 @@ void *m_tracked_calloc(size_t nmemb, size_t size) {
...
+        node->next->prev = node->prev;
+        return NULL;
+    }
+    #if MICROPY_TRACKED_ALLOC_STORE_SIZE
+    node->size = n_bytes;
+    #endif
+    node->prev->next = node;
+    node->next->prev = node;
```
projectgus, line 305: I think `node->next` and `node->prev` can be NULL at the head and tail of the list, respectively.

py/malloc.c
projectgus: FWIW I looked at this code last week, and my understanding is same for yours - this is only ever used for stats, either `MICROPY_TRACKED_ALLOC_STORE_SIZE` for debugging tracked alloc or complete memory stats when  `MICROPY_MALLOC_USES_ALLOCATED_SIZE` is set.

That said, maybe could we remove one of the config permutations by requiring `MICROPY_TRACKED_ALLOC_STORE_SIZE` to be set if `MICROPY_MALLOC_USES_ALLOCATED_SIZE` is set?

I also toyed with the idea of adding a `inline static size_t get_size(node)` function here to simplify some of these copy-paste ifdefs. I think I decided it was marginal for my PR, but if we're adding one more here it might be worth it!

projectgus (commented):
This looks potentially useful, for sure! Left a couple of comments inline.

I was worried about the code size impact but I guess this only gets compiled if C code links to `realloc`, yeah? So seems pretty worthwhile given that tradeoff.

andrewleech:
Thanks for the review @projectgus. I've addressed all the feedback.

The NULL prev/next dereference is fixed with proper NULL checks matching the pattern in `m_tracked_free()`. The code now safely handles realloc at the head (prev==NULL), tail (next==NULL), and middle of the list. Also added thread safety locks around the linked list updates that were missing in the original implementation.

Added a `get_tracked_node_size()` helper as you suggested to reduce the conditional code duplication across multiple functions. This abstracts away the `MICROPY_TRACKED_ALLOC_STORE_SIZE` logic.

Test coverage includes head/tail/middle list positions, shrinking realloc, single-node list edge case, and backward list traversal to verify prev pointers are maintained correctly. All tests pass.

tests/run-tests.py
```diff
@@ -1382,6 +1382,10 @@ def main():
   run-tests.py -i async - exclude all, then include tests containing "async" anywhere
   run-tests.py -e '/big.+int' - include all, then exclude by regex
   run-tests.py -e async -i async_foo - include all, exclude async, yet still include async_foo
+
+For tests that use cpython, unix micropython and mpy-cross, the default executables
+vcan be overridden with the environment variables MICROPY_CPYTHON3, MICROPY_MICROPYTHON and
```
projectgus:
```suggestion
can be overridden with the environment variables MICROPY_CPYTHON3, MICROPY_MICROPYTHON and
```

(Although not sure if this addition is better in another PR?)

ports/unix/coverage.c
```diff
@@ -20,6 +20,16 @@
 #include "py/binary.h"
 #include "py/bc.h"
 
+// Forward declaration for tracked allocation testing
```
projectgus: Re-declaring m_tracked_node_t here seems risky (and it already has potential ABI mismatch as the "real" `size` member is gated on `MICROPY_TRACKED_ALLOC_STORE_SIZE` not `!MICROPY_ENABLE_GC`.

Is it possible to rewrite the tests here so that they test the behaviour that we want to see from `m_tracked_realloc()` as a black box, rather than testing the internals of the implementation?

Perhaps something like: doing some tracked allocations, realloc them, run a GC pass, check everything is still valid, free them, allocate some more?

If it's definitely necessary to test the internals by walking the linked list, maybe we can create a `py/tracked_alloc.h` and put all the relevant declarations there, and then include that header from `py/misc.h` for compatibility?

Big preference for testing this as a black box here, though, if possible.

py/malloc.c
```diff
@@ -292,18 +301,57 @@ void *m_tracked_calloc(size_t nmemb, size_t size) {
...
+        #if MICROPY_MALLOC_USES_ALLOCATED_SIZE
+        sizeof(m_tracked_node_t) + prev_bytes,
+        #endif
+        sizeof(m_tracked_node_t) + n_bytes
+        );
+    // m_realloc raises on failure (never returns NULL), so no error handling needed.
+    #if MICROPY_TRACKED_ALLOC_STORE_SIZE
+    node->size = n_bytes;
+    #endif
+    m_tracked_node_lock();
```
projectgus: I think there may be a race between the `m_realloc()` above and this point. An interrupt (or CPU1 on rp2) could run and traverse the old version of the linked list, including the old buffer address which may have been freed if `m_realloc()` moved the buffer. Is it safe to move the call to `m_tracked_node_lock()` before `m_realloc()`? The gc_mutex is recursive now, I believe.

Otherwise, if `m_tracked_realloc()`  isn't going to be in a hot code path or used with very large buffers then a simpler but slower option would be to `m_tracked_calloc()` a whole new buffer each time, copy the contents of the old buffer into it, and then `m_tracked_free()` the old buffer and return a pointer to the new one. This is a lot less efficient but it's simpler as it's built using only the existing tracked_alloc() API, there wouldn't be any need to fixup the linked list so much of  the extra coverage could probably be avoided as well...?

projectgus (commented):
Thanks for keeping this ticking along, @andrewleech! How are your experiments with LVGL going?

(I think you're right that it's safer to use this API from LVGL rather than to rely on the GC pointer tracking. An upside is that if we eventually implement [selective collect ala CircuitPython](https://github.com/orgs/micropython/discussions/17142) then all the LVGL buffers would be automatically marked as "no pointers here" and skipped by the GC, which is probably very beneficial for GC times!)

andrewleech:
Thanks for the detailed feedback, I've updated the implementation based on your suggestions and rebased onto latest master.

Switched to the alloc+copy+free approach which eliminates the race condition entirely - each operation handles its own locking so there's no window where the list is in an inconsistent state. Also added the NULL check after allocation so if m_tracked_calloc fails the original pointer stays intact per standard realloc semantics.

The tests are now black-box, removed the m_tracked_node_t redeclaration and instead test data preservation during grow/shrink, GC stability after multiple reallocs, and the edge cases like realloc(NULL, n) and realloc(ptr, 0). The FLIP_POINTER test now always shows location_changed=1 since alloc+copy+free always moves the pointer, which is expected behaviour for this implementation.

Removed the unrelated run-tests.py documentation changes as you suggested.

The alloc+copy+free approach trades some efficiency for simplicity - always allocates a new buffer rather than potentially growing in-place, which increases peak memory pressure briefly while both buffers exist. The correctness and simplicity gains are worth it though.

The get_tracked_node_size() call happens outside the lock but that's safe since node->size is write-once and gc_nbytes reads stable metadata for live allocations, same pattern as m_tracked_free uses.

Known gap is the allocation failure path isn't tested since that would need deterministic heap exhaustion which is tricky in the coverage tests.

Another thing I noticed was the m_tracked_count_links byte counting has a pre-existing inconsistency between the STORE_SIZE and gc_nbytes paths but that's unrelated to this PR, can address it too though if desired.

py/malloc.c
```diff
@@ -292,18 +301,47 @@ void *m_tracked_calloc(size_t nmemb, size_t size) {
...
+    if (ptr_in == NULL) {
+        return m_tracked_calloc(1, n_bytes);
+    }
+
+    // Handle pure free
+    if (n_bytes == 0) {
+        m_tracked_free(ptr_in);
+        return NULL;
+    }
+
```
projectgus, line 317: Suggest a short comment here about the approach, i.e.


```suggestion
// To keep the implementation simple, we always allocate a new buffer and copy the old data into it.
// This could be optimised if faster performance or lower worst-case memory usage is required.
```

py/malloc.c
```diff
@@ -292,18 +301,47 @@ void *m_tracked_calloc(size_t nmemb, size_t size) {
...
+    // Allocate new buffer
+    void *new_ptr = m_tracked_calloc(1, n_bytes);
+    if (new_ptr == NULL) {
+        // Allocation failed, return NULL but leave original pointer intact
+        return NULL;
+    }
+
+    // Copy data (minimum of old and new size)
+    size_t copy_size = (old_size < n_bytes) ? old_size : n_bytes;
```
projectgus: Nitpick, but can use macro here:

```suggestion
    size_t copy_size = MIN(old_size, n_bytes);
```

projectgus (approved):
Thanks @andrewleech. It's certainly much simpler code now! I hope I didn't bulldoze you into this approach, and it still meets your needs for this API.

andrewleech:
>  I hope I didn't bulldoze you into this approach, and it still meets your needs for this API.

No pressure as all, I really appreciate the suggestion to simplify things and for your time reviewing it!

dpgeorge:
Thanks for this!

PR #16026, opened 2024-10-16
rp2/mpconfigport: Leave callable pointers alone on RV32.

agatti:
### Summary

The port configuration file tagged callable pointers' LSB on both Arm and RISC-V variants.  This is needed on Arm due to Thumb/Thumb2 code addresses having their LSB set, but on RISC-V this is not required.

### Testing

My RPI2 is still being shipped, but I've rebuilt the port with both board variants making sure the define in question is enabled only for the Arm port.  All the other Arm-based ports have that defined, as it is needed to properly handle Thumb/Thumb2 code, but non-Arm ports in fact do not have that define.

### DIFF
--- a/ports/rp2/mpconfigport.h
+++ b/ports/rp2/mpconfigport.h
@@ -275,7 +275,9 @@ extern const struct _mp_obj_type_t mod_network_nic_type_wiznet5k;
 #define MICROPY_HW_BOOTSEL_DELAY_US 8
 #endif
 
+#if PICO_ARM
 #define MICROPY_MAKE_POINTER_CALLABLE(p) ((void *)((mp_uint_t)(p) | 1))
+#endif
 
 #define MP_SSIZE_MAX (0x7fffffff)
 typedef intptr_t mp_int_t; // must be pointer size


agatti:
Would it actually make sense to set `PICO_ARM` even if `PICO_RP2040` is defined?  That way maybe some checks that look for either the Pico1 or the Pico2 in Arm mode would be simplified to just one single condition rather than two.

dpgeorge:
> Would it actually make sense to set `PICO_ARM` even if `PICO_RP2040` is defined?

I think that's already done, see `lib/pico-sdk/src/rp2040.cmake`.  Also, we rely on that in `mpconfigport.h` for enabling `MICROPY_EMIT_THUMB` etc.

dpgeorge:
The existing code works OK on RISC-V, but I guess that's because the CPU ignores the LSB of the program counter?

Anyway, this fix here looks good, thanks!

agatti:
> The existing code works OK on RISC-V, but I guess that's because the CPU ignores the LSB of the program counter?

Branch/Jump instructions don't encode either the last or the last two bits (compressed and regular instruction kinds, respectively), so that won't create issues.  However if that address with the LSB set is used as a base for something else and the offset applied to it is odd (by bug or by design) then it may or may not point to the wrong location.

Super-rare corner case, but better safe than sorry :)

ports/rp2/mpconfigport.h
```diff
@@ -275,7 +275,9 @@ extern const struct _mp_obj_type_t mod_network_nic_type_wiznet5k;
 #define MICROPY_HW_BOOTSEL_DELAY_US 8
 #endif
 
+#if PICO_RP2040 || PICO_ARM
```
dpgeorge, line 278: Can this just be `PICO_ARM`?

agatti: Ironically my first version of this PR only used `PICO_ARM`, but I wasn't sure whether that would cover the Pico1, hence the dual condition.

However, I see that it's set in `lib/pico-sdk/src/rp2040.cmake` too - I'll update the PR!

agatti: Done, thanks for the review!

dpgeorge (approved):

ports/rp2/mpconfigport.h
dpgeorge: Thanks for updating!

PR #16031, opened 2024-10-17
ports: Update board.json files for vendor/product consistency.

mattytrentini:
The `vendor` and `product` fields in the `board.json` files were somewhat inconsistent. In particular, I wanted to remove any duplication of the vendor name in the product field so that `f"{vendor} {product}"` reads well.

In addition to that I checked - and updated - most of the url's for `board.json`'s that I modified. I also tried to match case and spacing used by the manufacturers for the vendor and product names. 

In short, it was a general tidy-up of a lot of the `board.json` files. 

(I was going to add `features` to boards that didn't specify them, but that will have to come in a later PR...)

### DIFF
--- a/ports/esp32/boards/ARDUINO_NANO_ESP32/board.json
+++ b/ports/esp32/boards/ARDUINO_NANO_ESP32/board.json
@@ -17,8 +17,8 @@
         "ABX00092_01.iso_1000x750.jpg"
     ],
     "mcu": "esp32s3",
-    "product": "Arduino Nano ESP32",
+    "product": "Nano ESP32",
     "thumbnail": "",
-    "url": "https://store.arduino.cc/products/arduino-nano-esp32",
+    "url": "https://docs.arduino.cc/hardware/nano-esp32/",
     "vendor": "Arduino"
 }
--- a/ports/esp32/boards/OLIMEX_ESP32_EVB/board.json
+++ b/ports/esp32/boards/OLIMEX_ESP32_EVB/board.json
@@ -18,8 +18,8 @@
         "ESP32-EVB_Rev_K1.png"
     ],
     "mcu": "esp32",
-    "product": "Olimex ESP32 EVB",
+    "product": "ESP32 EVB",
     "thumbnail": "",
-    "url": "https://www.olimex.com/Products/IoT/ESP32/ESP32-EVB/open-source-hardware",
-    "vendor": "OLIMEX"
+    "url": "https://www.olimex.com/Products/IoT/ESP32/ESP32-EVB",
+    "vendor": "Olimex"
 }
--- a/ports/mimxrt/boards/OLIMEX_RT1010/board.json
+++ b/ports/mimxrt/boards/OLIMEX_RT1010/board.json
@@ -12,8 +12,8 @@
         "OLIMEX_RT1010Py.jpg"
     ],
     "mcu": "mimxrt",
-    "product": "Olimex_RT1010Py",
+    "product": "RT1010-Py",
     "thumbnail": "",
-    "url": "https://www.olimex.com/Products/ARM/NXP",
-    "vendor": "OLIMEX"
+    "url": "https://www.olimex.com/Products/MicroPython/RT1010-Py",
+    "vendor": "Olimex"
 }
--- a/ports/stm32/boards/HYDRABUS/board.json
+++ b/ports/stm32/boards/HYDRABUS/board.json
@@ -10,6 +10,6 @@
     "mcu": "stm32f4",
     "product": "HydraBus v1.0",
     "thumbnail": "",
-    "url": "",
+    "url": "https://github.com/hydrabus/hydrabus",
     "vendor": "HydraBus"
 }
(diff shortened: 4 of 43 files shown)


ports/esp32/boards/ARDUINO_NANO_ESP32/board.json
```diff
@@ -14,7 +14,7 @@
         "ABX00092_01.iso_1000x750.jpg"
     ],
     "mcu": "esp32s3",
-    "product": "Arduino Nano ESP32",
+    "product": "Nano ESP32",
     "thumbnail": "",
     "url": "https://store.arduino.cc/products/arduino-nano-esp32",
```
dpgeorge, line 19: Looks like this link changed to `https://store.arduino.cc/products/nano-esp32`

ports/esp32/boards/OLIMEX_ESP32_EVB/board.json
```diff
@@ -15,8 +15,8 @@
         "ESP32-EVB_Rev_K1.png"
     ],
     "mcu": "esp32",
-    "product": "Olimex ESP32 EVB",
+    "product": "ESP32 EVB",
     "thumbnail": "",
-    "url": "https://www.olimex.com/Products/IoT/ESP32/ESP32-EVB/open-source-hardware",
-    "vendor": "OLIMEX"
+    "url": "https://www.olimex.com/Products/IoT/ESP32/ESP32-EVB",
+    "vendor": "Olimex"
```
dpgeorge, line 21: Should this stay all-caps?  Eg on https://www.olimex.com/Products/IoT/ESP32/ESP32-EVB/ it says:
```
ESP32-EVB is originally designed by OLIMEX Ltd and a result of continual improvements.
```

ports/mimxrt/boards/OLIMEX_RT1010/board.json
```diff
@@ -12,8 +12,8 @@
         "OLIMEX_RT1010Py.jpg"
     ],
     "mcu": "mimxrt",
-    "product": "Olimex_RT1010Py",
+    "product": "RT1010-Py",
     "thumbnail": "",
-    "url": "https://www.olimex.com/Products/ARM/NXP",
-    "vendor": "OLIMEX"
+    "url": "https://www.olimex.com/Products/MicroPython/RT1010-Py",
+    "vendor": "Olimex"
```
dpgeorge, line 18: Stay all-caps?

ports/stm32/boards/HYDRABUS/board.json
```diff
@@ -8,8 +8,8 @@
         "hydrabus.jpg"
     ],
     "mcu": "stm32f4",
-    "product": "HydraBus v1.0",
+    "product": "v1.0",
```
dpgeorge, line 11: I think this is a special case where it should stay "HydraBus v1.0".

ports/stm32/boards/NETDUINO_PLUS_2/board.json
```diff
@@ -8,7 +8,7 @@
         "netduino_plus_2.jpg"
     ],
     "mcu": "stm32f4",
-    "product": "Netduino Plus 2",
+    "product": "Plus 2",
```
dpgeorge, line 11: Maybe this one should also retain the vendor name?

dpgeorge:
> In particular, I wanted to remove any duplication of the vendor name in the product field so that `f"{vendor} {product}"` reads well.

That does seem like a good goal, but it might be awkward in some cases.  Eg some boards like "Arduino Nano ESP32" explicitly have the vendor name in the product name (otherwise Nano ESP32 is quite generic).  And others like "TinyS3" don't, and in that case "Unexpected Maker TinyS3" isn't really the board's name.  And then for boards like HydraBus and LimiFrog, there's only one board for the vendor so the vendor name is the board name.

So instead maybe we should keep it as-is and have the policy be:
- "vendor" should be the vendor name
- "product" should be the full product name as given on the product webpage / URL
- "product" should be able to be used standalone to refer to the board (it may not be fully unique, but that's the problem of the vendor, to chose a good name!)
- if we want to use both then it could be rendered as `f"{product} by {vendor}"`

Note that `ports/esp32/boards/deploy_flashmode.md` is now a template that uses product, eg:
```
To flash or erase your {product}, you have to first put it into download mode.
```
and it would be good to keep that as, eg, "Arduino Nano ESP32", rather than "Nano ESP32", and also just have it as "TinyS3" not "Unexpected Maker TinyS3".

ports/mimxrt/boards/OLIMEX_RT1010/board.json
mattytrentini: Olimex use their name in a mixture of upper and Pascal case (like on their [about page](https://www.olimex.com/About/)). [Mouser](https://au.mouser.com/manufacturer/olimex/) and [Digikey](https://www.digikey.com.au/en/supplier-centers/olimex) list them using Pascal case.

I find the upper case visually ugly so prefer the Pascal case but can use upper if that's your preference?

dpgeorge: You're right, it is a bit inconsistent.

It looks like this is their business registration: https://portal.registryagency.bg/CR/en/Reports/ActiveConditionTabResult?uic=825241718 and since it's Bulgarian there's no real English source of truth for the name, except how they themselves translate it.

Os, let's go with "Olimex".

dpgeorge (approved):
Looks good now, thank you!

PR #16032, opened 2024-10-17
ports/unix: Freeze `asyncio`.

vshymanskyy:
### Summary

`asyncio` is available for most ports out of the box, this adds it to the unix binaries.

### Testing

This now works without referencing any libraries on the filesystem:

```py
import asyncio
```

Alternatively, one needs to put `asyncio` somewhere on the filesystem and maybe configure `MICROPYPATH` to use it.

### DIFF
--- a/ports/unix/variants/manifest.py
+++ b/ports/unix/variants/manifest.py
@@ -1,3 +1,4 @@
+include("$(MPY_DIR)/extmod/asyncio")
 add_library("unix-ffi", "$(MPY_LIB_DIR)/unix-ffi")
 require("mip-cmdline")
 require("ssl")


vshymanskyy:
Sorry, seems like this is no longer needed for the `standard` build. Closing

andrewleech:
It is missing in the coverage build though, I've been confused by asyncio not working on some variants. 

I'd support keeping this PR and remving it from just the standard build manfest. Unless there's a reason it was excluded there @dpgeorge ?

dpgeorge:
> It is missing in the coverage build though,

That's because the coverage variant is intended to test the pure Python implementation of asyncio Task/TaskQueue.  The standard variant tests the C implementation.

Also, it should remain excluded from the minimal variant.

andrewleech:
Ah ok, that does make sense. Thanks for the clarification @dpgeorge !

PR #16036, opened 2024-10-17
ports/rp2: Enable mbedtls PEM parser.

vshymanskyy:
### Summary

Enable `.pem` certificate parser for tls/ssl.

### Testing

Working fine for a few month in Blynk.Edgent.

Use `.der` certificates, but DER format does not allow having multiple certificates in a single file

### DIFF
--- a/ports/rp2/mbedtls/mbedtls_config_port.h
+++ b/ports/rp2/mbedtls/mbedtls_config_port.h
@@ -29,6 +29,10 @@
 // Set mbedtls configuration
 #define MBEDTLS_ECP_NIST_OPTIM
 
+// Enable mbedtls modules
+#define MBEDTLS_BASE64_C
+#define MBEDTLS_PEM_PARSE_C
+
 // Time hook
 #include <time.h>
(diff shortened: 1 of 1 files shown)


dpgeorge:
I'm not against this change, but:
1. It's a big code size increase.
2. It's not adding anything that you can't already do (you can load multiple der files by loading them separately, right?).
3. It would be best to unify this configuration across all ports that use mbedTLS, not just rp2.  Otherwise the configuration gets fragmented.

Maybe we need to have "big" and "small" mbedTLS default configurations that ports can choose between.  "Big" would have things like this enabled.  "Small" would have just the bare minimum.

Also would be good to add a test for PEM parsing, if we are going to officially support that.

dpgeorge:
@vshymanskyy are you able to update this PR to move the change to `extmod/mbedtls/mbedtls_config_common.h` so that all ports can benefit from it?

dpgeorge:
I'll close this in favour of #19569, which enables PEM parsing on all ports using mbedTLS.

PR #16039, opened 2024-10-18
ports/qemu: Fix inline-if to make it ignore the highest bit of `mcaus…

Plaque-fcc:
This patch isn't optimal, yet it makes it consistent to work with the same 31 bits of mcause inside this switch-block.


Plaque-fcc:
Replaced with #16040.

PR #16040, opened 2024-10-18
ports/qemu: Fix inline-if in `lookup_cause`.

Plaque-fcc:
This patch isn't optimal, yet it makes it consistent to work with the same 31 bits of mcause inside this switch-block.

### DIFF
--- a/ports/qemu/mcu/rv32/interrupts.c
+++ b/ports/qemu/mcu/rv32/interrupts.c
@@ -144,7 +144,7 @@ const char *lookup_cause(uint32_t mcause) {
             case 11:
                 return exception_causes[6];
             default:
-                return (mcause >= 16) ?
+                return ((mcause & 0x7FFFFFFF) >= 16) ?
                        exception_causes[7] :
                        exception_causes[0];
         }


agatti:
Good catch!  I missed that bit back when I wrote the mtvec handler, thanks for finding that.  

However now I'm a bit curious about what sort of exception did you bump into to actually trigger that code path...

Plaque-fcc:
Me and my friend were discussing this MPy@RV32 implementation (it's complicated, lol), and our other friend noticed this inconsistency and proposed filing bugreport. For me, it took less time to file a PR, lol.

He said this code can be optimized, yet I prefer more clarity, at least for now.

agatti:
The table lookups can be optimised indeed, but the Qemu port is meant for debugging and testing.  Considering that the bit of code in question is called exactly once at the very end of the execution flow, well... :)

The interrupt handling setup can be probably be simplified by switching to a non-vectored interrupt handling scheme, but then you'd have to make the interrupt handler itself a bit more complicated as it has to not terminate execution on certain interrupts and so on.

Still, I noticed the same issue is also present in #15931, I'll probably update that PR later today.  Thanks again.

agatti (approved):

dpgeorge (approved):

dpgeorge:
Thanks for the fix!

PR #16041, opened 2024-10-18
lib/micropython-lib: Update submodule to latest.

dpgeorge:
This brings in:
- requests improvements with overriding headers
- use non-u versions of built-in modules, including asyncio
- fix to logging so StreamHandler calls parent constructor
- various fixes to usb-device packages
- fixes to lora sx126x and sx127x drivers
- improvements to unix-ffi/sqlite3
- support additional gap_connect arguments in aioble

This is done for the upcoming 1.24.0 release.


projectgus (approved):
Marking as approved although as discussed there's not much to review here, these changes were all reviewed when going into micropython-lib.

PR #16042, opened 2024-10-18
tools: Self-hosting of `mpy-tool`.

vshymanskyy:
### Summary

This allows running `mpy-tool` using MicroPython itself.

### Testing

Currently it is used in `ViperIDE` to disassemble `.mpy` files:
- MicroPython VM runs in-browser using WebAssembly
- The filesystem is [populated with the required libraries](https://github.com/vshymanskyy/ViperIDE/tree/main/src/tools_vfs)
- `.mpy` file is copied to the VM as well
- The `mpy-tool` is [used to dissect](https://github.com/vshymanskyy/ViperIDE/blob/7e42f28da5394c273b34d8f04835961676a43d8d/src/python_utils.js#L218-L233) the `.mpy`

Also, a test is added to `ci.sh`

None that I'm aware of

### DIFF
--- a/py/makeqstrdata.py
+++ b/py/makeqstrdata.py
@@ -9,17 +9,54 @@ from __future__ import print_function
 import re
 import sys
 
-# Python 2/3 compatibility:
+# Python 2/3/MicroPython compatibility:
 #   - iterating through bytes is different
-#   - codepoint2name lives in a different module
-import platform
-
-if platform.python_version_tuple()[0] == "2":
+#   - codepoint2name from html.entities is hard-coded
+if sys.version_info[0] == 2:
     bytes_cons = lambda val, enc=None: bytearray(val)
-    from htmlentitydefs import codepoint2name
-elif platform.python_version_tuple()[0] == "3":
+elif sys.version_info[0] == 3:  # Also handles MicroPython
     bytes_cons = bytes
-    from html.entities import codepoint2name
+
+# fmt: off
+codepoint2name = {
+    198: "AElig", 193: "Aacute", 194: "Acirc", 192: "Agrave", 913: "Alpha", 197: "Aring", 195: "Atilde",
+    196: "Auml", 914: "Beta", 199: "Ccedil", 935: "Chi", 8225: "Dagger", 916: "Delta", 208: "ETH",
+    201: "Eacute", 202: "Ecirc", 200: "Egrave", 917: "Epsilon", 919: "Eta", 203: "Euml", 915: "Gamma",
+    205: "Iacute", 206: "Icirc", 204: "Igrave", 921: "Iota", 207: "Iuml", 922: "Kappa", 923: "Lambda",
+    924: "Mu", 209: "Ntilde", 925: "Nu", 338: "OElig", 211: "Oacute", 212: "Ocirc", 210: "Ograve",
+    937: "Omega", 927: "Omicron", 216: "Oslash", 213: "Otilde", 214: "Ouml", 934: "Phi", 928: "Pi",
+    8243: "Prime", 936: "Psi", 929: "Rho", 352: "Scaron", 931: "Sigma", 222: "THORN", 932: "Tau",
+    920: "Theta", 218: "Uacute", 219: "Ucirc", 217: "Ugrave", 933: "Upsilon", 220: "Uuml", 926: "Xi",
+    221: "Yacute", 376: "Yuml", 918: "Zeta", 225: "aacute", 226: "acirc", 180: "acute", 230: "aelig",
+    224: "agrave", 8501: "alefsym", 945: "alpha", 38: "amp", 8743: "and", 8736: "ang", 229: "aring",
+    8776: "asymp", 227: "atilde", 228: "auml", 8222: "bdquo", 946: "beta", 166: "brvbar", 8226: "bull",
+    8745: "cap", 231: "ccedil", 184: "cedil", 162: "cent", 967: "chi", 710: "circ", 9827: "clubs",
+    8773: "cong", 169: "copy", 8629: "crarr", 8746: "cup", 164: "curren", 8659: "dArr", 8224: "dagger",
+    8595: "darr", 176: "deg", 948: "delta", 9830: "diams", 247: "divide", 233: "eacute", 234: "ecirc",
+    232: "egrave", 8709: "empty", 8195: "emsp", 8194: "ensp", 949: "epsilon", 8801: "equiv", 951: "eta",
+    240: "eth", 235: "euml", 8364: "euro", 8707: "exist", 402: "fnof", 8704: "forall", 189: "frac12",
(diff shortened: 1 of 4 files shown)


py/makeqstrdata.py
```diff
@@ -9,17 +9,54 @@
 import re
 import sys
 
-# Python 2/3 compatibility:
+# Python 2/3/micro compatibility:
```
dpgeorge, line 12: "MicroPython" instead of "micro"

py/makeqstrdata.py
```diff
@@ -9,17 +9,54 @@
...
+if sys.version_info[0] == 2:
     bytes_cons = lambda val, enc=None: bytearray(val)
-    from htmlentitydefs import codepoint2name
-elif platform.python_version_tuple()[0] == "3":
+elif sys.version_info[0] == 3:  # Also handles MicroPython
     bytes_cons = bytes
-    from html.entities import codepoint2name
+
+# fmt: off
+codepoint2name = {
```
dpgeorge, line 21: Is this copied verbatim from the Python source?  Is there any licensing/copyright issue doing that?

Is it worth instead making a `micropython-lib` module for `html.entities` that has this dict there?

(In general I'm a fan of keeping things self contained, so having this dict verbatim here, but in this case maybe it's better to make a library for it.)

py/makeqstrdata.py
```diff
@@ -307,15 +344,17 @@ def compute_hash(qstr, bytes_hash):
 
 
 def qstr_escape(qst):
-    def esc_char(m):
-        c = ord(m.group(0))
+    def esc_char(c):
+        if re.match(r"^[a-zA-Z0-9_]$", c):
+            return c
+        c = ord(c)
```
dpgeorge, line 350: Instead of using `re` here, maybe it can simply be `c.isalnum() or c == "_"`?

py/makeqstrdata.py
```diff
@@ -307,15 +344,17 @@ def compute_hash(qstr, bytes_hash):
...
         try:
             name = codepoint2name[c]
         except KeyError:
             name = "0x%02x" % c
         return "_" + name + "_"
 
-    return re.sub(r"[^A-Za-z0-9_]", esc_char, qst)
+    return "".join(map(esc_char, qst))
```
dpgeorge, line 357: Why doesn't this work in MicroPython, is it simply because `re.sub` is not always enabled on MicroPython?

tools/ci.sh
```diff
@@ -96,9 +96,19 @@ function ci_mpy_format_test {
     python2 ./tools/mpy-tool.py -xd tests/frozen/frozentest.mpy
     python3 ./tools/mpy-tool.py -xd tests/frozen/frozentest.mpy
 
+    # Build MicroPython
+    ci_unix_standard_build
+    micropython=./ports/unix/build-standard/micropython
+    $micropython -m mip install --target . argparse __future__
+    export MICROPYPATH=.
+
+    # Test mpy-tool.py in MicroPython
```
dpgeorge, line 105: "Test mpy-tool.py running under MicroPython"

tools/mpy-tool.py
```diff
@@ -1765,7 +1765,7 @@ def copy_section(file, offset, offset2):
             f.write(merged_mpy)
 
 
-def main():
+def main(args=None):
```
dpgeorge, line 1768: This doesn't look like it's related to MicroPython compatibility.  What's it needed for?  Can it be in a separate commit?

dpgeorge (commented):
This is great, definitely good to be able to self-host these basic tools.

Can you please split out the change to `makeqstrdata.py` to a separate commit, because that's independently making `makeqstrdata.py` MicroPython-compatible.

dpgeorge:
@vshymanskyy any comments on the above review?

py/makeqstrdata.py
vshymanskyy: No, this is not copied from Python source, see: https://github.com/python/cpython/blob/74a517181a9bb65a1f6da149af7427a9fcb3add3/Lib/html/entities.py#L2503

Yes I think I can create a separate PR to add this to `micropython-lib`. Given that `argparse`, `__future__` are needed anyway, adding another dependency is not an issue.

However, I'm wondering if increasing the code size of `micropython-lib/html` is justified given that these modules are installed into microcontrollers most of the time.

py/makeqstrdata.py
vshymanskyy: Correct, `re.sub` is not always enabled. I think it was disabled in WebAssembly port when I was developing it.

tools/mpy-tool.py
vshymanskyy: This is a small change that enables importing and using mpy-tool from other modules.
If this is undesired, I'll simply exclude it from this PR.

py/makeqstrdata.py
dpgeorge: OK, for now let's just leave it as you have it, with this dict hard-coded here.

py/makeqstrdata.py
dpgeorge: I think for performance, either:
1. precompile the regex in `qstr_escape` (`re.compile` should be supported in all MicroPython builds that include the `re` module)
2. or use `c in "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz01234567890_"`

py/makeqstrdata.py
dpgeorge: OK.

tools/mpy-tool.py
```diff
@@ -26,9 +26,9 @@
 
 # Python 2/3 compatibility code
```
dpgeorge, line 27: Also add MicroPython: `2/3/MicroPython compatibility code`.

tools/mpy-tool.py
dpgeorge: OK, this change is fine.  But please put it in a separate commit in this PR, because it's logically doing something independent to the rest of the changes here.

dpgeorge:
Can you please also split out the change to `makeqstrdata.py` to a separate commit (or, I can do that during merge, but it's easier if you do it).

projectgus (approved):
This LGTM, I don't have anything to add.

@vshymanskyy Do you have time to run through the minor outstanding discussions, or would you rather one of us picked them up?

jonnor:
@projectgus I believe @vshymanskyy said on Discord that he is rather preoccupied at the moment (the situation in Ukraine remains difficult). So I am sure he would appreciate it if anyone can do the small things to get this unblocked and merged :)

projectgus:
I've done an update according to the review. I'd not rebased, so changes can be viewed via https://github.com/micropython/micropython/compare/0289e32a9afdb87d82cde42996f5bf924a17517e..efbd3f7

However I see that there's a merge conflict, so I'll rebase now.

projectgus:
Now rebased. Use the compare link above to see the pre-rebase changes.

dpgeorge (approved):

dpgeorge:
This looks good now, thanks @projectgus for updating.

PR #16043, opened 2024-10-18
tests: Adjust tests for axtls, esp8266 and targets with terse error messages

dpgeorge:
### Summary

Adjust the tests so esp8266 passes the test suite.  Does not change any firmware code.

### Testing

Tested on esp8266 and PYBD-SF2.


tests/micropython/ringio.py
```diff
@@ -45,4 +45,4 @@
     # Size must be int.
     micropython.RingIO(None)
 except TypeError as ex:
-    print(ex)
+    print(repr(ex)[:24], repr(ex)[-9:])
```
projectgus, line 48: This seems a bit opaque... what about `print("to int" in repr(ex)) # error string depends on port` or even `pass`?

projectgus (approved):
Looks good to me! Left one minor suggestion about error message wrangling.

tests/micropython/ringio.py
dpgeorge: It definitely needs to print something here, otherwise the test will pass even if the exception is not raised.

I think some other tests do this string mangling, but here it's not really necessary, so I changed it to `print(type(ex))`, which is the same on all builds.

Retested on PYBD-SF2 and ESP8266_GENERIC.

projectgus: Looks good to me!

PR #16050, opened 2024-10-21
tests/extmod: Support stm32 port in machine.UART TX test.

dpgeorge:
### Summary

Getting this test running on stm32-based boards requires:
- Adding a small delay after constructing the UART so that the initial idle frame has time to be transmitted before the test starts.
- Adding a timing margin because the time for a UART write is always going to be longer than the exact calculation due to overhead of loading and calling methods (rp2 doesn't have this issue because it doesn't wait for the last byte to go out before returning from `flush()`).

### Testing

This test now passes on PYBv1.0 and PYB-SF2 (and probably all other stm32 boards).  Also retested on Pico and it passes.

### DIFF
--- a/tests/extmod/machine_uart_tx.py
+++ b/tests/extmod/machine_uart_tx.py
@@ -9,12 +9,34 @@ except ImportError:
 
 import time, sys
 
+initial_delay_ms = 0
+bit_margin = 0
+timing_margin_us = 100
+
 # Configure pins based on the target.
-if "rp2" in sys.platform:
+if "esp32" in sys.platform:
+    uart_id = 1
+    pins = {}
+    timing_margin_us = 400
+elif "mimxrt" in sys.platform:
+    uart_id = 1
+    pins = {}
+    initial_delay_ms = 20  # UART sends idle frame after init, so wait for that
+    bit_margin = 1
+elif "pyboard" in sys.platform:
+    uart_id = 4
+    pins = {}
+    initial_delay_ms = 50  # UART sends idle frame after init, so wait for that
+    bit_margin = 1  # first start-bit must wait to sync with the UART clock
+elif "rp2" in sys.platform:
     uart_id = 0
-    tx_pin = "GPIO0"
-    rx_pin = "GPIO1"
+    pins = {"tx": "GPIO0", "rx": "GPIO1"}
     timing_margin_us = 180
+elif "samd" in sys.platform:
+    uart_id = 2
+    pins = {"tx": "D1", "rx": "D0"}
+    timing_margin_us = 300
+    bit_margin = 1
 else:
     print("SKIP")
     raise SystemExit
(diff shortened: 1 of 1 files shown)


robert-hh:
Looks good to me. I tested that with PICO2 (both MCUs) and the most recent change of #16052 and the test passes. Returning after a call to flush() varies between ports. Some are early, some are late. With PR #16052 Pico should be on time now. I tried this test with support for MIMXRT, ESP32 and SAMD as well. Code snippet below. MIMXRT needs an initial delay and a timing_margin of at least a half bit time. SAMD does not need an initial delay, but a timing margin 1200µs.
```
elif "mimxrt" in sys.platform:
    uart_id = 1
    pins = {}
    initial_delay_ms = 20  # UART sends idle frame after init, so wait for that
    timing_margin_us = 400
elif "samd" in sys.platform:
    uart_id = 3
    pins = {"tx": "D1", "rx": "D0"}
    timing_margin_us = 1500
elif "esp32" in sys.platform:
    uart_id = 1
    pins = {}
    timing_margin_us = 400
```
MIMXRT tested with Teensy 4 (MIMXRT1062), MIMXRT1020, MIMXRT1052 and some MIMXRT1011 boards.
SAMD tested with ItsyBitsy M4.
ESP32 tested with ESP32, ESP32S2, ESP32S3, ESP32C3
I can add more ports if you like.

robert-hh:
The test does fail at the NRF port, since it does not accept keyword arguments at the UART constructor. 
At the Renesas port, `timing_margin_us` for 2400 has to be set to ~5000µs to pass. At higher baud rates the `timing_margin_us` can be smaller.
For STM32 `timing_margin_us` must be set somewhat longer for slower boards like the F411 blackpill.
At the SAMD port `timing_margin_us` can be set to lower value of 500 if instead of uart.flush() a busy wait like `while uart.txdone() is False: pass` is used.

dpgeorge:
Now that #16052 is merged I've updated this PR, adding the above ports from @robert-hh .

I added a `bit_margin` parameter which adds to the timing margin a 1-bit worth of timing, which is needed on some ports, eg stm32 which must sync with the UART clock before sending out the first start bit, and that can take between 0 and 1/baudrate time to occur.  This setting is also useful on mimxrt.

I also changed the SAMD port to 2 so it would work on an Adafruit ItsyBitsy M0 Express.

Tested on: ESP32_GENERIC (TinyPICO), Teensy 4.0, ItsyBitsy M0 Express, Pico, PYBv1.0, PYBD-SF2.

@robert-hh Please take a look, thanks!

robert-hh:
Re-tested and confirmed to work with a similar set of boards. Added ESP32S3, PYBD-SF6, WEACT STM32F407, Olimex_RT1011. It fails with SAMD51 ItsyBitsy M4 Express after changing the UART to 3, caused by the `MICROPY_EVENT_POLL_HOOK` call in flush(). Surprisingly that does not apply to SAMD21.

dpgeorge:
> It fails with SAMD51 ItsyBitsy M4 Express after changing the UART to 3, caused by the `MICROPY_EVENT_POLL_HOOK` call in flush().

Interesting.  Let's investigate and fix that separately.

Does this otherwise look good to merge?

dpgeorge:
Great, thanks!

PR #16051, opened 2024-10-21
tests/extmod: Make invalid-blockdev test work consistently on all ports.

dpgeorge:
### Summary

Some ports (eg stm32) configure the FAT driver differently (eg with multi-partition support) and that leads to a slightly different sequence of block reads, compared to other configurations (eg rp2).  Adjust the test so it passes for the different configurations.

Also print out more information about the arguments passed to `readblocks` for easier debugging.  They should be deterministic for a given filesystem implementation.

### Testing

This test tested to now work on PYBD-SF2, Pico and ESP32 (generic).

### DIFF
--- a/tests/extmod/vfs_blockdev_invalid.py
+++ b/tests/extmod/vfs_blockdev_invalid.py
@@ -19,7 +19,7 @@ class RAMBlockDevice:
         self.write_res = 0
 
     def readblocks(self, block, buf, off=0):
-        print("readblocks")
+        # print("readblocks", block, len(buf), off)
         addr = block * self.ERASE_BLOCK_SIZE + off
         for i in range(len(buf)):
             buf[i] = self.data[addr + i]
--- a/tests/extmod/vfs_blockdev_invalid.py.exp
+++ b/tests/extmod/vfs_blockdev_invalid.py.exp
@@ -1,126 +1,28 @@
 <class 'VfsLfs2'>
-readblocks
-readblocks
-readblocks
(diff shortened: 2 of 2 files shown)


tests/extmod/vfs_blockdev_invalid.py
```diff
@@ -17,9 +17,16 @@ def __init__(self, blocks):
...
+        self.readblocks_history = []
 
     def readblocks(self, block, buf, off=0):
-        print("readblocks")
+        # FAT filesystem access may insert an additional read to blocks 0 and 2 in
+        # certain build configurations.  Don't print out those accesses to ensure that
+        # the test output is consistent across targets.
+        self.readblocks_history.append(block)
+        if self.readblocks_history not in ([0, 0, 2, 0], [0, 0, 2, 0, 2]):
+            print("readblocks", block, len(buf), off)
```
projectgus, line 28: I know I wrote this test to `print("readblocks")` here but I'm not sure if it's really necessary for anything this test verifies. Plus it couples the test very strongly to the implementation details of the filesystem, making it brittle (as per this issue).

I reckon it might be enough to replace all of this with a commented out print statement, that can be un-commented for debugging. What do you think?

dpgeorge: I added a commit to comment out the print (will squash before merge).

Retested on PYBD-SF2, Pico, ESP32_GENERIC.

projectgus (approved):

PR #16052, opened 2024-10-21
rp2/machine_uart: Make it so TX is done only when no longer busy.

dpgeorge:
### Summary

When flushing a UART on the rp2 port, it returns just before the last character is sent out the wire.

Fix this by waiting until the BUSY flag is cleared.

### Testing

TODO: the `tests/extmod/machine_uart_tx.py` test needs updating to cover this case.  Currently that test checks that transmit time is less than the expected time, and on rp2 the transmit+flush time is quite a bit less than expected (because it returns while there's still a byte left to transmit).  Rather it should be changed so transmit time is very close to the expected time.

Edit: that's now done.  Also docs updated.

### DIFF
--- a/tests/extmod/machine_uart_tx.py
+++ b/tests/extmod/machine_uart_tx.py
@@ -14,6 +14,7 @@ if "rp2" in sys.platform:
     uart_id = 0
     tx_pin = "GPIO0"
     rx_pin = "GPIO1"
+    timing_margin_us = 180
 else:
     print("SKIP")
     raise SystemExit
@@ -31,4 +32,5 @@ for bits_per_s in (2400, 9600, 115200):
     # 1(startbit) + 8(bits) + 1(stopbit) + 0(parity)
     bits_per_char = 10
     expect_us = (len(text)) * bits_per_char * 1_000_000 // bits_per_s
-    print(bits_per_s, duration_us <= expect_us)
+    delta_us = abs(duration_us - expect_us)
+    print(bits_per_s, delta_us <= timing_margin_us or delta_us)
--- a/docs/library/machine.UART.rst
+++ b/docs/library/machine.UART.rst
@@ -160,7 +160,7 @@ Methods
 
    .. note::
 
-       For the rp2, esp8266 and nrf ports the call returns while the last byte is sent.
+       For the esp8266 and nrf ports the call returns while the last byte is sent.
        If required, a one character wait time has to be added in the calling script.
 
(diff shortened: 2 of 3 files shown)


dpgeorge:
@robert-hh what do you think of this patch?

robert-hh:
> @robert-hh what do you think of this patch?

By reading the data sheet it looks good. I'll grabed a RP2 tested it with short and long messages. I will test it with PICO2 as well.
There is a note in the documentation about tx_done and flush() returning while the last byte is sent. That can be changed as well.

robert-hh:
The change works well too with Pico2 with both ARM and RISCV.

dpgeorge:
Thanks @robert-hh for the review and testing.  I have now updated the docs (thanks for pointing that out) and the test so that it checks `abs(actual - expected) <= margin`.

I tested on Pico, and Pico 2 in ARM and RISCV mode.  The test passes.

Please see what you think!

tests/extmod/machine_uart_tx.py
```diff
@@ -14,6 +14,7 @@
     uart_id = 0
     tx_pin = "GPIO0"
     rx_pin = "GPIO1"
+    timing_margin_us = 160
```
robert-hh, line 17: Looking at the achieved values of delta_us, it's 135 - 145µs for RP2 at 2400 Baud and it seems well distributed (without making real statistical tests). So 160µs does not look too tight.

grafik

robert-hh (approved):
Tested as well with PICO, PICO2_ARM and PICO2_RISCV. uart.flush() returns after the stop bit has been sent.

robert-hh:
I ran the test a little bit more often. The first call is after loading is a little bit slower, and then it happens that delay_us is larger than 160 at a RP2. So maybe it's better to increase`timing_margin_us` a little bit, say to 200.

grafik

dpgeorge:
> The first call is after loading is a little bit slower, and then it happens that delay_us is larger than 160 at a RP2.

Yeah, that's also what I saw, the first call is slow because it needs to load the code from flash (not just the Python code but also the relevant C code for the UART driver), and after that the code is in cache so it's faster (this is a general limitation of devices with external, executable flash for the firmware).

I'll increase the margin to 180us.  Making it too big and other errors could sneak in undetected.

tests/extmod/machine_uart_tx.py
dpgeorge: Thanks for the thorough testing!

PR #16057, opened 2024-10-21
Support for RM2 break out boards

peterharperuk:
Adds changes needed to support Pico 2 W. This requires the develop version of the sdk so I've kept the PR in draft.


peterharperuk:
I see one test failure for rp2040 Pico in rp2_dma which implies DMA has go slower. It does appear to be caused by the sleep change which doesn't make a lot of sense. I tried taking an average over a number of runs and get much better results - even if it's a loop of 1?!

dpgeorge:
> I see one test failure for rp2040 Pico in rp2_dma which implies DMA has go slower. It does appear to be caused by the sleep change which doesn't make a lot of sense. I tried taking an average over a number of runs and get much better results - even if it's a loop of 1?!

That could be due to flash caching effects, where the first run through code is slower but after that it's faster.

ports/rp2/mphalport.c
```diff
@@ -266,15 +266,13 @@ void soft_timer_init(void) {
...
-    soft_timer_remove(&timer);
+    best_effort_wfe_or_timeout(delayed_by_ms(get_absolute_time(), timeout_ms));
 }
+
+int mp_hal_is_pin_reserved(int n) {
+#if MICROPY_PY_NETWORK_CYW43
+    return (n == CYW43_PIN_WL_HOST_WAKE);
+#else
+    return false;
+#endif
```
dpgeorge, line 277: Please indent the `#` preprocessor lines to 4 spaces.

ports/rp2/mpnetworkport.c
```diff
@@ -77,11 +77,64 @@ void cyw43_irq_init(void) {
...
+            mp_obj_get_int(args[2]), //CYW43_DEFAULT_PIN_WL_CLOCK,
+            mp_obj_get_int(args[3]), //CYW43_DEFAULT_PIN_WL_CS
+        };
+        cyw43_irq_deinit();
+        cyw43_set_pins_wl(pins);
+        cyw43_irq_init();
+    }
+    return mp_const_none;
+}
+MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(network_cyw43_set_pins_wl_obj, 4, 6, network_cyw43_set_pins_wl);
```
dpgeorge, line 124: Not sure that this is the best way to expose this configuration as a Python-level API.  For this kind of thing the usual approach would be to use the WLAN constructor, eg:
```py
wl = network.WLAN(reg_on=Pin(4), data_out=Pin(5), ..., freq=1_000_000)
```
And then have sensible defaults so it works on Pico 2 W with just `wl = network.WLAN()`.

peterharperuk:
> That could be due to flash caching effects, where the first run through code is slower but after that it's faster.

Maybe. Resetting src before the copy makes it twice as fast. Maybe that puts it in the cache? I don't quite understand why this is faster however...

```
for i in range(0,1):
    dt = run_and_time_dma(dma)
```

ports/rp2/mpnetworkport.c
peterharperuk: That would be neater. But these parameters are pico specific and network_cyw43_make_new is "common". I'll see if I can override this.

peterharperuk: I've had a go at doing this. I had to make changes to extmod/network_cyw43.c to make network_cyw43_make_new configurable.  So now you can do something like this...

wlan = network.WLAN(network.STA_IF, pin_on=2, pin_out=4, pin_in=4, pin_wake=4, pin_clock=5, pin_cs=3, div=3, div_frac=0)

peterharperuk:
I've taken the liberty to add the ability to use cyw43 with normal Pico or Pico 2 devices as we're starting to see cyw43 breakout boards appearing https://shop.pimoroni.com/products/rm2-breakout?variant=53492995719547

If you wanted to do this you have to enable it and define the default pins to use...

```
cmaked .. -DPICO_BOARD=pico -DPICO_CYW43_SUPPORTED=1 -DCYW43_DEFAULT_PIN_WL_REG_ON=2 -DCYW43_DEFAULT_PIN_WL_DATA_OUT=4 -DCYW43_DEFAULT_PIN_WL_DATA_IN=4 -DCYW43_DEFAULT_PIN_WL_HOST_WAKE=4 -DCYW43_DEFAULT_PIN_WL_CLOCK=5 -DCYW43_DEFAULT_PIN_WL_CS=3 -DCYW43_PIO_CLOCK_DIV_INT=3
```

You can use different pins by configuring them at run time.

```
wlan = network.WLAN(network.STA_IF, pin_on=6, pin_out=8, pin_in=8, pin_wake=8, pin_clock=9, pin_cs=7, div=3)
```

ports/rp2/mpnetworkport.c
Gadgetoid: I like this, it's *much* nicer than my solution.

How does this play with the early pin setup that happens in `main.c` before any user code is run? It would be nice to decouple CYW43 from `main.c` altogether if possible, it doesn't feel like it should get any special treatment as a network interface, but it's doing all of this up front:

https://github.com/micropython/micropython/blob/18c9abc32930728d0403d8c13994adc3cabba8db/ports/rp2/main.c#L135-L155

Could this be moved to your custom `network_cyw43_make_new` somehow?

peterharperuk: That's probably there for "historical" reasons. I'll have to check if we can remove it or put it somewhere better.
I think I should add "pin_dat" that sets pin_in, pin_out and pin_wake in one go, as they should really always be the same.

Gadgetoid: > That's probably there for "historical" reasons.

Oof. Programming terms to make your blood run cold!

> I think I should add "pin_dat" that sets pin_in, pin_out and pin_wake in one go, as they should really always be the same.

I don't foresee any situation where they wouldn't be, at least for Pimoroni boards or breakouts. Is it even possible for them not to be with the PIO driver? Anyway - yes please, the shorter the better since we'll definitely drop my hacky `set_pins` function in favour of this.

peterharperuk: > Is it even possible for them not to be with the PIO driver?

It looks like it might work, but it's untested so who knows!

peterharperuk: I added pin_dat...

wlan = network.WLAN(network.STA_IF, pin_on=2, pin_cs=3, pin_dat=4, pin_clock=5, div_int=3, div_frac=0)

And moved the cyw43 init code from main.

Gadgetoid:
Awesome stuff. Will switch our wireless builds over to this and see how it fares!

Gadgetoid:
This *possibly* breaks PPP support, though I can't seem to figure out why. Might be a configuration quirk.

Gadgetoid:
Also seem to be having some trouble with WiFi not working on soft reset. If I run the same code twice I need to press reset on a board or I'll get connection aborted or `OSError -2`.

I'm also having a lot of trouble establishing a connection at all which might be related to my router, or my early-stage development boards, or something else, but I know I'm not the only one encountering strangeness. I can't seem to quantify it any better than that- though- and without a Pico 2 W as a reference it's tricky to eliminate any external factors.

peterharperuk:
I've tested this with a Pico 2 and your RM2 breakout. I sometimes have to slow the clock div to 3. I believe this was the command line...

import network
wlan = network.WLAN(network.STA_IF, pin_on=2, pin_cs=3, pin_dat=4, pin_clock=5, div_int=3)

I'll check what happens after a soft reset.

Gadgetoid:
Can't replicate on a Pico W, even if I enable dynamic pins. Either specifically a Pico 2 / RP2350 problem, or something up with our batteries-included builds.

peterharperuk:
The only problem I have "sometimes" is when I pull the plug on a connected device. It will often take a couple of attempts to reconnect "next time". But it sounds like you have a bigger issue. I can send you a Pico 2 W if you like. Otherwise I'll just spend some time playing around with your RM2 breakout.

Gadgetoid:
We're still seeing an issue establishing a connection on the first try, which I'm working on narrowing down. Otherwise, things are working much, much better. It was definitely related to MicroPython's soft reset pin cleanup trouncing the comms pin. (Might be fun with a wireless module for Yukon 😆)

Gadgetoid:
Since it would probably be difficult to support multiple wireless interfaces (though it would be really cool), should the `network.WLAN()` class currently prevent the user from creating more than *one* instance?

peterharperuk:
> multiple wireless interfaces

I wondered when someone was going to ask for that feature :) I guess it's more of a problem now you have the ability to change the pins.

peterharperuk:
Yeah it should work (I've had a wifi and lan connections working at the same time with a wiznet evb). The wifi driver relies on cyw43_state global all over the place - need to fix that somehow.

anecdata:
Some use cases for multiple network interfaces, fwiw:
https://gist.github.com/anecdata/456524f8e38c207931afd0dedf6bec89?permalink_comment_id=5040949#gistcomment-5040949

peterharperuk:
@anecdata try raising an issue in here https://github.com/georgerobotics/cyw43-driver (note: you can already connect Wifi in STA and AP at the same time).

ports/rp2/enable_cyw43.h
```diff
@@ -0,0 +1,25 @@
+// Enable networking.
+#define MICROPY_PY_NETWORK 1
+
+#ifndef MICROPY_PY_NETWORK_HOSTNAME_DEFAULT
+#define MICROPY_PY_NETWORK_HOSTNAME_DEFAULT "Pico"
```
peterharperuk, line 5: This should be set for PicoW? And Pico2W?

anecdata:
@peterharperuk Being limited to same channel is the limiting factor in most of those use cases I listed. One radio, one channel at a time, so probably not very useful as driver change. But perhaps MicroPython wants to be able to support more than one RM2 module (and maybe it will after these changes).

peterharperuk:
```
void mod_network_lwip_init(void) {
    soft_timer_static_init(
        &mp_network_soft_timer,
        SOFT_TIMER_MODE_PERIODIC,
        LWIP_TICK_RATE_MS,
        mp_network_soft_timer_callback
        );

    soft_timer_reinsert(&mp_network_soft_timer, LWIP_TICK_RATE_MS);
}
```

This needs to be fixed! or LwIP won't work properly!

Gadgetoid:
PPP tested okay on both Pico 2 W with the LTE module deadbugged onto GPIO, and Pico Plus 2 W using the SP/CE connector. Seems to have fixed it, thank you!

Gadgetoid:
Note: picotool `develop` has changes which means the `pico-sdk` module here needs bumped to avoid it breaking when it fetches/compiles.

@peterharperuk is there a means for initialising Bluetooth with custom pins? I must attest to not tinkering with it at all on these builds yet.

peterharperuk:
> is there a means for initialising Bluetooth with custom pins?

Sorry? What Bluetooth pins? I don't think there are any?

Gadgetoid:
Bluetooth also uses the CYW43 module, I think via this driver here? https://github.com/micropython/micropython/blob/master/drivers/cyw43/cywbt.c

And the bluetooth module: https://github.com/micropython/micropython/blob/master/extmod/modbluetooth.c

If you import and use Bluetooth I think it initialises the pins, PIO and wireless module so it can talk to it?

This is done with:

```
import bluetooth
ble = bluetooth.BLE()
```

So there's no way to tell the driver what the CYW43 pins should be (if they are dynamic).

peterharperuk:
We don't use cywbt.c, probably just by luck as it's not in the cmake file. But yes, this seems to have been an oversight. Let me take a look (and update the sdk).

Gadgetoid:
Thank you!

ports/rp2/rp2_network_cyw43.c
```diff
@@ -63,6 +63,11 @@ mp_obj_t rp2_network_cyw43_make_new(const mp_obj_type_t *type, size_t n_args, si
     mp_arg_parse_all_kw_array(n_args, n_kw, all_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
     rp2_network_cyw43_init();
 
+    // Can't call this again while the driver is initialised
+    if (cyw43_poll) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("wifi in use"));
```
peterharperuk, line 68: This is a bad idea

peterharperuk:
@Gadgetoid I've pushed a change that I think fixes bluetooth. So now you can do this...

```
ble = bluetooth.BLE(pin_on=6, pin_cs=7, pin_dat=8, pin_clock=9, div_int=3)
```

I need to do some more testing.

Gadgetoid:
> I need to do some more testing.

I'm still searching for a compelling thing to do with Bluetooth LE.

Will run some tests as soon as I can!

peterharperuk:
The led doesn't work with this change. I think moving cyw43 init from main, is causing more trouble than it's worth. And maybe network_cyw43_set_pins_wl was the best idea after all?

Gadgetoid:
> I think moving cyw43 init from main, is causing more trouble than it's worth. And maybe network_cyw43_set_pins_wl was the best idea after all?

There should be an underlying hardware init function that both Bluetooth and WLAN tap into that could be called in main if `CYW43_PIN_WL_DYNAMIC` is not set?

That said - and fully aware of the fact I've got a PSRAM PR that brings up PSRAM in `main` - it feels a little bit backwards to bring up non-standard peripherals in main, in C, rather than setting them up in `boot.py`. IE: AIUI the prevailing direction of MicroPython is *more Python less C*.

It's a tricky one... but perhaps we do need an `rp2.cyw32_init()` or similar?

There's certainly a bit of a conflict between CYW43 as an external peripheral and CYW43 as an integral part of Pico W. 😬 I'm very much in favour of making it as generic as possible, but the LED is a... painful edge case.

peterharperuk:
Well I've pushed a commit which might fix the led. And if you need to change pins you can do it like this...

ext_led = Pin('WL_GPIO0', Pin.OUT, pin_on=6, pin_cs=7, pin_dat=8,
pin_clock=9, div_int=3)

I'm not sure Damien will like all these changes?

Gadgetoid:
> I'm not sure Damien will like all these changes?

It's a lot of turmoil for something that's already battle-tested,  but I think there's demand for it from more than just us 😆 

If we're rolling another pre-release build for a while then... it wouldn't be the first time! Though it would be good to have the API locked down to keep any turmoil to a minimum.

It feels a little wrong to have three different places to be passing pins to init the CYW43, that's three different places to document, but it's at least explicit about what's going on. Could definitely use some more eyes on this...

peterharperuk:
> It feels a little wrong to have three different places to be passing pins to init the CYW43

First call wins, then you don't need to bother. And at the moment it looks like you can't reload cyw43 on different pins.
I may rewrite what I've done, the current code feels messy.
I may split all this into a separate PR - I bundled it with the the Pico 2 W stuff because it was supposed to be a small change.

Gadgetoid:
Agreed it does feel like it's got out of hand and a bit byzantine for a proper peer review.

I'm not sure we'd ever *need* to reload it on different pins, but I suspect someone will come along and find a means. It *feels* like the driver should be possible to instantiate as many times, on as many combinations of pins, but that is another monster of a change.

Thanks for doing all the legwork here, by the way! I'm usually on my own when we brush up against the bleeding edge.

peterharperuk:
I've rebased and squashed some commits, so the changes *might* make more sense now.

ports/rp2/mphalport.c
```diff
@@ -268,3 +268,11 @@ void soft_timer_init(void) {
 void mp_wfe_or_timeout(uint32_t timeout_ms) {
     best_effort_wfe_or_timeout(delayed_by_ms(get_absolute_time(), timeout_ms));
 }
+
+int mp_hal_is_pin_reserved(int n) {
+    #if MICROPY_PY_NETWORK_CYW43
+    return n == CYW43_PIN_WL_HOST_WAKE;
```
peterharperuk, line 274: CYW43_PIN_WL_REG_ON?

peterharperuk: Not needed as all this is used for is to avoid disabling irqs in machine_pin_deinit on a soft reset. Host wake is the only gpio that has this.

Gadgetoid:
Can this now be bumped to use Pico SDK 2.1.0?

Playing devil's advocate for the maintainers, it might also be worth pulling out the not-strictly-related-to-WiFi changes into 
one (or two?) separate PRs? - 

rp2: Bump to Pico SDK 2.1.0
* Bump SDK to 2.1.0
* 7d012c7202e3fa4ca96853989c26482703bf966d

rp2: Fix USB sleep on RP2350
* 4ac6c2ad3e1b6451985615bcedba6fde70bd6c91
* 1a83ea4a9cfeb30ba5cf7fbcb5c0ecb5f367af12

With the hope these could be merged relatively quickly and benefit all Pico W/non-W users.

@dpgeorge thoughts?

peterharperuk:
I've updated the sdk version.

These are updates for fixes that have been made to the sdk.

4ac6c2ad3 rp2: Fix USB sleep on 2350.
7d012c720 rp2: CMakeLists: Add components required by bootrom.h.
7dbe7f49e rp2: Sleep changes.
1a83ea4a9 tests/ports/rp2: Put back lightsleep/machine_idle test for RP2350.

These add a new board.

2921915d2 rp2: Add RPI_PICO2_W.
633f44e76 rp2: Add riscv variant for pico2.

The following are about supporting RM2 breakout boards.

82714913b rp2: Allow extra arguments in Makefile.
1dc54af2c rp2: Add mp_hal_is_pin_reserved.
bfc2b89bf rp2: Define cyw43 pins.
f13a6f40e rp2: Put cyw43 config in separate files.
ca42121f7 rp2: Allow pins.csv filename to be configured.
65a3f67a4 rp2: Allow pico and pico2 to use cyw43.
aad532f26 rp2: Dynamic cyw43 config.

dpgeorge:
@peterharperuk can you please rebase this now that #16313 is merged?

peterharperuk:
I re-ran the tests on all devices. extmod/machine_i2s_rate.py failed once on pico2/rm2 but I couldn't repeat this. rp2_dma is failing on pico2/rm2 as noted before (I guess I should look info fixing this) but not on pico2_w (bit weird)?

I see the thread_stdin issue on riscv as well but failures of select_poll_eintr, stress_schedule and thread_sleep2 when cyw43 is enabled. I think I missed these failures previously.

peterharperuk:
I've put the sleep changes over here https://github.com/micropython/micropython/pull/16454

ports/rp2/Makefile
```diff
@@ -34,26 +34,26 @@ ifeq ($(BUILD_VERBOSE),1)
 MAKE_ARGS += VERBOSE=1  # Picked up in Makefile generated by CMake
 endif
 
-CMAKE_ARGS += -DMICROPY_BOARD=$(BOARD) -DMICROPY_BOARD_DIR="$(abspath $(BOARD_DIR))"
+override CMAKE_ARGS += -DMICROPY_BOARD=$(BOARD) -DMICROPY_BOARD_DIR="$(abspath $(BOARD_DIR))"
```
dpgeorge, line 37: I really don't think it's a good idea to complicate the Makefile by adding `override`.  Combining make and cmake is already tricky enough.

Why exactly is this needed, what were you trying to achieve with it?  Can we do it a different way, eg just call cmake directly in such cases?

dpgeorge:
I'm glad that the Pico 2 W changes were pulled out to a separate commit (and merged!) because now this PR is a lot cleaner, and it's clearer what it's trying to do (allow dynamic cyw43 config).

And I now realize the complexity that's added here to the constructors of WLAN, BLE and Pin.  It wasn't clear to me before how deep these changes were, I thought they were only related to the WLAN part.

I really don't think the `machine.Pin` constructor should take cyw43 pin arguments (so the cyw43 can be set up to use the external GPIO).  That seems really convoluted and a violation of layering (pins needing pins).  I'm aware I suggested this, over the previous "set cyw43 pins" approach, but I didn't appreciate then that it would affect the `machine.Pin` constructor.

Taking a step back, what's needed here is a better abstraction/model of the hardware.  The cyw43 peripheral is not just WLAN, it's also BLE and GPIO.  So maybe it needs to be its own entity, and then you obtain the sub-entities from that.  Eg:
```py
import cyw43

# construct a cyw43 instance (this will persist over soft reset)
cyw = cyw43.CYW43(regon=2, cs=3, data=4, clock=5, div_int=3)

# get the sub-systems from the cyw43 instance
wlan = cyw.WLAN()
ble = cyw.BLE()
led = cyw.Pin(0)
```
The downside with this approach is that you can no longer use `network.WLAN()` or `bluetooth.BLE()`.  The former is not too bad, all the network tests will still work because they only access `socket`.  But the BLE tests won't work because they need `bluetooth.BLE()`.  Also, the user won't be able to access the LED (or other GPIO) via `machine.Pin`.  I don't think that's such a problem though, they know if the board has a cyw43 breakout or not and need to do things slightly differently constructing the objects.

Alternatively, when the `cyw43.CYW43` instance is created, it could activate the WLAN and BLE so that `network.WLAN()` and `bluetooth.BLE()` would "just work" and return the cyw43's instance of those peripherals.  In the case of multiple cyw43 breakout's (a future possibility) the WLAN and BLE entities could be specified by higher identifiers, eg `bluetooth.BLE(1)`.  Eg:
```py
import cyw43

# construct a cyw43 instance (this will persist over soft reset)
cyw = cyw43.CYW43(regon=2, cs=3, data=4, clock=5, div_int=3)

# access sub-systems as normal
wlan = network.WLAN()
ble = bluetooth.BLE()
led = machine.Pin("LED")
```
That's very similar to the "set cyw43 pins" approach, except that it's a cyw43 object that you create, rather than a function you call (to set the pins).  The benefit of this is that you can just put a `cyw43.CYW43()` line in boot.py/main.py somewhere, and then all existing code samples just work the same.

(Looking at how `network.WIZNET5K` works, that's much easier because that peripheral is just supplying LAN.  The CYW43 doesn't match this pattern so really needs something new, like suggested above.)

eightycc:
Over in CircuitPython we've run into an issue (https://github.com/adafruit/circuitpython/issues/9777) with the RM2 breakout that may need to be addressed in this issue. The RM2 (and probably the bare CYW43439) appears to automatically select phasing for RM2 to host transfers depending on the gSPI clock speed. Reducing the gSPI clock speed below a certain (unknown) threshold causes "normal mode" to be forced by the RM2. Writing a 1 to gSPI register 0x0000 bit 4 has no effect in this case.

peterharperuk:
@eightycc I've had to increase the pio divisor to 3 for reliable operation for reasons unknown. I don't know if that's the same issue.

eightycc:
@peterharperuk I'd suspect signal issues for your case. The problem I'm reporting kicks in around divide by 10 and up. I hit it while increasing the divisor to compensate for a nasty fly wire bodge I've been using for tracing.

eightycc:
Once I changed from `spi_gap01_sample0` to `spi_gap0_sample1` for the PIO program at slower clock speeds I've been able to go all the way down to 625kHz with no issues, so I don't believe there's a firmware issue.

eightycc:
@peterharperuk I've done some more testing with both the CYW43439 on the Pico W boards and the RM2 module.

The older CYW43439 operates as described in the Cypress Semi datasheet. It does not change gSPI data phasing at lower clock rates.

The newer RM2 modules change gSPI data phasing somewhere between 21.429MHz and 23.077MHz on the two RM2 modules I tested.

Assuming a 150MHz processor clock, a divisor of 3 results in a 25MHz gSPI clock. Depending on the accuracy of the RM2's gSPI clock measurement, a divisor of 3 could conceivably fail intermittently. At 23.077MHz (divisor 3, fraction 64) with a debug build I'm seeing instability.which leads me to think that there's a range of gSPI clock frequencies centered around 22MHz that are unusable.

Back to stable operation at 18.75MHz (divisor 4, fraction 0), but need to specify `CYW43_SPI_PROGRAM_NAME="spi_gap0_sample1"` to accommodate the change in phasing.

ports/rp2/Makefile
Gadgetoid: I think the intent is to pass compiler definitions through to a port, effectively skipping the need to change config files? Possibly to optionally build Pico with RM2 support.

We have variants for this, though there are few examples of that in practise. Here's one for switching to a RISCV build: https://github.com/micropython/micropython/tree/master/ports/rp2/boards/SEEED_XIAO_RP2350

Gadgetoid:
> The benefit of this is that you can just put a cyw43.CYW43() line in boot.py/main.py somewhere, and then all existing code samples just work the same.

This makes sense.

It feels like there should be a better approach to the pins.csv wrangling. It's a lot of duplication to accomplish not very much. My gut feeling would be to omit them, but perhaps something could be hung off the `CYW43` instance for manipulating these pins if they're needed.

Gadgetoid:
@peterharperuk this can be rebased to use the mainline custom pins.csv if/when you get time!

peterharperuk:
@ Gadgetoid Doesn't this need re-writiing to use cyw43.CYW43? I ran out of time which is why it hasn't been done. However I can rebase as it is now if it'll help.

Gadgetoid:
A straight rebase at least gets us a step closer, but yes it does need rewriting. I should probably try to pick that up, but am also time pressured 😆

peterharperuk:
@Gadgetoid Done - I just gave it a quick test with a Pico connected to a RM2 breakout

Gadgetoid:
Awesome, thank you. We're still "shipping" these changes in our custom builds, so it's useful for the time being.

ports/rp2/boards/RPI_PICO2/mpconfigboard.h
```diff
@@ -1,3 +1,22 @@
 // Board and hardware specific configuration
 #define MICROPY_HW_BOARD_NAME                   "Raspberry Pi Pico2"
 #define MICROPY_HW_FLASH_STORAGE_BYTES          (PICO_FLASH_SIZE_BYTES - 1024 * 1024)
+
+#if MICROPY_PY_NETWORK_CYW43
+#include "enable_cyw43.h"
+
+// Enable the ability to pass cyw43 pins into WiFi, Bluetooth and Pin constructors
```
iampremo, line 8: Will Pico W and Pico2 W also not need `CYW43_PIN_WL_DYNAMIC` and `CYW43_PIO_CLOCK_DIV_DYNAMIC` set to 1 so that their pins can be changed at runtime?

peterharperuk: How would you change the gpios the onboard cyw43 chip is using?

iampremo: I was thinking of it more as connecting up an external rm2 and not using the onboard one. I guess this probably isn't too common but for my case I have a surplus of pico2 w but need the rm2 in a different place than the pico

peterharperuk: You could just use the Pico 2 build on a Pico 2 W. The internal led won't work. Unfortunately having two cyw43 chips running at the same time is not currently supported.

iampremo: Ah I wasn't aware that you could do this. Tried it out over the weekend and it worked no problem. That makes it a lot easier for anyone who would want to use an external cyw43 chip on a W.

PR #16058, opened 2024-10-22
esp32: Move linker wrap directives to common cmake file.

projectgus:
### Summary

For in-tree builds, these are effectively equivalent. However for out-of-tree builds it's preferable to have as little as possible
in the top-level CMakeLists.txt file (as the out-of-tree build needs its own copy).

### Testing

Built the ESP32 port and used the linker map file to verify the `__wrap_`-prefixed symbols were still linked into the ELF file.


projectgus:
@karlp Noticed this while looking at #16013 last week. It may be relevant to your use case as well.

dpgeorge (approved):

dpgeorge:
Thanks!

PR #13011, opened 2023-11-18
examples/usercmodule: Add finaliser to cexample.

andrewleech:
There are a number of situations in user C modules where hardware or static C variables should be reset or de-init when the module is no longer being used, or particularly over a soft-reset.

This PR adds usage of the gc finaliser to the existing cexample module to demonstrate how this can be achieved.

Inspired by discussion in https://github.com/micropython/micropython/pull/12803 and https://ptb.discord.com/channels/574275045187125269/1012726673709469818/1173760875539206225

PR #16063, opened 2024-10-23
py/modatexit: Add atexit module for deinit functionality.

andrewleech:
### Summary

Edit: I pivoted this MR to instead implement atexit instead. Maybe it should have been a new PR.... Regardless I'll update this with more details later. 

Orig:

This PR proposes a pattern for adding a deinit function to any builtin module (including user c modules).

Similar to what I was trying to achieve in https://github.com/micropython/micropython/pull/13011 I've run into an issue safely developing a user c module such that it works correctly over soft-reset. In particular, this module needs to know when a soft-reset occurs so that it can re-initialise its global / MP_STATE_VM variables.

These builtin modules already have the ability to register an `__init__` function which gets run on import. This PR mirrors this with a `__del__` function which (if defined) will be run during soft reset.

I'm also inclined to think this mechanism could be used for a range of other builtin c modules which currently need to be manually listed in each ports main.c to deinit as needed. Using this automatic method could ease maintenance of ther ports.

Similarly, there's a lot of duplication in the soft-reset loop of most ports already; is there a reason the `gc_sweep_all();` and `mp_module_builtin_deinit();` shouldn't be included in the `mp_deinit()` function?

Lastly, should this mechanism possibly be extended to include all modules `MP_STATE_VM(mp_loaded_modules_dict)` to allow any module (inc python and dynamic c) to define a `__del__` function that's run on soft-reset? 

### Testing

I've been testing this with my work to make lvgl work cleanly as a user c module. If this mechanism seems reasonable, I'll also add this functionality to the c example module and try to figure some way to add a test to the coverage tests,

I initially tried to use the existing gc finaliser mechanisms demonstrated in https://github.com/micropython/micropython/pull/13011 however, similar to in the PR, I found it quite non-deterministic. It also required the use of a C class registered as a singleton which the module didn't otherwise have or need. I failed to find a way that ensured this singleton class would be actually deleted with the __del__ function run reliable during the soft restart, it just would not run for me.

I added this as a matching function to the existing builtin __init__ functionality, hoewver looking at cpython the atexit module is a more standard way to handle this kind of thing. I see there's already a sys.atexit micropython extension but it's only a single global function and it's only implemented on the windows port. Adding full atexit functionality that works with a linked list or similar would be more compatible, but would take more flash (to add a module / extend sys.atexit) and ram (holding the linked list) and run-time (to register the atexit functions).


py/objmodule.c
```diff
@@ -170,6 +170,21 @@ static const mp_module_delegation_entry_t mp_builtin_module_delegation_table[] =
 };
 #endif
 
+#if MICROPY_MODULE_BUILTIN_INIT
+void mp_module_builtin_deinit() {
```
dlech, line 174:
```suggestion
void mp_module_builtin_deinit(void) {
```

py/objmodule.c
```diff
@@ -170,6 +170,21 @@ static const mp_module_delegation_entry_t mp_builtin_module_delegation_table[] =
...
+    // Iterate through builtin modules. Run the __del__ function if defined.
+    for (size_t i = 0; i < mp_builtin_module_map.alloc; i++) {
+        if (mp_builtin_module_map.table[i].key != MP_OBJ_NULL) {
+            mp_obj_t dest[2];
+            mp_load_method_maybe(mp_builtin_module_map.table[i].value, MP_QSTR___del__, dest);
+            if (dest[0] != MP_OBJ_NULL) {
+                mp_call_method_n_kw(0, 0, dest);
```
dlech, line 181: What if this raises an exception?

dlech (commented):
This sounds more like `atexit()` than `__del__`. Could we implement it that way instead?

Once nice thing about `atexit()` is that the call order is deterministic, which could be important in some cases.

andrewleech:
> This sounds more like atexit() than __del__. Could we implement it that way instead?

Yeah as I was initially writing up the trade-offs/alternatives section above I was thinking an atexit approach probably would be much better.

I've been thinking about it more, my feeling is it would be quite low overhead to simply extend the existing system.atexit function to store all provided functions in a linked list, rather than being the single slot that gets replaced currently. Then enable that by rom level and integrate into all ports. 

If anyone needs it as a separate atexit module for compatibility that could be done as a wrapper in micropython-lib 

I'm not sure that there's really any great need to also implement the remove function, just registering functions is probably enough to start with

py/modsys.c
```diff
@@ -208,13 +208,42 @@ static MP_DEFINE_CONST_FUN_OBJ_1(mp_sys_getsizeof_obj, mp_sys_getsizeof);
 #endif
 
 #if MICROPY_PY_SYS_ATEXIT
+typedef struct _m_atexit_node_t {
+    struct _m_atexit_node_t *prev;
+    struct _m_atexit_node_t *next;
+    mp_obj_t fn;
+} m_atexit_node_t;
+
 // atexit(callback): Callback is called when sys.exit is called.
```
stinos, line 217: Can you add a comment that this is LIFO ?

py/modsys.c
```diff
@@ -208,13 +208,42 @@ static MP_DEFINE_CONST_FUN_OBJ_1(mp_sys_getsizeof_obj, mp_sys_getsizeof);
...
+void mp_sys_atexit_execute(void) {
+    // walk through the linked list and execute each function
+    // Beware, the sys.settrace callback should be disabled before running sys.atexit.
+    while (MP_STATE_VM(sys_exitfunc) != mp_const_none) {
+        if (mp_obj_is_callable(MP_STATE_VM(sys_exitfunc)->fn)) {
+            nlr_buf_t nlr;
+            if (nlr_push(&nlr) == 0) {
+                mp_call_function_0(MP_STATE_VM(sys_exitfunc)->fn);
+            } else {
+                // Uncaught exception: print it out.
```
stinos, line 240: In CPython a SystemExit would still cause an actual exit, is it worth adding that?

andrewleech: These are currently only being run during an exit / soft-restart so I wouldn't think the handling should change? 
Though maybe the exception _shouldn't_ be printed out if it's a `SystemExit`

ports/unix/main.c
```diff
@@ -753,10 +754,7 @@ MP_NOINLINE int main_(int argc, char **argv) {
     #endif
 
     #if MICROPY_PY_SYS_ATEXIT
-    // Beware, the sys.settrace callback should be disabled before running sys.atexit.
```
andrewleech, line 756: This comment was added in https://github.com/micropython/micropython/commit/310b3d1b81d561e19d719acd89ee47b759e3795c

Does anyone know why this would be the case and if it's still relevant?

py/modsys.c
stinos: > so I wouldn't think the handling should change?

It is actually changing: for the unix port an exception called during the atexit callback would result in exit code 1 (because of an 'FATAL: uncaught NLR', maybe not ideal, but at least it's a non-zero exit code and that is what matters). But with this implementation the exception would be swallowed (and printed). That should not change imo and this should remain CPython-compatible in the sense that an exception during atexit is treated as an exception. Note CPython does [this](https://docs.python.org/3/library/atexit.html):

> If an exception is raised during execution of the exit handlers, a traceback is printed (unless SystemExit is raised) and the exception information is saved. After all exit handlers have had a chance to run, the last exception to be raised is re-raised.

But given your question, I assume raising an exception here would actually change behavior?


>  Though maybe the exception shouldn't be printed out if it's a SystemExit

That's indeed what CPython does.

andrewleech: Ah, that detail helps thanks! The filter on whether to print the exception is easy enough to implement certainly. Maybe then just overriding the exit code in the exception handler here is enough to broadly match cpython, as opposed to keeping a copy of the exception details to pass them onto any default handler.

andrewleech:
I'm going to try reworking to the a cpython compatibly `atexit` module and check the size difference, see how much overhead this adds at least. It'd be preferable to actually match cpython api rather than this sys extension.

py/modsys.c
stinos: > Maybe then just overriding the exit code in the exception handler here is enough to broadly match cpython

Probably, but in the end it would be nicer to fix this properly:


> as opposed to keeping a copy of the exception details to pass them onto any default handler.

Ideally the only thing which would be needed is to store `nlr.ret_val` then if not NULL raise it again after having called all handlers. Should not be an issue code-size-wise. Passing it onto the default handler would then happen automatically because the main code knows how to do that; however, this requires that `mp_sys_atexit_execute` is being called within `execute_from_lexer` (or whatever the counterpart is named for the various ports).

This is not the case for the unix port now, nor for the other ports. I don't know why though, and all in all this is maybe not very logical/expected; e.g. for the stm32 port it runs after deinitializing pretty much all of the hardware: is there a specific reason you did it that way? Would it not be more sound if there's not many restrictions to what can run in the atexit handler i.e. can still use hardware if it chooses to etc?

andrewleech:
Try `raise SystemExit` at the end of the unit test. 
Capture the mingw ci script "print-failures job" in separate issue, why it sometimes re-runs tests, other times traceback
Look at 
```
Run python run-tests.py --print-failures
--- "D:/a/micropython/micropython/tests/results\\misc_atexit.py.exp"	2025-03-24 15:45:14.18[6](https://github.com/micropython/micropython/actions/runs/14039499915/job/39305801012?pr=16063#step:11:7)485700 +0000
+++ "D:/a/micropython/micropython/tests/results\\misc_atexit.py.out"	2025-03-24 15:45:14.186485[7](https://github.com/micropython/micropython/actions/runs/14039499915/job/39305801012?pr=16063#step:11:8)00 +0000
@@ -1,3 +1,5 @@
-printed before exit
-atexit: registered last with access to local var
-atexit: registered first
+  File "D:/a/micropython/micropython/tests/misc/atexit.py", line 4, in <module>
+  File "D:/a/micropython/micropython/tests/misc/atexit.py", line [12](https://github.com/micropython/micropython/actions/runs/14039499915/job/39305801012?pr=16063#step:11:13), in <module>
+AttributeError: 'module' object has no attribute 'register'
+CRASH
\ No newline at end of file

FAILURE D:/a/micropython/micropython/tests/results\misc_atexit.py
```

shared/runtime/pyexec.c
```diff
@@ -175,6 +189,11 @@ static int parse_compile_execute(const void *source, mp_parse_input_kind_t input
         mp_hal_stdout_tx_strn("\x04", 1);
     }
 
+    #if MICROPY_PY_ATEXIT
+    int atexit_code = mp_atexit_execute();
```
projectgus, line 193: I haven't tested this, but placing this here in `parse_compile_execute()` seems like it might have some unexpected consequences? For example, if you put an `atexit.register()` call in `boot.py` then will the handler be run after executing `boot.py` and before executing `main.py`?

(I think this draws attention to a shortcoming in the current architecture, which is that AFAIK there's no non-hardware-specific soft reset hook function that all ports can call for things which need cleanup in the soft reset path.)

projectgus (commented):
This implementation looks generally good to me, nice and clean. And potentially very useful!

I have a couple of inline questions. Also I'm a little surprised at the amount of code size impact on some ports given how little code seems to be compiled in (assuming register-only). But maybe that's actually inline with expectations.

ports/unix/main.c
```diff
@@ -756,6 +757,11 @@ MP_NOINLINE int main_(int argc, char **argv) {
         }
     }
 
+    #if MICROPY_PY_ATEXIT
+    int atexit_code = mp_atexit_execute();
+    ret = (atexit_code != 0) ? atexit_code : ret;
```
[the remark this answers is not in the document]
andrewleech: That's an interesting point, yeah I assumed we should capture any failures here but for code size and compatibility it's worth trying without

shared/runtime/pyexec.c
andrewleech: Yes I've realised similar issues with this location and been testing / looking at other ways to add this cleanly.

I've actually been fighting the urge to undertake a broad refactor & consolidation of soft-reset / shutdown handling in ports to create the "correct" place to run the handler code! I'd rather not block this change behind such a large refactor though. 

Perhaps for now I should go back to my earlier plan to just add the execute call to an appropriate place in each main.c 
Then they'll all get cleaned up anyway if I get to the refactor.

projectgus: > Perhaps for now I should go back to my earlier plan to just add the execute call to an appropriate place in each main.c Then they'll all get cleaned up anyway if I get to the refactor.

That sounds like a good approach to me. FWIW, I have also felt this urge... but better to do it as a pure refactor, for sure.

andrewleech:
@projectgus / @dpgeorge I'm also running into an issue with trying to run the unit test on qemu port, which also applies to running it on any live hardware that connects with pyboard.py behind the scenes;

Basically pyboard.py uses the raw mode and paste to drop the test code onto the target and runs it that way, capturing just the output from the code run in this raw mode session. 

I recently added the `raise SystemExit` at the end of the test to ensure a soft-reboot is triggered, which is working, however the first thing done on the target as part of this soft reset is exiting raw mode which ends the test result capture  for run_tests.py so it never receives the print outputs from the atexit functions.

Is there any other tests I can reference that connect over a soft-restart / keeps the raw connection open to capture all the output during this time?

Carglglz:
> I've been testing this with my work to make lvgl work cleanly as a user c module.

@andrewleech You may be already aware of this option but using `MICROPY_PORT_DEINIT_FUNC` works, see https://github.com/lvgl/lv_binding_micropython/pull/383/files, 

I still think this PR is a better solution in the long term, so looking forward to it 👍🏼

andrewleech:
> @andrewleech You may be already aware of this option but using `MICROPY_PORT_DEINIT_FUNC` works, see https://github.com/lvgl/lv_binding_micropython/pull/383/files,
> 
> I still think this PR is a better solution in the long term, so looking forward to it 👍🏼

Ah neat, yeah that's not bad as a workaround, better than having the C module crash over soft-reboot! It wont work for native modules though, or if there's multiple modules that need it.

Yeah I'm still working on this one intermittently, adding support to the unit test runner for tests that can continue over a number of  device soft-resets (or hard reset on devices with uart / persistent serial connection) to capture all the behaviour during the reset process.

PR #16068, opened 2024-10-24
rp2/datetime_patch: Fix year and month offset for mktime wrapper.

dpgeorge:
### Summary

Retrieving the RTC time on RP2040 is currently broken.  This is due to commit 8cc7c64d01a597da8190873432fcb6b267e6a1bb , which does not convert the year and month when passing it to `timeutils_mktime()`.  That is now fixed.

The root cause is that MicroPython's custom timeutils functions have year and month as 1-based, which differs to the standard `mktime()`.  So this is now documented explicitly.

Also, a simple test is added for `machine.RTC` which failed on RP2040 targets prior to the fix here.

### Testing

`mpremote rtc --set` and `mpremote rtc` now work as expected on Pico.

Pico W now passes the multi_net tests (prior to this fix it would fail the time checks for the TLS certificates).

Pico and Pico W now pass the new `machine_rtc.py` test added here.


projectgus (approved):
Oops, entirely my mistake - I had totally missed those distinctions! Fix and regression test look good to me.

dpgeorge:
> I had totally missed those distinctions!

Well, they have very similar names (the `mktime` bit of the name), so it is a bit confusing.

Conventions for time are always tricky :smile:

PR #16069, opened 2024-10-24
Update README.md

holdenweb:
Remove redundant `global` statement from example code. Since `led` is not being rebound inside the `tick` function the standard Python name resolution method will find it.

### Summary

A simple improvement in sample code Pythonicity.

### Testing

I felt this documentation change was unlikely to modify any test results or cause errors in the documentation toolchain.


dpgeorge (approved):

dpgeorge:
Thanks for the patch!

PR #16071, opened 2024-10-24
py/malloc,rp2: Add mutex for tracked allocations.

projectgus:
~~Note: This draft PR is currently one commit stacked on top of #16015, will need rebase after the linked PR is merged.~~ (Done.)

### Summary

As discussed in [this comment](https://github.com/micropython/micropython/pull/16015#discussion_r1802309307), non-GIL threaded ports (i.e. rp2) need some thread safety protection around tracked allocations. This PR reuses the GC mutex for this purpose.

### Testing

* This PR adds a unit test for thread safety with TLS sockets. Running the original version of #16015 this test would HardFault on rp2 quite regularly. The final version of #16015 doesn't use tracked node allocations as aggressively and I was unable to make it fail (the race condition window is quite small), but I've kept the test in for additional coverage.
* Also ran full test suite on rp2, unix coverage variant, and esp32 port. No failures and new test passes.

* This fix is based on an assumption that threaded ports with a GIL will always hold the GIL when calling tracked allocation functions. I believe this is true in all current ports using existing configurations, although will need to take care in the future (for example, the openamp module libmetal integration uses tracked allocations - it's currently not enabled on any threaded ports, but if it was then this might be unsafe.)
* Initial version of this fix added a new mutex for tracked alloc linked list only, instead of reusing the GC mutex. This has all the downsides of reusing the GC mutex (potential for deadlocks), plus the downside of requiring a `m_tracked_malloc_init` function to initialise the mutex.


projectgus:
Rebased, re-tested. The bug this fixes is no longer easy to reproduce, but I believe it still exists.

tests/extmod/ssl_threads.py
```diff
@@ -0,0 +1,58 @@
+# Ensure that SSL sockets can be allocated from multiple
```
dpgeorge, line 1: Since this test uses the `tls` module (and not `ssl`) I suggest calling it `tls_threads.py`.

tests/extmod/ssl_threads.py
```diff
@@ -0,0 +1,58 @@
...
+# threads without thread safety issues
+import gc, unittest
```
dpgeorge, line 3: It looks like `gc` is unused.  Also `sys` below is unused.

dpgeorge (commented):
I tested this on RPI_PICO2_W, running the new test both before and after the patch in this commit.  It passes.  Also tested the test on ESP32_GENERIC, and it passes there.

I agree that even if the test passes without the patch here, it's still worth having this fix merged.  It should be very minimal (unmeasurable) in terms of cost to performance.

projectgus:
Rebased, addressed comments, re-ran the new unit test on RPI_PICO_W (still passes).

dpgeorge (approved):

PR #16072, opened 2024-10-24
tests/extmod: Fix access of RTC class in machine.RTC test.

dpgeorge:
This previously passed on some targets that automatically import the `machine` module in `boot.py`.

Tested using `./run-tests.py --target pyboard --via-mpy` which forces a clean namespace, and fails without this patch.


dpgeorge:
Mistake introduced in #16068.

projectgus (approved):

PR #16073, opened 2024-10-24
rp2/modmachine: Re-sync time_ns offset when coming out of lightsleep.

dpgeorge:
### Summary

Prior to this fix, `tests/extmod/vfs_lfs_mtime.py` would fail when run after the `tests/ports/rp2/rp2_lightsleep.py` test, because `time.time_ns()` would have a large and constant offset from `time.time()`.

Fix this by re-syncing the time-ns offset when coming out of lightsleep.

### Testing

Running the test suite multiple times on a Pico, it now passes the `vfs_lfs_mtime.py` test.  Prior to this fix, it would fail it on the second (and later) test run.

### DIFF
--- a/tests/extmod/time_time_ns.py
+++ b/tests/extmod/time_time_ns.py
@@ -22,3 +22,10 @@ if 2000000 < t1 - t0 < 50000000:
     print(True)
 else:
     print(t0, t1, t1 - t0)
+
+# Check that time.time() and time.time_ns() are within a second of each other.
+# Note that time.time() may return an int or float.
+for _ in range(10):
+    t_s, t_ns = time.time(), time.time_ns()
+    print(abs(t_s * 1_000 - t_ns // 1_000_000) <= 1_000)
+    time.sleep_us(100_000)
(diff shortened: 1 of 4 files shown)


projectgus (approved):

dpgeorge:
This still doesn't fix the vfs_lfs_mtime test fully (it passes sometimes but not other times).

So I tweaked that test, and added a proper test to test the difference between `time_ns()` and `time()`.

tests/extmod/time_time_ns.py
```diff
@@ -22,3 +22,9 @@
     print(True)
 else:
     print(t0, t1, t1 - t0)
+
+# Check that time.time() and time.time_ns() are within a second of each other.
+for _ in range(10):
+    t_s, t_ns = time.time(), time.time_ns()
```
projectgus, line 28: are there any ports where `time.time()` result is an int not a float?

projectgus:
@dpgeorge The new checks still look good to me, just one question about the possible types of time.time() results

tests/extmod/time_time_ns.py
dpgeorge: Yes, the unix and webassembly ports.  Will fix.

dpgeorge: Actually, I think it's fine as-is, regardless of whether `time()` returns a float or int.  If it's a float then multipliying by 1_000_000_000 yields a float, then the subtraction is a float and so is the comparison.  And it passes.

But that can loose precision going to float (double).  I have tweaked the test a bit.  I don't want to rely on the port having float at all, but it should work in all cases.  See what you think.

PR #16110, opened 2024-10-29
Fix/mip documentation

honza-klu:
### Summary
There is small mistake in [Package management](https://docs.micropython.org/en/latest/reference/packages.html) documentation. It is stating that it can be used on Unix port like this:
` mpremote mip install --target=/flash/third-party pkgname`
But it has to be used like this:
` mpremote mip install --target /flash/third-party pkgname`
This error is probably caused by mip usage directly in python code (`mip.install("pkgname", target="third-party")`)


dpgeorge (approved):
Thank you for the fix.  This looks correct to me, and I did some simple tests to verify it.

PR #15547, opened 2024-07-25
lib/mbedtls: Update to mbedtls v3.6.2.

gstrauss:
### Summary

lib/mbedtls: Update to mbedtls v3.6.2 for security fixes

PR #12658, opened 2023-10-10
ci/ci_commit_formatting_run: Problem with checking older MR branch.

andrewleech:
I've noticed on a couple of MR's lately that the CI commit message check is failing, even when the MR commit looks fine.

One example of this is https://github.com/micropython/micropython/actions/runs/6059530353/job/16442749613?pr=12126
This an MR with 4 commits, however the CI check is looking at a _long_ list of different commits:
image

I made this dummy MR branched from the same point as #12126  to test / examine the issue, though in testing locally I'm pretty sure I've found the problem. Assuming this initial push fails the same way, I'll push a follow up commit to check the solution.

This MR, like many other active MR's, is branched from an old enough point in master that when `ci_commit_formatting_run` runs `git fetch --depth=100 upstream master` it doesn't actually grab enough commits from upstream to find the common ancestor between the MR branch and master. 

It ends up checking _only_ master... which includes some older commits that did not pass the check.

PR #16114, opened 2024-10-30
tools/ci: Fix commit message checking when PR branch HEAD behind master.

projectgus:
### Summary

Fixes the subtle problem [noted here](https://github.com/micropython/micropython/pull/15547#issuecomment-2434479702) - [workflow log here](https://github.com/micropython/micropython/actions/runs/11492562583/job/31986832560?pr=15547). Specifically, default CI HEAD for a PR is a (generated) merge commit into the master branch's current HEAD. If the PR branch isn't fully rebased (i.e. is one or more commits behind latest mater) then the commit check runs against commits from master as well!

Also drops running this check on push, the pull_request event is triggered by default on open and update, so that should cover all cases where this check needs to run.

### Testing

I made this PR branch the default branch in my fork and opened a series of dummy PRs:

1. Up to date with a valid commit message. Passes on master, [still passes](https://github.com/projectgus/micropython/actions/runs/11586370067/job/32256877702?pr=7).
2. Up to date with an invalid commit message. Fails on master, [still fails](https://github.com/projectgus/micropython/actions/runs/11586368061/job/32256874148?pr=8).
3. Old branch (100+ commits out of date) with invalid message. Fails on master but probably tests a lot of unrelated commits. [Tests correct commits and fails with this PR](https://github.com/projectgus/micropython/actions/runs/11587208838/job/32259052443). Note this run needed to do an "unshallow" fetch to find the merge base, due to being more than 100 commits out of date.
4. Old branch (one commit out of date) with valid message. This is the kind of branch which fails on master as it tests unrelated commits. [Checks one commit and passes with this PR](https://github.com/projectgus/micropython/actions/runs/11587538275/job/32259879338?pr=10).

### DIFF
--- a/.github/workflows/commit_formatting.yml
+++ b/.github/workflows/commit_formatting.yml
@@ -1,6 +1,6 @@
 name: Check commit message formatting
 
-on: [push, pull_request]
+on: [pull_request]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
@@ -12,7 +12,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
-        fetch-depth: '100'
+        fetch-depth: 100
     - uses: actions/setup-python@v5
     - name: Check commit message formatting
       run: source tools/ci.sh && ci_commit_formatting_run
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -39,12 +39,18 @@ function ci_c_code_formatting_run {
 # commit formatting
 
 function ci_commit_formatting_run {
-    git remote add upstream https://github.com/micropython/micropython.git
-    git fetch --depth=100 upstream master
+    # Default GitHub Actions checkout for a PR is a generated merge commit where
+    # the parents are the head of base branch (i.e. master) and the head of the
+    # PR branch, respectively. Use these parents to find the merge-base (i.e.
+    # where the PR branch head was branched)
+
     # If the common ancestor commit hasn't been found, fetch more.
-    git merge-base upstream/master HEAD || git fetch --unshallow upstream master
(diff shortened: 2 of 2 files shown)


projectgus:
@andrewleech I just saw you tackled this exact problem in https://github.com/micropython/micropython/pull/12658, but I think there are still some scenarios where it still checks the wrong commits.

dpgeorge:
Thanks, this looks good to me, and your testing is thorough!

As you noted, do we also need to update the related logic in `ci_code_size_build`?

andrewleech:
That looks great @projectgus , thanks for the clear description both here and in the script. 
I'm pretty sure I never realised GitHub ran its actions in a trial merge, that does complicate it just a little bit more!

andrewleech (approved):

projectgus:
> I'm pretty sure I never realised GitHub ran its actions in a trial merge, that does complicate it just a little bit more!

Yeah, I hadn't realised either until I looked at this. There is a way to [tell the GitHub checkout action to check out the PR branch HEAD instead of the merge commit](https://github.com/actions/checkout?tab=readme-ov-file#Checkout-pull-request-HEAD-commit-instead-of-merge-commit), but this actually gives us more work in this case because we still don't know what the merge base is! Whereas having the dummy merge commit (in this case) means we have all of the git history we need without needing to fetch anything new.

> As you noted, do we also need to update the related logic in ci_code_size_build?

I looked at this yesterday and thought it was OK, but I just looked again and I think there's a bug: it currently compares the PR branch's merge base against the generated merge commit. So that means any changes merged on master which aren't in the PR branch are included in the size comparison.

I think it should either compare the merge base with the PR branch HEAD, or compare the generated merge commit with the master branch. Probably the second one as that's exactly what the code size increase would be *if* the PR was merged.

Will submit another PR for this.

PR #16120, opened 2024-10-31
extmod/nimble: Remove asserts of ediv_rand_present and adjust comments.

dpgeorge:
### Summary

Recent versions of NimBLE (since release 1.6.0) removed this variable; see https://github.com/apache/mynewt-nimble/commit/7cc8c08d67d52b373eeeec6b33b113192cf2e996.

We never used it except in an assert, so remove those asserts to make the code compatible with newer NimBLE versions (eg for the esp32 port).

### Testing

Not tested on hardware, but I verified that the PYBD-SF6 firmware (which uses NimBLE) is binary equivalent before and after this commit (except for the change in git hash stored in the firmware).

### DIFF
--- a/extmod/nimble/modbluetooth_nimble.c
+++ b/extmod/nimble/modbluetooth_nimble.c
@@ -1911,24 +1911,21 @@ static int ble_secret_store_read(int obj_type, const union ble_store_key *key, u
                 // <type=peer,addr,*> (single)
                 // Find the entry for this specific peer.
                 assert(key->sec.idx == 0);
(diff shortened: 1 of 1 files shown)


dpgeorge:
@DvdGiessen this follows pretty much your patch.

projectgus (approved):
LGTM, I *think* I understand the linked NimBLE commit well enough to say that. :grin:

dpgeorge:
> I _think_ I understand the linked NimBLE commit well enough to say that.

I also had the same thought :smile:

PR #16121, opened 2024-10-31
extmod/modlwip: Fix IGMP address type when IPv6 is enabled.

### LINKED ISSUE #16100: Multicast Group Join Fails with ValueError from setsockopt() on v1.24.0

rp2 port, Raspberry Pi PICO W

### MicroPython version

MicroPython v1.24.0 on 2024-10-25; Raspberry Pi Pico W with RP2040

### Reproduction

```
import network
import time
import socket
import struct
import sys

ssid = None
password = None

MCAST_GROUP = '239.3.3.3'
MCAST_PORT = 50007

def inet_aton(ip_str):
    return bytes(map(int, ip_str.split('.')))

wlan.active(True)
if ssid:
    wlan.connect(ssid, password)
    while not wlan.isconnected():
        time.sleep(0.25)

sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)

mreq = struct.pack("4sL", inet_aton(MCAST_GROUP), 0) 
print ('mreq:', mreq)

try:
    sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, mreq)
    print("success")
except Exception as e:
    sys.print_exception(e)
    sys.exit()

while True:
    try:
        print('recv:', sock.recv(1024))
    except Exception as e:
        sys.print_exception(e)
        sys.exit()
```

### Expected behaviour

mreq: b'\xef\x03\x03\x03\x00\x00\x00\x00'
success

### Observed behaviour

mreq: b'\xef\x03\x03\x03\x00\x00\x00\x00'
  File "\<stdin>", line 30, in \<module>
ValueError: 

### Additional Information

This works on v1.23.0 provided wlan is active.

### Code of Conduct

Yes, I agree

dpgeorge:
### Summary

This was missed in 628abf8f25a7705a2810fffe2ca6ae652c532896

Fixes issue #16100.

### Testing

Tested with PYBD-SF2 and Pico W.  Multicast join works on v1.23.0 but not v1.24.0, but works again with this patch.

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -1432,7 +1432,7 @@ static mp_obj_t lwip_socket_setsockopt(size_t n_args, const mp_obj_t *args) {
         case IP_DROP_MEMBERSHIP: {
             mp_buffer_info_t bufinfo;
             mp_get_buffer_raise(args[3], &bufinfo, MP_BUFFER_READ);
-            if (bufinfo.len != sizeof(ip_addr_t) * 2) {
+            if (bufinfo.len != sizeof(MP_IGMP_IP_ADDR_TYPE) * 2) {
                 mp_raise_ValueError(NULL);
             }
 


dpgeorge:
Note: the bug was that, when IPv6 is enabled, the `sizeof(ip_addr_t)` is much larger than IPv4 size, which is what's needed for IGMP addressing.

projectgus (approved):

PR #16127, opened 2024-11-01
esp32: Fix machine.PWM on IDF <5.2.

projectgus:
### Summary

Restore PWM support for ESP-IDF v5.0.x and v5.1.x. The cleanup in #16090 relies on some functions not available in older ESP-IDF. Temporarily restore support for a possible V1.24.1 bugfix release, before we actually drop support for ESP-IDF <5.2.

(Edit: Dropped the other commit which was in this PR as it wasn't necessary.)

### Testing

* Tried briefly to make a unit test with `machine.time_pulse_us()` as suggested by @dpgeorge [here](https://github.com/micropython/micropython/pull/16090#issuecomment-2445811920). Unfortunately it looks like readback of the same pin that's outputting PWM doesn't currently work - need to either fix that or require a jumper between two pins to run the test.
* Instead, wired up the logic analyser and measured frequencies between 50Hz and 16.667Mhz with some different duty cycles on ESP32, ESP32-C3 and ESP32-S3 and ESP-IDF versions 5.2.2, 5.1 and 5.0.4.

* We could drop support for ESP-IDF <5.2 immediately instead of reinstating support here, but that's not very reasonable if there's going to be a V1.24.1 bugfix release.

### DIFF
--- a/ports/esp32/machine_pwm.c
+++ b/ports/esp32/machine_pwm.c
@@ -34,8 +34,11 @@
 #include "py/mphal.h"
 #include "driver/ledc.h"
 #include "esp_err.h"
-#include "esp_clk_tree.h"
 #include "soc/gpio_sig_map.h"
+
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 1, 0)
+#include "esp_clk_tree.h"
+#endif
 
 #define PWM_DBG(...)
 // #define PWM_DBG(...) mp_printf(&mp_plat_print, __VA_ARGS__); mp_printf(&mp_plat_print, "\n");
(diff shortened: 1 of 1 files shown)


IhorNehrutsa:
Features from this PR are implemented in #10854

ports/esp32/machine_pwm.c
```diff
@@ -208,7 +211,42 @@ static void configure_channel(machine_pwm_obj_t *self) {
...
+        res = HIGHEST_PWM_RES;
+    }
+
+    return res;
+}
+#endif
+
+
 static void set_freq(machine_pwm_obj_t *self, unsigned int freq, ledc_timer_config_t *timer) {
+    esp_err_t err;
```
dpgeorge, line 249: Any reason this was moved here, instead of just keeping it as it was, defined at the place of use below?

dpgeorge:
> tried briefly to make a unit test with `machine.time_pulse_us()` as suggested by @dpgeorge [here](https://github.com/micropython/micropython/pull/16090#issuecomment-2445811920). Unfortunately it looks like readback of the same pin that's outputting PWM doesn't currently work

Oh, that's a shame!  It would have been great to be able to test PWM on all ports without any hardware connections.  But I guess it's too much to ask of all ports to support reading a pin in PWM mode.

In that case we could do it with a single jumper wire, and still use `time_pulse_us()`.  Would be good to use the same two pins with a jumper that the UART tests use.

Anyway, that can be done separately to this PR.

dpgeorge:
> We could drop support for ESP-IDF <5.2 immediately instead of reinstating support here, but that's not very reasonable if there's going to be a V1.24.1 bugfix release.

I think we should fix this for a bugfix release.

dpgeorge:
See #16147 for a test for PWM.

ports/esp32/machine_pwm.c
projectgus: Yes, the first place of use is now inside an #ifdef block.

projectgus:
> Features from this PR are implemented in #10854

I appreciate you've been working steadily at that PR for a while without a lot feedback, however I don't quite understand what you mean. AFAIK this PR doesn't add any features, all the new code here is copy-paste of code that was removed in #16090 (but now with #ifdefs around it for older IDF).

What features do you mean?

ports/esp32/machine_pwm.c
dpgeorge: Oh, I see, I missed that it's also used for `ledc_timer_config()`.

I would have put this within the `if (freq != timer->freq_hz)` block to keep its scope limited, but here is also fine considering it's a generic error return variable.

IhorNehrutsa:
> What features do you mean?

I mean that **ledc_find_suitable_duty_resolution** is used like in this PR.

projectgus:
> > What features do you mean?
> 
> I mean that **ledc_find_suitable_duty_resolution** is used like in this PR.

Oh, I see. This function was already added in #16090 actually, but I see that if we'd merged #10854 earlier then we wouldn't have needed that one. `ledc_find_suitable_duty_resolution` function is not present in IDF<5.1 which is why this PR adds back the old logic temporarily. However after this PR (which is for v1.24.1) we can merge #16128 (for v1.25) and only support IDF>=5.2 which will make things simpler.

dpgeorge (approved):

IhorNehrutsa:
duty_resolution must be in `0 <= timer->duty_resolution <= HIGHEST_PWM_RES`

PR #16128, opened 2024-11-01
esp32: Remove support for ESP-IDF <5.2

projectgus:
### Summary

When we merged support for ESP-IDF V5.2.2 in #15523, we discussed dropping support for older versions after the new version was bedded in. ([Related Discord discussion here](https://discord.com/channels/574275045187125269/952412628926472273/1265452441932136550)). This is a draft for that PR!

In two commits, one to remove all the #ifdef-ed code and one that reverts 27279e69b4 from #13775 (adding a 5.2-specific sdkconfig variable).

### Testing

* Haven't done a lot of on-device testing, but built ports for ESP32_GENERIC, ESP32_GENERIC_C3, ESP32_GENERIC_S3, ESP32_GENERIC_C6 and verified the binary size was the same for both master and this branch (ESP-IDF V5.2.2).

* Was originally going to keep the `SDKCONFIG_IDF_VERSION_SPECIFIC` variable from #13775 in case we need it in the future, but I reckon it's better to revert it here and we can always un-revert it later if a need arises for it.


projectgus:
Sorry andrew and damien, I thought I'd cancelled the review request before opening (as realised it's still a draft).

dpgeorge:
You (and others) did a lot of testing of IDF 5.2.2 in #15523, so I'm happy to move forward with this as the minimum supported version.  Also, firmware size was reduced a lot, so that means users won't struggle to upgrade if they are short on flash.

Since we will make a v1.24.1 (and possibly further patch releases on that v1.24 branch), we can have v1.24 as the last MicroPython version to support IDF < 5.2.2, and users can stick to that release branch if necessary.

ports/esp32/boards/sdkconfig.base
```diff
@@ -50,6 +50,8 @@ CONFIG_ESP_SYSTEM_MEMPROT_FEATURE=n
 CONFIG_FREERTOS_THREAD_LOCAL_STORAGE_POINTERS=2
 CONFIG_FREERTOS_SUPPORT_STATIC_ALLOCATION=y
 CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP=y
+CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP=n
```
projectgus, line 53: Oops!

ports/esp32/README.md
```diff
@@ -28,7 +28,7 @@ manage the ESP32 microcontroller, as well as a way to manage the required
 build environment and toolchains needed to build the firmware.
 
 The ESP-IDF changes quickly and MicroPython only supports certain versions.
-Currently MicroPython supports v5.0.4, v5.0.5, v5.1.2, v5.2.0, v5.2.2.
+Currently MicroPython supports v5.2.0 and v5.2.2.
```
dpgeorge, line 31: The tag is actually `v5.2`, see #16195.

projectgus:
Rebased now the linked PRs are merged, this PR now includes a revert of #16127. Re-ran unit tests on S3 and verified Wi-Fi is working on C3.

ports/esp32/main_esp32/idf_component.yml
```diff
@@ -2,4 +2,4 @@
 dependencies:
   espressif/mdns: "~1.1.0"
   idf:
-    version: ">=5.0.4"
+    version: ">=5.2.0"
```
dpgeorge, line 5: Is this correct even though the IDF tag is `v5.2`?

dpgeorge: I'm pretty sure it's correct, just wanted to double check.

ports/esp32/mpthreadport.c
```diff
@@ -41,7 +41,7 @@
 #define MP_THREAD_DEFAULT_STACK_SIZE                    (MP_THREAD_MIN_STACK_SIZE + MICROPY_STACK_CHECK_MARGIN)
 #define MP_THREAD_PRIORITY                              (ESP_TASK_PRIO_MIN + 1)
 
-#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 2, 0) && !CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP
+#if !CONFIG_FREERTOS_ENABLE_STATIC_TASK_CLEAN_UP
```
dpgeorge, line 44: Is this if logic needed at all now that this option is always disabled?

ports/esp32/network_wlan.c
```diff
@@ -760,10 +756,8 @@ static MP_DEFINE_CONST_DICT(wlan_if_locals_dict, wlan_if_locals_dict_table);
 
 #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 2, 0)
 _Static_assert(WIFI_AUTH_MAX == 13, "Synchronize WIFI_AUTH_XXX constants with the ESP-IDF. Look at esp-idf/components/esp_wifi/include/esp_wifi_types.h");
-#elif ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 5) && ESP_IDF_VERSION < ESP_IDF_VERSION_VAL(5, 1, 0) || ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 1, 2)
```
dpgeorge, line 763: I think there's some if logic above for the new SEC constants that can be removed.

ports/esp32/main_esp32/idf_component.yml
projectgus: It is, these are SemVer rather than git tags: https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/versioning.html#version-precedence

ports/esp32/mpthreadport.c
projectgus: Oh yeah, have pulled all this out!

ports/esp32/network_wlan.c
projectgus: Thanks, have removed.

projectgus:
Have rebased onto the v5.3 changes, ran one lot of unit tests on C6 and also re-ran the before/after binary size comparison on the boards mentioned in the PR description.

ports/esp32/network_wlan.c
projectgus: Thanks, should be all cleaned out now.

dpgeorge (approved):

dpgeorge:
> ran one lot of unit tests on C6 and also re-ran the before/after binary size comparison on the boards mentioned in the PR description.

Great, thanks for re-testing!

Now merged.

PR #16140, opened 2024-11-02
cpydiff: Fix test case for json_nonserializable.

jepler:
The test case was producing the following error:
```
  File "<stdin>", line 12, in <module>
UnicodeError:
```
which did not demonstrate the intended difference. (this particular non-json-serializable object DID throw an exception! just not TypeError)

The updated version creates the non-conforming json string `"<module 'json'>"` instead, better illustrating the problem.

I tested by manually running the example using a coverage build & in cpython.

### DIFF
--- a/tests/cpydiff/modules_json_nonserializable.py
+++ b/tests/cpydiff/modules_json_nonserializable.py
@@ -6,10 +6,7 @@ workaround: Unknown
 """
 import json
 
-a = bytes(x for x in range(256))
 try:
-    z = json.dumps(a)
-    x = json.loads(z)
-    print("Should not get here")
+    print(json.dumps(b"shouldn't be able to serialise bytes"))
 except TypeError:
     print("TypeError")


tests/cpydiff/modules_json_nonserializable.py
```diff
@@ -8,8 +8,7 @@
 
 a = bytes(x for x in range(256))
 try:
-    z = json.dumps(a)
-    x = json.loads(z)
-    print("Should not get here")
+    z = json.dumps(json)
+    print("Produced non-conforming json string", repr(z))
```
dpgeorge, line 12: I think the original test was testing that `bytes` are not serialisable, which would be more of a subtle difference between CPython and MicroPython than trying to serialise a module.

Suggest to simply do:
```py
try:
    print(json.dumps(b"shouldn't be able to serialise this")
except TypeError:
    print("TypeError")
```

jepler:
There are CI errors but they appear unrelated to this change.
```
1 tests failed: time_time_ns
1 tests failed: thread_lock4
```

```diff
-- /home/runner/work/micropython/micropython/tests/results/thread_thread_lock4.py.exp	2024-11-07 16:44:53.393277488 +0000
+++ /home/runner/work/micropython/micropython/tests/results/thread_thread_lock4.py.out	2024-11-07 16:44:53.393277488 +0000
@@ -58,3 +58,4 @@
 77 145183092028285869634070784086308284983740379224208358846781574688061991349156420080065207861248000000000000000000
 78 11324281178206297831457521158732046228731749579488251990048962825668835325234200766245086213177344000000000000000000
 79 894618213078297528685144171539831652069808216779571907213868063227837990693501860533361810841010176000000000000000000
+CRASH
\ No newline at end of file

FAILURE /home/runner/work/micropython/micropython/tests/results/thread_thread_lock4.py
```

```diff
--- /home/runner/work/micropython/micropython/tests/results/extmod_time_time_ns.py.exp	2024-11-07 16:44:35.601734727 +0000
+++ /home/runner/work/micropython/micropython/tests/results/extmod_time_time_ns.py.out	2024-11-07 16:44:35.601734727 +0000
@@ -4,7 +4,7 @@
 True
 True
 True
-True
+False
 True
 True
 True

FAILURE /home/runner/work/micropython/micropython/tests/results/extmod_time_time_ns.py
```

dpgeorge:
I reran the failing CI and now it passes.

dpgeorge:
Rebased and merged in 3844733d604d83fbfec5592cee20aab065ebff48

PR #16147, opened 2024-11-04
tests/extmod_hardware: Add a test for machine.PWM freq and duty.

dpgeorge:
### Summary

This adds a hardware test for `machine.PWM`.  It requires a jumper wire between two pins, uses `machine.PWM` to output on one of them, and `machine.time_pulse_us()` to time the PWM on the other pin.

It times both the high and low duty cycle (and hence the frequency) for a range of PWM frequencies and duty cycles (including full on and full off).

### Testing

Currently supported on:
- rp2 (passes)
- mimxrt / Teensy 4.0 (has a few issues with constant on ie duty=100%)
- esp32 (passes with a minor hack for initialisation)
- esp8266 (passes for frequencies 1kHz and less)
- samd21 (has some timing issues at high frequencies)

Need a way to exclude higher frequencies on certain platforms (eg above 1kHz on esp8266):
- Could split the test up into low and high frequency tests (and skip high frequencies on certain platforms), but that duplicates a lot of code.
- Could just `assert` instead of printing, but that misses a lot of useful info.
- Could use the test runner ##### output fuzzy matching feature.
- Could improve the test runner so that it supports debug output within a test (eg prefixed by a magic string like "DEBUG") as well as normal output to compare for correctness.
- EDIT: test now reworked to use `unittest` which solves all the above problems.

### DIFF
--- a/tests/extmod_hardware/machine_pwm.py
+++ b/tests/extmod_hardware/machine_pwm.py
@@ -1,0 +1,166 @@
+# Test machine.PWM, frequncy and duty cycle (using machine.time_pulse_us).
+#
+# IMPORTANT: This test requires hardware connections: the PWM-output and pulse-input
+# pins must be wired together (see the variable `pwm_pulse_pins`).
+
+import sys
+import time
+
+try:
+    from machine import time_pulse_us, Pin, PWM
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+import unittest
+
+pwm_freq_limit = 1000000
+freq_margin_per_thousand = 0
+duty_margin_per_thousand = 0
+timing_margin_us = 5
+
+# Configure pins based on the target.
+if "esp32" in sys.platform:
+    pwm_pulse_pins = ((4, 5),)
+    freq_margin_per_thousand = 2
+    duty_margin_per_thousand = 1
+    timing_margin_us = 20
+elif "esp8266" in sys.platform:
+    pwm_pulse_pins = ((4, 5),)
+    pwm_freq_limit = 1_000
+    duty_margin_per_thousand = 3
+    timing_margin_us = 50
+elif "mimxrt" in sys.platform:
+    if "Teensy" in sys.implementation._machine:
+        # Teensy 4.x
+        pwm_pulse_pins = (
+            ("D0", "D1"),  # FLEXPWM X and UART 1
+            ("D2", "D3"),  # FLEXPWM A/B
+            ("D11", "D12"),  # QTMR and MOSI/MISO of SPI 0
+        )
+    else:
+        pwm_pulse_pins = (("D0", "D1"),)
+elif "rp2" in sys.platform:
+    pwm_pulse_pins = (("GPIO0", "GPIO1"),)
+elif "samd" in sys.platform:
+    pwm_pulse_pins = (("D0", "D1"),)
+    if "SAMD21" in sys.implementation._machine:
+        # MCU is too slow to capture short pulses.
+        pwm_freq_limit = 2_000
+else:
+    print("Please add support for this test on this platform.")
+    raise SystemExit
+
+
+# Test a specific frequency and duty cycle.
+def _test_freq_duty(self, pulse_in, pwm, freq, duty_u16):
+    print("freq={:<5} duty_u16={:<5} :".format(freq, duty_u16), end="")
+
+    # Check configured freq/duty_u16 is within error bound.
+    freq_error = abs(pwm.freq() - freq) * 1000 // freq
+    duty_error = abs(pwm.duty_u16() - duty_u16) * 1000 // (duty_u16 or 1)
+    print(" freq={} freq_er={}".format(pwm.freq(), freq_error), end="")
+    print(" duty={} duty_er={}".format(pwm.duty_u16(), duty_error), end="")
+    print(" :", end="")
+    self.assertLessEqual(freq_error, freq_margin_per_thousand)
+    self.assertLessEqual(duty_error, duty_margin_per_thousand)
+
+    # Calculate expected timing.
+    expected_total_us = 1_000_000 // freq
+    expected_high_us = expected_total_us * duty_u16 // 65535
+    expected_low_us = expected_total_us - expected_high_us
+    expected_us = (expected_low_us, expected_high_us)
+    timeout = 2 * expected_total_us
+
+    # Wait for output to settle.
+    time_pulse_us(pulse_in, 0, timeout)
+    time_pulse_us(pulse_in, 1, timeout)
+
+    if duty_u16 == 0 or duty_u16 == 65535:
+        # Expect a constant output level.
+        no_pulse = (
+            time_pulse_us(pulse_in, 0, timeout) < 0 and time_pulse_us(pulse_in, 1, timeout) < 0
+        )
+        self.assertTrue(no_pulse)
+        if expected_high_us == 0:
+            # Expect a constant low level.
+            self.assertEqual(pulse_in(), 0)
+        else:
+            # Expect a constant high level.
+            self.assertEqual(pulse_in(), 1)
+    else:
+        # Test timing of low and high pulse.
+        n_averaging = 10
+        for level in (0, 1):
+            t = 0
+            time_pulse_us(pulse_in, level, timeout)
+            for _ in range(n_averaging):
+                t += time_pulse_us(pulse_in, level, timeout)
+            t //= n_averaging
+            expected = expected_us[level]
+            print(" level={} timing_er={}".format(level, abs(t - expected)), end="")
+            self.assertLessEqual(abs(t - expected), timing_margin_us)
+
+    print()
+
+
+# Test a specific frequency with multiple duty cycles.
+def _test_freq(self, freq):
+    print()
+    self.pwm.freq(freq)
+    for duty in (0, 10, 25, 50, 75, 90, 100):
+        duty_u16 = duty * 65535 // 100
+        if sys.platform == "esp32":
+            # TODO why is this bit needed to get it working on esp32?
+            self.pwm.init(freq=freq, duty_u16=duty_u16)
+            time.sleep(0.1)
+        self.pwm.duty_u16(duty_u16)
+        _test_freq_duty(self, self.pulse_in, self.pwm, freq, duty_u16)
+
+
+# Given a set of pins, this test class will test multiple frequencies and duty cycles.
+class TestBase:
+    @classmethod
+    def setUpClass(cls):
+        print("set up pins:", cls.pwm_pin, cls.pulse_pin)
+        cls.pwm = PWM(cls.pwm_pin)
+        cls.pulse_in = Pin(cls.pulse_pin, Pin.IN)
+
+    @classmethod
+    def tearDownClass(cls):
+        cls.pwm.deinit()
+
+    def test_freq_50(self):
+        _test_freq(self, 50)
+
+    def test_freq_100(self):
+        _test_freq(self, 100)
+
+    def test_freq_500(self):
+        _test_freq(self, 500)
+
+    def test_freq_1000(self):
+        _test_freq(self, 1000)
+
+    @unittest.skipIf(pwm_freq_limit < 2000, "frequency too high")
+    def test_freq_2000(self):
+        _test_freq(self, 2000)
+
+    @unittest.skipIf(pwm_freq_limit < 5000, "frequency too high")
+    def test_freq_5000(self):
+        _test_freq(self, 5000)
+
+    @unittest.skipIf(pwm_freq_limit < 10000, "frequency too high")
+    def test_freq_10000(self):
+        _test_freq(self, 10000)
+
+
+# Generate test classes, one for each set of pins to test.
+for pwm, pulse in pwm_pulse_pins:
+    cls_name = "Test_{}_{}".format(pwm, pulse)
+    globals()[cls_name] = type(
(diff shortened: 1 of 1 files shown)


robert-hh:
@dpgeorge  What is the problem with MIMXRT and constant high setting. At least PWM is creating a constant high signal? 
For SAM21 i understand, that the device is too slow to time higher frequencies.

tests/extmod_hardware/machine_pwm.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test():
+    pulse_in = Pin(pulse_pin, Pin.IN)
+    pwm = PWM(pwm_pin)
+
+    for freq in (50, 100, 500, 1_000, 2_000, 5_000, 10_000):
+        pwm.freq(freq)
+        for duty in (0, 10, 25, 50, 75, 90, 100):
+            duty_u16 = duty * 65535 // 100
```
robert-hh, line 97: Must be `duty_u16 = duty * 65536 // 100` to get a high pulse at all devices.

robert-hh:
time_pulse_us() is not good at timing repeating pulses. If the level at start is identical to the requested level, then it will return the remaining duration of that pulse. Changing the code to wait for the opposite level first improves it.
```
MP_WEAK mp_uint_t machine_time_pulse_us(mp_hal_pin_obj_t pin, int pulse_level, mp_uint_t timeout_us) {
    mp_uint_t start = mp_hal_ticks_us();
    while (mp_hal_pin_read(pin) == pulse_level) {
        if ((mp_uint_t)(mp_hal_ticks_us() - start) >= timeout_us) {
            return (mp_uint_t)-1;
        }
    }
    while (mp_hal_pin_read(pin) != pulse_level) {
        if ((mp_uint_t)(mp_hal_ticks_us() - start) >= timeout_us) {
            return (mp_uint_t)-2;
        }
    }
    start = mp_hal_ticks_us();
    while (mp_hal_pin_read(pin) == pulse_level) {
        if ((mp_uint_t)(mp_hal_ticks_us() - start) >= timeout_us) {
            return (mp_uint_t)-1;
        }
    }
    return mp_hal_ticks_us() - start;
}
```

tests/extmod_hardware/machine_pwm.py
robert-hh: But that fails with the actual ESP32 firmware. With PR #10854 it works properly. That PR is  major improvement for PWM. Even with the current implementation of PWM for ESP32, the output is NOT constant high at duty_u16=65535, but time_pulse_us() may be too slow at ESP32 to sample it.

dpgeorge: That's not true according to the docs, and the rp2 implementation.  Both of those have 65535 as the upper limit.

This definitely needs to be clarified!

dpgeorge:
> time_pulse_us() is not good at timing repeating pulses. If the level at start is identical to the requested level, then it will return the remaining duration of that pulse. Changing the code to wait for the opposite level first improves it.

I don't think we should change `time_pulse_us()`.  It works well and as specified in the docs.

The test here uses a dummy call to `time_pulse_us()` at to synchronise with the PWM pulses.  And then times each low pulse 10 times in a row.  Then it times 10 high pulses in a row.  That should be accurate.

tests/extmod_hardware/machine_pwm.py
robert-hh: > This definitely needs to be clarified!

The docs have to be updated. Besides ESP32 all ports accept 65536 as setting the output constant high and 0 for setting the output constant low. Depending on the dividers and frequencies, there may be no output pulse at e.g. 65535 (or 1).

robert-hh:
I was not suggesting changing time_pulse_us(). And the test script seems to implement that behavior.

tests/extmod_hardware/machine_pwm.py
dpgeorge: A duty of 0 should definitely mean a duty of 0%, ie constant low output.

Then if the duty range is 16-bits, the maximum you can specify is 65535.  And so that must be a constant high output.  (For example, what if you are storing duty values in 2 bytes of data, you need to be able to specify the whole range in 2 bytes.)

robert-hh: But doing it that way there is no 50% or 25% duty. Just numbers close to it. We can of course define 65535 as the value of "constant on", but keep the ratio as duty_u16/65536.

dpgeorge: There's also no perfectly accurate 10%, whether the division is /65535 or /65536.

With /65535 there is an accurate 20%, 40%, 60% and 80%.

But if the resolution of the underlying hardware PWM counter is <= 65535 (eg on rp2), then it should always be possible to select all of the available hardware duty cycles, including an accurate 50% (with /65535).  Eg on rp2 a duty_u16 of 32768 should give you exactly 50%, due to integer rounding when calculating the hardware counter limit.

Josverl:
Would it makes sense to extract the pins configuration to a seperate file/input?

I dont want to over-engineer, but having a common, perhaps shared, config will make running these HIL against different MCU  more efficient

dpgeorge:
> Would it makes sense to extract the pins configuration to a seperate file/input?

That would be nice.  But I think it's quite valuable to have a self contained test that's easy to run, and also all the requirements/settings for the test are all within that test.  Eg even if the pin config was in a special `board_test_setup.py` file, the tests still need to tune various timing margins for different ports, so still need logic at the top of the test to do this (and may as well then set the pins).

Well, maybe we can improve this in the future.  For now I want to encode the behaviour of hardware peripherals in the tests to make it easier to test ports and get them all behaving the same.

tests/extmod_hardware/machine_pwm.py
```diff
@@ -0,0 +1,108 @@
+# Test machine.PWM, frequncy and duty cycle (using machine.time_pulse_us).
+#
+# IMPORTANT: This test requires hardware connections: the PWM-output and pulse-input
+# pins must be wired together.
```
projectgus, line 4:
```suggestion
# pins must be wired together (see variables "pwm_pin" and "pulse_pin").
```

This is evident after you read and understanding the code, but if someone is running the whole test suite and doesn't overly care about PWM then it's all the information they need.

projectgus (approved):
Limitations notwithstanding, this looks good. It's a lot more thorough than the test I wrote!

projectgus:
> I dont want to over-engineer, but having a common, perhaps shared, config will make running these HIL against different MCU more efficient

FWIW I'd also like to see this eventually too (split out the hardware config and the test). I don't think it has to be a complex architecture, but I'd appreciate being a way to add a new test case without either appending to an existing file or copy-pasting the config into a new file.

However, as damien says, the test environment is currently simple because very few tests rely on additional files. It's a non-trivial amount of work to change this.

IhorNehrutsa:
ESP32C3 passed this test in #10854.

tests/extmod_hardware/machine_pwm.py
```diff
@@ -0,0 +1,108 @@
...
+            expected = expected_us[level]
+            print("", level, abs(t - expected) <= timing_margin_us or (expected, t), end="")
+        print()
+
+
+def test():
+    pulse_in = Pin(pulse_pin, Pin.IN)
+    pwm = PWM(pwm_pin)
+
+    for freq in (50, 100, 500, 1_000, 2_000, 5_000, 10_000):
```
IhorNehrutsa, line 94: for freq in FREQS:

tests/extmod_hardware/machine_pwm.py
```diff
@@ -0,0 +1,108 @@
...
+# pins must be wired together.
+
+import sys
+
+try:
+    from machine import time_pulse_us, Pin, PWM
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
```
IhorNehrutsa, line 13: FREQS1 = (1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
FREQS2 = (50, 100, 500, 1_000, 2_000, 5_000, 10_000)
FREQS3 = (100_000, 1_000_000, 10_000_000, 20_000_000, 40_000_000)
FREQS = FREQS1 + FREQS2 + FREQS3
FREQS = FREQS2

print("Test frequencies:", FREQS)

IhorNehrutsa:
Please see some of the improvements of `tests/extmod_hardware/machine_pwm.py` in #16161

IhorNehrutsa:
@dpgeorge and all interested persons.

I want to express my opinion about the range of duty PWM.
I believe that the range 0-65356 is more promising.

**First**, many ports already have a range of 0-65356.
I'm not sure: mimxrt, same, esp8266, nrf, esp32

**Second**, ports that do not have the physical ability to accept the value 65356 can easily replace it with 65355 internally.
```
#define MAX_DUTY_UI16 65356
void set_duty_u16(machine_pwm_obj_t *self, int duty) {
    if ((duty < 0) || (duty > MAX_DUTY_UI16)) {
        mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("duty_u16 must be from 0 to %d"), MAX_DUTY_UI16);
    }
    if (duty > 65355)) {
        duty = 65355;
    }
    ...
}
```
At the same time, they will return the actual achieved duty value 65355.
```py
    pwm.duty_u16(65356)
    pwm.duty_u16()
    >>>65355
```

**Third**, at high frequencies the resolution decreases and most duty values ​​cannot be achieved. So, they are replaced by the closest possible ones. For example, when freq=20MHz the resolution decreases to 5 available duty value of all range 0-65356.
Available duty value are 0%, 25%, 50%, 75%, 100%, that's all.
All other duty values are converted to these values.

```
pwm = PWM(2, freq=20_000_000)
DUTY_MAX = 65536
DUTY_LEVELS = 8
for duty in range(0, DUTY_LEVELS + 1):
    duty_u16 = DUTY_MAX * duty // DUTY_LEVELS
    pwm.duty_u16(duty_u16)
    print(pwm)
```
output is 
```
>>>PWM(Pin(2), freq=20000000, duty_u16=0)     # duty_u16 = 0     =   0.00% Ok!
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=0)     # duty_u16 = 8192  =  12.50%    not achieved
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=16384) # duty_u16 = 16384 =  25.00% Ok!
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=16384) # duty_u16 = 24576 =  37.50%    not achieved
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=32768) # duty_u16 = 32768 =  50.00% Ok!
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=32768) #  duty_u16 = 40960 =  62.50%   not achieved
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=49152) #  duty_u16 = 49152 =  75.00% Ok!
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=49152) #  duty_u16 = 57344 =  87.50%   not achieved
                                              #                               not achieved
>>>PWM(Pin(2), freq=20000000, duty_u16=65536) #  duty_u16 = 65536 = 100.00% Ok!
```
So replacing 65356 with 65355 in the second paragraph works the same like here.

**Fourthly**, 65356 is a power of two and is more beautiful than 65355 purely aesthetically. :smile:

Sorry for the many words.
Thanks.

dpgeorge:
@IhorNehrutsa the method is named `duty_u16()` which means it takes a 16-bit unsigned integer as the argument.  So that means the duty must be in the range 0-65535.

For example you may want to store duty values in 2 bytes (eg in `array.array("H")`), and in that case you can't store 65536.  So the maximum needs to be 65535.

dpgeorge:
Once #16216 is merged, this test will be refactored to use `unittest`.  That will allow it to run and pass on esp8266, and also have more diagnostics output.

dpgeorge:
Teensy pins to be updated to test 3 output types, per https://github.com/micropython/micropython/pull/16153#issuecomment-2461903430

tests/extmod_hardware/machine_pwm.py
dpgeorge: Done.

dpgeorge:
This has been refactored to use `unittest`, and it's a lot cleaner and can now skip higher frequencies on esp8266.

I also added support for more than one set of pins, and when running on a Teensy board it uses the three sets suggested by @robert-hh .

andrewleech:
Now that it's using unittest it could print out the pins being used? Perhaps just on failure?

dpgeorge:
Updated to skip higher frequencies on SAMD21 (which is too slow to measure the short duty duration).  Retested on all supported ports.

dpgeorge:
> Now that it's using unittest it could print out the pins being used?

I have made a test class for each set of supported pins, which is automatically generated based on the list of pins provided by the port.  Eg on rp2 the test class is called `Test_GPIO0_GPIO1` (and the methods are `test_freq_50` etc).

And at the start of the test it now prints out the pins that are configured for that run, eg:
```
set up pins: GPIO0 GPIO1
test_freq_50 (__main__.TestBase) ...
freq=50    duty_u16=0     : freq=50 freq_er=0 duty=0 duty_er=0 :
freq=50    duty_u16=6553  : freq=50 freq_er=0 duty=6553 duty_er=0 : level=0 timing_er=1 level=1 timing_er=0
...
```

projectgus (approved):

PR #10850, opened 2023-02-24
ports/PWM: Reduce inconsistencies between the ports.

robert-hh:
This is the first set of commits, consisting of:

- mimxrt: Start PWM only if freq and duty are set. And straighten the duty handling.
- samd: Add the init() method to PWM. This is merely restructuring  existing code, at some costs for flash size and RAM usage. init() without argument restarts a PWM channel which had been stopped with deinit().
- rp2: Enable keyword arguments with the PWM constructor and add the init() method. init() restarts a PWM channel which had been stopped with deinit(). Added code to stop PWM on soft reset.
- Adapt the quickref examples to set freq and duty_u16 using keyword arguments in the constructor.
- ESP8266: Add the methods duty_u16() and duty_ns().
- rp2: Edd a invert=True|False keyword argument to init() and the constructor to invert channel. In order to use not more memory, some flags of the object are put into a bit field.

So Rp2, ESP8266, MIMXRT and SAMD support now keyword arguments with the constructor and an init() method with optional argumnents. All four do no start PWM until freq and duty are set. The behavior at deinit() is different. The state of the pins during set-up and after deinit() varies. For ESP32 there is a PR enforcing freq and duty settings with the constructor or init(). 

The STM32 and Renesas-RA ports do not have a machine.PWM class, but it could be added. The nrf port's PWM modules behaves completely different and requires a full rework to get compatible.

PR #16153, opened 2024-11-04
ports/machine_pwm: Set the PWM duty_u16() back to 65535.

robert-hh:
Tested that output is high at duty_u16(65565) and low at duty_u16(0). Also verified that at duty_u16(32768) the high and low pulse have the same length.

Tested with ESP8266, MIMXRT, NRF and SAMD, which are affected by this change. Other ports did not require a change.

Partially reverting PR #10850.

### DIFF
--- a/ports/mimxrt/hal/pwm_backport.c
+++ b/ports/mimxrt/hal/pwm_backport.c
@@ -36,7 +36,7 @@ void PWM_UpdatePwmDutycycle_u16(
 
     // Setup the PWM dutycycle of channel A or B
     if (pwmSignal == kPWM_PwmA) {
-        if (dutyCycle >= 65536) {
+        if (dutyCycle >= PWM_FULL_SCALE) {
             base->SM[subModule].VAL2 = 0;
             base->SM[subModule].VAL3 = pulseCnt;
         } else {
@@ -44,7 +44,7 @@ void PWM_UpdatePwmDutycycle_u16(
             base->SM[subModule].VAL3 = base->SM[subModule].VAL2 + pwmHighPulse;
         }
     } else {
-        if (dutyCycle >= 65536) {
+        if (dutyCycle >= PWM_FULL_SCALE) {
             base->SM[subModule].VAL4 = 0;
             base->SM[subModule].VAL5 = pulseCnt;
         } else {
@@ -160,7 +160,7 @@ status_t QTMR_SetupPwm_u16(TMR_Type *base, qtmr_channel_selection_t channel, uin
     if (dutyCycleU16 == 0) {
         // Clear the output at the next compare
         reg |= (TMR_CTRL_LENGTH_MASK | TMR_CTRL_OUTMODE(kQTMR_ClearOnCompare));
-    } else if (dutyCycleU16 >= 65536) {
+    } else if (dutyCycleU16 >= PWM_FULL_SCALE) {
         // Set the output at the next compare
         reg |= (TMR_CTRL_LENGTH_MASK | TMR_CTRL_OUTMODE(kQTMR_SetOnCompare));
     } else {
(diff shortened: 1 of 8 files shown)


ports/mimxrt/hal/pwm_backport.c
```diff
@@ -36,15 +36,15 @@ void PWM_UpdatePwmDutycycle_u16(
 
     // Setup the PWM dutycycle of channel A or B
     if (pwmSignal == kPWM_PwmA) {
-        if (dutyCycle >= 65536) {
+        if (dutyCycle >= PWM_FULL_SCALE) {
```
dpgeorge, line 39: This constant is still defined as 65536 in `pwm_backport.h`.  Does it need changing there?

projectgus (commented):
This looks good to me, thanks @robert-hh

ports/mimxrt/hal/pwm_backport.c
robert-hh: Of course.

robert-hh: Updated. Behavior still fine. At duty_u16(32768) the difference between low and high phases with a Teensy 4.x are 700ps (flexpwm) and 400ps (QTMR).

mattytrentini:
Thanks for taking care of all of these important details folks!

ports/mimxrt/hal/pwm_backport.c
dpgeorge: Can the hardware have even better symmetry with 50% duty cycle, if you choose the right upper limit for the counter?  If so, it may be possible to get this working with `machine.PWM` by using 65536 for the hardware full scale, but 65535 for the `duty_u16` max, and then scaling it via `x * 65536 / 65535`.  I think this is how the rp2 port works.

robert-hh: The difference between the high/low phase is 10% of the PWM clock or about 50% of the CPU clock. It hardly can get much better.

dpgeorge: Right... picoseconds are very small!  Great work testing it to that degree.

robert-hh:
1. A suitable pin pair on Teensy 4.x wich uses QTMR and has a UART is D14/D15 and UART3.
2. The PWM test passes with FLEXPWM_AB pins if duty and freq are set in a single call using pwm.init() and not  separate. That looks at first suprising, since all methods of setting PWM parameters use the same function for actually configuring the hardware and the sequence should not matter. I tested that initially, but not with the sequence duty_u16(65535) -> freq(x) -> duty_u16(0). That's the one which make trouble.

robert-hh:
The issues with the FLEXPM channels and duty_u16(0) are fixed. The changes are pushed. Tested with MIMXRT 1011, -1015, -1020, -1052, -1062 (Teensy 4.x) and -1176.

dpgeorge:
Excellent work @robert-hh !  We should probably make more of these hardware tests, they are really teasing out the bugs.

> The MIMXRT port has three different mechanisms for PWM

I think we should make the PWM test test all three of these mechanisms (for mimxrt boards).  Can you suggest three pairs of pins that can be used for this?  Maybe one pair can also have MISO/MOSI to test loopback SPI.

robert-hh:
>  Can you suggest three pairs of pins that can be used for this? Maybe one pair can also have MISO/MOSI to test loopback SPI.

Sure:
Pin D0/D1: FLEXPMW X and UART 1
Pin D2/D3: FLEXPWM A/B
Pin D11/D12: QTMR and MOSI/MISO of SPI 0

dpgeorge (approved):

dpgeorge:
Merged.  Thanks again @robert-hh for your efforts here.

robert-hh:
Thank you for setting up the test, which proves to be very useful.

PR #16156, opened 2024-11-05
tests: get `connect_nonblock_xfer.py` test working again and simplify it

dpgeorge:
### Summary

CPython changed its non-blocking socket behaviour recently and this test would not run under CPython anymore.  So the following steps were taken to get the test working again and then simplify it:
- Run the test against CPython 3.10.10 and capture the output into the .exp file for the test.
- Run this test on unix port of MicroPython and verify that the output matches the CPython 3.10.10 output in the new .exp file (it did).  From now on take unix MicroPython as the source of truth for this test when modifying it.
- Remove all code that was there for CPython compatibility.
- Make it print out more useful information during the test run, including names of the OSError errno values.
- Add polling of the socket before the send/write/recv/read to verify that the poll gives the correct result in non-blocking mode.

As part of this, it was found that bare-metal lwIP boards (eg PYBD_SF2, RPI_PICO_W) failed the test due to non-conforming behaviour of send/write on a non-blocking socket that's not yet connected.  That issue is also fixed in this PR.

### Testing
    
Tested on unix MicroPython, ESP32_GENERIC, PYBD_SF2 and RPI_PICO_W boards.  This test and all existing network tests pass.  Multinet tests pass.

Matching CPython behaviour here would be a lot of work and changes to socket code, which now uses `BlockingIOError`.  I don't think we need to do that, not a good use of time, is a big churn of functionality, 
and many things need to be retested and fixed (eg `asyncio`).

esp8266 can't pass this test because axtls doesn't fully support non-blocking mode, so the test is skipped on this platform.  It does actually pass the normal socket (not TLS) bits of this test, but separating those out into a separate test is not worth it, IMO.

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -701,7 +701,12 @@ static mp_uint_t lwip_tcp_send(lwip_socket_obj_t *socket, const byte *buf, mp_ui
 
     MICROPY_PY_LWIP_ENTER
 
-    u16_t available = tcp_sndbuf(socket->pcb.tcp);
+    // If the socket is still connecting then don't let data be written to it.
+    // Otherwise, get the number of available bytes in the output buffer.
+    u16_t available = 0;
+    if (socket->state != STATE_CONNECTING) {
+        available = tcp_sndbuf(socket->pcb.tcp);
+    }
 
     if (available == 0) {
         // Non-blocking socket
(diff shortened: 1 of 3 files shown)


projectgus (approved):
Agree with the rationale about not tracking CPython here.

dpgeorge:
> esp8266 can't pass this test because axtls doesn't fully support non-blocking mode, so the test is skipped on this platform. It does actually pass the normal socket (not TLS) bits of this test, but separating those out into a separate test is not worth it, IMO.

Converting this test to use `unittest` in #16216 allows it to run on esp8266 (skipping the TLS bits).

PR #16160, opened 2024-11-05
machine: Add nchanges param to time_pulse_us func.

IhorNehrutsa:
time_pulse_us(pin, pulse_level, timeout_us=1000000, **nchanges=2**, /)

If `nchanges` is 3, if the pin is initially equal to *pulse_level* then first
waits until the pin input becomes different from *pulse_level*.
Then if the current input value of the pin is different to *pulse_level*,
the function first waits until the pin input becomes equal to *pulse_level*,
then times the duration that the pin is equal to *pulse_level*.

**The advantage is that there is no need for additional synchronization before measuring the pulse duration.**
A little bit longer, but with higher accuracy.

The jitter of the pulse measurement process is about 5-15 μs on the ESP32 board.

Idia from @robert-hh https://github.com/micropython/micropython/pull/16147#issuecomment-2454287302

This function is used in DRAFT PR for tests
**PWM from 10845 and time_hardware_pulse_us() and test from 16147.** #16161
The tests results show the functionality of the function over the entire ESP32 PWM frequency range from 1Hz to 40MHz.

### DIFF
--- a/docs/library/machine.rst
+++ b/docs/library/machine.rst
@@ -316,6 +316,15 @@ Miscellaneous functions
    microseconds.  The *pulse_level* argument should be 0 to time a low pulse
    or 1 to time a high pulse.
 
+   If the current input value of the pin is different to *pulse_level*,
+   the function first (*) waits until the pin input becomes equal to *pulse_level*,
+   then (**) times the duration that the pin is equal to *pulse_level*.
+
+   The above behaviour is for when nchanges=2. That behaviour can be generalised by
+   passing in a larger value for nchanges which is the number of times the function
+   waits for the pin to change. On error the return value can be down to -nchanges and
+   indicates how many edges were missed.
+
    If *nchanges* is 3, if the pin is initially equal to *pulse_level* then first
    waits until the pin input becomes different from *pulse_level* (***).
    Then if the current input value of the pin is different to *pulse_level*,
--- a/docs/library/machine.rst
+++ b/docs/library/machine.rst
@@ -310,19 +310,23 @@ Miscellaneous functions
    varies by hardware (so use substring of a full value if you expect a short
    ID). In some MicroPython ports, ID corresponds to the network MAC address.
 
-.. function:: time_pulse_us(pin, pulse_level, timeout_us=1000000, /)
+.. function:: time_pulse_us(pin, pulse_level, timeout_us=1000000, nchanges=2, /)
 
    Time a pulse on the given *pin*, and return the duration of the pulse in
    microseconds.  The *pulse_level* argument should be 0 to time a low pulse
    or 1 to time a high pulse.
 
-   If the current input value of the pin is different to *pulse_level*,
-   the function first (*) waits until the pin input becomes equal to *pulse_level*,
-   then (**) times the duration that the pin is equal to *pulse_level*.
+   If *nchanges* is 3, if the pin is initially equal to *pulse_level* then first
+   waits until the pin input becomes different from *pulse_level* (***).
+   Then if the current input value of the pin is different to *pulse_level*,
+   the function first (**)(*nchanges* is 2) waits until the pin input becomes equal to *pulse_level*,
+   then (*)(*nchanges* is 1) times the duration that the pin is equal to *pulse_level*.
    If the pin is already equal to *pulse_level* then timing starts straight away.
+   Higher values of *nchanges* are possible, which allow more synchronisation.
 
+   The function returns -3 if there was timeout waiting for condition marked (***) above.
    The function will return -2 if there was timeout waiting for condition marked
-   (*) above, and -1 if there was timeout during the main measurement, marked (**)
+   (**) above, and -1 if there was timeout during the main measurement, marked (*)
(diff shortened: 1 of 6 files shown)


robert-hh:
Thanks for mentioning me. I considered creating that feature, but as optional argument to time_pulse_us() instead of a new method. That looks like a smaller change, even if the additional code size might be the same..

IhorNehrutsa:
Remove additionnal synchronization time_pulse_us() from tests/extmod_hardware/machine_pwm.py

dpgeorge:
After discussing with @projectgus , we feel that this is a useful enhancement but maybe not worth the cost in code size.

I think it would be possible to make this PR a lot simpler and less code size.  I opened #17346 as a basis for that.

If #17346 is merged, then the additional argument added here could be generalised to a counter, eg:
```py
machine.time_pulse_us(pin, pulse_level, timeout_us=1000000, num_levels=2)
```
With `num_levels=2` it's the same behaviour as before.  With `num_levels=3` it waits for 3 level changes and that implements the behaviour in this PR.  Higher values are possible, which allow more synchronisation.

robert-hh:
> then the additional argument added here

It's not (yet) visible in PR 17346.

dpgeorge:
Now that #17346 is merged, this PR can be updated on that.

docs/library/machine.rst
```diff
@@ -173,19 +173,23 @@ Miscellaneous functions
...
+   Then if the current input value of the pin is different to *pulse_level*,
+   the function first (**)(*nchanges* is 2) waits until the pin input becomes equal to *pulse_level*,
+   then (*)(*nchanges* is 1) times the duration that the pin is equal to *pulse_level*.
    If the pin is already equal to *pulse_level* then timing starts straight away.
+   Higher values of *nchanges* are possible, which allow more synchronisation.
 
+   The function returns -3 if there was timeout waiting for condition marked (***) above.
    The function will return -2 if there was timeout waiting for condition marked
-   (*) above, and -1 if there was timeout during the main measurement, marked (**)
+   (**) above, and -1 if there was timeout during the main measurement, marked (*)
```
dpgeorge, line 338: It's quite tricky to document the `nchanges` parameter.

I think the original description should remain unchanged.  And then add a paragraph at the end stating something like "The above behaviour is for when `nchanges=2`.  That behaviour can be generalised by passing in a larger value for `nchanges` which is the number of times the function waits for the pin to change.  On error the return value can be down to `-nchanges` and indicates how many edges were missed."  Or something like that.

extmod/machine_pulse.c
```diff
@@ -71,10 +70,14 @@ static mp_obj_t machine_time_pulse_us_(size_t n_args, const mp_obj_t *args) {
...
         timeout_us = mp_obj_get_int(args[2]);
     }
-    mp_uint_t us = machine_time_pulse_us(pin, level, timeout_us);
-    // May return -1 or -2 in case of timeout
+    mp_uint_t nchanges = 2;
+    if (n_args > 3) {
+        nchanges = mp_obj_get_int(args[3]);
+    }
+    mp_uint_t us = machine_time_pulse_us(pin, level, timeout_us, nchanges);
+    // May return -1 or -2 or -3 in case of timeout
```
dpgeorge: This comment needs adjusting because the return value can be between -nchanges and -1 (on timeout).

tests/extmod_hardware/machine_pwm.py
```diff
@@ -80,9 +77,8 @@ def _test_freq_duty(self, pulse_in, pwm, freq, duty_u16):
         n_averaging = 10
         for level in (0, 1):
             t = 0
-            time_pulse_us(pulse_in, level, timeout)
             for _ in range(n_averaging):
-                t += time_pulse_us(pulse_in, level, timeout)
+                t += time_pulse_us(pulse_in, level, timeout, 3)
```
dpgeorge: I think this code should stay as it was, because inside this loop you don't need to wait for an extra change.

dpgeorge:
@IhorNehrutsa there are some review comments above that need to be addressed before this PR can move forward.

tests/extmod_hardware/machine_pwm.py
IhorNehrutsa: nchanges=4 works, nchanges=2 don't works

tests/extmod_hardware/machine_pwm.py
```diff
@@ -80,12 +77,11 @@ def _test_freq_duty(self, pulse_in, pwm, freq, duty_u16):
         n_averaging = 10
         for level in (0, 1):
             t = 0
-            time_pulse_us(pulse_in, level, timeout)
             for _ in range(n_averaging):
-                t += time_pulse_us(pulse_in, level, timeout)
+                t += time_pulse_us(pulse_in, level, timeout, 4)
```
IhorNehrutsa, line 80: On esp32 generic board if nchanges=4 works, if nchanges=2 or 3 don't works.
@robert-hh 
If you have time, could you test tests/extmod_hardware/machine_pwm.py on the mimxrt port?
Thanks.

PR #11349, opened 2023-04-27
.github/workflows/mpremote.yml: Fetch full history.

jimmo:
Instead of doing the shallow checkout followed by an unshallow-with-tags, just set fetch-depth=0 to get the full history to start with.

This prevents the issue we saw with the v1.20.0 release where the github checkout action does

```
git fetch --no-tags --prune --progress --no-recurse-submodules --depth=1000 origin +294baf52b346e400e2255c6c1e82af5b978b18f7:refs/tags/v1.20.0
```

(in effect creating a "fake" `v1.20.0` tag in the local repo)

and then our workflow gets the tags and history by doing

```
git fetch --prune --tags --unshallow
```

which then attempts to clobber the `v1.20.0` tag with the real one.

PR #16163, opened 2024-11-05
py/mkrules.mk: Use partial clone for submodules if available.

andrewleech:
### Summary

Micropython relies on a number of submodules for third party and chip vendor libraries. Users need to check these out before building their desired ports and Github Actions CI here needs to clone them all multiple times for every build.

Many of these are getting significantly larger over time, slowing down usage and consuming more disk space.  There can also be resistance to adding larger libraries / sdk's here as submodules to avoid these sort of issues.

As suggested in https://github.com/micropython/micropython/pull/11349#issuecomment-1524305575 by @dhalbert newer version of git have features to avoid pulling all historic / blob data which can have a significant impact of total data use.

The change here has been directly inspired by https://github.com/adafruit/circuitpython/issues/7225 and https://github.com/adafruit/circuitpython/pull/7788

### Testing

Very little so far, it's been pushed here to trigger CI to ensure it still passes and hopefully get an idea of any speed improvements gained.

This is using a standard feature of git with automatic fallback to previous behavior on error.

Shallow clone is the more common way of achieving this but that often results in issues due to the lack of history metadata, particularly around versions derived from tags etc. While that's less often an issue in submodules this filter method is generally considered a safer means of gaining similar size improvements.

### DIFF
--- a/py/mkrules.mk
+++ b/py/mkrules.mk
@@ -252,7 +252,11 @@ submodules:
 	$(ECHO) "Updating submodules: $(GIT_SUBMODULES)"
 ifneq ($(GIT_SUBMODULES),)
 	$(Q)cd $(TOP) && git submodule sync $(GIT_SUBMODULES)
-	$(Q)cd $(TOP) && git submodule update --init $(GIT_SUBMODULES)
+	# If available, do blobless partial clones of submodules to save time and space.
+	# A blobless partial clone lazily fetches data as needed, but has all the metadata available (tags, etc.).
+	# Fallback to standard submodule update if blobless isn't available (earlier than 2.36.0)
+	$(Q)cd $(TOP) && git submodule update --init --filter=blob:none $(GIT_SUBMODULES) || \
+	  git submodule update --init $(GIT_SUBMODULES)
 endif
 .PHONY: submodules
 


projectgus (approved):
This looks handy, thanks @andrewleech!

dpgeorge:
@andrewleech running `make submodules` now prints out the comment added in this commit.  Is that intentional?  I think probably not and the comment should be moved out of the make target.

Eg in `ports/unix`:
```
$ make submodules
Updating submodules: lib/mbedtls lib/berkeley-db-1.xx lib/micropython-lib
Synchronizing submodule url for 'lib/berkeley-db-1.xx'
Synchronizing submodule url for 'lib/mbedtls'
Synchronizing submodule url for 'lib/micropython-lib'
# If available, do blobless partial clones of submodules to save time and space.
# A blobless partial clone lazily fetches data as needed, but has all the metadata available (tags, etc.).
# Fallback to standard submodule update if blobless isn't available (earlier than 2.36.0)
```

andrewleech:
Yeah I've seen that pop up in a couple of issue reports too! It certainly was unintentional, I didn't realise a comment there would be echo'd. I've seen comments above make targets before and thought that looked a bit ugly / out of context but now I understand why.

dpgeorge:
I made a PR to fix this comment in #16595.

Issue #2929, opened 2017-03-03
VFS FatFS: mkdir with trailing slash fails when it should succeed


dpgeorge:
On CPython and unix uPy, `os.mkdir('directory/')` makes the requested directory.  But on bare metal ports using VFS with FatFS it fails with ENOENT because of the trailing slash.  I'd say this should be fixed.  (And in fact CPy and unix uPy allow any number of trailing slashes).

ryannathans:
If this is occurring #2466 might be resolved

agatti:
This was fixed in 4fd5b72a8b71c6b3a074bcc3b9f6385b049c3dcb, closing as completed.

Issue #2466, opened 2016-09-30
stmhal: os directory operations throw cryptic errors on paths with trailing '/'


ryannathans:
It seems as though directory operations (listdir, mkdir, chdir, etc.) with the 'os' module throw cryptic errors when the path contains a trailing '/'. For example: '/sd/cache/' as opposed to the working '/sd/cache'.

```
>>> os.listdir('/flash/cache/')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] EINVAL

>>> os.listdir('/flash/cache')
[]
```

```
>>> os.mkdir('/sd/cache/test2/')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: Error creating directory '/sd/cache/test2/'
>>> os.mkdir('/sd/cache/test2')

```

```
>>> os.chdir('/flash/cache/')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OSError: No such file or directory: '/flash/cache/'
>>> os.chdir('/flash/cache')

```

hosaka:
Why "cryptic" errors? 
`man errno` shows EINVAL as "Invalid Argument", and there's of course pages on the web that describe errors in detail.

I don't know if the FatFS supports stripping trailing slashes, I can only see trailing spaces trim in directory related functions. Perhaps we just shouldn't use trailing spaces and add a note to the documentation or strip them in the OS/VFS module before passing them to FatFS. But knowing the maintainers vision of keeping the code tight, it will probably be the former.

jonnor:
There seems to be an error here, with the same degree of information as other errors in MicroPython. So I think we can close this.

Generally we cannot promise to have very descriptive/detailed/pedagogical errors because error strings take up a lot of program space, which is highly constrained in microcontroller environments.

PR #16164, opened 2024-11-06
tools/mpremote: Support trailing slash on dest for non-recursive copy.

dpgeorge:
### Summary

This fixes a regression in db59e55fe7a0b67d3af868990468e7b8056afe42: prior to that commit `mpremote` supported trailing slashes on the destination of a normal (non-recursive) copy.

Add back support for that, with the semantics that a trailing slash requires the destination to be an existing directory.

### Testing

A test has been added to the `mpremote` tests (also a test for force copy, in a separate commit).  As part of getting the test working, needed to flush stdout before printing error messages to stderr.

The `mpremote` tests were then run on a PYBD_SF2.  They passed.

### DIFF
--- a/tools/mpremote/mpremote/commands.py
+++ b/tools/mpremote/mpremote/commands.py
@@ -129,8 +129,15 @@ def _remote_path_basename(a):
 
 def do_filesystem_cp(state, src, dest, multiple, check_hash=False):
     if dest.startswith(":"):
-        dest_exists = state.transport.fs_exists(dest[1:])
-        dest_isdir = dest_exists and state.transport.fs_isdir(dest[1:])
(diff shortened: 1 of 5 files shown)


dpgeorge:
Note: this is related to #2466 and #2929: filesystems on bare-metal ports don't like trailing slashes in their filenames, even if it refers to a directory.  So that's why trailing slashes need to be stripped before passing them down to the target board (also doing the strip in `mpremote` allows enforcing that anything ending in a slash is indeed a directory).

dpgeorge:
This would also go in a v1.24.1 patch release.

projectgus (approved):
LGTM!

dpgeorge:
For some reason the windows mingw dev CI is failing, but that's also happening on master so it's nothing to do with this PR.  That needs separate investigation.

PR #8955, opened 2022-07-24
esp32: Fix TouchPad on esp32s2 and esp32s3.

Wind-stormger:
Fixed the issue that TouchPad failed on esp32s2 and esp32s3 
after compiling firmware with ESP-IDF v4.4 and higher versions.

Addresses issue #8048 .

Solution proposed by this contributor https://github.com/micropython/micropython/issues/8048#issuecomment-1107432527 .

Tested with this code passed:

```python=
from machine import TouchPad, Pin
# ESP32S3 TouchPad on GPIO 1~14
t = TouchPad(Pin(i))
print(t.read())
```

Specific readings depend on the PCB and its peripherals.

PR #16169, opened 2024-11-06
esp32: Fix stuck TouchPad readings on ESP32-S3, use version specifiers in code.

### LINKED ISSUE #13178: esp32s3 TouchPad not working
from machine import TouchPad, Pin
import time
t = TouchPad(Pin(5))
for i in range(100):
    time.sleep(0.5)
    print("%X"%t.read())

The same code changes the number with touch on ESP32, but the reading does not change on ESP32s3

Adjusting the initialization code as follows works fine, but I am not familiar enough with ESP32 and have not fully tested it, for reference only. Additionally, I suspect that the same issue also exists in ESP32s2
The changes are minimal, move the touch_Pad_Fsm_Start  to touch_Pad_config After 

STATIC mp_obj_t mtp_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw,
    const mp_obj_t *args) {
    mp_arg_check_num(n_args, n_kw, 1, 1, true);
    gpio_num_t pin_id = machine_pin_get_id(args[0]);
    const mtp_obj_t *self = NULL;
    for (int i = 0; i < MP_ARRAY_SIZE(touchpad_obj); i++) {
        if (pin_id == touchpad_obj[i].gpio_id) {
            self = &touchpad_obj[i];
            break;
        }
    }
    if (!self) {
        mp_raise_ValueError(MP_ERROR_TEXT("invalid pin for touchpad"));
    }

    static int initialized = 0;
    if (!initialized) {
        touch_pad_init();
        touch_pad_set_fsm_mode(TOUCH_FSM_MODE_TIMER);
        initialized = 1;
    }
    #if CONFIG_IDF_TARGET_ESP32
    esp_err_t err = touch_pad_config(self->touchpad_id, 0);
    #elif CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
    esp_err_t err = touch_pad_config(self->touchpad_id);
    touch_pad_fsm_start();
    #endif
    if (err == ESP_OK) {
        return MP_OBJ_FROM_PTR(self);
    }
    mp_raise_ValueError(MP_ERROR_TEXT("Touch pad error"));
}

projectgus:
### Summary

* Updates the `machine.TouchPad` driver to use hardware version specifiers instead of chip names. This should make adding future SoCs easier.
* Fixes #13178, by updating the fix applied in #8955.
* Update the docs for ESP32-S3 and ESP32-S3 TouchPad differences.

The new order of initialisation matches more closely the order in Espressif's example code, i.e.
https://github.com/espressif/esp-idf/blob/v5.2.2/examples/peripherals/touch_sensor/touch_sensor_v2/touch_pad_read/main/tp_read_main.c#L86

### Testing

* Tested on ESP32, ESP32-S2 and ESP32-S3 using the example code from the linked issue. Plus a similar example which initialised multiple channels.
* Unfortunately I had no working boards with PCB touch pads, so used dupont wires hanging off dev board connections... Works well enough.
 
The only buggy case of stuck max readings that I was able to reproduce was on ESP32-S3, and never reproduced once the fix was applied.

The hardware FSM did seem to get stuck once on ESP32-S3 in a different way (was reading correctly and then stopped updating and didn't start again). However I think this was due to my poor quality touch pad inputs (had moved the wires a lot and it would have changed capacitance by a huge factor).

* The Espressif touch sensor V2 hardware in S2 and S3 has a lot of features (denoise, background averaging, etc) which MicroPython is not using. It'd be good to support this, although it'd be a lot of work for someone and also may break code that relies on the current "raw" readings.

### DIFF
--- a/docs/esp32/quickref.rst
+++ b/docs/esp32/quickref.rst
@@ -751,20 +751,33 @@ APA102 (DotStar) uses a different driver as it has an additional clock pin.
 Capacitive touch
 ----------------
 
-Use the ``TouchPad`` class in the ``machine`` module::
+ESP32, ESP32-S2 and ESP32-S3 support capacitive touch via the ``TouchPad`` class
+in the ``machine`` module::
 
     from machine import TouchPad, Pin
 
     t = TouchPad(Pin(14))
     t.read()              # Returns a smaller number when touched
 
-``TouchPad.read`` returns a value relative to the capacitive variation. Small numbers (typically in
-the *tens*) are common when a pin is touched, larger numbers (above *one thousand*) when
-no touch is present. However the values are *relative* and can vary depending on the board
-and surrounding composition so some calibration may be required.
+``TouchPad.read`` returns a value proportional to the capacitance between the
+pin and the board's Ground connection. On ESP32 the number becomes smaller when
+the pin (or connected touch pad) is touched, on ESP32-S2 and ESP32-S3 the number
+becomes larger when the pin is touched.
 
-There are ten capacitive touch-enabled pins that can be used on the ESP32: 0, 2, 4, 12, 13
-14, 15, 27, 32, 33. Trying to assign to any other pins will result in a ``ValueError``.
+In all cases, a touch causes a significant change in the return value. Note the
+returned values are *relative* and can vary depending on the board and
+surrounding environment so some calibration (i.e. comparison to a baseline or
+rolling average) may be required.
+
+========= ==============================================
+Chip      Touch-enabled pins
+--------- ----------------------------------------------
+ESP32     0, 2, 4, 12, 13, 14, 15, 27, 32, 33
+ESP32-S2  1 to 14 inclusive
+ESP32-S3  1 to 14 inclusive
+========= ==============================================
+
+Trying to assign to any other pins will result in a ``ValueError``.
 
 Note that TouchPads can be used to wake an ESP32 from sleep::
 
--- a/ports/esp32/machine_touchpad.c
+++ b/ports/esp32/machine_touchpad.c
@@ -29,12 +29,14 @@
 #include "modmachine.h"
 #include "driver/gpio.h"
 
-#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
+#if SOC_TOUCH_SENSOR_SUPPORTED
 
-#if CONFIG_IDF_TARGET_ESP32
+#if SOC_TOUCH_VERSION_1 // ESP32 only
 #include "driver/touch_pad.h"
-#elif CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
+#elif SOC_TOUCH_VERSION_2 // All other SoCs with touch, to date
 #include "driver/touch_sensor.h"
+#else
+#error "Unknown touch hardware version"
 #endif
 
 typedef struct _mtp_obj_t {
@@ -70,6 +72,8 @@ static const mtp_obj_t touchpad_obj[] = {
     {{&machine_touchpad_type}, GPIO_NUM_12, TOUCH_PAD_NUM12},
     {{&machine_touchpad_type}, GPIO_NUM_13, TOUCH_PAD_NUM13},
     {{&machine_touchpad_type}, GPIO_NUM_14, TOUCH_PAD_NUM14},
+    #else
+    #error "Please add GPIO mapping for this SoC"
     #endif
 };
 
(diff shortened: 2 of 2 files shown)


projectgus:
@wangshujun-tj Thanks for submitting the bug report and the description of a fix, and for being patient while we followed up.

@wangshujun-tj @brainstorm @SkaveRat do any of you have hardware that you can use to confirm this fix before we merge it?

brainstorm:
Tested on my side, works like a charm 🚀

brainstorm:
Tested on an aditional (different design/prototype) board with an esp32s3-mini-1u module: the difference between having this PR applied or not means working touch or not.

docs/esp32/quickref.rst
```diff
@@ -750,20 +750,33 @@ APA102 (DotStar) uses a different driver as it has an additional clock pin.
...
+returned values are *relative* and can vary depending on the board and
+surrounding environment so some calibration (i.e. comparison to a baseline or
+rolling average) may be required.
+
+========= ==============================================
+Chip      Touch-enabled pins
+--------- ----------------------------------------------
+ESP32     0, 2, 4, 12, 13, 14, 15, 27, 32, 33
+ESP32-S2  1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14
+ESP32-S3  1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14
```
dpgeorge, line 776: Is it more readable to have `1 through 14 inclusive`?

ports/esp32/machine_touchpad.c
```diff
@@ -29,12 +29,16 @@
...
-#if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
+#if SOC_TOUCH_SENSOR_SUPPORTED
 
 #if CONFIG_IDF_TARGET_ESP32
 #include "driver/touch_pad.h"
-#elif CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
+#define TOUCH_HW_VER 1
+#elif SOC_TOUCH_VERSION_2
 #include "driver/touch_sensor.h"
+#define TOUCH_HW_VER 2
```
dpgeorge, line 39: Could the code below just use the IDF `SOC_TOUCH_VERSION_x` macros instead of defining a custom one?

And maybe to be sure the logic works, `#define SOC_TOUCH_VERSION_1 (0)` when `SOC_TOUCH_VERSION_2` is defined/enabled, and vice versa?

See also comments below.

ports/esp32/machine_touchpad.c
```diff
@@ -92,28 +98,29 @@ static mp_obj_t mtp_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_
     if (!initialized) {
         touch_pad_init();
         touch_pad_set_fsm_mode(TOUCH_FSM_MODE_TIMER);
-        #if CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
-        touch_pad_fsm_start();
-        #endif
         initialized = 1;
     }
-    #if CONFIG_IDF_TARGET_ESP32
+    #if TOUCH_HW_VER == 1
```
dpgeorge, line 103: Could it be `#if SOC_TOUCH_VERSION_1`?

ports/esp32/machine_touchpad.c
```diff
@@ -92,28 +98,29 @@ static mp_obj_t mtp_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_
...
     esp_err_t err = touch_pad_config(self->touchpad_id, 0);
-    #elif CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
+    #elif TOUCH_HW_VER == 2
```
dpgeorge, line 105: And this could be `#elif `SOC_TOUCH_VERSION_2`?

ports/esp32/machine_touchpad.c
projectgus: Sure. I think I looked for `SOC_TOUCH_VERSION_1` and thought it was never defined in ESP-IDF (only V2). Although I looked again now and it is defined. Will change.

docs/esp32/quickref.rst
projectgus: Yes, have changed!

projectgus:
Updated according to review, re-tested with dodgy dupont wire method on ESP32-S2 and original ESP32, rebased.

dpgeorge (approved):

PR #16191, opened 2024-11-08
ports/rp2/mphalport.h: Fix mp_hal_pin_xyz() for pin>=32

### LINKED ISSUE #16190: NeoPixel does not work on RP2350B with GPIO pin >=32

rp2 port with RP2350B

### MicroPython version

v1.24.0-preview.461.gb8227a3f7.dirty

### Reproduction

```
import neopixel
import machine

pin = 32

p = machine.Pin(pin)
n = neopixel.NeoPixel(p, 1)

n[0] = (127, 0, 0)
n.write()
```

### Expected behaviour

A NeoPixel connected to `pin` should illuminate.

### Observed behaviour

A NeoPixel connected to `pin` only illuminates if `pin` is <32. If `pin` is >=32, the NeoPixel does not illuminate.

### Additional Information

The root problem appears to be here:

https://github.com/micropython/micropython/blob/5dfbd4371480c175e6f54e57e4e5eb5d579cbd00/ports/rp2/mphalport.h#L184-L190

`gpio_set_mask()` requires a 32-bit value, so the shifted bit gets truncated when using a pin >= 32. There is a `gpio_set_mask64()`, however for some reason it didn't work when I tried it (I didn't dig into it). However changing to `gpio_put()` works fine on my end.

### Code of Conduct

Yes, I agree

sfe-SparkFro:
### Summary

Fixes #16190

### Testing

Tested on a board with an RP2350B, now NeoPixels on pins >=32 work.

~This uses `gpio_put()`, though `gpio_set_mask64()` might also work. The latter didn't work right away for me, and I did not bother testing further since `gpio_put()` worked fine, so it's possible I just did something wrong with `gpio_set_mask64()`.~


sfe-SparkFro:
Turns out the reason the 64-bit versions didn't work is because you need to explicitly typecast to a uint64_t. Updated to use mask functions since they should be faster than `gpio_put()`

Gadgetoid:
This slipped under my radar since we use our own PIO-based WS281X driver, but we might have a good excuse to give this a try and get some feedback here.

ports/rp2/mphalport.h
```diff
@@ -182,11 +182,11 @@ static inline void mp_hal_pin_od_high(mp_hal_pin_obj_t pin) {
 }
 
 static inline void mp_hal_pin_low(mp_hal_pin_obj_t pin) {
-    gpio_clr_mask(1 << pin);
+    gpio_clr_mask64((uint64_t) 1 << pin);
```
dpgeorge, line 185: Might be cleaner to use `1ULL` here instead of casting?  Same for the change below.

dpgeorge: Actually, even better, use `UINT64_C(1)`.

sfe-SparkFro: Only took me 3 weeks, but how's this now?

dpgeorge (approved):

PR #16197, opened 2024-11-09
ports/esp32/usb.c: Allow serial/jtag to be enabled.

bwhitman:
### Summary

I was having an issue getting USB serial going on an ESP32S3 LilyGO T-Deck, the flavor that is called "Serial/JTAG mode" in Micropython (not the one where there's an explicit serial -> USB chip onboard.) 

The REPL appears fine on the UART chip models. But on the T-Deck (and presumably other boards with no explicit UART chip) I cannot see or type into the REPL from the serial monitor. 

If i set in mpconfigport.h: 
```c
#define MICROPY_HW_USB_CDC (1)
#define MICROPY_HW_ESP_USB_SERIAL_JTAG (1)
```

I get a compiler error:  `#error "Invalid build config: Can't enable both native USB and USB Serial/JTAG peripheral"`. OK, so for the T-Deck I turn off HW_USB_CDC: 

```c
#ifdef TDECK
#define MICROPY_HW_USB_CDC (0)
#else
#define MICROPY_HW_USB_CDC (1)
#endif
```

But then ports/esp32/usb.c has a #ifdef for `MICROPY_HW_USB_CDC` bracketed around the entire code, including `usb_usj_mode` which is required in the bootloader. So, if i'm understanding it right -- we can't have both `USB_CDC` and `USB_SERIAL_JTAG` on at once. Makes sense. But `usb_usj_mode` requires `USB_CDC` to be enabled to switch the PHY back to Serial/JTAG mode. That doesn't make sense to me. So this PR moves the `#ifdef` out of the way of `usb_usj_mode`. 

### Testing

With this, everything compiles fine and I'm able to use `mpremote` etc to control the T-Deck over Serial/JTAG.

### DIFF
--- a/ports/esp32/usb.c
+++ b/ports/esp32/usb.c
@@ -30,13 +30,12 @@
 
 #include "esp_rom_gpio.h"
 #include "esp_mac.h"
-#include "esp_private/usb_phy.h"
 
-static usb_phy_handle_t phy_hdl;
 
 #if MICROPY_HW_USB_CDC
-
 #include "shared/tinyusb/mp_usbd.h"
+#include "esp_private/usb_phy.h"
+static usb_phy_handle_t phy_hdl;
 
 
 
@@ -67,10 +66,13 @@ void mp_usbd_port_get_serial_number(char *serial_buf) {
     mp_usbd_hex_str(serial_buf, mac, sizeof(mac));
 }
 
-#endif // MICROPY_HW_USB_CDC
-
+#else // !MICROPY_HW_USB_CDC
 
 #if CONFIG_IDF_TARGET_ESP32S3
+
+#include "esp_private/usb_phy.h"
+static usb_phy_handle_t phy_hdl;
+
 void usb_usj_mode(void) {
     // Switch the USB PHY back to Serial/Jtag mode, disabling OTG support
     // This should be run before jumping to bootloader.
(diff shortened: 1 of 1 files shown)


projectgus:
Hi @bwhitman,

Apologies, some of the esp32 build config for USB changed in v1.24 to make it more like the other ports but I realise there's no instructions on how to make it work. Will find a place to document this.

However, I think something is off with the T-Deck board config. We don't ship T-Deck as a board in MicroPython itself (I don't think), so I'm guessing the config is coming from somewhere externally and there might be other changes in the code?

The reason I say this is that the default ESP32_GENERIC_S3 board builds by default with both serial REPL (on UART0) and the internal native USB peripheral enabled (which gives a USB-CDC device on the native USB pins). This should work just as well as the Serial/JTAG support for getting a REPL on the same hardware, but sounds like it's not working on the T-Deck? This can probably be fixed by updating something in the T-Deck's build config or removing/tweaking some modifications in its source code.

Still, FWIW, to disable the native USB peripheral and enable the USB Serial/JTAG instead then all you should need to do (as of MicroPython v1.24) is:

```
#define MICROPY_HW_ENABLE_USBDEV            (0) 
```

Then `#define MICROPY_HW_ESP_USB_SERIAL_JTAG` will automatically default to 1, if the chip supports both (which ESP32-S3 does).

However, if you're building with a bunch of other config changes then some of these might be preventing both of these settings from working the same as they do here...

dpgeorge:
I think this was fully resolved by d1607598faf4ce2ef8e1a0385f75036ffb50ea9b

Issue #15841, opened 2024-09-12
Starting webrepl or binding to a socket on ESP32S3 crashes at modsocket.c:256


gampam2000:
### Port, board and/or hardware

esp32 esp32s3 

### MicroPython version

MicroPython v1.24.0-preview.309.gf1bdac375.dirty on 2024-09-12; ESP32-S3
ESP-IDF release/v5.2 commit c8c6fd1

### Reproduction

```
import webrepl
webrepl.start(password='test')
```

### Expected behaviour

webrepl is started
return to the micropython >>> promt

### Observed behaviour

it crashes instantly after  `webrepl.start(password='test')` is called.
It seem it is depended of the ESP-IDF v5.2 version, with an older version of ESP-IDF release/v5.2 I did not observe this for example with commit id: e92f89ca33624b79784f01dc3c1312cbf649210b lib/esp-idf (v5.2.2-384-ge92f89ca33)

```
���ESP-ROM:esp32s3-20210327
Build:Mar 27 2021
rst:0xc (RTC_SW_CPU_RST),boot:0x28 (SPI_FAST_FLASH_BOOT)
Saved PC:0x40382ed6
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fce3820,len:0x1108
load:0x403c9700,len:0x4
load:0x403c9704,len:0xbdc
load:0x403cc700,len:0x2f4c
entry 0x403c989c
Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x4005544b  PS      : 0x00060030  A0      : 0x82036505  A1      : 0x3fcbcb20  
A2      : 0x00000000  A3      : 0x3c148b5c  A4      : 0x3fcbcb70  A5      : 0x3fcbcbc0  
A6      : 0x00000000  A7      : 0x00000001  A8      : 0x82037ba4  A9      : 0x3fcbcad0  
A10     : 0x00000000  A11     : 0x3fca56a0  A12     : 0x00000001  A13     : 0xffffffff  
A14     : 0x00000000  A15     : 0x3fca56c8  SAR     : 0x0000001c  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000000  LBEG    : 0x400570e8  LEND    : 0x400570f3  LCOUNT  : 0x00000000  


[...]

PR #16210, opened 2024-11-11
esp32/modsocket: Fix getaddrinfo hints to set AI_CANONNAME.

dpgeorge:
### Summary

Because the `ai_canonname` field is subsequently used.

ESP32_GENERIC_S3 (at least) crashes with IDF 5.2.3 without this set.

See #15841, and prior attempt at a fix in #15999.

### Testing

Tested with the network tests, `./run-tests.py -t a0 net_*/*.py`.

- Tested with ESP32_GENERIC_S3 with IDF 5.2.3.  Tests crash the board prior to this fix, and pass after this fix (except, TLS seems to be broken in IDF 5.2.3.... that's a separate issue).
- Tested with ESP32_GENERIC_S3 with IDF 5.2.2.  Tests pass before and after this fix.

### DIFF
--- a/ports/esp32/modsocket.c
+++ b/ports/esp32/modsocket.c
@@ -213,7 +213,7 @@ static int mdns_getaddrinfo(const char *host_str, const char *port_str,
 #endif // MICROPY_HW_ENABLE_MDNS_QUERIES
 
 static void _getaddrinfo_inner(const mp_obj_t host, const mp_obj_t portx,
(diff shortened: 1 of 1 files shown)


projectgus (approved):
Ah, I [hacked around this](https://github.com/micropython/micropython/pull/15733#issuecomment-2316389875) to test something on IDF V5.3 a while back, too!

But this is the correct fix :)

Issue #16171, opened 2024-11-06
SDCard os.mount(sd,'/sd') OSError: [Errno 5] EIO 


FreddyBI:
### Port, board and/or hardware

ports esp32 build-ESP32_GENERIC Chip ESP32D0WDQ5 (revision 1)

### MicroPython version

MicroPython v1.25.0-preview.16.g919756cea on 2024-11-05; Generic ESP32 module with ESP32

### Reproduction

MicroPython v1.25.0-preview.16.g919756cea on 2024-11-05; Generic ESP32 module with ESP32

import machine
sd=machine.SDCard(slot=2)
import os
os.mount(sd,'/sd')

Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
OSError: [Errno 5] EIO

### Observed behaviour

no

### Additional Information

with build with IDF 5.0.4 and v1.2'.0-preview "no problem"

### Code of Conduct

Yes, I agree

dpgeorge:
Thanks for the report.  See #16222 and #16223 for candidate fixes.

dpgeorge:
Fixed by 5dc9eda1953668eb6861be01ca85f147dcf8d406

Issue #16214, opened 2024-11-12
Error mounting SDCARD in micropython v1.24.0 ESP32_GENERIC


dpgeorge:
### Discussed in https://github.com/orgs/micropython/discussions/16125

<div type='discussions-op-text'>

<sup>Originally posted by **AledelaCV** November  1, 2024</sup>
I have upgraded to version 1.23.0 to 1.24.0 encountering the error when mounting SDCard. The sample code in the previous version works without problem, but in the new version of micropython it generates error: 
```
Traceback (most recent call last):
  File "<stdin>", line 5, in <module>
OSError: [Errno 5] EIO
```

Here's the code I'm using:
```
import machine, os, vfs

# Slot 2 uses pins sck=18, cs=5, miso=19, mosi=23
sd = machine.SDCard(slot=2)
vfs.mount(sd, '/sd') # mount

print(os.listdir('/sd'))    # list directory contents
```

This code works perfectly in version 1.23.0</div>

dpgeorge:
Also possibly on stm32 (via @iabdalkader):

This commit broke something, the following code no longer works.
```
paste mode; Ctrl-C to cancel, Ctrl-D to finish
=== import pyb
=== sdcard = pyb.SDCard()
=== vfs = os.VfsFat(sdcard)
=== os.mount(vfs, "/sdcard")
=== 
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
OSError: [Errno 5] EIO
```

The same code works on 1.23.0 or without this commit

dpgeorge:
Related #16171.

dpgeorge:
The bug here does not manifest on stm32 by default (ie when you just let the SD card be mounted by the start-up code) because that uses `MP_BLOCKDEV_FLAG_NATIVE`.  The bug only manifests on stm32 when you manually access the SD card from Python.

dpgeorge:
See #16222 and #16223 for fixes.

dpgeorge:
Fixed by 5dc9eda1953668eb6861be01ca85f147dcf8d406

PR #16222, opened 2024-11-13
extmod/vfs_blockdev: Support bool return from Python read/write blocks.

dpgeorge:
### Summary

Commit f4ab9d924790581989f2398fe30bbac5d680577f inadvertently broke some Python block devices, for example esp32 and stm32 SDCard classes.  Those classes return a bool from their `readblocks` and `writeblocks` methods instead of an integer errno code.  With that change, both `False` and `True` return values are now be interpreted as non-zero and hence the block device call fails.

The fix in this commit is to allow a bool and explicitly convert `True` to 0 and `False` to `-MP_EIO`.

Fixes issues #16171 and #16214.

### Testing

Tested on PYBV11 using:
```py
import pyb, vfs
sdcard = pyb.SDCard()
vfs.VfsFat(sdcard)
```
That would previously fail, but with this patch passes.

Tested also on ESP32 (a Lolin 32 Pro).  Wit this patch mounting an SD card passes.

This is a simple fix and a candidate for a patch release (1.24.1).  An alternative is to make the Python blockdev functions return integers instead of bools, but that's an API change.

Edit: see #16223 for the alternative.

### DIFF
--- a/extmod/vfs_blockdev.c
+++ b/extmod/vfs_blockdev.c
@@ -58,6 +58,13 @@ static int mp_vfs_blockdev_call_rw(mp_obj_t *args, size_t block_num, size_t bloc
     if (ret == mp_const_none) {
         return 0;
     } else {
+        // Some block devices return a bool indicating success, so
+        // convert those to an errno integer code.
+        if (ret == mp_const_true) {
+            return 0;
+        } else if (ret == mp_const_false) {
+            return -MP_EIO;
+        }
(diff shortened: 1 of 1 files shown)


projectgus (approved):
Thanks for following this up. This looks good to me as the stop-gap for a patch release.

dpgeorge:
I tested on ESP32 and this patch fixes the issue.

PR #16223, opened 2024-11-13
ports: change block dev read/write methods to return integer errno instead of bool

dpgeorge:
### Summary

This is an alternative to #16222, and a more comprehensive fix.  It changes all Python-level block dev functions so they return an integer errno code instead of a bool.

Fixes issues #16171 and #16214.

### Testing

Tested on PYBV11 using:
```py
import pyb, vfs
sdcard = pyb.SDCard()
vfs.VfsFat(sdcard)
```
That would previously fail, but with this patch passes.

~~TODO: test on esp32.~~ now tested on ESP32_GENERIC_P4 with an SD card, running `vfs.VfsFat(machine.SDCard())`.  That passes.

This changes the API (the return value) of the stm32 and esp32 SDCard `readblocks` and `writeblocks` methods.  A non-API-changing fix is in #16222.

I propose to merge #16222 to fix the issue for a patch release, then this PR to properly fix things (and eventually revert #16222).


projectgus (commented):
This approach looks good to me for the long term. Let me know if I can help test on ESP32.

dpgeorge:
I've added a commit here to revert #16222.

dpgeorge:
This is now ready.  Tested on PYBV10 and ESP32_GENERIC_P4 with an SD card.

BUT!  It needs updating of the `tests/ports/stm32_hardwar/sdcard_dma_align.py` test because the return value of `readblocks()` changed.  Updating that depends on merging #19304 first.

projectgus (approved):
This looks good to me, good to standardise this! Only two questions:

* Are we concerned about changes in `extmod/vfs_blockdev.c` breaking existing Python block device drivers on update? It looks like trivial cases that always return `None` will still work, so the only risk is for any which return `True`/`False` - and maybe that's OK to change? Otherwise we could put it behind the MicroPython 2.0 feature flag.
* Is this an opportunity to update `reference/filesystem.rst` and `library/vfs.rst`? They currently don't mention the return type of readblocks & writeblocks at all.

projectgus:
> BUT! It needs updating of the tests/ports/stm32_hardwar/sdcard_dma_align.py test because the return value of readblocks() changed. Updating that depends on merging https://github.com/micropython/micropython/pull/19304 first.

I don't mind which order you want to do this in, if you prefer then I can rebase 19304 on this.

dpgeorge:
> I don't mind which order you want to do this in, if you prefer then I can rebase 19304 on this.

I think merge #19304 first.  That one is all ready to go.

projectgus:
@dpgeorge Merged!

dpgeorge:
> * Are we concerned about changes in `extmod/vfs_blockdev.c` breaking existing Python block device drivers on update? It looks like trivial cases that always return `None` will still work, so the only risk is for any which return `True`/`False` - and maybe that's OK to change? Otherwise we could put it behind the MicroPython 2.0 feature flag.

I'm not overly concerned, because:
- commit 5dc9eda1953668eb6861be01ca85f147dcf8d406 added this extra code specifically to work around esp32 and stm32 SDCard driver return values, which are fixed by this PR
- documentation for readblocks/writeblocks has examples that don't return anything (they return `None` implicitly), and it doesn't document anything about the return type, so hopefully no one was relying on returning anything

I think the biggest impact could actually be the return values of SDCard readblocks/writeblocks changing from bool to int, because success changes from True to 0.  I think we should just call that out in the release notes.

dpgeorge:
> * Is this an opportunity to update `reference/filesystem.rst` and `library/vfs.rst`? They currently don't mention the return type of readblocks & writeblocks at all.

Yes indeed.  I've now updated the docs.

dpgeorge:
> BUT! It needs updating of the `tests/ports/stm32_hardwar/sdcard_dma_align.py` test because the return value of `readblocks()` changed

That test is now updated.

PR #8991, opened 2022-07-31
ports/esp32: WLAN use esp_wifi_s/get_channel() to set and get wifi channel.

glenn20:
This PR updates `ports/esp32/network_wlan.c` to use the `esp_wifi_set_channel()` and `esp_wifi_get_channel()` to set and get the current wifi channel.

**Fixes:** WLAN.config('channel') returns incorrect channel for AP_IF if STA has connected to an external AP running on a different channel.

> Currently, WLAN.config('channel')  returns the configured channel for the AP_IF rather than the current channel. The current channel of the AP_IF will change if the STA_IF has connected to an Access Point on a different channel. eg:
> ```
> import network, time
> sta, ap = [network.WLAN(i) for i in (0, 1)]; sta.active(True); ap.active(True)
> ap.config(channel=6)
> sta.connect(ssid, password)    # Connect to AP running on channel 11
> while not sta.isconnected(): time.sleep(0.1)
> print("AP channel =", ap.config('channel'))
> ```
> will print "AP channel = 6" but the AP is actualy running on channel 11. With this PR, the correct "AP channel = 11" will be printed.
> 

**Extends**: Provides support for setting the secondary wifi channel as well as the primary channel.

> WLAN.config(channel=x) will also accept x as a tuple: (p, s) where p is the primary channel and s signifies the secondary channel by: 
> 
> - s < 0: WIFI_SECOND_CHAN_BELOW, s == 0: WIFI_SECOND_CHAN_NONE; s > 0: WIFI_SECOND_CHAN_ABOVE
> 
> Example:
> ```
> import network
> ap = network.WLAN(1); ap.active(True)
> ap.config(channel=6)
> ap.config(channel=(6,-1))
> ap.config(channel=(6,0))
> ap.config(channel=(6,+1))
> ```
> 

Also provides support for setting the channel of the STA_IF interface (supported in the IDF since IDFv4+).

PR #16225, opened 2024-11-13
esp32: Fix AP mode channel configuration.

### LINKED ISSUE #16165: wifi channel selection in access point mode is ignored

Waveshare ESP32-S3-ZERO

### MicroPython version

MicroPython v1.24.0 on 2024-10-25; Generic ESP32S3 module with ESP32S3

### Reproduction

Here is the code I use to test : 
```
import network

wlan = network.WLAN(network.AP_IF)
wlan.active(True)
wlan.config(ssid='potato',password='password',channel=6)

while True
    print(wlan.config('channel'))
```

### Expected behaviour

I expect this code to create an access point that will be on channel 6.

### Observed behaviour

The returned print command indicates the channel is "1", and a wifi scanner also confirms the access point is being broadcasted on channel 1.

### Additional Information

_No response_

### Code of Conduct

Yes, I agree

projectgus:
### Summary

* Seems at some point (probably following an ESP-IDF update), the channel setting code added in #8991 stopped working for the AP interface. This is because `esp_wifi_config()` is called after `esp_wifi_set_channel()`, the old channel gets re-configured and the AP doesn't change channels as expected.
* For AP mode, calling `esp_wifi_config()` seems to be all that's needed to get desired behaviour. The Wi-Fi stack configures 40MHz bandwidth and selects the secondary channel based on the available channels.
* However, the `esp_wifi_set_channel()` call is required in the particular case of ESP-NOW being used on the STA interface without an active connection to an AP. Testing suggests this is the only way to configure a particular channel in that case (unlike if the STA is associated to an AP, which means ESP-NOW has to use the same channel as the AP).
* Make some docs updates regarding Wi-Fi channels, in particular the role of channels in ESP-NOW (which is pretty restricted, I think this is something that has changed in later ESP-IDF versions).

Closes #16165.

### Testing

* Added a multi-test for explicitly configuring the ESP-NOW channel. This test passed on master, and still passes in this PR with the channel fix.
* Added a wlan multi-test for simple STA and AP behaviour. This test fails for esp32 in master as the AP chooses the wrong channel, and passes in this branch. Also tested on esp8266 port and CYW43 (RPI_PICO_W). CYW43 does not currently pass, but I have some fixes to submit separately so that it passes.


projectgus:
@glenn20 If you have the time then I'd be interested to know your thoughts on this PR.

tests/multi_wlan/01_ap_sta.py
```diff
@@ -0,0 +1,117 @@
...
+import os
+import sys
+import time
+
+CHANNEL = 8
+
+# Note that on slower Wi-Fi stacks this bumps up against the run-multitests.py
+# timeout which expects <10s between lines of output. We work around this by
+# logging something half way through the wait_for loop...
+CONNECT_TIMEOUT = 15000
```
dpgeorge, line 22: I've hit this 10s limit a few times.  Happy to just increase it to 15s to make this test simpler, and for future tests.

Alternatively (but probably out of scope for this PR) could add a special debug output function on the `multitest` class and just ignore that data, eg:
```py
multitest.output_debug("random data that shouldn't affect test result")
```
(See related `output_metric`.)

tests/multi_wlan/01_ap_sta.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # Wait for station
+    if not wait_for(ap.isconnected):
+        raise RuntimeError("Timed out waiting for station, status ", ap.status())
+
+    print("AP got station")
+    time.sleep(
+        3
+    )  # depending on port, may still need to negotiate DHCP lease for STA to see connection
+
```
dpgeorge, line 70: It might be worth testing the result of `ap.status('stations')` here... although that could open a can of inconsistent worms.

dpgeorge:
The esp32 fix looks good, thank you!

For the new `multi_wlan` test: this also looks good, although eventually it would probably be good to rewrite it using `unittest`, that would make it less awkward (eg for the esp8266 fake output).

tests/multi_wlan/01_ap_sta.py
dpgeorge: Let's leave this as-is.  In the future we can support `unittest` in the multitests to improve things.

tests/multi_wlan/01_ap_sta.py
dpgeorge: Let's add that in a separate test.

glenn20:
> @glenn20 If you have the time then I'd be interested to know your thoughts on this PR.

@projectgus sorry for the slow response. I have been on holiday overseas with family. I will be flying home today and can take a closer look when I get back. This looks like a good fix on first inspection.

projectgus:
@glenn20 No need to apologise, enjoy your holiday!

dpgeorge:
The required fixes for CYW43 are now merged, so the new `multi_wlan` test added here should now pass on boards using that driver.

dpgeorge:
I think we should merge this.  @projectgus I guess it's good to go in?

PR #16226, opened 2024-11-13
extmod/network_cyw43: Allow configuring active AP interface.

### LINKED ISSUE #14224: RPI-PICO-W (CYW43) Wi-Fi AP can only be configured when not active
### Checks

pico w

### MicroPython version

Micropython RP2040 Board CDC COM3

### Reproduction

my pico w's access point is locked with the wrong password, it still has a password when i remove it I CANNOT FIX THIS AND THE SSID IS WRONG IT IS SET AS PICO2F6F NOT PicoServer AS IT SHOULD AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA JUST USING ACCESS POINTS BREAK IT

### Expected behaviour

expected to create AP and create webserver from there

### Observed behaviour

no exception, just broken

### Additional Information

projectgus:
### Summary

Configuring the AP for cyw43 writes to some buffers that are only sent to the modem when the interface is brought up. This means you can't configure the AP after calling active(True), the new settings seem to be accepted but the radio doesn't change.

This is different to the documented WLAN behaviour which says to call active(True) before configuring the interface. The esp8266 ~~and esp32~~ ports require this, even. *EDIT: only esp8266 requires this, esp32 works in either order*

Fix this by bouncing the AP interface after a config change, if it's active. Configuring with active(False) still works the same as before.

Adds a static variable to track interface active state, rather than relying on the LWIP interface state. This is because the interface state is updated by a driver callback and there's a race: if code calls active(True) and then config(a=b) then the driver doesn't know it's active yet and the changes aren't correctly applied.

Closes #14224

### Testing

This is one of two fixes which are required for RPI_PICO_W board to pass the wlan test added in #16225.

It is possible this pattern will cause the AP to come up briefly with the default "PICOabcd" SSID before being reconfigured, however (due to the aforementioned race condition) it seems like this may not happen at all before the new config is applied. However, if that was a concern then we could hold off bringing the AP interface up until after it's configured. Doing it that way would be a potential breaking change, though.

### DIFF
--- a/extmod/network_cyw43.c
+++ b/extmod/network_cyw43.c
@@ -60,6 +60,10 @@ typedef struct _network_cyw43_obj_t {
 static const network_cyw43_obj_t network_cyw43_wl_sta = { { &mp_network_cyw43_type }, &cyw43_state, CYW43_ITF_STA };
 static const network_cyw43_obj_t network_cyw43_wl_ap = { { &mp_network_cyw43_type }, &cyw43_state, CYW43_ITF_AP };
 
+// Avoid race conditions with callbacks by tracking the last up or down request
+// we have made for each interface.
+static bool if_active[2];
+
 static void network_cyw43_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     network_cyw43_obj_t *self = MP_OBJ_TO_PTR(self_in);
     struct netif *netif = &self->cyw->netif[self->itf];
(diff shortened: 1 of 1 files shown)


projectgus:
> I think the ordering of active(True) vs configuration is quite a tricky issue. If esp32/8266 really do need the interface to be up before it's configured, then won't they always possibly appear briefly as an unconfigured AP first?

Actually I was wrong about esp32, this restriction only exists on esp8266. Will update the commit message and description. esp32 can have `active(True)` and `config()` in either order, same as CYW43 (after this commit).

It looks like calling `active(True)` on the esp8266 without configuring first doesn't result in any AP coming up until you call `ap.config()` afterward.

dpgeorge (approved):
Tested on PYBD_SF6, changing the `ssid` while the AP is active and it does indeed cycle the AP and the new SSID is then visible.

PR #16234, opened 2024-11-13
ports/rp2/boards: Add SparkFun Thing Plus RP2350

sfe-SparkFro:
### Summary

To add support for this board from us: https://www.sparkfun.com/products/25134

Depends on #16057

Also depends on #15620, but this can be merged before that if needed

### Testing

Tested with above dependencies merged locally on an initial production board, and all seems well.


sfe-SparkFro:
Also added image: https://github.com/micropython/micropython-media/pull/76

dpgeorge:
This is blocked on #16057.

sfe-SparkFro:
@dpgeorge No longer depends on #16057! I realized this was using an old implementation, so updated to match the Pico 2 W implementation. Still have the soft dependency on #15620, but that doesn't need to prevent this from being merged.

dpgeorge (approved):

dpgeorge:
Rebased, squashed and merged in 0a9cc9014ac228e5c51a067b7d83a97a539899da

PR #16235, opened 2024-11-14
ports/mimxrt: Restructure nxp_sdk to match official mcux-sdk.

andrewleech:
### Summary

The official mcux-sdk follows a slightly different structure to the current nxp_sdk submodule with many drivers moved to a common location.

To ease updating to newer versions of the SDK and/or adding new families the nxp_sdk submodule has been updated to follow the structure of mcux-sdk, just trimmed down to families used here to considerably reduce the size.

This PR is an update on https://github.com/micropython/micropython/pull/11516  
I've retargetted it to an updated nxp_driver submodule pointing directly at my fork; tinyusb are no longer using this repo so I believe it makes sense to move it here. I'd be quite happy for it to be moved from my github to micropython, but that is a separate discussion really.

The initial push of this in 4df2f8d4 is still based on mcux-sdk 2.11 which is equivalent to the existing version of nxp_driver, however it's been updated to match the official release from `https://github.com/nxp-mcuxpresso/mcux-sdk/archive/refs/tags/MCUX_2.11.0.tar.gz` using the automated update script in https://github.com/hathach/nxp_driver/pull/10

### Testing

I've built all existing boards to ensure they at least compile using:
``` bash
for board in $(cd ports/mimxrt/boards; find * -type d); do [ -e ports/mimxrt/boards/$board/mpconfigboard.mk ] && make -C ports/mimxrt BOARD="$board" -j18; done
```

I'll be continuing to work with this branch on an mimxrt-1176 so will update with test results here.

Originally https://github.com/micropython/micropython/pull/11516 was switching directly to the official sdk repo. Speaking to @dpgeorge there was a preference to keep using a trimmed down repo considering the dramatic difference in size (official ~1.6GB, trimmed < 100MB). 

The changes made by @imi415 to work with the official sdk layout of the repo are solid though and a better path forward than my previous efforts to use symlinks in the trimmed nxp_sdk repo to emulate the original layout, hence using their changes here 

The authorship of the commit has been kept as you @imi415 as the vast majority of the changes are still as per your PR. I hope you don't mind me making the minor tweaks for compatibility and squashing it together.


robert-hh:
Thanks for your ongoing work on this. It will take a few days before I can test it.  The wrapper for fsl_lpuart.c looks cool. Did you test that to fire at the right moment?
I'm working at the moment on changing the way flash devices are defined. At the moment it's a single flash_config.c file for all boards. But that does not properly account for the difference in the various flash devices. So I change than to have a flash_config.c per board. Alternatively it could be a flash_config.c per flash type, since it seem to be just three flash families being used: Winbond, ISSI and Adesto.

andrewleech:
> The wrapper for fsl_lpuart.c looks cool. Did you test that to fire at the right moment?

I was working in that last year now and can't quite remember what I tested on it. I did test a few things to do with the interrupts and wrappers because I was making a second similar change elsewhere.
I think I broke the functionality that depends on this idle hook though and then this wrapper fixed it so I decided it must be ok. 

> I'm working at the moment on changing the way flash devices are defined.

Neat! I'd love to consolidate spiflash handing across ports more, at least in terms of definition differences between the chip vendors. That's a much bigger job though...
In this case my vote would be for a common file/files that define the chips that can then just be referenced in the board mk/h files to select the one used.

robert-hh:
> In this case my vote would be for a common file/files that define the chips

Looking at it in  detail it gets less clean. The Winbond devices have minor difference, so it cannot be just a Winbond type of definition. So I'm working on making a flash_config per board and reducing the #defines for chip properties set in this definitions. Then everything is at a single place. The only #define that would stay is the one for the flash clock frequency, which may have to be changed at runtime.

robert-hh:
@andrewleech Did you make any progress in this task?

andrewleech:
Hi @robert-hh other than the new merge conflict to resolve I do think this PR is in pretty good shape.

robert-hh:
That is a trivial conflict about just one #include statement in mpconfigport.h being present or not.
```
#include "lib/cmsis/inc/core_cm7.h"
```
I tried to rebase onto the common master, with a rebase glitch in machine_rtc.c.
In May 2022 I had set up short list of to-do's for the switch from 2.10 to 2.11:

- enable retry-timer for I2C, SPI, UART and return timeout error
- format clock_config.c files with uncrustify
- reduce backport and fallback code from hal/backport.*, but not  sdcard.c

Not sure how much of that is still valid. The last bullet is a lot of work, but not for you.

robert-hh:
I made a few tests with that PR, building all boards and running functional tests with
- MIMXRT1010_EVK / Adafruit Metro M7
- MIMXRT1020_EVK
- Teensy 4.1
- MIMXRT1176_EVK

Tested festures:
- UART with UART_IRQ_RXIDLE
- SPI, using a SPI attached SD card
- I2C in combination with the I2S test
- I2S
- ADC
- PWM
- Ethernet, where applicable (not i.MX RT 1011)
- Built-in SD card supprt, where applicable (not i.MX RT 1011)

Result:
- All boards build.
- The above listed tests pass.

andrewleech:
Thanks for the testing @robert-hh that sounds great! I've just rebased to current master, hopefully it's essentially the same as what you tested (just resolving the conflict by removing the include, no other manual changes)

dpgeorge:
@robert-hh are you generally happy with this PR?  I think it's a good direction to go in, especially if we can host a trimmed-down version of the NXP SDK in the micropython GitHub org.

If you are happy, I'll do some basic testing, and then I think it's good to merge.

robert-hh:
@dpgeorge All boards build. I ran a few boards (1011, 1015, 1020, 1052, 1062 and 1170)  and they work including networking where supported, which is not too much of a surprise since the lib is still at version 2.11 as before. Moving the version up will be more work.

andrewleech:
Thanks @robert-hh I've got a local branch stacked on top of this that upgrades to 2.16 and includes some rework to the uart patch you've done previously, once this is merged I'll get that branch into a new clean PR to test and review

dpgeorge:
I've tested this on TEESY40 and there are no regressions.  And the PR itself looks good.

@andrewleech how do you want to handle the `nxp_driver` submodule, do you want to move that into the micropython org before merging this PR (and then update this PR to point to it), or do it as a separate step later on?  I'd prefer to move it to `micropython/nxp_driver` before merging, to minimise changes to `.gitmodules` which can sometimes lead to git difficulties.

andrewleech:
> how do you want to handle the `nxp_driver` submodule

Thanks @dpgeorge I've moved the repo to https://github.com/micropython/nxp_driver and merged the mcux 2.11 commit used/pinned here to master, then updated this PR to match.

dpgeorge (approved):
Thanks for moving across the `nxp_driver` repo.  This is now good to go in.

PR #16236, opened 2024-11-14
esp32/machine_pin.c: Fix the logic when to clear USB_SERIAL_JTAG_USB_PAD_ENABLE

### LINKED ISSUE #16414: Disabling USB_SERIAL_JTAG on esp32c3 to free up the pins.

esp32c3

### MicroPython version

MicroPython v1.25.0-preview.114.gbdda91fe7.dirty on 2024-12-14; ESP32C3 module with ESP32C3

### Reproduction

```import network
import espnow
from machine import Pin
from time import sleep

relay = Pin(19, Pin.OUT)

while True:
    relay.on()
    sleep(1)
    relay.off()
    sleep(1)
```

### Expected behaviour

Relay to Turn on and off.

### Observed behaviour

  File "<stdin>", line 13, in <module>
ValueError: invalid pin

### Additional Information

previous branch sdkconfig.c3usb
```
CONFIG_ESP32C3_REV_MIN_3=y
CONFIG_ESP32C3_BROWNOUT_DET=y
CONFIG_ESP32C3_BROWNOUT_DET_LVL_SEL_7=
CONFIG_ESP32C3_BROWNOUT_DET_LVL_SEL_4=y
CONFIG_ESP32C3_BROWNOUT_DET_LVL=4
CONFIG_ESP_CONSOLE_UART_DEFAULT=
CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG=y
```

master branch sdkconfig.c3usb
```
CONFIG_ESP32C3_REV_MIN_3=y

# Workaround for https://github.com/espressif/esp-idf/issues/14456
CONFIG_ESP_SYSTEM_HW_STACK_GUARD=n
```

Usually I can block ```CONFIG_ESP_CONSOLE_USB_SERIAL_JTAG=y``` and pin 18 and 19 work but not in this case? 

### Code of Conduct

Yes, I agree

garywill:
### Summary

When we don't use USB JTAG, we want to use the two USB pins (D+/D-) as GPIO. So, do clear USB_SERIAL_JTAG_USB_PAD_ENABLE when USB SERIAL JTAG is not enabled

### Testing

I was trying to build and use mpy v1.24.0 on ESP32-C3 (IDF 5.2.2). 

The code below shows GPIO 18 19 shoud be usable when `MICROPY_HW_ESP_USB_SERIAL_JTAG` is false:
https://github.com/micropython/micropython/blob/611d8f9ce82ab5e04fb86ab6cfc28d5ed98f33bf/ports/esp32/machine_pin.h#L90-L93

So I tried commenting out this:
https://github.com/micropython/micropython/blob/611d8f9ce82ab5e04fb86ab6cfc28d5ed98f33bf/ports/esp32/mpconfigport.h#L257-L261
to make sure `MICROPY_HW_ESP_USB_SERIAL_JTAG` is false.

Then the result is I could use `Pin(18)/Pin(19)` to get GPIO object, but could not do real GPIO output.

So I think we should apply this fix, which works for me on ESP32-C3

I don't have ESP32-C6 to test


projectgus:
Thanks @garywill for submitting this fix. I took a look at ESP32-C6 and it had a similar problem but had deviated from ESP32-C3 even more. Have pushed a commit here to make it the same behaviour as C3 (USB pins invalid if Serial/JTAG is enabled in the config, and can now be used otherwise).

I am currently working on some changes that will allow reconfiguring the USB pins for other uses on any SoC, without needing to make a custom build with USB disabled, provided that USB is not currently in use (related to #14217). However, that PR will not be ready for some time so this seems like a good fix to apply in the short term.

projectgus (approved):

dpgeorge (approved):

PR #16238, opened 2024-11-14
extmod/modframebuf.c: Fix 0 radius bug in FrameBuffer.ellipse.

### LINKED ISSUE #16053: `FrameBuffer.ellipse` goes into infinite loop with 0 radius

unix, RP2

### MicroPython version

MicroPython v1.23.0 on 2024-05-31; darwin [GCC 4.2.1] version

### Reproduction

In an interactive session (or otherwise), run the following code:
```
import framebuf
buf = bytearray(10*10*2)
f = framebuf.FrameBuffer(buf, 10, 10, framebuf.RGB565)
f.ellipse(5, 5, 0, 0, 0xffff, True)
```

### Expected behaviour

Nothing should be drawn.  Code should complete normally.

*Maybe* an exception should be raised, but I think it's OK to just draw nothing.

### Observed behaviour

Micropython hangs; 100% CPU usage observed in unix port.  Strong suspicion this is an infinite loop.

### Additional Information

I'm very suspicious of the following code: https://github.com/micropython/micropython/blob/1b89c503db690967d50699abe0bfa942f6f6b15e/extmod/modframebuf.c#L542-L554

In this case `two_asquare` and `two_bsquare` are both 0, so the `stoppingx` and `stoppingy` look like they are never changed, but I haven't 100% proven it to myself.

### Code of Conduct

Yes, I agree

corranwebster:
### Summary

This fixes a bug in `FrameBuffer.ellipse` where it goes into an infinite loop if both radii are 0.

This fixes the bug with a simple pre-check to see if both radii are 0, and in that case sets a single pixel at the center. This is consistent with the behaviour of the method when called with just one of the radii set to 0, where it will draw a horizontal or vertical line of 1 pixel width.

The pixel is set with `setpixel_checked` so it should handle out-of-bounds drawing and sector masks correctly.

Fixes #16053.

### Testing

This fix also includes three new tests: one for the default behaviour, one for drawing out-of-bounds, and one for drawing with all sector masks set to 0.

Tests were run on the unix port; they should run on any port which supports the `framebuf` module.

There was a question in my mind about whether this should simply not draw anything, but I think it is better to draw a single pixel since if one radius is 0 you get a line of width 1, and a single pixel is the intersection of those two cases.


corranwebster:
Closing until I can sort out why I'm getting failures here.

extmod/modframebuf.c
```diff
@@ -536,6 +536,10 @@ static mp_obj_t framebuf_ellipse(size_t n_args, const mp_obj_t *args_in) {
     } else {
         mask |= ELLIPSE_MASK_ALL;
     }
+    if (args[2] == 0 && args[3] == 0) {
+        setpixel_checked(self, args[0], args[1], args[4], 1);
```
dpgeorge, line 540: I think the last argument should be `mask`, in case the user specifies no fill and no quadrants (in which case nothing should be drawn).

corranwebster: Thanks, I've fixed this - it should also draw nothing if the user specifies fill and no quadrants, I think, so the last argument should actually be `mask  & ELLIPSE_MASK_ALL` since the 5th bit is for fill/no-fill.

Also I've added a test for this case.

dpgeorge (approved):

dpgeorge:
Rebased and merged in e70048cf5934bf3db65ec7a1b6490635716282f5

Issue #16259, opened 2024-11-17
Multiple complications with Termux (Android, armv7a, clang)


CardealRusso:
### Port, board and/or hardware

Unix, armv7a, android

### MicroPython version

1.24

### Reproduction

I had to go through a series of hacks and even change the micropython source code to get it to work in Termux. Having a deformed, broken, probably deficient Micropython in the end.  Hopefully, this experience can be resolved in the future?
Termux 0.119, clang 19.1, Galaxy a01

### “make” on a fresh/unmodified ports/unix:
The first error I get is this:
```
../../py/emitglue.c:118:47: error: passing 'uint8_t *' (aka 'unsigned char *') to parameter of type 'char *' converts between pointers to integer types where one is of the unique plain 'char' type and the other is not [-Werror,-Wpointer-sign]
  118 |     __builtin___clear_cache((void *)fun_data, (uint8_t *)fun_data + fun_len);
```
at

https://github.com/micropython/micropython/blob/611d8f9ce82ab5e04fb86ab6cfc28d5ed98f33bf/py/emitglue.c#L118
Which I was able to “ bypass” by exchanging for:

```c
__builtin___clear_cache((char *)fun_data, (char *)fun_data + fun_len);
```

### gchelper_generic:
```
CC ../shared/runtime/gchelper_generic.c
../shared/runtime/gchelper_generic.c:114:14: error: variable 'r4' is uninitialized when used here [-Werror,-Wuninitialized]
  114 |     arr[0] = r4;
      |              ^~
../shared/runtime/gchelper_generic.c:104:21: note: initialize the variable 'r4' to silence this warning
  104 |     register long r4 asm ("r4");
      |                     ^
      |                      = 0
...(10 or more similar hidden)
../shared/runtime/gchelper_generic.c:123:14: error: variable 'r13' is uninitialized when used here [-Werror,-Wuninitialized]
  123 |     arr[9] = r13;
      |              ^~~
../shared/runtime/gchelper_generic.c:113:22: note: initialize the variable 'r13' to silence this warning
  113 |     register long r13 asm ("r13");

```

fixed by setting #define MICROPY_GCREGS_SETJMP (1) at top of both mpconfigport.h from ports/unix and mpy-cross

### pthreads
Despite being present and functional, it refuses to compile with pthreads.

```
CC mpthreadport.c
[...]

PR #16261, opened 2024-11-18
unix/mpthreadport: Work around lack of thread cancellation on Android.

agatti:
### Summary

This commit fixes thread-related compilation issues under Android using [Termux](https://termux.dev/en/) as its runtime environment.

On Android's libc (Bionic) thread cancellation is not implemented, but the Unix port uses that mechanism to provide asynchronous thread termination.  In this commit there is a workaround for that, by adding a new signal handler to each newly created thread, whose callback simply exits the thread.  Threads are then sent the new signal rather than being explicitly cancelled, which in turn trigger the signal handler to stop the thread execution at the next possible occasion.

This makes the cancellation behaviour differ slightly on Android, as threads are probably going to linger a little bit more since the method introduced in this commit is equivalent to setting PTHREAD_CANCEL_DEFERRED as the thread cancellation type.  On the other hand there are no guarantees of immediate cancellation using PTHREAD_CANCEL_ASYNCHRONOUS either.

This fixes the pthread-related issues reported in #16259.

### Testing

This has been tested on a no-name Android tablet running Android 11 on AArch64.  On AArch64 these changes make MicroPython build and run, and make the standard tests suite pass, along with running all the extra tests in `tests/thread/` via `run-tests.py` multiple times in a row.

This fixes compilation issues for a niche environment and eventual binary size increases are limited to the Android platform, not impacting neither the Unix port nor MicroPython as a whole otherwise.

### DIFF
--- a/ports/unix/mpthreadport.c
+++ b/ports/unix/mpthreadport.c
@@ -45,8 +45,14 @@
 // potential conflict with other uses of the more commonly used SIGUSR1.
 #ifdef SIGRTMIN
 #define MP_THREAD_GC_SIGNAL (SIGRTMIN + 5)
+#ifdef __ANDROID__
+#define MP_THREAD_TERMINATE_SIGNAL (SIGRTMIN + 6)
+#endif
 #else
 #define MP_THREAD_GC_SIGNAL (SIGUSR1)
+#ifdef __ANDROID__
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thanks for submitting this fix, and sorry it took some time to get to.

It looks good to me!

PR #16264, opened 2024-11-18
build: Fix build errors on Clang for AArch32.

agatti:
### Summary

This PR contains minor changes that allow the code to build with Clang for AArch32.  There are test failures on Android, (`select_poll_fd`, `vfs_posix` on both 32 and 64 bits, `float_parse`, `float_parse_doubleprec` on 64 bits only), however the latter two are due to minor floating point rounding errors, whilst `select_poll_fd` cannot be tested due to Python crashing when generating the expected output, and `vfs_posix` failing when calling `os.listdir("/")`, as on Android the device root directory cannot be enumerated unless the process runs as root.

This PR fixes the remaining issues reported in #16259.

### Testing

GCC compatibility was tested by building and running tests for the Arm Unix target, using the `tools/ci.sh` script.

Android compatibility was tested by building and running tests inside Termux on both 32-bit and 64-bit Android instances.

### DIFF
--- a/py/emitglue.c
+++ b/py/emitglue.c
@@ -115,7 +115,7 @@ void mp_emit_glue_assign_native(mp_raw_code_t *rc, mp_raw_code_kind_t kind, cons
     #endif
     #elif MICROPY_EMIT_ARM
     #if (defined(__linux__) && defined(__GNUC__)) || __ARM_ARCH == 7
-    __builtin___clear_cache((void *)fun_data, (uint8_t *)fun_data + fun_len);
+    __builtin___clear_cache((void *)fun_data, (char *)fun_data + fun_len);
     #elif defined(__arm__)
     // Flush I-cache and D-cache.
     asm volatile (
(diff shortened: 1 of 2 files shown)


py/emitglue.c
```diff
@@ -115,7 +115,7 @@ void mp_emit_glue_assign_native(mp_raw_code_t *rc, mp_raw_code_kind_t kind, cons
     #endif
     #elif MICROPY_EMIT_ARM
     #if (defined(__linux__) && defined(__GNUC__)) || __ARM_ARCH == 7
-    __builtin___clear_cache((void *)fun_data, (uint8_t *)fun_data + fun_len);
+    __builtin___clear_cache((void *)fun_data, (void *)((uint8_t *)fun_data + fun_len));
```
dlech, line 118:
```suggestion
    __builtin___clear_cache((void *)fun_data, (char *)fun_data + fun_len);
```

Usually `char *` gets special treatment, so might work instead of the double cast.

agatti: It worked indeed.  I wonder if there's anything in the spec saying that `char*` can implicitly casted into `void*` whilst `unsigned char*` can't, and why.  The Clang error message was pretty explicit into mentioning "non-void" pointers anyway.

Anyway, thanks for the suggestion, I'll update the PR right away!

dpgeorge (approved):

dpgeorge:
Thank you!

PR #16265, opened 2024-11-19
extmod/network_cyw43: Fix uninitialised variable in status('stations').

dpgeorge:
### Summary

The `num_stas` was uninitialised and if it happened to take the value 0 then no results were returned.  It now has the correct maximum value.

### Testing

Tested with AP on PYBD_SF6 and two STA's connected (PYBD_SF2 and ESP32_GENERIC_S2).  With the AP running the following code:
```py
while True:
    print(ap.isconnected(), ap.status('stations'))
    time.sleep(1)
```
Prior to this fix `ap.status('stations')` would sometimes be empty.  After the fix it always contained the correct MAC's.

### DIFF
--- a/extmod/network_cyw43.c
+++ b/extmod/network_cyw43.c
@@ -361,13 +361,15 @@ static mp_obj_t network_cyw43_status(size_t n_args, const mp_obj_t *args) {
             if (self->itf != CYW43_ITF_AP) {
                 mp_raise_ValueError(MP_ERROR_TEXT("AP required"));
             }
-            int num_stas;
-            uint8_t macs[32 * 6];
+            static const unsigned mac_len = 6;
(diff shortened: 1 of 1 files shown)


projectgus (approved):
LGTM, one more to chalk up to automated testing! I'll send another PR to update the wlan test to call `status('stations')` and check the other ports.

It'd also be nice to `#define MAC_LEN 6` at the top of this file and replace some of the other magic number 6s, but can do that in a follow-up as well.

dpgeorge:
> It'd also be nice to `#define MAC_LEN 6` at the top of this file and replace some of the other magic number 6s, but can do that in a follow-up as well.

Yes, I agree.  Also worth checking the other network drivers to see if they can have a similar improvement.

PR #16270, opened 2024-11-20
esp32: Add missing network.STAT_CONNECT_FAIL constant.

### LINKED ISSUE #13167: Network - AttributeError: 'module' object has no attribute 'STAT_CONNECT_FAIL'
According to the micropython library and docs, network has various status constants that are return whenever WLAN.status() is called. However, one of them should be STAT_CONNECT_FAIL. While mapping the different status constants, I get the error:

```
AttributeError: 'module' object has no attribute 'STAT_CONNECT_FAIL'
```

I tried searching if anyone had this issue as well, and didn't seem like it, and I doubt I'm doing anything wrong given that it's pretty straight forward. Here is this code:

```python
import network

status_names = {
    network.STAT_CONNECT_FAIL: 'STAT_CONNECT_FAIL',  
}
```

I'm guessing this is some sort of bug. I apologize in advance if somehow it's not a bug, it's a known issue, or I'm implementing it wrong.

Thanks.

projectgus:
### Summary

The esp32 port had `network.STAT_ASSOC_FAIL` for the same purpose, but this is undocumented and different to all other ports which use the documented `STAT_CONNECT_FAIL`. This PR deprecates `STAT_ASSOC_FAIL` for removal in MicroPython 2.0.

Closes #13167

### Testing

Flashed ESP32 port, checked that both constants exist and have the same numeric value.

Having two `STAT_` constants with the same numeric value might break someone's code if they're building a lookup table with reflection (for example), but I think this is preferable to either having to document port-specific statuses, or removing `STAT_ASSOC_FAIL` which will definitely break someone's code.

### Future work

I see that a bunch of ESP32-specific constants have been added (`STAT_NO_AP_FOUND_IN_RSSI_THRESHOLD`, `STAT_NO_AP_FOUND_IN_AUTHMODE_THRESHOLD`, etc.) These are undocumented and make it hard to write portable code using the `WLAN` interface.

Suggest in a future PR to remove these (return `STAT_NO_AP_FOUND` for these cases), and add a `wlan.status('verbose')` or similar that can return additional port-specific details like this as a short string.


dpgeorge (approved):
Thanks, this looks good, good for consistency.

(These `STAT_xxx` constants probably need to be moved to `network.WLAN.STAT_xxx` similar to the recent moving of `SEC_xxx` and `IF_xxx`, but that's a separate discussion.)

PR #16272, opened 2024-11-20
ci: Windows CI fixes for recent MSVC and CPython 3.12.

projectgus:
### Summary

CI has been failing on master for Windows builds recently. These are four commits to make it pass again:

#### 1) MSVC NAN

Recent MSVC versions have introduced a bug where the `NAN` macro is no longer a constant expression in C! Have chosen the workaround approach from here: https://stackoverflow.com/a/79199887 

#### 2) asyncio_new_event_loop test w/ CPython 3.12

~~Added an .exp file for the `asyncio_new_event_loop` test on mingw.  The MSYS2 CPython 3.12.7 package prints some warnings about "Task was destroyed but it is pending!" on interpreter shutdown ([Sample failure](https://github.com/micropython/micropython/actions/runs/11926334558/job/33239930992)).~~ The fix commit updates the test to add `loop.close()` which means the warning appears consistently on other CPython builds, and then filters the warning out via a custom exception handler.

#### 3,4) CPython 3.12 settrace

There is a documented anomaly in CPython 3.12 settrace behaviour: opcode event tracing is missing in this version but has been restored in 3.13. ([Sample failure](https://github.com/micropython/micropython/actions/runs/11926334558/job/33239930992)). MicroPython's behaviour matches all CPython versions except 3.12.

To help clue in anyone running it, the test now prints a warning that it won't pass when run under CPython 3.12 (not possible to SKIP from the CPython run). More context available in CPython issue https://github.com/python/cpython/issues/114480 and the docs PR linked from that issue.

Also changed the mingw CI workflow to install CPython 3.11 from `setup-python` action rather than CPython 3.12.7 from MSYS2. This works around the settrace failure. This commit can be reverted in the future once MSYS2 repository updates to a newer CPython.

### Testing

Ran the failing tests locally with Python 3.12.7 and Python 3.10.

Ran the updated asyncio_new_event_loop test on unix port, rp2 port, and esp32 port.

Otherwise relying on CI to test this.

### Alternatives & Trade-Offs

* We could manually download and install an earlier MSYS2 Python package, there are a number available in their repo archive. However this is unsupported by MSYS2 (due to library dependency compatibility) . Moving over to the setup-python action is less likely to break randomly in the future.

### DIFF
--- a/py/objfloat.c
+++ b/py/objfloat.c
@@ -47,6 +47,13 @@
 #define M_PI (3.14159265358979323846)
 #endif
 
+// Workaround a bug in recent MSVC where NAN is no longer constant.
+// (By redefining back to the previous MSVC definition of NAN)
+#if defined(_MSC_VER) && _MSC_VER >= 1942
+#undef NAN
+#define NAN (-(float)(((float)(1e+300 * 1e+300)) * 0.0F))
+#endif
+
 typedef struct _mp_obj_float_t {
     mp_obj_base_t base;
     mp_float_t value;
--- a/.github/workflows/ports_windows.yml
+++ b/.github/workflows/ports_windows.yml
@@ -110,6 +110,11 @@ jobs:
       run:
         shell: msys2 {0}
     steps:
+    - uses: actions/setup-python@v5
+      # note: can go back to installing mingw-w64-${{ matrix.env }}-python after
+      # MSYS2 updates to Python >3.12 (due to settrace compatibility issue)
+      with:
+        python-version: '3.11'
     - uses: msys2/setup-msys2@v2
       with:
         msystem: ${{ matrix.sys }}
@@ -118,9 +123,9 @@ jobs:
           make
           mingw-w64-${{ matrix.env }}-gcc
           pkg-config
-          mingw-w64-${{ matrix.env }}-python3
           git
           diffutils
+        path-type: inherit  # Remove when setup-python is removed
     - uses: actions/checkout@v4
     - name: Build mpy-cross.exe
       run: make -C mpy-cross -j2
--- a/tests/extmod/asyncio_new_event_loop.py
+++ b/tests/extmod/asyncio_new_event_loop.py
@@ -12,6 +12,16 @@ if hasattr(asyncio, "set_event_loop"):
     asyncio.set_event_loop(asyncio.new_event_loop())
 
 
+def exception_handler(loop, context):
+    # This is a workaround for a difference between CPython and MicroPython: if
+    # a CPython event loop is closed while there are tasks pending (i.e. not finished)
+    # on it, then the task will log an error. MicroPython does not log this error.
+    if context.get("message", "") == "Task was destroyed but it is pending!":
+        pass
+    else:
+        loop.default_exception_handler(context)
+
+
 async def task():
     for i in range(4):
         print("task", i)
@@ -22,17 +32,21 @@ async def task():
 async def main():
     print("start")
     loop.create_task(task())
-    await asyncio.sleep(0)
+    await asyncio.sleep(0)  # yields, meaning new task will run once
     print("stop")
     loop.stop()
 
 
 # Use default event loop to run some tasks
 loop = asyncio.get_event_loop()
+loop.set_exception_handler(exception_handler)
 loop.create_task(main())
 loop.run_forever()
+loop.close()
 
 # Create new event loop, old one should not keep running
 loop = asyncio.new_event_loop()
+loop.set_exception_handler(exception_handler)
 loop.create_task(main())
 loop.run_forever()
+loop.close()
--- a/tests/misc/sys_settrace_features.py
+++ b/tests/misc/sys_settrace_features.py
@@ -6,6 +6,10 @@ except AttributeError:
     print("SKIP")
     raise SystemExit
 
+if sys.version.startswith("3.12"):
+    # There is a CPython change in settrace that is reverted in 3.13!
+    print("WARNING: this test will fail when compared to CPython 3.12.x behaviour")
+
 
 def print_stacktrace(frame, level=0):
     # Ignore CPython specific helpers.


dpgeorge (approved):
Fantastic, thanks for fixing this!  Looks good to me.

dpgeorge:
> Recent MSVC versions have introduced a bug where the `NAN` macro is no longer a constant expression in C!

Wow, that's very unexpected!  @stinos FYI

stinos:
> Also changed the mingw CI workflow to install CPython 3.11 from setup-python action rather than CPython 3.12.7 from MSYS2.

Wow, surprised this even works. I've had numerous weird issues in the past when using anything but the msys2-supplied python version. IIRC because the 'normal' Python would have issues detecting the platform ran on, or path translation not working and so on. But apparently this all just works now.


> The MSYS2 CPython 3.12.7 package prints some warnings about "Task was destroyed but it is pending!" on interpreter shutdown

To be clear: this also happens with the Python from `setup-python`? And we're sure it is in fact a Python issue/bug and not something in the test code?

py/objfloat.c
```diff
@@ -47,6 +47,13 @@
 #define M_PI (3.14159265358979323846)
 #endif
 
+// Workaround a bug in recent MSVC where NAN is no longer constant.
+// (By redefining back to the previous MSVC definition of NAN)
+#if defined(_MSC_VER) && _MSC_VER >= 1942
```
stinos, line 52: It's tempting to make this `_MSC_VER == 1942` instead though on the other hand that might just lead to endless adjusting if the bug doesn't get fixed soon.

projectgus:
> To be clear: this also happens with the Python from setup-python? And we're sure it is in fact a Python issue/bug and not something in the test code?

It's kind of tricky. The `asyncio_new_event_loop` test code's `main` coro creates another task (running `test`), yields once with an `await asyncio.sleep(0)` and then calls `loop.stop()` to stop the entire event loop both tasks are running on. So the created task is left pending without running to completion. That's part of the intent of the task, from what I can tell.

The tasks run exactly the same on CPython and MicroPython, all of the output they print is the same.

However, then the interpreter shuts down and runs a bunch of finalizers including for tasks which may still be pending and waiting to run. This prints the warnings that show up in the output.

I've only seen this happen on the mingw64 build of CPython 3.12.7 in CI, I can't make it happen here and I don't think it happens with the setup-python version. However, I also don't think that's a bug: during interpreter the finalizer order is known to be arbitrary (as everything is going away), so my guess is it depends on some detail of the runtime environment or the build that is hard to reproduce.

But I think this is probably the best way to handle it, if we want to have a test that stops the event loop when it has a pending task on it. Otherwise we'd need to find a way to force the warning to always occur on all CPython versions (for example, clearing references and forcing a GC pass) and then implement the same warnings in MicroPython.

projectgus:
> > Also changed the mingw CI workflow to install CPython 3.11 from setup-python action rather than CPython 3.12.7 from MSYS2.
> 
> Wow, surprised this even works. I've had numerous weird issues in the past when using anything but the msys2-supplied python version. IIRC because the 'normal' Python would have issues detecting the platform ran on, or path translation not working and so on. But apparently this all just works now.

It's been a while since I messed with MSYS2, both worth noting that CI was installing the "mingw" variant of Python which should be closest to native Windows build with the least unix-style path support (as opposed to the "msys2" variant which does Cygwin path translation.) But yeah, it just worked.

py/objfloat.c
projectgus: I don't know much attention Microsoft pay to C rather than C++ (I think the new expression is constant in C++). So I wouldn't bet money it gets fixed quickly. :laughing: 

This should be harmless if it does stay in too long, AFAIK it's a valid definition of NAN and covered by tests that will pick up if it ever stops being one for some unexpected reason.

dpgeorge:
> > The MSYS2 CPython 3.12.7 package prints some warnings about "Task was destroyed but it is pending!" on interpreter shutdown
> 
> To be clear: this also happens with the Python from `setup-python`? And we're sure it is in fact a Python issue/bug and not something in the test code?

I'm really not sure why MSYS CPython 3.12.7 prints this warning.  The test is quite simple and is running the asyncio event loop using `run_forever()`.  It uses `loop.stop()` to finish the loop, and `task` should be closed by that call even though `task` does not finish.  So maybe this CPython version is warning about that task being destroyed by `loop.stop()`??  That seems wrong though because isn't that what `loop.stop()` is intended to do, stop running tasks?

This test anyway has incompatibilities with CPython, which deprecated calling get_event_loop() when there is no current event loop.  So maybe we just make this test MicroPython-only, add the .exp file, and remove the CPython workaround at the top of this test altogether?

projectgus:
> So maybe this CPython version is warning about that task being destroyed by `loop.stop()`?? That seems wrong though because isn't that what `loop.stop()` is intended to do, stop running tasks?

This is a good question... and after reading the docs and CPython code more I now understand a bit better how stopping works in asyncio. Calling `loop.stop()` really only pauses the tasks, you can call `loop.run_forever()` again and the loop will resume where it was up to. So the tasks (as futures) keep all of their state at the point they are stopped.

There is [loop.close()](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.close) which does permanently close a stopped event loop. This clears the tasks from the event loop, but it also doesn't change anything about the tasks themselves - any running task is still a pending (i.e. not completed or cancelled) Future.

But it turns out that calling `loop.close()` in the test is the key to always seeing the warning on CPython. I still guesss it previously appears unpredictably during interpreter shutdown because of finalizer order (the loop finalizer calls self.close(), for example).

I've pushed a commit that adds `loop.close()` and then filters out the (now consistently generated) warning by setting an event loop exception handler.

> This test anyway has incompatibilities with CPython, which deprecated calling get_event_loop() when there is no current event loop. So maybe we just make this test MicroPython-only, add the .exp file, and remove the CPython workaround at the top of this test altogether?

Given how fiddly this stuff is I reckon it's worth maintaining tests that run the same code against CPython as much as we can. It helps us tease out little behavioural differences like this warning.

projectgus:
The macos failure in this CI run is fixed in #16304

stinos:
> I've pushed a commit that adds loop.close() and then filters out the (now consistently generated) warning by setting an event loop exception handler.

Sorry for being picky but is there really no way to make this code behave properly (well, according to CPython) such that the warning does not occur? From a quick look it seems it could be solved by having the task run another cycle to handle its cancellation. Or otherwise a somewhat more robust solution than compaing strings (i.e. `if context.get("message", "") == "Task was destroyed but it is pending!"`), checking for `asyncio.CancelledError` perhaps?

projectgus:
> Sorry for being picky but is there really no way to make this code behave properly (well, according to CPython) such that the warning does not occur? From a quick look it seems it could be solved by having the task run another cycle to handle its cancellation. Or otherwise a somewhat more robust solution than compaing strings (i.e. `if context.get("message", "") == "Task was destroyed but it is pending!"`), checking for `asyncio.CancelledError` perhaps?

It's OK, it's good to get to the bottom of understanding this.

An important point is that no task is ever cancelled in this test, by either `loop.stop()` or `loop.close()`. With `loop.stop()` the task stays pending and will resume immediately if you run the loop again. With `loop.cancel()` the loop goes into a state where the loop will never run again, but no task cancellation occurs either and the task object itself doesn't change.

To cancel a task then you have to explicitly call `task.cancel()`. It's also necessary to call `task.cancel()` before stopping the loop, as `task.cancel()` only sets a flag to inject a CancelledError the next time the task is run. If you cancel a task and never run the task again then you see the same warning ([More explanation here](https://xinhuang.github.io/posts/2017-07-31-common-mistakes-using-python3-asyncio.html#orgc09f339)).

If we were to change the test to cancel the task before closing the loop then we are no longer testing the same thing: this test (as I understand it) is that we can stop an event loop with active tasks on it and those tasks don't keep running on a new loop. If we were to change the test to cancel the tasks before stopping the loop then this is much better from a "best programming practices with async" perspective (which is why the warning exists in CPython and the blog post I linked above exists), but now we can't verify that the old loop has properly stopped running (as it has no tasks left to run). (We also already have other tasks covering async task cancellation.)

projectgus:
> more robust solution than compaing strings

If comparing strings is a sticking point then we could rewrite this to use the unittest module and actually test the behaviour rather than the output. But currently we have no way to verify a unittest also passes on CPython, and (as I said above) verifying compatible behaviour with CPython seems pretty important in these kind of fiddly corner cases.

dpgeorge:
Would it be possible to run the asyncio loop to completion (to make sure `task` finishes cleanly) at the end of the test, after testing the behaviour of `loop.stop()`?

projectgus:
> Would it be possible to run the asyncio loop to completion (to make sure `task` finishes cleanly) at the end of the test, after testing the behaviour of `loop.stop()`?

That makes sense to me, have pushed a commit to change this. Have kept it calling `loop.stop()` the second time, as `run_forever()` doesn't complete otherwise and it seems like a good way to test both possibilities for stopping and closing the loop.

dpgeorge:
> That makes sense to me, have pushed a commit to change this.

Oh actually I was thinking of something slightly different, if it would be possible to call `loop.run_forever()` twice to finish running `task`... but that won't test what this test is aiming to test, namely (as you state above) that it should test abruptly stopping a loop and then starting a new one without the old one finishing all its tasks.

How about this:
```diff
--- a/tests/extmod/asyncio_new_event_loop.py
+++ b/tests/extmod/asyncio_new_event_loop.py
@@ -17,6 +17,8 @@ async def task():
         print("task", i)
         await asyncio.sleep(0)
         await asyncio.sleep(0)
+    print("task stop")
+    loop.stop()
 
 
 async def main():
@@ -30,9 +32,17 @@ async def main():
 # Use default event loop to run some tasks
 loop = asyncio.get_event_loop()
 loop.create_task(main())
+print("run_forever")
 loop.run_forever()
+print("close")
+loop.close()
 
 # Create new event loop, old one should not keep running
 loop = asyncio.new_event_loop()
 loop.create_task(main())
+print("run_forever")
+loop.run_forever()
+print("run_forever")
 loop.run_forever()
+print("close")
+loop.close()
```
That still requires suppressing the error under CPython, but doesn't need `main` to `await t`.

Otherwise, `asyncio_loop_stop.py` is already testing `loop.stop()` behaviour so this `asyncio_new_event_loop.py` can just stay minimal and the only change in this PR would be to add `loop.close()` and suppress the CPython exception (and we fully understand why we need to suppress it, that it's a difference to MicroPython).

projectgus:
Oh, I see what you mean now.

> Otherwise, `asyncio_loop_stop.py` is already testing `loop.stop()` behaviour so this `asyncio_new_event_loop.py` can just stay minimal and the only change in this PR would be to add `loop.close()` and suppress the CPython exception (and we fully understand why we need to suppress it, that it's a difference to MicroPython).

OK, no worries. So I'll drop the last commit in this branch and we'll merge up to e08250e?

projectgus:
> > Otherwise, `asyncio_loop_stop.py` is already testing `loop.stop()` behaviour so this `asyncio_new_event_loop.py` can just stay minimal and the only change in this PR would be to add `loop.close()` and suppress the CPython exception (and we fully understand why we need to suppress it, that it's a difference to MicroPython).
> 
> OK, no worries. So I'll drop the last commit in this branch and we'll merge up to [e08250e](https://github.com/micropython/micropython/commit/e08250e9350cbcaa78ae7cd49daad4ad3988fee4)?

I went ahead and did this, with a rebase.

stinos:
Yes I understand, there doesn;t seem to be another way so should be good.


> If comparing strings is a sticking point

Given the circumstances it should be fine, it seems unlikely CPython is ever going to change that message text.

PR #16299, opened 2024-11-25
rp2/boards/SEEED_XIAO_RP2350: Add new SEEED_XIAO_RP2350 board definit…

Lesords:
### Summary

This PR adds initial support for the XIAO RP2350 board.

Depend on [this](https://github.com/raspberrypi/pico-sdk/pull/2098).

### Testing

- LED Blink
- RGB


Lesords:
Update: Made some changes to pass the commit message check.

dpgeorge:
Now that pico-sdk 2.1.0 is updated here, this board should work.

ports/rp2/boards/SEEED_XIAO_RP2350/pins.csv
```diff
@@ -0,0 +1,28 @@
...
+GP18,GPIO18
+GP19,GPIO19
+GP20,GPIO20
+GP21,GPIO21
+GP22,GPIO22
+GP25,GPIO25
+GP26,GPIO26
+GP27,GPIO27
+GP28,GPIO28
+LED,GPIO25
```
dpgeorge, line 28: Most of these pin names are wrong.

Please remove all the GPxx names because that's not the name printed on the PCB.

Instead, add the Dx names like:
```
D0,GPIO26
D1,GPIO27
```
And then also names for the `LED`, `LED_POWER` and `LED_RGB` connections.

Lesords: > Most of these pin names are wrong.
> 
> Please remove all the GPxx names because that's not the name printed on the PCB.
> 
> Instead, add the Dx names like:
> 
> ```
> D0,GPIO26
> D1,GPIO27
> ```
> 
> And then also names for the `LED`, `LED_POWER` and `LED_RGB` connections.

I'm very sorry, it looks like I made a stupid mistake.

I've made the changes now.

dpgeorge: That's OK, thanks for updating.

ports/rp2/boards/SEEED_XIAO_RP2350/pins.csv
```diff
@@ -0,0 +1,24 @@
...
+D12,GPIO20
+D13,GPIO17
+D14,GPIO16
+D15,GPIO11
+D16,GPIO12
+D17,GPIO10
+D18,GPIO9
+LED,GPIO25
+NEO_PWR,GPIO23
+NEOPIX,GPIO22
```
dpgeorge, line 22: I suggest calling these `NEOPIXEL_POWER` and `NEOPIXEL`.

That at least matches the M5STACK_NANOC6 board, and a lot of other boards also use `NEOPIXEL` for the Neopixel data pin.

Lesords: > I suggest calling these `NEOPIXEL_POWER` and `NEOPIXEL`.
> 
> That at least matches the M5STACK_NANOC6 board, and a lot of other boards also use `NEOPIXEL` for the Neopixel data pin.

I actually did it to be consistent with this schematic, look at the last page of this link
https://files.seeedstudio.com/wiki/XIAO-RP2350/res/Seeed-Studio-XIAO-RP2350-v1.0.pdf

I can also make this modification if necessary.

Lesords: > > I suggest calling these `NEOPIXEL_POWER` and `NEOPIXEL`.
> > That at least matches the M5STACK_NANOC6 board, and a lot of other boards also use `NEOPIXEL` for the Neopixel data pin.
> 
> I actually did it to be consistent with this schematic, look at the last page of this link https://files.seeedstudio.com/wiki/XIAO-RP2350/res/Seeed-Studio-XIAO-RP2350-v1.0.pdf
> 
> I can also make this modification if necessary.

It's been modified.

dpgeorge: > I actually did it to be consistent with this schematic

Oh, I see.

I guess either way is fine, following the schematics or following other boards.

> It's been modified.

OK, if you're happy to call it `NEOPIXEL` and `NEOPIXEL_POWER`, let's stick with that.  It makes it easier for users to discover, and then they don't even need to look at the schematics.

PR #16300, opened 2024-11-25
py/obj: Fix type mismatch when printing object.

mmilata:
### Summary

The `type` member of `mp_obj_type_t` is `uint16_t` while `mp_vprintf` expects `qstr` which is a typedef of `size_t`. This can lead to undefined behaviour. In my case, for qstr 0xaf4 the `va_arg` invocation returned 0x100000af4 which later lead to assertion failure when looking up the qstring.

### Testing

The bug manifests on arm64 Darwin with GCC 12 & GCC 13 on micropython-1.19 unix port when running something like `print("%s" % object)`. The commit fixes it on this configuration, and likely on git master as well.

### DIFF
--- a/py/obj.c
+++ b/py/obj.c
@@ -128,7 +128,7 @@ void mp_obj_print_helper(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t
     if (MP_OBJ_TYPE_HAS_SLOT(type, print)) {
         MP_OBJ_TYPE_GET_SLOT(type, print)((mp_print_t *)print, o_in, kind);
     } else {
-        mp_printf(print, "<%q>", type->name);
+        mp_printf(print, "<%q>", (qstr)type->name);
     }
 }
 


projectgus:
@mmilata Thanks for submitting this PR and pointing this out.

There are a lot of places in MicroPython where a `uint16_t` arg (aka `qstr_short_t`) is passed to `mp_printf()` for format specifier `%q`. If this is really a bug then it's going to be kind of annoying to fix.
 
I'm not certain that's the root cause, though. IIUC the bug as reported is that first a uint16_t is passed to a varargs function via:

```
        mp_printf(print, "<%q>", type->name);
```

and then the value is retrieved as a `qstr` type (aka `size_t`) in `mpprint.c`:

```
            case 'q': {
                qstr qst = va_arg(args, qstr);
```

... and that this is some kind of undefined behaviour that's causing the argument to contain junk upper bytes when retrieved by `var_arg` as the larger int.

As far as I understand the C standard, this should be valid C. From the C99 spec of `va_arg`:

> The parameter type shall be a type name specified such that the type of a pointer to an object that has the specified type
can be obtained simply by postfixing a * to type. If there is no actual next argument, or if type is not compatible with the type of the actual next argument (as promoted according to the default argument promotions), the behavior is undefined

So the `uint16_t` should be promoted to a `size_t` according to integer promotion rules for function arguments (i.e. the same as if the caller had called a function where the argument was `size_t` and passed a `uint16_t`). This is basically the same promotion that adding the cast on the calling side is doing in this PR.

I'm thinking the explanation is one of:

1) I don't understand the C standard as well as I think I do.
2) A compiler or libc bug. Seems unlikely, but a hassle if so.
3) Root cause is something other than what's explained above.

What do you think?

mmilata:
@projectgus I have missed the part about argument promotions - I believe your interpretation is correct and this should work. Given that GCC handles this fine on other platforms, and clang handles this fine on arm64 Darwin, I'm inclined to think this is a compiler/libc bug. Thank you for looking into this and feel free to close the PR.

mmilata:
Just tested with GCC 14 where this also works fine. Probably a compiler bug, looks vaguely similar to iains/gcc-darwin-arm64#18. Let's close this & sorry for the noise.

projectgus:
@mmilata No worries, thanks for following up!

PR #16304, opened 2024-11-26
ci: Remove macos install step for pkg-config

projectgus:
### Summary

* Fix for macos CI failures ([sample job](https://github.com/micropython/micropython/actions/runs/12023820688/job/33518713948))

* Brew renamed this package upstream and looks like it should be a smooth transition, except that the GitHub MacOS Runner also installs pkgconf and trying to install the old pkg-config on top of the new pkgconf causes it to fail.

* See https://github.com/actions/runner-images/pull/11015 for the runner change (and links from there for more detail).

* As pkg-config is already installed on GitHub macOS runners, there's no need to keep this step any more.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -651,9 +651,6 @@ function ci_unix_settrace_stackless_run_tests {
 }
 
 function ci_unix_macos_build {
-    # Install pkg-config to configure libffi paths.
-    brew install pkg-config
-
     make ${MAKEOPTS} -C mpy-cross
     make ${MAKEOPTS} -C ports/unix submodules
     #make ${MAKEOPTS} -C ports/unix deplibs


projectgus:
The windows failures in this CI run are fixed in #16272

dpgeorge:
> It looks like pkg-config comes pre-installed on GitHub MacOS runners for some time now. If we're not worried about other environments then we could remove this setup line entirely.

I reckon we should just do that, remove the brew line completely.

Other environments may have different needs anyway and so may already require additional set up steps.

projectgus:
Looks like brew may have done something to fix this problem anyway, as the macos builds have stopped failing (unless this is randomness because of different runner versions). Still, this PR may now be optional...

dpgeorge:
Even if CI is fixed, I reckon we should still merge this PR to simplify the set up (since the package is not needed explicitly).

dpgeorge (approved):

PR #11244, opened 2023-04-12
py/modsys.c: Add sys._exc_traceback.

DavidEGrayson:
This new function makes it easier to provide custom formatting of exception stack traces, for example to make them fit on a tiny 16x8-character display.  (Without this, I think you'd have to call `sys.print_exception`, somehow get the output as a string, and then scrape information from that string.)

I'm thinking of this as an experimental function that exposes internal details of MicroPython and therefore might change in the future.  That's why it has the underscore in the name.

Here is an example of its output:

    ['test.py', 4, 'foo', 'test.py', 7, '<module>']

The regular output from `sys.print_exception` for the same exception is:

```text
Traceback (most recent call last):
  File "test.py", line 7, in <module>
  File "test.py", line 4, in foo
ZeroDivisionError: divide by zero
```

Both outputs were generated by the following script:

```text
import sys

def foo():
  1/0

try:
  foo()
except Exception as e:
  sys.print_exception(e)
  print(sys._exc_traceback(e))
 ```

Is this the right approach for getting info about exception stack traces?   I'd be happy to change the name, change the output format, add tests, or add documentation.

PR #16311, opened 2024-11-27
py/objint,py/binary: Add int.to_bytes(signed) parameter, add common overflow checks.

projectgus:
### Summary

This PR is attempting to bring some CircuitPython changes upstream to MicroPython. From these PRs:
1. https://github.com/adafruit/circuitpython/pull/1860
2. https://github.com/adafruit/circuitpython/pull/1879
3.  https://github.com/adafruit/circuitpython/pull/2625

The changes:

* Implement overflow checks for buffer conversions, in a similar but more general way to #13087.
* Add optional `int.to_bytes(signed=bool)` parameter. This removes the CPython incompatibility where currently `int.to_bytes` in MicroPython behaves as if `signed=self < 0` when converting.

Note that this may require MicroPython code changes where it previously behaved differently to CPython:

* Anywhere that `int.to_bytes()` was being called on a negative number will now raise `OverflowError`. The fix is to add `signed=True`.
* Anywhere that `int.to_bytes()` was silently truncating the output will now raise `OverflowError`. The fix is either to ensure the length is sufficient for the value, or mask the value before converting it (i.e. `(x & 0xFFFF).to_bytes(2)`.

After cherry-picking commits directly from CircuitPython and resolving conflicts, some additional changes were needed (and in the end I've squashed most of them down into one):

* Update existing MicroPython tests to pass. Many tests are now simpler, as behaviour now closer to CPython. Some tests needed to be converted to unittest to verify both current and MP V2 behaviour is correct.
* Restore optional passing of arguments for `byteorder` and `length`, as per https://github.com/micropython/micropython/pull/14387 and CPython >=3.11.
* Do not raise `OverflowError` from array constructors if a value is out of range, truncate the value instead. This is a documented difference between CPython and MicroPython, and after discussing with @dpgeorge decided changing it here has the potential to break existing MP code. The overflow checks are kept if `MICROPY_PREVIEW_VERSION_2` is set, so that we can change behaviour to match CPython (and CircuitPython) in MicroPython 2. I guess Python code size may become a littler bigger if masking needs to be added, but it also eliminates a potential silent bug.
* Additional refactoring commit to merge similar code paths and bring the code size back down closer to current master.
* Double the allowed stack size on the Windows port. This seems necessary to allow debug builds to pass the new unit tests in this PR.
* Remove C function `mp_binary_set_val_array_from_int` - after the above refactors this was no longer used except in stm32/adc.c and a coverage test. Removing it reduces the stm32 port code size substantially.

### Testing

* [x] Unit tests passing on unix and rp2 ports.
* [x] Unit tests passing on Zephyr port (uses long long implementation for bigint).

* We already have some of this bounds checking functionality in master now, but the approach taken by CircuitPython is more general and has potential to apply the same checks in multiple places (i.e. array constructors) with no further impact on code size.


ports/unix/coverage.c
```diff
@@ -538,13 +538,32 @@ static mp_obj_t extra_coverage(void) {
...
+        int8_t i8ar[1];
+        uint8_t u8ar[1];
+        int16_t i16ar[1];
+        uint16_t u16ar[1];
+        int32_t i32ar[1];
+        uint32_t u32ar[1];
         mp_binary_set_val_array_from_int('f', far, 0, 123);
-        mp_printf(&mp_plat_print, "%.0f\n", (double)far[0]);
         mp_binary_set_val_array_from_int('d', dar, 0, 456);
+        mp_binary_set_val_array_from_int('b', i8ar, 0, -1);
```
projectgus, line 552: This function is only called from stm32/adc.c so an alternative fix for this would be to delete it from binary.h and binary.c and make it a static function in adc.c.

projectgus:
Only two things left here...

* Project coverage has somehow gone down even though the patch coverage is 100%. I've added some additional cove in coverage.c but it's still not quite enough.
* The Windows x64 Debug builds are failing in the uctypes_load_store test with `RuntimeError: maximum recursion depth exceeded`. I don't know if this is an actual stack usage problem,  a bug triggered when a unittest assert fails, or something else...

projectgus:
> * The Windows x64 Debug builds are failing in the uctypes_load_store test with `RuntimeError: maximum recursion depth exceeded`

Ah OK, there's a whole section about stack usage in the windows port README so I'm guessing we're just running out of stack.

projectgus:
> ```
>       stm32:  +156 +0.040% PYBV10
> ```

Had a quick look into why stm32 got bigger while all the other bare metal ports got smaller. Didn't find anything unusual: no significant new symbols are being linked compared to master. Several functions shrunk as they now call a common helper. I think maybe they just didn't shrink as much as on the other bare metal ports...

```diff
--- ./build-PYBV10-master/firmware.syms 2025-03-04 15:59:56.145991980 +1100
+++ ./build-PYBV10/firmware.syms        2025-03-04 16:00:08.656797386 +1100
@@ -5948,0 +5949 @@
+00000018 t allowed_args.0
@@ -6199 +6200 @@
-00000264 t compile_scope_inline_asm
+00000260 t compile_scope_inline_asm
@@ -6503 +6504 @@
-00000970 t emit_inline_thumb_op
+00000978 t emit_inline_thumb_op
@@ -7326 +7327 @@
-000000d8 t int_to_bytes
+0000005c t int_to_bytes
@@ -7770,3 +7771,3 @@
-0000002a T mp_binary_set_int
-00000114 T mp_binary_set_val
-0000008c T mp_binary_set_val_array
+0000005a T mp_binary_set_int
+000000f2 T mp_binary_set_val
+0000006e T mp_binary_set_val_array
@@ -8329 +8330 @@
-00000042 T mp_obj_exception_attr
+00000040 T mp_obj_exception_attr
@@ -8373 +8374 @@
-00000016 T mp_obj_int_to_bytes_impl
+00000114 T mp_obj_int_to_bytes
@@ -8521 +8522 @@
-0000091c T mp_qstr_const_hashes
+00000922 T mp_qstr_const_hashes
@@ -8523 +8524 @@
-0000048e T mp_qstr_const_lengths
+00000491 T mp_qstr_const_lengths
@@ -8525 +8526 @@
-00001250 T mp_qstr_const_pool
+0000125c T mp_qstr_const_pool
@@ -8891 +8892 @@
-000000d2 T mpz_as_bytes
+0000009e T mpz_as_bytes
@@ -9768 +9769 @@
-00000084 t set_aligned
+00000080 t set_aligned
@@ -10078 +10079 @@
-0000007c t task_attr
+00000078 t task_attr
@@ -10216 +10217 @@
-00000200 t uctypes_struct_attr_op
+00000204 t uctypes_struct_attr_op
```

(Files generated with `nm --print-size firmware.elf | cut -d' ' -f2- | tee firmware.syms`)

I don't see much we can do to improve this. Interestingly, the firmware size increase on PYBV10 is about the same if LTO is enabled (although the firmwares are overall smaller.)

dpgeorge:
> Had a quick look into why stm32 got bigger while all the other bare metal ports got smaller. Didn't find anything unusual

What about:
```diff
-00000016 T mp_obj_int_to_bytes_impl
+00000114 T mp_obj_int_to_bytes
```
That's 254 extra bytes.  Is that similar on other ports that did decrease?

projectgus:
> That's 254 extra bytes. Is that similar on other ports that did decrease?

I think so, this is the common function that the others now call. They grow by basically the same amount on both samd and rp2 ports. Here's samd:

```diff
@@ -6342 +6338 @@
-00000016 T mp_obj_int_to_bytes_impl
+00000114 T mp_obj_int_to_bytes
```

<details><summary>Full size diffs</summary>

```diff
--- build-ADAFRUIT_ITSYBITSY_M4_EXPRESS-master/firmware.syms    2025-03-04 17:06:29.044966391 +1100
+++ build-ADAFRUIT_ITSYBITSY_M4_EXPRESS/firmware.syms   2025-03-04 17:07:24.983870265 +1100
@@ -993 +992,0 @@
-t $d
@@ -4381,3 +4379,0 @@
-t $t
-t $t
-t $t
@@ -4404,0 +4401 @@
+00000018 t allowed_args.0
@@ -4620 +4617 @@
-00000260 t compile_scope_inline_asm
+00000264 t compile_scope_inline_asm
@@ -4822 +4819 @@
-00000968 t emit_inline_thumb_op
+00000970 t emit_inline_thumb_op
@@ -5312 +5309 @@
-000000d8 t int_to_bytes
+0000005c t int_to_bytes
@@ -5758,4 +5755,3 @@
-0000002a T mp_binary_set_int
-0000016c T mp_binary_set_val
-0000008c T mp_binary_set_val_array
-000000f2 T mp_binary_set_val_array_from_int
+0000005a T mp_binary_set_int
+00000150 T mp_binary_set_val
+0000006e T mp_binary_set_val_array
@@ -5777 +5773 @@
-0000008c t mp_builtin_compile
+00000088 t mp_builtin_compile
@@ -6300 +6296 @@
-0000003a T mp_obj_exception_attr
+00000038 T mp_obj_exception_attr
@@ -6342 +6338 @@
-00000016 T mp_obj_int_to_bytes_impl
+00000114 T mp_obj_int_to_bytes
@@ -6477 +6473 @@
-00000542 T mp_qstr_const_hashes
+00000536 T mp_qstr_const_hashes
@@ -6479 +6475 @@
-000002a1 T mp_qstr_const_lengths
+0000029b T mp_qstr_const_lengths
@@ -6481 +6477 @@
-00000a9c T mp_qstr_const_pool
+00000a84 T mp_qstr_const_pool
@@ -6802 +6798 @@
-000000d2 T mpz_as_bytes
+0000009e T mpz_as_bytes
@@ -7172 +7168 @@
-00000084 t set_aligned
+00000080 t set_aligned
@@ -7245 +7241 @@
-00000034 t slice_attr
+00000030 t slice_attr
@@ -7402 +7398 @@
-00000070 t task_attr
+0000007c t task_attr
@@ -7553 +7549 @@
-00000200 t uctypes_struct_attr_op
+00000204 t uctypes_struct_attr_op
@@ -7599 +7595 @@
-00000038 t usb_device_attr
+00000034 t usb_device_attr
```

```diff
--- build-RPI_PICO-master/firmware.syms        2025-03-04 17:01:17.898379768 +1100
+++ build-RPI_PICO/firmware.syms 2025-03-04 17:01:28.760234428 +1100
@@ -31,4 +30,0 @@
-t $d
-t $d
-t $d
-t $d
@@ -244 +239,0 @@
-t $d
@@ -630,0 +626,2 @@
+t $d
+r $d
@@ -1202,0 +1200 @@
+t $d
@@ -1536,0 +1535 @@
+t $d
@@ -5462,3 +5460,0 @@
-t $t
-t $t
-t $t
@@ -5503,0 +5500 @@
+00000018 r allowed_args.0
@@ -5826 +5823 @@
-00000278 t compile_scope_inline_asm
+00000274 t compile_scope_inline_asm
@@ -6144 +6141 @@
-00000498 t emit_inline_thumb_op
+00000494 t emit_inline_thumb_op
@@ -6781 +6778 @@
-000000d4 t int_to_bytes
+00000060 t int_to_bytes
@@ -6786 +6783 @@
-00000044 t iobase_ioctl
+00000040 t iobase_ioctl
@@ -7352,4 +7349,3 @@
-0000002a T mp_binary_set_int
-00000154 T mp_binary_set_val
-00000084 T mp_binary_set_val_array
-00000084 T mp_binary_set_val_array_from_int
+00000056 T mp_binary_set_int
+0000012e T mp_binary_set_val
+00000064 T mp_binary_set_val_array
@@ -7370 +7366 @@
-00000064 t mp_builtin_compile
+00000068 t mp_builtin_compile
@@ -7885 +7881 @@
-0000005c T mp_native_type_from_qstr
+00000064 T mp_native_type_from_qstr
@@ -7962 +7958 @@
-00000016 T mp_obj_int_to_bytes_impl
+00000110 T mp_obj_int_to_bytes
@@ -8109 +8105 @@
-000002ee R mp_qstr_const_hashes
+000002f1 R mp_qstr_const_hashes
@@ -8111 +8107 @@
-000002ee R mp_qstr_const_lengths
+000002f1 R mp_qstr_const_lengths
@@ -8113 +8109 @@
-00000bd0 R mp_qstr_const_pool
+00000bdc R mp_qstr_const_pool
@@ -8355 +8351 @@
-00000064 T mp_vfs_blockdev_init
+00000060 T mp_vfs_blockdev_init
@@ -8365 +8361 @@
-00000068 T mp_vfs_getcwd
+0000006c T mp_vfs_getcwd
@@ -8367,2 +8363,2 @@
-00000060 T mp_vfs_ilistdir
-00000070 t mp_vfs_ilistdir_it_iternext
+00000064 T mp_vfs_ilistdir
+00000074 t mp_vfs_ilistdir_it_iternext
@@ -8370 +8366 @@
-0000007c T mp_vfs_import_stat
+0000007a T mp_vfs_import_stat
@@ -8433 +8429 @@
-0000002e T mp_vfs_rename
+00000030 T mp_vfs_rename
@@ -8435 +8431 @@
-00000016 T mp_vfs_rmdir
+00000018 T mp_vfs_rmdir
@@ -8437 +8433 @@
-00000040 T mp_vfs_stat
+0000003c T mp_vfs_stat
@@ -8439 +8435 @@
-00000054 T mp_vfs_statvfs
+00000058 T mp_vfs_statvfs
@@ -8441 +8437 @@
-0000009c T mp_vfs_umount
+000000a0 T mp_vfs_umount
@@ -8450 +8446 @@
-00000140 T mpz_as_bytes
+000000ea T mpz_as_bytes
@@ -9025 +9021 @@
-00000084 t set_aligned
+0000007c t set_aligned
@@ -9102 +9098 @@
-00000038 t slice_attr
+00000034 t slice_attr
```

Ah! But the very big difference on these ports is that, unlike stm32, the `mp_binary_set_val_array_from_int` function is no longer linked in this PR! That's about 127 bytes.

That's the function which is still called from stm32/adc.c . So if we can find another way to implement that, the code would go down on stm32 (and we also wouldn't need the changes in #16858).

projectgus:
EDIT: Accidentally buried the useful info in the `<details>` tag above.

projectgus:
> That's the function which is still called from stm32/adc.c . So if we can find another way to implement that, the code would go down on stm32 (and we also wouldn't need the changes in #16858).

Specifically [pyb.ADC.read_timed and read_timed_multi](https://docs.micropython.org/en/latest/library/pyb.ADC.html#pyb.ADC.read_timed) are documented as supporting either an array of 8/16-bit integers or a bytearray.

`mp_binary_set_val_array_from_int` supports any numeric typecode. So I think there's potentially a much simpler version we can embed into adc.c and remove the other function entirely. But probably best to do in a separate PR.

py/binary.c
```diff
@@ -478,16 +483,12 @@ void mp_binary_set_val_array(char typecode, void *p, size_t index, mp_obj_t val_
...
-                mp_obj_int_to_bytes_impl(val_in, MP_ENDIANNESS_BIG,
-                    size, (uint8_t *)p + index * size);
-                return;
-            }
-            #endif
-            mp_binary_set_val_array_from_int(typecode, p, index, mp_obj_get_int(val_in));
+        default: {
+            size_t size = mp_binary_get_size('@', typecode, NULL);
+            p = (uint8_t *)p + index * size;
+            mp_obj_int_to_bytes(val_in, size, p, MP_ENDIANNESS_BIG, is_signed(typecode), OVERFLOW_CHECKS);
```
projectgus, line 489: This endianness choice seems wrong, though, at least `mp_binary_set_val_array_from_int` would have used native endianness...? Need to make sure this is being tested.

projectgus: This is fine, I misread the meaning of `MP_ENDIANNESS_BIG` here as being an argument to the function, but the argument is `bool big_endian` and the value of `MP_ENDIANNESS_BIG` macro is 0 on a little endian port.

IhorNehrutsa:
Are there any implementation plans?

py/objint.c
```diff
@@ -307,6 +312,185 @@ char *mp_obj_int_formatted(char **buf, size_t *buf_size, size_t *fmt_size, mp_co
     return b;
 }
 
+#if MICROPY_LONGINT_IMPL == MICROPY_LONGINT_IMPL_MPZ
+
+static void mp_obj_int_buffer_overflow_check(mp_obj_t self_in, size_t nbytes, bool is_signed) {
+    if (is_signed) {
+        // edge = 1 << (nbytes * 8 - 1)
+        mp_obj_t edge = mp_binary_op(MP_BINARY_OP_INPLACE_LSHIFT,
```
projectgus: Something of note is that this overflow check allocates two mpz integers, which might be too expensive?

I think we can write this in a less memory-intensive way, but it'll be a bit more code size.

projectgus: Rewrote this to not allocate, and it also seems to be similar to this version on code size (better on some ports).

projectgus:
The "optimised" implementation of `mp_obj_int_buffer_overflow_check` for MPZ is really hacky. Need to look at it with fresh eyes, I think.

py/binary.c
```diff
@@ -488,67 +494,12 @@ void mp_binary_set_val_array(char typecode, void *p, size_t index, mp_obj_t val_
...
-        #if MICROPY_PY_STRUCT_UNSAFE_TYPECODES
-        case 'P':
-            ((void **)p)[index] = (void *)(uintptr_t)val;
-            break;
-        #endif
+        // In all remaining cases the type code is an integer
+        default: {
+            size_t size = mp_binary_get_size('@', typecode, NULL);
+            p = (uint8_t *)p + index * size;
+            mp_obj_int_to_bytes(val_in, size, p, MP_ENDIANNESS_BIG, is_signed(typecode), OVERFLOW_CHECKS);
```
projectgus: Did some benchmarking of the cost to removing the switch/case table for setting integers in the array and instead delegating to `mp_obj_int_to_bytes()`:

<details>
<summary>Test code & results</summary>

```py
import micropython
import array, gc, time

EXTEND_LIST = [ 1 ] * 8192

def test_extend(arr):
    a = time.ticks_us()
    arr.extend(EXTEND_LIST)
    b = time.ticks_us()
    print("Array of", typecode, "extended to length", len(arr), "in", time.ticks_diff(b, a), "us")

def test_subscr_set(arr):
    a = time.ticks_us()
    for i in range(len(arr)):
        arr[i] = 3
    b = time.ticks_us()
    print("Array of", typecode, "called set", len(arr), "times in", time.ticks_diff(b, a), "us")

@micropython.native
def test_native_subscr_set(arr):
    a = time.ticks_us()
    for i in range(len(arr)):
        arr[i] = 2
    b = time.ticks_us()
    print("Array of", typecode, "called native set", len(arr), "times in", time.ticks_diff(b, a), "us")

for typecode in ('b', 'i', 'q', 'Q'):
    arr = array.array(typecode)
    gc.collect()
    test_extend(arr)
    test_subscr_set(arr)
    test_native_subscr_set(arr)
```

```
unix, master

Array of b extended to length 8192 in 177 us
Array of b called set 8192 times in 175 us
Array of b called native set 8192 times in 103 us
Array of i extended to length 8192 in 333 us
Array of i called set 8192 times in 178 us
Array of i called native set 8192 times in 104 us
Array of q extended to length 8192 in 533 us
Array of q called set 8192 times in 179 us
Array of q called native set 8192 times in 105 us
Array of Q extended to length 8192 in 535 us
Array of Q called set 8192 times in 176 us
Array of Q called native set 8192 times in 103 us

rp2, master

Array of b extended to length 8192 in 82713 us
Array of b called set 8192 times in 53549 us
Array of b called native set 8192 times in 34465 us
Array of i extended to length 8192 in 209051 us
Array of i called set 8192 times in 54763 us
Array of i called native set 8192 times in 34524 us
Array of q extended to length 8192 in 379239 us
Array of q called set 8192 times in 54068 us
Array of q called native set 8192 times in 34714 us
Array of Q extended to length 8192 in 380521 us
Array of Q called set 8192 times in 54321 us
Array of Q called native set 8192 times in 35111 us

unix, branch

Array of b extended to length 8192 in 186 us
Array of b called set 8192 times in 181 us
Array of b called native set 8192 times in 110 us
Array of i extended to length 8192 in 327 us
Array of i called set 8192 times in 188 us
Array of i called native set 8192 times in 110 us
Array of q extended to length 8192 in 519 us
Array of q called set 8192 times in 181 us
Array of q called native set 8192 times in 118 us
Array of Q extended to length 8192 in 549 us
Array of Q called set 8192 times in 182 us
Array of Q called native set 8192 times in 112 us

(these are within the margin of error for re-runs of the process)

rp2, branch

Array of b extended to length 8192 in 97745 us (+18%)
Array of b called set 8192 times in 60576 us (+13%)
Array of b called native set 8192 times in 41398 us (+20%)
Array of i extended to length 8192 in 225731 us (+7%)
Array of i called set 8192 times in 61302 us (+11%)
Array of i called native set 8192 times in 42186 us (+22%)
Array of q extended to length 8192 in 399358 us (+5%)
Array of q called set 8192 times in 65175 us (+20%)
Array of q called native set 8192 times in 45983 us (+32%)
Array of Q extended to length 8192 in 397886 us (+4%)
Array of Q called set 8192 times in 65501 us (+20%)
Array of Q called native set 8192 times in 46012 us (+31%)
```

</details>

Unix port is within the margin of error for multiple runs, but rp2 port (RPI_PICO_W):

- Extend is 8.5% slower on average
- Setting array from bytecode is 8.25% slower on average
- Setting array from native emitter is 26% slower on average!

The size cost on rp2 (without overflow checks) is only +192 bytes, so putting this back in.

dpgeorge: Oh wow!  That's a big drop in performance.  Thanks for testing.

projectgus: I thought it was interesting there was no measurable difference for Unix port running on my deskop. I guess all of the related code probably fits in L1 cache.

projectgus:
I think this is ready for review:

* Now the array setter routines are back there's a code size impact for most ports. Impact is still bigger on the really minimal ports, I think the reason is that the code size benefits were achieved by combining some code paths in modules/features that the minimal ports don't have enabled (like mpz or the array module).
* Commit history could probably be squashed a bit. I've tried to keep distinct changes separate and also keep the original CircuitPython commits as-is, but possibly the follow-up commits could be mostly combined.

@IhorNehrutsa If you have time then I'd be interested to know what you think, as you've been looking at the same feature in #16798.

projectgus:
(Pushed one more change to restore fast path in `mp_binary_set_val()`. It feels like `mp_binary_set_val()` and `mp_binary_set_val_array()` are similar enough they could be combined further, with the limitation (I think) being the latter requires `dest` is aligned.

dpgeorge:
@projectgus : via octoprobe, it looks like some of the updated tests here are failing due to MemoryError on targets with a small amount of memory (when run with the native emitter).  I'm not sure if we should try to solve that here, but it will need to be fixed at some point.

IhorNehrutsa:
```
>>> x=-2
>>> x.to_bytes(signed=True)
b'\xfe'
>>> x.to_bytes()
OverflowError: value would overflow a 1 byte buffer
>>> 
```
Is this normal behavior?

dpgeorge:
> Is this normal behavior?

Yes.  Although the error message is different to CPython, the exception type is the same and the behaviour is the same.

py/objint.c
```diff
@@ -307,6 +312,200 @@ char *mp_obj_int_formatted(char **buf, size_t *buf_size, size_t *fmt_size, mp_co
...
+            return;
+        }
+    }
+
+    // Corner case: Do a dry-run conversion of MPZ to bytes with no output buffer, and raise on failure
+    // (note: it's important to do this with no buffer as the destination buffer may be pre-allocated.)
+    //
+    // The two cases this catches are: room for correct sign bit, and the case where the top-most
+    // digit of mpz has upper bytes which are 0s.
+    if (mpz_as_bytes(mpz, false, is_signed, nbytes, NULL)) {
```
dpgeorge: It feels wrong that `mpz_as_bytes()` must be called twice if this path is hit (once here to check, and then a second time to do the conversion).

In fact, I think this entire function can be removed and just rely on the return value of `mpz_as_bytes()` to tell whether or not the conversion fit.  The only extra case to handle is `!is_signed && mpz->neg`, which needs to raise.

Something like this below in `mp_obj_int_to_bytes()`:
```c
        bool ok = false;
        if (!overflow_check || is_signed || !self->mpz.neg) {
            ok = mpz_as_bytes(&self->mpz, big_endian, is_signed, buf_len, buf);
        }
        if (overflow_check && !ok) {
            raise_to_bytes_overflow_error(buf_len);
        }
```
That should simplify things a fair bit, and `mpz_as_bytes()` can remove the code that allows `buf` to be NULL.

py/objint.c
```diff
@@ -307,6 +312,200 @@ char *mp_obj_int_formatted(char **buf, size_t *buf_size, size_t *fmt_size, mp_co
...
+        // Fall through to raise.
+    }
+
+raise:
+    raise_to_bytes_overflow_error(nbytes);
+}
+
+#endif // MICROPY_LONGINT_IMPL == MICROPY_LONGINT_IMPL_LONGLONG
+
+void mp_obj_int_to_bytes(mp_obj_t self_in, size_t buf_len, byte *buf, bool big_endian, bool is_signed, bool overflow_check) {
```
dpgeorge: This function has quite a bit of dependency on the int implementation (mpz, longlong), and maybe it and all its static helper functions that are in this file would have a better home in their respective `objint_mpz.c` / `objint_longlong.c` file?

That may give a little bit of code duplication (the small-int handling bits at the end of this function), but to fix that the small-int code could possibly be a `static inline` helper in `objint.h`?

py/objint.c
projectgus: Agree this would be a lot simpler.

Just to be clear on the trade-off: The reason for calling twice is that there are corner cases where this code path gets hit with a pre-existing buffer, via binary.c `mp_binary_set_val()` and `mp_binary_set_val_array()` when the source is a big int and the destination is an existing array or similar. (EDIT: clarified src/dest arguments.)

Without calling `mpz_as_bytes()` twice, there are failure paths where the destination buffer (i.e. array entry) gets written with a value but the function then raises `OverflowError`. (I don't think I actually have test coverage for this, should either add coverage for it or add coverage which helps document the difference depending on what we choose.)

There is another way we could handle this case, which would be for `buf_len <= 8` we initially call `mpz_as_bytes()` to write into a buffer on the stack and then copy it to `buf` following success. I don't know if this would be less or more complexity than supporting `buf == NULL`.

projectgus:
> Yes. Although the error message is different to CPython, the exception type is the same and the behaviour is the same.

Hmm. This exception is the most visible "breaking change" for MicroPython code which worked before this PR, so maybe we can add an explanatory error message similar to CPython (as this one is a bit obscure).

py/objint.c
projectgus: > There is another way we could handle this case, which would be for buf_len <= 8 we initially call mpz_as_bytes() to write into a buffer on the stack and then copy it to buf following success. I don't know if this would be less or more complexity than supporting buf == NULL.

Oh, of course - we can do this in the caller (binary.c), that way it doesn't clutter up any of the code paths where it doesn't matter.

dpgeorge:
> This exception is the most visible "breaking change" for MicroPython code which worked before this PR,

I don't follow.  In current master we get:
```
>>> x=-2
>>> x.to_bytes(signed=True)
TypeError: function doesn't take keyword arguments
```
So this never worked.  Or do you mean that `x.to_bytes()` works on current master and this PR makes that not work (now matching CPython)?

py/objint.c
dpgeorge: > Just to be clear on the trade-off: The reason for calling twice is that there are corner cases where this code path gets hit with a pre-existing buffer, via binary.c `mp_binary_set_val()` and `mp_binary_set_val_array()` when the source is a big int and the destination is an existing array or similar.

Ah, right.  I didn't realise that.  I guess something like:
```py
import array
a = array.array('H', [0])
a[0] = 12345678
```
As you say, we should definitely add a test for that (cpydiff or otherwise).

projectgus:
> Or do you mean that `x.to_bytes()` works on current master and this PR makes that not work (now matching CPython)?

That's the one. `(-2).to_bytes()` will start raising an exception with this change.

py/objint.c
projectgus: Rewrote along these lines, it's a lot cleaner. Added an `objint_impl.h` to hold the inline functions for small int conversions.

py/objint.c
projectgus: Moved the temporary value buffer into binary.c and expanded the test coverage to verify that the value previously in the array is kept after `OverflowErrror` is raised.

extmod/moductypes.c
```diff
@@ -463,8 +463,8 @@ static mp_obj_t uctypes_struct_attr_op(mp_obj_t self_in, qstr attr, mp_obj_t set
                 if (self->flags == LAYOUT_NATIVE) {
                     set_aligned_basic(val_type & 6, self->addr + offset, val);
                 } else {
-                    mp_binary_set_int(GET_SCALAR_SIZE(val_type & 7), self->flags == LAYOUT_BIG_ENDIAN,
-                        self->addr + offset, val);
+                    size_t item_size = GET_SCALAR_SIZE(val_type & 7);
+                    mp_binary_set_int(item_size, self->addr + offset, item_size, val, self->flags == LAYOUT_BIG_ENDIAN);
```
dpgeorge, line 467: As I understand it, this is a breaking change in behaviour because it can now raise if `val` doesn't fit.

Is it worth retaining existing behaviour simply by masking `val` as appropriate so that it always fits?  (Then the `uctypes_array_load_store.py` test does not need to change.)

projectgus: Not quite, `mp_binary_set_int()` doesn't ever raise - the API just changed a little so it could be called from more places.

The changes in `uctypes_array_load_store.py` are to account for bytearray setters gaining overflow checks in MicroPython V2.0 (because moductypes creates bytearrays internally).

dpgeorge: > Not quite, `mp_binary_set_int()` doesn't ever raise - the API just changed a little so it could be called from more places.

Ah, I see.



> The changes in `uctypes_array_load_store.py` are to account for bytearray setters gaining overflow checks in MicroPython V2.0 (because moductypes creates bytearrays internally).

Hmm, I don't see the original version of that test storing into a bytearray?  It creates a bytearray only as a backing buffer for the uctypes struct.

dpgeorge:
Following up on octoprobe failures: I ran the current version of this PR on ADAFRUIT_ITSYBITSY_M0_EXPRESS and there are 3 new failures introduced by this PR.

```
$ ./run-tests.py -t a0 --via-mpy --emit native
```

#### FAIL  `basics/array_int_repr.py`
```
  File "<stdin>", line 37, in <module>
  File "unittest/__init__.py", line 70, in <module>
  File "unittest/__init__.py", line 205, in TestCase
MemoryError: memory allocation failed, allocating 232 bytes
```

#### FAIL  `basics/array_limits_intbig.py`
```
  File "<stdin>", line 37, in <module>
MemoryError: memory allocation failed, allocating 160 bytes
```
#### FAIL  `extmod/uctypes_array_load_store.py`
```
  File "<stdin>", line 37, in <module>
MemoryError: memory allocation failed, allocating 302 bytes
```

extmod/moductypes.c
projectgus: Hmm, yeah... I would classify my confidence about moductypes internals as "very low" but my understanding is that the relevant code path is here:
https://github.com/micropython/micropython/blob/master/extmod/moductypes.c#L502

Can see it by doing something like this:

```
>>> b = bytearray(5)
>>> desc = {"arr": (uctypes.ARRAY | 0, uctypes.UINT8 | 5)}
>>> s = uctypes.struct(uctypes.addressof(b), desc, uctypes.LITTLE_ENDIAN)
>>> type(s.arr)
<class 'bytearray'>
>>> desc = {"arr": (uctypes.ARRAY | 0, uctypes.INT8 | 5)}
>>> s = uctypes.struct(uctypes.addressof(b), desc, uctypes.LITTLE_ENDIAN)
>>> type(s.arr)
<class 'struct'>
```

Which is why the `is_v2` differences in the test are all gated behind `if is_v2 and item_sz == 1`. (I'm not actually sure why the test passes here for `INT8` when expecting the OverflowError, maybe all of the test values are in-range for INT8 but not UINT8...?)

This is still a weird change I admit, because other kinds of moductypes arrays won't add overflow checks in V2 but UINT8 will...

So can think of two possible ways forward:

1. Change the test to explicitly check if `isinstance(s.arr, bytearray)` which will make the reason for the special case in the test more obvious.
2. Remove the moductypes optimisation of treating `UINT8` specially as a bytearray, so you always get back a `struct` object of some kind. I'm not sure what the other trade-offs of this might be.

Any preference?

projectgus:
> Following up on octoprobe failures: I ran the current version of this PR on ADAFRUIT_ITSYBITSY_M0_EXPRESS and there are 3 new failures introduced by this PR.

Thanks @dpgeorge. So if I understand correctly these are all running out of memory after the test module starts executing but before any test case runs?

dpgeorge:
> So if I understand correctly these are all running out of memory after the test module starts executing but before any test case runs?

Yes, I think that's the case.  Although I'm not 100% certain because it's hard to tell with the native emitter due to limited traceback info.  I'll see if I can dig deeper to see exactly which allocation is failing.

extmod/moductypes.c
dpgeorge: Interesting.  In your snip above the second one (INT8) is *not* a bytearray, so that's why the check is not done.

If you modify `extmod/uctypes_array_load_store.py` to add `print(s.arr)` just before the inner for loop, you'll see that everything *except* UINT8 are internal uctypes struct's, but the UINT8 case is a `bytearray`.

Hmm...

dpgeorge: Ok, commit 6d287a6a029df0d4b7a2ab8bdd73be77c0c04317 tells the story.

dpgeorge: > 2\. Remove the moductypes optimisation of treating `UINT8` specially as a bytearray, so you always get back a `struct` object of some kind. I'm not sure what the other trade-offs of this might be.

That would make things cleaner/simpler, and not give any regressions wrt overflow checks.

But, it would probably give regressions in other areas, eg `uctypes.struct` can't be sliced, iterated, etc.  Whereas `bytearray` can be.

Could we return a `memoryview` instead of `bytearray`?  Or does `memoryview` also do overflow checking with this PR?

> 1. Change the test to explicitly check if `isinstance(s.arr, bytearray)` which will make the reason for the special case in the test more obvious.

I think at the very least do this option.  Then it's clear what's going on.

projectgus: > Interesting. In your snip above the second one (INT8) is not a bytearray, so that's why the check is not done.

Yes, sorry that's what I meant - the part I didn't get is why `item_sz == 1` is a sufficient check given it matches both INT8 and UINT8 and they're different types.

I get it now, it's because the range of values used in each test is `[-2, -1, 0, 1, 2]`. So they're always in-range for all signed types, meaning that the full check `if is_v2 and item_sz == 1 and (n < item_min or n > item_max):` only hits the special-case for `UINT8`.

So, I can definitely rewrite the test to make the behaviour clearer. But that still depends on keeping the behaviour that an array of `UINT8` is a `bytearray` type not the same struct type as the other arrays.

projectgus: To reword the options I can think of:

1. Change the test to explicitly check if `isinstance(s.arr, bytearray)` which will make the reason for the special case in the test more obvious. MicroPython V2 will have a breaking change for moductypes code that uses arrays of `UINT8` and relies on implicit truncation, same as any other code which uses bytearray and relies on implicit truncation.
2. Remove the V2 bounds check for bytearray setters (making it different to the `array` module). Don't recommend this for the reasons given in the PR description (possibility for silent bugs, difference to CPython).
3. Remove the moductypes optimisation of having a UINT8 array be a bytearray, so the `arr` element always gives back a struct object of some kind. This will keep all moductypes behaviour the same in MicroPython V2. I'm not sure what the other trade-offs of not returning a `bytearray` in this case would be.

projectgus: Oops, our comments overlapped. I'll do option 1 for now, we can always rethink the V2 behaviour change later.

> Could we return a memoryview instead of bytearray? Or does memoryview also do overflow checking with this PR?

With V2 it will, yes. Same code path.

projectgus: Have pushed a new test which makes the reason for the behaviour more explicit.

dpgeorge:
> I'll see if I can dig deeper to see exactly which allocation is failing.

I looked deeper into the above 3 failing tests.  They all fail trying to `import unittest`.

I added a `gc.collect()` and `micropython.mem_info()` to all 3 tests just before they `import unittest`.  And then printed a message after `import unittest`.  I get (using #19329):
```
$ ./run-tests.py -t a0 --via-mpy --emit native --trace-output basics/array_int_repr.py basics/array_limits_intbig.py extmod/uctypes_array_load_store.py
platform=samd arch=armv6m inlineasm=thumb float=32-bit
TRACE: tests/basics/array_int_repr.py
START TEST
stack: 1276 out of 7168
GC: total: 21248, used: 10272, free: 10976
 No. of 1-blocks: 57, 2-blocks: 11, max blk sz: 229, max free sz: 324
  File "<stdin>", line 37, in <module>
  File "unittest/__init__.py", line 70, in <module>
  File "unittest/__init__.py", line 169, in TestCase
MemoryError: memory allocation failed, allocating 184 bytes
FAIL  tests/basics/array_int_repr.py
NOTE: tests/basics/array_int_repr.py may be a unittest that doesn't run unittest.main()
TRACE: tests/basics/array_limits_intbig.py
START TEST
stack: 1276 out of 7168
GC: total: 21248, used: 16080, free: 5168
 No. of 1-blocks: 73, 2-blocks: 18, max blk sz: 395, max free sz: 34
  File "<stdin>", line 37, in <module>
MemoryError: memory allocation failed, allocating 80 bytes
FAIL  tests/basics/array_limits_intbig.py
NOTE: tests/basics/array_limits_intbig.py may be a unittest that doesn't run unittest.main()
TRACE: tests/extmod/uctypes_array_load_store.py
START TEST
stack: 1276 out of 7168
GC: total: 21248, used: 12400, free: 8848
 No. of 1-blocks: 57, 2-blocks: 14, max blk sz: 295, max free sz: 152
  File "<stdin>", line 37, in <module>
MemoryError: memory allocation failed, allocating 302 bytes
FAIL  tests/extmod/uctypes_array_load_store.py
3 tests performed (34 individual testcases)
0 tests passed
3 tests failed: tests/basics/array_int_repr.py tests/basics/array_limits_intbig.py tests/extmod/uctypes_array_load_store.py
```
The message after `import unittest` never prints, so they all fail on that import.

`unittest` itself requires just a little over 10k to import, and at the point of import the above tests all have 10k or less free.  So that's why the import -- and hence the test -- fails.

There are some small optimisations we can make to `unittest` to get it down in size, but that'll probably only help `basics/array_int_repr.py` which is just on the edge of having enough heap memory.  The other 2 tests just cannot run on this device with 20k total heap.

tests/extmod/uctypes_array_load_store.py
```diff
@@ -1,4 +1,10 @@
 # Test uctypes array, load and store, with array size > 1
+try:
+    import unittest
+except (SyntaxError, ImportError):
+    # minimal unix variant can't import unittest
```
dpgeorge: This way of skipping is not used elsewhere.  We usually require unittest to be importable (so this is the first test that is picked up by a target that cannot run unittest).

Instead of having the logic here (which could be a little brittle, eg it'll accidentally skip on a target that forgot to install unittest) how about adding an exclusion like this:
```diff
--- a/tests/run-tests.py
+++ b/tests/run-tests.py
@@ -125,6 +125,7 @@ platform_tests_to_skip = {
     "minimal": (
         "basics/class_inplace_op.py",  # all special methods not supported
         "basics/subclass_native_init.py",  # native subclassing corner cases not support
+        "extmod/uctypes_array_load_store.py",  # requires unittest
         "micropython/opt_level.py",  # don't assume line numbers are stored
     ),
     "nrf": (
```
I think that'll work...

projectgus: Ah yeah, I should have picked this up. I think the first version of this only caught `SyntaxError` (as this is what minimal unix port will do if unittest is on the path), and then I expanded it to `ImportError` to work in CI but wasn't thinking of the implications.

Unfortunately the test runner reads the platform of the "minimal" unix variant as "linux".

I've changed it to test via `if getattr(sys.implementation, "_build", None) == "minimal":` which will at least not be brittle, but it doesn't feel like the ideal solution either.

projectgus:
> The message after import unittest never prints, so they all fail on that import.

In this case, how about:

```py
try:
    import unittest
except MemoryError:
    print("SKIP")
    raise SystemExit
```

... at least, for now. It's more boilerplate than I prefer, and there's a potential gap for a test which can import `unittest` but then can't run the test (although maybe the peak memory usage will still happen during the import on a lot of tests?) However, sounds like it will at least catch these particular failures.

tests/extmod/uctypes_array_load_store.py
dpgeorge: I investigated this a bit.  The minimal build can't import `unittest` due to missing `io` module, which is needed by `unittest` only for rendering exceptions nicely.  So we could actually get it working on the minimal port by tweaking `unittest` so that it can run without the `io` module (and just use `repr(exc)` to render the exception in that case).

(There's also #11244 which will allow rendering exceptions without `io`, but it still has issues that need resolving.)

BUT, I think this discussion is all moot because the minimal variant doesn't even have `uctypes` enabled?!  So, simply moving `import unittest` below the `import uctypes` (and also the `int("FF" * 8, 16)` check) should get this working under CI.

dpgeorge:
> In this case, how about:

I tried that (although used `"SKIP-TOO-LARGE"` instead for more detailed reporting) and it does skip correctly on all 3 tests with the native emitter (on ADAFRUIT_ITSYBITSY_M0_EXPRESS).

BUT, running just `--via-mpy` (bytecode emitter, not native emitter) on the `basics/array_limits_intbig.py` it fails with MemoryError part way through (as you predicted!):
```
test_bigint_overflow (__main__.Test) ... ok
test_unsigned_negative (__main__.Test) ... ok
test_zero (__main__.Test) ... ok
test_one (__main__.Test) ... ok
test_array (__main__.Test) ... ok
test_setter_overflow_Q (__main__.Test) ...Traceback (most recent call last):
  File "<stdin>", line 37, in <module>
  File "tests/basics/array_limits_intbig.py", line 118, in <module>
  File "unittest/__init__.py", line 464, in main
  File "unittest/__init__.py", line 269, in run
  File "unittest/__init__.py", line 254, in run
  File "unittest/__init__.py", line 432, in _run_suite
  File "unittest/__init__.py", line 407, in run_one
  File "unittest/__init__.py", line 350, in _handle_test_exception
  File "unittest/__init__.py", line 339, in _capture_exc
MemoryError: memory allocation failed, allocating 136 bytes
CRASH
```
Maybe we can live with that for now (so, include that try-import-unittest-except logic), and I can make a few optimisations to `unittest` to bring its size down.

tests/extmod/uctypes_array_load_store.py
projectgus: > BUT, I think this discussion is all moot because the minimal variant doesn't even have uctypes enabled?! So, simply moving import unittest below the import uctypes (and also the int("FF" * 8, 16) check) should get this working under CI.

:facepalm: I should have checked this. Easy fix!

projectgus:
@dpgeorge Thanks for working through these. I've updated the tests according to the discussion.

tests/basics/array_limits_intbig.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def _set_out_of_range(self, arr, index, value):
+        old_val = arr[index]
+        with self.assertRaises(OverflowError):
+            arr[index] = value
+        self.assertEqual(arr[index], old_val)  # should be unchanged after exception
+
+    @unittest.skipIf(not is_v2, "no overflow checks on v1")
+    def test_setter_overflow_Q(self):
+        arr = array("Q", [1,2,3,4])
```
dpgeorge: Unfortunately these tests are not caught by ruff format, but might be nice to format them regardless.

dpgeorge:
I made a PR to optimise `unittest` https://github.com/micropython/micropython-lib/pull/1125 .  Using that version of `unittest` the tests above can all run or skip on ADAFRUIT_ITSYBITSY_M0_EXPRESS.

tests/basics/array_limits_intbig.py
projectgus: Oops, I've gotten too dependent on pre-commit doing this for me. Formatted.

projectgus:
> I made a PR to optimise `unittest` [micropython/micropython-lib#1125](https://github.com/micropython/micropython-lib/pull/1125) . Using that version of `unittest` the tests above can all run or skip on ADAFRUIT_ITSYBITSY_M0_EXPRESS.

That's great! Do you want to keep the try/except guards in these tests, or pull them out?

dpgeorge:
> Do you want to keep the try/except guards in these tests, or pull them out?

Ah, maybe I wasn't clear enough: the tests can now either run and pass, or SKIP-TOO-LARGE.  There's no longer any "it can import but not run" cases like in https://github.com/micropython/micropython/pull/16311#issuecomment-4794934339

So, we need to keep the try/except guards around `import unittest`.  They still skip when run with the native emitter.  The `unittest` optimisations are needed to get the tests to run and pass with and standard run, and with `--via-mpy`.

tests/basics/array_limits_intbig.py
dpgeorge: Thanks!  Maybe one day we can work on formatting the `tests/basics/` directory.... it's a big job though.

dpgeorge (approved):
Thanks for persisting with this PR, I know it's taken a while to get to this stage with a lot of back and forth.

It looks really good now.  Nice to finally have `int.to_bytes(signed)` support.

Let's merge it!

projectgus:
Thanks @dpgeorge! Glad I could get there after almost two years :sweat_smile: 

@tannewt @dhalbert Heads-up that next time you update the CircuitPython core you'll be able to drop some of these int-related patches. There are lines for `#define OVERFLOW_CHECKS` that you should be able to easily patch to re-enable most of the same overflow checks you have at the moment.

Issue #16181, opened 2024-11-08
lightsleep on Pico W regression for 1.24.0


jeanlucmongrain:
### Port, board and/or hardware

PICO W

### MicroPython version

Micropython 1.24.0

### Reproduction

Micropython 1.23.0 can run that code:

```
import machine
import time

led = machine.Pin("LED", machine.Pin.OUT)
led.off()

while True:
    print("Loop instance")
    led.on()
    time.sleep(0.1)   
    led.off()
    machine.lightsleep(10000)
```

it will sleep for 10s.

if you try the same with 1.24.0 there is absolutely no sleep at all. the string `Loop instance` just appears every 0.1 seconds

### Expected behaviour

`machine.lightsleep` to do something on 1.24.0

### Observed behaviour

`machine.lightsleep` do something on 1.23.0 but not 1.24.0

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

jeanlucmongrain:
related to #16180 16180

CrabbyPete:
I got the same result using a regular PICO. Works on 1.23, not on 1.24.1

scruss:
* expected behaviour seen on a Raspberry Pi Pico (RP2040, **RP2-B1**, _no_ wifi) running v1.24.1
* observed behaviour seen on a Raspberry Pi Pico W (RP2040,  **RP2-B2**, _with_ wifi) running v1.24.1

I don't have a B2 revision Pico to try this on: all of mine are old.

scruss:
> I got the same result using a regular PICO. Works on 1.23, not on 1.24.1

Does your Pico say RP2-B1 on the chip, or RP2-B2?

projectgus:
I posted some explanation of this in one of the linked PRs that's since closed, but here's what is happening (mostly copied from #16431):

`lightsleep()` behaviour has always been to "sleep for the allotted time unless woken by an interrupt first". This issue was introduced in https://github.com/micropython/micropython/commit/74fb42aa82d537b6fefb463907274bd660c6ce8d (https://github.com/micropython/micropython/pull/13329) when we switched away from the pico-sdk alarm pool for soft timers and unintentionally made soft timer expiry a wakeup source.

[...]

Issue #16180, opened 2024-11-07
lightsleep() not working properly Pi Pico WH v1.24.0


RYoder97:
### Port, board and/or hardware

Raspberry Pi Pico WH

### MicroPython version

MicroPython v1.24.0 on 2024-10-25; Raspberry Pi Pico W with RP2040

### Reproduction

Attempt following code, both with lightsleep() having no ms argument and one with:

```python
import machine
from machine import Pin, deepsleep, lightsleep

def wake_up(pin):
    # Handle wake-up logic here
    led.toggle()
    print("Button pressed, waking up!")

# Configure buttons
button = Pin(11, Pin.IN, Pin.PULL_UP) # Example pin
led    = Pin(20, Pin.OUT)

led.on()

# Set up interrupt
button.irq(trigger=Pin.IRQ_FALLING | Pin.IRQ_RISING, handler=wake_up)

# Go to sleep until a button is pressed
lightsleep()

```

### Expected behaviour

Expect LED to react to button press when asleep and execute wake_up function. LED is on by default. LED should toggle/turn off when button is pressed down, both when machine is asleep and not asleep.

### Observed behaviour

USB power tester indicates that the board sleeps only when **NO MILLISECONDS ARGUMENT IS SPECIFIED**. 

When milliseconds **ARE SPECIFIED**, board does not appear to sleep as power consumption does not drop according to the USB tester. The board continues to work at full power, and the LED reacts to the button.

When **NO MILLISECONDS ARE SPECIFIED**, board appears to go to sleep (reducing power from >10ma to 0-1ma. But button does not do anything to LED.

### Additional Information

I have tried adding the ```wake``` argument to `button.irq`, but Micropython says that extra keyword arguments are present. 

I understand sleeping may have issues on the Pico W models. I have also tried turning off WiFi/setting the pin associated with Wifi to "low".  That does not seem to have any impact whatsoever, since my code never explicitly turns Wifi **on**

[https://forums.raspberrypi.com/viewtopic.php?p=2266590#p2266590](url)

### Code of Conduct

Yes, I agree

jeanlucmongrain:
@RYoder97 

```
import machine
import time

led = machine.Pin("LED", machine.Pin.OUT)
led.off()

while True:
    print("Loop instance")
    led.on()
    time.sleep(0.1)   
    led.off()
    machine.lightsleep(10000)
```

[...]

PR #8381, opened 2022-03-04
Add VfsMap filesystem, mpremote deploy-mapfs, and ability to import .mpy files from ROM (WIP)

dpgeorge:
This is part 2 of #8191 (new mpy v6 file format).  The PR adds the ability to place .mpy files in a special filesystem ROM and import them inplace so they don't take up much RAM.  This is essentially dynamically frozen .mpy files.

Things included here are:
- `os.mount(path)` can be used to query a mount point, to return the VFS object mounted there
- new `os.VfsMap` filesystem which is a very simple linear read-only filesystem that can be memory mapped
- when importing .mpy files from a mounted `VfsMap` filesystem, the data in the files there can be referenced directly (ie no need to load it into RAM)
- new `mpremote deploy-mapfs <directory>` command that takes a directory on the host PC, packs it into a mapfs filesystem and deploys it to the MicroPython device
- support on stm32, rp2 and esp32 for a small (~128k) mapfs filesystem which is mounted at `/mapfs` on boot (and also added to `sys.path`)

Workflow for the user:
- install new firmware from this PR
- create a `<directory>` on your PC containing all the code to go in the mapfs (must precompile .py to .mpy beforehand if you want to take advantage of RAM savings, but any types of files can go in this filesystem)
- `mpremote deploy-mapfs <directory>`
- `/mapfs` now contains all files from `<directory>` and they can be listed/read/imported as though it were a normal filesystem

PR #16313, opened 2024-11-27
Add Pico 2 W port and update Pico SDK version

peterharperuk:
### Summary

Add Pico 2 W port and changes made in pico-sdk 2.1.0. This fixes issues with lightsleep associated with sev handling and alarm pool problems.

### Testing

Ran all the micropython tests


peterharperuk:
I've split these changes out from this PR https://github.com/micropython/micropython/pull/16057

ports/rp2/boards/RPI_PICO2_W/board.json
```diff
@@ -0,0 +1,21 @@
...
+        "USB",
+        "WiFi"
+    ],
+    "images": [
+        "rp2-pico2-w.jpg"
+    ],
+    "mcu": "rp2350",
+    "product": "Pico 2 W",
+    "thumbnail": "",
+    "url": "https://www.raspberrypi.com/products/raspberry-pi-pico-2/",
```
dpgeorge, line 19: Should this be different, to point to a Pico 2 W specific page?

ports/rp2/boards/RPI_PICO2_W/mpconfigboard.h
```diff
@@ -0,0 +1,22 @@
+// Board and hardware specific configuration
+#define MICROPY_HW_BOARD_NAME                   "Raspberry Pi Pico 2 W"
+#define MICROPY_HW_FLASH_STORAGE_BYTES          (PICO_FLASH_SIZE_BYTES - 1024 * 1024)
```
dpgeorge, line 3: Might want to decrease the filesystem size by a bit more to make room for a future ROMFS (see #8381).  It has 4MiB flash (right?) so maybe reserve 1.5MiB for the firmware and ROMFS?

ports/rp2/mphalport.c
```diff
@@ -266,15 +266,13 @@ void soft_timer_init(void) {
...
 void mp_wfe_or_timeout(uint32_t timeout_ms) {
-    soft_timer_entry_t timer;
-
-    // Note the timer doesn't have an associated callback, it just exists to create a
-    // hardware interrupt to wake the CPU
-    soft_timer_static_init(&timer, SOFT_TIMER_MODE_ONE_SHOT, 0, NULL);
-    soft_timer_insert(&timer, timeout_ms);
-
-    __wfe();
+    best_effort_wfe_or_timeout(delayed_by_ms(get_absolute_time(), timeout_ms));
```
dpgeorge, line 269: It might be best to stick with our custom implementation here, if that's possible.  I'm still a bit concerned about issues like https://github.com/raspberrypi/pico-sdk/issues/2118

tests/ports/rp2/rp2_lightsleep.py
```diff
@@ -9,18 +9,13 @@
 # A range of sleep periods (1 to 512ms) are tested. The total nominal sleep time
 # is 10.23 seconds, but on most ports this will finish much earlier as interrupts
 # happen before each timeout expires.
-import sys
```
dpgeorge, line 12: Since `sys` is still used below, I suggest keeping this line here.

tests/ports/rp2/rp2_lightsleep.py
```diff
@@ -9,18 +9,13 @@
...
     from machine import lightsleep, Pin
 except ImportError:
     print("SKIP")
     raise SystemExit
 
-# RP2350 currently fails this test, needs further investigation.
-if "RP2350" in sys.implementation._machine:
-    print("SKIP")
-    raise SystemExit
+from sys import stdout, platform
```
dpgeorge, line 18: This line is not needed, just continue to use `sys.stdout.write` below.

dpgeorge (commented):
Thanks for this, it's a relatively concise PR!

ports/rp2/mphalport.c
peterharperuk: I don't think your custom implementation will work and allow the device to go to sleep on rp2350 for reasons we already discussed? https://github.com/raspberrypi/pico-sdk/issues/1812

ports/rp2/boards/RPI_PICO2_W/board.json
peterharperuk: It's the same URL, see the section "Now with added wireless"

ports/rp2/boards/RPI_PICO2_W/mpconfigboard.h
dpgeorge: > changing this also requires communicating to users that their files will go _poof_ when they update the firmware. That was my other rationale for going straight up to 2MB/2MB.

Yes, exactly, the choice made here will be "set in stone" for some time to come.

(Eventually we should support user defined partitions, but the firmware still needs some fixed allocated space.)

ports/rp2/mphalport.c
peterharperuk: @dpgeorge I think that issue can impact machine.Timer if you run the device at a clock speed above 160MHz via machine.freq and the alarm or repeated timer callback takes <1us. I will leave the commit in place for the moment. If you want me to put back the soft timer, I'll remove the changes I've made in this area. (Including the "fixes" for light sleep?). 

One possible workaround (while waiting for a fix to the sdk) would be to add a 1us busy wait to the alarm_callback function. Can you let me know what you prefer me to do?

ports/rp2/boards/RPI_PICO2/mpconfigboard.h
```diff
@@ -1,3 +1,3 @@
 // Board and hardware specific configuration
 #define MICROPY_HW_BOARD_NAME                   "Raspberry Pi Pico2"
-#define MICROPY_HW_FLASH_STORAGE_BYTES          (PICO_FLASH_SIZE_BYTES - 1024 * 1024)
```
peterharperuk, line 3: wrong board!

ports/rp2/boards/RPI_PICO2_W/mpconfigboard.h
peterharperuk: I reduced the flash storage by 1536 * 1024 as requested

ports/rp2/mphalport.c
dpgeorge: I'd rather be conservative here and not change anything around the alarm pool until that pico-sdk issue is fixed (or we can be confident the issue does not affect MicroPython).  Can we please take that change out of this PR and just make this PR about adding Pico 2 W support?  Then we can have a separate PR with the lightsleep fix (you can open that now to not lose track of it) and test and apply that separately.

peterharperuk: I've dropped the timer changes.

mattytrentini:
For anyone that wants to test but doesn't want to build the firmware, here is a [zipped UF2 file](https://github.com/user-attachments/files/18054596/firmware.zip) for the ARM build of 2d4b02b4d.

lgb-this:
I have a project with different topics (webserver, I2C, buttons, asynio, WLAN, MQTT, neo led, logging, ...) and this firmware runs stable for me. I could run the project without any modifications from the Pico W.

scruss:
Requesting a look at https://github.com/micropython/micropython/issues/16180 (and related https://github.com/micropython/micropython/issues/16181),  a confirmed lightsleep regression for Pico W and Pico 2 W

peterharperuk:
Yes. Light sleep is broken I'm afraid.

ports/rp2/boards/RPI_PICO2_W/mpconfigvariant_RISCV.cmake
```diff
@@ -0,0 +1 @@
+set(PICO_PLATFORM "rp2350-riscv")
```
dpgeorge, line 1: The RPI_PICO2 board also has a `mpconfigvariant.cmake` file which is used for the default variant (ARM) and contains:
```
set(PICO_PLATFORM "rp2350")
```
It looks like this is not actually needed because the pico-sdk automatically infers the PICO_PLATFORM based on PICO_BOARD (if PICO_PLATFORM is not given).  But maybe it's worth being explicit here and including `mpconfigvariant.cmake` with this line to set the platform to `rp2350`?

ports/rp2/modmachine.c
```diff
@@ -228,7 +228,7 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
             #if PICO_RP2040
             clocks_hw->sleep_en1 |= CLOCKS_SLEEP_EN1_CLK_USB_USBCTRL_BITS;
             #elif PICO_RP2350
-            clocks_hw->sleep_en1 |= CLOCKS_SLEEP_EN1_CLK_SYS_USBCTRL_BITS;
+            clocks_hw->sleep_en1 |= CLOCKS_SLEEP_EN1_CLK_USB_BITS;
```
dpgeorge, line 231: With this fix, it looks like we can enable the `tests/ports/rp2/rp2_lightsleep.py` test on rp2350, by removing the following lines in that file:
```py
# RP2350 currently fails this test, needs further investigation.
if "RP2350" in sys.implementation._machine:
    print("SKIP")
    raise SystemExit
```

ports/rp2/mpnetworkport.c
```diff
@@ -74,7 +76,9 @@ static void gpio_irq_handler(void) {
 void cyw43_irq_init(void) {
     gpio_add_raw_irq_handler_with_order_priority(CYW43_PIN_WL_HOST_WAKE, gpio_irq_handler, CYW43_SHARED_IRQ_HANDLER_PRIORITY);
     irq_set_enabled(IO_IRQ_BANK0, true);
+    #if !defined(__riscv)
```
projectgus, line 79: Mostly out of curiosity, does the PICO 2W with Wi-Fi work in RISC-V mode?

projectgus (approved):
These changes look good to me, thanks @peterharperuk . I don't have a 2 W to run this code myself, but I've ordered one.

Do I understand correctly that this pico-sdk version still has a buggy alarm pool, but after we update to 2.2.0 then we'd be able to switch the soft timer implementation back to the alarm pool if we wish? (Thanks to your PR https://github.com/raspberrypi/pico-sdk/pull/2127)

ports/rp2/mpnetworkport.c
peterharperuk: yes

peterharperuk: > we'd be able to switch the soft timer implementation back to the alarm pool

The "bug" is that an alarm handler taking <1us can cause future repeating alarms to not occur. This is fixed in develop by https://github.com/raspberrypi/pico-sdk/pull/2127. The proposal is to only replace the soft timer in mp_wfe_or_timeout with best_effort_wfe_or_timeout which handles repeated events properly and would allow the rp2350 to sleep. best_effort_wfe_or_timeout is not actually impacted by the issue addressed by https://github.com/raspberrypi/pico-sdk/pull/2127

ports/rp2/modmachine.c
peterharperuk: I thought we'd removed the lightsleep test due to it not working on pico2? Unfortunately I don't think this has changed. I'll check however.

dpgeorge: I tested this test using this PR and it works on all Pico boards (Pico, Pico2, Pico2 W, and in RISCV mode).

It tests USB staying alive during lightsleep, not the duration of lightsleep itself.

peterharperuk:
Should we include unittest in the list of installed modules? To avoid having to remember to manually install it to run the tests

dpgeorge:
> Should we include unittest in the list of installed modules? To avoid having to remember to manually install it to run the tests

I'd rather not do that.  The unittest package is pretty large when frozen in to firmware and not all ports have the space for it.  Also, it's not a very common package for users to use, so takes up flash space unnecessarily.

If you're using a particular board for running the MicroPython test suite, it's relatively easy to do `mpremote mip install unittest` once on that board and then leave it installed.  Then it will work with the tests from then on.

dpgeorge (approved):

dpgeorge:
I squashed down the last two changes you made @peterharperuk .  This looks good now, it's very clean thank you!

PR #16315, opened 2024-11-27
nrf/ticker: Fix the soft reboot lock-up of the micro:bit board.

robert-hh:
### Summary

The micro:bit board and probably other boards using the music or display module locked up on soft reboot. Reason was an buffer overflow caused by an index counter, which was not reset on soft_reboot.

### Testing

Tested with a micro:bit board, performing a series of soft reboots.


dpgeorge (approved):
Good find!

PR #16324, opened 2024-11-29
ports/zephyr: Create ability to use device_next, and so to use device_next CDC ACM as REPL

VynDragon:
This enables using the newer USB stack CDC ACM for the REPL

### Summary

This enables using the newer USB device_next stack for the mpy REPL via CDC ACM.
To switch to it board file must contain CONFIG_USB_DEVICE_STACK_NEXT=y and CONFIG_USBD_CDC_ACM_CLASS=y
In the case of a board that is a platform that supports the older device stack, CONFIG_USB_DEVICE_STACK=n may be necessary.

### Testing

Tested on promicro nrf52840 (https://github.com/zephyrproject-rtos/zephyr/pull/80168, device and device_next stacks), and the platforms I work on (https://github.com/zephyrproject-rtos/zephyr/pull/78795, device_next only)

### DIFF
--- a/ports/zephyr/CMakeLists.txt
+++ b/ports/zephyr/CMakeLists.txt
@@ -128,6 +128,7 @@ add_dependencies(${MICROPY_TARGET} zephyr_generated_headers)
 target_sources(app PRIVATE
     src/zephyr_start.c
     src/zephyr_getchar.c
+    src/usbd.c
 )
 
 target_link_libraries(app PRIVATE ${MICROPY_TARGET})
--- a/ports/zephyr/Kconfig
+++ b/ports/zephyr/Kconfig
@@ -49,6 +49,14 @@ config MICROPY_FROZEN_MANIFEST
 	depends on MICROPY_FROZEN_MODULES
 	default "boards/manifest.py"
 
+config MICROPY_USB_DEVICE_VID
+	hex "USB VID"
+	default 0x2fe3
+
+config MICROPY_USB_DEVICE_PID
+	hex "USB PID"
+	default 0x0001
+
 endmenu # MicroPython Options
 
 source "Kconfig.zephyr"
--- a/ports/zephyr/main.c
+++ b/ports/zephyr/main.c
@@ -63,6 +63,10 @@
 
 static char heap[MICROPY_HEAP_SIZE];
 
+#if defined(CONFIG_USB_DEVICE_STACK_NEXT)
+extern int mp_usbd_init(void);
+#endif // defined(CONFIG_USB_DEVICE_STACK_NEXT)
+
 void init_zephyr(void) {
     // We now rely on CONFIG_NET_APP_SETTINGS to set up bootstrap
     // network addresses.
@@ -141,6 +145,10 @@ soft_reset:
 
     #ifdef CONFIG_USB_DEVICE_STACK
     usb_enable(NULL);
+    #endif
+
+    #ifdef CONFIG_USB_DEVICE_STACK_NEXT
+    mp_usbd_init();
     #endif
 
     #if MICROPY_VFS
--- a/ports/zephyr/src/usbd.c
+++ b/ports/zephyr/src/usbd.c
@@ -1,0 +1,183 @@
+/*
+* This file is part of the MicroPython project, http://micropython.org/
+*
+* The MIT License (MIT)
+*
+* Copyright (c) 2024-2025 MASSDRIVER EI (massdriver.space)
+*
+* Permission is hereby granted, free of charge, to any person obtaining a copy
+* of this software and associated documentation files (the "Software"), to deal
+* in the Software without restriction, including without limitation the rights
+* to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+* copies of the Software, and to permit persons to whom the Software is
+* furnished to do so, subject to the following conditions:
+*
+* The above copyright notice and this permission notice shall be included in
+* all copies or substantial portions of the Software.
+*
+* THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+* IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+* FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+* AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+* LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+* THE SOFTWARE.
+*/
+#include <stdint.h>
+
+#include <zephyr/device.h>
+#include <zephyr/usb/usbd.h>
+#include <zephyr/usb/bos.h>
+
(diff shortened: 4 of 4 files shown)


projectgus (commented):
Thanks for contributing this as well, @VynDragon!

I have a couple of questions, mostly around making sure this feature is convenient for other boards to enable.

ports/zephyr/CMakeLists.txt
```diff
@@ -123,4 +123,10 @@ target_sources(app PRIVATE
     src/zephyr_getchar.c
 )
 
+if (CONFIG_DT_HAS_ZEPHYR_CDC_ACM_UART_ENABLED)
+target_sources(app PRIVATE
+    src/usbd_cdc_acm.c
```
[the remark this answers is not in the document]
VynDragon: Done

ports/zephyr/mpconfigport.h
```diff
@@ -160,3 +160,6 @@ typedef long mp_off_t;
         k_msleep(1); \
     } while (0);
 #endif
+
+#define MICROPY_HW_USB_VID  (0x1209)
+#define MICROPY_HW_USB_PID  (0xADDA)
```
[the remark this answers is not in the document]
VynDragon: Done

VynDragon:
Okay so this whole PR was bothering me, i revamped it to be basically usbd next init instead of just CDC ACM. In practice it still does the same thing except now it also initializes USB if other USB things are provided, not just CDC ACM.

ports/zephyr/main.c
```diff
@@ -62,6 +66,30 @@
 
 static char heap[MICROPY_HEAP_SIZE];
 
+#if defined(CONFIG_USB_DEVICE_STACK_NEXT)
+
+extern struct usbd_context *mpy_usbd_init_device(usbd_msg_cb_t msg_cb);
+
+static struct usbd_context *mpy_usbd;
+
+static int mp_usbd_init(void) {
```
dpgeorge, line 75: I'm curious why you put this helper function here in main.c, and not in src/usbd.c?  Maybe it can be moved to src/usbd.c?

ports/zephyr/src/usbd.c
```diff
@@ -0,0 +1,167 @@
...
+* OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+* THE SOFTWARE.
+*/
+#include <stdint.h>
+
+#include <zephyr/device.h>
+#include <zephyr/usb/usbd.h>
+#include <zephyr/usb/bos.h>
+
+#include "../mpconfigport.h"
```
dpgeorge, line 32: This file should not be included directly.  Instead include `py/mpconfig.h`.

ports/zephyr/src/usbd.c
```diff
@@ -0,0 +1,167 @@
...
+#include <zephyr/logging/log.h>
+LOG_MODULE_REGISTER(mpy_usbd);
+
+/* By default, do not register the USB DFU class DFU mode instance. */
+static const char *const blocklist[] = {
+    "dfu_dfu",
+    NULL,
+};
+
+USBD_DEVICE_DEFINE(mpy_usbd,
```
dpgeorge, line 45: The prefix `mpy_` is not used elsewhere in this code base, so I suggest this be changed to `mp_usbd`.  Similarly for all the other symbols in this file.

ports/zephyr/src/usbd.c
```diff
@@ -0,0 +1,167 @@
...
+*/
+#include <stdint.h>
+
+#include <zephyr/device.h>
+#include <zephyr/usb/usbd.h>
+#include <zephyr/usb/bos.h>
+
+#include "../mpconfigport.h"
+
+#if defined(CONFIG_USB_DEVICE_STACK_NEXT)
```
dpgeorge, line 34: This is `#if CONFIG_USB_DEVICE_STACK_NEXT` in main.c, which is subtly different.

Would be good to be consistent.  I think `#ifdef CONFIG_USB_DEVICE_STACK_NEXT` is more consistent with other config options used in this port.

dpgeorge (commented):
Thanks for this, it looks like a good improvement.

Ideally I'd like to enable this new USB stack by default.  Would that be too much of a breaking change?  (Boards could opt-out if needed.)

ports/zephyr/main.c
VynDragon: Yea it could, will do.

ports/zephyr/src/usbd.c
VynDragon: My bad, missed that when I switched it from just USB CDC to generic USB device next init.

VynDragon:
>Ideally I'd like to enable this new USB stack by default. Would that be too much of a breaking change? (Boards could opt-out if needed.)

It will become the default USB stack at some point soon on zephyr, so no need.

ports/zephyr/src/usbd.c
VynDragon: That doesnt work because it's a private

ports/zephyr/src/usbd.c
VynDragon: done

ports/zephyr/main.c
VynDragon: done

ports/zephyr/src/usbd.c
dpgeorge: I'm not sure what you mean by "private src"?

Note that `mpconfigport.h` may not be the correct file, sometimes it could be `mpconfigport_minimal.h`, or another custom file.  We need to find a way to make the include of `py/mpconfig.h` work.

VynDragon: Yes, I understand the issue, but I am not sure how to fix it,
It is in this section:
```
target_sources(app PRIVATE
    src/zephyr_start.c
    src/zephyr_getchar.c
    src/usbd.c
)
```
so it doesnt have access to the `py/mpconfig.h` file in its include path (not exactly sure why, but it doesnt). How do you suggest solving this? moving it to zephyr_usbd.c one folder up?

dpgeorge: It feels like this is a layering issue:
1. zephyr itself
2. these files in `ports/zephyr/src` which are a slight extension to zephyr, to add a little more functionality for use by MicroPython
3. the MicroPython port to zephyr, sitting on top of the above

So, layer 2 should not know anything about layer 3.  If there's anything in layer 2 that needs layer 3, it should either be abstracted out somehow (eg a config option passed from 3 into 2) or moved to layer 3.

Why exactly is this config file needed?  Maybe it was for the VID/PID which were originally in that file, but now moved to Kconfig?

ports/zephyr/Kconfig
```diff
@@ -49,6 +49,14 @@ config MICROPY_FROZEN_MANIFEST
 	depends on MICROPY_FROZEN_MODULES
 	default "boards/manifest.py"
 
+config MICROPY_USB_DEVICE_VID
+	hex "USB VID"
+	default 0x2fe3
+
+config MICROPY_USB_DEVICE_PID
+	hex "USB VID"
```
projectgus, line 57:
```suggestion
	hex "USB PID"
```

ports/zephyr/src/usbd.c
VynDragon: You are right, it is not even necessary anymore...

VynDragon: Done

ports/zephyr/Kconfig
VynDragon: Done

dpgeorge (approved):

dpgeorge:

PR #16328, opened 2024-11-29
ports/nrf/boards/ARDUINO_NANO_33_BLE_SENSE: Update config.

iabdalkader:
### Summary

Update board config to enable hardware SPI and timer, and define LED pins.


ports/nrf/boards/ARDUINO_NANO_33_BLE_SENSE/mpconfigboard.h
```diff
@@ -13,6 +13,8 @@
 
 #define MICROPY_PY_MACHINE_UART         (1)
 #define MICROPY_PY_MACHINE_HW_PWM       (1)
+#define MICROPY_PY_MACHINE_HW_SPI       (1)
```
dpgeorge, line 16: This define is not used anywhere in this repo, what did you intend it to do?

ports/nrf/boards/ARDUINO_NANO_33_BLE_SENSE/mpconfigboard.h
```diff
@@ -13,6 +13,8 @@
...
+#define MICROPY_PY_MACHINE_TIMER        (1)
```
dpgeorge, line 17: I don't think this will enable `machine.Timer`, you need to still add the correct entry to the `machine` globals table.  Maybe you meant to use `MICROPY_PY_MACHINE_TIMER_NRF`?

iabdalkader: Yes I did, thanks! I think it was renamed at some point.

dpgeorge (approved):

PR #16331, opened 2024-11-29
ports/mimxrt: Switch to shared TinyUSB descriptor.

iabdalkader:
### Summary

Switch mimxrt port to use shared TinyUSB descriptor. This removes duplicated tusb config/port.

### Testing

Tested on RT1062, CDC + MSC still working.
robert-hh: CDC works with -1011, -1015, -1020 and -1176
dpgeorge: Tested Teensy 4.0 with the USB CDC stress test and it works.


robert-hh:
CDC works with -1011, -1015, -1020 and -1176 as well. Did not test -1052, which is sufficiently close to -1060. How did you test MSC?

iabdalkader:
> How did you test MSC?

I have it enabled in my fork by default. The old tusb options/port are exactly the same as the ones in shared, so it works the same. I also have support for switching the medium between internal flash and SD-Card, both work fine.

dpgeorge (approved):
Thanks for this, it's a good clean up to make mimxrt more consistent with other TinyUSB ports.

I note that the serial number now uses lowercase hex, where it previously used uppercase hex.  But that should be an OK change, and makes it consistent with other ports.

dpgeorge:
I tested Teensy 4.0 with the USB CDC stress test and it works.

PR #16337, opened 2024-11-30
extmod/extmod.mk: Fix libmetal build prefix.

iabdalkader:
libmetal source files already have the build directory prefix, because they're auto-generated inside the build directory. When they're added to `SRC_THIRDPARTY_C`, another build directory prefix is added resulting
in the object files being generated in a nested build directory. This patch strips the build directory prefix before adding libmetal's source files.

This has no effect on functionality, it's just to avoid nested build directories which can be confusing.


dpgeorge (approved):
Thanks, I tested this and it does improve the build directory layout.

PR #16338, opened 2024-11-30
ports/stm32/sdram.c: Fix SDRAM refresh count setting.

iabdalkader:
### Summary

SDRAM refresh count calculations used a hard-coded SDRAM frequency and refresh cycles. The two boards that use it actually use a 100MHz SDRAM frequency, and 4096 refresh cycles.

### Testing

Tested by running exhaustive SDRAM test on Portenta-H7


dpgeorge:
This looks OK, but `STM32F7DISC` and `STM32F769DISC` (with optional SDRAM enabled) no longer build.

I suggest providing default values for `MICROPY_HW_SDRAM_REFRESH_CYCLES` and `MICROPY_HW_SDRAM_FREQUENCY` (or if you think they should always be provided by a board, so the board creator is forced to think about their values, then add them to those two boards).

Also, I suggest renaming `MICROPY_HW_SDRAM_FREQUENCY` to `MICROPY_HW_SDRAM_FREQUENCY_KHZ` to make it explicit what the units are.

iabdalkader:
> if you think they should always be provided by a board, so the board creator is forced to think about their values, then add them to those two boards

I think so. The SDRAM frequency and refresh cycles depend on the other SDRAM settings, exact SDRAM chip and speed grade. I think the original 90MHz and 8192 came from those two boards, so I used them in the board config files.

> Also, I suggest renaming MICROPY_HW_SDRAM_FREQUENCY to MICROPY_HW_SDRAM_FREQUENCY_KHZ to make it explicit what the units are.

Renamed.

dpgeorge (approved):

dpgeorge:
@iabdalkader this broke `STM32F429DISC`, we forgot to enable the SDRAM settings on that board.  Can you please make a PR to fix it?

iabdalkader:
Sure.

PR #16342, opened 2024-12-01
ports: Switch FATFS LFN to type 2.

iabdalkader:
### Summary

FatFS LFN type 2 uses the stack to allocate the internal working buffer for LFN, which is thread-safe and saves about 300bytes of BSS memory. The mimxrt port is also changed to use LFN2 and the configuration is aligned with the rest of the ports (enable multi-partition, use label etc..).

Note I only changed the major ports that actually use FatFs (for storage for example).

```
      stm32:  -436 -0.111% PYBV10[incl -512(bss)]
     mimxrt:  +312 +0.085% TEENSY40
      rp2:  -384 -0.042% RPI_PICO_W[incl -512(bss)]
```

### Testing

This config has been tested on rp2, mimxrt, stm32 ports.

### DIFF
--- a/ports/mimxrt/mpconfigport.h
+++ b/ports/mimxrt/mpconfigport.h
@@ -118,10 +118,12 @@ uint32_t trng_random_u32(void);
 #define MICROPY_PY_ONEWIRE                  (1)
 
 // fatfs configuration used in ffconf.h
-#define MICROPY_FATFS_ENABLE_LFN            (1)
-#define MICROPY_FATFS_RPATH                 (2)
-#define MICROPY_FATFS_MAX_SS                (4096)
+#define MICROPY_FATFS_ENABLE_LFN            (2)
 #define MICROPY_FATFS_LFN_CODE_PAGE         437 /* 1=SFN/ANSI 437=LFN/U.S.(OEM) */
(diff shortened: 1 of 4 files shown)


dpgeorge:
What is the motivation for the LFN change, is it just to save BSS?

It looks fine to me, and any port that can afford 512 bytes of BSS should have 512 bytes of stack available (and could now use that additional BSS to increase its stack if needed).

> saves about 300bytes of BSS memory

It should save exactly 512 bytes of BSS (255+1 wide chars worth).  The code size diff is text+bss change.

iabdalkader:
> What is the motivation for the LFN change, is it just to save BSS?

Yes, this is the main motivation. BSS memory gets reserved while stack is re-usable. Even if we can spare the BSS memory, it's best to use it for something else (ex: more heap). As a bonus, stack memory should be thread-safe.

> It should save exactly 512 bytes of BSS

I noticed that too. The comment in FatFS said 300 bytes, I thought it was rounded up somehow, but the more it saves the better. The mimxrt ended up increasing slightly, but that's because its default config was already different than other ports.

dpgeorge (approved):

PR #16346, opened 2024-12-02
ports/stm32: Add deinit_all functions for I2C and SPI.

iabdalkader:
### Summary

SPI and I2C objects can remain active after a soft-reboot because they are statically allocated and lack a finalizer to collect and deinitialize them. This patch adds `deinit_all` functions for SPI and I2C, similar to other peripherals such as UART, DAC, etc.

### DIFF
--- a/ports/stm32/pyb_i2c.c
+++ b/ports/stm32/pyb_i2c.c
@@ -428,6 +428,15 @@ int pyb_i2c_init_freq(const pyb_i2c_obj_t *self, mp_int_t freq) {
     return pyb_i2c_init(self->i2c);
 }
 
+void pyb_i2c_deinit_all(void) {
+    for (int i = 0; i < MP_ARRAY_SIZE(pyb_i2c_obj); i++) {
+        const pyb_i2c_obj_t *pyb_i2c = &pyb_i2c_obj[i];
+        if (pyb_i2c->i2c != NULL) {
+            i2c_deinit(pyb_i2c->i2c);
+        }
+    }
+}
+
 static void i2c_reset_after_error(I2C_HandleTypeDef *i2c) {
     // wait for bus-busy flag to be cleared, with a timeout
     for (int timeout = 50; timeout > 0; --timeout) {
--- a/ports/stm32/spi.c
+++ b/ports/stm32/spi.c
@@ -545,6 +545,15 @@ void spi_deinit(const spi_t *spi_obj) {
     }
 }
 
+void spi_deinit_all(void) {
+    for (int i = 0; i < MP_ARRAY_SIZE(spi_obj); i++) {
+        const spi_t *spi = &spi_obj[i];
+        if (spi->spi != NULL) {
+            spi_deinit(spi);
+        }
+    }
+}
+
 static HAL_StatusTypeDef spi_wait_dma_finished(const spi_t *spi, uint32_t t_start, uint32_t timeout) {
     volatile HAL_SPI_StateTypeDef *state = &spi->spi->State;
     for (;;) {
--- a/ports/stm32/i2c.h
+++ b/ports/stm32/i2c.h
@@ -50,6 +50,7 @@ void i2c_init0(void);
 int pyb_i2c_init(I2C_HandleTypeDef *i2c);
 int pyb_i2c_init_freq(const pyb_i2c_obj_t *self, mp_int_t freq);
 uint32_t pyb_i2c_get_baudrate(I2C_HandleTypeDef *i2c);
+void pyb_i2c_deinit_all(void);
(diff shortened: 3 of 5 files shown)


ports/stm32/pyb_i2c.c
```diff
@@ -408,6 +408,15 @@ void i2c_deinit(I2C_HandleTypeDef *i2c) {
     }
 }
 
+void i2c_deinit_all(void) {
```
dpgeorge, line 411: I would call this `pyb_i2c_deinit_all` to make it clear that it's not related to `i2c.c` functions.

ports/stm32/pyb_i2c.c
```diff
@@ -408,6 +408,15 @@ void i2c_deinit(I2C_HandleTypeDef *i2c) {
...
+    for (int i = 0; i < (sizeof(pyb_i2c_obj) / sizeof(pyb_i2c_obj_t)); i++) {
```
dpgeorge, line 412: Please use `MP_ARRAY_SIZE`.

ports/stm32/spi.c
```diff
@@ -545,6 +545,15 @@ void spi_deinit(const spi_t *spi_obj) {
     }
 }
 
+void spi_deinit_all(void) {
+    for (int i = 0; i < (sizeof(spi_obj) / sizeof(spi_t)); i++) {
```
dpgeorge, line 549: Please use `MP_ARRAY_SIZE`.

ports/stm32/pyb_i2c.c
iabdalkader: Done.

ports/stm32/pyb_i2c.c
iabdalkader: Done.

dpgeorge (approved):

dpgeorge:
Tested on PYBV10, all tests pass (including hardware ones).

dpgeorge:
@iabdalkader unfortunately this change does not work well with boards that use a hardware SPI instance for external SPI flash, because on soft reset the SPI bus is reset and so the SPI flash no longer works.  Boards that are affected are:
- ADAFRUIT_F405_EXPRESS
- LEGO_HUB_NO6
- MIKROE_QUAIL
- WEACT_F411_BLACKPILL

What was your motivation to reset SPI on soft reset?

iabdalkader:
> What was your motivation to reset SPI on soft reset?

This caused us a lot of issues in the past so we patched it a long time ago as it left peripherals configured and enabled, possibly with IRQs.

> boards that use a hardware SPI instance for external SPI flash

Shouldn't this be reserved? i.e., not usable in machine_spi or pyb_spi? Or do those boards use the MicroPython APIs to initialize SPI? If so, could we define a macro to not de-initialize certain peripherals?

dpgeorge:
> Shouldn't this be reserved? i.e., not usable in machine_spi or pyb_spi? Or do those boards use the MicroPython APIs to initialize SPI? If so, could we define a macro to not de-initialize certain peripherals?

There are a few ways to fix it:
1. For the SPI that's used for SPI flash, don't expose it to the user at all (in the board config).
2. Use `MICROPY_HW_SPI_IS_RESERVED()` to mark SPI used by the flash as reserved, and don't reset them on soft-reset.  This would also prevent the user from accessing it, so very similar to option 1 (only difference is that the SPI is still available for use on the C side).
3. Provide a `MICROPY_HW_SPI1_IS_STATIC` config option for all SPI instances, and if a board sets it to true then don't reset the SPI on soft-reset.  This allows the user to still access it, and also be accessed from C.  This kind of matches how UART works (see the `machine_uart_obj_t::is_static` variable).

Option 1 is the simplest, and probably the best from a user point of view, so users don't accidentally break the filesystem by reconfiguring the SPI.

On the other hand, maybe the user needs the SPI for something else, eg some other device is connected to that SPI bus with a separate CS pin.  I don't know if there are any boards like that?  It seems quite fragile to expose the SPI flash bus to the user.

iabdalkader:
> On the other hand, maybe the user needs the SPI for something else, eg some other device is connected to that SPI bus with a separate CS pin. I don't know if there are any boards like that? It seems quite fragile to expose the SPI flash bus to the user.

I agree, but I can't tell whether the bus is connected to something else or not. This requires knowledge of the boards in question. I'd say it's best to go with option 2, at least this way it's easier to undo if the bus is shared. Plus SPI in this port already supports `IS_RESERVED`, so all we need to do is Not allow deinit reserved buses and define flash bus as reserved for these boards. However, if some board shares a bus, as bad as it may be, we'll need another option for shared + don't deinit.

PR #16350, opened 2024-12-03
esp32,rp2: Support common BUILD_VERBOSE and V=1 build flags.

projectgus:
### Summary

Small build system consistency/quality-of-life improvement:

* Adds the same `V=1` or `BUILD_VERBOSE` flags that exist on other ports to the esp32 and rp2 ports.
* Previously, to get verbose build output had to run rp2 build with `VERBOSE=1` (undocumented) and esp32 with the same (for Makefile generator) or by running `idf.py` manually (for Ninja generator).


ports/rp2/Makefile
```diff
@@ -28,7 +30,9 @@ else
 BUILD ?= build-$(BOARD)
 endif
 
-$(VERBOSE)MAKESILENT = -s
```
projectgus, line 31: This PR no longer passes `-s` to the top-level make invocation, but this appears not necessary as the generated Makefile sets `.SILENT` at the top level based on the `VERBOSE` variable. I couldn't see a difference with or without `-s` at the top level.

ports/rp2/Makefile
```diff
@@ -28,7 +30,9 @@ else
...
+ifeq ($(BUILD_VERBOSE),1)
+    MAKE_ARGS += VERBOSE=1  # Picked up in Makefile generated by CMake
```
dpgeorge, line 34: Maybe un-indent this, to match the ifdef's below?

dpgeorge:
This looks good to me.

ports/rp2/Makefile
projectgus: Oops, yes thanks. Extra indenting in Makefiles is a dangerous game :laughing:

dpgeorge (approved):

PR #16352, opened 2024-12-03
py/misc: Fix msvc and C++ compatibility.

stinos:
### Summary

Use an explicit cast to suppress the implicit conversion which started popping up in recent compiler versions (and wasn't there yet in 07bf3179).

### Testing

Standard CI tests all run


dpgeorge (approved):

dpgeorge:
Looks correct, thanks.

Issue #16312, opened 2024-11-27
32bit unix port build fails to find module in same directory


stinos:
### Port, board and/or hardware

unix port on WSL, source code on a mounted windows directory

### MicroPython version

Latest master

### Reproduction

- clone and cd to unix port directory
- add a 32bit directory under variants containing

micropython/ports/unix/variants/32bit/mpconfigvariant.h:
```
#define MICROPY_CONFIG_ROM_LEVEL (MICROPY_CONFIG_ROM_LEVEL_EXTRA_FEATURES)
#define MICROPY_DEBUG_PRINTERS (1)
#define MICROPY_FLOAT_IMPL (MICROPY_FLOAT_IMPL_DOUBLE)
#define MICROPY_LONGINT_IMPL (MICROPY_LONGINT_IMPL_MPZ)
```

micropython/ports/unix/variants/32bit/mpconfigvariant.mk:
```
MICROPY_FORCE_32BIT = 1
MICROPY_PY_FFI = 0
MICROPY_PY_BTREE = 0
```
- build standard and 32bit variants, so that's `make` and `make VARIANT=32BIT`
- create a directory frob containing

micropython/ports/unix/frob/foo.py:
```
# empty
```
micropython/ports/unix/frob/bar.py:
```
import foo
```
- run `./build-standard/micropython ./frob/bar.py` -> no output, exepected
- run `./build-32bit/micropython ./frob/bar.py` -> output:

```
Traceback (most recent call last):
  File "./frob/bar.py", line 1, in <module>
ImportError: no module named 'foo'
```

### Expected behaviour

Behaviour should be the same as standard variant

### Observed behaviour

`import foo` fails to find `foo` even though it's in the same directory as the module where the import statement is

### Additional Information

Could be related to other features in this 32bit variant, though setting `MICROPY_FORCE_32BIT = 0` does make the problem go away. I'm caught up in getting other things working at the moment so don't have time to look into thi. Could just as well be a WSL bug.

### Code of Conduct

Yes, I agree

agatti:
It seems to not be a problem on my Linux machine, unless I followed your steps incorrectly:

```
~/src ➜ cat micropython/ports/unix/variants/32bit/mpconfigvariant.h
#define MICROPY_CONFIG_ROM_LEVEL (MICROPY_CONFIG_ROM_LEVEL_EXTRA_FEATURES)
#define MICROPY_DEBUG_PRINTERS (1)
[...]

PR #16354, opened 2024-12-04
tools/ci.sh: Fix listdir test by setting _FILE_OFFSET_BITS=64.

dpgeorge:
### Summary

As suggested in https://github.com/micropython/micropython/issues/16312#issuecomment-2514587579 it may be possible to fix the unix MIPS listdir test.

### Testing

Hopefully CI passes!

### DIFF
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -48,6 +48,10 @@ CWARN = -Wall -Werror
 CWARN += -Wextra -Wno-unused-parameter -Wpointer-arith -Wdouble-promotion -Wfloat-conversion
 CFLAGS += $(INC) $(CWARN) -std=gnu99 -DUNIX $(COPT) -I$(VARIANT_DIR) $(CFLAGS_EXTRA)
 
+# Force the use of 64-bits for file sizes in C library functions on 32-bit platforms.
+# This option has no effect on 64-bit builds.
+CFLAGS += -D_FILE_OFFSET_BITS=64
+
 # Debugging/Optimization
 ifdef DEBUG
 COPT ?= -Og
--- a/ports/windows/Makefile
+++ b/ports/windows/Makefile
@@ -42,6 +42,10 @@ INC += -I$(VARIANT_DIR)
 CFLAGS += $(INC) -Wall -Wpointer-arith -Wdouble-promotion -Werror -std=gnu99 -DUNIX -D__USE_MINGW_ANSI_STDIO=1 $(COPT) $(CFLAGS_EXTRA)
 LDFLAGS += -lm -lbcrypt $(LDFLAGS_EXTRA)
 
+# Force the use of 64-bits for file sizes in C library functions on 32-bit platforms.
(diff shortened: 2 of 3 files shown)


dpgeorge:
This is how the CI used to fail on the unix qemu MIPS port:
```diff
--- /home/runner/work/micropython/micropython/tests/results/extmod_vfs_posix_paths.py.exp	2024-12-04 02:01:22.215945136 +0000
+++ /home/runner/work/micropython/micropython/tests/results/extmod_vfs_posix_paths.py.out	2024-12-04 02:01:22.215945136 +0000
@@ -1,12 +1,12 @@
-listdir("/"): ['subdir']
-listdir("."): ['subdir']
+listdir("/"): []
+listdir("."): []
 getcwd() in {"", "/"}: True
 chdir("subdir"): None
 getcwd(): /subdir
 mkdir("two"): None
-listdir("/"): ['subdir']
-listdir("/subdir"): ['file.py', 'one', 'two']
-listdir("."): ['file.py', 'one', 'two']
+listdir("/"): []
+listdir("/subdir"): []
+listdir("."): []
 print('hello')
 hello
 <module 'file' from 'file.py'>
```
But now with `_FILE_OFFSET_BITS` defined to 64 the tests all pass.

Thanks @stinos for the hint!

projectgus:
@dpgeorge Good to have this fixed! Maybe I don't fully understand the root cause, but should we fix this in the port Makefile rather than the CI system - for non-CI builds of this port+config?

dpgeorge:
> Maybe I don't fully understand the root cause

Neither do I!



> should we fix this in the port Makefile rather than the CI system - for non-CI builds of this port+config?

Yes, that's a good idea.  Just for MIPS builds (at this stage, that's where we see the failure).

agatti:
Shouldn't this be applied to the Unix/Arm qemu build too?

I've seen this issue in #15800 for both MIPS and Arm; as in, the toolchains I was using in the CI-like environment had that define baked in somehow and made tests pass in my VM.

stinos:
This should probably be applied for all 32bit builds, preferably in the Makefile.

dpgeorge:
> This should probably be applied for all 32bit builds, preferably in the Makefile.

I've now added it unconditionally to the unix Makefile, so it is applied for all builds.  The CI actually passes with that (on 32-bit and 64-bit builds).

I'm not exactly sure how to make it apply for 32-bit only builds.  Any ideas?  Or maybe it can stay unconditionally enabled?

stinos:
It has no effects on 64bit builds (https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html).
Should also be added to at least the windows port Makefile, and maybe other 'large' ports as well?

dpgeorge:
I have updated this PR to force `_FILE_OFFSET_BITS=64` on both the unix and windows ports.  On 64-bit systems this should have no effect.  On 32-bit systems it will force 64-bit file sizes which should work on pretty much all systems these days.

projectgus (approved):

Issue #15564, opened 2024-07-28
Implement machine.USBDevice for ESP32-S3


tom709:
### Description

The ESP32-S3 chip has excellent USB support built-in. Please implement machine.USBDevice for ESP32-S3.

### Implementation

I hope the MicroPython maintainers or community will implement this feature

### Code of Conduct

Yes, I agree

projectgus:
I'd also like to see this supported!

I realised that we might be able to do it another way, but locking the GIL for each of the USB callbacks. But that's also a lot of work, and I like the approach in your PR better.

TomFahey:
I noticed that your commits were merged last week, and so took the opportunity to compile a couple of micropython images for the Arduino Nano ESP32 and a Waveshare Pico S2.

I've only tested the use of basic USB CDC instantiation (using the 
 higher level micropython-lib [usb-device](https://github.com/micropython/micropython-lib/tree/master/micropython/usb) drivers, which make use of machine.USBDevice, however so far everything has worked flawlessly.

Fantastic job and thank you very much.

andrewleech:
That's really interesting to hear @TomFahey did you just enable the runtime build config?
```
#define MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE    (1) // Support machine.USBDevice
```

TomFahey:
Pretty much - I also had to add `${MICROPY_DIR}/shared/tinyusb/mp_usbd_runtime.c` to

```c++
    list(APPEND MICROPY_SOURCE_TINYUSB
          ${TINYUSB_SRC}/tusb.c
          ${TINYUSB_SRC}/common/tusb_fifo.c
          ${TINYUSB_SRC}/device/usbd.c
          ${TINYUSB_SRC}/device/usbd_control.c
          ${TINYUSB_SRC}/class/cdc/cdc_device.c
          ${TINYUSB_SRC}/portable/synopsys/dwc2/dcd_dwc2.c
          ${MICROPY_DIR}/shared/tinyusb/mp_usbd.c
          ${MICROPY_DIR}/shared/tinyusb/mp_usbd_cdc.c
          ${MICROPY_DIR}/shared/tinyusb/mp_usbd_descriptor.c
          ${MICROPY_DIR}/shared/tinyusb/mp_usbd_runtime.c
      )
``` 

in the esp32_common.cmake file to get it to compile and also added 

```c++
    #if MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE
[...]

PR #16356, opened 2024-12-04
esp32: Enable support for machine.USBDevice.

sebromero:
### Summary

Recently support for machine.USBDevice was kindly added to the ESP32 port. However it's not currently enabled in the runtime yet by default.
After discussing it (#15564) I followed the recipe of @TomFahey and the suggestions of @andrewleech to enable the USB runtime support support the same way it's been added e.g. to the RP2 port.

### Testing

I tested it on an Arduino Nano ESP32 board by emulating a HID keyboard. In fact I had a whole class of students testing it and it worked beautifully.

I'm not sure how this affects the binary size but I would expect a minimal increase.


dpgeorge:
Thanks for this, it's very simple!  This is a great feature to enable.

I guess it's enabled on all ESP32-S2 and -S3 boards.

I will test it.

dpgeorge:
I tested ESP32_GENERIC_S2 and ESP32_GENERIC_S3 firmware with a HID keyboard example, and it works.

dpgeorge (approved):

PR #10607, opened 2023-01-28
ports/RTC: Attempt to reduce the inconsistencies between the port's RTC implementation.

robert-hh:
This PR follows the issue #10578 and #5553 and other discussions about the topic. It consists of the commits:

- `docs/quickref: Fix the documentation of rtc.datetime()`, which makes the documentation of rtc.datetime() consistent to the implementation for all ports.
- `mimxrt/machine_rtc.c: Drop rtc.now()` Keep it at the WiPy port for legacy,
- `ports/RTC: Rework the availability and interface of rtc.init()`, which  drops rtc.init() from the SAMD port and makes it consistent to the documentation and the WiPy port for the ports ESP32 and MIMXRT. But rtc.init() could as well be dropped for the latter ones and just kept for WiPy.

PR #16357, opened 2024-12-04
esp32: Fix RTC initialization from datetime.

sebromero:
### Summary

I was pulling my hair out trying to initialize the RTC though its `init` function as follows:
```py
rtc = RTC()

year = 2024
month = 12
day = 26
weekday = 3
hour = 6
minute = 30
second = 20
microsecond = 0
tzinfo = 0

rtc.init((year, month, day, hour, minute, second, microsecond, tzinfo))
```
However it kept being initialized with a wrong date / time. The documentation states:

> Initialise the RTC. Datetime is a tuple of the form:
> (year, month, day[, hour[, minute[, second[, microsecond[, tzinfo]]]]]

The problem lies in the fact that both the `datetime` setter and init use the same helper function. The 8-tuple for that one is different though according to the docs:

> The 8-tuple has the following format:
> (year, month, day, weekday, hours, minutes, seconds, subseconds)

It uses `weekday` which makes sense as the getter should deliver this information and both setter and getter shall use the same format. Now we can see that using the same helper function for both formats cannot work and the two need to be distinguished. This is what this PR does. It fixes the issue by treating the two differently.

### Testing

I tested the fix on an Arduino Nano ESP32. I wasn't sure where to add a unit test, but here is what I wrote in MicroPython:

```py
from machine import RTC

rtc = RTC()

year = 2024
month = 12
day = 26
weekday = 3
hour = 6
minute = 30
second = 20
microsecond = 0
tzinfo = 0

def test(source, should_fail=False):
    print(f"\nTesting {source}")
    # Read time
    t = rtc.datetime()
    print(t)
    try:
        assert t[0] == 2024, f"Year should be 2024 but was {t[0]}"
        assert t[1] == 12, f"Month should be 12 but was {t[1]}"
        assert t[2] == 26, f"Day should be 26 but was {t[2]}"
        assert t[3] == 3, f"Weekday should be 3 but was {t[3]}"
        assert t[4] == 6, f"Hour should be 6 but was {t[4]}"
        assert t[5] == 30, f"Minute should be 30 but was {t[5]}"
        assert t[6] == 20, f"Second should be 20 but was {t[6]}"
        if should_fail:
            print("❌ Test should have failed but didn't")
        else:
            print("✅ Test passed")
    except AssertionError:
        if should_fail:
            print("✅ Test failed as expected")
        else:
            print("❌ Test passed but should have failed")


# Initialise the RTC. Datetime is a tuple of the form:
# (year, month, day[, hour[, minute[, second[, microsecond[, tzinfo]]]]]
rtc.init((year, month, day, hour, minute, second, microsecond, tzinfo))
test("init in order")

rtc.init((year, month, day, tzinfo, hour, minute, second, microsecond))
test("init out of order", should_fail=True)

# The 8-tuple has the following format:
# (year, month, day, weekday, hours, minutes, seconds, subseconds)
rtc.datetime((year, month, day, weekday, hour, minute, second, microsecond))
test("datetime in order")
```

The output before applying the fix:

```
Testing init in order
(2024, 12, 27, 4, 6, 20, 0, 373)
❌ Test passed but should have failed

Testing init out of order
(2024, 12, 26, 3, 6, 30, 20, 245)
❌ Test should have failed but didn't

Testing datetime in order
(2024, 12, 26, 3, 6, 30, 20, 256)
✅ Test passed
```

The output after applying the fix:

```
Testing init in order
(2024, 12, 26, 3, 6, 30, 20, 389)
✅ Test passed

Testing init out of order
(2024, 12, 26, 3, 0, 6, 30, 365)
✅ Test failed as expected

Testing datetime in order
(2024, 12, 26, 3, 6, 30, 20, 235)
✅ Test passed
```

An alternative could be to change the documentation so that the `init` and the `datetime` function use the same format. `tzinfo` doesn't seem to be used anyway. At least not in this port.

### DIFF
--- a/ports/esp32/machine_rtc.c
+++ b/ports/esp32/machine_rtc.c
@@ -126,15 +126,20 @@ static mp_obj_t machine_rtc_datetime_helper(mp_uint_t n_args, const mp_obj_t *ar
         return mp_obj_new_tuple(8, tuple);
     } else {
         // Set time
-
         mp_obj_t *items;
-        mp_obj_get_array_fixed_n(args[1], 8, &items);
 
         struct timeval tv = {0};
         if (use_init_format) {
-            tv.tv_sec = timeutils_seconds_since_epoch(mp_obj_get_int(items[0]), mp_obj_get_int(items[1]), mp_obj_get_int(items[2]), mp_obj_get_int(items[3]), mp_obj_get_int(items[4]), mp_obj_get_int(items[5]));
-            tv.tv_usec = mp_obj_get_int(items[6]);
+            size_t seq_len;
+            mp_ob_get_array_min_max(args[1], 3, 8, &seq_len, &items);
+            mp_int_t hour = seq_len > 3 ? mp_obj_get_int(items[3]) : 0;
+            mp_int_t minute = seq_len > 4 ? mp_obj_get_int(items[4]) : 0;
+            mp_int_t second = seq_len > 5 ? mp_obj_get_int(items[5]) : 0;
+            mp_int_t subsecond = seq_len > 6 ? mp_obj_get_int(items[6]) : 0;
+            tv.tv_sec = timeutils_seconds_since_epoch(mp_obj_get_int(items[0]), mp_obj_get_int(items[1]), mp_obj_get_int(items[2]), hour, minute, second);
+            tv.tv_usec = subsecond;
         } else {
+            mp_obj_get_array_fixed_n(args[1], 8, &items);
             tv.tv_sec = timeutils_seconds_since_epoch(mp_obj_get_int(items[0]), mp_obj_get_int(items[1]), mp_obj_get_int(items[2]), mp_obj_get_int(items[4]), mp_obj_get_int(items[5]), mp_obj_get_int(items[6]));
             tv.tv_usec = mp_obj_get_int(items[7]);
         }
--- a/ports/esp32/machine_rtc.c
+++ b/ports/esp32/machine_rtc.c
@@ -101,7 +101,7 @@ static mp_obj_t machine_rtc_make_new(const mp_obj_type_t *type, size_t n_args, s
     return (mp_obj_t)&machine_rtc_obj;
 }
 
-static mp_obj_t machine_rtc_datetime_helper(mp_uint_t n_args, const mp_obj_t *args) {
+static mp_obj_t machine_rtc_datetime_helper(mp_uint_t n_args, const mp_obj_t *args, bool use_init_format) {
     if (n_args == 1) {
         // Get time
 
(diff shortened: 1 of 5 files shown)


sebromero:
Update: I also added the possibility to initialize the RTC only with the date. According to the docs the time related parameters are optional.

> (year, month, day[, hour[, minute[, second[, microsecond[, tzinfo]]]]]

New test:

```py
from time import sleep
from machine import RTC

rtc = RTC()

year = 2024
month = 12
day = 26
weekday = 3
hour = 6
minute = 30
second = 20
microsecond = 0
tzinfo = 0

def test(source, assert_time = True, should_fail=False):
    print(f"\nTesting {source}")
    # Read time
    t = rtc.datetime()
    print(t)
    try:
        assert t[0] == 2024, f"Year should be 2024 but was {t[0]}"
        assert t[1] == 12, f"Month should be 12 but was {t[1]}"
        assert t[2] == 26, f"Day should be 26 but was {t[2]}"
        assert t[3] == 3, f"Weekday should be 3 but was {t[3]}"
        if assert_time:
            assert t[4] == 6, f"Hour should be 6 but was {t[4]}"
            assert t[5] == 30, f"Minute should be 30 but was {t[5]}"
            assert t[6] == 20, f"Second should be 20 but was {t[6]}"
        if should_fail:
            print("❌ Test should have failed but didn't")
        else:
            print("✅ Test passed")
    except (AssertionError, ValueError) as e:
        if should_fail:
            print(f"✅ Test failed as expected: {e}")
        else:
            print("❌ Test passed but should have failed")

# Initialise the RTC. Datetime is a tuple of the form:
# (year, month, day[, hour[, minute[, second[, microsecond[, tzinfo]]]]]
rtc.init((year, month, day, hour, minute, second, microsecond, tzinfo))
test("init in order")

rtc.init((year, month, day, tzinfo, hour, minute, second, microsecond))
test("init out of order", should_fail=True)

try:
    rtc.init((year, month, day))
    test("init with only date", assert_time=False)
except ValueError as e:
    print(f"❌ Test failed: {e}")

try:
    print("\nTesting with too many arguments")
    rtc.init((year, month, day, hour, minute, second, microsecond, tzinfo, 1234))
    print("❌ Test passed but should have failed")
except ValueError as e:
    print(f"✅ Test failed as expected: {e}")


# The 8-tuple has the following format:
# (year, month, day, weekday, hours, minutes, seconds, subseconds)
rtc.datetime((year, month, day, weekday, hour, minute, second, microsecond))
test("datetime in order")
```

Output before the fix:
```
Testing init in order
(2024, 12, 27, 4, 6, 20, 0, 370)
❌ Test passed but should have failed

Testing init out of order
(2024, 12, 26, 3, 6, 30, 20, 248)
❌ Test should have failed but didn't
❌ Test failed: requested length 8 but object has length 3

Testing with too many arguments
✅ Test failed as expected: requested length 8 but object has length 9

Testing datetime in order
(2024, 12, 26, 3, 6, 30, 20, 223)
✅ Test passed
```

Output with the fix:
```
Testing init in order
(2024, 12, 26, 3, 6, 30, 20, 381)
✅ Test passed

Testing init out of order
(2024, 12, 26, 3, 0, 6, 30, 311)
✅ Test failed as expected: Hour should be 6 but was 0

Testing init with only date
(2024, 12, 26, 3, 0, 0, 0, 218)
✅ Test passed

Testing with too many arguments
✅ Test failed as expected: requested maximum length 8 but object has length 9

Testing datetime in order
(2024, 12, 26, 3, 6, 30, 20, 242)
✅ Test passed
```

To allow for optional parameters but still checking the required tuple size range I added a new function that checks for the min and max size of the given tuple.

dpgeorge:
Thanks for the patch, and the comprehensive test.

Sorry that you ran into this issue.  Indeed the docs don't match the behaviour, although it's not clear which is the "source of truth here" :sweat_smile: 

The mimxrt and samd ports also implement `rtc.init()` in the same way as the esp32 port (stm32 and renesas-ra have `rtc.init()` but it takes no arguments, it just resets the RTC peripheral).  So either they all need changing, or the docs need changing.

Actually, PR #10607 attempts to fix all the affected ports, and fix some other inconsistencies.  So I think that might be a good solution here?

dpgeorge:
I should also say that we do eventually want to fix this properly, by making all the methods follow the standard CPython tuple for date-time, namely `(year, month, day, hours, minutes, seconds, milliseconds[, tzinfo])`.  That should hopefully reduce confusion because there will be only one way to specify the date-time.

For this change we either need to define completely new RTC methods that use this CPython tuple format, or wait for MicroPython 2.0.

dpgeorge:
The `RTC.init()` argument order is fixed by f4e45995232ad52aeb4da270e6ca7571695e339b

@sebromero would be good to get your test merged, so it can test this feature.  And best if it was made generic, for any port, not just esp32.  Do you want to attempt that?  If not I can take it from here.

sebromero:
@dpgeorge Sure I can give it a shot. Some guidance needed though:

- Is it correct that we only test the C code by means of testing the uPython code that uses it behind the scenes?
- Is there a way to exclude the ports in the test that don't support RTC `init` with parameters? I could also try catch that case in the test but feels a bit quirky.
- Is there any value in keeping the `mp_ob_get_array_min_max` function that I created? It's meant to check for min/max tuple sizes in the case of a list of parameters with optional entries. I couldn't find such function, we only had `mp_obj_get_array_fixed_n`.

dpgeorge:
> * Is it correct that we only test the C code by means of testing the uPython code that uses it behind the scenes?

Yes.  Pretty much all tests are Python code running on the device (or unix port etc).  And that Python code runs the underlying C code and tests it.

(One exception: the unix coverage build has `coverage.c` and `coveragecpp.cpp` that do C-level testing.)

> * Is there a way to exclude the ports in the test that don't support RTC `init` with parameters? I could also try catch that case in the test but feels a bit quirky.

The best way is to use a SKIP mechanism at the start of the test, eg:
```py
try:
    from machine import RTC

    RTC().init(0, 0, 0, 0, 0, 0, 0, 0)
except:
    print("SKIP")
    raise SystemExit
```

That is a little quirky but lots of other tests do this.  Alternatively, for this case you know that the ports that don't conform are `stm32` and renesas-ra` so you could do:
```py
try:
    from machine import RTC
except:
    print("SKIP")
    raise SystemExit

import sys

if not hasattr(RTC, "init") or sys.platform in ("pyboard", "renesas-ra"):
    print("SKIP")
    raise SystemExit
```

You could also use `unittest` to implement the test, see eg `tests/extmod/machine1.py`.



> * Is there any value in keeping the `mp_ob_get_array_min_max` function that I created?

Not at this stage.



> * Would you like me to reuse this PR or should I create a new one just with the test?

It doesn't really matter.

dpgeorge:
A generic RTC test that runs on all ports now exists at `tests/extmod/machine_rtc.py` (actually, I think that was there prior to this PR being opened).  It tests pretty much the same things as the test added in this PR, except for too few/too many arguments to `RTC.init()`.  So I'll close this PR.

PR #16358, opened 2024-12-04
ports/stm32: Enable FIFO for H7 controller.

ennyKey:
On my board I sometimes have the problem that single bytes get lost during UART communication at 115200 baud. I use the `UART.read()` method and see that sometimes single bytes are missing in the middle of the data.
The reason is that sometimes the receive interrupt of the UART is not called fast enough before the next byte is received.
Fortunately, the H7 controller has a hardware FIFO. This was not used in the existing implementation.
I enabled the FIFO in the code for the H7 controller and have never had a problem with missing bytes since.

As far as I know, using the FIFO has no negative effect on communication behaviour. Therefore, I do not think a configuration option is necessary at this point, and it can always be enabled.

I have tested the new implementation on a STM32H753 succesfully.

### DIFF
--- a/ports/stm32/Makefile
+++ b/ports/stm32/Makefile
@@ -417,6 +417,12 @@ HAL_SRC_C += $(addprefix $(STM32LIB_HAL_BASE)/Src/stm32$(MCU_SERIES)xx_,\
 	)
 endif
 
+ifeq ($(MCU_SERIES),$(filter $(MCU_SERIES),h7))
+HAL_SRC_C += $(addprefix $(STM32LIB_HAL_BASE)/Src/stm32$(MCU_SERIES)xx_,\
+    hal_uart_ex.c \
+    )
+endif
+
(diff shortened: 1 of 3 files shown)


ennyKey:
**Side note:**
I have seen that there is an option in uart.c line 657 which suggests that the FIFO is enabled for the STM32G4.

However, the `HAL_UART_Init(&huart)` does not activate the FIFO.
This must be done explicitly by calling `HAL_UARTEx_EnableFifoMode(&huart)`.

As far as I can see, the current implementation for the G4 should run without FIFO despite this line.
Unfortunately I don't have a G4 to check this. So I left the G4 code as it is and made the necessary additions for the H7 only.

dpgeorge (approved):
This looks like a good improvement, thanks!

The G4 case can be fixed in the future by someone who needs that and has a board to test with.

PR #16369, opened 2024-12-06
esp32/machine_pwm.c: Correctly stop pwm timer

garywill:
#### Problem

ESP32 PWM (LEDC) timer wansn't correctly stopped. We were using `ledc_timer_rst()` to stop timer, which is incorrect.
[This section of IDF doc](https://docs.espressif.com/projects/esp-idf/en/release-v5.2/esp32/api-reference/peripherals/ledc.html?highlight=ledc_timer_rst#more-control-over-pwm) says:

> -  ledc_timer_set()
> - **ledc_timer_rst()**
> - ledc_timer_pause()
> - ledc_timer_resume()
> 
> The **first two** functions are called "behind the scenes" by `ledc_channel_config()` to provide a **startup** of a timer after it is configured.

So `ledc_timer_rst()` is for resetting timer counter to zero, not for stopping the timer. 

The ESP32 technical reference manual (PDF v5.2) also says in Section 14.4:
> Register 14.13. LEDC_LSTIMERx_CONF_REG 
> LEDC_LSTIMERx_RST Used to reset low speed timer counter. Counter value is 0 after resetting

#### Fix

According to [IDF doc](https://docs.espressif.com/projects/esp-idf/en/release-v5.2/esp32/api-reference/peripherals/ledc.html?highlight=deconfigure#timer-configuration), the correct way to stop a PWM timer is:

> When a timer is no longer needed by any channel, it can be **deconfigured** by calling the same function `ledc_timer_config()`. The configuration structure `ledc_timer_config_t` passes in should be:
> - `ledc_timer_config_t::deconfigure` Set this to true so that the timer specified can be **deconfigured**

And [this section of IDF doc](https://docs.espressif.com/projects/esp-idf/en/release-v5.2/esp32/api-reference/peripherals/ledc.html?highlight=deconfigure#_CPPv4N19ledc_timer_config_t11deconfigureE) says:

> bool deconfigure
> Set this field to **de-configure a LEDC timer** which has been configured before.
> Also, the timer has to be **paused first** before it can be de-configured. When this field is set, duty_resolution, freq_hz, clk_cfg fields are ignored.

So the correct way to stop a pwm timer is:
1. Check if the timer is not used by MPY anymore
2. Pause the timer using IDF function
3. Deconfigure the timer using IDF function

### Testing

Tested with IDF 5.2.2 on ESP32

### DIFF
--- a/ports/esp32/machine_pwm.c
+++ b/ports/esp32/machine_pwm.c
@@ -156,7 +156,13 @@ static void pwm_deinit(int channel_idx) {
         int timer_idx = chans[channel_idx].timer_idx;
         if (timer_idx != -1) {
             if (!is_timer_in_use(channel_idx, timer_idx)) {
-                check_esp_err(ledc_timer_rst(TIMER_IDX_TO_MODE(timer_idx), TIMER_IDX_TO_TIMER(timer_idx)));
+                check_esp_err(ledc_timer_pause(TIMER_IDX_TO_MODE(timer_idx), TIMER_IDX_TO_TIMER(timer_idx)));
+                ledc_timer_config_t timer_config = {
+                    .deconfigure = true,
+                    .speed_mode = TIMER_IDX_TO_MODE(timer_idx),
+                    .timer_num = TIMER_IDX_TO_TIMER(timer_idx),
+                };
+                check_esp_err(ledc_timer_config(&timer_config));
                 // Flag it unused
                 timers[chans[channel_idx].timer_idx].freq_hz = -1;
             }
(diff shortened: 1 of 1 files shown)


IhorNehrutsa:
Thanks @garywill 
Applied in #10854

dpgeorge (approved):
Thanks for the fix.  Tested on ESP32 with the `tests/extmod_hardware/machine_pwm.py` test.  It works.

(Even though this has been integrated into #10854, it's worth getting this small fix in now.)

PR #13090, opened 2023-11-29
ci: Enable caching for esp32 port builds, split into multiple jobs

projectgus:
Changes to improve performance of esp32 build (previously sitting at 7-8 minutes per run, slowest CI step along with Zephyr.)

1. Cache the ESP-IDF checkout (including all submodules) and the ESP-IDF tools install. Cache is keyed per ESP-IDF version, and the cache action also scopes to each branch (although the default branch cache is available to other branches.) These steps take a little over 1 minute to run on cache miss, but restoring from a cache hit takes approx 15 seconds.
2. Use GitHub Actions build matrix to parameterise which function in ci.sh should be run. This allows splitting up the build steps with minimum copy-paste of boilerplate. With a warm IDF cache the current build jobs each takes around 3 minutes, and will usually run in parallel.
3. Enable [ccache](https://ccache.dev/) for ESP-IDF builds, using hendrikmuhs' [ccache action](https://github.com/hendrikmuhs/ccache-action/) because the cache should be refreshed after each build, the built-in cache action isn't quite enough. A warm ccache, this saves an extra ~40-60 seconds per MicroPython build (i.e. 80-120 seconds per job.) For example see [this job](https://github.com/micropython/micropython/actions/runs/7094430432/job/19309688449?pr=13090), finished in 1m30s.

There is also a commit here to change stm32 build to use the build matrix, for consistency with esp32 rather than because it significantly reduces reduces copy/paste.

PR #16380, opened 2024-12-10
ci: Add GitHub Actions caching for Zephyr

projectgus:
### Summary

Reduces Zephyr CI build times by approx 50%.

### Details

Zephyr is one of the slower-running CI jobs, partially as it has to pull a docker image and then clone the Zephyr repository (including submodules). This PR tackles this in a few places:

1. Cache the Zephyr Workspace (similar to ESP-IDF caching in #13090). The Zephyr workspace is about 2.5GB!!
2. ~~Cache the Docker image.~~ Turns out not feasible as the image is 15GB(!), and each repository only gets 10GB of cache in total. Have removed from this PR.
3. Switch the Docker image source to the GitHub Container Registry rather than Docker Hub. Benefit seems marginal, the "Install Packages" step that pulls the image takes between 2 and 3 minutes each time with a lot of variation. However I did see the fastest time yet (1m52s) pulling from the GitHub Registry so there might be some benefit there. :shrug: 
4. Added ccache persistence, similar to the ESP32 CI workflow (but a little more complex as the ccache directory needs to be mounted into the container).

### Testing

All up, the total Zephyr build job is down from (for example) [11m4s on master](https://github.com/micropython/micropython/actions/runs/12343941960/job/34445781228) to [5m17s on this branch with a warm cache](https://github.com/micropython/micropython/actions/runs/12366470093/job/34514261104).

A cold cache run takes about 40s longer than before (due to creating the workspace cache), but this should happen very infrequently (for example: the ESP-IDF v5.2.2 cache object is now 4 months old and gets reused many times every day.)

There is a lot of variation between runs (as much as 2-3 minutes), but there are consistently speedups resulting from changes 1 and 4 and a possible small speedup from change 3.

* I don't see other easy performance improvements, as 50% of the build time is now pulling the Zephyr CI image. We could probably get rid of this step entirely by downloading just the Zephyr SDK tarball (and caching it), as we don't need most of [what's in the CI image](https://github.com/zephyrproject-rtos/docker-image/blob/main/Dockerfile.ci#L135). However then we have the maintenance task of keeping it working through updates, instead of using the supported image.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -754,12 +754,15 @@ function ci_zephyr_setup {
     # Directories cached by GitHub Actions, mounted
     # into the container
     ZEPHYRPROJECT_DIR="$(pwd)/zephyrproject"
+    CCACHE_DIR="$(pwd)/.ccache"
 
     mkdir -p "${ZEPHYRPROJECT_DIR}"
+    mkdir -p "${CCACHE_DIR}"
 
     docker run --name zephyr-ci -d -it \
       -v "$(pwd)":/micropython \
       -v "${ZEPHYRPROJECT_DIR}":/zephyrproject \
+      -v "${CCACHE_DIR}":/root/.cache/ccache \
       -e ZEPHYR_SDK_INSTALL_DIR=/opt/toolchains/zephyr-sdk-${ZEPHYR_SDK_VERSION} \
       -e ZEPHYR_TOOLCHAIN_VARIANT=zephyr \
       -e ZEPHYR_BASE=/zephyrproject/zephyr \
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -747,7 +747,7 @@ ZEPHYR_SDK_VERSION=0.16.8
 ZEPHYR_VERSION=v3.7.0
 
 function ci_zephyr_setup {
-    IMAGE=zephyrprojectrtos/ci:${ZEPHYR_DOCKER_VERSION}
+    IMAGE=ghcr.io/zephyrproject-rtos/ci:${ZEPHYR_DOCKER_VERSION}
 
     docker pull ${IMAGE}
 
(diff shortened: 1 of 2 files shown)


projectgus:
Was trying to also cache the Zephyr CI docker image here, but it is 15GB(!) and each repository is limited to 10G of total caches... so that was never going to work :laughing: 

Have switched from Docker Hub to the GitHub Container Registry to see if fetching gets any faster (suspect it won't).

projectgus:
That... took longer than I thought it would to get right. :sweat_smile:

tools/ci.sh
```diff
@@ -749,7 +749,7 @@ ZEPHYR_SDK_VERSION=0.16.8
 ZEPHYR_VERSION=v3.7.0
 
 function ci_zephyr_setup {
-    IMAGE=zephyrprojectrtos/ci:${ZEPHYR_DOCKER_VERSION}
+    IMAGE=ghcr.io/zephyrproject-rtos/ci:${ZEPHYR_DOCKER_VERSION}
```
dpgeorge, line 752: Is this mirror kept up-to-date?

dpgeorge:
Thanks, this is a great improvement, the zephyr CI is definitely the slowest at the moment.

It's a big win so I think we should merge this, and if it needs further tuning/fixing in the future then we'll do that then.

projectgus:
I realised the ccache action needs a bit of work, as implemented it doesn't update the cache data on a hit which means the contents will get progressively irrelevant. Should be a quick fix, though.

tools/ci.sh
projectgus: Should be, both registries are published by the official Zephyr project and they link to both in their docs:
https://github.com/zephyrproject-rtos/docker-image/?tab=readme-ov-file#using-pre-built-developer-docker-image

projectgus:
@dpgeorge This should be ready now.

tools/ci.sh
dpgeorge: I see, great!

dpgeorge (approved):

PR #16381, opened 2024-12-10
renesas-ra/modrenesas: Expose the Flash block device in a renesas module.

robert-hh:
### Summary

Allowing to use Python methods for creating and modifying the file system. The Flash block device for the file system can be accessed with:

    from renesas import Flash
    bdev = Flash(start=0)

With the block device accessible for Python, the startup code for mounting and eventually creating the file system can be implemented e.g. as `_boot.py` script, replacing the existing C code and allowing to use for instance LFS1 as file system. That saves >12k of flash space and increases the available file system size from 19k to 34k at a RA4M1 device. Deactivating the C startup code can be configured in mpconfigboard.h with the settings:
```
// Disable the C startup code for mounting the file system
#define MICROPY_HW_FLASH_MOUNT_AT_BOOT (0)
// Disable the boot mode select. If needed, use Python code
#define MICROPY_BOARD_BEFORE_SOFT_RESET_LOOP(x)
// Use _boot.py to mount flash at boot and to handle boot mode selection.
#define MICROPY_BOARD_FROZEN_BOOT_FILE "_boot.py"
```
### Testing

Tested with a RA4M1 board (Weact RA4M1 core module) and a RA6M2 board.

### DIFF
--- a/ports/renesas-ra/Makefile
+++ b/ports/renesas-ra/Makefile
@@ -250,6 +250,7 @@ SRC_C += \
 	machine_pin.c \
 	machine_rtc.c \
 	machine_sdcard.c \
+	modrenesas.c \
 	extint.c \
 	usrsw.c \
 	flash.c \
--- a/ports/renesas-ra/modrenesas.c
+++ b/ports/renesas-ra/modrenesas.c
@@ -1,0 +1,41 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2019 Damien P. George
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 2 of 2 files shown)


robert-hh:
A sample _boot.py:
```
# Startup code _boot.py for mouting the file system
import gc
import os
import renesas
import sys
import vfs

bdev = renesas.Flash(start=0)
# Try to mount the filesystem, and format the flash if it doesn't exist.
try:
    fs_type = (
        vfs.VfsLfs2
        if hasattr(vfs, "VfsLfs2")
        else (vfs.VfsLfs1 if hasattr(vfs, "VfsLfs1") else vfs.VfsFat)
    )
    vfs.mount(fs_type(bdev), "/flash")
except:
    try:
        fs_type.mkfs(bdev)
        vfs.mount(fs_type(bdev), "/flash")
    except:
        pass

os.chdir("/flash")
sys.path.append("/flash")
sys.path.append("/flash/lib")

del bdev, os, renesas, sys, vfs
gc.collect()
del gc
```

andrewleech:
Would it be better to work towards a new `machine.Flash` API rather than a new port specific module?

I guess not every port will be able to support a raw block device access on this kind of API, though it would be certainly useful for the same reasons you outline. 

On a separate note have you made a board profile for the RA4M1 core module with the  USB defined?
If not I've got a local profile for a physically modded RA4M1 evk with native USB I've been meaning to rework for the weact core board.

robert-hh:
It is obviously possible to add Flash to machine. Only that it is in a port- named module in other ports.
About USB: Since the beginning, the renesas port deserves using USB for REPL  and stdio.  The arguing board uses USB, so the jmplemtation exists and can be extended for the other boards. Can you share your implementation for RA4M1? I have also a RA6M2 board here for testing.

robert-hh:
@andrewleech Here is the set of board files I used for the WEACT board. The generated files are hand-modified. The config tool did not work at my Linux PC. I was too lazy to fix it or to get a Windows machine. I tried to make the appropriate changes in the XML file. REPL is at UART9, the 8 pin connector at the top of the board.
[WEACT_RA4M1.zip](https://github.com/user-attachments/files/18079614/WEACT_RA4M1.zip)

andrewleech:
> Can you share your implementation for RA4M1?

Ah drats, turns out the evk I modified was RA4W1 not RA4M1 - I was able to use the config tool to make the modifications, I'll do that for the M1 when I get a chance. I've got the weact RA4M1 core board and have been meaning to do this anyway.

I do know the built in USB works on these chips at least, the modified board (just wiring up the usb cable to the required pins and updating the config to enable USB) was enough to test my PR a while back consolidating the tinyusb stdio stuff.

dpgeorge (approved):
Thanks for this patch.  I think it's fine, it brings this port to parity with other ports like stm32 and rp2 that expose a `Flash` object.

And anyway this object is already defined here in the renesas-ra port, it just somehow was never exposed to the user.

Yes, eventually it would be good to unify all the ports to expose something like `vfs.Flash`, but for now we can keep the pattern of having it in the port-specific module.

robert-hh:
Thank you.

PR #16383, opened 2024-12-10
ports/stm32/boards/STM32F429DISC: Fix SDRAM configuration.

iabdalkader:
Define SDRAM frequency and refresh cycles.


dpgeorge:
Thanks.

PR #16384, opened 2024-12-10
stm32/pin: add config options to exclude legacy items, alt function support, and cpu/board objects

dpgeorge:
### Summary

stm32's `Pin` class has a lot of legacy entries in it.  This PR adds config options to exclude them if desired by a board.  Also adds some other related options to exclude other pin entries:
- `MICROPY_PY_MACHINE_PIN_LEGACY` can be disabled to exclude all the methods and constants that are not part of the standard `machine.Pin` set
- `MICROPY_PY_MACHINE_PIN_ALT_SUPPORT` can be disabled to exclude support for alternate functions
- `machine.Pin.cpu` and `machine.Pin.board` are excluded automatically if they contain no entries

These are all enabled by default so nothing should change on existing board configurations.

### Testing

Tested on PYBV10, by disabling the options, and by removing all `Pin.board` names in `pins.csv`.

This is useful for a board to have fine control over what's available in `machine.Pin`.  Without these config options a board would need to copy all the `Pin` code in order to customise/exclude these legacy items.

### DIFF
--- a/ports/stm32/mpconfigboard_common.h
+++ b/ports/stm32/mpconfigboard_common.h
@@ -52,6 +52,11 @@
 #define MICROPY_PY_PYB_LEGACY (1)
 #endif
 
+// Whether to include legacy methods and constants in machine.Pin (which is also pyb.Pin).
+#ifndef MICROPY_PY_MACHINE_PIN_LEGACY
+#define MICROPY_PY_MACHINE_PIN_LEGACY (!MICROPY_PREVIEW_VERSION_2)
+#endif
+
 // Whether machine.bootloader() will enter the bootloader via reset, or direct jump.
 #ifndef MICROPY_HW_ENTER_BOOTLOADER_VIA_RESET
 #define MICROPY_HW_ENTER_BOOTLOADER_VIA_RESET (1)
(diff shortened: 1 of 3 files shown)


ports/stm32/mpconfigboard_common.h
```diff
@@ -52,6 +52,16 @@
 #define MICROPY_PY_PYB_LEGACY (1)
 #endif
 
+// Whether to include legacy methods and constants in machine.Pin (which is also pyb.Pin).
+#ifndef MICROPY_PY_MACHINE_PIN_LEGACY
+#define MICROPY_PY_MACHINE_PIN_LEGACY (1)
```
projectgus, line 57: Do we want to phase out these legacy items? If so, could make the default here `!MICROPY_PREVIEW_VERSION_2`

projectgus (approved):
One optional suggestion but the changes LGTM!

ports/stm32/mpconfigboard_common.h
dpgeorge: Yes, good idea, updated as suggested.

PR #16389, opened 2024-12-11
stm32/extint: Fix EXTI IRQ handlers for H5 MCUs.

dpgeorge:
### Summary

EXTI IRQ handlers are currently not working on STM32H5xx MCUs.

In this PR, the existing EXTI IRQ handlers are moved from `stm32_it.c` to `extint.c` to keep them with related code.  A macro is defined to make it easier to define the handler function that handles one line, and correct handlers added for STM32H5xx MCUs.

Also, to prevent errors in the future, `MP_STATIC_ASSERT(<irqn> > 0)` is added to each handler function to check that the correct `IRQn` constant is used, which corresponds to the handler function name.

### Testing

Tested on a H5 board, EXTI now works.

Existing boards have not been tested, but if they build then they should work because they are effectively unchanged.

### DIFF
--- a/ports/stm32/extint.c
+++ b/ports/stm32/extint.c
@@ -241,6 +241,97 @@ static const uint8_t nvic_irq_channel[EXTI_NUM_VECTORS] = {
     #endif
 };
 
+#define DEFINE_EXTI_IRQ_HANDLER(line) \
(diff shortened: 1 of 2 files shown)


projectgus (approved):
Great idea with the static assert!

PR #16395, opened 2024-12-11
ports/renesas-ra/Makefile: Fix binary file generation.

iabdalkader:
The linker scripts for most of these microcontrollers contain non-contiguous flash sections that result in big binary files, that exceed the flash size. This fixes the binary files by copying only the text and data sections.

For example, every `.bin` hosted here is 16 Mbytes https://micropython.org/download/EK_RA4M1/

### Testing

Tested uploading binaries to two RA boards, before it was impossible to upload the binary due to its size.

### DIFF
--- a/ports/renesas-ra/Makefile
+++ b/ports/renesas-ra/Makefile
@@ -442,7 +442,8 @@ endef
 
 define GENERATE_BIN
 	$(ECHO) "GEN $(1)"
-	$(Q)$(OBJCOPY) -I ihex -O binary $(2) $(1)
+	$(Q)$(OBJCOPY) -O binary -j .id_code $(2) $(BUILD)/id_code.bin
+	$(Q)$(OBJCOPY) -O binary --remove-section=.id_code $(2) $(1)
 endef
 
 define GENERATE_HEX
@@ -452,7 +453,7 @@ endef
 
 .PHONY:
 
-$(BUILD)/firmware.bin: $(BUILD)/firmware.hex
+$(BUILD)/firmware.bin: $(BUILD)/firmware.elf
 	$(call GENERATE_BIN,$@,$^)
 
 $(BUILD)/firmware.hex: $(BUILD)/firmware.elf


dpgeorge:
Is it possible to fix this differently and remove the unwanted data from the hex file?  If it's included in the hex then technically it should be flashed to the device.  If it's data that should not be there then it should be removed from the .hex.

It looks like the culprit is `.id_code` at 0x0100a150, and can be removed via:
```diff
--- a/ports/renesas-ra/boards/EK_RA6M2/ra6m2_ek.ld
+++ b/ports/renesas-ra/boards/EK_RA6M2/ra6m2_ek.ld
@@ -275,13 +275,6 @@ SECTIONS
         KEEP(*(.data_flash*))
         __Data_Flash_End = .;
     } > DATA_FLASH
-
-    .id_code :
-    {
-        __ID_Code_Start = .;
-        KEEP(*(.id_code*))
-        __ID_Code_End = .;
-    } > ID_CODE
 }
 /* produce a link error if there is not this amount of RAM for these sections */
 /* _minimum_stack_size = 2K; */
```

iabdalkader:
> It looks like the culprit is `.id_code` at 0x0100a150, and can be removed via:

Yes, the ID code is used for flash protection feature, which is not enabled, so it should be safe to remove:

> When ID code protection is enabled, the device validates or invalidates the ID code sent from the host by comparing it with the ID code stored in the flash memory.

However, the hex file doesn't contain this gap, so it could be used to program it if it's needed. Either way is fine with me.

dpgeorge:
> Yes, the ID code is used for flash protection feature

OK, thanks for explaining.  Then I think it's best to keep the ID_CODE section, in case that feature is enabled in the future.

dpgeorge:
Checking the generated .bin files with the .hex converted to bin, there is a difference: the .bin file is missing some code.  Eg:
```
$ make
$ cd build-EK_RA6M2
$ arm-none-eabi-objcopy -I ihex -O binary firmware.hex firmware2.bin
$ xxd firmware.bin > a
$ xxd firmware2.bin > b
$ diff a b
(some differences)
```
It looks like the `FLASH_OP` section is missing from the bin and needs to be added to the list for the bin generation from elf.

But this is a bit worrying, what other sections might be missing?  It wasn't easy to tell that `FLASH_OP` was the missing part.

So this is why I think it's actually best to generate the bin from the hex, then you don't need to know what list of sections need to be included.

iabdalkader:
I'm not sure I follow. I think you want the second commit then? Which removes the ID section..

dpgeorge:
> I'm not sure I follow. I think you want the second commit then? Which removes the ID section..

I'm not sure exactly what's best here.  It would be good to keep the ID section, in case that's ever used/needed.  But then, how to cleanly generate the bin file, without missing any sections?

I guess if the ID section is needed, then bin files won't work at all because they won't contain the correct ID key.  Do we even need the bin files?  The deploy instructions specify how to use JLink to flash a .hex file.

iabdalkader:
> I guess if the ID section is needed, then bin files won't work at all because they won't contain the correct ID key.

Exactly, it's impossible to include the ID section in a bin anyway, so I'd say the first commit is right, which keeps the flash ID section, and drops it when generating bins. If the flash ID is needed and enabled, then `hex` can still be used.

dpgeorge:
> so I'd say the first commit is right, which keeps the flash ID section, and drops it when generating bins. If the flash ID is needed and enabled, then `hex` can still be used.

Yes, I agree.

But as I mention above, the first commit still needs fixing to at least include the `FLASH_OP` section.  I'm not sure if there are other sections that were also missed.

iabdalkader:
@dpgeorge Maybe also add `.noinit` ?

dpgeorge (approved):
This approach is much better.

PR #16398, opened 2024-12-11
C3 Qemu support

mzakharo:
### Summary

Add support for C3 target to run on QEMU emulator with LAN networking support

### Testing

Tested with IDF 5.3.2

`make BOARD=ESP32_GENERIC_C3 BOARD_VARIANT=QEMU qemu`
```python
>>> import network
>>> lan = network.LAN(phy_type=network.PHY_OPENETH, phy_addr=1)
>>> lan.active(True)
>>> lan.ipconfig('addr4')
('10.0.2.15', '255.255.255.0')
```

`esp_efuse_rtc_calib_get_ver()`, part of `efuse` component, does not play nicely with the emulator.  Presented here is an example workaround for the issue, outlined here: https://github.com/espressif/esp-insights/issues/38


ports/esp32/boards/ESP32_QEMU_C3/board.md
```diff
@@ -0,0 +1,7 @@
+The following files are firmware images that should work on most
+ESP32-C3-based boards with 4MiB of flash, including WROOM and MINI modules,
+that use the revision 3 silicon (or newer).
+
+USB serial/JTAG support is enabled on pin 18 and 19. Note that this
+is not a full USB stack, the C3 just provides a CDC/ACM class serial
+and JTAG interface.
```
dpgeorge, line 7: So, this board also works on hardware as well as qemu?  Won't the ETH break on hardware?

ports/esp32/network_lan.c
```diff
@@ -232,6 +235,13 @@ static mp_obj_t get_lan(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_ar
             self->phy = esp_eth_phy_new_ksz80xx(&phy_config);
             break;
         #endif
+        #if CONFIG_ETH_USE_OPENETH
```
dpgeorge, line 238: Please split this addition of `PYH_OPENETH` into a separate commit within this PR.

ports/esp32/main.c
```diff
@@ -257,3 +257,15 @@ void *esp_native_code_commit(void *buf, size_t len, void *reloc) {
     memcpy(p, buf, len);
     return p;
 }
+
+
+// Workaround for https://github.com/espressif/esp-insights/issues/38
+extern int  __real_esp_efuse_rtc_calib_get_ver(void);
+int __wrap_esp_efuse_rtc_calib_get_ver(void)
+{
+    #if MICROPY_PI_QEMU
```
dpgeorge, line 266: I don't know if this is the correct name for such a config option.

Is there a macro option that the IDF defines when building for qemu that can be used here instead?

ports/esp32/Makefile
```diff
@@ -106,6 +106,12 @@ size-components:
 size-files:
 	$(call RUN_IDF_PY,size-files)
 
+qemu:
+	$(call RUN_IDF_PY,qemu monitor)
+
+menuconfig:
+	$(call RUN_IDF_PY,menuconfig)
```
dpgeorge, line 113: Is this `menuconfig` target needed?

dpgeorge (commented):
Thanks, this looks interesting!  I did a very quick review so maybe more changes are needed.

ports/esp32/Makefile
mzakharo: Will remove

ports/esp32/main.c
mzakharo: The idf build has nothing special when targeting qemu. Enabling openeth is the only way to tell it's a qemu build

ports/esp32/network_lan.c
mzakharo: Will do

ports/esp32/boards/ESP32_QEMU_C3/board.md
mzakharo: We can make it a variant of esp32 generic c3 instead.

ports/esp32/main.c
mzakharo: switched it to CONFIG_ETH_USE_OPENETH instead and removed custom config option

ports/esp32/boards/ESP32_QEMU_C3/board.md
mzakharo: OPENETH usage will break on hardware, other than that the build is compatible with both.

projectgus: I've marked as un-resolved to keep track of this:  either this should be a board variant, or this text should be changed to describe what this "board" is rather than being copy-paste from the C3 board.

ports/esp32/boards/ESP32_QEMU_C3/board.json
```diff
@@ -0,0 +1,19 @@
...
+        "External Flash",
+        "WiFi"
+    ],
+    "images": [
+        "esp32c3_devkitmini.jpg"
+    ],
+    "mcu": "esp32c3",
+    "product": "ESP32-C3",
+    "thumbnail": "",
+    "url": "https://www.espressif.com/en/products/modules",
```
projectgus, line 17: Suggest changing some of these to reflect what this actually is,

i.e. `url` can link to https://docs.espressif.com/projects/esp-idf/en/latest/esp32c3/api-guides/tools/qemu.html, `product` can be "ESP32-C3 QEMU Emulator", etc.

ports/esp32/boards/ESP32_QEMU_C3/board.md
mzakharo: Changed to board variant

ports/esp32/boards/ESP32_QEMU_C3/board.json
mzakharo: changed to board variant instead

dpgeorge (approved):
Thanks for updating, and sorry this took ages to get back to.

It looks good now, nice and simple as a board variant.

I tested on IDF 5.5.2 and it works well (after installing their custom qemu fork).

PR #15937, opened 2024-09-30
qemu: Enable more via-mpy and via-mpy with native emitter tests

dpgeorge:
### Summary

The qemu port has recently undergone a lot of improvement, including adding of RISC-V support and improved test running to use the standard bare-metal test approach.

One of the reasons to improve the test runner was to be able to enable more native emitter tests on the qemu port, by leveraging the `--via-mpy` and `--emit native` arguments to `run-tests.py`.

That can now be done, and is now done in this PR.

### Testing

Tested locally with `make test_full`.  Updated CI to use this new `test_full` make target.

### Trade-offs and Alternatives

This will make CI a bit slower but not by much, and the benefit is well worth it (better test coverage of the ARM and RISC-V native emitters).

PR #16400, opened 2024-12-12
qemu/mpconfigport: Enable VFS reader, loading .mpy files and io.IOBase.

dpgeorge:
### Summary

This change allows importing from the VFS, and enables the following 7 additional tests: builtin_execfile, io_iobase, json_dump_iobase, import_mpy_invalid, import_mpy_native, import_mpy_native_gc, vfs_userfs.

This is split out from #8381 (ROMFS).  In that PR it's needed for the qemu port to be able to test ROMFS features.

### Testing

These 7 new tests will run on CI.

This increases the size of the qemu port, but IMO that's OK, it's intended for testing, and this gets it testing more things.


projectgus:
Is anything else needed for those tests to run in CI? I either see them not mentioned or skipped in https://github.com/micropython/micropython/actions/runs/12296239476/job/34314834928

dpgeorge:
> Is anything else needed for those tests to run in CI? I either see them not mentioned or skipped in

If you go to that action, and expand the "Build and run test suite" bit, and then search for "import_mpy" you should see at least those 3 new tests passing.  And the other 4 should also be there.

dpgeorge:
See also #15937, which enables more mpy tests on qemu.  This PR is one of the commits of that PR.

projectgus:
> If you go to that action, and expand the "Build and run test suite" bit, and then search for "import_mpy" you should see at least those 3 new tests passing. And the other 4 should also be there.

Ah, my mistake. That's what I did but I was searching with Firefox Ctrl-F which doesn't work (even after scrolling the whole log), because it's only partially loaded somehow. If I use the search box on the page, or "View Raw Log" and Ctrl-F in that, then I can see the tests. :facepalm:

projectgus (approved):

PR #16401, opened 2024-12-13
add MPY v11 board and acceleration is new ic LIS2DH12TR. And optimize pyb.Accel method filtered_xyz.

wangyoufeng8889:


wangyoufeng8889 (commented):
test done in stm32F405RGT6 LIS2DH12TR board

PR #16412, opened 2024-12-14
rp2/modmachine: Fix lightsleep returning early.

### LINKED ISSUE #16181: lightsleep on Pico W regression for 1.24.0

cpottle9:
fixes #16181

Commit 74fb42a added code using hardware timer 2.
Prior to that only lightsleep used hardware timers specifically timer 3.
It turned off almost all hardware blocks except hardware timers and waits for an interrupt.
It did not check if the interrupt is from hardware timers or if it is from timer 3.

With this change, lightsleep now loops calling __wfi() until timer 3 has fired. Without the loop lightsleep would return very early.

Ideally this pull request should be reviewed by user @projectgus  who made the changes in commit 74fb42a and has also made changes to modmachine.c lightsleep.

### Testing
Tested on PICO W using thonny with code from the issue. Works fine.

### DIFF
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -245,8 +245,11 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
         #endif
         #endif
 
-        // Go into low-power mode.
-        __wfi();
+        // A timer other than #3 could fire.
+        // Repeatedly go into low-power mode until timer 3 is no longer armed.
+        while (timer_hw->armed & (1u << alarm_num)) {
+            __wfi();
+        }
 
         if (!timer3_enabled) {
             irq_set_enabled(irq_num, false);
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -197,6 +197,7 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
         xosc_dormant();
     } else {
         bool timer3_enabled = irq_is_enabled(3);
+        uint64_t endtime_us_64 = timer_time_us_64(timer_hw) + delay_ms * 1000;
 
         const uint32_t alarm_num = 3;
         const uint32_t irq_num = TIMER_ALARM_IRQ_NUM(timer_hw, alarm_num);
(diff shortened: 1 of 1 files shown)


ports/rp2/modmachine.c
```diff
@@ -245,8 +245,11 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
         #endif
         #endif
 
-        // Go into low-power mode.
-        __wfi();
+        // A timer other than #3 could fire.
+        // Repeatedly go into low-power mode until timer 3 is no longer armed.
+        while (timer_hw->armed & (1u << alarm_num)) {
+            __wfi();
```
dpgeorge, line 251: I'm pretty sure there's a race condition in here: between the check of the timer and the wfi the interrupt may go off, in which case the wfi will sit there waiting "forever" because the interrupt already fired.

cpottle9: Agreed. 
Needs to be fixed, but it won't wait forever. Timer 2 is going off about every 100 ms which will cause the __wfi() to return.

@dpgeorge, it is not obvious to me how to resolve the race condition.
I still a relative neophyte on ARM and micropython internals.

dpgeorge: @peterharperuk tried to fix this before but there's not really any way to do it.

A proper fix needs the pico-sdk update (which fixes issues with the alarm pool and wfe), but the latest version of that has a new alarm pool bug...

peterharperuk: We've fixed the latest issue and have a tentative plan to do a small sdk update in the new year.

cpottle9: @peterharperuk, it does not prevent wakeup for USB interrupts it just delays it. Potentially a lot.

Without my change the single __wfi() wakes up every 60 to 100 milliseconds due to the change in https://github.com/micropython/micropython/commit/74fb42aa82d537b6fefb463907274bd660c6ce8d which added code using hardware timer 2.
USB interrupts will be masked for that time.

With the while loop I added USB interrupts will be masked for the entire delay time.
But they will get serviced eventually. 

I did notice while testing that if I did a lightsleep(10000) and hit the thonny stop button the prompt would not come back until the sleep expired.
Longer delays like lightsleep(30000) and thonny would timeout if I hit the stop button.
But once the lightsleep completed (after 30 seconds) hitting the stop button again would bring it back.

I suggest to you that someone calling lightsleep() for a long time (EG: several minutes) needs to be aware thonny will timeout.

cpottle9: @projectgus has a pull request open for the same issue.

Likely I will close this pull request without a merge and his fix will go in. 

It is end of day were I live. I will get back to this tomorrow.

cpottle9:
I tested @projectgus changes in #16431. 
It fixes the issue and is cleaner than my fix.

PR #16421, opened 2024-12-16
esp32: Add LAN8670.

dpgeorge:
### Summary

This adds support for LAN8670 to the esp32 port, when built with ESP-IDF v5.3 or newer.

Addresses issue #15731.

### Testing

This is untested.

Merging this PR won't cause LAN8670 support to be available in the nightly builds until we update the ESP-IDF version there to V5.3.x. This will probably happen after the MicroPython 1.25.0 release.


dpgeorge:
@JohannesMaierhofer @DuaneKaufman are you able to test this PR?

DuaneKaufman:
Dear All, 

I have a pair of Two-Wire ETH Click boards from Mikroe ( [ https://www.mikroe.com/two-wire-eth-click | https://www.mikroe.com/two-wire-eth-click ] ), which are Microchip LAN8651 based and SPI-connected to the micro-controller board. 

I am not sure how much the LAN8651 is similar to the LAN867x chip, but I was leaning towards the LAN8651 for its ability to use SPI. 

If the LAN8651 can be used just like a LAN867x, what would an example SPI-based Micropython look like? 

Sincerely, 
Duane

xorbit:
@dpgeorge I tested this on a board with LAN8671 chip and when I try to create the LAN object I get this error:

```
>>> lan = network.LAN(mdc = machine.Pin(8), mdio = machine.Pin(7),
...                   power = None, phy_type = network.PHY_LAN8670, phy_addr=0)
OSError: esp_eth_driver_install failed with invalid argument
```

Which is odd, because on the same board, a previously hacked together proof-of-concept MicroPython build initializes the LAN8671 as expected without error.  Not sure what the difference is.

xorbit:
Oh, if I run it again after that failure, I get an extra error:

```
>>> lan = network.LAN(mdc = machine.Pin(8), mdio = machine.Pin(7),
...                   power = None, phy_type = network.PHY_LAN8670, phy_addr=0)
E (2977611) esp_netif_lwip: esp_netif_new_api: Failed to configure netif with config=0x3ffd74fc (config or if_key is NULL or duplicate key)
OSError: esp_eth_driver_install failed with invalid argument
```

projectgus:
> I tested this on a board with LAN8671 chip and when I try to create the LAN object I get this error:

@xorbit There was a bug, the LAN8671 was accidentally grouped with the SPI/Ethernet PHYs so it failed to install. I've pushed an updated commit, but I'm not able to test it apart from verifying that the install step passes.

ports/esp32/main_esp32/idf_component.yml
```diff
@@ -1,5 +1,6 @@
 ## IDF Component Manager Manifest File
 dependencies:
+  espressif/lan867x: "^1.0.0"
```
projectgus, line 3: This is annoying because the component depends on idf >= 5.3.0 so the build will fail on older ESP-IDF, and AFAIK there's no notion of a conditional dependency in the IDF Component Manager Manifest format.

I think our options are:

1. Hold this PR until we only support IDF 5.3.0 or newer
2. Generate this file as part of the build. :grimacing: 
3. Add a second `main_esp32` component (i.e. `main_esp32_old`) and pick one or the other in the top-level CMakeLists.txt file depending on the ESP-IDF version.

I can have a crack at option 2 if you want to try that path.

xorbit:
I tried this new commit and it seems to work on my boards. 👍

However, I had some problem with the `make submodules` not pulling in berkeley db, I needed to apply this fix first:
https://github.com/orgs/micropython/discussions/10658

Anyway, that's not really related to this PR I think.

I was wondering if it would make sense to have the `idf_component.yml` defined per board?

If not, how many boards fail to build with IDF 5.3+?

DuaneKaufman:
Dear Patrick, 

Are your LAN8670 boards custom or purchased? I bought two LAN8670 eval boards, but have not gotten around to testing yet. 

Sincerely, 
Duane

xorbit:
@DuaneKaufman it is a custom board that hopefully soon will become a product.  The timing of which may hinge on this PR making it into mainline MicroPython. 😆

projectgus:
> I was wondering if it would make sense to have the `idf_component.yml` defined per board?

Probably not, we already have multiple copy-paste copies and changing to that we'd have dozens. However, we could maybe provide a way to override `idf_component.yml` for a particular board.

(I think you can already a custom component to a board via `mpconfigboard.cmake`, and it can have its own `idf_component.yml` file. However that's likely not even a per-board solution for this case as the "main" `network.LAN` implementation needs to know about the PHY component...)

> If not, how many boards fail to build with IDF 5.3+?

Probably none, but the question is is more "are there subtle behaviour changes or regressions to some when built with IDF 5.3+?"

Updating the minimum IDF version boils down to two things:

* Currently the nightly builds are done with IDF 5.2.2. We can switch this up to 5.3 at some point, but there's some risk of a subtle regression that tests haven't found. So ideally we don't do this right before a release (for example).
* After doing that, we could drop support for IDF 5.2.x and anyone building their own board will need to update ESP-IDF. We can do this whenever we want, but it's not particularly convenient for folks (and we only dropped support for <5.2 six weeks ago, in #16128).

There's also the bigger picture view, which is that if MicroPython has a need to keep adding support for features distributed via ESP-IDF component manager, and if these components may have particular version or platform requirements, then we may need a way to support them flexibly.

ports/esp32/main_esp32/idf_component.yml
dpgeorge: I suggest we go for option 3, it's the simplest way to move forward.

projectgus:
I've reworked this PR so that it builds a new `main_esp32_idf53` component if IDF>=5.3 is in use, and the old `main_esp32` component otherwise. When we drop support for ESP-IDF 5.2.x these components can be merged again.

Nightly builds currently use ESP-IDF 5.2.2 so they won't contain LAN8670 support immediately after this merges. To get support you'll need to build MicroPython locally with a newer ESP-IDF version (5.3.x or 5.4.x).

Suggest after 1.25.0 releases that we bump the build version (for nightly and CI) to use ESP-IDF 5.3.2.

projectgus:
> I've reworked this PR so that it builds a new `main_esp32_idf53` component if IDF>=5.3 is in use, and the old `main_esp32` component otherwise. When we drop support for ESP-IDF 5.2.x these components can be merged again.

Turns out I didn't look hard enough at the Component Manager docs, we can do [conditional dependencies just fine](https://docs.espressif.com/projects/idf-component-manager/en/latest/reference/manifest_file.html#conditional-dependencies). :facepalm:  So have simplified to this!

I think we can use this technique to collapse the per-target main_xyz components down to one main component again. Have made a note to look at this when I have a chance.

xorbit:
Is there a reason this is not being merged?  I'm waiting on this functionality to create a new board definition.

projectgus:
> Is there a reason this is not being merged? I'm waiting on this functionality to create a new board definition.

Thanks for being patient. The 1.25 release has run behind and this feature is slated for 1.26. 1.25 should be released very soon, at which point we'll start merging 1.26 changes.

I've just rebased this PR on master to resolve the merge conflict.

xorbit:
Thank you for the update, I just wanted to ensure this wasn't lost in the shuffle. 😊

dpgeorge:
IDF is now at 5.4.1 and this PR is now merged.  LAN8670 should be supported by the ESP32 (only) in latest downloads.

xorbit:
Awesome, thank you!

DuaneKaufman:
Yes, this is good news - thanks!

PR #16428, opened 2024-12-16
rp2/spi: Allow omitting of unused pins

msalau:
### Summary

The change makes it possible to omit unused SPI pins.
This can be achieved by setting a pin as `None`.

### Testing

The change as been tested on a custom RP2040 board with only SCK and MOSI used to drive a SPI IC.

The commit changes behavior if the constructor method.
Previously if a pin was set as `None` the default pin number for this interface has been used.

### DIFF
--- a/ports/rp2/machine_spi.c
+++ b/ports/rp2/machine_spi.c
@@ -41,12 +41,12 @@
 #ifdef MICROPY_HW_SPI_NO_DEFAULT_PINS
 
 // With no default SPI, need to require the pin args.
-#define MICROPY_HW_SPI0_SCK     (0)
-#define MICROPY_HW_SPI0_MOSI    (0)
-#define MICROPY_HW_SPI0_MISO    (0)
-#define MICROPY_HW_SPI1_SCK     (0)
-#define MICROPY_HW_SPI1_MOSI    (0)
-#define MICROPY_HW_SPI1_MISO    (0)
+#define MICROPY_HW_SPI0_SCK     (-1)
+#define MICROPY_HW_SPI0_MOSI    (-1)
+#define MICROPY_HW_SPI0_MISO    (-1)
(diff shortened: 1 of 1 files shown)


Gadgetoid:
Missed that I had effectively duplicated this here: https://github.com/micropython/micropython/pull/16922

In most cases users would be disabling MISO, for the reasons given in my PR, but it's feasible that MISO *or* MOSI could be disabled for single direction SPI (read a read only device, write a write only device).  Not to be confused for half duplex mode.

sck though? That's a tough sell!

robert-hh:
It seems that a similar change was merged today: commit dd7a950bb.

dpgeorge:
Ahh, sorry, I forgot about this PR!

@msalau if you want to update this PR to make all pins optional, please follow the logic used in dd7a950bbc27713df8cc461958757ec793189d72


> sck though? That's a tough sell!

Disabling SCK is potentially useful, if you want to generate a specific bit stream on MOSI, eg to drive a Neopixel.

robert-hh:
> Disabling SCK is potentially useful, if you want to generate a specific bit stream on MOSI,

I made a HX711 driver once which uses MOSI to create the clock signal of 25-27 clock pulses. So the SPI SCK was not used. But applications are rare, and it this case using GPIO bit banging was better. The only advantage of using SPI was a reliable timing.
As far as I recall, I tried once to use SPI/MOSI to drive a Neopixel, taking advantage of the fact, that only the leading high pulses duration is important. The low period can be extended. So you can use 4 clock periods per Neopixel bit. But it needs 12 or 16 byte data per Neopixel, so it's not memory-efficient.

msalau:
Thanks for pointing out that similar feature is already in repo.
I need only to skip MISO but added other pins 'just in case' :)
I'm closing the pull request then.

Issue #7035, opened 2021-03-15
rp2: Can't wake from lightsleep with Pin IRQ


peterhinch:
On a pin change this code sample toggles the LED but never outputs "got here".
```python
from machine import Pin, lightsleep
p25 = Pin(25, Pin.OUT)
p0 = Pin(0, Pin.IN, Pin.PULL_UP)
def foo(_):
    p25(not p25())  # Toggle the LED

p0.irq(foo, trigger=Pin.IRQ_FALLING)
while True:
    lightsleep()
    print('Got here')
```
The [docs](http://docs.micropython.org/en/latest/library/machine.Pin.html#machine.Pin.irq) indicate that to fix this, the IRQ should be instantiated with `wake=machine.SLEEP`. Unfortunately this constant does not exist.

cipy:
well, it looks like the docs is a bit ahead of its time. also there is no wake or priority parameter to .irq()

>>> import machine
>>> dir(machine)
['__class__', '__name__', 'ADC', 'I2C', 'PWM', 'PWRON_RESET', 'Pin', 'SPI', 'SoftI2C', 'SoftSPI', 'Timer', 'UART', 'WDT', 'WDT_RESET', 'bootloader', 'deepsleep', 'disable_irq', 'enable_irq', 'freq', 'idle', 'lightsleep', 'mem16', 'mem32', 'mem8', 'reset', 'reset_cause', 'soft_reset', 'time_pulse_us', 'unique_id']
>>>

tomjorquera:
For what I have seen, support does not seems implemented yet for the pico.

From my experiments, it seems that in the meantime you could try experimenting with using a function such as

```
@micropython.asm_thumb
def my_lightsleep():
    wfi()
```

`wfi` puts the pico into a suspended state until any interrupt is received (see https://docs.micropython.org/en/latest/reference/asm_thumb2_misc.html).

In itself it may not save a lot of energy, but it could serve as a workaround.

For my use case, I want to use the deepsleep mode, which does not seems to be supported yet either. So I have started cobbling together a small experimental lib to use while waiting for the feature to be added in micropython. If any tinkerer feels adventurous and want to give it a run and help improving things they are very welcome:

https://github.com/tomjorquera/pico-micropython-lowpower-workaround

jmfloyd:
[...]

PR #16431, opened 2024-12-17
rp2: Fix soft timer expiry waking early from lightsleep.

### LINKED ISSUE #16181: lightsleep on Pico W regression for 1.24.0

projectgus:
### Summary

Fixes #16181 and partial fix for #16180. Specifically: bug where PICO-W build won't lightsleep() for more than 64ms.

### Details

This is a regression introduced in 74fb42a (#13329) when we switched away from the pico-sdk alarm pool for soft timers and accidentally made soft timer expiry a wakeup source.

Before 74fb42a (including V1.23), both the pico-sdk alarm pool and the lightsleep wakeup timer uses alarm 3. The lightsleep timer would quietly clobber alarm 3's timeout, meaning softtimer events wouldn't wake the chip from lightsleep.

After 74fb42a (including V1.24), soft timer wakeup happens on timer alarm 2 so this interrupt can wake the chip from light sleep. On PICO-W builds this happens every lwIP tick (64ms) which is unexpected.

The change here is to go back to using the same timer alarm for both lightsleep wakeup and soft timer, but now being explicit about lightsleep wakeup clobbering any soft timer wakeup. Also adds a "catch up" call to soft timer handler as it's not currently there.

This also reworks the changes added in 19844b4 to enable the timer IRQ on CPU1. The functionality is basically the same, just being a bit more explicit and by reworking this and the other code there's no more "magic IRQ number 3" in this code.

*Thanks to @cpottle9 for investigating this issue and pinpointing 74fb42a as the regression commit.* :pray:

### Testing

* [x] Basic test case on PICO-W.
* [x] Unit tests pass on PICO-W.
* [x] Test wakeup from pin IRQ as also reported in #16180 (might need to be fixed separately). (Note: Tested this works provided timeout is provided, issue without timeout is tracked in #7035)
* [ ] Test on RP2350 (help wanted, have ordered hardware but probably won't get until next year).
* [x] Add unit test for light sleep running from thread on core1.
* [x] ~~Add unit test verifying that soft timer callbacks resume after waking from lightsleep.~~ EDIT: Can't do this actually, as `machine.Timer()` uses the alarm pool. However, lightsleep and Wi-Fi are compatible so soft timer is resuming OK.
* [x] Test Wi-Fi stays valid after lightsleep (with a timeout).  

* We were hoping to go back to the pico-sdk alarm pool rather than using our own hardware timer alarm now that https://github.com/raspberrypi/pico-sdk/issues/1812 is closed, however @dpgeorge [mentioned here](https://github.com/micropython/micropython/pull/16412#discussion_r1886101759) that the new version of the alarm pool may have another bug - maybe the one fixed in https://github.com/raspberrypi/pico-sdk/pull/2127 ?
* In general, it might be expected that soft timers added by Python code can wake the rp2 from lightsleep. In that case, we could fix this by disabling the 64ms lwIP tick wakeup timer before going to lightsleep. Lightsleep wakeup could then be just another soft timer entry.
* More extreme analysis would be that lwIP tick should wake from lightsleep so that TCP connections can stay up, etc. I'm not sure I buy this one as the rp2 port is already tickless so "idle" is close to this behaviour, but you could make the case. If doing that instead then we should at least disable the lwIP tick while no network interface is up, though.

### DIFF
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -196,17 +196,8 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
         #endif
         xosc_dormant();
     } else {
-        bool timer3_enabled = irq_is_enabled(3);
-
-        const uint32_t alarm_num = 3;
-        const uint32_t irq_num = TIMER_ALARM_IRQ_NUM(timer_hw, alarm_num);
+        uint32_t timer_irq_num;
         if (use_timer_alarm) {
-            // Make sure ALARM3/IRQ3 is enabled on _this_ core
-            if (!timer3_enabled) {
-                irq_set_enabled(irq_num, true);
-            }
-            hw_set_bits(&timer_hw->inte, 1u << alarm_num);
-            // Use timer alarm to wake.
             clocks_hw->sleep_en0 = 0x0;
             #if PICO_RP2040
             clocks_hw->sleep_en1 = CLOCKS_SLEEP_EN1_CLK_SYS_TIMER_BITS;
@@ -215,8 +206,18 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
             #else
             #error Unknown processor
             #endif
-            timer_hw->intr = 1u << alarm_num; // clear any IRQ
-            timer_hw->alarm[alarm_num] = timer_hw->timerawl + delay_ms * 1000;
+
+            // Clobber the soft timer hardware alarm (as soft timer doesn't wake lightsleep)
+            // and repurpose it for the lightsleep wakeup
+            //
+            // The timer peripheral will already be configured via softtimer_init().
+            if (get_core_num() == 1) {
+                // On CPU0 the hardware alarm interrupt is already enabled, but IRQs won't be serviced
+                // due to the critical section so we also need to temporarily enable this IRQ on our CPU
+                timer_irq_num = hardware_alarm_get_irq_num(PICO_DEFAULT_TIMER_INSTANCE(), MICROPY_HW_SOFT_TIMER_ALARM_NUM);
+                irq_set_enabled(timer_irq_num, true);
+            }
+            hardware_alarm_set_target(MICROPY_HW_SOFT_TIMER_ALARM_NUM, delayed_by_ms(get_absolute_time(), delay_ms));
         } else {
             // TODO: Use RTC alarm to wake.
             clocks_hw->sleep_en0 = 0x0;
@@ -247,16 +248,24 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
 
         // Go into low-power mode.
         __wfi();
-
-        if (!timer3_enabled) {
-            irq_set_enabled(irq_num, false);
-        }
         clocks_hw->sleep_en0 |= ~(0u);
         clocks_hw->sleep_en1 |= ~(0u);
+
+        if (use_timer_alarm && get_core_num() == 1) {
+            // We don't have an easy way to clear any pending IRQ on this CPU,
+            // so we might get a spurious soft timer interrupt handler run on
+            // CPU1 unfortunately - but this should be harmless
+            irq_set_enabled(timer_irq_num, false);
+        }
     }
 
     // Enable ROSC.
     rosc_hw->ctrl = ROSC_CTRL_ENABLE_VALUE_ENABLE << ROSC_CTRL_ENABLE_LSB;
+
+    // we might have missed at least one soft timer expiry that hasn't been processed
+    // depending on when we woke up, so force soft timer processing for when we exit
+    // the atomic section
+    hardware_alarm_force_irq(MICROPY_HW_SOFT_TIMER_ALARM_NUM);
 
     // Bring back all clocks.
     runtime_init_clocks_optional_usb(disable_usb);
--- a/ports/rp2/mphalport.c
+++ b/ports/rp2/mphalport.c
@@ -250,9 +250,13 @@ static void soft_timer_hardware_callback(unsigned int alarm_num) {
     // a second ISR, as PendSV may be currently suspended by the other CPU.
     pendsv_schedule_dispatch(PENDSV_DISPATCH_SOFT_TIMER, soft_timer_handler);
 
-    // This ISR only runs on core0, but if core1 is running Python code then it
-    // may be blocked in WFE so wake it up as well. Unfortunately this also sets
-    // the event flag on core0, so a subsequent WFE on this core will not suspend
+    // This ISR mostly only runs on core0 but may run on core1 if core1 calls
+    // machine.lightsleep().
+    //
+    // Either way, if the other core is running Python code then it may be
+    // blocked in WFE so wake it up as well. Unfortunately this also sets the
+    // event flag on this core, so a subsequent WFE on this core will not
+    // suspend
     #if MICROPY_PY_THREAD
     if (core1_entry != NULL) {
         __sev();
--- a/tests/ports/rp2/rp2_lightsleep.py
+++ b/tests/ports/rp2/rp2_lightsleep.py
@@ -9,7 +9,7 @@
 # A range of sleep periods (1 to 512ms) are tested. The total nominal sleep time
 # is 10.23 seconds, but on most ports this will finish much earlier as interrupts
 # happen before each timeout expires.
-import sys
+import sys, time
 
 try:
     from machine import lightsleep, Pin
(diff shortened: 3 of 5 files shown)


cpottle9:
@projectgus, I built micropython for RPI_PICO_W with your changes and tested them.
Looks great.

As part of that, I ran the attached example running a periodic Timer once per second during the lightsleep.
[lightsleep_with_timer.txt](https://github.com/user-attachments/files/18170836/lightsleep_with_timer.txt)

I just happened to have this code lying around for testing sleep_ms and I thought to give it a try.
This specific case worked correctly.
 lightsleep() returned after 1 second as expected and the Timer function ran.
 
 I know interrupts from other sources (EG: Pin) don't work but this one does.

projectgus:
Thanks for taking a look and testing this, @cpottle9!

> I just happened to have this code lying around for testing sleep_ms and I thought to give it a try. This specific case worked correctly. lightsleep() returned after 1 second as expected and the Timer function ran.

Ironically, `machine.Timer` can now wake the rp2 from lightsleep because it uses the alarm pool (timer interrupt 3). So I assume it doesn't work on either V1.23 or V1.24, and may stop working on future versions if we switch back to also using the alarm pool for the soft timer!

peterharperuk:
Having a constant lwIP tick counter of 64ms is probably a waste when it will tell us when it next needs to wake up via sys_timeouts_sleeptime.

cpottle9:
@projectgus and other maintainers of the rp2 port, I've been following changes related to lightsleep and the shoe just dropped for me. I did not see what you are trying to do until now.

You are trying to solve _world hunger_ for lightsleep and I don't think that is good use of your time.
By that, I mean you are trying to provide maximum power reduction and yet not break functionality.
I see the functionality you are trying to preserve includes USB, WIFI, and use of the second core.

If I'm running a PICO_W from battery to save power the first thing I need to do is disable the cyw43 (and WIFI) when I don't need it. That includes not using the system LED. I have an application where I do wlan.deinit() and then lightsleep.
Further, when my application is deployed I don't need USB (I know, you fixed that one).
Running on battery I probably don't want to make use of the second core either.

I propose you need make lightsleep simpler and empower developers(like me) to decide which hardware blocks need to enabled for a given application.
To that end, I created micropython code that allows a developer to modify the WAKE_ENx and SLEEP_ENx registers on the RP2040 and RP2350. If you have time, take a look at my github [RP2-PowerControl](https://github.com/cpottle9/RP2-PowerControl).
My idea is users stop using machine.lightsleep and use time.sleep_ms instead ( make machined.lightsleep do the same thing as time.sleep_ms).
Whenever a __WFI or __WFE happens the contents of SLEEP_ENx registers will be used.
Further, if a user knows they do not need specific blocks the can update WAKE_ENx registers to power them down all the time.

projectgus:
> Having a constant lwIP tick counter of 64ms is probably a waste when it will tell us when it next needs to wake up via sys_timeouts_sleeptime.

Ah, thanks @peterharperuk! The rp2 port is the only tickless port in MicroPython at the moment, so it's the only place this really counts. If we can easily make lwIP tickless as well then we have a lot more options. I'd assumed this wasn't simple but sounds like it may well be.

> You are trying to solve world hunger for lightsleep and I don't think that is good use of your time.
> By that, I mean you are trying to provide maximum power reduction and yet not break functionality.
> I see the functionality you are trying to preserve includes USB, WIFI, and use of the second core.

This is a good point, @cpottle9, although it's less "not break functionality" and more "not crash or hang" which is the bottom line here. For example, the reason that USB gets special handling in rp2 lightsleep is because disabling (or suspending) on sleep and re-enabling on wake caused intermittent crashes when either the peripheral or TinyUSB gets in some permanent bad state. Similarly, the workarounds for the second core are to prevent execution stalling indefinitely if the other CPU gets permanently stuck in WFI. If the impact is limited to "feature stops on sleep and starts on wake" or "power consumption is higher when this feature is in use" then these are probably reasonable trade-offs. Permanent hangs or hard faults aren't reasonable trade-offs.

> I propose you need make lightsleep simpler and empower developers(like me) to decide which hardware blocks need to enabled for a given application.
> To that end, I created micropython code that allows a developer to modify the WAKE_ENx and SLEEP_ENx registers on the RP2040 and RP2350. If you have time, take a look at my github [RP2-PowerControl](https://github.com/cpottle9/RP2-PowerControl).

That's a great project, and I think it's a good approach for people who want the absolute most control over the sleep modes.

Part of the goal of MicroPython, and the `machine` module in particular, is to provide some abstraction of the hardware. To date, low-power modes are not the best example of this in MicroPython as there's not much of a common design and things that are quite different between each port. However, long term that's the kind of direction we'd want to go. I would say that as a first goal, having a "lowest reasonable consumption" `machine.lightsleep()` and a "low power but max functionality" `time.sleep()` will meet a lot of people's needs.

That's not to say that something like RP2-PowerControl isn't a great option for folks who need that fine-grained register-level control, though.

projectgus:
> If the impact is limited to "feature stops on sleep and starts on wake" or "power consumption is higher when this feature is in use" then these are probably reasonable trade-offs

I realise there is another trade-off we could choose for some of these states, which would be to have `lightsleep` degrade to a normal WFI or even raise a `RuntimeException` if we know it's not safe to disable clocks for lightsleep.

There are still footguns we don't want to leave lying around, though. For example, even if someone's battery powered application isn't designed to run with USB in the field then they might want to sometimes plug USB in and do some debugging. If the behaviour in that state is significantly different then it can become quite hard to debug.

cpottle9:
@projectgus, your USB issue is valid. 
But dealing with _can_ be the developers problem not yours. 

Their application can determine if USB is present and have RP2-PowerCtrl disable USB only when it is not.

By the way, I initially coded RP2-PowerCtrl as part of modrp2.c.
If the owners of micropython agree I could finish that work and create a pull request.

cpottle9:
@projectgus one other thing. In prior to commit d1423ef7a2  on entry lightsleep saved the current value of sleep_en0 and sleep_en1.
Before exiting it restored the saved values.

Now before exiting lightsleep sets those register to the power on defaults.

That means if someone uses RP2-PowerCtrl and then calls lightsleep their changes to sleep_en0 and sleep_en1 are lost.

Should I open an issue?

projectgus:
> By the way, I initially coded RP2-PowerCtrl as part of modrp2.c.
> If the owners of micropython agree I could finish that work and create a pull request.

That could be one way forward, although if everything works the same in pure Python code then it's less appealing to add binary size for it in modrp2. I also just tagged you in this discussion, where I vaguely outlined another possible approach: 
https://github.com/micropython/micropython/pull/16183#discussion_r1893472482

> Now before exiting lightsleep sets those registers to the power on defaults.
>
> That means if someone uses RP2-PowerCtrl and then calls lightsleep their changes to sleep_en0 and sleep_en1 are lost.
>
> Should I open an issue?

Yes, we can definitely change this back. If possible then suggest you open a Pull Request restoring behaviour that works for RP2-PowerCtrl, and then we can go from there.

projectgus:
Couple of things have become apparent about this approach:

1. It seems much more sensible to merge #16454 and go back to using the Pico alarm pool. That PR currently switches to use a dedicated timer alarm for lightsleep wakeup, which seems like a good choice but it will mean that the lwIP tick timer still wakes us up from lightsleep.
2. Switching to "tickless" lwIP doesn't immediately solve the problem either. LWIP runs all its "cyclic" timers all of the time, the fastest of which run at 100ms for IGMP and MLD6. MicroPython will need to either calculate when no interface is enabled and disable timer callbacks (which means lightsleep+wifi can work seamlessly), or we disable the lwIP timer callbacks while we're in lightsleep (in which case we may as well keep the 64ms lwIP tick).

projectgus:
Quick update after discussing with @dpgeorge : Strong preference for https://github.com/micropython/micropython/pull/16454 to be the long-term fix. Hopeful that there will be a pico-sdk release with the final alarm pool fix included and we can go direct to that one.

So I'm not planning to pick this approach up again for now. If this is a major inconvenience for anyone, please leave a comment.

cpottle9:
@projectgus, thanks for the update. You prompted me to look at #16454. 
It explains the symptoms I described in #16562.

If I understand correctly it will be a while before the issues in pico-sdk are addressed.

So, this problem will not get resolved quickly. Correct?

projectgus:
> If I understand correctly it will be a while before the issues in pico-sdk are addressed. So, this problem will not get resolved quickly. Correct?

The issue in pico-sdk is addressed now, but there hasn't yet been a pico-sdk release that includes the fix. Once a release comes out, and assuming no other issues crop up when we update pico-sdk, then we'll be able to proceed with the changes in #16454.

projectgus:
Closing in favour of the approach from https://github.com/micropython/micropython/pull/16454

PR #8744, opened 2022-06-09
mpremote: Support bytecode raw paste for 'mpremote run module.mpy'

projectgus:
For resource constrained devices without `mpremote mount` support, it is useful to be able to directly run precompiled bytecode (`mpremote run`) when developing, rather having to send source code each time.

Currently this is possible using pyboard.py - the implementation loads some loader Python and injects the bytecode as a variable. This approach reduces RAM usage further by using the "raw paste" window to paste into the same code path that loads and executes .mpy files (thanks Damien for this tip).

## Summary of changes

- Adds a new raw paste command 'B' for 'raw paste bytecode' (raw paste command 'A' is the current/default paste command)
- Adds an escaping mechanism (`<char>` is escaped as `Ctrl-F` `<char + 8>`) during raw paste so byte values less than 8 can be sent without triggering Ctrl-C or Ctrl-D handlers. The two-byte escape sequence still counts as one byte in the paste window.
- Adds relevant support to mpremote.py
- Adds the same "loader Python" approach to mpremote.py which is already in pyboard.py, for devices without raw paste support. Unclear how useful this is, I think it's only needed if using newer mpremote.py with older firmware, but there's no device-side changes and minimal host changes.

## Impact

Building for PYBV11 with default settings, .text segment +128 bytes.

## TODOs

- [x] ~~My understanding might be wrong, but this still currently streams the entire module into a memory buffer (mp_raw_code_load, mp_make_function_from_raw_code) and then executes it ( mp_call_function_0). Maybe this can be made to execute it in chunks as each Python statement completes, meaning less RAM usage for modules which execute some statements on import and don't need to keep those around. But I'm very new to this and haven't dug right into it, could be way off.~~ (I think the soft reset before run makes this a bit of a moot point, although there are still potential savings for 'run'-ing a bytecode that doesn't define any code.)
- [x] Set up some protocol macro defines in the code so all of the possible command sequences are grouped together, to make the protocol easier to understand from the top down.
- [x] ~~Test this works with `MICROPY_REPL_EVENT_DRIVEN`~~ (Seems only JavaScript port uses this method, at least by default, and it doesn't support mpremote - if I should test this on a different port, let me know.)
- [x] Test this works with 'Ctrl-K inject file'
- [x] Set up compilation guards for ports which don't support .mpy loading
- [x] Check for regressions due to the new escape sequences anywhere that pyboard.py calls raw paste routines
[...]

PR #16433, opened 2024-12-17
tests/run-tests.py: Set name of injected test module to '__main__'.

dpgeorge:
### Summary

Running unittest-based tests with `--via-mpy` is currently broken, because the unittest test needs the module to be named `__main__`, whereas it's actually called `__injected_test`.

Fix this by changing the name when the file is opened.

### Testing

The following now works:
```
$ ./run-tests.py -t a0 --via-mpy extmod/machine1.py
```

This fix is a little hacky.

An alternative would be to just make unittest-based tests do this at the end of the file:
```py
unittest.main(__name__)
```
(Instead of `if __name__ == '__main__': unittest.main()`).

### DIFF
--- a/tests/run-tests.py
+++ b/tests/run-tests.py
@@ -59,10 +59,13 @@ DIFF = os.getenv("MICROPY_DIFF", "diff -u")
 os.environ["PYTHONIOENCODING"] = "utf-8"
 
 # Code to allow a target MicroPython to import an .mpy from RAM
+# Note: the module is named `__injected_test` but it needs to have `__name__` set to
+# `__main__` so that the test sees itself as the main module, eg so unittest works.
 injected_import_hook_code = """\
 import sys, os, io, vfs
 class __File(io.IOBase):
   def __init__(self):
+    sys.modules['__injected_test'].__name__ = '__main__'
     self.off = 0
   def ioctl(self, request, arg):
     return 0


dpgeorge:
@projectgus @andrewleech @stinos do you have any preference for this fix?

The options are:
1. this PR
2. make all unittest-based tests end with `unittest.main(__name__)`

Maybe option (2) is cleaner, because then it doesn't matter what the module is imported as?

projectgus:
I prefer the way in this PR, it seems "principle of least surprise" that a Python file executed by the test runner is called `__main__` in all cases. We can also hopefully some day move the implementation into mpremote so you can run code via .mpy there, too? (This has been tried by a few folks previously, my own attempt was #8744).

> Maybe option (2) is cleaner, because then it doesn't matter what the module is imported as?

IIUC that way will run the tests even if you `import` the test module from a REPL to inspect the contents or manually instantiate the test class, wouldn't it?

dpgeorge:
> I prefer the way in this PR, it seems "principle of least surprise" that a Python file executed by the test runner is called `__main__` in all cases.

Yes and no.  I would think that it doesn't really matter what module a test is imported as, it should just run.  Up until now that's been the case, tests just run no matter what.  It's only with unittest that things have changed, because of the bit at the bottom of the unittest:
```py
if __name__ == "__main__":
    unittest.main()
```
Maybe it was an oversight by me to require this code... it's not like these tests are ever going to be imported and used as a module, right?  You would always expect them to just run their test.  We aren't using auto-discovery here like `micropython -m unittest`.

> IIUC that way will run the tests even if you `import` the test module from a REPL to inspect the contents or manually instantiate the test class, wouldn't it?

Right, if you import the unittest-based test from the REPL then, currently, it won't execute.  But if we changed the last bit to read `unittest.main(__name__)` then it would run when imported.

stinos:
Is there anything which might rely on `__injected_test` ?

> Maybe it was an oversight by me to require this code... it's not like these tests are ever going to be imported and used as a module, right? You would always expect them to just run their test. We aren't using auto-discovery here like micropython -m unittest.

Even though I kinda dislike the `if __name__ == "__main__": unittest.main()`, construct it's a very standard way to write tests and I'm leaning towards sticking with that. Because at least for me it's not uncommon at all to run/debug a bunch of unit tests with CPython, so then it's better if they're fully compatible with it to cover all scenarios.

Which in this case would mean for example: editor plugins supporting unittest use discovery to then allow you to run/debug a single test in a file. Pretty sure that's only going to work if the file can be imported as a module.

andrewleech:
My first thought was why is it currently called `__injected_test` in the first place and not just `__main__`; TIL micropython already has a __main__ built-in. Side thought, `__main__.__dict__` could be an alternative location to place objects you don't want garbage collected, similar to a root pointer.

Back on topic, I think the way presented in this PR is best. The `__injected_test` "module name" is an arbitrary implementation detail, set and used in the same section of commented code that "renames" it to `__main__`. Sure it's a little hacky on the face of it but it's self contained and for a specific purpose to work around the fact that `__main__` is already taken. 

I definitely think it's worth keeping the `if __name__ == '__main__':` pattern. 

I do like how clean `unittest.main(__name__)` looks but it's a non-standard pattern that might confuse new users and be one extra thing that needs changing when trying to port existing cpython modules.  I think keeping things that be developers see as "standard" as possible will help get more people writing tests. 

Speaking of confusing people, the one negative side effect I foresee is if said unit test developer tries to use `import __main__` in their tests to get access to the "top test file" but hopefully that's not a common pattern, I can't say I've seen it myself.

dpgeorge:
> Is there anything which might rely on `__injected_test` ?

No, that's a private internal name used only by `run-tests.py`.

> My first thought was why is it currently called `__injected_test` in the first place and not just `__main__`

I did try changing `__injected_test` to `__main__` and it didn't work, because the built-in `__main__` overrides it.

----------

Thanks all for the quick and useful feedback here.  Let's go with what's in this PR.

projectgus (approved):
LGTM!

PR #16437, opened 2024-12-17
extmod/vfs: Guard mutating fs functions with MICROPY_VFS_WRITABLE.

dpgeorge:
### Summary

Enabled by default.  Useful for ports that need the VFS but don't have any writable filesystems.

This is broken out from #8381.  There it's used for esp8266-512k boards to reduce their size when enabling ROMFS (such boards don't have any writable filesystem).

### Testing

Tested as part of #8381, by disabling this option on esp8266-512k.


projectgus (approved):
LGTM! I see that ports building with this will arrive with #8381 :+1:

PR #16440, opened 2024-12-17
extmod/uctypes: fix addressof when returning large values

dpgeorge:
### Summary

This fixes `uctypes.addressof()` so it always returns positive values.

Prior to this fix it would return negative addresses for values that were larger than the maximum small-int value.

### Testing

Test added as part of this PR which runs under CI.  Also tested on PYBD_SF6 (test failed prior to the fix here, now passes).


extmod/moductypes.c
```diff
@@ -602,7 +602,7 @@ static mp_obj_t uctypes_struct_unary_op(mp_unary_op_t op, mp_obj_t self_in) {
                 uint agg_type = GET_TYPE(offset, AGG_TYPE_BITS);
                 if (agg_type == PTR) {
                     byte *p = *(void **)self->addr;
-                    return mp_obj_new_int((mp_int_t)(uintptr_t)p);
+                    return mp_obj_new_int_from_uint((mp_int_t)(uintptr_t)p);
```
projectgus, line 605:
```suggestion
                    return mp_obj_new_int_from_uint((uintptr_t)p);
```

(Takes out the unnecessary and maybe-UB-probably-not cast.)

I ran into this testing #16311 V2 changes on qemu where the RAM range is at the top of the 32-bit space, I'm guessing you probably hit the same. :grin:

dlech, line 605:
```suggestion
                    return mp_obj_new_int_from_uint((uintptr_t)p);
```

Probably don't need the 2nd cast since the parameter type is `mp_uint_t` already.

dlech: Heh, I'm too slow, Angus beat me to it.

projectgus: Two out of two reviewers agree! :grin:

dpgeorge:
This was split out from #8381.

dpgeorge:
Thanks for the quick review!  I made the recommended change.  It should be fine, because `mp_uint_t` is always guaranteed to be at least as big as a system pointer (so it can be used to store `mp_obj_t` which can point to arbitrary memory addresses).

extmod/moductypes.c
dpgeorge: Yes I hit this on qemu as well, testing #8381, which checks the address of ROMFS files are within the address of the ROMFS itself.

projectgus (approved):

PR #16442, opened 2024-12-18
WIP: rp2: Enable GPIO wakeup from indefinite lightsleep, document sleep modes.

### LINKED ISSUE #7035: rp2: Can't wake from lightsleep with Pin IRQ

### LINKED ISSUE #16180: lightsleep() not working properly Pi Pico WH v1.24.0

projectgus:
### Summary

Make the `Pin.irq()` wakeup behaviour more consistent on rp2. Closes https://github.com/micropython/micropython/issues/7035 and Closes https://github.com/micropython/micropython/issues/16180 (the last part of that issue to need fixing.)

* Make it so any pin which has a `Pin.irq()` trigger set will wake rp2 port from any `machine.lightsleep()` call. (Previously this only worked if lightsleep had a timeout *and* the IRQ had a handler set.)
* Don't use the lowest power DORMANT mode if USB is active. Similar to #15301 it seems like disabling the USB clock while a device is enumerated leads to intermittent failures on wakeup. If USB is active then rp2 will now enter the higher level sleep mode only, the same one which is used if a timeout is set (in addition, the board is usually woken almost immediately by the USB host so this configuration doesn't save much power.)
* Document the rp2 sleep and low power modes and their capabilities/limitations.
* Add some notes in the machine documentation about sleep mode differences between ports.

### TODOs

* [ ] Waking from lightsleep appears to clobber the UART settings
* [ ] Full testing

### Testing

* TBD

* Instead of not allowing DORMANT mode when USB is active, we could first disable the USB device entirely (so the host sees it disconnect) and then go to DORMANT mode. Unclear if there's any use case that this would be useful for, though...?

### DIFF
--- a/docs/library/machine.Pin.rst
+++ b/docs/library/machine.Pin.rst
@@ -196,6 +196,8 @@ Methods
        These values can also be OR'ed together to make a pin generate interrupts in
        more than one power mode.
 
+       .. note:: The ``wake`` parameter is only available on the cc3200 port.
+
      - ``hard`` if true a hardware interrupt is used. This reduces the delay
        between the pin change and the handler being called. Hard interrupt
        handlers may not allocate memory; see :ref:`isr_rules`.
--- a/docs/library/machine.rst
+++ b/docs/library/machine.rst
@@ -145,8 +145,9 @@ Power related functions
    that require processing.  Such events, or wake sources, should be configured before
    sleeping, like `Pin` change or `RTC` timeout.
 
-   The precise behaviour and power-saving capabilities of lightsleep and deepsleep is
-   highly dependent on the underlying hardware, but the general properties are:
+   The precise behaviour and power-saving capabilities of lightsleep and
+   deepsleep, as well as the available wake-up sources, are all highly dependent
+   on the underlying hardware. The general properties are:
 
    * A lightsleep has full RAM and state retention.  Upon wake execution is resumed
      from the point where the sleep was requested, with all subsystems operational.
@@ -156,6 +157,8 @@ Power related functions
      script, similar to a hard or power-on reset. The `reset_cause()` function will
      return `machine.DEEPSLEEP` and this can be used to distinguish a deepsleep wake
      from other resets.
+
+   For more details consult the quick reference for the port you are using.
 
 .. function:: wake_reason()
 
--- a/docs/rp2/quickref.rst
+++ b/docs/rp2/quickref.rst
@@ -119,6 +119,28 @@ Use the :mod:`time <time>` module::
     start = time.ticks_ms() # get millisecond counter
     delta = time.ticks_diff(time.ticks_ms(), start) # compute time difference
 
+Sleep Modes and Power Consumption
+---------------------------------
+
+On the rp2 port:
+
+* `time.sleep()` and `time.sleep_ms()` will suspend the CPU and reduce power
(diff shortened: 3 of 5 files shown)


Gadgetoid:
Does Pick SDK 2.3.0's `pico_low_power` have any bearing here?

Right now we're shouldering a near duplicate of `runtime_init_clocks` and it feels like there ought to be a way to... not do that 😆

projectgus:
> Does Pick SDK 2.3.0's `pico_low_power` have any bearing here?

I hadn't seen this API, thanks @Gadgetoid! This does look like the way forward.
 
> Right now we're shouldering a near duplicate of `runtime_init_clocks` and it feels like there ought to be a way to... not do that 😆

Yes. There's a ticket open on pico-sdk for this (https://github.com/raspberrypi/pico-sdk/issues/1746) - but it looks like pico_low_power might be the solution for it, already shipping. :tada:

PR #16444, opened 2024-12-18
stm32: Fix extraction of hse/hsi/pllm values from preprocessed source.

dpgeorge:
### Summary

Follow up to #15646.

The expressions for the `micropy_hw_hse_value` etc variables may contain parenthesis, eg `micropy_hw_hse_value = ((25) * 1000000)`.  To handle such a case, simplify the regex and always use `eval(found)` to evaluate the expression.

### Testing

The `WEACT_F411_BLACKPILL` board now builds.  Tested all 6 variants, they all build.

### DIFF
--- a/ports/stm32/boards/plli2svalues.py
+++ b/ports/stm32/boards/plli2svalues.py
@@ -132,12 +132,9 @@ def search_header(filename, re_define, lookup):
             m = regex_define.match(line)
             if m:
                 # found lookup value
-                found = m.group(3)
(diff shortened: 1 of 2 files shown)


dpgeorge:
@andrewleech please see what you think about this fix.

andrewleech:
Thanks this looks perfect, it's a sensible addition to the regex regardless.

andrewleech (approved):

PR #16445, opened 2024-12-18
ports/rp2: Enable PPP for Pico W.

Gadgetoid:
Enables `MICROPY_PY_NETWORK_PPP_LWIP` for Pico W so that it can take advantage of PPP-supporting networking modules.

This would probably be too onerous for a regular Pico build. Even though it could work, I suspect the networking stack will blow the tight firmware allocation.

Edit: Woof! That's quite a bit more code bloat than I'd expected!


ports/rp2/boards/RPI_PICO_W/mpconfigboard.h
```diff
@@ -8,6 +8,11 @@
 #define MICROPY_PY_NETWORK 1
 #define MICROPY_PY_NETWORK_HOSTNAME_DEFAULT     "PicoW"
 
+// Enable PPP networking.
+#define MICROPY_PY_NETWORK_PPP_LWIP             (1)
+extern const struct _mp_obj_type_t mp_network_ppp_lwip_type;
+#define MICROPY_BOARD_NETWORK_INTERFACES        { MP_ROM_QSTR(MP_QSTR_PPP), MP_ROM_PTR(&mp_network_ppp_lwip_type) },
```
dpgeorge, line 14: You don't need to add this entry, it's done automatically in `extmod/modnetwork.c`.

ports/rp2/boards/RPI_PICO_W/mpconfigboard.h
```diff
@@ -8,6 +8,11 @@
 #define MICROPY_PY_NETWORK 1
 #define MICROPY_PY_NETWORK_HOSTNAME_DEFAULT     "PicoW"
 
+// Enable PPP networking.
+#define MICROPY_PY_NETWORK_PPP_LWIP             (1)
```
dpgeorge, line 12: Could probably put this line in `mpconfigport.h` as:
```c
define MICROPY_PY_NETWORK_PPP_LWIP (MICROPY_PY_NETWORK)
```
Then it's available on all network-enabled boards, eg Pico 2 W.

Gadgetoid:
Update to enable in `mpconfigport.h` by default, with the option to set `MICROPY_PY_NETWORK_PPP_LWIP` and disable it.

Looks like there's a significant code size overhead, though!


Edit: Looks like I need to limit to ports with LWIP enabled 🤦 

```
[  2%] Generating genhdr/pins.h, pins_ARDUINO_NANO_RP2040_CONNECT.c
[  2%] Generating genhdr/qstr.i.last
/home/runner/work/micropython/micropython/micropython repo/extmod/network_ppp_lwip.c:38:10: fatal error: lwip/dns.h: No such file or directory
   38 | #include "lwip/dns.h"
      |          ^~~~~~~~~~~~
```

Actually or maybe I have an order of operations thing going on in the header.

Edit: I think this is the classic disagreement in patterns between:

1. Include *all* source files and use .h defines to determine whether their contents are enabled
2. Conditionally include things based on CMake options and *then* set a define, eg: https://github.com/micropython/micropython/blob/744270ac1b9ed3929cd41d1a6e1f6ea0e785745d/ports/rp2/CMakeLists.txt#L318-L327

Gadgetoid:
Wait a minute, what's the diference between:

`MICROPY_PY_LWIP_PPP` - https://github.com/search?q=repo%3Amicropython%2Fmicropython%20MICROPY_PY_LWIP_PPP&type=code

and

`MICROPY_PY_NETWORK_PPP_LWIP ` - https://github.com/search?q=repo%3Amicropython%2Fmicropython+MICROPY_PY_NETWORK_PPP_LWIP&type=code

The latter seems to be set *just* for `lwipopts_common.h` - https://github.com/micropython/micropython/blob/744270ac1b9ed3929cd41d1a6e1f6ea0e785745d/extmod/lwip-include/lwipopts_common.h#L63

dpgeorge:
@Gadgetoid this PR looks good now, but you still have it in draft status.  Any reason it's in draft?

If you move it out of draft and rebase, it should be good to merge.

dpgeorge:
@Gadgetoid see above.

Gadgetoid:
> Any reason it's in draft?

PR anxiety! And a dash of completely forgetting I raised this.

Rebased and converted.

ports/rp2/mpconfigport.h
```diff
@@ -199,6 +199,7 @@
 #define MICROPY_VFS_ROM                         (MICROPY_HW_ROMFS_BYTES > 0)
 #define MICROPY_SSL_MBEDTLS                     (1)
 #define MICROPY_PY_LWIP_PPP                     (MICROPY_PY_NETWORK_PPP_LWIP)
+#define MICROPY_PY_LWIP_SOCK_RAW                (MICROPY_PY_LWIP)
```
dpgeorge: This is no longer needed, see 7c1f1c4dbbd0ac03610947aaa73be59f3546c6df

Instead, you want to remove the previous line `MICROPY_PY_LWIP_PPP`.

ports/stm32/mpconfigport.h
```diff
@@ -110,6 +110,7 @@
 #define MICROPY_PY_TIME_TIME_TIME_NS (1)
 #define MICROPY_PY_TIME_INCLUDEFILE "ports/stm32/modtime.c"
 #define MICROPY_PY_LWIP_PPP         (MICROPY_PY_NETWORK_PPP_LWIP)
+#define MICROPY_PY_LWIP_SOCK_RAW    (MICROPY_PY_LWIP)
```
dpgeorge: Same comment as above.

dpgeorge:
> PR anxiety! And a dash of completely forgetting I raised this.

:smile: 

I think some wires got crossed during the rebase, see comments above.

Gadgetoid:
Okay these commits being conceptually backwards really hurt my brain, I've swapped them and hopefully fixed them too.

ports/rp2/mpconfigport.h
```diff
@@ -241,8 +240,7 @@
 #endif
 
 #ifndef MICROPY_PY_NETWORK_PPP_LWIP
-#define MICROPY_PY_NETWORK_PPP_LWIP     (0)
-#endif
+#define MICROPY_PY_NETWORK_PPP_LWIP     (1)
```
dpgeorge: missing `#endif`

Gadgetoid: I'm clearly having a bad day. 🤦

Gadgetoid: Right, third time's a charm, the cleanup commit hid that endif removal so I stared at the enable commit for a solid minute scratching my head before the penny dropped.

Sorry. I've been burning the midnight oil trying to land those GC changes, thanks for bearing with me.

Gadgetoid:
*thousand yard stare*

:sob:

edit: Okay I had confused this in my head with *enable LWIP PPP for all Pico boards*, since we have PPP-based modules that could conceivably run on a Pico, Pico 2 and CYW43 support is tangental to that. As such I changed the mpconfigport.h from a conditional enable to always enable 🤦 Fixed!

Of course this built locally for me because I didn't try a non-W build.

Th... fo.. fifth time lucky?

ports/rp2/mpconfigport.h
dpgeorge: Probably needs to be `MICROPY_PY_LWIP` instead of 1.

(The GC improvements look really good!)

dpgeorge (approved):
It's green!

PR #16452, opened 2024-12-19
ports/stm32: Reserve SPI flash bus.

iabdalkader:
### Summary

Reserved SPI buses must remain initialized during a soft reboot as they may be used for SPI flash storage or XIP. Fixes an issue introduced in #16346.

### DIFF
--- a/ports/stm32/boards/ADAFRUIT_F405_EXPRESS/mpconfigboard.h
+++ b/ports/stm32/boards/ADAFRUIT_F405_EXPRESS/mpconfigboard.h
@@ -28,6 +28,7 @@
 // External SPI Flash config
 #if !MICROPY_HW_ENABLE_INTERNAL_FLASH_STORAGE
 
+#define MICROPY_HW_SPI_IS_RESERVED(id)  (id == 1) // Reserve SPI flash bus.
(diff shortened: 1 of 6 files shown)


dpgeorge:
I've tested this on LEGO_HUB_NO6, and it fixes the problem (soft reset hangs).

But I don't think the approach will work.  The `MIKROE_QUAIL` board uses SPI3 for two of its click connectors as well as the external SPI flash.  So this PR will be a big breaking change for that board, users will no longer be able to use SPI on two of the click connectors.

Instead I suggest a `MICROPY_HW_SPI1_IS_STATIC(spi_id)` macro which a board can use to say that certain SPI buses should not be deinitialized.

iabdalkader:
> The `MIKROE_QUAIL` board uses SPI3 for two of its click connectors as well as the external SPI flash.
>  So this PR will be a big breaking change for that board, users will no longer be able to use SPI on two of the click connectors.

I could just add the static macro and use it for this particular board, but I don't think this SPI should be shared. If allow it to be used from machine_spi users could reconfigure it changing the polarity/phase or baud would would break the SPI flash.

dpgeorge (approved):
Looks good now, thank you.

I tested again on LEGO_HUB_NO6 and it still works.

PR #16459, opened 2024-12-20
shared/tinyusb, extmod: Add USB NCM network driver.

andrewleech:
> Depends on #19101 (network init) and #19102 (TinyUSB Makefile consolidation). The first 5 commits in this branch are from those PRs and will drop out once they merge.

### Summary

This PR provides a network interface over the USB connection which is connected to the existing LWIP network stack.

`dmesg` after plugging in a Pico with NCM enabled:
image

The regular USB serial port is still provided for mpremote. Use that to enable the network interface:
image

DHCP server is provided by the board, so the host gets an IP address automatically:
image

After that, sockets and services can be used as per any normal network connection:

image
image

Any port based on TinyUSB can use this with minimal effort, provided LWIP is also enabled. The board/chip needs to be large enough to support the ~20KB RAM needs of LWIP.

The feature is disabled by default. Enable it by setting `MICROPY_PY_NETWORK_NCM=1` at build time.

#### Building

On stm32 (also requires enabling the TinyUSB stack and LWIP):
```bash
make -j -C ports/stm32 BOARD=NUCLEO_WB55 \
  MICROPY_PY_LWIP=1 \
  CFLAGS_EXTRA="-DMICROPY_HW_TINYUSB_STACK=1 -DMICROPY_PY_NETWORK_NCM=1"
```

On a cmake-based port (rp2):
```bash
make -j -C ports/rp2 BOARD=RPI_PICO2_W \
  MICROPY_PY_LWIP=1 \
  CFLAGS_EXTRA="-DMICROPY_PY_NETWORK_NCM=1"
```

On mimxrt:
```bash
make -j -C ports/mimxrt BOARD=SEEED_ARCH_MIX \
  MICROPY_PY_LWIP=1 \
  CFLAGS_EXTRA="-DMICROPY_PY_NETWORK_NCM=1 -DLWIP_IPV6=1"
```

### Testing
- Pico (minimal connectivity testing)
- STM32F765 running microdot (with stm32 TinyUSB PR merged)
- NUCLEO_WB55 (stm32, CDC+NCM, ping verified)
- mimxrt/SEEED_ARCH_MIX (NCM device appears but both CDC and NCM were locked up - needs further investigation)

LWIP adds ~20KB RAM overhead, so for smaller parts a custom USB protocol may be more appropriate. A network connection over USB opens up use of existing servers and services (HTTP, MQTT, WebREPL, etc.) without requiring Ethernet hardware.

### Generative AI

I used generative AI tools when creating this PR, but a human has checked the
code and is responsible for the code and the description above.

---
This work was funded by Planet Innovation.

### DIFF
--- a/docs/library/network.USBD_NCM.rst
+++ b/docs/library/network.USBD_NCM.rst
@@ -1,0 +1,82 @@
+.. currentmodule:: network
+.. _network.USBD_NCM:
+
+class USBD_NCM -- USB NCM network interface
+===========================================
+
+This class provides a network interface over USB using the NCM (Network
+Control Model) protocol.  The host computer sees this device as a USB
+Ethernet adapter and assigns it an IP address via DHCP (served by the
+MicroPython device).
+
+.. note:: ``network.USBD_NCM`` requires a port with TinyUSB and NCM support,
+          enabled at build time by defining ``MICROPY_PY_NETWORK_USBD_NCM``
+          (off by default).
+
+Example usage::
+
+    import network
+
+    nic = network.USBD_NCM()
+    nic.active(True)
+    # wait for USB host to configure the NCM interface
+    while not nic.isconnected():
+        pass
+
+    print(nic.ipconfig("addr4"))
+
+Constructors
+------------
+
+.. class:: USBD_NCM()
+
+   Create and return a USBD_NCM object.  This initialises the NCM network
+   interface if it has not already been initialised.  Only one instance
+   exists (singleton).
+
+Methods
+-------
+
+.. method:: USBD_NCM.active([is_active])
+
+   Activate or deactivate the network interface.  Without argument returns
+   current state as a bool.
+
+   The interface is brought up automatically before USB enumeration, so this
+   returns ``True`` from boot.
+
+.. method:: USBD_NCM.isconnected()
+
+   Returns ``True`` if the USB host has configured the NCM interface,
+   ``False`` otherwise.
+
+   When USB is disconnected, this returns ``False`` and network traffic
+   stops. The interface remains registered with lwIP and can resume when
+   the host reconnects and re-enumerates the device.
+
+.. method:: USBD_NCM.status()
+
+   Returns the link status as an integer: ``1`` if the interface is up,
+   ``0`` otherwise.
+
+.. method:: USBD_NCM.ipconfig('param')
+            USBD_NCM.ipconfig(param=value, ...)
+
+   See `AbstractNIC.ipconfig`.
+
+.. method:: USBD_NCM.ifconfig([(ip, subnet, gateway, dns)])
+
+   See `AbstractNIC.ifconfig`.
+
+Notes
+-----
+
+**Link-local IP address:** The device IP (169.254.x.1) is derived
+deterministically from the device MAC address. RFC 3927 ARP probe/announce
+(conflict detection) is not implemented, so if two devices happen to derive
+the same address on the same network segment, the conflict will go undetected.
+
+**MAC address uniqueness:** The device and host-side MAC addresses are derived
+from the value returned by ``mp_hal_get_mac()``. If two boards have the same
+hardware MAC (e.g. the port does not use a hardware UID), they will present
+the same network addresses and cause ARP conflicts.
--- a/docs/library/network.USBD_NCM.rst
+++ b/docs/library/network.USBD_NCM.rst
@@ -1,0 +1,81 @@
+.. currentmodule:: network
+.. _network.USBD_NCM:
+
+class USBD_NCM -- USB NCM network interface
+===========================================
+
+This class provides a network interface over USB using the NCM (Network
+Control Model) protocol.  The host computer sees this device as a USB
+Ethernet adapter and assigns it an IP address via DHCP (served by the
+MicroPython device).
+
+.. note:: ``network.USBD_NCM`` requires a port with TinyUSB and NCM support,
+          enabled at build time by defining ``MICROPY_PY_NETWORK_USBD_NCM``
+          (off by default).  The USB NCM class must also be active - either
+          compiled as the default or enabled in ``boot.py`` via
+          ``machine.USBDevice.BUILTIN_NCM``.
+
+Example usage::
+
+    import network
+
+    nic = network.USBD_NCM()
+    nic.active(True)
+    # wait for USB host to configure the NCM interface
+    while not nic.isconnected():
+        pass
+
+    print(nic.ipconfig("addr4"))
+
+Constructors
+------------
+
+.. class:: USBD_NCM()
+
+   Create and return a USBD_NCM object.  This initialises the NCM network
+   interface if it has not already been initialised.  Only one instance
+   exists (singleton).
+
+Methods
+-------
+
+.. method:: USBD_NCM.active([is_active])
+
+   Activate or deactivate the network interface.  Without argument returns
+   current state as a bool.
+
+.. method:: USBD_NCM.isconnected()
+
+   Returns ``True`` if the USB host has configured the NCM interface,
+   ``False`` otherwise.
+
+   When USB is disconnected, this returns ``False`` and network traffic
+   stops. The interface remains registered with lwIP and can resume when
+   the host reconnects and re-enumerates the device.
+
+.. method:: USBD_NCM.status()
+
+   Returns the link status as an integer: ``1`` if the interface is up,
+   ``0`` otherwise.
+
+.. method:: USBD_NCM.ipconfig('param')
+            USBD_NCM.ipconfig(param=value, ...)
+
+   See `AbstractNIC.ipconfig`.
+
+.. method:: USBD_NCM.ifconfig([(ip, subnet, gateway, dns)])
+
+   See `AbstractNIC.ifconfig`.
+
+Notes
+-----
+
+**Link-local IP address:** The device IP (169.254.x.1) is derived
+deterministically from the device MAC address. RFC 3927 ARP probe/announce
+(conflict detection) is not implemented, so if two devices happen to derive
+the same address on the same network segment, the conflict will go undetected.
+
+**MAC address uniqueness:** The device and host-side MAC addresses are derived
+from the value returned by ``mp_hal_get_mac()``. If two boards have the same
+hardware MAC (e.g. the port does not use a hardware UID), they will present
+the same network addresses and cause ARP conflicts.
--- a/extmod/modnetwork.c
+++ b/extmod/modnetwork.c
@@ -55,6 +55,11 @@ extern const struct _mp_obj_type_t mod_network_nic_type_nina;
 extern const struct _mp_obj_type_t mod_network_esp_hosted_type;
 #endif
 
+#if MICROPY_PY_NETWORK_USBD_NCM
+extern const struct _mp_obj_type_t mod_network_nic_type_ncm;
+#include "extmod/network_usbd_ncm.h"
+#endif
+
 #ifdef MICROPY_PY_NETWORK_INCLUDEFILE
 #include MICROPY_PY_NETWORK_INCLUDEFILE
 #endif
@@ -75,6 +80,11 @@ char mod_network_hostname_data[MICROPY_PY_NETWORK_HOSTNAME_MAX_LEN + 1] = MICROP
 
 void mod_network_init(void) {
     mp_obj_list_init(&MP_STATE_PORT(mod_network_nic_list), 0);
+    #if MICROPY_PY_NETWORK_USBD_NCM
+    // Initialise the USB NCM netif early so it is ready when TinyUSB starts
+    // receiving packets during USB enumeration (before any Python code runs).
+    ncm_auto_init();
+    #endif
 }
 
 void mod_network_deinit(void) {
@@ -203,6 +213,10 @@ static const mp_rom_map_elem_t mp_module_network_globals_table[] = {
 
     #if MICROPY_PY_NETWORK_ESP_HOSTED
     { MP_ROM_QSTR(MP_QSTR_WLAN), MP_ROM_PTR(&mod_network_esp_hosted_type) },
+    #endif
+
+    #if MICROPY_PY_NETWORK_USBD_NCM
+    { MP_ROM_QSTR(MP_QSTR_USBD_NCM), MP_ROM_PTR(&mod_network_nic_type_ncm) },
     #endif
 
     // Allow a port to take mostly full control of the network module.
--- a/extmod/network_usbd_ncm.c
+++ b/extmod/network_usbd_ncm.c
@@ -1,0 +1,400 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2026 Andrew Leech
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "py/runtime.h"
+#include "py/mperrno.h"
+#include "py/mphal.h"
+
+#if MICROPY_PY_NETWORK_USBD_NCM
+
+// Include NCM configuration (MICROPY_PY_NETWORK_USBD_NCM_DHCP_SERVER).
+#include "extmod/network_usbd_ncm.h"
+
+#ifndef NO_QSTR
+#include "tusb.h"
+#endif
+
+#include "lwip/ethip6.h"
+#include "lwip/init.h"
+#include "lwip/timeouts.h"
+#include "lwip/etharp.h"
+#include "lwip/dns.h"
+#include "lwip/dhcp.h"
+#include "netif/ethernet.h"
+#include "lwip/apps/mdns.h"
+
+#include "extmod/modnetwork.h"
+#include "shared/netutils/netutils.h"
+#include "shared/netutils/dhcpserver.h"
+
+#define error_printf(...) mp_printf(&mp_plat_print, "network_usbd_ncm: " __VA_ARGS__)
+
+#if LWIP_IPV6
+#define IP(x) ((x).u_addr.ip4)
+#else
+#define IP(x) (x)
+#endif
+
+const mp_obj_type_t mod_network_nic_type_ncm;
+
+typedef struct _ncm_obj_t {
+    mp_obj_base_t base;
+
+    struct netif netif;
+    bool init;
+
+    ip_addr_t ipaddr;
+    ip_addr_t netmask;
+    ip_addr_t gateway;
+
+    #if MICROPY_PY_NETWORK_USBD_NCM_DHCP_SERVER
+    dhcp_server_t dhcp_server;
+    #endif
+} ncm_obj_t;
+
+// Global object holding the usb net state
+static ncm_obj_t ncm_obj;
+
+static mp_sched_node_t ncm_sched_node;
+
+#define IS_ACTIVE(self) ((self)->netif.flags & NETIF_FLAG_UP)
+
+/* This default mac will be updated during init from hardware unique ID (if available) */
+/* it is suggested that the first byte is 0x02 to indicate a link-local address */
+uint8_t tud_network_mac_address[6] = {0x02, 0x00, 0x00, 0x00, 0x00, 0x00};
+
+// Derive a link-local IPv4 address (169.254.X.1) from a MAC address.
+// Note: no ARP probe/announce is performed (RFC 3927 collision detection is not
+// implemented). The address is deterministic; if two devices share the same MAC,
+// they will get the same link-local IP and ARP conflicts will go undetected.
+// Assumes mp_hal_get_mac() returns a per-device unique value.
+static void generate_linklocal_ip_from_mac(const uint8_t mac_addr[6], uint8_t ip_addr[4]) {
+    ip_addr[0] = 169;
+    ip_addr[1] = 254;
+    ip_addr[3] = 1;
+
+    // Reuse the core string hash (djb2, the same hash() uses on bytes) to
+    // spread the MAC across the third octet, with no extra code compiled in.
+    size_t hash = qstr_compute_hash(mac_addr, 6);
+    uint8_t third_octet = (uint8_t)(hash ^ (hash >> 8));
+
+    // Avoid .0 and .255
+    if (third_octet == 0) {
+        third_octet = 1;
+    }
+    if (third_octet == 255) {
+        third_octet = 254;
+    }
+
+    ip_addr[2] = third_octet;
+}
+
+static err_t linkoutput_fn(struct netif *netif, struct pbuf *p) {
+    (void)netif;
+    if (!tud_ready() || !tud_network_can_xmit(p->tot_len)) {
+        // Drop the frame while the link is down or a previous transmit is
+        // still in flight.  An error return would propagate out of
+        // tcp_output() to the caller of socket.send(), even though tcp_write()
+        // has already queued that data, so the caller would resend bytes that
+        // are already in the stream.  Dropping leaves recovery to TCP
+        // retransmission, as for any other lossy link.
+        return ERR_OK;
+    }
+    tud_network_xmit(p, 0);
+    return ERR_OK;
+}
+
+static err_t netif_init_cb(struct netif *netif) {
+    LWIP_ASSERT("netif != NULL", (netif != NULL));
+    netif->name[0] = 'u'; // lwIP short name "un" for USB Network
+    netif->name[1] = 'n';
+    netif->mtu = CFG_TUD_NET_MTU;
+    netif->flags = NETIF_FLAG_BROADCAST | NETIF_FLAG_ETHARP | NETIF_FLAG_ETHERNET | NETIF_FLAG_IGMP;
+    netif->state = NULL;
+    netif->linkoutput = linkoutput_fn;
+    netif->output = etharp_output;
+    #if LWIP_IPV6
+    netif->output_ip6 = ethip6_output;
+    netif->flags |= NETIF_FLAG_MLD6;
+    #endif
+    return ERR_OK;
+}
+
+static void ncm_init(void) {
+    // Init the NCM object
+    ncm_obj.base.type = (mp_obj_type_t *)&mod_network_nic_type_ncm;
+
+    struct netif *netif = &(ncm_obj.netif);
+
+    MICROPY_PY_LWIP_ENTER
+
+    mp_hal_get_mac(MP_HAL_MAC_ETH0, tud_network_mac_address);
+    tud_network_mac_address[0] = 0x02;
+
+    netif->hwaddr_len = sizeof(tud_network_mac_address);
+    memcpy(netif->hwaddr, tud_network_mac_address, sizeof(tud_network_mac_address));
+
+    // This USB Network essentially creates a link with two interfaces, one for the
+    // host and another for the micropython/lwip end of the connection.
+    // The MAC configured above is use for one end of the link, flip a bit for the
+    // MAC on the other end of the link.
+    netif->hwaddr[5] ^= 0x01;
+
+    // Generate unique link-local IP address from MAC address
+    uint8_t ip_bytes[4];
+    generate_linklocal_ip_from_mac(tud_network_mac_address, ip_bytes);
+
+    // Convert to network byte order and set IP configuration
+    IP(ncm_obj.ipaddr).addr = PP_HTONL(((uint32_t)ip_bytes[0] << 24) | ((uint32_t)ip_bytes[1] << 16) | ((uint32_t)ip_bytes[2] << 8) | ip_bytes[3]);
+    IP(ncm_obj.netmask).addr = PP_HTONL(0xFFFF0000);  // 255.255.0.0 for link-local
+    IP(ncm_obj.gateway).addr = 0;  // No gateway for link-local
+
+    netif = netif_add(
+        netif,
+        ip_2_ip4(&ncm_obj.ipaddr),
+        ip_2_ip4(&ncm_obj.netmask),
+        ip_2_ip4(&ncm_obj.gateway),
+        NULL,
+        netif_init_cb,
+        ethernet_input
+        );
+    if (netif == NULL) {
+        MICROPY_PY_LWIP_EXIT
+        mp_raise_OSError(MP_ENOMEM);
+    }
+    #if LWIP_IPV6
+    netif_create_ip6_linklocal_address(netif, 1);
+    netif_set_ip6_autoconfig_enabled(netif, 1);
+    #endif
+    #if LWIP_NETIF_HOSTNAME
+    netif_set_hostname(netif, mod_network_hostname_data);
+    #endif
+    if (netif_default == NULL) {
+        netif_set_default(netif);
+    }
+
+    #if LWIP_MDNS_RESPONDER
+    mdns_resp_add_netif(netif, mod_network_hostname_data);
+    #endif
+
+    #if MICROPY_PY_NETWORK_USBD_NCM_DHCP_SERVER
+    dhcp_server_init(&ncm_obj.dhcp_server, &ncm_obj.ipaddr, &ncm_obj.netmask);
+    // Don't advertise a default gateway -- this is a point-to-point USB link
+    // and advertising a router causes the host to route all traffic over USB.
+    ncm_obj.dhcp_server.send_router = false;
+    #endif
+
+    // register with network module
+    mod_network_register_nic(&ncm_obj);
+    ncm_obj.init = true;
+
+    MICROPY_PY_LWIP_EXIT
+}
+
+static void _scheduled_tud_network_recv_renew(mp_sched_node_t *node) {
+    (void)node;
+    // Release the TinyUSB receive buffer and pick up the next datagram.
+    tud_network_recv_renew();
+}
+
+bool tud_network_recv_cb(const uint8_t *src, uint16_t size) {
+    if (!ncm_obj.init) {
+        return false;
+    }
+    struct netif *netif = &(ncm_obj.netif);
+    bool ret = true;
+    if (size) {
+        MICROPY_PY_LWIP_ENTER
+
+        struct pbuf *p = pbuf_alloc(PBUF_RAW, size, PBUF_POOL);
+        if (p == NULL) {
+            MICROPY_PY_LWIP_EXIT
+            // Return false without calling recv_renew: TinyUSB holds the buffer.
+            // NCM RX will stall until the host retransmits and memory is available.
+            return false;
+        }
+
+        // Copy buf to pbuf
+        pbuf_take(p, src, size);
+
+        if (netif->input(p, netif) != ERR_OK) {
+            pbuf_free(p);
+            ret = false;
+        }
+
+        MICROPY_PY_LWIP_EXIT
+    }
+    // Defer tud_network_recv_renew(): calling it here would pull the next
+    // datagram, and its full lwIP input path, into this USB interrupt.
+    mp_sched_schedule_node(&ncm_sched_node, _scheduled_tud_network_recv_renew);
+    return ret;
+}
+
+uint16_t tud_network_xmit_cb(uint8_t *dst, void *ref, uint16_t arg) {
+    // copy from network stack packet pointer to dst.
+    (void)arg;
+    MICROPY_PY_LWIP_ENTER
+    struct pbuf *p = (struct pbuf *)ref;
+    uint16_t len = pbuf_copy_partial(p, dst, p->tot_len, 0);
+    MICROPY_PY_LWIP_EXIT
+    return len;
+}
+
+// Set the lwIP link state and tell the host to match, so that a deactivated
+// interface shows up as disconnected rather than silently discarding traffic.
+static void ncm_set_link(bool up) {
+    if (up) {
+        netif_set_link_up(&ncm_obj.netif);
+    } else {
+        netif_set_link_down(&ncm_obj.netif);
+    }
+    tud_network_link_state(TUD_OPT_RHPORT, up);
+}
+
+// Called during netd_init() for the initial link state, which TinyUSB otherwise
+// defaults to up.  netd_init() also runs on every bus reset, so without this the
+// host would see the link come back up after a replug while the netif is down.
+bool tud_network_default_link_state_cb(void) {
+    return !ncm_obj.init || netif_is_link_up(&ncm_obj.netif);
+}
+
+// Called from mod_network_init() to set up the NCM netif before USB starts.
+// TinyUSB's NCM driver begins receiving as soon as the host sends
+// SET_INTERFACE(alt=1), which happens during USB enumeration -- before any
+// Python code runs.  The netif must exist by then so tud_network_recv_cb()
+// can deliver packets to lwIP.
+//
+// The netif is intentionally kept alive across MicroPython soft reset: lwIP
+// is not re-initialised on soft reset, so the netif remains valid and TinyUSB
+// can continue delivering packets. On re-entry we only re-register the singleton
+// with the NIC list (which mod_network_init() cleared) rather than re-running
+// netif_add (which must only be called once per netif).
+void ncm_auto_init(void) {
+    if (!ncm_obj.init) {
+        ncm_init();
+    } else {
+        // On soft-reset mod_network_init() clears the NIC list, so
+        // re-register the (still-live) singleton.
+        mod_network_register_nic(&ncm_obj);
+    }
+    // Bring the interface up here rather than leaving it to active(True): lwIP
+    // drops inbound packets on a down netif, and the host starts DHCP as soon
+    // as it enumerates.  As such active() reads True from boot.
+    netif_set_up(&ncm_obj.netif);
+    ncm_set_link(true);
+}
+
+/*******************************************************************************/
+// MicroPython bindings
+
+// Create and return a USBD_NCM object.
+static mp_obj_t ncm_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 0, 0, false);
+    if (!ncm_obj.init) {
+        ncm_init();
+    }
+    // Return NCM object
+    return MP_OBJ_FROM_PTR(&ncm_obj);
+}
+
+static mp_obj_t ncm_status(mp_obj_t self_in) {
+    ncm_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    return MP_OBJ_NEW_SMALL_INT(IS_ACTIVE(self) ? 1 : 0);
+}
+static MP_DEFINE_CONST_FUN_OBJ_1(ncm_status_obj, ncm_status);
+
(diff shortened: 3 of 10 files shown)


Josverl:
Interesting, is the assumption that the MCU is network capable, and will act as a bridge/router?

andrewleech:
With this feature any MCU with TinyUSB based USB connection _becomes_ network capable.

I hadn't thought of routing or bridging to other interfaces, though that's probably possible. 

My initial use case at Planet Innovation is for larger systems where we have a Linux or windows host PC with an embedded module running micropython attached. We wanted an option to use standard networking protocols to communicate between the application on the host computer and the device, however not every system warrants an MCU with an Ethernet connection, or the host computer Ethernet is used for external connection and we want the internal connection on a separate interface etc. 

With this feature the connection between host and device is still point to point like any USB connection, but you can easily run multiple services like a http server, mqtt client, webrepl etc. 

Currently I don't think you can monitor or query for the IP address/s handed out by the micropython dhcp server to active clients, I might want to add that if I'm correct so that any client library could easily find / talk to services running on the host.

robert-hh:
> With this feature any MCU with TinyUSB based USB connection becomes network capable.

In that respect, who is server, who is client?
- Does a MP board without network interface get network capable as a client and uses a service provided by the entity it connects to? In that case, how is the service provided?  or
- Is the MP board a server providing network services to client connected by USB? In that case the board must have already a network interface.

andrewleech:
> who is server, who is client?

There's no hard and fast rule here, both can run servers and be clients. 

This is just a network connection. 

Both the micropython device and the PC have their own network stacks.

The USB network connection kind of acts the same as if the micropython board has an Ethernet port that is connected to a PC Ethernet port via a crossover cable. So both ends have their own IP address that can be manually or automatically assigned but there's no router or hub in between.

Currently the micropython end is hardcoded to 192.168.7.1 though I'd plan to make that configurable eventually.

I have enabled the built-in dhcp server on micropython on the interface (so mpy end is a server from that perspective) and the PC end seems to get provisioned as 192.168.7.16 by default, though this isn't necessarily guaranteed.

If the PC happens to be running a web server then a micropython requests library will be able to access it on that 192.168.7.16 address. 

Conversely if you run microdot or similar on the micropython board the PC will be able to access it on 192.168.7.1

I'm planning on often having a Linux host running mosquito mqtt broker which both the micropython board and the host PC can both publish/subscribe to. I've also looked briefly at a pure python mqtt broker that I might try running on the device instead though! 

My point is server/client relationship is generally on a per-service basis and can work both ways.

andrewleech:
This has been updated to make `mdns` work so the device can be pinged from host based on hostname, though it seems pretty slow to resolve initially on windows.

The ip addresses used by default are now in the [link-local range](https://en.wikipedia.org/wiki/Link-local_address) which feels appropriate for this kind of isolated point-to-point link and should reduce the chance of ip collisions with other networks on the host.

Once the host is provided with an ip from the dhcp server in micropython, the host ip address is set as the gateway address on the micropython side of the interface which makes it easy for the micropython application to find/use the host ip address.

I've written up a quick demo of host to use mqtt between the device and host pc here, it works really well! https://github.com/andrewleech/mpy-usb-net-mqtt-demo/

xplshn:
Could you please introduce an example to use forward WLAN to RNDIS using this? It'd be a welcome thing in the usb examples folder.

andrewleech:
On most ports, features are generally enabled like: 
```
make -j BOARD=RPI_PICO2_W CFLAGS_EXTRA="-DMICROPY_PY_USBNET=1"
```

That plug gap alignment error is a strange one, I actually got it yesterday on a RPI build on an unrelated branch. I think it's related to the compilation of picotool more than the firmware issue, don't know where it came.

I haven't used this change on a board with wifi, though it probably should work fine. 

Not sure how to proxy interfaces with lwip / micropython but I'm guessing there will probably be some standard patterns for this.

xplshn:
> On most ports, features are generally enabled like:
> 
> ```
> make -j BOARD=RPI_PICO2_W CFLAGS_EXTRA="-DMICROPY_HW_NETWORK_USBNET=1"
> ```
> 
> That plug gap alignment error is a strange one, I actually got it yesterday on a RPI build on an unrelated branch. I think it's related to the compilation of picotool more than the firmware issue, don't know where it came.
> 
> I haven't used this change on a board with wifi, though it probably should work fine.
> 
> Not sure how to proxy interfaces with lwip / micropython but I'm guessing there will probably be some standard patterns for this.

Thanks a lot, I'm new to Python and really stumbled upon this out of sheer luck!

andrewleech:
Thanks for your testing information, I'll have to look into making it easier to include on-demand with the make command. 

At the moment I'm not expecting it to be built in by default as it adds a reasonable amount of flash/ram use; I've got a build variant of the original Pico added as an example but that's not really helping with your board.

andrewleech:
Great to hear @xplshn 
I finally figured out how to pass defines on the command line during make, it's different to the other ports: 
```
make  -j BOARD=RPI_PICO2_W clean

CFLAGS="-DMICROPY_HW_NETWORK_USBNET=1" make  -j BOARD=RPI_PICO2_W
```

xplshn:
@andrewleech do you think an example of RNDIS usage could be provided? I want to finish my wifi2usb.py program, I'd appreciate help, perhaps this could become _the_ RNDIS usage example :)

```python
import network
import rp2
import time
import sys
import machine
from machine import Pin
import select
import socket

pico_led = Pin("LED", Pin.OUT)

SCAN_TIMEOUT = 5000

class WiFi2USB:
    def __init__(self):
        self.usb_net = network.USB_NET()
        self.usb_net.active(True)

        self.wlan = network.WLAN(network.STA_IF)
        self.wlan.active(True)

        self.networks = []
        self.current_ssid = ""
        self.current_password = ""
        self.connected = False

    def scan_networks(self):
        """Scan for available WiFi networks"""
        pico_led.on()
        print("Scanning for networks...")
        self.networks = self.wlan.scan()
        pico_led.off()
        return self.networks

    def connect_wifi(self, ssid, password):
        """Connect to a WiFi network with given SSID and password"""
        pico_led.off()
        print(f"Connecting to {ssid}...")

        self.wlan.connect(ssid, password)

        start_time = time.ticks_ms()
        while not self.wlan.isconnected():
            # Check for cancel with BOOTSEL button
            if rp2.bootsel_button() == 1:
                print("\nConnection attempt cancelled")
                return False

            if time.ticks_diff(time.ticks_ms(), start_time) > SCAN_TIMEOUT:
                print("\nConnection timeout - please try again")
                return False

            pico_led.on()
            time.sleep(0.1)
            pico_led.off()
            time.sleep(0.1)

        ip = self.wlan.ifconfig()[0]
        print(f'Connected on {ip}')
        pico_led.on()
        self.connected = True
        self.current_ssid = ssid
        self.current_password = password
        return True

    def disconnect_wifi(self):
        """Disconnect from WiFi network"""
        if self.connected:
            self.wlan.disconnect()
            self.connected = False
            pico_led.off()
            print("Disconnected from WiFi")

    def get_network_info(self):
        """Return current network configuration"""
        if self.connected:
            return {
                'ssid': self.current_ssid,
                'ip': self.wlan.ifconfig()[0],
                'subnet': self.wlan.ifconfig()[1],
                'gateway': self.wlan.ifconfig()[2],
                'dns': self.wlan.ifconfig()[3],
                'status': 'Connected'
            }
        else:
            return {'status': 'Not connected'}

    def test_connectivity(self):
        """Test network connectivity"""
        if not self.connected:
            print("Not connected to any network")
            return False

        # Simple connectivity test (could be expanded with ping, etc.)
        try:
            ip = self.wlan.ifconfig()[0]
            print(f"Connected with IP: {ip}")
            print("Gateway: " + self.wlan.ifconfig()[2])

            # TODO: ping a reliable site here
            print("Connectivity test passed")
            return True
        except:
            print("Connectivity test failed")
            return False

    def get_usb_status(self):
        """Return USB RNDIS status"""
        return {
            'active': self.usb_net.active(),
            'ifconfig': self.usb_net.ifconfig()
        }

    def forward_traffic(self):
        """Forward traffic between WiFi and USB interfaces"""
        try:
            wlan_sock = socket.socket(socket.AF_INET, socket.SOCK_RAW, socket.IPPROTO_IP)
            usb_sock = socket.socket(socket.AF_INET, socket.SOCK_RAW, socket.IPPROTO_IP)

            wlan_sock.bind((self.wlan.ifconfig()[0], 0))
            usb_sock.bind((self.usb_net.ifconfig()[0], 0))

            poller = select.poll()
            poller.register(wlan_sock, select.POLLIN)
            poller.register(usb_sock, select.POLLIN)

            print("Forwarding traffic between WiFi and USB interfaces...")

            while True:
                events = poller.poll(1000)
                for fd, event in events:
                    if fd == wlan_sock.fileno():
                        data, addr = wlan_sock.recvfrom(1500)
                        usb_sock.sendto(data, (self.usb_net.ifconfig()[0], 0))
                    elif fd == usb_sock.fileno():
                        data, addr = usb_sock.recvfrom(1500)
                        wlan_sock.sendto(data, (self.wlan.ifconfig()[0], 0))
        except OSError as e:
            print(f"Socket error: {e}")
        except Exception as e:
            print(f"Unexpected error: {e}")

class CLI:
    def __init__(self, wifi2usb):
        self.wifi2usb = wifi2usb
        self.current_menu = self.main_menu

    def start(self):
        """Start the CLI interface"""
        print("\n" + "=" * 50)
        print("WiFi2USB Dongle - CLI Interface")
        print("=" * 50)

        while True:
            try:
                self.current_menu()
            except KeyboardInterrupt:
                print("\nExiting CLI Interface")
                break

    def main_menu(self):
        """Display the main menu"""
        print("\nMAIN MENU:")
        print("1. Reconfigure network")
        print("2. See network config")
        print("3. Test network connectivity")
        print("4. Enter microPython REPL")
        print("0. Exit")

        choice = input("\nEnter your choice: ")

        if choice == "1":
            self.current_menu = self.network_selection_menu
        elif choice == "2":
            self.network_config_menu()
        elif choice == "3":
            self.connectivity_test_menu()
        elif choice == "4":
            self.enter_repl()
        elif choice == "0":
            print("Exiting...")
            sys.exit()
        else:
            print("Invalid choice, please try again")

    def network_selection_menu(self):
        """Display network selection menu"""
        print("\nScanning for networks...")
        networks = self.wifi2usb.scan_networks()

        print("\nAvailable Networks:")
        print("   Network Name                [#Channel] |Other info|")
        print("-" * 60)

        for i, network in enumerate(networks, 1):
            try:
                ssid = network[0].decode('utf-8')
                bssid = network[1]  # MAC address in bytes
                channel = network[2]
                rssi = network[3]
                authmode = network[4]

                # Format MAC address
                mac = ':'.join('{:02x}'.format(b) for b in bssid)

                # Safe access to security types with fallback
                security_types = ["Open", "WEP", "WPA-PSK", "WPA2-PSK", "WPA/WPA2-PSK", "WPA3"]
                if 0 <= authmode < len(security_types):
                    security = security_types[authmode]
                else:
                    security = f"Unknown({authmode})"

                print(f"{i}. {ssid:<25} [#{channel}]     |mac:{mac}, db:{rssi}, {security}|")
            except Exception as e:
                print(f"{i}. [Error reading network info: {e}]")

        print("-" * 60)
        print("0: go back to last menu")

        try:
            choice = input("\nSelect network: ")

            if choice == "0":
                self.current_menu = self.main_menu
                return

            idx = int(choice) - 1
            if 0 <= idx < len(networks):
                selected_ssid = networks[idx][0].decode('utf-8')
                password = input(f"Enter password for {selected_ssid}: ")

                if self.wifi2usb.connect_wifi(selected_ssid, password):
                    print(f"Successfully connected to {selected_ssid}")
                    self.wifi2usb.forward_traffic()
                else:
                    print(f"Failed to connect to {selected_ssid}")

                # Return to main menu after connection attempt
                self.current_menu = self.main_menu
            else:
                print("Invalid selection")
        except ValueError:
            print("Please enter a valid number")

    def network_config_menu(self):
        """Display current network configuration"""
        print("\nCurrent Network Configuration:")
        print("-" * 50)

        # WiFi status
        wifi_info = self.wifi2usb.get_network_info()
        print("WiFi Status:", wifi_info['status'])

        if wifi_info['status'] == 'Connected':
            print(f"SSID: {wifi_info['ssid']}")
            print(f"IP Address: {wifi_info['ip']}")
            print(f"Subnet Mask: {wifi_info['subnet']}")
            print(f"Gateway: {wifi_info['gateway']}")
            print(f"DNS Server: {wifi_info['dns']}")

        # USB RNDIS status
        usb_status = self.wifi2usb.get_usb_status()
        print("\nUSB Network Interface:")
        print(f"Active: {usb_status['active']}")
        print(f"IP Configuration: {usb_status['ifconfig']}")

        print("-" * 50)
        input("Press Enter to return to main menu...")

    def connectivity_test_menu(self):
        """Test network connectivity"""
        print("\nTesting Network Connectivity:")
        print("-" * 50)

        self.wifi2usb.test_connectivity()

        print("-" * 50)
        input("Press Enter to return to main menu...")

    def enter_repl(self):
        """Enter MicroPython REPL"""
        print("\nEntering MicroPython REPL...")
        print('Type "exit()" to return to CLI')
        print("-" * 50)

        # TODO: Replace this, it sucks
        while True:
            try:
                code = input(">>> ")
                if code.strip() == "exit()":
                    break
                try:
                    result = eval(code)
                    if result is not None:
                        print(result)
                except SyntaxError:
                    exec(code)
            except Exception as e:
                print(f"Error: {e}")

        print("-" * 50)
        print("Returned from REPL")

def main():
    wifi2usb = WiFi2USB()
    cli = CLI(wifi2usb)
    cli.start()

if __name__ == "__main__":
    main()

```

andrewleech:
@xplshn I really don't know anything about how packet forwarding is supposed to work. Does your application there seem to work at all?
For your repl bit it would probably be better to integrate aiorepl from micropython-lib

FWIW I've got an example usage of this published at https://github.com/andrewleech/mpy-usb-net-mqtt-demo/ 

ps. it's _not_ RNDIS, that's the older standard protocol for usb network adapters. NCM is the newer one, it's a different protocol / driver behind the scenes. They should basically work the same as far as users are concerned though.

xplshn:
> @xplshn I really don't know anything about how packet forwarding is supposed to work. Does your application there seem to work at all?

My code works, it stablishes a WLAN connection, and an NCM connection (as you said), but packet forwarding isn't handled at all, because I do not know how, if someone can shed some light, I'd appreciate it

> For your repl bit it would probably be better to integrate aiorepl from micropython-lib
> 
> FWIW I've got an example usage of this published at https://github.com/andrewleech/mpy-usb-net-mqtt-demo/

Thank you! didn't know about these two

andrewleech:
I've been working on this PR on stm32 in conjunction with [stm32/usb: Add support for using TinyUSB stack. #15592](https://github.com/micropython/micropython/pull/15592) and it's looking quite stable now.

I found and fixed a data corruption issue (https://github.com/hathach/tinyusb/discussions/3131) and have improved the speed of the interface significantly in the process.

I'm also finding ipv6 to be highly beneficial for this use case as I don't need to worry about clashing ip addresses from multiple devices being attached to a PC. For windows testing I've got some powershell snippets that automatically find the correct ip address to communicate with for a given USB `vid/pid</serial>` which has helped. I'll clean them up and share them here too.

Josverl:
I tried to  build this PR to use in testing the `settrace` variants, but was unable to fix the build errors in both a rebase, and the original PR for multiple ports

Would it be possible for you to rebase it?

andrewleech:
Ah yeah the PR is in the middle of some large changes based on review feedback to make it more runtime configurable, however it doesn't work again yet. I've got time at work budgeted to fix it, but not yet booked in!

Josverl:
Ok, no rush

andrewleech:
# USB Dynamic Device Updates - Testing & Validation

I've been working through the USB dynamic device mode implementation and believe it's looking and working well.

This branch has gotten pretty extensive with the TinyUSB integration, USB networking, and all the device control features, it would be possible but not trivial to split this out into a separate PR if it wanted to be viewed and tested in isolation.

The branch is now / currently stacked on top of https://github.com/micropython/micropython/pull/15592 so I could use / test it on a NUCLEO board.

The main focus lately has been on runtime VID/PID configuration and making sure USB device state management works properly in both static and dynamic modes. Here's what's working:

- **Runtime VID/PID changes** - Can change USB vendor/product IDs on the fly
- **Bitfield builtin drivers** - Combine USB classes with `BUILTIN_CDC | BUILTIN_MSC | BUILTIN_NCM`
- **Static mode support** - Minimal footprint mode that still lets you control builtin drivers when `USB_RUNTIME_DEVICE=0`
- **USB NCM networking** - Full TCP/IP stack over USB with LWIP

## Testing Setup

Tested on NUCLEO_H563ZI (STM32H563) running both build modes:
- **Static**: `MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE=0` (~517KB firmware)
- **Dynamic**: `MICROPY_HW_ENABLE_USB_RUNTIME_DEVICE=1` (~519KB firmware)

**To date it hasn't been tested on other ports.**

## Test Results

### ✅ VID/PID Runtime Configuration

```python
import machine

usb = machine.USBDevice()
print(f'VID: 0x{usb.vid:04X}, PID: 0x{usb.pid:04X}')
# Output: VID: 0xF055, PID: 0x9802 (defaults)

# Change to custom VID/PID
usb.active(False)
usb.vid = 0x1234
usb.pid = 0x5678
usb.active(True)

# Device re-enumerates with new VID/PID
```

**Tested:**
- ✅ Setting custom VID/PID
- ✅ Device re-enumeration after changes
- ✅ Values persist across `active(False)/active(True)` cycles
- ✅ Resetting to defaults with `vid=0, pid=0`
- ✅ Multiple VID/PID change cycles
- ✅ Both static and dynamic modes

### ✅ Builtin Driver Combinations

The bitfield approach for combining USB classes works really well:

```python
usb = machine.USBDevice()

# Individual classes
usb.builtin_driver = usb.BUILTIN_CDC  # Just serial

# Combine with bitwise OR
usb.builtin_driver = usb.BUILTIN_CDC | usb.BUILTIN_MSC  # Serial + Mass Storage
usb.builtin_driver = usb.BUILTIN_CDC | usb.BUILTIN_NCM  # Serial + Networking

# Check dynamic properties
combo = usb.BUILTIN_CDC | usb.BUILTIN_MSC
print(combo.itf_max)  # 3 (2 for CDC + 1 for MSC)
print(combo.ep_max)   # 3
```

**Tested:**
- ✅ All individual classes (CDC, MSC, NCM)
- ✅ All combinations work
- ✅ `itf_max`, `ep_max`, `str_max` calculated correctly
- ✅ USB descriptors generated dynamically

### ✅ USB NCM Networking


```python
# Activate USB with networking
usb.active(False)
usb.builtin_driver = usb.BUILTIN_CDC | usb.BUILTIN_NCM
usb.active(True)

# Configure network
import network
usb_net = network.USB_NET()  # Important: not network.LAN()
usb_net.active(True)
usb_net.ifconfig(('192.168.7.2', '255.255.255.0', '192.168.7.1', '8.8.8.8'))

print(usb_net.active())        # True
print(usb_net.isconnected())   # True
```

**Host side:**
```bash
sudo ip addr add 192.168.7.1/24 dev enx020000000000
sudo ip link set enx020000000000 up
ping 192.168.7.2
```

**Ping results:**
```
64 bytes from 192.168.7.2: icmp_seq=1 ttl=255 time=2.88 ms
64 bytes from 192.168.7.2: icmp_seq=2 ttl=255 time=2.18 ms
5 packets transmitted, 5 received, 0% packet loss
```

**Tested:**
- ✅ USB NCM device enumeration
- ✅ LWIP stack initialization
- ✅ ICMP (ping) - works perfectly
- ✅ TCP sockets - connection established
- ✅ Network interface control

**Important:** Use `network.USB_NET()` for USB networking, not `network.LAN()` - the latter is for the Ethernet PHY on this board.

## Key Implementation Details

### VID/PID Property Pattern

The implementation uses a simple pattern where setting VID/PID to 0 restores defaults:

```c
// In machine_usb_device.c
if (attr == MP_QSTR_vid) {
    uint16_t vid = self->custom_vid;
    if (vid == 0) {
        vid = MICROPY_HW_USB_RUNTIME_VID;  // Use default
    }
    dest[0] = MP_OBJ_NEW_SMALL_INT(vid);
}
```

This means:
- Properties always return valid values
- Setting to 0 effectively resets to defaults
- Values persist across active() cycles

### Descriptor Callback

The descriptor callback checks for custom VID/PID at enumeration time:

```c
// In mp_usbd_descriptor.c
const uint8_t *tud_descriptor_device_cb(void) {
    if (MP_STATE_VM(usbd) != MP_OBJ_NULL) {
        mp_obj_usb_device_t *usbd = MP_OBJ_TO_PTR(MP_STATE_VM(usbd));
        if (usbd->custom_vid != 0 || usbd->custom_pid != 0) {
            static tusb_desc_device_t custom_desc;
            custom_desc = mp_usbd_builtin_desc_dev;
            if (usbd->custom_vid != 0) custom_desc.idVendor = usbd->custom_vid;
            if (usbd->custom_pid != 0) custom_desc.idProduct = usbd->custom_pid;
            return (const uint8_t*)&custom_desc;
        }
    }
    return (const void *)&mp_usbd_builtin_desc_dev;
}
```

This is called during enumeration, so runtime changes take effect immediately on the next `active(True)`.

### Disconnect/Connect Fix

Added proper USB disconnect/connect calls for static mode:

```c
// In machine_usb_device.c static mode
if (value) {
    mp_usbd_init();
    mp_usbd_update_class_state(builtin->flags);
    tud_connect();  // Trigger re-enumeration
} else {
    tud_disconnect();  // Clean disconnect
    mp_usbd_update_class_state(USB_BUILTIN_FLAG_NONE);
}
```

This ensures the host OS knows to re-enumerate when configuration changes.

## Common Usage Patterns

### Runtime VID/PID Change
```python
import machine

usb = machine.USBDevice()
usb.active(False)
usb.vid = 0x1234
usb.pid = 0x5678
usb.builtin_driver = usb.BUILTIN_CDC
usb.active(True)
```

### Reset to Defaults
```python
usb.active(False)
usb.vid = 0
usb.pid = 0
usb.active(True)
```

If anyone wants to repeat one of the tests, here's an example:
``` sh
#!/bin/bash
STLINK="/dev/ttyACM0"

echo ""
echo "=========================================="
echo "TEST 1.8: CDC + MSC + NCM (All Three)"
echo "=========================================="

echo "Configuring USB for CDC + MSC + NCM via ST-Link..."
mpremote connect $STLINK resume exec "
import machine
usb = machine.USBDevice()
usb.active(False)
usb.builtin_driver = usb.BUILTIN_CDC | usb.BUILTIN_MSC | usb.BUILTIN_NCM
usb.active(True)
print('USB configured: CDC + MSC + NCM')
"

echo "Waiting 12 seconds for USB re-enumeration..."
sleep 12

echo ""
echo "Checking for CDC (should be present):"
if ls /dev/ttyACM1 2>/dev/null; then
    echo "✅ PASS: CDC port present at /dev/ttyACM1"
    CDC_OK=1
else
    echo "❌ FAIL: CDC port not found"
    CDC_OK=0
fi

echo ""
echo "Checking for MSC block device:"
sleep 3  # Wait for SCSI enumeration
dmesg | tail -30 | grep -i "usb mass storage\|scsi.*micropython\|sd[a-z].*512-byte" | tail -5

MSC_DEV=$(dmesg | tail -30 | grep -oE 'sd[a-z]' | tail -1)
if [ -n "$MSC_DEV" ]; then
    echo "✅ PASS: MSC device found: /dev/$MSC_DEV"
    MSC_OK=1
else
    echo "❌ FAIL: MSC device not found"
    MSC_OK=0
fi

echo ""
echo "Checking for NCM network interface:"
NCM_IFACE=$(ip link show | grep -oE '(usb[0-9]|enx[0-9a-f]{12})' | head -1)
if [ -n "$NCM_IFACE" ]; then
    echo "✅ PASS: NCM interface found: $NCM_IFACE"
    ip link show $NCM_IFACE
    NCM_OK=1
else
    echo "Checking dmesg for NCM..."
    dmesg | tail -20 | grep -i "cdc_ncm\|ncm.*register"

    # Try again after delay
    sleep 2
    NCM_IFACE=$(ip link show | grep -oE '(usb[0-9]|enx[0-9a-f]{12})' | head -1)
    if [ -n "$NCM_IFACE" ]; then
        echo "✅ PASS: NCM interface found after delay: $NCM_IFACE"
        NCM_OK=1
    else
        echo "❌ FAIL: NCM interface not found"
        NCM_OK=0
    fi
fi

echo ""
echo "Testing CDC communication:"
if [ $CDC_OK -eq 1 ]; then
    mpremote connect /dev/ttyACM1 resume exec "print('✅ CDC communication works')" 2>&1 | head -5
fi

echo ""
echo "=========================================="
echo "TEST 1.8 SUMMARY"
echo "=========================================="
if [ $CDC_OK -eq 1 ] && [ $MSC_OK -eq 1 ] && [ $NCM_OK -eq 1 ]; then
    echo "✅ PASS: All three USB classes working simultaneously"
    echo "  - CDC: /dev/ttyACM1"
    echo "  - MSC: /dev/$MSC_DEV"
    echo "  - NCM: $NCM_IFACE"
else
    echo "❌ FAIL: Not all classes working"
    echo "  - CDC: $([ $CDC_OK -eq 1 ] && echo '✅' || echo '❌')"
    echo "  - MSC: $([ $MSC_OK -eq 1 ] && echo '✅' || echo '❌')"
    echo "  - NCM: $([ $NCM_OK -eq 1 ] && echo '✅' || echo '❌')"
fi
```

surfermarty:
Was pointed at this PR regarding a possible solution for an ESP32 project I'm working on. Having problems compiling machine_usb_device.c in extmod due to missing code for 'mp_usbd_enable_class_cdc', 'mp_usbd_enable_class_msc' and 'mp_usbd_enable_class_ncm' and the define for MICROPY_HW_USB_MSC is in shared code but not in all ports.

Happy to contribute/test for ESP32 if interested.

ant9000:
I would love to see this PR get included. I [rebased](https://github.com/ant9000/micropython/tree/usbd_net) it on current master and it does seem to compile correctly for STM32 (tried `BOARD=NUCLEO_WB55`) - but I have no board for testing it with.  My very basic test:
```
export BOARD=NUCLEO_WB55
export MICROPY_PY_LWIP=1
export CFLAGS_EXTRA="-DMICROPY_HW_TINYUSB_STACK=1 -DMICROPY_HW_NETWORK_USBNET=1"
make -C mpy-cross
cd ports/stm32
make -j clean
make -j submodules
make -j
```
goes well.

I then started fiddling a bit with ports/esp32 - here things get messy. Expressif already includes a custom TinyUSB implementation, that sports an NCM driver, under `managed_components/espressif__tinyusb/`. Moreover, ESP32 does not make use of `MICROPY_PY_LWIP` - and maybe even conflicts with it.

Basic test for a generic ESP32S3 board:

```
export BOARD=ESP32_GENERIC_S3
export CFLAGS_EXTRA="-DMICROPY_HW_NETWORK_USBNET=1"
make -C mpy-cross
cd ports/esp32
make -j clean
make -j submodules
make -j
```

@andrewleech - can I provide some help on pushing this PR further?

andrewleech:
The current direction I've been pursuing is trying to make this runtime configurable such that ncm can be enabled/disabled in boot.py like is common with other network interfaces. 

My initial attempts at this have been larger than expected (in terms of flash size) so need further work. 

That being said, I think that probably shouldn't have to be a blocking feature for this PR, I've already been able to work on that in isolation to this branch. 

I think there are still other changes in this branch though that will need to be split out into their own branch / PR - I'll have to have another go getting it cleaned up soon.

ant9000:
Thanks a lot for your continued effort in this topic - it will surely contribute to a much improved USB support for MP.

If you think I could help in any way to make it happen, please ask.

ant9000:
I managed to make ports/esp32 compile with `MICROPY_HW_NETWORK_USBNET=1`. The interface is not working, but I can see it's announcing itself as NCM to my Linux laptop - so I got one step further.

I lost an insane amount of time because configuring Espressif TinyUSB does not work with the same setup used by MicroPython CMake: namely, MP uses `CFLAGS_EXTRA` while `idf.py` requires `EXTRA_CFLAGS`. So the env needs something like:

```
BOARD=ESP32_GENERIC_S3
BOARD_VARIANT=SPIRAM_OCT
CFLAGS_EXTRA="-DMICROPY_HW_NETWORK_USBNET=1"
export BOARD BOARD_VARIANT CFLAGS_EXTRA
# idf.py uses a different env var!!!
export EXTRA_CFLAGS="$CFLAGS_EXTRA"
```

My branch is [here](https://github.com/ant9000/micropython/tree/usbd_net).

andrewleech:
This has been rebased and pretty significantly reworked. The STM32 TinyUSB enablement and a couple of standalone fixes have been split out into separate PRs (#19070, #19071) so this one now focuses on NCM networking and the runtime class selection API.

The big changes from the previous version:

The "per-driver runtime USB class control" approach with the opaque C type and internal flags byte is gone. Instead there are simple integer flag constants (`BUILTIN_FLAG_CDC`, `BUILTIN_FLAG_MSC`, `BUILTIN_FLAG_NCM`) that get OR'd together and assigned to `builtin_driver` in boot.py. This builds a dynamic configuration descriptor at runtime with compact interface/endpoint numbering. Much simpler at both the Python and C level.

`BUILTIN_DEFAULT` is now CDC-only. MSC and NCM are opt-in via the flag constants. Without `RUNTIME_DEVICE` (ie. boards that don't have `machine.USBDevice`) all compiled-in classes are exposed by default in the static descriptor.

The NCM netif is now auto-initialised at boot from `mod_network_init()` rather than waiting for Python code to create a `network.USB_NET()` object. This is necessary because TinyUSB's NCM driver (in v0.19+) starts receiving as soon as the host sends SET_INTERFACE during enumeration, which happens before any Python code runs. Without the netif being ready at that point, the receive path stalls.

The DHCP server no longer advertises a default gateway on the NCM link. The previous version caused the host to route all internet traffic over USB which broke outbound connectivity. There's a new `send_router` flag on the shared dhcpserver that controls this.

Port Makefiles for alif, mimxrt, renesas-ra and samd now use TinyUSB's upstream `tinyusb.mk` instead of hand-maintained file lists. NCM is auto-enabled on any port with lwIP and TinyUSB via `MICROPY_HW_NETWORK_USBNET`.

Tested on NUCLEO_WB55 with both `RUNTIME_DEVICE=1` and `RUNTIME_DEVICE=0`, ping working at ~1.8ms RTT over the NCM link. Haven't re-tested on rp2, mimxrt, renesas-ra or samd yet (build-tested only).

The branch has three phases that could potentially be split into separate stacked PRs if that would make review easier: background network support prerequisites, the NCM driver itself, and the runtime USB class selection API.

extmod/modnetwork.c
```diff
@@ -205,6 +215,10 @@ static const mp_rom_map_elem_t mp_module_network_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR_WLAN), MP_ROM_PTR(&mod_network_esp_hosted_type) },
     #endif
 
+    #if MICROPY_HW_NETWORK_USBNET
```
dpgeorge: Suggest this be consistent with other `network` config options above, eg `MICROPY_PY_NETWORK_USB_NET`.

extmod/network_usbd_ncm.c
```diff
@@ -0,0 +1,396 @@
+/*
```
dpgeorge, line 1: This filename should match the class name, eg `network_usb_net.c`.

But, also, maybe the class should be `network.USB_NCM`?

ports/rp2/mpconfigport.h
```diff
@@ -198,6 +198,9 @@
 #define MICROPY_VFS_FAT                         (1)
 #define MICROPY_VFS_ROM                         (MICROPY_HW_ROMFS_BYTES > 0)
 #define MICROPY_SSL_MBEDTLS                     (1)
+#ifndef MICROPY_HW_NETWORK_NCM
+#define MICROPY_HW_NETWORK_NCM               (MICROPY_PY_LWIP && MICROPY_HW_TINYUSB_STACK)
```
projectgus: I was wondering why there's no code size impact reported on rp2, and it's because `MICROPY_HW_TINYUSB_STACK` flag is only used on the stm32 port.

Not sure what the best default for mimxrt, renesas and rp2 is for this...?

andrewleech: Oh good catch thanks, that define isn't needed for rp2 at all.

Some time back I discussed the default on/off of NCM with @dpgeorge and I'm pretty sure we landed on having it on if the network stack is already in use. 

That being said, it could be confusing for people to suddenly get a network connection pop up after plugging in a pico-w and it is consuming USB resources that won't be available for dynamic USB devices so it probably should be off by default everywhere currently. 

I've now got runtime / boot.py configurability for all tinyusb built-in classes in a separate PR stacked on top of this one that should resolve this kind of confusion. 
https://github.com/micropython/micropython/pull/19103

extmod/network_usbd_ncm.c
```diff
@@ -0,0 +1,405 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "py/runtime.h"
+#include "py/mphal.h"
+
+#if MICROPY_PY_NETWORK_NCM
```
dpgeorge: Suggest this option be called `MICROPY_PY_NETWORK_USBD_NCM`, to match the class name, and this filename.

dpgeorge:
Testing this by enabling `#define MICROPY_PY_NETWORK_USBD_NCM (1)`, it has some build errors (after rebasing to latest master):
- rp2/RPI_PICO_W: 
```
micropython/extmod/network_usbd_ncm.c: In function 'ncm_init':
micropython/extmod/network_usbd_ncm.c:187:26: error: 'MP_ENOMEM' undeclared (first use in this function); did you mean 'MP_QSTR_ENOMEM'?
  187 |         mp_raise_OSError(MP_ENOMEM);
      |                          ^~~~~~~~~
      |                          MP_QSTR_ENOMEM
```
- stm32/PYBD_SF6:
```
../../shared/tinyusb/mp_usbd_descriptor.c:85:169: error: macro 'TUD_CDC_NCM_DESCRIPTOR' requires 11 arguments, but only 9 given
   85 | CM_EP_OUT, USBD_NCM_EP_IN, USBD_NCM_IN_OUT_MAX_SIZE, CFG_TUD_NET_MTU),
      |                                                                     ^
In file included from ../../lib/tinyusb/src/tusb.h:81,
                 from ../../shared/tinyusb/mp_usbd_descriptor.c:32:
../../lib/tinyusb/src/device/usbd.h:1068:9: note: macro 'TUD_CDC_NCM_DESCRIPTOR' defined here
 1068 | #define TUD_CDC_NCM_DESCRIPTOR(_itfnum, _desc_stridx, _mac_stridx, _ep_notif, _ep_notif_size, _epout, _epin, _epsize, _maxsegmentsize, _ep_notif_interval, _capability) \
      |         ^~~~~~~~~~~~~~~~~~~~~~
```

andrewleech:
Thanks @dpgeorge yes there was a #include needed and fixes for the newer tinyusb. I've compiled for both those boards in the latest push.

dpgeorge:
Thanks for updating.  I can now build for both RPI_PICO_W and PYBD_SF6.

I tested both of those boards and it works really well!  It basically works out of the box, `network.USBD_NCM()` is already active on start up, and running `dhcpcd enp0s20f0u2i3` on the host PC gives the PC an IP address.  From there the host can ping the board.

I could run all `tests/multi_net/*.py` tests against the host PC and get them to pass in both directions:
```
$ ./run-multitests.py -t a0 -p2 multi_net/*.py
multi_net/asyncio_tcp_client_rst.py on ttyACM0|micropython: pass
multi_net/asyncio_tcp_close_write.py on ttyACM0|micropython: pass
multi_net/asyncio_tcp_readall.py on ttyACM0|micropython: pass
...
39 tests performed
39 tests passed
1 tests skipped: multi_net/sslcontext_getpeercert.py
multi_net/asyncio_tcp_client_rst.py on micropython|ttyACM0: skip
multi_net/asyncio_tcp_close_write.py on micropython|ttyACM0: pass
multi_net/asyncio_tcp_readall.py on micropython|ttyACM0: pass
...
37 tests performed
37 tests passed
3 tests skipped: multi_net/asyncio_tcp_client_rst.py multi_net/sslcontext_getpeercert.py multi_net/tcp_client_rst.py
```

docs/library/network.USBD_NCM.rst
```diff
@@ -0,0 +1,81 @@
...
+This class provides a network interface over USB using the NCM (Network
+Control Model) protocol.  The host computer sees this device as a USB
+Ethernet adapter and assigns it an IP address via DHCP (served by the
+MicroPython device).
+
+.. note:: ``network.USBD_NCM`` requires a port with TinyUSB and NCM support,
+          enabled at build time by defining ``MICROPY_PY_NETWORK_USBD_NCM``
+          (off by default).  The USB NCM class must also be active - either
+          compiled as the default or enabled in ``boot.py`` via
+          ``machine.USBDevice.BUILTIN_NCM``.
```
dpgeorge: This constant doesn't exist at the moment.   I don't know if you intended to add it in this PR?

extmod/network_usbd_ncm.c
```diff
@@ -0,0 +1,406 @@
...
+ * THE SOFTWARE.
+ */
+
+#include "py/runtime.h"
+#include "py/mperrno.h"
+#include "py/mphal.h"
+
+#if MICROPY_PY_NETWORK_USBD_NCM
+
+// Include NCM configuration (MICROPY_PY_NETWORK_USBD_NCM_DHCP_SERVER, CFG_TUD_NCM_* sizes).
```
dpgeorge: The `CFG_TUD_NCM_*` sizes don't seem to be in the below header.  Maybe the comment can be updated?

extmod/network_usbd_ncm.c
```diff
@@ -0,0 +1,406 @@
...
+    #endif
+} ncm_obj_t;
+
+// Global object holding the usb net state
+static ncm_obj_t ncm_obj;
+
+static mp_sched_node_t ncm_sched_node;
+// Pending tud_network_recv_renew() flag. mp_sched_schedule_node is a
+// one-slot queue; this flag is consumed once per scheduler drain.
+static volatile bool ncm_recv_renew_pending;
```
dpgeorge: Is this flag needed at all?  From what I can tell it has a 1-1 correspondence with the node being on the schedule queue, so if the node is scheduled the flag is true.  Hence the flag is not needed.

extmod/network_usbd_ncm.c
```diff
@@ -0,0 +1,406 @@
...
+// netif_add (which must only be called once per netif).
+void ncm_auto_init(void) {
+    if (!ncm_obj.init) {
+        ncm_init();
+    } else {
+        // On soft-reset mod_network_init() clears the NIC list, so
+        // re-register the (still-live) singleton.
+        mod_network_register_nic(&ncm_obj);
+    }
+    netif_set_up(&ncm_obj.netif);
```
dpgeorge, line 306: This makes `USBD_NCM().active()` be True on boot.  Is that intentional?

extmod/network_usbd_ncm.c
```diff
@@ -0,0 +1,406 @@
...
+static void _scheduled_tud_network_recv_renew(mp_sched_node_t *node) {
+    // Acknowledge the pending receive renew. tud_network_recv_renew() must
+    // be called after each received packet to release the TinyUSB receive
+    // buffer; mp_sched_schedule_node is a single-slot queue so at most one
+    // call is pending at any time.
+    uint32_t atomic_state = MICROPY_BEGIN_ATOMIC_SECTION();
+    if (ncm_recv_renew_pending) {
+        ncm_recv_renew_pending = false;
+        MICROPY_END_ATOMIC_SECTION(atomic_state);
+        tud_network_recv_renew();
```
dpgeorge: As mentioned above, I don't think `ncm_recv_renew_pending` is needed.  Just call `tud_network_recv_renew()` unconditionally.

shared/tinyusb/tusb_config.h
```diff
@@ -94,55 +94,122 @@
...
+#ifndef CFG_TUD_NCM_IN_NTB_N
+#define CFG_TUD_NCM_IN_NTB_N  (2 + LWIP_MDNS_RESPONDER)
+#endif
+
+#ifndef MICROPY_PY_NETWORK_USBD_NCM_INTERFACE_STRING
+#define MICROPY_PY_NETWORK_USBD_NCM_INTERFACE_STRING "Board NET"
+#endif
+#endif // MICROPY_PY_NETWORK_USBD_NCM
+
+#define USBD_RHPORT (0) // Currently only one port is supported
```
dpgeorge: Looks like this macro is never used anywhere?

shared/tinyusb/tusb_config.h
```diff
@@ -94,55 +94,122 @@
...
+
+// define the matching in endpoints to each EP_OUT
 #if CFG_TUD_CDC
-#define USBD_ITF_CDC (0) // needs 2 interfaces
-#define USBD_CDC_EP_CMD (0x81)
-#define USBD_CDC_EP_OUT (0x02)
-#define USBD_CDC_EP_IN (0x82)
-#endif // CFG_TUD_CDC
-
+#define USBD_CDC_EP_OUT  ((USBD_CDC_EP_IN)&~TUSB_DIR_IN_MASK)
```
dpgeorge: I'm not sure why uncrustify doesn't pick this up, but maybe add a space around `&` to make it a little more readable?

docs/library/network.USBD_NCM.rst
andrewleech: Whoops, that's leaked in from the PR stacked on top of this one (#19103) which adds the `BUILTIN_*` flags. I've dropped that sentence here so this PR's note only covers the build time `MICROPY_PY_NETWORK_USBD_NCM` define, and moved the `BUILTIN_NCM` mention into the docs commit of #19103 where the constant actually gets added.

extmod/network_usbd_ncm.c
andrewleech: Yep you're right, they're over in `shared/tinyusb/tusb_config.h`. Comment fixed to only mention what the header actually provides.

extmod/network_usbd_ncm.c
andrewleech: Agreed, it's gone, along with the two atomic sections it needed. Tinyusb only ever calls recv_cb once per renew, so the flag could only ever be set once anyway.

Tested on a WB55: 4 MiB host to device arrived complete.

extmod/network_usbd_ncm.c
andrewleech: Done - see the reply on the earlier comment, it's just `tud_network_recv_renew()` unconditionally now.

extmod/network_usbd_ncm.c
andrewleech: Yes, intentional. The host asks for its DHCP address as soon as the USB interface starts, which is before boot.py runs, and lwIP drops anything arriving on a netif that isn't up.

shared/tinyusb/tusb_config.h
andrewleech: Good catch. `USBD_RHPORT` is left over from the original version of this branch - I'd missed that it got deleted in master @ 521b2f86be. Deleted.

shared/tinyusb/tusb_config.h
andrewleech: Turns out uncrustify was actively taking the spaces back out - with the operand parenthesised it reads `(X)&~Y` as an address-of rather than a bitwise and, so `codeformat.py` kept reverting it. Dropping the redundant parens around the enum makes it parse as binary and the spacing sticks:

```c
#define USBD_CDC_EP_OUT  (USBD_CDC_EP_IN & ~TUSB_DIR_IN_MASK)
```

Same for the MSC and NCM ones.

andrewleech:
Thanks for the review @dpgeorge, all six are addressed in the threads above.

Testing those changes on a NUCLEO_WB55 turned up a few things unrelated to the review,
all now fixed:

- `linkoutput_fn` returned `ERR_USE` when the USB transmit buffer was busy, which comes
  back out of `socket.send()` as `OSError: [Errno 98] EADDRINUSE`. Since `tcp_write()` has
  already queued the data at that point, an application that retries also duplicates it in
  the stream. It now drops the frame and lets TCP retransmit.
- `active(False)` only changed the lwIP state, so the host kept the link up and carried on
  sending into an interface that was discarding everything. It now signals link state to
  the host, including an override of tinyusb's `tud_network_default_link_state_cb()` so a
  bus reset doesn't bring the host side back up on its own.
- Dropped `tud_network_init_cb()`, which is dead code on tinyusb 0.21.0.

For anyone wanting to try it, no board in tree enables any of this by default, so the
WB55 build is:

```
make -C ports/stm32 submodules BOARD=NUCLEO_WB55
git submodule update --init lib/lwip
make -C mpy-cross
make -C ports/stm32 BOARD=NUCLEO_WB55 MICROPY_PY_LWIP=1 \
    CFLAGS_EXTRA='-DMICROPY_HW_TINYUSB_STACK=1 -DMICROPY_PY_NETWORK_USBD_NCM=1'
```

`MICROPY_PY_LWIP=1` is needed because NUCLEO_WB55 has no networking of its own, and
`lib/lwip` doesn't come down with the normal submodules target for that reason. Fits in
the 512K part with BLE still enabled, around 87% of flash.

dpgeorge (approved):
Thanks for updating.  I retested on PYBD_SF6 (enabling USBD_NCM) and the multi-tests pass.

Excellent work on this!

Issue #8819, opened 2022-06-25
ESP32 WPA2 Enterprise


MATTYGILO:
Micropython would feel more complete with WPA Enterprise for esp32. This is an essential function that is not yet met.

All help and suggestions would be great. If there are way around this for now.

MATTYGILO:
Been requested since 2019:

[https://forum.micropython.org/viewtopic.php?f=18&t=7219](url)
[https://forum.micropython.org/viewtopic.php?t=10543](url)

SL2021-Dev:
I am also in need of an enterprise login capabilities.
Arduino, Pycom (I know... GONE)  and CircuitPython have this ability.
v1.21 is about to come out, PLEASE INCLUDE
It has been 2 years 4 months since...
![image](https://github.com/micropython/micropython/assets/139779102/84c6b38e-1220-4cfe-bf4f-e2a065e0f2ce)

MATTYGILO:
@SL2021-Dev 

Step 1) 
Fork micropython repo
Add wpa_supplicant to set(IDF_COMPONENTS …) in /micropython/ports/esp32/main/CMakeLists.txt

Step 2)
Create a custom micropython module (Provided) and reference it with your build


enterprise.c:
```
#include <string.h>

#include "py/builtin.h"
#include "py/runtime.h"
#include "py/mpconfig.h"
#include "py/objstr.h"
#include "py/obj.h"
#include "py/stream.h"
#include "py/mperrno.h"

#include "esp_wifi.h"
#include "esp_wpa2.h"
#include "esp_log.h"
#include "esp_event.h"

#include "modnetwork.h"


// Set up EAP
STATIC mp_obj_t esp_seteap(mp_obj_t username, mp_obj_t password){

    // The length of username and password
    size_t Ilen;
    size_t Plen;

    // Converts from micropython form to C form
    const char *EAP_IDENTITY = mp_obj_str_get_data(username,&Ilen);
    const char *EAP_PASSWORD = mp_obj_str_get_data(password,&Plen);

    // Sets the identity and password
    esp_exceptions(esp_wifi_sta_wpa2_ent_set_identity((uint8_t *)EAP_IDENTITY, strlen(EAP_IDENTITY)));
    esp_exceptions(esp_wifi_sta_wpa2_ent_set_username((uint8_t *)EAP_IDENTITY, strlen(EAP_IDENTITY)));
[...]

Issue #8670, opened 2022-05-16
How can I use WPA2-Enterprise on ESP32?


MATTYGILO:
How do I connect to WPA2-Enterprise on a ESP32?

Issue #5705, opened 2020-02-27
ESP32: WPA2-Enterprise support 


wgaylord:
Hello, I am making this issue to request support for WPA2 enterprise network that use EAP.

I had late 2019 posted some code on how to do so that I used in a custom build of micropython so that someone with more knowledge of how to properly add it could take a look.

https://forum.micropython.org/viewtopic.php?f=18&t=7219

Looks like no one ever did which is why I am creating this issue to bring attention to this.

wgaylord:
Will there ever be the possibility of someone looking into this? I don't want to keep building custom firmware.  

Or should I spend my time trying to figure out the proper way to implement this in and make a PR? (Wish there was docs about how to do the exceptions, checks and stuff on the C side of things)

tve:
Yup, a pull request is the way these things happen. I wouldn't even have a WPA2 enterprise AP to test against...
WRT handling errors, there's a `check_esp_err` function (https://github.com/micropython/micropython/blob/master/ports/esp32/mphalport.c#L57-L87) used for example in https://github.com/micropython/micropython/blob/master/ports/esp32/esp32_partition.c#L154

GC-RnD:
We have a project in development that is now required to have WPA2-Enterprise encryption.
I see the Arduino and the Pycom groups have this ability.
Has anyone ever seen a library available for this.
This is becoming required for commercial networks and I hope the micropython group will move forward to including Enterprise connectivity.

wgaylord:
I will be trying to make a PR for this sometime soon. Have to set up my Micropython build system again. Plus I am probably going to be loosing access the the WPA2 EAP network I needed this for soon because I am graduating in a few days.

wgaylord:
I will be working on a PR for this. Finally get a good amount of experience working with Micropython's internals that I think i can make a PR properly.

ATM-HSW:
Anything news about implementing WPA2 EAP?

MATTYGILO:
How's this coming on?

one0410:
Hi, I really need WPA2-Enterprise feature on ESP32.
Please add this feature.
Thx.

SL2021-Dev:
Same...
I also have been using hacks with help from other members in order to provide PEAP .
It has been some time now and an offical release should be worked on.
I have also mentioned this in the Core Development forum....
https://github.com/orgs/micropython/discussions/12143

Please... seriously consider adding this to Micropython !!!

mattytrentini:
[...]

PR #16463, opened 2024-12-20
esp32: add WPA2 enterprise support

h-milz:
### Summary

This PR adds WPA2 Enterprise support to the ESP32 port. In particular, the patch supports EAP-PWD, EAP-PEAP, and EAP-TTLS with MSCHAPv2 and PAP. 

### Testing

The patch was developed and tested in the Technical University of Munich eduroam network on an ESP32_GENERIC_S3 board, namely, a CrowPanel 5.0"-HMI ESP32 Display board with a ESP32-S3-WROOM-1-N4R8 module. 

Usage example: 

```
import network

wlan.active(True)

identity = "anonymous@eduroam.mwn.de"
username = "my_username@eduroam.mwn.de" 
password = "my_password"
certfile = "/T-TeleSec_GlobalRoot_Class_2.pem"  # needs to be uploaded first
ssid = "eduroam"
method = wlan.EAP_method
        
with open (certfile, 'rb') as file:
    ca_cert = file.read()
            
try:
    if method == wlan.EAP_PWD:
        wlan.eap_connect(ssid=ssid, eap_method=method, 
                        username=username, password=password)
    elif method == wlan.EAP_PEAP:
        wlan.eap_connect(ssid=ssid, eap_method=method, 
                        username=username, password=password, 
                        identity=identity, ca_cert=ca_cert)        
    elif method == wlan.EAP_TTLS:  
        wlan.eap_connect(ssid=ssid, eap_method=method, 
                        username=username, password=password, 
                        identity=identity, ca_cert=ca_cert,
                        ttls_phase2_method=ttls_phase2_method)
except Exception as e:
    print (f"error: {e}")

```

If your board does not have a hardware RTC, odds are that the server certificate validation for EAP-PEAP, -TTLS and potentially -TLS will fail due to the system time being way off. As a workaround, you can set the system time to build time on system start like this: 

```
import sys
import machine

(year, month, day) = sys.version.split(" on ")[1].split("-")
rtc = machine.RTC()
date_time = (int(year), int(month), int(day), 0, 0, 0, 0, 0)
rtc.init(date_time) 

``` 

EAP-TTLS with MSCHAP or CHAP could not be tested so far, neither could EAP-TLS, lacking a corresponding network. User testing and feedback is highly appreciated! 

Development and testing were done with ESP-IDF 5.2.2 and MPY 1.23.0 because later versions or master gave me too many runtime or compilation errors (in other places than this patch). 

If requested, I can add more documentation for the README or the MPY documentation.


Carglglz:
related #8819 #8670 #5705 #12143

h-milz:
I would like to add that I am aware that my time setting workaround is really an ugly hack that is only for testing but should not be used in a production environment. The ESP32 RTC does not survive a reboot because it is not battery buffered. Hence, when the system boots, the system time will restart with the build date using the hack. In case a server certificate is renewed (they are usually valid for a year or even less, e.g. for LetsEncrypt), then the validation will fail. In this case you will have to build a new image and reflash the chip.  

(Mind you that in eduroam or other wifi networks supporting EAP-PWD this no issue at all - in this case it should just work. The rule of thumb is that if your IT department documents how to set up Android phones with EAP-PWD it will be just fine. But I'm told that not all eduroam networks do that which is why I wrote the code for PEAP et al. in the first place.)

The better solution whenever you deal with SSL certificates in production is to have a battery buffered RTC, which you can set once before shipping and then e.g. once per day using NTP. 

The function `esp_eap_client_set_disable_time_check()` cannot be used to circumvent this issue because it only suppresses the validity check for the local client certificate when using EAP-TLS. Again, this is a hack. It's better to (automatically) renew client certificates before expiry, also because the server may refuse validation when presented an expired client cert. 

This being said, although I cannot test ESP-TLS myself I would be surprised if my code did not work out of the box, because all I did is wrap the ESP-IDF `wpa_supplicant` API. And yes, it is port specific and will not help the STM32 or RP2 guys e.g. using CYW43 chips. ¯\_(ツ)_/¯

Carglglz:
> I would like to add that I am aware that my time setting workaround is really an ugly hack that is only for testing but should not be used in a production environment. The ESP32 RTC does not survive a reboot because it is not battery buffered. Hence, when the system boots, the system time will restart with the build date using the hack.

An alternative could be saving current date/time in NVS periodically (see https://docs.micropython.org/en/latest/library/esp32.html#esp32.NVS)

Josverl:
I would recommend to drop any and all support for MSCHAPv2 and CHAP,
As they are weak, are vulnerable to dictionary attracts, and are therefore rarely used anymore.

zcattacz:
Future proof maybe a good point , and CHAP does seem to have die out, but MSCHAPv2 is still not rare.
Not sure if by any chance, the sync clock timestamp can be captured from WiFi beacon management frames. Usually this is useless with home routers. Corp managed APs may be different.

Josverl:
Support Status
MSCHAPv2 is still a supported protocol, but its usage is becoming increasingly limited due to security concerns. Microsoft has recommended moving away from MSCHAPv2-based connections to more secure, certificate-based authentication methods such as PEAP-TLS or EAP-TLS

Security Risks
MSCHAPv2 has several known security vulnerabilities:
- **Weak Encryption**: The DES encryption used in MSCHAPv2 is considered weak and can be easily broken with modern hardware[1](https://en.wikipedia.org/wiki/MS-CHAP).
- **Credential Theft**: MSCHAPv2 is vulnerable to credential theft due to misconfigurations and weak encryption. Attackers can exploit misconfigured devices by imitating legitimate networks and decrypting user credentials[2](https://www.securew2.com/blog/peap-mschapv2-vulnerability).
- **Evil Twin Attacks**: MSCHAPv2 is susceptible to Evil Twin attacks, where attackers set up a rogue access point to intercept and decrypt user credentials[3](https://www.securew2.com/blog/security-analysis-of-peap-mschapv2).

Recommendations
Given these vulnerabilities, it is highly recommended to transition to more secure authentication methods:
- **EAP-TLS**: This method uses certificate-based authentication, eliminating password-related risks and increasing security[3](https://www.securew2.com/blog/security-analysis-of-peap-mschapv2).
- **PEAP-TLS**: Similar to EAP-TLS, this method also uses certificates for authentication, providing a secure alternative to MSCHAPv2[3](https://www.securew2.com/blog/security-analysis-of-peap-mschapv2).

My Conclusion
While MSCHAPv2 is still supported, its security risks make it an unsuitable choice for modern networks. Transitioning to EAP-TLS or PEAP-TLS is strongly recommended.

dpgeorge:
Thanks for the contribution, this is a very nice feature to add.

> And yes, it is port specific and will not help the STM32 or RP2 guys

Right, so this is the main thing to discuss and get right first: how this API will generalise to other WLAN drivers.

Should it be a new method like `eap_connect()`, or should the existing `connect()` method be reused and new security arguments/options added to it?  The latter seems the more obvious path to me.

Then, it would be great to try implementing this on at least one other WLAN driver, to see how general it can be.

zcattacz:
It's unlikely a client side choice. In order to plug their loopholes, organizations need to prioritize spendings on IT infrastructure, accelerate old device depreciation, enforce robust and consistent network policy.

Josverl:
> organizations need to prioritize spendings on IT infrastructure, accelerate old device depreciation,

Fully agree, let's give these orgs more reasons not to do the right thing

h-milz:
> > organizations need to prioritize spendings on IT infrastructure, accelerate old device depreciation,
> 
> Fully agree, let's give these orgs more reasons to do the right thing

As if universities or anyone else changed their policies because Micropython decided to not implement something. This stance is one of the reasons why people have a hard time using Micropython, 

In more than 30 years in open source, one thing I learned is that it's all about choice. Tell people what this is, and let them make an informed decision. But don't patronize them. 

A couple of people approached me in the meantime and asked if and how they could add the PR to their codebase because the MP developers seemed not to be willing to merge it. Maybe I should cancel the PR and enable people to more easily integrate it on their own.

Josverl:
@h-milz , 
Did you see that this PR was planned for inclusion in the next release?

robert-hh:
In that case it should be re-opened.

h-milz:
> @h-milz , 
> Did you see that this PR was planned for inclusion in the next release?

No but in this case I'll be more than happy to contribute. 

As I wrote above, some of the methods are so far untested anyway and I'd declare them experimental. There needs to be some documentation written, too. I'd suggest to extend ports/esp32/README.md. Shall I create a PR for this one too?

robert-hh:
You can add commits for the documentation to this PR. No need to create another PR. Just add the commits to you local branch and (force-)push it to your micropython fork. Then the existing PR will be updated.

h-milz:
> > some of the methods are so far untested anyway and I'd declare them experimental
> 
> Please test them. No untested code please.
> 

I do understand your concern, but as much as I would like to - as I wrote above, I do not have access to any network infrastructure offering anything other than EAP-PWD and EAP-PEAP. My uni does not implement anything else, and the only way I see to get this tested is to get the code in the hands of people who can. This is not going to happen when we remove the corresponding code.

Also, please note that I did not implement any of the network and authentication modes myself. My code is a wrapper around functions provided by the ESP-IDF, where these methods are officially supported. 

How about this. I'd restrict the code to EAP-PWD and EAP-PEAP by setting the appropriate max value for `eap_method` to `WIFI_AUTH_EAP_PEAP` in 

```
    // parameter check
    if (eap_method < WIFI_AUTH_EAP_PWD || eap_method > WIFI_AUTH_EAP_TLS) {
        mp_raise_ValueError(MP_ERROR_TEXT("unknown config value for eap_method"));
    }

```
in `ports/esp32/network_wlan.c` , leaving the other code as-is and documenting properly how to deal with it. That would facilitate others who want to test the code to do that in any easy way without jeopardizing the integrity of the next MP release. 

Alternatively, when a user selects EAP-TTLS or EAP-TLS, issue a proper warning that the code is experimental, and on top mark it clearly as experimental in the documentation (i.e., `ports/esp32/README.md`). For the reasons given, I'd prefer this route. 

Please advise. 

Actually, I will have some time tomorrow to implement and test the code including minor changes, given that my PR was created against v1.23.0, and re-commit. BTW what is the deadline for inclusion in v1.26.0?

robert-hh:
It's good that you documented what you have tested and what not. Maybe someone else follows your call to test other methods. You have section in `network_wlan_eap_connect()` for EAP-TTLS or EAP-TLS, so even if your network does not support these, you can test if the call handling in your code is working as expected.
You should rebase your PR on top of the actual repository state, and adding a section for the documentation is needed.
I do not know of a deadline for v1.26. It's on the list of commits to be included. (see https://github.com/micropython/micropython/milestone/10) So you will get feedback when the review for the inclusion happens.

Josverl:
To get clarification on what is tested, and where help in testing is needed I'd suggest editing the initial message in this PR , to add a list of configs to test, and check the ones that have been tested.
That makes it simpler for others to discover what needs testing.

robert-hh:
That's quite a few commits. Did you rebase your branch before pushing and resolve the conflicts? Then the number of commits should be few.

h-milz:
> That's quite a few commits. Did you rebase your branch before pushing and resolve the conflicts? Then the number of commits should be few. 

I'm afraid something went wrong when I tried to do this. I may have to start over. It should in fact only be 3 commits.

h-milz:
This PR is FUBAR and superseded by #17234.

PR #16468, opened 2024-12-22
Some improvement in esp32 readme.

garywill:
### Summary

1. To add user to Linux `dialout` group, `usermod` is the universal Linux way. `adduser` is Debian-based way.
2. When installing IDF, we don't have to install all toolchains for all chips.
3. List currently MPY supported chip models.
4. ~~Use `network.STA_IF` instead of `network.WLAN.IF_STA`~~
5. Other typo and gramma corrections

### DIFF
--- a/ports/esp32/README.md
+++ b/ports/esp32/README.md
@@ -5,6 +5,9 @@ This is a port of MicroPython to the Espressif ESP32 series of
 microcontrollers.  It uses the ESP-IDF framework and MicroPython runs as
 a task under FreeRTOS.
 
+Currently supports ESP32, ESP32-C3, ESP32-C6, ESP32-S2 and ESP32-S3
+(ESP8266 is supported by a separate MicroPython port).
+
 Supported features include:
 - REPL (Python prompt) over UART0.
 - 16k stack for the MicroPython task and approximately 100k Python heap.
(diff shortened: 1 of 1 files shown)


ports/esp32/README.md
```diff
@@ -195,7 +202,7 @@ quickly call `wlan_connect()` and it just works):
 ```python
 def wlan_connect(ssid='MYSSID', password='MYPASS'):
     import network
-    wlan = network.WLAN(network.WLAN.IF_STA)
+    wlan = network.WLAN(network.STA_IF)
```
projectgus, line 205: `network.WLAN.IF_STA` is actually preferred here, the top-level namespace constants like `network.STA_IF` are deprecated and will be removed at some point in the future. (This one was only changed recently in 285e1d0b)

projectgus (changes requested):
This is good, thanks @garywill for taking the time to submit these improvements!

I have one small request for a change, but otherwise looks good to me.

I think probably we should merge the "Getting a Python prompt on the device" and "Configuring WiFi" steps into the main tutorial documentation and link to there from here, instead of maintaining two near-identical sections of doc. However that can be left for someone to do in the future, doesn't have to be done here unless you're feeling very keen (as the ESP32 tutorial currently redirects into the esp8266 tutorial at about this point.)

ports/esp32/README.md
garywill: Corrected in newly commit

projectgus (approved):

PR #16475, opened 2024-12-24
docs/rp2: Add wlan information to the quickref.

mattytrentini:
### Summary

Some rp2 boards include wifi at least with the very popular Pico W and Pico 2 W. 
New users frequently ask how to set up wifi and are confused because it's not covered in the quickref; this PR adds the wlan section (copied and modified with notes from the ESP32 quickref).

### Testing

I built the docs locally with:

```bash
docker run --rm -v `pwd`:/mp -w /mp/docs -p 8000:8000 minidocks/sphinx-doc sphinx-autobuild --host 0.0.0.0 . _build/
```

And the changes look correct.

It would be better if the quickrefs were consolidated (for all ports) rather than duplicated, and I think the `machine` module is  close enough to parity between ports that this would be possible. However, in the meantime, adding wlan to the rp2 port seems reasonable. 

### Todo

~~This still needs work, particularly the LAN section, as I'm not familiar enough with rp2 LAN boards (particularly the Wiznet range) to know what's supported or how it works. I'm happy to make changes if other folks can provide appropriate information.~~

> [!NOTE]
> I've removed the LAN section in order to get moving with the (more important) wlan section.

### DIFF
--- a/docs/esp32/quickref.rst
+++ b/docs/esp32/quickref.rst
@@ -83,30 +83,30 @@ The :class:`network.WLAN` class in the :mod:`network` module::
 
     import network
 
-    wlan = network.WLAN(network.WLAN.IF_STA) # create station interface
-    wlan.active(True)       # activate the interface
-    wlan.scan()             # scan for access points
-    wlan.isconnected()      # check if the station is connected to an AP
+    wlan = network.WLAN()       # create station interface (the default, see below for an access point interface)
+    wlan.active(True)           # activate the interface
+    wlan.scan()                 # scan for access points
+    wlan.isconnected()          # check if the station is connected to an AP
     wlan.connect('ssid', 'key') # connect to an AP
-    wlan.config('mac')      # get the interface's MAC address
-    wlan.ipconfig('addr4')  # get the interface's IPv4 addresses
+    wlan.config('mac')          # get the interface's MAC address
+    wlan.ipconfig('addr4')      # get the interface's IPv4 addresses
 
     ap = network.WLAN(network.WLAN.IF_AP) # create access-point interface
-    ap.config(ssid='ESP-AP') # set the SSID of the access point
-    ap.config(max_clients=10) # set how many clients can connect to the network
-    ap.active(True)         # activate the interface
+    ap.config(ssid='ESP-AP')              # set the SSID of the access point
+    ap.config(max_clients=10)             # set how many clients can connect to the network
+    ap.active(True)                       # activate the interface
 
 A useful function for connecting to your local WiFi network is::
 
     def do_connect():
-        import network
-        wlan = network.WLAN(network.WLAN.IF_STA)
+        import machine, network
+        wlan = network.WLAN()
         wlan.active(True)
         if not wlan.isconnected():
             print('connecting to network...')
             wlan.connect('ssid', 'key')
             while not wlan.isconnected():
-                pass
+                machine.idle()
         print('network config:', wlan.ipconfig('addr4'))
 
 Once the network is established the :mod:`socket <socket>` module can be used
--- a/docs/rp2/quickref.rst
+++ b/docs/rp2/quickref.rst
@@ -55,6 +55,57 @@ The :mod:`rp2` module::
 
     import rp2
 
+Networking
+----------
+
+WLAN
+^^^^
+
+.. note::
+   This section applies only to devices that include WiFi support, such as the `Pico W`_ and `Pico 2 W`_.
+
+The :class:`network.WLAN` class in the :mod:`network` module::
+
+    import network
+
+    wlan = network.WLAN()       # create station interface (the default, see below for an access point interface)
+    wlan.active(True)           # activate the interface
+    wlan.scan()                 # scan for access points
+    wlan.isconnected()          # check if the station is connected to an AP
+    wlan.connect('ssid', 'key') # connect to an AP
+    wlan.config('mac')          # get the interface's MAC address
+    wlan.ipconfig('addr4')      # get the interface's IPv4 addresses
+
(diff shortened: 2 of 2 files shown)


docs/rp2/quickref.rst
```diff
@@ -55,6 +55,60 @@ The :mod:`rp2` module::
...
+
+Once the network is established the :mod:`socket <socket>` module can be used
+to create and use TCP/UDP sockets as usual, and the ``requests`` module for
+convenient HTTP requests.
+
+After a call to ``wlan.connect()``, the device will by default retry to connect
+**forever**, even when the authentication failed or no AP is in range.
+``wlan.status()`` will return ``network.STAT_CONNECTING`` in this state until a
+connection succeeds or the interface gets disabled.  This can be changed by
+calling ``wlan.config(reconnects=n)``, where n are the number of desired reconnect
```
dpgeorge, line 105: This "reconnects" option is not available for the cyw43 driver / Pico W.

docs/rp2/quickref.rst
```diff
@@ -55,6 +55,60 @@ The :mod:`rp2` module::
...
+^^^^
+
+.. note::
+   This section applies only to devices that include WiFi support, such as the `Pico W`_ and `Pico 2 W`_.
+
+The :class:`network.WLAN` class in the :mod:`network` module::
+
+    import network
+
+    wlan = network.WLAN(network.WLAN.IF_STA) # create station interface
```
dpgeorge, line 71: I suggest using just `wlan = network.WLAN()` because STA is (always) the default and it's simpler.

docs/rp2/quickref.rst
```diff
@@ -55,6 +55,60 @@ The :mod:`rp2` module::
...
+    wlan.scan()             # scan for access points
+    wlan.isconnected()      # check if the station is connected to an AP
+    wlan.connect('ssid', 'key') # connect to an AP
+    wlan.config('mac')      # get the interface's MAC address
+    wlan.ipconfig('addr4')  # get the interface's IPv4 addresses
+
+    ap = network.WLAN(network.WLAN.IF_AP) # create access-point interface
+    ap.config(ssid='RP2-AP') # set the SSID of the access point
+    ap.config(max_clients=10) # set how many clients can connect to the network
+    ap.active(True)         # activate the interface
```
dpgeorge, line 82: You could also add the line `ap.status("stations")` as an example of how to get a list of connected STAs.

docs/rp2/quickref.rst
```diff
@@ -55,6 +55,60 @@ The :mod:`rp2` module::
...
+
+    def do_connect():
+        import network
+        wlan = network.WLAN(network.WLAN.IF_STA)
+        wlan.active(True)
+        if not wlan.isconnected():
+            print('connecting to network...')
+            wlan.connect('ssid', 'key')
+            while not wlan.isconnected():
+                pass
```
dpgeorge, line 94: Better to use `machine.idle()` here so it doesn't use power unnecessarily while waiting.

mattytrentini:
I'm happy to make those changes but shouldn't we be striving to make the WLAN behaviour the same across all ports (at least ESP32/rp2).

Apart from where they actually differ, of course (like `reconnects`). So, should I update the ESP32 quickref too?

docs/rp2/quickref.rst
mattytrentini: Thanks, will delete.

docs/rp2/quickref.rst
mattytrentini: OK, and will update the ESP32 quickref.

docs/rp2/quickref.rst
mattytrentini: OK, but if so I'll also update the ESP32 quickref.

docs/rp2/quickref.rst
mattytrentini: Good idea - and I'll update the ESP32 quickref too.

docs/rp2/quickref.rst
mattytrentini: Done.

docs/rp2/quickref.rst
mattytrentini: Done.

docs/rp2/quickref.rst
mattytrentini: I didn't end up making this change because I couldn't find any documentation for the option. Seems like that should be in place first?

docs/esp32/quickref.rst
```diff
@@ -83,30 +83,30 @@ The :class:`network.WLAN` class in the :mod:`network` module::
...
-    ap.config(max_clients=10) # set how many clients can connect to the network
-    ap.active(True)         # activate the interface
+    ap.config(ssid='ESP-AP')              # set the SSID of the access point
+    ap.config(max_clients=10)             # set how many clients can connect to the network
+    ap.active(True)                       # activate the interface
 
 A useful function for connecting to your local WiFi network is::
 
     def do_connect():
         import network
```
dpgeorge, line 102: Is it worth making this `import machine, network` so the call to `machine.idle()` below works (and if the user hasn't got `machine` in the global scope)?

mattytrentini: Yes, an oversight on my behalf, fixed, thanks!

docs/rp2/quickref.rst
dpgeorge: Documentation for stations was just added now in 4d034f817c861b90cbe7ab0cb59d85036031750f

Do you want to add the example?  Happy either way.

dpgeorge (approved):

docs/rp2/quickref.rst
dpgeorge: We can add this another time.

PR #16480, opened 2024-12-24
f-string thousands separator for float

### LINKED ISSUE #11284: f-string thousands separator doesn't work with floats
On v1.19.1, f-string thousands separator doesn't work with floats. Tested on both `unix` and `rp2040`:

Demo code:
```
foo_int = 123456
foo_float = 1234.56789
print(f"{foo_int:,}   {foo_float:,}")
```

CPython (correct):
```
123,456   1,234.56789
```

MicroPython (incorrect):
```
123,456   1234.57
```

ludollaoo:
### Summary

Closes #11284 
thousand separator was not implemented for floating point number

### Testing

test the thousand separator for different float
test that the rounding up with the thousand separator (esp :  9,999.99 rounded to 10,000.0)

couldn't test that no buffer overflow occurs when using the thousand separator

### DIFF
--- a/ports/unix/coverage.c
+++ b/ports/unix/coverage.c
@@ -520,17 +520,17 @@ static mp_obj_t extra_coverage(void) {
 
         // format with inadequate buffer size
         char buf[5];
-        mp_format_float(1, buf, sizeof(buf), 'g', 0, '+');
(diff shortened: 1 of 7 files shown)


dpgeorge:
Thanks for the contribution.  Please check the CI results, there are quite a few failures which look related to the changes in this PR.

Issue #10027, opened 2022-11-20
RP2 PIO: access to EXECCTRL_SIDE_PINDIR to use side-set for pin dirs


pdg137:
The RP2040 PIO has a feature called "side-set" that can set the levels or dirs of pins as part of any PIO instruction. However, it appears that only the levels are supported, and there's no way to use side-set to set pin directions.  Here's the relevant quote from the datasheet:

![image](https://user-images.githubusercontent.com/466760/202914986-31b0d7b4-76b6-4eeb-9197-7fb2ba4caf55.png)

It seems like this should be a configuration parameter to [`rp2.StateMachine()`](https://docs.micropython.org/en/latest/library/rp2.StateMachine.html#rp2.StateMachine).

dpgeorge:
Implemented in 0662c551216db1a3ab40dafbeffcc3ee929d7690

PR #16506, opened 2024-12-31
rp2/rp2_pio: Add sideset_pindirs support for PIO.

magy00:
PIO side-setting can also be used to change pin directions instead of pin values.
This adds a parameter `sideset_pindirs` to decorator `asm_pio()` to configure it.

While C SDK 2.1.0 uses the singular for naming (`SIDE_PINDIR`), its `pioasm` directive
actually uses the plural (`.side_set <count> (opt) (pindirs)`), so it would look like:

```
.program test
    .side_set 1 opt pindirs
    .lang_opt python sideset_init = rp2.PIO.IN_LOW
    .lang_opt python sideset_opt = True
    .lang_opt python sideset_pindirs = True
...
```

Also replaces a few close-by 0s with corresponding `PIO.*` constants.

Addresses #10027.

### DIFF
--- a/docs/library/rp2.rst
+++ b/docs/library/rp2.rst
@@ -23,7 +23,7 @@ The ``rp2`` module includes functions for assembling PIO programs.
 
 For running PIO programs, see :class:`rp2.StateMachine`.
 
-.. function:: asm_pio(*, out_init=None, set_init=None, sideset_init=None, in_shiftdir=0, out_shiftdir=0, autopush=False, autopull=False, push_thresh=32, pull_thresh=32, fifo_join=PIO.JOIN_NONE)
+.. function:: asm_pio(*, out_init=None, set_init=None, sideset_init=None, side_pindir=False, in_shiftdir=PIO.SHIFT_LEFT, out_shiftdir=PIO.SHIFT_LEFT, autopush=False, autopull=False, push_thresh=32, pull_thresh=32, fifo_join=PIO.JOIN_NONE)
 
     Assemble a PIO program.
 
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the contribution.

I suggest calling the new keyword argument `side_pindir`.

Rationale: it matches the name used in the datasheet, ie `SIDE_PINDIR`.  The other keyword arguments like `in_shiftdir` and `autopull` also match the datasheet.  (And note that the datasheet uses `SIDESET_COUNT` as well as `SIDE_PINDIR` so it distinguishes between "sideset" and plain "side".)

docs/library/rp2.rst
```diff
@@ -35,8 +35,10 @@ For running PIO programs, see :class:`rp2.StateMachine`.
     - *out_init* configures the pins used for ``out()`` instructions.
     - *set_init* configures the pins used for ``set()`` instructions. There can
       be at most 5.
-    - *sideset_init* configures the pins used side-setting. There can be at
-      most 5.
+    - *sideset_init* configures the pins used for ``.side()`` modifiers. There
+      can be at most 5.
+    - *side_pindir* configures whether ``.side()`` modifiers are used for
+      pin directions instead of pin values.
```
dpgeorge, line 41: It's not clear which values are allowed for this argument.  Suggest something like:
```
*side_pindir* when set to ``True`` configures ``.side()`` modifiers to be used for pin directions, instead of pin values (the default, when ``False``).
```

dpgeorge (approved):

docs/library/rp2.rst
dpgeorge: Looks good now, thanks.

PR #16516, opened 2025-01-02
esp32/boards/LOLIN_C3_MINI/mpconfigboard.h: Enabled i2s.

StrayCatNZ:
### Summary

Enabled i2s for lolin_c3_mini as it works fine now.

### Testing

Tested on esp32_c3_SuperMini and it works great!

### DIFF
--- a/ports/esp32/boards/ESP32_GENERIC_C3/mpconfigboard.h
+++ b/ports/esp32/boards/ESP32_GENERIC_C3/mpconfigboard.h
@@ -4,7 +4,7 @@
 #define MICROPY_HW_MCU_NAME                 "ESP32C3"
 
 #define MICROPY_HW_ENABLE_SDCARD            (0)
-#define MICROPY_PY_MACHINE_I2S              (0)
+#define MICROPY_PY_MACHINE_I2S              (1)
 
 // Enable UART REPL for modules that have an external USB-UART and don't use native USB.
 #define MICROPY_HW_ENABLE_UART_REPL         (1)
--- a/ports/esp32/boards/ESP32_GENERIC_C3/mpconfigboard.h
+++ b/ports/esp32/boards/ESP32_GENERIC_C3/mpconfigboard.h
@@ -4,7 +4,6 @@
 #define MICROPY_HW_MCU_NAME                 "ESP32C3"
 
 #define MICROPY_HW_ENABLE_SDCARD            (0)
-#define MICROPY_PY_MACHINE_I2S              (1)
 
 // Enable UART REPL for modules that have an external USB-UART and don't use native USB.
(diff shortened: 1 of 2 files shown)


jonnor:
This should maybe be on all ESP32C3 devices? I do not see any reason why this feature would be board specific?

StrayCatNZ:
> This should maybe be on all ESP32C3 devices? I do not see any reason why this feature would be board specific?

Most likely yes, I can also change it in generic c3. I only have esp32c3 supermini to test it on though, if you think that will be enough of a test?

jonnor:
If you only have one, that is ok. Hopefully one other person is able to test on another C3! That would be ideal :)

StrayCatNZ:
> If you only have one, that is ok. Hopefully one other person is able to test on another C3! That would be ideal :)

I also flashed ESP32_GENERIC_C3 version into my esp32c3 mini and i2s works with this firmware too. 

So far I'm using my esp32c3 mini with this patch in a project where it plays audio with i2s, controls some leds and communicates with lego hub at the same time and no issues so far, so super happy about that! :)
https://github.com/user-attachments/assets/8462c1c0-13eb-4cad-86ff-dab7c2198bd9

zalves:
Just today I had to build a firmware for my ESP32C3 super mini to use with an INMP441 mic. Works great!

ports/esp32/boards/ESP32_GENERIC_C3/mpconfigboard.h
```diff
@@ -4,7 +4,7 @@
 #define MICROPY_HW_MCU_NAME                 "ESP32C3"
 
 #define MICROPY_HW_ENABLE_SDCARD            (0)
-#define MICROPY_PY_MACHINE_I2S              (0)
+#define MICROPY_PY_MACHINE_I2S              (1)
```
dpgeorge, line 7: Please just remove this line altogether.  Then it should pick up the default config, which is to enable I2S.

ports/esp32/boards/LOLIN_C3_MINI/mpconfigboard.h
```diff
@@ -3,7 +3,7 @@
 #define MICROPY_PY_NETWORK_HOSTNAME_DEFAULT "mpy-c3-mini"
 
 #define MICROPY_HW_ENABLE_SDCARD            (0)
-#define MICROPY_PY_MACHINE_I2S              (0)
+#define MICROPY_PY_MACHINE_I2S              (1)
```
dpgeorge, line 6: As above, just remove this line.

dpgeorge (approved):

dpgeorge:
Squashed, rebased and merged in 9d0a5ac7e9ac8b90d1bbea402fd16a410aa4008f

PR #16469, opened 2024-12-22
esp32: Add the ability to disable ULP module using mpconfigport.h and ULP images built into custom builds

purewack:
esp32: Add the ability to compile custom ULP image into the build by defining ulp_app_name, ulp_sources, and ulp_exp_dep_srcs

### Summary

I think it should be configurable to disable the ULP module, so that experimentation with the RISCV ULP can be done without causing conflicts in the esp-idf build steps, since only one type of ULP can be enabled at a time (for those chips that do have 2 types)

### Testing

I have tested this by writing my own test module which uses the RISCV ULP by firstly disabling the built in ULP module. I have done the test by creating a custom board definition for testing and defining ULP specific build steps in mpconfigboard.cmake as follows:

```cmake
set(ulp_app_name ulp_test)
set(ulp_riscv_sources ${MICROPY_BOARD_DIR}/ulp/main.c)
set(ulp_exp_dep_srcs "${MICROPY_BOARD_DIR}/cmodules/modulptest.c ${MICROPY_BOARD_DIR}/board_init.c")
```

PR #11572, opened 2023-05-20
ports/esp32/ulp: Initial draft PR for RISCV ULP support S2/S3.

ThinkTransit:
The S2 and S3 chips contain a RISCV ULP co-processor in addition to the FSM ULP.

This PR is an initial attempt to provide access to the RISCV co-processor from micropython.

### Operation

1. During build cmake checks for the existence of file **ports/esp32/ulp_riscv/main.c**
2. **main.c** is compiled and included in the micropython firmware bin file
3. Calling ULP.riscv_load_binary() will load binary to ULP co-processor
4. Calling ULP.riscv_run() will run RISCV ULP.

### Compilation

The ULP needs to be enabled in your sdkconfig.board

```
CONFIG_ESP32S3_ULP_COPROC_ENABLED=y
CONFIG_ESP32S3_ULP_COPROC_RISCV=y
CONFIG_ESP32S3_ULP_COPROC_RESERVE_MEM=4096
```

Build micropython as normal.

### Example - Blinky

Tested with IDF v4.4.4, micropython latest

**ports/esp32/ulp_riscv/main.c**

```c
#include <stdio.h>
#include <stdint.h>
#include <stdbool.h>
#include "esp32s3/ulp_riscv.h"
#include "ulp_riscv/ulp_riscv_utils.h"
#include "ulp_riscv/ulp_riscv_gpio.h"

/* this variable will be exported as a public symbol, visible from main CPU: */
bool gpio_level = false;

int main (void)
{
   ulp_riscv_gpio_init(14);
   ulp_riscv_gpio_set_output_mode(14, RTCIO_MODE_OUTPUT);
   ulp_riscv_gpio_output_enable(14);

   while(1) {
   // Turn on GPIO
       ulp_riscv_gpio_output_level(14, 1);
       gpio_level = true;
   // Delay
       ulp_riscv_delay_cycles(1000 * 1000);
   // Turn off GPIO
       ulp_riscv_gpio_output_level(14, 0);
       gpio_level = false;
   // Delay
       ulp_riscv_delay_cycles(1000 * 1000);
   }

    /* ulp_riscv_halt() is called automatically when main exits */
    return 0;
}
```

```python
MicroPython 3229791b6-dirty on 2023-05-20; INSIGHT_S3_GENERIC with ESP32-S3
Type "help()" for more information.
>>> import esp32
>>> ulp = esp32.ULP()
>>> ulp.riscv_load_binary()
>>> ulp.riscv_run()
>>>
>>> import machine
[...]

PR #16521, opened 2025-01-03
esp32: ULP RISCV optional build

purewack:
### Summary

I have updated the ADC and ULP modules to fix an issue with the RISCV adc function causing a conflict of drivers (as mentioned in my PR #16469), preventing my application from working. 
I have also used the work contributed by  #11572 (Thanks @ThinkTransit ) and extended it to suit it for the new way of building the firmware, as that PR is quite outdated now.

## ADC 
I have started by updating the ADC to use the new driver (esp_adc/adc_oneshot.h).
The reason for this is that when using custom c modules, and the need to call the `ulp_adc_init()` function from the esp-idf arises, an error `ADC: CONFLICT! driver_ng is not allowed to be used with the legacy drive` was displayed at runtime, but compilation was still successful.

There are several errata notes about not being able to change the bit-width of the ADCs certain chips.
The only chip that can switch resorption to a lower one is the normal esp32.
esp32 C2 & S3 are stuck at 12 bits, while S2 is at 13 bits.
On the S2, you can change the resolution, but it has no effect on the resolution, rather, it prevents attenuation from working at all!
I have set the resolution to the be maximum possible for each SoC, with the esp32 being the only one not throwing errors when trying to set the bit-width to 9,10,11 or 12 bits using `ADC.width(bits)`

https://github.com/espressif/esp-idf/issues/13128
https://github.com/espressif/esp-idf/blob/release/v5.2/components/soc/esp32s3/include/soc/soc_caps.h
https://docs.espressif.com/projects/esp-chip-errata/en/latest/esp32s2/03-errata-description/index.html
https://docs.espressif.com/projects/esp-idf/en/v4.2/esp32/api-reference/peripherals/adc.html

## ULP
To feel the benefits of the new ADC driver, I have in turn extended  and improved the functionality of the ULP module.
I went with the following strategy:
- Enable RISCV ULP for chips that support it by default (S2 & S3)
- Keep existing ULP functionality by using FSM where RISCV is not available
- Give users the option to recompile the firmware with a preference for FSM where RISCV is enabled by default

It is now also possible to compile in the binary image of the ULP code, by doing the following:
1. Create or modify a board definition and add a `ulp` folder for ULP source code
2. When this is done, edit `mpconfigboard.cmake` and add the following:
> ```cmake
> set(ulp_embedded_sources ${MICROPY_BOARD_DIR}/ulp/ulp_source_name.c)
> # or
> set(ulp_embedded_sources ${MICROPY_BOARD_DIR}/ulp/ulp_source_name.S)
> ```
3. If there are any custom c modules which want to use the generated `ulp_embedded.h` header, add:
> ```cmake
> set(ulp_dependants ${MICROPY_BOARD_DIR}/cmodules/module_name.c)
> ```
4. After compilation, the ULP module will have variables (ulp code variables prefixed with `var_`) exposed as the modules is an automatic dependent.
5. The ULP module will then have an extra function called `run_embedded()` which does what is says on the tin.

Users then can load & run the code using `ULP.run_embedded()` and interact with the code using `ULP.read(address)` and/or `ULP.write(address, value)`.

Users will have to option to compile for FSM preference by adding the lines: 
```cmake
set(SDKCONFIG_DEFAULTS
    ...
    boards/sdkconfig.ulp_fsm
)


set(PREFER_FSM_ULP ON)
```
to their `mpconfigboard.cmake` file.

I believe this is an elegant solution to this problem, as FSM and RISCV components are mutually exclusive in the esp-idf.

I also added missing functions such as 
- `ULP.pause()`
- `ULP.resume()`
for pausing and resuming ULP execution (in reality it only stops the wakeup timer)

## RTC options for ULP

I also ran into a number of issues with how the ULP code expects the main app to set up its IO.
From my testing, FSM has no way to set pins into RTC mode, therefore no outputs can be controlled by the ULP.
RISCV has functions to do that from within the ULP, but FSM does not seem to have it so easy.

I have added the following methods to the ULP class:
- `ULP.rtc_init(pin_number)`
- `ULP.rtc_deinit(pin_number)`
- `ULP.adc_init(unit, channel)`

These functions allow setting up the pins for IO, as well as setting up the ADC correctly for the ULP to sample away with!

## Example usage of embedded app:

ulp/main.c:
```c
#include <stdint.h>
#include "ulp_riscv_gpio.h"
#include "ulp_riscv_utils.h"
#include "ulp_riscv.h"

#define LED_GPIO_NUM 3

unsigned int var_counter = 0;
unsigned int var_count = 1;

void main(){
      if(var_count){
          var_counter++;
      }
      ulp_riscv_gpio_output_level(LED_GPIO_NUM, 1);
      ulp_riscv_gpio_output_level(LED_GPIO_NUM, 0);
}
```

The app is then compiled into the firmware by modifying, for example the GENERIC_S3 board.

main.py:
```python
import time
import esp32

ulp = esp32.ULP() #default wakeup period is 10ms
ulp.rtc_init(3)
ulp.run_embedded()

time.sleep(1)
print( ulp.read(ulp.VAR_COUNTER) ) # prints for example 325
ulp.write(ulp.VAR_COUNT, 0)

time.sleep(1)
print( ulp.read(ulp.VAR_COUNTER) ) # still prints 325, as the counter was stopped in the ulp code
```

## Example usage loading ADC sampler:

ulp/main.c:
```c
#include <stdint.h>
#include "ulp_riscv_adc_ulp_core.h"
#include "ulp_riscv_utils.h"
#include "hal/adc_types.h"

unsigned int var_reading;

void main(){
      var_reading = ulp_riscv_adc_read_channel(ADC_UNIT_1, ADC_CHANNEL_1);
}
```

The app is then compiled into the firmware by modifying, for example the GENERIC_S3 board.

main.py:
```python
import time
import esp32

ulp = esp32.ULP() 
ulp.adc_init(0,1) # ADC1, Channel 1, GPIO 2 on esp32 S3

with open("riscv,bin","rb") as file:
    bin = file.read()
ulp.run(bin)

var_reading = 0x500000dc # sourced from building the ulp firmware using the generated `.ld` file
print( ulp.read(var_reading) ) # print raw adc reading, resolution dependent on chip
```

### Testing

I have gone through a lot of builds and different combinations, all seem to work fine.
✅ esp32 with FSM embedded code
✅ loading test.bin on esp32 FSM (requires RTC pin setting)
✅ running esp32 s3 RISCV embedded code
✅ loading riscv.bin by first compiling embedded code, then with no embedded app on esp32 S2
✅ using the build ULP binary or an S3 and running it on another S3 that was build with no embedded binary

Test result of running a simple pin flip with no delays; FSM in an infinite loop on ESP32:
test2

Test result of running a simple pin flip with delays in an infinite loop; RISCV on ESP32 S3:
test1

I have included in [my example branch](https://github.com/purewack/micropython/tree/esp32_ulp_examples) two modified board definitions which include everything that I have used to test and confirm functionality of the ULP.

- ESP32_GENERIC
- ESP32_GENERIC_S3

Both include their respective ULP embedded code, along side a `ulp_example_fsm` and `ulp_example` module for ease of testing. 
Simply build, flash, and run `import ulp_example`

The test increments a counter variable, reads ADC, and quickly toggles a pin on and off.

### Side-effects
During testing of this ULP code: 
```c
#include <stdint.h>
#include "ulp_riscv_gpio.h"
#include "ulp_riscv_utils.h"
#include "ulp_riscv.h"

// GPIO number for the LED
#define LED_GPIO_NUM 3

// Delay value (adjust for desired blinking speed)
#define DELAY_CYCLES 200000

unsigned int var_counter = 0;
unsigned int var_count = 1;

void main(){
    // Configure the GPIO as an output
    ulp_riscv_gpio_init(LED_GPIO_NUM);
    ulp_riscv_gpio_output_enable(LED_GPIO_NUM);

    while (1) {
        if(var_count){
            var_counter++;
        }
        // Toggle the GPIO
        ulp_riscv_gpio_output_level(LED_GPIO_NUM, 1);
        ulp_riscv_delay_cycles(DELAY_CYCLES);
        ulp_riscv_gpio_output_level(LED_GPIO_NUM, 0);
        ulp_riscv_delay_cycles(DELAY_CYCLES);
    }
}
```
I was not able to reference the `var_counter` and `var_count` variables. They were compiled out as the function could never return. 
As another side effect of an infinite loop, you cannot stop the code by calling `ULP.pause()` because the function never returns to be then woken up by the timer.

I believe the RISCV ULP has matured enough to be useful to a number of users, as it is far simpler code rather than being stuck writing assembly for the FSM.


ThinkTransit:
Nice work @purewack I will close my unfinished PR as I think you have covered everything in mine?

stephanelsmith:
This looks fantastic.  I can't wait to try it out!  Greatly appreciate your pioneering efforts on this front!

dpgeorge:
Thanks for the contribution, this does look very good!

One initial comment/question: it looks like `ULP.load_binary()` is gone, and now `ULP.run()` both loads and starts the ULP program.  This is a breaking change to the API.

Is this a necessary change?  Is there a use-case where the ULP program is loaded just once and then run many times?  And possibly started some time after loading?

Also, are the ADC changes to use the new driver independent to the ULP improvements?  If so, it would be beneficial to move those changes (hopefully just cherry-pick a commit or two from here) to a separate PR, to make it easier to review and faster to merge.

purewack:
Hi @dpgeorge !

Thanks for looking through the PR.

What is the best approach for this now? I'm new to all this as I have only ever worked on my private repos before. 

Should I close this PR and open separate ADC and ULP ones ?

Regarding the ULP breaking change, I can revert the functionality back to separate load and run functions. The reason I merged them into one was I didn't see a use case where a user might load a binary to them later execute it as opposed to just calling load and run in a quick succession, it felt like an extra step to do the same thing. 

I'm am happy to do what is required in order not to break anything, but some pointers would be appreciated 😅

dpgeorge:
> What is the best approach for this now?

The work you've done here looks really good, thank you.  I think it's probably 90% ready to merge.  It's just the last 10% can take some time to go back and forth.

Considering there are ULP API changes which need a bit of discussion/work, it would be best to split out the ADC changes as a separate PR, as I think they can get merged relatively quickly.  Then leave this PR as-is until the ADC is merged, then we'll pick it up again.

purewack:
I have opened PR #16630 and singled out all the ADC changes there @dpgeorge

purewack:
I've included RTC pin functionality in the changes here too. 
I think given what I have learned now, it would be a good idea to extend either the RTC module or the Pin class itself to allow setting up pins in RTC mode so that the ULP can control the outputs / inputs.

The question is which approach is more appropriate: RTC or Pin class?
I think this would be the next step to focus on in terms of breaking up this PR.

purewack:
Ok some time has passed since the last time this was spoken of. I have made all the necessary changes (and corrected my mistakes @dpgeorge) to the ADC for the ULP module to work with riscv on the PR #16630.

I think what I will do is restore the functions reverting any breaking changes I have introduced to the ULP module, but extend the module to both add missing functions like `pause` and `resume` while also adding *embedded* firmware specific functions like they are now. 

Regarding the RTC pin functionality I am unsure whether is is best to keep them here or place them in a different module, but as was mentioned earlier this will require further discussion and some back and forth in this regard as well.

docs/library/esp32.rst
```diff
@@ -300,34 +300,147 @@ Constants
...
-ESP32-S2 and ESP32-S3 chips.
+This class gives access to the Finite State Machine (FSM) Ultra Low Power (ULP)
+co-processor on the ESP32, ESP32-S2 and ESP32-S3 chips. In addition the ESP32-S2
+and ESP32-S3 chips contain a RISCV co-processor which can be accessed through this
+class.
 
-.. warning::
+.. note::
+    By default RISCV is preferred on supported chips, however, you can configure a custom build 
+    of micropython to revert back to using the FSM ULP.
```
projectgus, line 310:
```suggestion
    of MicroPython to revert back to using the FSM ULP.
```

projectgus: In addition to the nitpick about capitalisation, this is something of a breaking change for anyone who was using the `ULP` class on S2 and S3 before, yes? Although I agree the RISC-V one is much nicer to work with, we might have to defer swapping defaults until MP 2.0 (or add a build variant for some boards).

docs/library/esp32.rst
```diff
@@ -300,34 +300,147 @@ Constants
...
+    print(u.write(u.VAR_DO_ADC_SAMPLING, 1))
 
-    This class does not provide access to the RISCV ULP co-processor available
-    on the ESP32-S2 and ESP32-S3 chips.
+.. note::
+   Exposed variables are relative address offsets from RTC_SLOW_MEM address, 
+   
+   however you can use full address values also, if loading ULP code from outside the compilation process.
 
 .. class:: ULP()
```
projectgus, line 371: Even keeping the API (and the implementation) mostly the same, what do you think about renaming the class to `RVULP`  if the build is configured for RISC-V?

Otherwise it's quite confusing to take Python code that works on one board and find it's valid Python on another board, but also doesn't work.

This also makes it easier to document some of the differences (like `ULP.set_wakeup_period`) between the two implementations.

ports/esp32/esp32_ulp.c
```diff
@@ -28,14 +28,33 @@
 
 #if CONFIG_IDF_TARGET_ESP32 || CONFIG_IDF_TARGET_ESP32S2 || CONFIG_IDF_TARGET_ESP32S3
 
+#if CONFIG_IDF_TARGET_ULP_PREFER_FSM || CONFIG_IDF_TARGET_ESP32
+    #define TYPE_RISCV 0
```
projectgus, line 32: Nitpick, we don't indent inside #if blocks in MicroPython.

ports/esp32/esp32_ulp_qstr.py
```diff
@@ -0,0 +1,78 @@
+"""
+This script reads in the given CMSIS device include file (eg stm32f405xx.h),
+extracts relevant peripheral constants, and creates qstrs, mpz's and constants
+for the stm module.
```
projectgus, line 4: This docstring (and the block comment below) need updating for this context.

projectgus (commented):
This is a nice clean implementation and I think it'll be really handy to be able to write small RISC-V programs and embed them in MicroPython.

I have been thinking a lot about supporting custom builds with an embedded RISC-V program as the "best" option for users, mostly the trade-off of functionality to complexity. The more "MicroPython-y" way would be to load the ULP binary at runtime. IIUC the trade-offs are:

* There is a temporary RAM cost to loading the binary from the filesystem. That'll be true until "ROMFS" is fully supported, but also the binary is small.
* Generating the `var_` variable entries is a really nice bit of functionality, and it'd be more work to bring this over to dynamically loaded ULP programs. Although you could imagine a relatively simple tool that takes a ULP .elf file as input and outputs some kind of .mpy file that embeds the variable constants and the binary code...

To be clear, I'm still OK with merging support for embedding the RISC-V ULP binaries and you should keep it in this PR for now. I am curious what you think about the good/bad aspects of an alternative approach, though?

docs/library/esp32.rst
projectgus: It's kind of a shame that ESP-IDF doesn't support enabling both ULP types... I'm guessing you looked into this and there's no way to make it work?

purewack: Unfortunately, yes I have looked deep into it and it looks like they are mutually exclusive so we only get to pick one or the other at build time :/ Maybe in the future IDF releases there will be an option for runtime selection as I don't know if this is a hardware of software limitation of the ESP32.

docs/library/esp32.rst
purewack: I agree and this was going through my head after implementing the changes. This change would be very breaking if the RISCV preference was a default on future builds. I agree the RVULP class is a great idea!

purewack:
I have started a thread regarding this PR: https://github.com/orgs/micropython/discussions/16819#discussioncomment-12380769 to try and get more opinions on the breaking changes but now that @projectgus has suggested the RVULP class, I think it is the solution for this.

purewack:
I've been doing a lot of work on my own board running micropython (esp32s3) and have learned a great deal about how things work under the hood. I've been using the ULP changes I have made, and noticed that sometimes my ULP code will not start (still trying to determine why, happens when imported using a wrapper module that actually interfaces with the ULP using VAR_*). 

I needed to first pause then resume the timer (pause halts the program too for RISCV) to get the binary to work after importing my wrapper module in a different script for example, as I have the wrapper frozen into the firmware.

**_Main question:_**
From my understanding the ULP is a singleton object. 
Would it benefit from a finaliser to stop the running ULP code upon GC or explicit __del__ ? 

I think the code should not run, especially after soft-reset, which it does, and always did even before the changes I had made.

This can be easily confirmed by writing a counter program for the ULP, running that binary then doing a soft-reset, and finally reading the counter value from RTC memory, observing that it has not reset, and is still increasing.

```python
import esp32
u = esp32.ULP()
u.run_embedded() # or any other way using open() and ulp.load() etc
u.read(u.VAR_COUNTER)

# soft reset here
import esp32
u = esp32.ULP
u.read(u.VAR_COUNTER)
```

projectgus:
> From my understanding the ULP is a singleton object.
> Would it benefit from a finaliser to stop the running ULP code upon GC or explicit **del** ?

Sorry, I missed this question when you asked it!

The most common pattern for these "hardware-adjacent" singletons is to add a deinit() function in the soft reset path of the main function (look for label `soft_reset_exit:` in `ports/esp32/main.c`). This is more deterministic than a finaliser, and has less overhead when you know you're going to clean up at most one of something.

However, there are some low level board functions (like the network connection) that remain running over a soft reset. I feel like conceptually ULP is closer to this, because you can even (for example) go into deep sleep and have the ULP keep running.

So maybe the cleanup code won't need to stop the ULP program running, but should still clean-up any Python-side state about the ULP so that it stays "safe" after the reset cycle when the code can re-initialise it and determine the current state? I'm not sure if that idea makes sense, let me know if it doesn't.

ports/esp32/esp32_ulp.c
```diff
@@ -28,14 +28,33 @@
...
 #elif CONFIG_IDF_TARGET_ESP32S2
-#include "esp32s2/ulp.h"
+    #include "esp32s2/ulp.h"
 #elif CONFIG_IDF_TARGET_ESP32S3
-#include "esp32s3/ulp.h"
+    #include "esp32s3/ulp.h"
 #endif
 
+#ifdef ULP_EMBEDDED_APP
+    #include "ulp_embedded.h"
```
adrian-aaen, line 48: I wanted to test out this PR for a personal project of mine, and after having followed the guide and examples to make a modified board definition, the compiling fails. It crashes with the following error:
```
ports/esp32/esp32_ulp.c:48:14: fatal error: ulp_embedded.h: No such file or directory
```
How can I make this work?

purewack: Hi I just realised that I must have made a mistake somewhere along the way when testing and somehow my build was using cached files when switching between the different ULP types. 

I will post a fix soon when I find out why it refuses to work on the first build. You may want to try building again despite the error.

Sorry about this!

stephanelsmith:
@purewack hope you are well!  Finally getting time to try this one out (some vacation relax time!)  I've added a pull request to update your branch with the micropython latest master, test with esp-idf v5.4.  I did find one change required to get build running.
https://github.com/purewack/micropython/pull/1

Still unsure why it takes two builds for it to work.  First build fails with:
```
No install step for 'ulp_embedded'
```
Checking the stdout:
```
make[3]: *** No rule to make target 'esp-idf/main/ulp_embedded/ulp_embedded.ld', needed by 'esp32_ulpconst_qstr.h'.  Stop.
```
Trying again with no changes passes.

Counter tested on UM TINYS3, I don't have a scope, so not able to test the ADC and GPIO.

purewack:
Nice to hear from you! @stephanelsmith 

I did see your fork so that encouraged me to work on this again.

I will soon be tackling this and trying to place all the RISCV stuff in its own class ULP_RV (or similar) as suggested by some pr comments and other posts. 

As to why the compilation problems persist, it may have to do with the fact I wrote this using IDF v5.2.x so I did some work updating the adc driver with IDF v5.4 now (#16630). I'm not sure yet if this is the reason but regardless I need to rebase anyway and 5.4 seems to be the desired IDF version now.

stephanelsmith:
> Nice to hear from you! @stephanelsmith
> 
> I did see your fork so that encouraged me to work on this again.
> 
> I will soon be tackling this and trying to place all the RISCV stuff in its own class ULP_RV (or similar) as suggested by some pr comments and other posts.
> 
> As to why the compilation problems persist, it may have to do with the fact I wrote this using IDF v5.2.x so I did some work updating the adc driver with IDF v5.4 now (#16630). I'm not sure yet if this is the reason but regardless I need to rebase anyway and 5.4 seems to be the desired IDF version now.

Rock on, the build process is still a bit novel for me, but happy to help where I can!  I'll keep playing around with it!

stephanelsmith:
> Nice to hear from you! @stephanelsmith 
> 
> I did see your fork so that encouraged me to work on this again.
> 
> I will soon be tackling this and trying to place all the RISCV stuff in its own class ULP_RV (or similar) as suggested by some pr comments and other posts. 
> 
> As to why the compilation problems persist, it may have to do with the fact I wrote this using IDF v5.2.x so I did some work updating the adc driver with IDF v5.4 now (#16630). I'm not sure yet if this is the reason but regardless I need to rebase anyway and 5.4 seems to be the desired IDF version now. 


Another question. I'm getting into energy harvest application and considering esp32c6 (s3 is my current goto).  It too looks to have a riscv ulp core!

https://docs.espressif.com/projects/esp-idf/en/stable/esp32c6/api-reference/system/ulp-lp-core.html

In terms of making a ULP_RV class, would compatibility seem plausible? I haven't had a moment to play with it, but possibility a consideration for the class.

dpgeorge:
Please rebase this PR now that #16630 is merged.

purewack:
I have ran into a problem when I was testing my patch. I created a simple ULP program to sample the ADC. 
The FSM variant works fine across multiple IDF versions, RISCV however hangs indefinitely on the recommended v5.4.1 version.

I have narrowed down the cause of the bug to be the IDF version which MicroPython suggests to compile for: v5.4.1

**On the IDF v5.4.1 this hyper-specific bug appears when the user code tries to call the `ulp_riscv_adc_read_channel(unit,channel)` function.**

I have compiled this bare-bones app to isolate the bug from MicroPython.

# RISCV variant test case

main.c:
```c
#include <stdio.h>
#include "esp_sleep.h"
#include "soc/rtc_cntl_reg.h"
#include "soc/sens_reg.h"
#include "soc/rtc_periph.h"
#include "driver/gpio.h"
#include "driver/rtc_io.h"
#include "ulp_riscv.h"
#include "ulp_adc.h"
#include "ulp_main.h"
#include "freertos/FreeRTOS.h"
#include "freertos/task.h"

extern const uint8_t ulp_main_bin_start[] asm("_binary_ulp_main_bin_start");
extern const uint8_t ulp_main_bin_end[]   asm("_binary_ulp_main_bin_end");

static void init_ulp_program(void);

void app_main(void)
{
    
    vTaskDelay(pdMS_TO_TICKS(1000));

    init_ulp_program();
    printf("ULP %p \n",(void*)ulp_main_bin_start);
    while(1){
    	printf("reading ULP (%ld) %ld - %ld\n",
         (ulp_var_vbat_raw & 0xffff),
         (ulp_var_counter & 0xffff),
         (ulp_var_counter2 & 0xffff)
      );
    	vTaskDelay(pdMS_TO_TICKS(200));
    }
}

static void init_ulp_program(void)
{
	esp_err_t err = 0;

    ulp_riscv_halt();
    ulp_riscv_timer_stop();
    ulp_riscv_reset();

   ulp_adc_cfg_t cfg = {
        .adc_n = ADC_UNIT_1,
        .channel = ADC_CHANNEL_0,
        .width = ADC_BITWIDTH_DEFAULT,
        .atten = ADC_ATTEN_DB_12,
        .ulp_mode = ADC_ULP_MODE_RISCV,
    };
    err = ulp_adc_init(&cfg);
    ESP_ERROR_CHECK(err);

    err = ulp_riscv_load_binary(ulp_main_bin_start, (ulp_main_bin_end - ulp_main_bin_start));
    ESP_ERROR_CHECK(err);

 	ulp_set_wakeup_period(0,1000 * 100);

    err = ulp_riscv_run();
    ESP_ERROR_CHECK(err);

    ulp_riscv_timer_resume();
}

```

ulp.c
```c
#include <stdio.h>
#include <stdint.h>
#include <stdbool.h>
#include "ulp_riscv.h"
#include "ulp_riscv_utils.h"
#include "ulp_riscv_gpio.h"
#include "ulp_riscv_adc_ulp_core.h"
#include "hal/adc_types.h"

unsigned int var_vbat_raw;
unsigned int var_toggle = 1;
unsigned int var_counter = 0;
unsigned int var_counter2 = 0;

int main (void)
{      
    var_counter2 += 1;

    var_vbat_raw = ulp_riscv_adc_read_channel(ADC_UNIT_1, ADC_CHANNEL_0);
    
    if(var_toggle){
        ulp_riscv_gpio_output_enable(13);
        ulp_riscv_gpio_output_level(13, 1);
        ulp_riscv_gpio_output_level(13, 0);
    }

    var_counter += 1;

    return 0;
}
```
 
# RISCV test results

Built and tested against these idf versions: 
```
REPOSITORY      TAG            IMAGE ID       CREATED        SIZE
espressif/idf   release-v5.4   6765a5376c4a   2 days ago     7.51GB
espressif/idf   release-v5.3   2cebfb4176a0   2 days ago     5.58GB
espressif/idf   release-v5.2   7183a8985d32   2 days ago     5.53GB
espressif/idf   release-v5.5   f83d2ef7987b   2 days ago     7.56GB
espressif/idf   v5.4.2         83dcdbc6e72d   2 days ago     7.5GB
espressif/idf   v5.4.1         fcd0c0acc420   2 days ago     7.26GB
```

| version | ADC sample ok? |
| ---- | ----- |
| v5.4.1 | hanging |
| v5.4.2 | ok |
| release-v5.5 | ok |
| release-v5.4 | ok |
| release-v5.3 | ok |
| release-v5.2 | ok |

I attach the log results for the important versions v5.4.1, v5.4.2, and release 5.4

[esp32s3_ulp_adc_log.md](https://github.com/user-attachments/files/21092083/esp32s3_ulp_adc_log.md)

In the logs, it can be seen that the counters which are before and after the adc function, only get incremented before the adc function on v5.4.1 
:/

The easiest option would be to compile for v5.4.2, but I'm not sure if MicroPython is written for it yet.
We could advise against using the ADC entirely, however that would defeat the purpose of the PR #16630 where the ADC driver was updated to allow exactly this kind of operation.

dpgeorge:
> I have compiled this bare-bones app to isolate the bug from MicroPython.

Is there an upstream issue report in IDF that was for this bug, or at least could you find the commit that fixed it?

> The easiest option would be to compile for v5.4.2, but I'm not sure if MicroPython is written for it yet.

I think MicroPython can be built with v5.4.2, see eg #17586.  So maybe updating the official version to v5.4.2 is the right way forward here?

purewack:
I was only able to find this PR which is a similar issue however I cant locate the exact commit that has fixed the issue: https://github.com/espressif/esp-idf/issues/15927

dpgeorge:
MicroPython now targets IDF 5.4.2, since 10ef3e4ac2767b68f746f0ebb995672076eee67d

stephanelsmith:
How is this PR looking, any help on this one?

purewack:
> How is this PR looking, any help on this one?

Hi sorry for the massive downtime :/ I didn't have the time for it, but after reviewing my changes and going down a rabbit hole, I have determined the following:

- Enabling RISCV ULP by default is a bad idea as it would break compatibility with existing unknown number of users' code
- embedding ULP code into the MicroPython binary is less flexible than already existing options and is far more problematic

Therefor I have modified my original idea and made the following changes:
- allow users to build custom MicroPython firmware with the option to enable RISCV type ULP, thus changing the class esp32.ULP to esp32.ULP_RV to avoid confusion. This is done by specifying BOARD_VARIANT=RV during build or modifying board definitions.
- Loading external binaries is easier than embedding them using esp-idf `ulp_embed_binary()` cmake function, which was my initial idea of allowing users to pre-bake the binary into MicroPython.

As for building the RISCV binary itself, I have yet to come up with a build script which uses esp-idf compilers to build the binary file, so for now the easiest option is to just build a bare-bones esp-idf project based on the ULP examples they have and extract the binary from the build files and reference the symbols for your own variables that way.

Also I didn't realize that MicroPython has a built in way to read memory, so I have removed memory manipulation functions which I have added along the way and instead users should use machine.mem32 to interface with the ULP (something like this [example](https://github.com/micropython/micropython-esp32-ulp/blob/master/examples/counter.py) for the FSM type).

This approach eliminates errors with missing auto generated ULP header files which I have had since I started working on this, it also eliminates Docker build problems which I ran into as well.

I will try and come up with an easy way to compile RISCV ULP code.

purewack:
I have created 2 examples which should run without errors and are standalone, meaning the binary ULP code is in the test file, which you can run with mpremote or similar. 

As I've stated, I have not found a more streamlined way of getting ULP binaries other than through the use of a separate minimal esp-idf project.

[ulp.riscv.test.py](https://github.com/user-attachments/files/24806256/ulp.riscv.test.py)
[ulp.fsm.test.py](https://github.com/user-attachments/files/24806257/ulp.fsm.test.py)

For the RISCV test, you need to `make BOARD=ESP32_GENERIC_S3 BOARD_VARIANT=RV` to enable RISCV ULP.

purewack:
@stephanelsmith You were asking about progress and your interest in the RISCV ULP before. 

For anyone who wants to test this PR, I created a repo to compile ULP code into a python script using a bare-bones esp-idf project (This could be refined later, but works for now), this will allow you to write ULP c code and try it out with this fork at ease, with mpremote or other means :)

[ULP compiler](https://github.com/purewack/micropython-ulp-compiler).

I also put in an example program in there to demonstrate the output.

docs/library/esp32.rst
```diff
@@ -557,10 +579,64 @@ ESP32-S2 and ESP32-S3 chips.
...
+    
+    Removes ULP exclusive access to the pin and restores default ADC behavior.
+
+    Throws a ``ValueError`` if you try to de-initialize an already de-initialized ADC block.
+
+.. warning::
+    When compiled for RISCV ULP, the default ``ULP`` class is renamed to ``ULP_RV``
+    and changes the signature of these functions
+
+.. class:: ULP_RV()
```
projectgus, line 623:
```suggestion
.. class:: ULP_RV()

    This class is only enabled if MicroPython is built with ``BOARD_VARIANT=RV`` (see above),
    and replaces the ``ULP`` class. It has all of the same methods as the ``ULP`` class, plus
    the following:
```

ports/esp32/esp32_ulp.c
```diff
@@ -84,21 +128,122 @@ static mp_obj_t esp32_ulp_run(mp_obj_t self_in, mp_obj_t entry_point_in) {
     if (_errno != ESP_OK) {
         mp_raise_OSError(_errno);
     }
+    ulp_timer_resume();
+
     return mp_const_none;
 }
 static MP_DEFINE_CONST_FUN_OBJ_2(esp32_ulp_run_obj, esp32_ulp_run);
+#endif
```
projectgus, line 136:
```suggestion
#endif // ULP_TYPE_RISCV
```

docs/library/esp32.rst
```diff
@@ -536,11 +536,33 @@ Ultra-Low-Power co-processor
...
-
+For hardware specifics and differences between different ULP types, see `ESP-IDF ULP documentation.
+<https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/ulp.html>`_.
+
+.. note::
+
+    In the default scenario, the FSM ULP type is used by default
+    To use RISCV ULP, users will need to compile custom versions of MicroPython with ::
+    
+        make BOARD=ESP32_GENERIC_S3 BOARD_VARIANT=RV
```
projectgus, line 547: I think `BOARD_VARIANT=RV` may be too ambiguous as Espressif has a number of MCUs where the main MCU is RISC-V (although now S2 and S3, of course!) How about `BOARD_VARIANT=ULP_RV`?

ports/esp32/modmachine.c
```diff
@@ -170,6 +171,15 @@ static void machine_sleep_helper(wake_type_t wake_type, size_t n_args, const mp_
...
     #if SOC_ULP_SUPPORTED
     if (machine_rtc_config.wake_on_ulp) {
+        #if CONFIG_IDF_TARGET_ESP32
+        /* Disconnect GPIO12 and GPIO15 to remove current drain through
+        * pullup/pulldown resistors on modules which have these (e.g. ESP32-WROVER)
+        * GPIO12 may be pulled high to select flash voltage.
+        */
+        rtc_gpio_isolate(GPIO_NUM_12);
+        rtc_gpio_isolate(GPIO_NUM_15);
+        #endif // CONFIG_IDF_TARGET_ESP32
```
projectgus, line 181: This looks like a potentially useful fix, but please extract it to a separate PR so we can discuss it separately.

projectgus (commented):
Hi @purewack,

Thanks for keeping this up to date, and sorry it's taken me so long to get back to it.

I have a few low-level comments below, but before doing too much work we need to consider whether we'll commit to supporting a different board variant for this. It's always a lot of work to add board variants, and hard to test. It's unfortunately that ESP-IDF won't allow building a binary that supports both types of ULP!

I think it probably is worth it, as the RISC-V ULP is quite powerful - even though not many people will probably do the work of building a separate ULP C binary to load. @dpgeorge what do you think?

dpgeorge:
> but before doing too much work we need to consider whether we'll commit to supporting a different board variant for this.

Yes, I think we should have board variants for this ULP RISC-V feature.

Since it's available only on ESP32-S2 and ESP32-S3, I suggest only adding ULP_RV variants for ESP32_GENERIC_S2 and ESP32_GENERIC_S3 board profiles.  Other S2/S3 boards will need to choose one of the FSM or RISC-V ULP configurations.

@purewack are you still interested in working on this PR?  It's a nice feature to have.

PR #7123, opened 2021-04-15
STM32L4: Support all STM32L4 MCUs with all features. Add NUCLEO boards.

chrismas9:
This corrects errors and omissions in the STM32L4 ports.
In particular it adds:

- USB for STM32L45x family.
- STM32L433/443 support.
- An easier way to fine tune memory allocation per board which is
    necessary on 256KB parts.
- Adds all series members instead of just one, eg L451, L452, L462
    instead of just L452.
- NUCLEO_L432KC with 28KB LFS1 file system.
- NUCLEO_L433-P board.
- NUCLEO_L433-P board with 28KB LFS1 file system.
- NUCLEO_L433-P board with USB and 16KB LFS1 file system.
    Requires custom USB connection.
- NUCLEO_L452-P board.

Note: The -P boards have a low power mode using on board SMPS.
      Low power mode is not supported. MCU runs from 3.3V.

PR #16523, opened 2025-01-04
stm32: Add support for all STM32L43x/L44x/L45x/L462/L47x/L486/L4A6.

chrismas9:
### Summary

This adds support for all 80 MHz STM32L4 parts.
There are no code changes, only preprocessor directives, af, ld and mk file changes. 
L431 and L433 are higher pincount versions of L432 with more peripherals.
All other variants are identical to existing parts except for USB, LCD and crypto.
The 256 kiB parts are short on Flash for File System and the FS size needs tuning depending on
enabled features (eg USB requires 14 KiB). The stm32l43x*.ld linker scripts simplify FS_SIZE tuning.

### Testing

The L431, L432, L433,  L442 and L443 have been tested on NUCLEO boards with and without USB. Only REPL, USB and FS have been tested. Other peripherals are the same as similar parts.
All parts have been compiled error free using modified board files from existing ports.
After these tests I consider the changes low risk.
I have been using many of these MCUs, including L433, commercially for 4 years based on https://github.com/micropython/micropython/pull/7123


chrismas9:
This is the first of three PRs to replace https://github.com/micropython/micropython/pull/7123
The next will add support for 120 MHz STM32L4 parts - L4P, L4Q, L4R and L4S.
The last one will add ST Microelectronics development board support for some of the new parts.

ports/stm32/mphalport.c
```diff
@@ -90,7 +90,8 @@ void mp_hal_ticks_cpu_enable(void) {
 #endif
 
 void mp_hal_gpio_clock_enable(GPIO_TypeDef *gpio) {
-    #if defined(STM32L476xx) || defined(STM32L496xx)
+    #if defined(STM32L471xx) || defined(STM32L475xx) || defined(STM32L476xx) || defined(STM32L486xx) || \
+    defined(STM32L496xx) || defined(STM32L4A6xx)
```
projectgus, line 94: There are too many datasheets to read them all, but it seems like all of the STM32L4s that have a GPIOG also have a VDDIO2:

```
❯ rg -c --include-zero "define PWR_CR2_IOSV" | sort
stm32l412xx.h:0
stm32l422xx.h:0
stm32l431xx.h:0
stm32l432xx.h:0
stm32l433xx.h:0
stm32l442xx.h:0
stm32l443xx.h:0
stm32l451xx.h:0
stm32l452xx.h:0
stm32l462xx.h:0
stm32l471xx.h:3
stm32l475xx.h:3
stm32l476xx.h:3
stm32l485xx.h:3
stm32l486xx.h:3
stm32l496xx.h:3
stm32l4a6xx.h:3
stm32l4p5xx.h:3
stm32l4q5xx.h:3
stm32l4r5xx.h:3
stm32l4r7xx.h:3
stm32l4r9xx.h:3
stm32l4s5xx.h:3
stm32l4s7xx.h:3
stm32l4s9xx.h:3
stm32l4xx.h:0
system_stm32l4xx.h:0

❯ rg -c --include-zero "define GPIOG" | sort
stm32l412xx.h:0
stm32l422xx.h:0
stm32l431xx.h:0
stm32l432xx.h:0
stm32l433xx.h:0
stm32l442xx.h:0
stm32l443xx.h:0
stm32l451xx.h:0
stm32l452xx.h:0
stm32l462xx.h:0
stm32l471xx.h:2
stm32l475xx.h:2
stm32l476xx.h:2
stm32l485xx.h:2
stm32l486xx.h:2
stm32l496xx.h:2
stm32l4a6xx.h:2
stm32l4p5xx.h:2
stm32l4q5xx.h:2
stm32l4r5xx.h:2
stm32l4r7xx.h:2
stm32l4r9xx.h:2
stm32l4s5xx.h:2
stm32l4s7xx.h:2
stm32l4s9xx.h:2
stm32l4xx.h:0
system_stm32l4xx.h:0
```

... Do you think it's safe to make this `#if defined(STM32L4) && defined(GPIOG)` ?

ports/stm32/adc.c
```diff
@@ -173,10 +173,13 @@
...
+#elif defined(STM32L431xx) || \
+    defined(STM32L432xx) || defined(STM32L433xx) || \
+    defined(STM32L442xx) || defined(STM32L443xx) || \
     defined(STM32L451xx) || defined(STM32L452xx) || \
-    defined(STM32L462xx) || defined(STM32L475xx) || \
-    defined(STM32L476xx) || defined(STM32L496xx) || \
+    defined(STM32L462xx) || defined(STM32L471xx) || \
+    defined(STM32L475xx) || defined(STM32L476xx) || \
+    defined(STM32L486xx) || defined(STM32L496xx) || \
     defined(STM32L4A6xx) || \
```
projectgus, line 183: There reference manual confirms this is true for all STM32L4s, so suggest collapsing it to be `defined(STM32L4)`

image

projectgus (commented):
Thanks for submitting this PR, @chrismas9!

I have a few comments, all about trying to avoid the "rightward creep" of needing to name the *many subtly different* variants of STM32L4! Sounds like you've used the STM32L4 family a lot, and I haven't, so please let me know if you think any of the proposed changes are too risky. (It's probably worth also waiting to see what @dpgeorge thinks about going in that direction before we commit to it.)

ports/stm32/dac.c
```diff
@@ -133,7 +133,7 @@ static void dac_deinit(uint32_t dac_channel) {
 
 void dac_deinit_all(void) {
     dac_deinit(DAC_CHANNEL_1);
-    #if !defined(STM32L452xx)
+    #if !defined(STM32L451xx) && !defined(STM32L452xx) && !defined(STM32L462xx)
```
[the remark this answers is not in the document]
projectgus: Hmm, looking at CMSIS some more this macro might be inconsistently defined. :roll_eyes: 

I think a safe (but ugly) variant will be `#if !defined(STM32L4) || !defined(DAC_CHANNEL2_SUPPORT))`

chrismas9:
@projectgus Thanks for the review and helpful comments.

I agree using STM32L4 in many places would simplify things and improve readability. I did that in my previous PR in some places. I looked for precedent from other commits and thought listing all parts was preferred . However with so many parts it is messy. When I first started I thought there might be newer parts coming with different features, but now I think that is unlikely. ST have moved on with newer families like G4 and I think L4 is stable. I will review your suggestions when I get a chance.

I also wanted to be careful not to touch anything that affects other families due to the amount of testing, especially for families I am not familiar with and don't have dev boards for. I will still consider your suggestions.

I'm happy to update all the linker scripts for parts with 2K sectors. It's a bit more complicated for parts with non uniform sectors sizes. I have another PR coming for them. It reduces the FS cache to 48K, moves the stack into CCRAM, increases the heap size and the FS by using more sectors. I would do the non uniform linker scripts in that PR. I have been using this for years, but it needs a bit of cleanup and a big rebase first.

projectgus:
Sounds good. Feel free to ignore any/all of the suggestions which might impact other chips, I will make a note to take a look at them later on. (I think I can write a script that builds all of the stm32 boards and verifies no differences in the binaries, and then get stuck in cleaning up macros without fear...)

> I'm happy to update all the linker scripts for parts with 2K sectors. It's a bit more complicated for parts with non uniform sectors sizes. I have another PR coming for them. It reduces the FS cache to 48K, moves the stack into CCRAM, increases the heap size and the FS by using more sectors. I would do the non uniform linker scripts in that PR. I have been using this for years, but it needs a bit of cleanup and a big rebase first.

That sounds great to me, if you're up for it!

projectgus:
>  thought listing all parts was preferred 

On this, @dpgeorge may disagree, but I think they've gradually grown from just a few chips supported to a lot now without it necessarily being the intention.

We've had good luck cleaning up the esp32 port to use capability-based macros rather than listing every chip, but that's an order of magnitude fewer chips with a lot less differences so I'm not sure how viable it will turn out to be on stm32.

ports/stm32/usbd_conf.c
```diff
@@ -93,7 +93,8 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd) {
         const uint32_t otg_alt = GPIO_AF10_OTG1_FS;
         #elif defined(STM32L0)
         const uint32_t otg_alt = GPIO_AF0_USB;
-        #elif defined(STM32L432xx) || defined(STM32L452xx)
+        #elif defined(STM32L432xx) || defined(STM32L442xx) || defined(STM32L433xx) || defined(STM32L443xx) || \
+        defined(STM32L452xx) || defined(STM32L462xx)
```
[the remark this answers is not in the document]
chrismas9: The STM32 HAL uses `GPIO_AF0_USB`, `GPIO_AF10_USB_FS`, `GPIO_AF10_USB`, `GPIO_AF10_OTG_FS`, `GPIO_AF10_OTG1_FS`, etc depending on the capabilities of each chip and which AF the USB is allocated to. The L432/L452 names differ to all other L4s because they don't support OTG. I think these need to stay as they are.

As for using `otg_alt `the pyboard hardware supports OTG, it has an OTG connector and provision for a power switch. I'm not sure what software support there is.

ports/stm32/mpconfigboard_common.h
```diff
@@ -629,7 +629,9 @@
 // Whether the USB peripheral is device-only, or multiple OTG
 // For STM32G0 and STM32H5 the USB peripheral supports device and host mode,
 // but otherwise acts like a non-multi-OTG peripheral.
-#if defined(STM32G0) || defined(STM32G4) || defined(STM32H5) || defined(STM32L0) || defined(STM32L1) || defined(STM32L432xx) || defined(STM32L452xx) || defined(STM32WB)
+#if defined(STM32G0) || defined(STM32G4) || defined(STM32H5) || defined(STM32L0) || defined(STM32L1) || \
+    defined(STM32L432xx) || defined(STM32L442xx) || defined(STM32L433xx) || defined(STM32L443xx) || \
+    defined(STM32L452xx) || defined(STM32L462xx) || defined(STM32WB)
```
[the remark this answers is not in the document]
chrismas9: The L332 and L452 families use `USB_FS`, so using `USB_FS `and `USB_OTG_FS `I can greatly simplify this.

ports/stm32/mphalport.c
chrismas9: I have checked one datasheet from each family (L432/33/5/7/9/P/R) and this look safe. Only L47x and above have `GPIOG `and they also have `VDDIO2`

ports/stm32/adc.c
chrismas9: That's what I did in https://github.com/micropython/micropython/pull/7123

chrismas9:
Another possible way to simplify the long list of parts would be to define family sub groups in `mpconfigboard_common.h`, eg:

```
#if defined(STM32L431xx) || \
    defined(STM32L432xx) || defined(STM32L433xx) || \
    defined(STM32L442xx) || defined(STM32L443xx)
#define STM32L43
#endif
```

There are only six sub groups of STM32L4 compared with over twenty individual parts. The main differences between parts in a sub group are USB, LCD and crypto support.

ports/stm32/boards/stm32l43x_12kFS.ld
```diff
@@ -1,12 +1,19 @@
...
-/* Specify the memory areas */
+/* Specify the memory areas.
+   FS_SIZE is not a region, it is only used to define the File System length.
+*/
 MEMORY
 {
-    FLASH (rx)      : ORIGIN = 0x08000000, LENGTH = 230K /* sectors 0-114 */
-    FLASH_FS (r)    : ORIGIN = 0x08039800, LENGTH = 26K /* sectors 115-127 */
+    ALL_FLASH (rx)  : ORIGIN = 0x08000000,  LENGTH = 256K
+    FS_SIZE (r)     : ORIGIN = 0,           LENGTH = 12K /* Adjust the File System size here */
```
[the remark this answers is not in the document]
dpgeorge: I think you can just use variables in ORIGIN and LENGTH?  See how mimxrt does it, eg `ports/mimxrt/boards/common.ld`.  @robert-hh has done a good job making that port quite configurable with the linker scripts.

chrismas9:
@dpgeorge I looked at how the mimxrt  port set MEMORY sizes and moved the MCU flash size and filesystem size to mpconfigboard.mk, passing the values to the linker script. This has several advantages:

1. Only requires one linker script per device type irrespective of flash size.
2. File system size can be easily tuned when options are enabled in mpconfigboard.h without a proliferation of .ld files. Important for 256kiB parts. For example enabling CANbus requires FS to be resiszd by 6K.
3. A board definition can be used with different MCU flash sizes by overriding the flash size on the make command line.

For example `make BOARD=my_board MICROPY_HW_MCU_FLASH_SIZE=1M MICROPY_HW_FS_SIZE=512K`

I have only implemented this for NUCLEO_L432KC, but will do it for all L4 boards if you approve. We would need to keep the old linker scripts to avoid breaking third party projects. I can also do the other STM families if you want.

@projectgus I looked through all the CMSIS definitions and found a way to identify all the parts without multi-OTG USB as you suggested.

ports/stm32/boards/NUCLEO_L432KC/mpconfigboard.mk
```diff
@@ -1,12 +1,20 @@
...
 CMSIS_MCU = STM32L432xx
 AF_FILE = boards/stm32l432_af.csv
-LD_FILES = boards/stm32l432.ld boards/common_basic.ld
+LD_FILES = boards/stm32l43x.ld boards/common_basic.ld
+
+# Specify MCU flash memory size and file system size here.
+MICROPY_HW_MCU_FLASH_SIZE ?= 256K
+MICROPY_HW_FS_SIZE ?= 26K
+LDFLAGS += -Wl,--defsym=MICROPY_HW_MCU_FLASH_SIZE=$(MICROPY_HW_MCU_FLASH_SIZE) \
+           -Wl,--defsym=MICROPY_HW_FS_SIZE=$(MICROPY_HW_FS_SIZE)
```
andrewleech, line 10: I really like this pattern of being able to define filesystem size / linker variables from a board folder, this could be useful on a lot of ports!

dpgeorge: Yes, this looks like a good pattern.

Other ports use lowercase names for linker symbols, but I guess it makes sense to just match the C-level name, ie use the uppercase like you've done here.

Maybe rename `MICROPY_HW_FS_SIZE` to `MICROPY_HW_MCU_FS_SIZE` to indicate that the filesystem is internal within the MCU flash?  Because some boards have external flash with a filesystem, and that's a whole different set of configuration.

ports/stm32/boards/stm32l43x.ld
```diff
@@ -1,12 +1,15 @@
...
 */
 
-/* Specify the memory areas */
+/* Specify the memory areas. */
 MEMORY
 {
-    FLASH (rx)      : ORIGIN = 0x08000000, LENGTH = 230K /* sectors 0-114 */
-    FLASH_FS (r)    : ORIGIN = 0x08039800, LENGTH = 26K /* sectors 115-127 */
+    MCU_FLASH (rx)  : ORIGIN = 0x08000000,  LENGTH = MICROPY_HW_MCU_FLASH_SIZE
+    FLASH (rx)      : ORIGIN = ORIGIN(MCU_FLASH), LENGTH = LENGTH(MCU_FLASH) - MICROPY_HW_FS_SIZE
```
dpgeorge, line 11: I suggest removing the `MCU_FLASH` entry and just have:
```
FLASH (rx)      : ORIGIN = 0x08000000, LENGTH = MICROPY_HW_MCU_FLASH_SIZE - MICROPY_HW_FS_SIZE
```

ports/stm32/mphalport.c
dpgeorge: Please use `#if defined(STM32L4) && defined(GPIOG)`.

ports/stm32/stm32_it.c
```diff
@@ -311,7 +311,7 @@ void USB_DRD_FS_IRQHandler(void) {
 }
 #endif
 
-#elif defined(STM32L0) || defined(STM32L432xx) || defined(STM32L452xx)
+#elif defined(STM32L0) || (defined(STM32L4)) && defined(USB_BASE) /* non-multi-OTG */
```
dpgeorge, line 314: I think the closing parenthesis needs to be at the end, to group the `&&` arguments together.

dpgeorge (commented):
Thanks @chrismas9 for keeping up with this PR, and sorry it's been hanging around for a while now.

> I have only implemented this for NUCLEO_L432KC, but will do it for all L4 boards if you approve. We would need to keep the old linker scripts to avoid breaking third party projects. I can also do the other STM families if you want.

The new scheme looks good.  If you want to apply it to all L4 boards that would be great.

But please leave the other families alone for now (mainly so we can get this PR through without further delay).

PR #16525, opened 2025-01-04
tests/extmod/re_sub.py: Fix test execution on Python 3.13.1.

agatti:
### Summary

This PR fixes a test failure for `extmod/re_sub.py` where the code, whilst being correct, would not make the test pass due to a newer Python version than expected.

On Python 3.13.1, running `tests/extmod/re_sub.py` would yield a deprecation warning about `re.sub` not providing the match count as a keyword parameter:

```bash
/micropython/tests/extmod/re_sub.py:50: DeprecationWarning: 'count' is passed as positional argument
  print(re.sub("a", "b", "1a2a3a", 2))
```

This warning would be embedded in the expected test result and thus the test would always fail.

The expected ouput file was generated by running `python tests/extmod/re_sub.py > tests/extmod/re_sub.py.exp`, with stderr output not ending up in the file.  The output is identical to older Python versions than 3.13.1.

### Testing

`extmod/re_sub.py` passes on an ESP32-C3 with ESP-IDF 5.2.0 with no code changes, using Python 3.13.1 to generate the expected test output.

I can see two alternatives to this.  The first one is to change the expected test output collector to ignore what comes out of `stderr` (and eventually print something alerting the user of that fact).  The second one is to make `re.sub`'s `count` parameter available also as a keyword argument and change the parsing code involved.

The first option isn't a long-term fix as the test will fail on newer Python versions anyway.  The second one may make the code size larger, although it has the potential to streamline argument parsing (by using `mp_arg_parse_all` instead of manual checking).

### DIFF
--- a/tests/extmod/re_sub.py
+++ b/tests/extmod/re_sub.py
@@ -10,6 +10,8 @@ except AttributeError:
     print("SKIP")
     raise SystemExit
 
+import sys
+
 
 def multiply(m):
     return str(int(m.group(0)) * 2)
@@ -47,7 +49,11 @@ print(re.sub("(abc)", r"\g<1>\g<1>", "abc"))
 print(re.sub("a", "b", "c"))
 
 # with maximum substitution count specified
-print(re.sub("a", "b", "1a2a3a", 2))
+if sys.implementation.name != "micropython":
+    # On CPython 3.13 and later the substitution count must be a keyword argument.
+    print(re.sub("a", "b", "1a2a3a", count=2))
+else:
+    print(re.sub("a", "b", "1a2a3a", 2))
 
 # invalid group
 try:


stinos:
Warning could be suppressed though that's also not longterm, but wouldn't the simplest (because of not using.exp files) and clearest (because the code makes it clear why the change is needed) solution be to make the code conditional on whether it's running on CPython? I mean like
```
if sys.implementation != "micropython":
  # On CPython versions >=3.13 the keyword argument is required.
  re.sub(...., count=2)
else:
  re.sub(...., 2)
```

dpgeorge:
> solution be to make the code conditional on whether it's running on CPython?

Yes, this sounds good to me.  And there's precedent to do this, eg the `tests/net_inet/test_tls_sites.py` test configures the SSLContext differently when running under CPython.

agatti:
> And there's precedent to do this, eg the `tests/net_inet/test_tls_sites.py` test configures the SSLContext differently when running under CPython.

As a counterargument, there's also `tests/float/math_domain_python311.py` that introduces a whole new test for a specific CPython version, along with its expected output :)

I'd say that providing a baseline for MicroPython's behaviour in this case is slightly more future-proof than adding a CPython version check that will probably linger around until somebody cleans the tests up to remove support for then-obsolete CPython versions.

Speaking of which, what's the earliest CPython version that is supported for running tests, 3.5?

dpgeorge:
> As a counterargument, there's also `tests/float/math_domain_python311.py` that introduces a whole new test for a specific CPython version, along with its expected output :)

If you want to follow the logic of that test and apply it here, then I think you would want to create `tests/extmod/re_sub_python313.py` (along with a .exp) and it would contain only the single `print(re.sub("a", "b", "1a2a3a", 2))` line.  Everything else still works fine and matches under Python 3.13, so doesn't need a .exp.

The problem with a .exp is that it's harder to update the test, adding new test cases to it, because then you also have to work out the expected output and add that as well (instead of it just being automatic, what CPython outputs).

> I'd say that providing a baseline for MicroPython's behaviour in this case is slightly more future-proof than adding a CPython version check that will probably linger around until somebody cleans the tests up to remove support for then-obsolete CPython versions.

But the CPython version check (actually just the fact it's running under CPython, there's no check of the version) will always exist if MicroPython's behaviour with `re.sub` doesn't change: newer CPython will require a keyword argument, while MicroPython does not accept a keyword argument.

> Speaking of which, what's the earliest CPython version that is supported for running tests, 3.5?

MicroPython originally targeted CPython 3.4.  But since the test `tests/basics/string_fstring.py` doesn't have a .exp, you'd need at least CPython 3.6 to run the tests.

agatti:
> If you want to follow the logic of that test and apply it here, then I think you would want to create `tests/extmod/re_sub_python313.py` (along with a .exp) and it would contain only the single `print(re.sub("a", "b", "1a2a3a", 2))` line. Everything else still works fine and matches under Python 3.13, so doesn't need a .exp.

Got it.  There may be one problem that neither my original patch nor this approach solves though.  `tests/extmod/re_sub.py` can be skipped if neither the `re` module nor the `re.sub` method are enabled.  If regex support is partially enabled or not enabled at all then the test will always fail.

That said, the original suggestion from @stinos is probably the better approach as it handles that case as well.  I'll update the PR accordingly, thanks!

dpgeorge (approved):

PR #16531, opened 2025-01-06
mpremote: Fix endless loops

hmaerki:
### Summary

In two situations, `mpremote` blocks forever.

### Testing

* Situation A
  * Precondition: Call mpremote.transport `read_until()`. The device does not return any data.
  * Observation: `read_until()` blocks forever.
  * Why: `serial.read(min_num_bytes)` is called without checking  `inWaiting() > 0` which then blocks forever.
  * Expected: `read_unit()` returns after timeout.

* Situation B
  * Precondition: Call mpremote.transport `read_until()`. The device is a brand new ESP32-C3-DevKitC. The factory installed firmware spits out `example: log -> UART` every second.
  * Observation: `read_until()` blocks forever.
  * Why: The `timeout_count` is reset every second.
  * Expected: `read_unit()` returns after timeout.

### Fix

**This PR just fixes 'Situation A'!**

* `serial.read()` is always guarded by  `inWaiting() > 0`. This will prevent blocking.

### Notes

* I added the uart-timeout to the constructor. This is not used but I may be used as an addition safeguard.

### Testing supplemental

This endless loop/blocker may be demonstrated using a RP2:
* Flash a RP2 with this firmware: https://github.com/raspberrypi/debugprobe/releases/tag/debugprobe-v2.2.0
* Reboot the RP2
* Call `mpremote a0`: Hangs forever.
* Call `mpremote a0 ls`: Hangs forever. Or timeout after 10s if this PR is applied.

### DIFF
--- a/tools/mpremote/mpremote/transport_serial.py
+++ b/tools/mpremote/mpremote/transport_serial.py
@@ -42,7 +42,7 @@ from .transport import TransportError, TransportExecError, Transport
 
 
 class SerialTransport(Transport):
-    def __init__(self, device, baudrate=115200, wait=0, exclusive=True):
+    def __init__(self, device, baudrate=115200, wait=0, exclusive=True, timeout=None):
         self.in_raw_repl = False
         self.use_raw_paste = True
         self.device_name = device
@@ -52,7 +52,11 @@ class SerialTransport(Transport):
         import serial.tools.list_ports
 
         # Set options, and exclusive if pyserial supports it
-        serial_kwargs = {"baudrate": baudrate, "interCharTimeout": 1}
+        serial_kwargs = {
+            "baudrate": baudrate,
+            "timeout": timeout,
+            "interCharTimeout": 1,
+        }
         if serial.__version__ >= "3.3":
             serial_kwargs["exclusive"] = exclusive
 
@@ -95,13 +99,20 @@ class SerialTransport(Transport):
         self.serial.close()
 
     def read_until(self, min_num_bytes, ending, timeout=10, data_consumer=None):
-        # if data_consumer is used then data is not accumulated and the ending must be 1 byte long
-        assert data_consumer is None or len(ending) == 1
+        """
+        min_num_bytes: Obsolete.
+        ending: Return if 'ending' matches.
+        timeout [s]: Return if timeout between characters. None: Infinite timeout.
+        data_consumer: Use callback for incoming characters.
+            If data_consumer is used then data is not accumulated and the ending must be 1 byte long
 
-        data = self.serial.read(min_num_bytes)
-        if data_consumer:
-            data_consumer(data)
-        timeout_count = 0
+        It is not visible to the caller why the function returned. It could be ending or timeout.
+        """
+        assert data_consumer is None or len(ending) == 1
+        assert isinstance(timeout, (type(None), int, float))
+
+        data = b""
+        begin_char_s = time.monotonic()
         while True:
             if data.endswith(ending):
                 break
@@ -112,10 +123,9 @@ class SerialTransport(Transport):
                     data = new_data
                 else:
                     data = data + new_data
-                timeout_count = 0
+                begin_char_s = time.monotonic()
             else:
-                timeout_count += 1
-                if timeout is not None and timeout_count >= 100 * timeout:
+                if timeout is not None and time.monotonic() >= begin_char_s + timeout:
                     break
                 time.sleep(0.01)
         return data


dpgeorge:
I think this will be a breaking change to the mpremote-as-a-library API:
- `SerialTransport.follow(10)` will now return after 10 seconds instead of waiting for input forever.  Users may be expected it to wait forever (eg follow the output for hours).  The 10 second timeout here is the time between successive output, and the function returns if nothing is seen for 10 seconds.
- `SerialTransport.exec()` will have a similar change in behaviour: prior to this patch it could wait for a long time for the output to complete.  Now it returns after 10 seconds maximum (with no way to control the timeout).

Really I think we need two or maybe three timeout values:
- timeout for first character received
- timeout between subsequent characters
- overall timeout measured from the start

tools/mpremote/mpremote/transport_serial.py
```diff
@@ -97,27 +101,33 @@ def close(self):
...
-        if data_consumer:
-            data_consumer(data)
-        timeout_count = 0
+        data = b""
+        begin_s = time.monotonic()
         while True:
+            if self.serial.inWaiting() <= 0:
+                # Wait for input
+                time.sleep(0.01)
+                continue
```
dpgeorge, line 112: Are you sure this won't wait forever if no characters come in at all at the beginning?

hmaerki:
> Really I think we need two or maybe three timeout values:
> timeout for first character received
> timeout between subsequent characters
> overall timeout measured from the start

Shall I make a proposal? Shall I start a new PR for the that?

dpgeorge:
Can we take it step-by-step and just fix Situation A first, in this PR?  That will go through much faster, because it shouldn't require any API decisions.

tools/mpremote/mpremote/transport_serial.py
```diff
@@ -95,29 +99,45 @@ def close(self):
...
 
-        data = self.serial.read(min_num_bytes)
-        if data_consumer:
-            data_consumer(data)
-        timeout_count = 0
+        data = b""
+        begin_char_s = time.monotonic()
         while True:
-            if data.endswith(ending):
+            if timeout and time.monotonic() > begin_char_s + timeout:
```
dpgeorge, line 117: What happens if `timeout` is 0?

tools/mpremote/mpremote/transport_serial.py
```diff
@@ -95,29 +99,45 @@ def close(self):
...
-                timeout_count += 1
-                if timeout is not None and timeout_count >= 100 * timeout:
-                    break
+
+            if self.serial.inWaiting() <= 0:
+                # Wait for input
                 time.sleep(0.01)
+                continue
+
+            new_data = self.serial.read(1)  # Will not block!
```
dpgeorge, line 125: IMO, this comment is not necessary.

tools/mpremote/mpremote/transport_serial.py
```diff
@@ -95,29 +99,45 @@ def close(self):
...
+            begin_char_s = time.monotonic()
+            if data_consumer:
+                data_consumer(new_data)
+                data = new_data
+            else:
+                data = data + new_data
+
+            if min_num_bytes > 1:
+                # Make sure 'min_num_bytes' are read before
+                # testing for 'ending'.
```
dpgeorge, line 135: There's no need for this check, you can simply do `if data.endswith(ending)` (and it'll be `False` if there aren't yet enough bytes in `data`).

In fact, `min_num_bytes` is no longer needed with this PR.  That parameter was originally used for the initial blocking call to `read(min_num_bytes)` at the start of this function.

tools/mpremote/mpremote/transport_serial.py
dpgeorge: A `timeout` of 0 is still considered the same as `None`, ie no timeout.  That's not the original semantics, where 0 would mean "non blocking" (except for the initial read...).

dpgeorge: @hmaerki this comment still needs addressing.  The code must handle timeout=0 to mean non-blocking.

dpgeorge: Sorry to be pedantic here, but I think there is an edge case where the timeout is small but non-zero and the PC is slow (eg switches to a different task which for some time) and this means the timeout is reached here before the serial port is checked.

The serial port should always be checked at least once, before the timeout is checked.

dpgeorge:
I think there's a much simpler change to the code that achieves the same thing as this PR.  Namely:
```diff
--- a/tools/mpremote/mpremote/transport_serial.py
+++ b/tools/mpremote/mpremote/transport_serial.py
@@ -98,10 +98,8 @@ class SerialTransport(Transport):
         # if data_consumer is used then data is not accumulated and the ending must be 1 byte long
         assert data_consumer is None or len(ending) == 1
 
-        data = self.serial.read(min_num_bytes)
-        if data_consumer:
-            data_consumer(data)
-        timeout_count = 0
+        data = b""
+        begin_char_s = time.monotonic()
         while True:
             if data.endswith(ending):
                 break
@@ -112,10 +110,9 @@ class SerialTransport(Transport):
                     data = new_data
                 else:
                     data = data + new_data
-                timeout_count = 0
+                begin_char_s = time.monotonic()
             else:
-                timeout_count += 1
-                if timeout is not None and timeout_count >= 100 * timeout:
+                if timeout is not None and time.monotonic() >= begin_char_s + timeout:
                     break
                 time.sleep(0.01)
         return data
```

hmaerki:
I like you simple solution! I rebased and squashed into one simple commit.

For my taste I would reshape the code slightly:

> Clan Code (Robert Martin)
> Blocks and Indenting
> ...
> This also implies that functions should not be large enough to hold nested structures.
> Therefore, the indent level of a function should not be greater than one or two. This, of
> course, makes the functions easier to read and understand.

But lets leave it as it is!

dpgeorge (approved):

dpgeorge:
> For my taste I would reshape the code slightly:

It's always best to keep the existing code style.

And also usually best to minimise a diff so that it's easier to review, and easier to go back to the commit and understand the change.

PR #16535, opened 2025-01-07
doc,esp32: Improve and de-duplicate the flashing instructions.

projectgus:
### Summary

This PR aims to clean up the flashing instructions for the esp32 port, and also make them easier to maintain in the future.

The big change for the download pages is using string formatting to expand the very few things (chip name, flash offset) which are different between most boards. This allows using the same two `deploy.md` and `deploy_nativeusb.md` for almost all ESP32 boards, which also made it easier to add some additional troubleshooting docs there.

The initial motivation comes from a few places:

* @andrewleech pointed out that esptool port auto-detection works well these days: https://github.com/micropython/micropython/pull/15956#discussion_r1802015892
* I noticed that the instructions on the download page are also pretty bare-bones for most boards. They also include redundant options like `-z` (which is enabled by default).
* @donburch888 pointed out that the setup instructions give the wrong flashing address for newer chips: https://github.com/orgs/micropython/discussions/16417 - that page is also missing a faster baud rate, etc.

Specific changes applied across here include:

* Remove unnecessary command line arguments.
* ~~Harmonise the flashing steps between the download pages and the ESP32 getting started doc.~~ EDIT: Now the tutorial page refers to the download page for flashing steps - it's much simpler as these are the steps for this particular board, rather than "any" board.
* Expanded our troubleshooting steps a bit, to include getting into download mode on most boards.
* Also link to the esptool.py docs when possible, as these are quite good.

### Testing

* Used `./build-downloads.py` to generate download pages from master and this branch, and diffed the directories to ensure no non-esp32-port files changed and that nothing important was missing from the updated esp32 download pages.
* I've uploaded the generated files from this branch to a gist, here: https://gist.github.com/projectgus/f2d24fa2e4689c41ce8002b70f0926bf#file-esp32c6-md

* Having lots of copy-paste deploy.md is less "clever", but much more painful to maintain and update.
* If we want even more flexibility in how we generate these in the future then we could migrate to something like [Jinja2 templating](https://jinja.palletsprojects.com/en/stable/templates/). This would let us generate quite sophisticated sets of download instructions (for example, we could add DFU flashing instructions for all boards where the chip supports USB DFU-mode). Probably overkill for now.


projectgus:
@UnexpectedMaker I'd be interested to know what you think of this, as your boards have the best download instructions included (:star_struck:). I think I picked up all the per-board variants (flashing mode buttons, and the boards which need external flashing mode wiring).

You can see the new generated deploy.md files for UM boards here: https://gist.github.com/projectgus/f2d24fa2e4689c41ce8002b70f0926bf#file-um_feathers2-md

and also these ones where the name has no UM_ prefix: [nanos3](https://gist.github.com/projectgus/f2d24fa2e4689c41ce8002b70f0926bf#file-nanos3-md) and [omgs3](https://gist.github.com/projectgus/f2d24fa2e4689c41ce8002b70f0926bf#file-omgs3-md)

Please let me know if you think anything here is wrong, or might lead to a worse user experience.

ports/esp32/boards/deploy.md
```diff
@@ -1,14 +1,44 @@
...
+Program your board using the esptool.py program, found [here](https://docs.espressif.com/projects/esptool/en/latest/{mcu}/).
+
+### Erasing
 
 If you are putting MicroPython on your board for the first time then you should
 first erase the entire flash using:
 
 ```bash
-esptool.py --chip esp32 --port /dev/ttyUSB0 erase_flash
+esptool.py --chip {mcu} erase_flash
```
projectgus, line 9: I wasn't sure what to do with the `--chip` argument here but I settled on keeping it for the erase_flash command (one less chance of erasing wrong chip), but omitting it for write_flash command (esptool.py will error out if it detects the .bin file doesn't match the target chip, anyway).

UnexpectedMaker:
@projectgus Hiya!

A few things... 
- I'm not sure what you mean by NANOS3 and OMGS3 not having UM_ ? All of my boards have an UM_ prefix
- For the S2/S3 boards (those that use native USB) there should be no `--baud` setting as it's fixed speed and adding that makes no difference and is confusing for people. The same thing applies to the message about dropping it to slow transfer if there are transfer issues.
- Seems to be inconsistencies with board naming within each doc - for instance in FeatherS2 - sometimes the board is called `FeatherS2` and sometimes `Feather S2`
- 
Otherwise, I think consistency across all of the boards is great, good job!

projectgus:
Thanks for the quick review, @UnexpectedMaker!

> * I'm not sure what you mean by NANOS3 and OMGS3 not having UM_ ? All of my boards have an UM_ prefix

Oh, right. I put direct links to those two as they're not grouped with the other files in the gist. All the other `UM_SOMETHING` boards generate markdown files named `UM_SOMETHING.md` but these ones are `nanos3.md` and `omgs3.md`. This name comes from the `board.json` contents, not the board directory name. I don't think it changes anything apart from the download URLs, i.e. there is currently:

* https://micropython.org/download/omgs3/
* https://micropython.org/download/UM_TINYS2/

Happy to change `omgs3` and `nanos3` to be consistent with the others in this PR, but I don't particularly mind (and leaving them as-is will avoid breaking any existing links).

>     * For the S2/S3 boards (those that use native USB) there should be no `--baud` setting as it's fixed speed and adding that makes no difference and is confusing for people. The same thing applies to the message about dropping it to slow transfer if there are transfer issues.

Good catch! I've added a separate `deploy_nativeusb.md` and pointed all the boards with native USB to this one. The gist is updated as well, please take a look.

>     * Seems to be inconsistencies with board naming within each doc - for instance in FeatherS2 - sometimes the board is called `FeatherS2` and sometimes `Feather S2`

Ah, this is where the `board.md` file uses one form and the `board.json` file uses the other. Which form do you prefer: with or without the space? I'll clean it up in this PR.

UnexpectedMaker:
> Thanks for the quick review, @UnexpectedMaker!
> 
> Oh, right. I put direct links to those two as they're not grouped with the other files in the gist. All the other `UM_SOMETHING` boards generate markdown files named `UM_SOMETHING.md` but these ones are `nanos3.md` and `omgs3.md`. This name comes from the `board.json` contents, not the board directory name. I don't think it changes anything apart from the download URLs, i.e. there is currently:
> 
> * https://micropython.org/download/omgs3/
> * https://micropython.org/download/UM_TINYS2/
> 
> Happy to change `omgs3` and `nanos3` to be consistent with the others in this PR, but I don't particularly mind (and leaving them as-is will avoid breaking any existing links).
> 

Ahh, hmm, my mistake then, sorry. They should all be `um_` - I can fix that later if you want, or if you are already doing stuff, happy for you to do it ;) 

> Good catch! I've added a separate `deploy_nativeusb.md` and pointed all the boards with native USB to this one. The gist is updated as well, please take a look.
> 

Excellent!

> > * Seems to be inconsistencies with board naming within each doc - for instance in FeatherS2 - sometimes the board is called `FeatherS2` and sometimes `Feather S2`
> 
> Ah, this is where the `board.md` file uses one form and the `board.json` file uses the other. Which form do you prefer: with or without the space? I'll clean it up in this PR.

Hmm, again, seems my mistake/inconsistency when adding the board data over time. There should be no spaces in my names other than when there is a `Neo` - so `FeatherS3` and `FeatherS3 Neo`  - All S2/S3 bits have no spaces before them.

Thanks for being so responsive Angus - Happy New Year btw! I hope you and the family are well :)

projectgus:
@UnexpectedMaker Great! I've pushed a commit to change the spacings, as this PR otherwise will make the content on those pages internally inconsistent.

I've left the omgs3/tinys3 naming alone, thanks for offering to pick that one up. It probably makes more sense to do it in a separate PR as it's separate from these changes. Feel free to tag me for review if you do submit a PR for it.

And Happy New Year to you and your family, as well! :grin:

Josverl (changes requested):
On >90% of Windows the command `esptool.py` will fail 
below is for `pipx install`ed esptool   
image

I suggest to clarify this here , or in another place where the installation of the esptool package is described.

and HNY 🥂

Josverl:
One other thing that I found when creating mpflash was that of the ESP32 port , the ARDUINO_NANO_ESP32 cannot be flashed with esptool , but needs dfutil 

The documentation is in place , but its easy to miss as its an outlier from most of the ESPxxx boards
https://github.com/micropython/micropython/blob/bf35eefc625b35aa301aefbc743dcf1395829865/ports/esp32/boards/ARDUINO_NANO_ESP32/deploy.md

https://github.com/Josverl/micropython-stubber/blob/main/src/mpflash/mpflash/flash/esp.py

projectgus:
Thanks @Josverl for the insights! I've added notes for Windows users, and updated the gist with the previews.

> The documentation is in place , but its easy to miss as its an outlier from most of the ESPxxx boards

Yes, that one is kept unchanged in this PR (I did add a spurious `flash_offset` key to the boards.json file, but the actual `deploy.md` is the same.

Do you suggest additional changes? We could mention this in the `intro.rst` file as well, I suppose, but I think if we go that far then we should probably remove the flashing instructions from there entirely and direct everyone to their board's download page instead.

Josverl:
> Do you suggest additional changes? We could mention this in the intro.rst file...

Not in detail, but perhaps a note like:

While many board in a port share the same flashing tools and common parameters, some boards will use different parameters or even different  flashing tools or methods.

ports/esp32/boards/ARDUINO_NANO_ESP32/board.json
```diff
@@ -2,6 +2,7 @@
     "deploy": [
         "deploy.md"
     ],
+    "flash_offset": "0",
```
dpgeorge, line 5: I can see why you put this key here, to make it very simple to do templating.

But, it doesn't seem right to have this key at the top-level like this in `board.json`.  Other ports probably need different keys, etc.

Instead, maybe have an entry here which is a dict for the templating: `"deploy_options": { "flash_offset": "0" }`.  Or allow such a dict to be inside the existing `deploy` key, eg:
```
"deploy": [
    "deploy.md",
    {
        "flash_offset": "0"
    }
]
```

So the template can see top-level keys like `product`, as well as deploy-specific keys, you can format using, eg:
```py
fin.format(**blob, **blob.get("deploy_options", {}))
```
Or the template can access nested options like `deploy_options[flash_offset]`.

There are quite a few ways it could be done.  IMO it should be simple, but I think having `flash_options` at the top level is not good for future generalisation of this templating feature.

dpgeorge (commented):
Thanks @projectgus for this clean up, it's definitely a good improvement and will make it easier to maintain these instructions.

Will you squash the two commits that modify `docs/esp32/tutorial/intro.rst`?  It seems like the latter one fixes the first.

ports/esp32/boards/ARDUINO_NANO_ESP32/board.json
dpgeorge: Note that Python allows subscripting dicts within `str.format` without using quotes, so you could do something like this in the template:
```
flash at offset {deploy_options[flash_offset]} on esp32
```
That only requires `fin.format(**blob)` in the template expansion code.

Or a very simple option would be to just use the key `deploy_flash_offset` (at the top level of `board.json`), which at the very least puts the options in a namespace by virtue of the `deploy_` prefix.

Side note: eventually it would be great to embed flashing code (esptool, dfu-util etc) within the download webpage so it's as simple as pressing a button to deploy firmware (as long as you're using Chrome...).  And for that, these deploy options will probably be used by the webpage to execute the correct command.

projectgus: Done, I went with the suggested `deploy_options` key.

projectgus:
Thanks @dpgeorge and @Josverl for the review comments. Have updated accordingly (and the preview gist is also refreshed).

I ended up removing the flashing steps from the tutorial entirely and directing users to the download page. In most cases they already have that page open, and it's much easier to follow because it's the specific steps for that board only.

I also made some more tweaks, particularly around the instructions for choosing a board to download as I realised these were also quite out of date. I moved the information about `ESP32_GENERIC` variants out of the tutorial as well, and expanded the descriptions in that board's `board.md` file.

docs/esp32/tutorial/intro.rst
```diff
@@ -36,106 +36,95 @@ Getting the firmware
...
-If the above commands run without error then MicroPython should be installed on
-your board!
+If you experience problems during flashing or with running firmware immediately
+after flashing, here are some troubleshooting recommendations:
+
+* Esptool will try to detect the serial port where your ESP32 is connected. If
+  this doesn't work, or you have multiple serial ports, then you may need to
+  manually specify the port by adding the ``--port`` option to the start of the
+  ``esptool.py`` command line. For example, ``esptool.py --port /dev/ttyUSB0
+  <rest of line>`` for Linux or ``esptool.py --port COM4 <rest of line>`` for
```
mattytrentini, line 89:
```suggestion
  <rest of line>`` for Linux or ``esptool --port COM4 <rest of line>`` for
```

mattytrentini (approved):
A nice improvement; queries about flashing ESP32's are _very_ common. I think this will help!

docs/esp32/tutorial/intro.rst
projectgus: Thanks @mattytrentini. Good catch on the command name here, have applied.

dpgeorge (approved):

PR #16544, opened 2025-01-08
stm32: Fix pyb.CAN BRS baud rate, make calculation more forgiving.

### LINKED ISSUE #16330: Stm32 CAN-FD faulty timing configuration

stm32

### MicroPython version

self compiled MicroPython v1.25.0-preview.10.g0e490b7c8.dirty on 2024-11-17 on custom Board using STM32H750VBT6

### Reproduction

The following code was running on the custom board to test the CAN-FD-Bus
```ruby
import pyb
import time

can_stb = pyb.Pin('PC11', pyb.Pin.OUT_PP)
can_stb.high()

can_en = pyb.Pin('PC12', pyb.Pin.OUT_PP)
can_en.high()

#can = pyb.CAN(1, pyb.CAN.NORMAL, prescaler=1, sjw=16, bs1=63, bs2=16, brs_prescaler=1, brs_sjw=6, brs_bs1=13, brs_bs2=6)
can = pyb.CAN(1, pyb.CAN.NORMAL, baudrate=500_000, brs_baudrate=2_000_000, sample_point=80, brs_sample_point=70)

while True:
    try:
        can.send('a'*64, 0x123, fdf=True, brs=True, extframe=False)
    except:
        print("FiFo full")
    pyb.delay(1000)
```
A Computer with an external USB to CAN-FD Converter device was also connected to the same CAN-FD-Bus. The CAN-Bus is also properly terminated using 120 Ohms resistors on each end.

### Expected behaviour

After enabling the CAN-Transceiver and configuring the CAN-Bus timing the code is supposed to periodically send a message on the CAN-Bus. The message has a data length of 64 Bytes which means it is not a regular CAN-Frame but rather a CAN-FD-Frame. The code also enables the so called Bit-Rate-Switch (BRS). This CAN-FD specific feature allows for a higher transmission speed during the data-field of the CAN-FD-Frame. 

### Observed behaviour

After a few seconds the Message "FiFo full" is printed in the terminal and there are no messages received on the other side. Instead a lot of Error-Frames are being send on the Bus which means there is either a timing or signal integrity issue.

### Additional Information

This problem only occures when the BRS is enabled. As soon as the BRS is disabled proper CAN-FD-Frames with 64 Data-Bytes but without BRS are send and received on the other side. This implies that the CAN-Controller might not be configured properly.

The Timing for the CAN-FD is set in the 'pyb_can.c' file
First in line 263 to 268 the timing is calculated.
```ruby
// Calculate CAN data bit timing from baudrate if provided
if (args[ARG_brs_baudrate].u_int != 0) {
    pyb_can_get_bit_timing(args[ARG_brs_baudrate].u_int, args[ARG_brs_sample_point].u_int,
        CAN_MAXIMUM_DBRP, CAN_MAXIMUM_DBS1, CAN_MAXIMUM_DBS2, 1,
        &args[ARG_brs_bs1].u_int, &args[ARG_brs_bs2].u_int, &args[ARG_brs_prescaler].u_int);
}
```
Please notice that the Timing values for the data-field are named after the scheme **ARG_brs_xxx**.
Next from line 269 to 273 the calculated timing values are handed over to the hardware abstraction layer (HAL).
```ruby
// Set BRS bit timings.
self->can.Init.DataPrescaler = args[ARG_brs_prescaler].u_int;
self->can.Init.DataSyncJumpWidth = args[ARG_brs_sjw].u_int;
self->can.Init.DataTimeSeg1 = args[ARG_bs1].u_int; // DataTimeSeg1 = Propagation_segment + Phase_segment_1
self->can.Init.DataTimeSeg2 = args[ARG_bs2].u_int;
```
Now here it can be seen that the last two timing values are named **ARG_bs1** and **ARG_bs2** which is not correct. Instead they should be called **ARG_brs_bs1** and **ARG_brs_bs2** because these are the names of the timing values that have been calculated in the last code snippet.
In my local copy of the micropython repository i have applied this patch, recompiled and uploaded to my custom board. I ran the code again and now proper CAN-FD-Frames with 64 Data-Bytes and BRS are send and received on the other side.
To me it looks like a copy and paste error because the code for setting up the standard can timing is identical just with values called **ARG_xxx**.
I guess that a micropython user who only has stm32h7 boards connected to a CAN-FD-Bus would not even notice this bug, because than all CAN-Controllers have the same faulty configuration and could still talk to each other. But in my case there was a third party device with a correct configuration connected to the bus.
It would be nice if someone could approve this fix and implement it in the standard repository.

### Code of Conduct

Yes, I agree

projectgus:
### Summary

* Fixes #16330. Thanks @albertstoewer for the very comprehensive bug report.
* As part of verifying the fix, I realised the baud rate calculations are very unforgiving. Have added a small amount of error tolerance, which allowed me to set desired baud rates on STM32H723G.
* Similarly, the error when a baudrate couldn't be achieved was hard to debug on CAN-FD as the calculation is repeated twice (once for the Classic baud rate and once for BRS). Added the failing rate to the error message.

### Testing

* Tested using a NUCLEO STM32H723G baud using the test code submitted in #16330 with sample points both changed to 75 in order to be accepted. For the device on the other side I modified the [embassy stm32g4 can example](https://github.com/embassy-rs/embassy/blob/main/examples/stm32g4/src/bin/can.rs) to use the same baud rates and flashed on the NUCLEO STM32G474RE. Confirmed no frames received by the STM32G4 without the patch, and received successfully with the patch.
* For the STM32H723 to work successfully I needed to cherry-pick the fix from #16543 first.

### DIFF
--- a/ports/stm32/pyb_can.c
+++ b/ports/stm32/pyb_can.c
@@ -269,8 +269,8 @@ static mp_obj_t pyb_can_init_helper(pyb_can_obj_t *self, size_t n_args, const mp
     // Set BRS bit timings.
     self->can.Init.DataPrescaler = args[ARG_brs_prescaler].u_int;
     self->can.Init.DataSyncJumpWidth = args[ARG_brs_sjw].u_int;
-    self->can.Init.DataTimeSeg1 = args[ARG_bs1].u_int; // DataTimeSeg1 = Propagation_segment + Phase_segment_1
-    self->can.Init.DataTimeSeg2 = args[ARG_bs2].u_int;
+    self->can.Init.DataTimeSeg1 = args[ARG_brs_bs1].u_int; // DataTimeSeg1 = Propagation_segment + Phase_segment_1
+    self->can.Init.DataTimeSeg2 = args[ARG_brs_bs2].u_int;
     #else
     // Init filter banks for classic CAN.
     can2_start_bank = args[ARG_num_filter_banks].u_int;
--- a/ports/stm32/pyb_can.c
+++ b/ports/stm32/pyb_can.c
@@ -222,7 +222,7 @@ static void pyb_can_get_bit_timing(mp_uint_t baudrate, mp_uint_t sample_point,
         }
     }
 
-    mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("couldn't match baudrate and sample point"));
+    mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("couldn't match baudrate %u and sample point %u"), baudrate, sample_point / 10);
 }
 
 // init(mode, prescaler=100, *, sjw=1, bs1=6, bs2=8)
--- a/ports/stm32/pyb_can.c
+++ b/ports/stm32/pyb_can.c
@@ -25,6 +25,7 @@
  */
 
 #include <string.h>
+#include <stdlib.h>
 
 #include "py/objarray.h"
 #include "py/runtime.h"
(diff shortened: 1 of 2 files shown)


ports/stm32/pyb_can.c
```diff
@@ -199,12 +200,19 @@ static void pyb_can_get_bit_timing(mp_uint_t baudrate, mp_uint_t sample_point,
...
         for (uint32_t bs1 = min_tseg; bs1 < max_bs1; bs1++) {
             for (uint32_t bs2 = min_tseg; bs2 < max_bs2; bs2++) {
-                if ((baudrate == (can_kern_clk / (brp * (1 + bs1 + bs2)))) &&
-                    ((sample_point * 10) == (((1 + bs1) * 1000) / (1 + bs1 + bs2)))) {
+                mp_int_t calc_baud = can_kern_clk / (brp * (1 + bs1 + bs2));
+                mp_int_t calc_sample = ((1 + bs1) * 1000) / (1 + bs1 + bs2);
+                mp_int_t baud_err = baudrate - calc_baud;
+                mp_int_t sample_err = sample_point - calc_sample;
+                if (abs(baud_err) < max_baud_error &&
+                    abs(sample_err) < MAX_SAMPLE_ERROR) {
```
dpgeorge, line 215: What happens if there's an exact parameter match but it's not used because this algorithm finds a "close enough" match first?  Maybe it needs to have a set of parameters "best match so far" and then it searches all options until it finds an exact match, or at the end falls back to a "close enough" match.

projectgus: I worried about this too, but I think the short answer is nothing noticeable. The error tolerances are pretty small: An unnecessary .1% baud rate error should still work fine unless the two nodes were already very marginal, and the chosen sample point should still match the argument when rounded to the nearest percent (the sample point argument doesn't support any higher precision than this).

(That said, the current version of the machine.CAN calculation *does* do the "track best option" you're describing, but that's mostly so that it can accept >.1% deviations rather than failing there as well. Although, after writing the simpler version here I was thinking about throwing that out to save code size. But I could add it to this PR instead, if you prefer. There's some small chance of regression though, as it's quite different.)

projectgus: Very Serious CAN users will likely always use the other configuration method (brp, bs1, bs2) anyway.

dpgeorge: > Very Serious CAN users will likely always use the other configuration method (brp, bs1, bs2) anyway.

Yes that's a good point.  So I guess it's fine to keep the code here as you have it, but maybe it's then worth documenting the fact that specifying `baudrate` will only get it to within 0.1% and if very precise timing is needed then it's better to specify bs1/bs2/brp.

dpgeorge (approved):

PR #16552, opened 2025-01-08
Add SparkFun IoT Node LoRaWAN RP2350

sfe-SparkFro:
### Summary

To add support for this board from us: https://www.sparkfun.com/sparkfun-iot-node-lorawan.html

Depends on #15620, but this can be merged before that.

### Testing

Tested with a board, all seems well.


ports/rp2/boards/SPARKFUN_IOTNODE_LORAWAN_RP2350/mpconfigboard.cmake
```diff
@@ -0,0 +1,3 @@
+# cmake file for SparkFun IoT Node LoRaWAN RP2350
+
+set(PICO_BOARD "sparkfun_iotnode_lorawan_rp2350")
```
dpgeorge, line 3: It looks like this board is not in pico-sdk (yet), so this won't build.  If you can't wait for pico-sdk then you can add the relevant header file in the directory here.

Also, do you need to set `PICO_PLATFORM`, at least for consistency with other boards?

sfe-SparkFro: > It looks like this board is not in pico-sdk (yet), so this won't build.

Correct, looks like it's slated for Pico SDK v2.1.1: https://github.com/raspberrypi/pico-sdk/pull/2143

> If you can't wait for pico-sdk then you can add the relevant header file in the directory here.

It's okay, no rush on this one.

> Also, do you need to set PICO_PLATFORM, at least for consistency with other boards?

Good catch, fixed!

dpgeorge: Just to be clear, you would like to wait for the pico-sdk release?  Note that it can take some time for those releases.  I'm happy to add the header file here as a temporary measure.

sfe-SparkFro: Adding the header here temporarily is totally fine, I just figured that would be more work for you. But if you're happy to handle it, that works for us!

dpgeorge: Adding it here is fine.

sfe-SparkFro: Pushed! How's this?

dpgeorge: Good, it now builds for me.

dpgeorge:
Squashed, rebased and merged in b2ce9b6fb03ed27943733235eac2f3f5e0d5b97d

PR #16564, opened 2025-01-10
github/workflows: Use Python 3.11 for unix settrace jobs.

dpgeorge:
### Summary

GitHub Actions has updated ubuntu-latest to 24.04, which now defaults CPython to 3.12, which has issues with our settrace test.

Fix that by explicitly using CPython 3.11.

Edit: also use ubuntu-22.04 for unix qemu CI jobs, because there are issues with newer libffi on those configurations (not sure exactly what's wrong there...).

Edit 2: also updated multi_test TLS test certificates and keys, because the older ones expired yesterday (13th Jan 2024).  Tested on unix, PYBD_SF6, RPI_PICO_W and ESP32.


projectgus (approved):

projectgus (approved):
Still LGTM!

PR #12293, opened 2023-08-23
esp32/machine_pin: Add mode and pull in machine_pin_print() as repr() function.

IhorNehrutsa:
**Summary**
This PR adds `mode`, `pull`, `drive` parameters to the Pin repr function.

The main advantage:
This PR allows to serialize Pin object to str and restore the Pin object from the string:
```
from machine import Pin

pin1 = Pin(27, mode=Pin.IN, pull=Pin.PULL_DOWN)
print(pin1)
pin2 = eval(str(pin1))
print(pin2)
```
Output is:
```
Pin(27, mode=Pin.IN, pull=Pin.PULL_DOWN)
Pin(27, mode=Pin.IN, pull=Pin.PULL_DOWN)
```

**Testing**
Test code is:
```
from time import sleep
from machine import Pin

PIN = 27

pin = Pin(PIN, mode=Pin.IN, pull=None)
sleep(2)
print(pin, pin())

pin = Pin(PIN, mode=Pin.IN, pull=Pin.PULL_UP)
sleep(2)
print(pin, pin())
assert pin() == 1

pin = Pin(PIN, mode=Pin.IN, pull=Pin.PULL_DOWN)
sleep(2)
print(pin, pin())
assert pin() == 0

pin = Pin(PIN, mode=Pin.OUT, value=1)
sleep(2)
print(pin, pin())
assert pin() == 1

pin = Pin(PIN, mode=Pin.OUT, value=0, drive=Pin.DRIVE_1)
sleep(2)
print(pin, pin())
assert pin() == 0
```
Output is:
```
Pin(27, mode=Pin.IN) 0
Pin(27, mode=Pin.IN, pull=Pin.PULL_UP) 1
Pin(27, mode=Pin.IN, pull=Pin.PULL_DOWN) 0
Pin(27, mode=Pin.OUT, pull=Pin.PULL_DOWN) 1
Pin(27, mode=Pin.OUT, pull=Pin.PULL_DOWN, drive=Pin.DRIVE_1) 0
```

PR #16565, opened 2025-01-10
esp32: IDF v5.4(v5.5-dev) compile.

IhorNehrutsa:
**Summary**
This PR attempts to fix the esp32 compilation with IDF v5.4 (v5.5-dev)

**Testing**
ESP32 generic, ESP32-S3 boards tested with **esp32/machine_pin: Add mode and pull in machine_pin_print() as repr() function.** #12293.

### DIFF
--- a/ports/esp32/network_ppp.c
+++ b/ports/esp32/network_ppp.c
@@ -100,7 +100,12 @@ static mp_obj_t ppp_make_new(mp_obj_t stream) {
 }
 MP_DEFINE_CONST_FUN_OBJ_1(esp_network_ppp_make_new_obj, ppp_make_new);
 
-static u32_t ppp_output_callback(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx) {
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+static u32_t ppp_output_callback(ppp_pcb *pcb, const void *data, u32_t len, void *ctx)
+#else
+static u32_t ppp_output_callback(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx)
+#endif
+{
     ppp_if_obj_t *self = ctx;
 
     mp_obj_t stream = self->stream;
@@ -109,7 +114,7 @@ static u32_t ppp_output_callback(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx)
     }
 
     int err;
-    return mp_stream_rw(stream, data, len, &err, MP_STREAM_RW_WRITE);
+    return mp_stream_rw(stream, (void *)data, len, &err, MP_STREAM_RW_WRITE);
 }
 
 static void pppos_client_task(void *self_in) {
--- a/ports/esp32/network_wlan.c
+++ b/ports/esp32/network_wlan.c
@@ -763,6 +763,10 @@ static const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
     #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
     { MP_ROM_QSTR(MP_QSTR_SEC_DPP), MP_ROM_INT(WIFI_AUTH_DPP) },
     #endif
+    #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+    { MP_ROM_QSTR(MP_QSTR_SEC_WPA3_ENT), MP_ROM_INT(WIFI_AUTH_WPA3_ENTERPRISE) },
+    { MP_ROM_QSTR(MP_QSTR_SEC_WPA2_WPA3_ENT), MP_ROM_INT(WIFI_AUTH_WPA2_WPA3_ENTERPRISE) },
+    #endif
 
     { MP_ROM_QSTR(MP_QSTR_PM_NONE), MP_ROM_INT(WIFI_PS_NONE) },
     { MP_ROM_QSTR(MP_QSTR_PM_PERFORMANCE), MP_ROM_INT(WIFI_PS_MIN_MODEM) },
@@ -770,7 +774,9 @@ static const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
 };
 static MP_DEFINE_CONST_DICT(wlan_if_locals_dict, wlan_if_locals_dict_table);
 
-#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+_Static_assert(WIFI_AUTH_MAX == 16, "Synchronize WIFI_AUTH_XXX constants with the ESP-IDF. Look at esp-idf/components/esp_wifi/include/esp_wifi_types_generic.h");
+#elif ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
 _Static_assert(WIFI_AUTH_MAX == 14, "Synchronize WIFI_AUTH_XXX constants with the ESP-IDF. Look at esp-idf/components/esp_wifi/include/esp_wifi_types_generic.h");
 #elif ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 2, 0)
 _Static_assert(WIFI_AUTH_MAX == 13, "Synchronize WIFI_AUTH_XXX constants with the ESP-IDF. Look at esp-idf/components/esp_wifi/include/esp_wifi_types.h");
--- a/py/persistentcode.c
+++ b/py/persistentcode.c
@@ -402,7 +402,7 @@ static mp_raw_code_t *load_raw_code(mp_reader_t *reader, mp_module_context_t *co
 
     #if MICROPY_EMIT_MACHINE_CODE
     } else {
-        const uint8_t *prelude_ptr;
+        const uint8_t *prelude_ptr = NULL;
         #if MICROPY_EMIT_NATIVE_PRELUDE_SEPARATE_FROM_MACHINE_CODE
         if (kind == MP_CODE_NATIVE_PY) {
             // Executable code cannot be accessed byte-wise on this architecture, so copy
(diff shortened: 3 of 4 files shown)


ports/esp32/network_ppp.c
```diff
@@ -145,7 +145,7 @@ static mp_obj_t ppp_active(size_t n_args, const mp_obj_t *args) {
                 return mp_const_true;
             }
 
-            self->pcb = pppapi_pppos_create(&self->pppif, ppp_output_callback, ppp_status_cb, self);
+            self->pcb = pppapi_pppos_create(&self->pppif, (void *)ppp_output_callback, ppp_status_cb, self);
```
DvdGiessen, line 148: Instead of erasing the callback type, since the function is IDF-specific anyway wouldn't it be better to update its signature? That way we'll still get compiler errors if there's potentially incompatible changes in the future.

```diff
diff --git a/ports/esp32/network_ppp.c b/ports/esp32/network_ppp.c
index c739264b9..a6e3b152f 100644
--- a/ports/esp32/network_ppp.c
+++ b/ports/esp32/network_ppp.c
@@ -101,7 +101,7 @@ static mp_obj_t ppp_make_new(mp_obj_t stream) {
 }
 MP_DEFINE_CONST_FUN_OBJ_1(esp_network_ppp_make_new_obj, ppp_make_new);
 
-static u32_t ppp_output_callback(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx) {
+static u32_t ppp_output_callback(ppp_pcb *pcb, const void *data, u32_t len, void *ctx) {
     ppp_if_obj_t *self = ctx;
 
     mp_obj_t stream = self->stream;
@@ -110,7 +110,7 @@ static u32_t ppp_output_callback(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx)
     }
 
     int err;
-    return mp_stream_rw(stream, data, len, &err, MP_STREAM_RW_WRITE);
+    return mp_stream_rw(stream, (void *)data, len, &err, MP_STREAM_RW_WRITE);
 }
 
 static void pppos_client_task(void *self_in) {
```

ports/esp32/network_wlan.c
```diff
@@ -770,7 +770,9 @@ static const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
 };
 static MP_DEFINE_CONST_DICT(wlan_if_locals_dict, wlan_if_locals_dict_table);
 
-#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+_Static_assert(WIFI_AUTH_MAX == 16, "Synchronize WIFI_AUTH_XXX constants with the ESP-IDF. Look at esp-idf/components/esp_wifi/include/esp_wifi_types_generic.h");
```
DvdGiessen, line 774: We should also actually add the new constants:

```diff
diff --git a/ports/esp32/network_wlan.c b/ports/esp32/network_wlan.c
index fed81d28c..d612d3cca 100644
--- a/ports/esp32/network_wlan.c
+++ b/ports/esp32/network_wlan.c
@@ -763,6 +763,11 @@ static const mp_rom_map_elem_t wlan_if_locals_dict_table[] = {
     #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
     { MP_ROM_QSTR(MP_QSTR_SEC_DPP), MP_ROM_INT(WIFI_AUTH_DPP) },
     #endif
+    #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+    { MP_ROM_QSTR(MP_QSTR_SEC_WPA3_ENTERPRISE), MP_ROM_INT(WIFI_AUTH_WPA3_ENTERPRISE) },
+    { MP_ROM_QSTR(MP_QSTR_SEC_WPA2_WPA3_ENTERPRISE), MP_ROM_INT(WIFI_AUTH_WPA2_WPA3_ENTERPRISE) },
+    #endif
 
     { MP_ROM_QSTR(MP_QSTR_PM_NONE), MP_ROM_INT(WIFI_PS_NONE) },
     { MP_ROM_QSTR(MP_QSTR_PM_PERFORMANCE), MP_ROM_INT(WIFI_PS_MIN_MODEM) },
```

DvdGiessen (changes requested):
I made the same changes locally to compile against IDF v5.4; found this PR so adding a few comments here instead of opening a new one.

I think this commit/PR should probably be referring to v5.4 instead of v5.5, since the latter is not out yet.

py/persistentcode.c
```diff
@@ -402,7 +402,7 @@ static mp_raw_code_t *load_raw_code(mp_reader_t *reader, mp_module_context_t *co
 
     #if MICROPY_EMIT_MACHINE_CODE
     } else {
-        const uint8_t *prelude_ptr;
+        const uint8_t *prelude_ptr = NULL;
```
dpgeorge, line 405: This change is unrelated to this PR, and shouldn't be needed.

dpgeorge:
@IhorNehrutsa when you open a pull request, can you please make sure you fill in the GitHub PR template, with a summary, testing that you did, and a list of trade-offs/alternatives.  Otherwise we may just close the PR.

py/persistentcode.c
DvdGiessen: This change is required to prevent an error while building against IDFv5.4 with a compiler error; perhaps a change in the compiler version included used by the IDF caused this issue to be exposed?

```
[1325/1467] Building C object esp-idf/main_esp32s3/CMakeFiles/__idf_main_esp32s3.dir/build/micropython/py/persistentcode.c.obj
FAILED: esp-idf/main_esp32s3/CMakeFiles/__idf_main_esp32s3.dir/build/micropython/py/persistentcode.c.obj (...)
/build/micropython/py/persistentcode.c: In function 'load_raw_code':
/build/micropython/py/persistentcode.c:439:38: error: 'prelude_ptr' may be used uninitialized [-Werror=maybe-uninitialized]
  439 |                 children[n_children] = (void *)prelude_ptr;
      |                 ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/build/micropython/py/persistentcode.c:405:24: note: 'prelude_ptr' was declared here
  405 |         const uint8_t *prelude_ptr; // = NULL;
      |                        ^~~~~~~~~~~
cc1: some warnings being treated as errors
(...)
ninja failed with exit code 1
```

dpgeorge: > Although I agree it _shouldn't_ be needed; reading it myself it appears correct regardless of which ifdef block is used. Not sure why the compiler disagrees.

Yeah, it looks like a compiler optimisation bug here.

I'm using gcc 14.2.1 and arm-none-eabi-gcc 14.2.0 here, and neither of them complain (with `MICROPY_EMIT_NATIVE_PRELUDE_SEPARATE_FROM_MACHINE_CODE` enabled).

ports/esp32/network_wlan.c
IhorNehrutsa: Thanks, done.

IhorNehrutsa:
@dpgeorge 
> @IhorNehrutsaколи ви відкриваєте запит на отримання, переконайтеся, що ви заповнюєте шаблон GitHub PR із підсумком, тестуванням, яке ви провели та списком компромісів/альтернатив. Інакше ми можемо просто закрити PR.

The latest line in the template is
> Delete this heading if not relevant (i.e. small fixes) -->
It seems that these small fixes are described in the PR/commit title.

Thanks.

py/persistentcode.c
IhorNehrutsa: In the end, are we forced to leave this fix as is?

dpgeorge:
> The latest line in the template is
> 
> > Delete this heading if not relevant (i.e. small fixes) -->

That line is only referring to the `Trade-offs and Alternatives` heading, not the entire template.

Please can you update the top post of this PR with the following template filled out:
```
### Summary

<!-- Explain the reason for making this change. What problem does the pull request
     solve, or what improvement does it add? Add links if relevant. -->


### Testing

<!-- Explain what testing you did, and on which boards/ports. If there are
     boards or ports that you couldn't test, please mention this here as well.

     If you leave this empty then your Pull Request may be closed. -->
```

DvdGiessen (commented):
Combined changes seem correct to me, but the commits can perhaps be organized to provide a bit more context to future readers.

- The first commit message still mentions this is for IDFv5.5 which isn't out yet; I think it should only mention v5.4?
- The other two commits could probably be squashed into the first one; they belong together after all.
- It might make sense to split out the change to `py/persistentcode.c` into a separate commit and add a short comment that this is to work around a compiler optimization bug in the compiler used by ESP-IDFv5.4. That way it'll be clear why that change is there, instead of being grouped in with the ESP32 port-specific changes. The change itself seems harmless enough to me; and with a clear separate commit documenting the reason for it I think it should be alright.

IhorNehrutsa:
@DvdGiessen 
Thanks for the detailed review.

py/persistentcode.c
dpgeorge: Yes, we can leave this change in.  It doesn't affect any other ports (the code size diff is zero for them all).

ports/esp32/network_ppp.c
```diff
@@ -100,7 +100,7 @@ static mp_obj_t ppp_make_new(mp_obj_t stream) {
 }
 MP_DEFINE_CONST_FUN_OBJ_1(esp_network_ppp_make_new_obj, ppp_make_new);
 
-static u32_t ppp_output_callback(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx) {
+static u32_t ppp_output_callback(ppp_pcb *pcb, const void *data, u32_t len, void *ctx) {
```
dpgeorge, line 103: This change now gives a compiler warning for IDF 5.2.2 (at least, probably also 5.3).  So I suggest having something like:
```c
#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
static u32_t ppp_output_callback(ppp_pcb *pcb, const void *data, u32_t len, void *ctx)
#else
static u32_t ppp_output_callback(ppp_pcb *pcb, u8_t *data, u32_t len, void *ctx)
#endif
{
    ...
```

IhorNehrutsa: Done

dpgeorge (approved):

dpgeorge:
I updated the esp32 README to note that IDF v5.4 is now supported.

IhorNehrutsa:
Thanks, Daniel.
Thanks, Damien.

PR #16549, opened 2025-01-08
esp32: Makefile, remove submodules sideeffect.

hmaerki:
### Summary

This PR removes an unwanted sideffect of
`make -C micropython/ports/esp32 submodules`
versus
`make -C micropython/ports/esp32 submodules BOARD=ESP32_GENERIC_S3`

This sideffect broke the build of ESP32_GENERIC_S3 when running on docker.io/espressif/idf:v5.2.2.

### Testing

This compile ok:
```bash
docker run -t docker.io/espressif/idf:v5.2.2 bash -c "git clone https://github.com/micropython/micropython.git && make -C micropython/mpy-cross && make -C micropython/ports/esp32 submodules && make -C micropython/ports/esp32 BOARD=ESP32_GENERIC_S3"
```

This compile fails (added: BOARD=ESP32_GENERIC_S3):
```bash
docker run -t docker.io/espressif/idf:v5.2.2 bash -c "git clone https://github.com/micropython/micropython.git && make -C micropython/mpy-cross && make -C micropython/ports/esp32 submodules BOARD=ESP32_GENERIC_S3 && make -C micropython/ports/esp32 BOARD=ESP32_GENERIC_S3"
```

Error message
```text
CMake Error at /opt/esp/idf/tools/cmake/build.cmake:544 (message):
  ERROR: Some components (espressif/tinyusb) in the "managed_components"
  directory were modified on the disk since the last run of the CMake.
  Content of this directory is managed automatically.
```

### Subfailure in cmake

In `ports/esp32/Makefile`

```Makefile
submodules:
	@GIT_SUBMODULES=$$(idf.py $(IDFPY_FLAGS) -B $(BUILD)/submodules -D ECHO_SUBMODULES=1 build 2>&1 | \
	                  grep '^GIT_SUBMODULES=' | cut -d= -f2); \
	rm -rf ./managed_components; \
	$(MAKE) -f ../../py/mkrules.mk GIT_SUBMODULES="$${GIT_SUBMODULES}" submodules
```

Here, `idf.py` crashes silently with

```text
CMake Error at /opt/esp/idf/tools/cmake/component.cmake:313 (message):
  Include directory '/home/maerki/2025-01-08b_espressif/ori/lib/tinyusb/src'
  is not a directory.
```

This error occurs using `docker.io/espressif/idf:v5.2.2` and the current version `hmaerki/build-micropython-esp32:v5.2.2`.

The second commit in this MR fixes this.

[...]

PR #16581, opened 2025-01-14
esp32: Fix component manager issue with "make submodules"

projectgus:
### Summary

This is an alternative version of #16549. I was testing it at @hmaerki's request and ended up with a slightly different approach.

The underlying bug is that some older versions of IDF Component Manager (definitely 1.4.2, maybe others) will leave the build and `managed_components` directories in an inconsistent state if CMake exits early. Running `make submodules` causes this early exit from CMake. The `make submodules` step appears to succeed, but a subsequent build with the same build directory fails with an error similar to:

```
CMake Error at /opt/esp/idf/tools/cmake/build.cmake:544 (message):
  ERROR: Some components (espressif/tinyusb) in the "managed_components"
  directory were modified on the disk since the last run of the CMake.
  Content of this directory is managed automatically.

  If you want to keep the changes, you can move the directory with the
  component to the "components"directory of your project.

  I.E.  for "espressif__tinyusb" run:

  mv /micropython/ports/esp32/managed_components/espressif__tinyusb
  /micropython/ports/esp32/components/espressif__tinyusb

  Or, if you want to discard the changes remove the ".component_hash" file
  from the component's directory.

  I.E.  for "espressif__tinyusb" run:

  rm
  /micropython/ports/esp32/managed_components/espressif__tinyusb/.component_hash


Call Stack (most recent call first):
  /opt/esp/idf/tools/cmake/project.cmake:605 (idf_build_process)
  CMakeLists.txt:68 (project)

```

Fix implemented as three commits:

1. Disable the component manager during the run of `idf.py` that collects the `GIT_SUBMODULES` list. This still completes successfully, and is a bit less heavy-handed (and faster) than deleting the managed_components directory afterwards.
2. Don't add the TinyUSB source files to a build which is only echoing submodules. This check is already applied elsewhere, was missed here.
3. As @hmaerki suggested in the linked PR, when the `GIT_SUBMODULES` run of CMake fails then it's possible for execution to appear to complete cleanly, but with an empty list of submodules. Then it fails at some later step. Add a flag to have it fail immediately in these cases.

### Testing

The latest IDF component manager doesn't exhibit the bug, so similar to @hmaerki I ran the build in a clean container. For speed I cross-mounted my local .git directory and then cloned from it:

```
podman run --rm -t -v "/home/gus/ry/george/micropython/.git:/gitdir:ro" -t docker.io/espressif/idf:v5.2.2 bash -c "git clone -b BRANCHNAME /gitdir micropython && make -j16 -C micropython/mpy-cross && make -C micropython/ports/esp32 submodules BOARD=ESP32_GENERIC_S3 V=1 && make -C micropython/ports/esp32 BOARD=ESP32_GENERIC_S3"
```

This command fails with the error above when BRANCHNAME is master but succeeds for this branch.


projectgus:
@hmaerki I was testing your PR and ended up finding another approach. Can you please check if this works for you?

ports/rp2/Makefile
```diff
@@ -76,4 +76,4 @@ submodules:
 	$(MAKE) -f ../../py/mkrules.mk GIT_SUBMODULES="lib/pico-sdk" submodules
 	@GIT_SUBMODULES=$$(cmake -B $(BUILD)/submodules -DECHO_SUBMODULES=1 ${CMAKE_ARGS} -S . 2>&1 | \
 	                  grep '^GIT_SUBMODULES=' | cut -d= -f2); \
-	$(MAKE) -f ../../py/mkrules.mk GIT_SUBMODULES="$${GIT_SUBMODULES}" submodules
+	$(MAKE) -f ../../py/mkrules.mk GIT_SUBMODULES="$${GIT_SUBMODULES}" FAIL_IF_NO_SUBMODULES=1 submodules
```
dpgeorge, line 79: Suggest using `GIT_SUBMODULES_FAIL_IF_EMPTY` as a variable name.

hmaerki:
> @hmaerki I was testing your PR and ended up finding another approach. Can you please check if this works for you?

I like your PR - thanks a lot!

I successfully tested (with and without your fix):
* `docker run -t docker.io/espressif/idf:v5.2.2 bash -c "git clone https://github.com/projectgus/micropython.git && git -C micropython checkout bugfix/esp32_components_submodules && make -C micropython/mpy-cross && make -C micropython/ports/esp32 submodules BOARD=ESP32_GENERIC_S3 && make -C micropython/ports/esp32 BOARD=ESP32_GENERIC_S3"`
* `mpbuild build ESP32_GENERIC_S3`

I will close now my initial https://github.com/micropython/micropython/pull/16549 in favor of this PR.

dpgeorge (approved):
Thanks @hmaerki for testing this PR.  Let's merge it.

PR #16582, opened 2025-01-14
py/asmarm: Make all SABRELITE tests pass.

agatti:
### Summary

This PR adds fixes for native code generation issues that made `tests/extmod/vfs_rom.py` and `tests/float/math_domain.py` fail when ran through the native emitter.

The commits in this PR address the same issue encountered in #15937 with regard to the Arm code generator, namely not checking for immediates' range when embedding them into opcodes.

The heap size for the Qemu port was also increased from 120KiB to 140KiB, as code generated for `tests/extmod/vfs_rom.py` would not fit into the available heap during a full test run.  This is due to the fact that no compressed opcodes are generated and thus a much larger block is needed.

### Testing

The full test suite passes when ran under Qemu with `make -C mpy-cross ; make -C ports/qemu BOARD=SABRELITE RUN_TESTS_EXTRA='--via-mpy --emit native --mpy-cross-flags="-march=armv6"' test`.

### DIFF
--- a/ports/qemu/Makefile
+++ b/ports/qemu/Makefile
@@ -19,7 +19,7 @@ QSTR_DEFS = qstrdefsport.h
 MICROPY_ROM_TEXT_COMPRESSION ?= 1
 
 ifeq ($(QEMU_ARCH),arm)
-MICROPY_HEAP_SIZE ?= 122880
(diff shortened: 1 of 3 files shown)


dpgeorge (approved):

dpgeorge:
Thank you!

PR #16583, opened 2025-01-14
qemu/boards: Change boards to use a subdirectory like other ports.

dpgeorge:
### Summary

This commit moves `<board>.mk` to `<board>/mpconfigboard.mk` for all qemu boards, making it the same as other bare-metal ports (same board directory layout, same way to define a board).

For example, it makes way to add a potential `mpconfigboard.h` file for each board, if that's ever needed in the future.

### Testing

This should be fully tested by CI.


mattytrentini:
Looks good! I tested MPS2_AN385 and SABRELITE; they worked exactly as expected. 

Not even any documentation updates required.

mattytrentini (approved):

dpgeorge:
Thanks for the review!

PR #16584, opened 2025-01-14
esp32/boards: Remove remaining "id" entries from board.json.

dpgeorge:
### Summary

This entry was originally used to override the firmware filenames generated by the build server, but these days all filenames should match the board directory name.  So, remove the "id" entry and let the default be used.

This is a follow-up to 1a99f74063569df0927e1ada0256059fcdef128c (these three boards were added after that change).

### Testing

No testing has been done.  The change is hopefully self evident.

### DIFF
--- a/ports/esp32/boards/ESP32_GENERIC_C6/board.json
+++ b/ports/esp32/boards/ESP32_GENERIC_C6/board.json
@@ -7,7 +7,6 @@
         "BLE",
         "WiFi"
     ],
-    "id": "esp32c6",
     "images": [
         "esp32c6_devkitmini.jpg"
     ],
--- a/ports/esp32/boards/UM_NANOS3/board.json
+++ b/ports/esp32/boards/UM_NANOS3/board.json
@@ -13,7 +13,6 @@
     "features_non_filterable": [
         "TinyPICO Nano Compatible"
     ],
-    "id": "nanos3",
     "images": [
         "unexpectedmaker_nanos3.jpg"
     ],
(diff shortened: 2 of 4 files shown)


dpgeorge:
This was triggered by https://github.com/micropython/micropython/pull/16535#issuecomment-2574491648

projectgus (approved):
Oh, I didn't notice this! Yes, this is better :grin: 

It looks like `"id"` is not used by other ports, either. Can we also remove support from `build-board.sh`, so that it can't creep back in?

projectgus:
Heads up for @UnexpectedMaker as well: no action required on your part after all.

dpgeorge:
> It looks like `"id"` is not used by other ports, either. Can we also remove support from `build-board.sh`, so that it can't creep back in?

OK, I've now done that.

But note that you can still define "id", it'll just be overwritten (similar to "port" and "build" entries which are always populated/overwritten with defaults).

mattytrentini:
Tangentially related to this; A long time ago I looked at using [json-schema](https://json-schema.org) to enforce the build.json format...is that something that would be useful? 

(Additional checks - size of images, valid urls etc - have already begun to be integrated into [mpbuild](https://github.com/mattytrentini/mpbuild) too.)

dpgeorge:
> Tangentially related to this; A long time ago I looked at using [json-schema](https://json-schema.org) to enforce the build.json format...is that something that would be useful?

If a check can be added to the CI here with minimal overhead and maintenance effort, then yes that could be good to have.

PR #12272, opened 2023-08-21
stm32/af_csv_gen: Add tool to generate af csv files from STM32_open_pin_data.

andrewleech:
This PR adds a new tool to automatically generate the alternate functions csv from official STM xml resources.

PR #16587, opened 2025-01-15
ports/stm32/boards: Add F427 AF file.

iabdalkader:


dpgeorge:
How did you generate these files, and verify they are correct?

Can you please reformat them to match the style of the other af.csv files, which is now fixed width columns, eg:
af_csv

iabdalkader:
> How did you generate these files, and verify they are correct?

@dpgeorge As far as I remember these were based on other AF files, and modified to remove missing functions. The only change for 427 right now is that doesn't have an LCD. I double checked the latest datasheet and the rest of the functions should exist for both 427 and 429. As for the 765, it was going to end up being identical to 767 so I removed it.
I'm not sure why, but both files had more missing functions, perhaps they were based on old datasheets that no longer exist.

andrewleech:
If you'd like to double check the af and/or automate the formatting this PR is for a tool that generates stm32 AF files https://github.com/micropython/micropython/pull/12272

iabdalkader:
@andrewleech Thanks! This should be very useful for future MCUs.

dpgeorge (approved):

PR #16591, opened 2025-01-15
ports/mimxrt: Add MSC support.

iabdalkader:
### Summary

Add MSC support using internal flash storage or SD card. This has been sitting for a while in my fork so I figured I should send it upstream. Note this is disabled by default, and can be enabled by boards if needed.

### Testing
Tested with:

- MIXRT1020_EVK, Teensy 4.1 and MIMXRT1176_EVK
- RT106x based board.

### DIFF
--- a/ports/mimxrt/main.c
+++ b/ports/mimxrt/main.c
@@ -39,6 +39,7 @@
 #include "led.h"
 #include "pendsv.h"
 #include "modmachine.h"
+#include "modmimxrt.h"
 
 #if MICROPY_PY_LWIP
 #include "lwip/init.h"
@@ -55,6 +56,7 @@
 
 #include "systick.h"
 #include "extmod/modnetwork.h"
+#include "extmod/vfs.h"
 
 extern uint8_t _sstack, _estack, _gc_heap_start, _gc_heap_end;
 
@@ -112,6 +114,19 @@ int main(void) {
 
         // Execute _boot.py to set up the filesystem.
         pyexec_frozen_module("_boot.py", false);
+
+        #if MICROPY_HW_USB_MSC
+        // Set the USB medium to flash block device.
+        mimxrt_msc_medium = &mimxrt_flash_type;
+
+        #if MICROPY_PY_MACHINE_SDCARD
+        const char *path = "/sdcard";
+        // If SD is mounted, set the USB medium to SD.
+        if (mp_vfs_lookup_path(path, &path) != MP_VFS_NONE) {
+            mimxrt_msc_medium = &machine_sdcard_type;
+        }
+        #endif
+        #endif
 
         // Execute user scripts.
         int ret = pyexec_file_if_exists("boot.py");
--- a/ports/mimxrt/modmimxrt.c
+++ b/ports/mimxrt/modmimxrt.c
@@ -31,6 +31,9 @@
 static const mp_rom_map_elem_t mimxrt_module_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR___name__),            MP_ROM_QSTR(MP_QSTR_mimxrt) },
     { MP_ROM_QSTR(MP_QSTR_Flash),               MP_ROM_PTR(&mimxrt_flash_type) },
+    #if MICROPY_HW_USB_MSC
+    { MP_ROM_QSTR(MP_QSTR_MSC),                 MP_ROM_TRUE },
+    #endif
 };
 static MP_DEFINE_CONST_DICT(mimxrt_module_globals, mimxrt_module_globals_table);
 
--- a/ports/mimxrt/msc_disk.c
+++ b/ports/mimxrt/msc_disk.c
@@ -1,0 +1,153 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020-2021 Damien P. George
+ * Copyright (c) 2024-2025 Ibrahim Abdelkader <iabdalkader@openmv.io>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+#include "tusb.h"
+#if CFG_TUD_MSC
+#include "flash.h"
+#include BOARD_FLASH_OPS_HEADER_H
+#include "stdlib.h"
+#include "modmimxrt.h"
+#if MICROPY_PY_MACHINE_SDCARD
+#include "sdcard.h"
+
+#ifndef MICROPY_HW_SDCARD_SDMMC
+#define MICROPY_HW_SDCARD_SDMMC (1)
+#endif
+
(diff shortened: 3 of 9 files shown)


iabdalkader:
> complaining about the symbol MICROPY_HW_SDCARD_SDMMC not being set. Maybe you have that set in your environment

Yes, this just lets the msc disk know which SD is used for the block device if you have more than one, and since it's board-specific you need to define `MICROPY_HW_SDCARD_SDMMC` if you enable MSC.

> The Teensy's file system is littlefs, but a block device should be transparent. Is your PR by some means restricted to FAT?

You probably need to match the flash sector size, maybe try `progsize=4096`?

robert-hh:
Not mounting LFS It was a problem with my PC. After reboot it worked. `MICROPY_HW_SDCARD_SDMMC` has to be set to 1. Setting it to 2 causes other compile error.

Maybe you get yourself MIMXRT boards and test the submission before posting a PR.

iabdalkader:
> MICROPY_HW_SDCARD_SDMMC has to be set to 1. Setting it to 2 causes other compile error.

Is this unexpected? If you only have one SD card, you can't set the ID to 2.

> Maybe you get yourself MIMXRT boards and test the submission before posting a PR.

As I've mentioned I have been using this code in production for year, so it's been tested, just never with an upstream board.

robert-hh:
I did a few more board tests after changing mpconfigport.h to:

```
// Enable USB Mass Storage with FatFS filesystem.
#ifndef MICROPY_HW_USB_MSC
#define MICROPY_HW_USB_MSC                  (1)
#if MICROPY_PY_MACHINE_SDCARD && !defined(MICROPY_HW_SDCARD_SDMMC)
#define MICROPY_HW_SDCARD_SDMMC             (1)
#endif
#endif
```

MIXRT1020_EVK, Teensy 4.1 and MIMXRT1176_EVK worked fine. It is a good idea to expose the inserted SD card.

MIMXRT1010 and MIMXRT1015 fail for two reasons:
1. Overflow of the m_dtcm segment. by 2800 resp 1700 bytes. It may be possible to reorganize the loader file for that,
2. Access to `machine_sdcard_type` in main.py. These devices do not support SD cards. I've changed that section to:
```
        #if MICROPY_HW_USB_MSC
        // Set the USB medium to flash block device.
        mimxrt_msc_medium = &mimxrt_flash_type;

        #if MICROPY_PY_MACHINE_SDCARD
        const char *path = "/sdcard";
        // If SD is mounted, set the USB medium to SD.
        if (mp_vfs_lookup_path(path, &path) != MP_VFS_NONE) {
            mimxrt_msc_medium = &machine_sdcard_type;
        }
        #endif
        #endif
```

iabdalkader:
@robert-hh I've incorporated both changes. As for the memory overflow, this should be fixed for individual boards if/when they enable this feature.

robert-hh:
I see that you disable MSC by default.
The memory overflow at least for the 1011 is caused by a buffer with a need 2k alignment, which is placed by the loader just after a 2k boundary. And then there is a ~2k gap. On option would be to reduce the stack size, which is 20k at the moment. The other would be to re-arrange the objects in .bss manually. I know that I did that once, but I forgot how.

iabdalkader:
> I see that you disable MSC by default.

Yes. If you like you can enable it by default or for certain boards if/when this PR gets merged.

As for the memory, I'd lower the stack, 16KBs is good probably good enough.

ports/mimxrt/modmimxrt.c
```diff
@@ -31,6 +31,9 @@
 static const mp_rom_map_elem_t mimxrt_module_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR___name__),            MP_ROM_QSTR(MP_QSTR_mimxrt) },
     { MP_ROM_QSTR(MP_QSTR_Flash),               MP_ROM_PTR(&mimxrt_flash_type) },
+    #if MICROPY_HW_USB_MSC
+    { MP_ROM_QSTR(MP_QSTR_MSC),                 MP_ROM_INT(1) },
```
dpgeorge, line 35: Maybe `MP_ROM_TRUE` instead of `MP_ROM_INT(1)`?

ports/mimxrt/main.c
```diff
@@ -113,6 +115,19 @@ int main(void) {
...
 
+        #if MICROPY_HW_USB_MSC
+        // Set the USB medium to flash block device.
+        mimxrt_msc_medium = &mimxrt_flash_type;
+
+        #if MICROPY_PY_MACHINE_SDCARD
+        const char *path = "/sdcard";
+        // If SD is mounted, set the USB medium to SD.
+        if (mp_vfs_lookup_path(path, &path) != MP_VFS_NONE) {
+            mimxrt_msc_medium = &machine_sdcard_type;
```
dpgeorge, line 126: It looks like this code can switch the medium type on soft reset, which would lead to inconsistencies with what the host PC thinks is attached.  Not sure there's any good way to improve that though.

ports/mimxrt/msc_disk.c
```diff
@@ -0,0 +1,152 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2020-2021 Damien P. George
+ * Copyright (c) 2024-2045 Ibrahim Abdelkader <iabdalkader@openmv.io>
```
dpgeorge, line 5: Year typo, should be 2025.

ports/mimxrt/msc_disk.c
```diff
@@ -0,0 +1,152 @@
...
+#define MSC_SDCARD_INDEX    (MICROPY_HW_SDCARD_SDMMC - 1)
+#endif
+
+// This implementation does Not support Flash sector caching.
+// MICROPY_FATFS_MAX_SS  must be identical to SECTOR_SIZE_BYTES
+#define BLOCK_SIZE          (SECTOR_SIZE_BYTES)
+#define BLOCK_COUNT         (MICROPY_HW_FLASH_STORAGE_BYTES / BLOCK_SIZE)
+#define FLASH_BASE_ADDR     (MICROPY_HW_FLASH_STORAGE_BASE)
+
+bool ejected = false;
```
dpgeorge, line 43: Make it a static variable, and maybe called `msc_ejected`?

ports/mimxrt/msc_disk.c
```diff
@@ -0,0 +1,152 @@
...
+
+const mp_obj_type_t *mimxrt_msc_medium = NULL;
+
+// Invoked when received SCSI_CMD_INQUIRY
+// Application fill vendor id, product id and revision with string up to 8, 16, 4 characters respectively
+void tud_msc_inquiry_cb(uint8_t lun, uint8_t vendor_id[8], uint8_t product_id[16], uint8_t product_rev[4]) {
```
dpgeorge, line 49: I suggest this function copies `ports/rp2/msc_disk.c` and allows the VID/PID/REV to be configured by a board.

ports/mimxrt/msc_disk.c
iabdalkader: Future-proofing 👍🏼

dpgeorge:
@robert-hh are you happy with this PR now?

iabdalkader:
> will you do that in mimxrt1011.ld and mimxrt1015.ld? It may be useful when adding a new feature to at least build the firmware for all applicable boards with that feature enabled.

Are you saying we should enable MSC for these boards by default?

robert-hh:
> Are you saying we should enable MSC for these boards by default?

Actually no. But if it's disabled, there should be no further hiccup. I enabled MSC by default and ran a build-all script with the 1011/1015 stack changed. All boards build.

The Teensy4.1 board has a MIMXRT1062 MCU, and it is already built in the CI. So you could use that one. But if you want to add a board, the MIMXRT1010 would be a good choice, because that one is always a candidate for trouble.

dpgeorge:
I don't want to enable MSC by default.  It's just that it should build without error if a user enables it themselves.

iabdalkader:
@dpgeorge You can just drop commit number 3 which enables it for some boards, and commit number 4 which adds 1060 to CI. However, I'd leave 1060 in CI as it tests more things.

iabdalkader:
> I don't want to enable MSC by default. It's just that it should build without error if a user enables it themselves.

@dpgeorge @robert-hh On second thought, if don't add at least 1 board with MSC enabled, we'll never notice if this ever gets broken by any change. I only care about the core functionality, but this is worth mentioning.

dpgeorge:
> On second thought, if we don't add at least 1 board with MSC enabled, we'll never notice if this ever gets broken by any change

You can build with MSC enabled just under CI using something like `make BOARD=... CFLAGS_EXRTA=-DMICROPY_HW_USB_MSC=1`.

iabdalkader:
Done. Seems to build fine.

dpgeorge (approved):
Tested with TEENSY40 and enabling MSC.  It works.

robert-hh:
As a note: This feature works also with the lfs file system at the board and littlefs-fuse at the PC.

mattytrentini:
> As a note: This feature works also with the lfs file system at the board and littlefs-fuse at the PC.

Oh nice, I've been meaning to try this - that's great to know!

iabdalkader:
Unrelated question, but does anyone know what the minimum LFSV2 volume size is? Or is there no such limitation ? For example, for FatFS it's 512 * 128 = 64K.

robert-hh:
Lfs uses at least 5 blocks (2+2+file data), with the minimal block size being 104 bytes. So yes, there is a minimal size, but that is rather small.

iabdalkader:
Amazing! Thanks @robert-hh

Issue #13439, opened 2024-01-16
rp2 PIO StateMachine jump table to produce instructions


jrmoserbaltimore:
An enhancement to the Pi Pico StateMachine, add a method producing `jmp` instructions referencing labels in the state machine.  This could then be used to direct a jump to various places without decisions in the PIO program.  This is already possible, just semi-difficult:  as far as I know, there is no way to get the address of a label besides manually counting instructions.

If this is already doable, it's not well documented that I could find, so this becomes just a documentation bug.

For example, consider communicating with the General Instruments AY-3-8930.  This has a simple enough interface and, for the most part, requires writing 8 bits to an address/data bus and 2 bits to a control bus indicating whether the bus is idle, read, write, or address.  The system can communicate with this through the TX FIFO by writing words of `[data][operation][address]`, each 8 bits, but this gets complex:

* The program first writes the address, then must evaluate [operation] to determine if this is read or write
* The program must use `out` to move the contents of the ISR into one register, and then use several `set` and `jmp` instructions to decide where in the program to jump

Instead of these 24 bits, one could use the whole 32 bits in the FIFO, making [operation] a `jmp` instruction:

* The program first writes the address, then sets `x` to 1
* On the next bus clock (next loop of the program), a `jmp (x_dec,"handle data")` directs flow away from setting the address, instead to data handling
* The data handling code uses `out (exec,16)` to jump to the appropriate point in the code

This replaces the coded jump table with a single `out` instruction.

To get the `jmp` instruction, the StateMachine object would provide a member function to produce a `jmp` instruction using the particular label as a destination.

As for the example, this is the output section of a write-only PIO driver for the AY-3-8930.  This uses 16-bit input of an address followed by data, and always writes.

```python
# Set address
label("handle_output")
jmp(x_dec, "set_data")
out(pins,8) .side(0b11)
set(x, 0b1)
jmp("handle_clock")

# Set data
label("set_data")
out(pins,8) .side(0b10) [1]
jmp("handle_clock")
```

To expand this to read/write, I'd need to send another flag, copy it into y, and evaluate it against x:

```python
# Set address
label("handle_output")
jmp(x_dec, "getset_data")
out(pins,8) .side(0b11)
set(x, 0b1)
jmp("handle_clock")

label("getset_data")
out(y,8)
set(x,0b0)
jmp(x_not_y, "get_data")

# Set data
label("set_data")
out(pins,8) .side(0b10) [1]
jmp("handle_clock")

# Get data
[...]

Issue #13248, opened 2023-12-21
rp2 PIO state machine pin config enhancements


nludban:
Several annoyances with trying to configure pins for PIO, lumping them into one ticket since they're minor and overlapping in the implementation.

The rp2040 datasheet, 3.5.6 GPIO Mapping, specifies pin base + count are "wrapping after GPIO31".  The rp2 port does not currently support this.  Since this is only executed once, a loop to configure one pin at a time would be fine:
https://github.com/micropython/micropython/blob/5d28bb4adb24094917038a506248ff7fb351ac44/ports/rp2/rp2_pio.c#L214-L215
https://github.com/micropython/micropython/blob/5d28bb4adb24094917038a506248ff7fb351ac44/ports/rp2/rp2_pio.c#L227

The rp2 port requires initial value and direction to be specified for every pin in every group.  When the pin groups (in, out, set, etc) overlap, initial value and direction must be duplicated and there are no checks for conflicts.  It may also be desirable for multiple state machines to use the same pin, one for output and one for input - seems to work, at least when on the same PIO.  Feature request is to allow `None` for the initial setting, the user is then responsible for making sure the configuration happens.

Pull (up and down) should be configurable in the same place as value and direction.  Workaround is to instantiate a `Pin` or poke the config registers directly.

Open drain can be emulated in the rp2040 with value=0 and changing the pin direction to drive or float the output.  It also needs a pull up to be configured.  Request is to be able to specify `PINDIR_OD_LOW` or `PINDIR_OD_HIGH` so the code is all in one place with obvious intent.

Issue #10210, opened 2022-12-13
RP2: Make PIO EXECCTRL_STATUS_SEL settable through API


pigrew:
Please add setting EXECCTRL_STATUS_SEL to the initializer of a PIO state machine for the RP2040.

I'd imagine this would be something like:

`StateMachine.init(program,...,status_sel=PIO.TX_FIFO, status_n=3)`

The new initialization routine could then call `sm_config_set_mov_status(...)` from the `pico-sdk`.

Alternatively, a new API function could be created. This seems less in line with the current API, but could be more flexible since it allows dynamically changing the parameters:

```
sm = StateMachine.init(...)
sm.set_mov_status(status_sel: PIO.TX_FIFO, n: 3)
```

Thanks!

nludban:
+1 for the feature request.

Alternate workaround, I hacked the `execctrl` value in the program before initializing the state machine.  Less code, but more magic values:

https://github.com/nludban/rpi-pico-psst/blob/2711839c95f376afb1a5104e03d3a1429ba55e71/psst.py#L268-L277

PR #16593, opened 2025-01-15
PIO enhancements poc

nludban:
### Summary

This oversized PR is a proof of concept that adds a new `PIOProgram` class to replace functionality provided by the `asm_pio` decorator.  By cleanly splitting the definition of a PIO program from its deployment to a `StateMachine`, more advanced use cases are easier to implement in `rp2` and the application.  The basic idea is to encapsulate the semi-opaque `prog` array and `labels` so future customization can be done with methods rather than expanding the set of supported keyword args.  It could also provide a common path for deploying hex output from `pioasm`.

https://github.com/micropython/micropython/issues/11593 - implemented via parameters to `PIOProgram.init_state_machine`. `rp2.calc_pio_clock_dividers` is also provided, this supports testing, verifying the values being calculated, and allows the application to round the  values any number of ways before using them.  Note a change in behavior: `freq` was previously an upper bound (divisor calculations had an implicit floor), now it may round up if it reduces the error.

https://github.com/micropython/micropython/issues/13439 is implemented via `PIOProgram.encode`.  Two bugs are fixed in the handling of the global `_pio_funcs`: bound functions for the last `PIOASMEmit` were left in after use (a memory leak) and not all replaced/deleted by `asm_pio_encode` (leaving old directives in place).

The new `PIOProgram` will make https://github.com/micropython/micropython/issues/13248 and https://github.com/micropython/micropython/issues/10210 much cleaner.

### Testing

The example code in [this comment](https://github.com/micropython/micropython/issues/11593#issuecomment-2585491108) was translated to C and the tests pass on a pico.

Nothing else is tested yet as this is a proof of concept PR.

### DIFF
--- a/ports/rp2/modrp2.c
+++ b/ports/rp2/modrp2.c
@@ -91,6 +91,8 @@ static const mp_rom_map_elem_t rp2_module_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR_DMA),                 MP_ROM_PTR(&rp2_dma_type) },
     { MP_ROM_QSTR(MP_QSTR_bootsel_button),      MP_ROM_PTR(&rp2_bootsel_button_obj) },
 
+    { MP_ROM_QSTR(MP_QSTR_calc_pio_clock_dividers), MP_ROM_PTR(&rp2_calc_pio_clock_dividers_obj) },
+
     #if MICROPY_PY_NETWORK_CYW43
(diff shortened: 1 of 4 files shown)


Josverl:
Can you provide an example how this works from a python perspective?

Also I'm interessed to understand if and how this may help USEs, linters or type checkers handle the Python form of the PIO code better than the current implementation

PR #16598, opened 2025-01-16
all: Rename the "NORETURN" macro to "MP_NORETURN".

agatti:
### Summary

During the review process for #15803 it was noticed that there was a MicroPython attribute macro whose name did not start with the "MP_" prefix like the other ones.

This commit renames the `NORETURN` macro, indicating to the compiler that a function does not return, into `MP_NORETURN` to maintain the same naming convention of other similar macros.

To maintain compatibility with existing external code `NORETURN` is aliased to `MP_NORETURN`, but it is also marked as deprecated for MicroPython v2.

This changeset was created using a similar process to decf8e6a8bb940d5829ca3296790631fcece7b21 ("all: Remove the "STATIC" macro and just use "static" instead."), with no documentation or python scripts to change to reflect the new macro name.

```bash
git ls-files | grep -E '\.[ch]$' | xargs sed -Ei "s/(^| )NORETURN($| )/\1MP_NORETURN\2/"
```

should yield an identical result to this PR except for the `MP_NORETURN` definition itself, and the misaligned lines in `mpy-cross/mpconfigport.h` and `ports/windows/mpconfigport.h` - those were fixed by hand.

### Testing

The automated CI tests should not fail due to compilation errors related to this change, as they should build all ports except PIC16.

There may be some PRs that haven't yet been merged that may require an update.

Ideally there should be an automated process to flag those PRs and notify their authors, similar to what was done for the `STATIC` &rarr; `static` migration.  Assuming it was not a manual process, maybe that script can be reused for this change.

### DIFF
--- a/extmod/modmachine.c
+++ b/extmod/modmachine.c
@@ -45,11 +45,11 @@
 static void mp_machine_idle(void);
 
 #if MICROPY_PY_MACHINE_BOOTLOADER
-NORETURN void mp_machine_bootloader(size_t n_args, const mp_obj_t *args);
+MP_NORETURN void mp_machine_bootloader(size_t n_args, const mp_obj_t *args);
 #endif
 
 #if MICROPY_PY_MACHINE_RESET
-NORETURN static void mp_machine_reset(void);
+MP_NORETURN static void mp_machine_reset(void);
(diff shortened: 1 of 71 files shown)


dpgeorge (approved):
Thanks for doing this!

PR #16606, opened 2025-01-18
esp8266/Makefile: Fix local toolchain builds on recent Linux systems.

agatti:
### Summary

This commit fixes compilation for the ESP8266 port when using a local toolchain on relatively recent Linux systems.

The documentation asks the user to delete the esptool instance that comes with the toolchain, in favour of using the one provided by the system.  On Linux systems that are at least two years old (looking at the CI Ubuntu image as an example), the version of esptool installed with the package manager isn't called `esptool.py` but just `esptool`. The Makefile didn't take that into account and used `esptool.py` without checking if such a command exists, making builds fail.

Now preference is given to the `esptool` command, falling back to `esptool.py` only if the former command does not exist or it is not available to the current user, to maintain compatibility with old setups.

### Testing

The 8266 port was successfully built twice on Ubuntu 22.04 using a local toolchain with its provided `esptool` instance deleted.  Once with `esptool` provided by `apt-get`, and once with `esptool` provided by `pip` (using the CI scripts).

### DIFF
--- a/ports/esp8266/Makefile
+++ b/ports/esp8266/Makefile
@@ -185,6 +185,11 @@ else ifneq ($(shell cat $(CONFVARS_FILE)), $(FROZEN_MANIFEST) $(UART_OS))
 $(shell echo $(FROZEN_MANIFEST) $(UART_OS) > $(CONFVARS_FILE))
 endif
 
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
I tested it locally (where I don't have `esptool` command) and it works for me.

PR #16607, opened 2025-01-18
machine.Pin: Implement Pin.toggle() for three ports.

robert-hh:
### Summary

The ESP32, ESP8266 and CC3200 ports did not provide the method Pin.toggle(). This PR adds it. 

### Testing

Tested with:

- ESP32
- ESP32C3
- ESP32C6
- ESP32S2
- ESP32S3
- ESP8266
- WiPY 1 (CC3200)

### DIFF
--- a/docs/library/machine.Pin.rst
+++ b/docs/library/machine.Pin.rst
@@ -209,13 +209,13 @@ The following methods are not part of the core Pin API and only implemented on c
 
    Set pin to "0" output level.
 
-   Availability: nrf, rp2, stm32 ports.
+   Availability: mimxrt, nrf, renesas-ra, rp2, samd, stm32 ports.
 
 .. method:: Pin.high()
 
    Set pin to "1" output level.
 
-   Availability: nrf, rp2, stm32 ports.
+   Availability: mimxrt, nrf, renesas-ra, rp2, samd, stm32 ports.
 
 .. method:: Pin.mode([mode])
 
@@ -242,7 +242,7 @@ The following methods are not part of the core Pin API and only implemented on c
 
    Toggle output pin from "0" to "1" or vice-versa.
 
-   Availability: mimxrt, samd, rp2 ports.
+   Availability: cc3200, esp32, esp8266, mimxrt, rp2, samd ports.
 
 Constants
 ---------
--- a/ports/esp32/mphalport.h
+++ b/ports/esp32/mphalport.h
@@ -36,6 +36,7 @@
 #include "freertos/task.h"
 
 #include "driver/spi_master.h"
+#include "soc/gpio_reg.h"
 
 #define MICROPY_PLATFORM_VERSION "IDF" IDF_VER
 
@@ -132,6 +133,15 @@ static inline void mp_hal_pin_od_high(mp_hal_pin_obj_t pin) {
 }
 static inline int mp_hal_pin_read(mp_hal_pin_obj_t pin) {
     return gpio_get_level(pin);
+}
+static inline int mp_hal_pin_read_output(mp_hal_pin_obj_t pin) {
+    #if defined(GPIO_OUT1_REG)
+    return pin < 32
+        ? (*(uint32_t *)GPIO_OUT_REG >> pin) & 1
+        : (*(uint32_t *)GPIO_OUT1_REG >> (pin - 32)) & 1;
+    #else
+    return (*(uint32_t *)GPIO_OUT_REG >> pin) & 1;
+    #endif
 }
 static inline void mp_hal_pin_write(mp_hal_pin_obj_t pin, int v) {
     gpio_set_level(pin, v);
--- a/ports/esp8266/esp_mphal.h
+++ b/ports/esp8266/esp_mphal.h
@@ -100,6 +100,13 @@ void mp_hal_pin_open_drain(mp_hal_pin_obj_t pin);
         else { gpio_output_set(1 << (p), 0, 1 << (p), 0); } \
 } while (0)
 #define mp_hal_pin_read(p) pin_get(p)
+static inline int mp_hal_pin_read_output(mp_hal_pin_obj_t pin) {
+    if (pin >= 16) {
+        return READ_PERI_REG(RTC_GPIO_OUT) & 1;
+    } else {
+        return (GPIO_REG_READ(GPIO_OUT_ADDRESS) >> pin) & 1;
+    }
+}
 #define mp_hal_pin_write(p, v) pin_set((p), (v))
 
 void *ets_get_esf_buf_ctlblk(void);
--- a/ports/cc3200/mods/pybpin.c
+++ b/ports/cc3200/mods/pybpin.c
@@ -680,6 +680,20 @@ static mp_obj_t pin_value(size_t n_args, const mp_obj_t *args) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(pin_value_obj, 1, 2, pin_value);
 
+// pin.toggle()
+static mp_obj_t pin_toggle(mp_obj_t self_in) {
+    pin_obj_t *self = self_in;
+    if (self->value) {
+        self->value = 0;
+        MAP_GPIOPinWrite(self->port, self->bit, 0);
+    } else {
+        self->value = 1;
+        MAP_GPIOPinWrite(self->port, self->bit, self->bit);
+    }
+    return mp_const_none;
+}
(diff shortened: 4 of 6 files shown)


mattytrentini (approved):
LGTM (we should check toggle exists for all other ports too...)

robert-hh:
`grep -r QSTR_toggle *` at the current master returns:
```
mimxrt/machine_pin.c:    { MP_ROM_QSTR(MP_QSTR_toggle),  MP_ROM_PTR(&machine_pin_toggle_obj) },
mimxrt/machine_led.c:    {MP_ROM_QSTR(MP_QSTR_toggle), MP_ROM_PTR(&led_obj_toggle_obj)},
nrf/modules/board/led.c:    { MP_ROM_QSTR(MP_QSTR_toggle), MP_ROM_PTR(&led_obj_toggle_obj) },
pic16bit/modpybled.c:    { MP_ROM_QSTR(MP_QSTR_toggle), MP_ROM_PTR(&pyb_led_toggle_obj) },
renesas-ra/led.c:    { MP_ROM_QSTR(MP_QSTR_toggle), MP_ROM_PTR(&led_obj_toggle_obj) },
rp2/machine_pin.c:    { MP_ROM_QSTR(MP_QSTR_toggle), MP_ROM_PTR(&machine_pin_toggle_obj) },
samd/machine_pin.c:    { MP_ROM_QSTR(MP_QSTR_toggle), MP_ROM_PTR(&machine_pin_toggle_obj) },
stm32/led.c:    { MP_ROM_QSTR(MP_QSTR_toggle), MP_ROM_PTR(&led_obj_toggle_obj) },
```
So ports with a machine module missing `Pin.toggle()` are:
- esp32 (done with this PR)
- esp8266
- cc3200
- zephyr

I can try to implement it for ESP8266 and CC3200, even it it may be waste of time for the latter. I have no hardware with Zephyr loaded, and I never dealt with Zephyr. And it depends a lot from the way, the functions `gpio_pin_get_raw()` and `gpio_pin_set_raw()` are implemented and the pins are initialized for the respective hardware. So even if a Pin.toggle() was implemented using these two functions, it may not work.

robert-hh:
Pin.toggle() is added for ESP8266 and CC3200 as well.

mattytrentini:
Fantastic work @robert-hh!

robert-hh:
That's right

dpgeorge:
Note that the stm32 port does not have `Pin.toggle()` (only `LED.toggle()`).

But in general it's not easy to implement this function.  To make it work correctly you must be able to read the current *output* state of the pin, not the input state.

For the three ports that you implemented `toggle()` for, does it work correctly and invert the state of the output pin regardless of the voltage level on the pin?

dpgeorge:
> Checking the implementation for mimxrt and samd: there the output driver itself supports toggling, which is used by the code.

OK, thanks for checking.  I think we should only implement `Pin.toggle()` for ports that can do it properly like this.


> I have to check whether is is supported by the ESP32 hardware.

Please do.  And if we need to do raw register access to read back the current state, I think that's what we should do.

robert-hh:
> Please do. And if we need to do raw register access to read back the current state, I think that's what we should do.

Done. ESP32 and ESP8266 use the output register value, while the CC3200 port already stored the actual set output value at the pin object. So that could be used.

docs/library/machine.Pin.rst
```diff
@@ -242,7 +242,7 @@ The following methods are not part of the core Pin API and only implemented on c
 
    Toggle output pin from "0" to "1" or vice-versa.
 
-   Availability: mimxrt, samd, rp2 ports.
+   Availability: cc3200, esp32, esp8266, mimxrt, nrf, pic16bit, renesas-ra, rp2, samd, stm32 ports.
```
dpgeorge, line 245: nrf, pic16bit, renesas-ra and stm32 do not have the `Pin.toggle` method (but they do have `LED.toggle`, I think that's what you saw when you did the grep).  So please remove them from this list.

ports/esp32/mphalport.h
```diff
@@ -136,6 +137,16 @@ static inline int mp_hal_pin_read(mp_hal_pin_obj_t pin) {
 static inline void mp_hal_pin_write(mp_hal_pin_obj_t pin, int v) {
     gpio_set_level(pin, v);
 }
+static inline int mp_hal_pin_output_level(mp_hal_pin_obj_t pin) {
```
dpgeorge, line 140: I suggest calling this `mp_hal_pin_read_output`.

ports/esp32/mphalport.h
```diff
@@ -136,6 +137,16 @@ static inline int mp_hal_pin_read(mp_hal_pin_obj_t pin) {
...
+    #if defined(GPIO_OUT1_REG)
+    return (pin < 32
+        ? (*(uint32_t *)GPIO_OUT_REG & (1 << pin))
+        : (*(uint32_t *)GPIO_OUT1_REG & (1 << (pin - 32))))
+            ? 1 : 0;
+    #else
+    return (*(uint32_t *)GPIO_OUT_REG & (1 << pin)) ? 1 : 0;
```
dpgeorge, line 147: You should be able to do away with the `?:` operator by right shifting the register, eg (schematically):
```c
return (*GPIO_OUT_REG >> pin) & 1
```
(Same comment applies to the more complicated case when `GPIO_OUT1_REG` is defined.)

ports/esp8266/esp_mphal.h
```diff
@@ -101,6 +101,12 @@ void mp_hal_pin_open_drain(mp_hal_pin_obj_t pin);
 } while (0)
 #define mp_hal_pin_read(p) pin_get(p)
 #define mp_hal_pin_write(p, v) pin_set((p), (v))
-
+static inline int mp_hal_pin_output_value(mp_hal_pin_obj_t pin) {
+    if (pin >= 16) {
+        return READ_PERI_REG(RTC_GPIO_OUT) & 1;
+    } else {
+        return (GPIO_REG_READ(GPIO_OUT_ADDRESS) & (1 << pin)) ? 1 : 0;
```
dpgeorge, line 108: As above, should be able to do `(GPIO_REG_READ(GPIO_OUT_ADDRESS) >> pin) & 1`.

ports/esp8266/esp_mphal.h
```diff
@@ -101,6 +101,12 @@ void mp_hal_pin_open_drain(mp_hal_pin_obj_t pin);
 } while (0)
 #define mp_hal_pin_read(p) pin_get(p)
 #define mp_hal_pin_write(p, v) pin_set((p), (v))
-
+static inline int mp_hal_pin_output_value(mp_hal_pin_obj_t pin) {
```
dpgeorge, line 104: As above, I suggest calling this `mp_hal_pin_read_output`.

docs/library/machine.Pin.rst
robert-hh: Done.

ports/esp32/mphalport.h
robert-hh: Done.

ports/esp32/mphalport.h
robert-hh: Yes. That looks simpler. I wonder what the optimizer would have created with the previous code.

ports/esp8266/esp_mphal.h
robert-hh: Done

robert-hh:
All suggested comments implemented & tested. For ESP32 I tested the ESP32_GENERIC and ESP32_GENERIC_C3 board to cover all changed code.

ports/esp32/mphalport.h
```diff
@@ -136,6 +137,15 @@ static inline int mp_hal_pin_read(mp_hal_pin_obj_t pin) {
 static inline void mp_hal_pin_write(mp_hal_pin_obj_t pin, int v) {
     gpio_set_level(pin, v);
 }
+static inline int mp_hal_pin_read_output(mp_hal_pin_obj_t pin) {
+    #if defined(GPIO_OUT1_REG)
+    return pin < 32
+        ? (*(uint32_t *)GPIO_OUT_REG >> pin) & 1
+        : (*(uint32_t *)GPIO_OUT1_REG >> (pin - 32)) & 1;
+    #else
+    return *(uint32_t *)GPIO_OUT_REG >> pin & 1;
```
dpgeorge, line 146: Please put parenthesis around the `>>` operator and its operands, just to be certain the calculation is done with the correct precedence.

robert-hh: Changed an just recompiled to be sure it builds.

dpgeorge (approved):

Issue #16604, opened 2025-01-17
Emitted code epilogue block contains a superflous jump to the next available code address.


agatti:
### Port, board and/or hardware

mpy-cross built from git commit 6db29978ac8954f3686f9eb59dd71b55c3495456 (current master)

### MicroPython version

MicroPython v1.25.0-preview.216.g6db29978a.dirty on 2025-01-17; mpy-cross emitting mpy v6.3

### Reproduction

1. Run `mpy-cross -X emit=native -march=debug tests/basics/0prelim.py`
2. Look at the last part of the output for a jump to `label_0`.

This applies to any supported architecture with a native emitter, but `-march=debug` is used in the command line to make the problem visible without using a disassembler.

### Expected behaviour

There should be no empty jump being emitted at the end of a block.

### Observed behaviour

At the end of an emitted block of code this can be seen (taken from `tests/basics/0prelim.py`):

```
    ...
    jump(label_0)
label(label_0)
EXIT(0)
```

This applies also to functions that need a more involved cleanup procedure:

(`tests/basics/async_for.py`):

```
    ...
    jump(label_0)
    dead_code load(r_temp0, r_fun_table, 0)
    dead_code store(r_temp0, r_local2, 5)
    dead_code mov_reg_imm(r_temp0, 40=0x28)
    dead_code add(r_temp0, r_local2)
    dead_code store(r_temp0, r_local2, 2)
    dead_code mov_reg_imm(r_temp0, 0=0x0)
    dead_code mov_local_reg(local_3, r_temp0)
    dead_code jump(label_0)
label(label_0)
    call_ind(nlr_pop)
    mov_reg_local(r_ret, local_3)
EXIT(0)
```

(`tests/basics/array_micropython.py`):

```
    ...
    mov_local_reg(local_3, r_ret)
    jump(label_0)
label(label_0)
    mov_reg_local(r_arg1, local_6)
    call_ind(native_swap_globals)
    call_ind(nlr_pop)
    mov_reg_local(r_ret, local_3)
EXIT(0)
```

### Additional Information

This is not specific to `mpy-cross` as the same issue occurs when emitting a block of code at runtime as well.

### Code of Conduct

Yes, I agree

PR #16608, opened 2025-01-19
py/emitnative: Optimise native code unwinding jumps.

agatti:
### Summary

This PR introduces an optimisation for emitting unwinding jump blocks in native code.  Currently the native emitter will emit a jump to the block's unwinding code even though that jump is located right before the unwinding sequence itself.  So the unwind code sequence would look like this (as emitted by mpy-cross using "-march=debug"):

```
    ...
    jump(label_0)
label(label_0)
    ...
EXIT(0)
```

or:

```
    ...
    jump(label_0)
    dead_code ...
label(label_0)
    ...
EXIT(0)
```

With this change, the native emitter only emits jumps to the exit label (label_0 in the examples above) when unwinding if both the exception stack and the emitter stack are empty, as it means the jump is at the very end of the code block.

This fixes issue #16604.

### Testing

The full test suite was executed with `make BOARD=VIRT_RV32 RUN_TESTS_EXTRA='--via-mpy --emit native --mpy-cross-flags="-march=rv32imc"' test` on the QEMU/VIRT_RV32 board, and all tests pass.  The lack of extra jumps to the unwind label was checked with `mpy-cross` emitting code for the `debug` architecture.

This change is dependent on the current native emitter behaviour w.r.t. where to place the unwinding code relative to the unwinding  jump.  If the assumption of adjacency for the unwinding jump and the unwinding code is invalidated due to other changes to the emitter then this PR will make it generate invalid code.

Ideally this should be turned into a patch that would optimise all jumps regardless their location in the code as long as the label is right before or right after the code sequence in question.  However from what I can see it is way more complicated to achieve as the only place where this optimisation can be done is at the lower level of the emitter framework (arch-dependent) and only after the label offsets have been stabilised.  Omitting instructions at that point would move label addresses around and thus invalidate computed jumps.

### DIFF
--- a/py/emitnative.c
+++ b/py/emitnative.c
@@ -2082,6 +2082,18 @@ static void emit_native_jump_if_or_pop(emit_t *emit, bool cond, mp_uint_t label)
 }
 
 static void emit_native_unwind_jump(emit_t *emit, mp_uint_t label, mp_uint_t except_depth) {
+    // Skip the jump only if the exception stack and the emitter stack are
+    // empty.  An unwind request with an empty exception stack and an empty
+    // emitter stack only occurs at the end of a code block and it is the
+    // last reference to the exit label from an unwind block.  In that case
+    // the unwind block is right before the exit label anyway, so the jump
+    // can be omitted.
+    if (except_depth == 0 && label == emit->exit_label && emit->stack_info->vtype == VTYPE_PTR_NONE) {
+        // Simulate emit_native_jump's side effects.
+        mp_asm_base_suppress_code(&emit->as->base);
+        return;
+    }
+
     if (except_depth > 0) {
         exc_stack_entry_t *first_finally = NULL;
         exc_stack_entry_t *prev_finally = NULL;


agatti:
Hrm, the async tests pass for the Unix on my Arch Linux machine do pass when running `cd ../../tests && MICROPY_MICROPYTHON=../ports/unix/build-standard/micropython ./run-tests.py --emit native` - not sure what's going on in the CI environment for that.

agatti:
So `modasyncio` fails tests due to DEBUG being set and some assertions triggering in the unix/coverage port, along with different behaviour in a couple of tests w.r.t. task execution order.  Weird that unix/standard without assertions in release mode passes the very same tests though...

I've marked the PR as draft, as I need to investigate further on what's going on here.

agatti:
Whilst the changes proposed here seem to work with production code, I couldn't find a way out to avoid the test failures when compiled with `DEBUG=1`.

Not sure if it's the tests that expect a specific execution order of the asyncio functions that, if perturbed, makes them fail.  In any case, this can be sidestepped as the recompiler already takes care of this.

@dpgeorge Is this still worth looking at or can I just close it instead?

dpgeorge:
I had a look at this and saw that `extmod/ssl_poll.py` was failing.  Then I reduced that failure down to the following:
```py
def f(x):
    print(1)
    if x:
        print(2)
        return None
        print(3)
    print(4)    
    
f(1)
```
That should print 1 then 2.  But if you run it with this PR using the unix coverage build with the native emitter it prints 1 then 2 then 4!

So there's something wrong with the logic in the commit here.

agatti:
Thanks for looking into this!

With your test case it's indeed not working as expected.  I found a bit surprising that it passed `standard` tests but not `coverage`, especially since CI runs the tests also through `mpy-cross` with and without emitting native code.  Hence why this PR is marked as draft rather than something ready to be reviewed/merged.

Anyway, I'll look into this a bit more using your test case as a reference as it's much easier to reason with than asyncio stuff.

agatti:
I'm starting to suspect this is not really worth pursuing, since the code size increase may outweigh saving an opcode in some cases.  Unless there can be "tier 1" and "tier 2" optimisations, as in ones that are worth having in the interpreter, and ones that are better suited to have in `mpy-cross` instead?

Anyway, if my understanding is correct, what it is needed here is to extend the added check in `emit_native_unwind_jump` to also:

* wait until the emitter backend state is MP_ASM_PASS_COMPUTE, that the target label has an assigned offset, and that the emitter is not suppressing opcodes
* find the last label whose offset is assigned and that said offset is lesser or equal to the backend's current code offset (ie. finding the "parent" label of the current emitted instruction pointer)
* (assuming defined labels are always in ascending offset order) check if the target label is the direct follower of the label obtained in the previous step, ignoring undefined labels when scanning entries' offsets

and only then omit the jump, emulating its eventual side effects.

I thought `*emit->label_slot` would hold the last defined label, so I could use that as a starting point to skip the first label table scan, but when this check needs to be made all labels are already laid out so I cannot use that value.

So, it all boils down to how much code increase on all platforms this can add.  The broken version added more or less ~30 bytes on most platforms except for RV32, this one will probably be around ~200 bytes on non-RV32, make it ~300 on RV32.

dpgeorge:
I think we should just close this.  There are lots of other things to work on :smile:

agatti:
I agree.  I'll leave the related issue up there, though, so when the backlog will shrink down to a handful of entries it may be something worth looking at.  Maybe :)

PR #16613, opened 2025-01-19
lwip: ipv6: Fix incorrect peer address.

greezybacon:
### Summary

For IPv6 UDP sockets, the peer address was previously defined as only the first four bytes of the IP address. For IPv6 addresses, this resulted in an incorrect address expressed as IPv4. For instance, receiving a packet via `::recvfrom` from `'fe80::87:e7ff:fe48:629a'` is returned as having a peer address of `'254.128.0.0'`

### Testing

I have tested this with a UDP-based IPv6 service using `::sendto` and `::recvfrom`, and the address are now returned correctly to Python.

The underlying sockets function works correctly if provided the correct IPv6 addresses, so this doesn't affect the underlying socket interface. Here is an example echo service which doesn't work with the current implementation:

```python
s = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
s.bind(('::', 4000))
while True:
    msg, addr = s.recvfrom(512)
    # NOTE: `addr` is incorrect
    s.sendto(msg, addr)
```

This uses slightly more memory for IPv4 UDP socket servers as memory is allocated based on the largest supported address family. However, it's only a few bytes per function call.

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -37,7 +37,6 @@
 
 #if MICROPY_PY_LWIP
 
-#include "shared/netutils/netutils.h"
 #include "modnetwork.h"
 
 #include "lwip/init.h"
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the fix, this is definitely needed.

I think the `#include "shared/netutils/netutils.h"` line can now be removed as well.

dpgeorge (approved):

dpgeorge:

PR #16616, opened 2025-01-21
mpremote: Introduce timeout_overall to prevent endless loops

hmaerki:
### Summary

This is a follow up of https://github.com/micropython/micropython/pull/16531.

### Testing

In this situation, mpremote blocks forever:

* Precondition: Connect a brand new ESP32-C3-DevKitC. The factory installed firmware spits out `Turning the LED ON!` every second.
* Stimuli: Call `mpremote u0 eval 42`
* Error: `mpremote` blocks forever.
* Why: The character timeout in `read_until()` is reset every second.
* Desired behaviour: `read_unit()` should return after 10s.

It might not be a problem when mpremote hangs forever on the command line. However it is a severe problem when `read_until()` is called in octoprobe and `read_until()` hangs forever. This PR fixes both situations.

### Compile/flash the firmware to reproduce above situation

Connect a esp32s3 and flash using below command:
```bash
docker run --rm -v /dev:/dev --privileged -t docker.io/espressif/idf:v5.2.2 bash -c "git clone --depth 1 https://github.com/espressif/esp-idf.git && cd esp-idf/examples/get-started/blink && idf.py set-target esp32s3 && idf.py flash -p /dev/ttyUSB0"
```
 
### Fix

* Introduce `timeout_overall` in `read_until()`.

This is a new optional parameter and defaults to NONE (infinite). It does not break existing code.

### DIFF
--- a/tools/mpremote/mpremote/transport_serial.py
+++ b/tools/mpremote/mpremote/transport_serial.py
@@ -98,11 +98,14 @@ class SerialTransport(Transport):
     def close(self):
         self.serial.close()
 
-    def read_until(self, min_num_bytes, ending, timeout=10, data_consumer=None):
+    def read_until(
+        self, min_num_bytes, ending, timeout=10, data_consumer=None, timeout_overall=None
+    ):
         """
         min_num_bytes: Obsolete.
         ending: Return if 'ending' matches.
         timeout [s]: Return if timeout between characters. None: Infinite timeout.
+        timeout_overall [s]: Return not later than timeout_overall. None: Infinite timeout.
         data_consumer: Use callback for incoming characters.
             If data_consumer is used then data is not accumulated and the ending must be 1 byte long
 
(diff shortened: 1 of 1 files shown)


tools/mpremote/mpremote/transport_serial.py
```diff
@@ -142,7 +151,7 @@ def enter_raw_repl(self, soft_reset=True):
         self.serial.write(b"\r\x01")  # ctrl-A: enter raw REPL
 
         if soft_reset:
-            data = self.read_until(1, b"raw REPL; CTRL-B to exit\r\n>")
+            data = self.read_until(1, b"raw REPL; CTRL-B to exit\r\n>", timeout_overall=10)
```
dpgeorge, line 154: I think `timeout_overall=10` should be passed to both of the other calls to `read_until` below in this `enter_raw_repl` function.  That will prevent waiting forever in the case (1) triggering the soft reset does not work correctly but does print out data continuously (rare but could happen); and (2) boot.py forever prints out data and never exits.

Also, I suggest making a global constant for this timeout, eg at the top of this file just after the imports:
```py
_ENTER_RAW_REPL_TIMEOUT = 10
```

hmaerki: I added a commit but did it slightly different: I added a default parameter 'timeout_overall' to 'enter_raw_repl()' - it is consistent with the rest of the code. But I am also open for your proposal:-)

dpgeorge: Looks good!

dpgeorge (approved):

PR #16620, opened 2025-01-21
ports/rp2: Workaround for multicore lockout.

### LINKED ISSUE #15192: RP2 Hanging on file write after soft reboot when a thread was started

RP2 Pico

### MicroPython version

MicroPython v1.23.0-5.g3613ad962 on 2024-05-31; Raspberry Pi Pico with RP2040

### Reproduction

- Copy this code to main.py
- Run the code
- Do soft reboot

If the thread is started before writing to the file, it works. If the thread has not been started, it hangs.

```
import _thread
import time
thread_running = True
def second_thread():
    while thread_running:
        print("Second thread running")
        time.sleep(1)
#thread = _thread.start_new_thread(second_thread,()) # Works if thread is started here

print("Writing file 1") # Hangs here after soft reboot
with open("test1.bin","w") as fh:
    fh.write(b'test')
    
print("Writing file 2") # Hangs here after soft reboot
with open("test.bin","ab") as fh:
    fh.write(b'test')
print("End writing file")

thread = _thread.start_new_thread(second_thread,())

try:
    while True:
        print("First thread running")
        time.sleep(1)
except KeyboardInterrupt:
    thread_running = False
    time.sleep(2)

```

### Expected behaviour

The file should be written without blocking after a soft reboot whether the thread is started before or after writing the file.

### Observed behaviour

After soft reboot, the micro hangs and it needs a hard reset.

### Additional Information

I think everything is here. I'm running the code on a custom hardware (compatible with pico), I'll try with a pico whenever I can access one. 
Please ask if you want me to try something or you need more info.
Also, I did not experience bugs on 1.21.0 (although I did not try this specific code yet). It seems it happens in 1.22/1.23.

### Code of Conduct

Yes, I agree

### LINKED ISSUE #16619: rp2: flash writes broken after starting core1 and performing a soft reset

rp2040, rp2350 (Pico, Pico 2, Pico W, Pico 2 W)

### MicroPython version

N/A, but tested on MicroPython v1.25.0-preview.217.gb4f53a0e51.dirty and various stable releases.

### Reproduction

If I run the following code (via Thonny) then it breaks my ability to save files (trying to write to flash will cause a lockup) until I hard reset:

```python
import _thread

def core1():
    pass

_thread.start_new_thread(core1, ())

while True:
    pass

```

If I *remove* `multicore_lockout_start_blocking();` and `multicore_lockout_end_blocking();` it will work fine. Since my core1 is already doing nothing (spinning on `WFI`) there's no need to make sure it's doing nothing.

### Expected behaviour

I expect it to soft reset and for flash IO to work as normal.

### Observed behaviour

A "Ctrl+C" will stop the main loop on core0 but leave core1 spinning, flash writes will work.

A soft reset (stop in Thonny) will stop/reset core1, and flash writes will break.

### Additional Information

The bug happens because a soft reset will *reset* core1, but *will not* reset the multicore lockout victim state. The check on `multicore_lockout_victim_is_initialized(1 - get_core_num())` will pass, but `multicore_lockout_start_blocking()` will block forever, since `multicore_lockout_victim_is_initialized` just returns an incorrect boolean value that isn't reset by `multicore_reset_core1()`

This is documented in Pico SDK here: https://github.com/raspberrypi/pico-sdk/blob/95ea6acad131124694cda1c162c52cd30e0aece0/src/rp2_common/pico_multicore/multicore.c#L26-L32

This is an upstream bug in Pico SDK, documented here https://github.com/raspberrypi/pico-sdk/issues/2201

There may be a downstream workaround in the interim. For example `MICROPY_BEGIN_ATOMIC_SECTION` checks for `core1_entry` 

### Code of Conduct

Yes, I agree

MichaelBell:
### Summary

This workaround fixes #16619 until https://github.com/raspberrypi/pico-sdk/issues/2201 is fixed.

### Testing

Tested on Pico W, with the repro scenario outlined in #16619.

The alternative would be to wait until the pico-sdk issue is fixed and a new release of pico-sdk can be taken.

While the workaround is not exactly pretty, it is very self-contained, so I think it is worth having until then.


dpgeorge (approved):

dpgeorge:
This temporary fix looks good to me, thanks.

I think we should add a test for it though.  We don't currently have a test for this because all the threading tests make sure the threads run to completion before finishing the test.  So we'd need to add a test that simply starts a thread and lets it run, eg following the original bug report for this issue: https://github.com/micropython/micropython/issues/16619#issue-2802612660

projectgus:
> So we'd need to add a test that simply starts a thread and lets it run, eg following the original bug report for this issue: [#16619 (comment)](https://github.com/micropython/micropython/issues/16619#issue-2802612660)

I've implemented a test as discussed and push it here. It's a little fiddly as after the runner soft-resets and forcibly stops the thread, the bug won't trigger until something tries to write to flash. So I've made the test as two parts. Without this fix, the `part2` test hangs indefinitely and the runner times out. With this workaround, it succeeds (on both RP2040 and RP2350).

Thanks for submitting this, @MichaelBell!

dpgeorge (approved):
Thanks @projectgus for adding the test.  I've tested it on Pico 2 before and after the fix here, and it fails before and passes after.

PR #16623, opened 2025-01-22
ports/*: increase LWIP timers for ports with mDNS support

tpwrules:
### Summary

Despite the code comments claiming one is sufficient, the mDNS application is capable of using up to twelve timers. Three per IP protocol are started at once in `mdns_start_multicast_timeouts_ipvX`, then another two per protocol can be started in `mdns_handle_question`. Further timers can be started for two additional callbacks.

Having certain timers, such as `MDNS_MULTICAST_TIMEOUT`, fail to start due to none being free will break mDNS forever as the app will never realize it's safe to transmit a packet.

### Testing

This fixes mDNS on W5500_EVB_PICO. Before, mDNS would work for a bit after connection until the host's cache expired a minute or two later. Then the board would never respond to further queries. With this patch, all works well.

This PR goes somewhat overkill and allocates the maximal amount of timers; it's uncertain if all can run simultaneously, or how many callback timers are needed.

Each timer struct is 16 bytes on standard 32 bit builds. Plus 8 bytes of allocater overhead, that's 288 more bytes of RAM used which shouldn't be too horrible. Users who don't need mDNS can manually disable it to recover the RAM if necessary.

### DIFF
--- a/extmod/lwip-include/lwipopts_common.h
+++ b/extmod/lwip-include/lwipopts_common.h
@@ -65,7 +65,9 @@
 
 #define LWIP_NUM_NETIF_CLIENT_DATA      LWIP_MDNS_RESPONDER
 #define MEMP_NUM_UDP_PCB                (4 + LWIP_MDNS_RESPONDER)
-#define MEMP_NUM_SYS_TIMEOUT            (LWIP_NUM_SYS_TIMEOUT_INTERNAL + LWIP_MDNS_RESPONDER)
+
+// The mDNS responder requires 5 timers per IP version plus 2 others. Not having enough silently breaks it.
+#define MEMP_NUM_SYS_TIMEOUT            (LWIP_NUM_SYS_TIMEOUT_INTERNAL + (LWIP_MDNS_RESPONDER * (2 + (5 * (LWIP_IPV4 + LWIP_IPV6)))))
 
 #define SO_REUSE                        1
 #define TCP_LISTEN_BACKLOG              1


dpgeorge:
Thanks for the contribution.

> Despite the code comments claiming one is sufficient

Can you point to where the comment says one is sufficient?  I could not find it.

> Three per IP protocol are started at once per IP protocol in `mdns_start_multicast_timeouts_ipvX`,

Yes, that looks correct, I can see 3 independent timers in those functions.


> then another two per protocol can be started in `mdns_set_timeout`.

I don't follow this bit, because `mdns_set_timeout` is a helper function.  I think you meant to say this is called from `mdns_handle_question()`?  Then the 5 timers per protocol are:
- `multicast_timeout`
- `multicast_probe_timeout`
- `multicast_timeout_25TTL`
- `unicast_msg_in_use`
- `multicast_msg_waiting`

And it definitely looks possible to have all 5 of those running at once.

But then there are also additional calls to `sys_timeout()` in `src/apps/mdns/mdns.c`:
```
2164:      sys_timeout(MDNS_RESPONSE_TC_DELAY_MS, mdns_handle_tc_question, pkt);
2319:        sys_timeout(MDNS_PROBE_MAX_CONFLICTS_TIMEOUT, mdns_probe_and_announce, netif);
2322:        sys_timeout(MDNS_PROBE_DELAY_MS, mdns_probe_and_announce, netif);
2348:        sys_timeout(announce_delay, mdns_probe_and_announce, netif);
2786:    sys_timeout(MDNS_PROBE_MAX_CONFLICTS_TIMEOUT, mdns_probe_and_announce, netif);
2790:    sys_timeout(delay, mdns_probe_and_announce, netif);
```
I'm not sure how to work out which of those can be running together... and in fact those calls could start an arbitrary number of timers if they are called over and over again.  But, since there are only two independent callbacks (namely `mdns_handle_tc_question` and `mdns_probe_and_announce`) it seems sensible to allow just two timers for these calls.  So maybe there needs to be `2 + 5 * (IP4+IP6)` timers in total?

(Also, please check the CI code formatting failure.)

tpwrules:
Thanks for checking my work and the quick response.

> Can you point to where the comment says one is sufficient? I could not find it.

Line 11 of `src/apps/mdns/mdns.c` says "You need to increase MEMP_NUM_SYS_TIMEOUT by one if you use MDNS!" I do not know why the number of timers has expanded so dramatically, it's not a super old comment.

> I don't follow this bit, because `mdns_set_timeout` is a helper function. I think you meant to say this is called from `mdns_handle_question()`? Then the 5 timers per protocol are:
> ...

Sorry, yes I mis-pasted the function name when writing the message. Those are the same 5 timers I identified.

> But then there are also additional calls to `sys_timeout()` in `src/apps/mdns/mdns.c`:

I missed these, thank you. It looks like these are used in rarer circumstances, but I agree with your assessment that these could take arbitrary numbers of timers. It is unfortunate that each callback could be queued several times and use up the timer slots for the other processes. I think upstream might need to put some attention here.

I could send an inquiry to their list but I don't think I have bandwidth to fix it up at this time.

> (Also, please check the CI code formatting failure.)

I will increase the number of timers and address the failure and test again.

tpwrules:
Okay, took a few tries but I think we are good now!

dpgeorge:
Thanks for updating.

In order to eliminate code duplication, I made PR #16647 which factors out the common lwIP configuration for these four ports.  I think we should merge that first, then this PR becomes simpler, it's just changing one location.

dpgeorge:
@tpwrules are you able to update this PR and just make a single change in the consolidated file at `extmod/lwip-include/lwipopts_common.h`?

tpwrules:
Not sure what to do about the CI failure?

dpgeorge:
> Not sure what to do about the CI failure?

I reran it, it now passes.

extmod/lwip-include/lwipopts_common.h
```diff
@@ -65,7 +65,8 @@
 
 #define LWIP_NUM_NETIF_CLIENT_DATA      LWIP_MDNS_RESPONDER
 #define MEMP_NUM_UDP_PCB                (4 + LWIP_MDNS_RESPONDER)
-#define MEMP_NUM_SYS_TIMEOUT            (LWIP_NUM_SYS_TIMEOUT_INTERNAL + LWIP_MDNS_RESPONDER)
+#define _MDNS_TIMERS                    (2 + (5 * (LWIP_IPV4 + LWIP_IPV6)))
```
dpgeorge, line 68: I know this is a bit picky, but I think this macro should either be named longer like `MEMP_NUM_SYS_TIMEOUT_MDNS`, or not defined at all and just write the expression inline in `MEMP_NUM_SYS_TIMEOUT`.

eightycc:
CircuitPython has had similar problems with static memory pool allocations in lwIP. Our solution was to configure lwIP to allocate all of its memory dynamically from our top-level tlsf heap. See https://github.com/adafruit/circuitpython/issues/9837.

extmod/lwip-include/lwipopts_common.h
tpwrules: What do you think about the leading underscore? I wanted to make it clear it was not an lwIP flag.

dpgeorge (approved):

dpgeorge:

PR #16629, opened 2025-01-23
samd/boards: Add definitions for Generic boards, Adafruit NeoKey Trinkey and Adafruit QT PY SAMD21.

robert-hh:
### Summary

Add definitions for Generic boards, Adafruit NeoKey Trinkey and Adafruit QT PY SAMD21.
The Generic boards/MCUs chosen are:

- SAMD21J18, which applies to SAM21E18 and SAMD21G18 as well
- SAMD51P19, which applies to SAMD51G19 and SAMD51J19 as well
- SAMD51P20. which applies to SAMD51J20 as well

The Generic board firmware uses the internal RC-oscillator for clock and the internal flash for the file system.

The firmware for the QT PY board includes a variant supporting the optional SPIFLASH. For that
purpose, the support for variants was added.

### Testing

All firmware images were built and loaded to the respective boards. REPL started and scripts could be executed. The respective Generic firmware was tested with the lower pin count MCU versions as well.


ports/samd/Makefile
```diff
@@ -22,6 +34,9 @@ MCU_SERIES_LOWER = $(shell echo $(MCU_SERIES) | tr '[:upper:]' '[:lower:]')
 
 include ../../py/mkenv.mk
 include $(BOARD_DIR)/mpconfigboard.mk
+ifneq ($(BOARD_VARIANT),)
+include $(BOARD_DIR)/mpconfigvariant_$(BOARD_VARIANT).mk
```
dpgeorge, line 38: I suggest following stm32 and supporting optional, default variant config files:
```
# Include board specific .mk file, and optional board variant .mk file.
include $(BOARD_DIR)/mpconfigboard.mk
ifeq ($(BOARD_VARIANT),)
-include $(BOARD_DIR)/mpconfigvariant.mk
else
include $(BOARD_DIR)/mpconfigvariant_$(BOARD_VARIANT).mk
endif
```

ports/samd/boards/ADAFRUIT_QTPY_SAMD21/board.json
```diff
@@ -0,0 +1,20 @@
...
+    ],
+    "docs": "",
+    "features": [
+        "USB-C"
+    ],
+    "images": [
+        "qt_py_samd21.jpg"
+    ],
+    "mcu": "samd21",
+    "product": "QT PY (SAMD21)",
```
dpgeorge, line 13: Looks like the name is "QT Py - SAMD21".

ports/samd/boards/ADAFRUIT_QTPY_SAMD21/mpconfigboard.h
```diff
@@ -0,0 +1,6 @@
+#define MICROPY_HW_BOARD_NAME "QT PY"
```
dpgeorge, line 1: "QT Py"

ports/samd/boards/GENERIC_SAMD21J18/board.json
```diff
@@ -0,0 +1,12 @@
+{
+    "deploy": [
+        "../deploy.md"
+    ],
+    "docs": "",
+    "features": [
+        "USB"
+    ],
+    "mcu": "samd21",
+    "product": "Generic SAM21G18",
```
dpgeorge, line 10: Missing "D" in "SAMD21G18".

ports/samd/boards/GENERIC_SAMD21J18/pins.csv
```diff
@@ -0,0 +1,10 @@
+# The lines contain pairs of Pin name and Pin number.
+# Pin names must be valid Python identifiers.
+# Pin numbers have the form Pxnn, with x being A, B, C or D.
+# Lines starting with # or empty lines are ignored.
+
```
dpgeorge, line 5: Is it worth exposing all standard GPIO, so they can be used if they are physically available?  Otherwise no pins can be used (or does samd always provide defaults?).

ports/samd/Makefile
robert-hh: Updated.

ports/samd/boards/GENERIC_SAMD51P19/board.json
```diff
@@ -0,0 +1,15 @@
+{
+    "deploy": [
+        "../deploy.md"
+    ],
+    "docs": "",
+    "features": [
+        "USB"
+    ],
+    "mcu": "samd51",
+    "product": "Generic SAM51P19",
```
dpgeorge, line 10: Missing "D".

ports/samd/boards/GENERIC_SAMD51P19/board.json
```diff
@@ -0,0 +1,15 @@
...
+    "variants": {
+        "WLAN": "WiFi support using a ESP32 module with NINA firmware"
```
dpgeorge, line 12: Looks like this board does not have a variant defined?

ports/samd/boards/GENERIC_SAMD51P20/board.json
```diff
@@ -0,0 +1,15 @@
+{
+    "deploy": [
+        "../deploy.md"
+    ],
+    "docs": "",
+    "features": [
+        "USB"
+    ],
+    "mcu": "samd51",
+    "product": "Generic SAM51P20",
```
dpgeorge, line 10: Missing "D".

ports/samd/boards/GENERIC_SAMD51P20/board.json
```diff
@@ -0,0 +1,15 @@
...
+    "variants": {
+        "WLAN": "WLAN with SSL support"
```
dpgeorge, line 12: Does this variant exist?

dpgeorge (commented):
Regarding the generic boards:
- I think they should be named `SAMD_GENERIC_<mcu>` so that they match esp8266/esp32 generic board names.
- Please add `"vendor": "Microchip"` to the board.json

dpgeorge:
> * SAMD21J18, which applies to SAM21E18 and SAMD21G18 as well

It would be good to mention somewhere that this generic firmware applies to all these MCU variants.  Maybe for now just add it as a comment to the top of the `mpconfigboard.mk` file.  Later on it can be put in `board.json` and rendered into the download webpage.

dpgeorge:
> It would be good to mention somewhere that this generic firmware applies to all these MCU variants. Maybe for now just add it as a comment to the top of the `mpconfigboard.mk` file. Later on it can be put in `board.json` and rendered into the download webpage.

Actually, you can put some text into `board.md` for the generic boards, something like "The following firmware should work on most boards with ...".  See `ports/esp32/boards/ESP32_GENERIC/board.md` for inspiration.

ports/samd/boards/GENERIC_SAMD51P20/board.json
robert-hh: Not yet, it's a different PR. And maybe never.

ports/samd/boards/GENERIC_SAMD21J18/pins.csv
robert-hh: All pins are available through Pin.cpu. So you can create a pin object e.g. as
Pin.cpu.PA01
Pin("PA01")
Pin(1)

ports/samd/boards/GENERIC_SAMD21J18/board.json
robert-hh: Done. And it should be "SAMD21J18"

ports/samd/boards/GENERIC_SAMD51P19/board.json
robert-hh: Done

ports/samd/boards/GENERIC_SAMD51P19/board.json
robert-hh: Removed

ports/samd/boards/GENERIC_SAMD51P20/board.json
robert-hh: Done

ports/samd/boards/ADAFRUIT_QTPY_SAMD21/board.json
robert-hh: Done

ports/samd/boards/ADAFRUIT_QTPY_SAMD21/mpconfigboard.h
robert-hh: Done

robert-hh:
I have changed the files following your comments and added board.md files to the generic board definitions. Sorry for all the typos, copied along the boards.

robert-hh:
Alternatively I could add variants for the other package sizes. Only that would increase the build time and storage space on the server. For generic boards, I sketched a little board with KiCad just to get a nice 3D picture of it for the web site to get a license free image.

generic_board

dpgeorge:
> Alternatively I could add variants for the other package sizes.

No.  The way you have it at the moment, with 3 generic boards, is good.

But, for the naming, how about:
- `SAMD_GENERIC_21X18`
- `SAMD_GENERIC_51X19`
- `SAMD_GENERIC_51X20`

?

There's no repeating "SAMD" in the name, and the "X" indicates it suits multiple boards of that form.

robert-hh:
The chip name is SAMD21X18. How about
SAMD21X18_GENERIC ....

dpgeorge:
I was trying to match the esp32 generic board names, like `ESP32_GENERIC_C6`.  And if we did have generic stm32 boards I would think they'd be called something like `STM32_GENERIC_F405`.  And rp2 might be `RP2_GENERIC_2040`.

How about `SAMD_GENERIC_D21X18` etc?

robert-hh:
OK. Repeating the D makes it better readable.

dpgeorge (approved):
Tested the `SAMD_GENERIC_D21X18` firmware on an Adafruit ItsyBitsy M0 Express.  It works.

docs/samd/pinout.rst
```diff
@@ -567,6 +567,66 @@ The default devices at the board are:
...
+
+For the definition of the table columns see the explanation at the table for
+Adafruit ItsyBitsy M0 Express :ref:`samd21_pinout_table`.
+
+The default devices at the board are:
+
+- UART 0 at pins PA07/PA06, labelled RX/TX
+- I2C 1 at pins PA16/PA17, labelled SDA/SCL
+- SPI 0 at pins PA09/PA10/PA11, labelled MISO, MOSI and SCK
+- DAC output on pin PA02, labelled A0
```
dpgeorge, line 607: This is not a blocker for this PR, but what exactly are "default devices"?  It seems that the samd port requires the user to always specify I2C pins (for example).

Maybe it would be good to follow rp2 and stm32 and allow default pins, so the user can then just do `machine.I2C(1)` and get the default pin settings.  Same with UART and SPI.

robert-hh: Default devices are the ones with according labels on the silkscreen of the board, as indicated by "labelled". Theses could be assigned in a mpconfigboard.h setting, but still the device number would differ between board. Or the numbers have to be re-mapped. That is done in the MIMXRT port, and I do not like it. The mpconfigboard.h files of the MIMXRT port are way to long and complicated. Out of that experience, may aim was to have the SAMD board config files, as small as possible.

dpgeorge: It's possible to use just `pins.csv` to configure default pins for peripherals.  Eg `ADAFRUIT_QTPY_SAMD21` has pin aliases `SCL` and `SDA` defined.  So in `machine_i2c.h` you could have something like:
```c
#if defined(pin_SCL) && defined(pin_SDA)
// board defines default pins for I2C0, so use them as default
#endif
```
Or if you need to distinguish between I2C0 and I2C1 as the default you could add to `pins.csv`:
```
-I2C1_SCL,PA16
-I2C1_SDA,PA17
```
Then use those symbols in `machine_i2c.c`.

The point is, it should be possible to (a) use just `pins.csv` and not touch `mpconfigboard.h`; (b) standardise a naming scheme for default pins and use it on all boards, to make things consistent.

Note that the pin names don't have to be exposed to the user (as above, just prepend with -), it's just an internal mechanism for specifying pin configuration.

mattytrentini:
An mpbuild user just [reported a bug](https://github.com/mattytrentini/mpbuild/issues/65) due to the GENERIC boards not having a URL in the `build.json` files.

I can easily patch mpbuild to work around it, but should we add a url for these boards? They're the only definitions that don't have one. MicroChip doesn't provide great URLs but even pointing the the [sam-d home page](https://www.microchip.com/en-us/products/microcontrollers-and-microprocessors/32-bit-mcus/sam-32-bit-mcus/sam-d) might be adequate.

My longer-term fix is to use json-schema to enforce fields in _every_ `build.json` files but the point still stands; should the url key be optional?

robert-hh:
The sam-d home page is a good choice.

robert-hh:
@dpgeorge I have a branch now that uses default pins for UART, I2C and SPI. I have to do some testing to check, that the assignments are right and that reasonable error messages appear when needed. Still, the right ID must be supplied with the constructor, and it differs among boards.

dpgeorge:
> I have a branch now that uses default pins for UART, I2C and SPI. I have to do some testing to check, that the assignments are right and that reasonable error messages appear when needed.

Sounds good!

> Still, the right ID must be supplied with the constructor, and it differs among boards.

That's OK, much easier to document a single ID per board, rather than ID and pins.  Also much easier for the user to just specify the ID.

For further improvements there, so you don't need to specify anything, see #16671.

PR #16630, opened 2025-01-23
esp32: ADC driver update to the new esp_adc.

purewack:
### Summary

This is a complementary PR to #16521, focusing only on ADC, updating it to fix an issue with the RISCV adc function causing a conflict of drivers mentioned in #16469

There are several errata notes about not being able to change the bit-width of the ADCs certain chips.
The only chip that can switch resolution to a lower one is the normal esp32.
esp32 C2 & S3 are stuck at 12 bits, while S2 is at 13 bits.
On the S2, you can change the resolution, but it has no effect on the resolution, rather, it prevents attenuation from working at all!
I have set the resolution to the be maximum possible for each SoC, with the esp32 being the only one not throwing errors when trying to set the bit-width to 9,10,11 or 12 bits using ADC.width(bits)

- https://github.com/espressif/esp-idf/issues/13128 
- https://github.com/espressif/esp-idf/blob/release/v5.2/components/soc/esp32s3/include/soc/soc_caps.h
- https://docs.espressif.com/projects/esp-chip-errata/en/latest/esp32s2/03-errata-description/index.html
- https://docs.espressif.com/projects/esp-idf/en/v4.2/esp32/api-reference/peripherals/adc.html

### Testing

I have built and flashed the following variants: Generic, S2, S3, C3

The ADC behaves correctly as expected, with readings following the position of a 10K potentiometer.
Attenuation works too where values above the maximum range get clipped and readings stay at the maximum attenuated level:
-  ~1v with 0db attenuation 
- ~3.1v using 11db attenuation


ports/esp32/boards/sdkconfig.base
```diff
@@ -115,7 +115,7 @@ CONFIG_ADC_CAL_LUT_ENABLE=y
 CONFIG_UART_ISR_IN_IRAM=y
 
 # IDF 5 deprecated
-CONFIG_ADC_SUPPRESS_DEPRECATE_WARN=y
+CONFIG_ADC_SUPPRESS_DEPRECATE_WARN=n
```
dpgeorge, line 118: I think you can just remove this option, it will take the default, which is "n".

ports/esp32/machine_adc.c
```diff
@@ -225,20 +207,45 @@ static mp_int_t mp_machine_adc_read(machine_adc_obj_t *self) {
...
     return u16;
 }
 
 static mp_int_t mp_machine_adc_read_uv(machine_adc_obj_t *self) {
-    adc_atten_t atten = madc_atten_get(self);
-    return madcblock_read_uv_helper(self->block, self->channel_id, atten);
+    return madcblock_read_uv_helper(self->block, self->channel_id, mp_machine_adc_atten_get_helper(self));
 }
 
+
```
dpgeorge, line 219: Please try to consistently have just one blank line between functions.

ports/esp32/machine_adc.c
```diff
@@ -225,20 +207,45 @@ static mp_int_t mp_machine_adc_read(machine_adc_obj_t *self) {
...
+mp_int_t mp_machine_adc_atten_get_helper(const machine_adc_obj_t *self) {
+    uint8_t value = madc_obj_atten[self - &madc_obj[0]];
+    return value == 0 ? ADC_ATTEN_MAX : value - 1;
+}
+void mp_machine_adc_atten_set_helper(const machine_adc_obj_t *self, mp_int_t atten) {
```
dpgeorge, line 224: Add a line between functions.

ports/esp32/machine_adc.c
```diff
@@ -225,20 +207,45 @@ static mp_int_t mp_machine_adc_read(machine_adc_obj_t *self) {
...
-    madc_atten_helper(self, atten);
+    mp_machine_adc_atten_set_helper(self, atten);
+    apply_self_adc_channel_atten(self, mp_machine_adc_atten_get_helper(self));
+}
+
+
+mp_int_t mp_machine_adc_width_get_helper(const machine_adc_obj_t *self) {
+    return self->block->bitwidth;
+}
+void mp_machine_adc_block_width_set_helper(machine_adc_block_obj_t *self, mp_int_t width) {
```
dpgeorge, line 240: As above, one blank line between all functions.

ports/esp32/machine_adc_block.c
```diff
@@ -29,25 +29,21 @@
 
 #include "py/mphal.h"
 #include "adc.h"
-#include "driver/adc.h"
+#include "esp_adc/adc_oneshot.h"
 
 machine_adc_block_obj_t madcblock_obj[] = {
-    {{&machine_adc_block_type}, ADC_UNIT_1, SOC_ADC_RTC_MAX_BITWIDTH, -1, {0}},
+    {{&machine_adc_block_type}, ADC_UNIT_1, NULL},
```
dpgeorge, line 35: I think it's worth initialising the `bitwidth` field here, eg `.bitwidth = ...`.

docs/esp32/quickref.rst
```diff
@@ -442,14 +442,27 @@ Legacy methods:
 
     Equivalent to ``ADC.block().init(bits=bits)``.
 
+.. method:: ADC.deinit()
+
+    Provided to deinit the adc driver.
```
projectgus, line 447: This documentation for `ADC.deinit()` has split the documentation for `ADC.width()`, suggest moving it down around line 466

projectgus (commented):
I have one small note but this looks really nice, thanks for submitting it @purewack!

projectgus (approved):

ports/esp32/machine_adc.c
```diff
@@ -182,18 +155,31 @@ void madc_init_helper(const machine_adc_obj_t *self, size_t n_pos_args, const mp
...
+
 }
 
 static void mp_machine_adc_init_helper(machine_adc_obj_t *self, size_t n_pos_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
     madc_init_helper(self, n_pos_args, pos_args, kw_args);
 }
 
+static void mp_machine_adc_deinit(machine_adc_obj_t *self) {
+    if (self->block->handle) {
+        check_esp_err(adc_oneshot_del_unit(self->block->handle));
```
dpgeorge, line 179: If I do `adc.deinit()` twice in a row then the second one fails with an exception.  Ideally it would not, because usually it's expected that `deinit()` (and others like `close()`) don't raise if called multiple times.

Maybe set `self->block->handle = NULL` here after deiniting?

purewack: Sorry, my mistake. Just added this and tested. Now no error, and adc.read() when deinit'ed returns 0 read values.

purewack: Are you using some more verbose build settings? I have tested this on the ESP32, S2, S3, C3 using this script:
```py
import machine
a = machine.ADC(0)

print(a.read(),a.read_uv())

a.deinit()
print(a.read(),a.read_uv())

# deinit again on purpose 
a.deinit()
print(a.read(),a.read_uv())

a.init()
print(a.read(),a.read_uv())
```

I do not get the ESP error, just the base reading for read_uv etc. I used the fix you suggested unless I misunderstood :/  
This is the output as per the test script: 
```
2559 2169000
0 128000
0 128000
2549 2145000
```

dpgeorge: I'm using IDF 5.4.1.

Checking out this PR, rebasing on latest master (rebase is clean), compiling ESP32_GENERIC and running your above test I get:
```
MicroPython v1.26.0-preview.129.g8e8f77718 on 2025-05-20; Generic ESP32 module with ESP32
>>> import platform
>>> platform.platform()
'MicroPython-1.26.0-preview-xtensa-IDFv5.4.1-with-newlib4.3.0'
>>> 
paste mode; Ctrl-C to cancel, Ctrl-D to finish
=== import machine
=== a = machine.ADC(0)
=== 
=== print(a.read(),a.read_uv())
=== 
=== a.deinit()
=== print(a.read(),a.read_uv())
=== 
=== # deinit again on purpose 
=== a.deinit()
=== print(a.read(),a.read_uv())
=== 
=== a.init()
=== print(a.read(),a.read_uv())
13 138000
E (17428) adc_oneshot: adc_oneshot_read(184): invalid argument: null pointer
E (17428) adc_oneshot: adc_oneshot_read(184): invalid argument: null pointer
0 128000
E (17438) adc_oneshot: adc_oneshot_read(184): invalid argument: null pointer
E (17438) adc_oneshot: adc_oneshot_read(184): invalid argument: null pointer
0 128000
176 265000
>>> 
```

dpgeorge: Instead of printing a message it should raise an exception, eg:
```c
if (!self->block->handle) {
    mp_raise_OSError(MP_EPERM);
}
```

ports/esp32/machine_adc.c
```diff
@@ -218,27 +213,63 @@ static mp_obj_t mp_machine_adc_block(machine_adc_obj_t *self) {
 }
 
 static mp_int_t mp_machine_adc_read(machine_adc_obj_t *self) {
+    mp_machine_adc_is_init_guard(self);
```
dpgeorge, line 216: Instead of putting these calls in 5 separate locations here, it would be simpler to just do it in two locations in:
- `apply_self_adc_channel_atten()`
- `madcblock_read_helper()`

That reduces code size, and keeps the code a little simpler.

ports/esp32/machine_adc.c
```diff
@@ -63,6 +64,10 @@
     { MP_ROM_QSTR(MP_QSTR_ATTN_2_5DB), MP_ROM_INT(ADC_ATTEN_DB_2_5) }, \
     { MP_ROM_QSTR(MP_QSTR_ATTN_6DB), MP_ROM_INT(ADC_ATTEN_DB_6) }, \
     { MP_ROM_QSTR(MP_QSTR_ATTN_11DB), MP_ROM_INT(ADC_ATTEN_DB_11) }, \
+    { MP_ROM_QSTR(MP_QSTR_ATTN_MIN), MP_ROM_INT(ADC_ATTEN_MIN) }, \
+    { MP_ROM_QSTR(MP_QSTR_ATTN_MAX), MP_ROM_INT(ADC_ATTEN_MAX) }, \
+    { MP_ROM_QSTR(MP_QSTR_WIDTH_MIN), MP_ROM_INT(ADC_WIDTH_MIN) }, \
+    { MP_ROM_QSTR(MP_QSTR_WIDTH_MAX), MP_ROM_INT(ADC_WIDTH_MAX) }, \
```
dpgeorge, line 70: I don't think these min/max constants should be added.  The existing constants should be sufficient, and are explicit in the db/bit-width that they configure.

dpgeorge (approved):
This looks good now, a nice improvement, thank you!

Tested with ESP32_GENERIC firmware.  Deinit'ing and then doing a read now raises EPERM.

PR #16631, opened 2025-01-24
py/emitnative: Code generation improvements for Xtensa and RV32.

agatti:
### Summary

This PR contains ~two~ three improvements to the emitter for native and viper code.

In the first one, address generation for Xtensa and Xtensawin (eg. calculating a byte offset in an array of words/halfwords) was done by calculating the final byte offset by hand without even using shifts, which would mean three opcodes for halfwords (add + add + load/store) and five opcodes for words (add + add + add + add + load/store).  Said platform offers the `ADDX2` and `ADDX4` opcodes that perform an addition between two registers but with one of them being pre-shifted by either 2 or 4 bits, making them ideal to replace the addition sequences in the examples described earlier.

The second one exploits the fact that RV32 has a register (X0/ZERO) that is hardwired to ignore writes and to always return zero when read.  For native functions that require an exception handler, the code sequence emitted when setting up the exception stack would zero a temporary register and then use that to clear some memory locations needed to be zeroed.  On RV32 the temporary register clearing is skipped and memory is written with the contents of X0/ZERO, obtaining the same result without an extra XOR operation.

Finally, the third one lets Xtensa and Xtensawin Viper load and stores that go through an immediate offset to emit a single opcode if the offset is small enough to be embedded in the opcode itself.  The regular code sequence is emitted otherwise.

### Testing

`micropython/viper*` and `micropython/native*` tests pass on an ESP8266, on a plain ESP32 with ESP-IDF 5.2.0, on QEMU/VIRT_RV32, and on QEMU/MPS2_AN385.

There are more places where a temporary register is zeroed and then used to clear memory, but rather than using an explicit 0 value they use `MP_OBJ_NULL` which is defined to 0 as well.  To keep things safe I left those bits alone, but if needed I can expand that by perform an `#if` check on the value of `MP_OBJ_NULL` and use X0/ZERO on RV32 in those cases.

### DIFF
--- a/py/asmrv32.h
+++ b/py/asmrv32.h
@@ -694,6 +694,7 @@ static inline void asm_rv32_opcode_xori(asm_rv32_t *state, mp_uint_t rd, mp_uint
 #define REG_LOCAL_1 ASM_RV32_REG_S3
 #define REG_LOCAL_2 ASM_RV32_REG_S4
 #define REG_LOCAL_3 ASM_RV32_REG_S5
+#define REG_ZERO ASM_RV32_REG_ZERO
 
 void asm_rv32_meta_comparison_eq(asm_rv32_t *state, mp_uint_t rs1, mp_uint_t rs2, mp_uint_t rd);
 void asm_rv32_meta_comparison_ne(asm_rv32_t *state, mp_uint_t rs1, mp_uint_t rs2, mp_uint_t rd);
@@ -756,6 +757,7 @@ void asm_rv32_emit_store_reg_reg_offset(asm_rv32_t *state, mp_uint_t source, mp_
 #define ASM_STORE_REG_REG(state, rs1, rs2) ASM_STORE32_REG_REG(state, rs1, rs2)
 #define ASM_SUB_REG_REG(state, rd, rs) asm_rv32_opcode_sub(state, rd, rd, rs)
 #define ASM_XOR_REG_REG(state, rd, rs) asm_rv32_emit_optimised_xor(state, rd, rs)
+#define ASM_CLR_REG(state, rd)
 
 #endif
 
--- a/py/emitnative.c
+++ b/py/emitnative.c
@@ -1550,6 +1550,11 @@ static void emit_native_load_subscr(emit_t *emit) {
                             asm_rv32_opcode_lbu(emit->as, REG_RET, reg_base, index_value);
                             break;
                         }
+                        #elif N_XTENSA || N_XTENSAWIN
+                        if (index_value > 0 && index_value < 256) {
+                            asm_xtensa_op_l8ui(emit->as, REG_RET, reg_base, index_value);
+                            break;
+                        }
                         #endif
                         need_reg_single(emit, reg_index, 0);
                         ASM_MOV_REG_IMM(emit->as, reg_index, index_value);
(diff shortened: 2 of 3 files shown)


py/emitnative.c
```diff
@@ -288,6 +288,14 @@ struct _emit_t {
     ASM_T *as;
 };
 
+#if N_RV32
+#define REG_ZERO ASM_RV32_REG_ZERO
+#define CLEAR_REGISTER(as, reg)
+#else
+#define REG_ZERO REG_TEMP0
+#define CLEAR_REGISTER(as, reg) ASM_XOR_REG_REG(as, reg, reg)
+#endif
```
dpgeorge, line 297: I feel like this should go in `py/asm*.h`, alongside the other `REG_xxx` definitions.  And only rv32 needs to define it, then here it could be:
```c
#ifndef REG_ZERO
#define REG_ZERO REG_TEMP0
...
#endif
```
Similar to how `REG_PARENT_RET` is defined in this file.

agatti: Fair point.  This should be addressed in the current revision of the PR.

agatti:
I've found another couple of places to speed things up on Xtensa - I'll just add the new commit to this PR rather than opening another one.

agatti:
This should be it, I guess :)

py/asmrv32.h
```diff
@@ -756,6 +757,7 @@ void asm_rv32_emit_store_reg_reg_offset(asm_rv32_t *state, mp_uint_t source, mp_
 #define ASM_STORE_REG_REG(state, rs1, rs2) ASM_STORE32_REG_REG(state, rs1, rs2)
 #define ASM_SUB_REG_REG(state, rd, rs) asm_rv32_opcode_sub(state, rd, rd, rs)
 #define ASM_XOR_REG_REG(state, rd, rs) asm_rv32_emit_optimised_xor(state, rd, rs)
+#define ASM_ZERO_REGISTER(state, reg)
```
dpgeorge, line 760: To match the naming of the other macros here, I suggest `ASM_CLR_REG(state, rd)`; "clear" is more of an action than "zero" which really means "zero out".

agatti: No worries, I've updated the commit in question to use your suggestion :)

py/emitnative.c
```diff
@@ -1591,6 +1606,17 @@ static void emit_native_load_subscr(emit_t *emit) {
...
+                        #elif N_XTENSA || N_XTENSAWIN
+                        if (index_value > 0) {
+                            if (index_value < 16) {
+                                asm_xtensa_op_l32i_n(emit->as, REG_RET, reg_base, index_value);
+                                break;
+                            }
+                            if (index_value < 256) {
+                                asm_xtensa_op_l32i(emit->as, REG_RET, reg_base, index_value);
+                                break;
+                            }
```
dpgeorge, line 1618: I suggest:
```c
if (index_value > 0 && index_value < 256) {
    asm_xtensa_l32i_optimised(emit->as, REG_RET, reg_base, index_value);
}
```
just to make it a little simpler.

Also, similar comment to the s32i below.

agatti: Missed that, thanks!  That should be sorted out now.

dpgeorge (approved):

PR #16633, opened 2025-01-24
ports/stm32: Make ETH DMA buffer attributes configurable.

kwagyeman:
This commit allows the attributes for the eth_dma buffer to be overridden by a port.

OpenMV's firmware for the Arduino Portenta stores .data in the DTCM which is inaccessible by the ethernet controller. By overriding the buffer's attributes with alignment and section specifications, we can place the ethernet buffer in the right SRAM location.

### DIFF
--- a/ports/stm32/eth.c
+++ b/ports/stm32/eth.c
@@ -121,7 +121,7 @@ typedef struct _eth_t {
     int16_t (*phy_get_link_status)(uint32_t phy_addr);
 } eth_t;
 
-static eth_dma_t eth_dma __attribute__((aligned(16384)));
+static eth_dma_t eth_dma MICROPY_HW_ETH_DMA_ATTRIBUTE;
 
 eth_t eth_instance;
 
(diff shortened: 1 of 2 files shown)


kwagyeman:
@iabdalkader

ports/stm32/mpconfigport.h
```diff
@@ -193,6 +193,10 @@ extern const struct _mp_obj_type_t network_lan_type;
 #define MICROPY_HW_NIC_ETH
 #endif
 
+#ifndef MICROPY_HW_ETH_DMA_ATTRIBUTE
+#define MICROPY_HW_ETH_DMA_ATTRIBUTE __attribute__((aligned(16384)));
+#endif
```
dpgeorge, line 198: I think this would be better in `mpconfigboard_common.h`, near the bottom of that file.

kwagyeman: Done

dpgeorge (approved):

dpgeorge:
Rebased and merged in 112f65776588ef0cc3307f8bef208a480037d7d0

Issue #15297, opened 2024-06-17
rp2 - mDNS not working


massimosala:
### Port, board and/or hardware

rp2040

### MicroPython version

mp 1.22.2

### Reproduction

There isn't a code sample, the feature is built-in in the firmware.

### Expected behaviour

The micro working on WiFi (STA_IF or AP_IF) should respond to mDNS queries.

### Observed behaviour

Not working.

### Additional Information

I code in Python a mDNS and LLMNR responder.

It works like a charm on the esp8266: the micro, with its hostname xxxxxx.local is recognized by any browser (browsers use mDNS) and by ping (on microsoft pc, ping use also LLMNR).

Running my module on the rp2:

- LLMNR works
- MDNS doesn't work: when my code binds a UDP socket to the local port 5353, mp raises an exception
_already in use_

So the firmware has already taken the mDNS port ... is trying to do something... but mDNS isn't working.

### Code of Conduct

Yes, I agree

massimosala:
Tested also with MP 1.23.0 : not resolved

felixdoerre:
On rp2 MDNS is enabled to be provided by lwip: https://github.com/micropython/micropython/blob/master/ports/rp2/lwip_inc/lwipopts.h#L35. Maybe you want to compile your own firmware with that flag disabled?

mseminatore:
I am looking for a fix for this as well. I would prefer not to build my own custom firmware and use slimdns.

mseminatore:
> On rp2 MDNS is enabled to be provided by lwip: https://github.com/micropython/micropython/blob/master/ports/rp2/lwip_inc/lwipopts.h#L35. Maybe you want to compile your own firmware with that flag disabled?

But I think the point being made is that LWIP is **not** providing a working mDNS implementation on rp2 and yet is holding onto the port. Could that perhaps be a result of a build order of operations with the LWIP library and the rps lwipopts.h? Or some missing glue code?

It is fair if mDNS support is not yet implemented for the rp2 port. But it seems very much like a bug for rp2 lwipopts.h to define the following if mDNS support is not properly included/implemented in the fw:

```
#define LWIP_DNS_SUPPORT_MDNS_QUERIES   1
#define LWIP_MDNS_RESPONDER             1
```

I also notice that the following lines are not included in the rp2 mpconfigport.h file:

```
#ifndef MICROPY_HW_ENABLE_MDNS_QUERIES
[...]

Issue #10957, opened 2023-03-07
RP2040: mDNS support in AP mode


ThinkTransit:
Implement mdns in AP mode on RP2 port to allow users to access micropython device by hostname.

Issue #8946, opened 2022-07-22
LWIP mDNS interface on pico port


omogenot:
Hi,

First of all, congrats for the tremendous work you guys have achieved !

I'm quite new to Micropython stuff, but succeeded however to compile it for a W5500_EVB_PICO board, and it's working like a charm. I even understood how to freeze some of my modules in the firmware to save RAM space.

As far as I can see in the rp2/lwip_inc/lwipopts.h file, there are two constants:
```
#define LWIP_DNS_SUPPORT_MDNS_QUERIES   1
#define LWIP_MDNS_RESPONDER             1
```
that supposedly implies some mDNS support to the lwip driver. This is confirmed by the fact that the port 5353 is busy as you cannot bind a socket to it. So far, so good. However, I do not see any API/Object/Function when doing an "import lwip" or any other module to take advantage of this mDNS support and publish my pico IP and services.

How can I do? Or shall these constants be put to 0 by default and I will have to write my own mDNS functions?

I thank you in advance for your advices.

Olivier.

jimmo:
> How can I do? Or shall these constants be put to 0 by default and I will have to write my own mDNS functions?

@omogenot the simple answer for right now is that yes you should disable these flags and then implement mDNS in Python. https://github.com/cbrand/micropython-mdns might be helpful.

In more detail, on rp2 the lwip mdns features are enabled by default because on the Pico W board the wlan driver supports setting a hostname, which it also advertises via LWIP's mDNS. (Ok it doesn't quite support configuring it, but it does set a default hostname, see https://github.com/micropython/micropython/issues/8906).
The wiznet LAN driver doesn't support this at all yet, so on rp2 boards that don't use the cyw43 wlan driver we should probably disable those lwip mdns flags by default until it does so as not to block the port.

But yes, longer term this should be more configurable from Python (e.g. modlwip or something...not sure).

omogenot:
Hi,

Thanks for the update. I therefore modified the file extmod/network_wiznet5k.c, stealing code from cyw43 driver to add at line 520:
```
    #if LWIP_MDNS_RESPONDER
    // TODO better to call after IP address is set
    char mdns_hostname[9];
    memcpy(&mdns_hostname[0], "PICO_EVB", 8);
    mdns_hostname[8] = '\0';
    mdns_resp_add_netif(&self->netif, mdns_hostname, 60);
    #endif
```
Et voilà ! 
[...]

PR #16641, opened 2025-01-25
ports/rp2: Fix rp2 mdns responder issue.

mseminatore:
### Summary
As a number of users have reported via issues and discussions, the RP2 port of MicroPython has an incomplete mDNS implementation. The code in main.c calls `mdns_resp_init()` which opens the UDP socket for mDNS. However, no code in the cyw43 driver of the RP2 port makes the proper calls to `mdns_resp_add_netif()` and `mdns_resp_remove_netif()` to send the announce packets. The wiznet5k driver does make these calls and was used as a model for these changes.

This PR attempts to address this by very small changes to the `extmod/network_cyw43.c` driver file. As far as I can tell this driver is currently used only by the RP2 port.

### Testing
This was tested by building the firmware and deploying to a PicoW, starting a WLAN connection and pinging PicoW.local from several devices on the local network. I also tested changing the hostname from the default.

Alternatively, if this PR is not accepted we should disable the partial implementation by setting `LWIP_MDNS_RESPONDER` to 0 in `lwipopts.h` so that users can implement their own mDNS solutions without E_ADDRINUSE failures. A user should not have to build their own firmware to disable a partial/broken mDNS solution.


mseminatore:
@dpgeorge @omogenot @felixdoerre My apologies for the formatting issues. I am a first time contributor and am not certain what I did incorrectly.

extmod/network_cyw43.c
```diff
@@ -96,8 +97,20 @@ static void network_cyw43_print(const mp_print_t *print, mp_obj_t self_in, mp_pr
 static mp_obj_t network_cyw43_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
     mp_arg_check_num(n_args, n_kw, 0, 1, false);
     if (n_args == 0 || mp_obj_get_int(args[0]) == MOD_NETWORK_STA_IF) {
+        #if LWIP_MDNS_RESPONDER
+            // NOTE: interface is removed in network_cyw43_deinit
```
dpgeorge, line 101: You need to dedent these lines to line up with the `#` (or maybe you used tabs?).

dpgeorge:
Thanks for the contribution.  Did you also see #16623?  Maybe that actually fixes the same issue this PR is trying to fix.

extmod/network_cyw43.c
mseminatore: Thank you, I've updated the PR to fix the dedent formatting issue.

mseminatore:
> Thanks for the contribution. Did you also see #16623? Maybe that actually fixes the same issue this PR is trying to fix.

This PR is trying to address some of the issues mentioned in #15297, #10957 and #8946 

I just looked at that PR. It appears that it only changes .h configuration for timers but does not add any missing mDNS functionality. The lwIP [documentation](https://www.nongnu.org/lwip/2_0_x/group__mdns.html#gaa619ac8f46a4b4021195720f0355cbeb) seems clear that a call to `mdns_resp_add_netif()` is required for mDNS announce packets to be sent.

Looking over the code in `extmod/network_wiznet5k.c` it demonstrates the same calling pattern.

tpwrules:
A PR recently went in (#16011) to correct those calls for the wiznet5k driver. With that, plus the timer PR, mDNS seems to work well on RP2 using wiznet5k Ethernet.

I don't have wireless devices to test but this seems like a similar fix for a similar issue to the wiznet5k PR. There is some discussion in that PR about the calling pattern which might be a good reference; I have not cross-checked the API usage.

mseminatore:
> A PR recently went in (#16011) to correct those calls for the wiznet5k driver. With that, plus the timer PR, mDNS seems to work well on RP2 using wiznet5k Ethernet.
> 
> I don't have wireless devices to test but this seems like a similar fix for a similar issue to the wiznet5k PR. There is some discussion in that PR about the calling pattern which might be a good reference; I have not cross-checked the API usage.

Yes, it makes sense that the wiznet5K and cyw32 drivers for the RP2 should be similar. And that's why I tried to follow the pattern. I am new to the MicroPython codebase and it looks some time to peruse the code and get a feel for the patterns. Not certain my solution is 100% correct but it does function.

Regarding the timers, I do find that the RP2 port, at least for PicoW, sometimes gets into a bad state where the WiFi is not responding at startup. Maybe the timers resolve that?

mseminatore:
@dpgeorge @omogenot @felixdoerre My apologies for the incorrectly formatted commit messages. Would you prefer that I yank this PR and resubmit? Doing so will lose the comments on this PR.

robert-hh:
You do not have to close the PR. Just fix the commit message locally with `git commit --amend` in combination with `git rebase -i commit-id` and do a force push to your branch. Then, the PR will be updated.

mseminatore:
> You do not have to close the PR. Just fix the commit message locally with `git commit --amend` in combination with `git rebase -i commit-id` and do a force push to your branch. Then, the PR will be updated.

Thanks, I did that and it appears to have worked.

projectgus:
I'm a little surprised this works, as the cyw43-driver only initialises the netif and adds it to LWIP in `cyw43_cb_tcpip_init()` (implemented in `lib/cyw43-driver/src/cyw43_lwip.c`), which is a callback when the low-level network interface goes "up".

At the time the object is constructed in Python, I don't think this has happened yet - so although the argument to `mdns_resp_add_netif()` points to the correct structure, its contents will be all zeroes (I think). In particular, the calls to add the interface to a multicast group for mDNS probably fail (unless I'm misunderstanding the context here). But maybe that doesn't really matter, if basic mDNS is working...

(If this is a problem, I think the only way around it with the current driver design would be to add/remove the interface to/from mdns inside the cyw43-driver callbacks instead of from micropython code. But maybe it's not actually a problem.)

mseminatore:
> I'm a little surprised this works, as the cyw43-driver only initialises the netif and adds it to LWIP in `cyw43_cb_tcpip_init()` (implemented in `lib/cyw43-driver/src/cyw43_lwip.c`), which is a callback when the low-level network interface goes "up".
> 
> At the time the object is constructed in Python, I don't think this has happened yet - so although the argument to `mdns_resp_add_netif()` points to the correct structure, its contents will be all zeroes (I think). In particular, the calls to add the interface to a multicast group for mDNS probably fail (unless I'm misunderstanding the context here). But maybe that doesn't really matter, if basic mDNS is working...
> 
> (If this is a problem, I think the only way around it with the current driver design would be to add/remove the interface to/from mdns inside the cyw43-driver callbacks instead of from micropython code. But maybe it's not actually a problem.)

Thanks. I will write some test code to see what the netif state is at the time of the `mdns_resp_add_netif()` call. If netif is uninitialized, then perhaps lwIP defaults to the default_netif? I will have a look at the lwIP code to see how it behaves and what validation it performs.

That said, in my testing the mDNS for the device was properly advertised to my local network as PicoW.local and it was responding properly to pings.

projectgus:
Sounds great @mseminatore, sorry I posted that concern without doing any follow-up myself.

The part that I suspect won't work is sending the announcement to add the interface to the IGMP group (for IPv4) or MLD (for IPv6). This probably won't matter when everything is on a single network with no routing, but it may matter if the mDNS multicast group is also routed between networks. That said, having mDNS work on a single local network is probably what most people are looking for.

(That's assuming nothing else gets corrupted by adding an incomplete netif to lwip mdns. I don't know about this part.)

mseminatore:
> Sounds great @mseminatore, sorry I posted that concern without doing any follow-up myself.
> 
> The part that I suspect won't work is sending the announcement to add the interface to the IGMP group (for IPv4) or MLD (for IPv6). This probably won't matter when everything is on a single network with no routing, but it may matter if the mDNS multicast group is also routed between networks. That said, having mDNS work on a single local network is probably what most people are looking for so it might be fine to simply ignore this.
> 
> (That's assuming nothing else gets corrupted by adding an incomplete netif to lwip mdns. I don't know about this part.)

@projectgus let me know if you'd prefer to see the approach you called out in #15297 instead.

dpgeorge:
I also have similar concerns as @projectgus that mDNS needs to be started *after* the netif is configured.

There are three ways we could do that:
1. Modify cyw43-driver to init/deinit the mDNS itself in the correct location, as @projectgus suggested.
2. Modify cyw43-driver to add callback hooks after a `netif_add` and before a `netif_remove` so that users of this library can do whatever they need at that point.  Then in this repo, implement those callback hooks to init/deinit mDNS.
3. Do the init here but put it in `network_cyw43_active()`: before the call to `cyw43_wifi_set_up() one would need to deinit mDNS, then after the call to `cyw43_wifi_set_up()` one would init mDNS (but only in the case that it was setting the interface up when it was previously down).

Option (3) only requires changes in this repo but it's a little tricky to get right.  I think option (2) is the right way forward.

mseminatore:
Thanks for the review @dpgeorge I agree that option 2 seems the most robust solution. Do you want me to draft a PR for option 2 or should someone else do that? Happy to pursue either path. 

If you'd like me to draft a cyw43 lib PR can you let me know whether you'd prefer to see support for a single callback, a list of callbacks, or chainable callbacks?

Thanks!

andrewleech:
On a related note, not meaning to make things more complicated but I think there should also be a call made to
```
    #if LWIP_MDNS_RESPONDER
    mdns_resp_netif_settings_changed(&netif);
    #endif
```
any time an ip address is changed, eg in `extmod/network_lwip.c` in ipconfig/ifconfig and any time the ip is set/changed by the dhcp client.

As I write this I realise it's really a new feature request which should probably be tracked in its own issue? I'll write one up if preferred.

dpgeorge:
@mseminatore I made a PR for cyw43-driver here: https://github.com/georgerobotics/cyw43-driver/pull/138

Hopefully that's enough to do what you need with mDNS.  If you could test that out with the required mDNS fixes, that would be great!

dpgeorge:
> On a related note, not meaning to make things more complicated but I think there should also be a call made to
> mdns_resp_netif_settings_changed

I don't think we need to do this because:
```c
/**
 * @ingroup mdns
 * Announce IP settings have changed on netif.
 * Call this in your callback registered by netif_set_status_callback().
 * No need to call this function when LWIP_NETIF_EXT_STATUS_CALLBACK==1,
 * this handled automatically for you.
 * @param netif The network interface where settings have changed.
 */
#define mdns_resp_netif_settings_changed(netif) mdns_resp_announce(netif)
```
and we have `LWIP_NETIF_EXT_STATUS_CALLBACK` enabled.

dpgeorge:
I tested this PR as-is, and it doesn't work for me.  The problem does look exactly like what @projectgus mentioned above, that the IGMP (multicast) mDNS group is not joined because the netif is not ready when mDNS is started on it.

@mseminatore will you get a chance to update this PR using the new hooks in cyw43-driver?

dpgeorge:
The original (first version) cyw43-driver did actually handle mDNS correctly, see 7b70ab7258986567dd970d5108cd45c706b36a96 .  It called `mdns_resp_add_netif` and `mdns_resp_remove_netif` in precisely the locations that https://github.com/georgerobotics/cyw43-driver/pull/138 added hooks for.

mseminatore:
@dpgeorge yes, I will take a look and see if it fixes this PR. Is the driver PR integrated or do I need to pull that directly?

dpgeorge:
> Is the driver PR integrated or do I need to pull that directly?

The cyw43-driver has just been updated, so you can use the macros now.

mseminatore:
@dpgeorge et al, I am sorry for the delay, I was on other projects and then ill. I've used the new hooks and it works nicely in my testing. I have decided to create a new clean PR since this one has gotten messy with several resets and rebases and I don't want to risk a bad PR/merge to main. I will link to this one to keep the history.

mseminatore:
@dpgeorge If OTOH you are OK with this PR as it currently stands that is fine as well. The other PR is the same code changes sans the path of messy checkins and improperly formatted log messages.

mseminatore:
See #17057 for new, cleaner version of this PR

PR #16647, opened 2025-01-25
extmod/lwip-include: Factor common lwIP config into lwipopts_common.h.

dpgeorge:
### Summary

This lwIP configuration file has options that are common to all ports, and the ports are updated to use this file.  This change is a no-op, the lwIP configuration remains the same for the four ports using this common file.

This reduces code duplication, keeps the ports in sync, and makes it easier to update the configuration for all ports at once.

### Testing

For each port built a board with lwIP enabled: `RPI_PICO_W`, `PYBD_SF6`, `TEENSY41`, `ARDUINO_PORTENTA_C33`.  Built both before and after the commit in this PR.  Then compared the binary firmware.  The only things that changed in the binary were (1) the git commit hash in the version string; (2) the ordering of some of the pins in the `machine.Pin.board` dict (for some reason the order of these are not deterministic).  So I could conclude that the lwIP config changes are a no-op.

Edit: also built an stm32 board with PPP enabled and it builds.

### DIFF
--- a/extmod/lwip-include/lwipopts_common.h
+++ b/extmod/lwip-include/lwipopts_common.h
@@ -1,0 +1,109 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 Damien P. George
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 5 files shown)


projectgus (approved):
Yay for standardisation! Change LGTM.

tpwrules:
Is `lwipopts.h` in the same directory of the new common file necessary anymore?

dpgeorge:
> Is `lwipopts.h` in the same directory of the new common file necessary anymore?

Probably not needed anymore.  That was there originally as a full lwIP configuration for ports if they wanted to use it, but none ever did, the ports all needed their own custom configuration (there are also `arch/` files that can similarly be removed).

PR #16655, opened 2025-01-27
parsenumbase: Favor clarity.

jepler:
### Summary

In #13576 I optimized some code in parsenumbase for code size (using `!(b & ~16)` instead of `(b == 0 || b == 16)` for example. However, based on some testing on godbolt, this seems unnecessary for code size, at least on gcc x86_64 and arm, and it's definitely not good for clarity.

### Testing

Ran the Unix tests.

If any ports see code size increased with this change, then it's probably undesirable.

### DIFF
--- a/py/parsenumbase.c
+++ b/py/parsenumbase.c
@@ -41,11 +41,11 @@ size_t mp_parse_num_base(const char *str, size_t len, int *base) {
     if (c == '0') {
         c = *(p++) | 32;
         int b = *base;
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thank you, this looks a lot clearer, and the code size did not change.

dpgeorge:
Rebased and merged in bfb1bee6feba9f06452a0e4b572ec4d15df325cf with an embellished commit message.

PR #16656, opened 2025-01-27
extmod/asyncio: Make SingletonGenerator create_task()-aware.

jjsuwa-sys3175:
### Summary

This commit allows the awaitable returned by `asyncio.sleep/sleep_ms()` to be passed directly to `asyncio.create_task()`.

Although the documentation says that `sleep/sleep_ms()` are coroutines and `create_task()` accepts coroutines, this is not the case; `sleep/sleep_ms()` actually return a `SingletonGenerator()`, which is missing a `send()` method.

```
=== import asyncio
=== asyncio.run(asyncio.sleep_ms(2000))
===
  File "<stdin>", line 3, in <module>
  File "asyncio/core.py", line 1, in run
  File "asyncio/core.py", line 1, in create_task
TypeError: coroutine expected
```

This is easily solved by implementing the missing `send()` method.

### Testing

The principle is as follows:

```
=== import asyncio
=== from time import ticks_ms as ticks, ticks_add
===
=== class SingletonGenerator2(asyncio.SingletonGenerator):
===   def send(self, x):
===     self.__next__()
===
=== def sleep_ms(t, sgen=SingletonGenerator2()):
===   sgen.state = ticks_add(ticks(), max(0, t))
===   return sgen
===
=== asyncio.run(sleep_ms(2000))
===
>>>
```

In exchange for adding just two lines of source code and one method definition, we no longer have to manually `async def` wrap awaitables that wait for a specified time; the benefits are obvious.

### DIFF
--- a/extmod/asyncio/core.py
+++ b/extmod/asyncio/core.py
@@ -49,6 +49,9 @@ class SingletonGenerator:
             self.exc.__traceback__ = None
             raise self.exc
 
+    def send(self, x):
+        self.__next__()
+
 
 # Pause task execution for the given time (integer in milliseconds, uPy extension)
 # Use a SingletonGenerator to do it without allocating on the heap


dpgeorge:
Thanks for the contribution.

Do you have any examples of code that would use this pattern, to create a task that simply waits for some time?

jjsuwa-sys3175:
@dpgeorge 
> Do you have any examples of code that would use this pattern, to create a task that simply waits for some time?

Before I get to the code examples, we should first clear up some of the issues that `asyncio` newbies (like me) might have.

As I mentioned before, [the documentation](https://docs.micropython.org/en/latest/library/asyncio.html) says:

- asyncio.sleep(t)
Sleep for t seconds (can be a float). **This is a coroutine.**

and:

- asyncio.run(coro)
Create a new task from **the given coroutine** and run it until it completes.

In the past, I tried to feed a coroutine that sleeps for a few seconds directly to `asyncio.run()` as a quick exercise - hoping to get a bit of a wait for the REPL prompt... No need to mention the result :(

```
async def aio_sleep(t):
  await asyncio.sleep(t)
```
For now, it would be fine to wrap it like this, but rather than burdening users with the trouble, it would be better to make `SingletonGenerator` acceptable to `Task` by default.

sosi-deadeye:
[asyncio.sleep](https://github.com/micropython/micropython/blob/master/extmod/asyncio/core.py#L62) calls [asyncio.sleep_ms](https://github.com/micropython/micropython/blob/master/extmod/asyncio/core.py#L55). This is a normal function, not async. Internally, it uses the [asyncio.SingletonGenerator](https://github.com/micropython/micropython/blob/master/extmod/asyncio/core.py#L35) to accomplish the task. It was used to prevent memory allocation.

You could patch it, if you want.

**patch_aio_sleep.py**
```python
import asyncio

from asyncio import sleep as _sleep_generator


async def asleep(delay: float | int):
    await _sleep_generator(delay)


asyncio.sleep = asleep
del asleep
```

**main.py**
```python
import patch_aio_sleep
import asyncio
import time


async def worker():
    while True:
        await asyncio.sleep(1)
        print(time.time())


print("Direct call of asyncio.sleep(1)")
asyncio.run(asyncio.sleep(1))

print("Usage of await asyncio.sleep in an async function.")
asyncio.run(worker())
```

I guess that this code would allocate memory somewhere in the asyncio module. I think this is also the cause, why it wasn't changed yet.

jjsuwa-sys3175:
@sosi-deadeye 

> You could patch it, if you want.
> ** snip **
> I guess that this code would allocate memory somewhere in the asyncio module. I think this is also the cause, why it wasn't changed yet.

At the moment, I am already trying to avoid the issue by wrapping things up in such a way, and of course this pull request is based on that.

To further explain the reason for this pull request, this situation is not compatible with CPython.

```
import asyncio
asyncio.run(asyncio.sleep(2))
```
If you type the above into the CPy REPL it will wait 2 seconds with no issues.

PR #16661, opened 2025-01-28
docs: Note which ports have default or optional network.PPP support.

projectgus:
### Summary

Someone was asking on Discord why only esp32 had network.PPP support. This PR adds a note in the docs about availability on custom builds of stm32 and rp2 ports. It also adds the default values of these macros to the respective mpconfigport.h files, to improve discoverability.

### Testing

* Built stm32 port `PYBD_SF2` board with the new macro line changed to 1, verified build succeeded. It was also necessary to disable VFAT to save binary size.
* Built rp2 port `RPI_PICO_W` board with the new macro line changed to 1, verified build succeeded and it was possible to instantiate a `network.PPP` object.

* It's a bit fiddly to have these notes in the docs about piecemeal support, but it gives a better user experience than following the documentation and finding the support is missing.

### DIFF
--- a/docs/library/network.PPP.rst
+++ b/docs/library/network.PPP.rst
@@ -5,7 +5,12 @@ class PPP -- create network connections over serial PPP
 =======================================================
 
 This class allows you to create a network connection over a serial port using
-the PPP protocol.  It is only available on selected ports and boards.
+the PPP protocol.
+
+.. note:: Currently only the esp32 port has PPP support enabled in the default
+          firmware build. PPP support can be enabled in custom builds of the
+          stm32 and rp2 ports by enabling networking support and setting
+          ``MICROPY_PY_NETWORK_PPP_LWIP`` to 1.
 
 Example usage::
 
--- a/ports/stm32/mpconfigport.h
+++ b/ports/stm32/mpconfigport.h
@@ -146,14 +146,22 @@
 #define MICROPY_HW_SOFTSPI_MAX_BAUDRATE (HAL_RCC_GetSysClockFreq() / 48)
 #define MICROPY_PY_WEBSOCKET        (MICROPY_PY_LWIP)
 #define MICROPY_PY_WEBREPL          (MICROPY_PY_LWIP)
-#ifndef MICROPY_PY_SOCKET
-#define MICROPY_PY_SOCKET           (1)
-#endif
 #ifndef MICROPY_PY_NETWORK
 #define MICROPY_PY_NETWORK          (1)
 #endif
 #ifndef MICROPY_PY_ONEWIRE
 #define MICROPY_PY_ONEWIRE          (1)
+#endif
+
+// optional network features
+#if MICROPY_PY_NETWORK
+#ifndef MICROPY_PY_SOCKET
+#define MICROPY_PY_SOCKET           (1)
+#endif
+
+#ifndef MICROPY_PY_NETWORK_PPP_LWIP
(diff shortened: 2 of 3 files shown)


ports/stm32/mpconfigport.h
```diff
@@ -146,16 +146,24 @@
 #define MICROPY_HW_SOFTSPI_MAX_BAUDRATE (HAL_RCC_GetSysClockFreq() / 48)
 #define MICROPY_PY_WEBSOCKET        (MICROPY_PY_LWIP)
 #define MICROPY_PY_WEBREPL          (MICROPY_PY_LWIP)
-#ifndef MICROPY_PY_SOCKET
-#define MICROPY_PY_SOCKET           (1)
-#endif
```
projectgus, line 151: Have moved this macro as this is how it's laid out in the rp2/mpconfigport.h

Gadgetoid:
Can confirm we have been shipping PPP support in RP2 builds for some time now and it works well!

It's something of a shame that it's not in the build as standard, possibly something to consider for Pico 2 builds with more flash to spare

docs/library/network.PPP.rst
```diff
@@ -1,6 +1,11 @@
 .. currentmodule:: network
 .. _network.PPP:
 
+.. note:: Currently only the esp32 port has PPP support enabled in the default
+          build. PPP support can be enabled in custom builds of the stm32 and
+          rp2 ports by enabling networking support and setting
+          ``MICROPY_PY_NETWORK_PPP_LWIP`` to 1.
+
```
dpgeorge, line 8: IMO this should be moved down after the main heading, after the first paragraph.  Note that the first paragraph below already states "only available on selected ports".

dpgeorge:
> It's something of a shame that it's not in the build as standard, possibly something to consider for Pico 2 builds with more flash to spare.

Do you mean enable PPP on Pico 2 W, or Pico 2?  The latter does not even have networking at all, but I can see how it might be useful to enable it, sockets and PPP support.  We could consider doing that.

Gadgetoid:
> Do you mean enable PPP on Pico 2 W, or Pico 2?

Both. I think it's fair to say they have the flash to spare. CYW43 isn't the only game in town- there are Ethernet, LTE, LORA, 5GHz and WiFi modules out there which it would be nice to eventually not require custom builds for.

Network drivers implemented in pure MicroPython? 😆  (I did try this with "PINT" using the raw sockets network functions and hooking every call out to a method on a MicroPython class: https://github.com/pimoroni/pimoroni-pico/pull/937 - more an an exercise in experimenting with network devices than as anything shippable. It skips all the niceties of LwIP if I understand correctly.)

Tangerino:
This example does not work at all
```python
>>> ppp = network.PPP(uart)

ppp.connect()

while not ppp.isconnected():
    pass

print(ppp.ipconfig("addr4"))

# use the socket module as usual, etc

ppp.disconnect()
  File "<stdin>", line 3, in <module>
OSError: must be active
>>> 
```

dpgeorge:
> This example does not work at all

@Tangerino If PPP is not working, or the documentation is incorrect, please open a separate issue for that with more details, thanks.

docs/library/network.PPP.rst
projectgus: Have moved it below the first paragraph. I also removed the "selected ports and boards" sentence (suggest we add something like this back if we enable on some other ports & boards in the future.)

dpgeorge (approved):

PR #16664, opened 2025-01-28
wiznet: Use a pull-up on the interrupt line.

greezybacon:
### Summary

Without one, if there isn't an external pull-up on the board, the non-asserted state of the line is completed undefined. Since the interrupt condition is falling edge triggered, if the voltage of the INTn line is not considered above digital 0 when an interrupt occurs, then the interrupt is not triggered. But worse, the floating pin may be leading to over-servicing the interrupt when there is in fact no interrupt condition.

According to the schematics of the eval boards, there may not be a pull-up on the INTn line, so there should be one enabled on the MCU. If there is one additionally on the board, it's harmless to add an additional pull-up in the MCU.

- [W5100S-EVB-Pico](https://docs.wiznet.io/assets/images/w5100s-evb-pico_sch_v110-50b55eac13c594678a5c536a6d371111.jpg) does not have one
- [W5100S-EVB-Pico2](https://docs.wiznet.io/Product/iEthernet/W5100S/w5100s-evb-pico2#schematic-v10) does not have one
- [W5500-EVB-Pico](https://docs.wiznet.io/Product/iEthernet/W5500/w5500-evb-pico#schematic) has one
- [W5500-EVB-Pico2](https://docs.wiznet.io/Product/iEthernet/W5100S/w5100s-evb-pico2#schematic-v10) has one

Additionally, the [W5100S Datasheet](https://docs.wiznet.io/img/products/w5100s/w5100s-ds-v128e.pdf), page 14 indicates there is no internal pull-up on the INTn pin.

### Testing

I've been battling intermittent delays or what appears to be dropped packets with the W5100S for some time. This seems to finally resolve the issue.

Oddly, I don't see a HAL method to configure an input with a pull-up, so it looks like both function calls are needed?

### DIFF
--- a/extmod/network_wiznet5k.c
+++ b/extmod/network_wiznet5k.c
@@ -234,6 +234,7 @@ static void wiznet5k_init(void) {
         #endif
 
         mp_hal_pin_input(wiznet5k_obj.pin_intn);
+        mp_hal_pin_config(wiznet5k_obj.pin_intn, GPIO_IN, MP_HAL_PIN_PULL_UP, 0);
         wiznet5k_config_interrupt(true);
     }
 


andrewleech:
This looks like a great find, great job finding the issue! I know that feeling all too well when a chip is intermittently not quite working.

Curiously it looks to me like the `mp_hal_pin_config` function doesn't exist for all ports, only being defined for stm32,  rp2 and renesas. I'm surprised too that I can't see a standardised way to configure a pin for input with pull-up from a C function this way.

I'm not sure if this wiznet driver is being actively used on other ports like esp32 or samd, maybe this function missing is ok first now? Or maybe the C pin helper API for all ports should be extended with a mp_hal_pin_input_pullup() function or similar for this kind of use case.

greezybacon:
> Or maybe the C pin helper API for all ports should be extended with a mp_hal_pin_input_pullup() function or similar for this kind of use case.

Or simply `mp_hal_pin_set_pull` to configure the pull-up or pull-down resistor on an already-configured input.

dpgeorge:
Thanks for the fix.

Unfortunately this won't work on stm32.  You need to use `MP_HAL_PIN_MODE_INPUT` instead of `GPIO_IN`.

Note that the wiznet driver only works on rp2 and stm32 ports (in lwIP mode) due to the needed `mp_hal_pin_interrupt()` function.

> Oddly, I don't see a HAL method to configure an input with a pull-up,

I think we should add such a function.  It could be one of:
1. `mp_hal_pin_input_pullup(pin)`
2. `mp_hal_pin_input_with_pull(pin, pull)`
3. `mp_hal_pin_config(pin, mode, pull, alt)`

Number 3 already exists, so maybe just go with that?  That should already work on stm32, but on rp2 I think it needs to be fixed to do `gpio_set_function(pin, GPIO_FUNC_SIO);`.

greezybacon:
@dpgeorge @andrewleech after verifying this with a scope, I don't believe this makes any difference. The 5100S must have an undocumented pull-up resistor on the INTn line.

Should I close this and propose the configure-pin-with-pull function in a different PR?

PR #12008, opened 2023-07-14
py/persistentcode: Store constants more efficiently

arbrauns:
Instead of being stored as strings:
- Floats are stored as little-endian bytes
- Integers are stored as a length byte and the minimum number of little-endian bytes required.

This yields .mpy file space savings of 0.8% to 17% for the scripts I tested with.

PR #16666, opened 2025-01-28
py/parse: Add support for math module constants and float folding

yoctopuce:
### Summary

This is the first of four pull requests providing enhancements to the MicroPython parser, mainly targeting mpy-cross, with the aim of reducing the footprint of compiled mpy files to save flash and RAM. I have previously opened a discussion in the [MicroPython discussion forum](https://github.com/orgs/micropython/discussions/16617) and asked for comments.

The first new feature is to extend the use of compile-time `const()` expressions to some unhandled cases, that we found useful in the MicroPython implementation of our programming API. our programming API uses named constants to refer to specific cases, such as `INVALID_MEASURE`. For floating-point methods, this requires a definition such as:
```
_INVALID_MEASURE = const(math.nan)
```
or
```
_INVALID_MEASURE = const(-math.inf)
```
However, this is not supported by the current implementation, because `MICROPY_COMP_MODULE_CONST` and `MICROPY_COMP_CONST_FOLDING` are restricted to integer constants.

So we have introduced a new `MICROPY_COMP_FLOAT_CONST` feature which reuses the code of `MICROPY_COMP_CONST_FOLDING` to also support folding of floating point constants, and to include math module constants when `MICROPY_COMP_MODULE_CONST` is defined. This makes it possible to use compile-time math constants such as:
```
_DEG_TO_GRADIANT = const(math.pi/180)
_INVALID_VALUE = const(math.nan)
```
The commit explicitely enables this feature for mpy-cross, as it makes the most sense there, but is otherwise limited to ports using MICROPY_CONFIG_ROM_LEVEL_FULL_FEATURES.

### Testing

We have verified that the new code in mpy-cross works properly both on Windows and Linux. As target for running mpy code, we have been testing various Windows and Linux versions, as well as our custom board, which uses a Texas ARM Cortex processor very similar to the cc3200 port.

Micropython integration testing has found some tricky corner cases that have been solved:

- Constants 0.0 and -0.0 should not be merged during code emission, even though they are identical according to `==`
- The string based encoding used for floats in the `.mpy` file must be done carefully, as the `mp_parse_num_float()` used when loading the .mpy constants has some quirks (due to the use of a float to build the mantissa from the decimal form) which can cause a *decrease* in precision when adding more decimals. For instance, the number returned by `mp_parse_num_float()` when parsing 2.7182818284590451 is *smaller* than 2.718281828459045, and therefore less accurate to represent `math.e` although it should actually be closer. But relying on 16 decimal places to represent double-precision does not work in all cases, such as properly encoding/decoding 2.0**100. So we ended up checking at compile time if `mp_parse_num_float()` would give the exact same number using the shortest representation, and only adding an extra digit only if this is not the case. This empirical method seems to work for all test cases. 

Another way to solve the `mp_parse_num_float()` problem would have been to avoid the mantissa overflow altogether by using a pair of floats instead of a single float, but this would have required a change to the runtime code that is otherwise not needed by this pull request, and causing
1. an increase in code size in the runtime code
2. possible loss of precision when using new mpy-cross binaries with unpatched runtime

There are two qemu ports for which the integration tests show failed test cases, that appear to be related to this  `mp_parse_num_float()` problem. We could investigate these further  if you can provide information on how to reproduce this test environment.

This Pull Request only affects the code size of mpy-cross and ports using MICROPY_CONFIG_ROM_LEVEL_FULL_FEATURES, for which the negative impact of increased code size is unlikely to be relevant.

Folding floating-point expressions at compile time generally reduces the memory footprint of `.mpy` files, by saving some opcodes and even some qstr for referencing math constants. The saving is even greater for use cases like ours, where global definition such as `_INVALID_MEASURE = -math.inf` can be replaced by a compile-time `const()` expression, removing all references to the qstr `_INVALID_MEASURE`.


dpgeorge:
Thanks for the contribution!  At first glance this looks like a good enhancement.

Please can you add tests to get 100% coverage of the new code.

yoctopuce:
Sure, I will update to get full coverage.

I have also identified the issue with qemu arm integration test failing and will fix it.

Should I convert to 'draft' or is it fine to leave the pull request as-is in between ?

yoctopuce:
Followup on the qemu arm integration failure: the problem is indeed linked to the suboptimal result provided by `mp_parse_num_float()`. Contrarily to what I initially thought, the problem is not linked to a mantissa overflow but to a round-up correction after multiplying by the exponent. 

I wrote a small piece of test code that compares the value provide by `mp_parse_num_float()` to the value computed by casting the result of bigint arithmetic to a float, which is expected to provided the closest value. The two resulting values are displayed using a format string showing extra digits to put the difference in evidence. As illustrated below run with MICROPY_FLOAT_IMPL_FLOAT, `mp_parse_num_float()` currently fails to provide the best floating point number even in some very simple cases:

```
float('1.2e30'):
    => 1.2e+30
    vs 1.2e+30
float('1.26e30'):
    => 1.2599999020249624675e+30  FAIL
    vs 1.2599999775828091928e+30
float('1.267e30'):
    => 1.2669999610545268366e+30
    vs 1.2669999610545268366e+30
float('1.2676e30'):
    => 1.2676000416158554855e+30  FAIL
    vs 1.2675999660720408409e+30
float('1.26765e30'):
    => 1.2676499097910229878e+30  FAIL
    vs 1.2676499853488888585e+30
float('1.267650e30'):
    => 1.2676499097910229878e+30  FAIL
    vs 1.2676499853488888585e+30
float('1.2676506e30'):
    => 1.2676505898138652462e+30
    vs 1.2676505898138652462e+30
float('1.26765060e30'):
    => 1.2676505898138652462e+30
    vs 1.2676505898138652462e+30
float('1.267650600e30'):
    => 1.2676505898138652462e+30
    vs 1.2676505898138652462e+30
float('1.2676506002e30'):
    => 1.2676505898138652462e+30
    vs 1.2676505898138652462e+30
float('1.26765060022e30'):
    => 1.2676505142560004287e+30  FAIL
    vs 1.2676505898138652462e+30
float('1.267650600228e30'):
    => 1.2676505142560004287e+30  FAIL
    vs 1.2676505898138652462e+30
float('1.2676506002282e30'):
    => 1.2676505142560004287e+30  FAIL
    vs 1.2676505898138652462e+30
float('1.26765060022822e30'):
    => 1.2676505898138652462e+30
    vs 1.2676505898138652462e+30
float('1.267650600228229e30'):
    => 1.2676504386981526132e+30  FAIL
    vs 1.2676505898138652462e+30
```

I will see if I can fix that rounding issue properly.

yoctopuce:
The code has been improved since previous review, and coverage fixed.

The two outstanding failed checks are due to inaccurate float parsing when compiling with MICROPY_FLOAT_IMPL_FLOAT, but they should be fixed once my other pull request [py/parsenum.c: reduce code footprint of mp_parse_num_float. #16672](https://github.com/micropython/micropython/pull/16672) is integrated

yoctopuce:
As mentionned in my previous comment, as this pull request depends on the improvement in parsenum.c in another pull request (https://github.com/micropython/micropython/pull/16672) , I have rebased it accordingly so that the unit test checks show the actual status once pulled in.

yoctopuce:
Rebased to master now that https://github.com/micropython/micropython/pull/16672 has been merged in.

yoctopuce:
rebased to head revision

py/parse.c
```diff
@@ -642,12 +664,32 @@ static const mp_rom_map_elem_t mp_constants_table[] = {
...
+    #endif
     // Extra constants as defined by a port
     MICROPY_PORT_CONSTANTS
 };
 static MP_DEFINE_CONST_MAP(mp_constants_map, mp_constants_table);
 #endif
 
+static bool binary_op_maybe(mp_binary_op_t op, mp_obj_t lhs, mp_obj_t rhs, mp_obj_t *res) {
+    nlr_buf_t nlr;
+    if (nlr_push(&nlr) == 0) {
```
dpgeorge, line 678: This `nlr_push` is relatively expensive.

What are the cases in which it's needed?  From what I can gather it's:
- divide by zero (both int and float)
- when complex numbers are produced without complex enabled (float only)
- negative shift (int only)
- negative power without float enabled (int only)
- unsupported operation (eg shifting a float)

The int cases are already handled explicitly.  Maybe it's not much extra code to explicitly guard against the invalid float operations, so that this `nlr_push` protection is not needed?

It's a bit of a trade-off here, whether to use `nlr_push` and let `mp_binary_op` dictate what's valid and what isn't, or to make the code in the parser here know what's valid.

py/parse.c
```diff
@@ -866,7 +911,7 @@ static bool fold_constants(parser_t *parser, uint8_t rule_id, size_t num_args) {
         }
         mp_obj_t dest[2];
         mp_load_method_maybe(elem->value, q_attr, dest);
-        if (!(dest[0] != MP_OBJ_NULL && mp_obj_is_int(dest[0]) && dest[1] == MP_OBJ_NULL)) {
+        if (!(dest[0] != MP_OBJ_NULL && (mp_obj_is_int(dest[0]) || mp_obj_is_float(dest[0])) && dest[1] == MP_OBJ_NULL)) {
```
dpgeorge, line 914: There's a comment a few lines up that should now road "see if it can be replaced with an integer or float".

tests/micropython/const_error.py
```diff
@@ -18,9 +18,13 @@ def test_syntax(code):
...
 test_syntax("A = const(1 << -2)")
 test_syntax("A = const(1 >> -2)")
 test_syntax("A = const(1 % 0)")
 test_syntax("A = const(1 // 0)")
+
+# Expressions below are supported if MICROPY_COMP_FLOAT_CONST is set.
+# They should not anymore be expected to always fail:
+#
+# test_syntax("A = const(1 / 2)")
+# test_syntax("A = const(1 ** -2)")
```
dpgeorge, line 30: I think it's worth putting these in a new test file `tests/micropython/const_float.py`, along with some other basic tests of float constant folding.

py/mpconfig.h
```diff
@@ -480,6 +480,11 @@
 #define MICROPY_COMP_CONST (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
 #endif
 
+// Whether to enable float module constants lookup and folding; eg const(-math.inf)
+#ifndef MICROPY_COMP_FLOAT_CONST
+#define MICROPY_COMP_FLOAT_CONST (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_FULL_FEATURES)
```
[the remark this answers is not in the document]
dpgeorge: Just to clarify: if a build has floats enabled and constant folding, it's highly likely that constant folding of floats would be desired and isn't much more code compared to how much code floats bring in already.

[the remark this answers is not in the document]
yoctopuce: Thanks for the nice comment :-)
I have made the suggested changes

py/objfloat.c
```diff
@@ -124,6 +124,18 @@ static void float_print(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t
...
     #endif
     mp_format_float(o_val, buf, sizeof(buf), 'g', precision, '\0');
+    #if MICROPY_COMP_FLOAT_CONST
+    if (kind == PRINT_REPR) {
+        // For .mpy files, try to add an extra (partial) digit for precision.
+        // If the length increases by more than one digit, it means we are
+        // digging too far (eg. 1.234499999999), so we keep the short form
+        char ebuf[32];
+        mp_format_float(o_val, ebuf, sizeof(ebuf), 'g', precision + 1, '\0');
+        if (strlen(ebuf) - strlen(buf) == 1) {
```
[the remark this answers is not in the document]
yoctopuce: Fully agree, done.

py/parse.c
yoctopuce: I thought about that initially, but I was afraid to miss a corner case which would then break compilation. For instance, on builds with limited integers (`longlong`),  an exception might be raised if the result of an operation is beyond the supported `longlong` range. I was afraid that going into such detailled tests somehow breaks the modularity of MicroPython, hence my preference for catching the exception.

Are you mostly concerned by the impact on stack space, or more on compilation time ?

py/parse.c
yoctopuce: Good point, done.

tests/micropython/const_error.py
yoctopuce: I have added two test files, one for simple const floats and one for const expression referencing math constants (if math is available).
I have also added a note to point out the the const folding code is actually mostly tested by running the full coverage tests via mpy, which involves lots of float constant folding.

dpgeorge: > I have also added a note to point out the the const folding code is actually mostly tested by running the full coverage tests via mpy, which involves lots of float constant folding.

Does it really?  Going via mpy doesn't change the parsing or compilation process.  Can you point to a specific existing test that tests float folding?

py/builtin.h
```diff
@@ -138,6 +138,9 @@ extern const mp_obj_module_t mp_module_sys;
 extern const mp_obj_module_t mp_module_errno;
 extern const mp_obj_module_t mp_module_uctypes;
 extern const mp_obj_module_t mp_module_machine;
+#if MICROPY_PY_MATH
```
dpgeorge, line 141: I don't think this guard is worth putting here.  The other modules just above don't have equivalent guards.

py/mpconfig.h
```diff
@@ -480,6 +480,11 @@
 #define MICROPY_COMP_CONST (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
 #endif
 
+// Whether to enable float module constants lookup and folding; eg const(-math.inf)
```
dpgeorge, line 483: I would make this comment a little more detailed:
```
// Whether to enable float constant folding like 1.2+3.4 (when MICROPY_COMP_CONST_FOLDING is also enabled)
// and constant optimisation like id = const(1.2) (when MICROPY_COMP_CONST is also enabled)
// and constant lookup like math.inf (when MICROPY_COMP_MODULE_CONST is also enabled)
```
(I think I understood the logic there?)

tests/micropython/const_error.py
```diff
@@ -18,9 +18,13 @@ def test_syntax(code):
...
 # these operations are not supported within const
 test_syntax("A = const(1 @ 2)")
-test_syntax("A = const(1 / 2)")
-test_syntax("A = const(1 ** -2)")
 test_syntax("A = const(1 << -2)")
 test_syntax("A = const(1 >> -2)")
 test_syntax("A = const(1 % 0)")
 test_syntax("A = const(1 // 0)")
+
+# Expressions below are supported if MICROPY_COMP_FLOAT_CONST is set.
```
dpgeorge, line 26: I think these two test cases can just be moved to `const_float.py`.

tests/float/float_parse_doubleprec.py
```diff
@@ -19,3 +19,13 @@
 print(float("1.00000000000000000000e-307"))
 print(float("10.0000000000000000000e-308"))
 print(float("100.000000000000000000e-309"))
+
+# ensure repr() adds an extra digit when needed for accurate parsing
+if float(repr(2.0**100)) != (2.0**100):
+    # this test is needed for coverage, but it will
+    # only work if MICROPY_COMP_FLOAT_CONST == 1,
```
dpgeorge, line 26: Should be `MICROPY_COMP_CONST_FLOAT`.

py/objfloat.c
```diff
@@ -124,6 +124,20 @@ static void float_print(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t
...
     mp_format_float(o_val, buf, sizeof(buf), 'g', precision, '\0');
+    #if MICROPY_COMP_CONST_FLOAT
+    if (kind == PRINT_REPR) {
+        // See if we can add an extra (partial) digit for precision.
+        // This is useful in particular for saving floats to .mpy files,
+        // but also for general `repr` support.
+        // If the length increases by more than one digit, it means we are
+        // digging too far (eg. 1.234499999999), so we keep the short form
+        char ebuf[32];
+        mp_format_float(o_val, ebuf, sizeof(ebuf), 'g', precision + 1, '\0');
```
dpgeorge, line 135: I'm still trying to understand why this code is necessary.

Is it only necessary when floats are double precision?  Or also single precision?

Can you give an example of a number that needs this additional digit of precision?

py/parse.c
```diff
@@ -336,6 +336,26 @@ static uint8_t peek_rule(parser_t *parser, size_t n) {
 }
 #endif
 
+#if MICROPY_COMP_CONST_FLOAT
+static bool mp_parse_node_get_number_maybe(mp_parse_node_t pn, mp_obj_t *o) {
```
dpgeorge, line 340: In order to eliminate code duplication below, I suggest:
```c
static bool mp_parse_node_get_number_maybe(mp_parse_node_t pn, mp_obj_t *o) {
    if (MP_PARSE_NODE_IS_SMALL_INT(pn)) {                               
        *o = MP_OBJ_NEW_SMALL_INT(MP_PARSE_NODE_LEAF_SMALL_INT(pn));
        return true;         
    } else if (MP_PARSE_NODE_IS_STRUCT_KIND(pn, RULE_const_object)) {
        mp_parse_node_struct_t *pns = (mp_parse_node_struct_t *)pn;
        *o = mp_parse_node_extract_const_object(pns);     
        return mp_obj_is_int(*o)                        
            #if MICROPY_COMP_CONST_FLOAT
            || mp_obj_is_float(*o) 
            #endif
            ;
    } else {       
        return false;
    }                                    
}                                           
bool mp_parse_node_get_int_maybe(mp_parse_node_t pn, mp_obj_t *o) {
    return mp_parse_node_get_number_maybe(pn, o)
        #if MICROPY_COMP_CONST_FLOAT                   
        && mp_obj_is_int(*o)                                   
        #endif                                           
        ;                    
}                      
```

py/parse.c
```diff
@@ -732,60 +774,63 @@ static bool fold_constants(parser_t *parser, uint8_t rule_id, size_t num_args) {
...
-                }
+            if (!binary_op_maybe(op, arg0, arg1, &arg0)) {
+                return false;
             }
-            arg0 = mp_binary_op(op, arg0, arg1);
         }
     } else if (rule_id == RULE_factor_2) {
         // folding for unary ops: + - ~
         mp_parse_node_t pn = peek_result(parser, 0);
-        if (!mp_parse_node_get_int_maybe(pn, &arg0)) {
```
dpgeorge, line 781: You could leave this here, change it to `get_number_maybe`, and then in the `MP_TOKEN_OP_TILDE` path below do a `if (!mp_obj_is_int(arg0)) { return false; }`.

That might lead to slightly smaller code size.

py/mpconfig.h
yoctopuce: Absolutely.

tests/micropython/const_error.py
yoctopuce: OK

py/builtin.h
dpgeorge: > I usually put the same `#if` in the declaration as in the definition file so that undefined symbols are caught at compilation time rather than at link time

Yes, that is done in some places in the MicroPython code base.

But here IMO it's better to stick with local coding style to get local consistency.

py/objfloat.c
yoctopuce: The numbers that require this extra digit are not the the same for single-precision and double-precision (as the rounding effects differs), but a typical example for double-precision is `2.0 ** 100`, which is the test case I have added to `float_parse_doubleprec.py`.  Without the improved `repr` code, number which _appear_ to be the same as per `repr()` would actually not match with regards to == operator:
```
>>> n = float('2.0')**100
>>> n2 = float(repr(n))
>>> n
1.267650600228229e+30
>>> n2
1.267650600228229e+30
>>> n2 == n
False
```
With the improved `repr` code, we get the expected behaviour:
```
>>> n = float('2.0')**100
>>> n2 = float(repr(n))
>>> n
1.2676506002282294e+30
>>> n2
1.2676506002282294e+30
>>> n2 == n
True
```
With float folding enabled, large expanded decimal numbers appear much quicker `mpy` files and this can quickly cause functional differences. Without this improved repr code, the coverage tests fail in `float_float2int_intbig` due to the missing _half-digit_.

py/parse.c
yoctopuce: Sure, I will use that version

py/parse.c
yoctopuce: Sure, will do.

yoctopuce:
@dpgeorge I've finished implementing your suggestions, and rebased on HEAD. Thank you so much for your attention to every detail. It's really what makes MicroPython such a great project!

dpgeorge:
> I've finished implementing your suggestions, and rebased on HEAD. Thank you so much for your attention to every detail. It's really what makes MicroPython such a great project!

Thanks for updating, and thanks for the kind words!

Note that even though the CI is all green, it's actually failing when building `bare-arm`:
```
../../py/parse.c:339:13: error: 'mp_parse_node_get_number_maybe' defined but not used [-Werror=unused-function]
  339 | static bool mp_parse_node_get_number_maybe(mp_parse_node_t pn, mp_obj_t *o) {
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
That'll need fixing.  (And we'll need to fix the CI so such errors show up...)

py/objfloat.c
dpgeorge: OK, thanks for the info.  I now understand the problem.  Let me restate it in two ways:
1. Printing a double with 16 digits of precision is not enough to represent every number.  17 digits is a little bit more than needed.
2. There exist multiple different numbers whose repr at 16 digits is the same.  At 17 digits they all differ.

MicroPython at the moment uses 16 digits which is not enough to faithfully represent doubles.

------

So we do need to fix something here, but it's not clear what or how.  My testing shows that in CPython doing `'{:.17g}'.format(n)` is enough digits to properly represent every number.  Ie:
```py
n == float('{:.17g}'.format(n))
```
The issue is that in MicroPython the above is not true.  In fact there's no repr precision in MicroPython that allows you to satisfy the above for all `n` (eg even `.20g` doesn't work).

For some numbers, eg `39.0**100`, the above is true in MicroPython for precision of 17 and 19 but false for 16, 18 and 20!

Numbers like `12.0**100` and `18.0**100` get worse going from 16 to 17 digits (in MicroPython).

(And others like `40.0**100` are accurate at 17, but at 16 are printed only with 15, and so the length increases by 2 digits to get full accuracy.)

The real issue here is that MicroPython's `mp_format_float()` code is inaccurate.  Eg:
```py
$ python -c "print(('{:.22g}'.format(2.0**100)))" 
1.267650600228229401497e+30
$ micropython -c "print(('{:.22g}'.format(2.0**100)))"
1.267650600228229428959e+30
```
That's a problem!  It means we can't really use `mp_format_float()` to store floats in .mpy files, because it's inaccurate.

----

What to do?

There is #12008 which stores floats as binary in .mpy files.  That's definitely one way to fix it, but I don't want to change the .mpy version number at this stage.

We could try to improve `mp_format_float()`, but I think that's very difficult, and probably needs a rewrite to make it more accurate.

Note that currently (without this PR) certain floats are already stored inaccurately in .mpy files, due to using 16 digit precision only.  So does this PR make things worse?  I think yes, because now more floats have the chance to be store and hence stored inaccurately.  Eg those in `tests/float/float2int_intbig.py`.

That said, I don't think this PR (constant folding) will make things that much worse in .mpy files.

My suggestion to move forward is:
1. Switch to using 17 digits of precision for all repr printing (and I guess 7 or 8 for single precision).  And don't worry about things like `1.2345` rendering as `1.23499999999`.  I think if you use `repr(a_float)` then you want accuracy.  If you want a set number of digits, you'd be using a precision specifier like `'{:.10g}'.format(a_float)`. (It's pretty expensive to call `mp_format_float()` so I'd rather not do it twice.)
2. Accept that floats in .mpy files are possibly inaccurate.  Users can use `array.array` with binary representation, or escape the const folding with eg `float("2.0") + 3.4` if they need 100% accuracy.
3. In the future, try to improve things.

If you don't like (1) because it makes `1.2345` print out as `1.234999999` at the REPL, then maybe we can adjust the code in `py/persistentcode.c:save_obj` to save floats explicitly with a higher precision (17 digits for double).  (From my testing, always using 17 digits is slightly better than using 17 only if it grows 16 by one digit.)

(I didn't investigate single precision floats, but I assume all the above logic/reasoning applies there as well.)

yoctopuce: Your understanding is correct, and indeed, single precision runs into the exact same kind of issues.

As you have guessed, I am not a fan of `1.234499999999`, but if your testing shows that using 17 digits provides overall better results than conditionally shorting to 16, then this is the way to go.

dpgeorge: I made a test:
```py
import os, random, array, math
stats = [0, 0, 0, 0, 0]   
N = 10000000
for _ in range(N):
    #f = random.random()       
    f = array.array("d", os.urandom(8))[0]
    while f == math.isinf(f) or math.isnan(f):
        f = array.array("d", os.urandom(8))[0]
    str_f_repr = repr(f)           
    str_f_16g = "{:.16g}".format(f)                
    str_f_17g = "{:.17g}".format(f)
    str_f_18g = "{:.18g}".format(f)                                       
    str_f_19g = "{:.19g}".format(f)                                        
    f_repr = float(str_f_repr)                                      
    f_16g = float(str_f_16g)                                            
    f_17g = float(str_f_17g)
    f_18g = float(str_f_18g)
    f_19g = float(str_f_19g)
    stats[0] += f_repr == f                                           
    stats[1] += f_16g == f                                        
    stats[2] += f_17g == f                                              
    stats[3] += f_18g == f                                        
    stats[4] += f_19g == f             
print(list(100 * s / N for s in stats))
```
That generates 10 million random doubles and checks `repr` and formatting with 16 through 19 digits of precision, to see if the rendered number is equivalent to the original.

On Python it gives:
```
[100.0, 54.61923, 100.0, 100.0, 100.0]
```
So that means 100% of the numbers are accurate using repr and 17 or more digits of precision.  Using 16 digits, only 54.6% represent correctly.

MicroPython unix port with this PR (ie repr using 16 or 17 digits) gives:
```
[52.24606, 37.99564, 53.94176, 54.37388, 54.12582]
```
That shows ... it's pretty terrible.  But at least we see that using repr with this PR is much better than master which just uses 16 digits: 52.2% for repr vs 38% for always using 16 digits.  But also, always using 17 digits gives a little more accuracy than repr, in this case 53.9% which is about 1.7% more than repr.

Note that MicroPython may also have inaccuracies converting str to float, so that may add to the errors here.  But that's representative of .mpy loading because the loader uses the same float parsing code.

Also note that MicroPython was faster than CPython in the above test!  That might be because MicroPython's float printing code is taking shortcuts.  It also means that rendering the float twice in repr to try and extract the extra half digit is not taking up that much extra time.

Open questions:
- is it worth the extra 1% or so accuracy always showing 17 digits, or is it better to show a nicer number and doing the 16/17 trick?
- is it worth trying to improve `mp_float_format()` using a different algorithm, eg fully integer based?

dpgeorge: I tried extending the 16/17 trick to the following:
```diff
--- a/py/objfloat.c
+++ b/py/objfloat.c
@@ -133,7 +133,7 @@ static void float_print(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t
         // digging too far (eg. 1.234499999999), so we keep the short form
         char ebuf[32];
         mp_format_float(o_val, ebuf, sizeof(ebuf), 'g', precision + 1, '\0');
-        if (strlen(ebuf) == strlen(buf) + 1) {
+        if (strlen(ebuf) - strlen(buf) <= 3) {
             memcpy(buf, ebuf, sizeof(buf));
         }
     }
```
That uses 17 digits if it extends up to 3 digits more.  That gives better results in the above test:
```
[53.85271, 37.94894, 53.94477, 54.36183, 54.16032]
```
That means `repr` is almost as good as unconditionally doing 17 digits, 53.85% vs 53.94%.  But note that these percentages change a bit across runs, due to the randomness.

yoctopuce: I have been working on this today. It is now clear to me as well that the proper solution is indeed to fix `mp_float_format()`, rather than adding a quick fix in `repr`.

I have implemented today an alternate algorithm for `mp_float_format()`, based on the same idea as my previous enhancement to `mp_parse_num_float`: using an `mp_float_uint_t` to convert the mantissa, rather than working directly on floats. This brings the correct percentage up to 60%, which is a good start.

Once the mantissa is computed using an integer, it is possible _fix it_  incrementally to ensure that the parse algorithm will get back to the original number. I have been working on that, and got up to 95% correct conversions using 18 digits. I still have a bit of work to fix corner cases (forced round-ups) before I can push that code.

Do you want me to send the new `mp_float_format` code in a separate PR ?

dpgeorge: > I have implemented today an alternate algorithm for `mp_float_format()`, based on the same idea as my previous enhancement to `mp_parse_num_float`: using an `mp_float_uint_t` to convert the mantissa, rather than working directly on floats.

Wow, very nice!

> Do you want me to send the new `mp_float_format` code in a separate PR ?

Yes please.

dpgeorge: This should now be fully resolved with the new float formatting code.

dpgeorge:
This can now be rebased on the latest master, to pick up the new float formatting code.

py/objfloat.c
yoctopuce: Absolutely, this is where it all started...

yoctopuce:
I will take care of that right now

py/objfloat.c
dpgeorge: This was a very good digression, printing floats is much, much better now.

And the change here is completely gone, which is great!

py/parse.c
dpgeorge: > I was afraid to miss a corner case which would then break compilation.

Yes, you're right, there could easily be corner cases which are missed.

> Are you mostly concerned by the impact on stack space, or more on compilation time ?

Stack space, and also performance.  But since this is in the parser/compiler it doesn't matter much for performance.  And even stack space is not critical because compilation usually happens only at the top level of code, not in deep recursion.

So, let's keep it using the NLR.  That should anyway lead to smaller code, which is good.

tests/micropython/const_error.py
dpgeorge: > By the way, this dump shows that float folding shortcuts the true intent of some of these tests, as some float operations were supposed to execute in runtime but are now executed at compile-time.

That's correct.

But, at the same time, they are still executed "at runtime" on the target, and by the same code as before (eg `float_binary_op`).  The only difference is that the computation is done at compile time instead of when the bytecode is executed.

I think it's OK, because we still get the same coverage of the same functions running the same tests.

A difference will be when running a test via mpy files.  Then the computation is done on the host.  So these mpy tests change and now exercise mpy-cross more and also saving/loading of floats in the mpy file.  That's probably a good thing, to test that more.

In the future we could add some tests for explicit float computation during bytecode execution (because maybe that somehow differs to doing it during compilation) but for now I'm happy that we still have the same coverage of functions like `float_binary_op` running on the target.

dpgeorge (approved):
This is looking really good now with the new float format changes (that are already merged).

This PR is now relatively simple, and the code size increase has dropped a lot.  Basically what's done here is (1) allow floats and ints in constant folding, and (2) use NLR to protect the binary_op, and don't fold the constant if that call fails.

Very simple and very effective!

PR #16668, opened 2025-01-29
doc: Add documentation for SPI Ethernet devices on esp32 port.

projectgus:
### Summary

This was inspired by some confusion on Discord about how to use WIZnet5K on ESP32.

As well as adding documentation for the SPI Ethernet "mode" of `network.LAN` on ESP32, this PR adds some cross-reference links between related network documentation pages.

### Testing

* I set up an ESP32-S3 with a W5500 breakout and ran the sample code which has been added as part of the new doc section.

* We could try to move this documentation out of esp32 quickref and into the `network.LAN` doc, but I think it'll be even messier to do it that way.

### DIFF
--- a/docs/esp32/quickref.rst
+++ b/docs/esp32/quickref.rst
@@ -121,11 +121,20 @@ calling ``wlan.config(reconnects=n)``, where n are the number of desired reconne
 attempts (0 means it won't retry, -1 will restore the default behaviour of trying
 to reconnect forever).
 
+.. _esp32_network_lan:
+
 LAN
 ^^^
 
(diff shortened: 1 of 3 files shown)


dpgeorge (approved):
Thanks, this does clear things up.  I now understand how LAN works on esp32 a bit better!

(As you noted in the docs, `network.LAN` is configured differently for different ports, and also Wiznet support is different comparing esp32 to non-esp32 ports.  Would be great to eventually make that all consistent, probably following how esp32 does it with `phy_type=PHY_WW5500`.)

dpgeorge (approved):

PR #16671, opened 2025-01-29
feature/optional_i2c_bus

malcolm-sparkfun:
### Summary

This pull request gives the option to not pass an I2C Bus ID when creating a machine I2C object on RP2 and ESP32. If the ID is not provided on RP2, the default bus ID (PICO_DEFAULT_I2C) is used. For ESP32, I2C_NUM_0  is used.  This would allow users on both platforms to simply declare an I2C object with machine.I2C() without passing any arguments, thus creating an object with the default I2C ID, SCL, and SDA. 

Our use case is a higher-level python driver that creates machine.I2C objects. We would like the driver to be portable to many different RP2 and ESP32 boards without having to explicitly specify these arguments. 

### Testing
Tested on RP2 with Raspberry Pi Pico W:

- Created machine.I2C object by passing no arguments to I2C(). Verified that it successfully could use the scan() method to detect a connected sensor.
- Exercised I2C object by using it with an example program performing reads and writes to a sensor.

Tested on ESP32 with [Sparkfun Thing Plus - ESP32 WROOM (Micro-B)](https://www.sparkfun.com/sparkfun-thing-plus-esp32-wroom-micro-b.html):

- Created machine.I2C object by passing no arguments to I2C(). Verified that it successfully could use the scan() method to detect a connected sensor.
- Exercised I2C object by using it with an example program performing reads and writes to a sensor.

- For ESP32, this required removal of the MP_MACHINE_I2C_CHECK_FOR_LEGACY_SOFTI2C_CONSTRUCTION() macro in machine_hw_i2c_make_new(). This macro would cause the I2C object to be created with an ID of -1 if not explicitly specified. This would cause a SoftwareI2C object to be created. Thus, this could break backwards compatibility for ESP32 if users have code where they don't specify the ID (or specify it as -1) and expect that to lead to creation of a SoftwareI2C object. However, this is a legacy construction and likely should be removed anyways. It is a more intuitive behavior for creation of an I2C object with no ID argument to get an I2C object with default ID rather than an entirely different SoftwareI2C object. 
- As with any default argument that is optionally supplied by a user, making it optional means that it may be more frequently overlooked. The default may sometimes be used when the user actually should look into it deeply enough to decide which ID/Bus they want to use.

### DIFF
--- a/ports/esp32/machine_i2c.c
+++ b/ports/esp32/machine_i2c.c
@@ -146,12 +146,15 @@ static void machine_hw_i2c_print(const mp_print_t *print, mp_obj_t self_in, mp_p
 }
 
 mp_obj_t machine_hw_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
-    MP_MACHINE_I2C_CHECK_FOR_LEGACY_SOFTI2C_CONSTRUCTION(n_args, n_kw, all_args);
+    // Create a SoftI2C instance if no id is specified (or is -1) but other arguments are given
+    if (n_args != 0) {
+        MP_MACHINE_I2C_CHECK_FOR_LEGACY_SOFTI2C_CONSTRUCTION(n_args, n_kw, all_args);
+    }
 
     // Parse args
     enum { ARG_id, ARG_scl, ARG_sda, ARG_freq, ARG_timeout };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id, MP_ARG_REQUIRED | MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
+        { MP_QSTR_id, MP_ARG_INT, {.u_int = I2C_NUM_0} },
         { MP_QSTR_scl, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
         { MP_QSTR_sda, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
         { MP_QSTR_freq, MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = 400000} },
@@ -161,7 +164,9 @@ mp_obj_t machine_hw_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_
     mp_arg_parse_all_kw_array(n_args, n_kw, all_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
 
     // Get I2C bus
-    mp_int_t i2c_id = mp_obj_get_int(args[ARG_id].u_obj);
+    mp_int_t i2c_id = args[ARG_id].u_int;
+
+    // Check if the I2C bus is valid
     if (!(I2C_NUM_0 <= i2c_id && i2c_id < I2C_NUM_MAX)) {
         mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("I2C(%d) doesn't exist"), i2c_id);
     }
--- a/ports/rp2/machine_i2c.c
+++ b/ports/rp2/machine_i2c.c
@@ -97,7 +97,11 @@ static void machine_i2c_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
 mp_obj_t machine_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
     enum { ARG_id, ARG_freq, ARG_scl, ARG_sda, ARG_timeout };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id, MP_ARG_REQUIRED | MP_ARG_OBJ },
+        #ifdef PICO_DEFAULT_I2C
+        { MP_QSTR_id, MP_ARG_INT, {.u_int = PICO_DEFAULT_I2C} },
+        #else
+        { MP_QSTR_id, MP_ARG_INT, {.u_int = -1} },
+        #endif
         { MP_QSTR_freq, MP_ARG_INT, {.u_int = DEFAULT_I2C_FREQ} },
         { MP_QSTR_scl, MICROPY_I2C_PINS_ARG_OPTS | MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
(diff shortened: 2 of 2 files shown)


ports/esp32/machine_i2c.c
```diff
@@ -146,12 +146,11 @@ static void machine_hw_i2c_print(const mp_print_t *print, mp_obj_t self_in, mp_p
 }
 
 mp_obj_t machine_hw_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
-    MP_MACHINE_I2C_CHECK_FOR_LEGACY_SOFTI2C_CONSTRUCTION(n_args, n_kw, all_args);
 
     // Parse args
     enum { ARG_id, ARG_scl, ARG_sda, ARG_freq, ARG_timeout };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id, MP_ARG_REQUIRED | MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
+        { MP_QSTR_id, MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
```
Gadgetoid, line 153: You can use `.u_int = I2C_NUM_0` here to save the default and `MP_OBJ_NULL` checks below.

ports/rp2/machine_i2c.c
```diff
@@ -97,7 +97,7 @@ static void machine_i2c_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
 mp_obj_t machine_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
     enum { ARG_id, ARG_freq, ARG_scl, ARG_sda, ARG_timeout };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id, MP_ARG_REQUIRED | MP_ARG_OBJ },
+        { MP_QSTR_id,  MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
```
Gadgetoid, line 100: Still not ideal but I'd probably do:

```python
#ifdef PICO_DEFAULT_I2C
    { MP_QSTR_id,  MP_ARG_INT, {.u_int = PICO_DEFAULT_I2C} },
#else
    { MP_QSTR_id,  MP_ARG_INT, {.u_int = -1} },
#endif
```

Rather than inline the checks below. Saves code.

(Not really sure why this was ever `MP_ARG_OBJ`?)

It doesn't feel like this breaks the API, but it seems to break expectations on Pico since - afaict - `machine.I2C()` will not give you a properly configured default I2C instance. (It doesn't handle `PICO_DEFAULT_I2C_SCL_PIN` and `PICO_DEFAULT_I2C_SDA_PIN` so I2C will wiggle into the void and thus your objective - portable code - wont quite be met?)

Gadgetoid:
I kinda like where you're going with this, though afaict it doesn't make `machine.I2C()` work out of the box on the Pico since, by default, no I2C pins are configured. I should probably poke that because it doesn't *feel* like I'm right, even though my foggy Monday morning sniffing around the code suggests as much.

Of course, to play devil's advocate here, your library shouldn't construct `machine.I2C()` instances internally at all- they should be user supplied! (and yes I say this with the full weight of understanding that I am throwing stones in a 12 year old glass house 😆)

dpgeorge:
I actually quite like this idea.  After all, you can already do things like `network.WLAN()` to get the default WLAN interface (which is usually STA, ie it's equivalent to `network.WLAN(network.WLAN.IF_STA)`).

> though afaict it doesn't make `machine.I2C()` work out of the box on the Pico since, by default, no I2C pins are configured

I think it will work on Pico.  There are default pins for all I2C objects which should get configured the first time `machine.I2C()` is instantiated.

> Of course, to play devil's advocate here, your library shouldn't construct `machine.I2C()` instances internally at all- they should be user supplied!

I agree with this.  A driver usually doesn't need to `import machine` at all.

Gadgetoid:
> I think it will work on Pico. There are default pins for all I2C objects which should get configured the first time `machine.I2C()` is instantiated.

Oof I think I see how this works, falls through to the check for `freq == 0` (the default value in `machine_i2c_obj`) and initialises `MICROPY_HW_I2Cn_SCL` and `MICROPY_HW_I2Cn_SDA`.

There *is* a comment saying as much 😆 but I can see why I missed that.

malcolm-sparkfun:
Thank you for your feedback. I have made the suggested changes. Let me know if there's anything else I should do before this is merged.

dpgeorge (approved):

ports/esp32/machine_i2c.c
```diff
@@ -146,12 +146,11 @@ static void machine_hw_i2c_print(const mp_print_t *print, mp_obj_t self_in, mp_p
 }
 
 mp_obj_t machine_hw_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
-    MP_MACHINE_I2C_CHECK_FOR_LEGACY_SOFTI2C_CONSTRUCTION(n_args, n_kw, all_args);
```
dpgeorge, line 149: I think we should still keep this legacy constructor behaviour.  And to get the new default-instance behaviour working as well, you can do:
```c
if (n_args != 0) {
    MP_MACHINE_I2C_CHECK_FOR_LEGACY_SOFTI2C_CONSTRUCTION(n_args, n_kw, all_args);
}
```
The cases are:
- `machine.I2C()` previously tried to create a SoftI2C instance but would fail because scl/sda are required arguments.  So this pattern won't be used in any existing working code.  With this PR it will now work and create the default hardware I2C instance.  I think that's an acceptable change that shouldn't break existing code.
- `machine.I2C(-1, ...)` will still work the same as before, allow creating SoftI2C instances.

dpgeorge: Great, thanks!

ricksorensen:
My experience with a XIAO C3 running with the GENERIC_C3 build:
`MicroPython v1.25.0-preview.531.gef8282c71 on 2025-04-09; XIAOC3 with ESP32C3`
is that both `i2=I2C()` and `i2=I2C(0)` hang the REPL with no return requiring a hardware reset.  

Using 1.24.1 release build, `i2=I2C(0)` hangs the REPL as above, but `i2=I2C()` fails with 
```>>> i2=I2C()
Warning: I2C(-1, ...) is deprecated, use SoftI2C(...) instead
TypeError: 'scl' argument required
>>> i2=I2C(0)
``` 
I will look at what is causing the hang.  Note that on the XIAO S3 both `i2=I2C()` and `i2=I2C(0)`  work.  Maybe because there are 2 HW I2C channels available?

PR #16672, opened 2025-01-30
py/parsenum.c: reduce code footprint of mp_parse_num_float.

yoctopuce:
### Summary

The existing mantissa parsing code uses a floating point variable to accumulate digits. 

Using an mp_float_uint_t variable instead, and casting to mp_float_t at the very end is more efficient and reduces code size. In some cases, it also improves the rounding behaviour as extra digits are taken into account by the int-to-float conversion code.

### Testing

The new code was unit-tested using a small program generating 100'000 random mantissa and exponents, parsing the resulting number as a float and comparing the result to the value obtained via bigint arithmetic.
- when using `MICROPY_FLOAT_IMPL_DOUBLE`, with exponents between -200 and 200
  - the max relative error of the old code is around 5.7e-16
  - the max relative error of the new code is around 4.3e-16
  - CPython max relative error is around 3.6e-16
- when using `MICROPY_FLOAT_IMPL_FLOAT`, with exponents between -20 and 50
  - the max relative error of the old code is around 4.3e-7
  - the max relative error of the new code is around 2.4e-7

The code size reduction was observed on binaries built using gcc-arm-none-eabi-9-2019-q4-major

A further improvement in precision for positive exponents could be obtained by leveraging bigint arithmetic up to the computation of the final value:
```
        #if MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_FLOAT
        mp_obj_t mant_obj = mp_obj_new_int_from_uint(mantissa);
        #else
        mp_obj_t mant_obj = mp_obj_new_int_from_ulonglong(mantissa);
        #endif
        mp_binary_op_t combine_op = MP_BINARY_OP_MULTIPLY;
        if (exp_val < 0) {
            combine_op = MP_BINARY_OP_TRUE_DIVIDE;
            exp_val = -exp_val;
        }
        mp_obj_t exp_obj = MP_OBJ_NEW_SMALL_INT(exp_val);
        mp_obj_t pow_obj = mp_binary_op(MP_BINARY_OP_POWER, MP_OBJ_NEW_SMALL_INT(10), exp_obj);
        mp_obj_t res_obj = mp_binary_op(combine_op, mant_obj, pow_obj);
        dec_val = mp_obj_get_float(res_obj);
```
This would however increase the code size, and cause dynamic object allocation in case the mantissa is larger then 9 digits or the absolute value of the exponent is bigger than 9, so this idea was dropped.

### DIFF
--- a/py/parsenum.c
+++ b/py/parsenum.c
@@ -179,39 +179,40 @@ typedef enum {
 } parse_dec_in_t;
 
 #if MICROPY_PY_BUILTINS_FLOAT
-// DEC_VAL_MAX only needs to be rough and is used to retain precision while not overflowing
+// MANTISSA_MAX is used to retain precision while not overflowing mantissa
 // SMALL_NORMAL_VAL is the smallest power of 10 that is still a normal float
 // EXACT_POWER_OF_10 is the largest value of x so that 10^x can be stored exactly in a float
 //   Note: EXACT_POWER_OF_10 is at least floor(log_5(2^mantissa_length)). Indeed, 10^n = 2^n * 5^n
 //   so we only have to store the 5^n part in the mantissa (the 2^n part will go into the float's
 //   exponent).
 #if MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_FLOAT
-#define DEC_VAL_MAX 1e20F
+#define MANTISSA_MAX 0x19999998U
 #define SMALL_NORMAL_VAL (1e-37F)
 #define SMALL_NORMAL_EXP (-37)
 #define EXACT_POWER_OF_10 (9)
 #elif MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_DOUBLE
-#define DEC_VAL_MAX 1e200
+#define MANTISSA_MAX 0x1999999999999998ULL
 #define SMALL_NORMAL_VAL (1e-307)
 #define SMALL_NORMAL_EXP (-307)
 #define EXACT_POWER_OF_10 (22)
 #endif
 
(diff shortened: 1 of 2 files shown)


dpgeorge (approved):
Thanks for this, it's a very neat improvement.

The stm32 board PYBD_SF6 uses double precision floats, and it's always failed `tests/float/float_parse.py` and `tests/float/float_parse_doubleprec.py` due to minor parsing errors, ie:
```diff
FAILURE tests/results/float_float_parse.py
--- tests/results/float_float_parse.py.exp
+++ tests/results/float_float_parse.py.out
@@ -2,9 +2,9 @@
 0.0
 1.0000e+00
 1.0000e+20
-1.0
-1e+20
-True
+1.000000000000001
+1.000000000000001e+20
+False
 True
 True
 True
```
and
```diff
FAILURE tests/results/float_float_parse_doubleprec.py
--- tests/results/float_float_parse_doubleprec.py.exp
+++ tests/results/float_float_parse_doubleprec.py.out
@@ -1,9 +1,9 @@
-1e+300
-1e+200
-1.0
-1.0
-1e+100
-1e-100
+1.000000000000003e+300
+1.000000000000003e+200
+1.000000000000003
+1.000000000000003
+1.000000000000003e+100
+1.000000000000003e-100
 9.00000000000000e-301
 9.00000000000000e-201
 9.00000000000000e-01
```
All the above differences are resolved by this PR, very good!

I see that code size is increased for boards like TEENSY40 which use double precision.  I assume this is because they now need to pull in 64-bit integer multiplication support.  I guess that's necessary.  If you want double precision floats then you want precision, and also probably have enough flash to fit the 64-bit integer arithmetic.

Issue #16424, opened 2024-12-16
USER_C_MODULES for ESP32


straga:
### Port, board and/or hardware

esp32

### MicroPython version

Master

### Reproduction

For building the example modules which come with MicroPython, set USER_C_MODULES to the examples/usercmodule directory for Make, or to examples/usercmodule/micropython.cmake for CMake.

The recent commit b65e89107c0194bfdaf92720786341a2047f36b6 enforces that USER_C_MODULES must point to a directory in the Make system. However, this causes confusion with the documentation, which mixes usage of relative paths and file references for both Make and CMake.

but that's half the battle. We can comment on this committee. 
Also, they no longer work as they used to.

#### Examples:

- The following command no longer works with Make after the commit:
```
cd micropython/ports/esp32
make USER_C_MODULES=../../examples/usercmodule/micropython.cmake
```

- Instead, absolute paths are required:
```
make USER_C_MODULES=/dev_iot/opt/upy/micropython/examples/usercmodule/micropython.cmake BOARD=ESP32_GENERIC all
```

dpgeorge:
> * The following command no longer works with Make after the commit:

This command should work, it's used in the CI to test building user C modules with the esp32 port.  Maybe you just need an extra `../`:
```
cd ports/esp32
make USER_C_MODULES=../../../examples/usercmodule/micropython.cmake
```

straga:
@dpgeorge 

```
[1045/1396] Generating /Users/dev/Documents/dev_iot/opt/upy/micropython/mpy-cross/build/mpy-cross
FAILED: /Users/dev/Documents/dev_iot/opt/upy/micropython/mpy-cross/build/mpy-cross
cd /Users/dev/Documents/dev_iot/opt/upy/micropython/ports/esp32/build-ESP32_GENERIC/esp-idf/main_esp32 && make -C /Users/dev/Documents/dev_iot/opt/upy/micropython/mpy-cross
Use make V=1 or set BUILD_VERBOSE in your environment to increase build verbosity.
../py/py.mk:38: *** Directory containing USER_C_MODULES doesn't exist: ../../../examples/usercmodule/.  Stop.
```

![CleanShot 2024-12-16 at 16 50 33@2x](https://github.com/user-attachments/assets/a7a9b762-f390-4908-9c55-e282955eaf74)

dpgeorge:
Ah, OK, so the issue is building mpy-cross with user C modules, not the esp32 port.

[...]

PR #16021, opened 2024-10-16
py/py.mk: Add check that any specified USER_C_MODULES folder exists.

andrewleech:
### Summary

When compiling with `USER_C_MODULES` set on the command line it's easy to get the path slightly wrong, particularly as it's usually given as a relative path to the port / build root directory.

Currently, if the provided path is incorrect the build proceeds as usual simply without including any of the user c modules which causes confusion for the developer.

This PR adds a basic check in the make rules to ensure that the provided path does indeed exist. 

After also getting confused by the different way this is specifed for make vs cmake I updated cmake script to also accept the folder (same as make) and simply append the `micropython.cmake` file to that if needed. This still works as it did previously if the full path to that file is given.

#### With invalid paths:
```
anl@STEP:~/micropython$ make -C ports/unix/ USER_C_MODULES=../../../examples/usercmodule
make: Entering directory '/home/anl/micropython/ports/unix'
Use make V=1 or set BUILD_VERBOSE in your environment to increase build verbosity.
../../py/py.mk:36: *** USER_C_MODULES doesn't exist: /home/anl/examples/usercmodule.  Stop.
make: Leaving directory '/home/anl/micropython/ports/unix'```
```

```
anl@STEP:~/micropython$ make -C ports/rp2/ USER_C_MODULES=../../../examples/usercmodule/micropython.cmake 
make: Entering directory '/home/anl/micropython/ports/rp2'
[ -e build-RPI_PICO/Makefile ] || cmake -S . -B build-RPI_PICO -DPICO_BUILD_DOCS=0 -DMICROPY_BOARD=RPI_PICO -DMICROPY_BOARD_DIR="/home/anl/micropython/ports/rp2/boards/RPI_PICO" -DUSER_C_MODULES=../../../examples/usercmodule/micropython.cmake
PICO_SDK_PATH is /home/anl/micropython/lib/pico-sdk<snip>
<snip>
CMake Error at /home/anl/micropython/py/usermod.cmake:47 (message):
  USER_C_MODULES doesn't exist:
  /home/anl/examples/usercmodule/micropython.cmake
Call Stack (most recent call first):
  CMakeLists.txt:92 (include)

-- Configuring incomplete, errors occurred!
```

#### With corrected path:
```
anl@STEP:~/micropython$ make -C ports/unix/ USER_C_MODULES=../../examples/usercmodule
make: Entering directory '/home/anl/micropython/ports/unix'
Use make V=1 or set BUILD_VERBOSE in your environment to increase build verbosity.
[...]

PR #16688, opened 2025-02-03
py/py.mk: Ignore USER_C_MODULES check when building mpy-cross.

andrewleech:
### Summary

In https://github.com/micropython/micropython/pull/16021 I introduced a new makefile check when setting `USER_C_MODULES` to ensure the provided path exists.

This PR skips the check logic when building mpy-cross as it does not use this variable.

As issue with mpy-cross build failing due to the `USER_C_MODULES` check was first raised by @Gadgetoid in https://github.com/micropython/micropython/pull/16021#issuecomment-2498127602.

I (somewhat unfairly) dismissed this issue at the time, flagging that this variable shouldn't really be set for a mpy-cross build that doesn't use it.

This has come back to bite me however when working on an esp32 project which uses `USER_C_MODULES`.
I didn't realise previously that the esp32 cmake process automatically compiles mpy-cross to ensure that particular xtensa settings are included correctly.
When compiled this way, the variables (like `USER_C_MODULES`) from the target build environment automatically flow down to the `mpy-cross` dependency build. 

When the `USER_C_MODULES` path is provided as a relative path this will be incorrect / broken in the mpy-cross directory so the error is raised.

To avoid this problem entirely this PR skips the check when mpy-cross is being built.

### Testing

I have been building an esp32-s2 project using `USER_C_MODULES` which was failing due to the check raising an error during the automatic dependency build of mpy-cross.

The change from this PR fixed the issue.

An incorrect path still is detected by the check during the target build.

Alternatively the mpy-cross Makefile could explicitely test and clear this variable before including `py.mk` however I thought it might be cleaner to keep this related logic located in the one place. Happy to change it if requested however!

### DIFF
--- a/py/mkrules.cmake
+++ b/py/mkrules.cmake
@@ -211,7 +211,7 @@ if(MICROPY_FROZEN_MANIFEST)
         endif()
         add_custom_command(
             OUTPUT ${MICROPY_MPYCROSS_DEPENDENCY}
-            COMMAND ${MICROPY_MAKE_EXECUTABLE} -C ${MICROPY_DIR}/mpy-cross
+            COMMAND ${MICROPY_MAKE_EXECUTABLE} -C ${MICROPY_DIR}/mpy-cross USER_C_MODULES=
         )
     endif()
 
--- a/py/mkrules.mk
+++ b/py/mkrules.mk
@@ -176,7 +176,7 @@ $(HEADER_BUILD):
 ifneq ($(MICROPY_MPYCROSS_DEPENDENCY),)
 # to automatically build mpy-cross, if needed
 $(MICROPY_MPYCROSS_DEPENDENCY):
-	$(MAKE) -C "$(abspath $(dir $@)..)"
+	$(MAKE) -C "$(abspath $(dir $@)..)" USER_C_MODULES=
 endif
 
 ifneq ($(FROZEN_DIR),)


Gadgetoid:
> This has come back to bite me however when working on an esp32 project which uses `USER_C_MODULES`. I didn't realise previously that the esp32 cmake process automatically compiles mpy-cross

Ooof!

This feels like the kind of thing CI should have caught, but I am acutely aware it's very easy to say that and entirely another problem to actually cover `USER_C_MODULES` comprehensively.

andrewleech:
Yeah it might be an absolute vs relative path to the folder issue too. That's the problem with adding flexibility to support both; it doubles the test space.

dpgeorge:
Thanks, this should fix #16424.

dpgeorge:
How about this as an alternative:
```diff
diff --git a/py/mkrules.cmake b/py/mkrules.cmake
--- a/py/mkrules.cmake
+++ b/py/mkrules.cmake
@@ -211,7 +211,7 @@ if(MICROPY_FROZEN_MANIFEST)
         endif()
         add_custom_command(
             OUTPUT ${MICROPY_MPYCROSS_DEPENDENCY}
-            COMMAND ${MICROPY_MAKE_EXECUTABLE} -C ${MICROPY_DIR}/mpy-cross
+            COMMAND ${MICROPY_MAKE_EXECUTABLE} -C ${MICROPY_DIR}/mpy-cross USER_C_MODULES=
         )
     endif()
 
diff --git a/py/mkrules.mk b/py/mkrules.mk
--- a/py/mkrules.mk
+++ b/py/mkrules.mk
@@ -176,7 +176,7 @@ $(HEADER_BUILD):
 ifneq ($(MICROPY_MPYCROSS_DEPENDENCY),)
 # to automatically build mpy-cross, if needed
 $(MICROPY_MPYCROSS_DEPENDENCY):
-       $(MAKE) -C "$(abspath $(dir $@)..)"
+       $(MAKE) -C "$(abspath $(dir $@)..)" USER_C_MODULES=
 endif
 
 ifneq ($(FROZEN_DIR),)
```
I tested that on esp32, rp2 and stm32 port (using `USER_C_MODULES=../../examples/usercmodule/cexample` as a relative path, which I think is necessary to see the issue) and it works with that patch.

It's a bit clearer then what's going on (it's the recursive make that's the issue), and doesn't accidentally build mpy-cross with user C modules enabled (which the current PR could still do).

Gadgetoid:
@dpgeorge's hack above is exactly what I did to fix our build systems when this issue popped up.

andrewleech:
Thanks @dpgeorge I've updated this PR with your suggestions.

dpgeorge (approved):

dpgeorge:

PR #16690, opened 2025-02-03
py/emitnative: Further (minor) improvements.

agatti:
### Summary

This PR contains two minor improvements to the native emitter that weren't part of my previous PR even though they should have been.

There are two commits, the first one prevents Xtensa Viper word load/store operations from performing the same operation twice - if the optimised code path was taken the function didn't terminate execution but went on to perform the regular unoptimised operation (a `break` statement was missing).  The second one explicitly marks the condition code lookup tables as `const`, which should help out some compilers a bit.

### Testing

Viper code generation was tested on an ESP8266 via `./run-tests.py -t <device> -d micropython -i "viper.*"`, and `mpy-cross` built successfully without warnings.


dpgeorge (approved):

dpgeorge:
Thanks!

PR #16691, opened 2025-02-04
docs: Add strings vs bytes to speed optimisation tips.

### LINKED ISSUE #8580: String versus bytes performance
Hello, with the example below on esp32 (V1.18 micropython), the string is 89 times slower than the bytes, unfortunately not only on this simple case, on another esp32 I can reach a factor of 280, it depends on the hardware, a pico pi is 20 times. An optimization would save time for many applications.

**Result :**
- concatenation strings 2328 ms
- concatenation bytes 26 ms

**Source :**

```py
from time import ticks_ms                                                                                    
start = ticks_ms()

for i in range(1000):
    a = "b" + "b"

end = ticks_ms()
print("concatenation strings %d ms"%(end-start))

start = ticks_ms()

for i in range(1000):
    a = b"b" + b"b"

end = ticks_ms()
print("concatenation bytes %d ms"%(end-start))
```

projectgus:
### Summary

Closes #8580. The speed difference has substantially reduced since that issue was opened, but it's still enough to make a difference in some cases and not particularly obvious.

Also add some additional context links, suggestions for alternative classes, etc.

### DIFF
--- a/docs/library/array.rst
+++ b/docs/library/array.rst
@@ -19,6 +19,10 @@ Classes
     array are given by *iterable*. If it is not provided, an empty
     array is created.
 
+    In addition to the methods below, array objects also implement the buffer
+    protocol. This means the contents of the entire array can be accessed as raw
+    bytes via a `memoryview` or other interfaces which use this protocol.
+
     .. method:: append(val)
 
         Append new element *val* to the end of array, growing it.
--- a/docs/reference/speed_python.rst
+++ b/docs/reference/speed_python.rst
@@ -57,6 +57,8 @@ and used in various methods.
 
 This is covered in further detail :ref:`Controlling garbage collection <controlling_gc>` below.
 
+.. _speed_buffers:
+
 Buffers
 ~~~~~~~
 
@@ -69,6 +71,13 @@ example, objects which support stream interface (e.g., file or UART) provide ``r
 method which allocates new buffer for read data, but also a ``readinto()`` method
 to read data into an existing buffer.
 
+Some useful classes for creating reusable buffer objects:
+
+- :class:`bytearray`
+- :mod:`array` (:ref:`discussed below<speed_arrays>`)
+- :class:`io.StringIO` and :class:`io.BytesIO`
+- :class:`micropython.RingIO`
+
 Floating point
 ~~~~~~~~~~~~~~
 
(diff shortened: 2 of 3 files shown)


docs/reference/speed_python.rst
```diff
@@ -118,6 +131,23 @@ of buffer and fills in entire buffer. What if you need to put data in the
...
+Strings vs Bytes
+~~~~~~~~~~~~~~~~
+
+MicroPython uses :ref:`string interning <qstr>` to save space when there are
+multiple identical strings. Each time a new string is allocated at runtime (for
+example, when two other strings are concatenated), MicroPython checks whether
+the new string can be interned to save RAM.
+
+If you have code which performs performance-critical string operations then
+consider using :class:`bytes` objects or byte string literals. This skips the
```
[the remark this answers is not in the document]
projectgus: Not only you but also Python - https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

Will fix

docs/reference/speed_python.rst
```diff
@@ -69,6 +71,12 @@ example, objects which support stream interface (e.g., file or UART) provide ``r
 method which allocates new buffer for read data, but also a ``readinto()`` method
 to read data into an existing buffer.
 
+Some useful classes for creating reusable buffer objects:
+
+- :class:`bytearray`
+- :mod:`array` (:ref:`discussed below<speed_arrays>`)
+- :class:`io.StringIO` and :class:`io.BytesIO`
```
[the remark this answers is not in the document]
projectgus: Memoryview is documented a bit further down, which I think is appropriate as it's more of a technique for working with buffers. 

Good call on RingIO, I should have thought of this!

docs/reference/speed_python.rst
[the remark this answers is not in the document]
projectgus: Have edited to something slightly different (also put a bytes literal example, for clarity)

projectgus:
@dpgeorge Thanks for the review! Have addressed, and also added a note about MicroPython `array` objects implementing the buffer protocol.

docs/library/array.rst
```diff
@@ -79,3 +79,8 @@ Classes
 
         **Note:** ``__repr__`` cannot be called directly (``a.__repr__()`` fails) and
         is not present in ``__dict__``, however ``str(a)`` and ``repr(a)`` both work.
+
+    Array objects also implement the buffer protocol, meaning the contents can
+    be accessed as raw bytes via a `memoryview` or other interfaces which use
+    this protocol.
```
dpgeorge, line 85: I reckon this is probably important enough that it could go up as the second paragraph of the `array.array` constructor, line 21.

projectgus: Good idea, have moved.

dpgeorge (approved):

docs/library/array.rst
```diff
@@ -19,6 +19,10 @@ Classes
     array are given by *iterable*. If it is not provided, an empty
     array is created.
 
+    Array objects also implement the buffer protocol, meaning the contents can
```
dpgeorge, line 22: Reading this again, the word "also" is out of place here.  I suggest "In addition to the methods below, array objects also ...".

projectgus: I've made an edit along these lines. I think this is why I originally decided to put this note at the end, because the array is primarily an array of typed objects and only secondarily a flat buffer. The version in d3932ad3f17c9a000d867f6ecb1d4bdb7c6925ca may be less confusing for someone who is new to the `array` module. However I don't feel very strongly either way.

dpgeorge (approved):

PR #16700, opened 2025-02-05
esp32: Fix invalid result of machine.SDCard readblocks().

projectgus:
### Summary

Small fix for invalid SD readblocks result - function would return NULL instead of mp_const_false if failed to init.

### Testing

Manually tested, calling `readblocks` on an invalid SD device would cause an invalid result to Python. Using with Vfs could cause a crash.


dpgeorge:
Related: #16223

dpgeorge (approved):

PR #16703, opened 2025-02-05
stm32: Fix hang when negative block_num passed to sdcard.

### LINKED ISSUE #10622: SDCard `readblocks()` method crashes pyboard when trying read block -1
On a Pyboard D series I've discovered the following bug:

```
>>> import pyb
>>> bd=pyb.SDCard()
>>> bd.power(True)
True
>>> bsize=bd.ioctl(5,None)
>>> nblocks=bd.ioctl(4, None)
>>> buf=bytearray(bsize)
>>> bd.readblocks(0, buf)
True
>>> bd.readblocks(1, buf)
True
>>> bd.readblocks(nblocks-1, buf)
True
>>> bd.readblocks(nblocks, buf)
False
>>> bd.readblocks(-1, buf)
[Pyboard fails to respond until watchdog or manual reset]
```

This is on Micropython 1.18.

`bd.writeblocks(-1, buf)` works fine (i.e. returns `False` since it failed the operation)

projectgus:
### Summary

* The ST HAL does a [bounds check on SD read operations](https://github.com/micropython/stm32lib/blob/vendor/STM32F4xx_HAL_Driver/Src/stm32f4xx_hal_sd.c#L1092), but there's an unsigned integer overflow bug if `(BlockAdd + NumberOfBlocks)` overflows. It seems like some SD cards then reject the out-of-bounds block number, but some become unresponsive. Closes #10622.
* The operation actually will time out (although the card won't respond to subsequent commands.) The timeout was set to 60 seconds and for some reason (unclear) my Linux system gave up on the USB serial connection before reaching this timeout. Pushed a commit to lower this to 30 seconds.

### Testing

Was able to reproduce by following the steps from the linked issue on PyBoard V1.1. However it seems like only some SD cards will reproduce, some will fail immediately as expected.

### DIFF
--- a/ports/stm32/sdcard.c
+++ b/ports/stm32/sdcard.c
@@ -498,13 +498,27 @@ static HAL_StatusTypeDef sdcard_wait_finished(uint32_t timeout) {
     return HAL_OK;
 }
 
-mp_uint_t sdcard_read_blocks(uint8_t *dest, uint32_t block_num, uint32_t num_blocks) {
+static HAL_StatusTypeDef sdcard_common_checks(uint32_t block_num, uint32_t num_blocks) {
     // check that SD card is initialised
     if (!(pyb_sdmmc_flags & PYB_SDMMC_FLAG_ACTIVE)) {
         return HAL_ERROR;
     }
 
-    HAL_StatusTypeDef err = HAL_OK;
+    // check that adding block_num & num_blocks don't overflow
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
This looks good.

I can't really test that the bug is fixed because I could not reproduce the error with my SD cards.

But, testing on PYBV10, SD card still works with this PR.

PR #16706, opened 2025-02-05
Stm32g4 timer20 interrupt

H-Grobben:
### Summary

Attaching an callback to timer 20 on a STM32 G4 will cause the repl to lock up until reset. After fixing the callback issue, a base frequency issue showed up. Both should be solved with this PR.

### Testing
board: NUCLEO_G474RE
``` py
def timer20_cb(_):
   print('Hello!')

import pyb
timer20 = pyb.Timer(20)
timer20.init(freq=1,callback=timer20_cb)
```
Before: lockup.
After: Hello!...

### DIFF
--- a/ports/stm32/stm32_it.c
+++ b/ports/stm32/stm32_it.c
@@ -775,6 +775,14 @@ void TIM17_FDCAN_IT1_IRQHandler(void) {
 }
 #endif
 
+#if defined(STM32G4)
+void TIM20_UP_IRQHandler(void) {
+    IRQ_ENTER(TIM20_UP_IRQn);
+    timer_irq_handler(20);
+    IRQ_EXIT(TIM20_UP_IRQn);
+}
+#endif
+
 #if defined(STM32H7)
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks, this fix looks correct.

I also see that TIM15, TIM16 and TIM17 are on APB2, so would similarly need fixing in the clock source function.  But that might be out of scope for this PR.

dpgeorge:
See how you go with the additional timers.  If it's too much out of scope it can be left for a separate PR.

dpgeorge:
> update: seems more STMs use timer 15 - 17, and not all have them on APB2 clock. Needs some more investigating.

OK, then let's merge this PR now, as is.  Follow up work can be done later.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 3b948893d827c39c967b0ca17da3e8159eaef176 and 91386b3d56b4ef3bb27af89846c1cc2712f6065f

PR #16710, opened 2025-02-06
extmod/vfs_rom: Remove ability to create VfsRom from an address.

dpgeorge:
### Summary

`VfsRom` objects can currently be constructed from an address (a pointer to memory containing a ROMFS).  It's not necessary to support this mode of construction, which allows an arbitrary memory address to be specified and potentially allows invalid memory accesses.

Requiring an object with the buffer protocol is safer, and also means that the length of the region is always specified.  It also doesn't lose any generality because you can always create a buffer from an arbitrary memory region using `uctypes.bytearray_at`.

### Testing

Corresponding test updated, will run under CI.

### DIFF
--- a/extmod/vfs_rom.c
+++ b/extmod/vfs_rom.c
@@ -198,15 +198,13 @@ static mp_obj_t vfs_rom_make_new(const mp_obj_type_t *type, size_t n_args, size_
     self->base.type = type;
     self->memory = args[0];
 
+    // Get the ROMFS memory region.
     mp_buffer_info_t bufinfo;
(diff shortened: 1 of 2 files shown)


dpgeorge:
This is split out from #8381.

robert-hh:
Why this splitout?

dpgeorge:
> Why this splitout?

I just don't think we can get #8381 merged before the 1.25.0 release.  So wanted to get at least this minor change in.

projectgus (approved):

Issue #16499, opened 2024-12-29
ESPNow.config fails to get param "rate" (unknown config param) 


martin-korinek-1985:
### Port, board and/or hardware

ESP32-C3 Super Mini

### MicroPython version

uname: (sysname='esp32', nodename='esp32', release='1.24.1', version='v1.24.1 on 2024-11-29', machine='ESP32C3 module with ESP32C3')

### Reproduction

```python
import os
import espnow
import network

sta = network.WLAN(network.STA_IF)
sta.active(True)
sta.disconnect()

e = espnow.ESPNow()
e.active(True)

print('uname     :', os.uname())
print('rxbuf     :', e.config('rxbuf'))
print('timeout_ms:', e.config('timeout_ms'))
print('rate      :', e.config('rate'))
```

### Expected behaviour

Expectation is NOT to get error "ValueError: unknown config param" but get value of the "rate" instead.

### Observed behaviour
```
Traceback (most recent call last):
  File "<stdin>", line 15, in <module>
ValueError: unknown config param
```
![image](https://github.com/user-attachments/assets/2952f501-3de3-4b1c-93e4-5a5110b8b2e7)

### Additional Information

![image](https://github.com/user-attachments/assets/b1d8e914-4cfa-46b0-aa86-358f8e986d2f)

### Code of Conduct

Yes, I agree

agatti:
As far as I know the ESP32 SDK doesn't provide any explicit function to get the current ESPNow wifi rate.  It only allows for it to be set (see `esp_wifi_config_espnow_rate` and `esp_now_set_peer_rate_config` in `components/esp_wifi/include/esp_now.h` in ESP-IDF 5.2 for example).

martin-korinek-1985:
I do appreciate all the work that is done on MicroPython and I very much like to use it. My intention is not to spend hours trying to do the same in low level language like C or ASM so I believe you there is no ESP-IDF API for what I was trying to do. Unfortunately I did not realize that based on the documentation:  
![image](https://github.com/user-attachments/assets/ee74fc39-89ed-454b-809c-38165d7942a4)  
and still have a few questions.
1) I think it would be great to update the documentation with note, that ESPNow.config(rate=XY) is only working on ESP32 && IDF>=4.3.0 && only the "set" part, not the "get" part. So the rate can be only set but the information is not possible to retrieve (at least not in this way).
[...]

PR #16730, opened 2025-02-10
docs/library/espnow.rst: Clarify usage of the "rate" configuration key.

agatti:
### Summary

This PR adds a clarification for the ESPNow module's documentation regarding its "config" method.

The original documentation for that method could be interpreted as having all its configuration keys being able to be queried, but the "rate" configuration key is actually write-only due to ESP-IDF's lack of a way to retrieve that bit of information from the radio's configuration.  The documentation changes highlight the fact that said configuration key is actually write-only.

The reference to the rate configuration key being available from ESP-IDF 4.4.0 and onwards was removed as now all supported versions of ESP-IDF have support for setting the radio interface rate.  The link to the `wifi_phy_rate_t` enum was also updated to point to the latest version of the documentation for the v5.2 series.

This change was meant to prevent issues like #16499, where the documentation not being clear enough caused some confusion on the user's end w.r.t. LR mode setting.

### Testing

The discussion for #16499 should present the problem in question and it being solved with the bit of information that's now part of the PR.

As an aside, maybe the new PR template should be updated to let people know the "Testing" section may be optional for docs-only changes?

### DIFF
--- a/docs/library/espnow.rst
+++ b/docs/library/espnow.rst
@@ -164,11 +164,13 @@ Configuration
         wait forever. The timeout can also be provided as arg to
         `recv()`/`irecv()`/`recvinto()`.
 
-        *rate*: (ESP32 only, IDF>=4.3.0 only) Set the transmission speed for
+        *rate*: (ESP32 only) Set the transmission speed for
         ESPNow packets. Must be set to a number from the allowed numeric values
         in `enum wifi_phy_rate_t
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Looks good, thank you!

Re testing for docs: you can actually test the docs by building them locally and checking your changes look correct, eg correct markup, links work :smile:

Issue #16594, opened 2025-01-16
Inline assembler/code.


0wwafa:
### Description

It would be great to have an inline assembler or a way to pass machine code (already precompiled).
Python is great but quite slow for many timing sensitive applications.
It would be also useful to add features on the fly.

I see two possible implementations:
The classic one  asm("nop; nop;")
Or a byte array containing the code.

### Code Size

Shouldn't be much.

### Implementation

I hope the MicroPython maintainers or community will implement this feature

### Code of Conduct

Yes, I agree

agatti:
In theory the ESP8266 inline assembler code could be repurposed for this (I believe the opcodes are almost the same between the ESP8266 and ESP32), but the way function calls are handled is quite different [1].

I'm not an expert on Xtensa matters, so I can't say whether supporting the windowed ABI is easy or not, maybe somebody with direct Xtensa experience can chime in with more detailed information.  This looks interesting though, maybe I can give it a try once I get some free time for non-RISCV work.

<hr>

[1] See also https://sachin0x18.github.io/posts/demystifying-xtensa-isa/#calling-convention

rkompass:
> This looks interesting though, maybe I can give it a try once I get some free time for non-RISCV work.

@agatti I would love it if you found the time to start work on this.

I studied the ESP8266 inline assembler for using it in my CRC code. There the benchmarks showed that the Xtensa architecture is on par with ARM wrt. performance.
But even the ESP8266 inline assembler is not complete yet. Some important instructions like bit shifts had to be done via data() statements.
From this the [preliminary asm_xtensa docs](https://github.com/orgs/micropython/discussions/12965) resulted.

I was wondering what prevented the completion of the inline assembler for the ESP architecture. I guess it was the necessity to prioritize other work at the time. Otherwise at least the 8266 assembler would be completed.

> I can't say whether supporting the windowed ABI is easy or not

The ESP32 viper emitter emits code according to this ABI. So from my limited understanding this problem seems to be solved already. It appears as if the completion of the inline assembler was _stepped over_.

Now as you completed the inline assembler for RISCV the Xtensa architecture is the only important one with missing inline assembler.
I will incorporate the inline assembler for RV-32 as optimization option for my CRC code soon.

I would love to help in development, testing and documentation on this. But didn't start to work on it myself, because I had the feeling that it is above my level of expertice.

agatti:
> The ESP32 viper emitter emits code according to this ABI.

[...]

PR #16731, opened 2025-02-10
py/emitinlinextensa: Add all supported LX3/LX106 opcodes to the inline assembler.

agatti:
### Summary

This PR expands the Xtensa inline assembler to support most if not all opcodes available on the ESP8266 and LX3/LX106 Xtensa cores.

This is meant as a stepping stone to add inline assembler support for the ESP32 and its LX6 core, along to windowed-specific opcodes and additional opcodes that are present only on the LX7 core (ESP32-S3 and later).  Rather than having a single monolithic PR with all sorts of assembler changes I chose to split this into ESP8266-only support first and then use this to add ESP32 support (see #16594).

New opcodes being added are covered by tests, and the provided tests were expanded to also include opcodes available in the existing implementation.  Given that the ESP8266 space requirements are tighter than ESP32's, certain opcodes that won't be commonly used are put behind a define (`MICROPY_EMIT_INLINE_XTENSA_UNCOMMON_OPCODES`, disabled by default) to save some space in the general use case.

### Testing

The new inline assembler additions were tested on an ESP8266 with `./run-tests.py -t <device> -d inlineasm/xtensa`.

This increases the footprint of the ESP8266 firmware, as it has inline assembler enabled by default.  I've tried to keep the size increase within reason but there's only so much you can do after all.

Also, some extmod tests would be emitted incorrectly when ran through `mpy-cross` to generate Xtensa/Xtensawin native code (some jumps are too far away to be encoded with a simple opcode) but now they fail at compile time rather than at runtime.  This will be fixed in a later PR.


rkompass:
Hello @agatti,
I started to try out this PR.

You noted in the tests:

> `# SLLI shift argument is actually (32 - shift)`

I suggest the following correction for this in emitinlinextensa.c (line 472ff):
```
        } else if (op == MP_QSTR_slli) {
            mp_uint_t r0 = get_arg_reg(emit, op_str, pn_args[0]);
            mp_uint_t r1 = get_arg_reg(emit, op_str, pn_args[1]);
            mp_uint_t bits = 32 - get_arg_i(emit, op_str, pn_args[2], 1, 31);  // correct role of i
            asm_xtensa_op24(&emit->as, ASM_XTENSA_ENCODE_RRR(0, 1, 0 | ((bits >> 4) & 0x01), r0, r1, bits & 0x0F));
```

agatti:
Right, turns out it's also what GAS does when assembling the SLLI opcode.  Thanks for pointing it out, I'll fix it right away.

rkompass:
Thanks. Another one which is purely cosmetic:
```
@micropython.asm_xtensa
def lsr31(a2):
    srli(a2, a2, 15)

print(hex(lsr31(0x80000000)))
```
in the tests should be titled lsr15, because you do the shift by 15 bits, right?

agatti:
Yep. SRLI only accepts from 1 to 15 as its shift parameter - good catch.  I'll fix this after the current test run passes.  Thanks!

rkompass:
I had in my [Preliminary asm_xtensa docs](https://github.com/orgs/micropython/discussions/12965#top)
noted that:
```
#   slli(Ax, Ay, imm4)  AX = Ay << imm4  Shift left imm.     data(3, 0x110000 | x<<12 | y<<8 | (16-imm4)<<4)
#      slli(Ax, Ay, imm5) with imm5 > 16 may be coded as     data(3, 0x010000 | x<<12 | y<<8 | (32-imm5)<<4)
#   srli(Ax, Ay, imm4)  AX = Ay >> imm4  L. sh. right imm.   data(3, 0x410000 | x<<12 | y<<4 | imm4<<8)
#   srai(Ax, Ay, imm4)  AX = Ay >> imm4  A. sh. right imm.   data(3, 0x210000 | x<<12 | y<<4 | imm4<<8)
#      srai(Ax, Ay, imm5) with imm5 > 15 may be coded as     data(3, 0x310000 | x<<12 | y<<4 | (imm5-8)<<8)
```
which includes that srli is only shiftable by 0..15 bits (i.e. imm4).
But the other two have a more complicated coding where the case i<16 has to be handled differently than i>=16.
This is not reflected in the PR yet.

Consequently now the first test in xtensa_asmshift.py fails and produces `-0x80000000` instead of `0x246`.

agatti:
I've implemented SRAI according to https://www.cadence.com/content/dam/cadence-www/global/en_US/documents/tools/silicon-solutions/compute-ip/isa-summary.pdf pages 613 and 614:

```c
        } else if (op == MP_QSTR_srai) {
            mp_uint_t r0 = get_arg_reg(emit, op_str, pn_args[0]);
            mp_uint_t r1 = get_arg_reg(emit, op_str, pn_args[1]);
            mp_uint_t bits = get_arg_i(emit, op_str, pn_args[2], 0, 31);
            asm_xtensa_op24(&emit->as, ASM_XTENSA_ENCODE_RRR(0, 1, 2 | ((bits
 >> 4) & 0x01), r0, bits & 0x0F, r1));
```

The case for shifts larger than 15 is handled, the third argument of ASM_XTENSA_ENCODE_RRR (`2 | ((bits >> 4) & 0x01)`) covers bits 20 to 24.  How should that work instead?

rkompass:
I added the snippet:

```
@micropython.asm_xtensa
def asr24(a2):
    srai(a2, a2, 24)

print(hex(asr1(0x12345678)), hex(0x12345678>>24))
```
 to the test file `xtensa_asmshift.py` and it also reports an error with the current code:
`0x91a2b3c`  is returned instead of `0x12`.

So the results seem to show that there is something wrong yet. Can you confirm that?
I got my info from [Xtensa® Instruction Set Architecture (ISA) Reference Manual ](https://0x04.net/~mwk/doc/xtensa.pdf)
pages 525 (slli) and 527 (srai).

agatti:
Both documents' instruction descriptions match indeed.  Still, thanks for the snippet, I'll look into it later today.

agatti:
(Just a quick test, I'll update this comment later.)

The opcode generated for `srai(a2, a2, 24)` is correct:

```

**Comment by agatti** [comment]
> ```
> @micropython.asm_xtensa
> def asr24(a2):
>     srai(a2, a2, 24)
> 
> print(hex(asr1(0x12345678)), hex(0x12345678>>24))
> ```

You're calling `asr1(...)` whilst the test is called `asr24(...)`... :)

**Comment by rkompass** [comment]
> Both documents' instruction descriptions match indeed.

Yeah. And I also checked the code. It is correct now. And the tests pass. I had the mistake that you found in it.

It might be useful to add the test for asr for 24 bits anyway, because of the complications.

```
@micropython.asm_xtensa
def asr24(a2):
    srai(a2, a2, 24)
```

My translation of the encoding is correct but made things more complicated.

I replaced the data statements by the slli and srli in my CRC code and the tests for correctness pass.

**Comment by agatti** [comment]
> It might be useful to add the test for asr for 24 bits anyway, because of the complications.

No worries, just pushed a new version with an additional ASR test.

**Comment by agatti** [comment]
I left over some code that was used by opcodes that don't work on the 8266, now removed.  Sorry!

**Comment by dpgeorge** [comment]
Thanks for this, it's a very nice enhancement.

I tested it on an esp8266 and the new tests all pass.

The code size increase for this PR is +2200 bytes on ESP8266_GENERIC.

File: py/asmxtensa.c [hunk 1]
```diff
@@ -37,6 +37,9 @@
 #define WORD_SIZE (4)
 #define SIGNED_FIT8(x) ((((x) & 0xffffff80) == 0) || (((x) & 0xffffff80) == 0xffffff80))
 #define SIGNED_FIT12(x) ((((x) & 0xfffff800) == 0) || (((x) & 0xfffff800) == 0xfffff800))
+#define SIGNED_FIT18(x) ((((x) & 0xfffc0000) == 0) || (((x) & 0xfffc0000) == 0xfffc0000))
```
dpgeorge, line 40: According to the existing pattern this should either be called `SIGNED_FIT19`, or the mask should be `0xfffe0000`.

py/asmxtensa.c
```diff
@@ -264,4 +267,31 @@ void asm_xtensa_call_ind_win(asm_xtensa_t *as, uint idx) {
...
+    int32_t rel = dest - as->base.code_offset - 3;
+    if (as->base.pass == MP_ASM_PASS_EMIT && (((rel & 0x3) != 0) || !SIGNED_FIT18(rel))) {
+        mp_obj_new_exception_msg_varg(&mp_type_RuntimeError, ET_OUT_OF_RANGE, MP_QSTR_call0);
+    }
+    asm_xtensa_op_call0(as, rel);
+}
+
+void asm_xtensa_l32r(asm_xtensa_t *as, mp_uint_t reg, mp_uint_t label) {
+    uint32_t dest = get_label_dest(as, label);
+    int32_t rel = dest - as->base.code_offset;
```
dpgeorge, line 290: Does this `rel` computation need to align things, as done in `asm_xtensa_op_l32r`?  Otherwise there could be an edge case in checking if the relative offset fits or not.

py/emitinlinextensa.c
```diff
@@ -237,21 +216,58 @@ static const opcode_table_3arg_t opcode_table_3arg[] = {
...
+};
+#endif
+
 static void emit_inline_xtensa_op(emit_inline_asm_t *emit, qstr op, mp_uint_t n_args, mp_parse_node_t *pn_args) {
     size_t op_len;
     const char *op_str = (const char *)qstr_data(op, &op_len);
 
     if (n_args == 0) {
-        if (op == MP_QSTR_ret_n) {
+        if (op == MP_QSTR_ret_n || op == MP_QSTR_ret) {
```
dpgeorge, line 250: Is `ret` really the same as `ret_n`?

agatti: According to Cadence's documentation it seems to be (page 565):

> **Description**
>
> `RET.N` is the same as `RET` in a 16-bit encoding. `RET` returns from a routine called by `CALL0` or `CALLX0`.

py/emitinlinextensa.c
```diff
@@ -263,17 +279,49 @@ static void emit_inline_xtensa_op(emit_inline_asm_t *emit, qstr op, mp_uint_t n_
...
-        if (op == MP_QSTR_beqz) {
-            int label = get_arg_label(emit, op_str, pn_args[1]);
+        for (size_t index = 0; index < MP_ARRAY_SIZE(BCCZ_OPCODES); index++) {
+            if (op == BCCZ_OPCODES[index]) {
+                mp_uint_t label = get_arg_label(emit, op_str, pn_args[1]);
+                asm_xtensa_bccz_reg_label(&emit->as, index, r0, label);
+                return;
+            }
+        }
+        if (op == MP_QSTR_beqz_n) {
```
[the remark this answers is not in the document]
agatti: That is indeed on purpose.  The idea was to still generate valid code even when the label is out of range, as adding the extra checks with an error message would increase the code size by a larger amount than silently converting the branch opcode into its wide equivalent.

I'd suggest keeping those two opcodes, as then users would need to remember that *almost* all narrow opcodes are supported and which ones aren't.  Now, users going as far as to write inline assembler probably know what they're doing, but for folks just starting to write Xtensa assembler code this saves a couple of headaches (this is probably going to be more important if ESP32 support ends up in MicroPython, given the platform's popularity).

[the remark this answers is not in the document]
dpgeorge: OK, so we can keep these names.

But, can they be put into `BCCZ_OPCODES` table, to reduce code size?  Maybe:
```c
 static const qstr_short_t BCCZ_OPCODES[] = { MP_QSTR_beqz, MP_QSTR_bnez, MP_QSTR_bltz, MP_QSTR_bgez,  MP_QSTR_beqz_n, MP_QSTR_bnez_n };

// then mask the `index`:
asm_xtensa_bccz_reg_label(&emit->as, index & 3, r0, label);
```

dpgeorge:
> 2.2KiB... I thought it took less space.

This is not a show stopper for this PR.  If we are short on space in the future, we can look into optimisations then.


> Given that the plan is to eventually submit a PR to make this feature available on ESP32 as well, would it make sense to add a PR to add a QEMU Xtensa port in order to be able to put the inline assembler (and natmods too) under CI?

If that's not too much work/effort, then yes that would be a good idea.

py/emitinlinextensa.c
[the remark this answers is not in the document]
agatti: Indeed, that will be in the next iteration.  Thanks for the hint.

py/asmxtensa.c
agatti: You're right - it will be in the final iteration.  I need to add a couple more tests for `J` and `JX` opcodes, a new commit to skip emitting the prologue's initial jump skipping the constant pool if said pool is empty (as it is a possibility with inline assembler blocks) and then I guess this PR should be pretty much ready on my end.

agatti:
I believe this should be in its final form.  Changes since last review round:

* `J` and `JX` opcodes have their own tests
* `NSA` and `NSAU` opcodes have been added, each with its own test
* Xtensa prologue will skip dealing with the constants table if it is empty - extremely unlikely for native code, but a common occurrence for inline assembler
* `BEQZ.N` and `BNEZ.N` have been folded into `BEQZ` and `BNEZ`, respectively
* `L32R` address checks also take alignment into account.

@rkompass If you want to start to work on documentation when you've got some time, that'd be great :)

Anyway, once I'm done with Picolibc stuff for natmods I'll start working on ESP32/ESP32S3 (LX6+LX7) opcodes.  I'll skip floating point opcodes as I don't think any other inline assembler implementation supports them anyway.

rkompass:
@agatti I will start in a few days, hopefully.

py/asmxtensa.c
```diff
@@ -150,7 +155,7 @@ void asm_xtensa_bccz_reg_label(asm_xtensa_t *as, uint cond, uint reg, uint label
     uint32_t dest = get_label_dest(as, label);
     int32_t rel = dest - as->base.code_offset - 4;
     if (as->base.pass == MP_ASM_PASS_EMIT && !SIGNED_FIT12(rel)) {
-        printf("ERROR: xtensa bccz out of range\n");
+        mp_obj_new_exception_msg_varg(&mp_type_RuntimeError, ET_OUT_OF_RANGE, MP_QSTR_bccz);
```
dpgeorge, line 158: These will need to actually raise the exception, ie be wrapped in `nlr_raise(...)`.

There are 5 instances here that need fixing.

tests/inlineasm/xtensa/asmloadstore.py
```diff
@@ -0,0 +1,93 @@
...
+DWORD_DATA = array.array("i", (100_000, -200_000, 300_000, -400_000))
+
+
+@micropython.asm_xtensa
+def tl32r() -> int:
+    j(CODE)
+    label(DATA)
+    data(1, 1, 2, 3, 4, 5, 6, 7)
+    label(CODE)
+    l32r(a2, DATA)
```
dpgeorge, line 18: This test fails the alignment requirements for `l32r`.

If you insert 3x nop's after `data()` then it works.

But then if you insert a nop after the `j(CODE)` it gives different answers!

dpgeorge (commented):
Apart from the comments above, this is looking very good now.

py/asmxtensa.c
agatti: My bad, that's a copy and paste gone wrong - I copied that bit from `emit_inline_xtensa_error_exc(emit, …);` in `py/emitinlinextensa.c` - I removed the emitter bit to replace with something appropriate and forgot to raise the exception.  Thanks for letting me know!

tests/inlineasm/xtensa/asmloadstore.py
agatti: This is weird, because I recall this working back then.  It might have been some changes that were probably stashed and then lost when I re-cloned the repo.

Anyway, I'll look into it today.

agatti: OK, the `l32r` test should be sorted out.  The code was pretty much correct, but it depends on the both the target address and the relative distance being word-aligned.

Unless I missed something, labels in inline asm blocks do not have a default alignment, so additional `align(4)` directives had to be strategically placed to make things work.  I wonder if there's a way to easily override the behaviour of `label()` just for xtensa, so those alignment directives can be made implicit.

py/asmxtensa.c
```diff
@@ -264,4 +271,31 @@ void asm_xtensa_call_ind_win(asm_xtensa_t *as, uint idx) {
...
+        mp_raise_msg_varg(&mp_type_RuntimeError, ET_OUT_OF_RANGE, MP_QSTR_call0);
+    }
+    asm_xtensa_op_call0(as, rel);
+}
+
+void asm_xtensa_l32r(asm_xtensa_t *as, mp_uint_t reg, mp_uint_t label) {
+    uint32_t dest = get_label_dest(as, label);
+    int32_t rel = dest - as->base.code_offset;
+    if (as->base.pass == MP_ASM_PASS_EMIT && (((dest & 0x03) != 0) || (rel >= 0) || ((rel & 0x03) != 0) || !SIGNED_FIT18(rel))) {
+        mp_raise_msg_varg(&mp_type_RuntimeError, ET_OUT_OF_RANGE, MP_QSTR_l32r);
```
dpgeorge, line 296: Is it worth having a different error message for the case `dest` or `rel` are not aligned?  It's quite confusing to get the message "out of range" when it's clearly in range.

Could also do the same in `asm_xtensa_call0` above.

tests/inlineasm/xtensa/asmloadstore.py
dpgeorge: I feel like it should be up to the user to know about alignment in this particular case.  Because if all labels are automatically aligned (and padded with `nop_n`?) then that could very well be unexpected to the user.  And alignment is only necessary for l32r, right?

Probably documentation is the best way to address this.  And good error messages, see above about emitting a different error for out-of-alignment.

agatti: From what I can see, word-alignment is only necessary for `l32r` and `call0`-like opcodes.  Looking for `32-bit aligned` in Cadence's ISA documentation brings up only those opcodes.

I agree on the out of alignment message, I'll make the necessary changes to have something like `ERROR: label "%q" %d bytes out of alignment` in those cases.

py/asmxtensa.c
dpgeorge: @agatti just in case it dropped off your list, this is still outstanding on this PR.  Once the error messages are fixed, I think this should be good to go.

agatti: Sorry, I just pushed the error messages change.

I've tried adding and removing nops/align directives from the `tl32r()` test in `tests/inlineasm/xtensa/asmloadstore.py` and I get appropriate alignment error messages depending on the change.

So,

```py
@micropython.asm_xtensa
def tl32r() -> int:
    nop()
    j(CODE)
    nop()
    label(DATA)
    data(1, 1, 2, 3, 4, 5, 6, 7)
    align(4)
    label(CODE)
    nop_n()
    nop_n()
    l32r(a2, DATA)

tl32r()
```

yields "RuntimeError: ERROR: l32r target not word-aligned", and

```py
@micropython.asm_xtensa
def tl32r() -> int:
    nop()
    j(CODE)
    align(4)
    label(DATA)
    data(1, 1, 2, 3, 4, 5, 6, 7)
    align(4)
    label(CODE)
    nop_n()
    l32r(a2, DATA)

tl32r()
```

yields "RuntimeError: ERROR: l32r location not word-aligned".

The alignment check code is the same for both `l32r` and `call0`.

dpgeorge: Thanks, this all works well now.

dpgeorge (approved):
Tested with ESP32_GENERIC and ESP8266_GENERIC firmware.  Works well.

Thank you!

PR #16733, opened 2025-02-11
rp2: Fix build failure if threads are disabled.

### LINKED ISSUE #16720: rp2: new mp_thread_recursive_mutex_t code breaks with MICROPY_PY_THREAD disabled

rp2

### MicroPython version

N/A (Build from master)

### Reproduction

This change breaks builds with threading disabled, since `mp_thread_recursive_mutex_t ` is never defined for `pendsv.c`:

https://github.com/micropython/micropython/commit/3bfedd0f4a9e8b11ff97851615ec4761663b0e94

To reproduce, build any rp2 board with `#define MICROPY_PY_THREAD                       (0)`

### Expected behaviour

_No response_

### Observed behaviour

```
micropython/ports/rp2/pendsv.c:50:8: error: unknown type name 'mp_thread_recursive_mutex_t'
   50 | static mp_thread_recursive_mutex_t pendsv_mutex;
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Additional Information

### Code of Conduct

Yes, I agree

projectgus:
### Summary

Regression in 3af006efb39ad0b7aa7c0401c93329b654bca617 meant that pendsv.c no longer compiled if threads were disabled in the build config. Add an implementation based on the earlier one (simple counter) for the non-threads case.

To add CI coverage for future breakage of this kind, Damien suggested building one of the boards in CI with threads disabled. To support this, I've added support for the `CFLAGS_EXTRA` environment variable in CMake-based ports, similar to other ports (with caveat that it only applies to clean builds).

Closes #16720

### Testing

* Built `RPI_PICO` default config before and after this change, disassembled with `arm-none-eabi-objdump -d`, and diffed the output to confirm no binary changes.
* Built `RPI_PICO` with `MICROPY_PY_THREAD` defined to 0, confirmed unit tests all pass (except for threads tests).
* Pushed a temporary commit d78f558 to check that the rp2 CI build is applying the config correctly. ([Confirmed](https://github.com/micropython/micropython/actions/runs/13257398311/job/37006503860?pr=16733))

* I originally implemented the new counter using atomic increment/decrement, but looking at the usage patterns I couldn't see a reason for this if there's no contention from the second core or multiple threads. Any interrupt handler should exit with the counter at the same value that it was when entering.
* Feels kind of hacky to be making a custom config just for CI, but there's no boards in MicroPython that depend on not having thread support.

### DIFF
--- a/ports/rp2/pendsv.c
+++ b/ports/rp2/pendsv.c
@@ -43,10 +43,14 @@
 
 static pendsv_dispatch_t pendsv_dispatch_table[PENDSV_DISPATCH_NUM_SLOTS];
 
+static inline void pendsv_resume_run_dispatch(void);
+
 void PendSV_Handler(void);
 
-// Using the nowait variant here as softtimer updates PendSV from the loop of mp_wfe_or_timeout(),
-// where we don't want the CPU event bit to be set.
+#if MICROPY_PY_THREAD
+
+// Important to use a 'nowait' mutex here as softtimer updates PendSV from the
+// loop of mp_wfe_or_timeout(), where we don't want the CPU event bit to be set.
 static mp_thread_recursive_mutex_t pendsv_mutex;
 
 void pendsv_init(void) {
@@ -62,7 +66,40 @@ void pendsv_suspend(void) {
 
 void pendsv_resume(void) {
     mp_thread_recursive_mutex_unlock(&pendsv_mutex);
+    pendsv_resume_run_dispatch();
+}
 
+static inline int pendsv_suspend_count(void) {
+    return pendsv_mutex.mutex.enter_count;
+}
+
+#else
+
+// Without threads we don't include any pico-sdk mutex in the build,
+// but also we don't need to worry about cross-thread contention (or
+// races with interrupts that update this counter).
+static int pendsv_lock;
+
+void pendsv_init(void) {
+}
+
+void pendsv_suspend(void) {
+    pendsv_lock++;
+}
+
+void pendsv_resume(void) {
+    assert(pendsv_lock > 0);
+    pendsv_lock--;
+    pendsv_resume_run_dispatch();
+}
+
+static inline int pendsv_suspend_count(void) {
+    return pendsv_lock;
+}
+
+#endif
+
+static inline void pendsv_resume_run_dispatch(void) {
     // Run pendsv if needed.  Find an entry with a dispatch and call pendsv dispatch
     // with it.  If pendsv runs it will service all slots.
     int count = PENDSV_DISPATCH_NUM_SLOTS;
@@ -76,7 +113,7 @@ void pendsv_resume(void) {
 
 void pendsv_schedule_dispatch(size_t slot, pendsv_dispatch_t f) {
     pendsv_dispatch_table[slot] = f;
-    if (pendsv_mutex.mutex.enter_count == 0) {
+    if (pendsv_suspend_count() == 0) {
         #if PICO_ARM
         // There is a race here where other core calls pendsv_suspend() before
         // ISR can execute, but dispatch will happen later when other core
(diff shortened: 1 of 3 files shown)


projectgus:
@Gadgetoid Are you folks in a position to do any testing of this fix?

Gadgetoid:
Oooh! Thank you, will test ASAP!

ports/rp2/pendsv.c
```diff
@@ -62,7 +66,41 @@ void pendsv_suspend(void) {
...
+static inline int pendsv_suspend_count(void) {
+    return pendsv_mutex.mutex.enter_count;
+}
+
+#else
+
+// Without threads we don't include any pico-sdk mutex in the build,
+// but also we don't need to worry about cross-thread contention (or
+// races with interrupts that update this counter).
+static int pendsv_counter;
```
dpgeorge, line 81: This was originally called `pendsv_lock`... is it more self-documenting to use that name again?

ports/rp2/pendsv.c
```diff
@@ -62,7 +66,41 @@ void pendsv_suspend(void) {
...
+
+void pendsv_init(void) {
+
```
dpgeorge, line 84: Please remove blank line here.

ports/rp2/pendsv.c
```diff
@@ -62,7 +66,41 @@ void pendsv_suspend(void) {
...
+}
+
+void pendsv_suspend(void) {
+    pendsv_counter += 1;
```
dpgeorge, line 88: I suggest using `++` to match the `--` used below.

ports/rp2/pendsv.c
```diff
@@ -76,7 +114,7 @@ void pendsv_resume(void) {
 
 void pendsv_schedule_dispatch(size_t slot, pendsv_dispatch_t f) {
     pendsv_dispatch_table[slot] = f;
-    if (pendsv_mutex.mutex.enter_count == 0) {
+    if (pendsv_suspend_count() == 0) {
```
dpgeorge, line 117: Maybe call it `pendsv_lock_count`?

ports/rp2/pendsv.c
```diff
@@ -97,13 +135,19 @@ void pendsv_schedule_dispatch(size_t slot, pendsv_dispatch_t f) {
...
     if (!mp_thread_recursive_mutex_lock(&pendsv_mutex, 0)) {
         // Failure here means core 1 holds pendsv_mutex. ISR will
         // run again after core 1 calls pendsv_resume().
         return;
     }
     // Core 0 should not already have locked pendsv_mutex
     assert(pendsv_mutex.mutex.enter_count == 1);
+    #else
+    if (pendsv_suspend_count() > 0) {
+        return; // Will run again after pendsv_resume()
```
dpgeorge, line 148: The original implementation simply had `assert(pendsv_lock == 0)` here, which is much stricter.  Did you find cases where this handler can be called with `pendsv_lock > 0` ?

dpgeorge (commented):
I checked this against 3af006efb39ad0b7aa7c0401c93329b654bca617 and (apart from my comments) it effectively reverts that commit, which is what we want.

ports/rp2/pendsv.c
projectgus: This function returns the count of times `pendsv_suspend()` has been called, but I can change the name if you prefer.

ports/rp2/pendsv.c
projectgus: Probably not on single threaded, assuming the PendSV runs without latency. Will change back to the assert.

ports/rp2/pendsv.c
dpgeorge: Oh, right.  Keep it as `pendsv_suspend_count()`.

dpgeorge (approved):

dpgeorge:
@Gadgetoid did you want to test this fix?  It looks good to me, but maybe worth testing with your build setup.

Gadgetoid:
Sorry I fell down a rabbit hole of distractions! I'll give it a shot now.

dpgeorge:
Thanks @Gadgetoid for testing!

PR #16735, opened 2025-02-11
samd: Allow using default devices for UART, I2C and SPI.

robert-hh:
### Summary

Some of the SAMD boards show pin names for UART, I2C and SPI on their silkscreen. At these
boards Pin name aliases existed already to use e.g. "SCL" or "TX" as pin name wherever a Pin had to be specified. This PR allows to omit the device ID and Pin definitions in the constructor if
the default device is going to be used. So the default  UART, I2C or SPI device can be instantiated
with:

```
from machine import UART, I2C, SPI
uart = UART()
i2c = I2C()
spi = SPI()
```
The documentation was updated accordingly to reflect the changes and fix bugs.

### Testing

The change has been tested to return the proper device with:

Adafruit Feather M0
Adafruit Feather M4
Adafruit ItsyBitsy M0
Adafruit ItsyBitsy M4
Adafruit Metro Express M4
Adafruit Trinket M0
Adafruit QT PY
Seeed XIAO SAMD21
Seeed WIO terminal
Minisam M4
Sparkfun SAMD51 Thing plus 

It has been tested to raise an error if the device is NOT fully specified with the SAMD Generic builds.


ports/samd/machine_uart.c
```diff
@@ -479,12 +486,18 @@ static void mp_machine_uart_init_helper(machine_uart_obj_t *self, size_t n_args,
...
+    int uart_id = MICROPY_HW_DEFAULT_UART_ID;
+    int arg_offset = 0;
+
+    if (n_args > 0 && mp_obj_get_int(args[0]) <= SERCOM_INST_NUM) {
+        uart_id = mp_obj_get_int(args[0]);
+        arg_offset = 1;
+    }
     if (uart_id < 0 || uart_id > SERCOM_INST_NUM) {
-        mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("UART(%d) doesn't exist"), uart_id);
+        mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("UART(%d) does not exist"), uart_id);
```
dpgeorge, line 500: Please keep this unchanged.  The word `doesn't` is slightly smaller in code size and we use contracted words in all other error messages (or at least that's the aim).

ports/samd/machine_uart.c
```diff
@@ -479,12 +486,18 @@ static void mp_machine_uart_init_helper(machine_uart_obj_t *self, size_t n_args,
...
 static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
-    mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
+    mp_arg_check_num(n_args, n_kw, 0, MP_OBJ_FUN_ARGS_MAX, true);
 
     // Get UART bus.
-    int uart_id = mp_obj_get_int(args[0]);
+    int uart_id = MICROPY_HW_DEFAULT_UART_ID;
+    int arg_offset = 0;
+
+    if (n_args > 0 && mp_obj_get_int(args[0]) <= SERCOM_INST_NUM) {
```
dpgeorge, line 495: No need for the second check here against `SERCOM_INST_NUM`, that will be handled just below.

(Comment also applied to other similar code in I2C/SPI.)

ports/samd/machine_uart.c
```diff
@@ -479,12 +486,18 @@ static void mp_machine_uart_init_helper(machine_uart_obj_t *self, size_t n_args,
 }
 
 static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
-    mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
+    mp_arg_check_num(n_args, n_kw, 0, MP_OBJ_FUN_ARGS_MAX, true);
```
dpgeorge, line 489: Maybe worth doing `MICROPY_HW_DEFAULT_UART_ID < 0 ? 1 : 0` for the third argument, so if there's no default UART ID for the given board, the error message is slightly better (it'll say that the constructor is missing a required positional argument).

ports/samd/machine_spi.c
```diff
@@ -132,14 +146,16 @@ static void machine_spi_init(mp_obj_base_t *self_in, size_t n_args, const mp_obj
...
+    if (args[ARG_miso].u_int != -1 && args[ARG_miso].u_obj != mp_const_none) {
         self->miso = mp_hal_get_pin_obj(args[ARG_miso].u_obj);
     }
-
     // Initialise the SPI peripheral if any arguments given, or it was not initialised previously.
     if (n_args > 0 || kw_args->used > 0 || self->new) {
         self->new = false;
 
+        if (self->sck == 0xff || self->mosi == 0xff) {
+            mp_raise_ValueError(MP_ERROR_TEXT("mosi or sck missing"));
```
dpgeorge, line 157: If you want to reduce code size, I suggest "missing sck/mosi".

ports/samd/machine_i2c.c
```diff
@@ -146,15 +152,21 @@ mp_obj_t machine_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
...
-    self->sda = mp_hal_get_pin_obj(args[ARG_sda].u_obj);
-    sda_pad_config = get_sercom_config(self->sda, self->id);
+    if (args[ARG_scl].u_obj != MP_ROM_NONE) {
+        self->scl = mp_hal_get_pin_obj(args[ARG_scl].u_obj);
+    }
+    if (args[ARG_sda].u_obj != MP_ROM_NONE) {
+        self->sda = mp_hal_get_pin_obj(args[ARG_sda].u_obj);
+    }
+    if (self->scl == 0xff || self->sda == 0xff) {
+        mp_raise_ValueError(MP_ERROR_TEXT("sda or scl missing"));
```
dpgeorge, line 166: Suggest "missing scl/sda".

ports/samd/machine_uart.c
robert-hh: OK.

ports/samd/machine_uart.c
robert-hh: Done. Changed for UART/SPI. I2C did not need the change.

ports/samd/machine_uart.c
robert-hh: Good suggestion. Done.

ports/samd/machine_spi.c
robert-hh: Reducing the code size is always good, even by a few bytes.

ports/samd/machine_uart.c
robert-hh: To have the same behavior for I2C a mp_arg_check_num() has to be added. That increases the code size by 20 bytes. Even without that, the missing argument is flagged, only with a different error message of `ValueError: I2C(-1) doesn't exist` instead of `TypeError: argument num/types mismatch`.

dpgeorge: For I2C you can do the following at zero code size cost:
```c
#if MICROPY_HW_DEFAULT_I2C_ID < 0            
{ MP_QSTR_id, MP_ARG_REQUIRED | MP_ARG_INT, {.u_int = -1} },             
#else
{ MP_QSTR_id, MP_ARG_INT, {.u_int = MICROPY_HW_DEFAULT_I2C_ID} },
#endif
```

ports/samd/machine_uart.c
```diff
@@ -479,10 +486,16 @@ static void mp_machine_uart_init_helper(machine_uart_obj_t *self, size_t n_args,
...
+    mp_arg_check_num(n_args, n_kw, MICROPY_HW_DEFAULT_UART_ID < 0 ? 1 : 0, MP_OBJ_FUN_ARGS_MAX, true);
 
     // Get UART bus.
-    int uart_id = mp_obj_get_int(args[0]);
+    int uart_id = MICROPY_HW_DEFAULT_UART_ID;
+    int arg_offset = 0;
+
+    if (n_args > 0) {
+        uart_id = mp_obj_get_int(args[0]);
+        arg_offset = 1;
```
dpgeorge, line 497: Instead of adding `arg_offset` variable, you could do:
```c
++args;
--n_args;
```

ports/samd/machine_uart.c
robert-hh: That's a good hint. Done.

ports/samd/machine_uart.c
robert-hh: Made that. But I changed the logic or probing the first positional argument, which could either be the device ID or the baudrate, in case that the default ID is used. If the value of the first argument is small, it's taken as the device ID. So I restored the first check and skipped the second.

dpgeorge: That seems a bit ill defined, small vs large values.  In this case I think it's best the user uses a keyword argument for the baudrate, eg `machine.UART(baudrate=115200)`.

dpgeorge: > Requiring a keyword for baudrate breaks existing Python code.

Sorry, I didn't mean to change any existing requirements for the baudrate parameter.  You can already specify this by keyword, although it can also be positional.

What I mean is that, if the user wants to use the default UART instance without specifying an id, and they also want to configure some parameters of that UART, then they need to use keyword arguments for all those parameters.  That's already supported in the code, so it's just a documentation issue.

As for the code change here, my suggestion is that it does not distinguish between small/large integers.  It just always assumes the first positional argument (if given) is the UART id.  Although it would be neat to just write `UART(9600)` that's too ambiguous when one also expects to be able to write `UART(2)` and `UART(2, 9600)`.  It's better to be clear of your intent in the code, so use `UART(baudrate=9600)` instead.

robert-hh: Thanks. I changed the code back and added a few words and examples to the quickref documentation. The changes are pushed.

dpgeorge: Thanks for changing the code, it looks good now.

This also matches the recently added bb4ec886f8994ad7b2db6f80c7b46e6eedbc40e7 and c143eb50248b3a3f521775f09f5073675095e48a

ports/samd/boards/SEEED_WIO_TERMINAL/mpconfigboard.h
```diff
@@ -3,4 +3,11 @@
 
 #define MICROPY_HW_XOSC32K          (1)
 
+#define MICROPY_HW_DEFAULT_UART_ID  (2)
+#define MICROPY_HW_DEFAULT_I2C_ID   (4)
+#define MICROPY_HW_DEFAULT_SPI_ID   (5)
+// Map SDA0/SCL0 to SDA/SCL
+#define pin_SDA  pin_SDA0
+#define pin_SCL  pin_SCL0
```
dpgeorge, line 11: You can put this in pins.csv instead, which I think is a little cleaner, as it keeps all the pin definitions in one place:
```
SDA,PA13
SCL,PA12
```

ports/samd/machine_uart.c
```diff
@@ -298,7 +299,7 @@ static void mp_machine_uart_print(const mp_print_t *print, mp_obj_t self_in, mp_
     }
 
     mp_printf(print, "UART(%u, baudrate=%u, bits=%u, parity=%s, stop=%u, "
-        "timeout=%u, timeout_char=%u, rxbuf=%d"
+        "rx=\"%q\", tx=\"%q\", timeout=%u, timeout_char=%u, rxbuf=%d"
```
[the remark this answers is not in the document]
robert-hh: Done. It is not critical.

ports/samd/boards/SEEED_WIO_TERMINAL/mpconfigboard.h
robert-hh: Done.

ports/samd/machine_i2c.c
```diff
@@ -119,25 +120,34 @@ void common_i2c_irq_handler(int i2c_id) {
...
+        #endif
         { MP_QSTR_freq, MP_ARG_INT, {.u_int = DEFAULT_I2C_FREQ} },
-        { MP_QSTR_scl, MP_ARG_REQUIRED | MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
-        { MP_QSTR_sda, MP_ARG_REQUIRED | MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
+        #if defined(pin_SCL) && defined(pin_SDA)
+        { MP_QSTR_scl, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = pin_SCL} },
+        { MP_QSTR_sda, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = pin_SDA} },
+        #else
+        { MP_QSTR_scl, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_sda, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
```
dpgeorge, line 141: Do you think in this case, of no default scl/sda, these args should retain the `MP_ARG_REQUIRED` flag?

(Also, if the default ID is 0 but you do `machine.I2C(1)` then it configures instance #1 with the default pins used by instance #0.  I guess that's OK, but maybe a better way would be to only use the default pins if the default instance is selected?)

ports/samd/machine_spi.c
```diff
@@ -82,8 +83,15 @@ void common_spi_irq_handler(int spi_id) {
...
 static void machine_spi_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     machine_spi_obj_t *self = MP_OBJ_TO_PTR(self_in);
-    mp_printf(print, "SPI(%u, baudrate=%u, firstbit=%u, polarity=%u, phase=%u, bits=8)",
-        self->id, self->baudrate, self->firstbit, self->polarity, self->phase);
+    mp_printf(print, "SPI(%u, baudrate=%u, firstbit=%u, polarity=%u, phase=%u, bits=8,"
+        " sck=\"%q\", mosi=\"%q\", miso=",
+        self->id, self->baudrate, self->firstbit, self->polarity, self->phase,
+        pin_find_by_id(self->sck)->name, pin_find_by_id(self->mosi)->name);
+    if (self->miso == 0xff) {
+        mp_printf(print, "None");
```
dpgeorge, line 91: This is missing a closing ")".

ports/samd/machine_spi.c
```diff
@@ -94,9 +102,15 @@ static void machine_spi_init(mp_obj_base_t *self_in, size_t n_args, const mp_obj
...
         { MP_QSTR_firstbit, MP_ARG_INT, {.u_int = -1} },
+        #if defined(pin_SCK) && defined(pin_MOSI) && defined(pin_MISO)
+        { MP_QSTR_sck, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = pin_SCK} },
+        { MP_QSTR_mosi, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = pin_MOSI} },
+        { MP_QSTR_miso, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = pin_MISO} },
+        #else
         { MP_QSTR_sck, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
         { MP_QSTR_mosi, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
-        { MP_QSTR_miso, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_miso, MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_int = -1} },
```
dpgeorge, line 112: This use of `MP_ARG_OBJ` with `.u_int` is invalid.  Needs to be one of:
1. `MP_ARG_INT` and `.u_int = -1`
2. `MP_ARG_OBJ` and `.u_rom_obj = MP_ROM_INT(-1)`

I think you want the latter, so the user can pass in a pin object.

ports/samd/machine_i2c.c
robert-hh: You're right, these must be required in case of missing defaults. 
About the second question: SAMD21 allow to use two different IDs at one pin pair for I2C and UART. For example the UART ID  for RX and TX of a ItsyBitsy M0 may be 0 or 2. At SAMD51 it's similar in that the same pin pair may be used for UART and I2C, but with different IDs the pins are swapped.

ports/samd/machine_spi.c
robert-hh: Good find.

ports/samd/machine_spi.c
robert-hh: OK. Somehow it worked before.

robert-hh: Looking at it, the -1 case can be dropped. Setting the default to MP_CONST_NONE is sufficient for the logic. And it simplifies the code.

dpgeorge (approved):
Thanks for updating.  I tested this on ADAFRUIT_ITSYBITSY_M0_EXPRESS and it works well.

I think this is a really nice improvement to the user experience, you no longer need to know about instance IDs to get started using the board.  You can just create the default I2C/UART/SPI and it'll work correctly on the pins printed on the silkscreen.  You only need to know about instance IDs if you want to delve deeper.

robert-hh:
Thank you for your careful review and merging. Yes, it's more convenient now. I had to look up the ID numbers myself when testing with one of the rarely used boards.

robert-hh:
> I did massage the commits a little before merging, because there were some review fix-ups in latter commits that belonged in earlier commits.

Thank you. I noticed that and was confused at first. It looks better now.

PR #16737, opened 2025-02-12
esp32/espnow: Add support for espnow v2.0.

AmirHmZz:
### Summary

2 years ago, [I've submitted a feature request](https://github.com/espressif/esp-idf/issues/9453) to increase maximum length of `ESP-NOW` messages. [Finally](https://github.com/espressif/esp-idf/commit/b4d102796f50e6fabce0f2a1c6ff5b8f39c902a2) the maximum length increased from **250** bytes to **1490** bytes in `ESP-NOW v2.0` which is implemented in `IDFv5.4`. This PR aims to migrate current implementation of `espnow` module to support `ESP-NOW v2.0` but also keep support for `ESP-NOW v1.0` for builds with older `IDF` versions.

### Testing

I've tested both sending and receiving examples on `ESP32` and `ESP32-S3` chips.

### DIFF
--- a/docs/library/espnow.rst
+++ b/docs/library/espnow.rst
@@ -32,23 +32,42 @@ ESP-NOW is a connection-less wireless communication protocol supporting:
 
 - Direct communication between up to 20 registered peers:
 
-  - Without the need for a wireless access point (AP),
+- Without the need for a wireless access point (AP),
 
 - Encrypted and unencrypted communication (up to 6 encrypted peers),
 
-- Message sizes up to 250 bytes,
+- Message sizes up to 1470 bytes (For ESP-NOW v2),
 
-- Can operate alongside Wifi operation (:doc:`network.WLAN<network.WLAN>`) on
+- Can operate alongside Wi-Fi operation (:doc:`network.WLAN<network.WLAN>`) on
   ESP32 and ESP8266 devices.
+
+- Track the Wi-Fi signal strength (RSSI) of ESP-NOW peer devices.
 
 It is especially useful for small IoT networks, latency sensitive or power
 sensitive applications (such as battery operated devices) and for long-range
 communication between devices (hundreds of metres).
 
-This module also supports tracking the Wifi signal strength (RSSI) of peer
-devices.
+ESP-NOW Versions
+~~~~~~~~~~~~~~~~
 
-A simple example would be:
+Since ESP-IDF V5.4, two ESP-NOW versions are supported when running on ESP32:
+V1 and V2.
+
+- The maximum packet length supported by V2 devices is 1470 bytes
+- The maximum packet length supported by V1 devices is 250 bytes.
+
+To check at runtime whether ESP-NOW V2 is available, check the value of
+`espnow.MAX_DATA_LEN`.
+
+ESP-NOW V2 devices are capable of receiving packets from both V2 and V1 devices.
+
+ESP-NOW V1 devices (including ESP8266) can receive packets from other V1
+devices, or from V2 devices if the packet length doesn't exceed 250 bytes. For
+packets exceeding this length, a V1 device will either truncate the data to the
+first 250 bytes or discard the packet entirely.
+
+Example
+~~~~~~~
 
 **Sender:** ::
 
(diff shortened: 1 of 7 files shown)


glenn20:
@AmirHmZz  Thanks for your persistence in getting this change into the espressif IDE and now to producing this PR. I will endeavour to provide an initial review in the next few days.

gretel:
thanks!

AmirHmZz:
@glenn20 still waiting for your review!

mrichar1:
I've just tested this sending and receiving between two ESP32 C3 devices, and it seems to be working as expected, but with one caveat. It looks like the max value of 1490 might no longer be correct.

Looking at the Espressif arduino-esp32 repository, the max length is now fixed at 1470, not 1490 bytes - which is also reflected in the documentation:

https://github.com/espressif/arduino-esp32/blob/631216af5dfd9ce46b6007e371415f24743b0fa5/libraries/ESP_NOW/src/ESP32_NOW.cpp#L12
https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_now.html

There are some hints that it was 1490 previously, but has been reduced, e.g.:

https://github.com/espressif/esp-wifi-remote/commit/09ffacc5a6c975179950238154a57c43cadd3822
https://github.com/espressif/esp-wifi-remote/blob/cbdedb76d28d3868378a73ce108ecc161dfe069b/components/esp_wifi_remote/idf_v5.5/include/injected/esp_now.h#L54


In testing I can only send a maximum of 1478 bytes. More than this and it silently fails (sender returns `True`, not exceptions raised at either end). I guess that 1490 might be a theoretical maximum, but they've decided to play it safe and limit to 1470 to allow for extra headers/metadata etc?

For completeness, I can confirm that testing this PR with all references changed from 1490 to 1470 also works as expected.

(My use-case has messages of max 1024 bytes, so this update would make my project very happy!)

docs/library/espnow.rst
```diff
@@ -45,6 +45,15 @@ It is especially useful for small IoT networks, latency sensitive or power
 sensitive applications (such as battery operated devices) and for long-range
 communication between devices (hundreds of metres).
 
+According to ESP-IDF, ESP-NOW supports two versions: v1.0 and v2.0. The maximum
```
projectgus:
```suggestion
Recent MicroPython releases support two ESP-NOW versions when running on ESP32: v1.0 and v2.0. The maximum
```

(This way the reader doesn't need to care about ESP-IDF specifically.)

docs/library/espnow.rst
```diff
@@ -148,10 +157,10 @@ Configuration
 
     .. data:: Options:
 
-        *rxbuf*: (default=526) Get/set the size in bytes of the internal
+        *rxbuf*: (default=528 or 3008) Get/set the size in bytes of the internal
```
projectgus: Changing the default here means an extra 2.5KB of RAM usage for ESPNow users, even if they don't use V2 features. That may cause problems when some people update their existing MicroPython and are tight on RAM.

How about keeping the default `rxbuf` size at the V1 level and documenting the need to call `config(rxbuf=)` with the recommended value if receiving V2-length messages?

ports/esp32/modespnow.c
```diff
@@ -420,7 +430,7 @@ static mp_obj_t espnow_recvinto(size_t n_args, const mp_obj_t *args) {
     #else
     uint8_t *peer_buf = _get_bytes_len_w(list->items[0], ESP_NOW_ETH_ALEN);
     #endif // MICROPY_PY_ESPNOW_RSSI
-    uint8_t *msg_buf = _get_bytes_len_w(msg, ESP_NOW_MAX_DATA_LEN);
+    uint8_t *msg_buf = _get_bytes_len_w(msg, ESP_NOW_MAX_POSSIBLE_DATA_LEN);
```
projectgus: This change will break existing Python code that calls `recvinto()` with a 250 byte buffer on ESP32.

To avoid the breaking change, need to find a way to support shorter buffers provided they're long enough for the incoming message. (I think it's OK to change the API to check the length per-message, as old code stays compatible with that, but need to make sure the ringbuffer doesn't get out of sync if a message is dropped due to length.)

projectgus (commented):
Thanks @AmirHmZz for submitting this. I was unaware of ESPNow V2 until I went to implement a fix for #18311 (unrelated), it looks quite useful!

Unfortunately there is some additional work needed to make sure we don't break existing MicroPython code when adding V2 support (see inline comments).

There is also a set of tests in `tests/multi_espnow` (run as `./run-multitests.py -p2 -t PORTA -t PORTB multi_espnow/*.py`). These used to pass on IDF <5.4, but at least one test (`20_send_echo.py`) is already broken in master (can already send long ESPNow messages with MicroPython, but can't receive them properly). It'd also be good to add a test specifically for the V2 long messages (that we can skip when running on esp8266). Are you able to work on the tests as part of this PR as well?

projectgus:
> at least one test (`20_send_echo.py`) is already broken in master

Oh, FWIW there's another failure in master + IDF 5.5.1 that I think is an ESP-IDF bug, `multi_espnow/70_channels.py` fails about 25% of the time *when transmitting from some chips only*. They seem to take a very long time to switch channels (so the message is sent to the old/wrong channel). If I have time I'll try to reproduce in C code and report to ESP-IDF.

projectgus:
@AmirHmZz Are you still interested in working on this one? I don't mind making a few updates and re-submitting it, if you're busy with other things.

AmirHmZz:
@projectgus Sorry for the late response. I'm still interested in this, but I live in Iran and the internet here was completely cut off for about 40 days. It just came back recently, but only GitHub and a few other services are accessible right now, and there's no guarantee it won't go down again.

Given the instability, please feel free to make those updates cause I definitely don't want this PR to stall because of my connectivity issues.

projectgus:
@AmirHmZz Wow, I appreciate you getting back to me at all. You've certainly got more important things to worry about than this PR, I'll try and make some small updates to get it merged. Hoping you can stay stay safe and well.

docs/library/espnow.rst
projectgus: Decided to keep this as-is, so updated ESP32s will get the larger buffer. Added a note in the docs about reducing this value to save RAM if only using ESP-NOW V1.

projectgus:
I've made some tweaks to @AmirHmZz's work and added a test.

I also noticed while working on this that Espressif reduced the maximum packet size for V2 to 1470 bytes during ESP-IDF v5.5 development: https://github.com/espressif/esp-idf/commit/a4106ef3b94ae2c90dec0ed5364b2290966c7a6e (!!!), so have updated this PR to match.

Trying to get the multi_espnow tests to pass I fixed some other test issues, which I'll submit in another PR.

dpgeorge (approved):
Thanks @projectgus for updating this PR.

I tested using IDF 5.5.2 with ESP32_GENERIC and ESP32_GENERIC_S2, and I can get all `multi_espnow` tests to pass (most of the time).

Good to see docs and tests updated/added.  This is good to merge now.

PR #16738, opened 2025-02-12
github/workflows: Stop using ubuntu-20.04.

dpgeorge:
### Summary

For GitHub Actions, ubuntu-20.04 is deprecated and will be removed by 1st April 2025.

See announcement here: https://github.com/actions/runner-images/issues/11101

This PR changes existing actions that use ubuntu-20.04 to a newer image.

### Testing

CI will test this.  I assume there will be some problems that need to be individually addressed.

### DIFF
--- a/.github/workflows/mpy_format.yml
+++ b/.github/workflows/mpy_format.yml
@@ -15,7 +15,7 @@ concurrency:
 
 jobs:
   test:
-    runs-on: ubuntu-20.04 # use 20.04 to get python2
+    runs-on: ubuntu-22.04 # use 22.04 to get python2
     steps:
     - uses: actions/checkout@v4
     - name: Install packages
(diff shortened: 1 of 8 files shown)


dpgeorge:
@projectgus going to need your help here with the esp32 cached IDF, seems like it's not rebuilding the cache when the Python version changed.

projectgus:
> @projectgus going to need your help here with the esp32 cached IDF, seems like it's not rebuilding the cache when the Python version changed.

Right, I forgot that the install directory is based on both the IDF version and the Python version. I've pushed a commit here to change the cache key to match.

dpgeorge:
> I've pushed a commit here to change the cache key to match.

Excellent!  Thank you.

projectgus (approved):

PR #16760, opened 2025-02-16
esp32: Switch to IDF v5.4

IhorNehrutsa:
### Summary
Supported IDF v5.4 chips: ESP32, ESP32-S2, ESP32-C3, ESP32-S3, ESP32-C2, ESP32-C6, ESP32-H2, ESP32-P4

### Testing
ESP32 generic board.

Several PRs require IDF v5.4.

### DIFF
--- a/ports/esp32/README.md
+++ b/ports/esp32/README.md
@@ -31,7 +31,7 @@ manage the ESP32 microcontroller, as well as a way to manage the required
 build environment and toolchains needed to build the firmware.
 
 The ESP-IDF changes quickly and MicroPython only supports certain versions.
-Currently MicroPython supports v5.2, v5.2.2, v5.3 and v5.4.
+Currently MicroPython supports v5.2, v5.2.2, v5.3, v5.4 and v5.4.1.
 
 To install the ESP-IDF the full instructions can be found at the
 [Espressif Getting Started guide](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/index.html#installation-step-by-step).
@@ -49,10 +49,10 @@ The steps to take are summarised below.
 To check out a copy of the IDF use git clone:
 
 ```bash
-$ git clone -b v5.2.2 --recursive https://github.com/espressif/esp-idf.git
+$ git clone -b v5.4.1 --recursive https://github.com/espressif/esp-idf.git
 ```
 
-You can replace `v5.2.2` with any other supported version.
+You can replace `v5.4.1` with any other supported version.
 (You don't need a full recursive clone; see the `ci_esp32_setup` function in
 `tools/ci.sh` in this repository for more detailed set-up commands.)
 
@@ -61,7 +61,7 @@ MicroPython and update the submodules using:
 
 ```bash
 $ cd esp-idf
-$ git checkout v5.2.2
+$ git checkout v5.4.1
 $ git submodule update --init --recursive
 ```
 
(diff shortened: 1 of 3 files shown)


projectgus:
Now v1.25 is out, it's a good time to look at updating ESP-IDF. IDF v5.4.1 has been released since this PR was opened, I ran an updated `metrics_esp32.py` script (will push update in a separate PR):

| BOARD | BOARD_VARIANT | IDF Version | Binary Size | Static IRAM Size | Static DRAM Size |
|-------|---------------|-------------|-------------|------------------|------------------|
| ESP32_GENERIC |  | v5.2.2 | 1637369 | 116154 | 55252 |
|  |  | v5.4.1 | 1647921 | 117307 | 56296 |
| ESP32_GENERIC | D2WD | v5.2.2 | 1370213 | 108882 | 54348 |
|  |  | v5.4.1 | 1369281 | 111167 | 55992 |
| ESP32_GENERIC | SPIRAM | v5.2.2 | 1468941 | 112782 | 54756 |
|  |  | v5.4.1 | 1477780 | 116371 | 56424 |
| ESP32_GENERIC_S3 |  | v5.2.2 | 1604077 | 16383 | 136643 |
|  |  | v5.4.1 | 1633206 | 16383 | 143184 |
| ESP32_GENERIC_S3 | SPIRAM_OCT | v5.2.2 | 1607689 | 16383 | 140251 |
|  |  | v5.4.1 | 1636634 | 16383 | 146556 |

About 1KB more IRAM and 1KB more DRAM on most boards, not great. But 6.5KB more DRAM usage on the S3! We should look at whether we can claim some RAM savings back before switching over.

> Testing
> ESP32 generic board.

@IhorNehrutsa can you be more specific please? Did you only check that it boots, did you check WiFi? Bluetooth? The unit test suite? Anything else?

projectgus:
> About 1KB more IRAM and 1KB more DRAM on most boards, not great. But 6.5KB more DRAM usage on the S3! We should look at whether we can claim some RAM savings back before switching over.

I had a look through the idf_size summary and didn't find much, it seems like lots of small increases in many components. 

Although there is a BT controller option [BT_CTRL_RUN_IN_FLASH_ONLY](https://docs.espressif.com/projects/esp-idf/en/v5.2.5/esp32s3/api-reference/kconfig.html#config-bt-ctrl-run-in-flash-only) that's available in all supported versions and which saves around 20KB of IRAM (translates to free internal DRAM on ESP32-S3). It can be applied separately to any version update, though.

dpgeorge:
I have tested ESP32_GENERIC board on IDF 5.4.1.  Summary of test results:
- `extmod_hardware/machine_uart_irq_break.py` now passes
- `extmod/deflate_decompress.py` when run with `--via-mpy --emit native` now passes
- `thread/stress_aes.py` now fails, but only because it takes a tiny bit longer than the 10 second timeout and so gets killed before it can finish
- everything else is the same (including WiFi tests)

I also tested ESP32_GENERIC_S2 firmware on a UM FeatherS2.  I'm getting random crashes when running the standard test suite :disappointed: .  But that might actually be unrelated to IDF 5.4.1.  Needs further investigation.

dpgeorge:
> I also tested ESP32_GENERIC_S2 firmware on a UM FeatherS2. I'm getting random crashes when running the standard test suite 😞 . But that might actually be unrelated to IDF 5.4.1. Needs further investigation.

See #17287 for an explanation of this crash, and a fix.

dpgeorge (approved):
Updated PR to use IDF 5.4.1, and tested on S2 and C3.  Everything seems to work.

dpgeorge:
Now merged!

IDF 5.4.1 is now the officially supported version and used to build firmware on the download page from now on.

PR #16764, opened 2025-02-16
docs/library/neopixel.rst: Update neopixel docs to hide dunders.

tobes:
### Summary

Change documentation to show using `len(np)` to access number of pixels rather than `NeoPixel.__len__()`.  Also use update for setting and getting items via `np[index]`.

I was looking at the NeoPixel documentation and saw that it was showing dunder methods for accessing number of pixels and also for setting/reading pixels ie `NeoPixel.__len__()`, `NeoPixel.__setitem__()`, `NeoPixel.__getitem__()`  rather than using `len(np)`, `np[index]`.

It is good practice especially for beginners that we should be showing correct access of these items rather than encouraging people to use private methods.  

I used the python documentation https://docs.python.org/3/library/stdtypes.html#mapping-types-dict as a basis for this update.

I chose `np` as a place holder for a NeoPixel instance as it felt appropriate, additionally I updated the code example to use `np` instead if `n` so that there was consistency accross the documentation. 

I think the Methods sections should also be combined but I have not included that change here but can if you think that is the way to go.

### Testing

This is a trivial documentation change so no testing was done

### DIFF
--- a/docs/library/neopixel.rst
+++ b/docs/library/neopixel.rst
@@ -24,14 +24,14 @@ For example::
 
     # 32 LED strip connected to X8.
     p = machine.Pin.board.X8
-    n = neopixel.NeoPixel(p, 32)
+    np = neopixel.NeoPixel(p, 32)
 
     # Draw a red gradient.
     for i in range(32):
-        n[i] = (i * 8, 0, 0)
+        np[i] = (i * 8, 0, 0)
 
     # Update the strip.
-    n.write()
+    np.write()
 
 Constructors
 ------------
@@ -54,17 +54,17 @@ Pixel access methods
     Sets the value of all pixels to the specified *pixel* value (i.e. an
     RGB/RGBW tuple).
 
-.. method:: NeoPixel.__len__()
+.. method:: len(np)
 
-    Returns the number of LEDs in the strip.
+    Returns the number of LEDs in the strip for NeoPixel *np*.
 
-.. method:: NeoPixel.__setitem__(index, val)
+.. method:: np[index]
 
-    Set the pixel at *index* to the value, which is an RGB/RGBW tuple.
+    Returns the pixel of NeoPixel *np* at *index* as an RGB/RGBW tuple.
 
-.. method:: NeoPixel.__getitem__(index)
+.. method:: np[index] =  value
 
-    Returns the pixel at *index* as an RGB/RGBW tuple.
+    Set the pixel of NeoPixel *np* at *index* to the value, which is an RGB/RGBW tuple.
 
 Output methods
 --------------


docs/library/neopixel.rst
```diff
@@ -54,17 +54,17 @@ Pixel access methods
     Sets the value of all pixels to the specified *pixel* value (i.e. an
     RGB/RGBW tuple).
 
-.. method:: NeoPixel.__len__()
+.. method:: len(np)
```
Josverl, line 57: There is no method `Neopixel.len`

docs/library/neopixel.rst
```diff
@@ -54,17 +54,17 @@ Pixel access methods
...
 
-    Returns the number of LEDs in the strip.
+    Returns the number of LEDs in the strip for NeoPixel *np*.
 
-.. method:: NeoPixel.__setitem__(index, val)
+.. method:: np[index]
```
Josverl, line 61: There is no method `Neopixel.np`

Josverl (changes requested):

docs/library/neopixel.rst
tobes: I realise it's not a method but the `.. method::` directive is as much about formatting as anything else.  What would you suggest.  As I say I was using pythons documentation as a guide.

Josverl: I recommend to leave the docs unchanged - as changing them will likely break the hyperlinks in the documentation , and surely break the generation of the micropython-stubs, which actually parse the directives. 
Please see [my comment](https://github.com/micropython/micropython/pull/16764#issuecomment-2661558578) for recommendations

tobes:
> The definition and documentation of the dunder methods should not be removed.

I'm not aware of dunder methods being documented as 'real' methods in any python documentation. They aren't public methods as such because of how you access them.  To me it is like saying `NeoPixel.__init__()` should be documented - it doesn't really make sense.

Part of the python philosophy is to respect private methods and imho the current documentation goes against that.

Josverl:
The NeoPixel class has a almost unique combination of runtime properties that AFAIK are not captures as a Protocol nor as a Base Class 
Removing these would hide these capabilities.

Als, Python **does** document dunder methods, just a few examples below.
- [1](https://docs.python.org/3/library/enum.html#enum.Enum.__init__)
- [2](https://docs.python.org/3/library/email.message.html#email.message.EmailMessage.__len__)

tobes:
Ok, but there seem to be very few dunder methods and mostly around email so it still feels fairly niche.  Eg very few `__len__()` methods are documented considering how many objects support `len()`

I don't know enough about the NeoPixel code but it still feels odd that the natural way to use an instance isn't documented apart from  one line the code example, which might confuse some users as there is no explanation.

I still feel that these ways to access the object should be documented for users less familiar with python.

dpgeorge:
Thanks @tobes for the patch.  As @Josverl says, I think it's best to keep the method names as they are (they are actually public methods, not private).  Instead, if you want to improve these neopixel docs, I suggest adding examples in the existing methods on how to use `len()` and subscript access.

I'll close the PR for now, but feel free to reopen if you want to keep working on it.

PR #16765, opened 2025-02-16
esp8266/network_wlan: Allow enumerating connected stations in AP mode.

### LINKED ISSUE #5395: List mac addresses connected to ESP8266 via AP_IF mode
Similar to [#186](https://github.com/micropython/micropython-esp32/issues/186), would it be possible to list the devices that are connected to an ESP8266 broadcasting a wireless access point?
```
import network
ap = network.WLAN(network.AP_IF)
ap.active(True)
ap.config(essid='mode_ap')
ap.status('stations')
```

agatti:
### Summary

This commit introduces the ability to obtain a list of stations connected to the device when in soft-AP mode.

A new parameter ("stations") to pass to WLAN.status is supported, returning a tuple of (bssid, ipv4) entries, one per connected station. An empty tuple is returned if no stations are connected, and an exception is raised if an error occurred whilst building the python objects to return to the interpreter.

Documentation is also updated to cover the new parameter.

This fixes #5395.

### Testing

I've tested this code as part of a project of mine on a NodeMCU with three stations connected to it, with the expected output being returned by calling `interface.status("stations")`.  I doubt this can be put in the CI pipeline or how this can have a test of its own.

This comes with a small footprint increase for the general firmware configuration, however if this is unacceptable maybe I can add a port define that disables AP support in MicroPython.

Also, no idea on how `wifi_softap_get_station_info` behaves with a large number of stations, all possible recoverable error conditions should be handled in this PR.

### DIFF
--- a/docs/library/network.WLAN.rst
+++ b/docs/library/network.WLAN.rst
@@ -85,7 +85,18 @@ Methods
         * ``STAT_GOT_IP`` -- connection successful.
 
     When called with one argument *param* should be a string naming the status
-    parameter to retrieve.  Supported parameters in WiFI STA mode are: ``'rssi'``.
+    parameter to retrieve, and different parameters are supported depending on the
+    mode the WiFi is in.
+
+    In STA mode, passing ``'rssi'`` returns a signal strength indicator value, whose
+    format varies depending on the port (this is available on all ports that support
+    WiFi network interfaces, except for CC3200).
+
+    In AP mode, passing ``'stations'`` returns a list of connected WiFi stations
+    (this is available on all ports that support WiFi network interfaces, except for
+    CC3200).  The format of the station information entries varies across ports,
+    providing either the raw BSSID of the connected station, the IP address of the
+    connected station, or both.
 
 .. method:: WLAN.isconnected()
 
--- a/ports/esp8266/network_wlan.c
+++ b/ports/esp8266/network_wlan.c
@@ -170,6 +170,29 @@ static mp_obj_t esp_disconnect(mp_obj_t self_in) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_1(esp_disconnect_obj, esp_disconnect);
 
+static void build_stations_list_cleanup_callback(void *station_info) {
+    if (station_info != NULL) {
+        wifi_softap_free_station_info();
+    }
+}
+
+static mp_obj_t build_stations_list(void) {
+    struct station_info *info = wifi_softap_get_station_info();
+    MP_DEFINE_NLR_JUMP_CALLBACK_FUNCTION_1(ctx, build_stations_list_cleanup_callback, (void *)info);
+    nlr_push_jump_callback(&ctx.callback, mp_call_function_1_from_nlr_jump_callback);
+    mp_obj_list_t *stations = MP_OBJ_TO_PTR(mp_obj_new_list(0, NULL));
+    struct station_info *current = info;
+    mp_obj_t entry[2] = { 0 };
+    while (current != NULL) {
+        entry[0] = mp_obj_new_bytes(current->bssid, sizeof(current->bssid));
+        entry[1] = netutils_format_ipv4_addr((uint8_t *)&current->ip.addr, NETUTILS_BIG);
+        mp_obj_list_append(stations, mp_obj_new_tuple(2, entry));
+        current = STAILQ_NEXT(current, next);
+    }
+    nlr_pop_jump_callback(true);
+    return MP_OBJ_FROM_PTR(stations);
+}
+
 static mp_obj_t esp_status(size_t n_args, const mp_obj_t *args) {
     wlan_if_obj_t *self = MP_OBJ_TO_PTR(args[0]);
     if (n_args == 1) {
@@ -185,6 +208,12 @@ static mp_obj_t esp_status(size_t n_args, const mp_obj_t *args) {
                 if (self->if_id == STATION_IF) {
                     return MP_OBJ_NEW_SMALL_INT(wifi_station_get_rssi());
                 }
+                break;
+            case MP_QSTR_stations:
+                if (self->if_id == SOFTAP_IF) {
+                    return build_stations_list();
+                }
+                break;
         }
         mp_raise_ValueError(MP_ERROR_TEXT("unknown status param"));
     }


dpgeorge:
Thanks for this addition, it's good because a few other WLAN interfaces also support `ap.status("stations")`, namely cyw43 (stm32 and rp2), esp-hosted, ninaw10 and esp32.

Unfortunately those existing implementations don't really follow any standard about what they return!  Some return a a list of 1-tuples of MAC, some return a list of IPs.

We need to standardise on something and document it.  What you have here, namely a list of 2-tuples of MAC and IP, seems like a good choice.  The thing with a tuple is that it could be extended in the future to add new entries, and that's backwards compatible.

But the first entry in the tuple should just be a bytes object of the MAC.  You can easily hexlify this in Python via `.hex(":")`.

ports/esp8266/network_wlan.c
```diff
@@ -170,6 +170,38 @@ static mp_obj_t esp_disconnect(mp_obj_t self_in) {
...
+    nlr_buf_t nlr;
+    if (nlr_push(&nlr) == 0) {
+        stations = MP_OBJ_TO_PTR(mp_obj_new_tuple(count, NULL));
+        struct station_info *current = info;
+        size_t station_index = 0;
+        mp_obj_t entry[2] = { 0 };
+        char bssid_str[18] = { 0 };
+        while (current != NULL && station_index < count) {
+            mp_uint_t bssid_len = snprintf(bssid_str, 18, "%02X:%02X:%02X:%02X:%02X:%02X", current->bssid[0], current->bssid[1], current->bssid[2], current->bssid[3], current->bssid[4], current->bssid[5]);
+            entry[0] = mp_obj_new_str(bssid_str, bssid_len);
```
dpgeorge, line 192: Suggest that this entry just be a bytes object of the 6-byte MAC.

docs/library/network.WLAN.rst
```diff
@@ -85,7 +85,10 @@ Methods
         * ``STAT_GOT_IP`` -- connection successful.
 
     When called with one argument *param* should be a string naming the status
-    parameter to retrieve.  Supported parameters in WiFI STA mode are: ``'rssi'``.
+    parameter to retrieve.  Supported parameters in WiFi STA mode are: ``'rssi'``.
+    In WiFi AP mode on ESP8266 the ``'stations'`` parameter is also supported,
```
dpgeorge, line 89: This needs to be updated to state that more ports support `stations`.  Maybe something like "some network interfaces in AP mode support the `stations` parameter which returns a list of connected STAs ...".

ports/esp8266/network_wlan.c
```diff
@@ -170,6 +170,38 @@ static mp_obj_t esp_disconnect(mp_obj_t self_in) {
...
+    if (count == 0) {
+        return mp_const_empty_tuple;
+    }
+    mp_obj_tuple_t *stations = NULL;
+    struct station_info *info = wifi_softap_get_station_info();
+    if (info == NULL) {
+        mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("no station info available"));
+    }
+    nlr_buf_t nlr;
+    if (nlr_push(&nlr) == 0) {
```
dpgeorge, line 184: There's no need to wrap this in NLR.  Just let any exception propagate out.

ports/esp8266/network_wlan.c
```diff
@@ -170,6 +170,38 @@ static mp_obj_t esp_disconnect(mp_obj_t self_in) {
...
 static MP_DEFINE_CONST_FUN_OBJ_1(esp_disconnect_obj, esp_disconnect);
 
+static mp_obj_t build_stations_tuple(void) {
+    mp_uint_t count = wifi_softap_get_station_num();
+    if (count == 0) {
+        return mp_const_empty_tuple;
+    }
+    mp_obj_tuple_t *stations = NULL;
+    struct station_info *info = wifi_softap_get_station_info();
+    if (info == NULL) {
```
dpgeorge, line 180: Can you omit this check, and then the loop below just won't execute any iterations.  Then it'll return an empty list.  I think that's fine?

agatti: I remember I looked at the disassembly for `wifi_softap_free_station_info` and I recall that function not making any checks before deallocating memory that was meant to be allocated in `wifi_softap_get_station_info`.  

If I understood the disassembled code correctly then the NULL check must still be there, as if `wifi_softap_get_station_info` failed to allocate memory then who knows what happens later.  I'm not 100% sure on this, though, as I don't know Xtensa assembler as much as RISC-V so I might have missed something there.

docs/library/network.WLAN.rst
agatti: I've expanded the documentation to add the list of ports involved, also for the `rssi` parameter in STA mode - as I split the text block into separate STA/AP paragraphs.

ports/esp8266/network_wlan.c
agatti: This is addressed in the latest version of the PR, thanks for pointing this out!

ports/esp8266/network_wlan.c
dpgeorge: OK.  My main point here is that instead of raising an exception if `info` is `NULL`, just return an empty list.  That will lead to smaller code.

ports/esp8266/network_wlan.c
dpgeorge: Ah, OK, so you need to clean up things if there's an exception.

But in that case you should still re-raise the exception, instead of printing it out.

You might be interested to look at `nlr_push_jump_callback()`, which is a relatively new feature to support this kind of use case.

ports/esp8266/network_wlan.c
```diff
@@ -170,6 +170,37 @@ static mp_obj_t esp_disconnect(mp_obj_t self_in) {
...
+            stations->items[station_index++] = mp_obj_new_tuple(2, entry);
+            current = STAILQ_NEXT(current, next);
+        }
+        nlr_pop();
+    } else {
+        mp_obj_print_exception(&mp_plat_print, MP_OBJ_FROM_PTR(nlr.ret_val));
+        stations = NULL;
+    }
+    wifi_softap_free_station_info();
+    return stations != NULL ? MP_OBJ_FROM_PTR(stations) : MP_OBJ_NULL;
```
dpgeorge, line 201: You can't return `MP_OBJ_NULL` from a method, that will crash the runtime.

agatti:
> We need to standardise on something and document it. What you have here, namely a list of 2-tuples of MAC and IP, seems like a good choice. The thing with a tuple is that it could be extended in the future to add new entries, and that's backwards compatible.

That makes sense, thanks for letting me know.  I admit I haven't looked at what other ports did as I thought that different ports had different WiFi capabilities exposed to user code.

I'll look at other ports' code later today to see what information is available for connected stations, and see if a followup PR can be submitted to align the ports on that regard.  Those changes, if made, will be put behind `MICROPY_PREVIEW_VERSION_2` of course.

ports/esp8266/network_wlan.c
agatti: That indeed looks like it is what I needed there, thanks!  It also simplified the code by quite a bit too.

I've extended callback support to also accept functions with no arguments, but if the (tiny) size increase in the core runtime is not wanted I can always wrap the call to `wifi_softap_free_station_info` in a function that accepts (and discards) an argument.

ports/esp8266/network_wlan.c
agatti: I remember doing that in the past for other code of mine and it worked, but I don't recall whether it worked by accident or not (I recall doing something with `None` return values, but it's been a while).

Anyway, using the NLR callback made this bit of code disappear :)

ports/esp8266/network_wlan.c
agatti: Well, my point was that `wifi_softap_get_station_info` returning NULL even though `wifi_softap_get_station_num` returned a non-zero value is an exceptional condition and should be treated as such.

Anyway, I've modified the code as you suggested, for the sake of a smaller footprint :)

py/runtime.c
```diff
@@ -202,6 +202,11 @@ void mp_globals_locals_set_from_nlr_jump_callback(void *ctx_in) {
     mp_locals_set(ctx->locals);
 }
 
+void mp_call_function_0_from_nlr_jump_callback(void *ctx_in) {
+    nlr_jump_callback_node_call_function_0_t *ctx = ctx_in;
+    ctx->func();
```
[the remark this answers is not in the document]
agatti: That's fine, I didn't know there was a native test suite.  I've wrapped the callback to keep all changes within the port's source tree - having a user variable bound to a callback context is a common use case after all.

ports/esp8266/network_wlan.c
```diff
@@ -170,6 +170,31 @@ static mp_obj_t esp_disconnect(mp_obj_t self_in) {
...
+    if (count == 0) {
+        return mp_const_empty_tuple;
+    }
+    struct station_info *info = wifi_softap_get_station_info();
+    if (info == NULL) {
+        return mp_const_empty_tuple;
+    }
+    MP_DEFINE_NLR_JUMP_CALLBACK_FUNCTION_0(ctx, wifi_softap_free_station_info);
+    nlr_push_jump_callback(&ctx.callback, mp_call_function_0_from_nlr_jump_callback);
+    mp_obj_tuple_t *stations = MP_OBJ_TO_PTR(mp_obj_new_tuple(count, NULL));
```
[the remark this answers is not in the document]
agatti: No problem, that's been addressed in the latest revision of the PR.

docs/library/network.WLAN.rst
```diff
@@ -85,7 +85,17 @@ Methods
         * ``STAT_GOT_IP`` -- connection successful.
 
     When called with one argument *param* should be a string naming the status
-    parameter to retrieve.  Supported parameters in WiFI STA mode are: ``'rssi'``.
+    parameter to retrieve, and different parameters are supported depending on the
+    mode the WiFi is in.
+
+    In STA mode, passing ``'rssi'`` returns a signal strength indicator value, whose
+    format varies depending on the port (currently supported on ESP32 and ESP8266).
```
dpgeorge, line 92: rssi is actually supported on all WLAN interfaces except the cc3200 port

docs/library/network.WLAN.rst
```diff
@@ -85,7 +85,17 @@ Methods
...
+
+    In AP mode, passing ``'stations'`` returns a list of connected WiFi stations
+    (currently supported on ports/boards that use the CYW43 WiFi driver, and the
+    ESP32 and ESP8266 ports).  The format of the station information entries varies
```
dpgeorge, line 96: with this PR, stations is supported the same as rssi, on everything except cc3200

ports/esp8266/network_wlan.c
```diff
@@ -170,6 +170,36 @@ static mp_obj_t esp_disconnect(mp_obj_t self_in) {
...
+    if (info == NULL) {
+        return mp_obj_new_list(0, NULL);
+    }
+    MP_DEFINE_NLR_JUMP_CALLBACK_FUNCTION_1(ctx, build_stations_list_cleanup_callback, NULL);
+    nlr_push_jump_callback(&ctx.callback, mp_call_function_1_from_nlr_jump_callback);
+    mp_obj_list_t *stations = MP_OBJ_TO_PTR(mp_obj_new_list(count, NULL));
+    struct station_info *current = info;
+    size_t station_index = 0;
+    mp_obj_t entry[2] = { 0 };
+    while (current != NULL && station_index < count) {
```
dpgeorge, line 193: Do you need to check both `current` and `station_index < count`?  If the loop exists early because `current` is `NULL` then the returned list will be short, and entries at the end will be NULL and lead to crashes in the runtime.

You could either not test for `current` being NULL, or use `mp_obj_list_append()`.

If the latter, you could possibly omit the checks for `count == 0` and `info == NULL` above (and not need to create and return an empty list), use `mp_obj_new_list(0, NULL)` to create the initial list, append to it, and then pass `info` through as the argument to the NLR callback and test for `info == NULL` in `build_stations_list_cleanup_callback`.

docs/library/network.WLAN.rst
agatti: Interesting, I've searched for `MP_QSTR_rssi` and only found mentions on ESP32 and ESP8266.  `rssi` appeared on CC3200 but from a quick look it didn't seem like it was handled, I guess I got it wrong.  I'll update the documentation for this and `stations` below.  Thanks!

ports/esp8266/network_wlan.c
dpgeorge: I'd say it's not a concern.  If someone finds that it crashes, we can add a NULL check.

agatti: OK, I pushed the (hopefully) final changes to the PR.  Thanks for the code suggestions, this looks much better now indeed.

dpgeorge (approved):
Looks good, thanks for updating.

I tested on an esp8266 and it works (after having issues with the esp8266 disassociating the STA, I think due to power saving timing problems).

PR #16207, opened 2024-11-10
zephyr: Upgrade to Zephyr v4.0.0.

MaureenHelm:
### Summary

Updates the Zephyr port to the latest Zephyr release tag. ~~It's currently a release candidate, not a final release, therefore this PR is in draft status until the Zephyr release is finalized.~~ [Zephyr v4.0.0](https://github.com/zephyrproject-rtos/zephyr/releases/tag/v4.0.0) was released on 15 Nov 2024.

### Testing

Tested on max32690fthr and frdm_k64f boards.

PR #16770, opened 2025-02-17
Fixes for zephyr 4.0

laodzu:
### Summary

Some fixes for Zephyr port for v4.0 and beyond.  Without some of those changes, neither the current version v4.0 nor the current development version even compile.  The commits clearly explain the reason for the change with links to the original commits.  The patches were tested with an SD card and a QSPI flash system.

### Testing

I tested the patches on all currently supported Zephyr versions, i.e. v3.7, v4.0 and top-of-tree v4.0.0-5122-gcb772b0f60c on two boards:
 - mimxrt1060_evk
 - frdm_mcxn947

On the mimxrt1060 I tested the SD support by injecting -DCONFIG_DISK_ACCESS=y into the build.  I did not change the default configuration as this hangs when no SD card is discovered and so people should only enable it if they want to use such a system.

For the frdm_mcxn947 I added a device tree overlay and a config fragment to enable the functionality out of the box as the flash is always present, and I did not see any slowdown with the functionality.

I blogged the whole journey on my blog:

https://blog.lazy-evaluation.net/posts/embedded/micropython-zephyr-imxrt1060.html
https://blog.lazy-evaluation.net/posts/embedded/micropython-zephyr-mcxn947.html


dpgeorge:
> Interesting. One day nothing happens and suddenly a "This branch must not contain merge commits" error/warning shows up

It looks like that error is gone now?  GitHub has recently updated the "merge experience" so maybe some teething issues which have been resolved.

Anyway, things look good now.

laodzu:
Hi Damien,

thanks fir looking into this!

> It looks like that error is gone now? GitHub has recently updated the "merge experience" so maybe some teething issues which have been resolved.

I still do see a red "Merging is blocked" warning/error, but I have no idea if this prevents anything.

> Anyway, things look good now.

That sounds good! I am looking forward to feedback on the series.

Best wishes
  Detlev

ports/zephyr/main.c
```diff
@@ -97,11 +98,17 @@ static void vfs_init(void) {
...
+    mp_obj_t args[] = { mp_obj_new_str_from_cstr(DT_INST_PROP(0, disk_name)) };
+    #else
     mp_obj_t args[] = { mp_obj_new_str_from_cstr(CONFIG_SDMMC_VOLUME_NAME) };
+    #endif
     bdev = MP_OBJ_TYPE_GET_SLOT(&zephyr_disk_access_type, make_new)(&zephyr_disk_access_type, ARRAY_SIZE(args), 0, args);
     mount_point_str = "/sd";
     #elif defined(CONFIG_FLASH_MAP) && FIXED_PARTITION_EXISTS(storage_partition)
-    mp_obj_t args[] = { MP_OBJ_NEW_SMALL_INT(FIXED_PARTITION_ID(storage_partition)), MP_OBJ_NEW_SMALL_INT(4096) };
+    #define DT_DRV_COMPAT zephyr_flash_disk
+    mp_obj_t args[] = { MP_OBJ_NEW_SMALL_INT(FIXED_PARTITION_ID(storage_partition)), MP_OBJ_NEW_SMALL_INT(DT_INST_PROP(0, cache_size)) };
```
dpgeorge, line 111: This change (to retrieve `cache-size` from the device tree) now requires all boards that use this bit of code to have this defined.  And that seems to be something you have to now manually do.

Eg, with this PR the existing `nucleo_h743zi` board fails to build (using Zephyr v3.7.0):
```
zephyr/include/zephyr/devicetree.h:336:40: error: 'DT_N_INST_0_zephyr_flash_disk_P_cache_size' undeclared (first use in this function)
```

laodzu: Yes, you are right.  I do not have an overview of how many boards are affected, but I see that this is a problem.  Would it be ok, to check if such a property exists and if not use the hard coded 4 Kib instead?  I'll try to rework that patch in this way.

dpgeorge: > Would it be ok, to check if such a property exists and if not use the hard coded 4 Kib instead?

Yes, if that's possible, that would be good.

laodzu: I just force-pushed an update with this modification.  In order to let people know that the device tree should be updated, I inserted a corresponding #warning. I hope this is ok?

Also, I rewrote the device tree access not to use the #DRV_COMPAT macro that I used previously. I think things should be self-contained, even if that takes longer to express in code.

I am having a lot of problems getting the code formatting right :facepalm:

MaureenHelm:
See #16207

laodzu:
Thanks, Maureen, that explains a few things :smile:.  Let me comment on possible ways to join forces here. So if I see things correctly, then #16207 was not merged because that represents an either v3.7 or v4.0 path for MicroPython. This PR supports both versions (and main as of recent) and so the question is probably what stance MicroPython will take with regard to the required Zephyr version.  For this patch set, my reasoning was that the currently supported versions of Zephyr are v3.7 (LTS) and v4.0, so MicroPython should work with both - if possible.  For the current patch set, this worked, but I am not sure if this will always be possible. #16207 is better with respect to thread_analyzer. I am not so sure if MicroPython will run on multiple cores in the near future, but on the other hand doing it the right way is not a lot more work, so I would prefer #16207. Lastly this PR includes the addition for reading the erase block size for flash_map which probably should also go in (it is required when switching between the internal flash on mcxn947 with an erase block size of 8192 and the external QSPI flash).
So the question is if we want to support v3.7 _and_ v4.0 or v4.0 only.  In the latter case, this PR is mostly moot, but in the first case I would volunteer to try and merge the best of the two together and propose as another PR.
What do you think?

ThinkTransit:
Nice work, I'm happy to help with testing as I have recently started doing some work on getting power management working in zephyr for nrf [see here](https://github.com/micropython/micropython/pull/16864). I have a nrf5340DK and nrf9151DK for testing.

dpgeorge:
Thanks @laodzu for working on this, much appreciated!  Now that #16879 is merged, I'll close this PR as superseded.

Issue #16759, opened 2025-02-15
Asyncio task scheduling issue when using run_until_completed


yoctopuce:
### Port, board and/or hardware

any port with asyncio support

### MicroPython version

micropython-1.25.0 running on win32 [MSC v.1916 64 bit]

### Reproduction

The issue is linked to the fact that `run_until_complete` does not complete the processing of the _awaited_ task before returning to hte caller. If other pending tasks are waiting it, they will be lost forever.

The code below will reproduce the issue.
- When the tasks are await-ed form an async function, everything works as expected (function `async_test`)
- When the tasks are started from a sync function using run_until_complete, scheduling is broken  (function `sync_test`)
This issue is specific to micropython asyncio implementation of `run_until_complete` and does not occur in CPython.

```py
import sys, asyncio, random

def get_event_loop():
    if sys.implementation.name == "micropython":
        return asyncio.get_event_loop()
    else:
        try:
            eventloop = asyncio.get_running_loop()
        except RuntimeError:
            eventloop = asyncio.new_event_loop()
        return eventloop

class Worker:
    def __init__(self):
        self._eventLoop = None
        self._tasks = []

    def launchTask(self, asyncJob):
        if self._eventLoop is None:
            self._eventLoop = get_event_loop()
        return self._eventLoop.create_task(asyncJob)

    async def job(self, prerequisite, taskName):
        if prerequisite:
            await prerequisite
        await asyncio.sleep(2)
        print(taskName, "work completed")

    def planTasks(self):
        self._tasks.append(self.launchTask(self.job(None, 'task0')))
        self._tasks.append(self.launchTask(self.job(self._tasks[0], 'task1')))
        self._tasks.append(self.launchTask(self.job(self._tasks[1], 'task2')))

    async def waitForTask(self, taskIdx):
        return await self._tasks[taskIdx]

    def syncWaitForTask(self, taskIdx):
        return self._eventLoop.run_until_complete(self._tasks[taskIdx])

async def async_test():
    print("--- Running the async test")
    worker = Worker()
    worker.planTasks()
    await worker.waitForTask(0)
    print("-> task0 done")
    await worker.waitForTask(2)
    print("-> task2 done")

def sync_test():
    print("--- Running the sync test")
    worker = Worker()
    worker.planTasks()
    worker.syncWaitForTask(0)
    print("-> task0 done")
    worker.syncWaitForTask(2)
    print("-> task2 done")

asyncio.run(async_test())
sync_test()
```

[...]

Issue #16569, opened 2025-01-11
asyncio terminates unexpectedly


peterhinch:
### Port, board and/or hardware

Unix build (also verified on RP2 2)

### MicroPython version

MicroPython v1.24.1 on 2024-11-29; Raspberry Pi Pico2 with RP2350

### Reproduction

Paste the following at the REPL
```py
import asyncio

evt = asyncio.Event()
async def foo():
    data = [1,2,3]
    for d in data:
        print(d)
        await asyncio.sleep(1)
    await evt.wait()  # Should hang

async def main():
    await foo()

asyncio.run(main())
```

### Expected behaviour

Should print 1, 2, 3 then hang. This happens under CPython 3.10.12
```
1
2
3
```

### Observed behaviour

Under MP, after expected output, it returns to the REPL.
```
1
2
3
>>> 
```

### Additional Information

This may be related to https://github.com/micropython/micropython/issues/16318

I am encountering the phenomenon of premature return in a number of situations, notably in test scripts where there are no pending tasks. I have never observed it in applications where there are multiple concurrent tasks. 

### Code of Conduct

Yes, I agree

peterhinch:
To add a few observations.

The problem goes away if a second task is launched that simply does a repeated async sleep. This explains why the fault only occurs in simple scripts.

When the fault occurs the `run_until_complete` loop quits [here](https://github.com/micropython/micropython/blob/bfb1bee6feba9f06452a0e4b572ec4d15df325cf/extmod/asyncio/core.py#L168). It seems that while paused on
```py
    await evt.wait()  # Should hang
```
there is no `cur_task`, causing `run_until_complete` to quit prematurely.

While the example script is artificial I've encountered this bug a few times in simple scripts intended to test applications. The work-round is trivial, but the tendency for premature exit may confuse beginners.

Baker68:
Also present on MicroPython v1.24.0 on ESP32

dpgeorge:
Fixed by bdb7e036d2400c25beeb2b4252ff51c5255c4020

Issue #16318, opened 2024-11-28
asyncio: Asynchronous iterator terminates unexpectedly


peterhinch:
### Port, board and/or hardware

Unix build (also verified on RP2)

### MicroPython version

MicroPython v1.24.0-preview.106.g55e75c4ad on 2024-07-15; linux [GCC 11.4.0] version

### Reproduction

Run the follwing:
```py
import asyncio

async def test(ai):
    for x in range(3):
        await asyncio.sleep(1)
        ai.fetch_data(f"bar {x}")

class AsyncIterable:
    def __init__(self):
        self.message = None
        self.evt = asyncio.Event()

    def __aiter__(self):
        return self

    async def __anext__(self):
        await self.evt.wait()
        self.evt.clear()
        return self.message

    def fetch_data(self, message):
        self.message = message
        self.evt.set()

async def main():
    ai = AsyncIterable()
    asyncio.create_task(test(ai))
    async for message in ai:
        print(message)  # Why does this terminate?

asyncio.run(main())
```

### Expected behaviour

The following occurs under CPython
```
$ python3
Python 3.10.12 (main, Nov  6 2024, 20:22:13) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import rats
bar 0
bar 1
bar 2
< hangs here awaiting further data>
```

### Observed behaviour

Under MicroPython the code terminates:
```
$ upython 
MicroPython v1.24.0-preview.106.g55e75c4ad on 2024-07-15; linux [GCC 11.4.0] version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> import rats
bar 0
bar 1
bar 2
>>> 
```

### Additional Information

Termination occurs when `test()` terminates. Proved by adding `await asyncio.sleep(4)` at the end of the coroutine.

### Code of Conduct

Yes, I agree

Issue #16589, opened 2025-01-15
RMT receive


Zibri:
### Description

I just found out that microphthon can send RMT but not receive. That's a shame which forces me to use nodemcu while I prefer microphthon by a long shot. RMT receive is very important and without it, receiving pulses is way more inaccurate.
In my project I need to determine the pulse length accurately. Please implement RMT receive!

### Code Size

Should not be so big. But honestly I don't know.

### Implementation

I hope the MicroPython maintainers or community will implement this feature

### Code of Conduct

Yes, I agree

0wwafa:
The reason:

def test():
     s=time.ticks_cpu()
     time.sleep(1)
     print(time.ticks_diff(time.ticks_cpu(),s))
     
 the reults will have the overhead of the code which is around 4700 nanoseconds (clock 20-120) and more than 8000 at 240 mhz.
 
 The same happens if you sent an IRQ on a pin falling (or raising) edge.

RMT instead will return the time of each edge (or more couples) each time is called and has a resolution up to 12.5 nanoseconds!
It would be amazing to be able to use it without ditching micropython and coding in C.

elvis-epx:
Proposal in https://github.com/micropython/micropython/pull/18454

PR #16772, opened 2025-02-18
extmod/asyncio/core.py: Fix early exit of asyncio scheduler.

yoctopuce:
### Summary

This commit fixes three open issues related to the asyncio scheduler exiting prematurely when the main task queue is empty, in cases where CPython would not exit.
In the first case, the scheduler exits because running a task via run_until_complete did not schedule any dependent tasks.
In the other two cases, the scheduler exits because the tasks are queued in an event queue.
https://github.com/micropython/micropython/issues/16759
https://github.com/micropython/micropython/issues/16569 
https://github.com/micropython/micropython/issues/16318

### Testing

The fix has been tested on the Windows port compiled with asyncio support.
It is very unlikely to be board-specific.
All asyncio unit tests have been run, as well as several specific test cases related to the fixed issues.

This fix adds a few opcodes to the asyncio scheduler, but increases reliability and compatibility with CPython.

### DIFF
--- a/extmod/asyncio/core.py
+++ b/extmod/asyncio/core.py
@@ -163,9 +163,16 @@ def run_until_complete(main_task=None):
                 # A task waiting on _task_queue; "ph_key" is time to schedule task at
                 dt = max(0, ticks_diff(t.ph_key, ticks()))
             elif not _io_queue.map:
-                # No tasks can be woken so finished running
+                # No tasks can be woken
                 cur_task = None
-                return
+                if not main_task or not main_task.state:
+                    # no main_task, or main_task is done so finished running
+                    return
+                # At this point, there is theoretically nothing that could wake the
+                # scheduler, but it is not allowed to exit either. We keep the code
+                # running so that a hypothetical debugger (or other such meta-process)
+                # can get a view of what is happening and possibly abort.
+                dt = 3
             # print('(poll {})'.format(dt), len(_io_queue.map))
             _io_queue.wait_io_event(dt)
 
@@ -187,31 +194,33 @@ def run_until_complete(main_task=None):
         except excs_all as er:
             # Check the task is not on any event queue
             assert t.data is None
-            # This task is done, check if it's the main task and then loop should stop
-            if t is main_task:
+            # If it's the main task, it is considered as awaited by the caller
+            awaited = t is main_task
+            if awaited:
                 cur_task = None
-                if isinstance(er, StopIteration):
-                    return er.value
-                raise er
+                if not isinstance(er, StopIteration):
+                    t.state = False
+                    raise er
+                if t.state is None:
+                    t.state = False
             if t.state:
                 # Task was running but is now finished.
-                waiting = False
                 if t.state is True:
                     # "None" indicates that the task is complete and not await'ed on (yet).
-                    t.state = None
+                    t.state = False if awaited else None
                 elif callable(t.state):
                     # The task has a callback registered to be called on completion.
                     t.state(t, er)
                     t.state = False
-                    waiting = True
+                    awaited = True
                 else:
                     # Schedule any other tasks waiting on the completion of this task.
                     while t.state.peek():
                         _task_queue.push(t.state.pop())
-                        waiting = True
+                        awaited = True
                     # "False" indicates that the task is complete and has been await'ed on.
                     t.state = False
-                if not waiting and not isinstance(er, excs_stop):
+                if not awaited and not isinstance(er, excs_stop):
                     # An exception ended this detached task, so queue it for later
                     # execution to handle the uncaught exception if no other task retrieves
                     # the exception in the meantime (this is handled by Task.throw).
@@ -229,6 +238,9 @@ def run_until_complete(main_task=None):
                 _exc_context["exception"] = exc
                 _exc_context["future"] = t
                 Loop.call_exception_handler(_exc_context)
+            # If it's the main task then the loop should stop
+            if t is main_task:
+                return er.value
 
 
 # Create a new task from a coroutine and run it until it finishes
--- a/tests/extmod/asyncio_event_queue.py
+++ b/tests/extmod/asyncio_event_queue.py
@@ -1,0 +1,64 @@
+# Ensure that an asyncio task can wait on an Event when the
+# _task_queue is empty
+# https://github.com/micropython/micropython/issues/16569
+
+try:
+    import asyncio
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+# This test requires checking that the asyncio scheduler
+# remains active "indefinitely" when the task queue is empty.
+#
+# To check this, we need another independent scheduler that
+# can wait for a certain amount of time. So we have to
+# create one using micropython.schedule() and time.ticks_ms()
+#
+# Technically, this code breaks the rules, as it is clearly
+# documented that Event.set() should _NOT_ be called from a
+# schedule (soft IRQ) because in some cases, a race condition
+# can occur, resulting in a crash. However:
+# - since the risk of a race condition in that specific
+#   case has been analysed and excluded
+# - given that there is no other simple alternative to
+#   write this test case,
+# an exception to the rule was deemed acceptable. See
+# https://github.com/micropython/micropython/pull/16772
+
+import micropython, time
+
+try:
+    micropython.schedule
+except AttributeError:
+    print("SKIP")
+    raise SystemExit
+
+
+evt = asyncio.Event()
+
+
+def schedule_watchdog(end_ticks):
+    if time.ticks_diff(end_ticks, time.ticks_ms()) <= 0:
+        print("asyncio still pending, unlocking event")
+        # Caution: about to call Event.set() from a schedule
+        #          (see the note in the comment above)
+        evt.set()
+        return
+    micropython.schedule(schedule_watchdog, end_ticks)
+
+
+async def foo():
+    print("foo waiting")
+    schedule_watchdog(time.ticks_add(time.ticks_ms(), 100))
+    await evt.wait()
+    print("foo done")
+
+
+async def main():
+    print("main started")
+    await foo()
+    print("main done")
+
+
+asyncio.run(main())
--- a/tests/extmod/asyncio_event_queue.py.exp
+++ b/tests/extmod/asyncio_event_queue.py.exp
@@ -1,0 +1,5 @@
+main started
+foo waiting
+asyncio still pending, unlocking event
+foo done
+main done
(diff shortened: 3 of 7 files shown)


Josverl:
@yoctopuce , would it be possible to add tests to verify that the observed error cases do not occur any more ?

yoctopuce:
@Josverl : I have added the requested test cases. Interestingly, the new test case appear to cause issues in three specific environments, but I suspect this is caused by an improper setup of these test environments.
- webassembly port complains that `Loop.run_until_complete()` does not exist. However this should be available according to the version of extmod/asyncio/core.py present in master branch. Where does webassembly take its runtime asyncio package from ?
- the two coverage builds fail on two of the new tests, with a timeout error. Is there anything special in these build that could explaine why they would work differently from all other unix builds ?

dpgeorge:
> * webassembly port complains that `Loop.run_until_complete()` does not exist

The webassembly port has a customized version of `asyncio` that integrates with JavaScript `await` and allows (among other things) top-level awaitables.  You can just exclude the tests that don't work under this port using the list in `tests/run-tests.py:platform_tests_to_skip`.

> the two coverage builds fail on two of the new tests, with a timeout error. Is there anything special in these build that could explaine why they would work differently from all other unix builds ?

You can reproduce this failure locally, by running `make VARIANT=coverage test_full` in the `ports/unix` directory.  It fails when running the test with the native emitter, because when running native code any scheduled functions do not run.

------

Actually, it's not allowed to call `Event.set()` from a scheduled function (via `micropython.schedule`).  This is because `Event.set()` is not thread safe.  Eg if the main thread is executing `Event.set()` and is interrupted by the execution of a scheduled function, and that scheduled function calls the same `Event.set()`, then this could lead to an exception being raised:
```py
    def set(self):                       
        # Event becomes set, schedule any tasks waiting on it
        # Note: This must not be called from anything except the thread running
        # the asyncio loop (i.e. neither hard or soft IRQ, or a different thread).
        while self.waiting.peek():
   <-------- if a scheduled function executes here and runs the same `set()` method
             then it will exhaust the `self.waiting` queue and when the main thread
             resumes it will try to pop off an empty queue, leading to an exception

            core._task_queue.push(self.waiting.pop())                        
        self.state = True
```
You can see there's a warning there in the comments about not calling `Event.set()` from anything other than the main thread.  A scheduled function counts as a soft IRQ.

You instead need to use `ThreadSafeFlag`.  But if you do that the test will work without the fixes in this PR.

It should actually be impossible to write Python code that doesn't use any IRQs/scheduled code and exits early from the asyncio scheduler (apart from the simple case of waiting on something that will never be set, in that case it's a pure deadlock and the scheduler exits because there's not much point in waiting forever).

yoctopuce:
> The webassembly port has a customized version of `asyncio` that integrates with JavaScript `await` and allows (among other things) top-level awaitables. You can just exclude the tests that don't work under this port using the list in `tests/run-tests.py:platform_tests_to_skip`.

Ok, good to know. I will skip the new asyncio tests along with other skipped asyncio tests for webassembly.
This is all we need to close https://github.com/micropython/micropython/issues/16759

> You can see there's a warning there in the comments about not calling `Event.set()` from anything other than the main thread. A scheduled function counts as a soft IRQ.
> It should actually be impossible to write Python code that doesn't use any IRQs/scheduled code and exits early from the asyncio scheduler
> (apart from the simple case of waiting on something that will never be set, in that case it's a pure deadlock and the scheduler exits because there's not much point in waiting forever).

These problematic test cases are needed to cover the fix for the other two issues "fixed" by this PR
https://github.com/micropython/micropython/issues/16569
https://github.com/micropython/micropython/issues/16318
Both correspond precisely to that "pure deadlock" case: the main task is waiting for an Event that no one will ever trigger. MicroPython scheduler exits silently, while CPython stays and waits forever. This PR would Make MicroPython behave more like CPython, which also makes it possible to diagnose the code using a `sys.settrace` debugger.

So here are the options I see:

1. I can revert the part of this PR that solves these two issues, assuming we consider them as "work as designed". The PR would then only solve https://github.com/micropython/micropython/issues/16759
2. I can change the code to raise an exception when we reach this deadlock situation, to avoid the misleading silent exit, and provide a clear error message instead. It would not be fully CPython compliant, but from my point of view it would be better than the status quo.
3. I can keep the PR code as is if we want to mimic the CPython behaviour, but then we need to find a solution for the test cases:
  3a. If, for the sole purpose of these test cases, it were deemed acceptable to use a global flag in the scheduled code to ensure that there is only one call  to `Event.set()`, then this would be a way of ensuring that the test code works properly, even though it is *breaking the rules*. These two test cases should then be skipped on platforms where scheduled code is not available.
  3b. Or I could add a success-on-timeout mechanism in run-tests.py, but this is probably a bigger change
  3c. Or I could try to make a testcase that uses `sys.settrace`, but that might be an overkill
  3d. Or maybe these test cases should simply be dropped ?

@dpgeorge @peterhinch what is your preference ?

peterhinch:
The repro scripts in #16569 and #16318 are intentionally minimal; the bugs have shown up in real applications. The fact that the `Event` in the #16589 script can never occur is not a necessary condition for the bug: see the [EDIT] in https://github.com/micropython/micropython/issues/16569#issuecomment-2628937138.

My view is that premature exit should be fixed.

dpgeorge:
> My view is that premature exit should be fixed.

Yes, I agree, it's come up enough times and it would be good to fix it.

> 3\. I can keep the PR code as is if we want to mimic the CPython behaviour, but then we need to find a solution for the test cases:

I think this is a good option, and I would also go for 3a (keep the test as they are in this PR) but add a comment at the top of the two "illegal" tests that they are technically breaking the rules, but they are doing it in a way that still works correctly.

yoctopuce:
> I think this is a good option, and I would also go for 3a (keep the test as they are in this PR) but add a comment at the top of the two "illegal" tests that they are technically breaking the rules, but they are doing it in a way that still works correctly.

OK, so I will update the PR accordingly early next week with the amended test cases.
Thanks for the quick answers!

yoctopuce:
I have pushed the updated test code.

yoctopuce:
I have updated the test code, all checks have now passed.

extmod/asyncio/core.py
```diff
@@ -163,9 +163,12 @@ def run_until_complete(main_task=None):
                 # A task waiting on _task_queue; "ph_key" is time to schedule task at
                 dt = max(0, ticks_diff(t.ph_key, ticks()))
             elif not _io_queue.map:
-                # No tasks can be woken so finished running
+                # No tasks can be woken
                 cur_task = None
-                return
+                if not main_task or not main_task.data:
```
dpgeorge, line 168: Should `not main_task.data` perhaps be `not main_task.state`?  To mirror `Task.done()`.

extmod/asyncio/core.py
```diff
@@ -163,9 +163,12 @@ def run_until_complete(main_task=None):
...
+                    # no main_task, or main_task is done so finished running
+                    return
+                dt = 1
```
dpgeorge, line 171: Since the `_io_queue.map` is empty, the call to `wait_io_event()` below is equivalent to `time.sleep_ms(dt)`.  And since dt is 1 it will wake very often, almost doing a busy-like loop.

But, I guess there's not really any other option.  If it was `time.sleep(forever)` then it wouldn't be able to process `micropython.schedule()` functions.  The point is that it's technically not allowed to schedule functions, so it's quite a gray area what to do here.

Maybe it's worth a small comment here, above the `dt = 1` mentioning this point (that's there's no better option than dt=1).

extmod/asyncio/core.py
```diff
@@ -191,7 +194,17 @@ def run_until_complete(main_task=None):
...
                 if isinstance(er, StopIteration):
+                    if t.state is True:
+                        t.state = False
+                    elif not callable(t.state):
+                        # Schedule any other tasks waiting on the completion of this task.
+                        while t.state.peek():
+                            _task_queue.push(t.state.pop())
+                        # Save return value of coro to pass up to caller.
+                        t.data = er
+                        t.state = False
```
dpgeorge, line 205: This code duplicates what's just below (line 196/209 and onward).

Would it be possible to let the below code run first, and move this `if t is main_task:` block to run after the below code?

extmod/asyncio/core.py
yoctopuce: Absolutely, can't belive I wrote that... It happens to work because when main_task is done it is generally removed from any task queue, but it is clearly not what I wanted to write

extmod/asyncio/core.py
yoctopuce: OK, will do. One thing we could is set dt = 3, as 3ms loop is not as bad when it comes to a busy loop, and there is anyway not hurry in getting out of the loop...

extmod/asyncio/core.py
yoctopuce: Factoring this code was actually my first implementation, but the two nested `if`s in the second copy and the slights differences between the two versions required  the addition of extra local variables, and made the code significantly less readable. Given how important that piece of code is, I chose readability over code factoring. But I can go back to the first implementation if code size is a higher priority.

Regarding the difference between the two "endings": one that I cannot fully justify is the fact that a  `callable(t.state)` would still be dropped in case `t is main_task`. This typically happens when the task is part of a `gather` operation, and not handling it means that if you `run_until_complete` a task that is part of a gather operation, the gather would never end (same as today). I ditd not fix that specific case because it had unexpected side effects that broke other test cases. I can investigate further if you like.

yoctopuce:
@dpgeorge I have pushed an updated commit taking into account all your comments. 

I eventually managed to reuse most of the task chaining code. All checks have passed.

extmod/asyncio/core.py
dpgeorge: The comment and use of `dt = 3` look good now.

extmod/asyncio/core.py
dpgeorge: > Regarding the difference between the two "endings": one that I cannot fully justify is the fact that a `callable(t.state)` would still be dropped in case `t is main_task`

We can leave this as-is for now.

Maybe in the future would be worth revisiting.  It feels like the code should be able to unconditionally set `main_task.state = False` and so skip a bit of the logic there.... not 100% sure though.

dpgeorge:
> I eventually managed to reuse most of the task chaining code.

Very nice!  It's quite a small change now, and the code size diff is much less.  Well done.

dpgeorge (approved):

peterhinch:
Really glad to see this finally fixed! :)

PR #16774, opened 2025-02-18
all: Upgrade to ruff v0.9.6.

cclauss:
### Summary

The Python linter and code formatter [`ruff`](https://docs.astral.sh/ruff) is run in pre-commit and GitHub Actions.  This pull request upgrades from ruff==0.1.3 to the current ruff==0.9.6 changing some configuration for new rules and code formatting stable style 2025.

### Testing

Running ruff locally and in pre-commit.

Linting:
Nine Python test files are excluded because ruff considers that they contain Python SyntaxErrors.
% `ruff rule PLC0206 ` # [dict-index-missing-items](https://docs.astral.sh/ruff/rules/dict-index-missing-items) will be enabled in a followup pull request.  #16776
% `ruff rule PLC1901 ` # [compare-to-empty-string](https://docs.astral.sh/ruff/rules/compare-to-empty-string) is now disabled by default.

Formatting:
Hex constants were converted to lowercase.
String literals were concatenated if they fit within line-length.
Numerous files merely had a single blank line added near the top of the file as part of the stable style for 2025.

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -8,6 +8,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     # ruff version should be kept in sync with .pre-commit-config.yaml
-    - run: pip install --user ruff==0.1.3
+    - run: pipx install ruff==0.9.6
     - run: ruff check --output-format=github .
     - run: ruff format --diff .
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,6 +27,17 @@ line-length = 99
 target-version = "py37"
 
 [tool.ruff.lint]
+exclude = [  # Ruff finds Python SyntaxError in these files
+  "tests/cmdline/repl_autocomplete.py",
+  "tests/cmdline/repl_autoindent.py",
+  "tests/cmdline/repl_basic.py",
+  "tests/cmdline/repl_cont.py",
+  "tests/cmdline/repl_emacs_keys.py",
+  "tests/cmdline/repl_words_move.py",
+  "tests/feature_check/repl_emacs_check.py",
+  "tests/feature_check/repl_words_move_check.py",
+  "tests/micropython/viper_args.py",
+]
 extend-select = ["C9", "PLC"]
 ignore = [
   "E401",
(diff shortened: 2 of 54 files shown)


.github/workflows/ruff.yml
```diff
@@ -8,6 +8,5 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     # ruff version should be kept in sync with .pre-commit-config.yaml
-    - run: pip install --user ruff==0.1.3
-    - run: ruff check --output-format=github .
-    - run: ruff format --diff .
+    - run: pipx run ruff==0.9.6 check --output-format=github .
```
dpgeorge, line 11: Why not keep the existing `pip install` and just change the version to 0.9.6?  At least then you only need to specify the version once.

.github/workflows/ruff.yml
```diff
@@ -8,6 +8,5 @@ jobs:
...
+    - run: pipx run ruff==0.9.6 format --diff .
```
cclauss, line 12:
```suggestion
    - run: pipx install --user ruff==0.9.6
    - run: ruff check --output-format=github .
    - run: ruff format --diff .
```

.github/workflows/ruff.yml
```diff
@@ -8,6 +8,6 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     # ruff version should be kept in sync with .pre-commit-config.yaml
-    - run: pip install --user ruff==0.1.3
+    - run: pipx install --user ruff==0.9.6
```
cclauss, line 11:
```suggestion
    - run: pipx install ruff==0.9.6
```

pyproject.toml
```diff
@@ -37,13 +48,11 @@ ignore = [
...
+  "PLC0206",
 ]
+mccabe.max-complexity = 40
 
-[tool.ruff.mccabe]
-max-complexity = 40
-
-[tool.ruff.per-file-ignores]
+[tool.ruff.lint.per-file-ignores]
 # Exclude all tests from linting (does not apply to formatting).
```
dpgeorge, line 56: I think this comment can be simplified to `# Exclude all tests from linting`, because this section is now explicitly for lint.

cclauss:
```suggestion
# Exclude all tests from linting.
```

dpgeorge (approved):
Thanks for doing this.  I've been using newer ruff versions for a while now and getting warnings about the toml settings being outdated, so it's good to have things updated here.

PR #16781, opened 2025-02-18
ports/rp2: Add SparkFun XRP Controller Beta board files.

sfe-SparkFro:
### Summary

Board files for the XRP Controller (Beta) board: https://www.sparkfun.com/sparkfun-experiential-robotics-platform-xrp-controller.html

### Testing

Tested with a board, all is well!


sfe-SparkFro:
Image added here: https://github.com/micropython/micropython-media/pull/80

ports/rp2/boards/SPARKFUN_XRP_CONTROLLER_BETA/pins.csv
```diff
@@ -0,0 +1,93 @@
...
+MOTOR_R_ENCODER_A,GPIO12
+MOTOR_R_ENCODER_B,GPIO13
+MOTOR_3_ENCODER_A,GPIO0
+MOTOR_3_ENCODER_B,GPIO1
+MOTOR_4_ENCODER_A,GPIO8
+MOTOR_4_ENCODER_B,GPIO9
+SERVO_1,GPIO16
+SERVO_2,GPIO17
+I2C_SDA_1,GPIO18
+I2C_SCL_1,GPIO19
```
dpgeorge, line 57: If `I2C(1)` is available on these pins, you probably want to add the following to `mpconfigboard.h`:
```c
#define MICROPY_HW_I2C1_SCL (19)
#define MICROPY_HW_I2C1_SDA (18)
```
Then users can just do `machine.I2C(1)`, without specifying the pins.

ports/rp2/boards/SPARKFUN_XRP_CONTROLLER_BETA/pins.csv
```diff
@@ -0,0 +1,93 @@
...
+GP16,GPIO16
+GP17,GPIO17
+GP18,GPIO18
+GP19,GPIO19
+GP20,GPIO20
+GP21,GPIO21
+GP22,GPIO22
+GP26,GPIO26
+GP27,GPIO27
+GP28,GPIO28
```
dpgeorge, line 27: Do you need these aliases?  They are there for Pico boards because that's the name of the pins printed on the (back) silk screen.  On your break-out board these are not visible and not the names users will use.  They will use the XRP names.

ports/rp2/boards/SPARKFUN_XRP_CONTROLLER_BETA/pins.csv
sfe-SparkFro: Good catch, fixed!

dpgeorge (approved):

dpgeorge:
Squashed, rebased and merged in cad62c20f2971b1248f55f27ab9f6d9bf8ccdafa

PR #12518, opened 2023-09-26
py/vm.c: Avoid heap-allocating slices for built-ins.

jimmo:
This is an alternative to #10160 and implements a different variation on @damz's idea. It's still worth considering making the compiler do this (via a new `SLICE_AND_SUBSCR` opcode), but this way also has the benefit of handing the is-instance-type check in a simple way.

Unlike #10160 though it only helps for bytecode and not the native emitter.

---

Fast path optimisation for when a BUILD_SLICE is immediately followed by a LOAD/STORE_SUBSCR for a native type to avoid needing to allocate the slice on the heap.

In some cases (e.g. `a[1:3] = x`) this can result in no allocations at all.

We can't do this for instance types because the get/set/delattr implementation may keep a reference to the slice.

This work was funded through GitHub Sponsors.

PR #16785, opened 2025-02-19
extmod/machine_usb_device: Add optional data length parameters to usb_device_submit_xfer()

XIVN1987:
### Summary

I think it's better to give USB submit_xfer() an extra parameter to specify the actual amount of data in the buffer.

This avoids calling submit_xfer as follows
``` python
self.usbd.submit_xfer(self.EP_IN, self.in_buf[:xfer_bytes])
```

Instead, it can be written as:
``` python
self.usbd.submit_xfer(self.EP_IN, self.in_buf, xfer_bytes)
```

The former cannot be used in ISR because it causes memory allocation, and allocating in ISR causes MemoryError.

### Testing

I test this change on rp2/RPI_PICO board.

### DIFF
--- a/docs/library/machine.USBDevice.rst
+++ b/docs/library/machine.USBDevice.rst
@@ -223,11 +223,14 @@ Methods
     and on the host. Returns ``True`` if remote wakeup was enabled and
     active and the host was woken up.
 
-.. method:: USBDevice.submit_xfer(self, ep, buffer /)
+.. method:: USBDevice.submit_xfer(self, ep, buffer, [nbytes] /)
 
             Submit a USB transfer on endpoint number ``ep``. ``buffer`` must be
             an object implementing the buffer interface, with read access for
             ``IN`` endpoints and write access for ``OUT`` endpoints.
+
+            The optional ``nbytes`` parameter can be used to specify the number
+            of bytes to submit from the buffer.
 
             .. note:: ``ep`` cannot be the control Endpoint number 0. Control
                transfers are built up through successive executions of
--- a/extmod/machine_usb_device.c
+++ b/extmod/machine_usb_device.c
@@ -87,8 +87,8 @@ static void usb_device_check_active(mp_obj_usb_device_t *usbd) {
     }
 }
 
-static mp_obj_t usb_device_submit_xfer(mp_obj_t self, mp_obj_t ep, mp_obj_t buffer) {
-    mp_obj_usb_device_t *usbd = (mp_obj_usb_device_t *)MP_OBJ_TO_PTR(self);
+static mp_obj_t usb_device_submit_xfer(size_t n_args, const mp_obj_t *args) {
+    mp_obj_usb_device_t *usbd = (mp_obj_usb_device_t *)MP_OBJ_TO_PTR(args[0]);
     int ep_addr;
     mp_buffer_info_t buf_info = { 0 };
     bool result;
@@ -96,8 +96,8 @@ static mp_obj_t usb_device_submit_xfer(mp_obj_t self, mp_obj_t ep, mp_obj_t buff
     usb_device_check_active(usbd);
 
     // Unmarshal arguments, raises TypeError if invalid
-    ep_addr = mp_obj_get_int(ep);
-    mp_get_buffer_raise(buffer, &buf_info, ep_addr & TUSB_DIR_IN_MASK ? MP_BUFFER_READ : MP_BUFFER_RW);
+    ep_addr = mp_obj_get_int(args[1]);
+    mp_get_buffer_raise(args[2], &buf_info, ep_addr & TUSB_DIR_IN_MASK ? MP_BUFFER_READ : MP_BUFFER_RW);
 
     uint8_t ep_num = tu_edpt_number(ep_addr);
     uint8_t ep_dir = tu_edpt_dir(ep_addr);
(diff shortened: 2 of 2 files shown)


Josverl:
In order to make this discoverable the documentation on this needs to be updated as well. 

https://docs.micropython.org/en/latest/library/machine.USBDevice.html#machine.USBDevice.submit_xfer

Can you add that in a separate commit?

Josverl (approved):

dpgeorge:
> The former cannot be used in ISR because it causes memory allocation, and allocating in ISR causes MemoryError.

I'm surprised that you can call `submit_xfer()` from within a hard ISR, there should be other reasons why that doesn't work.  Do you have some code that does this, calls it from an ISR, and has a `MemoryError`?

An alternative way around this, that doesn't allocate memory, is to pre-create a `memoryview` of the data, eg:
```py
self.mv = memoryview(self.in_buf)
self.mv_slice = self.mv[:xfer_bytes]

...

# then use `self.mv_slice` later on, it won't allocate any memory
```

XIVN1987:
> I'm surprised that you can call `submit_xfer()` from within a hard ISR, there should be other reasons why that doesn't work. Do you have some code that does this, calls it from an ISR, and has a `MemoryError`?

my demo code:
``` python
class CDC_UART(USB_CDC.USB_CDC):
    def __init__(self):
        super().__init__(PID=CDC_UART.PID, EP_IN=CDC_UART.EP_IN, EP_OUT=CDC_UART.EP_OUT, product_string='CDC UART')
        
        self.out_buf = bytearray(64)
        self.in_buf = bytearray(64)

        self.ser = UART(1, 115200, txd='PB0', rxd='PN5', irq=UART.IRQ_RXAny|UART.IRQ_TIMEOUT, callback=lambda ser: self.on_ser_rx())

        self.usbd.active(1)

    def on_ser_rx(self):
        if self.ser.any() >= 64 or self.ser.irq_flags() & UART.IRQ_TIMEOUT:
            xfer_bytes = min(self.ser.any(), 64)
            
            self.ser.readinto(self.in_buf, xfer_bytes)
            
            self.usbd.submit_xfer(self.EP_IN, self.in_buf, xfer_bytes)
```

XIVN1987:
> An alternative way around this, that doesn't allocate memory, is to pre-create a `memoryview` of the data, eg:
> 
> ```python
> self.mv = memoryview(self.in_buf)
> self.mv_slice = self.mv[:xfer_bytes]
> 
> ...
> 
> # then use `self.mv_slice` later on, it won't allocate any memory
> ```

[Slice operation on memoryview causes allocation](https://github.com/micropython/micropython/issues/7718)

I think `self.mv[:xfer_bytes]` will create a new slice object.

dpgeorge:
> my demo code:

Thanks for posting your demo code.  Note that the UART IRQ is actually a soft interrupt that runs at thread level, so it's possible to allocate memory in that callback.

Did you ever see a `MemoryError` in the callback?

> I think `self.mv[:xfer_bytes]` will create a new slice object.

Yes, you are correct, it will allocate.  But the idea is to do that slice outside the IRQ callback and store it for later use.

Well, in this case you need a slice object for all 63 different slice lengths, so it's not very optimal.

XIVN1987:
> > my demo code:
> 
> Thanks for posting your demo code. Note that the UART IRQ is actually a soft interrupt that runs at thread level, so it's possible to allocate memory in that callback.

thank you for response.

I'm not familiar with soft interrupt in micropython. In my port, the UART ISR is a hardware interrupt:
``` c
static void UARTX_Handler(uint32_t uart_id)
{
    pyb_uart_obj_t *self = &pyb_uart_obj[uart_id];

    if(UART_INTStat(self->UARTx, UART_IT_RX_THR | UART_IT_RX_TOUT))
    {
        while(!UART_IsRXFIFOEmpty(self->UARTx))
        {
            fetch_one_char(self);
        }

        if((self->irq_trigger & UART_IRQ_RXAny) && uart_rx_any(self))
            self->irq_flags |= UART_IRQ_RXAny;
        
        if(UART_INTStat(self->UARTx, UART_IT_RX_TOUT))
        {
            UART_INTClr(self->UARTx, UART_IT_RX_TOUT);

            if(self->irq_trigger & UART_IRQ_TIMEOUT)
                self->irq_flags |= UART_IRQ_TIMEOUT;
        }
    }

    if(self->irq_flags && (self->irq_callback != mp_const_none))
    {
        gc_lock();
        nlr_buf_t nlr;
        if(nlr_push(&nlr) == 0)
        {
            mp_call_function_1(self->irq_callback, self);
            nlr_pop();
        }
        else
        {
            self->irq_callback = mp_const_none;

            printf("uncaught exception in UART(%u) interrupt handler\n", self->uart_id);
            mp_obj_print_exception(&mp_plat_print, (mp_obj_t)nlr.ret_val);
        }
        gc_unlock();
    }

    self->irq_flags = 0;
}

void UART0_Handler(void)
{
    UARTX_Handler(PYB_UART_0);
}

void UART1_Handler(void)
{
    UARTX_Handler(PYB_UART_1);
}
```

I will learn soft interrupt to see if I can change the ISR of peripherals to soft inerrrupt.

dpgeorge:
> In my port, the UART ISR is a hardware interrupt

Oh, you have a custom port!  I thought you were using the rp2 port and a Pico board.

You can try to implement soft interrupts, just call `mp_sched_schedule(self->irq_callback, self)` instead of the `mp_call_function_1` call.

I don't think calling `submit_xfer()` is thread safe, I don't think it can be called from within a hard interrupt (although I could be wrong here).  So you may be required to use a soft interrupt.

--------------

In general though, I do support improving things so there is less memory allocation.  But in the case here, I think we should instead improve `memoryview` so you can slice it without memory allocation, see eg #12518.

XIVN1987:
> You can try to implement soft interrupts, just call `mp_sched_schedule(self->irq_callback, self)` instead of the `mp_call_function_1` call.
> 
> I don't think calling `submit_xfer()` is thread safe, I don't think it can be called from within a hard interrupt (although I could be wrong here). So you may be required to use a soft interrupt.
> 
> In general though, I do support improving things so there is less memory allocation. But in the case here, I think we should instead improve `memoryview` so you can slice it without memory allocation, see eg #12518.

Thanks for your reply, I think what you said is very reasonable.

PR #16786, opened 2025-02-19
tests/cpydiff: Test for PEP487 __init_subclass__.

AJMansfield:
### Summary

This change documents and verifies the current absence of `__init_subclass__` functionality, in anticipation of a possible future PEP487 'metaclasses lite' patch.

The main `core_class_initsubclass.py` test verifies if the method is automatically called, while the other three verify other orthogonal properties it should have: if the method is an implicit classmethod; if micropython supplies the base case for the usual recursive function body the PEP encourages; and if kwargs inheritance parameters work correctly.

### Testing

I've verified that these tests fail as expected, correctly detecting the feature's absence.

Some of the tests include code that regularizes across other unrelated differences between CPython and Micropython; but the complexity of this isn't large and it should be easy to verify that they wouldn't ever cause the tests to spuriously pass or fail.

### DIFF
--- a/tests/cpydiff/core_class_initsubclass.py
+++ b/tests/cpydiff/core_class_initsubclass.py
@@ -1,0 +1,21 @@
+"""
+categories: Core,Classes
+description: ``__init_subclass__`` isn't automatically called.
+cause: MicroPython does not currently implement PEP 487.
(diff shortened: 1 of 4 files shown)


AJMansfield:
related: #15511

dpgeorge (approved):
This looks fine, thank you.

PR #16787, opened 2025-02-19
tests/cpydiff: Test for PEP487 __set_name__.

AJMansfield:
### Summary

This change documents and verifies the current absence of `__set_name__` functionality, in anticipation of a possible future PEP487 'metaclasses lite' patch, or another enabling patch to add this missing descriptor feature.

### Testing

I've verified that these tests correctly fail and detect the feature's absence.


dpgeorge:
Thanks for this.

But I'd rather merge one of the implementations of `__set_name__` instead.

AJMansfield:
Ye, just implementing it is my preference as well since it seems like that won't be too far on the horizon.

Issue #16779, opened 2025-02-18
RP2: Pico 2 W Not connecting with thread running


coencoensmeets:
### Port, board and/or hardware

RP2 - Pico 2 W

### MicroPython version

MicroPython v1.25.0-preview.229.g0d46e45a1 on 2025-01-27; Raspberry Pi Pico 2 W with RP2350

### Reproduction

```import network
import time
import _thread
import urequests

SSID = ""
password = ""

def connect_normal():
    wlan = network.WLAN(network.STA_IF)
    wlan.active(True)
    wlan.connect(SSID, password)
    print("Connecting to network")

    start_time = time.time()
    while not wlan.isconnected() and time.time() - start_time < 20:
        print("Waiting for connection...")
        time.sleep(2)

    if wlan.isconnected():
        print("Connected to network")
    else:
        print("Failed to connect to network")

    print("Sleep for 2 seconds")
    time.sleep(2)
    wlan.disconnect()
    
def main_thread():
    connect_normal()

def second_thread():
    for (i) in range(10):
        print("Second thread")
        time.sleep(2)

_thread.start_new_thread(second_thread, ())
main_thread()
```

Other things tested but not working:
- Extending the 20-second time limit.
- Putting `wlan = network.WLAN(network.STA_IF)`, `wlan.active(True)`, and `wlan.connect(SSID, password)` before starting a thread.
- Switching the functions `connection_normal` and `second_thread` to run in the other thread. So the connection is made on the newly opened thread.
- Building firmware with the latest Pico SDK develop
- Adding `rp2.country("NL")` and `network.country(country)`.

### Expected behaviour

Expected to connect to the network and print `Connected to network`.

### Observed behaviour

No connection is established after the set time of 20 seconds. (Time is irrelevant as it never seems to connect). When the line `_thread.start_new_thread(second_thread, ())` is commented out, a connection is found almost every single time.

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

Gadgetoid:
There are stability patches for the WiFi in Pico SDK 2.1.1 (though they are not really mentioned in detail and you need to delve into closed issues and commits) - https://github.com/raspberrypi/pico-sdk/releases/tag/2.1.1

If you can build MicroPython, I'd suggest bumping the SDK and re-trying your tests.

I do suspect this is a different issue, though, (something something IRQs?) threads have been pretty fraught on RP2 and I lose a day every time something like this comes up.

[...]

PR #16788, opened 2025-02-19
rp2: pendsv add support for CPU affinity

Gadgetoid:
### Summary

The CYW43 driver appears to be very sensitive to CPU core affinity, and uses pendsv to schedule critical tasks.

If Core1 is enabled these appear to fire there more often than not, completely breaking wireless bringup.

See: https://github.com/micropython/micropython/issues/16779

This PR is a proposal of sorts, suggesting the addition of CPU affinity to pendsv so that tasks which *must* be run on core0 (or core1, or either) can be designated as such and avoid breakage when users do something daring and controversial like start core1 before attempting to do anything with wireless.

### Testing

Tested on an RP2350 with the included CYW43/mpnetworkport modifications.

It might be possible to fix the CYW43 driver to be less sensitive to these issues, outright block pendsv from running on core1 (I don't really understand the implications of this, so why I'm even proposing this change is a mystery!)

Honestly at this point I'm for disabling core1 on the Pico. It's a horrible bag of footguns that distracts people from asyncio and even when it's not outright buggy or broken (I have spent days tracking down, reporting and fixing issues) it's fertile ground for users to get themselves into all sorts of trouble.

### DIFF
--- a/ports/rp2/cyw43_configport.h
+++ b/ports/rp2/cyw43_configport.h
@@ -126,7 +126,7 @@ uint cyw43_get_pin_wl(cyw43_pin_index_t pin_id);
 #define cyw43_hal_get_mac_ascii         mp_hal_get_mac_ascii
 #define cyw43_hal_generate_laa_mac      mp_hal_generate_laa_mac
 
-#define cyw43_schedule_internal_poll_dispatch(func) pendsv_schedule_dispatch(PENDSV_DISPATCH_CYW43, func)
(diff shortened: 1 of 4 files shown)


projectgus:
> It might be possible to fix the CYW43 driver to be less sensitive to these issues, outright block pendsv from running on core1 (I don't really understand the implications of this, so why I'm even proposing this change is a mystery!)

I reckon this is actually a very good idea, although it turned out the CYW43 driver didn't need a lot of changes to work across cores. See #16915 for another possible fix.

Gadgetoid:
> I reckon this is actually a very good idea

Even a stopped clock is right twice a day 😆

Thank you for taking a shot at this.

PR #16789, opened 2025-02-20
mimxrt/boards/ADAFRUIT_METRO_M7/mpconfigboard.mk: Reduce flash freq.

robert-hh:
### Summary

The flash clock was set to 133Mhz and tested at that frequency.

At a re-test today it turned out to be right, even if the flash chip is specified for 133Mh. The board does not boot. Reducing the flash CLK to 100 MHz the board boots again and works reliable.

### Testing

Tested with the target board. The UF2 bootloader runs at 100MHz, so no need for a change of the bootloader..


dpgeorge (approved):

dpgeorge:
Thanks, merged.

robert-hh:
Thank you. I do not really understand why it failed, because once the CLK was set to 100Mhz, it can be set later to any frequency between 30MHz and 133MHz (verified with Oscilloscope). That's how I tested the divider table.

PR #16793, opened 2025-02-21
zephyr: Add watchdog timer implementation.

danicampora:
Simple implementation in-line with the rest of the Micropython ports.


jonnor:
Very nice! What hardware have you tested this on? Any particular concerns if I would want to test this on hardware? I think I have an NRF52 available, at least.

danicampora:
I tested it on the nRF52832 and the nRF5340

ports/zephyr/machine_wdt.c
```diff
@@ -0,0 +1,95 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016 Paul Sokolovsky
+ * Copyright (c) 2017 Eric Poulsen
+ * Copyright (c) 2025 Daniel Campora on behalf of REMOTE TECH LTD
```
dpgeorge, line 8: I appreciate that you want to retain authorship here.  But this file is very simple and it doesn't look like there's much code actually copied from other sources, most of the code here is original work made to integrate with Zephyr.

How much was actually copied verbatim and unchanged?

We should definitely keep authorship when relevant, but it doesn't really make sense to unnecessarily keep copyright lines.

ports/zephyr/machine_wdt.c
```diff
@@ -0,0 +1,95 @@
...
+    int32_t channel_id;
+} machine_wdt_obj_t;
+
+static machine_wdt_obj_t wdt_default = {
+    {&machine_wdt_type}, NULL, -1
+};
+
+static machine_wdt_obj_t *mp_machine_wdt_make_new_instance(mp_int_t id, mp_int_t timeout_ms) {
+    if (id != 0) {
+        mp_raise_ValueError(MP_ERROR_TEXT("Invalid WDT ID"));
```
dpgeorge, line 53: Please use lowercase at the start of error messages.

Also ID could be lowercase id because that matches the keyword-argument name.

ports/zephyr/mpconfigport.h
```diff
@@ -68,6 +68,8 @@
 #define MICROPY_PY_MACHINE_PIN_MAKE_NEW mp_pin_make_new
 #define MICROPY_PY_MACHINE_UART     (1)
 #define MICROPY_PY_MACHINE_UART_INCLUDEFILE "ports/zephyr/machine_uart.c"
+#define MICROPY_PY_MACHINE_WDT      (1)
```
dpgeorge, line 71: Maybe this could be wrapped in `#if CONFIG_WATCHDOG` ... `#endif` so that it's only enabled if the watchdog is enabled in the Kconfig?

dpgeorge (commented):
Thanks for this, it's nicely integrated into the common extmod bindings.

danicampora:
@dpgeorge made the changes you requested and re-pushed.

ports/zephyr/machine_wdt.c
```diff
@@ -0,0 +1,94 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2016 Paul Sokolovsky
+ * Copyright (c) 2017 Eric Poulsen
```
dpgeorge, line 7: I think you should definitely add your copyright line.

What I meant is that you can probably omit these existing ones.  The terms of the MIT license are "substantial portions of the Software", and I don't think there's anything of substance that was copied from another file to this one.  Unless I'm wrong?

danicampora: Ahh... somehow I understood your first comment completely different. Makes sense now. Will add it. Thanks.

ports/zephyr/mpconfigport.h
danicampora: I agree. Thanks!

dpgeorge:

PR #16794, opened 2025-02-21
zephyr: Create machine.Timer class implementation.

danicampora:
Simple machine.Timer implementation in-line with the rest of the Micropython ports.

Note: Only virtual timers are supported (not linked to any particular hardware peripheral).

### DIFF
--- a/ports/zephyr/machine_timer.c
+++ b/ports/zephyr/machine_timer.c
@@ -1,0 +1,201 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 Daniel Campora on behalf of REMOTE TECH LTD
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include <stdint.h>
+#include <stdio.h>
+
+#include "py/mperrno.h"
+#include "py/obj.h"
+#include "py/runtime.h"
+#include "modmachine.h"
+#include <zephyr/kernel.h>
+#include <zephyr/device.h>
+#include <zephyr/sys/printk.h>
+
+#define TIMER_MS_PER_TICK (1000)
+
+typedef enum _machine_timer_mode_t {
+    TIMER_MODE_ONE_SHOT = 0,
+    TIMER_MODE_PERIODIC
+} machine_timer_mode_t;
+
+typedef struct _machine_timer_obj_t {
+    mp_obj_base_t base;
+
+    struct k_timer my_timer;
+
+    machine_timer_mode_t mode;
+    uint32_t period_ms;
+
+    mp_obj_t callback;
+
+    struct _machine_timer_obj_t *next;
+} machine_timer_obj_t;
+
+const mp_obj_type_t machine_timer_type;
+
+static mp_obj_t machine_timer_init_helper(machine_timer_obj_t *self, mp_uint_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args);
+static mp_obj_t machine_timer_deinit(mp_obj_t self_in);
+
+static void machine_timer_callback(struct k_timer *timer) {
+    machine_timer_obj_t *self = (machine_timer_obj_t *)k_timer_user_data_get(timer);
+    if (self->mode == TIMER_MODE_ONE_SHOT) {
+        machine_timer_deinit(self);
+    }
+    if (self->callback != mp_const_none) {
+        mp_sched_schedule(self->callback, MP_OBJ_FROM_PTR(self));
+    }
+}
+
+static void machine_timer_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
+    machine_timer_obj_t *self = self_in;
+    qstr mode_str = (self->mode == TIMER_MODE_PERIODIC) ? MP_QSTR_PERIODIC : MP_QSTR_ONE_SHOT;
+    mp_printf(print, "Timer(-1, mode=%q, period=%lu)", mode_str, self->period_ms);
+}
+
+static mp_obj_t machine_timer_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
+
+    if (mp_obj_get_int(args[0]) != -1) {
+        mp_raise_ValueError(MP_ERROR_TEXT("only virtual timers are supported"));
+    }
+
+    // Create the new timer.
+    machine_timer_obj_t *self = mp_obj_malloc_with_finaliser(machine_timer_obj_t, &machine_timer_type);
+
+    // Add the timer to the linked-list of timers
+    self->next = MP_STATE_PORT(machine_timer_obj_head);
+    MP_STATE_PORT(machine_timer_obj_head) = self;
+
+    if (n_args > 1 || n_kw > 0) {
+        mp_map_t kw_args;
+        mp_map_init_fixed_table(&kw_args, n_kw, args + n_args);
+        machine_timer_init_helper(self, n_args - 1, args + 1, &kw_args);
+    }
+    return self;
+}
+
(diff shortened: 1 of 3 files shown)


jonnor:
From a very brief scan, this looks pretty nice and neat. Are there any particular caveats with this implementation compared to existing implementations? (apart from mentioned only virtual timers, not hardware).

The virtual timers should work on any Zephyr platform, right?

danicampora:
@jonnor thanks!

No caveats besides the only virtual timers part. It should definitely work with any Zephyr port. I tested it with the nRF5340 and the nRF52840. It doesn't require any additional config options on Zephyr.

ports/zephyr/machine_timer.c
```diff
@@ -0,0 +1,196 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * Development of the code in this file was sponsored by Microbric Pty Ltd
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2013-2015 Damien P. George
+ * Copyright (c) 2016 Paul Sokolovsky
+ * Copyright (c) 2025 Daniel Campora on behalf of REMOTE TECH LTD
```
dpgeorge, line 10: As with #16793, I suggest to only copy across copyright lines where you took "substantial portions of the Software" from somewhere else.

ports/zephyr/machine_timer.c
```diff
@@ -0,0 +1,196 @@
...
+
+static mp_obj_t machine_timer_init_helper(machine_timer_obj_t *self, mp_uint_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum {
+        ARG_mode,
+        ARG_callback,
+        ARG_period,
+        ARG_freq,
+    };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_mode,         MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = 1} },
```
dpgeorge, line 121: I suggest making constants:
```c
enum {
    TIMER_MODE_ONE_SHOT = 0,
    TIMER_MODE_PERIODIC = 1,
};
```
and use them here and in the locals dict, just be clear what's going on, and what the default is.

ports/zephyr/machine_timer.c
```diff
@@ -0,0 +1,196 @@
...
+        { MP_QSTR_callback,     MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_obj = mp_const_none} },
+        { MP_QSTR_period,       MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = 0xffffffff} },
+        #if MICROPY_PY_BUILTINS_FLOAT
+        { MP_QSTR_freq,         MP_ARG_KW_ONLY | MP_ARG_OBJ, {.u_obj = mp_const_none} },
+        #else
+        { MP_QSTR_freq,         MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = 0xffffffff} },
+        #endif
+    };
```
dpgeorge, line 129: There is also a `tick_hz` argument that could be implemented (see eg `extmod/machine_timer.c`, and `ports/rp2/machine_timer.c`).  But it's not documented so I'm happy if you want to leave it out.

ports/zephyr/machine_timer.c
danicampora: @done.

ports/zephyr/machine_timer.c
```diff
@@ -0,0 +1,195 @@
...
+    k_ticks_t ticks = k_timer_remaining_ticks(&self->my_timer);
+    return MP_OBJ_NEW_SMALL_INT(k_ticks_to_ms_near32(ticks));
+}
+static MP_DEFINE_CONST_FUN_OBJ_1(machine_timer_value_obj, machine_timer_value);
+
+static const mp_rom_map_elem_t machine_timer_locals_dict_table[] = {
+    { MP_ROM_QSTR(MP_QSTR___del__),     MP_ROM_PTR(&machine_timer_deinit_obj) },
+    { MP_ROM_QSTR(MP_QSTR_deinit),      MP_ROM_PTR(&machine_timer_deinit_obj) },
+    { MP_ROM_QSTR(MP_QSTR_init),        MP_ROM_PTR(&machine_timer_init_obj) },
+    { MP_ROM_QSTR(MP_QSTR_value),       MP_ROM_PTR(&machine_timer_value_obj) },
```
dpgeorge, line 180: This `value` method is non-standard, or at least it's undocumented.

Did you find a need for it?

danicampora: I saw it in other ports and I thought it was nice to have.

ports/zephyr/machine_timer.c
danicampora: I would prefer to leave it out.

ports/zephyr/machine_timer.c
```diff
@@ -0,0 +1,196 @@
...
+    machine_timer_obj_t *self = self_in;
+    qstr mode = self->repeat ? MP_QSTR_PERIODIC : MP_QSTR_ONE_SHOT;
+    mp_printf(print, "Timer(-1, mode=%q, period=%lu ms)", mode, self->period);
+}
+
+static mp_obj_t machine_timer_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
+
+    if (mp_obj_get_int(args[0]) != -1) {
+        mp_raise_ValueError(MP_ERROR_TEXT("Only virtual timers are supported!"));
```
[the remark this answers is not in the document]
danicampora: Fixed.

ports/zephyr/machine_timer.c
danicampora: Done.

ports/zephyr/machine_timer.c
dpgeorge: OK, it does seem useful.

dpgeorge (approved):
Tested on the qemu_cortex_m3 virtual board.

dpgeorge:
Actually, I just realised this PR has a few issues regarding finalisation:
- it defines `__del__` but doesn't use `mp_obj_malloc_with_finaliser()` so the finaliser won't ever be used
- when you call `Timer.deinit()` it doesn't remove the timer from the linked list
- if your application has a loop that creates a timer, uses it, discards it, then creates another one, eventually it will run out of memory because the old timer objects stay on the linked list

I think all these issues can be solved by following how the rp2 port works:
- use `mp_obj_malloc_with_finaliser()` to allocate the object
- get rid of the linked list of timers
- get rid of the `machine_timer_deinit_all()` function

That should work because `gc_sweep_all()` will call all of the finalisers on soft reset, which will call `k_timer_stop(...)` which should be enough to clean everything up.

The good thing about Zephyr here is that it doesn't have any memory on its side allocated to the timer (from what I understand).  So for the clean-up it's enough to just GC the `Timer` instance and call `k_timer_stop()`.

dpgeorge:
@danicampora I force pushed to this branch because I was about to merge it... so please use `git pull --rebase` when you check out your branch to update it.

danicampora:
@dpgeorge thanks. I'll work on your comments and will update this branch asap.

dpgeorge:
On further thought, I think you need to retain the linked list, because otherwise you have no way to trace the timers that have been registered with Zephyr and are awaiting a callback from Zephyr when they expire.

I think you need to:
- use `mp_obj_malloc_with_finaliser()` to allocate the timer instance
- keep the linked list but remove the `machine_timer_deinit_all()` function, it'll be handled by `gc_sweep_all()`
- remove the timer from the linked list when `Timer.deinit()` is called

VynDragon:
tested on nrf52840, it works.

order: 
- 16207 (Upgrade to Zephyr v4.0.0)
- 16770 (Fixes for zephyr 4.0)
merged thread analyzer as:
```c
#ifdef CONFIG_THREAD_ANALYZER
static mp_obj_t mod_thread_analyze(mp_obj_t cpu_in) {
    #if KERNEL_VERSION_NUMBER >= ZEPHYR_VERSION(4, 0, 0)
    unsigned int cpu = mp_obj_get_int(cpu_in);
    thread_analyzer_print(cpu);
    #else
    thread_analyzer_print();
    #endif
    return mp_const_none;
}
static MP_DEFINE_CONST_FUN_OBJ_1(mod_thread_analyze_obj, mod_thread_analyze);
#endif
```
- 16794 (this PR)

danicampora:
@dpgeorge thanks for the remarks. I just forced pushed again and I think it reflects everything you proposed.

@VynDragon thanks for testing!

ports/zephyr/machine_timer.c
```diff
@@ -0,0 +1,200 @@
...
+    }
+    if (self->callback != mp_const_none) {
+        mp_sched_schedule(self->callback, MP_OBJ_FROM_PTR(self));
+    }
+}
+
+static void machine_timer_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
+    machine_timer_obj_t *self = self_in;
+    qstr mode_str = (self->mode == TIMER_MODE_PERIODIC) ? MP_QSTR_PERIODIC : MP_QSTR_ONE_SHOT;
+    mp_printf(print, "Timer(-1, mode=%q, period=%lu ms)", mode_str, self->period_ms);
```
dpgeorge, line 75: Please remove the ` ms` bit from the printing, because in principal the things that's printed out should be valid Python.

danicampora: @dpgeorge fixed. Sorry about that.

ports/zephyr/machine_timer.c
```diff
@@ -0,0 +1,200 @@
...
+} machine_timer_obj_t;
+
+const mp_obj_type_t machine_timer_type;
+
+static mp_obj_t machine_timer_init_helper(machine_timer_obj_t *self, mp_uint_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args);
+
+static void machine_timer_callback(struct k_timer *timer) {
+    machine_timer_obj_t *self = (machine_timer_obj_t *)k_timer_user_data_get(timer);
+    if (self->mode == TIMER_MODE_ONE_SHOT) {
+        k_timer_stop(&self->my_timer);
```
dpgeorge, line 65: I think call `machine_timer_deinit(MP_OBJ_FROM_PTR(self))` here, so that the one-shot timer is removed from the linked list.

That way you should be able to create one-shot timers indefinitely without deinit'ing them explicitly.

danicampora: Yeah that's a good point. Done now. Thanks.

dpgeorge (approved):

ports/zephyr/machine_timer.c
dpgeorge: Great thanks.  I tested on `frdm_k64f` creating many one-shot timers in a loop, and it works.

Issue #16709, opened 2025-02-05
Delayed PIO Assambly with custom variables


HighImp:
### Description

Hi,

I would like to start a pull request for a alternative of the decorator: ```def asm_pio(**kw):``` that can be called afterward.
This change would allow all "constants" to be modified during runtime in Python, significantly improving flexibility.

We use PIOs in our project for various situations. See my [Reddit post](https://www.reddit.com/r/MicroPythonDev/comments/1i6utf6/comment/m8w57ek/) for a specific use case.
The discussed solution requires two state machines instead of one.

The adaptation is ready and working—see also #7903, where I posted my first proposal for the new behavior and the [Branch](https://github.com/HighImp/micropython/tree/topic/add_delay_assambly) if you are interested.

Best regards!

### Code Size

I am not sure how much code will be used, but the adaptation in the branch takes 63 additional lines. During operation, the fact that the function will not collapse instantly into a tuple of bytes may require more RAM than the original approach. However, I think it is worth it, considering the amount of saved space in the PIO modules. I would love to hear your thoughts!

### Implementation

I intend to implement this feature and would submit a Pull Request if desirable

### Code of Conduct

Yes, I agree

Gadgetoid:
This is interesting, and I'd probably use a feature like this, but to play devil's advocate I have a couple of concerns:

1. I'm not sure maintaining two separate functions with extremely similar behaviour is a reasonable approach.
2. By polluting the function namespace with `gl.update(variables)` you're missing the chance to be explicit about what's parametisable
3. The PIO compiler is pretty sophisticated if you think about it, and limiting variables to 0-31 ints is unnecessary

Consider this horrible hack of blink_1hz.py (the timings are off on RP2350 🤔 which throws an exception if the PIO freq is set to 2000Hz... but that's uh, another issue):

```python
# Example using PIO to blink an LED and raise an IRQ at 1Hz.
# Note: this does not work on Pico W because it uses Pin(25) for LED output.

# ruff: noqa: F821 - @asm_pio decorator adds names to function scope

import time
from machine import Pin
import param_rp2 as rp2

DELAY_TIME = 1


@rp2.asm_pio(set_init=rp2.PIO.OUT_LOW)
def blink_fixed():
    # fmt: off
    # Cycles: 1 + 1 + 6 + 32 * (30 + 1) = 1000
    irq(rel(0))
    set(pins, 1)
    set(x, 31)                  [5]
    label("delay_high")
    nop()                       [29]
[...]

PR #16800, opened 2025-02-24
rp2: add pio asm parameters

HighImp:
### Summary
The PIO can be used more flexibly by adding a runtime parameter to modify previously static values during code execution.
Without this parameter, every desired manipulation would require a separate PIO function.
See #16709 for the discussion and [Reddit](https://www.reddit.com/r/MicroPythonDev/comments/1i6utf6/comment/m8w57ek/) for a specific use case.

### Testing
- Added some PIO tests to the PR and ran them on an RP2040, but not yet on an RP2350.
- Existing tests do not seem to manipulate pins, can be added if required (Are tests expected to run on a bare board?) 

- Due to the improved proposal from @Gadgetoid, the changes consist of just two extra lines and six modifications in total.
- If the feature is not used, it will only consume space. However, consolidating two or more PIO codes into one significantly reduces flash usage.

### DIFF
--- a/ports/rp2/modules/rp2.py
+++ b/ports/rp2/modules/rp2.py
@@ -243,7 +243,7 @@ _pio_instructions = (
 def asm_pio(**kw):
     emit = PIOASMEmit(**kw)
 
-    def dec(f):
+    def dec(f, **kw):
         nonlocal emit
 
         gl = f.__globals__
@@ -257,10 +257,10 @@ def asm_pio(**kw):
             gl[name] = getattr(emit, name)
 
         emit.start_pass(0)
-        f()
+        f(**kw)
 
         emit.start_pass(1)
-        f()
+        f(**kw)
 
         gl.clear()
         gl.update(old_gl)
--- a/docs/rp2/tutorial/pio.rst
+++ b/docs/rp2/tutorial/pio.rst
@@ -124,3 +124,64 @@ the following:
 
 The entire routine takes exactly 2000 cycles of the state machine.  Setting the
 frequency of the state machine to 2000Hz makes the LED blink at 1Hz.
+
+Variables inside PIO
+----------
+
+Due to the process of converting pythonic PioASM-Code into HexCode, 
+it is not possible to access to global references inside a pio-function. 
+However, keyword arguments can be used to manipulate the asm-program.
+
+This mechanic allows to write a pio program "template" 
+and calculate/determine the necessary values for the PIO-variables at runtime.
+
+The following example demonstrates how to calculate the delay for a given frequency
+and how to use the calculated values in the PIO program.
+
+.. code-block:: python3
+
+    def calc_delay(freq):
+        """Calculate the pio-variables for a given frequency"""
+        half_period = int(1000/freq) # 2000Hz PIO frequency, devided in high and low phase
+        error = round(1000/half_period - freq, 3) # calculate the integer cast error
+        loop_count = (half_period // 31) # 31 cycles for the delay loop
+        set_delay = half_period - loop_count * 31 - 2 # 2 cycles for the set and irq instructions
+        loop_count -= 1  # compensate for the first loop
+        assert 31 >= loop_count >= 0, "Frequency out of range"
+        assert 31 >= set_delay >= 0, "Frequency out of range"
+        return loop_count, set_delay, error
+
+    
+    def blink_template(loop_count, set_delay):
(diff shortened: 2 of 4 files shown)


dpgeorge:
Thanks for the PR, this is a neat addition.

For reference, an alternative that already works but is kind of tricky: https://github.com/micropython/micropython/issues/16709#issuecomment-2688513118

ports/rp2/modules/rp2.py
```diff
@@ -241,10 +241,14 @@ def set(self, dest, data):
 
 
 def asm_pio(**kw):
+    fn = kw.pop("program") if "program" in kw else None
+    args = kw.pop("args") if "args" in kw else {}
```
dpgeorge, line 245: Even simpler, you can use the 2-arg form of `pop` to return the default, eg `fn = kw.pop("program", None)`.

dpgeorge:
> * Existing tests do not seem to manipulate pins, can be added if required (Are tests expected to run on a bare board?)

Most tests run on a bare board.  But some like `tests/extmod_hardware/machine_pwm.py` require hardware connections, mostly a jumper across a pair of pins.  It's nice to have tests work on a bare board.  You could add a second test that requires a pair of pins to be connected.

dpgeorge:
I think I found a neater way to do this:
```diff
--- a/ports/rp2/modules/rp2.py
+++ b/ports/rp2/modules/rp2.py
@@ -243,7 +243,7 @@ _pio_instructions = (
 def asm_pio(**kw):
     emit = PIOASMEmit(**kw)
 
-    def dec(f):
+    def dec(f, **kw):
         nonlocal emit
 
         gl = f.__globals__
@@ -257,10 +257,10 @@ def asm_pio(**kw):
             gl[name] = getattr(emit, name)
 
         emit.start_pass(0)
-        f()
+        f(**kw)
 
         emit.start_pass(1)
-        f()
+        f(**kw)
 
         gl.clear()
         gl.update(old_gl)
```

Then use it like this:
```py
def dynamic_pio_code(custom_parameter):
   ...

pio_prog = rp2.pio_asm()(dynamic_pio_code, custom_parameter=123)
```

mendenm:
@dpgeorge i have a simple syntactic question about the code you quote above.  
Why the `nonlocal` declaration? This declaration is, as far as I can tell, ineffective.  This would make sense if the contents of the variable `emit` were being replaced. In this case, though, the access to the variable is strictly a read, which should work by normal python scoping.  The state of the object inside the variable is being modified, not the variable itself.  I have seen this construct scattered through other upy code, too.  Am I missing something? Otherwise, the presence of the declaration actually implies to the reader that something can happen (replacement of the value of `emit`), which is not going to happen.

dpgeorge:
@mendenm you are correct, the `nonlocal` is not needed in this case.  I can see from the history that it was there at the very beginning so either (1) the original development code for this feature did write to the `emit` variable and needed the `nonlocal` and it was never changed when `emit` didn't need writing to; or (2) I did not realize that `nonlocal` wasn't needed in this case when I wrote the code, or I was being defensive.

If you'd like to make a PR to remove it, that would be good.

Where are other instances of this that you've seen?

Gadgetoid:
> I think I found a neater way to do this:
> 
> ```diff
> --- a/ports/rp2/modules/rp2.py
> +++ b/ports/rp2/modules/rp2.py
> @@ -243,7 +243,7 @@ _pio_instructions = (
>  def asm_pio(**kw):
>      emit = PIOASMEmit(**kw)
>  
> -    def dec(f):
> +    def dec(f, **kw):
>          nonlocal emit
>  
>          gl = f.__globals__
> @@ -257,10 +257,10 @@ def asm_pio(**kw):
>              gl[name] = getattr(emit, name)
>  
>          emit.start_pass(0)
> -        f()
> +        f(**kw)
>  
>          emit.start_pass(1)
> -        f()
> +        f(**kw)
>  
>          gl.clear()
>          gl.update(old_gl)
> ```
> 
> Then use it like this:
> 
> ```python
> def dynamic_pio_code(custom_parameter):
>    ...
> 
> pio_prog = rp2.pio_asm()(dynamic_pio_code, custom_parameter=123)
> ```

Beautiful. A perfect combination of the approach you mentioned in https://github.com/micropython/micropython/issues/16709#issuecomment-2688513118 and minimal supporting changes. Avoids the dict nastiness. 10/10 would PIO ASM again.

Looks like this PR is missing supporting documentation tweaks, too!

dpgeorge:
> Looks like this PR is missing supporting documentation tweaks,

Yes, this change will need some updates to the docs to describe how to use it.

PR #16804, opened 2025-02-24
docs: Fix double 'the' in documentation.

rxwb:
### Summary

In a couple of places a double 'the' snuck into the documentation. This PR fixes that.
There"s more of this in some TI and a lot of STM32 files, but I didn't think it was worth addressing this.  

### Testing

Not necessary. Only changes to documentation.


dpgeorge (approved):

dpgeorge:
Thanks for the fix, I think it's reasonable to just fix this in the docs.

PR #16808, opened 2025-02-24
Draft Catch and raise deadlock in adc config.

andrewleech:
### Summary

Seen on a project using the stm32wb55 occasionally the device was locking up, pausing on the attached debugger showed the code was stuck in this while loop waiting for the ADC to report ready which never occurred.

The ADC read in progress was of a pin that's regularly read in a looping async task, normally operates fine.

The lockup was a number of times in the same place, in all cases it seemed to occur shortly after a cpu speed change.

### Testing

I've not been able to reliably reproduce the original failure so I'm not yet certain if this exception raise works as expected.
After the issue occurs I'm not sure if re-trying the ADC read is enough to get it working or whether some broader reset of the ADC is required.

If the issue is caused by the cpu speed change then some kind of ADC block / clock reset might be possible at a lower level as part of the speed change logic.

### DIFF
--- a/ports/stm32/machine_adc.c
+++ b/ports/stm32/machine_adc.c
@@ -28,6 +28,8 @@
 // extmod/machine_adc.c via MICROPY_PY_MACHINE_ADC_INCLUDEFILE.
 
 #include "py/mphal.h"
+#include "py/mperrno.h"
+#include "py/runtime.h"
 #include "adc.h"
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
@andrewleech can we close this in favour of #19047?

andrewleech:
Thanks yes I forgot this was still open as a separate PR

PR #16809, opened 2025-02-25
Draft modmachine: Expose keyboard interrupt as a python function.

andrewleech:
### Summary

Allows throwing a keyboard interrupt from code, was used from a timer interrupt that was configured to act like a "soft watchdog" with this interrupt triggering a stack trace printout on the main thread.

### Testing

Currently this causes a fatal error if another keyboard interrupt comes in on serial line (I think)

A separate / explicit version of this might be better rather than reusing / overloading keyboard interrupt.

### DIFF
--- a/py/modmicropython.c
+++ b/py/modmicropython.c
@@ -166,6 +166,14 @@ static mp_obj_t mp_micropython_schedule(mp_obj_t function, mp_obj_t arg) {
 static MP_DEFINE_CONST_FUN_OBJ_2(mp_micropython_schedule_obj, mp_micropython_schedule);
 #endif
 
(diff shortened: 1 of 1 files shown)


kwagyeman:
Useful for https://github.com/micropython/micropython/issues/19424. Needs an extra unused argument so it can be scheduled without a wrapper function.

dpgeorge:
@andrewleech I propose closing this PR in favour of #19467.  The implementation here doesn't really work as expected, it's pretty much equivalent to just doing `raise KeyboardInterrupt`.

andrewleech:
Ah yes, that explains the issues I had with it! Your version looks much better thanks.

PR #16810, opened 2025-02-25
extmod/vfs_rom: Add bounds checking for all filesystem accesses.

dpgeorge:
### Summary

While testing ROMFS as part of #8381 I found that it was relatively easy to end up with a corrupt filesystem on the device (eg due to the ROMFS deploy process stopping half way through), which could lead to hard crashes.  Notably: boot loops trying to mount a corrupt filesystem, crashes when importing modules like `os` that first scan the filesystem for `os.py`, and crashing when deploying a new ROMFS in certain cases because the old one is removed while still mounted.

@robert-hh also encountered similar issues when testing #8381.

This PR adds full bounds checking for all ROMFS filesystem accesses.

### Testing

The new code should be fully covered by the new tests that are added here, and they run in CI.

I was hoping such bounds checking would not be necessary, because it increases the size and complexity of the ROMFS driver.  But after playing around with ROMFS in a project, it's definitely necessary to have full bounds checking.

### DIFF
--- a/extmod/vfs_rom.c
+++ b/extmod/vfs_rom.c
@@ -91,6 +91,7 @@
 #define ROMFS_RECORD_KIND_DATA_POINTER (3)
 #define ROMFS_RECORD_KIND_DIRECTORY (4)
 #define ROMFS_RECORD_KIND_FILE (5)
+#define ROMFS_RECORD_KIND_FILESYSTEM (0x14a6b1)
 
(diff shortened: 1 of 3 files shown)


dpgeorge:
I needed to increase the MicroPython heap for the qemu Sabrelite board, because the new ROMFS test is quite large when compiled using the ARM (non-Thumb) native emitter.

projectgus (approved):

PR #16812, opened 2025-02-25
py/obstr: Support tuples as pre/suffix in startswith() and endswith().

glenn20:
This PR adds support for passing tuples as the prefix/suffix argument to the `str.startswith()` and `str.endswith()` methods. This improves compatibility with CPython.

The methods will return `True` if the string starts/ends with any of the prefixes/suffixes in the tuple.

The PR also adds uniform support for the optional `start` and `end` arguments of `str.startswith()` and `str.endswith()`. 

Taken together, these provide full compatibility with CPython `str.startswith()` and `str.endswith()` methods (see [python docs](https://docs.python.org/3/library/stdtypes.html#str.startswith)).

The new signatures of the methods are:

```py
bool str.startswith(prefix: str | tuple[str] [, start: int | None [, end: int | None]])
bool str.endswith(suffix: str | tuple[str] [, start: int | None [, end: int | None]])
```

### Testing

The tests in `tests/basics/string_{starts,ends}with.py` have been extended to test the new features. These tests are passing on the unix port.

### Tradeoffs

This PR should be fully backward compatible with existing python code running on micropython.

While improving compliance wih cpython, this PR will provide a small increase in code size. This may be mitigated by reusing the new `objstr.c:get_substring_data()` function to reduce code size elsewhere (eg. `objstr.c:str_count()` and `objstr.c:str_finder()`).


dpgeorge:
Thanks, this is something I've wanted for a while but was afraid it would be a big feature.  It's quite impressive how small the code size change is here!

You will need to get the CI passing, I think just remove the `endswith` test from `misc/non_compliant.py`.

glenn20:
> Thanks, this is something I've wanted for a while but was afraid it would be a big feature. It's quite impressive how small the code size change is here!

Ta. It got smaller once I realised `startswith()` and `endswith()` can share the same implementation. May even reduce the code size if we reuse `get_substring_data()` in `str_count()` and `str_finder()`.

> You will need to get the CI passing, I think just remove the `endswith` test from `misc/non_compliant.py`.

Done. I'm still getting failures on `extmod/select_poll_fd`, but that is also true for master.

dpgeorge:
CI now passing.

glenn20:
> May even reduce the code size if we reuse `get_substring_data()` in `str_count()` and `str_finder()`.

I just pushed a commit to do this and there is a tiny overall code size reduction:

```
   bare-arm:    -8 -0.014% 
minimal x86:    -3 -0.002% 
   unix x64:   -72 -0.008% standard
      stm32:   -96 -0.024% PYBV10
     mimxrt:  -104 -0.028% TEENSY40
        rp2:   -72 -0.008% RPI_PICO_W
       samd:  -100 -0.037% ADAFRUIT_ITSYBITSY_M4_EXPRESS
  qemu rv32:   -78 -0.017% VIRT_RV32
```

cclauss:
This is great to see!! This should enable MicroPython code to pass [`ruff rule PIE810`](https://docs.astral.sh/ruff/rules/multiple-starts-ends-with).

Once this has been merged, you might also consider adding `str.replaceprefix()` or `str.replacesuffix()` which is suggested by [ruff rule FURB188 ](https://docs.astral.sh/ruff/rules/slice-to-remove-prefix-or-suffix).

glenn20:
> Once this has been merged, you might also consider adding [`str.removeprefix()`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) or [`str.removesuffix()`](https://docs.python.org/3/library/stdtypes.html#str.removesuffix) which is suggested by [ruff rule FURB188 ](https://docs.astral.sh/ruff/rules/slice-to-remove-prefix-or-suffix).

Done ;). See PR#16821. Thanks for the suggestion @cclauss.

dpgeorge:
@glenn20 I like the additional commit which refactors code using `get_substring_data()`.  I think that should stay as a separate commit, but to make that commit a bit smaller, I suggest placing the new `get_substring_data()` function near the top of the file in the first commit that adds it (so it doesn't need to be moved in the third commit).  Eg put it just after `check_is_str_or_bytes()` and before the `*********` comment line.

glenn20:
Commits split, reordered, squashed and rebased.

py/objstr.c
```diff
@@ -748,23 +768,11 @@ static mp_obj_t str_finder(size_t n_args, const mp_obj_t *args, int direction, b
 
     GET_STR_DATA_LEN(args[0], haystack, haystack_len);
```
dpgeorge, line 769: `haystack_len` is no longer used, so this could be maybe replaced with something slightly smaller/efficient?

Also, is `haystack` really needed now?  I guess it is to compute the correct offset for return value.

py/objstr.c
```diff
@@ -748,23 +768,11 @@ static mp_obj_t str_finder(size_t n_args, const mp_obj_t *args, int direction, b
...
-    const byte *end = haystack + haystack_len;
-    if (n_args >= 3 && args[2] != mp_const_none) {
-        start = str_index_to_ptr(self_type, haystack, haystack_len, args[2], true);
-    }
-    if (n_args >= 4 && args[3] != mp_const_none) {
-        end = str_index_to_ptr(self_type, haystack, haystack_len, args[3], true);
-    }
-
-    if (end < start) {
-        goto out_error;
```
dpgeorge, line 762: Are you sure removing this check is correct?  Granted the tests all still pass, but maybe the test suite misses something?

For example the following doesn't currently work correctly under MP:
```
"a".index("", 2)
```
The change here doesn't fix that (and fixing that is probably out of scope of this PR), but that's an example of something that's currently not tested.

In other words, does the change here have equivalent behaviour to before the change?

glenn20: I believe it is made redundant by the MAX(l line in get_substring_data.

dpgeorge: I think I found a case which works prior to this PR, but fails with it:
```py
"abc".find("", 1, 0)
```
That would return `-1` prior to this PR because `end < start` and it does the `goto out_error`.

But with this PR it truncates the string to search to the empty string and then successfully finds the empty string within that, returning `1`.

glenn20: You're right of course, now that im back at the computer and can check. I also found another example which is broken on this PR (and also master): `"abc".find("", 5, 6)` returns `3` in both versions of MP and `-1` on cpython. The problem in this case is the use of `is_slice = true` in `str_index_to_ptr()`. I'll take a deeper look. Thanks for picking this up. I'll also add all these cases to the tests.

dpgeorge: I suggest pulling out the optimisation commit to a separate PR.  So this PR is just the first commit, improving startswith/endswith.  That change is a clear improvement and easier to argue it's "bug free".

glenn20: OK. Shall do. Some further investigation shows strange behaviour in cpython for edge cases where the search string is `""`. I'll document them in the new PR, rather than clutter up this PR.

glenn20: Done.

dpgeorge (approved):

dpgeorge:
Very good, nice enhancement, thank you!

PR #16806, opened 2025-02-24
py/objtype: Add support for PEP487 __set_name__.

AJMansfield:
### Summary

This PR adds support for the `__set_name__` data model method. This includes support for methods that mutate the owner class, and avoids the modify-while-iterating hazard on the class locals dictionary encountered by a more naive implementation like #15503. It's also able to do this _faster_ than the naive implementation, thanks to clever data layout in the linked list it uses and the way this approach allows the capture step to run during an existing step that already iterates the class dict.

### Testing

This feature was developed test-first with comprehensive tests of the feature (originally cpydiffs proposed in #15500). It also includes additional tests that were needed to guarantee test sensitivity to the modify-while-iterating hazard.

I've run the test suite against both Unix and RP2 targets. I've verified that the tests are sensitive to the feature's presence, and that this implementation passes the full test suite.

### Trade-offs and Alternatives

Performing `__set_name__` in a way that allows both owner-class mutation and avoids the modify-while-iterating hazard, requires allocating some kind of additional memory for a structure that can be iterated; but within that there are several ways this can be done.

- python/cpython#72983
During the implementation of this, it was discovered that some versions of CPython also have this naieve modify-while-iterating bug, and this issue was raised with them. This was resolved in [bpo-28797](https://bugs.python.org/issue28797); their approach is now to make a complete flat copy of the class's dict and iterate on this copy. This is a decision that doesn't make very much sense on a microcontroller, but makes all the sense on a desktop platform where memcpy might actually be cheaper than a hard-to-branch-predict conditional; and in their case it's also motivated by error-tracing considerations.
- #16816
This is an equivalent implementation to CPython's approach that places this copy directly on the stack; however it is both slower and has larger code size than the approach taken here.
- #15503
The simplest implementation is to just not worry about it and just let the user face the consequences if they mutate the owner class. It's not a very friendly behavior, though, and it's not actually much more performant than this PR on time or code size.
- #17693
Another alternative is to do the same as #15503 but leverage Micropython's existing `is_fixed` field in its dict type to convert attempted mutations of the owner dict into `AttributeError`s. This is safer than just leaving the open hazard, but there's still important use-cases for owner-mutating descriptors, and the performance gain is small enough that it isn't worth missing support for those cases.

[...]

PR #16816, opened 2025-02-25
py/objtype: Add support for __set_name__. (dict-copy version)

AJMansfield:
### Summary

This implements the alternative "flat dict copy" logic discussed in #16806 as another way to eliminate the modify-while-iterating hazard in the original `__set_name__` implementation (#15503).

### Testing

This feature was developed test-first. All automated tests pass, including additional tests added to verify the additional behavior around whether `__set_name__` is called even on descriptors that another `__set_name__` removes beforehand.

This version has significantly lower code complexity than #16806, but at the cost of a larger runtime memory footprint. It's still a bit more complex than #15503, though.

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -1124,7 +1124,7 @@ typedef time_t mp_timestamp_t;
 #define MICROPY_PY_FUNCTION_ATTRS_CODE (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_FULL_FEATURES)
 #endif
 
-// Whether to support the descriptors __get__, __set__, __delete__
+// Whether to support the descriptors __get__, __set__, __delete__, __set_name__
 // This costs some code size and makes load/store/delete of instance
 // attributes slower for the classes that use this feature
 #ifndef MICROPY_PY_DESCRIPTORS
--- a/py/objtype.c
+++ b/py/objtype.c
@@ -661,8 +661,8 @@ static void mp_obj_instance_load_attr(mp_obj_t self_in, qstr attr, mp_obj_t *des
     // try __getattr__
     if (attr != MP_QSTR___getattr__) {
         #if MICROPY_PY_DESCRIPTORS
-        // With descriptors enabled, don't delegate lookups of __get__/__set__/__delete__.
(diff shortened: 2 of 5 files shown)


AJMansfield:
Just updated this to instead make its own copy of `locals_map` into a `mp_local_alloc` in order to avoid the need to set a NLR trap handler on non-`MICROPY_ENABLE_PYSTACK` ports, but I have a small anxiety about a potential use-after-free related to garbage collection and I can't say I know enough about how gc works to be confident it can't be a problem.

@dpgeorge Does anything extra need to be done to guarantee that objects referenced in the table I'm allocating on the stack can't be garbage collected?

The `__set_name__` call returns control back to user code and the bytecode executor during `mp_call_method_n_kw(2, 0, set_name_method)`. Theoretically, it's therefore possible for `gc.collect()` to run in the middle that, either explicitly or as a result of memory pressure.

If a descriptor's `__set_name__` method were to remove an attribute from its owner, that copy of the map sitting on the stack in `run_set_name_hooks` could be the only remaining live reference to that attribute's value and/or name. If that's not enough to prevent an object from being collectable, the garbage collector could collect them, and if it happens to be later in the iteration order trying to load and call `__set_name__` on it would become a use-after-free.

Outside of this `__set_name__` implementation, I have been able to find zero examples of code needing to allocate an object (other than and then call to user code before having done _something_ that gives it root-pointer provenance.

The only thing kinda close is in `mp_obj_instance_make_new`, where you have `o = mp_obj_new_instance` followed by `mp_call_method_n_kw` -- but that object o is being passed as the self-parameter and could be picking up provenance there. Everywhere else, the allocated object is returned back to the bytecode interpreter, which grants provenance with whatever variable it assigns it to, or it's a `MP_OBJ_NEW_SMALL_INT` and `MP_OBJ_NEW_QSTR` that could never be garbage-collected anyway.

dpgeorge:
This is my least favourite of the implementations.  It's stack heavy and unconditionally allocates.

I think #16806 could be made to be simpler than this version.

AJMansfield:
Closed in favor of #16806.

PR #16820, opened 2025-02-26
tests/run-tests: Two small quality-of-life improvements for unittest.

projectgus:
### Summary

These are two small quality-of-life improvements for making new tests with `unittest`, particularly for the failure mode of forgetting the call to `unittest.main()` (which I am prone to doing):

* Print a note if it looks like a test might have forgotten `unittest.main()`.
* Clean up any `.exp` file which was generated from an early mal-formed unittest run, after it runs correctly.

### DIFF
--- a/tests/run-tests.py
+++ b/tests/run-tests.py
@@ -1004,6 +1004,16 @@ def run_tests(pyb, tests, args, result_dir, num_threads=1):
 
         test_count.increment()
 
+        # Print a note if this looks like it might have been a misfired unittest
+        if not uses_unittest and not test_passed:
+            with open(test_file, "r") as f:
+                if any(re.match("^import.+unittest", l) for l in f.readlines()):
+                    print(
+                        "NOTE: {} may be a unittest that doesn't run unittest.main()".format(
+                            test_file
+                        )
+                    )
+
     if pyb:
         num_threads = 1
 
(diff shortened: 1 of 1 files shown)


tests/run-tests.py
```diff
@@ -999,9 +999,21 @@ def run_one_test(test_file):
             with open(filename_mupy, "wb") as f:
                 f.write(output_mupy)
             failed_tests.append((test_name, test_file))
+            if uses_unittest:
+                rm_f(filename_expected)  # in case left over from previous failed run
```
dpgeorge, line 1003: It seems that `uses_unittest` is True is equivalent to `output_expected` is None.  So might be slightly better logic to put this up above as:
```py
if output_expected is not None:
    with open(filename_expected, "wb") as f:
        f.write(output_expected)
else:
    rm_f(filename_expected)
```

tests/run-tests.py
```diff
@@ -999,9 +999,21 @@ def run_one_test(test_file):
...
 
         test_count.increment()
 
+        # Print a note if this looks like it might have been a misfired unittest
+        if not (uses_unittest or test_passed):
```
dpgeorge, line 1008: This logic was a bit hard for me to grok.  Maybe better as `if not uses_unittest and not test_passed`??

tests/run-tests.py
projectgus: Ah yes, that's much cleaner. Thanks!

tests/run-tests.py
projectgus: Yeah... I paused on this when I wrote it, but wasn't sure which one reads cleaner. Have swapped to the other one!

projectgus:
Updated from review comments, ran through a test with missing `unittest.main()` to double check everything still behaved as expected.

dpgeorge (approved):

PR #16821, opened 2025-02-26
py/objstr: Add str.removesuffix() and str.removeprefix() methods.

glenn20:
### Summary

Add `str.removesuffix()` and `str.removeprefix()` methods to the `str` type. 

These methods are used to remove a suffix or prefix from a string if it exists. If the suffix or prefix does not exist in the string, the original string is returned.

This provides compatibility with [Python 3.9+](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) and supports [ruff rule FURB188 ](https://docs.astral.sh/ruff/rules/slice-to-remove-prefix-or-suffix) (thanks @cclauss).

The methods are conditionally compiled based on the `MICROPY_PY_BUILTINS_STR_REMOVESUFFIX` C macro (which is enabled by default in `MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_EXTRA_FEATURES` builds).

### Testing

Tests are provided in `tests/basics/string_remove{pre,suf}fix.py`. These have been tested and passing in the unix port.

This PR provides improved compatibility with cpython (and ruff code rules) at the expense of increased code size. Both methods share a common implementation to reduce the impact on code size.


dpgeorge:
Thanks for making this PR.

It's a lot of code size for a relatively obscure (?) or at least new feature.  Since these methods were added in Python 3.9 there's not that much code out there using them.

Are these necessary for MicroPython to have?  Apart from ruff having a rule that suggests to use these methods, is there any other motivation for adding them?

I guess it doesn't hurt to add them but under a high feature level like FULL_FEATURES.  Although that means users can't really rely on them to write portable code (portable to many MicroPython platforms).

Relevant PEP: https://peps.python.org/pep-0616/

glenn20:
> Thanks for making this PR.
> 
> It's a lot of code size for a relatively obscure (?) or at least new feature. Since these methods were added in Python 3.9 there's not that much code out there using them.

I'm happy to not merge this PR as I agree that the pros/cons weighs against inclusion. I built it, as it was an easy follow-on from #16812 (Support tuples as pre/suffix in startswith() and endswith()) and was suggested by @cclauss. I couldn't find a more compact implementation in the end and thought I'd just defer the decision to you ;).

I agree that its probably not worth including if it is not going to be available on the great majority of ports.

I'm happy for you to close this PR if you prefer.

PR #16825, opened 2025-02-26
tests/internal_bench: Benchmarks for descriptor-related features.

AJMansfield:
### Summary

This includes new `internal_bench` benchmarks for how quickly micropython can process class definitions, created in order to gauge the performance penalty for implementing `__set_name__` in the different ways proposed in #15503, #16806, and #16816.

It also includes an update to `run-internalbench.py` that adds in the same ability to run the benchmarks on remote micropython instances (e.g. actual microcontroller hardware) that the main `run-tests.py` has.

### Testing

I've tested these benchmarks using the unix port inside WSL on my development machine, with the following results:

<pre><code><b> . . .
internal_bench/class_create:</b>
    <b>0.351s (+00.00%) internal_bench/class_create-0-empty.py</b>
    <b>0.476s (+35.91%) internal_bench/class_create-1-slots.py</b>
    <b>0.484s (+38.01%) internal_bench/class_create-1.1-slots5.py</b>
    <b>0.427s (+21.87%) internal_bench/class_create-2-classattr.py</b>
    <b>0.773s (+120.54%) internal_bench/class_create-2.1-classattr5.py</b>
    <b>0.458s (+30.62%) internal_bench/class_create-3-instancemethod.py</b>
    <b>0.500s (+42.76%) internal_bench/class_create-4-classmethod.py</b>
    <b>0.462s (+31.65%) internal_bench/class_create-4.1-classmethod_implicit.py</b>
    <b>0.498s (+41.93%) internal_bench/class_create-5-staticmethod.py</b>
    <b>0.456s (+29.97%) internal_bench/class_create-6-getattribute.py</b>
    <b>0.467s (+33.13%) internal_bench/class_create-6.1-getattr.py</b>
    <b>0.388s (+10.58%) internal_bench/class_create-6.2-descriptor.py</b>
    <b>0.529s (+50.89%) internal_bench/class_create-6.3-descriptor_setname.py</b>
    <b>0.425s (+21.22%) internal_bench/class_create-6.4-property.py</b>
    <b>0.368s (+05.02%) internal_bench/class_create-7-inherit.py</b>
    <b>0.359s (+02.35%) internal_bench/class_create-7.1-inherit_initsubclass.py</b>
 . . .
internal_bench/var:
    0.459s (+00.00%) internal_bench/var-1-constant.py
    0.620s (+35.05%) internal_bench/var-2-global.py
    0.443s (-03.42%) internal_bench/var-3-local.py
    0.441s (-03.96%) internal_bench/var-4-arg.py
    1.131s (+146.39%) internal_bench/var-5-class-attr.py
    0.539s (+17.34%) internal_bench/var-6-instance-attr.py
    0.555s (+20.85%) internal_bench/var-6.1-instance-attr-5.py
    <b>0.537s (+17.04%) internal_bench/var-6.2-instance-speciallookup.py</b>
    <b>2.436s (+430.54%) internal_bench/var-6.3-instance-property.py</b>
    <b>3.102s (+575.63%) internal_bench/var-6.4-instance-descriptor.py</b>
    <b>3.144s (+584.85%) internal_bench/var-6.5-instance-getattr.py</b>
    2.422s (+427.62%) internal_bench/var-7-instance-meth.py
    0.865s (+88.46%) internal_bench/var-8-namedtuple-1st.py
    0.924s (+101.30%) internal_bench/var-8.1-namedtuple-5th.py
 . . .
</code></pre>

Due to the overall slow speed of class creation in general, the `class_create` tests reduce the number of iterations from 20,000,000 to just 500,000; but this should still be wide enough to give the results adequate statistical significance.

For microcontroller remotes it uses 200,000 as the base default (which in the class_create case gets divided down to 5,000 runs).

### DIFF
--- a/tests/run-internalbench.py
+++ b/tests/run-internalbench.py
@@ -8,6 +8,10 @@ import re
 from glob import glob
 from collections import defaultdict
 
+run_tests_module = __import__("run-tests")
+sys.path.append(run_tests_module.base_path("../tools"))
+import pyboard
+
 if os.name == "nt":
     MICROPYTHON = os.getenv(
         "MICROPY_MICROPYTHON", "../ports/windows/build-standard/micropython.exe"
@@ -15,13 +19,39 @@ if os.name == "nt":
 else:
     MICROPYTHON = os.getenv("MICROPY_MICROPYTHON", "../ports/unix/build-standard/micropython")
 
+injected_bench_code = b"""
+import time
 
-def run_tests(pyb, test_dict):
+class bench_class:
+    ITERS = 20000000
+
+    @staticmethod
+    def run(test):
+        t = time.ticks_us()
+        test(bench_class.ITERS)
+        t = time.ticks_diff(time.ticks_us(), t)
+        s, us = divmod(t, 1_000_000)
+        print("{}.{:06}".format(s, us))
+
+import sys
+sys.modules['bench'] = bench_class
+"""
+
+
+def execbench(pyb, filename, iters):
+    with open(filename, "rb") as f:
+        pyfile = f.read()
+    code = (injected_bench_code + pyfile).replace(b"20000000", str(iters).encode("utf-8"))
+    return pyb.exec(code).replace(b"\r\n", b"\n")
+
+
+def run_tests(pyb, test_dict, iters):
     test_count = 0
     testcase_count = 0
 
     for base_test, tests in sorted(test_dict.items()):
         print(base_test + ":")
+        baseline = None
         for test_file in tests:
             # run MicroPython
             if pyb is None:
@@ -36,20 +66,25 @@ def run_tests(pyb, test_dict):
                 # run on pyboard
                 pyb.enter_raw_repl()
                 try:
-                    output_mupy = pyb.execfile(test_file).replace(b"\r\n", b"\n")
+                    output_mupy = execbench(pyb, test_file[0], iters)
                 except pyboard.PyboardError:
                     output_mupy = b"CRASH"
 
-            output_mupy = float(output_mupy.strip())
+            try:
+                output_mupy = float(output_mupy.strip())
+            except ValueError:
+                output_mupy = -1
             test_file[1] = output_mupy
             testcase_count += 1
 
-        test_count += 1
-        baseline = None
-        for t in tests:
             if baseline is None:
-                baseline = t[1]
-            print("    %.3fs (%+06.2f%%) %s" % (t[1], (t[1] * 100 / baseline) - 100, t[0]))
+                baseline = test_file[1]
+            print(
+                "    %.3fs (%+06.2f%%) %s"
+                % (test_file[1], (test_file[1] * 100 / baseline) - 100, test_file[0])
+            )
+
+        test_count += 1
 
     print("{} tests performed ({} individual testcases)".format(test_count, testcase_count))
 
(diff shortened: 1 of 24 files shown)


AJMansfield:
It looks like the unix port / settrace_stackless test failure is spurious? It failed in `thread/thread_gc1.py`, which should have nothing to do with anything this PR does.

tests/internal_bench/bench_hw.py
```diff
@@ -0,0 +1,12 @@
+import utime
```
dpgeorge: I suggest using just `import time`.  We are trying to move away from u-modules, and `utime` is rarely used now.

tests/internal_bench/bench_hw.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+ITERS = 200000
+
+
+def run(f):
+    t = utime.ticks_us()
+    f(ITERS)
+    t = utime.ticks_us() - t
```
dpgeorge: Need to use `ticks_diff`, or else this could come out negative.

tests/internal_bench/bench_hw.py
```diff
@@ -0,0 +1,12 @@
...
+    s, us = divmod(t, 1_000_000)
+    print(f"{s}.{us:06}")
```
dpgeorge: Please use `.format(...)`.  Not all targets enable f-strings.

tests/run-internalbench.py
```diff
@@ -8,6 +8,13 @@
 from glob import glob
 from collections import defaultdict
 
+import importlib
+
+run_tests_py = importlib.import_module("run-tests")
```
dpgeorge: Suggest using the pattern that other test runners use:
```py
run_tests_module = __import__("run-tests")
```
(No dependency on `importlib`.)

tests/run-internalbench.py
```diff
@@ -58,27 +71,70 @@ def run_tests(pyb, test_dict):
...
+  MICROPY_MICROPYTHON_MJS environment variable (which defaults to the standard
+  variant of the webassembly port)
+- port:<device> - connect to and use the given serial port device
+- a<n> - connect to and use /dev/ttyACM<n>
+- u<n> - connect to and use /dev/ttyUSB<n>
+- c<n> - connect to and use COM<n>
+- exec:<command> - execute a command and attach to its stdin/stdout
+- execpty:<command> - execute a command and attach to the printed /dev/pts/<n> device
+- <a>.<b>.<c>.<d> - connect to the given IPv4 address
+- anything else specifies a serial port
```
dpgeorge: Suggest making this string a variable in `run-tests.py` that can be used in that script and here as well, to reduce duplication of this blob of text (and make it easier to update in the future).

tests/run-internalbench.py
```diff
@@ -58,27 +71,70 @@ def run_tests(pyb, test_dict):
...
+
+Options -i and -e can be multiple and processed in the order given. Regex
+"search" (vs "match") operation is used. An action (include/exclude) of
+the last matching regex is used:
+  run-tests.py -i async - exclude all, then include tests containing "async" anywhere
+  run-tests.py -e '/big.+int' - include all, then exclude by regex
+  run-tests.py -e async -i async_foo - include all, exclude async, yet still include async_foo
```
dpgeorge: This should be `run-internalbench.py`.

tests/run-internalbench.py
```diff
@@ -95,6 +151,10 @@ def main():
             continue
         test_dict[m.group(1)].append([t, None])
 
+    if pyb is not None:
+        print("Uploading bench harness.")
+        pyb.fs_put('internal_bench/bench_hw.py', 'bench.py')
```
dpgeorge: This is a reasonable way to do it, copy the file.  But it does require a writable filesystem on the target, for the current directory to be in the path, and it will be a bit slow due to filesystem writing.

An alternative way is to prepend the harness code to the script before sending it to the target.  Something like this should work:
```py
class bench_class:
    def run(test):
        ...
import sys
sys.modules['bench'] = bench_class
< here goes the actual benchmark test >
```

dpgeorge (commented):
Thanks for this, it's a nice addition, quite a comprehensive set of tests.  Also great to be able to now run it on a remote target.

I noticed that tests like `internal_bench/var-1-constant.py` fail on remote targets because they take too long, about 1 minute.  I'd be OK to just leave that for now, unless you want to try and fix it.

tests/run-internalbench.py
```diff
@@ -58,27 +71,70 @@ def run_tests(pyb, test_dict):
...
     if len(args.files) == 0:
-        if pyb is None:
+        if args.test_dirs:
+            test_dirs = tuple(args.test_dirs)
+        elif pyb is None:
             # run PC tests
             test_dirs = ("internal_bench",)
         else:
             # run pyboard tests
             test_dirs = ("basics", "float", "pyb")
```
dpgeorge: This bit can be removed, it never worked.

The default `test_dirs` should always just be `("internal_bench",)` regardless of whether `pyb` is `None` or not.

Then, it should be that doing `./run-internalbench.py -t a0` works and runs all the internal bench tests.

AJMansfield:
> I noticed that tests like `internal_bench/var-1-constant.py` fail on remote targets because they take too long, about 1 minute.

Yeah, I figured it might just be a timeout --- the root cause here is actually just that the test functions in the `var-` tests can't just use the `num` parameter since the feature under test is actually the way that their loop bound is represented.

Fixing this probably means needing to rewrite them to generate the code under test with a string interpolation and then `eval`-ing the resulting string.

AJMansfield:
Ok actually I found a much dumber way of making the var tests run --- I can just have the runner string-replace `"20000000"` in the benchmark code with the actual number of iterations.

Probably an even worse syntactic sin than trying to parse HTML with regex --- but if you think about it, you can just define a sub-language with a grammar that only accepts the specific benchmarks that are actually in the folder as valid, and at least over _that_ sub-language the literal string replacement is a correct parse.

tests/run-internalbench.py
AJMansfield: Really it should just be removed, since I didn't actually add the `-i` and `-e` flags...

dpgeorge: Yeah, happy to limit the scope of these changes.  Can just add the `-t` option if that's easiest.

tests/run-internalbench.py
AJMansfield: Then I shall remove it!

dpgeorge (approved):

PR #16828, opened 2025-02-27
shared/runtime/pyexec: Add a function to execute vstr.

iabdalkader:
### Summary

Just a simple function to execute a script from a vstr.

### Testing

Example test:
```C
vstr_t script_buf;
vstr_init(&script_buf, 1024);

vstr_printf(&script_buf, "print('HelloWorld')");

pyexec_str(&script_buf, true);
```

While we're at it, I noticed that `pyexec_file` is not used anywhere in the codebase. Should it be removed or made static?

### DIFF
--- a/shared/runtime/pyexec.c
+++ b/shared/runtime/pyexec.c
@@ -720,6 +720,11 @@ int pyexec_frozen_module(const char *name, bool allow_keyboard_interrupt) {
 }
 #endif
 
+int pyexec_vstr(vstr_t *str, bool allow_keyboard_interrupt) {
+    mp_uint_t exec_flags = allow_keyboard_interrupt ? 0 : EXEC_FLAG_NO_INTERRUPT;
+    return parse_compile_execute(str, MP_PARSE_FILE_INPUT, exec_flags | EXEC_FLAG_SOURCE_IS_VSTR);
+}
+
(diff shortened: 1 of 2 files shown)


shared/runtime/pyexec.c
```diff
@@ -720,6 +720,11 @@ int pyexec_frozen_module(const char *name, bool allow_keyboard_interrupt) {
 }
 #endif
 
+int pyexec_str(vstr_t *str, bool allow_keyboard_interrupt) {
```
dpgeorge, line 723: I suggest calling it `pyexec_vstr`, in case one day we do add a function that takes a null-terminated string (although that would probably be called `pyexec_cstr`).

dpgeorge:
> I noticed that `pyexec_file` is not used anywhere in the codebase. Should it be removed or made static?

But... the same could be said about this new function you want to add :smile: 

I think we should keep it.  It's used by `pyexec_file_if_exists()` anyway.

shared/runtime/pyexec.c
iabdalkader: Renamed to `pyexec_vstr`

iabdalkader:
> But... the same could be said about this new function you want to add 😄

Fair enough :)

dpgeorge (approved):

PR #16841, opened 2025-03-01
Add board configuration for HEIA-FR PiCoMo V2 and V3

supcik:
### Summary

This is the configuration for the PiCoMo V2 and V3 boards made at HEIA-FR.

### Testing

I successfully flashed and booted the PiCoMo with the generated firmware.


dpgeorge:
Thanks for the contribution.

Are these boards available for the public to purchase?  If not, then it's a bit hard to justify adding the board definition, because it takes resources to maintain, build and host it.

supcik:
Thank you for your comment. I understand you point. The board is given for free during events that we host at our university. We already gave 200 boards, and we have 300 ready to be distributed this year. I was able to add this board to CircuitPython and I thought that it would be nice to have it for MicroPython as well.

supcik:
I will do a dedicated repo for my boards.

mattytrentini:
Hope you don't mind me linking to the project @supcik, for those that might be interested in what an external board definition can look like:

https://github.com/heia-picomo/micropython-picomo/

dpgeorge:
> I will do a dedicated repo for my boards.

That's a great way to do it, and then you can freeze in all your additional libraries.  Very good!

PR #16849, opened 2025-03-03
extmod/vfs: Add functions for querying mount information.

AJMansfield:
### Summary

This adds functions to the python API for obtaining information about the mounted status of directories, following a discussion with @lukaskremla in the discord about a use-case they have.

### Testing

This PR includes automated tests of the added functionality, and I have verified that these and all other tests pass with this patch applied.

This implementation uses `mp_obj_new_attrtuple` rather than a raw tuple for the return value. With QSTRs this is a low-cost abstraction, but it's still technically one additional `mp_obj_t` worth of space that has to be allocated compared to using  `mp_obj_new_tuple`.


AJMansfield:
> But, the implementation needs to be minimal. This does not really need to be user friendly because most likely it will be wrapped in some other tool/code which provides a higher-level user interface, eg list mounted filesystems and their type and size.
> 
> Instead of three different functions that do very similar things, one function should be plenty, if not zero 😄
> 
> I suggest reusing the existing `vfs.mount()` function and if it is passed just one argument then that is argument is a path and used to query the vfs at that path. Essentially your `mountinfo()` function but merged with `mount()`.

If anything, it feels like `listmounts` is the more relevant piece of functionality -- but it can definitely be made more minimal by removing the user-facing iterator and making the list output the result of zero-arguments `mount`. I'll implement this and submit a new PR with this alternative approach.

dpgeorge:
This functionality was implemented by #16939.

PR #16854, opened 2025-03-03
nrf/modules: Fix access of read-only buffer in Flash.writeblocks.

dpgeorge:
### Summary

When writing to flash, the source buffer only needs to be read-only, not writable.  This fix allows passing in `bytes` and other read-only buffer objects.

### Testing

Tested on `ARDUINO_NANO_33_BLE_SENSE`.  Prior to the fix passing in a `bytes` object would raise an exception that an object with the buffer protocol is required, which was very confusing.  With the fix it allows a `bytes` object.


robert-hh:
The change looks obvious and good, but .. I tried the before-after test. And even without that PR, the following sequence passes:
```
b=bytes(100)
f = open("test", "w")
f.write(b)
f.close()
```
Creating a 100 bytes file. `type(b)` indicates it as bytes object. So what causes the error?

robert-hh (approved):
The change looks good. I only had problems finding an example to verify the changed behavior.

robert-hh:
> This came up while testing ROMFS on nrf.

That what I assumed.

dpgeorge:
Thanks for the quick review!

PR #16858, opened 2025-03-04
unix/coverage: Increase coverage of mp_binary_set_val_array_from_int.

projectgus:
### Summary

This function is only used from stm32/adc.c, but for now expand the manual coverage to include more of the supported typecodes.

Split out from #16311 where I originally added it to try and pass the project coverage CI check, but it actually wasn't enough. 

* As this function is only called from stm32/adc.c in MicroPython then we could totally remove it from the binary API, make it a static function in adc.c, and then it wouldn't need specific coverage code either.

### DIFF
--- a/ports/unix/coverage.c
+++ b/ports/unix/coverage.c
@@ -538,13 +538,32 @@ static mp_obj_t extra_coverage(void) {
     {
         mp_printf(&mp_plat_print, "# binary\n");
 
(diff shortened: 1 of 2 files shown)


projectgus:
Closing this as it seems like #16311 will either support removing `mp_binary_set_val_array_from_int` or it'll turn out we need it...

PR #16863, opened 2025-03-05
natmod: Allow linking static libraries.

agatti:
### Summary

This is an update to #15838 - as requested in https://github.com/micropython/micropython/pull/15838#issuecomment-2696048612.

All the good stuff is from @vshymanskyy, the not-so-good stuff is mine:

* Addressed the review points raised in the original PR (cache salt key, linker script detection logic, long argument for -l, some text changes, ESP-IDF CI caching)
* Picolibc is now supported for RV32 via a rather nasty workaround
* Xtensa natmods are now built as part of the CI process
* Natmods have the appropriate platform exceptions in the CI build step to make it work out of the box for most platforms
* The object cache directory is removed as part of `make clean`
* The object cache directory now contains a `CACHEDIR.TAG` file, so most archival/backup programs will skip that directory (if rust can do it why can't we?)
* Removed the assumption that there's a `python` executable in the CI environment (minor regression from 8b1ed4473dd215d564498f577cb6f3339c3c2e57), blocking manual testing on my provisioned build VMs :)

### Testing

All natmod-related CI jobs ran without issues on my own repo's branch.

### DIFF
--- a/tools/ar_util.py
+++ b/tools/ar_util.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python3
+#
+# This file is part of the MicroPython project, http://micropython.org/
+#
+# The MIT License (MIT)
+#
+# Copyright (c) 2024 Volodymyr Shymanskyy
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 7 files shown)


agatti:
For the ESP32 target, successful build run logs can be found here https://github.com/agatti/micropython/actions/runs/13666850878/job/38209736600 and here https://github.com/agatti/micropython/actions/runs/13666850878/job/38209737465

The commit checker fails to see the sign-off line, but I guess fixing that is its own PR.

tools/ar_util.py
```diff
@@ -0,0 +1,239 @@
...
+    # Woah, some .a files are actually ld scripts!
+    def is_ld_script():
+        with open(fn, "rb") as f:
+            header = b"/* GNU ld script"
+            return header == f.read(len(header))
+
+    if not is_ld_script():
+        return [CachedArFile(fn)]
+    else:
+        return [CachedArFile(item) for item in expand_ld_script(fn)]
```
dpgeorge, line 239: I see you inverted the logic here by swapping the if/else.  But I think @projectgus had a different idea in mind in his comment https://github.com/micropython/micropython/pull/15838#discussion_r1942090278

I think what was intended was:
```py
ar_header = "!<arch>\012"
with open(fn, "rb") as f:
    is_ar_file = f.read(len(ar_header)) == ar_header

if is_ar_file:
    return [CachedArFile(fn)]
else:
    return [CachedArFile(item) for item in expand_ld_script(fn)]
```

dpgeorge (commented):
Thanks @agatti for putting this PR together and tidying up the final bits.  I have just one comment above, but otherwise this looks ready to go in.

tools/ar_util.py
agatti: I see that now - I'm on it.  I was in a bit of a hurry back then to get the RV32 stuff working (I assumed it worked out of the box like when I tried it last time).

agatti:
This should be it - since last time you've looked at the code I've added the proper logic reversal for the magic bytes check, the platform name comparisons in the mpy build CI step now use explicit strings (as in quoted platform names), and I've added a message to the original PR's commit explaining the picolibc workaround and providing a bit more information about the change itself.

dpgeorge (approved):
Thanks for updating!

PR #16864, opened 2025-03-05
ports/zephyr: Add machine lightsleep.

ThinkTransit:
### Summary

Machine lightsleep is currently not implemented in the zephyr port.

nrf socs have excellent low power capabilities however they do require some additional PM setup. In particular the uart module consumes ~1mA of power even when not being used. Powering down the uart modue can reduce sleep current to as low as 4uA (depending on hardware). 

This PR adds support for lightsleep in the zephyr port. It disables the main uart and calls k_sleep() to put the micropython thread to sleep. This allows Zephyr power management to reduce power to minimum. The uart is made active when the thread is resumed. 

lightsleep is included by default on all boards but can be disabled by setting:

```
CONFIG_PM=n
CONFIG_PM_DEVICE=n
```

Note: It is disabled on the beagleconnect_freedom as it does not seem to be supported.

### Testing

Tested on nrf5340dk with power profile kit.

```python
MicroPython b4cf82b2d-dirty on 2025-03-05; zephyr-nrf5340dk with nrf5340
>>> import machine
>>> machine.lightsleep(1000)
>>> 
```

image
Current reduced from idle 1mA to 4uA during lightsleep.

Confirmed that ram is retained as this is lightsleep

```python
MicroPython b4cf82b2d-dirty on 2025-03-05; zephyr-nrf5340dk with nrf5340
>>> import machine
>>> my_var="123456789"
>>> machine.lightsleep(1000)
>>> my_var
'123456789'
>>> 
```
Firmware has been built for all board variants in the mp repo. All boards were sucsefully built except for:

qemu_cortex_m3 - Failed due to unrelated issue
beagleconnect_freedom - Does not seem to support PM so it has been explicitly disabled in prj.conf

```
CONFIG_PM=n
CONFIG_PM_DEVICE=n
```

I don't believe there are any negative impacts of this change although I am not sure how to measure the increase in firmware size.

**machine.deepsleep**, **machine.get_freq**, **machine.set_freq** and **machine.unique_id** return "not implemented".


jonnor:
This looks like a very useful addition to the Zephyr port. Thanks!

VynDragon:
I think, instead of grabbing the console device and putting it to sleep, it should either rely on zephyr's power policy management or implement pm_policy_next_state, or provide a API to chose what devices to turn on and off.

ThinkTransit:
> I think, instead of grabbing the console device and putting it to sleep, it should either rely on zephyr's power policy management or implement pm_policy_next_state, or provide a API to chose what devices to turn on and off.
> 
> Otherwise this looks good.

Thanks @VynDragon 

I think the uart/console has to be manually shutdown on the nrf chips regardless of PM policy.

My understanding is that when CONFIG_SERIAL=y then serial logging is enabled for debugging which uses power, even in low power modes. We can't turn off CONFIG_SERIAL because it's needed for micropython repl.

There is some info on it here.

[Power Optimisation](https://docs.nordicsemi.com/bundle/ncs-2.1.3/page/nrf/app_power_opt.html)

You can see also in the power management examples they shut down the console device manually before power off.

[System off sample](https://github.com/nrfconnect/sdk-zephyr/blob/4040aa0bf581206dd68beb0758b39096342ee7e0/samples/boards/nrf/system_off/src/main.c#L63)

I'm very new to zephyr so I could be wrong, what do you think?

Also looking at the example if might be worthwhile also implementing deepsleep with sys_poweroff();?

ThinkTransit:
I have added machine.deepsleep() to this PR which completely powers down the system using ```sys_poweroff();```

ports/zephyr/modmachine.c
```diff
@@ -61,3 +64,52 @@ MP_DEFINE_CONST_FUN_OBJ_0(machine_reset_cause_obj, machine_reset_cause);
...
+static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
+    // Check if no argument is provided
+    if (n_args == 0) {
+        mp_raise_ValueError(MP_ERROR_TEXT("value must be provided"));
+    }
+    mp_int_t milliseconds = mp_obj_get_int(args[0]);
+    // Get the UART device
+    const struct device *uart0 = DEVICE_DT_GET(DT_CHOSEN(zephyr_console));
+    // Set UART device to low power state
+    pm_device_action_run(uart0, PM_DEVICE_ACTION_SUSPEND);
```
dpgeorge, line 78: Will this work on boards that use USB CDC as the console?  Maybe the code here will still be OK, but at least the variable should be renamed from `uart0` to `console`?

ports/zephyr/modmachine.c
```diff
@@ -61,3 +64,52 @@ MP_DEFINE_CONST_FUN_OBJ_0(machine_reset_cause_obj, machine_reset_cause);
...
+static void mp_machine_deepsleep(size_t n_args, const mp_obj_t *args) {
+    // Check if argument is provided
+    if (n_args > 0) {
+        mp_raise_ValueError(MP_ERROR_TEXT("timeout not supported"));
+    }
+    // Get the UART device
+    const struct device *uart0 = DEVICE_DT_GET(DT_CHOSEN(zephyr_console));
+    // Set UART device to low power state
+    pm_device_action_run(uart0, PM_DEVICE_ACTION_SUSPEND);
+    sys_poweroff();
```
dpgeorge, line 93: It looks like this is not always available, eg the CI is failing here.

dpgeorge:
I tested this on nucleo_wb55rg, and it breaks the UART REPL on that board: the UART responds only occasionally to input (output is fine though).  I'm guessing it's because power management is now enabled and the UART is mostly asleep?

I also tested it on frdm_k64f and rpi_pico.  `machine.lightsleep(2000)` waits 2 seconds but doesn't seem to change the power consumption at all on those boards, maybe they don't implement low power?

If we want to move forward with this, I suggest only enabling CONFIG_PM/CONFIG_PM_DEVICE on devices where it's known to work.

ThinkTransit:
> I tested this on nucleo_wb55rg, and it breaks the UART REPL on that board: the UART responds only occasionally to input (output is fine though). I'm guessing it's because power management is now enabled and the UART is mostly asleep?
> 
> I also tested it on frdm_k64f and rpi_pico. `machine.lightsleep(2000)` waits 2 seconds but doesn't seem to change the power consumption at all on those boards, maybe they don't implement low power?
> 
> If we want to move forward with this, I suggest only enabling CONFIG_PM/CONFIG_PM_DEVICE on devices where it's known to work.

Hi Damien,

Apologies for the late reply.

I did some more research into this and it turns out that this high power usage on the UART is only related to nordic chips, something to do with the HFCLK being required when the uart is enabled.

Did enabling CONFIG_PM break the repl or was it calling lightsleep()?

Is there a mechanism where we could include the feature on just the nrf family? Or would if be better to add an nrf module with uart_sleep function?

VynDragon:
> Is there a mechanism where we could include the feature on just the nrf family? Or would if be better to add an nrf module with uart_sleep function?

if you make the config proper on zephyr side, UART will go to sleep automaticlaly when calling k_sleep or power off

ThinkTransit:
> if you make the config proper on zephyr side, UART will go to sleep automaticlaly when calling k_sleep or power off

Not for the nordic chips unfortunately, it's a know limitation that requires the uart to be powered down manually at the discretion of the application.

https://devzone.nordicsemi.com/f/nordic-q-a/101872/how-to-make-zephyr-power-management-unit-work-with-nrf52

VynDragon:
~~If it is not standard within zephyr even, why add it to micropython code?~~

Sounds like that's true for all UART on all platforms though, so I'd be wrong. Regardless there should be some way to disable that right? It seems @dpgeorge 's issue is exactly what's described as happeneing when you turn off uart :)

RodLaird:
Seems this machine.lightsleep() accessible in Micropython 1.26 beta?. On a pico 2 at least I get a response from the function. However the time delay seems a bit random as / and seems to turn off the internal RTC. Cannot be totally sure as testing with Thonny so UART kicked into action by interaction. But when I set up some delays and print(time.time()) I get no change in the time. And I cannot set up predictable delay loops. I would just like to be able to sleep the machine and reboot it reliably after a relatively accurate time delay... Still a bridge too far it seems..

docs/zephyr/quickref.rst
```diff
@@ -119,6 +119,36 @@ Hardware SPI is accessed via the :ref:`machine.SPI <machine.SPI>` class::
...
+
+Deep-sleep mode
+----------------
+
+The following code can be used to power down the system::
+
+    import machine
+
+    # power down the system
+    machine.depsleep()
```
jonnor, line 144: Typo here. Should be deepsleep()

bikeNomad:
I just created PR #18197 that allows `Pin` IRQs to interrupt `machine.lightsleep()`

ant9000:
> Did enabling CONFIG_PM break the repl or was it calling lightsleep()?

CONFIG_PM_DEVICE breaks the REPL.  I have a different board (an Apollo3 from Ambiq), and tried to investigate this stuff directly at Zephyr level, starting with the [shell](https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/subsys/shell/shell_module) sample and adding PM configuration. On my board, the working config is the following:

```
CONFIG_PM=y
CONFIG_PM_DEVICE=y
CONFIG_PM_DEVICE_RUNTIME=y
```

Skipping the last line makes the UART unresponsive to user input (it does print messages, though). It seems that there is nothing that marks the REPL (or the Zephyr shell) as making use of the console. Adding the following to `main` is enough to make the shell usable even without `CONFIG_PM_DEVICE_RUNTIME`:

```
    const struct device *dev;
    dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_console));
    pm_device_busy_set(dev);
```

I think this line would solve the REPL issue also for MicroPython; but then, before going to sleep you would need to relinquish the console with a corresponding `pm_device_busy_clear()`.

ant9000:
I can confirm that the solution above works also for MicroPython on the board I have. Here is my patch:

```
diff --git a/ports/zephyr/main.c b/ports/zephyr/main.c
index 8d39935884..ef1c7590ae 100644
--- a/ports/zephyr/main.c
+++ b/ports/zephyr/main.c
@@ -89,6 +89,10 @@ static __noinit char heap[MICROPY_HEAP_SIZE];
 extern int mp_usbd_init(void);
 #endif // defined(CONFIG_USB_DEVICE_STACK_NEXT)
 
+#if CONFIG_PM_DEVICE
+#include <zephyr/pm/device.h>
+#endif
+
 void init_zephyr(void) {
     // We now rely on CONFIG_NET_APP_SETTINGS to set up bootstrap
     // network addresses.
@@ -163,6 +167,14 @@ soft_reset:
     }
     #endif
 
+#if CONFIG_PM_DEVICE
+    const struct device *dev;
+    dev = DEVICE_DT_GET(DT_CHOSEN(zephyr_console));
+    if (dev != NULL) {
+        pm_device_busy_set(dev);
+    }
+#endif
+
     for (;;) {
         if (pyexec_mode_kind == PYEXEC_MODE_RAW_REPL) {
             if (pyexec_raw_repl() != 0) {

```

PR #14145, opened 2024-03-21
esp32: Workaround for ringbuf functions required by ISRs.

jimmo:
See https://github.com/espressif/esp-idf/issues/13378

Workaround for #14005 (cc @Tangerino)

Note that this means that _all_ esp32-og builds get non-isr ringbuf functions placed in flash now, whereas previously it was just the spiram variant. This might be a good thing (e.g. free up some iram for native/viper).

_This work was funded through GitHub Sponsors._

PR #16865, opened 2025-03-05
esp32: Merge the per-SoC main components, remove linker workaround.

projectgus:
### Summary

I noticed that IDF Component Manager has a conditional dependencies language that allows us to shrink the per-target `main_espXYZ` components back down to a single `main` component.

Split into two commits:

1. Combines the main components but keeps the `linker.lf` workaround added in #14145 (but with a conditional to only do anything on ESP32).
2. Removes the `linker.lf` workaround entirely, as the [upstream issue](https://github.com/espressif/esp-idf/issues/13378) apparently only existed for ESP-IDF V5.1 which MicroPython no longer supports. The support for a user-specified linker fragment (added in #16658) is kept, although it may not have a clear use case any more.

### Testing

* With only commit 1 applied, built multiple chips and verified application code size remained the same (Xtensa targets varied by 32 bytes and RISC-V targets by 16 bytes, I think because the final linker script was subtly different and changed padding.)
* Ran unit tests on ESP32 board
* Ran a build with the Espressif IDF V5.2.2 container, to check this works OK with the older Component Manager version (it does).

* Keeping the individual main components is arguably more flexible and simpler, but it's also a bunch of duplicated info (that adds more duplicates each time we add a new SoC).
* Out-of-tree builds may need to update their main component, at least the ones which use RISC-V.


ports/esp32/esp32_common.cmake
```diff
@@ -14,6 +14,18 @@ if(NOT MICROPY_PORT_DIR)
...
+# RISC-V specific inclusions
+if(CONFIG_IDF_TARGET_ARCH_RISCV)
+    list(APPEND MICROPY_SOURCE_LIB ${MICROPY_DIR}/shared/runtime/gchelper_generic.c)
+    list(APPEND IDF_COMPONENTS riscv)
+endif()
+
+if(NOT DEFINED MICROPY_PY_TINYUSB)
+  if(CONFIG_IDF_TARGET_ESP32S2 OR CONFIG_IDF_TARGET_ESP32S3)
+    set(MICROPY_PY_TINYUSB ON)
+  endif()
```
dpgeorge, line 26: This needs to be indented to 4 spaces.

dpgeorge (commented):
Thanks, this is a very nice simplification!

I noticed that `CONFIG_RINGBUF_PLACE_FUNCTIONS_INTO_FLASH` is still commented out, and I think it should now be uncommented, ie enabled.  That will claw back a bit of IRAM.

projectgus:
> I noticed that `CONFIG_RINGBUF_PLACE_FUNCTIONS_INTO_FLASH` is still commented out, and I think it should now be uncommented, ie enabled. That will claw back a bit of IRAM.

Oops, yes! It's restored now.

PR #16869, opened 2025-03-06
stm32: On Arduino boards, enable 4MiB ROMFS partition in external flash

dpgeorge:
### Summary

Split out from #8381, and a follow up to https://github.com/micropython/micropython/pull/16857#issuecomment-2699648257

This enables a 4MiB ROMFS partition on all three stm32 Arduino boards, using external QSPI flash.  The ROMFS partition sits near the end of flash:
- 11MiB filesystem
- 4 MiB ROMFS
- 1 MiB WiFi firmware (compatibility with alternate firmware)

### Testing

TODO (although previously tested by @iabdalkader as part of #8381).

This is most likely a breaking change for users, because the filesystem partition size has shrunk by 4MiB.  And I don't think that will be easily detected because existing filesystems will mount and work correctly until the ROMFS is used.  And then when ROMFS is used and the flash for that is written to, the original filesystem will probably still mount correctly because the initial filesystem data is unchanged.

@iabdalkader how do you want to manage this when users update the firmware?  This is part of the reason why I didn't want to enable ROMFS on all the boards right now, because we don't have a good way of transitioning the flash partition from previous layout to new layout.

### DIFF
--- a/ports/stm32/boards/ARDUINO_GIGA/mpconfigboard.h
+++ b/ports/stm32/boards/ARDUINO_GIGA/mpconfigboard.h
@@ -32,6 +32,11 @@ typedef unsigned int mp_uint_t;     // must be pointer size
 #define MICROPY_HW_ENTER_BOOTLOADER_VIA_RESET   (0)
 #define MICROPY_HW_TIM_IS_RESERVED(id) (id == 1)
 
+// ROMFS config
+#define MICROPY_HW_ROMFS_ENABLE_EXTERNAL_QSPI       (1)
+#define MICROPY_HW_ROMFS_QSPI_SPIFLASH_OBJ          (&spi_bdev.spiflash)
+#define MICROPY_HW_ROMFS_ENABLE_PART0               (1)
+
 // Flash storage config
 #define MICROPY_HW_SPIFLASH_ENABLE_CACHE            (1)
 #define MICROPY_HW_SPIFLASH_SOFT_RESET              (1)
@@ -125,8 +130,8 @@ void GIGA_board_low_power(int mode);
 // QSPI flash #1 for storage
 #define MICROPY_HW_QSPI_PRESCALER           (2) // 100MHz
 #define MICROPY_HW_QSPIFLASH_SIZE_BITS_LOG2 (27)
-// Reserve 1MiB at the end for compatibility with alternate firmware that places WiFi blob here.
-#define MICROPY_HW_SPIFLASH_SIZE_BITS       (120 * 1024 * 1024)
+// Reserve 4MiB for romfs and 1MiB for WiFi/BT firmware.
+#define MICROPY_HW_SPIFLASH_SIZE_BITS       (88 * 1024 * 1024)
 #define MICROPY_HW_QSPIFLASH_CS             (pyb_pin_QSPI2_CS)
 #define MICROPY_HW_QSPIFLASH_SCK            (pyb_pin_QSPI2_CLK)
 #define MICROPY_HW_QSPIFLASH_IO0            (pyb_pin_QSPI2_D0)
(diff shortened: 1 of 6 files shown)


iabdalkader:
@sebromero @ubidefeo

iabdalkader:
> This is most likely a breaking change for users, because the filesystem partition size has shrunk by 4MiB. And I don't think that will be easily detected because existing filesystems will mount and work correctly until the ROMFS is used. And then when ROMFS is used and the flash for that is written to, the original filesystem will probably still mount correctly because the initial filesystem data is unchanged.

We will deploy "stock" romfs images along with 1.25.0, which will likely corrupt the filesystem as you've expected. However, romfs will be the main storage for all embedded assets moving forward, as we're already running out of flash on some boards (like the Nicla), so this change is unavoidable. When we deploy 1.25.0 we could display a warning and erase the filesystem as well.

ports/stm32/boards/ARDUINO_GIGA/mpconfigboard.h
```diff
@@ -125,8 +130,8 @@ void GIGA_board_low_power(int mode);
 // QSPI flash #1 for storage
 #define MICROPY_HW_QSPI_PRESCALER           (2) // 100MHz
 #define MICROPY_HW_QSPIFLASH_SIZE_BITS_LOG2 (27)
-// Reserve 1MiB at the end for compatibility with alternate firmware that places WiFi blob here.
-#define MICROPY_HW_SPIFLASH_SIZE_BITS       (120 * 1024 * 1024)
+// Reserve 4MiB for romfs and 1MiB at for WiFi/BT firmware.
```
iabdalkader, line 133:
```suggestion
// Reserve 4MiB for romfs and 1MiB for WiFi/BT firmware.
```
Typo, in all files.

sebromero:
We could indeed add a version check in our installer tool to at least warns the user and consequently perform an erase of the filesystem.

Josverl:
For Mpflash it should be possible to detect the step-ups to 1.25.0 and then, after configuration:
- extract the filesystem, or files on the filesystem to the host
- erase , and flash
Wait for reboot, and the MCU to mount and format the flash
- if needed: Extract the files from the downloaded image
- re-upload the files.

Other tools can do this as well, if they can check the fw version up-front

dpgeorge:
> When we deploy 1.25.0 we could display a warning and erase the filesystem as well.

OK, sounds good, you can manage user expectations on your side for this change.

ports/stm32/boards/ARDUINO_GIGA/mpconfigboard.h
dpgeorge: Done.

dpgeorge:
@iabdalkader I have now updated this after #16903, to name the partition #0.  Please can you review it?

iabdalkader:
> @iabdalkader I have now updated this after #16903, to name the partition #0. Please can you review it?

I can retest it again, later today.

iabdalkader:
Seems to be working fine. I tested build, deploy and read. Anything else to test? Otherwise it's good.

```
>>> import os
>>> os.listdir("/rom")
['test.txt']
>>> print(open("/rom/test.txt", "r").read())
HelloWorld!
``

**Comment by dpgeorge** [comment]
> Seems to be working fine. I tested build, deploy and read

Great, thanks, that should be enough.

**Comment by iabdalkader** [comment]
I tested the address as well, using `uctypes.addressof`. The ROMFS should start at `0x00B00000` (11MiB) so this seems about right.
```
mpremote run ls_romfs.py 
addr: 0x00B00012  size: 12        alignment: NOT aligned name: test.txt
```

dpgeorge:
The address should be more like `0x90B00000`, right?

iabdalkader:
> The address should be more like `0x90B00000`, right?

Yes, that's right. The CYW43 firmware for example is at `0x90F00000`.

PR #16874, opened 2025-03-07
Introduce auto-listing of FlashArea Partitions

VynDragon:
### Summary

This enables listing all flash area partitions automagically instead of just sotrage_partitions. It uses the label, and the ID when not present.

### Testing

Tested on nrf52840

### DIFF
--- a/ports/zephyr/zephyr_storage.c
+++ b/ports/zephyr/zephyr_storage.c
@@ -139,6 +139,20 @@ MP_DEFINE_CONST_OBJ_TYPE(
 #endif // CONFIG_DISK_ACCESS
 
 #ifdef CONFIG_FLASH_MAP
+
+#define FLASH_AREA_DEFINE_LABEL(part) CONCAT(MP_QSTR_ID_, DT_STRING_TOKEN(part, label))
+#define FLASH_AREA_DEFINE_NB(part) CONCAT(MP_QSTR_ID_, DT_FIXED_PARTITION_ID(part))
+
+#define FLASH_AREA_DEFINE_GETNAME(part) COND_CODE_1(DT_NODE_HAS_PROP(part, label), \
+    (FLASH_AREA_DEFINE_LABEL(part)), (FLASH_AREA_DEFINE_NB(part)))
+
+#define FLASH_AREA_DEFINE_DEFINE(part) { MP_ROM_QSTR(FLASH_AREA_DEFINE_GETNAME(part)), MP_ROM_INT(DT_FIXED_PARTITION_ID(part)) },
+
+#define FLASH_AREA_DEFINE(part) COND_CODE_1(DT_NODE_HAS_STATUS_OKAY(DT_MTD_FROM_FIXED_PARTITION(part)), \
+    (FLASH_AREA_DEFINE_DEFINE(part)), ())
+
+#define FOREACH_PARTITION(n) DT_FOREACH_CHILD(n, FLASH_AREA_DEFINE)
+
 const mp_obj_type_t zephyr_flash_area_type;
 
 typedef struct _zephyr_flash_area_obj_t {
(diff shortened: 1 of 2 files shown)


ThinkTransit:
This is a great addition and really helps with identifying partitions.

**Tested on the nrf5340DK (8MB qpsi external flash)**

External flash added as 'my_flash'

```
&mx25r64 {
    partitions {
        compatible = "fixed-partitions";
        #address-cells = <1>;
        #size-cells = <1>;

        external_partition: partition@0 {
            reg = <0x00000000 0x800000>;
            label = "my_flash";
        };
    };
};
```

In micropython the partition is now visible by label.

```
>>> from zephyr import FlashArea
>>> FlashArea.__dict__
{'readblocks': <function>, 'writeblocks': <function>, 'ioctl': <function>, 'my_flash': 9, 'mcuboot': 0, 'image_0': 1, 'image_0_nonsecure': 2, 'image_1': 3, 'image_1_nonsecure': 4, 'tfm_ps': 5, 'tfm_its': 6, 'tfm_otp': 7, 'storage': 8}
>>> 
```

and can be mounted with
```
>>> import vfs
>>> from zephyr import FlashArea
>>> block_dev = FlashArea(FlashArea.my_flash, 4096)
>>> vfs.VfsLfs2.mkfs(block_dev)
>>> vfs.mount(block_dev, '/my_flash')
```

ports/zephyr/zephyr_storage.c
```diff
@@ -139,6 +139,22 @@ MP_DEFINE_CONST_OBJ_TYPE(
...
+
+#define FLASH_AREA_DEFINE_GETNAME(part) COND_CODE_1(DT_NODE_HAS_PROP(part, label), \
+    (FLASH_AREA_DEFINE_LABEL(part)), (FLASH_AREA_DEFINE_NB(part)))
+
+#define FLASH_AREA_DEFINE_DEFINE(part) { MP_ROM_QSTR(FLASH_AREA_DEFINE_GETNAME(part)), MP_ROM_INT(DT_FIXED_PARTITION_ID(part)) },
+
+#define FLASH_AREA_DEFINE(part) COND_CODE_1(DT_NODE_HAS_STATUS_OKAY(DT_MTD_FROM_FIXED_PARTITION(part)), \
+    (FLASH_AREA_DEFINE_DEFINE(part)), ())
+
+#define FOREACH_PARTITION(n) DT_FOREACH_CHILD(DT_DRV_INST(n), FLASH_AREA_DEFINE)
```
laodzu, line 156: We can do without DT_DRV_COMPAT pretty easily. Just change this to 

```
#define FOREACH_PARTITION(n) DT_FOREACH_CHILD(DT_INST(n, fixed_partitions), FLASH_AREA_DEFINE)
```

and remove DT_DRV_COMPAT. I think this makes things actually easier to understand.

laodzu (changes requested):
This looks pretty cool. Will test this later on a frdm_mcxn947

ports/zephyr/zephyr_storage.c
VynDragon: This looks better yea, will look into it

VynDragon: That does not work, but i figured another slightly different solution

ports/zephyr/zephyr_storage.c
```diff
@@ -139,6 +139,20 @@ MP_DEFINE_CONST_OBJ_TYPE(
 #endif // CONFIG_DISK_ACCESS
 
 #ifdef CONFIG_FLASH_MAP
+
+#define FLASH_AREA_DEFINE_LABEL(part) CONCAT(MP_QSTR_, DT_STRING_TOKEN(part, label))
```
projectgus, line 143: I know the existing `STORAGE` lookup doesn't have this, but it might be good to add either a prefix or a suffix so the programmer can see this is an ID rather than the partition object itself (that's right, isn't it? I haven't tested this myself.)

ports/zephyr/zephyr_storage.c
```diff
@@ -244,9 +258,7 @@ static const mp_rom_map_elem_t zephyr_flash_area_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_readblocks), MP_ROM_PTR(&zephyr_flash_area_readblocks_obj) },
     { MP_ROM_QSTR(MP_QSTR_writeblocks), MP_ROM_PTR(&zephyr_flash_area_writeblocks_obj) },
     { MP_ROM_QSTR(MP_QSTR_ioctl), MP_ROM_PTR(&zephyr_flash_area_ioctl_obj) },
-    #if FIXED_PARTITION_EXISTS(storage_partition)
-    { MP_ROM_QSTR(MP_QSTR_STORAGE), MP_ROM_INT(FIXED_PARTITION_ID(storage_partition)) },
-    #endif
+    DT_FOREACH_STATUS_OKAY(fixed_partitions, FOREACH_PARTITION)
```
projectgus, line 261: Please put a high-level comment about what all this is doing, for future readers.

(Don't have to explain exactly, just something like `// Generate named lookup of all partition IDs from DT flash area definitions`)

projectgus (commented):
This looks very useful!

What do you think about adding something brief about it in the "Flash Area" section of docs/zephyr/quickref.rst?

ports/zephyr/zephyr_storage.c
VynDragon: Done

ports/zephyr/zephyr_storage.c
VynDragon: done

dpgeorge:
> What do you think about adding something brief about it in the "Flash Area" section of docs/zephyr/quickref.rst?

Yes, I think this is a good idea.

VynDragon:
Is this ok?

dpgeorge (approved):
Yes, thanks, that addition to the quickref is good.

projectgus (approved):

PR #16876, opened 2025-03-07
mimxrt: A set of small bug fixes, additions and enhancements.

robert-hh:
### Summary

- Add the method adc.read_uv(). Tested with a MIMXRT1020_EVK and a MIMXRT1176_EVK board.
- Enable exFAT support. Tested with MIMXRT1176_EVK, MIMXRT1061 (Teensy 4.1) and MIMXRT1011 (Olimex RT1010).
- Enable PPP support for all boards with LWIP. Currently all boards with Ethernet support. PPP can be enabled for other boards without Ethernet like the Teensy 4.0 and sufficient RAM as well in a second step. Enabling for MIMXRT101x boards is hardly possible due to the large RAM  demand of LWIP. Tested with a Teensy 4.1 and a MIMXRT1020_EVK board and a SimCom A7608 GPRS/LTE modem.
- Remove duplicate UART init and make UART IRQ optional.
- Use a safe common flash CS timing. The flash devices used by the MIMXRT board are specified either with 3ns or 5 ns CS setup and hold time. Since a single configuration file is used for all boards, use 5ns instead of 3ns to be safe, even if there were no problems so far.
- Fix a bug in UART buffer allocation.The buffer would be reset on every call to uart.init(). If no sizes were given, the buffer would be set to the default size 256. That made problems e.g. with PPP. Also, the RX buffer was not stored to the UART object and not visible to GC as being in use. Then a gc.collect() would eventually free the buffer. The fix was tested with the PPP support test, allowing incoming messages larger than 256 bytes to be handled reliably.
- Add the `timeout=xxx` keyword argument to the I2C constructor. It will check for bus stuck low conditions. The default value is 100ms. Tested with a Teensy 4.1 board and by building the firmware for all boards.
- Support default device IDs for I2C, SPI and UART. Since all boards are configured to have a I2C(0), SPI(0) and UART(1), these can be taken as default devices, allowing the instantiation of I2C(), SPI(), UART().  Tested with a Teensy 4.1 board and by building the firmware for all boards.
- Update the deploy instructions when using the .uf2 bootloader. They were fine, but maybe a little bit too concise.

### Testing

See above.


ports/mimxrt/machine_i2c.c
```diff
@@ -82,24 +84,26 @@ bool lpi2c_set_iomux(int8_t hw_i2c, uint8_t drive) {
...
 
 mp_obj_t machine_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
-    enum { ARG_id, ARG_freq, ARG_drive};
+    enum { ARG_id, ARG_freq, ARG_drive, ARG_timeout};
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id, MP_ARG_REQUIRED | MP_ARG_OBJ },
+        { MP_QSTR_id, MP_ARG_INT, {.u_int = DEFAULT_I2C_ID} },
         { MP_QSTR_freq, MP_ARG_INT, {.u_int = DEFAULT_I2C_FREQ} },
         { MP_QSTR_drive, MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = DEFAULT_I2C_DRIVE} },
+        { MP_QSTR_timeout, MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = DEFAULT_I2C_TIMEOUT} },
```
dpgeorge, line 98: stm32, rp2 and the common `machine.SoftI2C` have this timeout parameter measured in microseconds, and default it to 50ms.  So please follow that convention here as well.

The name of this kwarg should stay as `timeout`, but I suggest calling the default macro `DEFAULT_I2C_TIMEOUT_US` to be clear about units (and set it to 50000).

Then adjust the logic converting `pinLowTimeout_ns` to/from this timeout value, to use 1000 instead of 1000000.

dpgeorge (approved):

PR #9458, opened 2022-09-29
py/ringbuf: Add micropython.ringbuffer() interface for general use.

andrewleech:
I've had a number of desirable use cases for a python ringbuffer over the years, usually relating to logging and/or connecting streams together in various ways. 

This PR adds an optional python interface to the existing C ringbuffer library.

---
This work was funded by Planet Innovation.

PR #6172, opened 2020-06-19
stm32/usb_cdc: Double data speed sending from micropython to computer

andrewleech:
## stm32/usb_cdc: Fix tx ringbuffer to remove full == size-1 limitation.

The current usb virtual serial port tx ringbuffer is the basic implementation where it can only be filled to a maximum of `buffer size-1`. 
For a 1024 size buffer this means the largest packet that can be sent is 1023. 

Once this is sent though, the next byte copied in goes to the final byte in the buffer, so must be sent as a 1 byte packet before the read pointer can be wrapped around. So in large streaming transfers, watching the usb sniffer you basically get alternating 1023 byte packets then 1 byte packets. 

By switching the ringbuffer to the different scheme that doesn't have the full limitation, we can get constant 1023 packets.

For a simple explanation of the two schemes, look at the first and last implementations discussed here: https://www.snellman.net/blog/archive/2016-12-13-ring-buffers/

Note: I have not merged this with the circular buffer from `ringbuf.h` due to the need to read from the buffer in chunks rather than single bytes, as well as the de-coupled copy of the read pointer.

The RX buffer could likely do with a similar change, though as it's not read from in chunks like the TX buffer it doesn't present the same issue, all that's lost is one byte capacity of the buffer.

PR #4920, opened 2019-07-11
py/ringbuf.h: Avoid wasted byte in ringbuf and don't inline get/put.

jimmo:
For discussion... I'd like to consolidate `py/ringbuf.h` with the ringbuf implementation in modbluetooth.c in #4893 written by @aykevl, so this PR merges the two implementations into py/ringbuf.h.
The benefit of @aykevl's implementation is that when the size of the ringbuf is divisible into 2^16 (i.e. `2**16 % size == 0`), we can make the code smaller and avoid a wasted byte in the buffer.

I'm unsure whether the requirement that the buffer is divisible into 2^16 is a problem. It means that if the user sets the `rxbuf` kwarg on ESP8266, they may get a slightly larger buffer than expected (the code rounds up to the nearest power of two).

I've also changed ringbuf to no longer inline get/put -- this is only a saving if each of get/put are called once (on ESP8266/ESP32 this is not the case). On NRF, LTO does this for us anyway. (And NRF will benefit further when this gets used in modbluetooth).

This reduces code size on ESP8266 (-124) and NRF (-84) ports, and marginal increase on ESP32 (+24).

PR #4617, opened 2019-03-20
py/schedule: Convert micropythyon.schedule() to circular buffer.

andrewleech:
This means the schedule operates on a first-scheduled, first-executed manner rather than the current last-scheduled, first executed.

Benefit:
When application/interrupt schedules multiple functions they're guaranteed to be run in the order of scheduling.

Currently in a busy system that isn't sleeping very often, if multiple functions or interrupts schedule tasks before they can be serviced they will be run out of order, potentially causing confusion or unexpected side-effects.

Costs:
* This costs 32 bytes extra flash on my build (`-Os` or `-Og`), no extra ram usage.
* `runtime.h`/`mp_sched_num_pending()` is now a uint_8 subtraction rather than directly reporting a uint16_t.
* The implementation requires `MICROPY_SCHEDULER_DEPTH` to be order of 2 sized. This is enforced with a `MP_STATIC_ASSERT`.

PR #16884, opened 2025-03-09
ports SAMD and RP2: Fix allocation of UART buffers.

robert-hh:
### Summary

The UART buffers were reset on every call to uart.init(). If no sizes were given, the buffers ere set to the default size 256. That made problems e.g. with PPP.
This PR fixes it, keeping the buffer size if not deliberately changed and allocating new buffers only if the size was changed.
Fix a subtle problem of the SAMD port where the board would lock up in same UART loopback mode, if rxbuf is full and not emptied. Then sending of the data does not complete. 

### Testing

- Set the UART buffer sizes to non-default values and then call uart.init() without changing the buffer size. Verified that the sizes did not change and data still can be sent & received.
- Change the buffer sizes in calls to uart.init().  Verify that the size has changed and data still can be sent & received.

### DIFF
--- a/ports/rp2/machine_uart.c
+++ b/ports/rp2/machine_uart.c
@@ -116,8 +116,10 @@ typedef struct _machine_uart_obj_t {
     uint16_t timeout_char;  // timeout waiting between chars (in ms)
     uint8_t invert;
     uint8_t flow;
+    uint16_t rxbuf_len;
     ringbuf_t read_buffer;
     mutex_t *read_mutex;
+    uint16_t txbuf_len;
     ringbuf_t write_buffer;
     mutex_t *write_mutex;
     uint16_t mp_irq_trigger;   // user IRQ trigger mask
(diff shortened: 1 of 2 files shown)


Gadgetoid:
Slight tangent but the extra byte ringbuf uses for "tracking" (I had to dig to find out what on earth that was for) really makes a royal mess of this code.

(Even bigger tangent) It also suggests some invocations of `ringbuf_alloc` are incorrect, since they don't account for this extra byte. Eg: https://github.com/micropython/micropython/blob/3823aeb0f14c04084eba6566164d9b7dbd9e7ced/ports/esp8266/modespnow.c#L183

robert-hh:
> Slight tangent but the extra byte ringbuf uses for "tracking" (I had to dig to find out what on earth that was for) really makes a royal mess of this code.

It this a comment about the +/-1 byte is tedious (which I agree), or are you saying that the code is wrong?

Gadgetoid:
That said there's right and then there's correct.

I don't think: `self->read_buffer.size = DEFAULT_BUFFER_SIZE + 1;` is "correct." It feels like a trap ready to spring. Doesn't look like `read_buffer.size` is used anywhere without ` - 1` so these cancel out?

Also any code that tries to consume a `machine_uart_obj_t` without knowledge of the `ringbuf_alloc` wart would get an incorrect `read_buffer.size`?

(I appreciate you've inherited this wart 😬)

robert-hh:
> I don't think: self->read_buffer.size = DEFAULT_BUFFER_SIZE + 1; is "correct."

Technically it looks correct to me. The symbols `DEFAULT_BUFFER_SIZE` and `rx_len` are related to the external view of a user/user code. `read_buffer.size` is the internal value of the ringbuf API, which has always to be by 1 larger than the external view.
And yes, there are ringbuf implementation which avoids this fussiness( see e.g. https://www.snellman.net/blog/archive/2016-12-13-ring-buffers/), but with limitations.

Gadgetoid:
> `read_buffer.size` is the internal value of the ringbuf API

Aha that's the piece of the puzzle I was missing. Thank you.

andrewleech:
FWIW the "other" kind of ringbuffer that avoids the tracking byte is used & discussed in a couple of places and is quite elegant, but comes with a restriction that the size must be a x^2, ie 16, 32, 64, 128 etc. You can no longer have an arbitrary size.

* https://github.com/micropython/micropython/pull/6172
* https://github.com/micropython/micropython/pull/4920
* https://github.com/micropython/micropython/pull/4617

On a related note, I did add the +1 behind the scenes in https://github.com/micropython/micropython/pull/9458 when the user requests a 16byte buffer it'll allocate a 17byte array for use to avoid the confusion. If the user passes in a pre-allocated buffer though this isn't possible, so they get -1 size. I tried to address this in the docs on the function.

It turns out though the +1 is kinda inefficient in many cases, due to the gc allocation blocks in micropython. If the user asks for 16byte and get given a 17byte buffer behind the scenes, well micropython will have allocated two blocks in gc which are typically 16byte blocks - so 32bytes will be consumed for that 16/17byte ringbuffer. Implementation details...

dpgeorge (approved):
This fix looks good, thanks!

(I split out the samd loopback bug fix to a separate commit.)

robert-hh:
Thank you very much.

Issue #5384, opened 2019-12-05
Two argument form of iter() is not implemented.


hardkrash:
The builtin `iter(callable, sentinel)` from pep [234](https://www.python.org/dev/peps/pep-0234/#python-api-specification) is not implemented.

The use case is something like this.

```python
with open('file.bin', 'rb') as f:
    for chunk in iter(lambda: f.read(128), b''):
        process_chunk(chunk)
```
The two parameter form allows an easy way to make a callable into an iterable by using a sentinel.

It allows for replacing the do while workarounds with a simple iterator.

```python
with open('file.bin', 'rb') as f:
    chunk = f.read(128)
    while chunk != b'':
        process_chunk(chunk)
        chunk = f.read(128)
```

Credit to this stack overflow answer from [jfs](https://stackoverflow.com/users/4279/jfs)
https://stackoverflow.com/a/20014805

mikewadsten:
Sort of related to #4400.

dpgeorge:
Yes, this is simply not implemented, mainly because the same thing can be achieved in pure Python.  As a workaround the following can be used:
```python
def iter2(c, s):
    while True:
        v = c()
        if v == s:
            raise StopIteration
        yield v
```
(This would be simpler with the walrus operator!)

hardkrash:
While I agree on the fact that this can be worked around in many ways.  it is unfortunate that this is not explicitly called out in the documentation as a difference from python's core syntax.

This would be a compatibility wrapper needed to make the two argument form of iter work without changing source code.

`from compat import iter`

Where compat.py has something like this.

```python
S = some_singleton
_orig_iter = iter
def _compat_iter(c, s=S):
    if s is S:
        for x in _orig_iter(c):
            yield x
    else:
        while True:
            v = c()
            if v == s:
                raise StopIteration
            yield v

iter = _compat_iter
```

Would adding 2 argument form to the builtin iter function be an acceptable patch?

dpgeorge:
>  it is unfortunate that this is not explicitly called out in the documentation as a difference from python's core syntax.

It's not really a syntax difference, rather the fact that MicroPython implements a subset of the Python builtins.  It could be added to the docs via `tests/cpydiff` (which automatically generates differences for the docs).

> Would adding 2 argument form to the builtin iter function be an acceptable patch?

[...]

PR #16902, opened 2025-03-11
py/mpconfig: Enable 2-argument built-in next() at basic feature level.

dpgeorge:
### Summary

Enable the 2-argument form of the built-in `next()` function at the basic feature level (which is one above core features).

This is a pretty fundamental built-in and having CPython-compatible behaviour is beneficial.  The code size increase is not much, and ports/boards can still disable it if needed to save space.

Addresses issue #5384.

### Testing

CI should test this.

Could leave it disabled and document better the differences.  But I think it's worth the increase in code size, most big boards have enough space.


projectgus:
Change LGTM but we should probably update or remove `tests/cpydiff/builtin_next_arg2.py` at the same time. (I'm not sure what the approach is for CPython differences that only exist at certain feature levels?)

PR #16903, opened 2025-03-11
ports: make naming of ROMFS partition config variables consistent

dpgeorge:
### Summary

Change ROMFS partition configuration variables to be more consistent across the ports, and use index 0 as the starting partition number (instead of index 1).

Reason:
- `vfs.rom_ioctl()` numbers the partitions starting from 0
- `mpremote romfs -p <partition id>` numbers the partitions starting from 0

### Testing

CI should test this, building the boards that enable ROMFS on stm32 and esp8266.

### DIFF
--- a/ports/esp8266/boards/esp8266_2MiB_ROMFS.ld
+++ b/ports/esp8266/boards/esp8266_2MiB_ROMFS.ld
@@ -17,8 +17,8 @@ MEMORY
 }
 
 /* define ROMFS extents */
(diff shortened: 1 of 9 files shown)


dpgeorge:
@iabdalkader FYI

iabdalkader:
This is much better, thanks! Note, this PR will have to be updated or merged first https://github.com/micropython/micropython/pull/16869

PR #16904, opened 2025-03-11
extmod/vfs_rom: Implement minimal VfsRom.getcwd() method.

dpgeorge:
### Summary

This is needed if you chdir to a ROMFS and want to query your current directory.

Prior to this change, using `os.getcwd()` when in a ROMFS would raise:

    AttributeError: 'VfsRom' object has no attribute 'getcwd'

### Testing

A test has been added, which will run under CI.

I wanted to keep `VfsRom` as minimal as possible so omitted this `getcwd` method initially.  But without this it's pretty awkward to know that you are in a ROMFS directory.  Eg on a board that only has a ROMFS, that will always be the current directory, and you'd like `os.getcwd()` to actually tell you that.


projectgus (approved):
LGTM. The implementation confused me for a moment until I remembered `chdir()` doesn't allow changing directory into any ROMFS subdirectory.

PR #16907, opened 2025-03-11
rp2,esp32,extmod: Implement submodule update inside CMake.

### LINKED ISSUE #16870: make submodules fails in ports/rp2: Internal build error: The submodule list should not be empty.

projectgus:
### Summary

Alternative approach to #16901. Fixes #16870.

Rather than having `make submodules` run CMake to get the submodules list and then update git submodules, have CMake update submodules itself. This means any other build errors will be visible, rather than being masked (as currently the CMake submodules run has its output captured and parsed.)

This effectively reverts commit 22353e9 from https://github.com/micropython/micropython/pull/16581, as an empty submodule list is no longer a symptom of any deeper (masked) issue.

It's also possible to simplify out a lot of the checks for `ECHO_SUBMODULES` as CMake won't check if those files and directories exist until after the `UPDATE_SUBMODULES` pass finishes. A workaround is still needed for ESP32 as the ESP-IDF framework tests if paths exist when registering a new component, but the workaround can be contained in the esp32_common.cmake file.

This PR also contains a commit to fail with a helpful error message if the pico-sdk is missing. This was originally necessary because I was trying to run the `UPDATE_SUBMODULES` pass in the normal build directory (to save on running CMake twice), but that actually doesn't quite work as pico-sdk sets some other invalid cache values if directories are missing. However, the helpful error message is still helpful!

### Testing

* Ran some local rp2 and esp32 builds in a clean git clone, verified all successful.
* Ran a build with the older IDF component manager in the IDF V5.2.2 docker image, verified still OK (but does still require the workaround from ec527a11). Command line:  `podman run --rm -t -v "/home/gus/ry/george/micropython/.git:/gitdir:ro" -t docker.io/espressif/idf:v5.2.2 bash -c "git clone -b $(git branch --show-current) /gitdir micropython && make -j -C micropython/mpy-cross && echo '@@@@@@@@@@ submodules' && make -C microp
 ython/ports/esp32 submodules V=1 BOARD=ESP32_GENERIC_S3 && echo '@@@@@@@@@@ main build' && make -C micropython/ports/esp32 BOARD=ESP32_GENERIC_S3" `

* The `make submodules` output now includes a lot of CMake build output for a dummy run which never gets to actually build.
* We could make this even simpler and hard-code the `GIT_SUBMODULES` value in `rp2/Makefile` and `esp32/Makefile`. The upside is a much simpler build stage. There are two downsides I can see:
  - Conceptually, this means each port has to know about implementation details (the submodule path) of all its external MP modules.
  - Practically, can no longer filter the list of submodules based on the board to reduce network traffic and disk usage. On ESP32 the submodule list is current the same for all boards so that doesn't have any current impact. On rp2 the list is much shorter for boards without Wi-Fi or networking, so some additional repos would be fetched unnecessarily. I'm not sure how much more traffic this would actually be, though.

### DIFF
--- a/extmod/extmod.cmake
+++ b/extmod/extmod.cmake
@@ -107,62 +107,57 @@ set(MICROPY_SOURCE_LIB_LIBM_SQRT_HW ${MICROPY_DIR}/lib/libm/thumb_vfp_sqrtf.c)
 
 if(MICROPY_PY_BTREE)
     set(MICROPY_LIB_BERKELEY_DIR "${MICROPY_DIR}/lib/berkeley-db-1.xx")
-    string(CONCAT GIT_SUBMODULES "${GIT_SUBMODULES} " lib/berkeley-db-1.xx)
+    list(APPEND GIT_SUBMODULES lib/berkeley-db-1.xx)
 
-    if(ECHO_SUBMODULES)
-        # No-op, we're just doing submodule/variant discovery.
-        # Cannot run the add_library/target_include_directories rules (even though
-        # the build won't run) because IDF will attempt verify the files exist.
-    elseif(NOT EXISTS ${MICROPY_LIB_BERKELEY_DIR}/README)
+    if(NOT UPDATE_SUBMODULES AND NOT EXISTS ${MICROPY_LIB_BERKELEY_DIR}/README)
         # Regular build, submodule not initialised -- fail with a clear error.
         message(FATAL_ERROR " MICROPY_PY_BTREE is enabled but the berkeley-db submodule is not initialised.\n Run 'make BOARD=${MICROPY_BOARD} submodules'")
-    else()
-        # Regular build, we have the submodule.
-        add_library(micropy_extmod_btree OBJECT
-            ${MICROPY_LIB_BERKELEY_DIR}/btree/bt_close.c
-            ${MICROPY_LIB_BERKELEY_DIR}/btree/bt_conv.c
-            ${MICROPY_LIB_BERKELEY_DIR}/btree/bt_debug.c
-            ${MICROPY_LIB_BERKELEY_DIR}/btree/bt_delete.c
-            ${MICROPY_LIB_BERKELEY_DIR}/btree/bt_get.c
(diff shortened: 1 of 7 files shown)


dpgeorge (approved):
This is nice improvement, definitely a lot cleaner this way (than printing the submodules and using make to updated them).

Tested with various rp2 boards, including ARDUINO_NANO_RP2040_CONNECT which needs mynewt-nimble.

Issue #16908, opened 2025-03-11
rp2.bootsel_button() not working on Raspberry Pi Pico 2 W (RP2350)


sozoraemon:
### Port, board and/or hardware

RP2 port, Raspberry Pi Pico 2 W (RP2350)

### MicroPython version

3.4.0; MicroPython v1.25.0-preview.49.g0625f07ad.dirty on 2024-11-21
(name='micropython', version=(1, 25, 0, 'preview'), _machine='Raspberry Pi Pico 2 W with RP2350', _mpy=7942)

- On Raspberry Pi Pico (RP2040), `rp2.bootsel_button()` works as expected.
- On Raspberry Pi Pico 2 W (RP2350), `rp2.bootsel_button()` **always returns `1`**, regardless of whether the BOOTSEL button is pressed or not.
- Tested on **MicroPython v1.25.0-preview.49.g0625f07ad.dirty (2024-11-21)**.
- This issue **does not occur on RP2040**, suggesting that the BOOTSEL button may be wired differently on RP2350.
- GPIO scan (`GPIO0-29`) did not detect any changes when pressing the button.
- `SIO` registers (`GPIO_HI_IN` and `GPIO_IN`) also did not show a clear state change.
- Further investigation is needed to determine how to read the BOOTSEL button on RP2350.

### Steps to Reproduce
1. Copy and paste the following code into a new file:
   ```python
   import sys, time, rp2
   print(f"sys.implementation._machine={sys.implementation._machine}")
   for i in range(5):
       print(f"rp2.bootsel_button() = {rp2.bootsel_button()}")
       time.sleep(1)

2.Run the script on Raspberry Pi Pico 2 W (RP2350).
3.Press and release the BOOTSEL button during execution.

Expected behavior
・On Pico (RP2040): rp2.bootsel_button() returns 1 when pressed, 0 otherwise.
・On Pico 2 W (RP2350): Expected the same behavior.

Actual behavior
・Pico (RP2040):  Works correctly.
・Pico 2 W (RP2350): Always returns 1, even when the button is not pressed.


### Expected behaviour

### Expected Behaviour
- On **Raspberry Pi Pico (RP2040)**:
  - `rp2.bootsel_button()` should return `1` when the BOOTSEL button is pressed and `0` otherwise.
  - Example output when pressing and releasing the button:
    ```
    rp2.bootsel_button() = 0
    rp2.bootsel_button() = 1  # Button pressed
    rp2.bootsel_button() = 1
[...]

PR #16909, opened 2025-03-11
ports/rp2: Fix bootsel for RP2350.

Gadgetoid:
Fixes reading bootsel as a button functionality, as raised in #16908

Tested on a Pico 2 W using [v1.25.0-preview.365.g3823aeb0f (2025-03-07) .uf2](https://micropython.org/resources/firmware/RPI_PICO2_W-20250307-v1.25.0-preview.365.g3823aeb0f.uf2) to confirm the bug.

Tested on a Pico W to check for regressions.

Note: I did not actually author this change, it was cribbed form the original source of the bootsel code which had been updated with RP2350 support: https://github.com/raspberrypi/pico-examples/blob/84e8d489ca321a4be90ee49e36dc29e5c645da08/picoboard/button/button.c#L42-L46

### DIFF
--- a/ports/rp2/modrp2.c
+++ b/ports/rp2/modrp2.c
@@ -42,11 +42,17 @@
 MP_DECLARE_CONST_FUN_OBJ_VAR_BETWEEN(mod_network_country_obj);
 #endif
 
+#define CS_PIN_INDEX 1
+
+#if PICO_RP2040
+    #define CS_BIT (1u << CS_PIN_INDEX)
+#else
+    #define CS_BIT SIO_GPIO_HI_IN_QSPI_CSN_BITS
+#endif
+
 // Improved version of
 // https://github.com/raspberrypi/pico-examples/blob/master/picoboard/button/button.c
 static bool __no_inline_not_in_flash_func(bootsel_button)(void) {
-    const uint CS_PIN_INDEX = 1;
-
     // Disable interrupts and the other core since they might be
     // executing code from flash and we are about to temporarily
     // disable flash access.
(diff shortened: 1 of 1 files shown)


ports/rp2/modrp2.c
```diff
@@ -65,7 +65,12 @@ static bool __no_inline_not_in_flash_func(bootsel_button)(void) {
 
     // The HI GPIO registers in SIO can observe and control the 6 QSPI pins.
     // The button pulls the QSPI_SS pin *low* when pressed.
-    bool button_state = !(sio_hw->gpio_hi_in & (1 << CS_PIN_INDEX));
+    #if PICO_RP2040
+        #define CS_BIT (1u << 1)
```
dpgeorge, line 69: Can this be `1u << CS_PIN_INDEX`?

ports/rp2/modrp2.c
```diff
@@ -65,7 +65,12 @@ static bool __no_inline_not_in_flash_func(bootsel_button)(void) {
...
+    #else
+        #define CS_BIT SIO_GPIO_HI_IN_QSPI_CSN_BITS
+    #endif
```
dpgeorge, line 72: I suggest moving this #if block to just above this bootsel_button function (that's the general coding style, not to have #defines within code itself).

ports/rp2/modrp2.c
Gadgetoid: Any idea why `CS_PIN_INDEX` even exists at all? It's inherited from the original example code, so readability I would guess.

I'll move it to a define so that moving the `#if` block is possible.

ports/rp2/modrp2.c
Gadgetoid: Good shout, agreed!

Gadgetoid:
Thank you for the review, suggested changes done!

dpgeorge (approved):
Tested on RPI_PICO and RPI_PICO2.  It works.

dpgeorge:
Rebased and merged in e75ffc3bccb8b213704f0ddd2ce82731ab2069ba

PR #16914, opened 2025-03-12
Update cyw43-driver to v1.1.0, and add WPA3 constants to network.WLAN for cyw43

### LINKED ISSUE #16482: Pico W port Wifi Problems

RP2040 on Raspberry Pi Pico W

### MicroPython version

(sysname='rp2', nodename='rp2', release='1.24.1', version='v1.24.1 on 2024-11-29 (GNU 13.2.0 MinSizeRel)', machine='Raspberry Pi Pico W with RP2040')

### Reproduction

1. Clone https://github.com/edmcman/rpi-temp-server/tree/debug
2. Edit SSID/password in main.py
3. Run `pipkin --port /dev/ttyACM0 install -r requirements.txt`
4. Edit the IP address in test.bash to batch the Pico W's IP
5. Run test.bash a bunch of times.

### Expected behaviour

The test.bash script tests whether the web server remains accessible after progressively longer periods of times with no requests.  It is expected that the web server will remain available and the script will print "Test succeeded".

### Observed behaviour

When run 20 times with a limit of 2048 second periods, 8 out of the 20 runs failed, meaning that the web server became inaccessible.  Here is a breakdown of the power modes during failures:

```
fgrep -l -R Fail *.log | xargs fgrep Random | awk '{print $4}' | sort | uniq -c
      6 network.WLAN.PM_POWERSAVE
      2 None
```

And the overall distribution of tested pm values:
```
      2 0xa11140
      8 network.WLAN.PM_NONE
      1 network.WLAN.PM_PERFORMANCE
      7 network.WLAN.PM_POWERSAVE
      3 None
```

I am not sure why they are so uneven.

### Additional Information

# High level problems

I think there are two problems:
1. The default PM value may not be ideal.  My testing suggests that `network.WLAN.PM_NONE` may be more reliable.
2. The fact that the Pico W reports it is still connected and can't reconnect is very problematic.

# Other Information

When the Pico W becomes inaccessible, the logs contain the following messages repeated very often:
```
[ 2509685] ASYNC(0000,49,0,0,0)
```

It is unclear what the 49 message is.  Under normal operation, it appears to be correlated with PSK_SUP messages.  But as soon as it starts repeating by itself the Pico W is offline.

Also, an asynchronous task that calls `WLAN.isconnected()` and `WLAN.status()` shows that the Pico W is connected, even though it does not seem to be:
```
[snip]
[ 3505634] ASYNC(0000,49,0,0,0)
[ 3507681] ASYNC(0000,49,0,0,0)
[ 3508397] ASYNC(0000,49,0,0,0)
Still checking Wi-Fi connection...
Debug: connected: True status: 3
Actual PM: 0xa11142
```

Here is a zip file of my collected logs: 
[logs-2048.zip](https://github.com/user-attachments/files/18247013/logs-2048.zip)

I am currently conducting more and longer tests to hopefully provide more information.  In theory, if `PM_NONE` proves more reliable, changing the default to help a lot in practice.  But even the current behavior for other power management modes is very concerning as it means long term connections are basically impossible.

### Code of Conduct

Yes, I agree

### LINKED ISSUE #16624: Please update the version of cyw43-driver
### Description

Please update the version of cyw43-driver to include [this commit](https://github.com/georgerobotics/cyw43-driver/commit/c1075d4bc440422cf2b2fd12c64a1f53f77660ee) which fixes an issue I encountered in micropython (https://github.com/micropython/micropython/issues/16482).

### Code Size

N/A.  The code size difference should be minimal.

### Implementation

I intend to implement this feature and would submit a Pull Request if desirable, I hope the MicroPython maintainers or community will implement this feature

### Code of Conduct

Yes, I agree

dpgeorge:
### Summary

This updates the cyw43-driver to the latest v1.1.0 version.

This is a minor release of cyw43-driver, some bugs have been fixed and some features have been added.  Those things were all tested upstream for that driver.

A few configuration constants have been updated here to work with the new driver version (very minor changes).

And, now that cyw43-driver supports WPA3, the WPA3 security constants have been added to `network.WLAN` for that driver's bindings.  These user-facing Python constants match the esp32 constant names.

Will address #16624 and fix #16482.

### Testing

Tested this PR on PYBD_SF6 and RPI_PICO_W, running all available WLAN and BLE tests in this repo.  There were no regressions.


extmod/network_cyw43.c
```diff
@@ -583,6 +583,8 @@ static const mp_rom_map_elem_t network_cyw43_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_IF_AP), MP_ROM_INT(MOD_NETWORK_AP_IF) },
     { MP_ROM_QSTR(MP_QSTR_SEC_OPEN), MP_ROM_INT(CYW43_AUTH_OPEN) },
     { MP_ROM_QSTR(MP_QSTR_SEC_WPA_WPA2), MP_ROM_INT(CYW43_AUTH_WPA2_MIXED_PSK) },
+    { MP_ROM_QSTR(MP_QSTR_SEC_WPA3), MP_ROM_INT(CYW43_AUTH_WPA3_SAE_AES_PSK) },
+    { MP_ROM_QSTR(MP_QSTR_SEC_WPA2_WPA3), MP_ROM_INT(CYW43_AUTH_WPA3_WPA2_AES_PSK) },
```
projectgus, line 587: It's maybe out of scope for this PR, but `network.WLAN.rst` should probably have a list of these constants. I can do this in a follow-up if you prefer?

projectgus (approved):
I haven't reviewed the driver, but the few changes on the MP side LGTM!

extmod/network_cyw43.c
dpgeorge: Yes, would be good to add them to the docs.

If you could do it in a follow-up, that would be great, thanks!

(They are quite consistent now among ports, which is a relief.)

dpgeorge:
Note: increase in firmware size for RPI_PICO_W is partly due to larger BT patch firmware (about +800 bytes).  Might need to investigate the remaining increase, and check PYBD_SF6 as well.

dpgeorge:
The increase in firmware on PYBD_SF6 is +500 bytes.  This is due mainly to the driver now supporting WPA3 on both STA and AP interfaces.

For RPI_PICO_W, it's about +500 for the same WPA3 support, and an additional +800 for improved BT patch firmware.

PR #16915, opened 2025-03-12
rp2: Fix lost CYW43 WiFi events when using both cores.

### LINKED ISSUE #16779: RP2: Pico 2 W Not connecting with thread running

projectgus:
### Summary

Closes #16779. Thanks to @coencoensmeets for a clear and easy to reproduce bug report, and to @Gadgetoid for doing some deep digging to zero in on the root cause.

There's a very odd but predictable sequence of events that breaks Wi-Fi when using both cores:

1) CPU1 calls `pendsv_suspend()` - for example sleep() causes a softtimer node to be inserted, which calls `pendsv_suspend()`.
2) CYW43 sends wakeup IRQ. CPU0 GPIO IRQ handler schedules PendSV and disables the GPIO IRQ on CPU0, to re-enable after `cyw43_poll()` runs and completes.
3) CPU0 `PendSV_Handler` runs, sees pendsv is suspended, exits.
4) CPU1 calls `pendsv_resume()` and `pendsv_resume()` sees PendSV is pending and triggers it on CPU1.
5) CPU1 runs `PendSV_Handler`, runs `cyw43_poll()`, and at the end it re-enables the pin IRQ *but now on CPU1*.

However CPU1 has the overall GPIO IRQ line disabled, so the CYW43 interrupt never runs again...

Fix is to specifically only enable the GPIO interrupt on CPU0. This isn't supported in pico-sdk but appears to be supported by the hardware.

Plus two follow-up commits:

* Refactor out the `cyw43_has_pending` global flag to instead be a check of the pendsv table. This was necessary for the previous version of this PR. Not necessary now, but seems like a worthwhile change.
* Enable PendSV interrupt at the correct priority on both CPUs, and even if CYW43 is not in use. Also adds some comments in PendSV explaining how it can run on either CPU.

### Testing

* Ran the test Python program from the linked issue on both RPI_PICO_W and RPI_PICO2_W boards, verified the problem goes away.
* Ran the multi_wlan test between RPI_PICO_W and RPI_PICO2_W and an ESP32, verified tests still pass.

* First version of this PR instead changed the interrupt to rising edge, so it didn't need to be disabled. This appears to work fine, and simplifies the code, but actually the CYW43 interrupt pin on RP2 is muxed with the data pin so the GPIO edge interrupt was triggering spuriously at a very high frequency. Oops!
* @Gadgetoid has a PR #16788 which adds support for CPU affinity in PendSV callbacks. We could add this support, although we don't really need the feature (everything that calls `pendsv_schedule_dispatch()` runs on CPU0).
* As @Gadgetoid also suggested in #16788, we could change PendSV to only ever fire on CPU0. In fact, comments in pendsv.c suggest that's the current expectation. I actually implemented this fix first, see https://github.com/projectgus/micropython/commit/47723da831afb4d2bdd7484c8ab8537f986a6339 . That is still more complex than this fix as it needs a cross-core PendSV trigger for when CPU1 does `pendsv_resume()`. However it might be easier to reason about overall.

### DIFF
--- a/ports/rp2/pendsv.h
+++ b/ports/rp2/pendsv.h
@@ -42,9 +42,6 @@ enum {
 
 #define PENDSV_DISPATCH_NUM_SLOTS PENDSV_DISPATCH_MAX
 
-// PendSV IRQ priority, to run system-level tasks that preempt the main thread.
-#define IRQ_PRI_PENDSV PICO_LOWEST_IRQ_PRIORITY
-
 typedef void (*pendsv_dispatch_t)(void);
 
 void pendsv_init(void);
--- a/ports/rp2/pendsv.h
+++ b/ports/rp2/pendsv.h
@@ -51,5 +51,6 @@ void pendsv_init(void);
 void pendsv_suspend(void);
 void pendsv_resume(void);
 void pendsv_schedule_dispatch(size_t slot, pendsv_dispatch_t f);
+bool pendsv_is_pending(size_t slot);
 
 #endif // MICROPY_INCLUDED_RP2_PENDSV_H
--- a/ports/rp2/cyw43_configport.h
+++ b/ports/rp2/cyw43_configport.h
@@ -140,10 +140,12 @@ uint cyw43_get_pin_wl(cyw43_pin_index_t pin_id);
 #endif
 
 void cyw43_post_poll_hook(void);
-extern volatile int cyw43_has_pending;
+static inline bool cyw43_poll_is_pending(void) {
+    return pendsv_is_pending(PENDSV_DISPATCH_CYW43);
+}
 
 static inline void cyw43_yield(void) {
-    if (!cyw43_has_pending) {
+    if (!cyw43_poll_is_pending()) {
         best_effort_wfe_or_timeout(make_timeout_time_ms(1));
     }
 }
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -145,7 +145,7 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
 
     uint32_t my_interrupts = MICROPY_BEGIN_ATOMIC_SECTION();
     #if MICROPY_PY_NETWORK_CYW43
-    if (cyw43_has_pending && cyw43_poll != NULL) {
+    if (cyw43_poll_is_pending()) {
         MICROPY_END_ATOMIC_SECTION(my_interrupts);
         return;
     }
--- a/ports/rp2/mpnetworkport.c
+++ b/ports/rp2/mpnetworkport.c
@@ -59,13 +59,36 @@ static soft_timer_entry_t mp_network_soft_timer;
 
 volatile int cyw43_has_pending = 0;
 
+// The Pico SDK only lets us set GPIO wake on the current running CPU, but the
+// hardware doesn't have this limit. We need to always enable/disable the pin
+// interrupt on CPU0, regardless of which CPU runs PendSV and
+// cyw43_post_poll_hook(). See feature request at https://github.com/raspberrypi/pico-sdk/issues/2354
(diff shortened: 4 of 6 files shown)


projectgus:
> * hange PendSV to only ever fire on CPU0. In fact, comments in pendsv.c suggest that's the current expectation

I've pushed another commit here to update these comments, and also to ensure PendSV runs at the same priority on CPU0 and CPU1.

projectgus:
@dpgeorge let me know that the HOST_WAKE line on the rp2 is muxed with the (bidirectional) SPI data line. So with this PR the gpio irq triggers many times unnecessarily while `cyw43_poll` is running...

image
(Image is from toggling a GPIO in the `gpio_irq_handler()` network function.)

Will need to rethink...

projectgus:
New version pushed which keeps the level-triggered interrupt but hand-rolls a function to only ever enable it on CPU0.

Have also a raised a feature request in pico-sdk to have this in the GPIO API: https://github.com/raspberrypi/pico-sdk/issues/2354

ports/rp2/pendsv.h
```diff
@@ -27,6 +27,7 @@
 #define MICROPY_INCLUDED_RP2_PENDSV_H
 
 #include <stddef.h>
+#include "hardware/irq.h"
```
dpgeorge, line 30: I don't think this include is needed (or maybe needs to be moved to `pendsv.c`).

dpgeorge:
The latest changes look really good.  I will test them.

ports/rp2/pendsv.h
dpgeorge: > This header has `#define IRQ_PRI_PENDSV PICO_LOWEST_IRQ_PRIORITY` in it, so I added this on the basis of "include what you use".

Ah, I see.  That makes sense.

> But I now realise the macro define could also be moved to pendsv.c, as (as of this PR) this is the only file which uses it.

Yes indeed it could be moved.  I'm happy either way (move or keep here).

projectgus: Have moved!

dpgeorge (approved):

dpgeorge:
Thanks for splitting this up into a set of very clean, independent commits.

PR #16916, opened 2025-03-12
zephyr: Allow pins to support both hard and soft interrupts.

danicampora:
The Pin irq method has an option `hard` parameter that us used to select if the IRQ callback will be called from interrupt context or not. At the moment in the Zephyr port this parameter value is saved but nothing is done with it and all pin interrupts are treated as hard. This PR presents a small change that allows to also handle interrupts in soft mode where the interrupt callback is called out of interrupt context allowing to execute bigger and less time sensitive routines.

### DIFF
--- a/ports/zephyr/machine_pin.c
+++ b/ports/zephyr/machine_pin.c
@@ -63,22 +63,26 @@ void machine_pin_deinit(void) {
 static void gpio_callback_handler(const struct device *port, struct gpio_callback *cb, gpio_port_pins_t pins) {
     machine_pin_irq_obj_t *irq = CONTAINER_OF(cb, machine_pin_irq_obj_t, callback);
 
-    #if MICROPY_STACK_CHECK
-    // This callback executes in an ISR context so the stack-limit check must be changed to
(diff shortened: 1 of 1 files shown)


danicampora:
@jonnor yes, tested it on nRF52840 and nRF5340

projectgus:
@danicampora Thanks for contributing this fix. I want to double check, are you sure that Zephyr doesn't already support both hard and soft interrupts? Take a look at the implementation of `mp_irq_handler()` in `shared/runtime/mpirq.c`.

If soft interrupts are already working, I think that it probably is still worth updating the comment in `gpio_callback_handler` that implies this won't be the case. (Does it also matter that soft irqs briefly switch the stack check and then switch it back? Maybe, although it seems like a pretty lightweight operation so probably fine.)

danicampora:
@projectgus yes you are right, they are already working, not sure how I missed that. Thanks for pointing it out.

projectgus:
> they are already working, not sure how I missed that.

The comment is written in a way that makes it sound like all interrupts are hard IRQs. I'll submit a quick PR to update it.

PR #16917, opened 2025-03-12
ports/rp2: Prevent machine.SPI settings from persisting through a soft reset

Gadgetoid:
Replaced by, original info has moved there: https://github.com/micropython/micropython/pull/16922

This has gone off the rails somewhat into a discussion about whether RP2'S SPI (and other peripherals on RP2 and other devices) should be fully reset upon a "soft reset" (in MicroPython terms this is basically just bailing out of user code, doing some teardown and set up and returning back to user code/REPL again)

To state my case in brief: A "reset" should, you know, reset things and an "init" should... "init" things.

For my part, I represent a vendor of "educational" electronics targeted at beginners/hobbyists. We have many MicroPython boards out in the wild and most of these have some self evident physical "default" bus connector: QwSt/Quiic/Stemma for I2C or SP/CE for SPI. Unsurprising and consistent behaviour of these default instances is therefore in my best interests.

### SPI Settings Persisting Across Soft Resets

Running `SPI(mosi=23)` will change the behaviour of `SPI()`, which is roughly what the documentation confirms- however this change persists across soft resets. IE: `SPI()` will use pin 23 for MOSI until a new pin is specified, or until the board is hard reset. This happens because the current SPI pin configuration is saved into static RAM (not managed by MicroPython) and not re-initialised on soft reset.

It's not obvious (read: not sufficiently documented) that this is the case. Anyone idly tinkering on the REPL or writing a script could find themselves unable to talk to their device because they've inadvertently made a persistent pin change.

I would be interested in feedback from anyone who relies on the persistence of SPI settings between soft resets since I'm struggling to find many reasons *for* it (not glitching sensitive external peripherals on reset?) but that may just be a failure of imagination on my part.

As a side-effect of this, SPI claims a small amount of RAM for these persistent instances whether you use it or not. That may not be a big deal on RP2 but reserving RAM for a peripheral that a user might not even use is an odd choice, especially in the absence of any good argumentation for why someone would want SPI settings to be persistent.

### SPI().init() Does Not Initialise Anything

`SPI().init()` (and other hardware `init` functions) have been somewhat twisted in meaning. Despite the claims of the documentation `SPI().init()` does not "init" the SPI.

It seems "init" functions across various hardware peripherals have been misappropriated to mean "make some change to a user-supplied bus." Examples include changing the baud rate on SPI or UART (perhaps supplied by a user into a driver instance) without resetting ("initialising") other values. Again the documentation doesn't sufficiently explain this practice.

As a tertiary consequence of this, SPI's "init" on RP2 specifically doesn't contain all the parameters that `SPI()` has available; chiefly you cannot configure pins with it. As the docs suggest, `SPI(id, *args, *kwargs)` should be equivalent to `SPI(id).init(*args, *kwargs)`. It is not.

### DIFF
--- a/ports/rp2/machine_spi.c
+++ b/ports/rp2/machine_spi.c
@@ -80,6 +80,11 @@
 
 #endif
 
+// Define PICO_DEFAULT_SPI for a no-args call to machine.SPI()
+#ifndef PICO_DEFAULT_SPI
+#define PICO_DEFAULT_SPI 0
+#endif
+
 // SPI0 can be GP{0..7,16..23}, SPI1 can be GP{8..15,24..29}.
 #define IS_VALID_PERIPH(spi, pin)   ((((pin) & 8) >> 3) == (spi))
 // GP{2,6,10,14,...}
@@ -128,7 +133,7 @@ static void machine_spi_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
 mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
     enum { ARG_id, ARG_baudrate, ARG_polarity, ARG_phase, ARG_bits, ARG_firstbit, ARG_sck, ARG_mosi, ARG_miso };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id,       MP_ARG_REQUIRED | MP_ARG_OBJ },
+        { MP_QSTR_id,       MICROPY_SPI_PINS_ARG_OPTS | MP_ARG_INT, {.u_int = PICO_DEFAULT_SPI} },
         { MP_QSTR_baudrate, MP_ARG_INT, {.u_int = DEFAULT_SPI_BAUDRATE} },
         { MP_QSTR_polarity, MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = DEFAULT_SPI_POLARITY} },
         { MP_QSTR_phase,    MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = DEFAULT_SPI_PHASE} },
@@ -144,7 +149,7 @@ mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
     mp_arg_parse_all_kw_array(n_args, n_kw, all_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
 
     // Get the SPI bus id.
-    int spi_id = mp_obj_get_int(args[ARG_id].u_obj);
+    int spi_id = args[ARG_id].u_int;
     if (spi_id < 0 || spi_id >= MP_ARRAY_SIZE(machine_spi_obj)) {
         mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("SPI(%d) doesn't exist"), spi_id);
     }
--- a/ports/rp2/machine_spi.c
+++ b/ports/rp2/machine_spi.c
@@ -80,6 +80,8 @@
 
 #endif
 
+#define MAX_SPI_RP2 (2)
+
 // Define PICO_DEFAULT_SPI for a no-args call to machine.SPI()
 #ifndef PICO_DEFAULT_SPI
 #define PICO_DEFAULT_SPI 0
@@ -109,27 +111,13 @@ typedef struct _machine_spi_obj_t {
     uint8_t mosi;
     uint8_t miso;
     uint32_t baudrate;
-    bool initialised;
 } machine_spi_obj_t;
 
-static machine_spi_obj_t machine_spi_obj[] = {
-    {
-        {&machine_spi_type}, spi0, 0,
-        DEFAULT_SPI_POLARITY, DEFAULT_SPI_PHASE, DEFAULT_SPI_BITS, DEFAULT_SPI_FIRSTBIT,
-        MICROPY_HW_SPI0_SCK, MICROPY_HW_SPI0_MOSI, MICROPY_HW_SPI0_MISO,
-        DEFAULT_SPI_BAUDRATE, false
-    },
-    {
-        {&machine_spi_type}, spi1, 1,
-        DEFAULT_SPI_POLARITY, DEFAULT_SPI_PHASE, DEFAULT_SPI_BITS, DEFAULT_SPI_FIRSTBIT,
-        MICROPY_HW_SPI1_SCK, MICROPY_HW_SPI1_MOSI, MICROPY_HW_SPI1_MISO,
-        DEFAULT_SPI_BAUDRATE, false
-    },
-};
+static machine_spi_obj_t *machine_spi_obj[MAX_SPI_RP2];
 
 void machine_spi_init0(void) {
-    for (int i = 0; i < MP_ARRAY_SIZE(machine_spi_obj); i++) {
-        machine_spi_obj[i].initialised = false;
+    for (uint8_t i = 0; i < MAX_SPI_RP2; i++) {
+        MP_STATE_PORT(machine_spi_obj[i]) = NULL;
     }
 }
 
@@ -150,7 +138,6 @@ enum { ARG_baudrate, ARG_polarity, ARG_phase, ARG_bits, ARG_firstbit, ARG_sck, A
 static void machine_spi_configure_helper(machine_spi_obj_t *self, mp_arg_val_t *args) {
     // Initialise the SPI instance
     self->baudrate = spi_init(self->spi_inst, args[ARG_baudrate].u_int);
-    self->initialised = true;
 
     // Set SCK/MOSI/MISO pins if configured, otherwise reset to defaults.
     int sck = self->spi_id == 0 ? MICROPY_HW_SPI0_SCK : MICROPY_HW_SPI1_SCK;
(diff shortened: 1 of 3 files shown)


Gadgetoid:
While testing this I realised that configured SPI pins survive a soft reset. That's quite a pitfall for beginners doing on device tinkering. I will add a commit to fix that, might as well go the distance with breaking `machine.SPI` on RP2 if we're going to change it at all.

Gadgetoid:
Okay after way too much looking into this and feeling like I'm losing my mind

For `SPI(n)` the docs say:

> With no additional parameters, the SPI object is created but not initialised (it has the settings from the last initialisation of the bus, if any). If extra arguments are given, the bus is initialised. See init for parameters of initialisation.

This is not strictly true. SPI is initialised if it was not already initialised since the last hard reset.

If extra parameters are given you will get a mix of default values, your custom values, and whatever state the SPI instance was last in.

Additionally `SPI(n)` accepts pin parameters that `SPI(n).init()` does not.

For `SPI(n).init()` the docs say:

> Initialise the SPI bus

It does not under any circumstances initialise the bus, though it does change any parameters that are passed to it and has a sort of defacto shorthand where you can change any property without affecting the other... this is somewhat antithetical to what people would expect from an `init` function.

The docs also suggest `init` accepts `sck, mosi, miso`. It does not.

Right now there is a significant portion of repeated code between `SPI(n)` and `SPI(n).init`

In keeping with the lean toward shorthand methods `machine.SPI` on RP2 also accepts *baudrate* as a positional argument. The docs don't really suggest the arguments should be positional or otherwise, but on RP2 the `baudrate` argument is positional to allow for things like `machine.SPI(0, 8_000_000)` .

I think that:

1. Soft reset needs to - somehow - clear the SPI instances back to their default values in all cases. Possibly using MicroPython to create a `machine_spi_obj` with default or normal values, and hang it off a root pointer.
2. `machine.SPI(n)` - should return the SPI instance in whatever state the user last configured it
3. `machine.SPI(n).init()` - should initialise any unspecified value back to its default. If the user wishes to change the baudrate on the fly we really ought to have some more formal means for this.
4. `machine.SPI(n, *args, **kwargs)` - should function exactly like `machine.SPI(n).init(*args, **kwargs)`

peterhinch:
Re 3. There is a need to be able to change the baudrate without affecting other properties. An example is touch displays where an SPI bus is shared between a display and a touch controller, where these have radically different maximum baudrates. My library uses
```py
spi.init(baudrate=new_rate)  # spi is an SPI instance
```

Gadgetoid:
That's pretty much the structure I've been moving toward in my PR, with the exception that `_new()` still doesn't properly initialise the hardware. I'm not sure we have any consensus on whether it *should* yet?

Oh and that I am being weird and passing MicroPython argument collections instead of regular C types around 😬

For a succinct example of where this goes awry consider the following:

```
from machine import SPI

test = machine.Pin(16, machine.Pin.OUT)
spi = SPI(0)

test.value(1)
print(spi)
print(test.value())
```

`test.value()` will return 0 on the first run because `SPI(0)` has mux'd GPIO 16 into SPI mode.

All subsequent runs will return 1 because `SPI(0)` will no longer initialise anything.

Quoting you from the issue:

> My approach would be that Soft Reset would reset the peripherals as well and the device is in the state it was after hard reset and setting up clocks and memory devices.

I would tend to agree.

With my current code that means adding `machine_spi_init`  which would be called in `main.c` like:

https://github.com/micropython/micropython/blob/3823aeb0f14c04084eba6566164d9b7dbd9e7ced/ports/rp2/main.c#L168-L171

And it would just need to roll through the `machine_spi_obj[]` array and set initialised (which I have added versus using baudrate == 0) to `false` to make sure the next `SPI()` call triggers a full initialisation.... I hope 😆

robert-hh:
> With my current code that means adding machine_spi_init which would be called in main.c like

Init is one aspect, deinit the other. If you look at the code in main after the label `soft_reset_exit:` you see quite a few deinit calls. These are required for devices which use memory allocated from the Python heap, interrupts and DMA. That all has to be done before the call to `gc_sweep_all()`, which resets the Python heap. A long as SPI does not use DMA or interrupts there seems nothing that need to be deinit'ed.

Gadgetoid:
I think it's conceptually an `init` since SPI isn't doing any allocation, but because it uses a static array of SPI instances that don't live on gc_heap and thus aren't reset to their default values it needs a little gentle reminder to do so.

The culprit is here:  https://github.com/micropython/micropython/blob/58f822914c25c29e6809fca0b1f8494492f67e69/ports/rp2/machine_spi.c#L115-L128

These variables shadow the SPI instance's current state and are used to make it possible to reconfigure one setting without also changing the others.

I think I should switch to the `machine_i2s` approach and hang the SPI instances off a root pointer. Very similar to what I was planning to do, but actually an established pattern (consistently consistent!) rather than some new nonsense.

robert-hh:
As far as I understand memory allocation, It is not needed to register the SPI objects with MP_STATE_PORT(). This is only needed if the allocated object is not directly returned to the calling Python code and must be protected from clearing though gc by registering it as root pointer. Not that it hurts, but it seems not to be needed.

Besides that: why did you move away from the static SPI objects? RAM shortage is not a big issue with the RPI Pico.

ports/rp2/machine_spi.c
```diff
@@ -172,7 +180,12 @@ mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
         }
         self->mosi = mosi;
     }
-    if (args[ARG_miso].u_obj != mp_const_none) {
+
+    if (args[ARG_miso].u_obj == MP_ROM_INT(-1)) {
```
robert-hh: At this state of the commit series, this if clause is wrong. If the argument is not present, the content of the static SPI object must not be changed.

robert-hh:
Looking at your PR your seem to change things which were not broken. The only required change would have been to skip the default miso pin initialization in case of `miso=None`. 

Anything else like changing the behavior of the RP2 port's behavior on re-initialization should be deferred until there is a common policy about it. And then it affects not only SPI.

Gadgetoid:
> This is only needed if the allocated object is not directly returned to the calling Python code and must be protected from clearing though gc by registering it as root pointer.

AIUI if it’s allocated on MicroPython’s heap then you must store a pointer *somewhere* since unreachable allocs will be garbage collected eventually. This could be via a root pointed, tracked alloc, or if we expect the user to hang onto the SPI instance it would happen implicitly. Since the intent is a persistent singleton this is about the best of a bad bunch of ways to do it- in all fairness at least i2s has some real cleanup to do.

It could he allocated in the C heap but insofar as MicroPython is concerned that’s a code smell if not an outright recipe for disaster (a port can be configured with no C heap and all sorts of fun ensues.)

> Besides that: why did you move away from the static SPI objects?

Because it becomes easier to reason about. The state persistence between soft resets, was a (debatably) unwanted side-effect of the static objects not being managed by MicroPython itself.

if they are allocated with “make_new” then it becomes clearer that the intent is a sort of singleton pattern rather than persistence.

> RAM shortage is not a big issue with the RPI Pico.

it isn’t until it is, arguably any static should have some strong argumentation to support why it deserves to skirt runtime allocation- be that performance or simply because it’s used 100% of the time. SPI has no supporting argumentation it’s just - afaik - the coin toss the implementer made when writing it. (Or an explicit choice to support SPI config surviving soft reset in which case they were - arguably - wrong 🤣)

Gadgetoid:
“not broken” is extremely debatable given just how many quirks and pitfalls the current code permits, including code that behaves differently after a soft reset and no path back to default values shy of unplugging a Pico.

> Anything else like changing the behavior of the RP2 port's behavior on re-initialization should be deferred until there is a common policy about it. And then it affects not only SPI.

Where’s the common policy going to come from if it’s not advocated for in a pull request with code and rationale to back it up?

(yes granted this all got a bit off piste because SPI was staggeringly weird and it seemed remiss not to address the other issues… fix creep? Is that a thing?)

dpgeorge:
I don't want to comment at the moment on the soft reset behaviour of SPI.  But I strongly suggest to split up this PR into one for the original intent (make MISO optional, make id optional) and one for the other changes.  The first of those changes can be reviewed and merged much quicker than the latter.

robert-hh:
> Since the intent is a persistent singleton this is about the best of a bad bunch of ways to do it

If that's the intent why not keep the static objects.

> if we expect the user to hang onto the SPI instance it would happen implicitly. 

That is the usual Python behavior. Since with your implementation the SPI object is cleared on soft reset, the persistence would "only" allow to re-create the SPI object with the previous configuration. That's the way I2C onRP2 is implemented, because it lacks the `init()` method, but unusual among the other ports - and kind of confusing. What is the visible difference in code between `spi=SPI()` using the default values and `spi=SPI()`, using a different configuration set in a previous call to SPI()?

> The state persistence between soft resets, was a (debatably) unwanted side-effect of the static objects not being managed by MicroPython itself.

My impression is, that this effect was wanted.

Gadgetoid:
> I don't want to comment at the moment on the soft reset behaviour of SPI.

A diplomatic choice.

> But I strongly suggest to split up this PR into one for the original intent (make MISO optional, make id optional) and one for the other changes. The first of those changes can be reviewed and merged much quicker than the latter.

Agreed. It feels a bit like building on shaky foundations 😆 but at least something will get done.

Since this PR has gone to the wolves with discussion around the former I will (rename and) leave it as-is until we resolve the questions around SPI soft reset, and raise a new PR to fix the original issue(s).

Gadgetoid:
Raised another PR for the original issues - https://github.com/micropython/micropython/pull/16922 -  and updated the OG post to state my case more succinctly.

Gadgetoid:
It's been a year and nobody else seems to want to advocate for this- a really bad sign for such a complex change - so I'm closing for now.

No great loss, the small changes spun off from here have been immensely useful!

dpgeorge:
Thanks for closing, it makes my life easier :joy:

PR #16920, opened 2025-03-13
Support error message compression on CMake ports

dpgeorge:
### Summary

This PR adds support for error message text compression on CMake ports:
- fix cases where `MP_ERROR_TEXT()` was not used to wrap error messages
- add support to `py/mkrules.cmake` for enabling compression and generating the correct header files
- enable compression by default on esp32 and rp2 ports

This reduces rp2 firmware by about 3000 bytes, and esp32 firmware by about 3300 bytes.

### Testing

Built both RPI_PICO2 and ESP32_GENERIC, deployed and used the REPL to raise a few different exceptions, the error messages were as expected.

Almost all other ports enable this option, no reason why esp32 and rp2 can't as well, to save some code size.

Didn't try to enable it on zephyr, but it should work.

### DIFF
--- a/ports/esp32/esp32_common.cmake
+++ b/ports/esp32/esp32_common.cmake
@@ -28,6 +28,11 @@ if(NOT DEFINED MICROPY_PY_TINYUSB)
     endif()
 endif()
 
+# Enable error text compression by default.
+if(NOT MICROPY_ROM_TEXT_COMPRESSION)
+    set(MICROPY_ROM_TEXT_COMPRESSION ON)
+endif()
+
 # Include core source components.
 include(${MICROPY_DIR}/py/py.cmake)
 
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Wow. TIL I should be using `MP_ERROR_TEXT` everywhere because our user C module builds have EXPLODED from this.

Feels like `mp_raise_ValueError()` should just handle this if error message compression is enabled. I guess it doesn't because the preprocessing script (I'm making an educated guess here because looking things up takes effort) only picks up and changes instances of `MP_ERROR_TEXT`.

Welp, I'd better get fixing.

dpgeorge:
> Our Tufty build saves a total of 8492 bytes!

Very good!

Issue #16912, opened 2025-03-11
When miso=None is set, the default miso pin changed state.


kaso-0:
### Port, board and/or hardware

Raspberry Pi Pico W

### MicroPython version

MicroPython v1.24.1 on 2024-11-29; Raspberry Pi Pico W with RP2040

### Reproduction

RS = Pin(16, Pin.OUT)

spi = SPI(0,
    baudrate=400000,
    polarity=0,
    phase=0,
    bits=8,
    firstbit=SPI.MSB,
    sck=Pin(18),
    mosi=Pin(19),
    miso=None) 

RS.value(0)
time.sleep_ms(1)
RS.value(1)

### Expected behaviour

Expected to set up SPI without MISO, set the RS pin value to 0, and after 1 ms, set the RS pin value to 1.

### Observed behaviour

The SPI was set up correctly, but the RS pin's value never changed.

### Additional Information

If I set up the pin after initializing SPI, it works as expected, the used pin 16 is default pin for miso. I think this bug is on other types of RPIs Pico.

### Code of Conduct

Yes, I agree

robert-hh:
The key feature here is to differentiate in the code between `no argument given` and `miso=None`. For that, both the ESP32 and SAMD port set a small negative number as default value, instead of "None" at the other arguments. Then the case `miso=None` can be detected  and treated accordingly. In case of the RP2 port this is a simple change.

Gadgetoid:
> In case of the RP2 port this is a simple change.

I think I've got a commit that does this gracefully.

I found `machine.SPI()` not resetting pins back to the default value quite confusing, so I ended up with:

```
>>> machine.SPI()
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=19, miso=16)
>>> machine.SPI(miso=None)
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=19, miso=None)
>>> machine.SPI()
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=19, miso=16)
```

Which is not the way setting up pins typically works... I might be solving a problem that I'm imagining here 😆  will raise the PR.

robert-hh:
> Even more egregious, the configured SPI pins stick through a soft reset which is a heck of a nasty bug!

[...]

PR #16922, opened 2025-03-13
ports/rp2: Optional spi_id and optional disabling of MISO pin

Gadgetoid:
### Summary

This PR includes two changes:

1. Bring `machine.SPI()` into parity with `machine.I2C()`, selecting the default instance where none is specified. See: https://github.com/micropython/micropython/pull/16671
2. Make `machine.SPI(mosi=None)` explicitly disable MISO, for configurations which use this as a register-select or data/command pin (very common for write-only LCDs). See: https://github.com/micropython/micropython/issues/16912

I avoided making this change for all pins, since it rarely makes sense for MOSI or SCLK to be disabled.

### Testing

Tested various invocations of `machine.SPI()` on a Pico 2 W.

Notably this change introduces a difference in behaviour that  *always* resets MISO back to the default unless it's specified or set to `None`. Eg:

```
>>> machine.SPI()
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=19, miso=16)
>>> machine.SPI(miso=None)
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=19, miso=None)
>>> machine.SPI()
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=19, miso=16)
```

This is materially different from, for example, mosi where it's impossible to implicitly set a pin back to its default value:

```
>>> machine.SPI(mosi=23)
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=23, miso=16)
>>> machine.SPI()
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=23, miso=16)
>>> machine.SPI(mosi=None)
SPI(0, baudrate=1000000, polarity=0, phase=0, bits=8, sck=18, mosi=23, miso=16)
```

🤔 Maybe this changeset should also require that a pin is specified or that it is otherwise always set back to its default value.

The latter change is more about explicit vs implicit. It's possibly just to set upt he MISO pin *after* setting up SPI and it will work as expected. A documentation change establishing this as the accepted pattern might be an alternative... though it might not be possible in all cases for the pin setup to happen before SPI.

Consider the following very crude example driver pattern where a user wishes to save a pin (normally eaten by SPI) and reuse it for register select:

```python
class LCD:
    def __init__(self, spi=None, rs=None):
        self._spi = spi or machine.SPI()

my_led_screen = LCD(rs=machine.Pin(16))
```

However because the pin is constructed before it's passed into the driver class, its mux will be overwritten by `machine.SPI()`

With this change, a driver author could use:

```python
class LCD:
    def __init__(self, spi=None, rs=None):
        self._spi = spi or machine.SPI(miso=None)

my_led_screen = LCD(rs=machine.Pin(16))
```

And SPI would no longer trample the "rs" pin config by setting it up as MISO.

### DIFF
--- a/ports/rp2/machine_spi.c
+++ b/ports/rp2/machine_spi.c
@@ -128,7 +128,11 @@ static void machine_spi_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
 mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
     enum { ARG_id, ARG_baudrate, ARG_polarity, ARG_phase, ARG_bits, ARG_firstbit, ARG_sck, ARG_mosi, ARG_miso };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id,       MP_ARG_REQUIRED | MP_ARG_OBJ },
+        #ifdef PICO_DEFAULT_SPI
+        { MP_QSTR_id,       MP_ARG_INT, {.u_int = PICO_DEFAULT_SPI} },
+        #else
+        { MP_QSTR_id,       MP_ARG_REQUIRED | MP_ARG_INT, },
+        #endif
         { MP_QSTR_baudrate, MP_ARG_INT, {.u_int = DEFAULT_SPI_BAUDRATE} },
         { MP_QSTR_polarity, MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = DEFAULT_SPI_POLARITY} },
         { MP_QSTR_phase,    MP_ARG_KW_ONLY | MP_ARG_INT, {.u_int = DEFAULT_SPI_PHASE} },
@@ -144,7 +148,7 @@ mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
     mp_arg_parse_all_kw_array(n_args, n_kw, all_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
 
     // Get the SPI bus id.
-    int spi_id = mp_obj_get_int(args[ARG_id].u_obj);
+    int spi_id = args[ARG_id].u_int;
     if (spi_id < 0 || spi_id >= MP_ARRAY_SIZE(machine_spi_obj)) {
         mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("SPI(%d) doesn't exist"), spi_id);
     }
--- a/ports/rp2/machine_spi.c
+++ b/ports/rp2/machine_spi.c
@@ -80,6 +80,9 @@
 
 #endif
 
+// Assume we won't get an RP2-compatible with 255 GPIO pins
+#define MICROPY_HW_SPI_PIN_UNUSED UINT8_MAX
+
 // SPI0 can be GP{0..7,16..23}, SPI1 can be GP{8..15,24..29}.
 #define IS_VALID_PERIPH(spi, pin)   ((((pin) & 8) >> 3) == (spi))
 // GP{2,6,10,14,...}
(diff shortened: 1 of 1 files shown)


ports/rp2/machine_spi.c
```diff
@@ -167,7 +180,12 @@ mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
         }
         self->mosi = mosi;
     }
-    if (args[ARG_miso].u_obj != mp_const_none) {
+
+    if (args[ARG_miso].u_obj == MP_ROM_INT(-1)) {
+        self->miso = spi_id == 0 ? MICROPY_HW_SPI0_MISO : MICROPY_HW_SPI1_MISO;
```
dpgeorge, line 185: In this case of -1, `miso` is not given and so the pin should not change.

The default is already set in the static SPI instance, so this conditional is not needed.

ports/rp2/machine_spi.c
```diff
@@ -120,31 +128,36 @@ static machine_spi_obj_t machine_spi_obj[] = {
...
+    } else {
+        mp_printf(print, "%u)", self->miso);
+    }
 }
 
 mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
     enum { ARG_id, ARG_baudrate, ARG_polarity, ARG_phase, ARG_bits, ARG_firstbit, ARG_sck, ARG_mosi, ARG_miso };
     static const mp_arg_t allowed_args[] = {
-        { MP_QSTR_id,       MP_ARG_REQUIRED | MP_ARG_OBJ },
+        { MP_QSTR_id,       MICROPY_SPI_PINS_ARG_OPTS | MP_ARG_INT, {.u_int = PICO_DEFAULT_SPI} },
```
dpgeorge, line 144: IMO this should follow commit bb4ec886f8994ad7b2db6f80c7b46e6eedbc40e7

Gadgetoid: Agreed, though-

Interestingly that commit looks wrong, if `PICO_DEFAULT_I2C` is not set then rather than requiring `i2c_id` it will default to `-1` and `machine.I2C()` will fail with an `"I2C(-1) doesn't exist` error?

The second form should require the argument.

dpgeorge: Yeah, it probably should add `MP_ARG_REQUIRED`.

ports/rp2/machine_spi.c
```diff
@@ -80,6 +80,14 @@
 
 #endif
 
+// Define PICO_DEFAULT_SPI for a no-args call to machine.SPI()
+#ifndef PICO_DEFAULT_SPI
+#define PICO_DEFAULT_SPI 0
+#endif
```
dpgeorge, line 86: Not needed anymore.

Gadgetoid: 🤦 

Wood for the trees. Thank you.

ports/rp2/machine_spi.c
```diff
@@ -167,7 +179,10 @@ mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
         }
         self->mosi = mosi;
     }
-    if (args[ARG_miso].u_obj != mp_const_none) {
+
+    if (args[ARG_miso].u_obj == mp_const_none) {
+        self->miso = MICROPY_HW_SPI_PIN_UNUSED;
+    } else {
```
dpgeorge, line 185: I think this line needs to be `} else if (args[ARG_miso].u_obj != MP_OBJ_NEW_SMALL_INT(-1)) {`.

Otherwise in the default case (miso not specified) it'll try to access pin `-1` and raise an exception.

Gadgetoid: Ooof, you are correct- I think this got lost in the chaos of the other PR.

Indeed -1 is the new "not specified" and `None` is "do not use."

Fixed.

dpgeorge (approved):

dpgeorge:
Rebased and merged in a86122396d0cdb98ed0e72a4c208d7491581e3fa and dd7a950bbc27713df8cc461958757ec793189d72

PR #16927, opened 2025-03-13
rp2/machine_i2c: Require an I2C bus ID when no default is available.

Gadgetoid:
When PICO_DEFAULT_I2C is not set require an I2C bus ID instead of using -1 as a default, which would fail with a cryptic:

```
    "I2C(-1) doesn't exist"
```

Picked up in review of #16922


dpgeorge (approved):

dpgeorge:
Thanks, rebased and merged in f315a376b616fb195a18700a724715f0e7d250e3

PR #16939, opened 2025-03-14
extmod/vfs: Return mount table from no-args mount call.

AJMansfield:
### Summary

This is a more minimal version of #16849, that overloads `vfs.mount` to output the current vfs mount table when called with no arguments.

### Testing

This PR includes full automated tests; I have verified that these tests pass with the feature code and fail without it.

In order to overload `vfs.mount` to accept a zero-argument call, it's necessary to change the two existing arguments from fixed positional arguments into dynamic positional arguments handled by `mp_arg_parse_all`.

I've re-defined them as required nameless arguments in the table for now so that their current positional-only behavior doesn't change:

```
        { MP_QSTR_, MP_ARG_REQUIRED | MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
        { MP_QSTR_, MP_ARG_REQUIRED | MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
```

This avoids introducing new QSTRs for the argument names `fsobj` and `mount_point` and avoids committing to those specific argument names as a visible part of the API (i.e. `mount(fsobj=..., mount_point=...)` remains an error). This may be a design decision worth revisiting (perhaps along with the argument names themselves).

### DIFF
--- a/docs/library/vfs.rst
+++ b/docs/library/vfs.rst
@@ -34,6 +34,14 @@ represented by VFS classes.
 
     Will raise ``OSError(EPERM)`` if *mount_point* is already mounted.
 
+.. function:: mount()
+    :noindex:
+
+    With no arguments to :func:`mount`, return a list of tuples representing
+    all active mountpoints.
+    
+    The returned list has the form *[(fsobj, mount_point), ...]*.
+
 .. function:: umount(mount_point)
 
     Unmount a filesystem. *mount_point* can be a string naming the mount location,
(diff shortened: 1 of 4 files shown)


dpgeorge (approved):
This looks good!  And it's nice how it mirrors the Linux `mount` command, which with no arguments outputs a list of mounted filesystems.

Tested on PYBD_SF2 (with a ROMFS mounted) and RPI_PICO_W (which mounts at the root).

PR #16945, opened 2025-03-17
extmod/moddeflate: Keep DeflateIO state consistent on window alloc fail.

dpgeorge:
### Summary

Allocation of a large compression window may fail, and in that case keep the `DeflateIO` state consistent so its other methods (such as `close()`) still work.  Consistency is kept by only updating the `self->write` member if the window allocation succeeds.

Thanks to @jimmo for finding the bug.

### Testing

A new test is added, will run under CI.

### DIFF
--- a/py/dynruntime.h
+++ b/py/dynruntime.h
@@ -28,6 +28,14 @@
 
 // This header file contains definitions to dynamically implement the static
 // MicroPython runtime API defined in py/obj.h and py/runtime.h.
+//
+// All of the symbols made available in this header are overriding those defined
+// in py/obj.h and py/runtime.h.  This is done with macros.  For macros that
+// would be too complicated (usually more than a single expression), they call a
+// static-inline function for the implementation.  This function has the same
+// name as the macro (hence the same name as a public API function) but with
+// "_dyn" appended. For example, the m_malloc() macro calls the m_malloc_dyn()
+// static-inline function.
 
 #include "py/binary.h"
 #include "py/nativeglue.h"
@@ -37,6 +45,10 @@
 
 #if !MICROPY_ENABLE_DYNRUNTIME
 #error "dynruntime.h included in non-dynamic-module build."
+#endif
+
+#if MICROPY_MALLOC_USES_ALLOCATED_SIZE
+#error "MICROPY_MALLOC_USES_ALLOCATED_SIZE must be disable in a dynamic-module build."
 #endif
 
 #undef MP_ROM_QSTR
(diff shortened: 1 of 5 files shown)


dpgeorge:
CI is a little tricky to get it to all pass here.

tests/extmod/deflate_compress_memory_error.py
```diff
@@ -0,0 +1,39 @@
...
+# Then, use up most of the heap.
+l = []
+while True:
+    try:
+        l.append(bytearray(1000))
+    except:
+        break
+l.pop()
+
+# Try to compress.  This will try to allocate a large window and fail.
```
projectgus, line 29: Could you do this test by locking the heap before calling `g.write()`, instead? (My quick read of moddeflate doesn't show up any temporarily allocation that would fail first, but maybe one exists?)

py/dynruntime.h
```diff
@@ -52,22 +52,35 @@
 /******************************************************************************/
 // Memory allocation
 
+#define m_malloc_fail(num_bytes)        (m_malloc_fail_dyn((num_bytes)))
```
projectgus, line 55: This is kind of unrelated to this fix, but do these have to be macros here? It seems like the inline function could be named `m_malloc_fail`? Or is this necessary to "override" the default functions with these names?

(If there is a reason, suggest putting it in a comment.)

tests/extmod/deflate_compress_memory_error.py
dpgeorge: That's a nice idea, but it won't work: the existing code (on master) that has a crash does a small alloc first of `mp_obj_deflateio_write_t` and then the large alloc of the window.  You need the first to pass -- so the code sets `self->write` -- and then the second to fail.  In that case the object is in a bad state, with `self->write` non-NULL yet the window unallocated.

py/dynruntime.h
dpgeorge: The macro is necessary, because the existing `m_malloc_fail()` is declared as non-static, and so it'll clash here if declared `static inline`.

This pattern (using `xxx_dyn` functions with wrapper macros) is used a lot in this file, so any comment would need to apply to the whole file.

projectgus: Yes, a comment at the top of this file that explains what it's for and why everything follows this pattern would be really useful :)

tests/extmod/deflate_compress_memory_error.py
projectgus: I see, I was only thinking about the test in this PR branch where it does the large allocation first so this would fail now. I agree it's not strictly testing the same thing because on master it won't trigger the bug, though.

dpgeorge: It's a tricky one.

I guess the best way to look at is this test prevents a regression.  Eg if someone comes along and reorders the logic here to allocate the window later on in the function, that test will fail.

projectgus (approved):

py/dynruntime.h
dpgeorge: I've now added a comment, and slightly improved the new macros (a zero-byte allocation should not raise an exception even though it returns NULL).

projectgus: Thanks!

PR #16946, opened 2025-03-17
ports/rp2/boards: Fix SparkFun vendor name.

malcolm-sparkfun:
### Summary

Apologies as this is a bit of a small/pedantic change. With merge of the new SparkFun IoT RedBoard RP2350 board definition files which spell our vendor name correctly as "SparkFun", we noticed that the MicroPython downloads page is case-sensitive and that all previous SparkFun boards have used our name spelled as "Sparkfun". This led to two separate vendor links on the downloads page:

- [https://micropython.org/download/?vendor=SparkFun](https://micropython.org/download/?vendor=SparkFun)
- [https://micropython.org/download/?vendor=Sparkfun](https://micropython.org/download/?vendor=Sparkfun)

One of which only has the new IoT RedBoard RP2350. Our preferred/correct spelling is "SparkFun" so this PR updates all of our existing board definitions with that spelling such that everything will show on the same page.

### Testing

Did not test building or running as this is an "aesthetic" change.

Alternative could be to switch only the board.json for the new IoT Redboard RP2350 although that is not our preference as it is actually the only one with the correct spelling. Trade-off is that this will break the old link that we have sent to some users.

### DIFF
--- a/ports/rp2/boards/SPARKFUN_IOTNODE_LORAWAN_RP2350/board.json
+++ b/ports/rp2/boards/SPARKFUN_IOTNODE_LORAWAN_RP2350/board.json
@@ -21,5 +21,5 @@
     "product": "IoT Node LoRaWAN RP2350",
     "thumbnail": "",
     "url": "https://www.sparkfun.com/products/26060",
-    "vendor": "Sparkfun"
+    "vendor": "SparkFun"
 }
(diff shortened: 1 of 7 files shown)


dpgeorge (approved):
Looks good to me.

dpgeorge:
Rebased and merged in 1e92bdd206f6f87ba65ea05c5b2623fca0b926cd

--------

Doing `git grep Sparkfun` in this repo shows a few other instances that you may want to change, in a follow-up PR.

PR #16948, opened 2025-03-17
examples/embedding: Build embed sources in the same make command.

AJMansfield:
### Summary

This updates the example for the embed port to streamline the build process, reducing the number of commands needed to build it from 2 to 1.

In specific:
- The original `Makefile` is moved aside to become `makefile.mk`.
- It's replaced with a new `Makefile` that invokes `micropython_embed.mk` and `makefile.mk` as recursive sub-makes.

This indirection makes sure that the `SRC += $(wildcard $(EMBED_DIR)/*/*.c) $(wildcard $(EMBED_DIR)/*/*/*.c)` variable expansion is performed after the sources it's meant to pick up already exist. If the `mpy` recipe were added directly to the original makefile, these files wouldn't be picked up on first invocation and it would prevent the example from being built directly from clean.

### Testing

This commit updates `.github/workflows/examples.yml` to reflect the `Makefile -> makefile.mk` rename, and adds tests of the combined single-command clean and build operations. I've also separately tested building the example from a number of different clean and dirty build states.

This is a documentation change (with automated tests that verify the documentation's claims).
The minor increase in technical complexity could subtly impact the example's pedagogical utility, but the positive utility of demonstrating a complete combined source-gen and compilation build should outweigh that.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -20,6 +20,12 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Build
-      run: make -C examples/embedding -f micropython_embed.mk && make -C examples/embedding
+      run: make -C examples/embedding -f micropython_embed.mk && make -C examples/embedding -f makefile.mk 
     - name: Run
+      run: ./examples/embedding/embed | grep "hello world"
+    - name: Clean
+      run: make -C examples/embedding clean && [ ! -e embed ] && [ ! -e micropython_embed ]
+    - name: Combined Build
+      run: make -C examples/embedding
+    - name: Combined Run
       run: ./examples/embedding/embed | grep "hello world"
--- a/examples/embedding/.gitignore
+++ b/examples/embedding/.gitignore
@@ -1,0 +1,3 @@
+micropython_embed/
+embed
+*.o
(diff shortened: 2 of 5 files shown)


AJMansfield:
Coverage change seems to be spurious?

projectgus (commented):
Hi @AJMansfield,

Thanks for patient while someone took a look at this.

I see that this PR streamlines the embedding example so you can build the example in a single command. 

However, the purpose of this example is not really that - it's to get a set of files and a very simple build process that someone can put into a totally different project. Can you please walk me through the improvement from the point of view of that workflow?

AJMansfield:
> However, the purpose of this example is not really that - it's to get a set of files and a very simple build process that someone can put into a totally different project. Can you please walk me through the improvement from the point of view of that workflow?

There's a somewhat common pitfall I've seem people fall into when helping troubleshoot uses of the embed target (as in, I've seen it twice), where they'll try to modify the upstream MicroPython source, without realizing they've set up their Makefile in a way that requires them to manually re-build MicroPython's embedded source before rerunning the build process for the rest of their project.

I'm not dead-set on the idea that demonstrating how to avoid that problem is _definitely_ worth the extra complexity though. There's merit to keeping the example Makefile extra simple, just depends on what we understand the goal of that example to be.

PR #16953, opened 2025-03-18
py: Fixes and test coverage for 64-bit big integer representations.

### LINKED ISSUE #16932: Incorrect parse of large integers in LONGLONG mode

any port configured for MICROPY_LONGINT_IMPL_LONGLONG (tested on Unix, Windows and ARM Cortex M4)

### MicroPython version

MicroPython v1.25.0-preview.389.gf187c77da.dirty on 2025-03-14; linux [GCC 13.3.0]

### Reproduction

The LONGLONG implementation of `mp_obj_new_int_from_str_len `is broken. It does not take the sign into account, and assumes the string buffer is nul-terminated, which is wrong as `mp_parse_num_integer` is mostly used on vstr buffers.

To reproduce, make a standard Unix build with
```
#define MICROPY_LONGINT_IMPL                    MICROPY_LONGINT_IMPL_LONGLONG
```

1. Try to evaluate
```
json.loads("-9111222333444555666")
```
It should return a negative number, but returns a positive number.

2. Try to evaluate
```
["9111222333444555666777", 9111222333444555666]
```
It fails with `SyntaxError: invalid syntax for integer with base 10: '9111222333444555666'` 
Removing the first element of array will make the number parse properly.
This issue can be reproduced using `json.loads()` as well.

Test cases that exhibit the issues have been added to `tests/extmod/json_loads.py` and submitted as PR https://github.com/micropython/micropython/pull/16931

### Expected behaviour

1. Should return `-9111222333444555666`
2. Should return `["9111222333444555666777", 9111222333444555666]`

### Observed behaviour

1. Returns 9111222333444555666
2. Fails with `SyntaxError: invalid syntax for integer with base 10: '9111222333444555666'` 

### Additional Information

Replacing the call to `strtoll` by code similar to `mp_parse_num_integer` but using a `long long` to accumulate the result does fix the issue. It might however be more efficient in terms of code size to use directly a `long long` in `mp_parse_num_integer` when compiled with `MICROPY_LONGINT_IMPL_LONGLONG`

### Code of Conduct

Yes, I agree

projectgus:
### Summary

As pointed out by @yoctopuce [in this discussion](https://github.com/orgs/micropython/discussions/16933) and #16932, there is poor test coverage for the optional 64-bit bigint representation and this representation has some bugs.

This PR aims to improve this:

1. Add test naming convention specifically for 64-bit big integers. These will run if there is any big integer support (either 64-bit or arbitrary precision).
2. Duplicate the basic bigint tests to add `int_64` equivalent.
3. Fix bug where negative 64-bit integers were incorrectly parsed.
4. Fix bug where 64-bit integer parsing produced invalid results if the buffer wasn't null terminated and the byte after the buffer was a valid digit. Fixes #16932. This incorporates the tests submitted in #16931 which seem to be a reliable way to get a string buffer which fits this edge case. However, the new tests are moved to a separate file so that the json tests don't depend on bigint support.
5. Add a `longlong` unix build variant that enables 64-bit long int mode. Mostly useful for CI testing.
6. Fix saturating behaviour when 64-bit integer parsing overflows, now it fails instead. This needed further filtering of tests as the ffi_int tests all depend on parsing UINT64_MAX or similar. This happened to work before this check was added, I believe as they got cast back to uint64 when passing to the FFI interface.
7. Raise OverflowError if an arithmetic operation overflows 64-bit signed integer. Uses built-ins on gcc & clang, hand-rolled checks on other compilers.
8. Change `mp_parse_num()` to parse directly to `long long` in the 64-bit big integer configuration, saving code size.

Thanks to @yoctopuce for suggesting and demonstrating a bunch of these ideas, and improvements on the original version of this PR.

### Testing

* Ran unit tests for unix port 'standard', 'longlong', and a special build of 'standard' where the compiler built-in overflow functions were disabled (gcc 15.1 seems to produce the same code in this case, the compiler must recognise that the polyfill versions are equivalents - but doing this confirmed it). (Note: I haven't re-run the special build on the very latest version of this PR, but none of that code has changed.)
* Ran unit tests for stm32 PYBD_SF2, NUCLEO_H723ZG with NANBOX=1, and esp32 ESP32_GENERIC_S3 - both default tests and `--via-mpy`. All passing.

### Follow-up work (for new PRs)

* Extend `longlong` config to also test `MICROPY_OBJ_REPR_C`.
* Fix bug when adding a float to a 64-bit big integer.

* We could deprecate 64-bit big integers instead of improving support for it, but it does seem useful in small systems.


py/objint_longlong.c
```diff
@@ -265,18 +266,51 @@ mp_obj_t mp_obj_new_int_from_ll(long long val) {
...
     if (val >> (sizeof(unsigned long long) * 8 - 1) != 0) {
         mp_raise_msg(&mp_type_OverflowError, MP_ERROR_TEXT("ulonglong too large"));
     }
     return mp_obj_new_int_from_ll(val);
 }
 
 mp_obj_t mp_obj_new_int_from_str_len(const char **str, size_t len, bool neg, unsigned int base) {
-    // TODO this does not honor the given length of the string, but it all cases it should anyway be null terminated
-    // TODO check overflow
+    char temp_buf[65]; // Large enough to hold 64-bit base 2 str + NUL terminating byte
```
projectgus, line 276: This is a reasonably big hit on the stack, so for a size/speed trade-off then we could malloc the temp buffer instead. It's only used when parsing in-place fails, which is rare but not impossible.

ports/unix/variants/longlong/mpconfigvariant.h
```diff
@@ -0,0 +1,37 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+// This config exists to test that the MICROPY_LONGINT_IMPL_LONGLONG variant
+// (i.e. minimal form of "big integer" that's backed by 64-bit int only) builds
+// and passes tests.
+
+#define MICROPY_LONGINT_IMPL           (MICROPY_LONGINT_IMPL_LONGLONG)
+
```
projectgus, line 32: @yoctopuce suggested we test `MICROPY_OBJ_REPR_C` and `MICROPY_LONGINT_IMPL_LONGLONG` together, which makes sense - and I don't think we have CI coverage for the former at the moment.

@dpgeorge what do you think?

projectgus:
> * The only way to correctly detect overflow without using `strtoll` and `errno` would be to add a dependency on `sscanf` or to re-implement `strtoll` ourselves

The other thing that might be worth considering here is 64-bit arithmetic overflow, should this raise OverflowError? (Currently arithmetic operations return the overflow result, which is UB in C - but I think on all our architectures it will wrap.)

projectgus:
This is a bit troubling, my local `longlong` builds pass tests but now the CI build is failing them. And I think the CI test run definitely got further than this before...

py/objint_longlong.c
yoctopuce: The other option would be to borrow the parse code from `mp_parse_num_integer`, but using
```long long int_val = 0;````
Given that this function is the single place in micropython where `strtoll` is used, the overall code footprint might actually be smaller.

**Comment by yoctopuce** [comment]
> This is a bit troubling, my local `longlong` builds pass tests but now the CI build is failing them. And I think the CI test run definitely got further than this before...

The code only fails when run through mpy-cross (test option `--via-mpy`)
The cause is the very last test, which is causing an undetected overflow:
```
x = 1 << 63
```
2^63 is not a valid positive `long long`, but it is silently interpreted as -2^63 when interpreted directly in MicroPython.

When using `--via-mpy`, `mpy-cross` uses `mpz` bigints and can therefore create valid a constant object with the true decimal representation of 2^63 (thanks to `MICROPY_COMP_CONST_FOLDING`). When the `mpy` file is loaded, the runtime engine fails to create the constant object and fails with the ValueError....

**Comment by projectgus** [comment]
> The code only fails when run through mpy-cross (test option `--via-mpy`) The cause is the very last test, which is causing an undetected overflow:

Ah of course, thanks!

**Comment by projectgus** [comment]
> 2^63 is not a valid positive `long long`, but it is silently interpreted as -2^63 when interpreted directly in MicroPython.

I feel like this difference in behaviour is a good example of why we might consider changing the 64-bit arithmetic operations to raise on overflow.

**Comment by yoctopuce** [comment]
> > 2^63 is not a valid positive `long long`, but it is silently interpreted as -2^63 when interpreted directly in MicroPython.
> 
> I feel like this difference in behaviour is a good example of why we might consider changing the 64-bit arithmetic operations to raise on overflow.

I'm totally with you

**Comment by projectgus** [comment]
This PR now includes arithmetic overflow checks for 64-bit bigints!

**Comment by yoctopuce** [comment]
> This PR now includes arithmetic overflow checks for 64-bit bigints!

Excellent, thank you !

File: ports/unix/variants/longlong/mpconfigvariant.h (hunk 2 above)
Comment (dpgeorge) [reply ports/unix/variants/longlong/mpconfigvariant.h:32]: Yes, that's a very good idea to test `MICROPY_OBJ_REPR_C` in CI.

It's enabled on esp8266 and stm32 Arduino boards, so it does get tested for a release.  But would be great to test it in CI.

Enabling representation C in this new longlong variant is a good idea!

File: py/objint_longlong.c (hunk 1 above)
Comment (dpgeorge) [reply py/objint_longlong.c:321]: This 65-byte temp buffer is fine.  We allocate a lot bigger things in certain places (eg FAT FS temp buffers can be up to 4k!).

In general it's best in MicroPython to make good use of the C stack, rather than the heap, so that operations can be done more deterministically.  That means allocating a relatively large C stack, which gives room to use buffers like these.

File: py/objint_longlong.c [hunk 3]
```diff
@@ -25,6 +25,7 @@
  * THE SOFTWARE.
  */
 
+#include <errno.h>
```
dpgeorge, line 28: Yeah, it's a bit of a nuisance to need to include this header... it sometimes doesn't exist on embedded systems.

At the very least I'd suggest moving this include to within the `MICROPY_LONGINT_IMPL == MICROPY_LONGINT_IMPL_LONGLONG` guard, so it's only included if needed (because this file will be compiled on all systems regardless of the config).

In the long run, would be best to remove dependency on `errno`, but I'm not sure it's worth the effort given this config is likely not used much.

py/objint_longlong.c
```diff
@@ -144,9 +148,22 @@ mp_obj_t mp_obj_int_unary_op(mp_unary_op_t op, mp_obj_t o_in) {
     }
 }
 
+// 64-bit binary operations have overflow checking when supported by the compiler,
+// multiply is used in a few places
+#if __has_builtin(__builtin_mul_overflow)
```
dpgeorge, line 153: I think this needs to be wrapped in `#if defined __has_builtin`.  See `py/misc.h` and https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fbuiltin.html

dpgeorge: Or follow `lib/cmsis/inc/cmsis_gcc.h`:
```c
#ifndef __has_builtin
  #define __has_builtin(x) (0)
#endif
```

py/objint_longlong.c
```diff
@@ -43,6 +44,10 @@
 const mp_obj_int_t mp_sys_maxsize_obj = {{&mp_type_int}, MP_SSIZE_MAX};
 #endif
 
+static void raise_long_long_overflow(void) {
+    mp_raise_msg(&mp_type_OverflowError, MP_ERROR_TEXT("result overflows signed 64-bit"));
```
dpgeorge, line 48: `long long` is at least 64-bits, but could possibly be larger (is that every true in practice? not sure).  So maybe the message can be "result overflows long long storage"?

py/objint_longlong.c
yoctopuce: I also hit that __has_builtin issue when integrating this commit in my dev branch, and as `__has_builtin()` is sometime undefined on compilers that do have support for `__builtin_*_overflow`, I ended up using this code:

```c
#if defined(_MSC_VER)
#define ENABLE_INTSAFE_SIGNED_FUNCTIONS
#include <intsafe.h>
#define add_overflow(lhs,rhs,res)   LongLongAdd(lhs, rhs, res)
#define sub_overflow(lhs,rhs,res)   LongLongSub(lhs, rhs, res)
#define mul_overflow(lhs,rhs,res)   LongLongMult(lhs,rhs,res)
#elif defined(__EMSCRIPTEN__ ) || (defined(__GNUC__) && __GNUC__ >= 5)
#define add_overflow(lhs,rhs,res)   __builtin_add_overflow(lhs, rhs, res)
#define sub_overflow(lhs,rhs,res)   __builtin_sub_overflow(lhs, rhs, res)
#define mul_overflow(lhs,rhs,res)   __builtin_mul_overflow(lhs, rhs, res)
#endif

#if !defined(add_overflow)
#include <limits.h>

__inline bool add_overflow(long long lhs, long long rhs, long long *result) {
    // Check for overflow
    if ((rhs > 0) && (lhs > LLONG_MAX - rhs)) {
        return 1; // Overflow
    }
    if ((rhs < 0) && (lhs < LLONG_MIN - rhs)) {
        return 1; // Overflow
    }
    *result = lhs + rhs;
    return 0; // No overflow
}

__inline bool sub_overflow(long long lhs, long long rhs, long long *result) {
    // Check for overflow
    if ((rhs > 0) && (lhs < LLONG_MIN + rhs)) {
        return 1; // Overflow
    }
    if ((rhs < 0) && (lhs > LLONG_MAX + rhs)) {
        return 1; // Overflow
    }
    *result = lhs - rhs;
    return 0; // No overflow
}

__inline bool mul_overflow(long long lhs, long long rhs, long long * result)
{
    if (lhs > 0) {
        if (rhs > 0) {
            if (lhs > (LLONG_MAX / rhs)) {
                return 1; // Overflow
            }
        } else {
            if (rhs < (LLONG_MIN / lhs)) {
                return 1; // Overflow
            }
        }
    } else {
        if (rhs > 0) {
            if (lhs < (LLONG_MIN / rhs)) {
                return 1; // Overflow
            }
        } else {
            if ((lhs != 0) && (rhs <= (LLONG_MAX / lhs))) {
                return 1; // Overflow
            }
        }
    }
    *result = lhs * rhs;
    return 0; // No overflow
}
#endif
```

My embedded platforms are all using GCC >= 5, so the bigger+slower inline code only ends up in OS-based test builds.

projectgus: Impressive! @dpgeorge do you think it's worth pulling any/all of this into MP, or stick with `__has_builtin` and fallback to not checking overflow? (`__has_builtin` is [GCC 10 and up](https://stackoverflow.com/a/63668217), first released 2020.)

projectgus: (I've added the check for `defined __has_builtin` already.)

py/objint_longlong.c
projectgus: Ah C, so many little pitfalls... (Changed.)

tools/ci.sh
```diff
@@ -660,6 +660,15 @@ function ci_unix_nanbox_run_tests {
     ci_unix_run_tests_full_helper nanbox PYTHON=python2.7
 }
 
+function ci_unix_longlong_build {
+    ci_unix_build_helper VARIANT=longlong
+    ci_unix_build_ffi_lib_helper gcc -m32
```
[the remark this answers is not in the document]
projectgus: good catch, this is copy-pasta.

py/objint_longlong.c
projectgus: I've moved it down behind the preprocessor guard.

I did start writing a version that doesn't depend on strtoll, using the approach @yoctopuce suggested for `mp_parse_num_integer()` to handle 64-bit. It has potential to work quite nicely, but it wasn't particularly simple either so I shelved it for now.

yoctopuce: Here is my version, in case it helps:
```c
    const byte* restrict str = (const byte*)*str_;
    const byte* restrict top = str + len;
    long long int_val = 0;
    for (; str < top; str++) {
        // get next digit as a value
        mp_uint_t dig = *str;
        if ('0' <= dig && dig <= '9') {
            dig -= '0';
        } else if (dig == '_') {
            continue;
        } else {
            dig |= 0x20; // make digit lower-case
            if ('a' <= dig && dig <= 'z') {
                dig -= 'a' - 10;
            } else {
                // unknown character
                break;
            }
        }
        if (dig >= (mp_uint_t)base) {
            break;
        }

        // add next digit
        int_val = int_val * base + dig;
    }

    // negate value if needed
    if (neg) {
        int_val = -int_val;
    }
    mp_obj_t result = mp_obj_new_int_from_ll(int_val);
    *str_ = (char*)str;
```

projectgus: I got inspired to go and finish the version of this I was working on, have updated this PR with the result.

py/objint_longlong.c
yoctopuce: If we rely only on `__has_builtin`, we are going to significantly increase code size for all compiler version between 2015 ( GCC 5+) and 2020 (GCC 10+).

As `__builtin_*_overflow` functions have been officially included in GCC 5, you might want to override the default "not supported" setting by checking `(defined(__GNUC__) && __GNUC__ >= 5)`

projectgus: The other option is to disable the overflow check on those compiler versions (which is what the earlier version of this PR did, although the latest version disables the overflow check for some options and keeps it for multiply due to integer parsing.)

@yoctopuce I don't suppose you have any measurement of the difference? There are some cases where the compiler can peephole optimise in a native overflow check from equivalent C code (but unsure if this is one of those cases).

yoctopuce: Hi, I assumed that the builtin would be better but did not check before.
I have double-checked the generated code today and this confirms that the builtin code is **a long way** better, as it looks like the peephole optimiser cannot guess what this code is doing around the overflow limits.

Here is the generated code for my explicit "add_overflow" function, compiled using  `arm-none-eabi-gcc.exe (GNU Tools for Arm Embedded Processors 9-2019-q4-major) 9.2.1 20191025 (release) [ARM/arm-9-branch revision 277599]`
explicit_add
And here is the code generated by `builtin_overflow_add`:
builtin_add

The code for my explicit "mul_overflow" is too big to fit in a screenshot, while "builtin_mul_overflow" is simply:
builtin_mul

In total, using the "builtin_*_overflow" on GCC 9 saves me 76 bytes of code. It is not a lot, but for me it is clearly worth testing `(defined(__GNUC__) && __GNUC__ >= 5)` to decide if builtin_*_overflow should be used

projectgus: > If we rely only on `__has_builtin`, we are going to significantly increase code size for all compiler version between 2015 ( GCC 5+) and 2020 (GCC 10+).

I think the impact is not that bad. If I build the Zephyr `nucleo_wb55rg` board with this branch (and gcc 12) I get 242792 bytes flash size.

If I manually patch the three `#if __has_builtin(*overflow)` lines to `#if 0` then I get 242800 bytes flash size.

Admittedly, there is now no overflow check for addition and subtraction *and* it's possible GCC 12 has peephole optimised the manual `mp_mul_ll_overflow` function to an overflow check in a way that older GCC can't detect.

It is a bit odd that GCC version determines the overflow check, though. I'll see how hard it is to patch around that further.

projectgus: I've updated the implementation so the overflow checks are consistent regardless of whether `__has_builtin(*overflow)` is true or not. Based on your code, so thanks for that! I *think* my version of the manual implementations are all sound...

The Zephyr build still only goes up by 8 bytes in code size if I swap from the built-in overflow checked ops to the manual ones, though. Even on GCC 5.x the difference in size seems neglible: https://gcc.godbolt.org/z/jhhn16dW6 (One key difference may be that the builtins return a correct "infinite precision" result even in case of overflow, whereas the manual functions can fall back on UB for overflow.)

On balance I'd rather keep the checks simple and fallback to the manual implementations rather than keep adding per-compiler tweaks here. Unless there's a case where it really does balloon out the code size.

projectgus: @yoctopuce I'm sorry, I missed your reply from 3 days ago - GitHub hadn't loaded it into the tab I still had open so I just saw it now.

 That's interesting you see such a big difference by comparison to the things I checked against. Can put in the suggested extra check for built-ins, then.

ports/unix/variants/longlong/mpconfigvariant.h
projectgus: `MICROPY_OBJ_REPR_C` seems to need more work than just enabling it for this config, so I'll look at that for a follow-up PR.

py/objint_longlong.c
projectgus: @yoctopuce Have you had any time to try out these changes on the platforms you're using?

yoctopuce: In case it helps (I realize my comments about `parsed_uint_t` might not be that clear), you can compare with my versions havailable here: 
https://github.com/yoctopuce/micropython/blob/yocto_master/py/misc.h
https://github.com/yoctopuce/micropython/blob/yocto_master/py/parsenum.c

and for my comment regarding support for `MICROPY_PY_BUILTINS_FLOAT`:
https://github.com/yoctopuce/micropython/blob/yocto_master/py/objint_longlong.c

projectgus:
I've pushed new commits including changing `mp_parse_num()` to unsigned as suggested by @yoctopuce above. I've gone further and changed the default "small int" parsing to unsigned as well (using `uintptr_t`) and this seems to have a code size benefit on all ports.

The "adding float to 64-bit long int" case I haven't dealt with here as this PR is too big already, I'll push it in a follow-up (unless you're keen to submit the fix for that, @yoctopuce?)

More info (tests run, etc.) is in the updated description for this PR.

projectgus:
Looks like UBSan is proving its worth already... :sweat_smile:

dpgeorge:
> Looks like UBSan is proving its worth already

Yay!  Will be interesting to see how many of the warnings are actual errors.

yoctopuce:
> The "adding float to 64-bit long int" case I haven't dealt with here as this PR is too big already, I'll push it in a follow-up (unless you're keen to submit the fix for that, @yoctopuce?)

Thank you for asking. Feel free to do as you prefer. Let's say that if I don't see the new PR coming once this one gets integrated into the master, I will submit it :-)

projectgus:
>    bare-arm:    +8 +0.014% 
> minimal x86:   +29 +0.015% 
>  unix x64:    +0 +0.000% standard
>      stm32:   +12 +0.003% PYBV10
>     mimxrt:    +8 +0.002% TEENSY40
>        rp2:   +72 +0.008% RPI_PICO_W
>       samd:    +8 +0.003% ADAFRUIT_ITSYBITSY_M4_EXPRESS
>  qemu rv32:    -5 -0.001% VIRT_RV32

Drat, when I tested the code size improvement from unsigned `mp_parse_num()` with my compiler it was a benefit on all the ports I tested. But here it's mostly neutral, and a big size hit on rp2! :disappointed:

dpgeorge:
> Drat, when I tested the code size improvement from unsigned `mp_parse_num()` with my compiler it was a benefit on all the ports I tested. But here it's mostly neutral, and a big size hit on rp2! 😞

The issue is because `mp_small_int_mul_overflow()` now takes a third argument which is the address of a local variable, and that means the variable must now be on the stack, etc.

I've pushed a commit to try and improve the situation there, by adjusting which local has it's address taken.  That improves things for most ports, but others (unix, esp8266) increase a little with my change.  Still, I think that's a reasonable improvement.

That change to `mp_small_int_mul_overflow()` to add a third arg is actually not really needed, because that function is only used in `py/runtime.c` in the end.  So it could be reverted.  I tried that but it did not decrease code size compared to my first change mentioned above.

projectgus:
Alright, a bit of finagling and we've got a very minor code size reduction on all the bare metal ports! :tada: 

I am not sure which part is causing the minor increase in unix port sizes, but it's small enough to possibly be noise.

I've rebased and simplified the commit history (removed the parsing fixes and improvements in `objint_longlong.c` that were later removed entirely). Have re-run tests locally and updated the PR description with the details.

dpgeorge:
@projectgus I'm happy to squash my optimisation commit down into your commit (the one just prior to mine).  I don't need the attribution for that.

projectgus:
@dpgeorge OK, have squashed! (And rebased agian also.)

dpgeorge (approved):
Looking good!

dpgeorge:
Thanks @projectgus and @yoctopuce for your efforts here.  Lots of small details have been addressed which is great!

yoctopuce:
Thanks @projectgus, you deserve all the credit. This is a really nice improvement for bare metal ports with limited ressources.

PR #16956, opened 2025-03-18
mimxrt/machine_i2c.c: Make I2C bus ID arg optional with default.

malcolm-sparkfun:
### Summary

This PR adds for the mimxrt port the same functionality as the already approved [https://github.com/micropython/micropython/pull/16671](https://github.com/micropython/micropython/pull/16671). It gives the option to not  pass an I2C Bus ID when creating a machine I2C object on the mimxrt port. This would allow users on mimxrt platforms to simply declare an I2C object with machine.I2C() without passing any arguments, thus creating an object with a default I2C ID of 0 (freq and drive were already optional for this port).

### Testing
Tested on TEENSY4.0 and TEENSY4.1 Boards:

- Created machine.I2C object by passing no arguments to I2C(). Verified that it successfully could use the scan() method to detect a connected sensor.
- Exercised I2C object by using it with an example program performing reads and writes to a sensor.

I simply chose the default I2C ID to be 0, which may not be the desired default for some boards? As far as I could tell, this seemed like a fairly reasonable choice.

### DIFF
--- a/ports/mimxrt/machine_i2c.c
+++ b/ports/mimxrt/machine_i2c.c
@@ -89,7 +89,7 @@ static void machine_i2c_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
 mp_obj_t machine_i2c_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *all_args) {
     enum { ARG_id, ARG_freq, ARG_drive};
     static const mp_arg_t allowed_args[] = {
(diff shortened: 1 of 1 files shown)


robert-hh:
Thanks for the contribution. This change is already covered in a different PR #16876, last commit, which addresses I2C, SPI and UART.

malcolm-sparkfun:
Oh perfect, thank you. I will close this then.

dpgeorge:
It looks like both this and #16876 have the same changes, which is good, you both implicitly agreed that I2C instance 0 is a good default.

PR #16958, opened 2025-03-19
codeconventions: Document the static naming conventions.

projectgus:
### Summary

This is based on some notes I took a while back and checked with @dpgeorge. It's mostly consistent in MicroPython but it's not always obvious (at least to me) what convention to select where.

Also added some more heading structure so it's easier to deep-link.

EDIT: To clarify, this PR is not trying to define any rules but to write down what the existing ones are. This doesn't mean they can't be changed in the future, or that there's an expectation that all the existing code uses the same conventions at the moment.

### DIFF
--- a/CODECONVENTIONS.md
+++ b/CODECONVENTIONS.md
@@ -206,14 +206,21 @@ adhere to the existing style and use `tools/codeformat.py` to check any changes.
 The main conventions, and things not enforceable via the auto-formatter, are
 described below.
 
-White space:
+As the MicroPython code base is over ten years old, not every source file
+conforms fully to these conventions. If making small changes to existing code,
+then it's usually acceptable to follow the existing code's style. New code or
+major changes should follow the conventions described here.
+
+## White space
+
 - Expand tabs to 4 spaces.
 - Don't leave trailing whitespace at the end of a line.
 - For control blocks (if, for, while), put 1 space between the
   keyword and the opening parenthesis.
 - Put 1 space after a comma, and 1 space around operators.
 
-Braces:
+## Braces
+
 - Use braces for all blocks, even no-line and single-line pieces of
   code.
 - Put opening braces on the end of the line it belongs to, not on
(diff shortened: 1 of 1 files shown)


CODECONVENTIONS.md
```diff
@@ -229,6 +229,25 @@ Names:
...
+  the file name (i.e. items defined in `py/obj.c` are declared in `py/obj.h` and
+  should be prefixed `mp_obj_`). There are exceptions, for example where one
+  header file contains declarations implemented in multiple source files for
+  expediency.
+
+Static names:
+- For static functions and variables exposed to Python (i.e. a static C function
+  that is wrapped in `MP_DEFINE_CONST_FUN_...` and attached to a module), use
+  the file-level shared common prefix, i.e. name them as if the function or
+  variable was not static.
```
dpgeorge, line 243: I'm not sure about this point.  There are quite a few examples that don't follow this, or follow a variation of it.  (Or maybe I don't understand exactly what you wrote?)

Eg the `marshal` module has static `marshal_loads()` and `marshal_dumps()` that are exposed to Python.  But if they were non-static I think they'd be called `mp_marshal_loads()` etc.  Similarly the `time` module has static functions like `time_time()` and `time_time_ns()`.

A different example: `openamp.Endpoint` has static functions like `endpoint_send()` which could probably be `openamp_endpoint_send()`.

Maybe all that really needs changing in this paragraph is to say that such static functions don't have the `mp_` prefix?

dlech:
IMHO, static identifier names should use the same convention as non-static (use `mp_` prefix). It is one less rule to have to keep in your head when writing code and it is less likely to result in conflicts with identifiers included from 3rd party libraries, which has happened on several occasions.

CODECONVENTIONS.md
projectgus: Thanks for taking a look, Damien. This PR is me trying to document the most commonly used existing practices as a baseline, so I'm definitely happy to change it to what you think is most accurate here.

Related question: Should all non-static identifiers have an `mp_` prefix, as a general rule?

dpgeorge: Unfortunately I don't think existing code is consistent enough to have a general rule (for static functions).  There will be lots of exceptions to whatever rule is written down here.

Considering @dlech point below that everything should be prefixed with `mp_` (which I tend to agree with, makes it simpler), maybe we can just keep the text you've written.

> Related question: Should all non-static identifiers have an `mp_` prefix, as a general rule?

Yes, I think so.  At least in `py/` and `extmod/`.  For ports... they are less consistent.

projectgus: I've updated based on these comments and our discussion just now.

projectgus: Made another small update

dpgeorge (approved):

PR #16960, opened 2025-03-19
py/py.mk: Enable makefile for USER_C_MODULES path.

Carglglz:
This enables setting a file path to select user C modules in a `USER_C_MODULES` directory the same way CMake works. 

The file is a .mk file with the user modules required.

`USER_C_MODULES`=<path/to/file>/my_user_modules.mk

in my_user_modules.mk:
```
MODULES_PATH := ../../examples/usercmodule
MODULES := cexample
MODULES += cppexample
MODULES += subpackage

$(info USER_C_MODULES found in $(MODULES_PATH): $(MODULES))
$(foreach module, $(MODULES), \
    $(eval USERMOD_DIR = $(patsubst %/,%,$(MODULES_PATH))/$(module))\
    $(info Including User C Module from $(USERMOD_DIR))\
	$(eval include $(USERMOD_DIR)/micropython.mk)\
)

```
Note that this doesn't change current behaviour i.e. providing a directory path works as expected.

### DIFF
--- a/examples/usercmodule/user_c_modules.mk
+++ b/examples/usercmodule/user_c_modules.mk
@@ -1,0 +1,10 @@
+MODULES := cexample
+MODULES += cppexample
+MODULES += subpackage
+
+$(foreach module, $(MODULES), \
+    $(eval USERMOD_DIR = $(patsubst %/,%,$(USER_C_MODULES_PATH))/$(module))\
+    $(info Including User C Module from $(USERMOD_DIR))\
+	$(eval include $(USERMOD_DIR)/micropython.mk)\
+)
+
--- a/py/py.mk
+++ b/py/py.mk
@@ -29,13 +29,6 @@ LD += -m32
 endif
 
 # External modules written in C.
-ifneq ($(USER_C_MODULES),)
-# pre-define USERMOD variables as expanded so that variables are immediate
-# expanded as they're added to them
-
-# Confirm the provided path exists, show abspath if not to make it clearer to fix.
-$(if $(wildcard $(USER_C_MODULES)/.),,$(error USER_C_MODULES doesn't exist: $(abspath $(USER_C_MODULES))))
-
 # C/C++ files that are included in the QSTR/module build
 SRC_USERMOD_C :=
 SRC_USERMOD_CXX :=
@@ -52,11 +45,33 @@ LDFLAGS_USERMOD :=
 # added to SRC_USERMOD_C below
 SRC_USERMOD :=
 
+ifneq ($(USER_C_MODULES),)
+# pre-define USERMOD variables as expanded so that variables are immediate
+# expanded as they're added to them
+
+# Confirm the provided path exists, show abspath if not to make it clearer to fix.
+ifeq ($(wildcard $(USER_C_MODULES)/.),)
+
+# Check if the provided path is a file
+ifeq ($(wildcard $(USER_C_MODULES)),)
+$(error USER_C_MODULES doesn't exist: $(abspath $(USER_C_MODULES)))
+
+# USER_C_MODULES is a .mk file
+else
+USER_C_MODULES_PATH := $(dir $(USER_C_MODULES))
+$(eval include $(USER_C_MODULES))
+
+override USER_C_MODULES := $(patsubst %/,%,$(USER_C_MODULES_PATH))
+
+endif
+
+else
 $(foreach module, $(wildcard $(USER_C_MODULES)/*/micropython.mk), \
     $(eval USERMOD_DIR = $(patsubst %/,%,$(dir $(module))))\
     $(info Including User C Module from $(USERMOD_DIR))\
 	$(eval include $(module))\
 )
+endif
 
 SRC_USERMOD_C += $(SRC_USERMOD)
 
--- a/docs/develop/cmodules.rst
+++ b/docs/develop/cmodules.rst
@@ -127,8 +127,9 @@ To build such a module, compile MicroPython (see `getting started
 applying 2 modifications:
 
 1. Set the build-time flag ``USER_C_MODULES`` to point to the modules
-   you want to include.  For ports that use Make this variable should be a
-   directory which is searched automatically for modules.  For ports that
+   you want to include.  For ports that use Make this variable can be a
+   directory which is searched automatically for modules or a Makefile that
+   includes the modules to build.  For ports that
    use CMake this variable should be a file which includes the modules to
    build.  See below for details.
 
@@ -146,6 +147,13 @@ For example, here's how the to build the unix port with the example modules:
 
     cd micropython/ports/unix
     make USER_C_MODULES=../../examples/usercmodule
+
+or 
+
+.. code-block:: bash
+
+    cd micropython/ports/unix
+    make USER_C_MODULES=../../examples/usercmodule/user_c_modules.mk
 
 You may need to run ``make clean`` once at the start when including new
 user modules in the build.  The build output will show the modules found::


stinos:
> The file is a simple text file with the user modules required, e.g.

Can multiple text files be supplied?

> Note that this doesn't change current behaviour.

I think it does: originally there'de be an error message if passing a non-existing path?

Anyway, isn't it somewhat confusing that USER_C_MODULES can either be the modules or a .txt, and that implementing it requires 2 different code paths? How about allowing both? E.g. if there would be a USER_C_MODULES_DEF or something like that, to point to a text file, you could in the .mk probably build the USER_C_MODULES variable by appending each value from USER_C_MODULES_DEF to it and leave the rest of the code the same instead of having to loop over MODULE_PATHS in a slightly different way.

andrewleech:
Taking a step back, I gather the goal here is to provide a mechanism for specifying / listing multiple C modules to include in a build that don't necessarily need to be in flat folders side by side like is currently required. 

I've wanted something like this myself, particularly to support different folder depths / locations so some can be in arbitrarily structured submodules. 

For some reason specifying just a list of paths in a text file like this "feels" wrong to me, though I'm not sure why, I'm possibly being quite unfair. 

Basically what I want is the manifest.py system but for the C modules, which leads me to think perhaps it'd be cleaner to extend the existing manifest.py handlers to also support C modules?

The biggest benefit of the manifest system I can think of is that paths are always relative to the manifest file itself, or you have the variables you can use to start from the micropython root dir etc; having something like this would remove a lot of the confusion / discrepancy that currently exists between ports and the "current directory" that relative paths start from.

Gadgetoid:
I list multiple C modules frequently across a dozen or more builds just by including them into a CMake file... this feature is already implicitly supported since there's no restriction on what you can put into a `micropython.cmake` or `usercmodules.cmake` or `my_custom_modules.cmake` or whatever you want to name it.

Eg:

https://github.com/pimoroni/pimoroni-pico/blob/main/micropython/modules/micropython-pico.cmake

You can mix and match modules to your heart's content, conditionally enable modules, conditionally configure modules and pretty much anything CMake would ordinarily allow you to do.

Want to include modules based on a variable you've set in a board level configuration file? Sure!

https://github.com/pimoroni/unicorn/blob/54c20aa9d71e4c678f96db1d17ab046a6215edb7/boards/usermod-common.cmake#L38-L39

Want to conditionally include modules based on build target? Go for it!

https://github.com/pimoroni/unicorn/blob/54c20aa9d71e4c678f96db1d17ab046a6215edb7/boards/usermod-common.cmake#L45-L55

Want to enable modules based on the selected board? Go nuts!

https://github.com/pimoroni/pga/blob/98b58dea54636658968ec0f2a9bc67511e3d1527/modules/default.cmake#L21-L23

Reinventing the wheel here is wasted effort, but we should probably document (or provide an example of) this mechanism for supplying multiple modules.

Carglglz:
Based on the feedback I've decided to update this and now the file is a makefile so one can add its own logic, the only requirement is to set  `USER_C_MODULES` to the actual path and set `MODULE_PATHS`. 

> The file is a simple text file with the user modules required, e.g.
Can multiple text files be supplied?

>Note that this doesn't change current behaviour.
I think it does: originally there'de be an error message if passing a non-existing path?

@stinos I guess now you could include multiple makefiles recursively. And no, this still checks first if the path is a directory, if it isn't then checks if the path exists, assuming it is a makefile.

> For some reason specifying just a list of paths in a text file like this "feels" wrong to me, though I'm not sure why, I'm possibly being quite unfair.

@andrewleech I agree with this and I've updated the PR accordingly 👍🏼 

> Basically what I want is the manifest.py system but for the C modules, which leads me to think perhaps it'd be cleaner to extend the existing manifest.py handlers to also support C modules?

> The biggest benefit of the manifest system I can think of is that paths are always relative to the manifest file itself, or you have the variables you can use to start from the micropython root dir etc; having something like this would remove a lot of the confusion / discrepancy that currently exists between ports and the "current directory" that relative paths start from.

That would be nice but it would probably take considerable more effort than this tiny change.

@Gadgetoid I think that now with the makefile you could add the same logic right?

Gadgetoid:
@Carglglz this feels like a much more logical approach, thank you!

py/py.mk
```diff
@@ -34,8 +34,17 @@ ifneq ($(USER_C_MODULES),)
...
 # Confirm the provided path exists, show abspath if not to make it clearer to fix.
-$(if $(wildcard $(USER_C_MODULES)/.),,$(error USER_C_MODULES doesn't exist: $(abspath $(USER_C_MODULES))))
+ifeq ($(wildcard $(USER_C_MODULES)/.),)
 
+# Check if the provided path is a file
+ifeq ($(wildcard $(USER_C_MODULES)),)
+$(error USER_C_MODULES doesn't exist: $(abspath $(USER_C_MODULES)))
+else
+# USER_C_MODULES is a .mk file
+$(eval include $(USER_C_MODULES))
```
dpgeorge: Instead of this user .mk file setting `MODULE_PATHS`, wouldn't it be simpler if that .mk file just did what it needed to do, eg included other .mk files?  Then it would work almost the same as the cmake version.

In that case the user .mk file would need to be included below, after setting the SRC_USERMOD variables to empty.

dpgeorge:
@Carglglz thanks for updating.  Did you get a chance to test this new version?

Carglglz:
@dpgeorge yes, I've tested 

```bash
 $ make USER_C_MODULES=../../examples/usercmodule 
``` 
and
```bash
 $ make USER_C_MODULES=../../examples/usercmodule/user_c_modules.mk 
```
they both work, I've also updated the docs (feel free to change anything) and let me know if I need to add this method to unix port coverage somehow? 👀

examples/usercmodule/user_c_modules.mk
```diff
@@ -0,0 +1,12 @@
+MODULES_PATH := ../../examples/usercmodule
```
dpgeorge: Does this `../../` bit of the path need to be hard coded here?  Doesn't it mean that this example must be `../../` relative to where the port is?

If so, that won't work for general C libraries, eg ulab.  There needs to be a variable that's available in this file which is the current path, like `$(USERMOD_DIR)`.

examples/usercmodule/user_c_modules.mk
```diff
@@ -0,0 +1,12 @@
...
+MODULES := cexample
+MODULES += cppexample
+MODULES += subpackage
+
+$(info USER_C_MODULES found in $(MODULES_PATH): $(MODULES))
+$(foreach module, $(MODULES), \
+    $(eval USERMOD_DIR = $(patsubst %/,%,$(MODULES_PATH))/$(module))\
+    $(info Including User C Module from $(USERMOD_DIR))\
+	$(eval include $(USERMOD_DIR)/micropython.mk)\
+)
```
dpgeorge, line 9: For this example, I'd suggest keeping it as minimal as possible, and similar to the cmake version.  Eg just:
```
include cexample/micropython.mk
include cppexample/micropython.mk
include subpackage/micropython.mk
```

Carglglz: I agree with making it similar to the cmake version however then both versions will differ, i.e. this one won't show any information of which modules are included, and also `$(USERMOD_DIR)` needs to be set for each module. 

So you will end with something like this:
```
$(eval USERMOD_DIR = $(patsubst %/,%,$(USER_C_MODULES_PATH))/cexample)\
$(info Including User C Module from $(USERMOD_DIR))\
$(eval include $(USERMOD_DIR)/micropython.mk)\

$(eval USERMOD_DIR = $(patsubst %/,%,$(USER_C_MODULES_PATH))/cppexample)\
$(info Including User C Module from $(USERMOD_DIR))\
$(eval include $(USERMOD_DIR)/micropython.mk)\

$(eval USERMOD_DIR = $(patsubst %/,%,$(USER_C_MODULES_PATH))/subpackage)\
$(info Including User C Module from $(USERMOD_DIR))\
$(eval include $(USERMOD_DIR)/micropython.mk)\

```
So I think the single foreach loop is better, but let me know if you prefer this version 👍🏼

examples/usercmodule/user_c_modules.mk
Carglglz: > Does this ../../ bit of the path need to be hard coded here? Doesn't it mean that this example must be ../../ relative to where the port is?

No, `$(USER_C_MODULES_PATH)` can be used instead which is 
```
USER_C_MODULES_PATH := $(dir $(USER_C_MODULES))
``` 
at this point

andrewleech:
Hi @Carglglz before I realised how much you'd improved this PR I put together a different approach that lets the user specify arbitrary user c module in the manifest.py instead of the USER_C_MODULES make arg. I think this consolidates the configuration more for both mand and cmake and allows python to handle all the inconsistencies with path handling (abs vs rel etc).
https://github.com/micropython/micropython/pull/18229

Arguably both it and this one have merit and could both be included if desired, I'm pretty sure USER_C_MODULES still works correctly in my PR (the existing / original way at least)

I'd be interested in your thoughts on it!

dpgeorge:
Closing in favour of #18229, which I think will eventually replace USER_C_MODULES altogether (although right now you can still use both).

PR #16961, opened 2025-03-19
stm32: invalidate I and D caches after modifying QSPI flash

dpgeorge:
### Summary

stm32's QSPI driver supports memory-mapped mode.  The memory-mapped flash can also be erased/written to.  To support both these modes, it switches in and out of memory-mapped mode during an erase/write.

If you erase/write the flash and then go back to memory mapped mode, you must invalidate the cache related to the memory-mapped region that changed.  Otherwise you may end up reading old data.

That cache invalidation is currently not being done, and this PR fixes that.

This bug has been around ever since QSPI memory-mapped mode existed, but it's never really been observed because it's not common to use flash in memory-mapped mode and also erase/write it.  Eg PYBD_SF2 uses the memory-mapped flash in read-only mode to store additional firmware.

But since the introduction of ROMFS, things changed.  The `vfs.rom_ioctl()` command can erase/write memory-mapped flash.

### Testing

Running the following script on PYBD_SF2 and PYBD_SF6 shows the issue:
```py
import time, machine, vfs, uctypes

addr = uctypes.addressof(vfs.rom_ioctl(2, 0))
print(hex(addr))

for _ in range(4):
    print("---------")
    t0 = time.ticks_us()
    vfs.rom_ioctl(3, 0, 1)
    t1 = time.ticks_us()
    print('erase time', time.ticks_diff(t1, t0))
    print(machine.mem32[addr])

    t0 = time.ticks_us()
    vfs.rom_ioctl(4, 0, 0, b'1234')
    t1 = time.ticks_us()
    print('write time', time.ticks_diff(t1, t0))
    print(machine.mem32[addr])

    t0 = time.ticks_us()
    vfs.rom_ioctl(3, 0, 1)
    t1 = time.ticks_us()
    print('erase time', time.ticks_diff(t1, t0))
    print(machine.mem32[addr])
```
Without this fix the memory retrieved by `mem32` is usually always `-1` ie `0xffffffff`, even after writing `b'1234'` to it.

With the fix to invalidate the cache, he above code works as expected.

I originally implemented this in the existing `MP_QSPI_IOCTL_BUS_RELEASE` ioctl, invalidating all of the QSPI memory-mapped region.  But that takes ~20ms, far too long for small writes (which can take <100us).  So this fix only invalidates the regions as they change.  That is a very small overhead (unmeasurable compared to the cost of the erase/write).

### DIFF
--- a/drivers/bus/qspi.h
+++ b/drivers/bus/qspi.h
@@ -37,10 +37,11 @@ enum {
     MP_QSPI_IOCTL_DEINIT,
     MP_QSPI_IOCTL_BUS_ACQUIRE,
     MP_QSPI_IOCTL_BUS_RELEASE,
+    MP_QSPI_IOCTL_MEMORY_MODIFIED,
 };
 
 typedef struct _mp_qspi_proto_t {
-    int (*ioctl)(void *self, uint32_t cmd);
+    int (*ioctl)(void *self, uint32_t cmd, uintptr_t arg);
     int (*write_cmd_data)(void *self, uint8_t cmd, size_t len, uint32_t data);
     int (*write_cmd_addr_data)(void *self, uint8_t cmd, uint32_t addr, size_t len, const uint8_t *src);
     int (*read_cmd)(void *self, uint8_t cmd, size_t len, uint32_t *dest);
--- a/drivers/bus/softqspi.c
+++ b/drivers/bus/softqspi.c
@@ -56,8 +56,9 @@ static void nibble_write(mp_soft_qspi_obj_t *self, uint8_t v) {
     mp_hal_pin_write(self->io3, (v >> 3) & 1);
 }
 
-static int mp_soft_qspi_ioctl(void *self_in, uint32_t cmd) {
+static int mp_soft_qspi_ioctl(void *self_in, uint32_t cmd, uintptr_t arg) {
     mp_soft_qspi_obj_t *self = (mp_soft_qspi_obj_t*)self_in;
+    (void)arg;
 
     switch (cmd) {
         case MP_QSPI_IOCTL_INIT:
--- a/drivers/memory/spiflash.c
+++ b/drivers/memory/spiflash.c
@@ -61,14 +61,22 @@
 static void mp_spiflash_acquire_bus(mp_spiflash_t *self) {
     const mp_spiflash_config_t *c = self->config;
     if (c->bus_kind == MP_SPIFLASH_BUS_QSPI) {
-        c->bus.u_qspi.proto->ioctl(c->bus.u_qspi.data, MP_QSPI_IOCTL_BUS_ACQUIRE);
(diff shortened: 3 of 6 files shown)


projectgus (approved):
LGTM, I appreciate how the CMSIS implementations of the invalidate functions manage alignment for the caller!

pbleyer:
Please note that the current master branch with this merged commit is broken for STM32 with Cortex-Ms without an I/D cache (or where the caches are inaccessible through the SCB). In particular I was trying to compile for Cortex-M4 and the compiler complained for undefined SCB_InvalidateICache_by_Addr and SCB_InvalidateDCache_by_Addr functions.

dpgeorge:
@pbleyer the caching calls should only be seen by the compiler if `MICROPY_HW_QSPIFLASH_SIZE_BITS_LOG2` is enabled, which means that QSPI is used.  I'm not aware of an STM32 part that has QSPI yet does not have cache.  What part are you using and what is your QSPI configuration?

pbleyer:
I am in particular using a board with an STM32G4 that has an external QSPI Flash and therefore has the QSPI macros defined. However any STM32 that uses an M4 core should be impacted: The CMSIS core_m4.h header does not include the SCB cachel1_armv7.h file that is the one that defines the SCB_InvalidateXCache_by_Addr functions.

dpgeorge:
> I am in particular using a board with an STM32G4 that has an external QSPI Flash and therefore has the QSPI macros defined.

Ah, I didn't realise the G4 had QSPI and that the existing `ports/stm32/qspi.c` driver worked for this chip (AFAIK it's only been tested on F7 and H7).  That's good to know it works on G4!

> However any STM32 that uses an M4 core should be impacted

Only M4 MCUs with QSPI will be impacted, which looks like F4, G4, L4 and WB.

Since you are defining your own board, you should be able to work around this problem (until it's fixed) by defining the following in your board's `mpconfigboard.h` file:
```c
#define SCB_InvalidateICache_by_Addr(addr, len)
#define SCB_InvalidateDCache_by_Addr(addr, len)
```

pbleyer:
> 
> Ah, I didn't realise the G4 had QSPI and that the existing `ports/stm32/qspi.c` driver worked for this chip (AFAIK it's only been tested on F7 and H7). That's good to know it works on G4!

 Yep. ;-) I think the STM32G4 define was already added to the qspi driver by another commit. I had to apply a patch before to enable it.

> 
> > However any STM32 that uses an M4 core should be impacted
> 
> Only M4 MCUs with QSPI will be impacted, which looks like F4, G4, L4 and WB.

 Correct.

> 
> Since you are defining your own board, you should be able to work around this problem (until it's fixed) by defining the following in your board's `mpconfigboard.h` file:
> 
> ```c
> #define SCB_InvalidateICache_by_Addr(addr, len)
> #define SCB_InvalidateDCache_by_Addr(addr, len)
> ```

Indeed I did something similar.

Thanks for your help, Damien.

PR #16967, opened 2025-03-20
dynruntime.mk: fix musl's libm.a when LINK_RUNTIME=1

dubiousjim:
Like PICOLIBC, MUSL also has its math functions in libc.a. There is a libm.a, but it's empty.

### Summary

The commits in 9bbd40e had special workarounds for pico libc, because it doesn't have its 
math functions in libm.a but instead in libc.a. Some of these workarounds are also required for musl.

### Testing

Tested on a Raspberry Pi Zero 2W, with Alpine Linux. With this change, and setting LINK_RUNTIME = 1, 
enables building example/natmod/features2.

(I have to tweak other parts of dynruntime.mk to build on my system, for example to set the CROSS 
prefix correctly. But this is a variant of the armv7emdp ARCH.

An alternative would be to do no special handling for musl. Then devs building for such systems can 
still explicitly specify `MPY_LD_FLAGS += -l /usr/lib/libc.a` in their Makefiles.

### DIFF
--- a/py/dynruntime.mk
+++ b/py/dynruntime.mk
@@ -124,6 +124,10 @@ else
 $(error architecture '$(ARCH)' not supported)
 endif
 
+ifneq ($(findstring -musl,$(shell $(CROSS)gcc -dumpmachine)),)
+USE_MUSL := 1
+endif
+
 MICROPY_FLOAT_IMPL_UPPER = $(shell echo $(MICROPY_FLOAT_IMPL) | tr '[:lower:]' '[:upper:]')
 CFLAGS += $(CFLAGS_ARCH) -DMICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_$(MICROPY_FLOAT_IMPL_UPPER)
 
@@ -146,6 +150,8 @@ ifeq ($(LINK_RUNTIME),1)
 # version (that hopefully contains a newer RISC-V compiler) or to another Linux
 # distribution.
 ifeq ($(USE_PICOLIBC),1)
+LIBM_NAME := libc.a
+else ifeq ($(USE_MUSL),1)
 LIBM_NAME := libc.a
 else
 LIBM_NAME := libm.a


py/dynruntime.mk
```diff
@@ -124,6 +124,10 @@ else
 $(error architecture '$(ARCH)' not supported)
 endif
 
+ifneq ($(shell ldd --version 2>&1 | head -n 1 | grep -i musl),)
```
agatti, line 127: Wouldn't something like `$(lastword $(subst -, ,$(shell $(CROSS)gcc -dumpversion))` (not tested but it should work) be more accurate rather than looking at the default dynamic linker?

With that you'd be able to still use a musl (cross) toolchain on a system that defaults on glibc and vice versa.

dubiousjim: Doesn't work for me.

```
~ armv7-alpine-linux-musleabihf-gcc -dumpversion
14.2.0
```

agatti: `dumpversion` was a typo, I've amended the message with `-dumpmachine` a minute after, my bad.

```bash
~ # uname -a
Linux alpine 6.13.7-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Thu, 13 Mar 2025 18:11:42 +0000 x86_64 Linux
~ # gcc -dumpmachine
x86_64-alpine-linux-musl
```

dubiousjim: Cool, didn't know about that. I'll have to come back to this PR later, but this seems like a good improvement.

dubiousjim: I used `ifneq ($(findstring -musl,$(shell $(CROSS)gcc -dumpmachine)),)`. On my machine, the gcc output doesn't end with just "-musl", but with "-musleabihf".

agatti: I see - I'd have probably written something like $filter(musl%, $lastword($(subst -, ,$(shell ...)))) to only check whether the last element of the machine triplet starts with "musl", but that's just me being pedantic :)

After all, unless you re-implement bits of cmake/autoconf in that makefile there's no proper way to find out if you're using a musl toolchain anyway, is there?

dubiousjim:
I agree that's better, and agree also that any reasonable implementation here is going to be hacky and incomplete. So not clear what degree of betterness we should aim at.

agatti:
I'd say this is good enough for the majority of cases.  The other two major platforms supported by Alpine are Arm64 and RISCV64 if I recall - for the latter I believe the last triplet element is `-musl`, for the former it may very well be the same or start with `-musl` anyway.  Your code is able to handle all those situations :)

dpgeorge (approved):

Issue #16969, opened 2025-03-20
modulo operator on strings does not support namedtuple


yoctopuce:
### Port, board and/or hardware

any build with MICROPY_PY_BUILTINS_STR_OP_MODULO, including unix port

### MicroPython version

MicroPython v1.25.0-preview.393.gf1018ee5c.dirty on 2025-03-20; linux [GCC 13.3.0] version

### Reproduction

```
from collections import namedtuple
T = namedtuple("Tup", ["foo", "bar"])
t = T(1, 2)
"(%d, %d)" % t
```

### Expected behaviour

Last expression is expected to return `'(1, 2)'` as in CPython.
NamedTuples should be handled the same as tuples.

### Observed behaviour

An exception is raised:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can't convert Tup to int
```

### Additional Information

There is already a *fixme* in the code to deal with this issue.
I will file a PR with the fix shortly.

### Code of Conduct

Yes, I agree

dpgeorge:
Fixed by 0d2c18c299936a34cdeb43e2ef9cbdaecb35c610

PR #16970, opened 2025-03-20
py/objstr: Fix handling of OP_MODULO with namedtuple.

yoctopuce:
### Summary

This pull request fixes issue https://github.com/micropython/micropython/issues/16969

### Testing

The change is quite trivial, and I have verified that the modulo operator on strings now works with both tuple and namedtuple.
I have added the corresponding test case in the same commit.

namedtuples are a memory efficient alternative to objects as containers: they use no more memory than a tuple, but provide similar code readability as objects. It is therefore important to Micropython that they work as expected. 

The correct type-checking code is only slightly larger than the original version that only handled tuples, and it fixes a problem encountered when porting a library to Micropython.

### DIFF
--- a/py/objstr.c
+++ b/py/objstr.c
@@ -33,6 +33,7 @@
 #include "py/objlist.h"
 #include "py/runtime.h"
 #include "py/cstack.h"
+#include "py/objtuple.h"
 
 #if MICROPY_PY_BUILTINS_STR_OP_MODULO
 static mp_obj_t str_modulo_format(mp_obj_t pattern, size_t n_args, const mp_obj_t *args, mp_obj_t dict);
@@ -357,8 +358,7 @@ mp_obj_t mp_obj_str_binary_op(mp_binary_op_t op, mp_obj_t lhs_in, mp_obj_t rhs_i
         mp_obj_t *args = &rhs_in;
         size_t n_args = 1;
         mp_obj_t dict = MP_OBJ_NULL;
-        if (mp_obj_is_type(rhs_in, &mp_type_tuple)) {
-            // TODO: Support tuple subclasses?
+        if (mp_obj_is_tuple_compatible(rhs_in)) {
             mp_obj_tuple_get(rhs_in, &n_args, &args);
         } else if (mp_obj_is_type(rhs_in, &mp_type_dict)) {
             dict = rhs_in;
--- a/tests/basics/namedtuple1.py
+++ b/tests/basics/namedtuple1.py
@@ -21,6 +21,9 @@ for t in T(1, 2), T(bar=1, foo=2):
 
     print(isinstance(t, tuple))
 
+    # a NamedTuple can be used as a tuple
+    print("(%d, %d)" % t)
+
     # Check tuple can compare equal to namedtuple with same elements
     print(t == (t[0], t[1]), (t[0], t[1]) == t)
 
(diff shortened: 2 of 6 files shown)


py/objstr.c
```diff
@@ -357,8 +357,7 @@ mp_obj_t mp_obj_str_binary_op(mp_binary_op_t op, mp_obj_t lhs_in, mp_obj_t rhs_i
         mp_obj_t *args = &rhs_in;
         size_t n_args = 1;
         mp_obj_t dict = MP_OBJ_NULL;
-        if (mp_obj_is_type(rhs_in, &mp_type_tuple)) {
-            // TODO: Support tuple subclasses?
+        if (mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(mp_obj_get_type(rhs_in)), MP_OBJ_FROM_PTR(&mp_type_tuple))) {
```
AJMansfield, line 360: This should use `mp_obj_is_tuple_compatible` so that it's also compatible with anonymous `attrtuple` objects.

tests/basics/namedtuple1.py
AJMansfield, line 1: It looks like the only extant attrtuples available to test against though are `sys.implementation` (per tests/basics/attrtuple1.py) and `os.uname()`.

AJMansfield (changes requested):
This looks useful, but as written it still wouldn't work with attrtuples -- attrtuple has the same _name_ as the regular tuple type, but the types aren't actually the same or even a subclass.

tests/basics/namedtuple1.py
yoctopuce: `sys.implementation` is actually not a `tuple` in CPython, but a `namespace`. So testing against it does not look like the best way to validate compatibility between tuple and attrtuple.

So I chose to use `os.uname()` to validate the fix on `attrtuple`. In order to do so, I have added support for `os.uname()` on Unix coverage build, which is probably not a bad idea anyway.

AJMansfield (changes requested):
Could you also add a test of the attrtuple case?

e.g. something like this:

```
print("(%r, %r, %r, %r, %r)" % os.uname() == repr(tuple(os.uname())))
```

(With an appropriate separate skip check, probably in a `tests/basics/attrtuple2.py` to keep it orthogonal vs the skip check in `attrtuple1.py`.)

yoctopuce:
> Could you also add a test of the attrtuple case?
> 
> (With an appropriate separate skip check, probably in a `tests/basics/attrtuple2.py` to keep it orthogonal vs the skip check in `attrtuple1.py`.)

My bad, I had done exactly that, but forgot to add the new file to the commit...

AJMansfield (approved):
Good work on this one!

I've also gone through and tested this PR on my own system; I can confirm that the code compiles and functions as expected, and addresses the issue reported.

dpgeorge:
> `mp_obj_get_array` also uses `if (mp_obj_is_type(o, &mp_type_tuple))` to get items, but if it would use `mp_obj_is_tuple_compatible` it could be used with `namedtuple` (my usecase) etc as well. So I think it makes sense to look for other occurrences of this pattern and fix them all?

Yes, I agree, makes sense to fix other cases of this pattern.  But I think each one needs to be individually checked against CPython (or at least that it makes sense) and a test written for the change.

In the interests of making progress we can just merge this PR as-is (it has a test and is well motivated) and then can follow up later with other fixes.

dpgeorge (approved):

PR #16972, opened 2025-03-20
py/obj.h: Make function definition macros compatible with C++.

stinos:
Simply requires explicitly naming all initializers.

### Summary

Like ce491ab0d168a8278062e1fc7ebed3ca47ab89d2 I think this is the last change needed to make the public API macros C++-compatible. At least I didn't immediately see other places with unnamed initializers.

### Testing

I put this code in coveragecpp.cpp:

```
mp_obj_t a0();
MP_DEFINE_CONST_FUN_OBJ_0(a0_obj, a0);
mp_obj_t a1(mp_obj_t);
MP_DEFINE_CONST_FUN_OBJ_1(a1_obj, a1);
mp_obj_t a2(mp_obj_t, mp_obj_t);
MP_DEFINE_CONST_FUN_OBJ_2(a2_obj, a2);
mp_obj_t a3(mp_obj_t, mp_obj_t, mp_obj_t);
MP_DEFINE_CONST_FUN_OBJ_3(a3_obj, a3);
mp_obj_t avar(size_t, const mp_obj_t *);
MP_DEFINE_CONST_FUN_OBJ_VAR(avar_obj, 1, avar);
mp_obj_t avarbetween(size_t, const mp_obj_t *);
MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(avarbetween_obj, 1, 2, avarbetween);
mp_obj_t akw(size_t, const mp_obj_t *, mp_map_t *);
MP_DEFINE_CONST_FUN_OBJ_KW(akw_obj, 1, akw);
static const mp_rom_map_elem_t atable[] = {};
MP_DEFINE_CONST_MAP(amap, atable);
MP_DEFINE_CONST_DICT(adict, atable);
```

briefly discussed before in https://github.com/micropython/micropython/pull/14323#issuecomment-2071972127 the question is if this code should just be included.

### DIFF
--- a/ports/unix/coveragecpp.cpp
+++ b/ports/unix/coveragecpp.cpp
@@ -17,10 +17,61 @@ extern "C" {
 #include <py/runtime.h>
 }
 
+// Invoke all (except one, see below) public API macros which initialize structs to make sure
+// they are C++-compatible, meaning they explicitly initialize all struct members.
+mp_obj_t f0();
+MP_DEFINE_CONST_FUN_OBJ_0(f0_obj, f0);
+mp_obj_t f1(mp_obj_t);
+MP_DEFINE_CONST_FUN_OBJ_1(f1_obj, f1);
+mp_obj_t f2(mp_obj_t, mp_obj_t);
+MP_DEFINE_CONST_FUN_OBJ_2(f2_obj, f2);
+mp_obj_t f3(mp_obj_t, mp_obj_t, mp_obj_t);
+MP_DEFINE_CONST_FUN_OBJ_3(f3_obj, f3);
+mp_obj_t fvar(size_t, const mp_obj_t *);
+MP_DEFINE_CONST_FUN_OBJ_VAR(fvar_obj, 1, fvar);
+mp_obj_t fvarbetween(size_t, const mp_obj_t *);
+MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(fvarbetween_obj, 1, 2, fvarbetween);
+mp_obj_t fkw(size_t, const mp_obj_t *, mp_map_t *);
+MP_DEFINE_CONST_FUN_OBJ_KW(fkw_obj, 1, fkw);
+
+static const mp_rom_map_elem_t table[] = {
+    { MP_ROM_QSTR(MP_QSTR_f0), MP_ROM_PTR(&f0_obj) },
+};
+MP_DEFINE_CONST_MAP(map, table);
+MP_DEFINE_CONST_DICT(dict, table);
+
+static const qstr attrtuple_fields[] = {
+    MP_QSTR_f0,
+};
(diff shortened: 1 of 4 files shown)


dpgeorge:
> briefly discussed before in [#14323 (comment)](https://github.com/micropython/micropython/pull/14323#issuecomment-2071972127) the question is if this code should just be included.

Yes, I think it would be good to include the above bit of test code in `coveragecpp.cpp`.

dpgeorge (approved):

py/runtime.h
```diff
@@ -33,6 +33,7 @@
 // For use with mp_call_function_1_from_nlr_jump_callback.
 #define MP_DEFINE_NLR_JUMP_CALLBACK_FUNCTION_1(ctx, f, a) \
     nlr_jump_callback_node_call_function_1_t ctx = { \
+        .callback = {.prev = NULL, .fun = NULL}, \
```
dpgeorge, line 36: I'm not quite sure about this change... this macro is always followed by a call to `nlr_push_jump_callback()` which is responsible for initialising this sub-struct.  So it's a bit of a waste to set it to NULL here.

Maybe a better way is to change this macro to:
```py
nlr_jump_callback_node_call_function_1_t ctx;
ctx.func = f;
ctx.arg = a;
```
or something similar?

ports/unix/coveragecpp.cpp
```diff
@@ -17,10 +17,57 @@ extern "C" {
 #include <py/runtime.h>
 }
 
+// Invoke all (except one, see below) public API macros which initialize structs to make sure
```
dpgeorge, line 20: Might be worth splitting this change to `coveragecpp.cpp` out into a separate commit?

py/runtime.h
stinos: Ok.

ports/unix/coveragecpp.cpp
stinos: Yeah sure, done.

py/runtime.h
```diff
@@ -30,12 +30,12 @@
...
+// Initialize an nlr_jump_callback_node_call_function_1_t struct for use with
+// nlr_push_jump_callback(&ctx.callback, mp_call_function_1_from_nlr_jump_callback);
 #define MP_DEFINE_NLR_JUMP_CALLBACK_FUNCTION_1(ctx, f, a) \
-    nlr_jump_callback_node_call_function_1_t ctx = { \
-        .func = (void (*)(void *))(f), \
-        .arg = (a), \
-    }
+    nlr_jump_callback_node_call_function_1_t ctx; \
+    ctx.func = (void (*)(void *))(f); \
+    ctx.arg = (a);
```
dpgeorge, line 38: Very minor, but I think this line should not end with a `;`, ie should be `ctx.arg = (a)`.  That will force the user of this macro to add a semicolon after the macro.

dpgeorge (approved):

PR #16979, opened 2025-03-21
unix/main: Remove PATH_MAX from realpath.

yangfl:
POSIX.1-2008 ensures realpath() give a dynamically allocated buffer if NULL is passed, avoiding an explicit call to malloc() and use of PATH_MAX, which may be undefined on some systems.

### DIFF
--- a/ports/unix/main.c
+++ b/ports/unix/main.c
@@ -453,10 +453,13 @@ static void set_sys_argv(char *argv[], int argc, int start_arg) {
 
 #if MICROPY_PY_SYS_EXECUTABLE
 extern mp_obj_str_t mp_sys_executable_obj;
-static char executable_path[MICROPY_ALLOC_PATH_MAX];
+static char *executable_path = NULL;
 
 static void sys_set_excecutable(char *argv0) {
-    if (realpath(argv0, executable_path)) {
+    if (executable_path == NULL) {
+        executable_path = realpath(argv0, NULL);
+    }
+    if (executable_path != NULL) {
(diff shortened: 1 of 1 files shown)


stinos:
Since https://github.com/micropython/micropython/blob/f1018ee5c2dd0eabcc0cfe7e02c607594e489788/ports/windows/realpath.c#L48 is the implementation used for msvc (and which seems to do the expected thing here), it might be a good idea to mention this explicitly in the commit

ports/unix/main.c
```diff
@@ -453,10 +453,13 @@ static void set_sys_argv(char *argv[], int argc, int start_arg) {
...
 extern mp_obj_str_t mp_sys_executable_obj;
-static char executable_path[MICROPY_ALLOC_PATH_MAX];
+static char *executable_path = NULL;
 
 static void sys_set_excecutable(char *argv0) {
-    if (realpath(argv0, executable_path)) {
+    if (executable_path == NULL) {
+        executable_path = realpath(argv0, NULL);
+    }
+    if (executable_path != NULL) {
```
dpgeorge, line 462: The memory allocated for `executable_path` is never freed.  That's usually not a big problem, but in debug builds the code tries to free all memory before the program finishes, to help find memory leaks (see the end of `main_`).

So I suggest to either:
1. free `executable_path` at the end of `main_` when `!defined(NDEBUG)`; or
2. make `executable_path` a local variable in this function and free it at the end, and take a copy of the string on the MicroPython heap (`m_new(char, strlen(executable_path) + 1)` etc) to pass into `mp_obj_str_set_data`

dpgeorge (approved):

PR #16982, opened 2025-03-21
esp32/esp32_common.cmake: Use the tinyusb source files from ESP-IDF.

agatti:
### Summary

This PR removes the explicit dependency on the vendored tinyusb version for the ESP32S2 and ESP32S3 boards.

Tinyusb is still available to MicroPython through a dependency on the `espressif/esp_tinyusb` ESP-IDF component, which in turn depends on the `espressif/tinyusb` component itself.

### Testing

The `ESP32_GENERIC_S3` firmware image was built without errors and the USB example scripts in `examples/usb` were executed correctly.

It may happen that ESP-IDF's version of tinyusb lags behind the one brought in via the git submodule, and that may not be wanted in case of urgent changes in tinyusb that aren't picked up by ESP-IDF.  On the other hand this simplifies the build process when MicroPython is brought in as a custom IDF component into an existing ESP-IDF project as there's one less dependency to be manually checked out and configured, and not having the ESP-IDF build system complain after a forced `make submodules` step (which is my use case).

### DIFF
--- a/ports/esp32/esp32_common.cmake
+++ b/ports/esp32/esp32_common.cmake
@@ -80,9 +80,7 @@ list(APPEND MICROPY_SOURCE_DRIVERS
     ${MICROPY_DIR}/drivers/dht/dht.c
 )
 
-list(APPEND GIT_SUBMODULES lib/tinyusb)
 if(MICROPY_PY_TINYUSB)
-    set(TINYUSB_SRC "${MICROPY_DIR}/lib/tinyusb/src")
     string(TOUPPER OPT_MCU_${IDF_TARGET} tusb_mcu)
 
     list(APPEND MICROPY_DEF_TINYUSB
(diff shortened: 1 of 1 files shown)


andrewleech:
When you say the `examples/usb` scripts were executed correctly do you mean the runtime usb examples in https://github.com/micropython/micropython-lib/tree/master/micropython/usb ?

agatti:
No, I mean the ones that are in the `micropython` repo - https://github.com/micropython/micropython/tree/master/examples/usb, as in ran usb_dfu_device.py and performed DFU operations, plus ran usb_simple_device.py and executed usb_simple_host_pyusb.py on my machine to make sure communication happened.

I've also tested third party code as well but that's not really public :)  Still, I can give the micropython-lib examples you mention a go and report back, no problem.

agatti:
The examples in `micropython-lib/micropython/usb/examples` seem to work on an ESP32S3-DevKitC (I cannot test `keyboard_example.py` as I don't have anything to connect to the board's GPIOs right now, not even wires...), and `micropython-lib/micropython/usb/usb-device/tests/test_core_buffer.py` also passes when ran on the board via `mpremote`.

All of this was after removing `lib/tinyusb` on a fresh micropython checkout of this PR's branch.

andrewleech:
That's great thanks @agatti to be honest I never thought of directly using the copy of tinyusb in IDF when I made this integration, in my mind that was what I was trying to get away from! But yes it was really only their C integration / bindings I was replacing with the micropython ones,  the source of tinyusb itself should be fine from either location!

dpgeorge (approved):
This is a neat simplification, to reduce dependencies.

Tested on an ESP32S2 using IDF 5.4.1, with a USB CDC rate test, and tested the code in `examples/usb/` in this repo.  Everything still works.

PR #16984, opened 2025-03-21
document additional inconsistencies with CPython

k98kurz:
### Summary

In my development, I discovered four undocumented inconsistencies with CPython. MattyTrenti on Discord suggested I submit an issue or PR. The four undocumented inconsistencies are as follows:

- Unpacking lists or tuples raises SyntaxError: `[1, 2, *others]`, `[*others, 1, 2]`, `(1, 2, *others)`, `(*others, 1, 2)`
- Unpacking dicts raises SyntaxError: `{**others, 'c': 3}`, `{'c': 3, **others}`
- `time.time()` returns an int instead of a float
- `struct` does not support the "?" type code for bools

### Testing

Wrote code that worked in CPython and ran it on an ESP32, encountered the errors, and then manually tested in a REPL on ESP32.

### DIFF
--- a/tests/cpydiff/modules_struct_does_not_support_bool.py
+++ b/tests/cpydiff/modules_struct_does_not_support_bool.py
@@ -1,0 +1,11 @@
+"""
+categories: Modules,struct
+description: Struct ``pack`` and ``unpack`` do not support the "?" type code for bools.
+cause: Unknown
+workaround: Cast to int and pack with "B" type code, then unpack with "B" and cast to bool.
+"""
+
+import struct
+
+struct.pack("?", True)
+struct.unpack("?", b"\x01")
--- a/tests/cpydiff/modules_time_time_returns_int.py
+++ b/tests/cpydiff/modules_time_time_returns_int.py
@@ -1,0 +1,10 @@
+"""
+categories: Modules.time
+description: ``time`` returns an integer instead of a float.
+cause: Unknown
+workaround: use ``int(time())`` if you need a timestamp with just the seconds or ``time_ns()/1_000_000_000`` if you need a float.
+"""
+
+import time
+
+print(type(time.time()))
--- a/tests/cpydiff/syntax_dict_unpacking.py
+++ b/tests/cpydiff/syntax_dict_unpacking.py
@@ -1,0 +1,10 @@
+"""
+categories: Syntax,dict
+description: Unpacking one dict's elements into a new dict with ``{'new_elem': val, **old_dict}`` is not supported.
+cause: Not implemented. Dict unpacking is only implemented when passing dict elements as function kwargs.
+workaround: use ``new_dict = {}; new_dict.update(old_dict)`` to copy values into new dicts.
+"""
+
+old_dict = {'a': 1, 'b': 2}
+new_dict = {**old_dict, 'c': 3}
+new_dict = {'c': 3, **old_dict}
--- a/tests/cpydiff/syntax_list_tuple_unpacking.py
+++ b/tests/cpydiff/syntax_list_tuple_unpacking.py
@@ -1,0 +1,14 @@
+"""
+categories: Syntax,tuple,list
+description: Unpacking one tuple or list's elements into a new object with ``[*things]`` or ``(*things)`` is not supported.
+cause: Not implemented. Sequence unpacking is only implemented when passing arguments to functions.
+workaround: use ``new = []; for t in old: new.append(t)`` or ``new = [] + old`` to copy into new lists and then cast to tuple if a tuple is needed.
+"""
+
+old_list = [1, 2, 3]
+new_list = [*old_list, 4]
+new_list = [0, *old_list]
+
+old_tup = (1, 2, 3)
+new_tup = (*old_tup, 4)
+new_tup = (0, *old_tup)


tests/cpydiff/modules_time_time_returns_int.py
AJMansfield, line 1: There was no difference running this time type test agains tthe unix port, this seems to be specific to esp32.

tests/cpydiff/modules_struct_does_not_support_bool.py
AJMansfield, line 1: Can confirm the struct boolean cpydiff on the unix port.

tests/cpydiff/syntax_dict_unpacking.py
AJMansfield, line 1: Can confirm the dict unpacking cpydiff on the unix port.

tests/cpydiff/syntax_list_tuple_unpacking.py
AJMansfield, line 1: Can confirm tuple unpacking cpydiff on the unix port.

AJMansfield (approved):
The tuple and dict unpacking features are certainly ones I've been missing myself.

tests/cpydiff/modules_time_time_returns_int.py
k98kurz: Thanks for looking into these. Should I amend the file to include verbiage around it being a potential device/build-specific difference?

projectgus: This PR highlights a general shortcoming we have in the docs at the moment, which is that the "cpydiff" mechanisms only document differences between CPython and the unix port, and don't document differences that only exist at certain feature levels or on certain ports.

There is no automated mechanism for that in MicroPython at the moment, but in the case of module-specific changes it's usually noted in the module docs (which is arguably more discoverable for most people). This particular difference is already documented in a note on [the time module](https://docs.micropython.org/en/latest/library/time.html#time.time).

tests/cpydiff/modules_struct_does_not_support_bool.py
projectgus: Thanks for adding this, @k98kurz. FWIW if the code size impact is not too big then maybe we should support this type code, but documenting it's lack is a good start. :)

Meta-point, there is currently an inconsistency between the [struct module docs](https://docs.micropython.org/en/latest/library/struct.html#module-struct) - which note one difference to CPython - and the [code generated differences](https://docs.micropython.org/en/latest/genrst/modules.html#struct) which note four including this one. Probably similar inconsistencies exist across all of the modules in "cpydiff" and this should be made consistent somehow... I'm thinking maybe we move these generated sections into the bottom of the individual module's documentation pages, and leave the standalone "changes" page for only the language-language changes. Don't feel obliged to tackle that in this PR, unless you're particularly keen.

projectgus (commented):
Thanks for taking the time to document these differences, @k98kurz! I also learned something, I personally didn't know about that dict unpacking could be used this way. (I assumed it was a new feature, but it's been around for a long time... :facepalm:)

I made some suggestions to clarify that unpacking is still supported for function calls, and I think the `time.time()` difference documentation probably belongs in the module doc (feel free to submit a docs change there if you think the existing info is not clear enough).

tests/cpydiff/modules_struct_does_not_support_bool.py
k98kurz: @projectgus I see only three documented struct differences in the [code generated differences](https://docs.micropython.org/en/latest/genrst/modules.html#struct) link: too few args, too many args, and whitespace. But the struct module docs indeed note only the whitespace difference. I am still getting my bearings with micropython, so I am not sure what to change to bring all of them into alignment.

projectgus: > I see only three documented struct differences in the [code generated differences](https://docs.micropython.org/en/latest/genrst/modules.html#struct) link: too few args, too many args, and whitespace

Oh, sorry, I meant four including the one being added in this PR.

> so I am not sure what to change to bring all of them into alignment.

This is fairly deep internals of the MicroPython docs, so feel free to keep everything as-is for this PR and I'll address the inconsistency in a follow-up PR.

dpgeorge:
@k98kurz are you interested in updating this PR based on the above review?  There are just a few points left.

k98kurz:
@dpgeorge I applied the suggestions. Anything else I need to do?

tests/cpydiff/modules_struct_does_not_support_bool.py
```diff
@@ -0,0 +1,11 @@
+"""
+categories: Modules,struct
+description: Struct ``pack`` and ``unpack`` does not support the "?" type code for bools.
```
dpgeorge: does -> do

tests/cpydiff/modules_time_time_returns_int.py
```diff
@@ -0,0 +1,10 @@
+"""
```
dpgeorge, line 1: This difference with the time return type is well documented already in `docs/library/time.rst`.  I suggest removing this file.

tests/cpydiff/syntax_dict_unpacking.py
```diff
@@ -0,0 +1,12 @@
+"""
+categories: Syntax,dict
+description: Unpacking one dict's elements into a new dict with ``{'new_elem': val, **old_dict}`` is not supported.
+cause: Not implemented. Dict unpacking is only implemented when passing dict elements as function kwargs.
+workaround: Use ``new_dict = {'new_elem': val}; new_dict.update(old_dict)`` to copy values into new dicts.
+cause: Unknown
+workaround: use ``new_dict = {}; new_dict.update(old_dict)`` to copy values into new dicts.
```
dpgeorge, line 5: These two duplicated lines (cause/workaround) can be removed.

tests/cpydiff/syntax_list_tuple_unpacking.py
```diff
@@ -0,0 +1,15 @@
+"""
+categories: Syntax,tuple,list
+description: Unpacking one tuple or list's elements into a new object with ``[*things]`` or ``(*things)`` is not supported.
+cause: Not implemented. Sequence unpacking is only implemented when passing arguments to functions.
+cause: Unknown
```
dpgeorge: This line can be removed.

tests/cpydiff/syntax_list_tuple_unpacking.py
```diff
@@ -0,0 +1,15 @@
...
+workaround: use ``new = []; for t in old: new.append(t)`` to copy into new lists and then cast to tuple if a tuple is needed.
```
dpgeorge: Suggest an example like `new = [0] + old`

tests/cpydiff/modules_time_time_returns_int.py
k98kurz: It has been over a year since this bit me in the ass, but I spent multiple hours debugging something because I did not see the difference documented on the main micropython docs site. Maybe it is no longer a problem, but this was the corrective course of action for the specific documentation failure I encountered.

tests/cpydiff/syntax_list_tuple_unpacking.py
k98kurz: Done: `new = [] + old`

tests/cpydiff/syntax_list_tuple_unpacking.py
k98kurz: Done

tests/cpydiff/syntax_dict_unpacking.py
k98kurz: Done

tests/cpydiff/modules_struct_does_not_support_bool.py
k98kurz: Done

k98kurz:
I pressed a button on GitHub, and now GitHub is complaining that the commit messages generated by GitHub do not fit GitHub's enforced format.

> *sighs and takes out his git scalpel*

k98kurz:
@dpgeorge It looks like the `Build docs / build (pull_request)` check is failing due to the Unix port behaving differently from ESP and RPi Pico ports. Not sure what to do about that. The `codecov/project` can probably be remediated by rebasing. If the code formatting thing with ruff is an issue, I can update the file and push another commit. Just let me know.

tests/cpydiff/modules_time_time_returns_int.py
projectgus: Hi @k98kurz,

Regarding the current CI failure:

```
Error: Test has same output in CPython vs MicroPython: ../tests/cpydiff/modules_time_time_returns_int.py
```

... as mentioned above, cpydiff only works to document differences that exist between CPython and the MicroPython unix port. The [time.time() documentation](https://docs.micropython.org/en/latest/library/time.html#time.time) has a "Differences to CPython" highlight that includes these sentences:

> With MicroPython, only Unix port uses the same Epoch, and if floating-point precision allows, returns sub-second precision.

> Embedded hardware usually doesn’t have floating-point precision to represent both long time ranges and subsecond precision, so they use integer value with second precision.

So suggest these steps from here:

1) Remove `tests/cpydiff/modules_time_time_returns_int.py`, as cpydiff can't document this change automatically.
2) If the time.time() documentation is not clear enough, submit a change in `docs/library/time.rst` to improve the documentation there instead. Feel free to rewrite that entire block or split it into a couple of blocks if it makes it clearer to the reader what differences they could expect.

k98kurz: @projectgus My goal was to include it here: https://docs.micropython.org/en/latest/genrst/modules.html

Is there perhaps some way to update the cpydiff tool to fix its behavior? Perhaps some kind of comment override syntax to indicate it is hardware port-dependent. If not, I guess I can remove the file, but that's a bummer.

PR #16990, opened 2025-03-21
ports/modtime: Define and implement flags for int vs float return type from time.time().

AJMansfield:
### Summary

This adds two feature flags, and implementations of these flags to applicable ports:
- `MICROPY_PY_TIME_TIME_HAS_SUBSECONDS`, which different ports and boards can define to indicate that their `mp_time_time_get` implementation supports a floating-point output with meaningful sub-second information.
- `MICROPY_PY_TIME_TIME_FLOAT`, which controls whether the floating-point or integer version of an applicable port's `mp_time_time_get` are selected.

In the future, as a micropython 2.0 breaking change we might redefine `MICROPY_PY_TIME_TIME_FLOAT`; but at present the defaults for this flag are set so that the existing default behavior of all affected ports remains the same.

The esp32, esp8366, rp2 (rp2350 only), stm32 (rtc_us only), and zephyr ports have been updated to produce floating-point time when the latter flag is enabled.
The webassembly port has been updated to only output integers when this is disabled.
The unix port already had both behaviors, but has been refactored to use these feature flags for it rather than its own conditional.

Thanks to @k98kurz for the cpydiffs in #16984 that prompted me to explore this.

### Testing

A lot of the testing here involves custom board variants to enable double-precision support; see my [float_time_test2](https://github.com/AJMansfield/micropython/tree/float_time_test2) branch for the test configurations.

esp32: I've performed successful build tests for both DP and non-DP variants. Still need functional tests to verify the implementation produces meaningful subsecond information.

esp8266: I'm at the limit of my expertise here trying to get a usable softdouble library to fit in order to build a firmware image here. Except for the firmware-size related link error, though, my DP-enabled variant does successfully build with floating-point time.

rp2: I've performed functional on-hardware tests with each combination of DP/non-DP and float-time/int-time variants on both RPI_PICO and RPI_PICO2; works as expected.

stm32: I've performed successful build tests for three out of four relevant configurations; like rp2 there's a difference in support between different boards depending on their MICROPY_HW_RTC_USE_US support. The current missing case is a DP variant with MICROPY_HW_RTC_USE_US=1, like with esp8266 I've reached my limit on expertise trying to get a softdouble library to fit on an applicable MICROPY_HW_RTC_USE_US board.

unix: I've performed functional tests with both float-time and int-time board definitions; works as expected.

webassembly: 've performed functional tests with the standard DP-enabled variant. This still needs to be tested on a variant with MICROPY_PY_TIME_TIME_FLOAT explicitly disabled.

zephyr: I've performed successful build tests for both DP and non-DP variants. Still need functional tests.


robert-hh:
I do not really have the impression that this change is useful. For me, compatibility with CPython is less important than compatibility between and within ports. And this change makes it worse. Looking at which ports this change would be effective by default, it creates more differences. MICROPY_FLOAT_IMPL_FLOAT is enabled by default on a fraction of STM32 and MIMXRT boards. So at least, this option should be guarded by a dedicated build flag for those how want time.time() to return float.

Notes:
Only the last commit seems to belong to this PR topic. The other four are from a different commit of yours.

> I don't have access to hardware for most of these ports, so most of these changes are presently untested. The code in question is not complex, however, and it should be clear from inspection that the logic it implements is correct.

Untested usually means, that a PR cannot be accepted.

AJMansfield:
> I do not really have the impression that this change is useful. For me, compatibility with CPython is less important than compatibility between and within ports. And this change makes it worse. Looking at which ports this change would be effective by default, it creates more differences. MICROPY_FLOAT_IMPL_FLOAT is enabled by default on a fraction of STM32 and MIMXRT boards. So at least, this option should be guarded by a dedicated build flag for those how want time.time() to return float.

That does sound more useful. That would also provide a way to turn float output _off_ for unix and webassembly for anyone who wants a configuration that's more compatible with the other platforms.

I imagine we could add something like this to `mpconfig.h`:

```C
#ifndef MICROPY_EXTMOD_TIME_FLOAT
#define MICROPY_EXTMOD_TIME_FLOAT (MICROPY_PY_BUILTINS_FLOAT && MICROPY_FLOAT_IMPL >= MICROPY_FLOAT_IMPL_DOUBLE)
#endif
```

> Notes: Only the last commit seems to belong to this PR topic. The other four are from a different commit of yours.

Thanks for that catch, I was exploring this after looking at @k98kurz's cpydiffs and forgot to rebase.

> Untested usually means, that a PR cannot be accepted.
> 
> Edit: A few tests: Just setting MICROPY_FLOAT_IMPL to MICROPY_FLOAT_IMPL_DOUBLE at the ESP8266 causes the build to fail. Enabling MICROPY_FLOAT_IMPL_DOUBLE at a ESP32 works and produces changing subsecond values. Enabling MICROPY_FLOAT_IMPL_DOUBLE at the RP2 port builds, but the subsecond fraction is always 0 for RP2040. RP2350 has at least a ms resolution for the fraction. Building for as PYBD_SF6 succeeds and the result of time.time() is a float, but the subseconds count down instead of counting up.
> 
> So much about "The code in question is not complex, however, and it should be clear from inspection that the logic it implements is correct."

Thanks for taking a look at this and demonstrating the issue! Point taken: I should've at least tested to verify if it all compiled, with applicable defines on those ports, even without hardware -- that was an amateur move on my part. (I also do mean to acquire a wider range of the supported boards for the sake of being able to test things like this on hardware, I just haven't managed to yet.)

> Edit 2: For STM32 you can use `rtc_subsec_to_us(time.SubSeconds)` from rtc.c to get a proper up-counting us sequence. The resolution is 1/256 sec resp. 1/8196 sec.

Useful note, thanks for that!

robert-hh:
Using `rtc_subsec_to_us(time.SubSeconds)` requires a change to rtc.c/rtc.h as well, since it is not for all boards defined as a function. For some it is just an empty define.

robert-hh:
Looking more into it, `rtc_subsec_to_us(time.SubSeconds)` can only be used for ports which have `MICROPY_HW_RTC_USE_US` defined and set to non-zero. Otherwise the time.SubSeconds field just return the actual value of the RTC register. That will as well result in a value for the datetime() subseconds not expressing microseconds. Even if that is not specified as such, it seems odd.  `MICROPY_HW_RTC_USE_US`  is set to 0 for many STM32 boards!

robert-hh:
In all respect I do not agree that this PR makes "time.time() output type more consistent between ports." It looks more like the opposite. At the moment, the output type of time.time() is **float** for the unix port and **int** for all hardware ports. With that change active there is a mix between and within the various hardware ports of float and int for the time.time() return type. Some ports like CC3200 will never return a float, other ports like the MIMXRT, SAMD lack a subsecond fraction of the clock source used for time.time(). And at ports like RP2 and STM32 some MCU's firmware will return int, some will return float. In case of int they could return a float, with the fraction set to 0. But with the actual change this is not implemented.

AJMansfield:
Mhm, noting an additional latent error in my current stm32 port, TODO: need to provide a fallback code path to still emit a float return value in the specific case where `MICROPY_PY_TIME_TIME_FLOAT` is manually enabled but there's no `rtc_subsec_to_us` available to call...

k98kurz:
It might be simpler and less confusing to just make the Unix port return an int like all other ports, and then I can update the documentation PR to include information about optionally enabling float behavior during builds with a build flag.

peterhinch:
I would guess that the `time` and `RTC` APIs are likely targets for MP V2.0. There are numerous inconsistencies between ports, but fixing them will cause breakages.

robert-hh:
The latter sounds better. There are boards like ESP8266 which will most likely never are candidates for float time(), just because of the code size and RAM usage penalty that comes along with it.

> with a warning on ports that lack double precision or sub-second info, 

Please no runtime warnings at functions that are subject to frequent use. Just leave that to the documentation. Its not only ports, but also boards of a port.

I still do not see a mandatory use case for that, besides CPython compatibility. The subsecond information can be used to take the difference between time events that are longer than the range of ticks_us() (~18 Minutes) or ticks_ms() (~12 days). But in contrast to the ticks_xx() functions the resolution is MCU-specific. So it's of very limited use.

extmod/modtime.c
```diff
@@ -100,10 +101,26 @@ MP_DEFINE_CONST_FUN_OBJ_1(mp_time_mktime_obj, time_mktime);
...
+#error "MICROPY_PY_TIME_TIME_FLOAT: Port does not have floating point support."
+#endif
+
+#if MICROPY_PY_TIME_TIME_FLOAT && !(MICROPY_FLOAT_IMPL >= MICROPY_FLOAT_IMPL_DOUBLE) && MICROPY_PY_TIME_TIME_FLOAT_WARNIMPRECISE
+#warning "MICROPY_PY_TIME_TIME_FLOAT: Port does not have adequate float precision. (`#define MICROPY_PY_TIME_TIME_FLOAT_WARNIMPRECISE (0)` to suppress.)"
+#endif
+
+#if MICROPY_PY_TIME_TIME_FLOAT && !MICROPY_PY_TIME_TIME_HAS_SUBSECOND && MICROPY_PY_TIME_TIME_FLOAT_WARNFUTILE
+#warning "MICROPY_PY_TIME_TIME_FLOAT: Port does not have sub-second information. (`#define MICROPY_PY_TIME_TIME_FLOAT_WARNFUTILE (0)` to suppress.)"
+#endif
```
AJMansfield, line 114: > Please no runtime warnings at functions that are subject to frequent use. Just leave that to the documentation. Its not only ports, but also boards of a port.

I only mean a set of compile-time `#warning`s like this. (And one `#error`, to give a slightly more useful message about a configuration that could never be made to build but likely for more opaque reasons otherwise.)

py/mpconfig.h
```diff
@@ -1631,6 +1631,30 @@ typedef double mp_float_t;
...
+// - Short floats (i.e. REPR_C) would lose seconds past 4,194,304.0 i.e. unix Feb 18, 1970
+// - Double-precision floats will only lose seconds past 9,007,199,254,740,992.0 i.e. unix November 12, 285428751
+#ifndef MICROPY_PY_TIME_TIME_FLOAT_WARNIMPRECISE
+#define MICROPY_PY_TIME_TIME_FLOAT_WARNIMPRECISE (1)
+#endif
+
+// Emit a warning if MICROPY_PY_TIME_TIME_FLOAT is selected on a port without a sub-second RTC.
+#ifndef MICROPY_PY_TIME_TIME_FLOAT_WARNFUTILE
+#define MICROPY_PY_TIME_TIME_FLOAT_WARNFUTILE (1)
+#endif
```
AJMansfield, line 1656: (And since micropython is usually build with warning -> error, a set of flags for turning those warnings off.)

extmod/modtime.c
AJMansfield, line 114: > Please no runtime warnings at functions that are subject to frequent use. Just leave that to the documentation. Its not only ports, but also boards of a port.

I only mean a set of compile-time `#warning`s like this. (And one `#error`, to give a slightly more useful message about a configuration that could never be made to build but likely for more opaque reasons otherwise.)

py/mpconfig.h
AJMansfield, line 1656: With defines able to turn them off since micropython is usually compiled with flags to make all warnings errors.

projectgus (commented):
Appreciate the effort and thought you've put into this, @AJMansfield, but I have some concerns.

I agree with most of @robert-hh's comments above, but there is a more fundamental problem which is that changing in either direction (some ports now return float time, or unix now returns int time) will break some existing MicroPython code on update.

So if we were going to make such a change in existing port defaults then I think it would need to be gated behind `MICROPY_PREVIEW_VERSION_2` as a breaking 2.0 change.

However, I'm not convinced changing in either direction is an improvement. It's a lot easier to document "embedded ports return an int from time.time()" rather than having a list of ports and configs which *may* do this. It's a personal niggle to me that a lot of the docs already say *"Some ports do this differently, good luck!"* without any specifics.

(Although, as mentioned in the other PR, maybe part of the existing `time.time()` doc needs to be edited to highlight this difference more clearly.)

I'd also note that even having `time.time()` return a float doesn't make this function strictly CPython compatible, because CPython returns a float of seconds since Jan 1 1970. So there's always going to be *some* difference here for embedded ports, that some developers will need to pay attention to.

AJMansfield:
> it would need to be gated behind `MICROPY_PREVIEW_VERSION_2` as a breaking 2.0 change.

Not a flag I was aware of previously, but I agree: any change to the defaults should be gated thus.

At present, as of [e16afc7](https://github.com/micropython/micropython/pull/16990/commits/e16afc7af87126bd68a77cd9430c833e2aec9e38) this PR is implemented in a way that should have no observable effect, unless a user defines and builds against a custom board definition that sets `MICROPY_PY_TIME_TIME_FLOAT`.

Probably the most important use-case for this patch is just to allow building `unix` and `webassembly` in a way that lets them better approximate the behavior of other pieces of hardware when unit-testing application code -- i.e. the case where code that previously seemed to run fine now "correctly" breaking is in some sense a bugfix.

PR #16991, opened 2025-03-22
ports/rp2: Replace pico-sdk assertion faults with python exceptions.

AJMansfield:
### Summary

This change replaces the default pico-sdk assertion fault handlers with new handlers that raise micropython exceptions.

### Testing

I've verified that this works on real rp2040 hardware, using a custom module to trigger the pico-sdk assertion mechanism:

```C
#include "py/runtime.h"
#include "pico/assert.h"

#define PARAM_ASSERTIONS_ENABLED_FAIL (1)

static mp_obj_t rp2_param_fault(void) {
    invalid_params_if(FAIL, true);
    return mp_const_none;
}
MP_DEFINE_CONST_FUN_OBJ_0(rp2_param_fault_obj, rp2_param_fault);

static mp_obj_t rp2_hard_fault(void) {
    hard_assert_if(FAIL, true);
    return mp_const_none;
}
MP_DEFINE_CONST_FUN_OBJ_0(rp2_hard_fault_obj, rp2_hard_fault);

static const mp_rom_map_elem_t fail_module_globals_table[] = {
    { MP_ROM_QSTR(MP_QSTR___name__),            MP_ROM_QSTR(MP_QSTR_fail) },
    { MP_ROM_QSTR(MP_QSTR_param),               MP_ROM_PTR(&rp2_param_fault_obj) },
    { MP_ROM_QSTR(MP_QSTR_hard),                MP_ROM_PTR(&rp2_hard_fault_obj) },
};
static MP_DEFINE_CONST_DICT(fail_module_globals, fail_module_globals_table);

const mp_obj_module_t mp_module_fail = {
    .base = { &mp_type_module },
    .globals = (mp_obj_dict_t *)&fail_module_globals,
};

MP_REGISTER_MODULE(MP_QSTR_fail, mp_module_fail);
```

Which leads to:

```
MicroPython v1.25.0-preview.400.ge3adb89f8.dirty on 2025-03-21; Raspberry Pi Pico W with RP2040
>>> import fail
>>> fail.hard()
AssertionError: assertion failure in pico-sdk FAIL
>>> fail.param()
ValueError: invalid params in pico-sdk FAIL
```

### Tradeoffs and Alternatives

It's not entirely clear what the best choice of exception and exception message is, and I invite comments on this topic.

At present I'm using variadic exceptions with a parameter derived from the name of the associated assertion enable gate. `invalid_params_if` and `valid_params_if` both map to `ValueError("invalid params in pico-sdk %s")`, and `hard_assert_if` failures map to `AssertionError("assertion failure in pico-sdk %s")`. Should the latter perhaps be a `RuntimeError` rather than an `AssertionError`, though?

### DIFF
--- a/ports/rp2/CMakeLists.txt
+++ b/ports/rp2/CMakeLists.txt
@@ -24,6 +24,9 @@ set(PICO_BTSTACK_PATH ${MICROPY_DIR}/lib/btstack)
 # Set the location of this port's directory.
 set(MICROPY_PORT_DIR ${CMAKE_CURRENT_LIST_DIR})
 
+# Replace the pico/assert.h assertion mechanism with one that raises python exceptions.
+list(APPEND PICO_CONFIG_HEADER_FILES ${MICROPY_PORT_DIR}/pico_assert.h)
+
 # Set the board if it's not already set.
 if(NOT MICROPY_BOARD)
     set(MICROPY_BOARD RPI_PICO)
@@ -163,6 +166,7 @@ set(MICROPY_SOURCE_PORT
     mpthreadport.c
     mutex_extra.c
     pendsv.c
+    pico_assert.c
     rp2_flash.c
     rp2_pio.c
     rp2_dma.c
@@ -189,6 +193,7 @@ set(MICROPY_SOURCE_QSTR
     ${MICROPY_PORT_DIR}/machine_wdt.c
     ${MICROPY_PORT_DIR}/modrp2.c
     ${MICROPY_PORT_DIR}/modos.c
+    ${MICROPY_PORT_DIR}/pico_assert.c
     ${MICROPY_PORT_DIR}/rp2_flash.c
     ${MICROPY_PORT_DIR}/rp2_pio.c
     ${MICROPY_PORT_DIR}/rp2_dma.c
--- a/ports/rp2/pico_assert.c
+++ b/ports/rp2/pico_assert.c
@@ -27,9 +27,9 @@
 #include "pico_assert.h"
 #include "py/runtime.h"
 
-void pico_param_fault(char *x) {
-    mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("invalid params in pico-sdk %s"), x);
+void pico_param_fault(char *define, char *file, unsigned int line, char *test) {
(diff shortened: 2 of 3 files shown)


Gadgetoid:
Woof, is the 4k size gain because you've forced all assertions to be enabled here?

This would probably be extremely useful for debugging user C modules, of which we (Pimoroni) have way, way too many, but should *definitely* be optional. Only a developer should ever see/care about these asserts, and if MicroPython is hitting a condition where an end user can trigger one then I'd argue that's a bug in MicroPython.

But otherwise 👍 knowing how to hook this up might have saved us some real headaches.

AJMansfield:
> Woof, is the 4k size gain because you've forced all assertions to be enabled here?

Come to think of it, maybe `#define PARAM_ASSERTIONS_ENABLE_ALL 1` is _not_ the correct value to be setting here...

> This would probably be extremely useful for debugging user C modules, of which we (Pimoroni) have way, way too many, but should _definitely_ be optional. Only a developer should ever see/care about these asserts, and if MicroPython is hitting a condition where an end user can trigger one then I'd argue that's a bug in MicroPython.

Are there some debug assertion defines that already exist that could be used to gate this?

Gadgetoid:
> Are there some debug assertion defines that already exist that could be used to gate this?

Not as far as I'm aware, there's some debug stuff here but I've... surprisingly little experience with it 😬

https://github.com/micropython/micropython/blob/f1018ee5c2dd0eabcc0cfe7e02c607594e489788/py/mpconfig.h#L501-L544

The ESP8266 port has an interesting (read: uncomplicated) take on this: 

https://github.com/micropython/micropython/blob/f1018ee5c2dd0eabcc0cfe7e02c607594e489788/ports/esp8266/esp_mphal.c#L131-L134

There doesn't really seem to be a consensus on dealing with SDK assertions in MicroPython itself, presumably because the various SDKs all have their own way of doing things.

AJMansfield:
It took me way too long messing around with trying to convert the assert flag's name into a qstr instead of just a raw char*, but yes: all of that size increase did indeed come down to that `#define PARAM_ASSERTIONS_ENABLE_ALL 1` line.

On the plus side, it looks like with that set to 0 now, the the compiler even ends up even fully eliding the text for `pico_param_fault` and `pico_hard_fault`, so there's no size penalty unless you configure your port/variant with some or all of the assertions enabled.

PR #16993, opened 2025-03-22
esp32/esp32_common.cmake: RISC-V improvements.

agatti:
### Summary

This PR contains some RISC-V specific improvements to the ESP32 port's build script.

The main change is the migration of the ESP32C3/ESP32C6 gchelper implementation over to the native version.  This is to make the RISC-V boards' behaviour in line with what most other ports do, and to follow the request found in the generic gchelper source file: to use the native version if possible.  This should also be compatible with future ESP32C2 board support, as the RISC-V CPU version in use is the same as the ESP32C6.

Two other changes perform some minor cleanup work by removing obsolete directives, with the last providing the proper architecture name to `mpy-cross` when needed.  The old implementation did that only for Xtensa.

### Testing

A build for ESP32_GENERIC_C3 was built without errors and the full test suite was run on an ESP32C3 board with the exactly same behaviour:  `machine_i2s_rate` fails on my board both with current `master` and with my changes.


projectgus (approved):
Nice cleanups, thanks @agatti!

Issue #16845, opened 2025-03-03
mpremote rmdir /   - Doesn't work


ZacharyACoon:
### Port, board and/or hardware

esp32c6 super micro

### MicroPython version

ESP32_GENERIC_S3-20241129-v1.24.1.bin

### Reproduction

`mpremote rmdir /`

```
rmdir :/
mpremote: Error with transport:
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
OSError: [Errno 1] EPERM
```

### Expected behaviour

Expected to erase all files from micropython's root.

### Observed behaviour

`OSError: [Errno 1] EPERM`

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

dpgeorge:
mpremote's commands are based on unix commands of the same name.  Under unix, `rmdir` only removes *empty* directories.  So, it's not expected that it would remove the contents of that directory first.

Instead, on unix one would use `rm -rf /*`.  So if we wanted this feature to remove everything, we would need to implement `rm -r` (recursive remove).

ZacharyACoon:
> mpremote's commands are based on unix commands of the same name. Under unix, `rmdir` only removes _empty_ directories. So, it's not expected that it would remove the contents of that directory first.
> 
> Instead, on unix one would use `rm -rf /*`. So if we wanted this feature to remove everything, we would need to implement `rm -r` (recursive remove).

Is there an expected process to sync a directory or clear out the storage so that I can write the files I expect to be there?

ricksorensen:
Checkour https://github.com/orgs/micropython/discussions/9802 for a discussion of ways to clear out filesystem.  Also, `mpremote` uses hashes to check if destination file is same as source and will not copy  if they are the same.

peterhinch:
`rm -r` would be *very* useful.

dpgeorge:
Recursive remove has now been implemented in `mpremote`.  I suggest using `rm -rv` to show exactly what was removed.

PR #16994, opened 2025-03-22
mpremote: Add `rm -r` recursive remove functionality to filesystem commands.

Josverl:
### Summary

Recursive remove functionality for mpremote is frequently requested by users.

- [#16845](https://github.com/orgs/micropython/discussions/9802)
- https://github.com/micropython/micropython/issues/16845#issuecomment-2695120895

### Testing

On Windows 
- manually 
- using pytest, after porting the script based test suite to pytest ( not part of this PR) 

On Linux
- tests for linux have been added, and run against 5 ports from WSL2

- The current implementation does **not** ask for conformation, so it behaves somewhat like `rm -rf :foo` and therefore should be used with caution. 
  If desired such a check, and its override '-rf' could be added 
- No attempt is made to detect, or avoid special paths such as `/rom`  or `/sd`

Fixes : 16845

### DIFF
--- a/docs/reference/mpremote.rst
+++ b/docs/reference/mpremote.rst
@@ -229,7 +229,7 @@ The full list of supported commands are:
   - ``ls`` to list the current directory
   - ``ls <dirs...>`` to list the given directories
   - ``cp [-rf] <src...> <dest>`` to copy files
-  - ``rm <src...>`` to remove files on the device
+  - ``rm [-r] <src...>`` to remove files or folders on the device
   - ``mkdir <dirs...>`` to create directories on the device
   - ``rmdir <dirs...>`` to remove directories on the device
   - ``touch <file..>`` to create the files (if they don't already exist)
@@ -238,14 +238,34 @@ The full list of supported commands are:
   The ``cp`` command uses a convention where a leading ``:`` represents a remote
   path. Without a leading ``:`` means a local path. This is based on the
   convention used by the `Secure Copy Protocol (scp) client
-  <https://en.wikipedia.org/wiki/Secure_copy_protocol>`_. All other commands
-  implicitly assume the path is a remote path, but the ``:`` can be optionally
-  used for clarity.
+  <https://en.wikipedia.org/wiki/Secure_copy_protocol>`_.
 
   So for example, ``mpremote fs cp main.py :main.py`` copies ``main.py`` from
   the current local directory to the remote filesystem, whereas
   ``mpremote fs cp :main.py main.py`` copies ``main.py`` from the device back
   to the current directory.
+
+  The ``mpremote rm -r`` command accepts both relative and absolute paths.
+  Use ``:`` to refer to the current remote working directory (cwd) to allow a
+  directory tree to be removed from the device's default path (eg ``/flash``, ``/``).
+  Use ``-v/--verbose`` to see the files being removed.
+
+  For example:
+
+  - ``mpremote rm -r :libs`` will remove the ``libs`` directory and all its
+    child items from the device.
+  - ``mpremote rm -rv :/sd`` will remove all files from a mounted SDCard and result
+    in a non-blocking warning. The mount will be retained.
+  - ``mpremote rm -rv :/`` will remove all files on the device, including any
+    located in mounted vfs such as ``/sd`` or ``/flash``. After removing all folders
+    and files, this will  also return an error to mimic unix ``rm -rf /`` behaviour.
+
+  .. warning::
+    There is no supported way to undelete files removed by ``mpremote rm -r :``.
+    Please use with caution.
+
+  All other commands implicitly assume the path is a remote path, but the ``:``
+  can be optionally used for clarity.
 
   All of the filesystem sub-commands take multiple path arguments, so if there
   is another command in the sequence, you must use ``+`` to terminate the
--- a/tools/mpremote/mpremote/commands.py
+++ b/tools/mpremote/mpremote/commands.py
@@ -1,4 +1,5 @@
 import binascii
+import errno
 import hashlib
 import os
 import sys
@@ -300,6 +301,28 @@ def do_filesystem_recursive_cp(state, src, dest, multiple, check_hash):
         do_filesystem_cp(state, src_path_joined, dest_path_joined, False, check_hash)
 
 
+def do_filesystem_recursive_rm(state, path, args):
+    if state.transport.fs_isdir(path):
+        for entry in state.transport.fs_listdir(path):
+            do_filesystem_recursive_rm(state, _remote_path_join(path, entry.name), args)
+        if path:
+            try:
+                state.transport.fs_rmdir(path)
+                if args.verbose:
+                    print(f"removed directory: '{path}'")
+            except OSError as e:
+                if e.errno != errno.EINVAL:  # not vfs mountpoint
+                    raise CommandError(
+                        f"rm -r: cannot remove :{path} {os.strerror(e.errno) if e.errno else ''}"
+                    ) from e
+                if args.verbose:
+                    print(f"skipped: '{path}' (vfs mountpoint)")
+    else:
+        state.transport.fs_rmfile(path)
+        if args.verbose:
+            print(f"removed: '{path}'")
+
+
 def do_filesystem(state, args):
     state.ensure_raw_repl()
     state.did_action()
@@ -352,7 +375,10 @@ def do_filesystem(state, args):
             elif command == "mkdir":
                 state.transport.fs_mkdir(path)
             elif command == "rm":
-                state.transport.fs_rmfile(path)
+                if args.recursive:
+                    do_filesystem_recursive_rm(state, path, args)
+                else:
+                    state.transport.fs_rmfile(path)
             elif command == "rmdir":
                 state.transport.fs_rmdir(path)
             elif command == "touch":
--- a/tools/mpremote/mpremote/main.py
+++ b/tools/mpremote/mpremote/main.py
@@ -182,7 +182,7 @@ def argparse_rtc():
 
 def argparse_filesystem():
     cmd_parser = argparse.ArgumentParser(description="execute filesystem commands on the device")
-    _bool_flag(cmd_parser, "recursive", "r", False, "recursive copy (for cp command only)")
+    _bool_flag(cmd_parser, "recursive", "r", False, "recursive (for cp and rm commands)")
     _bool_flag(
         cmd_parser,
         "force",
--- a/tools/mpremote/mpremote/transport.py
+++ b/tools/mpremote/mpremote/transport.py
@@ -24,7 +24,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-import ast, hashlib, os, sys
+import ast, errno, hashlib, os, sys
 from collections import namedtuple
 
 
@@ -63,6 +63,10 @@ def _convert_filesystem_error(e, info):
         return FileExistsError(info)
     if "OSError" in e.error_output and "ENODEV" in e.error_output:
         return FileNotFoundError(info)
+    if "OSError" in e.error_output and "EINVAL" in e.error_output:
+        return OSError(errno.EINVAL, info)
+    if "OSError" in e.error_output and "EPERM" in e.error_output:
+        return OSError(errno.EPERM, info)
     return e
 
 
--- a/tools/mpremote/tests/README.md
+++ b/tools/mpremote/tests/README.md
@@ -4,11 +4,26 @@ This directory contains a set of tests for `mpremote`.
 
 Requirements:
 - A device running MicroPython connected to a serial port on the host.
+- The device you are testing against must be flashed with a firmware of the same build
+  as `mpremote`.
+- If the device has an SDcard or other vfs mounted, the vfs's filesystem must be empty
+  to pass the filesystem test.
 - Python 3.x, `bash` and various Unix tools such as `find`, `mktemp`, `sed`, `sort`, `tr`.
+- To test on Windows, you can either:
+    - Run the (Linux) tests in WSL2 against a USB device that is passed though to WSL2.
+    - Use the `Git Bash` terminal to run the tests against a device connected to a COM
+      port. _Note:_ While the tests will run in `Git Bash`, several will throw false
+      positive errors due to differences in the way that TMP files are logged and and
+      several other details.
 
 To run the tests do:
 
+    $ cd tools/mpremote/tests
     $ ./run-mpremote-tests.sh
+
+To run a single test do:
+
+    $ ./run-mpremote-tests.sh test_filesystem.sh
 
 Each test should print "OK" if it passed.  Otherwise it will print "CRASH", or "FAIL"
 and a diff of the expected and actual test output.
--- a/tools/mpremote/tests/test_filesystem.sh
+++ b/tools/mpremote/tests/test_filesystem.sh
@@ -170,3 +170,67 @@ EOF
 $MPREMOTE resume cp -r "${TMP}/package" :
 $MPREMOTE resume ls : :package :package/subpackage
 $MPREMOTE resume exec "import package; package.x(); package.y()"
+
+echo -----
+# Test rm -r functionality
+# start with a fresh ramdisk before each test
+# rm -r MCU current working directory
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume touch :a.py
+$MPREMOTE resume touch :b.py
+$MPREMOTE resume cp -r "${TMP}/package" :
+$MPREMOTE resume rm -r -v :
+$MPREMOTE resume ls :
+$MPREMOTE resume ls :/ramdisk
+
+echo -----
+# rm -r relative subfolder
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume touch :a.py
+$MPREMOTE resume mkdir :testdir
+$MPREMOTE resume cp -r "${TMP}/package" :testdir/package
+$MPREMOTE resume ls :testdir
+$MPREMOTE resume ls :testdir/package
+$MPREMOTE resume rm -r :testdir/package
+$MPREMOTE resume ls :/ramdisk
+$MPREMOTE resume ls :testdir
+
+echo -----
+# rm -r non-existent path
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume ls :
+$MPREMOTE resume rm -r :nonexistent || echo "expect error"
+
+echo -----
+# rm -r absolute root
+# no -v to generate same output on stm32 and other ports
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume touch :a.py
+$MPREMOTE resume touch :b.py
+$MPREMOTE resume cp -r "${TMP}/package" :
+$MPREMOTE resume cp -r "${TMP}/package" :package2
+$MPREMOTE resume rm -r :/ || echo "expect error"
+$MPREMOTE resume ls :
+$MPREMOTE resume ls :/ramdisk
+
+echo -----
+# rm -r relative mountpoint
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume touch :a.py
+$MPREMOTE resume touch :b.py
+$MPREMOTE resume cp -r "${TMP}/package" :
+$MPREMOTE resume exec "import os;os.chdir('/')"
+$MPREMOTE resume rm -r -v :ramdisk
+$MPREMOTE resume ls :/ramdisk
+
+echo -----
+# rm -r absolute mountpoint
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume touch :a.py
+$MPREMOTE resume touch :b.py
+$MPREMOTE resume cp -r "${TMP}/package" :
+$MPREMOTE resume exec "import os;os.chdir('/')"
+$MPREMOTE resume rm -r -v :/ramdisk
+$MPREMOTE resume ls :/ramdisk
+
+echo -----
--- a/tools/mpremote/tests/test_filesystem.sh.exp
+++ b/tools/mpremote/tests/test_filesystem.sh.exp
@@ -191,3 +191,79 @@ ls :package/subpackage
           23 y.py
 x
 y2
+-----
+touch :a.py
+touch :b.py
+cp ${TMP}/package :
+rm :
+removed: './a.py'
+removed: './b.py'
+removed: './package/subpackage/__init__.py'
+removed: './package/subpackage/y.py'
+removed directory: './package/subpackage'
+removed: './package/__init__.py'
+removed: './package/x.py'
+removed directory: './package'
+ls :
+ls :/ramdisk
+-----
+touch :a.py
+mkdir :testdir
+cp ${TMP}/package :testdir/package
+ls :testdir
+           0 package/
+ls :testdir/package
+           0 subpackage/
+          43 __init__.py
+          22 x.py
+rm :testdir/package
+ls :/ramdisk
+           0 a.py
+           0 testdir/
+ls :testdir
+-----
+ls :
+rm :nonexistent
+mpremote: rm: nonexistent: No such file or directory.
+expect error
+-----
+touch :a.py
+touch :b.py
+cp ${TMP}/package :
+cp ${TMP}/package :package2
+rm :/
+mpremote: rm -r: cannot remove :/ Operation not permitted
+expect error
+ls :
+ls :/ramdisk
+-----
+touch :a.py
+touch :b.py
+cp ${TMP}/package :
+rm :ramdisk
+removed: 'ramdisk/a.py'
+removed: 'ramdisk/b.py'
+removed: 'ramdisk/package/subpackage/__init__.py'
+removed: 'ramdisk/package/subpackage/y.py'
+removed directory: 'ramdisk/package/subpackage'
+removed: 'ramdisk/package/__init__.py'
+removed: 'ramdisk/package/x.py'
+removed directory: 'ramdisk/package'
+skipped: 'ramdisk' (vfs mountpoint)
+ls :/ramdisk
+-----
+touch :a.py
+touch :b.py
+cp ${TMP}/package :
+rm :/ramdisk
+removed: '/ramdisk/a.py'
+removed: '/ramdisk/b.py'
+removed: '/ramdisk/package/subpackage/__init__.py'
+removed: '/ramdisk/package/subpackage/y.py'
+removed directory: '/ramdisk/package/subpackage'
+removed: '/ramdisk/package/__init__.py'
+removed: '/ramdisk/package/x.py'
+removed directory: '/ramdisk/package'
+skipped: '/ramdisk' (vfs mountpoint)
+ls :/ramdisk
+-----


tools/mpremote/mpremote/commands.py
```diff
@@ -300,6 +300,15 @@ def _mkdir(a, *b):
         do_filesystem_cp(state, src_path_joined, dest_path_joined, False, check_hash)
 
 
+def do_filesystem_recursive_rm(state, path):
+    if state.transport.fs_isdir(path):
+        for entry in state.transport.fs_listdir(path):
+            do_filesystem_recursive_rm(state, _remote_path_join(path, entry.name))
+        state.transport.fs_rmdir(path)
```
AJMansfield, line 307: The argument is a directory, and this function removes that directory; I'd argue that this should therefore be part of the `rmdir` command rather than the `rm` command.

tools/mpremote/tests/test_filesystem.sh
```diff
@@ -170,3 +170,20 @@ EOF
...
+$MPREMOTE resume ls :testdir
+
+# Test rm -r on non-existent path
+echo -----
+$MPREMOTE resume rm -r :nonexistent || echo "expect error"
+
+# Test rm -r on read-only filesystem
+echo -----
+$MPREMOTE resume exec "import os; os.mount(os.VfsFat(RAMBlockDev(512, 50)), '/readonly', readonly=True)"
+$MPREMOTE resume rm -r :readonly || echo "expect error"
```
AJMansfield, line 189: If you're implementing a preserve-root protection, the test cases need to include tests for it.

AJMansfield (changes requested):

tools/mpremote/tests/test_filesystem.sh
Josverl: The goal goal was not to include.

AJMansfield: Then the goal should be to delete the code implementing the root protection check entirely -- not to leave it untested.

tools/mpremote/mpremote/commands.py
```diff
@@ -352,7 +361,17 @@ def do_filesystem(state, args):
...
             elif command == "rm":
-                state.transport.fs_rmfile(path)
+                if args.recursive:
+                    if path == "":
+                        confirm = input(
+                            "Are you sure you want to delete all files and directories in the root directory? This action cannot be undone. (yes/no): "
+                        )
+                        if confirm.lower() != "yes":
+                            print("Aborted")
+                            return
```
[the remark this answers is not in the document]
Josverl: Agree this should not be in - as per the description above. 

and on breaking the CLI - yeah , not meant to be wrapped per docs - but not usable as a module today either.
that is another PR in the making.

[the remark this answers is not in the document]
AJMansfield: > not meant to be wrapped per docs

Sure, but let's not break things just to break them.

tools/mpremote/mpremote/main.py
```diff
@@ -182,7 +182,7 @@ def argparse_rtc():
 
 def argparse_filesystem():
     cmd_parser = argparse.ArgumentParser(description="execute filesystem commands on the device")
-    _bool_flag(cmd_parser, "recursive", "r", False, "recursive copy (for cp command only)")
+    _bool_flag(cmd_parser, "recursive", "r", False, "recursive copy (for cp and rm commands only)")
```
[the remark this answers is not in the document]
Josverl: the help line is still significantly shorter (64 chars) and well with the 80 cols I think is reasonable for a cli tool.

```
PS D:\mypython\micropython> mpremote fs --help
usage: fs [-h] [--recursive | --no-recursive] [--force | --no-force] [--verbose | --no-verbose] command path [path ...] ...

execute filesystem commands on the device

positional arguments:
  command          filesystem command (e.g. cat, cp, sha256sum, ls, rm, rmdir, touch)
  path             local and remote paths
  next_command     Next mpremote command

options:
  -h, --help       show this help message and exit
  --recursive, -r  recursive copy (for cp and rm commands only)
  --no-recursive
  --force, -f      force copy even if file is unchanged (for cp command only)
  --no-force
  --verbose, -v    enable verbose output (defaults to True for all commands except cat)
  --no-verbose
  
  ```

**Comment by Josverl** [comment]
can someone give this a testrun on a physical device just to make sure ?
 - [ ] linux
 - [ ] Mac

File: tools/mpremote/mpremote/main.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (AJMansfield) [reply tools/mpremote/mpremote/main.py]: Fair enough, I don't disapprove if you think this is better.

Though, I do really like the way GNU coreutils rm describes their option:

```
-r, -R, --recursive   remove directories and their contents recursively
```

The "directories and their contents" wording is pretty keen IMO -- it makes it blatantly clear even to a complete beginner what it does (even if they're still learning verbiage like "recursive" or are unused to the convention).

**Comment by AJMansfield** [comment]
I'll test run this in my environment tomorrow -- it's WSL that I've mounted my devices into with usbipd, but that's still linux enough to count lol.

File: tools/mpremote/tests/test_filesystem.sh (hunk 2 above)
Comment (AJMansfield) [inline tools/mpremote/tests/test_filesystem.sh:189]: Found another issue when running the test suite: the expected outputs haven't been added to `test_filesystem.sh.exp`!

File: tools/mpremote/tests/test_filesystem.sh [hunk 5]
```diff
@@ -170,3 +170,20 @@ EOF
 $MPREMOTE resume cp -r "${TMP}/package" :
 $MPREMOTE resume ls : :package :package/subpackage
 $MPREMOTE resume exec "import package; package.x(); package.y()"
+
+# Test rm -r functionality
+echo -----
+$MPREMOTE resume mkdir :testdir
+$MPREMOTE resume cp -r "${TMP}/package" :testdir/package
```
AJMansfield, line 177: Aside from the non-empty diff with `test_filesystem.sh.exp`, this line also generates another exception, such that the remaining test lines aren't even run:

```
~/micropython/tools/mpremote/tests$ ./run-mpremote-tests.sh 
./test_eval_exec_run.sh: OK
./test_filesystem.sh: FAIL
194,201d193
< -----
< mkdir :testdir
< cp ${TMP}/package :testdir/package
< mpremote: Error with transport:
< Traceback (most recent call last):
<   File "<stdin>", line 1, in <module>
< OSError: 28
< 
./test_mip_local_install.sh: OK
./test_mount.sh: OK
./test_recursive_cp.sh: OK
./test_resume.sh: OK
```

AJMansfield (changes requested):
The tests need to be fixed; the test runner crashes on one of them, and even the parts that are run aren't properly reflected in the .exp file.

My own preference is to split this into two separate commits for test code vs implementation code, with the implementation commit coming after the test commit. That way, it's easy to do proper verification, i.e. not just that the tests pass _with_ the implementation, but that they also fail without the implementation. 

You don't have to specifically do it that way, but I do expect tests with _both_ properties, and right now your tests only manage the 'fail' part.

tools/mpremote/tests/test_filesystem.sh
Josverl: Ah , that's indeed a bummer when you have tests that can't run on some platforms. 
All see if I can generate these in WSL2 as well.

tools/mpremote/tests/test_filesystem.sh
Josverl: Thanks for the report ,  Ill  dig-in to the test setup on linux

tools/mpremote/tests/test_filesystem.sh
AJMansfield: After inspecting the serial traces, I think this is the same issue I've had before with mpremote basically being just completely unusably broken. My rp2040 is just constantly soft-rebooting after just about every command; it's perhaps more impressive that the test even gets as far as it does before failing.

tools/mpremote/tests/test_filesystem.sh
AJMansfield: > Thanks for the report , Ill dig-in to the test setup on linux

If it helps, OSError 28 is ENOSPC, No space left on device.

AJMansfield: Ok, yes, I can confirm, increasing the size of the filesystem from 50 blocks to 100 blocks fixes this first issue.

tools/mpremote/tests/test_filesystem.sh
AJMansfield: That's not the issue here: the test harness includes a step that compares the output of `test_filesystem.sh` to the contents of `test_filesystem.sh.exp`.

This PR does not include _anything_ that would update `test_filesystem.sh.exp` with the additional lines of output that the additional lines of test code would generate.

Did you forget to stage and commit this file?

tools/mpremote/tests/test_filesystem.sh
AJMansfield: As written, in the state that the preceding tests leave the ramdisk filesystem, there is not enough space on that ramdisk for this test to pass.

Adjusting the number of blocks in `ramdisk.py` at the start of the test from 50 to 100 fixes this issue, at least.

Please run the entire test script end-to-end yourself, on a _clean_ repository -- no uncommitted files or changes; as if you'd just run `git reset --hard`.

tools/mpremote/tests/test_filesystem.sh
```diff
@@ -170,3 +170,20 @@ EOF
...
+$MPREMOTE resume rm -r :testdir/package
+$MPREMOTE resume ls :testdir
+
+# Test rm -r on non-existent path
+echo -----
+$MPREMOTE resume rm -r :nonexistent || echo "expect error"
+
+# Test rm -r on read-only filesystem
+echo -----
+$MPREMOTE resume exec "import os; os.mount(os.VfsFat(RAMBlockDev(512, 50)), '/readonly', readonly=True)"
```
AJMansfield, line 188:
```
OSError: [Errno 19] ENODEV
```

The RAMBlockDev object requires an mkfs call on it before it can successfully be mounted.

There's been some movement in the vfs API recently; make sure you are running the full test harness against a firmware version you built from your exact commit, not just some other release firmware version.

tools/mpremote/tests/test_filesystem.sh
Josverl: the problem with one big sequential test is that not all tests clean up at teardown.
Avoided the `OSError 28 is ENOSPC` issue by restarting the ramdrive in order to be able to start in a clean known state.

tools/mpremote/tests/test_filesystem.sh
Josverl: I missed clear docs on how this could be done on Windows.

Josverl:
split in multiple commits ( code / tests / docs ) 
tested in wsl2 - tests show success
tested  in Git Bash - Success but tests show a False Negative due to platform differences

AJMansfield (changes requested):
There could be some merit to using `vfs.mount()` output to skip failing on mountpounts, but special-casing it based on the directory name really isn't appropriate.

tools/mpremote/mpremote/commands.py
```diff
@@ -300,6 +300,16 @@ def _mkdir(a, *b):
         do_filesystem_cp(state, src_path_joined, dest_path_joined, False, check_hash)
 
 
+def do_filesystem_recursive_rm(state, path):
+    if state.transport.fs_isdir(path):
+        for entry in state.transport.fs_listdir(path):
+            do_filesystem_recursive_rm(state, _remote_path_join(path, entry.name))
+        if path not in ["/", "/ramdisk"]:
+            state.transport.fs_rmdir(path)
```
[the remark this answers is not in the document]
AJMansfield: You might also want to catch that `PermissionError` exception in `do_filesystem_recursive_rm`, so that the command can still remove directories that come after a mountpoint in their parent directory -- but it should still accumulate those errors if it gets them, and eventually raise a `CommandError` that gives one "operation not permitted" per directory it couldn't remove.

AJMansfield:
One other detail; with the `--verbose` flag this should output a `rm -r :/file` line for each individual file that's removed.

Josverl:
> rm shouldn't mysteriously fail to remove it just because it has a name it considers special
magic is never nice. A Try / Except - continue is somewhat better, but has a risk hide other special cases.

I do not want the CLI to terminate on hitting a mountpoint , as that requires knowledge that a user cannot have with current firmwares. Raising a Exception is useful in an API , not to send information to a user in a CLI. 
I can log a info or verbose message.

> so that the command can still remove directories that come after a mountpoint in their parent directory 
That already works en is part of the tests 

> with the --verbose flag this should output a rm :/file line for each file and rm -r :/dir for each directory
that would not be consistent with `cp -r`  that does not do that either, and I do not think that is useful for `rm -r` either.

```
$ mpremote cp -r --verbose ports/zephyr/boards/ : 
cp ports/zephyr/boards/ :
# copied multiple files
```

AJMansfield:
> I do not want the CLI to terminate on hitting a mountpoint , as that requires knowledge that a user cannot have with current firmwares. Raising a Exception is useful in an API , not to send information to a user in a CLI. I can log a info or verbose message.

Sure; the main point is that the way to detect this EPERM error is to add it to the code that detects the rest of the device-side errors, and raise the appropriate exception that follows the pattern.

Then in `do_filesystem_recursive_rm` you can try/catch for a PermissionError around the `state.transport.fs_rmdir(path)` call in order to output the appropriate message and continue.

> that would not be consistent with `cp -r` that does not do that either, and I do not think that is useful for `rm -r` either.

Oh, indeed? IMO it'd be good to make both subcommands' behavior more closely resemble GNU coreutils, rsync, etc -- but you're right that that's out of scope for this PR; that can be the subject of a separate PR and discussion later.

tools/mpremote/mpremote/commands.py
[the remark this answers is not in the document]
Josverl: There is no EPERM raised on MicroPython , it is EINVAL --> ValueError
<ay need to chack across different ports for the same condition 
But agree on the general approach 

A primary use case is people wanting to completely clear the flash 
`mpremote: rm: /: Operation not permitted.` would them make this PR mostly useless.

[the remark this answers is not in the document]
AJMansfield: > There is no EPERM raised on MicroPython , it is EINVAL --> ValueError Will need to check across different ports for the same condition But agree on the general approach

What port are you testing on? On my RP2040 it's most definitely an EPERM error.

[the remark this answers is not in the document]
AJMansfield: > A primary use case is people wanting to completely clear the flash `mpremote: rm: /: Operation not permitted.` would make this PR mostly useless.

I don't mean that it should fail fast on that, I mean that it should delete the directory's contents _and then fail on attempting to delete `/`. i.e. to be consistent with how, if passed some other mountpoint, it would delete that mount's contents and then error on deleting that mount's root folder.

tools/mpremote/mpremote/transport.py
```diff
@@ -63,6 +63,9 @@ def _convert_filesystem_error(e, info):
         return FileExistsError(info)
     if "OSError" in e.error_output and "ENODEV" in e.error_output:
         return FileNotFoundError(info)
+    if "OSError" in e.error_output and "EINVAL" in e.error_output:
+        return ValueError(info)
+    FileNotFoundError(info)
```
AJMansfield, line 68: What is this extra dangling `FileNotFoundError(info)`?

tools/mpremote/mpremote/transport.py
```diff
@@ -63,6 +63,9 @@ def _convert_filesystem_error(e, info):
         return FileExistsError(info)
     if "OSError" in e.error_output and "ENODEV" in e.error_output:
         return FileNotFoundError(info)
+    if "OSError" in e.error_output and "EINVAL" in e.error_output:
+        return ValueError(info)
```
AJMansfield, line 67: These are not just arbitrary; the rest of these cases are the established python errno mapping for that errno value. See https://docs.python.org/3/library/errno.html.

AJMansfield: EINVAL should map to `OSError(errno.EINVAL, info)`, not `ValueError(info)`.

AJMansfield (changes requested):

tools/mpremote/mpremote/transport.py
AJMansfield: All of the other exceptions it currently returns are ones that would be caught by `except OSError:`; returning an exception that is no longer a "corresponding OSError-derived exception" as documented breaks the function's type contract.

AJMansfield:
Regarding EINVAL vs EPERM, per the vfs source code EPERM is the error that's meant to be raised for an operation on a mountpoint root, e.g:

https://github.com/micropython/micropython/blob/master/extmod/vfs.c#L120-L121
https://github.com/micropython/micropython/blob/master/extmod/vfs.c#L266-L267
https://github.com/micropython/micropython/blob/master/extmod/vfs.c#L490-L491

EINVAL does happen, but only for trying to unmount a path that's not a mountpoint:

https://github.com/micropython/micropython/blob/master/extmod/vfs.c#L303-L305

tools/mpremote/mpremote/commands.py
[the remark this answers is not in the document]
Josverl: Tested on
Rp2, stm32, samd, esp32.

All return the same EINVAL.

Josverl:
Tested on 5 different ports, all raise the same error.
It can't hurt to catch the other error, but would like a way to trigger that error IRL,  if only to verify.
Perhaps with an SD card, rather than a ramdisk ?

Josverl:
Thanks for confirming.

That clears up all review comments AFAIKT.

The CI error on Mac appears to be a flakey build that I can't re-run

AJMansfield:
> That clears up all review comments AFAIKT.

One thing I'd consider still mandatory to fix, is the dead code introduced in the latest push at [tools/mpremote/mpremote/transport.py:68](https://github.com/micropython/micropython/pull/16994/files/d66a4984fdd2b0e5379ac57a8fded3b777529fc4#diff-08d700e284a379cf1c7d638b44a60ac5ec4c892c492a2f60da22cb6cbeac46cfR68), per [this comment](https://github.com/micropython/micropython/pull/16994#discussion_r2017626652).

(There's also the choice of python exception type mapping for EINVAL, per [these comments](https://github.com/micropython/micropython/pull/16994#discussion_r2017631047); but that's a more minor nitpick.)

tools/mpremote/mpremote/transport.py
Josverl: Makes indeed sense to keep it to a subclass of OSError

Josverl:
You have a point on OSError. 
Code and messages adjusted 

As I indicated before : I disagree that trying to remove all files and folder down from root should result in an error, as that would reduce usability too much.

@peterhinch , @mattytrentini, 
Do you have a preference whether or not an error should be raised on `mpremote -rm -r :/` ?

peterhinch:
I would prefer no error: clearing the filesystem is a normal thing to do, and users of `rm -r` should be aware that (as on Unix) it has the potential to be drastic.

AJMansfield:
The point of reporting that error isn't to error in a way that fails to remove the contents of `/`, just to report that the directory `/` itself still exists despite the ostensible request that the directory be removed.

What I'm describing _is_ the standard behavior for `rm -r /` -- or at least, the behavior implemented by both GNU coreutils and BusyBox rm:

#### GNU coreutils 9.1
```
user@host:~$ podman run -it debian
root@ecdfda5df5ad:/# mkdir /jail
root@ecdfda5df5ad:/# cp -r /lib /jail/lib
root@ecdfda5df5ad:/# cp -r /lib64 /jail/lib64
root@ecdfda5df5ad:/# cp -r /bin /jail/bin
root@ecdfda5df5ad:/# cp -r /usr /jail/usr
root@ecdfda5df5ad:/# ls /jail
bin  lib  lib64  usr
root@ecdfda5df5ad:/# chroot /jail rm -r --no-preserve-root /
rm: cannot remove '/': Device or resource busy
root@ecdfda5df5ad:/# echo $?
1
root@ecdfda5df5ad:/# ls /jail
root@ecdfda5df5ad:/# 
```

#### BusyBox v1.37.0
```
user@host:~$ podman run -it alpine
/ # mkdir /jail
/ # cp -r /lib /jail/lib
/ # cp -r /bin /jail/bin
/ # ls /jail
bin  lib
/ # chroot /jail rm -r /
rm: can't remove '/': Resource busy
/ # echo $?
1
/ # ls /jail
/ #
```

AJMansfield (approved):
I've tested this from WSL2 on both my rp2040 and rp2350.

Assuming the exit-0 behavior is what the maintainers prefer, this PR is ready to merge.

mattytrentini:
> Do you have a preference whether or not an error should be raised on mpremote -rm -r :/ ?

I think @AJMansfield presents a pretty compelling argument why there _should_ be an error - and I think it also satisfies @peterhinch's expectation that the folders contents will be removed (just not the directory itself).

Personally I'm not particularly invested, as long as I can delete the contents!

tools/mpremote/mpremote/commands.py
```diff
@@ -300,6 +300,19 @@ def _mkdir(a, *b):
         do_filesystem_cp(state, src_path_joined, dest_path_joined, False, check_hash)
 
 
+def do_filesystem_recursive_rm(state, path, args):
+    if state.transport.fs_isdir(path):
+        for entry in state.transport.fs_listdir(path):
+            do_filesystem_recursive_rm(state, _remote_path_join(path, entry.name), args)
+        if path not in ["/"]:
```
dpgeorge, line 307: This can just be `if path != "/"` (although maybe this check is not needed at all, if it should print an error about unable to remove root).

tools/mpremote/mpremote/commands.py
```diff
@@ -352,7 +365,13 @@ def do_filesystem(state, args):
             elif command == "mkdir":
                 state.transport.fs_mkdir(path)
             elif command == "rm":
-                state.transport.fs_rmfile(path)
+                if args.recursive:
+                    if path:
+                        do_filesystem_recursive_rm(state, path, args)
+                    else:
+                        raise CommandError("rm: missing path")
```
dpgeorge, line 372: I don't think it's possible to get to this line of code, the argument parser should guarantee that path is specified, right?

tools/mpremote/mpremote/commands.py
```diff
@@ -375,6 +394,8 @@ def do_filesystem(state, args):
         raise CommandError("{}: {}: File exists.".format(command, er.args[0]))
     except TransportError as er:
         raise CommandError("Error with transport:\n{}".format(er.args[0]))
+    except OSError as er:
+        raise CommandError(f"Error: {er}") from er
```
dpgeorge, line 398: This should match the other lines above in coding style, eg:
```py
raise CommandError("Error: {} {}".format(er.args[0], er.args[1])
```

tools/mpremote/mpremote/transport.py
```diff
@@ -63,6 +64,8 @@ def _convert_filesystem_error(e, info):
         return FileExistsError(info)
     if "OSError" in e.error_output and "ENODEV" in e.error_output:
         return FileNotFoundError(info)
+    if "OSError" in e.error_output and "EINVAL" in e.error_output:
+        return OSError(errno.EINVAL, info)
```
dpgeorge, line 68: Is this needed because trying to remove a mount point returns EINVAL?

Probably the VFS code should be fixed so attempting to do `os.remove('/')` raises EPERM instead of EINVAL, but that's out of scope for this PR.  So I guess catching EINVAL is OK for now.

tools/mpremote/tests/README.md
```diff
@@ -3,12 +3,26 @@
 This directory contains a set of tests for `mpremote`.
 
 Requirements:
-- A device running MicroPython connected to a serial port on the host.
+- A device running MicroPython connected to a serial port on the host, with the filesystem mounted as `/`
```
dpgeorge, line 6: The tests currently work fine on ports such as stm32 where the filesystem is not mounted on `/` (they are mounted on `/flash`).  And they should continue to run OK on such ports/boards.

tools/mpremote/tests/test_filesystem.sh
```diff
@@ -170,3 +170,46 @@ EOF
...
+$MPREMOTE resume ls :
+$MPREMOTE resume ls :/
+
+echo -----
+# "Test rm -r on root, with ramdisk mounted"
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume touch :/a.py
+$MPREMOTE resume touch :b.py
+$MPREMOTE resume cp -r "${TMP}/package" :
+$MPREMOTE resume rm -r :/
```
dpgeorge, line 212: Could this be `rm -r :/ramdisk` so it works on boards that don't have the internal filesystem mounted at `/`?

dpgeorge:
Regarding errors when trying to remove the root: some boards/ports have the internal filesystem mounted at `/flash` (eg stm32).  On such boards if you do `mpremote rm -r :/` then it'll go through all mounted partitions and delete everything from them and then try to delete the mount points.  That's perhaps not what you want.

Maybe one way to support boards with `/` and `/flash` as the main filesystem is instead to allow:
```
mpremote rm -r :
```
which would use the current directory as the target and remove everything from it, but crucially it would not attempt to remove the directory itself.  So that's kind of like `rm -rf *` on unix.  Implementing that is pretty easy and would give a generic way to remove all files in the main working directory, without issuing any errors.

Then it would still support `mpremote rm -r :/` but in that case try to remove `/` and print an error about it.

tools/mpremote/mpremote/commands.py
dpgeorge: If we want to support `mpremote rm -r :` then this check just needs to be `if path:`.

tools/mpremote/mpremote/commands.py
dpgeorge: If we want to support `mpremote rm -r :` then this check for path being non-zero is not needed.

tools/mpremote/mpremote/transport.py
Josverl: > trying to remove a mount point returns EINVAL

Indeed, that is through observation on 5 different ports

tools/mpremote/mpremote/commands.py
Josverl: You are correct, overly cautious 

partly due the guard above: 
```
    # ls implicitly lists the cwd.
    if command == "ls" and not paths:
        paths = [""]
```

AJMansfield:
> WRT raising error on root deletion: on ubuntu in docker (`docker container run -it ubuntu bash`) the following command does not even try to delete the root ⚠️ ` rm -rfv --no-preserve-root /`. ⚠️don't try this outside a throwaway docker / vm

Interesting! Note that it looks like coreutils rm skips trying to remove a parent directory if it errored trying to remove any of its children, and happens with any directory, not just `/`:

```
user@host:~$ podman run -it --mount=type=tmpfs,dst=/dir/fs --mount=type=tmpfs,dst=/dir/subdir/fs ubuntu bash
root@9c63b3159882:/# rm -rfv /dir
rm: cannot remove '/dir/fs': Device or resource busy
rm: cannot remove '/dir/subdir/fs': Device or resource busy
root@9c63b3159882:/# 
```

(Note that there was never an attempt to remove `/dir/` or `/dir/subdir/`.)

AJMansfield:
Note that this seems to be implementation-specific. BusyBox's behavior is somewhat different:

```
podman run -it --mount=type=tmpfs,dst=/dir/fs --mount=type=tmpfs,dst=/dir/subdir/fs alpine
/ # rm -rfv /dir
rm: can't remove '/dir/fs': Resource busy
rm: can't remove '/dir/subdir/fs': Resource busy
removed directory: '/dir/subdir'
removed directory: '/dir'
/ # echo $?
1
/ # ls /dir
fs      subdir
/ # 
```

It still exits nonzero -- but curiously, it log a success at removing the parent directories even though they still exist afterward.

tools/mpremote/mpremote/commands.py
Josverl: I changed the format as that lead to a too verbose error message repeating the same information multiple times.
With the updated logic the OSErrors are caught and handlerd deeper in `do_filesystem_recursive_rm` so this clause is no longer needed.

Josverl:
Current iteration works : 

- without specifying a path , effectively using the MCU's cwd
- with a relative path to a folder 
- with an absolute path to a subfolder 
- with an absolute path to a mount point (not a terminating error, `-v` shows "skipped: '/ramdisk' (vfs mountpoint)") 
- with a relative  path to a mount point (not a terminating error, `-v` shows "skipped: '/ramdisk' (vfs mountpoint)") 
- with the absolute root (Raises error) and exitcode = 1  

```
$ mpremote rm -rv : 
rm -r :
removed: './a.py'
removed: './b.py'
removed: './package/subpackage/__init__.py'
removed: './package/subpackage/y.py'
removed directory: './package/subpackage'
removed: './package/__init__.py'
removed: './package/x.py'
removed directory: './package'
```

```
$ mpremote rm -rv : 
rm -r :/
removed: '/ramdisk/a.py'
removed: '/ramdisk/b.py'
removed: '/ramdisk/package/subpackage/__init__.py'
removed: '/ramdisk/package/subpackage/y.py'
removed directory: '/ramdisk/package/subpackage'
removed: '/ramdisk/package/__init__.py'
removed: '/ramdisk/package/x.py'
removed directory: '/ramdisk/package'
removed: '/ramdisk/package2/subpackage/__init__.py'
removed: '/ramdisk/package2/subpackage/y.py'
removed directory: '/ramdisk/package2/subpackage'
removed: '/ramdisk/package2/__init__.py'
removed: '/ramdisk/package2/x.py'
removed directory: '/ramdisk/package2'
skipped: '/ramdisk' (vfs mountpoint)
mpremote: rm -r: cannnot remove :/ Operation not permitted
```

Also supports -v , and exit = 1 on attempt to remove root.
tested on stm32 with / without SDcard , ESP32 with & without SD card , and SAMD with several versions of MicroPython

AJMansfield:
I have an alternative implementation of `do_filesystem_recursive_rm` I've been playing with as well, that combined with #17090 would make the way errors get propagated much more consistent:

```
def do_filesystem_recursive_rm(state, path, args):
    if state.transport.fs_isdir(path):
        for entry in state.transport.fs_listdir(path):
            try:
                do_filesystem_recursive_rm(state, _remote_path_join(path, entry.name), args)
            except OSError as er:
                from .main import _PROG

                # Make sure existing stdout appears before the error message on stderr.
                sys.stdout.flush()
                print(
                    "{}: rm: {}: {}.".format(_PROG, er.strerror, os.strerror(er.errno)),
                    file=sys.stderr,
                )
                sys.stderr.flush()
        state.transport.fs_rmdir(path)
    else:
        state.transport.fs_rmfile(path)
```

This restructures it to specifically _only_ catch the OSError generated by the recursive child calls, and intentionally allows any error generated by `state.transport.fs_rmdir(path)` or `state.transport.fs_rmfile(path)` to propagate up so that the base exception handlers in `do_filesystem` still see the final `OSError(errno.ENOTEMPTY, ...)` to trigger an overall CommandError and nonzero exit code. And the output message code is duplicated from mpremote's main.py (interpolated with the results of the OSError -> CommandError behavior in #17090), in order to produce output that precisely matches the output that would be produced by that final CommandError.

Josverl:
> I have an alternative implementation 

indeed, it has it merits, and can stand on its own.
i see no need to include that PR in this one , I think it will only diffuse matters, and do not plan to pull it in.

> This doesn't implement the rm -r : behavior @dpgeorge suggested

I'll await Damien's comments.

> you can just run it on the mountpoint's path, 

mpremote does not (yet) CLI to expose these mountpoints to users without writing code to a new and therefore mostly unknown API. 
thus I see very clear merits.

tools/mpremote/tests/README.md
Josverl: verified tests pass across all ports for which I have hardware, including stm32

dpgeorge:
> I now find a desire for a `mpremote fs tree` command , let's land this one first.

Yes, I was also thinking that `tree` would be very useful.  That can be the next PR :smile:

dpgeorge:
> This doesn't implement the `rm -r :` behavior @dpgeorge suggested, but IMO it's just unnecessary to have a way of clearing a mountpoint without triggering an error output -- you can just run it on the mountpoint's path, and if you _need_ to suppress the stderr message or nonzero exit you just `2>/dev/null || true`.

Clearing out an entire mount point / filesystem is probably one of the main uses of `rm -r`, and having that always error out in some way is not very user friendly.  Similarly, asking users to redirect stderr is just going to add extra friction and require documentation and more support.  `mpremote` is supposed to be a very user friendly front-end to a device.

Since `mpremote` doesn't support wildcards you can't do `mpremote rm -r :\*` to clear out the current directory, so `mpremote rm -r :` seems like a simple way to implement that feature.  And it leaves `mpremote rm -r :/` to have behaviour like unix and give an error if it can't remove `/`.

docs/reference/mpremote.rst
```diff
@@ -238,9 +238,17 @@ The full list of supported commands are:
...
+  directory tree to be removed from the default path of the device (e.g. 
+  ``/flash`` or ``/``).
+  Attempt to remove a mounted ```vfs`` will result in a non-blocking warning. 
+  The mount's content will be removed, but the mount itself will remain.
+  Attempting to remove the root directory (``/``) will result in an error.
+  
+  All other commands implicitly assume the path is a remote path, but the ``:`` 
+  can be optionally used for clarity.
 
   So for example, ``mpremote fs cp main.py :main.py`` copies ``main.py`` from
```
dpgeorge, line 253: I think this paragraph should move up one paragraph, because it follows on from the `cp` path description.

docs/reference/mpremote.rst
```diff
@@ -238,9 +238,17 @@ The full list of supported commands are:
...
   path. Without a leading ``:`` means a local path. This is based on the
   convention used by the `Secure Copy Protocol (scp) client
-  <https://en.wikipedia.org/wiki/Secure_copy_protocol>`_. All other commands
-  implicitly assume the path is a remote path, but the ``:`` can be optionally
-  used for clarity.
+  <https://en.wikipedia.org/wiki/Secure_copy_protocol>`_. 
+  
+  The ``rm -r`` command accepts ``:`` to refer to the current remote path to allow a 
+  directory tree to be removed from the default path of the device (e.g. 
+  ``/flash`` or ``/``).
```
dpgeorge, line 245: I think it would be good to explicitly have `mpremote rm -r :` as an example here.

tools/mpremote/mpremote/commands.py
```diff
@@ -338,7 +359,7 @@ def do_filesystem(state, args):
                 if command == "cp":
                     print("{} {} {}".format(command, path, cp_dest))
                 else:
-                    print("{} :{}".format(command, path))
+                    print("{} {}:{}".format(command, "-r " if args.recursive else "", path))
```
dpgeorge, line 362: The `cp` command doesn't echo the `-r` above when recursive is used, so I don't think `rm` needs to either.  It'll be obvious from the verbose output that it's recursive.

tools/mpremote/mpremote/transport.py
```diff
@@ -25,6 +25,7 @@
 # THE SOFTWARE.
 
 import ast, hashlib, os, sys
+import errno
```
dpgeorge, line 28: Please put this import on the line above to match style in this file.

tools/mpremote/tests/.gitignore
```diff
@@ -0,0 +1,2 @@
+# test outputs
+*.out
```
dpgeorge, line 2: Please don't add this file.  If the directory is dirty one can use `git clean -f` to get rid of them.

tools/mpremote/tests/README.md
```diff
@@ -4,11 +4,25 @@ This directory contains a set of tests for `mpremote`.
...
+
 
 To run the tests do:
 
+    $ cd tools/mpremote/tests
     $ ./run-mpremote-tests.sh
 
+To run a single test do:
+
+    $ ./run-mpremote-tests.sh <test_name>
```
dpgeorge, line 23: I don't think this line is needed, just the one below it is enough.

tools/mpremote/tests/.gitignore
Josverl: ok, but at the risk of accidently adding .out files created during testing.

tools/mpremote/mpremote/transport.py
Josverl: ok

docs/reference/mpremote.rst
Josverl: Added 2 exaples, one to remove a reltive folder `libs`, and one to clear out the entire device.

tools/mpremote/mpremote/commands.py
Josverl: adjusted , including the .exp files

Josverl:
The failing CI test is unrelated.

dpgeorge (approved):
Tested on stm32 and rp2 boards.  Works well!

The implementation is also very clean.  Thanks @Josverl for doing multiple passes to get it right, and @AJMansfield for testing/review/feedback.

AJMansfield:
It's been a worthwhile effort getting this into a shape we can all be glad to see land!

Issue #16998, opened 2025-03-23
docs: network: ppp: API does not match


thokis:
### Documentation URL

https://docs.micropython.org/en/v1.24.0/library/network.PPP.html

### Description

Using version `1.24.1` on targets `ESP32_GENERIC` and `ESP32_GENREIC_S3` the documented API does not match `network.PPP` module on the devices.

![Image](https://github.com/user-attachments/assets/c430fd55-f2f6-4844-b52e-b28dbd7d988a)
- `network.PPP.active()` is missing from the documentation
![Image](https://github.com/user-attachments/assets/22dfa038-da31-4abf-9157-35d497aea1b0)
- `network.PPP.connect()` takes different params then documented. Instead of the keywords `security`, `user` and `key` the actual keywords are `authmode`, `username` and `password`.
![Image](https://github.com/user-attachments/assets/d8fadb8d-5639-4080-9106-85d3eecae1b1)
- `network.PPP.SEC_NONE`, `network.PPP.SEC_PAP`, `network.PPP.SEC_CHAP`, `network.PPP.disconnect()` and `network.PPP.poll()` do not exist.
![Image](https://github.com/user-attachments/assets/337323f2-f580-4e86-8150-47b5172f1425)
- `network.PPP.status()` returns nothing, even if the device is connected
![Image](https://github.com/user-attachments/assets/34179e1d-9246-49c5-9d9f-889f2c2b49f2)

The same is true for `ESP32_GENERIC_S3` target.

### Code of Conduct

Yes, I agree

thokis:
And maybe also worth mentioning is that on target `ESP32_GENERIC_S3` the `network.PPP.active(True)` and `network.PPP.connect()` calls are non-functional if the underlying `machine.UART` was initialized with keyword argument `timeout` set to anything. It seems that the `machine.UART` object must have non-blocking reads, else the PPPoS setup will not work. But it seems this is a hardware/target related problem because `ESP32_GENERIC` does not behave like this at all.

Tangerino:
I've been a long time trying to make PPP works in my setup, I've always blamed my modem but when I read about the `timeout` about, I just removed it when opening the UART and all is working now.
I never saw any mention about it  :(

robert-hh:
There is a difference between ESP32 PPP and the PPP/LWIP support for other ports. The the keywords `security`, `user` and `key` are used by the PPP/LWIP version.

dpgeorge:
[...]

PR #17008, opened 2025-03-25
ports/esp32: Updated network_ppp.c to match current documentation

thokis:
### Summary

This MR updates the `network.PPP` module of the `ESP32` port to match the current documentation. The `status()` method should behave like the common LwIP implementation and the `poll()` method will raise an RuntimeError indicating that this is not really supported on this port. Also did some minor resorting to match the common LwIP implementation.

This would solve https://github.com/micropython/micropython/issues/16998

### Testing

Tested on `ESP32_GENERIC` and `ESP32_GENERIC_S3` boards with `SIMCOM800L` and `SIMCOM7080G` modules. At least with `ESP32_GENERIC` the `PPP` connection is connected at least `18` hours.
Screenshot from 2025-03-25 20-35-16-obfuscated

### DIFF
--- a/ports/esp32/network_ppp.c
+++ b/ports/esp32/network_ppp.c
@@ -47,16 +47,24 @@
 
 #define PPP_CLOSE_TIMEOUT_MS (4000)
 
+typedef enum {
+    STATE_INACTIVE,
+    STATE_ACTIVE,
+    STATE_ERROR,
+    STATE_CONNECTING,
+    STATE_CONNECTED,
+} network_ppp_state_t;
+
 typedef struct _ppp_if_obj_t {
     mp_obj_base_t base;
-    bool active;
-    bool connected;
+    int error_code;
     volatile bool clean_close;
     ppp_pcb *pcb;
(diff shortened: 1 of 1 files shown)


DvdGiessen:
I worked on similar changes previous year and finally submitted them as a PR last week, see https://github.com/micropython/micropython/pull/17138. That fixes a few crashes I was having, and as a side-effect also does address the your linked issue about the documentation not matching. Linking it here since you might find it interesting.

Note these two PR's conflict. In the other PR I replaced parts of the ESP32 version with the extmod version. So for example it uses IRQ/poll() system from the extmod version for reading data from the UART and completely removes the task previously used by the ESP32 version. Whereas this PR one adds additional enhancements to the task management logic, and adds explicit warnings about the poll() method being unsupported.

dpgeorge:
Thanks @thokis for the contribution.  While this PR is a good improvement, I'll close it in favour of #17138, because the latter is more maintainable in the long run, matches the documentation, and is also backwards compatible.

PR #17010, opened 2025-03-25
A few MICROPY_PREVIEW_VERSION_2 cleanups.

andrewleech:
### Summary

I was doing some testing on rp2 and stm32 with `MICROPY_PREVIEW_VERSION_2` enabled and ran into a few basic needed updates. 

### Testing

Builds from rp2 and stm32 were run briefly with no issues noticed. 
I wasn't able to test the nina driver however the change should be very low risk.

I don't really know enough about the stack limit stuff in stm32 to be sure this replacement is equivalient, however it felt like it made sense to me?

### DIFF
--- a/ports/stm32/main.c
+++ b/ports/stm32/main.c
@@ -515,11 +515,8 @@ soft_reset:
     mp_thread_init();
     #endif
 
-    // Stack limit should be less than real stack size, so we have a chance
-    // to recover from limit hit.  (Limit is measured in bytes.)
-    // Note: stack control relies on main thread being initialised above
-    mp_stack_set_top(&_estack);
-    mp_stack_set_limit((char *)&_estack - (char *)&_sstack - 1024);
+    // Stack limit init.
+    mp_cstack_init_with_top(&_estack, (char *)&_estack - (char *)&_sstack);
 
     // GC init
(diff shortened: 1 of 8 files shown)


ports/stm32/main.c
```diff
@@ -508,11 +508,8 @@ void stm32_main(uint32_t reset_mode) {
     mp_thread_init();
     #endif
 
-    // Stack limit should be less than real stack size, so we have a chance
-    // to recover from limit hit.  (Limit is measured in bytes.)
-    // Note: stack control relies on main thread being initialised above
-    mp_stack_set_top(&_estack);
-    mp_stack_set_limit((char *)&_estack - (char *)&_sstack - 1024);
+    // Stack limit init.
+    mp_cstack_init_with_top(&_estack, (char *)&_estack - (char *)&_sstack);
```
dpgeorge, line 512: I'm pretty sure you'll need to add `#define MICROPY_STACK_CHECK_MARGIN (1024)` to `mpconfigport.h` to make this work the same as before.

andrewleech: Thanks, I've added that now.

dpgeorge (approved):

PR #17011, opened 2025-03-26
py/modmicropython: Expose repl_autocomplete as python function.

andrewleech:
### Summary

Exposes `mp_repl_autocomplete()` and `mp_hal_stdio_mode_raw()`/`mp_hal_stdio_mode_orig()` to Python as `micropython.repl_autocomplete()` and `micropython.stdio_mode_raw()` respectively.

`repl_autocomplete(line)` returns the completion suffix string, empty string for no match, or `None` when multiple candidates are printed. Gated on `MICROPY_HELPER_REPL` which is already enabled on most ports.

`stdio_mode_raw(enabled)` switches the terminal between raw and original mode. Gated behind a new `MICROPY_PY_MICROPYTHON_STDIO_RAW` config option, defaulting to off, enabled on unix.

Both are used by micropython/micropython-lib#1081 to bring aiorepl closer to feature parity with the native REPL — tab completion via `repl_autocomplete`, and proper terminal mode management via `stdio_mode_raw`.

### Testing

Unit tests added for both functions under `tests/micropython/`. A cmdline REPL test for `stdio_mode_raw` verifies actual terminal attribute changes via `termios`.

Tested on unix port.

### Generative AI

I used generative AI tools when creating this PR, but a human has checked the code and is responsible for the description above.

### DIFF
--- a/py/modmicropython.c
+++ b/py/modmicropython.c
@@ -31,6 +31,7 @@
 #include "py/runtime.h"
 #include "py/gc.h"
 #include "py/mphal.h"
+#include "py/repl.h"
 
 #if MICROPY_PY_MICROPYTHON
 
@@ -166,6 +167,27 @@ static mp_obj_t mp_micropython_schedule(mp_obj_t function, mp_obj_t arg) {
 static MP_DEFINE_CONST_FUN_OBJ_2(mp_micropython_schedule_obj, mp_micropython_schedule);
 #endif
 
+#if MICROPY_HELPER_REPL
+static mp_obj_t mp_micropython_repl_autocomplete(mp_obj_t cur_line) {
+    const char *compl_str = "";
+    size_t str_len;
+    const char *str = mp_obj_str_get_data(cur_line, &str_len);
+
+    size_t compl_len = mp_repl_autocomplete(str, str_len, &mp_plat_print, &compl_str);
+    if (compl_len == (size_t)-1) {
+        // Multiple completions, printed to stdout.
+        return mp_obj_new_str_via_qstr("", 0);
+    } else if (compl_len == 0) {
+        // No match.
+        return mp_const_none;
+    } else {
+        // Single completion or common prefix.
+        return mp_obj_new_str_via_qstr(compl_str, compl_len);
+    }
+}
+static MP_DEFINE_CONST_FUN_OBJ_1(mp_micropython_repl_autocomplete_obj, mp_micropython_repl_autocomplete);
+#endif
+
 static const mp_rom_map_elem_t mp_module_micropython_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_micropython) },
     { MP_ROM_QSTR(MP_QSTR_const), MP_ROM_PTR(&mp_identity_obj) },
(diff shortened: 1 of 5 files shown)


mattytrentini:
Adding autocomplete for aiorepl is _very_ desirable, and this is looking promising! Here are some of the tests I ran:

```python
>>>import micropython
>>>micropython.repl_autocomplete("impo")  # Should complete 'import'
'rt '
>>> class Foo:
...     def _bar():
...         pass
...     def alpha():
...         pass
>>> micropython.repl_autocomplete("f = Fo") # Should complete Foo
'o'
>>>f = Foo()
>>> micropython.repl_autocomplete("f.") # Should complete alpha (ignoring _bar)
'alpha'
```
Note that this is correct and consistent with MicroPython's built-in tab completion - but _differs_ in some cases to CPython, at least at v3.12. CPython will also supply parentheses, ie it would return `'o()'` to complete `f = Fo` and `'alpha('` for `f.`. That's not a problem with this PR, but should be resolved in `mp_repl_autocomplete`.

I also tested a module with private members and they were also correctly filtered out.

Looks good!

Josverl:
> but differs in some cases to CPython, at least at v3.12.

Perhaps that should _just_ be documented as such. I'd be fine with consistent behavior with the MicroPython repl.

py/repl.c
```diff
@@ -218,7 +218,7 @@ static void print_completions(const mp_print_t *print,
     for (qstr q = q_first; q <= q_last; ++q) {
         size_t d_len;
         const char *d_str = (const char *)qstr_data(q, &d_len);
-        if (s_len <= d_len && strncmp(s_start, d_str, s_len) == 0) {
+        if (s_len <= d_len && strncmp(s_start, d_str, s_len) == 0 && d_str[0] != '_') {
```
dpgeorge, line 221: Please break this out into a separate PR.  It'll need changes to the tests as well.

andrewleech: Split out to https://github.com/micropython/micropython/pull/17108 with a unit test (which might need some more updating to ensure the test is consistent / works across ports etc?)

py/modmicropython.c
```diff
@@ -166,6 +167,18 @@ static mp_obj_t mp_micropython_schedule(mp_obj_t function, mp_obj_t arg) {
...
 #endif
 
+#if MICROPY_HELPER_REPL
+static mp_obj_t mp_micropython_repl_autocomplete(mp_obj_t cur_line) {
+    const char *compl_str;
+    size_t str_len = 0;
+    const char *str = mp_obj_str_get_data(cur_line, &str_len);
+
+    ssize_t compl_len = mp_repl_autocomplete(str, str_len, &mp_plat_print, &compl_str);
+    return (compl_len <= 0) ? mp_const_none : mp_obj_new_str_via_qstr(compl_str, compl_len);
```
dpgeorge, line 177: If you make this `(compl_len < 0)` then the caller should be able to distinguish between the 3 distinct cases (no match, one match, many matches).

andrewleech:
Updated the branch with a few fixes and an additional feature:

Fixed a type mismatch in `repl_autocomplete` — was using `ssize_t` (not available on MSVC, hence all the Windows CI failures) and `<= 0` which incorrectly returned `None` for zero-length completions. Now uses `size_t` and compares against `(size_t)(-1)` matching how `readline.c` handles the same sentinel. The test also now uses `repr()` on outputs which avoids the trailing whitespace comparison issue that was likely causing the unix CI failures.

Also added `micropython.stdio_mode_raw(enabled)` in a second commit. This wraps the existing `mp_hal_stdio_mode_raw()` / `mp_hal_stdio_mode_orig()` HAL functions so Python code can control terminal mode directly — needed for aiorepl to properly manage terminal I/O without relying on the native REPL's implicit mode switching. Gated behind `MICROPY_PY_MICROPYTHON_STDIO_RAW` with a default of 0 in `mpconfig.h`, enabled on unix.

Added docs entries for both functions in `micropython.rst`.

PR #17013, opened 2025-03-26
samd/boards: Add two SparkFun SAMD21 boards.

robert-hh:
### Summary

Add support for the boards: 

- SparkFun SAMD21 Dev Breakout
- SparkFun  RedBoard Turbo

Both boards are SAMD21 based and actively sold by SparkFun.

### Testing

Built the board firmware and tested it on the board.

These are "just another" two SAMD boards. The SparkFun SAMD21 Dev Breakout works as well with the Generic SAMD21x18 firmware. Only the board Pin labels are then not supported.

### DIFF
--- a/docs/samd/pinout.rst
+++ b/docs/samd/pinout.rst
@@ -650,6 +650,124 @@ Adafruit ItsyBitsy M0 Express :ref:`samd21_pinout_table`.
 
 The board does not provide access to UART, I2C, SPI or DAC.
 
+SparkFun Redboard Turbo assignment table
+----------------------------------------
+
+=== ==== ============ ==== ==== ====== ====== ====== ======
+Pin GPIO Pin name     IRQ  ADC  Serial Serial TCC/TC TCC/TC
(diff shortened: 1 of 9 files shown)


dpgeorge (approved):
Thanks for this.  I tested that both boards build, but did not run the firmware (I don't have the hardware).

robert-hh:
Thank you for merging. I tested it with both boards.

sfe-SparkFro:
Thanks for adding these!

FYI these boards use the wrong capitalization (`Sparkfun` instead of `SparkFun`), meaning there's now 2 different "sparkfun" vendors on the MicroPython download page:

image

If a user selects one, they don't see the rest of the boards (there are currently 11 other SparkFun boards):

image

Should an issue be made for this?

robert-hh:
There was a PR #17258 by @malcolm-sparkfun, which was approved and merged. it changed "Sparkfun" into "SparkFun".

dpgeorge:
That PR did not fix these two boards, so we will need a PR to fix them.

robert-hh:
Created.

Issue #16644, opened 2025-01-25
ports/stm32: Add memory map with attributes for different MCU families.


iabdalkader:
### Description

There's a general DMA/Cache-management issue with the MicroPython's stm32 port; most drivers make assumptions about the memory (cachable or not DAM-acessible or not) which probably works fine for all (most?) of the default linker scripts, which happen to place everything in one memory region, but when buffers are moved around, drivers starts failing.

A few examples to demonstrate the issue (there are many more in sdcard, I2C etc..):

1. DAC should flush buffer before DMA call *if memory is cacheable* and fail if it's not DMA-accesible: https://github.com/micropython/micropython/blob/a4ab847688406bb31eb531e07155bf173a8785aa/ports/stm32/dac.c#L201

2. SPI regular transfer (fallback) path should be taken *if the memory is not accessible by DMA*: https://github.com/micropython/micropython/blob/a4ab847688406bb31eb531e07155bf173a8785aa/ports/stm32/spi.c#L588
For example:
```c
if (len == 1 || query_irq() == IRQ_STATE_DISABLED || !dma_is_accessible(src)) {
```
It should also Not do cache-management if memory is Not cacheable.

The problem will only get worse as we add more MCUs with more memory regions. To fix this issue, I propose we implement some sort of memory map per MCU family, with attributes per region: cacheable, DAM-accessible, TCM, external etc.. And a new generic API to query those attributes. It would be great if this mechanism is generic somehow, so it can be reused by future ports with more complex memory. I'd be very happy to work on this, but I need some guidance with the API.

### Implementation

I intend to implement this feature and would submit a Pull Request if desirable

### Code of Conduct

Yes, I agree

kwagyeman:
This is cool and would be very useful.

iabdalkader:
@dpgeorge Any thoughts?

Bortania:
@iabdalkader can you point out on the "patches to all of those driver" that you mentioned in your post.
Here we have an issue with DAC buffer on nucleo F7 board. Maybe the reason is memory cache.

iabdalkader:
> can you point out on the "patches to all of those driver" that you mentioned in your post.

Those patches are not applicable for all cases, otherwise I would have sent them as fixes. For example DAC uses DMA, in `timed` mode, so the memory needs to be cleaned/flushed before DMA starts. But the memory could also be TCM, in this case it should Not be flushed.
[...]

PR #17019, opened 2025-03-27
ports/stm32: Add support for additional GC blocks. 

iabdalkader:
### Summary

This patch adds support for defining additional GC blocks (for split heap) via linker scripts. While split GC on its own is very useful, there isn't a standard way to actually use it, so maybe this will be a step towards that.

For the Giga board, a few blocks in SDRAM/SRAM are also enabled as an example.

### Testing

I tested on the Giga, checked the blocks addresses an sizes passed to `gc_add` and `mem_free`. This change is enabled conditionally when split heap is enabled, so should not have any effect on other boards as none use split heaps. 

If linker scripts were processed through CPP, we could define these blocks in board config files and make the linker script section common. Since this is not the case, this table structure will have to be duplicated if other boards want to use this. I've tried to add a common linker script (`common_gc_table.ld`) but without CPP it's very hard to make this generic (the conditionals supported by ld script are very basic).

Also, the blocks are defined in a particular order so that faster SRAM blocks are, hopefully, searched first. We wouldn't have to rely on this, if we had this feature: https://github.com/micropython/micropython/issues/16644

### DIFF
--- a/ports/stm32/boards/ARDUINO_GIGA/mpconfigboard.h
+++ b/ports/stm32/boards/ARDUINO_GIGA/mpconfigboard.h
@@ -31,6 +31,7 @@ typedef unsigned int mp_uint_t;     // must be pointer size
 #define MICROPY_HW_ENABLE_MMCARD    (0)
 #define MICROPY_HW_ENTER_BOOTLOADER_VIA_RESET   (0)
 #define MICROPY_HW_TIM_IS_RESERVED(id) (id == 1)
+#define MICROPY_GC_SPLIT_HEAP       (1)
 
 // ROMFS config
 #define MICROPY_HW_ROMFS_ENABLE_EXTERNAL_QSPI       (1)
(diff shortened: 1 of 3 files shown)


Gadgetoid:
Short term this would have made a lot of sense for PSRAM on the RP2, too, though we don't have the luxury of per-board linker scripts yet.

And additionally the linked issue relates to the wider problems with PSRAM on the RP2, a platform which exposes DMA and PIO to the end user in ways that could quickly trip them up when a buffer by chance falls into PSRAM.

iabdalkader:
I hope we can find a way to standardize this, eventually, at least the C code in `main.c` could be common code, and ports could define the "gc table" in any way that works (via linker scripts, in board.c etc... However, for now I'm just trying to add more heap to the Giga as we have an app that's running out of memory and an unused 8MBytes SDRAM :)

> And additionally the linked issue relates to the wider problems with PSRAM on the RP2, a platform which exposes DMA and PIO to the end user in ways that could quickly trip them up when a buffer by chance falls into PSRAM.

That's another strong case for implementing the memory map feature.

iabdalkader:
Can we unblock this PR if the current solution is acceptable? We have a demo/example that doesn't work without this.

dpgeorge (approved):
Sorry for the late review.

This looks fine to me, it's a reasonable solution that gets things working.  In the future it can be refined/improved and generalised to other ports.

I built ARDUINO_GIGA firmware and inspected the elf to see that the `_gc_blocks_table_start` memory looked correct.

iabdalkader:
@sebromero

PR #8761, opened 2022-06-15
ports/rp2: Per-board linker scripts to avoid producing self-corrupting builds.

Gadgetoid:
Add per-board linker scripts to avoid silently overflowing flash.

Excludes region allocated to user filesystem from FLASH section.

Rationale explained in https://github.com/micropython/micropython/issues/8680

TLDR: Prevents the accidental generation of MicroPython builds which self-destruct on first run by overwriting themselves with the configured LittleFS filesystem.

I'm not super happy about copying memmap_mp.ld for every board, but I don't really see a better way to do this and preprocessing is non-trivial CMake spaghetti.

PR #17029, opened 2025-03-28
ports/rp2: Make linker script configurable.

Gadgetoid:
Add MICROPY_BOARD_LINKER_SCRIPT to specify a custom linker script for RP2 boards/variants.

This may, for example, include a PSRAM region so that C buffers or otherwise can be allocated into PSRAM.

### Summary

This is a much leaner version of #8761, but broadly attempts to accomplish some of the same things.

I use this feature (with some small modifications to the PSRAM code shown here: https://github.com/pimoroni/micropython/commit/79d7cc2ac97336847a6765d0ee23b2ef4e97af72) for building Presto with some non-critical C buffers moved to PSRAM so we can fit a large framebuffer in SRAM and use the remaining PSRAM for GC heap.

This change reflects #17028 and aims to bring some edge feature which don't cause too much immediate upset to the MicroPython codebase, but which are useful to those of us pushing weird and experimental things into shipping products 😆

### Testing

Currently a critical part of the Presto build process.

As with #17028 it seems reasonable to make linker scripts configurable, however:

1. They are *complicated* (they're composable but not really in a way that makes sense here)
2. There may be precious few uses for this particular feature
3. If the canonical linker script ever changes, then that change must be merged into any boards/variants

### DIFF
--- a/ports/rp2/CMakeLists.txt
+++ b/ports/rp2/CMakeLists.txt
@@ -588,12 +588,16 @@ endif()
 # todo this is a bit brittle, but we want to move a few source files into RAM (which requires
 #  a linker script modification) until we explicitly add  macro calls around the function
 #  defs to move them into RAM.
-if (PICO_ON_DEVICE AND NOT PICO_NO_FLASH AND NOT PICO_COPY_TO_RAM)
+if (NOT MICROPY_BOARD_LINKER_SCRIPT)
     if(PICO_RP2040)
-        pico_set_linker_script(${MICROPY_TARGET} ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2040.ld)
+        set(MICROPY_BOARD_LINKER_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2040.ld)
     elseif(PICO_RP2350)
-        pico_set_linker_script(${MICROPY_TARGET} ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2350.ld)
+        set(MICROPY_BOARD_LINKER_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2350.ld)
     endif()
+endif()
+
+if (PICO_ON_DEVICE AND NOT PICO_NO_FLASH AND NOT PICO_COPY_TO_RAM)
+    pico_set_linker_script(${MICROPY_TARGET} ${MICROPY_BOARD_LINKER_SCRIPT})
 endif()
 
 pico_add_extra_outputs(${MICROPY_TARGET})


Gadgetoid:
Since we've had variant functionality merged since, I think this probably represents a good first step toward #8761 and, specifically, configuring flash sizes via the linker script so that Filesystem/Flash overlaps can be caught at compile time.

The blocker with flash/app sizes was an incompatibility with `bi_decl` which I'm duplicating in below in case anyone can shed some light on this:

> 1. What can we do about `bi_decl` not being compatible with dynamic linker symbols? The suggestion to make the .ld file the source of truth seems reasonable, but is incompatible with Pico SDK's binary descriptor macros. Does anyone know if there's a reasonable way to work around this?  - the problem code is https://github.com/micropython/micropython/blob/668a7bd28a49980b239fd7666684885382526988/ports/rp2/rp2_flash.c#L61-L70

Gadgetoid:
The more I think about this the more I think it would be ideal to get back some of what was discussed/proposed in #8761 - I need to delve into `bi_decl` and see if I can figure out whether it's possible to inerop with linker symbols. Could have been fixed, or could be fixed.

ports/rp2/CMakeLists.txt
```diff
@@ -588,14 +588,18 @@ endif()
...
     if(PICO_RP2040)
-        pico_set_linker_script(${MICROPY_TARGET} ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2040.ld)
+        set(MICROPY_BOARD_LINKER_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2040.ld)
     elseif(PICO_RP2350)
-        pico_set_linker_script(${MICROPY_TARGET} ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2350.ld)
+        set(MICROPY_BOARD_LINKER_SCRIPT ${CMAKE_CURRENT_LIST_DIR}/memmap_mp_rp2350.ld)
     endif()
 endif()
 
+if (MICROPY_BOARD_LINKER_SCRIPT AND PICO_ON_DEVICE AND NOT PICO_NO_FLASH AND NOT PICO_COPY_TO_RAM)
```
dpgeorge: I don't think the check for `MICROPY_BOARD_LINKER_SCRIPT` is needed here, it should always be valid (the above if-block makes sure of that) and we'll get an error if there's something other than RP2040 or RP2350.

dpgeorge:
This looks fine to me, a very easy way to make the linker script configurable.

> The more I think about this the more I think it would be ideal to get back some of what was discussed/proposed in #8761

That PR does have some good bits in it.  It's just really complicated and touches lots of things and so requires lots of review and decision making :sweat_smile:

ports/rp2/CMakeLists.txt
Gadgetoid: Oof. Agreed!

dpgeorge (approved):

dpgeorge:
Rebased and merged in c16a4db151c0fac7db45c7c896741f707e9ad364

Issue #17027, opened 2025-03-28
Wifi Unknown Error 0x0101


robtinkers:
### Port, board and/or hardware

esp32

### MicroPython version

MicroPython v1.24.1 on 2024-11-29; Generic ESP32 module with ESP32

### Reproduction

```
import gc, machine, network, sys, time

from microdot import Microdot # https://raw.githubusercontent.com/miguelgrinberg/microdot/refs/heads/main/src/microdot/microdot.py

print("FREE:", gc.mem_free())

gc.collect()

DATA=bytearray(10_000)

print("FREE:", gc.mem_free())

try:
    
    ap_config = {"essid":"testing!", "password":"", "authmode":network.AUTH_OPEN}
    ap_if = network.WLAN(network.AP_IF)
    ap_if.active(True)
    ap_if.config(**ap_config)
    print("Active", repr(ap_if.ifconfig()))
    
    while True:
        print(time.time(), repr(ap_if.status("stations")))
        time.sleep(1)

except KeyboardInterrupt:
    pass
except Exception as e:
    sys.print_exception(e)
machine.reset()
```

### Expected behaviour

Not `Wifi Unknown Error 0x0101`

### Observed behaviour

```MPY: soft reboot
FREE: 117424
FREE: 144880
Traceback (most recent call last):
  File "<stdin>", line 16, in <module>
RuntimeError: Wifi Unknown Error 0x0101
ets Jun  8 2016 00:22:57

rst:0xc (SW_CPU_RESET),boot:0x13 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
mode:DIO, clock div:2
load:0x3fff0030,len:4892
ho 0 tail 12 room 4
load:0x40078000,len:14896
load:0x40080400,len:4
load:0x40080404,len:3372
entry 0x400805b0
MicroPython v1.24.1 on 2024-11-29; Generic ESP32 module with ESP32
Type "help()" for more information.
```

### Additional Information

Note that the bug isn't microdot-specific, and might simply be triggered by large amounts of code. I have had similar issues with other libraries which don't use any external modules (i.e. no asyncio, network, socket).

The DATA size that triggers the bug will vary with the exact imports and whether gc.collect() is called.

### Code of Conduct

Yes, I agree

agatti:
[...]

PR #17030, opened 2025-03-28
esp32/network_common: Raise a memory error on ESP_ERR_NO_MEM.

agatti:
### Summary

This PR changes the error handler for WiFi operations to recognise out of memory conditions reported by ESP-IDF functions, and report them as MemoryError exceptions rather than a generic "error 0x0101".

The error handler only provided a human-readable error description for WiFi-specific error codes (codes in the ESP_ERR_WIFI_BASE range), but WiFi functions are known to return other codes.  Since an out of memory condition has a specific Python equivalent, using `MemoryError` in that case is a bit more user-friendly in exchange for a handful of extra code bytes used by the firmware.

### Testing

The reproduction test case for #17027 was used to make sure an out of memory condition raised by the WiFi component was recognised as such.

This change causes a minimal increase of the firmware size, which I believe is an acceptable trade-off from the user's point of view.  After all, an out of memory error is not a "Wifi Unknown Error".

### DIFF
--- a/ports/esp32/network_common.c
+++ b/ports/esp32/network_common.c
@@ -77,6 +77,8 @@ MP_NORETURN void esp_exceptions_helper(esp_err_t e) {
             mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Wifi Would Block"));
         case ESP_ERR_WIFI_NOT_CONNECT:
             mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Wifi Not Connected"));
+        case ESP_ERR_NO_MEM:
+            mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("WiFi Out of Memory"));
         default:
             mp_raise_msg_varg(&mp_type_RuntimeError, MP_ERROR_TEXT("Wifi Unknown Error 0x%04x"), e);
     }


Josverl (approved):

ports/esp32/network_common.c
```diff
@@ -77,6 +77,8 @@ NORETURN void esp_exceptions_helper(esp_err_t e) {
             mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Wifi Would Block"));
         case ESP_ERR_WIFI_NOT_CONNECT:
             mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("Wifi Not Connected"));
+        case ESP_ERR_NO_MEM:
+            mp_raise_msg(&mp_type_MemoryError, MP_ERROR_TEXT("WiFi Out of Memory"));
```
dpgeorge, line 81: MemoryError is usually related to a Python heap memory error.  I think it's worth keeping that distinction, so we can more easily diagnose "no Python heap left" from "no IDF heap left".

So I suggest making this an OSError with the same message, "Wifi Out of Memory".  That's still a lot better than "Wifi Unknown Error", and allows MemoryError to specifically relate to the Python GC heap.

agatti: Fair point!  My take on this was that the average user may or may not be aware that there are actually two heaps; playing with WiFi is probably the second thing a user tries after blinking a LED :)  

Still, I also see the value in maintaining consistency with the rest of the errors in that switch case - I updated the PR accordingly.

dpgeorge (approved):

PR #17031, opened 2025-03-28
mpy_ld.py: Improve error for ARMv6m absolute relocations

jonnor:
Improved error message for building native modules using soft-float on ARMv6m, which is not supported.

### Testing

Built native modules for armv6m using soft-float, see that new error is raised


jonnor:
The one failing test does not seem to be related to these commits.

> 1 tests failed: stress_recurse

projectgus (approved):
Fix LGTM, thanks @jonnor!

Issue #17035, opened 2025-03-29
Panic when calling timer.value() on an uninitialised timer (ESP32)


robert-hh:
### Discussed in https://github.com/orgs/micropython/discussions/17033

<div type='discussions-op-text'>

<sup>Originally posted by **SergeiFranco** March 29, 2025</sup>
I have encountered a peculiar problem:
if I call value() on a uninitialised timer I get a panic:
Here are steps to reproduce:
```
t0 = Timer(0)
t0.value()
```

Results in:
```
A fatal error occurred. The crash dump printed below may be used to help
determine what caused it. If you are not already running the most recent
version of MicroPython, consider upgrading. New versions often fix bugs.

To learn more about how to debug and/or report this crash visit the wiki
page at: https://github.com/micropython/micropython/wiki/ESP32-debugging

MPY version : v1.24.1 on 2024-11-29
IDF version : v5.2.2
Machine     : Generic ESP32 module with ESP32

Guru Meditation Error: Core  1 panic'ed (LoadProhibited). Exception was unhandled.

Core  1 register dump:
PC      : 0x400d6790  PS      : 0x00060330  A0      : 0x800dd400  A1      : 0x3ffcf370  
A2      : 0x3ffd3890  A3      : 0x00000000  A4      : 0x00020002  A5      : 0x00000000  
A6      : 0x0000005c  A7      : 0x3ffbc984  A8      : 0x00000000  A9      : 0x00000000  
A10     : 0x00000003  A11     : 0x00060323  A12     : 0x00000001  A13     : 0x0000000d  
A14     : 0x3ffd6f40  A15     : 0x00000000  SAR     : 0x0000001a  EXCCAUSE: 0x0000001c  
EXCVADDR: 0x00000008  LBEG    : 0x4000c46c  LEND    : 0x4000c477  LCOUNT  : 0xffffffff  


Backtrace: 0x400d678d:0x3ffcf370 0x400dd3fd:0x3ffcf390 0x400e4935:0x3ffcf3b0 0x400e49fd:0x3ffcf3d0 0x40084d22:0x3ffcf3f0 0x400dd51f:0x3ffcf490 0x400e4935:0x3ffcf4e0 0x400e494a:0x3ffcf500 0x400f2763:0x3ffcf520 0x400f2a58:0x3ffcf5b0 0x400d6274:0x3ffcf5f0




ELF file SHA256: 4ac361b7d

Rebooting...
```

Note: works on any timer (not just 0).

The micro details are:
[...]

PR #17036, opened 2025-03-29
esp32/machine.timer.c: Fix timer.value() of an uninitialized timer.

robert-hh:
### Summary

Raises a value error in that case, which happens after a timer was created but not initialized or after calling timer.deinit(). Fixes issue #17033 / #17035.

### Testing

Tested with a ESP32_GENERIC before and after the change, for both a) the state after creating the timer but not initializing it, and b) after calling timer.deinit() on a previously operating timer. In both states, the change causes the error to be raised.

Testing with a generic ESP32 seems to be sufficient, since the change is not specific for a ESP32 variant.

Alternatively a neutral value like 0 or -1 could be returned instead of raising an error.

### DIFF
--- a/ports/esp32/machine_timer.c
+++ b/ports/esp32/machine_timer.c
@@ -246,6 +246,9 @@ static MP_DEFINE_CONST_FUN_OBJ_KW(machine_timer_init_obj, 1, machine_timer_init)
 
 static mp_obj_t machine_timer_value(mp_obj_t self_in) {
     machine_timer_obj_t *self = self_in;
+    if (self->handle == NULL) {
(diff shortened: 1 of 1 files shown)


jonnor:
Seems like a simple and reasonable fix. Exception makes sense to me over an error code, as this can never do anything useful.

I guess an alternative would be to use a more generic exception text, to avoid needing to store a specific error string just for this rare case. Not sure if there are any conventions for such cases?

dpgeorge (approved):
Tested with ESP32_GENERIC_S2 and IDF 5.4.1.  Works.

PR #17037, opened 2025-03-29
tools/mpremote/commands.py: Validate romfs image.

iabdalkader:
### Summary
Currently the tool allows writing invalid romfs image, with a bad header or images smaller than minimum size, and only checks the image extension.
This patch allows deploying romfs with any extension, and validates the image header and minimum size before writing.

### Testing

Before this patch, I tested deploying 1 byte images, or all zero images, and in both cases the tool deploys them as long as they have the correct extension. Deeper validation could be performed in both the tool and runtime, as I've encountered a runtime issue with a corrupted image (during development) at least once.

### DIFF
--- a/tools/mpremote/mpremote/commands.py
+++ b/tools/mpremote/mpremote/commands.py
@@ -9,7 +9,7 @@ import serial.tools.list_ports
 
 from .transport import TransportError, TransportExecError, stdout_write_bytes
 from .transport_serial import SerialTransport
-from .romfs import make_romfs
+from .romfs import make_romfs, VfsRomWriter
 
 
 class CommandError(Exception):
@@ -555,7 +555,7 @@ def _do_romfs_deploy(state, args):
     romfs_filename = args.path
 
     # Read in or create the ROMFS filesystem image.
-    if romfs_filename.endswith(".romfs"):
+    if os.path.isfile(romfs_filename) and romfs_filename.endswith((".img", ".romfs")):
         with open(romfs_filename, "rb") as f:
             romfs = f.read()
     else:
@@ -580,6 +580,11 @@ def _do_romfs_deploy(state, args):
     else:
         rom_size = transport.eval("len(dev)")
         print(f"ROMFS{rom_id} partition has size {rom_size} bytes")
+
+    # Check if ROMFS image is valid
+    if not romfs.startswith(VfsRomWriter.ROMFS_HEADER):
+        print("Invalid ROMFS image")
+        sys.exit(1)
 
     # Check if ROMFS filesystem image will fit in the target partition.
     if len(romfs) > rom_size:


tools/mpremote/mpremote/commands.py
```diff
@@ -581,6 +581,12 @@ def _do_romfs_deploy(state, args):
         rom_size = transport.eval("len(dev)")
         print(f"ROMFS{rom_id} partition has size {rom_size} bytes")
 
+    # Check if ROMFS image is valid
+    ROMFS_HEADER = b"\xd2\xcd\x31"
```
dpgeorge, line 585: I suggest using `romfs.VfsRomWriter.ROMFS_HEADER` instead.

tools/mpremote/mpremote/commands.py
```diff
@@ -581,6 +581,12 @@ def _do_romfs_deploy(state, args):
...
+    if len(romfs) < 4 or not romfs.startswith(ROMFS_HEADER):
```
dpgeorge, line 586: I'm pretty sure you don't need to check the length, just do the `startswith`.

tools/mpremote/mpremote/commands.py
```diff
@@ -555,7 +555,7 @@ def _do_romfs_deploy(state, args):
     romfs_filename = args.path
 
     # Read in or create the ROMFS filesystem image.
-    if romfs_filename.endswith(".romfs"):
+    if os.path.isfile(romfs_filename):
```
dpgeorge, line 558: My intention here was to eventually support other extensions to mean different things.  Eg `.json` could be a JSON manifest of files to turn into a ROMFS, and `.py` could be a standard MicroPython manifest file.  This change will no longer allow any of that, we'd need to find a different way to specify the type of the file.

iabdalkader: How about `.img` for ready to deploy images? It's seems standard for filesystem images? 

https://en.wikipedia.org/wiki/IMG_(file_format)

dpgeorge: What's wrong with `.romfs`?

(Or, could possibly support both `.romfs` and `.img` if it's really necessary.)

iabdalkader: I updated it, but when/if other file types are supported, wouldn't something like the following be needed anyway? So perhaps leave `isfile` ?

```Python
    if romfs.startswith(VfsRomWriter.ROMFS_HEADER):
       # romfs image ready to deploy
    elif extension == "json":
     # parse json
    elif extension == "py":
     # python
    else:
     raise(...) 
```

> What's wrong with .romfs?

Nothing really, I just call the images `romfs<part>` so they'll end up `romfs0.romfs` which is just not as nice as `romfs0.img`.

dpgeorge: > I updated it, but when/if other file types are supported, wouldn't something like the following be needed anyway? So perhaps leave `isfile` ?

Yes, it's OK to leave `isfile` in.  But also check the extension, maybe just by using `romfs_filename.endswith((".img", ".romfs"))`?

iabdalkader: Updated!

tools/mpremote/mpremote/commands.py
```diff
@@ -555,7 +556,7 @@ def _do_romfs_deploy(state, args):
     romfs_filename = args.path
 
     # Read in or create the ROMFS filesystem image.
-    if romfs_filename.endswith(".romfs"):
+    if romfs_filename.endswith((".img", ".romfs")):
```
dpgeorge, line 559: Oh, I thought you wanted to keep `isfile` (which I think is a good idea, because maybe a directory ends with `.romfs` for some reason) and so this would be:
```py
if os.path.isfile(romfs_filename) and romfs_filename.endswith((".img", ".romfs")):`
```

tools/mpremote/mpremote/commands.py
```diff
@@ -10,6 +10,7 @@
 from .transport import TransportError, TransportExecError, stdout_write_bytes
 from .transport_serial import SerialTransport
 from .romfs import make_romfs
+from .romfs import VfsRomWriter
```
dpgeorge, line 13: Please put this on one line: `from .romfs import VfsRomWriter, make_romfs`.

tools/mpremote/mpremote/commands.py
iabdalkader: Okay applied the above.

dpgeorge (approved):

Issue #4379, opened 2018-12-26
[ESP32] machine.ADC should be able to be initialised with pin numbers as well as machine.Pin


mattytrentini:
See the [Documentation for ESP32](https://forum.micropython.org/viewtopic.php?f=18&t=5347#p32747) forum post. It appears that for at least the PyBoard/STM32 and ESP8266 ports `machine.ADC` can be initialized by either a pin number (int) or from an instance of `machine.Pin`. The ESP32 port only allows an instance of `machine.Pin` for ADC init.

I suggest we allow the ESP32 port to also accept pin numbers:

``` python
>>> import machine
>>> a = machine.ADC(35) # Currently fails. Instead must use:
>>> a = machine.ADC(machine.Pin(35))
```

_Either_ of those last two lines should operate correctly.

(Incidentally I _prefer_ the ESP32 behaviour of only accepting `machine.Pin` but consistency is more important in this case.)

Note that there are efforts to overhaul the ADC interface (See #3943 and #4213). The proposed change above could be introduced in a shorter time frame, maybe even for v1.10 (see #4356) and ought to complement the upcoming overhaul.

pfalcon:
> (Incidentally I prefer the ESP32 behaviour of only accepting machine.Pin but consistency is more important in this case.)

So, consistency is accepting a Pin instance everywhere (beyond ADC too). Anything else is port-specific/legacy/undocumented behavior.

mattytrentini:
Just to clarify @pfalcon, are you agreeing with the suggested change?

If so I'll get on to making the change...

pfalcon:
I don't. For classes like UART, I2C, SPI, they *must* take a pin object, because any of them can be implemented in bitbanging manner on absolutely any Pin, e.g. you can do SPI over an I2C GPIO extender connected to a UART port.

Then, consistency calls to use Pin objects to refer to pins everywhere in `machine` API, even if freedom of which pins to use with other blocks is not so pervasive. (But there's nothing wrong to have an ADC extender over SPI, so it's just the same actually).

mattytrentini:
I'm really not too concerned whether we do or don't allow pins to be specified by integer for init but I do want _consistency between ports_. If you think it should only be possible to init ADC by passing in a Pin object then the following needs to be done:

1. Modify the ESP8266 port to accept Pin objects and not pin integers
    - Currently only `machine.ADC(0)` works, not `machine.ADC(machine.Pin(0))`
[...]

PR #17038, opened 2025-03-29
docs/esp32/quickref.rst Document that a Pin can be passed to UART constructor on the ESP32 ports.

bixb922:
### Summary

The ```UART()``` constructor has the  ```tx=``` and ```rx=``` parameters to specify a pin number. On the ESP32 port, an initialized ```machine.Pin()``` object can also be passed. This is useful to specify parameters such as ```mode=Pin.OPEN_DRAIN``` or ```drive=Pin.DRIVE_3```

I could not find the documentation for this neither on the ESP32 specific page (https://docs.micropython.org/en/latest/esp32/quickref.html) nor on the generic UART() page ((https://docs.micropython.org/en/latest/library/machine.UART.html)

### Testing

This is a documentation change only. However, I tested on a ESP32-S3 that ```mode=Pin.OPEN_DRAIN``` and ```drive=Pin.DRIVE_3``` alter the voltages at the UART pins accordingly:
```
txpin = Pin(tx_number, mode=Pin.OUT, drive=Pin.DRIVE_3)
rxpin = Pin(rx_number, mode=Pin.IN)
while True:
      UART( 1, baudrate=baudrate, tx=txpin, rx=rxpin)
      time.sleep(5)
      UART( 1, baudrate=baudrate, tx=txpin, rx=rxpin, invert=UART.INV_TX)
      time.sleep(5)
```

See also discussion #16959

### DIFF
--- a/docs/esp32/quickref.rst
+++ b/docs/esp32/quickref.rst
@@ -372,6 +372,13 @@ tx     1      10     17
 rx     3      9      16
 =====  =====  =====  =====
 
+You can pass a initialized Pin to ``tx`` and ``rx``::
+
+    from machine import UART, Pin
+
+    uart1 = UART(1, baudrate=9600, tx=Pin( 10, Pin.OPEN_DRAIN ), rx=Pin(9, Pin.IN, Pin.PULL_UP )  )
+
+
 PWM (pulse width modulation)
 ----------------------------
 


Josverl:
I think this should be available on any/all ports 
See https://github.com/micropython/micropython/issues/4379#issuecomment-2487247935

I would recommend reading that thread and checking the linked PRs to avoid duplication

projectgus:
> I think this should be available on any/all ports

I think at least every port which has `tx` and `rx` parameters for init supports this (or if a port doesn't support it, that's a bug). So yes, I think we could mention this in https://docs.micropython.org/en/latest/library/machine.UART.html#machine.UART.init docs instead.

@bixb922 would you consider updating this PR?

robert-hh:
> I think at least every port which has tx and rx parameters for init supports this (or if a port doesn't support it, that's a bug).

There are two aspects: a) Accept a Pin object for the UART/SPI/I2C ... interface, like rx, tx, mosi, .. . That should be possible for all ports. b) Keep the Pin configuration made when instantiating the Pin object when using the UART/SPI/I2C .. device. That is highly port specific and has to be analyzed & tested.

bixb922:
I agree, point (b) will most probably need testing. I did not find clear evidence even for the RP2040 if the UART can be set, for example, to open drain mode. 

I see two reasons why documenting this can be useful:

Use case 1: For all ports that allow specifying a rx/tx pin  number, using ``rx=Pin(pyb.Pin.board.X1)``` or ```rx=Pin("X1")``` can be useful as opposed to  ```rx=10```

Use case 2: for the ESP32, we know that you can set parameters such as ```OPEN_DRAIN``` , and that's also useful to know. I concur that most probably this capability is specific to ESP32 hardware, but that needs more reading/testing.

I think it would be best to add documentation for case 1 for all ports in https://docs.micropython.org/en/latest/library/machine.UART.html#machine.UART.init  and to document case 2 only for the ESP32 as proposed in this PR.

If you agree, I could add another documentation PR for use case 1, to be reviewed separately.

projectgus:
> I think it would be best to add documentation for case 1 for all ports in https://docs.micropython.org/en/latest/library/machine.UART.html#machine.UART.init and to document case 2 only for the ESP32 as proposed in this PR.

That seems reasonable to me. I think you can add them both in separate commits in this one PR.

bixb922:
Just for completeness here, I tested this code on a RP2040 with MicroPython 1.25.0:
```py
from machine import Pin, UART
import time
txpin = Pin(4, mode=Pin.OPEN_DRAIN, pull=Pin.PULL_DOWN)
rxpin = Pin(5)
txpin.value(1)
print(txpin, "set to 1")
time.sleep(5)
txpin.value(0)
print(txpin, "set to 0")
time.sleep(5)
# Initializing the UART will set txpin to a logical 1
txpin = Pin(4, mode=Pin.OPEN_DRAIN, pull=Pin.PULL_DOWN)
u=UART( 1, baudrate=9600, tx=txpin, rx=rxpin)
print("UART initialized", u)
time.sleep(100)
```
The voltage set by ```txpin.value(1)``` is near 0V, as it should. After the call to ```UART()```, the voltage returns to 3.3V, indicating that the open drain and pull down parameters are lost.

So point (b) of the previous post by @robert-hh valid: here is at least one case where the ```Pin``` parameters are not kept by ```UART()```. This behavior seems to be a hardware characteristic of the RP2040. Should MicroPython raise  ```ValueError``` when passing a ```Pin``` with  these parameters to ```UART()```? Given all the hardware constraints, it's probably not worth the memory cost. Perhaps the documentation should state something?

>That seems reasonable to me. I think you can add them both in separate commits in this one PR.

Ok, never done that, will give a try.

robert-hh:
> Should MicroPython raise ValueError when passing a Pin with these parameters to UART()?

No, not at all. It is more an unexpected behavior that the Pin settings are not changed to the state expected for e.g. UART, which is `Pin.OUT` for rx and `Pin.IN` for rx. There are other parameters which are not defined for UART or SPI like `drive`.
So since ESP32 has this optional feature, it can be documented, but only there.

robert-hh:
The changed statement in the docs are not clear. For any pin needed by an interface the general rule is, that this pin can be provided as either
- a Pin object (Pin(xyz), or Pin.board.xyz, Pin.cpu.xyz, if existing)
- anything that the Pin class accepts as argument, e.g. a number or a string (e.g. rx=4, sda="I2C_SDA"). But that is highly port-specific. Some ports or boards do not accept a string, some ports do not accept a number. The string type arguments are those defined in a board's pins.csv file, and are as well listed in the Pin.board class.

The latter may not apply to all ports. So that has to be verified.

Besides that the ESP32's behavior is special, to keep the configuration of  a Pin if a configured Pin object is supplied to UART or SPI. I did not test that yet for I2C or I2S. And this behavior does NOT apply to other ports.

So: do not change `docs/library/machine.UART.rst`. Statements there have to be valid for all ports. You can change the quickref for ESP32 explaining the special option.

bixb922:
Just for completeness, the the pull up for the rx pin is enabled by UART(), here is a test:
```py
# adcpin, rxpin and outpin must be connected together.

from machine import Pin, UART, ADC
from time import sleep_ms

def toggle(expected_low, expected_high):
    for val, expected in [(0,expected_low), (1,expected_high)]:
        outpin.value(val)
        v = adcpin.read_uv()/1_000_000
        ok = 'ok' if abs(expected-v)<0.3 else '???unexpected'
        print(f"    out {val} voltage={v:.3f} V, expected {expected}, {ok}")

# Reset rxpin to known state
rxpin = Pin(9, mode=Pin.IN, pull=None)
# Set output pin to OPEN_DRAIN
# since if it is in push/pull mode, no effect is visible in this test.
# Add pull down, to avoid open drain connected to a high impedance input.
outpin = Pin(16, mode=Pin.OPEN_DRAIN, pull=Pin.PULL_DOWN)
# adcpin to measure voltage (approximately)
adcpin = ADC( Pin(8), atten=ADC.ATTN_11DB )   

print("No UART, output pin with open drain+pull_down")
# rx pin is high impedance, open drain + pull down means 0 volts
toggle(expected_high=0, expected_low=0)

u=UART( 1, baudrate=9600, tx=Pin(10), rx=Pin(9))
print("Output pin with open drain+pull down, rx pin with pull set by UART()")
# The rx pin now provides the pull up and outpin provides a pull down
# dividing the voltage by 2 (not a very useful configuration, but
# shows that rx now has a pull up)
toggle(expected_high=1.5, expected_low=0)

print("Output Pin with open drain+no pull, rx with pull up set by UART()")
outpin = Pin(16, mode=Pin.OPEN_DRAIN, pull=None)
# Open drain with a pull up provided by rx pin
toggle(expected_high=3.3, expected_low=0)

# Output
# No UART, output pin with open drain+pull_down
#    out 0 voltage=0.000 V, expected 0, ok
#    out 1 voltage=0.000 V, expected 0, ok 
# Output pin with open drain+pull down, rx pin with pull set by UART()
#    out 0 voltage=0.000 V, expected 0, ok
#    out 1 voltage=1.474 V, expected 1.5, ok
# Output Pin with open drain+no pull, rx with pull up set by UART()
#    out 0 voltage=0.000 V, expected 0, ok
#    out 1 voltage=3.126 V, expected 3.3, ok
```

All this is really a behavior of ESP-IDF. MicroPython extracts the pin numbers and calls the ESP-IDF function ```uart_set_pin()```. Perhaps it would be better to write a "show and tell" in the discussions rather than putting this in the documentation.

PR #17044, opened 2025-03-30
ports/webassembly/proxy_c.c: Allow python `bytes()` to be proxied to/from JS.

bwhitman:
### Summary

This allows python `bytes()` objects to be passed to JS. Example:

```js
 function set_audio_samples(samples) {
     var res_ptr = amy_module._malloc(2 * 256 * 2); // 2 channels, 256 frames, int16s
     amy_module.HEAPU8.set(samples, res_ptr);
     amy_module.ccall('amy_set_input_buffer', null, ["number"], [res_ptr]);
     amy_module._free(res_ptr);
 }
...
 await mp.registerJsModule('amy_set_input_buffer', set_audio_samples);
```

In MP:
```python
buffer = bytes(1024)
# do something to fill in the bytes...
amy_set_input_buffer(buffer)
```

It also works the other way (JS returning bytes() to MP):

```js
 function get_audio_samples() {
     var res_ptr = amy_module._malloc(2 * 256 * 2); // 2 channels, 256 frames, int16s
     amy_module.ccall('amy_get_input_buffer', null, ["number"], [res_ptr]);
     var view = amy_module.HEAPU8.subarray(res_ptr , res_ptr + 1024);
     amy_module._free(res_ptr);
     return view
 }
 await mp.registerJsModule('amy_get_input_buffer', get_audio_samples);
```

```python
>>> b = amy_get_input_buffer()
>>> type(t)
<class 'bytes'>
```

Previously, we would have to manually construct a `js.UInt8Array.new(1024)` and fill it in one by one.

### Testing

Tested on webassembly port.

### DIFF
--- a/ports/webassembly/proxy_c.c
+++ b/ports/webassembly/proxy_c.c
@@ -62,6 +62,7 @@ enum {
     PROXY_KIND_JS_STRING = 5,
     PROXY_KIND_JS_OBJECT = 6,
     PROXY_KIND_JS_PYPROXY = 7,
+    PROXY_KIND_JS_BYTES = 8,
 };
 
 MP_DEFINE_CONST_OBJ_TYPE(
@@ -171,6 +172,10 @@ mp_obj_t proxy_convert_js_to_mp_obj_cside(uint32_t *value) {
         mp_obj_t s = mp_obj_new_str((void *)value[2], value[1]);
         free((void *)value[2]);
         return s;
+    } else if (value[0] == PROXY_KIND_JS_BYTES) {
+        mp_obj_t s = mp_obj_new_bytes((void*) value[2], value[1]);
+        free((void*)value[2]);
+        return s;
     } else if (value[0] == PROXY_KIND_JS_PYPROXY) {
         return proxy_c_get_obj(value[1]);
     } else {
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the contribution, this is definitely a useful feature and makes it much more efficient to pass bytes between Python and JS.

But, because `Uint8Array` is a mutable object, it no longer allows you to share such an object between Python and JS (instead this PR would make a copy when passing it across the divide).

At the moment you can do something like this:
```js
const mp = await (await import(process.argv[2])).loadMicroPython();

globalThis.array = new Uint8Array(4);
globalThis.array[0] = 1;
console.log(globalThis.array);

mp.runPython(`
import js
print(js.array[0])  # access JS Uint8Array
js.array[0] = 2  # modify JS Uint8Array
`);

console.log(globalThis.array);
```
Running this, it outputs:
```
$ node test.mjs `pwd`/build-pyscript/micropython.mjs
Uint8Array(4) [ 1, 0, 0, 0 ]
1
Uint8Array(4) [ 2, 0, 0, 0 ]
```
I don't think such sharing of mutable Uint8Array will be possible with this PR.

-------

Maybe instead it could be part of `jsffi.to_js()` where the user explicitly converts the value?

dpgeorge:
Closing due to inactivity.  Feel free to reopen with a response to the above comment.

PR #17055, opened 2025-03-31
port esp32 fix for esp-idf v5.5

electronicsguy:
The watchdog timer reconfiguration function has been modified. The original function was never meant to be used publicly (https://github.com/espressif/esp-idf/issues/15687)

This CL fixes the build error.

### Summary

port esp32 build fails with the linker error 'undefined reference' (see log below) with esp-idf v5.5

### Testing

Setup esp-idf v5.5
clone micropython from: https://github.com/micropython/micropython
cd <MICROPYTHON_GIT_ROOT>/ports/esp32
get_idf
make submodules
make

### DIFF
--- a/ports/esp32/panichandler.c
+++ b/ports/esp32/panichandler.c
@@ -42,11 +42,11 @@
 #endif
 
 void __real_esp_panic_handler(void *);
-void esp_panic_handler_reconfigure_wdts(uint32_t timeout_ms);
+void esp_panic_handler_enable_rtc_wdt(uint32_t timeout_ms);
 void panic_print_str(const char *str);
 
 void MICROPY_WRAP_PANICHANDLER_FUN(__wrap_esp_panic_handler)(void *info) {
-    esp_panic_handler_reconfigure_wdts(1000);
+    esp_panic_handler_enable_rtc_wdt(1000);
 
     const static char *msg = MICROPY_WRAP_PANICHANDLER_STR(
         "\r\n"


ports/esp32/panichandler.c
```diff
@@ -42,11 +42,11 @@
 #endif
 
 void __real_esp_panic_handler(void *);
-void esp_panic_handler_reconfigure_wdts(uint32_t timeout_ms);
+void esp_panic_handler_enable_rtc_wdt(uint32_t timeout_ms);
 void panic_print_str(const char *str);
 
 void MICROPY_WRAP_PANICHANDLER_FUN(__wrap_esp_panic_handler)(void *info) {
-    esp_panic_handler_reconfigure_wdts(1000);
+    esp_panic_handler_enable_rtc_wdt(1000);
```
sudeep-mohanty, line 49: Hi @electronicsguy,
Just curious - May I know the intention behind this change in this PR? IIUC, it seems like you’re trying to prevent watchdog timers from firing while the panic handler banner is printing.

If that’s the case, just a heads-up: the API `esp_panic_handler_enable_rtc_wdt()` might not do what you are expecting. This API will enable the RTC WDT additionally. It does not alter the state of any existing watchdogs that might already be running at this point, so it won't stop them from timing out.

You might want to check out [esp_panic_handler_feed_wdts()](https://github.com/espressif/esp-idf/blob/5c51472e82a58098dda8d40a1c4f250c374fc900/components/esp_system/panic.c#L215) API instead. That one explicitly feeds the watchdogs to keep them from firing.

Hope this helps!

PS: This is in reference to the esp-idf issue - https://github.com/espressif/esp-idf/issues/15687#issuecomment-2764970905

cnadler86: I think the right function would be `esp_panic_handler_feed_wdts`, or?

cnadler86:
This PR would also fix idf 5.4.1, BTW.

ports/esp32/panichandler.c
agatti: > I also think the right function would be `esp_panic_handler_feed_wdts()`.

I believe neither of those functions are available in 5.3 anyway :|.

(`esp_panic_handler_feed_wdts` in 5.3 is referenced by some ESP32C3 bluetooth code but as an extern when built with a particular configuration option, and there's no obvious implementation of that function anywhere else, go figure)

cnadler86: They were introduced in the latest patch releases of the IDF, AFAIK. To solve this an version specific fix will be needed, i think.

agatti: I know - the point is that right now MPy supports ESP-IDF 5.2, 5.2.2, 5.3, and 5.4.  Usually MPy supports current and two previous versions of ESP-IDF, in this case you'd have to move the support window to cover 5.3, 5.4, and 5.5, but 5.3 won't build with this.

Taking the previous assumption as valid then this PR needs to be changed regardless whether `esp_panic_handler_feed_wdts` or `esp_panic_handler_enable_rtc_wdt` is used in the first place to build with all the abovemetioned versions :)

cnadler86: Best would be if espressif provides a backwards compatible wrapper function for the new api, but don't know if they would do that.

And yes, this PR should be reworked I'm any case.

projectgus: > I believe neither of those functions are available in 5.3 anyway :|.

If there's no API that works on both versions then add a guard of the form 

```c
#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 5, 0)
...
#else
...
#endif
``3

... to avoid repeating everything twice it might be better to make a single static function that calls the API (which has the two implementations) and then call that from the two places.

> Best would be if espressif provides a backwards compatible wrapper function for the new api, but don't know if they would do that.

These aren't public APIs (as evidenced by having to declare them inline in the .c file), so I don't think we're likely to get a wrapper for them. I guess this custom panic handler use case isn't exactly supported.

DvdGiessen:
> This PR would also fix idf 5.4.1, BTW.

This doesn't affect IDF v5.4.1 as the panic handler WDT change is not in there yet and the old function still exists, see https://github.com/espressif/esp-idf/blob/v5.4.1/components/esp_system/panic.c#L185. I've been building against v5.4.1 since release without need for a patch.

electronicsguy:
Hi all,
    I wrote the quick PR since all I wanted was to make the build possible with the latest version of ESP_IDF. I had mistakenly downloaded v5.5 (which isn't released as yet), and Micropython wouldn't even build for me, so I was trying to figure out how to fix it. Not knowing exactly how linker wrappers work, micropython build intracacies, plus the complex ESP_IDF build system, I just made a quick change which allowed me to build and flash micropython on my ESP32S3 board. 

The ESP_IDF guys replied that the function that the custom wrapper wraps, was not intended for public use. 

Thank you pointing out about esp_panic_handler_feed_wdts().

PR #17056, opened 2025-03-31
Docs: Minor amendments to the documentation.

robert-hh:
### Summary

A few documentation updates aiming to clarify topics which confused users.

- Amend the documentation of time.mktime() by showing the argument and refer to epoch instead of a fixed
  date. The note about epoch lists the ports using the POSIX epoch.
- ESP32: Mention the different timer counts show the number of timers for each ESP32 variant.
- ESP32: Mention the use of Timer(0) by UART.IRQ_RXIDLE.

### Testing

Build the HTML version of the documentation.


dpgeorge (approved):
Thanks, these are good little improvements.

(I added alif to the list of ports that use 1970 Epoch.)

PR #17060, opened 2025-04-01
tools/gen-cpydiff: Fail CPython difference generation if output matches.

projectgus:
### Summary

MicroPython uses a set of automated test cases (`tests/cpydiff`) to generate documentation of differences between CPython and MicroPython. If a MicroPython improvement made something CPython compatible, the generation tool would print a note but would keep running as normal including generating the "difference" (with the docs containing identical generated output for CPython vs MicroPython).

This PR adjusts the generation tool to fail if any differences are non-existent (which should also fail CI), and also removes one fixed difference (`str.endswith()` was made compatible as of #16812 :tada:).

### Testing

* Ran the docs build and generation tool locally, verified the new failure due to a non-difference and non-failure after removing.

* We could make this behaviour configurable in `gen-cpydiff.py`, but this tool is only ever run to generate this output for the docs and almost all of the other behaviour is hard-coded for this purpose.

### DIFF
--- a/tests/cpydiff/types_str_endswith.py
+++ b/tests/cpydiff/types_str_endswith.py
@@ -1,8 +1,0 @@
-"""
-categories: Types,str
-description: Start/end indices such as str.endswith(s, start) not implemented
-cause: Unknown
-workaround: Unknown
-"""
-
-print("abc".endswith("c", 1))
--- a/tools/gen-cpydiff.py
+++ b/tools/gen-cpydiff.py
@@ -69,7 +69,6 @@ Output = namedtuple(
         "code",
         "output_cpy",
         "output_upy",
-        "status",
     ],
 )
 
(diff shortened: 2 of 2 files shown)


AJMansfield:
How will this affect port-specific differences, e.g. [`modules_time_time_returns_int.py`](https://github.com/micropython/micropython/pull/16984/files/682808f7c1711b2ed412911db12318e380b89e7c#diff-07e6535192956cd4b610a6c87e7c480b360071bcdc0fa078f5b7e654fc2b2998) from #16984? This probably needs a "SKIP"-like mechanism, so platform-dependent cpydiffs for issues not present in unix don't create spurious CI failures.

projectgus:
> How will this affect port-specific differences,

Thanks for pointing this out, @AJMansfield. I've left a reply on the linked PR: https://github.com/micropython/micropython/pull/16984/files#r2008260448

dpgeorge (approved):
Thanks, this is a good improvement!

Issue #10895, opened 2023-03-01
core: Crash with  an example code with walrus use,  on V1.19 ?


robert-hh:
### Discussed in https://github.com/orgs/micropython/discussions/10879

<div type='discussions-op-text'>

<sup>Originally posted by **fca1** February 28, 2023</sup>
Hi All,!

I've encourated a strange behavior with micropython 1.19 (crash and reset), my target is an ESP32 with spiram. 
The piece of code to attempt to reproduce that, is very tiny. This code has been tested with cpython, without any problems. 

Please, if someone could confirm if there is really a problem ? Thanks a lot in advance, 
    </div>
```py

def for_test_purpose():
    lst= [ 0,1]
    v=0
    x=1  # crash if  this line is present. 
    print([v := i for i in lst if i > v])


if __name__ == "__main__":
    for_test_purpose()
```

dpgeorge:
Thanks for the report.  I can confirm the crash on the unix port with latest master.

It's a nasty bug, it's not computing the closure of `v` correctly in the list comprehension.

dpgeorge:
See #10902 for a fix.

dpgeorge:
Fixed by 7c1584aef10f80aabfeba242892ec5f85c57bead

PR #17061, opened 2025-04-01
doc: Update CPython difference for assign expression in nested comprehensions.

projectgus:
### Summary

Since 7c1584aef10f80aabfeba242892ec5f85c57bead MicroPython matches CPython in most cases of assigning to a comprehension variable, apart from in nested comprehensions. Both [generated examples in the current version of the docs](https://docs.micropython.org/en/latest/genrst/syntax.html#micropython-allows-using-to-assign-to-the-variable-of-a-comprehension-cpython-raises-a-syntaxerror) are SyntaxErrors with different error text.

Updates the test case to only show the remaining difference, updates the recommendation to a more blanket recommendation to not do this.

Investigating this was my original reason to look into #17060, and I was going to update `gen-cypdiff.py` to also fail if a test case raises an exception on both MicroPython and CPython. However some of the "exceptions" test cases do both raise an exception (from which the interpreter prints different output), and there's no easy way to rewrite them.

### DIFF
--- a/tests/cpydiff/syntax_assign_expr.py
+++ b/tests/cpydiff/syntax_assign_expr.py
@@ -1,8 +1,8 @@
 """
 categories: Syntax,Operators
-description: MicroPython allows using := to assign to the variable of a comprehension, CPython raises a SyntaxError.
-cause: MicroPython is optimised for code size and doesn't check this case.
-workaround: Do not rely on this behaviour if writing CPython compatible code.
+description: MicroPython allows := to assign to the iteration variable in nested comprehensions, CPython does not.
+cause: MicroPython is optimised for code size. Although it is a syntax error to assign to the iteration variable in a standard comprehension (same as CPython), it doesn't check if an inner nested comprehension assigns to the iteration variable of the outer comprehension.
+workaround: Do not use := to assign to the iteration variable of a comprehension.
 """
 
-print([i := -1 for i in range(4)])
+print([[(j := i) for i in range(2)] for j in range(2)])


dpgeorge:
It looks like the commit that changed this behaviour was 7c1584aef10f80aabfeba242892ec5f85c57bead ; see also #10895.

And that commit tells us that there are still some cases that work in MP but not CPy, eg:
```py
print([[(j := i) for i in range(2)] for j in range(2)])
```
So maybe this cpydiff should be updated to use the above example?

projectgus:
> So maybe this cpydiff should be updated to use the above example?

Thanks for the context! I've updated the PR according to my understanding of the current behaviour, please check.

tests/cpydiff/syntax_assign_expr.py
```diff
@@ -1,8 +1,8 @@
 """
 categories: Syntax,Operators
-description: MicroPython allows using := to assign to the variable of a comprehension, CPython raises a SyntaxError.
-cause: MicroPython is optimised for code size and doesn't check this case.
-workaround: Do not rely on this behaviour if writing CPython compatible code.
+description: MicroPython allows := to assign to the iteration variable of a nested comprehension, CPython does not.
```
dpgeorge, line 3: This sounds like it allows assigning to the *inner* comprehension variable, whereas it's the outer one that it wrongly allows.

Suggest "of a nested comprehension" -> "in nested comprehensions", which makes it a bit more of a general statement.  The full description which much clearer is given on the next line in the "cause".

projectgus: Updated!

dpgeorge (approved):

PR #17067, opened 2025-04-02
stm32: add support to PYBD_SF6 for boards with larger flash

dpgeorge:
### Summary

There are some newer PYBD_SF6 being produced (finally!) which have a larger flash, namely two of 8MiB (instead of the older ones with two of 2MiB).

This PR adds support for these boards.  The idea is to have the same PYBD_SF6 firmware run on both old and new boards.  That m(eans autodetecting the flash at start-up and configuring all the relevant SPI/QSPI parameters, including for ROMFS and mboot.

I tried many ways of doing this, some more general and others more board-specific.  The more general approaches touched a lot more files than this PR which is less general and has a lot of the auto-detection logic in the PYBD_SF6 board directory.

Eventually the logic here could be generalised so it can work with other stm32 boards and even ports, but for now it's mostly contained to PYBD_SF6.

### Testing

Tested both mboot and the main firmware on original PYBD_SF6 hardware, and newer hardware with the larger flash.

I added some info to `machine.info()` to print out the detected flash size, which helps with debugging.

As discussed above, there are many ways to achieve this goal, some more/less general than others.  I think this PR strikes a good balance there and doesn't impact existing boards that don't need this auto-detection feature (such as the stm32 Arduino boards which do use the QSPI interface; a more general solution would need to modify those boards quite a bit).

### DIFF
--- a/ports/stm32/boards/PYBD_SF6/mpconfigboard.h
+++ b/ports/stm32/boards/PYBD_SF6/mpconfigboard.h
@@ -69,3 +69,53 @@
 #define MICROPY_HW_ETH_RMII_TX_EN   (pyb_pin_W8)
 #define MICROPY_HW_ETH_RMII_TXD0    (pyb_pin_W45)
 #define MICROPY_HW_ETH_RMII_TXD1    (pyb_pin_W49)
+
+// The below code reconfigures SPI flash for dynamic size detection.
+
+#undef MICROPY_BOARD_EARLY_INIT
+#undef MICROPY_HW_SPIFLASH_SIZE_BITS
+#undef MICROPY_HW_QSPIFLASH_SIZE_BITS_LOG2
+#undef MBOOT_SPIFLASH_BYTE_SIZE
+#undef MBOOT_SPIFLASH_LAYOUT
+#undef MBOOT_SPIFLASH_ERASE_BLOCKS_PER_PAGE
+#undef MBOOT_SPIFLASH2_BYTE_SIZE
+#undef MBOOT_SPIFLASH2_LAYOUT
+#undef MBOOT_SPIFLASH2_ERASE_BLOCKS_PER_PAGE
+
+// These are convenience macros to refer to the SPI flash chip parameters for the external SPI flash.
+#define MICROPY_BOARD_SPIFLASH_CHIP_PARAMS0 (spi_bdev.spiflash.chip_params) // SPI flash #1, R/W storage
+#define MICROPY_BOARD_SPIFLASH_CHIP_PARAMS1 (spi_bdev2.spiflash.chip_params) // SPI flash #2, memory mapped
+
+// Early init is needed to initialise the default SPI flash chip parameters.
+#define MICROPY_BOARD_EARLY_INIT    board_early_init_sf6
+
+// Enable dynamic detection of SPI flash.
+#define MICROPY_HW_SPIFLASH_CHIP_PARAMS (1)
+#define MICROPY_HW_SPIFLASH_DETECT_DEVICE (1)
+
+// Settings for SPI flash #1.
+#define MICROPY_HW_SPIFLASH_SIZE_BITS (1 << (MICROPY_BOARD_SPIFLASH_CHIP_PARAMS0->memory_size_bytes_log2 + 3))
+
+// Settings for SPI flash #2.
+#define MICROPY_HW_QSPI_PRESCALER   (MICROPY_BOARD_SPIFLASH_CHIP_PARAMS1->qspi_prescaler)
+#define MICROPY_HW_QSPIFLASH_SIZE_BITS_LOG2 (MICROPY_BOARD_SPIFLASH_CHIP_PARAMS1->memory_size_bytes_log2 + 3)
+
+// ROMFS partition 0 is dynamically sized (SPI flash #2).
+#define MICROPY_HW_ROMFS_PART0_START (uintptr_t)(&_micropy_hw_romfs_part0_start)
+#define MICROPY_HW_ROMFS_PART0_SIZE (1 << MICROPY_BOARD_SPIFLASH_CHIP_PARAMS1->memory_size_bytes_log2)
+
+// Mboot SPI flash #1 configuration.
+#define MBOOT_SPIFLASH_LAYOUT_DYNAMIC_MAX_LEN (20)
+#define MBOOT_SPIFLASH_LAYOUT       (MICROPY_BOARD_SPIFLASH_CHIP_PARAMS0->memory_size_bytes_log2 == 21 ? "/0x80000000/512*4Kg" : "/0x80000000/2048*4Kg")
+#define MBOOT_SPIFLASH_BYTE_SIZE    (1 << MICROPY_BOARD_SPIFLASH_CHIP_PARAMS0->memory_size_bytes_log2)
+#define MBOOT_SPIFLASH_ERASE_BLOCKS_PER_PAGE (1)
+
+// Mboot SPI flash #2 configuration.
+#define MBOOT_SPIFLASH2_LAYOUT_DYNAMIC_MAX_LEN (20)
+#define MBOOT_SPIFLASH2_LAYOUT      (MICROPY_BOARD_SPIFLASH_CHIP_PARAMS1->memory_size_bytes_log2 == 21 ? "/0x90000000/512*4Kg" : "/0x90000000/2048*4Kg")
+#define MBOOT_SPIFLASH2_BYTE_SIZE   (1 << MICROPY_BOARD_SPIFLASH_CHIP_PARAMS1->memory_size_bytes_log2)
+#define MBOOT_SPIFLASH2_ERASE_BLOCKS_PER_PAGE (1)
+
+extern unsigned char _micropy_hw_romfs_part0_start;
+
+void board_early_init_sf6(void);
--- a/ports/stm32/qspi.c
+++ b/ports/stm32/qspi.c
@@ -62,6 +62,8 @@
 #define QSPI_ADSIZE 2
 #endif
 
+static uint8_t qspi_num_dummy;
+
 static inline void qspi_mpu_disable_all(void) {
     // Configure MPU to disable access to entire QSPI region, to prevent CPU
     // speculative execution from accessing this region and modifying QSPI registers.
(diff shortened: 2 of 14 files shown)


dpgeorge:
The changes here could do with a few more comments :smile:

iabdalkader:
> the capacity of the flash is not available in the SFDP mandatory section.

I guess this is not a valid alternative then.

ports/stm32/boards/PYBD_SF6/mpconfigboard.h
```diff
@@ -69,3 +71,50 @@
...
+#undef MBOOT_SPIFLASH_LAYOUT
+#undef MBOOT_SPIFLASH_ERASE_BLOCKS_PER_PAGE
+#undef MBOOT_SPIFLASH2_BYTE_SIZE
+#undef MBOOT_SPIFLASH2_LAYOUT
+#undef MBOOT_SPIFLASH2_ERASE_BLOCKS_PER_PAGE
+
+#define MICROPY_BOARD_EARLY_INIT    board_early_init_sf6
+
+#define MICROPY_HW_SPIFLASH_DETECT_DEVICE (1)
+#define MICROPY_HW_SPIFLASH_QREAD_NUM_DUMMY(spiflash) (chip_params[spiflash->config == &spiflash_config ? 0 : 1]->qread_num_dummy)
```
projectgus, line 90: Having to look up entries in the 2-element `chip_params` array seems a little awkward.

Did you look at allowing a board to optionally enable a `const mp_spiflash_chip_params_t *params` member in `mp_spiflash_config_t`? That would allow all these ternary expressions to become `spiflash->config->params`, I think?)

ports/stm32/qspi.c
```diff
@@ -370,7 +378,7 @@ static int qspi_read_cmd_qaddr_qdata(void *self_in, uint8_t cmd, uint32_t addr,
             | 0 << QUADSPI_CCR_SIOO_Pos // send instruction every transaction
             | 1 << QUADSPI_CCR_FMODE_Pos // indirect read mode
             | 3 << QUADSPI_CCR_DMODE_Pos // data on 4 lines
-            | 4 << QUADSPI_CCR_DCYC_Pos // 4 dummy cycles
+            | (2 * num_dummy) << QUADSPI_CCR_DCYC_Pos // 2N dummy cycles
```
projectgus, line 381: Is the num_dummy parameter doubled everywhere it's used? If so, would it be clearer to double the values stored in the chip params, so `num_dummy` is exactly the number of dummy cycles?

projectgus (commented):
I have a couple of comments/suggestions but in general this LGTM. It's fiddly but as far as I can tell it's necessarily fiddly. :/

ports/stm32/boards/PYBD_SF6/mpconfigboard.h
dpgeorge: It can't be in `mp_spiflash_config_t` because that's a `const` struct that defines the *bus* and not the thing attached to the bus.  That struct would need to become non-const, and its definition included in `mpconfigboard.h`.

dpgeorge: I did try putting such an element in the `mp_spiflash_t` struct, but it means that `drivers/memory/spiflash.h` needs to be included in `mpconfigboard.h` which gets messy (but it's doable).

projectgus: Oh, my mistake.

What about having a board-level flag like `define MICROPY_HW_SPIFLASH_PARAMS 1` (or `define MICROPY_HW_SPIFLASH_PARAMS mp_spiflash_chip_params_t` if the structure definition is chosen by the board)?

That doesn't seem to create more inversion in the headers than this approach, does it?

ports/stm32/qspi.c
projectgus: My mistake, missed that usage when I searched.

ports/stm32/boards/PYBD_SF6/mpconfigboard.h
dpgeorge: > What about having a board-level flag like `define MICROPY_HW_SPIFLASH_PARAMS 1`

Do you mean that that config option would add a `mp_spiflash_chip_params_t *` entry to `mp_spiflash_t`?

projectgus: > Do you mean that that config option would add a `mp_spiflash_chip_params_t *` entry to `mp_spiflash_t`?

Yes, like that.

dpgeorge: OK, this turned out quite good, and allowed me to move `mp_spiflash_chip_params_t` to `drivers/memory/spiflash.h`.

The trick to getting it working was to `#include "storage.h"` in `qspi.c`, so the latter knows about all the needed struct's (from `spi_bdev2`).  Because it's impossible to include `storage.h` in `mpconfigboard.h`, which is what tripped me up in my previous attempt.

ports/stm32/boards/PYBD_SF6/mpconfigboard.h
```diff
@@ -69,3 +71,44 @@
...
+#undef MBOOT_SPIFLASH_LAYOUT
+#undef MBOOT_SPIFLASH_ERASE_BLOCKS_PER_PAGE
+#undef MBOOT_SPIFLASH2_BYTE_SIZE
+#undef MBOOT_SPIFLASH2_LAYOUT
+#undef MBOOT_SPIFLASH2_ERASE_BLOCKS_PER_PAGE
+
+#define MICROPY_BOARD_EARLY_INIT    board_early_init_sf6
+
+#define CHIP_PARAMS0 (spi_bdev.spiflash.chip_params)
+#define CHIP_PARAMS1 (spi_bdev2.spiflash.chip_params)
```
dpgeorge, line 90: I'll rename these macros to something better if we agree this is a good approach.

robert-hh:
@iabdalkader Reading the SFDP spec again it seems I was not right. The size of the flash is coded there, and quite a few more important parameters, like how for write the status bits. Things that would have to be defined otherwise. So using that information looks promising for a more generic approach to flash. It should be possible to drop quite a few board variants/boards which just differ by flash parameters.

iabdalkader:
> @iabdalkader Reading the SFDP spec again it seems I was not right. The size of the flash is coded there, and quite a few more important parameters, like how for write the status bits. Things that would have to be defined otherwise. So using that information looks promising for a more generic approach to flash. It should be possible to drop quite a few board variants/boards which just differ by flash parameters.

Yes, if I remember correctly, it also includes many useful things for OSPI chips (opcodes to switch to octal mode, commands for erase/read/write, dummy cycles? etc..). This would be a very good alternative to many things, but it's quiet an undertaking to implement generically, so I also resorted to tables in another port.

projectgus:
> This would be a very good alternative to many things, but it's quiet an undertaking to implement generically, so I also resorted to tables in another port.

FWIW I fully agree that it's useful but also a big undertaking. It'd be interesting to see if we could find a way to do it generically across ports.

My other recollection of SFDP, from Espressif, is that eventually there are chips which need manually inserted "quirks" because the SFDP parameters are wrong or don't apply in a particularly host config, or something like that. That's not a reason to discount it, though.

ports/stm32/boards/PYBD_SF6/mpconfigboard.h
projectgus: This approach looks good to me!

ports/stm32/boards/PYBD_SF6/mpconfigboard.h
dpgeorge: OK, now renamed.

(I still need to squash the fixing commits, but otherwise this should be in its final state now.)

ports/stm32/boards/PYBD_SF6/mpconfigboard.h
```diff
@@ -86,28 +86,28 @@
 
 #define MICROPY_BOARD_EARLY_INIT    board_early_init_sf6
 
-#define CHIP_PARAMS0 (spi_bdev.spiflash.chip_params)
-#define CHIP_PARAMS1 (spi_bdev2.spiflash.chip_params)
+#define MICROPY_BOARD_SPIFLASH_CHIP_PARAMS0 (spi_bdev.spiflash.chip_params)
+#define MICROPY_BOARD_SPIFLASH_CHIP_PARAMS1 (spi_bdev2.spiflash.chip_params)
```
projectgus, line 90: Minor, but suggest a couple of inline comments here about `// QSPI Flash 1`, `// QSPI Flash 2` or similar, to link these back to the hardware naming scheme

projectgus (approved):

robert-hh:
> FWIW I fully agree that it's useful but also a big undertaking. It'd be interesting to see if we could find a way to do it generically across ports.

I see a few challenges:
- Verify, that all information a port needs is in the mandatory section of SFDP. That's the one which should always be there IF the flash device supports SFDP.
- Get a sufficient test coverage and test confidence over the various chip types that exist.
- How to deal with flash chips that do not support SFDP. I have seen a few, and those were chips with smaller capacity, like <= 1MByte. At least, all flash chips I have seen larger than 1Mbyte had SFDP support.
- Generalize it across ports, where possible, as you said.

robert-hh:
Looking again into the various versions of the SFDP spec and into the SFDP data of various chips the situation is not overly promising. Besides the flash size and the size of the addresses a required information that varies between chips is the method of enabling the quad write mode == setting the QE flag. This piece of information is not covered by the initial SFDP spec. Newer version of SFDP have it in DWORD15 of the data. The problem: None of flash chips which I looked at have this data. They all comply only to the first version of SFDP. That means, that either this QE flag method has to be configured manually, or Quad Write cannot be supported. That sounds worse than it may be. Writing is relatively slow, and at high clock frequencies the data transfer time is way smaller than the page program time, like ~50µs transfer time @ 40Mhz vs 500 µs.  page program time.

dpgeorge:
> Besides the flash size and the size of the addresses a required information that varies between chips is the method of enabling the quad write mode == setting the QE flag. This piece of information is not covered by the initial SFDP spec

You could still maintain some simple logic to work out how to set the QE flag.  There are probably only a handful of ways that need to be supported for that.

Regardless, I'm not sure it's worth the effort to do this right now.  There are many other things to work on!

robert-hh:
> here are probably only a handful of ways that need to be supported for that.

The flag has 6 possible settings, but three of them can be handled the same way, and I have never seen one of them. So it's 3 or 4 variants.

PR #17074, opened 2025-04-04
extmod/modtls_mbedtls: Add support for TLS PSK

keenanjohnson:
This pull request adds support to the ssl module for TLS-PSK (pre-shared key) authentication.

It enables the feature on the esp32 port as well, which fits my use case. 

This PR was originally based on the work in https://github.com/micropython/micropython/pull/17074. I've bumped the other thread a few times, but figured I would move the discussion into my actual implementation here and the implementation has since diverged.


keenanjohnson:
Hey @projectgus and @dpgeorge . I've created this draft of an approach to add support for TLS PSK based on the work in https://github.com/micropython/micropython/pull/17074. I've bumped the other thread a few times, but figured I would move the discussion into my actual implementation here. 

There was some discussion on the different possible approaches and I believe what I have implemented here follows @dpgeorge 's suggestion in 17074. How does it look and does this direction seem good? If so, I can finalize the paperwork to close this out.

extmod/modtls_mbedtls.c
```diff
@@ -382,10 +392,87 @@ static mp_obj_t ssl_context_get_ciphers(mp_obj_t self_in) {
...
+        if (strcmp(ciphername, "PSK") == 0) {
+            ssl_context->use_psk = true;
+            set_psk_ciphersuites(&ssl_context->conf);
+            return mp_const_none;
+        }
+
+        // Check if this is a PSK ciphersuite name
+        if (strncmp(ciphername, "PSK-", 4) == 0 ||
+            strncmp(ciphername, "TLS-PSK-", 8) == 0 ||
+            strncmp(ciphername, "TLS_PSK_", 8) == 0) {
```
AJMansfield: Is there a reference that can be cited here to justify this parse? This needs some authority behind it.

Also, why not just use mbedtls's own API to determine both the "is psk" and "is recognised" question together? 
```C
const mbedtls_ssl_ciphersuite_t *info = mbedtls_ssl_ciphersuite_from_string(ciphername);
if (mbedtls_ssl_ciphersuite_uses_psk(info)) {
    ...
```
(With some extra bits to correctly handle the 'failed lookup' case, of course.)
(Assuming that special-casing PSK here is even the right design anyway.)

extmod/modtls_mbedtls.c
```diff
@@ -382,10 +392,87 @@ static mp_obj_t ssl_context_get_ciphers(mp_obj_t self_in) {
...
+
+            // Try to look up the ciphersuite ID
+            const int id = mbedtls_ssl_get_ciphersuite_id(ciphername);
+            if (id != 0) {
+                // Enable PSK mode
+                ssl_context->use_psk = true;
+
+                // Create a ciphersuite array with just this one ciphersuite
```
AJMansfield: Why should PSK ciphers be singleton, rather than parsed and accumulated as part of the existing list code?

extmod/modtls_mbedtls.c
```diff
@@ -467,6 +554,22 @@ static mp_obj_t ssl_context_load_verify_locations(mp_obj_t self_in, mp_obj_t cad
...
+static MP_DEFINE_CONST_FUN_OBJ_2(ssl_context_set_psk_identity_obj, ssl_context_set_psk_identity);
+
+// SSLContext.set_psk_key(key)
+static mp_obj_t ssl_context_set_psk_key(mp_obj_t self_in, mp_obj_t key) {
+    mp_obj_ssl_context_t *self = MP_OBJ_TO_PTR(self_in);
+    self->psk_key = key;
+    return mp_const_none;
+}
+static MP_DEFINE_CONST_FUN_OBJ_2(ssl_context_set_psk_key_obj, ssl_context_set_psk_key);
+
```
AJMansfield: Why the intermediate `self->psk_identity` and `self->psk_key` variables? i.e. instead of a single function that combines the code with the code in `ssl_socket_make_new` to call `mbedtls_ssl_conf_psk` directly.

extmod/modtls_mbedtls.c
```diff
@@ -603,6 +708,22 @@ static mp_obj_t ssl_socket_make_new(mp_obj_ssl_context_t *ssl_context, mp_obj_t
 
     mbedtls_ssl_init(&o->ssl);
 
+    // Configure PSK if enabled
+    if (ssl_context->use_psk && ssl_context->psk_identity != mp_const_none && ssl_context->psk_key != mp_const_none) {
```
AJMansfield: This condition should be more paranoid so it can give a better error. Mbedtls would error eventually, but any case where you don't have either _all_ or _none_ of those conditions, should be an error already.

extmod/modtls_mbedtls.c
```diff
@@ -603,6 +708,22 @@ static mp_obj_t ssl_socket_make_new(mp_obj_ssl_context_t *ssl_context, mp_obj_t
...
+        // Get PSK identity and key
+        size_t psk_identity_len;
+        const byte *psk_identity = (const byte *)mp_obj_str_get_data(ssl_context->psk_identity, &psk_identity_len);
+        
+        size_t psk_key_len;
+        const byte *psk_key = (const byte *)mp_obj_str_get_data(ssl_context->psk_key, &psk_key_len);
```
AJMansfield: These conversions could hit the `bad_implicit_conversion` path and bypass cleanup --- neither this method nor the setters for those variables actually check `mp_obj_is_str_or_bytes`.

extmod/modtls_mbedtls.c
```diff
@@ -382,10 +392,87 @@ static mp_obj_t ssl_context_get_ciphers(mp_obj_t self_in) {
...
+            MBEDTLS_TLS_PSK_WITH_AES_128_GCM_SHA256,
+            MBEDTLS_TLS_PSK_WITH_AES_256_GCM_SHA384,
+            0  // Terminating zero
+        };
+        
+        // Count available PSK ciphersuites
+        int count = 0;
+        for (int i = 0; known_psk_ciphersuites[i] != 0; i++) {
+            count++;
+        }
```
AJMansfield: Get the actual authoritative list of supported PSK suites by filtering the output from `mbedtls_ssl_list_ciphersuites` using `mbedtls_ssl_ciphersuite_from_id` and `mbedtls_ssl_ciphersuite_uses_psk`.

extmod/modtls_mbedtls.c
```diff
@@ -382,10 +392,87 @@ static mp_obj_t ssl_context_get_ciphers(mp_obj_t self_in) {
...
+
+        // Allocate memory for PSK ciphersuites
+        psk_ciphersuites = m_new(int, count + 1);
+        if (psk_ciphersuites == NULL) {
+            mp_raise_OSError(MP_ENOMEM);
```
AJMansfield: Redundant with the `m_malloc_fail` path already present in the allocator.

extmod/modtls_mbedtls.c
```diff
@@ -382,10 +392,87 @@ static mp_obj_t ssl_context_get_ciphers(mp_obj_t self_in) {
...
+            // Try to look up the ciphersuite ID
+            const int id = mbedtls_ssl_get_ciphersuite_id(ciphername);
+            if (id != 0) {
+                // Enable PSK mode
+                ssl_context->use_psk = true;
+
+                // Create a ciphersuite array with just this one ciphersuite
+                ssl_context->ciphersuites = m_new(int, 2);
+                if (ssl_context->ciphersuites == NULL) {
+                    mp_raise_OSError(MP_ENOMEM);
```
AJMansfield: Redundant with the `m_malloc_fail` path already present in the allocator.

extmod/modtls_mbedtls.c
```diff
@@ -603,6 +708,22 @@ static mp_obj_t ssl_socket_make_new(mp_obj_ssl_context_t *ssl_context, mp_obj_t
...
+        size_t psk_identity_len;
+        const byte *psk_identity = (const byte *)mp_obj_str_get_data(ssl_context->psk_identity, &psk_identity_len);
+        
+        size_t psk_key_len;
+        const byte *psk_key = (const byte *)mp_obj_str_get_data(ssl_context->psk_key, &psk_key_len);
+
+        // Configure PSK
+        ret = mbedtls_ssl_conf_psk(&ssl_context->conf, psk_key, psk_key_len, psk_identity, psk_identity_len);
+        if (ret != 0) {
+            goto cleanup;
```
AJMansfield: This path _needs_ coverage, with test cases that set invalid keys and identites.

AJMansfield (changes requested):
The commits in this PR need to be squashed and reworded to match MicroPython's [Code Conventions](https://github.com/micropython/micropython/blob/master/CODECONVENTIONS.md). Aside from that:

AJMansfield:
For overall structure and direction, though:
- Drop the separate setters and wrap `mbedtls_ssl_conf_psk` a bit more thinly.
- Let the SSL/TLS cipher suite abstraction work _for_ you, instead of making PSK a special case in places it could just not be.

AJMansfield:
This should _probably_ be feature-gated? e.g. with a new `MICROPY_PY_SSL_PSK` added to `py/mpconfig.h` next to the other `MICROPY_PY_SSL_*` flags. +3kB is a lot, even for mbedtls builds.

Re: #14396 and the client-callback and server-callback API, in my opinion it could be appropriate to add those too, at least if behind another feature gate like `MICROPY_PY_SSL_PSK_CALLBACKS`; though that would be a PR for after the basic version lands.

keenanjohnson:
Squashed and working through review feedback now.

extmod/modtls_mbedtls.c
```diff
@@ -439,6 +528,46 @@ static mp_obj_t ssl_context_set_ciphers(mp_obj_t self_in, mp_obj_t ciphersuite)
...
+            id == 0xB2 ||  // MBEDTLS_TLS_DHE_PSK_WITH_AES_128_CBC_SHA256
+            id == 0xB3 ||  // MBEDTLS_TLS_DHE_PSK_WITH_AES_256_CBC_SHA384
+            id == 0xB4 ||  // MBEDTLS_TLS_DHE_PSK_WITH_NULL_SHA256
+            id == 0xB5 ||  // MBEDTLS_TLS_DHE_PSK_WITH_NULL_SHA384
+            id == 0xB6 ||  // MBEDTLS_TLS_RSA_PSK_WITH_AES_128_CBC_SHA256
+            id == 0xB7 ||  // MBEDTLS_TLS_RSA_PSK_WITH_AES_256_CBC_SHA384
+            id == 0xB8 ||  // MBEDTLS_TLS_RSA_PSK_WITH_NULL_SHA256
+            id == 0xB9);   // MBEDTLS_TLS_RSA_PSK_WITH_NULL_SHA384
+}
+
```
AJMansfield: It's a pity the actual `mbedtls_ssl_ciphersuite_uses_psk` is only part of `mbedtls/library/ssl_ciphersuites_internal.h` header rather than a public include --- but IMO it's still better to just link against it anyway instead of re-implementing it. Just declare it yourself the same way its header does _without_ defining it:

```C
int mbedtls_ssl_ciphersuite_uses_psk(const mbedtls_ssl_ciphersuite_t *info);
```

If that's too spicy, add `MBEDTLS_ALLOW_PRIVATE_ACCESS` to `mbedtls_config_common.h` and examine the value of `info->key_exchange` against the four values it can have that indicate PSK suites.

extmod/modtls_mbedtls.c
```diff
@@ -493,6 +622,23 @@ static mp_obj_t ssl_context_load_verify_locations(mp_obj_t self_in, mp_obj_t cad
...
+    return mp_const_none;
+}
+static MP_DEFINE_CONST_FUN_OBJ_2(ssl_context_set_psk_identity_obj, ssl_context_set_psk_identity);
+
+static mp_obj_t ssl_context_set_psk_key(mp_obj_t self_in, mp_obj_t key) {
+    psk_key = key;
+    return mp_const_none;
+}
+static MP_DEFINE_CONST_FUN_OBJ_2(ssl_context_set_psk_key_obj, ssl_context_set_psk_key);
+
```
AJMansfield: Did ChatGPT write this? Because _that's_ a lie:

> These methods now configure PSK directly with mbedtls instead of storing values

AJMansfield: The extra variables in `mp_obj_ssl_context_t` was a bit sub-optimal, but using shared static variables like this is _completely unacceptable_. Even if we ignore mutliple contexts clobbering each other, it's also a use-after-free bug. (Can you spot it? It's because BSS variables aren't visible to micropython's garbage collector unless you give them a root-pointer decorator. The key and identity values here could be eligible for collection the instant these functions return.)

AJMansfield: The point of my earlier comment was this: The `mp_obj_ssl_context_t` object _already has_ a place to store these values, in its contained `mbedtls_ssl_config` value --- i.e. the place that calling `mbedtls_ssl_conf_psk` stores them.
The API here should just be the thinnest possible veneer over that function to make it callable from micropython --- just a function that takes two string arguments, validates that they're strings, extracts the underlying string pointers, and calls `mbedtls_ssl_conf_psk` with them.

extmod/modtls_mbedtls.c
AJMansfield: Also, don't hardcode the numeric values, use the symbols! You can just `id == MBEDTLS_TLS_PSK_WITH_NULL_SHA`, you already even have the `mbedtls/ssl_ciphersuites.h` header that defines those.

AJMansfield: Also also, you missed a couple:
```
MBEDTLS_TLS_DHE_PSK_WITH_AES_128_CBC_SHA256
MBEDTLS_TLS_DHE_PSK_WITH_AES_256_CBC_SHA384
MBEDTLS_TLS_DHE_PSK_WITH_NULL_SHA256
MBEDTLS_TLS_DHE_PSK_WITH_NULL_SHA384
MBEDTLS_TLS_RSA_PSK_WITH_AES_128_CBC_SHA256
MBEDTLS_TLS_RSA_PSK_WITH_AES_256_CBC_SHA384
MBEDTLS_TLS_RSA_PSK_WITH_NULL_SHA256
MBEDTLS_TLS_RSA_PSK_WITH_NULL_SHA384
MBEDTLS_TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA
MBEDTLS_TLS_ECDHE_PSK_WITH_AES_256_CBC_SHA
MBEDTLS_TLS_ECDHE_PSK_WITH_AES_128_CBC_SHA256
MBEDTLS_TLS_ECDHE_PSK_WITH_AES_256_CBC_SHA384
MBEDTLS_TLS_ECDHE_PSK_WITH_NULL_SHA
MBEDTLS_TLS_ECDHE_PSK_WITH_NULL_SHA256
MBEDTLS_TLS_ECDHE_PSK_WITH_NULL_SHA384
MBEDTLS_TLS_PSK_WITH_ARIA_128_CBC_SHA256
MBEDTLS_TLS_PSK_WITH_ARIA_256_CBC_SHA384
MBEDTLS_TLS_DHE_PSK_WITH_ARIA_128_CBC_SHA256
MBEDTLS_TLS_DHE_PSK_WITH_ARIA_256_CBC_SHA384
MBEDTLS_TLS_RSA_PSK_WITH_ARIA_128_CBC_SHA256
MBEDTLS_TLS_RSA_PSK_WITH_ARIA_256_CBC_SHA384
MBEDTLS_TLS_PSK_WITH_ARIA_128_GCM_SHA256
MBEDTLS_TLS_PSK_WITH_ARIA_256_GCM_SHA384
MBEDTLS_TLS_DHE_PSK_WITH_ARIA_128_GCM_SHA256
MBEDTLS_TLS_DHE_PSK_WITH_ARIA_256_GCM_SHA384
MBEDTLS_TLS_RSA_PSK_WITH_ARIA_128_GCM_SHA256
MBEDTLS_TLS_RSA_PSK_WITH_ARIA_256_GCM_SHA384
MBEDTLS_TLS_ECDHE_PSK_WITH_ARIA_128_CBC_SHA256
MBEDTLS_TLS_ECDHE_PSK_WITH_ARIA_256_CBC_SHA384
MBEDTLS_TLS_PSK_WITH_CAMELLIA_128_GCM_SHA256
MBEDTLS_TLS_PSK_WITH_CAMELLIA_256_GCM_SHA384
MBEDTLS_TLS_DHE_PSK_WITH_CAMELLIA_128_GCM_SHA256
MBEDTLS_TLS_DHE_PSK_WITH_CAMELLIA_256_GCM_SHA384
MBEDTLS_TLS_RSA_PSK_WITH_CAMELLIA_128_GCM_SHA256
MBEDTLS_TLS_RSA_PSK_WITH_CAMELLIA_256_GCM_SHA384
MBEDTLS_TLS_PSK_WITH_CAMELLIA_128_CBC_SHA256
MBEDTLS_TLS_PSK_WITH_CAMELLIA_256_CBC_SHA384
MBEDTLS_TLS_DHE_PSK_WITH_CAMELLIA_128_CBC_SHA256
MBEDTLS_TLS_DHE_PSK_WITH_CAMELLIA_256_CBC_SHA384
MBEDTLS_TLS_RSA_PSK_WITH_CAMELLIA_128_CBC_SHA256
MBEDTLS_TLS_RSA_PSK_WITH_CAMELLIA_256_CBC_SHA384
MBEDTLS_TLS_ECDHE_PSK_WITH_CAMELLIA_128_CBC_SHA256
MBEDTLS_TLS_ECDHE_PSK_WITH_CAMELLIA_256_CBC_SHA384
MBEDTLS_TLS_PSK_WITH_AES_128_CCM
MBEDTLS_TLS_PSK_WITH_AES_256_CCM
MBEDTLS_TLS_DHE_PSK_WITH_AES_128_CCM
MBEDTLS_TLS_DHE_PSK_WITH_AES_256_CCM
MBEDTLS_TLS_PSK_WITH_AES_128_CCM_8
MBEDTLS_TLS_PSK_WITH_AES_256_CCM_8
MBEDTLS_TLS_DHE_PSK_WITH_AES_128_CCM_8
MBEDTLS_TLS_DHE_PSK_WITH_AES_256_CCM_8
MBEDTLS_TLS_PSK_WITH_CHACHA20_POLY1305_SHA256
MBEDTLS_TLS_ECDHE_PSK_WITH_CHACHA20_POLY1305_SHA256
MBEDTLS_TLS_DHE_PSK_WITH_CHACHA20_POLY1305_SHA256
MBEDTLS_TLS_RSA_PSK_WITH_CHACHA20_POLY1305_SHA256
```

keenanjohnson:
Hey @AJMansfield , I appreciate the detailed feedback, but I'm still working through it all, so I'm not sure that's very useful you providing feedback on my intermediate work. I'll ping you directly and set this PR out of draft to ready once I'm ready for another round of feedback. I just like to push to the branch often to get the CI to run and cross check my local setup.

AJMansfield:
> Hey @AJMansfield , I appreciate the detailed feedback, but I'm still working through it all, so I'm not sure that's very useful you providing feedback on my intermediate work. I'll ping you directly and set this PR out of draft to ready once I'm ready for another round of feedback. I just like to push to the branch often to get the CI to run and cross check my local setup.

Understood! Apologies for the pressure: you're right it does neither of us any good to have me constantly looking over your shoulder, we've all got our own workflows. (I certainly made extensive use of the CI for testing changes myself at one point, lol.) Just be careful with the LLMs --- this _is_ a cryptographic module, after all.

projectgus:
@keenanjohnson Feel free to ping me as well once you're done updating.

Thanks @AJMansfield for getting the review process underway.

projectgus:
Hi @keenanjohnson,

No rush on this as I assume you've been very busy with other things. I just now left a long comment on the related PR: https://github.com/micropython/micropython/pull/14396#issuecomment-4551334229 - and the considerations there apply here more or less as well.

One high-level concern for this PR is how we'd manage to add a CPython-compatible callback wrapper, as discussed in the other PR.

* Client side is probably OK, the related PR only adds a hard-coded client callback as mbedTLS doesn't support anything more sophisticated. I assume we could build equivalent functionality on top of this PR.
* Server side seems harder, I don't see a way we could have a callback fire on each client connection and make it function similarly to CPython. (It's debatable whether that's necessary for an embedded application, but the CPython consistency is valuable.) (EDIT: [Damien has commented on the linked PR with a way that this could work](EDIT: [Damien has commented on the linked PR with a way that this could work](https://github.com/micropython/micropython/pull/14396#issuecomment-4551451019).)

As I said, no rush or pressure to respond but I wanted to leave this comment while it was on my mind. If anyone wants to pick up either of these TLS-PSK PRs and work on a new one, that'd be much appreciated - the feature is potentially very useful to have.

keenanjohnson:
Thanks @projectgus and @dpgeorge ! I have stepped away from this for a bit, but I would very much like to see it upstreamed as it would be useful for my particular project! Let me re read everything and recontextuatlize myself here and I'll try to address the design decision / point you brought up in the other PR.

extmod/modtls_mbedtls.c
keenanjohnson: Resolved in latest

extmod/modtls_mbedtls.c
keenanjohnson: Resolved in latest

extmod/modtls_mbedtls.c
keenanjohnson: Yes fixed in the latest

extmod/modtls_mbedtls.c
keenanjohnson: Fixed in latest!

keenanjohnson:
Ok all. Thanks for the bump @projectgus to put this back on my radar.

I believe I have addressed everything and this PR is ready to go. The CI is all happy. There is one line of code coverage missing which I believe to be a bit impractical to consistently hit in test.

@AJMansfield did some initial review a long while ago. I believe I have addressed all of your feedback, but check me if I missed anything.

@projectgus and @dpgeorge I have taken your feedback and implemented roughly the design you proposed in https://github.com/micropython/micropython/pull/14396#issuecomment-4551451019 where there is a PSK Dict and should keep things as micropython-y as we can.

@projectgus you also mentioned adding some adversarial tests and I have added all the ones I initially thought of and helped me verify some of the security / linking concerns that @AJMansfield originally brought up (i first wrote the tests to prove the security hole and then fixed it until the tests were green).

I enabled the feature on the esp32 port as that's the usecase I personally have. 

Finally i added some basic docs update.

Let me know what you all think and I am happy to respond to review feedback in a timely manner as I want to finally push this over the finish line!

docs/library/ssl.rst
```diff
@@ -66,6 +66,31 @@ class SSLContext
...
+   A mapping used by a server to look up the key for the identity presented by a
+   connecting client.  Set it to accept PSK clients::
+
+      ctx.server_psk_keys = {"my-identity": b"my-key"}
+
+   It is indexed with the client's identity (a string) and should return the
+   corresponding key as a `bytes` object, or raise ``KeyError`` to reject an
+   unknown identity.  Any object implementing ``__getitem__`` may be used, so
+   keys can be computed or fetched on demand.  As with the client, the context
+   is restricted to PSK cipher suites while this is set.
```
dpgeorge: Did you test this scenario, when a user object is passed in and a callback is made into Python?  (I couldn't see anything in the tests below, would be good to add something.)

(Also, an alternative could be to use `dict.get(key)` instead of `dict[key]` which would return `None` if the key didn't exist.  That could be more efficient than raising and catching `KeyError`.)

extmod/modtls_mbedtls.c
```diff
@@ -493,6 +523,81 @@ static mp_obj_t ssl_context_load_verify_locations(mp_obj_t self_in, mp_obj_t cad
...
+static int ssl_context_psk_server_callback(void *p_ctx, mbedtls_ssl_context *ssl,
+    const unsigned char *identity, size_t identity_len) {
+    mp_obj_ssl_context_t *self = (mp_obj_ssl_context_t *)p_ctx;
+    if (self->server_psk_keys == mp_const_none) {
+        return -1;
+    }
+    nlr_buf_t nlr;
+    if (nlr_push(&nlr) == 0) {
+        mp_obj_t key = mp_obj_subscr(self->server_psk_keys,
+            mp_obj_new_str((const char *)identity, identity_len), MP_OBJ_SENTINEL);
```
dpgeorge: I'm not sure, but maybe this should be a `bytes` rather than a `str`.  Need to understand what happens when non-ascii characters are used in the identity.

dpgeorge:
Thanks @keenanjohnson for jumping back on this, it's good to see.  I didn't do a full review yet, but made a few comments above based on initial observations.

docs/library/ssl.rst
keenanjohnson: Good catch! I added a new test for that in tests/multi_net/sslcontext_server_client_psk_getitem.py just now. 

We definitely could switch to using dict.get(key). Honestly I didn't really consider the tradeoff too much when I implemented. I'm happy to switch to that if we think it's better.

extmod/modtls_mbedtls.c
keenanjohnson: Also a good catch. Upon testing any non-ascii characters would raise UnicodeError, so any non UTF-8 ID can never authenticate. Switching to bytes does fix this and lets us eliminate one special case. I have implemented that and just pushed.

docs/library/ssl.rst
keenanjohnson: I did a quick little benchmark and yes the get method was about 3X faster than the currently impelmeted version. I've gone ahead and switched to dict.get(key) and updated everything to match that.

keenanjohnson:
> Thanks @keenanjohnson for jumping back on this, it's good to see. I didn't do a full review yet, but made a few comments above based on initial observations.

Of course @dpgeorge ! I am glad that this finally feels very close. I responded to your initial feedback above.

projectgus (approved):
This looks good, thanks @keenanjohnson! Very useful feature to have, and I appreciate all the test coverage.

keenanjohnson:
You are very welcome @AJMansfield ! This is quite useful for me, as our project has been running on a fork of micropython for a long time due to this missing feature, and I am glad to finally upstream it for others. It's taken a while (mostly my fault for finding the time to get around to it), but I am glad it seems useful.

Hopefully I didn't over do it with the test coverage haha.

keenanjohnson:
Again thank you for the incredible work that you (@projectgus ) and @dpgeorge and the rest of the team and community around micropython do. I'm certain you likely don't get enough thanks or credit, but this project is incredible and enables a lot of learning and projects around the world.

dpgeorge:
I ran the included tests on PYBD_SF6 vs RPI_PICO_W.  Everything passed except one test, `sslcontext_server_client_psk_wrong_key_error.py`.  It hangs on the client `instance1()` in `wrap_socket()`.

The following fixes it for me and gets the test passing on those boards.  The test also still passes on the unix port.
```diff
--- a/tests/multi_net/sslcontext_server_client_psk_wrong_key_error.py
+++ b/tests/multi_net/sslcontext_server_client_psk_wrong_key_error.py
@@ -32,6 +32,7 @@ def instance0():
     except OSError:
         print("server: handshake failed")
     multitest.broadcast("finished")
+    s2.close()
     s.close()
```
It may be that this is needed due to our implementation of TCP/IP with lwIP on bare metal (because it passes without it on unix, and also esp32 I assume).  I don't have time to investigate why that close is needed, but it seems sensible enough to clean up all the sockets on the server side, and I guess it unblocks the client allowing it to see that the handshake failed.

keenanjohnson:
Huh very curious @dpgeorge . The theory seems reasonable and I can't see any downside to adding the close socket so I have added it to the test that you mentioned, but alos the other four tests that follow this pattern here to try and prevent any flakiness there as well.

dpgeorge:
> The theory seems reasonable and I can't see any downside to adding the close socket so I have added it to the test that you mentioned, but alos the other four tests that follow this pattern here to try and prevent any flakiness there as well.

Thanks for adding that close socket.  I tested again on PYBD_SF6 and RPI_PICO2_W, and all tests pass there.

extmod/modtls_mbedtls.c
```diff
@@ -493,6 +523,76 @@ static mp_obj_t ssl_context_load_verify_locations(mp_obj_t self_in, mp_obj_t cad
...
+#if defined(MBEDTLS_SSL_HANDSHAKE_WITH_PSK_ENABLED)
+// mbedtls_ssl_ciphersuite_uses_psk() is declared only in mbedTLS's private
+// library/ssl_ciphersuites_internal.h, but it is a regular (linkable) symbol.
+// Declare it here so it can be used without pulling in that private header.
+int mbedtls_ssl_ciphersuite_uses_psk(const mbedtls_ssl_ciphersuite_t *info);
+
+// Restrict the context to offer/accept only PSK cipher suites, so a PSK
+// connection cannot silently fall back to a non-PSK (e.g. certificate-based)
+// suite.  Does nothing if the user already chose ciphers via set_ciphers().
+static void ssl_context_restrict_to_psk_ciphersuites(mp_obj_ssl_context_t *self) {
```
dpgeorge, line 546: I don't think this function is tested by the existing tests: I added a simple `return` statement at the start of this function and all tests continued to pass.  I thought that `sslcontext_server_client_psk_client_only_error.py` was testing the logic here, but it seems not.

Are you able to write a test that exercises this function?  So that if this function just returns immediately then the test fails?

extmod/modtls_mbedtls.c
```diff
@@ -637,6 +737,29 @@ static mp_obj_t ssl_socket_make_new(mp_obj_ssl_context_t *ssl_context, mp_obj_t
 
     mbedtls_ssl_init(&o->ssl);
 
+    #if defined(MBEDTLS_SSL_HANDSHAKE_WITH_PSK_ENABLED)
+    // Configure PSK from the context.  A server looks up keys by the client's
```
dpgeorge: This block of logic that sets the PSK values in mbedTLS doesn't seem to depend at all on the actual socket state.  It only depends on the `SSLContext`.

So I guess all this logic could be moved to the `ssl_context_attr()` function:
- if the user stores into `server_psk_keys` then `mbedtls_ssl_conf_psk_cb()` is called
- if the user stores into `psk_identity`/`psk_key` and they are both non-None, then `mbedtls_ssl_conf_psk()` is called

Benefits of doing it that way are:
- only needs to configure things once (when the attributes are set), instead of each time a socket is created (and the whole point of SSLContext is to be able to configure everything in the context and then creating sockets is less work)
- allows errors to be caught earlier
- probably smaller code size
- makes it clearer in the code that setting PSK attributes affects all sockets created from that SSLContext (whereas the current code seems to indicate that PSK settings are stored per socket, rather than per context)
- matches how the existing `verify_mode` is configured

tests/multi_net/sslcontext_server_client_psk.py
```diff
@@ -0,0 +1,54 @@
+# Test TCP server and client using TLS-PSK (pre-shared key).  The client
+# presents an identity/key via the psk_identity/psk_key attributes, and the
+# server looks the identity up in its server_psk_keys mapping.
+
+try:
+    import socket
+    import tls
+except ImportError:
```
dpgeorge: All these multi_net tests need to skip if PSK is not supported (eg using the same skip logic from `extmod/tls_psk.py`).

Eg esp8266 which uses axTLS will attempt to run this test and needs to skip it.

extmod/modtls_mbedtls.c
keenanjohnson: Ah yes that's a good catch that I missed in my test coverage. I confirmed the same with an immediate return. I added a test for both the client and server side version of this for good measure where each direction negotiates the PSK cipher suite, but presents the incorrect key.

tests/multi_net/sslcontext_server_client_psk_no_fallback_error.py
tests/multi_net/sslcontext_server_client_psk_server_no_fallback_error.py

tests/multi_net/sslcontext_server_client_psk.py
keenanjohnson: Ah yes I see. I did not think about that. Implemented the block below in all tests:

```
try:
    import socket
    import tls

    # PSK support is optional; psk_identity only exists when it's enabled.
    tls.SSLContext(tls.PROTOCOL_TLS_CLIENT).psk_identity
except (ImportError, AttributeError):
    print("SKIP")
    raise SystemExit
```

extmod/modtls_mbedtls.c
keenanjohnson: Yes that is probably better. I created a new helper function ssl_context_set_psk() that is called as you have described in the socket context.

keenanjohnson:
I believe I have addressed all the feedback @dpgeorge

extmod/modtls_mbedtls.c
dpgeorge: Thanks for updating and adding the new tests.  I can confirm that they both test unique code paths (ie the tests are both necessary and the two calls to `ssl_context_restrict_to_psk_ciphersuites()` are both necessary).

extmod/modtls_mbedtls.c
dpgeorge: Very good, thanks!

tests/multi_net/sslcontext_server_client_psk.py
dpgeorge: Thanks, I confirm that skips correctly on esp8266.

dpgeorge (approved):
This PR is looking very good now, thanks for responding to all the feedback.

keenanjohnson:
Excellent. I am so glad we got this closed out after so long. Thank you for all the review help!

keenanjohnson:
Any other open items before this is ready for merge?

dpgeorge:
> Any other open items before this is ready for merge?

No, this is ready for merging now.

dpgeorge:
Merged!  Thank you very much @keenanjohnson for your efforts here.

keenanjohnson:
Excellent and thank you for your help! I hope others find this as useful as I do,

PR #17084, opened 2025-04-06
Introduce Zephyr Filesystem VFS interface

VynDragon:
### Summary

Adds VFS interface for Zephyr filesystems

This allows using Zephyr's supported VFS and interface to them via MPY, making mpy side VFS unecessary (eg you can disable LFS2 in mpy but still do LFS2 because it's using zephyr's)

### Testing

Basic tests on nrf52840, would recommend thorough testing.

15/08/2025: also tested on rp2040.


jonnor:
What is needed to set this up on a device? Any limitations, or can one use any file-system operations from Python like on other ports?

VynDragon:
Basically you need to setup a fstap and then format it: https://docs.zephyrproject.org/latest/services/file_system/index.html

After that it works like any mpy fs.

VynDragon:
rebased on main

VynDragon:
Added guardrail for automount / mkfs.

Here is to use it on rp2040, this was for a 16MB rp2040 board so adapt the sizes to yours.

boards/rpi_pico.conf:
```
# Disable floating point hardware.
CONFIG_FPU=n

# Configure serial console over USB CDC ACM.
CONFIG_USB_DEVICE_STACK_NEXT=y
CONFIG_USBD_CDC_ACM_CLASS=y
CONFIG_UART_LINE_CTRL=y

# Disable networking.
CONFIG_NETWORKING=n

# Hardware features.
CONFIG_FLASH=y
CONFIG_FLASH_MAP=y
CONFIG_I2C=y
CONFIG_I2C_TARGET=y
CONFIG_SPI=y

# MicroPython config.
CONFIG_MICROPY_HEAP_SIZE=100608
CONFIG_FILE_SYSTEM=y
CONFIG_FILE_SYSTEM_LITTLEFS=y
CONFIG_FILE_SYSTEM_MKFS=y
```

boards/rpi_pico.overlay:
```
/*
 * Copyright (c) 2025 Damien P. George
 *
 * SPDX-License-Identifier: Apache-2.0
 */

/ {
	chosen {
		/* Use USB CDC ACM as the console. */
		zephyr,console = &cdc_acm_uart0;
	};

	fstab {
		compatible = "zephyr,fstab";
		lfs21: lfs21 {
			compatible = "zephyr,fstab,littlefs";
			mount-point = "/flash";
			partition = <&storage_partition>;
			read-size=<16>;
			prog-size=<4096>;
			cache-size=<4096>;
			lookahead-size=<32>;
			block-cycles=<4>;
		};
	};
};

/* Delete defined partitions and make a layout matching the rp2 port RPI_PICO configuration. */
/delete-node/ &code_partition;
&flash0 {
	reg = <0x10000000 DT_SIZE_M(16)>;
	partitions {
		compatible = "fixed-partitions";
		#address-cells = <1>;
		#size-cells = <1>;

		/* Code slot: 640 KiB - 0x100 placed after second-stage bootloader. */
		code_partition: partition@100 {
			label = "code-partition";
			reg = <0x00000100 (DT_SIZE_K(640) - 0x100)>;
			read-only;
		};

		/* Storage slot: 1408 KiB placed after code_partition. */
		storage_partition: partition@a0000 {
			label = "storage";
			reg = <0x000a0000 DT_SIZE_M(2)>;
		};
	};
};

&zephyr_udc0 {
	cdc_acm_uart0: cdc_acm_uart0 {
		compatible = "zephyr,cdc-acm-uart";
	};
};

&i2c1 {
	clock-frequency = <I2C_BITRATE_STANDARD>;
	status = "okay";
	pinctrl-0 = <&i2c1_default>;
	pinctrl-names = "default";
};
```

In mpy:
```python
import zephyr
zfs = zephyr.FS(zephyr.FS.fstab()[0])
vfs.mount(zfs, "/zephyr")
```

Be aware you may have to disable the automatic vfs mounting in main or umount manually, because mpy's vfs will detect zephyr's and take ownership of it through its own vfs system so /flash is likely to be occupied.

Edit: and i mean the vfs.mount name, the '/flash' in the overlay is the name used in the fstab on zephyr side, not useful mpy side except when creating the class.

bikeNomad:
Since there's already code to mount a filesystem on the partition labeled `storage` as `/flash` (I did add some code to my PR #17679 to create a filesystem if the mount fails), I guess this is mostly useful for other partitions?

VynDragon:
> Since there's already code to mount a filesystem on the partition labeled `storage` as `/flash` (I did add some code to my PR #17679 to create a filesystem if the mount fails), I guess this is mostly useful for other partitions?

No, this is a completely different method as the existing one, and your PR doesnt even use zephyr lfs2, you could disable it in your code and it would still work.

dpgeorge:
This is a nice addition, but I have a few general comments:
1. It's going to get confusing for new users which filesystem to use (native zephyr vs MicroPython).  I don't think there's any way around that except through documentation and examples.
2. You also need to explicitly enable this, and maybe worth adding something to the zephyr README to show how to do that.
3. This won't be tested by CI so most likely will accrue bugs.  Zephyr CI is already slow, so maybe one of the existing CI builds (eg `mimxrt1050_evk`) could have a custom `CONF_FILE` specified for the CI to test building this FS code?

ports/zephyr/zephyr_fs.c
```diff
@@ -0,0 +1,767 @@
...
+    } else {
+        size_t l = vstr_len(&self->root_dir);
+        vstr_add_str(&self->root_dir, path);
+        path = vstr_null_terminated_str(&self->root_dir);
+        self->root_dir.len = l;
+    }
+    return path;
+}
+
+typedef struct _pyb_file_obj_t {
```
dpgeorge: please rename thing struct to something like `zephyr_file_obj_t`

VynDragon:
Thanks for the review!

I see, I will add a commit that adds to the README and docs.

On the testing, would you mind waiting until the flash controller code for the platform I maintain becomes available and i can add boards and build for it using this? Otherwise I can add rpi pico build to ci, which i think is a well supported platform on zephyr and useful for testing (i use either rp2, nrf52840, WCH, BFLB platforms on zephyr), in which it also fixes a bug related to FS that i've been encountering a lot (it fails to make a new lfs2 system via mpy for some unknown reason, couldn't identify why even with debugging)

ports/zephyr/zephyr_fs.c
VynDragon: done

VynDragon:
@dpgeorge I added a couple new features that means it can replace flash area on rpi_pico, i've set it up to use this new thing.
For enabling the tests it would only be needed to add rpi_pico to the build list (I have zero familiarity with any of the other boards... Will give it an attempt if enabling rpi_pico in tests is not acceptable)

VynDragon:
retargeted soem elements to use with https://github.com/micropython/micropython/pull/18185

ports/zephyr/README.md
```diff
@@ -177,3 +177,47 @@ run the following after you built an image with the previous command:
...
+
+The Zephyr Micropython port provides 2 options for handling filesystems on the device:
+The first is the Micropython filesystem management, which uses Micropython's filesystem code and
+relies on zephyr's FlashArea API, this is enabled by default when 
+`CONFIG_FLASH` and `CONFIG_FLASH_MAP` are turned on.
+The second option is using Zephyr's Filesystem management:
+This relies on Zephyr's File System features and enables sharing access between zephyr and
+micropython code. Several configuration options must be enabled:
+
+    CONFIG_FLASH_MAP=y # Requirement for the file system subsystem
```
dpgeorge, line 222: It looks like requiring this option be enabled, along with `MICROPY_VFS`, will cause the auto-mount code in `main.c:vfs_init()` to be called, which is probably not what you want when using the Zephyr file system.

Not sure how to deal with that, maybe don't auto-mount if `CONFIG_FILE_SYSTEM` is enabled?

ports/zephyr/modzephyr.c
```diff
@@ -86,6 +86,9 @@ static const mp_rom_map_elem_t mp_module_time_globals_table[] = {
     #ifdef CONFIG_FLASH_MAP
     { MP_ROM_QSTR(MP_QSTR_FlashArea), MP_ROM_PTR(&zephyr_flash_area_type) },
     #endif
+    #ifdef CONFIG_FILE_SYSTEM
+    { MP_ROM_QSTR(MP_QSTR_FS), MP_ROM_PTR(&zephyr_fs_type) },
```
dpgeorge: Could this be named `FileSystem` instead?  Or do you think it's OK to abbreviate it to `FS`?

ports/zephyr/zephyr_fs.c
```diff
@@ -0,0 +1,767 @@
...
+#include <string.h>
+#include "py/obj.h"
+#include "py/objstr.h"
+#include "py/runtime.h"
+#include "py/stream.h"
+#include "py/mperrno.h"
+#include "extmod/vfs.h"
+#include "shared/timeutils/timeutils.h"
+
+/* Declare each FSTAB entry */
```
dpgeorge: Please use `//` style comments to match other code.

ports/zephyr/zephyr_fs.c
```diff
@@ -0,0 +1,767 @@
...
+        mp_obj_list_append(list, mp_obj_new_str_from_cstr(zephyr_fs_mounts[i]->mnt_point));
+    }
+
+    return list;
+}
+static MP_DEFINE_CONST_FUN_OBJ_0(zephyr_fs_fstab_fun_obj, zephyr_fs_fstab);
+static MP_DEFINE_CONST_STATICMETHOD_OBJ(zephyr_fs_fstab_obj, MP_ROM_PTR(&zephyr_fs_fstab_fun_obj));
+
+static const mp_rom_map_elem_t zephyr_fs_locals_dict_table[] = {
+    { MP_ROM_QSTR(MP_QSTR_fstab), MP_ROM_PTR(&zephyr_fs_fstab_obj) },
```
dpgeorge: Maybe move this function to the end of this local's dict and add a comment that it's an extension to the base MicroPython VFS set of methods?

ports/zephyr/zephyr_fs.c
```diff
@@ -0,0 +1,767 @@
...
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_1(zephyr_fs_mkfs_obj, zephyr_fs_mkfs);
+#endif
+
+static mp_obj_t zephyr_fs_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, 1, false);
+
+    if (!mp_obj_is_str(args[0])) {
+        mp_raise_ValueError("Argument must be a zephyr FSTAB mountpoint string");
```
dpgeorge: Please use lowercase for the start of error messages.

Also, they need to be wrapped in `MP_ERROR_TEXT(...)`.

ports/zephyr/README.md
VynDragon: As you can see from history this had been setup to fix vfs_init and other vfs things in main, and now to be replaced with python code with https://github.com/micropython/micropython/pull/18185 (so that should go first to remove vfs_init) , I have provided a code snippet with support for this PR there as example, once that pr is merged ill add the snippet here

ports/zephyr/modzephyr.c
VynDragon: It can be whatever, I used FS because it's faster to type, it can be FileSystem, just tell me which you think is better.

ports/zephyr/README.md
dpgeorge: OK, I'll concentrate on #18185 first.

ports/zephyr/zephyr_fs.c
VynDragon: done

ports/zephyr/zephyr_fs.c
VynDragon: done

ports/zephyr/zephyr_fs.c
VynDragon: done

VynDragon:
@dpgeorge Rebased and added zephyr FS to python script, and fixed one littlefs usage issue that appeared out of nowhere.

VynDragon:
Added conversion from zephyr errno to mpy strings so mpy tools (mpremote, viper-ide, etc) works correctly

ports/zephyr/modules/_boot.py
```diff
@@ -54,7 +57,34 @@ def mount_all_disks():
     return retval
 
 
-if FlashArea and create_flash_partition():
+def mount_FS_flash():
```
dpgeorge: I suggest putting this function before `create_flash_partition()`, because that keeps the order they are called in.

ports/zephyr/modules/_boot.py
```diff
@@ -67,6 +97,6 @@ def mount_all_disks():
     sys.path.append(_FLASH_LIB)
 
 # Cleanup globals for boot.py/main.py
-del FlashArea, DiskAccess, zephyr
+del FlashArea, DiskAccess, FS, zephyr
 del sys, vfs, os, const
 del create_flash_partition, mount_all_disks, _FLASH_EXISTS
```
dpgeorge: `del mount_FS_flash`

ports/zephyr/modzephyr.c
dpgeorge: IMO it would be more consistent to call it `FileSystem`, to match the naming of `DiskAccess` and `FlashArea` (which are two fully spelled words).

ports/zephyr/modzephyr.h
```diff
@@ -37,4 +37,10 @@ extern const mp_obj_type_t zephyr_disk_access_type;
 extern const mp_obj_type_t zephyr_flash_area_type;
 #endif
 
+#ifdef CONFIG_FILE_SYSTEM
+extern const mp_obj_type_t zephyr_fs_type;
+size_t mp_zephyr_fs_get_mounts_size(void);
+struct fs_mount_t *const *mp_zephyr_fs_get_mounts(void);
```
dpgeorge: These two declarations are never used, they can be removed.

ports/zephyr/zephyr_fs.c
```diff
@@ -0,0 +1,782 @@
...
+            case 0:             // SEEK_SET
+                err = fs_seek(&self->fp, s->offset, FS_SEEK_SET);
+                break;
+
+            case 1:             // SEEK_CUR
+                err = fs_seek(&self->fp, s->offset, FS_SEEK_CUR);
+                break;
+
+            case 2:             // SEEK_END
+                err = fs_seek(&self->fp, s->offset, FS_SEEK_END);
```
dpgeorge: Suggest simplifying this switch:
```c
MP_STATIC_ASSERT(FS_SEEK_SET == 0);
MP_STATIC_ASSERT(FS_SEEK_CUR == 1);
MP_STATIC_ASSERT(FS_SEEK_END == 2);
err = fs_seek(&self->fp, s->offset, s->whence);
```
Alternatively, use a lookup table:
```c
static const uint8_t WHENCE[] = {FS_SEEK_SET, FS_SEEK_CUR, FS_SEEK_END};
... WHENCE[s->whence]...
```

ports/zephyr/zephyr_filesystem.c
```diff
@@ -0,0 +1,782 @@
...
+        return mp_const_none;
+    }
+
+    mp_obj_tuple_t *t = MP_OBJ_TO_PTR(mp_obj_new_tuple(10, NULL));
+
+    t->items[0] = MP_OBJ_NEW_SMALL_INT(stats.f_bsize);     // f_bsize
+    t->items[1] = MP_OBJ_NEW_SMALL_INT(stats.f_frsize);     // f_frsize
+    t->items[2] = MP_OBJ_NEW_SMALL_INT(stats.f_blocks);     // f_blocks
+    t->items[3] = MP_OBJ_NEW_SMALL_INT(stats.f_bfree);     // f_bfree
+    t->items[4] = t->items[3];     // f_bavail
```
dpgeorge, line 625: I don't think these values are correct.  If I do `mpremote df` I get:
```
$ mpremote df
filesystem     size     used    avail use% mounted on
<FS>          90112      512    89600    0 /flash
```
But the same filesystem with `FlashArea` is:
```
$ mpremote df
filesystem     size     used    avail use% mounted on
<VfsLfs2>   1441792     8192  1433600    0 /flash
```
So the FS implementation here is a factor of 16 too small.

ports/zephyr/zephyr_fs.c
```diff
@@ -0,0 +1,782 @@
...
+    t->items[0] = MP_OBJ_NEW_SMALL_INT(stats.f_bsize);     // f_bsize
+    t->items[1] = MP_OBJ_NEW_SMALL_INT(stats.f_frsize);     // f_frsize
+    t->items[2] = MP_OBJ_NEW_SMALL_INT(stats.f_blocks);     // f_blocks
+    t->items[3] = MP_OBJ_NEW_SMALL_INT(stats.f_bfree);     // f_bfree
+    t->items[4] = t->items[3];     // f_bavail
+    t->items[5] = MP_OBJ_NEW_SMALL_INT(0);     // f_files
+    t->items[6] = MP_OBJ_NEW_SMALL_INT(0);     // f_ffree
+    t->items[7] = MP_OBJ_NEW_SMALL_INT(0);     // f_favail
+    t->items[8] = MP_OBJ_NEW_SMALL_INT(0);     // f_flags
+    t->items[9] = MP_OBJ_NEW_SMALL_INT(2048);     // f_namemax. Value i picked
```
dpgeorge: Please remove the "Value i picked" comment.

Would be good to have a proper value here, but I can see how there might not be a proper value.

dpgeorge:
I see there's a commit with message:

    zephyr: Fix FS config options being ignored.
    
    This makes CONFIG_MICROPY_VFS_* actually do things.

But that reasoning sounds wrong, because those config options are used in `CMakeLists.txt`
```
    MICROPY_VFS_FAT=$<BOOL:${CONFIG_MICROPY_VFS_FAT}>
    MICROPY_VFS_LFS1=$<BOOL:${CONFIG_MICROPY_VFS_LFS1}>
    MICROPY_VFS_LFS2=$<BOOL:${CONFIG_MICROPY_VFS_LFS2}>
```
Is that commit needed at all?

ports/zephyr/zephyr_fs.c
josuah: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/limits.h.html

- `POSIX` picked `PATH_MAX` to 256 bytes min
- `XOPEN` picked `PATH_MAX` to 1024 bytes min

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/limits.h

- Linux picked `PATH_MAX` to 4096 bytes min

https://github.com/search?q=repo%3Azephyrproject-rtos%2Fzephyr%20%22%23define%20PATH_MAX%22&type=code

- Zephyr picks `PATH_MAX` to 256 bytes min, which can be used with `#include <limits.h>`

ports/zephyr/zephyr_filesystem.c
VynDragon: What platform is that on? RP2040? It's very possible this is coming from different settings so would need to be compared to the settings of the DTS instead

ports/zephyr/zephyr_fs.c
VynDragon: is PATH_MAX equivalent to namemax? or better to use'_XOPEN_NAME_MAX' which is 255?

VynDragon:
> Is that commit needed at all?

It didnt seem to apply without that commit (and resulted in mpy fatfs conflicting with zephyr fatfs on names), but could always use a double check

ports/zephyr/zephyr_fs.c
josuah: Woops! Good catch it's `NAME_MAX` like Linux does i.e. [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/libfs.c).

Also, it's about reporting to the python scripts what Zephyr supports, so it looks like `MAX_FILE_NAME` is what Zephyr supports ([[1]](https://github.com/zephyrproject-rtos/zephyr/blob/main/include/zephyr/fs/fs_interface.h#L16-L62), [[2]](https://github.com/zephyrproject-rtos/zephyr/blob/bf0e0b3cad2b317a055712fdb304443857108216/include/zephyr/fs/fs.h#L126)).

Already included from `<zephyr/fs/fs.h>`.

ports/zephyr/zephyr_filesystem.c
VynDragon: Okay so this is funnier than that: mpremote's maths are wrong:

```py
print(_f.format("filesystem", "size", "used", "avail", "use%", "mounted on"))
try:
 _ms = vfs.mount()
except:
 _ms = []
 for _m in [""] + os.listdir("/"):
  _m = "/" + _m
  _s = os.stat(_m)
  if _s[0] & 1 << 14:
   _ms.append(("<unknown>",_m))
for _v,_p in _ms:
 _s = os.statvfs(_p)
 _sz = _s[0]*_s[2]
 if _sz:
  _av = _s[0]*_s[3]
  _us = 100*(_sz-_av)//_sz
  print(_f.format(str(_v), _sz, _sz-_av, _av, _us, _p))
  ```
  
  It takes f_bsize, while it should use f_frsize, eg the *optimal block transfer size* != block size

Comment (VynDragon) [reply ports/zephyr/zephyr_filesystem.c:625]: Actually how tf is fragsize > f_bsize

File: ports/zephyr/modules/_boot.py (hunk 7 above)
Comment (VynDragon) [reply ports/zephyr/modules/_boot.py]: done

File: ports/zephyr/modules/_boot.py (hunk 8 above)
Comment (VynDragon) [reply ports/zephyr/modules/_boot.py]: done

File: ports/zephyr/modzephyr.h (hunk 9 above)
Comment (VynDragon) [reply ports/zephyr/modzephyr.h]: done

File: ports/zephyr/zephyr_fs.c (hunk 10 above)
Comment (VynDragon) [reply ports/zephyr/zephyr_fs.c]: done

File: ports/zephyr/zephyr_filesystem.c (hunk 11 above)
Comment (VynDragon) [reply ports/zephyr/zephyr_filesystem.c:625]: well added a if, but dont know what is happening on zephry side... done

File: ports/zephyr/zephyr_fs.c (hunk 12 above)
Comment (VynDragon) [reply ports/zephyr/zephyr_fs.c]: done

File: ports/zephyr/modzephyr.c (hunk 3 above)
Comment (VynDragon) [reply ports/zephyr/modzephyr.c]: done

**Comment by VynDragon** [comment]
Yea try adding CONFIG_FAT_FILESYSTEM_ELM=y and setting CONFIG_MICROPY_VFS_FAT=y in the .conf of rpi_pico, or removing the commit that enables disabling it.

File: ports/zephyr/zephyr_filesystem.c (hunk 11 above)
Comment (dpgeorge) [reply ports/zephyr/zephyr_filesystem.c:625]: > What platform is that on? RP2040?

Yes, sorry, should have mentioned it was a standard Pico with 2MiB flash, using the `rpi_pico` zephyr board definition from this repo.

File: ports/zephyr/CMakeLists.txt [hunk 13]
```diff
@@ -55,6 +55,7 @@ set(MICROPY_SOURCE_PORT
     uart_core.c
     zephyr_device.c
     zephyr_storage.c
+    zephyr_fs.c
```
dpgeorge: suggest calling the file `zephyr_filesystem.c` to match the type name

ports/zephyr/README.md
```diff
@@ -208,3 +208,47 @@ run the following after you built an image with the previous command:
...
+			cache-size=<4096>;
+			lookahead-size=<32>;
+			block-cycles=<4>;
+		};
+	};
+	
+It is then possible to use the FS like a normal Micropython filesystem:
+
+    import vfs, zephyr
+    zfs = zephyr.FS(zephyr.FS.fstab()[0])
```
dpgeorge: please update to use `zephyr.FileSystem`

ports/zephyr/modules/_boot.py
```diff
@@ -11,12 +11,39 @@
...
+# zephyr.FileSystem depends on CONFIG_FILE_SYSTEM and CONFIG_FLASH_MAP
+FS = getattr(zephyr, "FileSystem", None)
+
 _FLASH = const("/flash")
 _FLASH_LIB = const("/flash/lib")
 _STORAGE_KEY = const("storage")
 _FLASH_EXISTS = False
 
 
+def mount_FS_flash():
```
dpgeorge: suggest calling this function `mount_filesystem` (we don't really mix lowercase and uppercase in symbol names)

ports/zephyr/zephyr_fs.c
```diff
@@ -0,0 +1,782 @@
...
+static MP_DEFINE_CONST_DICT(zephyr_fs_locals_dict, zephyr_fs_locals_dict_table);
+
+static const mp_vfs_proto_t zephyr_fs_proto = {
+    .import_stat = zephyr_fs_import_stat,
+};
+
+const mp_obj_type_t zephyr_fs_type;
+
+MP_DEFINE_CONST_OBJ_TYPE(
+    zephyr_fs_type,
```
dpgeorge: suggest calling this new type `zephyr_filesystem_type`, to make it unambiguous, and to match the Python-level name

ports/zephyr/zephyr_filesystem.c
dpgeorge: This now works for me, I get:
```
$ mpremote df
filesystem     size     used    avail use% mounted on
<FileSystem>  1441792     8192  1433600    0 /flash
```

ports/zephyr/zephyr_fs.c
VynDragon: done

ports/zephyr/modules/_boot.py
VynDragon: done

ports/zephyr/README.md
VynDragon: done

ports/zephyr/CMakeLists.txt
VynDragon: done

dpgeorge (approved):
Thanks for updating, this is looking good now.

PR #17088, opened 2025-04-07
cpydiff: Document differences in errno and OSError.

AJMansfield:
### Summary

This PR documents the lack of OSError subtype dispatch.

This PR also includes another minor cpydiff I discovered while testing the former, to highlight the difference in the name for Errno 95: EOPNOTSUPP vs ENOTSUP.

### Testing

I've tested these cpydiffs on both unix and rp2, and can confirm they detect the difference in question. No other existing cpydiffs that I've been able to find appear to cover this difference.

### Trade-Offs

The latter test is technically redundant, as the difference it reports also occurs in the diff output of the former -- but it's also much more succinct and communicates the difference it checks for much more clearly.

### References
- [PEP 3151 – Reworking the OS and IO exception hierarchy](https://peps.python.org/pep-3151/)
- https://github.com/python/cpython/blob/e80ed2cf75384bab09e70f0f04d17e92183fdf12/Objects/exceptions.c#L2093-L2103
- python/cpython#107486
- https://docs.python.org/3/library/errno.html#errno.ENOTSUP

### DIFF
--- a/tests/cpydiff/modules_errno_enotsup.py
+++ b/tests/cpydiff/modules_errno_enotsup.py
@@ -1,0 +1,10 @@
+"""
+categories: Modules,errno
+description: MicroPython does not include ``ENOTSUP`` as a name for errno 95.
+cause: MicroPython does not implement the ``ENOTSUP`` canonical alias for ``EOPNOTSUPP`` added in CPython 3.2.
+workaround: Use ``errno.EOPNOTSUPP`` in place of ``errno.ENOTSUP``.
+"""
+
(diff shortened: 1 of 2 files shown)


AJMansfield:
See also:
- python/cpython#107486
- https://github.com/python/cpython/blob/e80ed2cf75384bab09e70f0f04d17e92183fdf12/Objects/exceptions.c#L2093-L2103

dpgeorge (approved):
Thanks!

PR #16141, opened 2024-11-03
ports/unix: Add raw REPL support.

glenn20:
### Summary

Add raw REPL support to the unix port. This make it possible to connect to a unix micropython instance using mpremote and utilise the functionality of the raw REPL. This is useful for debugging and CI tests of mpremote and related tools.

For example, one can run micropython behind a PTY with `socat` and connect to the slave pty (as if it were a serial port) with `mpremote`:

```bash
socat PTY,link=$HOME/.upy-pty,rawer EXEC:./build-standard/micropython,pty,stderr,onlcr=0 &
mpremote resume connect ~/.upy-pty ls
```
(The `resume` argument is required since micropython v1.24 as a soft reset terminates the micropython process).

or, using the micropython docker image (thanks to @mattytrentini ):

```bash
socat PTY,link=$HOME/.upy-pty,rawer EXEC:"docker run -ti --rm micropython/unix\:v1.23.0",pty,stderr,onlcr=0
```

### Testing

This has been tested using private CI workflows adapted from those in the [`mpremote-path`](https://github.com/glenn20/mpremote-path) tool (which uses `mpremote` as a library).

### Trade-offs and Alternatives

The changes impact only the unix port (adding `pyexec.c` to the build) and present a user-visible change through behaviour when a ctrl-A character is entered at the repl. However, this reflects expected behaviour on other ports and should not impact on expected or supported usage.

PR #17089, opened 2025-04-07
mpremote/tests: Rewrite test runner to run correctly under Windows.

AJMansfield:
### Summary

This PR rewrites the test runner script used for testing mpremote to run correctly under Windows, and adds some functionality present in the main test runner that it was previously missing.

This PR adds inheritance behavior for the `TESTS_DIR` and `MPREMOTE` variables in the `run-mpremote-tests.sh` test script. It adds a new `RESULT_DIR` variable that can be specified to direct the runner for a path to place output `.exp` and `.out` files. It expands the use of the input arguments to allow specifying multiple test scripts.

Among other things, this rewrite also enables multiple concurrent instances of the test runner to be run against multiple simultaneously-connected target boards, via:

```bash
SERIAL_PORT="/dev/serial/by-id/usb-MY_BOARD"
MPREMOTE="./mpremote.py connect ${SERIAL_PORT}" RESULT_DIR="tests/results/MY_BOARD" tests/run-mpremote-tests.sh
```

Previously, the test runner had no way to override its automatic serial port selection, and concurrent instances had no way to specify different output paths to avoid clobbering each other's `.out` files.

### Testing

I've tested these changes manually and as part of the local automations I've been using to test #16994. I can confirm they address the configurability and concurrency issues described above.

For the moment I'm relying on @Josverl to report on whether this rewrite runs appropriately under Windows implementations of bash.

Some attention should probably be paid to adapting this to also run against locally-runnable ports, so that the mpremote tests can be added to automated CI/CD.


Josverl:
## Windows - Git Bash 
Some test cases pass :

./test_eval_exec_run.sh: OK
./test_recursive_cp.sh: OK
./test_resume.sh: OK

all others fail 

Tests fail when 
- uses {$TMP} or assumes a host path convention 
- comparison of a sha256sum is slightly different with an additional asterix.
   `The default mode is to print a line with checksum, a space, a character indicating input mode ('*' for binary,` ' ' for text or where binary is insignificant), and name for each FILE`
    > can be mitigated with 'sha256sum -t' 

- the ramdisk creation in `test_mip_local_install.sh` fails while the one in `test_filesystem.sh` works. 
   it is trying to create file in a location where it should not attempt write at all. 
   ```
      > cp ${TMP}/package :
      ./test_mip_local_install.sh: FAIL
      2,3c2,11
      < mkdir :C:/Program Files/Git/__ramdisk/lib
      < mpremote: mkdir: C:/Program Files/Git/__ramdisk/lib: No such file or directory.
   ```
   I would suggest `ramdisk.py` that should not be created on the fly, but be part of the testsuite and referenced from both scripts.


<details><summary>Details</summary>
<p>

EUROPE+josverl@josverl-sb5 MINGW64 /d/mypython/micropython/tools/mpremote/tests (mpremote-testfix)
$ ./run-mpremote-tests.sh 
./test_eval_exec_run.sh: OK
./test_filesystem.sh: FAIL
11c11
< e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855 *-
---
> e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  -
13,15c13,15
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :b.py
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :c.py
---
> cp ${TMP}/a.py :
> cp ${TMP}/a.py :b.py
> cp ${TMP}/a.py :c.py
30c30
< 50f0a701dd6cd6125387b96515300c9d5294c006518f8e62fa9eea3b66587f21 *-
---
> 50f0a701dd6cd6125387b96515300c9d5294c006518f8e62fa9eea3b66587f21  -
34,35c34,35
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :aaa
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :bbb/b.py
---
> cp ${TMP}/a.py :aaa
> cp ${TMP}/a.py :bbb/b.py
41c41
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :aaa
---
> cp ${TMP}/a.py :aaa
45c45
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :aaa/
---
> cp ${TMP}/a.py :aaa/
47c47
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/a.py :aaa/a.py/
---
> cp ${TMP}/a.py :aaa/a.py/
66d65
< sed: cannot rename ./sed3Pkd5W: Invalid cross-device link
69,70c68,69
< 50f0a701dd6cd6125387b96515300c9d5294c006518f8e62fa9eea3b66587f21
69,70c68,69
< 50f0a701dd6cd6125387b96515300c9d5294c006518f8e62fa9eea3b66587f21
< Hello
---
> 612c7ddb88390ac86b4174b26a6e5b52fc2f2838b234efd8f6f7c41631a49d04
> Goodbye
< 50f0a701dd6cd6125387b96515300c9d5294c006518f8e62fa9eea3b66587f21
< Hello
---
> 612c7ddb88390ac86b4174b26a6e5b52fc2f2838b234efd8f6f7c41631a49d04
> Goodbye
< Hello
---
> 612c7ddb88390ac86b4174b26a6e5b52fc2f2838b234efd8f6f7c41631a49d04
> Goodbye
> 612c7ddb88390ac86b4174b26a6e5b52fc2f2838b234efd8f6f7c41631a49d04
> Goodbye
73c72
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
> Goodbye
73c72
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
73c72
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
> cp ${TMP}/package :
86c85
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :package2
---
> cp ${TMP}/package :
86c85
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :package2
> cp ${TMP}/package :
86c85
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :package2
---
> cp ${TMP}/package :package2
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :package2
---
> cp ${TMP}/package :package2
---
> cp ${TMP}/package :package2
100c99
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :test
> cp ${TMP}/package :package2
100c99
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :test
---
100c99
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :test
---
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :test
---
> cp ${TMP}/package :test
111c110
---
> cp ${TMP}/package :test
111c110
> cp ${TMP}/package :test
111c110
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :test/package2
---
111c110
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :test/package2
---
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :test/package2
---
> cp ${TMP}/package :test/package2
---
> cp ${TMP}/package :test/package2
> cp ${TMP}/package :test/package2
123c122
< cp :test/package C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/copy
123c122
< cp :test/package C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/copy
---
> cp :test/package ${TMP}/copy
136c135
< cp :test/package C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/copy
---
> cp :test/package ${TMP}/copy
136c135
---
> cp :test/package ${TMP}/copy
136c135
< cp :test/package C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/copy/package2
---
> cp :test/package ${TMP}/copy/package2
150c149
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
< cp :test/package C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/copy/package2
---
> cp :test/package ${TMP}/copy/package2
150c149
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
150c149
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
> cp ${TMP}/package :
---
> cp ${TMP}/package :
> cp ${TMP}/package :
163c162
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
> cp ${TMP}/package :
> cp ${TMP}/package :
179c178
< cp C:/Users/josverl/AppData/Local/Temp/tmp.ryMYe3dGdU/package :
---
> cp ${TMP}/package :
./test_mip_local_install.sh: FAIL
2,3c2,11
< mkdir :C:/Program Files/Git/__ramdisk/lib
< mpremote: mkdir: C:/Program Files/Git/__ramdisk/lib: No such file or directory.
---
> mkdir :/__ramdisk/lib
>
> ---- Install package
> Install ${TMP}/example/package.json
> Installing: /__ramdisk/lib/mip_example/__init__.py
> Installing: /__ramdisk/lib/mip_example/hello.py
> Done
>
> ---- Test package
> Hello, world!
./test_mount.sh: FAIL
4c4
< Local directory C:/Users/josverl/AppData/Local/Temp/tmp.zjhJcEqIxq is mounted at /remote
---
> Local directory ${TMP} is mounted at /remote
6c6
< Local directory C:/Users/josverl/AppData/Local/Temp/tmp.zjhJcEqIxq is mounted at /remote
---
> Local directory ${TMP} is mounted at /remote
10c10
< Local directory C:/Users/josverl/AppData/Local/Temp/tmp.zjhJcEqIxq is mounted at /remote
---
> Local directory ${TMP} is mounted at /remote
./test_recursive_cp.sh: OK
./test_resume.sh: OK

</p>
</details>

AJMansfield:
@Josverl could you test the latest push? I've now majorly rewritten most of the test runner script with far more paranoid/maximalist quoting, and updated it to use a separate results directory to match the behavior of the main test suite (with a `.gitignore` line matching the way the main test suite gets its results ignored).

AJMansfield:
I'm not sure what the best way to address the sha256sum issue is here, since I'm not sure what the root cause is.
The thing being compared by that hash is ultimately whatever copy of `a.py` the remote ended up with.

The file being created in that heredoc line is still LF-terminated; even running the same snippet under Windows Git Bash, those line endings still aren't being written to said file as CRLF.

```bash
cat << EOF > "${TMP}/a.py"
print("Hello")
print("World")
EOF
```

I suspect that the hash mismatches here may represent an actual mpremote bug -- i.e. that mpremote is converting LFs in an input .py file into CRLFs somewhere between reading that file from windows and writing the file out on the remote micropython.

AJMansfield:
I really need to spend some more time getting my Windows build environment working so we're not just playing telephone on this.

> There is a common patter to setup a ramdisk. This should be a common fixture. Instead it is setup in multiple slightly different ways , and only one (the filesystem test ) works on Windows.

I'll try to see about refactoring this.

> I think the bash syntax: `${target@Q} ` does not work or work the same. Simpler to just use the correct delimiters instead.

You're right that that that didn't fix that, but it's not actually that, looking again it's actually the missing `r` prefix to get python to not interpret the backslashes as escapes Theoretically [77b5a04](https://github.com/micropython/micropython/pull/17089/commits/77b5a0498491ef218c25486070eb263a42f6b5e0) should've removed those backslashes, but I'll still add the `r` anyway.

> I think the diffing on errors would be better to understand when `$DIFF "${result}.exp" "${result}.out" ` That shows the extra's as + and the missing lines as - , which makes more sense to my simple Dutch brain.

Good catch, the other test drivers diff also diff them in the order you suggest so that'll make it more consistent.

> 🪲While debugging I found this bug 🦗 Git bash + mpremote result in a transport error on `mpremote resume mkdir /ramdisk/lib` ( no leading :) #17093

Could you run that command with `set -x` enabled? That'll print a trace as commands run that includes the final interpolated value of everything.

> just testing on Git bash - does not appear to cover enough ( id say that about 2% of windows users use git bash for mpremote) Even with a perfect bash script , that would cover only about 20% of the [end user computing scenarios](https://gs.statcounter.com/os-market-share/desktop/worldwide/).

Agreed, this should also at least support `.ps1` test cases, and probably `.cmd` test cases too.
This might also make it easier to test windows-y behaviors, too, since there's a port of powershell for linux it could use.

> Personally I think that test are best written using a test framework, such as pytest. And that can handle .exp files, and has much more flexibility and parametrization and fuzzing capabilities, code coverage etc etc that will take so much more time to try to accomplish using bash, or any other shell language for that matter.

Almost definitely. There's a saying that, if you're using arrays in bash, it's time to rewrite it in a different language... and the rewrite most definitely uses arrays lol.

As a command-line tool, it makes sense to have the test case files themselves in shell languages, but I do think it'll be useful to rewrite the test runner in python so it can reuse and integrate with parts of the other test runners.

Josverl:
Is this ready for another test run?

AJMansfield:
> Is this ready for another test run?

Not really, no; just rebased this in order to more easily experiment with #12802 --- since its patch theoretically enables raw repl mode on the unix port, and might make it possible to run the mpremote tests as part of CI.

AJMansfield:
@Josverl Just rewrote the test driver to use more-or-less the same test code as the main suite's `run-tests.py`. 

I've tested it against my own boards and can confirm it works correctly on unix; and since it's now just doing exactly what the main test suite does it _should_ be compatible with windows too. Theoretically, hopefully you can confirm this.

It now also looks for and tries to run `test_*.bat` and `test_*.ps1` scripts in addition to `test_*.sh` scripts, though I've not been able to test that just yet.

Still need to test the merge with #12802 or #16141 to confirm that the micropython socket server it sets up actually works though, and add the relevant CI jobs for it.

On a separate note, does the Windows port of micropython support raw repl mode? If it's possible to add an mpremote CI job for windows that'd be pretty useful to have too, given that it's been one of the tool's main pain points.

Josverl:
Thanks ,
Ill give it a go later today or tomorrow. I'm away from home but have a few boards to test.

> does the Windows port of micropython support raw repl mode?

I do not know

tools/mpremote/tests/run-mpremote-tests.py
```diff
@@ -0,0 +1,479 @@
...
+        if os.path.exists(results_file):
+            with open(results_file, "r") as f:
+                tests = list(test[0] for test in json.load(f)["results"] if test[1] == "fail")
+        else:
+            tests = []
+    elif len(args.files) == 0:
+        test_extensions = ("test_*.sh",)
+        if sys.platform == "nt":
+            test_extensions += (
+                "test_*.bat",
```
Josverl: I would not bother with .bat, that is something from the previous century. if you insist on legendary support use .cmd, on Windows prefer .ps1 over anything older.

tools/mpremote/tests/run-mpremote-tests.py
```diff
@@ -0,0 +1,479 @@
...
+Options -i and -e can be multiple and processed in the order given. Regex
+"search" (vs "match") operation is used. An action (include/exclude) of
+the last matching regex is used:
+  run-tests.py -i async - exclude all, then include tests containing "async" anywhere
+  run-tests.py -e '/big.+int' - include all, then exclude by regex
+  run-tests.py -e async -i async_foo - include all, exclude async, yet still include async_foo
+""",
+    )
+    cmd_parser.add_argument(
+        "-t", "--test-instance", default="unix", help="the MicroPython instance to test"
```
Josverl, line 356: Is this the micropython executable ? ( micropython/micropython.exe) 
what other values can be used / default to current platform ?
How is this used in the mpremote tests ?

AJMansfield: This works almost the same way as running the test harness via `run-tests.py`, with the only difference being that the value is passed through for mpremote to interpret, rather than being interpreted by the pyboard functions used in the other test runners.

i.e. `run-mpremote-tests.py -t whatever` results in running each test script with the `MPREMOTE` environment variable set to `tools/mpremote/mpremote.py connect whatever` instead of just `tools/mpremote/mpremote.py`, turning the `$MPREMOTE test_command` invocations into `tools/mpremote/mpremote.py connect whatever test_command`.

The value `"unix"` is special, and causes the test script to instead launch a separate micropython process for each test using socat to open a pty or server socket, and then pass that pty path or socket address to mpremote instead. `MPREMOTE` becomes `tools/mpremote/mpremote.py connect socket://127.0.0.1:12345` or `tools/mpremote/mpremote.py connect /tmp/mpremote.AAAAAAA/pty`. (The micropython binary to launch in this case is taken from the `MICROPY_MICROPYTHON` environment variable --- defaulting to unix-standard if that's missing, consistent with the other test runners.)

A blank or whitespace-only value results in omitting this `connect <target>` part entirely, and just include the `mpremote.py` path, which IIRC is equivalent to passing `auto`.

tools/mpremote/tests/run-mpremote-tests.py
AJMansfield: Makes sense!

Also, just added a powershell test!

Not proud of it though, but it's the only thing I could even get to work in the version of powershell I have on my machine:

```powershell
#!/snap/bin/pwsh -File
iex "/usr/bin/python $Env:MPREMOTE eval 1+1"
```

I've no idea how to avoid that very non-portable `/usr/bin/python` part; but without it, the test just complains `/usr/bin/env: 'python3': No such file or directory` about the shebang in `mpremote.py`. Though editing `mpremote.py` to just use `#!python` as its own shebang _does_ manage to work...

Josverl:
I spend some effort to try to port a single test to windows 
- need to copy / change the .sh test
- extract common micropython.py test fixtures to reduce duplication 
- create a new some_test.ps1.exp file 
further 
- add an alternative to socat / PTY to the test script, while there is a ConPTY on windows 10+ , however you end up in WSL2 , which is not what we need at all.

I wonder if this is the correct direction.

AJMansfield:
> * add an alternative to socat / PTY to the test script, while there is a ConPTY on windows 10+ , however you end up in WSL2 , which is not what we need at all.

Tbh that'd probably be fine --- mpremote is the code under test here, not really the rest of micropython; only mpremote itself needs to run within windows here. For now, test it against hardware with `-t COM<x>` or `-t auto` or `-t ' '` --- but I'm already considering replacing `socat` with docker or qemu or something, just to isolate micropython's filesystem.

Josverl:
Mpflash just runs Mpremote in a sub process with some minor threading and tenacity to handle time outs and retries. Pure python, and Works reliable across platforms

PR #17090, opened 2025-04-07
mpremote: Refactor error handling to apply generically to any errno.

AJMansfield:
### Summary

This rewrites the code that previously manually emitted and caught various `OSError` subclasses with equivalent code that uses the errno name dictionary to do this generically, and updates the exception handler in `do_filesystem` to catch them in a similarly-generic fashion using `os.strerror` to retrieve an appropriate error message text equivalent to the current messages.

Note that in the CPython environments where mpremote runs, the call to the `OSError` constructor _already returns an instance of the corresponding mapped exception subtype_.

That is, this code:

```python
raise PermissionError(errno.EPERM, '')
```

is _exactly_ equivalent to:

```python
raise OSError(errno.EPERM, '')
```
 
i.e. they both raise a `PermissionError(1, '')` error.

(See also: #17088, documenting this difference with micropython.)

### Testing

This PR adds an automated test script, `test_errno.sh`. It includes test cases for all of the errnos that were previously special-cased and for raw TransportErrors not wrapping any OSError.

I've tested running the full mpremote test suite, including this new script against rp2040 and rp2350 targets, and have performed additional manual tests of all of the existing exception cases to make sure they were covered.

Manual testing also led to the discovery of the [modules_errno_enotsup.py](https://github.com/micropython/micropython/pull/17088/files#diff-ecb9788a74be865a5c79ff47ebda69bffb0ca106d8cf675e51586345b2fc4231) cpydiff submitted in #17088. I've also added a test to ensure mpremote correctly handles this corner case.

This PR removes the conversion from `errno.ENODEV` into `FileNotFoundError`; see https://github.com/micropython/micropython/pull/17090#discussion_r2049376106 for a discussion of the rationale.

This PR subtly changes the spelling of some of the error messages that get reported when filesystem operations fail. This may break external tools that depend on those specific spellings may break; such usages arguably deserve to break, but this change also inadvertently makes mpremote compatible with any other pre-existing jank that expects the CPython error spellings.

Many of the errors this code scans for in the mpremote output are ones that micropython could never generate. This is a mild performance pessimization. I previously suspected that this might also broaden a correctness issue around injecting errno symbol names into file paths, however I've tested and verified that no such vulnerability exists.

### DIFF
--- a/tools/mpremote/mpremote/commands.py
+++ b/tools/mpremote/mpremote/commands.py
@@ -393,12 +393,8 @@ def do_filesystem(state, args):
                     )
                 else:
                     do_filesystem_cp(state, path, cp_dest, len(paths) > 1, not args.force)
-    except FileNotFoundError as er:
-        raise CommandError("{}: {}: No such file or directory.".format(command, er.args[0]))
-    except IsADirectoryError as er:
-        raise CommandError("{}: {}: Is a directory.".format(command, er.args[0]))
-    except FileExistsError as er:
-        raise CommandError("{}: {}: File exists.".format(command, er.args[0]))
+    except OSError as er:
+        raise CommandError("{}: {}: {}.".format(command, er.strerror, os.strerror(er.errno)))
     except TransportError as er:
         raise CommandError("Error with transport:\n{}".format(er.args[0]))
 
--- a/tools/mpremote/mpremote/transport.py
+++ b/tools/mpremote/mpremote/transport.py
@@ -56,7 +56,10 @@ listdir_result = namedtuple("dir_result", ["name", "st_mode", "st_ino", "st_size
 # raises it as the corresponding OSError-derived exception.
 def _convert_filesystem_error(e, info):
     if "OSError" in e.error_output:
-        for code, estr in errno.errorcode.items():
+        for code, estr in [
+            *errno.errorcode.items(),
+            (errno.EOPNOTSUPP, "EOPNOTSUPP"),
+        ]:
             if estr in e.error_output:
                 return OSError(code, info)
     return e
--- a/tools/mpremote/mpremote/transport.py
+++ b/tools/mpremote/mpremote/transport.py
@@ -55,18 +55,10 @@ listdir_result = namedtuple("dir_result", ["name", "st_mode", "st_ino", "st_size
 # Takes a Transport error (containing the text of an OSError traceback) and
 # raises it as the corresponding OSError-derived exception.
 def _convert_filesystem_error(e, info):
-    if "OSError" in e.error_output and "ENOENT" in e.error_output:
-        return FileNotFoundError(info)
-    if "OSError" in e.error_output and "EISDIR" in e.error_output:
-        return IsADirectoryError(info)
-    if "OSError" in e.error_output and "EEXIST" in e.error_output:
-        return FileExistsError(info)
-    if "OSError" in e.error_output and "ENODEV" in e.error_output:
-        return FileNotFoundError(info)
-    if "OSError" in e.error_output and "EINVAL" in e.error_output:
-        return OSError(errno.EINVAL, info)
-    if "OSError" in e.error_output and "EPERM" in e.error_output:
-        return OSError(errno.EPERM, info)
+    if "OSError" in e.error_output:
+        for code, estr in errno.errorcode.items():
+            if estr in e.error_output:
+                return OSError(code, info)
     return e
 
 
--- a/tools/mpremote/tests/test_errno.sh
+++ b/tools/mpremote/tests/test_errno.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+set -e
+
+# Special ErrorFS so the test can induce arbitrary filesystem errors.
+cat << EOF > "${TMP}/fs.py"
+import os, vfs, errno
+
+class ErrorFS:
+    def mount(self, *a, **k):
+        pass
+    def umount(self, *a, **k):
+        pass
+    def chdir(self, *a, **k):
+        pass
+    def open(self, *a, **k):
+        raise self.error
+
+fs = ErrorFS()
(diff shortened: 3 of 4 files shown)


tools/mpremote/mpremote/transport.py
```diff
@@ -55,14 +56,10 @@ def __init__(self, status_code, error_output):
...
 # raises it as the corresponding OSError-derived exception.
 def _convert_filesystem_error(e, info):
-    if "OSError" in e.error_output and "ENOENT" in e.error_output:
-        return FileNotFoundError(info)
-    if "OSError" in e.error_output and "EISDIR" in e.error_output:
-        return IsADirectoryError(info)
-    if "OSError" in e.error_output and "EEXIST" in e.error_output:
-        return FileExistsError(info)
-    if "OSError" in e.error_output and "ENODEV" in e.error_output:
-        return FileNotFoundError(info)
```
dpgeorge, line 65: It looks like this one is not automatically converted, ie:
```py
OSError(errno.ENODEV, 'filename')
```
just returns:
```py
OSError(19, 'filename')
```
So maybe it needs to be handled explicitly?

tools/mpremote/mpremote/commands.py
```diff
@@ -367,14 +367,8 @@ def do_filesystem(state, args):
...
                 else:
                     do_filesystem_cp(state, path, cp_dest, len(paths) > 1, not args.force)
-    except FileNotFoundError as er:
-        raise CommandError("{}: {}: No such file or directory.".format(command, er.args[0]))
-    except IsADirectoryError as er:
-        raise CommandError("{}: {}: Is a directory.".format(command, er.args[0]))
-    except FileExistsError as er:
-        raise CommandError("{}: {}: File exists.".format(command, er.args[0]))
-    except TransportError as er:
-        raise CommandError("Error with transport:\n{}".format(er.args[0]))
```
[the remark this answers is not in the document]
AJMansfield: Good catch! I'll see if I can add a test case for this regression too.

tools/mpremote/mpremote/transport.py
dpgeorge: > I'm of the opinion that the conversion from `ENODEV` into `FileNotFoundError` ought to be _removed_.

Yes, that sounds reasonable to me.

AJMansfield:
I've added the TransportError handler back in, and a test script to verify it and everything else.

dpgeorge:
This looks good now, thank you.

I have some feedback on the test, which is just a suggestion, not necessary to get this merged: it's possible, and simpler in the case at hand, to define an actual filesystem in Python, rather than a block device.  And then make that filesystem raise the relevant exception.  Eg something like this (untested!):
```py
class FS:
    def setup_error(self, er):
        self._cur_error = er
    def mount(self, readonly, mkfs):
        pass
    def chdir(self, path):
        pass
    def open(self, path, mode):
        raise self._cur_error

fs = FS()
vfs.mount(fs, "/fs")
setup_error = fs.setup_error
```

AJMansfield:
> It's possible, and simpler in the case at hand, to define an actual filesystem in Python, rather than a block device. And then make that filesystem raise the relevant exception.

Good call on that, that'd definitely make the test shorter and less clumsy; I'll implement this and retest.

(Also, I should probably test this under Windows first too, just to be sure, now that I have an environment for that set up.)

AJMansfield:
I've updated the test with the simplification suggested -- which also let me replace the `touch` commands with even-more-noopish `cat` commands.

I've re-tested the suite again with a WSL2 Linux mpremote host and Rpi Pico2 device and can confirm the new test discriminates regressions on all issues previously discussed; tests from my Windows 11 mpremote host forthcoming momentarily.

AJMansfield:
Running the test from a Windows 11 mpremote host (using an MSYS2 bash shell) against the same RPi Pico2 device has revealed an interesting deviation:

```patch
--- test_errno.sh.exp   2025-04-24 10:02:53.218824300 -0400
+++ test_errno.sh.out   2025-04-24 10:02:57.736379400 -0400
@@ -20,6 +20,6 @@
 mpremote: cat: EPERM.py: Operation not permitted.
 expect error
 -----
-mpremote: cat: EOPNOTSUPP.py: Operation not supported.
+mpremote: cat: EOPNOTSUPP.py: Unknown error.
 expect error
 -----
```

It seems that `os.strerr` doesn't include a specific error message for `EOPNOTSUPP` under Windows (at least, not in Python 3.12.10, the version I tested with).

The fact there's a platform difference in what set of errno symbols is available leads me to think it may be better to rethink searching the error text for `EWHATEVER` symbol strings, and parse for the `[Errno xx]` part of the error text instead...

AJMansfield:
After a couple more hours working through other possible parsing strategies, I'm forced to conclude that trying to parse all the different ways micropython might output an OSError trace would go well outside the scope of this PR.

As proposed, this operates correctly on the subset of errors that the commands its used for can produce; let's just let that be sufficient, even if it does technically break when you artificially inject more pathological error conditions.

Josverl:
I was looking at test coverage for mpremote ( which is a hassle due to bash, but possible ) and noticed that the test cases do not exercise all errors that could be thrown.

image
I assume that you now have these covered by the additional test that you added.

mpremote is currently at ~55% test coverage. Ill work on a PR for coverage test and reporting, always nice to push the dial up.

dpgeorge (approved):

dpgeorge:
> mpremote is currently at ~55% test coverage. Ill work on a PR for coverage test and reporting, always nice to push the dial up.

Sounds good!

dpgeorge:
> As proposed, this operates correctly on the subset of errors that the commands it's used for can produce; let's just let that be sufficient, even if it does technically break when you artificially inject more pathological error conditions.

Yes, good to be pragmatic.  Now merged.

PR #17091, opened 2025-04-07
tools/mpy_ld.py: Resolve fixed-address symbols if requested.

agatti:
### Summary

This PR lets mpy_ld.py resolve symbols not only from the object files involved in the linking process, or from compiler supplied static libraries, but also from a list of symbols referenced by an absolute address (usually provided by the system's ROM).

This is needed for Xtensa/LX106 targets as some C stdlib functions are provided by the MCU's own ROM code to reduce the final code footprint.  Given that there are different methods to reference external symbols for different processors, this change is currently limited to the Xtensa target and it and it will need further modifications to be adapted to other targets.

Natmod builds can set the `EXTERN_SYMS` variable pointing to a file containing a series of symbols and their absolute address if they want to provide external symbols to the resolution process.  Each line must contain two whitespace-separated fields, the name and its address (which will be parsed as an hexadecimal number).  Empty lines are ignored, and lines starting with a "#" symbol will be treated as comment lines and thus ignored as well.

Future work will be needed when the ESP32-C2 MCU support is brought in as some code from libgcc and newlib is moved into ROM on that platform as well (see ESP-IDF's linker files in `components/esp_rom/esp32c2/ld/` - `newlib`, `newlib-nano`, and `libgcc`).

All natmods are now built for the Xtensa architecture as part of the CI build process.

### Testing

Regular natmod test were ran on a NodeMCU v2 ESP8266 board by injecting the following code block when importing natmods on the board (see https://github.com/micropython/micropython/issues/14430#issuecomment-2332648018):

```py
import esp, gc
gc.collect()
esp.set_native_code_location(200 * 4096, 30 * 4096)
```

To achieve this a new argument to `tests/run-natmodtests.py` is introduced so the above fragment is uploaded as part of the testing process.  All tests were executed successfully with `./run-natmodtests.py -p -d /dev/ttyUSB0 -a xtensa -b xtensa_prelude.py extmod/<testfiles>` with the following exceptions (`xtensa_prelude.py` is the extra code fragment being injected):

`extmod/btree1.py`: FAIL ("MemoryError: memory allocation failed, allocating 20734 bytes")
`extmod/btree_closed.py`: FAIL ("MemoryError: memory allocation failed, allocating 20734 bytes")
`extmod/btree_error.py`: FAIL ("MemoryError: memory allocation failed, allocating 20734 bytes")
`extmod/btree_gc.py`: FAIL ("MemoryError: memory allocation failed, allocating 20734 bytes")
`extmod/deflate_compress.py`: FAIL ("Error: memory allocation failed")

The btree natmod probably is a bit too much for an ESP8266 in its current form, probably the same is valid for the deflate natmod when it comes to compressing buffers.

All `tools/mpy_ld.py` changes are active only for the `EM_XTENSA` architecture and they are opt-in, so there shouldn't be any trade-offs to be made here.  As far as alternatives go, the only one is to move natmod code into one or more usermods and bring them in as part of the MicroPython build.

For `tests/run-natmodtests.py` the same applies (as in the injection being opt-in), although it is available to all platforms.

### DIFF
--- a/ports/esp8266/main.c
+++ b/ports/esp8266/main.c
@@ -49,6 +49,64 @@
 
 static char heap[38 * 1024];
 
+#if MICROPY_HW_HARD_FAULT_DEBUG
+
+static void format_hex(uint32_t hex, char *buffer) {
+    static const char table[] = { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E', 'F' };
+    int offset = 7;
+    uint32_t value = hex;
+    while (offset >= 0) {
+        buffer[offset--] = table[value & 0x0F];
+        value >>= 4;
+    }
+}
+
+static void print_reset_info(void) {
+    struct rst_info *rst_info = system_get_rst_info();
+    if ((rst_info->reason == REASON_WDT_RST) || (rst_info->reason == REASON_EXCEPTION_RST) || (rst_info->reason == REASON_SOFT_WDT_RST)) {
+        char buffer[8];
+        mp_hal_stdout_tx_str("\r\n\r\nThe system restarted due to an error.\r\n\r\nReason: ");
+        switch (rst_info->reason) {
+            case REASON_WDT_RST:
+                mp_hal_stdout_tx_str("WDT");
+                break;
+
+            case REASON_EXCEPTION_RST:
+                mp_hal_stdout_tx_str("EXCEPTION");
+                break;
+
+            case REASON_SOFT_WDT_RST:
+                mp_hal_stdout_tx_str("SOFT_WDT");
+                break;
+
+            default:
+                assert(!"Should not ever get here.");
+                break;
+        }
+        mp_hal_stdout_tx_str(" Cause: ");
+        format_hex(rst_info->exccause, buffer);
+        mp_hal_stdout_tx_strn(buffer, sizeof(buffer));
+        mp_hal_stdout_tx_str(" EPC1: ");
+        format_hex(rst_info->epc1, buffer);
+        mp_hal_stdout_tx_strn(buffer, sizeof(buffer));
+        mp_hal_stdout_tx_str(" EPC2: ");
+        format_hex(rst_info->epc2, buffer);
+        mp_hal_stdout_tx_strn(buffer, sizeof(buffer));
+        mp_hal_stdout_tx_str(" EPC3: ");
+        format_hex(rst_info->epc3, buffer);
+        mp_hal_stdout_tx_strn(buffer, sizeof(buffer));
+        mp_hal_stdout_tx_str(" Exception Vector address: ");
+        format_hex(rst_info->excvaddr, buffer);
+        mp_hal_stdout_tx_strn(buffer, sizeof(buffer));
+        mp_hal_stdout_tx_str(" DEPC: ");
+        format_hex(rst_info->depc, buffer);
+        mp_hal_stdout_tx_strn(buffer, sizeof(buffer));
+        mp_hal_stdout_tx_str("\r\n\r\n");
+    }
+}
+
+#endif
+
 static void mp_reset(void) {
     mp_stack_set_top((void *)0x40000000);
     mp_stack_set_limit(8192);
@@ -112,6 +170,10 @@ void init_done(void) {
     mp_hal_stdout_tx_str("\r\n");
     #if MICROPY_REPL_EVENT_DRIVEN
     pyexec_event_repl_init();
+    #endif
+
+    #if MICROPY_HW_HARD_FAULT_DEBUG
+    print_reset_info();
     #endif
 
     #if !MICROPY_REPL_EVENT_DRIVEN
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -524,22 +524,11 @@ function ci_native_mpy_modules_build {
     else
         arch=$1
     fi
-    for natmod in features1 features3 features4 heapq re
+    for natmod in deflate features1 features3 features4 framebuf heapq random re
     do
         make -C examples/natmod/$natmod clean
         make -C examples/natmod/$natmod ARCH=$arch
     done
-
-    # deflate, framebuf, and random currently cannot build on xtensa due to
-    # some symbols that have been removed from the compiler's runtime, in
-    # favour of being provided from ROM.
-    if [ $arch != "xtensa" ]; then
-        for natmod in deflate framebuf random
-        do
-            make -C examples/natmod/$natmod clean
-            make -C examples/natmod/$natmod ARCH=$arch
-        done
-    fi
 
     # features2 requires soft-float on armv7m, rv32imc, and xtensa.  On armv6m
     # the compiler generates absolute relocations in the object file
@@ -551,10 +540,8 @@ function ci_native_mpy_modules_build {
         make -C examples/natmod/features2 ARCH=$arch
     fi
 
-    # btree requires thread local storage support on rv32imc, whilst on xtensa
-    # it relies on symbols that are provided from ROM but not exposed to
-    # natmods at the moment.
-    if [ $arch != "rv32imc" ] && [ $arch != "xtensa" ]; then
+    # btree requires thread local storage support on rv32imc.
+    if [ $arch != "rv32imc" ]; then
         make -C examples/natmod/btree clean
         make -C examples/natmod/btree ARCH=$arch
     fi
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -526,14 +526,14 @@ function ci_native_mpy_modules_build {
     fi
     for natmod in deflate features1 features3 features4 framebuf heapq random re
     do
-        make -C examples/natmod/$natmod clean
+        make -C examples/natmod/$natmod ARCH=$arch clean
         make -C examples/natmod/$natmod ARCH=$arch
     done
 
     # features2 requires soft-float on armv7m, rv32imc, and xtensa.  On armv6m
     # the compiler generates absolute relocations in the object file
     # referencing soft-float functions, which is not supported at the moment.
-    make -C examples/natmod/features2 clean
+    make -C examples/natmod/features2 ARCH=$arch clean
     if [ $arch = "rv32imc" ] || [ $arch = "armv7m" ] || [ $arch = "xtensa" ]; then
         make -C examples/natmod/features2 ARCH=$arch MICROPY_FLOAT_IMPL=float
     elif [ $arch != "armv6m" ]; then
@@ -542,7 +542,7 @@ function ci_native_mpy_modules_build {
 
     # btree requires thread local storage support on rv32imc.
     if [ $arch != "rv32imc" ]; then
-        make -C examples/natmod/btree clean
+        make -C examples/natmod/btree ARCH=$arch clean
         make -C examples/natmod/btree ARCH=$arch
     fi
 }
--- a/tools/mpy_ld.py
+++ b/tools/mpy_ld.py
@@ -402,6 +402,7 @@ class LinkEnv:
         self.known_syms = {}  # dict of symbols that are defined
         self.unresolved_syms = []  # list of unresolved symbols
         self.mpy_relocs = []  # list of relocations needed in the output .mpy file
+        self.externs = {}  # dict of externally-defined symbols
 
     def check_arch(self, arch_name):
         if arch_name != self.arch.name:
@@ -491,10 +492,14 @@ def populate_got(env):
         sym = got_entry.sym
         if hasattr(sym, "resolved"):
             sym = sym.resolved
-        sec = sym.section
-        addr = sym["st_value"]
-        got_entry.sec_name = sec.name
-        got_entry.link_addr += sec.addr + addr
+        if sym.name in env.externs:
+            got_entry.sec_name = ".external.fixed_addr"
+            got_entry.link_addr = env.externs[sym.name]
+        else:
+            sec = sym.section
+            addr = sym["st_value"]
+            got_entry.sec_name = sec.name
+            got_entry.link_addr += sec.addr + addr
 
     # Get sorted GOT, sorted by external, text, rodata, bss so relocations can be combined
     got_list = sorted(
@@ -520,6 +525,9 @@ def populate_got(env):
             dest = int(got_entry.name.split("+")[1], 16) // env.arch.word_size
         elif got_entry.sec_name == ".external.mp_fun_table":
             dest = got_entry.sym.mp_fun_table_offset
+        elif got_entry.sec_name == ".external.fixed_addr":
+            # Fixed-address symbols should not be relocated.
+            continue
         elif got_entry.sec_name.startswith(".text"):
             dest = ".text"
         elif got_entry.sec_name.startswith(".rodata"):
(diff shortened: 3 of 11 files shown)


jonnor:
Looks like useful additions. Especially as it will be needed for ESP32-C2.

agatti:
In theory this can also be used for other platforms that need to call into ROM from a natmod (or into functions available at fixed address that have no stub for the linker), it just so happens that esp8266 is the only supported platform that has this specific need at the moment.  I've limited this functionality to be only available on Xtensa only because that's what can be validated on hardware on my end.

I thought about parsing `eagle.rom.addr.v6.ld`, but that's a linkerscript and its syntax is much more involved than what I've implemented.  A linkerscript parser that can be used from Python is available at https://github.com/tree-sitter-grammars/tree-sitter-linkerscript but that brings in py-tree-sitter as a requirement and thus a larger code footprint for its integration.

I don't mind to write either a parser that can understand linkerscripts just enough to parse `eagle.rom.addr.v6.ld`, or integrating py-tree-sitter and modify the CI script to install dependencies (looking back at this, documentation has to be updated either way to keep track of the new feature).  I'm partial to the former, but the latter has the potential to be more generic.

dpgeorge:
Following on from https://github.com/micropython/micropython/pull/17379#issuecomment-2917045813, it seems that enabling `LINK_RUNTIME=1` is enough to get all the xtensa natmods working on esp8266.  So the in-ROM version of functions is never actually used (I didn't check the output .mpy file but I assume this is the case, because it's only if a symbol is undefined that the in-ROM locations are used).

But, as you point out, from a memory-use point of view it would be better to prefer in-ROM routines.  Is it possible to make the in-ROM symbols have a preference over ones from elf files?  I guess as a first attempt the in-ROM symbols can always override elf symbols, and in the future (if needed) we can have a way to select which is preferred.

> I don't mind to write either a parser that can understand linkerscripts just enough to parse `eagle.rom.addr.v6.ld`

I think it's a good idea to be able to parse `eagle.rom.addr.v6.ld`.  So a user can use that file verbatim if needed, instead of extracting the symbols they need into a separate file.  I reckon just a custom parser that can parse just that file is enough.  Probably a single regex could do it.

agatti:
> But, as you point out, from a memory-use point of view it would be better to prefer in-ROM routines. Is it possible to make the in-ROM symbols have a preference over ones from elf files? I guess as a first attempt the in-ROM symbols can always override elf symbols, and in the future (if needed) we can have a way to select which is preferred.

Making ROM symbols override ELF ones isn't much of a deal (it's a matter of altering the symbol source lookup order in `link_objects`), but the unused ELF symbols code will still linger in the MPY file since they are already embedded in the input object file at resolution time.  I haven't yet looked at how difficult it'd be to remove the extra symbols before resolution occurs.

Regarding per-symbol resolution order, how about having a `--extern-blacklist` parameter containing a comma-separated list of symbols that should never be picked from linkerscripts (with multiple argument occurrences adding to the list of symbols to skip)?

In the meantime I've rewritten the linkerscript parser to also handle the ESP-IDF symbols lists so - once I enable this on RV32 as well - this should also work for the whole ESP32 MCU line.  The only caveat would be that some extra argument processing will be needed in the makefile to pick up possibly more than just a single linkerscript (same goes for `mpy_ld.py`), as libc/libgcc/newlib symbol definitions are split across multiple files.

dpgeorge:
> I don't know what's the 1.26.0 tentative release date,

It's 1st August 2025.  And the aim is to be strict with that date.

> I'm partial to have this with ROM symbols support and redundant ELF symbols' code in - the final size of the MPY file is exactly the same when ELF resolution is used, as the only change is the symbol address in opcodes/tables pointing to ROM rather than elsewhere.

Yes, I'm happy with that.  We can improve upon it later.

> Regarding per-symbol resolution order

I think we can skip that feature for now, and just always prefer ROM.

> Turns out stripping symbols from within `mpy_ld.py` is doable but not trivial

I don't fully understand why this is difficult: at the moment, if you use `LINK_RUNTIME=1` but don't use any symbols from libc/libgcc/etc then it doesn't add any code at all.  And because libc is an archive made of many .o object files, I assume if you reference a symbol in one of those object files then you only need to include that single object file, not the whole libc archive.  Right?  So that feature alone (of only included .o's when a symbol within them is referenced) should be enough to get optimal sized .mpy files (as long as the .o in the .a are fine grained enough).

agatti:
> I don't fully understand why this is difficult: at the moment, if you use `LINK_RUNTIME=1` but don't use any symbols from libc/libgcc/etc then it doesn't add any code at all. And because libc is an archive made of many .o object files, I assume if you reference a symbol in one of those object files then you only need to include that single object file, not the whole libc archive. Right? So that feature alone (of only included .o's when a symbol within them is referenced) should be enough to get optimal sized .mpy files (as long as the .o in the .a are fine grained enough).

That's indeed correct, but I've probably expressed myself incorrectly as my point applies to single symbols regardless the object files' granularity.

Let's look at the btree natmod as an example.

The ESP8266 ROM has mem{cpy,set,move,cmp,etc.} implementations, and they're exposed in the linkerscript file, so their addresses are used in the final MPY file if they're encountered, as expected when using an externs file.

However, if you build the natmod (even without `LINK_RUNTIME` in this case) and look at the generated `build/btree_c.o` file you can see those symbols being referenced, but their implementation is present in the object file itself:

```
$ nm build/btree_c.o | grep -i " mem"

0000042c T memcpy
0000045c T memmove
00000444 T memset
```

That object file is then used for linking by `mpy_ld.py`, and it contains one single `.text` section so, as far as I know, the only way for redundant symbols to not show up in the final MPY file in that case is to cut the ranges out from the raw .text section data bytes and then fix up offsets past the cut out points.

I haven't tried building with `-ffunction-sections` to see if `mpy_ld.py` handles that; if it does then it's much easier to strip redundant symbols away, by skipping the whole section rather than operating on section data buffers.

agatti:
This iteration of the patch should address the pending review issues and enforce the new resolution behaviour, as in ROM symbols override ELF ones in every case.

However, this made testing more interesting.  The modules that are too large to be imported as-is on an ESP8266 and require calling `esp.set_native_code_location` to be able to be loaded in now exhibited an unexpected behaviour, with "impossible" crashes that took me some time disassembling code to see whether the patch didn't operate correctly or not.  I've never had any luck in using a JTAG probe with an ESP8266, so that's all through trial and error.  If anybody knows how to get JTAG working on that please let me know.

The natmods in question are `btree`, `deflate`, and `framebuf`.  `btree` is simply too large to be loaded unless some more RAM is freed anyway and so it couldn't be tested (as before).  

`framebuf` tests work *except* for `framebuf_polygon` where it would crash inside `mp_obj_get_int` with a crash EPC1 pointing into the void (somewhere in the 0x8000_0000 address range).  I've looked at the MPY CODE segment to check if I was poking in the wrong values, but things look OK: see https://gist.github.com/agatti/4398c2a28ed786171c31e54cef2f18cf, the ROM division symbol addresses are correctly put at 0x0000_1008 and 0x0000_100C and the code for `framebuf_poly` before and slightly after the crash point (0x0000_1f57 in the disassembled code) looks correct.  

For `deflate` I've just looked at the first one, `deflate_compress`, crashing inside `io.BytesIO.write` when building the buffer to compress (appending b"micropython" 10 times).

Now, why it didn't crash before, when ROM symbols were picked only as a last resort?  I have two hypotheses that still do not fully explain this.

* Code (or something else) put in the memory area reserved by `esp.set_native_code_location` gets somehow corrupted at runtime (that'd explain the `framebuf` crash)
* Mixing libgcc code with ESP8266 ROM symbols isn't really supported and some registers are trashed behind the scenes or there's some stack weirdness going (`deflate` needs `__modsi3`, which is not present in the ROM and thus has to be fetched from libgcc.a).

Well, there's a third hypothesis, my board's flash has failing cells in the area used by `esp.set_native_code_location`, but I only have one ESP8266 board I can try this on at the moment.  I'd be really surprised if it's this one though.

This needs a bit more investigation, but I doubt I'd make some progress quickly without getting JTAG to work.  Again, if anybody knows how to get it to work on Linux, please speak up! :)

**Edit:** the target address range suspiciously looks like it's been relocated even though it shouldn't have been, but that might just be a coincidence, otherwise why would it crash *after* a ROM call just took place successfully?

**Edit2:** and yes, it does get relocated - I guess I'll have to claim relocation type 0x7E for `py/persistentcode.c` to leave it alone since this is not going to be Xtensa specific and it's rather unprobable that the function table will grow to 135 entries whilst keeping the current file format version number (plus I have no idea on how to expand the Xtensa literal section to get this to work).

agatti:
@dpgeorge I've had to assign a new relocation type number (126, 0x7E) for fixed address entries to prevent their relocation at load time.  It should not be a change that breaks existing code as the function table never reached that number of entries.  However, if that's something you want to delay for a more significant set of MPY format changes that would require a version number bump, that's understandable.

That said, now all `framebuf` tests pass - `deflate` tests behave as when using ELF symbols ~~except for `tests/extmod/deflate_stream_error.py` that crashes when using ROM symbols~~ (the crash was due to an improper code location offset for `esp.set_native_code_location`, trashing data).

The request for help for JTAG on an ESP8266 is still valid, even though I've added an optional exception reporter to the ESP8266 port.  It prints the contents of `struct rst_info` if the board rebooted due to an exception or a watchdog kicking in (hardware or software).

So, relocation type number concerns aside this should be ready for review - it incorporates the suggestions you've made earlier but there's some new code in.

ports/esp8266/main.c
```diff
@@ -49,6 +49,64 @@
 
 static char heap[38 * 1024];
 
+#if MICROPY_PY_EXCEPTION_REPORT
```
dpgeorge: Config names of the form `MICROPY_PY_xxx` are intended to be for things in the Python scope, like modules, types, etc.

Also, "exception" could be confused with Python exceptions.

I suggest a name like `MICROPY_HW_HARD_FAULT_DEBUG`

py/persistentcode.c
```diff
@@ -140,6 +140,9 @@ void mp_native_relocate(void *ri_in, uint8_t *text, uintptr_t reloc_text) {
         } else if (op == 8) {
             // Destination is mp_fun_table itself
             dest = (uintptr_t)&mp_fun_table;
+        } else if (op == 126) {
+            // Destination is a fixed address and must be left alone
+            dest = 0;
```
dpgeorge: Is this needed in the runtime?  Couldn't it be that `mpy_ld.py` just skips emitting relocations of this type altogether?

tools/ci.sh
```diff
@@ -524,37 +524,24 @@ function ci_native_mpy_modules_build {
...
-            make -C examples/natmod/$natmod ARCH=$arch
-        done
-    fi
-
     # features2 requires soft-float on armv7m, rv32imc, and xtensa.  On armv6m
     # the compiler generates absolute relocations in the object file
     # referencing soft-float functions, which is not supported at the moment.
     make -C examples/natmod/features2 clean
-    if [ $arch = "rv32imc" ] || [ $arch = "armv7m" ] || [ $arch = "xtensa" ]; then
+    if [ $arch = "armv7m" ] || [ $arch = "rv32imc" ] || [ $arch = "xtensa" ]; then
```
dpgeorge: To minimise diff, I suggest not changing this line.

ports/esp8266/main.c
agatti: Thanks!  Maybe the `MICROPY_PY_` / `MICROPY_HW_` notation details should be added to `CODECONVENTIONS.md` or to another document under `docs/develop/`.

I've noticed a few (I guess now historical) divergences though, which maybe could be normalised for v2?  E.g. `MICROPY_REPL_EMACS_KEYS` could really be named `MICROPY_PY_REPL_EMACS_KEYS` since it is entirely in the Python scope.

py/persistentcode.c
agatti: I'll need to have a closer look at the on-disk format to see if this can be avoided.  On the other hand an absolute relocation type sort of makes sense in the grand scheme of things, but I can see the size savings aspect of not having one using the current format.

I can think of two ways to achieve the first half (not emitting explicit relocations):

1 - Whilst reordering the relocations table to put the externs together at the end should be possible, some other file pointers/offsets may need to be updated if the MPY file data is loaded sequentially (as in, expecting an aligned code block right after the 0xFF marker in the relocation table).

2 - Remove extern relocations from the list, extend the new `.text` segment and put extern address words at the beginning there.  This would also need all sort of fixups across the board, I guess.

I suspect both approaches would also potentially require custom handling of most relocations in `mpy_ld.py`, splitting cases into regular and externs, though.  I still need to play with that idea a bit more to see what I can come up with.

dpgeorge: I was thinking it would be as simple as not adding the fixed address relocation to the `env.mpy_relocs` list?

tools/mpy_ld.py
```diff
@@ -520,6 +525,8 @@ def populate_got(env):
             dest = int(got_entry.name.split("+")[1], 16) // env.arch.word_size
         elif got_entry.sec_name == ".external.mp_fun_table":
             dest = got_entry.sym.mp_fun_table_offset
+        elif got_entry.sec_name == ".external.fixed_addr":
+            dest = ".fixed_addr"
```
dpgeorge: I think this could simply be:
```py
elif got_entry.sec_name == ".external.fixed_addr":
    continue
```
Then there doesn't need to be a special 126 code.

agatti: Yes, you're right.  For some reason I kept thinking the relocation table and the relocation information were interleaved somehow, which is funny since I've been staring at MPY's hex dumps and their disassembled output for quite some time now so I should have known better!

I've tested your suggestion on an ESP8266 for one of the previously failing tests (`framebuf_polygon`) and it works.  I'll rebuild the PR without the extra relocation type in.

tools/mpy_ld.py
```diff
@@ -1207,13 +1215,26 @@ def link_objects(env, native_qstr_vals_len):
...
             if sym.name in fun_table:
                 sym.section = mp_fun_table_sec
                 sym.mp_fun_table_offset = fun_table[sym.name]
             else:
                 undef_errors.append("{}: undefined symbol: {}".format(sym.filename, sym.name))
 
+    if env.externs:
+        for sym in env.unresolved_syms:
+            if sym.name in env.externs:
+                log(
```
dpgeorge: I'm not sure the logic of this logging is correct.  Eg building `framebuf` I get:
```
$ make ARCH=xtensa -B
GEN build/framebuf_xtensa.config.h
CC framebuf.c
LINK build/framebuf.o
Symbol __udivsi3 is a fixed-address symbol at 4000e21c and is also provided from an object file
Symbol __divsi3 is a fixed-address symbol at 4000dc88 and is also provided from an object file
arch:         EM_XTENSA
text size:    8108
bss size:     160
GOT entries:  29
GEN framebuf_xtensa.mpy
```
This example doesn't use libc, so I don't think these symbols are provided from an object file.

agatti: That's interesting - I tested that on the `deflate` module and it yielded the correct result as it's one where there was at least a redundant ELF symbol...  But yes, thanks for spotting this.

It is now sorted out, tested on `btree`, `deflate`, and `framebuf` modules.

tools/mpy_ld.py
```diff
@@ -1207,13 +1215,27 @@ def link_objects(env, native_qstr_vals_len):
...
+    seen_externs = set()
+    for sym in env.externs:
+        if sym in env.known_syms and sym not in seen_externs:
+            log(
+                LOG_LEVEL_1,
+                "Symbol {} is a fixed-address symbol at {:08x} and is also provided from an object file".format(
+                    sym, env.externs[sym]
+                ),
+            )
+            seen_externs.add(sym)
```
dpgeorge: Sorry to keep going back and forth here, but is `seen_externs` actually needed?  Because `env.externs` is guaranteed to have unique entries, you'll never see the same symbol twice.

agatti: It might be a leftover of the previous loop iteration, see https://github.com/micropython/micropython/compare/2b46c83bab15c342cf16bc498f4b01a018a6acb9..23515be466d5c6e1f780e4057a1439e6aab1e58e where the `btree` module would report redundant symbols multiple times.

Looking back at it, yes, it should have been removed along with the old loop.  I'll fix this right away, thanks!

dpgeorge (approved):
Thanks for updating, this looks good now.  Tested building all natmod examples for xtensa arch, and ran them on esp8266 (using the new prelude option to `run-natmodtests.py`).

PR #11397, opened 2023-05-02
ports/mimxrt: Add CYW43 WiFi/BT support.

iabdalkader:
This is a work in progress, attempt at adding CYW43 WiFi/BT support to `ports/mimxrt`, starting with WiFi first and Bluetooth will follow when this is working.

PR #16848, opened 2025-03-03
drivers/cywbt: Switch to new CYW43 BTHCI UART backend.

iabdalkader:
### Summary

Remove old BTCHI driver and switch to the new CYW43 BTHCI UART backend and update the CYW configuration accordingly. More details here: https://github.com/georgerobotics/cyw43-driver/pull/137#

### Testing

mimxrt: Tested WiFi/BT with MIMXRT1062 + CYW43 module.
stm32: Tested WiFi/BT with Giga H7.

Requires:

- [x] https://github.com/georgerobotics/cyw43-driver/pull/137#

This is also related https://github.com/micropython/micropython/pull/16833 I renamed the files to match the cyw43.

PR #17092, opened 2025-04-08
extmod: Create a common cyw43 config header, apply mDNS fix in it.

projectgus:
### Summary

* First commit creates a new `extmod/cyw43_config_common.h` header with the common macros from ports which use CYW43 Wi-fi driver. There's more refactoring that could be done here (for example a `cyw43_sdio_common.h` header) but I've limited myself to macros that are the same, or that appear effectively the same, on every port.
* Second commit moves the rp2 LWIP mDNS resolver fix added in #17057 into the new header, so it's now also applied on stm32 and mimxrt ports which use the CYW43 driver.

### Testing

* Connected a PYBD_SF2 to my local network, verified I could look up the host via mDNS.
* Ran the `multi_wlan` tests with a RPI_PICO_W board and a PYBD_SF2 as the two nodes. This uncovered two pre-existing issues (not added in this PR):
   - The stm32 port logs `DHCPS:` updates when running as the AP, so the test fails on the additional output (at least easy to manually check for).
   - On both ports, the cyw43 driver doesn't like being a STA first and then switching to become an AP (even though the unit tests de-activate the unused interface). To run the two tests in sequence then I needed to trigger a hard reset in between. Will investigate more when time permits.
* I have not been able to test mimxrt as I have no suitable hardware. If I understand #11397 correctly then no such widely available board exists, the only mimxrt+cyw43 users have added a CYW43 module to custom hardware or an eval board.

* Could keep the per-port config headers but it makes it harder to verify consistent behaviour between the different ports, and to apply non-port-specific changes like the mDNS resolver fix.

### DIFF
--- a/extmod/cyw43_config_common.h
+++ b/extmod/cyw43_config_common.h
@@ -107,4 +107,20 @@ static inline void cyw43_delay_ms(uint32_t ms) {
     }
 }
 
+#if LWIP_MDNS_RESPONDER == 1
+
+// Hook for any additional TCP/IP initialization than needs to be done.
+// Called after the netif specified by `itf` has been set up.
+#ifndef CYW43_CB_TCPIP_INIT_EXTRA
+#define CYW43_CB_TCPIP_INIT_EXTRA(self, itf) mdns_resp_add_netif(&self->netif[itf], mod_network_hostname_data)
+#endif
+
+// Hook for any additional TCP/IP deinitialization than needs to be done.
+// Called before the netif specified by `itf` is removed.
+#ifndef CYW43_CB_TCPIP_DEINIT_EXTRA
+#define CYW43_CB_TCPIP_DEINIT_EXTRA(self, itf) mdns_resp_remove_netif(&self->netif[itf])
+#endif
+
+#endif
+
 #endif // MICROPY_INCLUDED_EXTMOD_CYW43_CONFIG_COMMON_H
--- a/extmod/cyw43_config_common.h
+++ b/extmod/cyw43_config_common.h
@@ -1,0 +1,110 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 Damien P. George, Angus Gratton
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#ifndef MICROPY_INCLUDED_EXTMOD_CYW43_CONFIG_COMMON_H
+#define MICROPY_INCLUDED_EXTMOD_CYW43_CONFIG_COMMON_H
+
+// The board-level config will be included here, so it can set some CYW43 values.
+#include "py/mpconfig.h"
+#include "py/mperrno.h"
+#include "py/mphal.h"
+#include "py/runtime.h"
+#include "extmod/modnetwork.h"
+#include "lwip/apps/mdns.h"
+#include "pendsv.h"
+
+// This file is included at the top of port-specific cyw43_configport.h files,
+// and holds the MicroPython-specific but non-port-specific parts.
+//
+// It's included into both MicroPython sources and the lib/cyw43-driver sources.
+
+#define CYW43_IOCTL_TIMEOUT_US          (1000000)
+#define CYW43_NETUTILS                  (1)
+#define CYW43_PRINTF(...)               mp_printf(MP_PYTHON_PRINTER, __VA_ARGS__)
+
+#define CYW43_EPERM                     MP_EPERM // Operation not permitted
+#define CYW43_EIO                       MP_EIO // I/O error
+#define CYW43_EINVAL                    MP_EINVAL // Invalid argument
+#define CYW43_ETIMEDOUT                 MP_ETIMEDOUT // Connection timed out
+
+#define CYW43_THREAD_ENTER              MICROPY_PY_LWIP_ENTER
+#define CYW43_THREAD_EXIT               MICROPY_PY_LWIP_EXIT
+#define CYW43_THREAD_LOCK_CHECK
+
+#define CYW43_HOST_NAME                 mod_network_hostname_data
+
+#define CYW43_ARRAY_SIZE(a)             MP_ARRAY_SIZE(a)
+
+#define CYW43_HAL_PIN_MODE_INPUT        MP_HAL_PIN_MODE_INPUT
+#define CYW43_HAL_PIN_MODE_OUTPUT       MP_HAL_PIN_MODE_OUTPUT
+#define CYW43_HAL_PIN_PULL_NONE         MP_HAL_PIN_PULL_NONE
+#define CYW43_HAL_PIN_PULL_UP           MP_HAL_PIN_PULL_UP
+#define CYW43_HAL_PIN_PULL_DOWN         MP_HAL_PIN_PULL_DOWN
+
+#define CYW43_HAL_MAC_WLAN0             MP_HAL_MAC_WLAN0
+#define CYW43_HAL_MAC_BDADDR            MP_HAL_MAC_BDADDR
+
+#define cyw43_hal_ticks_us              mp_hal_ticks_us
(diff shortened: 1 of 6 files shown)


projectgus:
@dpgeorge BTW I haven't forgotten our discussion about `network_cyw43.h`. I have a proposed refactor that covers this, but it turned out independent from this one.

ports/rp2/cyw43_configport.h
```diff
@@ -26,34 +26,27 @@
...
-#include "py/runtime.h"
-#include "extmod/modnetwork.h"
-#include "lwip/apps/mdns.h"
-#include "pendsv.h"
+#include "extmod/cyw43_config_common.h"
 
 #define CYW43_INCLUDE_LEGACY_F1_OVERFLOW_WORKAROUND_VARIABLES (1)
 #define CYW43_WIFI_NVRAM_INCLUDE_FILE   "wifi_nvram_43439.h"
-#define CYW43_IOCTL_TIMEOUT_US          (1000000)
 #define CYW43_SLEEP_MAX                 (10)
```
projectgus, line 33: I haven't moved this macro into the common header because it's `50` in the mimxrt & stm32 ports, and `10` here in rp2. I'm unsure if this difference is necessary or a historical accident.

extmod/cyw43_config_common.h
```diff
@@ -0,0 +1,104 @@
...
+static inline void cyw43_delay_ms(uint32_t ms) {
+    // PendSV may be disabled via CYW43_THREAD_ENTER, so this delay is a busy loop.
+    uint32_t us = ms * 1000;
+    uint32_t start = mp_hal_ticks_us();
+    while (mp_hal_ticks_us() - start < us) {
+        mp_event_handle_nowait();
+    }
+}
+
+#define CYW43_EVENT_POLL_HOOK mp_event_handle_nowait()
```
iabdalkader, line 102: For `stm32` and `mimxrt` this was `MICROPY_EVENT_POLL_HOOK` which does something different, especially in the `stm32` port (see the old one in stm32 port). The same for the delay functions which called this macro. I know that eventually we'll use the new functions, but I'd suggest you leave this macro and the delay functions in the port config files for now, to avoid changing the behavior, unless it's time to remove this macro.

projectgus: Oh good catch, I mis-remembered the macro compatibility as being different! Will fix.

projectgus: I've swapped this around so stm32 should be using the same hook as before.

mimxrt should be using mostly the same hooks, although it previously had `cyw43_delay_us` and `cyw43_delay_ms` aliased directly to `mp_hal_delay_us` and `mp_hal_delay_ms` and now it uses the inline function definitions instead. This seems like either a no-op or a small improvement, but I'm not able to test it to verify.

iabdalkader:
Note that the old cywbt.c driver will be replaced with a new BTHCI Uart backend that's been recently added to cyw43 driver, which requires some additional config option (no big changes, see https://github.com/micropython/micropython/pull/16848). The new Alif port already uses the new BTHCI backend, and it could be switch to the common config: https://github.com/micropython/micropython/blob/master/ports/alif/cyw43_configport.h

ports/rp2/cyw43_configport.h
iabdalkader: It just controls how often the the cyw43 poll gets scheduled via pendsv (at least in stm32, mimxrt, alif). I'd say on a slower chip maybe 50 is better.

iabdalkader: Oh! Perhaps just have the default at 50 and in RP2 `undef` and redefine it.

projectgus: Right! What I'm unsure is whether it was set to 10 on rp2 to fix some problem (or improve performance), or whether it's simply an accident that it's been set to 10 here. I haven't done any tests to compare 50 vs 10 on rp2.

dpgeorge:
Sorry @projectgus but this now needs rebasing after merging #16848.

Gadgetoid:
Some possible overlap between this and https://github.com/micropython/micropython/pull/16057/commits/d6cfdcaf4e53e4b99e154aa8a629cebc4a525a3d from https://github.com/micropython/micropython/pull/16057 ?

ports/rp2/cyw43_configport.h
projectgus: > Oh! Perhaps just have the default at 50 and in RP2 `undef` and redefine it.

Good idea. I've gone even further and removed it from all the ports except rp2, as [50 is the default in the driver repo](https://github.com/georgerobotics/cyw43-driver/blob/main/src/cyw43_config.h#L126). I suspect it could also be set to 50 in the rp2 port without any impact (maybe even lower sleep consumption), but haven't tested this.

projectgus:
> The new Alif port already uses the new BTHCI backend, and it could be switch to the common config

> Sorry @projectgus but this now needs rebasing after merging #16848.

I've rebased on top the BTHCI changes, and migrated the Alif port to the common cyw43 config. All ports build after updating, but I only have hardware to test stm32 PYBD_SF2 (confirmed Wi-Fi & BLE both work but didn't run complete tests).

I think the BTHCI macros could also be put into the common config but have left it out of this PR for now, same as SDIO.

projectgus:
> Some possible overlap between this and [d6cfdca](https://github.com/micropython/micropython/commit/d6cfdcaf4e53e4b99e154aa8a629cebc4a525a3d) from #16057 ?

I think some overlap, but as those changes are scoped to rp2 I think they're more or less a subset of these.

ports/alif/cyw43_configport.h
```diff
@@ -53,51 +48,16 @@
...
-
-#define CYW43_THREAD_ENTER              MICROPY_PY_LWIP_ENTER
-#define CYW43_THREAD_EXIT               MICROPY_PY_LWIP_EXIT
-#define CYW43_THREAD_LOCK_CHECK
-
-#define CYW43_HOST_NAME                 mod_network_hostname_data
-
 #define CYW43_SDPCM_SEND_COMMON_WAIT    __WFE()
 #define CYW43_DO_IOCTL_WAIT             // __WFE()
 #define CYW43_EVENT_POLL_HOOK           mp_event_handle_nowait()
```
iabdalkader, line 55: This is redefined.

iabdalkader:
I checked all changes for all the ports that I care about (so all except rp2), and they all look good. Also, I tested Alif port, and stm32 and both WiFi/BT are still working. I can't test mimxrt any time soon, maybe @kwagyeman can test it.

kwagyeman:
@iabdalkader - Pulled the latest of master of OpenMV, and then the micropython branched referenced, and applied the two commits and ran on the IMX. Bluetooth and WiFi both work.

ports/alif/cyw43_configport.h
projectgus: Good catch, thanks!

projectgus:
Thanks @iabdalkader and @kwagyeman for testing the other ports. Have removed the duplicate definition and rebased.

extmod/cyw43_config_common.h
```diff
@@ -0,0 +1,143 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#ifndef MICROPY_INCLUDED_EXTMOD_CYW43_CONFIG_COMMON_H
+#define MICROPY_INCLUDED_EXTMOD_CYW43_CONFIG_COMMON_H
+
+// The board-level config will be included here, so it can set some CYW43 values.
+#include <stdio.h>
```
dpgeorge, line 30: Is stdio.h needed?  I see it came from the rp2 port but I don't think it's needed there.  Both RPI_PICO_W and RPI_PICO2_W build with this include removed.

extmod/cyw43_config_common.h
```diff
@@ -0,0 +1,143 @@
...
+#include "py/mphal.h"
+#include "py/runtime.h"
+#include "extmod/modnetwork.h"
+#include "lwip/apps/mdns.h"
+#include "pendsv.h"
+
+// This file is included at the top of port-specific cyw43_configport.h files,
+// and holds the MicroPython-specific but non-port-specific parts.
+//
+// It's included into both some MicroPython sources and the lib/cyw43-driver sources.
```
dpgeorge, line 42: Remove "some"?

extmod/cyw43_config_common.h
```diff
@@ -0,0 +1,143 @@
...
+
+// Hook for any additional TCP/IP deinitialization than needs to be done.
+// Called before the netif specified by `itf` is removed.
+#ifndef CYW43_CB_TCPIP_DEINIT_EXTRA
+#define CYW43_CB_TCPIP_DEINIT_EXTRA(self, itf) mdns_resp_remove_netif(&self->netif[itf])
+#endif
+
+#endif
+
+#if LWIP_MDNS_RESPONDER == 1
```
dpgeorge, line 127: This bit is added twice :joy:

dpgeorge (approved):
Apart from the few comments above, the refactoring looks good to me.  I tested it on PYBD-SF6 and don't see any regressions.

extmod/cyw43_config_common.h
```diff
@@ -0,0 +1,143 @@
+/*
```
dpgeorge, line 1: I'm not sure, but maybe this file should be `extmod/cyw43-include/cyw43_config_common.h` or `extmod/cyw43-driver-include/cyw43_config_common.h`?

There are other common include dirs with this naming scheme.

projectgus: The other way we could do this is to move all three of `extmod/network_cyw.*` and this file to a directory like `extmod/cyw43/`. What do you think?

extmod/cyw43_config_common.h
projectgus: Oops, I must have messed up a rebase somewhere

extmod/cyw43_config_common.h
projectgus: Good catch, I don't know how this got here (probably clangd and I didn't notice it).

projectgus: Oh I missed you noticed it was copied from the rp2 port! But yes, have removed.

extmod/cyw43_config_common.h
dpgeorge: Let's leave it where it is for now.  Can move it later if needed (eg in conjunction with factoring this into SDIO/SPI parts).

PR #17095, opened 2025-04-09
zephyr: Make it clearer that gpio irqs can be hard or soft.

projectgus:
### Summary

Quick follow-up to #16916 - the comment in the IRQ handler makes it sound like all Zephyr IRQs are hard, but actually hard and soft irqs are supported.

### Testing

None, this change only updates a code comment.

### DIFF
--- a/ports/zephyr/machine_pin.c
+++ b/ports/zephyr/machine_pin.c
@@ -64,8 +64,11 @@ static void gpio_callback_handler(const struct device *port, struct gpio_callbac
     machine_pin_irq_obj_t *irq = CONTAINER_OF(cb, machine_pin_irq_obj_t, callback);
 
     #if MICROPY_STACK_CHECK
-    // This callback executes in an ISR context so the stack-limit check must be changed to
-    // use the ISR stack for the duration of this function (so that hard IRQ callbacks work).
+    // For irqs with hard=True, the callback executes in an ISR context so the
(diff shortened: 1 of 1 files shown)


projectgus:
@danicampora Is that clearer?

ports/zephyr/machine_pin.c
```diff
@@ -64,8 +64,11 @@ static void gpio_callback_handler(const struct device *port, struct gpio_callbac
     machine_pin_irq_obj_t *irq = CONTAINER_OF(cb, machine_pin_irq_obj_t, callback);
 
     #if MICROPY_STACK_CHECK
-    // This callback executes in an ISR context so the stack-limit check must be changed to
-    // use the ISR stack for the duration of this function (so that hard IRQ callbacks work).
+    // For irqs with hard=True, the callback executes in an ISR context so the
```
dpgeorge, line 67: Technically `gpio_callback_handler` executes in an ISR context which is what I think I intended this comment to mean (and in the parenthesis the comment here mentions that it's so hard IRQ callbacks work).

I think it's worth keeping the statement that this C function executes in an ISR context and we want hard *Python* callbacks to behave correctly.

See #6146 for reference.

projectgus: Ah right I missed that "this callback" refers to the enclosing function and is not referring to the Python callback.

It's probably fine to leave it how it is.

PR #17107, opened 2025-04-10
mimxrt/Makefile: Fix dependencies for generation of flexram_config.s.

dpgeorge:
### Summary

Prior to this fix the following would fail:

    $ make build-TEENSY40/flexram_config.s

because it didn't create the build directory before generating the file.

Also, make `hal/resethandler_MIMXRT10xx.S` have an explicit dependency on `flexram_config.s` rather than the latter just being forced to be built before everything else.

### Testing

The following now works from a fresh start (ie no existing build):
```
$ make build-TEENSY40/flexram_config.s
```

Also, `make -j build-TEENSY40/hal/resethandler_MIMXRT10xx.o` still works.

### DIFF
--- a/ports/mimxrt/Makefile
+++ b/ports/mimxrt/Makefile
@@ -327,6 +327,8 @@ SRC_SS = \
 
 SRC_S += shared/runtime/gchelper_thumb2.s \
 
+hal/resethandler_MIMXRT10xx.S: $(GEN_FLEXRAM_CONFIG_SRC)
+
 # =============================================================================
 # QSTR Sources
 # =============================================================================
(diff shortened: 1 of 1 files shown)


robert-hh:
It looks fine, but why would one run a make for that single file? It worked as part of building the whole firmware.
`make build-TEENSY40/flexram_config.s`

robert-hh (approved):

dpgeorge:
> It looks fine, but why would one run a make for that single file?

Oh, sorry, I forgot to mention the motivation for this fix: when I do `make -j` from a clean build (no existing build directory) sometimes it generates `flexram_config.s` before creating the directories.  So this patch is necessary if you're using `make -j` (although it can be hard to trigger the build bug).

dpgeorge:
> But still, Makefile defines TEENSY40 as default, so the board should be created

No, I mean if it's a valid board (eg using the default) but the *build* directory for that board doesn't exist yet (because mkdir hasn't run yet) then `make -j` can fail.

PR #17108, opened 2025-04-10
py/repl: Skip private variables when printing tab completion options.

andrewleech:
### Summary

The tab completion on repl is intended to not show any attributes starting with a single `_` character, however variables/functions in frozen modules are are currently shown.

This PR introduces a change to py/repl to skip private variables when printing tab completion options.

Originally included in https://github.com/micropython/micropython/pull/17011

### Testing

A unit test has added which covers this, currently based on the asyncio module which is generally frozen in. 
However looking again now maybe it would be better to add a suitable function / attribute to `./tests/frozen/frozentest.py` and test against that, assuming it should have the same behavior?

### DIFF
--- a/py/repl.c
+++ b/py/repl.c
@@ -218,6 +218,10 @@ static void print_completions(const mp_print_t *print,
     for (qstr q = q_first; q <= q_last; ++q) {
         size_t d_len;
         const char *d_str = (const char *)qstr_data(q, &d_len);
+        // filter out words that begin with underscore unless there's already a partial match
+        if (s_len == 0 && d_str[0] == '_') {
+            continue;
+        }
         if (s_len <= d_len && strncmp(s_start, d_str, s_len) == 0) {
             if (test_qstr(obj, q)) {
                 int gap = (line_len + WORD_SLOT_LEN - 1) / WORD_SLOT_LEN * WORD_SLOT_LEN - line_len;
--- a/tests/cmdline/repl_autocomplete_underscore.py
+++ b/tests/cmdline/repl_autocomplete_underscore.py
@@ -1,0 +1,33 @@
+# Test REPL autocompletion filtering of underscore attributes
+
+# Start paste mode
+{\x05}
+class TestClass:
+    def __init__(self):
+        self.public_attr = 1
+        self._private_attr = 2
+        self.__very_private = 3
+    
+    def public_method(self):
+        pass
+    
+    def _private_method(self):
+        pass
+    
+    @property
+    def public_property(self):
+        return 42
+    
+    @property  
+    def _private_property(self):
+        return 99
+
+{\x04}
+# Paste executed
+
+# Create an instance
+obj = TestClass()
+
+# Test tab completion on the instance
+# The tab character after `obj.` and 'a' below triggers the completions
+obj.{\x09}{\x09}a{\x09}
--- a/tests/run-tests.py
+++ b/tests/run-tests.py
@@ -405,6 +405,10 @@ def run_micropython(pyb, args, test_file, test_file_abspath, is_special=False):
                                     return rv
 
                     def send_get(what):
+                        # Detect {\x00} pattern and convert to ctrl-key codes.
+                        ctrl_code = lambda m: bytes([int(m.group(1))])
+                        what = re.sub(rb'{\\x(\d\d)}', ctrl_code, what)
+
                         os.write(master, what)
                         return get()
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ target-version = "py37"
 [tool.ruff.lint]
 exclude = [  # Ruff finds Python SyntaxError in these files
   "tests/cmdline/repl_autocomplete.py",
+  "tests/cmdline/repl_autocomplete_underscore.py",
   "tests/cmdline/repl_autoindent.py",
   "tests/cmdline/repl_basic.py",
   "tests/cmdline/repl_cont.py",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ exclude = [  # Ruff finds Python SyntaxError in these files
   "tests/cmdline/repl_basic.py",
   "tests/cmdline/repl_cont.py",
   "tests/cmdline/repl_emacs_keys.py",
+  "tests/cmdline/repl_paste.py",
   "tests/cmdline/repl_words_move.py",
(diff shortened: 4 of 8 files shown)


py/repl.c
```diff
@@ -218,7 +218,8 @@ static void print_completions(const mp_print_t *print,
     for (qstr q = q_first; q <= q_last; ++q) {
         size_t d_len;
         const char *d_str = (const char *)qstr_data(q, &d_len);
-        if (s_len <= d_len && strncmp(s_start, d_str, s_len) == 0) {
+        // check if item matches prefix and it's not a hidden attribute (starting with single underscore)
+        if (s_len <= d_len && strncmp(s_start, d_str, s_len) == 0 && (d_str[0] != '_' || (d_len > 1 && d_str[1] == '_'))) {
```
dpgeorge, line 222: Wouldn't it be simpler and maybe smaller code to follow the pattern of `find_completions()` like this:
```c
        if (s_len == 0 && d_str[0] == '_') {
            continue;
        }                                        
        if (s_len <= d_len && strncmp(...) == 0) {
```
?

dpgeorge:
It's actually not easy to make a test for this that is robust because the bug relies on a certain ordering of qstrs in the frozen set of qstrs (it needs qstrs beginning with underscore to appear between the first and last items that match the tab completion).

So maybe it's not worth adding a test for?  Does the test you added fail prior to the fix in this PR?

andrewleech:
Curiosity - is the code size change different if it's all in one `if` statement?

dpgeorge:
CI needs fixing on this.

andrewleech:
@dpgeorge I've got a working test now I think should be reliable.

Added I minor change to the repl test support in run-tests.py to allow using ctrl-keys in these tests, can you think of any negative side effects of this?

dpgeorge:
> Added I minor change to the repl test support in run-tests.py to allow using ctrl-keys in these tests, can you think of any negative side effects of this?

I see it's only used for "special" tests, which limits its scope (that's good!).  So probably OK.

But, why does it need to run code using paste mode?  It seems like it should work just constructing that class normally?

tests/run-tests.py
```diff
@@ -405,6 +405,12 @@ def get(required=False):
                                     return rv
 
                     def send_get(what):
+                        # Re-parse lines that match \x(dd) pattern and convert to chr(dd)
+                        if m := re.match(rb'\\x(\d\d)', what):
+                            # hex_value = int(m.group(1), 16)
+                            # what = bytes([hex_value]) + what[m.end():]
+                            what = bytes([int(m.group(1))])
```
dpgeorge: Remove commented out code.

And maybe make the escape sequence `{\xDD}` to make it more obvious?

tests/cmdline/repl_autocomplete_underscore.py
```diff
@@ -0,0 +1,34 @@
...
+
+\x04
+# Paste executed
+
+# Create an instance
+obj = TestClass()
+
+# Test tab completion on the instance
+# The tab character after `obj.` below triggers the completion
+obj.	a
```
dpgeorge: Could it use the escape sequence for tab here, so it's clearer?  (Assuming the escape sequence is needed for paste mode, may as well use it here as well.)

andrewleech:
> But, why does it need to run code using paste mode? It seems like it should work just constructing that class normally?

When the class code it "typed" into repl via pty the auto-indent throws it off. I could flatten the class in the .py file, but then would still need backspace to de-dent in between each function.

dpgeorge:
> then would still need backspace to de-dent in between each function

The existing test `repl_autocomplete.py` has backspace in it as a raw byte.  You could follow that.

I'm not against adding an escape sequence parser, but if it's easy to work around that, might be better not to add such complexity?

dpgeorge:
I guess we don't have any tests for paste mode, so maybe worth adding that anyway, using escape sequences?

andrewleech:
I've updated / simplified the escape code parser a little in the latest push as suggested earlier.

> I guess we don't have any tests for paste mode, so maybe worth adding that anyway, using escape sequences?

I had thought that, I went looking and didn't find any. 
I didn't realise backspace and similar codes could just be added (with a hex editor etc) would you prefer that used for this / paste test, rather than the regex based parser? 
Would you prefer a separate explicit paste test as well?

Taking a step back... the test I've added here seems to pass even without the fix in both `./run-tests.py cmdline/repl_autocomplete_underscore.py` and `./run-tests.py --via-mpy cmdline/repl_autocomplete_underscore.py` - is it only when frozen into the build it fails? Assuming that's the case, I guess this test needs to be partly integrated into https://github.com/micropython/micropython/blob/b725c26b4e351b71b8f72ad8e55bbbd5ccce4602/ports/unix/variants/coverage/manifest.py#L4 instead.

dpgeorge:
> I didn't realise backspace and similar codes could just be added (with a hex editor etc) would you prefer that used for this / paste test, rather than the regex based parser?

No, I think what you added here is an improvement, much easier to see what's going on in the test with the escape sequences.

> Would you prefer a separate explicit paste test as well?

Yes, that would be nice, but not necessary.  Only if you have time.

> Taking a step back... the test I've added here seems to pass even without the fix in both `./run-tests.py cmdline/repl_autocomplete_underscore.py` and `./run-tests.py --via-mpy cmdline/repl_autocomplete_underscore.py` - is it only when frozen into the build it fails?

No, it's tricky.  It really depends on what qstrs are frozen and the order they appear in.  And the test behaviour may change if/when more scripts are frozen.  So I think we should just leave this PR as-is.

tests/cmdline/repl_autocomplete_underscore.py
```diff
@@ -0,0 +1,34 @@
+# Test REPL autocompletion filtering of underscore attributes
+# This test is designed to test behavior with frozen code when run with --via-mpy
```
dpgeorge: Please remove this second line comment.  It's not valid.

dpgeorge:
CI is failing again...

andrewleech:
> CI is failing again...

Fixed again; needed dos2unix on the exp files!.
I've added a repl paste test as well.

dpgeorge:
Thanks for fixing, it looks good now.

And thanks for the paste test, that's also looking good!

Note: doing tab completion on `import <tab>` will no longer print modules beginning with an underscore, eg `_thread`.  I guess that's unavoidable, and actually correct behaviour.

dpgeorge (approved):

PR #17109, opened 2025-04-10
py/emitinlinerv32: Move include of asmrv32.h to within feature guard.

dpgeorge:
### Summary

Otherwise, when compiling on 16-bit systems (where `mp_uint_t` is 16 bits wide) the compiler warns about "left shift count >= width of type", from the static inline functions that have RV32_ENCODE_TYPE_xxx macros which do a lot of bit shifting.

### Testing

Prior to this commit the `pic16bit` port would not compile (with XC 2.00).  With this commit it does compile.

CI will test other builds of this, when `MICROPY_EMIT_INLINE_RV32` is enabled.

### DIFF
--- a/py/emitinlinerv32.c
+++ b/py/emitinlinerv32.c
@@ -30,11 +30,12 @@
 #include <stdio.h>
 #include <string.h>
 
-#include "py/asmrv32.h"
 #include "py/emit.h"
 #include "py/misc.h"
 
 #if MICROPY_EMIT_INLINE_RV32
+
+#include "py/asmrv32.h"
 
 typedef enum {
 // define rules with a compile function


agatti:
Oh, thanks!  All those compile-time shifts and masks were mostly to help me in getting the bitmasks right (RV32 opcodes encoding isn't the most linear out there).

If this added complexity is no longer needed/desired I can send a PR that pre-bakes masks into constants for the next MPy release - there is a test suite exercising pretty much every opcode after all.

dpgeorge:
> If this added complexity is no longer needed/desired I can send a PR that pre-bakes masks into constants

I'm not sure what you mean by that?

An alternative fix to this PR would be to use `uint32_t` for all the arguments instead of `mp_uint_t`.  The latter can be 16-bit on 16-bit builds, but `uint32_t` is obviously always 32-bits.  But that's a very big change (in terms of LoC count).

agatti:
What I meant was to try to pre-compute shifted masks and the like and store the uint32_t equivalents in the code, basically doing the preprocessor's work (or, in other words, what the arm and thumb2 opcode emitters do).

That said, I believe the alternate fix you propose is still needed, as the bitmasks and whatnot do have a fixed 32-bits size regardless the processor the opcodes are generated on.  That may or may not become relevant for RV64 since opcodes are still 32 bits in size and RV32 opcodes are valid RV64 opcodes as well :)  But that's for the next release anyway...

dpgeorge:
> What I meant was to try to pre-compute shifted masks and the like and store the uint32_t equivalents in the code, basically doing the preprocessor's work

If that makes for smaller code, that could be a good change.  But at the same time it might not be worth the effort (there are lots of other interesting things to work on instead!).

> That said, I believe the alternate fix you propose is still needed, as the bitmasks and whatnot do have a fixed 32-bits size regardless the processor the opcodes are generated on. That may or may not become relevant for RV64 since opcodes are still 32 bits in size and RV32 opcodes are valid RV64 opcodes as well :)

OK, then indeed the type probably does need to be changed to `uint32_t`.

> But that's for the next release anyway...

Yes for sure.  Release v1.25.0 is imminent!

PR #17112, opened 2025-04-10
Add `mpremote fs tree` command to show a tree of the device's files.

Josverl:
### Summary
This PR adds a new command to `mpremote` that displays the file system of a connected device in a tree format.

While working on mpremote rm -r I developed a desire to see a tree-view of files on the device.
I build a PoC for this and [asked for feedback](https://github.com/orgs/micropython/discussions/17104) 
where there appears to be a common need for this.

The  `mpremote fs tree` command :
 - Shows treeview from current path or specified path
 -  uses the graph chars ("├── ", "└── ") ( not configurable)
 - options: 
    - -v/--verbose adds the serial device name to the top of the tree
    - -s/--size add a (human readable) size to the files. ( not using -h as that )

This PR also fixes a limitation where `ls` can be run without specifying a path, but this would result in an error if an option was specified.
As `tree` also uses the same logic I wanted to fix this for both commands.

**Example tree view:**
```
$ mpremote resume tree -sv /
tree :/
:/ on /dev/ttyACM0
├── [  10B] a.py
├── [ 1.9K] b.py
├── backup
│   ├── [  10B] a.py
│   ├── [ 1.9K] b.py
│   ├── foo
│   │   ├── [   0B] aa.py
│   │   └── [   0B] ba.py
│   └── lib
│       ├── [   0B] bar.py
│       └── [   0B] foo.py
├── [   0B] bar.py
├── foo
│   ├── [   0B] aa.py
│   └── [   0B] ba.py
├── [   0B] foo.py
├── lib
│   ├── [   0B] bar.py
│   └── [   0B] foo.py
└── ramdisk
    ├── [   0B] bar.py
    ├── [   0B] foo.py
    └── lib
        ├── [   0B] lib_bar.py
        └── [   0B] lib_foo.py
```

### Testing

Automated tests for valid and error cases have been added, but can only be verified on unix based systems.
The -v option cannot be tested reliably with the `.exp` output approach as devices can,  and will,  be assigned to different port names.  So this has been omitted from automated tests.

Note that there is now 3x duplication of the `ramdisk.py` script across multiple test scripts. 
It would be best to reduce this to a single file and use it in all tests.

Manual testing:
| port | WSL2/ubuntu| Windows|
|--------|--------|--------|
| stm32| PASS | PASS |
| samd | PASS | PASS |
| esp32| PASS | PASS | 

### Documentation

Documentation and samples for the `fs tree` command have been added to the mpremote doc page.
A few _sub sections_ have been added to the **fs section**  to improve readability, 
A few paragraphs have been moved to a more appropriate location to better fit this structure.

### DIFF
--- a/docs/reference/mpremote.rst
+++ b/docs/reference/mpremote.rst
@@ -234,6 +234,7 @@ The full list of supported commands are:
   - ``rmdir <dirs...>`` to remove directories on the device
   - ``touch <file..>`` to create the files (if they don't already exist)
   - ``sha256sum <file..>`` to calculate the SHA256 sum of files
+  - ``tree [-vsh] <dirs...>`` to print a tree of the given directories
 
   The ``cp`` command uses a convention where a leading ``:`` represents a remote
   path. Without a leading ``:`` means a local path. This is based on the
@@ -263,6 +264,13 @@ The full list of supported commands are:
   .. warning::
     There is no supported way to undelete files removed by ``mpremote rm -r :``.
     Please use with caution.
+
+  The ``tree`` command will print a tree of the given directories.
+  Using the ``--size/-s`` option will print the size of each file, or use
+  ``--human/-h`` to use a more human readable format.
+  Note: Directory size is only printed when a non-zero size is reported by the device's filesystem.
+  The ``-v`` option  can be used to include the name of the serial device in
+  the output.
 
   All other commands implicitly assume the path is a remote path, but the ``:``
   can be optionally used for clarity.
--- a/tools/mpremote/mpremote/commands.py
+++ b/tools/mpremote/mpremote/commands.py
@@ -334,6 +334,49 @@ def do_filesystem_recursive_rm(state, path, args):
             print(f"removed: '{path}'")
 
 
+def human_size(size, decimals=1):
+    for unit in ['B', 'K', 'M', 'G', 'T']:
+        if size < 1024.0 or unit == 'T':
+            break
+        size /= 1024.0
+    return f"{size:.{decimals}f}{unit}" if unit != 'B' else f"{int(size)}"
+
+
+def do_filesystem_tree(state, path, args):
+    """Print a tree of the device's filesystem starting at path."""
+    connectors = ("├── ", "└── ")
+
+    def _tree_recursive(path, prefix=""):
+        entries = state.transport.fs_listdir(path)
+        entries.sort(key=lambda e: e.name)
+        for i, entry in enumerate(entries):
+            connector = connectors[1] if i == len(entries) - 1 else connectors[0]
+            is_dir = entry.st_mode & 0x4000  # Directory
+            size_str = ""
+            # most MicroPython filesystems don't support st_size on directories, reduce clutter
+            if entry.st_size > 0 or not is_dir:
+                if args.size:
+                    size_str = f"[{entry.st_size:>9}]  "
+                elif args.human:
+                    size_str = f"[{human_size(entry.st_size):>6}]  "
+            print(f"{prefix}{connector}{size_str}{entry.name}")
+            if is_dir:
+                _tree_recursive(
+                    _remote_path_join(path, entry.name),
+                    prefix + ("    " if i == len(entries) - 1 else "│   "),
+                )
+
+    if not path or path == ".":
+        path = state.transport.exec("import os;print(os.getcwd())").strip().decode("utf-8")
+    if not (path == "." or state.transport.fs_isdir(path)):
+        raise CommandError(f"tree: '{path}' is not a directory")
+    if args.verbose:
+        print(f":{path} on {state.transport.device_name}")
+    else:
+        print(f":{path}")
+    _tree_recursive(path)
+
+
 def do_filesystem(state, args):
     state.ensure_raw_repl()
     state.did_action()
@@ -361,8 +404,8 @@ def do_filesystem(state, args):
         # leading ':' if the user included them.
         paths = [path[1:] if path.startswith(":") else path for path in paths]
 
-    # ls implicitly lists the cwd.
-    if command == "ls" and not paths:
+    # ls and tree implicitly lists the cwd.
+    if command in ("ls", "tree") and not paths:
         paths = [""]
 
     try:
@@ -404,6 +447,8 @@ def do_filesystem(state, args):
                     )
                 else:
                     do_filesystem_cp(state, path, cp_dest, len(paths) > 1, not args.force)
+            elif command == "tree":
+                do_filesystem_tree(state, path, args)
     except OSError as er:
         raise CommandError("{}: {}: {}.".format(command, er.strerror, os.strerror(er.errno)))
     except TransportError as er:
--- a/tools/mpremote/mpremote/main.py
+++ b/tools/mpremote/mpremote/main.py
@@ -181,7 +181,11 @@ def argparse_rtc():
 
 
 def argparse_filesystem():
-    cmd_parser = argparse.ArgumentParser(description="execute filesystem commands on the device")
+    cmd_parser = argparse.ArgumentParser(
+        description="execute filesystem commands on the device",
+        add_help=False,
+    )
+    cmd_parser.add_argument("--help", action="help", help="show this help message and exit")
     _bool_flag(cmd_parser, "recursive", "r", False, "recursive (for cp and rm commands)")
     _bool_flag(
         cmd_parser,
@@ -197,10 +201,26 @@ def argparse_filesystem():
         None,
         "enable verbose output (defaults to True for all commands except cat)",
     )
+    size_group = cmd_parser.add_mutually_exclusive_group()
+    size_group.add_argument(
+        "--size",
+        "-s",
+        default=False,
+        action="store_true",
+        help="show file size in bytes(tree command only)",
+    )
+    size_group.add_argument(
+        "--human",
+        "-h",
+        default=False,
+        action="store_true",
+        help="show file size in a more human readable way (tree command only)",
+    )
+
     cmd_parser.add_argument(
         "command",
         nargs=1,
-        help="filesystem command (e.g. cat, cp, sha256sum, ls, rm, rmdir, touch)",
+        help="filesystem command (e.g. cat, cp, sha256sum, ls, rm, rmdir, touch, tree)",
     )
     cmd_parser.add_argument("path", nargs="+", help="local and remote paths")
     return cmd_parser
@@ -355,6 +375,7 @@ _BUILTIN_COMMAND_EXPANSIONS = {
     "rmdir": "fs rmdir",
     "sha256sum": "fs sha256sum",
     "touch": "fs touch",
+    "tree": "fs tree",
     # Disk used/free.
     "df": [
         "exec",
@@ -552,8 +573,13 @@ def main():
                 command_args = remaining_args
                 extra_args = []
 
-            # Special case: "fs ls" allowed have no path specified.
-            if cmd == "fs" and len(command_args) == 1 and command_args[0] == "ls":
+            # Special case: "fs ls" and "fs tree" can have only options and no path specified.
+            if (
+                cmd == "fs"
+                and len(command_args) >= 1
+                and command_args[0] in ("ls", "tree")
+                and sum(1 for a in command_args if not a.startswith('-')) == 1
+            ):
                 command_args.append("")
 
             # Use the command-specific argument parser.
--- a/tools/mpremote/tests/test_filesystem.sh
+++ b/tools/mpremote/tests/test_filesystem.sh
@@ -237,3 +237,6 @@ echo -----
 # try to delete existing folder in mounted filesystem
 $MPREMOTE mount "${TMP}" + rm -rv :package || echo "expect error"
 echo -----
+# fs without command should raise error
+$MPREMOTE fs 2>/dev/null || echo "expect error: $?"
+echo -----
--- a/tools/mpremote/tests/test_filesystem.sh.exp
+++ b/tools/mpremote/tests/test_filesystem.sh.exp
@@ -272,3 +272,5 @@ rm :package
 mpremote: rm -r not permitted on /remote directory
 expect error
 -----
+expect error: 2
+-----
--- a/tools/mpremote/tests/test_fs_tree.sh
+++ b/tools/mpremote/tests/test_fs_tree.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+set -e
+
+# Creates a RAM disk big enough to hold two copies of the test directory
+# structure.
+cat << EOF > "${TMP}/ramdisk.py"
+class RAMBlockDev:
+    def __init__(self, block_size, num_blocks):
+        self.block_size = block_size
+        self.data = bytearray(block_size * num_blocks)
+
+    def readblocks(self, block_num, buf):
+        for i in range(len(buf)):
+            buf[i] = self.data[block_num * self.block_size + i]
+
+    def writeblocks(self, block_num, buf):
+        for i in range(len(buf)):
+            self.data[block_num * self.block_size + i] = buf[i]
+
+    def ioctl(self, op, arg):
+        if op == 4: # get number of blocks
+            return len(self.data) // self.block_size
+        if op == 5: # get block size
+            return self.block_size
+
+import os
+
+bdev = RAMBlockDev(512, 50)
+os.VfsFat.mkfs(bdev)
+os.mount(bdev, '/ramdisk')
+os.chdir('/ramdisk')
+EOF
+
+# setup 
+echo -----
+$MPREMOTE run "${TMP}/ramdisk.py"
+$MPREMOTE resume ls
+
+echo -----
+echo "empty tree"
+$MPREMOTE resume tree :
+
+echo -----
+$MPREMOTE resume touch :a.py + touch :b.py  
+$MPREMOTE resume mkdir :foo + touch :foo/aa.py + touch :foo/ba.py
+
+echo "small tree - :" 
+$MPREMOTE resume tree :
+
+echo -----
+echo "no path" 
+$MPREMOTE resume tree 
+
+echo -----
+echo "path = '.'" 
+$MPREMOTE resume tree .
+
+echo -----
+echo "path = ':.'" 
+$MPREMOTE resume tree :.
+
+
+echo -----
+echo "multiple trees" 
+$MPREMOTE resume mkdir :bar + touch :bar/aaa.py + touch :bar/bbbb.py
+$MPREMOTE resume mkdir :bar/baz + touch :bar/baz/aaa.py + touch :bar/baz/bbbb.py
+$MPREMOTE resume mkdir :bar/baz/quux + touch :bar/baz/quux/aaa.py + touch :bar/baz/quux/bbbb.py
+$MPREMOTE resume mkdir :bar/baz/quux/xen + touch :bar/baz/quux/xen/aaa.py
+
+$MPREMOTE resume tree
+
+echo -----
+echo single path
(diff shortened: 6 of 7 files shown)


Josverl:
Should be no problem to add both -s and -h

docs/reference/mpremote.rst
```diff
@@ -234,6 +234,9 @@ The full list of supported commands are:
   - ``rmdir <dirs...>`` to remove directories on the device
   - ``touch <file..>`` to create the files (if they don't already exist)
   - ``sha256sum <file..>`` to calculate the SHA256 sum of files
+  - ``tree [-vs] <dirs...>`` to print a tree of the given directories
```
dpgeorge, line 237: Please also document the `-h` option.

tools/mpremote/mpremote/main.py
```diff
@@ -197,10 +197,25 @@ def argparse_filesystem():
...
+    size_group = cmd_parser.add_mutually_exclusive_group()
+    size_group.add_argument(
+        "--size",
+        "-s",
+        default=False,
+        action="store_true",
+        help="show file size in bytes(tree command only)",
+    )
+    size_group.add_argument(
+        "--human",
```
dpgeorge, line 209: I suggest allowing this to take the short form `-h` as well as the long form, to match unix tree.

tools/mpremote/mpremote/main.py
```diff
@@ -552,8 +568,12 @@ def main():
                 command_args = remaining_args
                 extra_args = []
 
-            # Special case: "fs ls" allowed have no path specified.
-            if cmd == "fs" and len(command_args) == 1 and command_args[0] == "ls":
+            # Special case: "fs ls" and "fs tree" allowed have no path specified.
+            if (
+                cmd == "fs"
+                and command_args[0] in ["ls", "tree"]
```
dpgeorge, line 574: Please use a tuple for `("ls", "tree")`.

tools/mpremote/mpremote/main.py
Josverl: I tried initially, but that conflicts with the existing `-h/--help` of mpremote fs

I can change that, but that may be a breaking change to some of the wrapper modules although i think the risk is low.
Also de docs state clearly that it should not be done/ depended upon.

dpgeorge: Oh, right.  The `-h`/`--help` options are added automatically to all the subcommands...

I think it's fair enough to repurpose `-h` for whatever the subcommand needs, in this case it would be shorthand for `--human`.  Then there's always the long form `--help` to get help if the user needs that.

Josverl:
- [x] updated and rebased

tools/mpremote/mpremote/commands.py
```diff
@@ -334,6 +334,49 @@ def do_filesystem_recursive_rm(state, path, args):
...
+    def _tree_recursive(path, prefix=""):
+        entries = state.transport.fs_listdir(path)
+        entries.sort(key=lambda e: e.name)
+        for i, entry in enumerate(entries):
+            connector = connectors[1] if i == len(entries) - 1 else connectors[0]
+            is_dir = entry.st_mode & 0x4000  # Directory
+            if args.size and not is_dir:
+                print(f"{prefix}{connector}[{entry.st_size:>12}] {entry.name}")
+            elif args.human and not is_dir:
+                print(f"{prefix}{connector}[{human_size(entry.st_size):>8s}] {entry.name}")
```
dpgeorge, line 359: As above, the width here could probably be less.  I think 6 is enough to fit everything (without a space between the number and unit, see above).

tools/mpremote/mpremote/commands.py
```diff
@@ -334,6 +334,49 @@ def do_filesystem_recursive_rm(state, path, args):
             print(f"removed: '{path}'")
 
 
+def human_size(size, decimal_places=1):
+    for unit in ['B', 'K', 'M', 'G', 'T']:
+        if size < 1024.0 or unit == 'T':
+            break
+        size /= 1024.0
+    return f"{size:.{decimal_places}f} {unit}" if unit != 'B' else f"{int(size)} {unit}"
```
dpgeorge, line 342: Unix tree doesn't have a space between the number and unit, so maybe it doesn't need it here either?

Also, this won't work on boards without floating point support.  I think it's worth making it work on such boards.  Considering you only want 1 decimal place, you could multiply `size` by 10 to start with, then do `size // 10` and `size % 10` to format it.  Also, arguably you don't need the decimal if the integer part is >= 100 (only need 3 significant figures).

Josverl: > Also, this won't work on boards without floating point support.

This computation is not executed on the MCU though, or is it?

dpgeorge: > This computation is not executed on the MCU though, or is the intent for the host to lack FO capabilities?

Ah, you're right, this does not execute on the target MCU.

So, it's OK to use floats here.

tools/mpremote/mpremote/commands.py
```diff
@@ -334,6 +334,49 @@ def do_filesystem_recursive_rm(state, path, args):
...
+    # connectors = ("|-- ", "+-- ")
+    def _tree_recursive(path, prefix=""):
+        entries = state.transport.fs_listdir(path)
+        entries.sort(key=lambda e: e.name)
+        for i, entry in enumerate(entries):
+            connector = connectors[1] if i == len(entries) - 1 else connectors[0]
+            is_dir = entry.st_mode & 0x4000  # Directory
+            if args.size and not is_dir:
+                print(f"{prefix}{connector}[{entry.st_size:>12}] {entry.name}")
+            elif args.human and not is_dir:
```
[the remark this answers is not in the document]
Josverl: does work reliably across ports ?
I did not know for sure so I left it out , especially as `tree --help` in Ubuntu actually states : _"-s            Print the size in bytes of each file."_
but , it actually does print the size though ...
Ill add it in and test it across a few ports for consistency

tools/mpremote/mpremote/commands.py
```diff
@@ -334,6 +334,49 @@ def do_filesystem_recursive_rm(state, path, args):
...
+
+    # connectors = ("|-- ", "+-- ")
+    def _tree_recursive(path, prefix=""):
+        entries = state.transport.fs_listdir(path)
+        entries.sort(key=lambda e: e.name)
+        for i, entry in enumerate(entries):
+            connector = connectors[1] if i == len(entries) - 1 else connectors[0]
+            is_dir = entry.st_mode & 0x4000  # Directory
+            if args.size and not is_dir:
+                print(f"{prefix}{connector}[{entry.st_size:>12}] {entry.name}")
```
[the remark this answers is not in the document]
Josverl: I have been tinkering with the width as well. 
personally I think the space improves readability , but if the goal is full consistence with linux/ubuntu -h size should be 4 characters

<details><summary>Details</summary>
<p>


```
$ tree ports/zephyr/boards/ -h
[4.0K]  ports/zephyr/boards/
├── [  46]  96b_carbon.conf
├── [ 163]  bbc_microbit_v2.conf
├── [  97]  beagleconnect_freedom.conf
├── [  77]  disco_l475_iot1.conf
├── [ 264]  frdm_k64f.conf
├── [ 122]  frdm_kw41z.conf
├── [ 206]  mimxrt1050_evk.conf
├── [  33]  mimxrt685_evk_cm33.conf
├── [ 273]  nrf52840dk_nrf52840.conf
├── [ 146]  nucleo_h743zi.conf
├── [ 467]  nucleo_h743zi.overlay
├── [ 131]  nucleo_wb55rg.conf
├── [ 172]  qemu_cortex_m3.conf
├── [ 172]  qemu_x86.conf
├── [  81]  reel_board.conf
└── [  50]  rv32m1_vega_ri5cy.conf

0 directories, 16 files
(.venv) (.venv) jos@josverl-sb5:~/micropython$ tree ports/zephyr/boards/ -s
[       4096]  ports/zephyr/boards/
├── [         46]  96b_carbon.conf
├── [        163]  bbc_microbit_v2.conf
├── [         97]  beagleconnect_freedom.conf
├── [         77]  disco_l475_iot1.conf
├── [        264]  frdm_k64f.conf
├── [        122]  frdm_kw41z.conf
├── [        206]  mimxrt1050_evk.conf
├── [         33]  mimxrt685_evk_cm33.conf
├── [        273]  nrf52840dk_nrf52840.conf
├── [        146]  nucleo_h743zi.conf
├── [        467]  nucleo_h743zi.overlay
├── [        131]  nucleo_wb55rg.conf
├── [        172]  qemu_cortex_m3.conf
├── [        172]  qemu_x86.conf
├── [         81]  reel_board.conf
└── [         50]  rv32m1_vega_ri5cy.conf
....
│   │   └── [4.0K]  upstream
│   │       ├── [4.0K]  bin
│   │       │   ├── [   8]  clang -> clang-21
│   │       │   ├── [   5]  clang++ -> clang
│   │       │   ├── [130M]  clang-21
│   │       │   ├── [ 79M]  clang-scan-deps
│   │       │   ├── [ 72M]  lld
│   │       │   ├── [  15]  llvm-addr2line -> llvm-symbolizer
│   │       │   ├── [9.6M]  llvm-ar
│   │       │   ├── [1.1M]  llvm-cxxfilt
│   │       │   ├── [9.4M]  llvm-dwarfdump
│   │       │   ├── [ 33M]  llvm-dwp
│   │       │   ├── [ 11M]  llvm-nm
│   │       │   ├── [6.0M]  llvm-objcopy
│   │       │   ├── [ 17M]  llvm-objdump
│   │       │   ├── [   7]  llvm-ranlib -> llvm-ar
│   │       │   ├── [9.1M]  llvm-readobj
│   │       │   ├── [5.3M]  llvm-size
│   │       │   ├── [791K]  llvm-strings
│   │       │   ├── [  12]  llvm-strip -> llvm-objcopy
│   │       │   ├── [7.2M]  llvm-symbolizer
```

</p>
</details>

tools/mpremote/mpremote/commands.py
```diff
@@ -334,6 +334,49 @@ def do_filesystem_recursive_rm(state, path, args):
...
+    connectors = ("├── ", "└── ")
+
+    # connectors = ("|-- ", "+-- ")
+    def _tree_recursive(path, prefix=""):
+        entries = state.transport.fs_listdir(path)
+        entries.sort(key=lambda e: e.name)
+        for i, entry in enumerate(entries):
+            connector = connectors[1] if i == len(entries) - 1 else connectors[0]
+            is_dir = entry.st_mode & 0x4000  # Directory
+            if args.size and not is_dir:
```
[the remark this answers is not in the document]
Josverl: see below comment.
Changed code to improve alignment , but suppress showing folder sizes <= 0 for both -s and -h

tools/mpremote/mpremote/commands.py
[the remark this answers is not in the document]
dpgeorge: > if the goal is full consistence with linux/ubuntu -h

We don't need full consistency with the output from unix tree.  But following unix tree is a good place to start, then we can tweak to suit MicroPython's use cases.  9 for --size and 6 for --human looks good to me.

tools/mpremote/mpremote/commands.py
```diff
@@ -334,6 +334,50 @@ def do_filesystem_recursive_rm(state, path, args):
             print(f"removed: '{path}'")
 
 
+def human_size(size, decimals=1):
+    for unit in ['B', 'K', 'M', 'G', 'T']:
+        if size < 1024.0 or unit == 'T':
+            break
+        size /= 1024.0
+    return f"{size:.{decimals}f}{unit}" if unit != 'B' else f"{int(size)} {unit}"
```
dpgeorge, line 342: Did you intend to add the unit `B` in the case of bytes?  I think it can be left out, no unit means bytes.

tools/mpremote/mpremote/commands.py
```diff
@@ -334,6 +334,50 @@ def do_filesystem_recursive_rm(state, path, args):
...
+    def _tree_recursive(path, prefix=""):
+        entries = state.transport.fs_listdir(path)
+        entries.sort(key=lambda e: e.name)
+        for i, entry in enumerate(entries):
+            connector = connectors[1] if i == len(entries) - 1 else connectors[0]
+            is_dir = entry.st_mode & 0x4000  # Directory
+            size_str = ""
+            # most micropyhon filesystems don't support st_size on folders - reduce clutter
+            if args.size:
+                if entry.st_size > 0 or not is_dir:
```
dpgeorge, line 358: You could move this if check outside, to make it a bit simpler:
```py
if entry.st_size > 0 or not is_dir:
    if args.size:
        ...
    elif args.human:
        ...
```

tools/mpremote/mpremote/commands.py
```diff
@@ -361,8 +405,8 @@ def do_filesystem(state, args):
         # leading ':' if the user included them.
         paths = [path[1:] if path.startswith(":") else path for path in paths]
 
-    # ls implicitly lists the cwd.
-    if command == "ls" and not paths:
+    # ls and tree implicitly lists the cwd.
+    if command in ["ls", "tree"] and not paths:
```
dpgeorge, line 409: Please use tuple `("ls", "tree")`.

tools/mpremote/mpremote/commands.py
Josverl: The B for Bytes is actually supressed later so ot is not in the output
Indeed simpler not to add it in the first place

tools/mpremote/mpremote/commands.py
Josverl: No the intent was indeed to drop it.

tools/mpremote/mpremote/commands.py
Josverl: I'd swear I had done that before.
Sorry to make you repeat yourself.

Josverl:
@mattytrentini This was the reasoning 
| format| unix tree | mpremote tree | notes|
|--------|--------|-------|---|
| --size | 12 | 9 | smaller files on micropython
| --human | 4 | 6| allow for one decimal `[ 13.8K]` (for those smaller files) 

And one additional space coming up

tools/mpremote/mpremote/commands.py
dpgeorge: You did already do that, on a separate line :smile:

tools/mpremote/mpremote/main.py
```diff
@@ -552,8 +573,12 @@ def main():
                 command_args = remaining_args
                 extra_args = []
 
-            # Special case: "fs ls" allowed have no path specified.
-            if cmd == "fs" and len(command_args) == 1 and command_args[0] == "ls":
+            # Special case: "fs ls" and "fs tree" allowed have no path specified.
+            if (
+                cmd == "fs"
+                and command_args[0] in ("ls", "tree")
```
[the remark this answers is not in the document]
dpgeorge: For some reason this line is added but commented out :joy:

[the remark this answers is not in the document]
Josverl: Disabled it to test the test,
then never re-enabled it 🙃

Josverl:
Thanks for testing all the edge cases.

The [`and len(command_args) == 1` ](https://github.com/Josverl/micropython/blob/1c365473054dfc0c3c8879a03584e7e4f48993e3/tools/mpremote/mpremote/main.py#L579)  needed to be `>= 1` when using options without a path.

Added a test case for that as well.

docs/reference/mpremote.rst
```diff
@@ -234,6 +234,9 @@ The full list of supported commands are:
   - ``rmdir <dirs...>`` to remove directories on the device
   - ``touch <file..>`` to create the files (if they don't already exist)
   - ``sha256sum <file..>`` to calculate the SHA256 sum of files
+  - ``tree [-vsh] <dirs...>`` to print a tree of the given directories
+  
+  *mpremote fs cp*
```
dpgeorge, line 239: I'm not sure these sub-sub-headings make this section any clearer / easier to read.  They are rendered in italics and not easy to spot.

I also don't think it's necessary to have an example output of the tree command.  No other command has example output (only example usage of the command).

I think it's enough to just add a single paragraph about tree (eg after the bit about rm).

docs/reference/mpremote.rst
```diff
@@ -245,6 +248,20 @@ The full list of supported commands are:
   ``mpremote fs cp :main.py main.py`` copies ``main.py`` from the device back
   to the current directory.
 
+  .. code-block:: bash
+
+      $ mpremote fs cp main.py :main.py + repl
```
dpgeorge, line 253: This example should go below, the bit documenting the `+` character.

I think it would be better to just add a single paragraph for tree and not reorder anything else.

docs/reference/mpremote.rst
```diff
@@ -264,24 +281,37 @@ The full list of supported commands are:
...
-  The ``cp`` command supports the ``-r`` option to make a recursive copy.  By
-  default ``cp`` will skip copying files to the remote device if the SHA256 hash
-  of the source and destination file matches.  To force a copy regardless of the
-  hash use the ``-f`` option.
+  All other commands implicitly assume the path is a remote path, but the ``:``
+  can be optionally used for clarity.
+
+  All of the filesystem sub-commands take multiple path arguments, so if there
+  is another command in the sequence, you must use ``+`` to terminate the
+  arguments, e.g.
```
dpgeorge, line 314: This "eg" is missing its example code.  As mentioned above, this would be fixed by not reordering the file, just add a paragraph about tree.

docs/reference/mpremote.rst
```diff
@@ -264,24 +281,37 @@ The full list of supported commands are:
...
+        tree :
+        : /on /dev/ttyACM0
+        ├── [   90B] config.py
+        ├── data
+        │   └── [   122K] norwegian.log
+        ├── lib
+        │   ├── [  2.4K] parrot.py
+        │   └── [  3.4K] worker.py
+        ├── [  800B] main.py
+        └── [  2.5K] readme.md
```
dpgeorge, line 305: As mentioned above, I don't think this example is needed.

dpgeorge:
Thanks for updating the code, it works well now.

The last thing to do is finalise the docs changes.

docs/reference/mpremote.rst
Josverl: I'm in conflict on some of your comments .
- the example can go , no issue there 
I think that mpremote is under documented, and that the doc is hard to follow in places.
I tried to improve that by adding the sub headings, but indeed italics is not too clear.

As mpremote bundles a lot of different functionality I was considering that some parts , such as the `fs` command , and perhaps the `mip` commands are better documented in their own, linked, reference page. Similar to the structure for modules and Classes.

While the fs commands are probably obvious to a linux user and need no further explanation, I still find (as someone reared on Windows)  need to lookup detailed syntaxes, and when they are not in the mpremote docs , I need to know to search for a man page and mentally filter that. 

That would allow for some more room , without getting very long pages , allow to add some more detail about often used commands.

Perhaps such a restructure should be part of a different PR , or I can give it a try in a isolated commit.

dpgeorge: > I think that mpremote is under documented, and that the doc is hard to follow in places.
>       I tried to improve that by adding the sub headings, but indeed italics is not too clear.

OK, that's a fair comment.

Let's improve the documentation in a separate PR.  Trying to do that here means possibly blocking this PR for a lot longer.

So I suggest to just add the one-liner `tree [-vsh] <dirs...>` here (as you currently have it), and a single extra paragraph about how `fs tree` works.  Then that's good enough to start with.

Another PR can expand and improve this documentation.

Josverl:
@dpgeorge , just FYI, the docs have been reversed to the simpler format you requested

dpgeorge (approved):
Thanks for updating.

Issue #17103, opened 2025-04-09
esp32c3 default I2C initializers hang the REPL.


ricksorensen:
### Port, board and/or hardware

XIAO ESP32C3 

### MicroPython version

MicroPython v1.25.0-preview.531.gef8282c71 on 2025-04-09; XIAOC3 with ESP32C3

### Reproduction

```python
# no I2C devices attached
from machine import I2C
i2=I2C(0)
```

### Expected behaviour

Return of valid I2C object.

### Observed behaviour

REPL hangs

### Additional Information

My experience with a XIAO C3 running with the GENERIC_C3 build:
MicroPython v1.25.0-preview.531.gef8282c71 on 2025-04-09; XIAOC3 with ESP32C3
is that both `i2=I2C()` and `i2=I2C(0)` hang the REPL with no return, requiring a hardware reset.

Using 1.24.1 release build, `i2=I2C(0)` hangs the REPL as above, but `i2=I2C()` fails with

```
Warning: I2C(-1, ...) is deprecated, use SoftI2C(...) instead
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'scl' argument required
>>> i2=I2C(0)
```
I will look at what is causing the hang. Note that on the XIAO S3 both `i2=I2C()` and `i2=I2C(0)` work. Maybe because there are 2 HW I2C channels available?

### Code of Conduct

Yes, I agree

ricksorensen:
The error seems to be a result of using GPIO18 and GPIO19 for the default SCL and SDA.  These are default C3 USB-JTAG pins- which would explain the REPL hangs.  Simple fix  - add the defines for default SCL/SDA to the  mpconfigboard.h.  For example for my XIAO C3 I added 
```
#define MICROPY_HW_I2C0_SCL                 (7)
#define MICROPY_HW_I2C0_SDA                 (6)
```
to the `mpconfigboard.h` for the ESP32_GENERIC_C3 and the default I2C settings now behave.  

Probably a documentation fix as opposed to a managed code fix?

ricksorensen:
The error seems to be a result of using GPIO18 and GPIO19 for the default SCL and SDA.  These are default C3 USB-JTAG pins- which would explain the REPL hangs.  Simple fix  - add the defines for default SCL/SDA to the  mpconfigboard.h.  For example for my XIAO C3 I added 
```
#define MICROPY_HW_I2C0_SCL                 (7)
[...]

PR #17113, opened 2025-04-10
ESP32_GENERIC_C3/mpconfigboard.h: Add defaults for I2C pins.

ricksorensen:
### Summary

For the ESP32C3  (XIAO C3 for example) the default I2C pins conflicted with the
USB pins resulting in REPL hanging after using the new I2C initializer without
parameters. 

Reference:
#17103
#16671

### Testing

I tested on an XIAO ESP32C3 board, using the micropython master development branch.

While the values I defaulted to should allow for safe construction of  an I2C object,
they may not be correct for any particular wiring.  
I would suggest the documentation be updated for board configurations to include the 
necessary default pin parameters. 

Another valid approach would be to require the parameters (port id, scl, sda) for every
I2C object creation - my favorite for consistency.

### DIFF
--- a/ports/esp32/boards/ESP32_GENERIC_C3/mpconfigboard.h
+++ b/ports/esp32/boards/ESP32_GENERIC_C3/mpconfigboard.h
@@ -20,6 +20,6 @@
    Without defining, SDL/SDA default is 19/18 which
    conflict with USB
 **/
-#define I2C pin defaults
+// define I2C pin defaults
 #define MICROPY_HW_I2C0_SCL                 (7)
 #define MICROPY_HW_I2C0_SDA                 (6)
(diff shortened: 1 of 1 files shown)


dpgeorge:
I suggest to fix it in `machine_i2c.c` in a more generic way.  Similar to how the default pins for I2C1 are different for esp32 vs other MCUs.

Eg:
```c
#ifndef MICROPY_HW_I2C0_SCL                                             
#if CONFIG_IDF_TARGET_ESP32C3                                      
#define MICROPY_HW_I2C0_SCL (GPIO_NUM_6) 
#define MICROPY_HW_I2C0_SDA (GPIO_NUM_7) 
#else
#define MICROPY_HW_I2C0_SCL (GPIO_NUM_18)
#define MICROPY_HW_I2C0_SDA (GPIO_NUM_19)
#endif
#endif                 
```

PR #17132, opened 2025-04-15
ports/samd: Some small changes and bug fixes.

robert-hh:
### Summary

Another service pack with small changes and bug fixes.

-  SAMD_GENERIC_Dxxx/board.json: Add the Microchip URL.
-  samd/machine_i2c.c: Add the timeout keyword argument to the constructor to make it compliant to the documentation. The default value is now 50000us instead of 100ms.
- samd/modtime.c: Change time.time_ns() to follow the RTC time. That is done by adding the offset to epoch, following the scheme from the RP2 port. RTC and ticks_us() are not precisely in sync, and so the difference between time.time_ns()/1e9 and time_time() will increase by about 9 seconds/24h. So applications should avoid using time.time() and time.time_ns() in the same context.
- samd/SAMD51: Fix VFS settings for the internal flash. 
     + The wrong loader script was assigned for SAMD_GENERIC_D51X20, causing the VFS block count to be wrong.
    + Change the VFS block size from 1536 to 2048. With the setting of 1536, writing more that 1536 bytes at once failed. This applies to SAMD_GENERIC_D51X19 and SAMD_GENERIC_D51X20. No other SAMD51 board uses the internal flash for the file system.
- samd/samd_spiflash.c: Improve the flash type detection.
    + Check properly for the SFDP header signature.
    + Use the flash size information from SFDP, if present.
    + Use a JEDEC-ID table for special flash types instead of a series of conditional statements. Add two more special flash chips <= 1 MByte without SFDP.  
    + Add a compile flag MICROPY_HW_SPIFLASH_SIZE to set the size in mpconfigboard.h, which replaces getting the size from the JEDEC ID or the SFDP record in case that both do not return reasonable values.

### Testing

- Tested the I2C change for the timeout to change and be shown when the instance is printed.
- Tested with the SPIflash boards and various flash types.
- Tested SAMD51 internal flash system by writing and re-reading files with random data of various length.

- The SPIflash for flash better flash type detection increases the flash size by 64 bytes.
- The change of the SAMD51 VFS block size increase the RAM buffer accordingly. No problem for SAMD51.


dpgeorge (approved):
These changes/fixes look good, thanks!

I smoke tested the PR on an ADAFRUIT_ITSYBITSY_M0_EXPRESS, it still works.

robert-hh:
Thank you.

PR #8255, opened 2022-02-03
esp32/uart: Allow limited configuration of (repl) uart0.

andrewleech:
For many applications the hardcoded 115200 repl buadrate on esp32 is painfully slow. This function allows this to be adjusted dynamically.

UART0 is blocked from access via the existing machine.UART interface, presumable for good reason: https://github.com/micropython/micropython/blob/a3675294ae39a724aec6959238c45af0dea92f21/ports/esp32/machine_uart.c#L297

So this just provides a direct function via the `esp32` module to set just the baudrate.

Hey @UnexpectedMaker this should tick that box you were talking about :-)

PR #17135, opened 2025-04-16
esp32/machine_uart: Correctly manage UART queue and event task.

DvdGiessen:
### Summary

If the driver was reinitialised while there was already an event task running the queue that task is trying to receive from would be deleted, causing it to try to take a lock that no longer existed and deadlocking the CPU.

This change ensures the task is always shut down before recreating the queue and recreates the task afterwards.

It also:

- Allows setting an IRQ handler before the UART is initialized (like other ports allow)
- Removes the task when the UART is deinitialized (which was previously missing).
- Adds a check that no event task can be started when no queue exists (i.e. when the UART is already deinitialized or it's the built-in REPL UART which doesn't have a queue configured).
- Adds a check to prevent reinitialising the UART driver unnecessarily (we only need to reinitialize it when the buffer sizes change).

### Testing

Tested on three custom boards (1x original ESP32, 2x ESP32S3) to talk to three modem chips (1x Sequans, 2x different SIMCOM) all using the UART IRQ's to trigger the PPP.poll() method.

### DIFF
--- a/ports/esp32/machine_uart.c
+++ b/ports/esp32/machine_uart.c
@@ -59,6 +59,7 @@
 #define UART_IRQ_BREAK (1 << UART_BREAK)
 #define MP_UART_ALLOWED_FLAGS (UART_IRQ_RX | UART_IRQ_RXIDLE | UART_IRQ_BREAK)
 #define RXIDLE_TIMER_MIN (5000)  // 500 us
+#define UART_QUEUE_SIZE (3)
 
 enum {
     RXIDLE_INACTIVE,
@@ -174,6 +175,13 @@ static void uart_event_task(void *self_in) {
     }
 }
 
+static inline void uart_event_task_create(machine_uart_obj_t *self) {
+    if (xTaskCreatePinnedToCore(uart_event_task, "uart_event_task", 2048, self,
+        ESP_TASKD_EVENT_PRIO, (TaskHandle_t *)&self->uart_event_task, MP_TASK_COREID) != pdPASS) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("failed to create UART event task"));
+    }
+}
+
 static void mp_machine_uart_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     machine_uart_obj_t *self = MP_OBJ_TO_PTR(self_in);
     uint32_t baudrate;
@@ -250,7 +258,7 @@ static void mp_machine_uart_init_helper(machine_uart_obj_t *self, size_t n_args,
     // wait for all data to be transmitted before changing settings
     uart_wait_tx_done(self->uart_num, pdMS_TO_TICKS(1000));
 
-    if (args[ARG_txbuf].u_int >= 0 || args[ARG_rxbuf].u_int >= 0) {
+    if ((args[ARG_txbuf].u_int >= 0 && args[ARG_txbuf].u_int != self->txbuf) || (args[ARG_rxbuf].u_int >= 0 && args[ARG_rxbuf].u_int != self->rxbuf)) {
         // must reinitialise driver to change the tx/rx buffer size
         #if MICROPY_HW_ENABLE_UART_REPL
         if (self->uart_num == MICROPY_HW_UART_REPL) {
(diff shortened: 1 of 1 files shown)


ports/esp32/machine_uart.c
```diff
@@ -59,6 +59,7 @@
 #define UART_IRQ_BREAK (1 << UART_BREAK)
 #define MP_UART_ALLOWED_FLAGS (UART_IRQ_RX | UART_IRQ_RXIDLE | UART_IRQ_BREAK)
 #define RXIDLE_TIMER_MIN (5000)  // 500 us
+#define UART_QUEUE_SIZE (3)
```
DvdGiessen, line 62: This number 3 is the exact same magic number that was previously used when IRQ RXIDLE support was introduced in #14041. I only moved it to a define since it's now used in two places. I don't know why it's 3 and didn't see any mention in the previous MR about that, but decided to not change it since it seems to work fine.

ports/esp32/machine_uart.c
```diff
@@ -568,6 +587,15 @@ static const mp_irq_methods_t uart_irq_methods = {
 };
 
 static mp_irq_obj_t *mp_machine_uart_irq(machine_uart_obj_t *self, bool any_args, mp_arg_val_t *args) {
+    #if MICROPY_HW_ENABLE_UART_REPL
+    if (self->uart_num == MICROPY_HW_UART_REPL) {
+        mp_raise_ValueError(MP_ERROR_TEXT("UART does not support IRQs"));
+    }
+    #endif
+    if (self->uart_queue == NULL && args[MP_IRQ_ARG_INIT_handler].u_obj != mp_const_none) {
```
DvdGiessen, line 595: This does allow setting a None IRQ handler, which is to prevent cleanup code from raising errors needlessly.

For example if used with PPP when closing the PPP connection the cleanup will happen asynchronously triggered by a callback from the lwIP TCP/IP thread, which would normally unset the IRQ it previously configured on it's UART object. If the UART object was already cleaned up as well (deinited or GC collected) that cleanup callback would encounter a ValueError and subsequently crash because the lwIP thread can't handle a non-linear return. (The fact PPP/lwIP breaks like that is a separate issue this PR is not addressing.)

ports/esp32/machine_uart.c
```diff
@@ -275,9 +276,20 @@ static void mp_machine_uart_init_helper(machine_uart_obj_t *self, size_t n_args,
...
+            recreate_event_task = true;
+        }
         check_esp_err(uart_driver_delete(self->uart_num));
+        self->uart_queue = NULL;
         check_esp_err(uart_param_config(self->uart_num, &uartcfg));
-        check_esp_err(uart_driver_install(self->uart_num, self->rxbuf, self->txbuf, 0, NULL, 0));
+        check_esp_err(uart_driver_install(self->uart_num, self->rxbuf, self->txbuf, UART_QUEUE_SIZE, &self->uart_queue, 0));
+        if (recreate_event_task) {
+            xTaskCreatePinnedToCore(uart_event_task, "uart_event_task", 2048, self,
+                ESP_TASKD_EVENT_PRIO, (TaskHandle_t *)&self->uart_event_task, MP_TASK_COREID);
```
dpgeorge, line 291: Is it worth putting this in a (inline) helper function, now that it's called twice in this file?

ports/esp32/machine_uart.c
```diff
@@ -568,6 +587,15 @@ static const mp_irq_methods_t uart_irq_methods = {
 };
 
 static mp_irq_obj_t *mp_machine_uart_irq(machine_uart_obj_t *self, bool any_args, mp_arg_val_t *args) {
+    #if MICROPY_HW_ENABLE_UART_REPL
+    if (self->uart_num == MICROPY_HW_UART_REPL) {
+        mp_raise_ValueError(MP_ERROR_TEXT("UART does not support IRQs"));
```
dpgeorge, line 592: Why isn't this possible?  I would think IRQ still works on the UART REPL.

ports/esp32/machine_uart.c
dpgeorge: Hmm, I wonder if it is useful to be able to configure the IRQ before initialising?  It looks like other ports (eg stm32) support this.

I guess here it's quite hard, because the event queue won't exist until the UART is initialised.  Could the event task be started anyway and then wait until the event queue became non-NULL?  Maybe then the task could also stay alive over a UART re-init and just use a different queue (I guess it'd need to have a signal to stop waiting on the queue and then wait for a new queue to become available).

Alternatively, could set a flag in this function that the event task should be started when the UART is next initialised, so the IRQ starts working after initialisation.

But, maybe this is a very niche use case (configure IRQ before initialising) and can be left for a future enhancement, if anyone ever needs it.

ports/esp32/machine_uart.c
DvdGiessen: Currently that's because of that we never initialize the UART queue for the REPL UART. See https://github.com/micropython/micropython/pull/8255. This error is a more specific case of the error raised below: Namely `self->uart_queue` will always be `NULL` for the REPL UART. For clarity I'll move it into the check below.

dpgeorge: OK, I see.  So user/Python IRQs on the UART REPL didn't work prior to this change anyway.

ports/esp32/machine_uart.c
DvdGiessen: Also updated the PR and commit description to note this addition.

dpgeorge (approved):
Tested with ESP32_GENERIC firmware, running the `machine_uart_*.py` tests.

Everything passes!

PR #17137, opened 2025-04-16
lib/littlefs: Update LittleFS to v2.10.2.

DvdGiessen:
### Summary

This updates LittleFS from 2.8.1 to 2.10.2. It brings a few different bugfixes and has better POSIX-like path parsing (accepting trailing slashes for directories, empty string is no longer interpreted as the root directory (`/`)). See the release notes on [their GitHub releases page](https://github.com/littlefs-project/littlefs/releases) for more.

The new version does have support for replacing the built-in CRC32 function, so I was able to save a few bytes by reusing the already existing function from the `uzlib` library used for the deflate module:

Code size change by reusing CRC32 function (second commit):
```
   unix x64:  -160 standard
      stm32:  -120 PYBV10
     mimxrt:  -120 TEENSY40
       samd:  -120 ADAFRUIT_ITSYBITSY_M4_EXPRESS
```

Getting this to work did require fixing a small issue in `lfs2_util.h`, which I've also [submitted upstream](https://github.com/littlefs-project/littlefs/pull/1095).

### Testing

I've been running updated LittleFS on most of my boards since they were released. The changes in v2.9.3 are included in a production release firmware build that has had extensive testing, and the more recent v2.10 changes I've been running on at least 5+ different custom ESP32-based boards since beginning of 2025.

Going above the root directory (`/../foo`) now gives an error.

This is an intentional change made by LittleFS:
> This actually moves away from the behavior of other filesystems, but towards, in my opinion, more reasonable behavior. It should also be more consistent with future theoretical openat-like APIs.

Note that the POSIX spec allows this. In [previous versions](https://pubs.opengroup.org/onlinepubs/9699919799.2016edition/xrat/V4_xbd_chap04.html#tag_21_04_13) it explicitly mentioned it was "implementation-defined", in the [last version](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap04.html#tag_04_16) it uses the word "may":
> As a special case, in the root directory, dot-dot *may* refer to the root directory itself.

(Emphasis mine.)

This required a update of the testsuite, and is a (minor) compatibility break.


dpgeorge (approved):
Thanks for this, everything looks good here.

I think it's acceptable to follow littlefs and now raise an exception when trying to change a directory up out of the filesystem.

dpgeorge:
Tested on an RPI_PICO2 using `VfsLfs2` and it passes the tests, and the filesystem still works.

PR #17144, opened 2025-04-17
py/emitnative: Viper load/store code generation improvements.

agatti:
### Summary

This PR contains a few improvements to the Viper code generator, focussing on load/store code generation.

The original implementation was able to generate optimised sequences only for a subset of supported architectures and operation data sizes.  Optimised sequence generation has been extended to include the following:

* 8/16/32 bit register-indexed loads for Arm
* 8/16/32 bit register-indexed load/stores for Thumb
~~* 8/16/32 bit integer-indexed loads for Thumb with ArmV7 opcodes support~~
~~* 16/32 bit register-indexed load/stores for RV32 with Zba opcodes support~~

~~The latter entry introduces a configuration option to the RV32 code generation backend to optionally enable support for address generation opcodes ([`SH1ADD` and `SH2ADD`](https://github.com/riscv/riscv-isa-manual/blob/e5078e55ea977ef197066dcd3f7c8e09032cb348/src/b-st-ext.adoc?plain=1#L3045)), which perform a scaled addition operation (used for calculating halfword and word offsets).  Support for these two opcodes is disabled by default in the QEMU build due to the CI image's QEMU version not being recent enough to support those opcodes in the first place.~~

~~Since Zba opcodes are officially supported by the RP2350 ("3.8.1.15. Zba: Bit manipulation (address generation)"), the rp2 port will now generate optimised address generation code for Viper sequences if MicroPython is built for RV32.~~

*The RV32 changes are no longer part of this PR and will show up in another PR along with changes to `mpy-cross`.*
*Armv7 improvements are no longer part of this PR and will show up in anther PR along with more invasive changes to the Viper emitter*

As a side-effect of these improvements, the Viper generator code involved was also cleaned up and made more generic, moving architecture-specific operation implementations into their own backends.  The generic implementation for register-indexed load and stores is retained as a fallback if a backend does not provide an optimised implementation.  Plus, a bug in `asm_thumb_add_reg_reg_offset` was found and fixed - I guess not many people used Viper code with large lookup tables.

Some extra tests for this specific part of the Viper emitter have also been added, trying to exercise the code at boundary conditions when it comes to which optimised code sequence should be emitted.

### Testing

Viper-specific tests were executed successfully on a NodeMCU v2 ESP2866 board, and on QEMU for SABRELITE and MPS2_AN385 boards.  ~~The latter QEMU board was also tested with `MICROPY_EMIT_RV32_ZBA` enabled on a more recent version of QEMU (9.2.3) than what's available in the CI image.  My RP2350 is out of order so I was not able to test Zba-specific opcodes on a real board, although the generated sequences work on QEMU.~~

This code will probably behave differently than the original version when it comes to array accesses with an invalid integer index.  Viper didn't implement the indexing from the end of the array semantics with negative indices anyway, so in theory this should not be an issue as invalid code would still be invalid.  Still, this may or may not be something worth keeping in mind.

Moreover, there's a small complexity increase (along with its increased binary footprint) for certain native code emitter backends, but I believe the improved code generation capabilities of the Viper backend make up for that.  I'm still actively trying to reduce footprint to a minimum without making the whole lot unreadable, though.


ports/rp2/mpconfigport.h
```diff
@@ -109,6 +109,7 @@
 #endif
 #elif PICO_RISCV
 #define MICROPY_EMIT_RV32                       (1)
+#define MICROPY_EMIT_RV32_ZBA                   (1)
```
dpgeorge, line 112: This might also need to be enabled in `mpy-cross`, otherwise it won't use the extensions.

But of course that means we probably need an option to `mpy-cross` to enable/disable it, so it can work with targets that don't have this extension.  And this starts to make things quite complicated!

Is it possible to just enable this unconditionally in `mpy-cross`, and yet the CI (qemu tests) still pass??

tests/micropython/viper_ptr16_load.py
```diff
@@ -35,3 +40,27 @@ def memadd2(src_in) -> int:
 print(get(b), get1(b))
 print(memadd(b, 2))
 print(memadd2(b))
+
+# Test boundary conditions for various architectures
+
+import gc
+
+gc.collect()
```
dpgeorge, line 48: Is this `gc.collect()` necessary?

Actually I think it's best to split all these new tests out into separate files, eg `viper_ptr16_load_boundary.py`.  Because targets like esp8266 already have enough issues importing large-ish files with native code.

ports/rp2/mpconfigport.h
dpgeorge: > I thought `mpy-cross` didn't compile `@micropython.viper` functions

Yet, it definitely can.  It can do all of `@micropython.native`, `@micropython.viper` and `@micropython.asm_XXX`.

There are already options like `-march=armv7emsp` that you can pass to `mpy-cross` to select the correct arch, so I guess they need to be extended to select sub-features of RV32.

But that's definitely out of scope for this PR.

tests/micropython/viper_ptr16_load.py
agatti: The PR now has the boundary tests in their own files, thanks for the suggestion!

ports/rp2/mpconfigport.h
agatti: I'm aware of `mpy-cross` compiling inline asm code, I just assumed that viper code was geared towards interactive/compiled-at-runtime usage and thus left alone.  Good to know though :)

I've removed the RV32 code from this PR so it can be part of another one that will also bring in more extensive `mpy-cross` changes.

agatti:
I'm temporarily marking this as draft as I found a better way to clean changed emitnative code up - I'll mark it as ready for review once I'm done with that.

tests/micropython/viper_ptr16_store_boundary.py
```diff
@@ -0,0 +1,37 @@
...
+    (127, (0x4847, 0x4A49, 0x4C4B)),
+    (2047, (0x4E4D, 0x504F, 0x5251)),
+)
+
+for start, vals in test_data:
+    for i, v in enumerate(vals):
+        set_index(b, start + i, v)
+        print(b[(start + i) * 2 : (start + i + 1) * 2])
+
+gc.collect()
```
dpgeorge, line 23: What is the intention of this `gc.collect()`?  It's not really doing anything, as far as I can tell.  The `b` bytearray won't be collected because it's still reachable via the global variable `b`.

Why not just reuse `b` for the second part of the test, instead of reallocating it?  If it needs to be cleared out to all 0's, just do a loop to set each element to 0.

(Same comment applies to the other tests that do `gc.collect()`).

agatti: I guess I thought that global variables would be reclaimed anyway, since at the point of collection there are no other entities referencing it, maybe I got my wires crossed with how other GC languages I'm working with operate.

Now it loops over the array setting each element to 0, so this should be sorted.  Would a PR that adds `bytearray.clear` be useful?

tests/micropython/viper_ptr16_store_boundary.py
```diff
@@ -0,0 +1,37 @@
...
+import gc
+
+
+@micropython.viper
+def set_index(dest: ptr16, i: int, val: int):
+    dest[i] = val
+
+
+b = bytearray(5000)
+test_data = (
```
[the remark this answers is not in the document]
agatti: Good call, this should be fixed with the latest iteration of the PR - thanks!

tests/micropython/viper_ptr16_store_boundary.py
dpgeorge: It may be able to reclaim the memory in the following case:
```py
b = bytearray(5000)

foo(b)

# erase the reference to b
b = None

# this may reclaim the memory from b, if there are no stray pointers to b left around
gc.collect()
```

But, in general, it's better to write tests that don't need to rely on GC behaviour.



> Would a PR that adds `bytearray.clear` be useful?

Not really.  `bytearray.clear` doesn't set the elements to 0, it deletes all elements and leaves you with a zero-length bytearray.

py/emitnative.c
```diff
@@ -1873,27 +1868,28 @@ static void emit_native_store_subscr(emit_t *emit) {
...
+                        break;
+                    }
+                    #elif N_RV32
+                    if (FIT_SIGNED(index_value, 10)) {
+                        asm_rv32_opcode_sw(emit->as, reg_value, reg_base, index_value << 2);
+                        break;
+                    }
+                    #elif N_XTENSA || N_XTENSAWIN
+                    if (index_value >= 0 && index_value < 256) {
+                        asm_xtensa_s32i_optimised(emit->as, reg_value, reg_base, index_value);
```
dpgeorge, line 1883: I see you've fixed a subtle bug here: `reg_value` used to be `REG_RET`!

Nice one.

dpgeorge (approved):
Tested on PYBD_SF6 and RPI_PICO2_W-RISCV, running all native tests.  Everything passes.

Issue #17141, opened 2025-04-16
Json parser accepts uncomplete json objects


asdil12:
### Port, board and/or hardware

rp2, unix

### MicroPython version

1.24.1

### Reproduction

Micropython
```python
>>> buffer = b'{"getEnergy_0": 977.222, "getVelocity_1": 0.0, "getHeading_0": 1'
>>> import json
>>> json.loads(buffer)
{'getHeading_0': 1, 'getEnergy_0': 977.222, 'getVelocity_1': 0.0}
```

### Expected behaviour

cPython:
```python
>>> buffer = b'{"getEnergy_0": 977.222, "getVelocity_1": 0.0, "getHeading_0": 1'
>>> import json
>>> json.loads(buffer)
Traceback (most recent call last):
  File "<python-input-2>", line 1, in <module>
    json.loads(buffer)
    ~~~~~~~~~~^^^^^^^^
  File "/usr/lib64/python3.13/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "/usr/lib64/python3.13/json/decoder.py", line 345, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/json/decoder.py", line 361, in raw_decode
    obj, end = self.scan_once(s, idx)
               ~~~~~~~~~~~~~~^^^^^^^^
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 1 column 65 (char 64)
```

### Observed behaviour

See above

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

dpgeorge:
Fixed by 9ef16b466bf980a3b3e0980e7d010b0b9130569b

PR #17148, opened 2025-04-17
extmod/modjson: Detect unterminated composite entities.

agatti:
### Summary

This PR makes the JSON parser raise an exception when handling objects or arrays whose declaration is incomplete, as in missing the closing marker (brace or bracket) and if the missing marker would have been the last non-whitespace character in the incoming string.

Since CPython's JSON parser would raise an exception in such a case, unlike MicroPython's, this PR aligns MicroPython's behaviour with CPython.

This commit fixes issue #17141.

### Testing

This was tested with the Unix port, passing all JSON-related tests (existing and new).

There may be some code out there that relies on this peculiar behaviour, and thus this PR may introduce an incompatibility across MicroPython versions.

On the other hand, the code wouldn't have worked in CPython in the first place as an incomplete JSON object is not meant to be parseable by `json.load` or `json.loads` anyway since they expect a fully formed JSON object to be available (either from a string or from a file descriptor).

### DIFF
--- a/extmod/modjson.c
+++ b/extmod/modjson.c
@@ -160,7 +160,8 @@ static mp_obj_t mod_json_load(mp_obj_t stream_obj) {
     for (;;) {
     cont:
         if (S_END(s)) {
-            break;
+            // Input finished abruptly in the middle of a composite entity.
+            goto fail;
         }
         mp_obj_t next = MP_OBJ_NULL;
         bool enter = false;
--- a/tests/extmod/json_loads.py
+++ b/tests/extmod/json_loads.py
@@ -71,3 +71,27 @@ try:
     my_print(json.loads("[null]   a"))
 except ValueError:
     print("ValueError")
+
+# incomplete object declaration
+try:
+    my_print(json.loads('{"a":0,'))
+except ValueError:
+    print("ValueError")
+
+# incomplete nested array declaration
+try:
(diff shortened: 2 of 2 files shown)


extmod/modjson.c
```diff
@@ -336,6 +336,7 @@ static mp_obj_t mod_json_load(mp_obj_t stream_obj) {
             }
         }
     }
+    goto fail;
```
dpgeorge, line 339: This is very neat!  A simple change, very low cost.

But, I can't help but think it would be better to put this `goto` up on line 163, with a comment about what's happening:
```c
    ...
    for (;;) {
    cont:
        if (S_END(s)) {
            // Input finished abruptly in the middle of a composite entity.
            goto fail;
        }
        mp_obj_t next = MP_OBJ_NULL;
        ...
```

That should be equivalent, and easier to reason about.

agatti: The rationale for choosing that location in my commit was to separate the failure action from the main processing loop - as in, "the whole input buffer was consumed, execute the next step in the process", otherwise it might have gotten lost in a fairly visually dense bit of code.

I agree a comment should have been there to begin with though :)

Still, your request is reasonable and indeed equivalent code-wise, I've updated the PR to reflect those changes.

dpgeorge (approved):

dpgeorge:

PR #17149, opened 2025-04-17
Mpr/fix rm remote

### LINKED ISSUE #17147: "mpremote -rv :/" can remove files from host

any

### MicroPython version

mpremote v1.25.0

### Reproduction

- use mpremote mount . 
- Ctrl-C to terminate connection 
- run `mpremote rm -rv :`

Minimal repro: ⚠️ destructive ⚠️
`mpremote mount . + rm -rv :`

### Expected behaviour

-rm -r :/ not to affect the local filesystem

### Observed behaviour

apparently the mount was retained - and therefore the `/remote` , hosts filesystem,  started be removed 
```
rm :
removed directory: './.venv/Lib/site-packages/Deprecated-1.2.18.dist-info'
removed directory: './.venv/Lib/site-packages/adodbapi/examples'
removed directory: './.venv/Lib/site-packages/adodbapi/test'
removed directory: './.venv/Lib/site-packages/adodbapi'
removed directory: './.venv/Lib/site-packages/annotated_types-0.7.0.dist-info/licenses'
removed directory: './.venv/Lib/site-packages/annotated_types-0.7.0.dist-info'
removed directory: './.venv/Lib/site-packages/annotated_types/__pycache__'
removed directory: './.venv/Lib/site-packages/annotated_types'
removed directory: './.venv/Lib/site-packages/ansicon-1.89.0.dist-info'
removed directory: './.venv/Lib/site-packages/ansicon'
removed directory: './.venv/Lib/site-packages/anyio-4.9.0.dist-info'
removed directory: './.venv/Lib/site-packages/anyio/_backends'
removed directory: './.venv/Lib/site-packages/anyio/_core'
removed directory: './.venv/Lib/site-packages/anyio/abc'
removed directory: './.venv/Lib/site-packages/anyio/streams'
removed directory: './.venv/Lib/site-packages/anyio'

```

### Additional Information

I think `\remote` should be special cased in `mpremote rm -r` to guard against this behavior  

### Code of Conduct

Yes, I agree

Josverl:
### Summary

Removes the risk of inadvertently deleting files from the host by preventing the deletion of files from the host via `rm -r` on the `/remote` vfs as reported in #17147.

### Testing

The PR adds a test for `rm -r` on the `/remote` vfs to ensure that the deletion of files from the host is prevented.
Tested manually on Windows.

The fix only addresses the risk via `rm -r` as that has a potentially large impact, and not with just the `rm` command that has been in use for quite some time.
The same safeguard could be added to the `rm' command.

Fixes: https://github.com/micropython/micropython/issues/17147

### DIFF
--- a/tools/mpremote/mpremote/commands.py
+++ b/tools/mpremote/mpremote/commands.py
@@ -303,6 +303,17 @@ def do_filesystem_recursive_cp(state, src, dest, multiple, check_hash):
 
 def do_filesystem_recursive_rm(state, path, args):
     if state.transport.fs_isdir(path):
+        if state.transport.mounted:
+            r_cwd = state.transport.eval("os.getcwd()")
+            abs_path = os.path.normpath(
+                os.path.join(r_cwd, path) if not os.path.isabs(path) else path
+            )
+            if isinstance(state.transport, SerialTransport) and abs_path.startswith(
+                f'{SerialTransport.fs_hook_mount}/'
+            ):
+                raise CommandError(
+                    f"rm -r not permitted on {SerialTransport.fs_hook_mount} directory"
+                )
         for entry in state.transport.fs_listdir(path):
             do_filesystem_recursive_rm(state, _remote_path_join(path, entry.name), args)
         if path:
--- a/tools/mpremote/mpremote/transport_serial.py
+++ b/tools/mpremote/mpremote/transport_serial.py
@@ -42,6 +42,8 @@ from .transport import TransportError, TransportExecError, Transport
 
 
 class SerialTransport(Transport):
+    fs_hook_mount = "/remote"  # MUST match the mount point in fs_hook_code
+
     def __init__(self, device, baudrate=115200, wait=0, exclusive=True, timeout=None):
         self.in_raw_repl = False
         self.use_raw_paste = True
@@ -375,7 +377,11 @@ class SerialTransport(Transport):
         self.serial = self.serial.orig_serial
 
         # Provide a message about the remount.
-        out_callback(bytes(f"\r\nRemount local directory {self.cmd.root} at /remote\r\n", "utf8"))
+        out_callback(
+            bytes(
+                f"\r\nRemount local directory {self.cmd.root} at {self.fs_hook_mount}\r\n", "utf8"
+            )
+        )
 
         # Enter raw REPL and re-mount the remote filesystem.
         self.serial.write(b"\x01")
@@ -392,7 +398,7 @@ class SerialTransport(Transport):
 
     def umount_local(self):
         if self.mounted:
-            self.exec('os.umount("/remote")')
+            self.exec(f'os.umount("{self.fs_hook_mount}")')
             self.mounted = False
             self.serial = self.serial.orig_serial
 
@@ -413,7 +419,7 @@ fs_hook_cmds = {
     "CMD_RMDIR": 13,
 }
 
-fs_hook_code = """\
+fs_hook_code = f"""\
 import os, io, struct, micropython
 
 SEEK_SET = 0
@@ -746,8 +752,8 @@ class RemoteFS:
 
 
 def __mount():
-    os.mount(RemoteFS(RemoteCommand()), '/remote')
-    os.chdir('/remote')
+    os.mount(RemoteFS(RemoteCommand()), '{SerialTransport.fs_hook_mount}')
+    os.chdir('{SerialTransport.fs_hook_mount}')
 """
 
 # Apply basic compression on hook code.
--- a/tools/mpremote/tests/test_filesystem.sh
+++ b/tools/mpremote/tests/test_filesystem.sh
@@ -234,3 +234,6 @@ $MPREMOTE resume rm -r -v :/ramdisk
 $MPREMOTE resume ls :/ramdisk
 
 echo -----
+# try to delete existing folder in mounted filesystem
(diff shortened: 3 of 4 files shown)


AJMansfield:
The code seems like a correct implementation of this feature -- in that it gates this behind `state.transport.mounted`, so it won't make `rm -r` behave unpredictably when a user happens to have a directory by that name for some other reason -- but it does still duplicate the assumption that the path the host filesystem is mounted at will always be `/remote/`.

There should only be a single source of truth for what that path's name is. Even if it's ultimately still a hardcoded string, it should be one string in one place -- not a first copy of that string in the code setting up the mount, and a second copy of that string in the rm check.

This _could_ be done dynamically using `vfs.mount()` from #16939 (included in 1.25.0); though to keep compatibility with older firmware versions that might need to be augmented with some other bookkeeping strategy.

Josverl:
I would be more than happy to catch the 98% of cases where the remote vfs is created through mpremote, 
and leave the 2% case of very advanced users, that create their own vfs and backing host service, to their own devices.

The current Dual SOT is the `__mount()` method that is part of a "string_embedded" micropython script in transport_serial.py 
https://github.com/micropython/micropython/blob/master/tools/mpremote/mpremote/transport_serial.py#L748-L750

doing an AST parse of that to extract "/remote" would be a bit much IMO, 
I do not think that an f-string would be that much clearer either .

A KISS solution would be to add a `FS_HOOK_MOUNT = "/remote"` with a comment and a reference to transport_serial.py#L748-L750

AJMansfield:
> A KISS solution would be to add a `FS_HOOK_MOUNT = "/remote"` , or a class attribute `SerialTransport.fs_hook_mount` with a comment and a reference to transport_serial.py#L748-L750

Either of those seem pretty reasonable to me.

AJMansfield:
> The fix only addresses the risk via `rm -r` as that has a potentially large impact, and not with just the `rm` command that has been in use for quite some time.
> The same safeguard could be added to the `rm' command.

Safeguarding this on the command-issuer side feels like a somewhat backwards approach to the problem, though. The unix-like idiom to use for this type of protection would be to bail out with an EROFS 'Read-only file system' error _at the filesystem layer_ (and to then allow a `-f`/`--force` flag that circumvents this).


This error could potentially be triggered on the mpremote side inside the fs hook. To do that, would mean needing to make the presence or absence of `--force` available on the device, which is doable. Perhaps not this implementation strategy specifically, but on the mpremote side, something like:

```python
@contextlib.contextmanager
def fs_flags_set_on_remote(state, args):
    state.transport.exec(f'__mpremote_args={vars(args)!r}')
    try:
        yield
    finally:
        state.transport.exec('del __mpremote_args')


with fs_flags_set_on_remote(state, args):
    ... # actual command functions
```

Then in the `RemoteFS` class in the hook, use that to determine if we're in an mpremote command context and trigger the error from the actual removal operations themselves:

```python
class RemoteFS:
    ...

    def _err_if_unforced(err, path):
        global __mpremote_args
        try:
            if __mpremote_args['force'] is False:
                raise OSError(err, path)
        except NameError:
            pass # still allow other programmatic removal

    def remove(self, path):
        self._err_if_unforced(30, path) # errno.EROFS
        ...
    
    def rmdir(self, path):
        self._err_if_unforced(30, path) # errno.EROFS
        ...
```

This could also be added to `open` to stop another less-common but still undesirable snag: the way `mpremote mount ./local/ + fs cp ./local/a.py :/remote/a.py` ends up deleting the contents of `a.py`.

```
    def open(self, path, mode):
        if any(flag in mode for flag in 'wxa+'):
            self._err_if_unforced(errno.EROFS, path)
        ...
```

Though, that still wouldn't prevent the reverse operation from blanking the file (i.e. `mpremote mount ./local/ + fs cp :/remote/a.py ./local/a.py`).
Maybe it'd be better to push this back even further, to the mpremote side where it answers the hook's commands? Since that would create an opportunity to compare the actual resolved paths for the fs command and filesystem operation to stop that copy error; and it would mean we could generate _much_ more helpful error messages about it, e.g.:

```
$ mpremote mount ./local/ + fs cp :/remote/a.py ./local/a.py
mpremote: cp: ':/remote/a.py' and './local/a.py' are the same file
$ mpremote mount ./local/ + fs cp :/a.py :/remote/a.py
mpremote: cp: ':/remote/a.py' is a host file, use 'mpremote cp :/a.py ./local/a.py'
$ mpremote mount ./local/ + fs cp :/remote/a.py :/remote/b.py
mpremote: cp: ':/remote/a.py' and ':/remote/b.py' are host files, use 'cp ./local/a.py ./local/b.py'
$ mpremote mount ./local/ + fs rm :/remote/a.py
mpremote: rm: ':/remote/a.py' is a host file, use 'rm ./local/a.py'
```

dpgeorge:
I'd advocate to keep this change/fix simple, and the PR here is pretty simple.

Although eliminating the duplicated `/remote` path would be good, something simple like the following suggestion from above sounds reasonable:

> A KISS solution would be to add a FS_HOOK_MOUNT = "/remote" , or a class attribute SerialTransport.fs_hook_mount with a comment and a reference to transport_serial.py#L748-L750

Josverl:
Updated to use `SerialTransport.fs_hook_mount="/remote"` for all references to the mountpoint as used by mpremote.

dpgeorge (approved):

PR #17171, opened 2025-04-23
stm32: Add support for STM32N6xx MCUs and two N6 boards

dpgeorge:
### Summary

This PR adds preliminary support for ST's new STM32N6xx MCUs.

Supported features of this MCU so far are:
- basic clock tree initialisation, running at 800MHz
- fully working USB
- mboot support (required, because there's no internal flash)
- XSPI in memory-mapped mode
- `machine.Pin`
- `machine.UART`
- SD card
- filesystem
- ROMFS
- WiFi and BLE via cyw43-driver (SDIO and UART backend)

Supported boards:
- NUCLEO_N657X0
- ~~STM32N6570_DK~~ Edit: this board is no longer supported, can be added later if needed
- OPENMV_N6

Note that the N6 does not have internal flash, and has some tricky boot sequence, so using a custom bootloader (mboot) is almost a necessity.

~~The ST CMSIS and HAL files are added verbatim here, but will eventually be moved into `stm32lib`.~~  Edit: N6 CMSIS and HAL files are now in `stm32lib`.

OpenMV have generously sponsored the development of this port.

### Testing

This PR has been tested on the two N6 boards that are added here.


ports/stm32/boards/OPENMV_N6/board.c
```diff
@@ -0,0 +1,261 @@
...
+
+#include "mboot/mboot.h"
+
+#else
+
+void board_early_init(void) {
+    // TODO: if (HAL_PWREx_ConfigSupply(PWR_EXTERNAL_SOURCE_SUPPLY ) != HAL_OK)
+    //xspi_flash_init();
+}
+
```
iabdalkader, line 260:
```suggestion

#define OMV_BOOT_MAGIC_ADDR   (0x3401FFFCU)
#define OMV_BOOT_MAGIC_VALUE  (0xB00710ADU)

void board_enter_bootloader(void) {
    *((uint32_t *) OMV_BOOT_MAGIC_ADDR) = OMV_BOOT_MAGIC_VALUE;
    SCB_CleanDCache();
    NVIC_SystemReset();
}
```
To enter our bootloader.

dpgeorge: Do you want to have the OPENMV_N6 board definition included in MicroPython's stm32 port (like with the OPENMV_AE3)?  If so, it should work with mboot.  But probably also a good idea to work with the OpenMV bootloader, which means adding this code.

Related: how did you choose address 0x3401FFFC?  That's part way through SRAM1, in the FLEXRAM area.  You'd need to make sure that isn't cleared on reset like the rest of SRAM1/2, and also make sure it's not overwritten by the bootloader.

iabdalkader: > Do you want to have the OPENMV_N6 board definition included in MicroPython's stm32 port (like with the OPENMV_AE3)? If so, it should work with mboot. But probably also a good idea to work with the OpenMV bootloader, which means adding this code.

Either way is fine with me. If mboot support is kept, I can always use `MP_CONFIGFILE` to override and not build mboot when building our firmware.

> Related: how did you choose address 0x3401FFFC? That's part way through SRAM1, in the FLEXRAM area. You'd need to make sure that isn't cleared on reset like the rest of SRAM1/2, and also make sure it's not overwritten by the bootloader.

Good point. The bootloader uses the 128K (or 64K) DTCM for its memory (heap, stack etc..), so this address is typically `0x2001FFFCU` (last word of bootloader's memory) and it's the same for almost all boards. However, for some reason, this doesn't work on the N6, so I just used SRAM1. Note that SRAM1 and SRAM2 don't seem to get erased on reset, otherwise this wouldn't work, but it would still be better to use DTCM for this, but it's not working.

iabdalkader: Oh I think I just needed a DSB, seems the write might be buffered. The following works too:

```C
#define OMV_BOOT_MAGIC_ADDR   (0x3001FFFCU)
#define OMV_BOOT_MAGIC_VALUE  (0xB00710ADU)

void board_enter_bootloader(void) {
    *((uint32_t *) OMV_BOOT_MAGIC_ADDR) = OMV_BOOT_MAGIC_VALUE;
    __DSB();
    NVIC_SystemReset();
}
```

iabdalkader: @dpgeorge There seems to be some issue with DTCM, I can't access any address above ~0x1000. This causes a fault on boot when it tries accesses the magic number address (oddly enough, it doesn't always happen). Perhaps it has something to do with security config, flexram, clocks or something else that gets enabled by the main firmware, but even from the main firmware I still can't access this memory from gdb.  Anyway let's please keep the original boot address (with cache invalidate).

ports/stm32/boards/OPENMV_N6/mpconfigboard.mk
```diff
@@ -12,15 +12,26 @@ STM32_N6_HEADER_VERSION = 2.3
...
 
+# MicroPython settings
+MICROPY_PY_BLUETOOTH ?= 1
+MICROPY_BLUETOOTH_NIMBLE ?= 1
+MICROPY_BLUETOOTH_BTSTACK ?= 0
+MICROPY_PY_LWIP = 1
+MICROPY_PY_NETWORK_CYW43 = 1
+MICROPY_PY_SSL = 1
+MICROPY_SSL_MBEDTLS = 1
+MICROPY_VFS_LFS2 = 1
```
iabdalkader, line 37: I need to be able to disable mboot `USE_MBOOT ?= 1` and also we never use LFS2, just fat.

ports/stm32/usbd_conf.c
```diff
@@ -220,6 +229,23 @@ void HAL_PCD_MspInit(PCD_HandleTypeDef *hpcd) {
...
+        __HAL_RCC_USB1_OTG_HS_PHY_RELEASE_RESET();
+        __HAL_RCC_USB1_OTG_HS_RELEASE_RESET();
+
+        LL_AHB5_GRP1_EnableClock(LL_AHB5_GRP1_PERIPH_OTG1);
+        LL_AHB5_GRP1_EnableClock(LL_AHB5_GRP1_PERIPH_OTGPHY1);
+        LL_AHB5_GRP1_EnableClockLowPower(LL_AHB5_GRP1_PERIPH_OTG1);
+        LL_AHB5_GRP1_EnableClockLowPower(LL_AHB5_GRP1_PERIPH_OTGPHY1);
+
+        // Select 24MHz clock.
+        MODIFY_REG(USB1_HS_PHYC->USBPHYC_CR, USB_USBPHYC_CR_FSEL, 2 << USB_USBPHYC_CR_FSEL_Pos);
```
iabdalkader, line 245: There's a proper init sequence for this in the examples in the HAL, if you want to take a look. I think it's more or less the same, but there were some delays, more force_reset/release etc...

Also, could you please add this? I use it in HS mode.

```C
diff --git a/ports/stm32/usbdev/class/inc/usbd_cdc_msc_hid.h b/ports/stm32/usbdev/class/inc/usbd_cdc_msc_hid.h
index 3a87896b4..5906f95e0 100644
--- a/ports/stm32/usbdev/class/inc/usbd_cdc_msc_hid.h
+++ b/ports/stm32/usbdev/class/inc/usbd_cdc_msc_hid.h
@@ -11,7 +11,7 @@
 
 // Work out if we should support USB high-speed device mode
 #if MICROPY_HW_USB_HS \
-    && (!MICROPY_HW_USB_HS_IN_FS || defined(STM32F723xx) || defined(STM32F733xx))
+    && (!MICROPY_HW_USB_HS_IN_FS || defined(STM32F723xx) || defined(STM32F733xx) || defined(STM32N6))
 #define USBD_SUPPORT_HS_MODE (1)
 #else
 #define USBD_SUPPORT_HS_MODE (0)
```

dpgeorge: I've added the `USBD_SUPPORT_HS_MODE` option as above.

But testing it with #15909 shows that there is data corruption when HS is enabled.  That needs investigation.

dpgeorge:
I finally got the filesystem working.  Needed to make sure all relevant code and data structures to erase/write SPI flash is in RAM, and that interrupts are fully disabled during erase/write (obvious but tricky to do).

ports/stm32/boards/OPENMV_N6/mpconfigboard.mk
iabdalkader: But what if you don't want to build mboot, yet still want a flash-based image? Can `USE_MBOOT=1` define something like `MICROPY_FLASH_BASED=1`? For example:
```
USE_MBOOT ?= 1
MICROPY_FLASH_BASED ?= $(USE_MBOOT)
```

 This way I can define `USE_MBOOT=0 MICROPY_FLASH_BASED=1` to get a flash-based image without mboot.

iabdalkader:
I pulled this PR and tested it again. Things still seem to be working, and I noticed some improvements (like no longer calling `xspi_init()` from ioctl). However, I still couldn’t get the storage working. I get a hardfault when trying to initialize the XSPI PSRAM after `storage_init()` has been called. I did move the drivers/IRQs to ITCM, following what's done here, but I think it's a different issue, maybe because because it remaps XSPI back in 111 mode, when the bus is released, which is not what we use initially.

As for CYW43, I just skipped testing it altogether, as it's still using SPI. Other than that, I don’t think this is ready to coexist with older MCUs yet, for example, it comments out a large chunk of powerctrl.c, so we should probably wait for now.

ports/stm32/boards/common_n6_flash.ld
```diff
@@ -0,0 +1,83 @@
...
+SECTIONS
+{
+    .isr_vector :
+    {
+        . = ALIGN(4);
+        __isr_vector_ram_start = .;
+
+        KEEP(*(.isr_vector)) /* Startup code */
+
+        /* These functions need to run from ram to enable uart
```
iabdalkader, line 28: I assume this is because IRQs are being disabled for too long? If that’s the case, would it help to re-enable IRQs after each block erase/write operation, rather than keeping them disabled across all blocks?  If I’m not wrong, this is how it's done in the mimxrt and rp2 ports, and it might help avoid the need to move all those functions and their data to RAM.

dpgeorge:
> Yes it's the remap, if I change it to 888 it works but then it fails to create a filesystem

What read mode do you use for 888?  Do you enable the flash in single or double data rate for octal mode?

Note that all SPI flash data structures must be in RAM as well.  Check that your `mp_spiflash_config_t` is placed in RAM.

> Also, I'm not able to move memcpy to RAM, why is this needed?

memcpy and memset are sometimes used by the compiler to optimise code, and it's hard to stop the compiler doing that.  I'm not sure exactly where memcpy is used but that was necessary for me to get it waking up from deepsleep.

iabdalkader:
> What read mode do you use for 888? Do you enable the flash in single or double data rate for octal mode?
> Note that all SPI flash data structures must be in RAM as well. Check that your `mp_spiflash_config_t` is placed in RAM.

The bootloader maps the flash in octal DTR. Yes, I copied the `isr_vector` linker script section plus `*xspi.o(.text* .rodata*);`. I made sure they're in ITCM by checking the firmware map. I get a hardfault with 111 on the very first instruction of the first function call after the remap. If I uncomment and use 888, this doesn't happen, but fatfs fails to create a file system. I see that XSPI attempts to read (write?) in 111 later, so maybe that's why.

> memcpy and memset are sometimes used by the compiler to optimise code, and it's hard to stop the compiler doing that. I'm not sure exactly where memcpy is used but that was necessary for me to get it waking up from deepsleep.

`memcpy` hardfaults when moved to ram, I think it's because this is libc's `memcpy`, so maybe it has other deps that need to be moved too. Something here might work: https://chatgpt.com/share/683bf9a7-7a94-8013-9a7e-b4b3c91705dc

dpgeorge:
I nearly have 8-8-8 octal DTR mode working with the filesystem.  Just a few things to tidy up...

iabdalkader:
> I nearly have 8-8-8 octal DTR mode working with the filesystem. Just a few things to tidy up...

I have it working here, but I use HAL drivers: https://github.com/openmv/openmv/blob/2206dcb31c2a854c79e83cd62d6b55939f6c351a/boot/src/ports/stm32/stm32_xspi.c#L502

dpgeorge:
> I have it working here, but I use HAL drivers

Yes, thanks, I was studying that code (you use `HAL_XSPI_INSTRUCTION_8_BITS` when I think it should be `HAL_XSPI_INSTRUCTION_16_BITS`?).

iabdalkader:
> Yes, thanks, I was studying that code (you use `HAL_XSPI_INSTRUCTION_8_BITS` when I think it should be `HAL_XSPI_INSTRUCTION_16_BITS`?).

Yes, I think you're right. It should have been a test like the others:
```C
(dtr) ? HAL_XSPI_INSTRUCTION_16_BITS : HAL_XSPI_INSTRUCTION_8_BITS
```

It's weird but it seems to be working fine either way, but I've fixed it anyway. 

Note that we'll be using an 8/24MBs split for the FS/ROMFS (for both N6 and AE3). It's just more useful to have a bigger ROMFS for models. I'll send a PR later to update the AE3.

EDIT: For the N6, it's actually 4MBs (bootloader + firmware) 4MBs filesystem 24MBs ROMFS.

dpgeorge:
I have now finished XSPI 888 DTR mode support.

Mboot now uses it by default, and the main application supports both SPI and DTR mode and can auto-detect which mode is in use and continue to work in that mode (so the bootloader can configure it in 111 SPI mode or 888 DTR mode and the application will work).

I also got wake-from-deepsleep working, which was tricky to get the XSPI reconfigured in 888 DTR upon waking (see `board_leave_standby()` for details).

ROMFS is now also working!

I tested filesystem reads and writes extensively.  Reading from a FAT filesystem is now about 4x faster in 888 DTR mode compared with SPI mode.

dpgeorge:
Regarding `memcpy`: this really needs to go in RAM because when reading data from XSPI (eg filesystem read) it uses a `memcpy` to copy the data out of XSPI.  So if `memcpy` lives in XSPI then, although it will work, it will be slower due to fetching instructions from the `memcpy` code in XSPI and also the source data from a different location in XSPI.  Thus it's best to put `memcpy` in RAM.

ports/stm32/boards/common_n6_flash.ld
```diff
@@ -0,0 +1,83 @@
...
+        *(.text.storage_systick_callback)
+        *(.text.SysTick_Handler)
+        *(.text.uart_irq_handler)
+        *(.text.UART*_IRQHandler)
+        *(.text.USART*_IRQHandler)
+        *(.text.HAL_GetTick)
+        *drivers/memory/spiflash.o(.text* .rodata*)
+        *boards*(.rodata.spiflash_config*)
+
+        . = ALIGN(4);
```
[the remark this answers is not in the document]
dpgeorge: There are quite a few functions that need to go in RAM, and even some HAL functions (which are inline... but the compiler still makes them non-inline).  So I think it's simpler to configure that in a linker script.

There's already asm code in `resethandler_m3_iram.s` that does the copy.  The code in `main.c` that copies to RAM was already there before N6 support.

ports/stm32/boards/common_n6_flash.ld
dpgeorge: Actually, the UART functions don't need to go in RAM anymore.  I've fully disabled IRQs during XSPI flash erase/write.  It didn't seem to work if I just raised the IRQ priority level.

And the code in `spibdev.c` does enable IRQs after each block erase.

iabdalkader:
Fantastic, thanks! I'll give it another try as soon as I can and let you know how it goes.

> So if `memcpy` lives in XSPI then, although it will work, it will be slower due to fetching instructions from the `memcpy` code in XSPI and also the source data from a different location in XSPI. Thus it's best to put `memcpy` in RAM.

The RW filesystem doesn't really need to be fast, because there are two other fast options: ROMFS and SDCARD. However, will it really be that slow? The N6 models are basically relocatable binaries (data + code) stored in XSPI. I don't see any difference between running it in XIP mode (code in XSPI) and running it in copy mode (code copied to RAM). I think the icache does a pretty good job of hiding this.

Anyway, I'll try it again. But if it doesn't work, I can't think of any other fix for this (short of placing the entire libc in memory, if that is indeed the issue).

EDIT: Actually it's not an issue, we use custom linker scripts anyway, so I could just not move it to RAM, as long as things will work if it's placed in flash.

ports/stm32/boards/common_n6_flash.ld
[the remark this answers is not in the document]
iabdalkader: > even some HAL functions (which are inline... but the compiler still makes them non-inline).

Could it be the `-Os` ?

> There's already asm code in resethandler_m3_iram.s that does the copy. The code in main.c that copies to RAM was already there before N6 support.

So code in main is not really needed? Just for my own understanding, why do we need so much code (including HAL GPIO functions) in RAM? Is it needed when waking up from STOP mode? I assume standby doesn't really need that because it resets on wakeup.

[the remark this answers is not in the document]
dpgeorge: I've now cleaned this IRAM stuff up, and there's now only a single RAM section that needs to be copied.  That copy is done in asm in `resethandler_m3_iram.s`.




> Just for my own understanding, why do we need so much code (including HAL GPIO functions) in RAM? Is it needed when waking up from STOP mode? I assume standby doesn't really need that because it resets on wakeup.

Code in RAM is needed for two things:
- Writing to flash, which needs all the SPI flash logic from drivers/memory/spiflash.o and XSPI code
- Waking from **standby**!  The N6 has SRAM1 retention during standby and a special SYSCFG register that you can set to tell the MCU where to resume when waking from standby.  This allows it to wake up much faster from standby by executing code straightaway from SRAM1, instead of going through the bootloaders.  So there needs to be enough code in RAM to be able to act like a mini bootloader that can do just enough to enable XSPI memory mapped mode.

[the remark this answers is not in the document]
kwagyeman: @dpgeorge - This is for light sleep or deepsleep? Deepsleep should always result in a reboot after waking up. Awesome for light sleep though.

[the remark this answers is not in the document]
dpgeorge: This is for deepsleep!  It doesn't do a reboot after waking but rather executes from retention SRAM1.  That's why it's so complicated to make it work.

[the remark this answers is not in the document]
kwagyeman: Mmm, should there be an argument for the new behavior? While technically better, the bootloader running after deep sleep is exited could be something folks expect. Defer to Ibrahim's thoughts here. Petty cool.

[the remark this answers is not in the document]
dpgeorge: I'm not sure it's possible to run the ST ROM bootloader.  At least, I couldn't get that working initially.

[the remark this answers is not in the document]
dpgeorge: At the very least, the SPI flash needs to be reset to take it out of 888 DTR mode, if we want to run the ST ROM bootloader.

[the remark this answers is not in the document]
kwagyeman: Ah... do you have the reset fix on the board?

Removing these two parts fixes an issue with reset not resetting the flash:

image

After they are removed the flash should reset correctly on NVIC reset.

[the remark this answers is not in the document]
dpgeorge: My board has the capacitor (on the right) but not the resistor (on the left).

But I don't think that will make a difference, the device is not NVIC reset on wake from deepsleep/standby, so the flash remains in its configuration (I think).  Therefore you can't just jump directly to the ST ROM bootloader.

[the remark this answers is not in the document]
kwagyeman: Oh, okay, then I already updated your board to 10nF cap. So, it's not an issue with the reset line.

Your HW is fine then.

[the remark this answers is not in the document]
kwagyeman: https://www.st.com/resource/en/application_note/an5946-how-to-optimize-lowpower-modes-on-stm32n6-mcus-stmicroelectronics.pdf

Damien, did you see this doc?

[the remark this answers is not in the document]
iabdalkader: @dpgeorge @kwagyeman Thanks for explaining and for the app note! The app note seems to imply that this is optional, so if not retained it should NVIC reset. Perhaps this is worth a support question? My goal here is to minimize the RAM code to just xspi.c, to eventually use gcc section attribute on all functions and data that file, and then all we need in our firmware is a ram_func section and copy code to the startup code.

> Writing to flash, which needs all the SPI flash logic from drivers/memory/spiflash.o and XSPI code

Is `spiflah.o` really needed in RAM or is it a speed optimization? Could we get away with just the XSPI code as it does the actual erase/write?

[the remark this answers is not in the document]
dpgeorge: > Is `spiflah.o` really needed in RAM or is it a speed optimization? Could we get away with just the XSPI code as it does the actual erase/write?

Yes it's needed.  That code has all the logic to set WREN and wait for the erase to complete, for example.

Also, as mentioned, if the SPI flash is in DTR mode then it won't work with the ST ROM bootloader if the SPI flash is not fully reset upon waking from standby/deepsleep.  And I don't think it is fully reset when waking from standby.  If so, that would need code in SRAM1 in retention mode to reset it, before doing an NVIC reset (if that even works, I couldn't get it to bounce into the ST ROM bootloader).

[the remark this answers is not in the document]
dpgeorge: It's only using about 4k of SRAM1 at the moment.  Is that really such an issue?  We could try to optimise that down a bit.  Eg maybe it doesn't need the app ISR moved there (although I'd suggest doing that regardless, for IRQ performance reasons.)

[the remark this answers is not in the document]
iabdalkader: > if the SPI flash is in DTR mode then it won't work with the ST ROM bootloader if the SPI flash is not fully reset upon waking from standby/deepsleep

The boards are designed such that they reset the flash on NVIC reset (same should apply for devkits, we're following the reference design here), so it should fully reset the flash if an NVIC reset occurs. This feature is crucial for them to function, otherwise `machine.reset()` wouldn't work. So normally you just need to ensure an NVIC reset occurs somehow. How about using the address of something like this instead? Would this simplify things a bit?

```C
__attribute__((noreturn, section(".ram_function"))) void board_reset(void) {
    // NVIC_SystemReset doesn't get inlined here.
    SCB->AIRCR  = (uint32_t)((0x5FAUL << SCB_AIRCR_VECTKEY_Pos) |
                             (SCB->AIRCR & SCB_AIRCR_PRIGROUP_Msk) |
                             SCB_AIRCR_SYSRESETREQ_Msk);
    __DSB();
    for (;;) {
        __NOP();
    }
}
```

>  Is that really such an issue? We could try to optimise that down a bit. Eg maybe it doesn't need the app ISR moved there (although I'd suggest doing that regardless, for IRQ performance reasons.)

Probably not, it's not the memory size I'm concerned about, in fact I put all of that code in ITCM which is not used at all, it's just that I use a shared linker script so I'm trying to figure out the bare minimum customization needed for the N6.

kwagyeman:
Damien, can you enable UART3/4 on the N6? PA11/PA12 and PB10/PB11?

dpgeorge:
> can you enable UART3/4 on the N6? PA11/PA12 and PB10/PB11

OK, now done.

kwagyeman:
About the SD card interface. If SD_RESET is pulled high, by default, UHS-1 is available. If it's low by default, then it's not. This isn't something to worry about until trying to implement that...

dpgeorge:
SD card and WiFi are now working!

iabdalkader:
@dpgeorge I'm testing this now, and it's looking good so far. I just wanted to ask you not to force push the branch anymore, or rebase it, because it will be very hard for me to pull this again and test updates. This is just temporarily until this is merged, then you can just squash commits as you like.

ports/stm32/sdcard.c
```diff
@@ -40,7 +40,7 @@
 
 #if MICROPY_HW_ENABLE_SDCARD || MICROPY_HW_ENABLE_MMCARD
 
-#if defined(STM32F7) || defined(STM32H5) || defined(STM32H7) || defined(STM32L4)
+#if defined(STM32F7) || defined(STM32H5) || defined(STM32H7) || defined(STM32L4) || defined(STM32N6)
```
iabdalkader, line 43: `sdcard_reset_periph` access SDMMC regs before the clock is enabled. This doesn't work on N6, I'm surprised it worked for you. Maybe the clock is enabled somewhere else? Anyway, I get a hardfault as soon as `SDMMC_CLK_ENABLE` is called. The fix is to make sure `SDMMC_CLK_ENABLE` is called before any regs are accessed. Also, we need a `__HAL_RCC_SDMMCx_CLK_SLEEP_ENABLE` otherwise clocks will stop during `WFI`.

Note I can work around both issues in our fw, but this should probably be fixed here too.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
```diff
@@ -0,0 +1,168 @@
...
+#define MICROPY_HW_HAS_FLASH        (1)
+#define MICROPY_HW_FLASH_MOUNT_AT_BOOT (1)
+#define MICROPY_HW_SDCARD_MOUNT_AT_BOOT (0)
+#define MICROPY_HW_ENABLE_RNG       (1)
+#define MICROPY_HW_ENABLE_RTC       (1)
+#define MICROPY_HW_ENABLE_ADC       (0)
+#define MICROPY_HW_ENABLE_DAC       (0)
+#define MICROPY_HW_ENABLE_USB       (1)
+#define MICROPY_HW_ENABLE_SDCARD    (0) // TODO enable
+#define MICROPY_PY_PYB_LEGACY       (0)
```
iabdalkader, line 17:
```c
#define MICROPY_FATFS_EXFAT         (1)
```
We use exfat for all boards.

iabdalkader:
Re waking up from deepsleep, I got the following working (reusing the same trick I used for NV, by rigging a memory ISR):

```C
static char _boot_mem[1024] __attribute__((aligned(1024)));

__attribute__((naked, noreturn, section(".ram_function"))) void ram_reset(void) {
    // NVIC_SystemReset doesn't get inlined here.
    SCB->AIRCR  = (uint32_t)((0x5FAUL << SCB_AIRCR_VECTKEY_Pos) |
                             (SCB->AIRCR & SCB_AIRCR_PRIGROUP_Msk) |
                             SCB_AIRCR_SYSRESETREQ_Msk);
    __DSB();
    for (;;) {
        __NOP();
    }
}

void board_early_init(void) {
    HAL_PWREx_EnableTCMRetention();
    HAL_PWREx_EnableTCMFLXRetention();

    uint32_t *boot_mem = (uint32_t *)_boot_mem;
    boot_mem[0] = (uint32_t)(_boot_mem + 1024);
    boot_mem[1] = ((uint32_t)&ram_reset) | 1;
    SCB_CleanDCache_by_Addr((uint32_t *)_boot_mem, 32);

    SYSCFG->INITSVTORCR = (uint32_t) boot_mem;
    (void) SYSCFG->INITSVTORCR;
}
```

It causes an NVIC reset on wakeup, and the bootloader runs again, with flash out of 888. Though, I'm not 100% happy with it, there should be an easier way to get the old STM behavior (NVIC reset on wakeup). I'll ask around, but if this is the only way, it's still considerably less complex than the asm code, `iram_bootloader_reset`, etc... 

Note that this doesn't have to be used in MicroPython, it's self contained in board.c, just need an option to not change `INITSVTORCR`.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
```diff
@@ -0,0 +1,176 @@
...
+#define MICROPY_HW_SDIO_D0          (pyb_pin_WL_SDIO_D0)
+#define MICROPY_HW_SDIO_D1          (pyb_pin_WL_SDIO_D1)
+#define MICROPY_HW_SDIO_D2          (pyb_pin_WL_SDIO_D2)
+#define MICROPY_HW_SDIO_D3          (pyb_pin_WL_SDIO_D3)
+
+// USB config
+#define MICROPY_HW_USB_HS           (1)
+#define MICROPY_HW_USB_HS_IN_FS     (1)
+#define MICROPY_HW_USB_MAIN_DEV     (USB_PHY_HS_ID)
+#define MICROPY_HW_USB_MSC          (0)
```
iabdalkader:
```suggestion
#define MICROPY_HW_USB_MSC          (1)
```

We use MSC by default. BTW the SD card works fine with MSC, but internal flash doesn't. Is it disabled somewhere?

ports/stm32/sdcard.c
iabdalkader: Another clock issue with BT UART, it doesn't work without: `__HAL_RCC_USART2_CLK_SLEEP_ENABLE`. Any clock used must have it's SLEEP_ENABLE called, otherwise it won't work during WFI.

EDIT: I just noticed the `LL_xxx_xxx_EnableClockLowPower(0xffffffff);` calls in `main.c`, which I didn't copy because I enable individual clk_sleep when needed. I also saw SDMMC low power clocks enabled again in `sdio.c`.

dpgeorge: Yes, I just enabled low power clock for everything, because I was having so many issues with WFI!

Do you think it's better to enable them individually when needed?

Well, at least for things like RAM they need to be enabled in main.

iabdalkader: > Do you think it's better to enable them individually when needed?

I'm not sure what the best approach is, but that's what I do in our firmware (enable them on demand). I've added the corresponding `CLK_SLEEP_ENABLE` to every `CLK_ENABLE`.

Re RAMs, I agree they need to be enabled early on. I enable some RAMs in the reset handler (I think those are used by the bootloader), and others in `SystemInit`.

iabdalkader:
Just noticing a weird issue with WiFi, it seems to work fine with debug builds, but with non-debug builds only scan works.. Does this sound familiar? Perhaps I'm missing something or the memory layout changes and breaks something.

EDIT: Seems to be a lwip issue that has nothing to do with CYW43. If I change `MEM_ALIGNMENT` to 32 it works fine. I wanted to override those opts anyway to increase the memory, but I don't see a way to do that in `stm32` port. Is there way to provide a custom config for the N6? @dpgeorge  or would something like this be okay?
```C

#ifdef STM32N6
#undef MEM_SIZE
#undef MEM_ALIGNMENT
# etc..

#define MEM_SIZE                        (16 * 1024)
#define MEM_ALIGNMENT                   32
# etc..
#endif
```

ports/stm32/sdcard.c
iabdalkader: I've thought about this more and one reason Not to enable every clock in low-power mode, is that it will affect current draw in WFI. It's probably best to enable as needed. Note that I've also described a different issue with `sdcard_reset_periph`, though I managed to get around this one by enabling SDMMCx in early init code.

dpgeorge: I've enabled some sleep clocks unconditionally in main (SPI, UART, I2C, memory), but left others to enable them explicitly when needed.

I think if you enable a peripheral, you probably want it to stay awake during WFI/WFE.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
dpgeorge: Now enabled on OPENMV_N6.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
dpgeorge: MSC now enabled on OPENMV_N6, and I fixed it so it works with internal flash (needed to increase MSC buffer size from 2k to 4k because that's the flash erase size).

dpgeorge:
For deepsleep: I've now made it so the INITSVTORCR code is in MICROPY_BOARD_ENTER_STANDBY.  So you can override it.

For lwIP: you're right that it doesn't work properly.  I set MEM_ALIGNMENT to 8, that seems to be enough to at least allow it to connect to an AP.  I also increased the memory settings on N6.  But still it doesn't fully work for me, eg I can't ping the board (no response, although the board does see the incoming packets), and a `getaddrinfo` doesn't work either.  I don't know what the problem is.

(I pushed changes without doing a rebase.)

kwagyeman:
It needs 32 on the N6. Ibrahim got it working with that.

dpgeorge:
> It needs 32 on the N6

I did try with all powers of 2 between 4 and 64.  With 4 it couldn't connect to the AP, but with 8 it could connect.  Higher than 8 showed no change in behaviour (ie ping still didn't work).

iabdalkader:
I didn't test ping, just a simple scan + connect to AP. I'm also not sure why the alignment fixes the connect, but at least you can reproduce this issue too.

Thanks for not force pushing the updates!

dpgeorge:
> I didn't test ping, just a simple scan + connect to AP.

For me, scan always works (regardless of mem align).  Connect to AP only works with mem align 8 or more.  Ping never works.

We probably need to understand why mem alignment matters, because mem align of 4 should be enough.  (Actually I'd be surprised if it even needs to be 4, it should work at 1 because ARM allows unaligned accesses.)

iabdalkader:
> For me, scan always works (regardless of mem align). Connect to AP only works with mem align 8 or more. Ping never works.

Yes, the exact same thing here, except I didn't test ping.

> We probably need to understand why mem alignment matters, because mem align of 4 should be enough

I'm wondering if alignment always mattered and we were just lucky with other boards, or if it's just the N6. It could have something to do with cache, might be worth testing again with 4 + dcache disabled.

iabdalkader:
@dpgeorge Forgot to mention that compiling lwip objects with `-O2` (custom rule in `extmod.mk`) also makes AP connect work for me. This may indicate that it has something to do with memory layout.

EDIT: If SDIO writes directly into lwip buffers, it could have some strict alignment requirement. I have a way to easily test this by using a temporary aligned buffer.

iabdalkader:
Any updates on the LWIP issue? I think this is pretty much ready to merge (at least in our fork), but we probably should wait for the LWIP fix.

dpgeorge:
> Any updates on the LWIP issue?

I looked at the incoming data from the cyw43 when a ping packet comes in, and that is passed correctly up into lwIP.  Then there's no further action from lwIP.  So, the issue lies either in lwIP processing the ping, or lwIP trying to send out a response.

> I think this is pretty much ready to merge

Yes, it's quite close to being mergeable here (although there won't be support for all peripherals, eg ETH; they will need to be done at a latter stage).

iabdalkader:
> Yes, it's quite close to being mergeable here (although there won't be support for all peripherals, eg ETH; they will need to be done at a latter stage).

That's fine, I'm just waiting on the LWIP issue.

iabdalkader:
@dpgeorge If you build with `DEBUG=1` ping now works, perhaps this could help narrow down the issue.

EDIT more specifically, it's `inet_chksum.c` file for some reason. If you building with `-Og`, ping works.

```bash
$(BUILD)/$(LWIP_DIR)/core/inet_chksum.o: CFLAGS += -Og
```

And this might be related: https://lists.gnu.org/archive/html/lwip-users/2011-11/msg00114.html

Summary: the root cause is probably pbuf list corruption, likely due to some race condition.

dpgeorge:
Building without -Og, but with alignment of 8, incoming ping packets fail their checksum check.

I can't load that lwip issue you link to (it times out) but if the date of that is really 2011 then that's a long time ago and it should be fixed by now?  Also strange that the issue is not seen on other boards using lwIP.

iabdalkader:
> I can't load that lwip issue you link to (it times out) but if the date of that is really 2011 then that's a long time ago and it should be fixed by now? Also strange that the issue is not seen on other boards using lwIP.

That's weird, it was working just yesterday. Anyway, building the checksum function with -O0 or -Og just introduced an extra delay which masked the issue. The OP eventually found out that it was his DMA code corrupting frames. We're likely looking at the same issue, and if we assume it doesn't exist on other ports or with other MCUs then it has to be the new SDIO code.

EDIT: I can't entirely rule out the bug existing in other ports or with other MCUs. Note that the N6 is the fastest MCU so far, so the speed at which it's processing those packets/frames might be what's triggering the race condition. So the issue could be higher up the stack.

dpgeorge:
After a lot of debugging and very strange observations, my tentative conclusion is that the Cortex-M55 optimiser in gcc is broken... !!

Switching the build flags for the N6 to `-mtune=cortex-m33 -mcpu=cortex-m33`, everything starts working, even with the default alignment of 4 bytes for lwIP.

I was investigating the `lwip_standard_chksum()` function in lwIP because it was returning strangely incorrect values.  The memory that it was calculating the chksum for was fine, it had the right IP header, so somehow this function is doing the wrong thing.  Looking at its disassembly when compiled with `cortex-m55` tuning, the function is more complicated and longer than when compiled with `cortex-m33` tuning.  With `cortex-m55` gcc is using new ARMv8 opcodes like `csinc` and that seems to be confusing gcc's optimiser, because it's producing some very strange looking code.  I didn't yet fully work out the logic of this assembly, it would be worth doing that to prove that indeed the generated assembly is wrong (and then file a gcc bug report).

This would explain why things work when that function is compiled with `-Og` as @iabdalkader found out.

It's a bit scary if gcc is broken for `cortex-m55` because it means a lot of other code is wrong.  Eg the alif port uses `cortex-m55` tuning.

iabdalkader:
> I was investigating the `lwip_standard_chksum()` function in lwIP because it was returning strangely incorrect values

Should be easy to test the function with known input/output. I'll give it a shot. FWIW I quickly tested `gcc version 14.2.1 20241119` and it doesn't fix the issue, nor does changing the checksum algorithm (tried 1 to 3). Not sure how to proceed, can we just add an -Og for that checksum file for N6? Everything else seems to be working perfectly fine. I'm leaning towards not gcc bug because we have a ton of optimize, image processing code that works fine. Note that we always build our firmware with `-O2`.

iabdalkader:
@dpgeorge This produces different results with/without -Og, you were right something is definitely broken...
```diff
diff --git a/ports/stm32/modpyb.c b/ports/stm32/modpyb.c
index 3d8696e3c..d389bb5cd 100644
--- a/ports/stm32/modpyb.c
+++ b/ports/stm32/modpyb.c
@@ -133,6 +133,50 @@ static mp_obj_t pyb_country(size_t n_args, const mp_obj_t *args) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mod_network_country_obj, 0, 1, pyb_country);
 #endif
+#include "lwip/inet_chksum.h"
+#include "lwip/def.h"
+
+static u16_t __attribute__((optimize("Og")))  lwip_standard_chksum(const void *dataptr, int len) {
+  u32_t acc;
+  u16_t src;
+  const u8_t *octetptr;
+
+  acc = 0;
+  /* dataptr may be at odd or even addresses */
+  octetptr = (const u8_t *)dataptr;
+  while (len > 1) {
+    /* declare first octet as most significant
+       thus assume network order, ignoring host order */
+    src = (*octetptr) << 8;
+    octetptr++;
+    /* declare second octet as least significant */
+    src |= (*octetptr);
+    octetptr++;
+    acc += src;
+    len -= 2;
+  }
+  if (len > 0) {
+    /* accumulate remaining octet */
+    src = (*octetptr) << 8;
+    acc += src;
+  }
+  /* add deferred carry bits */
+  acc = (acc >> 16) + (acc & 0x0000ffffUL);
+  if ((acc & 0xffff0000UL) != 0) {
+    acc = (acc >> 16) + (acc & 0x0000ffffUL);
+  }
+  /* This maybe a little confusing: reorder sum using lwip_htons()
+     instead of lwip_ntohs() since it has a little less call overhead.
+     The caller must invert bits for Internet sum ! */
+  return lwip_htons((u16_t)acc);
+}
+
+static mp_obj_t pyb_checksum(mp_obj_t data) {
+    mp_buffer_info_t rbuf;
+    mp_get_buffer_raise(data, &rbuf, MP_BUFFER_READ);
+    return mp_obj_new_int(lwip_standard_chksum(rbuf.buf, rbuf.len));
+}
+static MP_DEFINE_CONST_FUN_OBJ_1(pyb_checksum_obj, pyb_checksum);
 
 static const mp_rom_map_elem_t pyb_module_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR___name__), MP_ROM_QSTR(MP_QSTR_pyb) },
@@ -260,6 +304,7 @@ static const mp_rom_map_elem_t pyb_module_globals_table[] = {
     #if MICROPY_HW_HAS_LCD
     { MP_ROM_QSTR(MP_QSTR_LCD), MP_ROM_PTR(&pyb_lcd_type) },
     #endif
+    { MP_ROM_QSTR(MP_QSTR_checksum), MP_ROM_PTR(&pyb_checksum_obj) },
 };
 
 static MP_DEFINE_CONST_DICT(pyb_module_globals, pyb_module_globals_table);
```

iabdalkader:
I tested Alif as well, they all produce the wrong output with `-Os`, even when using tune=m33 cpu=m33, so it might one of the options enabled/disabled with `-Os` that breaks things. I think the safer path here is to just avoid `-Os` with these MCUs for now, they don't really need to save code space anyway.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
```diff
@@ -0,0 +1,175 @@
...
+#define MICROPY_HW_ENABLE_DAC       (0)
+#define MICROPY_HW_ENABLE_USB       (1)
+#define MICROPY_HW_ENABLE_SDCARD    (1)
+#define MICROPY_PY_PYB_LEGACY       (0)
+
+#define MICROPY_HW_ENABLE_INTERNAL_FLASH_STORAGE (0)
+
+#define MICROPY_BOARD_ENTER_BOOTLOADER board_enter_bootloader
+#define MICROPY_BOARD_EARLY_INIT    board_early_init
+#define MICROPY_BOARD_LEAVE_STANDBY board_leave_standby
```
iabdalkader: I think this should have parentheses to match MICROPY_BOARD_ENTER_STANDBY. There's some inconsistency with these defines.
.
```suggestion
#define MICROPY_BOARD_LEAVE_STANDBY board_leave_standby();
```

dpgeorge:
I've done some more testing, on both STM32N6 and Alif AE3.  I made the following changes:
- made `lwip_standard_chksum()` copy the incoming data into a temporary (and aligned to 16 bytes) buffer, and then compute the checksum of that buffer
- add `machine.checksum(buf)` Python function to be able to call into the lwIP `lwip_standard_chksum()` function

I then run the following test:
```py
import os, machine

# this is 20 bytes of data that has an IP checksum of 0xffff
data = bytes.fromhex('45 00 00 54 09 3e 40 00 40 01 af 85 c0 a8 00 0d c0 a8 00 88')

machine.checksum(os.urandom(100))
machine.checksum(bytes(20))
print(hex(machine.checksum(data)))

machine.checksum(os.urandom(100))
machine.checksum(bytes(21))
print(hex(machine.checksum(data)))
```

Both prints should print the same thing, 0xffff.  But when it's broken the first print gives random results, and the second print gives a consistent 0xffff (which is the correct result).

It's broken on both STM32N6 and Alif AE3 when building with `-mcpu=cortex-m55 -mtune=cortex-m55` and `-Os`.  It works on both those MCUs using `-mtune=cortex-m33 -mcpu=cortex-m33` and `-Os`.

The difference between Cortex-M33 and Cortex-M55 is ARMv8-M and ARMv8.1-M.  The latter has a few new core instructions like `uxtah` (in addition to vector and half-precision float support).

What the above test shows is that having a random byte in the buffer after the data ends changes the result of the checksum.  And from inspection of the machine code of `lwip_standard_chksum()` it's because gcc is generating incorrect code that allows a buffer overrun, and that overrun means that it always computes a checksum of an odd number of bytes, even if the input length argument is even.

Here's the disassembly of broken machine code (for `LWIP_CHKSUM_ALGORITHM == 1`):
```
700c7dec <lwip_standard_chksum>:           
700c7dec:       2300            movs    r3, #0                             
700c7dee:       b510            push    {r4, lr}                   
700c7df0:       084c            lsrs    r4, r1, #1                         
700c7df2:       4299            cmp     r1, r3                             
700c7df4:       f104 0401       add.w   r4, r4, #1                    
700c7df8:       bfac            ite     ge
700c7dfa:       4622            movge   r2, r4                              
700c7dfc:       2201            movlt   r2, #1                             
700c7dfe:       4408            add     r0, r1                        
700c7e00:       3a01            subs    r2, #1                               
700c7e02:       eba0 0401       sub.w   r4, r0, r1                               
700c7e06:       d10d            bne.n   700c7e24                <-- strange
700c7e08:       bf04            itt     eq                      <-- strange
700c7e0a:       7822            ldrbeq  r2, [r4, #0]                           
700c7e0c:       eb03 2302       addeq.w r3, r3, r2, lsl #8
700c7e10:       b298            uxth    r0, r3
700c7e12:       eb00 4013       add.w   r0, r0, r3, lsr #16
700c7e16:       b283            uxth    r3, r0                   
700c7e18:       f5b0 3f80       cmp.w   r0, #65536      @ 0x10000
700c7e1c:       ea50 9033       csinc   r0, r0, r3, cc           
700c7e20:       b280            uxth    r0, r0                                
700c7e22:       bd10            pop     {r4, pc}                           
700c7e24:       8824            ldrh    r4, [r4, #0]  
700c7e26:       3902            subs    r1, #2
700c7e28:       ba64            rev16   r4, r4
700c7e2a:       fa13 f384       uxtah   r3, r3, r4
700c7e2e:       e7e7            b.n     700c7e00
```
The broken code can be seen at address `700c7e06`: there's a `bne.n` followed by a `itt eq`.  The problem here is that if the `bne.n` branch is not taken (because the compare is equal) then the following `itt eq` will always be taken!  And hence the `ldrbeq` will always be executed and read one more byte after the end of the input buffer (assuming the input length is even).

Note that exactly the same bug manifests for all 3 versions of `lwip_standard_chksum()`, which is quite interesting.  And in some cases the bug is there even though there aren't any new ARMv8.1-M instructions being used.

----------

Summary: gcc with `-mcpu=cortex-m55 -mtune=cortex-m55` and `-Os` produces incorrect machine code, at least in the example above.  Considering that `-Os` has been extensively used so far with other Cortex-M's and has given no issues, and that Cortex-M55 introduces some new instructions, I'd say the issue lies with gcc's optimiser and handling of the new ARMv8.1-M instructions.

Thus I don't think we can trust gcc's `-mcpu=cortex-m55 -mtune=cortex-m55` option and we should not use it at all until we know more.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
dpgeorge: I was trying to be consistent with `MICROPY_BOARD_STARTUP` and `MICROPY_BOARD_EARLY_INIT` etc.

The problem with having the `;` part of the macro is that the use of the macro doesn't end in a `;` and hence isn't really well formatted C code, and can give problems with the auto-formatter (uncrustify).

dpgeorge:
This gcc bug looks related: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116444 .  That's just a missed optimisation bug, but could have deeper implications.

iabdalkader:
I'm not sure what to do here, we could use `-O2`, it seems to work and we'll not lose the new m55 instructions, or m33 but it will be a very very long time before we see a fix. Anyway, you should definitely report this somewhere to gcc or Arm, I'm surprised that no one else reported it and it's been a while since the m55 was released.

dpgeorge:
> we could use `-O2`, it seems to work

I wouldn't be confident using `-O2` with `-mcpu=cortex-m55 -mtune=cortex-m55`.  The bug seems to be something to do with optimising the counter used in a loop, and it could still be there with `-O2`.

I will file a report to gcc then we can see what to do.

kwagyeman:
I can verify this issue is present on 13.2.rel1 and 13.3.rel1. O2 seems to fix the issue.

Damien, we use M55 Helium instructions heavily in our upstream software. Is there a way we could switch things to O2 globally? We have the firmware space on the AE3 and N6.

kwagyeman:
From ARM:

Hi all,

Thank you for reporting this, this had already been reported on a different target in PR 116799.

The good news is the issue has already been fixed on trunk and backported to all active branches.  This fix is in our latest Release Candidate for the 14.3.Rel1 release that should be coming out before the end of the month.

If you would like a way to work around this wrong-code gen for the existing releases you can disable the pass that introduces this issue by passing the option '-fdisable-rtl-loop2_doloop'.

Relevant links:
  Bugzilla: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116799
  Patches on:
    trunk: https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=f87fe2579b48d57c5f97bb91674b60808722855d

    gcc-14: https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=26615af93cf759999c5dfae0c51a827b05968cca

    gcc-13: https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=0eed81612ad6eac2bec60286348a103d4dc02a5a

    gcc-12: https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=c1e55bcc1075aa74bebd7e1bb87d3939da2e498b

ports/stm32/boards/OPENMV_N6/board.ld
```diff
@@ -0,0 +1,12 @@
+/*
+    Linker script for OPENMV_N6.
+*/
+
+MEMORY
+{
+    EXT_FLASH_ROMFS (rx) : ORIGIN = 0x70200000, LENGTH = 14M
```
kwagyeman: 24M?

ports/stm32/lwip_inc/lwipopts.h
```diff
@@ -10,9 +10,24 @@
 
 #define LWIP_RAND() rng_get()
 
+// Increase memory for lwIP to get better performance.
+#if defined(STM32N6)
+#define MEM_SIZE                        (16 * 1024)
```
kwagyeman, line 15: Make larger? We have 1GB/s Ethernet onboard.

dpgeorge:
I can confirm that using `-fdisable-rtl-loop2_doloop` fixes the issue.  So I suggest using that flag on compilers known to have the bug.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
```diff
@@ -0,0 +1,175 @@
...
+#define MICROPY_HW_ROMFS_XSPI_SPIBDEV_OBJ (&spi_bdev)
+#define MICROPY_HW_ROMFS_ENABLE_PART0 (1)
+
+#if 1
+// SPI flash, block device config.
+#define MICROPY_HW_BDEV_SPIFLASH                (&spi_bdev)
+#define MICROPY_HW_BDEV_SPIFLASH_EXTENDED       (&spi_bdev)
+#define MICROPY_HW_BDEV_SPIFLASH_CONFIG         (&spiflash_config)
+#define MICROPY_HW_BDEV_SPIFLASH_OFFSET_BYTES   (16 * 1024 * 1024)
+#define MICROPY_HW_BDEV_SPIFLASH_SIZE_BYTES     (16 * 1024 * 1024)
```
iabdalkader: @dpgeorge If I understand correctly, this is the offset at which the flash filesystem (writable fs) starts and its size, correct? If so, this needs to be 4MBs and 4MBs, and the romfs offset in the linker script needs to be at 8MBs and size 24MBs. This is the current flash layout we use:
4MBs -> firmware
4MBs-> RW FS
24MBs-> ROMFS

I can change that, if there's a good reason to put the rw FS last?

ports/stm32/lwip_inc/lwipopts.h
iabdalkader: I prefer if we use the config I made for mimxrt and alif, back then I checked it against the docs and every option is carefully calculated and it performs very well:
```
#define MEM_SIZE                        (16 * 1024)
#define TCP_MSS                         (1460)
#define TCP_OVERSIZE                    (TCP_MSS)
#define TCP_WND                         (8 * TCP_MSS)
#define TCP_SND_BUF                     (8 * TCP_MSS)
#define TCP_SND_QUEUELEN                (2 * (TCP_SND_BUF / TCP_MSS))
#define TCP_QUEUE_OOSEQ                 (1)
#define MEMP_NUM_TCP_SEG                (2 * TCP_SND_QUEUELEN)
```
Could increase the MEM_SIZE if you like.

iabdalkader:
> `-fdisable-rtl-loop2_doloop`

This option generates one warning/note per file, that seems impossible to suppress.

ports/stm32/lwip_inc/lwipopts.h
kwagyeman: Probably wait till performance testing things with ipref3 when ethernet is working.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
kwagyeman: Seconding Ibrahim here. We've shipped alpha N6 boards to a few customers. Would prefer Micropython matches OpenMV's flash layout so they don't have to update their bootloaders.

dpgeorge:
> This option generates one warning/note per file, that seems impossible to suppress.

Yes I saw that.  I don't think there's much choice, except to update to a non-broken compiler.

I will make sure `-fdisable-rtl-loop2_doloop` is only enabled for gcc versions that are known broken.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
dpgeorge: > I can change that, if there's a good reason to put the rw FS last?

It doesn't matter the order.  I will update this to match your settings.

ports/stm32/lwip_inc/lwipopts.h
dpgeorge: > I prefer if we use the config I made for mimxrt and alif,

When I updated this I checked that all options I left out here take their defaults and the defaults match that your configuration.  So these settings do match what you have.

iabdalkader:
> I will make sure `-fdisable-rtl-loop2_doloop` is only enabled for gcc versions that are known broken.

That would be anything less than the future 14.3 release. I enable it unconditionally for CM55 for now, but the following also works:

```Makefile
ifeq ($(CPU),cortex-m55)
# Check if GCC version is less than 14.3
GCC_VERSION := $(shell arm-none-eabi-gcc -dumpversion | cut -d. -f1-2)
GCC_MAJOR := $(shell echo $(GCC_VERSION) | cut -d. -f1)
GCC_MINOR := $(shell echo $(GCC_VERSION) | cut -d. -f2)

# Convert to comparable number (14.3 becomes 1403)
GCC_VERSION_NUM := $(shell echo $$(($(GCC_MAJOR) * 100 + $(GCC_MINOR))))

# Only add the flag if version < 14.3 (1403)
ifeq ($(shell test $(GCC_VERSION_NUM) -lt 1403 && echo yes),yes)
$(warning )
$(warning *** WARNING ***)
$(warning GCC $(GCC_VERSION) has known issues with Cortex-M55)
$(warning Recommend upgrading to GCC 14.3+ for proper CM55 support)
$(warning )
CFLAGS += -fdisable-rtl-loop2_doloop
endif
endif
```

Anyway, I've already merged this PR in our fork (with the fixed flash layout). Everything seems to be working fine so far. I don't have any more comments, I think this is ready to be merged.

ports/stm32/boards/OPENMV_N6/mpconfigboard.h
dpgeorge: Now updated.

ports/stm32/boards/OPENMV_N6/board.ld
dpgeorge: Now 24M.

dpgeorge:
I've pushed a few new commits here:
- improve ADC N6 code (although it's still not fully working... not sure why)
- enable ADC on OPENMV_N6
- remove lwip alignment to 8 bytes
- disable gcc optimisation on Cortex-M55 (we can reenable this once there's a release of gcc that fixes the issue)
- adjust OPENMV_N6 flash partitioning to match OpenMV settings

I just need to retest it with the NUCLEO and DK boards, then it should be good to merge as preliminary N6 support.  Any further improvements can be made in the future.

dpgeorge:
I have improved a few things here (eg mboot now uses XSPI in octal mode for writing, which speeds up DFU deployment), fixed and tested NUCLEO_N657X0 and removed support for the STM32N6570_DK board (I don't have time to test, get this board working and document it, that can be done later if needed/desired).

I've taken this PR out of draft/WIP, it should be pretty much ready to merge.

iabdalkader:
Note GCC 14.3.1 is out and it does seem to fix the issue. It adds a bit of text size to the firmware, not sure why, but that's okay for us at least. I suggest making Makefile fail if an older toolchain is used (For example, https://github.com/openmv/openmv/blob/master/common/check_toolchain.mk)

dpgeorge:
> Note GCC 14.3.1 is out and it does seem to fix the issue.

OK, good.  I've tested it as well and it does fix the problem.



> I suggest making Makefile fail if an older toolchain is used

OK, I've made an addition for this, similar to what you suggested.

dpgeorge:
Merged!

kwagyeman:
@dpgeorge Amazing! Looking forward to Ethernet and ADC PRs now!

anchung-chen:
1: For the ADC to work on STM32N6, the RIF(Resource Isolation Framework) attributes of ADC  should be set in main.c
    HAL_RIF_RISC_SetSlaveSecureAttributes(RIF_RISC_PERIPH_INDEX_ADC12 , RIF_ATTRIBUTE_SEC | RIF_ATTRIBUTE_PRIV);

2: The current clock frequency of ADC is 100MHz (in powerctrlboot.c), but the sample time of ADC is only 11.5 clock (in adc.c) which is far too small.

3: Also the calling of ADC calibration function is missing for STM32N6 in adc.c

dpgeorge:
@anchung-chen thanks for the hints!  I've now fixed N6 ADC in PR #18009.

PR #17187, opened 2025-04-25
rp2/boards/WEACTSTUDIO_RP2350B_CORE: Add board.pinout

mattytrentini:
### Summary

This adds an ANSI-rendered pinout for the WeAct Studio RP2350B Core board:

```python
>>> import board
>>> board.pinout()
```
image

Compare with:

image

From [WeActStudio/WeActStudio.RP2350BCoreBoard](https://github.com/WeActStudio/WeActStudio.RP2350BCoreBoard/blob/main/HDK/RP2350B_PINOUT.png).

I've created a few of these pinouts for some boards I use, but I'd be interested if other folks think that this it's a useful (optional) feature for a board definition to supply.

It would be neat to make this an optional pseudo-standard; ie always at `board.pinout()` - or whatever we choose. 

### Testing

Tested using mpremote on Windows to a WeAct RP2350B board. 
Also tested `board.py` on the unix port.

### Background

I used some (currently pretty hacky!) Python scripts to help generate the output, I can tidy them up and submit them if there is interest.


mattytrentini:
Compression has been added, which reduces the size of the diagram by 6x. I've also tidied up and added the scripts to generate the diagram - as well as another to compress the output. A README explains how to use it all. 

Thanks to @dpgeorge for help with compressing/decompressing!

tools/make_pinout_diagram/README.md
```diff
@@ -0,0 +1,34 @@
+# Pinout diagram
```
dpgeorge, line 1: Suggest moving to `tools/generate_pinout_diagram/`, to make it easier to find and use for other boards in the future.

tools/make_pinout_diagram/pinout.py
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    for symbol, colours in symbol_colours.items():
+        pinout_str = pinout_str.replace(symbol, ansi_colour(symbol, colours))
+    return pinout_str
+
+
+def colour_tags(matchobj: re.Match) -> str:
+    tag_colours = {
+        "5V": (15, 1),
```
dpgeorge: Suggest making these tuples constants, eg:
```py
COL_WHITE_ON_RED = (15, 1)
...
```

tools/make_pinout_diagram/README.md
dpgeorge: Actually, please name it `tools/make_pinout_diagram/`, to match all the other make-prefixed tools (eg `make-pins.py`, `make_root_pointers.py` etc).

mattytrentini:
Updated to address all the feedback. Thanks!

I'd like to add a `make_pinout_diagram/boards/WEACTSTUDIO_RP2350B_CORE` subfolder in which to store `pinout.py` but I think we can update this later when there are multiple pinout files.

ports/rp2/boards/WEACTSTUDIO_RP2350B_CORE/modules/board.py
```diff
@@ -0,0 +1,10 @@
+import micropython
+import deflate
+import io
+import binascii
```
dpgeorge: I think `micropython` and `binascii` can be removed.

mattytrentini: Done

dpgeorge (approved):

PR #17201, opened 2025-04-28
rp2/rp2_dma: Fix default value used in pack_ctrl on RP2350.

dpgeorge:
### Summary 

The bit position of CHAIN_TO is not the same as on RP2040.

### Testing

This was picked up by the `tests/ports/rp2/rp2_dma.py` test running on a RPI_PICO2 board.

(That test still doesn't pass, because of other reasons, to be fixed separately.)


dpgeorge:
See improved test in #17202.  This PR is needed for that test to pass fully.

PR #17202, opened 2025-04-28
tests/ports/rp2: Convert rp2.DMA test to a unittest and make it more reliable

dpgeorge:
### Summary

This PR converts the rp2.DMA test to a unittest (in the first commit, should be a no-op in terms of functionality of the test) and makes it more reliable across all platforms (in the second commit).

Changes to the test to make it more reliable are:
- Allow the DMA instance to be any instance, not just DMA(0); eg WLAN may be using DMA(0).
- Make the DMA timing test run a little faster by preloading `dma.active`.
- Run the DMA timing test 10 times and take the average time taken as the test result, to eliminate any big effects of caching.
- Change the expected time to `range(30, 80)` to cover RP2040, RP2350, RISC-V variants, and both bytecode and native emitter.
- Add a `sleep_ms(1)` after waiting for the IRQ to fire, so that any scheduled code gets a chance to run when the test is compiled with the native emitter.

### Testing

With these changes this test passes reliably on RPI_PICO, RPI_PICO_W, RPI_PICO2, RPI_PICO2_W, RPI_PICO2-RISCV and RPI_PICO2_W-RISCV, in both bytecode and native emitter mode, with and without WLAN enabled.

The changes here picked up the bug fixed by #17201.


projectgus (approved):
This is a really good test refactor!

PR #17205, opened 2025-04-29
tools/mpremote: For mip install, use hash to skip files that exist.

dpgeorge:
### Summary

When using `mip install`, if a file that needs to be downloaded already exists locally, then the hash of that local file will be computed and if it matches the known hash of the remote file it will not be downloaded.

This behaviour follows that of the `mip` package in `micropython-lib`.

### Testing

Tested on RPI_PICO2_W.  Prior to this change, with `unittest` already installed:
```
$ mpremote mip install unittest
Install unittest
Installing unittest (latest) from https://micropython.org/pi/v2 to /lib
Installing: /lib/unittest/__init__.mpy
Done                                    
```
With this change:
```
$ ./mpremote.py mip install unittest
Install unittest
Installing unittest (latest) from https://micropython.org/pi/v2 to /lib
Exists: /lib/unittest/__init__.mpy
Done
```

Hashes in `mip` are guaranteed unique, so this change should never leave stale files on the filesystem.

### DIFF
--- a/tools/mpremote/mpremote/mip.py
+++ b/tools/mpremote/mpremote/mip.py
@@ -34,6 +34,15 @@ def _ensure_path_exists(transport, path):
         prefix += "/"
 
 
+# Check if the specified path exists and matches the hash.
+def _check_exists(transport, path, short_hash):
+    try:
+        remote_hash = transport.fs_hashfile(path, "sha256")
+    except FileNotFoundError:
+        return False
+    return remote_hash.hex()[: len(short_hash)] == short_hash
+
+
 def _rewrite_url(url, branch=None):
     if not branch:
         branch = "HEAD"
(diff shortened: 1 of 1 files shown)


projectgus (approved):
This is comparing the short hash, which is 8 hex characters (32 bits), is that right?

I think this is OK, we can expect a 50% chance of a collision in a single source file after 65,536 iterations (cribbed from [here](https://www.bluebill.net/hash_collisions.html#what-does-collision-have-to-do-with-short-hashes), I didn't check their working but it sounds about right to me). Individual files won't have anywhere close to that many versions uploaded to the package archive, so that's probably OK.

(That's accidental collisions, if an attacker wanted to produce a source file with a short hash collision then this would be relatively easy. However, I can't think what the benefit to an attacker would be - the only use I can think of would be to put the generated source file on the device and ensure it wasn't replaced by the one from the package, but that doesn't seem particularly valuable.)

dpgeorge:
> I think this is OK, we can expect a 50% chance of a collision in a single source file after 65,536 iterations

This hash is a fundamental piece of the mip design, and mip will stop working if there are any collisions.  See the comment here: https://github.com/micropython/micropython-lib/blob/master/tools/build.py#L89-L95

So, if there's ever a collision we need to increase the hash length.

PR #17211, opened 2025-04-30
nrf: fix IRAM address for native code execution

dpgeorge:
### Summary

On nRF52, the physical SRAM is mapped to 0x20000000 for data access and 0x00800000 for instruction access.  So, while native code is allocated and written using addresses in the 0x20000000 range, it must execute from the 0x00800000 range.

This PR makes this work correctly on nRF52 MCUs by adjusting the address.

### Testing

Tested on PCA10040 and ARDUINO_NANO_33_BLE_SENSE using the standard test suite.

### DIFF
--- a/ports/nrf/main.c
+++ b/ports/nrf/main.c
@@ -39,6 +39,7 @@
 #include "py/stackctrl.h"
 #include "py/gc.h"
 #include "py/compile.h"
+#include "py/persistentcode.h"
 #include "extmod/modmachine.h"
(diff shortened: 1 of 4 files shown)


agatti:
Is this applicable to nRF51 MCUs too?

The MICROBIT board under QEMU has native code execution disabled due to a hardfault being raised when code is being executed from RAM in the 0x2000_0000 range.  I wonder if this can fix that issue as well.

PR #17212, opened 2025-04-30
nrf: Fix a few issues with UART REPL

dpgeorge:
### Summary

This PR fixes three things with the nRF UART REPL:
- Only process interrupt chars on UARTs used for REPL.  Otherwise a board without REPL on UART can have its code interrupted if ctrl-C is received on the UART.
- Fix UART write on parts that can't write more than 255 bytes.
- Use 64 byte raw-paste buffer on PCA10028 and PCA10040, due to bug in JLink CDC.

### Testing

Tested on PCA10028, PCA10040 and ARDUINO_NANO_33_BLE_SENSE, using the script from #15909.


robert-hh:
@dpgeorge I will look at that, but today I'm a little bit busy. Tomorrow I'll have plenty of time.

ports/nrf/modules/machine/uart.c
```diff
@@ -446,17 +456,30 @@ static mp_uint_t mp_machine_uart_write(mp_obj_t self_in, const void *buf, mp_uin
...
-    nrfx_err_t err = nrfx_uart_tx(self->p_uart, buf, size);
-    if (err == NRFX_SUCCESS) {
+
+    // Send data out, in chunks if needed.
+    mp_uint_t remaining = size;
+    while (remaining) {
+        #if defined(NRF52832)
+        // The nRF52832 cannot write more than 255 bytes at a time.
+        mp_uint_t chunk = MIN(255, remaining);
+        #else
```
robert-hh, line 466: Would it make sense to have this 255 number configurable?

robert-hh:
Lacking the PCA boards I could not do too much of testing. I tested some aspect with a Arduino Nanon 3 BLE, NRF 52840 and ran that PR on Micro:BIt V1. The latter uses the single UART for REPL.
- I verified that Ctrl-C does not act as keyboard-interrupt unless REPL is assigned ot UART.
- The UART still works within the limits of this port. Receive works if the small 64Byte RX buffer is taken into account and the speed does not exceed 230400 baud. The test code you mentioned is for the STM32 port, so I could not use it.
- The code looks correct.

A few notes not related to this PR:
- At the NRF52840 running code cannot be interrupted by Ctrl-C at the USB interface. Ctrl-C is only recognized once something is sent to the console output.
- Console output is not sent to the PC until at least 128 bytes are accumulated. Then 128 bytes are sent. Console Output via UART is sent immediately.

robert-hh (approved):

dpgeorge:
Thanks for testing and review!

> The test code you mentioned is for the STM32 port, so I could not use it.

I did update #15909 to work on all ports, or at least it should work.  It detects if `pyb` is available or not, and falls back to `sys.stdXXX`.  It also uses `sys.stdout.buffer` or just `sys.stdout` if buffer is not available.   So it should work with the `nrf` port.

> * At the NRF52840 running code cannot be interrupted by Ctrl-C at the USB interface.

Yes, I also noticed that.  I think we can fix that in a separate PR.

> * Console output is not sent to the PC until at least 128 bytes are accumulated

Oh, that's interesting, I didn't see that behaviour.  Is that with ARDUINO_NANO_33_BLE_SENSE?

robert-hh:
> Oh, that's interesting, I didn't see that behaviour. Is that with ARDUINO_NANO_33_BLE_SENSE?

That's with NRF52840 boards, tested ARDUINO_NANO_33_BLE_SENSE and SEEED_XIAO_NRF52.

dpgeorge:
> * Console output is not sent to the PC until at least 128 bytes are accumulated

I investigated this issue.  Indeed it is there.  The following script triggers the bug (run via `mpremote run`):
```py
import sys,time
sys.stdout.write('a'*317+'\r\n')
for _ in range(100):
    time.sleep(.01)
print('done')
```
It doesn't print anything until after the script finishes.  It seems that 320 bytes is the magic limit (the above test writes out 319 bytes before sleeping).

The issue is due to LTO in combination with `-Wl,--wrap=dcd_event_handler`.  Wrapping functions with the linker just doesn't work properly with LTO  enabled.

So that seems like a separate issue to the ones addressed in this PR.  This PR is related to UART REPL.

ports/nrf/modules/machine/uart.c
dpgeorge: I've now made this a macro.

PR #17213, opened 2025-04-30
nrf: Use common implementation of machine disable/enable IRQ.

dpgeorge:
### Summary

This is a breaking change due to the signature change of `enable_irq()`. Previously the signature was:

    machine.enable_irq()

Now the signature matches other ports, and the docs, and is:

    machine.enable_irq(state)

Where `state` is the return value from `machine.disable_irq()`.

### Testing

Tested on ARDUINO_NANO_33_BLE_SENSE.  That board now passes the `tests/extmod/machine_disable_irq.py` test.


andrewleech (commented):
Oh, were these already no-operation when the nrf Bluetooth stack is enabled? That's a surprising trap.

dpgeorge:
> were these already no-operation when the nrf Bluetooth stack is enabled?

Yes.  I don't think there's any way around that?

andrewleech:
> > were these already no-operation when the nrf Bluetooth stack is enabled?
> 
> Yes.  I don't think there's any way around that?

Yeah probably not, although there is a way to debug / step through application code on these chips without interrupting the stuff device called "monitor mode debugging". I bring this up because it presumably uses some form of interrupt masking which might work here to effectively mask out "application" interrupts without interrupting the soft device.
On the other hand, migrating to nimble probably removes these restrictions anyway.

andrewleech:
Regardless, the changes look good, consolidating functionality and signatures to match other ports is a worthy direction to follow certainly.

PR #17214, opened 2025-04-30
nrf/mpconfigport: Enable IO.IOBase as a core feature.

dpgeorge:
### Summary

IOBase is quite an important building block of other parts of the system, such as `mpremote mount` and running .mpy and native tests.  It's also used by `asyncio` which is enabled by default on all boards.

This feature costs +244 bytes of firmware size, which is worth the cost for the extra features it enables.

### Testing

Tested on PCA10028 and PCA10040, they can now run more tests.

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -1469,7 +1469,7 @@ typedef double mp_float_t;
 
 // Whether to provide "io.IOBase" class to support user streams
 #ifndef MICROPY_PY_IO_IOBASE
-#define MICROPY_PY_IO_IOBASE (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_EXTRA_FEATURES)
+#define MICROPY_PY_IO_IOBASE (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
 #endif
 
 // Whether to provide "io.BytesIO" class


Josverl:
it would be also helpful if there would be some clearer documentation regarding this class to provide 
- what methods are available in MicroPython 
- availability across ports 

I have seen : 
> Note that for efficiency, MicroPython doesn’t provide abstract base classes corresponding to the hierarchy above, and it’s not possible to implement, or subclass, a stream class in pure Python.

But I still I think the docs can be improved. ( also I find the terminology hard to understand, and while not a native speaker, I am confident to claim that my grasp of tech English is well above average.) 
One of the reasons is that I have been trying to improve type checking for stdlib.io, and find that the documentation alone does not suffice.

Currently, in the type stubs,  I am assuming that MicroPython's `io.IOBase`  matches CPython's , and that is not correct.
I would like to improve that , and in parallel I'm happy to create a PR for the doc-page.
For that I would like to understand a bit more on the thinking and use of the IOBase class) 

( implementation in the type-stubs has some additional challenges as MyPy has its own opinion on how `io` looks like as well )

andrewleech:
I certainly support it being on by default here; though more broadly should it be enabled by default at more levels in https://github.com/micropython/micropython/blob/79abdad9e97f18f45650e1abce64ee51c3372953/py/mpconfig.h#L1472 at least BASIC features if not CORE ?

dpgeorge:
> I certainly support it being on by default here; though more broadly should it be enabled by default at more levels,
> at least BASIC features if not CORE ?

Yes, maybe it is better to enable it by default at the CORE level.  (CORE is needed rather than BASIC to get it enabled on these low-end nrf boards.)

From what I can tell, enabling it by default at CORE level impacts only the nrf port with nRF51822 and nRF52832 (what this PR is trying to do) and also the stm32 board B_L072Z_LRWAN1.  For the latter, could explicitly disable IOBase if it doesn't fit.  Otherwise would be good to have it there as well.



> it would be also helpful if there would be some clearer documentation regarding this class

A PR to add docs for `io.IOBase` would be welcome!

dpgeorge:
I've updated this PR to enable `io.IOBase` in `py/mpconfig.h` at the CORE feature level.

The B_L072Z_LRWAN1 board still builds and increases its firmware by +248.

andrewleech:
First pass at some quick docs.... https://github.com/micropython/micropython/pull/17263

Issue #16692, opened 2025-02-04
Build fails with GCC 15 due to  -Wunterminated-string-initialization


stratakis:
### Port, board and/or hardware

x86_64

### MicroPython version

1.24.1

### Reproduction

When compiling micropython on Fedora with the latest GCC 15, a new warning is now enabled by default through -Wextra:  -Wunterminated-string-initialization.

This makes the compilation fail for the arrays at:
https://github.com/micropython/micropython/blob/master/py/emitinlinethumb.c#L152
https://github.com/micropython/micropython/blob/v1.24.1/py/emitinlinethumb.c#L370
https://github.com/micropython/micropython/blob/v1.24.1/py/emitinlinethumb.c#L391
https://github.com/micropython/micropython/blob/v1.24.1/py/emitinlinethumb.c#L430 
https://github.com/micropython/micropython/blob/master/py/emitinlinextensa.c#L121

### Expected behaviour

Compilation passes.

### Observed behaviour

These errors are observed for each array element.

<snippet>

`../py/emitinlinextensa.c:122:9: error: initializer-string for array of ‘unsigned char’ is too long [-Werror=unterminated-string-initialization]
  122 |     {0, "a0\0"}`

`../py/emitinlinethumb.c:153:9: error: initializer-string for array of ‘unsigned char’ is too long [-Werror=unterminated-string-initialization]
  153 |     {0, "r0\0"}`

`../py/emitinlinethumb.c:371:24: error: initializer-string for array of ‘unsigned char’ is too long [-Werror=unterminated-string-initialization]
  371 |     { ASM_THUMB_CC_EQ, "eq" }`

`../py/emitinlinethumb.c:394:34: error: initializer-string for array of ‘char’ is too long [-Werror=unterminated-string-initialization]
  394 |     { X(ASM_THUMB_FORMAT_4_LSR), "lsr" },`

`../py/emitinlinethumb.c:431:13: error: initializer-string for array of ‘char’ is too long [-Werror=unterminated-string-initialization]
  431 |     { 0x30, "add" },`
<snippet>

Full [build.log](https://github.com/user-attachments/files/18650365/build.log) is attached.

### Additional Information

Similar issues from other projects:
https://github.com/Mbed-TLS/mbedtls/issues/9944
https://github.com/besser82/libxcrypt/issues/194

### Code of Conduct

Yes, I agree

projectgus:
Thanks @stratakis for the heads-up.

[...]

PR #17215, opened 2025-04-30
py/emitinlinethumb: Use nonstring attribute to suppress compiler warnings

dpgeorge:
### Summary

Windows mingw CI is failing due to compiler warnings.  This PR attempts to fix that.

### Testing

Done by CI.

### DIFF
--- a/py/emitinlinethumb.c
+++ b/py/emitinlinethumb.c
@@ -35,6 +35,18 @@
 
 #if MICROPY_EMIT_INLINE_THUMB
 
+#ifndef MP_ATTR_NONSTRING
+#ifdef __has_attribute
+#if __has_attribute(nonstring)
+#define MP_ATTR_NONSTRING __attribute__((nonstring))
+#endif
+#endif
+#endif
+
+#ifndef MP_ATTR_NONSTRING
+#define MP_ATTR_NONSTRING
+#endif
+
 typedef enum {
 // define rules with a compile function
 #define DEF_RULE(rule, comp, kind, ...) PN_##rule,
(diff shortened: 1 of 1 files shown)


py/emitinlinethumb.c
```diff
@@ -35,6 +35,12 @@
 
 #if MICROPY_EMIT_INLINE_THUMB
 
+#if !defined(__clang__) && defined(__GNUC__) && __GNUC__ > 10
```
dpgeorge, line 38: I don't actually know when this attribute was supported in gcc, I just picked 10 as a guess that should work fine.

dlech: how about `#if __has_attribute(__nonstring__)`?

dpgeorge: Ah, I didn't know about that, thanks for the hint!

Now changed.

projectgus (approved):

dpgeorge:
I forgot about #16692, but this PR should fix that issue.

dpgeorge:
Should probably also fix the two cases in `extmod/moductypes.c` as well.

DvdGiessen:
Note: LittleFS version 1 also raises this new error, which blocks the Unix port build for me:

```
../../lib/littlefs/lfs1.c: In function ‘lfs1_format’:
../../lib/littlefs/lfs1.c:2144:24: error: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (9 chars into 8 available) [-Werror=unterminated-string-initialization]
 2144 |             .d.magic = {"littlefs"},
      |                        ^
```

Since that version isn't getting updates anymore (all development is now happening on version 2, the last release of version 1 was over 6 years ago), I think we can safely patch it.

dpgeorge:
Closing in favour of #17269.

PR #17230, opened 2025-04-30
stm32/machine_adc: Enable ADC re-read errata handling for STM32WB55.

andrewleech:
### Summary
For STM32WB errata 2.7.1:
"Wrong ADC result if conversion done late after calibration or previous conversion"
states an incorrect reading is returned if more than 1ms has elapsed since the last
reading or calibration. According to the errata, this can be avoided by performing
two consecutive ADC conversions and keeping the second result.

This matches existing handling for a comparable STM32G4 errata 2.7.7 

### Testing

Before this change the first read from an ADC channel often looked wrong, it turns out we'd previously added what was essentially a double-read in our application "to help the adc stabilise". Fixing it here in micropython is far more reliable and efficient.

I initially though it would be better to add some timing checks / "how long since last read" and only re-read if needed, however I temporarily added some timing checks with `ticks_us`/`ticks_diff` and on my STM32WB55 @ 64Mhz and each ADC read seemed to take only ~ 3us, certainly not worth the overhead of storing and checking the time of last read.

### DIFF
--- a/ports/stm32/machine_adc.c
+++ b/ports/stm32/machine_adc.c
@@ -445,10 +445,13 @@ static void adc_config_channel(ADC_TypeDef *adc, uint32_t channel, uint32_t samp
 
 static uint32_t adc_read_channel(ADC_TypeDef *adc) {
     uint32_t value;
-    #if defined(STM32G4)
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
This looks like a reasonable work around for the errata.

Issue #17229, opened 2025-04-30
RP2350 Watchdog and lightsleep issue


cpottle9:
### Port, board and/or hardware

Raspberry Pi Pico2

### MicroPython version

MicroPython v1.25.0 on 2025-04-15; Raspberry Pi Pico2 with RP2350

### Reproduction

Copy the following into wdt_lightsleep_2350.py.
```py
from machine import lightsleep, WDT
from time import sleep_ms

wdt = WDT(timeout=5000)

lightsleep(3000)

sleep_ms(3000)
print('Should not print')
sleep_ms(3000)
```
Do 'mpremote run wdt_lightsleep_2350.py'.

### Expected behaviour

Expect the same behavior seen when the lightsleep is replaced with sleep_ms.
File wdt_sleep_ms_2350.py
```py
from machine import lightsleep, WDT
from time import sleep_ms

wdt = WDT(timeout=5000)

sleep_ms(3000)

sleep_ms(3000)
print('Should not print')
sleep_ms(3000)
```
```
>> mpremote run wdt_sleep_ms_2350.py
Traceback (most recent call last):
  File "/home/picompute/.local/bin/mpremote", line 8, in <module>
    sys.exit(main())
```
Note: most of traceback removed.
This time the print did not execute because the watchdog timer triggered during the second sleep_ms(3000).

### Observed behaviour

The print statement does execute and the watchdog timer triggers during the last sleep_ms(3000).

```
>> mpremote run wdt_lightsleep_2350.py
Should not print
Traceback (most recent call last):
  File "/home/picompute/.local/bin/mpremote", line 8, in <module>
    sys.exit(main())
```
Note: most of traceback removed.

### Additional Information

I have changes to lightsleep in modmachine.c which fix this problem.
I will prepare a pull request with this fix.

The root cause is the watchdog timer does not count down while lightsleep() is running. 

### Code of Conduct

Yes, I agree

alke:
in RP2350 lightsleep is not working it just pass with no sleep

from machine import lightsleep
from time import sleep_ms
led = machine.Pin("LED", machine.Pin.OUT)
while True:
    led.on()
    lightsleep(3000)
#    sleep_ms(3000)
    led.off()
    sleep_ms(3000)

alke:
sorry for the hashtag this is the code with no remark
from machine import lightsleep
from time import sleep_ms
led = machine.Pin("LED", machine.Pin.OUT)
while True:
  led.on()
  lightsleep(3000)
[...]

Issue #17228, opened 2025-04-30
RP2040 Watchdog and lightsleep issue


cpottle9:
### Port, board and/or hardware

Raspberry PI Pico and Pico2

### MicroPython version

MicroPython v1.25.0 on 2025-04-15; Raspberry Pi Pico with RP2040

### Reproduction

Test file named wdt_lightsleep.py.
```py
from machine import lightsleep, WDT

wdt = WDT(timeout=3000)
lightsleep(5000)
```
Do 'mpremote run wdt_lightsleep.py'. 
After 3 seconds mpremote will terminate with an OSError.
Try to reconnect by running mpremote with no arguments.
mpremote will fail.

### Expected behaviour

Expect it to behave the same as if lightsleep were replaced with time.sleep_ms.
Call this test file wdt_sleep_ms.py
```py
from machine import WDT
from time import sleep_ms

wdt = WDT(timeout=3000)
sleep_ms(5000)
```
In this case, after do 'mpremote run wdt_sleep_ms.py'.
After 3 seconds mpremote will report an OSError because the RP2040 does a soft reset.
Run mpremote again and it will connect successfully.

### Observed behaviour

The RP2040 seems to get stuck. After running 'mpremote run wdt_lightsleep.py' attempts to connect running mpremote fail.
On my Raspberry Pi 4 I see an error message of 'mpremote: no device found'.

A power cycle is required to recover.

One more observation, I have a USB power meter connected between the PI 4 and RPI Pico.
Normally it reports about 19 milli-amps consumed, about 5 milli-amps while the lightsleep is running.
After the test completes it reports about 1.00 milli-amps.

### Additional Information

I have changes to lightsleep in modmachine.c which fix this problem.
I plan to prepare a pull request with this fix.

The behavior reported here is seen on RP2040 only.

There is a different problem with RP2350.
I will open another issue for that problem.

### Code of Conduct

Yes, I agree

dpgeorge:
Thanks for the very clear bug report.  I can reproduce the issue.

PR #17232, opened 2025-05-01
ports/rp2/modmachine.c: Fix lightsleep watchdog interactions.

cpottle9:
### Summary
Aside: I am a little hesitant to propose these changes. If the maintainers decide to reject this pull request I will understand.
Lightsleep on rp2 is gradually getting more complex and these changes would contribute to that complexity.

Fixes RP2040 issue #17228 and RP2350 issue #17229.

RP2040 issue is when watchdog timer expires while lightsleep is running the RP2040 will lock up during power-on startup.
This occurs because the power-on state machine is configured to not re-initialize the ROSC and lightsleep stopped the ROSC.

RP2350 issue is the watchdog timer does not decrement while lightsleep is running.
Once lightsleep returns the timer will start decrementing again. This occurs because lightsleep clears the bit
CLOCKS_SLEEP_EN1_CLK_SYS_WATCHDOG_BITS while sleeping.

### Testing
I ran the reproduction methods from the two issues.
I also ran tests/port/rp2/rp2_lightsleep.py.

General alternative. Document that lightsleep and watchdog don't play well together and leave the code as is.

Alternatives for RP2040 issue #17228:
- Raise issue for pico-sdk. Get them to change function _watchdog_enable() so the ROSC and XOSC do get reset on a watchdog timeout. The existing behavior in _watchdog_enable() has been there from day one. There might not be a good reason for it. Disadvantage: a possible long wait for a new pico-sdk release.
- Modify lightsleep to not stop the ROSC. I measure 4.96 milli-amps when lightsleep disables ROSC and 5.09 milli-amps when it does not. This is a small power increase.

Alternatives for the RP2350 issue #17229:
- I don't have any :-).


cpottle9:
It seems I don't know enough about github to resolve a merge conflict :-).

I will close this pull request and create a fresh pull request from the most recent commit.

PR #17237, opened 2025-05-01
esp32/network_lan: Add PHY_GENERIC device type.

elvis-epx:
### Summary

Adds support to the new PHY_GENERIC board type, added in ESP-IDF v5.4.0 [1].

This PHY driver was added to ESP-IDF to support "generic"/oddball PHY LAN chips like the JL1101, which offer no features beyond the bare 802.3 PHY standard and don't actually need a chip-specific driver (see discussion at [2]).

[1] https://github.com/espressif/esp-idf/commit/0738314308ad36a73601ddb8bb82f1dcbfe1f550
[2] https://github.com/espressif/esp-eth-drivers/pull/28

### Testing

Tested on DTWonder automation device, which is based on ESP32 and with JL1101 LAN chip.


ports/esp32/network_lan.c
```diff
@@ -221,6 +222,13 @@ static mp_obj_t get_lan(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_ar
         case PHY_IP101:
             self->phy = esp_eth_phy_new_ip101(&phy_config);
             break;
+        case PHY_GENERIC:
+            #if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 4, 0)
+            self->phy = esp_eth_phy_new_generic(&phy_config);
+            #else
+            mp_raise_msg(&mp_type_OSError, MP_ERROR_TEXT("PHY_GENERIC not supported in ESP-IDF < 5.4"));
```
dpgeorge, line 229: Instead of raising an exception like this, I would guard the entire feature, including the `PHY_GENERIC` constant, behind a compile-time option.

The same way the recent LAN8670 support was done in 10f6c0699e52857a4f7350a5d3a9561862c98a8d .

dpgeorge (approved):

Issue #17246, opened 2025-05-05
BLE init failures can cause infinite loop in scheduler


andrewleech:
### Port, board and/or hardware

stm32wb55

### MicroPython version

master branch, slightly newer than v1.25

### Reproduction

On a stm32wb55 based board with the rfcore (WS) firmware blob wiped / damaged.
```
from bluetooth import BLE
BLE().active(1)
```
Ends up stuck with `tl_ble_wait_resp: timeout` repeated on the command line every few seconds.

### Expected behaviour

The `tl_ble_wait_resp: timeout` message is expected due to lack of response from the rfcore, however there is a timeout in C so this should only retry a limited number of times (2 seconds worth) before aborting with a Timeout error: https://github.com/micropython/micropython/blob/79abdad9e97f18f45650e1abce64ee51c3372953/extmod/nimble/modbluetooth_nimble.c#L632

### Observed behaviour

Instead, the `tl_ble_wait_resp: timeout` message repeats forever. On my application this cannot be ctrl-c interrupted, it reboots after some time only because we have a WDT running.

### Additional Information

I've debugged the issue down to the scheduler, specifically the C STATIC_NODES section of it:

![Image](https://github.com/user-attachments/assets/02ef81de-9431-40c2-aabd-4e9c46e8a55b)

In this case the C scheduler is used to run `mp_bluetooth_hci_poll()`. Due to timeout trying to communicate with hci a  `ble_hs_reset()` gets run on the nimble os_eventq which starts by resetting the hci transport.
That flows through to mp_bluetooth_hci_uart_init() which itself starts the hci polling...

![Image](https://github.com/user-attachments/assets/bd502ea5-c1d6-4c9a-a3c6-d28db0a03535)

This re-inserts the hci poll function into head of the c schedule linked list meaing it's ready to be run immediately, forming an infinite loop in `mp_sched_run_pending()`

### Code of Conduct

Yes, I agree

andrewleech:
This is a bit of a trap with the C static scheduler; if the scheduled function itself triggers the scheduling of a c function it creates the infinite loop here. @dpgeorge should the scheduler have a limit to the number of funtions processed in a single loop? 
Or would it be enough / appropriate to document this limitation and just change the `mp_bluetooth_hci_start_polling()` implementation from `mp_bluetooth_hci_poll_now()` to a `mp_bluetooth_hci_poll_in_ms(100)`

dpgeorge:
This is now resolved.  See links above.

PR #17247, opened 2025-05-05
extmod/nimble: Add timeout to deinit.

andrewleech:
### Summary

If the BLE radio stops responding before deinit is called the function can get stuck waiting for an event that is never received, particularly if the radio is external or on a separate core.

This PR adds a timeout, similar to the timeout already used in the init function. That init function timeout was also
simplified as part of this change.

Relates to https://github.com/micropython/micropython/issues/17246

### Testing

This has been tested on a stm32wb55 where the deinit was intermittently locking up waiting for the state to change.
This happened on a semi-regular basis on an application automated test suite which was running through hundreds of cycles over a number of hours.

The main alternative to this is relying on watchdog to break the unit out of the infinite C loop once the failure occurs, but this is not appropriate for all applications.

### DIFF
--- a/extmod/nimble/modbluetooth_nimble.c
+++ b/extmod/nimble/modbluetooth_nimble.c
@@ -60,6 +60,7 @@
 static uint8_t nimble_address_mode = BLE_OWN_ADDR_RANDOM;
 
 #define NIMBLE_STARTUP_TIMEOUT 2000
+#define NIMBLE_SHUTDOWN_TIMEOUT 500
 
 // Any BLE_HS_xxx code not in this table will default to MP_EIO.
 static int8_t ble_hs_err_to_errno_table[] = {
@@ -554,7 +555,7 @@ static void ble_hs_shutdown_stop_cb(int status, void *arg) {
 
 static struct ble_hs_stop_listener ble_hs_shutdown_stop_listener;
 
-void mp_bluetooth_nimble_port_shutdown(void) {
+int mp_bluetooth_nimble_port_shutdown(void) {
     DEBUG_printf("mp_bluetooth_nimble_port_shutdown (nimble default)\n");
     // By default, just call ble_hs_stop directly and wait for the stack to stop.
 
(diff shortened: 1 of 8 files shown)


extmod/nimble/modbluetooth_nimble.c
```diff
@@ -629,13 +641,12 @@ int mp_bluetooth_init(void) {
 
     // Run the scheduler while we wait for stack startup.
     // On non-ringbuffer builds (NimBLE on STM32/Unix) this will also poll the UART and run the event queue.
-    mp_uint_t timeout_start_ticks_ms = mp_hal_ticks_ms();
+    volatile mp_uint_t start = mp_hal_ticks_ms();
```
andrewleech, line 644: `volatile` wasn't meant to be left here, I'll remove

andrewleech:
Should extend this to more closely match the init functionality with regard to raising exceptions on failure

extmod/nimble/modbluetooth_nimble.c
```diff
@@ -629,13 +636,12 @@ int mp_bluetooth_init(void) {
...
-    mp_uint_t timeout_start_ticks_ms = mp_hal_ticks_ms();
+    mp_uint_t start = mp_hal_ticks_ms();
     while (mp_bluetooth_nimble_ble_state != MP_BLUETOOTH_NIMBLE_BLE_STATE_ACTIVE) {
-        uint32_t elapsed = mp_hal_ticks_ms() - timeout_start_ticks_ms;
-        if (elapsed > NIMBLE_STARTUP_TIMEOUT) {
+        if ((mp_hal_ticks_ms() - start) > NIMBLE_STARTUP_TIMEOUT) {
             break;
         }
-        mp_event_wait_ms(NIMBLE_STARTUP_TIMEOUT - elapsed);
+        mp_event_wait_ms(1);
```
dpgeorge, line 644: I'm not sure exactly why this logic was changed?  It should be more efficient to do a sleep for `NIMBLE_STARTUP_TIMEOUT - elapsed` than many sleeps for 1ms.  The idea is that `mp_event_wait_ms()` can return early if there's a system event to process, so it won't wait that whole time if it doesn't need to.

andrewleech: I thought this might be over-complicating it for minimal gain, figured it might be better to just simplify it to the same timeout logic used most other places in the codebase. 

But yes it's true the original version should be slightly more efficient with likely no charge in code size so I could change birth these timeouts to the original pattern used here.

andrewleech: With the original `mp_event_wait_ms(NIMBLE_STARTUP_TIMEOUT - elapsed);` scheme for both timeouts:
```
 397964     368   42356  440688   6b970 build-NUCLEO_WB55/firmware.elf
```
With the "simpler" one here `mp_event_wait_ms(1);`:
```
 397956     368   42356  440680   6b968 build-NUCLEO_WB55/firmware.elf
```

So the calculated timeout costs 4 bytes of flash (per copy of the logic). Considering this is only on boards that already are big enough to hold the BLE stack that's probably reasonable for more efficient use - BLE application are often battery ones and these kind of optimisations do (slightly) improve power use.

dpgeorge (approved):

PR #17248, opened 2025-05-05
py/scheduler: warning about C callbacks scheduling new tasks.

andrewleech:
### Summary

If a `mp_sched_schedule_node` based C scheduler function re-schedules itself then the `mp_sched_run_pending` function can get stuck in an infinite loop re-running the function where ctrl-c cannot break out of it anymore.

~~This change adds a unit test to display the behavior (in a self-limiting fashion) along with a minimal change to auto-break out of `mp_sched_run_pending` if the node is re-scheduled. It will still be run on next run of `mp_sched_run_pending` but other aspects of the application / micropython will be able to run as well in this case.~~

This MR adds a coverage unit test showing the behavior and adds a comment in the scheduler warning developers of this behavior. It should only apply to people working on C code within the micropython codebase.

See https://github.com/micropython/micropython/issues/17246

### Testing

Tested on a STM32WB55 with manually broken rfcore, before this it was stuck in an infinite loop trying to init the radio.

If there are two C scheduled functions that continually re-schedule each other this fix won't catch that, but I tried to keep the minimal fix as simple as possible as I believe the goal is to keep `mp_sched_run_pending` as fast as possible as its run in the interpreter core?

### DIFF
--- a/ports/unix/coverage.c
+++ b/ports/unix/coverage.c
@@ -184,6 +184,17 @@ static void pairheap_test(size_t nops, int *ops) {
     mp_printf(&mp_plat_print, "\n");
 }
 
+static mp_sched_node_t mp_coverage_sched_node;
+
+static void coverage_sched_function(mp_sched_node_t *node) {
+    (void)node;
+    static mp_uint_t loop = 3;
+    mp_printf(&mp_plat_print, "scheduled function\n");
+    if (--loop) {
+        mp_sched_schedule_node(&mp_coverage_sched_node, coverage_sched_function);
+    }
+}
+
 // function to run extra tests for things that can't be checked by scripts
 static mp_obj_t extra_coverage(void) {
     // mp_printf (used by ports that don't have a native printf)
@@ -621,6 +632,12 @@ static mp_obj_t extra_coverage(void) {
             mp_obj_print_exception(&mp_plat_print, MP_OBJ_FROM_PTR(nlr.ret_val));
         }
         mp_handle_pending(true);
+
+        mp_sched_schedule_node(&mp_coverage_sched_node, coverage_sched_function);
+        for (int i = 0; i <= 3; ++i) {
+            mp_printf(&mp_plat_print, "loop\n");
+            mp_handle_pending(true);
+        }
     }
 
     // ringbuf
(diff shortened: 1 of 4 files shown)


andrewleech:
@dpgeorge  I've updated this to add a warning comment to future developers and updated the unit test to show the behaviour. I don't believe there was previously any unit test coverage of the C scheduler so this is beneficial for that reason alone.

projectgus:
FWIW, I bumped into this problem a while back (maybe while working on USB) and came up with a patch to allow C scheduler callbacks to "chain" a call to another callback while still supporting normal execution. Then I think I turned out not to need the patch, as I don't seem to have ever commited it.

From memory it was pretty simple, something like this (untested):

```diff
diff --git i/py/scheduler.c w/py/scheduler.c
index 2170b9577e..969202f5ce 100644
--- i/py/scheduler.c
+++ w/py/scheduler.c
@@ -88,6 +88,7 @@ static inline void mp_sched_run_pending(void) {
 
     #if MICROPY_SCHEDULER_STATIC_NODES
     // Run all pending C callbacks.
+    mp_sched_node_t *original_tail = MP_STATE_VM(sched_tail);
     while (MP_STATE_VM(sched_head) != NULL) {
         mp_sched_node_t *node = MP_STATE_VM(sched_head);
         MP_STATE_VM(sched_head) = node->next;
@@ -99,6 +100,9 @@ static inline void mp_sched_run_pending(void) {
         MICROPY_END_ATOMIC_SECTION(atomic_state);
         callback(node);
         atomic_state = MICROPY_BEGIN_ATOMIC_SECTION();
+        if (node == original_tail) {
+            break; // Don't execute any callbacks scheduled during this run
+        }
     }
     #endif

```

If there isn't somewhere we rely on the current behaviour then maybe this is an alternate way forward?

dpgeorge:
I'm happy to go with @projectgus suggestion to stop processing when the loop gets to the tail of the schedule linked-list.

Then it won't ever get into a loop, no matter how many new callbacks are scheduled while the scheduler is running.

@andrewleech if we had this solution, does it fix your original bug, and make #17264 unnecessary?

dpgeorge:
Fixed as discussed above by 7f274c7550654160f73b0de8fa73338dc58109a6

PR #17261, opened 2025-05-07
extmod/modlwip: Implement a queue of incoming UDP/raw packets

dpgeorge:
### Summary

The bare-metal lwIP socket interface is currently quite limited when used for UDP streams, because it only allows one outstanding incoming UDP packet.  If one UDP packet is waiting to be `socket.recv`'d and another one comes along, then the second one is simply dropped.

This PR implements a queue for incoming UDP and raw packets.  The queue depth is fixed at compile time, and is currently 4.

This allows better use of UDP connections, eg more efficient.  It also makes DTLS work better which sometimes has a queue of UDP packets (eg during the connection phase).

### Testing

A new network multi-test is added which sends a burst of UDP packets from the client, and the server doesn't `recv` them until they have all been sent.

Tested on PYBD-SF2 and RPI_PICO2_W.  The new test passes (it fails without the UDP queue added here).

Also tested on unix and ESP32_GENERIC, both of which pass (they already have a UDP queue).

- Some memory is needed for the queue, and is allocated on the MP heap.  That memory holds a ring buffer of pbufs.  Unfortunately pbufs don't support "packet queue" mode in lwIP yet, or maybe we could just link-link the pubfs together (also need to store the peer address and port, which probably requires additional memory anyway).
- The queue depth is fixed at compile time.  That's probably adequate.

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -286,6 +286,15 @@ static const int error_lookup_table[] = {
 #define MOD_NETWORK_SOCK_DGRAM (2)
 #define MOD_NETWORK_SOCK_RAW (3)
 
+// Total queue length for buffered UDP/raw incoming packets.
+#define LWIP_INCOMING_PACKET_QUEUE_LEN (4)
+
+typedef struct _lwip_incoming_packet_t {
+    struct pbuf *pbuf;
+    ip_addr_t peer_addr;
+    uint16_t peer_port;
+} lwip_incoming_packet_t;
+
 typedef struct _lwip_socket_obj_t {
     mp_obj_base_t base;
 
@@ -294,8 +303,11 @@ typedef struct _lwip_socket_obj_t {
         struct udp_pcb *udp;
         struct raw_pcb *raw;
     } pcb;
+
+    // Data structure that holds incoming pbuf's.
+    // Each socket type has different state that it needs to keep track of.
     volatile union {
-        struct pbuf *pbuf;
+        // TCP listening sockets have a queue of incoming connections, implemented as a ringbuffer.
         struct {
             uint8_t alloc;
             uint8_t iget;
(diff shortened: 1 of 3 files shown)


extmod/modlwip.c
```diff
@@ -286,6 +286,15 @@ static const int error_lookup_table[] = {
 #define MOD_NETWORK_SOCK_DGRAM (2)
 #define MOD_NETWORK_SOCK_RAW (3)
 
+// Total queue length for UDP/raw connections.
```
projectgus, line 289:
```suggestion
// Total queue length for buffered UDP/raw incoming packets.
```

(That's right, isn't it?)

extmod/modlwip.c
```diff
@@ -305,10 +317,24 @@ typedef struct _lwip_socket_obj_t {
...
             } tcp;
         } connection;
+
+        // Connected TCP sockets have a single incoming pbuf that is
+        // appended to when new data comes in.
+        struct {
+            struct pbuf *pbuf;
+        } tcp;
+
+        // UDP and raw sockets have a queue of incoming pbuf's.
```
projectgus, line 327:
```suggestion
        // UDP and raw sockets have a queue of incoming pbuf's, implemented as a ringbuffer
```

extmod/modlwip.c
```diff
@@ -305,10 +317,24 @@ typedef struct _lwip_socket_obj_t {
...
+        struct {
+            uint8_t iget;
```
projectgus, line 329:
```suggestion
            uint8_t iget;   // Ringbuffer read index
```

(and same for `iput`)

projectgus (approved):
Implementation and test look solid to me. I have a couple of minor comment suggestions for future reference, but that's all.

extmod/modlwip.c
dpgeorge: Yes, that's right.  Comment updated.

extmod/modlwip.c
dpgeorge: Updated.

extmod/modlwip.c
dpgeorge: Updated for both lines.

PR #17263, opened 2025-05-07
docs: Add documentation for the io.IOBase class.

andrewleech:
### Summary

Documents the `io.IOBase` class which is used to create custom stream objects in Python. Subclasses override `readinto`, `write`, and `ioctl` to create objects that work with `print()`, `json.dump()`, `select.poll()`, user filesystems, etc.

The docs focus on the three methods the C stream layer actually dispatches to, rather than documenting the full stream API which is provided automatically by the C infrastructure. The ioctl section covers the common operations (flush, poll, close, buffer size) in the same style as vfs.AbstractBlockDev.ioctl.

Also fixes a statement in io.rst that incorrectly claimed streams cannot be subclassed in pure Python.

### Testing

Built docs locally with `make html`, renders correctly.

### Generative AI

I used generative AI tools when creating this PR, but a human has checked the code and is responsible for the description above.

### DIFF
--- a/docs/library/io.rst
+++ b/docs/library/io.rst
@@ -71,8 +71,9 @@ buffered, they aren't in MicroPython. (Indeed, that's one of the cases
 for which we may introduce buffering support.)
 
 Note that for efficiency, MicroPython doesn't provide abstract base
-classes corresponding to the hierarchy above, and it's not possible
-to implement, or subclass, a stream class in pure Python.
+classes corresponding to the hierarchy above.  However, the
+:class:`IOBase` class can be subclassed to implement custom stream
+objects in pure Python.
 
 Functions
 ---------
@@ -85,6 +86,69 @@ Functions
 
 Classes
 -------
+
+.. class:: IOBase()
+
+   Base class for implementing custom stream objects in Python. Subclasses
+   can override ``readinto``, ``write``, and ``ioctl`` to create objects
+   that work with ``print()``, ``json.dump()``, ``select.poll()``,
+   ``open()`` via a user filesystem, and other stream consumers.
+
+   .. admonition:: Difference to CPython
+      :class: attention
+
+      In CPython, ``io.IOBase`` has a much larger API surface. MicroPython's
+      ``IOBase`` is minimal: the C stream infrastructure provides standard
+      methods like ``read()``, ``readline()``, ``seek()``, ``close()``, and
+      ``flush()`` automatically. Subclasses only need to implement the
+      methods below.
+
+   Subclasses implement some or all of the following methods. The C stream
+   layer calls these internally when user code calls standard stream
+   functions.
+
+   .. method:: IOBase.readinto(buf)
+
+      Read data into *buf* (a ``bytearray`` sized by the caller). Return the
+      number of bytes read, 0 at EOF, or ``None`` if no data is available on
+      a non-blocking stream. Return a negative errno value (e.g. ``-errno.EIO``
+      or ``-1``) to signal an error.
+
+   .. method:: IOBase.write(buf)
+
+      Write *buf* (a ``bytearray``) to the stream. Return the number of
+      bytes written, or ``None`` if a non-blocking stream cannot accept data.
+      Return a negative errno value to signal an error.
+
+   .. method:: IOBase.ioctl(op, arg)
+
+      Control the stream and query its properties. The operation to perform
+      is given by *op* which is one of the following integers:
+
+        - 1 -- flush write buffers (*arg* is unused)
+        - 3 -- poll for readiness; *arg* is a bitmask of events to check,
+          return a bitmask of ready events. Poll flags:
+
+          * ``0x0001`` -- data available for reading
+          * ``0x0004`` -- stream ready for writing
+          * ``0x0008`` -- error condition
+          * ``0x0010`` -- hang up (e.g. connection closed)
+          * ``0x0020`` -- invalid request
+
+        - 4 -- close the stream (*arg* is unused)
+        - 11 -- return the preferred read buffer size, or 0 (*arg* is unused)
+
+      As a minimum ``ioctl(4, ...)`` should be handled to support stream
+      closure. Implement ``ioctl(3, ...)`` if the stream will be used with
+      ``select.poll()`` or ``asyncio``.
+
+      Other operations exist for advanced use cases (2 = seek, 5 = timeout,
+      10 = fileno); see ``py/stream.h`` for the full list.
+
+      Must always return an integer. Return 0 for success, or ``-1`` for
+      unsupported operations. (Returning 0 for an unhandled operation tells
+      the C layer the operation was processed successfully, which may cause
+      incorrect behaviour.)
 
 .. class:: StringIO([string])
 .. class:: BytesIO([string])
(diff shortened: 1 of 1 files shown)


docs/library/io.IOBase.rst
```diff
@@ -0,0 +1,414 @@
...
+   * ``1`` -- current stream position; *offset* may be negative
+   * ``2`` -- end of the stream; *offset* is usually negative
+
+   Returns the new absolute position.
+
+.. method:: tell()
+
+   Return the current stream position.
+
+.. method:: ioctl(request, arg=0)
```
Josverl: `ioctl` is also described in other locations , and while a bit of duplication is not a bad thing, I think the parameter names should be consistent unless there is a specific reason not to.

- ioctl(request, arg=0)
- ioctl(op, arg)
- Flash.ioctl(cmd, arg)

https://docs.micropython.org/en/latest/library/vfs.html#vfs.AbstractBlockDev.ioctl
https://docs.micropython.org/en/latest/library/zephyr.DiskAccess.html#zephyr.DiskAccess.ioctl
[more...](https://docs.micropython.org/en/latest/search.html?q=ioctl&check_keywords=yes&area=default)

docs/library/io.IOBase.rst
```diff
@@ -0,0 +1,414 @@
...
+--------------
+
+If an operation is not supported by a stream, ``UnsupportedOperation`` (which
+is an alias for ``OSError`` in MicroPython) or ``TypeError`` may be raised.
+Other ``OSError`` exceptions may be raised for I/O errors.
+
+Example
+-------
+
+Here is an example of a custom stream class inheriting from `io.IOBase`,
```
Josverl: At 230 lines this is a quite large example. 
would it be possible to add a simpler example first and keep this as an advanced example ?

Josverl:
Thanks Andrew, I think this a very good addition.
I had been trying to add version of this to the micropython-stubs project earlier, but did not get the quality i was after.
.

Josverl (changes requested):

docs/library/io.IOBase.rst
andrewleech: Thanks for the references, I've tried to improve this here.

docs/library/io.IOBase.rst
andrewleech: Good point, I've got a smaller one to start with now

andrewleech:
Thanks for the feedback, I've rewritten to address suggestions and improve the overall writing style.

- Renamed `ioctl(request, arg=0)` to `ioctl(op, arg)` to match vfs.AbstractBlockDev. All ioctl methods across the codebase are positional-only so the parameter names are cosmetic, but they should be consistent in docs regardless.
- Replaced the 230-line example with two shorter ones: a minimal write-only stream (~15 lines) and a pollable ring buffer with select.poll() (~35 lines). Both tested on the unix port.
- Cut the methods section down to just the three methods that IOBase subclasses actually implement: `readinto(buf)`, `write(buf)`, `ioctl(op, arg)`. The previous version documented read(), readline(), seek(), tell(), flush(), close() which are all provided automatically by the C stream infrastructure and don't need to be implemented by subclasses.
- Trimmed the ioctl ops list to the ones Python subclasses commonly handle (flush, poll, close, buffer size) in the same terse numbered-list style as vfs.rst. The rarely-used ops (seek, timeout, fileno etc.) get a brief mention pointing to py/stream.h.
- Added a CPython difference admonition since MicroPython's IOBase is much more minimal.
- Fixed a statement in io.rst that incorrectly claimed streams can't be subclassed in pure Python.

Josverl (approved):

dpgeorge:
I suggest putting the docs for `IOBase` in `io.rst`, before `StringIO` and `BytesIO` (that follows CPython).  And put the two examples at the bottom of `io.rst`.

docs/library/io.rst
```diff
@@ -86,6 +87,73 @@ Functions
 Classes
 -------
 
+.. _io.IOBase:
```
dpgeorge: Is this label needed?

docs/library/io.rst
```diff
@@ -86,6 +87,73 @@ Functions
...
+
+   Subclasses implement some or all of the following methods. The C stream
+   layer calls these internally when user code calls standard stream
+   functions.
+
+   .. method:: IOBase.readinto(buf)
+
+      Read data into *buf* (a ``bytearray`` sized by the caller). Return the
+      number of bytes read, 0 at EOF, or ``None`` if no data is available on
+      a non-blocking stream.
```
dpgeorge: Also worth mentioning that a negative return value means error, and corresponds to negative of an `errno` code.  Eg `return -errno.EIO`, or just `return -1` for a general error if you don't want to rely on `errno`.

Edit: I see this is mentioned below... but maybe worth putting here instead, since the return value is already discussed here in detail?

docs/library/io.rst
```diff
@@ -86,6 +87,73 @@ Functions
...
+
+   .. method:: IOBase.write(buf)
+
+      Write *buf* (a ``bytearray``) to the stream. Return the number of
+      bytes written, or ``None`` if a non-blocking stream cannot accept data.
```
dpgeorge, line 120: As above, a negative return means error.

docs/library/io.rst
```diff
@@ -86,6 +87,73 @@ Functions
...
+      10 = fileno); see ``py/stream.h`` for the full list.
+
+      ``ioctl`` must always return an integer. Return 0 for success on
+      operations that don't return data.
+
+   For all three methods, a negative return value signals an error; the
+   absolute value is used as the ``errno`` code and raised as ``OSError``.
+   For ``readinto`` and ``write``, returning ``None`` indicates a
+   non-blocking stream with no data available (translated to ``EAGAIN``).
+   ``ioctl`` does not support ``None`` returns.
```
dpgeorge: This is partially repeated above (None).

Maybe worth just detailing the return values in each method (like it's already done for read/write)?  Because otherwise it's a bit awkward that you need to read the entire IOBase docs to know that negative return means error (better to have each method self contained in its docs IMO).

docs/library/io.rst
```diff
@@ -119,3 +187,75 @@ Classes
...
+            return n
+
+        def ioctl(self, op, arg):
+            if op == _MP_STREAM_POLL:
+                if arg & _MP_STREAM_POLL_RD and self._available() > 0:
+                    return _MP_STREAM_POLL_RD
+                return 0
+            if op == _MP_STREAM_CLOSE:
+                return 0
+            return 0
```
dpgeorge: This should be `return -1` to indicate an unsupported op.

docs/library/io.rst
```diff
@@ -119,3 +187,75 @@ Classes
...
+    class MyOutput(io.IOBase):
+        def __init__(self):
+            self.data = bytearray()
+
+        def write(self, buf):
+            self.data.extend(buf)
+            return len(buf)
+
+        def ioctl(self, op, arg):
+            return 0
```
dpgeorge: This needs to be `return -1`.  Otherwise ioctls like `MP_STREAM_SEEK` assume you put valid data in the memory pointed to by arg (because `return 0` means "yes I support this op and processed it").

And then probably you want to implement a no-op `MP_STREAM_CLOSE` that does return 0 so it can be closed without error.

andrewleech:
Thanks @dpgeorge for the review, they're really helpful suggestions. I've addressed all now.

dpgeorge (approved):

PR #17264, opened 2025-05-07
all/mpbthciport.c: Don't restart BLE polling if already running.

andrewleech:
### Summary

When nimble stack  schedules a reset (ble_hs_reset) it's handled in `mp_bluetooth_nimble_os_eventq_run_all()` which itself is run in `mp_bluetooth_hci_poll()`.

The stack reset process starts with running `mp_bluetooth_hci_uart_deinit` then `mp_bluetooth_hci_uart_init`. The init function ends with `mp_bluetooth_hci_start_polling();` which brings us to this PR, a C scheduled hci poll can result in a re-schedule of itself in the event of a stack reset (failure / non-resonse detected from radio).

This was first discovered in https://github.com/micropython/micropython/issues/17246

### Testing

This has been tested on a stm32wb55 with the radio firware wiped so the nimble stack continually attempts to re-init the hci connection.

This issue can also be avoided with the C scheduler loop break first proposed in https://github.com/micropython/micropython/pull/17248/commits/af4a8e9cfc6df2467bb492d4794e377a8d80cb37

### DIFF
--- a/ports/alif/mpbthciport.c
+++ b/ports/alif/mpbthciport.c
@@ -38,6 +38,8 @@ uint8_t mp_bluetooth_hci_cmd_buf[4 + 256];
 static soft_timer_entry_t mp_bluetooth_hci_soft_timer;
 static mp_sched_node_t mp_bluetooth_hci_sched_node;
 
+static bool mp_bluetooth_hci_poll_active = false;
+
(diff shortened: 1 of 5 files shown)


dpgeorge:
Should be fixed by 7f274c7550654160f73b0de8fa73338dc58109a6

PR #17138, opened 2025-04-16
esp32/network_ppp: Restructure to match extmod/network_ppp_lwip.

DvdGiessen:
### Summary

The ESP32 PPP implementation predates the generic implementation in `extmod`. The [new `extmod` implementation](https://github.com/micropython/micropython/pull/14461) has a few advantages such as a better deinit procedure (the ESP32 implementation would not clean up properly and cause crashes if recreated) and using the UART IRQ functionality instead of running a task to read data from the UART.

This change restructures the ESP implementation to be much closer to the new `extmod` version, while also bringing a few tiny improvements from the ESP32 version to the `extmod` version. The diff between `extmod/network_ppp_lwip.c` and `ports/esp32/network_ppp.c` is now a small set of easy to review ESP32 port-specific changes.

### Testing

I've been running these changes for months on a few custom boards (original ESP32 and ESP32S3) with three different modem chips to connect to the Internet via PPP. Those firmware versions do contain a bunch of other changes as well (for example #17135, which I highly recommend also gets considered if this PR gets merged since it fixes a broken behaviour in the UART that can now get triggered by this updated PPP version) so it's not a completely clean test. I'm fairly confident in the changes, and the approach makes sense, but I'd be happy to see other people giving it a go as well and letting me know if they spot anything.

Note for reviewers: Instead of reviewing the changes to the ESP32 version directly, it might be easier to review the differences to the `extmod` version first (they should make sense on their own) and then compare the diff between the `extmod` and ESP32 versions:
<details><summary><code>git diff --no-index extmod/network_ppp_lwip.c ports/esp32/network_ppp.c</code></summary>

```diff
diff --git a/extmod/network_ppp_lwip.c b/ports/esp32/network_ppp.c
index 2c3dac9201..8b700c98ef 100644
--- a/extmod/network_ppp_lwip.c
+++ b/ports/esp32/network_ppp.c
@@ -3,8 +3,11 @@
  *
  * The MIT License (MIT)
  *
+ * Copyright (c) 2018 "Eric Poulsen" <eric@zyxod.com>
  * Copyright (c) 2024 Damien P. George
  *
+ * Based on the ESP IDF example code which is Public Domain / CC0
+ *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
@@ -24,22 +27,25 @@
  * THE SOFTWARE.
  */
 
[...]

PR #17265, opened 2025-05-07
ports/esp32: Fix crash and improve timer interrupt allocation

DvdGiessen:
### Summary

`esp_intr_free` is not safe to call from the timer ISR because it requires the *current task* (the one the ISR interrupted) to be pinned to the same core as the interrupt was allocated on. The ISR is of course guaranteed to interrupt a task on the same core, but it's not guaranteed that that task is also pinned to that core.

This was causing a lockup followed by ISR watchdog timeout in the `machine_uart` RXIDLE timer (which disables the timer from its own callback) when the ISR happened to interrupt a task that was not pinned to a specific core (for example for me if often hit the lwIP TCP/IP thread task).

The first commit in this PR fixes that by merely disabling the interrupt, which *is* safe to do from the ISR since that only requires that we're currently running on the same core (which the ISR always is), regardless of the current task.

Additionally, the second commit makes repeated disabling and enabling of the interrupt (such as the UART RXIDLE does) a bit more efficient by re-enabling instead of reallocating it. That also allowed to removing some code duplication and simplify how `machine_uart` uses the timer.

### Testing

I've been using these fixes in combination with https://github.com/micropython/micropython/pull/17138 (which changed the PPP implementation to use that RXIDLE IRQ). Like that PR, I've been running this on three custom boards (1x original ESP32, 2x ESP32S3) for a few weeks at the moment of writing.

### DIFF
--- a/ports/esp32/machine_timer.c
+++ b/ports/esp32/machine_timer.c
@@ -50,12 +50,13 @@
 const mp_obj_type_t machine_timer_type;
 
 static mp_obj_t machine_timer_init_helper(machine_timer_obj_t *self, mp_uint_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args);
+static mp_obj_t machine_timer_deinit(mp_obj_t self_in);
 
 void machine_timer_deinit_all(void) {
     // Disable, deallocate and remove all timers from list
     machine_timer_obj_t **t = &MP_STATE_PORT(machine_timer_obj_head);
     while (*t != NULL) {
-        machine_timer_disable(*t);
+        machine_timer_deinit(*t);
         machine_timer_obj_t *next = (*t)->next;
         m_del_obj(machine_timer_obj_t, *t);
         *t = next;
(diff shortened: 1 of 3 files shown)


projectgus (commented):
These changes look good to me, @DvdGiessen, and even without the use-after-free bug I agree on principle that it's better to be disabling/re-enabling the interrupt rather than freeing and re-allocating it where we can.

My only concern is that it'd be nice to have either a regression test or test coverage for this. I'm guessing that's non-trivial, yeah? i.e. It doesn't sound like the crash would be easy to reproduce in a unit test, and there's no existing unit test for machine.Timer on esp32 so "extending" test coverage would really mean "implementing" test coverage which is a much bigger task. What do you think?

DvdGiessen:
Thanks for reviewing! Note I don't think it's a use-after-free, it's due to the use of `esp_ipc_call_blocking` that it's more of a deadlock-because-the-ISR-waits-on-IPC-which-cant-run-until-the-ISR-is-finished. While triggering the bug is nondeterministic (due to it depending on which task is currently scheduled) once triggered the behaviour is deterministic.

I agree that the ESP32 could use some machine.Timer tests, but I don't think it'll be easy to test this specific bug. Due to it being dependant on scheduling (and having a non-pinned task in the first place) it won't be easily reproduced with a unit test; best we could do is either engineer the circumstances with a modified build (which seems a bit pointless) or run a reproduction in a loop with the standard build and assert it doesn't crash within some timeout.

projectgus (approved):

dpgeorge (approved):
Thanks, this looks good.

Regarding tests: I'm OK not adding any new tests.  We already have UART IRQ tests which should test part of the changes here.  And eventually we'll have better `machine.Timer` tests that cover all ports.

PR #17270, opened 2025-05-09
rp2: Add temporary workaround for GCC 15.1 build failure.

projectgus:
### Summary

This is a workaround for building `pioasm` for the host on the RP2 port when the compiler is GCC 15.1. This happens transparently during the MicroPython build for boards with CYW43 wireless chip.

The root cause is tracked in https://github.com/raspberrypi/pico-sdk/issues/2448 but patching the tool provides a stop-gap until a new pico-sdk release is available.

### Testing

* Verified `RPI_PICO2_W` build succeeds again with this change and GCC 15.1

* Could wait for pico-sdk to apply a fix and update, however would either need to update to a pre-release SDK or wait a fairly long time. This workaround can be removed when we next update pico-sdk.

### DIFF
--- a/ports/rp2/tools_patch/Findpioasm.cmake
+++ b/ports/rp2/tools_patch/Findpioasm.cmake
@@ -1,0 +1,58 @@
+# Finds (or builds) the pioasm executable
+#
+# This will define the following imported targets
+#
+#     pioasm
+#
+
+# This is a temporary patched copy of pico-sdk file Findpioasm.cmake to work around
+# a host toolchain issue with GCC 15.1:
+# https://github.com/raspberrypi/pico-sdk/issues/2448
+
+if (NOT TARGET pioasm)
+    # todo we would like to use pckgconfig to look for it first
+    # see https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/
+
+    include(ExternalProject)
+
+    set(PIOASM_SOURCE_DIR ${PICO_SDK_PATH}/tools/pioasm)
+    set(PIOASM_BINARY_DIR ${CMAKE_BINARY_DIR}/pioasm)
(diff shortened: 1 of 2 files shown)


ports/rp2/tools_patch/Findpioasm.cmake
```diff
@@ -0,0 +1,58 @@
...
+                INSTALL_DIR ${PIOASM_INSTALL_DIR}
+                CMAKE_ARGS
+                    "--no-warn-unused-cli"
+                    "-DCMAKE_MAKE_PROGRAM:FILEPATH=${CMAKE_MAKE_PROGRAM}"
+                    "-DPIOASM_FLAT_INSTALL=1"
+                    "-DCMAKE_INSTALL_PREFIX=${PIOASM_INSTALL_DIR}"
+                    "-DCMAKE_RULE_MESSAGES=OFF" # quieten the build
+                    "-DCMAKE_INSTALL_MESSAGE=NEVER" # quieten the install
+                    # Toolchain workaround follows
+                    "-DCMAKE_CXX_FLAGS=-include cstdint"
```
projectgus, line 41: This line is the only non-comment change in this file, compared to upstream.

dpgeorge: I did not know about `-include` option to gcc!

dpgeorge (approved):
Tested locally with host GCC 15.1.1, building RPI_PICO_W.  It now works.

ports/rp2/tools_patch/Findpioasm.cmake
projectgus: Supported by clang as well, thankfully.

dpgeorge: Great!

PR #16454, opened 2024-12-19
Alarm pool sleep changes

peterharperuk:
rp2350 generates a sev when using spin locks, which can prevent the device sleeping, see https://github.com/raspberrypi/pico-sdk/issues/1812 for more background.

The pico-sdk alarm pool handles this.

However Micropython moved away from using the alarm pool due to issues like this...
Fixed in sdk 2.0.0: https://github.com/raspberrypi/pico-sdk/issues/1552
Fixed in sdk 2.1.0: https://github.com/raspberrypi/pico-sdk/issues/1953
Fixed in develop: https://github.com/raspberrypi/pico-sdk/pull/2127

This change puts back the use of the alarm pool in mp_wfe_or_timeout to hopefully fix sleep issues.

PR #17272, opened 2025-05-09
rp2: Disable the LWIP tick timer when not needed.

### LINKED ISSUE #16181: lightsleep on Pico W regression for 1.24.0

### LINKED ISSUE #17204: lightsleep(n) returns immediately on pico-w or pico with network imported

pico and pico-w

### MicroPython version

MicroPython v1.25.0 on 2025-04-15; Raspberry Pi Pico W with RP2040

### Reproduction

copy paste this code:
```
# light_sleep_test.py
from machine import lightsleep
from time import sleep_ms, ticks_ms, ticks_diff

def test(n):
    for i in range(n):
        t0 = ticks_ms()
        sleep_ms(50)
        lightsleep(5000)
        sleep_ms(50)
        print(i, ticks_diff(ticks_ms(), t0))

test(10)
```
If the test works it should print 
0 5101
1 5101
2 5101
3 5101
4 5101
5 5101
6 5101
7 5100
8 5101
9 5100

Instead it prints:
0 142
1 127
2 127
3 128
4 129
5 127
6 128
7 128
8 128
9 128

### Expected behaviour

If the test works it should print 
0 5101
1 5101
2 5101
3 5101
4 5101
5 5101
6 5101
7 5100
8 5101
9 5100

### Observed behaviour

Instead it prints:
0 142
1 127
2 127
3 128
4 129
5 127
6 128
7 128
8 128
9 128

### Additional Information

See [discussion here](https://github.com/orgs/micropython/discussions/17185#discussioncomment-12951488)

### Code of Conduct

Yes, I agree

projectgus:
### Summary

Closes #16181. Prevents lightsleep being woken up every 64ms to service LWIP timers, when:

1. No netif is up, and
2. No TCP sockets are active

The TCP socket check may not be strictly necessary, but without ticking the tcp timer they won't ever time out by themselves.

It means that it's possible to do:

```py
wlan.disconnect()
machine.lightsleep(5000)
wlan.connect(...)
```

... and get the full sleep time, but also possible to do 

```py
while some_condition():
    machine.lightsleep(5000)
```

... and Python will wake up every 64ms if Wi-Fi is active, but power consumption is still lower than it would be otherwise.

(Not massively lower, on RPI_PICO_W I measure 53mA idle at REPL with Wi-Fi associated, and 40mA in a `while True: machine.lightsleep(1000)` loop. There is probably potential to make this better by having lightsleep put the CYW43 driver into a lower power mode.)

### Testing

**Important: This PR has only been tested cherry-picked onto the branch for #16454. It is marked as a draft until that PR is merged and it can be rebased.**

* Ran `multi_net` and `multi_wlan` tests on RPI_PICO_W and RPI_PICO2_W. Same failures as reported against master.

* There is some risk that some part of LWIP doesn't like being starved of timeouts for a long period after a network interface goes down. However, it looks like it will be OK - the main thing is TCP/IP socket timeouts and LWIP will keep ticking until these time out.
* Could make LWIP timer not a wake-up source, but this is quite difficult and does prevent using light sleep effectively with Wi-Fi.
* Could make LWIP timer wake-ups not return from lightsleep (i.e. the hardware wakes up, services LWIP, then goes back to sleep), but this is also quite fiddly to do (need to check the wake source each time).
* Semi-related: there is also a LWIP function `sys_timeouts_sleeptime()` which could be used to have the LWIP timer run at an exact time instead of at a constant tick rate. However, a number of LWIP "cyclic" timers run every 100ms right now so this change seems like it would be of limited benefit, it would add some code size, *and* the lwip tick callback is also used to poll the wiznet5k driver.

### DIFF
--- a/ports/rp2/mpnetworkport.c
+++ b/ports/rp2/mpnetworkport.c
@@ -31,6 +31,7 @@
 #if MICROPY_PY_LWIP
 
 #include "shared/runtime/softtimer.h"
+#include "lwip/netif.h"
 #include "lwip/timeouts.h"
 
 // Poll lwIP every 64ms by default
@@ -38,6 +39,9 @@
 
 // Soft timer for running lwIP in the background.
 static soft_timer_entry_t mp_network_soft_timer;
+
+// Callback for change of netif state
+NETIF_DECLARE_EXT_CALLBACK(netif_callback)
 
 #if MICROPY_PY_NETWORK_CYW43
 #include "lib/cyw43-driver/src/cyw43.h"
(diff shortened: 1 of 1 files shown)


madozu:
I did a lot of testing with the current master branch and pull requests #16454 and #17272 applied.

All the "early wakeup" scenarios I was experiencing when using machine.lightsleep() are solved --> **Great news, many thanks** to @projectgus for providing a good solution 😄 

So far, I did not encounter any negative side effects.

ports/rp2/mpnetworkport.c
```diff
@@ -147,7 +170,16 @@ void mod_network_lwip_init(void) {
...
-    soft_timer_reinsert(&mp_network_soft_timer, LWIP_TICK_RATE_MS);
+    if (netif_callback.callback_fn == NULL) {
+        netif_add_ext_callback(&netif_callback, mp_network_netif_status_cb);
+    }
+}
+
+static void mp_network_netif_status_cb(struct netif *netif, netif_nsc_reason_t reason, const netif_ext_callback_args_t *args) {
+    // Start the network soft timer any time an interface comes up
+    if (reason == LWIP_NSC_LINK_CHANGED && args->link_changed.state) {
+        soft_timer_reinsert(&mp_network_soft_timer, LWIP_TICK_RATE_MS);
```
dpgeorge, line 181: Maybe this needs a check `... && mp_network_soft_timer.mode == SOFT_TIMER_MODE_ONE_SHOT` so that it's only reinserted if it was disabled.

That would prevent pathological cases like an interface going up/down continuously and constantly pushing the timer out further and further.

projectgus: Check added. Also needed to switch the `mode` here to avoid the pathological case (otherwise too-fast interface flapping can still starve the timeout).

projectgus:
Rebased, addressed review comments, re-ran `multi_net` and `multi_wifi` tests on RPI_PICO_W & RPI_PICO2_W with same results.

* On my local setup `multi_net/udp_data.py` fails intermittently to packet loss - but I have a mediocre wifi signal here.
*  `multi_net/tls_dtls_server_client.py` fails when the rp2 is instance 1.
* `multi_wlan` I realised has a bug when STA is rp2 and a wifi connection is already active. Will submit separate PR to fix.

dpgeorge (approved):
Tested on RPI_PICO_W and RPI_PICO2_W, running all possible tests:
- `tests/ports/rp2/rp2_lightsleep_thread.py` fails on both of them, but that's expected (and fails also on master)
- `tests/multi_net/udp_data_multi.py` fails when one of the boards is instance 0, and also fails on master.  I was sure this was working when I wrote the test so maybe it's down to local WLAN variability.

All other tests pass for me.

PR #17287, opened 2025-05-12
esp32/mpthreadport: Fix double delete of tasks on soft reset.

dpgeorge:
### Summary

Python threads (created via the `_thread` module) are backed by a FreeRTOS task.  Managing the deletion of the task can be tricky, and there are currently some bugs with this in the esp32 port.

The actual crash I saw was in FreeRTOS' `uxListRemove()`, and that's because of two calls to `vTaskDelete()` for the same task: one in `freertos_entry()` when the task ran to completion, and the other in `mp_thread_deinit()`.  The latter tried to delete the task a second time because it was still in the linked list, because `vTaskPreDeletionHook()` had not yet been called.  And the reason `vTaskPreDeletionHook()` was yet to be called is because the FreeRTOS idle task was starved.

This PR attempts to fix that.

There are two things done by this fix:
- make sure `vTaskDelete()` is only called once for each task, either when the Python thread finishes, or in `mp_thread_deinit()` (otherwise a double free leads to memory corruption within FreeRTOS)
- make `mp_thread_deinit()` to wait for all remaining tasks to have their `vTaskPreDeletionHook()` called (otherwise `vTaskPreDeletionHook` may be called after the soft reset, when the MP GC heap no longer contains the valid `mp_thread_t` structs)

### Testing

The bugs here were found by running the standard test suite over and over again on an ESP32S2, with ESP32_GENERIC_S2 firmware.  But it was very difficult to reproduce.   The following was needed to see a crash:
- ESP32S2, because it's single core (on dual core, task cleanup seems to happen much faster)
- run a test that outputs lots of data, that seems to saturate the USB CDC task and starve the idle task
- then run a test that creates a thread
- then run another test so soft reset occurs
- the timing of the host PC processing USB CDC data also has an effect!  eg if the host PC is very busy then the bug may show up more easily or not
- IDF 5.4.1 showed the bug much more easily than IDF 5.2.2

I found the following test command allowed the bug to manifest:
```
$ cd tests
$ while true; do ./run-tests.py -t a0 extmod/framebuf16.py extmod/framebuf_ellipse.py extmod/select_poll_eintr.py || break; sleep 1s; done
```
With the fix in this PR the bug seems to be gone, I can no longer reproduce it.  Tested on ESP32 and ESP32S2, with IDF 5.4.1.


projectgus (commented):
This fix looks fine, I like the automatic state transitions as a way to model it.

It did make me wonder why the pre-delete hook is needed in the first place.

From what I can tell, it dates from a `vPortCleanUpTCB()` callback in the original ESP32 support commit bc08c884a2e7d1d3a4476c224f230b90ee2bf1e3. However at that time it looks like Task resources - including stacks - were allocated from the Python heap, which is why they needed to be cleaned up out of band.

Now all the pre-delete hook does is remove the thread from the linked list, and AFAIK there's no other FreeRTOS task callback. Which makes me wonder if we could simplify this back to:

Executing task ends:

1. Lock thread mutex
2. Remove self from linked list
3. Unlock thread mutex
4. vTaskDelete(NULL)

... and the soft reset loop:

1. Lock thread mutex
2. Iterate task list. For any task that's not the main task call vTaskDelete() immediately and remove from the list.
3. Unlock thread mutex

Calling `vTaskDelete(tid)` while holding the thread mutex should be safe from deadlock as nothing should run directly as a result of that call any more, and calling `vTaskDelete(NULL)` can't corrupt any entry in the linked list as the task has already been removed from the list by that time...

projectgus:
> Calling vTaskDelete(tid) while holding the thread mutex should be safe from deadlock

If there's potential for a deadlock here then another way around it would be to take the linked list nodes off the shared list while holding the mutex, put them on a temporary list on the stack, and then iterate that list after releasing the mutex. But I think no longer necessary if the hook goes away...

dpgeorge:
> It did make me wonder why the pre-delete hook is needed in the first place.

Yes, I also wondered this.  But wasn't brave enough to remove it!

I've now changed this PR to do what you suggest.  The new run-state enum is not really needed anymore, but I left it there because it's a minor improvement to code readability.

So now this PR essentially makes the following changes:
- a task removes itself from the linked list, just before it deletes itself
- on soft reset, all tasks are deleted and removed from the linked list in one swoop (while the mutex is held)

For this to work, I assume that once `vTaskDelete(task_id)` is called, that task will never be rescheduled/run.  I think that's a valid assumption to make?  (If it's not valid then there'll be a race.)

Also, there are no assumptions about how tasks behave when they delete themselves, I don't actually know how FreeRTOS handles this case (does the task resume until the next task switch??), but it doesn't really matter for the correctness of the code here.

ports/esp32/mpthreadport.c
```diff
@@ -221,32 +212,17 @@ void mp_thread_mutex_unlock(mp_thread_mutex_t *mutex) {
...
+        if (th != &thread_entry0) {
+            vTaskDelete(th->id);
         }
     }
-}
-
-#else
-
-void vTaskPreDeletionHook(void *tcb) {
+    thread = &thread_entry0;
```
projectgus, line 224: I think `thread_entry0.next == NULL` is an invariant (because new threads always pushed to the head of the list), but it might be worth either asserting that condition here or setting `thread->next = NULL` in case something changes in the future.

projectgus (approved):
This looks good to me, fewer moving parts!

ports/esp32/mpthreadport.c
```diff
@@ -116,11 +120,24 @@ void mp_thread_start(void) {
...
+            *th = (*th)->next;
+        }
+    }
+    mp_thread_mutex_unlock(&thread_mutex);
+
+    // Delete this FreeRTOS task.
     vTaskDelete(NULL);
-    for (;;) {;
+    for (;;) {
+        vTaskSuspend(NULL);
```
projectgus, line 140: Should be safe to remove this infinite loop.

ports/esp32/mpthreadport.c
dpgeorge: Added an assertion.

ports/esp32/mpthreadport.c
dpgeorge: Now removed, and comment updated.

dpgeorge:
Tested again on ESP32_GENERIC_C3.  No failures detected.

Issue #17241, opened 2025-05-02
RP2350B Core Board: says GPIO pin > 31 declared mode = OUTPUT  is shown as INPUT


Flipje1955:
### Port, board and/or hardware

RP2350B Core WeAct Studio

### MicroPython version

Using MicroPython v1.25.0 on 2025-04-15; WeAct Studio RP2350B Core with RP2350:

### Reproduction

'''
pin = Pin(31, Pin.OUT)
pin.on()
print(pin, pin.value())
pin.off()
print(pin, pin.value())

pin = Pin(32, Pin.OUT)
pin.on()
print(pin, pin.value())
pin.off()
print(pin, pin.value())
'''

### Expected behaviour

Expected all pins defined as OUTPUT to be shown as such. 

### Observed behaviour

Output:

MPY: soft reboot
Pin(GPIO31, mode=OUT) 1
Pin(GPIO31, mode=OUT) 0
Pin(GPIO32, mode=IN) 1
Pin(GPIO32, mode=IN) 0

This applies to all pins #32  up to # 47. 

nb. Same behaviour on [v1.26.0-preview.59.g79abdad9e (2025-05-02) .uf2](https://micropython.org/resources/firmware/WEACTSTUDIO_RP2350B_CORE-20250502-v1.26.0-preview.59.g79abdad9e.uf2)

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

Gadgetoid:
Looks like something I overlooked when porting to the RP2350. I think this relates to the macros defined in the top of `machine_pin.c` which directly access registers:

https://github.com/micropython/micropython/blob/69993daa5c2eae6055f0b3b2330e95e78d6e3738/ports/rp2/machine_pin.c#L49-L53

Specifically `GPIO_IS_OUT` which ignores the existence of `gpio_hi_oe` .

Should be a quick fix.

dpgeorge:
Fixed by 28c8fff6d8f920eb6199cc5db59da1baf1b00843

PR #17288, opened 2025-05-12
rp2/machine_pin: Replace macros with Pico SDK functions.

Gadgetoid:
Replace custom macros with Pico SDK functions, enabling support for RP2350B variant chips with > 32 GPIOs.

I had missed this when porting to RP2350/RP2350B, since it only affects the string representation of a pin and has no functional implications (at least where users aren't inspecting that text to detect pin state).

### Summary

Fixes an issue raised in  #17241 where `machine_pin_print` was not reporting the correct pin direction for GPIOs >= 32, since the output states are stored in `gpio_oe_hi`

### Testing

A basic test on an RP2350B board (Pimoroni Pico Plus 2) to make sure `machine_pin_print` produces the correct string:

```python
>>> machine.Pin(31, machine.Pin.OUT)
Pin(GPIO31, mode=OUT)
>>> machine.Pin(33, machine.Pin.OUT)
Pin(GPIO33, mode=OUT)
>>> machine.Pin(33, machine.Pin.OUT, machine.Pin.PULL_UP)
Pin(GPIO33, mode=OUT, pull=PULL_UP)
>>> machine.Pin(33, machine.Pin.OUT, machine.Pin.PULL_DOWN)
Pin(GPIO33, mode=OUT, pull=PULL_DOWN)
```

And for funcsel:

```python
>>> a = machine.Pin(0)
>>> b = machine.Pin(1)
>>> i2c = machine.I2C(0, sda=a, scl=b)
>>> print(a)
Pin(GPIO0, mode=ALT, pull=PULL_UP, alt=I2C)
>>> print(b)
Pin(GPIO1, mode=ALT, pull=PULL_UP, alt=I2C)
```

And for open-drain `machine_pin_toggle`, observe the output enable bit being toggled:

```python
>>> od = machine.Pin(3, machine.Pin.OPEN_DRAIN)
>>> print(od)
Pin(GPIO3, mode=OPEN_DRAIN)
>>> od.toggle()
>>> f"{machine.mem32[0xd0000030] & 0xffffffff:32b}"
'10000000000000000000000000000000'
>>> od.toggle()
>>> f"{machine.mem32[0xd0000030] & 0xffffffff:32b}"
'10000000000000000000000000001000'
```

This is a low-impact change affecting (fixing) only the output of `machine_pin_print`. It might actually result in a small size increase (🤞) but is a better choice than trying to reimplement SDK functions and dealing with future repercussions.

Edit: I lied. This affects `machine_pin_toggle` too!

### DIFF
--- a/ports/rp2/machine_pin.c
+++ b/ports/rp2/machine_pin.c
@@ -46,12 +46,6 @@
 
 #define GPIO_IRQ_ALL (0xf)
 
-// Macros to access the state of the hardware.
-#define GPIO_GET_FUNCSEL(id) ((iobank0_hw->io[(id)].ctrl & IO_BANK0_GPIO0_CTRL_FUNCSEL_BITS) >> IO_BANK0_GPIO0_CTRL_FUNCSEL_LSB)
-#define GPIO_IS_OUT(id) (sio_hw->gpio_oe & (1 << (id)))
-#define GPIO_IS_PULL_UP(id) (padsbank0_hw->io[(id)] & PADS_BANK0_GPIO0_PUE_BITS)
-#define GPIO_IS_PULL_DOWN(id) (padsbank0_hw->io[(id)] & PADS_BANK0_GPIO0_PDE_BITS)
-
 // Open drain behaviour is simulated.
 #define GPIO_IS_OPEN_DRAIN(id) (machine_pin_open_drain_mask & (1 << (id)))
 
@@ -198,13 +192,13 @@ const machine_pin_obj_t *machine_pin_find(mp_obj_t pin) {
 
 static void machine_pin_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     machine_pin_obj_t *self = self_in;
-    uint funcsel = GPIO_GET_FUNCSEL(self->id);
+    uint funcsel = gpio_get_function(self->id);
     qstr mode_qst;
     if (!is_ext_pin(self)) {
         if (funcsel == GPIO_FUNC_SIO) {
             if (GPIO_IS_OPEN_DRAIN(self->id)) {
                 mode_qst = MP_QSTR_OPEN_DRAIN;
-            } else if (GPIO_IS_OUT(self->id)) {
+            } else if (gpio_is_dir_out(self->id)) {
                 mode_qst = MP_QSTR_OUT;
             } else {
                 mode_qst = MP_QSTR_IN;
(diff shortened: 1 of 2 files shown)


Gadgetoid:
I think I need to run some more tests on open-drain toggling with pins >= 32 since I'm not seeing the register bits twiddled that I would expect.

Gadgetoid:
Fixing the uint64_t mask handling for RP2350B support adds 80 bytes to the binary size on regular RP2.

We can probably set `machine_pin_open_drain_mask` to uint32_t on non-B variant builds and replace `1ULL` with a define that defaults to `1U` on non-B builds without too much hassle.

Gadgetoid:
Commit 1: +16 bytes
Commit 2: +96 bytes
Commit 3: -132 bytes

Did a bit better than free!

ports/rp2/machine_pin.c
```diff
@@ -89,7 +83,13 @@ static const mp_irq_methods_t machine_pin_irq_methods;
 static const int num_intr_regs = sizeof(iobank0_hw->intr) / sizeof(iobank0_hw->intr[0]);
 
 // Mask with "1" indicating that the corresponding pin is in simulated open-drain mode.
+#if NUM_BANK0_GPIOS > 32
 uint64_t machine_pin_open_drain_mask;
+#define MACHINE_PIN_OD_BIT 1ULL
```
dpgeorge, line 88: I don't think these defines need to be redefined here, they should be picked up from `mphalport.h`.

Gadgetoid: Fixed, thanks.

I do wonder if this would perhaps be better as a much more explicit set of macros:

```c
// Open drain behaviour is simulated.
#if NUM_BANK0_GPIOS > 32
extern uint64_t machine_pin_open_drain_mask;
#define MACHINE_PIN_OD_MASK(pin) (1ULL << pin)
#define GPIO_IS_OPEN_DRAIN(id) (machine_pin_open_drain_mask & (1ULL << id))
#define GPIO_SET_OPEN_DRAIN(id) (machine_pin_open_drain_mask |= (1ULL << id))
#define GPIO_CLR_OPEN_DRAIN(id) (machine_pin_open_drain_mask &= ~(1ULL << id))
#else
extern uint32_t machine_pin_open_drain_mask;
#define MACHINE_PIN_OD_MASK(pin) (1U << pin)
#define GPIO_IS_OPEN_DRAIN(id) (machine_pin_open_drain_mask & (1U << id))
#define GPIO_SET_OPEN_DRAIN(id) (machine_pin_open_drain_mask |= (1U << id))
#define GPIO_CLR_OPEN_DRAIN(id) (machine_pin_open_drain_mask &= ~(1U << id))
#endif
```

For instance this turns:

```c
static inline void mp_hal_pin_input(mp_hal_pin_obj_t pin) {
    gpio_set_dir(pin, GPIO_IN);
    machine_pin_open_drain_mask &= ~(1ULL << pin);
    gpio_set_function(pin, GPIO_FUNC_SIO);
}
```

into:

```c
static inline void mp_hal_pin_input(mp_hal_pin_obj_t pin) {
    gpio_set_dir(pin, GPIO_IN);
    GPIO_CLR_OPEN_DRAIN(pin);
    gpio_set_function(pin, GPIO_FUNC_SIO);
}
```

dpgeorge (approved):
Looks good now, a nice refactor!

PR #17289, opened 2025-05-12
unix/modsocket: Accept a host+port array for socket.connect.

agatti:
### Summary

This PR lets `socket.connect` accept an array containing a hostname string and a port number as the address to connect to, if the socket family is either `AF_INET` or `AF_INET6`.  The existing behaviour of accepting a serialised sockaddr structure as the connection address is still present, maintaining compatibility with existing code.

This brings the behaviour of socket.connect almost in line with the other embedded ports' versions that use LWIP, and somewhat with CPython as well.  The differences are as follows:

* Addresses in numeric form are not supported: this is not supported in LWIP either but CPython can handle them
* Connection argument can be either a two-elements list or tuple: this matches LWIP's implementation, but CPython can only accept a tuple for `AF_INET`/`AF_INET6`
* A regular hostname can be passed as the address: this is not supported in LWIP (dotted quads only, see `netutils_parse_inet_addr` in `shared/netutils/netutils.c` as invoked by `extmod/modsocket.c`), but is supported in CPython.

Hostname resolution is performed internally via a call to `socket.getaddrinfo`, and in case of multiple compatible entries mapped to a single hostname the first one returned by `getaddrinfo(2)` will be used.

### Testing

This has been tested manually as such (numeric addresses are what www.google.com resolves to me from here):

```python
import socket

data = {
    socket.AF_INET: ["www.google.com", "64.233.184.99"],
    socket.AF_INET6: ["www.google.com", "2a00:1450:400c:c0b::69"],
}

for k, v in data.items():
    sf = (k,)
    for addr in v:
        s = socket.socket(*sf)
        s.connect((addr, 80))
        s.write("GET /\n\n")
        if len(s.recv(100)) != 100:
            raise Exception
        print(sf, addr, ": SUCCESS [STRING]")
        s.close()
        s = socket.socket(*sf)
        a = socket.getaddrinfo(addr, 80, k)[0][-1]
        s.connect(a)
        s.write("GET /\n\n")
        if len(s.recv(100)) != 100:
            raise Exception
        print(sf, addr, ": SUCCESS [SOCKADDR]")
        s.close()
```

with the output:

```
(2,) www.google.com : SUCCESS [STRING]
(2,) www.google.com : SUCCESS [SOCKADDR]
(2,) 64.233.184.99 : SUCCESS [STRING]
(2,) 64.233.184.99 : SUCCESS [SOCKADDR]
(10,) www.google.com : SUCCESS [STRING]
(10,) www.google.com : SUCCESS [SOCKADDR]
(10,) 2a00:1450:400c:c0b::69 : SUCCESS [STRING]
(10,) 2a00:1450:400c:c0b::69 : SUCCESS [SOCKADDR]
```

~~I see `extmod/ssl_basic.py` uses `test.example.com` for its tests - if a test file is required I may use that host myself as well.~~

Edit: a test file with ipv4/ipv6 connection tests using both host+port pair and a pre-resolved sockaddr buffer have been added.

There is a slight footprint increase for the Unix port but, considering the average computing power and storage facilities of the average environment the Unix port runs in, the quality of life improvements, and increased compatibility should be enough benefits to make these code changes useful.

### DIFF
--- a/ports/unix/modsocket.c
+++ b/ports/unix/modsocket.c
@@ -37,6 +37,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/un.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <netdb.h>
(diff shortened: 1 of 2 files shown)


agatti:
Are IPv6 connections blocked in the CI environment?  Unless I'm missing something the IPv6 tests should succeed (they do from my LAN at least).

agatti:
Interesting, `socket.send(str)` works on Unix - although it shouldn't.  Wonder if that behaviour should be deprecated for v2.

agatti:
Thanks for the explanation!  I've never really used ffi nor micropython-lib when dealing with the Unix port, I usually develop directly on device myself and use Unix just to validate some things that would take longer to test elsewhere.

From what I've seen (and I hope I didn't miss anything) there are only two ports that have sockets support and can't natively handle a host+port tuple in `socket.connect`: Unix and zephyr.  Some ports that can handle tuples are limited to dotted quads, whilst others call an internal implementation of getaddrinfo.

So, I wonder whether adding at least dotted quad support to Zephyr (and maybe to Unix without ffi to make it easier for interactive use) would make sense at this point, to actually make things more compatible across ports for simpler cases.

If that's not something that can help, no worries, I'll close the PR :)  On the other hand, maybe the sockaddr-only unit test cases can still be saved.

PR #17290, opened 2025-05-12
/esp32/boards: Add LOLIN_S3_PRO board definition.

### LINKED ISSUE #17217: Request: Port for Lolin ESP32-S3 Pro (16MB flash, 8MB SPRAM)
### Description

I notice a few ports for the ESP32 family, including a couple of lower-spec Lolin boards, plus generic ESP32-S3 ports. But these don't accommodate boards like the Lolin ESP32-S3 Pro with its 16MB Flash and 8MB PSRAM.

Is there any possibility of ports in the near future, for this board, and other newer/higher models of ESP32?

### Code Size

_No response_

### Implementation

I hope the MicroPython maintainers or community will implement this feature

### Code of Conduct

Yes, I agree

mattytrentini:
### Summary

Added a board definition for the [Wemos Lolin S3 Pro](https://www.wemos.cc/en/latest/s3/s3_pro.html).

### Testing

I deployed the firmware to my S3 Pro, it appeared to work fine. Tested the filesystem (since a key feature of this board is the 16MB flash) and the GRB (_not_ RGB) LED:

```bash
MicroPython v1.25.0-preview.575.g076e07197 on 2025-05-12; LOLIN_S3_PRO with ESP32S3
>>>
paste mode; Ctrl-C to cancel, Ctrl-D to finish
=== import os
=== fs_stat = os.statvfs('/')
=== fs_size = fs_stat[0] * fs_stat[2]
=== fs_free = fs_stat[0] * fs_stat[3]
=== print(f"File System Size {fs_size:,} - Free Space {fs_free:,}")
===
File System Size 14,680,064 - Free Space 14,667,776
>>> from machine import Pin
>>> from neopixel import NeoPixel
>>> neo = NeoPixel(Pin.board.RGB_LED, 1)
>>> NeoPixel.ORDER = (0, 1, 2, 3)  # GRB rather than RGB
>>> neo[0] = (40, 0, 0)
>>> neo.write() # Red
>>> neo[0] = (0, 40, 0)
>>> neo.write() # Green
>>> neo[0] = (0, 0, 40)
>>> neo.write() # Blue
```

A user could define their own custom board. We could also extend ESP32_GENERIC_S3 to support 16MB models (which may be wroth doing as well).

See #17217 for a request for this board.


dpgeorge:
> We could also extend ESP32_GENERIC_S3 to support 16MB models

This is now done, see commit 6201e77999b3614518abc4b21773e735d9b0b0ee

Given that, does it still make sense to add this Lolin board definition?  If so, please update this PR so it works with the new auto-sizing feature (basically, you don't need to specify the flash size or partitions csv).

mattytrentini:
>Given that, does it still make sense to add this Lolin board definition?

It doesn't add much on top of ESP32_GENERIC_S3. Really the only benefit is that it's more obvious which firmware is appropriate for Lolin S3 Pro users. Maybe we need to list 'community tested' boards for the ESP32_GENERIC* boards?

In any case, I'm ok with closing this and directing Lolin S3 Pro users to the generic alternative.

dpgeorge:
> Maybe we need to list 'community tested' boards for the ESP32_GENERIC* boards?

Yes, I think that's a good idea.  Can easily just list them in the `board.md` for those generic boards.

PR #17293, opened 2025-05-13
py/asmthumb: Implement long jumps on Thumb/armv6m architecture.

dpgeorge:
### Summary

Implement long jumps (greater than +/- 2k code distance) for armv6m targets, eg RPI_PICO.

With this change, all tests (except thread tests) now pass on RPI_PICO when using the native emitter:

    (plug in RPI_PICO)
    $ cd tests
    $ ./run-tests.py -t a0 --via-mpy --emit native

### Testing

Tested on an RPI_PICO as above.  Prior to this PR 8 tests would fail due to large jumps not being implemented.  Now they pass.

All forwards jumps now take 8 or 10 bytes, whereas previously they took only 2 bytes.  That's just because it's not possible to know in advance how long the jump is, and the emitter must commit to the size of the instruction in early passes (one day that could be improved by adding more assembler passes).

The jump can now clobber register `r1`, which it uses as a temporary to calculate the long jump.  Clobbering this register is OK, it's anyway used as a temporary in the native emitter and is never needed across jumps (but `r0` is because it's used to return a value from the function, so that can't be used here as a temporary).

### DIFF
--- a/py/asmthumb.c
+++ b/py/asmthumb.c
@@ -499,6 +499,7 @@ void asm_thumb_store_reg_reg_offset(asm_thumb_t *as, uint reg_src, uint reg_base
 #define OP_BW_HI(byte_offset) (0xf000 | (((byte_offset) >> 12) & 0x07ff))
 #define OP_BW_LO(byte_offset) (0xb800 | (((byte_offset) >> 1) & 0x07ff))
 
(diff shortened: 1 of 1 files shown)


dpgeorge:
@agatti FYI

PR #17295, opened 2025-05-13
extmod/vfs_lfsx: Fix errno value raised from chdir.

dpgeorge:
### Summary

Littlefs was raising a negative errno if chdir failed, when it should be a positive errno.

### Testing

Tests have been updated to cover this case (and other cases which were already correct).

### DIFF
--- a/tests/extmod/vfs_lfs_error.py
+++ b/tests/extmod/vfs_lfs_error.py
@@ -1,7 +1,7 @@
 # Test for VfsLittle using a RAM device, testing error handling
 
 try:
-    import vfs
+    import errno, vfs
 
     vfs.VfsLfs1
     vfs.VfsLfs2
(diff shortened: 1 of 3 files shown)


tests/extmod/vfs_lfs_error.py
```diff
@@ -1,7 +1,7 @@
 # Test for VfsLittle using a RAM device, testing error handling
 
 try:
-    import vfs
+    import errno, vfs
```
projectgus, line 4: Is it necessary to add SKIP logic for ports without `MICROPY_PY_ERRNO` set (defaults to `MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_EXTRA_FEATURES`), or is any port that can run extmod tests likely to have this set already?

projectgus (approved):
~~One question about the test, but~~ the fix and expanded test covertage looks good to me

tests/extmod/vfs_lfs_error.py
projectgus: (`vfs` defaults to `MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES && MICROPY_VFS`, so there is a potential window there but I don't know if there's actually many boards configured in this window).

projectgus: Oh my mistake, I misread the diff summary and didn't notice there was already SKIP logic here. :+1:

dpgeorge: This test actually runs only on the unix port because that's the only port with both lfs1 and lfs2 enabled.... might be good to refactor this test as a unittest (one day) so it can run on ports with lfs1 and/or lfs2.

But regardless of that, the skip logic should work for any configuration, as you noticed.

PR #17301, opened 2025-05-14
docs: Improve documentation for writing and running tests

andrewleech:
### Summary

I wanted to use some LLM assitance to create new unit tests. To aid in this process, it helps to have comprehensive docs to inform the LLM's how best to write and run unit tests, so I started with that.

These updates were mostly written with Google's gemini-2.5-pro-exp-03-25 though some final review and cleanup was done with Anthropic Claude 3.7

The updated documentation provides:
- More detailed explanation of the test directory structure and key runner scripts.
- Clearer instructions on writing simple tests, including the use of print() and .exp files.
- Expanded coverage of run-tests.py options for targeting specific devices, filtering tests, and handling failures.
- An overview of how run-tests.py works internally (feature detection, skipping, execution, comparison).
- Guidance on writing advanced tests (skipping, unittest, special handling, float/endian considerations).
- Introduction to multi-instance testing with run-multitests.py, including how to write tests with instanceN() functions and use the `multitest` helper for coordination.
- Examples of running multi-instance tests with different instance types.

### Testing

Just manual review of the content.


agatti:
Not sure if it's important enough to be featured in that document, but maybe it'd be useful to document the special test file name prefixes/suffixes recognised by `tests/run-tests.py` that act as rudimentary feature checks to skip certain tests?

For example, test files whose name starts with `native_` or `viper_` will be automatically skipped if `@micropython.native` or `@micropython.viper` support was not detected in the target interpreter, and so on.  There's a dozen or so of those tags so maybe it's not worth documenting them all, but again, the LLM may sort all that stuff out by itself :)

andrewleech:
> special test file name prefixes/suffixes

Yeah I do think it'd be helpful to have these itemised somewhere as it's certainly not abvious for new developers! 

I'm not too sure about how best to divide what does in the rst docs and what goes in the readme here... I kinda think the readme is perhaps better for "internals" things like that which are maybe of more interest to low level developers, with the rst docs more focussing on "end users" interested in running tests?

agatti:
I personally use the README files as a quickstart guide myself, always favouring explicit documentation if it's available.  It used to be like that back then, when dinosaurs still roamed freely and whatnot, so maybe somebody from a younger demographics should weigh in on this.

Considering the users/developers split, I wonder if this possible TOC for three (or two) documents - names are generic on purpose - can be a starting point for a discussion on this:

* "Quickstart"/"End users"
	- Standard tests
		- How to run tests (simplified version, no need to bring emitters and mpy in, f.ex.)
		- How a successful run should look like
		- Where to look for failure info to send back to a developer
	- Performance benchmarks
		- How to run benchmarks (simplified version)
		- How to read output and compare with previous runs
* "Developers"
	- How to run tests (what's left to document)
	- Standard tests (almost everything that's in `writingtests.rst` for standard tests, maybe skip more niche sub-topics)
	- Multi-instance tests
	- SSL tests
	- Performance benchmarks (what's left to document)
* "Internals" (fold this into "Developers" if needed)
	- How `run-tests.py` enumerates tests and how it decides which ones to skip
	- Natmod tests
	- Internal benchmarks
	- Everything else that's not in "Developers" ?

I can help out on the natmod tests bit if needed, as long as somebody else fixes the grammar :)

Josverl:
for context : I just got an inflatable Dino for my birthday.
There is no need to settle on just a single topilogy 
With Sphinx & MyST it is simple to pull in the Readme.MD and other markdown docs into the online documentation.  
That allows the information  to be found and accessed va different ways and personal preferences , without increasing the effort to maintain it.
( possibly that is also a way to pull in some of the information in the Wiki , which is rather hard to search )

agatti:
FYI, maybe you may want to update the documentation to also cover the extra argument introduced in 6fd7422a8e5b6c8952f7b67713f0551992d247cc.  This is probably something that should be put into the "advanced" section :)

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,355 @@
...
+1. **Feature Detection**: Runs scripts in ``feature_check/`` to determine:
+   
+   - Architecture and platform
+   - Available modules and features
+   - Code emitter support (native, viper)
+   - Float precision
+   - Endianness
+
+2. **Automatic Skipping**: Tests are skipped based on:
+
```
projectgus: Suggest documenting the explicit SKIP output first, as that's the preferred method. Something like:

---

Where possible, a test should determine at runtime if it should be skipped. If so, it prints a single line `SKIP` before existing with `raise SystemExit()`. The test runner parses this output and skips the test.

This isn't always possible, particularly if the test relies on optional MicroPython features that prevent parsing the .py file (i.e. importing the file raises SyntaxError before the skip logic can run.) In these cases, there is additional logic for skipping tests:

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,355 @@
...
+--------------
+
+1. **Test Isolation**: Each test should be independent
+2. **Deterministic Output**: Avoid timing-dependent output
+3. **Resource Awareness**: Consider memory constraints on embedded targets
+4. **Clear Failure Messages**: Make failures easy to diagnose
+5. **Documentation**: Comment complex test logic
+6. **Cross-platform**: Test on multiple architectures when possible
+
+Debugging Test Failures
```
projectgus: Maybe this section could be merged into test_quickstart.rst, as it's relevant for people running tests as well?

projectgus (commented):
Hi @andrewleech,

This is really good, and I think overdue! (I noted that it's inspired by improving the docs for bots not for humans, but either way it's good to have better docs!)

There's some newer features and changes that would need to be folded back in here (like `target_wiring.py` and I think some of the skip checks have been simplified out since this was written), but it's looking very solid.

It's taken a while for me to notice it (sorry), are you still interested in updating it?

Josverl:
>  I haven't got the rst docs importing the readme, not sure how much adjusting would be needed in the sphinx stuff to support that?

_Sorry: I forgot to reply earlier_ 

### .rst in .rst  using [include](https://docutils.sourceforge.io/docs/ref/rst/directives.html#include)
- `.. include:: path/to/file.rst`
- all or part of the file contents

### ..md in .rst 
- `.. include:: ../shared/snippet.md`
- same functionality as above 

- use of .md needs [Myst_parser](https://myst-parser.readthedocs.io/en/latest/intro.html) extension ( symbiotic .Rst and .md )
- conf.py : Add pyst_parser
```py
extensions = [
   "myst_parser",
]
```

andrewleech:
Thanks for the feedback everyone, I've reworked the docs in the latest push.

@projectgus yes definitely still interested in this, sorry for the delay.

Main changes:
- Restructured into quickstart (running tests) and developer guide (writing tests, advanced usage), with writingtests.rst as a landing page. The split follows @agatti's suggested topology.
- Improved accuracy of CLI examples and test file references throughout, verified against the actual test runner source.
- Added the filename-based skip patterns list per @agatti's suggestion.
- Added `--begin` injected code customisation per @agatti's note about 6fd7422a8e.
- Added `output_metric` to the multitest API listing.
- Tightened the prose throughout, removed generic filler sections.
- README kept as a lean GitHub-browsable quickstart with a link to the full RST docs.

I haven't added rst include for the README as @Josverl suggested since it would need myst_parser added as a sphinx extension, and the relative links in the README would break when included. The duplication between README and quickstart is minimal now and intentional for the two different audiences.

docs/develop/test_quickstart.rst
```diff
@@ -0,0 +1,148 @@
...
+On Windows use ``python run-tests.py``.
+
+Understanding Results
+~~~~~~~~~~~~~~~~~~~~~
+
+A test run produces output like::
+
+   pass  basics/array1.py
+   skip  basics/async_await.py
+   FAIL  basics/builtin_abs.py
```
agatti, line 49: Should LRGE tests be mentioned too?

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,281 @@
+.. _test_developers:
+
+Writing and Maintaining Tests
+=============================
+
+This covers writing tests and advanced test runner usage. For running
+tests, see :ref:`test_quickstart`.
+
+Test Directory Structure
```
agatti, line 10: There's also `inlineasm/<arch>/`, but not sure if it's worth mentioning.

docs/develop/test_quickstart.rst
andrewleech: Good call, added `lrge` / `SKIP-TOO-LARGE` to the results listing.

docs/develop/test_developers.rst
andrewleech: Added it, three arch subdirs (thumb, xtensa, rv32) are worth knowing about.

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,387 @@
...
+
+A basic print-based test:
+
+.. code-block:: python
+
+   # tests/basics/my_feature.py
+
+   print("Testing feature X")
+   result = some_function()
+   print(result)
```
projectgus: I didn't think of this the first time I looked, but maybe it's worth fully demonstrating how the output comparison works when not using CPython. i.e.

```suggestion
   print("some_function", bool(result))
   
   # tests/basics/my_feature.py.exp
   
   some_function True
```

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,387 @@
...
+- ``int_big*``, ``*_intbig``, ``ffi_int*`` - require arbitrary-precision integers
+- ``bytearray*``, ``*_bytearray`` - require bytearray support
+- ``set_*``, ``frozenset*``, ``*_set`` - require set type
+- ``*slice*`` - require slice support
+- ``async_*``, ``asyncio_*``, ``*_async`` - require async/await
+- ``const*`` - require ``const()`` keyword
+- ``int_64*``, ``*_int64`` - require 64-bit integer support
+- ``*reverse_op*`` - require ``__rOP__`` special methods
+- ``string_fstring*`` - require f-string support
+- ``asm*`` - require inline assembly for the target architecture
```
projectgus: Somewhat minor, but I think this at least easier for *humans* to read 😁 


```suggestion
+-----------------------+----------------------------------------------+
| Pattern               | Requires                                     |
+=======================+==============================================+
| ``native_*``          | Native code emitter                          |
+-----------------------+----------------------------------------------+
| ``viper_*``           | Viper code emitter.                          |
+-----------------------+----------------------------------------------+
| ``*_endian``          | Byte order to match between host and target. |
+-----------------------+----------------------------------------------+
| ``int_big*``,         | Arbitrary-precision integer support.         |
| ``*_intbig``,         |                                              |
| ``ffi_int*``          |                                              |
+-----------------------+----------------------------------------------+
| ``bytearray*``,       | `bytearray` support.                         |
| ``*_bytearray``       |                                              |
+-----------------------+----------------------------------------------+
| ``set_*``,            | `set` type support.                          |
| ``frozenset*``,       |                                              |
| ``*_set``             |                                              |
+-----------------------+----------------------------------------------+
| ``*slice*``           | Slice support.                               |
+-----------------------+----------------------------------------------+
| ``async_*``,          | ``async``/``await`` keyword support.         |
| ``asyncio_*``,        |                                              |
| ``*_async``           |                                              |
+-----------------------+----------------------------------------------+
| ``const*``            | `micropython.const()` keyword support.       |
+-----------------------+----------------------------------------------+
| ``int_64*``,          | 64-bit integer support.                      |
| ``*_int64``           |                                              |
+-----------------------+----------------------------------------------+
| ``*reverse_op*``      | ``__rOP__`` special methods                  |
+-----------------------+----------------------------------------------+
| ``string_fstring*``   | f-string support.                            |
+-----------------------+----------------------------------------------+
| ``asm*``              | Inline assembler for the target architecture.|
+-----------------------+----------------------------------------------+

```

(Same suggestion applies to most of these tabular-ish lists, but I ran out of steam for reformatting - I'm sure Claude has plenty of steam left though!)

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,387 @@
...
+       import socket
+       s = socket.socket()
+       s.connect((IP, 8000))
+       s.send(b'Hello')
+       print(s.recv(1024))
+       s.close()
+
+The ``multitest`` module provides:
+
+- ``next()`` - synchronize instances at a stage boundary
```
projectgus:
```suggestion
  - ``next()`` - Synchronise instances at the beginning of the test. Each instance function should call this function exactly once. This is because the test runner calls each instance function in sequence, and doesn't start the next instance until the current one has called ``next()``. After the last instance has called `next()` then the runner waits for all instances to complete.
```

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,387 @@
...
+- ``broadcast(msg)`` - send message to all instances
+- ``wait(msg)`` - block until a specific broadcast is received
+- ``globals(**kwargs)`` - share variables between instances
+- ``get_network_ip()`` - get the instance's IP address
+- ``expect_reboot(resume, delay_ms=0)`` - handle device reboots during test
+- ``output_metric(data)`` - emit a metric string for collection by the runner
+- ``skip()`` - skip the test from any instance
```
projectgus:
```suggestion
- ``skip()`` - skip the entire test. Can be called by any instance.
```

docs/develop/test_quickstart.rst
```diff
@@ -0,0 +1,150 @@
+.. _test_quickstart:
+
+Running Tests
+=============
+
```
projectgus, line 5: Suggest adding a note here about how these are the tests *for MicroPython itself*, and why you might want to run them.

(We don't really have a good story on how to write tests for your own MicroPython-based application. Can worry about that later, but for now it's good to avoid confusion.)

docs/develop/test_developers.rst
```diff
@@ -0,0 +1,387 @@
+.. _test_developers:
+
+Writing and Maintaining Tests
+=============================
+
+This covers writing tests and advanced test runner usage. For running
```
projectgus: Similar to top of `test_quickstart.rst`, suggest adding a few words about this being the test runners *for MicroPython's own tests*.

tests/README.md
```diff
@@ -1,251 +1,134 @@
-# MicroPython Test Suite
+# MicroPython Test Suite - Quick Start Guide
```
projectgus:
```suggestion
# MicroPython Test Suite
```

(Avoid confusion between this and the other Quick Start guide.)

projectgus (commented):
This is really great @andrewleech, thanks again for submitting it!

I have a bunch more comments but they're all quite minor, I'd also be very happy merging this in its current form if you don't have time to follow up.

andrewleech:
> This is really great @andrewleech, thanks again for submitting it!
> 
> I have a bunch more comments but they're all quite minor, I'd also be very happy merging this in its current form if you don't have time to follow up.

I'll always make time to follow it up, I appreciate the effort you've made to review it in detail. 

I've attempted to address all your notes from the last review thanks.

docs/develop/test_developers.rst
andrewleech: Reordered, the runtime SKIP is documented first now with the filename patterns as the fallback for when the file cant be parsed. Used your wording pretty much as-is, thanks.

docs/develop/test_developers.rst
andrewleech: Agreed, that section is in test_quickstart.rst now.

docs/develop/test_developers.rst
andrewleech: Good idea, there is a full example now: micropython/heap_locked.py with the contents of its .py.exp alongside it.

docs/develop/test_developers.rst
andrewleech: Hehe, converted to a grid table. Did the target selection and multitest API lists the same way while I was in there.

docs/develop/test_developers.rst
andrewleech: Applied, thanks.

docs/develop/test_developers.rst
andrewleech: Applied.

docs/develop/test_quickstart.rst
andrewleech: Added, the intro now says these are the tests for MicroPython itself rather than for your own application, and why you would run them.

docs/develop/test_developers.rst
andrewleech: Same here, there is a scope note at the top now which refers over to test_quickstart for actually running them.

tests/README.md
andrewleech: Applied.

andrewleech:
Rebased onto current master.

Note that 641fa01c0b added a paragraph to `writingtests.rst` about covering non-ASCII / Unicode text in string-related tests, which landed after this branch was originally based. As this PR reduces `writingtests.rst` to a toctree, I have carried that guidance across into `test_developers.rst` as a short Unicode section rather than dropping it.

Josverl:
don't want to cause scope creep but I miss 2 test topics.
- testing mpremote
- testing modules in micropython-lib

these can/should be their own pages in the tree - but I wonder if the current page names are clear enough fur a future addition

- Testing MicroPython
  - Running tests
  - Writing and Maintaining Tests
  - _Running `mpremote` tests_
  - _Testing micropython-lib modules_
  - _Writing and Maintaining Tests for micropython-lib modules_

PR #13390, opened 2024-01-09
Add a BufferedReader and allow BufferedWriter to handle partial writes and errors after some data was written.

klondi:
The commits should already describe much of what is going on but in summary.

Fix a missing include in py/mpconfig.h
Update BufferedWriter to handle partial writes and errors after some data is written.
Add a BufferedReader class handling also partial reads and errors after some data is read.

This is my first time sending code to micropython so I hope I did not make any mistakes.

(Also in case it is necessary, yes the code is mine, it is owned by me and not by any of my employers and I agree to release it under the MIT License).

PR #17302, opened 2025-05-14
tools/verifygitlog.py: Allow long co-author and sign-off names.

DvdGiessen:
While rebasing https://github.com/micropython/micropython/pull/13390 I noticed that the commit messages were being rejected due to the author name on the required `Signed-off-by:` line being longer than the maximum line length. I don't think we should be enforcing a maximum length for human names.

This change disables the line length check for lines that start with `Co-authored-by: ` or `Signed-off-by: `.

I've opened a sibling PR on for updating micropython-lib as well: https://github.com/micropython/micropython-lib/pull/1007

### DIFF
--- a/tools/verifygitlog.py
+++ b/tools/verifygitlog.py
@@ -105,8 +105,12 @@ def verify_message_body(raw_body, err):
 
     # Message body lines.
     for line in raw_body[2:]:
-        # Long lines with URLs are exempt from the line length rule.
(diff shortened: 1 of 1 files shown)


projectgus (approved):
Good idea, we should not punish the long-named (and long-emailed-addressed) authors amongst us. :grin:

dpgeorge (approved):

PR #17306, opened 2025-05-15
tools/mpy-tool.py: Allow dumping MPY segments into their own files.

agatti:
### Summary

This PR lets `tools/mpy-tool.py` extract MPY segments into their own files, one file per segment.

This is something I wrote some time ago but I guess it cannot hurt to be upstreamed.  When debugging issues related with compiled code generated by `@micropython.viper` or `@micropython.native`, it is of great help being able to get hold of generated code segments to pass to objdump or ghidra/idapro/cutter/etc., without having to dump memory from gdb or writing custom file/hex dumpers.

A pair of new command line arguments were added, namely "-e"/"--extract" that takes a filename prefix to use as a base for the generated files' name, and "--extract-only" that - combined with "--extract" - allows selecting which kind of segments should be dumped to the filesystem.

So, for example, assuming there's a file called "module.mpy", running "./mpy-tool.py --extract segments module.mpy" would yield a series of files with names like "segments_0_module.py_QSTR_module.py.bin", "segments_1_module.py_META__module_.bin",
"segments_2_module.py_QSTR_function.bin", etc.  In short the file name format is `<base>_<count>_<sourcefile>_<segmentkind>_<segmentname>.bin`, with `<segmentkind>` being META, QSTR, OBJ, or CODE.  Source file names and segment names will only contain characters in the range "a-zA-Z0-9_-." to avoid having output file names with unexpected characters.

The "--extract-only" option can accept one or more kinds, separated by commas and treated as case insensitive strings.  The supported kinds match what is currently handled by the "MPYSegment" class in "tools/mpy-tool.py": "META", "QSTR", "OBJ", and "CODE".  The absence of this command line option implies dumping every segment found.

If "--extract" is passed along with "--merge", dumping is performed after the merge process takes place, in order to dump all possible segments that match the requested segment kinds.

### Testing

Besides my own usage, I've attached a zipfile containing the compiled version of `tests/micropython/native_try_deep.py` for x64 and its dumped output.  To reproduce those files the commands to run are:

```bash
mpy-cross -X emit=native -march=x64 tests/micropython/native_try_deep.py -o native_try_deep.mpy
mpy-tool.py --extract native_try_deep native_try_deep.mpy
```

To check that the `CODE` segments actually contain executable code, running `objdump -b binary -M x86-64 -m i386:x86-64 --adjust-vma=0x1000 -z --start-address=0x1008 -D native_try_deep_7_native_try_deep.py_CODE_f.bin` should dump valid x64 code to STDOUT, as generated by `mpy-cross` (it skips the first two header words).

[native_try_deep.zip](https://github.com/user-attachments/files/20218071/native_try_deep.zip)

Given that this bit of code isn't executed unless explicitly required and for a niche scenario, the only issue it has would be that it increases the overall code complexity by a tiny amount and potential security issues when the output file prefix is used in a malicious way.

As far as alternatives go, I used to run `mpy-tool.py -x -d <mpyfile>` to figure out the binary code start offset by looking at the hex pairs on screen (and good luck if somebody remapped their terminal colour scheme :) no idea if the output is colourblind safe though).  After a while I wrote my own cut-down `mpy-tool.py` equivalent to run as a ghidra plugin, but then it would require keeping up with MPY format changes and whatnot, and I wasn't sure it would work in all possible cases.

Having `mpy-tool.py` dump the segments itself is probably the best compromise for the time being, it is tool-agnostic and doesn't require anything special to get it working.

### DIFF
--- a/tools/mpy-tool.py
+++ b/tools/mpy-tool.py
@@ -1765,6 +1765,44 @@ def merge_mpy(compiled_modules, output_file):
             f.write(merged_mpy)
 
 
+def extract_segments(compiled_modules, basename, kinds_arg):
+    import re
+
+    kind_str = ("META", "QSTR", "OBJ", "CODE")
+    kinds = set()
+    if kinds_arg is not None:
+        for kind in kinds_arg.upper().split(","):
+            if kind in kind_str:
+                kinds.add(kind)
+            else:
+                raise Exception('unknown segment kind "%s"' % (kind,))
+    segments = []
+    for module in compiled_modules:
+        for segment in module.mpy_segments:
+            if not kinds or kind_str[segment.kind] in kinds:
+                segments.append((module.mpy_source_file, module.source_file.str, segment))
+    count_len = len(str(len(segments)))
+    sanitiser = re.compile("[^a-zA-Z0-9_.-]")
+    for counter, entry in enumerate(segments):
+        file_name, source_file, segment = entry
+        output_name = (
+            basename
+            + "_"
+            + str(counter).rjust(count_len, "0")
+            + "_"
+            + sanitiser.sub("_", source_file)
+            + "_"
+            + kind_str[segment.kind]
+            + "_"
+            + sanitiser.sub("_", str(segment.name))
+            + ".bin"
+        )
+        with open(file_name, "rb") as source:
+            with open(output_name, "wb") as output:
+                source.seek(segment.start)
+                output.write(source.read(segment.end - segment.start))
+
+
 def main(args=None):
     global global_qstrs
 
@@ -1780,6 +1818,14 @@ def main(args=None):
     cmd_parser.add_argument("-f", "--freeze", action="store_true", help="freeze files")
     cmd_parser.add_argument(
         "--merge", action="store_true", help="merge multiple .mpy files into one"
+    )
+    cmd_parser.add_argument(
+        "-e", "--extract", metavar="BASE", type=str, help="write segments into separate files"
+    )
+    cmd_parser.add_argument(
+        "--extract-only",
+        metavar="KIND[,...]",
+        help="extract only segments of the given type (meta, qstr, obj, code)",
     )
     cmd_parser.add_argument("-q", "--qstr-header", help="qstr header file to freeze against")
     cmd_parser.add_argument(
@@ -1847,6 +1893,9 @@ def main(args=None):
 
     if args.merge:
         merge_mpy(compiled_modules, args.output)
+
+    if args.extract:
+        extract_segments(compiled_modules, args.extract, args.extract_only)
 
 
 if __name__ == "__main__":


Josverl:
I think it is worth having this, and for your explanation above to be added to the tools documentation.

tools/mpy-tool.py
```diff
@@ -1765,6 +1765,55 @@ def copy_section(file, offset, offset2):
             f.write(merged_mpy)
 
 
+def parse_extract_segments_arg(arg):
+    kinds = set()
+    if arg is not None:
+        for kind in arg.lower().split(','):
```
dpgeorge: If you use `arg.upper().split(","):` then it could be simply:
```py
try:
    kinds.add(getattr(MPYSegment, kind))
except AttributeError:
    raise Exception("unknown kind")
```
Then, this function could be written inline in the one below.  (A little simpler to have everything self contained in one function, IMO.)

tools/mpy-tool.py
```diff
@@ -1765,6 +1765,55 @@ def copy_section(file, offset, offset2):
...
+
+
+def extract_segments(compiled_modules, basename, kinds_arg):
+    import re
+
+    kinds = parse_extract_segments_arg(kinds_arg)
+    segments = []
+    for module in compiled_modules:
+        for segment in module.mpy_segments:
+            if not kinds or segment.kind in kinds:
```
dpgeorge: If you didn't want to bother validating `kinds_arg`, this could simply be `if not kinds or segment.kind in kinds_arg`.

tools/mpy-tool.py
```diff
@@ -1795,6 +1844,14 @@ def main(args=None):
         default=16,
         help="mpz digit size used by target (default 16)",
     )
+    cmd_parser.add_argument(
```
dpgeorge: I suggest moving this up to just after the `--merge` option, so they appear together in the help output (the action commands will then come before the tweaking options).

dpgeorge (commented):
Thanks, this looks good.  I would have used this more than once myself if it existing already :)

Please rebase on latest master, that will pick up changed ruff rules (namely double quotes).

tools/mpy-tool.py
agatti: Oh, I'm used to use `lower` to work around internationalisation issues in input (f.ex the Turkish I problem).

I agree this is probably not really needed here, but old habits tend to die hard don't they.

tools/mpy-tool.py
agatti: Validation is probably the easier option here, as the extract operation would still proceed with `mpy-tool.py -e module --extract-only= file.mpy`.  This can be interpreted either as "don't extract any segment" or maybe "extract all segments", and in both cases this is still the wrong set of arguments to pass.

Same thing if you accidentally type `cod` instead of `code`.  Since there's no output during the extraction process, as the final user I'd appreciate more having an error telling me I messed up rather than falsely assume the mpy file had no raw code segments in there, for example. 

The alternative would be to add a custom argparse argument type that performs its own validation, which would probably be more code overall.

tools/mpy-tool.py
agatti: Makes sense, thanks!  This will be addressed in the next PR iteration.

agatti:
Turns out I already had the segment kind strings mapped somewhere else in the new code, so I just reused those to simplify the validation.

I believe the patch is now shorter overall, and it should also be compliant with the new ruff formatting rules too.

dpgeorge (approved):
Looks good now, thanks for updating.

PR #17307, opened 2025-05-15
lib/littlefs: Update LittleFS to v2.11.

DvdGiessen:
### Summary

See https://github.com/littlefs-project/littlefs/releases/tag/v2.11.0. Follow-up to #17137.

Upstream now includes the patch we made in 2b29b1b8f9dc7014c27d9df43ca0ebd3ccc86060 for reusing our existing CRC function, so our version no longer contains any changes.

This version also contains a few other small bugfixes (possible overflow on 16bit architectures, double deorphan when removing a directory) and improvements to the build. The only new feature (which is why this is a minor not a patch bump) is behind a new compile-time option, so does not affect us.

### Testing

The release just a few hours old but I've been running it for a few days on a pre-release branch of this release.

No hurry to merge this; it mostly brings a few small bugfixes.

### DIFF
--- a/lib/littlefs/lfs2.c
+++ b/lib/littlefs/lfs2.c
@@ -3932,7 +3932,9 @@ static int lfs2_remove_(lfs2_t *lfs2, const char *path) {
     }
 
     lfs2->mlist = dir.next;
-    if (lfs2_tag_type3(tag) == LFS2_TYPE_DIR) {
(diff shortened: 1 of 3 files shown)


dpgeorge (approved):
Thanks!  This looks fine, and good to keep up-to-date with this driver.

PR #17311, opened 2025-05-16
extmod/asyncio/stream.py: Add ipv6 support to start_server().

andrewleech:
### Summary

I'm starting to use ipv6 more with micropython devices as their discoverability can actually be much better on a local network, without worrying about needing to manually configure (ipv4) ip addresses.

To that end, this PR adds support to start_server() to automatically use the parsed protocol, ie setting `socket.AF_INET` or `socket.AF_INET6` on the socket as parsed by `getaddrinfo()`

When adding unit tests for this change, I found the multi-test runner was missing support for ipv6 ip address detection, so that's been added as well (in a separate commit)

The ip address detection in multi-test runner has also been updated to use the newer / preferred `network.ipconfig("addr4")` style interface by default, falling back to ipconfig if needed.

### Testing

Unit tests are included which check ipv6 support in both `net_hosted` style test and `mult_net` tests. 

Without the change to `extmod/asyncio/stream.py` the tests fails with `OSError: 97`

### DIFF
--- a/extmod/asyncio/stream.py
+++ b/extmod/asyncio/stream.py
@@ -180,11 +180,11 @@ async def start_server(cb, host, port, backlog=5, ssl=None):
     import socket
 
     # Create and bind server socket.
-    host = socket.getaddrinfo(host, port)[0]  # TODO this is blocking!
-    s = socket.socket()
+    addr_info = socket.getaddrinfo(host, port)[0]  # TODO this is blocking!
+    s = socket.socket(addr_info[0])  # Use address family from getaddrinfo
     s.setblocking(False)
     s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-    s.bind(host[-1])
+    s.bind(addr_info[-1])
     s.listen(backlog)
 
     # Create and return server object and task.


extmod/asyncio/stream.py
```diff
@@ -180,11 +180,11 @@ async def start_server(cb, host, port, backlog=5, ssl=None):
     import socket
 
     # Create and bind server socket.
-    host = socket.getaddrinfo(host, port)[0]  # TODO this is blocking!
-    s = socket.socket()
+    addr_info = socket.getaddrinfo(host, port)[0]  # TODO this is blocking!
```
dpgeorge, line 183: It would be possible to add the `family` argument to `start_server()` and pass it through to `getaddrinfo()` here.  Then you could force it to use IPv6 (or IPv4).  But maybe that's a separate PR?

tests/multi_net/asyncio_tcp_server_client_ipv6.py
```diff
@@ -0,0 +1,72 @@
...
+try:
+    import asyncio
+    import socket
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+try:
+    # Check if IPv6 is supported
+    socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
```
dpgeorge: This detection won't work.  All ports based on bare-metal lwIP will let this pass even if they don't have IPv6 enabled.

I don't know if it's possible to detect IPv6??

tests/multi_net/asyncio_tcp_server_client_ipv6.py
```diff
@@ -0,0 +1,72 @@
...
+except (AttributeError, OSError):
+    print("SKIP")
+    raise SystemExit
+
+PORT = 8001  # Different from other tests to avoid conflicts
```
dpgeorge: I don't think this needs to use a different port, 8000 should be fine.

tests/multi_net/asyncio_tcp_server_client_ipv6.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+async def handle_connection(reader, writer):
+    # Test that peername exists
+    peer = writer.get_extra_info("peername")
```
dpgeorge: The `peer` variable is unused.  I don't think you need to call this function at all.

tests/multi_net/asyncio_tcp_server_client_ipv6.py
```diff
@@ -0,0 +1,72 @@
...
+async def tcp_client_ipv6(message):
+    # Connect to the IPv6 server
+    reader, writer = await asyncio.open_connection(IPV6, PORT)
+    print("write:", message)
+    writer.write(message)
+    await writer.drain()
+
+    data = await reader.read(100)
+    print("read:", data)
+    assert data == message, "Data mismatch"
```
dpgeorge: Don't need this assert, the .exp will test the data is correct.

tests/multi_net/asyncio_tcp_server_client_ipv6.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def instance0():
+    # Get the IPv6 address using the new parameter
```
dpgeorge: This comment is not needed, it won't be a new parameter after the PR is merged.

tests/multi_net/asyncio_tcp_server_client_ipv6.py
```diff
@@ -0,0 +1,72 @@
...
+    ipv6 = multitest.get_network_ip(ipv6=True)
+    multitest.globals(IPV6=ipv6)
```
dpgeorge: Please use just a single line, like the other tests:
```py
multitest.globals(IP=multitest.get_network_ip(ipv6=True))
```

tests/net_hosted/asyncio_start_server_ipv6.py
```diff
@@ -0,0 +1,79 @@
...
+try:
+    import asyncio
+    import socket
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+try:
+    # Check if IPv6 is supported
+    socket.socket(socket.AF_INET6, socket.SOCK_STREAM)
```
dpgeorge: As above, this detection won't work.

tests/net_hosted/asyncio_start_server_ipv6.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Start server with IPv6 address
+    print("create ipv6 server")
+    server = await asyncio.start_server(handle_connection, "::", PORT)
+    print("server running")
+
+    try:
+        # Connect with IPv6 client
+        print("connect to ipv6 server")
+        reader, writer = await asyncio.open_connection("::", PORT)
```
dpgeorge: For this to work requires loopback mode in the TCP/IP driver.  That's fine on unix, but it'll fail on most bare-metal ports.

tests/net_hosted/asyncio_start_server_ipv6.py
```diff
@@ -0,0 +1,79 @@
...
+        # Send test data
+        test_msg = b"ipv6 test data"
+        print("write:", test_msg)
+        writer.write(test_msg)
+        await writer.drain()
+
+        # Read response
+        data = await reader.read(100)
+        print("read:", data)
+        assert data == test_msg, "Data mismatch"
```
dpgeorge: Assert not needed.

tests/run-multitests.py
```diff
@@ -79,19 +79,59 @@ def globals(**gs):
             print("SET {{}} = {{!r}}".format(g, gs[g]))
         multitest.flush()
     @staticmethod
-    def get_network_ip():
+    def _get_ip_from_ifconfig(_nic, ipv6=False):
+        # Helper to get IP address from an interface object using appropriate format
```
dpgeorge: This comment doesn't add much, and I'm confused what "appropriate format" means?

tests/run-multitests.py
```diff
@@ -79,19 +79,59 @@ def globals(**gs):
...
+        try:
+            if 'nic' in globals():
+                ip = multitest._get_ip_from_ifconfig(nic, ipv6)
+                if ip:
+                    return ip
+        except:
+            pass
+            
+        # Find active network interface
+        try:
```
dpgeorge: The code below is very hard to understand.  Why so many try/except's?

andrewleech:
The change to the multi test runner and new (duplicated) ipv6 tests have been removed from this PR to simplify it.

Updates to support testing ipv6 have been added in a different way in https://github.com/micropython/micropython/pull/17793

dpgeorge (approved):

dpgeorge:
Actually, on second thoughts, I'm wondering if the `family` argument to `start_server()` is needed at all?  It looks like that's not used in the tests in #17793, so maybe it's a useless (or rarely used) argument?

Adding that argument does add a fair amount to the code size.  Whereas just passing through the family from getaddrinfo to the socket costs almost nothing.

So, if you want to get this in sooner rather than later, I suggest reverting the family argument addition.

andrewleech:
> I suggest reverting the family argument addition.

I've removed that now thanks. If we end up needing it back in for some future use, it can be added then if its adding value!

dpgeorge (approved):
Thanks for updating.  The code size diff is now basically +0, which is great.

PR #17312, opened 2025-05-16
extmod/modlwip,esp32,unix: Add support for socket recv flags argument

projectgus:
### Summary

Goal of this PR is to have support for the `MSG_DONTWAIT` and `MSG_PEEK` flags for socket `recv` & `recvfrom` on unix, esp32, and all "bare metal" LWIP ports.

* Bare metal LWIP adds support for these flags in modlwip.c, with some refactoring to try and reduce the code size impact.
* esp32 port now passes these flags through to the LWIP BSD socket layer, which already supports them.
* unix port already had support for flags argument. `MSG_PEEK` constant is now exposed to Python code.
* Adds multi_net tests for the new flags.
* Adds docs for the `flags` argument.

Note: Zephyr & CC3200 ports have their own `socket` module implementations, support not added to these yet. Zephyr looks like it would be trivially easy, CC3200 I didn't look into.

### Testing

* Ran new and existing multi_net tests on ESP32_GENERIC_S3, RPI_PICO2_W, RPI_PICO_W and PYBD_SF2 boards. One outstanding issue that seems lower level than these changes, see [comment](https://github.com/micropython/micropython/pull/17312#issuecomment-2921051299).

* Support for `MSG_PEEK` is motivated by wanting to improve the DTLS support. Adding `MSG_DONTWAIT` as well was relatively easy, and both of these flags are potentially useful for socket programming on MicroPython. However I guess the obvious alternative is not to add them, or to implement them via a Python wrapper in the Python socket module (which would be hacky but possible).

### DIFF
--- a/tests/multi_net/udp_recv_dontwait.py
+++ b/tests/multi_net/udp_recv_dontwait.py
@@ -1,0 +1,59 @@
+# Test UDP recv and recvfrom with MSG_DONTWAIT
+import random
+import socket
+
+try:
+    import errno, time
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+
+# Server
+def instance0():
+    PORT = random.randrange(10000, 50000)
+    multitest.globals(IP=multitest.get_network_ip(), PORT=PORT)
+    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    s.bind(socket.getaddrinfo("0.0.0.0", PORT)[0][-1])
+    multitest.next()
+    begin = time.ticks_ms()
+
+    # do some recvs before instance1 starts, when we know no packet is waiting
+    try:
+        print(s.recvfrom(8, socket.MSG_DONTWAIT))
+    except OSError as e:
+        print(e.errno == errno.EAGAIN)
+    try:
+        print(s.recv(8, socket.MSG_DONTWAIT))
+    except OSError as e:
+        print(e.errno == errno.EAGAIN)
+
+    # the above steps should not have taken any substantial time
+    elapsed = time.ticks_diff(time.ticks_ms(), begin)
+    print(True if elapsed < 50 else elapsed)
+
+    # Now instance1 will send us a UDP packet
+    multitest.broadcast("0-ready")
+    multitest.wait("1-sent")
+
+    for _ in range(10):  # retry if necessary, to allow for network delay
+        time.sleep_ms(100)
+        try:
+            print(s.recv(8, socket.MSG_DONTWAIT))
+            break
+        except OSError as er:
+            if er.errno != errno.EAGAIN:
+                raise er
+    s.close()
+
+
+# Client
+def instance1():
+    multitest.next()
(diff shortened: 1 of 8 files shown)


dpgeorge (commented):
This looks good, thanks!

I tested on ESP8266, RPI_PICO2_W-RISCV and ESP32_GENERIC, and all the new tests pass.

projectgus:
Have updated the tests and run on various ports, but have found what looks like a lower level issue on RPI_PICO and RPI_PICO_W.

The new `udp_recv_dontwait.py` test fails when instance0 is one of these rp2 boards. The UDP `recv` succeeds, then UDP `sendto()` appears to succeed and the frame appears to be passed into the CYW43 module but packet capture shows it is almost never sent (maybe <10% success). Unclear why, other tests show occasional packet loss but nothing consistent like this.

tests/multi_net/udp_recv_dontwait.py
```diff
@@ -0,0 +1,55 @@
...
+        print(e.errno == errno.EAGAIN)
+
+    # the above steps should not have taken any substantial time
+    elapsed = time.ticks_diff(time.ticks_ms(), begin)
+    print(True if elapsed < 50 else elapsed)
+
+    # Now instance1 will send us a UDP packet
+    multitest.broadcast("0-ready")
+    multitest.wait("1-sent")
+    time.sleep(2)  # allow some extra time for packet to arrive
```
projectgus, line 42: This seems like it shouldn't be necessary, but it is...

dpgeorge: I can understand that this delay is necessary, to let the packet go through the network and be processed by the target device.

But I can reduce this to 0.5 and the tests still run reliably in both directions on RPI_PICO_W and RPI_PICO2_W.  Reducing it down to 0.1 and the tests are less reliable.

Alternatively, could wrap the `s.recv` below in a loop and retry if it raises EAGAIN?

projectgus: Ah, interesting again - anything less than 2 seconds fails on my setup

tests/multi_net/udp_recv_dontwait.py
```diff
@@ -0,0 +1,55 @@
+# Test UDP recv and recvfrom with MSG_DONTWAIT
+import random
+import socket
+
+try:
+    import errno, time
+
+    if not hasattr(time, "ticks_ms"):
```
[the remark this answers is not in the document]
projectgus: Oh, this test has an .exp file that I haven't added to git!

If you want to test against CPython then I can change it though, it's just hard to test the instant return for ports which return int from `time.time()`.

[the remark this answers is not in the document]
dpgeorge: Adding an exp file is fine.

tests/multi_net/udp_recv_dontwait.py
projectgus: Done

dpgeorge (approved):
I've re-reviewed this, it looks good.

I also tested on PYBD_SF2, RPI_PICO_W and ESP32_GENERIC.  The new tests all pass, and so do existing ones (with known failures).

PR #17316, opened 2025-05-17
Calc ESP32 TIMER_SCALE scale by GPTIMER_CLK_SRC_DEFAULT

BSoD123456:
Firstly, I apologize for not knowing how to PR only a portion of the files. So please forgive the confusion of this PR.
Please only focus on the last 2 commits of "ports/esp32/machine_timer.h" and "ports/esp32/machine_timer.c".

### Summary

In esp-edf v5.2, for esp32-c6, the default clock source used by timers is not XTAL_CLK or APB_CLK  (they're all 40mhz in c6), but F80M_CLK (it's 80mhz). Therefore, using APB_CLK_FREQ to calculate TIMER_SCALE will cause the timer to run at double speed.

### Testing

I ran the pre and post modification version on my ESP32C6 dev board, and confirmed that the pre modification timer was at double speed, while the post modification timer at normal speed.


robert-hh:
You have to set up a clean branch first at your side where the change you made are based on latest master branch. If you make  a PR from this branch, then it should show only the last two commits.

dpgeorge:
This PR is not workable.

@BSoD123456 feel free to just post a diff/patch of the bug fix.

PR #9573, opened 2022-10-11
mpremote: Allow user configuration on Windows

Josverl:
Simplifies the use of mpremote's user configuration on Windows, by also considering the Windows standard environment variable `APPDATA`, which is the OS default for storing any user/application related file-based configuration. ( changed from USERPROFILE) 

This avoids the need for a Windows user to set the `HOME` or `XDG_CONFIG_HOME` environment variable, both of which are non-standard on Windows.

Per Microsoft documentation:

> The recommended location for a Windows application to store user settings is in the %APPDATA% folder. This folder is a hidden folder in the user's profile directory that is designed to store application-specific data, including user settings.

However Windows users also are used to %USERPROFILE%, which was established as a location in a previous century.
in order to adhere to both recommended usage ( $XDG_CONFIG_HOME and %APPDATA%) as well as common user practices ($HOME and %USERPROFILE%) , all variables used in a scan for a valid configuration file, and this first config file is used.

Simply stated : newer standards are preferred over the older standards, and Linux environment variables take precidence.

| # | ENV_VAR         | env_value                    | Path_checked                              | platform.system()              |  
| - | --------------- | ---------------------------- | ----------------------------------------- | ------------------------------ | 
| 1 | XDG_CONFIG_HOME | $HOME/**.config**            | /home/foo/**.config**/mpremote                | Linux, Windows(MSYS32), Darwin | 
| 2 | HOME            | /home/boo                    | /home/foo/**.config**/mpremote                | Linux, Windows(MSYS32), Darwin | 
| 3 | APPDATA         | C:\Users\foo\AppData\Roaming | C:\Users\foo\AppData\Roaming\mpremote | Windows                        | 
| 4 | USERPROFILE     | C:\Users\foo                 | C:\Users\foo\mpremote                     | Windows                        |

PR #17322, opened 2025-05-19
tools/mpremote: Add automatic reconnection feature

andrewleech:
### Summary
This PR adds automatic reconnection capability to mpremote and makes it the default behavior. When a MicroPython device disconnects, mpremote automatically reconnects to the same device, eliminating the need to manually restart the tool during development.

Reconnection is now enabled by default. Users can opt-out via:
- The `once` command for per-invocation control
- `MPREMOTE_RECONNECT` environment variable (supports 0/1/true/false/yes/no/on, case-insensitive)
- Config file setting (`reconnect = False` in `~/.config/mpremote/config.py`)
- Priority: `once` command > env var > config file > default (enabled)

Key features:
- Always reconnects to the exact same port that was previously connected when originally started in `connect auto` mode
- Always reconnects to the exact same device (on any port) when started in `connect id:abc123` mode
- Preserves resume state through reconnections
- Handles Windows COM port timing issues with retry logic

Usage:
```bash
# Reconnect is now the default
mpremote repl

# Disable reconnect for this invocation
mpremote once repl

# Disable via environment variable
MPREMOTE_RECONNECT=0 mpremote repl

# Disable via config file
echo "reconnect = False" >> ~/.config/mpremote/config.py
```

This builds on the existing disconnect handling improvements that were previously merged.

### Testing
Manual testing was performed on Linux with Pyboard D and Windows 11 with ESP32S2 devices. Tests used scripted USB power cycling (via uhubctl) to simulate disconnects and verify reconnection behavior. Testing covered default auto-reconnect, the `once` command opt-out, environment variable control (including priority override behavior where `once` takes precedence over `MPREMOTE_RECONNECT=1`), auto mode port-specific reconnection, ID mode device tracking across port changes, multiple device scenarios, resume state preservation, and environment variable parsing with various formats.

Making reconnect the default is a breaking change for scripts that expect mpremote to exit on disconnect. These scripts will need to use the `once` command or set `MPREMOTE_RECONNECT=0`. However, the new default is more intuitive for the primary use case (interactive development).

Design decisions:
- Default changed based on maintainer feedback
- Always reconnects to the exact same port (even in auto mode) rather than any available device, prioritizing safety in multi-device environments
- Added retry logic with delays specifically for Windows COM port timing issues
- Multiple opt-out mechanisms with clear priority order provide flexibility for different use cases
- `once` command takes highest priority as it's the most explicit user intention

The code size increase from the reconnect logic is justified by the significant improvement in developer workflow.

### DIFF
--- a/docs/reference/mpremote.rst
+++ b/docs/reference/mpremote.rst
@@ -54,6 +54,10 @@ If no command is specified, the default command is ``repl``. Additionally, if
 any command needs to access the device, and no earlier ``connect`` has been
 specified, then an implicit ``connect auto`` is added.
 
+By default, ``mpremote`` will automatically reconnect to the device if it
+disconnects. You can disable this behavior using the ``once`` command, the
+``MPREMOTE_RECONNECT`` environment variable, or the user config file.
+
 In order to get the device into a known state for any action command
 (except ``repl``), once connected ``mpremote`` will stop any running program
 and soft-reset the device before running the first command. You can control
@@ -67,6 +71,7 @@ The full list of supported commands are:
 - `connect <mpremote_command_connect>`
 - `disconnect <mpremote_command_disconnect>`
 - `resume <mpremote_command_resume>`
+- `once <mpremote_command_once>`
 - `soft_reset <mpremote_command_soft_reset>`
 - `repl <mpremote_command_repl>`
 - `eval <mpremote_command_eval>`
@@ -134,6 +139,20 @@ The full list of supported commands are:
 
   This disables :ref:`auto-soft-reset <mpremote_reset>`. This is useful if you
   want to run a subsequent command on a board without first soft-resetting it.
+
+.. _mpremote_command_once:
+
+- **once** -- disable automatic reconnection on disconnect:
+
+  .. code-block:: bash
+
+      $ mpremote once
+
+  This disables automatic reconnection, causing mpremote to exit when the
+  device disconnects. By default, mpremote will automatically reconnect to
+  the device if it disconnects. The ``once`` command is useful for scripts
+  that should exit on disconnect (e.g., in automated testing or CI/CD
+  environments).
 
 .. _mpremote_command_soft_reset:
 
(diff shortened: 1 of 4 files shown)


Josverl:
Thanks for the effort, Ill 
> Implemented as a standalone command rather than a --reconnect flag for consistency with other mpremote commands like resume

I have never liked that I find myself needing to type 'resume' to 95% of the commands. 
I think that that resume and reconnect should be the default state, as that is what most users will expect.
 - I think the main use for soft resetting before each command is automated testing, and there adding a reset is simple.  But changing that behavior will need some thought, as it may be a breaking change for tools that depend on it ( although the docs say they should not) 

But for this very welcome reconnect behavior I would propose this to be the default, and to be opted out via `--no-reconnect/--nrc` or environment variable.

andrewleech:
Thanks @Josverl for bringing up the defaults! 

Yes I too have thought it would be better to have `resume` the default, but not entirely sure of what the range of user effects this might have. I do think resume would be less confusing for many people, with an optional `reset` to be used to invert that.

I'm certainly interested in this new reconnect feature being on by default, I expect to use it as _my_ default by adding it to the local config file etc at the very least. I'd certainly be happy to make that change if we get such a concensus from the maintainers!

Josverl:
> adding it to the local config file etc at the very least.

AFAIK that is still broken windows - #9573 would be needed for that

Josverl:
As I usually have multiple boards connected, I frequently want mpremote to (re)connect to a specific one.
I find the CLI for that different to my initial expectations.

| Use Case| connect syntax| Expected | Actual | |
|--------|--------|--------|--------| -
| auto| `mpremote` | `mpremote reconnect` | `mpremote reconnect` | 👌
| COM3:| `mpremote com3` or `mpremote connect com3` | `mpremote reconnect com3` | `mpremote com3 reconnect` | 😕
| COM8:| `mpremote connect COM8` | `mpremote reconnect COM8` | `mpremote connect com8 reconnect` | 😕😕

Similar on linux 

Even without changing default behavior, it would make a lot more sense to me if I can use the same syntax for both the `connect` and `reconnect` commands.
(possibly sharing the same implementation.)

The more complex the commands get, the more confused I get;

As an example:
To set time before connecting:
 -   `mpremote connect com8 rtc --set  exec "import time;print(time.localtime())" repl`
 
perhaps becomes :
 -  `mpremote connect com8 rtc --set  exec "import time;print(time.localtime())" reconnect repl` ❓
 -   `mpremote connect com8 reconnect rtc --set  exec "import time;print(time.localtime())" reconnect repl` ❓

I have not been able to find a form that actually reconnects for this use case,
But I think it would be quite useful to set the time / connect to wifi and similar in such a scenario.

andrewleech:
Thanks for the feedback @Josverl yes having reconnect as a separate command is not as intuitive as I'd first thought.
However I'm actually just about to push a change to make reconnect the default, with a `connect --once <auto/port>` option to reverse the behaviour.
I'm also going to push a separate PR to make resume the default, I believe there's support for that more broadly now.

projectgus:
@andrewleech Changing the defaults like this sounds very sensible to me! Feel free to request a review from me after you've had time to update.

andrewleech:
@Josverl and @projectgus I've reworked this to reconnect by default, with a few options to disable it to revert to "current" behavior.

Hopefully the usage of this will now make better logical sense.

The PR description has been updated to match the new behavior.

PR #17328, opened 2025-05-20
extmod/modframebuf: Add attributes for width, height, format and stride.

andrewleech:
### Summary

  This PR adds direct attribute access to the FrameBuffer class by implementing an attribute handler in the C code. Users can now access the width, height, format, and stride properties directly from
  framebuffer objects using standard attribute syntax (e.g., fb.width, fb.height, etc.).

  This makes it easier to work with framebuffers in applications where these properties need to be accessed frequently, eliminating the need to store and track these values separately. It's particularly
  useful when passing framebuffers between different functions or modules.

Prompted by https://github.com/orgs/micropython/discussions/17266

###  Testing

  I've added a comprehensive test suite in tests/extmod/framebuf_attr.py that validates all the framebuffer attributes with all supported formats:
  - MONO_VLSB, MONO_HLSB, MONO_HMSB
  - RGB565
  - GS2_HMSB, GS4_HMSB, GS8

  The test also checks various dimensions (8x4, 16x8, 128x64) and verifies that attempting to access non-existent attributes raises an AttributeError as expected.

  The implementation was tested on the Unix port, and since this change is contained within the framebuf module and doesn't interact with hardware-specific features, it should work consistently across all
   ports that support the framebuf module.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/library/framebuf.rst
+++ b/docs/library/framebuf.rst
@@ -120,6 +120,25 @@ Drawing text
     dimensions of 8x8 pixels and there is currently no way to change the font.
 
 
+Attributes
+----------
+
+.. attribute:: FrameBuffer.width
+
+    Gets the width of the FrameBuffer in pixels.
+
+.. attribute:: FrameBuffer.height
+
+    Gets the height of the FrameBuffer in pixels.
+
+.. attribute:: FrameBuffer.format
+
+    Gets the format of the FrameBuffer, which is one of the constants defined in this module.
+
+.. attribute:: FrameBuffer.stride
+
+    Gets the stride of the FrameBuffer in pixels (horizontal offset between lines).
+
 Other methods
 -------------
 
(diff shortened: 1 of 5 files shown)


peterhinch:
Some display drivers enable rotation where width and height may be transposed.

My subclasses maintain `.width` and `.height` attributes which match the attributes of the displayed image. These are derived from args passed to the constructor (`height`, `width`, some means of specifying orientation). 

To expand on this, where rotation is done in code there may be no problem if the framebuf is instantiated with `height` and `width` attributes matching the image. But some hardware (e.g. ST7789) performs rotation in hardware, so the framebuffer's dimensions may not match those of the image.

Arguably no action is required here with such subclasses overriding the base class attributes, but comments are welcome.

dpgeorge:
> Some display drivers enable rotation where width and height may be transposed.

Rotation feels like a problem best solved at a layer above `FrameBuffer`.

Or at least right now `FrameBuffer` has no concept of rotation, maybe in the future it could but I'm not sure exactly how that would work.  For now the `width`, `height` and `stride` attributes are those used in clipping calculations, so seem sensible to expose.

PR #17329, opened 2025-05-20
tests/float/math_constants.py: Test actual e and pi constant values.

dpgeorge:
### Summary

The existing test for `math.e` and `math.pi` constants can fail on certain targets if the functions `math.exp()` and/or `math.cos()` are not accurate enough (eg out by an LSB of float precision).  Eg it currently fails on PYBD_SF6 which uses double precision floats.

This PR changes the test for these constants so that it tests the actual constant value, not the evaluation of `exp()` and `cos()` functions.

### Testing

New test ran successfully on the following targets:
- ESP8266 (30-bit precision with object representation C)
- ESP32, RP_PICO, PYBD_SF2 (32-bit precision)
- CPython, unix, PYBD_SF6 (64-bit precision)

EDIT: updated test to work on both little and big endian targets.  CI tests big endian on qemu-mips.
 
The target now requires `array.array` to run the test, but this is enabled at the same level as the `math` module so should always be available when testing these constants.

I did try improving our implementation of double-precision `exp()` so that `exp(1) == e` but that turned out to be very difficult (unsurprisingly).  There is a newer version of musl's math functions, and they do have an improved `exp()` function, but it costs 2k code size just to upgrade that function, and 10k to upgrade the entire double-precision library.  So I ruled out this option.

### DIFF
--- a/tests/float/math_constants.py
+++ b/tests/float/math_constants.py
@@ -1,11 +1,30 @@
 # Tests various constants of the math module.
+
+import sys
+
 try:
-    import math
-    from math import exp, cos
+    from array import array
+    from math import e, pi
 except ImportError:
     print("SKIP")
     raise SystemExit
 
-print(math.e == exp(1.0))
+# Hexadecimal representations of e and pi constants.
+e_truth_single = 0x402DF854
+pi_truth_single = 0x40490FDB
+e_truth_double = 0x4005BF0A8B145769
+pi_truth_double = 0x400921FB54442D18
 
-print(cos(math.pi))
+# Detect the floating-point precision of the system, to determine the exact values of
+# the constants (parsing the float from a decimal string can lead to inaccuracies).
+if float("1e300") == float("inf"):
+    # Single precision floats.
+    e_truth = array("f", e_truth_single.to_bytes(4, sys.byteorder))[0]
+    pi_truth = array("f", pi_truth_single.to_bytes(4, sys.byteorder))[0]
+else:
+    # Double precision floats.
+    e_truth = array("d", e_truth_double.to_bytes(8, sys.byteorder))[0]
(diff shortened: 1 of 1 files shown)


projectgus (approved):
The test approach looks fine to me. I am wondering if this is the kind of test that's likely to ever find a bug, though? Is there any chance that the floating point constants backing `e` and `pi` in the C code will not be the same bit patterns that are shown here?

The original test seems more useful in that it does turn up a bug (the floating-point results for `exp` is very slightly wrong on some platforms), it's just that's also a bug that's really difficult to fix...

The only other path I can think of would be to test something like `abs(math.e - exp(1.0)) <= math.ulp(math.e)`, though. (And MicroPython doesn't have math.ulp() so would probably need to set up the acceptable delta in the test).

dpgeorge:
> I am wondering if this is the kind of test that's likely to ever find a bug, though? Is there any chance that the floating point constants backing `e` and `pi` in the C code will not be the same bit patterns that are shown here?

You are right, it's unlikely to find a bug.  It's mostly a sanity check that the constants exist and aren't broken in some strange way.  Eg they could be swapped e<->pi, or on a big-endian machine maybe the bytes of the representation are swapped.

The original intent of this test was to test the constants have the correct values, so I tried to retain that intent.

> The original test seems more useful in that it does turn up a bug

Yes, and I thought about other ways to test it.  But there are other tests like `tests/float/math_fun.py` which are already explicitly testing values of math functions, including `math.exp`.  So it doesn't seem necessary to retest that here.  This test is about the constants, not the functions.

> The only other path I can think of would be to test something like `abs(math.e - exp(1.0)) <= math.ulp(math.e)`, though.

But that's less strict than what is done here now.  The constants should have the *exact* values.  The functions are the things that can have a small error (and are tested by `math_fun.py`).

projectgus:
@dpgeorge Makes sense, thanks for the additional explanation!

Issue #7266, opened 2021-05-15
Import / AttributeError error in new asyncio 


DrTom:
If some module imports with `from uasyncio import *` and then the this module
is imported  somewhere else there will occur an AttributeError if e.g. `start_server` is invoked: 
```
AttributeError: 'module' object has no attribute 'start_server'
```
The culprit is the "lazy loading" in: https://github.com/micropython/micropython/blob/master/extmod/uasyncio/__init__.py#L8-L30 
Replacing the lazy loading with regular `import` solves the problem.

dpgeorge:
I think we should fix `from uasyncio import *` so that it actually works, using `__all__` to do it (see `mp_import_all()` in `py/runtime.c`).

For now, the workaround would be to just import things explicitly.

yoctopuce:
I believe this issue should be closed as resolved now that the support for `__all__` in star import has been integrated.

PR #17331, opened 2025-05-20
py/runtime.c: Add support for using __all__ in star import.

yoctopuce:
### Summary

If a module defines the symbol `__all__`, that definition should be used when doing a wildcard import: all listed symbols and no others should be added to the parent global environment. This is the standard in CPython.

This pull requests fixes a longtime *FIXME* in `mp_parse_all` to implement this behaviour.

Each item is loaded in the same way as if it would be an explicit import statement, including invoking the the module's `__getattr__` function if needed. This provides a straightforward solution for fixing star import of modules using a dynamic loader, such as `extmod/asyncio` (see issue #7266, dated 2022).

### Testing

The corresponding test cases have been added to `tests/import`, including the default behaviour when `__all__` is not defined.

As star imports are not expected to be useful on systems with very limited ressources, this improvement is only enabled for ports at BASIC_FEATURES level. In order to skip the new test cases on builds at lower feature level, the test file checks for the availability of another feature enabled at the same level (next's 2nd argument).

This improvement will slightly increase the code size, for the sake of CPython compatibility.

On the other hand, limiting imports to a subset of symbols rather than importing most of them in the parent globals will be reduce RAM usage. The feature also fixes a longstanding issue with dynamic loading of modules, which is another method used to reduce RAM usage.


py/mpconfig.h
```diff
@@ -603,6 +603,11 @@
 #define MICROPY_ENABLE_EXTERNAL_IMPORT (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
 #endif
 
+// Whether to enable parsing __all__ when importing all public symbols from modules
+#ifndef MICROPY_ENABLE_MODULE_ALL
```
dlech, line 607:
```suggestion
#ifndef MICROPY_ENABLE_MODULE_IMPORT_ALL
```

perhaps a more informative name?

py/runtime.c
```diff
@@ -1247,6 +1247,19 @@ void mp_load_method(mp_obj_t base, qstr attr, mp_obj_t *dest) {
             mp_raise_msg_varg(&mp_type_AttributeError,
                 MP_ERROR_TEXT("type object '%q' has no attribute '%q'"),
                 ((mp_obj_type_t *)MP_OBJ_TO_PTR(base))->name, attr);
+        #if MICROPY_ERROR_REPORTING >= MICROPY_ERROR_REPORTING_DETAILED
```
dlech, line 1250: This seems like it should be in a separate commit.

py/runtime.c
```diff
@@ -1593,8 +1606,28 @@ mp_obj_t mp_import_from(mp_obj_t module, qstr name) {
 void mp_import_all(mp_obj_t module) {
     DEBUG_printf("import all %p\n", module);
 
-    // TODO: Support __all__
     mp_map_t *map = &mp_obj_module_get_globals(module)->map;
+
+    #if MICROPY_ENABLE_MODULE_ALL
+    mp_map_elem_t *elem = mp_map_lookup(map, MP_OBJ_NEW_QSTR(MP_QSTR___all__), MP_MAP_LOOKUP);
+    if (elem != NULL && mp_obj_is_type(elem->value, &mp_type_list)) {
```
dlech, line 1613: Would it be OK to just let it throw an exception later if the type is wrong?

py/runtime.c
```diff
@@ -1593,8 +1606,28 @@ mp_obj_t mp_import_from(mp_obj_t module, qstr name) {
...
+        // When __all__ is defined, we must explicitly load all specified
+        // symbols, possibly invoking the module __getattr__ function
+        size_t len;
+        mp_obj_t *items;
+        mp_obj_list_get(elem->value, &len, &items);
```
dlech, line 1618:
```suggestion
        mp_obj_get_array(elem->value, &len, &items);
```

Could be a bit more flexible to allow tuple as well.

dlech (commented):
> On the other hand, limiting imports to a subset of symbols rather than importing most of them in the parent globals will be reduce RAM usage

If the module containing `__all__` has to be loaded in RAM (e.g. it is a .py file), then wouldn't RAM usage actually increase because of the extra list (unless there are significantly more "private" members not starting with "_" than public)?

(Not saying we shouldn't accept this, just want to make sure the details are clear.)

py/mpconfig.h
yoctopuce: Agreed, I have changed it accordingly

py/runtime.c
yoctopuce: I have included that error handling code in the same pull request because it is clearly specific to the handling of `__all__` : invalid imports unrelated to `__all__` trigger a separate ImportError exception.

To make this more explicit - and to avoid including this reporting code in case `MICROPY_ENABLE_MODULE_IMPORT_ALL` is not enabled, I have added tham symbol in the `#if` test.

py/runtime.c
yoctopuce: `mp_obj_is_type()` was necessary as I was using `mp_obj_list_get()` afterwards, which does not perform any type checking. But as I have now switched to `mp_obj_get_array()` as per your next comment, this check can indeed be removed,

py/runtime.c
yoctopuce: You are absolutely right. This is actually CPython behaviour, and tuple have the big advantage of being treated as constants in bytecode and beeing fully frozen in ROM as `mp_rom_obj_tuple_t` for frozen modules, requiring no RAM storage.

I have force-pushed the new version with suggested changes.

yoctopuce:
> > On the other hand, limiting imports to a subset of symbols rather than importing most of them in the parent globals will be reduce RAM usage
> 
> If the module containing `__all__` has to be loaded in RAM (e.g. it is a .py file), then wouldn't RAM usage actually increase because of the extra list (unless there are significantly more "private" members not starting with "_" than public)?
> 
> (Not saying we shouldn't accept this, just want to make sure the details are clear.)

It will of course depend on the cases. You are correct in saying that for simple modules, default import is likely to work as well. However for more complex modules that use imports, those imported symbols would typically be unnecessarily re-exported in the parent global dictionary by a star import.

dpgeorge:
Thanks for this.  The "TODO" for this feature has definitely been there a long time, more than 11 years!

It would be nice to see this take up less code size, but I can't see a way to reduce it.  Well, at least it's guarded by a config option, and I agree "basic" level is a good level to enable it at.

py/mpconfig.h
dpgeorge: I think the config name needs to start with `MICROPY_MODULE_`, because that's the existing prefix used for module/package/import configuration, eg `MICROPY_MODULE_GETATTR` enables support for module-level `__getattr__`.

So maybe `MICROPY_MODULE_ALL`?  That does seem a bit obscure, so it could be `MICROPY_MODULE___ALL__` (following `MICROPY_PY___FILE__`) or `MICROPY_MODULE_IMPORT_ALL`

yoctopuce: Sorry for the delay, I was working on the formatfloat PR.
I like this last suggestion, so changed the name to `MICROPY_MODULE___ALL__`, moved the setting next to `MICROPY_PY___FILE__`, and rebased to master.

dpgeorge (approved):

PR #17337, opened 2025-05-22
esp32: auto-detect the SPI flash size and automatically size the filesystem

dpgeorge:
### Summary

Currently in the esp32 port the size of the SPI flash must be configured at build time, eg 4MiB, 8MiB, etc.  Also, the esp32 partition table must be configured at build time, which depends on the size of the SPI flash.  A bigger flash means more can be allocated to the user filesystem.

This PR makes it so the SPI flash size is automatically determined at runtime, and the filesystem size is automatically set to take up as much room as possible.

This works by:
1. Setting the SPI flash size to be 4MiB in the build (or some other value, as long as the firmware app fits).
2. Removing the `vfs` partition from the esp32 partition table (only nvs, phy_init and firmware remains in the partition table).
3. At boot, query the physical size of the SPI flash and use that as the actual size in the code.
4. If it doesn't already exist, automatically create a `vfs` partition which takes up the flash from the end of the firmware partition to the end of flash.

This allows simplifying a lot of board configurations, and removing some board variants that just change the flash size.

It's also fully backwards compatible, in the following sense:
1. Existing boards with MicroPython firmware will continue to work with the same filesystem, ie the filesystem won't be erased when the firmware is updated.
2. If a user has a custom esp32 partition table and installs MicroPython as a bare app into the app partition, the new MicroPython firmware will honor the esp32 partition table and use either `vfs` or `ffat` partitions as the filesystem.

Eventually this mechanism will be extended:
- to other ports, probably rp2 next
- to allow users to customise the filesystem layout, and enable ROMFS partition(s) without rebuilding firmware
 
### Testing

Tested with ESP32_GENERIC on a board with 4MiB of flash, and ESP32_GENERIC_S2 on a board with 16MiB flash.  The filesystem size was correctly detected.

- ~~Boards configured for OTA have not been changed, they still define the filesystem partition explicitly.~~ PR updated to now cover all boards.
- It may be that the physical flash size cannot be detected correctly, but I think the IDF has covered most cases by now.
- Custom board definitions can override the auto detect feature by defining their own `MICROPY_BOARD_STARTUP` macro.


Josverl:
An observation and suggestions.

It took me a long time to understand how variants worked, and what options there are for configuration as there is little documentation, other than in comments in individual make and header files, and in PR comments such as this, which are hard to find later.
(Or perhaps I just failed to find the docs)

In this case there is a new macro defined, that can be overridden, and that is clearly described here.

I think it would be very helpful to add that to the Documentation, on a common or port level. 
That would help tick the 'How can we teach this' checkbox.

dpgeorge:
> I think it would be very helpful to add that to the Documentation, on a common or port level.

@Josverl what exactly are you referring to to add to the documentation?  How board variants work, or how this new auto-sizing feature works?

This PR is intended to simplify things so there are less variants and less things the user needs to worry about (ie selecting the right firmware based on the flash size of the board).  But I agree it should have a little documentation added.  At least on the download page it should say that the firmware works with flash size 4MiB and above.

Josverl:
@dpgeorge , I think that it would be useful to document how the auto-sizing feature works. So that developers can read and understand how it works icw the defied boards and variants.

While I think it's good that fewer variants and boards are needed, this also increases the cognitive effort needed to make use of the existing functionality and (auto) configuration

dpgeorge:
I've now updated this to be more transparent to the user.  It now creates a partition called "vfs" only if there's no such existing partition, and there is either a "factory" or "ota_1" partition.  The latter is needed to work out where the filesystem starts.

This works for all boards/partitions: standard, 2MiB, OTA and ROMFS.

ports/esp32/main.c
```diff
@@ -214,11 +215,15 @@ void boardctrl_startup(void) {
         nvs_flash_erase();
         nvs_flash_init();
     }
+
+    // Query the physical size of the SPI flash and store it in the size
+    // variable of the global, default SPI flash handle.
+    esp_flash_get_physical_size(NULL, &esp_flash_default_chip->size);
```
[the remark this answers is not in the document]
dpgeorge: I've updated to use `CONFIG_SPI_FLASH_SIZE_OVERRIDE=y`.

dpgeorge:
I've now added some documentation to the esp32 README, and also updated the relevant `board.md` files to account for the changes.

ports/esp32/README.md
```diff
@@ -221,6 +221,28 @@ import machine
 antenna = machine.Pin(16, machine.Pin.OUT, value=0)
 ```
 
+Partition table and filesystem size
+-----------------------------------
+
+ESP32 firmware contains a bootloader, partition table and main application
+firmware, which are all stored in (external) SPI flash.  The user filesystem
+is also store in the same SPI flash.  By default, MicroPython does not have
```
Josverl: typo: store --> stored

ports/esp32/README.md
```diff
@@ -221,6 +221,28 @@ import machine
...
+a fixed entry in the ESP32 partition table for the filesystem.  Instead it
+will automatically determine the size of the SPI flash upon boot and then
+create a partition for the filesystem called "vfs" which takes all of the
```
Josverl, line 233: I would suggest 
 "On boot the firmware will look for an existing virtual file system ("vfs" or "ffat") , and use that. If it cannot find a file system, then it will automatically ....

ports/esp32/main.c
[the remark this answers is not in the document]
DvdGiessen: Having thought about it some more I don't think we should enable `CONFIG_SPI_FLASH_SIZE_OVERRIDE=y` by default.

Note it only matters when users are only flashing the application (else the bootloader and application would contain the same value anyway) and when `esp_flash_get_physical_size` fails to detect the flash size. Then we can make a choice: Do we want the size stored in the bootloader, or the size stored in the application?

In cases where the application was built specifically for a certain board/chip it makes sense to default to the value stored in the application. I've specifically needed that in some cases when building a application for a specific board with an incorrect bootloader.

However, in the general case for end-users flashing MicroPython on their boards, the value in the bootloader has a better chance of being correct. After all, the value in the application is most likely completely unrelated to their actual flash size, as our application images are per this change generic and not specific to any flash size.

And if they're not flashing the bootloader (the only case in which this matters), there's a chance that their original bootloader has a correct value because unlike the application it *was* set specifically for their board. Thus, I think by default we should prefer the bootloader value, and thus not enable this option.

> There's also ESPTOOLPY_HEADER_FLASHSIZE_UPDATE which I didn't use because it disables image protection via SHA256.

Hm, I wonder if that comment is up to date. Seems like `esptool.py` [re-calculates the SHA256 if the header was updated](https://github.com/espressif/esptool/blob/0b56f85b17cbf5c32ef4cb7c9b4d3430e9b12ea0/esptool/cmds.py#L454-L455), as I expect it would (since that's fairly easy to do if you're already modifying the image before flashing).

ports/esp32/README.md
dpgeorge: Fixed, thanks.

ports/esp32/README.md
dpgeorge: I've adjusted the text, but not exactly as you suggested (because in the paragraph below, it mentions in more detail about overriding the behaviour).

ports/esp32/main.c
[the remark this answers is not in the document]
dpgeorge: Your reasoning makes sense.  And I would feel more comfortable leaving this option disabled.

Josverl (approved):

ports/esp32/main.c
```diff
@@ -214,11 +215,39 @@ void boardctrl_startup(void) {
...
+
+        // Try to find the application partition, to work out where to start the filesystem.
+        const esp_partition_t *part_factory = esp_partition_find_first(ESP_PARTITION_TYPE_APP, ESP_PARTITION_SUBTYPE_ANY, "factory");
+        const esp_partition_t *part_ota_1 = esp_partition_find_first(ESP_PARTITION_TYPE_APP, ESP_PARTITION_SUBTYPE_ANY, "ota_1");
+        size_t offset = 0;
+        if (part_factory != NULL && part_ota_1 == NULL) {
+            offset = part_factory->address + part_factory->size;
+        } else if (part_factory == NULL && part_ota_1 != NULL) {
+            offset = part_ota_1->address + part_ota_1->size;
+        }
```
projectgus, line 236: I don't know if it's worth worrying about, but it might be more robust for custom layouts to relax this requirement to "after the last partition in the flash". i.e. (untested):

```c
// Find the end of the last existing partition
size_t offset = 0
esp_partition_iterator_t it = esp_partition_find(ESP_PARTITION_TYPE_ANY, ESP_PARTITION_SUBTYPE_ANY, NULL);
while (it != NULL) {
  const esp_partition_t *part = esp_partition_get(it);
  offset = MAX(offset, part->address + part->size);
  it = esp_partition_next(it);
}
```

projectgus (approved):
This is a really neat change, simplifies so much!

I left one possible suggestion although I'm not sure if it's worth the change, LGTM either way.

ports/esp32/main.c
dpgeorge: Yes, that's a good idea, better than what I have here.  It will also be slightly faster because then it only iterates once through the partitions (which requires an alloc in the IDF...).

Your suggestion will also fix a bug in the current code in that it doesn't respect the "romfs" partition and will overwrite it :grimacing:

dpgeorge: Now updated, and tested with ESP32_GENERIC firmware on a 4MiB board.

ports/esp32/README.md
```diff
@@ -221,6 +221,29 @@ import machine
...
+-----------------------------------
+
+ESP32 firmware contains a bootloader, partition table and main application
+firmware, which are all stored in (external) SPI flash.  The user filesystem
+is also stored in the same SPI flash.  By default, MicroPython does not have
+a fixed entry in the ESP32 partition table for the filesystem.  Instead it
+will automatically determine the size of the SPI flash upon boot, and then --
+so long as there is no existing partition called "vfs" or "ffat" -- it will
+create a partition for the filesystem called "vfs" which takes all of the
+remaining flash between the end of the application up until the end of flash.
```
projectgus:
```suggestion
remaining flash between the end of the last partition defined in the partition table up until the end of flash.
```

dpgeorge: Thanks, now updated.

dpgeorge:
Tested again on a board using `partitions-4MiB-romfs.csv`, it works correctly, the ROMFS is intact and the "vfs" partition starts after ROMFS.

PR #17338, opened 2025-05-22
shared/tinyusb: Use device event hook to schedule USB task.

projectgus:
### Summary

Previously MicroPython ports would linker-wrap dcd_event_handler in order to schedule the USB task callback to run when needed.

TinyUSB 0.16 added proper support for an event hook to do the same thing without the hacky linker wrapping. (See https://github.com/hathach/tinyusb/pull/2303)

(The vendored TinyUSB is 0.17.0 since 09fa90e, the version in ESP-IDF is 0.18.0 for both ESP-IDF V5.2 and V5.4.1.)

As a bonus: 

* This resolves a possible reported issue with LTO+linker wrap on the nrf port (as there's no more linker wrapping).
* Can revert commit 62e0fa04 from #15158 for the same reason (no more linker wrap in the nrf port)

### Testing

* Built `RPI_PICO` and `ESP32_GENERIC_S3` boards, interacted with the default REPL, and also ran the `usb/device/mouse` example on both boards.

### DIFF
--- a/ports/alif/alif.mk
+++ b/ports/alif/alif.mk
@@ -103,10 +103,6 @@ CFLAGS += -Wl,-T$(BUILD)/ensemble.ld \
           -Wl,--print-memory-usage \
           -Wl,--no-warn-rwx-segment
 
-ifeq ($(MCU_CORE),M55_HP)
-CFLAGS += -Wl,--wrap=dcd_event_handler
-endif
-
 ################################################################################
 # Source files and libraries
 
--- a/ports/esp32/esp32_common.cmake
+++ b/ports/esp32/esp32_common.cmake
@@ -97,10 +97,6 @@ if(MICROPY_PY_TINYUSB)
     list(APPEND MICROPY_INC_TINYUSB
         ${MICROPY_DIR}/shared/tinyusb/
     )
-
-    list(APPEND MICROPY_LINK_TINYUSB
-        -Wl,--wrap=dcd_event_handler
-    )
 endif()
 
 list(APPEND MICROPY_SOURCE_PORT
@@ -259,10 +255,6 @@ target_compile_options(${MICROPY_TARGET} PUBLIC
     -Wno-clobbered
     -Wno-deprecated-declarations
     -Wno-missing-field-initializers
-)
-
-target_link_options(${MICROPY_TARGET} PUBLIC
-     ${MICROPY_LINK_TINYUSB}
 )
 
 # Additional include directories needed for private NimBLE headers.
--- a/ports/nrf/Makefile
+++ b/ports/nrf/Makefile
@@ -268,7 +268,6 @@ SRC_C += $(addprefix lib/tinyusb/src/,\
 	portable/nordic/nrf5x/dcd_nrf5x.c \
 	)
 
-LDFLAGS += -Wl,--wrap=dcd_event_handler
 endif
 
 DRIVERS_SRC_C += $(addprefix modules/,\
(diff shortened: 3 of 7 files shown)


andrewleech (approved):
Very nice find @projectgus ! That's much cleaner.

I do see that `queue_event()`/`tud_event_hook_cb()` is called from a number of places in the ISR for different types of events, my initial thought was maybe we only want to trigger on some events or prevent repeat calls if there are multiple events in one ISR. 

But now I've confirmed that's not an issue, this new hook is queing up the usb task in the c-scheduler which has built in protection against re-scheduling of the same node, subsequent schedules are just ignored if it's already queued to be run (`py/scheduler.c` -> `mp_sched_schedule_node()`)

dpgeorge:
Tested on ARDUINO_NANO_33_BLE_SENSE with the following:
```py
>>> import time
>>> print('test');time.sleep(1)
test
>>> 
```
Prior to this PR, the "test" output would not appear until after the 1 second elapsed.  With this PR it now works properly, "test" is printed immediately, and then it waits 1 second before the prompt appears.

Great work @projectgus getting this hook merged to TinyUSB!

dpgeorge (approved):

dpgeorge:
Could we now get rid of `MICROPY_LINK_TINYUSB` in `ports/esp32/esp32_common.cmake`?

projectgus:
> Could we now get rid of MICROPY_LINK_TINYUSB in ports/esp32/esp32_common.cmake?

Oh, good catch! Have taken it out as well.

> Prior to this PR, the "test" output would not appear until after the 1 second elapsed. With this PR it now works properly, 

Huh, that's good! I see the nrf port of TinyUSB sometimes calls `usb_defer_func()` to defer an ISR callback, and this now also queues processing of the TinyUSB task.

projectgus:
Pushed one more small tweak, can remove the dcd.h include line from mp_usbd.c

dpgeorge:
> can remove the dcd.h include line from mp_usbd.c

Very good!

robert-hh:
@dpgeorge I tried to replicate your test with `print('test');time.sleep(1)` by fetching the actual PR and building it. It still prints "test" after one second. Did you update the tinyusb lib?

dpgeorge:
> Did you update the tinyusb lib?

No, I didn't change anything else.  Just this PR on latest master.  Maybe you need to `make clean` first?  And did you verify that it flashed the device correctly, eg check the date of the build on the device?

robert-hh:
I always start with `make clean`. But I had to update the tinyusb lib to 0.18.0. That did not happen with the PR. So maybe the PR misses the version change of tinyUSB.

dpgeorge:
> But I had to update the tinyusb lib to 0.18.0.

TinyUSB should be at 0.17.0 on current master, and that's all that is needed for this PR to work.

Can you please try with a fresh clone of this repo?  It really should work without needing to update TinyUSB.

robert-hh:
If I make a fresh clone of MicroPython, run `make submodules` in the nrf directory, checkout the PR 17338 and create a firmware, it works. But I do not understand why.
- the git hashes of the PR17338 MicroPython branch and the tinyusb and nrfx submodules are the same.
- the files of both ports/nrf subdirectories are identical.
- the compiler used for building is the same.
So the output ought to be the same. But the firmware files differ in size and content.

dpgeorge:
> So the output ought to be the same. But the firmware files differ in size and content. Do you have a hint for a reason?

Do you have a `makefile` or `GNUmakefile` in `ports/nrf` that is affecting the build?  Or maybe some other "hidden" files?  If you want to dig deeper, you can start diff'ing the object files (eg see which ones changed size) and even diff'ing the assembly dump of firmware.elf.  But that's a fair bit of work.  If it works with a fresh clone then I'd say it works.

robert-hh:
Thank you for your answer that late in the night.
I had a GNUmakefile, but removing it did not change anything. I compared with diff the tinyusb & nrf libs.
I compared the two map files of good vs bad build. There are some differences mostly in the order of the various files in created firmware. Comparing object files just tells, that all of them are different. So I went the simply way by re-cloning MP and setting up the environment as I had it before. Doing that I rebased & pushed all my branches, causing some noise with github messages.

PR #17341, opened 2025-05-22
STM32G4: QSPI fix

H-Grobben:
### Summary

Adding a QSPI memory chip on a STM32G4 does not work due to some small issues. These are fixed in this PR.

### Testing

Our custom development board (AEMICS PYglet) has QSPI Flash. Tested with that.
Nucleo G4 board does not natively have a QSPI flash, so that is not tested.


dpgeorge (approved):
Looks good, thanks.

PR #17342, opened 2025-05-22
py/emitnative: Further Viper integer-indexed load/store improvements.

agatti:
### Summary

This PR introduces improvements in the code that gets generated for integer-indexed load and store operations for various architectures.

The Viper integer-indexed generator bits are now able to utilise a specialised operation implementation if the platform emitter exposes one.  This means that macros named `ASM_{LOAD,STORE}{8,16,32}_REG_REG_OFFSET` are used if they are present in the platform backend.

`ASM_LOAD16_REG_REG_OFFSET`, `ASM_LOAD32_REG_REG_OFFSET`, and `ASM_STORE32_REG_REG_OFFSET` were already present since they are used by the native emitter, so their implementations are used also by the Viper emitter now.  This may or may not reduce the code size, but it streamlines the current behaviour by having one single optimised implementation on each platform.

However, those three implementations are now exposed to scenarios where accessing larger than usual offsets is a somewhat probable occurrence - requiring some changes to make them work in all possible cases on Arm, Thumb, and Xtensa.  The full set of platform-optimised integer indexed load/store operations is not implemented for all Viper operations on all platforms, as the focus of this PR is ARMv7-M improvements.  The rest of the changes is there to adapt the other platform backends to the infrastructure changes needed for ARMv7-M.  The remaining load/store operations' code will come later in a different PR if the code size increase can be kept to a minimum.

The Arm emitter is now able to generate `LDR` and `STR` integer-indexed opcodes whose offsets can span the full 32 bits address space.  The original behaviour would be undefined if the offset was larger than 12 bits, as it would corrupt the upper bits of the generated opcode itself.

The Thumb emitter is now able to generate ARMv7-M opcodes for all data sizes if both ARMv7-M opcodes are allowed and if the opcode offset spans between 6 and 12 bits.  In that case a single 32 bits opcode will be generated (`LDR{B,H}.W` and `STR{B,H}.W`), falling back to the existing generic code sequence generator if the offset is too big to be embedded in a single opcode.

The Xtensa emitter is now able to generate opcode sequences spanning the full 32 bits address space for 32 bits loads and stores, and for 16 bit loads.  The existing implementations would raise a `RuntimeError` if the offset could not be embedded in a single opcode.

### Testing

The native and viper tests in `tests/micropython` ran successfully on QEMU `MPS2-AN385`, `SABRELITE`, and `VIRT_RV32` boards, on a NodeMCU v3 8266 board, on a generic ESP32 D0WD board, and on x64 via the Unix port.

I've tried to keep the extra code footprint as limited as possible without having to make the code too complex to follow (leaving some potential ARMv7-M only optimisations out as well, taking up a bit more space).

This is mostly evident for armv6/armv7 targets, where the overhead should be ~200 bytes.  This was offset a bit by moving out some code from the native emitter to the platform backends, where a simpler implementation could suffice.

However there are some benefits in more optimised code being generated, and in a slight decrease of complexity in the Viper emitter as less platform-specific code is present there.


py/asmx86.h
```diff
@@ -200,16 +200,18 @@ void asm_x86_call_ind(asm_x86_t *as, size_t fun_id, mp_uint_t n_args, int temp_r
 #define ASM_SUB_REG_REG(as, reg_dest, reg_src) asm_x86_sub_r32_r32((as), (reg_dest), (reg_src))
 #define ASM_MUL_REG_REG(as, reg_dest, reg_src) asm_x86_mul_r32_r32((as), (reg_dest), (reg_src))
 
-#define ASM_LOAD_REG_REG_OFFSET(as, reg_dest, reg_base, word_offset) asm_x86_mov_mem32_to_r32((as), (reg_base), 4 * (word_offset), (reg_dest))
+#define ASM_LOAD_REG_REG_OFFSET(as, reg_dest, reg_base, word_offset) ASM_LOAD32_REG_REG_OFFSET((as), (reg_base), (word_offset), (reg_dest))
```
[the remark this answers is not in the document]
agatti: OK, this should be fixed - I've just run `ci_unix_coverage_32bit_run_tests` on a VM and it went through the whole thing without errors.

The reason for the x86 argument issue could probably be a cut and paste operation gone awry, from a different emitter or from somewhere else within the same emitter.  Anyway, with the test runner changes I was able to let the test run fail.

The test runner changes exposed a different issue when iterating through `ci_unix_coverage_32bit_run_tests` test runs without cleaning everything from scratch, when the binary changes gcov prints a warning to STDOUT that gets intercepted by the feature check output parser, and makes it raise an Exception.  Granted, the fix is to run `ci_unix_coverage_32bit_run_tests` again so gcov won't complain, I wonder if that's something worth looking into or not (not in this PR, of course).

agatti:
Hrm, the test runner change is clashing with `nanbox`?  Looking into it now...

Yep, `nanbox` doesn't support native emitters and CI requests a full test run, which forces native code output, and you can guess the rest.

agatti:
There are lots of ways to work around the `nanbox` CI issue - please let me know if the one I chose is OK with you.

py/asmx86.h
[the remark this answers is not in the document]
dpgeorge: > when the binary changes gcov prints a warning to STDOUT that gets intercepted by the feature check output parser, and makes it raise an Exception.

Yes, I've seen that a lot.  Simply by running `make VARIANT=coverage test` in the unix port directory.  I simply run it a second time if the gcov output interferes.

py/asmxtensa.c
```diff
@@ -42,6 +42,10 @@
 #define ET_OUT_OF_RANGE MP_ERROR_TEXT("ERROR: xtensa %q out of range")
 #define ET_NOT_ALIGNED MP_ERROR_TEXT("ERROR: %q %q not word-aligned")
 
+// A7 is part of the incoming parameter registers set in both CALL0 and
+// windowed ABIs, thus is not callee-saved and can be safely clobbered.
+#define REG_TEMP ASM_XTENSA_REG_A7
```
dpgeorge: This is problematic: A7 is used as a base pointer to the function table `mp_fun_table` when in windowed mode.  So clobbering it will lead to a crash if the function table is used afterwards.

Eg the following crashes on esp32:
```py
@micropython.native
def f(p: ptr32):
    p[256] = 1
    print('done')

f(bytearray(2048))
```
It's a bit tricky to fix.  The native emitter does allocate a register for calculating the index `reg_index` but that's not passed through to the backend emitter.

py/asmthumb.h
```diff
@@ -462,16 +464,21 @@ void asm_thumb_b_rel12(asm_thumb_t *as, int rel);
...
-#define ASM_LOAD_REG_REG_OFFSET(as, reg_dest, reg_base, word_offset) asm_thumb_ldr_reg_reg_i12_optimised((as), (reg_dest), (reg_base), (word_offset))
-#define ASM_LOAD8_REG_REG(as, reg_dest, reg_base) asm_thumb_ldrb_rlo_rlo_i5((as), (reg_dest), (reg_base), 0)
-#define ASM_LOAD16_REG_REG(as, reg_dest, reg_base) asm_thumb_ldrh_rlo_rlo_i5((as), (reg_dest), (reg_base), 0)
-#define ASM_LOAD16_REG_REG_OFFSET(as, reg_dest, reg_base, uint16_offset) asm_thumb_ldrh_reg_reg_i12_optimised((as), (reg_dest), (reg_base), (uint16_offset))
-#define ASM_LOAD32_REG_REG(as, reg_dest, reg_base) asm_thumb_ldr_rlo_rlo_i5((as), (reg_dest), (reg_base), 0)
-
-#define ASM_STORE_REG_REG_OFFSET(as, reg_src, reg_base, word_offset) asm_thumb_str_rlo_rlo_i5((as), (reg_src), (reg_base), (word_offset))
-#define ASM_STORE8_REG_REG(as, reg_src, reg_base) asm_thumb_strb_rlo_rlo_i5((as), (reg_src), (reg_base), 0)
-#define ASM_STORE16_REG_REG(as, reg_src, reg_base) asm_thumb_strh_rlo_rlo_i5((as), (reg_src), (reg_base), 0)
-#define ASM_STORE32_REG_REG(as, reg_src, reg_base) asm_thumb_str_rlo_rlo_i5((as), (reg_src), (reg_base), 0)
```
dpgeorge, line 474: These inline functions like `asm_thumb_strb_rlo_rlo_i5()` are now unused, so could be removed.  But maybe worth keeping them around in case they are needed again one day?

py/asmarm.c
```diff
@@ -333,9 +333,16 @@ void asm_arm_asr_reg_reg(asm_arm_t *as, uint rd, uint rs) {
...
-    // ldr rd, [rn, #off]
-    emit_al(as, 0x5900000 | (rn << 16) | (rd << 12) | byte_offset);
+void asm_arm_ldr_reg_reg_offset(asm_arm_t *as, uint rd, uint rn, uint byte_offset) {
+    if (byte_offset < 0x1000) {
+        // ldr rd, [rn, #off]
+        emit_al(as, 0x5900000 | (rn << 16) | (rd << 12) | byte_offset);
+    } else {
+        // mov r8, #off
+        // ldr rd, [rn, r8]
+        asm_arm_mov_reg_i32_optimised(as, ASM_ARM_REG_R8, byte_offset);
```
dpgeorge: Is it worth doing:
```c
#define REG_TEMP ASM_ARM_REG_R8
```
and then using `REG_TEMP` here?  And also change the other locations in this file that use r8.  Then at least it's clear that this is an available temporary register.

py/asmxtensa.c
agatti: This is weird, as I'm quite sure I ran the full set of `tests/micropython/` tests on a D0WD.  `tests/micropython/viper_ptr32_store_boundary.py` would have crashed when using the last set of offsets (around the 12-bits boundary mark).

But I've seen the constant in question so this needs fixing.  

I believe that using `REG_TEMP2` could be a temporary workaround - don't have an ESP32 available at the moment to test it.

`REG_TEMP2` is only used for implementing the `rot_three` opcode, and the native implementation whilst performing register stores would only use the stack for with that register.  How big can a single emitted function stack grow?  If that number is below 1024 bytes then `REG_TEMP2` should be it, since it would not touch the extra register; otherwise I'll see how this can be worked around differently.

Or I can just track an additional register's usage and load/store it on demand in the entry and exit functions like what I've done for `INTERNAL_TEMPORARY` on RV32 and that would leave `rot_three` alone.

dpgeorge: > `tests/micropython/viper_ptr32_store_boundary.py` would have crashed when using the last set of offsets (around the 12-bits boundary mark).

The reason it doesn't crash is because it doesn't access A7 after clobbering it.  In my test above you need the `print('done')` to trigger the crash, because that line requires A7.

py/asmarm.c
agatti: I briefly thought about it when touching the code but I decided to leave things alone to minimise the amount of code touched by that particular commit - same thing for all the checks for `(reg < ASM_THUMB_REG_R8)` in `py/asmthumb.c` that can be moved into a macro.

Anyway, I've added a commit to do just that, no big deal.

py/asmthumb.h
agatti: They're not adding any code to the final binary and they're under version control anyway.  I'd personally take them out but maybe some external code is relying on those functions?

py/asmxtensa.c
agatti: OK, using `REG_TEMP2` (or the same register anyway) seems to work on a regular ESP32 board when running the test suite with `--via-mpy --emit native`.

I've had to extend the range of some branch opcodes to make `basics/frozenset_binop.py`, `basics/set_binop.py`, `extmod/framebuf1.py`, and `extmod/json_dumps_separators.py` actually compile for xtensa/xtensawin in the first place.

I haven't tested this on an ESP8266 yet (need to fetch the board from somewhere else) but, unless checking for `N_XTENSAWIN` when picking which register number to use is not correct, in this case if it works on ESP32 then it should also work for ESP8266 unless this throws alignment off for inlineasm tests if I didn't handle things right when updating branch opcodes.

agatti: Oh, and thanks for the test case - I wonder why I didn't notice that.

dpgeorge: > I've had to extend the range of some branch opcodes to make `basics/frozenset_binop.py`, `basics/set_binop.py`, `extmod/framebuf1.py`, and `extmod/json_dumps_separators.py` actually compile for xtensa/xtensawin in the first place.

Excellent, thanks for that.  I had noticed for some time that these tests fail, but with all the other native changes in flight, I was waiting before mentioning it.  See related #17293.

agatti: > I haven't tested this on an ESP8266 yet (need to fetch the board from somewhere else) but, unless checking for `N_XTENSAWIN` when picking which register number to use is not correct, in this case if it works on ESP32 then it should also work for ESP8266 unless this throws alignment off for nowinlineasm tests if I didn't handle things right when updating branch opcodes.

I've just had the chance to test this on an ESP8266 - inline asm branches aren't affected, some tests crash with `--via-mpy --emit=native`.  I'm looking into it.

agatti: OK, I've misread the ISA document - `BEQZ.N` and `BNEZ.N` can only operate with an *unsigned* displacement, and on an ESP32 there were no backward jumps below the (signed) 6-bits threshold that triggered the emission of those opcodes.

This should be fixed now!

py/asmthumb.h
dpgeorge: Let's leave them in for now, until all these changes to the native emitter settle down.

dpgeorge (approved):
Thanks for updating/fixing.  It looks like everything is resolved now.

Tested on ESP8266_GENERIC and ESP32_GENERIC, with `run-tests.py --via-mpy --emit native`.  Not everything passes, but those that fail did already fail before this PR.

PR #17344, opened 2025-05-22
ports/rps: Make FLASH LENGTH match PICO_FLASH_SIZE_BYTES in .ld files.

### LINKED ISSUE #17343: ports/rp2: Make `FLASH` length configurable in linker script.
### Description

Right now, the `FLASH` length in `memmap_mp_rp2040.ld` and `memmap_mp_rp2350.ld` are hard-coded to the flash size of the Pico (2MB) and Pico 2 (4MB) respectively.

https://github.com/micropython/micropython/blob/7f6fedef2aeb0234f3c7d35a735ade75b9fdeaf3/ports/rp2/memmap_mp_rp2040.ld#L26

https://github.com/micropython/micropython/blob/7f6fedef2aeb0234f3c7d35a735ade75b9fdeaf3/ports/rp2/memmap_mp_rp2350.ld#L26

However, there are a number of RP2 boards the have different amounts of flash. For example, all of our (SparkFun) RP2 boards have 16MB of flash. This feature request is to enable a method for the flash size in the linker script to be configurable, ideally matching whatever is set by `PICO_FLASH_SIZE_BYTES`.

While these hard-coded values are fine for base MicroPython, we're working on a project (can't say what 🤫) that results in the firmware being over 2MB, possibly even over 4MB. This results in the linker failing because the firmware size exceeds the hard-coded flash length values:

```
firmware.elf section `.text' will not fit in region `FLASH'`
region `FLASH' overflowed by xxxx bytes
```

### Code Size

This should have no effect on the code size, other than enabling larger projects to actually fit into boards with larger flash chips. And this will ensure the flash size matches what is actually on each board.

I believe it should be trivial to define a symbol for the linker script, similar to `__micropy_c_heap_size__` for example:

https://github.com/micropython/micropython/blob/49f81d5046aaeb31f90626426363ae2518dbd810/ports/rp2/CMakeLists.txt#L527

### Implementation

I intend to implement this feature and would submit a Pull Request if desirable

### Code of Conduct

Yes, I agree

sfe-SparkFro:
### Summary

Resolves #17343

This dynamically sets the `FLASH` `LENGTH` value in the `.ld` files to match `PICO_FLASH_SIZE_BYTES`, meaning it should match the size of the actual flash chip on whatever board is being built.

### Testing

I have a project that requires over 4MB of flash for the firmware. I'm building the project for a board that has a 16MB flash chip, which is set by `PICO_FLASH_SIZE_BYTES`. Without these changes, I get the following error from the linker:

```
firmware.elf section `.text' will not fit in region `FLASH'`
region `FLASH' overflowed by xxxx bytes
```

After these changes, it works fine.

I do not believe this has any tradeoffs.

However, a possible alternative is to make the `FLASH` `LENGTH` truly configurable instead of dynamically set to match the board's flash size. I don't know whether that would actually be useful for anything though, so IMO best to keep it simple.

### DIFF
--- a/ports/rp2/CMakeLists.txt
+++ b/ports/rp2/CMakeLists.txt
@@ -525,6 +525,7 @@ target_compile_options(${MICROPY_TARGET} PRIVATE
 
 target_link_options(${MICROPY_TARGET} PRIVATE
     -Wl,--defsym=__micropy_c_heap_size__=${MICROPY_C_HEAP_SIZE}
+    -Wl,--defsym=__micropy_flash_size__=${PICO_FLASH_SIZE_BYTES}
     -Wl,--wrap=dcd_event_handler
     -Wl,--wrap=runtime_init_clocks
 )
--- a/ports/rp2/CMakeLists.txt
+++ b/ports/rp2/CMakeLists.txt
@@ -525,10 +525,23 @@ target_compile_options(${MICROPY_TARGET} PRIVATE
 
 target_link_options(${MICROPY_TARGET} PRIVATE
     -Wl,--defsym=__micropy_c_heap_size__=${MICROPY_C_HEAP_SIZE}
-    -Wl,--defsym=__micropy_flash_size__=${PICO_FLASH_SIZE_BYTES}
     -Wl,--wrap=dcd_event_handler
     -Wl,--wrap=runtime_init_clocks
 )
+
+if(PICO_FLASH_SIZE_BYTES GREATER 0)
+    target_link_options(${MICROPY_TARGET} PRIVATE
+        -Wl,--defsym=__micropy_flash_size__=${PICO_FLASH_SIZE_BYTES}
+    )
+elseif(PICO_RP2040)
+    target_link_options(${MICROPY_TARGET} PRIVATE
+        -Wl,--defsym=__micropy_flash_size__=2048k # Default to 2MB
+    )
+elseif(PICO_RP2350)
+    target_link_options(${MICROPY_TARGET} PRIVATE
+        -Wl,--defsym=__micropy_flash_size__=4096k # Default to 4MB
+    )
+endif()
 
 if(PICO_RP2350)
     target_link_options(${MICROPY_TARGET} PRIVATE
--- a/ports/rp2/memmap_mp_rp2040.ld
+++ b/ports/rp2/memmap_mp_rp2040.ld
@@ -23,7 +23,7 @@
 
 MEMORY
 {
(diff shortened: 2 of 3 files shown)


dpgeorge:
In principle this looks reasonable, although I have two concerns:
1. It requires board header files to correctly use `pico_cmake_set_default PICO_FLASH_SIZE_BYTES`, and not all of them do (the pico-sdk has some magic to grep for these strings and then expose the settings to cmake...).  Eg NULLBITS_BIT_C_PRO doesn't have that definition and doesn't build with this PR.
2. Eventually I want to auto-detect the size of the SPI flash, and allow the user to specify the size of filesystems (ROMFS, littlefs, FAT, etc) using a partition table.  That might require a bit of trickery in the linker script, similar to what's done here.

(1) is pretty easy to fix for the boards in this repo (there's also POLOLU_ZUMO_2040_ROBOT, WEACTSTUDIO and WEACTSTUDIO_RP2350B_CORE to fix), although anyone with a custom board outside this repo might also need to add such a definition.

(2) is probably not an issue but worth considering.

sfe-SparkFro:
@dpgeorge How about this? If `PICO_FLASH_SIZE_BYTES` is not set (actually, it appears to get set to `""`), then it uses the same defaults as before (2MB for RP2040 or 4MB for RP2350).

IMO the default probably doesn't need to be different between the RP2040 and RP2350, but I'm just keeping the same behavior as before.

dpgeorge (approved):
Yes, this looks good now.  I tested NULLBITS_BIT_C_PRO and it now builds.

dpgeorge:
Squashed, rebased and merged in 6bfb83e30aa28e7bbfb0f77f378da05b32574f3d

sfe-SparkFro:
Hmm, this check is not working for me:

https://github.com/micropython/micropython/blob/6bfb83e30aa28e7bbfb0f77f378da05b32574f3d/ports/rp2/CMakeLists.txt#L531

This does seem to work though:

```
if(DEFINED PICO_FLASH_SIZE_BYTES)
```

This statement:

> If `PICO_FLASH_SIZE_BYTES` is not set (actually, it appears to get set to `""`)

appears to be wrong. Not sure if something changed since I last looked at this, but I tested a few boards, and `if(DEFINED PICO_FLASH_SIZE_BYTES)` seems to work fine.

I didn't catch this until now, because my project's firmware size dropped below 4MB shortly after this PR, and only just came back up over 4MB.

dpgeorge:
> Hmm, this check is not working for me:

I think that's because `PICO_FLASH_SIZE_BYTES` evaluates to something like `(2 * 1024 * 1024)`.

PR #17346, opened 2025-05-23
Optimise `machine.time_pulse_us` implementation for code size, and switch esp8266 port to use it

dpgeorge:
### Summary

The esp8266 has a custom implementation of `machine.time_pulse_us()`, which is needed so it can feed the WDT during long waits for a pin change.  It doesn't need to have a separate implementation, instead we can use the common one with an optional hook for a port to do work during the busy wait loops.

This PR:
1. Reworks the common implementation of `machine.time_pulse_us()` to be more like the esp8266 version.  That cuts down code size a little, and improves accuracy a little.
2. Switches esp8266 to use this common implementation.

Also, motivation comes from #16160: this PR should make that PR much simpler (and lower cost) to implement.

### Testing

Tested with a range of square waves from 125Hz to 4kHz on:
- PYBD_SF2: this has a very accurate measurement and accuracy is unchanged with this PR
- RPI_PICO2_W: same as PYBD_SF2, very accurate and unchanged with this PR
- ESP8266: for the frequencies it can handle (up to about 1kHz) this PR is slightly more accurate than before: for example a pulse that is 1000us long, it would measure between 998 and 1000us with an average of 999us.  Now it measures between 999us and 1001us with an average of 1000us.

Also tested the timeout argument, a value of 5000000 still takes 5s and doesn't lead to a WDT timeout on esp8266.

This is an optimisation.  There is no change in the API.  As tested above, the accuracy is at least the same if not better due to better sampling of ticks and pin at the same point in the code.

### DIFF
--- a/extmod/machine_pulse.c
+++ b/extmod/machine_pulse.c
@@ -31,19 +31,34 @@
 #if MICROPY_PY_MACHINE_PULSE
 
 MP_WEAK mp_uint_t machine_time_pulse_us(mp_hal_pin_obj_t pin, int pulse_level, mp_uint_t timeout_us) {
+    mp_uint_t nchanges = 2;
     mp_uint_t start = mp_hal_ticks_us();
-    while (mp_hal_pin_read(pin) != pulse_level) {
-        if ((mp_uint_t)(mp_hal_ticks_us() - start) >= timeout_us) {
-            return (mp_uint_t)-2;
+    for (;;) {
+        // Sample ticks and pin as close together as possible, and always in the same
+        // order each time around the loop.  This gives the most accurate measurement.
+        mp_uint_t t = mp_hal_ticks_us();
+        int pin_value = mp_hal_pin_read(pin);
+
+        if (pin_value == pulse_level) {
+            // Pin is at desired value.  Flip desired value and see if we are done.
+            pulse_level = 1 - pulse_level;
+            if (--nchanges == 0) {
+                return t - start;
+            }
+            start = t;
+        } else {
+            // Pin hasn't changed yet, check for timeout.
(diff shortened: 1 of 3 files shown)


extmod/machine_pulse.c
```diff
@@ -30,20 +30,35 @@
...
+        // Sample ticks and pin as close together as possible, and always in the same
+        // order each time around the loop.  This gives the most accurate measurement.
+        mp_uint_t t = mp_hal_ticks_us();
+        int pin_value = mp_hal_pin_read(pin);
+
+        if (pin_value == pulse_level) {
+            // Pin is at desired value.  Flip desired value and see if we are done.
+            pulse_level = 1 - pulse_level;
+            if (--nchanges == 0) {
+                return t - start;
```
robert-hh, line 46: Shouldn't that be `--nchanges <= 0`. Otherwise the mechanism of waiting for the opposite pulse by setting nchanges to 3 would not work for all start levels. I know that  nchanges was set hard to 2 here, but you discussed making that an optional argument.

dpgeorge:
@robert-hh regarding your comment that I think you deleted: it should be possible to allow `nchanges` to be any value, but the initial `pulse_value` would need to be twiddled like: `pulse_value ^= nchanges & 1`.  Ie if there are an odd number of changes, toggle the initial value.

robert-hh:
Yes, I delete it because at  second though it was wrong. The `nchanges` counter can not underrun.
Tests with a SAMD21 also returned improved accuracy. A 1000µs pulse showed most values being either 999 or 1002, and an average over 100 samples of 1000 +/- 0.15.

projectgus (approved):
Nice cleanup!

PR #17347, opened 2025-05-23
py/scheduler: Allow C scheduler callbacks to re-queue themselves.

### LINKED ISSUE #17246: BLE init failures can cause infinite loop in scheduler

projectgus:
### Summary

Currently if a scheduler callbacks queues a new C scheduled callback, that callback will run in the same schedule pass. This means if a scheduler callback re-queues itself, it can "infinite loop" inside the scheduler.

This change means that any call to `mp_handle_pending()` will only process the callbacks which were queued when `mp_handle_pending()` started running. Therefore any callback which re-queues itself should have that callback handled the next time `mp_handle_pending()` is called.

This is an alternative to #17248 and should remove the need for #17264. Fixes #17246.

This does create potential for some additional callback latency (i.e. if an interrupt fires while a different scheduler callback is running, the interrupt callback is now deferred until the next scheduler run). However the worst-case scheduler latency remains similar (interrupt fires at end of one scheduler pass, has to wait until the next one). I think most MicroPython systems only sparsely call C scheduler callbacks, so this shouldn't be noticeable in most cases.

There is also potential for a misbehaving callback that re-schedules itself continuously to degrade performance of MicroPython (as the callback runs continuously with some allowance for Python code to run), whereas before it would have locked up MicroPython entirely which is more obviously broken.

### Testing

* I adapted the test case @andrewleech added in #17248 for the new behaviour and resubmitted here, so coverage test now includes C scheduler callbacks (testing the new logic).
* Manually ran the rp2 unit tests on RP2_PICO board. As this is a tickless port I felt it had the most potential for an interrupt timing bug to surface due to this change. All passed.

* Possible to manually add anti-recursion checks in callbacks, i.e. approach in #17264, but it can get fiddly quickly.

### DIFF
--- a/py/scheduler.c
+++ b/py/scheduler.c
@@ -88,17 +88,21 @@ static inline void mp_sched_run_pending(void) {
 
     #if MICROPY_SCHEDULER_STATIC_NODES
     // Run all pending C callbacks.
-    while (MP_STATE_VM(sched_head) != NULL) {
-        mp_sched_node_t *node = MP_STATE_VM(sched_head);
-        MP_STATE_VM(sched_head) = node->next;
-        if (MP_STATE_VM(sched_head) == NULL) {
-            MP_STATE_VM(sched_tail) = NULL;
-        }
-        mp_sched_callback_t callback = node->callback;
-        node->callback = NULL;
-        MICROPY_END_ATOMIC_SECTION(atomic_state);
-        callback(node);
-        atomic_state = MICROPY_BEGIN_ATOMIC_SECTION();
+    mp_sched_node_t *original_tail = MP_STATE_VM(sched_tail);
+    if (original_tail != NULL) {
+        mp_sched_node_t *node;
+        do {
+            node = MP_STATE_VM(sched_head);
+            MP_STATE_VM(sched_head) = node->next;
+            if (MP_STATE_VM(sched_head) == NULL) {
+                MP_STATE_VM(sched_tail) = NULL;
+            }
+            mp_sched_callback_t callback = node->callback;
+            node->callback = NULL;
+            MICROPY_END_ATOMIC_SECTION(atomic_state);
+            callback(node);
+            atomic_state = MICROPY_BEGIN_ATOMIC_SECTION();
+        } while (node != original_tail); // Don't execute any callbacks scheduled during this run
     }
     #endif
 
--- a/ports/unix/coverage.c
+++ b/ports/unix/coverage.c
@@ -184,6 +184,18 @@ static void pairheap_test(size_t nops, int *ops) {
     mp_printf(&mp_plat_print, "\n");
 }
 
+static mp_sched_node_t mp_coverage_sched_node;
+static bool coverage_sched_function_continue;
+
+static void coverage_sched_function(mp_sched_node_t *node) {
+    (void)node;
+    mp_printf(&mp_plat_print, "scheduled function\n");
+    if (coverage_sched_function_continue) {
+        // Re-scheduling node will cause it to run again next time scheduled functions are run
+        mp_sched_schedule_node(&mp_coverage_sched_node, coverage_sched_function);
+    }
+}
+
 // function to run extra tests for things that can't be checked by scripts
 static mp_obj_t extra_coverage(void) {
     // mp_printf (used by ports that don't have a native printf)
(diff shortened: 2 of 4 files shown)


projectgus:
@andrewleech Are you able to easily reproduce the issue that #17264 is fixing? Am interested to know if this fixes it OK, I noticed just now that PR links to your commit https://github.com/micropython/micropython/commit/af4a8e9cfc6df2467bb492d4794e377a8d80cb37 which implies the problem is actually with recursive calls to `mp_handle_pending()` itself rather than callbacks re-adding themselves, so that might need a different fix...

andrewleech:
Thanks for this @projectgus yes it's easy to reproduce with a stm32wb55 that's had its rfcore firmware wiped, when trying to start BLE it gets stuck trying to init forever in a loop, the higher level timeout code never runs. 
I'll get a unit set up to test the change.

py/scheduler.c
```diff
@@ -99,6 +100,9 @@ static inline void mp_sched_run_pending(void) {
         MICROPY_END_ATOMIC_SECTION(atomic_state);
         callback(node);
         atomic_state = MICROPY_BEGIN_ATOMIC_SECTION();
+        if (node == original_tail) {
+            break; // Don't execute any callbacks scheduled during this run
+        }
```
dpgeorge, line 105: Can this be rewritten as:
```c
mp_sched_node_t *original_tail = MP_STATE_VM(sched_tail);
while (MP_STATE_VM(sched_head) != original_tail) {
    ...
}
```
?

andrewleech:
Current master:
``` python
MicroPython v1.26.0-preview.148.g49f81d5046 on 2025-05-24; NUCLEO-WB55 with STM32WB55RGV6
>>> 
>>> 
>>> from bluetooth import BLE
>>> BLE().active()
False
>>> BLE().active(True)
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout

```
Note: Ctrl-C did not working here, I couldn't break out of the error loop other than hard reset.

With this PR:
``` python
MicroPython v1.26.0-preview.150.g8b14be1c02 on 2025-05-24; NUCLEO-WB55 with STM32WB55RGV6
>>> from bluetooth import BLE
>>> BLE().active()
False
>>> BLE().active(True)
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
tl_ble_wait_resp: timeout
OSError: [Errno 110] ETIMEDOUT
>>> 

```

andrewleech (approved):

py/scheduler.c
projectgus: No, that way the scheduler always skips the last entry in the list. I think it could be written slightly differently as:

```c
mp_sched_node_t *original_tail = MP_STATE_VM(sched_tail);
if (MP_STATE_VM(sched_head) != NULL) {
    do {
      ..
  } while (MP_STATE_VM(sched_head) != original_tail);
}
```

(Provided the list entries aren't ever modified during the run - apart from appending - which is currently the case.)

This takes the comparison to NULL out of the loop body. Unsure if that'd make much difference in the binary, but I can test it out if you like this way.

projectgus: (Edited the above as I think can remove a pointer lookup by comparing `original_tail` to `NULL` instead of `MP_STATE_VM(sched_head)`.)

projectgus:
Thanks @andrewleech for confirming this fixes the WB55 issue!

py/scheduler.c
projectgus: In a local build, code size for rp2 was the same as the current version after I moved the `original_tail` assignment outside of the loop.

dpgeorge (approved):
Looks good now!

PR #17358, opened 2025-05-25
extmod/modframebuf: Add support for blit'ing read-only data.

dpgeorge:
### Summary

Currently the `FrameBuffer.blit(buf, x, y)` method requires the *buf* argument to be another `FrameBuffer`, which is quite restrictive because it doesn't allow blit'ing read-only memory/data.

This PR extends `blit()` to allow the `buf` argument to be a tuple or list of the form:
```py
(buffer, width, height, format[, stride])
```
where *buffer* can be anything with the buffer protocol and may be read-only, eg `bytes`.

Also, the *palette* argument to `blit()` may be of the same form.

The form of this tuple/list was chosen to be the same as the signature of the `FrameBuffer` constructor (that saves quite a bit of code size doing it that way).

### Testing

A test is added to CI that tests this new capability.

This is an alternative to #15285, which added a new `StaticBuffer` type that was a `FrameBuffer` without any methods and could have a read-only buffer.

The advantage of the approach in this PR is:
- It's smaller code size.
- It doesn't require creating a new `StaticBuffer` object each time the underlying data changes, because one can use a list and change its elements each time the data/width/height changes.  Eg when drawing text using a custom font, a single 4-list object can be created at the start and reused for each character.  That allows drawing text with the font stored in ROM without allocating any extra memory.

The disadvantage of this PR is:
- It's arguably less Pythonic, it would probably be more ergonomic to have a new `StaticBuffer` type rather than a tuple.
- It uses a little more C stack in the blit code, because the source and palette frame buffer objects are now full struct's, not pointers to structs.


projectgus (approved):
Looks good to me. The implementation is a bit fiddly, but I can't think of a simple way to make it less fiddly!

PR #17363, opened 2025-05-26
rp2/machine_timer: Support hard IRQ timer callbacks.

arachsys:
### Summary

When using `Timer` on the pyboard (and stm32 more generally), the callback is run in hard IRQ context and is reliably punctual.

On a Pi Pico 2 board, I was surprised to find that the rp2 port instead queues timer callbacks via `mp_sched_schedule()` which means they are subject to GC delays and potentially unbounded jitter.

This probably makes sense as an 'easy mode' default, allowing allocation and less disciplined code in a timer callback, but having it as the only option makes tight timing more awkward on rp2 than it is on stm32. Searching discussions, some have worked around the unpunctual timers by repurposing a PWM output with a hard trigger IRQ. A PIO state machine in an `irq()` loop with a hard IRQ handler would also do the job, but in each case we're wasting limited resources and adding complexity.

Rather than hacking around the problem once again, it feels sensible to add a (non-default) option to make the rp2 timers dispatch callbacks in hard IRQ context in the same way stm32 timers do.

Add a `hard=` parameter to the `Timer()` constructor and init, matching the `hard=` parameter on `Pin.irq()` and `rp2.DMA.irq()` and `rp2.PIO.irq()`. When set true, this dispatches the callback in hard IRQ context in the same way as the other hard IRQ handlers for rp2 and the timer callbacks for stm32, taking the same precautions to lock the scheduler and GC, catch unhandled exceptions, and disable the callback on unhandled exceptions.

Does this make sense? I've tested this reasonably heavily in my own application without uncovering any problems (see below for a demo) but would be grateful for any thoughts. For example, should this patch make the behaviour configurable in the other ports at the same time? Obviously, I'll update the patch to update the documentation to match wherever it ends up as well.

### Testing

To demonstrate the GC jitter with the default soft callbacks on a Pi Pico or Pico 2, try:

```python
from machine import Pin, Timer

pin = machine.Pin(0, machine.Pin.OUT)
timer = Timer(freq = 1000, mode = Timer.PERIODIC, callback = lambda t: pin.toggle())

x = 0.3
while True:
    x = 4 * x * (1 - x)
```
with a scope on pin GP0. The fp loop allocates and gcs heavily, so you'll see the 500Hz square wave often lengthen out from 1 ms to 2-4 ms when the callback is delayed.

With this patch and the new hard callbacks, this effect disappears and the jitter is down in the single digit microseconds:

```python
from machine import Pin, Timer

pin = machine.Pin(0, machine.Pin.OUT)
timer = Timer(freq = 1000, mode = Timer.PERIODIC, callback = lambda t: pin.toggle(), hard = True)

x = 0.3
while True:
    x = 4 * x * (1 - x)
```

I've tested with multiple concurrent timers of disjoint frequencies up to 20kHz (pushing towards 100% utilisation in interrupt handling) on a variety of different RP2350 boards.


Josverl:
> adding a port-specific tests/ports/rp2/rp2_machine_timer.py as part of this patch

In order to ensure coverage, and make sure that the feature remains operational in the future, an additional test would be needed. 
I tend to use separate commits for code, tests and documentation. 

I would start with a port specific test, but possibly that could be lifted to a port-agnostic test if/when other ports opt-in to this new keyword.

arachsys:
I've added tests for both this hard= option and the existing rp2-specific tick_hz option, documented the option of soft- vs hard-IRQ, and rebased against the current HEAD of the master branch.

arachsys:
Rebased against the current HEAD of the master branch again.

dpgeorge:
Thanks for the contribution, this is a very nice enhancement!

The Timer class is a bit inconsistent across ports, as you have noticed.  We can't fix that here, but at least we can try not to make things any worse.

One question is whether `Timer` should have an `irq()` method to match other classes like `Pin`.  Then the `hard` argument would be easy to add.  Well, the fact that the `Timer` constructor already takes a `callback` argument, and usually the only thing you want to do (or can do) with a timer is to have it periodically call a callback, I guess adding ` hard` argument alongside the `callback` argument makes the most sense.

Overall, I think this PR has got it right in terms of API decision, ie simply add the `hard` argument to the constructor/init method.  Eventually that can be extended to other ports pretty easily.  And then one day we can make things fully consistent across ports and have a common test.  But for now this PR is a good step forward.

docs/rp2/quickref.rst
```diff
@@ -135,6 +135,11 @@ Use the :mod:`machine.Timer` class::
     tim = Timer(period=5000, mode=Timer.ONE_SHOT, callback=lambda t:print(1))
     tim.init(period=2000, mode=Timer.PERIODIC, callback=lambda t:print(2))
 
+By default, timer callbacks run as soft IRQs so they can allocate but
+are prone to GC jitter and delays. Pass ``hard=True`` to the ``Timer()``
+constructor or ``init()`` method to run the callback in hard-IRQ context
+instead.
+
```
dpgeorge, line 143: Might be worth adding a link/reference to `_isr_rules`.

ports/rp2/machine_timer.c
```diff
@@ -40,13 +41,36 @@ typedef struct _machine_timer_obj_t {
...
+        mp_sched_lock();
+        gc_lock();
+        nlr_buf_t nlr;
+        if (nlr_push(&nlr) == 0) {
+            mp_call_function_1(self->callback, MP_OBJ_FROM_PTR(self));
+            nlr_pop();
+        } else {
+            // Uncaught exception; disable the callback so it doesn't run again.
+            self->mode = TIMER_MODE_ONE_SHOT;
+            mp_printf(MICROPY_ERROR_PRINTER, "Uncaught exception in timer callback\n");
```
dpgeorge: Please use lowercase "uncaught" (matches stm32).

tests/ports/rp2/rp2_machine_timer.py
```diff
@@ -0,0 +1,21 @@
...
+# Test the rp2-specific adjustable tick_hz and hard/soft IRQ handlers
+# for both one-shot and periodic timers.
+
+modes = {Timer.ONE_SHOT: "one-shot", Timer.PERIODIC: "periodic"}
+kinds = {False: "soft", True: "hard"}
+
+for mode in modes:
+    for hard in kinds:
+        for hz in 1000, 2000:
+            timer = machine.Timer(
```
dpgeorge: You need to use `Timer` here (`machine` is not imported!).

tests/ports/rp2/rp2_machine_timer.py
```diff
@@ -0,0 +1,21 @@
...
+                period=2,
+                mode=mode,
+                hard=hard,
+                tick_hz=hz,
```
dpgeorge: It's quite confusing what the actual frequency is of this timer.  IMO this test should separate out the test for changing `ticks_hz` into a separate part of this file.

For this hard/soft one-shot/periodic test, I suggest just using `period` to set the period (eg to 1 and 2 ms).

tests/ports/rp2/rp2_machine_timer.py
```diff
@@ -0,0 +1,21 @@
...
+                callback=lambda t: print("callback", modes[mode], kinds[hard], hz),
+            )
+            sleep_ms(5)
```
dpgeorge: This sleep duration is a little unreliable, because sometimes the IRQ fires 4 times and other times it fires 5 times, making the test fail.

You can test this with:
```
./run-tests.py -t a0 --via-mpy -d ports/rp2
```
That's a bit more of an involved test, going via compiled .mpy files.

I suggest making the sleep not a multiple of the period, so it's more reliable to get the exact number of desired callbacks.

robert-hh:
> And then one day we can make things fully consistent across ports

Quite a few ports use Softimer. At a quick glance it seems to be alif, mimxrt, renesas, samd and stm32.

dpgeorge:
> Quite a few ports use Softimer.

We can add `hard` to those ports (in a separate PR).  But it needs to remain backwards compatible, eg on stm32 it would need to be `hard=True` by default.

tests/ports/rp2/rp2_machine_timer.py
arachsys: Yes, that's much clearer. I think the stray `period = 2` was carelessly left from an earlier version of the test which just used a fixed period of 2.

tests/ports/rp2/rp2_machine_timer.py
arachsys: Originally I had a fixed period of 2ms here which was chosen not to divide the 5ms sleep. Not sure why I switched to hz instead of 2ms, 4ms, but you're quite right, it's better with period = 2, 4.

docs/rp2/quickref.rst
arachsys: Ooh yes, good plan.

arachsys:
I've pushed a corrected version with all the fixes you spotted - thanks!

If you're happy with the API choice, I'm very happy to have a go at a follow-up PR generalising this to the other ports, preserving the current default for hard vs soft in each case?

tests/ports/rp2/rp2_machine_timer.py
```diff
@@ -0,0 +1,20 @@
...
+
+# Test the rp2-specific adjustable tick_hz and hard/soft IRQ handlers
+# for both one-shot and periodic timers.
+
+modes = {Timer.ONE_SHOT: "one-shot", Timer.PERIODIC: "periodic"}
+kinds = {False: "soft", True: "hard"}
+
+for mode in modes:
+    for hard in kinds:
+        for period in 2, 4:
```
dpgeorge, line 12: This is technically not testing period=4 for periodic mode, because it's only executed once in the 5ms waiting window.

I suggest increasing the waiting window to 9ms.

arachsys: Fair point! I'll push a version that does this.

dpgeorge (approved):

And thanks for making clean commits with good git commit messages!  Makes it easier to just merge.

dpgeorge:
> If you're happy with the API choice, I'm very happy to have a go at a follow-up PR generalising this to the other ports, preserving the current default for hard vs soft in each case?

Yes, feel free to do that!

Issue #3935, opened 2018-07-10
RFC: defining an I2C slave class and behaviour


dpgeorge:
I2C slave mode is an often-requested feature.  See eg #3680 and #3858, and #3114 for a previous discussion that was tied to esp8266 implementation.  It would be good to define an I2C slave class and how it would behave, and then implementation for various ports could follow.

stm32 recently got some I2C slave code and the interface is in this header: https://github.com/micropython/micropython/blob/master/ports/stm32/i2cslave.h .  In this C implementation there are 4 external functions that should be provided and which are essentially callbacks from the low-level I2C slave driver:
```c
// the slave's address was matched, and the master requested a read/write (via rw param)
// return 0 to ack address or 1 to nack it
int i2c_slave_process_addr_match(int rw);

// master sent a byte, slave received it (val param)
// return 0 to ack the byte or 1 to nack it
int i2c_slave_process_rx_byte(uint8_t val);

// master stopped sending data via a stop or repeated start condition
void i2c_slave_process_rx_end(void);

// master requests data, slave needs to send it
// return the byte to send back to the master
uint8_t i2c_slave_process_tx_byte(void);
```

This describes a very low level interface, at the byte level.  With a bus at 100kHz bytes are transferred at about 10kHz.  Python code might be able to respond at this rate to individual bytes, but only on fast hardware.  And a 400kHz SCL will make it more difficult to respond to individual bytes.  So, while the above interface allows to implement (almost) any I2C device, it's not really a solution for MicroPython.

We need to define a I2C slave API at a higher level than bytes.  A callback for each full bus transaction would be the way to go.

The most obvious I2C behaviour to implement is a memory-mapped device, eg `I2CSlaveMemory`, where the master send the memory address (1 or 2 bytes) then either reads or writes the data.  For example:
```python
mem_buf = bytearray(32)
i2c_mem = I2CSlaveMemory(scl=scl, sda=sda, addr=67, mem_addr_len=8, mem=mem_buf)
# master can now access the mem_buf array
while True:
    mem_buf[0] += 1
    time.sleep(1)
```
For simple cases that doesn't even need a callback.  But `I2CSlaveMemory` would support callbacks like `memory_read(offset, len)` and `memory_written(offset, len)` to signal when the master did a transaction.

Other types of high-level devices include:
[...]

PR #17365, opened 2025-05-27
Add `machine.I2CTarget` implementing a generic I2C device

dpgeorge:
### Summary

(Edit: this summary was updated to reflect the most recent state of this PR.)

This PR implements a generic I2C target/peripheral/"slave" device.  It can work in two separate modes:
- a general device with interrupts/events/callbacks for low-level I2C operations like address match, read request and stop
- a memory device that allows reading/writing a specific region of memory (or "registers") on the target I2C device

The class is called `machine.I2CTarget`.  To make a memory device is very simple:
```py
from machine import I2CTarget

mem = bytearray(8)
i2c = I2CTarget(addr=67, mem=mem)
```
That's all that's needed to start the I2C target.  From then on it will respond to any I2C controller on the bus, allowing reads and writes to the `mem` bytearray.

You can also register to receive events.  For example to be notified when the memory is read/written:
```py
from machine import I2CTarget

def irq_handler(i2c_target):
    flags = i2c_target.irq().flags()
    if flags & I2CTarget.IRQ_END_READ:
        print("controller read from target starting at", i2c_target_memaddr)
    if flags & I2CTarget.IRQ_END_WRITE:
        print("controller wrote to target starting at", i2c_target_memaddr)

mem = bytearray(8)
i2c = I2CTarget(addr=67, mem=mem)
i2c.irq(irq_handler)
```

Instead of a memory device, an arbitrary I2C device can be implemented using all the events (see docs).

This is based on the discussion in #3935.

An implementation is provided for:
- rp2 (direct register access)
- alif (direct register access, same I2C hardware as rp2)
- stm32 (direct register access)
- esp32s2/esp32s3 (using the IDF, does not support all events)
- zephyr (using zephyr i2c_target API, all events supported)
- mimxrt
- samd

### Testing

An extmod test is added, and also a multitest.

The tests pass RPI_PICO, RPI_PICO2_W, PYBV10, PYBD_SF6, ALIF_ENSEMBLE.

This is a very simple implementation, but it works and is probably enough for most use cases.  There are lots of things that could be enhanced:
- [x] add Python callbacks to notify when the I2C controller reads/writes memory
- [x] implement 16-bit wide memory addressing (currently restricted to 8-bit addresses)
- [ ] implement other I2C targets, eg FIFO, or generic
- [ ] add support for `asyncio`, eg polling the device for events


chrismas9:
Interesting timing. I have an immediate need for this. I have just written an implementation of https://github.com/micropython/micropython/issues/3935 in MicroPython. It implements callbacks and 16-bit addressing, but needs to be called regularly in the main loop to service I2C requests from the controller. It has some other limitations, but is enough for my applications.

I see this failed on STM32F091. I am using STM32L432. I will try it and report back.

Callbacks would be useful, but I can live without them. My inputs need to be debounced or filtered and that logic can update the read registers. For write registers I will maintain a backup copy and diff them periodically to update outputs.

Would the `memory_read `callback be called before the I2C reply so that the memory can be updated on demand, eg from a pin, ADC, etc?

Is it necessary to name it other than `I2CTarget `? If it were named `I2CTarget `now, with maybe `mode=memory` it would allow applications to keep working after a full featured `I2CTarget `was added to MicroPython.

kwagyeman:
@dpgeorge - Alif support? :)

ports/rp2/machine_i2c_target.c
```diff
@@ -0,0 +1,185 @@
...
+#endif
+#endif
+
+// SDA/SCL on even/odd pins, I2C0/I2C1 on even/odd pairs of pins.
+// TODO: should be in a common i2c header
+#define IS_VALID_SCL(i2c, pin) (((pin) & 1) == 1 && (((pin) & 2) >> 1) == (i2c))
+#define IS_VALID_SDA(i2c, pin) (((pin) & 1) == 0 && (((pin) & 2) >> 1) == (i2c))
+
+static machine_i2c_target_data_t i2c_target_data[4];
+
```
robert-hh, line 74: It seems that the code only uses 2 of these data objects.

robert-hh:
I could start to implement this feature with the MIMXRT port. The NXP lib has support for the slave mode.

chrismas9:
I would prefer callbacks to a flag as they allow for a much  faster response for I2C reads if the register is updated in the callback.  How would a flag work? Would the read function block until the flag is polled, the register updated and the flag is cleared, or would it immediately reply with stale data? The other problem, which I have with my python library, is that the Controller timeout has to be larger than the slowest poll time. I have use cases where such latency would be a problem, eg measuring motor current in a servo controller

chrismas9:
I have added all STM32 families. I will submit the changes after testing them all on Eval boards.

chrismas9:
I was hoping for a blocking read callback that could update a register with fresh data before the reply using clock stretching during the callback. I am making a series of modules which should behave like standard ICs. Most ADCs respond to a read with fresh data, either by clocking a SAR with SCL or using clock stretching until a conversion is finished. Any background loop that has to continuously update registers will result in reading data that is probably 10 to 100ms out of date. I'm not sure it's very useful to have a read callback after the event.

Write callbacks after a write are very useful as long as they contain the range of registers written.

robert-hh:
Thanks. I have seen the test and will run it. And yes, I expect the API to change, but that can be adapted easily.

robert-hh:
There were some issues in the wrap tests, that are fixed now. After adapting `machine_i2c_target_memory.py` with the settings for this board, all tests pass. The code can still be optimized somewhat and tested at the various board types, especially the slower i.mx1011 variants.

dpgeorge:
> I was hoping for a blocking read callback that could update a register with fresh data before the reply using clock stretching during the callback.

This is an important question to discuss further.  With a read transaction (point of view from the controller) we have:
1. generate start condition
2. send address byte, with direction flag (read)
3. read byte from the target, so the target must drive SDA
4. generate stop condition

If the target is not ready in step 3 to send the data (eg an ADC that needs to be sampled) then the target could stretch the clock by holding SCL low.  Then release it when it has the new data ready.

If we want to support such clock stretching then the API and implementation will be quite different, because we need a way for the Python code to signal that it wants to stretch the clock (stall the bus), and then indicate when it is ready to unstall the bus.

There are real devices out there that do this clock stretching, eg ADS7128 (and it seems that device stretches SCL for quite some time when it's averaging many readings).  Other devices like ADS1115 require you to poll a register to see when the ADC conversion is ready.  And something like MCP3021 will do the ADC conversion during the first few bits of the data read (and those bits always read 0) so don't need to stretch SCL.

Looking at Zephyr's I2C target API, it does not support clock stretching, so it's not possible to implement a clock stretching target with Zephyr.

Do we need the MicroPython I2C target API to be able to do clock stretching?

Other considerations:
- clock stretching for writes to the target; this could be used for example to implement flow control when a lot of data is written out to the target (it holds SCL low if it's not ready to receive more data)
- support atomic reads from memory device, eg if the I2C controller is reading a 16-bit value from the I2C target, you don't want the I2C target to update that memory in the middle of the read or else the upper and lower bytes of the 16-bit value will be out of sync
- support atomic writes

Things can get complicated pretty quickly when looking at all these things.

kwagyeman:
Clock stretching is a very undesirable behavior of I2C.

Do not replicate that. It makes the bus unreliable as it means I2C transfers are not accomplished in fixed time. This kills any MCUs ability to meet a schedule with an I2C shared with multiple devices.

robert-hh:
>  Teensy already needs D0/D1 and D2/D3 and D11/D12 connected 
With the test of PAM and UART, these Pins are pairwise connected, D0 to D1, D2 to D3 and D11 to D12. For I2C, SDA (A4) and and SCL(A5)  have to be connected each to a GPIO pin. So I suggest to use D4 and D5. Could be as well A2 and A3. At the Teensy, SCL and SDA need an external pull-up resistor for a reliable test. The internal pull-up is not sufficiently strong.

robert-hh:
@dpgeorge The MIMXRT implementation is at a stable state with the feature set matching the RP2. It works fine so far.

> @chrismas9 I was hoping for a blocking read callback that could update a register with fresh data before the reply using clock stretching during the callback.

Looking at the timing constraints this feature is hard to achieve. Hardware which allows that uses often a double buffer scheme, such that when a value is requested the most recent ADC value is latched to the transmit buffer. Or they need a dedicated command to start a new sampling and offer a status bit that has to be polled, or an INT signal.
What's less critical is a callback that's called after receiving the STOP condition. With that you may be able to emulate a readfrom_mem() by a combo of writeto() and readfrom() calls.

robert-hh:
There is an implementation for SAMD at https://github.com/robert-hh/micropython/tree/samd_i2c_target. Works well. Tested so far with SAMD51. Tests with SAMD21 follow.

robert-hh:
@dpgeorge I'm still unsure about the wrapping behavior for writing and reading. I could not find a source where this is defined as to be expected. Do you have a reference?

For writing controller->target the target can signal if the memory overflows by sending NAK. Then the controller would stop sending. For receiving there is no such mechanism. So the alternative to wrapping would be sending dummy data like 0x00 or 0xff.

robert-hh:
I looked into other ports, whether they can support the I2CTargetMemory class. 
- The NRF port it looks promising. The library provides suitable functions. 
- ESP8266 does not provides I"C hardware support at all.
- The Renesas port seems to have I2C target mode hardware, but the HAL layer does not offer interfaces for using it.
- The ESP32 situation is crooked for several reasons.
 
  + The ESP32 API has three versions of I2C support. 1. The legacy API, which is currently used for I2C controller support. 2. The ng API Version 1. 3, The ng API version 2. Option b) is declared as getting expired with idf 6.x. The ng API V2 is available starting version 5.4. So using is at the moment is not possible. The legacy API and ng API must not be used at the same time. So if that is going to be used, the I2C class has to be updated.
  + Implementing the class  I2CTargetMemory for trial it turned out that it cannot properly support the target memory mode. The API offers two callback functions or receive and transmit. Each of them is called AFTER the respective bus transactions is finished. That makes it possible to support the writeto_mem() method properly, but fails to support readfrom_mem(). The only option which could be supported is getting the proper content from the previous state of the memory object. After it has been written, one has to call readfrom_mem() twice to get the proper content. It may be useful to have a I2CTargetMemory.write() method which fills the transmit buffer with data from a provided object.
  + One has to declare a suitable large size for the temporary buffer. Otherwise, if on writeto_mem() the sent data is longer than the declared buffer, the amount of data actually provided to the callback is erratic. On readfrom_mem() the reading does not wrap. Instead for the excess data 0xff is sent.

dpgeorge:
Thanks @robert-hh for working on this for the other ports, that's very helpful!

Right now I'm reworking the API here to take into account the feedback, mainly to provide IRQs/callbacks for various I2C target events.  After that's done we can integrate your changes.

> I'm still unsure about the wrapping behavior for writing and reading. I could not find a source where this is defined as to be expected. Do you have a reference?

It's not really defined as a standard thing, it's just that a lot of devices will wrap around if you read beyond the end of their register range.  Eg I think the MMA on the PYBv1.x does this?

dpgeorge:
I've now updated this PR:
- renamed the class to `machine.I2CTarget`
- now supports creating either a memory device, or an arbitrary device
- now supports IRQs for various events

It's not finished but the included test is comprehensive and passes on stm32, rp2 and zephyr.

@robert-hh are you able to update the ports you worked on so far, to match the new code here?

robert-hh:
@dpgeorge Will do that.

docs/library/machine.I2CTarget.rst
```diff
@@ -0,0 +1,78 @@
...
+Example usage::
+
+    from machine import I2CTarget
+
+    # Create the backing memory for the I2C target.
+    mem = bytearray(8)
+
+    # Create an I2C target.  Depending on the port, extra parameters
+    # may be required to select the peripheral and/or pins to use.
+    i2c = I2CTargetMemory(addr=67, mem=mem)
```
robert-hh: Shouldn't that be: `i2c = I2CTarget(addr=67, mem=mem)`

dpgeorge: correct, should be `I2CTarget`

robert-hh:
After being kept busy by my family for a few days, I found now some time to adapt the code, using the RP2 port as reference. I have a question on the implementation:
- The function `mp_machine_i2c_target_read_bytes()` can deal with values of len > 1, but not `mp_machine_i2c_target_write_bytes()`. Since the whole execution flow from hard interrupt to actually moving the data got longer, is is as well acceptable to implement `mp_machine_i2c_target_read_bytes() ` with a length of 1?
- Using the I2C mem functions thing look fine. But when using i2c.writeto() and i2c.readfrom(), always one byte is skipped. for instance, with the object set tp "01234567", calling`i2c.readfrom(67, 1)` returns varying data with one value skipped. Actually I do not know what to expect.
```
>>> i2c.readfrom(67, 1)
b'0'
>>> i2c.readfrom(67, 1)
b'2'
>>> i2c.readfrom(67, 1)
b'4'
>>> i2c.readfrom(67, 1)
b'6'
```

hoihu:
> Do we need the MicroPython I2C target API to be able to do clock stretching?

I would be in favour of that. 

If the "Target" device is an "intelligent device" it's very likely it has other tasks to do as well (usb, FS, network, sensor measurements...) -> It might not be ready all the time to serve the (read) interrupt since it's busy with other high priority tasks. We actually use such a setup in some of our products. I believe it's pretty common actually.

> support atomic reads from memory device, eg if the I2C controller is reading a 16-bit value from the I2C target, you don't want the I2C target to update that memory in the middle of the read or else the upper and lower bytes of the 16-bit value will be out of sync

oh yes, that's a concern I know of first hand ;) So yes - such a mutex or similar would be really nice to have. I tend to think of that as a shared RAM (via I2C, and directly on the MCU). With all it's pros and cons.

Other Comments
- blocking wait(s) on the controller side: would it be worth to add **async** API for read/write operations on the controller?
- Instead of "Target", perhaps use "Peripheral"? I2CPeripheralMemory?

dpgeorge:
> * The function `mp_machine_i2c_target_read_bytes()` can deal with values of len > 1, but not `mp_machine_i2c_target_write_bytes()`

I may need to change this to only deal with one byte at a time.  It depends on how it all turns out with the other ports.

> * is is as well acceptable to implement `mp_machine_i2c_target_read_bytes() ` with a length of 1?

Yes, that's fine to make this function deal only with one byte.

> * But when using i2c.writeto() and i2c.readfrom(), always one byte is skipped. for instance, with the object set tp "01234567", calling`i2c.readfrom(67, 1)` returns varying data with one value skipped

I will look into this.  But for now that should not impact the port-specific implementation.  As long as the port can pass the test that's added in this PR, it should be fully correct.

robert-hh:
The ports for MIMXRT and SAMD are adapted now. It's the last commit in https://github.com/robert-hh/micropython/tree/mimxrt_i2c_target and the last two commits in https://github.com/robert-hh/micropython/tree/samd_i2c_target. That includes an adapted test for MIMXRT and SAMD. Both ports pass the supplied test. 
I'm not sure how to proceed with ESP32. Useful slave mode requires new I2C support V2, which needs IDF v5.4.1 at least, and the master mode I2C has to be rewritten as well. And even then the behavior is different to the other ports.
I can start to adapt it for the NRF port. At least the library API set looks promising.

chrismas9:
With regard to clock stretching I would still like to see an optional callback `update_mem=update_mem` called during clock stretching to update the buffer before it is read and sent to the controller. I would use this for fast updates such as reading a GPIO. This would also solve the atomic read problem as the driver would be stalled during the callback. If you don't want clock stretching use `update_mem=None`.

With regard to atomic multi-byte read and write what about adding the `readfrom_mem` and `writeto_mem` methods from `machine.I2C`? Buffer access method would be the same for controller and target and I2CTarget can handle the interleaving of reads and writes by controller and target.

robert-hh:
> With regard to clock stretching I would still like to see an optional callback update_mem=update_mem called during clock stretching to update the buffer before it is read and sent to the controller.

WIth the actual version there are quite a few IRQ events that can be set: `'IRQ_ADDR_MATCH', 'IRQ_END', 'IRQ_READ_REQ', 'IRQ_STOP', 'IRQ_WRITE_REQ'`. These could be used to update the buffer.

Edit: In order to grant the time for a callback execution, clock stretching is enabled at the SAMD, RP2 and MIMXRT ports. I assume as well for the other ports.

robert-hh:
@dpgeorge There seems to be a hiccup with the RP2 port and the value of data->status. When calling e.g. `i2c.readfrom_mem(67, 0, 8)` at the controller, and a callback is active for  IRQ_ADDR_MATCH, then this callback is  called for every single data byte, not only for the first. Timing below. The gap between the data bytes is the time, when the callback is active, and it shows b.t.w., that the RP2 uses clock stretching.

grafik

Edit: That is caused by the Restart, which sets the state back to IDLE.

extmod/machine_i2c_target.c
```diff
@@ -0,0 +1,356 @@
...
+#include "shared/runtime/mpirq.h"
+
+enum {
+    I2C_TARGET_IRQ_ADDR_MATCH = 1 << 0,
+    I2C_TARGET_IRQ_READ_REQ = 1 << 1,
+    I2C_TARGET_IRQ_WRITE_REQ = 1 << 2,
+    I2C_TARGET_IRQ_END = 1 << 3,
+    I2C_TARGET_IRQ_STOP = 1 << 4, // should be able to support this on all targets (rp2 needs custom code though), but maybe can't support addr_match?
+
+    I2C_TARGET_IRQ_MEM_ADDR_MATCH = 1 << 5, // needed?
```
robert-hh: These look not necessary. The ..START ones are at the same time that the ADDR_MATCH, and the ..END ones are related to STOP or END:

extmod/machine_i2c_target.c
```diff
@@ -0,0 +1,356 @@
...
+        }
+    } else {
+        // Have a buffer.
+        uint8_t val;
+        mp_machine_i2c_target_read_bytes(self, 1, &val);
+        if (data->state == STATE_IDLE) {
+            // TODO allow N bytes for address, with N=0 allowed
+            data->state = STATE_MEM_ADDR_SELECTED;
+            data->mem_addr = val % data->mem_len;
+            event(data, I2C_TARGET_IRQ_MEM_ADDR_MATCH, data->mem_addr, 0);
```
robert-hh: Seems to cause a duplicate call, because it follows the address byte handling, which already calls the event handler for I2C_TARGET_IRQ_MEM_ADDR_MATCH.

extmod/machine_i2c_target.c
```diff
@@ -0,0 +1,356 @@
...
+    bool event_handled = event(data, I2C_TARGET_IRQ_READ_REQ, 0, 0);
+    if (data->mem_buf == NULL) {
+        if (!event_handled) {
+            // No data source, just write out a zero.
+            uint8_t val = 0;
+            mp_machine_i2c_target_write_bytes(self, 1, &val);
+        }
+    } else {
+        // Have a buffer.
+        if (data->state == STATE_MEM_ADDR_SELECTED) {
```
robert-hh: The additional calls for ADDR_MATCH events disappear, if the test is:
`        if (data->state == STATE_MEM_ADDR_SELECTED || data->state == STATE_IDLE) {`
This is the smallest change I could find, but may not be the best method to fix the state machine here.
But that seems to be only a problem of the RP2 port, because it does not have a specific ADDR_MATCH event.
The event MEM_ADDR_SELECTED seems to be independent from the other states and it must survive a STOP and RESTART. SO maybe its better to put it into a separate flag or drop it. It is anyhow only needed to make a difference between the ADDR_MATCH  event and the WRITE_START event.

robert-hh:
When setting up a test with a MIMXRT1020 as controller and RP2 as target, the RP2 had a data loss once longer messages were sent from Controller to Target, causing data to get lost. That is with a minimal callback on ADDR_MATCH, freq=400_000. Without a callback the Target is sufficiently fast. Both is not a surprise. Interestingly, while upon send the Target uses clock-stretching, it does not do so on receive. But maybe there is a configuration which enables it upon FIFO overrun.
Edit: This setting seems to be available. and it is set!
Edit 2: This data loss happens already with 8 byte messages, while the FIFO size is 16 bytes. That explains why the mechanism for holding the bus does not trigger.
Edit3: When sending  messages longer than 16 byte clock stretching happens. Data loss is to be analyzed.

dpgeorge:
I have now implemented `I2CTarget` on the alif port.  The test included here passes, all IRQs are supported in hard IRQ mode.

dpgeorge:
> Interestingly, while upon send the Target uses clock-stretching, it does not do so on receive. But maybe there is a configuration which enables it upon FIFO overrun.

I tested clock stretching on the alif port.  The I2C hardware on alif has a 32-byte deep FIFO, for both TX and RX.  So you don't see clock stretching for controller -> target with just a few bytes being sent by the controller, because the alif target is queuing these bytes in its RX FIFO.  OTOH I did observe clock stretching for target -> controller: as long as the TX FIFO is empty the alif will stretch the clock.  So you don't need to respond immediately to `IRQ_READ_REQ`, you can delay that indefinitely and it'll stretch the clock until `i2c_target.write(...)` is called which fills the TX FIFO.

I guess all the ports are going to behave differently here, depending on whether they have a FIFO or not, and whether they can do clock stretching.

I'm not sure how useful these events/IRQs are going to be in practice to make real I2C targets/devices.  Still needs investigation.

robert-hh:
That matches my observation: 
- clock stretching appears for target -> controller, verified for RP2, MIMXRT and SAMD
- clock stretching may appear for controller -> target, once the RXFIFO get's full. Seen at RP2. But a different limit may hit before, and that is queuing of interrupts, if the received data is not processed sufficiently fast. I may have seen that with the RP2 port. For SAMD and MIMXRT with a ADDR_MATCH event handler the clock is held low as long as the event handler is active. That's good. Further tests needed.

robert-hh:
> I'm not sure how useful these events/IRQs are going to be in practice to make real I2C targets/devices. 

For me the ADDR_MATCH and STOP/END events seem to be the most useful ones. Maybe ADDR_MATCH could already be split into READ and WRITE. That is known already at tha time of address match. I2C_TARGET_IRQ_READ_REQ and I2C_TARGET_IRQ_WRITE_REQ seems less useful for being tested and triggered with every byte. That's a lot of overhead.

ports/rp2/machine_i2c_target.c
```diff
@@ -106,8 +106,8 @@ static void i2c_target_handler(i2c_inst_t *i2c, i2c_slave_event_t event) {
             if (data->state == STATE_IDLE) {
                 machine_i2c_target_data_addr_match(data, true);
             }
-            //i2c_write_byte_raw(i2c, 0x12);
-            //break;
+            // i2c_write_byte_raw(i2c, 0x12);
+            // break;
```
robert-hh: Are these commented out lines needed? I guess not.

robert-hh:
There is still the unexpected behavior that on reading from the Target the ADDR_MATCH handler is called for every byte.
Reason: The condition in line 165 `if (data->state == STATE_MEM_ADDR_SELECTED)` is never true, since between setting the state to `STATE_MEM_ADDR_SELECTED` and getting to that line a restart happens in the protocol, setting state to `STATE_IDLE`.
Edit: It should be sufficient to make the test in line 165 for `STATE_IDLE`.

extmod/machine_i2c_target.c
robert-hh: Just ` if (data->state ==  STATE_IDLE)` is sufficient.

dpgeorge:
> There is still the unexpected behavior that on reading from the Target the ADDR_MATCH handler is called for every byte.

I don't see this on the rp2 port.  Using RPI_PICO_W and running the test in this PR, the `TestIRQ` class tests this.  At the moment those tests don't full pass, but they show only one ADDR_MATCH per read or write.

Well, I need to rework the rp2 implementation to use a lower-level hardware interface, instead of the higher-level `pico_i2c_slave` component.  That's needed to get events for a simple scan (0-byte write) transaction.

robert-hh:
> I don't see this on the rp2 port. Using RPI_PICO_W,

Sorry. My old eyes. The second call is for I2C_TARGET_IRQ_MEM_ADDR_MATCH, not I2C_TARGET_IRQ_ADDR_MATCH. Anyhow I see the callback being called twice and data loss. But that's a different problem. It happens when sending data from a separate device at freq=400_000. With freq=100_000 all is fine. And of course it does not happen in the test script set-up, where the data is sent by SoftI2C and being at the same CPU, there cannot be any timing conflict between controller and target.

dpgeorge:
A nice surprise: the I2C peripheral for alif and rp2 is the same Synopsis silicon IP.  So they can have the same low-level implementation and their semantics will match.

dpgeorge:
I have now reworked the rp2 I2CTarget implementation.  It's now the same as the alif one, and passes the included tests.

ports/rp2/machine_i2c_target.c
dpgeorge: Removed.

robert-hh:
Downloaded and tested with a RP2 Pico and a second board as Controller and a single callback for ADDR_MATCH. Runs fine at freq=100_000. Works fine with freq=400_000 for data requested from the slave. Behaves weird at freq=400_000 when data is sent from Controller:
- the callback is called twice instead of once.
- single bytes of the transfer are lost. Either the old data i still in the target memory at that place, or a NULL byte is inserted.
No improvement with that aspect compared to the previous version.
Without a callback data is transferred properly.

dpgeorge:
> * the callback is called twice instead of once.
> 
> * single bytes of the transfer are lost. Either the old data i still in the target memory at that place, or a NULL byte is inserted.

Thanks for testing.  I can also see the same issue (using a special test I wrote that uses software SPI for the controller and can detect clock stretching).

Should be fixed now.  I had to enable clock stretching for the RX path.  And also fix out-of-order events, when the STOP occurs before the RX FIFO is fully drained.

kwagyeman:
Hi Damien, will the current API be usable for replacing `pyb` in this case? https://github.com/openmv/openmv/blob/master/scripts/libraries/rpc.py#L502

It looks like fo get_bytes() I can use the read_into() in a loop while tracking time to receive data... However, for put_bytes(), I can't do the same kind of thing because mp_machine_i2c_target_write_bytes() seems to always write 1 byte even if there's no space in the FIFO on alif for example.

Can you adjust mp_machine_i2c_target_write_bytes() to return 0 if there's no space in the FIFO? Otherwise it looks like some type of interrupt driven thing is necessary to write data out. Also, should mp_machine_i2c_target_write_bytes() do more than 1 byte at a time?

dpgeorge:
> It looks like fo get_bytes() I can use the read_into() in a loop while tracking time to receive data...

Yes, you might be able to do that.  Although the FIFO is only 16 or 32 bytes deep, so if you expect a lot of data coming in, then it could overflow before you have a chance to read it out.

> Can you adjust mp_machine_i2c_target_write_bytes() to return 0 if there's no space in the FIFO? 

I tried this, but on rp2 and I assume also alif (they use the same Synopsis IP) the TX (outgoing) FIFO is flushed when there is an address match (also the RX FIFO).  So that means if you queue the data in the TX FIFO beforehand, it is all lost once the target is addressed by the controller.

To fix that either needs buffering outside the TX FIFO (ie in RAM), or use an interrupt based mechanism.

----

Maybe it's possible for you to use `I2CTarget` in memory device mode.  Just set an address size of 0 and a large memory buffer, eg:
```py
buf = bytearray(65536)
i2c_target = I2CTarget(0, addr, mem=buf, mem_addrsize=0)
```
Then for `get_bytes()` you just copy out of `buf`, and for `put_bytes()` you just copy into `buf`.  The actual transfer is then handled in the background.

Of course, that scheme would require having some way to know that new data has arrived in `buf`, and that data has been read out of `buf`.  Would having that capbility work for your case?

How much data do you have coming in and going out?  Do you need to be able to send large amounts of data from ROM (ie not copy to a temporary RAM buffer)?  What exactly is the protocol you use over I2C?

kwagyeman:
Using an address size of 0 (if that removes the address part) would work for me.

> Of course, that scheme would require having some way to know that new data has arrived in buf, and that data has been read out of buf. Would having that capbility work for your case?

Well, I need the method to transfer the requested data size or timeout. So, I'd need to know when the data has all been transferred within a certain time limit.

> How much data do you have coming in and going out? Do you need to be able to send large amounts of data from ROM (ie not copy to a temporary RAM buffer)? What exactly is the protocol you use over I2C?

For the OpenMV RPC library things are more complex at the upper layers of the protocol. For the lowest layer for I2C, the target device is just receiving an address match and then an expected number of bytes via I2C or timeout. It also will transmit a known number of bytes after an address match or timeout. The sizes of the buffers change all the time along with the timeouts.

dpgeorge:
> Well, I need the method to transfer the requested data size or timeout. So, I'd need to know when the data has all been transferred within a certain time limit.

With the memory device, you'd set up the data buffer and then just let it do its thing.  If you need a timeout I guess you could do something like:
```py
def transfer_bytes(buf, timeout):
    i2c_target = I2CTarget(0, addr, mem=buf, mem_addrsize=0)
    t0 = time.ticks_ms()
    while True:
        if i2c_target.transfer_done():
            # controller read our buffer or wrote into our buffer
            i2c_target.deinit()
            return True
        if time.ticks_diff(time.ticks_ms(), t0) >= timeout:
            # timeout
            i2c_target.deinit()
            return False
        time.sleep(0)
```
That would handle both reads and writes.  The I2C target would disappear between transfers (I think that's how you have it at the moment?).

> The sizes of the buffers change all the time along with the timeouts.

What's the approximate maximum buffer size?  Is it OK for all buffers to be in RAM?

kwagyeman:
The largest buffer size can be 2^32 bytes. All would be in RAM or whatever is addressable via byte arrays or memory views. I think I split things up by 65K chunks though in the code.

For:

```
def transfer_bytes(buf, timeout):
    i2c_target = I2CTarget(0, addr, mem=buf, mem_addrsize=0)
    t0 = time.ticks_ms()
    while True:
        if i2c_target.transfer_done():
            # controller read our buffer or wrote into our buffer
            i2c_target.deinit()
            return True
        if time.ticks_diff(time.ticks_ms(), t0) >= timeout:
            # timeout
            i2c_target.deinit()
            return False
        time.sleep(0)
```

Could a method be added to machine class to do all this for you in C? Maybe add a poll method which you could select and wait on. It would return or timeout then. Each return could mark the completion of a I2C read/write.

dpgeorge:
> Could a method be added to machine class to do all this for you in C?

What's wrong with having it in Python, eg in a helper module in `micropython-lib`?  That would be more flexible than in C, and not any slower.

robert-hh:
The new commits show some improvements. The double call of the ADDR_MATCH callback on data controller -> target is gone. The data corruption is still present, and then two subsequent readfrom_mem() calls return different data.

Log at the controller:
```
>>> i2c.readfrom_mem(67, 0, 8)
b'01234567'
>>> i2c.writeto_mem(67, 0, "abcdefgh")
>>> i2c.readfrom_mem(67, 0, 8)
b'bcdefg7\x00'
>>> i2c.readfrom_mem(67, 0, 8)
b'\x00bcdefg7'
>>> i2c.readfrom_mem(67, 0, 8)
b'\x00bcdefg7'
```
Set-up: Controller: MIMXRT1020-EVK, hard I2C, freq=400_000, Target RP2 Pico, mem-object preset with "01234567", callback for ADDR_MATCH, which just prints the irq.flags().
At freq=200_000 and of course below this test passes, even with longer text.

Screen shot of the first readfrom_mem():
grafik

Screen shot of writeto_mem():
grafik

Screen shot of the last readfrom_mem():
grafik

dpgeorge:
@robert-hh thanks for further testing.  I could reproduce your issue with a PYBv1.0 as the controller (at 330kHz), and a Pico W as the target (although I had to add a `sleep_us(10)` in the I2C IRQ handler to trigger the bug).

The issue was the stop condition being processed before the last incoming byte, which lead to that last byte being missed.  And that messed up subsequent transactions.  Should be fixed now.

kwagyeman:
@dpgeorge - A helper in python is fine. Just thinking that the cleanest way to make it to have something that's pollable. https://docs.micropython.org/en/latest/library/select.html

Like, the way the OpenMV's RPC library works now is okay with a blocking loop... but, it needs to be rewritten eventually to support asyncio.

robert-hh:
>  Should be fixed now.

Confirmed. The above test passes at freq=400_000.
At longer messages clock stretching kicks in after 16 bytes.

robert-hh:
Both the implementation for SAMD and MIMXRT are updated and rebased onto the actual state of this PR17365.

dpgeorge:
@robert-hh I've added your commits for samd and mimxrt (and the test change) to this PR.  Thanks for the work on that!

I see you noted that the mimxrt implementation moves a lot of things into IRAM.  In your opinion is the additional amount of RAM used an acceptable amount?

robert-hh:
I "just" replaced other code in IRAM with the I2C driver. That section of RAM is dedicated to code caching, so swapping does not affect the space used for MicroPython.

dpgeorge:
I've now added support for ESP32.  It was not easy, as @robert-hh discusses above in https://github.com/micropython/micropython/pull/17365#issuecomment-2979913776 :

> The ESP32 API has three versions of I2C support. 1. The legacy API, which is currently used for I2C controller support. 2. The ng API Version 1. 3, The ng API version 2. Option b) is declared as getting expired with idf 6.x. The ng API V2 is available starting version 5.4. So using is at the moment is not possible. The legacy API and ng API must not be used at the same time. So if that is going to be used, the I2C class has to be updated.

The esp32 port now uses IDF 5.4.2, so ng v2 I2C slave is available.  And there's an option `CONFIG_I2C_SKIP_LEGACY_CONFLICT_CHECK=y` which I used to allow it to co-exist with the existing I2C controller code.  That seems to work OK, at least as long as you only use I2C or I2CTarget (not both at the same time).

> Implementing the class I2CTargetMemory for trial it turned out that it cannot properly support the target memory mode. The API offers two callback functions or receive and transmit. Each of them is called AFTER the respective bus transactions is finished

Yes, that's pretty strange, and it confused me to start with.  I tried to get it working (by prefilling the TX FIFO before the transaction) but gave up when I realised that behaviour seems to only occur on ESP32 (original), not the other SoCs.  So I switched to development on ESP32S2 and it works much better, the transmit callback comes in before the transaction.  (It's still not ideal, but at least it can be made to work.)

> One has to declare a suitable large size for the temporary buffer

Yes, and for that I used the size of the user-supplied memory buffer.  This gets the tests working (most of them).

----

So ESP32 (ESP32S2 at least) now supports the memory buffer mode, and also the `IRQ_END_READ` and `IRQ_END_WRITE` events, as soft callbacks.  That means the code can be notified after the memory has been read/written.  And I added a `I2CTarget.memaddr()` function so the code can see which memory address was read/written.

I updated the documentation and other ports to work with the tweaked API / bindings.

The tests are also updated.

robert-hh:
Thanks for picking up the work on ESP32. I did not know how to continue. About I2C V2. There was a PR #17696 this morning to update I2C support to V2.

dpgeorge:
> About I2C V2. There was a PR #17696 this morning to update I2C support to V2.

That can be independent to the work here.  My testing on ESP32S2 shows that the existing I2C code works alongside this new I2C target code.

dpgeorge:
@projectgus I've now updated and tested this and it's ready to review.  Note that the commits still need to be squashed, but otherwise the code is in good shape.

@robert-hh there's a problem with mimxrt running the tests: it can do a read and a write transaction separately, but a read-then-write doesn't work, it doesn't respond to the write (ENODEV is raised).

robert-hh:
@dpgeorge I'll look into that. Which exact sequence caused the issue?

dpgeorge:
> Which exact sequence caused the issue?

When doing a `readfrom_mem(...)` on the controller side, when mimxrt is the target.  The controller will raise with ENODEV (works fine with `writeto_mem()` and with a simple `readfrom()`).

The included `tests/extmod_hardware/machine_i2c_target.py` test will show the issue.)

robert-hh:
> When doing a readfrom_mem(...) on the controller side

That is strange. I'm sure that `readfrom_mem(...)` and `writeto_mem` were my favorite test cases and that the test scripts `tests/extmod_hardware/machine_i2c_target.py` worked. I ran that test routinely.  I'll look into it.

ports/esp32/mpconfigport.h
```diff
@@ -138,6 +138,9 @@
 #define MICROPY_PY_MACHINE_PWM_INCLUDEFILE  "ports/esp32/machine_pwm.c"
 #define MICROPY_PY_MACHINE_I2C              (1)
 #define MICROPY_PY_MACHINE_I2C_TRANSFER_WRITE1 (1)
+#define MICROPY_PY_MACHINE_I2C_TARGET       (SOC_I2C_SUPPORT_SLAVE && !CONFIG_IDF_TARGET_ESP32)
```
projectgus, line 142: Suggest a comment here explaining why we don't support this feature on original ESP32

ports/samd/machine_i2c.c
```diff
@@ -72,16 +70,18 @@ typedef struct _machine_i2c_obj_t {
     uint8_t *buf;
 } machine_i2c_obj_t;
 
+uint8_t machine_i2c_pin_config(mp_obj_t pin_obj, uint8_t id, uint8_t pad_nr);
```
projectgus: Suggest adding this declaration to one of the port headers (maybe `ports/samd/modmachine.h` - that one has a similar one-off declaration already) and including that from here, `machine_i2c_target.c:53`, and `pin_af.c` so that the compiler can check the signatures don't deviate.

projectgus (approved):
I think this will be really useful!

I haven't looked closely at all the port implementations but I think this is worth including, the main API looks sound (and flexible) and other details can be cleaned up later.

I left a couple of small comments, and agree with some of @robert-hh's outstanding ones as well - but they're all pretty minor.

robert-hh:
@dpgeorge I cannot replicate the problem. I fetched the last state from the repository and made a fresh build for a Teensy 4.1.  `readfrom_mem(...) ` works, and `tests/extmod_hardware/machine_i2c_target.py` runs fine as well:

Edit: I should note that I have 5k6 pull-up resistors at SCL and SDA. Otherwise the internal pull-up would not be string enough when attaching the logic analyzer probes or using an external board as controller.

```
micropython/tests$ mpremote run extmod_hardware/machine_i2c_target.py 
test_scan (__main__.TestMemory) ... ok
test_write (__main__.TestMemory) ... ok
test_write_wrap (__main__.TestMemory) ... ok
test_write_wrap_large (__main__.TestMemory) ... ok
test_read (__main__.TestMemory) ... ok
test_read_wrap (__main__.TestMemory) ... ok
test_read_wrap_large (__main__.TestMemory) ... ok
test_write_read (__main__.TestMemory) ... ok
test_write_read_read (__main__.TestMemory) ... ok
test_scan (__main__.TestMemoryIRQ) ... ok
test_write (__main__.TestMemoryIRQ) ... ok
test_read (__main__.TestMemoryIRQ) ... ok
test_scan (__main__.TestIRQ) ... ok
test_write (__main__.TestIRQ) ... ok
test_read (__main__.TestIRQ) ... ok
test_write_read (__main__.TestIRQ) ... ok
----------------------------------------------------------------------
Ran 16 tests

OK
```

```
micropython/tests$ ./run-tests.py -t a0 extmod_hardware/machine_i2c_target.py 
platform=mimxrt arch=armv7emdp inlineasm=thumb
pass  extmod_hardware/machine_i2c_target.py (unittest: 16 ran)
1 tests performed (16 individual testcases)
1 tests passed
```

dpgeorge:
Thanks @robert-hh for testing.

> I should note that I have 5k6 pull-up resistors at SCL and SDA

This is what I was missing (rookie mistake!).  All other boards that I've tested with don't need additional pull-ups, it's enough to enable pull-up on both controller and target to get it working.  But with Teensy 4.0 (which I'm using) I need to add external resistors to get all tests passing.

Thanks for the hint.  Now mimxrt looks good.

ports/rp2/machine_i2c_target.c
dpgeorge: Now resolved with a configurable macro parameter.

ports/rp2/machine_i2c_target.c
```diff
@@ -0,0 +1,185 @@
...
+    mp_hal_pin_obj_t sda;
+} machine_i2c_target_memory_obj_t;
+
+static machine_i2c_target_memory_obj_t machine_i2c_target_memory_obj[] = {
+    {{&machine_i2c_target_memory_type}, i2c0, 0, MICROPY_HW_I2C0_SCL, MICROPY_HW_I2C0_SDA},
+    {{&machine_i2c_target_memory_type}, i2c1, 0, MICROPY_HW_I2C1_SCL, MICROPY_HW_I2C1_SDA},
+};
+
+MP_REGISTER_ROOT_POINTER(mp_obj_t pyb_i2cslave_mem[4]);
+
```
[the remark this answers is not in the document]
dpgeorge: Now uses a configurable macro parameter.

extmod/machine_i2c_target.c
```diff
@@ -0,0 +1,107 @@
...
+#include "py/runtime.h"
+
+#if MICROPY_PY_MACHINE_I2C_TARGET_MEMORY
+
+#include "extmod/modmachine.h"
+
+typedef struct _machine_i2c_target_data_t {
+    uint8_t first_rx;
+    uint8_t addr;
+    uint8_t len; // stored as len-1 so it can represent up to 256
```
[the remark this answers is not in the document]
dpgeorge: These are now 32-bit variables.

docs/library/machine.I2CTarget.rst
dpgeorge: Now fixed.

ports/esp32/mpconfigport.h
dpgeorge: OK, done.

ports/samd/machine_i2c.c
dpgeorge: I've renamed this to `pin_config_for_i2c()` and put it in `pin_af.h`.

ports/rp2/machine_i2c_target.c
```diff
@@ -0,0 +1,185 @@
...
+static void mp_machine_i2c_target_memory_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
+    machine_i2c_target_memory_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    mp_printf(print, "I2CTargetMemory(%u, addr=%u, scl=%u, sda=%u)",
+        self - &machine_i2c_target_memory_obj[0], self->addr, self->scl, self->sda);
+}
+
+static void mp_machine_i2c_target_memory_deinit(machine_i2c_target_memory_obj_t *self) {
+    i2c_slave_deinit(self->i2c_inst);
+}
```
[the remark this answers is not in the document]
dpgeorge: Yes, you are right.  New fixed: the common extmod code clears out everything upon deinit and soft reset.

extmod/machine_i2c_target.c
dpgeorge: Unnecessary events have now been removed.

extmod/machine_i2c_target.c
dpgeorge: This should now be fixed.

extmod/machine_i2c_target.c
dpgeorge: This has been reworked and should now be fixed.

dpgeorge:
Apart from squashing commits, this PR is now ready to merge.

(I retested on PYBV10, PYBD_SF2, RPI_PICO, TEENSY40, ALIF_ENSEMBLE and zephyr with RPI_PICO.  Everything looks OK, just need to remember to use pull-ups!)

iabdalkader:
I tested the Alif port with AE3 and a bus pirate. Does the read/write address wrap around or something like that? I'm can't get reading/writing to work in any order that makes sense.

1- Reading seems to return bytes from mem randomly every time:
```
>>> from machine import I2CTarget
>>> mem = bytearray([0, 1, 2, 3, 4, 5, 6, 7])
>>> i2c = I2CTarget(id=1, addr=0x60, mem=mem)
>>> mem
bytearray(b'\x00\x01\x02\x03\x04\x05\x06\x07')

I2C> [0xC1 r:8]
I2C START
TX: 0xC1 ACK 
RX: 0x07 ACK 0x01 ACK 0x02 ACK 0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 NACK     
I2C STOP

I2C> [0xC1 r:8]
I2C START
TX: 0xC1 ACK 
RX: 0x02 ACK 0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x01 ACK 0x02 NACK     
I2C STOP
```

2- Writes are partial, some data gets dropped and always seem to pick a random offset to write:

```
I2C> [0xC0 0xAA 0xBB 0xCC 0xDD 0xFF 0xFF 0xFF 0xFF]
I2C START
TX: 0xC0 ACK 0xAA ACK 0xBB ACK 0xCC ACK 0xDD ACK 0xFF ACK 0xFF ACK 0xFF ACK 0xFF ACK 
I2C STOP

>>> mem
bytearray(b'\xff\x01\xbb\xcc\xdd\xff\xff\xff')

I2C> [0xC0 0 1 2 3 4 5 6 7]                           
TX: 0xC0 ACK 
TX: 0 ACK 1 ACK 2 ACK 3 ACK 4 ACK 5 ACK 6 ACK 7 ACK 

>>> mem
bytearray(b'\x01\x02\x03\x04\x05\x06\x07\xff')
```

3- Scanning returns the correct address (0x60) but also some other addresses reply which is weird because nothing else is on the bus:
```
I2C> scan
I2C address search:
0x00 (0x00 W)   #? 
0x60 (0xC0 W) (0xC1 R)
0x7C (0xF8 W) (0xF9 R)    #?

Found 5 addresses, 2 W/R pairs.
```

dpgeorge:
@iabdalkader do you have pull-ups on SCL and SDA?

ALIF_ENSEMBLE works fine for me, passing the included tests.  I tested it connected to a PYBD_SF2, with the PYBD_SF2 a controller and the ALIF_ENSEMBLE the I2C target.  Also tested the ALIF_ENSEMBLE against itself, using `tests/extmod_hardware/machine_i2c_target.py`.  That passes all tests.

It should be fine anyway, because it's exactly the same as the rp2 port, the silicon IP is the same (Design Ware / Synopsis).  And rp2 also passes all tests.

iabdalkader:
> @iabdalkader do you have pull-ups on SCL and SDA?

Yes, the BP has onboard pullups connected to 3.3v. I enabled them for all those tests.

> I tested it connected to a PYBD_SF2, with the PYBD_SF2 a controller and the ALIF_ENSEMBLE the I2C target.

Does a scan from PYBD_SF2 also return more than one address? Assuming the Ensemble board has nothing else on the bus, you should see just one address.

> It should be fine anyway, because it's exactly the same as the rp2 port, the silicon IP is the same (Design Ware / Synopsis). And rp2 also passes all tests.

While it is the same IP, we know that it's configured differently. For example, the STOP control is configured differently for AE3. Could that be related or some other IP config option that's affecting this?

I also tested another bus (i2c2) to be able to connect the probes directly instead of via jumpers, thought it might have something to do with, but it's the same. I don't know what else to try. Any suggestions?

dpgeorge:
> Yes, the BP has onboard pullups connected to 3.3v.

The AE3 is 1.8v, so you must be going through a level convertor.  Maybe that's the problem?

> Does a scan from PYBD_SF2 also return more than one address?

No, it returns just the single address.  I was using pins P3_7 and P3_6 on the Ensemble board.

> I don't know what else to try. Any suggestions?

Do you have an Ensemble board?  If so I suggest trying that because it's known to work (and don't use level shifters).  Try running `tests/extmod_hardware/machine_i2c_target.py`.  Be sure to connect P1_1 to P0_3, and P1_0 to P0_2.

You can also try a RPI_PICO connected to the Ensemble to run the multitest, `tests/multi_extmod/*.py`.

robert-hh:
> I retested on PYBV10, PYBD_SF2, RPI_PICO, TEENSY40, ALIF_ENSEMBLE and zephyr with RPI_PICO

Tested with SAMD using using `tests/extmod_hardware/machine_i2c_target.py`. It locks up at TestPollling. Maybe the statement to skip it is missing in the test script, like it was added e.g. for mimxrt.

Edit: When it locks up, the SCL line is low and the last action  are two bytes on the bus, address_with_write_flag and a ASCII "0", not ACKed. The debugger tells that C the firmware is running and it looks like python code is executed. When stopping the code, the first entry on the call stack is `mp_execute_bytecode`, and further up is `mp_machine_i2c_target_event_callback`.
So that's the irq handler in the test where it tries to execute `i2c_target.write(b"0123")`. That must fail. The SAMD implementation cannot write to I2C in the I2C interrupt handler.

iabdalkader:
> The AE3 is 1.8v, so you must be going through a level convertor. Maybe that's the problem?

Yes, but I don't think it's causing any problems. Reading doesn't return wrong/shifted data. For example, if you set the memory array to 0xAB, 0xAB,... reading always returns 0xAB, 0xAB... not any other value. Second reading more than len(mem) seems to wrap around correctly, it just uses the last_mem_addr + 1 for following reads. Do you see the pattern? Is this expected?

```
TX: 0xC1 ACK 
RX: 0x01 ACK 0x02 ACK 0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 ACK 
    0x01 ACK 0x02 ACK 0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 NACK 

I2C> [0xC1 r:16]
TX: 0xC1 ACK 
RX: 0x02 ACK 0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 ACK 0x01 ACK 
    0x02 ACK 0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 ACK 0x01 NACK 

I2C> [0xC1 r:16]
TX: 0xC1 ACK 
RX: 0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 ACK 0x01 ACK 0x02 ACK 
    0x03 ACK 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 ACK 0x01 ACK 0x02 NACK 
    
I2C> [0xC1 r:16]
TX: 0xC1 ACK 
RX: 0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 ACK 0x01 ACK 0x02 ACK 0x03 ACK 
    0x04 ACK 0x05 ACK 0x06 ACK 0x07 ACK 0x00 ACK 0x01 ACK 0x02 ACK 0x03 NACK 
```

> No, it returns just the single address. I was using pins P3_7 and P3_6 on the Ensemble board.

That could indicate some problem.

robert-hh:
@iabdalkader Did you use i2c.readfrom() or i2c.readfrom_mem()? You have to use the latter.

iabdalkader:
> @iabdalkader Did you use i2c.readfrom() or i2c.readfrom_mem()? You have to use the latter.

I use a bus pirate to read/write. I'm just not sure what the expected behavior here is, should it wrap around like that or restart the offset/pointer if it sees a stop?

Ah I see, so I have to write the address first. That makes sense. Probably why write was broken too.

iabdalkader:
@robert-hh That was the problem indeed. Thank you very much!

@dpgeorge Read/write is working fine, I was just using it wrong not writing the memory address. Scan however, still produces many addresses for me, not sure why.

EDIT: Just noticed that those other addresses are general call and invalid (0x7C) they normally don't show in a scan with other boards.

robert-hh:
> Ah I see, so I have to write the address first. 

Yes. You have to write the start offset of the memory byte first and then read or write the data.

kwagyeman:
Can confirm that Alif responds to the invalid I2C address 0x7C, and general call 0x00.

For general call, that's enabled by default to be responded to:

image

For 0x7C, this seems to be a reserved address but may be related to 10-bit addressing or some other feature of the IP that's enabled by default. Probably not an issue that the IP responds to it.

image

dpgeorge:
> Tested with SAMD using using `tests/extmod_hardware/machine_i2c_target.py`. It locks up at TestPollling. Maybe the statement to skip it is missing in the test script, like it was added e.g. for mimxrt.

Thanks @robert-hh for testing on samd.  That TestPolling is intended to testing reading/writing the I2C FIFO outside the IRQ.  I'm happy for this test to fail, it's not a use case that we support yet, even if some ports can support it (eg because they have a relatively large I2C hardware FIFO, like rp2 and alif).

So I'll just skip this test on samd.  Can be fixed later if needed.

dpgeorge:
> Read/write is working fine, I was just using it wrong not writing the memory address

OK, thanks for confirming that.

> Can confirm that Alif responds to the invalid I2C address 0x7C, and general call 0x00.

I think that's acceptable.  These addresses are not usually used (in a 7-bit addressing system).  Responding to general call doesn't hurt.  If it turns out to be an issue, we can fix it then.

dpgeorge:
Commits have been squashed and docs improved.

iabdalkader:
@dpgeorge When using the memory backed mode/API, the memory buffer is read/written from IRQs, correct? Wouldn't this lead to race conditions? If so, how could we avoid this? Is it possible to schedule the reads/writes instead? Or if the API allows it somehow, we could maintain shadow regs and swap them atomically?

dpgeorge:
> When using the memory backed mode/API, the memory buffer is read/written from IRQs, correct?

Yes, correct.

> Wouldn't this lead to race conditions?

I guess that depends on how your Python code works, and what the I2C protocol is like.  Eg you could make it so that a `writeto_mem()` will set up an operation and then the target will perform that operation and eventually write out a result into the memory buffer.  Then it can set a special byte (say the first one) to indicate it's done.  The controller can poll the first byte to check, and when the operation is done it can read the rest of memory for the result.

> Is it possible to schedule the reads/writes instead?

It depends what MCU you are using.  Most of them support clock stretching (alif, rp2, stm32) so you can postpone the I2C transaction until the data is ready.  But to do that you probably need to implement a memory device by hand using the `I2CTarget.IRQ_xxx` callbacks.

If you have full control over the I2C controller you should be able to make it work using the memory-backed mode.

iabdalkader:
> Then it can set a special byte (say the first one) to indicate it's done.

Thanks. We'll try to implement something very similar using shadow regs and a special UPDATE register.  Each side will have a copy of the regs which will be synchronized atomically. I couldn't think of a way to avoid critical sections, as the target could get interrupted by another update from the controller while it's reading an update.

kwagyeman:
Hi Damien, I tried to get our RPC library working. However, the AE3 seems to always crash. Is this code problematic?

```
class rpc_i2c_slave(rpc_slave):
    def __init__(self, slave_addr=0x12, i2c_bus=2):  # private
        self.__addr = slave_addr
        self.__bus = i2c_bus
        rpc_slave.__init__(self)
        self._stream_writer_queue_depth_max = 1

    def transfer(self, data, timeout_ms):
        view = memoryview(data)
        for i in range(0, len(view), 65535):
            target = machine.I2CTarget(self.__bus, self.__addr, mem=view[i : i + 65535], mem_addrsize=0)
            done = False
            def done_callback(trigger):
                nonlocal done
                done = True
            target.irq(done_callback)
            start = time.ticks_ms()
            while not done:
                time.sleep_ms(0)  # Is this needed??? Calls mp_event_handle_nowait() internally.
                if time.ticks_diff(time.ticks_ms(), start) >= timeout_ms:
                    target.deinit()
                    return None
            target.deinit()
        return view

    def get_bytes(self, buff, timeout_ms):  # protected
        buff = self.transfer(buff, timeout_ms)
        return buff

    def put_bytes(self, data, timeout_ms):  # protected
        self.transfer(data, timeout_ms)
```

There's no way to tell if a transfer has finished without using the irq() stuff. Having a clear on read flag with the end_write/read status would be helpful.

Going to try what Ibrahim has in mind next (not part of our RPC library). But, the above should have worked for RPC. It seems to work about 6 times and then the system dies. Given it doesn't die on the first time it works... makes me think there's some memory corruption happening.

image

System seems to terminate with the clock being held low. Afterwards it falls off USB:

image

It appears in this last transfer the number of bytes that should have been sent is off, should be one more byte... there's no NACK either.

kwagyeman:
Found the issue, alif has infinite loops in their i2c_enable/disable:

image

Disable doesn't seem to go low in some cases after being told to.

dpgeorge:
> Disable doesn't seem to go low in some cases after being told to

Hmm. Maybe the I2C bus is locked up (held low) and so the hardware is waiting (forever) for it to be released?  Can you measure SCL/SDA when the lockup occurs?

kwagyeman:
They are both held low. I've verified alif is doing it via unplugging it from the bus and seeing the lines rise. SDA appears to be controlled by the master, though. Alif is just holding clock low.

image

And then if I zoom out:

image

The reference manual says to reset to the whole subsystem when this happens... 

image

The recovery bits mention they only work for master mode. When I looked into resetting I2C it appears it has a permanent clock and no reset lines... I guess I need to ask their app engineer for help.

kwagyeman:
Note that modifying/commenting out the disable while loop appears to break I2C across the whole system. Adding a timeout counter in the loop also appears to not work.

PR #17366, opened 2025-05-27
shared/timeutils: Standardize supported date range on all platforms.

yoctopuce:
This is pull request is to ensure that time functions (`mktime`, `localtime` and `gmtime`) work properly on a reasonable date range, on all platforms, regardless of the epoch. 

It does fix issue #17340

### Summary

The most important point is to have a test case that checks conversion from timestamp to tuple and back for every day in the _supported time range_. This was currently missing. The new test code starts from 2001, first iterates *backward* on tuple to find the earliest supported date and then goes *forward* by adding 86400 to the timestamp, to verify that the date computations are correct, including leap years, and do not raise an overflow.

The second part of the pull request is to ensure that this test does actually pass on all platforms. 

The most severely affected platforms were 32 bit platforms using a 1970 epoch, as

1. there was an assumption that timestamps can be stored in a `mp_int_t`
2. conversion from the 1970 epoch to the 2000 epoch used internally by timeutils breaks computations for dates before 2000

On some of these platforms, the resulting supported date range was only 2001 to 2037. As 2038 is now only 13 years away, this clearly has to be fixed quickly.

Small code footprint beeing a key value for MicroPython, the suggested minimal _supported time range_ for all platforms is 1970 to 2099. This range can be covered using 32 bit timestamps on ressource-limited platforms, fixes the 1970 epoch problems and requires less code for handling of leap years.

Support for years 2100 and beyond is only enabled for platforms using 64 bit pointers, which are typically less ressource-constrained.

### Testing

The PR includes new test cases that verify the supported time range using architecture-independent Python code.

The new `timeutils` code has a smaller footprint than the original for 32 bit machines, as it does not have to handle signed inputs and does not have to handle century-based leap years. 

We could have opted to use signed timestamps centered on 2000, as was probably the original intent of the timeutils code. We felt however that it was more useful to fully support the 21st century rather going supporting dates back to 1940's but breaking in 2068. Handling timestamps as unsigned also simplifies the code by removing some checks for negative values.

We could have opted to upgrade all platforms to use 64-bit signed integers for time functions, which would have offered the broadest possible range for abstract date operations, spanning from the year 1600 to the year 3000 and beyond. However, this would have had a code size impact on 32-bit platforms, with little added value. We guessed that, if a specific application on a 32-bit platform required handling _exotic_ dates, a Python-based implementation would remain a good option.

### DIFF
--- a/extmod/modtime.c
+++ b/extmod/modtime.c
@@ -58,7 +58,7 @@ static mp_obj_t time_localtime(size_t n_args, const mp_obj_t *args) {
         return mp_time_localtime_get();
     } else {
         // Convert given seconds to tuple.
-        mp_int_t seconds = mp_obj_get_int(args[0]);
+        mp_timestamp_t seconds = timeutils_obj_get_timestamp(args[0]);
         timeutils_struct_time_t tm;
         timeutils_seconds_since_epoch_to_struct_time(seconds, &tm);
         mp_obj_t tuple[8] = {
@@ -90,7 +90,7 @@ static mp_obj_t time_mktime(mp_obj_t tuple) {
         mp_raise_TypeError(MP_ERROR_TEXT("mktime needs a tuple of length 8 or 9"));
     }
 
-    return mp_obj_new_int_from_uint(timeutils_mktime(mp_obj_get_int(elem[0]),
+    return timeutils_obj_from_timestamp(timeutils_mktime(mp_obj_get_int(elem[0]),
         mp_obj_get_int(elem[1]), mp_obj_get_int(elem[2]), mp_obj_get_int(elem[3]),
         mp_obj_get_int(elem[4]), mp_obj_get_int(elem[5])));
 }
--- a/ports/esp8266/modtime.c
+++ b/ports/esp8266/modtime.c
@@ -31,7 +31,7 @@
 
 // Return the localtime as an 8-tuple.
 static mp_obj_t mp_time_localtime_get(void) {
-    mp_int_t seconds = pyb_rtc_get_us_since_epoch() / 1000 / 1000;
+    mp_uint_t seconds = pyb_rtc_get_us_since_epoch() / 1000u / 1000u;
     timeutils_struct_time_t tm;
     timeutils_seconds_since_epoch_to_struct_time(seconds, &tm);
     mp_obj_t tuple[8] = {
@@ -50,5 +50,5 @@ static mp_obj_t mp_time_localtime_get(void) {
 // Returns the number of seconds, as an integer, since the Epoch.
 static mp_obj_t mp_time_time_get(void) {
     // get date and time
-    return mp_obj_new_int(pyb_rtc_get_us_since_epoch() / 1000 / 1000);
+    return timeutils_obj_from_timestamp(pyb_rtc_get_us_since_epoch() / 1000 / 1000);
 }
--- a/ports/unix/modtime.c
+++ b/ports/unix/modtime.c
@@ -34,6 +34,7 @@
 
 #include "py/mphal.h"
 #include "py/runtime.h"
+#include "shared/timeutils/timeutils.h"
 
 #ifdef _WIN32
 static inline int msec_sleep_tv(struct timeval *tv) {
@@ -130,12 +131,7 @@ static mp_obj_t mod_time_gm_local_time(size_t n_args, const mp_obj_t *args, stru
     if (n_args == 0) {
         t = time(NULL);
     } else {
-        #if MICROPY_PY_BUILTINS_FLOAT && MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_DOUBLE
-        mp_float_t val = mp_obj_get_float(args[0]);
-        t = (time_t)MICROPY_FLOAT_C_FUN(trunc)(val);
-        #else
-        t = mp_obj_get_int(args[0]);
-        #endif
+        t = (time_t)timeutils_obj_get_timestamp(args[0]);
     }
     struct tm *tm = time_func(&t);
 
@@ -196,7 +192,7 @@ static mp_obj_t mod_time_mktime(mp_obj_t tuple) {
     if (ret == -1) {
         mp_raise_msg(&mp_type_OverflowError, MP_ERROR_TEXT("invalid mktime usage"));
     }
-    return mp_obj_new_int(ret);
+    return timeutils_obj_from_timestamp(ret);
 }
 MP_DEFINE_CONST_FUN_OBJ_1(mod_time_mktime_obj, mod_time_mktime);
 
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -893,6 +893,64 @@ typedef double mp_float_t;
 #define MICROPY_FULL_CHECKS (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_CORE_FEATURES)
 #endif
 
+// Ports can choose to use timestamps based on 2000-01-01 or 1970-01-01
+// Default is timestamps based on 2000-01-01
+#if !defined(MICROPY_EPOCH_IS_2000) && !defined(MICROPY_EPOCH_IS_1970)
+#define MICROPY_EPOCH_IS_2000 (1)
+#define MICROPY_EPOCH_IS_1970 (0)
+#elif !defined(MICROPY_EPOCH_IS_1970)
+#define MICROPY_EPOCH_IS_1970 (1 - (MICROPY_EPOCH_IS_2000))
+#elif !defined(MICROPY_EPOCH_IS_2000)
+#define MICROPY_EPOCH_IS_2000 (1 - (MICROPY_EPOCH_IS_1970))
+#endif
+
+// To maintain reasonable compatibility with CPython on embedded systems,
+// and avoid breaking anytime soon, time functions are defined to work
+// at least between 1970 and 2099 (included) on any machine.
+//
+// Specific ports can enable extended date support
+// - after 2099 using MICROPY_TIME_SUPPORT_Y2100_AND_BEYOND
+// - before 1970 using MICROPY_TIME_SUPPORT_Y1969_AND_BEFORE
+// The largest possible range is year 1600 to year 3000
+//
+// By default, extended date support is only enabled for machines using 64 bit pointers,
+// but it can be enabled by specific ports
+#ifndef MICROPY_TIME_SUPPORT_Y1969_AND_BEFORE
+#if MP_SSIZE_MAX > 2147483647
+#define MICROPY_TIME_SUPPORT_Y1969_AND_BEFORE (1)
+#else
+#define MICROPY_TIME_SUPPORT_Y1969_AND_BEFORE (0)
+#endif
+#endif
+
+// When support for dates <1970 is enabled, supporting >=2100 does not cost anything
+#ifndef MICROPY_TIME_SUPPORT_Y2100_AND_BEYOND
+#define MICROPY_TIME_SUPPORT_Y2100_AND_BEYOND (MICROPY_TIME_SUPPORT_Y1969_AND_BEFORE)
+#endif
+
+// The type to be used to represent platform-specific timestamps depends on the choices above
+#define MICROPY_TIMESTAMP_IMPL_LONG_LONG (0)
+#define MICROPY_TIMESTAMP_IMPL_UINT (1)
+#define MICROPY_TIMESTAMP_IMPL_TIME_T (2)
+
+#ifndef MICROPY_TIMESTAMP_IMPL
+#if MICROPY_TIME_SUPPORT_Y2100_AND_BEYOND || MICROPY_TIME_SUPPORT_Y1969_AND_BEFORE || MICROPY_EPOCH_IS_2000
+#define MICROPY_TIMESTAMP_IMPL (MICROPY_TIMESTAMP_IMPL_LONG_LONG)
+#else
+#define MICROPY_TIMESTAMP_IMPL (MICROPY_TIMESTAMP_IMPL_UINT)
+#endif
+#endif
+
+// `mp_timestamp_t` is the type that should be used by the port
+// to represent timestamps, and is referenced to the platform epoch
+#if MICROPY_TIMESTAMP_IMPL == MICROPY_TIMESTAMP_IMPL_LONG_LONG
+typedef long long mp_timestamp_t;
+#elif MICROPY_TIMESTAMP_IMPL == MICROPY_TIMESTAMP_IMPL_UINT
+typedef mp_uint_t mp_timestamp_t;
+#elif MICROPY_TIMESTAMP_IMPL == MICROPY_TIMESTAMP_IMPL_TIME_T
+typedef time_t mp_timestamp_t;
+#endif
+
 // Whether POSIX-semantics non-blocking streams are supported
 #ifndef MICROPY_STREAMS_NON_BLOCK
 #define MICROPY_STREAMS_NON_BLOCK (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_EXTRA_FEATURES)
--- a/py/obj.c
+++ b/py/obj.c
@@ -314,6 +314,36 @@ mp_int_t mp_obj_get_int(mp_const_obj_t arg) {
     return val;
 }
 
+#if MICROPY_LONGINT_IMPL != MICROPY_LONGINT_IMPL_NONE
+mp_uint_t mp_obj_get_uint(mp_const_obj_t arg) {
+    if (!mp_obj_is_exact_type(arg, &mp_type_int)) {
+        mp_obj_t as_int = mp_unary_op(MP_UNARY_OP_INT_MAYBE, (mp_obj_t)arg);
+        if (as_int == MP_OBJ_NULL) {
+            mp_raise_TypeError_int_conversion(arg);
+        }
+        arg = as_int;
+    }
+    return mp_obj_int_get_uint_checked(arg);
+}
+
+long long mp_obj_get_ll(mp_const_obj_t arg) {
+    if (!mp_obj_is_exact_type(arg, &mp_type_int)) {
+        mp_obj_t as_int = mp_unary_op(MP_UNARY_OP_INT_MAYBE, (mp_obj_t)arg);
+        if (as_int == MP_OBJ_NULL) {
+            mp_raise_TypeError_int_conversion(arg);
+        }
+        arg = as_int;
+    }
+    if (mp_obj_is_small_int(arg)) {
+        return MP_OBJ_SMALL_INT_VALUE(arg);
+    } else {
+        long long res;
+        mp_obj_int_to_bytes_impl((mp_obj_t)arg, MP_ENDIANNESS_BIG, sizeof(res), (byte *)&res);
+        return res;
+    }
+}
+#endif
+
 mp_int_t mp_obj_get_int_truncated(mp_const_obj_t arg) {
     if (mp_obj_is_int(arg)) {
         return mp_obj_int_get_truncated(arg);
--- a/py/objint_longlong.c
+++ b/py/objint_longlong.c
@@ -295,6 +295,22 @@ mp_int_t mp_obj_int_get_checked(mp_const_obj_t self_in) {
     return mp_obj_int_get_truncated(self_in);
 }
 
+mp_uint_t mp_obj_int_get_uint_checked(mp_const_obj_t self_in) {
+    if (mp_obj_is_small_int(self_in)) {
+        if (MP_OBJ_SMALL_INT_VALUE(self_in) >= 0) {
+            return MP_OBJ_SMALL_INT_VALUE(self_in);
+        }
+    } else {
+        const mp_obj_int_t *self = self_in;
+        long long value = self->val;
+        mp_uint_t truncated = (mp_uint_t)value;
+        if (value >= 0 && (long long)truncated == value) {
+            return truncated;
+        }
+    }
+    mp_raise_msg(&mp_type_OverflowError, MP_ERROR_TEXT("overflow converting long int to machine word"));
+}
+
 #if MICROPY_PY_BUILTINS_FLOAT
 mp_float_t mp_obj_int_as_float_impl(mp_obj_t self_in) {
     assert(mp_obj_is_exact_type(self_in, &mp_type_int));
--- a/shared/timeutils/timeutils.h
+++ b/shared/timeutils/timeutils.h
@@ -27,9 +27,23 @@
 #ifndef MICROPY_INCLUDED_LIB_TIMEUTILS_TIMEUTILS_H
 #define MICROPY_INCLUDED_LIB_TIMEUTILS_TIMEUTILS_H
 
+#include "py/obj.h"
+#if MICROPY_PY_BUILTINS_FLOAT && MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_DOUBLE
+#include <math.h> // required for trunc()
+#endif
+
+// `timeutils_timestamp_t` is the type used internally by timeutils to
+// represent timestamps, and is always referenced to 1970.
+// It may not match the platform-specific `mp_timestamp_t`.
+#if MICROPY_TIME_SUPPORT_Y2100_AND_BEYOND || MICROPY_TIME_SUPPORT_Y1969_AND_BEFORE
+typedef long long timeutils_timestamp_t;
+#else
+typedef mp_uint_t timeutils_timestamp_t;
+#endif
+
 // The number of seconds between 1970/1/1 and 2000/1/1 is calculated using:
 // time.mktime((2000,1,1,0,0,0,0,0,0)) - time.mktime((1970,1,1,0,0,0,0,0,0))
-#define TIMEUTILS_SECONDS_1970_TO_2000 (946684800ULL)
+#define TIMEUTILS_SECONDS_1970_TO_2000 (946684800LL)
 
 typedef struct _timeutils_struct_time_t {
     uint16_t tm_year;       // i.e. 2014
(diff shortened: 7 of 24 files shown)


Josverl:
Good to have the tests, and the improvements and consistency across ports.

Does that now also make the tuple order and sizes across all ports the same?

yoctopuce:
I wasn't even aware that some ports did not follow the documentation for the tuple content. Given how the test code is written, my guess is that if the order is not the expected one for year, month and day the test will certainly fail. The code does not check the content of time, this could be added, at least to ensure some level of consistency (tuple order).

As for the extra numbers and tuple size 8/9, this was not in the scope of this PR. I was considering submitting a separate PR to provide an optional support for a simple TZ strings with DST support, so that devices can provide proper localtime support for logs for instance. But I first have to implement it :-)

extmod/modtime.c
```diff
@@ -58,7 +58,15 @@ static mp_obj_t time_localtime(size_t n_args, const mp_obj_t *args) {
...
-        mp_int_t seconds = mp_obj_get_int(args[0]);
+        mp_obj_t timestamp = args[0];
+        if (!mp_obj_is_exact_type(timestamp, &mp_type_int)) {
+            mp_obj_t as_int = mp_unary_op(MP_UNARY_OP_INT_MAYBE, (mp_obj_t)timestamp);
+            if (as_int == MP_OBJ_NULL) {
+                mp_raise_TypeError_int_conversion(timestamp);
+            }
+            timestamp = as_int;
+        }
+        mp_uint_t seconds = mp_obj_int_get_uint_checked(timestamp);
```
dpgeorge: I feel like this should be factored out into a function `mp_obj_get_uint(arg)`, and then reused below in the unix port's changes.

(Or call it `mp_obj_get_int_as_uint(arg)` or similar, but probably `mp_obj_get_uint(arg)` describes it best.)

ports/esp8266/modtime.c
```diff
@@ -50,5 +50,5 @@ static mp_obj_t mp_time_localtime_get(void) {
 // Returns the number of seconds, as an integer, since the Epoch.
 static mp_obj_t mp_time_time_get(void) {
     // get date and time
-    return mp_obj_new_int(pyb_rtc_get_us_since_epoch() / 1000 / 1000);
+    return mp_obj_new_int_from_uint((mp_uint_t)(pyb_rtc_get_us_since_epoch() / 1000u / 1000u));
```
dpgeorge: Does this need a cast, since it's already unsigned (`uint64_t`)?

ports/unix/modtime.c
```diff
@@ -132,9 +132,17 @@ static mp_obj_t mod_time_gm_local_time(size_t n_args, const mp_obj_t *args, stru
     } else {
         #if MICROPY_PY_BUILTINS_FLOAT && MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_DOUBLE
         mp_float_t val = mp_obj_get_float(args[0]);
-        t = (time_t)MICROPY_FLOAT_C_FUN(trunc)(val);
+        t = (time_t)(mp_uint_t)MICROPY_FLOAT_C_FUN(trunc)(val);
```
dpgeorge: Is the cast to `mp_uint_t` needed?  (Maybe it is, just trying to understand.)

shared/timeutils/timeutils.h
```diff
@@ -46,42 +46,47 @@ bool timeutils_is_leap_year(mp_uint_t year);
...
-
-// Year is absolute, month/mday are 1-based, hours/minutes/seconds are 0-based.
-static inline uint64_t timeutils_mktime(mp_uint_t year, mp_int_t month, mp_int_t mday, mp_int_t hours, mp_int_t minutes, mp_int_t seconds) {
-    return timeutils_mktime_2000(year, month, mday, hours, minutes, seconds) + TIMEUTILS_SECONDS_1970_TO_2000;
+    timeutils_seconds_since_1970_to_struct_time((mp_uint_t)(t + TIMEUTILS_SECONDS_1970_TO_2000), tm);
 }
 
 // Year is absolute, month/date are 1-based, hour/minute/second are 0-based.
-static inline uint64_t timeutils_seconds_since_epoch(mp_uint_t year, mp_uint_t month,
+static inline uint64_t timeutils_seconds_since_2000(mp_uint_t year, mp_uint_t month,
```
dpgeorge: It looks like this `uint64_t` is now used by ports that have 2000 as the epoch.  So that means `timeutils_seconds_since_epoch` will return a `uint64_t` type (see eg stm32 which uses this function).  I guess that's OK, but I thought the intention was to use only 32-bit range for bare-metal ports?

shared/timeutils/timeutils.h
```diff
@@ -46,42 +46,47 @@ bool timeutils_is_leap_year(mp_uint_t year);
...
     mp_uint_t date, mp_uint_t hour, mp_uint_t minute, mp_uint_t second) {
     // TODO this will give incorrect results for dates before 2000/1/1
```
dpgeorge: Is this TODO resolved now?

shared/timeutils/timeutils.h
```diff
@@ -46,42 +46,47 @@ bool timeutils_is_leap_year(mp_uint_t year);
...
+#define timeutils_seconds_since_epoch_to_struct_time timeutils_seconds_since_1970_to_struct_time
+#define timeutils_seconds_since_epoch timeutils_seconds_since_1970
+#define timeutils_mktime timeutils_mktime_1970
+
+static inline mp_uint_t timeutils_seconds_since_epoch_from_nanoseconds_since_1970(int64_t ns) {
     return (mp_uint_t)(ns / 1000000000ULL);
 }
 
-static inline uint64_t timeutils_nanoseconds_since_epoch_to_nanoseconds_since_1970(uint64_t ns) {
+static inline int64_t timeutils_nanoseconds_since_epoch_to_nanoseconds_since_1970(int64_t ns) {
```
dpgeorge, line 138: Should the arg and return type stay as unsigned here?  (I guess there's a good reason you changed it to signed.)

tests/extmod/time_mktime.py
```diff
@@ -0,0 +1,112 @@
+# test conversion from date tuple to timestamp and back
+
+try:
+    import time
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+# Range of date expected to work on all Micropython platforms
```
dpgeorge: Please use uppercase "P" in "MicroPython" :smile:

dpgeorge (commented):
Thanks for this PR.  This is something that's been on my to-do list for a while, so it's great to see it finally get fixed properly.

What's required to switch bare-metal ports to use 64-bit range for timestamps?

extmod/modtime.c
yoctopuce: That makes perfect sense.

ports/esp8266/modtime.c
yoctopuce: Some C compilers issue a warning when assigning 64bit int into 32bit without explicit cast, so I tend to always add this cast. But obviously this is not the case for this port, as the same expression a few lines above does not cause a warning, so I will revert this hunk.

ports/unix/modtime.c
yoctopuce: You are correct, it does not bring anything in this case.

tests/extmod/time_mktime.py
yoctopuce: Sure :-)

shared/timeutils/timeutils.h
yoctopuce: This commit is clearly lacking clarity with regard to what format should be used by which port.
I will push a new version which clarifies
- the type of timestamps used internally by timeutils
- the type of timestamps used by the port

shared/timeutils/timeutils.h
yoctopuce: Yes, I will remove the comments

yoctopuce:
> Thanks for this PR. This is something that's been on my to-do list for a while, so it's great to see it finally get fixed properly.
> 
> What's required to switch bare-metal ports to use 64-bit range for timestamps?

Any bare-metal port with an epoch at 2000 would actually need a 64 signed timestamp to properly encode time on the whole range.

Internally, timeutils can still use the unsigned 32 bit representation to reduce code footprint, but the port should be using 64 bit signed timestamps when interfacing it. I will work on checking this on all platforms using `timeutils`. However I don't have the machines available at hand to test my changes... What kind of integration test is run on baremetal platforms during CI tests ? I am looking at tests for `vfs_lfs`, `fatfs`, `modtime` and `machine_rtc`

yoctopuce:
I have pushed the new version, which should bring proper datetime support for most platforms.
There are a several platforms that still use an `uint64_t` when reading their RTC clock, but as the code setting the clock cannot use dates before 2000, this should not be an issue. When creating an mp_obj_t to hold a timestamp , they now use a proper function that will not fail when the timestamp is past a signed int.

robert-hh:
With which port did you test the PR?

yoctopuce:
> With which port did you test the PR?

I tested on ports/unix (using int64_t timestamps) and on our bare-metal port (an ARM platform using 1970 epoch, using 32 bit timestamps), but I don't have MicroPython classic bare metals available here. This is the reason I was asking if there was any possibility of running CI test remotely on bare metal hardware.

yoctopuce:
Anyone could suggest an idea to explein why the qemu_mips would be failing ?
The failing code is a new bit of the extra coverage written in C, which works properly in the standard `ports/unix`:
```c
        // convert a large timestamp value (stored in a mpz) to ll
        mp_printf(&mp_plat_print, "%lx\n", mp_obj_get_ll(obj_timestamp));

        // convert a small timestamp value (as a smallint) to a ll;
        obj_timestamp = timeutils_obj_from_timestamp(0xc0ffee);
        mp_printf(&mp_plat_print, "%lx\n", mp_obj_get_ll(obj_timestamp));
```
On qemu_mips, this is returning 0 instead of the proper values. The problem is likely the implementation of `mp_obj_get_ll()`, which uses `mp_obj_int_to_bytes_impl()` to retrieve the 64 bit value from the mpz number:
```c
int64_t res;
mp_obj_int_to_bytes_impl((mp_obj_t)arg, MP_ENDIANNESS_BIG, sizeof(res), (byte *)&res);
return res;
```
It is very similar to `ffifunc_print()` implementation, so I don't see why it would not work...

edit 1 : ffi is apparently skipped on mips. But I believe I found the problem, pushing a new version...
edit 2 : that was it. My mistake was using mp_printf "%lx" with an int64_t, which does not seems to work on qemu_mips build.

robert-hh:
I compiled this PR for a few boards, installed them on the board and ran the test script `time_mktime.py` with these boards. Result:

The test passes for STM32, ESP32 and MIMXRT
The test fails for CC3200, ESP8266, SAMD, Renesas-Ra, RP2 and NRF

The fail does not necessary tell that the change is wrong. Chances are better that it is a problem of the test script. Can you provide a few test cases and the expected result, which I can feed manually to the boards at REPL?

yoctopuce:
> I compiled this PR for a few boards, installed them on the board and ran the test script `time_mktime.py` with these boards. Result:
> 
> The test passes for STM32, ESP32 and MIMXRT The test fails for CC3200, ESP8266, SAMD, Renesas-Ra, RP2 and NRF
> 
> The fail does not necessary tell that the change is wrong. Chances are better that it is a problem of the test script. Can you provide a few test cases and the expected result, which I can feed manually to the boards at REPL?

Thank you very much, that helps a lot.

If you could try to run the script in `tests/extmod/time_mktime.py`, it should either report an error if I did make something really wrong, or display the properly supported date range for the platform. Knowing where it fails would problably give a fair amount of information on the issue.

If this test pass completely, maybe the problem is with another function.
You could for instance try
```py
import time
print(time.time())
```
to check if I broke something there.

robert-hh:
As suspected the problem is with running the test e.g. on RP2 with:
`./run-tests.py -t a0 extmod/time_mktime.py`

When I run the test with `mprmote run extmod/time_mktime.py` almost all ports return the line:
`tested from (1970, 1, 1) to (2100, 1, 1)`.

Exceptions:
NRF: `AttributeError: 'module' object has no attribute 'mktime'`
CC3200: is a little bit unreliable when executing code with `mpremote run`. Sometimes it fails.

Runtime of the test between 1.7s (Teensy 4.1) and 26.9s (Renesas-RA EK6M2).

yoctopuce:
> As suspected the problem is with running the test e.g. on RP2 with: `./run-tests.py -t a0 extmod/time_mktime.py`
> 
> When I run the test with `mprmote run extmod/time_mktime.py` almost all ports return the line: `tested from (1970, 1, 1) to (2100, 1, 1)`.
> 
> Exceptions: NRF: `AttributeError: 'module' object has no attribute 'mktime'` CC3200: is a little bit unreliable when executing code with `mpremote run`. Sometimes it fails.
> 
> Runtime of the test between 1.7s (Teensy 4.1) and 26.9s (Renesas-RA EK6M2).

Excellent, thank you very much for running these all these tests. So it looks like the objective of having an uniform supported date range is met.

The NRF error makes perfect sense: NRF port has very limited support for module `time`, it does not include `mktime`. I have just added a check at the beginning of my test to make sure to properly reports a `SKIP` rather than a failure when `time.mktime()` is unavailable.

Thank you again for you help testing this PR.

robert-hh:
It is still strange that the test looks fine when executed directly via `mpremote run`, and fails for some boards when executed by `run-test`. I reran some tests. It seems to pass for boards where the runtime of the test is below 10s, and fail for others. But the default timeout in run-tests.py is 30 seconds.
 Other observations:
- this morning the RP2 passed the test through run-tests.py. It's net execution time is <10s.
- looking at the output of the ESP8266 and Renesas RA when running the test with run-tests.py the behavior is strange. First I see a long series of reboots, at the end of that series a "START TEST" message and then the proper test result. Before that that run-tests-py had already reported after ~11 seconds that the test failed. So it seems that run-tests.py has somewhere a ~10 second timeout.

Edit: If I modify `time_mktime.py` such that it's runtime on SAMD is below 10s, then it passes with `run-tests.py`.
Edit 2: The test is properly skipped for the NRF port.

yoctopuce:
I did not anticipate that the test would take that long to run on bare metals. It might be a problem if it slows down the whole testing process.

I have just pushed a new version that skips faster over first days of the month until the 27th. This should run significantly faster. Can you give it a try ?

robert-hh:
All boards except cc3200 pass the test now when performed using run-tests.py. The script time_mktime.py itself runs fine at the CC3200 board. So it's a test run issue.  The problem seems to be in the code transport mechanism with pyboard.py. Even running the code with `mpremote run` is not reliable.
The CC3200 is already a little bit outdated. A fix of the test runner may not be needed.

yoctopuce:
> All boards except cc3200 pass the test now when performed using run-tests.py. The script time_mktime.py itself runs fine at the CC3200 board. So it's a test run issue. The problem seems to be in the code transport mechanism with pyboard.py. Even running the code with `mpremote run` is not reliable. The CC3200 is already a little bit outdated. A fix of the test runner may not be needed.

Excellent, thank you very much for your assistance !

While you have all the platforms available at hand, would you ave a chance to try to the other PR that I pushed yesterday, https://github.com/micropython/micropython/pull/17444 ? Same scenario as here: there is a new test file `tests/float/float_format_accuracy.py` that performs a 10'000 float conversions and tests the accuracy, but I now suspect my test code might be too slow to run on some platforms, as it was the case here.

robert-hh:
The slowest board in the previous test was the ESP8266. So I ran `run-tests.py -t u0 float/*.py`. It ran to an end, with the test summary below:

```
42 tests performed (2417 individual testcases)
34 tests passed
24 tests skipped: float/cmath_dunder.py float/cmath_fun.py float/cmath_fun_special.py float/complex1_intbig.py float/complex1.py float/complex_dunder.py float/complex_reverse_op.py float/complex_special_methods.py float/float2int_doubleprec_intbig.py float/float2int_intbig.py float/float_divmod.py float/float_format_ints_doubleprec.py float/float_format_ints_power10.py float/float_parse_doubleprec.py float/float_struct_e_doubleprec.py float/int_big_float.py float/math_constants_extra.py float/math_domain_special.py float/math_factorial_intbig.py float/math_fun_special.py float/math_isclose.py float/string_format.py float/true_value.py float/types.py
8 tests failed: float/builtin_float_round.py float/float1.py float/float_format_accuracy.py float/float_format_ints.py float/float_format.py float/float_parse.py float/lexer.py float/python36.py
```

Running the test on a SAMD21, only `float/float_format_accuracy.py` failed.
Renesas-RA EK-RA6M2 is slow as well, but all tests eithe pass or are skipped.

Besides that, the runtime should be fine.

yoctopuce:
> The tested ESP32, RP2 and NRF boards fail the `float_format_accuracy.py` test, because it takes too long. ESP8266 and SAM21 fail this test with:
> 
> ```
>   File "<stdin>", line 19, in pseudo_randouble
> AttributeError: 'module' object has no attribute 'crc32'
> ```

Thank you very much for the feedback, it looks like these platforms don't have built-in CRC32, I will put a Python implementation in my code. Would you mind copy-pasting your comments into the other PR so that things don't get mixes ?

Thanks again.

robert-hh:
I deleted my comments. You have to delete yours.

yoctopuce:
Excellent, thank you [robert-hh](https://github.com/robert-hh) for your help in testing this PR.
- We now have all automated CI checks passed
- Running `run-test` on STM32, ESP32, MIMXRT, ESP8266, SAMD, Renesas-Ra, RP2 and NRF is successful
- Running `time_mktime.py` manually on CC3200 is also successful
I guess we are ready for the next code review

py/obj.c
```diff
@@ -314,6 +314,36 @@ mp_int_t mp_obj_get_int(mp_const_obj_t arg) {
     return val;
 }
 
+#if MICROPY_LONGINT_IMPL != MICROPY_LONGINT_IMPL_NONE
+mp_uint_t mp_obj_get_uint(mp_const_obj_t arg) {
```
dpgeorge, line 318: Please can you put this change (adding these two methods) into a separate commit?  It's really quite independent to the change to timeutils.

Please also put the new coverage test for these functions into the same commit.

Also, I don't think this will build correctly when longlong is enabled, because `mp_obj_int_get_uint_checked()` is not implemented for longlong...

shared/timeutils/timeutils.h
```diff
@@ -27,6 +27,46 @@
 #ifndef MICROPY_INCLUDED_LIB_TIMEUTILS_TIMEUTILS_H
 #define MICROPY_INCLUDED_LIB_TIMEUTILS_TIMEUTILS_H
 
+#if MICROPY_PY_BUILTINS_FLOAT && MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_DOUBLE
+#include <math.h> // required for trunc()
```
dpgeorge, line 32: I think this if-guard and include should come after the include of `py/obj.h` below, so the MICROPY_xxx configurations are available.

shared/timeutils/timeutils.h
```diff
@@ -27,6 +27,46 @@
...
+// To maintain reasonable compatibility with CPython on embedded systems,
+// and avoid breaking anytime soon, time functions are defined to work
+// at least between 1970 and 2099 on any machine.
+//
+// Specific ports can set MP_SUPPORT_Y2100_AND_BEYOND to 1 in order
+// to get extended date support beyond 2099
+//
+// By default this is enabled for machines using 64 bit pointers only,
+// but it can be enabled by specific ports
+#ifndef MP_SUPPORT_Y2100_AND_BEYOND
```
dpgeorge: This is a configuration setting, so should start with `MICROPY_`.

I think it should be called `MICROPY_EPOCH_SUPPORT_Y2100_AND_BEYOND` and put in `py/mpconfig.h`, along with `MICROPY_EPOCH_IS_1970` and `MICROPY_EPOCH_IS_2000`.

shared/timeutils/timeutils.h
```diff
@@ -27,6 +27,46 @@
...
+#endif
+
+#define MICROPY_EPOCH_IS_2000 (!(MICROPY_EPOCH_IS_1970))
+
+// `platform_timestamp_t` is the type that should be used by the port
+// to represent timestamps, and is referenced to the platform epoch
+#if MP_SUPPORT_Y2100_AND_BEYOND || MICROPY_EPOCH_IS_2000
+typedef int64_t platform_timestamp_t;
+#else
+typedef mp_uint_t platform_timestamp_t;
```
dpgeorge: This type is used in `extmod` code so (due to code layering) I think it's better to define it in the core in `py/mpconfig.h`.

And also call it `mp_platform_timestamp_t` or even just `mp_timestamp_t` (that matches eg `mp_off_t`).

shared/timeutils/timeutils.h
```diff
@@ -27,6 +27,46 @@
...
+#endif
+
+// `timeutils_timestamp_t` is the type used internally by timeutils to
+// represent timestamps, and is always referenced to 1970.
+#if MP_SUPPORT_Y2100_AND_BEYOND
+typedef int64_t timeutils_timestamp_t;
+#else
+typedef mp_uint_t timeutils_timestamp_t;
```
dpgeorge, line 41: This type is only used in `timeutils` so can stay named as is and defined here :+1:

py/obj.c
```diff
@@ -314,6 +314,36 @@ mp_int_t mp_obj_get_int(mp_const_obj_t arg) {
...
+        mp_obj_t as_int = mp_unary_op(MP_UNARY_OP_INT_MAYBE, (mp_obj_t)arg);
+        if (as_int == MP_OBJ_NULL) {
+            mp_raise_TypeError_int_conversion(arg);
+        }
+        arg = as_int;
+    }
+    return mp_obj_int_get_uint_checked(arg);
+}
+
+int64_t mp_obj_get_ll(mp_const_obj_t arg) {
```
dpgeorge: I would call this `mp_obj_get_int64()` because that's what it is.

Or, this function should return `long long` type.

dpgeorge:
Running the tests on PYBD_SF6 everything passes except `tests/ports/stm32/modtime.py` which has the following diff for its output:
```diff
 time.localtime( -1 ) returned (1999, 12, 31, 23, 59, 59, 4, 365) (pass)
 time.localtime( 447549467 ) returned (2014, 3, 7, 23, 17, 47, 4, 66) (pass)
 time.localtime( -940984933 ) returned (1970, 3, 7, 23, 17, 47, 5, 66) (pass)
-time.localtime( -1072915199 ) returned (1966, 1, 1, 0, 0, 1, 5, 1) (pass)
-time.localtime( -1072915200 ) returned (1966, 1, 1, 0, 0, 0, 5, 1) (pass)
-time.localtime( -1072915201 ) returned (1965, 12, 31, 23, 59, 59, 4, 365) (pass)
+time.localtime( -1072915199 ) returned (2102, 2, 6, 6, 28, 17, 1, 37) expecting (1966, 1, 1, 0, 0, 1, 5, 1)
+time.localtime( -1072915200 ) returned (2102, 2, 6, 6, 28, 16, 1, 37) expecting (1966, 1, 1, 0, 0, 0, 5, 1)
+time.localtime( -1072915201 ) returned (2102, 2, 6, 6, 28, 15, 1, 37) expecting (1965, 12, 31, 23, 59, 59, 4, 365)
```
I'm not sure what to do here?  The new behaviour is probably more correct, but it's a breaking change.

py/obj.c
yoctopuce: OK, will do. Should that be a separate commit be in a separate PR as well, or a separate commit in the same PR ?

shared/timeutils/timeutils.h
yoctopuce: Makes sense. I wasn't sure how much visibility you wanted to give to that option, but I am happy to put in `mpconfig.h`

py/obj.c
yoctopuce: I used `int64_t` because this is what other functions in `timeutils` were already using, and because I find it clear. However, from another perspective, there is an existing `mpz` function which is using the type `long long` to do the reverse operation (`mp_obj_new_int_from_ll()`), so it might make more sense to use `long long` as well.

Which option do you prefer ?

yoctopuce:
> Running the tests on PYBD_SF6 everything passes except `tests/ports/stm32/modtime.py` which has the following diff for its output:
> 
> ```diff
>  time.localtime( -1 ) returned (1999, 12, 31, 23, 59, 59, 4, 365) (pass)
>  time.localtime( 447549467 ) returned (2014, 3, 7, 23, 17, 47, 4, 66) (pass)
>  time.localtime( -940984933 ) returned (1970, 3, 7, 23, 17, 47, 5, 66) (pass)
> -time.localtime( -1072915199 ) returned (1966, 1, 1, 0, 0, 1, 5, 1) (pass)
> -time.localtime( -1072915200 ) returned (1966, 1, 1, 0, 0, 0, 5, 1) (pass)
> -time.localtime( -1072915201 ) returned (1965, 12, 31, 23, 59, 59, 4, 365) (pass)
> +time.localtime( -1072915199 ) returned (2102, 2, 6, 6, 28, 17, 1, 37) expecting (1966, 1, 1, 0, 0, 1, 5, 1)
> +time.localtime( -1072915200 ) returned (2102, 2, 6, 6, 28, 16, 1, 37) expecting (1966, 1, 1, 0, 0, 0, 5, 1)
> +time.localtime( -1072915201 ) returned (2102, 2, 6, 6, 28, 15, 1, 37) expecting (1965, 12, 31, 23, 59, 59, 4, 365)
> ```
> 
> I'm not sure what to do here? The new behaviour is probably more correct, but it's a breaking change.

The proper solution is probably to add an extra setting `MP_SUPPORT_Y1969_AND_BEFORE`, that provides a similar extension of the supported timerange to older date. Given that the bit size of timestamps is already parametric, it should not be a big problem to add that.

py/obj.c
dpgeorge: A separate commit in this PR is fine.

py/obj.c
dpgeorge: I suggest changing he type to `long long` and keeping the function name.

dpgeorge:
> The proper solution is probably to add an extra setting `MP_SUPPORT_Y1969_AND_BEFORE`, that provides a similar extension of the supported timerange to older date.

OK, that sounds reasonable.  If you can keep that test passing, that would make things simpler.

py/objint_longlong.c
```diff
@@ -295,6 +295,22 @@ mp_int_t mp_obj_int_get_checked(mp_const_obj_t self_in) {
...
+        }
+    } else {
+        const mp_obj_int_t *self = self_in;
+        long long value = self->val;
+        mp_uint_t truncated = (mp_uint_t)value;
+        if (value >= 0 && (long long)truncated == value) {
+            return truncated;
+        }
+    }
+    mp_raise_msg(&mp_type_OverflowError, MP_ERROR_TEXT("result overflows uint storage"));
```
dpgeorge: For the error message, for consistency please use the same as `objint_mpz.c`: "overflow converting long int to machine word".

py/mpconfig.h
```diff
@@ -888,6 +888,64 @@ typedef double mp_float_t;
...
+#define MICROPY_TIMESTAMP_IMPL (MICROPY_TIMESTAMP_IMPL_LONG_LONG)
+#else
+#define MICROPY_TIMESTAMP_IMPL (MICROPY_TIMESTAMP_IMPL_UINT)
+#endif
+#endif
+
+// `timestamp_t` is the type that should be used by the port
+// to represent timestamps, and is referenced to the platform epoch
+#if MICROPY_TIMESTAMP_IMPL == MICROPY_TIMESTAMP_IMPL_LONG_LONG
+typedef long long timestamp_t;
```
dpgeorge: Please call this new (public) type `mp_timestamp_t`.

tests/extmod/time_mktime.py
```diff
@@ -0,0 +1,124 @@
+# test conversion from date tuple to timestamp and back
+
+try:
+    import time
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+try:
+    time.localtime(1)
```
dpgeorge: I don't think the function needs to be called, just `time.localtime` is enough to check it exists.

Also, you can simplify both checks like this:
```py
try:
    import time

    time.localtime
except (ImportError, AttributError):
    print("SKIP")
    raise SystemExit
```

dpgeorge (commented):
This is looking very good now.

Tested on PYBD_SF6 and all tests now pass.

Also tested on ESP8266_GENERIC and all tests pass.

Just a few comments above to address.

py/objint_longlong.c
yoctopuce: Sure, done.

py/mpconfig.h
yoctopuce: Makes sense, I have changed it.

yoctopuce:
Rebased to master to fix the codecov report compared to master. Should be fine now.

dpgeorge (approved):
This is really nice improvement, thanks for your efforts here!

yoctopuce:
> This is really nice improvement, thanks for your efforts here!

You're welcome, thank you for helping to make it better !

PR #17369, opened 2025-05-27
unix/modsocket: Add "sendall" method to socket class.

agatti:
### Summary

This PR adds an implementation for "socket.sendall" to the Unix port.

Right now the implementation is a wrapper over a single "socket.send" call, since it wasn't possible to let "socket.send" perform a partial transfer.  With no partial transfers it is not possible to have a testable implementation following the expected behaviour, so a single send operation is done and OSErr(EINTR) is raised if a partial transfer occurs.

The discussion for the issue linked to this PR contains more information about the efforts made to let partial transfers happen.

This fixes #16803.

### Testing

A new test is introduced to exercise `socket.sendall`, called `tests/multi_net/tcp_sendall.py`.  This could have been folded into `tests/multi_net/tcp_data.py`, but the Zephyr port doesn't have a `socket.sendall` implementation.

There's a modest size increase.  The footprint increase could have been smaller if `socket.sendall` was aliased to `socket.send` like the CC3200 port does, but then the behaviour would be slightly incompatible with CPython (`socket.sendall` is supposed to return `None`, not the sent bytes count).

### DIFF
--- a/ports/unix/modsocket.c
+++ b/ports/unix/modsocket.c
@@ -333,6 +333,12 @@ static mp_obj_t socket_send(size_t n_args, const mp_obj_t *args) {
     mp_obj_socket_t *self = MP_OBJ_TO_PTR(args[0]);
     int flags = 0;
 
+    bool is_sendall = false;
+    if (n_args < 2) {
+        is_sendall = true;
+        n_args += 2;
+    }
+
     if (n_args > 2) {
         flags = mp_obj_get_int(args[2]);
     }
@@ -342,9 +348,18 @@ static mp_obj_t socket_send(size_t n_args, const mp_obj_t *args) {
     ssize_t out_sz;
     MP_HAL_RETRY_SYSCALL(out_sz, send(self->fd, bufinfo.buf, bufinfo.len, flags),
         mp_raise_OSError(err));
+    if (is_sendall && ((size_t)out_sz != bufinfo.len)) {
+        mp_raise_OSError(MP_EINTR);
+    }
     return MP_OBJ_NEW_SMALL_INT(out_sz);
 }
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(socket_send_obj, 2, 3, socket_send);
+
+static mp_obj_t socket_sendall(size_t n_args, const mp_obj_t *args) {
+    socket_send(n_args - 2, args);
+    return mp_const_none;
+}
+static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(socket_sendall_obj, 2, 3, socket_sendall);
 
 static mp_obj_t socket_sendto(size_t n_args, const mp_obj_t *args) {
     mp_obj_socket_t *self = MP_OBJ_TO_PTR(args[0]);
@@ -511,6 +526,7 @@ static const mp_rom_map_elem_t socket_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_recv), MP_ROM_PTR(&socket_recv_obj) },
     { MP_ROM_QSTR(MP_QSTR_recvfrom), MP_ROM_PTR(&socket_recvfrom_obj) },
     { MP_ROM_QSTR(MP_QSTR_send), MP_ROM_PTR(&socket_send_obj) },
+    { MP_ROM_QSTR(MP_QSTR_sendall), MP_ROM_PTR(&socket_sendall_obj) },
     { MP_ROM_QSTR(MP_QSTR_sendto), MP_ROM_PTR(&socket_sendto_obj) },
     { MP_ROM_QSTR(MP_QSTR_setsockopt), MP_ROM_PTR(&socket_setsockopt_obj) },
     { MP_ROM_QSTR(MP_QSTR_setblocking), MP_ROM_PTR(&socket_setblocking_obj) },
--- a/tests/multi_net/tcp_sendall.py
+++ b/tests/multi_net/tcp_sendall.py
@@ -1,0 +1,71 @@
+# Simple test of a TCP server and client transferring data using sendall.
+
+import socket
+
+if not hasattr(socket.socket, "sendall"):
+    print("SKIP")
+    raise SystemExit
+
+PORT = 8000
+
+# This is the smallest send buffer size supported by Linux, see socket(7).
+SEND_SIZE = 1024
+BUFFERS = 8
+
+
+# G. D. Nguyen, "Fast CRCs", vol. 18, no. 10, pp. 1321-1331, Oct. 2009
+# https://dl.acm.org/doi/abs/10.1109/TC.2009.83
+def calculate_checksum(buffer):
+    checksum = 0
+    for byte in buffer:
+        checksum = (checksum ^ byte) & 0xFFFF
+        for step in range(16):
+            if (checksum & 0x8000) != 0:
+                checksum = ((checksum << 1) ^ 0x8005) & 0xFFFF
+            else:
+                checksum = (checksum << 1) & 0xFFFF
+    return checksum
+
+
+# Server
+def instance0():
+    multitest.globals(IP=multitest.get_network_ip())
+    s = socket.socket()
+    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    s.bind(socket.getaddrinfo("0.0.0.0", PORT)[0][-1])
+    s.listen()
+    multitest.next()
+    s2, _ = s.accept()
+    buffer = b""
+    while len(buffer) < SEND_SIZE * BUFFERS:
+        buffer += s2.recv(SEND_SIZE * BUFFERS)
+    s2.send(b"%08x" % len(buffer))
+    s2.send(b"%04x" % calculate_checksum(buffer))
+    s2.close()
+    s.close()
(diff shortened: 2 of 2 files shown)


ports/unix/modsocket.c
```diff
@@ -323,26 +323,50 @@ static mp_obj_t socket_recvfrom(size_t n_args, const mp_obj_t *args) {
...
+    mp_buffer_info_t bufinfo;
+    mp_get_buffer_raise(args[1], &bufinfo, MP_BUFFER_READ);
+    return socket_send_buffer(n_args, args, &bufinfo);
+}
 static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(socket_send_obj, 2, 3, socket_send);
 
+// Note: besides flag param, this differs from write() in that
+// this does not swallow blocking errors (EAGAIN, EWOULDBLOCK) -
+// these would be thrown as exceptions.
+static mp_obj_t socket_sendall(size_t n_args, const mp_obj_t *args) {
```
dpgeorge, line 358: A tricky way to reduce code size would be to implement this function as:
```c
 static mp_obj_t socket_sendall(size_t n_args, const mp_obj_t *args) {
    socket_send(n_args - 2, args);
    return mp_const_none;
}
static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(socket_sendall_obj, 2, 3, socket_sendall);
```
Then `socket_send` would have some logic along the lines of:
```c
    bool is_sendall = false;
    if (n_args < 2) {
        n_args += 2;
        is_sendall = true;
    }
```
then if `is_sendall` is true, it will check the number of bytes send equals the size of the buffer (and raise EINTR if not).

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,75 @@
...
+# Server
+def instance0():
+    multitest.globals(IP=multitest.get_network_ip())
+    s = socket.socket()
+    s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    s.bind(socket.getaddrinfo("0.0.0.0", PORT)[0][-1])
+    s.listen()
+    multitest.next()
+    s2, _ = s.accept()
+    buffer = s2.recv(SEND_SIZE * BUFFERS)
```
dpgeorge: I tested this on PYBD_SF6 and it doesn't work.  This `recv` returns a short read (it's allowed to).

The following patch gets it working:
```diff
--- a/tests/multi_net/tcp_sendall.py
+++ b/tests/multi_net/tcp_sendall.py
@@ -40,7 +40,9 @@ def instance0():
     s.listen()
     multitest.next()
     s2, _ = s.accept()
-    buffer = s2.recv(SEND_SIZE * BUFFERS)
+    buffer = b""
+    while len(buffer) < SEND_SIZE * BUFFERS:
+        buffer += s2.recv(SEND_SIZE * BUFFERS)
     s2.send(rjust_hex(len(buffer)))
     checksum = 0
     for byte in buffer:
```
But, probably it should add some explicit timeout so that while loop doesn't run forever.

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,75 @@
...
+
+PORT = 8000
+
+# This is the smallest send buffer size supported by Linux, see socket(7).
+SEND_SIZE = 1024
+BUFFERS = 8
+
+
+def rjust_hex(v):
+    return (("00000000" + hex(v)[2:])[-8:]).encode("us-ascii")
```
dpgeorge: This seems a bit complicated.  Does `b"%08x" % v` work instead?

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,75 @@
+# Simple test of a TCP server and client transferring data using sendall.
+
+import socket
+
+try:
+    s = socket.socket()
+    if not hasattr(s, "sendall"):
```
dpgeorge: This doesn't need to create the socket, because `socket.socket` is a type.  Eg can just do `not hasattr(socket.socket, "sendall")`.

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,75 @@
...
+# This is the smallest send buffer size supported by Linux, see socket(7).
+SEND_SIZE = 1024
+BUFFERS = 8
+
+
+def rjust_hex(v):
+    return (("00000000" + hex(v)[2:])[-8:]).encode("us-ascii")
+
+
+def calculate_checksum(size):
```
dpgeorge: Suggest this takes the buffer and computes the checksum on the buffer (see also below).

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,75 @@
...
+    s.bind(socket.getaddrinfo("0.0.0.0", PORT)[0][-1])
+    s.listen()
+    multitest.next()
+    s2, _ = s.accept()
+    buffer = s2.recv(SEND_SIZE * BUFFERS)
+    s2.send(rjust_hex(len(buffer)))
+    checksum = 0
+    for byte in buffer:
+        checksum += byte
+    s2.send(rjust_hex(checksum))
```
dpgeorge: Suggest this uses the checksum function, eg `s2.send(rjust_hex(calculate_checksum(buffer)))`.

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,75 @@
...
+    for index in range(len(buffer)):
+        buffer[index] = index & 0xFF
+    s.sendall(buffer)
+    length = s.recv(8)
+    if int(length, 16) != len(buffer):
+        print("sendall len fail", length, len(buffer))
+    else:
+        print("sendall len pass")
+    checksum = s.recv(8)
+    if checksum != CHECKSUM:
```
dpgeorge: Instead of "hard-coding" the CHECKSUM, suggest using the function like `calculate_checksum(buffer)`.

dpgeorge (commented):
Thanks for opening this PR.  If we can get the code size down, that would be great (see my suggestion).

dpgeorge:
@agatti you may have missed the above review :smile:

agatti:
Looks like I did indeed :)  Thanks for the reminder (and for the extensive review).

I've revised both the code and the multitest, using your suggestion to alias `sendall` to `send`, along rebasing it to current master to resolve merge conflicts.  The multitest now expects a proper implementation of `socket.settimeout` to work reliably, but if that's not the case on all ports I can make this multithreaded/async.

Still, I've tested this on both linux and an esp32 - hope that is good enough now.

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,84 @@
...
+    s2.settimeout(RECV_TIMEOUT)
+    buffer = b""
+    time_start = int(time.time())
+    while len(buffer) < SEND_SIZE * BUFFERS:
+        buffer += s2.recv(SEND_SIZE * BUFFERS)
+        current_time = int(time.time())
+        if (current_time - time_start) > TEST_TIMEOUT:
+            raise Exception("test didn't stop after %d seconds" % TEST_TIMEOUT)
+        elif current_time < time_start:
+            raise Exception("time zone change?")
```
dpgeorge: I know I suggested it, but.... I don't think this test should have a timeout.  The `run-multitests.py` script already has a timeout: if one of the instances doesn't print anything for 10 seconds then it aborts the test.

So `TEST_TIMEOUT` here will never be hit.

Also `s2.settimeout(RECV_TIMEOUT)` will never be reached and is not needed.

Removing both those timeouts will simplify the test.

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,84 @@
...
+    s2.send(b"%08x" % len(buffer))
+    s2.send(b"%04x" % calculate_checksum(buffer))
+    s2.close()
+    s.close()
```
dpgeorge, line 45: Suggest adding `print("received {} bytes".format(len(buffer)))` at the end, just so there's at least one line of output.

tests/multi_net/tcp_sendall.py
```diff
@@ -0,0 +1,84 @@
...
+    s = socket.socket()
+    s.connect(socket.getaddrinfo(IP, PORT)[0][-1])
+    # Try our best to trigger a partial send in socket.sendall
+    if hasattr(socket, "SO_SNDBUF"):
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, SEND_SIZE)
+    buffer = bytearray(SEND_SIZE * BUFFERS)
+    for index in range(len(buffer)):
+        buffer[index] = index & 0xFF
+    expected_checksum = calculate_checksum(buffer)
+    s.sendall(buffer)
```
dpgeorge, line 61: Suggest adding `print("sent {} bytes".format(len(buffer)))`.

tests/multi_net/tcp_sendall.py
agatti: Got it - unfortunately it's the first time I actually write a multinet test so I missed all the existing timeout bits.

I've updated the test with the current suggestions and it still works on Unix at least.

dpgeorge (approved):
Thanks for updating.

Tested on PYBD_SF6 <-> unix and the new test passes.

PR #17296, opened 2025-05-13
tests/run-tests.py: Add test statistics to output _result.json file.

dpgeorge:
### Summary

This commit add some simple test statistics to the output `_result.json` file generated at the end of a test run.

This will be useful for Octoprobe (automated hardware testing framework), to report a summary of the number of tests passed/skipped/failed.

### Testing

Running a few tests on an esp8266:
```
$ ./run-tests.py -t u0 basics/0prelim.py extmod/vfs_rom.py extmod/vfs_lfs.py extmod/vfs_posix.py extmod/vfs_blockdev_invalid.py
platform=esp8266 arch=xtensa inlineasm=xtensa
pass  basics/0prelim.py 
FAIL  extmod/vfs_rom.py 
NOTE: extmod/vfs_rom.py may be a unittest that doesn't run unittest.main()
skip  extmod/vfs_lfs.py
skip  extmod/vfs_posix.py
FAIL  extmod/vfs_blockdev_invalid.py 
3 tests performed (31 individual testcases)
1 tests passed
2 tests skipped: vfs_lfs vfs_posix
2 tests failed: vfs_blockdev_invalid vfs_rom
```
The output `_result.json` is:
```
$ cat results/_results.json|jq
{                             
  "args": {
    "test_instance": "u0",
    "baudrate": 115200,
    "user": "micro",
    "password": "python",
    "test_dirs": null,
    "result_dir": "micropython/tests/results",
    "filters": [],
    "emit": "bytecode",
    "heapsize": null,
    "via_mpy": false,
    "mpy_cross_flags": "-march=xtensa",
    "keep_path": false,
    "jobs": 8,
    "files": [
      "basics/0prelim.py",
      "extmod/vfs_rom.py",
      "extmod/vfs_lfs.py",
      "extmod/vfs_posix.py",
      "extmod/vfs_blockdev_invalid.py"
    ],
    "print_failures": false,
    "clean_failures": false,
    "run_failures": false,
    "platform": "esp8266",
    "arch": "xtensa",
    "inlineasm_arch": "xtensa"
  },
  "failed_tests": [
    "extmod/vfs_blockdev_invalid.py",
    "extmod/vfs_rom.py"
  ],
  "statistics": {
    "total": 5,
    "pass": 1,
    "fail": 2,
    "skip": 2
  }
}
```

EDIT: implementation now changed to output the entire list of passed/skipped/failed tests.  The above output is now changed to:
```
{                             
  "args": {
    "test_instance": "u0",
[...]

PR #17373, opened 2025-05-28
tests/run-tests.py: Change _results.json to have a combined result list.

dpgeorge:
### Summary

The `_results.json` output of `run-tests.py` was recently changed in #17296 to add a list of passed and skipped tests.

The way this was done turned out to be not general enough, because in #17361 we want to add another type of result, namely tests that are skipped because they are too large.

Instead of having separate lists in `_results.json` for each kind of result (pass, fail, skip, skip too large, etc), this PR changes the output form of `_results.json` so that it stores a single list of 3-tuples of all tests that were run:
```
[(test_name, result, reason), ...]
```
That's more general and allows adding a reason for skipped and failed tests.  At the moment this reason is just an empty string, but can be improved in the future.

### Testing

Ran `run-tests.py` on a board with passing, failing and skipping tests and verified the output.  Also ran `run-tests.py --run-failures` to check that it could re-read the failed test list.

Could instead keep three separate lists (pass, fail, skip) and have them be a list of 2-tuples of the form `(test_name, reason)`.  But I think the approach in this PR is more general, with just one data structure of results.

Note that the change in #17296 was very recent and not in a release so can be changed without backwards compatibility issues.

### DIFF
--- a/tests/run-tests.py
+++ b/tests/run-tests.py
@@ -605,9 +605,7 @@ class PyboardNodeRunner:
 def run_tests(pyb, tests, args, result_dir, num_threads=1):
     test_count = ThreadSafeCounter()
     testcase_count = ThreadSafeCounter()
-    passed_tests = ThreadSafeCounter([])
-    failed_tests = ThreadSafeCounter([])
-    skipped_tests = ThreadSafeCounter([])
+    test_results = ThreadSafeCounter([])
 
     skip_tests = set()
     skip_native = False
@@ -896,7 +894,7 @@ def run_tests(pyb, tests, args, result_dir, num_threads=1):
 
         if skip_it:
             print("skip ", test_file)
-            skipped_tests.append((test_name, test_file))
+            test_results.append((test_name, test_file, "skip", ""))
             return
 
         # Run the test on the MicroPython target.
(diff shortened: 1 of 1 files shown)


dpgeorge:
@hmaerki FYI

tests/run-tests.py
```diff
@@ -1069,9 +1069,15 @@ def to_json(obj):
...
             {
+                # The arguments passed on the command-line.
                 "args": vars(args),
-                "passed_tests": [test[1] for test in passed_tests],
-                "skipped_tests": [test[1] for test in skipped_tests],
+                # A list of all results of the form [(test, result, reason), ...].
+                "results": (
+                    list([test[1], "pass", ""] for test in passed_tests)
+                    + list([test[1], "skip", ""] for test in skipped_tests)
+                    + list([test[1], "fail", ""] for test in failed_tests)
```
hmaerki, line 1078: pytest (and other tools) use 'passed', 'skipped', 'failed'.

hmaerki:
You use 3 lists: passed_tests, skipped_tests and failed_tests.

If it would be one list, the order of test execution would be preserved. So it becomes visible if one test breaks the following test.

But in the context of 'run-tests.py' the current implementation with 3 lists is completely fine..

tests/run-tests.py
dpgeorge: I was trying to keep the output concise, so it takes up less space.  I think it's important not to take up resources if you don't need to.

dpgeorge:
I've now updated this PR to accumulate all tests results in one big list.  So that means the order that the tests are run is retained in the `_results.json` file.

(Your recent adaptation in octoprobe should still work fine with this adjusted PR, the only difference is the order of tests in `_results.json`.  The format stays the same.)

@hmaerki do you like this new approach better?

hmaerki:
I reviewed and tested this PR: https://reports.octoprobe.org/github_selfhosted_testrun_24/octoprobe_summary_report.html

@dpgeorge: Please merge into master.

dpgeorge:
Thanks for review and testing!  Now merged.

PR #17374, opened 2025-05-28
fix: solve the problem that the flash clock frequency is too high and…

Lesords:
### Summary

Reduce the flash clock frequency to avoid the problem of XIAO RP2350 not responding and having no serial port after flashing the firmware.

### Testing

Tested on XIAO RP2350

### DIFF
--- a/ports/rp2/rp2_flash.c
+++ b/ports/rp2/rp2_flash.c
@@ -104,8 +104,8 @@ static bool use_multicore_lockout(void) {
 // and core1 locked out if relevant.
 static void __no_inline_not_in_flash_func(rp2_flash_set_timing_internal)(int clock_hz) {
 
-    // Use the minimum divisor assuming a 133MHz flash.
-    const int max_flash_freq = 133000000;
+    // Use the minimum divisor assuming a 60MHz flash.
+    const int max_flash_freq = 60000000;
     int divisor = (clock_hz + max_flash_freq - 1) / max_flash_freq;
 
     #if PICO_RP2350
@@ -115,7 +115,7 @@ static void __no_inline_not_in_flash_func(rp2_flash_set_timing_internal)(int clo
     }
 
     // RX delay equal to the divisor means sampling at the same time as the next falling edge of SCK after the
-    // falling edge that generated the data.  This is pretty tight at 133MHz but seems to work with the Winbond flash chips.
+    // falling edge that generated the data.  This is pretty tight at 60MHz but seems to work with the Winbond flash chips.
     const int rxdelay = divisor;
     qmi_hw->m[0].timing = (1 << QMI_M0_TIMING_COOLDOWN_LSB) |
         rxdelay << QMI_M1_TIMING_RXDELAY_LSB |


Lesords:
You can get more information through this issue: https://github.com/micropython/micropython/issues/17375

Lesords:
Before this commit https://github.com/micropython/micropython/commit/91cff8e4f10ea665c5f3f4b16d62c98d6ca22037, everything was working fine.

nspsck:
> I think a cleaner approach would be something like:
> 
> ```c
>   // Use the minimum divisor assuming a 133MHz flash.
> #ifdef WAVESHARE_RP2350_ZERO
>   const int max_flash_freq = 60000000;
> #else
>   const int max_flash_freq = 133000000;
> #endif
> ```
> 
> Then we would just add boards to
> 
> ```c
> #ifdef WAVESHARE_RP2350_ZERO
> ```
> 
> And whenever we add a board that is officially supported by `pico-sdk`, we should be adding something like this to `mpconfigboard.cmake`:
> 
> ```
> set(PICO_BOARD "waveshare_rp2350_zero")
> ```

Nvm, something better, assuming `pico-sdk` will not have something like spi clkdiv = 5 or so:
```C
  // Use the minimum divisor assuming a 133MHz flash.
  const int max_flash_freq = 133000000;
  int divisor = (clock_hz + max_flash_freq - 1) / max_flash_freq;

  // For boards with PICO_FLASH_SPI_CLKDIV = 3
#if PICO_FLASH_SPI_CLKDIV == 3
  divisor++;
#endif

  // For boards with PICO_FLASH_SPI_CLKDIV = 4
#if PICO_FLASH_SPI_CLKDIV == 4
  divisor += 2;
#endif
```
or a little harder to read, but more future prove
```C
  // Use the minimum divisor assuming a 133MHz flash.
  const int max_flash_freq = 133000000;
  int divisor = (clock_hz + max_flash_freq - 1) / max_flash_freq + PICO_FLASH_SPI_CLKDIV - 2;
```
This should cover all future problems with other vendors boards, if `pico-sdk` supports their boards. And this results higher flash speed in almost all sense making cases.

dpgeorge:
This definitely needs to be fixed.

It looks like `PICO_FLASH_SPI_CLKDIV` is only used by the stage 2 bootloader.  And it defaults to 4.  Boards that have defined this usually define it to 2 to increase performance of the SPI flash.

> I suggest we make this value a board configuration option that _defaults_ to `133000000 / PICO_FLASH_SPI_CLKDIV`.

Yes, I think this is the right approach.  So a board would define something like:
```c
#define MICROPY_HW_SPIFLASH_MAX_FREQ (133000000)
```
and it would default to:
```c
#ifndef MICROPY_HW_SPIFLASH_MAX_FREQ
#if PICO_RP2040
#define MICROPY_HW_SPIFLASH_MAX_FREQ (125000000 / PICO_FLASH_SPI_CLKDIV)
#else
#define MICROPY_HW_SPIFLASH_MAX_FREQ (150000000 / PICO_FLASH_SPI_CLKDIV)
#endif
#endif
```

Gadgetoid:
> Yes, I think this is the right approach. So a board would define something like:

Basically what I've done here: https://github.com/micropython/micropython/pull/17389

`MICROPY_HW_SPIFLASH_MAX_FREQ` might be a more explicit name for the variable, though!

dpgeorge:
> Basically what I've done here: #17389

Ah, sorry, I missed that!

Let's close this PR then, in favour of that one.

Gadgetoid:
> Ah, sorry, I missed that!

No worries, if you're coming up with basically the same fix then I can't have gone far wrong... y'know apart from adopting the change that broke this in the first place 😆

dpgeorge:
> y'know apart from adopting the change that broke this in the first place

Haha, indeed!

PR #17377, opened 2025-05-28
tools/mpy_ld.py: support absolute relocations in text section on Thumb targets

dpgeorge:
### Summary

Add support for `R_ARM_ABS32` relocations in native .mpy files.

Also, enable float by default on `armv6m` and `armv7m` native mpy targets, since soft-float now works correctly.

Should fix issue #14430.

### Testing

Tested with `ARCH=armv6m` on RPI_PICO_W, and `ARCH=armv7emdp` on PYBD_SF6.  The update `features2` -- calling `sin()` and `cos()` -- works on both these targets.

This is a relatively minimal change, but nonetheless `tools/mpy_ld.py` is getting quite unwieldy now and pretty hard to follow what relocations are/aren't allowed.  Maybe in the future it can be cleaned up, but we'd need more tests before doing that to ensure something doesn't break.

### DIFF
--- a/examples/natmod/random/Makefile
+++ b/examples/natmod/random/Makefile
@@ -14,4 +14,9 @@ ifeq ($(ARCH),xtensa)
 MPY_EXTERN_SYM_FILE=$(MPY_DIR)/ports/esp8266/boards/eagle.rom.addr.v6.ld
 endif
 
+ifeq ($(ARCH),$(filter $(ARCH),armv6m armv7m))
+# Link with libm.a for soft-float helper functions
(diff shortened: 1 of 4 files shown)


jonnor:
I have tested this in emlearn-micropython, with RP2040. All the native modules now build. I checked quickly two of them (emlearn_fft and emlearn_trees) on device, and it also seems to execute successfully. So this does indeed look like it solves #14430 !

dpgeorge:
Thanks @jonnor for testing.  I have now fixed the CI, so this should be ready to merge.

PR #17379, opened 2025-05-28
tools/ci.sh: Enable a few more natmod tests on rv32imc and xtensa.

dpgeorge:
### Summary

The `btree` natmod example can now build for rv32imc, and `random` can build for xtensa if it's linked with the standard C library.

### Testing

Tested by CI.

### DIFF
--- a/examples/natmod/btree/Makefile
+++ b/examples/natmod/btree/Makefile
@@ -32,6 +32,11 @@ SRC += $(addprefix $(realpath $(BTREE_DIR))/,\
 	mpool/mpool.c \
 	)
 
+ifeq ($(ARCH),xtensa)
+# Link with libm.a and libgcc.a from the toolchain
+LINK_RUNTIME = 1
+endif
+
 include $(MPY_DIR)/py/dynruntime.mk
 
 # btree needs gnu99 defined
--- a/examples/natmod/deflate/Makefile
+++ b/examples/natmod/deflate/Makefile
@@ -10,4 +10,9 @@ SRC = deflate.c
 # Architecture to build for (x86, x64, armv7m, xtensa, xtensawin, rv32imc)
 ARCH = x64
 
+ifeq ($(ARCH),xtensa)
+# Link with libm.a and libgcc.a from the toolchain
+LINK_RUNTIME = 1
+endif
+
 include $(MPY_DIR)/py/dynruntime.mk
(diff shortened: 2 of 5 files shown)


dpgeorge:
> The errno fix may not be specific to RV32 but rather to Ubuntu's 24.04 packaged picolibc libraries.

@agatti thanks for the hint, but I gave up on getting things working with RV32.  I don't think it's possible with the picolibc used by the CI (although on my local machine with `nano.specs` it builds fine).

But I did get everything building for xtensa, just by linking with the libc runtime.  Take a look now and see what you think.

agatti:
The Xtensa changes sort of clash with #17091, as in they bring in symbols from libc/libgcc/libm that with #17091 would be picked up from ROM whenever possible.  Maybe for ESP32-series MCUs it's less of an issue (even though pretty much every chip has parts of stdio/libc in ROM, see `components/esp-rom/<mcu>/ld/<mcu>.rom.{libgcc,newlib*,syscalls}.ld` files in ESP-IDF), but on an 8266 saving memory is almost always worthwhile.

That doesn't fully invalidate said PR, but defaulting to libc/libgcc/libm symbols doesn't exercise the PR changes in an automated fashion (I'll have to remove the makefile changes and symbol list fragments from natmods as they aren't useful anymore).

For RV32, I actually got the Ubuntu version wrong - I meant 22.04, but I see the workflow is running on 24.04 now and still has the same issue.  Maybe it'd be easier to patch the berkeleydb submodule to add a picolibc+tls compatible definition of errno instead?

dpgeorge:
> The Xtensa changes sort of clash with #17091

Oh, yes, it does indeed.  Sorry I kind of forgot about #17091 in this context :sweat_smile: 

I can see the changes in this PR are almost exactly a sub-set of #17091 (the only difference is `LINK_RUNTIME=1` is only set here for xtensa, whereas in #17091 it's unconditionally set).

So I'll close this PR then, and we can continue the xtensa discussion over in #17091.


> For RV32, I actually got the Ubuntu version wrong - I meant 22.04, but I see the workflow is running on 24.04 now and still has the same issue. Maybe it'd be easier to patch the berkeleydb submodule to add a picolibc+tls compatible definition of errno instead?

Oh, if we could patch berkeleydb to improve the way it uses errno, that would be great!  Do you want to have a look at doing that?

PR #17382, opened 2025-05-28
zephyr: Complete the UART driver and make it interrupt driven.

danicampora:
### Summary

Before this PR the UART would only work in very simple use cases. Receiving large amounts of data would result
in lost bytes. Plus the print function would crash due to `uart_config_get()` returning incorrect values. Additionally, receiving data with timeout==0 would fail even if data was already available in the internal UART Rx FIFO.

The non-implemented functions have also been made usable.

### Testing

This was tested on the nRF5340DK and the nRF52840DK.

Code size is slightly increased for all Zephyr ports.


ports/zephyr/machine_uart.c
```diff
@@ -60,16 +73,30 @@ static void mp_machine_uart_print(const mp_print_t *print, mp_obj_t self_in, mp_
...
     mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
     mp_arg_parse_all(n_args, pos_args, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
 
+    self->txbuf_size = args[ARG_txbuf].u_int;
+    self->rxbuf_size = args[ARG_rxbuf].u_int;
     self->timeout = args[ARG_timeout].u_int;
     self->timeout_char = args[ARG_timeout_char].u_int;
+
+    self->tx_buffer = m_tracked_calloc(self->txbuf_size, sizeof(uint8_t));
+    ring_buf_init(&self->tx_ringbuffer, self->txbuf_size, self->tx_buffer);
```
dpgeorge: Any reason not to use MicroPython's ringbuffer (see `py/ringbuf.h`)?  I guess it doesn't really matter much, I'm just more familiar with our version rather than zephyr's.

With zephyr's implementation, if you allocate N bytes, does it only give you N-1 slots in the ringbuffer?

ports/zephyr/machine_uart.c
```diff
@@ -86,65 +113,122 @@ static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_arg
 }
 
 static void mp_machine_uart_deinit(machine_uart_obj_t *self) {
-    (void)self;
+    m_tracked_free(self->tx_buffer);
+    m_tracked_free(self->rx_buffer);
```
dpgeorge: Do you need to deinit the IRQ callback here?  Otherwise if it continues to receive characters and the IRQ fires, it puts data in the ringbuf which is now freed.

Also, I think there's a memory leak if you continuously allocate UART objects but never deinit them.  The GC will reclaim the UART but the ringbuffer memory will never be freed.  To fix that can add a finaliser that calls this deinit function (same as you did with `machine.Timer`).

dpgeorge (commented):
This is a nice improvement, thanks!

I tested on a frdm_k64f and the UART works a lot better now, you can actually send and received data.

But I noticed in passing that it's still not possible to set the baudrate.  Is that an easy thing to add?

ports/zephyr/machine_uart.c
danicampora: I forgot we had our own implementation of a ringbuffer. I'll switch to it.

ports/zephyr/machine_uart.c
danicampora: Good points. I'll fix all that.

danicampora:
> But I noticed in passing that it's still not possible to set the baudrate. Is that an easy thing to add?

Yes that's easy to add. I'll do it.

VynDragon:
Definitely looking forward to that one! this has been a significant issue.

danicampora:
@VynDragon working on it. Should be ready to merge very soon.

ports/zephyr/machine_uart.c
```diff
@@ -32,16 +33,35 @@
...
+
+static void uart_interrupt_handler(const struct device *dev, void *user_data);
 
 typedef struct _machine_uart_obj_t {
     mp_obj_base_t base;
     const struct device *dev;
     uint16_t timeout;       // timeout waiting for first char (in ms)
     uint16_t timeout_char;  // timeout waiting between chars (in ms)
+    ringbuf_t rx_ringbuffer;
+    ringbuf_t tx_ringbuffer;
```
danicampora, line 60: @dpgeorge can you confirm that this objects put there like that will be tracked by the gc?

danicampora:
@dpgeorge I made the changes requested. Using py/ringbuf.h now instead of Zephyr's, and the UART can be (re)configured at runtime. One difference to other ports is that separate RTS and CTS flow control methods do not exist with Zephyr, so flow control can only be `None` or `(UART.RTS | UART.CTS)`

ports/zephyr/machine_uart.c
dpgeorge: They will be.

But because this object is used in a Zephyr IRQ handler, if the user code no longer references the UART object -- but Zephyr does in the callback data -- then the object will be reclaimed and the IRQ handler will access invalid memory.

I suggest following the pattern used by `machine_timer.c`.

ports/zephyr/machine_uart.c
```diff
@@ -60,22 +80,93 @@ static void mp_machine_uart_print(const mp_print_t *print, mp_obj_t self_in, mp_
...
+
+    uart_irq_callback_user_data_set(self->dev, uart_interrupt_handler, (void *)self);
+    uart_irq_rx_enable(self->dev);
 }
 
 static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
     mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
 
-    machine_uart_obj_t *self = mp_obj_malloc(machine_uart_obj_t, &machine_uart_type);
+    machine_uart_obj_t *self = mp_obj_malloc_with_finaliser(machine_uart_obj_t, &machine_uart_type);
```
dpgeorge, line 167: You'll also need to add a `__del__` method.  I suggest following `machine_timer.c`.

dpgeorge:
> One difference to other ports is that separate RTS and CTS flow control methods do not exist with Zephyr, so flow control can only be `None` or `(UART.RTS | UART.CTS)`

That's OK.

ports/zephyr/machine_uart.c
danicampora: I did, it's in:

```
#define MICROPY_PY_MACHINE_UART_CLASS_CONSTANTS \
    { MP_ROM_QSTR(MP_QSTR_RTS), MP_ROM_INT(MACHINE_UART_RTS) }, \
    { MP_ROM_QSTR(MP_QSTR_CTS), MP_ROM_INT(MACHINE_UART_CTS) }, \
    { MP_ROM_QSTR(MP_QSTR___del__), MP_ROM_PTR(&machine_uart_deinit_obj) },
```

danicampora: Or is there a cleaner way to add it?

ports/zephyr/machine_uart.c
danicampora: > I suggest following the pattern used by machine_timer.c.

I thought I did. The `__del__` will be called and then interrupts will be disabled, so that scenario should never happen.

dpgeorge: Ah, I didn't see the `__del__`!  That's a neat way to do it.

Let me do some testing of this PR.

dpgeorge:
I tested this on `nucleo_wb55rg` using a modified version of `tests/extmod/machine_uart_tx.py`:
```py
import time
from machine import UART

uart_id = "lpuart1"
pins = {}  # TX/RX=PC0/PC1 -- pins need to be bridged

for bits_per_s in (2400, 9600, 115200):
    text = "Hello World"
    uart = UART(uart_id, bits_per_s, bits=8, parity=None, stop=1, **pins)

    start_us = time.ticks_us()
    uart.write(text)
    uart.flush()
    duration_us = time.ticks_diff(time.ticks_us(), start_us)
    data = uart.read()

    bits_per_char = 10
    expect_us = (len(text)) * bits_per_char * 1_000_000 // bits_per_s
    print("bps:", bits_per_s, "data:", data, "duration:", duration_us, "expect:", expect_us, "delta:", duration_us - expect_us)
```
That configures the UART at different baudrates, writes some data, reads it back (UART should have its TX/RX pins connected via a wire) and times how long the TX took.

Results with this PR:
```
bps: 2400 data: b'Hello Worl' duration: 46310 expect: 45833 delta: 477
bps: 9600 data: b'Hello Worl' duration: 11380 expect: 11458 delta: -78
bps: 115200 data: b'Hello World' duration: 1235 expect: 954 delta: 281
```
You can see that the flush does not fully work, it's returning too early (for the lower baudrates) and the full "Hello World" is not yet transmitted.

This isn't a show stopper but would be good to fix.  Maybe use `uart_irq_tx_complete()` to set a flag, and wait on that flag when waiting for the FLUSH.  There's some code in `subsys/modbus/modbus_serial.c` that uses `uart_irq_tx_complete()` which might be useful.

danicampora:
@dpgeorge thanks for testing. That should be easy to fix. I'll do it asap.

ports/zephyr/machine_uart.c
```diff
@@ -199,7 +202,9 @@ static mp_uint_t mp_machine_uart_read(mp_obj_t self_in, void *buf_in, mp_uint_t
     do {
         int _rx_len = MIN(ringbuf_avail(&self->rx_ringbuffer), size - bytes_read);
         if (_rx_len > 0) {
+            uint32_t key = irq_lock();
```
danicampora: @dpgeorge while I was fixing the flush problem it came to my mind that we need this critical section around `ringbuf_get_bytes()` here. Because if this call is interrupted while characters are being received then things can go wrong with the ringbuffer. Or am I wrong?

danicampora:
@dpgeorge I added a flag that is set in the ISR and now the test also passes at low baudrates. But I have a doubt if the ringbuffer implementation is interrupt/thread-safe...

ports/zephyr/machine_uart.c
dpgeorge: The ringbuf API is designed to be inherently race free.

The getting and putting from/to a ring buffer is independent -- because `iput` and `iget` indices are independent -- so it's possible for the thread-level code here to do `ringbuf_get_bytes()` and be interrupted by a call to `ringbuf_put_bytes()`.  (Note that the result of `ringbuf_avail()` may grow due to the IRQ, but it will never shrink, so it's safe to read as many bytes as `ringbuf_avail()` previously returned.)

So, the `irq_lock()` is unnecessary.

ports/zephyr/machine_uart.c
```diff
@@ -86,65 +180,127 @@ static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_arg
...
 static mp_int_t mp_machine_uart_any(machine_uart_obj_t *self) {
-    (void)self;
-    mp_raise_NotImplementedError(NULL); // TODO
+    return ringbuf_avail(&self->rx_ringbuffer);
 }
 
 static bool mp_machine_uart_txdone(machine_uart_obj_t *self) {
-    (void)self;
-    mp_raise_NotImplementedError(NULL); // TODO
+    return (ringbuf_avail(&self->tx_ringbuffer) || (self->tx_complete == false)) ? false : true;
```
dpgeorge: IMO it's simpler to write:
```c
return self->tx_complete && !ringbuf_avail(...);
```

danicampora: done, thanks!

ports/zephyr/machine_uart.c
danicampora: Nice, thanks for explaining!

danicampora:
@dpgeorge I removed the irq lock and made the other improvement, so this should be ready to merge.

dpgeorge:
OK, thanks, I'll retest and review.

ports/zephyr/machine_uart.c
```diff
@@ -60,23 +77,95 @@ static void mp_machine_uart_print(const mp_print_t *print, mp_obj_t self_in, mp_
...
+    uart_irq_callback_user_data_set(self->dev, uart_interrupt_handler, (void *)self);
+    uart_irq_rx_enable(self->dev);
 }
 
 static mp_obj_t mp_machine_uart_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
     mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
 
-    machine_uart_obj_t *self = mp_obj_malloc(machine_uart_obj_t, &machine_uart_type);
+    machine_uart_obj_t *self = mp_obj_malloc_with_finaliser(machine_uart_obj_t, &machine_uart_type);
     self->dev = zephyr_device_find(args[0]);
```
dpgeorge: If `zephyr_device_find()` raises an exception (eg unknown UART id) then `self->dev` has an undefined value, and when the GC reclaims this object it tries to disable IRQs on an undefined object, leading to a crash.

Eg on frdm_k64f, do `machine.UART(1)` and then soft reset, and it'll lock up.

Suggested fix:
```c
const struct device *dev = zephyr_device_find(args[0]);

machine_uart_obj_t *self = mp_obj_malloc_with_finaliser(machine_uart_obj_t, &machine_uart_type);
self->dev = dev;
self->tx_complete = true;
```
(That follows how `machine_i2c.c`, `machine_spi.c` etc do it.)

danicampora: Oh I see... Thanks for testing such scenario. I'll fix that.

dpgeorge: I only found it by accident, running a test that was meant for a different board!

dpgeorge (approved):

PR #17383, opened 2025-05-28
Fix parsing complex numbers like -1+1j

jepler:
### Summary

Over in https://github.com/adafruit/circuitpython/pull/10297 a contributor noticed a bug and offered a fix for parsing of complex numbers like `complex("-1+1j")`. The fix looked minimal and based on testing seems correct, so I decided to offer it over here.

### Testing

I added a pair of initially failing tests to the test suite, and then checked that the fix made them pass.

I also noticed that the non-standard complex literal `complex("1 1j")` is accepted by micropython but not python3. It is accepted as an argument to the built in `complex` function but not as a literal in source code. A more complete fix might make this into an error too.


dpgeorge (approved):
I like this fix.  It's definitely needed to fix the bug, and is very minimal.

dpgeorge:
Rebased, squashed and merged in 2ce63b142068fc04aae701ae78af683736dc179c

Thanks for the neat fix!

PR #17384, opened 2025-05-28
Alternate fix for complex() parsing

jepler:
In contrast to https://github.com/micropython/micropython/pull/17383 this one _also_ fixes parsing for `complex("1 1j")` where a sign character does not occur between the real and imaginary parts of a number with both parts. It almost certainly makes the code grow more. Please feel free to close this if the code growth isn't worth the benefit. If so, a section should probably be added to the difference page in the docs.


py/parsenum.c
```diff
@@ -233,6 +233,8 @@ mp_obj_t mp_parse_num_float(const char *str, size_t len, bool allow_imag, mp_lex
     unsigned int real_imag_state = REAL_IMAG_STATE_START;
     mp_float_t dec_real = 0;
 parse_start:
+    bool dec_pos = false;
```
dpgeorge, line 236: You can't put new variable declarations right after a label.

jepler:
Updated. I have seen this technique used but since I didn't see a diagnostic on the unix build (which allows it as an extension) I didn't realize it was needed here.

Future gcc will have a way to enable a diagnostic about this incorrect label usage even when not running in pedantic mode: https://gcc.gnu.org/pipermail/gcc-patches/2024-November/669243.html

dpgeorge:
Thanks for opening this for a comparison with #17383.  Considering the relatively large increase in code size, I'll close this in favour of #17383.

At least with this PR, `complex("1 1j")` isn't supposed to be a correct number, so it's not like MicroPython is giving back an unexpectedly incorrect result (unlike `complex("-1+1j")` which was clearly wrong).

jepler:
Makes perfect sense, thanks!

PR #17386, opened 2025-05-29
ports/esp32: esp32.wake_on_ulp should only be available on boards

meirarmon:
ports/esp32: esp32.wake_on_ulp should only be available on boards
          that have SOC_ULP_SUPPORTED=y
docs/library/esp32.rst: update docs to reflect that wake_on_ulp
          is only available for boards that support ulp coprocessor

### Summary

Having code wrapped by definitions of whether or not the capability is supported on the
board instead of what board it is, is better practice.
I introduced this so that when using the method wake_on_ulp on boards that don't
support it, it will be clear that it is not supported, instead of silently failing.

### Testing

I tested this on the ESP32C3

### DIFF
--- a/docs/library/esp32.rst
+++ b/docs/library/esp32.rst
@@ -23,6 +23,8 @@ Functions
     Configure whether or not the Ultra-Low-Power co-processor can wake the
     device from sleep. *wake* should be a boolean value.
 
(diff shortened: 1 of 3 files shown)


dpgeorge (approved):
This is good improvement, thanks!

Issue #17375, opened 2025-05-28
RP2350: Flash clock frequency problem


Lesords:
### Port, board and/or hardware

XIAO RP2350

### MicroPython version

MicroPython v1.26.0-preview.159.g0c623fe on 2025-05-28

And most of the firmware for this interface:
https://micropython.org/download/SEEED_XIAO_RP2350/

### Reproduction

When using the firmware of this interface, you will find that the device has no serial port.

https://micropython.org/download/SEEED_XIAO_RP2350/

### Expected behaviour

Merge this PR：https://github.com/micropython/micropython/pull/17374

Or give better suggestions

### Observed behaviour

When using the firmware of this interface, you will find that the device has no serial port. 
https://micropython.org/download/SEEED_XIAO_RP2350/

After my test, I found that it is not available after this commit: https://github.com/micropython/micropython/commit/91cff8e4f10ea665c5f3f4b16d62c98d6ca22037

And this commit will cause the PICO_FLASH_SPI_CLKDIV macro definition to be invalid, you can view this macro through this link: https://github.com/raspberrypi/pico-sdk/blob/master/src/boards/include/boards/seeed_xiao_rp2350.h#L137

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

dpgeorge:
Fixed by b725c26b4e351b71b8f72ad8e55bbbd5ccce4602

PR #17389, opened 2025-05-29
rp2/rp2_flash: Add MICROPY_HW_FLASH_MAX_FREQ.

Gadgetoid:
### Summary

⚠️  This is a critical fix for a regression introduced in https://github.com/micropython/micropython/commit/91cff8e4f10ea665c5f3f4b16d62c98d6ca22037 which rendered some (at least XIAO RP2350) boards unbootable.

I've tested [v1.25.0 (2025-04-15) .uf2](https://micropython.org/resources/firmware/SEEED_XIAO_RP2350-20250415-v1.25.0.uf2) on a XIAO RP2350 and, indeed, it's doesn't boot.

Replaces: https://github.com/micropython/micropython/pull/17374

Breakage reported: https://github.com/micropython/micropython/issues/17375

Assuming a 133MHz capable flash caused rp2_flash_set_timing_internal to set out of range dividers for some boards (anything with value of 4 and flash that doesn't tolerate higher speeds.)

This affected the XIAO RP2350 board, making it non-bootable.

Since Pico SDK's PICO_FLASH_SPI_CLKDIV is entirely unreliable on a system with a variable system clock (users can change it at runtime) then use it only to work out a default MICROPY_HW_FLASH_MAX_FREQ.

This value can be overridden in board config.

Note that RP2350's default clock is 150MHz, RP2040's is 125MHz and it has been certified at 200MHz so it's quite possible that PICO_FLASH_SPI_CLKDIV is unreliable even at standard RP2 clocks.

(If flash timings are marginal then this can manifest as instability rather than outright failure.)

### Testing

I tested boards at 250MHz to make sure we hit the worst case.

| Board              | Result                |
|---------------|----------------|
| RPI_PICO2      | 883/884 pass  |
| RPI_PICO         | 903/904 pass  |
| RPI_PICO2_W | 891/892 pass  |
| RPI_PICO_W    | 872/871 pass   |
| SEED_XIAO_RP2350 | 883/884 pass |

Note the failing test was `ports/rp2/rp2_dma.py` which, entirely predictably, failed with some variation of:

```
test_printing (__main__.Test) ... ok
test_pack_unpack_ctrl (__main__.Test) ... ok
test_register_access (__main__.Test) ... ok
test_close (__main__.Test) ... ok
test_simple_memory_copy (__main__.Test) ... ok
test_time_taken_for_large_memory_copy (__main__.Test) ... FAIL
test_config_trigger (__main__.Test) ... ok
test_irq (__main__.Test) ...irq fired
 ok

======================================================================
FAIL: test_time_taken_for_large_memory_copy <class 'Test'>
----------------------------------------------------------------------
  File "unittest/__init__.py", line 399, in run_one
  File "<stdin>", line 114, in test_time_taken_for_large_memory_copy
  File "unittest/__init__.py", line 187, in assertIn
AssertionError: Expected 25 to be in range(30, 80)

----------------------------------------------------------------------
Ran 8 tests

FAILED (failures=1, errors=0)
```

(Because the chip was clocked too fast)

I'm pretty sure we can crank the Pico, Pico W, Pico 2 and Pico 2 W chips faster, based on nothing but the lack of any loud and consistent shouting about them not working. (Though there might be marginal cases.)

The test suite probably doesn't do much to exercise the flash, save for `ports/rp2/rp2_thread_reset_part1.py` and `ports/rp2/rp2_thread_reset_part2.py`...

---

@Lesords I don't have a XIAO RP2350. Can you confirm this change works for you?

Since you seemed to suggest that your flash chips work (reliably and repeatably?) at 60MHz, would you like me to include 

```c
#define MICROPY_HW_FLASH_MAX_FREQ               (60000000)
```

in `SEEED_XIAO_RP2350/mpconfigboard.h`?

Note: I should probably avail myself of some non-Pimoroni RP2 boards to test. This was an unintended consequence of this change, but we definitely could have caught it with more contentious testing and less tunnel vision.

We don't really have much wiggle room here. `PICO_FLASH_SPI_CLKDIV` can't be relied upon where the end-user has control of the system clock. We can, however, at least use it to derive a default maximum clock value for the flash. Therefore - by my estimation - it's not appropriate to simply revert the breaking change.

### DIFF
--- a/ports/rp2/rp2_flash.c
+++ b/ports/rp2/rp2_flash.c
@@ -49,6 +49,14 @@
 static_assert(MICROPY_HW_ROMFS_BYTES % 4096 == 0, "ROMFS size must be a multiple of 4K");
 static_assert(MICROPY_HW_FLASH_STORAGE_BYTES % 4096 == 0, "Flash storage size must be a multiple of 4K");
 
+#ifndef MICROPY_HW_FLASH_MAX_FREQ
+// Emulate Pico SDK's SYS_CLK_HZ / PICO_FLASH_SPI_CLKDIV behaviour by default
+// On RP2040 if PICO_USE_FASTEST_SUPPORTED_CLOCK is set then SYS_CLK_HZ can be
+// 200MHz, potentially putting timings derived from PICO_FLASH_SPI_CLKDIV
+// out of range.
+#define MICROPY_HW_FLASH_MAX_FREQ (SYS_CLK_HZ / PICO_FLASH_SPI_CLKDIV)
+#endif
+
 #ifndef MICROPY_HW_FLASH_STORAGE_BASE
 #define MICROPY_HW_FLASH_STORAGE_BASE (PICO_FLASH_SIZE_BYTES - MICROPY_HW_FLASH_STORAGE_BYTES)
 #endif
@@ -104,9 +112,8 @@ static bool use_multicore_lockout(void) {
 // and core1 locked out if relevant.
 static void __no_inline_not_in_flash_func(rp2_flash_set_timing_internal)(int clock_hz) {
 
-    // Use the minimum divisor assuming a 133MHz flash.
-    const int max_flash_freq = 133000000;
-    int divisor = (clock_hz + max_flash_freq - 1) / max_flash_freq;
+    // Use the minimum divisor based upon our target MICROPY_HW_FLASH_MAX_FREQ
+    int divisor = (clock_hz + MICROPY_HW_FLASH_MAX_FREQ - 1) / MICROPY_HW_FLASH_MAX_FREQ;
 
     #if PICO_RP2350
     // Make sure flash is deselected - QMI doesn't appear to have a busy flag(!)


Lesords:
Thanks for what you did, I can help test it tomorrow (it's already 9pm here)

Lesords:
This is our flash datasheet: [310091605_Puya_P25Q16SH-UXH_IR_Datasheet.pdf](https://github.com/user-attachments/files/20516664/310091605_Puya_P25Q16SH-UXH_IR_Datasheet.pdf)

Gadgetoid:
> Maybe it would make sense to add more parameters to control all this timing?

Definitely a question for the MicroPython maintainers. For PSRAM I don't think we'll have a choice (we could have chip-specific config includes to handle this) but for flash it might be a struggle to eke out performance at the cost of configuration busywork 😆 that said, config includes for flash chips could work there too.

This feels more like a Pico SDK problem 😆 

> set the max freq to 100MHz it should work OK

Can confirm that 100MHz works for me with that change. If we can dial in the timings then we can potentially add:

* `flash_generic.h`
* `flash_P25QXXSH` ?
* `flash_whatever-it-is-we-use`

And so on, assuming performing tuning flash is something MicroPython on the whole is prepared to care about!

That's something for a future PR, though, since this one needs to be minimal enough to serve as a hotfix for the broken XIAO firmware. (Assuming the flash timing change isn't just reverted...  I think it can be?)

nspsck:
I find this to be hard to maintain, and may confuse ppl when they were trying to port boards to micropython. I would suggest to take [this approach](https://github.com/micropython/micropython/pull/17374#issuecomment-2925797582). As `PICO_FLASH_SPI_CLKDIV` will always be defined no matter what and you always get the fastest possible flash speed.
@Gadgetoid @Lesords 
You gotta decide who to change their PR.

Gadgetoid:
> As `PICO_FLASH_SPI_CLKDIV` will always be defined

This PR already relies on this, and gives setting a specific target max frequency as an *optional* config value.

Nonetheless `PICO_FLASH_SPI_CLKDIV` is wrong more often than it’s right, since it is chosen depending on the “default” RP2 clock speed (presumably only really for lazy second stage bootloader config) and there are currently at least three official values for that.

MichaelBell:
Possibly I shouldn't have confused things on this PR by talking about the wider range of flash timing settings available on RP2350 - we can raise another issue/PR for taking advantage of that.

For now, the important thing is to fix the regression on that XIAO RP2350 board.  Basing the frequency on the existing `PICO_FLASH_SPI_CLKDIV` but providing the opportunity to override to a more exact value in the board config as this PR does seems like the best approach to me.

nspsck:
> This PR already relies on this, and gives setting a specific target max frequency as an optional config value.

Yes, I get that part.

> Nonetheless PICO_FLASH_SPI_CLKDIV is wrong more often than it’s right, since it is chosen depending on the “default” RP2 clock speed (presumably only really for lazy second stage bootloader config) and there are currently at least three official values for that.

No, well, not if the one who is porting a board is doing the right job. As I mentioned in that post, the vendor only have to get [this part](https://github.com/micropython/micropython/blob/f5d10c322ecacc6be208388fa93adba5bee6eb0c/ports/rp2/boards/SEEED_XIAO_RP2350/mpconfigboard.cmake#L2) in `mpconfigboard.cmake` correct:

```
set(PICO_BOARD "seeed_xiao_rp2350")
```


In addition to that, your approach will actually limit the boards flash speed to a fixed value and leave no room for scaling. Since `PICO_FLASH_SPI_CLKDIV ` is more than `2` for the most boards, let's assume it's value is `2` and do some math (that is, if I actually can do math....... which is not always the case.......):
The default pico sysclk:
```bash
SYS_CLK_MHZ = 125
PICO_FLASH_SPI_CLKDIV = 2
# hence
MICROPY_HW_FLASH_MAX_FREQ = 62.5 MHz
```
Which means, if the clock is later increased to `250MHz`, we have
```bash
clk_sys = 250 MHz
MICROPY_HW_FLASH_MAX_FREQ = 62.5 MHz
# hence 
divider = 4.984f = 4
```
This limits the flash clock to `62.5 MHz`

If you took my approach:
```bash
SYS_CLK_MHZ = 125
PICO_FLASH_SPI_CLKDIV = 2
# hence, in MHz
divisor = (125 + 133 - 1) / 133 + 2 - 2 = 1.93f = 1
```
The flash speed is `125 MHz`, if the clock is later increased to `250MHz`, we have
```bash
SYS_CLK_MHZ = 250
PICO_FLASH_SPI_CLKDIV = 2
# hence, in MHz
divisor = (250+ 133 - 1) / 133 + 2 - 2 = 2.87f = 2
```
Flash clock is `125 MHz`.

I actually dig a bit further and got almost all max flash speed for all `pico-sdk` supported venders. They pretty much always can do 120 MHz. Means, my methode will even work at roughly `clk_sys = (120 / (133 - 120)) * 133= 1.22 GHz`. Which is a clock, let's be honest here, no rp2040 nor rp2350 is going to achieve under normal usage.

## And lemme mention the most significant benefit again (your approach does this part already, but I just want to point it out again, that mine wouldn't change this nature):
Vendors only has to set `mpconfigboard.cmake` correctly and expect their port to work. Which should be obvious for them at the first place, since they definitively have worked with the raspberry pi foundation in order to get their boards included to the `pico-sdk`.

Gadgetoid:
> No, well, not if the one who is porting a board is doing the right job

It doesn't matter what they're doing, `PICO_FLASH_SPI_CLKDIV` was only intended to provide a reasonable flash clock for Pico SDK's boot_stage2 and does not get used by default in RP2350 builds.

This is cause for surprise enough that Raspberry Pi are talking about fixes for it - https://github.com/raspberrypi/pico-sdk/issues/1903

Asking the original board vendor for the number they were attempting to hit when they picked "2" (or "3" or "4") (assuming they didn't just copy and paste the pico.h and/or guess) is not an unreasonable burden.

(Notably `PICO_FLASH_SPI_CLKDIV` only used in MicroPython's RP2350 builds because @MichaelBell mentioned it: https://github.com/micropython/micropython/pull/15619#issuecomment-2330272860)

> In addition to that, your approach will actually limit the boards flash speed to a fixed value and leave no room for scaling

This is intentional. `PICO_FLASH_SPI_CLKDIV` is configured by the vendor to indicate a divisor against the default Pico clock (or some other unknown arbitrary clock they had in mind at the time), thus calculating a fixed maximum frequency using `CLOCK_HZ / PICO_FLASH_SPI_CLKDIV` is the safest and most reasonable interpretation of their intent.

We are explicitly *avoiding* scaling because `PICO_FLASH_SPI_CLKDIV`, ie: Pico SDK itself, is fundamentally broken and doesn't make any reliable guarantees about flash config, maximum supported timings or otherwise.

> Since `PICO_FLASH_SPI_CLKDIV ` is more than `2` for the most boards

Where are you getting this information? There are 112 boards with `PICO_FLASH_SPI_CLKDIV` explicitly configured (and one - Metrotech Xerxes RP2040 - with some indirection) and 101 of these are configured with a div of 2.

> I actually dig a bit further and got almost all max flash speed for all pico-sdk supported venders.

This could be useful information, got a spreadsheet or otherwise?

I can grep for `PICO_BOOT_STAGE2_CHOOSE_` and make some assumptions based on that, but I've very little confidence that these values are reliable. The subject of this very PR is configured as `PICO_BOOT_STAGE2_CHOOSE_W25Q080` for example.

As a vendor, you have entirely too much faith in vendors 😆

nspsck:
> It doesn't matter what they're doing, PICO_FLASH_SPI_CLKDIV was only intended to provide a reasonable flash clock for Pico SDK's boot_stage2 and does not get used by default in RP2350 builds.

Ugh... not entirely sure about this part... I think this actually will be used as soon as you compile anything using the `pico-sdk.` But I use it as a indicator anyway. My thought is pretty straight forward. If a vendor set `PICO_FLASH_SPI_CLKDIV` to a value greater than `2`, then the flash is on a slower side, so we have to lower the flash frequecy by increasion the `divisor`.

> Asking the original board vendor for the number they were attempting to hit when they picked "2" (or "3" or "4") (assuming they didn't just copy and paste the pico.h and/or guess) is not an unreasonable burden.

Probably the boards were having some issues at higher cpu clocks so they had to increase the divider by a bit. I.e. vendor probably tested for `250MHz` and saw they spi-flash failed (max 120 MHz) with `PICO_FLASH_SPI_CLKDIV=2`, hence they increased the divider to 3, so when others are overclocking the pico (which is not uncommon) they do not have to manually change the divider.

> (Notably PICO_FLASH_SPI_CLKDIV only used in MicroPython's RP2350 builds because @MichaelBell mentioned it: https://github.com/micropython/micropython/pull/15619#issuecomment-2330272860)

Ugh.... to my understanding, again, as soon as you compile the firmware using `pico-sdk`, `PICO_FLASH_SPI_CLKDIV` will be defined anyway, no matter what. That's actually why you have to set `PICO_FLASH_SPI_CLKDIV=4` or so for overclocking with micropython.

> This is intentional. PICO_FLASH_SPI_CLKDIV is configured by the vendor to indicate a divisor against the default Pico clock (or some other unknown arbitrary clock they had in mind at the time), thus calculating a fixed maximum frequency using CLOCK_HZ / PICO_FLASH_SPI_CLKDIV is the safest and most reasonable interpretation of their intent.

Then we could have just removed the commit that brought the dynamic flash clocking feature, set the flash to a fixed frequency, but a little pessimistic i think.

> We are explicitly avoiding scaling because PICO_FLASH_SPI_CLKDIV, ie: Pico SDK itself, is fundamentally broken and doesn't make any reliable guarantees about flash config, maximum supported timings or otherwise.

We are using a fixed version of pico-sdk, so won't be a huge problem, but you are right.

> Where are you getting this information? There are 112 boards with PICO_FLASH_SPI_CLKDIV explicitly configured (and one - Metrotech Xerxes RP2040 - with some indirection) and 101 of these are configured with a div of 2.

My bad, should be "more or equal than 2". And I went through the `.h` files [here](https://github.com/raspberrypi/pico-sdk/tree/master/src/boards/include/boards). It's exactly what you have described.

> This could be useful information, got a spreadsheet or otherwise?

I.e. [P25Q32SH-UXH-IR](https://www.alldatasheet.com/datasheet-pdf/pdf/1656473/PUYA/P25Q32SH-UXH-IR.html) which is used by Seeed Studio and Waveshare for some of their boards. Basically I just checked boards with a default divider higher than 2, checked they schematics, see the flash model, searched these.

> As a vendor, you have entirely too much faith in vendors 😆

I trust you! And others!

Gadgetoid:
On RP2040 it’s *always* used, as it’s used to configure the flash during boot stage2.

On RP2350 you have to explicitly ask for Pico SDK to embed the XIP setup. Otherwise it’s just a dangling define that you *could* use, but it’s a poor substitute for configuring the flash - including all the other timing settings - properly.

And yes the reasonable alternative to this PR is just to revert the breaking change. We have reasons to not want to do that but they’re definitely a bit niche.

Probably wouldn’t hurt to find out what Raspberry Pi have planned for flash config because it’s currently - at the risk of causing a diplomatic incident - a bit of a half baked mess. 🤣

dpgeorge:
> Probably wouldn’t hurt to find out what Raspberry Pi have planned for flash config because it’s currently - at the risk of causing a diplomatic incident - a bit of a half baked mess. 🤣

I wouldn't go so far to say that... but, yes, I agree that the situation is not great with the way flash speed is configured in pico-sdk.  I guess it's a case of designing something simple to fit a need (configurable flash speed on RP2040 only) and then that design doesn't scale well in the future (RP2350, max frequency being updated beyond 125MHz).

I think the original intent of `PICO_FLASH_SPI_CLKDIV` was a simple way to specify the maximum SPI flash frequency, within the constraints of the RP2040's hardware.  But it doesn't generalise:
- it's based on a fixed clock frequency of 125MHz
- it's based on an integral divider
- even worse, the multiplier must be even

We don't have much room to move here.  We must assume that a vendor has chosen `PICO_FLASH_SPI_CLKDIV` so that `SYS_CLK_HZ / PICO_FLASH_SPI_CLKDIV` is the maximum frequency of the SPI flash.  I know that's most likely wrong, or at least far from exact (as @nspsck argues above) but we don't really have any choice.  Unless we want to go through all boards and check their schematics/BoM for the maximum SPI flash frequency -- and further have a guarantee that all boards produced actually match that BoM -- we have to go with this suboptimal max frequency calculation.

Really pico-sdk need to solve this issue and require boards to specify at least the flash frequency, if not other timing parameters.  But until that happens I think the fix in this PR is the right way forward.  It allows boards to get better performance if they tune the value of `MICROPY_HW_FLASH_MAX_FREQ`.

dpgeorge (approved):

Gadgetoid:
> Really pico-sdk need to solve this issue and require boards to specify at least the flash frequency, if not other timing parameters.

I had planned to make a case for this once I understood the problem and its implications within MicroPython. It's a chicken and egg problem, and having a good case - like this - to point at is as useful starting point to advocate for change.

dpgeorge:
Rebased and merged in b725c26b4e351b71b8f72ad8e55bbbd5ccce4602

dpgeorge:
Unfortunately this broke SIL_RP2040_SHIM:
```
$ make BOARD=SIL_RP2040_SHIM
...
[ 48%] Building C object CMakeFiles/firmware.dir/rp2_flash.c.o
micropython/ports/rp2/rp2_flash.c: In function 'rp2_flash_set_timing_internal':
micropython/ports/rp2/rp2_flash.c:57:49: error: 'PICO_FLASH_SPI_CLKDIV' undeclared (first use in this function)
   57 | #define MICROPY_HW_FLASH_MAX_FREQ (SYS_CLK_HZ / PICO_FLASH_SPI_CLKDIV)
      |                                                 ^~~~~~~~~~~~~~~~~~~~~
micropython/ports/rp2/rp2_flash.c:116:31: note: in expansion of macro 'MICROPY_HW_FLASH_MAX_FREQ'
  116 |     int divisor = (clock_hz + MICROPY_HW_FLASH_MAX_FREQ - 1) / MICROPY_HW_FLASH_MAX_FREQ;
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~
micropython/ports/rp2/rp2_flash.c:57:49: note: each undeclared identifier is reported only once for each function it appears in
   57 | #define MICROPY_HW_FLASH_MAX_FREQ (SYS_CLK_HZ / PICO_FLASH_SPI_CLKDIV)
      |                                                 ^~~~~~~~~~~~~~~~~~~~~
micropython/ports/rp2/rp2_flash.c:116:31: note: in expansion of macro 'MICROPY_HW_FLASH_MAX_FREQ'
  116 |     int divisor = (clock_hz + MICROPY_HW_FLASH_MAX_FREQ - 1) / MICROPY_HW_FLASH_MAX_FREQ;
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~~
```
This boards does `set(PICO_BOARD none)`, which is kinda strange, but I guess we need to support that by providing a default for `PICO_FLASH_SPI_CLKDIV`.

@Gadgetoid are you able to make a PR to fix that?  If not then I can.

Gadgetoid:
🤦 testing tunnel vision strikes again.

Might not get a chance until later this week, but will look into that board either way!

Edit: I would have thought Pico SDK would set a default `PICO_FLASH_SPI_CLKDIV`. However this doesn't happen outside of `boot2_XXXX.s` It doesn't make a lot of sense to do this *again* so perhaps the more correct approach would be to default MICROPY_HW_FLASH_MAX_FREQ to `SYS_CLK_HZ / 4`

`none.h` (TIL that's a valid board) would default to `PICO_BOOT_STAGE2_CHOOSE_GENERIC_03H` which has a default `PICO_FLASH_SPI_CLKDIV` of 4 on both RP2040 and RP2350.

I know this is kinda potato potäto but it feels slightly more correct.

Gadgetoid:
Opened as https://github.com/micropython/micropython/pull/17510

PR #17392, opened 2025-05-30
parsenum: Reduce code size.

jepler:
### Summary

By avoiding two different checks of the string length, code size is reduced without changing behavior: Some invalid float/complex strings like "ix" will get handled just like "xx" in the main number literal parsing code instead.

The optimizer alone couldn't remove the reundant comparisons because it couldn't make a transformation that let an invalid string like "ix" pass into the generic number parsing code.

### Testing

Locally, I checked that the unix coverage tests still passed. I looked at the `size` of `parsenum.o` on the 64-bit coverage build as well as the `MICROBIT` board in the qemu port.

### DIFF
--- a/py/parsenum.c
+++ b/py/parsenum.c
@@ -252,24 +252,18 @@ parse_start:
     const char *str_val_start = str;
 
     // determine what the string is
-    if (str + 2 < top && (str[0] | 0x20) == 'i') {
-        // string starts with 'i', should be 'inf' or 'infinity' (case insensitive)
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
This looks good to me, and CI passes so it's very likely to be correct.

dpgeorge:
Rebased and merged in 5eb97552591c6d2681b3e452b4b7d8445354b138 and c1629dc2ff12a92d600b5988827566f8d693cad1

PR #17393, opened 2025-05-30
Wake on touch only if supported

meirarmon:
esp32: esp32.wake_on_touch should only be available on boards
that have SOC_TOUCH_SENSOR_SUPPORTED =y
docs/library/esp32.rst: update docs to reflect that wake_on_touch
is only available for boards that have touch sensor support

### Summary

Having code wrapped by definitions of whether or not the capability is supported on the
board instead of what board it is, is better practice.
I introduced this so that when using the method wake_on_touch on boards that don't
support it, it will be clear that it is not supported, instead of silently failing.

### Testing

I tested this on the ESP32C3

PR #17394, opened 2025-05-30
Wake on ext0 only if supported

meirarmon:
ports/esp32: esp32.wake_on_ext0 should only be available on boards
          that have SOC_PM_SUPPORT_EXT0_WAKEUP=y
docs/library/esp32.rst: update docs to reflect that wake_on_ext0
          is only available for boards that have support for ext0

### Summary

Having code wrapped by definitions of whether or not the capability is supported on the
board instead of what board it is, is better practice.
I introduced this so that when using the method wake_on_ext0 on boards that don't
support it, it will be clear that it is not supported, instead of silently failing.

### Testing

I tested this on the ESP32C3

Merge this after https://github.com/micropython/micropython/pull/17393

### DIFF
--- a/docs/library/esp32.rst
+++ b/docs/library/esp32.rst
@@ -33,6 +33,8 @@ Functions
     or a valid Pin object.  *level* should be ``esp32.WAKEUP_ALL_LOW`` or
     ``esp32.WAKEUP_ANY_HIGH``.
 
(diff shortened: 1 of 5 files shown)


dpgeorge:
Closing in favour of #17397.

PR #17400, opened 2025-05-31
tools/ci.sh: Allow errors in code-size build to fail the CI.

dpgeorge:
### Summary

PR #16666 showed that it's possible for CI to pass even if the `bare-arm` port fails to build.

This fixes that.

### Testing

Tested locally with both bash and zsh.

There are other ways to propagate errors out of a pipeline (eg `$PIPESTATUS`), but this seems the simplest.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -93,23 +93,30 @@ function ci_code_size_build {
     function code_size_build_step {
         COMMIT=$1
         OUTFILE=$2
-        IGNORE_ERRORS=$3
 
         echo "Building ${COMMIT}..."
         git checkout --detach $COMMIT
         git submodule update --init $SUBMODULES
         git show -s
         tools/metrics.py clean $PORTS_TO_CHECK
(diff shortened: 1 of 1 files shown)


ports/bare-arm/main.c
```diff
@@ -67,6 +67,7 @@ void bare_main(void) {
 void nlr_jump_fail(void *val) {
     for (;;) {
     }
+    fail
```
dpgeorge, line 70: This is just to test that the CI actually fails now.

projectgus (approved):

dpgeorge:
@projectgus I realised it's possible to simplify this a bit further because the `IGNORE_ERRORS` bit can now be removed.  Does that look OK?

projectgus (approved):
This still looks good to me.

I don't understand how this ever worked as expected without pipefail (even the older version before 787c424cf has the same pattern of piping output to tee).

dpgeorge:
> I don't understand how this ever worked as expected without pipefai

Agree.  I don't think this ever worked.

PR #17407, opened 2025-06-01
modre: Report if regex is too complex.

### LINKED ISSUE #17150: docs: re,compile fails due ops > 127 but the error is misleading
### Documentation URL

_No response_

### Description

Apparently this is expected to fail:

```python
import re

# this works
re.compile('^(?:a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v)$')

# this fails
re.compile('^(?:a|b|c|d|e|f|g|h|i|j|k|l|m|n|o|p|q|r|s|t|u|v|w)$')
```

but at least the error could hint that's because there are too many operations to handle, instead of:

```
ValueError: error in regex
```

'cause there is no error, strictly speaking, in that *regex*, it's a hard (default) limitation on *MicroPython* side.

Thanks.

### Code of Conduct

Yes, I agree

jepler:
### Summary

If the error reporting mode is at least "normal", report a failure due to a complex regex with a different message.

Closes: #17150

### Testing

I ran the unix coverage tests locally, as well as testing that the OP's complex regular expression showed the new message.

The `MICROPY_ERROR_REPORTING` level to enable this message is up for debate.

### DIFF
--- a/extmod/modre.c
+++ b/extmod/modre.c
@@ -427,6 +427,9 @@ static mp_obj_t mod_re_compile(size_t n_args, const mp_obj_t *args) {
     const char *re_str = mp_obj_str_get_str(args[0]);
     int size = re1_5_sizecode(re_str);
     if (size == -1) {
+        #if MICROPY_ERROR_REPORTING >= MICROPY_ERROR_REPORTING_NORMAL
+        mp_raise_ValueError(MP_ERROR_TEXT("regex too complex"));
+        #endif
         goto error;
     }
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
This looks OK to me, definitely much better for the user.

dpgeorge:
Rebased and merged in 745bec9ce3cc4c95d8e80fcc8f64ffecde1d91bc

Issue #7408, opened 2021-06-17
uarray extend doesn't accept an iterable


hlovatt:
The `extend` method of a `uarray` doesn't accept an iterable only a buffer. Documentations says it should be an iterable and the constructor accepts an iterable and therefore would expect `extend` to accept an iterable.

EG:

```python
MicroPython v1.14-68-gd4b45898f on 2021-02-18; PYBv1.1 with STM32F405RG
Type "help()" for more information.
>>> from uarray import array
>>> a=array('b',[1,2,3])
>>> a.extend([4,5,6])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object with buffer protocol required
```

dpgeorge:
It looks like this can be added with minimal additional code by reusing the constructor code for extension.

dpgeorge:
Fixed by 0a98f3a91130d127c937d6865ead24b6693182eb

PR #17408, opened 2025-06-01
Allow array.extend to take iterables

jepler:
### Summary

As outlined in #7408, CPython can extend an array with an iterable, while micropython doesn't. Fix this, and test it.

### Testing

I added tests for the refactored code, and ran the tests locally.

Extending an array with a list (or other known length sequence) could do better, by resizing the array once first.

### DIFF
--- a/py/objarray.c
+++ b/py/objarray.c
@@ -113,6 +113,19 @@ static mp_obj_array_t *array_new(char typecode, size_t n) {
 #endif
 
 #if MICROPY_PY_BUILTINS_BYTEARRAY || MICROPY_PY_ARRAY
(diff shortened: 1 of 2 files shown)


dpgeorge (approved):
Thanks, this looks reasonable.  Gets better compatibility with CPython.

I don't think there's anyway to make the code size smaller.

dpgeorge:
Rebased, squashed and merged in 0a98f3a91130d127c937d6865ead24b6693182eb

dpgeorge:
I see that this PR number is exactly 10000 greater than the issue it's fixing :smile:

PR #17409, opened 2025-06-01
Enable gcc undefined behavior sanitizer during coverage build

jepler:
### Summary

gcc's "undefined behavior" sanitizer can catch a range of misbehaviors at runtime that normally go unnoticed. These include integer and pointer operations that are "undefined" per the relevant C specification.

This PR fixes current undefined behavior detected under gcc 12.2.0 (debian stable/bookworm) on an x64 system, then enables it during unix coverage builds.

### Testing
I built and ran the unix tests locally, iterating until there were no remaining diagnostics.

I double checked the places I used the `mem*0` alternative functions for "passing the sniff test" (i.e., is it reasonably expected that zero-lengths would occur here, etc)

I believe that the core changes (memcpy0 etc) are implemented so that they avoid code growth except in the coverage build. (but the CI will tell us for sure)

Not all gcc sanitizers can be enabled simultaneously. So, a choice has to be mad (mainly between -fsanitize=undefined and -fsanitize=memory). I chose the undefined checker, but implemented it so that an override of the makefile variable is possible.

-fsanitize=memory is also nearly error free except for https://github.com/micropython/berkeley-db-1.xx/pull/1

I read that a future C specifiction will make e.g., `memset(NULL, 0, 0)` (setting zero bytes of a NULL pointer) not-undefined-behavior. When this comes to pass, the `mem*0` macros can possibly be removed.

Different gcc versions might be too different in what they affect, as any diagnostics will make `make test` fail due to the unexpected output.

Well, except pattern-based tests which might inadvertently skip over output text that comes from the sanitizer, giving false negatives.

### DIFF
--- a/ports/unix/variants/coverage/mpconfigvariant.h
+++ b/ports/unix/variants/coverage/mpconfigvariant.h
@@ -51,3 +51,6 @@
 #define MICROPY_HW_MCU_NAME            MICROPY_PY_SYS_PLATFORM
 // Keep the standard banner message
 #define MICROPY_BANNER_MACHINE MICROPY_PY_SYS_PLATFORM " [" MICROPY_PLATFORM_COMPILER "] version"
+
+// Add extra checks to satisfy gcc -fsanitize-undefined
+#define MICROPY_NONNULL_COMPLIANT (1)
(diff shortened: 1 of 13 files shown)


jepler:
Having done this I don't know how it contrasts with https://github.com/micropython/micropython/pull/15303 which I didn't take the time to properly review at the time and had forgotten about until I went looking just now. Duplicated effort? nah, never.

PR #17423, opened 2025-06-03
esp32/main: Fix mp_deinit order.

Carglglz:
Move `mp_deinit` to be called before `gc_sweep_all` to prevent `mp_deinit` from double freeing memory if `MICROPY_PORT_DEINIT_FUNC` is defined and the function tries to free memory already freed by `gc_sweep_all`

### Summary

Prevent a core panic on soft-reset caused by "double freeing" memory.

### Testing

Tested with esp32 and https://github.com/lvgl/lv_binding_micropython/pull/383

Probably this PR https://github.com/micropython/micropython/pull/16063#

### DIFF
--- a/ports/esp32/main.c
+++ b/ports/esp32/main.c
@@ -187,6 +187,10 @@ soft_reset_exit:
     mp_usbd_deinit();
     #endif
 
+    #ifdef MICROPY_PORT_DEINIT_FUNC
+    mp_deinit();
+    #endif
+
     gc_sweep_all();
 
     // Free any native code pointers that point to iRAM.
@@ -203,7 +207,10 @@ soft_reset_exit:
     socket_events_deinit();
     #endif
 
+    #ifndef MICROPY_PORT_DEINIT_FUNC
     mp_deinit();
+    #endif
+
     fflush(stdout);
     goto soft_reset;
 }


dpgeorge:
I'm not sure this is the right thing to do.  Once `mp_deinit()` is called, no more MicroPython API functions should be called (eg because the GIL is released).

Can you provide a minimal example of the problem you are trying to solve?

Carglglz:
> Can you provide a minimal example of the problem you are trying to solve?

Deinit a `USER_C_MODULE`, in this case `lvgl` on soft-reset, @andrewleech explains this with much more detail in https://github.com/micropython/micropython/pull/16063#issue-2606860351

So I've found `MICROPY_PORT_DEINIT_FUNC` could be used as a simple alternative for this, however
I  probably need to investigate it further since stm32 port seems to work fine without this change, but the esp32 port raises a core panic and backtrace points to `mp_deinit`. 

What happens is that lvgl uses micropython garbage collector to allocate/free memory, so at soft-reset `gc_sweep_all()` frees everything and then `lvgl.deinit()` at `mp_deinit` will try to free already freed memory which in esp32 case raises a core panic.

Note that moving `gc_sweep_all` and `esp_native_code_free_all` after `mp_deinit` works too.

> Once mp_deinit() is called, no more MicroPython API functions should be called

This includes the garbage collector I guess?

dpgeorge:
> This includes the garbage collector I guess?

Correct.  It can't run because the GIL is released.

> What happens is that lvgl uses micropython garbage collector to allocate/free memory, so at soft-reset `gc_sweep_all()` frees everything and then `lvgl.deinit()` at `mp_deinit` will try to free already freed memory which in esp32 case raises a core panic.

It sounds like lvgl should be changed so that it doesn't free memory in its deinit function.  It should rely on `gc_sweep_all()` to do that for it.

lvgl can also use finalisers, which will always be run as part of `gc_sweep_all()` if there are any objects remaining at soft reset time.

Carglglz:
> It sounds like lvgl should be changed so that it doesn't free memory in its deinit function. It should rely on gc_sweep_all() to do that for it.

Ah that was it, I've only needed to reset lvgl root pointers, `gc_sweep_all()` handles the rest, @dpgeorge thanks for the support 🙏🏼

PR #17424, opened 2025-06-03
github/workflows: Split QEMU/Arm builds into separate entries.

agatti:
### Summary

This PR takes the QEMU/Arm CI build and test step and splits it into three separate steps (bigendian, sabrelite, thumb), to allow them to run in parallel.

Currently the QEMU/Arm CI build step would take up to 16 minutes, often being the last step blocking a full test run.  With this commit, when the steps run in parallel the time it takes to complete the QEMU/Arm build and test procedure is cut in half - taking between 8 to 9 minutes depending on the CI runner load.

### Testing

The various build types have been run locally in a CI-like environment, and the GitHub workflow was run successfully on my local branch.

QEMU/Arm runs are usually scheduled together to run in the same batch, saving up to 8 minutes on each full CI run.  However, in the worst case when the three build steps are run sequentially the overall time it takes to run the whole QEMU/Arm tests might be a couple dozen seconds longer than having a single monolithic step; this is due to `mpy-cross` being built three times and submodules being fetched three times as well.

This also increases the number of CI jobs by two entries - no idea if there's a hard cap on that.

### DIFF
--- a/tools/ci.sh
+++ b/tools/ci.sh
@@ -324,13 +324,24 @@ function ci_qemu_setup_rv32 {
     qemu-system-riscv32 --version
 }
 
-function ci_qemu_build_arm {
+function ci_qemu_build_arm_prepare {
     make ${MAKEOPTS} -C mpy-cross
     make ${MAKEOPTS} -C ports/qemu submodules
+}
+
+function ci_qemu_build_arm_bigendian {
+    ci_qemu_build_arm_prepare
     make ${MAKEOPTS} -C ports/qemu CFLAGS_EXTRA=-DMP_ENDIANNESS_BIG=1
-    make ${MAKEOPTS} -C ports/qemu clean
-    make ${MAKEOPTS} -C ports/qemu test_full
+}
+
+function ci_qemu_build_arm_sabrelite {
+    ci_qemu_build_arm_prepare
     make ${MAKEOPTS} -C ports/qemu BOARD=SABRELITE test_full
+}
+
+function ci_qemu_build_arm_thumb {
+    ci_qemu_build_arm_prepare
+    make ${MAKEOPTS} -C ports/qemu test_full
 
     # Test building and running native .mpy with armv7m architecture.
     ci_native_mpy_modules_build armv7m
--- a/.github/workflows/ports_qemu.yml
+++ b/.github/workflows/ports_qemu.yml
@@ -20,13 +20,20 @@ concurrency:
 
 jobs:
   build_and_test_arm:
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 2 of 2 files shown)


agatti:
For example, for the CI test run triggered by this commit:

| Configuration | Start timestamp | End timestamp |
|--------|--------|--------|
| Bigendian | 2025-06-03T21:51:53.5684634Z | 2025-06-03T21:53:29.2741578Z |
| Sabrelite | 2025-06-03T21:51:03.3429400Z | 2025-06-03T22:00:53.2601634Z |
| Thumb | 2025-06-03T21:52:03.5608825Z | 2025-06-03T21:59:20.8603767Z |

So, taking the earliest start time (2025-06-03T21:51:03.3429400Z) and the latest end time (2025-06-03T22:00:53.2601634Z) we see the QEMU/Arm step took 9 minutes and 50 seconds, compared to 15/16 minutes it usually takes.

tools/ci.sh
```diff
@@ -317,19 +317,54 @@ function ci_qemu_setup_rv32 {
...
+
+function ci_qemu_build_arm_thumb {
+    make ${MAKEOPTS} -C ports/qemu test_full
 
     # Test building and running native .mpy with armv7m architecture.
     ci_native_mpy_modules_build armv7m
     make ${MAKEOPTS} -C ports/qemu test_natmod
 }
 
+function ci_qemu_build_arm {
```
dpgeorge: IMO, it's simpler (and definitely less code) to just have three functions `ci_qemu_build_arm_bigendian`, `ci_qemu_build_arm_sabrelite` and `ci_qemu_build_arm_thumb` and make those functions call `ci_qemu_build_arm_prepare` at the start of them.

Then the workflow can just call `ci_qemu_build_arm_${{ matrix.ci_func }}` directly (that's how stm32 workflow does it).

There's no need to support the old way of calling `ci_qemu_build_arm`.

agatti: > There's no need to support the old way of calling `ci_qemu_build_arm`.

I know at least one person whose tooling would break due to this :)

But that's fine - I'll make the necessary changes.

dpgeorge: Haha, I see!

MicroPython has a very big user surface area.  I think it's just adding to our work if we add `tools/ci.sh` to that list of things we can't change.

dpgeorge (approved):

dpgeorge:
Taking a quick look at timings I see:
- container set up ~ 1 minute
- fetch submodules, mpy-cross and port build ~20 seconds
- test suite run ~7-9 minutes

So, yes, it's a good idea to split this up so the tests can run in parallel.

PR #17391, opened 2025-05-30
extmod/modnetwork: Consolidate definition of common drivers.

andrewleech:
### Summary

Most extmod network drivers were being defined on a per-port basis, duplicating code and making enabling a driver on a new port harder.

This consolidates extmod driver declarations and removes the existing per-port definitions of them.

### Testing

Quick sanity test done on `PYBD_SF6`, `RPI_PICO2_W` & `W5100S_EVB_PICO`
``` bash
anl@STEP:~/micropython/tests$ ./run-tests.py -t a0 ./net_inet/test_tls_sites.py
platform=pyboard arch=armv7emdp inlineasm=thumb
pass  ./net_inet/test_tls_sites.py
1 tests performed (4 individual testcases)
1 tests passed

anl@STEP:~/micropython/tests$ ./run-tests.py -t a1 ./net_inet/test_tls_sites.py
platform=rp2 arch=armv6m inlineasm=thumb
pass  ./net_inet/test_tls_sites.py
1 tests performed (4 individual testcases)
1 tests passed

anl@STEP:~/micropython/tests$ ./run-tests.py -t a2 ./net_inet/test_tls_sites.py
platform=rp2 arch=armv7emsp inlineasm=thumb
pass  ./net_inet/test_tls_sites.py
1 tests performed (4 individual testcases)
1 tests passed
```

PR #17433, opened 2025-06-05
tools/boardgen: Ensure locals_dict is generated with consistent order.

andrewleech:
### Summary

`tools/boardgen.py` is used by the `make-pins.py` scripts in many ports to generate the pin definitions for the machine module.

In https://github.com/micropython/micropython/pull/17391#discussion_r2127729964 I found that this is currently generating the C structs for pin definitions with inconsistent ordering which makes it sometimes impossible to get a consistent built binary file even for no change in source files.

### Testing

Reviewed the generated files from running to see the content is the same as previously, however now with different / repeatable order.
``` bash
cd ports/stm32; 
python boards/make-pins.py --board-csv boards/PYBD_SF6/pins.csv --af-csv boards/stm32f767_af.csv --prefix boards/stm32f4xx_prefix.c          │
│   --output-source build-PYBD_SF6/pins_PYBD_SF6.c --output-header build-PYBD_SF6/genhdr/pins.h --output-af-const build-PYBD_SF6/genhdr/pins_af_const.h --output-af-defs build-PYBD_SF6/genhdr/pins_af_defs.h
```

### DIFF
--- a/tools/boardgen.py
+++ b/tools/boardgen.py
@@ -108,6 +108,10 @@ class Pin:
             )
         )
 
+    # Iterate over board pin names in consistent sorted order.
+    def board_pin_names(self):
+        return sorted(self._board_pin_names, key=lambda x: x[0])
+
     # Override this to handle an af specified in af.csv.
     def add_af(self, af_idx, af_name, af):
         raise NotImplementedError
@@ -295,7 +299,7 @@ class PinGenerator:
             file=out_source,
         )
         for pin in self.available_pins():
-            for board_pin_name, board_hidden in pin._board_pin_names:
+            for board_pin_name, board_hidden in pin.board_pin_names():
                 if board_hidden:
                     # Don't include hidden pins in Pins.board.
                     continue
@@ -389,7 +393,7 @@ class PinGenerator:
 
             # #define pin_BOARDNAME (pin_CPUNAME)
             if board:
-                for board_pin_name, _board_hidden in pin._board_pin_names:
+                for board_pin_name, _board_hidden in pin.board_pin_names():
                     # Note: Hidden board pins are still available to C via the macro.
                     # Note: The RHS isn't wrapped in (), which is necessary to make the
                     # STATIC_AF_ macro work on STM32.


tools/boardgen.py
```diff
@@ -186,7 +186,7 @@ def add_cpu_pin(self, cpu_pin_name, available=True):
 
     # Iterate just the available pins (i.e. ones in pins.csv).
     def available_pins(self, exclude_hidden=False):
-        for pin in self._pins:
+        for pin in sorted(self._pins, key=lambda p: p.name()):
```
dpgeorge: Does this order by CPU name or board name?  I think it's the CPU name.  And there can be multiple aliases for the same CPU name.  And I think the issue is that those aliases need to be sorted in order as well.

andrewleech: It's cpu name as I thought that was likely most consistent / easy to grab from the pin object. I didn't think of aliases though, I'll switch to board name then.

dpgeorge: You can see in https://github.com/micropython/micropython/pull/17391#discussion_r2127729964 that both before and after are already sorted by CPU name!  So it doesn't do anything to resort by that.

andrewleech: Yes looking closer at the code I see the overall Pin ordering comes directly as read from file into a list. It's the board names that are stored in an unordered set that needs explicit sorting.

dpgeorge: Maybe don't sort, but retain the order they are read in?

andrewleech:
Updated to just order the board name iterating (over a set)

Also created some unit tests for this module, I don't believe there was any automated testing on it previously?

dpgeorge:
> Also created some unit tests for this module, I don't believe there was any automated testing on it previously?

Correct, there's no existing test for it.  Except of course that boards build without error.

I can see it's AI generated, and not actually run as part of the CI (yet).  It's quite a large test and would take time to review.  Maybe put that in a separate PR, so this one (sorting pin names) is easier and quicker to get merged?

andrewleech:
> I can see it's AI generated, and not actually run as part of the CI (yet). It's quite a large test and would take time to review. Maybe put that in a separate PR, so this one (sorting pin names) is easier and quicker to get merged?

Yes it was generated, I used it to check the results of the change here and gave a quick sanity check over the other tests thinking they looked better than nothing. I didn't think a build / no build of ports is quite enough coverage as doens't show if things are missed, nor the ordering as found today!

It does warrant separate discussion though so good idea, I've split it out into its own PR here https://github.com/micropython/micropython/pull/17434

dpgeorge (approved):
Tested on PYBD_SF2.  The board pins are now in a sorted order (sorting is first done on CPU name, then boards name).

PR #17435, opened 2025-06-05
ports/alif/boards/ALIF_ENSEMBLE: Add deploy instructions.

iabdalkader:
### Summary

Add deploy/build instructions for Ensemble E7 DevKit.

- [ ] Board images: https://github.com/micropython/micropython-media/pull/95

### Testing

I've tested these instructions, with another board, and they work fine. Testing with the devkit itself would be helpful.

### DIFF
--- a/ports/alif/README.md
+++ b/ports/alif/README.md
@@ -19,3 +19,61 @@ The following more advanced features will follow later:
 - Ethernet support.
 - SDRAM support.
 - Other machine modules.
+
+Build instructions
+------------------
+
+Before building the firmware for a given board the MicroPython cross-compiler
+must be built; it will be used to pre-compile some of the built-in scripts to
+bytecode.  The cross-compiler is built and run on the host machine, using:
+```bash
+$ make -C mpy-cross
+```
+
+This command should be executed from the root directory of this repository.
+All other commands below should be executed from the ports/alif/ directory.
+
+An ARM compiler is required for the build, along with the associated binary
+utilities.  The recommended toolchain version to use with this port is
+Arm GNU toolchain version 13.3.Rel1. The compiler can be changed using the
+`CROSS_COMPILE` variable when invoking `make`.
+
+Next, the board to build must be selected. The default board is `ALIF_ENSEMBLE`
+but any of the names of the subdirectories in the `boards/` directory is valid.
+The board name must be passed as the argument to `BOARD=` when invoking `make`.
+
+All boards require certain submodules to be obtained before they can be built.
+The correct set of submodules can be initialised using (with `ALIF_ENSEMBLE`
+as an example of the selected board):
+```bash
+make BOARD=ALIF_ENSEMBLE submodules
+```
+
+Then to build the board's firmware run:
+```bash
+make BOARD=ALIF_ENSEMBLE
+```
+
+The above command should produce binary images in the `build-ALIF_ENSEMBLE/`
+subdirectory (or the equivalent directory for the board specified).
+
+### Update the SE Firmware
+
+The SE firmware must be updated **before** flashing the main firmware to match
+the version used by MicroPython. This step only needs to be performed once.
+Connect the board to your PC via the **SE UART USB** port (on the Ensemble kit,
+this is labeled **PRG USB**), then run:
+
+```bash
+make update-system-package
+```
+
+**Note:** The board must be power-cycled after this step.
+
+### Deploy MicroPython
+
+To flash the firmware, run:
+
+```bash
+make BOARD=ALIF_ENSEMBLE deploy
+```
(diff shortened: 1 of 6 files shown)


iabdalkader:
@dpgeorge I'm not sure how to add the firmware to the hosted files, we probably need to add the entire build dir. Could you please add it? Also, I don't have a devkit to test with, but I did test these instructions with AE3.

> ./ports/alif/boards/ALIF_ENSEMBLE/board.json unknown features: {'OSPI Flash', 'I2S/PDM Microphone', 'SD Card', 'OSPI HyperRAM', 'Joystick'}

I'm not really sure what the purpose of the "features" tag is, but I can just remove these.

ports/alif/boards/ALIF_ENSEMBLE/deploy.md
```diff
@@ -0,0 +1,63 @@
...
+
+```bash
+python alif-security-toolkit/toolkit/app-write-mram.py --port /dev/ttyACM0 --cfg-part AE722F80F55D5XX --cfg-rev B4 --pad --images file:build-ALIF_ENSEMBLE/application_package.ds
+```
+
+---
+
+### Build MicroPython from Source
+
+Alternatively, you can build the firmware from source. The MicroPython repository includes all necessary tools.
```
dpgeorge: I don't think it's a good idea to put these build instructions here.  They should instead be in the `ports/alif/README.md` file, similar to other ports.

dpgeorge:
> I'm not sure how to add the firmware to the hosted files, we probably need to add the entire build dir.

I think we should add just the files that are needed for deployment over serial.  To do that, we should create a target in the `Makefile` like `package` which packages up all the needed files into a .zip (or .tgz?).  And then `tools/autobuild/build-boards.sh` can have a function added to it which builds and collects this .zip.  That will be very similar to how the cc3200 port works.

ports/alif/boards/ALIF_ENSEMBLE/deploy.md
iabdalkader: Updated.

iabdalkader:
> I think we should add just the files that are needed for deployment over serial. To do that, we should create a target in the `Makefile` like `package` which packages up all the needed files into a .zip (or .tgz?). And then `tools/autobuild/build-boards.sh` can have a function added to it which builds and collects this .zip. That will be very similar to how the cc3200 port works.

This is fine with me, but it's probably best if you do this step. We could merge the build instructions and board images contained in this PR or feel free to push to the branch/cherry-pick the commits here.

dpgeorge:
I've now pushed 2 commits to this PR:
- create `firmware.zip` as part of the alif build process, containing relevant files needed to deploy over SEUART
- add alif to the auto-build scripts, so the boards are built along with all the other boards

I didn't test the zip yet!

ports/alif/boards/ALIF_ENSEMBLE/deploy.md
```diff
@@ -0,0 +1,32 @@
...
+
+---
+
+### Update the SE Firmware (Optional)
+
+If needed, update the SE firmware to match the version used by MicroPython. Ensure you have the correct port, part number, and chip revision.
+
+```bash
+cd alif-security-toolkit/toolkit
+python updateSystemPackage.py --port /dev/ttyACM0 --cfg-part AE722F80F55D5XX --cfg-rev B4
```
kwagyeman, line 17:
```suggestion
python updateSystemPackage.py --port /dev/ttyACM0
```

The other arguments aren't needed to update to B4.

ports/alif/README.md
```diff
@@ -19,3 +19,40 @@ The following more advanced features will follow later:
...
+as an example of the selected board):
+```bash
+make BOARD=ALIF_ENSEMBLE submodules
+```
+
+Then to build the board's firmware run:
+```bash
+make BOARD=ALIF_ENSEMBLE
+```
+
```
kwagyeman, line 56: For brand new boards you need to update the system package:

```
make update-system-package
```

This only needs to be done once. Then, afterwards you can deploy the image, which you must do each time via:

```
make deploy
```

ports/alif/boards/ALIF_ENSEMBLE/deploy.md
```diff
@@ -0,0 +1,32 @@
+### Alif Security Toolkit
```
kwagyeman, line 1: This section isn't really needed as the makefile has support for both of these steps.

kwagyeman:
Tested the build and everything works.

ports/alif/boards/ALIF_ENSEMBLE/deploy.md
iabdalkader: Yes it is if you download the firmware and not clone the repo.

kwagyeman: Yes, good point.

iabdalkader:
@dpgeorge Is the readme/deploy here clear enough? Should we move forward with this PR?

dpgeorge (approved):
I tested this on an Alif dev kit by following the instructions here, and the system package update works, along with the firmware install.

PR #17441, opened 2025-06-06
extmod/mbedtls: Implement DTLS HelloVerify, update test for MSG_PEEK

projectgus:
### Summary

* Updates the DTLS support initially added in #15764.
* DTLS 1.2 requires the HelloVerify protocol. This was already enabled on esp32 port (part of ESP-IDF config), and this PR enables it on bare metal ports as well as anti-replay attack protection.
* HelloVerify requires a cookie store to be implemented in the server application that calls mbedTLS. We're using the cookie implementation from mbedTLS (implemented in C), the only Python-facing addition is a new `client_id` argument for `wrap_socket()`. MicroPython DTLS server code has to pass a transport-specific client identifier (as bytes) here, in order for HelloVerify to work correctly. The same SSLContext should be reused for all incoming connections.
* Add a config flag to disable DTLS on bare metal ports that don't need it. Enabled on "Extra" and above by default, so I think all ports with mbedTLS will get DTLS by default.
* Updated the multi_net test to work with HelloVerify and the new `MSG_PEEK` receive flag added in #17312. No longer needs an additional byte to be sent at the start of each connection.

### Testing

* Ran multi_net tests, including updated dtls test, on rp2, stm32 (PYBD_SF2) and esp32 ports. I'm still having (seemingly unrelated) issues with high outgoing UDP packet loss on rp2, but stm32 and esp32 ports passed the test in both directions.
* Manually tested the mbedTLS in-tree `dtls_server` and `dtls_client` programs against a MicroPython client and server (adapted from the multi_net test, with the mbedTLS test certificates swapped in). Tests worked, suggesting MicroPython should now work with any DTLS 1.2 peer.

* It's not clear how high the risk of MicroPython DTLS servers amplifying a DDoS attack is, so perhaps the HelloVerify support is not needed on the server side. However, connecting to a (compliant) DTLS 1.2 server requires the DTLS client to implement HelloVerify, and mbedTLS (reasonably) treats HelloVerify as enabled for both server and client concurrently so we need to implement something on the server side or remove DTLS server support (see below for some discussion of this). The "best case" code size impact of only supporting DTLS client is still about half of what's added by this PR.

### DIFF
--- a/extmod/modtls_mbedtls.c
+++ b/extmod/modtls_mbedtls.c
@@ -62,6 +62,9 @@
 #include "mbedtls/ecdsa.h"
 #include "mbedtls/asn1.h"
 #endif
+#ifdef MBEDTLS_SSL_DTLS_HELLO_VERIFY
+#include "mbedtls/ssl_cookie.h"
+#endif
 
 #ifndef MICROPY_MBEDTLS_CONFIG_BARE_METAL
 #define MICROPY_MBEDTLS_CONFIG_BARE_METAL (0)
@@ -92,6 +95,9 @@ typedef struct _mp_obj_ssl_context_t {
     #if MICROPY_PY_SSL_ECDSA_SIGN_ALT
     mp_obj_t ecdsa_sign_callback;
     #endif
+    #ifdef MBEDTLS_SSL_DTLS_HELLO_VERIFY
+    mbedtls_ssl_cookie_ctx cookie_ctx;
+    #endif
 } mp_obj_ssl_context_t;
 
 // This corresponds to an SSLSocket object.
@@ -117,7 +123,8 @@ static const mp_obj_type_t ssl_socket_type;
 static const MP_DEFINE_STR_OBJ(mbedtls_version_obj, MBEDTLS_VERSION_STRING_FULL);
 
 static mp_obj_t ssl_socket_make_new(mp_obj_ssl_context_t *ssl_context, mp_obj_t sock,
-    bool server_side, bool do_handshake_on_connect, mp_obj_t server_hostname);
+    bool server_side, bool do_handshake_on_connect, mp_obj_t server_hostname,
+    mp_obj_t client_id);
 
 /******************************************************************************/
 // Helper functions.
@@ -320,6 +327,16 @@ static mp_obj_t ssl_context_make_new(const mp_obj_type_t *type_in, size_t n_args
     mbedtls_ssl_conf_dbg(&self->conf, mbedtls_debug, NULL);
     #endif
 
+    #ifdef MBEDTLS_SSL_DTLS_HELLO_VERIFY
+    mbedtls_ssl_cookie_init(&self->cookie_ctx);
+    ret = mbedtls_ssl_cookie_setup(&self->cookie_ctx, mbedtls_ctr_drbg_random, &self->ctr_drbg);
+    if (ret != 0) {
+        mbedtls_raise_error(ret);
+    }
+    mbedtls_ssl_conf_dtls_cookies(&self->conf, mbedtls_ssl_cookie_write, mbedtls_ssl_cookie_check,
+        &self->cookie_ctx);
+    #endif
+
     return MP_OBJ_FROM_PTR(self);
 }
 
@@ -366,6 +383,11 @@ static mp_obj_t ssl_context___del__(mp_obj_t self_in) {
     mbedtls_ctr_drbg_free(&self->ctr_drbg);
     mbedtls_entropy_free(&self->entropy);
     mbedtls_ssl_config_free(&self->conf);
+    #ifdef MBEDTLS_SSL_DTLS_HELLO_VERIFY
+    if (self->is_dtls_server) {
+        mbedtls_ssl_cookie_free(&self->cookie_ctx);
+    }
+    #endif
     return mp_const_none;
 }
 static MP_DEFINE_CONST_FUN_OBJ_1(ssl_context___del___obj, ssl_context___del__);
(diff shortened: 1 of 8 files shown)


projectgus:
While reviewing this myself I realised the server DTLS cookie support I implemented is far too naive. Have marked as draft until I have time to improve it.

projectgus:
> * The current session cookie algorithm is too naive and doesn't achieve the purpose of HelloVerify (and can be used to DoS the server by creating too many cookies). It should be updated along the lines of the suggestion in the DTLS specification (`Cookie = HMAC(Secret, Client-IP, Client-Parameters)`).

This is a lot harder than I initially thought, because in MicroPython `modmbedtls` works with an abstract stream and doesn't have any knowledge of the client IP. Also, adding this will probably add even more code size. But what's currently in this draft is woefully inadequate.

There's two possible ways forward:

1) Add another Python API extension, so the server code does something like `s = ctx.wrap_socket(s, server_side=1, client_cookie=hmac_result)` or `s = ctx.dtls_wrap_socket(...)`. This offloads the responsibility (and code size cost) for calculating a secure cookie to Python code (and we could provide a sample implementation or a thin wrapper in micropython-lib that does it correctly). The server code can choose to pass a constant string here, at the cost of potentially becoming a DDoS vector if exposed to untrusted clients.
2) Drop DTLS server support, support DTLS client only. This makes testing a little harder as we'll need a DTLS server endpoint, but a simple no-op server should be easy enough to stand up (and safe to stand up on the public internet if it implements Hello Verify!)

Leaning towards (2) at the moment because I think the use case for DTLS client in MicroPython is much stronger than the case for needing a DTLS server, but that would mean removing functionality that was in v1.25 (albeit not fully compliant in that version). @keenanjohnson you implemented the original DTLS support in #15764, do you have any thought about this?

extmod/modtls_mbedtls.c
```diff
@@ -669,6 +685,51 @@ static mp_obj_t ssl_socket_make_new(mp_obj_ssl_context_t *ssl_context, mp_obj_t
     mbedtls_raise_error(ret);
 }
 
+#if defined(MBEDTLS_SSL_DTLS_HELLO_VERIFY) && defined(MBEDTLS_SSL_SRV_C)
+// DTLS Server session cookie support
+static int ssl_cookie_write(void *ctx, unsigned char **p, unsigned char *end,
```
dpgeorge: Would it be simpler and more general to require that these functions be implemented in Python, and the user passes them through when wrapping the socket?  Eg:
```py
cookie_dict = {}
def dtls_cookie_handler(ctx, key, value):
    if value is None:
        # create and return new cookie
        cookie = os.urandom(32)
        cookie_dict[key] = cookie
        return cookie
    else:
        # check cookie
        return cookie_dict.get(key) == value
ctx.wrap_socket(..., dtls_cookie=dtls_cookie_handler)
```

keenanjohnson:
In regards to your question @projectgus , I personally don't see a lot of use for the DTLS server on an embedded device. However, I do think that having the server implemented in micropython does make the unit testing vastly less brittle. Every time I have tied unit tests to some sort of live public service (like you suggest in option 2), I have personally regretted the choice in the long run as it tends to be brittle, people start to encounter all sorts of spurious local test failiures, etc etc. 

Given that our main usecase for the server is just unit tests, I feel like moving the cookie calculation to python isn't the worst as long as we caveat the potential for a DDOS in the docs or something in case anyone ever comes along and has a real use case for the server.

Thanks for thinking of me and great work on the project as always.

projectgus:
Thanks @dpgeorge and @keenanjohnson for your thoughts:

### Alternative 1

I've pushed a version that uses the mbedTLS ssl_cookie implementation in C. All that the Python code needs to do is supply a client transport identifier (i.e. bytes representing IP & port) when it wraps the socket.

Code size cost is less bad than I thought: `rp2: +1240`. ESP32 impact will be a little lower as some of this was compiled in already. But still a lot of code size!

### Alternative 2

Alternative to this, along the lines suggested by @dpgeorge above, would be to implement the cookie support in Python - but it probably still need to be a two step process in order to pass through the client's transport id:

```py
cookie_dict = {}
def dtls_cookie_handler(ctx, key, value):
    encoded_cookie = binary encoding of (timestamp, HMAC(secret, key + timestamp)
    if value is None:
        return encoded_cookie
    else:
        # compare encoded_cookie matches `value` and also check the prepended timestamp isn't too old

ctx.set_dtls_cookie_handler(dtls_cookier_handler)

while True:
        _, client_addr = s.recvfrom(1, socket.MSG_PEEK)
        print("incoming connection")
        s.connect(client_addr)  # Connect back to the client

        # Wrap the UDP socket in server mode.
        try:
            s = ctx.wrap_socket(s, server_side=1, client_id=repr(client_addr).encode())
```

### Alternative 3

There's a slightly hackier version where we pass it all through to wrap_socket() in one go, i.e.:

```py
  client_id = repr(client_addr).encode()
  s = ctx.wrap_socket(s, server_side=1,
            cookie_handler= lambda ctx, key, value: dtls_client_handler(ctx, key, value, client_id))
```

... but the necessary implementation is hacky as would need to pass the mp_obj_t pointer via the `client_id` bytes and then decode it on the other side. If `cookie_handler` needs to stay valid after wrap_socket returns (unclear) then it'd need to also be stored somewhere in the ssl socket object so GC doesn't pick it up.

projectgus:
After discussing with @dpgeorge we're going to stick with alternative 1. The code size impact is not that much higher than the other alternatives (maybe 400 out of the +1200 bytes on rp2) and this way we get a correct secure implementation (via mbedTLS).

projectgus:
> Attention: Patch coverage is 95.00000% with 1 line in your changes missing coverage. Please review.

The line missing coverage here will only be hit if a heap allocation fails part-way through setting up the SSL context, not an easy code path to cover with unit testing.

dpgeorge:
Tested the DTLS multi_net test on:
- micropython unix <-> micropython unix
- PYBD_SF6 <-> micropython unix
- PYBD_SF6 <-> ESP32_GENERIC
- PYBD_SF6 <-> RPI_PICO2_W
- ESP32_GENERIC <-> micropython unix
- ESP32_GENERIC <-> RPI_PICO2_W
- RPI_PICO2_W <-> micropython unix

Also tested against openssl (which has a mode where it can be a DTLS client or server):
- micropython unix <-> openssl s_server/s_client
- PYBD_SF6 <-> openssl s_server/s_client
- ESP32_GENERIC <-> openssl s_server/s_client
- RPI_PICO2_W <-> openssl s_server/s_client

Everything works!

dpgeorge (approved):

PR #6024, opened 2020-05-10
py/parsenum: Implement exact float parsing using integer mpz (WIP)

dpgeorge:
This PR attempts to implement exact parsing of floats in `mp_parse_num_decimal()`.  There have been quite a few issues related to this, and parsing floats correctly is important, so I thought it'd be worth trying to fix this properly.

Other implementations I could find (eg musl's) are pretty complex and use a lot of stack.  The implementation here is deigned and written from scratch (but it's probably not original) and reuses the existing MicroPython big-int functions to do the bulk of the conversion.

The point is that all operations during the conversion must be exact (ie not do any rounding or else the rounding will compound), except as an optimisation the very last bit can use the FP hardware because it'll round in the correct way.

The idea is to parse the input number as a big-int and then apply the exponent also using big-int arithmetic.  The trick is that the exponent of `10**e` can be factored into `(2*5)**e` of which the `2**e` part can be done exactly in hardware because that's the natural base of the FP representation.  So only the `5**e` calculation needs to be done using big-ints.  If `e` is negative then the big-int is pre-multiplied by a large enough number (namely `8**e`) so it can then be divided by `5**e` without loss.

The entire parsing could be done without any FP hardware, ie just using big-ints to construct the bits in the floating point representation (mantissa, exponent, sign bit).  But it seems OK to do the last bit using standard float operations (converting the big-int to a float then adjusting by the power-of-2 exponent) and this makes the code a bit simpler (reuses existing functions, like `ldexp`).

Tested against a set of known difficult numbers (from CPython test suite) and it parses all of them correctly (in double precision mode).

TODO/outstanding:
- do more tests
- test with 32-bit floats
- don't allocate heap memory for temporary mpz's, use the stack instead
- benchmark to see if it has acceptable performance

PR #17444, opened 2025-06-06
py/formatfloat: Improve accuracy of float formatting code.

yoctopuce:
### Summary

Following discussions in [PR #16666](https://github.com/micropython/micropython/pull/16666), this pull request updates the float formatting code to reduce the `repr` reversibility error, i.e. the percentage of valid floating point numbers that do not parse back to the same number when formatted by `repr`.

The baseline before this commit is an error rate of ~46%, when using double-precision floats.

This new code is available in two flavors, based on a preprocessor definition:
- In the simplest version, it reduces the error down to ~40%, using an integer representation of the decimal mantissa rather than working on floats. It is also slightly faster, and improves the rounding in some conditions.
- In the most complete version, it reduces the error down to ~5%. This extra code works by iterative refinement, and makes the code slightly slower than CPython when tested on ports/unix.

### Testing

The new formatting code was tested for reversibility using the code provided by Damien in [PR #16666](https://github.com/micropython/micropython/pull/16666)
A variant using formats `{:.7g}`, `{:.8g}` and `{:.9g}` was used for single-precision testing.

Compatibility with CPython on the various float formats was tested by comparing the output using the following code:
```py
for mant in [34567, 76543]:
    for exp in range(-16, 16):
        print("Next number: %de%d" % (mant, exp))
        num = mant * (10.0**exp)
        for mode in ['e', 'f', 'g']:
            maxprec = 16
            # MicroPython has a length limit in objfloat.c
            if mode == 'f' and 6 + exp + maxprec > 31:
                maxprec = 31 - 6 - exp
            for prec in range(1, maxprec):
                fmt = "%." + str(prec) + mode
                print("%5s: " % fmt, fmt % num)
```
The integration tests have also found some corner cases in the new code which have been fixed.
For single-precision floats, some test cases had to be adapted:
- `float_format_ints` is tapping into an ill-defined partial digit of the mantissa (the 10th), which is not available in single-precision floats with the new code due to integer limitations. So the display range has been updated accordingly.
- similarly, `float_struct_e` uses a 15-digit representation which is meaningless on single-precision floats. A separate version for double-precision has been made instead
- in `float_format_ints`, there is one test case specific to single-precision floats which verifies that the largest possible mantissa value `16777215` can be used to store that exact number and retrieve it as-is. Unfortunately the rounding in the simplest version of the new algorithm makes it display as a slightly different number. This would cause the CI test to fail on single-precision floats when the improved algorithm is not enabled.

It is unclear at that point if the simplest version of this improvement is worth the change: 
- going from 46% error to 40% error in double precision is not a big improvement.
- there is no improvement for single-precision
- the new code is only marginally faster

The full version of the enhancement makes much more difference in terms of precision, both for double-precision and single-precision floats, but it causes about 20% overhead on conversion time, and makes the code a bit bigger

Looking forward to reading your feedback...

Edit 1: See https://github.com/micropython/micropython/pull/17444#issuecomment-2979968660 for updates on accuracy results
Edit 2: Updated values in https://github.com/micropython/micropython/pull/17444#issuecomment-2987116217


yoctopuce:
Additional results after a few more experiments in double-precision :
- with the simple method, %17g provides the fastest `repr` value with ~40% error 
- with the elaborate method enabled, we can bring `repr` error down to 0.5% as follows:
  1) try first %17g, and stop if a perfect match is found (90% of the cases)
  2) otherwise fallback to %18g (7% of the cases)
  3) otherwise fallback to %19g (remaining cases)
  The overall cost is only 20% more than the fastest `repr` that had 40% error.

dpgeorge:
Thanks for this.  I will study it in detail.

I also found the corresponding CPython issue and long discussion about this topic: https://github.com/python/cpython/issues/45921 .  There's a lot of useful information there.  It looks like this is a hard thing to get right.

yoctopuce:
> I also found the corresponding CPython issue and long discussion about this topic: [python/cpython#45921](https://github.com/python/cpython/issues/45921) . There's a lot of useful information there. It looks like this is a hard thing to get right.

I read that thread, it is indeed an interesting input. But I am afraid that we will not be able to use much of it due to ressource constraints in MicroPython.

yoctopuce:
> * with the elaborate method enabled, we can bring `repr` error down to 0.5% as follows:

Out of curiosity, I have been investigating a bit more...

Half of these remaining errors are caused by `{:g}` falling back to `{:f}` for small negative exponents and therefore loosing significant digits in the mantissa. If we run the test using `{:e}`, the `repr`error goes down to 0.25%. So I am looking at changing the code for `{:f}` to insert leading zeroes without integrating them in the mantissa. This should properly fix the issue and improve the accuracy of both `{:f}` and `{:g}`.

About 20% of the remaining errors are caused by large negative exponents (exp < -255). There is probably something we could improve there as well, but I am not sure it is a big concern for MicroPython if such seldom-used numbers have accuracy problems in `repr`. 

I am also looking at cleaning a bit the code, to reduce code footprint. While doing this, one things that I noticed is that the function includes lots of checks on `buf_size` and `buf_remaining` at various places, but MicroPython always uses a 32 bytes buffer (or 16 bytes buffer for single-precision floats). Don't you think it would make sense to document this function as requiring a 32 bytes output buffer, adding an `assert(buf_size >= 32)` at the beginning and simplifying these checks ?

dpgeorge:
> Don't you think it would make sense to document this function as requiring a 32 bytes output buffer, adding an `assert(buf_size >= 32)` at the beginning and simplifying these checks ?

Yes, that makes a lot of sense.

dpgeorge:
Related: #6024 has been around for a while and attempt to improve the other side of this, ie parsing of floats, using higher precision ints.

yoctopuce:
Thanks, I will look at it as well. I have a big update to submit for this PR, which I believe makes things look much better.
But I just found an accuracy issue in the float parse code for very small numbers, so I need to fix it first and check the impact on results...

yoctopuce:
I have just force-pushed my new code.
In addition to previous test code and to running the CI tests, I have been testing using the following routine:
```py
import array, math, time, binascii

seed = 42


def pseudo_randouble():
    global seed
    ddef = []
    for _ in range(8):
        ddef.append(seed & 0xFF)
        seed = binascii.crc32(b'\0', seed)
    arr = array.array('d', bytes(ddef))
    return ddef, arr[0]


# The largest errors come from seldom used very small numbers, near the
# limit of the representation. So we keep them out of this test to keep
# the max relative error display useful.
if float('1e-100') == 0:
    # single-precision
    min_expo = -96  # i.e. not smaller than 1.0e-29
    # Expected results:
    # HIGH_QUALITY_REPR=1: 99.71% exact conversions, relative error < 1e-7
    # HIGH_QUALITY_REPR=0: 94.89% exact conversions, relative error < 1e-6
else:
    # double-precision
    min_expo = -845  # i.e. not smaller than 1.0e-254
    # Expected results:
    # HIGH_QUALITY_REPR=1: 99.83% exact conversions, relative error < 2.7e-16
    # HIGH_QUALITY_REPR=0: 64.01% exact conversions, relative error < 1.1e-15

ttime = 0
stats = 0
N = 10000000
max_err = 0
for _ in range(N):
    (ddef, f) = pseudo_randouble()
    while f == math.isinf(f) or math.isnan(f) or math.frexp(f)[1] <= min_expo:
        (ddef, f) = pseudo_randouble()

    start = time.time_ns()
    str_f = repr(f)
    ttime += time.time_ns() - start
    f2 = float(str_f)
    if f2 == f:
        stats += 1
    else:
        error = abs(f2 - f) / f
        if max_err < error:
            max_err = error
            print("{:.19e}: repr='{:s}' err={:.4e}".format(f, str_f, error))

print("{:d} values converted in {:d} [ms]".format(N, int(ttime / 1000000)))
print("{:.2%} exact conversions, max relative error={:.2e}".format(stats / N, max_err))
```
It is similar to the one Damien posted before, but the version tests specifically the `repr` function, which features additional refinements compared to a `%.g` format with fixed precision. This code also makes sure that no single case shows a relative error greater than the precision expected from the mantissa.

This new code brings the error rate to a really low level, which should be acceptable for MicroPython intended use.

yoctopuce:
The unix / nanbox variant did originally crash at the end of the accuracy test, due to the randomly generated float number that create invalid nanbox objects.

I have therefore fixed the nanbox code in `obj.h` to prevent the creating bad `nan`objects

py/mpprint.h
```diff
@@ -37,7 +37,9 @@
 #define PF_FLAG_CENTER_ADJUST     (0x040)
 #define PF_FLAG_ADD_PERCENT       (0x080)
 #define PF_FLAG_SHOW_OCTAL_LETTER (0x100)
-#define PF_FLAG_SEP_POS           (9) // must be above all the above PF_FLAGs
+#define PF_FLAG_USE_OPTIMAL_PREC  (0x200)
+#define PF_FLAG_ALWAYS_DECIMAL    (0x400)
+#define PF_FLAG_SEP_POS           (16) // must be above all the above PF_FLAGs
```
jepler: Over in the PR where the code that added PF_FLAG_SEP_POS we discussed the position of this flag. There is a port with 16-bit `int`s. It was possible to position PF_FLAG_SEP_POS at 9, because there were enough bits in an unsigned 16-bit value to squeeze in `_` and `,` as shifted values without widening the type of the flags argument.

It's my mistake that I didn't comment on the requirement here, nor write a check that would trigger during CI. There is an assertion about the situation in `objstr.c` but because the pic16bit port is not built during CI (non-free toolchain) the assertion is only checked on platforms with 32-bit ints.

yoctopuce: Goot point. I saw your (unrelated) PR on `mpprint.h` yesterday and was wondering about this.
I have just updated the code to move these new flags above the `SEP` bits.
As they are strictly related to the use of floats, which are obviously not enabled in the pic16 port, this should be safe.

yoctopuce:
While fixing `nanbox` CI to avoid a crash in case of handcrafted `nan` floats, I run into another small nanbox bug: `math.nan` was incorrectly defined as a signaling `nan`. Using that value caused a failure of `float/math_domain.py` when testing `copysign`. So I have included that fix as well in the PR, since this was breaking CI tests.

yoctopuce:
Heads up: I found a trick to get 100% good conversions, significantly faster then CPython, and the code appears to be smaller than my previous version. Working on it...

yoctopuce:
I have finally managed to complete cleaning my new implementation. It is a much bigger rewrite than I anticipated originally...

This new code offers a choice of 3 float conversion methods, depending on the desired tradeoff between code size and conversion precision:
- BASIC method is the smallest code footprint: it achieves better results than current method through the use of `mp_float_uint_t` to work on the mantissa, and also by trying multiple precisions to find the best `repr`.
- APPROX method iteratively improves the mantissa, which is a bit slower but but does not have a big impact on code size. It provides `repr` reversibility on >99.70% of the cases.
- EXACT method uses higher-precision floats during conversion,  which provides best results, but has a higher impact on code size. It is faster than APPROX method, and faster than  CPython equivalent implementation.

Here is the table comparing the impact of the three conversion methods on code footprint on PYBV10 (using single-precision
floats) and reversibility rate for both single-precision and double-precision floats. The table includes current situation as a baseline for the comparison:

|               | PYBV10  code size  |  IMPL_FLOAT |  IMPL_DOUBLE |
|----------|------------|--------|--------|
|current implementation | 364136 |  85.47% |  37.90% |
|MICROPY_FLOAT_CONV_BASIC| 364188  | 97.78%  | 62.18% |
|MICROPY_FLOAT_CONV_APPROX| 364396 |  99.70% |  99.84% |
|MICROPY_FLOAT_CONV_EXACT| 365608  |100.00% | 100.00% |

**edit**: Newer stats here: https://github.com/micropython/micropython/pull/17444#issuecomment-2987116217

Unfortunately, on ARM gcc, the EXACT method is significantly bigger than the APPROX method, so I am not sure this is worth the code size increase for limited platforms. However on platforms that use IMPL_DOUBLE, I guess it might be worth enabling it when possible. But not all platforms have support for true `long double` floats.

The commit also include two minor fix for nanbox, that were preventing the new CI tests to run properly on that port.

I will look for any CI test failing.

robert-hh:
The slowest board in the previous test was the ESP8266. So I ran run-tests.py -t u0 float/*.py. It ran to an end, with the test summary below:

42 tests performed (2417 individual testcases)
34 tests passed
24 tests skipped: float/cmath_dunder.py float/cmath_fun.py float/cmath_fun_special.py float/complex1_intbig.py float/complex1.py float/complex_dunder.py float/complex_reverse_op.py float/complex_special_methods.py float/float2int_doubleprec_intbig.py float/float2int_intbig.py float/float_divmod.py float/float_format_ints_doubleprec.py float/float_format_ints_power10.py float/float_parse_doubleprec.py float/float_struct_e_doubleprec.py float/int_big_float.py float/math_constants_extra.py float/math_domain_special.py float/math_factorial_intbig.py float/math_fun_special.py float/math_isclose.py float/string_format.py float/true_value.py float/types.py
8 tests failed: float/builtin_float_round.py float/float1.py float/float_format_accuracy.py float/float_format_ints.py float/float_format.py float/float_parse.py float/lexer.py float/python36.py

Running the test on a SAMD21, only float/float_format_accuracy.py failed.
Renesas-RA EK-RA6M2 is slow as well, but all tests eithe pass or are skipped.

The tested ESP32, RP2 and NRF boards fail the float_format_accuracy.py test, because it takes too long.
ESP8266 and SAM21 fail this test with:

```
  File "<stdin>", line 19, in pseudo_randouble
AttributeError: 'module' object has no attribute 'crc32'
```

yoctopuce:
> The tested ESP32, RP2 and NRF boards fail the float_format_accuracy.py test, because it takes too long. 
> ESP8266 and SAM21 fail this test with: `AttributeError: 'module' object has no attribute 'crc32'`

I have just pushed a new version that runs less tests, this should now be much faster
The new version uses a Python implementation of crc32 in case `binutils.crc32` is not available.

> 8 tests failed: float/builtin_float_round.py float/float1.py float/float_format_accuracy.py float/float_format_ints.py float/float_format.py float/float_parse.py float/lexer.py float/python36.py
> Running the test on a SAMD21, only float/float_format_accuracy.py failed. 

Thank you very  much. Could you send me the output of `float/float_format_accuracy.py` on platforms where it fails ? It should be just two lines.

> Renesas-RA EK-RA6M2 is slow as well, but all tests eithe pass or are skipped.
It looks like this board has single-precision floating point enabled.

Can you confirm that `float/float_format_accuracy.py` is not in the list of skipped tests ?

Thanks again for your help.

robert-hh:
> Could you send me the output of float/float_format_accuracy.py on platforms where it fails ?

ESP8266:
```
mpremote run float/float_format_accuracy.py 
  File "<stdin>", line 48, in <module>
  File "<stdin>", line 19, in pseudo_randouble
AttributeError: 'module' object has no attribute 'crc32'
```
SAMD21:
```
mpremote run float/float_format_accuracy.py 
  File "<stdin>", line 48, in <module>
  File "<stdin>", line 19, in pseudo_randouble
AttributeError: no such attribute
```
RP2040:
```
time mpr run float/float_format_accuracy.py 
10000 values converted
float format accuracy OK

real	0m20,374s
user	0m0,141s
sys	0m0,076s
```
ESP32:
```
time mpr run float/float_format_accuracy.py 
10000 values converted
float format accuracy OK

real	0m19,573s
user	0m0,198s
sys	0m0,025s
```
NRF (Arduino Nano Connect)
```
time mpr run float/float_format_accuracy.py 
10000 values converted
float format accuracy OK

real	0m16,019s
user	0m0,164s
sys	0m0,034s
```
Renesas-RA EK-RA6M2:
```
time mpremote u0 run float/float_format_accuracy.py 
10000 values converted
float format accuracy OK

real	0m9,177s
user	0m0,127s
sys	0m0,022s
```

yoctopuce:
Thank you again. Could you try the new version that I pushed 1h ago ? It should run properly on ESP8266 and SAMD21 using a the built-in crc32 implementation in Python, and it should run faster on all platforms as I have reduced the number of conversions to 2000.

robert-hh:
The test  `float_format_accuracy.py` now passes for ESP32, NRF and RP2. The runtime is betwee 4-5 seconds. 
The test `float/float_format_accuracy.py` runs fine to an end with ESP8266 and SAMD21 as well, but the execution time is too long with 15 and 17 seconds. So running it with run-test.py will fail.

Besides that I wonder about the function pseudo_randouble(). As far as I can tell, it fills a list with pseude-random numbers with the first value always being 42, converts that list to a bytes-object and that to a float and returns it. So my question:

a) why do you not use random.getrandbits() for the number? It is available at most ports & boards. It uses a pseudo-rng. Trying to use random.getrandbits leads to spurious fails like this one: `FAILED: repr rate=99.950% max_err=2.720e-07`

b) pseudo_randouble() returns both that random number list ddef and the float value. But ddef is not used. Dropping ddef as return value improves the speed a little bit.

robert-hh:
> Trying to use random.getrandbits leads to spurious fails like this one: FAILED: repr rate=99.950% max_err=2.720e-07

Your solution for pseudo_randouble() generates always the same sequence of numbers with the same number of differences. In my tests the number sequence was always different with different start values. Seeding the RNG returns the same number sequence.

Another observation: If I print f and f2 in the test branch where the match, the majority of numbers print a inf or -inf. Is that intentional? That happens only at a single precisio board. With a double precision board all is fine. If I change line 83 to
`    while math.isinf(f) or math.isnan(f) or math.frexp(f)[1] <= min_expo:`, I get 0.0 values instead of inf. The changed expression seems to be the proper one.
```
-inf -inf
inf inf
-inf -inf
-inf -inf
-inf -inf
-inf -inf
-inf -inf
-2.9860916e-09 -2.9860916e-09
5.985439e+33 5.985439e+33
inf inf
-14969399.0 -14969399.0
4.66570656e+27 4.66570656e+27
inf inf
inf inf
1.4599386e-09 1.4599386e-09
-inf -inf
-3.0602446e-10 -3.0602446e-10
-inf -inf
-inf -inf
2.970024e-24 2.970024e-24
1.5661973e-16 1.5661973e-16
inf inf
-inf -inf
-inf -inf
inf inf
-inf -inf
-inf -inf
-inf -inf
inf inf
```

yoctopuce:
> The test `float_format_accuracy.py` now passes for ESP32, NRF and RP2. The runtime is betwee 4-5 seconds. The test `float/float_format_accuracy.py` runs fine to an end with ESP8266 and SAMD21 as well, but the execution time is too long with 15 and 17 seconds. So running it with run-test.py will fail.

OK, I will try to reduce the number of samples even more.

> a) why do you not use random.getrandbits() for the number? It is available at most ports & boards. It uses a pseudo-rng. 

I was afraid that `getrandbits()` would not be available everywhere, as it depends on `MICROPY_PY_RANDOM_EXTRA_FUNCS` which is only available with EXTRA features. By using a crc32-based pseudo generator, I was doing about the same, in a way that would be for sure reproductible accross all platforms.

> Trying to use random.getrandbits leads to spurious fails like this one: `FAILED: repr rate=99.950% max_err=2.720e-07`

If the number of samples if too small, the status may not converge soon enough. My guess is that if you let it run for 10'000 samples or more, the stats converge to
- 99.70% for IMPL_FLOAT
- 99.83% for IMPL_DOUBLE
regardless of the random generator.

> b) pseudo_randouble() returns both that random number list ddef and the float value. But ddef is not used. Dropping ddef as return value improves the speed a little bit.

Good point. I used to save ddef in order to be able to analyze number that caused problems, but this is no more needed. I will remove it.

Thank you again for all the time you spend running these test, it really helps a lot !

yoctopuce:
> Your solution for pseudo_randouble() generates always the same sequence of numbers with the same number of differences.

Correct, that was the intent. We want to test a variety of floats, but we want the test to be reproductible, to avoid random failures.

> Another observation: If I print f and f2 in the test branch where the match, the majority of numbers print a inf or -inf. 

Oho, good point. As I was developing on `ports/unix` with a fast machine, I did not realize that my random generator was always based on doubles, which means a large quantity of generated numbers are actually dropped when testing on a single-precision port. This is stupid, as it waste precious testing. I will change the generator to generate 32 bit floats on 32 bit machines, rather than generating 64 bit floats and converting them to 32 bits. This should speed up things a lot !

Thank you again, will post a new version tomorrow morning.

robert-hh:
Some more results:
- Testing proper single precision floats shows a  higher fail rate: `FAILED: repr rate=99.300% max_err=2.810e-07`
- Using getrandbits() at ESP8266 fails with a strange error after 418 attempts `TypeError: can't convert NoneType to float`.  It seems that sometimes other types that float are returned by pseudo_randsingle() (I made two functions. If I change the test for proper numbers to: `while type(f) is not float or math.isinf(f) or math.isnan(f) or math.frexp(f)[1] <= min_expo:`, then it passes on ESP8266, and the time for 2000 values is below 8 seconds using getrandbits().
- No need to support SAMD21 for the test. SAMD51 can be used for the SAMD port and is much faster (1s instead of 12s).

robert-hh:
Attached is the version of the test I use right now.
```
# Test accuracy of `repr` conversions.
# This test also increases code coverage for corner cases.

try:
    import array, math
except ImportError:
    print("SKIP")
    raise SystemExit

seed = 42

import random
random.seed(seed)

def pseudo_randouble():
    global seed
    ddef = bytearray(8)
    for _ in range(8):
        ddef[_] = random.getrandbits(8)
    arr = array.array('d', ddef)
    return arr[0]


def pseudo_randsingle():
    global seed
    ddef = bytearray(4)
    for _ in range(4):
        # ddef[_] = (seed & 0xFF)
        # seed = crc32(b'\0', seed)
        ddef[_] = random.getrandbits(8)
    arr = array.array('f', ddef)
    return arr[0]

# The largest errors come from seldom used very small numbers, near the
# limit of the representation. So we keep them out of this test to keep
# the max relative error display useful.
if float('1e-100') == 0:
    # single-precision
    min_expo = -96  # i.e. not smaller than 1.0e-29
    # Expected results:
    # - MICROPY_FLOAT_CONV_EXACT: 100% exact conversions
    # - MICROPY_FLOAT_CONV_APPROX: 99.70% exact conversions, max relative error < 1e-7
    min_success = 0.997
    max_rel_err = 1e-7
    float_conversion = 'f'
    float_size = 4
    pseudo_randfloat = pseudo_randsingle
else:
    # double-precision
    min_expo = -845  # i.e. not smaller than 1.0e-254
    # Expected results:
    # - MICROPY_FLOAT_CONV_EXACT: 100% exact conversions
    # - MICROPY_FLOAT_CONV_APPROX: >=99.83% exact conversions, max relative error <= 2.7e-16
    min_success = 0.998
    max_rel_err = 2.7e-16
    pseudo_randfloat = pseudo_randouble

stats = 0
N = 2000
max_err = 0

for _ in range(N):
    f = pseudo_randfloat()
    while type(f) is not float or math.isinf(f) or math.isnan(f) or math.frexp(f)[1] <= min_expo:
        f = pseudo_randfloat()

    str_f = repr(f)
    f2 = float(str_f)
    if f2 == f:
        stats += 1
    else:
        error = abs(f2 - f) / f
        if max_err < error:
            max_err = error

print(N, "values converted")
if stats / N >= min_success and max_err < max_rel_err:
    print("float format accuracy OK")
else:
    print("FAILED: repr rate=%.3f%% max_err=%.3e" % (100 * stats / N, max_err))
```

yoctopuce:
> Some more results:
> 
> * Testing proper single precision floats shows a  higher fail rate: `FAILED: repr rate=99.300% max_err=2.810e-07`
> * Using getrandbits() at ESP8266 fails with a strange error after 418 attempts `TypeError: can't convert NoneType to float`.  It seems that sometimes other types that float are returned by pseudo_randsingle() (I made two functions. If I change the test for proper numbers to: `while type(f) is not float or math.isinf(f) or math.isnan(f) or math.frexp(f)[1] <= min_expo:`, then it passes on ESP8266, and the time for 2000 values is below 8 seconds using getrandbits().
> * No need to support SAMD21 for the test. SAMD51 can be used for the SAMD port and is much faster (1s instead of 12s).

Thank you for these additional tests.

I have just pushed a new version of the test code, that fixes , with a fixed version of the floating point generator and should make it faster by avoiding garbage-collecting the list.

I realized that the previous version that was generating doubles, then casting them to single-precision floats was significantly biasing the results for single precision floats, by running a lots of tests on `inf` (due to the wrong test) and on '0.0'. This also changes the stats for the baseline using existing code, which is much worth than initially thought. So although the accuracy obtained by the APPROX method is now significantly lower than initially thought, it also shows that the new code is a big improvement over the existing one.

Here are the updated stats (based on 1 Mio random numbers):

|               | PYBV10  code size  |  IMPL_FLOAT |  IMPL_DOUBLE |
|----------|------------|--------|--------|
|current implementation | 364136 |  27.57% |  37.90% |
|MICROPY_FLOAT_CONV_BASIC| 364188  | 91.01%  | 62.18% |
|MICROPY_FLOAT_CONV_APPROX| 364396 |  98.50% |  99.84% |
|MICROPY_FLOAT_CONV_EXACT| 365608  |100.00% | 100.00% |

Would you mind giving a try to this new test code ?

robert-hh:
Runtimes with the code above:

STM32 (pyboard V1.1) 0.76s
ESP32 2.62s
MIMXRT (Teensy 4.1) 0.35s
CC3200 (WiPy) 10.2s
ESP8266 7.5s
SAMD51 1.04 s
SAMD21 12.7s
Renesas-RA (EK-RA6M2) 4.04s
RP2 (RP2040) 4.39s
NRF (Arduino Nano Connect NRF82540) 1.9s

So only CC3200 and SAMD21 would be above the 10s limit. That's OK. We might not have to take care about them. All single float tests fail by exceeding the error limits at `FAILED: repr rate=98.350% max_err=8.675e-08`.

robert-hh:
I will test that during the day. Just a quick compare for the two ends of the range of your recent test accuracy code with the code I used:

MIMXRT 1.4s (your code) vs 0.37s
STM32 3.06s vs 0.76s
ESP8266 39s vs 7.5 s

Note that even with your code the ESP8266 needs the changed test, that the value returned by pseudo_randouble() is a float.

tests/float/float_format_accuracy.py
```diff
@@ -0,0 +1,115 @@
...
+    # print(rnd_buff, "%.17e" % arr[0])
+    return arr[0]
+
+
+stats = 0
+N = 10000
+max_err = 0
+for _ in range(N):
+    f = pseudo_randouble()
+    while math.isinf(f) or math.isnan(f) or math.frexp(f)[1] <= min_expo:
```
robert-hh: ESP8266 needs this line to be changed to:
`    while type(f) is not float or math.isinf(f) or math.isnan(f) or math.frexp(f)[1] <= min_expo:`

yoctopuce: Sorry, I did not realize the problem would not be specific to `randbits`. I have now force-pushed a new version reusing many of yours ideas, including using `randbits`. I have limited the number of tests to 1500, so it should run on all platforms I guess.

Would you give it a try ?

yoctopuce:
> I will test that during the day. Just a quick compare for the two ends of the range of your recent test accuracy code with the code I used:
> 
> MIMXRT 1.4s (your code) vs 0.37s STM32 3.06s vs 0.76s ESP8266 39s vs 7.5 s
> 
> Note that even with your code the ESP8266 needs the changed test, that the value returned by pseudo_randouble() is a float.

`randbits` is definitely the best solution, agreed. I have changed the test code to use it, and added the ESP8266 check. But I will see if I can fix the `array.array` function (as I did for nanbox) rather than adding this extra guard, I believe it would make more sense.

robert-hh:
There is not much difference between rnd_buf[] being local or global. The local version is slightly faster, since it does not need a symbol lookup. So for code clarity the local version seems better. Interestingly the results for ESP8266 and SAMD21, RP2, STM32 or ESP32 at N=1200 are different. Except for SAMD21, you could go back to N=2000.

SAMD21 at N=1200:
```
1200 values converted
FAILED: repr rate=98.167% max_err=8.675e-08

real	0m7,237s
user	0m0,109s
sys	0m0,038s
```
ESP8266, N=1200:
```
1200 values converted
FAILED: repr rate=98.750% max_err=3.082e-07

real	0m4,653s
user	0m0,114s
sys	0m0,027s
```
ESP8266, N=2000:
```
2000 values converted
FAILED: repr rate=98.900% max_err=3.082e-07

real	0m7,240s
user	0m0,131s
sys	0m0,021s
```
RP2 at N=1200:
```
1200 values converted
FAILED: repr rate=98.167% max_err=8.675e-08

real	0m2,802s
user	0m0,106s
sys	0m0,024s
```
RP2 at N=2000:
```
2000 values converted
float format accuracy OK

real	0m4,436s
user	0m0,126s
sys	0m0,009s
```

robert-hh:
On a side note: At single precision the function pseudo_randfloat() is called 2301 times, at double precision it's 2171.

robert-hh:
So they all pass fine at N=2000 except for the ESP8266, where the error stays too high. I wonder what's the difference using  the same float code in MP. Obviously the compiler & C-libs are different.

yoctopuce:
> So they all pass fine at N=2000 except for the ESP8266, where the error stays too high. I wonder what's the difference using the same float code in MP. Obviously the compiler & C-libs are different.

The problem has the same root cause as the reason why you had to add ` type(f) is not float`: REPR_C. 
- This is a nan-boxing representation, and when creating an object from float, the code does not check if the input float is a `nan`different from *the* `nan`. This is the reason why the function randfloat created non-float objects
- Two digits are lost in the mantissa when storing objects in REPR_C...

I am looking at fixing the first issue as I did for REPR_D, but I am afraid that this would increase code size significantly given this is an inline function. So maybe we should leave your test in the test case, and I should submit this as a different PR to make the impact on code size more obvious.

Regarding the second issue, I guess I can detect REPR_C by the fact that some floats don't exist, and adjust the expectations accordingly.

dpgeorge:
> I guess I can detect REPR_C by the fact that some floats don't exist, and adjust the expectations accordingly.

Yes, that's possible.  `tests/feature_check/float.py`  uses `float("1.0000001") == float("1.0")` to see if REPR_C is active.

yoctopuce:
> > I guess I can detect REPR_C by the fact that some floats don't exist, and adjust the expectations accordingly.
> 
> Yes, that's possible. `tests/feature_check/float.py` uses `float("1.0000001") == float("1.0")` to see if REPR_C is active.

Thanks for the hint. So I have added the exact same test in `float_format_accuracy.py`

Regarding the problem of hand-crafted REPR_C floats getting interpreted as objects:
1) do you prefer that I include the fix in this PR as I did for nanbox, or that I move both to a different PR ?
2) I am a bit concerned by the impact on code size of the extra check to add in `mp_obj_new_float` inline function.
    Should I rather make it non-inline for REPR_C ? I don't think the performance hit would be that big

tests/float/float_format_accuracy.py
```diff
@@ -0,0 +1,73 @@
...
+    # - MICROPY_FLTCONV_IMPL_EXACT: 100% exact conversions
+    # - MICROPY_FLTCONV_IMPL_APPROX: >=99.83% exact conversions, max relative error <= 2.7e-16
+    min_success = 0.997  # with only 1200 samples, the success rate is lower
+    max_rel_err = 2.7e-16
+
+
+# Deterministic pseudorandom generator. Designed to be uniform
+# on mantissa values and exponents, not on the represented number
+def pseudo_randfloat():
+    global rnd_seed, rnd_buff
```
robert-hh: If at all, this line should be 
`    global float_size`

robert-hh: Besides that the test passes for all boards.

dpgeorge:
> 1. do you prefer that I include the fix in this PR as I did for nanbox, or that I move both to a different PR ?

Please move to a different PR, this one is getting big :)



> 2\. I am a bit concerned by the impact on code size of the extra check to add in `mp_obj_new_float` inline function.
>         Should I rather make it non-inline for REPR_C ? I don't think the performance hit would be that big

I need to think about this.

yoctopuce:
> So they all pass fine at N=2000 except for the ESP8266, where the error stays too high. I wonder what's the difference using the same float code in MP. Obviously the compiler & C-libs are different.

I have pushed the code with the new REPR_C tolerance. I think this should now work on all platforms. To keep the test duration short, I have left the number of numbers to test set to 1200. As demonstrated with REPR_C, 1200 steps appear to be sufficient to detect any platform-specific discrepancy.

tests/float/float_format_accuracy.py
yoctopuce: Sure :-)

yoctopuce: Thank you VERY much for all time you have spent helping me to improve this PR, this was really helpful.

I have pushed the new version without the useless `global` (and moving away `math.nan` fix on REPR_C, as it was not strictly needed for this PR). No need to rerun the tests for now, nothing else has changed.

I will be sending a separate PR for fixing `math.nan`and crafted nan-floats on REPR_C and REPR_D

yoctopuce:
> > 1. do you prefer that I include the fix in this PR as I did for nanbox, or that I move both to a different PR ?
> 
> Please move to a different PR, this one is getting big :)

@dpgeorge I have submitted a PR with the trivial fix for `math.nan` and hand-crafted floats in `nanbox`. Once it will be integrated into master, I can remove that part of the code from this PR and rebase to the new master. 

I have intentionally kept the REPR_C crafted-floats issue separate, as more time will be needed to find the most efficient implementation. The test code is already available here: https://github.com/micropython/micropython/commit/edf5cab28ab7ebd953cbb2e91b73a98be834bab5

yoctopuce:
I have rebased the PR to master and removed nanbox-related changes, now that they have been integrated in master.

py/mpprint.c
```diff
@@ -353,6 +353,14 @@ int mp_print_float(const mp_print_t *print, mp_float_t f, char fmt, unsigned int
 
     char *s = buf;
 
+    if ((flags & PF_FLAG_ALWAYS_DECIMAL) && strchr(buf, '.') == NULL && strchr(buf, 'e') == NULL && strchr(buf, 'n') == NULL) {
+        if ((size_t)(len + 2) < sizeof(buf)) {
```
dpgeorge: Is it possible to remove this buffer size check by allocating 2 more bytes to `buf`, and then passing `sizeof(buf) - 2` to `mp_format_float()` above, thereby guaranteeing there will always be enough space to add these characters?

Same comment for the `PF_FLAG_ADD_PERCENT` check below, so add 3 bytes in total to `buf` and actually pass `sizeof(buf) - 3` to `mp_format_float()`.

py/mpconfig.h
```diff
@@ -856,6 +856,23 @@ typedef double mp_float_t;
 #define MICROPY_PY_BUILTINS_COMPLEX (MICROPY_PY_BUILTINS_FLOAT)
 #endif
 
+// Float conversion to/from string implementations
+#define MICROPY_FLTCONV_IMPL_BASIC (0)  // smallest code, but inexact
+#define MICROPY_FLTCONV_IMPL_APPROX (1) // slightly bigger, almost perfect
+#define MICROPY_FLTCONV_IMPL_EXACT (2)  // bigger code, but 100% exact repr
```
dpgeorge: I would use "and 100% exact repr" (instead of "but") because it's a positive thing that it's an exact representation.

py/mpconfig.h
```diff
@@ -856,6 +856,23 @@ typedef double mp_float_t;
...
+#if MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_FLOAT
+#define MICROPY_FLOAT_DEFAULT_FLTCONV_IMPL (MICROPY_FLTCONV_IMPL_APPROX)
+#elif defined(__SIZEOF_LONG_DOUBLE__) && __SIZEOF_LONG_DOUBLE__ > __SIZEOF_DOUBLE__
+#define MICROPY_FLOAT_DEFAULT_FLTCONV_IMPL (MICROPY_FLTCONV_IMPL_EXACT)
+#else
+#define MICROPY_FLOAT_DEFAULT_FLTCONV_IMPL (MICROPY_FLTCONV_IMPL_APPROX)
+#endif
+
+#ifndef MICROPY_FLTCONV_IMPL
+#define MICROPY_FLTCONV_IMPL (MICROPY_FLOAT_DEFAULT_FLTCONV_IMPL)
```
dpgeorge: Is this `MICROPY_FLOAT_DEFAULT_FLTCONV_IMPL` needed?  Why not just wrap the above bit of code in `#ifndef MICROPY_FLTCONV_IMPL` and then define `MICROPY_FLTCONV_IMPL` directly?

Also, maybe call it `MICROPY_FLOAT_FORMAT_IMPL`?

dpgeorge:
This is looking very good now!  My testing on PYBD-SF2 (single prec) and PYBD-SF6 (double prec) shows that everything is good, all tests pass.

And around +250 bytes code size is pretty decent for the benefit provided by the approx algorithm.

Definitely `mpy-cross` should use the exact algorithm, because that benefits all targets and the cost (code size and speed) is host side.  Could you enable that?

yoctopuce:
> Definitely `mpy-cross` should use the exact algorithm, because that benefits all targets and the cost (code size and speed) is host side. Could you enable that?

The default settings in `mpconfig.h` should enable `EXACT` mode automatically by default for `mpy-cross` when using `gcc`, because `mpy-cross` is using `IMPL_DOUBLE` and `long double` is available. On Windows, there is no true `long double`, so `APPROX` is the best we can do anyway...

py/mpconfig.h
yoctopuce: OK, done

py/mpprint.c
yoctopuce: Excellent idea, I have changed that.

py/misc.h
```diff
@@ -272,6 +272,57 @@ typedef union _mp_float_union_t {
     mp_float_uint_t i;
 } mp_float_union_t;
 
+static inline int fp_signbit(mp_float_t x) {
```
dpgeorge: I'm wondering a few things about this and the below helper functions:
1. Do they need to be in this public header, or can they just all be put in formatfloat.c where they are used (and only used)?
2. If they are public, might be better to namespace them by prefixing them with `mp_`, like eg `mp_clz()` below.
3. Why have a custom signbit function, why not just use `signbit`?  Note that `py/objfloat.c` already uses `signbit`.

py/mpconfig.h
```diff
@@ -856,6 +856,21 @@ typedef double mp_float_t;
 #define MICROPY_PY_BUILTINS_COMPLEX (MICROPY_PY_BUILTINS_FLOAT)
 #endif
 
+// Float conversion to/from string implementations
```
dpgeorge: Might be worth adding a comment here that EXACT is only possible when long double exists (is that true?).

py/parsenum.c
```diff
@@ -178,27 +180,77 @@ typedef enum {
...
 #elif MICROPY_FLOAT_IMPL == MICROPY_FLOAT_IMPL_DOUBLE
-#define MANTISSA_MAX 0x1999999999999998ULL
-#define SMALL_NORMAL_VAL (1e-307)
-#define SMALL_NORMAL_EXP (-307)
-#define EXACT_POWER_OF_10 (22)
+#define MAX_EXACT_POWER_OF_5 (22)
 #endif
 
+// Helper to compute `num * (10.0 ** dec_exp)`
+mp_large_float_t mp_decimal_exp(mp_large_float_t num, int dec_exp) {
```
dpgeorge, line 223: This refactoring of `parsenum.c` -- and also it looks like it slightly improves things by doing 5**exp instead of 10**exp -- can you put it in a separate commit?  It would be good to have that change self-contained so that it's easier to understand what is going on.

yoctopuce: Good idea. This is now done.

py/mpconfig.h
yoctopuce: Let me know if the added text is OK

py/misc.h
yoctopuce: You definitely have a talent to find where code could be simplified :-)

These definitions were directly inherited _as is_ from the previous version of the format code. But that code was clearly not up-to-date with current usage of floating point tools in MicroPython: it was duplicating many helper functions which are now available from other places. I have now cleaned up most of these, and kept the last ones which might be useful directly in `py/formatfloat.c`

yoctopuce:
OK, it looks like we are all set for the next review, the last failing test is a false positive.

py/mpconfig.h
dpgeorge: Yes, that's good now.

py/misc.h
dpgeorge: > These definitions were directly inherited _as is_ from the previous version of the format code.

Ah, I see.



> I have now cleaned up most of these, and kept the last ones which might be useful directly in `py/formatfloat.c`

Very good!

dpgeorge:
One last thing: it would be good to have a build that uses the BASIC mode.  I think esp8266 is the ideal candidate for that because it has little flash available, not much computing resources, and if we have room to spare on that port it's best to use that room for other things.

Could you enable BASIC on that port, in `ports/esp8266/mpconfigport.h`?

I tested and with APPROX (current situation) that port grows by +380 bytes, with BASIC it grows by +132.

dpgeorge:
> Could you enable BASIC on that port, in `ports/esp8266/mpconfigport.h`?

Actually, maybe that's easier said than done.  Running the tests on esp8266 with BASIC enabled gives 8 failures:
```
FAILURE tests/results/float_float_format_ints.py
--- tests/results/float_float_format_ints.py.exp	2025-07-05 08:58:24.111382479 +1000
+++ tests/results/float_float_format_ints.py.out	2025-07-05 08:58:24.111478972 +1000
@@ -478,6 +478,6 @@
 23456 x 10^4 with format {:.8e} gives 2.34560000e+08
 23456 x 10^4 with format {:.8f} gives 234560000.00000000
 23456 x 10^4 with format {:.8g} gives 2.3456e+08
-16777215.000000
+16777214.400000
 2147483520.000000
-1.000000e+38
+9.999998e+37

FAILURE tests/results/float_float_parse.py
--- tests/results/float_float_parse.py.exp	2025-07-05 08:58:24.481610409 +1000
+++ tests/results/float_float_parse.py.out	2025-07-05 08:58:24.481699777 +1000
@@ -8,9 +8,9 @@
 True
 True
 True
-1e-37
-1e-37
-1e-37
+9.999998e-38
+9.999998e-38
+9.999998e-38
 inf
 0.0
 inf

FAILURE tests/results/float_python36.py
--- tests/results/float_python36.py.exp	2025-07-05 08:58:33.011527395 +1000
+++ tests/results/float_python36.py.out	2025-07-05 08:58:33.012626615 +1000
@@ -1,5 +1,5 @@
-1000.18
+1000.17992
 1e+12
 123.0
-123.4
+123.399992
 1e+13

FAILURE tests/results/float_lexer.py
--- tests/results/float_lexer.py.exp	2025-07-05 08:58:26.874129703 +1000
+++ tests/results/float_lexer.py.out	2025-07-05 08:58:26.874202664 +1000
@@ -1 +1 @@
-0.1
+0.099999992

FAILURE tests/results/float_float_format_accuracy.py
--- tests/results/float_float_format_accuracy.py.exp	2025-07-05 08:58:21.583842985 +1000
+++ tests/results/float_float_format_accuracy.py.out	2025-07-05 08:58:21.583925345 +1000
@@ -1,2 +1,2 @@
 1200 values converted
-float format accuracy OK
+FAILED: repr rate=93.167% max_err=4.488e-07

FAILURE tests/results/float_float_format.py
--- tests/results/float_float_format.py.exp	2025-07-05 08:58:18.242580224 +1000
+++ tests/results/float_float_format.py.out	2025-07-05 08:58:18.242648482 +1000
@@ -116,13 +116,13 @@
 0.00006
 0.000060
 0.0000600
-10000000000000000000
-10000000000000000000.0
-10000000000000000000.00
-10000000000000000000.000
-10000000000000000000.0000
-10000000000000000000.00000
-10000000000000000000.000000
-10000000000000000000.0000000
+9999996800000000000
+9999996800000000000.0
+9999996800000000000.00
+9999996800000000000.000
+9999996800000000000.0000
+9999996800000000000.00000
+9999996800000000000.000000
+9999996800000000000.0000000
 1.00e+12
 1.00e+19

FAILURE tests/results/float_builtin_float_round.py
--- tests/results/float_builtin_float_round.py.exp	2025-07-05 08:58:12.110944351 +1000
+++ tests/results/float_builtin_float_round.py.out	2025-07-05 08:58:12.162301735 +1000
@@ -7,7 +7,7 @@
 -123
 -124
 1.23457
-1.2
+1.1999998
 1.0
 1200.0
 -2
@@ -24,6 +24,6 @@
 0.0
 1.0
 1.5
-1.47
+1.4699998
 <class 'OverflowError'>
 <class 'ValueError'>

FAILURE tests/results/float_float1.py
--- tests/results/float_float1.py.exp	2025-07-05 08:58:14.905048967 +1000
+++ tests/results/float_float1.py.out	2025-07-05 08:58:14.905123244 +1000
@@ -1,15 +1,15 @@
-0.12
+0.119999976
 1.0
-1.2
+1.1999998
 0.0
 0.0
 0.0
-1.2
-1.2
+1.1999998
+1.1999998
 1.0
 10.0
 10.0
-0.1
+0.099999992
 inf
 -inf
 inf
@@ -21,12 +21,12 @@
 ValueError
 ValueError
 ValueError
-1.2
-3.4
+1.1999998
+3.3999996
 False
 True
-1.2
--1.2
+1.1999998
+-1.1999998
 0.5
 0.5
 0.0
```

Although, with APPROX there are also 7 failures:
```
FAILURE tests/results/float_float_format_ints.py
--- tests/results/float_float_format_ints.py.exp	2025-07-05 09:00:28.728158119 +1000
+++ tests/results/float_float_format_ints.py.out	2025-07-05 09:00:28.728308144 +1000
@@ -480,4 +480,4 @@
 23456 x 10^4 with format {:.8g} gives 2.3456e+08
 16777215.000000
 2147483520.000000
-1.000000e+38
+9.999999e+37

FAILURE tests/results/float_float_parse.py
--- tests/results/float_float_parse.py.exp	2025-07-05 09:00:29.110447738 +1000
+++ tests/results/float_float_parse.py.out	2025-07-05 09:00:29.110559388 +1000
@@ -8,9 +8,9 @@
 True
 True
 True
-1e-37
-1e-37
-1e-37
+9.999998e-38
+9.999998e-38
+9.999998e-38
 inf
 0.0
 inf

FAILURE tests/results/float_python36.py
--- tests/results/float_python36.py.exp	2025-07-05 09:00:37.893613383 +1000
+++ tests/results/float_python36.py.out	2025-07-05 09:00:37.893859481 +1000
@@ -1,5 +1,5 @@
-1000.18
+1000.17992
 1e+12
 123.0
-123.4
+123.399992
 1e+13

FAILURE tests/results/float_lexer.py
--- tests/results/float_lexer.py.exp	2025-07-05 09:00:31.453041007 +1000
+++ tests/results/float_lexer.py.out	2025-07-05 09:00:31.453109019 +1000
@@ -1 +1 @@
-0.1
+0.099999992

FAILURE tests/results/float_float_format.py
--- tests/results/float_float_format.py.exp	2025-07-05 09:00:21.947870630 +1000
+++ tests/results/float_float_format.py.out	2025-07-05 09:00:21.948010941 +1000
@@ -116,13 +116,13 @@
 0.00006
 0.000060
 0.0000600
-10000000000000000000
-10000000000000000000.0
-10000000000000000000.00
-10000000000000000000.000
-10000000000000000000.0000
-10000000000000000000.00000
-10000000000000000000.000000
-10000000000000000000.0000000
+9999996160000000000
+9999996160000000000.0
+9999996160000000000.00
+9999996160000000000.000
+9999996160000000000.0000
+9999996160000000000.00000
+9999996160000000000.000000
+9999996160000000000.0000000
 1.00e+12
 1.00e+19

FAILURE tests/results/float_builtin_float_round.py
--- tests/results/float_builtin_float_round.py.exp	2025-07-05 09:00:15.756557259 +1000
+++ tests/results/float_builtin_float_round.py.out	2025-07-05 09:00:15.756698139 +1000
@@ -7,7 +7,7 @@
 -123
 -124
 1.23457
-1.2
+1.1999998
 1.0
 1200.0
 -2
@@ -24,6 +24,6 @@
 0.0
 1.0
 1.5
-1.47
+1.4699998
 <class 'OverflowError'>
 <class 'ValueError'>

FAILURE tests/results/float_float1.py
--- tests/results/float_float1.py.exp	2025-07-05 09:00:18.611995322 +1000
+++ tests/results/float_float1.py.out	2025-07-05 09:00:18.612098402 +1000
@@ -1,15 +1,15 @@
-0.12
+0.119999976
 1.0
-1.2
+1.1999998
 0.0
 0.0
 0.0
-1.2
-1.2
+1.1999998
+1.1999998
 1.0
 10.0
 10.0
-0.1
+0.099999992
 inf
 -inf
 inf
@@ -21,12 +21,12 @@
 ValueError
 ValueError
 ValueError
-1.2
-3.4
+1.1999998
+3.3999996
 False
 True
-1.2
--1.2
+1.1999998
+-1.1999998
 0.5
 0.5
 0.0
```

Hmm...

yoctopuce:
I was able to reproduce the issue with a IMPL_FLOAT 32bit unix build.
The problem is not in the format code, but in the float parse code: with the latest code, in single-precision,
`1.2`
is parsed as
`1.19999980926513671875`
instead of
`1.2000000476837158203125`

The format code properly retrieves the exact `1.1999998` representation, which obviously breaks the test :-)

I will try to solve that in the coming days.

yoctopuce:
> I was able to reproduce the issue with a IMPL_FLOAT 32bit unix build. The problem is not in the format code, but in the float parse code

Erratum: the parse code works as intended. The problem is caused by REPR_C, which I had activated in my 32 bit test build.
With REPR_A, the `APPROX` implementation pass all tests, and the `BASIC` implementation only fails `float_format_accuracy` (makes sense...) and two minor rounding issues.

yoctopuce:
I started a discussion about improving the accuracy of REPR_C (or adding a new non-boxed 32-bit REPR) to avoid this kind of problem. My idea was to allocate more bits to floats at the expense of a lower limit for smallints. However, the feedback was not enthusiastic. See https://github.com/orgs/micropython/discussions/17566

yoctopuce:
To recap: the new formatting code works as intended: as it looks for the most accurate representation, it actually shows the "lost bits" of REPR_C (that would not have been shown with the old code). 

However if the APPROX code would take into account the fact that REPR_C has two bits less, it would not push the precision that high. So I will update the validation checks in the formatting code to to take into account the lost bits when REPR_C is active, and stop refining earlier. That should solve the issue.

yoctopuce:
Quick update: still working on it, I am almost done (but it takes a bit longer as I am out of office these days...)

yoctopuce:
I have pushed a new version which fixes most of the issues with `REPR_C`. The new code applies the optimal rounding method by adding mantissa digits progressively to all formatting operations where it could be beneficial, rather than just for `repr`. This provides a much nicer output for all numbers based on decimal exponents. I have been comparing the output of the following test code with CPython and MicroPython, using all float formats:
```py
for mant in [34567, 76543, 999999]:
    for exp in range(-16, 16):
        strnum = "%de%d" % (mant, exp)
        print("Next number: %s" % strnum)
        num = float(strnum)
        for mode in ['e', 'f', 'g']:
            maxprec = 16
            # MicroPython has a length limit in objfloat.c
            if mode == 'f' and 6 + exp + maxprec > 31:
                maxprec = 31 - 6 - exp
            for prec in range(0, maxprec):
                fmt = "%." + str(prec) + mode
                print("%5s: " % fmt, fmt % num)
```
With the latest code, even REPR_C outperforms CPython in providing a better representation without "phantom" digits.

Here is the code size and format reversibility score, including current method as a baseline for comparison:

|           |  PYBV10 code size | REPR_C score |  FLOAT score | DOUBLE score |
|--------|---------|---------|--------|---------|
|current | 364596 |  12.9% |  27.6% |  37.9% |
|basic   | 364712  | 85.6% |   60.5% |  85.7% |
| approx  | 364964 | 100.0% |  98.5% |  99.8% |
| exact  | 366408  | 100.0% | 100.0% |  100.0% |

I think there is still some room for improvement in this code, at least for one REPR_C rounding case, and to reduce the code size. But you can already give a try of the new code on esp8266 with REPR_BASIC.

dpgeorge:
Testing the latest code here on esp8266, BASIC mode costs +208 bytes, APPROX mode costs +448 bytes.

With BASIC mode, I get failures on 2 tests:
```diff
FAILURE micropython/tests/results/float_float_format_ints.py                     
--- micropython/tests/results/float_float_format_ints.py.exp
+++ micropython/tests/results/float_float_format_ints.py.out
@@ -478,6 +478,6 @@
 23456 x 10^4 with format {:.8e} gives 2.34560000e+08
 23456 x 10^4 with format {:.8f} gives 234560000.00000000
 23456 x 10^4 with format {:.8g} gives 2.3456e+08
-16777215.000000
+16777214.400000
 2147483520.000000
 1.000000e+38

FAILURE micropython/tests/results/float_float_format_accuracy.py                     
--- micropython/tests/results/float_float_format_accuracy.py.exp
+++ micropython/tests/results/float_float_format_accuracy.py.out
@@ -1,2 +1,2 @@
 1200 values converted
-float format accuracy OK
+FAILED: repr rate=85.583% max_err=5.934e-07
```

With APPROX mode, I get failures on 1 test:
```diff
FAILURE micropython/tests/results/float_float_format_ints.py                     
--- micropython/tests/results/float_float_format_ints.py.exp
+++ micropython/tests/results/float_float_format_ints.py.out
@@ -478,6 +478,6 @@
 23456 x 10^4 with format {:.8e} gives 2.34560000e+08
 23456 x 10^4 with format {:.8f} gives 234560000.00000000
 23456 x 10^4 with format {:.8g} gives 2.3456e+08
-16777215.000000
-2147483520.000000
+16777212.000000
+2147483200.000000
 1.000000e+38
```

But everything else passes, which is great!

yoctopuce:
@dpgeorge I have just pushed a new version that fixes the last rounding issues that I had. 

I have reintroduced some of the `fp_` functions (internal to `formatfloat`) which I suspect were originally added to reduce code size on platforms with inlined soft float support. Let me know what the code size looks like on esp8266.

I have tested with various combinations for the implementation options, including with the improved REPR_C algorithm, and I believe this new version should now work fine in all cases. I have intentionally lowered the accuracy requirement of REPR_C to 85% to allow REPR_C builds to use FORMAT_IMPL_BASIC and pass all CI tests.

dpgeorge:
I ran this PR (after rebasing on latest master) on esp8266 and got 1 failure:
```
FAILURE micropython/tests/results/float_float_format_ints.py
--- micropython/tests/results/float_float_format_ints.py.exp	
+++ micropython/tests/results/float_float_format_ints.py.out	
@@ -538,6 +538,6 @@
 23456 x 10^4 with format {:.9e} gives 2.345600000e+08
 23456 x 10^4 with format {:.9f} gives 234560000.000000000
 23456 x 10^4 with format {:.9g} gives 234560000
-16777215.000000
+16777212.000000
 2147483520.000000
 1.000000e+38
```

yoctopuce:
Interesting, this failure is not really surprising but I could not reproduce it on a unix build, even when using REPR_C in BASIC or APPROX mode.

I have nevertheless added the tolerance to the missing bits for this test as well, so now it should now be ok.

dpgeorge:
esp8266 now works OK in APPROX mode.

In BASIC mode I get:
```
FAILURE micropython/tests/results/float_float_format_ints.py                     
--- micropython/tests/results/float_float_format_ints.py.exp
+++ micropython/tests/results/float_float_format_ints.py.out
@@ -538,6 +538,6 @@
 23456 x 10^4 with format {:.9e} gives 2.345600000e+08
 23456 x 10^4 with format {:.9f} gives 234560000.000000000
 23456 x 10^4 with format {:.9g} gives 234560000
-16777215.000000
+16777214.400000
 2147483520.000000
 1.000000e+38

FAILURE micropython/tests/results/float_float_format_accuracy.py                      
--- micropython/tests/results/float_float_format_accuracy.py.exp
+++ micropython/tests/results/float_float_format_accuracy.py.out
@@ -1,2 +1,2 @@
 1200 values converted
-float format accuracy OK
+FAILED: repr rate=83.000% max_err=5.737e-07
```
I couldn't reproduce those errors on unix longlong variant in BASIC mode... so maybe that's due to the esp8266's math library and/or soft float math functions?

It would be nice to have the test suite pass in BASIC mode, although maybe it's not fully necessary.

We need to also decide if esp8266 uses BASIC or APPROX.  I was pushing for BASIC (see above), but would be fine if it used APPROX (and then the tests already pass).  BASIC costs +120, APPROX costs +468 (compared to master).  What's your opinion there?

yoctopuce:
> I couldn't reproduce those errors on unix longlong variant in BASIC mode... so maybe that's due to the esp8266's math library and/or soft float math functions?

Yes, I am quite sure that a small difference in the implementation of `pow()` function was also the reason for the previous failure in APPROX mode.

> It would be nice to have the test suite pass in BASIC mode, although maybe it's not fully necessary.

I think BASIC mode definitely makes sense on ports with limited ressources, not only for code space but also for speed. In particular, when used with REPR_C, the >80% reversibility rate is quite good compared to previous implementation which did <13%. It is quite interesting to see that BASIC mode does not add a single failure to `rge_sm.py` test.

> We need to also decide if esp8266 uses BASIC or APPROX. I was pushing for BASIC (see above), but would be fine if it used APPROX (and then the tests already pass). BASIC costs +120, APPROX costs +468 (compared to master). What's your opinion there?

I agree that it would be good to have at least one standard build using REPR_C + BASIC, to make sure that nobody breaks it. We could use the `unix/longlong` build for that, but given that we found out that the math library of esp8266 is less accurate, maybe it is a good idea to use it as a worst case scenario, also because code size and speed is a concern on that platform.

So I did push a new version which should pass BASIC and REPR_C on esp8266.

dpgeorge:
> It is quite interesting to see that BASIC mode does not add a single failure to `rge_sm.py` test.

Maybe because that test uses only 4 digits to print `{:4f}`?


> So I did push a new version which should pass BASIC and REPR_C on esp8266.

Thanks.  I tested on esp8266 in both APPROX and BASIC mode, and they both pass.  Well done!

Now, I think the only thing left is to enable BASIC on esp8266:
```diff
--- a/ports/esp8266/mpconfigport.h
+++ b/ports/esp8266/mpconfigport.h
@@ -106,6 +106,7 @@
 #define MICROPY_PY_OS_URANDOM       (1)
 #define MICROPY_LONGINT_IMPL        (MICROPY_LONGINT_IMPL_MPZ)
 #define MICROPY_FLOAT_IMPL          (MICROPY_FLOAT_IMPL_FLOAT)
+#define MICROPY_FLOAT_FORMAT_IMPL   (MICROPY_FLOAT_FORMAT_IMPL_BASIC)
 #define MICROPY_WARNINGS            (1)
 #define MICROPY_PY_STR_BYTES_CMP_WARN (1)
 #define MICROPY_STREAMS_POSIX_API   (1)
```
I think it makes sense to do that in this PR.

yoctopuce:
I have just pushed the switch to BASIC for esp8266.

dpgeorge (approved):

dpgeorge:
Now merged!  Thanks for all your effort here!

PR #17446, opened 2025-06-06
lwip: Support `family` specification in getaddrinfo.

greezybacon:
### Summary

socket.getaddrinfo() supports the specification of an address family; however, the LwIP implementation does not use it. This change allows the application to specify the address family request in DNS resolution. If no family is specified, it falls back to the default preference configured with network.ipconfig().

### Testing

I've done some minimal testing on this using an RP2040 and a Wiznet W5100S. It works as expected and returns the address of the family requested. For example:
```
>>> import socket
>>> socket.getaddrinfo('www.google.com', 80)
[(2, 1, 0, '', ('216.239.38.120', 80))]
>>> socket.getaddrinfo('www.google.com', 80, socket.AF_INET6)
[(2, 1, 0, '', ('2001:4860:4802:32::78', 80))]
>>> socket.getaddrinfo('www.google.com', 80, socket.AF_INET)
[(2, 1, 0, '', ('216.239.38.120', 80))]
```

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -1802,10 +1802,11 @@ static mp_obj_t lwip_getaddrinfo(size_t n_args, const mp_obj_t *args) {
     mp_obj_t host_in = args[0], port_in = args[1];
     const char *host = mp_obj_str_get_str(host_in);
     mp_int_t port = mp_obj_get_int(port_in);
+    mp_int_t family = 0;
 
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thanks, this is a nice improvement.

Tested on RPI_PICO2_W using the same test described above.  It works.

PR #17447, opened 2025-06-06
wiznet: lwip: Implement zero-copy for socket buffers.

greezybacon:
### Summary

This is closer to a zero-copy mechanism for the Wiznet chip and LwIP stack. Instead of using a static Ethernet frame buffer, it copies from the device directly into LwIP pbufs and visa versa.

Additionally, it avoids an issue for WIZnet devices:

  1. In the `wiznet5k_poll` method, if the chip indicates there is data available to receive and it is received but no buffer space is available in LwIP, then the packet will be fetched from the device and discarded. Instead, now it will remain on the device until sufficient memory is available in the MCU.

Furthermore, it optimizes and simplifies the usage of the WIZnet device based on the particular usage that the MACRAW mode will be used on exactly one socket on the device. It frees 1514 bytes of a static Ethernet frame buffer over the baseline.

### Testing

I've tested this pretty extensively on the RP2040 and WIZnet W5100S. Based on interest, I'm happy to test it further on other devices.

As written, it removes support for the W5200 and W5300 chips- although there is no reference board or firmware in the project for these devices.

However, insourcing a simplified driver could make for a simpler platform to add support for the W6100.

If Ethernet tracing is enabled, when traces are written to the console, if the LwIP pbuf is fragmented, only the first fragment is traced. This is to prevent the need for a copy of the buffer.

### DIFF
--- a/drivers/wiznet/macraw.c
+++ b/drivers/wiznet/macraw.c
@@ -1,0 +1,141 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 Jared Hancock <jared@greezybacon.me>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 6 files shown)


dpgeorge:
This looks like a really good improvement!

> As written, it removes support for the W5200 and W5300 chips

Hmm, that might be a problem.  The stm32 port has supported W5200 and W5500 ever since the beginning, and I still have a few lying around that can test with.  I guess W5200 is very old by now so we could remove support for it.  Users need to explicitly enable it for custom builds, so they could add it back themselves if needed (or stick with and older version of MicroPython).

greezybacon:
@dpgeorge I don't mind adding support back for the 5200 and/or 5300. Where could I get a demo/eval board to test against?

dpgeorge:
> Where could I get a demo/eval board to test against?

I don't know!

I'm happy to test.  If you can just add some code that looks reasonable and builds for the W5200, then I can test and feedback any changes that need to be made.

MilhouseVH:
#18035 adds more chips (W6100, W6300), boards (W55RP20, W5500/W6100/W6300-EVB-PICO[2]), and migrates from wiznet5k to wiznet6k.

Would it make more sense to land #18035 first, then rebase this PR wth support for the new chips, boards and wiznet6k?

PR #17452, opened 2025-06-07
ports: Provide VFS ROM support for the MIMXRT, SAMD, NRF and Renesas port.

robert-hh:
### Provide VFS ROM support for the MIXRT, SAMD, NRF and Renesas port.

This is a combined PR with commits for each of the mentioned ports. It's based on the implementation already merged
for ESP8266, ESP32 and RP2. The PR is meant as reminder that the implementation is still due for these ports.

### Testing

Tested with a set of boards from each of the ports. Details are shown in the respective commit messages.


dpgeorge:
Thanks @robert-hh for working on this, it's great to have ROMFS enabled for all the ports.

I would have reviewed and merged this earlier, but I really wanted to implement a partition table before going down the path of enabling ROMFS on all the ports and boards.  IMO a partition table is needed because there should be a way for the user to define how big the ROMFS and writable FS's are.

Prior to ROMFS there was just the firmware and the filesystem.  For boards with an external flash filesystem, the decision is easy: the firmware takes all the internal flash, and the filesystem takes all the external flash.  For boards with just internal (eg stm32 Nucleo boards) or just external flash (eg esp32, rp2) it's also not that hard of a decision, we just pick a boundary between firmware and filesystem such that the firmware fits and has a bit of room to grow.

But with ROMFS we now need to make two decisions: the boundary between firmware and ROMFS, and then between ROMFS and writable filesystem.  And it gets more complicated when some of the flash is memory mappable and some is not, because ROMFS must go in mappable flash.  There's really no one-size-fits all choice for how big the ROMFS should be.  Eg I have some projects when I use ROMFS and I need it to be very large (up to 1MiB) to store lots of font data.

So, we need a partition table to define the layout of the flash.  I have been slowly working on that, but it's a lot harder than I originally expected it would be.  When that lands, we will need to reorganise the flash layout of pretty much every board.  That's why I thought it best to wait until fully enabling ROMFS on all the boards, so we don't have to change everything again.

Anyway, I don't know when I'll be able to get a working implementation of a partition table.  So I'm happy to move forward with this PR now and merge it.

ports/mimxrt/Makefile
```diff
@@ -267,6 +267,11 @@ ifeq ($(USE_UF2_BOOTLOADER),1)
 	CFLAGS += -DMICROPY_MACHINE_UF2_BOOTLOADER=1
 endif
 
+# Set the default size of the VfsRom file system. Can be changed at board level
+ifeq ($(MICROPY_HW_ROMFS_BYTES),)
+	MICROPY_HW_ROMFS_BYTES ?= 0x40000
```
dpgeorge, line 316: Can we increase this a bit?  Most boards have a lot of flash and the firmware is not that big, so this could be maybe 512k?  Or at least 384k?

ports/nrf/modules/nrf/flashbdev.c
```diff
@@ -202,4 +216,34 @@ void flashbdev_init(void) {
     nrf_flash_obj.len = num_pages * FLASH_PAGESIZE;
 }
 
+#if MICROPY_VFS_ROM
+
+extern byte _vfsrom_start[];
+extern byte _vfsrom_size[];
+
+#define MICROPY_HW_ROMFS_BASE ((uint32_t)_vfsrom_start)
```
dpgeorge: Maybe export the symbols `_micropy_hw_romfs_part0_start/size` from the linker scripts, to be consistent with other ports like alif and stm32?

ports/nrf/modules/scripts/_boot.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+
+try:
+    bdev = nrf.Flash()
+    fs_type = (
+        vfs.VfsLfs2
+        if hasattr(vfs, "VfsLfs2")
+        else (vfs.VfsLfs1 if hasattr(vfs, "VfsLfs1") else vfs.VfsFat)
+    )
+    vfs.mount(fs_type(bdev), "/flash")
```
dpgeorge: It's great to move all this logic from C to Python.

But, the C version would actually auto-detect the FS type.  You can achieve the same thing by using `vfs.mount(bdev, "/flash")`, ie mount the `bdev` directly.

ports/nrf/modules/scripts/_boot.py
```diff
@@ -0,0 +1,31 @@
...
+import sys
+import nrf
+import os
+
+try:
+    bdev = nrf.Flash()
+    fs_type = (
+        vfs.VfsLfs2
+        if hasattr(vfs, "VfsLfs2")
+        else (vfs.VfsLfs1 if hasattr(vfs, "VfsLfs1") else vfs.VfsFat)
```
dpgeorge: Can do this, might be a little more efficient?
```py
getattr(vfs, "VfsLfs2", getattr(vfs, "VfsLfs1", getattr(vfs, VfsFat)))
```

ports/nrf/modules/scripts/_boot.py
```diff
@@ -0,0 +1,31 @@
+import gc
```
dpgeorge: Putting everything (including the imports) in a function and calling that can be more efficient and produce smaller bytecode.  It also means you don't need to del everything at the end, just del the function name.

ports/nrf/boards/nrf52840_1M_256k.ld
```diff
@@ -4,6 +4,7 @@
 
 _flash_size = 1M;
 _ram_size   = 256K;
+_vfsrom_size = 128K;
```
dpgeorge: For these 1M parts, can we increase ROMFS to maybe 256k?

ports/nrf/boards/nrf9160_1M_256k.ld
```diff
@@ -7,6 +7,7 @@ _ram_size   = 256K;
 _sd_size = 0x00008000;
 _sd_ram  = 0x00020000;
 _fs_size = 80K;
+_vfsrom_size = 128K;
```
dpgeorge: Increase to 256k?

ports/renesas-ra/boards/ARDUINO_PORTENTA_C33/ra6m5.ld
```diff
@@ -6,7 +6,8 @@
 MEMORY
 {
   FLASH_BOOT (r)     : ORIGIN = 0x00000000, LENGTH = 0x00010000  /* 64K */
-  FLASH (rx)         : ORIGIN = 0x00010000, LENGTH = 0x000f0000  /* 960KB */
+  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x000C0000  /* 768KB/2MB */
+  VFSROM_FS (r)      : ORIGIN = 0x000D0000, LENGTH = 0x00030000  /* 192KB/2MB */
```
dpgeorge: Suggest calling this section `FLASH_ROMFS` to match stm32 and esp8266.

Also, this MCU has a lot of flash so could increase ROMFS to 256k, if not 320k.

ports/renesas-ra/boards/ARDUINO_PORTENTA_C33/ra6m5.ld
```diff
@@ -304,3 +305,6 @@ _heap_end = __HeapLimit;	/* tunable */
 
 _micropy_hw_internal_flash_storage_start = ORIGIN(FLASH_FS);
 _micropy_hw_internal_flash_storage_end   = ORIGIN(FLASH_FS) + LENGTH(FLASH_FS);
+
+_micropy_hw_vfsrom_storage_start         = ORIGIN(VFSROM_FS);
+_micropy_hw_vfsrom_storage_len           = LENGTH(VFSROM_FS);
```
dpgeorge: Please use `_micropy_hw_romfs_part0_start/size` to match other ports.

ports/renesas-ra/boards/EK_RA4M1/ra4m1_ek.ld
```diff
@@ -3,9 +3,14 @@
...
 /* Linker script to configure memory regions. */
+/*
+ * VFSROM_FS and FLASH_FS must start and length must be a multiple
+ * of the physical sector size of that region (2k).
+ */
 MEMORY
 {
-  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x00037000  /* 220KB/256KB */
+  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x00034000  /* 208KB/256KB */
+  VFSROM_FS (r)      : ORIGIN = 0x00034000, LENGTH = 0x00003000  /* 12KB/256KB */
```
dpgeorge: Suggest using FLASH_ROMFS.

ports/renesas-ra/boards/EK_RA4M1/ra4m1_ek.ld
```diff
@@ -300,3 +305,5 @@ _heap_end = __HeapLimit;	/* tunable */
 
 _micropy_hw_internal_flash_storage_start = ORIGIN(FLASH_FS);
 _micropy_hw_internal_flash_storage_end   = ORIGIN(FLASH_FS) + LENGTH(FLASH_FS);
+_micropy_hw_vfsrom_storage_start         = ORIGIN(VFSROM_FS);
```
dpgeorge: Please rename as per comment above.

ports/renesas-ra/boards/EK_RA6M2/ra6m2_ek.ld
```diff
@@ -3,9 +3,15 @@
...
+
+/*
+ * VFSROM_FS and FLASH_FS must start and length must be a multiple
+ * of the physical sector size of that region (32k).
+ */
 MEMORY
 {
-  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x000e0000  /* 896KB/1MB */
+  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x000e0000  /* 768KB/1MB */
+  VFSROM_FS (r)      : ORIGIN = 0x000c0000, LENGTH = 0x00020000  /* 128KB/1MB */
```
dpgeorge: Could be a bit bigger?

ports/renesas-ra/boards/VK_RA6M5/vk_ra6m5.ld
```diff
@@ -5,7 +5,8 @@
 /* Linker script to configure memory regions. */
 MEMORY
 {
-  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x00100000  /* 1MB/2MB */
+  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x000C0000  /* 768KB/2MB */
+  VFSROM_FS (r)      : ORIGIN = 0x000C0000, LENGTH = 0x00040000  /* 256KB/2MB */
```
dpgeorge: Could be a bit bigger?

ports/renesas-ra/boards/EK_RA4W1/mpconfigboard.h
```diff
@@ -67,3 +67,5 @@
 #define MICROPY_HW_LED_ON(pin)      mp_hal_pin_low(pin)
 #define MICROPY_HW_LED_OFF(pin)     mp_hal_pin_high(pin)
 #define MICROPY_HW_LED_TOGGLE(pin)  mp_hal_pin_toggle(pin)
+
+#define MICROPY_VFS_ROM             (1)
```
dpgeorge: I think this can be removed, it's done in `mpconfigboard_common.h`.

ports/samd/Makefile
```diff
@@ -90,7 +90,7 @@ CFLAGS += $(CFLAGS_EXTRA)
 CFLAGS += -DMICROPY_HW_CODESIZE=$(strip $(subst K,' ', $(MICROPY_HW_CODESIZE)))
 
 LDFLAGS += -nostdlib $(addprefix -T,$(LD_FILES)) -Map=$@.map --cref
-LDFLAGS += --defsym=_codesize=$(MICROPY_HW_CODESIZE)
+LDFLAGS += --defsym=_codesize=$(MICROPY_HW_CODESIZE) --defsym=_vfsromsize=$(MICROPY_HW_VFSROMSIZE)
```
dpgeorge: Suggest using `MICROPY_HW_ROMFS_BYTES` here to match other ports.

ports/samd/boards/samd21x18a.ld
```diff
@@ -21,8 +21,18 @@ MEMORY
...
+That does not matter since these devices do not use the MCU flash file
+system.
+*/
+
 _oflash_fs = ORIGIN(FLASH) + _codesize;
-_sflash_fs = _flashsize - _codesize - _bootloader;
+_sflash_fs = _flashsize - _codesize - _bootloader - _vfsromsize;
+
+_oflash_vfsrom = ORIGIN(FLASH) + _flashsize - _vfsromsize - _bootloader;
+_sflash_vfsrom = _vfsromsize;
```
dpgeorge: Suggest using `_micropy_hw_romfs_part0_start/size` to match stm32 etc.

dpgeorge (commented):
I tested this on TEENS40 and it works well!

For samd: if possible I'd rather not change the size of the existing writable filesystems, which is the case when the board does not have external SPI flash.  Is it possible to only enable ROMFS on samd boards that have external SPI flash?  That way, users won't have to reformat their filesystem, and if they have external SPI flash then their boards will just be upgraded to have a new ROMFS (which takes up unused space at the end of the existing firmware section of flash).

robert-hh:
Thank you Damien for reviewing  the PR. I will try to implement the changes after Christmas. 
I agree that it would be best to have common approach for all ports & boards. That would also have an API to handle a partition table, making modifications independent from building and updating the firmware.

dpgeorge:
Thanks, there's no hurry.

ports/mimxrt/Makefile
robert-hh: Yes and no. I can be increased by reducing the VFS size. The VFS size if (flash size - 1M) except for MIMXRT1052 and MIMXRT1176, where it is (flash_size - 2M). If the VFS size is not touched, the VFSROM size  can be increased for MIMXRT101x boards and Teensy 4.0 to 512k, for other MIMXRT106x boards to 384k, increasing the use %age of the text section to 91%, but for MIMXRT1021 only 256k fits. MIMXRT 1052 and MIMXRTT1176 allow much larger VFSROM size.
so we have a few options:
1. Reduce the VFS size. That will cause all files on most boards to get lost. The board with the smallest flash size of 2M are Teensy 4.0 and OLIMEX RT1010.
2. Set different VFS sizes for each board. 
3. Use 384k for all boards except MIMXRT1021, which use 256k.But increasing the use %age of the text section to 91% does impose a limit on further changes.

ports/nrf/boards/nrf52840_1M_256k.ld
robert-hh: Done.

ports/nrf/modules/nrf/flashbdev.c
robert-hh: Done.

ports/nrf/boards/nrf9160_1M_256k.ld
robert-hh: Done

ports/nrf/modules/scripts/_boot.py
robert-hh: The statement should maybe be:
`getattr(vfs, "VfsLfs2", getattr(vfs, "VfsLfs1", getattr(vfs, "VfsFat",None)))`

ports/nrf/modules/scripts/_boot.py
robert-hh: Works. tested with a board supporting VfsLfs2.

ports/nrf/modules/scripts/_boot.py
robert-hh: Done. That is a good hint.

ports/nrf/modules/scripts/_boot.py
robert-hh: Simplified _boot.py accordingly.

ports/mimxrt/Makefile
dpgeorge: OK, I see.  I thought there was more flash available on all boards.

We definitely don't want to change the size of the existing writable filesystem.

So, I suggest keeping the default at 256k (which is quite a reasonable size), and then overriding that in MCU/board-specific files for those with large enough flash.  Eg MIMXRT1052 and MIMXRT1176 can go to 512k.  The rest can stay at 256k.

In the end it doesn't matter that much.  Eventually having a user-defined partition table will make this much better, the user can decide how much flash to use for what purpose.

robert-hh: OK. So I've set the ROMFS size to 512k for MIMXRT1176 and SEEED Arch Mix and disabled ROMFS for MIMXRT1052_EVKB. There seems to be a problem with Hyperflash and it's 256kB sector size. Must be analyzed separately.

ports/renesas-ra/boards/ARDUINO_PORTENTA_C33/ra6m5.ld
robert-hh: Done.

ports/renesas-ra/boards/EK_RA4M1/ra4m1_ek.ld
robert-hh: Done for all boards.

ports/renesas-ra/boards/EK_RA4M1/ra4m1_ek.ld
robert-hh: Done for all boards.

ports/renesas-ra/boards/ARDUINO_PORTENTA_C33/ra6m5.ld
robert-hh: The text segment has already a size of ~600k, so 256kB is better than 320kB.

ports/renesas-ra/boards/EK_RA6M2/ra6m2_ek.ld
robert-hh: Set to 256kB.

ports/renesas-ra/boards/VK_RA6M5/vk_ra6m5.ld
robert-hh: Set to 384kB.

ports/renesas-ra/boards/EK_RA4W1/mpconfigboard.h
robert-hh: Done

ports/samd/boards/samd21x18a.ld
robert-hh: Done

ports/samd/Makefile
robert-hh: Done

robert-hh:
@dpgeorge  Your review comments are implemented. There is the observation with MIMXRT: romfs fails with Hyperflash, probably due to the huge sector size of 256kB.

Note: The text segment for SAMD21 devices is full, sometimes 98-99%. ROMFS is consuming the space left at devices with SPI flash. Time for a diet and maybe drop ROMFS support for SAMD21.
Edit: Another option would be to drop MATH support, which gains 8k of code.

ports/mimxrt/Makefile
robert-hh: There is another commit now switching to use the symbols _micropy_hw_romfs_part0_start/size.

dpgeorge (approved):

I've squashed the commits to one per port, and made a few minor fixes:
- increased ROM partition size from 128k to 256k in `nrf9160_1M_256k.ld` (as we agreed, but I think you forgot to push that change)
- fixed `ARDUINO_PORTENTA_C33` so that the FLASH starts after the bootloader:
```diff
--- a/ports/renesas-ra/boards/ARDUINO_PORTENTA_C33/ra6m5.ld
+++ b/ports/renesas-ra/boards/ARDUINO_PORTENTA_C33/ra6m5.ld
@@ -6,7 +6,7 @@
 MEMORY
 {
   FLASH_BOOT (r)     : ORIGIN = 0x00000000, LENGTH = 0x00010000  /* 64K */
-  FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x000B0000  /* 704KB/2MB */
+  FLASH (rx)         : ORIGIN = 0x00010000, LENGTH = 0x000B0000  /* 704KB/2MB */
   FLASH_ROMFS (r)    : ORIGIN = 0x000C0000, LENGTH = 0x00040000  /* 256KB/2MB */
   FLASH_FS (r)       : ORIGIN = 0x00100000, LENGTH = 0x00100000  /* 1MB */
   RAM (rwx)          : ORIGIN = 0x20000000, LENGTH = 0x00080000  /* 512KB */
```
- fixed `VK_RA6M5` so FLASH_ROMFS doesn't run into FLASH_FS:
```diff
--- a/ports/renesas-ra/boards/VK_RA6M5/vk_ra6m5.ld
+++ b/ports/renesas-ra/boards/VK_RA6M5/vk_ra6m5.ld
@@ -6,7 +6,7 @@
 MEMORY
 {
   FLASH (rx)         : ORIGIN = 0x00000000, LENGTH = 0x000A0000  /* 640KB/2MB */
-  FLASH_ROMFS (r)    : ORIGIN = 0x000A0000, LENGTH = 0x000A0000  /* 384KB/2MB */
+  FLASH_ROMFS (r)    : ORIGIN = 0x000A0000, LENGTH = 0x00060000  /* 384KB/2MB */
   FLASH_FS (r)       : ORIGIN = 0x00100000, LENGTH = 0x00100000  /* 1MB/2MB */
   RAM (rwx)          : ORIGIN = 0x20000000, LENGTH = 0x00080000  /* 512KB */
   OSPI_RAM (rwx)     : ORIGIN = 0x68000000, LENGTH = 0x00800000  /* 8MB/8MB */
```

PR #17468, opened 2025-06-10
github/workflows: Stop using windows-2019 runner for Windows CI.

dpgeorge:
### Summary

This runner has been deprecated by GitHub.

### Testing

CI should test this.

### DIFF
--- a/.github/workflows/ports_windows.yml
+++ b/.github/workflows/ports_windows.yml
@@ -28,13 +28,10 @@ jobs:
         visualstudio: ['2017', '2019', '2022']
         include:
         - visualstudio: '2017'
-          runner: windows-latest
           vs_version: '[15, 16)'
         - visualstudio: '2019'
-          runner: windows-2019
           vs_version: '[16, 17)'
         - visualstudio: '2022'
-          runner: windows-2022
           vs_version: '[17, 18)'
         # trim down the number of jobs in the matrix
         exclude:
@@ -42,9 +39,9 @@ jobs:
           configuration: Debug
         - visualstudio: '2019'
           configuration: Debug
+    runs-on: windows-latest
     env:
       CI_BUILD_CONFIGURATION: ${{ matrix.configuration }}
-    runs-on: ${{ matrix.runner }}
     steps:
     - name: Install Visual Studio 2017
       if: matrix.visualstudio == '2017'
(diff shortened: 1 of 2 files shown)


dpgeorge:
@stinos FYI.  I'm not sure what the best thing to do is, but I simply replaced `windows-2019` with `windows-2025`.

dlech:
The windows-2025 builder only has VS 2022 installed. So if we drop windows-2019, we will need to install VS 2019 manually like we do for VS 2017.

I.e. duplicate this and adjust for Visual Studio 2019. (And we can just make it window-latest.)

https://github.com/micropython/micropython/blob/5f4abeb3859c948cc4a65221af6013a95533cdc4/.github/workflows/ports_windows.yml#L49-L54

.github/workflows/ports_windows.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       with:
         vs-version: ${{ matrix.vs_version }}
     - uses: actions/setup-python@v5
-      if: matrix.runner == 'windows-2019'
+      if: matrix.runner == 'windows-2025'
       with:
         python-version: '3.9'
```
dlech: Last time I checked, windows-2025 still uses python 3.9 by default, so we can just drop this whole section if we are dropping use of the windows-2019 runner. This was only needed because windows-2019 used an older python version by default.

dpgeorge: Thanks, I've now removed this setup-python action.

dpgeorge:
> The windows-2025 builder only has VS 2022 installed. So if we drop windows-2019, we will need to install VS 2019 manually like we do for VS 2017.

Thanks for the hint, that's now done.

But now we get errors again about NAN not being a constant... any ideas how to extend the existing `defined(_MSC_VER) && _MSC_VER >= 1942` check to make the CI work again?

.github/workflows/ports_windows.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           runner: windows-latest
           vs_version: '[15, 16)'
         - visualstudio: '2019'
-          runner: windows-2019
+          runner: windows-latest
           vs_version: '[16, 17)'
         - visualstudio: '2022'
           runner: windows-2022
```
stinos: Maybe this should also be `windows-latest` for consistency? In the end we don't really care on what windows (or whatever GH makes from it) version this runs. The build tools are the key difference.

stinos:
> But now we get errors again about NAN not being a constant... any ideas how to extend the existing `defined(_MSC_VER) && _MSC_VER >= 1942` check to make the CI work again?

Based on https://stackoverflow.com/questions/79195142/recent-msvc-versions-dont-treat-nan-as-constant-workaround/79324199#79324199 the original commit only worked by accident at that time, because this is not actually related to compiler version (i.e. `_MSC_VER`)  but to the window SDK version. I think

```
#if defined(_MSC_VER) && !defined(_UCRT_NOISY_NAN)
#define _UCRT_NOISY_NAN 
#endif
```

should be good.

.github/workflows/ports_windows.yml
dpgeorge: I thought the intention was to test on different Windows version?  But I'm happy to change them all to `windows-latest`.

stinos: > I thought the intention was to test on different Windows version?

@dlech was it? Or rather just a byproduct of which image contained which VS version ?

dpgeorge:
I've updated it to use `_UCRT_NOISY_NAN` and indeed it looks like that's a good solution.

.github/workflows/ports_windows.yml
dlech: It was just a byproduct of VS version. So, yes, we could pull that out of the matrix and just set the runner to always be `windows-latest`.

dpgeorge: OK, I've made everything use `windows-latest` and factored that out of the matrix.

dlech (approved):

dpgeorge:
Thanks for the quick review on this @stinos and @dlech .

PR #17470, opened 2025-06-10
cpydiff: Document complex() parsing difference.

jepler:
### Summary

In https://github.com/micropython/micropython/pull/17384 it was decided that fixing this difference was not worth the code size increase.

### Testing

Built & viewed the documentation locally
image

### DIFF
--- a/tests/cpydiff/types_complex_parser.py
+++ b/tests/cpydiff/types_complex_parser.py
@@ -1,0 +1,14 @@
+"""
+categories: Types,complex
+description: MicroPython's complex() accepts certain incorrect values that CPython rejects
+cause: MicroPython is highly optimized for memory usage.
+workaround: Do not use non-standard complex literals as argument to complex()
+
+MicroPython's ``complex()`` function accepts literals that contain a space and
+no sign between the real and imaginary parts, and interprets it as a plus.
+"""
+
+try:
+    print(complex("1 1j"))
+except ValueError:
+    print("ValueError")


tests/cpydiff/types_complex_parsing.py
```diff
@@ -0,0 +1,14 @@
+"""
```
dpgeorge: I suggest calling this file `types_complex_parser.py`, to match `core_fstring_parser.py`.

tests/cpydiff/types_complex_parsing.py
```diff
@@ -0,0 +1,14 @@
...
+categories: Types,complex
+description: uPy's complex() accepts certain incorrect values that CPython rejects
```
dpgeorge: Please use "MicroPython" instead of "uPy", because not everyone knows that the latter is shorthand.

(Would be good to change remaining uPy's in the cpydiff directory, eventually.)

tests/cpydiff/types_complex_parsing.py
```diff
@@ -0,0 +1,14 @@
...
+cause: MicroPython is highly optimized for memory usage.
+workaround: Do not use non-standard complex literals as argument to complex()
+
+MicroPython's ``complex()`` function accepts literals that contain a space and
+no sign between the real and imaginary parts.
```
dpgeorge: ", and interprets it as a plus."

dpgeorge (approved):

dpgeorge:
Rebased and merged in 5901b8d149d63d74354bfb3f7f44a0b942eda23e

PR #17472, opened 2025-06-10
tests: create a helper function to generate a test report and use it in all test runners

dpgeorge:
### Summary

This PR factors existing code in `run-tests.py` into a new helper function `create_test_report()`.  That function prints out a summary of the test run (eg number of tests passed, number failed, number skipped) and creates the `_results.json` file.

Then that function is reused by the three other test runners: `run-multitests.py`, `run-natmodtests.py` and `run-perfbench.py`.  So those tests runners now also generate a `_results.json` file with exactly the same format as `run-tests.py`.

This is useful, eg, for Octoprobe, which aims to print a concise summary of all tests runs that were performed.  It's very useful in that case that the test summary is in a computer-readable file and follows the same format.

For consistency this PR makes a minor change to the printed output of `run-tests.py`: instead of printing a shorthand name for tests that failed/skipped, it now prints the full name.  Eg what was previously printed as `attrtuple2` is now printed as `basics/attrtuple2.py`.  This makes the output a little longer (when there are failed/skipped tests) but helps to disambiguate the test name, eg which directory it's in.

### Testing

Ran the four test runners on the unix port and observed the output, and inspected the `_results.json` file.

CI will also test this change.

I tried to reuse code as much as possible, so there's less code duplication and the output format is consistent across the runners.  This had some minor trade-offs:
1. as mentioned above, `run-tests.py` now prints the full filename of failed/skipped tests
2. `run-multitests.py` when run with multiple permutations only saves the last run as `_results.json`; this isn't an issue for Octoprobe which only runs this with a single permutation
3. the three tests runners that now save a `_results.json` file will use the same output directory `tests/results/` by default, and hence overwrite any existing file (eg from `run-tests.py`); a user can always provide a different result directory to overcome this

### DIFF
--- a/tests/run-multitests.py
+++ b/tests/run-multitests.py
@@ -15,6 +15,8 @@ import itertools
 import subprocess
 import tempfile
 
+run_tests_module = __import__("run-tests")
+
 test_dir = os.path.abspath(os.path.dirname(__file__))
 
 if os.path.abspath(sys.path[0]) == test_dir:
@@ -488,9 +490,7 @@ def print_diff(a, b):
 
 
 def run_tests(test_files, instances_truth, instances_test):
-    skipped_tests = []
-    passed_tests = []
-    failed_tests = []
+    test_results = []
 
     for test_file, num_instances in test_files:
         instances_str = "|".join(str(instances_test[i]) for i in range(num_instances))
@@ -526,13 +526,13 @@ def run_tests(test_files, instances_truth, instances_test):
         # Print result of test
         if skip:
             print("skip")
-            skipped_tests.append(test_file)
+            test_results.append((test_file, "skip", ""))
         elif output_test == output_truth:
             print("pass")
-            passed_tests.append(test_file)
+            test_results.append((test_file, "pass", ""))
         else:
             print("FAIL")
-            failed_tests.append(test_file)
+            test_results.append((test_file, "fail", ""))
             if not cmd_args.show_output:
                 print("### TEST ###")
                 print(output_test, end="")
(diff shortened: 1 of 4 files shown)


dpgeorge:
@hmaerki hopefully this can be used to fill in the last entries in the summary table.  Please try it out and let me know!

hmaerki:
I run tests against this PR.

tests/run-natmodtests.py does not add results to the 'results', see https://reports.octoprobe.org/github_selfhosted_testrun_64/RUN-NATMODTESTS@1133-TEENSY40/_results.json.

In https://reports.octoprobe.org/github_selfhosted_testrun_64/RUN-NATMODTESTS@1133-TEENSY40/testresults.txt we see, that all tests have been skipped (extmod/random_seed_default.py - mpy file not compiled).

A) I would expect all tests to be skipped. Could you fix that?
B) How to I have to run the tests correctly in Octoprobe? Do I have to mpy-compile the files first?

dpgeorge:
> I run tests against this PR.

Great, thanks!  I can see that the results from `run-perfbench.py` and `run-multitests.py` work correctly.

For `run-natmodtests.py`: to get these to run we need to compile the module before running.  This is a bit tricky because each module needs to be compiled for each architecture, and you need to make sure the compiler is in the path.

Eg:
```
$ cd examples/natmod/heapq
$ make ARCH=armv6m
```
That will build the `heapq_armv6m.mpy` file which can then be used on boards that are ARM based, eg RPI_PICO.

You need to do that in each of the `heapq`, `random`, `re`, `framebuf`, `btree` and `deflate` subdirs within `examples/natmod`.  And for each of those you need to go through all values of ARCH: `arm6vm`, `armv7m`, `armv7emsp`, `armv7emdp`, `xtensa`, `xtensawin` and `rv32imc`.

We should discuss that issue over at the Octoprobe repo.

hmaerki:
Could you add this line?

```python
        except OSError:
            **test_results.append((test_file, "skip", "mpy file not compiled"))**
            print("----  {} - mpy file not compiled".format(test_file))
            continue
```

Then I will run another test.

dpgeorge:
> Could you add this line?

Yes, now done.

hmaerki:
This testrun includes the last commit: https://reports.octoprobe.org/github_selfhosted_testrun_69/octoprobe_summary_report.html

I reviewed the PR - looks fine to me - please merge!

dpgeorge:
Octoprobe summary is looking really good now!

PR #17477, opened 2025-06-11
py/emitnative: Some more load/store improvements.

agatti:
### Summary

Here's the last - I hope - PR with Viper load/store improvements (well, for native too since they share those emitter primitives).

Some ports should, in theory, actually shrink down in size:

- Xtensa has one single function per operation, each covering all operand sizes by flipping bits in a single instruction template (like the Thumb emitter in the previous PR in the series), still using narrow opcodes whenever possible
- RV32 also has one single function per operation with a single function template each, with even more duplicated code removed, now emitting compressed opcodes whenever possible (even for store operations)
- x86 already had the necessary opcode emitters to cover all load/store cases but they weren't exposed to Viper, so all operations should be one opcode each and - in theory - emit faster code
- x64, like x86, also had all the necessary opcode emitters already there, with the same performance benefits

However, the Arm emitter is probably slightly bigger since there's one function per operation/size combination.  I've tried to offset this by merging some single-opcode emitter functions into entities that generate the shortest sequence for all offsets.  Halfword opcodes have a different format so I'd still have four functions rather than six - not really worth the effort for such a platform that, as I understand it, isn't as memory constrained as other ones supported by MicroPython.

Thumb only got redundant single-opcode emitters removed, so it's equivalent as far as emitted code goes.

I haven't replaced all the bit-width fit checks in the various emitters with the new macros, in order to keep this patch series more readable.  For the time being I'll replace those as I encounter them in code that gets changed.

Boundary tests were cleaned up and improved as well - I didn't go as far as using unittest's sub-tests (one per bit width tested), but if that's desired I can migrate those tests to that framework.

Who would have thought that getting and setting simple variables would have triggered all this work? :)

### Testing

`tests/mcropython/viper_ptr*` were executed successfully on the following ports:

- QEMU (`VIRT_RV32`, `SABRELITE`, and `MPS_AN385`) both directly and with `--via-mpy --emit native`
- ESP8266 (NodeMCU v3) both directly and with `--via-mpy --emit native` using the same prelude injection method introduced for natmod tests (`-b` was already taken in `tests/run-tests.py`, so only `--begin` is available)
- ESP32 (generic D0WD board), both directly and with `--via-mpy --emit native`
- x64 Unix using the STANDARD variant directly
- x86 Unix using the COVERAGE variant directly

As mentioned, the Arm emitter is slightly larger and it can be made a bit smaller if it's required - adding some complexity to that emitter backend.


py/asmrv32.c
```diff
@@ -450,31 +450,37 @@ void asm_rv32_emit_mov_reg_local_addr(asm_rv32_t *state, mp_uint_t rd, mp_uint_t
...
 }
 
-void asm_rv32_emit_load_reg_reg_offset(asm_rv32_t *state, mp_uint_t rd, mp_uint_t rs, mp_int_t offset) {
-    mp_int_t scaled_offset = offset * sizeof(ASM_WORD_SIZE);
+static const uint8_t RV32_LOAD_OPCODE_TABLE[3] = {
+    0x04, 0x05, 0x02
+};
 
-    if (scaled_offset >= 0 && RV32_IS_IN_C_REGISTER_WINDOW(rd) && RV32_IS_IN_C_REGISTER_WINDOW(rs) && FIT_UNSIGNED(scaled_offset, 6)) {
+void asm_rv32_emit_load_reg_reg_offset(asm_rv32_t *state, mp_uint_t rd, mp_uint_t rs, int32_t offset, mp_uint_t shift) {
```
dpgeorge: It might be worth adding a comment for this function (and the store function, and for other archs that are the same) mentioning that `shift` is both a shift of the `offset` (ie `offset << shift`) **and** determines the size of the load/store, ie byte/half-word/word.

tests/run-tests.py
```diff
@@ -297,6 +300,14 @@ def prepare_script_for_target(args, *, script_text=None, force_plain=False):
         rm_f(mpy_filename)
         rm_f(script_filename)
 
+        prologue = ""
+        if args.begin:
+            with open(args.begin, "rt") as source:
+                prologue = source.read()
+        injected_import_hook_code = injected_import_hook_code.replace(
```
dpgeorge: This code looks like it does the replace for each test, and only the first time will it do anything.

I suggest following 9ef7322a5d11937e061b85da12900eeee62e9891 and replacing the prologue only once at the start of the entire test run.

Also could use `type=argparse.FileType("rt")` as in that commit?

dpgeorge (commented):
Thanks for this, it looks really good!  A nice clean up and enhancement.

py/asmrv32.h
```diff
@@ -751,13 +751,20 @@ void asm_rv32_emit_store_reg_reg_offset(asm_rv32_t *state, mp_uint_t source, mp_
...
 #define ASM_STORE32_REG_REG(state, rs1, rs2) ASM_STORE32_REG_REG_OFFSET(state, rs1, rs2, 0)
-#define ASM_STORE32_REG_REG_OFFSET(state, rd, rs, offset) asm_rv32_emit_store_reg_reg_offset(state, rd, rs, offset)
-#define ASM_STORE8_REG_REG(state, rs1, rs2) asm_rv32_opcode_sb(state, rs1, rs2, 0)
+#define ASM_STORE8_REG_REG_OFFSET(state, rd, rs, offset) asm_rv32_emit_store_reg_reg_offset(state, rd, rs, offset, 0)
+#define ASM_STORE16_REG_REG_OFFSET(state, rd, rs, offset) asm_rv32_emit_store_reg_reg_offset(state, rd, rs, offset, 1)
+#define ASM_STORE32_REG_REG_OFFSET(state, rd, rs, offset) asm_rv32_emit_store_reg_reg_offset(state, rd, rs, offset, 2)
 #define ASM_SUB_REG_REG(state, rd, rs) asm_rv32_opcode_sub(state, rd, rd, rs)
 #define ASM_XOR_REG_REG(state, rd, rs) asm_rv32_emit_optimised_xor(state, rd, rs)
 #define ASM_CLR_REG(state, rd)
+#define ASM_LOAD8_REG_REG_REG(state, rd, rs1, rs2) \
```
dpgeorge, line 763: It looks like this particular macro isn't needed?  If it's omitted then I think `emitnative.c` will generate the same code using its generic fall back code.

(Same comment for `ASM_STORE8_REG_REG_REG`.)

tests/run-tests.py
agatti: Oh OK, I'll look into this - thanks!  I cannot really test this particular change right now, though, as most of my dev hardware is still packed up after a move.

> Also could use type=argparse.FileType("rt") as in that commit?

It makes thing slightly more complicated: the bits generating the JSON report files also serialise the parsed input arguments, and `argparse.FileType` requires passing a fallback serialisation function to `json.dump`/`json.dumps`.

py/asmrv32.c
agatti: Good point.  Would an enum rather than a comment make things clearer for that?

An earlier iteration of this PR had this in `py/asmbase.h`:

```c
// By how many bits an offset should be shifted left to obtain its efffective
// byte offset.
//
// Terminology follows non-x86/x64 architectures.  For x86/x64 the names 
// should read BYTE, WORD, DWORD, and QWORD instead.
typedef enum {
    MP_ASM_OFFSET_SHIFT_BYTE = 0,
    MP_ASM_OFFSET_SHIFT_HALFWORD,
    MP_ASM_OFFSET_SHIFT_WORD,
    MP_ASM_OFFSET_SHIFT_DOUBLEWORD
} mp_asm_shift_amount_t;
```

but then I just kept the raw `mp_uint_t`s in there to not make define lines too long.

If the enum is fine then I'll just tweak the names (`OFFSET_SHIFT` → `OPERAND_SIZE` and `shift_amount` → `operand_size`) and add a commit to the PR with the relevant changes.

tests/run-tests.py
dpgeorge: > It makes thing slightly more complicated: the bits generating the JSON report files also serialise the parsed input arguments, and `argparse.FileType` requires passing a fallback serialisation function to `json.dump`/`json.dumps`.

OK, then just leave that bit as-is.

But still move the prologue replacement code either to `main()` or near the start of `run_tests()`.

py/asmrv32.c
dpgeorge: Even just renaming `shift` to `operation_size` would probably be documenting enough.

py/asmrv32.h
agatti: Right now they act the same as the fallback indeed, the reason for their existence is that it was some preparatory work for further changes in the RV32 emitter I planned to introduce (optional support for more RV32 C sub-extensions that introduce more compressed opcodes).  This technically applies to `ASM_LOAD16_REG_REG_REG` and `ASM_STORE16_REG_REG_REG` too, as whilst the emitted code sequence is different the number of opcodes and the final sequence size is the same.

I don't mind taking them out now to potentially re-introduce them later (even if under a different form).

tests/run-tests.py
agatti: Done - I managed to find a stray ESP8266 board in one of the boxes to test the new change with.

py/asmrv32.c
agatti: OK, that should be addressed in the latest iteration of the PR.

py/asmrv32.h
dpgeorge: OK, it' fine to keep them.

dpgeorge (approved):

PR #17414, opened 2025-06-02
unix: Introduce sanitize_undefined variant, use during CI.

jepler:
### Summary

gcc's "undefined behavior" sanitizer can catch a range of misbehaviors at runtime that normally go unnoticed. These include integer and pointer operations that are "undefined" per the relevant C specification. Over time, most of these problems have been fixed through other PRs but if micropython desn't have regular CI time checks, regressions are inevitable.

This PR fixes current undefined behavior detected under gcc 12.2.0 (debian stable/bookworm) on an x64 system, then enables it during a new unix "coverage-like" build.

### Testing

I built and ran the unix tests locally, iterating until there were no remaining diagnostics.

### Trade-offs and Alternatives

#15303 is an alternate implementation, but it's become stale and did not cleanly merge with current micropython. That's why I'm re-opening my version of the changes.

Not all gcc sanitizers can be enabled simultaneously. So, a choice has to be made (mainly between -fsanitize=undefined and -fsanitize=memory). I chose the undefined checker, but implemented it so that an override of the makefile variable is possible.

-fsanitize=memory is may be error free now and can be added as a separate PR.

I read that [a future C specifiction](https://developers.redhat.com/articles/2024/12/11/making-memcpynull-null-0-well-defined) will make e.g., `memset(NULL, 0, 0)` (setting zero bytes of a NULL pointer) not-undefined-behavior. For this reason, and because my inspection didn't find any current incorrect optimizations due to `-fdelete-null-pointer-checks` [for sites that hit UB sanitizer messages], I chose to simplify this PR by disabling the use of nonnull-pointer annotations by the sanitizer.

Different gcc versions might be too different in what they affect, as any diagnostics will make `make test` fail due to the unexpected output.

Pattern-based tests which might inadvertently skip over output text that comes from the sanitizer, giving false negatives.

PR #17480, opened 2025-06-11
mpz: Avoid undefined behavior decrementing NULL.

jepler:
### Summary

In the case where an mpz number is zero, its `len` is 0 and its `dig` is NULL. In that case, decrementing NULL via `d--` is undefined behavior according to the C specification.

Restructuring the loop in this way avoids undefined behavior.

### Testing

I ran the unix coverage tests locally

A different and probably worse fix is in https://github.com/micropython/micropython/pull/17414 which only changes the loop when a special coverage-build flag is enabled.

### DIFF
--- a/py/mpz.c
+++ b/py/mpz.c
@@ -1537,7 +1537,8 @@ mp_int_t mpz_hash(const mpz_t *z) {
     mp_uint_t val = 0;
     mpz_dig_t *d = z->dig + z->len;
 
-    while (d-- > z->dig) {
+    while (d > z->dig) {
+        d--;
         val = (val << DIG_SIZE) | *d;
     }
 
@@ -1552,11 +1553,12 @@ bool mpz_as_int_checked(const mpz_t *i, mp_int_t *value) {
     mp_uint_t val = 0;
     mpz_dig_t *d = i->dig + i->len;
 
-    while (d-- > i->dig) {
+    while (d > i->dig) {
         if (val > (~(MP_OBJ_WORD_MSBIT_HIGH) >> DIG_SIZE)) {
             // will overflow
             return false;
         }
+        d--;
         val = (val << DIG_SIZE) | *d;
     }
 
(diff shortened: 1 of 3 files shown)


dpgeorge (approved):
This looks fine to me.

jepler:
There are two similar loops, only one is hit in coverage testing in a way that triggers the undefined behavior sanitizer. I should probably fix both at once, if you want to hold off merging this PR.

jepler:
There are two CI failures but I don't know how my changes would be related to them.

dpgeorge:
Failing CI was re-run, now all green.

py/mpz.c
```diff
@@ -1537,8 +1537,8 @@ mp_int_t mpz_hash(const mpz_t *z) {
     mp_uint_t val = 0;
     mpz_dig_t *d = z->dig + z->len;
 
-    while (d-- > z->dig) {
-        val = (val << DIG_SIZE) | *d;
+    while (d > z->dig) {
+        val = (val << DIG_SIZE) | *--d;
```
projectgus:
```suggestion
        d--;
        val = (val << DIG_SIZE) | *d;
```

This is kinda nitpicky so feel free to disregard, but this arrangement seems more readable to me: no longer necessary to parse out the order of operations in the expression.

(I know damien has previously expressed a preference for prefix decrements, but I'd be shocked if any optimising C compiler produces different final output between these two.)

projectgus (approved):
Nice change, thanks @jepler! Thanks for resurrecting the UBSan work, as well!

py/mpz.c
jepler: style changes made.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 0ef5ede3823ef00e95d1cc1a5b475d8d30c9caa4

PR #17483, opened 2025-06-12
objlist: Reduce code size in slicing.

jepler:
### Summary

By refactoring the code to separate out the slicing operation from the regular indexing operation, code can be shared between the various types of slice operations (read/assign/delete)

### Testing

I ran the unix coverage tests locally. I also checked the size improvement as 36 bytes on `ports/qemu` building `BOARD=MPS2_AN385` (improved by subsequent tweaks)

We investigated several alternatives and the current one saves more code but is the larger change.

I did a quick check and don't think any other core type would benefit from this. array objects don't have slice deletion.

### DIFF
--- a/py/objlist.c
+++ b/py/objlist.c
@@ -159,66 +159,63 @@ static mp_obj_t list_binary_op(mp_binary_op_t op, mp_obj_t lhs, mp_obj_t rhs) {
 }
 
 static mp_obj_t list_subscr(mp_obj_t self_in, mp_obj_t index, mp_obj_t value) {
-    if (value == MP_OBJ_NULL) {
-        // delete
-        #if MICROPY_PY_BUILTINS_SLICE
-        if (mp_obj_is_type(index, &mp_type_slice)) {
-            value = mp_const_empty_tuple;
-            goto slice_assign;
-        }
-        #endif
-        mp_obj_t args[2] = {self_in, index};
-        list_pop(2, args);
-        return mp_const_none;
-    } else if (value == MP_OBJ_SENTINEL) {
-        // load
+    #if MICROPY_PY_BUILTINS_SLICE
+    if (mp_obj_is_type(index, &mp_type_slice)) {
         mp_obj_list_t *self = MP_OBJ_TO_PTR(self_in);
-        #if MICROPY_PY_BUILTINS_SLICE
-        if (mp_obj_is_type(index, &mp_type_slice)) {
-            mp_bound_slice_t slice;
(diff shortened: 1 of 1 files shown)


py/objlist.c
```diff
@@ -163,19 +163,8 @@ static mp_obj_t list_subscr(mp_obj_t self_in, mp_obj_t index, mp_obj_t value) {
...
-
-            mp_int_t len_adj = slice.start - slice.stop;
-            assert(len_adj <= 0);
-            mp_seq_replace_slice_no_grow(self->items, self->len, slice.start, slice.stop, self->items /*NULL*/, 0, sizeof(*self->items));
-            // Clear "freed" elements at the end of list
-            mp_seq_clear(self->items, self->len + len_adj, self->len, sizeof(*self->items));
-            self->len += len_adj;
-            return mp_const_none;
+            value = mp_const_empty_tuple;
+            goto slice_assign;
```
dlech: Saving a few bytes is always nice, but this kind of jumping is how goto got a bad name, yikes!

dpgeorge: I don't think this is too bad coding style.

But it might be able to improve the style by shifting the "load" check to be first, then combining "delete" and "store" into one check.  That may also help reduce code size further by being able to combine the `if (mp_obj_is_type(index, &mp_type_slice)) {` check.

jepler: I went further and changed the whole function so that slices are handled separately from non-slices. It saved a further 16 bytes in a qemu arm build in my testing.

Here are the values from the first iteration, for comparison:
```
   bare-arm:    +0 +0.000% 
minimal x86:    +0 +0.000% 
   unix x64:  -120 -0.014% standard
      stm32:   -40 -0.010% PYBV10
     mimxrt:   -40 -0.011% TEENSY40
        rp2:   -40 -0.004% RPI_PICO_W
       samd:   -40 -0.015% ADAFRUIT_ITSYBITSY_M4_EXPRESS
  qemu rv32:   -46 -0.010% VIRT_RV3
```

If this variant is preferred I can squash it, there would be little reason to preserve the intermediate step that uses a goto.

dpgeorge: OK!  Let's see how it goes with the CI here and the code size comparison.

(Note: you don't need to save the code size from previous runs, it's possible to see previous ones by looking at the edits by the bot on the comment.)

dpgeorge (approved):
Very nice!

dpgeorge:
Rebased, squashed and merged in 73b1cbc17abe6da6f913d30dad960e5401577c0f

PR #17487, opened 2025-06-13
rp2: Increase configurability of boards

PepperoniPingu:
### Summary

I'm making my own board definition where there is a need to modify some of the default definitions: disabling i2c, spi, i2s, and changing the float implementation and the gc stack entry type. Since most of the parameters are hard coded in `mpconfigboard.h` this pr adds some `#ifndef ... #endif`. These changes are already scattered in other ports so nothing new.

### Testing

Tested on a custom rp2350 board and passes CI. 

`mpconfigboard.h` is a bit more cluttered now. Ideally this treatment would be done to every configurable parameter but I don't have the resources to test that so starting small.


ports/rp2/mpconfigport.h
```diff
@@ -86,12 +86,14 @@
 #endif
 
 // Memory allocation policies
+#ifndef MICROPY_GC_STACK_ENTRY_TYPE
```
Gadgetoid: Is there a reason for this change in particular? Defining `MICROPY_GC_STACK_ENTRY_TYPE` would disable PSRAM unconditionally, which seems counter-intuitive.

Possibly should be moved to the `#else` branch of `#if MICROPY_HW_ENABLE_PSRAM`?

In cases where someone configures their board with:

```
#define MICROPY_HW_ENABLE_PSRAM                   1
#define MICROPY_GC_STACK_ENTRY_TYPE             uint16_t
```

Then we should probably fail the build with an error. An example of that in practise is:

https://github.com/micropython/micropython/blob/398da22492d9e593f87f0b9ae4fc043f8b2b8a11/ports/rp2/rp2_psram.h#L34-L37

Gadgetoid:
> I'm making my own board definition

I am intrigued.

> Ideally this treatment would be done to every configurable parameter

I tend to agree. For user-facing, firmware-like deployments there are good reasons to disable interfaces- though afaik there's only really precedent for this in ports (nRF for example) where the SOC might be missing these interfaces. 

> but I don't have the resources to test

Probably no harm in allowing PRs like this to enable port config overrides as and when they are needed.

ports/rp2/mpconfigport.h
PepperoniPingu: > Is there a reason for this change in particular? Defining `MICROPY_GC_STACK_ENTRY_TYPE` would disable PSRAM unconditionally, which seems counter-intuitive.

Good catch, thanks! Moved it and added an error now.

PepperoniPingu:
> For user-facing, firmware-like deployments there are good reasons to disable interfaces

yes both since there is no need for the interfaces to be reachable and also since they might be occupied by other c-code.

AJMansfield (approved):
I can confirm that this PR builds and runs the test suite on hardware with both my Pico W and my Pico 2.

The test suite passes in the default configurations, and when disabling i2c, softi2c, i2s, and spi:

<details>

<summary>boards/RPI_PICO_W_TEST/mpconfigvariant_NOPERI.cmake</summary>

```cmake
list(APPEND MICROPY_DEF_BOARD
    MICROPY_PY_MACHINE_I2C=0
    MICROPY_PY_MACHINE_SOFTI2C=0
    MICROPY_PY_MACHINE_I2S=0
    MICROPY_PY_MACHINE_SPI=0
)
```

</details>

There's a number of test failures on both my Pico W and my Pico 2 when enabling double-precision floats, though I believe the failures here aren't the result of this PR per-se.

<details>

<summary>boards/RPI_PICO_W_TEST/mpconfigvariant_DP.cmake</summary>

```cmake
list(APPEND MICROPY_DEF_BOARD
    MICROPY_FLOAT_IMPL=MICROPY_FLOAT_IMPL_DOUBLE
)
```

</details>

<details>

<summary>Test Failures for rp2-RPI_PICO_W_TEST-DP</summary>

```diff
FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_domain_special.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_domain_special.py.exp   2025-07-03 17:26:38.870785536 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_domain_special.py.out   2025-07-03 17:26:38.870785536 -0400
@@ -1,56 +1,56 @@
 expm1(inf) = inf
 expm1(-inf) = 0.0000
-expm1(nan) = nan
+expm1(nan) = ValueError
 log2(-1.0000) = ValueError
 log2(0.0000) = ValueError
 log2(inf) = inf
-log2(-inf) = ValueError
-log2(nan) = nan
+log2(-inf) = -inf
+log2(nan) = ValueError
 log10(-1.0000) = ValueError
 log10(0.0000) = ValueError
 log10(inf) = inf
-log10(-inf) = ValueError
-log10(nan) = nan
+log10(-inf) = -inf
+log10(nan) = ValueError
 sinh(inf) = inf
 sinh(-inf) = -inf
-sinh(nan) = nan
+sinh(nan) = ValueError
 cosh(inf) = inf
 cosh(-inf) = inf
-cosh(nan) = nan
+cosh(nan) = ValueError
 tanh(inf) = 1.0000
 tanh(-inf) = -1.0000
-tanh(nan) = nan
+tanh(nan) = 1.0000
 asinh(inf) = inf
 asinh(-inf) = -inf
-asinh(nan) = nan
-acosh(-1.0000) = ValueError
+asinh(nan) = ValueError
+acosh(-1.0000) = 0.0000
 acosh(0.9000) = ValueError
 acosh(1.0000) = 0.0000
 acosh(inf) = inf
-acosh(-inf) = ValueError
-acosh(nan) = nan
+acosh(-inf) = inf
+acosh(nan) = ValueError
 atanh(-1.0000) = ValueError
 atanh(1.0000) = ValueError
-atanh(inf) = ValueError
-atanh(-inf) = ValueError
-atanh(nan) = nan
+atanh(inf) = -inf
+atanh(-inf) = -inf
+atanh(nan) = ValueError
 erf(inf) = 1.0000
 erf(-inf) = -1.0000
-erf(nan) = nan
+erf(nan) = 1.0000
 erfc(inf) = 0.0000
 erfc(-inf) = 2.0000
-erfc(nan) = nan
+erfc(nan) = 0.0000
 gamma(-2.0000) = ValueError
 gamma(-1.0000) = ValueError
 gamma(0.0000) = ValueError
 gamma(1.0000) = 1.0000
 gamma(inf) = inf
-gamma(-inf) = ValueError
-gamma(nan) = nan
+gamma(-inf) = inf
+gamma(nan) = ValueError
 lgamma(-2.0000) = ValueError
 lgamma(-1.0000) = ValueError
 lgamma(0.0000) = ValueError
 lgamma(1.0000) = 0.0000
 lgamma(inf) = inf
 lgamma(-inf) = inf
-lgamma(nan) = nan
+lgamma(nan) = ValueError

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_complex1.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_complex1.py.exp      2025-07-03 17:26:19.875498741 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_complex1.py.out      2025-07-03 17:26:19.875498741 -0400
@@ -42,7 +42,7 @@
 True True
 True True
 False
-False
+True
 1.0
 2.2361
 1
@@ -50,8 +50,8 @@
 (1.2+3j)
 0.58779 0.80902
 -0.16625 0.51167
-(nan+nanj)
-(nan+nanj)
+(inf+infj)
+(-inf-infj)
 (inf+infj)
 (-inf-infj)
 ValueError 1+2

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_inf_nan_arith.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_inf_nan_arith.py.exp 2025-07-03 17:26:37.348430015 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_inf_nan_arith.py.out 2025-07-03 17:26:37.348430015 -0400
@@ -34,9 +34,9 @@
   ** pow 0.0 0.0
   == != < <= > >= False True False False True True
 -2 nan
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf -inf -inf
+  / -0.0
+  ** pow inf inf
   == != < <= > >= False True False False False False
 -1 -2
   + - * -3 1 2
@@ -74,9 +74,9 @@
   ** pow 1.0 1.0
   == != < <= > >= False True False False True True
 -1 nan
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf -inf -inf
+  / -0.0
+  ** pow 1.0 1.0
   == != < <= > >= False True False False False False
 0 -2
   + - * -2 2 0
@@ -104,19 +104,19 @@
   ** pow 0 0
   == != < <= > >= False True True True False False
 0 inf
-  + - * inf -inf nan
+  + - * inf -inf inf
   / 0.0
   ** pow 0.0 0.0
   == != < <= > >= False True True True False False
 0 -inf
-  + - * -inf inf nan
+  + - * -inf inf -inf
   / -0.0
   ** pow inf inf
   == != < <= > >= False True False False True True
 0 nan
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf -inf inf
+  / 0.0
+  ** pow 0.0 0.0
   == != < <= > >= False True False False False False
 1 -2
   + - * -1 3 -2
@@ -154,8 +154,8 @@
   ** pow 1.0 1.0
   == != < <= > >= False True False False True True
 1 nan
-  + - * nan nan nan
-  / nan
+  + - * inf -inf inf
+  / 0.0
   ** pow 1.0 1.0
   == != < <= > >= False True False False False False
 2 -2
@@ -194,9 +194,9 @@
   ** pow 0.0 0.0
   == != < <= > >= False True False False True True
 2 nan
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf -inf inf
+  / 0.0
+  ** pow inf inf
   == != < <= > >= False True False False False False
 inf -2
   + - * inf inf -inf
@@ -209,7 +209,7 @@
   ** pow 0.0 0.0
   == != < <= > >= False True False False True True
 inf 0
-  + - * inf inf nan
+  + - * inf inf inf
   / ZeroDivisionError
   ** pow 1.0 1.0
   == != < <= > >= False True False False True True
@@ -224,19 +224,19 @@
   ** pow inf inf
   == != < <= > >= False True False False True True
 inf inf
-  + - * inf nan inf
-  / nan
+  + - * inf inf inf
+  / inf
   ** pow inf inf
   == != < <= > >= True False False True False True
 inf -inf
-  + - * nan inf -inf
-  / nan
+  + - * inf inf -inf
+  / -inf
   ** pow 0.0 0.0
   == != < <= > >= False True False False True True
 inf nan
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf inf
+  / inf
+  ** pow inf inf
   == != < <= > >= False True False False False False
 -inf -2
   + - * -inf -inf inf
@@ -249,7 +249,7 @@
   ** pow -0.0 -0.0
   == != < <= > >= False True True True False False
 -inf 0
-  + - * -inf -inf nan
+  + - * -inf -inf -inf
   / ZeroDivisionError
   ** pow 1.0 1.0
   == != < <= > >= False True True True False False
@@ -264,57 +264,57 @@
   ** pow inf inf
   == != < <= > >= False True True True False False
 -inf inf
-  + - * nan -inf -inf
-  / nan
+  + - * inf -inf -inf
+  / -inf
   ** pow inf inf
   == != < <= > >= False True True True False False
 -inf -inf
-  + - * -inf nan inf
-  / nan
+  + - * -inf inf inf
+  / inf
   ** pow 0.0 0.0
   == != < <= > >= True False False True False True
 -inf nan
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf -inf -inf
+  / -inf
+  ** pow inf inf
   == != < <= > >= False True False False False False
 nan -2
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf -inf
+  / -inf
+  ** pow 0.0 0.0
   == != < <= > >= False True False False False False
 nan -1
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf -inf
+  / -inf
+  ** pow 0.0 0.0
   == != < <= > >= False True False False False False
 nan 0
-  + - * nan nan nan
+  + - * inf inf inf
   / ZeroDivisionError
   ** pow 1.0 1.0
   == != < <= > >= False True False False False False
 nan 1
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf inf
+  / inf
+  ** pow inf inf
   == != < <= > >= False True False False False False
 nan 2
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf inf
+  / inf
+  ** pow inf inf
   == != < <= > >= False True False False False False
 nan inf
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf inf
+  / inf
+  ** pow inf inf
   == != < <= > >= False True False False False False
 nan -inf
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf -inf
+  / -inf
+  ** pow 0.0 0.0
   == != < <= > >= False True False False False False
 nan nan
-  + - * nan nan nan
-  / nan
-  ** pow nan nan
+  + - * inf inf inf
+  / inf
+  ** pow inf inf
   == != < <= > >= False True False False False False

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_float_parse_doubleprec.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_float_parse_doubleprec.py.exp        2025-07-03 17:26:36.334860357 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_float_parse_doubleprec.py.out        2025-07-03 17:26:36.334860357 -0400
@@ -1,9 +1,9 @@
-1e+300
+1.000000000000003e+300
 1e+200
 1.0
 1.0
 1e+100
-1e-100
+1.000000000000003e-100
 9.00000000000000e-301
 9.00000000000000e-201
 9.00000000000000e-01

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_fun_special.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_fun_special.py.exp      2025-07-03 17:26:39.936201093 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_fun_special.py.out      2025-07-03 17:26:39.936201093 -0400
@@ -57,7 +57,7 @@
 asinh(-2.5) = -1.647
 asinh(-1) = -0.8814
 asinh(-0.5) = -0.4812
-asinh(0) = 0
+asinh(0) = -0
 asinh(0.5) = 0.4812
 asinh(2.5) = 1.647
 asinh(8) = 2.776

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_builtin_float_round.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_builtin_float_round.py.exp   2025-07-03 17:26:18.409612217 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_builtin_float_round.py.out   2025-07-03 17:26:18.409612217 -0400
@@ -26,4 +26,4 @@
 1.5
 1.47
 <class 'OverflowError'>
-<class 'ValueError'>
+<class 'OverflowError'>

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_float_parse.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_float_parse.py.exp   2025-07-03 17:26:36.237948842 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_float_parse.py.out   2025-07-03 17:26:36.237948842 -0400
@@ -4,7 +4,7 @@
 1.0000e+20
 1.0
 1e+20
-True
+False
 True
 True
 True

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_builtin_float_pow.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_builtin_float_pow.py.exp     2025-07-03 17:26:18.321560683 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_builtin_float_pow.py.out     2025-07-03 17:26:18.321560683 -0400
@@ -5,4 +5,4 @@
 0.0625
 0.0
 inf
-nan
+0.0

FAILURE /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_domain.py
--- /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_domain.py.exp   2025-07-03 17:26:38.616541257 -0400
+++ /home/anson/mpy/micropython/tests/results/rp2-RPI_PICO_W_TEST-DP/float_math_domain.py.out   2025-07-03 17:26:38.616541257 -0400
@@ -36,8 +36,8 @@
 sqrt(-1.0000) = ValueError
 sqrt(-1.1200) = ValueError
 sqrt(inf) = inf
-sqrt(-inf) = ValueError
-sqrt(nan) = nan
+sqrt(-inf) = -inf
+sqrt(nan) = ValueError
 exp(0.0000) = 1.0000
 exp(1.0000) = 2.7183
 exp(1.1200) = 3.0649
@@ -45,7 +45,7 @@
 exp(-1.1200) = 0.3263
 exp(inf) = inf
 exp(-inf) = 0.0000
-exp(nan) = nan
+exp(nan) = ValueError
 log(0.0000) = ValueError
 log(1.0000) = 0.0000
 log(1.1200) = 0.1133
@@ -53,47 +53,47 @@
 log(-1.1200) = ValueError
 log(inf) = inf
 log(-inf) = ValueError
-log(nan) = nan
+log(nan) = inf
 sin(0.0000) = 0.0000
 sin(1.0000) = 0.8415
 sin(1.1200) = 0.9001
 sin(-1.0000) = -0.8415
 sin(-1.1200) = -0.9001
-sin(inf) = ValueError
-sin(-inf) = ValueError
-sin(nan) = nan
+sin(inf) = -0.1585
+sin(-inf) = -0.1585
+sin(nan) = -0.1585
 cos(0.0000) = 1.0000
 cos(1.0000) = 0.5403
 cos(1.1200) = 0.4357
 cos(-1.0000) = 0.5403
 cos(-1.1200) = 0.4357
-cos(inf) = ValueError
-cos(-inf) = ValueError
-cos(nan) = nan
+cos(inf) = 0.9874
+cos(-inf) = 0.9874
+cos(nan) = 0.9874
 tan(0.0000) = 0.0000
 tan(1.0000) = 1.5574
 tan(1.1200) = 2.0660
 tan(-1.0000) = -1.5574
 tan(-1.1200) = -2.0660
-tan(inf) = ValueError
-tan(-inf) = ValueError
-tan(nan) = nan
+tan(inf) = -0.1606
+tan(-inf) = -0.1606
+tan(nan) = -0.1606
 asin(0.0000) = 0.0000
 asin(1.0000) = 1.5708
 asin(1.1200) = ValueError
 asin(-1.0000) = -1.5708
 asin(-1.1200) = ValueError
-asin(inf) = ValueError
-asin(-inf) = ValueError
-asin(nan) = nan
+asin(inf) = inf
+asin(-inf) = inf
+asin(nan) = ValueError
 acos(0.0000) = 1.5708
 acos(1.0000) = 0.0000
 acos(1.1200) = ValueError
 acos(-1.0000) = 3.1416
 acos(-1.1200) = ValueError
-acos(inf) = ValueError
-acos(-inf) = ValueError
-acos(nan) = nan
+acos(inf) = inf
+acos(-inf) = inf
+acos(nan) = ValueError
 atan(0.0000) = 0.0000
 atan(1.0000) = 0.7854
 atan(1.1200) = 0.8419
@@ -101,7 +101,7 @@
 atan(-1.1200) = -0.8419
 atan(inf) = 1.5708
 atan(-inf) = -1.5708
-atan(nan) = nan
+atan(nan) = 1.5708
 ldexp(0.0000) = 0.0000
 ldexp(1.0000) = 1.0000
 ldexp(1.1200) = 1.1200
@@ -117,7 +117,7 @@
 radians(-1.1200) = -0.0195
 radians(inf) = inf
 radians(-inf) = -inf
-radians(nan) = nan
+radians(nan) = inf
 degrees(0.0000) = 0.0000
 degrees(1.0000) = 57.2958
 degrees(1.1200) = 64.1713
@@ -125,7 +125,7 @@
 degrees(-1.1200) = -64.1713
 degrees(inf) = inf
 degrees(-inf) = -inf
-degrees(nan) = nan
+degrees(nan) = inf
 pow(0.0000, 0.0000) = 1.0000
 pow(0.0000, 2.0000) = 0.0000
 pow(0.0000, -1.0000) = ValueError
@@ -154,13 +154,13 @@
 pow(inf, -inf) = 0.0000
 pow(-inf, inf) = inf
 pow(-inf, -inf) = 0.0000
-pow(0.0000, nan) = nan
+pow(0.0000, nan) = 0.0000
 pow(nan, 0.0000) = 1.0000
 pow(1.0000, nan) = 1.0000
-pow(nan, 1.0000) = nan
-pow(inf, nan) = nan
-pow(nan, inf) = nan
-pow(nan, nan) = nan
+pow(nan, 1.0000) = ValueError
+pow(inf, nan) = inf
+pow(nan, inf) = inf
+pow(nan, nan) = ValueError
 log(0.0000, 0.0000) = ValueError
 log(0.0000, 2.0000) = ValueError
 log(0.0000, -1.0000) = ValueError
@@ -185,17 +185,17 @@
 log(inf, 0.0000) = ValueError
 log(inf, 1.2000) = inf
 log(inf, -1.2000) = ValueError
-log(inf, inf) = nan
+log(inf, inf) = inf
 log(inf, -inf) = ValueError
 log(-inf, inf) = ValueError
 log(-inf, -inf) = ValueError
 log(0.0000, nan) = ValueError
 log(nan, 0.0000) = ValueError
-log(1.0000, nan) = nan
+log(1.0000, nan) = 0.0000
 log(nan, 1.0000) = ZeroDivisionError
-log(inf, nan) = nan
-log(nan, inf) = nan
-log(nan, nan) = nan
+log(inf, nan) = inf
+log(nan, inf) = inf
+log(nan, nan) = inf
 fmod(0.0000, 0.0000) = ValueError
 fmod(0.0000, 2.0000) = 0.0000
 fmod(0.0000, -1.0000) = 0.0000
@@ -217,20 +217,20 @@
 fmod(-0.9000, -inf) = -0.9000
 fmod(-1.2000, inf) = -1.2000
 fmod(-1.2000, -inf) = -1.2000
-fmod(inf, 0.0000) = ValueError
-fmod(inf, 1.2000) = ValueError
-fmod(inf, -1.2000) = ValueError
-fmod(inf, inf) = ValueError
-fmod(inf, -inf) = ValueError
-fmod(-inf, inf) = ValueError
-fmod(-inf, -inf) = ValueError
-fmod(0.0000, nan) = nan
-fmod(nan, 0.0000) = nan
-fmod(1.0000, nan) = nan
-fmod(nan, 1.0000) = nan
-fmod(inf, nan) = nan
-fmod(nan, inf) = nan
-fmod(nan, nan) = nan
+fmod(inf, 0.0000) = inf
+fmod(inf, 1.2000) = inf
+fmod(inf, -1.2000) = inf
+fmod(inf, inf) = inf
+fmod(inf, -inf) = inf
+fmod(-inf, inf) = inf
+fmod(-inf, -inf) = inf
+fmod(0.0000, nan) = 0.0000
+fmod(nan, 0.0000) = ValueError
+fmod(1.0000, nan) = 1.0000
+fmod(nan, 1.0000) = ValueError
+fmod(inf, nan) = inf
+fmod(nan, inf) = inf
+fmod(nan, nan) = ValueError
 atan2(0.0000, 0.0000) = 0.0000
 atan2(0.0000, 2.0000) = 0.0000
 atan2(0.0000, -1.0000) = 3.1416
@@ -246,11 +246,11 @@
 atan2(0.9000, -inf) = 3.1416
 atan2(1.2000, inf) = 0.0000
 atan2(1.2000, -inf) = 3.1416
-atan2(-0.5000, inf) = -0.0000
+atan2(-0.5000, inf) = 0.0000
 atan2(-0.5000, -inf) = -3.1416
-atan2(-0.9000, inf) = -0.0000
+atan2(-0.9000, inf) = 0.0000
 atan2(-0.9000, -inf) = -3.1416
-atan2(-1.2000, inf) = -0.0000
+atan2(-1.2000, inf) = 0.0000
 atan2(-1.2000, -inf) = -3.1416
 atan2(inf, 0.0000) = 1.5708
 atan2(inf, 1.2000) = 1.5708
@@ -259,13 +259,13 @@
 atan2(inf, -inf) = 2.3562
 atan2(-inf, inf) = -0.7854
 atan2(-inf, -inf) = -2.3562
-atan2(0.0000, nan) = nan
-atan2(nan, 0.0000) = nan
-atan2(1.0000, nan) = nan
-atan2(nan, 1.0000) = nan
-atan2(inf, nan) = nan
-atan2(nan, inf) = nan
-atan2(nan, nan) = nan
+atan2(0.0000, nan) = 0.0000
+atan2(nan, 0.0000) = 1.5708
+atan2(1.0000, nan) = 0.0000
+atan2(nan, 1.0000) = 1.5708
+atan2(inf, nan) = 0.7854
+atan2(nan, inf) = 0.7854
+atan2(nan, nan) = 0.7854
 copysign(0.0000, 0.0000) = 0.0000
 copysign(0.0000, 2.0000) = 0.0000
 copysign(0.0000, -1.0000) = -0.0000
make: Leaving directory '/home/anson/mpy/boards/RPI_PICO_W_TEST'
```

</details>

I also can't confirm any of the SPIRAM options, as I don't have an rp2 board with this feature.

PepperoniPingu:
Hi again!

Ran into some problems when setting `#define MICROPY_PY_MACHINE_SOFTI2C (0)` while also having `#define MICROPY_PY_MACHINE_I2C (1)`. This is because the rp2 i2c driver depends on [soft i2c for zero length writes](https://github.com/micropython/micropython/blob/master/ports/rp2/machine_i2c.c#L165).

Therefore I propose to remove the configurability that I added on `MICROPY_PY_MACHINE_SOFTI2C` and have it always set to same value as `MICROPY_PY_MACHINE_I2C`.

dpgeorge (approved):
Thanks for the contribution.  This looks fine to me.

We can leave the SOFTI2C change.  It allows a board to disable hardware I2C and then keep or also disable software I2C.

PR #17491, opened 2025-06-13
modio.BufferedWriter: Fix the case where write() fails.

jepler:
### Summary

Previously, there was no test coverage of the "write failed" path. In fact, the assertion would fire instead of gracefully raising a Python exception.

Slightly re-organize the code to place the assertion later. Add a test case, and update the expected output.

### Testing

I ran the unix coverage tests locally.

I considered writing `assert(err || (out_sz == self->len));` but thought this was preferable.

### DIFF
--- a/py/modio.c
+++ b/py/modio.c
@@ -169,12 +169,13 @@ static mp_obj_t bufwriter_flush(mp_obj_t self_in) {
         int err;
         mp_uint_t out_sz = mp_stream_write_exactly(self->stream, self->buf, self->len, &err);
         (void)out_sz;
-        // TODO: try to recover from a case of non-blocking stream, e.g. move
-        // remaining chunk to the beginning of buffer.
-        assert(out_sz == self->len);
-        self->len = 0;
         if (err != 0) {
             mp_raise_OSError(err);
+        } else {
+            // TODO: try to recover from a case of non-blocking stream, e.g. move
+            // remaining chunk to the beginning of buffer.
+            assert(out_sz == self->len);
+            self->len = 0;
         }
     }
 
--- a/tests/basics/io_buffered_writer.py
+++ b/tests/basics/io_buffered_writer.py
@@ -31,13 +31,24 @@ print(type(hash(buf)))
 
 # Test failing flush()
 class MyIO(io.IOBase):
+    def __init__(self):
+        self.count = 0
+
     def write(self, buf):
-        return None
+        self.count += 1
+        if self.count < 3:
+            return None
+        print("writing", buf)
+        return len(buf)
+
 
 buf = io.BufferedWriter(MyIO(), 8)
+
 buf.write(b"foobar")
-try:
-    buf.flush()
-    print("flushed")
-except OSError:
-    print("OSError")
+
+for _ in range(4):
+    try:
+        buf.flush()
+        print("flushed")
+    except OSError:
+        print("OSError")
--- a/tests/basics/io_buffered_writer.py
+++ b/tests/basics/io_buffered_writer.py
@@ -28,3 +28,16 @@ print(bts.getvalue())
 
 # hashing a BufferedWriter
 print(type(hash(buf)))
+
+# Test failing flush()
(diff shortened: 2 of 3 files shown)


py/modio.c
```diff
@@ -169,12 +169,13 @@ static mp_obj_t bufwriter_flush(mp_obj_t self_in) {
...
         mp_uint_t out_sz = mp_stream_write_exactly(self->stream, self->buf, self->len, &err);
         (void)out_sz;
-        // TODO: try to recover from a case of non-blocking stream, e.g. move
-        // remaining chunk to the beginning of buffer.
-        assert(out_sz == self->len);
-        self->len = 0;
         if (err != 0) {
             mp_raise_OSError(err);
+        } else {
+            // TODO: try to recover from a case of non-blocking stream, e.g. move
```
dpgeorge, line 175: Should this TODO stay where it was, because it's related to the OSError being raised in the case of a non-blocking error?

tests/basics/io_buffered_writer.py
```diff
@@ -28,3 +28,19 @@
 
 # hashing a BufferedWriter
 print(type(hash(buf)))
+
+# Test failing flush()
+class MyIO(io.IOBase):
+    def write(self, buf):
+        return None
+
+    def writable(self):
```
dpgeorge: I don't think this method is needed (this test has a .exp file so won't be run under CPython).

py/modio.c
jepler: I think the comment is referring to the case where e.g., 8 bytes were requested to write, but only 6 bytes were written. This is the case that hits the assertion error, and where moving the "remaining" bytes to the head of the buffer would be a possible way to recover. If that understanding is right, then I think the comment's in the right place.

 * err == 0, out_sz == self->len: data was written & flushed
 * err != 0: data is preserved, OSError is raised (what this PR implements)
 * err == 0, out_sz < self->len: data is partially written. assertion will fire. If assertions disabled, some data is lost.

tests/basics/io_buffered_writer.py
jepler: you're right, will change. I had toyed with *NOT* printing the value on line 17 and getting rid of the exp file; this is a remnant of that, I guess. Will remove.

py/modio.c
dpgeorge: Yes, you're right.

tests/basics/io_buffered_writer.py
```diff
@@ -28,3 +28,16 @@
...
+    def write(self, buf):
+        return None
+
+buf = io.BufferedWriter(MyIO(), 8)
+buf.write(b"foobar")
+try:
+    buf.flush()
+    print("flushed")
+except OSError:
+    print("OSError")
```
dpgeorge: I think this needs to test a little more, to ensure we have the logic correct for the case of a failed flush.

In particular we want to test that `self->len = 0` should not be executed if there's an error.  Because if we turn the assertion off then the change here changes semantics.

I suggest to loop over this try-except 4 times:
- the first two should fail with errors and print OSError
- the third should succeed and flush all 6 bytes (and `MyIO.write` should print those bytes out)
- the fourth should succeed and flush nothing, and not even call `MyIO.write`

To do this you can add some simple state to `MyIO` so it fails the first two calls to `write`, then succeeds the remaining calls.

dpgeorge:
> I considered writing `assert(err || (out_sz == self->len));` but thought this was preferable.

As I mention above, changing the assert to do that is subtly different to the change made in this PR.  The change done in this PR is better!

tests/basics/io_buffered_writer.py
jepler: That's a much more comprehensive test. Let me know if the updated version captures what you wanted.

dpgeorge (approved):
Yes, that's exactly what I was thinking.  Looking good now!

dpgeorge:
Squashed and merged in b6b7d64bd9257cc1c138d37033108dcff5ceb89e

PR #17492, opened 2025-06-13
mpprint: Improve coverage and remove dead code

jepler:
### Summary

I was looking at coverage misses in mpprint .c. This PR is the result of investigating some un-covered lines.

mpprint: Remove unused "PF_FLAG_NO_TRAILZ" flag and re-number the other flags.

Leave PF_FLAG_SEP_POS at 9 (the highest position that probably works with 16-bit integers like the pic16bit port)

coverage: Test left adjusted print.

### Testing

I ran the Unix coverage tests locally.

I looked at git history and found no indication that the NO_TRAILZ flag was ever implemented or that "%!" was used as an mp_printf format string in practice.

### DIFF
--- a/py/mpprint.h
+++ b/py/mpprint.h
@@ -32,11 +32,11 @@
 #define PF_FLAG_SHOW_SIGN         (0x002)
 #define PF_FLAG_SPACE_SIGN        (0x004)
 #define PF_FLAG_NO_TRAILZ         (0x008)
-#define PF_FLAG_SHOW_PREFIX       (0x010)
-#define PF_FLAG_PAD_AFTER_SIGN    (0x020)
-#define PF_FLAG_CENTER_ADJUST     (0x040)
-#define PF_FLAG_ADD_PERCENT       (0x080)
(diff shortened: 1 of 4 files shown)


py/mpprint.h
```diff
@@ -32,11 +32,11 @@
 #define PF_FLAG_SHOW_SIGN         (0x002)
 #define PF_FLAG_SPACE_SIGN        (0x004)
 #define PF_FLAG_NO_TRAILZ         (0x008)
```
dlech:
```suggestion
```

jepler: Fixed now. Thank you!

dpgeorge (approved):
Thanks for this, always nice to have a clean up, even a small one!

I guess technically this is a breaking change, if someone was using `!` then it previously ignored that flag but now will print something different.  But that's doesn't affect any Python API, only the C-level `mp_printf` uses.

dpgeorge:
Rebased and merged in 1a8f4b290bc19cc3981f6deb2cb10a40a7b3bbea and fc6205c4f005dcc9d0560cfe1619f022db8ff12f

PR #17493, opened 2025-06-13
tests: Improve coverage of platform module.

jepler:
### Summary

I noticed two trivial functions in the platform module were not covered in the tests.

### Testing

I ran the tests locally & updated the expected results

### DIFF
--- a/tests/extmod/platform_basic.py
+++ b/tests/extmod/platform_basic.py
@@ -1,0 +1,8 @@
+try:
+    import platform
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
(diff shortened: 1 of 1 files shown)


dpgeorge:
I think it would be better to create a new `tests/extmod/platform_basic.py` test that has these tests in it.  Then it can run on all targets, not just unix coverage.

jepler:
Good idea. Updated.

dpgeorge (approved):

dpgeorge:
Thanks for updating.  Tested on RPI_PICO2_W.  Rebased and merged in 5ade8b7058f95e77a0d62d523c8443647f176532

PR #17498, opened 2025-06-14
py/misc: Fix fallback implementation of mp_popcount.

dpgeorge:
### Summary

The fallback `mp_popcount()` was broken, missing a shift.

### Testing

Tested using gcc 7.3.1 which does not have the popcount built-in and uses this fallback version.  Without the fix, `mpy-cross -march=rv32imc -X emit=native` produces mpy files with corrupt RISC-V machine code.  With the fix, `mpy-cross` output is correct.

Note: gcc 7.3.1 can be tested using docker image `larsks/esp-open-sdk`.

This bug was found as part of Octoprobe integration: https://github.com/octoprobe/testbed_micropython/issues/27

### DIFF
--- a/py/misc.h
+++ b/py/misc.h
@@ -390,7 +390,7 @@ static inline uint32_t mp_popcount(uint32_t x) {
     x = x - ((x >> 1) & 0x55555555);
     x = (x & 0x33333333) + ((x >> 2) & 0x33333333);
     x = (x + (x >> 4)) & 0x0F0F0F0F;
-    return x * 0x01010101;
+    return (x * 0x01010101) >> 24;
 }
 #endif
 #endif


dpgeorge:
@agatti FYI.  Please can you double check this `popcount` implementation and my fix here, thanks!

jepler (approved):
I plugged this into [cbmc](https://www.cprover.org/cbmc/) and it is equivalent to a naive for-loop popcnt implementation for all 32-bit unsigned integers. The original formulation encounters a verification error.

hmaerki:
I run octoprobe tests against this PR: https://reports.octoprobe.org/github_selfhosted_testrun_76/octoprobe_summary_report.html

The error has gone!

These are the results: [tests_mpy-cross_native.txt](https://github.com/user-attachments/files/20739922/tests_mpy-cross_native.txt)

hmaerki:
I run the octoprobe tests against this PR: https://reports.octoprobe.org/github_selfhosted_testrun_76/octoprobe_summary_report.html

These are the results: [tests_mpy-cross_native.txt](https://github.com/user-attachments/files/20739949/tests_mpy-cross_native.txt)

Now all md5 equal: **The error has gone!**

agatti:
I don't have access to any of my boards/dev machines until the end of the month to see what's going on there (along with my copy of "Hacker's Delight" to double check things against), but if this PR is reported by multiple sources to fix the issue then please go ahead!

I took that code from an old project of mine running on a ColdFire board that's been working for a few years now, so now I wonder whether this works due to code consuming that value that either does the shift itself or masks it out and fills the other bits with something else - no idea until I take a look.

PR #17499, opened 2025-06-15
asmbase: Fix assertion error with viper code.

jepler:
### Summary

In the case of viper code it's possible to reach MP_ASM_PASS_EMIT with a code size of 0 bytes. Update the assertion accordingly.

After this change, `mpy-cross -march=debug' on the viper tests creates output that looks plausible to me, though I am not experienced in reading the output.

Closes: #17467

### Testing

I locally ran mpy-cross on some viper files.

As this just changes an assertion it's unlikely to affect real ports.

I briefly considered that the code size could also be forced non-zero but this would add code and doesn't appear to be necessary. Or, that the debug backend emitter could set its own length, but this does not seem to be necessary.


dpgeorge (approved):

dpgeorge:
Rebased and merged in 5ff2ae5a6c955a5123082ba441ad57fe4908b690

PR #17503, opened 2025-06-16
workflows: Run the address sanitizer (ASan) build during CI.

jepler:
### Summary

ASan, the address sanitizer can find interesting bugs using additional runtime checks. For instance, it detected the memory leak now fixed at https://github.com/micropython/berkeley-db-1.xx/pull/1 and possibly other interesting findings I've previously submitted.

This patch set enables ASan for the 64-bit unix coverage build.

### Testing

Lots and lots of builds until it was green!

I had to disable one entire class of check, "stack use after return". I wasn't able to find good information about how this mode works, but it clearly violates assumptions that micropython makes in nlr, gc, and stack space checking.


jepler:
aha https://github.com/google/sanitizers/wiki/AddressSanitizerUseAfterReturn explains how it works. I'm not sure how you'd make this mode work with all micropython's stack trickery, there was a link to research on how to make it interoperate with some other kind of gc but it's dead and not archived that I could find.

jepler:
ah, found it in their git history: llvm [introduced two functions](https://github.com/llvm/llvm-project/commit/1ee681305f1ddf2fd82e2ff5acb0d711ba23879b) to allow gc to work with the stack use sanitizer. I don't know if gcc did likewise, and I'm not sure how exactly they'd be used in practice.

jepler:
Here's a program that seems to know how to gather pointers in the fake stack for gc: https://github.com/karolba/serenity/blob/2a563b9de65886c0a5644b8ea41bef14e5587735/Userland/Libraries/LibJS/Heap/Heap.cpp#L163

jepler:
(just to be clear: I don't think this PR should be held for want of being able to use that specific sanitizer.)

dpgeorge (approved):
This is a really nice addition to the testing, thank you!

dpgeorge:
Rebased and merged in 9a5cf0e3dbb7dbc1b66ef67dd2fec0245d24e37e through 1eb27e11f3bf8e5e0ee13191ef1e9c30b00f9ead

PR #17397, opened 2025-05-30
Wake on ext1 only if supported

meirarmon:
ports/esp32: esp32.wake_on_ext1 should only be available on boards
that have SOC_PM_SUPPORT_EXT1_WAKEUP=y
docs/library/esp32.rst: update docs to reflect that wake_on_ext1
is only available for boards that have support for ext1

### Summary
Having code wrapped by definitions of whether or not the capability is supported on the
board instead of what board it is, is better practice.
I introduced this so that when using the method wake_on_ext1 on boards that don't
support it, it will be clear that it is not supported, instead of silently failing.

### Testing
I tested this on the ESP32C3

Can either be merged after https://github.com/micropython/micropython/pull/17393 and https://github.com/micropython/micropython/pull/17394, or can be merged instead of thos PRs (and close them out)

PR #17504, opened 2025-06-16
esp32/modesp32: Fix access to ext0_pin only if defined.

meirarmon:
In different functions `machine_rtc_config.ext0_pin` is accessed where SOC_PM_SUPPORT_EXT0_WAKEUP is not defined, fix that.

### Summary

Previous PRs (https://github.com/micropython/micropython/pull/17397#) wrapped defining the ext0 pins only if SOC_PM_SUPPORT_EXT0_WAKEUP, but some of the spots were missed. This PR fixes that.

### Testing

Compiled for ESP32_GENERIC_C6


dpgeorge (approved):
Thanks for the fix.

I tested that ESP32_GENERIC_C6 and UM_TINYC6 now build without error.

PR #17510, opened 2025-06-17
rp2/rp2_flash: Add default MICROPY_HW_FLASH_MAX_FREQ.

Gadgetoid:
Set a default MICROPY_HW_FLASH_MAX_FREQ when PICO_FLASH_SPI_CLKDIV is unset.

Use a divider of 4, which is the default in boot2_generic_03h.S.

As requested in https://github.com/micropython/micropython/pull/17389


dpgeorge (approved):
Thanks for getting to this quickly.  What you've done here looks very sensible.

I tested that SIL_RP2040_SHIM now builds.

PR #17511, opened 2025-06-17
alif: Add support for machine_pin IRQ.

iabdalkader:
### Summary

Adds support for machine.Pin.IRQ().

### Testing

Rising/falling edge seems to work fine. Also made sure IRQs are disabled on soft-reset.

I think `mpirq.c` needs a finaliser, this way we won't need `machine_pin_irq_deinit();`: when the object is collected, it could just disable the IRQ by calling `trigger(0)`. Also, it would be nice if `mp_irq_obj_t` could just add the `flags` and `trigger` as they seem to be used by all ports, and all ports use this weird object to wrap `mp_irq_obj_t` to add the fields.

Finally, CYW43 should use the Python API to create a pin and register an IRQ, and set it to reserved. For now I just avoided redefining the IRQ handler so it builds.

### DIFF
--- a/ports/alif/machine_pin.c
+++ b/ports/alif/machine_pin.c
@@ -30,8 +30,76 @@
 #include "extmod/virtpin.h"
 #include "shared/runtime/mpirq.h"
 
+#ifndef MACHINE_PIN_NUM_VECTORS
+#define MACHINE_PIN_NUM_PORT_IO (8)
+#define MACHINE_PIN_NUM_VECTORS (16 * MACHINE_PIN_NUM_PORT_IO)
+#endif
+
+typedef struct _machine_pin_irq_obj_t {
+    mp_irq_obj_t base;
+    uint32_t flags;
+    uint32_t trigger;
+    IRQn_Type irq_num;
+    bool reserved; // for use by other drivers
+} machine_pin_irq_obj_t;
+
+#define MACHINE_PIN_IRQ_INDEX(port, pin) \
+    ((port) * MACHINE_PIN_NUM_PORT_IO + (pin))
+
+#define MACHINE_PIN_IRQ_OBJECT(port, pin) \
+    (MP_STATE_PORT(machine_pin_irq_obj[MACHINE_PIN_IRQ_INDEX((port), (pin))]))
+
+// Defines a single GPIO IRQ handler
+#define DEFINE_GPIO_IRQ_HANDLER(pname, port, pin) \
+    void pname##_IRQ##pin##Handler(void) { \
+        machine_pin_irq_obj_t *irq = MACHINE_PIN_IRQ_OBJECT(port, pin); \
+        machine_pin_obj_t *self = MP_OBJ_TO_PTR(irq->base.parent); \
+        gpio_interrupt_eoi(self->gpio, pin); \
+        irq->flags = irq->trigger; \
+        mp_irq_handler(&irq->base); \
+    }
+
+// Defines all 8 pin IRQ handlers for a port
+#define DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(gpio, port) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 0) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 1) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 2) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 3) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 4) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 5) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 6) \
+    DEFINE_GPIO_IRQ_HANDLER(gpio, port, 7)
+
+// Generate handlers for GPIO ports 0 to 14 + LPGPIO
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO0, 0)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO1, 1)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO2, 2)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO3, 3)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO4, 4)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO5, 5)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO6, 6)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO7, 7)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO8, 8)
+
+DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 0)
+DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 1)
+DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 2)
+DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 3)
+DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 4)
+DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 5)
+// DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 6)  // Reserved for WiFi
+DEFINE_GPIO_IRQ_HANDLER(GPIO9, 9, 7)
+
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO10, 10)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO11, 11)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO12, 12)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO13, 13)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(GPIO14, 14)
+DEFINE_GPIO_IRQ_HANDLERS_FOR_PORT(LPGPIO, 15)
+
 extern const mp_obj_dict_t machine_pin_cpu_pins_locals_dict;
 extern const mp_obj_dict_t machine_pin_board_pins_locals_dict;
+static const mp_irq_methods_t machine_pin_irq_methods;
 
 static const machine_pin_obj_t *machine_pin_find_named(const mp_obj_dict_t *named_pins, mp_obj_t name) {
     const mp_map_t *named_map = &named_pins->map;
@@ -171,6 +239,7 @@ mp_obj_t mp_pin_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, 
         mp_map_init_fixed_table(&kw_args, n_kw, args + n_args);
         machine_pin_obj_init_helper(self, n_args - 1, args + 1, &kw_args);
     }
+
     return MP_OBJ_FROM_PTR(self);
 }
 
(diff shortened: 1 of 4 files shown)


dpgeorge:
Can you please show your testing code?  I tried this on ALIF_ENSEMBLE with the following code:
```py
from machine import Pin
p = Pin("JOY_LEFT", Pin.IN, Pin.PULL_UP)
p.irq(lambda x: print(x), Pin.IRQ_RISING | Pin.IRQ_FALLING)
```
It didn't do anything, it didn't run the IRQ lambda.

iabdalkader:
> Can you please show your testing code? I tried this on ALIF_ENSEMBLE with the following code:

This is my test script, maybe you need to set the trigger with the keyboard? Should I try it again maybe it's broken?

```Python
from machine import Pin
import time


int_pin = Pin("P9", Pin.IN, Pin.PULL_UP)
def int_callback(line):
    print("callback")

int_pin.irq(handler=int_callback, trigger=Pin.IRQ_FALLING, hard=False)

while True:
    time.sleep_ms(1000)
```

I think LPGPIO is not working for some reason.

dpgeorge:
OK, so it works on P1_3, but not on P15_0.

iabdalkader:
> OK, so it works on P1_3, but not on P15_0.

Yes, I think I see the issue.

iabdalkader:
Should be fixed now. You could just use `int_pin = Pin("USR_SW", Pin.IN, Pin.PULL_UP)`

ports/alif/machine_pin.c
```diff
@@ -30,8 +30,70 @@
...
+} machine_pin_irq_obj_t;
+
+// Defines a single GPIO IRQ handler
+#define DEFINE_GPIO_IRQ_HANDLER(pname, port, pin) \
+    void pname##_IRQ##pin##Handler(void) { \
+        size_t irq_idx = port * MACHINE_PIN_NUM_PORT_IO + pin; \
+        machine_pin_irq_obj_t *irq = MP_STATE_PORT(machine_pin_irq_obj[irq_idx]); \
+        machine_pin_obj_t *self = MP_OBJ_TO_PTR(irq->base.parent); \
+        gpio_interrupt_eoi(self->gpio, pin); \
+        mp_irq_handler(&irq->base); \
```
dpgeorge, line 59: Need to set `irg->flags` to the flags that triggered the IRQ, before calling `mp_irq_handler()`.  Python code should be able to retrieve the flags (rising or falling or both) using `irq.flags()`.

ports/alif/machine_pin.c
```diff
@@ -225,6 +288,125 @@ static mp_obj_t machine_pin_toggle(mp_obj_t self_in) {
...
+        irq->reserved = false;
+        MP_STATE_PORT(machine_pin_irq_obj[irq_idx]) = irq;
+    }
+
+    return irq;
+}
+
+static mp_uint_t machine_pin_irq_trigger(mp_obj_t self_in, mp_uint_t trigger) {
+    machine_pin_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    machine_pin_irq_obj_t *irq = machine_pin_get_irq(self);
```
dpgeorge: If the IRQ object doesn't exist, then this function should just return 0.  This function should never try to allocate a new object (it might be running in a hard interrupt).

So probably you want a simple `machine_pin_get_irq()` which may return NULL, and `machine_pin_get_irq_or_create()` which creates if it doesn't exist.

ports/alif/machine_pin.c
```diff
@@ -225,6 +288,125 @@ static mp_obj_t machine_pin_toggle(mp_obj_t self_in) {
...
+    NVIC_ClearPendingIRQ(irq->irq_num);
+    NVIC_SetPriority(irq->irq_num, IRQ_PRI_GPIO);
+    NVIC_EnableIRQ(irq->irq_num);
+    return 0;
+
+}
+
+static mp_uint_t machine_pin_irq_info(mp_obj_t self_in, mp_uint_t info_type) {
+    machine_pin_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    machine_pin_irq_obj_t *irq = machine_pin_get_irq(self);
```
dpgeorge: As above, should never allocate an object.

ports/alif/machine_pin.c
```diff
@@ -225,6 +288,125 @@ static mp_obj_t machine_pin_toggle(mp_obj_t self_in) {
...
+    if (n_args > 1 || kw_args->used != 0) {
+        machine_pin_irq_obj_t *irq = machine_pin_get_irq(self);
+        if (irq->reserved) {
+            mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("Pin IRQ is reserved"));
+        }
+        irq->base.handler = args[ARG_handler].u_obj;
+        irq->base.ishard = args[ARG_hard].u_bool;
+        machine_pin_irq_trigger(self, args[ARG_trigger].u_int);
+    }
+    return mp_const_none;
```
dpgeorge: This needs to return the irq object, `return MP_OBJ_FROM_PTR(irq)`.

ports/alif/machine_pin.c
```diff
@@ -225,6 +288,125 @@ static mp_obj_t machine_pin_toggle(mp_obj_t self_in) {
...
+    .trigger = machine_pin_irq_trigger,
+    .info = machine_pin_irq_info,
+};
+
+// pin.irq(handler=None, trigger=IRQ_FALLING|IRQ_RISING, hard=False)
+static mp_obj_t machine_pin_irq(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    enum { ARG_handler, ARG_trigger, ARG_hard };
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_handler, MP_ARG_OBJ, {.u_rom_obj = MP_ROM_NONE} },
+        { MP_QSTR_trigger, MP_ARG_INT, {.u_int = MP_HAL_PIN_TRIGGER_FALL | MP_HAL_PIN_TRIGGER_RISE} },
```
dpgeorge, line 371: I tried setting the trigger to rising only, and falling only, but in both those cases it seems to always trigger on both edges.

Can you test to see if you can get it triggering on just one edge?

ports/alif/machine_pin.c
iabdalkader: > This function should never try to allocate a new object (it might be running in a hard interrupt).

I don't think it's possible for this function to allocate memory when called from callback, because you'd have to have an irq object first to call any function that calls this function, and `machine_pin_get_irq` already allocates the memory. Or am I missing something? It's possible to just add a `bool alloc` but I don't think it's needed.

ports/alif/machine_pin.c
iabdalkader: Yes, it works really well if you enable:
```C
gpio_enable_debounce(self->gpio, self->pin);
```

With the switch on the AE3, if you use FALLING, the callback triggers when you press and hold once. If you use rising, the callback triggers when you press and lift. I still get some spurious IRQs but you can control the debounce clock with `set_gpio_debounce_clkdiv` (only for regular I/Os though).

ports/alif/machine_pin.c
dpgeorge: > I don't think it's possible for this function to allocate memory when called from callback, because you'd have to have an irq object first to call any function that calls this function, and `machine_pin_get_irq` already allocates the memory.

That might be true, but IMO it's hard to reason about, especially if you come to this code with fresh eyes.

I suggest making it explicit that these functions should never allocate, eg with a `bool alloc` argument would be fine.

dpgeorge:
I've tested this again but it's still not working properly:
- now only P15_0 works, and P1_3 doesn't
- it always triggers on rising and falling edges, regardless of the `trigger` value

Here's my test:
```py
import time
from machine import Pin


def callback(pin):
    print("[irq", pin, pin.irq().flags(), "]")


p_out = Pin("P0_0", Pin.OUT, value=0)
p_in = Pin("P15_0", Pin.IN, Pin.PULL_UP)

for hard in (False, True):
    for trigger in (0, Pin.IRQ_FALLING, Pin.IRQ_RISING, Pin.IRQ_FALLING | Pin.IRQ_RISING):
        print(f"TEST {hard=} {trigger=}")
        p_in.irq(callback, trigger=trigger, hard=hard)
        for _ in range(4):
            p_out(0)
            print(f"out={p_out()} in={p_in()}")
            time.sleep_ms(10)
            p_out(1)
            print(f"out={p_out()} in={p_in()}")
            time.sleep_ms(10)
```

P0_0 is the output pin that toggles.  P15_0 is the input pin and you must connect them together with a wire.  The output I see is:
```
TEST hard=False trigger=0
out=0 in=0
out=1 in=1
out=0 in=0
out=1 in=1
out=0 in=0
out=1 in=1
out=0 in=0
out=1 in=1
TEST hard=False trigger=1
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
TEST hard=False trigger=2
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
TEST hard=False trigger=3
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=0 in=0[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
TEST hard=True trigger=0
out=0 in=0
out=1 in=1
out=0 in=0
out=1 in=1
out=0 in=0
out=1 in=1
out=0 in=0
out=1 in=1
TEST hard=True trigger=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 1 ]
out=1 in=1
TEST hard=True trigger=2
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 2 ]
out=1 in=1
TEST hard=True trigger=3
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=0 in=0
[irq Pin(P15_0, mode=IN, pull=PULL_UP) 3 ]
out=1 in=1
```
You can see that it always triggers on both edges.

If P15_0 is replaced with P1_3 (and P1_3 is wired to P0_0) then the callback is never called.

I'm testing on an Alif dev kit with MCU revision B4.

ports/alif/machine_pin.c
iabdalkader: Alternatively, I could add a macro to calculate the IRQ index and move the alloc to make_new? The reason for adding this function to begin with is to not have that code repeated everywhere just in case it changes in the future (for another MCU).

dpgeorge: > I could add a macro to calculate the IRQ index and move the alloc to make_new?

Yes, that would be good.

iabdalkader:
> I've tested this again but it's still not working properly:

Yes I see it too, but I can't fix that. It seems that regardless of the edge you configure, BOTH is stuck (see `GPIO_CONFIG_REG1` bit 21):
```C
$4 = {
  GPIO_SWPORTA_DR = 0,
  GPIO_SWPORTA_DDR = 1,
  GPIO_SWPORTA_CTL = 0,
  RESERVED = {0, 0, 0, 0, 0, 0, 0, 0, 0},
  GPIO_INTEN = 8,
  GPIO_INTMASK = 0,
  GPIO_INTTYPE_LEVEL = 8,
  GPIO_INT_POLARITY = 0,
  GPIO_INTSTATUS = 0,
  GPIO_RAW_INTSTATUS = 0,
  GPIO_DEBOUNCE = 8,
  GPIO_PORTA_EOI = 0,
  GPIO_EXT_PORTA = 0,
  RESERVED1 = {0, 0, 0},
  GPIO_LS_SYNC = 0,
  RESERVED2 = 0,
  GPIO_INT_BOTHEDGE = 0,
  GPIO_VER_ID_CODE = 842085162,
  GPIO_CONFIG_REG2 = 236775,
  GPIO_CONFIG_REG1 = 4157682
}
```

ports/alif/machine_pin.c
iabdalkader: Refactored.

dpgeorge:
It looks like you fixed this?  The test I wrote above is now working.

iabdalkader:
> It looks like you fixed this? The test I wrote above is now working.

Some bits get stuck and should be cleared before reconfiguring the trigger, so I cleared them, but `GPIO_CONFIG_REG1` is still wrong and I don't get any IRQs when using P1 and P2.

iabdalkader:
The wire I used to short the pins for testing didn't make good contact. Never mind it's working now.

dpgeorge:
Testing again with my test above, it works fine with P15_0 as the input (P0_0 is output), but does not work at all if P1_3 is the input.  I can see that the logic value read on input pin P1_3 is the same as what P0_0 is outputting, but there's no IRQ occurring.

iabdalkader:
> but does not work at all if P1_3 is the input

Is it used for something else on the devkit? Note that some pins were 1.8v on our board. Anyway, I don't have access to that pin, maybe @kwagyeman can test.

dpgeorge:
I tested P1_2 and P0_5 as inputs as well, and neither of them work.

kwagyeman:
P1_3 is P9 on the AE3. This definitely works with commits to our micropython fork. Tested with the OLED shield:

https://github.com/openmv/micropython/commit/f71dd03dfad000996750631147c6b5cd2e0254f1

dpgeorge:
Testing this PR on OPENMV_AE3 with the following pairs of pins connected:
- P8/P9 = P1_2/P1_3
- P6/P7 = P7_2/P7_3

Neither pair works (in either order, eg P8 being input or P9).

iabdalkader:
> Testing this PR on OPENMV_AE3 with the following pairs of pins connected:

Does it work without debounce? That's one of the changes added to this PR. Note that I'm waiting for this to fix and sync `machine_pin.c` in our fork, it's actually outdated and broken.

dpgeorge:
> Does it work without debounce?

No, still doesn't work on OPENMV_AE3 with pins P6/P7.

iabdalkader:
> > Does it work without debounce?
> 
> No, still doesn't work on OPENMV_AE3 with pins P6/P7.

It is the debounce, it needs a debounce clock enabled per port first otherwise IRQs don't work. It's not enough to comment it out in this case, you must also power cycle the board to clear/reset the debounce setting. I don't think it's worth the the trouble, is it? I mean it didn't actually fix the issue, clearing the stale regs did, also doesn't work for LPGPIO. We could still support it, but should probably add an arg to `make_new` to do so... FWIW this is how you enable it:
```C
typedef size_t GPIO_INSTANCE;
#include "sys_ctrl_gpio.h"

enable_gpio_debounce_clk(self->port);
// May also need to configure a div here
set_gpio_debounce_clkdiv(1, self->port);
gpio_enable_debounce(self->gpio, self->pin);
```

dpgeorge:
OK, I can confirm that it now works with the latest PR here, which has the debounce line removed (and I did a full power cycle, that's what I missed last time).

I tested both OPENMV_AE3 with P7_2/P7_3, and ALIF_ENSEMBLE with P0_0/P0_1 and P0_0/P15_0.  I tested those pairs in both directions OUT/IN and IN/OUT.  The IRQ works in all cases.

iabdalkader:
I have nothing else to add to this or the SPI PR, so feel free to merge when ready.

dpgeorge (approved):

PR #17513, opened 2025-06-17
alif/machine_spi: Improve transfer function.

iabdalkader:
### Summary

Improve SPI transfer function to poll MP events and time out if blocking transfer fails.

### Testing

Tested with OLED display, works fine. This follows the driver's SPI transfer function.

It is possible to optimize the FIFO transfer further, but I chose to keep it simple.

### DIFF
--- a/ports/alif/machine_spi.c
+++ b/ports/alif/machine_spi.c
@@ -39,6 +39,7 @@ typedef struct _machine_spi_obj_t {
     uint8_t id;
     SPI_Type *inst;
     bool is_lp;
+    uint32_t bits;
 } machine_spi_obj_t;
 
 static machine_spi_obj_t machine_spi_obj[] = {
@@ -246,6 +247,9 @@ mp_obj_t machine_spi_make_new(const mp_obj_type_t *type, size_t n_args, size_t n
     // Get static peripheral object.
     machine_spi_obj_t *self = &machine_spi_obj[spi_id];
 
+    // Set args
+    self->bits = args[ARG_bits].u_int;
+
     // here we would check the sck/mosi/miso pins and configure them, but it's not implemented
     if (args[ARG_sck].u_obj != MP_OBJ_NULL ||
         args[ARG_mosi].u_obj != MP_OBJ_NULL ||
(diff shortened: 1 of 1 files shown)


kwagyeman:
@iabdalkader - All good. Works for read/write.

dpgeorge:
Thanks, the implementation looks fine, and the existing test `tests/extmod/machine_spi_rate.py` still passes.

But I'm not sure about having a `timeout` argument in the SPI construction.  No other ports have this, and it's not documented (so documentation should be added here).

The problem is, the timeout depends on the transfer size.  You can see in stm32 that the SPI timeout is calculated depending on the length (and it's not user configurable).  Maybe for API simplicity the alif port can do the same thing, just have a fixed timeout that depends on the input transfer length?

(Note that I2C and UART do have timeouts that can be set by the user, but they are not related to the transfer size, rather to the character size.  Also note that stm32 spi send/recv has a timeout so it can be specified per transaction.)

iabdalkader:
Yes, no problem we can remove it. It's just so SPI doesn't get stuck forever. How do I calculate the timeout based on transfer size? Example? Like 100uS * len for example?

dpgeorge:
> Like 100uS * len for example?

Yes, something simple like that would be OK, that's how stm32 does it.  Could also take into account the frequency, then it would be something like:
```
predicted_time_us = (1000000 * len * 8 +freq - 1) / freq;
timeout_us = 2 * predicted_time_us + 100;
```
Just need to think about integer overflow in the above for large `len` and small `freq`.  Maybe better to calculate it in milliseconds.

iabdalkader:
> Just need to think about integer overflow in the above for large `len` and small `freq`. Maybe better to calculate it in milliseconds.

We could use:
```C
    // Estimate timeout in milliseconds
    uint32_t timeout = 2 * (((len * 8 * 1000) + freq - 1) / freq) + 1;
```
But actually the timeout here is intended to just make sure the polling times out eventually, if for any reason SPI gets stuck, that's why it starts over every time poll_flag is called. So I replaced this with a fixed 100ms.

ports/alif/machine_spi.c
```diff
@@ -294,22 +299,50 @@ static void machine_spi_deinit(mp_obj_base_t *self_in) {
     }
 }
 
+static void machine_spi_poll_flag(SPI_Type *spi, uint32_t flag, uint32_t timeout) {
+    mp_uint_t tick_start = mp_hal_ticks_ms();
+    while (!(spi->SPI_SR & flag)) {
+        if (mp_hal_ticks_ms() - tick_start >= timeout) {
+            mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("SPI transfer timeout"));
```
dpgeorge: Better if this is an OSError: `mp_raise_OSError(MP_ETIMEDOUT)`.

dpgeorge:
> But actually the timeout here is intended to just make sure the polling times out eventually, if for any reason SPI gets stuck, that's why it starts over every time poll_flag is called. So I replaced this with a fixed 100ms.

OK, that's a good solution.  Nice and simple.

ports/alif/machine_spi.c
iabdalkader: Done.

dpgeorge (approved):

dpgeorge:
Side note: I think this also fixes the `bits` parameter which probably never worked (was always 8 bits) because `spi_xfer.frame_size` was never set :grimacing:

iabdalkader:
> Side note: I think this also fixes the `bits` parameter which probably never worked (was always 8 bits) because `spi_xfer.frame_size` was never set

Yes it fixes that too.

PR #17388, opened 2025-05-29
Create board folder for pybstick26-esp32c3 board.

garatronic:
### Summary
esp32c3 based board at pybstick26 form factor from Garatronic.
based on lolin c3 board, with extra feature : rgb leds, two dual role buttons.

### Testing

Test, compilated with esp-idf 4.5.1

     If you leave this empty then your Pull Request may be closed. -->

PR #17514, opened 2025-06-17
Create garatronic pybstick26-esp32c3

garatronic:
### Summary
Add Garatronic pybstick26-esp32c3 board to micropython deposit

### Testing

Including #17388 issues and correct them.


ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/modules/pybstick26_esp32c3.py
```diff
@@ -0,0 +1,29 @@
+# PYBSTICK26-ESP32C3 MicroPython Helper Library
```
dpgeorge: I think it would be better to call this file `board.py` to make it a bit easier to import and use.  Eg:
```py
import board
board.led.on()
```

ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/modules/pybstick26_esp32c3.py
```diff
@@ -0,0 +1,29 @@
...
+
+from micropython import const
+from machine import Pin
+
+# Pin Assignments
+
+# SPI
+SPI_MOSI = const(7)
+SPI_MISO = const(2)
+SPI_CLK = const(6)
```
dpgeorge: You shouldn't really need to add any of these pin aliases.  Instead use a `pins.csv` file (see eg `ports/esp32/boards/UM_TINYC6/pins.csv`).  You could do the following:
```csv
SPI_MOSI,GPIO7
SPI_MISO,GPIO2
SPI_CLK,GPIO6

I2C_SDA,GPIO0
I2C_SCL,GPIO1

LED,GPIO8

BUTTON_A,GPIO9
BUTTON_B,GPIO5
```
You can also add other things like `IO6,GPIO6`.  Generally you want to add all aliases for pins that are labelled on the silk screen, and use the silk screen name for the alias.  That way someone looking at the board can know which name to use for each pin.

Then you could do:
```py
led = Pin("LED", Pin.OUT, value=0)
button = Pin("BUTTON_B", Pin.IN, Pin.PULL_UP)
```

dpgeorge (commented):
You don't need to close the PR and create a new one.  You can just add commits to this one that address the review.

ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/modules/pybstick26_esp32c3.py
garatronic: Yes. Of course. I take the wrong example with LOLIN_C3_MINI and follow it as a mindless.
Thanks for your help.

garatronic:
Hope this will be OK now. Works fine locally. Thanks

dpgeorge:
> Hope this will be OK now.

It looks like you didn't commit all of your changes?  You've only added `pins.csv` with the latest commit.

garatronic:
I double check local and github deposit. All seems to be OK. I also pull image of the board in micropython-media.

dpgeorge:
> I double check local and github deposit. All seems to be OK

You still have the file named `pybstick26_esp32c3.py` file here, and it contains the pin definitions that are no longer needed.

You might need to do `git add `modules/board.py` to add the new file?  Or `git mv modules/pybstick26_esp32c3.py modules/board.py` to rename it via git.

garatronic:
Many sorry. This should be OK this time.

dpgeorge:
Did you intend to keep the `board.py` file?

If not, you can remove the custom `manifest.py` lines, because it's now just using the default one.

garatronic:
No, i don't want to keep it.
I think (and hope) have already removed the relevant line in `manifest.py` file

ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/manifest.py
```diff
@@ -0,0 +1 @@
+include("$(PORT_DIR)/boards/manifest.py")
```
dpgeorge: this file can be removed, it uses the default

ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/mpconfigboard.cmake
```diff
@@ -0,0 +1,9 @@
+set(IDF_TARGET esp32c3)
+
+set(SDKCONFIG_DEFAULTS
+    boards/sdkconfig.base
+    boards/sdkconfig.ble
+    boards/GARATRONIC_PYBSTICK26_ESP32C3/sdkconfig.board
+)
+
+set(MICROPY_FROZEN_MANIFEST ${MICROPY_BOARD_DIR}/manifest.py)
```
dpgeorge: this line can be removed

garatronic:
Sorry to borrow you. Do you see anything else.

ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/mpconfigboard.cmake
dpgeorge: this line still needs to be removed

garatronic:
It would be long to explain but ... line 9 of `mpconfigboard.cmake` is removed.

ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/board.json
```diff
@@ -0,0 +1,23 @@
...
+    },
+    "docs": "",
+    "features": [
+        "BLE",
+        "External Flash",
+	"RGB LED",
+        "WiFi"
+    ],
+    "images": [
+        "GAR-PYBSTICK26-C3-mUSB-00.JPG"
```
dpgeorge, line 16: I see there are two images submitted at https://github.com/micropython/micropython-media/pull/96

It's best to have just one image (only one is shown on the download page).  Which image do you prefer to have?  I see you selected the one with the coin for size reference, but having the coin there will not match any of the other board images.

So I suggest either using the first image (with the two USB variants of the boards), or use the second image but remove the coin.

dpgeorge:
> It would be long to explain but ... line 9 of `mpconfigboard.cmake` is removed.

The configuration files are looking good now!

The only thing left is to work out which image to use, see above.

ports/esp32/boards/GARATRONIC_PYBSTICK26_ESP32C3/board.json
garatronic: Ok, it seems to be clear for me. Try to do it with ... two push only :)

garatronic: It should be ok now. Did only changes in micropython-media. Thanks again for your time

dpgeorge (approved):

dpgeorge:
Rebased, squashed and merged in b4b7c0a0033b128b488dae99251e7492fbb24008

PR #17517, opened 2025-06-17
tools/mpremote: Feature add control for rts / dts running in linux

poWer4aiX:
### Summary

`mpremote` is not working on my laptop running Ubuntu 24.04.2 LTS when trying to connect to an `ESP32-CAM` with the `CAM-MB`. This board has DTR connected to RST and requires DTR to be LOW for the ESP to be running. By adding the new options `--dtr [0|1|off|on]` and `--rts [0|1|off|on]` you can steer the logical level of the two lines. 

In addition, many drivers seem to pull the signals up either when the device is closed or opened. This causes the ESP32 to perform a hard reset, which can take more than 1s. A special loop has been added if entering raw REPL mode is required.

The PR also solves #11451.

### Testing
For the test I'm using the modified code since some weeks with an ESP32-CAM and MB, ESP32 NodeMCU Dev Kit C V2 and ESP8266 NodeMCU with ESP-12F module. I used it to copy files, install modules and with the REPL, or cause.

When running without providing one of the new options, the code runs as before.

### DIFF
--- a/tools/mpremote/README.md
+++ b/tools/mpremote/README.md
@@ -85,3 +85,15 @@ Examples:
     mpremote mip install aioble
     mpremote mip install github:org/repo@branch
     mpremote mip install gitlab:org/repo@branch
+
+Note for ESP-32 CAM with MB
+
+When using mpremote with the ESP-32 CAM attached to the ESP-32 CAM MB you might
+get into the situation that the communication is not working (i.e. you might
+get a timeout). To avoid this try using the following options:
+
+    mpremote connect --dtr 0 --rts 0 auto
+    mpremote connect --dtr 0 --rts 0 auto fs ls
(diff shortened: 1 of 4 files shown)


poWer4aiX:
FYI, I closed this one as I did not knew how to solve the pre-commit errors... I have opened a new one (#17532) after cleaning up my local fork...

robert-hh:
You can always fix the commit errors locally and the force-push to the existing branch on your fork. The PR will then updazed.

poWer4aiX:
> You can always fix the commit errors locally and the force-push to the existing branch on your fork. The PR will then updazed.

Good to know; this is something I have never done till now. I will give it a try the next time (hopefully not required :-D).

PR #15498, opened 2024-07-19
py/port/esp32 support on esp32c3 for deepsleep wakeup with gpio pins

m-cas:
Adding support on esp32c3 port for deepsleep wakeup with gpio pins and retriving the pin(s) that triggered the wakeup

### Summary

The current support for the ESP32C3 platform contains errors preventing the usage of deep sleep wakeup with pins. Two older PRs (#13333 and #7990) were created to fix this using gpio pins and to add the necessary calls to retrieve the pin(s) that triggered the wake. I have refactored the code, updated to the latest master (e.g. STATIC -> static) and tested.

### Testing

I created a new version of the firmware for a project of mine and it has been working since months.

### Trade-offs and Alternatives

No tradeoffs that I am aware of.

PR #13333, opened 2024-01-03
Implemented GPIO wake-up for ESP32-C3

puppet13th:
updated #9583 to current codebase.

PR #17518, opened 2025-06-17
esp32: Add esp32.wake_on_gpio.

meirarmon:
### Summary

Some boards support waking up from deepsleep via GPIO pins (for instance ESP32C3, ESP32C6), but this is not currently supported by MicroPython. This commit adds support for waking with GPIO in a similar interface to waking with ext0, ext1, touch and ulp. This commit adds documentation for this new function as well.

This is a long standing request, with a lot of chatter online. Over the years there have been several attempts to add this (https://github.com/micropython/micropython/pull/15498, https://github.com/micropython/micropython/pull/13333 and https://github.com/micropython/micropython/pull/9583), but with no success. I also explain below why my approach is better (it is based on the existing PRs, so thanks to @m-cas, @puppet13th and @ondiiik).

### Testing

Tested ESP32_GENERIC_C3with:

```
import machine
import esp32
esp32.wake_on_gpio((machine.Pin(5),machine.Pin(2)), esp32.WAKEUP_ANY_HIGH)
machine.deepsleep()
```

and checked that the relevant pins wake up from deepsleep.

The existing PRs mentioned above modify `wake_on_ex1`, testing if the port is specifically ESP32C3. My solution adds a `wake_on_gpio` which is a similar interface, and tests for the SOC capabilities.

### DIFF
--- a/docs/library/esp32.rst
+++ b/docs/library/esp32.rst
@@ -43,6 +43,15 @@ Functions
 
     .. note:: This is only available for boards that have ext1 support.
 
+.. function:: wake_on_gpio(pins, level)
+
+    Configure how GPIO wakes the device from sleep.  *pins* can be ``None``
+    or a tuple/list of valid Pin objects.  *level* should be ``esp32.WAKEUP_ALL_LOW``
+    or ``esp32.WAKEUP_ANY_HIGH``.
+
+    .. note:: Some boards don't support waking on GPIO from deep sleep,
+       on those boards, the pins set here can only be used to wake from light sleep.
+
 .. function:: gpio_deep_sleep_hold(enable)
 
     Configure whether non-RTC GPIO pin configuration is retained during
--- a/ports/esp32/machine_rtc.c
+++ b/ports/esp32/machine_rtc.c
@@ -86,8 +86,9 @@ machine_rtc_config_t machine_rtc_config = {
     .ext1_pins = 0,
     #endif
     #if SOC_PM_SUPPORT_EXT0_WAKEUP
-    .ext0_pin = -1
+    .ext0_pin = -1,
     #endif
+    .gpio_pins = 0,
 };
 
 static mp_obj_t machine_rtc_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
--- a/ports/esp32/machine_rtc.h
+++ b/ports/esp32/machine_rtc.h
@@ -37,6 +37,7 @@ typedef struct {
     #if SOC_PM_SUPPORT_EXT0_WAKEUP
     int8_t ext0_pin;   // just the pin#, -1 == None
     #endif
+    uint64_t gpio_pins; // set bit == pin#
     #if SOC_TOUCH_SENSOR_SUPPORTED
     bool wake_on_touch : 1;
     #endif
@@ -50,6 +51,7 @@ typedef struct {
     #if SOC_PM_SUPPORT_EXT1_WAKEUP
     bool ext1_level : 1;
     #endif
+    bool gpio_level : 1;
 } machine_rtc_config_t;
 
 extern machine_rtc_config_t machine_rtc_config;
--- a/ports/esp32/modesp32.c
+++ b/ports/esp32/modesp32.c
@@ -46,12 +46,13 @@
 #define MULTI_HEAP_FREERTOS
 #include "../multi_heap_platform.h"
 #include "../heap_private.h"
+#include "driver/rtc_io.h"
 
 #if SOC_TOUCH_SENSOR_SUPPORTED
 static mp_obj_t esp32_wake_on_touch(const mp_obj_t wake) {
 
     #if SOC_PM_SUPPORT_EXT0_WAKEUP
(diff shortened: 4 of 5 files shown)


meirarmon:
@dpgeorge as specified in the description, this is a feature request that is very long standing (several years at least). I think that my implementation should satisify your standards, and if it does, I would appreciate serious consideration and merging this for the next release. Thank you for maintaining MicroPython and all of your hard work :-)

meirarmon:
@dpgeorge @jimmo — any blockers left? Happy to iterate or split this into smaller PRs if the size is an issue. Thanks for reviewing!

meirarmon:
@dpgeorge is there any way to oush this? Is there anything that I can do?

VikingMultimedia:
Would love to see this merged! I'm stuck with a bunch of C3 and C6 boards laying around that would be significantly more useful with this implemented.

Zil0:
Same here, been running this PR on some C3 for a while, and it works very well.

meirarmon:
@jimmo maybe you can help push this? As you can see, many people will benefit by merging this

kastaneda:
Just tested the patch on my ESP32-C3. Works as expected! Many thanks!

My board is small, labelled "ESP32-C3 Super Mini" on back side, no vendor information, and those boards are sold widely on marketplaces. There are 16 legs, and no one of them are connected to the RST, so that feature is really, really needful.

Hopefully this PR will be in the next release!

Josverl:
Needed to rebase on top of master in order to build with a recent IDF, but indeed allows to wakeup from multiple Pins fine.

**Testing :** 
- Used `deepsleep()` and `deepsleep(20_000)` + Manual wakeup tests on multiple GPIOs
- Ran full tests suite using both UART and USB-CDC interfaces.
  * tests/extmod/machine_i2s_rate.py needed adjusting the pin numbers used as the C3 exposes less pins.
  I have opened a separate [PR](https://github.com/micropython/micropython/pull/18386) for that.


> [!NOTE]
> AFAIKT None of the existing `/tests` actually test `machine.deepsleep()` or `machine.lightsleep()` 
> It would be a good addition to the PR to at least add a test for `machine.lightsleep(10)`  or similar

jkorte-dev:
Rebase with master needed otherwise it will compile but linking fails.
 
I did manual testing with esp32c3 and wake up pin 2

``esp32.wake_on_gpio((Pin(2),), esp32.WAKEUP_ALL_LOW)``
and 
``esp32.wake_on_gpio((Pin(2),), esp32.WAKEUP_ANY_HIGH)``

Is works fine. Thanks a lot!!! You made me happy.

Some remarks:

- It should be documented, that it works only with gpio pins of the VDD3P3_RTC domain (GPIO 0-5). 
- If you choose a pin outside the range (0-5) it fails where entering deepsleep (`machine.deepsleep()`). I would expect that the exception occurs when setting wake_on_gpio(), but the exception message is clear: "E (66622) sleep: gpio 8 is an invalid deep sleep wakeup IO"
- You always need to specify a tuple even with only one pin. It's ok, but maybe there is a better way to allow setting only one pin.

I did the same tests with esp32c6 and it also works. Valid wake up pins are: GPIO 0-7

dpgeorge (approved):
Thanks for adding this feature, and sorry it took a while to get to.

It's a very straightforward extension of the existing `wake_on_XXX` functions, so it fits in nicely with them (from an API point of view).

dpgeorge:
> If you choose a pin outside the range (0-5) it fails where entering deepsleep (`machine.deepsleep()`). I would expect that the exception occurs when setting wake_on_gpio()

I don't think it's possible to validate the pins in `wake_on_gpio()` because the validation criteria is different depending on whether `deepsleep()` or `lightsleep()` is used.


> You always need to specify a tuple even with only one pin. It's ok, but maybe there is a better way to allow setting only one pin.

That's also how the existing `wake_on_ext1()` works, you must specify a tuple.  That could probably be improved to allow a single pin, or a tuple of pins.  But let's leave that for a follow up PR, to improve both those methods (if someone wants to do it).

meirarmon:
Thank you!
I will be happy to get to the enhancements that you mention, as soon as I get a chance.
Maybe [this](https://github.com/micropython/micropython/pull/17542) enhancement could get merged as well

meirarmon:
> > If you choose a pin outside the range (0-5) it fails where entering deepsleep (`machine.deepsleep()`). I would expect that the exception occurs when setting wake_on_gpio()
> 
> I don't think it's possible to validate the pins in `wake_on_gpio()` because the validation criteria is different depending on whether `deepsleep()` or `lightsleep()` is used.
> 
> > You always need to specify a tuple even with only one pin. It's ok, but maybe there is a better way to allow setting only one pin.
> 
> That's also how the existing `wake_on_ext1()` works, you must specify a tuple. That could probably be improved to allow a single pin, or a tuple of pins. But let's leave that for a follow up PR, to improve both those methods (if someone wants to do it).

@dpgeorge I did the API enhancement in [this PR](https://github.com/micropython/micropython/pull/18522).

@jkorte-dev can you please try this branch out and see that it works for you as well?

Issue #17509, opened 2025-06-17
RP2350: PIO2 State Machines do not work


IntergalacticMicrosystems:
### Port, board and/or hardware

ports-rp2 Pico 2

### MicroPython version

MicroPython v1.26.0-preview.251.g5ade8b705 on 2025-06-16; Raspberry Pi Pico2 with RP2350

### Reproduction

pio_1hz.py from examples does not blink the LED or trigger the interrupt when switched to StateMachine IDs 8, 9, 10, or 11.

### Observed behaviour

What I've found so far:
PIO2_BASE + SM0_EXECCTRL is set incorrectly
PIO2 instruction memory is not loaded
Instruction start address is not set

### Additional Information

This code mostly works.
The IRQ handler only triggers once, not repeatedly as expected.

```
import time
from machine import Pin
import rp2

@rp2.asm_pio(set_init=rp2.PIO.OUT_LOW)
def blink_1hz():
    # fmt: off
    # Cycles: 1 + 1 + 6 + 32 * (30 + 1) = 1000
    irq(rel(0))
    set(pins, 1)
    set(x, 31)                  [5]
    label("delay_high")
    nop()                       [29]
    jmp(x_dec, "delay_high")

    # Cycles: 1 + 1 + 6 + 32 * (30 + 1) = 1000
    nop()
    set(pins, 0)
    set(x, 31)                  [5]
    label("delay_low")
    nop()                       [29]
    jmp(x_dec, "delay_low")
    # fmt: on


# Create the StateMachine with the blink_1hz program, outputting on Pin(25).
sm = rp2.StateMachine(8, blink_1hz, freq=2300, set_base=Pin(25))

# Set the IRQ handler to print the millisecond timestamp.
sm.irq(lambda p: print(time.ticks_ms()))

# -------- MANUAL REGISTER SETTING START -----------
import machine
PIO2_BASE = 0x50400000
SM0_EXECCTRL = 0xCC
INSTR_MEM0 = 0x48
SM0_INSTR = 0xD8
machine.mem32[PIO2_BASE + SM0_EXECCTRL] = 129920
instrs = [49168, 57345, 58687, 48450, 89, 41026, 57344, 58687, 48450, 94]
instr_start = 0x48 + (4 * (32-len(instrs)))
for idx,instr in enumerate(instrs):    
    machine.mem32[PIO2_BASE + instr_start + (idx*4)] = instr
[...]

PR #17519, opened 2025-06-18
rp2/pio: Fix the RP2350 (PICO2) Port PIO2 prog data structure

IntergalacticMicrosystems:
### Summary

The RP2350 PIO2 State Machines ( 8, 9, 10, 11 ) did not work.
The data structure used to pass the PIO arguments was missing an entry for PIO2, thus causing the PIO2 instances to write wrong data to wrong locations.
Fixes issue [https://github.com/micropython/micropython/issues/17509](https://github.com/micropython/micropython/issues/17509)

### Testing

RP2530 Pico 2 board - Tested a modified pio_1hz.py, now works on all 12 RP2350 state machines at once or 1 at a time.
RP2040 Pico board - still works, tested a modified pio_1hz.py, works on all 8 state machines at once or 1 at a time.


jepler:
What happens if an attempt to use PIO2 is made on an RP2040?

IntergalacticMicrosystems:
> What happens if an attempt to use PIO2 is made on an RP2040?
For PIO ID > 1:  ValueError: invalid PIO
StateMachine ID > 7:  ValueError: invalid StateMachine

dpgeorge (approved):
Thanks for the fix, this looks correct.

I tested on RPI_PICO_W and RPI_PICO2_W, using `StateMachine` with various arguments.  It works as expected now.

robert-hh:
@IntergalacticMicrosystems Thanks.

PR #17531, opened 2025-06-19
py/obj: Fix nan handling in REPR_C and REPR_D.

yoctopuce:
### Summary

CPython `math.nan` is positive with regards to `math.copysign()`.
In MicroPython, this is only true with REPR_A and REPR_B.

In addition, REPR_C and REPR_D (nanbox) should only use the _true_ `nan` to prevent system crash in case of hand-crafted floats.

This problem has been discovered while working on a test code for https://github.com/micropython/micropython/pull/17444
Fixing the issue without changing the constant used for `math.nan` did break an existing `copysign` test case.

### Testing

A new test case has been added to specific test the behaviour of `math` constants with regard to `copysign`.
Test cases creating special nans that were improperly interpreted in REPR_C have also been added.

Fixing `nan` signaling flag does not change code size.

The extra check in  `mp_obj_new_float` will increase code size, but 
1. `nanbox` platform is probably not very sensitive to larger code as it uses 64 bit objects. As the added code prevents a crash, it is probably worth the small size increase. 
2. The impact on REPR_C appears to be quite small (as tested on ARMv4), and is actually more than compensated by removing the unnecessary bit-clearing code in the opposite function (`mp_obj_float_get`). Indeed, leaving bit 1 provides a better estimation of the original float value than always biasing the result toward zero by clearing both low bits.

### DIFF
--- a/py/obj.h
+++ b/py/obj.h
@@ -184,13 +184,15 @@ static inline bool mp_obj_is_small_int(mp_const_obj_t o) {
 #define MP_OBJ_NEW_SMALL_INT(small_int) ((mp_obj_t)((((mp_uint_t)(small_int)) << 1) | 1))
 
 #if MICROPY_PY_BUILTINS_FLOAT
-#define MP_OBJ_NEW_CONST_FLOAT(f) MP_ROM_PTR((mp_obj_t)((((((uint64_t)f) & ~3) | 2) + 0x80800000) & 0xffffffff))
+#include <math.h>
+// note: MP_OBJ_NEW_CONST_FLOAT should be a MP_ROM_PTR but that macro isn't available yet
+#define MP_OBJ_NEW_CONST_FLOAT(f) ((mp_obj_t)((((((uint64_t)f) & ~3) | 2) + 0x80800000) & 0xffffffff))
 #define mp_const_float_e  MP_OBJ_NEW_CONST_FLOAT(0x402df854)
 #define mp_const_float_pi MP_OBJ_NEW_CONST_FLOAT(0x40490fdb)
+#define mp_const_float_nan MP_OBJ_NEW_CONST_FLOAT(0x7fc00000)
 #if MICROPY_PY_MATH_CONSTANTS
 #define mp_const_float_tau MP_OBJ_NEW_CONST_FLOAT(0x40c90fdb)
 #define mp_const_float_inf MP_OBJ_NEW_CONST_FLOAT(0x7f800000)
-#define mp_const_float_nan MP_OBJ_NEW_CONST_FLOAT(0xffc00000)
 #endif
 
 static inline bool mp_obj_is_float(mp_const_obj_t o) {
@@ -207,6 +209,10 @@ static inline mp_float_t mp_obj_float_get(mp_const_obj_t o) {
     return num.f;
 }
 static inline mp_obj_t mp_obj_new_float(mp_float_t f) {
+    if (isnan(f)) {
+        // prevent creation of bad nanboxed pointers via array.array or struct
+        return mp_const_float_nan;
+    }
     union {
         mp_float_t f;
         mp_uint_t u;
@@ -257,12 +263,13 @@ static inline bool mp_obj_is_immediate_obj(mp_const_obj_t o) {
 #error MICROPY_OBJ_REPR_D requires MICROPY_FLOAT_IMPL_DOUBLE
 #endif
 
+#include <math.h>
 #define mp_const_float_e {((mp_obj_t)((uint64_t)0x4005bf0a8b145769 + 0x8004000000000000))}
 #define mp_const_float_pi {((mp_obj_t)((uint64_t)0x400921fb54442d18 + 0x8004000000000000))}
+#define mp_const_float_nan {((mp_obj_t)((uint64_t)0x7ff8000000000000 + 0x8004000000000000))}
 #if MICROPY_PY_MATH_CONSTANTS
 #define mp_const_float_tau {((mp_obj_t)((uint64_t)0x401921fb54442d18 + 0x8004000000000000))}
 #define mp_const_float_inf {((mp_obj_t)((uint64_t)0x7ff0000000000000 + 0x8004000000000000))}
-#define mp_const_float_nan {((mp_obj_t)((uint64_t)0xfff8000000000000 + 0x8004000000000000))}
 #endif
 
 static inline bool mp_obj_is_float(mp_const_obj_t o) {
@@ -276,6 +283,13 @@ static inline mp_float_t mp_obj_float_get(mp_const_obj_t o) {
     return num.f;
 }
 static inline mp_obj_t mp_obj_new_float(mp_float_t f) {
+    if (isnan(f)) {
+        // prevent creation of bad nanboxed pointers via array.array or struct
+        struct {
+            uint64_t r;
+        } num = mp_const_float_nan;
+        return num.r;
+    }
     union {
         mp_float_t f;
         uint64_t r;
--- a/tests/float/float_array.py
+++ b/tests/float/float_array.py
@@ -19,4 +19,10 @@ def test(a):
 test(array("f"))
 test(array("d"))
 
-print("{:.4f}".format(array("f", bytes(array("I", [0x3DCCCCCC])))[0]))
+# hand-crafted floats, including non-standard nan
+for float_hex in (0x3DCCCCCC, 0x7F800024, 0x7FC00004):
+    f = array("f", bytes(array("I", [float_hex])))[0]
+    if type(f) is float:
+        print("{:.4e}".format(f))
+    else:
+        print(f)
--- a/ports/windows/mpconfigport.h
+++ b/ports/windows/mpconfigport.h
@@ -266,6 +266,11 @@ typedef long mp_off_t;
 #endif
 #endif
 
+// VC++ 2017 fixes
+#if (_MSC_VER < 1920)
+#define MICROPY_PY_MATH_COPYSIGN_FIX_NAN (1)
+#endif
+
 // CL specific definitions
 
 #ifndef __cplusplus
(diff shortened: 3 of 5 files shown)


yoctopuce:
I ended up adding the REPR_C fix in the same PR as the impact on code footprint does not seems to be a problem on my ARMv4 platform, even when keeping the function inline.

tests/float/float_array.py
```diff
@@ -19,4 +19,10 @@ def test(a):
 test(array("f"))
 test(array("d"))
 
-print("{:.4f}".format(array("f", bytes(array("I", [0x3DCCCCCC])))[0]))
+# hand-crafter floats, including non-standard nan
```
dpgeorge: "crafted"

tests/float/float_array.py
```diff
@@ -19,4 +19,10 @@ def test(a):
...
+for _ in (0x3DCCCCCC, 0x7F800024, 0x7FC00004):
```
dpgeorge: Since the loop variable is being used, I suggest giving it a name like `float_hex`.

dpgeorge (commented):
Thanks for splitting this out into a separate PR.

For esp8266 it saves 16 bytes of code size!

But, 20 tests now fail on esp8266 :disappointed: 

Here are the failing tests: 

    20 tests failed: basics/nanbox_smallint.py float/builtin_float_abs.py float/builtin_float_pow.py float/builtin_float_round.py float/bytearray_construct_endian.py float/bytes_construct_endian.py float/float1.py float/float2int_fp30_intbig.py float/float_divmod_relaxed.py float/float_parse.py float/float_struct.py float/float_struct_e.py float/inf_nan_arith.py float/math_domain.py float/math_domain_python311.py float/math_fun.py float/math_fun_bool.py float/math_fun_int.py float/string_format_fp30.py float/string_format_modulo.py

Eg the output of `float/builtin_float_abs.py` is:
```
1.0 1.0
-1.0 1.0
0.0 2e-45
-0.0 2e-45
nan nan
-nan nan
inf nan
-inf nan
```
Lines 3 and 4 are wrong, they should be 0.0, but because bit 1 is not being cleared they are now very small numbers.

Lines 7 and 8 are wrong, they should be inf, but because bit 1 is not being cleared they are now nan.

So I don't think it's a good idea to drop the `& ~3u`.  I can understand the motivation (to round up some numbers), and it actually makes `tests/misc/rge_sm.py` nearly pass now!  That would be great, but I don't think this is the right approach.

(I did actually try something similar myself but instead in `mp_obj_new_float()` I tried rounding up the 30-bit float by adding 0x04 if bit 1 was set in the original float.  That gave quite good results, but I didn't get a chance to tidy that up for a PR yet.)

py/obj.h
```diff
@@ -203,14 +204,18 @@ static inline mp_float_t mp_obj_float_get(mp_const_obj_t o) {
...
     return num.f;
 }
 static inline mp_obj_t mp_obj_new_float(mp_float_t f) {
     union {
         mp_float_t f;
         mp_uint_t u;
     } num = {.f = f};
+    if (isnan(f)) {
+        // prevent creation of bad nanboxed pointers via array.array or struct
+        return (mp_obj_t)((((uint64_t)0xffc00000 & ~0x3u) | 2u) + 0x80800000u);
```
dpgeorge: How about moving this check before the union, and just returning `mp_const_float_nan`?  (You could remove the `MICROPY_PY_MATH_CONSTANTS` guard for that.)

py/obj.h
```diff
@@ -276,6 +284,10 @@ static inline mp_float_t mp_obj_float_get(mp_const_obj_t o) {
     return num.f;
 }
 static inline mp_obj_t mp_obj_new_float(mp_float_t f) {
+    if (isnan(f)) {
+        // prevent creation of bad nanboxed pointers via array.array or struct
+        return mp_obj_true_nan;
```
dpgeorge: Can we somehow just reuse `mp_const_float_nan`, eg:
```c
struct {
    uint64_t r;
} num = mp_const_float_nan;
return num.r;
```
?

py/obj.h
yoctopuce: Using `mp_const_float_nan` was my first idea, but `MP_ROM_PTR` is not defined at that point, so it cannot be used within an `inline` function. I will try another approach to avoid including the constant twice.

py/obj.h
yoctopuce: I like that :-)

yoctopuce:
@dpgeorge Sorry for the first commit yesterday that failed so many tests. I did not immediately figure out how I could get a 32bit Linux build on Ubuntu 24.04 for easily running the full CI tests.

I have now pushed a new version that fixes it, and takes your comments into account.
Can you tell how much bigger the esp8266 binary gets ?
I tried to build it myself, but was unable to find a place to download the proper tools (links in the ReadMe are now 404)

dpgeorge:
> Can you tell how much bigger the esp8266 binary gets ?

It's now +44 bytes.  That seems OK, considering it fixes a bug.

Running the tests with ESP8266_GENERIC firmware, they are now all passing, including the modified `float_array.py` test (but note that `math_constants_extra.py` doesn't run on this target -- it skips -- because it has these extra constants disabled).

yoctopuce:
> > Can you tell how much bigger the esp8266 binary gets ?
> 
> It's now +44 bytes. That seems OK, considering it fixes a bug.

Excellent, thank you !

> Running the tests with ESP8266_GENERIC firmware, they are now all passing, including the modified `float_array.py` test (but note that `math_constants_extra.py` doesn't run on this target -- it skips -- because it has these extra constants disabled).

I have been able to verify that `math_constants_extra.py` works using a 32bit unix build with REPR_C. Testing this as part of standard CI comes back to the discussion we had with @projectgus about having a REPR_C + longlong variant. I can work on it once his PR will be pulled in.

tests/float/float_array.py
dpgeorge: This still needs attention.

py/obj.h
```diff
@@ -184,13 +184,16 @@ static inline bool mp_obj_is_small_int(mp_const_obj_t o) {
 #define MP_OBJ_NEW_SMALL_INT(small_int) ((mp_obj_t)((((mp_uint_t)(small_int)) << 1) | 1))
 
 #if MICROPY_PY_BUILTINS_FLOAT
-#define MP_OBJ_NEW_CONST_FLOAT(f) MP_ROM_PTR((mp_obj_t)((((((uint64_t)f) & ~3) | 2) + 0x80800000) & 0xffffffff))
+#include <math.h>
+#define mp_nan_obj_bits (0x7fc00000)
+#define MP_OBJ_NEW_FLOAT(f) (mp_obj_t)((((((uint64_t)f) & ~3) | 2) + 0x80800000) & 0xffffffff)
+#define MP_OBJ_NEW_CONST_FLOAT(f) MP_ROM_PTR(MP_OBJ_NEW_FLOAT(f))
```
dpgeorge: I can understand why you've done things this way, but IMO it's a little complex and this code here is already complex enough.

So I suggest just using the known definition of `MP_ROM_PTR` for repr C, which is just the identity, and doing:
```c
#define MP_OBJ_NEW_CONST_FLOAT(f) ((mp_obj_t)((((((uint64_t)f) & ~3) | 2) + 0x80800000) & 0xffffffff))
```
ie drop `MP_ROM_PTR`.  Then `mp_const_float_nan` can be used directly below.

I think that will lead to the most readable code.

You could add a small comment above the definition of `MP_OBJ_NEW_CONST_FLOAT` saying something like "should be a MP_ROM_PTR but that macro isn't available yet".

tests/float/float_array.py
yoctopuce: Oops, sorry for that, I must have done it on the wrong branch. It is now fixed.

py/obj.h
```diff
@@ -207,6 +209,10 @@ static inline mp_float_t mp_obj_float_get(mp_const_obj_t o) {
     return num.f;
 }
 static inline mp_obj_t mp_obj_new_float(mp_float_t f) {
+    if (isnan(f)) {
+        // prevent creation of bad nanboxed pointers via array.array or struct
+        return MP_OBJ_NEW_FLOAT(mp_nan_obj_bits);
```
dpgeorge: Should be `return mp_const_float_nan`.

yoctopuce: I pushed it too fast :-( 
It's fixed now.

py/obj.h
```diff
@@ -257,12 +263,13 @@ static inline bool mp_obj_is_immediate_obj(mp_const_obj_t o) {
...
 #endif
 
+#include <math.h>
 #define mp_const_float_e {((mp_obj_t)((uint64_t)0x4005bf0a8b145769 + 0x8004000000000000))}
 #define mp_const_float_pi {((mp_obj_t)((uint64_t)0x400921fb54442d18 + 0x8004000000000000))}
 #if MICROPY_PY_MATH_CONSTANTS
 #define mp_const_float_tau {((mp_obj_t)((uint64_t)0x401921fb54442d18 + 0x8004000000000000))}
 #define mp_const_float_inf {((mp_obj_t)((uint64_t)0x7ff0000000000000 + 0x8004000000000000))}
-#define mp_const_float_nan {((mp_obj_t)((uint64_t)0xfff8000000000000 + 0x8004000000000000))}
+#define mp_const_float_nan {((mp_obj_t)((uint64_t)0x7ff8000000000000 + 0x8004000000000000))}
```
dpgeorge: This will need to come outside the `#if MICROPY_PY_MATH_CONSTANTS` guard.

dpgeorge: Although the CI passes, I think it's worth putting this line outside the `#if` guard, in case repr D is enabled without math constants.

yoctopuce: Yes, of course!

dpgeorge (approved):
Thanks for updating, looks good now!

Tested again on esp8266 and the tests pass.

PR #17538, opened 2025-06-20
Coverage test sys.settrace & improve coverage

jepler:
### Summary

I noticed that `py/profile.c` was not covered at all according to codecov, due to the fact it was never enabled in the unix port.

I enabled it, and then added a new test to cover some previously un-covered lines.

### Testing

I ran the tests locally and reviewed the coverage of profile.c.

Note: I suspect one or more Windows x64 builds will fail **due to improved coverage in this PR**, exposing a latent problem.

I wrote a [rambling post in the discussions area about passing qstrs to mp_printf](https://github.com/orgs/micropython/discussions/17537). The TL;DR explanation is, there appears to be undefined behavior that occurs when a "small qstr" (`uint16_t`) argument such as `type->name` is printed via the `%q` format specifier, which retrieves a `ssize_t` value. The specific behavior I saw with x64 msvc on godbolt looked like it would trigger on one specific format string in the core, the repr of frame objects. That's what spurred me to ensure that this code was covered.

If the failure does occur, I'll try to write it up properly as an issue.


jepler:
While there are some other errors going on that also leave me scratching my head, the "expected error" happens on Windows CI, specifically "build-vs (x64, Debug, dev, 2017)".

```
Run python run-tests.py --print-failures
--- D:/a/micropython/micropython/tests/results\misc_sys_settrace_cov.py.exp	2025-06-20 18:41:30.413152600 +0000
+++ D:/a/micropython/micropython/tests/results\misc_sys_settrace_cov.py.out	2025-06-20 18:41:30.413152600 +0000
@@ -1,2 +1,2 @@
-FRAME <frame at 0x\[0-9a-f\]\+, file 'misc/sys_settrace_repr\.py', line \\d\+, code f>
-LASTI \\d\+
+FRAME <frame at 0xe06df4a0, file 'D:/a/micropython/micropython/tests/misc/sys_settrace_cov.py', line 20, code Assertion failed: *q < pool->len, file d:\a\micropython\micropython\py\qstr.c, line 198
+CRASH
\ No newline at end of file

FAILURE D:/a/micropython/micropython/tests/results\misc_sys_settrace_cov.py
```

jepler:
I can't come up with any good theory as to what's making those unix builds fail. It's also a surprise that it's only ONE of the msvc x64 builds.

Josverl:
While working on extending sys.settrace() with additional capabilities I also noticed that the debug builds are not tested across ports. I first noticed Errors first popping up when building an ESP32 locally.

Is there a way to determine coverage locally?

jepler:
Yes, you can go through the steps of the unix coverage build locally. In `ports/unix`, `make VARIANT=coverage test_gcov`. The build output will finish with a summary of coverage, and a large number of ".gcov" files will be written. 

gcov files show coverage on a line by line basis, with "#####" indicating no coverage of the line, "-" indicating no coverage possible, and a number indicating the number of times the line was covered. The format is described in `man gcov`.

jepler:
The nanbox build were failing like this:
```
-FRAME <frame at 0x\[0-9a-f\]\+, file '\.\*/sys_settrace_cov.py', line \\d\+, code f>
+FRAME <frame at 0xf4d622e0, file 'misc/sys_settrace_cov.py', line 16, code >
```
The reason is that a `mp_uint_t` is passed in to a `%d` format specifier. But in nanbox, `sizeof(mp_uint_t) > sizeof(int)`, so again there's the UB problem that arises when the promoted argument type does not match the va_arg type.

jepler:
This PR is now failing only because it decreases the overall coverage percentage. In reality, it adds coverage to 21751 previously not covered lines.

ports/unix/mpconfigport.h
```diff
@@ -154,6 +154,11 @@ typedef long mp_off_t;
 // Don't default sys.argv and sys.path because we do that in main.
 #define MICROPY_PY_SYS_PATH_ARGV_DEFAULTS (0)
 
+// Enable sys.settrace by default
```
dlech: I though this wasn't previously enabled by default because it significantly affects performance. Would it make more sense to just enable in one of the unix port builds on CI?

jepler: I can change it so it's only in the coverage build, sure.

jepler:
I've updated the PR so that settrace is only enabled in the coverage build, not the standard build or other variants.

we should consider whether this makes the standard+settrace CI build redundant. I can remove that if desired. the standard+stackless+settrace CI job is probably still useful.

dpgeorge:
> I've updated the PR so that settrace is only enabled in the coverage build, not the standard build or other variants.

Yes, that's a good idea.  As mentioned, it really decreases performance.

OTOH, it's important to get coverage of the settrace code, so enabling it on the coverage build makes sense.

> we should consider whether this makes the standard+settrace CI build redundant. I can remove that if desired. the standard+stackless+settrace CI job is probably still useful.

Yes, we can now simplify those CI jobs.

These settrace jobs were added in 0b85b5b8b3357860c426cede526f48fdd84b56fd .  Maybe we can replace those two with just a test for stackless mode?  Or at least just remove the standard+settrace CI and keep standard+stackless+settrace (as you suggest).

Josverl:
wrt the work on `micropython-lib/python-ecosys/debugpy` I have been contemplating a few changes as well 
- moving the `misc/sys_settrace*.*` tests to their own folder: `tests/sys_settrace`
- there I have been updating and adding tests to cover the new functionality @andrewleech and I have been working on 

That makes it simpler to run that subset during development.
- debug variant ?
Also if you have guidance how to test the combination of a sys,settrace firmware together with `micropython-lib/python-ecosys/debugpy` that would be appreciated. Creating a separate 'settrace/debug' variant seems the simplest approach to me.

dpgeorge:
> * moving the `misc/sys_settrace*.*` tests to their own folder: `tests/sys_settrace`

That sounds reasonable, although not in scope for this PR.

> * Also if you have guidance how to test the combination of a sys,settrace firmware together with `micropython-lib/python-ecosys/debugpy` that would be appreciated. Creating a separate 'settrace/debug' variant seems the simplest approach to me.

This PR will enable settrace on the unix coverage build, so once this PR is merged you could use that.

dpgeorge:
@jepler there's a failing test here on the CI which I think is due to the code taking longer to run now that settrace is enabled.  It looks like the combination of already-long-running-test + settrace + ubsan = tests now takes longer than the 30s timeout.

I guess the simplest way to fix this is just increase `TEST_TIMEOUT` in `tests/run-tests.py`.

jepler:
> I guess the simplest way to fix this is just increase `TEST_TIMEOUT` in `tests/run-tests.py`.

I'll increase the timeout globally if that's ok. Or I could pipe through a timeout multiplier used only for 1 build.

dpgeorge:
> Or I could pipe through a timeout multiplier used only for 1 build.

Might be worth adding a command line argument for this timeout?  And then increase it just for this build?

Because if it's increased globally, eg to 60 seconds, then that's a long time to wait to know that a small test failed, if it locks up.

jepler:
I added an environment variable override and set it from ci.sh when running coverage tests. I tested that locally, by making sure that a timeout of 2 produced a handful of failures.

.github/workflows/ports_unix.yml
```diff
@@ -169,23 +169,6 @@ jobs:
...
 
-  settrace:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-    - uses: actions/setup-python@v5
-      # Python 3.12 is the default for ubuntu-24.04, but that has compatibility issues with settrace tests.
-      # Can remove this step when ubuntu-latest uses a more recent Python 3.x as the default.
-      with:
-        python-version: '3.11'
```
dpgeorge, line 180: I suggest copying this setup-python rule over to the coverage job.  Then it's possible to remove the `sys_settrace_features.py.exp` file.

I'd rather not have that .exp file because we want to retain compatibility with CPython with the output of settrace, and having an exp file makes it too easy to accidentally introduce an unwanted difference.

py/runtime.h
```diff
@@ -169,6 +169,12 @@ static inline void mp_thread_init_state(mp_state_thread_t *ts, size_t stack_size
     ts->nlr_jump_callback_top = NULL;
     ts->mp_pending_exception = MP_OBJ_NULL;
 
+    #if MICROPY_PY_SYS_SETTRACE
+    ts->prof_trace_callback = MP_OBJ_NULL;
+    ts->prof_callback_is_executing = 0;
```
dpgeorge: I think this should be `false` rather than `0`, to match the type of the destination.

py/objcode.h
```diff
@@ -72,7 +72,7 @@ static inline const void *mp_code_get_proto_fun(mp_obj_code_t *self) {
 
 #include "py/emitglue.h"
 
-#define MP_CODE_QSTR_MAP(context, idx) (context->constants.qstr_table[idx])
+#define MP_CODE_QSTR_MAP(context, idx) (qstr)(context->constants.qstr_table[idx])
```
dpgeorge: Please wrap the LHS here in parenthesis, so the thing the macro expands to is entirely in parenthesis.

dpgeorge:
Let me know when this is ready.

dpgeorge (approved):
Looking good now!

dpgeorge:
Squashed the fix-up commits and merged in fcfed6a0ea763ff14ea2bf819179fdf7eecda889 through a9801f9960ea2b8e94d5626840c97bc45286e1a3

jepler:
Thank you!

PR #17542, opened 2025-06-21
modmachine: Add `machine.wake_pins`.

meirarmon:
### Summary

When waking from deep sleep, it could be helpful to know what pins triggered the wake up since the wake pins could be configured to multiple pins.

Based on work that @m-cas did in this PR - https://github.com/micropython/micropython/pull/15498.

### Testing

Tested ESP32_GENERIC_C3 with:

```
import machine
import esp32
esp32.wake_on_gpio((machine.Pin(5),machine.Pin(2)), esp32.WAKEUP_ANY_HIGH)
machine.deepsleep()
```

After wakeup checked that `machine.wake_pins()` returned the correct pin that trigerred the wake.

I decided to place the method in `machine` together with `wake_reason`, since it is feasible that other ports
will want the exact same interface.

### DIFF
--- a/ports/esp32/modmachine.c
+++ b/ports/esp32/modmachine.c
@@ -36,6 +36,7 @@
 #include "esp_sleep.h"
 #include "esp_pm.h"
 
+#include "py/objtuple.h"
 #include "modmachine.h"
 #include "machine_rtc.h"
 
@@ -73,6 +74,7 @@
     \
     /* Wake reasons */ \
     { MP_ROM_QSTR(MP_QSTR_wake_reason), MP_ROM_PTR(&machine_wake_reason_obj) }, \
+    { MP_ROM_QSTR(MP_QSTR_wake_pins), MP_ROM_PTR(&machine_wake_pins_obj) }, \
     { MP_ROM_QSTR(MP_QSTR_PIN_WAKE), MP_ROM_INT(ESP_SLEEP_WAKEUP_EXT0) }, \
     { MP_ROM_QSTR(MP_QSTR_EXT0_WAKE), MP_ROM_INT(ESP_SLEEP_WAKEUP_EXT0) }, \
     { MP_ROM_QSTR(MP_QSTR_EXT1_WAKE), MP_ROM_INT(ESP_SLEEP_WAKEUP_EXT1) }, \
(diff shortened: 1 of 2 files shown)


ports/esp32/modmachine.c
```diff
@@ -254,6 +255,40 @@ static mp_obj_t machine_wake_reason(size_t n_args, const mp_obj_t *pos_args, mp_
...
+    #endif
+
+    // Only a few (~8) pins might cause wakeup.
+    // Therefore, we don't allocate 64*4 = 256 bytes on the stack and calculate the
+    // required space in a first pass.
+    for (index = 0, len = 0; index < 64; index++) {
+        len += (status & ((uint64_t)1 << index)) ? 1 : 0;
+    }
+    if (len) {
+        tuple = alloca(len * sizeof(*tuple));
```
projectgus: Rather than using `alloca` here suggest doing something like:

```c
#include "py/objtuple.h"

mp_obj_tuple_t *tuple = MP_OBJ_TO_PTR(mp_obj_new_tuple(len, NULL));
[...]
   tuple->items[i] = MP_OBJ_NEW_SMALL_INT(index)
[...]
return MP_OBJ_FROM_PTR(tuple)
```

projectgus (commented):
This looks like a useful function to me, and a good abstraction to eventually support on other ports as well. Thanks @meirarmon!

ports/esp32/modmachine.c
meirarmon: @projectgus Done, shall we get this merged :-) ?

ports/esp32/modmachine.c
```diff
@@ -301,6 +256,39 @@ static mp_obj_t machine_wake_reason(size_t n_args, const mp_obj_t *pos_args, mp_
...
+    #endif
+
+    #if SOC_PM_SUPPORT_EXT1_WAKEUP && SOC_RTCIO_PIN_COUNT > 0
+    status |= esp_sleep_get_ext1_wakeup_status();
+    #endif
+
+    // Only a few (~8) pins might cause wakeup.
+    // Therefore, we calculate the required space in a first pass.
+    for (index = 0, len = 0; index < 64; index++) {
+        len += (status & ((uint64_t)1 << index)) ? 1 : 0;
```
projectgus:
```suggestion
        len += (status & (1ULL << index)) ? 1 : 0;
```

(This is nitpicky, but it's easier to read.)

ports/esp32/modmachine.c
```diff
@@ -301,6 +256,39 @@ static mp_obj_t machine_wake_reason(size_t n_args, const mp_obj_t *pos_args, mp_
...
+    }
+    if (len) {
+        mp_obj_tuple_t *tuple = MP_OBJ_TO_PTR(mp_obj_new_tuple(len, NULL));
+
+        for (index = 0, len = 0; index < 64; index++) {
+            if (status & ((uint64_t)1 << index)) {
```
projectgus:
```suggestion
            if (status & (1ULL << index)) {
```

projectgus (approved):
Hey @meirarmon, thanks for being so patient while I got back to this.

I have one very nitpicky optional suggestion, but this looks good to me and I think we should merge it either way. Thank you for contributing it!

dpgeorge:
Now merged!

PR #17546, opened 2025-06-21
port/rp2/modmachine: Fix lightsleep watchdog interactions.

### LINKED ISSUE #17228: RP2040 Watchdog and lightsleep issue

### LINKED ISSUE #17229: RP2350 Watchdog and lightsleep issue

cpottle9:
### Summary
_Second try at this change. I had trouble resolving a merge conflict in an earlier pull request and gave up._

Aside: I am a little hesitant to propose these changes. If the maintainers decide to reject this pull request I will understand.
Lightsleep on rp2 is gradually getting more complex and these changes would contribute to that complexity.

Fixes RP2040 issue #17228 and RP2350 issue #17229.

RP2040 issue is when watchdog timer expires while
lightsleep is running the RP2040 will lock up during power-on startup.
This occurs because the power-on state machine is configured to not re-initialize the ROSC and lightsleep stopped the ROSC.

RP2350 issue is the watchdog timer does not decrement while lightsleep is running.
Once lightsleep returns the timer will start decrementing again. This occurs because lightsleep clears the bit
CLOCKS_SLEEP_EN1_CLK_SYS_WATCHDOG_BITS while sleeping.

### Testing
I ran the reproduction tests from the two issues.
I also ran tests/port/rp2/rp2_lightsleep.py.

General alternative. Document that lightsleep and watchdog don't play well together and leave the code as is.

Alternatives for RP2040 issue #17228:
- Raise issue for pico-sdk. Get them to change function _watchdog_enable() so the ROSC and XOSC do get reset on a watchdog timeout. The existing behavior in _watchdog_enable() has been there from day one. There might not be a good reason for it. Disadvantage: a possible long wait for a new pico-sdk release.
- Modify lightsleep to not stop the ROSC. I measure 4.96 milli-amps when lightsleep disables ROSC and 5.09 milli-amps when it does not. This is a small change.

Alternatives for the RP2350 issue #17229:
- I don't have any :-).

### DIFF
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -36,8 +36,12 @@
 #include "clocks_extra.h"
 #include "hardware/pll.h"
 #include "hardware/structs/rosc.h"
+#if PICO_RP2040
+#include "hardware/structs/psm.h"
+#endif
 #include "hardware/structs/scb.h"
 #include "hardware/structs/syscfg.h"
+#include "hardware/structs/watchdog.h"
 #include "hardware/watchdog.h"
 #include "hardware/xosc.h"
 #include "pico/bootrom.h"
@@ -192,6 +196,8 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
         clock_stop(clk_usb);
     }
 
+    bool watchdog_active = (watchdog_hw->ctrl & WATCHDOG_CTRL_ENABLE_BITS) != 0;
+
     clock_stop(clk_adc);
     #if PICO_RP2350
     clock_stop(clk_hstx);
@@ -218,6 +224,12 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
     }
 
     // Disable ROSC.
+    #if PICO_RP2040
+    if (watchdog_active) {
+        // Configure Power-On State Machine to reset the ROSC on a watchdog timeout.
+        psm_hw->wdsel |= PSM_WDSEL_ROSC_BITS;
+    }
+    #endif
     rosc_hw->ctrl = ROSC_CTRL_ENABLE_VALUE_DISABLE << ROSC_CTRL_ENABLE_LSB;
 
     #if DEBUG_LIGHTSLEEP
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thanks for the fix.  This looks clean and works well!  I tested it on both RPI_PICO and RPI_PICO2 (ARM and RISC-V modes).

I think the fix here for RP2040 is the most sensible thing to do, given all the constraints.  I don't think pico-sdk should enable PSM_WDSEL_ROSC_BITS by default.

PR #17548, opened 2025-06-22
py/emitndebug: Indicate signedness for setcc opcodes.

agatti:
### Summary

This commit changes the output of the `setcc` opcode in the debug emitter, in order to know whether a comparison operates on signed or unsigned values.

With this change the `setcc` opcode turns from being described as `setcc(<cc>, <dest>, <lhs>, <rhs>)` into `setcc(<cc>, <dest>, <lhs>, <rhs>, <unsigned>)`.  As integer variables in Python are signed by default, the flag is made to be `true` for unsigned comparisons, as they are the special case.

I'm working on a tool that generates retargettable binary code from the output of the debug emitter, and not knowing whether a `setcc` operation is supposed to work on signed or unsigned variables is sort of a problem.

Given that this seems to be the only piece missing for third parties to consume the debug emitter's output, it'd be a shame for that opcode to block that.

### Testing

Testing was performed by printing the output of `mpy-cross -X emit=native -march=debug test.py`, with `test.py` being as such:

```py
@micropython.viper
def signed(c: int, n: int):
    print(c == n)

@micropython.viper
def unsigned(c: uint, n: uint):
    print(c == n)
```

and checking that the first function would contain a `setcc` opcode with `false` as its last argument, and the second function would contain a `setcc` opcode with `true` as its last argument.

The (limited) size increase is limited to `mpy-cross`, and thus is not really impacting any existing port.  There's no other way to access the pre-emitted opcodes list, short of adding a new emitter that outputs that list in a structured data format like JSON or XML.

### DIFF
--- a/py/emitnative.c
+++ b/py/emitnative.c
@@ -2634,7 +2634,7 @@ static void emit_native_binary_op(emit_t *emit, mp_binary_op_t op) {
                     break;
             }
             #elif N_DEBUG
-            asm_debug_setcc_reg_reg_reg(emit->as, op_idx, REG_RET, REG_ARG_2, reg_rhs);
+            asm_debug_setcc_reg_reg_reg(emit->as, op_idx, REG_RET, REG_ARG_2, reg_rhs, vtype_lhs == VTYPE_UINT);
             #else
             #error not implemented
             #endif
--- a/py/emitndebug.c
+++ b/py/emitndebug.c
@@ -166,8 +166,8 @@ static void asm_debug_reg_reg_label(asm_debug_t *as, const char *op, int reg1, i
     asm_debug_printf(as, "%s(%s, %s, label_%u)\n", op, reg_name_table[reg1], reg_name_table[reg2], label);
 }
 
-static void asm_debug_setcc_reg_reg_reg(asm_debug_t *as, int op, int reg1, int reg2, int reg3) {
-    asm_debug_printf(as, "setcc(%d, %s, %s, %s)\n", op, reg_name_table[reg1], reg_name_table[reg2], reg_name_table[reg3]);
+static void asm_debug_setcc_reg_reg_reg(asm_debug_t *as, int op, int reg1, int reg2, int reg3, bool unsigned_comparison) {
+    asm_debug_printf(as, "setcc(%d, %s, %s, %s, %s)\n", op, reg_name_table[reg1], reg_name_table[reg2], reg_name_table[reg3], unsigned_comparison ? "true" : "false");
 }
 
 // The following macros provide a (mostly) arch-independent API to


agatti:
Turns out there *is* another way to get that information out - my debug output parser silently ignored any condition code for signed comparisons :(

Apologies for submitting this - closing now.

dpgeorge:
> I'm working on a tool that generates retargettable binary code from the output of the debug emitter

Oh, that's interesting!

> There's no other way to access the pre-emitted opcodes list, short of adding a new emitter that outputs that list in a structured data format like JSON or XML.

I never thought the debug emitter could be parsed and used programatically, it was really just for human debugging.  We could eventually add a way for it to output to JSON (or S-expressions?) if that would be beneficial.  Or even just improve the debug output, eg at least print the function name and maybe arguments.

agatti:
> I never thought the debug emitter could be parsed and used programatically, it was really just for human debugging. We could eventually add a way for it to output to JSON (or S-expressions?) if that would be beneficial. Or even just improve the debug output, eg at least print the function name and maybe arguments.

Well, how about both?  I'm up for writing the JSON emitter (there's already a JSON serialiser in MicroPython after all) and to improve the debug emitter output.

<hr>

Right now I've got something that translates the debug emitter's output into LLVM IR for around 40% of the MicroPython test suite, and then turns the IR to object code using the rest of the LLVM toolchain/libraries.  Making the output run on device currently requires some binary cut and paste into an existing MPY file and the process is nowhere near being automatic at the moment.

Once I'm done with the IR generation and made sure things work properly then I see two possible options:

* the tool stays external and works as a MPY file post-processor, but this needs changes in `mpy-cross` and `mpy-tool.py`: the serialised opcodes form from either the new JSON emitter or the existing debug emitter should be placed as an additional META chunk in the output MPY file by `mpy-cross` if asked and optionally stripped away by `mpy-tool.py` (a way to replace MPY sections from `mpy-tool.py` would be an added bonus here).  The drawback here is that maybe all these changes are just extra complexity for something used by just a single external tool.

* if the lack of Xtensa support is not a deal-breaker for the time being (assuming Espressif's Xtensa LLVM support upstreaming is eventually carried out) this can be rewritten in C as a separate emitter that gets brought in only from `mpy-cross`, just like the debug emitter does.  It will require some makefile tinkering to conditionally link LLVM in, but it is somehow doable.  However, this brings in some possibly unwanted cognitive load: LLVM's C API isn't that complicated (the tool is written in Rust, and it uses a rather thin wrapper over LLVM's C API anyway), but debugging why things don't work can sometimes get "interesting" really quick.

That said, things look promising - right now, when things work, the size savings are noticeable (~30-35% size decrease on RV32 for selected functions, I can give figures for other architectures if you want).  But what I'm really looking forward to is being able to tell the arch backend to emit code that uses (or doesn't use) extra CPU features or targets a specific CPU model.

Once I reach a sizeable percentage of the MicroPython test suite that gets correctly compiled I'll push this on its own repo and see where things go from there :)

PR #17551, opened 2025-06-23
tools/mpremote: Support OSError's on targets without errno.

dpgeorge:
### Summary

Targets without the `errno` module enabled will not render `OSError`s with the name of the error.  Instead they just print the numeric error code.

Add support for such targets by explicitly recognising certain error codes.

### Testing

Tested on SEEED_XIAO_SAMD21: prior to this fix doing an `mpremote mip install` would crash with an exception.  Now it works.

This hard codes the errno values based on `py/mperrno.h`, but I don't think there's any other option.  This is anyway what most bare-metal ports use, especially if they are minimal and have the `errno` module disabled.


AJMansfield:
This check could be made generic by doing it as part of the for-loop, using that loop's numeric `code` value. Something like:

```python
def _convert_filesystem_error(e, info):
    if "OSError" in e.error_output:
        error_lines = e.error_output.splitlines()
        for code, estr in [
            *errno.errorcode.items(),
            (errno.EOPNOTSUPP, "EOPNOTSUPP"),
        ]:
            if estr in e.error_output:
                return OSError(code, info)
            
            if error_lines[-1] == f"OSError: {code}":
                return OSError(code, info)
    return e
```

dpgeorge:
> This check could be made generic by doing it as part of the for-loop, using that loop's numeric `code` value

Unfortunately this won't be portable, because different host PCs can have different errno codes.  Eg 2 is not always ENOENT on a PC, but we know that on MicroPython bare-metal that it is (with `MICROPY_USE_INTERNAL_ERRNO` enabled).

dpgeorge:
I've added support for a bunch of errno codes, which should cover all the important cases.

AJMansfield:
Why not just the entire list from mperrno.h?

dpgeorge:
> Why not just the entire list from mperrno.h?

That's a fair point.  It's about 50 entries, which isn't too many to add.

dpgeorge:
I've now added all errno's from `py/mperrno.h`.

Tested again on SEEED_XIAO_SAMD21, still works.

AJMansfield (approved):
I've tested this against my Pico W and Pico 2 and can confirm all tests in the mpremote suite pass.

PR #17552, opened 2025-06-23
zephyr: a few quality-of-life improvements, and additional board configurations

dpgeorge:
### Summary

The main aim here was to get `mpremote mip install unittest` working on the zephyr port, so it could run more of the test suite.  And that lead to a lot of little changes.

Overall this PR aims to bring the zephyr port a bit closer to other ports in terms of functionality and usability:
- execute `boot.py` and `main.py` like other ports
- add `/flash/lib` to `sys.path` on start up if there's a filesystem (eg `mpremote mip install` now works)
- enable `sys.stdin/out/err`
- improve REPL reliability and performance, it now passes the test from #15909, and copying files with `mpremote` is now reliable
- fix USB device_next driver compilation error
- add `Pin.OPEN_DRAIN` support
- allow constructing a Pin from an existing Pin (follows other ports)
- implement C-level pin HAL
- enable `machine.SoftI2C`
- enable importing of .mpy files

It also tweaks/adds a few board configurations:
- frdm_k64f: get ctrl-C working, enable PWM, increase heap size
- nucleo_wb55rg: enable BLE, I2C, SPI and add filesystem
- rpi_pico: use USB for REPL, enable I2C, SPI, add filesystem, increase heap size

In particular, the filesystem settings for `rpi_pico` now match the rp2 port RPI_PICO board, so you can switch between zephyr and rp2 firmware and retain the filesystem (and even .mpy files on that filesystem work across the two ports!).  That allows some nice side-by-side comparisons between the zephyr and rp2 ports for the same hardware.  Eg USB CDC throughput is quite a lot higher on the rp2 port using TinyUSB, compared with zephyr's USB device (but maybe this is due to zephyr configuration issues, I didn't investigate further).

### Testing

Tested with frdm_k64f, nucleo_wb55rg and rpi_pico zephyr firmware:
- The serial test from #15909 now passes on all three boards.
- Can do `mpremote mip install unittest` and it works.
- Can now run a lot more of the test suite, including `--via-mpy` tests.
- Tested `machine.Pin.OPEN_DRAIN` and `machine.SoftI2C` (actually as part of #17365).

This increases the firmware size of zephyr builds a bit, but (1) these boards have enough flash and RAM; and (2) it's easy enough for custom boards to reconfigure/disable features if needed.


ports/zephyr/machine_pin.c
```diff
@@ -271,6 +279,7 @@ static const mp_rom_map_elem_t machine_pin_locals_dict_table[] = {
     // class constants
     { MP_ROM_QSTR(MP_QSTR_IN),        MP_ROM_INT(GPIO_INPUT) },
     { MP_ROM_QSTR(MP_QSTR_OUT),       MP_ROM_INT(GPIO_OUTPUT) },
+    { MP_ROM_QSTR(MP_QSTR_OPEN_DRAIN), MP_ROM_INT(GPIO_OUTPUT | GPIO_OPEN_DRAIN) },
```
bikeNomad: Should this be:
```c
{ MP_ROM_QSTR(MP_QSTR_OPEN_DRAIN), MP_ROM_INT(GPIO_OUTPUT | GPIO_OPEN_DRAIN | GPIO_INPUT) },
```

ports/zephyr/mphalport.h
```diff
@@ -35,3 +36,33 @@ static inline uint64_t mp_hal_time_ns(void) {
...
+mp_hal_pin_obj_t mp_hal_get_pin_obj(mp_obj_t pin_in);
+
+static inline unsigned int mp_hal_pin_name(mp_hal_pin_obj_t pin) {
+    // TODO make it include the port
+    return pin->pin;
+}
+
+static inline void mp_hal_pin_open_drain(mp_hal_pin_obj_t pin) {
+    (void)gpio_pin_configure(pin->port, pin->pin, GPIO_OUTPUT | GPIO_OPEN_DRAIN);
+}
```
bikeNomad, line 75: Should this be:
```c
    (void)gpio_pin_configure(pin->port, pin->pin, GPIO_OUTPUT | GPIO_OPEN_DRAIN | GPIO_INPUT);
```

bikeNomad (changes requested):
Just a couple of comments regarding being able to read from output/open-drain Pins.

ports/zephyr/machine_pin.c
dpgeorge: Done.

ports/zephyr/mphalport.h
dpgeorge: Done.

ports/zephyr/src/usbd.c
```diff
@@ -137,7 +131,7 @@ struct usbd_context *mp_usbd_init_device(usbd_msg_cb_t msg_cb) {
         return NULL;
     }
 
-    err = usbd_register_all_classes(&mp_usbd, USBD_SPEED_FS, 1, blocklist);
+    err = usbd_register_all_classes(&mp_usbd, USBD_SPEED_FS, 1);
```
danicampora: In the latest Zephyr API docs, this function receives 4 params, which includes the blocklist. Which Zephyr version did you compile against?

https://docs.zephyrproject.org/apidoc/latest/group__usbd__api.html

dpgeorge: I'm using 4.0.0, which definitely has only 3 parameters... I guess we need to have a `#if` here to support that and newer versions at the same time?

dpgeorge: OK, they added the blocklist in https://github.com/zephyrproject-rtos/zephyr/commit/3bfe16381157019a70c57d1e363172decad25949, which is in v4.1.0 but not v4.0.0.

dpgeorge: I've now made this compatible with both v4.0.0 and v4.1.0.

Thanks for the review!

dpgeorge:
I've added 3 more commits/changes here:
- enable `sys.maxsize`; that costs +48 bytes but allows introspection to the target, needed at least for some tests to work
- enable emergency exception buffer, again needed for a few tests to run and pass (otherwise the device can lock up running the `tests/micropython/heapalloc_exc_compressed_emg_exc.py` test)
- make the id argument to `machine.Timer` optional, defaulting to -1 (they only allowed value); that matches other ports and gets the `tests/extmod/machine_soft_timer.py` test passing

Testing was done with `rpi_pico` firmware.

danicampora:
> I've added 3 more commits/changes here:

Looks good!

jonnor:
Very nice! Zephyr support seems to be soaring ahead lately :D

PR #17553, opened 2025-06-23
stm32/boards/ARDUINO_PORTENTA_H7: Free reserved timer.

iabdalkader:
### Summary

This patch frees the timer pin used for the camera clock on Portenta carrier, and vision shield, allowing users to use it with `tim`. The pin doesn't strictly need to be reserved, as the timer module only deinitializes timers that were used. See https://github.com/micropython/micropython/commit/b6649b922e0ff7fa263b05d560f4c11a3910d1d9#commitcomment-160466116

### Testing

Tested Portenta firmware still works.


razvanphp:
LGTM. Thank you!

dpgeorge (approved):

PR #17556, opened 2025-06-24
Add compile-time checking of mp_printf format strings

jepler:
### Summary

It's always nice when errors can be detected at compile time. In traditional C programs, gcc can check that the printf argument types match the printf format string. This has not been possible up to now with mp_printf, because it has both extensions to standard printf (e.g., the `%q` format type) and is missing things in standard printf (e.g., `%zd` is not supported).

To that end, I have developed a GCC plugin that does this checking at compile time, and enabled it during the ci build process where possible.

### Testing

I built the unix port coverage variant & ran the tests locally. The plugin itself should cause no code changes, and the code size check (which does NOT turn on the plugin) reports no size change.

A total of 45 sites in ci.sh use the new plugin.

As a gcc plugin this can only support gcc-based toolchains. clang and proprietary compilers cannot use the plugin. This does not seem important, as this feature only produces diagnostics.

The plugin is GPL licensed. I started with a GPL-licensed plugin template, and plugins need to be GPL-or-compatible in license in order to be loaded in gcc anyway. The plugin code IMO does not affect the license situation of the output object code, as you'd get the exact same code with or without the plugin.

Missing support for:
 * Knowing whether `%ll` is runtime-supported
 * handling enum argmuents properly (just one enum was %d printed across all checked builds and I added a cast instead of fixing the checker)
 * non-gnumake based builds
 * mingw-based builds on windows (mingw-based cross builds are checked though)
 * mac-based builds

I never specifically measured the build-time impact of enabling the plugin.

### DIFF
--- a/.github/workflows/ports_unix.yml
+++ b/.github/workflows/ports_unix.yml
@@ -171,6 +171,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v6
+    - name: Install packages
+      run: source tools/ci.sh && ci_unix_float_setup
     - name: Build
       run: tools/ci.sh unix_float_build
     - name: Run main test suite
--- a/docs/develop/gettingstarted.rst
+++ b/docs/develop/gettingstarted.rst
@@ -234,6 +234,24 @@ You can also specify which board to use:
 See `ports/stm32/boards <https://github.com/micropython/micropython/tree/master/ports/stm32/boards>`_
 for the available boards. e.g. "PYBV11" or "NUCLEO_WB55".
 
+Compile-time format string checking
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When gcc is used to build MicroPython, a plugin can be used for compile-time
+checking of ``mp_printf`` format strings. The plugin is enabled by setting the
+Makefile variable ``MICROPY_USE_COMPILER_PLUGIN=gcc`` before including
+``py/mkrules.mk``.
+
+The plugin doesn't work:
+ * With non-gcc compilers (including clang, which is sometimes installed on
+   Macs with the name "gcc")
+ * On Windows systems, where the steps for building a plugin are more complicated
+ * With MicroPython builds that use cmake rather than traditional make.
+ * If the necessary files for plugin development are not installed. In Debian
+   bookworm, for instance, the files to build plugins for ``gcc`` are in
+   ``gcc-12-plugin-dev``, and the files to build plugins for
+   ``riscv64-unknown-elf-gcc`` are in ``gcc-12-plugin-dev-riscv64-linux-gnu``.
+
 Building the documentation
 --------------------------
 
--- a/extmod/machine_can.c
+++ b/extmod/machine_can.c
@@ -664,7 +664,7 @@ static void machine_can_print(const mp_print_t *print, mp_obj_t self_in, mp_prin
             break;
     }
 
-    mp_printf(print, "CAN(%d, bitrate=%u, mode=CAN.%q, sjw=%u, tseg1=%u, tseg2=%u, f_clock=%u)",
+    mp_printf(print, "CAN(" INT_FMT ", bitrate=%u, mode=CAN.%q, sjw=%u, tseg1=%u, tseg2=%u, f_clock=%u)",
         self->can_idx + 1,
         actual_bitrate,
         mode,
--- a/ports/nrf/Makefile
+++ b/ports/nrf/Makefile
@@ -147,7 +147,7 @@ CFLAGS += -fno-strict-aliasing
 CFLAGS += -I$(BOARD_DIR)
 CFLAGS += -DNRF5_HAL_H='<$(MCU_VARIANT)_hal.h>'
 
-LDFLAGS += $(CFLAGS) -nostdlib
+LDFLAGS += $(filter-out -fplugin%, $(CFLAGS)) -nostdlib
 LDFLAGS += -Xlinker -Map=$(@:.elf=.map)
 LDFLAGS += -mthumb -mabi=aapcs $(addprefix -T,$(LD_FILES)) -L boards/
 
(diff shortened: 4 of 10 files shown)


jepler:
Example diagnostic:
```
coverage.c: In function ‘extra_coverage’:
coverage.c:206:9: error: argument 3: expected ‘long int’ or ‘long unsigned int’, not ‘int’ [-
Werror=format=]
  206 |         mp_printf(&mp_plat_print, "%ld\n", 123); // long
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

jepler:
the plugin can also run during the gcc windows builds. I tested it locally using the cross-building steps but I assume it'd work on windows too. I won't complicate this PR by adding it, but I'd plan to add it in a subsequent PR. Some of the format string "findings" came out of me doing that locally.

jepler:
@dpgeorge Please let me know if you think this is worth pursuing.

jepler:
I did some checks in a sibling branch, and on macos, `gcc` is clang (!) which has a different incompatible plugin api; gcc-14 is also installed via brew in the runner, but it is missing some indirect dependency required for plugin building (`/opt/homebrew/Cellar/gcc@14/14.3.0/bin/../lib/gcc/14/gcc/aarch64-apple-darwin23/14/plugin/include/system.h:700:10: fatal error: 'gmp.h' file not found`). this may be solvable with brew, but as this would not cover any additional source files or integer size models it's probably not worth the time to figure out how to fix it.

jepler:
[Plugin support on Windows/MinGW has a number of limitations and additional requirements](https://gcc.gnu.org/onlinedocs/gccint/Plugins-building.html#:~:text=Plugin%20support%20on%20Windows/MinGW%20has%20a%20number%20of%20limitations%20and%20additional%20requirements) so adding support for that would better be postponed to a subsequent PR.

dpgeorge:
> Please let me know if you think this is worth pursuing.

I'm mildly in favour of this.

The three main things to consider would be:
1. The code being GPL licensed: no GPL code in this repo can be part of compiled firmware/executables.  But the code added here is only part of the toolchain, which is OK.  After all, gcc itself is GPL, and having this new plugin under GPL is the same as that, it has the same scope.
2. Increased complexity of the build process: before building any object files the plugin needs to be built.  I guess this is fine, although it does add yet more rules to Make (and eventually CMake).
3. Being enabled by default: this will probably annoy some developers who would either need to install `gcc-plugin-dev` or disable it each time with `DISABLE_PLUGIN=1`.  Luckily Arch Linux includes `gcc-plugin-dev` by default (for both gcc and arm-none-eabi-gcc) so that lowers the bar somewhat.

I did test out this PR locally with the unix coverage build and it works well.

It looks like this did find same cases of `mp_printf` that need to be fixed, so I guess that's a big reason to have it.

jepler:
Thanks for the feedback.

I did consider trying to automatically enable the plugin if possible, and disable it if not; but this looked fragile.

One option would be to switch it to requring the plugin to be *enabled* (and enabling it during as many CI jobs as possible). A developer who runs into a diagnostic during CI would then have the option to install the plugin and use ENABLE locally, or whether to make a stab at fixing it and submit another job to CI.

Would it make more sense to work on fixing the diagnosed format problems in a separate PR (some of them do seem to be "real bugs" that will bite at runtime) and then bring the format checker in later? Or are you content to let the fixes and the checker land at the same time, which would make the fixes later?

Final question: C99 "solves" some format string problems by using macros like PRId64 that expand to a correct format string depending on the underlying types (e.g., it might expand to "lld" on an LLp64 sytem or "ld" on an LP64 system). What do you think of introducing such macros in micropython for mp_int_t/mp_uint_t? It looks like there's also a problem integer-printing pointer-width types which leads to the current Windows build failures (and which I'll correct)..

(Some issues are getting fixed in #17538 because the problems DID turn up during CI; this branch will need to be rebased when that one goes in)

dpgeorge:
> One option would be to switch it to requring the plugin to be _enabled_

Yes, I also thought about that.

At the very least, I think the option should be in the positive tense, eg `GCC_ENABLE_MP_PRINTF_PLUGIN`, and that could either be disabled by default, or enabled by default on selected builds (eg just unix coverage).

> Would it make more sense to work on fixing the diagnosed format problems in a separate PR

Yes, please.  That PR would be a much easier thing to review and merge.

> Final question: C99 "solves" some format string problems by using macros like PRId64
> ...
> What do you think of introducing such macros in micropython for mp_int_t/mp_uint_t?

Yes, I think that's a good idea.

I have many times considered using the PRIxxx macros for all printf strings.  But it's a fair bit of work to do that.  But a good idea to start with them for `mp_int_t`/`mp_uint_t`.

jepler:
My thoughts on sequencing the work:
 1. Land https://github.com/micropython/micropython/pull/17538 because it has some initial format-string fixes
 2. Cherry pick just the fixes from this PR into a new branch and get it to pass CI.
 3. Introduce PRI-macros if it helps. Internally I'll use the format checker during this step.
 4. Finally, once that new format string fix PR is in, rebase & return to this PR.

jepler:
huh. two wrongs make a "fascinating". I didn't expect to see code size savings. Is it coming from 3f9b9c4, which I had actually intended to rebase out? Let's find out over at https://github.com/micropython/micropython/pull/17618

jepler:
More sequencing: Once #17618 goes in, I'll rebase this, add XINT_FMT (the format string to print mp_int_t as hex), and use it for cell printing (@jepler [objcell: Fix printing cell ID.](https://github.com/micropython/micropython/pull/17556/commits/f34089b31255047ecc3bd404f14a13c0b2275f47)). Once that's done and all green with the checker in this PR, I'll create a separate "fixes only" PR.

jepler:
Rebased. I'll get it back to green, then submit a 2nd PR with "just the bug fixes".

jepler:
OK, everything this PR depends on has been merged. I've updated the initial comment to reflect the current state of things.

jepler:
@dpgeorge I think this is ready for review.

jepler:
I have rebased this. I found some newly introduced mistakes in mp_printf argument types (#18923) which did not lead directly to test failures on any tested platform. Please let me know if there is interest in incorporating this compile time check. Otherwise, I can close up the PR.

PR #17557, opened 2025-06-24
py: Add PEP 750 template strings support

koxudaxi:
Implements [PEP 750](https://peps.python.org/pep-0750/) template strings for MicroPython.

Started in discussion #17497. Template strings return Template objects instead of strings, allowing access to literal parts and expressions separately.

Changes:
- Lexer/parser support for t-string syntax
- Template and Interpolation types
- Public type definitions in py/objtstring.h
- Minimal builtin string.templatelib module
- Conditional build with MICROPY_PY_TSTRINGS

Usage:
```python
from string.templatelib import Template

t = t"Hello {name}!"
t.strings         # ('Hello ', '!')
t.values          # ('World',)
t.interpolations[0].expression  # 'name'

t"{x!r:>10}"      # Conversions: !r, !s, !a
rt"C:\{file}"     # Raw t-strings
```

Implementation:
- Lexer (py/lexer.c) and parser (py/parse.c) handle tokenization
- Template/Interpolation types in py/modtstring.c, py/objinterpolation.c
- Public header py/objtstring.h eliminates code duplication
- Minimal string module in py/modstring.c (no manifest.py changes)
- Dynamic array doubling (4→8→16→...→2048), GC-safe
- CPython 3.14 compatible error messages

Testing:

Tested on unix (coverage, standard, minimal), windows, webassembly.

New tests in tests/basics/:
- string_module_tstring.py
- tstring_basic.py, tstring_parser.py, tstring_errors.py
- tstring_format.py, tstring_operations.py, tstring_coverage.py

Feature detection in tests/feature_check/tstring.py - automatic skip when MICROPY_PY_TSTRINGS=0.

All existing tests pass.

Trade-offs:

Code size: ~10 KB when enabled, zero when disabled.

String module: minimal builtin with only templatelib blocks micropython-lib's full string module. Future: coordinate with micropython-lib.

Config: enabled when MICROPY_CONFIG_ROM_LEVEL >= EXTRA_FEATURES, requires MICROPY_PY_FSTRINGS=1.

To disable: `make CFLAGS_EXTRA=-DMICROPY_PY_TSTRINGS=0`


WebReflection:
for what is worth it, we'd love to have this available at least for the *PyScript* *WASM* variant as this unlocks tons of UI related use cases we'd like to deliver to our users.

/cc @dpgeorge @ntoll

docs/library/btree.rst
```diff
@@ -130,31 +130,50 @@ Methods
    to get access to all keys in order.
 
 .. method:: btree.keys([start_key, [end_key, [flags]]])
-            btree.values([start_key, [end_key, [flags]]])
```
dpgeorge, line 133: This is an unrelated change, so please put it in a separate PR.

(This PR itself will be big enough, so let's aim to make it as easy as possible to review.)

py/modbuiltins.c
```diff
@@ -525,6 +525,326 @@ static mp_obj_t mp_builtin_sum(size_t n_args, const mp_obj_t *args) {
 }
 MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(mp_builtin_sum_obj, 1, 2, mp_builtin_sum);
 
+#if MICROPY_PY_TSTRINGS
```
dpgeorge: I think this new code should go in its own, separate file.

dpgeorge (commented):
Thanks for the contribution!  This will be nice to have for the webassembly port.

I didn't do a review yet, but there will need to be tests to get full coverage of the new code.

koxudaxi:
@dpgeorge Thank you for the feedback and for taking time to look at this! I'm glad this will be useful for the webassembly port.

I'll add more tests to ensure full coverage when I find time.

ntoll:
@koxudaxi slightly off topic - but I notice you'll be at EuroPython in Prague, as will I. We should look out for each other and have a coffee or lunch together! :eu: :snake:

koxudaxi:
@ntoll Sounds good! See you in Prague. ☕

frozen/string/templatelib.py
```diff
@@ -0,0 +1,10 @@
+"""
```
dpgeorge: I don't think you need this frozen code anymore, because it's now implemented in C in `py/modstring.c`.

koxudaxi: Thank you. I forgot to delete the files.

koxudaxi:
@dpgeorge 

> I didn't do a review yet, but there will need to be tests to get full coverage of the new code.

I tried really hard to make 100% coverage but some code is never called and I cannot cover it. Do you know how to fix this?

docs/library/btree.rst
```diff
@@ -151,10 +151,10 @@ Constants
...
 .. data:: INCL
 
-   A flag for `keys()`, `values()`, `items()` methods to specify that
+   A flag for :meth:`btree.keys`, :meth:`btree.values`, :meth:`btree.items` methods to specify that
    scanning should be inclusive of the end key.
 
 .. data:: DESC
 
-   A flag for `keys()`, `values()`, `items()` methods to specify that
+   A flag for :meth:`btree.keys`, :meth:`btree.values`, :meth:`btree.items` methods to specify that
```
koxudaxi, line 159: I faced an error when building with Sphinx because it conflicted with `string.templatelib.Template.values`. To solve this problem, I renamed the values attribute in btree to be more explicit.

koxudaxi:
@dpgeorge 
I'm not very familiar with the micropython codebase, so please let me know if you notice any issues with how I'm handling memory constraints. I think everything looks good, but I'd appreciate your review.

py/objtype.c
```diff
@@ -542,7 +542,18 @@ const byte mp_binary_op_method_name[MP_BINARY_OP_NUM_RUNTIME] = {
 };
 
 static mp_obj_t instance_binary_op(mp_binary_op_t op, mp_obj_t lhs_in, mp_obj_t rhs_in) {
+    #if MICROPY_PY_TSTRINGS
+    if (op == MP_BINARY_OP_ADD || op == MP_BINARY_OP_INPLACE_ADD) {
```
dpgeorge: This whole bit can be removed, it's not needed.  There is already a check in `template_binary_op()`.

ports/unix/coverage.c
```diff
@@ -822,6 +847,356 @@ static mp_obj_t extra_coverage(void) {
             MICROPY_STACK_CHECK == 0 || old_stack_limit == new_stack_limit);
     }
 
+
+    #if MICROPY_PY_TSTRINGS
```
dpgeorge: I'm surprised all of this coverage code is needed here, in C.

You should be able to move most (if not all) of this to pure Python tests.  Did you have trouble doing that?

py/modtstring.c
```diff
@@ -0,0 +1,600 @@
...
+                    vstr_init_print(&str_vstr, 16, &str_print);
+                    mp_obj_print_helper(&str_print, value, PRINT_STR);
+                    conv_value = mp_obj_new_str_from_vstr(&str_vstr);
+                } else if (*conversion_str == 'a') {
+                    vstr_t repr_vstr;
+                    vstr_init(&repr_vstr, 16);
+                    mp_print_t repr_print;
+                    vstr_init_print(&repr_vstr, 16, &repr_print);
+                    mp_obj_print_helper(&repr_print, value, PRINT_REPR);
+                    conv_value = mp_obj_new_str_from_vstr(&repr_vstr);
```
dpgeorge: These 3 cases for r/s/a all look the same.  Can you factor the code to reduce code duplication?

py/modtstring.c
```diff
@@ -0,0 +1,600 @@
...
+
+            if (interps->len <= 4) {
+                mp_obj_t values[4];
+                for (size_t i = 0; i < interps->len; i++) {
+                    mp_obj_interpolation_t *interp = MP_OBJ_TO_PTR(interps->items[i]);
+                    values[i] = interp->value;
+                }
+                dest[0] = mp_obj_new_tuple(interps->len, values);
+            } else {
+                mp_obj_t *values = m_new(mp_obj_t, interps->len);
```
dpgeorge: You can simplify the code here and not have any temporary memory by the following:
```c
mp_obj_tuple_t *tuple = MP_OBJ_TO_PTR(mp_obj_new_tuple(interps->len, NULL));
for (size_t i = 0; i < interps->len; i++) {
    tuple->items[i] = interp->value;
}
dest[0] = MP_OBJ_FROM_PTR(tuple);
```

py/modtstring.c
```diff
@@ -0,0 +1,600 @@
...
+            m_del(mp_obj_t, new_strings_items, new_strings_len);
+            m_del(mp_obj_t, new_interps_items, new_interps_len);
+
+            mp_obj_template_t *result = mp_obj_malloc(mp_obj_template_t, &mp_type_template);
+            result->strings = new_strings_tuple;
+            result->interpolations = new_interps_tuple;
+            return MP_OBJ_FROM_PTR(result);
+        }
+
+        case MP_BINARY_OP_REVERSE_ADD: {
```
dpgeorge, line 338: You can just remove this `case`, it's not needed.

py/parse.c
```diff
@@ -626,6 +643,379 @@ static void push_result_token(parser_t *parser, uint8_t rule_id) {
...
+        }
+
+        // Estimate memory usage and check against limit
+        size_t est_total = est_seg_cnt + est_interp_cnt;
+        size_t est_bytes = sizeof(mp_parse_node_struct_t) + sizeof(mp_parse_node_t) * est_total;
+        // Add overhead for dynamic arrays during parsing
+        est_bytes += est_seg_cnt * sizeof(mp_parse_node_t) * 2; // Assume some growth
+        est_bytes += est_interp_cnt * sizeof(mp_parse_node_t) * 2;
+
+        if (est_bytes > MICROPY_PY_TSTRING_MAX_BYTES) {
```
dpgeorge: I don't think this test is needed, what's wrong with the tstring growing large?

dpgeorge:
I've made a few comments, mostly regarding coverage.

About the `MICROPY_PY_TSTRING_MAX_xxx` constants: are these really needed?  What's wrong with just letting the t-strings grow as much as the user needs them to?  (Note: if memory runs out then the runtime will already automatically raise an exception, that the memory allocation did not succeed.)

docs/library/btree.rst
AJMansfield: It seems like a useful change, then --- it's just not a change related to PEP-750 t-string support, so it needs to be a separate commit and separate PR.
Git `cherry-pick` and `rebase -i` are your friends!

py/lexer.c
```diff
@@ -355,10 +357,10 @@ static void parse_string_literal(mp_lexer_t *lex, bool is_raw, bool is_fstring)
             while (is_fstring && is_char(lex, '{')) {
                 next_char(lex);
                 if (is_char(lex, '{')) {
-                    // "{{" is passed through unchanged to be handled by str.format
+                    // "{{" is passed through unchanged for f-strings
                     vstr_add_byte(&lex->vstr, '{');
                     next_char(lex);
-                } else {
+                } else if (is_fstring) {
```
AJMansfield: This seems to be redundant with the `while (is_fstring` condition.

docs/library/micropython.rst
```diff
@@ -79,6 +79,9 @@ Functions
    Return an integer representing the current amount of stack that is being
    used.  The absolute value of this is not particularly useful, rather it
    should be used to compute differences in stack usage at different points.
+   
+   Note: When building with sanitizers (ASan/UBSan), an 8 KiB stack margin is
+   automatically reserved unless the port overrides MICROPY_STACK_CHECK_MARGIN.
```
AJMansfield: Documentation change not related to PEP-750.

ports/unix/coverage.c
```diff
@@ -19,6 +19,24 @@
...
+extern const mp_obj_fun_builtin_fixed_t mp_builtin___template___obj;
+extern const mp_obj_type_t mp_type_template;
+extern const mp_obj_type_t mp_type_interpolation;
+extern mp_obj_t mp_obj_new_interpolation(mp_obj_t value, mp_obj_t expression, mp_obj_t conversion, mp_obj_t format_spec);
+
+typedef struct _mp_obj_template_t {
+    mp_obj_base_t base;
+    mp_obj_t strings;
+    mp_obj_t interpolations;
+} mp_obj_template_t;
```
AJMansfield: If you need all of these symbols just to test this, these symbols probably shouldn't be private.

ports/unix/mpconfigport.h
```diff
@@ -134,6 +134,11 @@ typedef long mp_off_t;
 #define MICROPY_STACKLESS_STRICT    (0)
 #endif
 
+// Reserve extra C-stack headroom for overflow checks.
+// Sanitizer builds enlarge call frames; 8 KiB prevents
+// false positives without noticeably shrinking usable stack.
+#define MICROPY_STACK_CHECK_MARGIN  (8192)
+
```
AJMansfield: This seems like a tooling change that's unrelated to PEP-750.

Though if I'm wrong about that, this needs more than just this comment to justify the claim that the sanitizer errors this bypasses are actually "false" positives.

py/lexer.c
```diff
@@ -109,6 +103,12 @@ static bool is_following_odigit(mp_lexer_t *lex) {
     return lex->chr1 >= '0' && lex->chr1 <= '7';
 }
 
+#if MICROPY_PY_FSTRINGS
+static bool is_char_or4(mp_lexer_t *lex, byte c1, byte c2, byte c3, byte c4) {
+    return lex->chr0 == c1 || lex->chr0 == c2 || lex->chr0 == c3 || lex->chr0 == c4;
+}
+#endif
+
```
AJMansfield: Move this up next to `is_char_or` and `is_char_or3`.
Also, this function doesn't need a macro guard -- it already has `static` linkage, so the compiler can already elide it if it's not used anywhere else in the file.

py/lexer.c
```diff
@@ -393,7 +395,7 @@ static void parse_string_literal(mp_lexer_t *lex, bool is_raw, bool is_fstring)
                         vstr_add_byte(&lex->fstring_args, c);
                         next_char(lex);
                     }
-                    if (lex->fstring_args.buf[lex->fstring_args.len - 1] == '=') {
+                    if (lex->fstring_args.len > 0 && lex->fstring_args.buf[lex->fstring_args.len - 1] == '=') {
```
AJMansfield: This should be part of a separate commit.

Also, I think this is probably unnecessary. I'll not go through the whole proof, but it should be possible to statically prove that `lex->fstring_args.buf[lex->fstring_args.len - 1]` is always a valid reference even when the length is 0, as a consequence of the fact that we only ever arrive in this parsing case if `fstring_args.buf[-1]` is validly `'{'` anyway.

If the compiler is smart enough to find that proof and elide this check to +0 code size, I see no reason to not have it --- but otherwise there's a discussion to be had here.

py/lexer.c
```diff
@@ -520,10 +522,9 @@ static void parse_string_literal(mp_lexer_t *lex, bool is_raw, bool is_fstring)
         }
         next_char(lex);
     }
-
-    // check we got the required end quotes
     if (n_closing < num_quotes) {
         lex->tok_kind = MP_TOKEN_LONELY_STRING_OPEN;
+        return;  // Exit early to prevent further processing
```
AJMansfield: Early-return isn't forbidden, but this is an unnecessarily confusing way of skipping a single line of code --- just wrap the the `vstr_cut_tail_bytes` operation that's being skipped in an `else` block.

This should be part of a separate commit, and needs some sort of test or explanation to justify _why_ 'prevent further processing' is the correct behavior.

py/parse.h
```diff
@@ -46,6 +46,29 @@ struct _mp_lexer_t;
...
+#define TSTR_HDR_MAKE(seg_cnt, int_cnt) \
+    ((MP_PARSE_NODE_TEMPLATE_STRING | \
+    ((((seg_cnt) - 1) << TSTR_HDR_SEG_SHIFT)) | \
+    (((int_cnt) << TSTR_HDR_INT_SHIFT))))
+
+#define TSTR_HDR_GET_SEG_CNT(hdr) \
+    (((((hdr) >> TSTR_HDR_SEG_SHIFT) & UINT32_C(0xFFF)) + 1))
+
+#define TSTR_HDR_GET_INT_CNT(hdr) \
+    ((((hdr) >> TSTR_HDR_INT_SHIFT) & UINT32_C(0xFFF)))
```
AJMansfield: Symbols in public headers need the appropriate name prefixes --- these macros should probably be `MP_PARSE_TSTR_(whatever)`.

py/parse.h
```diff
@@ -46,6 +46,29 @@ struct _mp_lexer_t;
...
+
+// Ensure header bit packing fits in 32 bits
+typedef char _tstr_assert1[(TSTR_HDR_INT_SHIFT + 12 <= 32) ? 1 : -1];
+typedef char _tstr_assert2[(TSTR_MAX_SEG <= 0xFFF) ? 1 : -1];
+typedef char _tstr_assert3[(TSTR_MAX_INT <= 0xFFF) ? 1 : -1];
```
AJMansfield: Interesting idiom, I wasn't aware you could do static asserts like that!
But you should use `MP_STATIC_ASSERT` (from `misc.h`) instead.

py/qstrdefs.h
```diff
@@ -76,3 +76,26 @@ Q(/rom/lib)
...
+Q(string)
+Q(templatelib)
+Q(values)
+Q({)
+Q(})
+Q(__template__)
+Q(__lt_tstring_expr_gt_)
+Q(__iter__)
+Q(template_iterator)
+#endif
```
AJMansfield: This is unnecessary, these should all already be picked up automatically from any `MP_QSTR_*` symbols, during the qstr preprocessing build step. (If you're getting IDE errors it's because you haven't set your include path to include the generated includes folder in the build output directory.)

py/runtime.h
```diff
@@ -169,6 +169,13 @@ static inline void mp_thread_init_state(mp_state_thread_t *ts, size_t stack_size
     ts->nlr_jump_callback_top = NULL;
     ts->mp_pending_exception = MP_OBJ_NULL;
 
+    #if MICROPY_PY_SYS_SETTRACE
+    // Initialize profiling state
+    ts->prof_callback_is_executing = false;
+    ts->prof_trace_callback = MP_OBJ_NULL;
+    ts->current_code_state = NULL;
+    #endif
+
```
AJMansfield, line 190: This is a tooling change unrelated to PEP-750.

tests/thread/thread_stacksize1.py
```diff
@@ -7,7 +7,11 @@
 
 # different implementations have different minimum sizes
 if sys.implementation.name == "micropython":
-    sz = 2 * 1024
+    # Use larger stack size for desktop platforms to accommodate sanitizers
+    if sys.platform in ("linux", "darwin", "emscripten", "win32"):
+        sz = 32 * 1024
+    else:
+        sz = 2 * 1024
```
AJMansfield: This is a tooling change unrelated to PEP-750.

py/lexer.c
```diff
@@ -567,9 +568,17 @@ void mp_lexer_to_next(mp_lexer_t *lex) {
...
+        if (lex->fstring_args.len >= 3) {
+            lex->chr0 = lex->fstring_args.buf[0];
+            lex->chr1 = lex->fstring_args.buf[1];
+            lex->chr2 = lex->fstring_args.buf[2];
+        } else {
+            // This should not happen with well-formed f/t-strings
+            lex->chr0 = lex->fstring_args.len > 0 ? (unichar)lex->fstring_args.buf[0] : MP_LEXER_EOF;
+            lex->chr1 = lex->fstring_args.len > 1 ? (unichar)lex->fstring_args.buf[1] : MP_LEXER_EOF;
+            lex->chr2 = lex->fstring_args.len > 2 ? (unichar)lex->fstring_args.buf[2] : MP_LEXER_EOF;
+        }
```
AJMansfield: This should be part of a separate commit, and needs a non-tstring test.

AJMansfield (changes requested):
This PR needs to be rebased into several distinct commits so that it can be properly reviewed. This is a very complicated set of changes that adds a lot of potential surface for errors, requiring careful scrutiny --- and the way it mixes concerns does not inspire confidence.

To make this easier to review, please split your commit into multiple commits. If you still have a non-squashed head you can submit instead, that'd be helpful; and especially if not, you're going to need to learn to love `git cherry-pick` and `git rebase -i`.

I'd consider several of the changes in this PR to still be untested _despite_ appearing to have 100% test coverage. 100% coverage is just the bare minimum --- the standard to aspire to is 100% test sensitivity, i.e. tests don't just have to pass when the change is present, they also are also _required to fail_ when the change in question is absent. (100% test sensitivity is, in general, not a possible goal; but it's still an essential tool for verifying things that are a matter of correctness rather than just style.)

Several of the changes seem to be fixes to string parsing generally, beyond just t-strings. It's not surprising that you might discover latent string-parsing bugs while implementing something like this, but those fixes need to be split into their own commit. I also want to see test sensitivity on those changes, or at least better explanations why.

Several of the changes are tooling changes: the `MICROPY_STACK_CHECK_MARGIN` and related changes that seem to be meant to accommodate running micropython with a sanitizer. Sanitizers are definitely helpful on features like this, and if those are changes that allow some sanitizer that previously didn't work on micropython _at all_ to be used, that's potentially quite useful and should be submitted as its own PR. And if they're specific to bypassing errors introduced by _this_ feature, this needs more discussion here.

I'd also appreciate separate commits for test code vs feature code, so that it's more straightforward to verify test sensitivity --- Doing it with separate commits makes it easy to just checkout the test commit on its own and run the suite from there.

All of the tests have `.exp` files associated with them --- it's understandable since this is a syntax feature not present in most of the CPython versions used for testing, either.
Having those as part of the same commit makes it less convenient for testing, though --- please separate out the `.exp` files that don't represent actual micropython-specific behaviors and put them in a separate commit from their tests.

koxudaxi:
@dpgeorge @AJMansfield 
Thank you for the code review. I'll address the pointed out issues and clean up the commits. 
I'm planning to work on this over the weekend.

AJMansfield:
Thanks for putting in the effort @koxudaxi, this will be a useful feature to have!

py/parse.h
koxudaxi: @AJMansfield 
Thanks for the suggestion! I tried using `MP_STATIC_ASSERT`, but it  causes build errors at file scope:
```
error: expected declaration specifiers or '...' before 'sizeof'
```
Looks like `MP_STATIC_ASSERT` expands to an expression, not a declaration, so it can't be used outside functions.
Any ideas how to work around this?

py/lexer.c
koxudaxi: After analyzing the code, I found that the boundary check is actually unnecessary. 
The fstring_args buffer always contains at least ".format()" (9 characters) even for empty f-strings, making the 3-character access always safe.
I've removed this dead code in a new commit.

dpgeorge:
@koxudaxi thank you for working on this template string implementation.  What is the current status from your side?

If it's ready for a review then please rebase on latest master (there are some conflicts that need resolving).

koxudaxi:
@dpgeorge 
Sorry,I was busy and I did not get the time for PR. I will update the PR on this week or next week.

WebReflection:
FYI (and you all know already) ... Python 3.14 with template strings has been released, it'd be awesome to have that feature backed into MicroPython *next* but I've seen a lot of back and forward around commits in here ... is the CI pipe to blame, as impossible to reproduce locally, or something else? Happy to have a look at anything I could help with, looking forward to see this merged for the best next version of MicroPython to date 🥳

koxudaxi:
Sorry for the delay. I'm writing boundary tests (like memory limits) to improve test coverage, but the results vary slightly depending on the CI execution environment. In the last run, only some Windows 64-bit tests and Unix 32-bit tests failed. I am testing locally on Linux, but checking all environments isn't really practical 😓

dpgeorge:
@koxudaxi if you rebase on latest master, it'll fix the zephyr CI.

py/parse.h
AJMansfield: If there's really no particular obvious function to put those asserts in, you can define a `static inline` function just for the sake of it:

```C
static inline void _tstr_asserts() {
    MP_STATIC_ASSERT(TSTR_HDR_INT_SHIFT + 12 <= 32);
    MP_STATIC_ASSERT(TSTR_MAX_SEG <= 0xFFF);
    MP_STATIC_ASSERT(TSTR_MAX_INT <= 0xFFF);
}
```

With `static`, a body that's free from any kind of effect (side or otherwise), and never taking the function's address, this will give 0 additional code gen, not even a call site that just returns; and `inline` ensures you don't get a `warning: ‘void _tstr_asserts()’ defined but not used [-Wunused-function]` for never calling it, either.

AJMansfield:
Feel free to ping me once you've finished working through this to a state that's ready to be re-reviewed.

WebReflection:
@kumekay are you paying attention to hints? if you need any help at all, mind stating what is it?

I am following pushes and I feel like you are stuck in some testing loop and there's zero shame in that ... but maybe others can help with? Surely not helping your case but I see tons of pushes daily and I kinda sympathize with your possible frustration, specially when I see tests added or removed to make CI green .... maybe there's a better way, as everything else feels right in place so that asking "*how do I make CI happy in there?*" might be a faster way out.

kumekay:
> @kumekay are you paying attention to hints? if you need any help at all, mind stating what is it?
> 
> I am following pushes and I feel like you are stuck in some testing loop and there's zero shame in that ... but maybe others can help with? Surely not helping your case but I see tons of pushes daily and I kinda sympathize with your possible frustration, specially when I see tests added or removed to make CI green .... maybe there's a better way, as everything else feels right in place so that asking "_how do I make CI happy in there?_" might be a faster way out.

@WebReflection 
You probably need wanted to mention @koxudaxi here ;)

WebReflection:
@kumekay apologies and thank you for bringing my comment to @koxudaxi attention ... too many `@k`, my mistake 😅

koxudaxi:
@WebReflection 
Thank you for your kind comment.
I'm indeed struggling a bit, but I'm fine-tuning the detailed behavior to match CPython and PEP 750 as closely as possible while also aligning with MicroPython's design principles.

I think I've squashed most of the bugs, and what remains is the module path issue.

@AJMansfield @dpgeorge 
I'd like to discuss the module structure.
In t-strings, we need to provide the string.templatelib namespace from which classes like Template and Interpolation are imported.
`from string.templatelib import Template`
In CPython, this is provided as a Python file in the stdlib, but I'm wondering what the best way to arrange this in MicroPython would be. I was thinking of placing string.templatelib as a builtin, but I realized that if someone imports string from micropython-lib, it would overwrite these attributes. Is there a good approach for this?
I also considered moving templatelib entirely to micropython-lib, but then I realized this could complicate the build configuration.

https://docs.python.org/3/library/string.templatelib.html
https://github.com/python/cpython/blob/95a3564869daa3cc083e4d4603f5f03ee5f53e31/Lib/string/templatelib.py#L5

I believe everything else should be fine, so I'd appreciate another review when you have a chance.

AJMansfield:
> I'd like to discuss the module structure. In t-strings, we need to provide the string.templatelib namespace from which classes like Template and Interpolation are imported. `from string.templatelib import Template` In CPython, this is provided as a Python file in the stdlib, but I'm wondering what the best way to arrange this in MicroPython would be. I was thinking of placing string.templatelib as a builtin, but I realized that if someone imports string from micropython-lib, it would overwrite these attributes. Is there a good approach for this? I also considered moving templatelib entirely to micropython-lib, but then I realized this could complicate the build configuration.

Normally, implementing a module as frozen mpy bytecode will actually be smaller than the equivalent implementation in C. It's also a common pattern to define an underscore-prefixed module in C to provide a minimal set of primitives to access some interpreter feature or api or perform the performance-critical hot loop of a package, and then implement the rest of that module in python. For an example of this, take a look at how the `asyncio` module is implemented. `extmod/modasyncio.c` doesn't actually define a python-api-compatible `asyncio` package, it just defines `_asyncio` to give C definitions for `Task` and `TaskQueue`, and then implements everything else in a python module --- `extmod/asyncio/__init__.py` and its siblings.

Instead of defining the recursive module structure manually in `modstring.c` and special-casing the import handling, this package structure could be implemented in python code, with a `extmod/string/__init__.py`, a `extmod/string/templatelib.py`, and then a `modtstring.c` that exports a bare `_tstring` module that exports the `mp_obj_interpolation_t` and `mp_obj_template_t` types.


Also, given the massive code size increase at stake here, it might even be worth considering if there's a way to move even more of the parse logic into something that can run from bytecode --- perhaps even, for some ports, as an externally mip-installable component rather than something to freeze in. For instance, have MicroPython's parser just minimally recognize `t"(whatever)"`, and then invoke something along the lines of:
```C
static mp_obj_t tstring_parse_func = NULL;
mp_obj_t tstring_parse(vstr_t *str) {
    if (tstring_parse == NULL) {
        mp_obj_t modtstring = mp_builtin___import__(1, &MP_OBJ_NEW_QSTR(MP_QSTR_mp_tstring_support));
        // NLR this to a "tstring support not installed" error? Maybe with directions to run `mip.install("mp_tstring_support")`?
        tstring_parse_func = mp_obj_dict_get(mp_obj_module_get_globals(modtstring), MP_QSTR_parse);
    }
    return mp_call_function_1(tstring_parse_func, mp_obj_new_str_from_vstr(str));
}
```

koxudaxi:
@AJMansfield 
Thank you for the detailed comment. I applied your suggestions. Could you please review it?
Regarding the parser implementation, I had a misunderstanding. I thought writing it in C was better aligned with MicroPython's approach.
However, I tried to follow the f-strings implementation. Since we have enough tests, rewriting the parser in Python as you suggested is feasible.
About the mip-installable approach, my original plan was making t-strings a build option to avoid size increase for unnecessary targets. Could you clarify your suggestion?
 Should it be mip-installable instead of a build option, or in addition to it?
 
 (If I understand correctly, no changes to `micropython-lib` are needed.)

AJMansfield:
> Thank you for the detailed comment. I applied your suggestions. Could you please review it?

Will do! Hopefully I'll be able to get to this on Monday.

> I thought writing it in C was better aligned with MicroPython's approach.

That's my assumption as well --- I wouldn't put too much stock in my musings on that idea, for whether or not it might be possible to let portions of the parser operate from bytecode. If that's anything at all, it's certainly not something I'd put in the way of the current version being merged.

koxudaxi:
@AJMansfield 
I've reorganized the commits based on your previous feedback. If the commit granularity is still too coarse or inappropriate, please let me know and I'll revise them.

AJMansfield:
Apologies for the delay reviewing, finally found the time though. And that's a well-organized set of commits I'll have no trouble examining.

tests/feature_check/tstring.py.exp
AJMansfield, line 1: This exp should be in the commit that adds the feature check; it's needed to even run the tstring tests at all.

tests/micropython/tstring_validation.py
AJMansfield, line 1: This should probably be a `micropython` test rather than a `basics` test, as `__template__` is a micropython-specific implementation detail.

py/modbuiltins.c
```diff
@@ -757,6 +762,9 @@ static const mp_rom_map_elem_t mp_module_builtins_globals_table[] = {
     { MP_ROM_QSTR(MP_QSTR_ViperTypeError), MP_ROM_PTR(&mp_type_ViperTypeError) },
     #endif
     { MP_ROM_QSTR(MP_QSTR_ZeroDivisionError), MP_ROM_PTR(&mp_type_ZeroDivisionError) },
+    #if MICROPY_PY_TSTRINGS
+    { MP_ROM_QSTR(MP_QSTR___template__), MP_ROM_PTR(&mp_builtin___template___obj) },
```
AJMansfield: What's the reasoning for putting this at the end, rather than with the other double-underscore builtins? (Note that `__print_repl__` is actually also a micropython-specific builtin.)

AJMansfield (changes requested):
On examination of the code and after running it on my machine, I can say that the functional code in this PR seems substantively correct, but there's a few details to still clean up especially with regards to testing.

Most of the tests here do not pass when run in comparison with a real CPython 3.14 instance --- but with only a small number of lines differing, on error behavior and other areas where differences are commonly accepted. Usually, when testing a feature that's broadly compatible but with a minor deviation, it's a common pattern to separate out the compatible and incompatible cases between e.g. `some_test.py` and `some_test1.py`, so that you can provide a `some_test1.py.exp` without removing the cross-compare for all of it.

It's reasonable enough to also be using `.exp` files here to bridge the gap to 3.14-less testing environments, but I'd like to still see these tests split into separate fully-3.14-compatible and non-3.14-compatible test files. Note that the `.exp`'s for the non-3.14-compatible ones should go in the same commit with the tests --- i.e. it should be a complete commit with everything needed to run and pass a comparison in an environment that _does_ have the right CPython, with the follow-up commit containing only exactly the files generated by running e.g. `python3.14 some_test.py &> some_test.py.exp` for the relevant tests.

(As a related note, we've actually been needing a mechanism to address this scenario for a while --- ever since CPython changed tracing formats making MicroPython's outputs on the tests for `sys.settrace` no longer match on newer versions. If you'd be interested in taking on adding that --- maybe a table similar to `tests_with_regex_output` that gives CPython version ranges to use or disregard exp files on --- it would be quite useful indeed to have. If not though, I'll be satisfied with just having the tests split.)

py/modbuiltins.c
koxudaxi: I've moved `__template__` to group it with the other double-underscore builtins.

tests/feature_check/tstring.py.exp
koxudaxi: The feature-check exp file is now part of the same commit as tests/feature_check/tstring.py so the test works immediately

tests/micropython/tstring_validation.py
koxudaxi: The validation test has been moved into tests/micropython so the template coverage is clearly  marked as MicroPython-specific.

koxudaxi:
Regarding the behavioral differences, I recall changing the wording due to resource constraints on 32-bit systems.
I'll review and organize the behavioral differences from CPython at the test level. Please bear with me.

tests/basics/tstring_basic1.py
```diff
@@ -0,0 +1,143 @@
...
+print("=== Parser error tests ===")
+try:
+    exec('t_empty = t"{}"')
+    print(f"Empty expr: {t_empty.interpolations[0].value}")
+except SyntaxError as e:
+    print(f"Empty expr: SyntaxError - {e}")
+
+# Whitespace in expression (MicroPython preserves, CPython strips trailing)
+t_ws = t"{ 42 }"
+print(f"Whitespace: {str(t_ws)}")
```
koxudaxi, line 22: This behaves differently from CPython, but keeping it unchanged for now since CPython may have a bug.
Currently checking with Core developers to confirm if it's a bug.

koxudaxi:
@AJMansfield 
I've addressed your feedback and reorganized the commits. Could you please review it again?

koxudaxi:
After taking another look at this PR for the first time in a while, I noticed that coverage was insufficient due to some dead code that had been left behind, so I removed it. However, there now appear to be errors in the Windows environment and the esp32 port that seem unrelated to this change.
Please review when you have time.

dpgeorge:
@koxudaxi I had a good look at this PR and how it parses t-strings.

From what I understand, it creates a new nested parser when it encounters a t-string, in order to parse the expression parts of the t-string.

That seems rather complicated and probably unnecessary, at least for MicroPython's implementation of t-strings.

As far as I understand, t-strings are syntatically equivalent to f-strings (except that the f prefix is a t).  So I'm wondering if it would be possible to reuse the f-string parser that we already have?

The way f-strings work at the moment is that the lexer dynamically transforms the incoming f-string bytes into a new set of bytes for a corresponding str-format expression, like this:
```
f"ab{12}cd{34}"  -->  "ab{}cd{}".format(12, 34)
```
And then the lexer itself tokenizes the transformed bytes.  So the parser only ever sees the RHS of the above, the parser never sees any f-strings.

Could we make it so t-strings worked the same way?  Something like:
```
t"ab{12}cd{34}"  -->  __template__("ab", 12, "cd", 34, "")
```
That could reuse most of the code for f-string parsing, and make the implementation minimal and quite efficient.

That doesn't support nested t-strings, but I don't think we need that.  We don't support nested f-strings.

WebReflection:
@dpgeorge 

> That doesn't support nested t-strings, but I don't think we need that. We don't support nested f-strings.

... ouch ... for the most common/popular use case, which is X/HTML representation, does that mean that this code would not work neither, because you cannot have an `f` string within a `t` one?

```python
div = html(t"<div>{f"Hello {user}!"}</div>")
```

This might be a blocker in terms of ergonomics / capabilities, but maybe it won't be the end of the world (as in: one simply factors out via lambdas all partial `f` or `t` and call it a day), thanks!

koxudaxi:
@dpgeorge 
Thanks for looking at this in detail!
A clarification on "nesting", based on recent discussion around HTML/tdom-style usage with @WebReflection.
There are (at least) two different things people call "nesting":
1) using an f-string or t-string literal as a normal expression inside a t-string interpolation, e.g. `t"<div>{f\"Hello {user}!\"}</div>"` or `t"outer { t'inner {x}' }"` (tdom relies on this pattern)
2) nesting inside format specs, e.g. `t"{value:{width}.{precision}f}"`
For tdom/HTML composition, (1) is the non‑negotiable case, while (2) is nice-to-have at best.

Even if we ignore (2) entirely, I don’t think a simple lexer-rewrite approach that only scans for delimiters (similar to the current f-string handling) is robust enough for (1), because inner `{...}` from the f/t-string literal can interfere with outer boundary detection. That’s why this PR treats the `{...}` portions as real expressions and parses them accordingly.
I understand the RAM/flash concerns, especially around retaining `Interpolation.expression`. To keep this “micro”, I’m limiting enablement to larger ports (unix/windows/pyscript) for now. If we still want this on more constrained targets, we could also gate `expression` retention behind a build flag/config option.

dpgeorge:
> 2\. nesting inside format specs, e.g. `t"{value:{width}.{precision}f}"`

Actually, MicroPython does already support this form of nesting in f-strings!  We just had a test added for that in #18495.  So it's not out of the question that f-strings can support nested f-strings using the current lexer-rewrite approach.  And by extension, nested t-strings could also work with that approach.

Eg the lexer would have both an `inject` and `fstring_args` vstr, where `inject` holds characters that should be injected into the stream, while `fstring_args` holds characters that are currently being extracted from the f-string.  Then when the f-string parsing is done (by the lexer) it copies `fstring_args` over to `inject` (possibly inserting it at the current location in `inject` to handle the nesting scenario), and then clears `fstring_args` to be ready for the next one.

> I understand the RAM/flash concerns, especially around retaining `Interpolation.expression`. To keep this “micro”, I’m limiting enablement to larger ports (unix/windows/pyscript) for now. If we still want this on more constrained targets, we could also gate `expression` retention behind a build flag/config option.

IMO the `.expression` attribute has limited use, and would consume quite a bit of RAM.  So might be good to have a compile-time option to turn that off.

dpgeorge:
For the record, the f-string feature currently costs about 470 bytes (on ARM Cortex-M), and this t-string implementation costs about 7400 bytes (on ARM Cortex-M, including the frozen code for `string.templatelib`).

dpgeorge:
In #18588 I modified the existing f-string parser to support nested f-strings, with nesting to arbitrary depth.  I think the same algorithm could now be used for a t-string parser.

dpgeorge:
In #18650 I've provided a proof-of-concept implementation of t-strings which reuses the f-string lexer-rewrite technique, and supports nested t-strings and f-strings.  It's not finished yet, but I think it can implement all required features of t-strings.

koxudaxi:
@dpgeorge 
Sorry for the delayed response. Thank you for providing the proof-of-concept implementation. Over the holidays, I also started working on modifications to reuse the f-string lexer based on your advice, but I haven't pushed it yet since it's still incomplete. I'll take another look at your WIP branch after work.

dpgeorge:
> Over the holidays, I also started working on modifications to reuse the f-string lexer based on your advice, but I haven't pushed it yet since it's still incomplete. I'll take another look at your WIP branch after work.

Oh, very good!  Hopefully we didn't duplicate too much effort here :)

I did reuse a lot of your code from this PR for the `Template` and `Interpolation` classes.  And I reused your tests.

You'll also see that I implemented the `string.templatelib` module entirely in C, which I think is a bit simpler than having to freeze in some Python code for it.

koxudaxi:
@dpgeorge 
No worries about the duplication!
I looked at my work-in-progress code and while it reuses the f-string functions, it uses a two-pass approach rather than a single pass.
Your PR has a much cleaner integration with f-strings.
As for the test cases, I carefully verified them against CPython and the PEP, so hopefully they are useful for your implementation as well.

dpgeorge:
@koxudaxi thanks to your efforts here I have taken your tests, docs and the code for `Template` and `Interpolation` classes and made them part of #18650.  That PR is now finished and I think in a good state to replace this one.  It's a lot smaller in its implementation, doesn't place any limitations/restrictions on the size of t-strings, and provides almost all the same features as this PR (see the top post in #18650 for details of that).

koxudaxi:
@dpgeorge Thanks for finishing this up and for using the tests and code! Glad my work was helpful to the project. I'll close this PR.

PR #17572, opened 2025-06-27
py/emitnative: Let emitters know the compiled entity's name.

agatti:
### Summary

This PR introduces an optional feature to provide to native emitters the fully qualified name of the entity they are compiling.

This is achieved by altering the generic ASM API to provide a third argument to the entry function, containing the name of the entity being compiled.  Currently only the debug emitter uses this feature, as it is not really useful for other emitters for the time being; in fact the macros in question just strip the name away.

As discussed in #17548, this brings more descriptive output to the debug emitter, in preparation for a new machine-readable JSON output emitter that would need to know the entity name being processed as well.

### Testing

Testing had to be performed manually, by checking the output of `mpy-cross -X emit=native -march=debug` for a variety of existing MicroPython test files containing all possible entities to be handled by the compiler framework along with their nested variants.  The usual battery of CI tests was also ran through Github to be sure nothing out of the ordinary would happen.

In theory there shouldn't be any changes in any other emitter that actually generates binary code, this code should only affect `mpy-cross`.

### DIFF
--- a/py/emitnative.c
+++ b/py/emitnative.c
@@ -419,6 +419,30 @@ static void emit_native_start_pass(emit_t *emit, pass_kind_t pass, scope_t *scop
         emit->stack_info[i].vtype = VTYPE_UNBOUND;
     }
 
+    char *qualified_name = NULL;
+
+    #if N_DEBUG
+    scope_t *current_scope = scope;
+    vstr_t *qualified_name_vstr = vstr_new(qstr_len(current_scope->simple_name));
+    size_t fragment_length = 0;
+    const byte *fragment_pointer;
+    for (;;) {
+        fragment_pointer = qstr_data(current_scope->simple_name, &fragment_length);
+        vstr_hint_size(qualified_name_vstr, fragment_length);
+        memmove(qualified_name_vstr->buf + fragment_length, qualified_name_vstr->buf, qualified_name_vstr->len);
+        memcpy(qualified_name_vstr->buf, fragment_pointer, fragment_length);
+        qualified_name_vstr->len += fragment_length;
+        if (current_scope->parent == NULL || current_scope->parent->simple_name == MP_QSTR__lt_module_gt_) {
+            break;
+        }
+        vstr_ins_char(qualified_name_vstr, 0, '.');
+        current_scope = current_scope->parent;
+    }
+    qualified_name = vstr_null_terminated_str(qualified_name_vstr);
+    #else
+    (void)qualified_name;
+    #endif
+
     mp_asm_base_start_pass(&emit->as->base, pass == MP_PASS_EMIT ? MP_ASM_PASS_EMIT : MP_ASM_PASS_COMPUTE);
 
     // generate code for entry to function
@@ -465,7 +489,7 @@ static void emit_native_start_pass(emit_t *emit, pass_kind_t pass, scope_t *scop
         }
 
         // Entry to function
-        ASM_ENTRY(emit->as, emit->stack_start + emit->n_state - num_locals_in_regs);
+        ASM_ENTRY(emit->as, emit->stack_start + emit->n_state - num_locals_in_regs, qualified_name);
 
         #if N_X86
         asm_x86_mov_arg_to_r32(emit->as, 0, REG_PARENT_ARG_1);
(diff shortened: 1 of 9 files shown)


py/emitnative.c
```diff
@@ -377,6 +377,43 @@ static void emit_native_mov_reg_qstr_obj(emit_t *emit, int reg_dest, qstr qst) {
         emit_native_mov_state_reg((emit), (local_num), (reg_temp)); \
     } while (false)
 
+#if MICROPY_EMIT_PROVIDE_ENTITY_NAMES
```
dpgeorge: I think it's fine to just use `MICROPY_EMIT_NATIVE_DEBUG` here.  No real need to add (yet another) configuration option for this.  And then it's almost self-documenting, that this code is primarily here for the native debug output.

py/emitnative.c
```diff
@@ -377,6 +377,43 @@ static void emit_native_mov_reg_qstr_obj(emit_t *emit, int reg_dest, qstr qst) {
...
+    size_t size = 0;
+    for (;;) {
+        size += qstr_len(current_scope->simple_name) + 1;
+        if (current_scope->parent == NULL || current_scope->parent->simple_name == MP_QSTR__lt_module_gt_) {
+            break;
+        }
+        current_scope = current_scope->parent;
+    }
+
+    char *buffer = m_malloc(size);
```
dpgeorge: I suggest using a `vstr_t` here, to make it much simpler to do the dynamic buffer alloc.  Then you only need to loop over the name once to build it, using `vstr_add_strn()`.  Then do `return vstr_null_terminated_str()` at the end.

In fact, it might even be simple enough code that you don't need to put it in a separate function.

py/emitnative.c
agatti: I put the define in there because the JSON output printer also needs that feature, but I can remove this.

py/emitnative.c
agatti: The `vstr_t` remark is valid, I'll rework memory management in that bit of code to follow that - thanks!

agatti: I've had to extend the vstr API just a bit to get there - there was no facility for inserting strings at arbitrary locations - but now it should look much cleaner.

py/emitnative.c
dpgeorge: I'd be happy to replace the native debug emitter with a native JSON emitter.  I don't think we need both.  But that's a discussion for follow-up PRs.

ports/unix/coverage.c
```diff
@@ -364,6 +364,11 @@ static mp_obj_t extra_coverage(void) {
         } else {
             mp_obj_print_exception(&mp_plat_print, MP_OBJ_FROM_PTR(nlr.ret_val));
         }
+
+        vstr_ins_str(vstr, 0, "\"tests\"");
```
dpgeorge: If this is just to test `vstr_ins_str` maybe it should clear the vstr to start with, instead of using `vstr_cut_tail_bytes`?

py/vstr.c
```diff
@@ -210,6 +210,12 @@ void vstr_ins_char(vstr_t *vstr, size_t char_pos, unichar chr) {
     *s = chr;
 }
 
+void vstr_ins_str(vstr_t *vstr, size_t str_pos, const char *str) {
```
dpgeorge: It looks like this addition is increasing code size for all ports, which is unexpected (due to gc sections).

Needs investigation.

ports/unix/coverage.c
agatti: There's no other `vstr_ins_*` function that's called with 0 as its insertion point in this test file.  Since all the insertion functions share the same extra prefix allocation bits I thought it'd be nice to cover that specific behaviour too.

But I can let it perform an insertion on an empty vstr too, or better yet cover both cases.

py/vstr.c
dpgeorge: Maybe for simplicity, just write this function inline where it's used (once)?

agatti: To get there I'd have to make `vstr_ins_blank_bytes` non-static, I'll check if that would increase the code footprint or not.

agatti: Unfortunately even a version with `vstr_ins_blank_bytes` non-static and the reworked string build loop seems to increase all ports' size by the same amount as the current iteration of the PR.

Considering the niche use case and the limited scope of the size/time overhead in the initial version, would it be OK to revert to that version (with a comment on why things are like that)?  It may not look as nice but it won't take up space where it shouldn't be.

dpgeorge: Ah, that's annoying.  I guess it's because the current `vstr_ins_blank_bytes()` is only ever called with a length of 1, so the compiler can optimise it.

I have two suggested options:
1. Use `vstr_hint_size()` to implement the insertion in `emitnative.c` (it's really just a call to that function, then a `memmove` then a `memcpy`).
2. Go back to the initial version but use `vstr_t` instead of explicitly managing the memory (so I guess it would have 2 loops, one to calculate the size and one to build the vstr).

agatti: Now things should be back to normal, I've opted to follow your first suggestion - thanks for that!

dpgeorge (approved):

dpgeorge:
Thanks for updating!

PR #17578, opened 2025-06-29
stm32: Change SPI IRQ priority.

yn386:
### Summary

On STM32H5/H7, SPI flash cannot use as storage device with DMA.
SPI interruption may not be genearated even if DMA transfer has been done.
This is due to lower priority of SPI interruption than DMA.

This PR changes SPI interrupt priority more higher than DMA's priority.

### Testing

Tested:
* NUCLEO-H563ZI with W25Q32BVSSIG
* Custom STM32H723VG board with W25Q64JVSIQ

There is no negative impact because only STM32H5/H7 HAL requires SPI IRQs to be enabled and handled.


dpgeorge (approved):
Thanks, this looks fine to me.

PR #17580, opened 2025-06-29
Various ports: add support for Timer hard= option

arachsys:
### Summary

7816b1f, 8f85eda and 8162468 (PR #17363) added a `hard=` option to the rp2 `Timer` constructor and initialiser to allow the callback to be switched between hard-IRQ and soft-IRQ mode.

The behaviour of other ports varies: either hard-IRQ only or soft-IRQ only. @dpgeorge suggested that it would be good to make this consistently configurable across all ports, preserving the current default but allowing both behaviours everywhere.

This patch series is an attempt to do that, adding support to every port that has a `machine.Timer` except nrf (see below), ensuring there is cross-port test coverage of both variants, and documenting as appropriate.

### Notes and issues

nrf is the only port with a `machine.Timer` that isn't covered by this patch series. This port doesn't have `mp_sched_schedule()` enabled so I don't think I can support `hard=False` there? I've left it out altogether, assuming that extra code size isn't worth it just to add an ignored `hard=True` option for consistency. (I also note that nrf doesn't support the `freq=` option to the `Timer` constructor/init, just `period=`, so it's already a cut-back timer implementation compared to other ports.)

I've added hard IRQ support to `shared/runtime/softtimer`, but have gated it to avoid bloat on platforms which have their own timer infrastructure and don't need hard IRQ support from the software timer. alif, mimxrt and samd are the three ports which turn this on.

### Testing

Although I have compile tested every patch and every port via the Micropython test infrastructure, I don't have hardware to test ports other than rp2. The generalised test I have written in `tests/extmod/machine_timer.py` would exercise old and new codepaths on every modified platform if run there.

I would greatly appreciate help in verifying that this series does indeed work correctly on every port it claims to cover! Obviously I will happily fix any issues that do show up.

For the purposes of writing tests:
- every port with `machine.Timer` except nrf supports both `hard=True` and `hard=False`
- every port with `machine.Timer` except nrf supports `freq=N` (`N` in Hz)
- every port with `machine.Timer` except renesas-ra supports `period=N` (`N` in ms)
- esp32 and esp8266 don't support software timers (`id = -1`)

To try to cover the various cases a bit more thoroughly, my new test covers all ports with a `machine.Timer` including esp32/esp8266 but excluding nrf, exercising hard and soft timers with varying `freq=` and explicitly checking that the heap is locked in hard callbacks and unlocked in soft callbacks. (The preexisting `tests/extmod/machine_soft_timer.py` already covers all the functionality available on nrf, and tests period= instead of freq=.)

### Additional consolidation

As suggested by @andrewleech, I've added an additional (somewhat bolder!) patch to show what consolidating the hard/soft callback from mp_irq_handler() with the various timer implementations would look like. The test suite suggests this gives small binary size improvements on ports like stm32 and rp2 as well as having a pleasingly negative diffstat.


arachsys:
Rebased onto HEAD of master.

ports/esp32/machine_timer.c
```diff
@@ -156,8 +157,28 @@ static void machine_timer_isr(void *self_in) {
...
+            self->repeat = 0;
+            mp_printf(MICROPY_ERROR_PRINTER, "uncaught exception in timer callback\n");
+            mp_obj_print_exception(MICROPY_ERROR_PRINTER, MP_OBJ_FROM_PTR(nlr.ret_val));
+        }
+        gc_unlock();
+        mp_sched_unlock();
+    } else {
+        mp_sched_schedule(self->callback, MP_OBJ_FROM_PTR(self));
+        mp_hal_wake_main_task_from_isr();
+    }
```
andrewleech: Perhaps this common snippet could be consolidated into the extmod machine timer C file for reuse by all the ports.

Though arguably it could be used by any hardware peripheral interrupts so perhaps it should be somewhere even more common?

tests/extmod/machine_timer.py
```diff
@@ -0,0 +1,51 @@
...
+    # The nrf port doesn't support selection of hard and soft callbacks,
+    # and only allows Timer(period=N), not Timer(freq=N).
+    print("SKIP")
+    raise SystemExit
+elif sys.platform in ("esp32", "esp8266"):
+    # Test with a hardware timer as software timers aren't implemented
+    # on the esp32 and esp8266 ports. Timer ID 0 is guaranteed to exist.
+    timer_id = 0
+else:
+    timer_id = -1
```
andrewleech, line 17: It would be good to test with both hardware and software timers if possible on platforms that support it?

andrewleech (commented):
Great work broadening the timer features and port consistency here!

tests/extmod/machine_timer.py
arachsys: Yes, it'd be great to make this more general to cover both on platforms that support both. Perhaps I can
```python
try:
  t = machine.Timer(0)
  # hardware timers available; test them
except:
  # hardware timers not available
```
and ditto for `Timer(-1)` for the soft timers. Is there anywhere where timer ID 0 is available but unsafe to use? Maybe
```python
ids = []
for i in -1, 0:
  try:
    t = machine.Timer(i)
    ids.append(i)
  except:
    pass

for i in ids:
  [...]
```

ports/esp32/machine_timer.c
arachsys: I originally hoped to consolidate all these stanzas into a single place, ideally merging into the similar IRQ object too, and I have provided a shared implementation of this in the patch to `extmod/machine_timer.c`.

However, many ports roll their own and don't use `extmod/machine_timer.c` at all, and there are a lot of incidental differences between these ports. For example, in the esp32 patch you quote above, we have to `mp_hal_wake_main_task_from_isr()` after `mp_sched_schedule()`, unlike any other port. Some ports (renesas-ra, stm32) have the notion of channels which change the shape of the error messages. Different ports have different structures for the timer object with different ways to disable callbacks on uncaught exceptions.

A lot of these incidental differences could be smoothed out, e.g. using `callback = NULL` as the disable technique everywhere with appropriate modifications to the callers. But I confess that without testing on real hardware, I chickened out and kept the code in existing ports as similar as I could in this first step, planning a clean-up-and-unify patch as a follow-up!

I expect it would be possible to pull out something sufficiently generalised like

```c
    if (hard) {
        mp_sched_lock();
        gc_lock();
        nlr_buf_t nlr;
        if (nlr_push(&nlr) == 0) {
            mp_call_function_1(callback, MP_OBJ_FROM_PTR(argument));
            nlr_pop();
        } else {
            /* caller has to disable and print exception */
            return nlr.ret_val;
        }
        gc_unlock();
        mp_sched_unlock();
    } else {
        mp_sched_schedule(callback, MP_OBJ_FROM_PTR(argument));
    }
    return NULL;
```

and then use it as

```c
    ret_val = dispatch_callback(self->callback, self);
    if (ret_val) {
        self->repeat = 0;
        mp_printf(MICROPY_ERROR_PRINTER, "uncaught exception in timer callback\n");
        mp_obj_print_exception(MICROPY_ERROR_PRINTER, MP_OBJ_FROM_PTR(ret_val));
    }
    if (!self->hard) {
        /* special case that isn't handled by dispatch_callback for esp32 */
        mp_hal_wake_main_task_from_isr()
    }
```

or something like that. Perhaps pass the "timer callback" string from the error message as an argument so the exception handling can all live in dispatch_callback too. If that covered the IRQ case as well as the Timer case for each port, it might give us a slight code shrinkage too, albeit slightly slower at interrupt time.

(I think I'm assuming that callbacks can't raise a VM abort in the above, or at least ignoring the exception if they do, by assuming `nlr.ret_val` is non-NULL in an exception return. Possibly we're doing that already by calling mp_obj_print_exception with them?)

arachsys: If we're being really brave about eliminating repetition here: where ports roll their own soft-timers instead of using the standard implementation (esp32, rp2, etc)... should they? Or, once hard=True and hard=False are properly supported on the shared `extmod/machine_timer.c` implementation, should everything 'custom' be replaced by that? Probably a bit controversial for this PR though! :)

tests/extmod/machine_timer.py
arachsys: As you suggested, I've added an extra test to cover hardware timers on the platforms that support them. I think `timer_id = 0` is the only choice that's guaranteed to be available on esp32 and esp8266, depending on the chip, whereas `timer_id = 1` is a safe choice on stm32, so I've not tried to do anything too clever/automatic.

ports/esp32/machine_timer.c
andrewleech: Yeah it's always hard to know how far to take a change like this! 

FWIW most ports already have a [stub `mp_hal_wake_main_task_from_isr` ](https://github.com/micropython/micropython/blob/e993f53877027bc46b226d026845e523c2cc94ca/ports/rp2/mphalport.h#L80 )for exactly this kind of consolidation.

arachsys: Okay, you've emboldened me! I've pushed an additional patch as a suggestion for how we might consolidate this, factoring out mp_irq_dispatch() as a generalised version of mp_irq_handler() and then using it throughout the timer implementations. Hopefully this should give a small size reduction on quite a few ports. What do you think?

I see esp8266 doesn't link against mpirq.c at all, so I've not tried to consolidate that one for now. (But mpirq.c is quite minimal apart from mp_irq_handler(), so maybe the penalty for linking it would be negligible...)

ports/esp32/machine_timer.c
```diff
@@ -238,6 +246,7 @@ static mp_obj_t machine_timer_init_helper(machine_timer_obj_t *self, mp_uint_t n
     self->repeat = args[ARG_mode].u_int;
     self->handler = machine_timer_isr_handler;
     self->callback = args[ARG_callback].u_obj;
+    self->ishard = args[ARG_hard].u_bool;
```
andrewleech: It might be just me but I read `ishard` as "I shard", personally I'd go for `is_hard` or just `hard`. Minor detail though...

shared/runtime/softtimer.c
```diff
@@ -88,7 +89,11 @@ void soft_timer_handler(void) {
         soft_timer_entry_t *entry = heap;
         heap = (soft_timer_entry_t *)mp_pairheap_pop(soft_timer_lt, &heap->pairheap);
         if (entry->flags & SOFT_TIMER_FLAG_PY_CALLBACK) {
-            mp_sched_schedule(entry->py_callback, MP_OBJ_FROM_PTR(entry));
+            if (mp_irq_dispatch(entry->py_callback, MP_OBJ_FROM_PTR(entry),
+                entry->flags & SOFT_TIMER_FLAG_HARD_CALLBACK)) {
+                // Uncaught exception; disable the callback so it doesn't run again.
```
andrewleech, line 94: Looks like the callback disable is missing now?

shared/runtime/softtimer.h
```diff
@@ -30,6 +30,7 @@
 
 #define SOFT_TIMER_FLAG_PY_CALLBACK (1)
 #define SOFT_TIMER_FLAG_GC_ALLOCATED (2)
+#define SOFT_TIMER_FLAG_HARD_CALLBACK (4)
```
andrewleech, line 33: Was 4 used here for a reason? Is 3 used elsewhere?

andrewleech (commented):
The consolidation looks good to me!

ports/esp32/machine_timer.c
arachsys: I'm not a great fan of 'isfoo' naming either, but the ishard in the timer structs is copying the existing ishard bool in the irq structs. I've tried to be as consistent with the existing code as possible, especially any conventions in the 'core' parts of micropython.

shared/runtime/softtimer.c
arachsys: It's the line below that comment where I set entry->mode to SOFT_TIMER_MODE_ONE_SHOT.

shared/runtime/softtimer.h
arachsys: These are bit masks that get ORed together, so powers of two.

arachsys:
> The consolidation looks good to me!

Yes, I like it too to be honest. The size reduction was better than I expected. Thanks for encouraging me not to be too timid!

arachsys:
[Rebased to HEAD of master.]

dpgeorge:
Thanks for the contribution!  Always good to clean up things and consolidate across ports.

Unfortunately esp32 cannot support hardware interrupts, because the hardware interrupt does not run on the main thread.  The interrupt handler needs to acquire the GIL which is non-trivial.  Testing the `extmod/machine_hard_timer.py` test on ESP32 it indeed crashes running the `hard=True` part of the test, and the REPL locks up.

So, I think for esp32 just raise an exception if `hard=True` is passed in.

For esp8266, the `tests/extmod/machine_hard_timer.py` doesn't currently work because esp8266 doesn't support passing in arguments to the constructor.  Instead it needs to do `timer=Timer(0)` then `timer.init(...)`.  Might be worth a separate PR to make that work the same as other ports?

I didn't check other boards yet.

dpgeorge:
I tested PYBV10 (stm32 board).  This port doesn't support non-soft `machine.Timer`, ie the `id` must be `-1` on this port, so `machine_hard_timer.py` test fails.

Also, `machine_timer.py` fails because the `modes` and `kinds` dicts evaluate to a different order than what they are defined in (eg it does periodic before one-shot).  Probably needs to use a tuple for those to ensure consistent order across all ports.

tests/extmod/machine_timer.py
```diff
@@ -0,0 +1,51 @@
+import sys
+
+try:
+    from machine import Timer
+    from time import sleep_ms
+except:
+    print("SKIP")
+    raise SystemExit
+
+if sys.platform in ("nrf"):
```
dpgeorge: This won't work, it should be `sys.plaftorm in ("nrf",)`.

arachsys:
> I tested PYBV10 (stm32 board). This port doesn't support non-soft `machine.Timer`, ie the `id` must be `-1` on this port, so `machine_hard_timer.py` test fails.

Ah, so my check for 'pyboard' in the hard timer test is not the right thing at all. Are there two versions of the pyboard, one of which supports hard timers and one of which does not? (pybv10 and pybv11 maybe?) Or is the situation more complicated than that?

> Also, `machine_timer.py` fails because the `modes` and `kinds` dicts evaluate to a different order than what they are defined in (eg it does periodic before one-shot). Probably needs to use a tuple for those to ensure consistent order across all ports.

Oops, that's embarrassing! I'll fix.

While I'm doing this, what do you think about the final 'consolidation' patch? It reduces code size a bit and cleans things up, but I wasn't sure what you'd make of splitting mp_irq_dispatch() like that. If you're happy with it, I could tap into mp_irq_dispatch() at the start of the series and use it straight away in the patches that extend each port. That would reduce diff noise and make things a bit clearer, while still leaving things in exactly the same state at the end of the series?

I didn't do that when I added it in case you wanted the rest of the changes but didn't like the last one.

arachsys:
> Unfortunately esp32 cannot support hardware interrupts, because the hardware interrupt does not run on the main thread. The interrupt handler needs to acquire the GIL which is non-trivial. Testing the `extmod/machine_hard_timer.py` test on ESP32 it indeed crashes running the `hard=True` part of the test, and the REPL locks up.
> 
> So, I think for esp32 just raise an exception if `hard=True` is passed in.

Thanks, I'll do that. Getting the GIL will definitely be hard as it might already be held by the code that got interrupted!

> For esp8266, the `tests/extmod/machine_hard_timer.py` doesn't currently work because esp8266 doesn't support passing in arguments to the constructor. Instead it needs to do `timer=Timer(0)` then `timer.init(...)`. Might be worth a separate PR to make that work the same as other ports?

Sounds like a good plan! I'll get the various issues with this one straightened out (including using timer.init() in this test) and then open that and link it.

dpgeorge:
> Are there two versions of the pyboard, one of which supports hard timers and one of which does not?

They all support `pyb.Timer` which is the stm32 specific way to do hardware timers.  For `machine.Timer`, only soft timer is supported by stm32 at the moment.

This is starting to dig up the main issue that the `machine.Timer` class is not well specified nor very consistent across ports, see #2971 for that discussion.

Ideally we'd address #2971 first, and then work on implementing it.  But --for the sake of making progress -- this PR is good.  It just might end up being a bit imperfect due to the incomplete nature of the `Timer` class on various ports.

> While I'm doing this, what do you think about the final 'consolidation' patch?

Yes, that's quite nice.  Always good to factor code, eliminate duplication, and reduce code size!  Eventually we might want to add `Timer.irq` (like eg `Pin.irq` and `UART.irq`) and then it would use the standard `mpirq.c` interface/code and not need any refactoring like that.  But -- again in the interest of making progress -- your refactoring is good and I think we should do it.  So feel free to expose `mp_irq_dispatch()` in an earlier commit and then use it immediately when updating each port.

arachsys:
> So feel free to expose `mp_irq_dispatch()` in an earlier commit and then use it immediately when updating each port.

I've restructured the commits, fixed the various incompatibilities you found, and have updated the patch series. Sorry I wasn't able to test these properly before sending the original series. My hardware collection clearly needs to grow!

dpgeorge:
> Sorry I wasn't able to test these properly before sending the original series. My hardware collection clearly needs to grow!

Not a problem.  There are many hardware combinations that need testing, and that's why we are trying to get Octoprobe up and running, for automated hardware testing.

I've started an Octoprobe run for this PR, to see how it goes.  In maybe 5 hours (yes it takes that long!) it'll appear at the top of the list here: https://reports.octoprobe.org/ (under the "275" on the left there will eventually be a "success" link that'll take you to the report).

dpgeorge:
I tested this on esp32, esp8266, stm32 and rp2 ports.  It works well.

I also tested on zephyr and the hard IRQ timer callback fails there.  That's because the hard IRQ executes on a different thread using a different C stack, and to make that work requires setting the stack limit variables.

@arachsys do you think you can fix that?  You can look at `gpio_callback_handler()` to see how the `machine.Pin` hard IRQ handles it, the bits of code guarded by `#if MICROPY_STACK_CHECK`.  Also `mp_machine_i2c_target_event_callback()` does a similar thing.  I guess it makes sense to factor that common code into some helper macros to be reused.

arachsys:
> I tested this on esp32, esp8266, stm32 and rp2 ports. It works well.

Great, thanks!

> I also tested on zephyr and the hard IRQ timer callback fails there. That's because the hard IRQ executes on a different thread using a different C stack, and to make that work requires setting the stack limit variables.
> 
> @arachsys do you think you can fix that? You can look at `gpio_callback_handler()` to see how the `machine.Pin` hard IRQ handles it, the bits of code guarded by `#if MICROPY_STACK_CHECK`. Also `mp_machine_i2c_target_event_callback()` does a similar thing. I guess it makes sense to factor that common code into some helper macros to be reused.

Sure! I've pushed an update to the existing zephyr patch which wraps with the same stack check adjustment as the two other places you pointed out where hard IRQ callbacks can be dispatched.

When it comes to consolidating, I wondered if it might make sense to push that stack-check adjustment down as far as `mp_irq_dispatch()` in `shared/runtime/mpirq.c`, wrapped in an appropriate `#ifdef`? Here, we'll eliminate code duplication between all the different sites and can easily wrap just the hard-IRQ case to make the common case of scheduling a soft-callback a bit cheaper. And presumably zephyr won't be the only port with a separate ISR-stack long-term?

If you think this is a good idea, I don't think there's an appropriate `mpconfigport.h` `#define` yet to indicate that ISRs run with their own stack, so I might have to add one. Maybe it should be something like `MICROPY_ISR_STACK`, looking at the style of the other feature flags?

(Are mechanisms like this what is missing to make hard IRQs work on esp32, or is it harder than zephyr there? Seems a pity to have a platform without hard IRQ support at all, although I know things get complicated when there's an RTOS in the picture too.)

dpgeorge:
> When it comes to consolidating, I wondered if it might make sense to push that stack-check adjustment down as far as `mp_irq_dispatch()` in `shared/runtime/mpirq.c`, wrapped in an appropriate `#ifdef`?

Yes, that sounds like a good approach.  Putting the stack adjustment only around the hard dispatch would be the most optimal thing to do.

If you do that, please use `mp_cstack_init_with_sp_here()`.  And then in zephyr's mpconfigport.h you'll need to define `MICROPY_STACK_CHECK_MARGIN` to 512.

> Maybe it should be something like `MICROPY_ISR_STACK`, looking at the style of the other feature flags?

It's tricky to name it... maybe `MICROPY_STACK_SIZE_HARD_IRQ`?  Default it to 0 and then if it's >0 then the stack check needs to be adjusted.

arachsys:
> It's tricky to name it... maybe `MICROPY_STACK_SIZE_HARD_IRQ`? Default it to 0 and then if it's >0 then the stack check needs to be adjusted.

I like that. I've added an extra commit that does this consolidation, moving the stack limit adjustment into `mp_irq_dispatch()`, adding the new `mpconfig.h` `#define`, and adjusting the zephyr `mpconfigport.h`.

tests/extmod/machine_timer.py
```diff
@@ -0,0 +1,48 @@
...
+    except MemoryError:
+        print("locked")
+
+
+modes = [(Timer.ONE_SHOT, "one-shot"), (Timer.PERIODIC, "periodic")]
+kinds = [(False, "soft"), (True, "hard")]
+
+for mode in modes:
+    for kind in kinds:
+        for freq in 500, 250:
```
dpgeorge: I ran this test on zephyr using the rpi_pico board.  It fails for the hard=True case because too many hard IRQs are executed (the sleep_ms(9) actually goes for around 32ms).  That seems to be because the IRQ callback takes too long to execute, and IRQs get backlogged.

I suggest decreasing the frequency of the test here by 10, ie do `for freq in 50, 25:` ... `sleep_ms(90)`.  That works on zephyr rpi_pico, and should be a lot more lenient for other ports and boards which are slow or don't have as accurate timing.

dpgeorge: It might be a good idea to (eventually) add a second timer test to check for more fine-grained accuracy of the period/frequency.  Eg using just hard=True, and store into a pre-allocated list the time of each callback.  Then at the end make sure those times were all within a certain percentage of the set period.

arachsys: Ah damn, sorry this has been such a pain to get debugged. I'll push a correction to make this 50Hz/25Hz now. I'll leave the additional fine-grained test for a follow-up 'improve tests' PR to avoid delaying this one any more, as you imply.

I'd assumed 500Hz was slow enough to be safe everywhere, but I guess I underestimated how wide the microcontroller performance range really is. (E.g. I'm using 10kHz hard timer interrupts on an rp2350 with a handler containing an ADC lookup, some integer maths, filling a pre-allocated array then triggering DMA to PIO in the handler, and any slow-down in foreground performance is too small to measure. If I dropped interrupts, the 1500V output capacitor would probably explode, so I'm sure I'd notice. ;-) )

I wonder if the much poorer performance on rp2 zephyr vs rp2 native is because of interrupt delivering latency, or just because we're calling print() (and therefore the USB stack?) from an isr?

dpgeorge: I think the issue with zephyr is the slow printing to USB.  Doing that in the hard IRQ is really slowing it down.  Because otherwise -- as you say -- it should be fine at 500Hz.

(And nice to hear you can get a 10kHz Python IRQ!!)

dpgeorge (approved):
This is looking really good now.  I tested on esp32, esp8266, stm32 (PYBV10), rp2 and zephyr and the timer tests work as expected.

Overall this is a neat clean up and consolidation of code, and makes the ports more consistent.  And thanks for making well defined commits with good commit comments, and including docs and tests updates.  Very good!

PR #17583, opened 2025-06-30
mpprint: Add %R (print object) & %K (print exception).

jepler:
As well as adding some use sites to get an idea of whether this is saving code size.

This also makes that `mp_obj_print_helper` print `MP_OBJ_NULL` objects as `(nil)` (regardless of NDEBUG), which is otherwise written at several use sites.

With just a few sites modified to use %R & %K, it already saved around 28 bytes on MPS2_AN385 according to a quick local test.

### Summary

I noticed that often there is an alternation between calling `mp_printf` & `mp_print_obj_helper`. I hypothesized that overall code size might be reduced by adding a %-format for whole objects to mp_printf.

### Testing

I locally ran the unix coverage suite & built a qemu build for size comparison.

* Somewhat arbitrarily I chose `%R` for objects and `%K` for exceptions. I initially decided to pass in the print kind for `%R` but this may be unneeded as in the sites I converted so far, the print type is statically known to be `PRINT_REPR`.
 * Printing some deeply nested objects will terminate sooner when it goes through a path that uses `%R`, as now `mp_printf` will be on the stack too.
 * This would require updates to the printf format string checker in #17556, which I will do if this PR is accepted and that PR is deemed desirable.
 * There are probably more sites amenable to conversion. I only converted enough sites to show that there was a size savings to be had.

### DIFF
--- a/py/modthread.c
+++ b/py/modthread.c
@@ -189,10 +189,8 @@ static void *thread_entry(void *args_in) {
             // swallow exception silently
         } else {
             // print exception out
-            mp_printf(MICROPY_ERROR_PRINTER, "Unhandled exception in thread started by ");
-            mp_obj_print_helper(MICROPY_ERROR_PRINTER, args->fun, PRINT_REPR);
-            mp_printf(MICROPY_ERROR_PRINTER, "\n");
-            mp_obj_print_exception(MICROPY_ERROR_PRINTER, MP_OBJ_FROM_PTR(exc));
+            mp_printf(MICROPY_ERROR_PRINTER, "Unhandled exception in thread started by %R\n%K",
+                PRINT_REPR, args->fun, MP_OBJ_FROM_PTR(exc));
         }
     }
 
--- a/py/mpprint.c
+++ b/py/mpprint.c
@@ -475,6 +475,17 @@ int mp_vprintf(const mp_print_t *print, const char *fmt, va_list args) {
                 chrs += mp_print_strn(print, &str, 1, flags, fill, width);
                 break;
             }
+            case 'K': {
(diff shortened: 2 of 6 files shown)


jepler:
The code size is a bit all over the place. This is because the more minimal ports aren't benefitting from (m)any of the reductions, but are paying the full cost of the added format codes.

the cc3200 failure is interesting. It looks like `mp_printf` but not `mp_obj_print_helper` is built into the "bootmgr" program. So, support for these new format strings would need to be conditional.

Making it conditional would be nice, as some ports might not use `%R`; the plugin printf checker could probably consult the compiler flags to check if it's available when checking format strings, but that's a hypothetical at this point.

the nanbox failure is due to printing an object of type mp_obj_t via the %p format specifier. This mistake wasn't detected at compile time by #17556 because the plugin was only enabled for coverage builds, but the nanbox build is its own variant.

`%K` should likely be dropped, I'm guessing there are too few sites where it would be used.

This is feeling overly complicated for a small savings only on larger builds where clawing for those last few bytes doesn't make as much difference usually. Please feel free to close if you're feeling the same way.

dpgeorge:
I can see how this could be useful in decreasing code size.  And it's definitely a lot cleaner to use `%R` than explicitly calling the function to print the object.

But, to merge it, the change can't increase size of the `bare-arm` build.

Note that unix and mimxrt are the ones with the biggest savings because they are the only ones on the list with `MICROPY_ERROR_REPORTING_DETAILED` enabled, which is where most of the use of `%R` is.

jepler:
Gonna close this up for now even though I'd like to revisit it someday.

PR #17604, opened 2025-07-02
webassembly/objpyproxy: Avoid throwing on symbols or iterator.

WebReflection:
Not only our latest stack uses internally symbols to brand check its own proxies, there are various 3rd party libraries that might perform similar checks and you also have:

```js
if (prop === Symbol.iterator) ...
```

in your `get` trap but a check before such as:

```js
if (Symbol.iterator in ref) ...
```

would fail with an error that states that a string was expected but a symbol was received.

This MR would like to fix the situation for at least known symbols you also handle internally.

### DIFF
--- a/ports/webassembly/objpyproxy.js
+++ b/ports/webassembly/objpyproxy.js
@@ -148,6 +148,12 @@ const py_proxy_handler = {
         };
     },
     has(target, prop) {
+        // avoid throwing on `Symbol() in proxy` checks
+        if (typeof prop !== "string") {
+            // returns true only on iterator because other
+            // symbols are not considered in the `get` trap
+            return prop === Symbol.iterator;
+        }
         return Module.ccall(
             "proxy_c_to_js_has_attr",
             "number",
--- a/tests/ports/webassembly/py_proxy_has.mjs
+++ b/tests/ports/webassembly/py_proxy_has.mjs
@@ -9,3 +9,5 @@ x = []
 const x = mp.globals.get("x");
 console.log("no_exist" in x);
 console.log("sort" in x);
+console.log(Symbol.toStringTag in x);
+console.log(Symbol.iterator in x);
--- a/tests/ports/webassembly/py_proxy_has.mjs.exp
+++ b/tests/ports/webassembly/py_proxy_has.mjs.exp
@@ -1,2 +1,4 @@
 false
 true
+false
+true


WebReflection:
for quick check about current state VS state after this MR (published on my npm account with same version + this MR in) and for history sake, here a test that shows the issue:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>MicroPython VS JS Symbols</title>
  <script type="module">
    import { loadMicroPython } from 'https://cdn.jsdelivr.net/npm/@micropython/micropython-webassembly-pyscript/micropython.mjs';
    // change this with webreflection instead to see it working -- ^^^^^^^^^^^
    import codedent from 'https://esm.run/codedent';
    const mpy = await loadMicroPython();
    // sync or async it doesn't matter
    await mpy.runPythonAsync(codedent(document.querySelector('script[type=micropython]').textContent));
  </script>
  <script>
    globalThis.hasSymbol = (symbol, ref) => symbol in ref;
    globalThis.getSymbol = (symbol, ref) => ref[symbol];

    // some 3rd party JS library might use symbols to brand-check
    // so it's not about symbols traveling from MicroPython
    // it's about MicroPython proxies traps not understanding symbols
    globalThis.hasIterator = ref => Symbol.iterator in ref;
  </script>
  <script type="micropython">
    import js

    # change this to Symbol.toStringTag to see it throwing all over
    symbol = js.Symbol.iterator

    # this passes only with Symbol.iterator
    print('get', js.getSymbol(symbol, []))

    # these both fail miserably
    print('has', js.hasSymbol(symbol, []))
    print('direct', js.hasIterator([]))
  </script>
</head>
</html>
```

dpgeorge:
Thanks for the patch.

Was this a regression that worked in a prior version?

ntoll:
Yes, it's a regression. If you compare MicroPython bundled in the previous version of PyScript 2025.5.1 with current PyScript (2025.7.1) there's a difference in behaviour.

WebReflection:
@dpgeorge 

> Was this a regression that worked in a prior version?

I don't think so ... I've never seen `symbol` in this code to my memory and I don't think that part has been changed/touched

@ntoll 

> Yes, it's a regression. If you compare MicroPython bundled in the previous version of PyScript 2025.5.1 with current PyScript (2025.7.1) there's a difference in behaviour.

that's my miscommunication ... the regression is introduced in our latest PyScript because we use better primitives to handle worker/main use cases and those better primitives revealed this issue with MicroPython which has been broken to date but now it starts hurting when paired with more modern libraries and code.

As mentioned in Discord, there's much more to do to have MicroPython fully handling symbols but for the time being, at least having it not throwing and return `false` for `has` traps unless it's `Symbol.iterator` which again, it doesn't handle well but it's needed to make list and tuples iterables, would be ideal and I'd be more than happy to provide a way wider MR that touches tons of things around the Proxy dance done in MicroPython, as an attempt to contribute the JS/Python interoperability story and never have surprises, if that's welcomed and desired in here too.

ntoll:
@WebReflection ah, my bad. I misconstrued the context.

The MicroPython/JS proxy dance could be something to discuss in our next call (21st July)..? I think if we (PyScript - i.e. Andrea) can contribute back to upstream MicroPython we're telling and embodying a good FLOSS open source story.

WebReflection:
@dpgeorge last time this was touched is in here: https://github.com/micropython/micropython/pull/15293 I should've reviewed that (or provided hints around this topic) ... symbols are very special (and weird) beast in JS and particularly from workers these cannot ever travel but even on main we have issues in checking if special, libraries defined, symbols are present in objects to simply brand check these or guard against undesired access (private symbols in JS are also a thing, these are all `typeof "symbol"`).

This MR prevents those checks from failing but it also grant that the only one understood and implemented in the `get` trap is reflected ... and about that, it should be reflected in `get` only if the underlying python object is iterable, which luckily enough for Python is probably dictionaries too, but surely not callbacks ... so there's much more to consider to have 100% symbol support but at least please let's not throw errors for traps that by specifications expect either *strings* or *symbols* (which is all traps that accept a `key` among the target): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/has#parameters

- - -

**edid** I did review that and forgot checking the `has` trap 🤦🤦🤦

https://github.com/micropython/micropython/pull/15293#issuecomment-2173461171

dpgeorge (approved):
This fix looks good to me, and the new test covers the new code and passes.

ntoll:
Bravo and thank you..! Has this made it into a release..?

dpgeorge:
> Has this made it into a release..?

Yes: https://www.npmjs.com/package/@micropython/micropython-webassembly-pyscript/v/1.26.0-preview-293

ntoll:
Awesome work. As always. :tada:

Issue #17608, opened 2025-07-02
zephyr: Ctrl-C does not cause KeyboardInterrupt


bikeNomad:
### Port, board and/or hardware

zephyr port, XIAO BLE NRF52840 SENSE

### MicroPython version

MicroPython v1.26.0-preview.293.g7989a1488 on 2025-07-02; zephyr-xiao_ble with nrf52840

built with
```
west build -b xiao_ble/nrf52840/sense   ../submodules/micropython/ports/zephyr
```

### Reproduction

At the REPL, enter this:
```
while True:
    pass
```
Then try to interrupt with Ctrl-C.

### Expected behaviour

Expected to interrupt the loop and return to the REPL prompt.

### Observed behaviour

Kept running, no response to Ctrl-C

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

dpgeorge:
I've opened #17610 which should fix this.

bikeNomad:
Fixed by #17610

PR #17610, opened 2025-07-03
zephyr: Use UART for console as default, not CONSOLE_SUBSYS.

dpgeorge:
### Summary

Most zephyr boards enable the UART console because it's needed for USB (where USB CDC creates a virtual UART), and for ctrl-C to work.  This PR makes that option the default (prior to this PR, CONSOLE_SUBSYS was the default).
    
The `prj_minimal.conf` settings still use CONSOLE_SUBSYS, and show how it can be configured.

Fixes issue #17608.

(I also updated the qemu board configurations so they work with `prj.conf`, which was needed for testing.)

### Testing

Tested a REPL on:
- qemu_x86, both prj and prj_minimal settings
- qemu_cortex_m3, both prj and prj_minimal settings
- nucleo_wb55rg


bikeNomad:
Works on XIAO BLE NRF52840 SENSE board.

dpgeorge:
> Works on XIAO BLE NRF52840 SENSE board.

That's good, thanks for testing.

PR #17613, opened 2025-07-04
stm32/eth: Improve Ethernet driver with link detection and static IP support.

andrewleech:
## Summary

This PR implements background link detection, hot-plug support, and multipleminor bug fixes for the STM32 Ethernet driver. The changes eliminate blocking timeouts, enable static IP configuration workflows, and fix reliability issues discovered during testing and code review.

**Key Features**:
-  Static IP configuration before `LAN.active(True)` with proper preservation
-  Automatic DHCP restart on cable replug
-  Background PHY polling for cable connect/disconnect detection
-  Non-blocking interface activation (56× faster with cable, no timeout without)
-  Dynamic MAC speed/duplex configuration matching PHY autonegotiation
-  Network initialization before boot.py execution

**Critical Fixes**:
-  Fixed static IP being overwritten after MAC reconfiguration
-  Fixed DHCP restart logic to work in all states
-  Eliminated 30ms IRQ blackout during MAC reconfiguration
-  Fixed IEEE 802.3 PHY_BSR latched-low bit handling for reliable hot-plug
-  Fixed race conditions in link detection and MAC configuration

**Breaking Change**:
 `LAN.active()` now returns interface enabled state instead of link status. Use `LAN.status()` or `LAN.isconnected()` to check cable connection.

## Testing

**Hardware Tested**:
- NUCLEO_H563ZI (STM32H563ZI) - Primary development board
- NUCLEO_F429ZI (STM32F429ZI) - Regression testing

**Test Coverage - 11 Scenarios**:

*Basic Functionality*:
1. Fresh boot with cable → DHCP obtains IP in 2-3 seconds 
2. `active(True)` without cable → Returns in 28ms, no timeout 
3. Link status before `active()` → Returns current state via direct PHY read 
4. Fresh boot without cable → `active(True)` succeeds immediately 

*Static IP Configuration*:
5. Static IP before `active()` → IP preserved, DHCP not started 
6. Static IP after DHCP → Static IP replaces DHCP 
7. Static IP through toggle → IP survives `active(False)`/`active(True)` 
8. Static IP during hot-plug → IP preserved during cable unplug/replug 

*Hot-Plug Detection*:
9. Boot without cable + plug-in → Link detected, DHCP starts immediately 
10. Hot-unplug detection → Status 3→0 within 1 second 
11. Hot-plug DHCP recovery → Status 0→3, DHCP completes in 2-4 seconds 

**Performance Improvements**:

| Operation | Before | After | Improvement |
|-----------|--------|-------|-------------|
| `active(True)` with cable | 1586ms | 28ms | **56× faster** |
| `active(True)` without cable | 10s timeout | 28ms | **No blocking** |
| DHCP acquisition | 6 seconds | 2-3 seconds | **2× faster** |
| IRQ blackout during reconfig | 30ms | 0ms | **Eliminated** |

**Board-Specific Issues Found & Fixed**:
- **H563ZI**: ETH clock stopped during CPU WFI → Enabled clock continuation during sleep
- **F429ZI**: MAC hardcoded 100Mbps, PHY negotiated different → Dynamic MAC reconfiguration

#### 1. Network Initialization Order (1307179e60)
**Problem**: `network.LAN()` couldn't be instantiated in boot.py due to initialization order.

**Solution**: Moved `mod_network_init()` before boot.py execution. Safe because LWIP initialization already occurs earlier; mod_network_init only sets up the NIC list.

**Impact**: Enables network configuration in boot.py.

**Files**: `stm32/main.c`, `stm32/mpnetworkport.c`

---

#### 2. Background Link Detection (90297bc82a)
**Problem**: No mechanism to detect cable unplug/replug events. Blocking PHY initialization caused multi-second delays.

**Architectural Changes**:
- Split LWIP netif initialization into early phase (eth_init) and late phase (eth_start)
- Added `eth_phy_link_status_poll()` called every ~250ms from background timer
- Removed blocking PHY autonegotiation loops from initialization path
- Integrated with LWIP via `netif_set_link_up/down()`

**Rationale for Polling**: STM32 Nucleo boards don't wire PHY interrupt lines to MCU. Polling is the only method to detect cable state changes without hardware modifications.

**Functional Changes**:
- `active(True)` returns immediately (~28ms) regardless of cable state
- Static IP can be configured before calling `active(True)`
- PHY autonegotiation happens asynchronously in background
- `active()` now returns interface enabled state (not link status) - **BREAKING CHANGE**

**Performance**: 56× faster activation, 2× faster DHCP

**Files**: `stm32/eth.c` (+262 lines), `stm32/eth.h`, `stm32/network_lan.c`, `stm32/mpnetworkport.c`

---

#### 3. MAC Speed/Duplex and DHCP Hot-Plug (e75650ea86)
**Problem**: MAC speed/duplex hardcoded at initialization. DHCP didn't restart after cable replug.

**MAC Speed/Duplex**:
- Poll for PHY autonegotiation completion in background
- Reconfigure MAC speed/duplex to match negotiated values
- 5-second timeout with fallback to 10Mbps Half-Duplex
- Protected MAC reconfiguration with IRQ disable (later improved in Fix 1)

**DHCP Hot-Plug**:
- Detect cable replug via link status change
- Restart DHCP using `dhcp_stop()` + `dhcp_start()` pattern
- Preserve static IP by checking DHCP state

**Issue Fixed**: F429ZI regression where MAC/PHY speed mismatch prevented DHCP.

**Files**: `stm32/eth.c` (+100 lines), `stm32/eth_phy.h`

---

### Breaking Change: `active()` Method

**Old Behavior** (v1.26.1 and earlier):
```python
lan = network.LAN()
lan.active(True)
if lan.active():  # Returned True only if cable connected
    print("Cable is plugged in")
```

**New Behavior** (this PR):
```python
lan = network.LAN()
lan.active(True)
if lan.active():  # Returns True if interface is enabled
    print("Interface is enabled")

# To check cable connection, use:
if lan.isconnected():  # or lan.status() == 3
    print("Cable is plugged in and IP acquired")
```

**Rationale**: Separates interface administrative state from physical link state, consistent with other network stacks and allows `active(True)` to succeed without blocking on cable presence.

---

### Static IP Before Activation
```python
import network

lan = network.LAN()
lan.active(False)

# Configure static IP while interface is down
lan.ifconfig(('192.168.1.100', '255.255.255.0', '192.168.1.1', '8.8.8.8'))

# Activate interface - IP is preserved
lan.active(True)

# Check status
print(f"Status: {lan.status()}")  # 0=down, 1=link, 2=link+no IP, 3=connected
print(f"IP: {lan.ifconfig()[0]}")  # 192.168.1.100
```

### Boot.py Network Configuration
```python
# In boot.py - now works!
import network

network.country('US')
network.hostname('my-device')

lan = network.LAN()
lan.active(True)

# Wait for connection (cable may not be plugged yet)
import time
for _ in range(30):  # 30 second timeout
    if lan.isconnected():
        print(f'Connected: {lan.ifconfig()[0]}')
        break
    time.sleep(1)
```

### Hot-Plug Detection
```python
import network, time

lan = network.LAN()
lan.active(True)

while True:
    status = lan.status()
    if status == 3:
        print(f"Connected: {lan.ifconfig()[0]}")
    elif status == 0:
        print("Cable unplugged")
    time.sleep(1)
```

### DIFF
--- a/ports/stm32/eth.c
+++ b/ports/stm32/eth.c
@@ -107,6 +107,7 @@
 // Configuration values
 
 #define PHY_INIT_TIMEOUT_MS (10000)
+#define PHY_AUTONEG_TIMEOUT_MS (5000)
 
 // These buffer sizes need to be a multiple of 8 (for STM32N6 at least).
 #define RX_BUF_SIZE (1528) // includes 4-byte CRC at end
@@ -162,6 +163,11 @@ typedef struct _eth_t {
     uint32_t phy_addr;
     void (*phy_init)(uint32_t phy_addr);
     int16_t (*phy_get_link_status)(uint32_t phy_addr);
+    bool last_link_status;
+    bool enabled;
+    bool mac_speed_configured;
+    uint32_t autoneg_start_ms;
+    volatile bool mac_reconfig_in_progress;
 } eth_t;
 
 // This struct contains RX and TX buffers shared with the DMA, and they may need
@@ -181,6 +187,9 @@ eth_t eth_instance;
 
 static void eth_mac_deinit(eth_t *self);
 static void eth_process_frame(eth_t *self, size_t len, const uint8_t *buf);
+static void eth_lwip_init(eth_t *self);
+static int eth_phy_init(eth_t *self);
+static void eth_dhcp_restart_if_needed(struct netif *netif);
 
 void eth_phy_write(uint32_t phy_addr, uint32_t reg, uint32_t val) {
     #if defined(STM32H5) || defined(STM32H7) || defined(STM32N6)
@@ -239,9 +248,18 @@ uint32_t eth_phy_read(uint32_t phy_addr, uint32_t reg) {
 }
 
 int eth_init(eth_t *self, int mac_idx, uint32_t phy_addr, int phy_type) {
+    if (self->netif.input != NULL) {
+        // Already initialised.
+        return 0;
+    }
     mp_hal_get_mac(mac_idx, &self->netif.hwaddr[0]);
     self->netif.hwaddr_len = 6;
     self->phy_addr = phy_addr;
+    self->last_link_status = false;
+    self->enabled = false;
+    self->mac_reconfig_in_progress = false;
+    self->mac_speed_configured = false;
+    self->autoneg_start_ms = 0;
     self->phy_init = eth_phy_generic_init;
     if (phy_type == ETH_PHY_DP83825 || phy_type == ETH_PHY_DP83848) {
         self->phy_get_link_status = eth_phy_dp838xx_get_link_status;
@@ -301,6 +319,10 @@ int eth_init(eth_t *self, int mac_idx, uint32_t phy_addr, int phy_type) {
     #else
     __HAL_RCC_ETH_CLK_ENABLE();
     #endif
+
+    // Register netif with LWIP so the interface is visible before active(True).
+    eth_lwip_init(self);
+
     return 0;
 }
 
@@ -356,6 +378,11 @@ static int eth_mac_init(eth_t *self) {
     SYSCFG->PMC |= SYSCFG_PMC_MII_RMII_SEL;
     #endif
 
+    // Release ETH peripheral from reset and enable clocks during CPU sleep.
+    // Note: CLK_SLEEP_ENABLE means clocks stay ON during sleep (not OFF).
+    // Clocks must continue during sleep to allow the ETH peripheral to receive
+    // packets and generate interrupts when the CPU enters sleep mode (WFI),
+    // which is necessary for DHCP and other network traffic.
     #if defined(STM32H5)
     __HAL_RCC_ETH_RELEASE_RESET();
 
@@ -467,52 +494,6 @@ static int eth_mac_init(eth_t *self) {
     ETH->DMA_CH[RX_DMA_CH].DMACCR &= ~(ETH_DMACxCR_DSL_Msk);
     ETH->DMA_CH[TX_DMA_CH].DMACCR &= ~(ETH_DMACxCR_DSL_Msk);
     #endif
-
-    // Reset and initialize the PHY.
-    self->phy_init(self->phy_addr);
-
-    // Wait for the PHY link to be established
-    int phy_state = 0;
-    t0 = mp_hal_ticks_ms();
-    while (phy_state != 3) {
-        if (mp_hal_ticks_ms() - t0 > PHY_INIT_TIMEOUT_MS) {
-            eth_mac_deinit(self);
-            return -MP_ETIMEDOUT;
-        }
-        uint16_t bcr = eth_phy_read(self->phy_addr, PHY_BCR);
-        uint16_t bsr = eth_phy_read(self->phy_addr, PHY_BSR);
-        switch (phy_state) {
-            case 0:
-                if (!(bcr & PHY_BCR_SOFT_RESET)) {
-                    phy_state = 1;
-                }
-                break;
-            case 1:
-                if (bsr & PHY_BSR_LINK_STATUS) {
-                    // Announce all modes
-                    eth_phy_write(self->phy_addr, PHY_ANAR,
-                        PHY_ANAR_SPEED_10HALF |
-                        PHY_ANAR_SPEED_10FULL |
-                        PHY_ANAR_SPEED_100HALF |
-                        PHY_ANAR_SPEED_100FULL |
-                        PHY_ANAR_IEEE802_3);
-                    // Start autonegotiate.
-                    eth_phy_write(self->phy_addr, PHY_BCR, PHY_BCR_AUTONEG_EN);
-                    phy_state = 2;
-                }
-                break;
-            case 2:
-                if ((bsr & (PHY_BSR_AUTONEG_DONE | PHY_BSR_LINK_STATUS))
-                    == (PHY_BSR_AUTONEG_DONE | PHY_BSR_LINK_STATUS)) {
-                    phy_state = 3;
-                }
-                break;
-        }
-        mp_hal_delay_ms(2);
-    }
-
-    // Get register with link status
-    uint16_t phy_scsr = self->phy_get_link_status(self->phy_addr);
 
     // Burst mode configuration
     #if defined(STM32H5) || defined(STM32H7) || defined(STM32N6)
@@ -654,23 +635,15 @@ static int eth_mac_init(eth_t *self) {
     ETH->MACA0LR = mac[3] << 24 | mac[2] << 16 | mac[1] << 8 | mac[0];
     mp_hal_delay_ms(2);
 
-    // Work out the line speed configuration for MACCR.
-    uint32_t maccr = 0;
-    if (phy_scsr & PHY_SPEED_100HALF) {
-        maccr |= ETH_MACCR_FES;
-    }
-    if (phy_scsr & PHY_DUPLEX) {
-        maccr |= ETH_MACCR_DM;
-    }
+    // Set MAC control register to a safe default (100Mbps Full Duplex).
+    // The actual speed/duplex is configured by eth_phy_link_status_poll() once
+    // PHY autonegotiation completes.
+    uint32_t maccr = ETH_MACCR_FES | ETH_MACCR_DM;
 
     #if defined(STM32N6)
 
-    if (!(phy_scsr & PHY_SPEED_1000HALF)) {
-        maccr |= ETH_MACCR_PS;
-    }
-
-    maccr |=
-        ETH_MACCR_IPG_96BIT
+    maccr |= ETH_MACCR_PS // 100/10Mbit, reconfigured by poll if 1000Mbit negotiated
+        | ETH_MACCR_IPG_96BIT
         | ETH_MACCR_SARC_REPADDR0
         | ETH_MACCR_IPC
         | ETH_MACCR_BL_10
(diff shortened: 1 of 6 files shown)


panuph:
Thanks @andrewleech for this PR. I've done some test based on my use case around the use of `network.LAN().isconnected()`. 

This is the test result from the micropython version (branch master commit 35f15cfdf) that I am using:
```
# Disconnect the Ethernet cable and wait for 10 seconds

import network

network.LAN().active(False)

network.LAN().ifconfig()  # ('0.0.0.0', '0.0.0.0', '0.0.0.0', '0.0.0.0')
network.LAN().active()  # False
network.LAN().isconnected()  # False

# Connect the Ethernet cable and wait for 10 seconds

network.LAN().ifconfig()  # ('0.0.0.0', '0.0.0.0', '0.0.0.0', '0.0.0.0')
network.LAN().active()  # True
network.LAN().isconnected()  # False

network.LAN().active(True)

network.LAN().ifconfig()  # ('0.0.0.0', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # False

network.LAN().ipconfig(addr4=("192.168.1.1", "255.255.255.0"))

network.LAN().ifconfig()  # ('192.168.1.1', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # True

# Disconnect the Ethernet cable and wait for 10 seconds

network.LAN().ifconfig()  # ('192.168.1.1', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # True (this is most likely a bug!)

# Connect the Ethernet cable and wait for 10 seconds

network.LAN().ifconfig()  # ('192.168.1.1', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # True

network.LAN().active(False)

network.LAN().ifconfig()  # ('0.0.0.0', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # False
network.LAN().isconnected()  # False

# Disconnect the Ethernet cable and wait for 10 seconds

network.LAN().active(True)  # OSError: [Errno 110] ETIMEDOUT

```

This is the test result from your PR:
```
# Disconnect the Ethernet cable and wait for 10 seconds

import network

network.LAN().active(False)

network.LAN().ifconfig()  # ('0.0.0.0', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # False
network.LAN().isconnected()  # False

# Connect the Ethernet cable and wait for 10 seconds

network.LAN().ifconfig()  # ('0.0.0.0', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # False
network.LAN().isconnected()  # False

network.LAN().active(True)

network.LAN().ifconfig()  # ('0.0.0.0', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # False

network.LAN().ipconfig(addr4=("192.168.1.1", "255.255.255.0"))

network.LAN().ifconfig()  # ('192.168.1.1', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # True

# Disconnect the Ethernet cable and wait for 10 seconds

network.LAN().ifconfig()  # ('192.168.1.1', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # False

# Connect the Ethernet cable and wait for 10 seconds

network.LAN().ifconfig()  # ('192.168.1.1', '255.255.255.0', '192.168.0.1', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # True

network.LAN().active(False)

network.LAN().ifconfig()  # # ('0.0.0.0', '0.0.0.0', '0.0.0.0', '8.8.8.8')
network.LAN().active()  # False
network.LAN().isconnected()  # False

# Disconnect the Ethernet cable and wait for 10 seconds

network.LAN().ifconfig()  # # ('0.0.0.0', '0.0.0.0', '0.0.0.0', '8.8.8.8')
network.LAN().active()  # True
network.LAN().isconnected()  # False
```

`network.LAN().isconnected()` works correctly with your fix. Please note `OSError: [Errno 110] ETIMEDOUT` is no longer raised with your code when activating the network with no ethernet cable connected.

mattytrentini:
Should this still be in draft? It's looking pretty complete from our testing!

ports/stm32/eth_phy.h
```diff
@@ -56,6 +56,14 @@
 #define PHY_SPEED_100FULL  (6)
 #define PHY_DUPLEX         (4)
 
+// PHY interrupt registers (common for LAN87xx and DP838xx)
```
dpgeorge: These constants aren't used anywhere.

ports/stm32/eth.c
```diff
@@ -259,21 +275,21 @@ static int eth_mac_init(eth_t *self) {
     #if defined(STM32H5)
     __HAL_RCC_ETH_RELEASE_RESET();
 
-    __HAL_RCC_ETH_CLK_SLEEP_ENABLE();
-    __HAL_RCC_ETHTX_CLK_SLEEP_ENABLE();
-    __HAL_RCC_ETHRX_CLK_SLEEP_ENABLE();
+    __HAL_RCC_ETH_CLK_SLEEP_DISABLE();
+    __HAL_RCC_ETHTX_CLK_SLEEP_DISABLE();
+    __HAL_RCC_ETHRX_CLK_SLEEP_DISABLE();
```
dpgeorge: Is this change really necessary?  I would think that it's OK to just enable the clock sleep feature at the start, and not need to disable/enable it during reset.

andrewleech: Having the ENABLE lines the way he was previously is actually what I found prevented .active(True) from working when the ethernet cable was connected (on the H5 at least). The reset just after this would timeout unless the cable was connected. 

I'm not sure if explicitly disabling it here is necessary, though I thought it might be a good idea because I don't really know what state the system is in at this point.

dpgeorge: OK, I see.  Maybe it has to do with the WFI hiding in the `mp_hal_delay_ms(2)` (because the loop below that line is a busy loop and does not do WFI).

Well, probably it's OK to just disable clock-sleep altogether, because we do want ETH to run "in the background" when the CPU does WFI.  All other peripherals do that (disable clock-sleep), eg USB.

ports/stm32/eth_phy.h
andrewleech: Ah yes I'd initially thought I'd be able to use phy interrupts to trigger the link up/down events, however in many/most cases (including the H5 nucleo) the phy used shares a pin for both ref clock out and interrupt out aka if you're using the reference clock (like we are) you can't get interrupts from the phy.

ports/stm32/eth.c
```diff
@@ -258,22 +273,10 @@ static int eth_mac_init(eth_t *self) {
 
     #if defined(STM32H5)
     __HAL_RCC_ETH_RELEASE_RESET();
-
-    __HAL_RCC_ETH_CLK_SLEEP_ENABLE();
```
andrewleech, line 362: I'll test again with 
    __HAL_RCC_ETH_CLK_SLEEP_DISABLE(); for these to be explicit, with comment why.

andrewleech:
## CLK_SLEEP Configuration Investigation and Fix

During code review, we discovered that the restructure commit (90297bc82a) inadvertently removed all `CLK_SLEEP_ENABLE`/`DISABLE` calls, leaving the driver dependent on the HAL default state.

### Investigation: CLK_SLEEP Semantics

The macro naming is counter-intuitive:

- `__HAL_RCC_ETH_CLK_SLEEP_ENABLE()` → Clocks stay **ON** during CPU sleep (WFI)
- `__HAL_RCC_ETH_CLK_SLEEP_DISABLE()` → Clocks turn **OFF** during CPU sleep

From STM32H5 HAL documentation:
> "Enable or disable the AHB1 peripheral clock during Low Power (Sleep) mode.  
> By default, all peripheral clocks are enabled during SLEEP mode."

**Why clocks must stay enabled:** The ETH peripheral needs clocks during CPU sleep to receive packets and generate interrupts, which is necessary for DHCP and other network traffic.

### Fix Applied

Restored explicit `CLK_SLEEP_ENABLE` calls in `eth_mac_init()` for all STM32 platforms (H5, H7, F4) with clarifying comment:

```c
// Release ETH peripheral from reset and enable clocks during CPU sleep.
// Note: CLK_SLEEP_ENABLE means clocks stay ON during sleep (not OFF).
// Clocks must continue during sleep to allow the ETH peripheral to receive
// packets and generate interrupts when the CPU enters sleep mode (WFI),
// which is necessary for DHCP and other network traffic.
```

### Testing on NUCLEO_H563ZI

**Test 1: Interface activation without cable**
- ✅ `active(True)` completed in 28ms
- Status: 0 (link down, as expected)
- No timeout, immediate return

**Test 2: Hot-plug cable detection (2-hour monitoring)**
- ✅ Cable plugged at 536s
- Link detected immediately: Status 0 → 2
- DHCP acquired in 4 seconds: IP 192.168.0.10
- Background polling and DHCP working correctly

**Confirmed:** ETH peripheral receiving packets and processing DHCP during CPU sleep as expected. No regressions from explicit CLK_SLEEP configuration.

Commit: 095f4db460

andrewleech:
Rebased onto current upstream master, now sits as 2 commits matching the structure here.

**Behaviour change summary** (verified on NUCLEO_H563ZI point-to-point with AX88179 USB NIC + dnsmasq, and on OPENMV_N6 with the same rig):

| | Before | After |
|---|---|---|
| `active(True)` with cable | ~1.5s blocking | 68ms (H563ZI), 66ms (OPENMV_N6) |
| `active(True)` no cable | 10s timeout then OSError | 68ms, status=0 |
| DHCP acquisition after link-up | n/a (active blocked until cable) | 5s (H563ZI), 6s (OPENMV_N6) |
| Cable unplug detection | not detected; `isconnected()` stayed True forever | status 3->0 in 4s (H563ZI), 2s (OPENMV_N6) |
| Cable replug DHCP recovery | manual `active(False)/active(True)` cycle required | automatic, 13s (H563ZI) |
| Static IP through unplug/replug | cleared by driver | preserved (DHCP state checked via `dhcp_supplied_address()`) |
| `network.LAN()` in `boot.py` | not possible (init ran after boot.py) | works |
| Autoneg fallback to 10Mb/Half | silent | prints warning |

OPENMV_N6 testing exercises the gigabit/RGMII + RTL8211 path that landed via #18161. Autoneg completed cleanly at 1000Mb without falling back to the warning path.

Addressing the review comments:

**`eth.c:280` CLK_SLEEP** - kept the explicit `CLK_SLEEP_ENABLE` calls with a comment block explaining the counter-intuitive macro naming (`ENABLE` means clocks stay ON during sleep, not OFF). Without explicit ENABLE the driver depends on HAL default state, and the comment captures why the ETH clocks must stay running through WFI for packet RX. See the [Oct 30 comment](https://github.com/micropython/micropython/pull/17613#issuecomment-3466556561) for the investigation.

**`eth_phy.h:59` PHY_SPEED_* constants** - these are now used by the MAC speed/duplex reconfig path in `eth_phy_link_status_poll()`. PHY interrupt support to avoid polling isn't feasible on most STM32 Nucleo boards, where the PHY's INTRP pin is shared with the reference clock output.

One fix folded in beyond the previous draft: on link-down in DHCP mode the netif's DHCP-assigned IP is now cleared via `ip4_addr_set_zero()` before `dhcp_stop()`. Without this, the link-up restart check (`eth_dhcp_restart_if_needed`) sees a non-zero IP and skips the DHCP restart, so the lease was never renewed on reconnect. Static IPs are unaffected since `dhcp_supplied_address()` gates the clear.

Builds clean on NUCLEO_H563ZI, NUCLEO_F429ZI, NUCLEO_H743ZI, NUCLEO_N657X0.

@kwagyeman heads-up: this was also tested on OPENMV_N6 by cherry-picking the branch into the openmv micropython submodule; the same test set passed there.

kwagyeman:
@dpgeorge can you add this to the OpenMV Jira board?

dpgeorge (approved):
I've tested this on NUCLEO_F767ZI.  It works very well!  LAN bring up is much more usable and responsive now.

In particular it's very nice that `lan.active(True)` returns almost immediately and then negotiation and DHCP happen in the background.  That makes it similar to WLAN behaviour.

Also really nice how you can unplug and replug the ETH cable, `lan.status()` gives you immediate indication of that, and everything comes back online when you replug the cable.

Well done @andrewleech !

(Now we just need to do the same thing with mimxrt 😂 ... actually probably a lot of the code in `stm32/eth.c` could be made common and shared with mimxrt and other ports, eg the `eth_phy_link_status_poll()` function is almost port-agnostic.)

andrewleech:
> I've tested this on NUCLEO_F767ZI. It works very well! LAN bring up is much more usable and responsive now.

Thanks for the testing! Yes my goal was to use stm to prove out the pattern before rolling out across other ports, I was hoping to possibly consolidate at least some of the handling into a common file.

ports/stm32/eth.c
```diff
@@ -1102,28 +1075,174 @@ static void eth_lwip_init(eth_t *self) {
...
+        ETH->MACCR = maccr;
+        ETH->MACCR |= ETH_MACCR_TE | ETH_MACCR_RE;
+
+        self->mac_reconfig_in_progress = false;
+        self->mac_speed_configured = true;
+
+        // MAC was reconfigured; restart DHCP if needed.
+        struct netif *netif = &self->netif;
+        MICROPY_PY_LWIP_ENTER
+        eth_dhcp_restart_if_needed(netif);
```
dpgeorge, line 1243: @andrewleech I noticed that sometimes the DHCP client is stopped and started twice in a row from this function (here and above in the link up/down if-logic).

It still works, but probably better one day to improve this so it only calls it once (eg set a flag in the two locations here, and at the end of the function check if it needs to be done).

PR #17599, opened 2025-07-02
zephyr/machine_pin: Configure OUT pin also as input so it's readable.

dpgeorge:
### Summary

Zephyr allows setting both GPIO_OUTPUT and GPIO_INPUT on a pin, which means it's an output pin that can have its current value read.

Fixes issue #17596.

### Testing

Tested with `bbc_microbit_v2` and `nucleo_wb55rg` board definitions.  Both allow reading an output pin:
```
from machine import Pin
p = Pin(("gpioa", 0), Pin.OUT)
print(p())
p(1)
print(p())
p(0)
print(p())
```
Note: `nucleo_wb55rg` already works without this change, but `bbc_microbit_v2` requires the change.

PR #17616, opened 2025-07-05
Zephyr: Add ADC support.

bikeNomad:
### Summary

This PR adds support for ADC peripherals in the Zephyr port.

As is typical for Zephyr, the ADC channel setup is done in the devicetree (typically using an overlay).

This code requires ADC channels to be listed in the `io-channels` property of the `zephyr,user` root node like this:
```
/ {
	zephyr,user {
		io-channels = <&adc 0>, <&adc 1>, <&adc 4>, <&adc 5>, <&adc 7>;
	};
};

&adc {
	#address-cells = <1>;
	#size-cells = <0>;

	channel@0 {
		reg = <0>;
		zephyr,gain = "ADC_GAIN_1_6";
		zephyr,reference = "ADC_REF_INTERNAL";
		zephyr,acquisition-time = <ADC_ACQ_TIME_DEFAULT>;
		zephyr,input-positive = <NRF_SAADC_AIN0>; /* P0.02 */
		zephyr,resolution = <12>;
	};
       // other channels defined here (1, 4, 5, 7)
};
```

### Testing

I have tested this on the Seeed Studio XIAO BLE NRF52840 SENSE board.

This code may be enabled and disabled by the usual means in `mpconfigport.h`:
```
#define MICROPY_HW_ENABLE_ADC (1)
#define MICROPY_PY_MACHINE_ADC (1)
#define MICROPY_PY_MACHINE_ADC_INCLUDEFILE "ports/zephyr/machine_adc.c"
#define MICROPY_PY_MACHINE_ADC_READ (1)
#define MICROPY_PY_MACHINE_ADC_READ_UV (1)
```

### DIFF
--- a/ports/zephyr/README.md
+++ b/ports/zephyr/README.md
@@ -19,7 +19,8 @@ Features supported at this time:
 * `machine.Pin` class for GPIO control, with IRQ support.
 * `machine.I2C` class for I2C control.
 * `machine.SPI` class for SPI control.
-* `machine.PWM` class for PWM control
+* `machine.PWM` class for PWM control.
+* `machine.ADC` class for ADC control.
 * `socket` module for networking (IPv4/IPv6).
 * "Frozen modules" support to allow to bundle Python modules together
   with firmware. Including complete applications, including with
@@ -119,7 +120,7 @@ To blink an LED:
         time.sleep(0.5)
 
 The above code uses an LED location for a FRDM-K64F board (port B, pin 21;
-following Zephyr conventions port are identified by their devicetree node
+following Zephyr conventions ports are identified by their devicetree node
 label. You will need to adjust it for another board (using board's reference
 materials). To execute the above sample, copy it to clipboard, in MicroPython
 REPL enter "paste mode" using Ctrl+E, paste clipboard, press Ctrl+D to finish
@@ -153,6 +154,13 @@ Example of using SPI to write a buffer to the MOSI pin:
     spi.init(baudrate=500000, polarity=1, phase=1, bits=8, firstbit=SPI.MSB)
     spi.write(b'abcd')
 
+Example of using ADC to read a pin's analog value (the 'zephyr,user' node must contain
+the 'io-channels' property with all the ADC channels):
+
+    from machine import ADC
+
+    adc = ADC(("adc", 0))
+    adc.read_uv()
 
 Minimal build
 -------------
--- a/ports/zephyr/machine_adc.c
+++ b/ports/zephyr/machine_adc.c
@@ -1,0 +1,217 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 NED KONZ <NED@METAMAGIX.TECH>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+// This file is never compiled standalone, it's included directly from
+// extmod/machine_adc.c via MICROPY_PY_MACHINE_ADC_INCLUDEFILE.
+
+/*
+ * The ADC peripheral and pinmux is configured in the board's .dts file. Make
+ * sure that the ADC is enabled (status = "okay";).
+ *
+ * In addition to that, this driver requires an ADC channel specified in the
+ * io-channels property of the zephyr,user node. This is usually done with a
+ * devicetree overlay.
+ *
+ * Configuration of channels (settings like gain, reference, or acquisition time)
+ * also needs to be specified in devicetree, in ADC controller child nodes. Also
+ * the ADC resolution and oversampling setting (if used) need to be specified
+ * there.
+ *
+ * Here is an excerpt from a devicetree overlay that configures an ADC
+ * with one channel that would be referred to as ('adc', 0) in the constructor
+ * of the ADC object:
+ *
+ *  / {
+ *      zephyr,user {
+ *          io-channels = <&adc 0>;
+ *      };
+ *  };
+ *
+ *  &adc {
+ *      #address-cells = <1>;
+ *      #size-cells = <0>;
+ *
+ *      channel@0 {
+ *          reg = <0>;
+ *          zephyr,gain = "ADC_GAIN_1_6";
+ *          zephyr,reference = "ADC_REF_INTERNAL";
+ *          zephyr,acquisition-time = <ADC_ACQ_TIME_DEFAULT>;
+ *          zephyr,input-positive = <NRF_SAADC_AIN0>;
+ *          zephyr,resolution = <12>;
+ *      };
+ *      // other channels here (1, 4, 5, 7)
+ *  };
+ *
+ */
+
+#include <inttypes.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <zephyr/device.h>
+#include <zephyr/devicetree.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/printk.h>
+#include <zephyr/sys/util.h>
+#include <zephyr/drivers/adc.h>
+#include "py/mphal.h"
+#include "py/mperrno.h"
+#include "zephyr_device.h"
+
+static uint16_t sample_buffer;
+
+static struct adc_sequence adc_sequence = {
+    .options = NULL, // No options set by default
+    .channels = 0,
+    .buffer = &sample_buffer,
+    .buffer_size = sizeof(sample_buffer),   // bytes, not samples
+    .resolution = 12, // Default resolution, can be changed later
+    .oversampling = 0, // Default oversampling, can be changed later
+    .calibrate = false, // Default to no calibration
+};
+
+#define USER_NODE DT_PATH(zephyr_user)
+
+#define DT_SPEC_AND_COMMA(node_id, prop, idx) \
+    ADC_DT_SPEC_GET_BY_IDX(node_id, idx),
+
+// Data of ADC io-channels specified in devicetree.
+static const struct adc_dt_spec adc_channels[] = {
+    // We require that the user node has an io-channels property.
+    #if DT_NODE_HAS_PROP(USER_NODE, io_channels)
+    DT_FOREACH_PROP_ELEM(USER_NODE, io_channels,
+        DT_SPEC_AND_COMMA)
+    #endif
+};
+
+#define NUM_CHANNELS ARRAY_SIZE(adc_channels)
+
+typedef struct _machine_adc_obj_t {
+    mp_obj_base_t base;
+    char const *name;   // name given by the user in the tuple
+    const struct adc_dt_spec *spec; // ADC device channel specification
+} machine_adc_obj_t;
+
+
+static void mp_machine_adc_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
+    machine_adc_obj_t *self = MP_OBJ_TO_PTR(self_in);
+
+    mp_printf(print, "ADC(%s.%d)", self->name, self->spec->channel_id);
+}
+
+// constructor((adcblock, channel) | adc_obj, ...)
+static mp_obj_t mp_machine_adc_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, MP_OBJ_FUN_ARGS_MAX, true);
+
+    machine_adc_obj_t *self = NULL;
+    if (mp_obj_is_type(args[0], &machine_adc_type)) {
+        // Already an ADC object, just return it
+        self = MP_OBJ_TO_PTR(args[0]);
+    } else if (mp_obj_is_type(args[0], &mp_type_tuple)) {
+        // Get the wanted (adcblock, channel) values.
+        mp_obj_t *items;
+        mp_obj_get_array_fixed_n(args[0], 2, &items);
+        const char *dev_name = mp_obj_str_get_str(items[0]);
+        int channel_id = mp_obj_get_int(items[1]);
+        const struct device *wanted_port = zephyr_device_find(items[0]);
+
+        // Find the desired channel by device name and channel ID
+        struct adc_dt_spec const *wanted_adc_channel = NULL;
+
+        for (size_t i = 0U; i < ARRAY_SIZE(adc_channels); i++) {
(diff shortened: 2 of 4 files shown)


bikeNomad:
This PR includes @dpgeorge 's PRs #17610, #17599,  and #17552. I don't think it'll work as well without those.

jonnor:
Hi @bikeNomad and thank you for this merge request. There seems to be a number of other commits included. Would it be possible to have just the commits related to the ADC here?

dpgeorge:
I've now merged those other PRs, so please can you rebase this PR on latest master?

bikeNomad:
> I've now merged those other PRs, so please can you rebase this PR on latest master?

Done.

dpgeorge:
Thanks for the contribution here.  This looks pretty good.

Did you see that MicroPython has a `machine.ADCBlock` class?  That could possibly map better to zephyr's ADC model, eg:
```py
import machine

adc_block = machine.ADCBlock("adc")
adc = adc_block.connect(0)  # get channel 0
```
Although, admittedly, it is simpler to just write `machine.ADC(("adc", 0))` which is what this PR implements.  So maybe the `ADCBlock` is overkill here (it's more useful when you need to control the bit resolution, and connect pins to channels dynamically).

----

Regarding configuration: I suggest either enabling this feature in `mpconfigport.h` this way:
```c
#ifdef CONFIG_ADC
#define MICROPY_PY_MACHINE_ADC (1)
#define MICROPY_PY_MACHINE_ADC_INCLUDEFILE "ports/zephyr/machine_adc.c"
#define MICROPY_PY_MACHINE_ADC_READ (1)
#define MICROPY_PY_MACHINE_ADC_READ_UV (1)
#endif
```
That follows other things, like watchdog.

ports/zephyr/README.md
```diff
@@ -153,6 +154,13 @@ Example of using SPI to write a buffer to the MOSI pin:
     spi.init(baudrate=500000, polarity=1, phase=1, bits=8, firstbit=SPI.MSB)
     spi.write(b'abcd')
 
+Example of using ADC to read a pin's analog value (the 'zephyr,user' node must contain
+the 'io-channels' property with all the ADC channels):
+
+    from machine import ADC
+
+    adc = ADC(('adc', 0))
```
dpgeorge: Please use double quotes `ADC(("adc", 0))` to match examples above.

michalpiekos (approved):
LGTM

bikeNomad:
I don't understand the build error on `qemu_x86`; I've built that successfully with this branch. But I've been using Zephyr v4.1.0, and the build script seems to use v4.0.0 instead.

dpgeorge:
The build failure is actually with `frdm_k64f` (look closely at the CI output), and I can reproduce the error locally with Zephyr 4.0.0.

The problem is that `frdm_k64f` doesn't have `io-channels` defined.

But it does have an ADC and does define `adc0_default` and `adc1_default` in its device overlay/tree.  So do many other boards.

Can you make it so that the ADC driver here uses `adc0_default` (and `adc1_default` etc) so it just works out of the box with a majority of the Zephyr boards?

(I tried disabling the ADC by adding `CONFIG_ADC=n` to `ports/zephyr/boards/frdm_k64f.conf` but that didn't work for some reason... the ADC was still enabled.)

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,238 @@
...
+
+
+
+#if !DT_NODE_EXISTS(DT_PATH(zephyr_user)) || \
+    !DT_NODE_HAS_PROP(DT_PATH(zephyr_user), io_channels)
+#error "No suitable devicetree overlay specified"
+#endif
+
+#define DT_SPEC_AND_COMMA(node_id, prop, idx) \
+    ADC_DT_SPEC_GET_BY_IDX(node_id, idx),
```
michalpiekos, line 98: I am working on DAC extension for zephyr port. It also uses io-channels. Your code would not work if some of the io-channels are actually DAC channels.

ned-pcs: What do you recommend?

michalpiekos: Below is what I did for DAC. It should be easy to re-work to ADC. `make_new` takes index of a channel or channel name then I check if `dac_channels` has proper channel under given index or INVALID_CHANNEL. This is how I know it is DAC.
If DAC object is initialized with channel name then I check its index in `io_channel_names` and the rest is as above.

If something is unclear or you find some improvement just let me know. It is still work in progress.

#define INVALID_CHANNEL (-1)                                                                                                                                                                                           
#define N_CHANNELS  DT_PROP_LEN(USER_NODE, io_channels)                                                                                                                                                                
#define NAME_ELEM(node, prop, idx)  [idx] = DT_PROP_BY_IDX(node, prop, idx),                                                                                                                                           
#define IO_DEV(node, prop, idx)  DEVICE_DT_GET(DT_PHANDLE_BY_IDX(node, prop, idx)),
#define IS_DAC(node, prop, idx)  DT_PHA_HAS_CELL_AT_IDX(node, prop, idx, output),
#define DAC_ONLY(node_id, prop, idx)                                         \
    DT_PHA_BY_IDX_OR(node_id, prop, idx, output, INVALID_CHANNEL),

static const struct device *io_devs[N_CHANNELS] = {
    DT_FOREACH_PROP_ELEM(USER_NODE, io_channels, IO_DEV)
};

static const char *io_channel_names[] = {
    DT_FOREACH_PROP_ELEM(USER_NODE, io_channel_names, NAME_ELEM)
};

static const bool io_is_dac[N_CHANNELS] = {
    DT_FOREACH_PROP_ELEM(USER_NODE, io_channels, IS_DAC)
};

static const int dac_channels[] = {
    DT_FOREACH_PROP_ELEM(USER_NODE, io_channels, DAC_ONLY)
};

bikeNomad: I guess I don't understand the magic of `io-channels`. I'll look at your code.

bikeNomad:
> The problem is that `frdm_k64f` doesn't have `io-channels` defined.
> 
> But it does have an ADC and does define `adc0_default` and `adc1_default` in its device overlay/tree. So do many other boards.
> 
> Can you make it so that the ADC driver here uses `adc0_default` (and `adc1_default` etc) so it just works out of the box with a majority of the Zephyr boards?

The issue is a little more complicated than that, as @michalpiekos points out. I can easily make it so that the stock `frdm_k64f` compiles OK by checking for the existence of `io-channels` within the `/zephyr,user` node.  This would be something that would be set for a specific build (so that MicroPython, for instance, could have a `frdm_k64f.overlay` that provides access to specific ADC channels in the `io-channels` list. But there's some magic in that label, because `#io-channel-cells` is defined within the ADC and DAC nodes.

`adc0_default` and `adc1_default` refer to something else (the ADCs in the FRDM_K64F are named `adc0` and `adc1`).

I see some ADC channels being referred to in the same DTS from FRDM_K64F, but I _think_ we'd see these in the `zsensor` interface:
```
	temp0: temp0 {
		compatible = "nxp,kinetis-temperature";
		io-channels = < &adc0 0x1a >, < &adc0 0x1b >;
		io-channel-names = "SENSOR", "BANDGAP";
		bandgap-voltage = < 0xf4240 >;
		vtemp25 = < 0xaece0 >;
		sensor-slope-cold = < 0x654 >;
		sensor-slope-hot = < 0x654 >;
		status = "disabled";
	};
	temp1: temp1 {
		compatible = "nxp,kinetis-temperature";
		io-channels = < &adc1 0x1a >, < &adc1 0x1b >;
		io-channel-names = "SENSOR", "BANDGAP";
		bandgap-voltage = < 0xf4240 >;
		vtemp25 = < 0xaece0 >;
		sensor-slope-cold = < 0x654 >;
		sensor-slope-hot = < 0x654 >;
		status = "okay";
	};
```

I will try to figure out @michalpiekos 's code and adapt it for the ADC so we can both use the `io-channels` list.

ports/zephyr/machine_adc.c
ned-pcs: @michalpiekos How does your `IS_DAC()` macro work? What would be an equivalent `IS_ADC()` macro?

Would something like `DT_NODE_HAS_PROP(node_id,zephyr_gain)` work to find ADC channels (since ADC channels have a required `zephyr,gain` property)?

ned-pcs:
After hours of struggle, I can't figure out how to make this work with a mixture of ADC and DAC channels in the `zephyr,user` node's `io-channels` property. So I'm sticking with my prior code.

Without seeing @michalpiekos 's code, I can't suggest an alternative for his DAC work.

@dpgeorge I think this is ready for merging; please take a look when you have time.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+ * io-channels property of the zephyr,user node. This is usually done with a
+ * devicetree overlay.
+ *
+ * Configuration of channels (settings like gain, reference, or acquisition time)
+ * also needs to be specified in devicetree, in ADC controller child nodes. Also
+ * the ADC resolution and oversampling setting (if used) need to be specified
+ * there.
+ *
+ * Here is an excerpt from a devicetree overlay that configures an ADC
+ * with one channel that would be referred to as ('adc', 0) in the constructor
```
dpgeorge, line 44: Doesn't the below define 5 channels?  0, 1, 4, 5 and 7?

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+ *          zephyr,acquisition-time = <ADC_ACQ_TIME_DEFAULT>;
+ *          zephyr,input-positive = <NRF_SAADC_AIN0>;
+ *          zephyr,resolution = <12>;
+ *      };
+ *      // other channels here (1, 4, 5, 7)
+ *  };
+ *
+ */
+
+// #include "extmod/modmachine.h" // included from extmod/machine_adc.c
```
dpgeorge: This comment can be removed.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include <zephyr/device.h>
+#include <zephyr/devicetree.h>
+#include <zephyr/kernel.h>
+#include <zephyr/sys/printk.h>
+#include <zephyr/sys/util.h>
+#include <zephyr/drivers/adc.h>
+#include "py/mphal.h"
+#include "extmod/modmachine.h"
```
dpgeorge: This include should be able to be removed.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+#include "modmachine.h"
```
dpgeorge: I don't think this include is needed either.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+
+#define NUM_CHANNELS ARRAY_SIZE(adc_channels)
+
+typedef struct _machine_adc_obj_t {
+    mp_obj_base_t base;
+    char const *name;   // name given by the user in the tuple
+    const struct adc_dt_spec *spec; // ADC device channel specification
+} machine_adc_obj_t;
+
+// The port must provide implementations of these low-level ADC functions.
```
dpgeorge: You can probably remove this comment.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+        self = MP_OBJ_TO_PTR(args[0]);
+    } else if (mp_obj_is_type(args[0], &mp_type_tuple)) {
+        // Get the wanted (adcblock, channel) values.
+        mp_obj_t *items;
+        mp_obj_get_array_fixed_n(args[0], 2, &items);
+        const char *dev_name = mp_obj_str_get_str(items[0]);
+        int channel_id = mp_obj_get_int(items[1]);
+        const struct device *wanted_port = zephyr_device_find(items[0]);
+
+        /* Find the desired channel by device name and channel ID */
```
dpgeorge: Please use `//` style comments consistently for single line comments.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+                break;
+            }
+        }
+
+        if (!wanted_adc_channel) {
+            mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("ADC channel (%s, %d) not found"), dev_name, channel_id);
+        }
+
+        if (!adc_is_ready_dt(wanted_adc_channel)) {
+            mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("ADC device (%s) not ready"), dev_name);
```
dpgeorge: It feels like this is more of an `OSError` with errno `EBUSY` or `EIO`.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+        }
+
+        int err = adc_channel_setup_dt(wanted_adc_channel);
+        if (err < 0) {
+            mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("Could not setup ADC %s channel %d (err=%d)"), dev_name, channel_id, err);
```
dpgeorge: This also feels more like an `OSError(EIO)`.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+        mp_raise_ValueError(MP_ERROR_TEXT("ADC must be initialized with a tuple of (adcblock, channel) or an ADC object"));
+    }
+
+    return MP_OBJ_FROM_PTR(self);
+}
+
+static mp_uint_t zephyr_adc_read(const struct adc_dt_spec *spec) {
+    int err = adc_sequence_init_dt(spec, &adc_sequence);
+    if (err < 0) {
+        mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("ADC %s channel %d invalid (err=%d)"), spec->dev->name, spec->channel_id, err);
```
dpgeorge: This is definitely an `OSError`.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    err = adc_read_dt(spec, &adc_sequence);
+    if (err < 0) {
+        mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("ADC %s channel %d read failed (err=%d)"), spec->dev->name, spec->channel_id, err);
```
dpgeorge: Should be `OSError(EIO)`.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+    mp_uint_t raw = zephyr_adc_read(self->spec);
+    // Scale raw reading to 16 bit value using a Taylor expansion (for 8 <= bits <= 16)
+    mp_int_t bits = self->spec->resolution;
+    mp_uint_t u16 = raw << (16 - bits) | raw >> (2 * bits - 16);
+    return u16;
+}
+
+
+#if MICROPY_PY_MACHINE_ADC_INIT
+static void mp_machine_adc_init_helper(machine_adc_obj_t *self, size_t n_pos_args, const mp_obj_t *pos_args, mp_map_t *kw_args);
```
dpgeorge: This and the two below can be removed.  They aren't needed, and it'll correctly be a compile error if these options are enabled.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+        mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("ADC %s channel %d uv conversion failed (err=%d)"),
+            self->name, self->spec->channel_id, err);
+    }
+    return (mp_int_t)raw * 1000UL; // Convert to microvolts
+}
+#endif
+
+#if MICROPY_PY_MACHINE_ADC_ATTEN_WIDTH
+static void mp_machine_adc_atten_set(machine_adc_obj_t *self, mp_int_t atten);
+static void mp_machine_adc_width_set(machine_adc_obj_t *self, mp_int_t width);
```
dpgeorge: These two can also be removed.

ports/zephyr/machine_adc.c
```diff
@@ -0,0 +1,240 @@
...
+#if MICROPY_PY_MACHINE_ADC_BLOCK
+static mp_obj_t mp_machine_adc_block(machine_adc_obj_t *self);
+#endif
+
+#if MICROPY_PY_MACHINE_ADC_READ_UV
+static mp_int_t mp_machine_adc_read_uv(machine_adc_obj_t *self) {
+    int32_t raw = (int32_t)zephyr_adc_read(self->spec);
+    int err = adc_raw_to_millivolts_dt(self->spec, &raw);
+    if (err < 0) {
+        mp_raise_msg_varg(&mp_type_ValueError, MP_ERROR_TEXT("ADC %s channel %d uv conversion failed (err=%d)"),
```
dpgeorge: `OSError`

ports/zephyr/machine_adc.c
ned-pcs: Zephyr will return EINVAL on failure.

dpgeorge (approved):
Thanks for updating based on the review feedback.

PR #17618, opened 2025-07-05
mpprint: Rework integer vararg handling.

jepler:
### Summary

While working on format string errors in mp_printf (#17556) I experimented with adding support for %llx printing to mp_printf. In doing so, I re-organized mp_printf to increase the commonality of the various number printing possibilities.

When CI on 17556 turned up with a substantial size savings, it made me want to check if this change alone was the cause of the size savings. If so, and if the change is right, it might be worth taking independently of 17556.

### Testing

I added new test cases to the unix coverage build and ran the tests locally. However, because the controls to enable/disable support for %l and %ll are different now, this really does need testing that is sure to cover all the cases and I'm not sure I've achieved that.

In the current iteration of 17556, I am not actually using the llx format, so if this is not actually the source of the size savings reported there, then there's little reason to take it.

~~The changes to add MP_OBJ_FROM_PTR I think are "right" but they're incidental to this PR and should probably be rebased out. I briefly tried to enable the extra coverage code in the nanbox build and the build errored at these sites.~~ Some other changes to the coverage tests were broken out to https://github.com/micropython/micropython/pull/17619


dpgeorge:
Wow!  Nice code savings.

py/mpprint.c
```diff
@@ -446,16 +446,36 @@ int mp_vprintf(const mp_print_t *print, const char *fmt, va_list args) {
...
+        #if defined(MICROPY_UNIX_COVERAGE) || (LONG_MAX > INT_MAX)
+        #define SUPPORT_L_FORMAT (1)
+        #else
+        #define SUPPORT_L_FORMAT (0)
         #endif
+        #if SUPPORT_L_FORMAT
         bool long_arg = false;
+        #endif
+
+        #if ((MICROPY_OBJ_REPR == MICROPY_OBJ_REPR_D) || defined(_WIN64) || defined(MICROPY_UNIX_COVERAGE))
```
dpgeorge: The extra parenthesis around the whole expression aren't technically needed.

py/mpprint.c
```diff
@@ -501,35 +521,48 @@ int mp_vprintf(const mp_print_t *print, const char *fmt, va_list args) {
...
-                    val = va_arg(args, long int);
-                } else {
-                    val = va_arg(args, int);
-                }
-                chrs += mp_print_int(print, val, 1, 10, 'a', flags, fill, width);
-                break;
-            }
+            case 'p':
+            case 'P':
+            case 'd':
```
dpgeorge: Maybe alphabetically sort these cases?

py/mpprint.c
```diff
@@ -501,35 +521,48 @@ int mp_vprintf(const mp_print_t *print, const char *fmt, va_list args) {
...
+                #endif
+                else {
+                    if (sizeof(int) != sizeof(mp_uint_t) && fmt_chr == 'd') {
+                        val = va_arg(args, int);
+                    } else {
+                        val = va_arg(args, unsigned);
+                    }
+                }
+                int base;
+                char fmt_c = (fmt_chr & 0xf0) - 'P' + 'A'; // maps char u/x/X to char a/a/A
```
dpgeorge: If the format is 'd' then this evaluates to 'Q'.  That doesn't matter because it's not used for base 10, but needs a comment.

Also would be good to comment about the p/P formats and what they map to.

dpgeorge:
I tested this locally on PYBV10 using `./run-tests.py -t a0`.  Nothing fails.

jepler:
Thanks! It may be a few days before I can implement your review suggestions.

jepler:
Updated.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 0a4f9ec46bc0c962b5d0eff00e4ea49287fd9112 and 628d53d23cf42986bc3ae55fda2062e2016a7c2d

PR #17619, opened 2025-07-05
coverage: Add missing MP_OBJ_FROM_PTR casts.

jepler:
### Summary

An attempt to build the coverage module into the nanbox binary failed, but pointed out that these sites needed explicit conversion from pointer to object.

### Testing

I locally did a coverage build on Linux. However, only building this file as part of the nanbox build (which is not done) would reveal whether the added casts corrected the problem.


dpgeorge (approved):

dpgeorge:
Rebased and merged in da3709a73810d054bb93baf8509dcc5d246962b1

PR #17620, opened 2025-07-05
stm32: Add machine.ADC implementation for STM32L1.

yn386:
### Summary

machine.ADC needs implementation for each platform.
This PR adds implementation for STM32L1.

### Testing

Tested whether machine.ADC works with NUCLEO-L152RE.
```python
from machine import ADC

# Internal temperature sensor
adc=ADC(ADC.CORE_TEMP)
adc.read_u16()

# Internal Vrefint
adc=ADC(ADC.CORE_VREF)
adc.read_u16()
```

There is no negative impact because this change affects only for STM32L1.


dpgeorge (approved):
Thank you!

Issue #17621, opened 2025-07-06
Raspberry Pico W: Hangs after WLAN deinit()


SmokeyBrandie:
### Port, board and/or hardware

pico w

### MicroPython version

MicroPython v1.25.0 on 2025-04-15; Raspberry Pi Pico W with RP2040

### Reproduction

```
import network
import time
from time import sleep
import urequests
import gc

WLAN_SSID = 'MYSSID'
WLAN_PASSWORD = '123'

wlan = network.WLAN(network.STA_IF)

counter = 1
while True:
    print('Start round {}'.format(counter))

    wlan.active(True)
    wlan.connect(WLAN_SSID, WLAN_PASSWORD)

    for i in range(20):
        if wlan.isconnected():
            break
        print('Trying to connect... Status is: ', wlan.status())
        time.sleep(1)

    if wlan.isconnected():
        response = urequests.get('https://www.google.com')
        response.close()

    print('Deactivate WLAN...')
    print('Memory alloc {}'.format(gc.mem_alloc()))
    print('Memory free {}'.format(gc.mem_free()))

    wlan.disconnect()
    wlan.active(False)
    wlan.deinit()

    print('Sleeping...')
    time.sleep(5)
    counter += 1
```

### Expected behaviour

Should not hang

### Observed behaviour

This script works approx. 10-30 times and then it suddenly hangs. It seems to be the 
deinit() function which causes the freeze.

I tested several pico w hardware, but same problem. 

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

agatti:
For people following along at home:

```
#0  HardFault_Handler () at /micropython/lib/pico-sdk/src/rp2_common/pico_crt0/crt0.S:170
#1  <signal handler called>
#2  mdns_add_query_question_to_outpacket (outpkt=outpkt@entry=0x20041d34, msg=msg@entry=0xac) at /micropython/lib/lwip/src/apps/mdns/mdns_out.c:559
#3  0x100450fa in mdns_create_outpacket (netif=netif@entry=0x20002e98 <cyw43_state+2276>, msg=msg@entry=0xac, outpkt=outpkt@entry=0x20041d34)
    at /micropython/lib/lwip/src/apps/mdns/mdns_out.c:588
[...]

PR #17624, opened 2025-07-07
rp2/mpnetworkport: Deregister all sys timeouts when netif is removed.

dpgeorge:
### Summary

When mDNS is active on a netif it registers a lot of timeouts, namely:
    
        mdns_probe_and_announce
        mdns_handle_tc_question
    
        mdns_multicast_probe_timeout_reset_ipv4
        mdns_multicast_timeout_25ttl_reset_ipv4
        mdns_multicast_timeout_reset_ipv4
        mdns_send_multicast_msg_delayed_ipv4
        mdns_send_unicast_msg_delayed_ipv4
    
        mdns_multicast_probe_timeout_reset_ipv6
        mdns_multicast_timeout_25ttl_reset_ipv6
        mdns_multicast_timeout_reset_ipv6
        mdns_send_multicast_msg_delayed_ipv6
        mdns_send_unicast_msg_delayed_ipv6

These may still be active after a netif is removed, and if they are called they will find that the mDNS state pointer in the netif is NULL and they will crash.

These functions could be explicitly removed using `sys_untimeout()`, but `mdns_handle_tc_question()` is static so it's not possible to access it.  Instead use the new `sys_untimeout_all_with_arg()` helper to deregister all timeout callbacks when a netif is removed.

Fixes issue #17621.

### Testing

Tested with RPI_PICO_W firmware, running the test from #17621.  It's been running for over 100 cycles without any issue (prior to the fix it would crash within 10 cycles).

There's also `LWIP_NETIF_REMOVE_CALLBACK` which is a more lightweight way to do something when a netif is removed.  But `LWIP_NETIF_EXT_STATUS_CALLBACK` is already enabled so easier to use that.

This fix needs to be generalised to all ports that use lwIP, but for now I just wanted something minimal that works and is easy to review / reason about.

Note that we could probably use `LWIP_NETIF_EXT_STATUS_CALLBACK` instead of the hooks `CYW43_CB_TCPIP_INIT_EXTRA` and `CYW43_CB_TCPIP_DEINIT_EXTRA`, but that's also for follow-up work.

### DIFF
--- a/extmod/lwip-include/lwipopts_common.h
+++ b/extmod/lwip-include/lwipopts_common.h
@@ -28,6 +28,9 @@
 
 #include "py/mpconfig.h"
 
+// This is needed to access `next_timeout` via `sys_timeouts_get_next_timeout()`.
+#define LWIP_TESTMODE                   1
+
 // This sys-arch protection is not needed.
 // Ports either protect lwIP code with flags, or run it at PendSV priority.
 #define SYS_ARCH_DECL_PROTECT(lev) do { } while (0)
--- a/extmod/modnetwork.h
+++ b/extmod/modnetwork.h
@@ -82,6 +82,9 @@ extern const struct _mp_obj_type_t mp_network_ppp_lwip_type;
 #endif
 
 struct netif;
+
+void sys_untimeout_all_with_arg(void *arg);
+
 void mod_network_lwip_init(void);
 void mod_network_lwip_poll_wrapper(uint32_t ticks_ms);
 mp_obj_t mod_network_nic_ifconfig(struct netif *netif, size_t n_args, const mp_obj_t *args);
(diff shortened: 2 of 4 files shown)


dpgeorge:
A possible alternative to this PR would be to never remove netif's only add them once.  I'm not sure how easy that would be to do, but it could simplify things quite a bit, to be able to delete all the code that removes netifs.

projectgus (approved):
LGTM to fix the bug for now. The idea of never unregistering interfaces from lwip seems like a good long-term goal, though (less code size, etc).

andrewleech:
Yes I agree that never de-registering the netif  is a worthwhile future change; this is a change I've made in https://github.com/micropython/micropython/pull/17613#pullrequestreview-3010490920 for just the stm32 eth driver. 

Doing this has a few benefits; avoiding issues like this one certainly but also it means you can set / adjust the network interface config regardless of whether the interface is active. 

If/when the stm32 pr there is finalised and the architecture agreed on I'd be happy to refactor all the other network interfaces in the same way to make them all consistent.

PR #17635, opened 2025-07-07
py/asmrv32: Add support for RV32 Zba opcodes.

agatti:
### Summary

This PR adds support for RV32 [Zba](https://github.com/riscv/riscv-isa-manual/blob/main/src/b-st-ext.adoc#zba) address generation opcodes to several places in MicroPython (scaled addition operation used for calculating halfword and word offsets):

* the native emitter will emit Zba opcodes if configured to do so for Viper/Native 16- and 32-bits register-indexed load/store operations
* the inline assembler is now able to emit `SH1ADD`, `SH2ADD`, and `SH3ADD` opcodes
* the `rp2` port will emit `Zba` opcodes at runtime for RP2350 firmwares when built using the RISCV variant
* `mpy-cross` now has an extra implementation-dependent option to enable generation of `Zba` opcodes when pre-compiling native or viper code into a MPY file

~~With these changes, each load/store operation that can use these opcodes will be 2 bytes shorter.~~

With these changes, all non-byte load/store operations have been shortened by two bytes, and operations that can use these opcodes will be one instruction shorter (fewer CPU cycles and fewer register stalls).

### Testing

The inline assembler changes were tested on QEMU (to check they work), and on an ESP32C3 (to check whether the appropriate tests are skipped if those opcodes are not supported).

Runtime changes were tested on QEMU, built with and without `MICROPY_EMIT_RV32_ZBA` being set, running the full RV32 test suite.

`mpy-cross` changes were tested by looking at the emitted output code of viper ptr tests through a disassembler; both with and without `-X riscv=zba` in `mpy-cross` command line.

There's a fixed 4 bytes (or 8 bytes on 64 bit targets) RAM overhead to let the dynamic compiler settings structure hold a pointer to an emitter-specific options structure.

This mechanism, in theory, could lead to the deprecation of the various Armv7 sub-types, as they could be command line switches instead.


agatti:
Weird, the `sanitize_address` test failure doesn't seem related to my changes - my code doesn't touch threading at all (the failing test is `thread/stress_recurse.py`).

Also, QEMU's size increase wont't affect the ESP32C3 build.  The C3 doesn't have neither the inline assembler nor Zba opcodes support enabled in its default configuration (and will crash if Zba opcodes are enabled and emitted, so...).  The RP2350 in RV32 mode, however, will have a similar footprint increase as QEMU.  Considering the amount of space available on that board, it shouldn't be too much of a problem, I hope.

dpgeorge:
This looks good, thanks!  Tested on RPI_PICO2_W in RISCV mode, running the entire test suite with `--via-mpy --emit native`.  It works.

The only issue is how to deal with the selection of this zba option.  It's going to get a bit out of hand with all the possibilities that RISC-V has.

Note that `-X` options to `mpy-cross` are really only there to match the unix `micropython` executable (where, to match CPython, we need to put our special options under `-X`).  So it's possible to add other options to `mpy-cross` outside the `-X`, ie as normal options.

My suggestion would be to follow gcc and allow the `-march` option to have suffixes that specify the architecture extensions/options.  Gcc needs a base architecture, and then you can add as many options after that, optionally separated by `_`.

So this could be:
```
$ mpy-cross -march=rv32imc_zba ...
```
That exact option works with `riscv64-unknown-elf-gcc`.

What do you think?

agatti:
That could work, although extra care should be taken when displaying the available extensions from `mpy-cross`'s help to make sure those extensions are discoverable.

As a thought experiment, let's say that for MPY v7 you want to coalesce all ArmV7 targets into one single platform identifier and use this backend option handling mechanism for the thumb emitter (that may simplify the loader and natmod tooling, and maybe even save some space on the firmware side [1]).  If you follow the same rationale then you'd have to add "-mfloat-abi" and "-mfpu" command line arguments to `mpy-cross` to follow GCC.

I'd personally go for something more platform-neutral, like "-mfeatures" for example.  This leaves the door open for other platforms' configuration options in the future.

[1] And after you coalesce the armv7 platforms, if you run an armv7emdp natmod on a Netduino then you'd have the same protections as if you're running a Zba-enabled natmod on an ESP32C3 with this PR merged in :)

dpgeorge:
> If you follow the same rationale then you'd have to add "-mfloat-abi" and "-mfpu" command line arguments to `mpy-cross` to follow GCC.

Is that so bad?  Maybe it's a feature to be able to cut-and-paste GCC architecture options directly to mpy-cross?

Well, GCC does have a lot of history/legacy and so maybe its options aren't the best thing to copy.

I guess the general thing we want to do here is add "architecture feature flags", to say that the target has some feature we can utilise.

We could allow `-march` to have trailing feature flags, separated by some delimiter.

Or, as you suggest, a new option which would also be a delimiter-separated list, maybe `-march-flags`?  Or `-mfeatures` seems fine as well.

agatti:
I've implemented the architecture flags passing method using `-march-flags=...`, it actually reduced the amount of changes in `mpy-cross` whilst keeping the same amount validation.  ~~While the help message announces the fact that "-march-flags" exists, it doesn't feature detailed information yet - I'll fill that in soon, maybe along with some documentation for this flag.~~  

*Edit:*

`mpy-cross` now prints reasonable help message text for this new feature, but documentation-wise I couldn't really see a good place to fit this feature in.  The process to build native code out of a python source file via `mpy-cross` doesn't seem to be documented anywhere (f.e. looking for `-X` in `docs/` yields no mention of `mpy-cross`), but the native compilation flags are part of the Unix port's quick reference file.

The rest of `mpy-cross` features is documented in bits and pieces scattered across the various documentation files: maybe there should be a stand-alone `mpy-cross` documentation file to consolidate the documentation efforts for that tool.  I hope the current docs situation is not a blocker for this PR :)

<hr>

I also noticed the RP2 port would set the wrong `mpy-cross` flags for the RP2350 in both Arm and RV32 mode - I've added a commit to the PR to correct that.  It's in this PR because the flags it set depend on the rest of this code being merged in when building for RV32.

ports/rp2/CMakeLists.txt
```diff
@@ -487,7 +487,16 @@ list(APPEND MICROPY_SOURCE_QSTR
 )
 
 # Define mpy-cross flags
-set(MICROPY_CROSS_FLAGS -march=armv6m)
+if (${PICO_PLATFORM} STREQUAL "rp2040")
+    set(MICROPY_CROSS_FLAGS "-march=armv6m")
+elseif (${PICO_PLATFORM} STREQUAL "rp2350-arm-s")
+    set(MICROPY_CROSS_FLAGS "-march=armv7emsp")
```
dpgeorge: RP2350 uses softfp calling conventions (unfortunately).  But that's something which is out of scope for this PR.

tests/run-tests.py
```diff
@@ -763,6 +763,12 @@ def run_tests(pyb, tests, args, result_dir, num_threads=1):
                 skip_tests.add("inlineasm/thumb/asmfpmuldiv.py")
                 skip_tests.add("inlineasm/thumb/asmfpsqrt.py")
 
+        if args.inlineasm_arch == "rv32":
+            # Check if @micropython.asm_rv32 supports Zba instructions, and skip such tests if it doesn't
+            output = run_feature_check(pyb, args, "inlineasm_rv32_zba.py")
+            if output != b"rv32_zba\n":
+                skip_tests.add("inlineasm/rv32/asmzba.py")
```
dpgeorge, line 823: I don't think this skip logic works as expected: if doesn't matter if MICROPY_EMIT_RV32_ZBA is enabled or not, the inline assembler will still happily allow the `sh1add` etc opcodes to be used.

dpgeorge:
I retested this on RP2350 (in ARM and RISC-V mode) and it works.

But, I really wonder if `-march-flags` is the right direction to go in.

The benefit of the existing scheme (having many different `-march` options) is that the target will check the .mpy before running it to see if it is compatible.  Eg you can't run an `armv7m` .mpy on an `armv6m` target.  That check is not done with the `-march-flags` scheme introduced here, because the selected flags (eg whether zba is enabled or not) is not stored in the .mpy file.

A simpler approach would be to just add a new architecture `rv32imc_zba` with enum `MP_NATIVE_ARCH_RV32IMC_ZBA`.  That would be straightforward to add to mpy-cross, and would gain run-time checking of the architecture capability for free (and, eg, would allow `rv32imc` code to run on zba targets, but not `rv32imc_zba` code to run on non-zba targets).

It's true that eventually we need to add many, many different targets if we want to support all the possible combinations of RISC-V options.  But:
1. If we want to store in the .mpy the options used in the compilation, it's actually more efficient to use one big arch enum (like it is at the moment) rather than one arch enum and one set of arch-flags bits.
2. We don't really need to handle all possible combinations of RISC-V options.  We could just say we support a certain set of extension options, in the same way that the baseline RISC-V that we support is IMC.  So instead of adding `MP_NATIVE_ARCH_RV32IMC_ZBA` we could add `MP_NATIVE_ARCH_RV32IMC_Z1` which we define to be IMC+Zba+maybe-a-few-more-Z's.  We could see if there are a common set of extension options shared by RP2350 and more recent ESP32 RISC-V SoCs which could use this `MP_NATIVE_ARCH_RV32IMC_Z1`.

If instead we do opt to have two fields to specify the arch -- a base arch and feature flags -- then:
1.  We need a new .mpy format that can handle this (probably, would be hard to fit an extra field into the existing .mpy format).
2. How do we decide what goes in the base arch vs what goes in the feature flags?  Since they would both be checked when loading the .mpy they both basically play the same role.

My aim here is to try and keep things relatively simple.  Yes we are faced with quite complex architectures and instruction sets, but IMO we don't need to be as sophisticated as a C compiler.  We need to find a pragmatic balance where it's relatively easy to select the right architecture, where we can make good use of the available instruction set (but not necessarily optimal use), and where there is run-time checking of loaded .mpy files so that things don't crash unexpectedly.

agatti:
Yes, this is not an easy issue to solve indeed, and as you mention there are a number of ways to get through this and not all of them are suited to MicroPython.

TL;DR: 

It is my understanding that unless there's a mechanism to mark all the used extensions somewhere in a MPY file and it is mandatory to have a finer-grained set of checks on module load, the RV32 backend is effectively frozen.  Nothing wrong with that, really, it still performs fine given the circumstances and won't break in the future as far as RISC-V ISA or ABI are concerned.

Following RISC-V [ELF files](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#attributes), would it be acceptable to add a raw code element to MPY files containing extension information?

Longer version:

Adding the flags mechanism was supposed to also sidestep (for now) the need to bump the MPY format version due to the extra CPU features set, and all that it may entail.  A new platform enum value for each extension or combination of extensions isn't sustainable, and adding predefined extension sets doesn't really solve the problem, as you cannot really be sure a target implements *all* the extensions present in the set unless you do runtime detection at startup (which may or may not be possible depending on the environment the code runs in - also as far as I know top-level extensions can theoretically be enabled and disabled at runtime).

There is a ratified extension subset for 32 bit targets: [RVM23](https://github.com/riscv/riscv-profiles/blob/main/src/rvm23-profile.adoc).  Despite being a couple of years old, its mandatory extensions list is a *superset* of what is supported across RV32 microcontrollers currently targeted by MicroPython: RVM23-compliant code can happily crash on ESP32Cx *and* RP2350 due to unsupported opcodes that are part of the Zca extension.

Still, there's quite a bit of performance left on the table, so to speak, but users can always write C code and go wild with `-march` and `-misa` flags.  Or, if I manage to figure out a way to shoehorn the native compiler's exception handling into something that fits the SSA flow model, they can use my recompiler :)

That said, to play devil's advocate here, is a finer-grained architecture check really needed in practice?

The only case I can think of where there may not be a workaround is if somebody makes available a MIP package containing pre-built native code that used extra flags on purpose, and said package is installed on an incompatible target.  Not impossible, but that's really hard to do it by accident: assuming this PR gets in as it is now, if no special flags are passed to `mpy-cross` then nothing will change as far as tooling and generated code is concerned.

This also means that if I build a native code MPY file to run on a RP2350 and move it to an ESP32C3, if I don't explicitly add the extra flags it will still work just fine: if I'm using special platform flags it is implied that the output may or may not work on all possible targets (just with static libraries built with GCC or LLVM - although ELF files have section with RV32-specific information in there, that can be replicated in MPY files without breaking the format I guess).  Probably some more scary messages can be added to `mpy-cross`'s help and documentation to warn users when they use extra flags?

Anyway, ultimately it is not my call to make - I just want to give RV32 better chances to be more competitive w.r.t. compiled code performance (so one can write Python code to get fast enough code rather than dropping down to C).

If the necessary changes to make it happen can be detrimental to the rest of the project in its current state and there's no way out, then I'll happlily shelve this PR, maybe this can become relevant at another time :)

dpgeorge:
> A new platform enum value for each extension or combination of extensions isn't sustainable,

You don't think it's sustainable?  I think we could support 3 or 4 options pretty easily, that's only 8 to 16 new enum values, which is not too bad (even 5 options with 32 enum values would be manageable).  Is there really going to be support in the near future in MicroPython's RV32 native emitter/assembler for more than 4 Z-options?

> adding predefined extension sets doesn't really solve the problem, as you cannot really be sure a target implements _all_ the extensions present in the set unless you do runtime detection at startup

But isn't the same problem still there if flags are used?  Whether you're checking a set of flags, or a predefined set, isn't that the same code?

> There is a ratified extension subset for 32 bit targets: [RVM23](https://github.com/riscv/riscv-profiles/blob/main/src/rvm23-profile.adoc). Despite being a couple of years old, its mandatory extensions list is a _superset_ of what is supported across RV32 microcontrollers currently targeted by MicroPython:

Oh, that is a shame (that RP2350 and ESP32-C don't implement RVM23).  That would be ideal to support as a baseline.

> Or, if I manage to figure out a way to shoehorn the native compiler's exception handling into something that fits the SSA flow model, they can use my recompiler :)

Yeah, the native compiler's exception handling is highly unconventional.  It used to be different/simpler and use nested NLR blocks, but now it just uses (at most) one NLR.

> is a finer-grained architecture check really needed in practice?

Yes, I think it is.  We have all the infrastructure there already to check the main arch, so why not also check the arch flags?

> I just want to give RV32 better chances to be more competitive w.r.t. compiled code performance

Don't get me wrong, support for Zba is great!  And you've already done the majority of the hard work there.

-------------------

I see two options to move forward here (plus a third semi-option):

1. IMO the simplest thing to do is just add a new architecture `MP_NATIVE_ARCH_RV32IMC_ZBA`.  That's straightforward to do and fits in with all the existing code and tooling in a simple way.  And that will work for another couple of extensions, if any more are added.

2. In the .mpy header there is one byte for the native information.  Currently that has the lower two bits as the sub-version, and the upper 6 bits as the arch.  But only 4 of those 6 bits are used.  So we have:
```
mpy header[2] byte (as a bit pattern):
0b00aaaa00 -- aaaa = architecture, sub-version=0
0b00aaaa01 -- aaaa = architecture, sub-version=1
0b00aaaa10 -- aaaa = architecture, sub-version=2
0b00aaaa11 -- aaaa = architecture, sub-version=3
0b01xxxxxx -- 64 unused bit patterns
0b1xxxxxxx -- 128 unused bit patterns
```
So there are still 192 unused values in the space, which we can use for anything.    (Note aaaa currently only goes up to a maximum of 11.)

One way to repurpose this byte (in a backwards compatible way) is: if bit 6 is set then we are on sub-version 4 and bits 0-3 (inclusive) are the arch, and bits 4 and 5 are architecture flags.  Actually, it would be a smaller change (to existing code) to keep bits 2-5 being the arch and bits 0 and 1 being the arch flags.  Those extra two flag bits could get us quite a long way.

Alternatively, if bit 6 is clear we have sub-version=3 as before, but if bit 6 is set then we have a special RV32 sub-version and bits 0 through 5 are flags for RV32.  Or a simpler version of that: bit 6 being set means bits 2-5 are the RV32 arch flags, and bits 0-1 are still just "11" to indicate sub-version=3  (in this case we are really just defining 16 new enums for RV32 starting at 16 and going up to 31 ... and it would even be possible to use the same numbering scheme in mpy-cross to select the arch flags, `mp_dynamic_compiler_t.native_arch` is enough to store the relevant arch flags information).

3. (Not really a good option, but I'll mention it anyway) Only enable Zba support in the on-device compiler/emitter, and disable it in mpy-cross.  Then it will only be enabled on targets that support this feature and there's no need to ever manually select it, or store information about it in .mpy files.  Of course, that means you can't take advantage of Zba in precompiled (off device) code.

agatti:
> Is there really going to be support in the near future in MicroPython's RV32 native emitter/assembler for more than 4 Z-options?

I can think of four ratified candidates off the top of my head for RV32 - all supported only by the RP2340 for now (and possibly by recent versions of QEMU):
- `Zba`: address generation opcodes
- `Zcmp`: multiple-registers stack push/pop opcodes, just like Arm/Thumb does
- `Zcb`: compressed logical operations, multiplications, and non-word load/stores
- (Bonus) `B`/`Zbb`: single bit manipulation instructions (eg. bit set/bit clear) but that needs much more work to identify situations where they can be used and may take up more space than it's worth - Thumb and Xtensa also have specialised opcodes that can be used in this situation.

Although the moment you look at floating point, then there's `F` (single precision) and `D` (double precision) too.  They're not for binary code generation but would mark MPY file requirements, just like you've got separate `armv7emsp` and `armv7emdp` targets.  Right now there are no current MicroPython targets that support these - *maybe* the ESP32P4 does, but I'm on a metered connection at the moment to check that with an ESP-IDF tree checkout.  Ignoring `B` we're looking at 5 meaningful extensions that could show up in the short or mid term.

> But isn't the same problem still there if flags are used? Whether you're checking a set of flags, or a predefined set, isn't that the same code?

If you add extra meaning to some bits of the flags byte then it is effectively the same as you say.  My idea was to leave the flags byte alone, and thus having to find a different place to save that data.

Anyway, with 5 extensions then option 2 (give more structured meaning to the flags byte) seems the only viable choice.

> Alternatively, if bit 6 is clear we have sub-version=3 as before, but if bit 6 is set then we have a special RV32 sub-version and bits 0 through 5 are flags for RV32.

That fits quite nicely with the number of extensions mentioned earlier.  RV64 also supports the same extensions set, just with the addition of different opcodes to indicate usage of a wider data type (in fact adding a RV64 emitter would be relatively little work, as all RV32 code is RV64 code too).  In theory for RV64 you could assume all RV32 extensions mentioned earlier are there, as the `RVA20` profile and further revisions mandate the minimum availability of floating point code for both single and double precision values, multiplication/division, and compressed opcodes - in which case two bits could just be an index into a `[RVA20, RVA22, RVA23]` profiles set array (or just use three bits to give yourself some room for further profiles).  Yes, you *could* have a RV64 chip with no floating point, but that's uncommon to say the least - plus it wouldn't follow a ratified profile.  RV64 isn't really meant for microcontrollers anyway.

But how would the user let `mpy-cross` fill those bits in?  I can probably extend the current changes to `mpy-cross` found in this PR to encode the individual extensions set for now and then make the appropriate changes to the persistent code loader and saver and to `mpy-tool`.

If that's OK with you I can work on this and then start updating the RV64 QEMU port to use the scheme mentioned earlier and have floating point always enabled - I already made some of the necessary makefile changes when adding Arm hard floating-point targets.

dpgeorge:
> If that's OK with you I can work on this

Yes, that's fine with me.  Although I have some further suggestions:
1. We can get this PR merged much earlier by following my option (3) above and only enable Zba support in the on-device compiler/emitter, and disable it in mpy-cross (therefore no need to encode anything in the .mpy file).  That gets most of the functionality in (all of the relevant native emitter / assembler code) and doesn't need any further discussion/design.  (Or even keep the new mpy-cross `-march-flags` option and users should be aware that it may generate code which is incompatible with their target, because there are no checks for the flags at this stage.)
2. Similarly, we can merge the RV64 qemu PR #15931 without float support.
3. I have some extra things I want to add to native .mpy files (eg to properly support bss and data sections) and that needs some additional bytes of information in the .mpy file.  So I would suggest we use the mpy `header[2]` byte as follows: bit pattern `0b00aaaa11` is the existing format, but the loader also supports bit pattern `0b01aaaa11 which means that there are some extra bytes that follow in the .mpy file (or even that the rest of the .mpy file is totally different).  Then we are free to add whatever extra fields we need, eg I would suggest a variable length encoded uint for the arch flags, to allow arbitrary room for growth in the number of flags.

agatti:
Sure, MPY flag changes will be in their own PR then, once this is in.

In the meantime I'll update this to keep the changes to a minimum: this means gating the inline assembler opcodes, removing the rp2 commit (it will show up in its own PR after MPY format changes are in), and leaving `-march-flags` code in but with plenty of warnings if that stuff is used.

Still, thanks for the multiple review passes and the suggestions!

ports/rp2/CMakeLists.txt
agatti: I expected the 2350 to make use of the FPU, though!  Still, this bit is not present anymore, an updated version of this using `armv7m` as the `mpy-cross` arch target for the 2350 will show up at a later date.

tests/run-tests.py
agatti: I didn't expect the assembler to perform those checks... Oh well, this should be addressed now.

Since I've let the QEMU port use `Zba` opcodes by default and made `mpy-cross` aware of `Zba` only at runtime, the latter can be used to test whether the gating I've added actually works:

```bash
[/micropython/ports/qemu] $ /micropython/mpy-cross/build/mpy-cross -X emit=native -march=rv32imc -march-flags=zba ../../tests/inlineasm/rv32/asmzba.py
Warning: Using architecture-specific flags may create a MPY file whose code
         won't run on all target boards!
         Currently there are no checks in the module file loader for whether
         the chosen flags used to build the MPY file are compatible with the
         running target.

[/micropython/ports/qemu] $ /micropython/mpy-cross/build/mpy-cross -X emit=native -march=rv32imc ../../tests/inlineasm/rv32/asmzba.py  
  File "../../tests/inlineasm/rv32/asmzba.py", line 3, in test_sh1add
SyntaxError: invalid RV32 instruction 'sh1add'
```

agatti:
OK, I think this is it.  Inline assembler checks are in, `mpy-cross` produces a warning on STDERR whenever `-march-flags` would affect the output MPY file, and QEMU generates `Zba` opcodes by default so that bit can be tested.  I even fixed a couple of issues that went under the radar (`-march-flags` didn't like not having an architecture being set, and now *all* RV32 load/stores for halfwords/words that do not use `Zba` are two bytes shorter :)).

I'm still not 100% sure if [`43eedba` (#17635)](https://github.com/micropython/micropython/pull/17635/commits/43eedba1a1b716bc4e279a17f75761dccdc119eb) is the right way to keep the interpreter and `mpy-cross` synchronised when it comes to architecture flags for QEMU, but I didn't really find any other way to achieve that with minimal code changes.

mpy-cross/main.c
```diff
@@ -131,6 +139,8 @@ static int usage(char **argv) {
         "-msmall-int-bits=number : set the maximum bits used to encode a small-int\n"
         "-march=<arch> : set architecture for native emitter;\n"
         "                x86, x64, armv6, armv6m, armv7m, armv7em, armv7emsp, armv7emdp, xtensa, xtensawin, rv32imc, host, debug\n"
+        "-march-flags=<flags> : set architecture-specific flags (OUTPUT FILE MAY NOT WORK ON ALL TARGET BOARDS)\n"
```
dpgeorge: Suggest using just "TARGETS" instead of "TARGET BOARDS".

mpy-cross/main.c
```diff
@@ -350,6 +364,29 @@ MP_NOINLINE int main_(int argc, char **argv) {
...
+            }
+        }
+        #endif
+        if (!processed) {
+            mp_printf(&mp_stderr_print, "unrecognised arch flags\n");
+            exit(1);
+        }
+        mp_printf(&mp_stderr_print,
+            "Warning: Using architecture-specific flags may create a MPY file whose code\n"
+            "         won't run on all target boards!\n"
```
dpgeorge: "targets" instead of "target boards"

py/emitinlinerv32.c
```diff
@@ -37,6 +37,23 @@
...
+#define EXTENSION_BIT_ZBA (0)
+
+#if MICROPY_EMIT_RV32_ZBA
+#define EXTENSION_ZBA_FRAGMENT (1 << EXTENSION_BIT_ZBA)
+#else
+#define EXTENSION_ZBA_FRAGMENT (0)
+#endif
+
+#define ALLOWED_EXTENSIONS (EXTENSION_ZBA_FRAGMENT)
+static uint32_t emit_inline_rv32_allowed_extensions = 0;
```
dpgeorge: It looks like this variable is only needed for `MICROPY_DYNAMIC_COMPILER`, and so I suggest guarding it in such a `#if MICROPY_DYNAMIC_COMPILER`.

Actually, that's still a bit messy and essentially duplicates the `asm_rv32_allow_zba_opcodes()` function added to `asmrv32.c`.

So my suggestion would be to follow and extend how Thumb does it with the public `asm_thumb_allow_armv7m()` function (which is used by `asmthumb.c` and `emitinlinethumb.c`, and even `emitnative.c`):
- standardise in `asmrv32h` a public set of bit fields (eg enum) for supported RV32 extensions
- change the `asm_rv32_backend_options_t` struct to have a `uint8_t rv32_ext;` entry which holds a bit mask of enabled extensions (instead of bools for each flag)
- expose a public function in `asmrv32.h` (maybe `static inline`) called something like `asm_rv32_allowed_extensions()` which is similar to how `asm_rv32_allow_zba_opcodes` is currently implemented and can replace (or be used to implement) that function

Eg in `asmrv32.h`:
```c
enum {
   RV32_EXT_ZBA = 1 << 0,
   ...
};

static inline uint8_t asm_rv32_allowed_extensions(void) {
    uint8_t ext = 0;
    #if MICROPY_DYNAMIC_COMPILER
    if (mp_dynamic_compiler.backend_options != NULL) {
        ext |= ((asm_rv32_backend_options_t *)mp_dynamic_compiler.backend_options)->ext;
    }
    #else
    #if MICROPY_EMIT_RV32_ZBA
    ext |= RV32_EXT_ZBA;
    #endif
    #endif
    return ext;
}

static inline bool asm_rv32_allow_zba_opcodes(void) {
    return asm_rv32_allowed_extensions() & RV32_EXT_ZBA;
}

File: py/emitinlinerv32.c [hunk 6]
```diff
@@ -220,6 +245,9 @@ typedef void (*call_i_t)(asm_rv32_t *state, mp_int_t immediate);
 typedef void (*call_r_t)(asm_rv32_t *state, mp_uint_t rd);
 typedef void (*call_n_t)(asm_rv32_t *state);
 
+#define EXT_NIL (0)
+#define EXT_ZBA (1 << (EXTENSION_BIT_ZBA))
```
dpgeorge: As suggested above, these could be moved to `asmrv32.h` and reused by mpy-cross etc.

py/emitinlinerv32.c
```diff
@@ -785,10 +817,11 @@ static void emit_inline_rv32_opcode(emit_inline_asm_t *emit, qstr opcode, mp_uin
         }
     }
 
-    if (!opcode_data) {
-        emit_inline_rv32_error_exc(emit,
-            mp_obj_new_exception_msg_varg(&mp_type_SyntaxError,
-                MP_ERROR_TEXT("unknown RV32 instruction '%q'"), opcode));
+    if (!opcode_data ||
+        (((opcode_data->required_extensions & ALLOWED_EXTENSIONS) != opcode_data->required_extensions) &&
+         ((opcode_data->required_extensions & emit_inline_rv32_allowed_extensions) != opcode_data->required_extensions))) {
```
dpgeorge: As suggested above, this could be simplified to `opcode_data->required_extensions & asm_rv32_allowed_extensions() != opcode_data->required_extensions`.

py/emitinlinerv32.c
agatti: It looks better the way you suggested - and that's what I've followed in the newest iteration of the PR.  It also saved 8 bytes.

The original plan was to have an integer-backed bitfield as the backend options payload, but since at the moment there's just one single entry there wasn't much benefit in having a `uint8_t emit_zba : 1;` declaration instead of using a single boolean value.

py/emitinlinerv32.c
agatti: This should have been addressed in the latest version of the PR.

py/emitinlinerv32.c
agatti: This should have been addressed in the latest version of the PR.

mpy-cross/main.c
agatti: Thanks!  In my mind target boards was more appropriate as loading a RV32 MPY file on anything but a RV32 board wouldn't have worked anwyay :)

mpy-cross/main.c
agatti: This should have been addressed in the latest version of the PR.

dpgeorge (approved):
Thanks for following my suggestion and updating the code.  It looks good now.

And thanks overall for persisting with this PR, it now sets the direction to add arch-specific extensions to the emitter.  And having zba for RV32 targets is a nice enhancement.

PR #17638, opened 2025-07-08
unix/make: Drop i686-linux-gnu path.

yf13:
### Summary

This drops use of non-existing path `/usr/include/i686-linux-gnu` as default toolchain include paths shall suffice for 32bit.

### Testing

Checked locally on amd64 PC host running Ubuntu 22.04 with gcc v11.4 and following configs:

- `make -C ports/unix VARIANT=nanbox MICROPY_PY_FFI=1 test -j4`
- `make -C ports/unix MICROPY_FORCE_32BIT=1 test -j4`
- `make -C ports/unix test -j4`

### DIFF
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -120,16 +120,6 @@ LDFLAGS += $(LDFLAGS_MOD) $(LDFLAGS_ARCH) -lm $(LDFLAGS_EXTRA)
 # Flags to link with pthread library
 LIBPTHREAD = -lpthread
 
-ifeq ($(MICROPY_FORCE_32BIT),1)
-# Note: you may need to install i386 versions of dependency packages,
-# starting with linux-libc-dev:i386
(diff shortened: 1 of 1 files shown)


tests/extmod/select_poll_eintr.py
```diff
@@ -24,6 +24,7 @@
 
 
 def thread_main():
+    global lock
```
dlech: `lock` is not assigned in this function, so we don't need a `global` statement.

dpgeorge:
Thanks for the contribution.

The problem with `select_poll_eintr.py` is described in https://github.com/micropython/micropython/pull/11604#issuecomment-1955358151 .  And I worked around that in #17655.  So I think you can drop your changes to that file from this PR and it'll get fixed separately.

yf13:
@dpgeorge, thank you for the information. I've dropped it. Please let me know if there are more things to do with this PR.

dpgeorge (approved):

PR #17639, opened 2025-07-08
py/objcode: Remove co_lnotab from v2 preview.

AJMansfield:
### Summary

Starting in CPython 3.12, accessing the `co_lnotab` attribute on code objects results in a deprecation message; and in theory, attempting to use it will be an AttributeError in CPython 3.15. Likewise, the corresponding provision in [PEP-626](https://peps.python.org/pep-0626/#backwards-compatibility) states that "the co_lnotab attribute will be deprecated in 3.10 and removed in 3.12."

Once #16989 is merged (to implement the `co_lines` method that `co_lnotab` is being removed in favor of), Micropython should follow suit to deprecate and remove `co_lnotab` from our v2 previews (and eventually, the Micropython 2.0 release).

This PR does this, adding `MICROPY_PREVIEW_VERSION_2` macro gates to the `co_lnotab` attr entry and corresponding function implementation.

### Testing

This PR adds a new set of automated test cases to validate the v2 removal and pre-v2 non-removal of `co_lnotab`, replacing a test from #16989 that only applies to pre-v2.
I've run these tests and verified correct behavior on both v2-preview and non-preview settrace builds of the unix port.

Aside from the expected compatibility break with v2 preview, there are no downsides to this. This is a strict improvement to code size on settrace builds, and should have no effect on standard builds.

### DIFF
--- a/tests/basics/fun_code_lnotab.py
+++ b/tests/basics/fun_code_lnotab.py
@@ -1,0 +1,34 @@
+# Test deprecation of co_lnotab
+
+try:
+    (lambda: 0).__code__.co_code
+except AttributeError:
+    print("SKIP")
+    raise SystemExit
+
+
+import unittest
+import sys
+
+
+mpy_is_v2 = getattr(sys.implementation, '_v2', False)
+
+
+def f():
+    pass
+
+
+class Test(unittest.TestCase):
+
+    @unittest.skipIf(mpy_is_v2, "Removed in MicroPython v2 and later.")
+    def test_co_lnotab_exists(self):
+        self.assertIsInstance(f.__code__.co_lnotab, bytes)
+
+    @unittest.skipUnless(mpy_is_v2, "Not removed before MicroPython v2.")
+    def test_co_lnotab_removed(self):
+        with self.assertRaises(AttributeError):
+            f.__code__.co_lnotab
+
+
+if __name__ == "__main__":
+    unittest.main()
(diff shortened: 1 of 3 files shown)


AJMansfield:
The intended commits here are only the last two:
-  tests/basics/fun_code_lnotab: Test removal of co_lnotab in v2.
- py/objcode: Remove co_lnotab from v2 preview.

Once #16989 is merged I'll rebase this so it doesn't include the commits from it that this PR depends on.

dpgeorge:
This can now be rebased.

tests/basics/fun_code_lnotab.py
```diff
@@ -0,0 +1,30 @@
...
+import unittest
+import sys
+
+
+def f():
+    pass
+
+
+class Test(unittest.TestCase):
+    @unittest.skipIf(getattr(sys.implementation,'_v2',False), "Removed in Micropython v2 and later.")
```
dpgeorge: Suggest factoring the v2 test out to a global variable so it's a little more readable:
```py
mpy_is_v2 = getattr(sys.implementation,'_v2',False)
```

Also, please use uppercase P in MicroPython.

AJMansfield: > Also, please use uppercase P in MicroPython.

Noted for the future!

dpgeorge (approved):

PR #17642, opened 2025-07-08
examples/natmod/btree: Fix build on RV32 with Picolibc.

agatti:
### Summary

This PR fixes building the "btree" example natmod on RV32 when Picolibc is being used and uses thread-local storage for storing the errno variable.

The fix is surprisingly simple: Picolibc allows overriding the function that will provide a pointer to the "errno" variable, and the btree natmod integration code already has all of this machinery set up as part of its library integration.  Redirecting Picolibc to the already existing pointer provider function via a compile-time definition is enough to let the module compile and pass QEMU tests.
 
This workaround will work on any Picolibc versions (Arm, RV32, Xtensa, etc.) even if TLS support was not enabled to begin with, and will effectively do nothing if the toolchain used will rely on Newlib to provide standard C library functions.

Given that the btree module now builds and passes the relevant natmod tests, said module is now part of the QEMU port's natmod testing procedure, and CI now will build the btree module for RV32 as part to its checks.

### Testing

The `btree` natmod was built successfully when targeting RV32, and then QEMU's `make test_natmod` btree tests pass.  Relevant CI tests also pass in my repo's branch (and hopefully will pass on this branch too!).

### DIFF
--- a/examples/natmod/btree/Makefile
+++ b/examples/natmod/btree/Makefile
@@ -36,6 +36,9 @@ ifeq ($(ARCH),xtensa)
 MPY_EXTERN_SYM_FILE=$(MPY_DIR)/ports/esp8266/boards/eagle.rom.addr.v6.ld
 endif
 
+# Use our own errno implementation if Picolibc is used
+CFLAGS += -D__PICOLIBC_ERRNO_FUNCTION=__errno
+
 include $(MPY_DIR)/py/dynruntime.mk
 
 # btree needs gnu99 defined
--- a/docs/develop/natmod.rst
+++ b/docs/develop/natmod.rst
@@ -67,6 +67,9 @@ The known limitations are:
 
 * static BSS variables are not supported; workaround: use global BSS variables
 
+* thread-local storage variables are not supported on rv32imc; workaround: use
+  global BSS variables or allocate some space on the heap to store them
+
 So, if your C code has writable data, make sure the data is defined globally,
 without an initialiser, and only written to within functions.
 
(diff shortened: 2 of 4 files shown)


dpgeorge:
Very nice!

examples/natmod/btree/Makefile
```diff
@@ -36,6 +36,10 @@ ifeq ($(ARCH),xtensa)
 MPY_EXTERN_SYM_FILE=$(MPY_DIR)/ports/esp8266/boards/eagle.rom.addr.v6.ld
 endif
 
+ifeq ($(ARCH),rv32imc)
+CFLAGS += -D__PICOLIBC_ERRNO_FUNCTION=__errno
+endif
```
dpgeorge: Maybe this can go in `py/dynruntime.mk` so it's available for everyone?  It seems pretty harmless to add unconditionally.

agatti: Not totally sure on that.  This works for `btree` only because there's this bit in the natmod's code:

```c
int native_errno;
#if defined(__linux__)
int *__errno_location (void)
#else
int *__errno (void)
#endif
{
    return &native_errno;
}
```

Hardcoding the function name for something like this maybe it's not the best way, especially if you've had to reimplement `errno`'s behaviour to be able to integrate a much more complex bit of code in a natmod.  Sure, this can work through another layer of indirection using a configuration define so you can easily change `__errno` with something else only if picolibc is enabled, but at this point we're just recreating CMake and not in a good way :)

Something I can do, though, is make this define unconditional for this specific natmod - if you use picolibc for Arm or Xtensa you may still encounter the same issue if picolibc was built with TLS support enabled and its errno implementation is configured to use it.

Oh, I almost forgot, I should probably document this behaviour somewhere in `docs/develop/natmod.rst` as it's not entirely obvious how to solve this issue if you encounter it and expect something as simple as `errno` to just work.

**Edit:** Done!

dpgeorge: > Something I can do, though, is make this define unconditional for this specific natmod

OK, that's a good solution.  The `btree` natmod is after all the only one using `errno`.

> Oh, I almost forgot, I should probably document this behaviour somewhere in `docs/develop/natmod.rst`

Docs update looks good!

dpgeorge (approved):

PR #17643, opened 2025-07-08
py/parse: Recognise const int assignments with type hints.

### LINKED ISSUE #15608: Const with type hints is not optimized

Raspberry Pi Pico W, esp32

### MicroPython version

MicroPython v1.23.0 on 2024-06-02; Raspberry Pi Pico W with RP2040

### Reproduction

**Not optimized**

1. Save file
```
from micropython import const

_MY_CONST1: int = const(1)
_MY_CONST2: int = const(2)
_MY_CONST3: int = const(2)


def print_const():
    print(f"1:{_MY_CONST1} 2:{_MY_CONST2} 3:{_MY_CONST3}")

```
2. run `mpy-cross c.py`
3. run `mpy-tool.py -d c.mpy`
```
mpy_source_file: c.mpy
source_file: c.py
header: 4d:06:00:1f
qstr_table[10]:
    c.py
    <module>
    const
    micropython
    print_const
    format
    _MY_CONST1
    _MY_CONST2
    _MY_CONST3
    print
obj_table: ['1:{} 2:{} 3:{}']
simple_name: <module>
  raw bytecode: 46 08:0a:01:4c:27:27:67:80:10:02:2a:01:1b:03:1c:02:16:02:59:11:02:81:34:01:16:06:11:02:82:34:01:16:07:11:02:82:34:01:16:08:32:00:16:04:51:63
  prelude: (2, 0, 0, 0, 0, 0)
  args: []
  line info: 4c:27:27:67
  80          LOAD_CONST_SMALL_INT 0
  10:02       LOAD_CONST_STRING const
  2a:01       BUILD_TUPLE 1
  1b:03       IMPORT_NAME micropython
  1c:02       IMPORT_FROM const
  16:02       STORE_NAME const
  59          POP_TOP
  11:02       LOAD_NAME const
  81          LOAD_CONST_SMALL_INT 1
  34:01       CALL_FUNCTION 1
  16:06       STORE_NAME _MY_CONST1
  11:02       LOAD_NAME const
  82          LOAD_CONST_SMALL_INT 2
  34:01       CALL_FUNCTION 1
  16:07       STORE_NAME _MY_CONST2
  11:02       LOAD_NAME const
  82          LOAD_CONST_SMALL_INT 2
  34:01       CALL_FUNCTION 1
  16:08       STORE_NAME _MY_CONST3
  32:00       MAKE_FUNCTION 0
  16:04       STORE_NAME print_const
  51          LOAD_CONST_NONE
  63          RETURN_VALUE
  children: ['print_const']
simple_name: print_const
  raw bytecode: 24 28:06:04:80:08:12:09:23:00:14:05:12:06:12:07:12:08:36:03:34:01:59:51:63
  prelude: (6, 0, 0, 0, 0, 0)
  args: []
  line info: 80:08
  12:09       LOAD_GLOBAL print
  23:00       LOAD_CONST_OBJ '1:{} 2:{} 3:{}'
  14:05       LOAD_METHOD format
  12:06       LOAD_GLOBAL _MY_CONST1
  12:07       LOAD_GLOBAL _MY_CONST2
  12:08       LOAD_GLOBAL _MY_CONST3
  36:03       CALL_METHOD 3
  34:01       CALL_FUNCTION 1
  59          POP_TOP
  51          LOAD_CONST_NONE
  63          RETURN_VALUE
  children: []
```

**Optimized**

**Not optimized**

1. Save file
```
from micropython import const

_MY_CONST1 = const(1)
_MY_CONST2 = const(2)
_MY_CONST3 = const(2)


def print_const():
    print(f"1:{_MY_CONST1} 2:{_MY_CONST2} 3:{_MY_CONST3}")

```
2. run `mpy-cross c.py`
3. run `mpy-tool.py -d c.mpy`
```
mpy_source_file: c.mpy
source_file: c.py
header: 4d:06:00:1f
qstr_table[7]:
    c.py
    <module>
    const
    micropython
    print_const
    format
    print
obj_table: ['1:{} 2:{} 3:{}']
simple_name: <module>
  raw bytecode: 23 08:06:01:8c:07:80:10:02:2a:01:1b:03:1c:02:16:02:59:32:00:16:04:51:63
  prelude: (2, 0, 0, 0, 0, 0)
  args: []
  line info: 8c:07
  80          LOAD_CONST_SMALL_INT 0
  10:02       LOAD_CONST_STRING const
  2a:01       BUILD_TUPLE 1
  1b:03       IMPORT_NAME micropython
  1c:02       IMPORT_FROM const
  16:02       STORE_NAME const
  59          POP_TOP
  32:00       MAKE_FUNCTION 0
  16:04       STORE_NAME print_const
  51          LOAD_CONST_NONE
  63          RETURN_VALUE
  children: ['print_const']
simple_name: print_const
  raw bytecode: 21 28:06:04:80:08:12:06:23:00:14:05:81:82:82:36:03:34:01:59:51:63
  prelude: (6, 0, 0, 0, 0, 0)
  args: []
  line info: 80:08
  12:06       LOAD_GLOBAL print
  23:00       LOAD_CONST_OBJ '1:{} 2:{} 3:{}'
  14:05       LOAD_METHOD format
  81          LOAD_CONST_SMALL_INT 1
  82          LOAD_CONST_SMALL_INT 2
  82          LOAD_CONST_SMALL_INT 2
  36:03       CALL_METHOD 3
  34:01       CALL_FUNCTION 1
  59          POP_TOP
  51          LOAD_CONST_NONE
  63          RETURN_VALUE
  children: []

```

### Expected behaviour

`_MY_CONST1: int = const(1)` should be optimized as `_MY_CONST1 = const(1)`

### Observed behaviour

PyCharm supports type hints, which is very useful. When using type hints e.g. `_MY_CONST1: int = const(1)` optimizer does not work as expected (used `mpy-cross` and `mpy-tool` to show the content of .mpy file). In `.mpy` file you can see `_MY_CONST1` in qstr table, `STORE_NAME _MY_CONST1` and `LOAD_NAME const`. The result file has 162 bytes

When not using type hints e.g. `_MY_CONST1 = const(1)` everything is fine, the code has been optimized, and you can see in `.mpy` file `LOAD_CONST_SMALL_INT 1` instead of  `_MY_CONST1` (100 bytes), 

### Additional Information

Used `mpy_cross 1.24.0rc0` as optimizer

### Code of Conduct

Yes, I agree

agatti:
### Summary

This PR extends the parser's behaviour to also recognise constant integer assignments even when the assignment itself has an "int" type annotation.

Now declarations like "var: int = const(val)" can be treated as constants by the compiler and thus be folded in the rest of the program as it sees fit.  Before this change, that line would generate a variable creation and its value assignment, without any folding being performed.

This fixes #15608.

### Testing

This was tested using the same verification methodology described in #15608, and the Unix test suite was ran locally without errors with this change applied.

I've limited the type checking to `int` types, which I hope it's all that's needed.  It can probably be improved style-wise, but from what I reckon the logic *should* be correct - I haven't looked at the parser since the rv32 inline asm days.


agatti:
Forgot to add tests... thank you, codecov!  I've also had the chance to reword the commit message as it wasn't entirely clear.  I'll push the lot once the current test run is over to avoid having to receive loads of cancelled run reports.

Anyway, I wonder how difficult it would be to have something like `mpy-tool.py`'s disassembler available in the tests' running environment.  This way the tests could also verify that, for a given bit of code, optimised bytecode was emitted instead of regular code sequences - that would make some sense for this specific PR at least.

dlech:
Would it work to do this in the Python code instead?

```python
_MY_CONST: int
_MY_CONST = const(123)
```

If it works, it might be worth it to save those 80+ bytes.

agatti:
I've just tried, and it does indeed work - after all the type annotations are discarded anyway.  Also, I've simplified the code and turned out the final overhead is bigger, go figure.  I'll revert the PR to what it looked like then.

From an end user's point of view, having annotated variables exhibit the same behaviour as non-annotated ones is probably a reasonable expectation, isn't it?

dlech:
Yes, a reasonable expectation. And the compiler can be disabled on super-constrained systems. So the code size isn't as much of a concern as I initially made it to be.

dpgeorge:
> I've limited the type checking to `int` types, which I hope it's all that's needed.

Actually, const supports bool and tuple.  And also now float!  So the type annotation could be rather complicated if you have nested tuples etc.

So... the parser could just support any type annotation, ie any expression.  Or just any identifier.



> Anyway, I wonder how difficult it would be to have something like `mpy-tool.py`'s disassembler available in the tests' running environment. This way the tests could also verify that, for a given bit of code, optimised bytecode was emitted instead of regular code sequences

There are tests for generated bytecode, in particular one for const generation.  See `tests/cmdline/cmd_showbc_const.py`.

py/parse.c
```diff
@@ -797,21 +806,26 @@ static bool fold_constants(parser_t *parser, uint8_t rule_id, size_t num_args) {
         if (!MP_PARSE_NODE_IS_NULL(pn1)
             && !(MP_PARSE_NODE_IS_STRUCT_KIND(pn1, RULE_expr_stmt_augassign)
                  || MP_PARSE_NODE_IS_STRUCT_KIND(pn1, RULE_expr_stmt_assign_list))) {
-            // this node is of the form <x> = <y>
+            // this node is of the form <x> = <y> or <x> : int = <y>
             mp_parse_node_t pn0 = peek_result(parser, 1);
             if (MP_PARSE_NODE_IS_ID(pn0)
-                && MP_PARSE_NODE_IS_STRUCT_KIND(pn1, RULE_atom_expr_normal)
```
dpgeorge, line 826: Could you reduce code size by doing something like this:
```c
if (MP_PARSE_NODE_IS_ID(pn0)) {
    if (MP_PARSE_NODE_IS_STRUCT_KIND(pn1, RULE_annassign)
        && MP_PARSE_NODE_IS_ID(((mp_parse_node_struct_t *)pn1)->nodes[0])
        && MP_PARSE_NODE_LEAF_ARG(((mp_parse_node_struct_t *)pn1)->nodes[0]) == MP_QSTR_int) {
        // extract annassign rhs
        pn1 = ((mp_parse_node_struct_t *)pn1)->nodes[1];
    }
    if (is_const_atom_expr_normal(pn1)) {
        // code to assign dynamic constants: id = const(value)
        ... existing code
    }
}
```

agatti:
> Actually, const supports bool and tuple. And also now float! So the type annotation could be rather complicated if you have nested tuples etc.

Indeed.  Luckily I limited myself to just handle `int` types with this!  I've tried to extend the const condition check to cover both `MP_QSTR_int` and `MP_QSTR_float` as the checked annotation type but, unless I did the test incorrectly, it didn't seem to work as expected.

I can revisit this at another time though, so both `float` and `bool` can be handled this way as well.

> There are tests for generated bytecode, in particular one for const generation. See tests/cmdline/cmd_showbc_const.py.

Thanks, I didn't know this was possible.  Although what I had in mind was more or less a MicroPython implementation of CPython's `dis` module so unrelated future changes in the compiled output would not break the test unless they affect const loads.

py/parse.c
agatti: This commit's footprint was cut in half with this, thanks!  I've implemented this change in the current version of the PR.

dlech:
Why do we need to check the annotation value and limit it to only `int`? Can't we just remove `&& MP_PARSE_NODE_LEAF_ARG(((mp_parse_node_struct_t *)pn1)->nodes[0]) == MP_QSTR_int)` and basically ignore the annotation to allow all const expressions with annotations?

dpgeorge:
@agatti what do you think about @dlech 's suggestion above, to just allow any annotation?  That's needed because MicroPython's consts can be more than just `int`.

agatti:
Gosh, I missed that - sorry!  I'm all for it, let me update the PR right away then.

py/parse.c
```diff
@@ -820,50 +829,52 @@ static bool fold_constants(parser_t *parser, uint8_t rule_id, size_t num_args) {
...
-                // If the constant starts with an underscore then treat it as a private
-                // variable and don't emit any code to store the value to the id.
-                if (qstr_str(id)[0] == '_') {
-                    pop_result(parser); // pop const(value)
-                    pop_result(parser); // pop id
-                    push_result_rule(parser, 0, RULE_pass_stmt, 0); // replace with "pass"
-                    return true;
+            if (MP_PARSE_NODE_IS_ID(pn0)) {
+                if (MP_PARSE_NODE_IS_STRUCT_KIND(pn1, RULE_annassign)
+                    && MP_PARSE_NODE_IS_ID(((mp_parse_node_struct_t *)pn1)->nodes[0])) {
```
dpgeorge: I think this `MP_PARSE_NODE_IS_ID()` can be dropped altogether, and then this code will work with any annotations, eg:
```py
x: tuple[int, float] = const((1, 2.3))
```

agatti: In this case then the code in `is_const_atom_expr_normal` can be folded back into the main function body, as it is only invoked once.  I'll look into it, thanks!

dpgeorge:
Thanks for updating, the code changes look good now.

Regarding the tests: it looks like you've copied two existing tests and added annotations.  I think that's overkill and makes it harder to maintain the tests.  Instead I suggest adding one new test called, eg, `const_annotated.py` which just tests annotations of various kinds, eg:
```py
# Test type annotations in combination with const.

from micropython import const

_X0: bool = const(True)
_X1: int = const(123)
_X2: str = const("test")
_X3: tuple = const((1, 2))
_X4: tuple[bool, int] = const((True, 4))
```
(Keep in mind not to use float because not all targets support floats.)

agatti:
Thanks for the test template.  The previous iteration duplicated the original tests because they only dealt with integers in the first place, so that would fit with the initial reduced scope of these changes.

Anyway, I've added a missing const-able type (`bytes`) to your suggestion and removed the rest of the duplicated code.

dpgeorge (approved):
Thanks for updating the test.

PR #17649, opened 2025-07-09
build/sanitizer: reserve 8 KiB C-stack margin for sanitizer builds

koxudaxi:
### Summary

This PR moves the sanitizer-only changes that reviewers asked to split out of
#17557.  
When a build is compiled with AddressSanitizer or UndefinedSanitizer the
compiler enlarges each C call frame; MicroPython’s stack-overflow guard can
therefore trigger before the real end of the stack.  To avoid that, the patch
sets `MICROPY_STACK_CHECK_MARGIN` to **8192 bytes** whenever the usual sanitizer
macros are present.

Files touched

* `py/mpconfig.h` – default margin for builds that define the sanitizer macros  
* `ports/unix/mpconfigport.h` – same margin applied to the unix port  
* `tests/thread/thread_stacksize1.py` – larger stack requested on desktop
  targets where sanitizers are common  
* `docs/library/micropython.rst` – short note explaining the new default

Builds without sanitizers behave exactly as before.

### Testing

* **unix port (macOS, Clang 18.1)**  
  * normal build: `tools/run-tests.py` → all tests pass  
  * `-fsanitize=address,undefined`:  
    * `tools/run-tests.py tests/thread/thread_stacksize1.py` passes  
    * no premature “stack overflow” exceptions  
    * UBSan prints the same null-pointer warnings that are already present on
      the main branch; they are unrelated to this change
* Other ports were not rebuilt; they keep the default margin of zero and are
  unaffected.

### Trade-offs and alternatives

Sanitizer builds lose 8 KiB of usable C stack.  A smaller margin (4 KiB) was
occasionally insufficient in deep recursion tests; larger values offered no
extra benefit.  Each port can still override `MICROPY_STACK_CHECK_MARGIN` if a
different value is required.

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -701,6 +701,16 @@
 // Additional margin between the places in the runtime where Python stack is
 // checked and the actual end of the C stack. Needs to be large enough to avoid
 // overflows from function calls made between checks.
+#if !defined(MICROPY_STACK_CHECK_MARGIN)
+#if defined(__SANITIZE_ADDRESS__) || defined(__SANITIZE_UNDEFINED__)
+#define MICROPY_STACK_CHECK_MARGIN (8192)
+#elif defined(__has_feature)
+#if __has_feature(address_sanitizer) || __has_feature(undefined_sanitizer)
+#define MICROPY_STACK_CHECK_MARGIN (8192)
+#endif
+#endif
+#endif
+
 #ifndef MICROPY_STACK_CHECK_MARGIN
 #define MICROPY_STACK_CHECK_MARGIN (0)
 #endif
--- a/tests/thread/thread_stacksize1.py
+++ b/tests/thread/thread_stacksize1.py
@@ -7,7 +7,11 @@ import _thread
 
 # different implementations have different minimum sizes
 if sys.implementation.name == "micropython":
-    sz = 2 * 1024
+    # Use larger stack size for desktop platforms to accommodate sanitizers
+    if sys.platform in ("linux", "darwin", "emscripten", "win32"):
+        sz = 32 * 1024
+    else:
+        sz = 2 * 1024
 else:
     sz = 512 * 1024
 
--- a/docs/library/micropython.rst
+++ b/docs/library/micropython.rst
@@ -80,6 +80,9 @@ Functions
    used.  The absolute value of this is not particularly useful, rather it
    should be used to compute differences in stack usage at different points.
 
+   Note: When building with sanitizers (ASan/UBSan), an 8 KiB stack margin is
+   automatically reserved unless the port overrides MICROPY_STACK_CHECK_MARGIN.
+
 .. function:: heap_lock()
 .. function:: heap_unlock()
 .. function:: heap_locked()
--- a/ports/unix/mpconfigport.h
+++ b/ports/unix/mpconfigport.h
@@ -137,6 +137,11 @@ typedef long mp_off_t;
 #define MICROPY_STACKLESS_STRICT    (0)
 #endif
 
+// Reserve extra C-stack headroom for overflow checks.
+// Sanitizer builds enlarge call frames; 8 KiB prevents
+// false positives without noticeably shrinking usable stack.
+#define MICROPY_STACK_CHECK_MARGIN  (8192)
+
 // Implementation of the machine module.
 #define MICROPY_PY_MACHINE_INCLUDEFILE "ports/unix/modmachine.c"
 


dpgeorge:
Thanks for the patch.

Can you describe a test which shows that this is necessary?  The CI has been passing, and it does recursive depth tests, so I wonder why it's needed.

AJMansfield:
I'm not very confident I understand what's going on here, as I've not been able to figure out how to successfully link a clang build of micropython with the asan and ubsan sanitizers in question on linux, but after digging in on what the stack check margin actually _does_...

> the compiler enlarges each C call frame; MicroPython’s stack-overflow guard can therefore trigger before the real end of the stack.

This is actually the _opposite_ of what I believe is happening. Setting `MICROPY_STACK_CHECK_MARGIN` to a larger value makes micropython's stack-checking _more_ paranoid, i.e. larger values make it trigger earlier, while the stack is still smaller.

```c
inline static void mp_cstack_init_with_top(void *top, size_t stack_size) {
    MP_STATE_THREAD(stack_top) = (char *)top;

    #if MICROPY_STACK_CHECK
    assert(stack_size > MICROPY_STACK_CHECK_MARGIN); // Should be enforced by port
    MP_STATE_THREAD(stack_limit) = stack_size - MICROPY_STACK_CHECK_MARGIN;
    #else
    (void)stack_size;
    #endif
}
```

```c
void mp_cstack_check(void) {
    if (mp_cstack_usage() >= MP_STATE_THREAD(stack_limit)) {
        mp_raise_recursion_depth();
    }
}
```

```c
MP_NORETURN void mp_raise_recursion_depth(void) {
    mp_raise_type_arg(&mp_type_RuntimeError, MP_OBJ_NEW_QSTR(MP_QSTR_maximum_space_recursion_space_depth_space_exceeded));
}
```

Which then leaves margin for C functions called between the stack checks to use up to 8192 bytes more than the specified stack limit, before those operations actually start to overrun --- i.e. converting a situation that might result in a hard stack overflow, or memory corruption on platforms without stack guard pages and the like.

But that still leaves an unanswered question: 
What is is micropython actually _doing_ that would convince a sanitizer that some of its stack accesses are out-of-bounds?

tests/thread/thread_stacksize1.py
AJMansfield, line 1: This test change makes sense though --- if the sanitizer is padding out every stack frame with guard pages, then in a thread context where the standard library can't just freely syscall `brk` to expand, it makes sense that the test might need to explicitly request larger stacks that can hold all the extra padding.

py/mpconfig.h
```diff
@@ -701,6 +701,16 @@
...
 // overflows from function calls made between checks.
+#if !defined(MICROPY_STACK_CHECK_MARGIN)
+#if defined(__SANITIZE_ADDRESS__) || defined(__SANITIZE_UNDEFINED__)
+#define MICROPY_STACK_CHECK_MARGIN (8192)
+#elif defined(__has_feature)
+#if __has_feature(address_sanitizer) || __has_feature(undefined_sanitizer)
+#define MICROPY_STACK_CHECK_MARGIN (8192)
+#endif
+#endif
+#endif
```
projectgus, line 712: While we only support sanitizers on the unix port does it makes sense to move this block into `ports/unix/mpconfigport.h`?

(It's also redundant in this version of the PR, where the margin is already added unconditionally on the unix port.)

projectgus (commented):
This seems like a reasonable change, thanks @koxudaxi! Although I'm also curious how to reproduce a stack-related sanitizer failure on the current master.

koxudaxi:
@dpgeorge @AJMansfield @projectgus 
Thank you for reviewing this PR.
During the last two days I fixed some bugs and added tests on the t‑strings branch. After this work my view on this patch changed.
Now t‑strings is enabled only on ports with a larger stack (macOS, WebAssembly, Windows, etc.). In these builds the tests pass without changing `MICROPY_STACK_CHECK_MARGIN`, so the PR looks unnecessary.
But when I made a test with many interpolations, I still got a stack‑overflow in a sanitizer build. On smaller targets we avoid this by disabling the feature, but it might point to a hidden issue between sanitizer builds and our stack checker.
Should I keep improving this PR based on the review, or should we close it because we don’t need it now?
Feedback from people who often run sanitizer builds would be very helpful.

projectgus:
@koxudaxi Appreciate the update, thanks. I think that until someone encounters this problem again then we should close this, but keep this PR in mind if they do run into stack issues when using sanitizers. (Others may recommend differently, though.)

PR #17655, opened 2025-07-10
tests/run-tests.py: Auto detect whether the target has threading and the GIL or not.

dpgeorge:
### Summary

This PR makes the following changes to the test runner and CI:
- detect if the target has the `_thread` module and, if so, add the threading tests to the set of tests to run
- if the target has threading, detect if the target has the GIL enabled or not, and use that information to exclude thread mutating tests if the GIL is disabled
- work around threading bug in `tests/extmod/select_poll_eintr.py` test
- make `tests/thread/stress_aes.py` test do less work on stackless threading builds, so it finishes before the test timeout limit
- increase test timeout for unix qemu test runs
- skip some threading tests on macOS and on unix qemu

Currently the information whether a port has the GIL is hard-coded: unix/PC targets and the rp2 port are considered to be GIL-less, and all other targets have the GIL enabled (if they have threading enabled).  With the change here, some code is run to detect the GIL.  That uses the fact that native code won't bounce the GIL and can hog it, and times how long code runs.

Overall the changes here mean that the threading tests are run in many more configurations:
- on the unix port with the native emitter
- on the unix port with single precision float
- on the unix port with stackless mode enabled
- on the unix port using clang as a compiler
- on unix qemu, the architectures MIPS, ARM and RISCV-64
- on macOS
 
### Testing

Tested locally on the unix port, esp32 and rp2.  Let's also see how the CI works.

This uses a tricky technique to determine GIL/non-GIL.  An alternative would be to add some indication in a function whether the GIL is enabled, eg `sys.implementation._thread_info`.  But that will increase code size.

### DIFF
--- a/docs/library/sys.rst
+++ b/docs/library/sys.rst
@@ -77,6 +77,8 @@ Constants
    * *_mpy* - supported mpy file-format version (optional attribute)
    * *_build* - string that can help identify the configuration that
      MicroPython was built with
+   * *_thread* - optional string attribute, exists if the target has threading
+     and is either "GIL" or "unsafe"
 
    This object is the recommended way to distinguish MicroPython from other
    Python implementations (note that it still may not exist in the very
@@ -94,6 +96,14 @@ Constants
      name, for example ``'standard'``.
    * On microcontroller targets, the first element is the board name and the second
      element (if present) is the board variant, for example ``'RPI_PICO2-RISCV'``
+
+   The *_thread* entry was added in version 1.26.0 and if it exists then the
+   target has the ``_thread`` module.  If the target enables the GIL (global
+   interpreter lock) then this attribute is ``"GIL"``.  Otherwise the attribute
+   is ``"unsafe"`` and the target has threading but does not enable the GIL,
+   and mutable Python objects (such as `bytearray`, `list` and `dict`) that are
+   shared amongst threads must be protected explicitly by locks such as
+   ``_thread.allocate_lock``.
 
    .. admonition:: Difference to CPython
       :class: attention
--- a/tests/thread/stress_aes.py
+++ b/tests/thread/stress_aes.py
@@ -277,7 +277,7 @@ if __name__ == "__main__":
         n_thread = 2
         n_loop = 2
     else:
-        n_thread = 20
+        n_thread = 10
         n_loop = 5
     for i in range(n_thread):
         _thread.start_new_thread(thread_entry, (n_loop,))
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -259,7 +259,6 @@ test: $(BUILD)/$(PROG) $(TOP)/tests/run-tests.py
 test_full_no_native: $(BUILD)/$(PROG) $(TOP)/tests/run-tests.py
 	$(eval DIRNAME=ports/$(notdir $(CURDIR)))
 	cd $(TOP)/tests && MICROPY_MICROPYTHON=../$(DIRNAME)/$(BUILD)/$(PROG) ./run-tests.py
-	cd $(TOP)/tests && MICROPY_MICROPYTHON=../$(DIRNAME)/$(BUILD)/$(PROG) ./run-tests.py -d thread
 	cd $(TOP)/tests && MICROPY_MICROPYTHON=../$(DIRNAME)/$(BUILD)/$(PROG) ./run-tests.py --via-mpy $(RUN_TESTS_MPY_CROSS_FLAGS) -d basics float micropython
 	cat $(TOP)/tests/basics/0prelim.py | ./$(BUILD)/$(PROG) | grep -q 'abc'
 
--- a/py/modsys.c
+++ b/py/modsys.c
@@ -103,6 +103,18 @@ static const MP_DEFINE_STR_OBJ(mp_sys_implementation__build_obj, MICROPY_BOARD_B
 #define SYS_IMPLEMENTATION_ELEMS__BUILD
 #endif
 
+#if MICROPY_PY_THREAD
+#if MICROPY_PY_THREAD_GIL
+#define SYS_IMPLEMENTATION_ELEMS__THREAD \
+    , MP_ROM_QSTR(MP_QSTR_GIL)
+#else
+#define SYS_IMPLEMENTATION_ELEMS__THREAD \
+    , MP_ROM_QSTR(MP_QSTR_unsafe)
+#endif
+#else
+#define SYS_IMPLEMENTATION_ELEMS__THREAD
+#endif
+
 #if MICROPY_PREVIEW_VERSION_2
 #define SYS_IMPLEMENTATION_ELEMS__V2 \
     , MP_ROM_TRUE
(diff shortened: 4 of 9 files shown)


tests/thread/stress_aes.py
```diff
@@ -263,11 +263,30 @@ def thread_entry(n_loop):
...
+
+        def stackless():
+            pass
+
+        micropython.heap_lock()
+        stackless()
+        micropython.heap_unlock()
+    except RuntimeError:
+        micropython.heap_unlock()
+        is_stackless = True
```
dlech:
```suggestion
    def stackless():
        pass

    micropython.heap_lock()
    try:
        stackless()
    except RuntimeError:
        is_stackless = True
    finally:
        micropython.heap_unlock()
```

does it work to make the try/except a bit more focused on where the actual exception is expected?

ports/unix/mpthreadport.c
```diff
@@ -56,7 +56,7 @@
 #endif
 
 // This value seems to be about right for both 32-bit and 64-bit builds.
-#define THREAD_STACK_OVERFLOW_MARGIN (8192)
+#define THREAD_STACK_OVERFLOW_MARGIN (4 * 8192)
```
dlech:
```suggestion
#define THREAD_STACK_OVERFLOW_MARGIN (16 * 1024)
```

Also, the commit message could use an explanation of why this is changed.

tests/feature_check/thread_gil.py
```diff
@@ -0,0 +1,41 @@
...
+        pass
+
+
+def thread():
+    global thread_done
+    busy()
+    thread_done = True
+
+
+global thread_done
```
dlech: Do we actually need `global` statement at the top level?

tests/feature_check/thread_gil.py
```diff
@@ -0,0 +1,41 @@
...
+thread_done = False
+t0 = time.ticks_ms()
+_thread.start_new_thread(thread, ())
+busy()
+while not thread_done:
+    time.sleep(0)
+dt = time.ticks_diff(time.ticks_ms(), t0)
```
dlech: Depending on time measurements seems like it could be flaky, especially on desktop systems. I'm not awake enough yet though to try to come up with an alternative that doesn't involve creating a deadlock.

dpgeorge:
Thanks @dlech for the review here.  After I initially submitted this PR and all the CI failed on things I couldn't test locally, the scope of the PR increased quite a bit... so there was a lot of back and forth getting it to work.

tests/thread/stress_aes.py
dpgeorge: You are right, it's probably better to be more focused like that.  I will change it.

(I was just copying existing code for this from eg `tests/micropython/heapalloc.py`.)

tests/feature_check/thread_gil.py
dpgeorge: You are right, now fixed.

tests/feature_check/thread_gil.py
dpgeorge: Yes, it could be flaky.  But from testing on bare-metal and PC, it looks reasonable.  The timing is usually either very close to T or 2*T (depending if GIL is disable or enabled).

If it fails due to timing errors (running for too long), it will incorrectly think the GIL is enabled when it is not.  Then it will run the mutate tests and they'll probably all fail.  In such a case we can just rerun CI...

If it gets to be a real issue, could increase the timing here, eg double it (I just don't want the auto detection to take too long, because it's run at the start of each run of `run-tests.py`.)

ports/unix/mpthreadport.c
dpgeorge: I'm currently trying to work out if this change is even needed.

dpgeorge: This change is not needed, it doesn't fix the issue I thought it would fix.  So I took it out.

dpgeorge:
After a lot of back-and-forth with the CI, this PR is now in its final state.  I updated the comment at the top with a better description of what this PR is aiming for.

tests/feature_check/thread_gil.py
projectgus: I have some similar concerns, and generally prefer the approach of explicitly adding this information in `sys.implementation` or similar. Suggest we keep this approach now, but keep the explicit option in mind if it turns out to be flaky.

projectgus (approved):

tests/feature_check/thread_gil.py
dpgeorge: Having some information in `sys.implementation` could actually be useful for users, instead of them having to run a tricky detection test like this (if they ever need to know the threading implementation...).

And I guess if you have threading enabled you have some amount of flash to spare (although not necessarily on cc3200 which does enable threading...).

So maybe it's worth adding `sys.implementation._thread` after all?

projectgus: I think it'd be useful to add. Especially if we don't add any `_thread` property on builds without threading support.

dpgeorge: I've now changed this GIL detection to use `sys.implementation._thread`.  It's a lot simpler.

dpgeorge: And retested on PICO_W, ESP32_GENERIC and unix port.  They detect the GIL correctly and the relevant tests run.

docs/library/sys.rst
```diff
@@ -90,6 +92,11 @@ Constants
    access this field using ``sys.implementation._build.split("-")``.  The
    elements that are currently used are:
 
+   The *_thread* entry was added in version 1.26.0 and if it exists then the
+   target has the ``_thread`` module.  The target enables the GIL (global
+   interpreter lock) if this attribute equals ``"GIL"``, otherwise the attribute
+   is ``"no-GIL"`` and the target does not enable the GIL.
```
projectgus: Do you think MicroPython will ever have a variant that is non-GIL but thread safe (ala CPython `--disable-gil`?)

If yes then maybe we could make the attribute values `GIL` and `unsafe` (or similar), keeping the door open for a `safe` value in some future release.

tests/feature_check/thread_gil.py
projectgus: Much neater, I agree worth the extra size & complexity!

docs/library/sys.rst
projectgus: (Would mean tests which test for no-GIL will need to test `_thread != "GIL"`, I guess

dpgeorge: > Do you think MicroPython will ever have a variant that is non-GIL but thread safe

Yes, that's a possibility.  But, I would suspect that we gradually evolve the current no-GIL mode to a more safer behaviour, rather than have separate options for GIL/unsafe-no-GIL/safe-no-GIL.

That said, I'm happy to change "no-GIL" to "unsafe".  Although that could scare users, but OTOH it might push them to read the docs to understand what "unsafe" actually means.

Is there another word that could better describe the situation, eg "racey" or "raceful"?

projectgus: I think it's OK if this scares users! It's quite difficult to write threaded Python correctly with a thread unsafe runtime.

> But, I would suspect that we gradually evolve the current no-GIL mode to a more safer behaviour, rather than have separate options for GIL/unsafe-no-GIL/safe-no-GIL.

Agree this would be the best way to evolve it. From perspective of someone targeting MicroPython it may still be useful to programmatically tell if the current runtime is thread safe or not (i.e. even if no MicroPython release supports all three values, there might still be code that checks for all three.)

dpgeorge: OK, I've now renamed "no-GIL" to "unsafe", and updated the docs.

Also tested again on esp32, rp2 and unix.

PR #17660, opened 2025-07-11
tests/extmod: Close UDP timely.

yf13:
### Summary

This adds call to release UDP port timely so that to reuse it  in other tests. Otherwise, one could face issue like https://github.com/orgs/micropython/discussions/17623.

### Testing

Tested with VARIANT coverage of UNIX port on Ubuntu 22.04.

### DIFF
--- a/tests/extmod/select_poll_udp.py
+++ b/tests/extmod/select_poll_udp.py
@@ -29,3 +29,5 @@ print(poll.poll(0)[0][1] == select.POLLOUT)
 if hasattr(select, "select"):
     r, w, e = select.select([s], [], [], 0)
     assert not r and not w and not e
+
+s.close()
--- a/tests/extmod/socket_udp_nonblock.py
+++ b/tests/extmod/socket_udp_nonblock.py
@@ -19,3 +19,5 @@ try:
     s.recv(1)
 except OSError as er:
     print("EAGAIN:", er.errno == errno.EAGAIN)
+
+s.close()


dpgeorge:
Thanks for the contribution.  This is a good fix.

But I think the socket also needs to be closed in the initial detection part of the script at the top.

This bit of code:
```py
try:
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    s.bind(socket.getaddrinfo("127.0.0.1", 8000)[0][-1])
except OSError:
    print("SKIP")
    raise SystemExit
```
should probably be changed to:
```py
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
try:
    s.bind(socket.getaddrinfo("127.0.0.1", 8000)[0][-1])
except OSError:
    s.close()
    print("SKIP")
    raise SystemExit
```

yf13:
@dpgeorge thanks for the comment.  But I  have some concerns regarding this close-before-SKIP:
- This `close()` before SKIP doesn't help to free the UDP port earlier, right?  
- The unbound socket will be closed (e.g. by GC) later, so not closing it here may help to reveal more issues?
- The scope of changes will grow as same pattern appear in more places (e.g. `select_ipoll.py` etc).

So do we really want that?

dpgeorge:
> * This `close()` before SKIP doesn't help to free the UDP port earlier, right?

Correct.  The `s.close()` still needs to be added at the end of the test, as done in this PR.

> * The unbound socket will be closed (e.g. by GC) later, so not closing it here may help to reveal more issues?

Correct, it will eventually be closed by the GC (most likely on soft reset).

Usually we like to make the tests clean up properly after themselves.  You are right that it may help reveal more issues, although there should be other tests for that.

> * The scope of changes will grow as same pattern appear in more places

Yes there are a few other tests where this pattern appears, and would need to be fixed for consistency.

If you want to keep this PR small and of limited scope, then I'm happy for it to stay as-is, just closing the socket at the end of the test.  At least that helps with the situation you identified, running the tests in parallel.

yf13:
@dpgeorge, thank you for the confirmations. as I am too new to the code base, it is better to stick with small changes fow now. We can do more cleanups later in separate PR, how do you think?

dpgeorge (approved):

dpgeorge:
> We can do more cleanups later in separate PR, how do you think?

Yes, I'm always happy to have smaller changes.

PR #17668, opened 2025-07-13
Allow the GIL to be enabled on the unix port, and add a CI job to test this configuration

dpgeorge:
### Summary

Up until now, thread tests have only been run (under CI at least) on a small number of build configurations.  PR #17655 enables the thread tests on a lot more of the existing builds, and this PR adds a new job which builds and tests the unix port with the GIL enabled.

(This PR builds on #17655 and includes all the commits from that PR.)

Running tests with the GIL enabled will help to test the esp32 port's threading implementation, which also uses the GIL.

Edit: updated some thread tests to add `time.sleep(0)` to the busy-wait loop, so that they pass when used with the native emitter.

### Testing

Will be tested by CI.

Edit: new CI job runs correctly, and takes 1 minute 55 seconds.  It runs the thread tests in bytecode and native emitter mode.

Adds a new CI job which adds to the CI time, but it should only be a couple of minutes, and run in parallel.


tests/thread/mutate_bytearray.py
```diff
@@ -36,7 +37,7 @@ def th(n, lo, hi):
 
 # busy wait for threads to finish
```
dlech, line 38: Missed one. 

But, TBH, I would still call all of them busy waits since the arg to `sleep()` is 0.

tests/feature_check/target_info.py
```diff
@@ -20,4 +20,12 @@
     "xtensawin",
     "rv32imc",
 ][sys_mpy >> 10]
-print(platform, arch)
+
+try:
+    import _thread
+
+    have_thread = "thread"
```
dlech: I was going to suggest to just make this "yes" here, but I see run-tests.py changes in a later commit, so it doesn't really matter.

tests/thread/mutate_bytearray.py
dpgeorge: OK, I've reverted the change to the comment and just left them all as "busy wait for threads to finish".

tests/feature_check/target_info.py
dpgeorge: Yes, it's changed later to be GIL or no-GIL.

projectgus (approved):
Changes look good to me. Suggest as part of this PR we also add some brief docs about the native emitter thread/scheduler starvation issue, and the `sleep(0)` workaround. I think the appropriate place is the dot point list in `reference/speed_python.rst`?

(While looking for that I also noticed that `refernece/micropython.rst` doesn't document the `native` annotation at all - it should probably have a short entry that links to the explanatory section in `speed_python.rst`? I can submit this in a follow-up, though.)

dpgeorge:
> Suggest as part of this PR we also add some brief docs about the native emitter thread/scheduler starvation issue, and the `sleep(0)` workaround.

I've now added this in the last commit.


> While looking for that I also noticed that `refernece/micropython.rst` doesn't document the `native` annotation at all - it should probably have a short entry that links to the explanatory section in `speed_python.rst`?

Yes, would be good to document bytecode/native/viper decorators in `docs/library/micropython.rst`.  Feel free to do that (but not urgent).

PR #17670, opened 2025-07-13
py: Make math and cmath modules extensible.

dlech:
### Summary

Change math and cmath module registration from normal to extensible. This way users can extend the modules to add any missing functions.

At Pybricks, we actually want this to be able to import `umath` for backwards compatibility purposes, but this seems like something that would be useful for other users as well since they will likely expect these modules to be extensible like most other standard library modules are in MicroPython.

### Testing

We tested this in Pybricks where we were able to import `umath` to get the `math` module.

We'll see if there is any code size change. Alternative would be for Pybricks to keep carrying this patch in their fork.

### DIFF
--- a/py/modcmath.c
+++ b/py/modcmath.c
@@ -149,6 +149,6 @@ const mp_obj_module_t mp_module_cmath = {
     .globals = (mp_obj_dict_t *)&mp_module_cmath_globals,
 };
 
-MP_REGISTER_MODULE(MP_QSTR_cmath, mp_module_cmath);
+MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR_cmath, mp_module_cmath);
 
 #endif // MICROPY_PY_BUILTINS_FLOAT && MICROPY_PY_BUILTINS_COMPLEX && MICROPY_PY_CMATH
--- a/py/modmath.c
+++ b/py/modmath.c
@@ -450,6 +450,6 @@ const mp_obj_module_t mp_module_math = {
     .globals = (mp_obj_dict_t *)&mp_module_math_globals,
 };
 
-MP_REGISTER_MODULE(MP_QSTR_math, mp_module_math);
+MP_REGISTER_EXTENSIBLE_MODULE(MP_QSTR_math, mp_module_math);
 
 #endif // MICROPY_PY_BUILTINS_FLOAT && MICROPY_PY_MATH


dlech:
unix coverage test failure is just a flaky test: `1 tests failed: thread/thread_gc1.py`

dpgeorge:
The downside here is that importing an extensible module is (a lot) slower than a non-extensible built in module.  It's slower because it has to search the filesystem (everything in `sys.path`) first, and then fall back to searching built-in modules.

> At Pybricks, we actually want this to be able to import `umath` for backwards compatibility purposes

Is that the only reason you need this change?  Maybe there's another way to support this without slowing down the import.

Note that if this PR were accepted then the docs need updating to add these modules to the list of extensible ones; see the "Extending built-in libraries from Python" section in `docs/library/index.rst`.

dlech:
> It's slower because it has to search the filesystem (everything in `sys.path`) first, and then fall back to searching built-in modules.

Ah, we didn't notice because we don't have a filesystem.

> Is that the only reason you need this change? Maybe there's another way to support this without slowing down the import.

Sure, open to suggestions.

dpgeorge:
Maybe something like this (untested!):
```diff
--- a/py/objmodule.c
+++ b/py/objmodule.c
@@ -207,6 +207,12 @@ mp_obj_t mp_module_get_builtin(qstr module_name, bool extensible) {
         if (!elem) {
             return MP_OBJ_NULL;
         }
+        #if MICROPY_LEGACY_U_MODULE
+        elem = mp_map_lookup((mp_map_t *)&mp_builtin_module_map, MP_OBJ_NEW_QSTR(qstr_from_strn(module_name_str + 1, module_name_len - 1)), MP_MAP_LOOKUP);
+        if (!elem) {
+            return MP_OBJ_NULL;
+        }
+        #endif
     }
 
     #if MICROPY_MODULE_BUILTIN_INIT
```

dpgeorge:
Or can you simply do:
```c
MP_REGISTER_MODULE(MP_QSTR_umath, mp_module_math);
MP_REGISTER_MODULE(MP_QSTR_ucmath, mp_module_cmath);
```
in your `main.c`?

dlech:
> ```c
> MP_REGISTER_MODULE(MP_QSTR_umath, mp_module_math);
> ```

It works, but takes up precious flash memory and makes `help("modules")` list both `math` and `umath`.



> ```diff
> #if MICROPY_LEGACY_U_MODULE
> ```

We are overriding `mp_builtin___import___default()`, so could do something like that there. 



> The downside here is that importing an extensible module

If this is a concern, it seems like it would be beneficial to add config options to optionally make any module extensible or not. Then minimal ports could completely turn off the extensible module mechanism to save on size a bit.

PR #17678, opened 2025-07-14
examples/rp2/pio_uart_rx: Fix error running example.

AJMansfield:
Running the unmodified `pio_uart_rx` example by uploading the file and importing it doesn't succeed, and instead emits a NameError at line 26.

Discovered while using this and other examples to help verify #16851.

With this fix, and the specified jumper between GPIO4 and GPIO3, this example runs successfully on both my PicoW and my Pico2.

### DIFF
--- a/examples/rp2/pio_uart_rx.py
+++ b/examples/rp2/pio_uart_rx.py
@@ -23,7 +23,7 @@ PIO_RX_PIN = Pin(3, Pin.IN, Pin.PULL_UP)
 @asm_pio(
     autopush=True,
     push_thresh=8,
-    in_shiftdir=rp2.PIO.SHIFT_RIGHT,
+    in_shiftdir=PIO.SHIFT_RIGHT,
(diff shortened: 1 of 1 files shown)


dpgeorge (approved):
Thanks for the fix.

(The existing example did work because `_boot.py` imports `rp2` into the global namespace and doesn't delete it, but we shouldn't rely on that.)

AJMansfield:
> The existing example did work because `_boot.py` imports `rp2` into the global namespace and doesn't delete it, but we shouldn't rely on that.

Ah, right! In my case I was running the examples like:

```bash
mpremote \
    cp examples/rp2/pio_uart_rx.py :/pio_uart_rx.py \
    exec "import pio_uart_rx"
```

But running as `mpremote run examples/rp2/pio_uart_rx.py` should work since that's basically equivalent to pasting the contents of the file into the REPL (i.e. inheriting the main namespace which inherits the boot namespace).

PR #17653, opened 2025-07-10
Zephyr/zsensor: Add sensor attribute set/get methods.

bikeNomad:
### Summary

This PR adds `zsensor.Sensor.attr_set()` and `zsensor.Sensor.attr_get()` methods that are
necessary to set various sensor attributes if they haven't been set statically in the device tree.

I needed this because the LSM6DS3TR-C sensor in my XIAO BLE NRF52840 SENSE board wouldn't
work with `zsensor` because it didn't have any default configuration for sampling frequency.

I added the various `SENSOR_ATTR_`* constants from `zephyr/incude/zephyr/drivers/sensor.h` as
`ATTR_`* constants in `zsensor`.

I also added the below test script to the `README.md`.

### Testing

Here is a test script that I ran on my board:
```python
    from zsensor import *
    sensor = Sensor('lsm6ds3tr_c')  # name from Devicetree
    # Set full-scale to 2g (19.613300 m/sec^2)
    # units are micro-m/s^2
    sensor.attr_set(ACCEL_XYZ, ATTR_FULL_SCALE, 19, 613300)
    # Set sampling frequency to 104 Hz
    sensor.attr_set(ACCEL_XYZ, ATTR_SAMPLING_FREQUENCY, 104)
    sensor.measure()
    sensor.get_float(ACCEL_X) # -0.508 (m/s^2)
    sensor.get_float(ACCEL_Y) # -3.62 (m/s^2)
    sensor.get_float(ACCEL_Z) # 9.504889 (m/s^2)
```

### Trade-offs and Alternatives

This adds a small amount of code size to the `zsensor` module; the whole module may be eliminated
by not #defining `MICROPY_PY_ZSENSOR`.

Issue #17615, opened 2025-07-04
zephyr: Add ADC support


bikeNomad:
### Description

The Zephyr port is currently lacking `machine.ADC`. The Zephyr ADC API is relatively simple, though it requires the ADC channels to be set up in the devicetree (typically using an overlay). See this example:
https://github.com/zephyrproject-rtos/zephyr/tree/main/samples/drivers/adc/adc_dt
We could use the same tuple method as the `Pin` constructor to access channels: `ADC(('adc0', 1), ...)`.
There is also an API for channel sequencing; perhaps this API could be made available through (new) `ADCBlock` methods.

### Code Size

This should be optional code, enabled by `MICROPY_PY_MACHINE_ADC`.
The `ADCBlock` sequencing/control API can be enabled by `MICROPY_PY_MACHINE_ADC_BLOCK`.

### Implementation

I intend to implement this feature and would submit a Pull Request if desirable

### Code of Conduct

Yes, I agree

bikeNomad:
#17616 adds code for the ADC.

PR #17679, opened 2025-07-14
zephyr: Add Seeed Studio's XIAO BLE NRF52840 SENSE board.

bikeNomad:
### Summary

This PR adds Zephyr support for Seeed Studio's XIAO BLE NRF52840 SENSE board.

It also provides a good example of a richer Zephyr port than the default, adding:
  - Frozen modules (including `asyncio`, `upysh`, `aioble`, `neopixel`, and `aiorepl`).
  - Enough MicroPython features to support using `aioble` (at least for the `temp_sensor.py` example)
  -  JSON, `random`, `re`, `struct`, etc.
  - Support for my Zephyr ADC code (see PR #17615)
  - Mounted LFS2 filesystem using the 2MiB flash mounted as `/flash`
 
You can build it in the usual way:
```
$ cd ports/zephyr
$ west build -b xiao_ble/nrf52840/sense . 
```
This will produce a `.uf2` file in `build/zephyr/zephyr.uf2` that can be copied into the XIAO-SENSE USB drive that appears when you double-click the `RESET` button.

### Testing

I have tested this on several of these boards, with and without my ADC (#17615) and `zensor.Sensor.attr_set()` (#17653) PRs.

I have run the `aioble` `temp_sensor.py` example but haven't done much more with the BLE.

The build size is as follows:
```
Memory region         Used Size  Region Size  %age Used
           FLASH:      428864 B       788 KB     53.15%
             RAM:      165908 B       256 KB     63.29%
        IDT_LIST:          0 GB        32 KB      0.00%
```


dpgeorge:
Thanks for the contribution, this is a nice addition.

I've opened #17680 which should help simplify this PR.

ned-pcs:
@dpgeorge OK, simplified using your PR #17680. Will re-base when you merge it into mainline.

ned-pcs:
The additional features defined in `mpconfigport-xiao.h` add 12224 bytes of flash and 12 bytes of RAM size.

ports/zephyr/boards/manifest_xiao_ble_nrf52840_sense.py
```diff
@@ -0,0 +1,11 @@
+# This is an example frozen module manifest. Enable this by configuring
+# the Zephyr project and enabling the frozen modules config feature.
```
dpgeorge: I think this comment can be removed, it's not really an example anymore for this board.

ports/zephyr/boards/mpconfigport-xiao.h
```diff
@@ -0,0 +1,51 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "mpconfigport.h"
+
+#define MICROPY_PY_ASYNCIO (1)
```
dpgeorge: Instead of defining all these options, it may be simpler to do:
```c
#define MICROPY_CONFIG_ROM_LEVEL (MICROPY_CONFIG_ROM_LEVEL_EXTRA_FEATURES)
```
Then you automatically get the new features at that level as they are added in the future, and the config matches other boards at this level.

ports/zephyr/boards/xiao_ble_nrf52840_sense.conf
```diff
@@ -0,0 +1,66 @@
...
+
+CONFIG_DISK_ACCESS=y
+
+# Enable MicroPython's VFS and mount a littlefs filesystem on boot
+CONFIG_MICROPY_VFS_LFS2=y
+CONFIG_MICROPY_VFS_FAT=y
+
+CONFIG_MICROPY_FROZEN_MODULES=y
+CONFIG_MICROPY_FROZEN_MANIFEST="boards/manifest_xiao_ble_nrf52840_sense.py"
+CONFIG_MICROPY_CONFIGFILE="boards/mpconfigport-xiao.h"
```
dpgeorge: The `boards` directory is going to get rather messy with all these files.  I suggest making a subdir for these files:
```
boards/xiao_ble_nrf52840_sense/manifest.py
boards/xiao_ble_nrf52840_sense/mpconfigport.h
```
Note that the conf and overlay files need to remain where they are so west can find them.

ports/zephyr/boards/mpconfigport-xiao.h
ned-pcs: Good idea, though it adds about 26kB to the flash size.

Blafy:
Great !
Do you have a plan to implement a disable flag for the p25q16h ?
This little flash is IMO a bad addition on this board, as it requires to be put into suspend mode to keep the nice low power of the NRF chip ...

bikeNomad:
@blafy asks:
>  Great ! Do you have a plan to implement a disable flag for the p25q16h ? This little flash is IMO a bad addition on this board, as it requires to be put into suspend mode to keep the nice low power of the NRF chip ...

Good question. Not using this flash means that instead of 2MiB filesystem we'd have only 32KiB, at least without partition modification.

The stock "Storage" partition (given the original overlay; see `nrf52840_partition_uf2_sdv7.dtsi` ) on the NRF chip is only 32kB in size (156kB is given to the SoftDevice that we don't use, and 48kB is saved for the UF2 bootloader). The way these boards are set up at the factory includes the UF2 bootloader and SoftDevice. Using MCUBoot and two app partitions (see `nrf52840_partition.dtsi`) still only leaves 32kB of storage space.

Although the newest Adafruit UF2 bootloader allows for having no SoftDevice (thus maybe letting us use the 156kiB space for `/flash` instead), I haven't tried this.

As for putting it into suspend mode, I would hope that the `nordic,qspi-nor` driver would do this when possible. I've seen current consumption in the 1-2mA range even without the new `lightsleep` code.

ports/zephyr/boards/mpconfigport-xiao.h
dpgeorge: > it adds about 26kB to the flash size.

I think that's fine.  This part has enough flash, and it keeps the configuration simple.

dpgeorge:
This can now be rebased, because the `mpconfigport.h` changes were merged.

Also, this PR is currently blocked behind #17616 (ADC support), which may take some time to resolve.  If you want to get this PR in now (before ADC support) then please remove all the ADC and io-channel stuff from this PR.  Otherwise this PR can wait, no hurry.

ned-pcs:
> This can now be rebased, because the `mpconfigport.h` changes were merged.
> 
> Also, this PR is currently blocked behind #17616 (ADC support), which may take some time to resolve. If you want to get this PR in now (before ADC support) then please remove all the ADC and io-channel stuff from this PR. Otherwise this PR can wait, no hurry.

OK, rebased and removed ADC stuff from `mpconfigport.h` and the overlay.

ports/zephyr/boards/xiao_ble_nrf52840_sense.conf
```diff
@@ -0,0 +1,67 @@
...
+CONFIG_FLASH_MAP=y
+CONFIG_FLASH_PAGE_LAYOUT=y
+CONFIG_FILE_SYSTEM=y
+CONFIG_FILE_SYSTEM_LITTLEFS=y
+
+CONFIG_DISK_ACCESS=n
+
+# Enable MicroPython's VFS and mount a littlefs filesystem on boot
+CONFIG_MICROPY_VFS_LFS2=y
+CONFIG_MICROPY_VFS_FAT=y
```
dpgeorge: I think these will be enabled by default from `prj.conf`, right?

If so, I suggest removing everything from here that's already the default in `prj.conf`, to keep this file minimal.  Otherwise in the future when people add other boards, they copy this file and there are more and more duplicated settings.

ports/zephyr/boards/xiao_ble_nrf52840_sense/mpconfigport.h
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "../mpconfigport.h"
+
+#undef MICROPY_CONFIG_ROM_LEVEL
+#define MICROPY_CONFIG_ROM_LEVEL (MICROPY_CONFIG_ROM_LEVEL_EXTRA_FEATURES)
+
+#define MICROPY_PY_SYS_MODULES (1)
```
dpgeorge: This is already the default (in `py/mpconfig.h`), so can be removed.

ports/zephyr/boards/xiao_ble_nrf52840_sense/mpconfigport.h
```diff
@@ -0,0 +1,32 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "../mpconfigport.h"
+
+#undef MICROPY_CONFIG_ROM_LEVEL
+#define MICROPY_CONFIG_ROM_LEVEL (MICROPY_CONFIG_ROM_LEVEL_EXTRA_FEATURES)
```
dpgeorge, line 25: If you set this before including `../mpconfigport.h` then it doesn't need to be undef'd.

ports/zephyr/boards/xiao_ble_nrf52840_sense/manifest.py
```diff
@@ -0,0 +1,8 @@
+# freeze("../modules")
```
dpgeorge: This comment can be removed.

ports/zephyr/boards/xiao_ble_nrf52840_sense/manifest.py
```diff
@@ -0,0 +1,8 @@
...
+include("$(MPY_DIR)/extmod/asyncio")
+
+# Require a micropython-lib module.
```
dpgeorge: This comment is not really correct, it was copied from the example manifest.

You could remove it, or change it to something like "Require some useful micropython-lib modules."

jonnor:
I have tested this working on device. So that is a +1 for me. 

In general, I think this is a good target hardware for a standard build on a Zephyr board. For the following reasons:

- It is widely available (incl via distributors like Digikey etc) and affordable, with schematics etc available
- Easy to flash without special tools, since it implements UF2 over USB
- It covers NRF52, which is now in some areas better supported with Zephyr port than "nrf" port - notably Bluetooth Low Energy
- Has I2C devices on-board, enables testing that key module

jonnor:
@bikeNomad there are some checks failing for the commit formatting and code spelling. Can you check them out and resolve them?

One of the failing check is for the name "Ned", which seems not-very-useful. But I have no idea how to make codespell ignore that, unfortunately. Or if that is critical wrt merging.

bikeNomad:
> @bikeNomad there are some checks failing for the commit formatting and code spelling. Can you check them out and resolve them?
> 
> One of the failing check is for the name "Ned", which seems not-very-useful. But I have no idea how to make codespell ignore that, unfortunately. Or if that is critical wrt merging.

I'm working on getting the checks to pass. I found that I had to use all upper-case letters for `NED` because `codespell` is configured to ignore three-letter all-caps words.

bikeNomad:
I re-based this on the ADC PR (#17616) to simplify.

The other thing that I added was automatic filesystem creation in case of an unformatted flash storage area (which will be the case when applying to a new board). I did this in the same way that other ports do it: by adding a `_boot.py` module that first tries to mount an existing formatted filesystem, then formats and mounts if that fails.

Users of other zephyr ports have noticed the lack of automatic filesystem creation, as well.

I could separate this into a separate PR if desired.

VynDragon:
I dont think the bits on the FS should be accepted, the current code is fine for its purpose and this is just translating it into python for the most part, and disabling/breaking the feature for every other supported boards because you are not adding a generic manifest.py, which might break other boards in other ways anyway.

```
CONFIG_FILE_SYSTEM=y
CONFIG_FILE_SYSTEM_LITTLEFS=y
```
are unecessary, you are only using flash map from zephyr's side. 

```
#if MICROPY_MODULE_FROZEN && MICROPY_VFS
    // Mount and/or create the filesystem
    pyexec_frozen_module("_boot.py", false);
    #endif
 ```
should be a separate PR that adds the ability to run _boot.py at start in general.

**Comment by bikeNomad** [comment]
> I dont think the bits on the FS should be accepted, the current code is fine for its purpose and this is just translating it into python for the most part, and disabling/breaking the feature for every other supported boards because you are not adding a generic manifest.py, which might break other boards in other ways anyway.

@VynDragon thanks for looking at the code. I'll remove the unnecessary lines in the config.

There is a generic `manifest.py` (which you wrote :grin:) that freezes the `modules` directory (which in this PR contains only `_boot.py`). The current code doesn't bother to create a filesystem in the case where there isn't one already (it just mounts it if it's already formatted).

This lack of filesystem formatting has bitten myself and others (for instance: #16240). It will cause frustration for future users installing to boards for the first time.

I will, however, separate the FS code from the PR.

**Comment by VynDragon** [comment]
> There is a generic manifest.py (which you wrote 😁) that freezes the modules directory (which in this PR contains only _boot.py)

Ah, sorry then...

**Comment by bikeNomad** [comment]
OK, I separated the filesystem mounting/creation changes into another PR (#18185) per @VynDragon 's suggestion.

**Comment by dpgeorge** [comment]
Please rebase on latest master now that the ADC code is merged.

File: ports/zephyr/boards/xiao_ble_nrf52840_sense.conf [hunk 9]
```diff
@@ -0,0 +1,56 @@
...
+CONFIG_FLASH_PAGE_LAYOUT=y
+
+CONFIG_DISK_ACCESS=n
+
+CONFIG_MICROPY_FROZEN_MODULES=y
+CONFIG_MICROPY_FROZEN_MANIFEST="boards/xiao_ble_nrf52840_sense/manifest.py"
+CONFIG_MICROPY_CONFIGFILE="boards/xiao_ble_nrf52840_sense/mpconfigport.h"
+
+# Using PicoLibC instead of Newlib saves around 7k of flash.
+CONFIG_NEWLIB_LIBC=n
```
dpgeorge: newlib is now disabled by default, so you can remove this line.

Also, maybe just remove this section (which selects picolibc) altogether?  It's good to keep the config minimal when possible.

bikeNomad: @dpgeorge Thanks for all the review comments on my PRs!

dpgeorge (approved):

PR #17680, opened 2025-07-15
zephyr: use feature levels for minimal and standard build configurations

dpgeorge:
### Summary

There are two commits here, both simplifying the zephyr configuration to use ROM feature level settings.

The minimal configuration now uses MICROPY_CONFIG_ROM_LEVEL_MINIMUM.  This decreases firmware size by about 15k, and the tests still pass.

The standard configuration now uses MICROPY_CONFIG_ROM_LEVEL_BASIC_FEATURES.  This increases firmware size by about 6.5k, but adds some useful things:
- async/await keyword support
- `filter`, `property` and `reversed` builtins
- `range` attributes
- `str.count()` method
- `array` module with `array.array` object
- `collections` module with `collections.namedtuple` object
- `struct` module with everything
- `id = const()` and constant folding in the compiler
- `__all__` support in modules

Using feature levels simplifies the configurations and allows boards to better fine-tune their config, eg #17679 should be much simpler with this PR.

### Testing

Tested locally with qemu_cortex_m3 build, both minimal and standard.  Tested also with frdm_k64f and the standard configuration.

Will also be tested by CI.

This should be a win all round.  After all, this is what the feature levels are for (and zephyr so far didn't convert to use them, that's what's done here).

### DIFF
--- a/ports/zephyr/mpconfigport_minimal.h
+++ b/ports/zephyr/mpconfigport_minimal.h
@@ -30,41 +30,36 @@
 // Included here to get basic Zephyr environment (macros, etc.)
 #include <zephyr/kernel.h>
 
+// Use the minimum configuration level to get a small but useful system.
+#ifndef MICROPY_CONFIG_ROM_LEVEL
+#define MICROPY_CONFIG_ROM_LEVEL (MICROPY_CONFIG_ROM_LEVEL_MINIMUM)
+#endif
+
 // Usually passed from Makefile
 #ifndef MICROPY_HEAP_SIZE
 #define MICROPY_HEAP_SIZE (16 * 1024)
 #endif
 
+#define MICROPY_ENABLE_COMPILER     (1)
+#define MICROPY_ENABLE_EXTERNAL_IMPORT (1)
 #define MICROPY_STACK_CHECK         (1)
 #define MICROPY_ENABLE_GC           (1)
 #define MICROPY_HELPER_REPL         (1)
 #define MICROPY_REPL_AUTO_INDENT    (1)
 #define MICROPY_KBD_EXCEPTION       (1)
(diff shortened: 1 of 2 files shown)


ports/zephyr/mpconfigport_minimal.h
```diff
@@ -30,41 +30,36 @@
...
-#define MICROPY_PY_SYS_MODULES      (0)
 #define MICROPY_LONGINT_IMPL (MICROPY_LONGINT_IMPL_LONGLONG)
 #define MICROPY_FLOAT_IMPL (MICROPY_FLOAT_IMPL_FLOAT)
 #define MICROPY_PY_BUILTINS_COMPLEX (0)
 
-// Saving extra crumbs to make sure binary fits in 128K
-#define MICROPY_COMP_CONST_FOLDING  (0)
-#define MICROPY_COMP_CONST (0)
-#define MICROPY_COMP_DOUBLE_TUPLE_ASSIGN (0)
+// These features are enabled to get the test suite passing.
```
projectgus, line 54: Long term, I wonder if we should aim to have `MICROPY_CONFIG_ROM_LEVEL_MINIMUM` pass the test suite - presumably with a mixture of automatically skipping some tests and adding some features to the minimum level...

Anyway, that's not relevant for this PR which looks like a good improvement.

projectgus (approved):

ports/zephyr/mpconfigport_minimal.h
dpgeorge: > Long term, I wonder if we should aim to have `MICROPY_CONFIG_ROM_LEVEL_MINIMUM` pass the test suite

Yes, totally agree.  We do have unix minimal variant that runs the tests under CI, but only the `tests/basics` directory, and it explicitly skips some tests.

So, something to improve in the future.

PR #17685, opened 2025-07-15
objrange: Allow return of non-small ints.

### LINKED ISSUE #17684: range() returns wrong results across 1<<31

68k private port.  rp2 version 1.21, js version on micropython homepage.

### MicroPython version

Whatever the JS on the micropython homepage uses ?

### Reproduction

Running:

       for i in range(0x3c000000, 0x44000000, 0x04000000):
           print("A %x" % i)
       b = list(range(0x3c000000, 0x44000000, 0x04000000))
       for i in b:
           print("B %x" % i)
 
Produces:
 
       A 3c000000
       A 40000000
       B 3c000000
       B -40000000

### Expected behaviour

Same output for both sides of the example.

### Observed behaviour

See example above.

### Additional Information

### Code of Conduct

Yes, I agree

jepler:
### Summary

Closes #17684.

### Testing

I wrote new tests & ran the unix coverage tests locally.

Given that values between the "small int" range and "machine int" range never worked with range(), maybe non "small int"
range() arguments should be rejected instead. However, this would be a bit contradictory as you expect to have ranges with elements up to `sys.maxsize` at least.

### DIFF
--- a/py/objrange.c
+++ b/py/objrange.c
@@ -41,9 +41,9 @@ typedef struct _mp_obj_range_it_t {
 static mp_obj_t range_it_iternext(mp_obj_t o_in) {
     mp_obj_range_it_t *o = MP_OBJ_TO_PTR(o_in);
     if ((o->step > 0 && o->cur < o->stop) || (o->step < 0 && o->cur > o->stop)) {
-        mp_obj_t o_out = MP_OBJ_NEW_SMALL_INT(o->cur);
+        mp_int_t cur = o->cur;
         o->cur += o->step;
-        return o_out;
+        return mp_obj_new_int(cur);
     } else {
         return MP_OBJ_STOP_ITERATION;
     }
@@ -132,7 +132,7 @@ static mp_obj_t range_unary_op(mp_unary_op_t op, mp_obj_t self_in) {
         case MP_UNARY_OP_BOOL:
             return mp_obj_new_bool(len > 0);
         case MP_UNARY_OP_LEN:
-            return MP_OBJ_NEW_SMALL_INT(len);
+            return mp_obj_new_int(len);
         default:
             return MP_OBJ_NULL;      // op not supported
     }
@@ -173,7 +173,7 @@ static mp_obj_t range_subscr(mp_obj_t self_in, mp_obj_t index, mp_obj_t value) {
         }
         #endif
         size_t index_val = mp_get_index(self->base.type, len, index, false);
-        return MP_OBJ_NEW_SMALL_INT(self->start + index_val * self->step);
+        return mp_obj_new_int(self->start + index_val * self->step);
     } else {
         return MP_OBJ_NULL; // op not supported
     }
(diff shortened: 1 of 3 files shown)


bsdphk:
This PR works in the code where #17684 was found.

Thanks a lot!

bsdphk:
Do I need to do more to get this PR merged ?

jepler:
I rebased the branch to bring it up to date.

@dpgeorge is there anything else I can do to move this forward?

dpgeorge:
Somehow the code size diff is now mostly negative, which is great!

tests/basics/builtin_range.py
```diff
@@ -54,6 +54,32 @@
 print(range(1, 100, -5)[5:15:-3])
 print(range(1, 100, -5)[15:5:3])
 
+for i in range(24, 67):
```
dpgeorge: It feels like this new set of tests should be in a separate file, because:
- I don't think it'll work on ports without big int (maybe I'm wrong there??), so should be suffixed with `_intbig.py`.
- Since it doesn't really work on CPython, it should probably have a .py.exp file.

Also, maybe a short comment at the top describing what the test is testing.

jepler:
There are still some oddities:
 * a micropython range doesn't seem capable of including `-sys.maxsize-1` while a Python range can
 * When `start-stop` exceeds `sys.maxsize` operations don't behave properly. (in python 3.13, mostly the operations are correct though you can't `len()` a range with more than `sys.maxsize` elements)

tests/basics/builtin_range.py
jepler: I updated the test so that it does not depend on long int support or a specific sys.maxsize value. I also added a cpydiff for some additional oddities I discovered while testing.

jepler:
`r = range(-maxsize, 0, maxsize//2)` is another failure, because this calculation overflows: `mp_int_t len = self->stop - self->start + self->step;`

jepler:
some ranges that triggered integer overflow checks in the UBsan build:
```
r = range(-maxsize, 0)
r = range(0, maxsize)
```
like so (values don't necessarily match, but line numbers should)
```
../../py/objrange.c:115:14: runtime error: signed integer overflow: 9223372036854775807 + 1 cannot be represented in type 'long int'
../../py/objrange.c:117:13: runtime error: signed integer overflow: -9223372036854775808 - 1 cannot be represented in type 'long int'
```
it would be nice to deal with these but it's a bit scope-creepy.

dpgeorge:
> it would be nice to deal with these but it's a bit scope-creepy.

Yes, it is scope creep.

The fix done in this PR is nice and self contained, fixes obvious issues, and has a good code size change.  So let's keep it as-is.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 7be82f8b7c9ee83696cb08f8f2936f0ed93cd52c and 72867bce1d2cf39abe164bebf4d239c6974c9122

Thanks for the fix @jepler !

jepler:
Thank you as always for the helpful reviews!

bsdphk:
Thanks!

PR #17688, opened 2025-07-15
Test REPR_B & fix test failures

jepler:
### Summary

For m68k-micropython I'm interested in REPR_B. However, CI doesn't check any REPR_B builds, so of course there were test failures.

This patch series adds a 32-bit REPR_B build (i686 Linux) and fixes the test failures that arose.

### Testing

I ran the tests locally with a REPR_B build. I'll check the CI for other failures.

Just for the benefit of one test in the testsuite, I added `micropython.small_int_max`. Instead, a few tests could be changed so that they just use numbers sized to work on "any repr that gets tested".


jepler:
Native & viper tests fail if --via-mpy even after accounting for small int size. I did some digging and I'm guessing it's because mpy-cross uses a fixed representation for integers in the code emitter, e.g., 
```
        } else if (si->vtype == VTYPE_INT || si->vtype == VTYPE_UINT) {
            ASM_MOV_REG_IMM(emit->as, reg_dest, (uintptr_t)MP_OBJ_NEW_SMALL_INT(si->data.u_imm));
```
I chose to simply skip running any of the "native" tests, even though locally they succeed when not going via mpy. Because this was the easiest of the choices available.

dpgeorge:
> Just for the benefit of one test in the testsuite, I added `micropython.small_int_max`. Instead, a few tests could be changed so that they just use numbers sized to work on "any repr that gets tested".

I'd prefer not to add this constant, if we can help it.  Could the tests automatically detect the max int?  Note there is already `sys.maxsize` which may be useful.  Could also add a feature check in `tests/feature_check/` to detect the representation.

If we do need to add a constant, it might be better as `sys.implementation._small_int`.

jepler:
I reworked it to not require the added property.

jepler:
is this a known-to-be-flaky test? My changes shouldn't have touched it. `1 tests failed: extmod/select_poll_eintr.py`

tests/basics/int_big_to_small.py
```diff
@@ -5,6 +5,15 @@
     print("SKIP")
     raise SystemExit
 
+micropython.heap_lock()
+try:
+    k = (1 << 29)
+    k = k + 20
+except MemoryError:
+    print("SKIP")
```
dpgeorge, line 14: This needs a `micropython.heap_unlock()`.

dpgeorge:
> is this a known-to-be-flaky test? My changes shouldn't have touched it. `1 tests failed: extmod/select_poll_eintr.py`

Yes, it's known flaky.  PR #17655 works around the issue.

jepler:
The one remaining failure appears to be `1 tests failed: extmod/select_poll_eintr.py`

jepler:
@dpgeorge would it be easier if I split this up? The uctypes change to fit the integer width of REPR_B is the most important part for me in m68k-micropython. The float conversion bit is nice to have. The rest is "just" needed so that the tests of the added REPR_B unix build can pass.

extmod/moductypes.c
```diff
@@ -43,14 +44,32 @@
 #define LAYOUT_BIG_ENDIAN    (1)
 #define LAYOUT_NATIVE        (2)
 
+#define LEN_BITS (OFFSET_BITS + BITF_OFF_BITS)
+#if MICROPY_OBJ_REPR == MICROPY_OBJ_REPR_B
+#define VAL_TYPE_BITS 4
```
dpgeorge: Maybe add a comment why things need to change here for repr B.

extmod/moductypes.c
```diff
@@ -43,14 +44,32 @@
...
+#define BITF_LEN_BITS 5
+#define BITF_OFF_BITS 5
+#define OFFSET_BITS 16
+#if VAL_TYPE_BITS + BITF_LEN_BITS + BITF_OFF_BITS + OFFSET_BITS != 30
+#error Invalid encoding field length
+#endif
+// Here we need to set sign bit right
+#define TYPE2SMALLINT(x, nbits) ((((int)x) << (32 - nbits)) >> 2)
+#define GET_TYPE(x, nbits) (((x) >> (30 - nbits)) & ((1 << nbits) - 1))
+#define VALUE_MASK(type_nbits) (~((int)0xc0000000 >> type_nbits))
```
dpgeorge: Can any of these constants be computed using values from `py/smallint.h` (eg `MP_SMALL_INT_POSITIVE_MASK`) so that the definitions can be common for both object repr's?

extmod/moductypes.c
```diff
@@ -94,6 +109,15 @@ static MP_NORETURN void syntax_error(void) {
 }
 
 static mp_obj_t uctypes_struct_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    MP_STATIC_ASSERT(MP_SMALL_INT_FITS(TYPE2SMALLINT(BFINT32, VAL_TYPE_BITS)));
```
dpgeorge, line 104: Maybe a comment that these asserts are placed here because they don't have anywhere else to go (and this is just the first function in this file)?

Also I suggest adding a blank line after them all.

tests/basics/int_big_to_small.py
```diff
@@ -5,6 +5,16 @@
     print("SKIP")
     raise SystemExit
 
+micropython.heap_lock()
+try:
+    k = (1 << 29)
+    k = k + 20
```
dpgeorge: The compiler does a lot of constant folding, in particular the `1<<29` is done at compile time, not runtime.  To make it clear that it's the add that may fail, I suggest:
```py
k = 1 << 29
micropython.heap_lock()
try:
    k += 1
except MemoryError:
    print("SKIP")
    raise SystemExit
finally:
    micropython.heap_unlock()
```

tests/basics/int_big_to_small_int29.py
```diff
@@ -0,0 +1,32 @@
+try:
```
dpgeorge, line 1: Please add a comment that this test is specifically for repr B, where small ints are maximum (1<<29)-1 (and that it also works fine for other reprs).

tests/basics/int_big_to_small_int29.py
```diff
@@ -0,0 +1,32 @@
...
+    import micropython
+    micropython.heap_lock
+except:
+    print("SKIP")
+    raise SystemExit
+
+micropython.heap_lock()
+try:
+    k = (1 << 28)
+    k = k + 20
```
dpgeorge: Won't this test always pass?  I mean, it's not necessary at all for repr B.  Maybe remove it?

tests/perf_bench/core_import_mpy_multi.py
```diff
@@ -23,7 +23,7 @@ def f():
     x = ("const tuple", None, False, True, 1, 2, 3)
 result = 123
 """
-file_data = b'M\x06\x00\x1f\x14\x03\x0etest.py\x00\x0f\x02A\x00\x02f\x00\x0cresult\x00/-5#\x82I\x81{\x81w\x82/\x81\x05\x81\x17Iom\x82\x13\x06arg\x00\x05\x1cthis will be a string object\x00\x06\x1bthis will be a bytes object\x00\n\x07\x05\x0bconst tuple\x00\x01\x02\x03\x07\x011\x07\x012\x07\x013\x81\\\x10\n\x01\x89\x07d`T2\x00\x10\x024\x02\x16\x022\x01\x16\x03"\x80{\x16\x04Qc\x02\x81d\x00\x08\x02(DD\x11\x05\x16\x06\x10\x02\x16\x072\x00\x16\x082\x01\x16\t2\x02\x16\nQc\x03`\x1a\x08\x08\x12\x13@\xb1\xb0\x18\x13Qc@\t\x08\t\x12` Qc@\t\x08\n\x12``Qc\x82@ \x0e\x03\x80\x08+)##\x12\x0b\x12\x0c\x12\r\x12\x0e*\x04Y\x12\x0f\x12\x10\x12\x11*\x03Y#\x00\xc0#\x01\xc0#\x02\xc0Qc'
```
dpgeorge, line 26: Please update the comment above "This is the test.py file that is compiled to test.mpy below" so that it now mentions that mpy-cross must be invoked with `-msmall-int-bits=30`.

tests/perf_bench/core_import_mpy_single.py
```diff
@@ -78,7 +78,7 @@ def f1():
...
...\\x12]\x12^\x12_\x12`\x12a\x12b\x12c\x12d*\nY\x12e\x12f\x12g\x12h\x12i\x12j\x12k\x12l\x12m\x12n\x12o*\x0bY\x12p\x12q\x12r\x12s\x12t\x12u\x12v\x12w\x12x\x12y\x12z*\x0bY\x12{\x12|\x12}\x12~\x12\x7f\x12\x81\x00\x12\x81\x01\x12\x81\x02\x12\x81\x03\x12\x81\x04*\nY\x12\x81\x05\x12\x81\x06\x12\x81\x07\x12\x81\x08\x12\x81\t\x12\x81\n\x12\x81\x0b\x12\x81\x0c\x12\x81\r\x12\x81\x0e\x12\x81\x0f*\x0bY\x12\x81\x10\x12\x81\x11\x12\x81\x12\x12\x81\x13\x12\x81\x14\x12\x81\x15\x12\x81\x16\x12\x81\x17\x12\x81\x18\x12\x81\x19*\nY\x12\x81\x1a\x12\x81\x1b\x12\x81\x1c\x12\x81\x1d\x12\x81\x1e\x12\x81\x1f\x12\x81 \x12\x81!\x12\x81\"\x12\x81#\x12\x81$*\x0bY\x12\x81%\x12\x81&*\x02Y\x12\x81'\x12\x81(\x12\x81)\x12\x81*\x12\x81+\x12\x81,\x12\x81-\x12\x81.\x12\x81/\x12\x810*\nY\x12\x811\x12\x812\x12\x813\x12\x814*\x04Y\x12\x815\x12\x816\x12\x817\x12\x818*\x04Y\x12\x819\x12\x81:\x12\x81;\x12\x81<*\x04YQc\x87p\x08@\x05\x80###############################\x00\xc0#\x01\xc0#\x02\xc0#\x03\xc0#\x04\xc0#\x05\xc0#\x06\xc0#\x07\xc0#\x08\xc0#\t\xc0#\n\xc0#\x0b\xc0#\x0c\xc0#\r\xc0#\x0e\xc0#\x0f\xc0#\x10\xc0#\x11\xc0#\x12\xc0#\x13\xc0#\x14\xc0#\x15\xc0#\x16\xc0#\x17\xc0#\x18\xc0#\x19\xc0#\x1a\xc0#\x1b\xc0#\x1c\xc0#\x1d\xc0Qc"
```
dpgeorge, line 81: Please update the comment above for `-msmall-int-bits`.

tests/stress/fun_call_limit.py
```diff
@@ -16,7 +16,7 @@ def test(n):
 
 
 # If the port has at least 32-bits then this test should pass.
-print(test(29))
+print(test(28))
```
dpgeorge, line 19: I don't understand why this change is needed.

tools/ci.sh
```diff
@@ -522,6 +522,14 @@ CI_UNIX_OPTS_SANITIZE_UNDEFINED=(
     LDFLAGS_EXTRA="-fsanitize=undefined -fno-sanitize=nonnull-attribute"
 )
 
+CI_UNIX_OPTS_REPR_B=(
+    VARIANT=coverage
+    CFLAGS_EXTRA="-DMICROPY_OBJ_REPR=MICROPY_OBJ_REPR_B -Dmp_int_t=int32_t -Dmp_uint_t=uint32_t"
+    MICROPY_FORCE_32BIT=1
+    RUN_TESTS_MPY_CROSS_FLAGS="--mpy-cross-flags=\"-march=x86 -msmall-int-bits=29\""
```
dpgeorge: Shouldn't it be `-msmall-int-bits=30`?

tools/ci.sh
```diff
@@ -668,6 +676,15 @@ function ci_unix_32bit_setup {
     python3 --version
 }
 
+function ci_unix_coverage_repr_b_build {
+    ci_unix_build_helper "${CI_UNIX_OPTS_REPR_B[@]}"
+    ci_unix_build_ffi_lib_helper gcc -m32
+}
+
+function ci_unix_coverage_repr_b_run_tests {
+    ci_unix_run_tests_helper "${CI_UNIX_OPTS_REPR_B[@]}"
```
dpgeorge, line 705: Can we use `ci_unix_run_tests_full_helper` here?  Or at least `ci_unix_run_tests_full_no_native_helper`?

dpgeorge:
> would it be easier if I split this up?

No, it's best together to see exactly what's needed to get repr B working under CI.

extmod/moductypes.c
jepler: VAL_TYPE_BITS, BITF_LEN_BITS, and BITF_OFF_BITS are the same in both sides of the `#if` now !? This makes zero sense, and I have no idea why I did that. Will revert.

extmod/moductypes.c
jepler: In principle, yes. In practice, I spent some time trying to do so and failed. I will try some more tomorrow, because simplifying this would be good. Failing that, I will comment it better.

dpgeorge: Actually I don't think it's possible.  Because on 64-bit targets we still want this to use 32-bit small-int ranges.  I think that's so the constants are the same across "all" targets.

Well, now with this change they won't be, repr B will have different constants.  I guess we could in principle change all targets to use the same, smaller, sizing.  But probably that's not a good idea.

So... how you have it looks like the best way to do it.

tools/ci.sh
jepler: As noted above
> [Native & viper tests fail if --via-mpy](https://github.com/micropython/micropython/pull/17688#issuecomment-3075383706) even after accounting for small int size.

full_no_native still runs viper tests via mpy.

extmod/moductypes.c
jepler: I did go ahead and do this in a way that widened the values to be whatever mp_int_t can accommodate. I guess this is the cause of the slight growth in the x86_64 build?

tests/stress/fun_call_limit.py
jepler: This test fails in REPR_B with the argument 29:
```
FAILURE /home/jepler/src/micropython/tests/results/stress_fun_call_limit.py
--- /home/jepler/src/micropython/tests/results/stress_fun_call_limit.py.exp	2025-09-16 09:50:39.647389928 -0500
+++ /home/jepler/src/micropython/tests/results/stress_fun_call_limit.py.out	2025-09-16 09:50:39.647389928 -0500
@@ -1,2 +1,2 @@
-32
+SyntaxError
 SyntaxError
make: Leaving directory '/home/jepler/src/micropython/ports/unix'
```

tools/ci.sh
jepler: yes, fixed.

tests/basics/int_big_to_small_int29.py
jepler: what are the characteristics of mp_int_t on pic16bit? if this is not needed I'll zap it but that's what I had in mind.

tests/perf_bench/core_import_mpy_multi.py
jepler: done (both places)

tests/stress/fun_call_limit.py
jepler: The test is `if (i >= MP_SMALL_INT_BITS - 1)` so the upper bound depends on the number of bits stolen by the obj_repr, not the actual number of bits in the underlying C type. Maybe this is another spot where MP_INT_BITS (introduced by me in this PR but local to uctypes) would be good to use, since this is about the C type not Python small ints.

tests/basics/int_big_to_small_int29.py
dpgeorge: pic16bit is not run or tested, so we don't have anything that would fail this test.  So I think just remove it.

If we ever have a 16-bit (or 8-bit?) port that needs to run the test suite, I think there are a lot of other things that'll need tweaking.

tests/perf_bench/core_import_mpy_multi.py
dpgeorge: I don't see these changes, maybe you forgot to push?

jepler:
[actually pushed now]

tools/ci.sh
dpgeorge: Ah, I see.

Your remark is correct, that viper makes assumptions about small int encoding.

Maybe add a comment to that effect here, so we know in the future why it doesn't test mpy?  (And maybe it can be fixed one day...)

tests/stress/fun_call_limit.py
dpgeorge: OK, thanks for the reasoning.



> since this is about the C type not Python small ints.

Actually, it is about small ints because `star_args` (which holds `1 << i`) is encoded as a small int in the bytecode.  So the C code is correct.

I suggest therefore:
- change the comment above this line to something like "This is the biggest number that will pass on a 32-bit port using object representation B, which has 1<<28 still fitting in a positive small int."
- the `for i in range(30, 70):` below should be changed to `for i in range(28, 70):`, so that it tests the transition from non-syntaxerror to syntaxerror (it should've been `range(29,70)` prior to this PR!)

tests/basics/int_big_to_small_int29.py
dpgeorge: Please add comment.

extmod/moductypes.c
```diff
@@ -69,11 +73,11 @@ enum {
...
 // Here we need to set sign bit right
-#define TYPE2SMALLINT(x, nbits) ((((int)x) << (32 - nbits)) >> 1)
-#define GET_TYPE(x, nbits) (((x) >> (31 - nbits)) & ((1 << nbits) - 1))
-// Bit 0 is "is_signed"
+#define TYPE2SMALLINT(x, nbits) ((mp_int_t)(((mp_uint_t)x) << (MP_INT_BITS - nbits)) >> LOST_INT_BITS)
+#define GET_TYPE(x, nbits) (((x) >> (MP_SMALL_INT_BITS - nbits)) & ((1 << nbits) - 1))
+// Bit 0 is "is_signed", other bits give log2 of scalar size
 #define GET_SCALAR_SIZE(val_type) (1 << ((val_type) >> 1))
-#define VALUE_MASK(type_nbits) ~((int)0x80000000 >> type_nbits)
+#define VALUE_MASK(type_nbits) ~(MP_SMALL_INT_MIN >> type_nbits)
```
dpgeorge, line 80: It's nice to have these constants generalised like this.  But, I think it's a (subtle) breaking change: this will change `uctypes` constant values and those can be embedded into .mpy files because the parser can optimise them to their int values (see `mp_constants_table`).

So, eg, someone using `uctypes` in a compiled .mpy file on a 64-bit arch, it won't work as expected after this change.

This change also means that, moving forward, building a 32-bit .mpy file on a 64-bit arch will use large ints for the substituted constants, rather than small ints.  So it'll be suboptimal.

It's annoying, but we are stuck here.

Maybe the simplest way to handle this is just hard-code values like this:
```c

#define MP_INT_BITS (32)
#define MP_UINT_MAX (0xffffffff)
#define MP_INT_MAX ((mp_int_t)(MP_UINT_MAX / 2))
#define MP_INT_MIN (-MP_INT_MAX - 1)
```

dpgeorge: Upon further thought, this just isn't going to work at all.  Consider the situation where mpy-cross builds a .mpy file using `uctypes`.  That will use 31-bit small ints to encode these `uctypes` constants.  Then if that .mpy is loaded on a obj-repr-B system the constant won't match the constant values used by that target.

There's not really an easy way to fix this.  Either we need to change the constants for all targets so they fit on repr B.  Or we need to have a way of storing in the .mpy which bit size was used.  Both those options require a bump to the .mpy version (and the latter option is pretty tricky).

dpgeorge: The main point here is, if we want to allow saving of .mpy files that use `uctypes`, the constants must be the same across all platforms.

Some ways to make it work for repr B, or work around the issue :
- Store the constants as big ints, instead of small ints; see eg `sys.maxsize` for how to do that.  I tried this and it can work, but it requires quite a few changes (eg MP_OBJ_SMALL_INT_VALUE -> mp_obj_get_int_truncated everywhere in `moductypes.c`)
- disable `uctypes` for obj repr B
- disable loading of .mpy files for obj repr B

jepler:
Thanks for all that background, I was unaware of `mp_constants_table`.

I think that for my purposes, I can accept the cost of storing these as long ints and the code/memory overhead of doing so. I'll spin a fresh PR that will use macros to switch between "generic integer" APIs specifically in the case of REPR_B and the SMALL_INT APIs for all other REPRs.

The other alternative I could think of would be to disable uctypes in REPR_B builds here in micropython and I just carry the incompatible mpy constant changes over in m68k-micropython.

I'm good with either of those alternatives, but honestly as I think about it the 2nd is almost preferable (and I can put back REPR_B uctypes testing over in my m68k-micropython fork).

jepler:
Closing in favor of #18109

PR #17693, opened 2025-07-16
py/objtype: Add support for __set_name__. (no-self-modification version)

AJMansfield:
### Summary

This PR implements the feature described in https://github.com/micropython/micropython/issues/15501, adding support for the `__set_name__` data model method.

In particular, this implementation leverages the existing `is_fixed` flag bit on micropython's flag type to avoid the potential modify-while-iterating hazard in the simplest possible way, i.e. by preventing `__set_name__` from modifying the owner class.

### Testing

This PR includes and passes the unit test originally submitted in https://github.com/micropython/micropython/pull/15500 to verify the feature's absence, as well as additional tests to specifically verify that attempting to mutate the descriptor's owner triggers an error.

This is the least powerful of the four ways I've tried implementing this (just allowing the hazard in #15503, or copying a list of calls to make in #16806, or making a private copy of the dict in #16816), as the ability for a descriptor to add siblings or transparently delete or replace itself isn't an especially-rarely-used functionality in descriptor-heavy code.

The functionality in question can, though, for the most part be replaced by inheriting the owner class from a suitable parent class with an `__init_subclass__` method that handles this on behalf of its descriptors. This will need to be documented as a cpydiff workaround, though that can only really be considered a workaround once we have actual `__init_subclass__` (#15511) or at least documented workarounds to that method's nonfunctionality (#16786).

As an alternative, maybe this PR could be combined with either #16806 or #16816 with a feature flag that defaults to this minimal 'disallow modifying the owner' implementation, but can be set for platforms/applications where the full cost of making the extra copies is tolerable?

### DIFF
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -1124,7 +1124,7 @@ typedef time_t mp_timestamp_t;
 #define MICROPY_PY_FUNCTION_ATTRS_CODE (MICROPY_CONFIG_ROM_LEVEL_AT_LEAST_FULL_FEATURES)
 #endif
 
-// Whether to support the descriptors __get__, __set__, __delete__
+// Whether to support the descriptors __get__, __set__, __delete__, __set_name__
 // This costs some code size and makes load/store/delete of instance
 // attributes slower for the classes that use this feature
 #ifndef MICROPY_PY_DESCRIPTORS
--- a/py/objtype.c
+++ b/py/objtype.c
@@ -661,8 +661,8 @@ static void mp_obj_instance_load_attr(mp_obj_t self_in, qstr attr, mp_obj_t *des
     // try __getattr__
     if (attr != MP_QSTR___getattr__) {
         #if MICROPY_PY_DESCRIPTORS
-        // With descriptors enabled, don't delegate lookups of __get__/__set__/__delete__.
-        if (attr == MP_QSTR___get__ || attr == MP_QSTR___set__ || attr == MP_QSTR___delete__) {
+        // With descriptors enabled, don't delegate lookups of __get__/__set__/__delete__/__set_name__.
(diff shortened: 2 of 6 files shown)


dpgeorge:
I like this approach, it's nice and simple, and safe.



> the ability for a descriptor to add siblings or transparently delete or replace itself isn't an especially-rarely-used functionality in descriptor-heavy code.

Do you think we'd be severely limiting this feature by disallowing locals dict updates?

AJMansfield:
> Do you think we'd be severely limiting this feature by disallowing locals dict updates?

The use cases aren't overwhelming, so there's definitely merit to just disallowing locals dict updates; or maybe having a feature flag for this.

Just to describe a few, though:

Some descriptors need to coordinate with other descriptor instances on the same class (e.g. a serialization or ORM libraries); in these cases, it's common to have the first descriptor create an attribute on the class to hold a data structure that they all register in. (One of my own libraries use owner class attrs like this for the tables it uses to work out what registers are contiguous, for I2C transaction pooling.)

```python
class Registered:
    def __set_name__(self, owner, name):
        try:
            owner.__registry[name] = self
        except AttributeError:
            owner.__registry = {name: self}
```

In CPython, a library writer can also do this using a mixin class, with an `__init_subclass__` method --- but that's extra boilerplate that users easily forget, and limits extensibility since the mixin class needs to know how all of the descriptors should be processed. (Also, Micropython doesn't support multiple inheritance, so regular mixins are already out for a lot of use cases.)

It's also potentially a performance optimization in the case where a descriptor can replace itself with a concrete value at class-creation time, to avoid the indirection of dispatching access through a getter:

```python
class Configured:
    def __set_name__(self, owner, name):
        concrete_value = query_config_database(name)
        setattr(owner, name, concrete_value)

class MyApp:
    TIME_BETWEEN_WIDGETS: int = Configured()
```

AJMansfield:
Combined into #16806.

PR #17697, opened 2025-07-17
tests/run-tests.py: Use TEST_TIMEOUT as timeout for bare-metal tests.

dpgeorge:
### Summary

This TEST_TIMEOUT parameter is already used for PC-based tests (eg unix and webassembly ports), and it makes sense for it to be used for bare-metal ports as well.  That way the timeout is configurable for all targets.

Because this increases the default timeout from 10s to 30s, this fixes some long-running tests that would previously fail due to a timeout such as `thread/stress_aes.py` on ESP32.

### Testing

Tested with ESP32_GENERIC firmware.  The `thread/stress_aes.py` test now passes (previously it would fail due to timeout).

### DIFF
--- a/tests/run-tests.py
+++ b/tests/run-tests.py
@@ -15,7 +15,7 @@ from multiprocessing.pool import ThreadPool
 import threading
 import tempfile
 
-# Maximum time to run a PC-based test, in seconds.
+# Maximum time to run a single test, in seconds.
 TEST_TIMEOUT = float(os.environ.get('MICROPY_TEST_TIMEOUT', 30))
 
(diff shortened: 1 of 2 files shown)


dpgeorge:
@hmaerki FYI.  This should fix both https://github.com/octoprobe/testbed_micropython/issues/37 and  https://github.com/octoprobe/testbed_micropython/issues/40

hmaerki:
stress_aes.py / pow3_intbig.py.out does not timeout anymore in https://reports.octoprobe.org/github_selfhosted_testrun_185/octoprobe_summary_report.html.

Please merge this PR - thanks!

PR #17698, opened 2025-07-17
esp32/machine_uart: Change `sendbreak()` implementation to simply pull the pin low for the break period

dpgeorge:
### Summary

Currently, `UART.sendbreak()` on esp32 will reconfigure the UART to a slower baudrate and send out a null byte, to synthesise a break condition.  That's not great because it changes the baudrate of the RX path as well, which could miss incoming bytes while sending the break.

This PR changes the sendbreak implementation to just reconfigure the TX pin as GPIO in output mode, and hold the pin low for the required duration.

### Testing

Existing test is updated (simplified) and works well on ESP32_GENERIC.

Also tested that RPI_PICO_W still passes this test.

There's `uart_write_bytes_with_break()` but that requires at least 1 byte to be sent before the break, so we can't use that.

There's also the lower level `uart_hal_tx_break()` but I'm not sure how to use that properly in conjunction with the higher level driver.

### DIFF
--- a/ports/esp32/machine_uart.c
+++ b/ports/esp32/machine_uart.c
@@ -508,20 +508,21 @@ static bool mp_machine_uart_txdone(machine_uart_obj_t *self) {
 }
 
 static void mp_machine_uart_sendbreak(machine_uart_obj_t *self) {
-    // Save settings
+    // Calculate the length of the break, as 13 bits.
     uint32_t baudrate;
     check_esp_err(uart_get_baudrate(self->uart_num, &baudrate));
+    uint32_t break_delay_us = 13000000 / baudrate;
 
-    // Synthesise the break condition by reducing the baud rate,
-    // and cater for the worst case of 5 data bits, no parity.
-    check_esp_err(uart_wait_tx_done(self->uart_num, pdMS_TO_TICKS(1000)));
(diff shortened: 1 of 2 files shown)


dpgeorge:
@hmaerki FYI this should fix the final issue with https://github.com/octoprobe/testbed_micropython/issues/36

hmaerki:
There is a testrun against this PR: https://reports.octoprobe.org/github_selfhosted_testrun_189/octoprobe_summary_report.html
All test concerning this PR pass!

Please merge this PR.

projectgus (approved):
Looks like a good improvement.

The only practical difference I can think of between doing it this way and using the UART HAL break function is that the break signal length will be exact if the peripheral hardware generates it whereas `mp_hal_delay_us` may delay longer than `break_delay_us` due to interrupts or a context switch. However, as the UART break signal is defined as a minimum pulse length this doesn't seem like a problem for correctness.

dpgeorge:
> `mp_hal_delay_us` may delay longer than `break_delay_us` due to interrupts or a context switch. However, as the UART break signal is defined as a minimum pulse length this doesn't seem like a problem for correctness.

Yes, that's also my understanding, that the break is a *minimum* low pulse, so it's OK for it to be longer than 13 bits.

PR #17699, opened 2025-07-17
asyncio: slight optimizations for `run_until_complete` and `sleep_ms`

greezybacon:
### Summary

This is aimed at improving the loop timing of the asyncio core loop. It makes a few small optimizations to the core and realizes about a 20% impact in overall performance.

  - In the IO poll method, the POLLIN and POLLOUT constants are looked up in the local module context rather than in the `select` module when used.
  - In `sleep_ms`, `max` is not used for each call. Instead, an `if` expression handles the case when `t` is negative.
  - In `run_until_complete`, a call to `max` is avoided
  - In `run_until_complete` the methods for the task and IO queues are only looked up once.

### Testing

I ran two tests on three platforms. Source code is given below. The tight-loop just runs a single task as quickly as possible. The second task uses a ThreadSafeFlag to run two tasks as quickly as possible but requires IO polling between the tasks.

| test       | platform                      | base (v1.25.0) | PR     | Change |
|------------|-------------------------------|----------------|--------|--------|
| tight-loop | unix (ubuntu 22.04 on Mac M2) | 1.65us         | 1.26us | +23.6% |
|            | mimxrt (Teensy 4.1)           | 47us           | 42us   | +10.6% |
|            | rp2 (W5100S EVB PICO @ 125MHz)| 554us          | 495us  | +10.6% |
| io-poll    | unix                          | 2982us         | 3004us | -0.7%  |
|            | mimxrt                        | 253us          | 222us  | +12.2% |
|            | rp2                           | 2107us         | 1972us | +6.4%  |

tight-loop test
```python
import asyncio

async def count():
    global counter
    while True:
        await asyncio.sleep_ms(0)
        counter += 1

try:
    counter = 0
    asyncio.run(asyncio.wait_for(count(), timeout=2))
finally:
    print(counter, 2e6/counter)
```

io-poll test
```python
import asyncio

flag = asyncio.ThreadSafeFlag()
async def sender():
    while True:
        flag.set()
        await asyncio.sleep_ms(0)

async def recv():
    global counter
    while True:
        await flag.wait()
        counter += 1

counter = 0
try:
    asyncio.create_task(sender())
    asyncio.run(asyncio.wait_for(recv(), timeout=2))
finally:
    if counter:
        print(counter, 2e6/counter)
```

### DIFF
--- a/extmod/asyncio/core.py
+++ b/extmod/asyncio/core.py
@@ -3,6 +3,7 @@
 
 from time import ticks_ms as ticks, ticks_diff, ticks_add
 import sys, select
+from select import POLLIN, POLLOUT
 
 # Import TaskQueue and Task, preferring built-in C code over Python code
 try:
@@ -66,6 +67,9 @@ def sleep(t):
 ################################################################################
 # Queue and poller for stream IO
 
+nPOLLIN = ~POLLIN
+nPOLLOUT = ~POLLOUT
+
 
 class IOQueue:
     def __init__(self):
@@ -77,13 +81,13 @@ class IOQueue:
             entry = [None, None, s]
             entry[idx] = cur_task
             self.map[id(s)] = entry
-            self.poller.register(s, select.POLLIN if idx == 0 else select.POLLOUT)
+            self.poller.register(s, POLLIN if idx == 0 else POLLOUT)
         else:
             sm = self.map[id(s)]
             assert sm[idx] is None
             assert sm[1 - idx] is not None
             sm[idx] = cur_task
-            self.poller.modify(s, select.POLLIN | select.POLLOUT)
+            self.poller.modify(s, POLLIN | POLLOUT)
         # Link task to this IOQueue so it can be removed if needed
         cur_task.data = self
 
@@ -114,20 +118,20 @@ class IOQueue:
         for s, ev in self.poller.ipoll(dt):
             sm = self.map[id(s)]
             # print('poll', s, sm, ev)
-            if ev & ~select.POLLOUT and sm[0] is not None:
+            if ev & nPOLLOUT and sm[0] is not None:
                 # POLLIN or error
                 _task_queue.push(sm[0])
                 sm[0] = None
-            if ev & ~select.POLLIN and sm[1] is not None:
+            if ev & nPOLLIN and sm[1] is not None:
                 # POLLOUT or error
                 _task_queue.push(sm[1])
                 sm[1] = None
             if sm[0] is None and sm[1] is None:
                 self._dequeue(s)
             elif sm[0] is None:
-                self.poller.modify(s, select.POLLOUT)
+                self.poller.modify(s, POLLOUT)
             else:
-                self.poller.modify(s, select.POLLIN)
+                self.poller.modify(s, POLLIN)
 
 
 ################################################################################
@@ -153,12 +157,15 @@ def run_until_complete(main_task=None):
     global cur_task
     excs_all = (CancelledError, Exception)  # To prevent heap allocation in loop
     excs_stop = (CancelledError, StopIteration)  # To prevent heap allocation in loop
+    queue_peek = _task_queue.peek
+    queue_pop = _task_queue.pop
+    wait_io_event = _io_queue.wait_io_event
     while True:
         # Wait until the head of _task_queue is ready to run
         dt = 1
         while dt > 0:
             dt = -1
-            t = _task_queue.peek()
+            t = queue_peek()
             if t:
                 # A task waiting on _task_queue; "ph_key" is time to schedule task at
                 dt = max(0, ticks_diff(t.ph_key, ticks()))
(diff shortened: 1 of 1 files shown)


extmod/asyncio/core.py
```diff
@@ -54,7 +55,8 @@ def __next__(self):
 # Use a SingletonGenerator to do it without allocating on the heap
 def sleep_ms(t, sgen=SingletonGenerator()):
     assert sgen.state is None
-    sgen.state = ticks_add(ticks(), max(0, t))
+    now = ticks()
+    sgen.state = ticks_add(now, t) if t > 0 else now
```
dpgeorge: Does this give a measurable speed improvement?  Is it worth it for the cost in code size?

I measure this change here as +5 bytes to the bytecode.  The most taken path will be when `ticks_add()` needs to be called, which goes from 12 opcodes previously to now 16 opcodes.  It's usually the opcode overhead that's slow, rather than the actual call (eg out to `max`, which should be quick with two small int args).  So I would guess that this change actually makes things a little slower.

extmod/asyncio/core.py
```diff
@@ -153,24 +157,29 @@ def run_until_complete(main_task=None):
...
+        while dt:
             dt = -1
-            t = _task_queue.peek()
+            t = queue_peek()
             if t:
                 # A task waiting on _task_queue; "ph_key" is time to schedule task at
-                dt = max(0, ticks_diff(t.ph_key, ticks()))
+                dt = ticks_diff(t.ph_key, ticks())
+                if dt < 0:
+                    dt = 0
```
dpgeorge: As above, does this change here actually make things faster?

extmod/asyncio/core.py
greezybacon: @dpgeorge thank you for asking. I spent too long working on this. It turns out that it does make a performance improvement on all platforms (ie. both Unix and MCUs), but it isn't substantial. I'm happy to remove it and search for a more substantial impact.

dpgeorge: Thanks for investigating.  Unless it's a significant improvement, I'd prefer to leave it as-is (ie prefer shorter bytecode over performance).

extmod/asyncio/core.py
```diff
@@ -3,6 +3,7 @@
 
 from time import ticks_ms as ticks, ticks_diff, ticks_add
 import sys, select
+from select import POLLIN, POLLOUT
```
jepler, line 6: alas I don't think it works to write `POLLIN = const(select.POLLIN)`. I wonder if `POLLIN = const(1); assert POLLIN == select.POLLIN` benefits performance enough that it would be worth doing. (is micropython asyncio supposed to be cpython compatible? I guess there's no guarantee of the value of POLLIN/POLLOUT constants there. But there's no `time.ticks_ms` so probably this is a non-goal)

extmod/asyncio/core.py
```diff
@@ -77,13 +80,13 @@ def _enqueue(self, s, idx):
...
             self.map[id(s)] = entry
-            self.poller.register(s, select.POLLIN if idx == 0 else select.POLLOUT)
+            self.poller.register(s, POLLIN if idx == 0 else POLLOUT)
         else:
             sm = self.map[id(s)]
             assert sm[idx] is None
             assert sm[1 - idx] is not None
             sm[idx] = cur_task
-            self.poller.modify(s, select.POLLIN | select.POLLOUT)
+            self.poller.modify(s, POLLIN | POLLOUT)
```
jepler, line 90: any measurable benefit to having `POLLANY = POLLIN | POLLOUT` to avoid the calculation here?

jepler (approved):
A few suggestions, but please feel free to ignore.

extmod/asyncio/core.py
dpgeorge: That's an interesting suggestion, to just use plain constants, eg `_POLLIN = const(1); _POLLOUT=const(4)`.  Those constants were chosen to be compatible with Linux, and actually the unix port (and windows) uses the common `extmod/modselect.c` along with the common POLLxx constant values.

So... if we assume no one else is implementing their own `select` module... it might actually be OK to hard-code these constants here.  That would be more performant and reduce code size.

extmod/asyncio/core.py
peterhinch: If they are declared `const` the calculation is done at compile time.

extmod/asyncio/core.py
```diff
@@ -153,12 +156,15 @@ def run_until_complete(main_task=None):
...
     excs_all = (CancelledError, Exception)  # To prevent heap allocation in loop
     excs_stop = (CancelledError, StopIteration)  # To prevent heap allocation in loop
+    queue_peek = _task_queue.peek
+    queue_pop = _task_queue.pop
+    wait_io_event = _io_queue.wait_io_event
     while True:
         # Wait until the head of _task_queue is ready to run
         dt = 1
-        while dt > 0:
+        while dt:
```
greezybacon: oops- I didn't mean to make this edit.

greezybacon:
@dpgeorge I recalculated the timings without the `max` function call avoidance. The numbers are a bit less but still a sizable improvement. I have another idea I found to improve the performance further I'll propose as a separate PR.

extmod/asyncio/core.py
dpgeorge: @greezybacon what do you think about that, about defining these constants here, instead of importing them?

PR #17704, opened 2025-07-17
Fix mpprintf argument type errors

jepler:
### Summary

I have developed a gcc plugin for checking mp_printf format strings vs argument types at compile time. https://github.com/micropython/micropython/pull/17556 is the fixes plus the checker, while this PR is only the fixes.

### Testing

I added tests to the coverage build for stuff I felt was not yet adequately covered.

### DIFF
--- a/extmod/modlwip.c
+++ b/extmod/modlwip.c
@@ -907,7 +907,7 @@ static const mp_obj_type_t lwip_socket_type;
 
 static void lwip_socket_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     lwip_socket_obj_t *self = MP_OBJ_TO_PTR(self_in);
-    mp_printf(print, "<socket state=%d timeout=%d incoming=", self->state, self->timeout);
+    mp_printf(print, "<socket state=%d timeout=" UINT_FMT " incoming=", self->state, self->timeout);
     if (self->type == MOD_NETWORK_SOCK_STREAM) {
         mp_printf(print, "%p off=%d>", self->incoming.tcp.pbuf, self->recv_offset);
     } else {
--- a/py/objcell.c
+++ b/py/objcell.c
@@ -30,7 +30,7 @@
 static void cell_print(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t kind) {
     (void)kind;
     mp_obj_cell_t *o = MP_OBJ_TO_PTR(o_in);
-    mp_printf(print, "<cell %p ", o->obj);
+    mp_printf(print, "<cell " HEX_FMT " ", (mp_uint_t)o->obj);
     if (o->obj == MP_OBJ_NULL) {
         mp_print_str(print, "(nil)");
     } else {
--- a/examples/usercmodule/cexample/examplemodule.c
+++ b/examples/usercmodule/cexample/examplemodule.c
@@ -86,12 +86,12 @@ static void example_AdvancedTimer_print(const mp_print_t *print, mp_obj_t self_i
     mp_uint_t elapsed = mp_obj_get_int(example_Timer_time(self_in));
 
     // We'll make all representations print at least the class name.
-    mp_printf(print, "%q()", MP_QSTR_AdvancedTimer);
+    mp_printf(print, "%q()", (qstr)MP_QSTR_AdvancedTimer);
 
     // Decide what else to print based on print kind.
     if (kind == PRINT_STR) {
         // For __str__, let's attempt to make it more readable.
-        mp_printf(print, "  # created %d seconds ago", elapsed / 1000);
+        mp_printf(print, "  # created %d seconds ago", (int)(elapsed / 1000));
     }
 }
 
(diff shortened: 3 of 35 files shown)


py/objcell.c
```diff
@@ -30,7 +30,8 @@
 static void cell_print(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t kind) {
     (void)kind;
     mp_obj_cell_t *o = MP_OBJ_TO_PTR(o_in);
-    mp_printf(print, "<cell %p ", o->obj);
+    MP_STATIC_ASSERT(sizeof(mp_obj_t) == sizeof(mp_uint_t));
+    mp_printf(print, "<cell " HEX_FMT " ", (mp_uint_t)o->obj);
```
dpgeorge, line 33: It looks like this is the only place that uses the new `HEX_FMT`.  Can it just stay as `%p`, and if needed cast `o->obj` to `(void *)`?

dpgeorge:
If `HEX_FMT` is really needed, maybe a simpler change would be to delete as many defines of `INT_FMT` and `UINT_FMT` as possible and let the defaults from `py/mpconfig.h` be used instead?

jepler:
Introducing and using HEX_FMT allows the nanbox variant to print cells correctly:
```
<closure <function <lambda> at 0xf79e7300> at f79e7320, n_closed=1 <cell f79e72a0 []> >
<closure <function <lambda> at 0xf79e7360> at f79e7380, n_closed=1 <cell 1000000000003 1> >
<closure <function <lambda> at 0xf79e73c0> at f79e73e0, n_closed=1 <cell c00d1eb851eb851f 3.14> >
<closure <function <lambda> at 0xf79e7440> at f79e7460, n_closed=1 <cell 200000000002b '__main__'> >
```
If the argument is cast to a pointer type and the `%p` format is used, then only the lower 32 bits are printed:
```
<closure <function <lambda> at 0xf79d8300> at f79d8320, n_closed=1 <cell f79d82a0 []> >
<closure <function <lambda> at 0xf79d8360> at f79d8380, n_closed=1 <cell 3 1> >
<closure <function <lambda> at 0xf79d83c0> at f79d83e0, n_closed=1 <cell 51eb851f 3.14> >
<closure <function <lambda> at 0xf79d8440> at f79d8460, n_closed=1 <cell 2b '__main__'> >
```

jepler:
I noticed I introduced a bug with printing pointers in https://github.com/micropython/micropython/pull/17618. I put a fix for it here but can break it out to its own PR if desired: 333665bc140d494132c7b888a64a4e138a09c831

dpgeorge:
> I noticed I introduced a bug with printing pointers in #17618. I put a fix for it here but can break it out to its own PR if desired: [333665b](https://github.com/micropython/micropython/commit/333665bc140d494132c7b888a64a4e138a09c831)

It's OK to keep it here.

dpgeorge:
> Introducing and using HEX_FMT allows the nanbox variant to print cells correctly:

Oh!  In that case how about my suggestion to let the defaults in `py/mpconfig.h` be used?

jepler:
The one remaining failure is `1 tests failed: extmod/select_poll_eintr.py`

py/objcell.c
```diff
@@ -30,7 +30,8 @@
 static void cell_print(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t kind) {
     (void)kind;
     mp_obj_cell_t *o = MP_OBJ_TO_PTR(o_in);
-    mp_printf(print, "<cell %p ", o->obj);
+    MP_STATIC_ASSERT(sizeof(mp_obj_t) == sizeof(mp_uint_t));
```
dpgeorge: I don't think it makes sense to have this assert here, because this property is already fundamental to the entire runtime.

It's documented somewhat at the top of `py/obj.h`, where `mp_obj_t` is defined (although the comment there could be extended to note that `mp_uint_t` and `mp_int_t` need to be the same size as `mp_obj_t` (see also the definition of `MP_BYTES_PER_OBJ_WORD` in `py/mpconfig.h`).

If anything, this assert could go at the start of `mp_init()`.

extmod/modlwip.c
```diff
@@ -907,7 +907,7 @@ static const mp_obj_type_t lwip_socket_type;
 
 static void lwip_socket_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
     lwip_socket_obj_t *self = MP_OBJ_TO_PTR(self_in);
-    mp_printf(print, "<socket state=%d timeout=%d incoming=", self->state, self->timeout);
+    mp_printf(print, "<socket state=%d timeout=" INT_FMT " incoming=", self->state, self->timeout);
```
dpgeorge: I think this should be `UINT_FMT`?  Because `self->timeout` is `mp_uint_t`.

jepler: done

py/objcell.c
jepler: static assert removed.

extmod/modlwip.c
dpgeorge: I can't see this change?

py/objcell.c
dpgeorge: Did you commit this change?

jepler:
I now actually pushed those two changes.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 0c8d35b322554293c0847a618fa90760cefa1b60 through ab620f40844a837546497252849933ef2f685f4d

jepler:
Thank you!

PR #17708, opened 2025-07-18
shared/tinyusb: Fix build errors with CDC support disabled.

agatti:
### Summary

This PR makes it possible to build selected MicroPython ports that would have USB support enabled but CDC support disabled as well (whilst keeping MSC support for example).

The original code does support such a configuration but missed a few spots where build errors would arise.  These changes fix the remaining issues, fixing also warnings caused by the changes needed to make the build succeed (like unused variables and so on).

### Testing

This was checked by building the ESP32-S3 firmware with `MICROPY_HW_USB_CDC` set to 0, and then making sure the build process would succeed with that flag set back to 1.  I also have another (unreleased) port where this behaviour was originally discovered and then fixed.

There shouldn't be any impact, as these changes only affect a specific configuration whilst leaving the rest of the codebase as it is otherwise.

### DIFF
--- a/ports/esp32/usb.c
+++ b/ports/esp32/usb.c
@@ -28,15 +28,15 @@
 #include "py/mphal.h"
 #include "usb.h"
 
-#if MICROPY_HW_USB_CDC
-#include "esp_rom_gpio.h"
+#if MICROPY_HW_ENABLE_USBDEV
+
 #include "esp_mac.h"
+#include "esp_rom_gpio.h"
 #include "esp_private/usb_phy.h"
 
 #include "shared/tinyusb/mp_usbd.h"
 
 static usb_phy_handle_t phy_hdl;
-
 
 void usb_init(void) {
     // ref: https://github.com/espressif/esp-usb/blob/4b6a798d0bed444fff48147c8dcdbbd038e92892/device/esp_tinyusb/tinyusb.c
@@ -77,4 +77,4 @@ void mp_usbd_port_get_serial_number(char *serial_buf) {
     mp_usbd_hex_str(serial_buf, mac, sizeof(mac));
 }
 
-#endif // MICROPY_HW_USB_CDC
+#endif // MICROPY_HW_ENABLE_USBDEV
--- a/ports/rp2/mphalport.c
+++ b/ports/rp2/mphalport.c
@@ -50,7 +50,7 @@
 static uint64_t time_us_64_offset_from_epoch;
 #endif
 
-#if MICROPY_HW_ENABLE_UART_REPL || MICROPY_HW_USB_CDC
+#if MICROPY_HW_ENABLE_UART_REPL || MICROPY_HW_USB_CDC || MICROPY_PY_OS_DUPTERM_NOTIFY
 
 #ifndef MICROPY_HW_STDIN_BUFFER_LEN
 #define MICROPY_HW_STDIN_BUFFER_LEN 512
@@ -83,11 +83,12 @@ int mp_hal_stdin_rx_chr(void) {
         #if MICROPY_HW_USB_CDC
         mp_usbd_cdc_poll_interfaces(0);
         #endif
-
+        #if MICROPY_HW_USB_CDC || MICROPY_HW_ENABLE_UART_REPL || MICROPY_PY_OS_DUPTERM_NOTIFY
         int c = ringbuf_get(&stdin_ringbuf);
         if (c != -1) {
             return c;
         }
+        #endif
         #if MICROPY_PY_OS_DUPTERM
         int dupterm_c = mp_os_dupterm_rx_chr();
         if (dupterm_c >= 0) {
--- a/shared/tinyusb/mp_usbd.h
+++ b/shared/tinyusb/mp_usbd.h
@@ -38,13 +38,16 @@
 #ifndef NO_QSTR
 #include "tusb.h"
 #include "device/dcd.h"
+#include "class/cdc/cdc_device.h"
 #endif
 
 // Initialise TinyUSB device.
 static inline void mp_usbd_init_tud(void) {
     tusb_init();
+    #if MICROPY_HW_USB_CDC
     tud_cdc_configure_fifo_t cfg = { .rx_persistent = 0, .tx_persistent = 1 };
     tud_cdc_configure_fifo(&cfg);
+    #endif
 }
 
 // Run the TinyUSB device task
--- a/ports/mimxrt/mphalport.c
+++ b/ports/mimxrt/mphalport.c
@@ -47,7 +47,9 @@ ringbuf_t stdin_ringbuf = {stdin_ringbuf_array, sizeof(stdin_ringbuf_array), 0, 
 
 uintptr_t mp_hal_stdio_poll(uintptr_t poll_flags) {
     uintptr_t ret = 0;
+    #if MICROPY_HW_USB_CDC
     ret |= mp_usbd_cdc_poll_interfaces(poll_flags);
+    #endif
     #if MICROPY_PY_OS_DUPTERM
     ret |= mp_os_dupterm_poll(poll_flags);
     #endif
@@ -56,7 +58,9 @@ uintptr_t mp_hal_stdio_poll(uintptr_t poll_flags) {
 
 int mp_hal_stdin_rx_chr(void) {
     for (;;) {
+        #if MICROPY_HW_USB_CDC
         mp_usbd_cdc_poll_interfaces(0);
+        #endif
         int c = ringbuf_get(&stdin_ringbuf);
         if (c != -1) {
             return c;
@@ -74,11 +78,13 @@ int mp_hal_stdin_rx_chr(void) {
 mp_uint_t mp_hal_stdout_tx_strn(const char *str, mp_uint_t len) {
     mp_uint_t ret = len;
     bool did_write = false;
+    #if MICROPY_HW_USB_CDC
     mp_uint_t cdc_res = mp_usbd_cdc_tx_strn(str, len);
     if (cdc_res > 0) {
         did_write = true;
         ret = MIN(cdc_res, ret);
     }
+    #endif
     #if MICROPY_PY_OS_DUPTERM
     int dupterm_res = mp_os_dupterm_tx_strn(str, len);
     if (dupterm_res >= 0) {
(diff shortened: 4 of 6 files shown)


shared/tinyusb/mp_usbd_descriptor.c
```diff
@@ -95,10 +97,12 @@ const uint16_t *tud_descriptor_string_cb(uint8_t index, uint16_t langid) {
     if (desc_str == NULL) {
         // Fall back to the "static" string
         switch (index) {
+            #if CFG_TUD_CDC
             case USBD_STR_SERIAL:
```
dpgeorge: USB devices should have a serial string, regardless of what descriptors are defined.

So I think this should stay unconditionally in the code.

Probably the fix is then to pull the definition of `mp_usbd_port_get_serial_number()` in the esp32 port outside it's `#if` conditional.

agatti:
I found build issues also on mimxrt, rp2 and samd, need to look into all the USB-enabled ports' behaviour without CDC so I can send a single PR fixing this for good.

There's a chance I may not be able to finish this in time for 1.26.0 - I suggest postponing this to 1.26.1 or 1.27.0 to be on the safe side.  If I happen to finish it in time, then feel free to merge it in 1.26.0, I won't complain :)

shared/tinyusb/mp_usbd_descriptor.c
agatti: Indeed.  Haven't looked at USB protocol details in a decade, I assumed that `USBD_STR_SERIAL` was related to CDC support somehow.  Thanks for pointing it out, it should be fixed now.

agatti:
I believe this should be it for now - I expected it to take longer but:

- `alif`, `nrf`, and `renesas-ra` do the right thing when it comes to CDC
- `stm32` looks like it needs at least one CDC port in order to remap that to MSC if needed?  I'm not sure if making it compile with `MICROPY_HW_USB_CDC_NUM` set to 0 is the right thing to do; there's no other obvious way to disable CDC on that port (at least for my `NUCLEO_F767ZI`, maybe I've missed something).  **Edit:** #15592 already takes care of that, so it should be good to go w.r.t. building with CDC disabled :)

This will also fix cases where CDC is disabled but 1200BPS_TOUCH is not - given that they are handled separately in the integration code I assume they can be enabled independently of each other.  If it's not the case, then it's a matter of removing "class/cdc/cdc_device.h" from `shared/tinyusb/mp_usbd.h`'s includes.

There are also other minor issues I've found, like `samd` not building if USB is disabled altogether and so on, but that's something I can send a PR for at another time (or even better, somebody else doing it :)).  To keep the PR scope limited, I've only looked at making ports build using their default configuration but with USB CDC disabled.

ports/esp32/usb.c
```diff
@@ -28,13 +28,16 @@
...
-#if MICROPY_HW_USB_CDC
-#include "esp_rom_gpio.h"
+#if MICROPY_HW_ENABLE_USBDEV
+
 #include "esp_mac.h"
-#include "esp_private/usb_phy.h"
 
 #include "shared/tinyusb/mp_usbd.h"
 
+#if MICROPY_HW_USB_CDC
```
dpgeorge: If MICROPY_HW_ENABLE_USBDEV is enabled but MICROPY_HW_USB_CDC disabled, then `usb_init()` below is never compiled in.  So that gives a link error if you disable CDC (and also USB_SERIAL_JTAG) on, eg, ESP32S3.

ports/rp2/mphalport.c
```diff
@@ -83,11 +83,12 @@ int mp_hal_stdin_rx_chr(void) {
         #if MICROPY_HW_USB_CDC
         mp_usbd_cdc_poll_interfaces(0);
         #endif
-
+        #if MICROPY_HW_ENABLE_UART_REPL || MICROPY_PY_OS_DUPTERM_NOTIFY
```
dpgeorge: I'm not sure why this guard is added, because `shared/tinyusb/mp_usbd_cdc.c` pushes to `stdin_ringbuf`.

shared/tinyusb/mp_usbd.h
```diff
@@ -38,13 +38,16 @@
 #ifndef NO_QSTR
 #include "tusb.h"
 #include "device/dcd.h"
+#include "class/cdc/cdc_device.h"
```
dpgeorge, line 41: Is this extra include really needed?

agatti: This fixes `samd` (and `mimxrt` if memory serves) if CDC is disabled but `MICROPY_HW_USB_CDC_1200BPS_TOUCH` is left enabled.  With `MICROPY_HW_USB_CDC` set to 0 and this include commented out this error occurs on the `ADAFRUIT_ITSYBITSY_M4_EXPRESS` board (the default):

```
../../shared/tinyusb/mp_usbd_cdc.c: In function 'tud_cdc_line_state_cb':
../../shared/tinyusb/mp_usbd_cdc.c:184:9: error: unknown type name 'cdc_line_coding_t'
  184 |         cdc_line_coding_t line_coding;
      |         ^~~~~~~~~~~~~~~~~
../../shared/tinyusb/mp_usbd_cdc.c:185:9: error: implicit declaration of function 'tud_cdc_n_get_line_coding' [-Werror=implicit-function-declaration]
  185 |         tud_cdc_n_get_line_coding(itf, &line_coding);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~
../../shared/tinyusb/mp_usbd_cdc.c:186:24: error: request for member 'bit_rate' in something not a structure or union
  186 |         if (line_coding.bit_rate == 1200) {
      |                        ^
cc1: all warnings being treated as errors
```

ports/rp2/mphalport.c
agatti: But so do `uart_irq` and `mp_os_dupterm_notify`.  This guard covers those usages (not to mention it replicates the one used when defining `stdin_ringbuf` once the CDC part is taken care of).

ports/esp32/usb.c
agatti: I just built this using ESP-IDF 5.4 with this PR applied:

```
[/micropython/ports/esp32] $ git diff .
diff --git a/ports/esp32/mpconfigport.h b/ports/esp32/mpconfigport.h
index 721f22de1..89d08eae4 100644
--- a/ports/esp32/mpconfigport.h
+++ b/ports/esp32/mpconfigport.h
@@ -258,7 +258,7 @@
 
 // Enable stdio over native USB peripheral CDC via TinyUSB
 #ifndef MICROPY_HW_USB_CDC
-#define MICROPY_HW_USB_CDC                  (MICROPY_HW_ENABLE_USBDEV)
+#define MICROPY_HW_USB_CDC                  (0)
 #endif
 
 // Enable stdio over USB Serial/JTAG peripheral
[/micropython/ports/esp32] $ make BOARD=ESP32_GENERIC_S3 clean
Executing action: fullclean
Executing action: remove_managed_components
Done
[/micropython/ports/esp32] $ make BOARD=ESP32_GENERIC_S3 -j6      
Executing action: all (aliases: build)
Running cmake in directory /micropython/ports/esp32/build-ESP32_GENERIC_S3
...
[1468/1469] Generating binary image from built executable                             
esptool.py v4.9.0
Creating esp32s3 image...
Merged 3 ELF sections
Successfully created esp32s3 image.
Generated /micropython/ports/esp32/build-ESP32_GENERIC_S3/micropython.bin
[1469/1469] cd /micropython/ports/esp32/build-ESP32_GENERIC_S3/esp-idf/esptool_py &&.../partition-table.bin /micropython/ports/esp32/build-ESP32_GENERIC_S3/micropython.bin
micropython.bin binary size 0x18bd90 bytes. Smallest app partition is 0x1f0000 bytes. 0x64270 bytes (20%) free.

Project build complete. To flash, run:
 idf.py flash
or
 idf.py -p PORT flash
or
 python -m esptool --chip esp32s3 -b 460800 --before default_reset --after hard_reset write_flash --flash_mode dio --flash_size 4MB --flash_freq 80m 0x0 build-ESP32_GENERIC_S3/bootloader/bootloader.bin 0x8000 build-ESP32_GENERIC_S3/partition_table/partition-table.bin 0x10000 build-ESP32_GENERIC_S3/micropython.bin
or from the "/micropython/ports/esp32/build-ESP32_GENERIC_S3" directory
 python -m esptool --chip esp32s3 -b 460800 --before default_reset --after hard_reset write_flash "@flash_args"
bootloader  @0x000000    19152  (   13616 remaining)
partitions  @0x008000     3072  (    1024 remaining)
application @0x010000  1621392  (  410224 remaining)
total                  1686928
```

I'll try with `USB_SERIAL_JTAG` disabled right now.

ports/rp2/mphalport.c
dpgeorge: But if CDC is enabled, and UART_REPL and DUPTERM_NOTIFY disabled, won't that mean input from CDC will no longer work?

ports/esp32/usb.c
agatti: You were right - it did fail to build with JTAG Serial disabled.  This has been taken care of, thank you!

ports/rp2/mphalport.c
agatti: Right!  However without that guard it won't compile if CDC and UART and dupterm are disabled (often having to disable UART/debugging support of any kind is part of the security checklist of a product).

I'll update things to make it work in that case as well.

agatti: I've updated the guard so it should compile with all possible valid combinations of cdc, uart, and dupterm.  Peeking from the ringbuffer is now performed if one or more of those three flags are set, matching the ringbuffer's backing buffer definition.

agatti:
The `nrf` compile error doesn't seem to be related to this PR.  (Although I was worried I messed something up in the tinyusb integration...)

ports/rp2/mphalport.c
dpgeorge: But this whole section of code is guarded with `#if MICROPY_HW_ENABLE_UART_REPL || MICROPY_HW_USB_CDC || MICROPY_PY_OS_DUPTERM_NOTIFY` so it's redundant to have that same guard here.  What does it gain?

dpgeorge: Oh, I'm wrong about that.  It looks good now.

dpgeorge (approved):
This looks all correct now.  Thanks!

PR #17713, opened 2025-07-19
Fix incomplete implementation of readonly for VfsPosix

jepler:
### Summary

I noticed that operations such as unlink could be performed on a nominally read-only VfsPosix.

Fix all these operations, and then add a compile-time configuration option MICROPY_VFS_POSIX_WRITABLE. Disabling this option ensures that a VfsPosix instance is *ALWAYS* read-only. This may be useful when fuzzing micropython, which can [otherwise make modifications to the host filesystem](https://emergent.unpythonic.net/01522108310).

### Testing

I added a new test for VfsPosix read-only mode.

Initially, I structured the test as an importable module so that other filesystems could potentially re-use the same test code; however, most(all?) other filesystems are based on block devices, and I don't think they have the same problems with needing to add readonly checks in each code path since they just ensure the block write function cannot be called. It also turns out ci_webassembly_run_tests failed due to inability to import the auxiliary file, so I went with the one-file implementation after all.

### DIFF
--- a/extmod/vfs_posix.c
+++ b/extmod/vfs_posix.c
@@ -160,10 +160,21 @@ static mp_obj_t vfs_posix_umount(mp_obj_t self_in) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_1(vfs_posix_umount_obj, vfs_posix_umount);
 
+static bool vfs_posix_is_readonly(mp_obj_vfs_posix_t *self) {
+    return self->readonly;
+}
+
+static void vfs_posix_require_writable(mp_obj_t self_in) {
+    mp_obj_vfs_posix_t *self = MP_OBJ_TO_PTR(self_in);
+    if (vfs_posix_is_readonly(self)) {
+        mp_raise_OSError(MP_EROFS);
+    }
+}
+
 static mp_obj_t vfs_posix_open(mp_obj_t self_in, mp_obj_t path_in, mp_obj_t mode_in) {
     mp_obj_vfs_posix_t *self = MP_OBJ_TO_PTR(self_in);
     const char *mode = mp_obj_str_get_str(mode_in);
-    if (self->readonly
+    if (vfs_posix_is_readonly(self)
         && (strchr(mode, 'w') != NULL || strchr(mode, 'a') != NULL || strchr(mode, '+') != NULL)) {
         mp_raise_OSError(MP_EROFS);
     }
@@ -303,6 +314,7 @@ typedef struct _mp_obj_listdir_t {
 } mp_obj_listdir_t;
 
 static mp_obj_t vfs_posix_mkdir(mp_obj_t self_in, mp_obj_t path_in) {
+    vfs_posix_require_writable(self_in);
     mp_obj_vfs_posix_t *self = MP_OBJ_TO_PTR(self_in);
     const char *path = vfs_posix_get_path_str(self, path_in);
     MP_THREAD_GIL_EXIT();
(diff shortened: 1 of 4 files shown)


extmod/vfs_posix.c
```diff
@@ -137,7 +137,7 @@ static mp_obj_t vfs_posix_make_new(const mp_obj_type_t *type, size_t n_args, siz
         vstr_add_char(&vfs->root, '/');
     }
     vfs->root_len = vfs->root.len;
-    vfs->readonly = false;
+    vfs->readonly = MICROPY_VFS_POSIX_READONLY;
```
dpgeorge: This might be better called `MICROPY_VFS_POSIX_WRITABLE`, because:
1. it matches the existing `MICROPY_VFS_WRITABLE`
2. it's something you turn on to enable more features, rather than turning something on to disallow things (ie more a positive config rather than a negative config option)

extmod/vfs_posix.c
```diff
@@ -160,10 +160,21 @@ static mp_obj_t vfs_posix_umount(mp_obj_t self_in) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_1(vfs_posix_umount_obj, vfs_posix_umount);
 
+static bool vfs_posix_is_readonly(mp_obj_vfs_posix_t *self) {
```
dpgeorge: maybe make this `static inline`?

tests/extmod/vfs_posix_readonly.py
```diff
@@ -0,0 +1,99 @@
...
+    import gc, os, vfs, errno
+
+    vfs.VfsPosix
+except (ImportError, AttributeError):
+    print("SKIP")
+    raise SystemExit
+
+# We need a directory for testing that doesn't already exist.
+# Skip the test if it does exist.
+temp_dir = "micropy_readonly_test_dir"
```
dpgeorge: suggest `vfs_posix_readonly_test_dir` (follows the name of this tests file) to match most of the other tests that do this

tests/extmod/vfs_posix_readonly.py
```diff
@@ -0,0 +1,99 @@
...
+try:
+    os.stat(temp_dir)
+    raise SystemExit("Target directory {} exists".format(temp_dir))
```
dpgeorge, line 16: other tests just print "SKIP" here and `raise SystemExit`; suggest following those

extmod/vfs_posix.c
jepler: will do

extmod/vfs_posix.c
jepler: OK

tests/extmod/vfs_posix_readonly.py
jepler: I saw this convention in other tests but I'm not sure how I feel about it.

It feels like "a directory happens to exist in the filesystem that prevents a test from running" is an error, not a skip-this-test situation. For instance, if I'm not keeping an eye on things, a botched run of this test that leaves the temporary directory will just SKIP and not actually test anything.

If you're happy with the way other tests do it, I of course can change it.

dpgeorge: > It feels like "a directory happens to exist in the filesystem that prevents a test from running" is an error, not a skip-this-test situation. For instance, if I'm not keeping an eye on things, a botched run of this test that leaves the temporary directory will just SKIP and not actually test anything.

Yes, you're right.  It is easy to accidentally have tests SKIP when you actually want them to run, see eg c1c73d966ea2d3ee20dbde8c5cd6ea5bd99bc94f

So, let's leave it as you have it.

dpgeorge (approved):

dpgeorge:
Rebased and merged in d79000df70d2fe50a7912da7f52813ad00584b3c through 59ee59901b9223697cf96f5859fcbf49fe9d21f4

PR #17716, opened 2025-07-19
objint_mpz: Fix pow3(1,2,0).

jepler:
### Summary

This finding is based on fuzzing micropython. I manually minimized the test case it provided.

### Testing

A fuzzer found a crash in 3-arg pow where the 3rd argument was zero. I added a test case for it, and fixed it.

I initially wanted to throw ZeroDivisionError and re-use the error string but I now throw ValueError with a string that matches CPython (3.11.2)

### DIFF
--- a/py/objint_mpz.c
+++ b/py/objint_mpz.c
@@ -356,9 +356,10 @@ static mpz_t *mp_mpz_for_int(mp_obj_t arg, mpz_t *temp) {
 mp_obj_t mp_obj_int_pow3(mp_obj_t base, mp_obj_t exponent,  mp_obj_t modulus) {
     if (!mp_obj_is_int(base) || !mp_obj_is_int(exponent) || !mp_obj_is_int(modulus)) {
         mp_raise_TypeError(MP_ERROR_TEXT("pow() with 3 arguments requires integers"));
+    } else if (modulus == MP_OBJ_NEW_SMALL_INT(0)) {
+        mp_raise_ValueError(MP_ERROR_TEXT("divide by zero"));
     } else {
-        mp_obj_t result = mp_obj_new_int_from_ull(0); // Use the _from_ull version as this forces an mpz int
-        mp_obj_int_t *res_p = (mp_obj_int_t *)MP_OBJ_TO_PTR(result);
+        mp_obj_int_t *res_p = mp_obj_int_new_mpz();
 
         mpz_t l_temp, r_temp, m_temp;
         mpz_t *lhs = mp_mpz_for_int(base,     &l_temp);
@@ -376,7 +377,7 @@ mp_obj_t mp_obj_int_pow3(mp_obj_t base, mp_obj_t exponent,  mp_obj_t modulus) {
         if (mod == &m_temp) {
             mpz_deinit(mod);
         }
-        return result;
+        return MP_OBJ_FROM_PTR(res_p);
     }
 }
 #endif
--- a/tests/basics/builtin_pow3_intbig.py
+++ b/tests/basics/builtin_pow3_intbig.py
@@ -20,3 +20,7 @@ print(hex(pow(2, x-1, x))) # Should be 1, since x is prime
 print(hex(pow(y, x-1, x))) # Should be 1, since x is prime
 print(hex(pow(y, y-1, x))) # Should be a 'big value'
 print(hex(pow(y, y-1, y))) # Should be a 'big value'
+try:
+    print(pow(1,2,0))
+except ValueError:
+    print("ValueError")


py/objint_mpz.c
```diff
@@ -364,6 +364,9 @@ mp_obj_t mp_obj_int_pow3(mp_obj_t base, mp_obj_t exponent,  mp_obj_t modulus) {
         mpz_t *lhs = mp_mpz_for_int(base,     &l_temp);
         mpz_t *rhs = mp_mpz_for_int(exponent, &r_temp);
         mpz_t *mod = mp_mpz_for_int(modulus,  &m_temp);
+        if (mpz_is_zero(mod)) {
+            mp_raise_msg(&mp_type_ZeroDivisionError, MP_ERROR_TEXT("divide by zero"));
```
AJMansfield: If CPython raises a `ValueError` here, why `mp_raise_msg(&mp_type_ZeroDivisionError` instead of just `mp_raise_ValueError(`?

py/objint_mpz.c
```diff
@@ -364,6 +364,9 @@ mp_obj_t mp_obj_int_pow3(mp_obj_t base, mp_obj_t exponent,  mp_obj_t modulus) {
         mpz_t *lhs = mp_mpz_for_int(base,     &l_temp);
         mpz_t *rhs = mp_mpz_for_int(exponent, &r_temp);
         mpz_t *mod = mp_mpz_for_int(modulus,  &m_temp);
+        if (mpz_is_zero(mod)) {
```
AJMansfield: This should be checked right after the typeerror check, rather than after it (potentially) already allocated memory for the mpz buffers for its arguments.

Note that integer 0 is never a concrete object when represented as a `mp_obj_t`, it's always a small int --- so the condition to check for zero is just `mp_obj_is_small_int(modulus) && MP_OBJ_SMALL_INT_VALUE(modulus) == 0`

AJMansfield (changes requested):

py/objint_mpz.c
```diff
@@ -357,8 +357,10 @@ mp_obj_t mp_obj_int_pow3(mp_obj_t base, mp_obj_t exponent,  mp_obj_t modulus) {
     if (!mp_obj_is_int(base) || !mp_obj_is_int(exponent) || !mp_obj_is_int(modulus)) {
         mp_raise_TypeError(MP_ERROR_TEXT("pow() with 3 arguments requires integers"));
     } else {
-        mp_obj_t result = mp_obj_new_int_from_ull(0); // Use the _from_ull version as this forces an mpz int
-        mp_obj_int_t *res_p = (mp_obj_int_t *)MP_OBJ_TO_PTR(result);
+        if (modulus == MP_OBJ_NEW_SMALL_INT(0)) {
```
AJMansfield: De-nest this so it's parallel as an `else if` with the line above.

py/objint_mpz.c
jepler: I changed it, including introducing a new MP_ERROR_TEXT.

py/objint_mpz.c
jepler: Makes sense, done. 

Thinking about this led to the discovery of #17730.

py/objint_mpz.c
```diff
@@ -357,8 +357,10 @@ mp_obj_t mp_obj_int_pow3(mp_obj_t base, mp_obj_t exponent,  mp_obj_t modulus) {
...
+            mp_raise_msg(&mp_type_ValueError, MP_ERROR_TEXT("pow() 3rd argument cannot be 0"));
```
AJMansfield: Use `mp_raise_ValueError` for smaller code size, that's the whole reason those shortcut raise methods exist.

AJMansfield (changes requested):

py/objint_mpz.c
jepler: done.

py/objint_mpz.c
jepler: done, thanks.

AJMansfield (approved):
Looks good! The tests seem adequate for the change, and I've confirmed sensitivity with unix-standard: tests correctly fail on https://github.com/micropython/micropython/commit/b12af7ddd69b2a88426ba5c6a23ac7c037c34a27 and pass on https://github.com/micropython/micropython/commit/902411afd5c184750bdaabfe5d3733ec44594227.

Unless someone else has a concern, this PR is ready to merge from f7d26f1.

jepler:
Something must be wrong with the change, rv32 failed (timeout) specifically on basics_builtin_pow3_intbig. I can't figure why, and I failed at performing the rv32 build locally on my debian bookworm system to reproduce it here.

dpgeorge:
I ran the rv32 tests locally with this PR, and did not get any failure.  That `basics_builtin_pow3_intbig` test takes only a few seconds to run.  I'm using newlib though, so maybe that's the difference?

AJMansfield:
I'll see if I can corroborate whether or not it works on my Pico2 in Hazard3 mode.

AJMansfield:
I've run the test suite against my Pico2 running in Hazard3 mode and can confirm that `basics/builtin_pow3_intbig.py` passes, along with the rest of the test suite (with the exception of some unrelated extmod/machine and other feature-related failures).

AJMansfield:
Can confirm that the tests pass on my Pico2 running in Hazard3 mode, not sure what the cause for the failure in qemu rv32 might be.

dpgeorge:
I suggest rebasing this on latest master to see if it fixes the CI.

jepler:
Done.

dpgeorge:
CI now passing.

py/objint_mpz.c
```diff
@@ -356,9 +356,10 @@ static mpz_t *mp_mpz_for_int(mp_obj_t arg, mpz_t *temp) {
 mp_obj_t mp_obj_int_pow3(mp_obj_t base, mp_obj_t exponent,  mp_obj_t modulus) {
     if (!mp_obj_is_int(base) || !mp_obj_is_int(exponent) || !mp_obj_is_int(modulus)) {
         mp_raise_TypeError(MP_ERROR_TEXT("pow() with 3 arguments requires integers"));
+    } else if (modulus == MP_OBJ_NEW_SMALL_INT(0)) {
+        mp_raise_ValueError(MP_ERROR_TEXT("pow() 3rd argument cannot be 0"));
```
dpgeorge: This new error string adds quite a bit of code size, for arguably not much gain.

Suggest making it smaller, maybe reuse the divide-by-zero message `"divide by zero"` but keep it as a ValueError?

jepler: done.
```>>> 1/0
ZeroDivisionError: divide by zero
>>> pow(1,1,0)
ValueError: divide by zero
```

dpgeorge: Thanks.  That's a much more acceptable code-size increase now.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 062e82a7cd60165db7edd8ee32105ea19a31ae2f

PR #13360, opened 2024-01-05
embed: Fix alloca include for FreeBSD and NetBSD.

darcagn:
When building the embedded port on FreeBSD, I receive the following error:
```
fatal error: alloca.h: No such file or directory
   36 | #include <alloca.h>
      |          ^~~~~~~~~~
```
FreeBSD/NetBSD doesn't include `alloca.h`, but `alloca()` is provided via `stdlib.h` instead. This PR allows it to build properly on those platforms.

PR #17717, opened 2025-07-19
`embed`: Fix `alloca` include for Windows.

sizious:
### Summary

When building the embedded port on MinGW-w64, I receive the following error:

```
fatal error: alloca.h: No such file or directory
```

MinGW-w64 (used on [MSYS2](https://www.msys2.org/)) doesn't include `alloca.h`, but `alloca()` is provided via `malloc.h` instead. This PR allows the build on this system. 

This fix needs be applied on other Windows build systems.

**Note:** A similar problem affected BSD in the past (see: https://github.com/micropython/micropython/pull/13360).

### Testing

We are using MicroPython through [KallistiOS](https://github.com/KallistiOS/), a library for developing on Sega Dreamcast, and more specifically a port in KallistiOS ([kos-ports](https://github.com/KallistiOS/kos-ports/)). Previously, it was impossible to compile this port with MinGW-w64 (the foundation of the [DreamSDK](https://github.com/dreamsdk/) environment), thanks to this fix, compilation is now possible and the problem is fixed.

image

See: https://github.com/KallistiOS/kos-ports/issues/112

### DIFF
--- a/ports/embed/port/mpconfigport_common.h
+++ b/ports/embed/port/mpconfigport_common.h
@@ -34,8 +34,13 @@ typedef long mp_off_t;
 
 // Need to provide a declaration/definition of alloca()
 #if defined(__FreeBSD__) || defined(__NetBSD__)
+// BSD
 #include <stdlib.h>
+#elif defined(_WIN32)
+// Windows
+#include <malloc.h>
 #else
+// Other OS
 #include <alloca.h>
 #endif
 


ports/embed/port/mpconfigport_common.h
```diff
@@ -34,8 +34,13 @@ typedef long mp_off_t;
 
 // Need to provide a declaration/definition of alloca()
 #if defined(__FreeBSD__) || defined(__NetBSD__)
+// BSD
 #include <stdlib.h>
+#elif defined(_WIN32) && (defined(__MINGW64_VERSION_MAJOR) || defined(__MINGW32_VERSION_MAJOR))
+// MinGW-w64; others Windows compilers probably need this too
```
stinos: Other compilers do need this, e.g. ports/windows/mpconfigport.h also has it for this reason. So I'd leave only the `#elif defined(_WIN32)` bit and mention explicitly it is needed for `alloca()`.

sizious: Thank you! Good catch. I’ll update this as suggested.

sizious: @stinos, the file has been updated.

dpgeorge (approved):

PR #17723, opened 2025-07-20
map: Always hash the object in mp_map_lookup.

jepler:
### Summary

Fuzzing produced a crash in OrderedDict which reduced to the following:
```
from collections import OrderedDict
x = OrderedDict()
x[: 'a'] = 1
x['b'] = 2
```

This occurs because, as an optimization, the key is not hashed when searching an OrderedDict, so an unhashable key can be stored. Even worse, the slice in this case is from `build_slice_stack_allocated` so by the time the 2nd indexing operation is being performed, it's nonsense instead of a Python object.

### Testing

I added a test for this and ran the testsuite locally.

This might slightly lower the performance of OrderedDict objects because before this change they avoided hashing the keys at all. However, I'm not aware of a way to check whether an object is hashable short of actually hashing it.

### DIFF
--- a/py/map.c
+++ b/py/map.c
@@ -188,6 +188,16 @@ mp_map_elem_t *MICROPY_WRAP_MP_MAP_LOOKUP(mp_map_lookup)(mp_map_t * map, mp_obj_
         }
     }
 
+    // get hash of index, with fast path for common case of qstr
+    // The object is hashed regardless of whether the map is ordered or not,
+    // because performing the hash is also performing the "is object hashable" check.
+    mp_uint_t hash;
+    if (mp_obj_is_qstr(index)) {
+        hash = qstr_hash(MP_OBJ_QSTR_VALUE(index));
+    } else {
+        hash = MP_OBJ_SMALL_INT_VALUE(mp_unary_op(MP_UNARY_OP_HASH, index));
+    }
+
     // if the map is an ordered array then we must do a brute force linear search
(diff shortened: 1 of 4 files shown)


jepler:
The test fails on some platforms [because Python 3.12 made slice objects hashable](https://docs.python.org/3/library/functions.html#slice).

```py
>>> d = collections.OrderedDict()
>>> d[:] = "oops"
>>> d
OrderedDict({slice(None, None, None): 'oops'})
```

jepler:
I pulled the new test into a separate file with a .exp included so that it can run regardless of Python version.

dpgeorge:
> This might slightly lower the performance of OrderedDict objects because before this change they avoided hashing the keys at all. However, I'm not aware of a way to check whether an object is hashable short of actually hashing it.

This change will most likely impact performance quite a lot.  That's because all built-in dicts (method tables and module dicts) have `is_ordered = 1`, so are essentially an `OrderedDict`.  See `py/obj.h:MP_DEFINE_CONST_DICT`.

So we should at least measure the impact of this change on performance.  And then probably need to find another way to fix this issue.

jepler:
That makes sense. Maybe the alternative is to add a specific check for slice objects. It's the (new?) temporary nature of the slice object which turns it from a "well don't do that then" bug into a crashing bug.

dpgeorge:
See #17736 for an alternative "fix".

PR #17729, opened 2025-07-21
webassembly: Fix binding of self to JavaScript methods.

dpgeorge:
### Summary

Fixes a bug in the binding of self/this to JavaScript methods.

The new semantics match Pyodide's behaviour, at least for the included tests.

### Testing

A test has been added for the case that is fixed.


dpgeorge:
Add test for `ar.push.call(ar, 3)` and `ar.push.call(ar2, 3)`.

dpgeorge:
@ntoll @WebReflection FYI

WebReflection:
I had a look and if the test works fine then I think it's OK if also these two work as expected https://github.com/micropython/micropython/pull/17729#issuecomment-3096544493 which I believe is the case, as it shouldn't be too different from borrowing methods via the `prototype` ... *however*, I am also not super sure why there are so many invariants for the same thing but in general, at least in JS, I would use the `apply` trap for functions or something like `__call__` trap in Python per each accessed reference, something to consider for the future?

In short, if you end up **always** with `fn.call(js_context or js_undefined, ...args)` a lot of code might be simplified but also with a proper trap there would be less guessing around and just natural consequences out of the meant code.

arrow functions and/or most global constructors would ignore the context, but in case of namespaces, context is ignored too, that is methods from `Math` or `JSON` or even `window` (or `globalThis`) so it feels like this dance might be reduced by far to me, but I let you decide if or when it'd be the right time to do so, also I need to find time myself to eventually contribute with a POC or something.

Again, if latest tests work, I think for the time being we're all good!

dpgeorge:
Thanks @WebReflection for the review.

I've now added those two extra tests and they pass.

> so it feels like this dance might be reduced by far to me, but I let you decide if or when it'd be the right time to do so, also I need to find time myself to eventually contribute with a POC or something.

If the code can be simplified so it doesn't have to select between `f(args)` and `f.call(args)` that would be great, and I'd change it.  But I just don't know how else to do it.

Anyway, the existing and new tests pass, so at least the semantics seem correct, even if they are not optimally implemented.

WebReflection:
please remember to update *npm* if you get a chance, thank you!

WebReflection:
@dpgeorge sorry for the ping but we'd really like to push these changes before vacations time ... by any chance these could land on *npm* ?

Thank you!

dpgeorge:
Now published here: https://www.npmjs.com/package/@micropython/micropython-webassembly-pyscript/v/1.26.0-preview-464

PR #17734, opened 2025-07-22
py/objint_longlong: Fix left shift of negative values.

projectgus:
### Summary

Closes #17702, thanks to @jepler for running these additional checks.

* The overflow check for 'longlong' representation left shift added in #16953 was wrong, and depended on UB. Rewrote to be the same as the small int left shift overflow check.
* Skip the viper "boundary" store tests if arbitrary integer support is disabled, these tests depend on generating very long integers and then truncating them.

Note: First version of this PR enabled UBSan for longlong tests, but this surfaced another issue so has been spun out to #17735.

### Testing

* Re-ran unix port (with UBSan), observed tests pass and no errors.

### Trade-offs and Alternatives

* The viper "boundary" store tests could probably be rewritten to also work withouth arbitrary precision integers, but relatively fiddly to do so.

PR #17735, opened 2025-07-22
ci: Enable UBSan for 'longlong' builds in CI, add stack size for sanitizer builds.

projectgus:
### Summary

* Spun out from #17734 - enable UBSan in CI as suggested by @jepler 
* Scale up the unix port stack sizes substantially when sanitizers are enabled. The size of some MicroPython C stack frames with UBSan and GCC 12 (as used in CI) appears to be quite high. Whatever causes this seems to be fixed in later GCC.
* To support uniform scaling of stack sizes, added some macros to apply the same stack size multiplier to both the main thread and the default thread stack size on unix port. Previously the unix port main thread stack was doubled on ARM builds but the default thread stack size was not, now they are both doubled (and the same macros are used when sanitizers are enabled).
* Added a custom MicroPython macro to mark if sanitizers are enabled. As far as I can tell there's no way to programmatically detect if sanitizers are enabled on gcc<14, the other macros are all Clang-only.
* Rewrote the scaling mechanism for the `stress/recurse_iternext.py` test, as this test fails on sanitizer builds with the increased stack. (Sanitizer build stack usage is only increased due to certain C function calls, but this test - I think - consumes stack in a way which isn't impacted by sanitizers so it can now handle many more iterations compared to a standard build). This took a while to find a good approach, but what's there seems more accurate than the previous heuristic and doesn't take noticeably more time to execute.

### Testing

* Tested in CI and by running CI functions locally with an ubuntu-22.04 container and with my local newer GCC.
* Re-ran the updated `stress/recurse_iternext.py` test on unix port with and without sanitizers, esp32 port with SPIRAM, and esp8266 port.

* Could leave UBSan disabled in CI for the 'longlong' variant, but (as demonstrated) it is useful to find bugs.

*This work was funded through GitHub Sponsors*

### DIFF
--- a/ports/unix/stack_size.h
+++ b/ports/unix/stack_size.h
@@ -1,0 +1,54 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 Angus Gratton
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+#ifndef MICROPY_INCLUDED_UNIX_STACK_SIZE_H
+#define MICROPY_INCLUDED_UNIX_STACK_SIZE_H
+
+#include "py/misc.h"
+
+// Define scaling factors for the stack size (also applies to main thread)
+#ifndef UNIX_STACK_MULTIPLIER
+
+#if defined(__arm__) && !defined(__thumb2__)
+// ARM (non-Thumb) architectures require more stack.
+#define UNIX_STACK_MUL_ARM 2
+#else
+#define UNIX_STACK_MUL_ARM 1
+#endif
+
+#if MP_SANITIZER_BUILD
+// Sanitizer features consume significant stack in some cases
+// This multiplier can probably be removed when using GCC 12 or newer.
+#define UNIX_STACK_MUL_SANITIZERS 4
+#else
+#define UNIX_STACK_MUL_SANITIZERS 1
+#endif
+
+// Double the stack size for 64-bit builds, plus additional scaling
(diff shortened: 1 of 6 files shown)


dpgeorge:
I'm confused why the stack changes are needed here:
1. For the main unix thread, its allocated stack is probably megabytes long (and might even automatically grow), so `MICROPY_STACK_CHECK_MARGIN` shouldn't make any difference for the main thread.
2. `ports/unix/mpthreadport.c` has:
```c
    // ensure there is enough stack to include a stack-overflow margin
    if (*stack_size < 2 * THREAD_STACK_OVERFLOW_MARGIN) {
        *stack_size = 2 * THREAD_STACK_OVERFLOW_MARGIN;
    }                                                        
```
which is 16k, so I don't think the change to `tests/thread/thread_stacksize1.py` is needed (is 16k really not enough?).
3. `ports/unix/mpthreadport.c` has:
```c
    // adjust stack_size to provide room to recover from hitting the limit
    *stack_size -= THREAD_STACK_OVERFLOW_MARGIN;
```
and that's already 8k, so I don't think `MICROPY_STACK_CHECK_MARGIN` is needed.


It's a bit hard to follow the logic of the stack limit for threads.  So maybe we should either get rid of `THREAD_STACK_OVERFLOW_MARGIN` and use `MICROPY_STACK_CHECK_MARGIN` instead.  Or just increase `THREAD_STACK_OVERFLOW_MARGIN` if it really needs to be increased for sanitizer builds.

projectgus:
> I'm confused why the stack changes are needed here:

This PR with the cherry-picked commit also hit "RuntimeError: maximum recursion depth exceeded" when testing longlong+UBSan, so you're right that it's not the right fix for this case.

> It's a bit hard to follow the logic of the stack limit for threads. So maybe we should either get rid of THREAD_STACK_OVERFLOW_MARGIN and use MICROPY_STACK_CHECK_MARGIN instead. Or just increase THREAD_STACK_OVERFLOW_MARGIN if it really needs to be increased for sanitizer builds.

Seems reasonable, I'll have a look along these lines.

dpgeorge:
> This PR with the cherry-picked commit also hit "RuntimeError: maximum recursion depth exceeded" when testing longlong+UBSan

So maybe there just needs to be more stack overall, rather than a change in stack margin?  Maybe `ports/unix/mpthreadport.c:mp_thread_create` just needs to have a larger minimum stack size for sanitizer builds?

projectgus:
@jepler Curious of your opinion on this PR, if you have any time to look?

dpgeorge:
I tested the new `recursive_iternext.py` test on PYBD_SF6, where it previously failed when used with the native emitter, and it now passes.

Also tested on PYBD_SF2, PYBV10, RPI_PICO.  It passes on all of those.  Very good!

py/misc.h
```diff
@@ -43,6 +43,11 @@ typedef unsigned int uint;
 #ifndef __has_builtin
 #define __has_builtin(x) (0)
 #endif
+#ifndef __has_feature
+// This macro is supported by Clang and gcc>=14
+#define __has_feature(x) (0)
```
jepler, line 48: gcc isn't keen on __has_feature. Their docs say
>  Note that ‘__has_feature’ and ‘__has_extension’ are not recommended for use in new code, and are only provided for compatibility with Clang.

That said, I don't see a better way to autodetect the sanitizers, though. (gcc has a predefined macro for asan but clang doesn't; neither defines a macro for ubsan)

Given that you defensively check whether has_feature is available and that this works in practice in clang and gcc-14 (I checked gcc-14) and obey the MP_UBSAN / MP_ASAN in preference to detection I think it's fine. There's also explicit documentation from clang that this is how to check for asan. The corresponding bit was missing from the ubsan documentation but it does work too (tested in gcc-14 and clang-19).

jepler (commented):
I'm in favor of it.

I looked into alternatives for detecting ubsan/asan and while gcc documentation shows they were grumpy at having to implement `__has_feature`, it does work as needed.

jepler (approved):

ports/unix/stack_size.h
```diff
@@ -0,0 +1,29 @@
+#ifndef MICROPY_INCLUDED_UNIX_STACK_SIZE_H
```
dpgeorge, line 26: Please add a copyright and license header.

dpgeorge:
I retested the revised `recursive_iternext.py` test on PYBD_SF2 and PYBD_SF6, in bytecode, mpy and native mode, and it passes.

Apart from the above comment about the license, this looks good to go in.

ports/unix/stack_size.h
projectgus: Oops, rebased and added!

dpgeorge (approved):
Thanks for updating.

PR #17736, opened 2025-07-22
py/obj: Add new type flag to indicate subscr accepts slice-on-stack.

dpgeorge:
### Summary

The recently merged PR #12518 allowed temporary slices to be allocated on the C stack, which is much better than allocating them on the GC heap.

Unfortunately there are cases where the C-allocated slice can escape and be retained as an object, which leads to crashes (because that object points to the C stack which now has other values on it).

See #17723 and #17733.

The fix here is to add a new `MP_TYPE_FLAG_SUBSCR_ALLOWS_STACK_SLICE`.  Native types should set this flag if their `subscr` method is guaranteed not to hold on to a reference of the slice object.

### Testing

The existing tests pass.  A new test is added to test the bugs reported in #17723 and #17733.

The function `mp_obj_is_native_type()` actually tests a type flag, so we replace the test of one flag with another flag.  Thus the fix here should be no change in code size and no change in performance.

We do use up another `MP_TYPE_FLAG_xxx` bit, but there are still 5 left.

Eventually more native types (eg tuple, list) can have this flag set, but I'll leave that for a follow-up PR.

### DIFF
--- a/py/obj.h
+++ b/py/obj.h
@@ -558,6 +558,8 @@ typedef mp_obj_t (*mp_fun_kw_t)(size_t n, const mp_obj_t *, mp_map_t *);
 // If MP_TYPE_FLAG_ITER_IS_STREAM is set then the type implicitly gets a "return self"
 //   getiter, and mp_stream_unbuffered_iter for iternext.
 // If MP_TYPE_FLAG_INSTANCE_TYPE is set then this is an instance type (i.e. defined in Python).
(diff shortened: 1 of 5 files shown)


jepler:
This makes sense and seems worth spending a flag bit on.

projectgus (approved):
Looks like a good fix to me, agree with jeff that it's worth spending a flag bit on this given the optimisation potential.

One minor thing, the test name has a typo (optimse instead of optimise).

dpgeorge:
> One minor thing, the test name has a typo (optimse instead of optimise).

Ah, thanks!  Now fixed.

PR #17739, opened 2025-07-22
py/objint_longlong: Fix overflow check in mp_obj_int_get_checked.

yoctopuce:
### Summary

This is to fix an outstanding TODO. The test cases is using a range expression, as this is available all builds, but `mp_obj_get_int` is used by various extmod where an overflow is more likely to occur.

### Testing

The code is pretty simple, and has been tested on unix/longlong varient.
An test case has been added in CI tests.

I am not sure that this exception is likely to occur in the MicroPython core code ,unless the Python code includes a serious bug. Therefore, the extra code size might not be worthwhile for everyone. However, there are many hardware extensions that use this function, and in those cases, the likelihood of an overflow is much higher.

### DIFF
--- a/tests/basics/int_64_basics.py
+++ b/tests/basics/int_64_basics.py
@@ -125,6 +125,22 @@ else:
 x = 1 << 62
 print('a' * (x + 4 - x))
 
+# test overflow check in mp_obj_get_int_maybe
+x = 1 << 32
+r = None
+try:
+    r = range(0, x)
+except OverflowError:
+    # 32-bit target, correctly handled the overflow of x
+    print("ok")
+if r is not None:
+    if len(r) == x:
+        # 64-bit target, everything is just a small-int
+        print("ok")
+    else:
+        # 32-bit target that did not handle the overflow of x
+        print("unhandled overflow")
+
 # negative shifts are invalid
 try:
     print((1 << 48) >> -4)
--- a/py/objint_longlong.c
+++ b/py/objint_longlong.c
@@ -308,8 +308,17 @@ mp_int_t mp_obj_int_get_truncated(mp_const_obj_t self_in) {
 }
 
 mp_int_t mp_obj_int_get_checked(mp_const_obj_t self_in) {
-    // TODO: Check overflow
-    return mp_obj_int_get_truncated(self_in);
+    if (mp_obj_is_small_int(self_in)) {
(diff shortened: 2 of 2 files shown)


yoctopuce:
@dpgeorge Can you help me understand what could cause zephir port to cause a syntax error when compiling `int_64_basics.py` ? 
I have the impression that the new code is super safe... Could it be a problem with constant folding of the expression `1 << 32` ?

dpgeorge:
> Can you help me understand what could cause zephir port to cause a syntax error when compiling `int_64_basics.py` ?

I think it's because the zephyr minimal build doesn't enable slicing, ie `a[:]`.

I suggest finding another way to test it, that doesn't use slicing.

yoctopuce:
Thank you, I forgot that it was even possible to make a build without slicing support...
I have updated the test code using a plain `range()`, that should do the trick.

tests/basics/int_64_basics.py
```diff
@@ -125,6 +125,16 @@
...
+# test overflow check in mp_obj_get_int_maybe
+x = 1 << 32
+try:
+    if len(range(0,x)) == x:
+        print('ok')
+    else:
+        print('unhandled overflow')
+except OverflowError as e:
+    print('ok')
+
```
dpgeorge, line 143: It might be worth adjusting this test to make it a bit clearer what is going on, eg:
```py
x = 1 << 32
r = None
try:
    r = range(0, x)
except OverflowError:
    # 32-bit target, correctly handled the overflow of x
    print("ok")
if r is not None:
    if len(r) == x:
        # 64-bit target, everything is just a small-int
        print("ok")
    else:
        # 32-bit target that did not handle the overflow of x
        print("unhandled overflow")
````

**Comment by yoctopuce** [comment]
Makes sense, this is much easier to understand. I just pushed the updated code.

**Comment by yoctopuce** [comment]
Note: the CI failure is unrelated to this PR:
```diff
--- /home/runner/work/micropython/micropython/tests/results/thread_thread_gc1.py.exp	2025-07-24 08:53:02.717730694 +0000
+++ /home/runner/work/micropython/micropython/tests/results/thread_thread_gc1.py.out	2025-07-24 08:53:02.717730694 +0000
@@ -1 +1,2 @@
-True
+qemu: uncaught target signal 11 (Segmentation fault) - core dumped
```

dpgeorge (approved):

PR #17743, opened 2025-07-22
stm32: Fix 12bit DAC issue on STM32H5.

yn386:
### Summary

The following code does not work with NUCLEO-H563ZI.
```python
import math
from array import array
from pyb import DAC

# create a buffer containing a sine-wave, using half-word samples
buf = array('H', 2048 + int(2047 * math.sin(2 * math.pi * i / 128)) for i in range(128))

# output the sine-wave at 400Hz
dac = DAC(2, bits=12)
dac.write_timed(buf, 400 * len(buf), mode=DAC.CIRCULAR)
```

For STM32H5, DMA parameter should set:
 * Actual DMA source datawidth to CTR1.
 * The length is the amount of data to be transferred from source to destiniation in bytes. to use 12bit DAC.

Also, this PR modifies the definition of `DMA_CIRCULAR` for STM32H5 to prevent conflict with data width specification.

### Testing

Tested code above with NUCLEO-H563ZI.
Also, DAC with 8bit still works after applying this PR.

The value of `DAC_CIRCULAR` is changed:
from
```shell
>>> hex(DAC.CIRTUCLAR)
'0x1'
```
to
```shell
>>> hex(DAC.CIRCULAR)
'0x20000000'
```

Indeed, it won't be a problem because users do not need to be aware of value of `DAC.CIRCULAR` in most cases.

### DIFF
--- a/ports/stm32/dac.c
+++ b/ports/stm32/dac.c
@@ -174,7 +174,7 @@ static void dac_start_dma(uint32_t dac_channel, const dma_descr_t *dma_descr, ui
         // For STM32G4, DAC registers have to be accessed by words (32-bit).
         dma_align = DMA_MDATAALIGN_BYTE | DMA_PDATAALIGN_WORD;
         #elif defined(STM32H5)
-        dma_align = 0;
+        dma_align = DMA_SRC_DATAWIDTH_BYTE | DMA_DEST_DATAWIDTH_WORD;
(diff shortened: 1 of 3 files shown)


dpgeorge (approved):
Thanks for the fix, this looks good to me (but I did not test it).

PR #17744, opened 2025-07-22
all: Go back to using default ruff quote style.

dpgeorge:
### Summary

Commit dc2fcfcc5511a371ff684f7d7772e7a7b479246d seems to have accidentally changed the ruff quote style to "preserve", instead of keeping it at the default which is "double".

Put it back to the default and update relevant .py files with this rule.

### Testing

This is a just a Python formatting change, so shouldn't need any tests.  But CI should anyway test the changes.


dpgeorge:
@cclauss FYI, did you have a reason for changing the quote style to "preserve"?

cclauss:
Sorry if I made an error.  The reason that I would use "preserve" is I say numerous quotes changes without it.  I prefer the default setting so I am happy to see this PR.

cclauss (approved):

dpgeorge:
@cclauss thanks for the approval.  I want to get this in now, because already it needs updating based on recent commits that had incorrect use of quotes.

PR #17746, opened 2025-07-22
rp2/modmachine: Do not use deprecated XOSC_MHZ.

langchr86:
### Summary

I replaced the crystal on a `W5500_EVB_PICO` board by a custom clock source which is not 12 MHz. Therefore I need to configure the SYS and USB PLL with custom dividers. The script `vcocalc.py` is used to calculate the needed values and to generate the compiler definitions that need to be added to cmake. 

But when using such custom defines the `XOSC_MHZ` is not defined and the build fails. 

### Testing

I tested this change while compiling for a `W5500_EVB_PICO` board with custom PLL settings as described above.

### DIFF
--- a/ports/rp2/clocks_extra.c
+++ b/ports/rp2/clocks_extra.c
@@ -83,8 +83,8 @@ void runtime_init_clocks_optional_usb(bool init_usb) {
     clock_configure(clk_ref,
         CLOCKS_CLK_REF_CTRL_SRC_VALUE_XOSC_CLKSRC,
         0,             // No aux mux
-        XOSC_KHZ * KHZ,
-        XOSC_KHZ * KHZ);
+        XOSC_HZ,
+        XOSC_HZ);
 
     /// \tag::configure_clk_sys[]
     // CLK SYS = PLL SYS (usually) 125MHz / 1 = 125MHz
--- a/ports/rp2/modmachine.c
+++ b/ports/rp2/modmachine.c
@@ -163,8 +163,6 @@ static void mp_machine_lightsleep(size_t n_args, const mp_obj_t *args) {
         }
     }
 
-    const uint32_t xosc_hz = XOSC_MHZ * 1000000;
-
     uint32_t my_interrupts = MICROPY_BEGIN_ATOMIC_SECTION();
     #if MICROPY_PY_NETWORK_CYW43
     if (cyw43_poll_is_pending()) {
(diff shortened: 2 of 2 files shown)


dpgeorge:
Thanks for the contribution.  This looks fine to me.

But note that `clocks_extra.c` uses `XOSC_KHZ`... so maybe that needs changing as well?  Why didn't that lead to a build error in your case?

Also, pico-sdk provides defines of `XOSC_MHZ` in terms of `XOSC_HZ` so I'm also uncertain why that define wasn't provided in your case?  What value is your `XOSC_HZ`?

langchr86:
My added definitions are the following:

~~~~~~
target_compile_definitions(${MICROPY_TARGET} PRIVATE
        XOSC_HZ=30720000

        PLL_SYS_REFDIV=4
        PLL_SYS_VCO_FREQ_HZ=960000000
        PLL_SYS_POSTDIV1=4
        PLL_SYS_POSTDIV2=2
        SYS_CLK_HZ=120000000

        PLL_USB_REFDIV=4
        PLL_USB_VCO_FREQ_HZ=1344000000
        PLL_USB_POSTDIV1=7
        PLL_USB_POSTDIV2=4
        USB_CLK_HZ=48000000
~~~~~~

There is code for backwards compatibility in `platform_defs.h`:

~~~~~~
// For backwards compatibility define XOSC_KHZ if the frequency is indeed an integer number of Khz.
#if defined(XOSC_HZ) && !defined(XOSC_KHZ) && (XOSC_HZ % 1000 == 0)
#define XOSC_KHZ (XOSC_HZ / 1000)
#endif

// For backwards compatibility define XOSC_MHZ if the frequency is indeed an integer number of Mhz.
#if defined(XOSC_KHZ) && !defined(XOSC_MHZ) && (XOSC_KHZ % 1000 == 0)
#define XOSC_MHZ (XOSC_KHZ / 1000)
#endif
~~~~~~

Because of the test `XOSC_KHZ % 1000 == 0` the `XOSC_MHZ` is not defined. 

The other point about `XOSC_KHZ` is true. I will later update my MR to also replace this in `clocks_extra.c`.

In theory we could/should also replace all usages of `SYS_CLK_KHZ`/`USB_CLK_KHZ` and the `MHZ` variants. The thing is, that those are not only used for `clock_configure()` (where they are converted to a `HZ` value) but also for `set_sys_clock_khz()` where we could replace the whole call with `set_sys_clock_hz()`. But I am not sure if this would have some side effects because of the comment in `check_sys_clock_khz()`:

> Note this impl is kept to preserve previous rounding behavior, vs calling check_sys_clock_hz

What is your suggestion?

dpgeorge:
> Because of the test `XOSC_KHZ % 1000 == 0` the `XOSC_MHZ` is not defined.

Ah, right, that makes sense then.  Your crystal is not an exact MHz multiple.  So you definitely need the change in this PR to get the correct timing.

> The other point about `XOSC_KHZ` is true. I will later update my MR to also replace this in `clocks_extra.c`.

Yes, please make this change.  Looking at pico-sdk, it seems they also updated their code to use `XOSC_HZ`, which `clocks_extra.c` is adapted from.

> In theory we could/should also replace all usages of `SYS_CLK_KHZ`/`USB_CLK_KHZ` and the `MHZ` variants.

No, let's leave those as they are for now.  Too much of a risk to change that just before a release.

langchr86:
I now also removed the usage of `XOSC_KHZ`. And I improved the change for `XOSC_MHZ` even more by removing the unneeded definition of the `xosc_hz` variable in `modmachine.c`.

I hope this matches your expectations. And thanks again for this great project!

dpgeorge (approved):
This looks good now, thanks!

Tested on RPI_PICO2_W, everything seems fine.

dpgeorge:
> And thanks again for this great project!

:smile:

PR #17751, opened 2025-07-23
unix/mpconfigport: Include time.h to get definition of time_t.

dpgeorge:
### Summary

For some reason macOS builds have just started failing, due to undefined `time_t` (related to df05caea6c6437a8b4756ec502a5e6210f4b6256).

Include "time.h" to fix that.

### Testing

CI will test this.

### DIFF
--- a/ports/unix/mpconfigport.h
+++ b/ports/unix/mpconfigport.h
@@ -29,6 +29,9 @@
 // features to work on Unix-like systems, see mpconfigvariant.h (and
 // mpconfigvariant_common.h) for feature enabling.
 
+// For time_t, needed by MICROPY_TIMESTAMP_IMPL_TIME_T.
+#include <time.h>
+
 // For size_t and ssize_t
 #include <unistd.h>
 


dpgeorge:
@yoctopuce FYI

yoctopuce:
> @yoctopuce FYI

Makes sense. I see that  `<time.h>` was included directly in the `.c` files present in `ports/unix`, hence the fact that the error did not come out earlier on Linux/windows builds.

dpgeorge:
I'm not sure if there's any better place to put this.  But since `typedef time_t mp_timestamp_t;` is in `py/mpconfig.h` it seems prudent to put `<time.h>` in `mpconfigport.h`.

yoctopuce:
I agree, and it shouldn't hurt since this is port specific.

dpgeorge:
Ah, this was probably due to 9b7d85227e67a7edd608aab4ff7eb4a838651f75 which included `mpconfig.h` in `mbedtls_config_common.h`.

PR #17754, opened 2025-07-23
core: Add MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC.

jepler:
And enable it on platforms where I am aware an efficient 32x32->64 bit multiply instruction exists.

### Summary

In the discussion of #17734 I became aware there was some existing use of the builtin overflow intrinsics, particularly for the longlong build.

This PR tests using it in place of `mp_small_int_mul_overflow`.

### Testing

I ran the testsuite locally (64-bit standard build). However, I don't know if the testsuite adequately checks multiplications "at the boundary" of the short integer range.

I also did some investigating and found a check for riscv, x86/x86_64, and arm that seems to capture the "is there a suitable multiply instruction". A check for xtensa is missing but could be beneficial.

I think there might be a modest performance benefit (avoiding multiple divisions per multiplication) but I did not attempt to measure it.

In an earlier iteration, the structure of int*int multiply in mp_binary_op ended up more complicated than needed, because I was still trying to use "small int overflow" in some places and "c type mul overflow" in others in a misguided attempt to keep the change minimal. That's no longer the case.

### DIFF
--- a/py/misc.h
+++ b/py/misc.h
@@ -35,7 +35,11 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stddef.h>
+#if __cplusplus // Required on at least one compiler to get ULLONG_MAX
+#include <climits>
+#else
 #include <limits.h>
+#endif
 
 typedef unsigned char byte;
 typedef unsigned int uint;
@@ -454,7 +458,7 @@ static inline uint32_t mp_clz_mpi(mp_int_t x) {
     #endif
 }
 
-// Overflow-checked operations for long long
+// Overflow-checked operations
 
 // Integer overflow builtins were added to GCC 5, but __has_builtin only in GCC 10
 //
@@ -462,45 +466,28 @@ static inline uint32_t mp_clz_mpi(mp_int_t x) {
 // functions below don't update the result if an overflow would occur (to avoid UB).
 #define MP_GCC_HAS_BUILTIN_OVERFLOW (__GNUC__ >= 5)
 
-#if __has_builtin(__builtin_umulll_overflow) || MP_GCC_HAS_BUILTIN_OVERFLOW
+#if MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC
+
 #define mp_mul_ull_overflow __builtin_umulll_overflow
+#define mp_mul_ll_overflow __builtin_smulll_overflow
+static inline bool mp_mul_mp_int_t_overflow(mp_int_t x, mp_int_t y, mp_int_t *res) {
+    // __builtin_mul_overflow is a type-generic function, this inline ensures the argument
+    // types are checked to match mp_int_t.
+    return __builtin_mul_overflow(x, y, res);
+}
+
 #else
-inline static bool mp_mul_ull_overflow(unsigned long long int x, unsigned long long int y, unsigned long long int *res) {
+
+bool mp_mul_ll_overflow(long long int x, long long int y, long long int *res);
+bool mp_mul_mp_int_t_overflow(mp_int_t x, mp_int_t y, mp_int_t *res);
+static inline bool mp_mul_ull_overflow(unsigned long long int x, unsigned long long int y, unsigned long long int *res) {
     if (y > 0 && x > (ULLONG_MAX / y)) {
         return true; // overflow
     }
     *res = x * y;
     return false;
 }
-#endif
 
-#if __has_builtin(__builtin_smulll_overflow) || MP_GCC_HAS_BUILTIN_OVERFLOW
-#define mp_mul_ll_overflow __builtin_smulll_overflow
-#else
-inline static bool mp_mul_ll_overflow(long long int x, long long int y, long long int *res) {
-    bool overflow;
-
-    // Check for multiply overflow; see CERT INT32-C
-    if (x > 0) { // x is positive
-        if (y > 0) { // x and y are positive
-            overflow = (x > (LLONG_MAX / y));
-        } else { // x positive, y nonpositive
-            overflow = (y < (LLONG_MIN / x));
-        } // x positive, y nonpositive
-    } else { // x is nonpositive
-        if (y > 0) { // x is nonpositive, y is positive
-            overflow = (x < (LLONG_MIN / y));
-        } else { // x and y are nonpositive
-            overflow = (x != 0 && y < (LLONG_MAX / x));
-        } // End if x and y are nonpositive
-    } // End if x is nonpositive
-
-    if (!overflow) {
-        *res = x * y;
-    }
-
-    return overflow;
-}
 #endif
 
 #if __has_builtin(__builtin_saddll_overflow) || MP_GCC_HAS_BUILTIN_OVERFLOW
--- a/py/mpconfig.h
+++ b/py/mpconfig.h
@@ -2336,4 +2336,23 @@ typedef time_t mp_timestamp_t;
 #define MP_WARN_CAT(x) (NULL)
 #endif
 
+// If true, use __builtin_mul_overflow (a gcc intrinsic supported by clang) for
+// overflow checking when multiplying two small ints. Otherwise, use a portable
+// algorithm.
+//
+// Most MCUs have a 32x32->64 bit multiply instruction, in which case the
+// intrinsic is likely to be faster and generate smaller code. The main exception is
+// cortex-m0 with __ARM_ARCH_ISA_THUMB == 1.
+//
+// The intrinsic is in GCC starting with version 5.
+#ifndef MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC
+#if defined(__ARM_ARCH_ISA_THUMB) && (__GNUC__ >= 5)
+#define MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC (__ARM_ARCH_ISA_THUMB >= 2)
+#elif (__GNUC__ >= 5)
+#define MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC (1)
+#else
+#define MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC (0)
+#endif
+#endif
+
 #endif // MICROPY_INCLUDED_PY_MPCONFIG_H
--- a/py/parsenum.c
+++ b/py/parsenum.c
@@ -28,6 +28,7 @@
 #include <stdlib.h>
 
 #include "py/runtime.h"
+#include "py/misc.h"
 #include "py/parsenumbase.h"
 #include "py/parsenum.h"
 #include "py/smallint.h"
@@ -52,7 +53,7 @@ static MP_NORETURN void raise_exc(mp_obj_t exc, mp_lexer_t *lex) {
 // to bigint parsing if supported)
 typedef mp_int_t parsed_int_t;
 
-#define PARSED_INT_MUL_OVERFLOW mp_small_int_mul_overflow
+#define PARSED_INT_MUL_OVERFLOW mp_mul_mp_int_t_overflow
 #define PARSED_INT_FITS MP_SMALL_INT_FITS
 #else
 // In the special case where bigint support is long long, we save code size by
--- a/py/runtime.c
+++ b/py/runtime.c
@@ -430,7 +430,7 @@ mp_obj_t MICROPY_WRAP_MP_BINARY_OP(mp_binary_op)(mp_binary_op_t op, mp_obj_t lhs
             // Operations that can overflow:
             //      +       result always fits in mp_int_t, then handled by SMALL_INT check
             //      -       result always fits in mp_int_t, then handled by SMALL_INT check
-            //      *       checked explicitly
+            //      *       checked explicitly for fit in mp_int_t, then handled by SMALL_INT check
             //      /       if lhs=MIN and rhs=-1; result always fits in mp_int_t, then handled by SMALL_INT check
             //      %       if lhs=MIN and rhs=-1; result always fits in mp_int_t, then handled by SMALL_INT check
             //      <<      checked explicitly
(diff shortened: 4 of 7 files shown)


jepler:
I'm surprised the code size on rv32 is unchanged. I must have the wrong preprocessor check. Locally I found that `__riscv_m` was defined for riscv CPUs that have the "m" (multiply) instruction set extension. However, my compiler is broken and can't actually link an executable, so I don't trust it very far. (`riscv64-unknown-elf-gcc (12.2.0-14+11+b1) 12.2.0` from debian stable, different from both ubuntu 22.04 and 24.04)

rp2 (rp2040) is expected to not change; the new code path is not enabled on Cortex M0 CPUs.

dpgeorge:
That's a very nice code size decrease!

(I have a MicroPython project running on a very small MCU which has run out of space, even using LTO, and I'll definitely be applying this patch to it.)

jepler:
Any suggestion how to structure this change better?

py/mpconfig.h
```diff
@@ -2324,4 +2324,24 @@ typedef time_t mp_timestamp_t;
 #define MP_WARN_CAT(x) (NULL)
 #endif
 
+// If true, use __builtin_mul_overflow (a gcc intrinsic supported by clang) for
+// overflow checking when multiplying two small ints. Otherwise, use a portable
+// algorithm.
+//
+// Most MCUs have a with a 32x32->64 bit multiply instruction, in which case the
```
dlech:
```suggestion
// Most MCUs have a 32x32->64 bit multiply instruction, in which case the
```

Or perhaps was missing a word?

py/parsenum.c
```diff
@@ -52,7 +53,11 @@ static MP_NORETURN void raise_exc(mp_obj_t exc, mp_lexer_t *lex) {
 // to bigint parsing if supported)
 typedef mp_int_t parsed_int_t;
 
-#define PARSED_INT_MUL_OVERFLOW mp_small_int_mul_overflow
+#if MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC
+#define PARSED_INT_MUL_OVERFLOW __builtin_mul_overflow
```
dlech: Why not always use `mp_mul_mp_int_t_overflow` here since we have the static inline function when `MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC` is true.

py/misc.h
```diff
@@ -454,16 +454,36 @@ static inline uint32_t mp_clz_mpi(mp_int_t x) {
...
+#define ULLONG_MAX (~0ULL)
+#endif
+
+#ifndef LLONG_MAX
+#define LLONG_MAX ((long long)(ULLONG_MAX >> 1))
+#endif
+
+#ifndef LLONG_MIN
+#define LLONG_MIN (-LLONG_MAX - 1)
+#endif
```
dlech: Seems like an unrelated change that deserves a separate commit.

jepler: I can break it out as its own commit if you prefer. It became necessary in this PR, it's not needless.

It became necessary there because the header itself uses the LLONG_MAX and related macros for mp_mul_ll_overflow. 

The following error occurs during the rp2 build without it (this is in my fork, which is named circuitpython):
```
In file included from /home/runner/work/circuitpython/circuitpython/micropython repo/py/mpstate.h:33,
                 from /home/runner/work/circuitpython/circuitpython/micropython repo/py/runtime.h:29,
                 from /home/runner/work/circuitpython/circuitpython/micropython repo/examples/usercmodule/cppexample/examplemodule.h:2,
                 from /home/runner/work/circuitpython/circuitpython/micropython repo/examples/usercmodule/cppexample/example.cpp:2:
/home/runner/work/circuitpython/circuitpython/micropython repo/py/misc.h: In function 'bool mp_mul_ull_overflow(long long unsigned int, long long unsigned int, long long unsigned int*)':
/home/runner/work/circuitpython/circuitpython/micropython repo/py/misc.h:475:23: error: 'ULLONG_MAX' was not declared in this scope
  475 |     if (y > 0 && x > (ULLONG_MAX / y)) {
      |                       ^~~~~~~~~~
/home/runner/work/circuitpython/circuitpython/micropython repo/py/misc.h:346:1: note: 'ULLONG_MAX' is defined in header '<climits>'; did you forget to '#include <climits>'?
  345 |     #include "genhdr/compressed.data.h"
  +++ |+#include <climits>
  346 | 
make[3]: *** [CMakeFiles/firmware.dir/build.make:6761: CMakeFiles/firmware.dir/home/runner/work/circuitpython/circuitpython/micropython_repo/examples/usercmodule/cppexample/example.cpp.o] Error 1
```
It's a bit of a surprise since LLONG_MIN and friends appeared before this commit in this header, but they were in blocks that were compiled out via MP_GCC_HAS_BUILTIN_OVERFLOW.

On M0(+) MCUs the non-inline/intrinsic version of mul_overflow is preferable when optimizing for size because the intrinsic expands to a lengthy sequence. The open-coded version is smaller in code size than the intrinsic, assuming __aeabi_ldivmod is already brought in to perform division of python long ints. The same is true of the 32 bit variant. godbolt: https://godbolt.org/z/c391ebfcv

py/parsenum.c
jepler: I think you're right, this can be simplified.

py/misc.h
jepler: this brings up the fact that mp_mul_ll_overflow should probably be moved out of line, because it's used 3 times in objint_longlong and as noted above it is still a pretty lengthy function.

updated godbolt, with m0plus and m33 builds: https://godbolt.org/z/s8oj7qTjG

so actually on m33 if we're optimizing for size we still want the division based long long version, and the intrinsic only for int/long.

jepler:
For future folks wondering about why the unsigned long long maximum constant is written the way it is:
```
In file included from /home/runner/work/micropython/micropython/micropython repo/py/mpstate.h:33,
                 from /home/runner/work/micropython/micropython/micropython repo/py/runtime.h:29,
                 from /home/runner/work/micropython/micropython/micropython repo/examples/usercmodule/cppexample/examplemodule.h:2,
                 from /home/runner/work/micropython/micropython/micropython repo/examples/usercmodule/cppexample/example.cpp:2:
/home/runner/work/micropython/micropython/micropython repo/py/misc.h: In function 'bool mp_mul_ull_overflow(long long unsigned int, long long unsigned int, long long unsigned int*)':
/home/runner/work/micropython/micropython/micropython repo/py/misc.h:482:23: error: 'ULLONG_MAX' was not declared in this scope
  482 |     if (y > 0 && x > (ULLONG_MAX / y)) {
      |                       ^~~~~~~~~~
/home/runner/work/micropython/micropython/micropython repo/py/misc.h:346:1: note: 'ULLONG_MAX' is defined in header '<climits>'; did you forget to '#include <climits>'?
  345 |     #include "genhdr/compressed.data.h"
  +++ |+#include <climits>
  346 | 
make[3]: *** [CMakeFiles/firmware.dir/build.make:6776: CMakeFiles/firmware.dir/home/runner/work/micropython/micropython/micropython_repo/examples/usercmodule/cppexample/example.cpp.o] Error 1
```

py/misc.h
```diff
@@ -454,53 +458,38 @@ static inline uint32_t mp_clz_mpi(mp_int_t x) {
...
+    // types are checked to match mp_int_t.
+    return __builtin_mul_overflow(x, y, res);
+}
+
 #else
+
+extern bool mp_mul_ll_overflow(long long int x, long long int y, long long int *res);
+extern bool mp_mul_mp_int_t_overflow(mp_int_t x, mp_int_t y, mp_int_t *res);
+
+// only called once, so inline it
```
dpgeorge: I'd remove this comment because:
1. we don't comment other `static inline` functions as such
2. in the future code may call it more than once
3. if someone makes a change where this function is called multiple times, they can decide at that time if it's then better to make this non-inline to improve code size

py/misc.h
```diff
@@ -454,53 +458,38 @@ static inline uint32_t mp_clz_mpi(mp_int_t x) {
...
 // Note that the builtins has a defined result when overflow occurs, whereas the custom
 // functions below don't update the result if an overflow would occur (to avoid UB).
 #define MP_GCC_HAS_BUILTIN_OVERFLOW (__GNUC__ >= 5)
 
-#if __has_builtin(__builtin_umulll_overflow) || MP_GCC_HAS_BUILTIN_OVERFLOW
+#if MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC
+
 #define mp_mul_ull_overflow __builtin_umulll_overflow
+#define mp_mul_ll_overflow __builtin_smulll_overflow
+inline static bool mp_mul_mp_int_t_overflow(mp_int_t x, mp_int_t y, mp_int_t *res) {
```
dpgeorge: please use `static inline`

py/misc.h
```diff
@@ -454,53 +458,38 @@ static inline uint32_t mp_clz_mpi(mp_int_t x) {
...
+    // __builtin_mul_overflow is a type-generic function, this inline ensures the argument
+    // types are checked to match mp_int_t.
+    return __builtin_mul_overflow(x, y, res);
+}
+
 #else
+
+extern bool mp_mul_ll_overflow(long long int x, long long int y, long long int *res);
+extern bool mp_mul_mp_int_t_overflow(mp_int_t x, mp_int_t y, mp_int_t *res);
```
dpgeorge: please don't use `extern` for functions, it's not needed

py/runtime.c
```diff
@@ -490,30 +490,15 @@ mp_obj_t MICROPY_WRAP_MP_BINARY_OP(mp_binary_op)(mp_binary_op_t op, mp_obj_t lhs
                 case MP_BINARY_OP_MULTIPLY:
                 case MP_BINARY_OP_INPLACE_MULTIPLY: {
```
dpgeorge: I think you can remove this empty line, because this `case` statement is now quite small (and style will then match other `case` statements here).

py/runtime.c
```diff
@@ -490,30 +490,15 @@ mp_obj_t MICROPY_WRAP_MP_BINARY_OP(mp_binary_op)(mp_binary_op_t op, mp_obj_t lhs
...
-                        return mp_obj_new_int_from_ll(res);
-                    } else {
-                        // use standard precision
-                        lhs_val = (mp_int_t)res;
-                    }
-                    #endif
-
                     mp_int_t int_res;
-                    if (mp_small_int_mul_overflow(lhs_val, rhs_val, &int_res)) {
-                        // use higher precision
```
dpgeorge, line 510: I suggest leaving this comment here, it's still valid.

py/runtime.c
```diff
@@ -490,30 +490,15 @@ mp_obj_t MICROPY_WRAP_MP_BINARY_OP(mp_binary_op)(mp_binary_op_t op, mp_obj_t lhs
...
+                    if (mp_mul_mp_int_t_overflow(lhs_val, rhs_val, &int_res)) {
                         lhs = mp_obj_new_int_from_ll(lhs_val);
                         goto generic_binary_op;
                     } else {
-                        // use standard precision
-                        return MP_OBJ_NEW_SMALL_INT(int_res);
+                        lhs_val = int_res;
                     }
+
```
dpgeorge: You can remove this empty line.

py/runtime.c
```diff
@@ -490,30 +490,15 @@ mp_obj_t MICROPY_WRAP_MP_BINARY_OP(mp_binary_op)(mp_binary_op_t op, mp_obj_t lhs
...
+                    break; // result fits in mp_int_t but might not be MP_SMALL_INT_FITS
```
dpgeorge: Suggest putting this comment up in the `else` above.  Or just leave the existing "use standard precision" comment (because it is just using standard int*int precision).

See eg the `MP_BINARY_OP_LSHIFT` case to match styling of that.

py/runtime.c
```diff
@@ -1784,3 +1769,63 @@ MP_NORETURN void mp_raise_recursion_depth(void) {
     mp_raise_type_arg(&mp_type_RuntimeError, MP_OBJ_NEW_QSTR(MP_QSTR_maximum_space_recursion_space_depth_space_exceeded));
 }
 #endif
+
+#if !MICROPY_USE_GCC_MUL_OVERFLOW_INTRINSIC
+bool mp_mul_ll_overflow(long long int x, long long int y, long long int *res) {
```
dpgeorge: Maybe these functions can go in `py/runtime_utils.c`?

py/parsenum.c
dpgeorge: I think this can still be simplified, per the suggestion.

py/runtime.c
dpgeorge: I guess we could add `py/misc.c` for this (and maybe other such functions in the future), because they feel like lower-level / lower-layer functions than others in `runtime.c`/`runtime_utils.c`.  But I'm really not keen to add new source files if we don't need to.

dpgeorge:
> I am not happy with the _structure_ of how this ended up implemented, particularly for the `int*int` multiply in mp_binary_op. It's more complicated than I would like due to the fact that `mp_small_int_mul_overflow` also implicitly checks for SMALL_INT_FITS while ` __builtin_mul_overflow` just checks if the C type (e.g., mp_int_t) overflows. However, if/when tests pass & code size comes in smaller, it may be worth looking for a way to structure the change that's acceptable that still gets the size benefit.

@jepler do you still have reservations about how the `MP_BINARY_OP_MULTIPLY` case turned out?  I think it's actually very clean now, and matches how most of the other case's work with checking for overflow.

Also, I think both `MP_BINARY_OP_MULTIPLY` and `MP_BINARY_OP_POWER` might be more efficient now when computing an answer that overflows a small-int but stays within a `mp_int_t`: now it does the calculation using `mp_int_t` and only at the end converts that to a big-int object (previously it would abort the calculation and redo it using big int arithmetic).

jepler:
I think I got all the review comments addressed this time. I also updated the initial comment about the structure of the code, which was about an earlier version that tried to retain the "small int multiply overflow" and the new "mp_int_t overflow" cases, which was not needed.

Thank you!

dpgeorge (approved):
Thanks for updating.  The new `MP_BINARY_OP_MULTIPLY` case is now very clean and matches other cases.  Very good!

dpgeorge:
Rebased and merged in a80913292153a14424b29bdb9ca8847e8d35cf73

PR #17737, opened 2025-07-22
tools/ci.sh: Test building all natmod examples with all ARM-M archs.

dpgeorge:
### Summary

This adds a few more tests to CI, to build all natmod examples with all possible ARM-M architectures.

### Testing

To be tested by CI.

PR #17757, opened 2025-07-24
qemu/arm: Extend CI testing to include ArmV7-M hardfp targets.

agatti:
### Summary

As a followup to #17737, this PR introduces a QEMU Arm target that has a hardware floating point unit (`MPS2_AN500`), and then lets CI use it for selected operations.

`MPS2_AN500` is a Cortex-M7 with a single- and double-precision floating point unit, but it is otherwise equivalent to the `MPS2_AN385` target for MicroPython's needs.

To further validate that the FPU is used and initialised correctly, the machine exception handler for ArmV7-M targets was extended to dump the full set of debug registers.  If the FPU is used incorrectly (as in, access to the appropriate machine coprocessor was not enabled beforehand) then the `CFSR.NOCP` bit should appear to be set in the exception handler dump.

In order to correctly test the hardfp natmods we need to have a validated hardfp interpreter, so we must have a full `MPS2_AN500` build and test cycle as well.  To reduce the impact on CI times I've split the original CI step into two discrete units, as in "ci_qemu_build_arm_thumb_softfp" and "ci_qemu_build_arm_thumb_hardfp", so they can run in parallel whenever possible.

Manual building and testing of the QEMU port isn't otherwise affected, the default Arm machine is still `MPS2_AN385` and it will retain the same floating point configuration as before (`MICROPY_FLOAT_IMPL_FLOAT`).

Concidentally, I've made QEMU's Makefile a bit more flexible when it comes to floating point settings.  I've replicated the setup from the `stm32` port, so the same flag names apply here as well.

### Testing

Besides manual testing, CI should (hopefully) run the relevant step without errors; tests passed on my branch.

There will be a (hopefully) minor increase in CI build times.  However, having two separate targets is necessary to catch issues with FPU opcodes being used in softfp contexts for example.  Making the hardfp variant the default would still call for a softfp variant to be there anyway.

Things could still be improved though, as in ArmV6-M builds could be tested in an emulated Cortex-M0+ machine to make sure no ArmV7-M opcodes are used by GCC or by viper/native emitters.


dpgeorge:
This all looks really good to me, a nice extension of the existing qemu port and tests!

But, building the new board locally, I run into a strange problem:
```
$ cd ports/qemu
$ make BOARD=MPS2_AN500
...
CC ../../lib/libm_dbl/trunc.c
CC ../../lib/libm_dbl/thumb_vfp_sqrt.c
/usr/lib/gcc/arm-none-eabi/14.2.0/../../../../arm-none-eabi/bin/ld: section .ARM.exidx.text.__udivmoddi4 LMA [0002f940,0002f947] overlaps section .data LMA [0002f940,0002f94f]
make: *** [Makefile:232: build-MPS2_AN500/firmware.elf] Error 1
```
I'm using Arch Linux with arm-none-eabi-gcc 14.2.0.  Any idea what to do about that?

agatti:
I'm on Arch as well - I've just installed the arm-none-eabi-{binutils,gcc,newlib} packages and attempted to build that outside my usual VMs... and it works:

```bash
~/src/micropython/ports/qemu $ rm -rf build-* ; make BOARD=MPS2_AN500 -j4
mkdir -p build-MPS2_AN500/extmod/
[...]
CC ../../lib/libm_dbl/trunc.c
CC ../../lib/libm_dbl/thumb_vfp_sqrt.c
CC build-MPS2_AN500/frozen_content.c
/usr/lib/gcc/arm-none-eabi/14.2.0/../../../../arm-none-eabi/bin/ld: warning: build-MPS2_AN500/firmware.elf has a LOAD segment with RWX permissions
 193200      16  144156  337372   525dc build-MPS2_AN500/firmware.elf
```

For the record, my `arm-none-eabi-binutils` package is version 2.43-1, and `arm-none-eabi-newlib` is 4.5.0.20241231-1.

Can you please try to apply this diff and see if it works on your machine without reinstalling/updating your Arm toolchain?

```diff
diff --git a/ports/qemu/mcu/arm/mps2.ld b/ports/qemu/mcu/arm/mps2.ld
index dffefa3b0..1597a7de3 100644
--- a/ports/qemu/mcu/arm/mps2.ld
+++ b/ports/qemu/mcu/arm/mps2.ld
@@ -19,7 +19,15 @@ SECTIONS
         *(.rodata*)
         . = ALIGN(4);
         _etext = .;
-        _sidata = _etext;
+    } > RAM
+
+    .ARM.exidx : AT ( _etext ) {
+        . = ALIGN(4);
+        __exidx_start = .;
+        *(.ARM.exidx*)
+        *(.gnu.linkonce.armexidx.*)
+        __exidx_end = .;
+        _sidata = __exidx_end;
     } > RAM
 
     .data : AT ( _sidata )
```

I can still build and run things on Arch with that linkerscript change, if it doesn't cause issues under Ubuntu and it works on your machine too I'll update the PR with it.

What's happening is that for some reason on Arch's GCC14 (and probably later versions too) hardfp math libraries depend on exception handling.  What that diff does is adding a section to hold exception unwind tables data.  `__exidx_start` and `__exidx_end` aren't strictly needed for the compiler version in question, but I've found reports saying that just the section being present doesn't cut it and those two symbols are needed by something else inside libgcc/libc/libm/whatever.

Still, I can build and run tests with that diff applied on all machines/environments I have available, but if it doesn't fix the issue on your machine it's all moot :)

dpgeorge:
> For the record, my `arm-none-eabi-binutils` package is version 2.43-1, and `arm-none-eabi-newlib` is 4.5.0.20241231-1.

I have exactly the same versions.

And I just found that a fresh checkout of this PR -- WITHOUT REBASING -- builds fine.  It's after a rebase on latest master that it doesn't link.  I suspect you'll also see the problem after rebasing (and it's maybe because of changes to floating point constant folding, bringing in some additional libm code?).

Applying the above exidx patch gets it linking after a rebase.

dpgeorge:
Note that `tests/run-tests.py` has the following:
```
    "qemu": (
        # Skip tests that require Cortex-M4.
        "inlineasm/thumb/asmfpaddsub.py",
        "inlineasm/thumb/asmfpcmp.py",
        "inlineasm/thumb/asmfpldrstr.py",
        "inlineasm/thumb/asmfpmuldiv.py",
        "inlineasm/thumb/asmfpsqrt.py",
    ),
```
and so these tests are skipped on the MPS2_AN500, even though they could run and pass on it.  Not sure if that's worth improving in this PR, or for a follow-up PR (you can decide).

agatti:
> And I just found that a fresh checkout of this PR -- WITHOUT REBASING -- builds fine. It's after a rebase on latest master that it doesn't link. I suspect you'll also see the problem after rebasing (and it's maybe because of changes to floating point constant folding, bringing in some additional libm code?).
> 
> Applying the above exidx patch gets it linking after a rebase.

Indeed I was able to reproduce (and fix) the issue after a rebase - thanks for the hint.  I've added a commit to the PR with the linkerscript changes addressing this issue.

> Note that `tests/run-tests.py` has the following:
> [...]
> and so these tests are skipped on the MPS2_AN500, even though they could run and pass on it. Not sure if that's worth improving in this PR, or for a follow-up PR (you can decide).

No worries, I've just updated the test runner to put those tests into the block list if the reported arch is neither `armv7emsp` nor `armv7emdp`.  I've tested it with `MPS2_AN385`, `MPS2_AN500`, and `SABRELITE` and the set of tests being executed matches what they are supposed to run.

dpgeorge (approved):
Looks good now thanks!

Tested locally on qemu/MPS2_AN500, and stm32/PYBV11.  Tests pass.

PR #17758, opened 2025-07-24
webassembly: reuse JavaScript proxy references if possible, to improve identity/equality relationships

dpgeorge:
### Summary

In 77bd8fe5b80b0e7e02cdb6b4272c401ae3dca638 PyProxy's were reused when the same Python object was proxied across to JavaScript.

This PR does the same thing but for JsProxy's going from JS to Python.  If an existing JsProxy reference exists for the JS object about to be proxied across, then it's reused.  This helps keep down the number of alive objects, and, more importantly, improves equality relationships.  Eg we now get, on the Python side:
```py
import js

print(js.Object == js.Object)
```
that prints `True`.  Previously it was `False`.

### Testing

Tests have been added to CI.  They pass locally.

This does not make identity work with `is`, eg `js.Object is js.Object` is actually `False`.  With more work that could be made `True` but for now we leave that as-is.  This matches Pyodide semantics (only == is True, while is is False).

This needs to maintain a separate `Map` on the JS side, of all the proxied JS objects.  But that's (a) necessary and (b) worth it because of the reduced churn of JS objects when proxying them across to Python.


dpgeorge:
@WebReflection FYI.  This improved object equality as we discussed.  See the tests for what improvements to expect.  Note that `is` (on the Python side) is still `False`, per Pyodide.

ports/webassembly/proxy_js.js
```diff
@@ -97,13 +98,19 @@ function proxy_js_check_existing(c_ref) {
 
 // js_obj cannot be undefined
 function proxy_js_add_obj(js_obj) {
+    // See if there is an existing JsProxy reference, and use that if there is.
+    if (proxy_js_ref_map.has(js_obj)) {
```
WebReflection: I know it's an ugly workaround but AFAIK there are no optimizations for `has` immediately followed by `get` in the JS *Map* world (runtimes) so that when performance is meant to be "best" and/or on the critical path, *JS* devs tend to abuse `undefined` returned when no entry is found.

That is:

```js
const known = proxy_js_ref_map.get(js_obj);
if (known) return known;
```

We're still talking *O(1)* with JS *Map* lookup for the key but while *O(1)* + *O(1)* is still *O(1)* in the theoretical side of *BigO* notation, you can skip that "*O(2)*" by reaching directly the value, if available, via `.get(key)`.

Not a blocker, of course, but years of perf tuning makes me always use that ugly *fast-path* whenever it's crucial.

- - -

**edit** for correctness and history sake ... the suggested improvements fails when:

  * a `key` in a *Map* is meant to be *falsy* or even `undefined` ... for whatever reason; IIRC this would never be the case in here
  * if you prefer strict `known !== undefined` in that `if` statement, that's OK, the mentioned perf boost won't be affected

dpgeorge: Thanks for the review.

I was wondering if there was a more efficient way to do this, and I've now changed it to what you suggested.

Note that `proxy_js_ref_map` contains integers and could potentially return 0, so we do need the `known !== undefined` check.

WebReflection: yes, that's the falsy value I was referring to ... if `0` could be returned, you definitively need that `!== undefined` check 👍

PR #17761, opened 2025-07-24
tests/micropython: Remove big ints dependence for viper boundary tests.

agatti:
### Summary

This PR provides an implementation for viper boundary tests that can work even without big int support.

Since it uses a fixed-size buffer to hold values to work with, this should work on any platform as long as its integers are at least 32 bits wide, regardless its configuration on how big integers can get.

Feel free to merge this whenever you see fit, to avoid unnecessary churn on the repo.

@projectgus Apologies for causing you extra work - all the targets I routinely test my Viper/native changes on already have big int support, so I didn't catch that :|

### Testing

Viper boundary tests were executed successfully on the Unix port using its `longlong` variant, running on a x64 Linux system.  CI should also take care of this.


dpgeorge (approved):
Thanks, this looks good.  It's good to make it run on more targets.

PR #17763, opened 2025-07-25
stm32: Add DAC feature for STM32G0.

yn386:
### Summary

This PR adds DAC feature for STM32G0.
`DAC.write()` and `DAC.write_timed()` is now available.

### Testing

The following code works with NUCLEO-G01BRE.
A sine wave is output via PA4 when ch=1, PA5 when ch=2.

```python
import math
from pyb import DAC

ch = 1

# create a buffer containing a sine-wave
buf = bytearray(100)
for i in range(len(buf)):
    buf[i] = 128 + int(127 * math.sin(2 * math.pi * i / len(buf)))

# output the sine-wave at 600Hz
dac = DAC(ch, bits=8)
dac.write_timed(buf, 600 * len(buf), mode=DAC.CIRCULAR)
```

```python
import math
from array import array
from pyb import DAC

ch = 2

# create a buffer containing a sine-wave, using half-word samples
buf = array('H', 2048 + int(2047 * math.sin(2 * math.pi * i / 128)) for i in range(128))

# output the sine-wave at 400Hz
dac = DAC(ch, bits=12)
dac.write_timed(buf, 400 * len(buf), mode=DAC.CIRCULAR)
```

This PR increases code size by 1800 Bytes.

Before:
```
   text	   data	    bss	    dec	    hex	filename
 297528	    108	   4004	 301640	  49a48	build-NUCLEO_G0B1RE/firmware.elf
```
After:
```
   text	   data	    bss	    dec	    hex	filename
 299312	    108	   4020	 303440	  4a150	build-NUCLEO_G0B1RE/firmware.elf
```

### DIFF
--- a/ports/stm32/boards/NUCLEO_G0B1RE/mpconfigboard.h
+++ b/ports/stm32/boards/NUCLEO_G0B1RE/mpconfigboard.h
@@ -5,7 +5,7 @@
 #define MICROPY_HW_HAS_FLASH        (1)
 #define MICROPY_HW_ENABLE_RNG       (0)
 #define MICROPY_HW_ENABLE_RTC       (1)
-#define MICROPY_HW_ENABLE_DAC       (0)
+#define MICROPY_HW_ENABLE_DAC       (1)
 #define MICROPY_HW_ENABLE_USB       (0) // can be enabled if USB cable connected to PA11/PA12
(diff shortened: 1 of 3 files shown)


dpgeorge (approved):
A nice addition, thank you!

PR #17766, opened 2025-07-25
stm32/boards: Add WeAct_H743VI board support.

yn386:
### Summary

This change adds WeAct STM32H743 Core Board support to the STM32 port.

WeAct STM32H743 Core Board:
https://github.com/WeActStudio/MiniSTM32H7xx

This board uses STM32H743VI:
https://www.st.com/en/microcontrollers-microprocessors/stm32h743vi.html

### Testing

Tested:
 * REPL via USB VCP works.
 * SPI flash is available (8MB) as an internal storage.
 * QSPI flash is available (8MB) as ROMFS.
 * USB Storage is available.
 * Some peripherals (GPIO, SPI, I2C, Timer) works but not all peripherals are not tested.
 * run-tests.py passed.

### Trade-offs and Alternatives

There is no negative impact because this change contains only adding board definition.

PR #17770, opened 2025-07-27
stm32: Fix DAC issue for MCUs those have D-Cache.

yn386:
### Summary

The following code does not work with NUCLEO-F767ZI.
Incorrect wave may be output on PA4. STM32H7 has same issue.
```python
import math
from pyb import DAC

# create a buffer containing a sine-wave
buf = bytearray(100)
for i in range(len(buf)):
    buf[i] = 128 + int(127 * math.sin(2 * math.pi * i / len(buf)))

# output the sine-wave at 600Hz
dac = DAC(1, bits=8)
dac.write_timed(buf, 600 * len(buf), mode=DAC.CIRCULAR)
```

For MCUs which has D-Cache (such as STM32F7, STM32H7), clean D-cache before starting DMA.
For more details, please refer follwing document:
https://www.st.com/resource/en/application_note/DM00272913.pdf

This PR cleans D-cache before starting DMA.

### Testing

Tested code above with:
* NUCLEO-F767ZI (STM32F767ZI)
* WeAct H743VI (STM32H743VI, PR: #17766) 

-> Both boards output wave as expected.

### DIFF
--- a/ports/stm32/dac.c
+++ b/ports/stm32/dac.c
@@ -485,6 +485,9 @@ mp_obj_t pyb_dac_write_timed(size_t n_args, const mp_obj_t *pos_args, mp_map_t *
     #endif
     }
 
+    // To prevent invalid dac output, clean D-cache before starting dma.
+    MP_HAL_CLEAN_DCACHE(bufinfo.buf, bufinfo.len);
+
     uint32_t align;
     if (self->bits == 8) {
         align = DAC_ALIGN_8B_R;


ports/stm32/dac.c
```diff
@@ -485,6 +485,11 @@ mp_obj_t pyb_dac_write_timed(size_t n_args, const mp_obj_t *pos_args, mp_map_t *
     #endif
     }
 
+    #if defined(STM32F7) || defined(STM32H7)
```
dpgeorge: You should be able to remove this if conditional.  The `MP_HAL_CLEAN_DCACHE` macro is defined to nothing on MCUs that don't have a D-cache.

yn386: Thank you for your comment.
Indeed, MP_HAL_CLEAN_DCACHE can use without condition with STM32F7/H7.
https://github.com/micropython/micropython/blob/c9b52b2b7f164745eda7210ccf6640fce3737f59/ports/stm32/mpconfigboard_common.h#L705-L716

I removed condition for STM32F7/H7.

dpgeorge (approved):

PR #17771, opened 2025-07-27
py/asmthumb: Don't corrupt base register in large offset store.

arachsys:
### Summary

On rp2, I discovered that viper stores offset from a register variable were clobbering that variable:
```python
@micropython.viper
def test():
  x = ptr8(ba)
  print(hex(uint(x)))
  x[0x1000] = 0
  # The register x has now been clobbered, incremented by 0x1000 - 31.
  # It gets corrupted only if the index/offset is >= 1 << 10.
  print(hex(uint(x)))
ba = bytearray(0x2000)
test()
```

@agatti kindly pointed me at `asm_thumb_store_reg_reg_offset()`. It turns out this modifies the base register when storing with a large offset which triggers the generic path. If a variable lives in that register, this corrupts it. This patch fixes this by using R2 (REG_TEMP2) instead.

The only VM opcode which uses REG_TEMP2 is `emit_native_rot_three()`, which only calls `ASM_STORE_REG_REG()` via `emit_access_stack()` with an offset of free = start + size - pos for the stack. This should never be large enough to trigger the generic path which clobbers REG_TEMP2.

The rv32 emitter uses REG_TEMP2 in `asm_rv32_emit_store_reg_reg_offset()` already, with the same justification: values of the free stack capacity are too small to trigger the affected large offset path.

However...

### Testing

...before extending the existing store boundary tests to check for unwanted modification of the base registers, I double-checked them against this commit. It turns out to break them!

```
FAILURE /home/root/mp/tests/results/micropython_viper_ptr32_store_boundary_intbig.py
--- /home/root/mp/tests/results/micropython_viper_ptr32_store_boundary_intbig.py.exp    2025-07-27 13:03:13.618624946 +0100
+++ /home/root/mp/tests/results/micropython_viper_ptr32_store_boundary_intbig.py.out    2025-07-27 13:03:13.618759975 +0100
@@ -17,4 +17,4 @@
 0x10111213 0x11121314 0x12131415
 0x13141516
 0x14151617
-0x15161718
+0x20012b54

FAILURE /home/root/mp/tests/results/micropython_viper_ptr16_store_boundary_intbig.py
--- /home/root/mp/tests/results/micropython_viper_ptr16_store_boundary_intbig.py.exp    2025-07-27 13:04:43.689570377 +0100
+++ /home/root/mp/tests/results/micropython_viper_ptr16_store_boundary_intbig.py.out    2025-07-27 13:04:43.689900357 +0100
@@ -17,4 +17,4 @@
 0x1213 0x1314 0x1415
 0x1516
 0x1617
-0x1718
+0x2ae2

FAILURE /home/root/mp/tests/results/micropython_viper_ptr8_store_boundary_intbig.py
--- /home/root/mp/tests/results/micropython_viper_ptr8_store_boundary_intbig.py.exp     2025-07-27 13:34:04.778117813 +0100
+++ /home/root/mp/tests/results/micropython_viper_ptr8_store_boundary_intbig.py.out     2025-07-27 13:34:04.778436804 +0100
@@ -17,4 +17,4 @@
 0x13 0x14 0x15
 0x16
 0x17
-0x18
+0x61
```

I prodded a bit further, and have the following simple repro of the problem:

```python
>>> b = bytearray(5000)
>>> @micropython.viper
... def test():
...     x = ptr8(b)
...     x[0x1000]=123
...
>>> test()
>>> b[4096]
17
>>> sum(b)
17
```

So maybe there's a problem with using R2 after all @agatti - it looks like the correct location always gets written, but with the wrong value? (Strangely, it *didn't* break my more complicated test case that I used as a first quick check. I must have got lucky.) I've left this as a draft for now while I stare at it a bit more!

### DIFF
--- a/py/asmthumb.c
+++ b/py/asmthumb.c
@@ -491,9 +491,11 @@ void asm_thumb_store_reg_reg_offset(asm_thumb_t *as, uint reg_src, uint reg_base
         // Can use T3 encoding
         asm_thumb_op32(as, (OP_LDR_STR_W_HI(operation_size, reg_base) | OP_STR_W), OP_LDR_STR_W_LO(reg_src, (offset << operation_size)));
     } else {
-        // Must use the generic sequence
-        asm_thumb_add_reg_reg_offset(as, reg_base, reg_base, offset - 31, operation_size);
-        asm_thumb_op16(as, ((OP_LDR_STR_TABLE[operation_size] | OP_STR) << 11) | (31 << 6) | (reg_base << 3) | reg_src);
+        // Must use the generic sequence. It is safe to use R2 (REG_TEMP2)
+        // here because that is only used by emit_native_rot_three(), which
+        // can't call us with a large enough offset to end up in this branch.
+        asm_thumb_add_reg_reg_offset(as, ASM_THUMB_REG_R2, reg_base, offset - 31, operation_size);
+        asm_thumb_op16(as, ((OP_LDR_STR_TABLE[operation_size] | OP_STR) << 11) | (31 << 6) | (ASM_THUMB_REG_R2 << 3) | reg_src);
     }
 }
 


agatti:
This is interesting!  I'll take a look at it right now.

arachsys:
> This is interesting! I'll take a look at it right now.

A more expensive version which does work, fixing the problem and passing those tests is:

```diff
diff --git a/py/asmthumb.c b/py/asmthumb.c
index 18c3db9e..1b994437 100644
--- a/py/asmthumb.c
+++ b/py/asmthumb.c
@@ -494,6 +494,7 @@ void asm_thumb_store_reg_reg_offset(asm_thumb_t *as, uint reg_src, uint reg_base
         // Must use the generic sequence
         asm_thumb_add_reg_reg_offset(as, reg_base, reg_base, offset - 31, operation_size);
         asm_thumb_op16(as, ((OP_LDR_STR_TABLE[operation_size] | OP_STR) << 11) | (31 << 6) | (reg_base << 3) | reg_src);
+        asm_thumb_add_reg_reg_offset(as, reg_base, reg_base, 31 - offset, operation_size);
     }
 }
```
but that's potentially emitting multiple instructions to undo the reg_base modification which isn't great. (Might still be a lot cheaper than touching the stack, though?)

agatti:
With your change the code getting emitted is this (here's just the relevant part):

```asm
   0x184496 <gc_heap+9550>:     movs    r2, #123        @ 0x7b
   0x184498 <gc_heap+9552>:     movw    r2, #4065       @ 0xfe1
   0x18449c <gc_heap+9556>:     adds    r2, r2, r4
   0x18449e <gc_heap+9558>:     strb    r2, [r2, #31]
```

So, yep, more work to be done here I guess.  I'll also see what I can do for this but feel free to get there first :)

agatti:
> but that's potentially emitting multiple instructions to undo the reg_base modification which isn't great. (Might still be a lot cheaper than touching the stack, though?)

It's interesting that REG_TEMP2 is chosen as the destination register.  That said, I think hitting the stack isn't as bad a choice for the time being - depending on the register combination the stack may get hit anyway when calling `asm_thumb_add_reg_reg_offset`.  There will be time to make this faster in 1.27.0.

**Edit:** `REG_ARG_3` is aliased to `REG_TEMP2` on Thumb so there's why this happens.

agatti:
Incidentally, if you use R0 you can avoid hitting the stack but I'm not feeling 100% sure this is OK to do in all cases.

I mean, it looks like it passes the whole test suite using `--via-mpy --emit=native` but this requires more consideration.

arachsys:
> It's interesting that REG_TEMP2 is chosen as the destination register. That said, I think hitting the stack isn't as bad a choice for the time being - depending on the register combination the stack may get hit anyway when calling `asm_thumb_add_reg_reg_offset`. There will be time to make this faster in 1.27.0.

I'll give this a test now:

```diff
diff --git a/py/asmthumb.c b/py/asmthumb.c
index 18c3db9e..3b996449 100644
--- a/py/asmthumb.c
+++ b/py/asmthumb.c
@@ -492,8 +492,10 @@ void asm_thumb_store_reg_reg_offset(asm_thumb_t *as, uint reg_src, uint reg_base
         asm_thumb_op32(as, (OP_LDR_STR_W_HI(operation_size, reg_base) | OP_STR_W), OP_LDR_STR_W_LO(reg_src, (offset << operation_size)));
     } else {
         // Must use the generic sequence
+        asm_thumb_op16(as, OP_PUSH_RLIST((1 << reg_base)));
         asm_thumb_add_reg_reg_offset(as, reg_base, reg_base, offset - 31, operation_size);
         asm_thumb_op16(as, ((OP_LDR_STR_TABLE[operation_size] | OP_STR) << 11) | (31 << 6) | (reg_base << 3) | reg_src);
+        asm_thumb_op16(as, OP_POP_RLIST((1 << reg_base)));
     }
 }
```

Edit: yes, that works and passes tests. Maybe I should open as a separate PR to avoid confusion in the comments applying to a different with this one?

agatti:
> Edit: yes, that works and passes tests. Maybe I should open as a separate PR to avoid confusion in the comments applying to a different with this one?

It shouldn't be a big deal - as long as the updated commit message and an updated PR description reflect reality, that should be fine I guess.  Just force push and update this PR description.

arachsys:
> > Edit: yes, that works and passes tests. Maybe I should open as a separate PR to avoid confusion in the comments applying to a different with this one?
> 
> It shouldn't be a big deal - as long as the updated commit message and an updated PR description reflect reality, that should be fine I guess. Just force push and update this PR description.

Ah, damn, I opened https://github.com/micropython/micropython/pull/17772 before I read your reply.

agatti:
No worries - going forward, you may want to investigate if using your original change but using R0 is safe to use.

That may give back the speed hit taken by using the stack for larger offsets, but R0 is also aliased to `REG_RET` and that may be used in ways that aren't covered in the current test suite.

arachsys:
> No worries - going forward, you may want to investigate if using your original change but using R0 is safe to use.
> 
> That may give back the speed hit taken by using the stack for larger offsets, but R0 is also aliased to `REG_RET` and that may be used in ways that aren't covered in the current test suite.

Thanks, I'll have a play with that too. I imagine that functions which return their first argument unmodified might want to assume that R0 isn't clobbered?

arachsys:
Closed (for now) in favour of https://github.com/micropython/micropython/pull/17772.

agatti:
I don't think that case could happen, given that it's only triggered by `emit_native_store_subscr`.  If the compiler duplicates the value to be set on the stack and then pops it off once the store is performed, then it should be fine.

Consider that R0 is *also* aliased to `REG_ARG1` on Thumb which I believe it can be clobbered by emitters.

arachsys:
> I don't think that case could happen, given that it's only triggered by `emit_native_store_subscr`. If the compiler duplicates the value to be set on the stack and then pops it off once the store is performed, then it should be fine.
> 
> Consider that R0 is _also_ aliased to `REG_ARG1` on Thumb which I believe it can be clobbered by emitters.

Ah okay, so the variables in REG_ARG_n are explicitly allowed to be used by the emitters, hence why they were fine with using R2 - because R2 was also REG_ARG_3... although does that equally mean they have permission to clobber our use of R0 in the same way they clobber R2?

agatti:
> > Consider that R0 is _also_ aliased to `REG_ARG1` on Thumb which I believe it can be clobbered by emitters.
> 
> Ah okay, so the variables in REG_ARG_n are explicitly allowed to be used by the emitters, hence why they were fine with using R2 - because R2 was also REG_ARG_3... although does that equally mean they have permission to clobber our use of R0 in the same way they clobber R2?

My understanding is that you can clobber temporaries if they're not used by the VM opcode being emitted after your emitter chunk is called, so for opcodes that perform multiple operations you can clobber temporaries if your opcode sequence is the last step in the opcode implementation.

So, in this case, `MP_EMIT_SUBSCR_STORE` generates a single opcode sequence so I suppose it is safe to clobber all argument registers - there are no intermediate steps.  The fact argument registers are aliased to temporaries doesn't really matter I think.

I don't recall right now if the native emitter framework keeps temporary registers around across opcodes rather than always use stack locals (especially if you alias arguments to temporaries).  If stack locals are always used then it *might* be safe to clobber R0/ARG1 in this case.

@dpgeorge Is my understanding correct here?  It makes sense in my head but you obviously know more about this.

arachsys:
I built and tested the variant using R0 as a temporary and it doesn't seem to cause any problems, exactly as you predicted @agatti. It passes the entire test suite and all the test cases I used to try to pinpoint/characterise the clobbering problem originally.

However, as you say, before doing this we'd need to know that it's definitely safe to use R0 in this context, rather than just getting lucky, especially as the test suite wasn't designed to test for collateral clobbering from offset stores.

Maybe using R0 here is obviously valid (or obviously invalid) to @dpgeorge? :-)

```diff
diff --git a/py/asmthumb.c b/py/asmthumb.c
index 18c3db9e4..d5ff113b1 100644
--- a/py/asmthumb.c
+++ b/py/asmthumb.c
@@ -492,8 +492,8 @@ void asm_thumb_store_reg_reg_offset(asm_thumb_t *as, uint reg_src, uint reg_base
         asm_thumb_op32(as, (OP_LDR_STR_W_HI(operation_size, reg_base) | OP_STR_W), OP_LDR_STR_W_LO(reg_src, (offset << operation_size)));
     } else {
         // Must use the generic sequence
-        asm_thumb_add_reg_reg_offset(as, reg_base, reg_base, offset - 31, operation_size);
-        asm_thumb_op16(as, ((OP_LDR_STR_TABLE[operation_size] | OP_STR) << 11) | (31 << 6) | (reg_base << 3) | reg_src);
+        asm_thumb_add_reg_reg_offset(as, ASM_THUMB_REG_R0, reg_base, offset - 31, operation_size);
+        asm_thumb_op16(as, ((OP_LDR_STR_TABLE[operation_size] | OP_STR) << 11) | (31 << 6) | (ASM_THUMB_REG_R0 << 3) | reg_src);
     }
 }
```

dpgeorge:
> Maybe using R0 here is obviously valid (or obviously invalid) to @dpgeorge? :-)

I don't think you can use R0 for the following reason (line 1270 of emitnative.c):
```c
ASM_STORE_REG_REG_OFFSET(emit->as, REG_TEMP0, REG_GENERATOR_STATE, OFFSETOF_CODE_STATE_STATE);
```
So R0 is used as the source register here.

Now, the offset there is small enough that you don't hit the code path that you need a large immediate offset, but still it's very risky to assume you can clobber R0.

PR #17772, opened 2025-07-27
py/asmthumb: Don't corrupt base register in large offset store.

arachsys:
### Summary

On rp2, I discovered that viper stores offset from a register variable were clobbering that variable:

```python
@micropython.viper
def test():
  x = ptr8(ba)
  print(hex(uint(x)))
  x[0x1000] = 0
  # The register x has now been clobbered, incremented by 0x1000 - 31.
  # It gets corrupted only if the index/offset is >= 1 << 10.
  print(hex(uint(x)))
ba = bytearray(0x2000)
test()
```

@agatti kindly pointed me at `asm_thumb_store_reg_reg_offset()`. It turns out this modifies the base register when storing with a large offset which triggers the generic path. If a variable lives in that register, this corrupts it.

Fix this by saving the base register on the stack before modifying it. (I tried using REG_TEMP2 which would be faster but this turned out to collide with the generated code.)

### Testing

The existing viper ptr8, ptr16 and ptr32 tests all pass, and the second commit adds assertions to the viper store boundary tests to verify the base register is not corrupted during the store.

### DIFF
--- a/py/asmthumb.c
+++ b/py/asmthumb.c
@@ -492,8 +492,10 @@ void asm_thumb_store_reg_reg_offset(asm_thumb_t *as, uint reg_src, uint reg_base
         asm_thumb_op32(as, (OP_LDR_STR_W_HI(operation_size, reg_base) | OP_STR_W), OP_LDR_STR_W_LO(reg_src, (offset << operation_size)));
     } else {
         // Must use the generic sequence
+        asm_thumb_op16(as, OP_PUSH_RLIST(1 << reg_base));
         asm_thumb_add_reg_reg_offset(as, reg_base, reg_base, offset - 31, operation_size);
         asm_thumb_op16(as, ((OP_LDR_STR_TABLE[operation_size] | OP_STR) << 11) | (31 << 6) | (reg_base << 3) | reg_src);
+        asm_thumb_op16(as, OP_POP_RLIST(1 << reg_base));
     }
 }
 
--- a/tests/micropython/viper_ptr16_store_boundary_intbig.py
+++ b/tests/micropython/viper_ptr16_store_boundary_intbig.py
@@ -3,7 +3,9 @@
 SET_TEMPLATE = """
 @micropython.viper
 def set{off}(dest: ptr16):
+    saved = dest
     dest[{off}] = {val}
+    assert int(saved) == int(dest)
 set{off}(buffer)
 print(hex(get_index(buffer, {off})))
 """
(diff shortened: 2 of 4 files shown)


dpgeorge:
See previous discussion in #17771.

arachsys:
I benchmarked the speed hit of saving the base register on a real rp2350. Given
`ba = bytearray(0x10000)` and `x = ptr32(ba)`, a viper function executes `x[0x3805] = -1` in 5 cycles if we use a scratch register as discussed in https://github.com/micropython/micropython/pull/17771, and in 9 cycles with this patch.

If we're writing to a memory mapped register (e.g. if `x = ptr(-0x20000000)` which was my original test case), those timings are 5 cycles and 11 cycles instead.

agatti:
> If we're writing to a memory mapped register (e.g. if `x = ptr(-0x20000000)` which was my original test case), those timings are 5 cycles and 11 cycles instead.

An alternative could be to use `machine.mem32` (see `docs/library/machine.rst`), as in the documentation it is introduced as an easier way to read/write CPU registers from Python.  It's probably slower, but sometimes allows for cleaner code.

arachsys:
Absolutely, using viper for accessing registers is definitely a low-level speed optimisation - though ironically, it ends up looking at least as readable as mem32 in code that's already using viper.

For interest, `machine.mem32[0xe000e014] = - 1` takes around 400 cycles, so 40-80x slower depending on which offset store fix we use. If you pre-bind `mem32 = machine.mem32`, `mem32[0xe000e014] = -1` would take only 144 cycles.

Amusingly, `ppb = ptr32(-0x20000000)` or `ppb = ptxr32(0xe000 << 16)` are cheap (3 cycles or 8 cycles respectively), but `ppb = ptr32(0xe0000000)` is 104 cycles. Although it looks like `asm_thumb_mov_reg_i32()` would generate good code for 31- and 32-bit immediates, I can't find any way of getting a constant 32-bit value to it directly without 100+ cycles of object dereference. `x = const(0xe0000000)` at top level then `p = ptr32(x)` in viper doesn't work, because x becomes a constant object reference not a constant large i32 nowadays.

dpgeorge (approved):
This is a necessary change to fix the current code, which is a regression (of sorts) since v1.25.0.

In the future we could think of ways to improve this, maybe pass in a register to `asm_thumb_store_reg_reg_offset()` which can be optionally used as a temporary.

dpgeorge:
Tested on PYBV10, that the issue is present on that target and that this PR fixes it.

dpgeorge:
Rebased and merged in f39434e9fb17405754277dd62dfea788a42d5afb and 953da2080eb9115536101627c7406727c5d44f8b.

Thanks for the fix!

PR #17775, opened 2025-07-28
mpremote/mp_errno: Fix errno.ENOTBLK error on Windows.

Josverl:
### Summary
not all errors defined in stdlib `errno` are available on Windows.
Specifically `errno.ENOTBLK` is not.
Not 100% sure about availability of all other codes across platforms either, so would be good if someone can confirm that MacOS has no problems either.

This was not caught in CI, as there is no CI , nor tests for mpremote on Windows.

closes: https://github.com/micropython/micropython/issues/17773

### Testing
Manual testing on Windows 11 and WSL2.

### DIFF
--- a/tools/mpremote/mpremote/mp_errno.py
+++ b/tools/mpremote/mpremote/mp_errno.py
@@ -1,4 +1,5 @@
 import errno
+import platform
 
 # This table maps numeric values defined by `py/mperrno.h` to host errno code.
 MP_ERRNO_TABLE = {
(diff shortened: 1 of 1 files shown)


Josverl:
IMO its quite unlikely that the failing test (thread/thread_gc1.py) is caused by this PR.
it has been reported as a flaky test in other PRs as well.

dpgeorge (approved):
Thanks for the fix, this looks fine.

I confirmed that `errno.ENOTBLK` exists on macOS on Python 3.8.9.

Issue #9659, opened 2022-10-17
mpremote: all mpremote commands are followed by a hard reboot


Josverl:
according to @jimmo  mpremote should be able to resume a repl session on a connected mcu without a reboot by using the following command pattern:
``` sh
mpremote connect exec "start_something()"
mpremote connect resume exec "continue_here()"
```
where the `resume` is the important part.
see [Discord discussion](https://discord.com/channels/574275045187125269/1029416691488264252/1031558675900407818) 

however when I try this : 
 1. I see that any variables set in the the first `mpremote` commands are lost 
 2. I see that the esp32 performs a soft reset after exiting the `raw REPL` 
     (observed through logic analyzer)

So even though the 2nd mpremote command does not force a reset, by that time a soft reboot has already occurred.
I also see a `soft reset` after: 

- exiting the repl
- eval "42"
- exec "print(42)"

## Actual result on windows : 
```
C:\develop\usb>mpremote exec "foo=bar=42;print(foo)"
42

C:\develop\usb>mpremote resume exec "print(bar)"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'bar' isn't defined
```
## Expected 
the MCU not to reset between two mpremote commands similar to the behavour on linux 

```
$ mpr exec "foo=bar=42; print(foo)"
42
$ mpr exec "print(bar)"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'bar' isn't defined
$ mpr exec "foo=bar=42; print(foo)"
42
$ mpr resume exec "print(bar)"
42
```

results as seen by a logic analyzer hooked up to the RX/TX pins of the esp32 
![image](https://user-images.githubusercontent.com/981654/196273853-8dcf6a8d-9942-495d-ad7e-6234ef50400c.png)

``` text
\x01\x03
>>> 
>>> 
>>> 
>>> 
raw REPL; CTRL-B to exit
\x02OKx05A\x01foo=bar=43;print(foo)\x04
MPY: soft reboot
raw REPL; CTRL-B to exit
>R\x01\x80\0\x01\x0443
\x04\x04>ets Jun  8 2016 00:22:57

rst:0x1 (POWERON_RESET),boot:0x17 (SPI_FAST_FLASH_BOOT)
configsip: 0, SPIWP:0xee
clk_drv:0x00,q_drv:0x00,d_drv:0x00,cs0_drv:0x00,hd_drv:0x00,wp_drv:0x00
[...]

PR #17776, opened 2025-07-28
mpremote: fix esp usb reset

Josverl:
### Summary

Detection of ESP-XX devices was based on just the names of the USB driver name.
and did not account for the switch of the newer ESP-xx devices to USB-CDC.
On Windows this caused unwanted/unneeded resets as the DTR/RTS signals are also used for automatic device reset over USB-CDC as reported by @crackwitz in https://github.com/micropython/micropython/issues/9659#issuecomment-3124704572.

This PR uses the Espressif registered VID : 0x303A to detect USB-CDC ports ,
to enable the same DTR/RTS settings as used on the UART-USB connection.

Also improved the robustness of the code using `getattr()`

Depends on : #17775

### Testing
Manual testing on Windows and WSL2

| device| platform| USB-CDC| UART|
|--------|--------|--------|--------|
| Win11 | ESP32-C3| ✅| ✅|
| WSL2-Ubuntu | ESP32-C3| ✅| ✅|
| Win11 | ESP32-C2| ✅ | n/a |
| WSL2-Ubuntu| ESP32-C2| ✅ | n/a |
| Win11 | rp2040| n/a |  ✅ | 
| Win11 | ESP32-S3| ✅| ✅|
| WSL2-Ubuntu | ESP32-S3| ✅| ✅|

Detection of the ESPxx devices via the USB-CDC VID is more deterministic than `getattr(portinfo[0], "manufacturer", "") != "Microsoft"`, and will cover most of not all Espressif boards.

One possible drawback is that if/when other board manufacturers register their own VIDs for devices powered by an ESP32 - and access over USB-CDC that they will be exposed to unexpected resets.

### DIFF
--- a/tools/mpremote/mpremote/transport_serial.py
+++ b/tools/mpremote/mpremote/transport_serial.py
@@ -40,6 +40,8 @@ from errno import EPERM
 from .console import VT_ENABLED
 from .transport import TransportError, TransportExecError, Transport
 
+VID_ESPRESSIF = 0x303A  # Espressif Incorporated
+
 
 class SerialTransport(Transport):
     fs_hook_mount = "/remote"  # MUST match the mount point in fs_hook_code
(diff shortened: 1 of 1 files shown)


projectgus (approved):
I don't have a Windows setup to easily verify the fix but this seems very reasonable, thanks @Josverl!

dpgeorge (approved):

dpgeorge:
Thank you!

PR #17777, opened 2025-07-29
esp32: Update the README details to account for newer chips.

projectgus:
### Summary

Noted while adding C2 support in #15906 that some of the points in the README are a little out of date. spun out to its own PR to avoid clogging up that PR. Have also updated to mention C5 as well.

This PR also adds some recommendation on which ESP32 board to pick, as we occasionally see issues or questions that would be non-issues on a board with more RAM (and for small production or personal projects the savings of picking a cheaper ESP32 chip are basically neglible).

### DIFF
--- a/ports/esp32/README.md
+++ b/ports/esp32/README.md
@@ -5,22 +5,43 @@ This is a port of MicroPython to the Espressif ESP32 series of
 microcontrollers.  It uses the ESP-IDF framework and MicroPython runs as
 a task under FreeRTOS.
 
-Currently supports ESP32, ESP32-C2 (aka ESP8684), ESP32-C3, ESP32-C6, ESP32-S2
-and ESP32-S3 (ESP8266 is supported by a separate MicroPython port).
+Currently supports ESP32, ESP32-C2 (aka ESP8684), ESP32-C3, ESP32-C5, ESP32-C6,
+ESP32-S2 and ESP32-S3. ESP8266 is supported by a separate MicroPython port.
 
 Supported features include:
-- REPL (Python prompt) over UART0.
-- 16k stack for the MicroPython task and approximately 100k Python heap.
+- REPL (Python prompt) over UART0 and/or the integrated USB peripheral.
 - Many of MicroPython's features are enabled: unicode, arbitrary-precision
   integers, single-precision floats, complex numbers, frozen bytecode, as
   well as many of the internal modules.
-- Internal filesystem using the flash (currently 2M in size).
+- Internal filesystem using the remaining flash area.
+- Threading support via the _thread module (built on native FreeRTOS tasks),
+  with GIL enabled.
 - The machine module with GPIO, UART, SPI, software I2C, ADC, DAC, PWM,
   TouchPad, WDT and Timer.
 - The network module with WLAN (WiFi) support.
 - Bluetooth low-energy (BLE) support via the bluetooth module.
 
 Initial development of this ESP32 port was sponsored in part by Microbric Pty Ltd.
+
+Choosing a suitable chip
+------------------------
+
+ESP32 chips are not all the same. The different ESP32 families have different
+capabilities and resources available. In particular, the ESP32-C2 and ESP32-S2
+(without external SPIRAM) have the least RAM. They can still run MicroPython
+well but may run out of RAM if a program uses a lot of resources, eg if it
+needs many complex code modules, multiple TLS connections, large memory
+buffers for a display, etc.
+
+ESP32 boards with external "SPIRAM" (also called PSRAM) have megabytes of RAM
+available - significantly more than any board without external SPIRAM. SPIRAM is
+supported on original ESP32, ESP32-S2 and ESP32-S3 chips. Not every ESP32 board
+has SPIRAM included, even if the chip supports it.
+
+If you don't need particular hardware features but are looking for a board with
+the usual Wi-Fi and BLE support, many peripherals and plenty of RAM then
+recommend finding a board based on ESP32-S3 with SPIRAM included (usually 2MB,
+4MB or 8MB).
 
 Setting up ESP-IDF and the build environment
 --------------------------------------------


ports/esp32/README.md
```diff
@@ -10,18 +10,40 @@ Currently supports ESP32, ESP32-C3, ESP32-C6, ESP32-S2 and ESP32-S3
...
+- Threading support via the _thread module (built on native FreeRTOS tasks),
+  with GIL enabled.
 - The machine module with GPIO, UART, SPI, software I2C, ADC, DAC, PWM,
   TouchPad, WDT and Timer.
 - The network module with WLAN (WiFi) support.
 - Bluetooth low-energy (BLE) support via the bluetooth module.
 
 Initial development of this ESP32 port was sponsored in part by Microbric Pty Ltd.
 
+Choosing Correct Chip
```
dpgeorge: Suggest "Choosing the correct chip" or "Choosing a chip"

ports/esp32/README.md
```diff
@@ -10,18 +10,40 @@ Currently supports ESP32, ESP32-C3, ESP32-C6, ESP32-S2 and ESP32-S3
...
+---------------------
+
+ESP32 chips are not all the same. The different ESP32 families have different
+capabilities and resources available. In particular, the ESP32-C2 and ESP32-S2
+(without external SPIRAM) have the least RAM. They can still run MicroPython
+well but may run out of RAM if a program uses a lot of resources (i.e. if it
```
dpgeorge: Suggest removing the parenthesis for this last sentence (it's important information), and changing "ie" to "eg".

ports/esp32/README.md
```diff
@@ -10,18 +10,40 @@ Currently supports ESP32, ESP32-C3, ESP32-C6, ESP32-S2 and ESP32-S3
...
+needs many complex code modules, multiple TLS connections, or large memory
+buffers for a display, etc.)
+
+ESP32 chips with external "SPIRAM" (supported on ESP32, ESP32-S2, ESP32-S3) have
+megabytes of RAM available - significantly more than any ESP32 without external
+SPIRAM. Not every ESP32 board has SPIRAM included.
+
+The supported ESP32 chip with the most hardware resources is the ESP32-S3 with
+external SPIRAM included (usually 2MB, 4MB, or 8MB).
+
```
dpgeorge, line 45: Is it worth mentioning some other higher-level items, eg native USB, WiFi and BLE support?  Also the CPU architecture?

ports/esp32/README.md
projectgus: I went with "Choosing a suitable chip"

ports/esp32/README.md
projectgus: Have added a short summary. I mentioned dual CPU cores rather than architecture because at the level "picking first dev board for MicroPython" the underlying architecture doesn't make much difference.

ports/esp32/README.md
```diff
@@ -10,18 +10,41 @@ and ESP32-S3 (ESP8266 is supported by a separate MicroPython port).
 
 Supported features include:
 - REPL (Python prompt) over UART0.
```
Josverl: Should this mention usb-cdc as well?
( for the cpu/ boards that provide that)

ports/esp32/README.md
```diff
@@ -10,18 +10,41 @@ and ESP32-S3 (ESP8266 is supported by a separate MicroPython port).
...
-- 16k stack for the MicroPython task and approximately 100k Python heap.
+- 16KiB stack for the MicroPython task and approximately 110KiB of free RAM for
+  MicroPython heap and other usage. (This number is for the original ESP32
```
Josverl: I suggest keeping the feature list simple, and move the different memory sizes to the paragraph below where this is already discussed

Josverl (approved):

ports/esp32/README.md
dpgeorge: Yes, I think that's a good idea, to mention native USB REPL.

peterhinch (commented):
The statement "The ESP32-S3 has many other hardware features including USB, Wi-Fi, BLE and dual CPU cores." might be worth expanding.

1. As far as I know all ESP32's have WiFi.
2. The mention of USB may confuse beginners. Most ESP32x boards have USB: the distinction is between native and non-native USB.
3. The dual core feature has already caused confusion, particularly in people coming from RP2. Many variants (including the original ESP32) have dual core but the second core is not accessible to MP.

robert-hh:
> As far as I know all ESP32's have WiFi.

Yes and no. The ESP32P4 MCU for onstance does not support Wifi or Bluetooth,  but boards or modules with ESP32P4 carry a second ESP32 which provides wireless communication.

ports/esp32/README.md
projectgus: I agree, these numbers vary widely these days.

ports/esp32/README.md
projectgus: Done!

projectgus:
> The statement "The ESP32-S3 has many other hardware features including USB, Wi-Fi, BLE and dual CPU cores." might be worth expanding.

Fair points, thanks @peterhinch. I've rewritten it to say what I actually meant, which is essentially "if you're new to this and you don't know what ESP32 chip to buy, you should probably get an ESP32-S3 with PSRAM". Please take a look at the new paragraph.

(I guess the alternative is to make a full feature matrix, but keeping one up to date seems ambitious and there's plenty of detailed technical docs out there already.)

dpgeorge (approved):

pi-mst:
> > As far as I know all ESP32's have WiFi.
> 
> Yes and no. The ESP32P4 MCU for onstance does not support Wifi or Bluetooth, but boards or modules with ESP32P4 carry a second ESP32 which provides wireless communication.

Just a minor clarification...

Certainly not all ESP32-P4s have wifi; I have units that have either no connectivity or only wired ethernet.

Regardless, this README update LGTM!

PR #13586, opened 2024-02-02
tools\mpremote\mpremote\transport_serial.py: Fix encoding error in PyboardCommand class.

Josverl:
This is a fix for utf-8 decoding errors that are thrown when non-utf8 content is received. For instance during a reboot of an ESP8266 module.

The fix is to handle conversion errors by replacing illegal characters.
Note that these illegal characters most often occur during an MCU reboot sequence when the MCU is using baudrates different from115200

<details>
<summary>Error Details</summary>

```
11:52:31|INFO    |board_stubber        - Local directory C:\Users\josverl\AppData\Local\Temp\board_stubberagrvu06k is mounted at /remote
11:52:31|INFO    |board_stubber        - b' \r\npaste mode; Ctrl-C to cancel, Ctrl-D to finish\r\n=== A\x01'
11:52:31|WARNING |board_stubber        - Traceback (most recent call last):

11:52:31|WARNING |board_stubber        -   File "C:\develop\MyPython\micropython-stubber\.venv\Lib\site-packages\mpremote\main.py", line 502, in main

11:52:31|WARNING |board_stubber        -     handler_func(state, args)

11:52:31|WARNING |board_stubber        -   File "C:\develop\MyPython\micropython-stubber\.venv\Lib\site-packages\mpremote\commands.py", line 203, in do_exec

11:52:31|WARNING |board_stubber        -     _do_execbuffer(state, args.expr[0], args.follow)

11:52:31|WARNING |board_stubber        -   File "C:\develop\MyPython\micropython-stubber\.venv\Lib\site-packages\mpremote\commands.py", line 191, in _do_execbuffer

11:52:31|WARNING |board_stubber        -     ret, ret_err = state.pyb.follow(timeout=None, data_consumer=pyboard.stdout_write_bytes)

11:52:31|WARNING |board_stubber        -                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

11:52:31|WARNING |board_stubber        -   File "C:\develop\MyPython\micropython-stubber\.venv\Lib\site-packages\mpremote\pyboard.py", line 388, in follow

11:52:31|WARNING |board_stubber        -     data = self.read_until(1, b"\x04", timeout=timeout, data_consumer=data_consumer)

[...]

PR #17780, opened 2025-07-29
tests/run-multitests: Escape encoding errors instead of crashing.

projectgus:
### Summary

This is another small quality-of-life tweak while testing #15906.

It's possible for a test to output non-ASCII characters (for example, due to a hard fault or serial noise or memory corruption). Rather than crashing the test runner, backslash escape those characters and treat them as program output.

Refactors the string encoding step to a single helper to avoid copy-paste.

### Testing

* Re-ran multitests between ESP32-C2 and unix micropython, verified expected behaviour.

### DIFF
--- a/tests/run-multitests.py
+++ b/tests/run-multitests.py
@@ -132,6 +132,11 @@ def get_host_ip(_ip_cache=[]):
     return _ip_cache[0]
 
 
+def decode(output):
+    # Convenience function to convert raw process or serial output to ASCII
+    return str(output, "ascii", "backslashreplace")
+
+
 class PyInstance:
     def __init__(self):
         pass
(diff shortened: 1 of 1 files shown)


tests/run-multitests.py
```diff
@@ -264,7 +269,7 @@ def run_script(self, script):
             output = self.pyb.exec_(script)
         except pyboard.PyboardError as er:
             err = er
-        return str(output.strip(), "ascii"), err
+        return decode(output), err
```
dpgeorge: I think you accidentally removed the strip.

projectgus: oops, thanks! Fixed

Josverl:
a fix for similar problems in mpremote
https://github.com/micropython/micropython/pull/13586/files

dpgeorge (approved):

PR #17783, opened 2025-07-29
py/mphal: Add header for size_t.

yf13:
### Summary

This adds `#include <stddef.h>` for definition of `size_t` to solve compilation issue encountered with NuttX integration:

```
py/mphal.h:59:50: error: unknown type name 'size_t'
   59 | mp_uint_t mp_hal_stdout_tx_strn(const char *str, size_t len);
      |                                                  ^~~~~~
py/mphal.h:34:1: note: 'size_t' is defined in header '<stddef.h>'; did you forget to '#include <stddef.h>'
```

### Testing

Checked with NuttX integration on QEMU riscv.

### DIFF
--- a/py/mphal.h
+++ b/py/mphal.h
@@ -27,6 +27,7 @@
 #define MICROPY_INCLUDED_PY_MPHAL_H
 
 #include <stdint.h>
+#include <stddef.h>
 #include "py/mpconfig.h"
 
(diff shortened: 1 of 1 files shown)


projectgus (approved):
Thanks @yf13! Seems sensible from an "include what you use" perspective.

PR #17785, opened 2025-07-29
Fix more operations that can return small ints

jepler:
### Summary

#17730 identified some code paths that could still return long ints when small ints would suffice. Fix them.

### Testing

I wrote a new test based on sites that called new_int_from_{u,}ll, including the case I'd noticed of pow3 (already reported in #9531 as well).

I haven't seen the code size change summary yet, but this is likely to increase code size. I'm open to suggestions about how to reduce the increase.

Making the new `long_from` functions always return longs, and changing the existing `new_int_from` functions to return ints when possible is just one alternative. Another alternative would be to add calls to `mp_int_maybe_narrow` in paths that used `new_int_from`, but this seemed like the worse alternative, with more added code.

### DIFF
--- a/py/obj.h
+++ b/py/obj.h
@@ -996,8 +996,9 @@ mp_obj_t mp_obj_new_cell(mp_obj_t obj);
 mp_obj_t mp_obj_new_int(mp_int_t value);
 mp_obj_t mp_obj_new_int_from_uint(mp_uint_t value);
 mp_obj_t mp_obj_new_int_from_str_len(const char **str, size_t len, bool neg, unsigned int base);
-mp_obj_t mp_obj_new_int_from_ll(long long val); // this must return a multi-precision integer object (or raise an overflow exception)
-mp_obj_t mp_obj_new_int_from_ull(unsigned long long val); // this must return a multi-precision integer object (or raise an overflow exception)
+mp_obj_t mp_obj_new_int_from_ll(long long val); // returns a small int if value fits
+mp_obj_t mp_obj_new_int_from_ull(unsigned long long val); // returns a small int if value fits
+mp_obj_t mp_obj_new_bigint_from_ll(long long val); // this must return a multi-precision integer object (or raise an overflow exception)
 mp_obj_t mp_obj_new_str(const char *data, size_t len); // will check utf-8 (raises UnicodeError)
 mp_obj_t mp_obj_new_str_from_cstr(const char *str); // // accepts null-terminated string, will check utf-8 (raises UnicodeError)
 mp_obj_t mp_obj_new_str_via_qstr(const char *data, size_t len); // input data must be valid utf-8
--- a/ports/unix/coverage.c
+++ b/ports/unix/coverage.c
@@ -528,9 +528,9 @@ static mp_obj_t extra_coverage(void) {
         mp_printf(&mp_plat_print, "%f\n", mpz_as_float(&mpz));
 
         // convert a large integer value (stored in a mpz) to mp_uint_t and to ll;
-        mp_obj_t obj_bigint = mp_obj_new_int_from_uint((mp_uint_t)0xdeadbeef);
+        mp_obj_t obj_bigint = mp_obj_new_int_from_ull((unsigned long long)0xdeadbeef);
         mp_printf(&mp_plat_print, "%x\n", (int)mp_obj_get_uint(obj_bigint));
-        obj_bigint = mp_obj_new_int_from_ll(0xc0ffee777c0ffeell);
+        obj_bigint = mp_obj_new_bigint_from_ll(0xc0ffee777c0ffeell);
         long long value_ll = mp_obj_get_ll(obj_bigint);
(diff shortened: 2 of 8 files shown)


jepler:
the nrf port is failing due to a "403 Forbidden" error downloading some files from nordic:
```

####################################
### Downloading s140_nrf52_6.1.1 ###
####################################

--2025-07-30 15:05:28--  https://www.nordicsemi.com/api/sitecore/Products/MedialibraryZipDownload2
Resolving www.nordicsemi.com (www.nordicsemi.com)... 104.18.26.60, 104.18.27.60, 2606:4700::6812:1b3c, ...
Connecting to www.nordicsemi.com (www.nordicsemi.com)|104.18.26.60|:443... connected.
HTTP request sent, awaiting response... 403 Forbidden
2025-07-30 15:05:29 ERROR 403: Forbidden.

mv: cannot stat 'MedialibraryZipDownload2': No such file or directory
unzip:  cannot find or open temp.zip, temp.zip.zip or temp.zip.ZIP.
unzip:  cannot find or open s140nrf52611.zip, s140nrf52611.zip.zip or s140nrf52611.zip.ZIP.
rm: cannot remove 's140nrf52611.zip': No such file or directory
rm: cannot remove 'temp.zip': No such file or directory
```
furthermore, the failure is not detected at the time, but the build continues until this message is shown
```

drivers/bluetooth/bluetooth_common.mk:48: *** 

###### ERROR: Bluetooth LE Stack not found ############
#                                                     #
# The build target requires a Bluetooth LE stack.     #
# s140_nrf52_6.1.1 Bluetooth LE stack not found.      #
#                                                     #
# Please run the download script:                     #
#                                                     #
#      drivers/bluetooth/download_ble_stack.sh        #
#                                                     #
#######################################################
.  Stop.
```

with any luck this is a transient error and nordic will fix the download link.

robert-hh:
The same error popped up at PR #17365.

dhalbert:
> with any luck this is a transient error and nordic will fix the download link.

CircuitPython GitHub actions builds are also seeing this. It looks like `wget` has become a forbidden user agent for certain nordicsemi.com URL's (though not all). I reported this to NordicSemi: https://devzone.nordicsemi.com/f/nordic-q-a/123372/https-developer-nordicsemi-com-is-blocking-some-user-agent-values-on-downloads

py/obj.h
```diff
@@ -995,8 +995,9 @@ mp_obj_t mp_obj_new_cell(mp_obj_t obj);
 mp_obj_t mp_obj_new_int(mp_int_t value);
 mp_obj_t mp_obj_new_int_from_uint(mp_uint_t value);
 mp_obj_t mp_obj_new_int_from_str_len(const char **str, size_t len, bool neg, unsigned int base);
-mp_obj_t mp_obj_new_int_from_ll(long long val); // this must return a multi-precision integer object (or raise an overflow exception)
-mp_obj_t mp_obj_new_int_from_ull(unsigned long long val); // this must return a multi-precision integer object (or raise an overflow exception)
+mp_obj_t mp_obj_new_int_from_ll(long long val); // returns a small int if value fits
+mp_obj_t mp_obj_new_int_from_ull(unsigned long long val); // returns a small int if value fits
+mp_obj_t mp_obj_new_long_from_ll(long long val); // this must return a multi-precision integer object (or raise an overflow exception)
```
projectgus: It's fully separate to this PR, but at some point we should probably rename all the `longint` references in MicroPython to `bigint` to avoid this ambiguity about what `long` means.

projectgus (approved):
This looks good to me, thanks @jepler.

I tried to think of ways to combine `mp_obj_new_int`, `mp_obj_new_int_from_ll`, and `mp_obj_new_int_from_ull` into a single inner static implementation but they were all going to involve some level of compiler crimes I think. :laughing: 

I've linked this PR to close the issue as IIUC it addresses everything enumerated there, but let me know if I'm missing a case.

py/obj.h
dpgeorge: I really think this PR should already start using the name `mp_obj_new_bigint_from_ll()` for this new function, because the word "long" is too overloaded here.

Then a follow-up PR (maybe by @projectgus ) can rename `MICROPY_LONGINT_IMPL` to `MICROPY_BIGINT_IMPL` etc.

dpgeorge:
Thanks @jepler for this PR.  I'm a little on the fence with it because there's nothing technically broken that it's fixing.  It's just improving things (in a good way).  But OTOH it has a modest size increase to the code.

Is there anyway to reduce code size?  Some ideas:
- move `mp_obj_int_pow3()` to `py/modbuiltins.c` and make it static so it can be inlined there
- remove unreachable code from mpz, see https://github.com/micropython/micropython/pull/9531#issuecomment-1269305359 and https://github.com/micropython/micropython/pull/9531#issuecomment-1274519922

dpgeorge:
I also want to run performance tests with these changes to see what the impact is.

py/obj.h
jepler: I have renamed the added function.

PR #17795, opened 2025-07-31
esp32/boards/SEEED_XIAO_ESP32_C3: Add new board definition.

Jasionf:


ports/esp32/boards/SEEED_XIAO_ESP32_C3/board.json
```diff
@@ -0,0 +1,22 @@
...
+        "flash_offset": "0"
+    },
+    "docs": "",
+    "features": [
+        "BLE",
+        "External Flash",
+        "WiFi"
+    ],
+    "images": [
+        "xiao_esp32c3-font.jpg"
```
pyramids: Where does this file name come from?
As far as I can tell, it corresponds neither to a file in the repo or (in an obvious way) forms a complete URL.

And yet, in the ports/rp2/boards/SEED_XIAO_RP2350 board, a similarly inexplicable image file name results in the official MicroPython download page [https://micropython.org/download/SEEED_XIAO_RP2350/] having an actual image of the same file name!

I am asking primarily to self-educate (for creating more XIAO board definitions myself). Can I arbitrarily correct the apparent spelling error ("font" to "front") or will that break the magic of the image somehow appearing in the official MicroPython download web page?

pyramids:
The `pin.csv` definitions (D0 through D10) are **complete** despite that being rather atypical for XIAO boards:

This board does not have any GPIO-controlled LEDs and all the other GPIO-associated pads (the fairly standardized ones on the bottom of XIAO boards are present but) are, on this particular board, not extra GPIOs. These pads are instead connected to the same pins already accessible in the D0 to D10 range.

An potential alternative / addition:
I am secretly wishing for some sort of standard for accessing the boot button, at least across esp32xxx XIAO boards, but I do not know if that is worth starting to introduce pin aliases to `pin.csv`---if one wanted to provide such an alias, GPIO9 would have to be defined as both D9 and as BTN or the like.

PR #17796, opened 2025-07-31
Xiao esp32 s3

Jasionf:


sosi-deadeye:
The pin-mapping is correct. I did it 2 years ago: https://pastebin.com/zLmjTh3k

ricksorensen:
Should not the XIAO S3 build have SPIRAM OCT enabled?

pyramids:
For the fairly standardized XIAO boards (especially regarding output pin definitions in defined locations and hence interchangeable across boards), as a user, I would wish for more consistency across XIAO board definitions, including across MicroPython ports.

As I'm not sure how to raise that issue, I'll link to my discussion comment about inconsistent and missing XIAO pin definitions where I ask what to do about it before even more new board definitions perpetuate this issue: https://github.com/orgs/micropython/discussions/18186#discussioncomment-14734274

My main complaint is that PRs forget to define the---consistently physically present!---pads on the bottom side of the board and the---often but not always present---internal GPIO functionality (LEDs, antenna switches, etc.).

PR #17813, opened 2025-08-02
py: Add MICROPY_PY_STRUCT_UNSAFE_TYPECODES.

jepler:
### Summary

This adds a compile-time flag to disable some "unsafe" and non-standard typecodes in struct, array & related modules.

The typecodes "P", "O", and "S" can allow unchecked access to arbitrary memory. This is undesired in the context of fuzzing, so it's useful to specifically disable these codes.

The typecodes are also not part of CPython, and as such CircuitPython has disabled them for years.

This item is from a list prepared by @dhalbert and reviewed by @jimmo about a year ago of CircuitPython changes that might be worth offering upstream to MicroPython. The implementation is not taken from CircuitPython because I reimplemented it in the context of fuzzing instead of finding the right commit to cherry-pick.

### Testing

Locally I have tested this in builds intended to run under a fuzzer. No CI build tests with this setting changed from its default (which is enabled, leading to no behavioral change compared to the status quo)

### DIFF
--- a/py/binary.c
+++ b/py/binary.c
@@ -69,11 +69,13 @@ size_t mp_binary_get_size(char struct_type, char val_type, size_t *palign) {
                 case 'Q':
                     size = 8;
                     break;
+                #if MICROPY_PY_STRUCT_UNSAFE_TYPECODES
                 case 'P':
                 case 'O':
                 case 'S':
                     size = sizeof(void *);
                     break;
+                #endif
                 case 'e':
                     size = 2;
                     break;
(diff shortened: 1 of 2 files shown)


jepler:
The name of the flag is somewhat inaccurate, since this is also an array typecode.

dpgeorge:
> The name of the flag is somewhat inaccurate, since this is also an array typecode.

I think that's OK.

dpgeorge (approved):
Looks good, thank you!

dpgeorge:
Rebased and merged in 4614ee9e682dfffa26a4226645baca0208ecfac6

PR #17814, opened 2025-08-02
unix: Add aflplusplus variant for fuzzing. 

jepler:
This also includes #17813

### Summary

Recently I found and reported a number of core issues found by the AFLplusplus fuzzer. This PR introduces a variant that is useful for running under AFLplusplus, and provides a small overview of the process of actually running the fuzzer.

### Testing

I've run the fuzzing build locally for a few dozen CPU hours and reported most of the interesting findings.

There are other fuzzers out there. It would probably be infeasible to support them all. AFL is the one I started with, and today its successor AFLplusplus is an active project.

### DIFF
--- a/ports/unix/variants/aflplusplus/README.md
+++ b/ports/unix/variants/aflplusplus/README.md
@@ -1,0 +1,24 @@
+# Unix Fuzzing Variant
+
+This variant is for use with the [AFLplusplus](https://github.com/AFLplusplus/AFLplusplus)
+fuzzer.
+
+ 1. Install AFLplusplus so that the program `afl-cc` is on $PATH
+ 1. `cd ports/unix && make VARIANT=aflplusplus -j$(nproc)`
+ 1. Gather your inputs (e.g., from test cases in `tests`)
+ 1. Optionally, minimize them e.g., with `afl-cmin` (see AFLplusplus docs)
+ 1. Run the fuzzer. The simplest single process way is: `afl-fuzz -i inputs -o findings -- ports/unix/build-aflfuzz/micropython @@`
+
+Eventually, if crashing test cases are found, the crashing program(s) are placed in
+`findings/default/crashes` and invocations that were determined to hang go to
+`findings/default/hangs`.
+
+There are many more advanced ways to run the fuzzer; see the AFLplusplus documentation for info.
+
+# Safety
+
+Functionality that is known to be unsafe (host filesystem write access via vfs_posix)
+or is accepted as causing crashes when used improperly (ctypes, certain struct
+& array typecodes) is build-time disabled. However, it's always possible that
+the fuzzer will find a condition that can cause an unexpected modification to
+the runtime environment.
--- a/ports/unix/main.c
+++ b/ports/unix/main.c
@@ -627,6 +627,10 @@ MP_NOINLINE int main_(int argc, char **argv) {
     sys_set_excecutable(argv[0]);
     #endif
 
+    #ifdef __AFL_HAVE_MANUAL_CONTROL
+    __AFL_INIT();
+    #endif
+
     const int NOTHING_EXECUTED = -2;
     int ret = NOTHING_EXECUTED;
     bool inspect = false;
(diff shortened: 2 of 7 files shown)


jepler:
Are crashes in `@micropython.native` decorated functions of interest? If not, I'll disable it in this variant.

This crashes, but would be an UnboundLocalError in standard Python:

```py
@micropython.native
def f1(n):
    for i in range(i):
        print(i)


f1(4)
```

`@micropython.viper` looks like another license to crash, I'll disable it.  Here's one that looks more interesting than accessing arbitrary memory, though:

```py
@micropython.viper
def f(dest_in):
    6(dest_in)
```

This hits an assertion error:
```
micropython: ../../py/emitnative.c:2846: void emit_native_call_function(emit_t *, mp_uint_t, mp_uint_t, mp_uint_t): Assertion `vtype_fun == VTYPE_PYOBJ' failed.
```

There are more assertions of that type to hit:
```
micropython: ../../py/emitnative.c:2747: void emit_native_unpack_sequence(emit_t *, mp_uint_t): Assertion `vtype_base == VTYPE_PYOBJ' failed.
micropython: ../../py/emitnative.c:1497: void emit_native_load_attr(emit_t *, qstr): Assertion `vtype_base == VTYPE_PYOBJ' failed.
micropython: ../../py/emitnative.c:1497: void emit_native_load_attr(emit_t *, qstr): Assertion `vtype_base == VTYPE_PYOBJ' failed.
```

ports/unix/variants/aflplusplus/README.md
```diff
@@ -0,0 +1,24 @@
+# Unix Fuzzing Variant
+
+This variant is for use with the [AFLplusplus](https://github.com/AFLplusplus/AFLplusplus)
+fuzzer.
+
+ 1. Install AFLplusplus so that the program `afl-cc` is on $PATH
+ 1. `cd ports/unix && make VARIANT=aflplusplus -j$(nproc)`
+ 1. Gather your inputs (e.g., from test cases in `tests`)
```
dlech, line 8: How do you gather inputs?

Maybe this obvious to anyone who has used a fuzzer before, in which case, we don't need any more details here. 

Since I saw all of the bugs you have been posting recently, I was curious what your process was since I've never done anything like that before. So nice to see this PR to shed a little light on how to at least get started.

jepler: That's a fine question to ask. What I wrote above was pretty terse!

As you probably know by now, a fuzzer creates new test cases largely by making modifications to existing test cases according to various strategies. Then, it runs the test case in the hopes of seeing new behavior or especially a crash.

The Python files in MicroPython's `tests` directory are a great starting point, because they give the fuzzer a set of inputs that can reach a large variety of spots in the MicroPython source -- Around 98% of source code lines, according to the coverage report.

On the other hand, some tests are not good for fuzzing. For instance, it's highly desirable that the fuzzing tests are quick. Like, millisecond quick. Any test that calls "sleep" with a non-zero argument, for instance, is probably slow.

In various runs I've done things like use all tests in basics/, all tests in extmod/, etc. I've used grep with a pattern like 'sleep.*[1-9]' to find and remove scripts that sleep. Lengthy scripts and "stress" scripts are also good ideas to remove.  I've also used afl-cmin according to AFLplusplus directions to remove tests that do not add any "coverage" (e.g., if test3.py doesn't _actually_ reach any paths in the C code that test1.py and test2.py also reach, then remove test3.py entirely)

Usually an example of a crash from AFLplusplus is much bigger than it strictly needs to be. And the fuzzer might discover many variations on the same theme -- For instance, in the recent run that turned up https://github.com/micropython/micropython/issues/17815 there were probably 50+ variations and one of them was 144 lines of code, where the human-minimized version was just 2 lines. The afl-tmin program from AFLplusplus can also minimize a test case, but I haven't been using it lately.

jepler:
@dpgeorge is this something that might be mergeable or should I do something else with it like a newer personal blog write-up?

jepler:
Closing due to apparent lack of interest.

PR #17826, opened 2025-08-03
docs/esp32: Use new esptool syntax.

nmattia:
### Summary

This updates references `esptool` arguments to use kebab-case (`write-flash`) instead of snake_case (`write_flash`). The latter seems deprecated and shows warnings.

```bash
$ esptool write_flash
Warning: Deprecated: Command 'write_flash' is deprecated. Use 'write-flash' instead.
esptool v5.0.1
```

### DIFF
--- a/docs/esp8266/tutorial/intro.rst
+++ b/docs/esp8266/tutorial/intro.rst
@@ -97,11 +97,11 @@ to the documentation for your board to see its recommendations.
 
 Using esptool.py you can erase the flash with the command::
 
-    esptool.py --port /dev/ttyUSB0 erase_flash
+    esptool.py --port /dev/ttyUSB0 erase-flash
 
 And then deploy the new firmware using::
 
-    esptool.py --port /dev/ttyUSB0 --baud 460800 write_flash --flash_size=detect 0 esp8266-20170108-v1.8.7.bin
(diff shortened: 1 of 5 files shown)


dpgeorge:
From what I can tell, the IDF is still using `esptool.py` v4.x which does not support the new naming style.  So I think we need to keep the old style around for some time.

nmattia:
I see! The instructions however suggest installing it from source or with `pip` which will install the latest version, I believe that's why I ran into the issue

dpgeorge:
> The instructions however suggest installing it from source or with `pip` which will install the latest version

Yes, indeed they do.

But if we change it to use the new syntax, it won't work at all with old versions (< version 5) or with the esptool that is installed by the current version of ESP32 IDF that we use.

The only thing I could suggest at this stage is to put a note that newer versions will give a warning unless you use the newer syntax.

nmattia:
> put a note that newer versions will give a warning unless you use the newer syntax.

Will do!

> it won't work at all with old versions (< version 5)

I _think_ in one case I got an error in the new version when using the old syntax, but I can't remember where (it was a while back). I'll try to reproduce but either way will just add a note for now.

dpgeorge:
Closing due to inactivity.

PR #17830, opened 2025-08-04
websocket: Implement 64-bit frames, enable split frames.

jepler:
### Summary

This fixes several assertion errors that were found in fuzz testing, for unimplemented portions of the websocket spec. The assertions were either turned into Python exceptions, or the missing functionality was implemented.

Frames larger than 4GB are unsupported. Initial reception of such a frame will result in OSError(EIO) and subsequent operations on the same websocket will result in OSError(ECONNRESET), because framing has been lost.

Transmitting frames larger than 64kB is unsupported. Attempting to transmit such a frame will result in OSError(ENOBUFS). Subsequent operations on the websocket are possible.

### Testing

I added tests for the new & updated functionality and endorsed the output because I think it's correct.

This will increase code size somewhat. There are other alternatives that would have less of a code size impact (e.g., not even partially implementing the 64 bit size option) while still turning assertion failures into proper runtime errors. This might be a better choice, as I assume the current websocket implementation is actually useful talking to websocket servers in the wild.


jepler:
Locally, excluding any support for 64-bit frame payload sizes saves back 36 bytes on RPI_PICO_W.

jepler:
Closing in favor of #18211.

PR #17831, opened 2025-08-04
unix/modsocket: Don't crash on invalid argument type.

jepler:
### Summary

Fuzz testing found crashes in the unix socket implementation when wrong argument types were passed in, because the argument type was an assertion, not a runtime check.

### Testing

An added test passes with the updated code.

### DIFF
--- a/tests/extmod/socket_badconstructor.py
+++ b/tests/extmod/socket_badconstructor.py
@@ -1,0 +1,20 @@
+try:
+    import socket
+except:
+    print("SKIP")
+    raise SystemExit
+
+try:
+    s = socket(None)
+except TypeError:
+    print("TypeError")
+
+try:
+    s = socket(socket.AF_INET, None)
+except TypeError:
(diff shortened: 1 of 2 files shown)


dpgeorge (approved):
Looks good, thanks.

dpgeorge:
Rebased and merged in e9da4c9c98c329c85eb7273437ff9e567d6cd006

tests/extmod/socket_badconstructor.py
```diff
@@ -0,0 +1,20 @@
+try:
+    import socket
+except:
+    print("SKIP")
+    raise SystemExit
+
+try:
+    s = socket(None)
```
dpgeorge, line 8: I fixed this to be `socket.socket(None)` during merge.  Same with the 2 calls below.

jepler: Thanks!

PR #17836, opened 2025-08-05
esp32: Support building against IDFv5.5.

DvdGiessen:
### Summary

These changes allow building the ESP32 port against IDFv5.5.

### Testing

The build works, I've been using it on some of my custom boards, and I ran the unit tests on ESP32-S3 and ESP32-C3 GENERIC boards, as well as the BLE and WLAN multi-tests.

Also tested these changes do not break anything (build or tests) when building against IDFv5.4.2.


dpgeorge (approved):
Thanks.  These changes look harmless for v5.4.2.  And always good to have support for newer IDF versions.

PR #17837, opened 2025-08-05
tools/codeformat.py: Print filename + linenumber when dedenting fails.

DvdGiessen:
When the code formatter doesn't like what you've done with the preprocessor statements, it previously crashed with a `IndexError: list index out of range`, which isn't very helpful for figuring out where the offending code is.

Example: The formatted fails for https://github.com/micropython/micropython/commit/da5dac7abfb5f55bd59f3073fd9e4631d6d47658 ([GitHub action output](https://github.com/micropython/micropython/actions/runs/16746609751/job/47406546657#step:5:1808)).

Now it prints the filename and linenumber to help you figure out what you need to fix.

Also made a slight change to make the code use `enumerate` instead of modifying the `lines` list in-place, as suggested by @dlech.

### DIFF
--- a/tools/codeformat.py
+++ b/tools/codeformat.py
@@ -92,21 +92,20 @@ def fixup_c(filename):
     # Write out file with fixups.
     with open(filename, "w", newline="") as f:
         dedent_stack = []
-        while lines:
-            # Get next line.
-            l = lines.pop(0)
-
+        for line_number, line in enumerate(lines, 1):
             # Dedent #'s to match indent of following line (not previous line).
-            m = re.match(r"( +)#(if |ifdef |ifndef |elif |else|endif)", l)
+            m = re.match(r"( +)#(if |ifdef |ifndef |elif |else|endif)", line)
             if m:
                 indent = len(m.group(1))
                 directive = m.group(2)
                 if directive in ("if ", "ifdef ", "ifndef "):
-                    l_next = lines[0]
-                    indent_next = len(re.match(r"( *)", l_next).group(1))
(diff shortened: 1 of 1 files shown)


tools/codeformat.py
```diff
@@ -92,9 +92,11 @@ def fixup_c(filename):
     # Write out file with fixups.
     with open(filename, "w", newline="") as f:
         dedent_stack = []
+        line_number = 0
         while lines:
             # Get next line.
             l = lines.pop(0)
+            line_number += 1
```
dlech:
```suggestion
        for line_number, l in enumerate(lines):
```

And `l_next = lines[0]` changes to `l_next = lines[line_number]`

tools/codeformat.py
```diff
@@ -112,6 +114,10 @@ def fixup_c(filename):
                         # This #-line does not need dedenting.
                         dedent_stack.append(-1)
                 else:
+                    if len(dedent_stack) == 0:
+                        raise IndexError(
```
dlech: I would be inclined to make a custom `CodeFormatError` since `IndexError` isn't really what the actual problem is.

tools/codeformat.py
DvdGiessen: I wanted to keep the changes to a minimum, but indeed using `enumerate` is a bit cleaner.

dlech (approved):
LGTM

dpgeorge (approved):

dpgeorge:
This looks good, thank you.

PR #17842, opened 2025-08-05
framebuf: Fix crash in scroll().

jepler:
### Summary

Fuzzing found a crash in framebuf scroll. If mp_int_t is wider than int, then the tests such as `xend < 0` can fail even when the amount of scrolling requested is out of range. This resulted in a segmentation fault when attempting an out-of-bounds access to the framebuffer.

### Testing

I added a new test for the problem. The try/except is for systems with 32-bit mp_int_t (they throw an exception with an argument this large)

Widening the other locals to mp_int_t makes no difference on platforms where it's the same width as int, which is the case for the most constrained devices. However, it might expand nanbox builds a bit in code size.

### DIFF
--- a/extmod/modframebuf.c
+++ b/extmod/modframebuf.c
@@ -787,39 +787,40 @@ static mp_obj_t framebuf_scroll(mp_obj_t self_in, mp_obj_t xstep_in, mp_obj_t ys
     mp_obj_framebuf_t *self = MP_OBJ_TO_PTR(self_in);
     mp_int_t xstep = mp_obj_get_int(xstep_in);
     mp_int_t ystep = mp_obj_get_int(ystep_in);
-    int sx, y, xend, yend, dx, dy;
+    unsigned int sx, y, xend, yend;
+    int dx, dy;
     if (xstep < 0) {
-        sx = 0;
-        xend = self->width + xstep;
-        if (xend <= 0) {
+        if (-xstep >= self->width) {
             return mp_const_none;
         }
+        sx = 0;
+        xend = self->width + (int)xstep;
         dx = 1;
     } else {
-        sx = self->width - 1;
-        xend = xstep - 1;
-        if (xend >= sx) {
+        if (xstep >= self->width) {
             return mp_const_none;
         }
+        sx = self->width - 1;
+        xend = (int)xstep - 1;
         dx = -1;
     }
     if (ystep < 0) {
-        y = 0;
-        yend = self->height + ystep;
-        if (yend <= 0) {
+        if (-ystep >= self->height) {
             return mp_const_none;
         }
+        y = 0;
+        yend = self->height + (int)ystep;
         dy = 1;
     } else {
-        y = self->height - 1;
-        yend = ystep - 1;
-        if (yend >= y) {
+        if (ystep >= self->height) {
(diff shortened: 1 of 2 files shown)


dpgeorge:
Would be good to get the code size increase for unix x64 down to zero if possible.

Maybe worth thinking a bit about the types needed here:
- `dx` and `dy` can probably stay as `int`
- `sx` and `y` can probably be `unsigned int`, they are guaranteed to be non-negative
- maybe `x` (in the for loop) can be `unsigned int` as well, at least that matches the argument type for `getpixel` and `setpixel`.

jepler:
Updated. The best variant I could find locally was +3 bytes in unix standard build on x86_64. Let's see what CI says.

jepler:
I could pay for the size increase with this, unless the check was previously inlined for speed reasons:
```diff
diff --git a/extmod/modframebuf.c b/extmod/modframebuf.c
index 16606109cc..182e170191 100644
--- a/extmod/modframebuf.c
+++ b/extmod/modframebuf.c
@@ -467,13 +467,9 @@ static void line(const mp_obj_framebuf_t *fb, mp_int_t x1, mp_int_t y1, mp_int_t
     mp_int_t e = 2 * dy - dx;
     for (mp_int_t i = 0; i < dx; ++i) {
         if (steep) {
-            if (0 <= y1 && y1 < fb->width && 0 <= x1 && x1 < fb->height) {
-                setpixel(fb, y1, x1, col);
-            }
+            setpixel_checked(fb, y1, x1, col, 1);
         } else {
-            if (0 <= x1 && x1 < fb->width && 0 <= y1 && y1 < fb->height) {
-                setpixel(fb, x1, y1, col);
-            }
+            setpixel_checked(fb, x1, y1, col, 1);
         }
         while (e >= 0) {
             y1 += sy;
@@ -483,9 +479,7 @@ static void line(const mp_obj_framebuf_t *fb, mp_int_t x1, mp_int_t y1, mp_int_t
         e += 2 * dy;
     }
 
-    if (0 <= x2 && x2 < fb->width && 0 <= y2 && y2 < fb->height) {
-        setpixel(fb, x2, y2, col);
-    }
+    setpixel_checked(fb, x2, y2, col, 1);
 }
 
 static mp_obj_t framebuf_line(size_t n_args, const mp_obj_t *args_in) {
```

dpgeorge:
> I could pay for the size increase with this, unless the check was previously inlined for speed reasons:

Hmm, interesting!

I'm not sure about the speed.

But, maybe just changing the call at the end of the function to use `setpixel_checked()` would be enough to gain most of the code savings, and then we don't need to think about/measure the speed impact of the other calls?

jepler:
> But, maybe just changing the call at the end of the function to use `setpixel_checked()` would be enough to gain most of the code savings, and then we don't need to think about/measure the speed impact of the other calls?

OK, trying that in the newest build.

dpgeorge:
This is looking good now, a nice little win overall in terms of code size.

dpgeorge (approved):

dpgeorge:
Rebased and merged in e15219800e4874a6c2653cfbdd0553c37610aeff and 803da9645f735edd3d1a80a3f920039e0f0298f2

PR #17844, opened 2025-08-06
esp32: Fix UART rxidle and machine.Timer on ESP32-C2,C6

projectgus:
### Summary

* Fixes a regression in ESP32-C6 when updating to IDF v5.4.1 - the 80MHz PLL is no longer enabled by default, so the RXIDLE UART interrupt never fires. Found via `tests/extmod_hardware/machine_uart_irq_rxidle.py`.
* Adds a related fix, which is that the ESP32 `machine.Timer` implementation assumed the default timer source clock was the APB clock but it's a different clock on some chips, leading to incorrect timer periods (ESP32-C2 used 26MHz but should be 40MHz, ESP32-C6 used 40MHz but should be 80MHz).

### Testing

* Ran the `extmod_hardware/machine_uart_*.py` tests on ESP32-C2,C3,C6,S2 and original. All passing now.
* Ran a homemade periodic 1 second `machine.Timer` test on ESP32-C2,C6 and S2. Verified incorrect tick period on C2 and C6 that is fixed with this PR. Will submit a proper unit test for this, later.

(Note: Some of these tests were run on the first version of this PR. I re-ran the latest version on C2, C6 and original ESP32.)

* Calling `esp_clk_tree_src_get_freq_hz()` to get a constant clock frequency seems inelegant compared to calculating it at compile time, but this isn't called very often (initially I thought it was being called from the ISR, which made it seem worse, but it's not!)

*This work was funded through GitHub sponsors*

### DIFF
--- a/ports/esp32/machine_timer.h
+++ b/ports/esp32/machine_timer.h
@@ -34,13 +34,6 @@
 #include "hal/timer_ll.h"
 #include "soc/timer_periph.h"
 
-#define TIMER_DIVIDER  8
-
-// TIMER_BASE_CLK is normally 80MHz. TIMER_DIVIDER ought to divide this exactly
-#define TIMER_SCALE    (APB_CLK_FREQ / TIMER_DIVIDER)
-
-#define TIMER_FLAGS    0
-
 typedef struct _machine_timer_obj_t {
     mp_obj_base_t base;
 
@@ -63,5 +56,7 @@ typedef struct _machine_timer_obj_t {
 machine_timer_obj_t *machine_timer_create(mp_uint_t timer);
 void machine_timer_enable(machine_timer_obj_t *self);
 void machine_timer_disable(machine_timer_obj_t *self);
+
+uint32_t machine_timer_freq_hz(void);
 
 #endif // MICROPY_INCLUDED_ESP32_MACHINE_TIMER_H
--- a/ports/esp32/machine_uart.c
+++ b/ports/esp32/machine_uart.c
@@ -58,7 +58,7 @@
 #define UART_IRQ_RXIDLE (0x1000)
 #define UART_IRQ_BREAK (1 << UART_BREAK)
 #define MP_UART_ALLOWED_FLAGS (UART_IRQ_RX | UART_IRQ_RXIDLE | UART_IRQ_BREAK)
-#define RXIDLE_TIMER_MIN (5000)  // 500 us
+#define RXIDLE_TIMER_MIN (machine_timer_freq_hz() * 5 / 10000) // 500us minimum rxidle time
 #define UART_QUEUE_SIZE (3)
 
 enum {
@@ -535,7 +535,7 @@ static void uart_irq_configure_timer(machine_uart_obj_t *self, mp_uint_t trigger
         self->mp_irq_obj->ishard = false;
         uint32_t baudrate;
         uart_get_baudrate(self->uart_num, &baudrate);
-        mp_int_t period = TIMER_SCALE * 20 / baudrate + 1;
+        mp_int_t period = machine_timer_freq_hz() * 20 / baudrate + 1;
         if (period < RXIDLE_TIMER_MIN) {
             period = RXIDLE_TIMER_MIN;
         }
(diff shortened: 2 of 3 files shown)


projectgus:
@dpgeorge While testing this I reproduced a hard fault which might be the same one that you were seeing. I think that's a pre-existing bug not a regression, so I'll submit in a second PR that could be left until after the 1.26 release.

ports/esp32/machine_timer.h
```diff
@@ -30,16 +30,46 @@
...
-#define TIMER_FLAGS    0
+// Returns the default timer tick frequency
+// (i.e. frequency of TIMER_CLK_SRC / TIMER_DIVIDER)
+//
+// This function should evaluate at compile time, or produce a fatal warning if
+// there's a new chip with different default clock source.
+//
+// This is a hacky way to avoid calling esp_clk_tree_src_get_freq_hz() from the
+// timer interrupt (as it's slow, and these clocks are all constant frequency.)
+inline static uint32_t machine_timer_freq_hz(void) {
```
dpgeorge: Should be `static inline` for consistency.

ports/esp32/machine_uart.c
```diff
@@ -58,7 +58,7 @@
 #define UART_IRQ_RXIDLE (0x1000)
 #define UART_IRQ_BREAK (1 << UART_BREAK)
 #define MP_UART_ALLOWED_FLAGS (UART_IRQ_RX | UART_IRQ_RXIDLE | UART_IRQ_BREAK)
-#define RXIDLE_TIMER_MIN (5000)  // 500 us
+#define RXIDLE_TIMER_MIN (machine_timer_freq_hz() * 500 / 1000000) // 500us minimum rxidle time
```
dpgeorge: Could this overflow?  If it uses the return type of `machine_timer_freq_hz()` which is `uint32_t`, and the frequency is 10MHz, then 500*10MHz will overflow 32 bits.

Maybe do `machine_timer_freq_hz() * 5 / 10000`?

dpgeorge:
Thanks for debugging and fixing this so quickly!

I tested on ESP32-C6 and it does indeed fix the `uart_irq_rxidle` test.  I'll need to eventually run the full test suite on C6 and some other SoCs, but this does look good.

> 2\. Call `esp_clk_tree_get_src_clk_freq()` at runtime - which will work to return the same (constant) value, but currently the timer implementation would call this from the ISR and it seems a bit heavyweight for that.

How about calling `esp_clk_tree_src_get_freq_hz()` at start-up and caching the result in a variable?  Then could use that variable directly, instead of calling a function.

ports/esp32/machine_uart.c
projectgus: :facepalm: Thanks!

projectgus:
> How about calling `esp_clk_tree_src_get_freq_hz()` at start-up and caching the result in a variable? Then could use that variable directly, instead of calling a function.

Yes, this is better - have changed (and re-tested on OG ESP32 and C6, same timer results.) I got a bit hung up on "this should be compile-time calculated!" :laughing: 

Possibly caching it isn't even worth it, it's not a particularly expensive function call to make relative to the frequency of the timer ISR. (But, then again, caching it only costs 4 bytes of RAM so it's not a big trade-off either way.)

projectgus:
> Possibly caching it isn't even worth it, it's not a particularly expensive function call to make relative to the frequency of the timer ISR. (But, then again, caching it only costs 4 bytes of RAM so it's not a big trade-off either way.)

Uh, I looked again and it doesn't call this function from the Timer ISR anyway (in retrospect it's unclear why it would need to), so I've pulled out the caching as well.

ports/esp32/machine_uart.c
dpgeorge: I'm actually curious how this code worked with this overflow?  On a board with 10MHz timer freq, the calculation here would have resulted in 705 instead of 5000.

dpgeorge: Oh, I guess it's because it's a minimum, and that minimum was lower so not being triggered by the if logic.

dpgeorge:
I retested this on ESP32 and ESP32-C6, and the UART IRQ tests pass.

dpgeorge (approved):
Tested again with the full test suite on ESP32 and ESP32-C6.  There are no new failures, and `extmod_hardware/machine_uart_irq_rxidle.py` now passes on C6.

dpgeorge:
Now merged.  Thanks again for the fast turn around on this!

PR #17849, opened 2025-08-06
ringio: Detect incorrect constructor calls.

### LINKED ISSUE #17847: ringbuf: Zero-length ringbuf creation allowed, causes crash

unix port, coverage variant, x86_64 linux

### MicroPython version

MicroPython v1.26.0-preview.524.g255d74b5a8 on 2025-08-06; linux [GCC 12.2.0] version

### Reproduction

Run the following script:
```
import micropython
rb = micropython.RingIO(bytearray(0))
rb.write(b'\1')
```

### Expected behaviour

An exception is thown on line 2, because a zero-byte ring buffer is nonsense.

### Observed behaviour

Division by zero (reported by Linux as SIGFPE) occurs on line 3:
```
>>> micropython.RingIO(bytearray(0)).write(b'\1')

Program received signal SIGFPE, Arithmetic exception.
0x0000555555605f33 in ringbuf_free (r=0x7ffff7c318a8) at ../../py/ringbuf.h:86
86	    return (r->size + r->iget - r->iput - 1) % r->size;
```

### Additional Information

Found via fuzzing, manually minimized.

### Code of Conduct

Yes, I agree

jepler:
### Summary
Buffer size must be at least 2, and is a 16-bit quantity.

This fixes several cases including the one the fuzzer discovered, which would lead to a fatal signal when accessing the object.

Closes: #17847

### Testing

I added 5 new test cases for invalid constructor calls.

I suspect that the re-organization saves a little code size, but I could always be wrong. It at least allowed to have a single path for the new size checks.


jepler:
I'm quite surprised by the code size delta on this one.

andrewleech:
Thanks for the cleanup here, it's a really good catch!

I too am surprised at the code size increase... to be fair I said the same thing about the original MR to add this feature; adding the functions to expose something to python is surpsiringly large, however yeah +144 for just this re-organise and exception raise does seem strange.

I did go through a few iterations of the function originally to try to get the compiled size down, though I wouldn't have thought it's that sensitive.

dpgeorge:
There's something strange going on with the code size increase.  It doesn't increase at all on rp2, which does have this feature enabled.

Maybe the new code is bringing in a 64-bit arithmetic function?  Or inlining a call to memset/memcpy?

jepler:
The code delta is due to calling `mp_obj_get_uint` which was apparently not otherwise called on these builds.

tests/micropython/ringio.py
```diff
@@ -46,3 +46,33 @@
...
+
+try:
+    # Size may not be too small
+    micropython.RingIO(0)
+except ValueError as ex:
+    print(type(ex))
+
+try:
+    # Buffer may not be too big
+    micropython.RingIO(bytearray(65536))
```
dpgeorge: This is a very large buffer to allocate, and will probably make this test fail on ports with a small heap (or even with a large heap but fragmented).

Probably the easiest thing to do here is just remove this part of the test (unless you can think of a better way to do it?).  The "too large" path is already covered by passing in a large integer.

jepler: to save code the `RingIO(65535)` path is also allocating the whole buffer. So it might fail with a MemoryError instead as well.

should this be split to a separate test so we can raise SKIP if a bytearray of 65536 can't be allocated?

dpgeorge: > to save code the `RingIO(65535)` path is also allocating the whole buffer

Ah, right.



> should this be split to a separate test so we can raise SKIP if a bytearray of 65536 can't be allocated?

Yes, it will need to be separated out.  Then skip if any of the `RingIO()` calls raise `MemoryError`.

jepler: I split those tests to a new "ringio_big.py" test and added tests that the maximum expected size can be constructed.

tests/micropython/ringio_big.py
```diff
@@ -0,0 +1,32 @@
...
+
+try:
+    micropython.RingIO
+except AttributeError:
+    print("SKIP")
+    raise SystemExit
+
+try:
+    # Ensure the heap is big enough for the test.
+    bytearray(70000)
```
dpgeorge: I'm not sure this is going to work.

It could pass, yet the below allocations fail, due to the GC being conservative.  Or due to fragmentation that happens after this allocation succeeds.

I suggest just wrapping the whole thing below in try/except MemoryError and skipping if any of the parts fail.  That's not ideal, but the main thing here is to pass to get coverage, which will always be the case on the unix port.

jepler: updated again. is this what you're looking for?

dpgeorge: Yes, thanks, that should be robust across all targets.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 0615d13963ba33650a5b2070368584c7aec9c9dd

Issue #17732, opened 2025-07-21
rp2: Can't send KeyboardInterrupt with os.dupterm()


sfe-SparkFro:
### Port, board and/or hardware

`rp2`

### MicroPython version

`MicroPython v1.25.0 on 2025-04-15; Raspberry Pi Pico 2 W with RP2350`

### Reproduction

Call `os.dupterm()` with UART or other compatible stream object:

```python
from machine import UART
import os
os.dupterm(UART(0), 0)
```

Run some code:

```python
import time
i = 0
while True:
    print(i)
    i += 1
    time.sleep(1)
```

### Expected behaviour

While code is running, pressing `Ctrl+C` in the USB REPL works as expected:

```
0
1
2
Traceback (most recent call last):
  File "<stdin>", line 6, in <module>
KeyboardInterrupt:
>>>
```

### Observed behaviour

While code is running, pressing `Ctrl+C` in the UART REPL fails to send the `KeyboardInterrupt` and halt execution.

Below, I pressed `Ctrl+C` a few times in the UART REPL, then once in the USB REPL. It then output extra `>>>`'s for each time I hit `Ctrl+C` in the UART REPL, indicating the `KeyboardInterrupt`'s were getting buffered or something?

```
0
1
2
3
Traceback (most recent call last):
  File "<stdin>", line 6, in <module>
KeyboardInterrupt:
>>>
>>>
>>>
>>>
>>>
```

### Additional Information

Although the provided replication code uses UART, this also happens with a Bluetooth REPL connection and is actually what we need to be fixed. This was just the simplest way to replicate.

### Code of Conduct

Yes, I agree

felixdoerre:
For interrupts to work, you need to set up your stream to receive interrupts. By default micropython only reads from the UART stream, when the repl is idle, and while the command is executing no-one reads from the stream, so any received bytes are just buffered in the UART receive buffer.

For webrepl there is a mechanism implemented to react to Ctrl+C: `dupterm_notify`, see here:

https://github.com/micropython/micropython-lib/blob/master/micropython/net/webrepl/webrepl.py#L134

So what I believe you need to do is to use that, and plug it into `UART.irq`, so when a byte arrives, it is read, checked for an interrupt, and then buffered by the repl while the code is running.

sfe-SparkFro:
[...]

PR #17850, opened 2025-08-06
docs/library/os: Document dupterm_notify function.

Josverl:
### Summary

Add documentation for `os.dupterm_notify()`
Based on the information in 

https://github.com/micropython/micropython/issues/17732
and 
https://github.com/micropython/micropython/issues/17732
and the source code.

### Testing

n/a

### DIFF
--- a/docs/library/os.rst
+++ b/docs/library/os.rst
@@ -133,6 +133,37 @@ Terminal redirection and duplication
 
    The function returns the previous stream-like object in the given slot.
 
+.. function:: dupterm_notify(obj_in, /)
+
+    Notify the MicroPython REPL that input is available on a stream-like object
+    previously registered via `os.dupterm()`.
+
+    This function should be called by custom stream implementations (e.g., UART,
+    Bluetooth, or other non-USB REPL streams) to inform the REPL that input is
+    ready to be read. Proper use ensures that special characters such as
+    Ctrl+C (used to trigger KeyboardInterrupt) are processed promptly by the
+    REPL, enabling expected interruption behavior for user code.
+
+    The *obj_in* parameter is ignored by `os.dupterm_notify()`, but is required to allow calling
+    dupterm_notify from an interrupt handler such as `UART.irq()`.
+
+    Example:
+
+    .. code-block:: python
+
+        from machine import UART
+        import os
+        uart = UART(0)
+        os.dupterm(uart, 0)
+        uart.irq(os.dupterm_notify, machine.UART.IRQ_RX)
+
+    .. note::
+        If the ``dupterm_notify()`` function is not called, input from the custom stream
+        may not be detected or processed until the next REPL poll, potentially delaying
+        KeyboardInterrupts or other control signals.
+        This is especially important for UART, Bluetooth and other
+        non-standard REPL connections, where automatic notification is not guaranteed.
+
 Filesystem mounting
 -------------------
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,8 +13,7 @@ Building the documentation locally
 If you're making changes to the documentation, you may want to build the
 documentation locally so that you can preview your changes.
 
-Install Sphinx, and optionally (for the RTD-styling), sphinx_rtd_theme,
-preferably in a virtualenv:
+Install Sphinx and sphinx_rtd_theme, preferably in a virtualenv:
 
      pip install sphinx
      pip install sphinx_rtd_theme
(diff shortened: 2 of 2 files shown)


docs/library/os.rst
```diff
@@ -133,6 +133,37 @@ Terminal redirection and duplication
 
    The function returns the previous stream-like object in the given slot.
 
+.. function:: dupterm_notify(obj_in, /)
+    Notify the MicroPython REPL that input is available on a stream-like object
```
dpgeorge, line 138: Please put a blank line between the function decl and the first paragraph here.

docs/library/os.rst
```diff
@@ -133,6 +133,37 @@ Terminal redirection and duplication
...
+    previously registered via `os.dupterm()`.
+
+    This function should be called by custom stream implementations (e.g., UART,
+    Bluetooth, or other non-USB REPL streams) to inform the REPL that input is
+    ready to be read. Proper use ensures that special characters such as
+    Ctrl+C (used to trigger KeyboardInterrupt) are processed promptly by the
+    REPL, enabling expected interruption behavior for user code.
+
+    Args:
+        obj_in: Is ignored by `dupterm_notify`, but is required to allow calling
```
dpgeorge: Since there is only 1 argument I suggest writing it like: "The *obj_in* parameter is ignored ...".

Also put `*` around parameter names.

docs/library/os.rst
```diff
@@ -133,6 +133,37 @@ Terminal redirection and duplication
...
+        dupterm_notify from an interrupt handler such as ``UART.irq()``.
+
+    Note:
+        - If input is available (including control characters like Ctrl+C),
+          call this function to ensure responsive REPL behavior.
+        - If omitted, input from the custom stream may not be detected or
+          processed until the next REPL poll, potentially delaying KeyboardInterrupts
+          or other control signals.
+        - This is especially important for UART, Bluetooth, or other
+          non-standard REPL connections, where automatic notification is not guaranteed.
```
dpgeorge: These notes read more like a paragraph rather than dot points.  Suggest removing the first dot point (that information is already present in the paragraphs above) and combining the other two into a single paragraph.

Also "If omitted" -> "If not called".

docs/library/os.rst
```diff
@@ -133,6 +133,37 @@ Terminal redirection and duplication
...
+
+    Example::
+        from machine import UART
+        import os
+        uart = UART(0)
+        uart.irq(os.dupterm_notify, machine.UART.IRQ_RX)
+        os.dupterm(uart, 0)
```
dpgeorge, line 157: Probably better to register the IRQ after doing the `os.dupterm()`.

Josverl:
@dpgeorge, 
Adjusted as requested, and fixed additional RST formatting.

docs/library/os.rst
```diff
@@ -133,6 +133,37 @@ Terminal redirection and duplication
...
+    Notify the MicroPython REPL that input is available on a stream-like object
+    previously registered via `os.dupterm()`.
+
+    This function should be called by custom stream implementations (e.g., UART,
+    Bluetooth, or other non-USB REPL streams) to inform the REPL that input is
+    ready to be read. Proper use ensures that special characters such as
+    Ctrl+C (used to trigger KeyboardInterrupt) are processed promptly by the
+    REPL, enabling expected interruption behavior for user code.
+
+    Args:
```
dpgeorge: To match the style of the `dupterm` function above, I don't think the arguments should be in an indented paragraph like this.  Rather, just remove the "Args:" bit and make the paragraph below just a normal paragraph of this function definition.

docs/library/os.rst
```diff
@@ -133,6 +133,37 @@ Terminal redirection and duplication
...
+        The *obj_in* parameter is ignored by `os.dupterm_notify()`, but is required to allow calling
+        dupterm_notify from an interrupt handler such as `UART.irq()`.
+
+    Note:
+        If omitted, input from the custom stream may not be detected or
```
dpgeorge: As I mentioned before, I think "If omitted" is confusing.  Suggest writing it as "If this function is not called, input from ...".

Also, the "Note:" styling won't match other notes in the docs.  Either make this a normal paragraph (without the "Note:") or use the `.. note::` styling.

Josverl: Reworded, 
updated to .. note style, 
and move the note after the example.

docs/library/os.rst
Josverl: dropped Args and indent.

dpgeorge (approved):
Thanks for updating.  Looks good now.

Issue #11269, opened 2023-04-15
docs: Building embedding example using shared/static lib with cmake (with example)


shakfu:
I was recently interested to build static and shared library versions of the `micropython/examples/embedding` case using cmake and I managed to get it roughly done ( see this [public repo](https://github.com/shakfu/micropython-embed-cmake) in case anyone is interested).

Along the way I found that micropython has quite a number of cmake-related files in its repo (see below). 

Since I did not use any of these in my above example, is there any documentation guidance on building MicroPython with cmake that I can use to improve my example? Thanks!

```text
ports/esp32/CMakeLists.txt
ports/esp32/main/CMakeLists.txt
ports/rp2/CMakeLists.txt
ports/zephyr/CMakeLists.txt

AND

examples/usercmodule/cexample/micropython.cmake
examples/usercmodule/cppexample/micropython.cmake
examples/usercmodule/micropython.cmake
extmod/extmod.cmake
extmod/nimble/nimble.cmake
ports/esp32/boards/ESP32_S2_WROVER/mpconfigboard.cmake
ports/esp32/boards/GENERIC/mpconfigboard.cmake
ports/esp32/boards/GENERIC_C3/mpconfigboard.cmake
ports/esp32/boards/GENERIC_C3_USB/mpconfigboard.cmake
ports/esp32/boards/GENERIC_D2WD/mpconfigboard.cmake
ports/esp32/boards/GENERIC_OTA/mpconfigboard.cmake
ports/esp32/boards/GENERIC_S2/mpconfigboard.cmake
ports/esp32/boards/GENERIC_S3/mpconfigboard.cmake
ports/esp32/boards/GENERIC_S3_SPIRAM/mpconfigboard.cmake
ports/esp32/boards/GENERIC_S3_SPIRAM_OCT/mpconfigboard.cmake
ports/esp32/boards/GENERIC_SPIRAM/mpconfigboard.cmake
ports/esp32/boards/GENERIC_UNICORE/mpconfigboard.cmake
ports/esp32/boards/LILYGO_TTGO_LORA32/mpconfigboard.cmake
ports/esp32/boards/LOLIN_C3_MINI/mpconfigboard.cmake
ports/esp32/boards/LOLIN_S2_MINI/mpconfigboard.cmake
ports/esp32/boards/LOLIN_S2_PICO/mpconfigboard.cmake
ports/esp32/boards/M5STACK_ATOM/mpconfigboard.cmake
[...]

PR #17860, opened 2025-08-07
embed: CMake find_package support.

Gadgetoid:
This PR adds CMake support for embed, by adding a `micropython-embed-config.cmake` file, which makes the project compatible with:

```
find_package(micropython-embed CONFIG REQUIRED PATHS ${MICROPY_EMBED_DIR})
```

The user is then free to link the dynamic `micropython` library to their project in the normal CMake fashion.

Currently the variable `${MICROPY_EMBED_CONFIG_DIR}` must be set so that `mpconfigport.h` can be found. We might want to explore the nomenclature around this, and potentially also add support for `mpconfigport.cmake` customisation of the embed project- including support for USER_C_MODULES and so on.

Note: The CMake port crucially differs from the Makefile version in that it does not attempt to produce a stand-alone collection of source files for vendoring into a project. Instead it asks that you submodule MicroPython or otherwise include it into your project wholesale. It feels like it warrants a separate port to highlight this difference. But since shares a common codebase with `embed` that might not be tidy or feasible. Input welcome!

This PR answers- to some extent- https://github.com/micropython/micropython/issues/11269 by providing a complete example of using existing MicroPython CMake tooling to bring up a CMake compatible embed port.

Raised as a draft, since there's a lot of ground yet to cover and details to hash out. Input very much appreciated.

### Generative AI (Added in post, since I use AI to punch up my original proposal)

I used generative AI tools when creating this PR, but a human has checked the
code and is responsible for the code and the description above.

### DIFF
--- a/examples/embedding/.gitignore
+++ b/examples/embedding/.gitignore
@@ -2,3 +2,6 @@
 embed
 main.o
 micropython_embed
+
+# CMake build directory
+build
--- a/examples/embedding/CMakeLists.txt
+++ b/examples/embedding/CMakeLists.txt
@@ -1,0 +1,23 @@
+cmake_minimum_required(VERSION 3.13)
+
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    set(CMAKE_C_COMPILER "clang")
+    set(CMAKE_CXX_COMPILER "clang++")
+endif()
+
+# Set build type to reduce firmware size
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE MinSizeRel)
+endif()
+
+project(embed)
+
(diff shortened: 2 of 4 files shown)


Gadgetoid:
For resolving modules this probably has to be a *static* library rather than a *shared* one, which is more functionally equivalent to what the Makefile approach is doing. This means that library code - ie: MicroPython and all extmods and usermods - can access the HAL defined in user code.

It sounds redundant to support "USER_C_MODULES", but it can make adding existing modules (such as ulab) easier.

Edit: I think I'm stretching the limits of what constitutes an "embed" port, and what constitutes a more complete port here. I'm aiming for something in between. While that may be achievable it's probably a bit more complex than we'd want a simple CMake variant of "embed" to be.

Gadgetoid:
Closing because this was incomplete, I ended up not actually using it, and it seems to have garnered no other interest in the interim. I think this is better left as busywork for someone who has a real need.

PR #17861, opened 2025-08-07
zephyr: Read erase block size from dts instead of using fixed value.

ant9000:
### Summary

On my Apollo3 board, which is supported by Zephyr, I can create a filesystem in flash but cannot get Micropython automount to work.

I've tracked the issue down to the fact that not all boards use the 4096 byte erase block size  that MP is assuming in `vfs_init()`. 

My patch extracts the correct configuration value directly from the devicetree, so it should work out of the box for every Zephyr board (with a flash).

### Testing

With my modification filesystem automount works correctly for my board, which needs 8192 as erase block size as specified in the devicetree configuration.

I've checked with a script, and AFAICT all Zephyr boards with a `zephyr,flash` compatible node have the required property explicitly defined in the devicetree.

### DIFF
--- a/ports/zephyr/main.c
+++ b/ports/zephyr/main.c
@@ -112,7 +112,9 @@ static void vfs_init(void) {
     mount_point_str = "/sd";
     path_lib_qstr = MP_QSTR__slash_sd_slash_lib;
     #elif defined(CONFIG_FLASH_MAP) && FIXED_PARTITION_EXISTS(storage_partition)
-    int block_size = DT_PROP(DT_GPARENT(DT_NODELABEL(storage_partition)), erase_block_size);
+    #define STORAGE_DEVICE   DT_GPARENT(DT_NODELABEL(storage_partition))
+    #define ERASE_BLOCK_SIZE COND_CODE_1(DT_NODE_HAS_PROP(STORAGE_DEVICE, erase_block_size), (DT_PROP(STORAGE_DEVICE, erase_block_size)), (4096))
+    int block_size = ERASE_BLOCK_SIZE;
     mp_obj_t args[] = { MP_OBJ_NEW_SMALL_INT(FIXED_PARTITION_ID(storage_partition)), MP_OBJ_NEW_SMALL_INT(block_size) };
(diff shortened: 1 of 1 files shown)


dpgeorge:
This is a nice improvement, but I found at least two boards that no longer build: `beagleconnect_freedom` and `nrf9151dk`.

Is it possible to only use this DT setting if it exists?

ant9000:
> Is it possible to only use this DT setting if it exists?

I've added a check: if `erase_block_size` property is not defined, the code will use 4096 as default (the original fixed value). Now MP for both `beagleconnect_freedom` and `nrf9151dk` boards does compile correctly.

dpgeorge:
Thanks for updating.

This is actually going to clash badly with #18185, which moves all this logic into Python.  Both this PR and PR #18185 are needed, but #18185 should go in first because that's more of a structural change.  Then we need to consider how to implement this one.

My suggestion: make the `block_size` parameter to the `FlashArea` constructor optional, and if it's not given then default to either the value from dts, or if that doesn't exist then fallback to 4096.

bikeNomad:
In #18185 I've re-added the original `vfs_init()` code but only for the case where `MICROPY_MODULE_FROZEN_MPY` is not set. This PR could be appliied to that code, but it doesn't cover the case where _boot.py is doing the mounting. 

I've added a `TODO` comment in `_boot.py` to remind us.

Perhaps `FlashArea.areas` could hold `(index, block_size)` tuples instead of just `index`.

@VynDragon @ant9000 what do you think about this?

ant9000:
> Perhaps `FlashArea.areas` could hold `(index, block_size)` tuples instead of just `index`.

I second this. Since the block size can be provided via devicetree, having it in FlashArea looks like the most sensible way to go.

bikeNomad:
I just added logic to my PR #18185 to change `FlashArea.areas` to hold `(ID, block_size)` tuples, and used them in the filesystem creation.

@ant9000 please give this a try on your board.

dpgeorge:
Closing in favour of #18185.

Issue #17855, opened 2025-08-07
Crash printing exception detail when source code is not valid UTF-8


jepler:
### Port, board and/or hardware

unix port, coverage build, x86_64 linux

### MicroPython version

MicroPython v1.26.0-preview.524.g255d74b5a8 on 2025-08-06; linux [GCC 12.2.0] version

### Reproduction

```py
# Slight changes (like removing the derived exception type) move the misbehavior
# around. For instance, in my local build, not having this triggers
# 'NotImplementedError: opcode' instead.
class Dummy(BaseException):
    pass

# Smuggle invalid UTF-8 string into decompress_error_text_maybe
# This invalid UTF-8 string acts matches the test MP_IS_COMPRESSED_ROM_STRING
# This can also happen if the input file is not a valid UTF-8 file.
b = eval(b"'\xff" + b"\xfe" * 4096 + b"'")
try:
    raise BaseException(b)
except BaseException as good:
    print(type(good), good.args[0])
```

### Expected behaviour

CPython fails the eval() with `SyntaxError: (unicode error) 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte`.

### Observed behaviour

The invalid utf-8 string can be successfully created. When fetching the exception's `args` property, a crash occurs inside of `mp_decompress_rom_string` (which should never have been called). The call occurs because the first byte of the invalid UTF-8 string is \xff, the marker for compressed ROM strings.

### Additional Information

Found via fuzzer, manually minimized.

### Code of Conduct

Yes, I agree

jepler:
The problem can also occur when the original program text is not valid UTF-8, which is what the fuzzer generated. The use of eval() was to avoid problems representing the invalid UTF-8 sequences in the inline reproducer script.

Issue #9044, opened 2022-08-12
various ways to read() don't check for valid utf-8, weirdness ensues


jepler:
[Initial comment heavily revised because the original was unclear. Apologies if this breaks any of the context of the replies below]

Over in https://github.com/adafruit/circuitpython/issues/6752 a user reported that various ways to read from files didn't actually validate that the content was utf-8. Instead, it blithely stores the data in a `str`, which in Micropython must always be a valid utf-8 encoded chunk of memory.

This leads to weird behavior, such as when the file with the byte content `\xf7` is read:
```python
>>> with open("notutf8.txt", "rt", -1, "utf-8") as f: content = f.read()
... 
>>> f
<io.TextIOWrapper -1>
>>> content
'\x07'
>>> content[0]
'\x07\x00\x00\x00'
>>> 
```

The behavior is the same no matter what coding is specified, or if no coding is explicitly specified. As discussed below, it does not appear to be explicit anywhere that the encoding of a text file is utf-8. It is implicit in the way that reading a file with UTF-8 content works, and reading a file with non-UTF8 content yields up dragons.

I prepared a possible fix relative to CircuitPython at https://github.com/adafruit/circuitpython/pull/6754 -- is there interest in changing this behavior in micropython? if so, I can prepare a PR for y'all including hopefully adding tests.

basically, each code path in file reading will now check that the content is valid utf-8 in the case that a str, rather than a bytes, is being produced.

robert-hh:
IMHO it is not valid to assume that a file content always has to be utf-8 encoded. It may be anything. So such a check would only be reasonable if the file is opened requiring a specific encoding, like utf-8.

jimmo:
My interpretation of https://docs.python.org/3/library/functions.html#open is that by not specifying an encoding to `open()` we get the default encoding. We only support one encoding, so it makes sense that we're implicitly opening the file in utf-8 mode, as as such it does make sense that we should be doing the validation if it's a text-mode stream.

jepler:
Weird stuff starts happening when a str object has non-utf8 content. The original poster over in circuitpython found, for instance, that this value doesn't subscript right ... here, "notutf8.txt" contains the single byte \xf7:
```
>>> with open("notutf8.txt") as f: content = f.read()
... 
>>> content
'\x07'
>>> content[0]
'\x07\x00\x00\x00'
>>> ord(content[0])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

Issue #13084, opened 2023-11-28
formatting character values >= 128 gives unexpected results


jepler:
```
$ ./build-coverage/micropython
MicroPython 9c7067d9ad on 2023-11-28; linux [GCC 12.2.0] version
Use Ctrl-D to exit, Ctrl-E for paste mode
>>> s = f"{160:c}"
>>> len(s)
0
>>> s
' '
>>> print(s)
�
```

The same for `s = "%c" % 160`.

I think this is because the `'c'` formatter in objstr.c doesn't handle non-ASCII characters properly. It looks like this ends up being another way to get improper UTF-8 into a str() object, too.

dpgeorge:
Can probably use a fixed (stack allocated) vstr along with vstr_add_char(), and then copy the resulting vstr into the printer used in the formatter.

jepler:
There are at least 4 code paths affected:
 * "%c" % integer
 * "%c" % string
 * "{:c}".format()
 * format's "fill character" can also be unicode

An additional complication is whether the target type is a bytes or str.

I started down the road of implementing this but things seemed to be getting out of hand and I don't actually need this so I'll just put it aside.

jepler:
micropython counts precision/field width in bytes, not code points.
```
>>> print("|{:_^5}|".format(chr(225) * 2))
|_áá__|
```
vs
```
>>> print("|{:_^5}|".format(chr(225) * 2))
|áá_|
```

As with other operations, this means you can form a string with invalid UTF8 structure.

The format padding "character" is also treated as a byte, so attempting to pad with a non-ASCII code just fails (today)
```
>>> print("|{:\xe1^5}|".format(" " * 2))
ValueError: invalid format specifier
```

vs CPython
```
|á  áá|
```

PR #17862, opened 2025-08-07
various: Don't allow creation of invalid UTF8 strings or identifiers

jepler:
### Summary

Fuzz testing found that it was possible to create invalid UTF-8 strings when the program input was not UTF-8. This could occur because a disk file was not UTF-8, or because a byte string passed to eval()/exec() was not UTF-8.

Besides leading to the problems that the introduction of `utf8_check` was intended to fix (#9044), the fuzzer found an actual crash when the first byte was `\xff` and the string was used as an exception argument (#17855).

I also noticed that the check could be generalized a little to avoid constructing non-UTF-8 identifiers, which could also lead to problems.

I re-organized the code to pay for the size cost of the new check in the lexer.

### Testing

I added a new test, using `eval()` and `exec()` of byte strings, to ensure that these cases are caught by the lexer.

Could check that the whole code buffer is UTF-8 instead.

### DIFF
--- a/py/objstr.c
+++ b/py/objstr.c
@@ -2461,7 +2461,7 @@ mp_obj_t mp_obj_new_bytes_iterator(mp_obj_t str, mp_obj_iter_buf_t *iter_buf) {
 }
 
 #if MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
-static bool mp_utf8_check(const byte *p, size_t len) {
+bool mp_utf8_check(const byte *p, size_t len) {
     uint8_t need = 0;
     const byte *end = p + len;
     for (; p < end; p++) {
--- a/py/objstr.c
+++ b/py/objstr.c
@@ -2251,6 +2251,7 @@ mp_obj_t mp_obj_new_str_via_qstr(const char *data, size_t len) {
 static mp_obj_t mp_obj_new_str_type_from_vstr(const mp_obj_type_t *type, vstr_t *vstr) {
     // if not a bytes object, look if a qstr with this data already exists
     if (type == &mp_type_str) {
+        mp_utf8_require((byte *)vstr->buf, vstr->len);
         qstr q = qstr_find_strn(vstr->buf, vstr->len);
         if (q != MP_QSTRnull) {
             vstr_clear(vstr);
--- a/py/objstr.c
+++ b/py/objstr.c
@@ -208,11 +208,7 @@ mp_obj_t mp_obj_str_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_
                 if (str_hash == 0) {
                     str_hash = qstr_compute_hash(str_data, str_len);
                 }
-                #if MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
-                if (!utf8_check(str_data, str_len)) {
-                    mp_raise_msg(&mp_type_UnicodeError, NULL);
-                }
-                #endif
+                mp_utf8_require(str_data, str_len);
 
                 // Check if a qstr with this data already exists
                 qstr q = qstr_find_strn((const char *)str_data, str_len);
@@ -2285,17 +2281,13 @@ static mp_obj_t mp_obj_new_str_type_from_vstr(const mp_obj_type_t *type, vstr_t 
 }
 
 mp_obj_t mp_obj_new_str_from_vstr(vstr_t *vstr) {
-    #if MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
-    if (!utf8_check((byte *)vstr->buf, vstr->len)) {
-        mp_raise_msg(&mp_type_UnicodeError, NULL);
-    }
-    #endif // MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
+    mp_utf8_require((byte *)vstr->buf, vstr->len);
     return mp_obj_new_str_type_from_vstr(&mp_type_str, vstr);
 }
 
 #if MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
 mp_obj_t mp_obj_new_str_from_utf8_vstr(vstr_t *vstr) {
-    // bypasses utf8_check.
+    // bypasses utf8_require.
     return mp_obj_new_str_type_from_vstr(&mp_type_str, vstr);
 }
 #endif // MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
@@ -2305,11 +2297,7 @@ mp_obj_t mp_obj_new_bytes_from_vstr(vstr_t *vstr) {
 }
 
 mp_obj_t mp_obj_new_str(const char *data, size_t len) {
-    #if MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
-    if (!utf8_check((byte *)data, len)) {
-        mp_raise_msg(&mp_type_UnicodeError, NULL);
-    }
-    #endif
+    mp_utf8_require((byte *)data, len);
     qstr q = qstr_find_strn(data, len);
     if (q != MP_QSTRnull) {
         // qstr with this data already exists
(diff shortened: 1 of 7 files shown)


py/objstr.h
```diff
@@ -119,4 +119,13 @@ extern const mp_obj_dict_t mp_obj_bytearray_locals_dict;
 extern const mp_obj_dict_t mp_obj_array_locals_dict;
 #endif
 
+#if MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
+// Throws an exception if string content is not UTF-8
+void utf8_require(const byte *p, size_t len);
```
dlech: I suggest to give this function the `mp_` namespace prefix. We haven't always been consistent about this, but I think any new functions should have it.

jepler: good suggestion. done.

tests/unicode/unicode_parser.py.exp
```diff
@@ -0,0 +1,5 @@
+UnicodeError
```
dpgeorge: Can we make the output match CPython so this test doesn't need a .exp file?

I see that CPython raises SyntaxError though, which is different to MicroPython here... not sure what the best way forward is.

jepler: Updated.

dpgeorge: Would it be better to make MicroPython raise SyntaxError, to match CPython?

I don't know if that really matters though, this is a pretty rare case, and saving code size (reusing the same UnicodeError) is probably more important.

OTOH, if you had some code like this:
```py
try:
    exec(some_code)
except SyntaxError:
    handle_syntax_error()
```
you might be surprised that the UnicodeError is raised and escapes the error handling.

py/parse.c
```diff
@@ -598,6 +598,9 @@ static mp_parse_node_t make_node_const_object_optimised(parser_t *parser, size_t
 static void push_result_token(parser_t *parser, uint8_t rule_id) {
     mp_parse_node_t pn;
     mp_lexer_t *lex = parser->lexer;
+    if (lex->tok_kind == MP_TOKEN_NAME || lex->tok_kind == MP_TOKEN_STRING) {
+        mp_utf8_require((byte *)lex->vstr.buf, lex->vstr.len);
```
dpgeorge: This is going to impact performance for compilation of all code.  I guess there's not really any way around that if we want to validate the utf8 properly.

jepler: Agreed on both counts. If the overhead is intolerable for some use case, it can be compile-time disabled just like the existing uses of mp_utf8_require (gated by MICROPY_PY_BUILTINS_STR_UNICODE_CHECK, on by default when unicode is on).

(It's been rattling around in my head to have a build flag for "remove all checks a good static checker would diagnose", e.g., code that passes `mypy --strict`, is checked for valid utf-8, etc. but I think there would still end up being a lot of judgement calls and also you can't quite type check micropython/circuitpython code because of lack of exactly matching pyi stubs…)

jepler:
It's not necessarily a great comparison but I made a file called "huge.py" from 16 copies of tests/perf_bench/bm_hexiom.py (total size 268928 bytes) then benchmarked compiling it with the old and new mpy-cross on my x86_64 linux desktop system. Statistically, it was a wash.
```
Benchmark 1: ./mpy-cross-pr17862 huge.py
  Time (mean ± σ):      66.1 ms ±   0.3 ms    [User: 58.6 ms, System: 7.4 ms]
  Range (min … max):    65.3 ms …  67.0 ms    100 runs
 
Benchmark 2: ./mpy-cross-v1.27.0-preview-42-gb7cfafc1ee huge.py
  Time (mean ± σ):      66.3 ms ±   0.3 ms    [User: 59.6 ms, System: 6.6 ms]
  Range (min … max):    65.8 ms …  67.7 ms    100 runs
 
Summary
  ./mpy-cross-pr17862 huge.py ran
    1.00 ± 0.01 times faster than ./mpy-cross-v1.27.0-preview-42-gb7cfafc1ee huge.py
```

Linux `perf stat` counted 4 million more instructions (+0.59%) but another way to look at is 15 added instructions per byte of source code. (baseline is around 2554 instructions per byte of source code)

```
 Performance counter stats for './mpy-cross-pr17862 huge.py' (30 runs):

             66.82 msec task-clock                       #    0.990 CPUs utilized               ( +-  0.09% )
                 0      context-switches                 #    0.000 /sec                      
                 0      cpu-migrations                   #    0.000 /sec                      
               329      page-faults                      #    4.924 K/sec                       ( +-  0.09% )
       282,271,960      cycles                           #    4.224 GHz                         ( +-  0.05% )  (57.31%)
        41,553,651      stalled-cycles-frontend          #   14.72% frontend cycles idle        ( +-  0.33% )  (63.35%)
       690,916,317      instructions                     #    2.45  insn per cycle            
                                                  #    0.06  stalled cycles per insn     ( +-  0.09% )  (63.77%)
       193,477,661      branches                         #    2.896 G/sec                       ( +-  0.08% )  (60.80%)
         1,477,191      branch-misses                    #    0.76% of all branches             ( +-  1.10% )  (54.76%)

         0.0674732 +- 0.0000710 seconds time elapsed  ( +-  0.11% )


 Performance counter stats for './mpy-cross-v1.27.0-preview-42-gb7cfafc1ee huge.py' (30 runs):

             65.99 msec task-clock                       #    0.991 CPUs utilized               ( +-  0.05% )
                 0      context-switches                 #    0.000 /sec                      
                 0      cpu-migrations                   #    0.000 /sec                      
               328      page-faults                      #    4.971 K/sec                       ( +-  0.07% )
       283,490,040      cycles                           #    4.296 GHz                         ( +-  0.05% )  (55.98%)
        42,627,980      stalled-cycles-frontend          #   15.04% frontend cycles idle        ( +-  0.37% )  (62.06%)
       686,832,657      instructions                     #    2.42  insn per cycle            
                                                  #    0.06  stalled cycles per insn     ( +-  0.13% )  (63.52%)
       194,634,958      branches                         #    2.950 G/sec                       ( +-  0.17% )  (62.25%)
         1,374,874      branch-misses                    #    0.71% of all branches             ( +-  1.75% )  (56.18%)

         0.0665979 +- 0.0000371 seconds time elapsed  ( +-  0.06% )
```

jepler:
I changed things a bit so SyntaxError can be raised in this case. I agree it seems preferable if the cost isn't high.

jepler:
That takes e.g., RP2040 from -16 bytes to +16 bytes (net 32 bytes difference to throw the correct exception). Other ports are in the same range of increase.

jepler:
Because it naturally built on the code re-org done here, I added a UTF8 validity check in a path that catches the problem cases of #13084 as well as a cpydiff example for the docs. It doesn't implement CPython compatibility for formatting unicode code points via %c or{:c} but it does prevent invalid strings to be formed in that way and escape.

py/objstr.c
```diff
@@ -2255,6 +2251,7 @@ mp_obj_t mp_obj_new_str_via_qstr(const char *data, size_t len) {
 static mp_obj_t mp_obj_new_str_type_from_vstr(const mp_obj_type_t *type, vstr_t *vstr) {
     // if not a bytes object, look if a qstr with this data already exists
     if (type == &mp_type_str) {
+        mp_utf8_require((byte *)vstr->buf, vstr->len);
```
dlech, line 2254: Does it make a measurable difference in performance if we move this after the if statement?

py/objstr.c
```diff
@@ -2285,17 +2282,13 @@ static mp_obj_t mp_obj_new_str_type_from_vstr(const mp_obj_type_t *type, vstr_t
 }
 
 mp_obj_t mp_obj_new_str_from_vstr(vstr_t *vstr) {
-    #if MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
-    if (!utf8_check((byte *)vstr->buf, vstr->len)) {
-        mp_raise_msg(&mp_type_UnicodeError, NULL);
-    }
-    #endif // MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
+    mp_utf8_require((byte *)vstr->buf, vstr->len);
```
dlech, line 2285: This is now a duplicate check.

py/objstr.c
```diff
@@ -2285,17 +2282,13 @@ static mp_obj_t mp_obj_new_str_type_from_vstr(const mp_obj_type_t *type, vstr_t
...
     return mp_obj_new_str_type_from_vstr(&mp_type_str, vstr);
 }
 
 #if MICROPY_PY_BUILTINS_STR_UNICODE && MICROPY_PY_BUILTINS_STR_UNICODE_CHECK
 mp_obj_t mp_obj_new_str_from_utf8_vstr(vstr_t *vstr) {
-    // bypasses utf8_check.
+    // bypasses utf8_require.
```
dlech, line 2291: This comment is no longer true.

py/objstr.c
jepler: I'm not familiar with how to measure performance but it does seem like it could have a beneficial effect.

py/objstr.c
jepler: hm in fact with this change it's no longer possible to bypass the check, so the entire existence of this function becomes moot.

jepler:
maybe I should back the string formatting stuff out again, it's not as clean a change as I thought. What do you think @dlech ?

dlech:
We could add a `bool` parameter to `mp_obj_new_str_type_from_vstr()` to optionally enable the check, but of course that comes as some small cost in code size and a few extra instructions executed.

jepler:
Closing this up, as it stands this is not a cleanup, it's just making things uglier. #17855 and (new) #18609 track the two issues of creating non UTF-8 string literals & identifiers.

PR #17866, opened 2025-08-08
zephyr: Make time.sleep() and friends thread safe.

danicampora:
Currently the delay function is using a single static k_poll_event object which means that multiple threads sleeping at similar times will try to use the same event variable leading to a eventual crash.

I created a new `mp_hal_delay_ms()` function that is heavily inspired in the one from the ESP32 port.

### DIFF
--- a/ports/zephyr/mphalport.c
+++ b/ports/zephyr/mphalport.c
@@ -29,14 +29,6 @@
 #include "extmod/modmachine.h"
 
 static struct k_poll_signal wait_signal;
-static struct k_poll_event wait_events[2] = {
-    K_POLL_EVENT_INITIALIZER(K_POLL_TYPE_SIGNAL,
-        K_POLL_MODE_NOTIFY_ONLY,
-        &wait_signal),
-    K_POLL_EVENT_STATIC_INITIALIZER(K_POLL_TYPE_SEM_AVAILABLE,
-        K_POLL_MODE_NOTIFY_ONLY,
-        NULL, 0),
-};
 
 void mp_hal_init(void) {
     k_poll_signal_init(&wait_signal);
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the PR.  I can see why this is needed for the case of more than one thread.

But, I did spend quite a bit of effort implementing the existing sleep code, see 916c3fd23fc4b636a8007525c3bd6308d01509d0 and d120859857e9e34bded146977f18a2be8c53ce7b

The idea is that it should be possible to tell the zephyr OS to sleep for the maximum possible time, so it could maybe use a low-power sleep.  And then the semaphore can wake up the thread early if needed, eg when the user sends ctrl-C, or a I2CTarget IRQ fires and needs to run a Python callback.  That's handled by `mp_hal_signal_event()`.

IMO it would be good to keep all those capabilities (sleep for a long time, wake early via a semaphore).  The question is how to do that when there are multiple threads using the sleep mechanism.  My simple proposal for that would be to have 2 sleep functions: the existing sleep implementation would be used for the main thread, and the implementation in this PR would be used for all other threads.  You can select which one via something like this:
```c
void mp_hal_delay_ms(mp_uint_t delay_ms) {
    if (mp_thread_get_state() == &mp_state_ctx.thread) {
        // main thread
        mp_hal_wait_sem(NULL, delay_ms);
    } else {
        // other threads
        mp_hal_delay_ms_polling(delay_ms);
    }
}
```
(Or maybe you can think of a better way to detect the main thread?)

dpgeorge:
Actually, maybe the sleep on non-main threads could be a simple `k_msleep(delay_ms)`?  They don't need to wake to process events, that can be handled by the main thread (that would also mean ctrl-C only goes to the main thread, but that actually matches CPython semantics).

danicampora:
@dpgeorge thanks for explaining the reasons of having the original implementation! I will play a bit with your suggestions and update this PR asap.

danicampora:
@dpgeorge I tried both of the suggestions you made, but they don't produce the desired behaviour.

First problem is that `k_msleep(delay_ms)` cannot be used directly as otherwise the GIL is never released between sleeps. So this approach:

```
void mp_hal_delay_ms(mp_uint_t delay_ms) {
    if (mp_thread_get_state() == &mp_state_ctx.thread) {
        // main thread
        mp_hal_wait_sem(NULL, delay_ms);
    } else {
        // other threads
        mp_hal_delay_ms_polling(delay_ms);
    }
}
```

Seems to work better except that the sleep for the main function does not work at all. Need do continue debugging...

This is my simple test:

```
import time
import _thread

def thread_entry(num):
    for i in range(4):
        print('Running thread num', num + 1)
        time.sleep_ms(500 * (num + 1))

for i in range(4):
    _thread.start_new_thread(thread_entry, (i,))

time.sleep(8)
print("done")
```

Issue #17852, opened 2025-08-06
pin vs stream confusion leads to undefined behavior and Asan diagnostic


jepler:
### Port, board and/or hardware

unix port, coverage variant w/address sanitizer, x86_64 linux

### MicroPython version

MicroPython v1.26.0-preview.524.g255d74b5a8 on 2025-08-06; linux [GCC 12.2.0] version

### Reproduction

Build with `. tools/ci.sh; ci_unix_sanitize_address_build`. Then, run the following code (I did it at the repl):
```py
>>> import machine
>>> import ssl
>>> ctx = ssl.SSLContext(1)
>>> p = machine.PinBase()
>>> ctx.wrap_socket(p)
```

### Expected behaviour

`ctx.wrap_socket` should raise an exception, because a PinBase does not implement the stream protocol. For instance, `ctx.wrap_socket(1)` throws `OSError: stream operation not supported`.

### Observed behaviour

```
>>> ctx.wrap_socket(p)
=================================================================
==2165677==ERROR: AddressSanitizer: global-buffer-overflow on address 0x5585c4169a28 at pc 0x5585c3f6a4fb bp 0x7ffe114fe930 sp 0x7ffe114fe928
READ of size 8 at 0x5585c4169a28 thread T0
    #0 0x5585c3f6a4fa in mp_get_stream_raise ../../py/stream.c:102
    #1 0x5585c3fb01cc in ssl_socket_make_new ../../extmod/modtls_mbedtls.c:623
    #2 0x5585c3fb0a30 in ssl_context_wrap_socket ../../extmod/modtls_mbedtls.c:519
    #3 0x5585c3f3c338 in fun_builtin_var_call ../../py/objfun.c:118
    #4 0x5585c3f1c0d8 in mp_call_function_n_kw ../../py/runtime.c:727
    #5 0x5585c3f1ca56 in mp_call_method_n_kw ../../py/runtime.c:743
    #6 0x5585c3f7e435 in mp_execute_bytecode ../../py/vm.c:1069
    #7 0x5585c3f3c700 in fun_bc_call ../../py/objfun.c:295
    #8 0x5585c3f1c0d8 in mp_call_function_n_kw ../../py/runtime.c:727
    #9 0x5585c3f1ca56 in mp_call_method_n_kw ../../py/runtime.c:743
    #10 0x5585c3f7e435 in mp_execute_bytecode ../../py/vm.c:1069
[...]

PR #17868, opened 2025-08-08
vfs_reader: Check that open() resulted in a file-like object.

### LINKED ISSUE #17841: execfile: segmentation fault with incorrect Python vfs implementation

unix port, coverage build, x86_64

### MicroPython version

MicroPython v1.26.0-preview.524.g255d74b5a8.dirty on 2025-08-05; linux [GCC 12.2.0] version

### Reproduction

Use the following script.
```py
import os
import vfs

class Filesystem:
    def mount(self, readonly, mkfs):
        pass

    def open(self, file, mode):
        return None # violates vfs contract

fs = Filesystem()
vfs.mount(fs, "/test_mnt")
execfile("/test_mnt/test.py")
```

### Expected behaviour

Prints a Python exception because the return type from open() is incorrect (not a VFS file)

### Observed behaviour

Crashes with a segmentation fault

### Additional Information

gdb says:
```
Program received signal SIGSEGV, Segmentation fault.
mp_reader_new_file (reader=reader@entry=0x7fffffffd820, 
    filename=filename@entry=876) at ../../extmod/vfs_reader.c:86
86	    const mp_stream_p_t *stream_p = mp_get_stream(file);
(gdb) where
#0  mp_reader_new_file (reader=reader@entry=0x7fffffffd820, 
    filename=filename@entry=876) at ../../extmod/vfs_reader.c:86
#1  0x00005555555c4ca3 in mp_lexer_new_from_file (filename=876)
    at ../../py/lexer.c:897
#2  0x000055555561b5bf in eval_exec_helper (n_args=1, args=0x7fffffffda50, 
    parse_input_kind=parse_input_kind@entry=MP_PARSE_SINGLE_INPUT)
    at ../../py/builtinevex.c:167
#3  0x000055555561b782 in mp_builtin_execfile (n_args=<optimized out>, 
    args=<optimized out>) at ../../py/builtinevex.c:195
```
possibly `mp_get_stream_raise` should be used instead?

### Code of Conduct

Yes, I agree

jepler:
### Summary

Check that open() results in a file-like object, preventing crashes in callers to `mp_reader_new_file` when an object that is "not actually a file" is returned. This includes, for example, `execfile`.

Closes: #17841

### Testing

I added a test derived from the reproducer in #17841.

The code can only ensure that the object's type defines a protocol slot. Due to protocol confusion, a variant of the original crasher that returned e.g., a machine.Pin instance could still lead to a crash. (#17852)

### DIFF
--- a/extmod/modopenamp_remoteproc_store.c
+++ b/extmod/modopenamp_remoteproc_store.c
@@ -76,6 +76,7 @@ static int openamp_remoteproc_store_open(void *store, const char *path, const vo
 
     openamp_remoteproc_filestore_t *fstore = store;
     fstore->file = mp_vfs_open(MP_ARRAY_SIZE(args), args, (mp_map_t *)&mp_const_empty_map);
+    (void)mp_get_stream_raise(fstore->file, MP_STREAM_OP_READ);
 
     int error = 0;
     mp_uint_t bytes = mp_stream_read_exactly(fstore->file, fstore->buf, RPROC_FILE_STORE_BUF_SIZE, &error);
--- a/extmod/vfs_reader.c
+++ b/extmod/vfs_reader.c
@@ -83,7 +83,7 @@ void mp_reader_new_file(mp_reader_t *reader, qstr filename) {
     };
     mp_obj_t file = mp_vfs_open(MP_ARRAY_SIZE(args), &args[0], (mp_map_t *)&mp_const_empty_map);
 
-    const mp_stream_p_t *stream_p = mp_get_stream(file);
+    const mp_stream_p_t *stream_p = mp_get_stream_raise(file, MP_STREAM_OP_READ);
     int errcode = 0;
 
     #if MICROPY_VFS_ROM
--- a/tests/micropython/builtin_execfile.py
+++ b/tests/micropython/builtin_execfile.py
@@ -75,3 +75,24 @@ except TypeError:
 
 # Unmount the VFS object.
 vfs.umount(fs)
+
+
+class EvilFilesystem:
+    def mount(self, readonly, mkfs):
+        print("mount", readonly, mkfs)
+
+    def umount(self):
+        print("umount")
+
+    def open(self, file, mode):
+        return None
+
+
+fs = EvilFilesystem()
+vfs.mount(fs, "/test_mnt")
+try:
+    execfile("/test_mnt/test.py")
+    print("ExecFile succeeded")
+except OSError:
+    print("OSError")
+vfs.umount(fs)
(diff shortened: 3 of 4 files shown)


jepler:
Multiple tests fail because their vfs open method returns None (or other unsuitable objects, I didn't look any further than the first one).

Should the tests be corrected so they return a minimal file-like object, or should I take a different approach to resolving the crash in `execfile()`?

dpgeorge:
> Multiple tests fail because their vfs open method returns None (or other unsuitable objects, I didn't look any further than the first one).

Looking at the unix coverage job failures, only `extmod/vfs_basic.py` returns `None` for its custom `open()` method.  The other 5 tests return valid pure-Python-implemented files.

Is it only `execfile` that can trigger this bug?  If so, it should probably be fixed there.

jepler:
Anytime you can flow from `mp_vfs_open` to `mp_get_stream` (non-raising variant) I think you can get a similar crash.

Besides mp_reader_new_file this looks like it could crash in principle (extmod/modopenamp_remoteproc_store.c):
```
    fstore->file = mp_vfs_open(MP_ARRAY_SIZE(args), args, (mp_map_t *)&mp_const_empty_map);                         
    
    int error = 0;                                                                                                  
    mp_uint_t bytes = mp_stream_read_exactly(fstore->file, fstore->buf, RPROC_FILE_STORE_BUF_SIZE, &error);
```
since the `mp_stream_read_exactly` macro ends up calling `mp_get_stream`. (openamp is not tested in CI)

Looking through the source those are the only two _obvious_ ones but I'm not sure there aren't some subtle cases.

I went ahead and changed this to put the check down in the two sites I identified rather than directly in vfs_open for the reasons discussed.

dpgeorge:
> I went ahead and changed this to put the check down in the two sites I identified rather than directly in vfs_open for the reasons discussed

Excellent, this feels like the correct fix now.  There are quite a few places already using `mp_get_stream_raise()` just to validate the stream, eg `modbtree.c`, `moddeflate.c`, `modjson.c`.

tests/extmod/vfs_open_error.py.exp
```diff
@@ -0,0 +1 @@
+OSError
```
dpgeorge: This exp file is either missing a corresponding test, or should just be removed.

jepler: Thanks, I'll remove it, I merged the new test into (I hope) an appropriate existing test.

dpgeorge: Yes, the new execfile test is good.

dpgeorge (approved):

dpgeorge:
Rebased and merged in 4b013ec3ff688b8dc3af0cc9159a5fc4c27d8d20 and 0a41838becec8af2d80b90388072090ef2d0a4ad

Thanks @jepler, this turned out to be a very neat fix.

PR #17870, opened 2025-08-08
general: Remove Python 2.7 support.

jepler:
### Summary

Python 2.7 has been EOL since January 2020.

 * Ubuntu oldoldlts (Focal Fossa, 20.04) has Python 3.8. 
 * Debian oldoldstable (Buster, from 2019) has Python 3.7. 
 * RHEL 8 (from 2019) has Python 3.6.

It's easier than ever to install a modern Python using uv.

Given this, it seems like a fine idea to drop Python 2.7 support.

I don't really expect this to be merged right away; perhaps when development officially shifts to Micropython2.

### Testing

Relying on the CI to test this, but there's really not much code change here.

Even though the build is not tested on Python as old as 3.3, I left comments stating that "3.3+" is the baseline Python version. However, it might make sense to bump this to e.g., 3.10, the oldest Python 3 version used during CI. Or, using uv or another method actually test on the oldest Python interpreter that is desirable to support (uv goes back to Python 3.7 easily; in October 2025, the oldest supported Python interpreter version will be 3.10)

### DIFF
--- a/py/makeqstrdata.py
+++ b/py/makeqstrdata.py
@@ -4,8 +4,6 @@ Process raw qstr file and output qstr data with length, hash and data bytes.
 This script works with Python 3.3+.
 """
 
-from __future__ import print_function
-
 import re
 import sys
 
--- a/py/makeqstrdata.py
+++ b/py/makeqstrdata.py
@@ -1,7 +1,7 @@
 """
 Process raw qstr file and output qstr data with length, hash and data bytes.
 
-This script works with Python 2.6, 2.7, 3.3 and 3.4.
+This script works with Python 3.3+.
 """
 
 from __future__ import print_function
@@ -9,13 +9,7 @@ from __future__ import print_function
 import re
 import sys
 
-# Python 2/3/MicroPython compatibility:
-#   - iterating through bytes is different
-#   - codepoint2name from html.entities is hard-coded
-if sys.version_info[0] == 2:
-    bytes_cons = lambda val, enc=None: bytearray(val)
-elif sys.version_info[0] == 3:  # Also handles MicroPython
-    bytes_cons = bytes
+bytes_cons = bytes
 
 # fmt: off
 codepoint2name = {
@@ -57,7 +51,6 @@ codepoint2name = {
     253: "yacute", 165: "yen", 255: "yuml", 950: "zeta", 8205: "zwj", 8204: "zwnj"
 }
 # fmt: on
-# end compatibility code
 
 codepoint2name[ord("-")] = "hyphen"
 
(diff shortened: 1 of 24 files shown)


py/makeqstrdata.py
```diff
@@ -1,21 +1,15 @@
 """
 Process raw qstr file and output qstr data with length, hash and data bytes.
 
-This script works with Python 2.6, 2.7, 3.3 and 3.4.
+This script works with Python 3.3+.
 """
 
 from __future__ import print_function
```
dpgeorge: I think this can be removed.

py/makeqstrdefs.py
```diff
@@ -2,7 +2,7 @@
 This script processes the output from the C preprocessor and extracts all
 qstr. Each qstr is transformed into a qstr definition of the form 'Q(...)'.
 
-This script works with Python 2.6, 2.7, 3.3 and 3.4.
+This script works with Python 3.3+.
 """
 
 from __future__ import print_function
```
dpgeorge: Can remove.

py/makeversionhdr.py
```diff
@@ -1,7 +1,7 @@
 """
 Generate header file with macros defining MicroPython version info.
 
-This script works with Python 2.6, 2.7, 3.3 and 3.4.
+This script works with Python 3.3+.
 """
 
 from __future__ import print_function
```
dpgeorge: Can remove.

dpgeorge (commented):
I guess it's time to make this change!

And we can do it now.  If anyone really needs to use Python 2.7 then they have options:
1. Stick with MicroPython v1.26.x.
2. Copy the old generation scripts (probably only `py/makeqstrdata.py` and `py/makeqstrdefs.py`) from MicroPython v1.26 to the newer version of MicroPython that they want to use.

jepler:
I've updated the branch & removed all the `from __future__ import`s.

tools/pydfu.py
```diff
@@ -77,7 +75,7 @@
 
 # Python 3 deprecated getargspec in favour of getfullargspec, but
 # Python 2 doesn't have the latter, so detect which one to use
-getargspec = getattr(inspect, "getfullargspec", getattr(inspect, "getargspec", None))
+getargspec = inspect.getfullargspec
```
jepler: outdated comment to zap.

dpgeorge (approved):
Looks good to me!

dpgeorge:
Rebased and merged in f0c6f16b9eeedc11f3b042704109fddb9aa3532f through 36ab1c26400a68c9f8c8c24cce03f487e403c312

PR #17879, opened 2025-08-10
mpy-cross: Fix output to stdout on Windows.

dlech:
### Summary

Set stdout to binary mode on Windows to avoid corruption of MPY output.

By default, Windows uses text mode for stdout which will convert any newline byte (\n) to a carriage return + newline (\r\n) pair. This can corrupt the MPY output if the MPY contains any 0x0A byte.

### Testing

I didn't see a way to test this using our existing test framework. But I have made a test in the python-mpy-cross project I maintain.

Failing test before the fix: https://github.com/pybricks/python-mpy-cross/actions/runs/16865780805/job/47772458449#step:4:538

Passing test after the fix: https://github.com/pybricks/python-mpy-cross/actions/runs/16866365087/job/47773824320#step:4:863

Test source: https://github.com/pybricks/python-mpy-cross/blob/f376b412573a0c8c825abc300c5e36a716a85f98/tests/test_mpy_cross.py#L40-L50

### DIFF
--- a/mpy-cross/main.c
+++ b/mpy-cross/main.c
@@ -24,6 +24,11 @@
  * THE SOFTWARE.
  */
 
+#ifdef _WIN32
+#include <fcntl.h>
+#include <io.h>
+#endif
+
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


mpy-cross/main.c
```diff
@@ -92,7 +97,28 @@ static int compile_and_save(const char *file, const char *output_file, const cha
...
+            fflush(stdout);
+            _setmode(fileno(stdout), _O_BINARY);
+
+            nlr_buf_t nlr2;
+            if (nlr_push(&nlr2) == 0) {
+                mp_raw_code_save(&cm, (mp_print_t *)&mp_stdout_print);
+                nlr_pop();
+
+                // Restore the original mode in case we have to print unhandled exceptions
+                fflush(stdout);
```
dpgeorge, line 112: Would the code be a little cleaner if, instead of having a second NLR handler here, just set a flag `restore_stdout = true` and then do the restore down below just before returning from this function / printing the error?

That would mean the call to `mp_raw_code_save()` is not duplicated here either.

dpgeorge: @dlech any thoughts on the above?

dlech: Having the cleanup so far away is not ideal from a readability point of view, but technically should be OK, so no objection from me.

We also have `nlr_push_jump_callback()` now that would clean this up a bit, but still wouldn't avoid duplicate calls to `mp_raw_code_save()`.

dpgeorge: Removing the second NLR would be best IMO.  Then the `flush(stdout); _setmode(...)` can also just have a single instance (at the end of the function, and have a `return ret;` with `ret` set as appropriate).

PR #17880, opened 2025-08-11
tests: improve autodetection of target features and simplify how tests are selected

dpgeorge:
### Summary

There are five changes in this PR, all aimed to automate, simplify and generalise which tests are run on a given target:

1. automatically include `float/` tests if the target supports floating point
- this removes the code which explicitly adds this test directory for certain targets
- it also means the tests won't be added incorrectly, eg on a unix build without float

2. always include `stress/` tests
- this simplifies the customisation of which tests to run
- stress tests are now run on all targets, improving test coverage

3. generalise addition of port specific test directory
- no more explicit code to add, eg, `ports/stm32` to the list of test dirs; it's now done in a general way based on `sys.platform`

4. autodetect if the target has unicode support
- a simple piece of code is added to `target_info.py` to detect unicode support
- unicode tests are now run on all targets (previously only run on unix) which improves test coverage
- other unicode tests (such as `extmod/json_loads.py`) are now properly skipped if the target doesn't have unicode support

5. skip `basics/io_*.py` tests using the SKIP mechanism, rather than having a feature test for the io module
- this is consistent with how all other tests for optional modules are skipped

EDIT: I added a 6th commit, to enable C-stack checking on the webassembly port.  That's needed for it to pass the now-enabled stress tests on this port.  This is definitely an improvement to this port, otherwise users get strange Wasm errors if they overflow the C stack. 

### Testing

Tested locally on unix, webassembly, qemu, PYBD_SF6 and ADAFRUIT_ITSYBITSY_M0_EXPRESS.

Will also be tested by CI.

This should be a win all round, to simplify `run-tests.py` and improve test coverage.

It will also help with Octoprobe, to get more tests passing there (it tests a lot of different configurations, especially smaller targets without unicode).


projectgus (approved):
These look like good simplifications to me!

PR #17882, opened 2025-08-11
tests: add support for native-specific .exp test output files

dpgeorge:
### Summary

There are currently a few tests that are excluded when using the native emitter because they test printing of exception tracebacks, which includes line numbers.  And the native emitter doesn't store line numbers, so gets these tests wrong.

But we'd still like to run these tests using the native emitter, because they test useful things even if the line number info is not in the traceback (eg that threads which crash print out their exception).

This PR adds support for native-specific .exp files, which are of the form `<test>.py.exp.native`.  If such an .exp file exists then it take precedence over any normal `<test>.py.exp` file.  (Actually, the implementation is general enough that it also supports `<test>py.exp.bytecode` as well, if bytecode ever needs a specific exp file.)

In this PR:
- support for `run-tests.py` to look for `<test>.py.exp.native` files (a very simple change)
- adding of a few such .exp.native files to tests that need them
- remove excluded native tests from `run-tests.py`
- tweak a few other tests so they can run with the native emitter

### Testing

Tested locally on PYBD_SF6, RPI_PICO2_W, ESP32 and unix port.  There were no regressions, and the new tests (now no longer excluded) pass with the native emitter.

Will also be tested by CI.

- Could change the test so it doesn't rely on line numbers, but that really defeats the purpose of some of these tests, which aim to test line numbering.
- Could make the .exp have special regex pattern matching, but again that would mean the actual line number for the bytecode emitter is not tested.
- Could make the native emitter support line numbers in tracebacks... but that's very difficult!

### DIFF
--- a/tests/basics/sys_tracebacklimit.py.native.exp
+++ b/tests/basics/sys_tracebacklimit.py.native.exp
@@ -1,0 +1,22 @@
+ValueError: value
+
+limit 4
+ValueError: value
+
+limit 3
+ValueError: value
+
+limit 2
+ValueError: value
+
+limit 1
+ValueError: value
+
+limit 0
+ValueError: value
+
+limit -1
+ValueError: value
+
+True
+False
--- a/tests/micropython/emg_exc.py.native.exp
+++ b/tests/micropython/emg_exc.py.native.exp
@@ -1,0 +1,2 @@
+ValueError: 1
+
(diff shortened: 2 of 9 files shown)


Josverl:
Minor tidbit: To me the naming `<test>.py.native.exp` would make more sense, as that keeps the extension consistent.

projectgus (commented):
This LGTM! Although also have a slight preference for `filename.native.exp` rather than the other way around.

dpgeorge:
Thanks for the reviews.  I've now changed the file pattern to `<test>.py.native.exp` as suggested.

PR #17884, opened 2025-08-11
py/objtype: Make mp_obj_new_type a static function.

dpgeorge:
### Summary

It's only used once, in the same file it's defined, and making it static reduces code size.

Along with this, the associated example code comment in `ports/unix/main.c` has been removed.

### Testing

CI will test this.

This removes the public `mp_obj_new_type()` API function.  But, I doubt that's ever used (@dlech?) and it's more important to optimise code size rather than make this a public facing function.

(If it's really needed, it can be added as a static-inline wrapper.)

### DIFF
--- a/ports/unix/main.c
+++ b/ports/unix/main.c
@@ -616,19 +616,6 @@ MP_NOINLINE int main_(int argc, char **argv) {
     }
     #endif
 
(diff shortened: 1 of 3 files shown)


dlech (approved):
Nope, we're not using it.

projectgus (approved):

PR #17886, opened 2025-08-11
py/mpconfig: Rename MICROPY_PY___FILE__ to MICROPY_MODULE___FILE__.

dpgeorge:
### Summary

There are now quite a few module related configuration options, of the form `MICROPY_MODULE_xxx`.

The old `MICROPY_PY___FILE__` option has been around for a long time and it actually relates to modules, whether `__file__` is set on a module when import it.

So, rename it to `MICROPY_MODULE___FILE__` and organise it alongside the other module options in `py/mpconfig.h`.

Also, move `MICROPY_MODULE___ALL__` so it's with all the other module options as well.

### Testing

Should be a no-op, but will be tested by CI anyway.

Could leave things as they are, but IMO this is a worthy clean up.


ports/esp8266/mpconfigport.h
```diff
@@ -23,6 +23,7 @@
 #define MICROPY_OPT_MATH_FACTORIAL              (0)
 #define MICROPY_REPL_EMACS_KEYS                 (0)
 #define MICROPY_PY_BUILTINS_COMPLEX             (0)
```
dlech, line 25: Looks like this one should be moved with the other `PY_BUILTINS`

dpgeorge: I agree this looks strange, but the order here actually matches the ordering in `py/mpconfig.h`, and that's what I was aiming for.

Probably `MICROPY_PY_BUILTINS_COMPLEX` could be renamed, or moved around in `py/mpconfig.h`.

projectgus (approved):

PR #17887, opened 2025-08-11
stm32: Use IC9 clock source for slow peripherals.

iabdalkader:
### Summary

SPI5 doesn't work with clock source IC14. Instead of using another IC clock, switch to IC9 which is known to work with all peripherals.

### Testing

Tested SPI5 and USART2.

It's possible to configure another IC clock, there are so many of them, but IC9 seems to work.


dpgeorge:
> SPI5 doesn't work with clock source IC14.

Do you know why SPI5 doesn't work with IC14?  The reference manual (rev 2) says it should work.  And IC14 seems like the correct choice for lower speed peripherals, it's routed specifically to those lower speed SPI and UARTs.

iabdalkader:
> Do you know why SPI5 doesn't work with IC14

No idea, I think it's a silicon bug, maybe IC14 connection is disconnected internally. I found another issues with PLLs too (one of them is coupled to the other and won't start without the other). I'm not sure if these issues are specific to this particular package/part or all of them. Note that SPI4 and SPI5 support lower frequencies, that might have something to do with it. Anyway, if you have one of our boards, or a devkit that has SPI5 broken out, it should be very easy to test this.

>  And IC14 seems like the correct choice for lower speed peripherals

They're all interchangeable, IC9 works just the same, but makes SPI5 work.

dpgeorge:
I tested SPI5 on NUCLEO_N657X0 and it works OK for me.

Test procedure:
- deploy latest firmware v1.27.0-preview.58.g989abae12
- connect D11 and D12 (PG1 and PG2) with a wire
- run the following script:
```py
import machine
    
spi = machine.SPI(5, baudrate=5_000_000)
print(spi)   
             
buf = bytearray(8)
print(buf)
spi.write_readinto(b"01234567", buf)
print(buf)   
```
Output:
```
SPI(5, baudrate=3125000, polarity=0, phase=0, bits=8)
bytearray(b'\x00\x00\x00\x00\x00\x00\x00\x00')
bytearray(b'01234567')
```
You can see that the data is sent out MOSI and read back in on MISO, into `buf`.

How did you test it?

iabdalkader:
> I tested SPI5 on NUCLEO_N657X0 and it works OK for me.

Could that board be using a different chip/package? I think we're using the VFBGA223.

> How did you test it?

I test it using our board, SPI5 is connected to the image sensor module (B2B connector). If I switch to IC14, it just stops working. Switch back to IC9 it works (no other changes whatsoever). I realized that even if you have a board, I don't think you have an SPI sensor to test with. Do things work with IC9? If so, then there's no harm in switching to that, regardless of the test.

dpgeorge:
I did a test with the OPENVM_N6 board that I have, and SPI5 seems to work fine.

I enabled it with these pins in `pins.csv`:
```
-SPI5_CS,PA3
-SPI5_SCK,PG12
-SPI5_MISO,PG1
-SPI5_MOSI,PG2
```
and also added the relevant AF entry in `stm32n657_af.csv` for GP12.

I then ran the existing `tests/extmod/machine_spi_rate.py` test (modified to use `SPI(5)`) and it passed, showing that the timing of the `SPI.write` and `SPI.read` is correct for various baudrates.

I then also measured SPI5_SCK on PG12 = P9 on the header, and I could measure a clock rate of 390kHz when configured as `SPI(5, baudrate=390625, polarity=0, phase=0, bits=8)`.

Can you try to reproduce those tests on your board to see if it also works for you?

iabdalkader:
> Can you try to reproduce those tests on your board to see if it also works for you?

Note that we use the following pins, on the B2B header (not the GPIO headers), which makes it hard to measure:

```
E, 15, AF5, SPI5)       // SCLK
D, 4,  AF5, SPI5)       // MISO
A, 4,  AF5, SPI5)       // MOSI
A, 3,  AF5, SPI5)       // SSEL
```

I do have a breakout, and can use the scope to test it, but is it worth the time? IC9 fixes SPI5 problem for us, but doesn't break anything in MicroPython, any reason not to change it? If there's, I will do the test.

EDIT: Alternatively, we could let boards configure which IC to use? Does that work?

dpgeorge:
> but is it worth the time?

IMO, yes.  It would be good to work out the root cause of the issue.

From my testing, it looks like IC14 works just fine with SPI5 (and there's no errata listed for this), so I'm concerned that there's some other subtle bug somewhere else.  Maybe in the mods you have for OpenMV?  Maybe to do with clock sleep?  Maybe it doesn't work with DMA, or large transfers, and it'll start failing with IC9 as well?  Maybe it's related to #17947 somehow?

Also, if we do change it to IC9 then it needs a detailed comment as to why IC14 doesn't work, so someone doesn't revert it in the future (and we don't actually know why IC14 doesn't work for you).

You should be able to run `tests/extmod/machine_spi_rate.py` without any firmware or hardware changes on your board.  Just change the test to use SPI(5).

iabdalkader:
> IMO, yes. It would be good to work out the root cause of the issue.

I tested it again with two different modules, and both IC9 and IC14 seem to work now! Whatever the reason was, it's now gone. You were right that would've been a useless change, sorry for the noise.

dpgeorge:
OK, very good!  Thanks for following up with the test, glad this is resolved.

PR #16112, opened 2024-10-30
tests: add full, automated, top-level test runner for hardware

dpgeorge:
### Summary

With the aim to speed up testing for a release, I put together a script to automatically run tests on devices running MicroPython.  We have a lot of different tests these days, with different ways to run them.  The idea with this top-level test runner is:
- detect all serial devices connected
- detect what platform is running on each device and what its capabilities are (eg WLAN, BLE, ability to import .mpy files)
- run as many tests as possible given the connected devices (eg if there are two WLAN boards connected, then run the network multitests against them, among other things)

Basically all you need to do is connect boards to your PC via USB and run:
```
$ cd tests
$ ./hwtest.py
```
Then it does the rest.  It takes between 10 and 15 minutes per board, so if you have many connected it can take a few hours.  You can also restrict to given devices by specifying them on the command line, eg:
```
$ ./hwtest.py a0 a1 u0
```
(Probably the name of the script `hwtest.py` can change to something better.)

### Testing

This was used to test 16 distinct boards for the recent v1.24.0 release.

### Trade-offs and Alternatives

This is a bit of a hacky script at the moment but can be improved over time.

Currently there are no boards that pass all tests (!!) and so over time we need to fix things and improve the tests so more of them can pass.

PR #17897, opened 2025-08-12
tests: Require SSL certificate file to be available for test to run.

dpgeorge:
### Summary

Previously, any test needing an SSL certificate file would automatically skip if the file could not be found.  But that makes it too easy to accidentally skip tests.

Instead, change it so that the test fails if the certificate file doesn't exist.  That matches, for example, the fact that the test fails if networking (LAN, WiFi) is not active.

### Testing

Tested on the unix port (files always available, passes), and PYBD_SF6 with and without the files on the local filesystem.  It fails without them, passes with them.

IMO this is a good improvement, it simplifies the test and makes it impossible to accidentally skip it.  But it does require users to copy the files across first (as mentioned in the test README).

### DIFF
--- a/tests/multi_net/asyncio_tls_server_client.py
+++ b/tests/multi_net/asyncio_tls_server_client.py
@@ -14,13 +14,6 @@ PORT = 8000
 cert = cafile = "ec_cert.der"
 key = "ec_key.der"
 
-try:
-    os.stat(cafile)
-    os.stat(key)
-except OSError:
-    print("SKIP")
-    raise SystemExit
-
 
 async def handle_connection(reader, writer):
     data = await reader.read(100)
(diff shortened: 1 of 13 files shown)


dpgeorge:
@projectgus I think you had an opinion on this the last time it came up?

projectgus (approved):
I think this a good improvement, as it doesn't mask the failures.

IMO the best thing we could do would be to have the test runner check this (maybe by running on the host some well-formed `multi_test_setup.py` found in the directory before running any tests from that directory, and therefore avoid the whole problem of false-positive failures due to preconditions.

But this is still a good PR, I think. :)

dpgeorge:
> IMO the best thing we could do would be to have the test runner check this (maybe by running on the host some well-formed `multi_test_setup.py` found in the directory before running any tests from that directory, and therefore avoid the whole problem of false-positive failures due to preconditions.

I guess we could add a `multi_net/00_preconditions.py` test which would run first and fail.

To actually copy across certificates, there is #16112 which does that.

> But this is still a good PR, I think. :)

We can still make all those other improvements (eventually) on top of this PR.

PR #17898, opened 2025-08-12
extmod/modtime: Move tuple creation to common localtime implementation.

dpgeorge:
### Summary

This factors code out of the ports and into the common `time.localtime` implementation in `extmod/modtime.c`.  That helps to reduce code duplication, prevent errors in implementation, and hopefully reduce code size.

### Testing

Tested on PYBD_SF6, running the test suite (which includes tests for `localtime()`).  It passes.

CI will test the unix port.

This is a code refactoring, I don't see any downsides to it.

### DIFF
--- a/extmod/modtime.c
+++ b/extmod/modtime.c
@@ -53,26 +53,26 @@
 // - weekday is 0-6 for Mon-Sun
 // - yearday is 1-366
 static mp_obj_t time_localtime(size_t n_args, const mp_obj_t *args) {
+    timeutils_struct_time_t tm;
     if (n_args == 0 || args[0] == mp_const_none) {
         // Get current date and time.
-        return mp_time_localtime_get();
+        mp_time_localtime_get(&tm);
     } else {
         // Convert given seconds to tuple.
         mp_timestamp_t seconds = timeutils_obj_get_timestamp(args[0]);
-        timeutils_struct_time_t tm;
         timeutils_seconds_since_epoch_to_struct_time(seconds, &tm);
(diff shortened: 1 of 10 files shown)


andrewleech:
`unix` isn't covered here, it has a different localtime tuple layout iirc? Should it also be consolidated as a V2 breaking change?
Do you know how much size-add it would be to make this a namedtuple?

dpgeorge:
> `unix` isn't covered here, it has a different localtime tuple layout iirc?

You're right, unix isn't touched by this PR.  But it still returns the same layout (although it has an extra field for "isdst").

> Do you know how much size-add it would be to make this a namedtuple?

Do you mean for all ports, or just unix?

I'm not really sure, but is making it a namedtuple important/useful?

andrewleech:
> extra field for \"isdst\

Ah yes that's it. The inconsistency does add some difficulty when writing modules / tests that are intended to run on both Unix and hardware, though noting DST can be quite important when the platform supports it.

> I'm not really sure, but is making it a namedtuple important/useful?

Yeah sorry it's a bit off topic for this PR, but localtime and os.stat are two things I run into semi-regularly that have named attributes in cpython but not in micropython; it makes them harder to access individual elements in a documented way in code and harder to manually introspect on repl.

Every time I want to I need to research either online or by finding in the mpy source what the fields are, then for code use define field mappings to then reference the right field in a reviewable way.

andrewleech:
Regardless of the above comments, I'm in full support of the changes made already here, they look good and are a really logical cleanup.

PR #17901, opened 2025-08-12
alif/mpconfigport: Enable cryptolib and hashlib.md5/sha1.

dpgeorge:
### Summary

They are enabled when SSL/mbedTLS is included in the firmware (following, eg, the stm32 port).  These new features cost around +1400 bytes of code size.

### Testing

Tested on OPENMV_AE3, running the `tests/extmod/` tests.  They pass.

Could leave these disabled, but this port has plenty of flash and this brings it up to par with other networking boards.

### DIFF
--- a/ports/alif/mpconfigport.h
+++ b/ports/alif/mpconfigport.h
@@ -112,6 +112,9 @@
 
 // Extended modules
 #define MICROPY_EPOCH_IS_1970                   (1)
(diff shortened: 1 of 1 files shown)


iabdalkader:
Thanks! I agree we should enable all features as the flash allows it. I'll be cherry-picking these changes once they're merged.

iabdalkader (approved):

Issue #14068, opened 2024-03-11
Pin.DRIVE_0 mentioned in documentation but not exist in current code


nuschpl:
### Checks

- [X] I agree to follow the MicroPython [Code of Conduct](https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md) to ensure a safe and respectful space for everyone.

- [X] I've searched for [existing issues](https://github.com/micropython/micropython/issues) matching this bug, and didn't find any.

### Port, board and/or hardware

Raspbery Pi Pico ( RP2040)

### MicroPython version

On MicroPython v1.22.2 on 2024-02-22; Raspberry Pi Pico with RP2040

### Reproduction

1. Install forementioned micropython version on the formentioned board.
2. Launch `rshell`
3.  Launch `repl`
4. Reproduce below in interactive prompt:
```python
>>> import machine
>>> machine.Pin.PULL_UP
1
>>> machine.Pin.DRIVE_0
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'Pin' has no attribute 'DRIVE_0'
>>> dir(machine.Pin)
['__class__', '__name__', 'value', 'ALT', 'ALT_GPCK', 'ALT_I2C', 'ALT_PIO0', 'ALT_PIO1', 'ALT_PWM', 'ALT_SIO', 'ALT_SPI', 'ALT_UART', 'ALT_USB', 'IN', 'IRQ_FALLING', 'IRQ_RISING', 'OPEN_DRAIN', 'OUT', 'PULL_DOWN', 'PULL_UP', '__bases__', '__dict__', 'board', 'cpu', 'high', 'init', 'irq', 'low', 'off', 'on', 'toggle']
```

### Expected behaviour

`Pin.DRIVE_0` returns enum value and can be used to configure `Pin` in line with the documentation.
Ideally files:
https://github.com/micropython/micropython/blob/master/ports/rp2/mphalport.h
and
https://github.com/micropython/micropython/blob/master/ports/rp2/machine_pin.h
would contain full implmentation of above bitfield

### Observed behaviour

The ability documented [here](https://docs.micropython.org/en/latest/library/machine.Pin.html) regarding `Pin.DRIVE_0`, `Pin.DRIVE_1` despite supported by hardware is not implemented. It could be achieved with quirk like this:
`machine.mem32[0x4 + 0x04*gpio_pin_number + 0x4001c000] = 0b01110001`
[...]

PR #17906, opened 2025-08-12
rp2/machine_pin: Add support for setting pin drive strength & slew rate

picmaster:
### Summary

Currently when micropython uses GPIO pins as outputs, their actual configuration for pin drive strength and slew rate is the default one, which was set at power-on/reset (4mA, slow slew rate). There's no way to change these pin configurations in the micropython environment (other than low-level hacks with mem32[]).

This patch adds the ability to explicitly configure rp2040/rp2350 output pins' drive strength and slew rate, if/when needed, like this:

    pin1 = Pin(1, Pin.OUT, drive=Pin.DRIVE_12, slew=Pin.SLEW_FAST)

As the new drive/slew parameters are optional kwargs, all existing code and examples will still work as expected, like this:

    pin2 = Pin(2, Pin.OUT)

All relevant documentation files are updated.

### Testing

Tested on rpi-pico and rpi-pico2 boards.

### DIFF
--- a/docs/library/machine.Pin.rst
+++ b/docs/library/machine.Pin.rst
@@ -42,7 +42,7 @@ Usage Model::
 Constructors
 ------------
 
-.. class:: Pin(id, mode=-1, pull=-1, *, value=None, drive=0, alt=-1)
+.. class:: Pin(id, mode=-1, pull=-1, *, value=None, drive=-1, alt=-1, slew=-1)
 
    Access the pin peripheral (GPIO pin) associated with the given ``id``.  If
    additional arguments are given in the constructor then they are used to initialise
@@ -87,15 +87,20 @@ Constructors
        output pin value if given, otherwise the state of the pin peripheral remains
        unchanged.
 
-     - ``drive`` specifies the output power of the pin and can be one of: ``Pin.DRIVE_0``,
-       ``Pin.DRIVE_1``, etc., increasing in drive strength.  The actual current driving
-       capabilities are port dependent.  Not all ports implement this argument.
+     - ``drive`` specifies the drive strength of the pin (applicable only for output pins)
+       and can be one of: ``Pin.DRIVE_0``, ``Pin.DRIVE_1``, etc., increasing in drive strength.
+       The actual current driving capabilities are port dependent.  Not all ports implement this
+       argument.
 
      - ``alt`` specifies an alternate function for the pin and the values it can take are
        port dependent.  This argument is valid only for ``Pin.ALT`` and ``Pin.ALT_OPEN_DRAIN``
        modes.  It may be used when a pin supports more than one alternate function.  If only
        one pin alternate function is supported the this argument is not required.  Not all
        ports implement this argument.
+
+     - ``slew`` specifies the slew rate of the pin (applicable only for output pins) and can
+       be one of: ``Pin.SLEW_SLOW``, ``Pin.SLEW_FAST``, etc.  The actual slew rate capabilities
+       are port dependent.  Not all ports implement this argument.
 
    As specified above, the Pin class allows to set an alternate function for a particular
    pin, but it does not specify any further operations on such a pin.  Pins configured in
@@ -108,7 +113,7 @@ Constructors
 Methods
 -------
 
-.. method:: Pin.init(mode=-1, pull=-1, *, value=None, drive=0, alt=-1)
+.. method:: Pin.init(mode=-1, pull=-1, *, value=None, drive=-1, alt=-1, slew=-1)
 
    Re-initialise the pin using the given parameters.  Only those arguments that
    are specified will be set.  The rest of the pin peripheral state will remain
@@ -270,9 +275,20 @@ not all constants are available on all ports.
           Pin.DRIVE_1
           Pin.DRIVE_2
 
-   Selects the pin drive strength.  A port may define additional drive
-   constants with increasing number corresponding to increasing drive
-   strength.
+   Selects the output pin drive strength. Each port may define it's own drive
+   constants following only one of these approaches to avoid confusion:
+
+      - Sequential drive constants with increasing number corresponding to
+        increasing drive strength, starting at 0 for lowest drive strength
+        (like DRIVE_0, DRIVE_1, etc.);
+      - non-sequential drive constants that identify specific pin currents
+        (like DRIVE_4 for 4mA, DRIVE_12 for 12mA, etc.).
+
+.. data:: Pin.SLEW_SLOW
+          Pin.SLEW_FAST
+
+   Selects the output pin slew rate.  A port may define additional or entirely
+   custom drive constants with constant corresponding to specific slew rate.
 
 .. data:: Pin.IRQ_FALLING
           Pin.IRQ_RISING
--- a/docs/rp2/quickref.rst
+++ b/docs/rp2/quickref.rst
@@ -162,6 +162,9 @@ Use the :ref:`machine.Pin <machine.Pin>` class::
     p4 = Pin(4, Pin.IN, Pin.PULL_UP) # enable internal pull-up resistor
     p5 = Pin(5, Pin.OUT, value=1) # set pin high on creation
 
+    # create output pin on GPIO6 with 8mA drive strength and slow slew rate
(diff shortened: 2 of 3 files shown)


picmaster:
Related to issue [#14068](https://github.com/micropython/micropython/issues/14068).

pyramids:
This might be better(?) documented than my overlapping re-implementation ([PR 18302](https://github.com/micropython/micropython/pull/18302), in the sense that it documents changes to machine.Pin outside the port (which mine does not need because it does not change the API of that class).

It is also 16 bytes smaller (in code size) than my PR, despite defining one more constant and one more argument.

I think this PR does not conform to the naming sequence of drive strength constants (DRIVE_0, DRIVE_1, DRIVE_2, etc., without gaps). Despite the advantage of this PR's approach for someone only targeting the RP2xxx chips, I think it would be better to stick to consecutive, gapless constants.

picmaster:
This patch adds support for **both** drive strength and slew rate, that's why the Pin class interface had to be updated. Care was taken for this change to be non-intrusive and default behavior to be preserved, as explained in the patch comments. Also I tried to extend the interface in a way so other ports can benefit and build on it, not just rp2.

Regarding the specific values for drive strength (aka DRIVE_2, etc), my intention was to define values that look like the actual drive currents, helping the class' user. If other ports decide to use this approach, the user's python code will preserve it's design intent (e.g. gpio pins will have same drive strength across different platforms, which I think is great). The current solution of defining sequential values like DRIVE_0, DRIVE_1, etc has the small disadvantage that these constants don't actually mean anything specific, and also DRIVE_0 will most probably mean different things on different ports.

> I think this PR does not conform to the naming sequence of drive strength constants (DRIVE_0, DRIVE_1, DRIVE_2, etc., without gaps).

There's no such naming sequence for these constants. The file machine.Pin.rst actually says the following, quoting:

> data:: Pin.DRIVE_0
>          Pin.DRIVE_1
>          Pin.DRIVE_2
>
> Selects the pin drive strength.  A port may define additional drive constants with increasing number corresponding to increasing drive strength.

Let's check the text carefully. There's no mentioning of **sequential values**, only **increasing number corresponding to the increasing drive strength**. Exactly what I've done here - my proposed values make it obvious for the class\'s user about the actual drive strengths.

> Despite the advantage of this PR's approach for someone only targeting the RP2xxx chips, I think it would be better to stick to consecutive, gapless constants.

I intentionally tried to update the interface and docs in a way to not imply that it's RP2-only. Other MCUs are also capable of configuring their pin drive strengths and some of their current values can happily be exactly the same (like 2mA, 4mA, etc).

And regarding the patch submission - before writing it I carefully checked the PR queue for other patches adressing the same issue. If there are preceding patches that I missed, I'll happily throw it away.

pyramids:
If that is the correct interpretation, could we (perhaps by adding to your PR?) make that explicit in the documentation?

Say change (what you already quoted)

> data:: Pin.DRIVE_0
> Pin.DRIVE_1
> Pin.DRIVE_2
> 
> Selects the pin drive strength. A port may define additional drive constants with increasing number corresponding to increasing drive strength.

to something like

> data:: Pin.DRIVE_0
> Pin.DRIVE_1
> Pin.DRIVE_2
> 
> Selects the pin drive strength. A port may define additional drive constants with increasing number corresponding to increasing drive strength. It does not need to define Pin.DRIVE_0 or Pin.DRIVE_1 as long as some other, higher indexed constant is defined; for example, the drive strengths for the rp2 port start at Pin.DRIVE_2 and are not consecutively numbered; so if you ever migrate from the rp2 port to one with a more traditional drive strength numbering approach, be aware that what selects the absolutely weakest drive strength on RP2xxx chips is, if used on other ports, likely to select a drive strength actually stronger rather than weaker than that port's default value.

That would fix the issue I tried to point out in the very minimal sense that at least it is documented in a way that I would not be surprised by it (if I carefully read the documentation). IMHO a better approach would be to follow what maybe has only been implied by the existing documentation and, presumably, other ports, namely that consecutive Pin.DRIVE_0, Pin.DRIVE_1, etc., constants exist and that they start at the weakest strength a chip supports (rather than at 0 mA, as your naming would suggest).

picmaster:
I see your point. But see, the current documentation has already addressed this just a little bit earlier, at the same file machine.Pin.rst, quoting:
> **Constants**
> The following constants are used to configure the pin objects. Note that not all constants are available on all ports.

Which I think is quite clear. But still, let me try to address your concern, with something like this:

> data:: Pin.DRIVE_0
          Pin.DRIVE_1
          Pin.DRIVE_2
> 
> Selects the pin drive strength. Each port may define it's own drive constants following only one of these approaches to avoid confusion:
> - sequential drive constants with increasing number corresponding to increasing drive strength, starting at 0 for lowest drive strength (like DRIVE_0, DRIVE_1, etc);
> - non-sequential drive constants that identify specific pin currents (like DRIVE_4 for 4mA, DRIVE_12 for 12mA, etc).

Is this better?

pyramids:
It's very clear and hence, yes, better.

**EDIT: Referring to your suggestion for improvement, I think that is very clear. Regarding the original and existing documentation, I actually think it contradicts your interpretation (rather than clearly supporting it) because it can also be read as declaring that DRIVE_0, DRIVE_1, DRIVE_2, etc., exists, except that the list might go on or stop before reaching DRIVE_2 if the chip doesn't have that many different drive levels. I think this interpretation is more natural than your interpretation.**

**EDIT 2: I read the quoted "Note that not all constants are available on all ports" to mean that constants such as `Pin.OPEN_DRAIN` or `Pin.PULL_DOWN` might be missing from some ports, particularly if they are unavailable due to hardware limitations (but potentially also because nobody has bothered to implement them, such as `Pin.ANALOG` for the rp2 port).**

If we have to keep your numbering approach, it's as good as I think we will get.

But if your numbering approach---naming the DRIVE_x constants to remind the user that it selects a drive strength nominally corresponding to x mA---is better for RP2xxx chips, then this approach is better for every port! AFAIK, all microcontrollers implementing a drive strength selection name the selection option by the nom. current, in mA, that it will be able to drive without leaving (some) TTL voltage level specifications.

So if the consecutive numbering approach already used by some(?) other ports does not work for the rp2 port, we absolutely must inform those other ports that their approach is just as flawed for them and they must change their ways to also reflect the actual, nominal drive strength in mA in their DRIVE_x constants. In fact, IMHO a better approach would be to have the user specify the nom. current (or current range) as an integer in mA that they are looking for---but that is incompatible with the existing `machine.Pin(..)` and `machine.Pin.init(..)` interface, at least using the `drive` keyword argument.

Frankly, I think mimicking the existing, consecutive DRIVE_x constants is best. If I ever program a RP2xxx device to not use the default drive strength of nom. 4 mA  but instead the minimal drive strength of 2 mA, I would like it to also select a minimal drive strength if I ever try running  the same MicroPython code on an ESP32Cx chip, and vice versa. Hence I have a strong preference for copying the consecutive, 0-indexed DRIVE_x constants used on the esp32 port to rp2. Or at least l would want to also change the esp32 port to not interpret a DRIVE_2 [mA] constant to _increase_ the drive strength to way more than 2 mA---even throwing an exception for a missing constant would be better than potentially frying external circuitry in an unexpected way.

picmaster:
IMHO it's wrong to enforce such type of constraints to the ports' developers, especially if not everyone has added drive strength/slew rate support as of today and the existing knowledge doesn't show an easy/obvious solution for everyone.

Also this discussion is heading exactly nowhere. I don't have any interest in continuing.

I'll submit the doc changes and will leave the final decision to the responsible micropython maintainer, whoever that is:
- either accept the patch with the non-contiguous drive ranges as-is and merge it;
- notify me to rework it for contiguous drive values.

picmaster:
Description of force-pushed changes:
- Edit subject line to comply with git commit conventions;
- Rework Pin.DRIVE_xx description in machine.Pin.rst to better clarify the proposed change.

agatti:
Apologies for this to be checked only now :|

I'd personally split this PR in two, one for the pin strength and one for the slew rate.  That would give us the chance to figure out a set of constant names that best suits all ports that do support changing the slew rate, whilst eventually let the drive strength changes go through as the constants set is already defined (for better or for worse).

W.r.t. strength level numbering, all other ports use 0, 1, 2, etc. without gaps if I'm not mistaken, so it's better to stick to that.  Adding aliases to indicate the current may be an option, for which I'm actually neutral about.

Issue #17896, opened 2025-08-11
@micropython.viper doesn't work on Alif.


kwagyeman:
### Port, board and/or hardware

OpenMV AE3, Alif E3

### MicroPython version

V1.26

### Reproduction

Any function with the viper decorator throws and error when being compiled:

```
MPY rpc.py
error compiling rpc.py:
Traceback (most recent call last):
  File "rpc.py", line 22, in rpc
SyntaxError: invalid arch
```

Something to do with this:

<img width="654" height="192" alt="Image" src="https://github.com/user-attachments/assets/8999baee-f63f-4b57-9e13-01da7babcdf2" />

### Expected behaviour

Code compiles.

### Observed behaviour

Code doesn't compile.

### Additional Information

No, I've provided everything above.

### Code of Conduct

Yes, I agree

dpgeorge:
I did run native and viper and mpy tests for alif as part of the release testing, and they all passed.

Running them now again on OPENMV_AE3, and they still pass.  Eg:
```
$ ./run-tests -t a0 --via-mpy micropython/viper*.py
platform=alif arch=armv7emdp inlineasm=thumb
pass  micropython/viper_addr.py 
pass  micropython/viper_args.py 
pass  micropython/viper_binop_arith.py 
pass  micropython/viper_binop_arith_uint.py 
...
pass  micropython/viper_unop.py 
pass  micropython/viper_with.py 
39 tests performed (403 individual testcases)
39 tests passed
```

Can you please describe in more detail how you arrive at the error?  A minimal reproduction would be great.

(Maybe you forgot the `-march=armv7emdp` option?)

kwagyeman:
Mmm, maybe it's an issue with our build system in upstream OpenMV. I'll check and then close if that's the case.

kwagyeman:
@dpgeorge

It's because the MPY_CROSS_MCU_ARCH is not set. E.g. in lib/micropython/ports/stm32/stm32.mk:

```
MPY_CROSS_MCU_ARCH_f0 = armv6m
MPY_CROSS_MCU_ARCH_f4 = armv7m
MPY_CROSS_MCU_ARCH_f7 = armv7m
MPY_CROSS_MCU_ARCH_g0 = armv6m
MPY_CROSS_MCU_ARCH_g4 = armv7m
MPY_CROSS_MCU_ARCH_l0 = armv6m
MPY_CROSS_MCU_ARCH_l1 = armv7m
[...]

PR #17908, opened 2025-08-13
alif/alif.mk: Add MPY_CROSS_FLAGS setting.

dpgeorge:
### Summary

The HP and HE CPUs have double-precision hardware floating point, so can use the armv7emdp architecture.

This allows frozen code to use native/viper/asm_thumb decorators.

Fixes issue #17896.

### Testing

Tested on OPENMV_AE3, putting native/viper/asm_thumb code in a frozen module.  It works.

### DIFF
--- a/ports/alif/alif.mk
+++ b/ports/alif/alif.mk
@@ -22,6 +22,8 @@ include $(TOP)/extmod/extmod.mk
 ################################################################################
 # Project specific settings and compiler/linker flags
 
+MPY_CROSS_FLAGS += -march=armv7emdp
+
 CROSS_COMPILE ?= arm-none-eabi-
 ALIF_DFP_REL_TOP ?= lib/alif_ensemble-cmsis-dfp
 ALIF_DFP_REL_HERE ?= $(TOP)/lib/alif_ensemble-cmsis-dfp


kwagyeman:
@dpgeorge - Awesome!

ports/alif/alif.mk
```diff
@@ -22,6 +22,8 @@ include $(TOP)/extmod/extmod.mk
 ################################################################################
 # Project specific settings and compiler/linker flags
 
+MPY_CROSS_FLAGS += -march=armv7emdp
```
iabdalkader, line 25: Should this be set for float implementation or is it unrelated?
```suggestion
ifeq ($(MICROPY_FLOAT_IMPL),float)
MPY_CROSS_FLAGS += -march=armv7emdp
else
MPY_CROSS_FLAGS += -march=armv7emsp # ?
endif
```
Note that we use we use `MICROPY_FLOAT_IMPL=float` when building. Would that affect the loaded frozen code?

dpgeorge: This setting is unrelated to the MICROPY_FLOAT_IMPL setting.

This setting is about the hardware capabilities, not any API/ABI (at least for frozen code, which is what matters here).  It means you can use floating-point assembly instructions in `@micropython.asm_thumb` functions.

> Note that we use we use `MICROPY_FLOAT_IMPL=float` when building. Would that affect the loaded frozen code?

No, it won't matter.

(I'm curious why you don't use double though?  Float objects still the same amount of heap as double (16 bytes) and you'd get more precision with doubles.)

kwagyeman: We use the object representation that has floats as 4-bytes.

Matters a lot of large arrays. You'd run out of memory quick.

dpgeorge: You can still make 32-bit float arrays (in C and Python) while still using double for MicroPython's float object.

iabdalkader: >  It means you can use floating-point assembly instructions in @micropython.asm_thumb functions.

But wouldn't `armv7emdp` emit double-precision floating point instructions?

> I'm curious why you don't use double though?

Historically we only had single-precision FPUs so we hard-coded `float` in many places (not `mp_float_t`) switching would break a lot of code, but that's easily fixable. However, I still don't think we should use double because it uses more bandwidth and more cycles. I couldn't easily find a reference for this, but I think double-precision has to be slower, at the very least it's wider so more memory bandwidth on loads/stores. The CM55 has a performance monitoring unit (PMU), it would be interesting to use it to benchmark float vs double.

> You can still make 32-bit float arrays (in C and Python) while still using double for MicroPython's float object.

You mean by casting back and forth? We don't have control over all modules, for example ulab arrays would double in size, and its object files would use double-precision instructions. Also, I think the same goes for MicroPython for any float operations performed in Python.


EDIT: Not sure if I'm using it correctly:

```C
#include "pmu_armv8.h"

uint32_t test_float(size_t iterations) {
    ARM_PMU_CYCCNT_Reset();
    volatile float x = 1.234f, y = 5.678f, z = 0.0f;
    for (size_t i = 0; i < iterations; i++) {
        z += x * y;   // FP32 multiply-add
    }
    return ARM_PMU_Get_CCNTR();
}

uint32_t test_double(size_t iterations) {
    ARM_PMU_CYCCNT_Reset();
    volatile double x = 1.234, y = 5.678, z = 0.0;
    for (size_t i = 0; i < iterations; i++) {
        z += x * y;   // FP64 multiply-add
    }
    return ARM_PMU_Get_CCNTR();
}

int benchmark_float(void) {
    const size_t N = 1000000;

    __disable_irq();
    ARM_PMU_Enable();
    ARM_PMU_CNTR_Enable(PMU_CNTENSET_CCNTR_ENABLE_Msk);
   
    uint32_t flt_cycles = test_float(N);
    uint32_t dbl_cycles = test_double(N);

    __enable_irq();
    printf("Float cycles: %lu\n", (unsigned long)flt_cycles);
    printf("Double cycles: %lu\n", (unsigned long)dbl_cycles);

    while (1);
}
```
Output with `-Og`:

```
Float cycles: 1000358
Double cycles: 4300843
```

Output with `-O2`:
```
Float cycles: 800347
Double cycles: 3000302
```

dpgeorge: > But wouldn't `armv7emdp` emit double-precision floating point instructions?

`armv7emdp` describes what the hardware supports, not necessarily what instructions to use.  If you write `@micropython.asm_thumb` functions, you need to explicitly name the asm instructions you use, and you can use single or double (or even half) precision instructions.  All instructions would be compatible with firmware using either single or double floats as MicroPython objects, because assembler functions can't directly take/return floats (only integers).

Note also that you can run native .mpy modules (eg `examples/natmod/random`) built with either single or double float, and they'll be compatible with firmware using either of those configurations.  That's because it doesn't matter what precision the native .mpy uses internally, the numbers are still correctly marshalled to/from the Python side.

> However, I still don't think we should use double because it uses more bandwidth and more cycles

For Python expression using float objects, the overhead of the float vs double would be negligible.  It would only matter for large operations, eg ulab/numpy operations.

Regarding the benchmark above: that's interesting!  But I think it's maybe a little inaccurate because all the variables are volatile so there's a lot of memory operations going on.  It should still work to remove the volatile.

iabdalkader: >  But I think it's maybe a little inaccurate because all the variables are volatile so there's a lot of memory operations going on

Without the volatile the compiler optimizes away the functions or the calls, but I don't think that matters because both tests would have the same overhead, and I just wanted to show the relative difference not absolute cycle count.

iabdalkader: I tried without volatile, some ratio (~4x) but somehow uses more cycles, or maybe I ran the test before with 100_000 not 1_000_000
```
Float cycles: 5000706
Double cycles: 18000238
```

iabdalkader: Yeah I think I ran the test the first time with less iteration. I changed the test a bit so it doesn't get optimized out, and ran it again with 1_000_000 iterations and `-O2`:

With volatile:
```
Float cycles: 9000756
Double cycles: 31001274
```

Without volatile:
```
Float cycles: 6000008
Double cycles: 18000003
```
In either case you see it takes about 3-4x more cycles.

dpgeorge: > In either case you see it takes about 3-4x more cycles.

OK, that's good information to have.

But my point still stands: for Python code this won't make much of a difference.  It will however make a big difference for things like ulab and if you have custom C code that is floating-point heavy and uses `mp_float_t` instead of explicitly using `float`.

iabdalkader (approved):

ports/alif/alif.mk
iabdalkader: >  It will however make a big difference for things like ulab and if you have custom C code that is floating-point heavy and uses mp_float_t instead of explicitly using float.

ulab uses `mp_float_t`, while our code still uses hard-coded `float`. However, it still won't build with `IMPL=double` because double will need to be cast back to float. I'd rather refactor the code to use `mp_float_t` instead of adding the casts, this way if a port/board uses double it builds. For now, all we need is single-precision.

dpgeorge: > our code still uses hard-coded `float`. However, it still won't build with `IMPL=double` because double will need to be cast back to float. I'd rather refactor the code to use `mp_float_t` instead of adding the casts, this way if a port/board uses double it builds

I'd actually suggest to stick with hard-coded `float` everywhere, because that's what you've designed your algorithm about and optimised it with.  Then, use the following provided functions to interoperate with MicroPython objects: `mp_obj_get_float_to_f`, `mp_obj_get_float_to_d`, `mp_obj_new_float_from_f`, `mp_obj_new_float_from_d`.  They adjust themselved based on single/double settings.

(You could also define your own float type, eg `omv_float_t`, and use that in all your code, at least then it's configurable.)

iabdalkader: > interoperate with MicroPython objects: `mp_obj_get_float_to_f`, `mp_obj_get_float_to_d`, `mp_obj_new_float_from_f`

Didn't know about those, thanks! Yes, I'll do that instead.

Issue #17865, opened 2025-08-08
Uploading to esp32-s3 (4MB) corrupts file


gohai:
### Port, board and/or hardware

esp32-s3 (on a ESP32-S3 (QFN56) (revision v0.2), with 4 MB flash)

### MicroPython version

MicroPython v1.26.0-preview.527.g593ae04ee on 2025-08-07; Generic ESP32S3 module with ESP32S3

mpremote 1.25.0, on macOS

### Reproduction

1. Upload a file to the board (e.g. `mpremote cp bar.py :/bar.py`)
2. Download the file again (`mpremote cp :/bar.py bar2.py`)
3. Compare the file content (`diff -u bar.py bar2.py`)

The files reproducibly differ for me. (Seeing syntax errors in repl as well, so the corruption doesn't seem to be caused by the second transfer back to the computer.)

### Expected behaviour

No file corruption. (Content should be identical.)

### Observed behaviour

File corruption, e.g.

```diff
 def create_wav_header(sampleRate, bitsPerSample, num_channels, num_samples):
     datasize = num_samples * num_channels * bitsPerSample // 8
     o = bytes("RIFF", "ascii")  # (4byte) Marks file as RIFF
     o += (datasize + 36).to_bytes(
         4, "little"
+    )  # (4byte) File size in bytes excluding this and Rittle"
     )  # (4byte) File size in bytes excluding this and RIFF marker
     o += bytes("WAVE", "ascii")  # (4byte) File type
     o += bytes("fmt ", "ascii")  # (4byte) Format Chunk Marker
     o += (16).to_bytes(4, "little")  # (4byte) Length of above format data
+    o += (1).to_bytes(2, "little")  # (2byte) Format tyat data
     o += (1).to_bytes(2, "little")  # (2byte) Format type (1 - PCM)
     o += (num_channels).to_bytes(2, "little")  # (2byte)
     o += (sampleRate).to_bytes(4, "little")  # (4byte)
     o += (sampleRate * num_channels * bitsPerSample // 8).to_bytes(4, "little")  # (4byte)
+    o += (num_channels * bitsPerSample /, "little")  # (4byte)
     o += (num_channels * bitsPerSample // 8).to_bytes(2, "little")  # (2byte)
     o += (bitsPerSample).to_bytes(2, "little")  # (2byte)
     o += bytes("data", "ascii")  # (4byte) Data Chunk Marker
     o += (datasize).to_bytes(4, "little")  # (4byte) Data size in bytes
[...]

Issue #17560, opened 2025-06-25
ESP32-S3 `mpremote` sends corrupted fs hook when mounting local directory


adzierzanowski:
### Port, board and/or hardware

ESP32-S3 N8R8

### MicroPython version

```MicroPython 6fee099ca-dirty on 2025-06-25; Generic ESP32S3 module with Octal-SPIRAM with ESP32S3```

Compiled using
* IDF 5.4.1
* python 3.13.1
* gcc-15
* macOS Sequoia 15.5

`mpremote` is used directly from the micropython repository (but the results are same as with latest PyPI version)

### Reproduction

1. Have a local folder named `local`
2. Run `mpremote mount local`

### Expected behaviour

Expected to mount local directory on the device and enter the REPL

### Observed behaviour

```python
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File ".../micropython/tools/mpremote/mpremote/__main__.py", line 6, in <module>
    sys.exit(main.main())
             ~~~~~~~~~^^
  File ".../micropython/tools/mpremote/mpremote/main.py", line 615, in main
    handler_func(state, args)
    ~~~~~~~~~~~~^^^^^^^^^^^^^
  File ".../micropython/tools/mpremote/mpremote/commands.py", line 521, in do_mount
    state.transport.mount_local(path, unsafe_links=args.unsafe_links)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../micropython/tools/mpremote/mpremote/transport_serial.py", line 304, in mount_local
    self.exec(fs_hook_code)
    ~~~~~~~~~^^^^^^^^^^^^^^
  File ".../micropython/tools/mpremote/mpremote/transport_serial.py", line 293, in exec
    raise TransportExecError(ret, ret_err.decode())
mpremote.transport.TransportExecError: Traceback (most recent call last):
  File "<stdin>", line 11
SyntaxError: invalid syntax
```

### Additional Information

After some digging I found a recent thread #17465 which helped to resolve the issue.

After changing the size of a data chunk sent at once over REPL, mounting works as expected:

```python
        for i in range(0, len(command_bytes), 32):
            self.serial.write(
                command_bytes[i : min(i + 32, len(command_bytes))]
            )
            time.sleep(0.01)
```

[...]

PR #17913, opened 2025-08-13
esp32/esp32_common.cmake: Switch back to the vendored TinyUSB copy.

agatti:
### Summary

This PR reverts d7371124d288cc85d2386b63ea39cfa4fa803194, fixing USB issues reported by macOS users on ESP32S2/ESP32S3 boards.

The commit's intention was to rely on the TinyUSB tree provided by Espressif as a dependency of their USB support library (so maintenance and other bug fixes would come from them rather than from MicroPython).

However, this seems to have opened a Pandora's box of incompatibilities triggered by macOS's USB stack behaviour (see discussion #17465 and issues #17560 and #17865).

Whilst this may look like a permanent solution, it is not.  There are some important race condition fixes added in Espressif's patched USB code that are not available in MicroPython's TinyUSB copy (along other bug fixes and feature additions).  Still, cutting macOS users out of their boards probably isn't the best thing to do.

### Testing

Before this change, copying a file to the board from a macOS machine via USB using `mpremote` would fail either on the first or on the second iteration.

With this change applied, repeated copies to and from an ESP32S3-DevKitC-R8N16 succeed without problems.

The only trade-off I can see right now is that by reverting to that particular TinyUSB version we're missing on race condition fixes being added both by the TinyUSB authors and by Espressif.

Alternatives?  Where to start...  There seems to be a bunch of incompatibilities manifesting at the same time only in a very specific situation:

- This is triggered by macOS's USB stack, things seem to be fine on Linux and Windows
- The issue shows up as either losing data and deadlocking TinyUSB, not deadlocking anything but silently duplicating data instead, or just hang on the REPL when printing the help message; the whole lot seems to be board-dependent to a certain extent
- By forcing different versions of the `espressif/esp_tinyusb` and `espressif/tinyusb` components in `ports/esp32/main/idf_component.yml` things seem to be fixed for some users although with the same versions set I could still get my board to lock up during a stress test
- When letting `espressif/tinyusb` match the version of what's present in `lib/tinyusb`, the issue still appears - so this seems related to `espressif/esp_tinyusb` instead, but again, changing version numbers changes its final behaviour
- This was reported on both ESP32S3 and ESP32S2, so race conditions and multi-core issues do not seem to be related

I got a macOS machine on loan from a friend and I can at least replicate the mpremote issues on my board so I can slowly work on this in the meantime (I've been busy with a family situation the past couple of weeks).

I've only got one S3 board and to have JTAG exposed on GPIOs in order to have USB CDC available an eFuse must be burnt (thanks @projectgus for the help on that!).  I'm a bit hesitant to burn the eFuse on my board right now, being my only S3 specimen; I've ordered a new one but the shipping times combined with a lackluster postal service plus holiday season here means somebody else may fix this issue before I am ever able to get a JTAG probe attached to the new board (if that happens I won't mind at all!).

Incidentally, I am able to replicate the same issue on a whole different set of conditions (WCH CH32V208 single-core MCU using STM's DCD while connecting from Linux) using the vendored TinyUSB copy, but that may be caused by a coding error on my end.

### DIFF
--- a/ports/esp32/esp32_common.cmake
+++ b/ports/esp32/esp32_common.cmake
@@ -80,7 +80,9 @@ list(APPEND MICROPY_SOURCE_DRIVERS
     ${MICROPY_DIR}/drivers/dht/dht.c
 )
 
+list(APPEND GIT_SUBMODULES lib/tinyusb)
 if(MICROPY_PY_TINYUSB)
+    set(TINYUSB_SRC "${MICROPY_DIR}/lib/tinyusb/src")
     string(TOUPPER OPT_MCU_${IDF_TARGET} tusb_mcu)
 
     list(APPEND MICROPY_DEF_TINYUSB
@@ -88,6 +90,12 @@ if(MICROPY_PY_TINYUSB)
     )
 
     list(APPEND MICROPY_SOURCE_TINYUSB
+        ${TINYUSB_SRC}/tusb.c
+        ${TINYUSB_SRC}/common/tusb_fifo.c
+        ${TINYUSB_SRC}/device/usbd.c
+        ${TINYUSB_SRC}/device/usbd_control.c
+        ${TINYUSB_SRC}/class/cdc/cdc_device.c
+        ${TINYUSB_SRC}/portable/synopsys/dwc2/dcd_dwc2.c
         ${MICROPY_DIR}/shared/tinyusb/mp_usbd.c
         ${MICROPY_DIR}/shared/tinyusb/mp_usbd_cdc.c
         ${MICROPY_DIR}/shared/tinyusb/mp_usbd_descriptor.c
@@ -95,6 +103,7 @@ if(MICROPY_PY_TINYUSB)
     )
 
     list(APPEND MICROPY_INC_TINYUSB
+        ${TINYUSB_SRC}
         ${MICROPY_DIR}/shared/tinyusb/
     )
 endif()


projectgus:
Thanks @agatti for digging into this.

> By forcing different versions of the espressif/esp_tinyusb and espressif/tinyusb components in ports/esp32/main/idf_component.yml things seem to be fixed for some users although with the same versions set I could still get my board to lock up during a stress test

Just to check all the details, you don't get the stress test lockup with the version in this PR? Has anyone else reported problems when running the newer versions of the Espressif components?

Given that Espressif are likely to continue to debug and patch issues in their forks and then submit them upstream periodically into the future, strategically it might still be better to use their fork rather than upstream TinyUSB (especially with things like ESP32-P4 support looming on the horizon). What do you think?

If we can isolate the stress test lockup to TinyUSB then we can also report this to them (although one thing I can't account for is neither ESP-IDF or esp-tinyusb have similar bug reports regarding macOS+USB that I can see, which makes me think our unusual TinyUSB integration in MicroPython might be a contributing factor...)

Regarding JTAG debugging the stress test, I can borrow a Mac laptop and do some testing here if I don't need to set up a full development environment on the Mac. Is it easy to explain how to run it?

dpgeorge:
I can run stress tests on an old Macbook Air if needed.

agatti:
> Just to check all the details, you don't get the stress test lockup with the version in this PR? Has anyone else reported problems when running the newer versions of the Espressif components?

That's correct, I don't encounter lockups with this commit applied.

Regarding newer component versions: yes, there have been problem reports.  https://github.com/micropython/micropython/issues/17865#issuecomment-3181759500 for example, in which an S3 devkit board that used to work before now hangs at REPL with the updated Espressif components.  However that same issue is also *fixed* by those same components on two different boards that also had the issue in the first place.

> Given that Espressif are likely to continue to debug and patch issues in their forks and then submit them upstream periodically into the future, strategically it might still be better to use their fork rather than upstream TinyUSB (especially with things like ESP32-P4 support looming on the horizon). What do you think?

Personally I'm all for it.  If once this gets sorted out a similar issue arises it should be easy to switch to a local TinyUSB repo copy, that's what we've been doing after all.

> If we can isolate the stress test lockup to TinyUSB then we can also report this to them (although one thing I can't account for is neither ESP-IDF or esp-tinyusb have similar bug reports regarding macOS+USB that I can see, which makes me think our unusual TinyUSB integration in MicroPython might be a contributing factor...)

That's the interesting bit.  The PHY initialisation code in MPy doesn't set external GPIO pin indices in the configuration structure, but since only the internal interface is used it shouldn't really matter.  I'd wager if that was the problem it would have shown up on Linux and WSL too.

Everything else is virtually the same, except for how the TinyUSB task is run.  That could be related to the fact that the issue mainly presents on macOS as TinyUSB's state machine may be a step or two behind what it expects to be?  I don't think modern macOS versions let you dump USB traffic, but I can take a look at that to see if a particular pattern emerges.

> Regarding JTAG debugging the stress test, I can borrow a Mac laptop and do some testing here if I don't need to set up a full development environment on the Mac. Is it easy to explain how to run it?

This should be enough:

```sh
dd if=/dev/urandom of=file bs=250000 count=1
for i in $(seq 0 50); do mpremote cp file :file-$i; mpremote cp :file-$i file-$i; diff file file-$i; done
```

On the laptop/board I run this it hangs on file number 35.  Sometimes on smaller runs it'd complete all transfers but then hang on `mpremote ls`.  I haven't tried to add delays between transfers or things like that - the situation was already puzzling to begin with.

> I can run stress tests on an old Macbook Air if needed.

For the record, I can make things hang on macOS 13.7.6 - maybe older versions (or newer, that is) may not exhibit the issue.

I wish I could provide more help on this but things are "interesting" on my end to say the least.

agatti:
If that helps, here's a wireshark dump of the USB interface for a failed transfer between macOS 13.7.6 and an ESP32S3 using the release v1.26.0 firmware off micropython.org.  There's some fluff at the beginning, but if you need any pointers in the timeline:

* Packet 587 should be the first packet sent by mpremote
* Packet 12205 is the last valid file data chunk sent from macOS
* Packet 12217 is the error packet reported by the board

[failed_mpremote_copy.pcapng.gz](https://github.com/user-attachments/files/21778284/failed_mpremote_copy.pcapng.gz)

agatti:
@projectgus given that you sorted the issue out, this PR can be closed, right?

dpgeorge:
Yes, I think we can close this PR.

I prefer to stick with Espressif's TinyUSB component (as you @agatti originally did in d7371124d288cc85d2386b63ea39cfa4fa803194) and just make sure that works correctly.

agatti:
Sure, no problem - closing the issue.

PR #12347, opened 2023-08-31
mimxrt: Add Quadrature Encoder and Pulse Counter classes.

robert-hh:
This PR adds Quadrature Encoder and Pulse Counter classes based on the Encoder hardware of the mimxrt MCUs. The base methods are as simple as possible, with options to make use of the hardware features supporting fast encoder sensors, like index pulses.

Tested with a slow manual encoder and a simulation of fast encoder/counter signals up to a input frequency of 25 MHz.

The PR is a re-submit of PR #7911, which could not be reopened. The conversation at that PR applies here as well (except for the UART part).

PR #12346, opened 2023-08-31
esp32: Add machine.Counter and machine.Encoder.

jimmo:
This builds on the extensive work done by @IhorNehrutsa and @jonathanhogg (with contributions from many others including @robert-hh and @harbaum). Thank you to everyone involved!!

It combines the design + documentation in PR #8072 with the implementation in PR #7582. This allows us to provide the full-featured interface to the ESP32 hardware, while also providing a lightweight wrapper providing the simplified portable Counter and Encoder classes.

I note that compared to the implementation in #8766 this does not support setting IRQs on the Encoder and Counter classes, however this is still possible with the esp32.PCNT class (and I think we should add support for this to Encoder and Counter anyway).

Also, as @jonathanhogg has pointed out, this uses the now-deprecated IDF pcnt driver, but updating it to the new one isn't that difficult. @projectgus suggests that we should consider moving to the HAL instead, as the driver doesn't offer a significant benefit.

I have made the following additional changes:

esp32.PCNT:
 - Remove the ability to set a value, instead just a reset flag. (The hardware doesn't support it, and requiring the value being set to zero is a bit awkward). (The Encoder and Counter classes still support setting a value though).
 - Minor code cleanups and commenting (especially around some of the more trickier aspects around avoiding race conditions with IRQ handlers).
 - Remove the auto-assigning of unit IDs. We don't do this for other peripherals.

machine.Counter:
 - Simplify the way `__new__` and `__init__` work to implement the singletons.
 - Make the `id` passed to Counter (and Encoder) correspond directly to the PCNT id. This makes Encoder and Counter share the same "namespace" which I think is natural considering they do use the same underlying resource.
 - Make it count overflows rather than pulses, so that it can count for longer before overflowing small integer. This is important to avoid needing to allocate in the IRQ handler. This also changes the way we handle setting a value (and I think also fixes a race condition).
 - Comments to explain the subtleties in avoiding race conditions.

machine.Encoder:
 - Didn't previously work for phases=1 or 2. (Due to the way the pin arguments were handled in PCNT).
 - Make phases=1 the default (not 4). We should provide the same default on all ports, and a port may not be able to provide 2x or 4x quadrature.

Docs:
 - Minor clarifications to ESP32-specific support & availability.
 - Explain the overflow and small integer behavior.

Other:
 - Make this enabled by default on OG,S2,S3 (rather than explicitly disabling on C3).
[...]

PR #17914, opened 2025-08-13
tests/machine_encoder.py: Fix esp32 Encoder: only 3 pulses to count first rotation.

IhorNehrutsa:
Add pin initial level.
Add test Encoder phases x2 and x4.

Explanation to the https://github.com/micropython/micropython/pull/12346#discussion_r2241771478

Output waveform of industrial quadrature encoders:
image

Output waveform of rotary encoders is slightly different:
image
They have several fixed positions per revolution with channel A and channel B output levels equal to 1.
I suspect that @jonathanhogg, @jimmo used exactly this kind of encoder.

He has high levels of input channels and  increment the Encoder on the falling edge of channel A.
https://github.com/micropython/micropython/blob/744270ac1b9ed3929cd41d1a6e1f6ea0e785745d/ports/esp32/modules/machine.py#L161

So, the esp32 test requires initial high levels of channels before starting the count.

Also added mimxrt configuration from the #12347
@robert-hh You can cherry-pick from here too.

### DIFF
--- a/tests/extmod_hardware/machine_encoder.py
+++ b/tests/extmod_hardware/machine_encoder.py
@@ -11,22 +11,16 @@ except ImportError:
     raise SystemExit
 
 import sys
+import unittest
 from machine import Pin
+from target_wiring import encoder_loopback_id, encoder_loopback_out_pins, encoder_loopback_in_pins
 
 PRINT = False
 PIN_INIT_VALUE = 1
 
-if "esp32" in sys.platform:
-    id = 0
-    out0_pin = 4
-    in0_pin = 5
-    out1_pin = 12
-    in1_pin = 13
-else:
-    print("Please add support for this test on this platform.")
-    raise SystemExit
-
-import unittest
+id = encoder_loopback_id
+out0_pin, out1_pin = encoder_loopback_out_pins
+in0_pin, in1_pin = encoder_loopback_in_pins
 
 out0_pin = Pin(out0_pin, mode=Pin.OUT)
 in0_pin = Pin(in0_pin, mode=Pin.IN)
--- a/tests/extmod_hardware/machine_encoder.py
+++ b/tests/extmod_hardware/machine_encoder.py
@@ -13,6 +13,9 @@ except ImportError:
 import sys
 from machine import Pin
 
+PRINT = False
+PIN_INIT_VALUE = 1
+
 if "esp32" in sys.platform:
     id = 0
     out0_pin = 4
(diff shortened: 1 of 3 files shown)


tests/extmod_hardware/machine_encoder.py
```diff
@@ -13,12 +13,22 @@
 import sys
 from machine import Pin
 
+PRINT = False
+PIN_INIT_VALUE = 1
```
projectgus, line 19: The test fails if this is set to any other value, yes? If so, I think it's OK to hard-code it at line 46

projectgus (approved):
Thanks @IhorNehrutsa for the explanation and the updated test. This looks good to me.

@robert-hh do you have any concerns about these changes?

robert-hh:
> @robert-hh do you have any concerns about these changes?

I have, indeed. It works with ESP32. It fails with MIMXRT. The previous test version worked. The reason is about what should be the result of the encoder value with different settings of phases. Looking at a few test lines:
```
        self.assertPosition(0)
        self.rotate(1)
        self.assertPosition(1, 1, 1)
        self.rotate(1)
        self.assertPosition(1, 1, 2)
        self.rotate(1)
        self.assertPosition(1, 2, 3)
        self.rotate(1)
        self.assertPosition(1, 2, 4)  # +4
```
The tests sets up three instances with phases 1, 2, and 4. After one pulse, the test expects all three counter to advance by one. So the expected position is (1,1,1). The way the phases emulation is implemented in MIMXRT, the result is (0,0,1). MIMXRT counts in the positive range at phases=2 at the last one of every 2 transitions, and likewise for phases=1 in the last of every 4 transitions.ESP32 seems to advance the counter at the first of every 2 resp 4 transition block. The phases emulation can be changed, that's not a problem. but what it the proper behavior. Counting before an event is finished like the ESP32 does seems wrong to me. But it may be the behavior defined in a standard. Do you know any reference?

P.S. And yes, setting the `PIN_INIT_VALUE = 0` causes the ESP32 to fail the test as well.

jonathanhogg:
@robert-hh:

Best way to think about the current ESP32 logic is to divide the diagram into four edges, **t1** to **t4**, from left-to-right. In the clockwise direction, those edges are passed through **t1** > **t2** > **t3** > **t4**. In the anti-clockwise direction, they are passed through **t4** > **t3** > **t2** > **t1**.

The 1-phase logic is to always count on **t1**. That means the start of the step in the clockwise direction and the _end_ of the step in the anti-clockwise direction. The 2-phase logic is to count on **t1** and **t3** – so a similar difference depending on which direction you're turning. This is because the logic counts only on edges of the A signal. For 4-phase logic, a second counter channel is used to count on both edges of the B signal as well.

I guess this sort of makes sense if you think about the counting occurring at a place rather than a time: if I very carefully turn the encoder in one direction and then back again, the increment and decrement will happen at exactly the same physical position.

Staring at the setup, I _think_ it would be possible to count on the later edge in either direction for 1- and 2-phase logic, but it would require using both channels and a more complicated configuration. I have no idea whether there is any "right" answer to which edges one should count on with 1- and 2-phase logic. I can find no particular reference. Perhaps because anyone who cares actually uses 4-phase logic…

robert-hh:
@jonathanhogg I do of yourse understand the diagram at the first post. My question is, whether this behavior is a kind of standard and all implementations must match that, or is it the ESP32 implementation and other vendors may implement it differently.

The MIMXRT hardware implements solely phase=4. So the emulation would divide the internal counter value by 2 or 4 to emulate phases=2 or phases=1  And I added now offsets to meet the expected behavior of this specific test script. 

@projectgus Speaking of a connection check. The MIMXRT encoder hardware allows to get the actual state of the input signals. That could be uses as a means to check the connection, if these values are exposed with an API.

jonathanhogg:
@robert-hh: sorry, I wasn't trying to teach you how to suck eggs, but I wanted to be explicit about the ESP32 behaviour with regards to clockwise vs anticlockwise turning in case the MIMXRT behaviour is/was different.

As noted, I'm not aware of any standard for 1-phase or 2-phase counting behaviour. I suspect it is arbitrary choice which edge to count on.

I honestly can't even remember where I got my original logic from. I don't think it would be particularly difficult to change if there is some feeling that the current behaviour is suboptimal. I think matching your floor-divide behaviour would just require swapping to using the reversed B signal edges for 1- and 2-phase logic.

tests/extmod_hardware/machine_encoder.py
robert-hh: The test passes for ESP32 with exactly this setting for `PIN_INIT_VALUE`, and I made the MIMXRT Encoder/Counter emulation for phases that it passes the test as well. 
Changing PIN_INIT_VALUE to 0 lets the ESP32 fail the test.
The improvement of this test is, that it tests all phases options.
I wonder if this PR should not be deferred until the target_wiring PR is merged, and then a suitable target wiring is defined here.

robert-hh (approved):
Besides the aspect of target_wiring this test looks OK.

robert-hh:
On a side note: since support for Encoder/Counter in the Mimxrt por5 is not merged yet, this test will fail for teensy.

tests/extmod_hardware/machine_encoder.py
dpgeorge: The `target_wiring` feature is now merged, so I think it would be good to update this test to use that.  Although it may be better to just get this PR in first, then update later on.

@IhorNehrutsa I'll leave the decision up to you: you can either leave this as-is, or use the new `target_wiring` mechanism.  That would involve moving the pin definitions from this test and instead using something like:
```py
from target_wiring import encoder_loopback_id, encoder_loopback_out_pins, encoder_loopback_in_pins
```
Then in `tests/target_wiring/esp32.py` add:
```py
encoder_loopback_id = 0
encoder_loopback_out_pins = (4, 12)
encoder_loopback_in_pins = (5, 13)
```

projectgus (approved):
Thanks @IhorNehrutsa!

PR #17915, opened 2025-08-13
extmod/modlwip: Remove unused include and functions.

tpwrules:
### Summary

The lwipopts.h is now port-specific.

The `sys_arch_{,un}protect` functions are not used according to line 34 of `extmod/lwip-include/lwipopts_common.h`.

Neither have been touched in nearly a decade.

### Testing

Ensured that W5500-EVB-Pico builds and the resulting binary is unchanged.


tpwrules:
For clarity the motivation for this PR is that I was chasing down some problems and got led down the wrong path because I thought this stuff was still used. So removing it will avoid that chance for others in the future.

dpgeorge (approved):
Thanks for the clean up, this looks good.

(Long ago, it was possible to build the unix port with lwIP support, and use SLIP to create network interfaces, but that support seems to be long lost.)

dpgeorge:
Actually, I think `extmod/lwip-include/arch/*` can also be removed.  Could you do that in this PR as well?

tpwrules:
Done.

dpgeorge (approved):

dpgeorge:
Thanks for updating.

PR #17921, opened 2025-08-14
unix: Add additional testing targets.

jepler:
### Summary

These are convenience targets for running specific tests as a developer. They are more useful that invoking run-tests directly as they take account of the VARIANT= specified on the make command-line.

For instance, you can run all tests matching the regular expression "int" with `make VARIANT=... test//int`. (the new targets are all documented in README.md)

### Testing

I used each of the added targets locally and verified that they seemed to work as intended.

This doesn't do anything that isn't possible to day by properly setting `MICROPY_MICROPYTHON` and invoking run-tests.py directly but I find it's a good QOL enhancement. CircuitPython carries something similar but not quite the same. (note to @dhalbert in the future: if this is accepted, you should take this version and drop the one in circuitpython)

### DIFF
--- a/ports/unix/Makefile
+++ b/ports/unix/Makefile
@@ -254,11 +254,29 @@ endif
 
 include $(TOP)/py/mkrules.mk
 
-.PHONY: test test_full_no_native test_full
+.PHONY: test test_full_no_native test_full test//% test/% test-failures print-failures clean-failures
 
 test: $(BUILD)/$(PROG) $(TOP)/tests/run-tests.py
 	$(eval DIRNAME=ports/$(notdir $(CURDIR)))
 	cd $(TOP)/tests && MICROPY_MICROPYTHON=../$(DIRNAME)/$(BUILD)/$(PROG) ./run-tests.py
+
+test//%: $(BUILD)/$(PROG) $(TOP)/tests/run-tests.py
+	$(eval DIRNAME=ports/$(notdir $(CURDIR)))
+	cd $(TOP)/tests && MICROPY_MICROPYTHON=../$(DIRNAME)/$(BUILD)/$(PROG) ./run-tests.py -i "$*"
+
+test-failures: $(BUILD)/$(PROG) $(TOP)/tests/run-tests.py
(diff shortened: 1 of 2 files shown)


jepler:
There are of course different naming conventions that could be used for the "run a test subset" rules test/% and test//%. Here's the reasoning that led to the current forms:
 * I started with `test-a-%` and `test-d-%` to mirror the commandline flags, but for sub directories to be accepted by the directory test pattern, the `%` had to follow a `/`, so it would have to be `test/%` `test-d/%` or similar. I went with the shortest `test/%`.
 * similarly the regular expression based search would need to be of a similar form if patterns with an embedded / are important. Such patterns might occur, like if you want to test everything that matches the pattern `basics/.*int` for example. It turns out that with `//%` and `/%` patterns, make picks the correct (longest prefix?) one, at least in my testing. If one "/" introduces "here's what to test" (and has to for technical reasons of GNU make), the vi user in me finds "//" as "ditto, but it's a search" to be natural.

Josverl:
And would that pattern work with WindowsPaths \ as well ?

jepler:
That's a good question! I don't have any Windows machines available to test.

Since most path like stuff is handled just fine by make even though Windows uses a different directory separator, I think the part where it gets the test selection into `$*` will be the same.

The run-tests script is careful to replace windows-style directory separators with forward slash ones before applying filters (see around line 884), so I think there's a good chance that `make test//basics/.*int` would work in a cross-platform way.

It would be great if someone can actually test on Windows and advise how to update the readme about any caveats that may exist. If someone does I'll update it.

Josverl:
I found some time.

under MSYS2 + the same changes to the windows makefile - the test selection also works fine.
while make itself runs on plain powershell - the rest of the toolchain not.

So while it works, I don't think its worth the effort to add to the windows makefile.

jepler:
thanks for testing.

dpgeorge (approved):
A nice improvement!  I tested it and it works as expected.  The regex version is neat!

dpgeorge:
Rebased and merged in a69425b533932bbcac0ef463f9e27f79ff2150e3

PR #17922, opened 2025-08-14
mkrules: Always rebuild preprocessor targets.

jepler:
### Summary
These files are only built on demand for developers, and it is a quick process.

Without this, a sequence like this would leave the developer with an outdated `main.pp` to inspect:

```
make build-standard/main.o  # Create dependency information
make build-standard/main.pp
touch input.h
make build-standard/main.pp # Rebuilds now, wouldn't have before
```

### Testing

I locally used the `.pp` targets and they behaved as I stated above.

Initially I wanted to use dependency analysis but this seems the better way based on discussion.

### DIFF
--- a/py/mkrules.mk
+++ b/py/mkrules.mk
@@ -104,7 +104,7 @@ vpath %.cpp . $(TOP) $(USER_C_MODULES)
 $(BUILD)/%.o: %.cpp
 	$(call compile_cxx)
 
-$(BUILD)/%.pp: %.c
+$(BUILD)/%.pp: %.c FORCE
 	$(ECHO) "PreProcess $<"
 	$(Q)$(CPP) $(CFLAGS) -Wp,-C,-dD,-dI -o $@ $<
 


dpgeorge:
> It might be better to just make the `.pp` targets `.PHONY` so that they're always re-made. They're quick enough, and the use case I'm aware of is for a developer who is explicitly invoking make for just that purpose.

Yes, that's probably a better and simpler solution.  They are anyway only built when explicitly asked for, they are never build automatically.

py/mkrules.mk
```diff
@@ -104,7 +104,12 @@ vpath %.cpp . $(TOP) $(USER_C_MODULES)
 $(BUILD)/%.o: %.cpp
 	$(call compile_cxx)
 
-$(BUILD)/%.pp: %.c
+# GNU make ".PHONY" always interprets the prerequisites as literal target names, so just using
+# ".PHONY: $(BUILD)/%.pp" does not work. This is the recommended workaround from the GNU make
+# documentation.
+.PHONY: FORCE
+FORCE:
```
dpgeorge: I don't think these 2 lines are needed, FORCE is already defined in `py/py.mk` (and used elsewhere).

jepler: updated again. sorry for getting in a hurry to update this & not checking whether FORCE was already defined.

dpgeorge:
Thanks for updating.  Rebased and merged in b6cd577cf6ea1de28092a7abe58613c66debd412

PR #17926, opened 2025-08-15
py/mpconfig: Enable CRYPTOLIB, HASHLIB_MD5, HASHLIB_SHA1 if SSL enabled.

dpgeorge:
### Summary

This commit unifies the configuration of MICROPY_PY_CRYPTOLIB, MICROPY_PY_HASHLIB_MD5 and MICROPY_PY_HASHLIB_SHA1, so they are enabled by default if MICROPY_PY_SSL is enabled.  This matches the existing configuration of most of the ports.

With this change, all ports remain the same except:
- reneses-ra now enables MICROPY_PY_CRYPTOLIB, MICROPY_PY_HASHLIB_MD5 and MICROPY_PY_HASHLIB_SHA1.
- rp2 now enables MICROPY_PY_HASHLIB_MD5.

### Testing

Builds will be tested by CI.  Code size diff will tell the story.

Could leave this up to the port, but IMO it's better to have the config unified so ports are more uniform in their configuration.


py/mpconfig.h
```diff
@@ -1860,19 +1860,19 @@ typedef time_t mp_timestamp_t;
...
+#define MICROPY_PY_HASHLIB_SHA1  (MICROPY_PY_SSL)
 #endif
 
 #ifndef MICROPY_PY_HASHLIB_SHA256
 #define MICROPY_PY_HASHLIB_SHA256 (1)
 #endif
 
 #ifndef MICROPY_PY_CRYPTOLIB
-#define MICROPY_PY_CRYPTOLIB (0)
+#define MICROPY_PY_CRYPTOLIB (MICROPY_PY_SSL)
```
projectgus, line 1875: I don't know if this is worth worrying about, but the default definition of `MICROPY_PY_SSL` is lower down in the file than these lines.

projectgus (approved):
Yay for standardisation!

py/mpconfig.h
dpgeorge: I think it should be fine?  It's already the case for some other options (for better or worse), eg `MICROPY_PY_SYS_SETTRACE` is used before the default definition.

dpgeorge:
This PR accidentally disabled these features on non-networking rp2 boards, because they don't have SSL enabled.

See #18492 for a fix for that.

PR #17927, opened 2025-08-15
py/mpconfig: Enable the sys module at all feature levels by default.

dpgeorge:
### Summary

The `sys` module is a pretty fundamental module, and even minimal ports like unix and zephyr minimal have it enabled.  So, enabled it by default at the lowest feature level.

Most things in the `sys` module are configurable, and off by default, so it shouldn't add to much to ports that don't already have it enabled (which is just the minimal port).

### Testing

To be tested by CI.

The aim here is to simplify configuration, and also eventually make it so the minimal configuration can run and pass the test suite.  The `sys` module is something that's needed for introspection of the target, so it seems reasonable to include it by default on everything.  Of course that costs code size, and it's still easy to disable if needed, you just need to do it explicitly now.

Also note that it's still disabled on the bare-arm port, to keep that ultra minimal.  It means we now have bare-arm without `sys` but minimal port does have `sys`.  That will allow different code size comparisons if/when new sys features are added.


dpgeorge:
I'm not 100% sure this is the right thing to do.  It's a choice of:
1. Enable `sys` by default and then have to disable where it's not wanted (eg embedded example, bare-arm).
2. Enable `sys` at a higher feature level and then have to enable it where needed (eg unix, zephyr minimal).

I lean towards option 1, hence this PR.

ports/bare-arm/mpconfigport.h
```diff
@@ -37,6 +37,9 @@
 // Python internal features
 #define MICROPY_ERROR_REPORTING                 (MICROPY_ERROR_REPORTING_NONE)
 
+// Fine control over Python builtins, classes, modules, etc.
+#define MICROPY_PY_SYS                          (0)
```
projectgus, line 41: What's the code size cost if we make the `bare-arm` config the same as `ports/minimal` here?

(I guess what I'm hinting it as whether "minimal" can itself be a feature level?)

projectgus (commented):
This seems like a good idea to me.

There's a follow-up we can do to take some of the `import sys` in tests out of try/except/"SKIP" blocks as well. About 11 cases of that (although only 3 have `import sys` as the only line in the block, the others try to import multiple modules.)

ports/bare-arm/mpconfigport.h
dpgeorge: > What's the code size cost if we make the `bare-arm` config the same as `ports/minimal` here?

It's +412 bytes.

But I kind of wanted to have a build without `sys` (bare-arm) and one with "minimal" `sys` (minimal port), if only to test that MicroPython can build with such configuration.

dpgeorge:
> There's a follow-up we can do to take some of the `import sys` in tests out of try/except/"SKIP" blocks as well.

Yes, good idea.  Although it does mean the test suite requires `sys` to work, but I think that's a reasonable assumption.

ports/bare-arm/mpconfigport.h
projectgus: I see, makes sense!

projectgus (approved):

Issue #17925, opened 2025-08-15
crash when producing a large string via "*"


jepler:
### Port, board and/or hardware

unix port, coverage variant, x86_64 linux

### MicroPython version

v1.27.0-preview-32-g141f7d0c35

### Reproduction

Run micropython with the following script (via `-c` for example): `'a' * 8 * (1 << 62)`

### Expected behaviour

An error is printed because it's not possible to contain the resulting sequence

### Observed behaviour

A crash inside memcpy
```
Starting program: /home/jepler/src/micropython/ports/unix/build-coverage/micropython -c "'aaaa' * (1 << 62)"
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
__memcpy_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:366
warning: 366	../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S: No such file or directory
(gdb) where
#0  __memcpy_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:366
#1  0x0000555555619c62 in mp_seq_multiply (items=items@entry=0x7ffff7c15c20, item_sz=item_sz@entry=1, len=8, 
    times=4611686018427387904, dest=0x7ffff7c97000) at ../../py/sequence.c:41
#2  0x0000555555610ef1 in mp_obj_str_binary_op (op=MP_BINARY_OP_MULTIPLY, lhs_in=<optimized out>, 
    rhs_in=<optimized out>) at ../../py/objstr.c:391
```

### Additional Information

Many variants (such as `1<<59`) instead give the message `MemoryError: memory allocation failed, allocating 1 bytes` but I'm not sure where the value 1 is coming from.

### Code of Conduct

Yes, I agree

GitHubsSilverBullet:
> Run micropython with the following script (via `-c` for example): `'a' * 8 * (1 << 62)`

Wouldn't that try to allocate 32 ExbiByte (36,9 ExaByte)
Plus all the intermediate allocations.
Sound like a lot to me. Really, a lot!

jepler:
Yes it's completely unreasonble allocation attempt and should fail with MemoryError.

GitHubsSilverBullet:
Do you want to keep it open?

[...]

PR #17931, opened 2025-08-15
runtime: Fix printing failed allocation amounts.

jepler:
### Summary

On LP64 and LLP64 systems, size_t is bigger than unsigned. Printing the failed allocation as mp_uint_t allows the correct failed allocation size to be shown. I noticed this in #17925.

However, there are occasions where mp_uint_t is bigger than size_t (nanbox).  In that case, preserve the existing code path to avoid growth in executable size.

Example where this affects the failed allocation message (on x86_64 coverage build):
```
>>> "a" * (1 << 54)
```

Before, this would print the size as 1.  Now it prints it as 18014398509481985 (2**54 + 1).

### Testing

I built locally and verified that large failed allocations showed the correct value.

The code ends up more complicated than it was previously, with the hope of avoiding code growth.

I can remove the added size assertions if preferred. I did want to see if the assertions were correct across all builds done by CI.

### DIFF
--- a/py/runtime.c
+++ b/py/runtime.c
@@ -1684,14 +1684,21 @@ mp_obj_t mp_parse_compile_execute(mp_lexer_t *lex, mp_parse_input_kind_t parse_i
 #endif // MICROPY_ENABLE_COMPILER
 
 MP_NORETURN void m_malloc_fail(size_t num_bytes) {
-    DEBUG_printf("memory allocation failed, allocating %u bytes\n", (uint)num_bytes);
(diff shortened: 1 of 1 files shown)


dpgeorge:
I guess we really need a `%zu`?  That would seem to be the "proper" fix here, although it would definitely increase code size.

Maybe instead we can have a format macro like `SIZE_FMT`?

dpgeorge:
@jepler what do you think about adding a `SIZE_FMT` macro?

PR #17938, opened 2025-08-16
machine.UART: Fix the double timeout with uart.read(), uart.write() and uart.readinto().

robert-hh:
### Summary

When uart.read(), uart.write() and uart_readinto() timed out because less data was transferred than expected, but some, the delay was uart.timeout + uart.timeout.char instead of just uart,timeout_char().

This PR changes the behavior into the expected one. It consists of 3 commits. The first changes uart.read() and uart.write(). The impact of that change is limited to the machine.UART class. The second commit changes stream.readinto() and may affect all uses of that function.
The third commit changes the ESP32's  machine_uart.c to deal properly with the timeout_char argument for uart.read().

uart.readline() is unchanged.

This PR addresses issue #17611.

### Testing

Tested with STM32(Pyboard V1.1 and PYBD_SF6), RP2 Pico, MIMXRT(Teensy 4.1), SAMD51, ESP32, NRF (nrf52840) and a CC3200 board using strings of varying length. Typical results:

```
RP2

UART(0, baudrate=38400, bits=8, parity=None, stop=1, tx=0, rx=1, txbuf=256, rxbuf=256, timeout=200, timeout_char=50, invert=None, irq=0)

uart.read() msg=""  timeout=201 ms
uart.read() msg="123"  timeout=52 ms
uart.read() msg="12345"  timeout=1 ms

uart.readline() msg=""  timeout=201 ms
uart.readline() msg="12345"  timeout=202 ms
uart.readline() msg="12345\n"  timeout=2 ms

uart.readinto(bytearray(5)) msg="" timeout=201 ms
uart.readinto(bytearray(5)) msg="123" timeout=52 ms
uart.readinto(bytearray(5)) msg="12345" timeout=1 ms

CC3200

UART(1, pins=(Pin("GP16"), Pin("GP17")))

uart.read() msg=""  timeout=36 ms
uart.read() msg="123"  timeout=38 ms (from 74 ms without the change)
uart.read() msg="12345"  timeout=5 ms

uart.readline() msg=""  timeout=35 ms
uart.readline() msg="12345"  timeout=41 ms
uart.readline() msg="12345\n"  timeout=6 ms

uart.readinto(bytearray(5)) msg="" timeout=35 ms
uart.readinto(bytearray(5)) msg="123" timeout=39 ms
uart.readinto(bytearray(5)) msg="12345" timeout=5 ms
```

If the impact of the changes to a stream module seems unsafe, the second commit may be omitted, causing the behavior of uart.readinto() unchanged.


py/stream.c
```diff
@@ -303,7 +303,7 @@ static mp_obj_t stream_readinto(size_t n_args, const mp_obj_t *args) {
     }
 
     int error;
-    mp_uint_t out_sz = mp_stream_read_exactly(args[0], bufinfo.buf, len, &error);
+    mp_uint_t out_sz = mp_stream_rw(args[0], bufinfo.buf, len, &error, MP_STREAM_RW_READ | MP_STREAM_RW_ONCE);
```
dpgeorge: I need to think in detail about this change.  It will change behaviour for all users of the `readinto()` method.

CPython actually provides `readinto1()` for this case, but that's defined on `io.BufferedIOBase`: https://docs.python.org/3/library/io.html#io.BufferedIOBase.readinto1 .

I think we match `io.RawIOBase` better, and actually the semantics of `RawIOBase.readinto()` seem to be that it does a single underlying call to do the read, ie read-once.  See the explanation for `BufferedIOBase`:

> The main difference with [RawIOBase](https://docs.python.org/3/library/io.html#io.RawIOBase) is that methods [read()](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read), [readinto()](https://docs.python.org/3/library/io.html#io.BufferedIOBase.readinto) and [write()](https://docs.python.org/3/library/io.html#io.BufferedIOBase.write) will try (respectively) to read as much input as requested or to emit all provided data.

dpgeorge: Maybe we can introduce an underlying `mp_stream_readinto1_obj` method and then use it for UART `readinto`?  That would leave other streams that use readinto unchanged.

Adding readinto1 would add a bit more code size, but I think there are ways we could mitigate that by factoring code a bit more.  That code size optimisation can be done in a follow up PR.

robert-hh: Sure. I can implement that.

robert-hh: Can the `mp_stream_readinto1_obj` be simplified? like e.g. not accepting the second argument?
And should it have a configuration switch for boards with low memory, like `MICROPY_PY_STREAM_READINTO1`?
Edit: The size increment is 96 bytes for ESP32_GENERIC, 112bytes for RP2.

dpgeorge: > Can the `mp_stream_readinto1_obj` be simplified? like e.g. not accepting the second argument?

Probably not, unfortunately.  Because it would be a breaking change.  Users may be already using `UART.readinto` with a second argument.

> And should it have a configuration switch for boards with low memory, like `MICROPY_PY_STREAM_READINTO1`?

No.  Only ports that use this method will include it in their build/firmware, so it doesn't need a config switch.

robert-hh: PR updated with the change, which is just duplicating the `mp_stream_readinto_obj` with a single changed line. Size increase about 100 bytes among the ports.

dpgeorge: Thanks, that looks good.

If you want to make a quick and easy optimisation (which would also make this PR shorter), I suggest following how `stream_read` and `stream_read1` work, calling `stream_read_generic` (so you'd define a `stream_readinto_generic`, etc).

robert-hh: Done.  The size increase is now about 50 bytes.

robert-hh: One behavior that is changed and may cause some irritation for existing user code:
If `timeout_char` is not set, it is 0. Before it was then set to the duration of a byte. But that did not matter because it was effectively the same as `timeout`. Now the values matter, which was the intention of this change. 
Now I'm wondering if the default for `timeout_char` should be set to `timeout` to keep consistency with the previous behavior. However, that needs to change every port's `machine_uart` implementation and the docs.

Edit: On a second thought we should leave it as it is. Code that relied on a wrong behavior must be changed.

extmod/machine_uart.c
```diff
@@ -140,10 +140,10 @@ static const mp_rom_map_elem_t machine_uart_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_deinit), MP_ROM_PTR(&machine_uart_deinit_obj) },
 
     { MP_ROM_QSTR(MP_QSTR_flush), MP_ROM_PTR(&mp_stream_flush_obj) },
-    { MP_ROM_QSTR(MP_QSTR_read), MP_ROM_PTR(&mp_stream_read_obj) },
+    { MP_ROM_QSTR(MP_QSTR_read), MP_ROM_PTR(&mp_stream_read1_obj) },
     { MP_ROM_QSTR(MP_QSTR_readline), MP_ROM_PTR(&mp_stream_unbuffered_readline_obj) },
-    { MP_ROM_QSTR(MP_QSTR_readinto), MP_ROM_PTR(&mp_stream_readinto_obj) },
-    { MP_ROM_QSTR(MP_QSTR_write), MP_ROM_PTR(&mp_stream_write_obj) },
+    { MP_ROM_QSTR(MP_QSTR_readinto), MP_ROM_PTR(&mp_stream_readinto1_obj) },
+    { MP_ROM_QSTR(MP_QSTR_write), MP_ROM_PTR(&mp_stream_write1_obj) },
```
dpgeorge, line 146: Unfortunately using `write1` here will be a breaking change, because it doesn't accept the additional arguments like `write` does.

So, we'll need to use the same approach as you did with `stream_readinto_generic`, to make a `stream_write_generic(size_t n_args, const mp_obj_t *args, byte flags)`.

ports/cc3200/mods/pybuart.c
```diff
@@ -579,13 +579,13 @@ static const mp_rom_map_elem_t pyb_uart_locals_dict_table[] = {
     { MP_ROM_QSTR(MP_QSTR_txdone),      MP_ROM_PTR(&machine_uart_txdone_obj) },
 
     /// \method read([nbytes])
-    { MP_ROM_QSTR(MP_QSTR_read),        MP_ROM_PTR(&mp_stream_read_obj) },
+    { MP_ROM_QSTR(MP_QSTR_read),        MP_ROM_PTR(&mp_stream_read1_obj) },
     /// \method readline()
     { MP_ROM_QSTR(MP_QSTR_readline),    MP_ROM_PTR(&mp_stream_unbuffered_readline_obj) },
     /// \method readinto(buf[, nbytes])
     { MP_ROM_QSTR(MP_QSTR_readinto),    MP_ROM_PTR(&mp_stream_readinto_obj) },
```
dpgeorge: Should this use `readinto1` as well?

py/stream.c
```diff
@@ -303,7 +303,12 @@ static mp_obj_t stream_readinto(size_t n_args, const mp_obj_t *args) {
     }
 
     int error;
-    mp_uint_t out_sz = mp_stream_read_exactly(args[0], bufinfo.buf, len, &error);
+    mp_uint_t out_sz;
+    if (flags & MP_STREAM_RW_ONCE) {
```
dpgeorge: This `if` can be simplified to a single line:
```
mp_stream_rw(args[0], bufinfo.buf, len, &error, flags);
```
and then pass in `MP_STREAM_RW_READ` as flags from `stream_readinto()`, and `MP_STREAM_RW_READ | MP_STREAM_RW_ONCE` from `stream_readinto1()`.

py/stream.c
dpgeorge: > If `timeout_char` is not set, it is 0

I don't follow.  On eg the rp2 port, `timeout_char` is always at least as long as a single byte, even if the user sets it to 0.

I think the only behaviour that changes with the UART is that, after reading some bytes and timing out, it doesn't try to read more bytes (and hence wait another `timeout` time).

ports/cc3200/mods/pybuart.c
robert-hh: Indeed. The CC3200 is always a little bit out of sight.

robert-hh: Changed & tested. The timeout and timeout_char situation have the same timeout. These values are fixed with the CC3200 port and are set to 16 character times, as far as I read the code. What I see in testing is a delay of about 36 character times.

py/stream.c
robert-hh: Done, good hint.

extmod/machine_uart.c
robert-hh: Which additional arguments are expected in the context of machine_uart's write()? Obviously uart.write() works in the test I made.

dpgeorge: The signature is:
```
stream.write(buf)
stream.write(buf, max_len)
stream.write(buf, off, max_len)
```
I don't think this is actually documented anywhere, but it may be relied upon in some code.

robert-hh: Thank you. I did not find any documentation about it. It it safe then to change write1 to accept the additional arguments? The only places where it seems to be referenced were `extmod/modtls_mbedtls.c` and `unix/coverage.c`.

dpgeorge: > It it safe then to change write1 to accept the additional arguments?

Yes, it's safe, because it's an extension of the existing signature.

Ideally we would not support these extra arguments, which are a MicroPython specific extension (extension to CPython).  Maybe we can remove them in MicroPython v2.0.

But for now, they are needed.  Eg `umqtt.simple` uses the second argument, eg `self.sock.write(pkt, 2)`.

robert-hh: Change implemented & tested with a RP2 board. Both timeout and additional write() arguments work as expected.

dpgeorge (approved):

Tested on PYBV10, the `machine.UART` tests all pass.

robert-hh:
Thank you very much for merging and guiding me.

PR #17939, opened 2025-08-16
zephyr: Introduce Zephyr displays

VynDragon:
### Summary
Allows using zephyr displays as framebuf

### Testing
SSD1327 + RP2040
and SSD1357 + RP2040

### DIFF
--- a/docs/library/zephyr.Display.rst
+++ b/docs/library/zephyr.Display.rst
@@ -1,0 +1,68 @@
+.. currentmodule:: zephyr
+.. _zephyr.Display:
+
+class Display -- access to Zephyr Displays
+==========================================
+
+Uses the `Zephyr Display API <https://docs.zephyrproject.org/latest/doxygen/html/group__display__interface.html>`_.
+
+This class allows access to Zephyr-provided Displays
+(`zephyr,display chosen Node <https://docs.zephyrproject.org/latest/build/dts/api/api.html#zephyr-specific-chosen-nodes>`_
+and `zephyr,displays Node <https://docs.zephyrproject.org/latest/build/dts/api/bindings/display/zephyr%2Cdisplays.html>`_)
+via an API reproducing the Zephyr one.
+
+Constructors
+------------
+
+.. class:: Display(id)
+
+   Gets an object for accessing a Display identified by ``id``.
+
+   ``id`` can be an integer (``0``, ``1``...) or a string (``"ssd1306@3c"``) identifying a display node by its position or by its node identifiers.
+
+Methods
+-------
+
+.. method:: Display.write(buf[, x[, y[, size_x[, size_y]]]])
+
+            Write a buffer-protocol object in the Display's Pixel Format to the display.
+
+            Optionally x and y position, x size, and y size can be specified.
+
+.. method:: Display.rgb(r, g, b)
+
+            Convert a RGB color to the Display's Pixel Format.
+
+.. method:: Display.capabilities()
+
+            Retrieve a tuple describing the display in the format:
+
+            ``(X Size, Y Size, Supported PFs, Current PF, Current Orientation, Misc Characteristics, Current PF as framebuf format)``
+
+.. method:: Display.format([format])
+
+            Get and set the Pixel Format of the Display.
+
+.. method:: Display.blanking(value)
+
+            Enable or disable blanking.
+
+.. method:: Display.clear()
+
+            Clear the Display.
+
+.. method:: Display.set_brightness(value)
+
+            Set the Display's brightness from ``0`` to ``255``.
+
+.. method:: Display.set_contrast(value)
+
+            Set the Display's contrast from ``0`` to ``255``.
+
+.. method:: Display.orientation([orientation])
+
+            Get and set the Orientation of the Display.
+
+.. method:: Display.as_framebuf()
+
+            If :mod:`framebuf` is enabled, generate a :class:`framebuf.FrameBuffer` instance augmented with a ``show()`` function that directly maps to the display with the currently configured settings.
--- a/ports/zephyr/modzephyr.c
+++ b/ports/zephyr/modzephyr.c
@@ -89,6 +89,9 @@ static const mp_rom_map_elem_t mp_module_time_globals_table[] = {
     #ifdef CONFIG_FILE_SYSTEM
     { MP_ROM_QSTR(MP_QSTR_FileSystem), MP_ROM_PTR(&zephyr_filesystem_type) },
     #endif
+    #if defined(CONFIG_DISPLAY)
+    { MP_ROM_QSTR(MP_QSTR_Display), MP_ROM_PTR(&zephyr_display_type) },
+    #endif
 };
 
 static MP_DEFINE_CONST_DICT(mp_module_time_globals, mp_module_time_globals_table);
--- a/ports/zephyr/zephyr_display.c
+++ b/ports/zephyr/zephyr_display.c
@@ -1,0 +1,354 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2025 MASSDRIVER EI (massdriver.space)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include <zephyr/device.h>
+#include <zephyr/drivers/display.h>
+#include <stdlib.h>
+
+#include "modzephyr.h"
+#include "zephyr_device.h"
+#include "py/runtime.h"
+
+#if defined(CONFIG_DISPLAY)
+
+#if DT_HAS_COMPAT_STATUS_OKAY(zephyr_displays)
+#define ENUMERATE_DISPLAY_DEVS(node_id, prop, idx) DEVICE_DT_GET(DT_PROP_BY_IDX(node_id, prop, idx)),
+
+static const struct device *zephyr_display_devices[DT_ZEPHYR_DISPLAYS_COUNT] = {
+    DT_FOREACH_PROP_ELEM(DT_COMPAT_GET_ANY_STATUS_OKAY(zephyr_displays), displays, ENUMERATE_DISPLAY_DEVS)
+};
+#elif DT_HAS_CHOSEN(zephyr_display)
+static const struct device *zephyr_display_devices[] = {
+    DEVICE_DT_GET(DT_CHOSEN(zephyr_display)),
+};
+#else
+static const struct device *zephyr_display_devices[1];
+#endif
+
+const mp_obj_type_t zephyr_display_type;
+
+typedef struct _zephyr_display_obj_t {
+    mp_obj_base_t base;
+    const struct device *dev;
+} zephyr_display_obj_t;
+
+static void zephyr_display_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    mp_printf(print, "Display(\"%s\")", self->dev->name);
+}
+
+static mp_obj_t zephyr_display_make_new(const mp_obj_type_t *type,
+    size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, 1, false);
+    const struct device *dev;
+
+    if (mp_obj_is_str(args[0])) {
+        dev = zephyr_device_find(args[0]);
+    } else {
+        int id = mp_obj_get_int(args[0]);
+
+        if (id < 0 || id >= DT_ZEPHYR_DISPLAYS_COUNT) {
+            mp_raise_ValueError(MP_ERROR_TEXT("Invalid display ID"));
+        }
+        dev = zephyr_display_devices[id];
+    }
+    if (!device_is_ready(dev)) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Display is not ready"));
+    }
+
+    zephyr_display_obj_t *self = mp_obj_malloc(zephyr_display_obj_t, type);
+    self->dev = dev;
+    return MP_OBJ_FROM_PTR(self);
+}
+
+static mp_obj_t zephyr_display_write(size_t n_args, const mp_obj_t *args_in) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(args_in[0]);
+    struct display_capabilities capabilities;
+    int ret;
+    size_t len;
+    mp_buffer_info_t bufinfo;
+    int x = 0;
+    int y = 0;
+    int sx;
+    int sy;
+
+    mp_get_buffer_raise(args_in[1], &bufinfo, MP_BUFFER_READ);
+
+    display_get_capabilities(self->dev, &capabilities);
+    sx = capabilities.x_resolution;
+
+    if (n_args > 2) {
+        x = mp_obj_get_int(args_in[2]);
+        y = mp_obj_get_int(args_in[3]);
+    }
+
+    if (n_args > 4) {
+        sx = mp_obj_get_int(args_in[4]);
+    }
+    sy = bufinfo.len / (sx * DISPLAY_BITS_PER_PIXEL(capabilities.current_pixel_format) / 8);
+
+    if (n_args > 5) {
+        sy = mp_obj_get_int(args_in[5]);
+    }
+
+    len = sx * sy * DISPLAY_BITS_PER_PIXEL(capabilities.current_pixel_format) / 8;
+
+    if (len > bufinfo.len) {
+        mp_raise_ValueError(MP_ERROR_TEXT("Buffer is shorter than size requires"));
+    }
+
+    const struct display_buffer_descriptor desc = {
+        .buf_size = len,
+        .width = sx,
+        .height = sy,
+        .pitch = sx,
+    };
+
+    ret = display_write(self->dev, x, y, &desc, bufinfo.buf);
+    if (ret < 0) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Couldn't write to display"));
+    }
+    return mp_obj_new_int(ret);
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(zephyr_display_write_obj, 2, 6, zephyr_display_write);
+
+
+static mp_obj_t zephyr_display_rgb(size_t n_args, const mp_obj_t *args_in) {
+    (void)n_args;
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(args_in[0]);
+    struct display_capabilities capabilities;
+
+    int r = mp_obj_get_int(args_in[1]);
+    int g = mp_obj_get_int(args_in[2]);
+    int b = mp_obj_get_int(args_in[3]);
+
+    display_get_capabilities(self->dev, &capabilities);
+
+    switch (capabilities.current_pixel_format) {
+        case PIXEL_FORMAT_MONO10:
+            if (r > 127 || g > 127 || b > 127) {
+                return mp_obj_new_int(0x0);
+            } else {
+                return mp_obj_new_int(0xFF);
+            }
+        case PIXEL_FORMAT_MONO01:
+            if (r > 127 || g > 127 || b > 127) {
+                return mp_obj_new_int(0xFF);
+            } else {
+                return mp_obj_new_int(0x0);
+            }
+        case PIXEL_FORMAT_L_8:
+            return mp_obj_new_int((r + g + b) / 3);
+        case PIXEL_FORMAT_RGB_565:
+            return mp_obj_new_int((r & 0xe0) | ((g >> 3) & 0x1c) | (b >> 6));
+        default:
+            break;
+    }
+    mp_raise_ValueError(MP_ERROR_TEXT("Not a framebuf pixel format"));
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(zephyr_display_rgb_obj, 4, 4, zephyr_display_rgb);
+
+static mp_obj_t zephyr_display_framebuf_current_format(const mp_obj_t self_in) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    struct display_capabilities capabilities;
+    int format = -1;
+
+    display_get_capabilities(self->dev, &capabilities);
+
+    switch (capabilities.current_pixel_format) {
+        case PIXEL_FORMAT_MONO10:
+        case PIXEL_FORMAT_MONO01:
+            if (capabilities.screen_info & SCREEN_INFO_MONO_VTILED
+                && capabilities.screen_info & SCREEN_INFO_MONO_MSB_FIRST) {
+                format = -1; // no VMSB
+            } else if (capabilities.screen_info & SCREEN_INFO_MONO_VTILED) {
+                format = 0; // FRAMEBUF_MVLSB
+            } else if (capabilities.screen_info & SCREEN_INFO_MONO_MSB_FIRST) {
+                format = 4; // FRAMEBUF_MHMSB
+            } else {
+                format = 3; // FRAMEBUF_MHLSB
+            }
+            break;
+        case PIXEL_FORMAT_L_8:
+            format = 6; // FRAMEBUF_GS8
+            break;
+        case PIXEL_FORMAT_RGB_565:
+            format = 1; // FRAMEBUF_RGB565
+            break;
+        default:
+            format = -1;
+    }
+    if (format < 0) {
+        mp_raise_ValueError(MP_ERROR_TEXT("Not a framebuf pixel format"));
+    }
+    return mp_obj_new_int(format);
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_1(zephyr_display_framebuf_current_format_obj, zephyr_display_framebuf_current_format);
+
+static mp_obj_t zephyr_display_formats(const mp_obj_t self_in) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    struct display_capabilities capabilities;
+    mp_obj_t tuple_buf[32];
+    size_t tuple_cnt = 0;
+
+    display_get_capabilities(self->dev, &capabilities);
+
+    for (int i = 0; i < 32; i++) {
+        if (capabilities.supported_pixel_formats & 1U << i) {
+            tuple_buf[tuple_cnt] = mp_obj_new_int(1U << i);
+            tuple_cnt++;
+        }
+    }
+
+    return mp_obj_new_tuple(tuple_cnt, tuple_buf);
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_1(zephyr_display_formats_obj, zephyr_display_formats);
+
+static mp_obj_t zephyr_display_current_format(size_t n_args, const mp_obj_t *args_in) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(args_in[0]);
+    struct display_capabilities capabilities;
+    int ret;
+
+    if (n_args > 1) {
+        ret = display_set_pixel_format(self->dev, mp_obj_get_int(args_in[1]));
+        if (ret < 0) {
+            mp_raise_ValueError(MP_ERROR_TEXT("Invalid pixel format"));
+        }
+    }
+
+    display_get_capabilities(self->dev, &capabilities);
+
+    return mp_obj_new_int(capabilities.current_pixel_format);
+}
+static MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(zephyr_display_current_format_obj, 1, 2, zephyr_display_current_format);
+
+static mp_obj_t zephyr_display_blanking_on(const mp_obj_t self_in) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+
+    if (display_blanking_on(self->dev) < 0) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Couldn't set blanking"));
+    }
+
+    return mp_const_none;
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_1(zephyr_display_blanking_on_obj, zephyr_display_blanking_on);
+
+static mp_obj_t zephyr_display_blanking_off(const mp_obj_t self_in) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+
+    if (display_blanking_off(self->dev) < 0) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Couldn't set blanking"));
+    }
+
+    return mp_const_none;
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_1(zephyr_display_blanking_off_obj, zephyr_display_blanking_off);
+
+static mp_obj_t zephyr_display_set_brightness(const mp_obj_t self_in, mp_obj_t arg) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+
+    if (display_set_brightness(self->dev, mp_obj_get_int(arg)) < 0) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Couldn't set brightness"));
+    }
+
+    return mp_const_none;
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_2(zephyr_display_set_brightness_obj, zephyr_display_set_brightness);
+
+static mp_obj_t zephyr_display_set_contrast(const mp_obj_t self_in, mp_obj_t arg) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+
+    if (display_set_contrast(self->dev, mp_obj_get_int(arg)) < 0) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Couldn't set contrast"));
+    }
+
+    return mp_const_none;
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_2(zephyr_display_set_contrast_obj, zephyr_display_set_contrast);
+
+static mp_obj_t zephyr_display_set_orientation(const mp_obj_t self_in, mp_obj_t arg) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+
+    if (mp_obj_get_int(arg) > DISPLAY_ORIENTATION_ROTATED_270 || mp_obj_get_int(arg) < 0) {
+        mp_raise_ValueError(MP_ERROR_TEXT("Invalid orientation"));
+    }
+
+    if (display_set_orientation(self->dev, mp_obj_get_int(arg)) < 0) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Couldn't set Orientation"));
+    }
+
+    return mp_const_none;
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_2(zephyr_display_set_orientation_obj, zephyr_display_set_orientation);
+
+static const mp_rom_map_elem_t zephyr_display_locals_dict_table[] = {
+    /* this class */
+    { MP_ROM_QSTR(MP_QSTR_write), MP_ROM_PTR(&zephyr_display_write_obj) },
+    { MP_ROM_QSTR(MP_QSTR_rgb), MP_ROM_PTR(&zephyr_display_rgb_obj) },
+    { MP_ROM_QSTR(MP_QSTR_framebuf_current_format), MP_ROM_PTR(&zephyr_display_framebuf_current_format_obj) },
+    { MP_ROM_QSTR(MP_QSTR_formats), MP_ROM_PTR(&zephyr_display_formats_obj) },
+    { MP_ROM_QSTR(MP_QSTR_current_format), MP_ROM_PTR(&zephyr_display_current_format_obj) },
+    { MP_ROM_QSTR(MP_QSTR_blanking_on), MP_ROM_PTR(&zephyr_display_blanking_on_obj) },
+    { MP_ROM_QSTR(MP_QSTR_blanking_off), MP_ROM_PTR(&zephyr_display_blanking_off_obj) },
+    { MP_ROM_QSTR(MP_QSTR_set_brightness), MP_ROM_PTR(&zephyr_display_set_brightness_obj) },
+    { MP_ROM_QSTR(MP_QSTR_set_contrast), MP_ROM_PTR(&zephyr_display_set_contrast_obj) },
+    { MP_ROM_QSTR(MP_QSTR_set_orientation), MP_ROM_PTR(&zephyr_display_set_orientation_obj) },
+    { MP_ROM_QSTR(MP_QSTR_display_count), MP_ROM_INT(DT_ZEPHYR_DISPLAYS_COUNT) },
+    { MP_ROM_QSTR(MP_QSTR_ORIENTATION_NORMAL), MP_ROM_INT(DISPLAY_ORIENTATION_NORMAL) },
+    { MP_ROM_QSTR(MP_QSTR_ORIENTATION_90), MP_ROM_INT(DISPLAY_ORIENTATION_ROTATED_90) },
+    { MP_ROM_QSTR(MP_QSTR_ORIENTATION_180), MP_ROM_INT(DISPLAY_ORIENTATION_ROTATED_180) },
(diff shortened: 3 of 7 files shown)


Gadgetoid:
Looking over this briefly, the duplicate methods with delegation to framebuf feel like they would be a nuisance to maintain. Anyone changing framebuf would need to check for compatibility breaks in Zephyr Display and I'm not sure that's feasible.

Is it possible to retrieve the buffer from a Zephyr Display and use that to initialise a vanilla instance of framebuf module? If so, some of the complexity of doing so could be swept under the rug with a Python module that sets the correct pixel type, dimensions, etc.

In such a case, ZephyrDisplay should just handle providing the buffer and some basic display config information and it would be compatible with not just framebuf, but any means by which a user cares to set pixels (ulab's numpy implementation for example.)

VynDragon:
Zephyr doesnt provide a buffer to write into so it's not possible to use that directly.

But it sounds like maybe it should be possible to have the zephyrdisplay allocate the appropriate buffer and provide that instead, with other attributes to provide to framebuf (display format, etc? that's what you mean i think)

Maybe though, allocating the framebuf in the zephyrdisplay, and then having a method return that framebuf instead would be a better solution?

VynDragon:
Actually thinking about it there really is no point allocating the buffer for the user anyway if it's not fully automated /subclassing framebuf, instead the API from zephyr should be duplicated more closely, and allow mpy to retrieve the display infos, then send whatever buffer the user wants to send at x,y coordinate and of z size, like the zephyr API.

Maybe both should coexist (ZephyrDisplayFramebuf and ZephyrDisplay?)

josuah:
Possibly relevant: LVGL Zephyr integration for MicorPython: `modlvzephyr.c` in `lv_binding_micropython`
https://github.com/lvgl/lv_binding_micropython/tree/master/driver/zephyr

Gadgetoid:
> Zephyr doesnt provide a buffer to write into

Ah! In that case yes - FrameBuf should allocate the buffer, and ZephyrDisplay should supply the dimensions and format of buffer the display needs. (Looks like it supplies compatible formats and the user picks one, which is nice)

> the API from zephyr should be duplicated more closely, and allow mpy to retrieve the display infos, then send whatever buffer the user wants to send at x,y coordinate and of z size, like the zephyr API.

Absolutely this. It should be completely agnostic to whatever provides the framebuffer, so long as it is in the correct (or a supported?) pixel format and dimensions. It could be a `numpy.array` or a `bytearray`, it's all just bytes.

It looks like Zephyr's display interface (this? https://docs.zephyrproject.org/latest/doxygen/html/group__display__interface.html) is already well suited to this pattern.

For stuff like `supported_pixel_formats` I'd just to return the Zephyr ones and - as you suggest - supply a `ZephyrDisplayFrameBuf` module. Crucially I'd make this a Python module (which should be added to micropython-lib) since that's sort of the prevailing pattern for HAL vs glue code/user-facing API stuff (See Bluetooth for example).

So yes- agnostic, 1:1 implementation of the Zephyr display interface, and a `micropython-lib` library to glue `FrameBuf` and `ZephyrDisplay` together feels right in my very humble opinion.

Nice work though, `FrameBuf` needs more love 😆

VynDragon:
@Gadgetoid Updated to be like the API

ports/zephyr/modzephyr.c
```diff
@@ -86,6 +86,9 @@ static const mp_rom_map_elem_t mp_module_time_globals_table[] = {
     #ifdef CONFIG_FLASH_MAP
     { MP_ROM_QSTR(MP_QSTR_FlashArea), MP_ROM_PTR(&zephyr_flash_area_type) },
     #endif
+    #if defined(CONFIG_DISPLAY) && defined(MICROPY_PY_FRAMEBUF)
+    { MP_ROM_QSTR(MP_QSTR_Display), MP_ROM_PTR(&zephyr_display_type) },
+    #endif
```
josuah: Now that there is no inheritance from framebuf, the `&& defined(MICROPY_PY_FRAMEBUF)` can go?

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,333 @@
...
+
+static void zephyr_display_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
+    zephyr_display_obj_t *self = self_in;
+    mp_printf(print, "Display(%d)", self->id);
+}
+
+static mp_obj_t zephyr_display_make_new(const mp_obj_type_t *type,
+    size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, 1, false);
+    int id = mp_obj_get_int(args[0]);
```
josuah: Looking at other drivers, they use strings for creating new instances:

https://github.com/micropython/micropython/blob/b5fcb33eaa682bb666c839cd4fb301175cc3564f/ports/zephyr/machine_spi.c#L85

https://github.com/micropython/micropython/blob/b5fcb33eaa682bb666c839cd4fb301175cc3564f/ports/zephyr/machine_uart.c#L166

https://github.com/micropython/micropython/blob/b5fcb33eaa682bb666c839cd4fb301175cc3564f/ports/zephyr/modzsensor.c#L45

It seems like the first PR that uses `DT_CHOSEN()` mapped to numbers instead of strings, but it feels the right way to map chosen nodes IMHO.

To help the user figure whether to use a string or integer, some rule of thumb would be something like this?

- `Display(0)` valid whenever a module supports `DT_CHOSEN()` (like you did now)
- `Display("labelname")` always available to get a Zephyr device by string name

Using strings can be easier to implement as it's just a [`zephyr_device_find()`](https://github.com/micropython/micropython/blob/b5fcb33eaa682bb666c839cd4fb301175cc3564f/ports/zephyr/zephyr_device.h#L30).

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,333 @@
...
+}
+
+static MP_DEFINE_CONST_FUN_OBJ_2(zephyr_display_set_contrast_obj, zephyr_display_set_contrast);
+
+static mp_obj_t zephyr_display_set_orientation(const mp_obj_t self_in, mp_obj_t arg) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+
+    if (mp_obj_get_int(arg) > MP_QSTR_ROTATION_270 || mp_obj_get_int(arg) < 0) {
+        mp_raise_ValueError("Invalid orientation");
+    }
```
josuah: Maybe it is just me who misunderstood QSTR, but it looks like this is a comparison between a string index and a value as an integer here?

Maybe you meant to have an extra set of define like this:

```c
#define DISPLAY_ROTATION_NONE 0
#define DISPLAY_ROTATION_90   1
#define DISPLAY_ROTATION_180  2
#define DISPLAY_ROTATION_270  3
```

And then use them in the `zephyr_display_locals_dict_table`:
```c
    { MP_ROM_QSTR(MP_QSTR_ROTATION_NONE), MP_ROM_INT(DISPLAY_ROTATION_NONE) },
    { MP_ROM_QSTR(MP_QSTR_ROTATION_90), MP_ROM_INT(DISPLAY_ROTATION_90) },
    { MP_ROM_QSTR(MP_QSTR_ROTATION_180), MP_ROM_INT(DISPLAY_ROTATION_180) },
    { MP_ROM_QSTR(MP_QSTR_ROTATION_270), MP_ROM_INT(DISPLAY_ROTATION_270) },
```

So that they can be compared here as integers?

```c
if (mp_obj_get_int(arg) > DISPLAY_ROTATION_270 ||
    mp_obj_get_int(arg) < DISPLAY_ROTATION_NONE) {
```

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,333 @@
...
+};
+#else
+static const struct device *zephyr_display_devices[1];
+#endif
+
+const mp_obj_type_t zephyr_display_type;
+
+typedef struct _zephyr_display_obj_t {
+    mp_obj_base_t base;
+    uint8_t id;
```
josuah: If wanting to support displays from other sources than `chosen { zephyr,display = &xyz; };` for any reason, then this can be defined as `const struct device *dev;` like done elsewhere:

https://github.com/micropython/micropython/blob/b5fcb33eaa682bb666c839cd4fb301175cc3564f/ports/zephyr/machine_spi.c#L52

https://github.com/micropython/micropython/blob/b5fcb33eaa682bb666c839cd4fb301175cc3564f/ports/zephyr/modzsensor.c#L39

Then the code can be a bit shorter using `self->dev` instead of `zephyr_display_devices[self->id]`.
There might be other reasons to use IDs that I missed though.

josuah (commented):
I will need to revive my MicroPython memory and give it another look...
I hope this helps in the meantime.

ports/zephyr/zephyr_display.c
Gadgetoid: Agreed. Using `MP_QSTR_ROTATION_270` like this might actually work, but isn't very idiomatic of MicroPython bindings.

(Of course now I'm tempted to try.)

Edit: Of course I'm talking nonsense, `mp_obj_get_int(arg)` would raise a `TypeError` when given a string, which of course makes sense. And there's no guarantee that `MP_QSTR_ROTATION_270 == 270` or even that the QSTR integer values are in any particular order.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,333 @@
...
+} zephyr_display_obj_t;
+
+static void zephyr_display_print(const mp_print_t *print, mp_obj_t self_in, mp_print_kind_t kind) {
+    zephyr_display_obj_t *self = self_in;
```
Gadgetoid: Needs `MP_OBJ_TO_PTR` I think!

ports/zephyr/modzephyr.c
VynDragon: Yea, forgot to remove.

ports/zephyr/zephyr_display.c
VynDragon: Well it can do both depending on input type, so ill add the string method.

ports/zephyr/zephyr_display.c
VynDragon: That is supposed to be **DISPLAY_ORIENTATION_ROTATED_270** ...

Gadgetoid: Should be able to expose those constants to MicroPython too, perhaps with the same naming scheme so Zephyr users are at home?

```c
    { MP_ROM_QSTR(MP_QSTR_DISPLAY_ORIENTATION_NORMAL), MP_ROM_INT(DISPLAY_ORIENTATION_NORMAL) },
    { MP_ROM_QSTR(MP_QSTR_DISPLAY_ORIENTATION_ROTATED_90), MP_ROM_INT(DISPLAY_ORIENTATION_ROTATED_90) },
    { MP_ROM_QSTR(MP_QSTR_DISPLAY_ORIENTATION_ROTATED_180), MP_ROM_INT(DISPLAY_ORIENTATION_ROTATED_180) },
    { MP_ROM_QSTR(MP_QSTR_DISPLAY_ORIENTATION_ROTATED_270), MP_ROM_INT(DISPLAY_ORIENTATION_ROTATED_270) },
 ```

File: ports/zephyr/zephyr_display.c [hunk 6]
```diff
@@ -0,0 +1,333 @@
...
+    sy = bufinfo.len / (sx * DISPLAY_BITS_PER_PIXEL(capabilities.current_pixel_format) / 8);
+
+    if (n_args > 5) {
+        sy = mp_obj_get_int(args_in[5]);
+    }
+
+    len = sx * sy * DISPLAY_BITS_PER_PIXEL(capabilities.current_pixel_format) / 8;
+
+    if (len > bufinfo.len) {
+        mp_raise_ValueError("Buffer is shorter than size requires");
```
Gadgetoid: This and other errors should use `MP_ERROR_TEXT` so it doesn't explode the build if/when MicroPython's error text compression is enabled, eg:

```c
mp_raise_ValueError(MP_ERROR_TEXT("Buffer is shorter than size requires"));
```

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,333 @@
...
+                return mp_obj_new_int(0x0);
+            }
+        case PIXEL_FORMAT_L_8:
+            return mp_obj_new_int((r + g + b) / 3);
+        case PIXEL_FORMAT_RGB_565:
+            return mp_obj_new_int((r & 0xe0) | ((g >> 3) & 0x1c) | (b >> 6));
+        default:
+            return mp_obj_new_int(0);
+    }
+    return mp_obj_new_int(0);
```
Gadgetoid: Unreachable return.

Possibly better to return `mp_const_none` in `default:` or even raise an exception so the user isn't left with a valid-looking colour in the exceptional case.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,333 @@
...
+            break;
+        case PIXEL_FORMAT_RGB_565:
+            format = 1; // FRAMEBUF_RGB565
+            break;
+        default:
+            format = -1;
+    }
+    if (format < 0) {
+        mp_raise_msg(&mp_type_RuntimeError, "Display is not of a compatible pixel format");
+        return mp_const_none;
```
Gadgetoid: Unreachable return, there's no need to return after an exception.

I'm not totally sure a `RuntimeError` is the right pattern here, either.  Should probably just return `mp_const_none` and let documentation indicate this means "incompatible." Otherwise it bails user code before they have a chance to (maybe?) set a compatible pixel format.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,333 @@
...
+
+static MP_DEFINE_CONST_FUN_OBJ_1(zephyr_display_framebuf_current_format_obj, zephyr_display_framebuf_current_format);
+
+static mp_obj_t zephyr_display_formats(const mp_obj_t self_in) {
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    struct display_capabilities capabilities;
+
+    display_get_capabilities(zephyr_display_devices[self->id], &capabilities);
+
+    return mp_obj_new_int(capabilities.supported_pixel_formats);
```
Gadgetoid: *Could* maybe unpack this into a tuple so it's more idiomatic of Python? eg:

```python
if display.FORMAT_MONO01 in display.zephyr_formats():
    ...
```

ports/zephyr/zephyr_display.c
Gadgetoid: Just scrolled up here to make this same point! 👍

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/modzephyr.c
VynDragon: done

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,351 @@
...
+    { MP_ROM_QSTR(MP_QSTR_ORIENTATION_180), MP_ROM_INT(DISPLAY_ORIENTATION_ROTATED_180) },
+    { MP_ROM_QSTR(MP_QSTR_ORIENTATION_270), MP_ROM_INT(DISPLAY_ORIENTATION_ROTATED_270) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_RGB_888), MP_ROM_INT(PIXEL_FORMAT_RGB_888) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_MONO01), MP_ROM_INT(PIXEL_FORMAT_MONO01) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_MONO10), MP_ROM_INT(PIXEL_FORMAT_MONO10) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_ARGB_8888), MP_ROM_INT(PIXEL_FORMAT_ARGB_8888) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_RGB_565), MP_ROM_INT(PIXEL_FORMAT_RGB_565) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_BGR_565), MP_ROM_INT(PIXEL_FORMAT_BGR_565) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_L_8), MP_ROM_INT(PIXEL_FORMAT_L_8) },
+    { MP_ROM_QSTR(MP_QSTR_FORMAT_AL_88), MP_ROM_INT(PIXEL_FORMAT_AL_88) },
```
Gadgetoid, line 511: `PIXEL_FORMAT_AL_88` is undefined since it was only added five days ago:

https://github.com/zephyrproject-rtos/zephyr/commit/8c6c761744ca10e0e23bfb962893336127677a07

josuah: Something like this to work around?

```c
    { MP_ROM_QSTR(MP_QSTR_FORMAT_L_8), MP_ROM_INT(PIXEL_FORMAT_L_8) },
#ifdef PIXEL_FORMAT_AL_88 /* will be in Zephyr 4.3 */
    { MP_ROM_QSTR(MP_QSTR_FORMAT_AL_88), MP_ROM_INT(PIXEL_FORMAT_AL_88) },
#endif
```

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,351 @@
...
+    zephyr_display_obj_t *self = MP_OBJ_TO_PTR(self_in);
+    mp_printf(print, "Display(\"%s\")", self->dev->name);
+}
+
+static mp_obj_t zephyr_display_make_new(const mp_obj_type_t *type,
+    size_t n_args, size_t n_kw, const mp_obj_t *args) {
+    mp_arg_check_num(n_args, n_kw, 1, 1, false);
+    const struct device *dev;
+
+    if (mp_obj_get_type(args[0]) == &mp_type_str) {
```
Gadgetoid: `mp_obj_is_type()` is more conventional in these cases.

ports/zephyr/zephyr_display.c
VynDragon: Yea I work off main, applied josuah's solution.

ports/zephyr/zephyr_display.c
VynDragon: done

ports/zephyr/zephyr_display.c
josuah: I forgot to indent the `#ifdef` MicroPython-style as you noticed.

VynDragon:
Also fyi you can just give it the framebuf due to buffer protocol it's really nice:
```
disp = zephyr.Display("ssd1357@0")
fb =framebuf.FrameBuffer(bytearray(64*128*2), 64, 128, framebuf.RGB565)
fb.fill(0xF0F0)
disp.write(fb, 0, 0)
```

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,353 @@
...
+    const struct device *dev;
+
+    if (mp_obj_is_str(args[0])) {
+        dev = zephyr_device_find(args[0]);
+    } else {
+        int id = mp_obj_get_int(args[0]);
+
+        if (id < 0 || id >= DT_ZEPHYR_DISPLAYS_COUNT) {
+            mp_raise_ValueError(MP_ERROR_TEXT("Invalid display ID"));
+            return mp_const_none;
```
josuah: Following @Gadgetoid's [previous comment](https://github.com/micropython/micropython/pull/17939/files#r2286136666), I discovered that `mp_raise_msg()` returns in a different function using MicroPython [magic](https://github.com/micropython/micropython/blob/b5fcb33eaa682bb666c839cd4fb301175cc3564f/py/nlr.h#L30), via [`longjmp()`](https://en.wikipedia.org/wiki/Setjmp.h).

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,353 @@
...
+        }
+        dev = zephyr_display_devices[id];
+    }
+    if (!device_is_ready(dev)) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Display is not ready"));
+        return mp_const_none;
```
josuah: AFAIU, `return` is not needed after `mp_raise_msg()`.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,353 @@
...
+            }
+        case PIXEL_FORMAT_L_8:
+            return mp_obj_new_int((r + g + b) / 3);
+        case PIXEL_FORMAT_RGB_565:
+            return mp_obj_new_int((r & 0xe0) | ((g >> 3) & 0x1c) | (b >> 6));
+        default:
+            break;
+    }
+    mp_raise_ValueError(MP_ERROR_TEXT("Not a framebuf pixel format"));
+    return mp_const_none;
```
josuah: AFAIU, `return` is not needed after `mp_raise_...()`.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,353 @@
...
+    struct display_capabilities capabilities;
+    int ret;
+    size_t len;
+    mp_buffer_info_t bufinfo;
+    int x = mp_obj_get_int(args_in[2]);
+    int y = mp_obj_get_int(args_in[3]);
+    int sx;
+    int sy;
+
+    mp_get_buffer_raise(args_in[1], &bufinfo, MP_BUFFER_READ);
```
josuah: At [line 134](https://github.com/micropython/micropython/pull/17939/files#diff-52d15353d62a3e3d31e86d070c43f96560bb080eb9fc5080a42ed1d85a4a21f7R134) it is possible to have between 2 and 6 arguments.

Maybe you meant to allow `disp.write(fb)` and `disp.write(fb, 0)`?

```suggestion
    int x = 0;
    int y = 0;
    int sx;
    int sy;

    mp_get_buffer_raise(args_in[1], &bufinfo, MP_BUFFER_READ);

    if (n_args > 2) {
        x = mp_obj_get_int(args_in[2]);
    }

    if (n_args > 3) {
        y = mp_obj_get_int(args_in[3]);
    }
```

josuah (commented):
Just a few extra checks...

josuah:
Looking at other display drivers around:

- microypython-lib has only [2 drivers](https://github.com/micropython/micropython-lib/tree/master/micropython/drivers/display) with different APIs
- circuitpython have their circuitpython [`framebufferio`](https://docs.circuitpython.org/en/latest/shared-bindings/framebufferio/index.html) thing tightly bound to their drawing library, not really applicable there.
- LVGL uses a [framebuffer and a custom callback to flush](https://github.com/lvgl/lv_binding_micropython/blob/40eb3411ad4f3c296c371d31fa7cef85d6b4ab36/examples/example3.py#L19-L25), it looks like it would work. It's possible to use it through LVGL Zephyr binding anyway.

I like how it uses the buffer protocol now, it feels very native to MicroPython. :)

ports/zephyr/zephyr_display.c
VynDragon: Yes, that is a mistake, it should be able to take only the buffer argument.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,354 @@
...
+
+    display_get_capabilities(self->dev, &capabilities);
+    sx = capabilities.x_resolution;
+
+    if (n_args > 2) {
+        x = mp_obj_get_int(args_in[2]);
+        y = mp_obj_get_int(args_in[3]);
+    }
```
josuah, line 117: You're right, allowing only `x` without `y` (like I proposed) did not make sense... That could have an extra check like this but maybe that's not needed either (increases ROM size too...).
https://github.com/micropython/micropython/blob/a3f9dec78852caeb56b70f1cf5cd2bd71f36c1ee/ports/nrf/boards/MICROBIT/modules/microbitimage.c#L410-L412

p.s. you meant `if (n_args > 3)` for `args_in[3]`?

VynDragon: no, it will throw a exception at the get_int if there is less than 4 args, eg it can take 2 args (self, buffer), 4 args (self, buffer, x,y) , 5 args (self, buffer, x, y, sx) or 6 args (self, buffer, x, y, sx, sy)

josuah: MicroPython <3

josuah (approved):
My reviews are not official on MP but just to say I can't find any glitch here.
I've learned plenty in the process! Thanks again!

VynDragon:
@dpgeorge Could you check out this one please? It's been complete for a very long time and i've just rebased it over the FS changes.

ports/zephyr/zephyr_display.c
maass-hamburg: `PIXEL_FORMAT_AL_88` is a enum, I don't think ifdef works on enums.

but you can check for the zephyr version instead:
```
#include <zephyr/version.h>

#if ZEPHYR_VERSION_CODE >= ZEPHYR_VERSION(4, 3, 0)
```

josuah: It's zephyr 4.4 now so problem solved I think

VynDragon:
@dpgeorge Could you please review this?

dpgeorge:
Yes, I will review.

I don't know anything about zephyr's display, so will need to learn, and maybe try to test.

Might I suggest in the meantime to add some basic docs about how to use it in `docs/zephyr/quickref.rst`?

VynDragon:
Sure. What board and display combination do you have available? I can provide a configuration overlay.

dpgeorge:
I have many boards (usually use rpi_pico, rpi_pico2, nucleo_wb55rg, frdm_k64f, bbc_microbit_v2) but not many displays.  Probably SSD1306 is the easiest for me to use.

VynDragon:
do you know the specific kind of 'ssd1306' you have? There are some that are CH111x or SH110x and they need a different set of init settings and a different writing method, SSD1315 and some other variants will work as ssd1306 but it's better if they are initialized as the proper model (some power settings); So it's better to set exactly which one when using the zephyr driver for those.

dpgeorge:
I'm not sure!  I don't have it at hand right now.  I definitely have one corresponding to the driver https://github.com/micropython/micropython-lib/blob/master/micropython/drivers/display/ssd1306/ssd1306.py and if you could provide overlays/etc for that, then I should be able to get something working.

VynDragon:
This should be good for rpi_pico with SSD1315 0.96 (black PCB) on gpio 4  and 5:


```dts
/ {
	chosen {
		zephyr,display = &display;
	};

};

&i2c0 {
	status = "okay";
	//clock-frequency = <1000000>;

	display: ssd1306@3c {
		compatible = "solomon,ssd1315";
		reg = <0x3c>;
		width = <128>;
		height = <64>;
		segment-offset = <0>;
		page-offset = <0>;
		display-offset = <0>;
		multiplex-ratio = <63>;
		segment-remap;
		com-invdir;
		prechargep = <0x22>;
	};
};

```

However MPY is not building upstream on the pico so looking at that

VynDragon:
```dts
/ {
	chosen {
		zephyr,display = &display;
	};

};

&i2c0 {
	status = "okay";
	clock-frequency = <I2C_BITRATE_FAST>;

	display: ssd1306@3c {
		compatible = "solomon,ssd1315";
		reg = <0x3c>;
		width = <128>;
		height = <64>;
		segment-offset = <0>;
		page-offset = <0>;
		display-offset = <0>;
		multiplex-ratio = <63>;
		segment-remap;
		com-invdir;
		prechargep = <0x22>;
	};
};
```
```python
import zephyr, framebuf, sys, time

dispdev = zephyr.Display("ssd1306@3c")

print(dispdev)

display = framebuf.FrameBuffer(bytearray(int(64*128/8)), 128, 64, framebuf.MVLSB)

dispdev.blanking_off()

display.text('Hello World', 0, 0, 1)
display.text('----------------', 0, 8, 1)
display.text('  From Python', 0, 16, 1)
display.text('   On Zephyr', 0, 24, 1)
display.text('Running on', 18, 32, 1)
name = ""
for c in sys.implementation._machine[7:]:
	if c != ' ':
		name = name + c
	else:
		break
display.text('{}'.format(name), 64 - int(len(name) * 8 / 2), 40, 1)
dispdev.write(display, 0, 0)

time.sleep(1)

dispdev.blanking_on()
time.sleep(0.25)
dispdev.blanking_off()
time.sleep(0.25)

del display

dd = dispdev.as_framebuf()
print(dd)

dd.fill(0)
dd.show()

caps = dispdev.capabilities();

print(caps)

for i in range(1, caps[0]):
	dd.vline(i, 0, caps[1], 0xff)
	dd.vline(i-1, 0, caps[1], 0)
	dd.show()

dd.fill(0)
dd.show()

for i in range(1, caps[1]):
	dd.hline(0, i, caps[0], 0xff)
	dd.hline(0, i-1, caps[0], 0)
	dd.show()

dd.fill(0)
dd.show()

print("Done")
```
```
west build -p -c -b rpi_pico micropython/ports/zephyr/ -DEXTRA_DTC_OVERLAY_FILE=app.overlay -DCONFIG_DISPLAY=y
```

VynDragon:
Update with memory checks:
```python
import zephyr, framebuf, sys, time, gc

def free_mem_is():
	mem_free1 = gc.mem_free()
	print("Free Memory:{}".format(mem_free1))
	gc.collect()
	mem_free2 = gc.mem_free()
	print("Collected {}, Free Memory:{}".format((mem_free2 - mem_free1), mem_free2))

dispdev = zephyr.Display("ssd1306@3c")

print(dispdev)
free_mem_is()

display = framebuf.FrameBuffer(bytearray(int(64*128/8)), 128, 64, framebuf.MVLSB)

dispdev.blanking_off()

display.text('Hello World', 0, 0, 1)
display.text('----------------', 0, 8, 1)
display.text('  From Python', 0, 16, 1)
display.text('   On Zephyr', 0, 24, 1)
display.text('Running on', 18, 32, 1)
name = ""
for c in sys.implementation._machine[7:]:
	if c != ' ':
		name = name + c
	else:
		break
display.text('{}'.format(name), 64 - int(len(name) * 8 / 2), 40, 1)
dispdev.write(display, 0, 0)

time.sleep(1)

dispdev.blanking_on()
time.sleep(0.25)
dispdev.blanking_off()
time.sleep(0.25)

free_mem_is()

del display

free_mem_is()

dd = dispdev.as_framebuf()
print(dd)

dd.fill(0)
dd.show()

caps = dispdev.capabilities();

print(caps)

free_mem_is()

for i in range(1, caps[0]):
	dd.vline(i, 0, caps[1], 0xff)
	dd.vline(i-1, 0, caps[1], 0)
	dd.show()

dd.fill(0)
dd.show()

for i in range(1, caps[1]):
	dd.hline(0, i, caps[0], 0xff)
	dd.hline(0, i-1, caps[0], 0)
	dd.show()

free_mem_is()

dd.fill(0)
dd.show()
free_mem_is()
del dd
free_mem_is()
print("Done")

```

VynDragon:
@dpgeorge Docs added

dpgeorge:
Thanks for the instructions on how to test.  I managed to get it all working on `rpi_pico` with the SSD1306 display, running the example Python code above.

Although this PR with the above overlay did not compile with Zephyr 4.4.0, it did after I updated to Zephyr main.  I guess this PR would compile with Zephyr 4.4.0 if the display was configured correctly?

As for the API, it's looking pretty good.  Even though the `.write()` and `.as_framebuf()` functionality somewhat overlap, I can see how they are both useful.

docs/library/zephyr.Display.rst
```diff
@@ -0,0 +1,73 @@
...
+.. class:: Display(id)
+
+   Gets an object for accessing a Display identified by ``id``.
+
+   ``id`` can be an integer (``0``, ``1``...) or a string (``"ssd1306@3c"``) identifying a display node by its position or by its node identifiers.
+
+Methods
+-------
+
+.. method:: Display.write(buf[, x, y[, size_x[, size_y]]])
```
dpgeorge: Maybe call this method `blit()`?

"write" kind of implies stream semantics, whereas "blit" is more meaningful and matches `FrameBuffer.blit`.

docs/library/zephyr.Display.rst
```diff
@@ -0,0 +1,73 @@
...
+
+            Write a buffer-protocol object in the Display's Pixel Format to the display.
+
+            Optionally x and y position, x size, and y size can be specified.
+
+.. method:: Display.rgb(r, g, b)
+
+            Convert a RGB color to the Display's Pixel Format.
+
+.. method:: Display.framebuf_current_format()
```
dpgeorge: Would it make sense to merge this method into `capabilities()`, by appending it at the end of the tuple returned by that method?  I'm just thinking of ways to simplify the API.

VynDragon:
> Although this PR with the above overlay did not compile with Zephyr 4.4.0, it did after I updated to Zephyr main. I guess this PR would compile with Zephyr 4.4.0 if the display was configured correctly?

What was the issue? ssd1306 didn't exist but ssd1306fb did? Or something else?

And yes as_framebuf is explicitly a convenience feature, could be made optional if it has some cost

docs/library/zephyr.Display.rst
```diff
@@ -0,0 +1,73 @@
...
+
+            Retrieve a tuple describing the display in the format:
+
+            ``(X Size, Y Size, Supported PFs, Current PF, Current Orientation, Misc Characteristics)``
+
+.. method:: Display.format([format])
+
+            Get and set the Pixel Format of the Display.
+
+.. method:: Display.blanking_on()
```
dpgeorge: How about `Display.blanking(value)` and pass in a bool?  Then you do `disp.blanking(False)` etc.

That'll reduce the number of methods by 1.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,555 @@
...
+    mp_obj_t base_tuple;
+    mp_obj_t args[4];
+    mp_obj_t locals;
+    mp_obj_t fb_type;
+    uint8_t *buf;
+    size_t buf_size;
+    mp_obj_t instance;
+    mp_obj_t bound_show;
+
+    base[0] = mp_load_attr(mp_load_global(MP_QSTR_framebuf), MP_QSTR_FrameBuffer);
```
dpgeorge: `import framebuf` is needed for this to work.

Maybe just access the global `mp_module_framebuf` variable instead?

docs/library/zephyr.Display.rst
VynDragon: I don't think this is good semantically, blit means writing into a framebuffer usually, and it's used in framebuf for that meaning. The practical operation done here as well is effectively writing in the sense of SPI or I2C rather than a stream

ports/zephyr/zephyr_display.c
VynDragon: Is it better for memory usage?

docs/library/zephyr.Display.rst
VynDragon: The other naming I have considered though is 'show' (from the popular ui framework using framebuf), but went with the zephyr naming in the api, and for show in the framebuf overlay

dpgeorge:
> What was the issue? ssd1306 didn't exist but ssd1306fb did? Or something else?

```
...
[442/447] Linking C executable zephyr/zephyr_pre0.elf; L...command for additional byproducts on target: zephyr_pre
FAILED: [code=1] zephyr/zephyr_pre0.elf zephyr/zephyr_pre0.map micropython/ports/zephyr/build/zephyr/zephyr_pre0.map                     
: && zephyrproject/zephyr-sdk-1.0.1/gnu/arm-zephyr-eabi/bin/arm-zephyr-eabi-gcc  -gdwarf-4 -Os zephyr/CMakeFiles/zephyr_pre0.dir/misc/empty_file.c.obj -o zephyr/zephyr_pre0.elf  zephyr/CMakeFiles/offsets.dir/./arch/arm/core/offsets/offsets.c.obj  -T  zephyr/linker_zephyr_pre0.cmd  -Wl,-Map,micropython/ports/zephyr/build/zephyr/zephyr_pre0.map  -Wl,--whole-archive  app/libapp.a  zephyr/libzephyr.a  zephyr/arch/common/libarch__common.a  zephyr/arch/arch/arm/core/libarch__arm__core.a  zephyr/arch/arch/arm/core/cortex_m/libarch__arm__core__cortex_m.a  zephyr/lib/libc/picolibc/liblib__libc__picolibc.a  zephyr/lib/libc/common/liblib__libc__common.a  zephyr/lib/posix/c_lib_ext/liblib__posix__c_lib_ext.a  zephyr/lib/net_buf/liblib__net_buf.a  zephyr/subsys/fs/libsubsys__fs.a  zephyr/subsys/random/libsubsys__random.a  zephyr/subsys/usb/device_next/libsubsys__usb__device_next.a  zephyr/drivers/usb/udc/libdrivers__usb__udc.a  zephyr/drivers/usb/common/buf/libdrivers__usb__common__buf.a  zephyr/drivers/clock_control/libdrivers__clock_control.a  zephyr/drivers/console/libdrivers__console.a  zephyr/drivers/flash/libdrivers__flash.a  zephyr/drivers/gpio/libdrivers__gpio.a  zephyr/drivers/hwinfo/libdrivers__hwinfo.a  zephyr/drivers/i2c/libdrivers__i2c.a  zephyr/drivers/pinctrl/libdrivers__pinctrl.a  zephyr/drivers/reset/libdrivers__reset.a  zephyr/drivers/serial/libdrivers__serial.a  zephyr/drivers/spi/libdrivers__spi.a  zephyr/drivers/timer/libdrivers__timer.a  zephyr/drivers/watchdog/libdrivers__watchdog.a  modules/hal_rpi_pico/libmodules__hal_rpi_pico.a  modules/littlefs/libmodules__littlefs.a  -Wl,--no-whole-archive  zephyr/kernel/libkernel.a  -Lmicropython/ports/zephyr/build/zephyr  libmicropython.a  zephyr/kernel/libkernel.a  zephyr/arch/common/libisr_tables.a  -fuse-ld=bfd  -mcpu=cortex-m0plus  -mthumb  -mabi=aapcs  -mfp16-format=ieee  -mtp=soft  -Wl,--gc-sections  -Wl,--build-id=none  -Wl,--sort-common=descending  -Wl,--sort-section=alignment  -Wl,-u,_OffsetAbsSyms  -Wl,-u,_ConfigAbsSyms  -nostdlib  -static  -znoexecstack  -Wl,-X  -Wl,-N  -Wl,--orphan-handling=warn  -Wl,-no-pie  -Wl,--undefined=_sw_isr_table  -Wl,--undefined=_irq_vector_table  -specs=picolibc.specs  -DPICOLIBC_LONG_LONG_PRINTF_SCANF -L"zephyrproject/zephyr-sdk-1.0.1/gnu/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/14.3.0/thumb/v6-m/nofp/space" -lc -lgcc && cd micropython/ports/zephyr/build/zephyr && /usr/bin/cmake -E true                                                                                                       
zephyrproject/zephyr-sdk-1.0.1/gnu/arm-zephyr-eabi/bin/../lib/gcc/arm-zephyr-eabi/14.3.0/../../../../arm-zephyr-eabi/bin/ld.bfd: libmicropython.a(zephyr_display.c.obj): in function `zephyr_display_make_new':                    
micropython/ports/zephyr/zephyr_display.c:93:(.text.zephyr_display_make_new+0x6c): undefined reference to `__device_dts_ord_79'                     
ninja: build stopped: subcommand failed.
```

docs/library/zephyr.Display.rst
dpgeorge: OK, fair enough.  Matching zephyr naming is a good reason to keep it `write`.

ports/zephyr/zephyr_display.c
dpgeorge: At the moment doing `disp.as_framebuf()` does not work, it raises an exception about unknown name "framebuf" (if the code didn't already import framebuf).

Replacing `mp_load_global(MP_QSTR_framebuf)` with `MP_OBJ_FROM_PTR(&mp_module_framebuf)` will make it work without importing framebuf, and will be faster, and probably a tiny bit smaller in codesize.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,555 @@
...
+    int y = 0;
+    int sx;
+    int sy;
+
+    mp_get_buffer_raise(args_in[1], &bufinfo, MP_BUFFER_READ);
+
+    display_get_capabilities(self->dev, &capabilities);
+    sx = capabilities.x_resolution;
+
+    if (n_args > 2) {
```
dpgeorge, line 111: I think this should be `n_args > 3`, otherwise `args_in[3]` is invalid data.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,555 @@
...
+        x = mp_obj_get_int(args_in[2]);
+        y = mp_obj_get_int(args_in[3]);
+    }
+
+    if (n_args > 4) {
+        sx = mp_obj_get_int(args_in[4]);
+    }
+    sy = bufinfo.len / (sx * DISPLAY_BITS_PER_PIXEL(capabilities.current_pixel_format) / 8);
```
dpgeorge: Don't need to compute this line if `n_args > 5`, so it could go in an else-clause of the if-statement below.

ports/zephyr/zephyr_display.c
```diff
@@ -0,0 +1,555 @@
...
+
+    if (len > bufinfo.len) {
+        mp_raise_ValueError(MP_ERROR_TEXT("Buffer is shorter than size requires"));
+    }
+
+    const struct display_buffer_descriptor desc = {
+        .buf_size = len,
+        .width = sx,
+        .height = sy,
+        .pitch = sx,
```
dpgeorge, line 139: Is it worth allowing an optional argument to specify pitch/stride?  Ie:
```py
Display.write(buf[, x, y[, size_x[, size_y[, stride]]]])
```

VynDragon: No, I don't know of a single driver that doesn't reject an input buffer where pitch != width.

docs/library/zephyr.Display.rst
VynDragon: done

docs/library/zephyr.Display.rst
VynDragon: done

ports/zephyr/zephyr_display.c
VynDragon: okay, done

ports/zephyr/zephyr_display.c
VynDragon: fixed

ports/zephyr/zephyr_display.c
VynDragon: done

VynDragon:
> error

Probably what i mentioned yea, cant tell without seeing the full log or looking at the generated file in the build folder.

VynDragon:
I also redid test with a color display, ssd1331 this time, to check i didnt flip the byte order around in zephyr_display_rgb

dpgeorge (approved):
Thanks for updating.  Good to see the API a little smaller but still the same functionality.

PR #17945, opened 2025-08-17
STM32: Enable I2CTarget for STM32L4.

chrismas9:
STM32L4 uses the same I2C controller as STM32WB.
Change `defined(STM32WB)` to `defined(STM32WB) || defined(STM32L4) `in:
stm32/i2cslave.c
stm32/i2cslave.h
stm32/mpconfigboard_common.h

### Testing

Tested with custom STM32L496 board. Created I2C and I2CTarget with
wire link betweenports and read data from I2CTarget buffer.

### DIFF
--- a/ports/stm32/i2c.c
+++ b/ports/stm32/i2c.c
@@ -279,10 +279,6 @@ int i2c_write(i2c_t *i2c, const uint8_t *src, size_t len, size_t next_len) {
 #elif defined(STM32L4)
 #define APB1ENR            APB1ENR1
 #define RCC_APB1ENR_I2C1EN RCC_APB1ENR1_I2C1EN
-#if defined(STM32L432xx)
-// Not a real peripheral, only needed for i2c_id calculation in i2c_init.
-#define I2C2_BASE          (APB1PERIPH_BASE + 0x5800UL)
-#endif
 #endif
 
 static uint16_t i2c_timeout_ms[MICROPY_HW_MAX_I2C];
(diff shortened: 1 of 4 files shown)


dpgeorge (approved):
Very good!  Nice to hear the L4 works with I2CTarget.

dpgeorge:
Actually, testing this PR on NUCLEO_L432KC it doesn't build.  I added a commit to fix that.  See what you think.

chrismas9:
Your change works. I have hardware tested NUCLEO_L432KC and NUCLEO_L452RE.
I have built NUCLEO_L432KC  without I2CTarget and it still compiles.
Since `I2C2_BASE` is required by i2c would it be better to keep the definition in `i2c.c` rather then `i2cslave.h`?

dpgeorge:
> Your change works. I have hardware tested NUCLEO_L432KC and NUCLEO_L452RE.
> I have built NUCLEO_L432KC without I2CTarget and it still compiles.

Great, thanks for testing.



> Since `I2C2_BASE` is required by i2c would it be better to keep the definition in `i2c.c` rather then `i2cslave.h`?

Yes.  It would be better to have `I2C2_BASE` defined in `i2c.h` and shared with `i2cslave.[ch]`, but that leads to problems with building mboot, which needs to include `i2cslave.h`, and that in turn can't include `i2c.h`.

All that could do with a clean up, eg make a shared inline function that computes `((uint32_t)i2c - I2C1_BASE) / (I2C2_BASE - I2C1_BASE)`.  But let's do that in a follow-up PR.

PR #17946, opened 2025-08-17
py/asmthumb: Fix T3 encoding of conditional branches.

### LINKED ISSUE #17940: Large Viper code causing hang/crash on RP2350 but OK on RP2040

MicroPython v1.26.0 on 2025-08-09; Raspberry Pi Pico2 with RP2350

### MicroPython version

MicroPython v1.26.0 on 2025-08-09; Raspberry Pi Pico2 with RP2350
Large viper functions have been improved in v26 and shown to work on RP2040. The same code on a RP2350 however causes the pico VM to hang or crash.
In the code below a DEPTH of 500 works but 600 crashes. 

### Reproduction

```
DEPTH = const(600)

src = "@micropython.viper\ndef big_jump() -> int:\n"
src += "    x = 0\n"
src += "    while x < 10:\n"
for i in range(DEPTH):
    src += f"        x += 1\n"
src += "    return x"

#print(src)
exec(src)
print(big_jump())

### Expected behaviour

_No response_

### Observed behaviour

Observed behaviour is that the RP2350 is non responsive and requires a hard reset.

### Additional Information

### Code of Conduct

Yes, I agree

agatti:
### Summary

This PR fixes the encoding of conditional branch opcodes emitted for ARMv7-M targets, when the emitter decides to use the T3 encoding for said operation.

Fields J1 and J2 are now present in the generated opcode word, along with correcting some minor issues in bitmasks and shifts computation.

This fixes #17940.

### Testing

The full MicroPython test suite was run on a RP2350 (as in, regular, `--via-mpy`, and `--via-mpy --emit-native`).

The `micropython/viper_large_jump.py` test was run on ESP8266, ESP32, and ESP32C3 in the same three ways as the RP2350 just to make sure the newly added test won't fail.

<hr>

Incidentally, when testing this I found out the `micropython/viper_ptrX_store_boundary.py` tests aren't aware of `tests/run-tests.py` ability to detect large tests.  I guess I'd have to rewrite those three tests using unittest and bake in the expected results in the test files themselves.


dpgeorge (approved):
Thanks for the fix!

I see that this is my fault, and in fact the offending code was in the very first commit of MicroPython 429d71943d6b94c7dc3c40a39ff1a09742c77dc2 !  I really did mess up the calculation to encode the byte-offset, although at least the lower bits were correct, and I guess that's why it's gone unnoticed until now.

Also, the comment (now removed) was right, the macros needed testing :joy:

dpgeorge:
> Incidentally, when testing this I found out the `micropython/viper_ptrX_store_boundary.py` tests aren't aware of `tests/run-tests.py` ability to detect large tests. I guess I'd have to rewrite those three tests using unittest and bake in the expected results in the test files themselves.

This SKIP-TOO-LARGE option is a very new feature.  Feel free to adjust those tests to properly skip when out-of-memory (although I don't think they need to be unittest tests, that would make them use even more memory).

agatti:
> This SKIP-TOO-LARGE option is a very new feature. Feel free to adjust those tests to properly skip when out-of-memory (although I don't think they need to be unittest tests, that would make them use even more memory).

Right now those tests generate compiled blocks on the fly, which makes them fail in the middle of execution on low-memory platforms (ie. ESP8266 without a prologue file that maps some flash into IRAM).  `unittest` has the ability to either split a single test file into separate cases that may or may not be skipped on their own, or even provide subtests (in this case the address width in bits) for a single fixture.

I can look into making the test generate all compiled blocks upfront so it would fail right away and not require `unittest` for a clean run on the 8266, but it may make those tests more complex than they are now. Still, thanks for the remark, I'll see what I can do with this!

dpgeorge:
My preference is always to keep things simple if possible (but at the same time be pragmatic).

And it would actually be good to improve these viper ptr tests, because I'm seeing them fail on small boards like ADAFRUIT_ITSYBITSY_M0_EXPRESS, eg see this Octoprobe output: https://reports.octoprobe.org/github_selfhosted_testrun_255/RUN-TESTS_STANDARD,a@5f2a-ADA_ITSYBITSY_M0/micropython_viper_ptr8_store_boundary.py.out

PR #17948, opened 2025-08-17
Fix NULL pointer dereference in mp_obj_get_type #17944

### LINKED ISSUE #17941: Segmentation fault in array sorting with GC interaction

unix port

### MicroPython version

v1.26.0-46-g1588c455c on 2025-08-17

### Reproduction

1. Create a comparison class that modifies the array being sorted
2. Call gc.collect() within the comparison function
3. Attempt to sort the array
```python
import gc
class check:
    def __init__(self, arr1, arr2):
        self.arr1 = arr1
        self.arr2 = arr2
    
    def __lt__(self, other):
        # Clear one array, modify another
        self.arr1.clear()
        self.arr2.extend([777] * 5)
        gc.collect()
        self.arr2.clear()
        return True

d1 = []
d2 = []
contaminators = [check(d1, d2) for _ in range(3)]
d1.extend(contaminators)
d2.extend(contaminators)
d1.sort()
```

### Expected behaviour

The code should either:
1. Raise a Python exception (e.g., RuntimeError, ValueError)
2. Handle the edge case gracefully without crashing

### Observed behaviour

MicroPython crashes with a segmentation fault when array.clear() and gc.collect() are called within comparison functions during array sorting operations. This appears to be caused by a NULL pointer dereference in py/obj.c at line 61.
```shell
AddressSanitizer:DEADLYSIGNAL
=================================================================
==1758094==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x64c7a2bd4b4b bp 0x000000000000 sp 0x7fffd2494c90 T0)
==1758094==The signal is caused by a READ memory access.
==1758094==Hint: address points to the zero page.
    #0 0x64c7a2bd4b4b in mp_obj_get_type ../../py/obj.c:61
    #1 0x64c7a2bccdac in mp_binary_op ../../py/runtime.c:640
    #2 0x64c7a2bf012d in mp_quicksort ../../py/objlist.c:288
    #3 0x64c7a2bf0405 in mp_obj_list_sort ../../py/objlist.c:330
    #4 0x64c7a2be8192 in fun_builtin_var_call ../../py/objfun.c:118
    #5 0x64c7a2bc6984 in mp_call_function_n_kw ../../py/runtime.c:727
    #6 0x64c7a2bc7547 in mp_call_method_n_kw ../../py/runtime.c:743
    #7 0x64c7a2c2b5cd in mp_execute_bytecode ../../py/vm.c:1069
    #8 0x64c7a2be85a9 in fun_bc_call ../../py/objfun.c:295
    #9 0x64c7a2bc6984 in mp_call_function_n_kw ../../py/runtime.c:727
    #10 0x64c7a2bcadb3 in mp_call_function_0 ../../py/runtime.c:701
    #11 0x64c7a2d4adc6 in execute_from_lexer /home/kai/project/hypothesmith/micropython/ports/unix/main.c:162
    #12 0x64c7a2d4aedc in do_file /home/kai/project/hypothesmith/micropython/ports/unix/main.c:311
    #13 0x64c7a2d4c825 in main_ /home/kai/project/hypothesmith/micropython/ports/unix/main.c:728
    #14 0x64c7a2d4ce63 in main /home/kai/project/hypothesmith/micropython/ports/unix/main.c:494
    #15 0x7b69f7c2a3b7 in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58
    #16 0x7b69f7c2a47a in __libc_start_main_impl ../csu/libc-start.c:360
    #17 0x64c7a2b69734 in _start (micropython/ports/unix/build-coverage/micropython+0x1a8734) (BuildId: bb66085681e6f5a07b78002d60ef2779d5eae802)
```

The AddressSanitizer output shows a NULL pointer dereference at address 0x000000000000. The call stack reveals:

`mp_quicksort (objlist.c:288)` - Sorting algorithm in progress
`mp_binary_op (runtime.c:640)` - Calls comparison operation
`mp_obj_get_type (obj.c:61)` - CRASH HERE - Dereferencing NULL pointer

### Additional Information

### Code of Conduct

Yes, I agree

MaksimFeng:
### Summary

py/obj:Against null during list.sort,add testcase.
If a list's comparison function calls array.clear() and gc.collect() during sort, an element slot can
become MP_OBJ_NULL. Guard mp_obj_get_type() to avoid a NULL dereference.

Closes: #17941

### Testing

Testing on unix port.

### DIFF
--- a/py/objlist.c
+++ b/py/objlist.c
@@ -41,6 +41,11 @@ static mp_obj_t list_pop(size_t n_args, const mp_obj_t *args);
 
 /******************************************************************************/
 /* list                                                                       */
+static inline void list_sort_mutation_check(mp_obj_list_t *self, mp_obj_t *orig_items, size_t orig_len) {
+    if (self->items != orig_items || self->len != orig_len) {
+        mp_raise_ValueError(MP_ERROR_TEXT("List modified during sort"));
+    }
+}
 
 static void list_print(const mp_print_t *print, mp_obj_t o_in, mp_print_kind_t kind) {
     mp_obj_list_t *o = MP_OBJ_TO_PTR(o_in);
@@ -277,7 +282,8 @@ static mp_obj_t list_pop(size_t n_args, const mp_obj_t *args) {
     return ret;
 }
 
-static void mp_quicksort(mp_obj_t *head, mp_obj_t *tail, mp_obj_t key_fn, mp_obj_t binop_less_result) {
+static void mp_quicksort(mp_obj_list_t *self, mp_obj_t *head, mp_obj_t *tail, mp_obj_t key_fn, mp_obj_t binop_less_result, mp_obj_t *orig_items, size_t orig_len) {
+
     mp_cstack_check();
     while (head < tail) {
         mp_obj_t *h = head - 1;
@@ -285,12 +291,15 @@ static void mp_quicksort(mp_obj_t *head, mp_obj_t *tail, mp_obj_t key_fn, mp_obj
         mp_obj_t v = key_fn == MP_OBJ_NULL ? tail[0] : mp_call_function_1(key_fn, tail[0]); // get pivot using key_fn
         for (;;) {
             do {++h;
+                list_sort_mutation_check(self, orig_items, orig_len);
             } while (h < t && mp_binary_op(MP_BINARY_OP_LESS, key_fn == MP_OBJ_NULL ? h[0] : mp_call_function_1(key_fn, h[0]), v) == binop_less_result);
             do {--t;
+                list_sort_mutation_check(self, orig_items, orig_len);
             } while (h < t && mp_binary_op(MP_BINARY_OP_LESS, v, key_fn == MP_OBJ_NULL ? t[0] : mp_call_function_1(key_fn, t[0])) == binop_less_result);
             if (h >= t) {
                 break;
             }
+            list_sort_mutation_check(self, orig_items, orig_len);
             mp_obj_t x = h[0];
             h[0] = t[0];
             t[0] = x;
@@ -300,10 +309,10 @@ static void mp_quicksort(mp_obj_t *head, mp_obj_t *tail, mp_obj_t key_fn, mp_obj
         tail[0] = x;
         // do the smaller recursive call first, to keep stack within O(log(N))
         if (t - head < tail - h - 1) {
-            mp_quicksort(head, t, key_fn, binop_less_result);
+            mp_quicksort(self, head, t, key_fn, binop_less_result, orig_items, orig_len);
             head = h + 1;
         } else {
-            mp_quicksort(h + 1, tail, key_fn, binop_less_result);
+            mp_quicksort(self, h + 1, tail, key_fn, binop_less_result, orig_items, orig_len);
             tail = t;
         }
     }
@@ -327,9 +336,13 @@ mp_obj_t mp_obj_list_sort(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_
     mp_obj_list_t *self = MP_OBJ_TO_PTR(pos_args[0]);
 
     if (self->len > 1) {
-        mp_quicksort(self->items, self->items + self->len - 1,
+        mp_obj_t *orig_items = self->items;
+        size_t orig_len = self->len;
+        mp_quicksort(self,
+            self->items, self->items + self->len - 1,
             args.key.u_obj == mp_const_none ? MP_OBJ_NULL : args.key.u_obj,
-            args.reverse.u_bool ? mp_const_false : mp_const_true);
+            args.reverse.u_bool ? mp_const_false : mp_const_true,
+            orig_items, orig_len);
     }
 
     return mp_const_none;
--- a/tests/basics/list_sort_sideeffect.py
+++ b/tests/basics/list_sort_sideeffect.py
@@ -1,0 +1,29 @@
+try:
+    import gc
+except ImportError:
+    print("SKIP")
+    raise SystemExit
+
+
+class Evil:
+    n_calls = 0
+
+    def __init__(self, arr, val):
+        self.arr = arr
+        self.val = val
+
+    def __lt__(self, other):
+        Evil.n_calls += 1
+        if Evil.n_calls == 16:
+            self.arr.clear()
+            gc.collect()
+        return self.val < other.val
+
+
+a = []
+a.extend(Evil(a, i) for i in range(32))
+
+try:
+    a.sort()
+except ValueError:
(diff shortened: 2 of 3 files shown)


py/obj.c
```diff
@@ -57,6 +57,9 @@ const mp_obj_type_t *MICROPY_WRAP_MP_OBJ_GET_TYPE(mp_obj_get_type)(mp_const_obj_
     #if MICROPY_OBJ_IMMEDIATE_OBJS && MICROPY_OBJ_REPR == MICROPY_OBJ_REPR_A
 
     if (mp_obj_is_obj(o_in)) {
+        if (o_in == MP_OBJ_NULL) {
```
agatti: I believe this should also be applied to the other instance of `mp_obj_is_obj` later in the function, and to the final `->type` dereference at the end of the function.

If I understood this correctly, this only fixes the issue for `MICROPY_OBJ_REPR_A`, doesn't it?.

MaksimFeng: Thanks for the review — you’re right on both points.
My original guard lived only in the `MICROPY_OBJ_REPR_A`/`mp_obj_is_obj()` branch, so it would miss other branches and other representations. I’ve moved the `MP_OBJ_NULL` check to the top of `mp_obj_get_type()`, before any branching.
To address the crash at its source without changing list layout, I implemented a mutation check in `objlist.c`. Pass this check to `mp_quicksort` and check after any call that can run Python code (key function, comparator) and before swaps. If the pointer or length changed, raise `ValueError("list modified during sort")`.

py/objlist.c
```diff
@@ -41,6 +41,11 @@ static mp_obj_t list_pop(size_t n_args, const mp_obj_t *args);
 
 /******************************************************************************/
 /* list                                                                       */
+static inline void list_sort_mutation_check(mp_obj_list_t *self, mp_obj_t *orig_items, size_t orig_len){
+    if (self->items != orig_items || self->len != orig_len){
+        mp_raise_ValueError(MP_ERROR_TEXT("list modified during sort."));
```
agatti: Error messages shouldn't have a trailing period (I've got bitten by this as well in the past).

py/objlist.c
```diff
@@ -277,33 +282,42 @@ static mp_obj_t list_pop(size_t n_args, const mp_obj_t *args) {
...
 
-static void mp_quicksort(mp_obj_t *head, mp_obj_t *tail, mp_obj_t key_fn, mp_obj_t binop_less_result) {
+static void mp_quicksort(mp_obj_list_t *self, mp_obj_t *head, mp_obj_t *tail, mp_obj_t key_fn, mp_obj_t binop_less_result, mp_obj_t *orig_items, size_t orig_len){
+
     mp_cstack_check();
     while (head < tail) {
         mp_obj_t *h = head - 1;
         mp_obj_t *t = tail;
         mp_obj_t v = key_fn == MP_OBJ_NULL ? tail[0] : mp_call_function_1(key_fn, tail[0]); // get pivot using key_fn
+        list_sort_mutation_check(self, orig_items, orig_len);
```
agatti: This can probably be removed, as I think the execution flow should always pass through the call to `list_sort_mutation_check` at line 296.

py/objlist.c
```diff
@@ -277,33 +282,42 @@ static mp_obj_t list_pop(size_t n_args, const mp_obj_t *args) {
...
             } while (h < t && mp_binary_op(MP_BINARY_OP_LESS, v, key_fn == MP_OBJ_NULL ? t[0] : mp_call_function_1(key_fn, t[0])) == binop_less_result);
             if (h >= t) {
                 break;
             }
+            list_sort_mutation_check(self, orig_items, orig_len);
             mp_obj_t x = h[0];
             h[0] = t[0];
             t[0] = x;
         }
+        list_sort_mutation_check(self, orig_items, orig_len);
```
agatti: This can probably be removed too, at this point the list should have been checked by the call to `list_sort_mutation_check` on line 306, I think.

py/obj.c
agatti: I see.  Considering that your change is affecting all ports you may want to move the test from `tests/ports/unix` to `tests/basics` instead.  If you don't have a board for each port involved to test this just ask, folks here can help you out :)

Also, there's a non-trivial size increase on all ports caused by this change, see the comment for `py/objlist.c` below to see whether things can be made shorter.

py/objlist.c
```diff
@@ -277,33 +282,42 @@ static mp_obj_t list_pop(size_t n_args, const mp_obj_t *args) {
...
         mp_obj_t v = key_fn == MP_OBJ_NULL ? tail[0] : mp_call_function_1(key_fn, tail[0]); // get pivot using key_fn
+        list_sort_mutation_check(self, orig_items, orig_len);
         for (;;) {
-            do {++h;
+            do {
+                ++h;
+                list_sort_mutation_check(self, orig_items, orig_len);
             } while (h < t && mp_binary_op(MP_BINARY_OP_LESS, key_fn == MP_OBJ_NULL ? h[0] : mp_call_function_1(key_fn, h[0]), v) == binop_less_result);
-            do {--t;
+            list_sort_mutation_check(self, orig_items, orig_len);
```
agatti: This can probably be removed too, as I think execution should pass through line 301 anyway.

py/obj.c
dpgeorge: This check should not be needed, there should be no way for MP_OBJ_NULL to get here (at least, if the mutate-while-sorting issue is solved).

dpgeorge:
Thanks for the contribution.

For this to be considered for merging, it needs to be as minimal as possible in terms of code size increase (see above "Code size report", which will change as you push to this PR).

py/obj.c
MaksimFeng: Thank you so much for reviewing the code. There're indeed some duplicate checks and I've removed them. Please help me test it on different boards. Currently, I don't have an extra board with me.

agatti:
@MaksimFeng For the `unix/minimal` test error you see, said configuration doesn't have the garbage collector compiled in - apologies for missing that during the first review pass.

To work around that, relevant module inclusions that may fail can be wrapped like this:

```python
try:
    import gc
except ImportError:
    print("SKIP")
    raise SystemExit
```

Existing documentation on writing tests can be a bit scarce, but there's #17301 that aims to flesh that out.  You may want to scan those changes as they may clarify a few things.

*Edit:* the same thing makes the Zephyr test fail too?  I thought that had the garbage collector enabled, go figure.

MaksimFeng:
@agatti, thank you for the reminder and the supplementary document. It was very helpful. It appears the Zephyr test failure was caused by the gc import....

PR #17949, opened 2025-08-18
tests: change all CLI test runners to use the new `--test-instance` argument scheme

dpgeorge:
### Summary

Back in PR #16111 (see commit 8978102f3595ae321484a6be44c1dcf25c8909a9) the `run-tests.py` script was changed to use an improved way of selecting the test instance, eg:
```py
$ ./run-tests.py -t a0
```
that would run on `/dev/ttyACM0`.  This has been a very nice improvement, makes it easier to specify the target.

This PR updates all other test runners to use the same scheme:
- `run-multitests.py` -- previously used `-i`, eg `-i pyb:a0`
- `run-natmodtests.py` -- previously used `-d` and `-p`, eg `-p -d /dev/ttyACM0'
- `run-perfbench.py` -- previously used `-d` and `-p`

The `run-internalbench.py` runner was already updated in 88cb6bc818546e7fe8b862c784e1e4a6b8a12b8a

Now, everything uses the same scheme.

### Testing

Tested locally running on the unix port and some boards attached to serial ports.

CI will also test this.

Octoprobe will need updating.

Users will need to learn the new CLI options.  And some existing ones had to be renamed because they clashed with `-t`.

We could leave things as they are, but it's pretty awkward how the test runners all use different CLI options to select the test target.

There's still quite a bit of refactoring that could be done, to consolidate more code between the test runners.  But in the interest of making small, manageable steps, I've only done the first step of converting the command line options.

### DIFF
--- a/ports/qemu/Makefile
+++ b/ports/qemu/Makefile
@@ -225,7 +225,7 @@ test_natmod: $(BUILD)/firmware.elf
 	$(eval DIRNAME=ports/$(notdir $(CURDIR)))
 	cd $(TOP)/tests && \
 		for natmod in btree deflate framebuf heapq random_basic re; do \
-		./run-natmodtests.py -p -d execpty:"$(QEMU_SYSTEM) $(QEMU_ARGS) -serial pty -kernel ../$(DIRNAME)/$<" $(RUN_TESTS_EXTRA) extmod/$$natmod*.py; \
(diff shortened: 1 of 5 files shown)


projectgus (approved):
Yay, consistency!

It seems to me that some of this very similar functionality could be combined into a single support module next, but this is a great improvement on its own!

PR #17954, opened 2025-08-19
tests: Improve test hardware configuration by putting board-specific settings in `target_wiring` module

dpgeorge:
### Summary

There are currently a few "hardware" tests that need external board connections to be made for them to work, eg bridging a pair of pins.  Eg all tests in `tests/extmod_hardware` need some kind of connection.

Along with the physical connections -- which are different for each board -- there needs to be corresponding Python settings, eg which UART instance to use and which pins for TX/RX.

These settings are currently hard-coded in each test file.  That has a few problems:
- settings are repeated, eg all the UART tests have pretty much the same settings code duplicated across them
- changing the settings means changing many files
- adding a new board means adding a lot of code
- the tests get bigger and bigger with each new board that they support, meaning they may not fit on targets with a small amount of RAM (that's already the case with `tests/extmod_hardware/machine_pwm.py`)
- if you have a custom board you have to manually edit the test to add your own settings

This PR aims to solve all the above problems by splitting the board-specific settings out into separate files, one for each board (or port).  They are placed in the `tests/target_wiring/` directory.  The `run-tests.py` test runner then loads the appropriate configuration for the target that is being tested, sends it to the board so it's available as `import target_wiring` (without needing a filesystem), and then executes the test (only tests that need `target_wiring` have it loaded).

This PR adds support for this `target_wiring` module, and converts all the existing `machine.UART` tests to use it.

### Testing

Tested locally on PYBV10 and ESP32-S2.  Still needs further testing once we decide this is the way forward.

Although the idea is straightforward, there were many decisions to make in the implementation:
1.  Selection of the correct `target_wiring` module follows a hierarchy: if an exact match for the board name is found (via `sys.implementation._build`) then that's used, otherwise there's a possible wildcard match on the board name (eg for PYB* boards, they can all use the same config), otherwise there's a match on the port name (eg rp2 boards can all use the same config).  This allows a lot of flexibility (maybe too much?).
2. I used a trick to send the `target_wiring` module to the board without using the filesystem or VFS hook.  It just evaluates a new class called `target_wiring` with the body of the class the script with the settings.  This should be pretty minimal and work on targets without a filesystem/VFS.
3. The same configuration is reused for `extmod/machine_uart*.py` tests ... they are not strictly hardware tests (they don't require external connections) but they do need to know which UART instance to use, and they can reuse the same instance used by `extmod_hardware/machine_uart*.py` tests.  Hopefully this won't be an issue, but it does confuse somewhat which tests require `target_wiring` configuration and which don't (ie it's no longer as simple as saying that only `extmod_hardware/` tests need `target_wiring`).
4. Eventually there may be other UART tests that need to be connected to a separate board.  As such I named the config options `uart_loopback_args` and `uart_loopback_kwargs`.  This shows the general problem that the approach may not scale well if lots of tests need subtly different configurations, eg `I2CTarget` will need a loopback configuration (4 pins) and a non-loopback configuration (another 2 pins).  To make it work, the config for these either need to be named differently, or selected very carefully so the pins can overlap.
5. Tests like `extmod_hardware/machine_pwm.py` can reuse the UART loopback pins, but probably they should be named differently, in case they can't be reused.
6. Need to think about how to handle zephyr boards, how to detect which configuration to use (need to either add special handling in `run-tests.py` for zephyr, or expose `sys.implementation._build` on that port). --> DONE, zephyr now has `sys.implementation._build`
7. To manually run a test using this new `target_wiring` module (without using `run-tests.py`) requires the user to explicitly copy across the correct `target_wiring.py` file to their device.
8. If there's a `target_wiring.py` on the device then it is overridden by the one injected by `run-tests.py`.  I think this is desired behaviour, so you don't have a stale/old file on there that messes up the test. --> CHANGED, any files on the board take precedence.
9. Need to add a way in `run-tests.py` for the user to select/override which `target_wiring.py` module to use, eg for custom boards, eg for Octoprobe. --> DONE, `--target-wiring` option added

### DIFF
--- a/tests/target_wiring/samd.py
+++ b/tests/target_wiring/samd.py
@@ -1,0 +1,7 @@
+# Target wiring for general samd board.
+#
+# Connect:
+# - D0 to D1
+
+uart_loopback_args = ()
+uart_loopback_kwargs = {"tx": "D1", "rx": "D0"}
--- a/ports/zephyr/CMakeLists.txt
+++ b/ports/zephyr/CMakeLists.txt
@@ -30,6 +30,7 @@ project(micropython)
 set(MICROPY_PORT_DIR    ${CMAKE_CURRENT_SOURCE_DIR})
 set(MICROPY_DIR         ${MICROPY_PORT_DIR}/../..)
 set(MICROPY_TARGET      micropython)
+string(TOUPPER          ZEPHYR_${BOARD} MICROPY_BOARD)
 
 include(${MICROPY_DIR}/py/py.cmake)
 include(${MICROPY_DIR}/extmod/extmod.cmake)
--- a/tests/extmod/machine_uart_irq_txidle.py
+++ b/tests/extmod/machine_uart_irq_txidle.py
@@ -10,32 +10,7 @@ except (ImportError, AttributeError):
     raise SystemExit
 
 import time, sys
-
-# Configure pins based on the target.
-if "alif" in sys.platform:
-    uart_id = 1
-    tx_pin = None
-elif "rp2" in sys.platform:
-    uart_id = 0
-    tx_pin = "GPIO0"
-    rx_pin = "GPIO1"
-elif "samd" in sys.platform and "ItsyBitsy M0" in sys.implementation._machine:
-    uart_id = 0
-    tx_pin = "D1"
-    rx_pin = "D0"
-elif "samd" in sys.platform and "ItsyBitsy M4" in sys.implementation._machine:
-    uart_id = 3
-    tx_pin = "D1"
-    rx_pin = "D0"
-elif "mimxrt" in sys.platform:
-    uart_id = 1
-    tx_pin = None
-elif "nrf" in sys.platform:
-    uart_id = 0
-    tx_pin = None
-else:
-    print("Please add support for this test on this platform.")
-    raise SystemExit
+from target_wiring import uart_loopback_args, uart_loopback_kwargs
 
 
 def irq(u):
@@ -46,11 +21,7 @@ text = "Hello World" * 20
 
 # Test that the IRQ is called after the write has completed.
 for bits_per_s in (2400, 9600, 115200):
-    if tx_pin is None:
-        uart = UART(uart_id, bits_per_s)
-    else:
-        uart = UART(uart_id, bits_per_s, tx=tx_pin, rx=rx_pin)
-
+    uart = UART(*uart_loopback_args, baudrate=bits_per_s, **uart_loopback_kwargs)
     uart.irq(irq, uart.IRQ_TXIDLE)
 
     # The IRQ_TXIDLE shall trigger after the message has been sent. Thus
--- a/tests/extmod/machine_uart_tx.py
+++ b/tests/extmod/machine_uart_tx.py
@@ -8,50 +8,35 @@ except ImportError:
     raise SystemExit
 
 import time, sys
+from target_wiring import uart_loopback_args, uart_loopback_kwargs
 
 initial_delay_ms = 0
 bit_margin = 0
 timing_margin_us = 100
 
-# Configure pins based on the target.
+# Tune test parameters based on the target.
 if "alif" in sys.platform:
-    uart_id = 1
-    pins = {}
     bit_margin = 1
 elif "esp32" in sys.platform:
-    uart_id = 1
-    pins = {}
     timing_margin_us = 400
 elif "mimxrt" in sys.platform:
-    uart_id = 1
-    pins = {}
     initial_delay_ms = 20  # UART sends idle frame after init, so wait for that
     bit_margin = 1
+elif "nrf" in sys.platform:
+    timing_margin_us = 130
 elif "pyboard" in sys.platform:
-    if "STM32WB" in sys.implementation._machine:
-        uart_id = "LP1"
-    else:
-        uart_id = 4
-    pins = {}
     initial_delay_ms = 50  # UART sends idle frame after init, so wait for that
     bit_margin = 1  # first start-bit must wait to sync with the UART clock
 elif "rp2" in sys.platform:
-    uart_id = 0
-    pins = {"tx": "GPIO0", "rx": "GPIO1"}
     timing_margin_us = 180
 elif "samd" in sys.platform:
-    uart_id = 2
-    pins = {"tx": "D1", "rx": "D0"}
     timing_margin_us = 300
     bit_margin = 1
-else:
-    print("SKIP")
-    raise SystemExit
 
 # Test that write+flush takes the expected amount of time to execute.
 for bits_per_s in (2400, 9600, 115200):
     text = "Hello World"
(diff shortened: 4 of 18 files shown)


dpgeorge:
@projectgus as we discussed, this isn't 100% complete but should be good enough for a first review.

@hmaerki this should solve a lot of our issues with Octoprobe and configuring the wiring for each board.

dpgeorge:
> 8\. If there's a `target_wiring.py` on the device then it is overridden by the one injected by `run-tests.py`.  I think this is desired behaviour, so you don't have a stale/old file on there that messes up the test.

Actually, maybe the other way around is better: if there's a `target_wiring.py` on the filesystem on the device then that should take precedence?  That way, you can set up your own boards for testing with custom wiring and run the tests as normal (without any special arguments) and they just work.

The alternative would be new CLI options to `run-tests.py` to override the behaviour, eg `run-tests.py --skip-target-wiring`.

projectgus (commented):
This approach looks very good and simple to me, cuts out a lot of complex code!

It also lowers the bar for adding more tests, I've hesitated to add small hardware tests in the past because it seems "expensive" to copy the big block of hardware config and pay the cost of maintaining it.

There is a bit of new user documentation we should probably repeat even though it's redundant, for example in each file where we have `# For UART loopback tests` we should be explicit about `# For UART loopback tests (connect these pins together)`.

> Actually, maybe the other way around is better: if there's a target_wiring.py on the filesystem on the device then that should take precedence? That way, you can set up your own boards for testing with custom wiring and run the tests as normal (without any special arguments) and they just work.

I think that's a good approach (test runner defaults to a module on the target filesystem if it's there). Two additional suggestions:

* `run-tests.py` should log a line i.e. `Using host file PATH as target_wiring` or `Using target_wiring module from target` so you never have to guess which file is being used.
* Add a command line option for `run-test.py --target-wiring PATH` so can also override the selection for custom boards or experiments without needing to copy anything to the target.

dpgeorge:
Thanks for looking at this!

> It also lowers the bar for adding more tests, I've hesitated to add small hardware tests in the past because it seems "expensive" to copy the big block of hardware config and pay the cost of maintaining it.

Yes, definitely will make it much easier to add new hardware-based tests.

> I think that's a good approach (test runner defaults to a module on the target filesystem if it's there).

OK, I'll implement that.


> * `run-tests.py` should log a line i.e. `Using host file PATH as target_wiring` or `Using target_wiring module from target` so you never have to guess which file is being used.

It actually already does log this info (which file is used from the host), it's in the first line that also includes other target info (eg arch, threading, unicode, float).  Do you think it should be a separate line?

(Also, I think I should make it so that info is only printed if it's going to run one of the tests that requires `target_wiring.py`.  That provides further feedback about what's actually going to be tested.)

dpgeorge:
Added a commit that:
- only detects `target_wiring` if needed, ie if it's going to run a test that needs it
- if `import target_wiring` passes on the target then don't inject anything, to let the target override it (eg via a file on its filesystem, or frozen into the firmware)
- print an explicit message on a separate line about which `target_wiring` is used

Example output when host file is used:
```
$  ./run-tests.py -t a0 extmod/machine_uart_*.py
platform=pyboard arch=armv7emsp inlineasm=thumb float=32-bit unicode
using target_wiring module from host file PYBx.py
skip  extmod/machine_uart_irq_txidle.py
pass  extmod/machine_uart_tx.py 
1 tests performed (3 individual testcases)
1 tests passed
1 tests skipped: extmod/machine_uart_irq_txidle.py
```
and when there's already a file on the target filesystem:
```
./run-tests.py -t a0 extmod/machine_uart_*.py
platform=pyboard arch=armv7emsp inlineasm=thumb float=32-bit unicode
using target_wiring module from target's filesystem
skip  extmod/machine_uart_irq_txidle.py
pass  extmod/machine_uart_tx.py 
1 tests performed (3 individual testcases)
1 tests passed
1 tests skipped: extmod/machine_uart_irq_txidle.py
```

projectgus:
> It actually already does log this info (which file is used from the host), it's in the first line that also includes other target info (eg arch, threading, unicode, float). Do you think it should be a separate line?

Oh sorry, I didn't see that was already in there when I reviewed the code. I think that's also fine, if you prefer to do it that way. It might actually be easier to read out quickly that way.

projectgus:
> (Also, I think I should make it so that info is only printed if it's going to run one of the tests that requires target_wiring.py. That provides further feedback about what's actually going to be tested.)

Fully agree!

dpgeorge:
Added a commit that puts the `target_wiring` information back on the first line with all the other target info.  (Easy to revert that change if we don't like it.)

dpgeorge:
I made two changes/additions:
1. The zephyr port now sets `sys.implementation._build` to `ZEPHYR_<uppercase zephyr board name>`, which allows zephyr boards to use the target_wiring scheme in the same way as other ports.
2. Added `--target-wiring` CLI option to `run-tests.py` to explicitly provide a file to use for the target_wiring module.

dpgeorge:
I've rebased this and retested, including testing on ESP32-C6 (UART tests pass).

The `target_wiring` files for each board have a small comment at the top saying what needs to be connected for that board to work.  Otherwise, I tried to keep these files minimal so they don't take too long to send across to the board.

This PR should be ready to merge now.

tests/target_wiring/samd.py
```diff
@@ -0,0 +1,7 @@
+# Target wiring for general samd board.
+#
+# Connect:
+# - D0 to D1
+
+uart_loopback_args = 2
```
robert-hh: The UART ID for the UART at D0 and D1 is not always 2. It's better if it can be omitted.

robert-hh: The fixed argument can here be replaced by `tuple()`, which will work for both ITSYBITSY_M0 and ITSYBITSY_M4 and other samd boards.

tests/target_wiring/samd.py
```diff
@@ -0,0 +1,7 @@
...
+uart_loopback_kwargs = {"tx": "D1", "rx": "D0"}
```
robert-hh, line 7: For SAMD and the default UART, these can as well be replaced by the empty dict: `uart_loopback_kwargs = {}`, unless you want to keep these for documentation.

hmaerki: I prefer to see the pin number. Rationale: To look up the defaults is not trivial and a source of errors.

tests/target_wiring/samd.py
dpgeorge: I did have specific target_wiring files for the ItsyBitsy boards.  But you are right, it's better to just omit this specific setting and use `()`, so the default is picked up.  Then we don't need specific settings for the ItysBitsy boards.

I updated the PR as such, and retested on ITSYBITSY_M0.

Thanks for the idea!

tests/target_wiring/samd.py
dpgeorge: I have left these explicitly included, to make it clear what is wired up (in addition to the comment which tells what to wire up).

PR #17963, opened 2025-08-20
docs/espnow: Add  parameters for `ESPNow`.

Josverl:
While researching a question I noticed that the documentation for ESPNow was missing the default parameters for the initialization.

@glenn20 Does this change make sense ?

### Testing

Cross checked :
- with the ESPNow source code 
- with the CircuitPython documentation of the same module.

### DIFF
--- a/docs/library/espnow.rst
+++ b/docs/library/espnow.rst
@@ -96,10 +96,16 @@ class ESPNow
 Constructor
 -----------
 
-.. class:: ESPNow()
+.. class:: ESPNow(buffer_size: int = 526, phy_rate: int = 0)
 
     Returns the singleton ESPNow object. As this is a singleton, all calls to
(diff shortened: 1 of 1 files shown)


dpgeorge:
> Cross checked :
> 
>     * with the ESPNow source code

Where did you see these parameters to the constructor?  Looking at `ports/esp32/modules/espnow.py` and `ports/esp8266/modules/espnow.py` their `ESPNow.__init__()` methods only take `self` as a parameter.

Josverl:
[Removing egg from face] 
@dpgeorge 
I think I have mixed up CircuitPython with Micropyhon snippets/repos 
Tracking back though history I was not able to track back what my mistaken inspiration was - but CP is the most likely.

Sorry for causing confusion.

dpgeorge:
No worries!

